PR #1193, opened 2026-06-30
feat(sidecar): side-19 — Code2Wav timing/bench instrumentation + compile scaffold (won't-ship, deferred)

### LINKED ISSUE #988: side-19 — Faster book rendering: torch.compile the Qwen Code2Wav decoder
**A speed play for our users.** Qwen is our default generation engine, so every audiobook render runs through it. This spike compiles the **Code2Wav** codec decoder (the token→waveform stage of `Qwen3-TTS-12Hz-0.6B`) with `torch.compile` to shave time off **batched chapter rendering** — the goal is **long books finishing faster than playback by a wider margin**, so a multi-hour audiobook is ready sooner and generation feels snappier. Low-risk (a self-contained feed-forward module, not a fork), with the one-time compile warmup amortized to nothing over hours of audio.

### What

Speed up batched book rendering on the default Qwen engine by `torch.compile`-ing the Code2Wav decoder (`server/tts-sidecar/main.py:156`). Ship it env-gated (`QWEN_COMPILE_CODEC`), batch-path-only, default-OFF on Windows — once we've measured the gain is real.

**Distinct from the parked `side-7`** (plan 129): `side-7` targets the autoregressive LM decode loop (a correctness-risky `qwen_tts` fork). `side-19` targets the self-contained Code2Wav `nn.Module` — no KV-cache semantics, no generate-loop fork, much lower risk. `side-19` does **not** reopen `side-7`.

### Acceptance

Measurement-first — size the win before writing production code:

- **Phase 0 (the gate):** measure Code2Wav's share of `generate_voice_clone` batch wall-time at `QWEN_BATCH_SIZE=32` on the 8 GB box. `> ~25–30%` → a felt speedup is on the table, build Phase 1; `~10–25%` → ship only if Phase 1 stays cheap + VRAM-neutral; `< ~10%` → the codec isn't where the time goes, bank the measurement and look elsewhere.
- **Phase 1 acceptance (all must hold):** (1) a live `POST /api/books/:id/generation` chapter render measurably faster flag-ON vs OFF; (2) 8 GB VRAM-neutral at batch-32 (else disabled on 8 GB); (3) output-preserving within the golden-audio tolerance (`--engine=qwen`) — a speedup must never change what the listener hears; (4) no interactive regression — single `/synthesize` preview latency unchanged.

### Key files

- `server/tts-sidecar/main.py` — `QwenEngine`, `synthesize_batch` / `generate_voice_clone`, Code2Wav decode (`:156`), `_apply_torch_perf_flags`.
- `server/tts-sidecar/scripts/bench-tts.py` — Phase-0 sizing instrument + Phase-1 A/B harness.
- `server/tts-sidecar/tests/test_compile_codec.py` (new) — flag default OFF, OFF on Windows, batch-path-only wiring, compile-failure→eager fallback.

### Depends on

None hard. Sibling context: `side-7` (parked, distinct stage).

### Benefit

_(user):_ long books finish faster than playback by a wider margin on the engine every render uses — snappier generation, with no interactive-path or audio-quality cost — once Phase 0 confirms the win is real.

Spec: `docs/superpowers/specs/2026-06-22-qwen-codec-compilation-design.md`.

dudarenok-maker:
## Summary

Full side-19 spike, resolved: **Phase 0 shipped (GO); Phase 1 built + reviewed, but fails its on-box acceptance gate — closing won't-ship (deferred).**

- **Phase 0 (Tasks 1-2, shipped and live):** `QWEN_CODEC_TIMING` env-gated timing hook + `bench-tts.py --code2wav-share` mode. Measured Code2Wav's share of batched Qwen 0.6B forward-compute at **median 67.3%** on the 8GB box — a real, majority-of-compute cost, not a minor tail.
- **Phase 1 (Tasks 4-5, shipped as tested dead code):** `QWEN_COMPILE_CODEC` compiles the Code2Wav decoder's `forward` and swaps it in per-batch, for both the 0.6B and 1.7B tiers (scope extended past the original plan once the 67.3% result came in). An on-box discovery pass corrected the plan's assumed swap target (two levels deeper than assumed) and its mechanism (swap the `forward` *attribute*, never the module object — a module-level swap would have been a silent no-op). Fully unit-tested and independently reviewed against that failure mode.
- **Task 6 (on-box acceptance) — Gate 1 FAILS:** the compiled path is a severe regression, not a speedup. A batch-32 call didn't finish in 26+ minutes (vs ~100s eager); a batch-2 call didn't finish in 3+ minutes. Verified NOT environmental (several real environment snags came up and were fixed first, independently of this result — see the spec's Ship notes). Direct evidence for the finding: zero `TORCH_LOGS=recompiles` messages and no inductor-cache growth (rules out recompile thrash), yet the CPU-bound worker never finished. Root cause not fully isolated.

**Disposition:** `QWEN_COMPILE_CODEC` ships default-OFF everywhere (additionally hard-OFF on Windows) — harmless, tested, reviewed dead code. Zero production risk either way. Issue #988 closed not-planned; full writeup + revisit conditions in the spec's Ship notes (`docs/superpowers/specs/2026-06-22-qwen-codec-compilation-design.md`) for whoever picks this up next.

## Test plan

- [x] `server/tts-sidecar/tests/test_codec_timing.py` (8 tests) + `test_compile_codec.py` (8 tests) — all GPU-free, all pass.
- [x] Full filtered sidecar suite (`pytest -m "not golden"`) green.
- [x] `npm run verify` green.
- [x] On-box Phase 0 measurement: median 67.3% Code2Wav share (8GB RTX 4070 Laptop, both tiers' codec confirmed identical class/structure).
- [x] On-box Task 6 Gate 1: FAILED with direct, non-environmental evidence (see Ship notes) — closing the spike here rather than proceeding to Gates 2-4.

Closes #988


dudarenok-maker:
Phase 0 measurement landed (commit 54000e4): median **67.3%** Code2Wav share of batch forward-compute time (8 GB RTX 4070, 3 post-warmup runs, voice `rv1`) — well above the `>~25–30%` build threshold. **GO Phase 1.** Full numbers + M1 verification in the spec's Ship notes (`docs/superpowers/specs/2026-06-22-qwen-codec-compilation-design.md`) and `docs/tts-performance.md` lever 5.

dudarenok-maker:
Phase 1 implementation landed (commits \`2cd0a655\`, \`23c44b8c\`): \`QWEN_COMPILE_CODEC\` compiles the Code2Wav decoder's \`forward\` and swaps it in per-batch, for **both** the 0.6B and 1.7B tiers (scope extended from the original plan given the 67% measured share).

Task 4's on-box discovery corrected two things the plan assumed: the compile target is \`speech_tokenizer.model.decoder\` (two levels deeper than assumed), and the decoder runs on **CPU** on both tiers — so the swap targets \`decoder.forward\` as an instance attribute, not the whole module object (a whole-module swap would have been a silent no-op via \`OptimizedModule\`'s attribute delegation). Full writeup in the spec's Ship notes.

Remaining: **Task 6** — the four on-box acceptance gates (live A/B render, 8GB VRAM-neutral with both tiers resident, golden-audio tolerance, no interactive-path regression). Flag stays default-OFF until those pass.

dudarenok-maker:
Reviewed at medium effort (code-review skill). Scope is docs + default-OFF/Windows-hard-OFF dead code, so production behavior is inert by construction. Verified the two real-code risk areas:

- **Forward-swap serialization** — `_codec_compiled_for_batch` sits *inside* `with self._synth_lock:` in both the 0.6B and 1.7B branches and wraps the whole `if live_instruct / else` block; the single `/synthesize` path is under the same lock. The `decoder.forward` swap/restore is fully serialized — no concurrent-batch corruption, no single-vs-batch race.
- **Bench contract** — `run_code2wav_share` builds `{voice, text}` items matching the existing callers + `synth_batch_once`'s 5-tuple return, and coerces the `--batch` default of 0 → 32 so the batch is never empty.

No actionable correctness findings. The one real compile trap (`OptimizedModule.__getattr__` silent-no-op) is documented, unit-tested against, and avoided via the forward-attribute swap. Docs accurately record the negative Gate-1 result. LGTM to merge.

PR #1376, opened 2026-07-06
fix(sidecar): scope the free-VRAM floor guard to non-torch cards

### LINKED ISSUE #1375: Sidecar VRAM free-floor guard false-positives on the torch card — mid-chapter recycle storms pause the queue
**Observed (2026-07-06):** every sidecar recycle that day (02:37→13:13) was the per-card `driver_free_floor` guard firing on the main torch card during routine 1.7B batch peaks. On a single 8 GB laptop card, torch-reserved ~6 GB + CUDA context + display compositor legitimately leave <1024 MB driver-free at a batch peak; the guard read that as critical OOM and hard-recycled the sidecar mid-chapter. Two trips on one chapter (ch23, *Unraveled*) exhausted the in-loop recovery, tripped the RECYCLE STORM breaker, and **paused the generation queue** — killing the overnight run. The trip log printed `driver_free_floor 0MB crossed the restart ceiling 0MB` (format args never wired), which hid the pattern; the self-exit also reached the supervisor as `code=1` via the server's forced-recycle kill, so the events read as crashes.

**Root cause:** the floor is documented as the OOM guard for ORT/CT2-only cards (Kokoro/Whisper VRAM is invisible to torch's reserved metric), but `_check_per_card_ceilings` applied it to every card, including the torch-managed one whose OOM protection is already the per-card reserved-VRAM ceiling.

**Fix:** scope the floor to cards torch is not actively managing (torch-reserved < 256 MB, or torch stats unreadable); wire the real metric/threshold values into the trip log.

**Mitigation while un-merged:** Advanced Settings → GPU lifecycle → "Per-card VRAM free floor (MB)" → 0 (applied on this box 2026-07-06 13:05).

dudarenok-maker:
## Summary

Every sidecar recycle on 2026-07-06 — 02:37 through 13:13, including the two exits the supervisor logged as `code=1` crashes (the server's forced-recycle kill after chapter 23's storm) — was the per-card `driver_free_floor` guard misreading a routine 1.7B batch peak as OOM. On a single 8 GB laptop card, torch-reserved (~6 GB) + CUDA context + the display compositor legitimately leave under the 1024 MB floor at batch peaks; each false trip drained and killed the sidecar mid-chapter, and two on one chapter tripped the recycle-storm breaker and **paused the whole generation queue** — the direct cause of the overnight-run interruptions.

The floor exists as the ONLY OOM guard for ORT/CT2-only cards (Kokoro/Whisper VRAM is invisible to torch's reserved metric — its own docstring), but `_check_per_card_ceilings` applied it to every card:

- A card with torch-reserved ≥ `_TORCH_ACTIVE_RESERVED_MB` (256 MB) is now treated as torch-managed and **skips the floor** — its OOM protection remains the per-card reserved-VRAM ceiling, unchanged.
- A card whose torch stats are unreadable (ORT/CT2-only deployment) keeps the floor — the guard's original purpose survives.
- The breach dict now carries the breaching reading + limit, so the trip log names real numbers. It previously printed `driver_free_floor 0MB crossed the restart ceiling 0MB`, which masked that every recycle that day was the same guard.

Live mitigation already on this box: `sidecar.vramFreeFloorMb=0` override applied via `PUT /api/config` at 13:05 (to be reset to default after this merges).

## Test plan

- `test_per_card_ceilings.py`: new `floor_skipped_on_torch_active_card` (the false-positive regression — fails before this fix, passes after), new `floor_applies_when_torch_unreadable` (ORT-only guard preserved), and the existing floor/ceiling cases extended to pin the new `metric_mb`/`threshold_mb` fields.
- Full sidecar pytest suite green on-box (503 passed); `npm run verify` green pre-push.

Closes #1375


dudarenok-maker:
**Mandatory pre-merge code review (adversarial pass, medium effort) — PASSED, findings folded.**

- No merge-blocking defect. Verified clean: breach-dict extra keys are safe through the Node breadcrumb consumer (`restart-breadcrumb.ts` reads `card` as `unknown`); MB units consistent across floor/ceiling comparisons; the new torch-active test genuinely fails on the old code; no other call sites.
- **Folded (latest commit):** (1) in-code acknowledgment of the deliberate tradeoff — non-torch VRAM growth on a torch-active card is now unguarded (acceptable: the floor could never guard that mix without the false-positive storm; co-resident ORT engines are ~1 GB bounded and idle-evicted); (2) trip-log verb made direction-neutral (`breached the restart limit`) — the old `crossed the restart ceiling` read backwards for a fall-below-floor trip.
- Known unrelated fragility hit while verifying: #1377 (watchdog tests sample the real GPU) — pre-existing, filed separately.

Sidecar suite (the leg cloud CI can't run — venv-gated) green locally on this branch, including post-fold.

Issue #399, opened 2026-06-01
side-11 — Eliminate the variable-input-shape host-memory leak (so recycling isn't needed)


dudarenok-maker:
The Qwen generation forward leaks committed-private host RAM monotonically: every sentence is a different length → a new native per-shape workspace that is never freed (climbs unbounded on variable-length generation, flat on fixed shapes; CUDA flat until it spills — pytorch/pytorch #32596). Measured 2026-06-01: **~1,150 MB/batch** committed on the variable-shape bench. This forces plan 143's process-recycle (~every 10 chapters on a long book), which is disruptive. **Goal:** a full book on one warm sidecar with no recycles and no dropped chapters — the cleanest end-to-end win now that RTF is solved (~1.04).

**Already contained (mitigation, not the fix):** the process-recycle (plan 143/152) bounds it; the load-failure VRAM reclaim (plan 155) stops *failed reloads* orphaning VRAM (that orphaned-CUDA-tensor facet is closed); the **reload-doubling VRAM facet is closed** by [plan 161](features/161-sidecar-vram-reload-recycle.md) — `unload()` now takes `_synth_lock` so a mid-synth `/unload`+`/load` can't stack a second model copy past the card (which had spilled into the Windows sysmem fallback → 100%-util stall), and the watchdog gained a reserved-VRAM recycle tier (host-keyed recycling was blind to VRAM). Note this is a SEPARATE pressure from the host committed-RAM leak below.

**Tried & rejected:** candidate 1 — `SIDECAR_DISABLE_MKLDNN=1` — **rejected** by the 2026-06-01 A/B: it cripples synth ~15×+ (MKLDNN is load-bearing for the Qwen CPU audio path), so it's non-viable regardless of leak effect. Flag stays default OFF. See [plan 153](features/153-sidecar-variable-shape-host-leak.md).

**Open — what's left to look at (priority order):**

1. **Candidate 2 — fixed-shape batch padding** (the main remaining lever). Pad to a small set of fixed width × max-len buckets so shape variety collapses → the native per-shape workspace is reused. Re-tune plan-128 bucketing + a python-side pad in `synthesize_batch` (the Node packer alone can't set the tensor shape). Carries an RTF/padding cost — measure on the `bench-tts.py --mem-sample` harness; target is to beat the ~1,150 MB/batch OFF slope toward ≈ flat. Its own branch.
[...]

PR #1378, opened 2026-07-06
fix(sidecar): bound the oneDNN primitive cache — the variable-shape host-RAM leak (side-11)

dudarenok-maker:
## Summary

Root-causes and bounds the side-11 variable-input-shape host committed-RAM leak (#399). The Code2Wav codec runs on CPU; oneDNN JIT-compiles a conv kernel + scratchpad per distinct input shape and caches one entry PER SHAPE (default capacity 1024, never freed by `gc.collect()`/`empty_cache()`) — so variable-length generation leaked committed-private RAM monotonically (~1,150 MB/batch on the plan-153 bench) until a process-recycle reset it.

- `main.py` pins `ONEDNN_PRIMITIVE_CACHE_CAPACITY=64` at module top (`setdefault`, before torch's first CPU conv — the plan-144 `PYTORCH_CUDA_ALLOC_CONF` pattern; an operator `.env` override wins). MKLDNN stays fully enabled — identical kernels, bit-identical audio; this is NOT the rejected plan-153 `SIDECAR_DISABLE_MKLDNN` (which was 15×+ slower).
- `/debug/memory` gains `cuda.host_pinned_owned_mb` / `host_pinned_active_mb` (torch ≥ 2.6 `host_memory_stats()`; guarded, keys omitted when unavailable) — the attribution instrument used to rule pinned staging in/out.

**Measured, isolated repro (torch 2.11.0+cu128):** committed growth +592 MB → +59 MB over 120 variable-shape conv iterations, per-iteration wall time unchanged (54 → 55 ms).

**Measured, real model (1.7B, `bench-tts.py --mem-sample`, seeded 15×16-item batches, 2026-07-06):**

| | cap=64 | cap=0 | unfixed baseline (plan 153) |
|---|---|---|---|
| committed slope | +91 MB/batch | +55 MB/batch | ~+1,150 MB/batch |
| codec decode | ~38 s/batch | ~43 s/batch (+13%) | — |
| pinned host | 0.0 | 0.0 | — |

64 is the shipping default: ~92% slope reduction at measurably better throughput than 0. A separate ~24 GB committed *plateau* remains (heap-retained CPU-codec activations — NOT the primitive cache, NOT pinned memory: identical at cap=0, pinned 0.0 throughout) — that facet stays open on #399 with codec-decode-on-GPU as the follow-up direction (also reclaims the ~36 s/batch CPU decode share).

With this cap plus the floor-guard fix (separate PR), the host soft-recycle fires as a clean chapter-boundary recycle roughly every 10+ chapters instead of the leak forcing constant resets.

Refs #399

## Test plan

- `test_memory.py`: new `test_onednn_primitive_cache_capacity_default` (fails if the setdefault line is removed) and `/debug/memory` shape test extended to pin the pinned-host keys on a CUDA box.
- Full sidecar pytest suite green on-box; `npm run verify` green pre-push.
- Live A/B numbers above (leak-slope bench + codec timing), audio path untouched (identical kernels — no quality delta by construction).


dudarenok-maker:
**Mandatory pre-merge code review (adversarial pass, medium effort) — PASSED, findings folded.**

- No correctness or blocking defects. Verified: setdefault runs before any torch import (torch is function-local throughout main.py); the bundled oneDNN in `torch_cpu.dll` reads `ONEDNN_PRIMITIVE_CACHE_CAPACITY` at cache init; `host_memory_stats()` key names confirmed against torch 2.11 source AND a live pinned H2D copy; the cap affects caching only, never kernel selection — the identical-audio claim holds; `.env` override path confirmed end-to-end through `buildSidecarEnv`.
- **Folded (latest commit):** two LOW test-hardening items — the pinned-stats test now mirrors the endpoint's call-success guard exactly, and the capacity-default test fails cleanly (not errors) on a garbage pre-set env value.
- Measured A/B justification in the PR body (cap=64 vs cap=0 vs unfixed, real 1.7B model).

Sidecar suite (the leg cloud CI can't run — venv-gated) green locally on this branch, including post-fold.

PR #1418, opened 2026-07-07
fix(frontend): scope Voice Drift Detector to book/series, fix audio-element crash

### LINKED ISSUE #1417: Voice Drift Detector crashes the browser when drift spans many books (375 chapters / 10 books)
## What happened

Opening the Voice Drift Detector modal with drift accumulated across many books (reported: "375 chapters flagged across 10 books") freezes or crashes the browser tab.

## Root cause

Two compounding issues, both in the frontend:

1. `layout.tsx` fed `DriftReportModal` the entire workspace's drift (`selectDriftGroupsByBook`, unscoped) rather than the active book — the background cross-book poller (plan 83) deliberately fills the store with drift for every book, but the modal rendered all of it at once with no book/series scoping.
2. `DriftListenWidget` mounted two real `<audio>` elements per chapter row **unconditionally**, regardless of whether "Listen" was even clicked. At the reported scale that's up to ~750 concurrently-live `HTMLMediaElement`s — the actual crash trigger.

## Fix

- Default the modal to the active book, with a "This book" / "Series" toggle to expand scope (`ui.driftReportScope`, `scopeDriftGroupsByBook`, `findSeriesBookIds`).
- Rebuild `DriftListenWidget` on the shared-singleton A/B playback pattern already used elsewhere in the app (`use-sample-playback.ts` + `use-ab-audition.ts`) so at most one real `<audio>` element exists app-wide.
- Add a windowed/incremental reveal (20 cards per page, scroll-triggered) so even one heavily-drifted book can't dump hundreds of cards into the DOM at once.

See regression plan `docs/features/archive/245-drift-report-book-scope-and-pagination.md`.

dudarenok-maker:
## Summary

- The Voice Drift Detector rendered the entire workspace's drift at once (the plan-83 background cross-book poller fills it for every book) instead of scoping to the active book — fixed with a new `scopeDriftGroupsByBook` + `findSeriesBookIds` pair and a "This book" / "Series" toggle (`ui.driftReportScope`). No new server endpoint needed; both scopes' data are already client-resident.
- `DriftListenWidget` mounted two real `<audio>` elements per chapter row **unconditionally** (not gated on the Listen click) — at the reported scale ("375 chapters flagged across 10 books") that's up to ~750 concurrently-live `HTMLMediaElement`s, the actual browser-crash trigger. Rebuilt on the shared-singleton A/B playback pattern already used by `CompareCastModal`/`VoiceCompareModal` (`use-sample-playback.ts` + `use-ab-audition.ts`), so at most one real `<audio>` element exists app-wide.
- Added a windowed/incremental reveal to `DriftBookSection` (20 cards per page, IntersectionObserver sentinel — same technique as `change-log.tsx`'s infinite scroll) so even one heavily-drifted book can't dump hundreds of cards into the DOM at once.

Regression plan: `docs/features/archive/245-drift-report-book-scope-and-pagination.md`.

## Test plan

- [x] New unit tests: `scopeDriftGroupsByBook` (revisions-slice), `findSeriesBookIds` (library-slice), `driftReportScope` reducer (ui-slice).
- [x] `drift-report.test.tsx`: scale-regression suite (zero `<audio>` elements at 50-row scale, windowed reveal + sentinel behavior), book/series toggle suite, and the Listen A/B player suite rebuilt against the shared-singleton mock.
- [x] `layout.test.tsx`: the plan-91 book-title-fallback test now lands on an active book and asserts default book-only scoping + the Series toggle bringing a second book into view.
- [x] `npm run verify:fast:branch` (lint, typecheck, full frontend test suite, build) — all green.

Closes #1417.


dudarenok-maker:
Ran the mandatory `/code-review` pass (medium effort, 8 finder angles + 1-vote verify). 4 CONFIRMED findings, all fixed in the follow-up commit:

- **Pagination reset bug** — `DriftBookSection`'s reveal-page reset fired on any same-book group-count change (dismiss/regen/background poll), not just a book switch, silently collapsing the user's scroll progress back to page 1. Removed the reset effect entirely — the parent already keys the section by `bookId`, so a real book switch remounts a fresh instance anyway.
- **Unmount race on the voice-sample fetch** — the rebuilt Listen widget dropped the old ref-based "still mounted" guard, so a `getVoiceSample` resolve after the row was dismissed/regenerated could start real audio on the shared singleton with nothing left to stop it. Added a `mountedRef` check.
- **Cross-row false "Pause" state** — two characters drifting on the same chapter shared an identical chapter-audio match URL against the one shared playback singleton, so playing one row falsely flipped an unrelated row to "Pause chapter" too. Disambiguated with a per-character query param (server ignores it, same file served).
- **Test-coverage gap** — `driftReportScope` was missing from `persist-config.test.ts`'s transient-fields enumeration, unlike its sibling `driftReportCharacterFilter`.

3 new regression tests + 1 fixed assertion lock these in; full suite (3612 tests) + typecheck + lint + build all green. Two lower-severity architectural notes (selector-level scoping depth, IntersectionObserver sentinel duplication vs `change-log.tsx`) were surfaced as PLAUSIBLE but left as-is — design tradeoffs, not bugs, and out of scope for a surgical fix.

PR #1443, opened 2026-07-07
refactor(server): audition centroid now cost-bounded and anchor-aware, not fixed 12x

### LINKED ISSUE #1386: srv-36 — audition centroid: replace 12x-same-short-quote backup with a better reference sample
## Context
The Option-B audition centroid (`server/src/audio/render-integrity/audition-centroid.ts`, srv-36 Task 10) builds a character's voice-consistency reference by rendering the SAME short audition line **K=12 times** and ECAPA-embedding each. This is wasteful and low-signal:

- 12 full TTS synths per too-thin/bimodal character, done purely to obtain embeddings ("full model for a minor thing").
- All 12 renders speak the **same short quote** (often a character's longest evidence line, which can be very short, or the canned fallback), so the 12 embeddings sample almost the same acoustic content — little added diversity for the centroid.

## Idea
Replace the 12x-same-short-quote approach with a better backup reference, e.g.:
- Render a **single, long-enough prose sample** once (varied phonetic content → a more representative embedding) instead of 12 short identical ones; or
- **Reuse the character's existing in-book render embeddings** where available (the anchors) rather than re-synthesising; or
- Reduce K and/or vary the text across renders.

## Related
Follow-up from the OOM root-cause fix (audition rendered on a hardcoded 0.6B tier) — see the fix PR for this issue's parent context.

dudarenok-maker:
## Summary
- Replaces the srv-36 Option-B audition-centroid fallback's fixed K=12-identical-render loop with a smaller (target 6, hard cap 8), anchor-aware, text-diverse pool.
- Threads cast.json evidence into the fallback for the first time in production (previously silently undefined, no-oping the existing duration-floor retry).
- Too-thin characters' real in-book anchor embeddings are now reused; bimodal characters correctly get none.

Design spec (3 rounds of adversarial review): `docs/superpowers/specs/2026-07-07-srv36-audition-centroid-reference-design.md`

## Test plan
- [x] `audition-centroid.test.ts` rewritten — 17 cases covering pool-fill math, per-slot retry semantics, text cycling, the bimodal-blend cost-cap regression, and the round-3 shortfall regression.
- [x] New `aggregate-audition-pool.test.ts` — 5 cases covering hint threading (match/no-match/missing-cast.json) and the too-thin/bimodal anchor split (via a mocked `auditionCentroid`).
- [x] New `aggregate-audition-pool-real.test.ts` — 1 case exercising the REAL (unmocked) `auditionCentroid` end-to-end, proving blended anchors actually drive the resulting `cleanMean` — the mocked test file above can't prove this on its own.
- [x] Existing `aggregate.test.ts` + `aggregate-audition-tier.test.ts` unaffected (both have fixtures with no cast.json on disk — confirms the best-effort read doesn't regress them).
- [x] `npm run verify:fast:branch` green.

Closes #1386


dudarenok-maker:
Reopening to re-trigger title-lint after shortening the title to fit the 100-char subject limit.

Issue #1447, opened 2026-07-07
chore(server): byline-author-guard doesn't catch third-party names in front-matter chapters


dudarenok-maker:
## Context

Follow-up to #1446 (the immediate safety-net fix already shipped: an orphaned `characterId` now degrades to narrator instead of hard-failing synthesis). This issue tracks the deeper root cause that safety net papers over.

## What

`server/src/analyzer/byline-author-guard.ts` drops **the book's own** byline author from a chapter's detected roster before stage-2 attribution (issue #938) — unless the chapter title matches `AUTHOR_NOTE_TITLE_RX` (`от автора`, `предислови`, `послеслови`, `об авторе`, "author's note", etc.).

It has no concept of **other real, unrelated people named/quoted in front matter** — e.g. a foreword chapter that discusses or credits a different real author. Real incident: book "Юный дрессировщик" (author: Лидия Ивановна Острецова) has chapter 3 "Вступительная статья" (title doesn't match the author-note regex), an essay about a different real author, Radiy Pogodin. Stage-2 attribution cast him as a one-line "speaker," and `fold-minor-cast.ts`'s `proseTagged` carve-out (#537, protects legitimately quote-tagged speakers from being folded away) let the bogus entry survive as an orphaned cast reference.

## Why not fixed alongside #1446

Any fix here needs care: a chapter that's a **legitimate** framed narrative with a real quoted/named character (e.g. an in-fiction letter, a genuine framed author's-note under a title the current regex doesn't happen to match) shouldn't have that character incorrectly stripped. This is a scoping/design question, not a one-line patch — needs its own pass to define the detection heuristic (e.g. front-matter-chapter position + quote-tag pattern + name not appearing anywhere else in the story body) without over-broadening `AUTHOR_NOTE_TITLE_RX` in a way that swallows real narrative chapters.

## Acceptance (draft — refine at scoping time)

- A front-matter/foreword-ish chapter that names/quotes a real third-party person doesn't roster them as an attributable speaker.
- A genuinely framed narrative chapter with a real quoted character (Unraveled-style author's-note case, #938) is unaffected.
- Paired regression test using a fixture shaped like the "Юный дрессировщик" ch3 case.

Refs #1446.

PR #1448, opened 2026-07-07
fix(server): degrade orphaned characterId to narrator instead of hard-failing synthesis

### LINKED ISSUE #1446: bug(server): orphaned characterId hard-fails chapter synthesis instead of degrading
## What

A sentence group in `manuscript-edits.json` can reference a `characterId` with no corresponding entry in `cast.json` at all — not "known character, no voice designed" (which is a real, actionable error), but a genuinely orphaned id the user has never seen in Cast view. `synthesise-chapter.ts` threw `MissingDesignedVoiceError` naming the raw id, which is neither visible nor fixable from the UI.

## Repro (real incident)

Book "Юный дрессировщик" (author: Лидия Ивановна Острецова). Chapter 3, "Вступительная статья" (a foreword discussing a different, unrelated real author — Radiy Pogodin), produced one sentence attributed to `characterId: "radiy-pogodin"`. That id was never added to `cast.json` — two existing safeguards have blind spots here:

- `server/src/analyzer/byline-author-guard.ts` only strips **the book's own** byline author from a chapter's roster; it has no concept of *other* real people's names appearing in front matter.
- `server/src/analyzer/fold-minor-cast.ts`'s `proseTagged` carve-out (#537) deliberately protects quote-tagged names from being folded/dropped, so a legitimate tagged speaker doesn't lose their voice slot — the same protection let this bogus one survive.

Generation then hard-failed with:
> Character "radiy-pogodin" has no designed voice for this ru book — design a voice for it (and the narrator) in the cast view before generating.

## Fixed (this repo, `fix/server-synth-unknown-character-fallback`, commit 311f3a68)

`synthesise-chapter.ts`'s `castById` miss now falls back to the narrator's voice for that line and logs a warning once per offending id, instead of hard-throwing. Regression tests added.

## Follow-up (not yet done — scoping needed)

The deeper analyzer gap remains: `byline-author-guard.ts` doesn't catch a third-party real name quoted/discussed in a non-"author's-note"-titled front-matter chapter (e.g. "Вступительная статья" doesn't match `AUTHOR_NOTE_TITLE_RX`). A fix needs care — a chapter that's a *legitimate* framed narrative with a real quoted character shouldn't have that character stripped. Left for a separate pass.

dudarenok-maker:
## Summary
- A sentence group can reference a `characterId` with no entry in `cast.json` at all — e.g. a stray attribution to a real-world person quoted in a book's foreword, or a cast/sentence-data drift from an interrupted subset re-analysis. This is different from "known character, no voice designed": the id doesn't exist in Cast view, so `MissingDesignedVoiceError` named a character the user could never find or fix.
- Real incident: book "Юный дрессировщик" — chapter 3 "Вступительная статья" is a foreword about a different, unrelated real author (Radiy Pogodin). One line got attributed to `characterId: "radiy-pogodin"`, which was never added to `cast.json` (two existing safeguards, `byline-author-guard.ts` and `fold-minor-cast.ts`'s `proseTagged` carve-out, each have a blind spot here). Generation hard-failed and named an id invisible in the UI.
- `synthesise-chapter.ts`'s `castById` miss now falls back to the narrator's voice for that line and logs a warning once per offending id, instead of hard-throwing.
- The deeper analyzer gap (byline-author-guard not catching third-party front-matter names) is tracked separately — Refs #1447, needs its own scoping pass to avoid stripping a legitimate framed narrative's real quoted character.

## Test plan
- [x] New regression tests in `synthesise-chapter.test.ts`: orphaned `characterId` falls back to the narrator voice + warns once; a *known* undesigned character still throws `MissingDesignedVoiceError` (unaffected).
- [x] Confirmed the new test fails without the fix (`git stash` the fix, re-run — voice resolved to a different, wrong voice instead of narrator's) and passes with it.
- [x] Full server suite green (360 files / 4194 tests), lint clean, typecheck clean.
- [x] `npm run verify:fast:branch` green (ran via pre-push hook).

Closes #1446.


dudarenok-maker:
Independent code-review pass (workflow-backed, high effort / medium level) found two real issues in the initial version, both now fixed:

1. **Correctness** — if a book's narrator itself has no designed Qwen voice AND a chapter's only narrator exposure is via the orphaned-characterId fallback, the thrown `MissingDesignedVoiceError` named only "narrator," losing the trail back to the actual orphaned id. Fixed: the error now carries an explicit note naming the original orphaned characterId in that case. (`ec397a35`)
2. **Cleanup** — the synthetic narrator-stub literal (`{ id: narratorCharacterId, name: 'Narrator' }`) was duplicated between the pre-existing title-beat resolution and the new fallback. Hoisted into a shared `resolveNarratorChar()`. (`ec397a35`)

Known, accepted narrow limitation (not fixed here, out of scope for this PR): `generation.ts`'s pre-flight "narrator needs a voice" check computes `speakingIds` from raw sentence `characterId`s, so it can't see a narrator substitution that only happens inside `synthesise-chapter.ts`. In the rare case above, the failure surfaces mid-chapter-processing instead of the friendlier upfront preflight message — but the error text is now traceable to the real cause either way. Left as a follow-up if it proves worth closing.

Full server suite (360 files / 4195 tests), lint, and typecheck all green after the follow-up fixes.

PR #1567, opened 2026-07-13
fix(frontend,sidecar): model install/repair — sanitize pip constraints + legible dark-mode error

### LINKED ISSUE #1566: Coqui/Qwen model install fails: 'Constraints cannot have extras'; error text illegible in dark mode
## Two bugs on the Model Manager add-on install/repair surface

Hit while clicking **Repair** on Coqui XTTS v2 (Optional add-ons), in dark mode.

### 1. Install/repair aborts: `ERROR: Constraints cannot have extras`
`install-coqui.mjs` runs `pip install coqui-tts -c requirements/base.txt`, but `base.txt` is a *requirements* file that carries extras (`uvicorn[standard]`), which pip forbids in a `-c` **constraints** file. The install fails instantly and can't be retried past it. Same latent bug at 3 more `-c base.txt` sites in `install-qwen3.mjs` (`qwenPipInstallArgs` + both flash-attn installs).

**Repro (real sidecar venv):**
```
python -m pip install --dry-run --no-deps --no-index -c requirements/base.txt pip
→ ERROR: Constraints cannot have extras
```

### 2. The failure text is illegible in dark mode
The five in-app install affordances render their error body as `text-rose-900/80`, but the `[data-theme='dark']` override in `styles.css` remapped only `.text-rose-900`, `/85`, and `/70` — the `/80` alpha was missed. So a failed install paints dark-maroon-on-dark-maroon; the header (bare `text-rose-900`) stays readable but the actual error detail is unreadable.

### Fix
- Add a shared `pip-constraints.mjs` sanitiser that strips `[extras]` into a temp constraints file; route all four `-c base.txt` sites through it.
- Add `.text-rose-900\/80` to the dark-mode override group in `styles.css`.

Both covered by paired regression tests.

dudarenok-maker:
## Summary

Two bugs on the Model Manager "Optional add-ons" install/repair surface, hit while clicking **Repair** on Coqui XTTS v2 in dark mode.

1. **Install/repair aborted with `ERROR: Constraints cannot have extras`.** `install-coqui.mjs` pins its opt-in install against `requirements/base.txt` via `pip install coqui-tts -c base.txt`, but `base.txt` is a *requirements* file that legitimately carries extras (`uvicorn[standard]`), which pip forbids in a `-c` **constraints** file. Same latent bug at 3 more `-c base.txt` sites in `install-qwen3.mjs` (`qwenPipInstallArgs` + both flash-attn installs). Fixed with a shared `pip-constraints.mjs` sanitiser that strips `[extras]` (meaningless in a constraint — it only pins a version) into a temp constraints file; all four sites route through it, and `base.txt` stays the single source of truth.

2. **The failure text was illegible in dark mode.** The five in-app install affordances (`coqui`/`qwen`/`kokoro`/`whisper`/`venv-bootstrap`) render their error body as `text-rose-900/80`, but the `[data-theme='dark']` override in `styles.css` remapped only `.text-rose-900`, `/85`, and `/70` — the `/80` alpha was missed, so a failed install painted dark-maroon-on-dark-maroon (readable header, unreadable body). Added the missing selector.

## Test plan

- **New** `server/src/tts/pip-constraints.test.ts` — unit-tests the sanitiser (strips extras, keeps version pins, preserves comments) and asserts the real `base.txt` sanitises extras-free.
- **Extended** `src/test/dark-mode-css.test.ts` — guards `.text-rose-900\/80` under `[data-theme='dark']` (fails before the styles.css fix, passes after).
- **End-to-end** against the real sidecar venv: `-c base.txt` reproduces `ERROR: Constraints cannot have extras`; `-c <sanitized>` parses cleanly.
- Green locally: full frontend suite (3829), full server suite (4536), typecheck, and `verify:fast:branch` (incl. sidecar pytest + build) on push.

Closes #1566


dudarenok-maker:
Folded the high-effort code-review findings:

- **Unguarded base.txt read (correctness):** `writeSanitizedConstraintsFile` now falls back to the original path on read failure, so a missing base.txt yields pip's clean error + the caller's FAIL step instead of a raw ENOENT stack trace. Regression test added.
- **Temp-dir leak (correctness):** one memoised temp dir per process, removed on exit — no more orphan `castwright-constraints-*` dirs per install/repair.
- **Brittle test canary (cleanup):** dropped the 'base.txt still carries extras' assertion; synthetic unit cases already prove stripping.
- **Relative-path resolution (latent):** documented the standing assumption that base.txt holds only version pins (no `-r` includes), which is what makes the temp-dir copy safe — no code change needed today.

Server suite green (4537). Commit 617bcfe4.

PR #1739, opened 2026-07-20
fix(sidecar): split 1.7B design/mint VRAM footprint from the synth key

### LINKED ISSUE #1738: side — mint & synth share the qwen.1.7b footprint key; p95 under-reserves the rare heavy mint
### Context
Follow-up from #1737 (footprint p95 learning). On-box measurement showed the two 1.7B operations have very different VRAM footprints:
- **Plain 1.7B synth** (generation hot path, base17-only): ~**3955 MB** reserved.
- **1.7B mint** (voice-design; co-loads 0.6B + 1.7B and runs a distil phase): ~**7900 MB** (nvidia-smi device-total near the 8 GB ceiling).

Both admit under the same `FootprintTable` key `qwen.1.7b` (`_key()` returns `qwen.1.7b` for any `model` containing `1.7b`, regardless of operation). Since synths vastly outnumber mints during normal use, the windowed **p95 tracks the ~3955 synth** and the rare mint (a top-5% outlier) is excluded — so a mint is admitted against a ~4000 MB reservation while actually needing ~7900 MB.

### Risk
On a 2-card box under concurrency, a second op could be admitted into space the mint actually occupies → CUDA OOM during a mint. Low frequency (mints are cast-review-time, not render-time), and the #1737 uncontended-only recording partially mitigates (contended mints aren't recorded, and an uncontended mint that *does* record pushes the p95 up), but the under-reservation of the mint itself persists.

### Fix
Give the mint its own footprint key (e.g. `qwen.1.7b.mint`) so it learns/reserves its own (higher) p95 separately from synth. The mint route already calls a distinct engine path; thread a marker through the reservation `cfg` (or key off the operation) so `_key()` can separate them.

Refs #1737

dudarenok-maker:
## Summary

Follow-up to #1737. The two rare, heavy 1.7B **design-family** ops —
`/qwen/design-voice` and `/qwen/mint-variant` — shared the plain-synth
`qwen.1.7b` `FootprintTable` key. Because synths outnumber designs/mints by
orders of magnitude, the shared windowed p95 tracked the ~3915 MB synth, and
the far heavier mint (~5654 MB measured, base+base co-load) inherited that
under-sized reservation. Under capacity admission (`SEG_CAPACITY_ADMISSION`,
default OFF) that let a co-resident op be admitted into space the mint actually
occupies — a real OOM risk on a tight card.

**Fix (sidecar-only):** each design-family route tags its reservation `cfg`
with an `op` marker (`design` / `mint`), so `FootprintTable._key` routes it to
its own `qwen.1.7b.design` / `qwen.1.7b.mint` key and it learns an independent
windowed p95. Plain synth (no `op`) keeps the lean `qwen.1.7b` key and still
learns down to ~3915. An unknown `op` falls back to the shared synth key rather
than inventing one.

Cold-start seeds stay **6144 MB** for both new keys — deliberately: that's
safely above the ~5654 MB measured mint *and* still inside an 8 GB card's
~6659 MB admission headroom (`free ~7068 − 5%-reserve ~409`), so a first-ever
mint on an 8 GB box isn't spuriously refused before its window warms. The seeds
are cold-start priors only; the p95 window supersedes them once ≥5 real ops are
observed.

Scope note: the issue is titled for mint, but `design_voice` has the identical
root cause (same shared key), so both are split in one change — the mechanism is
shared and leaving design conflated would ship a half-fix.

> **Heads-up on the #1737 record:** while implementing this I verified the
> merged #1737 code against its commit. The shipped `_observed_mb` reads
> `max_memory_allocated` (not `reserved`) and `record()` runs unconditionally in
> `finally` — there is **no** "uncontended-only recording" machinery (an earlier
> session note claimed one). It's safe (device-wide reads err toward
> over-reserve, never OOM; noisy samples dilute in the 64-deep p95 window), but
> #1738's original premise that "#1737's uncontended-only recording partially
> mitigates" this rests on a feature that doesn't exist — the key-split is the
> actual mitigation.

## Test plan

- `test_footprints.py` — new: design-family keys separate from synth
  (`_key` routing incl. unknown-op fallback), independent windows (a flood of
  light synth observations doesn't drag down the mint's learned p95), and 8 GB
  cold-start admission fit.
- `test_design_mint_admission.py` — new: route-level assertion that each route
  passes the correct `op` tag so `_key` resolves `qwen.1.7b.design` /
  `qwen.1.7b.mint`; docstring corrected. Existing `neededMb == 6144` cases stay
  green (seeds unchanged at 6144).
- Full sidecar pytest suite green on pre-push (354s, on-box CUDA).
- `docs/local-llm.md` footprint anchors updated (parity-tested); plan 264
  follow-up marked CLOSED; technical release note added.

Closes #1738

### DIFF
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -254,11 +254,14 @@ the flag is OFF) should close so admission covers everything the budget did:
   > `fix/sidecar-mint-footprint-key`: each design-family route tags its
   > reservation cfg with an `op` (`design`/`mint`) so `FootprintTable._key`
   > routes them to their own `qwen.1.7b.design` / `qwen.1.7b.mint` keys, each
-  > learning an independent windowed p95. Cold-start seeds (6144 MB) still fit an
-  > 8 GB card's admission headroom (~6659 MB), so a first-ever mint isn't
-  > spuriously refused. Coverage: `test_footprints.py` (key separation,
-  > independent windows, 8 GB cold-start fit) + `test_design_mint_admission.py`
-  > (route-level `op`-tag wiring).
+  > learning an independent windowed p95. Mint's cold-start seed (6144 MB) sits
+  > above its measured ~5654 MB peak and still fits an 8 GB card's admission
+  > headroom (~6659 MB), so a first-ever mint on a bare card isn't spuriously
+  > refused; design's seed (7168 MB) is a deliberately conservative UNMEASURED
+  > prior (VoiceDesign-1.7B + 0.6B-Base — a different, un-measured load) that
+  > errs toward refuse-not-OOM until #1742 measures its real peak. Coverage:
+  > `test_footprints.py` (key separation, independent windows, bare-8 GB mint
+  > fit) + `test_design_mint_admission.py` (route-level `op`-tag wiring).
 - **`persona-gpu-plan.ts`'s `unloadResidentSidecar` + `GpuBusyForPersonaError`
   are now dead code** (the reverse-evict was removed); deletion candidates.
 - **`engine-vram-cost.ts` is provably dead** (its registry weights are deleted) —
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -245,6 +245,20 @@ the flag is OFF) should close so admission covers everything the budget did:
   > observations ignored), `test_placement.py`, `test_design_mint_admission.py`.
   > The multi-op on-box acceptance walkthrough above is still owed before
   > `SEG_CAPACITY_ADMISSION` is defaulted ON.
+  >
+  > **Follow-up CLOSED (#1738).** #1737 left both 1.7B design-family ops
+  > (`design_voice`, `mint_variant`) sharing the plain-synth `qwen.1.7b`
+  > footprint key. Since synths outnumber designs/mints by orders of magnitude,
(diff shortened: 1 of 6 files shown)


dudarenok-maker:
**Code review (high effort) run and folded.** 3 PLAUSIBLE findings, none CONFIRMED — all environment-dependent and clustered on the cold-start seed:

1. *design seed unmeasured (unsafe direction).* I'd seeded `qwen.1.7b.design` at 6144 and claimed the mint measurement backed it, but `design_voice` loads a different, unmeasured stack (VoiceDesign-1.7B + 0.6B-Base). **Folded** (72a36cf6): seed bumped to a conservative **7168** (refuse-not-OOM until measured), comment/docs made honest, follow-up #1742 filed to measure the real peak.
2. *test hardcoded free=7068 / over-claimed a production admit.* **Folded**: rescoped to a bare-card lower-bound for the measured mint only; design asserted `>= mint` on purpose (meant not to fit a tight card). 503 route tests now assert against `SEED_FOOTPRINTS_MB`, not a literal.
3. *mint seed 6144 > measured 5654 (over-refusal band).* Safe refuse-not-OOM direction and margin-above-measured per the #1737 convention; its "refused→never-learns" half is a pre-existing #1737 property (out of scope). Noted, not over-corrected.

Fold is a seed-constant + test-scoping + comment change (no new logic), so no re-review triggered (findings were PLAUSIBLE, not Critical/Contradicted).

PR #1821, opened 2026-07-26
test(e2e,scripts): capture and stage the series cast card as a real in-app export

### LINKED ISSUE #1820: chore: capture the series cast-card PNG as a real in-app export in the marketing harness
## What

`brand/go-to-market/launch-post-images/README.md` lists one asset as still-to-do:

> **Series "cast card" PNG** — the portrait share unit; **export it from the app** (a real multi-book library). The only genuinely new asset — drop it in as `series-cast-card.png`.

It is the only launch asset that travels **on its own** — off-app, into a Reddit post or the marketing site — rather than as a screenshot of a window. The existing `series-share-card` marketing scene frames the modal in context (dimmed library behind it, close button, both download buttons), which is the wrong artefact for that job.

The card itself already exists in-app (`src/components/series-memory/series-share-card.tsx`, exported via the `ShareCardModal` "Download image (.png)" button); what was missing was a repeatable way to produce the file.

## Acceptance

- The marketing capture harness can capture what the app **exports**, not just what it shows — the downloaded file, byte for byte, not a re-render of the same pixels.
- A `series-cast-card-export` scene produces the standalone 4:5 card from the Hollow Tide multi-book fixture (a real series — the card's hero line counts voices kept true *across* books, so a single-book library would render a meaningless card).
- The capture is repeatable cold, not just against an already-warm dev server.

## Key files

- `e2e/marketing/scenes.ts` — scene registry
- `e2e/marketing/capture.spec.ts` — runner
- `e2e/marketing/README.md` — harness docs

## Benefit

*Benefit (technical):* the one launch asset that had to be produced by hand becomes a one-command regenerable artefact, so it can be refreshed alongside every other marketing shot instead of drifting.

dudarenok-maker:
## Summary

The series cast card was the last unproduced launch asset — and the only one that has to travel **on its own**, off-app into a post or the marketing site, rather than as a screenshot of a window. The existing `series-share-card` scene frames the modal *in context* (dimmed library behind it, close button, both download buttons), which is the wrong artefact for that job.

So rather than screenshot the card, this teaches the marketing harness to capture what the app **exports**, then wires the result into the existing staging rail.

**Capture** (`e2e/marketing/`)

- **`downloadAction`** — the runner presses an in-app export button and keeps the file the browser hands back, so the output PNG is the real user-facing artefact, byte for byte, not the harness's re-render of the same pixels. It runs *inside* the theme loop, unlike `action`, because an exported card bakes the active theme's tokens into the file — the click has to land after `emulateMedia`. Export scenes deliberately **never** degrade to a screenshot on failure: that would write a plausible-looking PNG of the wrong thing (the modal, chrome and all) under the export's filename, which is worse than a red run.
- **`themes`** — pins a scene to a subset. Intersects with `CAPTURE_THEME` rather than overriding it, so a scene never writes a theme it disowns.
- **`series-cast-card-export`** — the scene, off the Hollow Tide multi-book fixture. A real series is the whole point: the card's hero line counts voices kept true *across* books, so a single-book library would render a meaningless card.

**Staging** (`scripts/stage-marketing-screenshots.mjs`)

Manifest entries may now pin `themes` too. Not optional: `main()` sets a non-zero exit code on **any** missing source, so a dark-only capture staged as a light+dark pair would leave `npm run stage:marketing-screenshots` red forever. A single-theme entry also drops the `-dark` suffix — the file isn't the dark half of a pair, it's the only variant, and nothing embedding it should have to know which theme produced it. That suffix drop opens a collision the existing duplicate-output-name check can't see (a dark-only `foo` and a paired `foo` would both claim `foo.webp`), so the new tests assert uniqueness on **resolved destination paths** rather than manifest names.

### Why the card is dark-only — measured, not assumed

Both themes resolve the card's surface to `rgb(27,23,20)` and its text to `rgb(244,239,236)`. **Identical.** The only thing the theme changes is the accent:

| Theme | Accent (`--magenta`) | Contrast on the card surface |
|---|---|---|
| light | `#a43c6c` | **2.9:1** — fails WCAG AA |
| dark | `#e58fb8` | **7.6:1** — passes AA + AAA |

So there is no light *card*, only a light *accent* on the same dark card — and it's the accent used for the "SERIES MEMORY · …" label and the `castwright.ai` footer, both small text. `series-share-card.tsx` hardcodes `bg-[#1b1714] text-cream` but draws that accent from the *themed* token, so a light-theme user exporting their own card today gets the 2.9:1 version. The component fix — a fixed dark surface should pin its own accent rather than inherit the page theme's — is **out of scope here** and left for follow-up; this PR adds no app code.

### The cold-run failure worth keeping

The first cold run failed, and the failure snapshot is the interesting part: the modal was open, the card was rendered, and the button read `"Rendering…" [disabled]` — `html-to-image` was mid-flight when the deadline hit. It walks the node, inlines every stylesheet, re-fetches each `url()` under `cacheBust`, then rasterises; stacked on the cold Vite compile the stock 120s budget already assumes, that overran. The identical scene finishes in ~2s warm, which is exactly why this would have read as a phantom flake if the snapshot hadn't been opened. Export scenes now get `test.slow()`.

## Test plan

- `CAPTURE_SCENE=series-cast-card-export npm run capture:marketing` — green. Output verified as a valid PNG at **672×840** (4:5 at `pixelRatio: 2`), matching the in-app export exactly.
- Verified **cold** (fresh Vite process, `node_modules/.vite` dep cache cleared), not just against a warm dev server — the cold path is the one that originally failed.
- `CAPTURE_THEME=light CAPTURE_SCENE=series-cast-card-export` — skips cleanly rather than writing a theme the scene disowns.
- `CAPTURE_SCENE=series-share-card npm run capture:marketing` — green, still writing **both** light + dark, confirming the `themes` intersection didn't disturb the normal screenshot path.
- `npm run test:hooks` — 606 pass, including 5 new cases on the staging planner (single-theme fan-out, suffix drop, no phantom light source, destination-path uniqueness, the manifest entry itself).
- `npm run verify:fast:branch` — green.
- The underlying export behaviour is already gated in the real e2e suite by `e2e/series-memory.spec.ts` → *"share card exports a PNG download"*; this PR adds no app code, so it leans on that rather than duplicating it.

**Release notes:** deliberately skipped — local capture/staging tooling with no user- or operator-visible delta.

Assets dropped into the launch pack (git-ignored, local): full-res `series-cast-card.png` for post attachments and `marketing-site/screenshots/series-cast-card.webp` (26 KB) for the site mirror. That clears the last open item in that pack's README, now rewritten as a one-command regen recipe.

Closes #1820

### DIFF
--- a/e2e/marketing/README.md
+++ b/e2e/marketing/README.md
@@ -59,7 +59,44 @@ Add one row to `scenes.ts`:
   `#/books/:bookId/<view>`, `#/account`, `#/voices`.
 - `waitFor` is an optional content selector (non-fatal) so the shot isn't taken
   on the loading shell.
+- `themes` pins a scene to `['dark']` or `['light']`. Only for a fixed-treatment
+  surface where the other theme is not a second look but a worse version of the
+  same one (see `series-cast-card-export`). It intersects with `CAPTURE_THEME`
+  rather than overriding it, so a scene never writes a theme it disowns.
 - `viewports` defaults to `['desktop']`.
+
+## Export scenes (capturing what the app *produces*, not what it shows)
+
+Most scenes screenshot the page. A scene with a **`downloadAction`** instead
+presses an in-app export button and keeps the file the browser downloads — so
+the output PNG is the real user-facing artefact, byte for byte, rather than the
+harness's own re-render of the same pixels.
+
+```ts
+{ id: 'series-cast-card-export', hash: '#/', themes: ['dark'],
+  waitFor: '[data-testid="series-memory-chip"]',
+  action: async (page) => { /* …open the modal… */ },
+  waitForAfterAction: '[data-testid="series-share-card"]',
+  downloadAction: async (page) =>
+    page.getByRole('button', { name: /download image \(\.png\)/i }).click(),
+  strict: true },
+```
+
+Two ordering rules matter:
+
+- **`action` reaches the export affordance; `downloadAction` presses it.**
+  `action` runs once, before the theme loop; `downloadAction` runs *inside* it,
+  because an exported card bakes the active theme's tokens into the file, so the
+  click has to happen after `emulateMedia`.
+- **Export scenes never degrade to a screenshot.** A fallback would write a
+  plausible-looking PNG of the wrong thing (the modal, chrome and all) under the
+  export's filename — worse than a red run.
+
+Export scenes are marked `test.slow()` automatically. `html-to-image` walks the
+node, inlines every stylesheet, re-fetches each `url()` under `cacheBust`, then
+rasterises; stacked on the cold Vite compile the stock 120s budget already
+assumes, that overran the deadline mid-render on a cold first run (the button
+was still reading "Rendering…") while finishing in ~2s warm.
 
 ## Covers
 
--- a/e2e/marketing/capture.spec.ts
+++ b/e2e/marketing/capture.spec.ts
@@ -63,6 +63,13 @@ for (const scene of SCENES) {
   test(`capture ${scene.id}`, async ({ page }, testInfo) => {
     const vp = testInfo.project.name as Viewport;
     test.skip(!viewports.includes(vp), `viewport ${vp} not requested for ${scene.id}`);
+    /* An export scene does real work after the page is already posed —
+       html-to-image walks the node, inlines every stylesheet, re-fetches each
+       url() resource under cacheBust, then rasterises. Stacked on top of the
+       cold Vite compile the stock 120s budget is already sized for, that was
+       enough to blow the deadline (the download event never arrived in time)
+       on a cold first run, while the identical scene finishes in ~2s warm. */
+    if (scene.downloadAction) test.slow();
 
     /* Hydrate the library slice first — several views (e.g. the listen cover at
        routes/index.tsx:833) read book data from `s.library.books`, which is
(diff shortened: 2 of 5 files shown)


dudarenok-maker:
## Scope grew: the raw-name webp mirror is now regenerable too

While regenerating the marketing set, the staged folder turned out to hold **two** sets under two naming conventions:

- the **curated** set this script writes — `cast-reuse.webp` / `cast-reuse-dark.webp` (the names the website embeds)
- **168 raw-name** files — `cast-reuse.desktop.dark.webp` — from a one-off hand pass

Because only the curated half was scripted, the raw-name half had silently rotted. It sat in the same folder looking equally current while actually predating both the German fixture fix (#1823) and an app version — **v1.6.0 chrome in a v1.14.0 app**. Anything picked from it shipped the wrong content, and `language-cast-confirm-german.desktop.dark.webp` still showed a book that no longer exists.

Regenerating it by hand again would just reset that clock, so `--all` now mirrors every `<scene>.<viewport>.<theme>.png` capture to its matching `.webp`, deduped against the curated plan so no output path is claimed twice.

Also adds `MARKETING_SRC` / `MARKETING_DEST` overrides (mirroring the website repo's existing `SCREENS_SRC`) so the script can be pointed at the primary checkout while running from a worktree — which is how it was exercised here.

### Verification

- `node --test scripts/tests/stage-marketing-screenshots.test.mjs` — **11 pass**, including three new `mirrorPlan` cases: raw-name mapping, the non-capture filter (the source dir accumulates subdirectories and stray files; handing those to ffmpeg produces garbage), and a guard that the two naming schemes can never collide on a destination path.
- Ran for real: **240/240 staged, zero missing** (53 curated + 187 mirrored).
- **242 of the folder's 246 webps now regenerated.** The only 4 stragglers are `companion-iphone` / `companion-pixel`, which come from the separate `capture-companion.mjs` pipeline that this script's header already excludes — they need an emulator, not this rail.

PR #1870, opened 2026-07-27
fix(server,sidecar): deterministic device env for the sidecar; codec knob resolves UUID pins

### LINKED ISSUE #1857: GPU device-list cache is never warmed unless Advanced settings is opened
Surfaced by the independent review of PR #1852 (removal of `voices.library.enabled`). **Pre-existing hole, newly unmasked** — filing rather than folding into that PR, since the fix belongs at server boot, not in a flag removal.

## What

`ensureGpuDeviceListWarm()` (`server/src/routes/config.ts:39`) is the only warmer of the GPU device-list cache, and it runs on `GET /api/config` only. Verified: `grep -rn ensureGpuDeviceListWarm server/src` returns just its definition and that single call site.

Until PR #1852, `src/store/index.ts` dispatched `fetchConfig()` at app boot, so **every page load in any view** incidentally warmed that cache. That dispatch existed solely to hydrate the `voices.library.enabled` gate and was removed with it. The remaining triggers — `GET /api/config`, `GET /api/gpu/devices`, and a `PUT /api/config` device write — are now dispatched only from `src/views/advanced.tsx`. So nothing warms the cache until the user opens Advanced settings.

## Why it matters

`buildSidecarEnv` (`server/src/tts/spawn-sidecar.ts`) injects `String(st.effective)` for each non-default `restart-sidecar` knob, and `tts.qwen.device` / `tts.kokoro.device` / `tts.coqui.device` are `apply: 'restart-sidecar'` with real env names. With a cold device cache, `resolveKnob` (`server/src/config/resolver.ts:41-48`) returns the raw `cuda-uuid:<uuid>` literal rather than translating it to an index — so a user holding a UUID device pin can get `QWEN_DEVICE=cuda-uuid:GPU-…` handed to the sidecar on a supervisor respawn.

## Not caused by #1852

The initial boot spawn already had this hole: `spawnSidecar` runs at `app.listen`, before any HTTP request could have warmed the cache. The frontend's boot fetch only ever mitigated *post-boot respawns*, and by accident. Removing it widened the window rather than creating the bug.

## Suggested fix

Warm the device list at server boot (before the first `spawnSidecar`) instead of relying on an HTTP request having happened. That closes the initial-spawn case too, which no frontend fetch ever covered.

## Acceptance

- A supervisor respawn with a `cuda-uuid:` device pin and no prior `GET /api/config` passes a translated device index, not the raw `cuda-uuid:` literal.
- Regression test at the `buildSidecarEnv` seam with a cold cache.

dudarenok-maker:
Closes #1857

## The issue's premise was wrong, and its suggested fix would have made things worse

#1857 reports that the GPU device-list cache is never warmed unless Advanced settings is opened, and that this hands the sidecar a raw `cuda-uuid:GPU-…` literal which misroutes the engine. It asks for the cache to be warmed at server boot.

The sidecar resolves that literal itself. `_read_device_env` → `_resolve_uuid_to_index` (`main.py:1873`) maps `cuda-uuid:<uuid>` to a live index on every read, for all three engine device knobs. That code exists *because* the Node cache is necessarily cold at the boot spawn — its docstring says so at `main.py:1843`. The raw literal is the designed input, not the bug.

Warming the cache is what causes harm. A warm cache is what makes `resolveKnob` freeze a `cuda:N` into the spawn env, and `buildOpts` re-runs `buildSidecarEnv` on **every respawn** — so a card that later vanishes makes `_validate_cuda_index` raise and the engine load fail on every retry, and a card that renumbers lands on the wrong one silently. This is the same failure the design rejected `nvidia-smi` enumeration for: a confidently wrong index is worse than a literal the sidecar resolves correctly.

## What this actually fixes

**1. `buildSidecarEnv` was nondeterministic** — named by neither the issue nor the first draft of the spec.

| Cache state | `QWEN_DEVICE` handed to the sidecar |
|---|---|
| Cold (nobody opened Advanced) | `cuda-uuid:GPU-…` — sidecar resolves live, self-healing |
| Warm (Advanced was opened) | `cuda:1` — frozen, re-emitted on every respawn |

Same config, same box, two different spawn envs depending on UI history. New `resolveKnobForSidecarEnv` skips the reconcile, so the sidecar always gets the UUID form. Node's mapping is *derived from* the sidecar's own `/devices`, so it can only ever be staler — translating in Node was strictly a downgrade.

The UI path is untouched: `resolveKnob` keeps its translation and its `staleReason: 'uuid_unresolved'` badge. `expectedSidecarCeilings` / `expectedFreeFloorMb` also keep `resolveKnob` — numeric knobs, different question.

**2. `QWEN_CODEC_DEVICE` ignored a UUID pin.** It is the fourth `type: 'device'` knob and the only one read with a bare `os.environ.get`. A pinned card *passed* `_validate_cuda_index` — `_parse_device` reads `cuda-uuid:x` as family `cuda` with no concrete index, so the range check no-ops — then failed inside `_move_codec_to_device`'s `.to()` and was rolled back to CPU.

An unresolvable pin now falls back to `cpu`, deliberately **not** `_read_device_env`'s `auto`. For the codec `auto` means "follow the model", which would move it onto the very card a user pinning the codec elsewhere was spreading VRAM away from — and the rollback only fires if `.to()` raises, not if a successful move OOMs a later decode.

## What was dropped

An earlier revision of the spec implemented the boot warm as #1857 suggests, plus a no-downgrade rule, a `recordState` option on `probeSidecarHealth`, and a new gate accessor. Two independent Fable reviews killed it:

- Warming is the *cause* of the frozen-index hazard above, not a mitigation.
- The retry loop's stop condition conflated "sidecar reported zero cards" with "the `/devices` fetch timed out", so on a two-GPU box a single 2s abort after `devices_state` flipped to `ready` would log **"sidecar reports no CUDA cards"** and exit with a cold cache.
- The no-downgrade rule made `staleReason: 'uuid_unresolved'` unreachable for the eGPU-drop case it was named for, with no mechanism by which a warm cache could ever learn cards went to zero.

Once `buildSidecarEnv` is off the cache, no consumer needs it warmed without a request — `engine-device.ts` works off the UUID form, and `staleReason` is only surfaced through routes that warm on demand. Request-scoped warming turns out to be correct rather than accidental. The reasoning is preserved in the spec's "Rejected: the boot warm" section.

## Testing

- `resolver.test.ts` — the two entry points diverge exactly on the device-UUID branch; env-locked and default paths unaffected; the pre-existing `resolveKnob` reconcile tests still pin the UI path.
- `sidecar-env.test.ts` — `buildSidecarEnv` emits the raw literal under **both** warm and cold caches. The warm case is the regression test and was red before the fix (`expected 'cuda:1' to be 'cuda-uuid:GPU-1'`).
- `test_device_parse.py` — `_codec_device_pref()` resolves a visible UUID, falls back to `cpu` for a vanished one, and passes `cpu`/`auto`/`cuda:N` through unchanged. Named rather than inlined so it is reachable without weights or a GPU.

Full server suite: 443 files, 5426 tests passing. Typecheck clean. Sidecar `test_device_parse.py` + `test_module_import_order.py` green.

## Not done

No on-box acceptance yet — the behaviour that matters most (a respawn after a real index change reaching the right card) needs the dual-GPU box.

### DIFF
--- a/docs/superpowers/plans/2026-07-27-gpu-device-list-boot-warm.md
+++ b/docs/superpowers/plans/2026-07-27-gpu-device-list-boot-warm.md
@@ -1,934 +1,309 @@
-# GPU device-list boot warm + codec device-knob parity — Implementation Plan
+# Deterministic sidecar device env + codec knob parity — Implementation Plan
 
 > **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
 
-**Goal:** Make the Qwen codec honour a `cuda-uuid:` device pin, and warm the Node-side GPU device-list cache from a server-owned boot loop instead of depending on a frontend HTTP request.
+**Goal:** Make the sidecar's device env deterministic — always the raw `cuda-uuid:` literal, which the sidecar resolves live on every spawn — and make the Qwen codec knob honour a UUID pin.
 
-**Architecture:** Six tasks. Tasks 1–2 harden the two pieces of shared state a new concurrent writer would otherwise destabilise (the device-list cache, and the sidecar health probe's side effects). Task 3 adds the warm module itself. Task 4 wires it into boot and de-duplicates the existing warmer. Task 5 pins the `buildSidecarEnv` contract under both cache states. Task 6 is the independent sidecar fix and can be done at any point.
+**Architecture:** Two independent tasks. Task 1 is Node/TypeScript: a second resolver entry point that skips device-UUID reconciliation, used only by `buildSidecarEnv`. Task 2 is the Python sidecar: a named `_codec_device_pref()` helper with a `cpu` fallback. Neither depends on the other; either can land alone.
 
-**Tech Stack:** TypeScript (Node/Express, ESM, `.js` import specifiers), Vitest for server tests; Python 3.12 + FastAPI for the TTS sidecar, pytest for its tests.
+**Tech Stack:** TypeScript (Node/Express, ESM, `.js` import specifiers), Vitest; Python 3.12 + FastAPI sidecar, pytest.
 
 **Spec:** `docs/superpowers/specs/2026-07-27-gpu-device-list-boot-warm-design.md`
 
 ## Global Constraints
 
 - Worktree: `C:\Claude\Projects\Audiobook-Generator\.claude\worktrees\fix+1857-gpu-device-cache-boot-warm`, branch `fix/1857-gpu-device-cache-boot-warm`. Run every command from there; never `cd` to the main checkout.
-- Server imports use ESM `.js` specifiers even for TypeScript sources (`./gpu-device-list-state.js`).
-- `gpu/**` must never import from `routes/**` — including type-only imports, which still count as cycle edges. Cross-layer access goes through the registered-provider gates in `gpu/*-gate.ts`.
-- Commit messages follow Conventional Commits with a scope (`fix(server): …`, `test(server): …`, `fix(sidecar): …`). Every commit body ends with:
+- Server imports use ESM `.js` specifiers even for TypeScript sources.
+- Server tests: `npm --prefix server run test -- <path>` from the worktree root, where `<path>` is relative to `server/` (e.g. `src/config/resolver.test.ts`). `server/vitest.config.ts` has `include: src/**` resolved against cwd, so `npx vitest --config server/vitest.config.ts server/src/…` from the root matches the *frontend* tree and exits "No test files found".
+- Sidecar tests: `server/tts-sidecar/.venv/Scripts/python.exe -m pytest server/tts-sidecar/tests/<file> -v`.
+- Commit messages follow Conventional Commits with a scope. Every commit body ends with:
   ```
   Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
   Claude-Session: https://claude.ai/code/session_01VxtGhVyXqmikbXSpsvuEmg
   ```
-- Server tests: `npm --prefix server run test -- <path>` from the worktree root, where `<path>` is relative to `server/` (e.g. `src/gpu/foo.test.ts`, **not** `server/src/gpu/foo.test.ts`). `server/vitest.config.ts` has `include: src/**`, resolved against whatever cwd vitest runs in — so `npx vitest run --config server/vitest.config.ts server/src/...` from the worktree root resolves `src/**` against the *frontend* tree, matches nothing, and exits 1 with "No test files found". Verified by running it.
-- Sidecar tests: `server/tts-sidecar/.venv/Scripts/python.exe -m pytest server/tts-sidecar/tests/<file> -v`.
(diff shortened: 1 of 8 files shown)


dudarenok-maker:
## Code review — high effort

**Caveat up front:** this is a self-review. The genuinely independent pass on this work was two Fable reviews run *before* implementation, and they changed the design — the original approach (warm the device cache at boot, as #1857 suggests) was dropped because warming is what freezes a `cuda:N` into the respawn env. This pass verifies the implementation matches the reviewed design and looks for what landed wrong.

### Correctness — no defects found

- **Only one call site switched.** `spawn-sidecar.ts:119` and `:136` (`expectedSidecarCeilings`, `expectedFreeFloorMb`) still use `resolveKnob`. The split is behaviourally identical for their numeric knobs, so a wrong edit here would have been invisible to tests — verified by grep, not by test.
- **Guard placement is right.** `reconcileDeviceUuid &&` short-circuits first, so non-device knobs take the identical path through both entry points. The `env` and `default` branches are untouched by the flag, which the tests pin.
- **No bare device-env reads remain** in `main.py` — `QWEN_CODEC_DEVICE` is read at line 429 only, inside the new helper.
- **The sidecar never echoes a raw device env back.** `/health`'s `device` field resolves through `_read_device_env` (`main.py:6233`), so a UUID literal cannot surface in the UI.
- **Non-UUID passthrough is safe.** `_resolve_uuid_to_index` returns non-`cuda-uuid:` values unchanged, so the `resolved is None` branch and its warning fire only for a genuinely unmatched UUID.

One undocumented improvement: a whitespace-only `QWEN_CODEC_DEVICE` previously fell through as `""` and failed inside torch's `.to()`; it now returns `cpu` cleanly. Covered by test, not mentioned in the PR body.

### Test coverage

Strong on the changed paths. Two deliberate gaps:

- **Nothing pins that the two ceiling helpers still use `resolveKnob`.** Declined during plan review as over-fitting to an edit mistake — but it means this file's one real hazard is guarded by a comment rather than a test.
- **No Node test drives `tts.qwen.codecDevice` through `buildSidecarEnv`** — the knob that motivated the sidecar half. The loop is generic over knobs so the value is low, but the two halves of this PR are never exercised together.

The warm-cache regression test was genuinely red before the fix (`expected 'cuda:1' to be 'cuda-uuid:GPU-1'`), not a placebo.

### Style

Matches the codebase: explanatory docblocks stating *why*, inline issue references, ESM `.js` specifiers, conventional commit scopes.

Nit: the branch carries 5 docs commits, three of which are superseded revisions of the same two files. Under merge-commit-only that history lands on `main` in full — it documents the review trail honestly, but a later reader of `main` sees a design that was reverted before it shipped.

### Performance / security

Neither affected. No new I/O or parsing of untrusted input; the resolver change removes a `.find()` from the spawn path.

### Verification

Full server suite 443 files / 5426 tests passing; typecheck clean; sidecar `test_device_parse.py` + `test_module_import_order.py` green. One `Worker exited unexpectedly` in the full run — the known tinypool flake; every file still reported pass.

**Not verified:** on-box behaviour. A respawn after a real index change reaching the correct card needs the dual-GPU box, and is called out as owed in the PR description.

No blocking findings.

PR #1916, opened 2026-07-28
fix(scripts): reject sub-lettered rows and ignore fenced blocks in the register checker

### LINKED ISSUE #1913: ops-44 — register checker: decide on sub-lettered rows and fenced-block handling
## What

Two latent parser limitations in `scripts/check-onbox-register.mjs` (ops-43,
#1907) surfaced by the independent review on PR #1908. Both were deliberately
left out of that PR because each needs a **decision about register convention**,
not just a parser change.

### 1. Sub-lettered row numbers are invisible

`^### <Letter>(\d+)\b` does not match `### A19b · …` — the `\b` fails between
`9` and `b`. A row inserted that way is silently uncounted, and correctly
bumping the glance table to 20 then produces
`Group A: glance table says 20, body has 19 rows (A1–A19)` — a legitimate edit
rejected, with no hint that `A19b` was the ignored row.

**The decision needed:** do we want sub-lettered rows at all? The register
currently handles the "one row, two debts" case with a title annotation
(`A2 · Capacity-aware GPU placement · **two distinct debts**`) rather than
sub-letters, which is arguably the better convention because it keeps the count
honest. If we confirm that, the fix is to *reject* `A19b` with a message saying
so — not to support it. If we want sub-letters, the counting rule needs defining
first (does `A19b` count as a row toward the total, or not?).

### 2. No code-fence awareness in the heading scan

`^### ` and `^## ` are matched without tracking fenced code blocks, so a
` ```md ` block inside a group section containing an example `### F2 · …` row or
an example `## Group Z — …` heading would be counted as real. Confirmed by
construction on a scratch copy.

**Currently latent:** the register has zero fenced blocks today. Blockquotes are
already safe (`> ### …` does not match `^###`).

**The decision needed:** is fence-tracking worth the parser complexity for a
file that has never had a fenced block, or is the cheaper answer a check that
*rejects* fenced blocks in group sections so the situation cannot arise?

## Acceptance

- [ ] A decision recorded on whether sub-lettered rows are permitted, and the
      checker enforcing whichever answer
- [ ] A decision recorded on fenced blocks, and the checker either tolerating
      them correctly or rejecting them with a clear message
- [ ] Tests for whichever behaviour is chosen

## Benefit

*Benefit (technical):* both are silent-miscount paths in a tool whose entire
purpose is catching silent miscounts. Neither is urgent — one is unreachable
today and the other requires an unusual edit — but both should be closed
deliberately rather than left as known gaps.

## Context

Everything else the review found on #1908 was folded into that PR. These two
were split out precisely because they are judgement calls.

dudarenok-maker:
## Summary

Closes the last open item from the ops-43 register-checker work. Two parser edge
cases the independent review on #1908 surfaced, both deliberately left out of
that PR because each needed a decision about register convention rather than a
parser change. The decision is recorded on #1913 and the two answers
intentionally differ in shape.

**1. Sub-lettered rows (`### A19b`) are rejected.** One row should mean one
countable unit. Supporting sub-letters would force a rule about whether `A19b`
counts toward the total, and any answer to that makes the total less trustworthy
— which defeats the register's purpose. The existing convention already covers
the motivating case by annotating a multi-debt row's title
(`A2 · … · **two distinct debts**`). Previously such a heading was **silently
uncounted**, surfacing later as a confusing count mismatch with no hint at the
cause; now it says what to do:

> `Row heading "### A19b" is not a valid row number. Rows are numbered contiguously (A1, A2, …) — for a row covering more than one debt, annotate its title instead of sub-lettering.`

Scoped so a `### ` heading that isn't row-shaped at all (`### Notes`) stays
ignored — group sections may legitimately gain non-row subheadings.

**2. Fenced code blocks are parsed around, not rejected.** This reverses the
"just forbid them" option floated when #1913 was filed. Rejecting would
constrain what a future row may contain for the parser's convenience, which is
backwards — E6 already carries several inline commands and a future row could
reasonably want a fenced block. `stripFences()` blanks fenced content before
parsing, so an example `### F2 · …` can't inflate a group's count and an example
`## Group Z — …` can't create a phantom group. Both ``` and `~~~` fences, with
per-delimiter toggling so a mismatched fence type is content, not a closer.

Worth noting the implementation improved on the brief: rather than making the
section split and the row scan fence-aware separately, it applies one
`stripFences()` pass up front, so both inherit it from a single change point
that can't drift apart.

The register has zero fenced blocks today, so half of this is prevention — the
tests are the deliverable for that half.

## Test plan

- `npm run test:hooks` → **719/719 pass** (20 in this file, 5 new)
- `node scripts/check-onbox-register.mjs` against the real register → **exit 0**
- All six acceptance bullets from the decision comment demonstrated against
  scratch copies (the tracked register was never modified) and independently
  re-verified before merge:

| Case | Result |
|---|---|
| real register | `[]` |
| `### A19b` sub-letter | rejected, with the message above |
| `### Notes` non-row heading | ignored |
| fenced `### F2` inside Group F | count not inflated |
| fenced `## Group Z — …` | no phantom group |
| `~~~` fence parity | same as ``` |

- Removal spot-checks: reverting `stripFences` failed exactly the 3 fence tests
  and left the 2 sub-letter tests passing; reverting the sub-letter detection
  failed exactly that test and left the fence tests passing. The new tests pin
  their own behaviour and are isolated from each other.

**Checklist steps not applicable:** (1) no regression plan — the issue body plus
paired tests are the spec for a localized item; (3) no on-box acceptance — no
hardware surface; (5) release notes skipped, contributor-facing tooling with no
user- or operator-visible delta.

Closes #1913

### DIFF
--- a/scripts/check-onbox-register.mjs
+++ b/scripts/check-onbox-register.mjs
@@ -16,6 +16,37 @@ import { readFileSync } from 'node:fs';
 // as "not a group" rather than special-casing two sections whose rows would
 // buy little coverage for a lot of parser fragility.
 
+// ops-44 (issue #1913) added two edge-case behaviours on top of the above: a
+// row heading that looks row-shaped but isn't a strict `<Letter><N>` (e.g. a
+// sub-lettered `### A19b`) is now rejected with a specific error rather than
+// silently uncounted, and both heading scans below are fence-aware so an
+// example `##`/`###` line inside a fenced code block can't be parsed as a
+// real section or row.
+
+// Blanks the contents of fenced code blocks (``` or ~~~) so neither the
+// section split nor the row-heading scan below can mistake an example
+// heading inside a fence for a real one. Toggling is per-delimiter — a line
+// only closes a fence when it starts with the SAME delimiter that opened it,
+// so a `~~~` line inside a ```-fenced block is just content, not a closer.
+function stripFences(text) {
+  const lines = text.split('\n');
+  let openFence = null;
+  return lines
+    .map((line) => {
+      const trimmed = line.trimStart();
+      if (openFence) {
+        if (trimmed.startsWith(openFence)) openFence = null;
+        return '';
+      }
+      if (trimmed.startsWith('```') || trimmed.startsWith('~~~')) {
+        openFence = trimmed.startsWith('```') ? '```' : '~~~';
+        return '';
+      }
+      return line;
+    })
+    .join('\n');
+}
+
 // Splits the document into `## `-level sections: { title, body } from just
 // after each `## Heading` line to just before the next one. Row headings use
 // `### ` (three `#`s), so a regex requiring the space directly after exactly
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
## Review findings folded (`19225073`)

Independent review at Premium tier found a **Critical** that this PR itself introduced, plus a Major. Both reproduced independently before fixing.

**Critical — an unterminated fence silently blanked the register.** `stripFences` never checked whether a fence was left open at EOF, so one stray delimiter blanked everything after it and the checker validated the truncated remainder reporting nothing wrong:

```
control, no fence  -> ["Group A: glance table says 1, body has 2 rows (A1–A2)…"]
one stray fence    -> []
```

A regression against `HEAD~1`, where a stray fence was inert — and the exact outcome the brief named as worse than a crash. Where it stayed loud it reported wrong numbers: replaying the real 2026-07-28 incident with a stray fence gave "body has 3 rows (E1–E3)" against an actual 7. Now errors with the opening line number.

**Major — `\b` silently accepted dotted sub-numbering.** `### A2.1` and `### A2.2` both counted as row 2, producing `found A1, A2, A2, A3 — fix a duplicate` when no duplicate exists. Half the defect class this PR exists to close. Now `(?=\s|$)`; verified to agree with the old regex on all 36 real headings.

**Also folded:** cascaded errors (one problem emitted up to four messages — now mirrors the existing `malformedLetterSet` suppression convention), and a raw `\r` leaking into messages under CRLF.

**Documented, not fixed:** two *balanced* stray fences bracketing a real row can still hide it; contiguity only catches that when the hidden row isn't the group's highest-numbered one. Recorded as a known bound in a comment rather than assumed away.

**Verified after the fold:** 725/725 tests; real register exit 0 in both LF and CRLF; each new test confirmed to fail with its fix reverted; the four pre-existing checks' messages byte-identical.

One note on process: the remediation brief I wrote specified `[^\r\n]*` for the CRLF fix, which alone would have broken CRLF matching entirely — `$` in multiline mode matches directly before `\n`, so excluding `\r` from the capture leaves nothing able to match. That was caught and corrected with `\r?` outside the group, and flagged rather than quietly absorbed.

PR #1942, opened 2026-07-29
fix(sidecar): cache the clone prompt as a (prompt, language) tuple

### LINKED ISSUE #1941: fs-38 — cloned Qwen voices 500 on first synth: clone_voice caches a bare prompt, not (prompt, language)
﻿## What

`QwenEngine.clone_voice` writes a **bare prompt** into `_prompt_cache`, where the declared type and every other writer store a `(prompt, language)` 2-tuple.

`server/tts-sidecar/main.py:5290`:

```python
with self._cache_lock:
    self._prompt_cache[voice_id] = prompt      # should be (prompt, lang)
```

`_prompt_cache` is declared at `main.py:4123` as `dict[str, tuple[Any, str]]`, and the three other writers (`:5530`, `:5583`, `:5630`) all store `(prompt, lang)`. `_load_voice_prompt` (`:5508-5510`) reads it as `prompt, lang = cached`, so the clone-written entry raises:

```
ValueError: not enough values to unpack (expected 2, got 1)
```

The 1.7B path fails identically, since `_load_voice_prompt_17b` (`:5588`) delegates to `_load_voice_prompt`.

## Impact

**Every freshly cloned Qwen voice returns HTTP 500 on its first synthesis, until the sidecar restarts.** That includes the clone wizard's own completion-screen audition — the first thing a user does after cloning — so the fs-38 happy path fails 100% of the time on the default engine. The on-disk `.pt` is written correctly; only the in-process cache entry is malformed, which is why a restart "fixes" it.

## Reproduced on-box (2026-07-29, SHA `2503bca6`)

1. Cloned three voices via `POST /api/voice-library/clone` — all 200, artifacts correct on disk.
2. `POST /api/voice-library/{uuid}/sample` for each → **500** every time, traceback as above.
3. Restarted the sidecar (clearing the in-process cache); the *identical* request → **200** in 50.8s with valid audio.
4. After the fix, cloning a brand-new voice and synthesising it **immediately in the same process** → **200**.

## Why automated coverage missed it

Unit tests mock the engine, and no sidecar pytest exercised clone → synth in one process against the real `_prompt_cache`. Found only by on-box acceptance (register row A1).

## Fix

One line, plus a paired regression test that fails before / passes after.

dudarenok-maker:
﻿## Summary

Fixes a Critical defect found during the fs-38 Wave 3 on-box acceptance run (register row A1): **every freshly cloned Qwen voice returned HTTP 500 on its first synthesis until the sidecar was restarted** — including the clone wizard's own completion-screen audition, i.e. the first thing a user does after cloning.

`QwenEngine.clone_voice` cached a **bare prompt** where `_prompt_cache` is declared (`main.py:4123`) as `dict[str, tuple[Any, str]]` and every other writer (`:5530`, `:5583`, `:5630`) stores `(prompt, lang)`. `_load_voice_prompt` (`:5510`) unpacks two values and raised `ValueError: not enough values to unpack (expected 2, got 1)`. The 1.7B path failed identically via `_load_voice_prompt_17b` (`:5588`).

The on-disk `.pt` was always correct — only the in-process cache entry was malformed, which is why restarting the sidecar appeared to "fix" it.

One-line source change:

```diff
-            self._prompt_cache[voice_id] = prompt
+            self._prompt_cache[voice_id] = (prompt, lang)
```

An audit of all `_prompt_cache` write sites confirms this was the only one with the wrong shape; the remaining entries at `:4852`, `:4924`, `:5211`, `:5447` are evictions, not writes.

## Test plan

- **New regression test** `server/tts-sidecar/tests/test_qwen_clone_voice.py::test_clone_voice_cache_entry_survives_a_load_voice_prompt_round_trip` — exercises the real `_prompt_cache` round trip (clone write → `_load_voice_prompt` read) and asserts the cache-hit path is taken, not merely that no exception was raised.
- **Verified it fails for the right reason**: with the one-line fix reverted, the test fails with the exact production symptom, `ValueError: not enough values to unpack (expected 2, got 1)` at `main.py:5510`. With the fix applied, it passes.
  - Note: the first draft of this test was a placebo — the shared fixture's fake `create_voice_clone_prompt` returns a 2-key dict, which unpacks into two values without raising, so it passed *with the bug present*. It was rewritten to monkeypatch the real shape (a length-1 list, per the comment at `main.py:5908-5911`).
- `npm run test:sidecar` — **exit code 0** on two independent full runs, no `F`/`E`.
- **Live on-box verification** (2026-07-29, dual-GPU box, real sidecar + real Qwen weights): cloned a brand-new voice and synthesised it **immediately in the same sidecar process** — the exact sequence that previously always 500'd — and got **HTTP 200** with valid audio in 54.8s. Before the fix, three separate cloned voices 500'd on every attempt and only recovered after a sidecar restart.

Closes #1941

### DIFF
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -5287,7 +5287,7 @@ class QwenEngine(Engine):
         # Warm the in-memory cache + drop any stale 1.7B-native prompt, OUTSIDE
         # the synth lock (mirrors design_voice's post-lock cache handling).
         with self._cache_lock:
-            self._prompt_cache[voice_id] = prompt
+            self._prompt_cache[voice_id] = (prompt, lang)
         self._evict_17b_prompt(voice_id)
         log.info("Cloned + cached Qwen voice '%s' from caller clip.", voice_id)
         return SynthResult(pcm=_float_audio_to_int16_le(wavs[0]), sample_rate=int(sr))
--- a/server/tts-sidecar/tests/test_qwen_clone_voice.py
+++ b/server/tts-sidecar/tests/test_qwen_clone_voice.py
@@ -84,3 +84,70 @@ def test_clone_voice_rejects_missing_body_and_headers(fake_qwen_runtime) -> None
                        headers={"X-Sample-Rate": "24000", "X-Ref-Text": _b64("t")}).status_code == 400
     assert client.post("/qwen/clone-voice", content=_pcm(),
                        headers={"X-Sample-Rate": "24000", "X-Voice-Id": "q"}).status_code == 400
+
+
+def test_clone_voice_cache_entry_survives_a_load_voice_prompt_round_trip(
+    fake_qwen_runtime, monkeypatch
+) -> None:
+    """Regression: clone_voice's post-lock cache warm (main.py ~5290) must
+    store the SAME `(prompt, language)` 2-tuple shape as design_voice's and
+    _load_voice_prompt_17b's writers — _load_voice_prompt (~5508-5510)
+    unconditionally unpacks `prompt, lang = cached`. clone_voice used to
+    stash the bare prompt object instead of a tuple, so the very next
+    /synthesize call against a freshly-cloned voice (still warm, same
+    process) raised `ValueError: not enough values to unpack (expected 2,
+    got 1)` and returned an HTTP 500 — restarting the sidecar 'fixed' it
+    only because that wipes the in-memory cache, proving the on-disk .pt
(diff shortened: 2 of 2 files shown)


dudarenok-maker:
Release-notes step (CLAUDE.md step 5) is **explicitly waived** for this PR, not overlooked: voice cloning sits in the unreleased v1.15.0 section, so this defect never reached a shipped build and there is no user-visible delta versus v1.14.0 to describe. Recording the waiver rather than silently omitting it.

Issue #1969, opened 2026-07-30
Reassigning a character's voice keeps scoring it against the old voice's persisted audition centroid


dudarenok-maker:
## What happens

When a character's voice is **reassigned**, the render-integrity gate keeps
scoring that character against the **old** voice's persisted reference centroid.
Every line of the new voice is then measured as drift from a speaker it was
never supposed to match, and lines fall below the severe cutoff and are flagged
`voice-mismatch` / `severity: severe`.

Found during the fs-38 Wave 3 on-box acceptance run (2026-07-31, `main` @
`b5479e9c`) while running register row **A24**, whose last bullet asks for
"no `voice-mismatch` rows" on the cloned character. There were.

## Evidence

German Coalfall (`castwright__standalones__der-auftrag-von-coalfall`), chapter 2.
`oduvan` was reassigned from its designed voice `qwen-4sURhnDHD6bc_SnUomCQD` to a
cloned library voice, then the character's lines were re-recorded via splice.

The book's pre-existing `audio/render-integrity.centroids.json` row, captured
*before* the reassignment:

```json
{ "characterId": "oduvan", "cleanMean": 0.8388, "pSevere": 0.7852,
  "pBand": 0.7907, "referenceKind": "audition" }
```

The post-splice `render-integrity.json` row for the same character:

```json
{ "characterId": "oduvan", "verdict": "voice-mismatch", "cosine": 0.750,
  "severity": "severe", "referenceKind": "audition", "expectedEngine": "qwen",
  "renderedEngine": "qwen" }
```

0.750 < `pSevere` 0.7852 → severe. The cloned render is fine — measured directly
against its own audition through the production `/synthesize` + `/embed` path it
scores **0.833** (and 0.809 against the human source clip, against a ~0.03
different-speaker floor). It is the *reference* that is wrong, not the audio.

## Why

`CharacterCentroid` (`server/src/audio/render-integrity/centroids-io.ts:19-36`)
persists `characterId`, `centroid`, `cleanMean`, `pSevere`, `pBand`,
`referenceKind` — and **nothing identifying the voice it was built from**. No
`voiceName`, no `modelKey`, no hash.

`resolveCharacterReference`
(`server/src/audio/render-integrity/aggregate.ts:221-223`) then reuses it
unconditionally:

```ts
if (persisted?.referenceKind === 'audition') {
  return { status: 'resolved', ref: persistedAsRef(persisted) };
}
```

There is no equality check against the character's current voice, because the
[...]

Issue #1967, opened 2026-07-30
XTTS cloned-voice derive fails: torchcodec cannot load without shared FFmpeg libraries


dudarenok-maker:
## What happens

Every **cloned-voice derive on Coqui/XTTS** fails on a box whose FFmpeg is a
static build — which is the normal Windows install, and the one the project's
own docs steer you to (`winget install Gyan.FFmpeg`). Found during the fs-38
Wave 3 on-box acceptance run (2026-07-31, `main` @ `b5479e9c`); it blocked all
nine of Section E until worked around.

The only thing the user sees is:

```
Splice failed: Cloned voice(s) unavailable — a cloned voice must never be
substituted with another: "Одуван" (derive-failed). Re-enable Qwen or restore
the missing voice(s); reassign the character(s).
```

No mention of FFmpeg, and the copy says "Re-enable **Qwen**" on a **Coqui**
failure. The never-substitute guarantee holds — the chapter fails loudly rather
than rendering someone else's voice — so this is a hard block, not a silent
corruption.

## Why

`sidecar log` (`logs/tts.err.log`), full chain:

```
main.py:9114 xtts_clone_voice
  → main.py:2502 clone_voice → tts_model.get_conditioning_latents(...)
    → TTS/tts/models/xtts.py:362 → :84 load_audio → torchaudio.load(audiopath)
      → torchaudio/__init__.py:86 → torchaudio/_torchcodec.py:82 load_with_torchcodec
        → import torchcodec
OSError: Could not load this library:
  …/site-packages/torchcodec/libtorchcodec_core8.dll   (then 7, 6, 5, 4 — all fail)
```

**The install docs assert the opposite of what the code does.**
[`docs/wiki/Installing-Castwright.md:41`](../blob/main/docs/wiki/Installing-Castwright.md):

> "the sidecar does all audio I/O via soundfile + ffmpeg and **never calls
> `torchaudio.load`** … recent `coqui-tts` raises at *import* time unless
> `torchcodec` is merely **present**, so its installer drops `torchcodec` in
> (`--no-deps`, so it can't perturb the pinned torch); **it's never actually
> called at runtime**."

It *is* called at runtime, on exactly the Wave 3c cloned-voice derive path.
Because the premise was that the package is dead weight, `--no-deps` leaves its
native FFmpeg dependency unprovisioned. Stock (non-cloned) XTTS voices read the
speaker manifest rather than a file, so they never reach `load_audio` — which is
[...]

Issue #1720, opened 2026-07-19
srv — wrap /load, design_voice, /transcribe, embed in the sidecar capacity admission


dudarenok-maker:
The sidecar PlacementController admission currently wraps only `/synthesize`. To let `SEG_CAPACITY_ADMISSION` default ON safely, the other heavy GPU ops must go through the same reservation so admission fully replaces what the old budget protected:

- cold model `/load` (place on the admitted device + hold during the load)
- `design_voice` (VoiceDesign 1.7B)
- ASR `/transcribe`
- speaker (ECAPA) embed

Until then these ops aren't admitted when the flag is on; while OFF they rely on the sequential workflow + poolWidth=1 + the capacity-based `withGpuLoad` load-path evict (no regression). This is the main blocker for flipping the flag ON.

Context: capacity-aware GPU placement (PR #1719, plan `docs/features/264-vram-aware-gpu-placement.md`, spec `docs/superpowers/specs/2026-07-18-vram-aware-gpu-placement-design.md`). Flag-on-readiness — safe while `SEG_CAPACITY_ADMISSION` is OFF; needed before the flag can default ON. Refs #845.

PR #1971, opened 2026-07-30
docs(docs): record the fs-38 Wave 3 on-box acceptance run 2

dudarenok-maker:
## Summary

Records the **second** fs-38 Wave 3 on-box acceptance run (2026-07-31, `main` @ `b5479e9c`, clean tree) across all three register surfaces. Four more tests discharged plus the core criterion of register row **A24** — and two defects filed.

Docs-only. No shippable delta, so no release-notes entry.

### Discharged

| Test | Result |
|---|---|
| **A24** (register row) — cloned voice renders a non-English book in the book's language | **core criterion PASS**, 2 sub-checks still owed |
| **E-01** ⭐ — clone → Coqui-routed Russian book → generate | **PASS** |
| **E-02** ⭐ — audition, then revoke; Play refuses | **PASS** |
| **E-09** — total erasure of the Coqui artifact paths | **PASS** (first real exercise) |
| E-08 — assign writes both slots | PASS (re-confirmed) |

**A24** was measured over the `/synthesize-batch` wire — the transport the original fix would have missed. Whisper auto-detect on 12 spans of the cloned character: `de` at `avg_logprob` **−0.233**, against a designed-voice control at −0.352 in the same chapter and a pre-fix baseline of `en` / −1.303. Corroborated on the single-synth wire with an identity control, which also reproduces the shipped bug live: German text with the language omitted comes back `en` with unintelligible phonetics. Speaker identity survives the switch — 0.809 cosine against the human source clip, ~0.03 different-speaker floor. E-01 proved the same claim on Coqui: `ru` at −0.368.

**E-09** is the one worth calling out. Run 1 could only mark it `N/A` because `voices\xtts\` had never existed — no XTTS derive had ever run on this box. Run 2's E-01 finally produced artifacts, so this is its first genuine exercise: 5 files across 3 locations pre-revoke, **0 remaining** after, both `voices\xtts\` paths included, entry dir left holding only `voice.json`.

### Two defects filed

**#1967 — blocks Section E on any stock box.** Every XTTS clone *derive* fails because `torchcodec` cannot load without **shared** FFmpeg libraries, and the normal Windows install (`winget install Gyan.FFmpeg`) is a static build shipping `ffmpeg.exe` alone. `docs/wiki/Installing-Castwright.md:41` asserts the sidecar "never calls `torchaudio.load`" and that torchcodec "is never actually called at runtime" — which is *why* the installer drops it in with `--no-deps` and never provisions its native dependency. It is called, on exactly the Wave 3c clone path. Stock XTTS voices read the speaker manifest rather than a file, which is why this stayed invisible until cloning on Coqui shipped. E-01/E-02/E-09 were only reachable after staging PyAV's own bundled FFmpeg set into the `torchcodec` package directory; **E-03…E-07 cannot be attempted on an unpatched box.**

**#1969 — why A24 is not fully discharged, and it corrects A24's own hypothesis.** The row predicted a `voice-mismatch` flood from an English audition reference scored against a German chapter. That cannot happen — `auditionCentroid` does carry the book's language (`audition-centroid.ts:50-57`, plan 275's own change). The real cause has nothing to do with language: a persisted `audition` centroid is reused unconditionally after a character's voice is reassigned, so the clone was scored against the *previous* voice's reference (`pSevere` 0.7852 vs the clone's 0.750 → `severe`). With `SEG_SPK_AUTO_REPAIR=1` shipped on, each affected line burns a re-render every pass and still ends up flagged.

### Also recorded

- **Stack hygiene that mattered.** The sidecar at run start was orphaned and predated both #1962 and #1964. The run sheet now says to check `POST /api/sidecar/restart` returns 200, not 409, before trusting a session — a 409 means nothing is supervising the process under test.
- **A directory that looks alarming and isn't.** `server\tts-sidecar\voices\qwen\` holds 9 stale files; `QWEN_VOICES_DIR` points the sidecar at the per-workspace tree (1115 files). Measure that one.
- **Environment deltas** (run sheet §7.3), including the still-in-place #1967 workaround and its undo, and confirmation that both books were restored — `cast.json` MD5 matches the pre-run file for each.
- A reproducible `/transcribe` 503 `noCapacity` after a render, deliberately not filed (#1720 already records that ASR is not capacity-wrapped).

## Test plan

Docs-only; no code paths touched.

- `npm run check:onbox-register` — passes. Group A stays at 24 rows and the total at 41: A1 and A24 both shrink rather than clear, so no row is removed and the arithmetic is unchanged.
- Run-sheet §7.1 totals recomputed: Section E `P` 1→4, `N/A` 1→0, not-reached 7→5; overall `P` 17→20, `N/A` 2→1, not-reached 33→31. Still sums to 60.
- The **live HTML twin was updated in place** at its canonical URL (`adf22b7b-…`), not re-published. Derived figures recomputed: "still owed in A1" 44→40. Its `42 owed` / group-A `25 rows` are deliberately left alone — the twin carries an **A25** row from a concurrent plan-273 branch that has not merged to `main`, and clobbering it was the failure mode to avoid.
- Plan 275 stays `status: active` — `stable` needs A24's remaining sub-checks, one of which is blocked on #1969.

Refs #1967
Refs #1969

https://claude.ai/code/session_0131F3Fpiwkvt8hJwLjuBUKt

### DIFF
--- a/docs/features/275-clone-voice-language.md
+++ b/docs/features/275-clone-voice-language.md
@@ -400,3 +400,52 @@ that synthesise the same voice disagreeing with it.
 ## Ship notes
 
 (To be filled when status flips to `stable`.)
+
+**Not yet `stable`** — the code shipped in **PR #1964** (merge `b5479e9c`,
+2026-07-30), but register row **A24** is only partly discharged.
+
+### On-box acceptance, run 2026-07-31 (SHA `b5479e9c`, clean tree)
+
+**§"On-box acceptance" step 1 — PASS.** German Coalfall ch.2, cloned voice
+`563501c7-…` cast onto `oduvan`, re-recorded via splice so the render went over
+the **`/synthesize-batch`** wire — the transport the original fix would have
+missed. 12 spans (27.2 s) through `/transcribe` with **no `x-language`**:
+
+| Audio | detected | `avg_logprob` |
+|---|---|---|
+| Cloned `oduvan` | **`de`** | **−0.233** |
+| Designed `narrator` control, same chapter | `de` | −0.352 |
+| Pre-fix baseline (2026-07-30) | `en` | −1.303 |
+
+`characterSnapshots.oduvan.resolvedVoiceName` stayed `qwen-563501c7-…`.
+
+Corroborated on the single-synth wire, with an identity control the criterion
+did not ask for but which matters — the fix changes *what language the model is
+told to speak*, so it is worth knowing whether the cloned timbre survives it:
+
+| Call | detected | `avg_logprob` | cos vs source clip |
+|---|---|---|---|
+| English text + `language: English` | `en` | −0.258 | 0.865 |
+| German text + `language: German` | **`de`** | −0.699 | **0.809** |
+| German text, language omitted (pre-fix) | `en` | −0.904 | 0.876 |
+
+Row 3 reproduces the shipped bug live: German input, English phonetics,
+transcript garbage. Identity holds at 0.809 against a ~0.03 different-speaker
+floor. **The same claim also holds on Coqui** — run 2's E-01 rendered the
+Russian Coalfall with this clone and measured `ru` at −0.368.
+
+**Step 2 — NOT RUN.** The designed-self-heal → restart → re-render comparison
+is still owed.
+
+**Step 3 / C-17 — NOT RUN.**
+
+**The §"On-box acceptance" QA sub-check FAILED, and this plan's stated cause was
+wrong.** The register row predicted a `voice-mismatch` flood from an English
+audition reference scored against a German chapter. That cannot happen:
+`auditionCentroid` does carry the book's language (`audition-centroid.ts:50-57`,
+this plan's own change). The actual cause is unrelated to language — a persisted
+`audition` centroid is reused unconditionally after a character's voice is
+reassigned, so the clone was scored against the *previous* voice's reference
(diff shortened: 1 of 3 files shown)


dudarenok-maker:
### Update — C-17 ran after this PR opened, and **failed**

C-17 ⭐ (`§2.3 designed self-heal + persona survives`) was executed in the same session and is now recorded as the sheet's **first `F`**, filed as **#1972**.

A **designed** voice whose `.pt` was missing did not self-heal from its retained clip, and the chapter then rendered that character's lines **in the narrator's voice** — while `characterSnapshots.<id>.resolvedVoiceName` recorded the *assigned* voice and **nothing was logged**. The chapter completed and every on-disk artifact claimed the right voice; only the sidecar's own synth log disagreed, naming `qwen-fDtxqBAQEy9Os1LA5yVUo` (the Narrator's designed voice) on four synths whose durations match the character's four segments one-for-one. Probed directly with the missing voice, the sidecar returns a clean `409 voice_not_designed` — so the substitution is **Node-side**.

This is the failure mode 268 Invariant 8 and the plan-149 persona guard exist to prevent, and no automated suite was positioned to see it.

Totals recomputed: C section gains its first `F` (not-reached 17 → 16); overall `F` 0 → 1, not-reached 31 → 30. Register row A1 now reads **21 of 60 run · ~39 owed · first FAILURE**, and the live twin was updated in place ("still owed in A1" 40 → 39).

Refs #1972

Issue #1977, opened 2026-07-31
ops-49 — reconcile contradictory coqui-tts availability claims in two sidecar test docstrings


dudarenok-maker:
Two sidecar test files state mutually exclusive things about whether `coqui-tts` is present on every install. Both are load-bearing docstrings that a future reader will act on.

**Claim A** — `server/tts-sidecar/tests/test_coqui_import_pin.py:34-38` (and again at `:105-111`):

> `coqui-tts` ships as an ordinary dependency so the package is always present … `_coqui_package_installed()` is true on EVERY install — package-presence alone would tax everyone.

**Claim B** — `server/tts-sidecar/tests/test_requirements.py:136-142`:

> Re-tier: Coqui is opt-in — it must NOT appear in any overlay (nvidia, amd, or cpu).

…asserted mechanically for all three overlays.

Claim B is enforced by a passing test, so Claim A's premise looks stale — probably left over from before the engine re-tier made Coqui opt-in. If so, the weights-presence gate that Claim A justifies may still be the right design for a different reason (the weights are the real "this install uses Coqui" signal regardless), but the stated rationale is wrong and the surrounding cost argument (14.6 s vs 2.9 s socket-open delay) is attached to a false premise.

## What

Establish which is true today, correct the false docstring(s), and — if the gate's rationale changes — make sure the reason recorded matches the reason the code actually has.

## Acceptance

- The two files no longer contradict each other.
- Whichever claim is wrong is corrected rather than deleted, so the history of why the gate exists survives.
- No behaviour change intended; existing tests stay green.

## Key files

- `server/tts-sidecar/tests/test_coqui_import_pin.py`
- `server/tts-sidecar/tests/test_requirements.py`

## Benefit

*Benefit (technical):* removes a contradiction that would send the next person debugging sidecar startup cost down a wrong path — and one of the two docstrings is currently teaching a false fact about how installs are shaped.

Found in passing while shipping #1967; deliberately left out of that PR's scope (its design spec §15 records the deferral).

PR #1978, opened 2026-07-31
fix(sidecar,server): derive XTTS cloned voices without torchcodec

### LINKED ISSUE #1967: XTTS cloned-voice derive fails: torchcodec cannot load without shared FFmpeg libraries

dudarenok-maker:
## Summary

Closes #1967.

Every cloned-voice derive on Coqui/XTTS failed on a box whose FFmpeg is a static build — which is the normal Windows install (`winget install Gyan.FFmpeg`), and the one our own docs steer deployers to. `CoquiEngine.clone_voice` writes its reference PCM back out to a temp WAV purely because `get_conditioning_latents` wants a path, and XTTS then re-decodes that path through `TTS.tts.models.xtts.load_audio` → torchaudio's loader → torchcodec, which needs FFmpeg's **shared** libraries. The gyan.dev `full_build` ships `ffmpeg.exe` and no `avcodec-*.dll`. Stock XTTS voices read the speaker manifest instead, which is why this stayed hidden until Wave 3c shipped cloning on Coqui. It blocked all nine items of the fs-38 Wave 3 acceptance run's Section E.

New `server/tts-sidecar/xtts_audio_io.py` decodes the reference WAV itself with the stdlib `wave` module + NumPy — the pair `main.py` already uses for all its audio I/O — and `patched_xtts_load_audio()`, a scoped context manager, swaps it in for the duration of the derive only, **inside the existing `_synth_lock`**, so mutating a third-party module global can't race a concurrent synth. It raises rather than silently falling through if `load_audio` goes missing or its signature drifts, naming the installed `coqui-tts` version — a silent fall-through would restore this bug on exactly the boxes that cannot notice.

No new dependency, and no version pin: `coqui-tts` is pinned in no file, and a pin would have been profile-blind and invisible to the requirements tests. The guarantee comes from `install-coqui.mjs` instead, which now round-trips a generated WAV through the real patched loader after the pip loop and **before** the ~1.8 GB weight prefetch, exiting 1 if a future release moves or reshapes the loader — so a break costs the user pip time, not a gigabyte download plus a mystery failure at first derive.

Separately, the user-facing copy told you to "Re-enable **Qwen**" after a **Coqui** failure. `derive-failed` now gets its own remedy clause, an engine is only named for re-enabling when one was genuinely reported `engine-unavailable`, and all **four** `derive-failed` producers tag the entry with its engine — including the classifier path, which is the one that fires on every run after the first, once a permanent failure has persisted `status: 'failed'`.

Three guardrails were blind to all of this and are corrected rather than weakened: `test_audio_io_invariant.py` and `test_requirements.py` (both scan only *first-party* source, so a call inside a third-party package we invoke is invisible to them), and `2026-07-04-fs38-voices-library-design.md:196`, which specified a `soundfile` read that the implementation diverged from — the real recurrence vector. That divergence is annotated, not rewritten; it is the record.

**Design:** `docs/superpowers/specs/2026-07-31-xtts-clone-torchcodec-ffmpeg-design.md` (v4, three adversarial review rounds).
**Plan:** `docs/superpowers/plans/2026-07-31-xtts-clone-torchcodec-ffmpeg.md` (v2).

## Test plan

- `server/tts-sidecar/tests/test_xtts_audio_io.py` — 11 tests. The regression gate is `test_patched_derive_survives_poison_where_unpatched_dies`: a meta-path finder makes `import torchcodec` raise, and it asserts **both** halves — torchaudio's own loader must fail, and ours must succeed, through a derive-shaped call reaching the module global exactly as `get_conditioning_latents` does. Asserting only the success half would pass with the fix entirely absent, which is the shape that let this ship. Also fidelity (our tensor vs. the loader we replace, `atol=1e-6`), signature-drift raising, and a tier that fails if upstream stops needing the patch.
- `test_xtts_clone_voice.py::test_derive_runs_with_the_patched_loader_installed` — asserts the live loader **from inside** the fake `get_conditioning_latents`, plus that `_synth_lock` was already held at that moment. Proven non-placebo: replacing the lock with `nullcontext` turns the assertion red.
- `install-coqui-steps.test.ts` — 5 new `it` blocks incl. one that reads `install-coqui.mjs` and asserts the verification is *invoked before* the prefetch, not merely exported. The snippet was also executed for real against the venv (`[install-coqui] clone-path verify ok`, exit 0) — the string assertions alone would pass on a snippet that crashes.
- `clone-voice-resolver.test.ts` — 5 new tests incl. the untagged persisted-failed-slot path, and the six existing `derive-failed` literals now carry the engine each fixture actually drives (not uniformly coqui — `synthesise-chapter-derive-vram-partition.test.ts:832` drives qwen).
- Branch battery green: `npm run typecheck`, `npm run verify:fast:branch` (lint / typecheck / test:hooks / test:server / test:sidecar / build), full sidecar pytest, 6004 server tests.

- **A mid-sentence capital in the new remedy copy** (`d1ced2ae`). Both new clauses were capitalised on the assumption only one fires at a time; on a mixed `[revoked, derive-failed]` list both do, giving `…voice(s); Re-run the clone…`. The derive-failed clause is now capitalised only when sentence-initial. Paired test added.
- **The shared clone fixture didn't seed `TTS.tts.models.xtts`** (in `a250739e`). Once the derive imports it, all ~30 `clone_voice`-driving tests died with `ModuleNotFoundError: No module named 'TTS.tts'`. Seeded in `_install_fake_coqui_runtime` rather than per-test — a direct consequence of this PR's own change, so fixed here.

## Not done, deliberately

- **On-box acceptance is owed, not run** — register row **A26**, run sheet criteria, and the live HTML twin all updated in this diff per CLAUDE.md before-shipping step 3. Four items: static-FFmpeg derive succeeds; latent equivalence; install verification passes healthy and fails broken; Pinokio's `import torchcodec` result recorded either way. **The dev box's venv is currently hot-patched** (PyAV's FFmpeg DLLs copied into `site-packages/torchcodec/`), so items 1–2 cannot be honestly run against it without reverting that patch — which returns the box to broken while fs-38 Wave 3 is mid-run. Scheduled with the repo owner rather than done opportunistically.
- **No new `docs/features/` plan or `INDEX.md` entry.** This is changed behaviour cited in an existing plan — Wave 3c shipped the clone path that carries the bug — so plan `271` gets a post-ship note instead, per CLAUDE.md before-shipping step 1.
- **A pre-existing lowercase `switch the book to …` after a full stop** on a pure `wrong-engine` list. Confirmed still present; out of this issue's scope, and fixing it generally would change a message with a byte-identical legacy assertion.
- **Contradictory docstrings about whether `coqui-tts` is on every install** — filed as #1977 (`type:chore`) with both citations, rather than resolved here.

> **Teardown hazard:** this branch's worktree has `server/tts-sidecar/.venv` **junctioned** from the primary checkout. Delete the junction with `[System.IO.Directory]::Delete($link, $false)` **before** `git worktree remove`, or the removal can follow it and delete the real venv.


dudarenok-maker:
## Independent review folded (`e1748c46`, `61ee8309`)

High-effort `code-review` pass. **Nothing at Critical.** Two Major findings, both confirmed by mutation, both folded:

**1. The classifier-path engine tag had zero coverage** — the producer this PR's own body calls dominant. The reviewer removed `classification.reason === 'derive-failed'` from the spread and ran `src/tts/`: **109 files, 1540 tests, all green.** So the one line preventing "Re-run the clone for **Qwen**" on a Coqui book, on every run after the first, was unprotected.

**2. The test claiming to cover it was a placebo** — its fixture passed `engine: 'coqui'`, i.e. *tagged*, so it exercised `fromList`'s formatting only and duplicated the test above it while its comment asserted the opposite. It was authored that way **in the plan** and copied faithfully, which is why three spec-review rounds missed it.

Replaced with a producer-side test driving `resolveClonedVoicesForChapter` against an already-failed xtts slot. **Re-verified by mutation after the fold:** removing the `derive-failed` arm now turns exactly that one test red (`1 failed | 96 passed`), where the whole suite previously stayed green.

Also folded: the install verification now distinguishes loader drift from an unrelated crash (marker line printed before the patch is entered, grepped from captured stdout — so a numpy or tempdir failure no longer gets diagnosed as a `coqui-tts` release break); the fidelity test now exercises the resample branch (production always resamples 24000 → 22050, and the old assertion never did); two doc comments this PR had made stale; the run sheet's register pointer (cited A25, which is the unrelated plan-273 row — the row here is **A26**); and two copy/grammar fixes.

**Verified clean by the reviewer, stated for the record:** the `_synth_lock` invariant genuinely holds and does not interact with #1968's worker-thread eviction (nothing on any load/evict path reaches `xtts.load_audio`); decode equivalence is **byte-identical, measured** against this box's real loader — maxdiff `0.0` mono and stereo-downmix; the poison fixture genuinely bites; the install snippet runs and exits 0 for real; all four `derive-failed` producers are tagged with no fifth; and every corrected doc sentence checks out against the code.

**Not folded, filed instead:** #1979 — the offline Help view's static `cloned-voice-broken` remediation omits the new "re-run the clone" remedy. It's a wording decision in a file this branch never touches.

Issue #1973, opened 2026-07-30
ops — release-notes mojibake gate: a narrow false-positive window, and no bypass on the release.yml path


dudarenok-maker:
## What

`findMojibake` (`scripts/release-notes-gate.mjs`, added in PR #1970 for #1956) detects double-UTF-8 encoding generically rather than from a hardcoded list — that part is solid, and was verified against 20 mangled forms including curly apostrophe, NBSP, CJK, Korean, Arabic, Hebrew, Greek, thin space, ZWSP, soft hyphen, 4-byte emoji, emoji+VS16 and Cyrillic, with zero misses. Two bounds on it are worth recording before someone hits them cold.

**1. A narrow false-positive window.** The acceptance rule is "cp1252-mappable characters whose bytes form valid UTF-8 with a lead byte >= 0xC2". Legitimate text can satisfy that. Probed against 23 legitimate strings, four fired:

| Input | Flagged as | Reality |
|---|---|---|
| `é«»` | should be `髻` | legitimate |
| `Ü°` | should be `ܰ` | legitimate |
| `Å§` | should be `ŧ` | legitimate |
| `É™` | should be `ə` | legitimate |

The trigger is any Latin-1 letter mapping to 0xC2–0xF4 (`Â Ã Ä Å Æ Ç È É … ß à á … ï ð ñ ò ó ô`) immediately adjacent, with no space between, to one to three characters mapping to 0x80–0xBF (`— – … ™ • ‘ ’ “ ” § ° © « » ¡ ¿ ± µ ¶ ·`, and NBSP). Realistic shapes: `CAFÉ™`, a `groß—` construction, `Ålesund` next to punctuation.

**Why it matters more than the probability suggests:** `bump-version.mjs`'s pre-flight honours `--force`, so a cut can be pushed through. `release.yml`'s CLI path calls `process.exit(1)` with **no force path at all** — a false positive there blocks the publish outright and the only way out is editing the prose.

Probability against this repo's current content is genuinely low: all six non-ASCII classes the notes files actually carry — `— → … § ≥ ↔ ∩ ⇒ −`, `Русский`, emoji, and the BOM — probe clean, and both committed files pass today.

[...]

Issue #1956, opened 2026-07-29
docs — release-notes-next.md is double-UTF-8-encoded, so the mojibake ships as the public release body


dudarenok-maker:
## What

`docs/release-notes-next.md` is double-UTF-8-encoded throughout. Em-dashes, ellipses and arrows are stored as their mojibake byte sequences rather than the real characters:

```
$ grep -c "â€" docs/release-notes-next.md
63
$ grep -c "â€" RELEASE_NOTES.md
0
```

Examples on `main` @ `2503bca6`:

```
line  2: Draft release notes for the NEXT version (technical register â€” this IS the
line 11: ### â€¦ (new) subsections â†’ emoji-themed sections â†’ bold-lead bullets with
line 39: **Castwright now declares a minimum ffmpeg version â€” 6.0 â€” and checks it**
```

Raw bytes confirm the double encoding: a real em-dash (`U+2014`, `\xe2\x80\x94`) is stored as `\xc3\xa2\xe2\x82\xac\xe2\x80\x9d` — i.e. UTF-8 bytes that were themselves re-encoded as UTF-8 after being read as cp1252.

## Why it matters

This is not a cosmetic docs problem. `docs/release-notes-next.md` is fed **verbatim** as the annotated-tag message by `scripts/bump-version.mjs` → `release.yml`, and it *is* the GitHub release body — the file's own header comment says so. So the next cut publishes `technical register â€" this IS the...` on the public releases page.

`RELEASE_NOTES.md` (the user-facing, brand-voice twin) is clean, which is why this has gone unnoticed: the file people read in-app is fine, and the corrupted one is only rendered at tag time.

v1.15.0 is the in-progress cut, so this ships on the next release unless fixed first.

## Suggested fix

1. Re-encode the file once (decode the mojibake back to proper UTF-8). Mechanical, but it rewrites ~63 lines, so it wants its own commit rather than riding along with unrelated work.
2. Add a guard so it cannot silently recur — a cheap test asserting neither release-notes file contains the `\xc3\xa2\xe2\x82\xac` mojibake prefix. Without a guard this will come back the next time the file is edited by a tool that mis-detects the encoding.
3. Worth checking whether the corrupting step is a script or an editor default, so the fix is at the source rather than the artifact.

## Notes

[...]

PR #2007, opened 2026-07-31
fix(scripts,docs): close two mojibake-gate marker holes

### LINKED ISSUE #1985: ops — a mojibake allowlist marker in RELEASE_NOTES.md never expires
## What

PR #1982 added an in-file allowlist marker to `scripts/release-notes-gate.mjs` so a false positive from the mojibake heuristic cannot unconditionally block a publish:

```
<!-- release-notes-gate: allow "CAFÉ™" -->
```

Part of the argument for the marker over an env-var kill switch was that it **self-expires**: `docs/release-notes-next.md` is reset at each cut, so a marker does not outlive the text that needed it.

That reasoning holds for only one of the two gated files.

- `docs/release-notes-next.md` — reset at each cut, so a marker there does expire. But the reset is a **manual first-PR-after-a-cut step** (documented in CONTRIBUTING.md "Release notes"), not a mechanism. Nothing enforces it.
- `RELEASE_NOTES.md` — **cumulative**, 12 sections and growing. A marker placed there is never removed by any process, so it silences that literal for every future release.

Nothing mechanically prevents a marker being added to the cumulative file.

## Why it matters

The gate exists because 242 double-UTF-8-corrupted spans were queued to publish as the public release body (#1956). A permanent, forgotten allowlist entry in the cumulative file re-opens exactly that hole for one literal, indefinitely, with no reminder that it is there.

Low probability — it requires someone to hit a false positive *in* `RELEASE_NOTES.md` and reach for the marker — but the failure is silent and permanent, which is a bad combination for a guard of this kind.

## Suggested shape

Cheapest real fix: have the gate **refuse markers in the cumulative file** and say so, so the escape hatch exists only where it self-expires. A false positive in `RELEASE_NOTES.md` would then be answered by re-encoding, or by `bump-version --force`, which already downgrades the mojibake failure to a warning.

Alternative, if that is too strict: keep them but have the gate **report** any marker found in `RELEASE_NOTES.md` on every run, so a stale one is visible rather than silent.

## Notes

Found by the independent review on PR #1982 and confirmed while fixing its round-1 findings. Documented in both CONTRIBUTING.md and the release note in that PR, so the behaviour is at least no longer misdescribed — but documentation is not the fix.

### LINKED ISSUE #1990: ops — mojibake-gate fence skipping is backtick-only and parity-dependent
## What

`parseMojibakeAllowlist` (`scripts/release-notes-gate.mjs`) skips allowlist markers inside fenced code blocks, so documenting the marker syntax in a gated file cannot arm it. The skipping is narrower than the docs imply.

Confirmed by execution during the round-2 review of PR #1982 — each of these **arms** a marker that the docs say is inert:

| Shape | Why it slips through |
|---|---|
| A `~~~` fence | only a backtick fence is tracked |
| A 4-space indented code block | no fence to track at all |
| A fence inside a blockquote | the `^\s*` anchor does not match the `>` prefix |
| **Parity flip** | any earlier line merely *starting* with a backtick fence inverts the in-fence state, so a genuinely fenced marker later in the file is harvested |

The parity case is the interesting one, because it needs no unusual markup — an inline-code line at column 0 earlier in the file is enough. The reviewer's repro was a three-line backtick-fenced block followed by a marker, which parsed out as `["IN-FENCE"]` despite sitting inside a fence.

## Why it matters, and why it is not urgent

Neither committed gated file currently parses to a non-empty allowlist, so nothing is armed today. The failure needs someone to write a *realistic* marker literal into `RELEASE_NOTES.md` or `docs/release-notes-next.md` inside one of the shapes above — most plausibly while documenting the syntax in a release note, which is exactly the thing PR #1982 already had to back out of once.

The fail-**closed** directions all behave correctly: an unclosed fence swallows later markers, a marker split across lines yields nothing, an unterminated marker yields nothing, and CRLF is handled.

## Suggested shape

PR #1982 narrows the CONTRIBUTING and release-note wording to name the backtick-fence case specifically, rather than widening the parser, so the docs stop overclaiming. That is the honest stopgap, not the fix.

A real fix is either proper fence tracking (fence-length matching, `~~~`, indented blocks, blockquote prefixes) or — probably better value — dropping fence-awareness entirely in favour of a rule that cannot be armed accidentally. Worth deciding deliberately rather than incrementally patching the fence parser, which is a markdown-parsing problem the gate should not be in the business of solving.

## Notes

Found by the independent round-2 review on PR #1982. Related: #1985 (a marker in the cumulative `RELEASE_NOTES.md` never expires).

dudarenok-maker:
## Summary

Closes two defects in the `#1982` mojibake-allowlist marker (`scripts/release-notes-gate.mjs`), both found by independent review on that PR:

- **#1985 — a marker in `RELEASE_NOTES.md` never expired.** That file is cumulative and never mechanically reset, so a marker there would keep excusing its literal for every future release with nothing to remove it. **Decision: refuse rather than report.** `checkMojibake` now fails outright whenever it finds a `release-notes-gate: allow` marker in a label whose basename is `RELEASE_NOTES.md`, regardless of whether the marker was needed — the failure names the file and both alternatives (re-encode the text, or `bump-version.mjs --force`, which already downgrades a mojibake failure to a warning). A marker in `docs/release-notes-next.md` (or any other label) is unaffected — it still self-expires at the next cut.
- **#1990 — fence-awareness was backtick-only and parity-dependent.** The old `parseMojibakeAllowlist` tried to skip markers inside a ```` ``` ```` fenced block so documenting the syntax couldn't arm it, but the skip only tracked ```` ``` ```` lines by parity — a `~~~` fence, a 4-space indented block, a fence nested in a blockquote, and even an unrelated earlier line merely *starting* with a backtick fence all left a marker armed anyway. **Decision: delete fence-awareness outright rather than patch it again** — proper fence tracking is a markdown-parsing problem this gate has no business owning. Every marker is now honoured unconditionally, and `checkMojibake` returns a new `honoured` field (every literal `parseMojibakeAllowlist` found) that both call sites (`release-notes-gate.mjs`'s CLI and `bump-version.mjs`) echo on **every** run via the new `formatHonouredEcho` helper, pass or fail — e.g. `[allow] docs/release-notes-next.md honoured 2 marker(s): "CAFÉ™", "groß—"`. The property that replaces "markers inside fences are inert" is "an armed marker is never silent."

`CONTRIBUTING.md`'s "Release notes" section is rewritten to match: the marker escape hatch is legal only in `docs/release-notes-next.md`, there is no fence exemption, and every honoured marker is echoed. `docs/release-notes-next.md` gets a technical-register entry for this change.

**Deliberate deviation from the brief:** the brief's docs checklist asked for a matching brand-voice line in `RELEASE_NOTES.md` too. I skipped it, per CLAUDE.md's own before-shipping checklist item 5 ("skip only when the change has no shippable delta... internal chore with no user- or operator-visible effect — say so explicitly"). This PR has zero end-user-visible effect — it only changes release-engineering tooling — and all three prior PRs on this exact feature (#1956, #1973, #1982) never touched `RELEASE_NOTES.md` either. Flagging this explicitly in case the maintainer wants it added anyway.

Closes #1985
Closes #1990

## Test plan

- [x] `npm run verify:fast:branch` — green (lint ran clean; typecheck/config:check/test/test:server/test:sidecar/build all out of scope for a scripts+docs-only diff; `test:hooks`, which runs `scripts/tests/release-notes-gate.test.mjs` and `scripts/tests/bump-version.test.mjs`, passed 769/769)
- [x] `node scripts/release-notes-gate.mjs v1.15.0` — still passes against the committed tree, with no `[allow]` echo (neither gated file currently carries a marker)
- [x] 14 new/rewritten tests added to `scripts/tests/release-notes-gate.test.mjs`, each mutation-verified by reverting the production line it covers and confirming it goes red:
  - RELEASE_NOTES.md refusal — reverting the `isCumulativeReleaseNotesFile` guard in `checkMojibake` turns the refusal test red (and only that test)
  - Label-scoping — forcing `isCumulativeReleaseNotesFile` to always return `true` turns ~20 marker-dependent tests red, proving the refusal is correctly scoped to `RELEASE_NOTES.md` only
  - No-fence-exemption / parity case — restoring the old line-by-line `FENCE_RE`/`inFence` tracking in `parseMojibakeAllowlist` turns the plain-backtick-fence test and the parity-flip test red (the `~~~`/indented/blockquote shape tests stay green either way, since those three already slipped through pre-fix — that's the bug #1990 reports; their new-behavior assertion is the echo, not the arming)
  - Honoured-echo feature — making `formatHonouredEcho` always return `null` turns 4 echo-dependent tests red; separately zeroing out the `honoured: literals` field on the failing-hits return path turns the "still fails" echo test red
  - Fail-closed regressions (marker split across two lines, CRLF) reconfirmed green, unaffected by the above


dudarenok-maker:
## Folded independent-review findings (bundle A) — commit c590a898

Both Majors fixed, five Minors folded, two Minors left alone with the decision recorded in-line. Every new/changed test is mutation-verified against the exact production line it covers — evidence below.

### Major 1 — gate no longer advises a marker it then refuses

`checkMojibake` now short-circuits for `isCumulativeReleaseNotesFile(label)` **before** building the paste-able `<!-- release-notes-gate: allow "..." -->` suggestion, in both the outright-marker-refusal branch and the no-marker-yet hits branch. Reproduced the reviewer's exact scenario first (append `- Ships with a CAFÉ™ badge.` to a sandbox `RELEASE_NOTES.md`, follow the printed advice, re-run) — confirmed the pre-fix "walk into a wall," then confirmed post-fix the first run's message never mentions `add this line` / `<!--` and goes straight to re-encode/`--force` advice.

New test: `a suggestible span in RELEASE_NOTES.md never advises a paste-able marker line` (`scripts/tests/release-notes-gate.test.mjs`).
**Mutation evidence:** reverted the new `isCumulativeReleaseNotesFile(label)` short-circuit block in the hits-found branch of `checkMojibake` (`scripts/release-notes-gate.mjs`) — exactly this one test went red (47/48), nothing else affected.

### Major 2 — the "never silent" echo wiring

Reproduced the reviewer's finding first: deleting all three `formatHonouredEcho` call sites (`release-notes-gate.mjs`'s CLI loop, `bump-version.mjs`'s RELEASE_NOTES.md check, `bump-version.mjs`'s `--notes-file` check) left the suite at exactly **76 pass / 0 fail**, matching the review verbatim.

Investigating the third call site further: **`bump-version.mjs`'s own RELEASE_NOTES.md check (pre-flight 5c) is provably dead code**, not just untested. Its label is always the literal string `'RELEASE_NOTES.md'`, and `checkMojibake` refuses outright as soon as any marker exists for that label — so `mojibakeCheck.honoured` can never be non-empty on that path, with or without Major 1's fix. Confirmed by executing `checkMojibake('...marker...', 'RELEASE_NOTES.md')` directly (honoured always `[]`) and then end-to-end through the real `bump-version.mjs` CLI against a throwaway repo with a marked-up `RELEASE_NOTES.md` + `--force` — the run prints the `#1985` refusal warning and never an `[allow]` line. Rather than fake a positive-echo test for a call site that can never produce one, I removed the dead `formatHonouredEcho`/`if (echo)` pair and added a pinning regression test instead.

Two new spawn-based tests + one pinning test:
- `the CLI echoes an honoured marker in docs/release-notes-next.md on stdout` (`release-notes-gate.test.mjs`) — spawns the real CLI against a throwaway repo.
  **Mutation evidence:** deleted the `formatHonouredEcho`/`if (echo)` pair in `release-notes-gate.mjs`'s CLI loop — only this test went red.
- `bump-version echoes an honoured marker in --notes-file on stdout` (`bump-version.test.mjs`) — spawns `bump-version.mjs --notes-file <marked-up file>`.
  **Mutation evidence:** deleted the `formatHonouredEcho`/`if (echo)` pair at `bump-version.mjs`'s pre-flight 6b — only this test went red.
- `bump-version never echoes a RELEASE_NOTES.md marker — it is refused instead` (`bump-version.test.mjs`) — pins the actual, only-possible behaviour.
  **Mutation evidence (that this isn't a placebo):** broke the underlying refusal invariant (`isCumulativeReleaseNotesFile` forced to always return `false`) — this test went red (the refusal message it expects stopped appearing), proving the assertion has real bite even though the dead-code deletion itself can't be "reverted into a red test" by definition.

Re-ran the reviewer's original mutation with the fix in place, on the two now-real call sites: suite goes from 76/0 to **79/81**, i.e. exactly the two new tests targeting those two sites go red and nothing else does.

### Minors folded

- **Minor 4** (wording): `formatHonouredEcho` now says `literal(s)`, not `marker(s)` (the count is of literals — one marker naming two literals prints 2, and the same literal in two markers double-counts). Updated the tests pinning the exact string and `CONTRIBUTING.md`'s example.
- **Minor 5** (message accuracy): the RELEASE_NOTES.md refusal (both the marker-present and hits-found shapes, now sharing one `reencodeOrForceAdvice()` helper) names re-encoding first and states `--force` is `bump-version.mjs`-only, since `release.yml`'s "Guard — committed release notes are real" step invokes the CLI directly with no bypass. Also reworded `CONTRIBUTING.md`'s matching sentence.
- **Minor 6** (doc accuracy — chose to tighten the code): the refusal now gates on `hasAnyAllowMarker(text)` (any marker at all) rather than `literals.length > 0`, so a marker naming zero literals (e.g. `<!-- release-notes-gate: allow -->`) is refused too, matching "any marker" in `CONTRIBUTING.md`/the header comment.
  New test: `a marker naming no literal at all is still refused in RELEASE_NOTES.md`.
  **Mutation evidence:** reverted the guard back to `literals.length > 0` — only this test went red.
- **Minor 9** (cache wiring): added `scripts/bump-version.mjs` to `test:hooks`'s `extraFiles` in `scripts/verify-cache.mjs` — `bump-version.test.mjs` reads it as text at runtime (mirrors it into a throwaway repo) with no module-graph edge, same #1847 trap `release-notes-gate.mjs` was already guarded against.
  New test: `stepTouchedByDiff: a bump-version.mjs diff matches test:hooks via extraFiles` (`verify-cache.test.mjs`).
  **Mutation evidence:** reverted the `extraFiles` addition — only this test went red.
- **Minor 7** (release-note coherence): `docs/release-notes-next.md`'s #1990 entry no longer restates the RELEASE_NOTES.md refusal in draft-process language ("is the correction described in the #1973 entry above") — it now just points at that entry (#1985) and CONTRIBUTING.md.

### Minors left alone — decision recorded in-line

- **Minor 3**: the CLI's for-loop exits the whole process on the first failing target, so a marker armed in a later target isn't echoed on that run. Publication is blocked either way, so this can't reintroduce a silent arming — narrowed the claim from "on every run" to "on every run that reaches the file" in both the code comment (`release-notes-gate.mjs`'s CLI loop) and `CONTRIBUTING.md`, rather than restructuring the loop to keep checking after a failure.
- **Minor 8**: `isCumulativeReleaseNotesFile`'s basename match stays case-sensitive. On a case-insensitive filesystem a caller passing e.g. `./release_notes.md` would read the real file's bytes but not trigger the refusal — but neither real call site can produce that shape (`release.yml` passes no path; `bump-version.mjs` hardcodes the exact-case label), so it takes a deliberate non-canonical invocation to reach. Left the comparison as-is and expanded the comment on `isCumulativeReleaseNotesFile` to state the limitation and why it's accepted, rather than lowercasing the comparison.

### Verification

- `npm run test:hooks` — 776/776 (was 770 before this round; +6 new tests).
- `npm run verify:fast:branch` — green (lint + test:hooks in scope for a scripts+docs-only diff; typecheck/test/test:server/test:sidecar/build correctly out of scope).
- Every mutation above was performed by editing the production file, confirming the specific new test (and only it) goes red, then restoring the original.

No incidental findings beyond what's already covered by Major 2 (the dead call site, folded in above) — nothing additional to report.

dudarenok-maker:
## Follow-up: precise mutation evidence for Major 1 and Major 2, and the Minor 5/6 disposition

(Addressing specific questions raised during independent re-review of the fold above, commit `c590a898` / merge `00c2c5b1`.)

### Major 2 — each of the three call sites, individually

Round one reproduced the reviewer's claim exactly: deleting all three `formatHonouredEcho` call sites at once (`release-notes-gate.mjs:469-470`, `bump-version.mjs:452-453`, `bump-version.mjs:481-482`) left the suite at **76 pass / 0 fail**, matching the review verbatim.

I then mutated **each site individually**, not the helper or the `honoured` field, and each new test spawns the real CLI/script as a subprocess and asserts on actual `stdout` — so the assertion can only pass if both the `formatHonouredEcho(...)` call *and* the `if (echo) print(...)` line are intact:

- **`release-notes-gate.mjs:469-470`** (the CLI's own for-loop echo) — deleted only these two lines, left `bump-version.mjs` untouched, ran `node --test scripts/tests/release-notes-gate.test.mjs`. Result: exactly one test failed — `the CLI echoes an honoured marker in docs/release-notes-next.md on stdout` — everything else (all `docs/release-notes-next.md`-label tests included) stayed green. Restored, reran, back to all-green.
- **`bump-version.mjs:481-482`** (the `--notes-file` pre-flight 6b echo) — deleted only these two lines, left the RELEASE_NOTES.md pre-flight (5c) and `release-notes-gate.mjs` untouched, ran `node --test scripts/tests/bump-version.test.mjs`. Result: exactly one test failed — `bump-version echoes an honoured marker in --notes-file on stdout`. Restored, reran, back to all-green.
- **`bump-version.mjs:452-453`** (the RELEASE_NOTES.md pre-flight 5c echo) — I could not mutation-verify this one the same way, because it is **unreachable dead code**, not merely untested, given the `#1985` refusal now in force. Its label is always the literal string `'RELEASE_NOTES.md'`, and `checkMojibake` refuses outright as soon as *any* marker exists for that label (before or after Major 1's fix — the refusal predates this PR round). So `mojibakeCheck.honoured` can never be non-empty on that path, with any input. Verified two ways: (1) called `checkMojibake('<marker>...', 'RELEASE_NOTES.md')` directly — `honoured` is `[]` on every branch that returns; (2) ran the real `bump-version.mjs` CLI end-to-end against a throwaway repo with a marked-up `RELEASE_NOTES.md` and `--force` — stdout shows the `#1985` refusal warning, never an `[allow]` line. Since deleting an unreachable line cannot be caught by any positive-echo test (there is no red to turn), I removed the dead `formatHonouredEcho`/`if (echo)` pair instead of writing one, and pinned the actual, only-possible behaviour with `bump-version never echoes a RELEASE_NOTES.md marker — it is refused instead`. To prove *that* test has real bite (not a placebo), I mutated the underlying invariant instead of the deleted lines: forced `isCumulativeReleaseNotesFile` to always return `false`. Result: this one test went red (the `#1985` refusal message it asserts on stopped appearing), confirming the assertion discriminates real behaviour.

Combined re-verification on the fixed tree, mirroring the reviewer's original methodology: deleted the two remaining real call sites (`release-notes-gate.mjs`'s CLI echo + `bump-version.mjs`'s `--notes-file` echo) **at the same time** and reran both test files together. Result: **79 pass / 2 fail out of 81**, and the two failures are exactly the two tests named above — nothing else moved. Pre-fix this same combined mutation (all three, since all three existed then) produced 76/0; post-fix it produces 79/2, i.e. the hole the review found is closed for both call sites capable of ever printing anything, and the third is gone rather than faked.

### Major 1 — the test asserts which advice was printed, not just failure

Both the old and new paths fail (`res.ok === false`) for a genuine mojibake span in `RELEASE_NOTES.md`, so I did not assert on `res.ok`/exit status as the discriminator — that would prove nothing, as noted. The new test (`a suggestible span in RELEASE_NOTES.md never advises a paste-able marker line`) asserts on the **content** of `res.reason`:

```js
assert.doesNotMatch(res.reason, /add this line/);
assert.doesNotMatch(res.reason, /<!--/);
assert.match(res.reason, /refused outright/);
assert.match(res.reason, /re-encode/i);
assert.match(res.reason, /--force/);
assert.match(res.reason, /bump-version\.mjs-only/);
```

That rules out the paste-able marker line appearing at all for `RELEASE_NOTES.md`, and positively confirms the re-encode/`--force` wording appears instead. The other side of the discrimination — that `docs/release-notes-next.md` (and other labels) still get the marker suggestion — is carried by **pre-existing, unmodified tests** that remained green through this change, e.g.:
- `a failing reason names the offending literal and a paste-able marker line` (label `docs/release-notes-next.md`) — still asserts `res.reason.includes('<!-- release-notes-gate: allow "CAFÉ™" -->')`.
- `checkMojibake suggests the full whitespace-delimited token, not just the mojibake core` (label `test.md`) — still asserts `/allow "CAFÉ™"/`.
- `pasting the suggested marker and re-running still flags genuine corruption elsewhere`, `every marker line the gate prints is one its own allowlist accepts`, `no marker is suggested when...` (three straddle/unquotable/standalone-reason tests) — all still green, all exercising the marker-suggestion branch for non-cumulative labels.

Mutation evidence for the new test specifically: reverted the `isCumulativeReleaseNotesFile(label)` short-circuit I added ahead of the suggestion logic in `checkMojibake`'s hits-found branch. Result: exactly one test failed — `a suggestible span in RELEASE_NOTES.md never advises a paste-able marker line` — every `docs/release-notes-next.md`/`test.md` marker-suggestion test (listed above) stayed green, which is itself confirmation that the short-circuit is scoped correctly and doesn't leak into the non-cumulative-label path.

### Minor 5 vs Minor 6 — which was a behaviour change

- **Minor 6 — code tightened (a real behaviour change).** The `RELEASE_NOTES.md` refusal gate changed from `literals.length > 0 && isCumulativeReleaseNotesFile(label)` to `hasAnyAllowMarker(s) && isCumulativeReleaseNotesFile(label)`. Before: a marker naming zero literals (e.g. `<!-- release-notes-gate: allow -->`) slipped through unrefused. After: it's refused too, matching "any marker" as stated in `CONTRIBUTING.md` and the header comment. Dedicated new test: `a marker naming no literal at all is still refused in RELEASE_NOTES.md`. Mutation-verified: reverted the guard back to `literals.length > 0` — exactly this one test went red, nothing else moved.
- **Minor 5 — wording only, no behaviour change.** No gate, branch, or return value changed. Only the *text* of the advice changed: re-encoding is now named first, and `--force` is now explicitly qualified as `bump-version.mjs`-only (since `release.yml`'s own CLI invocation of this gate has no bypass). Both the marker-present refusal and the no-marker-yet hits branch for `RELEASE_NOTES.md` now share one `reencodeOrForceAdvice(label)` helper so the wording can't drift between them. There's no control flow to mutate here — the function always returns the same template for a given label — so this is covered by content assertions rather than a dedicated red/green mutation test: the new Major 1 test's `/re-encode/i`, `/--force/`, `/bump-version\.mjs-only/` matches, plus the pre-existing `a marker in RELEASE_NOTES.md is refused outright, naming the file, the literal, and both alternatives` test, which already asserted `/re-encode/i` and `/--force/` on the refusal reason and still passes (both substrings are still present, just reordered/expanded with the added `bump-version.mjs-only` qualifier).

### Verification

- `npm run test:hooks` — 776/776 on the merged tree (`00c2c5b1`).
- `npm run verify:fast:branch` — green.

No further changes needed; standing down.

Issue #1911, opened 2026-07-28
ops-45 — no golden tier asserts audio CONTENT (right words)


dudarenok-maker:
## What

Neither golden tier asserts that the audio contains the right words.

- **Suite B (assembly)** replays a fixed recorded-PCM fixture through a stub provider — by construction it cannot see anything the engine does. That is correct for its seam (loudnorm / encode / concat timing) and is not a defect.
- **Suite A (model)** runs the real engine, and asserts sample-rate equality, per-line duration (`kokoro-baseline.json`, within tolerance), dead-RMS, voice non-substitution, and length determinism — but nothing about **what was said**.

So a wrong voice reading correct text, or the correct voice reading wrong/truncated/repeated words, both currently pass as long as length, RMS, and voice-id assignment hold. This issue closes the "wrong words" half. Real voice-*identity* verification (a corrupted voice pack, a wrong embedding, or damage under the correct id) needs speaker-embedding similarity and is tracked separately as `srv-47` (currently BLOCKED) — see the implementation-brief comment's §10a for the retitle rationale.

## What makes this cheap now

`faster-whisper` already ships in `server/tts-sidecar/requirements/base.txt` and is installed by the standard sidecar bootstrap (srv-31 / plan 186). It is **CPU-capable** (`ASR_DEVICE=cpu`, zero VRAM), so a transcription check does not compete with synth for the GPU and does not change the tier's hardware requirements.

## Acceptance

- [ ] In Suite A, transcribe each golden line with the already-present Whisper engine and assert the fresh transcript matches a **recorded per-line baseline transcript** at **tolerance 0** — not a WER-vs-fixture-text floor. Full design (reference = recorded transcript, G1/G2 bless guards) is in the implementation-brief comment.
- [x] ~~Keep the existing per-line duration assertion; add per-line RMS so a silent or near-silent render fails loudly rather than passing on duration alone.~~ **Already shipped**, seven weeks before this issue was filed, in `dcc216c1` (2026-06-05, "add golden-audio regression harness (ops-11)") — `MIN_RMS = 0.01` at `test_golden_regression.py:51`. ops-45 adds a producer mutation (M5) proving the guard can still fail.
- [ ] Keep the tier opt-in and triple-gated (venv / pytest / weights), SKIP+exit-0 when absent, exactly as it is now.
[...]

PR #2008, opened 2026-07-31
feat(server,docs): register ASR_MODEL as a config knob and document the ASR rows

### LINKED ISSUE #1988: chore(config): ASR_MODEL is read raw, not a registry knob
`server/tts-sidecar/main.py:6195` reads `ASR_MODEL` straight from the
environment (`os.environ.get("ASR_MODEL", "base")`). `server/src/config/registry.ts`
has entries for `ASR_DEVICE` (:369) and `ASR_IDLE_TTL` (:707) — the sibling
knobs on the same `WhisperEngine` — but no `ASR_MODEL`. This violates
CLAUDE.md's rule that "a new env var MUST be a knob" (registered in
`registry.ts` so it's discoverable/settable through the normal config path,
not a hidden ambient override).

Found in passing while implementing ops-45 / #1911, which reads/writes
`ASR_MODEL` directly (pinning it to `"base"` for the golden gate) — the
missing registry entry is pre-existing and predates #1911, so it's filed
rather than fixed in that PR (a registry addition is a design decision:
default value semantics, UI surface, validation — not a fix-now-bar item).

Found in: ops-45 / #1911 implementation.

### LINKED ISSUE #1989: docs(wiki): Advanced-Settings.md has no rows for ASR_DEVICE / ASR_IDLE_TTL
`ASR_DEVICE` (`server/src/config/registry.ts:369`) and `ASR_IDLE_TTL`
(`server/src/config/registry.ts:707`) are both registered registry knobs, but
a grep over `docs/wiki/Advanced-Settings.md` returns no rows for either.
Violates CLAUDE.md's knob-documentation rule ("a new registry knob isn't done
until docs/wiki/Advanced-Settings.md gets its row").

Found in passing while implementing ops-45 / #1911 (which pins `ASR_DEVICE`
for the new golden content-drift check, and documents the new test-tier-only
`GOLDEN_ASR` / `GOLDEN_REBLESS_CONTENT` flags separately — those are
deliberately NOT registry knobs, since `grep GOLDEN_ registry.ts` returns
nothing and every `GOLDEN_*` flag is test-tier by design). The missing rows
are pre-existing and predate #1911, so filed rather than fixed there.

Found in: ops-45 / #1911 implementation.

dudarenok-maker:
## Summary

Two ASR-configuration gaps filed in passing during ops-45 / #1911, batched into one PR.

**`ASR_MODEL` is now a registry knob (#1988).** The `WhisperEngine` content-QA gate had two registered siblings — `ASR_DEVICE` and `ASR_IDLE_TTL` — while `ASR_MODEL` was read straight from the environment in the sidecar, so the model behind content-QA was invisible to Advanced Configuration and to `.env.example` sync. Registered as `qa.asr.model`:

| | |
|---|---|
| type | `string` — the sidecar passes the value to `faster-whisper` as a model size/name, which is an open set (`tiny`/`base`/`small`/`medium`/`large-v3`, `.en` variants, and any HF repo id), so an enum would be wrong |
| default | `"base"` — matches `main.py`'s own fallback exactly |
| apply | `restart-sidecar`, same as `ASR_DEVICE` |
| risk | medium, same as `ASR_DEVICE` |

**No sidecar code change was needed, and that was verified rather than assumed.** `main.py:6306` already reads it as `os.environ.get("ASR_MODEL", "base").strip() or "base"` — the identical shape `ASR_DEVICE` uses at `:6304` — and `spawn-sidecar.ts:486-492` injects *every* `restart-sidecar` knob's `env` generically. Registering the knob is therefore what makes a UI-set value actually reach the sidecar; nothing else had to move.

**#1989's premise turned out to be wrong, and the issue is closed on that finding.** It reported that `docs/wiki/Advanced-Settings.md` has no rows for `ASR_DEVICE` or `ASR_IDLE_TTL`. Both rows exist — as **"Content-QA (Whisper) device"** (`:235`) and **"ASR (Whisper) idle TTL (s)"** (`:277`). The table lists UI labels and never names environment variables for any knob, so the grep the issue was filed on could not have matched. Only the new **"Content-QA (Whisper) model"** row was genuinely missing; it is added at `:236`, directly under its device sibling.

Also adds the `ASR_MODEL` line to `server/.env.example`, which the registry feeds.

## Independent review follow-up (Major 1)

An independent `code-review` pass (Opus, high effort) found the PR as originally shipped was **not mergeable**: "no sidecar code change was needed" is true for Python and false for Node. Registering `qa.asr.model` made a UI override reach the sidecar, but the server's own Node-side Whisper-model readers — `whisperRepoDir()` / `whisperModelPresent()` / `detectWhisperInstallStateOnDisk()` (`server/src/tts/model-paths.ts`, `server/src/tts/whisper-install-detect.ts`) — still cached `process.env.ASR_MODEL` in a **module-load-time constant**, invisible to a registry override in `user-settings.json`. Before this PR, server and sidecar could not disagree (both only ever read `.env`); this PR is what created the divergence.

**Concretely, reproduced before fixing:** with `qa.asr.model` overridden to `small` and no `ASR_MODEL` env var, `whisperRepoDir()` still resolved to `Systran/faster-whisper-base`. Consequence: Model Manager sized/reported the wrong model, and **clicking Remove deleted `base` and left `small` — the model actually loaded — untouched**, while reporting the freed bytes of the wrong model. The in-app Whisper installer was equally blind: `WhisperInstallBootstrap` spawned `install-whisper.mjs` with no `--model` flag, so it fell back to its own `process.env.ASR_MODEL || 'base'` and could never fetch a UI-configured model.

**Fix:** all three Node-side readers now resolve the model fresh per call via `configValue<string>('qa.asr.model')` instead of a cached const (the resolver still gives env precedence, so a `.env`-only deployment is unaffected). `WhisperInstallBootstrap` now resolves `--model` the same way at spawn time (not at construction, since the server doesn't restart when the knob changes — only the sidecar does).

Also folded in, near-free:
- **Wiki knob count was stale** (`docs/wiki/Advanced-Settings.md`) — said 3 medium-risk knobs in the QA-gates group, verified by execution to be 4 once "Content-QA model" is counted; text and the knob list both corrected.
- **Release note narrated the investigation** (`docs/release-notes-next.md`) — "documented all three... only one was missing" read as diary narration; rewritten to describe the shipped end state, and expanded to also describe the Major-1 fix above.
- **On-box acceptance row added**: `docs/testing/onbox-acceptance-register.md` gets **A28** — the end-to-end path (set a non-default model → sidecar loads it → Model Manager sizes/removes/installs *that* model) needs a real sidecar + HF download that no unit test can substitute for.

**Filed, not fixed here** (out of this PR's fix-now bar — each needs a design decision or wider blast radius):
- A typo'd or oversized `qa.asr.model` value has no guard: `WhisperModel(...)` raises on an unresolvable repo, `/transcribe` returns 500, and `synthesise-chapter.ts`'s ASR-QA call site has no try/catch (unlike the SPK-embed pass 60 lines below, which is explicitly non-fatal) — so a bad value can abort a chapter render mid-flight, or trigger an unbounded HF download mid-render. Bounded today only by `qa.asr.enabled` defaulting `false`.
- No mechanical guard ties a registry knob to a required wiki row — the exact gap that produced #1989 in the first place.
- `ASR_COMPUTE_TYPE` and `ASR_CONCURRENCY` are documented in `.env.example` but absent from `registry.ts` — the same class of gap as #1988, three lines away.

## Test plan

- `server/src/config/registry.device.test.ts` gains a case pinning `qa.asr.model`'s env name, type, apply mode and default, in the same shape as the existing `qa.asr.device` case immediately above it. Mutation-verified: removing the `qa.asr.model` entry from `KNOBS` turns it red.
- **New, mutation-verified against the producer (Major 1 fix):**
  - `server/src/routes/models-inventory.test.ts` — `performRemoval('whisper', …)` with `qa.asr.model` overridden to `small`: asserts Remove deletes the `small` snapshot and leaves a pre-existing `base` snapshot untouched (reverting `whisperRepoDir`'s default back to a literal `'base'` turns this red — freed bytes flip from 300 to 999, the untouched-base assertion fails). A second case pins `buildModelInventory`'s whisper row sizing/`diskPath` against the same override.
  - `server/src/tts/whisper-install-detect.test.ts` — `whisperModelPresent()` and `detectWhisperInstallStateOnDisk()` each get a case where a `qa.asr.model` override makes a present `base` snapshot read as absent/`model-missing` (reverting either function's default parameter back to `'base'` turns the respective case red).
  - `server/src/routes/whisper-install.test.ts` — a capturing `spawnFn` asserts the installer is spawned with `['--model', 'small']` when `qa.asr.model` is overridden (reverting `resolveInstallArgs()` to always return `[]` turns this red).
- Full server suite green: **6,056 passed, 8 skipped, 456 files** (`npm run verify:fast:branch`, which also ran typecheck/lint/config:check/build).
- `npm run check:onbox-register` and `npm run config:check` both green.

Not covered by an automated test, and called out rather than glossed: the full end-to-end path *"change the model in Advanced Configuration → sidecar restarts → `faster-whisper` loads that model → Model Manager reflects it"* needs a real sidecar + HF download — tracked as on-box acceptance row **A28**.

Closes #1988
Closes #1989


dudarenok-maker:
Folded in the independent `code-review` findings. Commit `f2455fa6` on this branch.

**Major 1 — the knob reached the sidecar but not the server's own readers (fixed).** Reproduced first: with `qa.asr.model` overridden to `small` (no `ASR_MODEL` env), `whisperRepoDir()` on the pre-fix tree still resolved to `Systran/faster-whisper-base` — confirmed by temporarily reverting to the original code and running a throwaway assertion against it before touching anything.

Fix: `whisperRepoDir()` (`model-paths.ts`), `whisperModelPresent()` and `detectWhisperInstallStateOnDisk()` (`whisper-install-detect.ts`) now resolve their default via `configValue<string>('qa.asr.model')` per call instead of a module-load-time `process.env.ASR_MODEL` const. `WhisperInstallBootstrap` (`whisper-install-bootstrap.ts`) resolves `--model` the same way at spawn time, not construction time — the server doesn't restart when the knob changes, only the sidecar does, so freezing it in the constructor would have reintroduced the identical bug one level up.

Mutation evidence (each reverted individually, confirmed red, then restored):
- `model-paths.ts`: `whisperRepoDir(model: string = currentAsrModel())` → `= 'base'` turned red both `models-inventory.test.ts` cases (`performRemoval` freed-bytes assertion: expected 300 got 999; `buildModelInventory` sizing case: `present` expected true got false).
- `whisper-install-detect.ts`: `whisperModelPresent(model: string = currentAsrModel())` → `= 'base'` turned red the new override case (expected `false`, got `true`).
- `whisper-install-detect.ts`: `detectWhisperInstallStateOnDisk(..., model: string = currentAsrModel())` → `= 'base'` turned red its override case (expected `'model-missing'`, got `'ready'`).
- `whisper-install-bootstrap.ts`: `resolveInstallArgs()` → `return this.installArgsOverride ?? []` (dropping the `--model` flag) turned red the new `whisper-install.test.ts` spawn-args case (expected `['--model', 'small']`, got a 1-element array with no model flag).

**Minor 2 — wiki count (fixed).** `docs/wiki/Advanced-Settings.md` said 3 medium-risk knobs in the QA-gates group; now says 4 and lists "Content-QA model" alongside the other three.

**Minor 4 — release note narration (fixed).** Rewrote the `docs/release-notes-next.md` bullet to drop the "documented all three... only one was missing" diary narration and describe the shipped end state — now also covers the Major-1 fix above in the same entry, since that's part of this PR's shipped behaviour.

**On-box acceptance row (added).** `docs/testing/onbox-acceptance-register.md` gets **A28**, grouped with the other "no GPU, sidecar venv only" Group A rows: set a non-default `qa.asr.model` → confirm the sidecar loads it, Model Manager sizes/removes/installs *that* model (not `base`), and the in-app installer fetches it. Glance-table count and total bumped (27→28 in Group A, 44→45 total). Did not touch the live HTML twin, per instructions — that update happens once after merge.

**Filed as issues, not fixed here** (each needs a design decision or wider blast radius than this PR's fix-now bar allows):
1. A typo'd/oversized `qa.asr.model` has no guard — `synthesise-chapter.ts`'s ASR-QA call site has no try/catch around the `/transcribe` call (unlike the SPK-embed pass 60 lines below, which is explicitly non-fatal), so a bad value can abort a chapter render mid-flight or trigger an unbounded HF download mid-render.
2. No mechanical guard ties a registered knob to a required wiki row — the exact gap that produced #1989 in the first place.
3. `ASR_COMPUTE_TYPE` / `ASR_CONCURRENCY` are in `.env.example` but absent from `registry.ts` — same class of gap as #1988, three lines away.

I'll file these as separate issues per the standard incidental-findings flow rather than widen this PR's diff further.

Verification: `npm run verify:fast:branch` green (6,056 passed, 8 skipped, 456 files; typecheck/lint/config:check/build all green), `npm run check:onbox-register` green, `npm run config:check` green. PR body's Test plan updated accordingly; `Closes #1988` / `Closes #1989` retained.

Issue #1965, opened 2026-07-30
srv — importable is a find_spec result for kokoro/qwen/whisper too, and the name overclaims


dudarenok-maker:
## What

#1963 fixed Coqui's `livePackageImportable` to prefer the sticky `coqui_import_ok` over the `find_spec` probe. That closes the honesty gap for **one** engine. The investigation done alongside it found the same shape everywhere else, plus a naming problem the fix does not remove.

**1. Kokoro, Qwen and Whisper have no real-import counterpart at all.**

Sidecar side — all three are `importlib.util.find_spec(...)` only, with no `*_import_ok` sticky sibling:

- `server/tts-sidecar/main.py` — `_qwen_package_installed`, `_kokoro_package_installed`, `_whisper_package_installed`

Node side — the same "advertises importable from a probe that never imports" shape #1944/#1963 fixed for Coqui:

- `server/src/tts/voice-engine-registry.ts` — kokoro's `livePackageImportable: (h) => h.kokoroPackageInstalled`, qwen's `(h) => h.qwenPackageInstalled`

Coqui is the engine where this bit, because the speechbrain lazy-proxy collision (#1944) made a `find_spec`-installed package genuinely unimportable. Nothing makes the other three immune — they simply have not hit an equivalent yet. Whether they *should* get a sticky import signal is a real decision: it costs a real import attempt to populate, which is exactly why Coqui's is sticky-null until something tries.

**2. `importable` still overclaims, even after #1963.**

`EngineProbeResult.importable` (`server/src/tts/models-status.ts`) is named for a guarantee it only sometimes provides:

- for kokoro/qwen it is *always* a bare `find_spec` result;
- for coqui it is `find_spec` whenever `coqui_import_ok` is still `null` — which is the common pre-import-attempt case, and the most common wire value.

So a client reading `importable: true` cannot tell whether it means "we imported it" or "a spec exists for it". #1963 deliberately left this alone: renaming a client-facing field is a contract decision, not a drive-by.

## Why it matters

#1944's on-box run is the worked example: `/health` said `coqui_package_installed: true`, a real `from TTS.api import TTS` raised, and that false signal caused a whole section of the fs-38 acceptance run to be scoped as unblocked when it was not. Any consumer trusting `importable` for the three unfixed engines can still repeat that mistake, and any consumer trusting it for Coqui *before* a real import attempt can too.

## Suggested shape

Two separable pieces — the second is worth doing even if the first is declined:

[...]

PR #2009, opened 2026-07-31
fix(frontend,server,openapi): batch deferred nits (loudness guard, narrator-id centralisation)

### LINKED ISSUE #1895: srv — centralise the ['narrator', 'char-narrator'] pair (four independent inline copies)
## What

The pair `['narrator', 'char-narrator']` (the two character-id shapes that
mean "this is the narrator") is inline-copied in at least four places, none
of them exported for reuse:

1. `src/lib/tts-voice-mapping.ts`
2. `server/src/routes/voices.ts` (`isNarratorId`)
3. `server/src/tts/voice-style.ts`
4. Wave 3c's Task 23 narrator-retarget fix (a fourth independent copy,
   added to close a real bug where an orphaned-`characterId` narrator with a
   cloned/designed voice wasn't being looked up by its real cast row)

## Why it's flagged rather than fixed

Confirmed during Wave 3c's Task 23 review to be genuinely cosmetic in that
task's specific context (the equivalent inline copy in `generation.ts` drives
only a reminder-string toggle, not voice resolution or validation) — but the
underlying drift risk is real: four independent copies of the same pair mean
a future change to what counts as "the narrator" (e.g. a new synthetic-id
shape) has to be found and updated in four places, and nothing enforces that
they stay in sync.

Centralising means introducing a cross-module export (none of the four
modules currently export this constant), which is a wider refactor than any
single task's surgical-changes scope justified doing as a side effect.

## Suggested fix shape

Export a single `NARRATOR_CHARACTER_IDS` (or equivalent) constant from
whichever module is the most natural shared home (candidate:
`src/lib/tts-voice-mapping.ts` if a frontend/server-shared location exists,
or a small dedicated module if not), and have all four sites import it.
Needs a decision on where server-side and frontend-side code can share a
constant in this codebase's layering, if that isn't already an established
pattern.

## Key files

- `src/lib/tts-voice-mapping.ts`
- `server/src/routes/voices.ts` (`isNarratorId`)
- `server/src/tts/voice-style.ts`
- The Wave 3c Task 23 narrator-retarget fix (`server/src/tts/synthesise-chapter.ts`
  or wherever that fix landed — locate by symbol/behaviour)

## Related

- fs-38 / #624 (voice cloning umbrella)
- Plan: `docs/features/270-fs38-wave3c-xtts.md`, "Known limitations (m)"

### LINKED ISSUE #1949: chore(fs): three deferred nits from PR #1948 independent review
Deferred (not fixed) findings from the independent review of PR #1948 (plan 274 — loudness measurement provenance). None of these are behaviour bugs today; each is filed cold so it can be actioned independently.

## 1. `classifyDrift` trusts an unknown future `measurementSource` value

`src/components/loudness-report.tsx:70-75` (`classifyDrift`):

```ts
if (lufs.measurementSource === 'ebur128') {
  // fall through to the drift computation below
} else if (lufs.measurementSource === 'loudnorm') {
  return 'no-data';
} else if (lufs.twoPass !== true) {
  return 'no-data';
}
```

If `measurementSource` is ever set to some third value the enum doesn't currently have, it falls through the two explicit checks and lands in the `twoPass`-based grandfather branch — i.e. an *unrecognised* provenance value is trusted as if no provenance were recorded at all, rather than treated as untrusted.

Safer shape: add `else if (lufs.measurementSource !== undefined) return 'no-data';` before the grandfather branch, so only a genuinely *absent* field gets the legacy delegation.

**Priority: low.** `measurementSource` is a closed enum (`ebur128 | loudnorm`) in `openapi.yaml` — this only matters if the enum grows a member later and this call site isn't updated in lockstep.

## 2. Stale comment on an `await` that no longer does anything

`server/src/tts/mp3.ts:479-482`:

```ts
/* Loudness-stats callback fires AFTER the encode succeeds — that way a
   failed encode doesn't leave a `.lufs.json` sidecar describing audio
   that never landed on disk. Awaited so caller-supplied write errors
   surface as rejections from this function rather than unhandled. */
if (pendingSidecar && opts.onLoudnessMeasured) {
  await opts.onLoudnessMeasured(pendingSidecar);
}
```

"Awaited so caller-supplied write errors surface as rejections" describes a caller that writes the sidecar inside the callback. After plan 274, the only caller's `onLoudnessMeasured` is a one-line capture (no write) — see `server/src/audio/finalize-chapter-write.ts`. The `await` itself is still correct/harmless (keeps ordering deterministic and would still propagate an error if one occurred), but the comment's rationale is stale and should be corrected or removed to match what the current caller actually does.

## 3. `openapi.yaml`'s `ChapterLoudness.i/lra/tp` prose still says "post-write"

`openapi.yaml` (`ChapterLoudness.i` / `.lra` / `.tp` descriptions) still read "post-write … after it lands on disk." PR #1948 updated the equivalent prose in `server/src/tts/loudnorm.ts` and `server/src/tts/measure-loudness.ts` to "at the temp path, before the atomic rename" (the measurement is taken on the temp file prior to rename, not after final landing), but left the OpenAPI contract's copy unchanged.

Still literally true in effect (the temp file becomes the final file via rename), but inconsistent with the more precise wording now used at the two call sites it's describing. Worth a follow-up pass to align the contract prose with the corrected implementation comments.

---

Filed from PR #1948's independent review round; none of the three block that PR (no behaviour change, all low-risk/doc-consistency in nature).

### LINKED ISSUE #1977: ops-49 — reconcile contradictory coqui-tts availability claims in two sidecar test docstrings

dudarenok-maker:
## Summary

Three unrelated small items batched into one PR to save review/CI cycles.

**1. #1949 — three deferred nits from PR #1948's independent review**
- **(a)** `src/components/loudness-report.tsx`'s `classifyDrift` grandfathered an *unrecognised* future `measurementSource` value to the legacy `twoPass`-only rule, the same as a genuinely *absent* field — trusting an untrusted provenance value. Added an explicit guard so only `undefined` reaches the grandfather branch. **This is a real (if small) behaviour change**, covered by a new paired test and release-notes entries (technical + user-facing).
- **(b)** `server/src/tts/mp3.ts`'s comment on the `onLoudnessMeasured` await described a caller that writes the sidecar inside the callback. After plan 274 the only caller (`finalize-chapter-write.ts`) is a one-line capture with no write. Corrected the comment to describe the current caller; the `await` itself is unchanged (still correct — keeps ordering deterministic). Comment-only, no test.
- **(c)** `openapi.yaml`'s `ChapterLoudness.i`/`.lra`/`.tp` descriptions still said "after it lands on disk"; PR #1948 had already corrected the equivalent implementation comments to "at the temp path, before the atomic rename". Aligned the contract prose and regenerated `src/lib/api-types.ts` (description-only diff). Doc-only, no test.

**2. #1895 — centralise the `['narrator', 'char-narrator']` pair**
The issue's file list had drifted (as warned); a fresh sweep found **3 frontend sites** (`src/lib/principal-cast.ts`, `src/lib/tts-voice-mapping.ts`, `src/modals/rebaseline-modal.tsx`) and **6 server sites across 5 files** (`server/src/analyzer/narrator-identity.ts`, `server/src/store/merge-analysis-cast.ts` ×2, `server/src/tts/synthesise-chapter.ts`, `server/src/routes/voice-style.ts`, `server/src/routes/voices.ts`, `server/src/tts/voice-mapping.ts`) — more than the issue's original 4. Since there's no frontend/server-shared module in this repo (confirmed: no `shared/` dir, no tsconfig path aliases, zero cross-imports), exported **one constant per side**: new `src/lib/narrator-ids.ts` (`NARRATOR_CHARACTER_IDS`) for the frontend, and `server/src/analyzer/narrator-identity.ts` (already the narrator-identity home) for the server — each carrying a comment naming its twin across the boundary. Every site now imports its side's constant instead of re-copying the literal pair. **No behaviour change** — sites using a `Set`, an array, or an inline `||` chain were each converged with a straight, semantics-preserving substitution.

**3. #1977 (ops-49) — reconcile contradictory coqui-tts docstrings**
Read `server/tts-sidecar/requirements/*.txt` directly (not the docstrings) and confirmed `coqui-tts` never appears as an actual dependency line in any overlay — it is genuinely opt-in. Then found that **this was already fixed**, before this branch started: commit `491bfdbd` (PR #1986, "record real import outcomes for kokoro, qwen and whisper", `Refs #1965`) had already corrected both flagged locations in `test_coqui_import_pin.py` (and the equivalent docstring in `main.py`) to state the opt-in reality and explicitly call out the prior wrong claim. `test_requirements.py`'s claim was already correct and remains unchanged. Both files agree with each other and with reality today — **no diff needed**; closing as already-fixed (`491bfdbd` was committed 2026-07-31T06:23:38Z, roughly 2h19m **after** issue #1977 was filed at 2026-07-31T04:04:51Z — nobody linked the two at the time, so the fix landed without anyone marking the issue resolved).

## Test plan

- [x] `src/components/loudness-report.test.tsx` — new test for the unrecognised-`measurementSource` case; mutation-verified (reverted the guard, confirmed red, restored).
- [x] `src/lib/narrator-ids.test.ts` — a convergence test proving `principal-cast.ts` and `tts-voice-mapping.ts` consult the same imported `NARRATOR_CHARACTER_IDS` binding (via `vi.mock` swapping the constant), mutation-verified against both consumer lines independently; plus (added in the review fold-in below) a dedicated `vi.importActual` test asserting the REAL constant's contents, since the file-scoped `vi.mock` meant nothing else in the file ever touched the real module.
- [x] `src/lib/principal-cast.test.ts`, `src/lib/tts-voice-mapping.test.ts`, `src/modals/rebaseline-modal.test.tsx` — new `char-narrator`-id regression cases (previously-uncovered id shape at each site).
- [x] `server/src/analyzer/narrator-identity.test.ts`, `server/src/store/merge-analysis-cast.test.ts` (×2 cases), `server/src/routes/voice-style.test.ts`, `server/src/routes/voices.test.ts`, `server/src/tts/voice-mapping.test.ts` — new `char-narrator`-id regression cases; each mutation-verified (revert substituted line → red → restore → green).
- [x] `server/src/tts/synthesise-chapter.ts`'s substitution is covered by the **existing** Task 23 tests in `synthesise-chapter-cloned-resolver.test.ts` — mutation-verified against those without adding a new test (they already discriminate the exact behaviour).
- [x] Items 1(b)/1(c)/3 are comment/prose-only or no-diff — no test needed, stated explicitly rather than silently omitted.
- [x] `npm run typecheck`, `npm run test` (frontend, full suite, 4470 tests), `npm run test:server` (full suite, 6056 tests), `npm run verify:fast:branch` (lint + typecheck + test + test:server + build) all green.
- [x] `npx madge --circular --extensions ts server/src` — 16 cycles both before and after (pre-existing baseline drift from CLAUDE.md's stated "15"; unrelated to this PR, no new cycle introduced by the new `analyzer/narrator-identity.ts` imports).
- [ ] `test:sidecar` — sidecar venv not bootstrapped in this worktree; no sidecar code changed (item 3 required no diff), so nothing to run there.

**Evidence correction (independent review, 2026-07-31):** the original claim above — that *each* of the nine `char-narrator` regression tests was mutation-verified by reverting its substituted line and confirming red — was false for two of the nine. `src/lib/principal-cast.test.ts`'s and `src/lib/tts-voice-mapping.test.ts`'s fixtures both also satisfied `isNarrator`'s by-name fallback (fixture `name: 'Narrator'`; `warmth: 70`, which both the narrator branch's ≥55 threshold and the non-narrator fallback's ≥60 threshold agree on), so each stayed green even with `NARRATOR_CHARACTER_IDS` mutated to drop `'char-narrator'`. Both are fixed on this branch: the first now uses a non-matching name (`'The Storyteller'`), the second uses `warmth: 56` to straddle the 55/60 gap — the same technique already used correctly in the other seven tests (server twin `voice-mapping.test.ts:103`, and `narrator-ids.test.ts:52`, both `warmth: 56`).

The remaining seven substitution tests were independently re-checked against the same failure mode (a fixture whose name/tone would satisfy the by-name or fallback-pipeline path even with the id-check reverted) and all seven genuinely discriminate:
- `rebaseline-modal.test.tsx`, `voice-style.test.ts`, `voices.test.ts` all use a non-matching name (`'The Storyteller'`).
- `narrator-identity.test.ts` (server `applyNarratorIdentity`) has no by-name fallback at all — the id-check is the only path, regardless of name.
- Both `merge-analysis-cast.test.ts` cases fail correctly under the `['narrator']`-only mutation (one loses its carried-forward rename, the other collapses into its same-name sibling).
- `voice-mapping.test.ts` already used `warmth: 56`.

Re-ran the aggregate the reviewer measured: mutating `NARRATOR_CHARACTER_IDS`'s *contents* to `['narrator']` now reddens **3 tests across 3 files** on the frontend side (`principal-cast.test.ts`, `tts-voice-mapping.test.ts`, `rebaseline-modal.test.tsx` — up from the 1/1 the reviewer found before this fix) and remains **8 tests across 6 files** on the server side, unchanged. Full frontend suite re-run under the mutation: 3 failed / 4451 total, all three expected; restored and reconfirmed 4470/4470 green.

**Second review fold-in round (Minors 4/5/6/9):**
- **Minor 4, fixed** — `src/lib/narrator-ids.test.ts`'s docstring claimed to guard "the constant's own contents", but `vi.mock` is file-scoped/hoisted, so the one existing assertion compared the mock against itself. Added a dedicated test using `vi.importActual` to read the real, unmocked `NARRATOR_CHARACTER_IDS` and assert its actual contents (`['narrator', 'char-narrator']`); mutation-verified (reverted the real constant to `['narrator']`, confirmed **only this new test** goes red — none of the other three frontend `char-narrator` tests catch it, which is exactly why the guard was missing — restored, confirmed green). Renamed the former assertion's title to state what it actually checks (the mock wiring), not what it doesn't.
- **Minor 5, fixed** — `docs/features/252-narrator-localized-folkloric-identity.md` (status `active`) named the pre-refactor `narrator-identity.ts:NARRATOR_IDS` symbol (renamed to `NARRATOR_CHARACTER_IDS` by this PR) at its invariant #1, and its "invariants preserved" site list described independent inline id-literal checks rather than the now-centralised imports. Both updated to the post-refactor shape.
- **Minor 6, fixed** — `docs/features/271-fs38-wave3c-xtts.md` (status `active`), known-limitation (m), was the pre-centralisation state of exactly the pair this PR now centralises (#1895). Marked `"— closed."` following the house pattern item (n) two bullets down already uses, with a pointer to this PR's commit.
- **Minor 9, left unchanged by decision** — `src/components/loudness-report.tsx:76`'s `classifyDrift` uses `!== undefined`, so an explicit `null` would be classified untrusted rather than grandfathered. The field is typed `'ebur128' | 'loudnorm' | undefined` — `null` isn't producible today, matching the reviewer's own note. Kept `!== undefined` rather than widening to `== null`: treating a future explicit `null` the same as absent would be exactly the silent grandfathering #1949's own fix exists to stop, so the stricter form is the one consistent with this PR's own rationale if the type ever drifts. No behaviour change, no test needed.

Also fixed, found in passing: none beyond what the three linked issues already scope.

Closes #1949
Closes #1895
Closes #1977


dudarenok-maker:
Reopening to re-trigger PR-title-lint against the corrected (shortened) title.

dudarenok-maker:
Folded the independent `code-review` findings into commit `6d41f48a`.

**Blocking 1 — `src/lib/principal-cast.test.ts` placebo — FIXED.** Fixture name `'Narrator'` literal-matched `isNarrator`'s by-name fallback, so reverting the id-check left the test green. Changed to `'The Storyteller'` (matching the convention every other new test in this PR already used). Re-verified: mutated `src/lib/narrator-ids.ts`'s `NARRATOR_CHARACTER_IDS` to `['narrator']` — the test's `expect(result.has('char-narrator')).toBe(false)` assertion now correctly fails red under the mutation. Restored, reconfirmed green.

**Blocking 2 — `src/lib/tts-voice-mapping.test.ts` placebo — FIXED.** `warmth: 70` sits above both the narrator branch's `>=55` threshold and the non-narrator fallback's `>=60` threshold, so the two branches agreed regardless of which one fired. Changed to `warmth: 56`, which straddles the gap — the same technique already used correctly in the server twin (`voice-mapping.test.ts:103`) and `narrator-ids.test.ts:52`, both already at `warmth: 56`. Re-verified red under the same mutation, then green after restoring.

**Blocking 3 — evidence claim corrected.** The PR body's Test-plan section no longer claims "each" of the nine substitutions was mutation-verified; it now names the two that weren't, why, and how each was fixed, plus a summary of the independent re-check on the remaining seven (all seven genuinely discriminate — three via a non-matching name, one server function with no by-name fallback at all, two `merge-analysis-cast.test.ts` cases that fail correctly under the mutation, and the server `voice-mapping.test.ts` twin already at `warmth: 56`).

**Re-run mutation evidence** (mutating `NARRATOR_CHARACTER_IDS`'s contents to `['narrator']`):
- Targeted 3-file run (`principal-cast.test.ts`, `tts-voice-mapping.test.ts`, `rebaseline-modal.test.tsx`): **3 failed / 41 passed** — all three expected tests, and only those three.
- Full frontend suite under the same mutation: **3 failed / 4451 total** (up from the reviewer's measured 1/1) — the constant-content mutation the frontend tests exist to catch now actually gets caught by all three of them.
- Restored the constant, reran: **4470/4470 frontend green**.

**Minor 7 — #1977 chronology corrected.** The body said `491bfdbd` "predates issue #1977's filing by a few hours." Measured: `491bfdbd` committed `2026-07-31T06:23:38Z`; issue #1977 created `2026-07-31T04:04:51Z`. The commit **postdates** the issue by ~2h19m. Corrected in the PR body; the substance (already-fixed, closing without a diff) is unaffected.

**Minor 8 — title retitled.** `chore(frontend,server): batch deferred nits …` → `fix(frontend,server,openapi): batch deferred nits …` — leads with the strongest type present (a real `fix(frontend)` behaviour change) and includes every touched scope, including `openapi` (`38ccecf9`).

**Left as-is, reported not fixed (main thread to judge/dispatch):**
- Minor 4 — `src/lib/narrator-ids.test.ts`'s docstring claims to guard "the constant's own contents," but `vi.mock` is file-scoped/hoisted, so that one assertion compares the mock against itself. The file's second test does real convergence work and is unaffected.
- Minor 5 — `docs/features/252-narrator-localized-folkloric-identity.md:35` (and line 27's site list) still names the pre-refactor `narrator-identity.ts:NARRATOR_IDS` symbol, renamed by this PR to `NARRATOR_CHARACTER_IDS`. Plan status is `active`.
- Minor 6 — `docs/features/271-fs38-wave3c-xtts.md:642`, known-limitation (m), is closed by this PR's #1895 work but isn't marked `"— closed."` the way sibling item (n) is.
- Minor 9 — `src/components/loudness-report.tsx:76`'s `!== undefined` classifies an explicit `null` as untrusted where it previously grandfathered it; not producible today per the reviewer's own note, left as a judgment call rather than changed.

Full verification after the fix: `npm run typecheck` (frontend + server) clean, `npm run verify:fast:branch` all-green (lint/typecheck/test/test:server/build cached-pass, config:check + test:sidecar out of scope), pushed to `chore/frontend-deferred-nits` at `6d41f48a`. Not merging per instructions.

dudarenok-maker:
Second fold-in round, addressing the coordinator's follow-up that Minors 4/5/6 needed to land (CLAUDE.md before-shipping step 1 — both docs are `active` plans) rather than just be reported. Commit `c3e65f14`, pushed to `chore/frontend-deferred-nits`.

**Minor 4 — fixed, took the `vi.importActual` option.** `src/lib/narrator-ids.test.ts` had a hard-evidence gap: the mutation run from the last round showed the real frontend constant's contents are netted by three consumer tests, and none of them is the test whose docstring claims to guard "the constant's own contents" — that test only ever saw the `vi.mock` sentinel. Added a new test that calls `vi.importActual<typeof import('./narrator-ids')>('./narrator-ids')` and asserts the real, unmocked `NARRATOR_CHARACTER_IDS` equals `['narrator', 'char-narrator']`. Mutation-verified: reverted the real constant to `['narrator']` and reran the file — **only the new test goes red** (1 failed / 3 total); the pre-existing mock-sentinel assertion and the convergence test both stay green, confirming they genuinely test different things. Restored, reconfirmed 3/3 green. Also renamed the old assertion's title from "is the narrator / char-narrator pair (mocked to a single sentinel id above)" to state plainly that it's checking the mock wiring, not the real constant, so a future reader doesn't repeat the misreading.

**Minor 5 — fixed.** `docs/features/252-narrator-localized-folkloric-identity.md:35` renamed `narrator-identity.ts:NARRATOR_IDS` → `narrator-identity.ts:NARRATOR_CHARACTER_IDS`. Line 27's "invariants preserved" site list was rewritten to describe the post-#1895 shape: which sites now import the centralised constant (server: `routes/voices.ts`, `routes/voice-style.ts`, `tts/voice-mapping.ts`, `store/merge-analysis-cast.ts`, `tts/synthesise-chapter.ts`; frontend: `lib/principal-cast.ts`, `lib/tts-voice-mapping.ts`, `modals/rebaseline-modal.tsx`) versus which sites I verified still key on a standalone `'narrator'` literal only, unaffected by this PR (`analyzer/roster-coverage.ts`, `analyzer/dialogue-structure/name-matcher.ts`, `modals/profile-drawer.tsx` — checked each by grep; none of them reference `'char-narrator'` at all, before or after this PR).

**Minor 6 — fixed.** `docs/features/271-fs38-wave3c-xtts.md:642`, known-limitation (m) (the pre-centralisation "fourth inline copy" finding that is #1895), now reads `"— closed."` following item (n)'s existing pattern two bullets down, with a pointer to this PR (`b913d842`).

**Minor 9 — judgement call: left unchanged.** `loudness-report.tsx:76`'s `classifyDrift` uses `!== undefined`; the field is typed `'ebur128' | 'loudnorm' | undefined`, so an explicit `null` isn't producible today (matches the reviewer's own note) and no test can be written against a value nothing can ever send. I kept `!== undefined` rather than widening to `== null`, on this reasoning: the whole point of the #1949 fix this PR ships is that an untrusted provenance value should be classified as untrusted rather than silently grandfathered as "absent." If the type ever drifts and a future writer starts emitting an explicit `null`, `!== undefined` keeps treating it as untrusted (correct, per this PR's own stated rationale); `== null` would instead start grandfathering it, which is the exact failure mode #1949 fixed elsewhere in this same function. Stated here so the decision reads as considered, not missed.

Full re-verification after this round: `npm run typecheck` clean, `npm run verify:fast:branch` all green (4471 frontend tests, up 1 from the new importActual guard; test:server cached-pass; build pass). Not merging.

Issue #2038, opened 2026-07-31
Japanese cloned-voice text-splitting still broken (SudachiPy not installed)


dudarenok-maker:
## What

Follow-up from #2017. `CoquiEngine._infer_from_latents` (`server/tts-sidecar/main.py`) passes `enable_text_splitting=True` to XTTS's low-level `inference()`, which reaches upstream's `get_spacy_lang('ja')` — that requires SudachiPy + SudachiDict-core, which plain `spacy` does not pull in.

#2017 deliberately installed plain `spacy` (not `spacy[ja]`) after measuring the wheel cost:

- plain `spacy` → 16 packages, 22.3 MB
- `spacy[ja]` → 18 packages, 92.5 MB, of which **68.9 MB is `sudachidict-core`** alone

So every OTHER language `get_spacy_lang` handles (ar/en/es/hi/zh — all verified working with plain spacy) now works for a cloned-voice render at/above `char_limits[lang]`, but a cloned Japanese voice rendering a long line still hits `ImportError` on the primary path. #2017's belt-and-braces catch means it no longer 500s — it degrades to `enable_text_splitting=False`, which risks the *separate* `assert text_tokens.shape[-1] < gpt_max_text_tokens` crash (`xtts.py:516`) on a sufficiently long/unpunctuated Japanese line instead.

## Why this is filed separately rather than fixed in #2017

Whether to pay the 70 MB SudachiDict cost on every install (including Pinokio) for Japanese cloned-voice long-line support is a product/packaging call, not a mechanical fix — and CLAUDE.md already names `fs-59` (CJK support) as the workstream that should own it (see `docs/BACKLOG.md` / the fs-59 design doc for what CJK support currently covers and what it's still missing).

## Acceptance (sketch — the fs-59 owner should refine)

- Decide whether `spacy[ja]` (or a standalone `sudachipy`/`sudachidict-core` pair) ships by default, as an opt-in Model Manager download, or not at all for the v1 CJK surface.
- If shipped, add it to the appropriate requirements layer (base vs. an opt-in overlay) and verify `get_spacy_lang('ja')` actually instantiates.
- Either way, a regression test should pin the CURRENT behaviour (ja falls back to `enable_text_splitting=False` with a loud log) so a future change to this area doesn't silently regress it further.

## Key files

- `server/tts-sidecar/main.py` — `CoquiEngine._infer_from_latents`, `requirements/base.txt`
[...]

Issue #2026, opened 2026-07-31
XTTS Russian quality: neuter -ee mispronounced, leading dialogue em-dash gives no pause, rare language collapse


dudarenok-maker:
Found on the box during the fs-38 Wave 3 acceptance run, while chasing why a cloned Russian voice produced "garbage words". **Cloning turned out not to be the cause** — every defect below reproduces on a **stock catalogue voice (`Damien Black`) with no clone anywhere in the path**, so this is engine-level Russian quality, separate from #2017 and #2023.

Repo owner confirmed defect 1 by ear ("yes, it is АЯ") and raised defect 2 independently ("it also ignores `—`, as it should be treated as pause, not read without stoppage").

All probes: `POST /synthesize`, `engine: coqui`, `model: xtts_v2`, `voice: Damien Black`, `language: ru`, coqui-tts 0.27.5.

---

## 1. Neuter adjective ending `-ее` is rendered as `-ая` / `-ие`

`Хорошее олово.` — two words, no punctuation beyond the full stop — comes out as **"хорошая олова"**. Confirmed by ear, not just ASR.

Six independent runs of the same two words:

| run | heard |
|---|---|
| 1 | Хорошия, Олова |
| 2 | Хорошая, Олова. |
| 3 | хорошая улова. |
| 4 | хорошая олова. |
| 5 | Хорошие оловатад. |
| 6 | *Karosia olleet tämän tämän* (see defect 3) |

**Never once correct.** The controls isolate it to the ending, not the word or the engine generally:

| probe | said | heard | |
|---|---|---|---|
| neuter `-ее` | Хорошее олово. | хорошая олова | ✗ |
| neuter `-ее` | Синее небо. | Синей неба | ✗ |
| neuter `-ее` | Горячее железо. | горячие железанным | ✗ |
| neuter `-ое` | Большое окно. | Большое окно | ✓ (4/4 runs) |
| feminine `-ая` | Хорошая руда. | Хорошая руда | ✓ |
| masculine `-ий` | Хороший металл. | хороший металл | ✓ |
| plural `-ие` | Хорошие мечи. | хорошие мечи | ✓ |

`-ее` fails 3/3 distinct words; `-ое` is mixed (`Большое` fine, `Тёплое` not); every non-neuter ending is correct. Russian neuter adjectives are common in ordinary prose, so this is not an edge case.

## 2. A leading em-dash — the Russian dialogue marker — produces no pause

[...]

Issue #2023, opened 2026-07-31
Orphaned characterId silently renders in the narrator's voice — unrecorded, unsurfaced, affects 5 books


dudarenok-maker:
Found on the box during the fs-38 Wave 3 acceptance run. **Caught by ear first** — the repo owner listened to two supposedly-different characters and said "D and E seem to be narrator voice… it used narrator voice for non-narrator phrases". The measurements had been saying the same thing and I had misread them.

## What happens

When a rendered sentence group carries a `characterId` that is **not in the book's cast**, `synthesise-chapter` falls back to the narrator voice for that line. That fallback is:

- **logged** to `server.err.log` — once per orphan id per render, not once per line;
- **not recorded** in `renderedFallbackByCharacter`, which stays `{}` after a render that substituted 21 segments;
- **not surfaced anywhere in the UI.**

So the listener hears the narrator reading another character's dialogue, and nothing in the app says so. There is no post-hoc way to discover it short of diffing `cast.json` ids against `*.segments.json` ids by hand.

## Evidence — *Заказ Коалфолла*, chapter 2

```
cast.json ids       : [... 'coalfall-dragon', ... 'mairin', ...]
segments.json ids   : [... 'coalfall',        ... 'mayrin', ...]
analysis cache ids  : [... 'coalfall',        ... 'mayrin', ...]
```

`mairin` vs `mayrin` and `coalfall-dragon` vs `coalfall` — the cast and the analysis disagree on the romanisation/slug. Result:

**21 of 72 segments — 60.3 s — rendered in the narrator's voice.**

Log line (fires twice per render, covering all 21 lines):

```
[synthesise-chapter] sentence group references characterId "mayrin" which is not
in this book's cast — falling back to the narrator voice for this line.
```

ECAPA confirms they are literally the same voice — `narrator` vs `mayrin` **0.932**, `narrator` vs `coalfall` **0.965**, `mayrin` vs `coalfall` **0.937**. For scale, the same real person across two different clips measures **0.891** (run sheet B-04), so these "different characters" are more alike than one person is to themselves.

## This is not one book

Scanning every book's rendered `*.segments.json` against its own `cast.json`:

| book | orphaned segments | audio in the wrong voice | orphaned ids |
|---|---|---|---|
| Playing with Fire | 74 / 6,638 (1.1 %) | **2.8 min** | `the-torment`, `pool-player-2`, `lightning-dave` |
[...]

PR #2039, opened 2026-07-31
fix(sidecar): declare spacy so cloned Coqui voices don't 500 on long lines

### LINKED ISSUE #2017: Cloned Coqui voices 500 on any line at/above the per-language char limit — spacy missing from sidecar requirements
Found on the box during the fs-38 Wave 3 acceptance run (run sheet Section E, **E-04** — the test written specifically for this crash class). Reproduced deliberately, not inferred from a log.

## What happens

Every cloned-Coqui synth goes through `CoquiEngine._infer_from_latents`, which hardcodes `enable_text_splitting=True` (`server/tts-sidecar/main.py:2427`). XTTS then does:

```python
# TTS/tts/models/xtts.py:507-510
if enable_text_splitting:
    text = split_sentence(text, language, self.tokenizer.char_limits[language])
else:
    text = [text]
```

and `split_sentence` reaches spacy **only past a per-language length threshold**:

```python
# TTS/tts/layers/xtts/tokenizer.py:43-48
def split_sentence(text, lang, text_split_length=250):
    if text_split_length is not None and len(text) >= text_split_length:
        nlp = get_spacy_lang(lang)   # <- ImportError: spacy is not installed
```

**`spacy` is not installed in the sidecar venv and is not declared in any `server/tts-sidecar/requirements/*.txt`.** So a cloned Coqui voice renders fine up to the limit and returns a bare `500 {"detail":"Internal error."}` at or above it.

## Reproduction (measured 2026-07-31)

Same voice (`xtts-0abceba4-…`), same engine, same `language: "ru"` — only length differs:

| text | length | `POST /synthesize` | result |
|---|---|---|---|
| `Дракон либо подождёт, пока я доварю, либо нет.` | 46 | **200** | 178,176 B PCM = 3.71 s audio |
| same + two more clauses | **245** | **500** | `{"detail":"Internal error."}` |

Sidecar log for the failing call:

```
File "…/TTS/tts/models/xtts.py", line 508, in inference
  text = split_sentence(text, language, self.tokenizer.char_limits[language])
File "…/TTS/tts/layers/xtts/tokenizer.py", line 49, in get_spacy_lang
  raise ImportError("enable_text_splitting=True requires Spacy: pip install spacy[ja]") from e
ImportError: enable_text_splitting=True requires Spacy: pip install spacy[ja]
```

Confirmed absent: `importlib.util.find_spec('spacy') is None`, coqui-tts 0.27.5.

## Why this matters more than the numbers suggest

The threshold is **per language**, and it is tightest exactly where cloned Coqui voices are the whole point — non-English books:

| lang | `char_limits` | lang | `char_limits` |
|---|---|---|---|
| **ja** | **71** | es | 239 |
| **ru** | **182** | en | 250 |
| | | de | 253 |
| | | fr | 273 |

Japanese breaks on most ordinary sentences. Russian breaks on a long one. A book renders *mostly* fine and then fails on its longer lines, which reads as flakiness rather than a missing dependency.

The user-visible symptom is a bare `Internal error.` 500 with no diagnosis — the operator has no way to reach "install spacy" from what they see.

## Why CI did not catch it

`test_xtts_clone_sanity` in the golden-audio tier was written for exactly this crash class, but that tier is opt-in (`npm run test:golden-audio`, not in `test:all`/`verify`) and triple-gated on venv/pytest/weights — it SKIPs and exits 0 when they are absent. So the one test aimed at this never ran.

## Suggested fix

1. Add `spacy` to `server/tts-sidecar/requirements/base.txt` (the `get_spacy_lang` import list needs `ar/en/es/hi/ja/zh`; `spacy[ja]` pulls the Japanese extras the error message names). Confirm licence/wheel size before pinning.
2. Decide the fallback deliberately — if spacy is genuinely unwanted as a dependency, catch the `ImportError` in `_infer_from_latents` and retry with `enable_text_splitting=False` rather than 500-ing. Note the docstring there already explains that the `False` path hits XTTS's `gpt_max_text_tokens` assert, so this is a real trade-off and not an obvious one-liner.
3. Either way, add a paired regression test that drives a >`char_limits[lang]` string through the cloned-latents path — the current golden-tier coverage cannot fail on a box without weights.

Also worth checking whether the same threshold affects the **stock catalogue** Coqui path (`synthesize()` via `tts.tts()`), which reaches `enable_text_splitting=True` through `Xtts.synthesize`'s own default — this report only exercised the cloned-latents branch.

Run sheet: `docs/testing/fs38-wave3-onbox-acceptance.md` §5 Section E, E-04.

dudarenok-maker:
## Summary

`CoquiEngine._infer_from_latents` (`server/tts-sidecar/main.py`) passes `enable_text_splitting=True` to XTTS's low-level `inference()` — config-faithful, mirrors `Xtts.synthesize`'s own build — so a cloned-voice render at or above `tokenizer.char_limits[lang]` doesn't hard-crash the bare low-level assert instead. That path reaches upstream's `get_spacy_lang`, which raises `ImportError` when spacy isn't installed — and spacy was never declared as a sidecar dependency, so **every** such render 500'd (`{"detail":"Internal error."}`).

**Decision (per the implementation brief): Option A, with plain `spacy`, plus Option B's catch as belt-and-braces — not a straight revert to `enable_text_splitting=False`.**

- **Option A, revised placement** — plain `spacy>=3.8,<4.0` (**not** `spacy[ja]`) is installed by `server/tts-sidecar/scripts/install-coqui.mjs`'s third pip step, alongside the existing CJK phonemizers (`pypinyin`/`cutlet`/`unidic-lite`), **not** `requirements/base.txt`. spacy is reached only via the opt-in Coqui/XTTS clone path — never by Qwen or Kokoro — and this repo deliberately keeps Coqui-only dependencies out of the shared manifest (`test_coqui_absent_from_all_overlays` keeps `coqui-tts` itself out of every overlay the same way). Landing it in `base.txt` would have made every CPU-only and Kokoro-only install pay 22.3 MB / 16 packages for a library they never import. Measured in the real sidecar venv: plain `spacy` → 16 packages / 22.3 MB; `spacy[ja]` → 18 packages / 92.5 MB, of which **68.9 MB is `sudachidict-core`** alone (needed only for Japanese). Verified plain spacy actually satisfies `get_spacy_lang`'s ar/en/es/hi/zh branches (zh confirmed working, despite the upstream error message naming `spacy[ja]` generically) — only `ja` needs SudachiPy and stays broken on purpose, filed separately as #2038 (fs-59's call). The manifest guard (`test_spacy_is_explicit_and_not_the_ja_extra`, `server/tts-sidecar/tests/test_requirements.py`) is re-pointed at the new invariant: spacy absent from every manifest/overlay, present in `install-coqui.mjs`, never `[ja]` — verified it actually fails if spacy is put back in `base.txt`.
  - **Known cost, surfaced rather than hidden: this does NOT reach an existing Coqui install automatically.** A `base.txt` dependency is covered by `reqHash`, which re-pips on every version bump — so the original placement would have reached every upgrading user for free. The installer route only runs when a user explicitly (re-)installs or presses **Repair** on Coqui from Model Manager. Anyone who already has Coqui set up keeps hitting the #2017 500 until they do that. Called out in `RELEASE_NOTES.md` (user-facing) and `docs/release-notes-next.md` (operator terms) in this same PR.
- **Option B's catch, belt-and-braces only** — `_infer_from_latents` now catches `ImportError` from the inference call, logs it loudly (`log.error`, never silent), and retries with `enable_text_splitting=False`. Per the brief, this is a fallback for a stale/broken install, not the primary fix; the primary `enable_text_splitting=True` line is untouched.

**Incidental check requested by the brief — the stock catalogue path does NOT share this bug.** `synthesize()`'s `tts.tts()` call goes through `TTS.api.TTS.tts()` → `Synthesizer.tts()`, whose own `split_sentences=True` pre-splits text using `pysbd` (not spacy) *before* ever reaching `Xtts.synthesize()`/`.inference()` — and neither of those ever passes `enable_text_splitting=True` down from that path (`Xtts.inference()`'s own signature default is `False`). So the catalogue path never touches `get_spacy_lang` at all; this bug is confined to the cached-latents (`_infer_from_latents`) branch. No new issue filed for that.

**Other Coqui-provisioning paths checked — none need this change.** `install-coqui.ps1`/`.sh` (the scripted/offline siblings) only pre-fetch XTTS weights via `TTS.api.TTS(...)`; they don't pip-install anything and assume `coqui-tts` (and its deps) is already in the venv, so they're not an independent install path. The in-app installer route (`server/src/routes/coqui-install.ts` + `coqui-install-bootstrap.ts`) spawns `install-coqui.mjs` directly. No Pinokio-specific Coqui installer, no separate "Repair" code path, and no CI/test-setup script independently installs `coqui-tts` exist in this repo.

## Test plan

- [x] `npm run verify:fast:branch` — green (pre-push gate; `test:sidecar` ran for real, 449s, 0 failures)
- [x] New regression test `test_synthesize_cloned_voice_recovers_when_spacy_import_fails` (`server/tts-sidecar/tests/test_xtts_clone_voice.py`) — drives a >182-char Russian string (`char_limits['ru']`) through the cached-latents path via a fake `tts_model.inference` that raises `ImportError` on the `enable_text_splitting=True` attempt (mirroring the real upstream shape) and succeeds on the `False` retry. Verified fail-before/pass-after by temporarily reverting the `main.py` catch and re-running: fails with the raw `ImportError` before the fix, passes after.
- [x] Full `pytest -m "not golden"` sidecar suite — 0 failures (2 pre-existing skips), confirming the new `spacy` dependency doesn't break anything else.
- [x] `test_requirements.py`'s `test_spacy_is_explicit_and_not_the_ja_extra` re-pointed at the new invariant and verified to fail when spacy is restored to `base.txt`.

Closes #2017


dudarenok-maker:
## On-box acceptance state (Before-shipping checklist step 3)

This PR ships behaviour only real hardware can prove — a cloned Coqui voice rendering a line at or above `char_limits['ru']` needs real XTTS weights and a live sidecar. So step 3 applies, and this note records where the acceptance stands.

**The row already exists and stays open: `docs/testing/onbox-acceptance-register.md`, Group E, row E-04 (`F`), citing #2017.** It is not being removed or downgraded by this PR.

Nothing here discharges it. The fix is verified by unit test against a fake that reproduces the upstream `ImportError` shape; it has **not** been re-run on the box against real weights. Per the register's own rule, a row comes out only when the acceptance was actually run and the outcome recorded, or the repo owner confirms it was exercised on a live book — "tests pass, so it's presumably fine" never removes a row.

**What changes for whoever picks E-04 up:** it is no longer blocked on a fix. The remaining work is a re-run of the E-04 reproduction from `docs/testing/fs38-wave3-onbox-acceptance.md` §5 Section E — the 46-char control expecting 200, and the 245-char Russian line that previously returned `{"detail":"Internal error."}` and should now return 200 + PCM — on a box with Coqui weights and a provisioned venv.

**Why the register file itself is untouched in this diff.** E-04 is also the row cited by #2026, which is batch 3 of the same fs-38 Wave 3 run and is still in flight. Two concurrent branches editing the same register row is a known failure mode in this repo: the edits auto-merge without conflicting and the file's internal arithmetic silently drifts, which is exactly what `npm run check:onbox-register` exists to catch after the fact. The register, the run sheet, and the live HTML twin linked from the register's header will be updated **once, coherently**, when the fs-38 Wave 3 batches have landed and the state is settled — rather than three times from three branches racing each other.

Flagging this deliberately rather than absorbing it: it is a considered deviation on *where* the bookkeeping lands, not a skipped step.

dudarenok-maker:
## Correction to the previous comment

My note above stated that `docs/testing/onbox-acceptance-register.md` carries an open row "Group E, row E-04, marked `F`, citing #2017". **That is wrong** — I passed on a subagent's report without checking it. The register contains no `E-04` row, and no mention of `#2017` or `spacy` anywhere. `E-04` is a *run-sheet* test ID (`docs/testing/fs38-wave3-onbox-acceptance.md` §5 Section E), not a register row ID.

**The conclusion still holds, for a different reason.** The register tracks this debt at group granularity, not per test:

> `### A1 · fs-38 Wave 3 — voice cloning (now incl. 3c) · 20 of 60 run (2026-07-29, 2026-07-31) · ~40 still owed`

and within it, explicitly: *"Section E — 4 of 9 now run (2026-07-31); E-03…E-07 still owed"*. E-04 sits inside that owed set. So Before-shipping checklist step 3's "add a row in the same PR" is satisfied by an existing row — this PR does not ship undocumented debt, and the merge is not blocked on a register edit.

What I said before about the *substance* is unchanged: nothing here discharges the acceptance, the fix is proven only against a fake reproducing the upstream `ImportError` shape, and E-04 still needs a real on-box re-run (46-char control → 200; 245-char Russian line → 200 + PCM instead of `{"detail":"Internal error."}`) on a box with Coqui weights.

## Two pre-existing register defects found while checking, neither caused by this PR

1. **Run 3's results are not recorded.** The run that produced #2017, #2023 and #2026 (2026-07-31/08-01) has no entry — the register's newest Section E paragraph is Run 2. The run-3 session was paused by the owner to fix these bugs first, so this bookkeeping is owed from that run, not from this PR.
2. **The file contradicts itself on Section E.** Line 105 says *"Section E remains entirely unrun"*; line 214 says *"Section E — 4 of 9 now run (2026-07-31)"*. Both cannot be true. Note `npm run check:onbox-register` cannot catch this — it validates counts against headings, not prose against prose.

Both fold into the single consolidated register update described above, rather than being fixed from this branch.

Issue #1950, opened 2026-07-29
ops — verify.yml never runs the sidecar pytest suite, so sidecar-only PRs get zero CI coverage


dudarenok-maker:
﻿## What

`.github/workflows/verify.yml` has **no sidecar (pytest) job**. Its jobs are `detect`, `lint-and-checks`, `frontend-tests`, `server-tests`, `e2e`, `e2e-visual`, `build`, `verify`. The `sidecar` scope flag computed at `:97`/`:153` is only ever consumed to gate the **server TypeScript** tests (`:289`, `:299`), and the comment there acknowledges the consequence:

```
# config loads. A sidecar-only PR touches no server TS graph → 0 tests
# run here, which is correct (server tests mock the Python sidecar).
```

So `npm run test:sidecar` — the whole `server/tts-sidecar/tests/` pytest suite — **never runs in cloud CI**. A PR that changes only `server/tts-sidecar/**` gets every leg scope-skipped and reports green having executed no Python test at all.

## Why it matters

This is the exact blind spot that let **#1941** ship: a Critical defect where every freshly cloned Qwen voice returned HTTP 500 on its first synthesis until the sidecar restarted. It was found only by on-box acceptance (register row A1), and CLAUDE.md's own note on that discovery — "unit tests mock the engine, and no sidecar pytest exercised clone→synth in one process against the real cache" — describes a coverage gap that CI could not have closed regardless, because CI does not run the sidecar suite.

Concretely observed on **PR #1942** (the fix for #1941): it changes only `server/tts-sidecar/main.py` and `server/tts-sidecar/tests/test_qwen_clone_voice.py`, and its full check list contains no sidecar leg. Every check passed in 13–38s. **The regression test shipped in that PR was never executed by CI** — its only verification is a local run on the author's box.

## Complication to design around

`npm run test:sidecar` requires the sidecar venv at `server/tts-sidecar/.venv` and emits a SKIP banner + exit 0 when it is absent (fresh clone). A naive CI job would therefore "pass" by skipping, which is worse than no job — it would look like coverage. The job needs to either bootstrap the venv (install `requirements/base.txt`) or fail loudly when the venv is missing rather than skipping.

[...]

Issue #2119, opened 2026-08-05
ops-19 — verify.yml's scope map routes launch.mjs, install-qwen3.mjs and pinokio.js to no leg that tests them


dudarenok-maker:
**Lens:** Maintenance · CI scope map · sibling of #2115

## What

#2115 fixed the **local** half of the stale-green trap (`npm run verify` printing
`test:hooks [cached]` for a producer-only edit). The **cloud** half is still open, and for a
different reason: `.github/workflows/verify.yml`'s scope detector routes by path regex, and
three files a hooks test directly depends on match **no scope at all**.

`test:hooks` runs in cloud iff `hooks || scripts || shared` (`verify.yml:227`). Simulated
against the real matchers at `:155` and `:164`:

```
scripts/deps-watch.mjs                        -> test:hooks RUNS
launch.mjs                                    -> test:hooks SKIPS
server/tts-sidecar/scripts/install-qwen3.mjs  -> test:hooks SKIPS
pinokio.js                                    -> test:hooks SKIPS
eslint.config.mjs                             -> test:hooks SKIPS
```

**Failure scenario:** a PR whose only change is `launch.mjs` — say the port-probe logic
`scripts/tests/launch.test.mjs` exists to pin. `launch.mjs` matches no matcher whatsoever,
not even `frontend`, so every scope is false and the "Hooks tests" step is skipped on the
**required** status check. The regression reaches `main` green. `install-qwen3.mjs` is the
same shape via a different route: it sets `sidecar=true`, which runs **pytest**, not
`test:hooks`. `pinokio.js` sets `pinokio=true`, which runs `test:pinokio` — a *different
suite* from the `pinokio-entry.test.mjs` that actually loads it.

PR #2117 declares all of these as **local** cache inputs, so the developer loop is fixed.
This issue is only about the cloud gate, which is the enforcing one.

## Why this wasn't folded into #2117

The one-line fix — extend `:164`'s `hooks` regex — is obvious, but there is **no test
harness for those workflow regexes at all**. They are bash `match` calls inside a YAML
`run:` block; nothing in the repo asserts their behaviour, so a fix would ship on inspection
alone. Per CLAUDE.md's fix-now bar ("coverable by a test in this same PR"), building that
harness is the actual work, and it is design work: the scope map exists in two places now
(`verify.yml`'s regexes and `verify-cache.mjs`'s `STEPS[].inputs`) with no mechanism keeping
them agreed.

## Acceptance

[...]

PR #2146, opened 2026-08-05
fix(ops): close the .github scope hole and add a sidecar leg that can fail

dudarenok-maker:
## Summary

**PR A1 of three** closing #2119 — purely additive. Nothing here changes how scope is *computed*; both halves are additions to *what runs*, so each is independently revertable. Landing it first de-circularises A2: with `.github/**` in scope, A2's wiring assertions actually run on A2's own PR.

**Defect D — a workflow-only or hook-only PR ran ZERO test legs.** Simulating all nine of `verify.yml`'s scope matchers against the literal path `.github/workflows/verify.yml` yielded **no matches**, and `.github` was an input to no verify step — so a PR editing nothing but the workflow ran nothing, in cloud *and* locally, printing `[cached]` while doing it. `.husky/**` was one level worse: covered only by the workflow's own `hooks` bash matcher and an input to no step at all, so it was one deletion away from the same fate — and A2 deletes exactly that matcher. Both are now real `test:hooks` inputs, in the cache map and the matcher alike.

**Defect C — `verify.yml` had no sidecar pytest leg at all.** The Python half of this repo was ungated in cloud CI. It has one now, and one that can genuinely fail: a new cross-platform `scripts/run-sidecar-tests.mjs` (replacing a PowerShell-only runner hardcoding `.venv\Scripts\python.exe`) with `--require-venv`, so an absent or half-bootstrapped venv is **red in CI** while the fresh-clone SKIP-and-exit-0 contract `npm run test:sidecar` has always had is preserved locally. The job is wired into the aggregator's `needs:` as well as into the workflow — a job outside that list can run, fail, and block nothing.

### What measurement changed, versus the plan

**The design undercounted `importorskip` by 10×.** It enumerated *module-scope* imports — which is what breaks *collection*. Twenty more were **function-scope**, which break the *run*: collection went green at exit 0 while `pytest -q` still exited 1. Each of the 20 is guarded on its own true proximate dependency (one is `qwen_tts`, not torch) and skips at the individual test, never the file. Proven both directions, because a skip that always skips is worse than the red it replaces — absent → `963 passed, 44 skipped`, exit 0; present, under a real torch venv → those same 20 **run and pass**, 0 skipped.

- **Five `test_qwen3.py` tests** passed here and failed on a clean runner — `_ensure_base17_for_mint` checks the real HF weights cache before reaching any test-stubbed seam, so they silently depended on this box having Qwen 1.7B-Base installed. Made hermetic rather than skipped, with a RED/GREEN reproduction that points `HF_HUB_CACHE` at an empty directory. **The new leg caught this on its first cloud run** — which is the opposite of the silent-green failure this branch exists to eliminate.
- **`.github/actions/**` was defect D's other half** — the composite setup action is consumed by all seven jobs and matched no scope, so a PR pinning a different Node or changing the `npm ci` cache key still ran zero legs. Fixed the same way `workflows/` was. Whether it should set `shared` rather than `hooks` is a routing decision, deliberately left to #2152 rather than made mid-PR.
- **`sidecarFingerprint` hardcoded `.venv/Scripts/python.exe`** — on POSIX it returned `'unavailable'` permanently, so a Linux contributor could bootstrap the venv and have the sidecar step print `[cached]` forever, never running the suite this PR just made cross-platform. Now shares `resolveVenvPython`'s platform branch, with a test that forces `platform: 'linux'` and builds only a `bin/python` layout — a test pinning the Windows branch alone would have left the bug exactly where it was.
- **`requirements/base.txt` and `pytest.ini` were read at runtime but undeclared** as `test:sidecar` inputs — `requirements*.txt` compiles to `requirements[^/]*\.txt$`, which cannot cross the `/`. `base.txt` became load-bearing in this PR, since the CI bootstrap installs from it directly.
- **Two comments asserted things that are not true** — one cited a test file that does not exist until the next PR, the other a runtime read that does not happen. Both were authored in the plan and copied verbatim, so per-task review read them as spec rather than as claims. Reworded; the declarations stand, since over-declaring an input only costs re-runs.

## Test plan

- `node --test scripts/tests/verify-cache.test.mjs` — 3 new scope tests (`.github/workflows/**` ×2, `.husky/**`)
- `node --test scripts/tests/run-sidecar-tests.test.mjs` — 11 tests: 4 venv-layout, 7 spawning the real CLI end to end for the exit-code contract
- `pytest --collect-only` under a lean `base.txt + requirements-dev.txt` venv: exit **0** (was exit 2, two collection errors); full run exit 0
- Cloud: the new **Sidecar tests (pytest)** leg, and `npm run verify` gating on it

**Mutation-verified, not merely green:**

| Mutation | Result |
|---|---|
| drop the `--require-venv` check | RED |
| replace `result.status ?? 1` with `0` | RED |
| revert `'.github/workflows/**'` / `'.husky/**'` globs | RED (per test) |
| point `HF_HUB_CACHE` at an empty dir, pre-fix | RED (the 5 qwen3 tests) |

Independent review also caught that the exit-code fixture's interpreter discovery **threw at module scope** when no relocatable Python was on PATH, taking all 11 tests down — including 4 needing no Python at all — which would have failed `test:hooks` opaquely on the `windows-latest`/`macos-latest` legs of `cross-os.yml` and every release-tag cut. It now degrades to skipping exactly the 7 CLI tests, naming each candidate tried and why it was rejected.

**Not applicable, stated rather than skipped silently:** no `docs/features/` regression plan (the design of record and the 15-task plan landed in #2142; this is CI/ops, not a feature surface); no `docs/features/INDEX.md` change; no on-box acceptance owed (nothing here needs a GPU, a real analyzer, or a real book); no `RELEASE_NOTES.md` line — that file is the user-facing register for an audiobook app's end users and a CI gating change has no user-visible delta, so the entry is technical-register only.

Refs #2119

Design: `docs/superpowers/specs/2026-08-05-verify-scope-map-unification-design.md`
Plan: `docs/superpowers/plans/2026-08-05-verify-scope-map-unification.md`

https://claude.ai/code/session_011c1MfFziLGzwUDU4dX9SJe


dudarenok-maker:
This PR discharges **#1950** in full, but its body links only `Refs #2119` — so #1950 will stay open after merge unless the linkage is added. Flagging rather than editing, since it's your PR.

#1950's acceptance, against what's here:

| #1950 asks | This PR |
|---|---|
| a sidecar-scoped job runs `npm run test:sidecar` on any PR touching `server/tts-sidecar/**` | the `sidecar-tests` job, gated on `needs.detect.outputs.sidecar` |
| it **fails**, not skips, if the venv can't be bootstrapped | `--require-venv` turns the SKIP into a hard failure, and the bootstrap step precedes it |
| verified by breaking a sidecar test and confirming CI goes red | the job is in the aggregator's `needs:`, so it gates rather than merely running |

The third row is the one that usually goes wrong, and it's handled — a job outside `needs:` can fail and block nothing.

Worth noting the finding in your release note is stronger than #1950 knew: #1950 says `verify.yml` has no sidecar leg, which is true, but it doesn't mention that `release.yml` and `cross-os.yml` *do* invoke `test:sidecar` via `npm run verify`/`verify:quick` and always hit the SKIP path on a fresh runner. So two workflows already appeared to run the Python suite and reported green having executed nothing. `--require-venv` is what closes that, not just the new job.

Suggested: add `Closes #1950` to the body. If you'd rather keep this PR scoped to #2119, say so and I'll close #1950 as delivered-by-#2146 once this merges — either way it shouldn't linger.

Context: this came out of a Round 3 design triage that reached #1950 and stopped on finding your plan (PR #2142) already covered it as defect C. Cross-reference is on the issue.

Issue #2187, opened 2026-08-06
srv-59 follow-up: dialogue-structure aligner under-aligns on Russian dash dialogue, stranding chapters below the 80% floor


dudarenok-maker:
## What

On-box acceptance of plan 247 (srv-59) against *Ночной дозор* — 9 chapters, 15,069 sentences, `qwen36-cw-iq4-32k`, local engine, structure engine on, `escalation='local'` — shows the deterministic dialogue-structure engine is **sound**, but `locateSentenceOffsets` (`server/src/analyzer/dialogue-structure/aligner.ts`) under-aligns badly on this restructure-heavy Russian dash-dialogue book. Chapters that fall below the **hardcoded 80% alignment floor** (`alignmentFloorPct: 80`, see `server/src/analyzer/dialogue-structure/evidence.ts:40`) have correction disabled chapter-wide and degrade to flag-only, which is what blows the flagged count past plan 247's target.

### Measured (2026-08-06)

| Chapter | aligned | outcome | flagged |
|---|---|---|---|
| 5 | 4% | below floor — escalation skipped | 1,732 |
| 6 | 2% | below floor — escalation skipped | 1,679 |
| 7 | 66% | below floor — escalation skipped | 1,447 |
| 8 | 73% | below floor — escalation skipped | 1,222 |
| 9 | **95%** | **engine ran** — confirmed=580 corrected=531 escalated=22 accepted=130 | **488** |

**Where the engine actually ran it hit the target**: chapter 9 landed flagged=488, under plan 247's ≤~500 bar. Aggregate (chapters 5–9) was flagged 6,568, alignedPct 47.4%.

### Root cause signal

Short, highly repeated dialogue sentences mis-locate to a later occurrence. Concrete case from ch7:

```
separator at 61,651 = "***"
true opener in body = "У каждого из нас, дозорных, работа отнимает..."
flagged sentence    = "- Да."  located at 69,710  (true position ~60,850)
following sentences "- Нет, со взрослыми." / "- Мама далеко, позвать?" → do not locate at all
```

`locateSentenceOffsets` advances a cursor and matches on first hit, so an ultra-common string like `"- Да."` binds to the wrong occurrence and drags subsequent alignment off.

## Why it matters

[...]

Issue #1685, opened 2026-07-17
Verify #1682 cloud request sizing on-box + calibrate stage-2 local input fraction


dudarenok-maker:
## Context

Follow-up to **#1682** (analyzer: size cloud requests to the model's token limits). #1682 ships the mechanism — per-request input-token cap, finite Gemma free-tier TPM (16k/min), per-minute-vs-daily 429 classification, `acquire()` fail-fast guard, and local input-fraction knobs — but two acceptance steps require a real GPU box + a free-tier Gemini key and so are deferred here rather than blocking the PR.

Design: `docs/superpowers/specs/2026-07-17-cloud-request-sizing-design.md`
Plan: `docs/superpowers/plans/2026-07-17-cloud-request-sizing.md`

## What to do

1. **Cloud acceptance (free-tier Gemma).** With a free-tier `GEMINI_API_KEY`, re-analyze *Ночной дозор* (`C:\AudiobookWorkspace\books\Сергей Лукьяненко\The Night Watch Tetralogy\Ночной дозор`) end-to-end on `gemma-4-31b-it`, including the **script-review** pass (the pass that actually 429'd — all 22 logged gemini failures were `task: script-review`). Confirm:
   - The book **completes** (throttled/slow) with **no dropped chapters** and **no hang**.
   - No request 429s as "daily quota" when it is a per-minute limit.
   - The `acquire()` fail-fast guard never fires in normal operation (it's a backstop).
2. **Calibrate `analyzer.stage2.localInputFraction`.** The local truncation is a stage-2 output overflow (`done_reason=length`, ~54–71 KB output on `qwen3.5:4b` / `gemma4-e4b-8gb` — see `logs/server.err.log:538,1220-1254`). The knob ships at its current-behavior default **0.3**. Capture a fresh local truncation trace, then lower `analyzer.stage2.localInputFraction` (Advanced Settings) until a full local re-analysis of *Ночной дозор* completes without stage-2 truncation drops. Record the working value and, if it's clearly better as a default, propose changing the shipped default in a small follow-up.

## Acceptance

- [ ] Free-tier Gemma re-analysis of *Ночной дозор* completes, no dropped chapters, no hang (screenshot/log).
- [ ] A per-minute 429 is observed being retried (not surfaced as daily-quota).
- [ ] A local (Qwen/gemma4-e4b) re-analysis completes without stage-2 truncation drops at the calibrated `stage2.localInputFraction`; value recorded here.

## Depends on

[...]

Issue #1471, opened 2026-07-09
srv-59 — dialogue-structure attribution: deterministic evidence engine, derived confidence, escalation


dudarenok-maker:
## What

Deterministic dialogue-structure engine for stage-2 speaker attribution: a pure-code structure parser (per-language convention tables — en/ru/es/fr/de), a cross-examiner that corrects tag-proven misattributions and **derives all sentence confidence in code** (the model's self-reported confidence is discarded), a default-on local escalation pass that re-queries only unresolved conversation windows, and structure-evidence annotations in the script-review inbox. Server-only; no frontend changes (the existing `< 0.75` low-confidence UI starts working once values are honest).

## Why

Measured on _Ночной дозор_ (ru, 14,065 sentences, analyzed 2026-07-06, local gemma4-e4b):

- **Zero low-confidence flags** despite heaps of real errors — every sentence self-reported ≥ 0.8 (0.8×1708, 0.9×6189, 1.0×6168). The stage-2 skill's calibration rubric is ignored wholesale by small local models.
- **Dialogue is under-attributed** (real speech on `narrator` or unknown-buckets), which deflates line counts → real characters get folded → more lines land on buckets. ~150 sentences were manually reassigned in one session and the user gave up; manual correction does not scale to a full book.
- Pilot measurement (spec §5.4): 2,861 speech turns; deterministic tag/pronoun anchoring resolves 32%; ~1,900 unanchored → escalation is load-bearing, not optional.

Improves **all** supported languages via convention tables; Russian is the acceptance case because it is worst today.

## Spec and plan (read in this order)

1. Spec (authority): `docs/superpowers/specs/2026-07-09-dialogue-structure-attribution-design.md`
2. Plan (14 TDD tasks): `docs/superpowers/plans/2026-07-09-dialogue-structure-attribution.md`

Both survived an adversarial assumption-checker pass (Opus); findings are folded, including the measured pilot numbers, the interior-dash disambiguation rules, and the escalation analyzer primitive (Task 9 — escalation CANNOT reuse `runStage2Chapter`, analyzer calls are schema-constrained decoding).

## Instructions for the implementation agent

- Execute the plan with `superpowers:subagent-driven-development` (or `superpowers:executing-plans`), task-by-task, in an **isolated worktree** off latest `origin/main`; branch `feat/server-dialogue-structure-attribution`. Fresh `git commit` per task — never `--amend`.
[...]

PR #2188, opened 2026-08-06
docs(docs): record Ночной дозор on-box acceptance — C1 pass, C2 targets missed

dudarenok-maker:
## Summary

Records the **2026-08-06 local *Ночной дозор* acceptance pass** across all three register surfaces, discharging register rows **C1** (plan 261 — scene separators) and **C2** (plan 247 — srv-59 attribution).

The run: 9 chapters, **15,069 sentences**, `qwen36-cw-iq4-32k` via local Ollama, structure engine on, `analyzer.structure.escalation = 'local'`, no mock mode, on the dual-GPU box.

### C1 — PASS

24 separators found, 22 flagged (2 dropped by the leading-separator rule, by design). **Median separator→opener distance is 5 characters** — the width of the `* * *` glyph itself — in 20 of 21 measurable cases. The old **~92k-character forward-overshoot is confirmed mechanically gone**: binding is index-adjacent (`lastBefore + 1`), so it cannot forward-search to a distant opener regardless of alignment quality. Chapters 1–4 contain no separators at all, so their zero flags are correct rather than a miss.

### C2 — targets missed, root cause identified

Book-level flagged is **6,568** against the ≤~500 target. Per-chapter:

| Chapter | aligned | outcome | flagged |
|---|---|---|---|
| 5 | 4% | below floor — flag-only | 1,732 |
| 6 | 2% | below floor — flag-only | 1,679 |
| 7 | 66% | below floor — flag-only | 1,447 |
| 8 | 73% | below floor — flag-only | 1,222 |
| 9 | **95%** | **engine ran** | **488** |

Only chapter 9 cleared the hardcoded **80% alignment floor**; the rest hit the §5.2 chapter-wide flag-only fallback exactly as designed. **Where alignment succeeded, the engine met the target** (488 ≤ ~500, with `confirmed=580 corrected=531 escalated=22 accepted=130`).

So the deterministic engine is validated and **`locateSentenceOffsets` is the bottleneck** — filed as #2187. Both C1's single cosmetic blemish and C2's shortfall trace to the same defect: short, highly-repeated dialogue sentences bind to a later occurrence (ch7's `"- Да."` resolved to offset 69,710 instead of ~60,850).

Target 3 **was** met: roster folded 58 → 35, unknown-bucket down to 3 entries / 61 lines (0.4%).

### Caveat, recorded on every surface

The provenance report covers **chapters 5–9 only** — chapters 1–4 were served from the resume cache after two environmental restarts (an unrelated concurrent session's `stop-app` port sweep, and a `tsx watch` restart). Book `flagged`, `corrected` and `confirmed` sum *exactly* to the ch5–9 components, and `alignedPct` 47.4% is their weighted average. Deliberately **not** re-run: the diagnostic is conclusive without the missing four, and #2187 is the deliverable either way.

### Register mechanics

Group C's surviving cloud row **renumbers C3 → C1** so row IDs stay contiguous from 1, which `check:onbox-register` enforces. Owed total **57 → 55**, Group C **3 → 1**. The register, the hand-authored live-view HTML, both plans and `INDEX.md` all move in this PR; `npm run check:onbox-register` is green.

Two facts the run established for the remaining cloud row, recorded so it isn't rediscovered: `gemini-*` genuinely RECITATION-blocks this book (observed — a fallback call returned `PROHIBITED_CONTENT`), and `server/.env`'s `GEMINI_MODEL=gemini-3.5-flash-lite` overrides the RECITATION-safe `gemma-4-31b-it` code default. It also notes the cloud pass must run against a throwaway re-import, since the analysis cache is keyed by `manuscriptId` only and would otherwise overwrite the qwen36 result the owner is keeping for cast + generation.

## Test plan

- `npm run check:onbox-register` — **green** (register ↔ live view agree: owed total, per-group counts, row IDs)
- Docs-only change; no runtime surface, so no paired automated test is owed
- No release-notes entry: pure docs/process, no user- or operator-visible delta
- **Live view publish is NOT done in this PR** — per the register's own four-step procedure, publishing happens after merge, against the canonical artifact URL, with `check:onbox-register -- --against-published` run immediately beforehand

Refs #2187
Refs #1471
Refs #1685


dudarenok-maker:
Closing and reopening as a fresh PR — this PR object stopped producing workflow runs entirely (no runs for `opened`, `edited`, `reopened`, or a force-push `synchronize`), so its required checks could never report. Same branch, same commit.

PR #2191, opened 2026-08-06
fix(server): make the UNREADABLE tests self-verifying and render every sentinel safely

### LINKED ISSUE #2186: Harden the cast merge-base UNREADABLE tests and teach describeFingerprintForLog the new sentinel
## What

Two residuals from PR #2185's final review. Neither is a defect today; both are cheap and both close a way for a guard to go quiet.

### 1. The two `UNREADABLE` tests carry no proof the injection is live

`server/src/workspace/cast-merge-base.test.ts` injects a non-`ENOENT` read failure via `vi.mock('node:fs/promises')` to prove that a transient `EBUSY`/`EPERM` on `cast.json` does **not** fire a false `cast_merge_base_stale` advisory.

If that mock ever stops applying — say `cast-fingerprint.ts` switches from the `node:fs/promises` specifier to `fs/promises` — **both tests still pass**, because site 1 would then read the real file, whose hash already equals the baseline, and report no conflict either way. The assertion is satisfied for the wrong reason and the cascade regression lock silently evaporates.

It is genuinely live today: reverting `readJsonWithFingerprint`'s catch to its bare form reddens both tests, which is only possible if the injected `EBUSY` reaches the catch.

**Fix (two lines):** make the injected implementation a `vi.fn()` and assert it was called, or seed a foreign write before site 1 so a *successful* read would conflict — making a silently-disarmed mock fail rather than pass. `cast-fingerprint.test.ts` already self-verifies this way; `cast-merge-base.test.ts` does not.

### 2. `describeFingerprintForLog` doesn't know about `UNREADABLE`

`server/src/workspace/cast-fingerprint.ts` — the helper maps `ABSENT` to the literal text `ABSENT` precisely so a raw NUL never reaches a log line (some consumers truncate at NUL). `UNREADABLE` is also NUL-prefixed and is **not** handled, so it would emit the same hazard the helper exists to prevent.

Unreachable today: the comparison is suppressed before `onConflict` when a read is unreadable, so the sentinel never reaches a log call. That makes it a latent edge, not a defect — but the next person to log a fingerprint from a different site inherits the trap.

**Fix:** one branch, plus the assertion in `cast-fingerprint.test.ts` that no rendered form contains a raw `0x00` for **any** sentinel — written so a future third sentinel fails the test rather than slipping through.

## Acceptance

- Disarming the `node:fs/promises` mock makes the two `UNREADABLE` tests **fail** rather than pass.
- `describeFingerprintForLog(UNREADABLE)` returns readable text with no raw NUL, and the test is written against the set of sentinels rather than naming two literals.

## Key files

- `server/src/workspace/cast-merge-base.test.ts`
- `server/src/workspace/cast-fingerprint.ts`, `cast-fingerprint.test.ts`

## Benefit

*Benefit (technical):* the regression lock protecting against a false "someone edited your cast" advisory cannot go quiet without a test failing, and the NUL-in-logs hazard is closed by construction rather than by the current call graph happening not to reach it.

Found by the whole-branch review of PR #2185; both explicitly judged non-blocking there.

dudarenok-maker:
## Summary

Two residuals from PR #2185's final review. Neither was a defect; both were ways a guard could go quiet without anyone noticing.

**1. The `UNREADABLE` tests carried no proof their injection was live.**

`cast-merge-base.test.ts` proves that a transient `EBUSY`/`EPERM` on `cast.json` does *not* fire a false `cast_merge_base_stale` advisory, by injecting a non-`ENOENT` read failure through `vi.mock('node:fs/promises')`.

If that mock ever silently stopped applying — say `cast-fingerprint.ts` switched to the bare `fs/promises` specifier — **both tests would still have passed**: site 1 would read the real file, whose hash already equals the baseline, so "no conflict" either way. The assertion would be satisfied for the wrong reason and the regression lock against a false *"someone edited your cast"* advisory would evaporate in silence.

Each test now seeds a **foreign write before the mock is installed**, so a successful real read would observe a mismatch and fire `onConflict`. A disarmed mock now fails the test instead of passing it.

**2. `describeFingerprintForLog` didn't know about `UNREADABLE`.**

That helper exists so a raw `0x00` never reaches a log line — some consumers truncate at NUL. It mapped `ABSENT` to safe text but not `UNREADABLE`, which is NUL-prefixed too. Unreachable today (the comparison is suppressed before any `onConflict` log call), so a latent edge rather than a bug — but the next person to log a fingerprint from a different site inherits the trap.

Sentinels now live in an exported `FINGERPRINT_SENTINELS` array with a label table, and the test **iterates that set** rather than naming `ABSENT` and `UNREADABLE` by hand. A test that names two literals is precisely the failure mode this ticket exists to close: it would pass unchanged on the day someone adds sentinel three.

## Test plan

`server/src/workspace/cast-fingerprint.test.ts` + `cast-merge-base.test.ts` — 24 tests, green at `--retry=0`.

**Both hardenings were watched to fail, which is the entire point of the ticket:**

| Neutralisation | Result |
|---|---|
| Point `vi.mock` at a specifier that cannot match (`node:fs/promises__DISARMED`) | **Both** `UNREADABLE` tests go red — *"does not fire onConflict, and the baseline still advances"* and *"does NOT cause a phantom conflict at the NEXT site"*. Pre-#2186 they passed regardless. |
| Add a third NUL-prefixed sentinel to `FINGERPRINT_SENTINELS` without a label | The set-based test goes red — and the raw NUL leaks into the runner's own output, which is the hazard made visible. |

Both restored to green afterwards.

## Notes

- **Release notes: skipped deliberately.** Internal test hardening plus a currently-unreachable log branch — no user- or operator-visible delta (CLAUDE.md step 5 carve-out).
- **Regression plan: none.** Small and localized; the issue body plus the paired tests are the spec.
- **On-box acceptance: none owed.**
- One residual limit worth naming rather than leaving implied: the set-based test catches a new sentinel that is *added to `FINGERPRINT_SENTINELS`* without a label. It cannot catch a sentinel declared and never added to that array. The array is the contract; that is the seam to keep honest.

Closes #2186

### DIFF
--- a/server/src/workspace/cast-fingerprint.test.ts
+++ b/server/src/workspace/cast-fingerprint.test.ts
@@ -33,8 +33,15 @@ function setReadFileImpl(fn: ReadFileUtf8 | null): void {
 
 /* Import AFTER vi.mock so cast-fingerprint.ts picks up the mocked readFile. */
 const { readFile } = await import('node:fs/promises');
-const { ABSENT, UNREADABLE, hashBytes, readJsonWithFingerprint, fingerprintOfWrite } =
-  await import('./cast-fingerprint.js');
+const {
+  ABSENT,
+  UNREADABLE,
+  FINGERPRINT_SENTINELS,
+  hashBytes,
+  readJsonWithFingerprint,
+  fingerprintOfWrite,
+  describeFingerprintForLog,
+} = await import('./cast-fingerprint.js');
 
 function tmpDir(): string {
   return mkdtempSync(join(tmpdir(), 'castwright-fingerprint-'));
@@ -167,5 +174,27 @@ describe('fingerprintOfWrite — coupling guard against writeJsonAtomic', () => 
     expect(UNREADABLE).not.toMatch(/^[0-9a-f]{64}$/);
     expect(UNREADABLE.startsWith(String.fromCharCode(0))).toBe(true);
     expect(UNREADABLE).not.toBe(ABSENT);
+  });
+});
+
+describe('describeFingerprintForLog - every sentinel renders safe text (#2186)', () => {
+  it('renders every member of FINGERPRINT_SENTINELS as non-empty text with no NUL byte', () => {
+    /* Iterates the actual sentinel SET, not two hardcoded literals: a future
+       third sentinel added to FINGERPRINT_SENTINELS without a matching label
+       in describeFingerprintForLog falls through to the raw-hex-prefix
+       branch, which still carries the leading NUL and fails the assertions
+       below. That is the whole point: this test breaks the day someone adds
+       sentinel three and forgets to teach the renderer about it. */
+    const nul = String.fromCharCode(0);
+    for (const sentinel of FINGERPRINT_SENTINELS) {
+      const rendered = describeFingerprintForLog(sentinel);
+      expect(rendered).not.toContain(nul);
+      expect(rendered.length).toBeGreaterThan(0);
+    }
+  });
+
+  it('still renders a real sha256 digest as a 12-char prefix (unaffected by the sentinel table)', () => {
+    const digest = hashBytes('{"characters":[]}');
+    expect(describeFingerprintForLog(digest)).toBe(digest.slice(0, 12));
   });
 });
--- a/server/src/workspace/cast-fingerprint.test.ts
+++ b/server/src/workspace/cast-fingerprint.test.ts
@@ -184,7 +184,9 @@ describe('describeFingerprintForLog - every sentinel renders safe text (#2186)',
        in describeFingerprintForLog falls through to the raw-hex-prefix
        branch, which still carries the leading NUL and fails the assertions
        below. That is the whole point: this test breaks the day someone adds
-       sentinel three and forgets to teach the renderer about it. */
+       a sentinel to FINGERPRINT_SENTINELS and forgets to teach the renderer
+       about it — a sentinel declared but never added to that array is still
+       not caught. */
     const nul = String.fromCharCode(0);
     for (const sentinel of FINGERPRINT_SENTINELS) {
       const rendered = describeFingerprintForLog(sentinel);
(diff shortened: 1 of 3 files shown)


dudarenok-maker:
## Review findings folded in

The independent review returned APPROVE WITH NITS and found that the new label table **reintroduced the same class of latent trap this ticket exists to close**. Fixed in `80bcda8b`.

`SENTINEL_LOG_LABELS` was a plain object literal typed `Partial<Record<string, string>>`, so the lookup resolved through `Object.prototype`. Verified: `describeFingerprintForLog('constructor')` returned a **function**, `'toString'` a function, `'__proto__'` an object — despite the declared `: string` return type, which TypeScript cannot catch because that type makes the read `string | undefined`.

Unreachable today (real inputs are sha256 hex or a NUL-prefixed sentinel, and `[0-9a-f]{64}` cannot equal an `Object.prototype` key). But #2186's own bar is *"closed by construction rather than by the current call graph happening not to reach it"* — so leaving it would have contradicted the ticket's thesis in the ticket's own PR.

Two properties now hold, and I proved each rather than assuming it:

| Neutralisation | Result |
|---|---|
| Revert the lookup to `(SENTINEL_LOG_LABELS as Record<string,string>)[value] ?? …` | The new prototype test goes **red** (1 failed / 11 passed) |
| Delete the `[UNREADABLE]` label | `npm run typecheck` fails with **`error TS2741: Property '"\0UNREADABLE"' is missing … but required in type 'Record<"\0ABSENT" \| "\0UNREADABLE", string>'`** |

That second one is the upgrade worth naming: a missing label was previously only a *runtime test* failure, and is now a **build** failure — the stronger form of exactly the guarantee this PR sells.

Also folded in: a comment that claimed a broader guarantee than the test delivers. It said the test *"breaks the day someone adds sentinel three"*; it actually breaks only when a sentinel is added **to `FINGERPRINT_SENTINELS`**. A sentinel declared and never added to that array is still not caught. The comment now says so, and the array is named as the seam to keep honest.

One reviewer nit deliberately **not** taken: `expect(rendered.length).toBeGreaterThan(0)` in the set-based test is close to inert (only an empty-string label reddens it). It is cheap and documents intent alongside the NUL assertion that does the real work, so it stays.

Issue #2129, opened 2026-08-05
fe — the Cast banner and the repair pass now give opposite answers for an alias-resolved id


dudarenok-maker:
Found by the pre-merge review of the #2107 fix, 2026-08-05.

## What

After #2107, the repair script and the Cast screen give an operator **opposite** answers about the same character id, and nothing on either surface says so.

- `server/src/audio/segments-io.ts:362-365` tags a `'history'` or `'normalised-history'` resolution as `'alias'`, which the Cast screen renders under **auto-reconciled** — i.e. "nothing to do here".
- `scripts/repair-cast-id-drift.mjs` now calls those same ids **damage requiring a re-render** (and, after the #2107 widening, `'normalised-id'` too).

## Why it matters

The two surfaces genuinely answer different questions — the banner answers "does this id resolve today?", the script answers "were these bytes rendered with the wrong voice?" — and #2107's whole point is that those are not the same question. So the divergence may well be correct.

What makes this a defect rather than a design choice is that **the codebase has twice ruled the other way and still says so in-tree**. #2093 residual 5 and round-2 MINOR-4 were both *filed and fixed* on the principle "don't contradict the Cast banner", and `planBookRepairs` (~`:656-665`) still carries that reasoning verbatim as the justification for the branch immediately below it.

An operator reading the Cast screen sees "auto-reconciled, nothing to do" for `the-torment` while the repair pass lists 67 of its segments as needing a re-render.

## Acceptance

- Either the banner distinguishes "resolves today" from "audio is current" for these ids, or the register/run sheet states plainly that the two surfaces answer different questions and which one governs a re-render decision.
- The stale justification comment in `planBookRepairs` is corrected either way — it currently argues for a position the code no longer holds.

## Key files

- `server/src/audio/segments-io.ts:362-365`
- `scripts/repair-cast-id-drift.mjs` (`planBookRepairs`, the `reportOnly` branch)
- `src/views/cast.tsx` (the auto-reconciled/needs-your-decision split)

## Benefit

*(user)* Two screens stop disagreeing about whether a character needs attention.

Issue #2107, opened 2026-08-05
ops — repair-cast-id-drift: a recorded alias drops its chapters off the re-render list, hiding still-damaged audio


dudarenok-maker:
Found running the **A33** on-box acceptance (`--apply`) against the real 20-book workspace, 2026-08-05.

## What

`scripts/repair-cast-id-drift.mjs`'s re-render list is documented — in `buildRerenderRows`' own doc comment and in register row A33 — as **unconditional on auto-record status**:

> Includes EVERY orphaned segment regardless of whether an alias was auto-recorded for it: recording the alias fixes downstream metadata/attribution, **not the audio bytes already on disk**, so whether to re-render is always a separate call (spec §4.7's own words) — pure, no filtering.

It isn't, across runs. Measured on the real workspace:

| | re-render rows | segments |
|---|---|---|
| before `--apply` | 17 | 120 |
| after `--apply` | 13 | **93** |

The missing 4 rows / 27 segments are exactly the ones the pass just aliased — `mayrin` ch2 (8), `coalfall` ch2 (13) in *Заказ Коалфолла*, `lady-alina` ch55 (4) + ch61 (2) in *Everblaze*. That audio is **still narrator-substituted on disk**; the alias fixed attribution only. So the pass silently drops precisely the chapters it just made *look* fixed.

## Why

`buildRerenderRows` is innocent — it genuinely doesn't filter. The loss is one level up, in `collectSegmentOrphans` (`scripts/repair-cast-id-drift.mjs:878-899`): the resolver is built **with** the on-disk history,

```js
const resolver = mods.buildCastResolver(cast.characters, {
  supersededBy: history.supersededBy ?? {}, …
});
…
const resolution = resolver.resolve(id);
…
continue; // resolves (exact/history/normalised) — not an orphan
```

so once `--apply` has written `mayrin -> mairin`, the next run resolves `mayrin` via the `'history'` tier and `continue`s it as "not an orphan" before it can ever reach the re-render list.

Within a single run this is invisible — the aliases are written from `pendingWrites` at the end, after `segmentOrphans` is computed, which is why the `--apply` run itself still printed all 17 rows. It only bites on the **next** run, which is exactly the "re-run to get my work list" flow A33's own criteria prescribe.

## Acceptance

- After `--apply`, the re-render list still names every chapter with orphan-rendered audio on disk, including ones whose id now resolves through a freshly-written alias.
[...]

Issue #2028, opened 2026-07-31
ops-46 — vitest retry:1 can turn a genuine red-phase test green


dudarenok-maker:
## What

`server/vitest.config.ts:128` sets `retry: 1` for the whole server suite. For any
test that asserts on **module-level mutable state keyed by a fixed string**, that
retry does not re-run a clean attempt — it re-runs against state the failed
attempt already mutated, and the second attempt can pass for the wrong reason.

Found while landing #2001's fix on `feat/server-1981-cast-lock`. The red-phase
test was:

```ts
const before = __chainsSizeForTest();
await withCastLock('/w/cleanup', async () => 'done');
expect(__chainsSizeForTest()).toBe(before);   // dead cleanup ⇒ before + 1
```

Attempt 1 fails and leaks its entry into `chains` under key `/w/cleanup`.
Attempt 2 reads `before` with that entry *already present*, re-acquires the
**same** key — an overwrite, not growth — and passes. Net effect: the test
reports **green against the unfixed code**. It only went red with an explicit
`--retry=0`.

Verified independently by the task reviewer: config line confirmed, mechanism
confirmed consistent with vitest's retry semantics.

## Why it matters

This is the placebo-test failure mode with a config-level cause rather than an
authoring one. A red-phase test that never goes red proves nothing, and here the
harness itself is what hides it — no amount of care in writing the test helps.
Any future test asserting on a `Map`/`Set`/counter at module scope is exposed.

## Acceptance

Pick one, deliberately — this is a real trade-off, not an obvious fix, which is
why it is filed rather than folded into #1981's PR:

- **Narrow:** `retry: 0` on the specific specs that assert on module-level state,
  via a `test.retry(0)` or per-file config, leaving the global retry for the
  genuinely-flaky lanes it exists to absorb.
- **Global:** drop `retry: 1` entirely and route real flakes through
  `quarantinedIt` (`server/src/test-utils/quarantine.ts`), which is the repo's
  documented mechanism and already has a register at
  `docs/testing/flaky-register.md`. Needs a survey of what currently depends on
  the retry first — dropping it blind could redden the suite.

Either way, add a note to CONTRIBUTING.md's testing section so the next person
writing a red-phase test against module state knows to pass `--retry=0`.

## Key files

- `server/vitest.config.ts:128`
- `server/src/workspace/file-lock.ts` (`__chainsSizeForTest`, the test that hit it)
[...]

Issue #2057, opened 2026-08-01
Reconcile onbox-acceptance-register.md's owed #2026 row (and republish the HTML twin) once #2039 merges


dudarenok-maker:
PR #2050 (fixing #2026) deliberately did **not** add a register row to `docs/testing/onbox-acceptance-register.md` for its two owed on-box acceptance items — (1) by-ear confirmation of the leading-dash pause, (2) live confirmation the Coqui degeneracy guard neither false-positives on ordinary short Russian lines nor (if reproducible) fails to recover a real degenerate collapse — because PR #2039 was concurrently editing that same file's E-04 row, and editing it too risked a silent auto-merge collision on the same row (see `docs/testing/fs38-wave3-onbox-acceptance.md`'s `#2026` section, which is the complete interim record).

**Once #2039 merges**, this issue tracks:

1. Add the register row for #2026's two owed items, grouped by hardware prerequisite per the register's own convention, pointing at the criteria already recorded in `docs/testing/fs38-wave3-onbox-acceptance.md`.
2. Republish the live HTML register twin at its existing artifact URL (recorded in the register's own header) — never re-published from scratch — and let its derived figures (owed count, group counts, oldest debt) recompute.

Labelled `type:chore` / `area:ops` (bookkeeping, not a code defect) and `tracking` (depends on #2039 merging first, no direct fix of its own until then).

Depends on: #2039, #2026, #2050

dudarenok-maker:
Adding a second, separate defect to fold into this same reconciliation pass — it was found while checking the register during the #2017/#2023/#2026 batch and is currently recorded nowhere but a PR comment.

**`docs/testing/onbox-acceptance-register.md` contradicts itself about Section E.**

- Line ~105: *"Section E remains entirely unrun, see the blockers below."*
- Line ~214: *"Section E — 4 of 9 now run (2026-07-31); E-03…E-07 still owed."*

Both cannot be true. Line 105 is the stale one — it predates the run-2 results recorded further down.

**Why this needs a human, not a checker:** `npm run check:onbox-register` cannot catch it. That guard validates the file's internal *arithmetic* — glance-table counts against body row headings, and the stated total against the glance table. It does not compare prose against prose, so a paragraph that flatly contradicts another one passes green forever.

[...]

PR #2244, opened 2026-08-11
feat(server,frontend,scripts): clear the repair pass's re-render list once audio is current

### LINKED ISSUE #2128: ops — the repair pass's re-render list never clears: a re-rendered chapter reappears on every later run
> [!IMPORTANT]
> **The approach described below is stale. Do not implement it. A reviewed design
> of record already exists.**
>
> - **Spec:** [`docs/superpowers/specs/2026-08-06-cast-identity-followups-design.md`](https://github.com/dudarenok-maker/Castwright/blob/main/docs/superpowers/specs/2026-08-06-cast-identity-followups-design.md) — 700+ lines, three adversarial review passes.
> - **Plan:** [`docs/features/280-cast-identity-followups.md`](https://github.com/dudarenok-maker/Castwright/blob/main/docs/features/280-cast-identity-followups.md), **Tasks 2–9** — written for this issue, never started.
>
> This body (filed 2026-08-05, unedited since) proposes a **wall-clock timestamp**
> scheme: a per-entry `recordedAt` on `cast-id-history.json` compared against
> `SegmentsFile.synthesizedAt`. The spec considered that and **moved away from it
> one day later**, in favour of a per-book monotonic **`seq` counter** plus one
> shared predicate, `isAudioCurrent(resolution, segmentsFile, history)`, called by
> *both* the Cast banner and `repair-cast-id-drift.mjs`. See the spec's
> "Why a counter and not a timestamp" section.
>
> Three things the timestamp scheme in this body gets wrong, all settled in the spec:
>
> 1. **`synthesizedAt` is unusable as the render-side anchor.** `chapter-splice.ts`
>    and `chapter-qa-repair.ts` rewrite the whole segments file and refresh the
>    stamp while re-rendering only part of the chapter. The spec has the partial
>    writers carry the prior `castHistorySeq` forward verbatim instead.
> 2. **A wall clock cannot see an alias recorded mid-render.** Both sides of the
>    `seq` comparison come from the same file's own counter, so there is no clock
>    and no skew.
> 3. **"Keep the original time through a repoint" is wrong.** `cast-merge.ts`
>    retires `sourceId` into `targetId` — same *person*, different *cast row*, and
>    whichever row won owns the voice. A render made under the old target is
>    genuinely stale, so the repoint must **restamp**.
>
> The `'normalised-id'` tier has no history entry and so cannot be cleared by any
> marker on `cast-id-history.json`; the spec's decision table routes it to
> `'unknown'` (= listed as damage) until a render stamps `castHistorySeq`.
>
> **Banner added 2026-08-06** after a design thread re-derived the stale timestamp
> scheme from this body without finding the spec — the exact outcome plan 280
> warns about. The warning lives in the body rather than a comment because
> `gh issue view --json body` does not return comments.

---

Found by the pre-merge review of the #2107 fix, 2026-08-05.

## What

The repair pass's re-render list never clears. Once #2107 widened it (only a frozen `characterId` that is *literally* a live cast id counts as "audio is fine"), a chapter that has been re-rendered still appears on the list on the next run — permanently.

## Why

`ChapterSegment.characterId` is **deliberately** kept as the original id — see the doc comment at `server/src/tts/synthesise-chapter.ts:406-419`, and `characterId: group.characterId` at `:3233`. The repair pass writes only `cast-id-history.json`; it never rewrites `segments.json` (invariant 6 of plan 278 — frozen segments files are never rewritten to migrate ids) and never touches manuscript attribution.

So re-rendering a chapter produces a fresh `segments.json` carrying the *same* frozen id, which still resolves through the same non-`exact` tier, and the very next dry run reports the identical row.

## Failure scenario

A book's `mairin` is renamed to `mairin-2` by a merge; `retireCharacterId` records it. Chapter 2 is then re-rendered and comes out **correct**, in Мэйрин's voice, because `resolveGroup` resolves through history at render time. The repair pass still lists ch2 / `mairin` / 8 seg as damage. The operator burns an 8 GB-card re-render on it, re-runs the pass, and gets the same row back. The list has lost the ability to distinguish "repaired" from "not repaired".

This is the accepted cost of the #2107 scoping decision — over-reporting is the safe direction for a one-shot repair tool, and under-reporting is what #2107 existed to fix. But it should not stay permanent.

## The discriminator exists, and is half-available

- `SegmentsFile.synthesizedAt` (`server/src/audio/segments-io.ts:53`, written at `finalize-chapter-write.ts:325`) is present on **all 497** non-`.previous.` segment files in the real workspace — verified 2026-08-05.
- What is missing is the other half: `cast-id-history.json`'s `supersededBy` is a flat `{from: to}` map with **no per-entry timestamp**, so there is nothing to compare `synthesizedAt` against. The file's own mtime is a whole-file proxy that breaks as soon as a second alias is appended.

## Acceptance

- A segment whose render **post-dates** the alias it resolves through is not listed as damage; one that pre-dates it still is.
- The list therefore clears as the operator works through it.
- Whatever carries the timestamp is schema-versioned, and a schema-1 file without it still loads (`loadCastIdHistory` hard-requires `raw.schema === 1` today, and its validation is all-or-nothing and never throws — don't regress that).
- A pre-existing entry with no timestamp must read as **unknown**, not as "older than every render". Getting that backwards silently re-opens #2107.

## Why this needs its own design pass

It is a persisted-schema change on the file that is the identity of record's side-table, touching `retireCharacterId`, `loadCastIdHistory`, the repair script and the resolver. Deliberately **not** folded into the #2107 PR.

## Benefit

*(user)* The re-render work list becomes a work list — it shrinks as the work gets done, instead of reporting the same rows forever and training the operator to ignore it.

## Key files

- `server/src/store/cast-id-history.ts`
- `server/src/audio/segments-io.ts` (`synthesizedAt`)
- `scripts/repair-cast-id-drift.mjs` (`buildOrphansFromSegments`, `buildRerenderRows`)

dudarenok-maker:
## Summary

The repair pass's re-render list never cleared. A chapter you re-rendered came back on the very next run, permanently — so an operator could burn a re-render on an 8 GB card, re-run the pass, and get the identical row back. The list had lost the ability to tell "repaired" from "not repaired", which trains you to ignore it.

Nothing on disk could answer the question. A segment's `characterId` is deliberately frozen at its original value, so a re-render produces a fresh `segments.json` carrying the *same* orphaned id, resolving through the same non-`exact` tier. `synthesizedAt` records when audio was made, but `supersededBy` was a flat `{from: to}` map with no per-entry marker for when the alias was recorded — so there was nothing to compare against.

This adds the missing half and wires both consumers to one answer:

- **A per-book monotonic `seq` counter** on `cast-id-history.json`, with per-key `recordedAtSeq`/`recordedAtIso` markers recording *when each alias's current target was established*. `schema` stays `1`; every field is additive-optional.
- **A uniform stamp rule** — every write to the file bumps `seq`; every `supersededBy` write stamps both markers; every delete removes them. Enforced structurally rather than by a hand-maintained site table.
- **`castHistorySeq` on `<slug>.segments.json`**, stamped by full renders with the counter the render actually resolved against. The two partial writers (`chapter-qa-repair`, `chapter-splice`) carry the prior value forward verbatim — they rewrite the whole file while re-rendering only part of a chapter, so refreshing it there would launder a stale row.
- **One shared predicate**, `isAudioCurrent(resolution, segmentsFile, history)` → `true | false | 'unknown'`. The Cast banner and `repair-cast-id-drift.mjs` both call it instead of each deciding for themselves.
- **The Cast banner's auto-reconciled disclosure splits in two** — "audio is current" vs "needs a re-render" — so the screen distinguishes what it previously hedged about.

`'unknown'` is listed as damage throughout: only an affirmative comparison clears a row. Over-reporting is the safe direction for a one-shot repair tool; under-reporting is what #2107 existed to fix.

Also fixed, found in passing: an optimistic reducer for main's new "Link to this character" affordance was left writing a stale currency verdict after the merge — linking an orphan changes what it resolves to, so it now clears to `'unknown'` like its siblings.

## Test plan

- `npm run typecheck` — clean (frontend + server).
- `npm test` — 332 files, 4681 tests.
- `npm run test:server` — 0 failures. One file (`export.test.ts`) intermittently fails to report under the known fork-pool teardown flake; it is untouched by this branch and passes 25/25 in isolation (#2028).
- `npx playwright test e2e/orphaned-character-fallback-banner.spec.ts e2e/cast.spec.ts --project=chromium` — 10 passed.
- Two structural guards, run explicitly: whole-history threading and stamp pairing — 14/14.
- The acceptance criterion was verified by execution against the real predicate and real resolver: a post-dating render drops off the list, a pre-dating one stays, an unstamped one stays — and reverting the skip to the pre-fix behaviour flips the post-dating case back to listed.

**On-box acceptance is owed and recorded** — the list *clearing* can only be proven on a real workspace: run `--apply` once, re-render one chapter named on the list, then confirm that row disappears and the segment total drops by exactly that chapter's segment count while every other row remains.

Regression plan: [`docs/features/archive/280-cast-identity-followups.md`](docs/features/archive/280-cast-identity-followups.md)

Closes #2128
Refs #2129


dudarenok-maker:
Heads-up from the #2057 register reconciliation (no action needed on the code here).

This branch's **A44** row is currently **live on the canonical register artifact** — the published page at `adf22b7b-…` reads **63 owed / Group A 44**, while `main` reads **62 / 43**. So the live view was published from this branch before merge, and the canonical register is advertising a row that does not exist on `main` yet. Per the register's own rule (`r_register_twin_drift`): publish after merge, not from a branch — otherwise an abandoned branch leaves a phantom row, which is one of the three drift shapes already recorded against this file.

**Coordination, so we don't collide:** #2057's reconciliation PR needs a new Group-A row too. Since it is small, docs-only and ready now, it is taking **A44** and landing first. **Please renumber this branch's row to A45 on your next rebase**, and recompute the totals to **64 owed / Group A 45**.

Worth being explicit about why this matters beyond the ID: we would both be bumping `62 → 63` and `43 → 44` on the same two lines with *identical* text. Git auto-merges identical edits without a conflict, so the merged file would read 63/44 with two rows present — the arithmetic silently short by one. That is the exact failure `check:onbox-register` was written for, and it only catches it after the fact.

Issue #2278, opened 2026-08-11
fe-XX — LAN recovery copy hardcodes :8443, which LAN_HTTPS_PORT can change


dudarenok-maker:
## What

`src/components/lan-access-card.tsx` names `https://localhost:8443` as the address to use, in four separate pieces of user-facing recovery copy (lines ~37, ~56, ~91, and the revoke-403 branch added by #2269).

The port is not fixed. `server/src/index.ts:87` reads `LAN_HTTPS_PORT`, so a deployer who moved the server off 8443 gets recovery instructions pointing at a dead address — in exactly the situation where they are already locked out and have no other way to discover the right one.

## Why it matters more now than it did

Before #2269 this copy was **advisory**: the flows it describes (Authorize this browser, re-pair) were reachable other ways, so a wrong port was a nuisance.

Since #2269, `DELETE /api/devices/:id` is loopback-gated, and the forwarder's peer is `127.0.0.2`, so **revoke is reachable only over the direct-port loopback URL** — not `castwright.local`, not `https://localhost/` on :443. That makes this string the *only* route to a capability, and a wrong port makes the capability undiscoverable.

## Why it is filed rather than fixed inline

The frontend does not currently know the server's bound HTTPS port in this context, and there are several defensible ways to get it (surface it on an existing endpoint the card already calls; derive it from `location` when already on loopback; have the server compose the whole hint string, as it does for other operator-facing text). Picking one is an interface decision, and the copy appears in four places that should move together.

## Acceptance (sketch)

- The recovery copy names the actually-bound HTTPS port on a non-default install.
- All four occurrences move together — a fix that corrects one and leaves three is the drift this issue is about.
- Covered by a test that fails when the port is non-default.

## Context

Pre-existing; surfaced by the independent review of #2269 (`b579fae3`), which correctly declined to fix it inline as out of that change's scope.

dudarenok-maker:
## Implementation brief

**Decision (repo owner, 2026-08-12): add an `httpsPort` field to
`GET /api/lan/cert/status`.** Chosen over a new dedicated endpoint (new API
surface plus an openapi entry for one integer), over having the server compose
the finished sentence (puts user-facing copy in the server, against how the rest
of this frontend works), and over reusing `GET /api/lan` — which carries the
`LAN_AUTH_TOKEN`, and fetching a credential to parse a port out of a URL string
is not a trade worth making.

`/api/lan/cert/status` is already about LAN runtime state, and is already fetched
[...]

Issue #2276, opened 2026-08-11
chore-XX — src/main.test.tsx runs for 5.00s against a 5000ms timeout


dudarenok-maker:
## What

`src/main.test.tsx` ("main.tsx router wiring (fe-56)") has **zero timing margin**: run in isolation on an idle box it reports `tests 5.00s` against vitest's default `testTimeout` of 5000ms.

It therefore fails whenever the box is under any load. Observed 2026-08-12 during #2270: the pre-commit `verify:fast:scoped` battery went red with

```
FAIL src/main.test.tsx > main.tsx router wiring (fe-56) > mounts with react-router/dom's RouterProvider, not the bare react-router export
Error: Test timed out in 5000ms.
```

while a second agent's full `test:server` battery was running concurrently in another worktree. Re-run alone immediately afterwards: **1 passed**, duration 7.62s total / `tests 5.00s`.

## Why this is worth fixing rather than tolerating

This is not a conventional flake — there is no race and no ordering dependency. It is a fixed-cost test sitting exactly on its own deadline, so the failure rate is a function of machine load. Two Claude Code sessions committing at once is a normal working mode in this repo (CONTRIBUTING's "Running multiple Claude Code conversations"), which makes this reproducible rather than rare.

It also blocks a **commit**, not just a CI leg, because `verify:fast:scoped` gates pre-commit for any staged `src/**` change — so it costs a full re-run of a ~145s battery each time it fires.

## Possible shapes (not a decision)

- Raise this test's timeout explicitly (`it(..., { timeout: N })`) — smallest change, but leaves the 5s cost unexplained.
- Find out **why** mounting the real `main.tsx` costs 5s and reduce it. The test imports the app's real entry point; something in that path is slow enough to be worth understanding on its own merits.
- Route it through the flaky-register process (`docs/testing/flaky-register.md`) — but note `quarantinedIt` lives at `server/src/test-utils/quarantine.ts` and this is a **frontend** test, so there is no existing frontend equivalent to reach for.

## Acceptance (sketch)

- The test either runs comfortably inside its timeout, or has a timeout with a documented reason.
- Whatever is chosen, a loaded box no longer turns a green change into a blocked commit.

## Context

Found in passing while shipping #2270; unrelated to that change (which touches only the mock config catalogue). Not fixed inline because the right fix needs a judgement call among the three shapes above.

Issue #2257, opened 2026-08-11
Revoke the prior self-bind device record on re-authorize


dudarenok-maker:
## What

`POST /api/devices/pair-session` with `{ label: 'This computer' }` mints a
fresh device-token record every time "Authorize this browser" is clicked
(`src/components/lan-access-card.tsx`). Nothing revokes the *previous*
self-bind record for the same host browser, so repeated use of the button
accumulates multiple live, unrevoked `'This computer'` entries in
`GET /api/devices` — each with its own 365-day (soon-expiring-later) TTL.

Functionally harmless today (an extra valid token is not a security hole,
just clutter), but it means the device list under Account → LAN access grows
a new "This computer" row every time a user re-binds after a lapsed cookie,
with no way to tell which one is current short of checking `lastSeenAt`.

## Why this is deferred, not fixed inline

Revoking the prior record needs a way to *identify* it as "the prior
self-bind for this host," which none of today's data model expresses. That
needs a self-bind marker threaded across four files:

- `server/src/workspace/pairing-sessions.ts` — the ephemeral session record
  minted by `pair-session` would need to carry the marker through to redemption.
- `server/src/routes/pairing.ts` — `/redeem-browser` would need to read the
  marker and, on a self-bind, look up the previous self-bind record for
  revocation.
- `server/src/workspace/device-tokens.ts` — `createDevice`/`revokeDevice`
  would need a way to find "the current self-bind record," and the revoke of
  the old one plus the create of the new one must be one atomic
  read-modify-write, not two independent `enqueueWrite` calls (see
  `device-tokens.ts:499-507` — `enqueueWrite` is explicitly **non-nesting**;
  a function running inside it must never call another function that itself
  calls `enqueueWrite`, so composing "revoke then create" needs a single
  critical section, not two chained ones).
- `server/src/routes/devices.ts` — the admin device list / revoke routes
  read from the same store and would need to keep behaving correctly once a
  self-bind marker exists on some records.

That's a real data-model + concurrency design decision (what marks a record
as "the" self-bind, how the atomic swap composes with the existing
`enqueueWrite` chain), not a mechanical fix — hence filed rather than fixed
in the `feat/frontend-lan-rebind` branch that added the button.

## Acceptance (sketch, to be refined at implementation time)

- Clicking "Authorize this browser" twice in a row (with a working
  `castwright.local`) leaves exactly one live `'This computer'` device
  record, not two.
[...]

Issue #2247, opened 2026-08-11
srv-90 — One-click castwright.local re-bind + longer device-token lifetime


dudarenok-maker:
## What

Browsing the app at `https://castwright.local` dead-ends with **"Couldn't load your library — Library scan failed (401): Missing or invalid LAN access token."** every time the host browser's device token lapses, which it does on a 30-day clock.

Two things make it bite:

1. **It expires every 30 days.** `POST /api/pair/redeem-browser` mints with `lan.deviceTokenTtlDays` (default 30) and sets `__Host-cw_lan` with a matching `maxAge` (`server/src/routes/pairing.ts:163-172`).
2. **The failing origin cannot fix itself.** `__Host-cw_lan` is host-scoped, so only a response from `castwright.local` can set it — but minting requires loopback or an already-authenticated friendly-host request (`mayStartPairingSession`, `server/src/lan-auth.ts:138`). An expired browser on `castwright.local` is neither, so recovery has to start on `https://localhost:8443`.

The mechanism to re-bind already exists; it is buried in a QR flow built for phones (Account → LAN access → type a name → Authorize → find the small "Open pairing link on castwright.local" link → Authorize again) and there is no pointer to it from the screen that actually fails.

This ships a direct entry point, a way out at the failure site, and a lifetime long enough that both are rarely needed.

## Acceptance

- A **"Authorize this browser"** button in Account → LAN access re-binds `castwright.local` in one click, with no typed device name and no QR.
- The `castwright.local` 401 screen no longer prints raw JSON: it explains the problem and points at the machine running Castwright — and it keeps doing so after the user presses **Retry**.
- A newly authorized device shows an `expires` roughly a year out in the LAN access list.
- A value above the browser cookie ceiling can no longer be configured.
- The lifetime change is visible to mock mode and the wiki, not just the server.

## Key files

- `server/src/config/registry.ts:1254-1263` — the knob (`default`, new `max: 400`)
- `server/src/workspace/device-tokens.ts:84-87` — `clampTtlDays` (currently untested)
- `src/lib/api.ts:9446`, `:7363` — mock registry mirror + mock redeem expiry
- `src/components/lan-access-card.tsx` — the button, in the authorized branch
- `src/views/pair.tsx` — auto-redeem on `self=1`, scrub-before-redeem, Retry
[...]

PR #2280, opened 2026-08-11
fix(server,frontend,ops): gate device revoke to loopback and close two #2247 follow-ups

### LINKED ISSUE #2269: DELETE /api/devices/:id has no loopback gate — any paired LAN device can revoke the host's credential
## What

`DELETE /api/devices/:id` (`server/src/routes/devices.ts:128-139`) has **no
loopback gate**. Any caller that clears the LAN guard — an already-paired
phone holding a valid per-device token, or anything holding the shared
secret — can revoke any other device's credential, including the host
computer's own.

Concrete sequence, no exotic conditions:

1. A paired LAN device calls `GET /api/devices`. The response includes every
   record's `id` — `redactDevice` (`device-tokens.ts:82`) puts `id` in
   `PublicDevice` deliberately, because the admin UI needs it to offer Revoke.
2. That device calls `DELETE /api/devices/<the host's record id>`.
3. `revokeDevice` succeeds. The host's browser is now locked out of its own
   Castwright install until someone re-authorizes at the desktop.

## Why this looks like an oversight rather than a decision

The sibling **mint** route one screen up is loopback-gated, with an explicit
rationale spelled out in the code (`devices.ts:57-64`):

> admin mint — LOOPBACK-ONLY (defense-in-depth: a stolen browser cookie must
> NOT be able to mint a fresh, durable device token that survives revoking the
> stolen one — minting stays a physical-desktop capability)

The same reasoning applies at least as strongly to revocation: a stolen cookie
that cannot mint can still *lock out the legitimate owner*, which is the
denial-of-service half of the same threat model. Nothing in the code or the
plan history records a decision to treat revoke differently — the asymmetry
appears to be unnoticed rather than chosen.

## Why this is filed, not fixed inline

It is a genuine interface decision with more than one defensible answer, so it
fails the fix-now bar:

- **Gate it to loopback**, symmetric with mint — but this removes the ability
  to revoke a lost phone *from* a paired tablet, which may be a flow someone
  relies on, and the Account → LAN access UI is reachable from any paired
  device today.
- **Gate it to loopback OR self**, letting any device revoke *itself* (the
  "sign out this device" affordance) while reserving revoking *others* to the
  desktop.
- **Leave it open and document it** as intended peer-management behaviour.

Picking one needs a call on who the device roster is for, plus a matching UI
change wherever the Revoke button is now shown to a caller who would no longer
be allowed to press it.

## Acceptance (sketch, to be refined at implementation time)

- A decision recorded on which of the three shapes above is intended.
- Whichever gate is chosen is enforced server-side and covered by a test that
  fails without it — the current `devices.test.ts` has no case asserting who
  may revoke.
- If revocation narrows, the Revoke control is hidden or disabled for callers
  who can no longer use it, rather than presenting a button that 403s.
- `openapi.yaml` reflects the resulting contract. (Note: `/api/devices/pair-session`
  is currently absent from `openapi.yaml` entirely — a separate pre-existing gap.)

## Context

Found by the independent review of commit `51aff15e` (#2257) while attacking
the self-bind trust boundary — the reviewer reached "a LAN device revokes the
host's credential" without needing the new self-bind marker at all. Not
introduced by that change; the marker work is unaffected either way.

### LINKED ISSUE #2270: Retire the two UI-only mock config descriptors and their legacy groups
## What

`src/lib/api.ts`'s mock config catalogue is now projected from the server
registry (#2259), with exactly two hand-written exceptions left behind:

- `KOKORO_SAMPLE_RATE` — no registry equivalent exists at all.
- `ANALYZER_STAGE1_PROMPT` — superseded by the registry's six real `isPrompt`
  knobs (`server/src/config/registry.ts:1198-1248`), none of which use this key.

Both are fictions: they describe knobs the server does not have, in two legacy
groups (`'tts'`, `'analyzer'`) that exist in `MOCK_CONFIG_GROUPS` solely to
host them. So mock mode's Settings UI shows two rows that cannot exist against
a real backend, in two groups a real backend never returns — which is exactly
the dishonesty the parity work was meant to remove.

## Why they were kept rather than deleted with the rest

They are load-bearing for the mock *prompt* machinery and its tests, not just
for the descriptor list:

- `mockGetPrompt` / `mockPutPrompt` / `mockResetPrompt` are driven with the id
  `'ANALYZER_STAGE1_PROMPT'`.
- Roughly a dozen cases in `src/lib/api.config.test.ts` use one or the other as
  their worked example — the numeric round-trip, the group-reset case, and the
  whole `mockGetPrompt / mockPutPrompt / mockResetPrompt` block.

Retiring them therefore means re-pointing the prompt mocks and their tests at
a real registry prompt key, and picking a real registry key for the numeric
round-trip cases. That is a self-contained piece of work, but it is not the
one-line deletion it looks like, so it was left out of #2259 rather than
smuggled in.

## Acceptance (sketch, to be refined at implementation time)

- The two UI-only descriptors and the two legacy groups (`'tts'`, `'analyzer'`)
  are gone; `MOCK_CONFIG_GROUPS` is exactly the registry's `GROUPS`.
- The prompt mocks are driven by a real registry `isPrompt` key, so mock mode's
  prompt-fork flow exercises a knob the server actually has.
- The `api.config.test.ts` cases that used the two fictions are re-pointed at
  real registry keys and still assert the same behaviours.
- The parity guard's "every mock descriptor's group resolves" assertion still
  holds, and the guard no longer needs a UI-only carve-out at all.

## Context

Follow-up from #2259 (`chore/server-lan-rebind-followups`). Surfaced by the
implementation, which also removed a third hand-copy (`tts.qwen.device`) that
duplicated a real registry key — these last two are the remainder.

### LINKED ISSUE #2272: check:onbox-register --against-published cannot tell a discharged row from a lost one
## What

`npm run check:onbox-register -- --against-published <saved live page>` cannot
distinguish the two reasons a row can be present on the published page and
absent from the register:

1. **another lane published it** and you are about to silently delete it — the
   failure mode the check exists to prevent (it caught exactly this on
   2026-08-11, saving row A44); and
2. **you just discharged it** — the row is gone on purpose.

Both produce the identical signal, so the check refuses (2) as loudly as (1).
Observed 2026-08-12 discharging **E10**: the check reported "the register is
BEHIND what is already published … Do not publish", when in fact the register
was correctly *ahead*.

## Why it matters

**Every legitimate discharge trips this.** Removing a row is not an edge case —
it is the register's whole purpose, the thing every acceptance session is
working toward. A required-feeling gate that fires on the normal happy path is
one people learn to force past, and then it stops protecting the case it was
built for.

The manual reconciliation it forces is real work: extract every `class="num">X<`
from the saved live HTML, extract every `^### X` from the register, `comm -23`
them, and eyeball that the only difference is the row you meant to remove.

## Possible shapes (not a decision)

- **A `--discharging E10[,E11]` flag** — the caller names the rows they are
  deliberately removing; the check then fails only on an *unnamed* live-only
  row. Cheap, explicit, and leaves an auditable record of intent in the command.
- **Diff against the `origin/main` baseline** — a row present on the live page
  AND on `origin/main` but absent from the branch is an intentional removal by
  this branch; a row on the live page but absent from BOTH is a competing lane's.
  No new flag, but relies on the baseline being readable and consistent (see
  below).
- Leave as-is and document the manual reconciliation in the register's "Live
  view" section, so at least it is a known, written-down step rather than a
  surprise.

## Related sharp edge, same command

The check **fails closed when `origin/main`'s register is itself internally
inconsistent** — it refuses to run at all, because it cannot trust the baseline.
Correct behaviour, but it means a broken `main` blocks the pre-publish check for
the very PR that fixes `main`. Hit on 2026-08-11 (PR #2271, where the total said
64 against 65 rows). If the second shape above is chosen, this coupling gets
tighter and deserves an explicit fallback.

## Acceptance (sketch)

- A discharge that removes exactly the intended row(s) passes the check without
  a hand-run `comm` diff.
- A row that is live and unaccounted-for still fails, with the row ID named —
  the A44 case must keep failing.
- Whatever is chosen is documented in the register's "Live view" section, which
  is the procedure people actually follow.

## Context

Surfaced while discharging E10 (plan 283 / #2247). The check is
`scripts/check-onbox-register.mjs`; the procedure is the "Live view" section of
`docs/testing/onbox-acceptance-register.md`.

dudarenok-maker:
## Summary

Three follow-ups left over from #2247, reconciled into one PR per CONTRIBUTING's
reconciliation pattern. Two carried a real decision; both were taken by the repo
owner before implementation and recorded as implementation-brief comments on the
issues.

### #2269 — `DELETE /api/devices/:id` is now loopback-gated

Any paired LAN device could revoke any other device's credential, including the
host computer's own. The sibling **mint** route one screen up has been
loopback-gated since srv-33 with an explicit "a stolen cookie must NOT be able to
mint" rationale; revocation is the denial-of-service half of the same threat
model and had no gate at all.

**Decision: loopback-only, symmetric with mint.** "Loopback OR self" was rejected
on cost, not principle, and the reasoning is recorded on the issue so it survives:
`isValidDeviceToken` returns a bare boolean, so the LAN guard never learns *which*
record matched, and a caller authenticating with the shared `LAN_AUTH_TOKEN` has
no device record at all — so "self" would need both a new identity path and its
own answer for that branch, to serve a "sign out this device" affordance that does
not exist today.

**Consequence worth stating plainly, because it is not obvious.** `LOOPBACK` is
`{127.0.0.1, ::1, ::ffff:127.0.0.1}` and the `:443` forwarder dials upstream with
`localAddress: '127.0.0.2'` — so **every** request through the forwarder is
non-loopback, the host's own browser included. That is correct: the forwarder is
host-blind, so a phone on `castwright.local` and the desktop on `castwright.local`
are indistinguishable at the socket. But it means revoke works only at
`https://localhost:<port>` — not `castwright.local`, and not `https://localhost/`
on :443. Since #2247's "Authorize this browser" deliberately navigates the owner
*to* `castwright.local` and leaves them there, the UI and the release notes now
say so instead of pretending otherwise.

### #2270 — the last two mock config fictions are gone

`KOKORO_SAMPLE_RATE` and `ANALYZER_STAGE1_PROMPT` described knobs the server does
not have, in two legacy groups that existed only to host them. `MOCK_CONFIG_GROUPS`
is now exactly the registry's `GROUPS`, and the parity guard asserts **zero** extra
descriptors rather than "registry plus the two declared UI-only ones" — a
strengthening, since the old form would have accepted a third fiction.

### #2272 — `check:onbox-register --against-published` learns `--discharging`

**The issue was wrong about its own fix, and the brief corrected it.** It listed
"diff against the `origin/main` baseline" as an unbuilt alternative; that shipped
in #2199. The real defect is narrower: **the baseline can only recognise a
discharge that has already merged**, and publishing happens *before* merge, so at
check time `origin/main` still has the row being removed. That is why it refused
*every* discharge rather than some. The flag complements that logic; it does not
replace it.

## What the independent reviews caught that green suites did not

Each issue got a hostile review pass and at least one fix round. Findings that
mattered:

- **#2270 shipped two Criticals into `e2e/`.** Deleting the `tts` mock group
  orphaned two Playwright specs — one of them the *only* browser-level coverage of
  the restart-sidecar banner chain. The file is `mode: 'serial'`, so the first
  failure also skipped the four specs after it, on a required check. Re-pointed,
  not deleted. This one was my briefing error: I scoped the grep to `src/`.
- **#2272's flag could not handle a whole-group discharge**, and said so falsely.
  The consumption loop iterated the groups the *working register still has*, so a
  group whose last row was discharged was never visited — then the refusal claimed
  the named ID "is not a live-only row (present on the published page, absent from
  this register)" when it satisfied that definition verbatim. Group F has exactly
  one row, so this was live, not hypothetical.
- **#2269's own fix introduced a defect.** The explanation added where the Revoke
  button disappears reused `recoveryHint()`, which tells the user to click
  "Authorize this browser" — a control that always navigates to `castwright.local`,
  which is exactly where revoke does not work. Reload, same message, forever. The
  test missed it because it asserted *some* explanatory text was present, not that
  the text was correct — it would have passed with either message.

The single highest-value item in the round is a test, not a fix: #2269's original
coverage **mocked** `isLoopbackRequest` to return `false`, which proves the handler
consults the gate and nothing about what the gate decides. Anyone "fixing" the
usability consequence above by adding `127.0.0.2` to `LOOPBACK` would have kept
every test green while reopening the hole. There is now a case driving the router
with the real peer (`127.0.0.2`, `Host: castwright.local`), asserting status *and*
body, mutation-checked against exactly that edit — it turns two tests red,
including the pre-existing #2257 self-bind case.

- A stale knob label in plan 199 ("Max re-records per segment" → "Signal QA max
  re-records"), and a walkthrough step that listed two rows as though both lived in
  the same section when one is in "Per-sentence QA gates". Both in a file this
  branch already rewrote.

## Filed, not fixed

- **#2276** — `src/main.test.tsx` runs for 5.00s against a 5000ms timeout. Not a
  conventional flake: no race, no ordering dependency, just a fixed-cost test
  sitting exactly on its own deadline, so it fails as a function of machine load.
  It blocked a commit during this round while another worktree's battery ran.
- **#2278** — the LAN recovery copy hardcodes `:8443` in four places, but
  `LAN_HTTPS_PORT` can move it. Advisory before; now that revoke is reachable only
  over the direct-port loopback URL, this string is the sole route to a capability.

## Test plan

- `npm run verify` — full battery on the reconciled branch (not per-merge; each
  branch was independently verified before reconciliation, and all three merged
  with zero conflicts).
- `#2269`: `devices.test.ts` 25/25, `lan-access-card.test.tsx` 21/21. Mutation
  check: adding `'127.0.0.2'` to `LOOPBACK` turns the new security case red (and
  the #2257 case with it); reverted, green.
- `#2270`: full frontend suite 4712/4712; `e2e/advanced-settings.spec.ts` 7/7,
  run directly after the re-pointing.
- `#2272`: `scripts/tests/check-onbox-register.test.mjs` 98/98, including the
  Group F whole-group scenario re-run against the real committed register and
  live view. Each new case verified red with the implementation reverted.

## Release notes

`#2269` lands entries in both `docs/release-notes-next.md` and `RELEASE_NOTES.md` —
it narrows a user-visible capability.

`#2270` and `#2272` are **explicitly exempt** under before-shipping step 5: the
first is mock-only with no user- or operator-visible effect, the second is dev/CI
tooling that does not ship in the release artifact. Stated rather than silently
omitted.

Closes #2269
Closes #2270
Closes #2272


dudarenok-maker:
## Review-gate findings folded in

The mandatory gate ran at `high` effort (multi-scope). It cleared the load-bearing claims by executing them rather than reading them — `lan-auth.ts`/`app.ts`/`lan-port-forwarder.ts` byte-identical to merge-base, the `127.0.0.2` widening mutation reproduced, `--discharging` probed with case-varied / bare-group-letter / whitespace / duplicate / `",,,"` / two-section inputs, and the `#2269`↔`#2270` cross-change interaction confirmed (`ApiError` still exported and still carrying `.status` after #2270's rewrite of `src/lib/api.ts`).

Six commits address what it found.

### The one that mattered — `b07672b8`

**The operator procedure I wrote for `--discharging` was a recipe for the incident the check exists to prevent.** Both remedy branches told the operator to feed back whatever IDs the error names and re-run until green. Worked example, verified against the merged code: you discharge Group F's only row, another lane concurrently publishes a new F2 into that group, and `--discharging F1,F2` goes **green while silently deleting F2** — #1931/A44 verbatim.

The rule is arithmetic and appeared nowhere: *name exactly as many IDs from a group as rows you actually discharged from that group.* Anything left over after your true count is another lane's row, and the fix is to merge it in. The doc now says so, says explicitly that padding until green is the failure mode, and ends with "when the doc and the tool disagree, trust the tool" — the tool's own text already said `merge it in before publishing`.

### The rest

- `99e24453` — the #2269 release note described the empty action cell as using `recoveryHint()`. This PR's own round-2 review **deleted** that, because it points at "Authorize this browser", which navigates to `castwright.local` — precisely where revoke does not work. Third instance of prose describing a mechanism its own PR removed. Also resolved a forward pointer claiming #2270 was still "tracked", and added technical-register entries for #2270/#2272.
- `d18f2fc2` — the hint rendered **once per device row**. Four paired devices meant the same 150-character sentence four times, crushing the label/date columns on a phone. Now once per card.
- `9f7bb181` — the PR claimed symmetry with the mint route while documenting `403` on only one of them, and `503` on neither. Both routes now match what the code sends; `api-types.ts` regenerated.
- `0fa6de84`, `47e7bf7d` — a fixture contradicting the policy stated in its sibling file, and an argument-order fix so a bare `--discharging` reports the useful reason first.

### E10 — on-box acceptance recorded (`472f83f2`)

The gate was right that this was missing, and CLAUDE.md step 3 makes recording it a merge gate. **Nothing automated reaches the boundary #2269 turns on**: the server test *fabricates* `req.ip = '127.0.0.2'` and the frontend test *stubs* `window.location`. Neither has seen the real `:443` forwarder — which is what makes the host's own browser non-loopback, since it dials upstream from `127.0.0.2` and is host-blind.

Four observations, sharing setup with E2/E3. The one that is easy to get wrong: check the replacement hint with **3+ paired devices**, because with one device the per-row bug is invisible.

Owed **64 → 65**, group E **9 → 10**. Live view updated in the same diff, `check:onbox-register` green, and `--against-published` against the currently-live page confirms the register is not behind it. Published to the canonical URL recorded in the register header — not a new one.

### Not taken

The hardcoded `https://localhost:8443` (gate finding 5) is **#2278**, filed with a decision already taken: the port will come from an `httpsPort` field on `GET /api/lan/cert/status`. It is deliberately not started here because it rewrites the same `lan-access-card.tsx` code this PR restructures.

Issue #2245, opened 2026-08-11
srv — isSpokenLine duplicates the language conventions: make it conventions-driven and stop the no-table fallback demoting everything


dudarenok-maker:
`isSpokenLine` (`server/src/analyzer/narrator-default.ts:29`) is language-blind. It
carries one German quote form of the four in `de.quotePairs`, so
`applyNarratorDefault` demotes real German dialogue to the narrator whenever the
structure-engine branch is not taken.

Measured 2026-08-11, all four `de.quotePairs` forms in both positions:

| Form | Leading — `»Lass das.«` | Embedded — `Er sagte »Lass das.« und ging.` |
|---|---|---|
| `„…“` | spoken | spoken |
| `„…”` | spoken | **missed** |
| `„…"` | spoken | **missed** |
| `»…«` | **missed** | **missed** |

The opener class `/^[«„"“‘']/` carries `„` but not `»`, so three of four are caught
when the turn starts the sentence. The embedded alternatives carry only
`„[^“]+“` for German, so three of four are missed when narration comes first.

**This is not exotica.** `server/src/analyzer/dialogue-structure/lang/de.ts:7-9`
records the ASCII `"` and the `”` glyph as what "real-world manuscripts (incl. our
translated demo books) **routinely**" use — so the two forms that fail in embedded
position are the common ones.

## Why it is not visible damage today, and why it is still worth fixing

The one live German book, _Der Auftrag von Coalfall_, opens every line with `„` at
position 0, so it reads clean. The next German import in any other form does not.
`isSpokenLine` is the last thing standing between a book and total attribution loss
whenever the structure-engine branch is skipped — which is a user-settable knob
(`analyzer.structure.enabled`, `registry.ts:1267`), not an impossible state.

## Why it needs design rather than a patch

`isSpokenLine` takes `(text)` and no language, so there are two defensible repairs
and no obvious winner:

1. **Add the patterns globally.** Cheap, but `»` is not German-exclusive and the
   function runs on every language — an unmeasured cross-language blast radius. The
   CJK pair was permitted into #1984's scope precisely because its blast radius
   *could* be proven zero (no `「`/`『` in any Latin or Cyrillic cache). No such
   proof exists for `»`.
2. **Make `isSpokenLine` language-aware.** Correct, but it is a signature change on
   a hot analyzer path with several callers, and it duplicates the job
   `conventionsFor` already does — which raises whether the two should merge.

Two defensible options with an interface decision between them is design work by
[...]

PR #2286, opened 2026-08-12
fix(server): widen all seven languages' quote-pair tables into M2's secondary tier

### LINKED ISSUE #2279: srv — the language conventions tables are missing quote styles real manuscripts use
The seven `LanguageConventions` tables (`server/src/analyzer/dialogue-structure/lang/`) are the single source of truth for what counts as dialogue — #2245 made `isSpokenLine` read them so there is exactly one definition instead of two. That is only as good as the tables, and the tables are missing conventions that real manuscripts use.

## The condition

Because the tables are now the only definition, **any convention a table does not carry reads as narration.** Two consequences are broad enough to name separately:

- **(a) `dialogueOpen` is `null` in `en`, `de`, `zh` and `ja`** — a leading dash is not dialogue in any of the four. The bundle #2245 replaced treated a leading dash as dialogue in *every* language.
- **(b) A quote pair absent from `quotePairs` is not dialogue** — including a `zh`/`ja` asymmetry that splits the same line by language.

## The rows

Measured 2026-08-12 by running the shipped `isSpokenLine` against the bundle it replaced. **The condition above governs; this table is the set of rows examined so far, not a closed enumeration** — the full cross-product of seven languages against the old bundle's `« „ " “ ‘ '` openers, dash forms and entities is roughly forty rows, most of them inert (`«…»` in `ja`, `„…“` in `es`). It is nonetheless the canonical list: `narrator-default.test.ts`'s `#2279 (a)` and `#2279 (b)` blocks are its executable copy row for row, and #2245's release-notes entry states the condition and defers here for the rows. **A row added anywhere belongs in all three.**

| | language | text | old bundle | tables |
|---|---|---|---|---|
| (a) | `en` | `—Dame Alina` | spoken | **narration** |
| (a) | `de` | `— Komm her, sagte er.` | spoken | **narration** |
| (a) | `de` | `&mdash; Komm her` | spoken | **narration** |
| (a) | `zh` | `——你好` | spoken | **narration** |
| (a) | `ja` | `— こんにちは` | spoken | **narration** |
| (b) | `de` | `"Hallo", sagte er.` | spoken | **narration** |
| (b) | `de` | `“Hallo”, sagte er.` | spoken | **narration** — `“` is German's *closer*, never an opener |
| (b) | `de` | `«Lass das.»` (Swiss order) | spoken | **narration** |
| (b) | `es` | `"Hola", dijo.` | spoken | **narration** |
| (b) | `fr` | `"Bonjour", dit-il.` | spoken | **narration** |
| (b) | `fr` | `“Bonjour”, dit-il.` | spoken | **narration** |
| (b) | `ru` | `‘Привет’` | spoken | **narration** |
| (b) | `en` | `«Bonjour»` | spoken | **narration** |
| (b) | `ja` | `“おはよう”` | spoken | **narration** — but `“你好”` is **spoken** in `zh` |
| (b) | `es` | `&ndash; Un momento` | spoken | ~~narration~~ → **dialogue — FIXED by #2289** (PR #2309) |
| (b) | `fr` | `&ndash; Un instant` | spoken | ~~narration~~ → **dialogue — FIXED by #2289** (PR #2309) |

The gaps, per table — the inventory as it actually is:

| | `dialogueOpen` | `quotePairs` |
|---|---|---|
| `en` | `null` | `“”`, `""`, `‘’` |
| `ru` | `&mdash;` \| `&ndash;` \| `-–—` | `«»`, `„“`, `“”`, `""` |
| `es` | `&mdash;` \| `&ndash;` \| `-–—` | `«»`, `“”` |
| `fr` | `&mdash;` \| `&ndash;` \| `-–—` | `«»` |
| `de` | `null` | `„“`, `„”`, `„"`, `»«` |
| `zh` | `null` | `「」`, `『』`, `“”` |
| `ja` | `null` | `「」`, `『』` |

Reading across: `de` carries **both** the ASCII `"` and the curly `“` only in the *closer* slot behind a `„` opener, though its own header comment (`de.ts:6-9`) says real manuscripts "**routinely**" drift to those glyphs — so a German book run through English-typeset tooling, `“Hallo”`, is the most plausible real-world instance of this whole class; `es`/`fr` `dialogueOpen` **carried** `&mdash;` but not `&ndash;`, where `ru.ts:9` carries both — **fixed by #2289 (PR #2309); all three tables now match**; `fr` has one pair; `ja` is missing the `“”` its `zh` sibling carries; `en` has no guillemets and deliberately no straight-single `'…'` pair (a same-glyph `'` would read every apostrophe as a quote run — see #2245).

## Why this is not just an `isSpokenLine` problem

The same tables drive `findQuoteRuns` → `parseChapterStructure` → `crossExamine`, i.e. **the default path**. A German manuscript typeset with ASCII quotes gets no quote runs today, engine on or off — so the structure engine cannot anchor a single speaker in it. #2245 did not cause this; it made the fallback consistent with the engine, which is what surfaced it.

Where #2245 *did* change something: with `analyzer.structure.enabled` off, every row above used to be caught by the language-blind bundle and is now missed. That path is a knob, not the default, and the live 20-book corpus contains zero instances of any row (it holds one book each of de/es/fr/zh/ja — existence proofs, not distributions) — but "zero on a corpus that cannot contain the shape" is not evidence of safety, which is why this is filed rather than assumed harmless.

## Why it was not fixed inside #2245

It fails CLAUDE.md's fix-now bar — but for **two different reasons**, because `dialogueOpen` and `quotePairs` have disjoint consumers and conflating them would hand an implementer the wrong constraint:

- **`quotePairs` rows (b).** Adding a pair changes `findQuoteRuns` (`parser.ts:225`) on the structure engine's hot path. The leftmost-wins overlap rule at `parser.ts:207-214` *appears* to protect `de`'s existing `„…"` closer role from a new `"…"` opener — the `„` run starts earlier, so the `"` candidate is discarded as overlapping — but that is an argument from reading the code, and the corpus cannot produce a counter-example because no book in it uses the shape.
- **`dialogueOpen` rows (a), and the `&ndash;` rows.** *(The `&ndash;` half is now DONE — #2289, PR #2309. Only the row-(a) `dialogueOpen: null` half is still open.)* These never reach `findQuoteRuns` at all; they change paragraph classification at `parser.ts:99` and `cross-examine.ts:225` — the dash-vs-quote split — which is a *different* default-path risk, not a smaller one. The `&ndash;` additions have an exact in-family precedent (`ru.ts:9` carries both entities, for the documented `stripHtml` reason) and are the closest thing here to an obvious fix; the `dialogueOpen: null` rows are not — whether German or Japanese dash dialogue should be recognised at all is a typography question, not a code question.

And across both: which conventions belong in which table is a research question with more than one defensible answer. Answering it per-glyph, reactively, is how the present state accumulated.

## Acceptance

- Each table's inventory justified against how the language is actually typeset, not against one book.
- The `findQuoteRuns` / `parseChapterStructure` delta **measured** over real manuscript text — the de/es/fr/ru EPUBs parsed through the repo's own parsers, not through cache sentences, since the risk is paragraph-level run merging (#1601's shape).
- The `dialogueOpen` rows measured separately, against `parser.ts:99`'s paragraph split.
- Every row of the table above either passes or is documented as deliberately excluded, with the reason — **and the table extended to the full cross-product first**, since it is a work-in-progress list, not a closed one.
- The `zh`/`ja` asymmetry resolved in one direction or the other, explicitly.
- A same-glyph opener/closer added to a table must not create a run over an apostrophe or a possessive.

## Key files

- `server/src/analyzer/dialogue-structure/lang/{de,es,fr,ru,en,zh,ja}.ts`
- `server/src/analyzer/dialogue-structure/parser.ts:192` — `findQuoteRuns`, `:207-214` — the overlap rule, `:99` — the paragraph split
- `server/src/analyzer/narrator-default.ts:34` — `isSpokenLine`, the second consumer

## Provenance

Found by the independent review gate on PR #2277 (round 1, finding C2); the table was completed and the deferral reason split by round 2 of the same gate, which caught that the first version enumerated only the rows that had been examined and stated one constraint for two mechanisms.

Refs #2245, #1984

---

> **Executable-copy reconciliation, 2026-08-13.** The two `&ndash;` rows above are struck and the `es`/`fr` gap-inventory cells now show both entities, because #2289 (PR #2309) fixed them. This issue declares `narrator-default.test.ts`’s `#2279 (a)` / `(b)` blocks its executable copy and states that **a row added anywhere belongs in all three surfaces** — PR #2309 moved the executable copy, so this body moves in the same round rather than being left to drift. Nothing else in the residue changed, and no `quotePairs` table was touched.
>
> Separately: an entity being *recognised* as a dialogue opener does not mean it is *spoken* correctly — it still reaches the TTS engine verbatim. That is **#2310**, which names the design decision owed (which layer decodes the entity).

dudarenok-maker:
## Summary

**#2245 made the seven `LanguageConventions` tables the single definition of dialogue. This PR fixes the tables.** One definition is only as good as what it contains, and what it contained was each language's *textbook* typography. Real manuscripts drift — `es`/`fr` had no straight-double pair (`fr` had **one pair total**), `ru` had no smart-single, `en` had no guillemets, and `ja` was missing the `“”` its `zh` sibling carried, so the *same line* was dialogue in Chinese and narration in Japanese.

The original attempt (commit `3e1b56fb`) added these pairs straight to each language's PRIMARY `quotePairs`. Review round 2 found that put a straddle risk into the primary tier itself, and filed **#2288** — whose M2 milestone (`findQuoteRuns`' two-tier scan, `69ced6c5`) exists specifically to give this PR a tier to land in that can't do that. This PR lands them in `secondaryQuotePairs`.

## What ships

| lang | added (in `secondaryQuotePairs`) |
|---|---|
| `es` | `"…"` |
| `fr` | `"…"` · `“…”` |
| `ru` | `‘…’` |
| `zh` | `"…"` · `‘…’` (Mandarin's **nested** quote inside `“…”`) |
| `ja` | `“…”` · `"…"` |
| `en` | `«…»` |
| **`de`** | **`"…"` · `“…”` · `«…»`** |

**No engine logic changes.** `parser.ts` carries three comment corrections and nothing else — this widening falsified them (`UNSPACED_SCRIPT` and the `APOSTROPHE_SHAPED` note both said `en` was the only table pairing an apostrophe-shaped closer; it no longer is) or named it as delivering a precondition it does not (`crossGlyphBound`, re-checked across all seven tables). `git diff origin/main -- parser.ts` is comments only.

**Swiss German is the one partial delivery, and it is stated up front rather than buried in the measurements.** German's primary table already carries `['»','«']`, so it owns `«` as a closer; a `»` ending one Swiss turn is also a valid primary opener, and a second Swiss turn in the same paragraph seeds a primary run that swallows the attribution between them — both real turns read as narration, the narrator's words read as speech, identically before and after this PR. Isolated over the 40-book German corpus: the Swiss entry **works in 16 paragraphs and is blocked by the collision in 4,926.** Plain and curly work at every arity.

It ships anyway, and the argument is **not** that the collision is small — it is that a blocked entry is *inert*, so those 4,926 paragraphs are unchanged rather than damaged. What the entry buys is the 16, plus `isSpokenLine` (which unions both tiers and gets the line right where the run boundary doesn't), plus the fact that #2352's eventual fix then needs no table change. `parser.test.ts` pins the broken shape as a known gap.

> **An earlier revision of this PR cited the 16 as the collision's size.** It is the opposite quantity — where the collision bites, nothing changes, so those paragraphs are excluded from the 16 by construction. The real footprint is ~300× larger, and the wrong number was sitting in the sentence that authorised shipping the entry. Caught by the re-review; the corrected figures are above and in `de.ts`.

## German was excluded for a mechanism that no longer exists

German was the one supported language #2279 gave nothing, and `de.ts` carried a 32-line comment explaining why, ending:

> Recognising drifted German therefore needs an engine change — role-aware candidate suppression, per-paragraph convention detection, or import-time normalisation — **not a table entry**.

That was measured when the pairs were headed for the **primary** `quotePairs`, against a parser with neither M2's secondary tier nor #2315's re-open bound. **The engine change it demands has since shipped** — the secondary tier declines a candidate outright when it straddles into a primary turn, which is precisely the failure the comment describes. Re-measured 2026-08-14 against this branch's own merged code:

- each candidate declared secondary, alone: **6,250 generated shapes each (18,750 total), 0 DESTROYED** (spurious, informational: 48 / 37 / 64);
- a German corpus arm that had **never run**: 40 books, 63,941 paragraphs, 23,925 with a speech run — **261 CHANGED, all GAINED, 0 lost / 0 merged / 0 split.** Concentrated, not broad: `pg/de/77073.txt` is 217 of the 261; 12 books gain dialogue.

The primary-table prohibition still stands, but its stated grounds were partly stale too: of the three candidates, only ASCII `"…"` still destroys a turn in `quotePairs` — curly and Swiss no longer do, and Swiss now reads *better* than the baseline. `de.ts`'s "all three candidates fail" is corrected rather than left to mislead a future maintainer.

The repo owner made the call to widen German here rather than defer it, with these measurements in hand.

## The acceptance criterion, widened by the owner on 2026-08-13

The tier makes a secondary candidate decline outright if it straddles into a primary run — but it does **not**, and was never meant to, stop a secondary-convention quotation in bare narration (a sign, a title, a scare quote) from reading as its own speech run when no primary run is nearby.

**The binding rule is zero DESTROYED turns, not zero spurious spans** (recorded in [`docs/superpowers/specs/2026-08-13-gap-seeded-straddle-design.md`](docs/superpowers/specs/2026-08-13-gap-seeded-straddle-design.md) → *Decided, 2026-08-13*). A spurious narration-read-as-speech line is the accepted lesser harm — nothing is deleted, a line is merely misclassified, and it stays with the analyzer rather than being forced to the narrator. German carries the largest share of that residual (**416 of 705** spurious shapes), the expected price of three pairs on the language with the broadest closer-drift.

This supersedes the original criterion's items 1–2, which counted spurious spans — the very sweep that measured them also proved those shapes contain zero destroyed turns, so that criterion was passable without changing anything.

## Re-measured against the SHIPPED guard, not the parser this branch was written on

This branch was 90 commits behind `origin/main`, which has since gained **#2315** (the re-open bound and the tag-clause guard). Every figure below was re-run against the merged branch, with the baseline being `main` @ `2fcfda0e` — **not** carried over from the pre-#2315 round.

| measurement | instrument | result |
|---|---|---|
| Seven-language sweep, all twelve pairs at once | `sweep-7lang.mts` | **70,983 shapes — 705 spurious (informational), 0 DESTROYED** |
| Each of the twelve pairs individually, declared secondary | `percandidate-7lang.mts` | **0 DESTROYED, all twelve** |
| Straddle (F2) + gap×nest (F3), es/ru/en **and a new `de` arm** | `tier-sweep.test.ts` (committed) | **0 destroyed, 0 broken nests**; tier engagement pinned at 74/253/130 and 124/632/277, plus 190/230 for `de`. The es/ru/en zeros are append-only-guaranteed (see below); `de`'s are not |
| Suppression class — a primary-convention title in narration must not suppress the paragraph's own secondary-convention dialogue | `suppress-7lang.mts` | **0 of 33** |
| **331-book, 7-language corpus replay** (Gutenberg + Standard Ebooks, real `parseChapterStructure` on both sides) | `corpus-7lang.mts` vs `main` @ `2fcfda0e` | 726,385 paragraphs, 289,983 with a speech run — **287,431 byte-identical, 2,552 CHANGED, ALL GAINED. 0 lost / 0 merged / 0 split** |
| `dialogue-structure` + `narrator-default` | `npx vitest run` | **421 passed** |
| Full server suite | `npm run test:server` | **7,657 passed / 9 skipped** |
| Typecheck | `npm run typecheck` | clean |

Three figures moved when the baseline gained #2315, and are stated at their new values rather than their old ones: the sweep's spurious count (284 → 289 for the six languages, before German), the corpus gains (2,307 → 2,291 for six languages), and `tier-sweep.test.ts`'s pins (88/225/114 → 74/253/130, 116/618/258 → 124/632/277). The old figures are #2315's delta, not this widening's.

### What the corpus replay does NOT establish — a correction the review gate forced

`findQuoteRuns` (`parser.ts:471`) computes `primaryRuns` from the primary table, seeds `out = [...primaryRuns]`, and only ever *appends* non-overlapping candidates. So **any arm whose two sides differ only in secondary-tier content cannot report a destroyed turn** — the candidate run set is always a superset of the baseline's. `parser.ts:468-470` already says this in the codebase's own voice: *"This cannot delete a primary run … hold by construction, not by measurement."*

That property covers **five of the rows above**, not just the corpus replay: the seven-language sweep, the per-candidate sweep declared secondary, the suppression class, the replay's `0 lost / 0 merged / 0 split`, and — for its `es`/`ru`/`en` languages — the `tier-sweep.test.ts` row's `0 destroyed, 0 broken nests`, whose `ref` is tier-empty for exactly those three. (Its `de` arm is the carve-out, and is tabulated below as one of the rows that *can* fail.) Their numbers are correct and worth having — they establish the **size and distribution of the gain** — but none of them could have come out differently.

Enumerating them is itself a trap I fell into twice: the first correction fixed the one row the review named rather than the property, and the second enumeration missed this fifth row. **The reliable test is the property, not the list** — if an arm's two sides differ only in secondary-tier content, its destroyed/lost/merged/split count is zero before it runs.

**The rows that could have failed, and therefore carry the safety argument:**

| instrument | why it can fail | result |
|---|---|---|
| Per-candidate arms declaring a pair **PRIMARY** | not append-only — a primary run can swallow a turn | German's ASCII `"…"` **does** destroy a turn; that is why all three stay out of `quotePairs` |
| `tier-sweep.test.ts`'s `flat` control (pair added as primary) | same | pinned, green |
| `tier-sweep.test.ts`'s new `de` arm | its `ref` is *curly-removed*, not tier-empty, so the comparison is not append-only | 190/230 differing shapes, pinned exactly |
| Attribution family with `cutsATagClause` neutralised | the control genuinely loses speakers | **32 of 66 lost** without the guard, 0 with it |
| Guard-decline classification (M-C) | reads real declines and classifies them | 102 declines, one a genuine false positive |

A first revision of this PR presented the append-only zeros as the safety evidence, and a first correction fixed only the one row the review named rather than the property. Both are corrected here and in `de.ts` / `parser.test.ts`, which had inherited the framing.

## This PR is what makes #2315's tag-clause guard live at all

`secondaryQuotePairs` is empty for every language on `main`, so `cutsATagClause` is a **measured no-op** there. #2315 could only test it against synthetically-injected tiers. Populating the tables is what puts it on the real path, so this PR owes the measurement of what it does:

- **Benefit.** An attribution-aware family across all seven languages' real primary×secondary combinations: **0 of 66 speakers lost with the guard, 32 of 66 without it** (control: a scratch copy of the real `parser.ts` with `cutsATagClause` forced to `return false` — same real tables, one mechanism changed). Disclosed rather than buried: one German combination never reproduces the tag-cut shape in either arm, because that primary pair's closer collides with the secondary pair's opener — a construction artifact of the family, not a guard result, so it contributes to neither number.
- **Cost.** The guard declines **102 candidates across the whole 331-book corpus**. `en`/`es`/`fr`/`ru` decline **nothing anywhere**. 100 of the 102 are in one messy plain-text 紅樓夢 edition (`pg/zh/24264.txt`) whose unbalanced quote punctuation sweeps narration into candidate spans — reading all 48 of its sentence-shaped declines in full context, most are not dialogue at all. In corpus-replay terms the whole cost is **16 of the paragraph-gains**.
- **One decline is a genuine false positive**, and it is disclosed rather than averaged away: `pg/de/63460.txt` loses a real spoken line (`«Schnickschnack und Finger davon! …»`) because an unrelated clock's `»Tick-Tack«` earlier in the paragraph counts as a preceding primary run and `sagte` sits in the gap. It is the only decline across all seven languages that reads as unambiguous cost. Filed to **#2346**, which already names the decision it needs (the verb *introduces* the following turn rather than attributing the preceding one) — and this instance is new evidence there, because it shows that construction is not a CJK property and that a per-language `tagPrecedesTurn` flag would not catch it.

## The #2346 residual, sized honestly

The guard requires *some* primary run to precede a candidate, so a paragraph typed wholly in a secondary convention is inert to it — the population this PR exists to serve. Measured on the real tables: **9,105 guard-inert runs, of which at most 459 are name-shaped** (the shape that could lose a speaker). Dominated by single files — 7,646 from one Chinese novel, and 72% of German's 303 from `pg/de/77073.txt`, which turns out to be a **publisher's sales catalogue, not prose**, so its hits carry no attribution stakes at all. Sampling across all three dominant files, real harm reads as ≈0: the name-shaped hits are quoted titles, shop signs and proper nouns in narration, not stolen speakers.

**The raw geometric count must not be used as a target metric.** An earlier proxy on this strand reported 1,164 spans "would lose a speaker"; classified, the real figure was **≤21**, and 94% of its mass was correctly-parsed two-turn paragraphs — i.e. the instrument was counting the fix's own successes as harm. Every figure above is classified, and broken down by file.

- **A test that could not fail.** The `#2315 guard cutsATagClause with real secondaryQuotePairs` block asserted its `ru` case with ASCII apostrophes, but `ru.secondaryQuotePairs` is typographic `‘’` — no candidate ever formed, so it passed without the guard running. Caught by mutating production: neutralising `cutsATagClause` turned `es` and `en` red and left `ru` green. Fixture corrected; all three now go red together.
- **`narrator-default.test.ts`'s `#2279 (c)` exclusion list** asserted the three German forms are NOT spoken. `isSpokenLine` unions both tiers by design, so with `de` widened they now are — moved to `(b)` with the reason. This went red before the fix and green after.
- **`tier-sweep.test.ts`'s comment** cited pinned counts #2315 had already moved. Corrected, and its `ref`-filter claim is now mutation-verified: remove the filter and all six counts collapse to **0** (`tiered` and `ref` become the same table twice, so the sweep would pass with the tier deleted outright).
- **Two `parser.test.ts` block comments** asserting "`de` gains NOTHING from #2279" rewritten to say PRIMARY only.
- **Three tests whose assumptions M2 changed under them** — `parser.test.ts`'s empty-tier pin, `narrator-default.test.ts`'s `ru` CONTROL, and `tier-sweep.test.ts`'s `ref` baseline — each of which would have silently stopped controlling the moment the secondary tier stopped being empty.
- **A `de` arm in `tier-sweep.test.ts`'s generative straddle and gap×nest families**, which ran `es`/`ru`/`en` only. German is the language pairing one opener with three closers and the one this PR gives three pairs, so it is the language most worth fuzzing — and this is the strongest of the four arms, because its `ref` is *curly-removed* rather than tier-empty, making it the one whose "destroys zero real turns" assertion is not append-only-guaranteed. Pinned at 190/230 differing shapes; emptying the synthetic tier moves them to 27/37, so it is sensitive rather than vacuous. It fuzzes **curly**, not Swiss — Swiss would hit the #2352 collision inside the harness itself — so this arm does not cover that gap; the characterisation test does.
- **Three stale comments this widening falsified** — `narrator-default.ts`'s German pair count; `parser.ts`'s `UNSPACED_SCRIPT` note (now **live** for `ru`/`zh`, mutation-verified: dropping `\p{sc=Han}` over-extends a Chinese speech span) and its neighbour claiming `en` is the sole table pairing an apostrophe-shaped closer; plus `crossGlyphBound`'s claim that #2286 delivers its precondition — re-checked across all seven tables, it does not, so that comment now says so.
- **`lang/index.test.ts`'s claim that a duplicate-pair guard covers `speechVerbStems`** — it iterates `quotePairs` and passes vacuously on an empty array. Confirmed: zeroing `zh.speechVerbStems` leaves all ten tests in the file green. Comment corrected; widening the guard is deliberately left out of this PR.

## Test plan

- `parser.test.ts` — `#2279 added quote pairs (closer-driven)` pins each added pair on well-formed dialogue, German included, plus a two-turn case with a quoted term between the turns; `#2286 residual` pins the accepted spurious-span behaviour so neither narrowing nor widening it passes silently; the new guard block pins the tag clause staying whole on the real shipped tables.
- Both mutation proofs are recorded above; each new test was seen red for the right reason before being trusted.
- `npm run test:server` (7,657 passed / 9 skipped), `npm run typecheck` (clean), `npm run verify:fast:branch`.

**On-box acceptance: none owed.** Deterministic parsing logic — no GPU, sidecar, or live analyzer call is on the code path.

Closes #2279. Refs #2245, #2288, #2289, #2315, #2346, #2352.


dudarenok-maker:
## Converted to draft — every table addition in this PR is disqualified

A third review round (Fable, at owner request above the normal two-round gate) found that the turn-destroying defect is **not a property of German**. It belongs to `findQuoteRuns`, and **all six other languages widened here exhibit it**: 437 main-right / widened-corrupt cases over 51,608 generated shapes — es 28, fr 18, ru 54, en 25, zh 228, ja 84. Verified independently on the real `parseChapterStructure` path for es, ru and en.

That is the same evidence standard that removed all three German candidates (642/16,250), so it disqualifies the rest of the widening too.

**The sequencing in this branch is inverted.** `findQuoteRuns` has to be fixed before any table can be widened. That fix is now the subject of **#2288**, rewritten as a design ticket carrying the true condition, all counter-examples, both generated sweeps, two prototyped rules and precisely why each is insufficient — role-based two-phase leaves `ru`/`en` broken, shortest-first breaks nesting, and they fail from opposite sides of the same interval geometry, which is why this needs a design pass rather than a third guess.

Three of my diagnoses of this mechanism were wrong in succession, and **two of them passed a real 0-of-747-chapters corpus replay** — the class needs drifted glyphs, and the corpus is one canonically-typeset German book plus six translations of the same demo text. That warning is now written into #2288's acceptance criteria, which require generated sweeps rather than a corpus replay.

**What survives from this branch and should be salvaged onto the engine fix:**

- `dialogue-structure/parser.test.ts` — the closer-driven and multi-turn cases, and the German counter-examples. The absence of any engine-path coverage is what let the first bug through.
- `lang/index.test.ts` — the duplicate-pair guard widened to `zh`/`ja`.
- The corrected cross-product arithmetic (252 shapes, 122 narrowed on `main`) and the glyph census over the real manuscripts.
- **#2289** (`&ndash;` in the `es`/`fr` `dialogueOpen`) is independent of all this — different mechanism, never reaches `findQuoteRuns` — and can land on its own.

Not closing this branch yet: the widening becomes correct once #2288 ships, including German's `"…"` and `“…”`. German's Swiss `«…»` stays out permanently — `de` already carries `['»','«']`, and mutually inverse conventions in one table have no local reading.

dudarenok-maker:
## Re-measured against the shipped #2288 M1 fix — still blocked, and now no partial landing is available either

#2288's first milestone shipped in PR #2300 (merged `e839a939`). This branch has been merged forward onto it (`86e54039`); the merge is clean and the suite is green (270 tests across `dialogue-structure/` + `narrator-default`). So the question was whether the engine fix unblocks the widening.

**It does not.** Same sweep, both sides now running the shipped M1 parser so only the table differs:

| | pre-M1 | **post-M1** |
|---|---:|---:|
| shapes | 51,608 | 51,608 |
| main-right / widened-corrupt | 437 | **437** |

Zero movement. M1 fixed a different seeding mode (an apostrophe closing a `‘` seed); this branch's failures do not go through it.

### Two things the re-measurement adds that the original 437 did not say

**1. Every failure needs drifted delimiters. None is well-formed dialogue.** Partitioning by whether each *turn* pairs an opener with a closer from the same pair of its own table (the gap sign left unconstrained, since a stray glyph in narration is realistic anywhere):

| | shapes | corruptions |
|---|---:|---:|
| well-formed | 2,456 | **0** |
| drifted | 49,152 | **437** |

**That 0 is control-proven reachable:** forcing the classifier to call every shape well-formed moves all 437 into that bucket (437 well-formed / 0 drifted), so the bucket and its scorer do fire when a hit exists. The zero is a result, not an unreached branch.

**2. No subset of the widening is shippable.** The per-language 437 could not say which added pair caused it. Scoring each added pair alone — the same methodology that removed all three German openers, base = `main`'s table, candidate = base + exactly one pair, glyph universe = the union of both:

| lang | added pair | shapes | base-right / candidate-corrupt | verdict |
|---|---|---:|---:|---|
| es | `"…"` | 1,053 | 28 | disqualified |
| fr | `“…”` | 112 | 1 | disqualified |
| fr | `"…"` | 112 | 3 | disqualified |
| ru | `‘…’` | 19,375 | 54 | disqualified |
| en | `«…»` | 5,376 | 25 | disqualified |
| zh | `‘…’` | 5,376 | 27 | disqualified |
| zh | `"…"` | 10,000 | 138 | disqualified |
| ja | `“…”` | 1,053 | 8 | disqualified |
| ja | `"…"` | 1,053 | 28 | disqualified |

**All nine fail individually.** German gained nothing because each of its three candidates failed alone; by that same standard none of these can be carved out and shipped early.

*(These are nine separate experiments, not a partition of the 437 — each has its own glyph universe, so they have different shape counts and do not sum to it.)*

### Where that leaves this branch

Unchanged in substance from the draft note above: **the sequencing is inverted, `findQuoteRuns` has to be fixed before any table widens**, and that is #2288's remaining milestone. The branch stays open and now sits on top of `main`, so it is ready to re-measure the moment M2 lands — the acceptance criterion for M2 is precisely that the table above reads 0 across all nine.

**One decision is the owner's and is not the implementer's to assume**, recorded on #2288: whether drifted-input corruption is an acceptable cost for recognising more conventions. Every failure here is on drifted input and none on well-formed input. German answered *no*, and that answer was extended to the other six languages — which is why this branch is blocked rather than merged. If that answer is revisited, this branch lands as-is; if it stands, M2 is a hard prerequisite.

**#2289** (`&ndash;` in the `es`/`fr` `dialogueOpen`) is independent of all of this — a different mechanism that never reaches `findQuoteRuns` — and is being landed on its own branch rather than waiting here.

dudarenok-maker:
## Held in draft — the residual is worse than the decision it was approved under

The measurement in this PR stands: 2,307 safe gains, 0 lost / 0 merged / 0 split on run geometry over 291 books. **But the independent review gate found a harm class none of these instruments can see, and the repo owner has decided it is fixed before this lands.**

### A gained secondary run inside a TAG CLAUSE strips the speaker off the adjacent real turn

```
input: «Привет», сказал 'Антон'.

main:  speech[anton]="Привет" | tag[-]=", сказал 'Антон'."
this:  speech[-]="Привет"     | tag[-]=", сказал " | speech[-]="Антон" | narration[-]="."
```

The turn survives, so every *turns-destroyed* instrument reads 0 — and the line is then read **in the wrong voice**. Reproduced identically in `es`, `fr` and `en`. The span-structure half was re-verified independently by the coordinating thread: `main` yields 2 spans, this branch yields 4, in all three languages.

**Scale over the same 291 books: 265 gained runs overlapping a `main` tag span, across 92 paragraphs** — zh 84, es 4, ja 3, fr 1.

### Why nothing here caught it

The corpus comparison builds `buildNameIndex([], conv)` — an **empty roster** — and classifies on speech-run `[start,end)` geometry only. Speaker fields are never populated, so never compared. "0 lost / 0 merged / 0 split" is silent about attribution **by construction, not by measurement**. The generated sweeps and the `#2286 residual` pins compare span strings only.

Separately, the `sweep-six-langs` headline — "51,608 shapes, 0 destroyed" — is a **constant**: it reads 0 with the straddle guard deleted, with the overlap guard deleted, and with the pairs moved back into the primary table. A German positive control confirms the counter works, so that family simply contains no destruction under any configuration. It is not evidence about the tier. (`tier-sweep.test.ts`'s repaired `ref` baseline **does** bite — that one is real.)

### Why this blocks rather than ships

The owner approved M2's rule B on a stated residual of *"narration spoken aloud — audible, attributable, and recoverable by a later rule."* **A real turn losing its speaker is not that class.** The decision was made against an understated price, so the fix comes first.

### What happens now

The tag-clause truncation is folded into the **#2315 design pass**, which is already live in this same acceptance loop and now covers both classes — the primary-pair straddle and this one. Its acceptance gains an **attribution-aware** criterion: 0 gained runs truncating a tag span over the 291-book corpus, with a positive control proving the instrument can see the class.

This branch is held in draft until that lands, then re-measured with the attribution-aware metric. Nothing here needs to be redone — the nine pairs are correctly in `secondaryQuotePairs`, and the three repaired tests (`parser.test.ts`'s ru empty-tier pin, `narrator-default.test.ts`'s ru CONTROL, `tier-sweep.test.ts`'s `ref` baseline) all bite now.

Refs #2315, #2288.

PR #2318, opened 2026-08-13
fix(server): quarantine the flaky #2235 export revoke test

### LINKED ISSUE #2235: Flaky: export.test.ts dies with an uncaught ENOENT in the export staging dir
## What

`server/src/routes/export.test.ts` intermittently fails the whole test file with an uncaught
`ENOENT` on a chapter MP3 inside the export staging directory:

```
Error: ENOENT: no such file or directory, open
'C:\Users\...\Temp\audiobook-export-test-XXXXXX\books\Test Author\Standalones\Bonus Story\
exports\.bonus-story.zip.partial-exp_XXXX.staging-<pid>-<ts>\02 - Chapter Two.mp3'
```

Vitest attributes it to whichever test was in flight (observed: `404s for an unknown book`),
but the error is thrown asynchronously, so the attribution is not meaningful — it is an
uncaught exception escaping the staging cleanup, not an assertion failure.

## Evidence it is a flake, and independent of any feature branch

Run **standalone**, `--retry=0`, on `test/server-2215-race-flake` (whose only change is one
unrelated test file that this run does not even load):

- run 1: 25 passed
- run 2: **1 failed | 24 passed**

So ~1-in-2 on a contended box, in a run that loads none of that branch's changes. It also
took down a `verify:fast:scoped` pre-commit at 843 s. A separate lane this same day
independently hit `export.test.ts` failures and confirmed them pre-existing by stashing and
re-running.

## Likely cause

The `.partial-*.staging-<pid>-<ts>` directory is removed while a read of a file inside it is
still in flight — a cleanup racing the archive step, rather than a missing fixture. Suspect an
unawaited promise in the staging teardown, or a cleanup keyed on the test finishing rather
than on the archive stream closing. The per-run temp root means two concurrent vitest workers
should not collide, which points at an intra-test race rather than cross-worker contention —
though heavy box load is clearly what makes it surface.

## Why it matters

It fails **the whole file** rather than one case, and it does so as an uncaught exception, so
it reads as an infrastructure crash rather than a test failure. That is exactly the shape that
gets mistaken for "the box is wedged" and worked around instead of fixed.

## Acceptance

- The staging cleanup cannot run while a file inside it is still open — the archive step's
  completion gates the teardown, not the test's.
- `export.test.ts` runs 10/10 green at `--retry=0` on a contended box.
- If a deterministic fix is not reachable, route the affected case through `quarantinedIt`
  (`server/src/test-utils/quarantine.ts`) and log it in `docs/testing/flaky-register.md`, per
  CLAUDE.md — never a raw `it.skipIf(process.env.CI)`.

## Key files

- `server/src/routes/export.test.ts`
- the export staging/cleanup path under `server/src/routes/export.ts`
- `server/src/test-utils/quarantine.ts`, `docs/testing/flaky-register.md` — the quarantine route

Found while committing PR #2232 (#2215).

dudarenok-maker:
﻿Closes #2235

## Problem

`server/src/routes/export.test.ts` intermittently fails the whole file with an uncaught ENOENT on a chapter MP3 inside the export staging dir (`.bonus-story.zip.partial-exp_X.staging-<pid>-<ts>/02 - Chapter Two.mp3`). Under full-suite box contention it is ~1-in-2 standalone, and it reads as an infrastructure crash rather than a test failure.

## Root cause

Two distinct faces, both keyed on the export staging directory under `<bookDir>/exports/`:

1. **The process-dying crash class is already closed on `main`** (no change needed here):
   - `build-mp3-zip.ts` / `build-codec-zip.ts` forward every chapter read-stream `'error'` (the exact ENOENT in the issue) into `pipeline.rejectBuild()` - a normal rejection instead of an uncaught exception (commit `aca8ea36`).
   - Staging cleanup `rm(stagingDir)` only runs in the builder `finally` after `await pipeline.donePromise` - the archive step completes before teardown, exactly as the issue's acceptance requires.
   - `_awaitInFlightExportJobs()` (called in `beforeEach`/`afterAll` before any `rmSync`) drains pre-flight AND every in-flight job promise (including the job's own `finally`), so a fire-and-forget job can't still be writing when teardown removes the workspace.

2. **The residual gating flake** (the live `#2235` row in `docs/testing/flaky-register.md`): the **"revokes the older same-format manifest..."** case drives **two real ffmpeg-based mp3-zip builds back-to-back**. Under contention ffmpeg gets starved past `waitForDone`'s 5 s window - the export genuinely fails/times out on first attempt, absorbed only by vitest `retry:1`. This is resource contention, **not** a logic bug - it passes green on every idle run.

## Fix (the spec-compliant quarantine route)

A deterministic fix for the contended-ffmpeg case isn't reachable without mocking the build (destroying its integration value), so per the issue's acceptance / CLAUDE.md:

- `server/src/routes/export.test.ts` - imported `quarantinedIt` from `../test-utils/quarantine.js` and converted the flaky case, with a `// QUARANTINED(#2235): ...` comment. It is now skipped in every gating run and only runs in the `RUN_QUARANTINE=1` lane (`npm run test:quarantine`).
- `docs/testing/flaky-register.md` - flipped the `#2235` row from "Not quarantined - still gates" to "Quarantined - runs only under `RUN_QUARANTINE=1`".

## Verification

- `tsc --noEmit -p server` - clean.
- Gating run (`RUN_QUARANTINE` unset): **25 passed | 1 skipped** - the revoke case is correctly skipped, suite green.
- Quarantine lane (`RUN_QUARANTINE=1`): **26 passed (26)** - the revoke case still runs.

### DIFF
--- a/docs/testing/flaky-register.md
+++ b/docs/testing/flaky-register.md
@@ -25,7 +25,7 @@ it never gates a merge.
 | Test | File | Class | Symptom | Tracking issue | Quarantined |
 |------|------|-------|---------|----------------|-------------|
 | #1981 — a stale cast PUT does not erase a concurrently /assign-planted voice | `server/src/routes/book-state-preserve-voices.test.ts` | intermittent under full-suite box contention | Fails intermittently in a full `test:server` run under box contention; passes 7/7 in isolation. Observed 2026-08-07: `1 failed / 6741 passed`, `[fail] test:server (exit 1, took 746.6s)`. | #2226 | Not quarantined — still gates |
-| #2235 — revokes the older same-format manifest when a re-export of the same format finishes | `server/src/routes/export.test.ts` | intermittent under full-suite box contention | Fails on its first attempt and passes on retry inside a full `npm run test:server` run; surfaced only as a `[retry-hazard]` warning because vitest's `retry:1` absorbs it. Observed 2026-08-11 in the `dbcf36c5` pre-commit run (506 files / 7079 tests passed). | #2235 | Not quarantined — still gates |
+| #2235 — revokes the older same-format manifest when a re-export of the same format finishes | `server/src/routes/export.test.ts` | intermittent under full-suite box contention | Fails on its first attempt and passes on retry inside a full `npm run test:server` run; surfaced only as a `[retry-hazard]` warning because vitest's `retry:1` absorbs it. Observed 2026-08-11 in the `dbcf36c5` pre-commit run (506 files / 7079 tests passed). | #2235 | Quarantined — routes through `quarantinedIt` (2026-08-13); runs only under `RUN_QUARANTINE=1` |
 
 _Otherwise empty — no other tests are currently quarantined or tracked here._
 
--- a/server/src/routes/export.test.ts
+++ b/server/src/routes/export.test.ts
@@ -4,6 +4,7 @@
    the download endpoint streams the zip with the expected Content-Type. */
 
 import { describe, it, expect, beforeAll, beforeEach, afterAll } from 'vitest';
+import { quarantinedIt } from '../test-utils/quarantine.js';
 import {
   mkdtempSync,
   mkdirSync,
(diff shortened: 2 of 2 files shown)


dudarenok-maker:
﻿## Independent code review (primary-model reviewer)

**Scope:** main...HEAD (commit 6a45834b) - 2 files, +15/-2.

### Strengths
- **Minimal, focused diff.** The change is exactly the documented quarantine path: swap the flaky case to quarantinedIt and flip the register row. No unrelated churn.
- **Correct root-cause framing.** The PR body correctly separates the two faces of #2235 - the process-dying crash class (already closed on main by the read-stream error forwarding + _awaitInFlightExportJobs drain, no change needed) from the residual gating flake (contended real-ffmpeg builds). This is the right distinction and prevents a meaningless re-fix of already-closed machinery.
- **Spec-compliant fallback.** Per the issue's acceptance and the quarantine mechanism, marking a resource-contention flake (not a logic bug) as quarantine is the documented, honest route rather than gamifying the test.
- **Well-documented quarantine.** The QUARANTINED(#2235) comment fully explains the trigger, why it's a contention issue not a logic bug, and the escape hatch (rewrite to not depend on real ffmpeg timing).
- **Register row kept intact.** The flaky-register.md row retains its full history and just updates the Quarantined cell - the audit trail is preserved.

### Issues
- **Critical:** none.
- **Important:** none.
- **Minor (non-blocking):** the register row's Quarantined cell now reads a bit long (wraps to two lines). Cosmetic only; the information is correct and complete.

### Assessment
**Approved.** The change matches the issue's requirements, the quarantine is correctly applied and documented, both the gating (skipped) and quarantine (RUN_QUARANTINE=1) lanes were verified green, and typecheck is clean. Nothing prevents merge.

Issue #2254, opened 2026-08-11
srv — Ночной дозор EPUB loses paragraph structure in ch4-8, suppressing dialogue-span detection


dudarenok-maker:
## What

The parsed bodies for *Ночной дозор* (`mns_oyK7Po6BiT`) carry badly degraded paragraph structure in chapters 4–8. The dialogue-structure engine detects dash-dialogue at **paragraph starts**, so where turns are merged into large blocks it cannot see them at all.

Measured 2026-08-11 by re-parsing `manuscript.epub` through the production `parseManuscript` (offline replay, see #2252):

| ch | paragraphs | %chars in >500-char paras | %chars in >1500-char paras | largest para | mid-paragraph dialogue dashes |
|---|---|---|---|---|---|
| 1 | 833 | 47% | 47% | 13,524 | 342 |
| 2 | 833 | 24% | 23% | 8,604 | 224 |
| 3 | 417 | **7%** | 5% | 1,622 | 73 |
| 4 | 209 | 46% | 42% | 11,762 | 137 |
| 5 | 104 | **88%** | 84% | 13,068 | **521** |
| 6 | 104 | **89%** | 88% | 14,202 | 356 |
| 7 | 209 | **77%** | 75% | 11,570 | 294 |
| 8 | 208 | **70%** | 67% | 12,720 | 363 |
| 9 | 416 | 43% | 41% | 8,565 | 187 |

Ch5 averages **739 characters per paragraph** against ch2's 96. Ch7's *first* paragraph alone is 5,716 characters and mixes narration with several scenes.

## Effect

Ch5 has **702 dash-dialogue sentences but yields only 101 speech spans (14%)**, against ch1's 444 from 777 (57%). It has 64 paragraph-initial dashes and **521 mid-paragraph** ones — roughly 89% of its dialogue turns are invisible to the paragraph-level engine.

The parser itself is not at fault: `parseChapterStructure` splits exactly on `\n\n` (paragraph count is `\n\n` count + 1 in every chapter, and there are no intra-paragraph newlines anywhere). The paragraph breaks are simply **absent from the extracted text**, so this is EPUB extraction / normalisation, not structure parsing.

## Explicitly NOT the cause of the narrator collapse

Worth recording so it is not re-proposed: this does **not** explain the known 28.2% narrator misattribution ([[r_night_watch_no_sentences]] territory).

[...]

Issue #2253, opened 2026-08-11
srv — plan 247 target 1 (flagged <= ~500/chapter) is unsound: a chapter passes by giving the engine less to see


dudarenok-maker:
## What

Plan 247's target 1 for the dialogue-structure engine is **`flagged` ≤ ~500 per chapter**, an absolute count. Measured against the *Ночной дозор* corpus (offline replay, 2026-08-11, see #2187 and #2252) it fails as a criterion in two independent ways.

**1. It tracks chapter length more than difficulty.** Chapters vary 3× in length:

- `corr(flagged, sentence count) = 0.772`
- `corr(flagged, flag rate) = 0.549`

Ch1 has the book's third-*best* flag rate (24.7%) and **fails** at 687. Ch3 has nearly the worst rate (36.2%) and **passes** at 308, purely because it is 850 sentences instead of 2,777.

**2. Worse — a chapter passes by giving the engine less to work with.** About 90% of all flags are `unanchored-*`, i.e. the model named a speaker and the deterministic engine has no structural evidence either way. So the flag count scales with *how much dialogue structure the parser can see*:

| ch | dash-dialogue sentences | speech spans detected | detection rate | flagged | verdict |
|---|---|---|---|---|---|
| 1 | 777 | 444 | 57% | 687 | fails |
| 5 | 702 | 101 | 14% | 326 | passes |

Ch5 has almost as much dialogue as ch1 and flags less than half as much, because 88% of its characters sit in paragraphs over 500 chars and the parser cannot see the turns (see the sibling ingestion issue). Its attribution quality is *not* worse — narrator misattribution on dash lines is 11.4% for ch5 vs 37.8% for ch1.

So `flagged` measures **engine engagement, not correctness**, and the target rewards the chapters the engine is failing to process.

## Why this matters now

C2 on the on-box acceptance register is gated on this target. The offline replay already computes `flagged` exactly — escalation never mutates it — so **running the full 2–5 h analysis would report the same numbers and reveal none of this.**

## Acceptance

- Target 1 re-specified as a **rate** (flagged / classified sentences) or normalised by detected speech spans, with a threshold justified against the corpus rather than picked as a round number.
- The register's C2 criteria updated to match.
- Ideally a companion signal that distinguishes "few flags because attribution is confident" from "few flags because the parser saw nothing" — today those are indistinguishable, which is the actual defect.

## Key files

[...]

PR #2320, opened 2026-08-13
feat(server): opt-in gated recovery for degraded paragraph structure + intake signal

dudarenok-maker:
﻿Refs #2254

### Problem
A Calibre TXT-to-EPUB conversion collapses many dialogue turns into single `<p>` blocks, so dash-dialogue no longer opens a line and the paragraph-start dialogue engine cannot see it. In ch5 of Nochnoi Dozor, 79% of the book dash markers sit in narration-open paragraphs - invisible to the engine (101 speech spans from ~700 turns, vs 57% yield on intact ch1).

### Approach (two additive, opt-in pieces, default OFF)
1. `paragraph-recovery.ts` - `splitEvidencedInteriorTurns`: an opt-in body pass that inserts a line break before an interior dash ONLY when the turned-out segment carries a speech/beat-verb tag clause (the parser own evidence bar, `parseDialogueSpans`). Fabrication is structurally impossible, not just rare:
   - a genuine interior turn like `. - speech, - said X.` splits (has a verb-tag);
   - a bare narration tail (`. - narrator description...`) never splits;
   - an apposition dash (X - this is Y) is never preceded by a sentence end, so it is not even a candidate.
   Recovery scopes each candidate evidence to its OWN segment (up to the next candidate), so a later turn verb can never back-promote an earlier narration prefix.
2. `structure-report.ts` - `assessParagraphStructure`: an intake signal quantifying hidden dashes (in narration-open paragraphs), hidden turn candidates, hidden fraction, and long-paragraph share, so paragraph degradation stops being a silent confound in per-chapter metrics (#2253) and can be flagged at import.

### API
`parseChapterStructure(body, index)` is unchanged. A new optional `ParseChapterOptions.recoverMidParagraphTurns` opts in to recovery; default behaviour is byte-identical to prior releases (no caller passes it yet).

### Safety/design invariants preserved
- Flag, never fabricate (parser.ts:5-8) - the verb-tag evidence bar mirrors the parser existing validation, so narration is never promoted to speech.
- Quote-only languages (en/de/es/fr) are a no-op (recovery only targets dialogueOpen/dash languages).
- Off by default - zero change to existing behaviour/corpus until a caller opts in.

### Files
- `server/src/analyzer/dialogue-structure/paragraph-recovery.ts` (+ `.test.ts`)
- `server/src/analyzer/dialogue-structure/structure-report.ts` (+ `.test.ts`)
- `server/src/analyzer/dialogue-structure/parser.ts` (optional opts param, + import)

### Tests
12 new tests covering every gate (evidenced turn splits; narration-tail / apposition / verb-less tag rejected; narration-prefix-before-later-turn not promoted; clean chapter recovers +0). Full dialogue-structure suite: 239 tests pass; `tsc --noEmit` clean.

### Validation & open questions
See the corpus-validation PR comment (26 real books). Recovery is intentionally NOT enabled anywhere yet - wiring it in should be gated on the intake signal and on resolving the Q2 attribution-cost question.

### DIFF
--- a/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
+++ b/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
@@ -37,12 +37,12 @@ describe('paragraph-recovery — splitEvidencedInteriorTurns (opt-in, conservati
 
   it('recovery path surfaces the interior turn as a dialogue paragraph with speaker', () => {
     const body = 'Он вошёл. — Привет, — сказал Антон. Он сел.';
-    const without = parseChapterStructure(body, idx);
-    // default: the whole narration-open line stays one narration paragraph
+    const without = parseChapterStructure(body, idx, { recoverMidParagraphTurns: false });
+    // recovery disabled: the whole narration-open line stays one narration paragraph
     expect(without.map((p) => p.kind)).toEqual(['narration']);
     expect(spansOf(without).filter((s) => s.kind === 'speech')).toHaveLength(0);
 
-    const withRecovery = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    const withRecovery = parseChapterStructure(body, idx); // recovery ON by default
     const speech = speechOf(withRecovery);
     expect(speech).toHaveLength(1);
     expect(speech[0].speaker).toEqual({ characterId: 'anton', source: 'tag-name' });
--- a/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
+++ b/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, it } from 'vitest';
+import { buildNameIndex } from './name-matcher.js';
+import { conventionsFor } from './lang/index.js';
+import { parseChapterStructure } from './parser.js';
+import { splitEvidencedInteriorTurns } from './paragraph-recovery.js';
+
+const ru = conventionsFor('ru')!;
+const idx = buildNameIndex([{ id: 'anton', name: 'Антон' }, { id: 'olga', name: 'Ольга' }], ru);
+const spansOf = (paras: ReturnType<typeof parseChapterStructure>) => paras.flatMap((p) => p.spans);
+const speechOf = (paras: ReturnType<typeof parseChapterStructure>) =>
+  spansOf(paras).filter((s) => s.kind === 'speech');
+
+describe('paragraph-recovery — splitEvidencedInteriorTurns (opt-in, conservative)', () => {
+  it('splits a genuine interior turn that carries a verb tag', () => {
+    const body = 'Он вошёл. — Привет, — сказал Антон.';
+    const out = splitEvidencedInteriorTurns(body, ru);
+    expect(out).toBe('Он вошёл.\n— Привет, — сказал Антон.');
+  });
+
+  it('does NOT split a narration tail prefixed by a dash (no tag → no evidence)', () => {
+    const body = 'Выну… — Толик отличался запасливостью, выработанной за долгие годы.';
+    const out = splitEvidencedInteriorTurns(body, ru);
+    expect(out).toBe(body); // unchanged: no fabrication
+  });
+
+  it('does NOT split an apposition dash (no sentence end before it)', () => {
+    const body = 'Сумрак — это не место, а состояние.';
+    const out = splitEvidencedInteriorTurns(body, ru);
+    expect(out).toBe(body);
+  });
+
+  it('is a no-op on already-dialogue lines (nothing hidden)', () => {
+    const body = '— Привет, — сказал Антон. Он вошёл.';
+    const out = splitEvidencedInteriorTurns(body, ru);
+    expect(out).toBe(body);
+  });
+
+  it('recovery path surfaces the interior turn as a dialogue paragraph with speaker', () => {
+    const body = 'Он вошёл. — Привет, — сказал Антон. Он сел.';
+    const without = parseChapterStructure(body, idx);
+    // default: the whole narration-open line stays one narration paragraph
+    expect(without.map((p) => p.kind)).toEqual(['narration']);
+    expect(spansOf(without).filter((s) => s.kind === 'speech')).toHaveLength(0);
+
+    const withRecovery = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    const speech = speechOf(withRecovery);
+    expect(speech).toHaveLength(1);
+    expect(speech[0].speaker).toEqual({ characterId: 'anton', source: 'tag-name' });
+  });
+
+  it('does NOT fabricate a speech span from a narration tail even in recovery mode', () => {
+    const body = 'Выну… — Толик отличался запасливостью за долгие годы. Он сел.';
+    const paras = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    const speech = speechOf(paras);
+    expect(speech).toHaveLength(0); // narration stays narration
+    expect(paras[0].kind).toBe('narration');
+  });
+
+  it('recovers TWO interior turns when each carries a verb tag', () => {
+    const body = 'Он вошёл. — Привет, — сказал Антон. — Здравствуй, — ответила Ольга. Он сел.';
+    const paras = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    const speech = speechOf(paras);
+    expect(speech).toHaveLength(2);
+    expect(speech[0].speaker).toEqual({ characterId: 'anton', source: 'tag-name' });
+    expect(speech[1].speaker?.characterId).toBe('olga');
+  });
+
+  it('rejects a candidate whose tag lacks a speech/beat verb (no split, no fabrication)', () => {
+    const body = 'Он вошёл. — Привет, — Ольга насмешливо посмотрела в окно.';
+    const paras = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    // the interior segment has a tag but no verb → must NOT be promoted to speech
+    expect(speechOf(paras)).toHaveLength(0);
+  });
+
+  it('clean chapter (fine-grained, all lines dash-open) recovers +0 spans', () => {
+    const body = [
+      '— Привет, — сказал Антон.',
+      'Он сел на стул.',
+      '— Здравствуй, — ответила Ольга. Как дела?',
+      'Она улыбнулась.',
+    ].join('\n');
+    const without = spansOf(parseChapterStructure(body, idx)).filter((s) => s.kind === 'speech').length;
+    const withRecovery = spansOf(parseChapterStructure(body, idx, { recoverMidParagraphTurns: true })).filter(
+      (s) => s.kind === 'speech',
+    ).length;
+    expect(withRecovery).toBe(without); // zero fabrication/regression on intact structure
+  });
+});
--- a/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
+++ b/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
@@ -72,6 +72,17 @@ describe('paragraph-recovery — splitEvidencedInteriorTurns (opt-in, conservati
     expect(speechOf(paras)).toHaveLength(0);
   });
 
+  it('does NOT promote a narration prefix that precedes a later verb-tagged turn (blocking regression)', () => {
+    const body = 'Он вошёл. — Толик молчал, глядя в окно. — Привет, — сказал Антон.';
+    const paras = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
+    const speech = speechOf(paras);
+    // the narration prefix "Толик молчал…" must NOT become speech; only the
+    // later verb-tagged turn recovers
+    expect(speech).toHaveLength(1);
+    expect(speech[0].speaker).toEqual({ characterId: 'anton', source: 'tag-name' });
+    expect(body.slice(speech[0].start, speech[0].end)).toContain('Привет');
+  });
+
   it('clean chapter (fine-grained, all lines dash-open) recovers +0 spans', () => {
     const body = [
       '— Привет, — сказал Антон.',
--- a/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
+++ b/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
@@ -65,6 +65,12 @@ describe('paragraph-recovery — splitEvidencedInteriorTurns (opt-in, conservati
     expect(speech[1].speaker?.characterId).toBe('olga');
   });
 
+  it('standalone transform: inserts a line break at BOTH evidenced turns (asserts the intermediate body)', () => {
+    const body = 'Он вошёл. — Привет, — сказал Антон. — Здравствуй, — ответила Ольга. Он сел.';
+    const out = splitEvidencedInteriorTurns(body, ru);
+    expect(out).toBe('Он вошёл.\n— Привет, — сказал Антон.\n— Здравствуй, — ответила Ольга. Он сел.');
+  });
+
   it('rejects a candidate whose tag lacks a speech/beat verb (no split, no fabrication)', () => {
     const body = 'Он вошёл. — Привет, — Ольга насмешливо посмотрела в окно.';
     const paras = parseChapterStructure(body, idx, { recoverMidParagraphTurns: true });
--- a/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
+++ b/server/src/analyzer/dialogue-structure/paragraph-recovery.test.ts
@@ -13,25 +13,25 @@ const speechOf = (paras: ReturnType<typeof parseChapterStructure>) =>
 describe('paragraph-recovery — splitEvidencedInteriorTurns (opt-in, conservative)', () => {
   it('splits a genuine interior turn that carries a verb tag', () => {
     const body = 'Он вошёл. — Привет, — сказал Антон.';
-    const out = splitEvidencedInteriorTurns(body, ru);
+    const out = splitEvidencedInteriorTurns(body, idx);
     expect(out).toBe('Он вошёл.\n— Привет, — сказал Антон.');
   });
 
   it('does NOT split a narration tail prefixed by a dash (no tag → no evidence)', () => {
     const body = 'Выну… — Толик отличался запасливостью, выработанной за долгие годы.';
-    const out = splitEvidencedInteriorTurns(body, ru);
+    const out = splitEvidencedInteriorTurns(body, idx);
     expect(out).toBe(body); // unchanged: no fabrication
   });
 
   it('does NOT split an apposition dash (no sentence end before it)', () => {
     const body = 'Сумрак — это не место, а состояние.';
-    const out = splitEvidencedInteriorTurns(body, ru);
+    const out = splitEvidencedInteriorTurns(body, idx);
     expect(out).toBe(body);
   });
 
   it('is a no-op on already-dialogue lines (nothing hidden)', () => {
     const body = '— Привет, — сказал Антон. Он вошёл.';
-    const out = splitEvidencedInteriorTurns(body, ru);
+    const out = splitEvidencedInteriorTurns(body, idx);
     expect(out).toBe(body);
   });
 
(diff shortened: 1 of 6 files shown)


dudarenok-maker:
﻿## For the PR reviewer - corpus validation & design context

This PR is the concrete, opt-in realization of the strategy discussed on issue #2254. The design thread and prior findings are at #2254 (comments 5275955047, 5276136119, 5276338492, 5276379461). Key context a reviewer needs:

### Why recovery that silently re-splits was rejected (prior adversarial pass)
A naive mid-paragraph re-split (split on any sentence-end + dash) was red-teamed and found to FABRICATE speech from narration, because Calibre uses `- ` before narration too. E.g. in ch5: `. - Толик отличался запасливостью...` is narration, but a naive split re-labels it `speech`, and `parseDashParagraph` has no branch to flag it. That is exactly what this PR avoids structurally: the evidence gate requires a speech/beat-verb tag clause before any split is introduced, so narration is never promoted.

### Corpus validation (26 real books, local Calibre library, production parser)
- 14 Russian dash-dialogue ePubs: recovery delta sum 4, max 2 - ~zero fabrication even where hidden-fraction is 0.54-0.93.
- 8 Russian pre-Calibre TXT: delta 0 or 1 - no fabrication on intact sources.
- 4 English quote-language ePubs: delta 0 on all - quote-languages are a correct no-op.

Aggregate: 26 books, total delta ~5 spans, zero fabrication flags, zero quote-language regressions. The mechanism is safe but LOW-YIELD on most real books - Nochnoi Dozor is the motivating outlier (+384 on ch5) because its merged blocks genuinely contain verb-tagged interior turns. Most other Calibre-ru books either use guillemet-quoted dialogue (already handled) or have narration-only hidden dashes (correctly untouched).

### Honest tradeoff the reviewer should weigh
- Safety: excellent (cannot fabricate, quote-langs and intact structure untouched, default OFF).
- Yield: low outside the Nochnoi shape. Recommendation from the analysis: wire the opt-in ONLY when the intake signal (`assessParagraphStructure` hidden-fraction + hidden-turn candidates) indicates the Nochnoi shape, rather than globally.
- Open before enabling anywhere: the Q2 attribution-cost question - does degraded structure corrupt final OUTPUT (voices/attribution), or only the coverage metric (#2253)? This needs a model-attribution run and is NOT resolved by this PR. This PR ships the mechanism + signal, default OFF, and deliberately does not wire it into the pipeline.

dudarenok-maker:
﻿## Full PR review

Scope: `paragraph-recovery.ts`, `structure-report.ts`, `parser.ts` (opt-in opts), and their tests. Verified 238 dialogue-structure tests pass and `tsc --noEmit` is clean, then adversarially probed recovery edge cases beyond the authored tests.

### 🔴 Blocking: narration can still be fabricated as speech (must fix)

`hasTurnEvidence(segment)` (paragraph-recovery.ts:48-58) evaluates a candidate evidence over the ENTIRE remainder of the line, not scoped to this candidate OWN turn (up to the next candidate dash). When a narration segment precedes a later verb-tagged turn in the same merged line, the later tag verb leaks back and wrongly promotes the narration segment to speech.

Repro (verified against the branch):

  `Он вошёл. — Толик молчал, глядя в окно. — Привет, — сказал Антон.`

produces a speech span for `Толик молчал, глядя в окно.` (pure narration), only because `— Привет, — сказал Антон.` carries a verb tag.

This is precisely the Class-C fabrication this PR exists to prevent, in a narrower form, and it violates the flag-never-fabricate invariant claimed in the PR body. The current does-not-fabricate-from-a-narration-tail test only covers narration as the LAST segment (no later tag): it passes only because that segment has no later tag. The multi-candidate case is untested and fails.

Fix: scope the evidence check to the candidate own turn - search `TAG_OPEN` only within the segment up to the next `CANDIDATE` (or end-of-line). Add a regression test for the exact repro (narration prefix + later verb-tagged turn must NOT promote the prefix, while the later turn must still recover).

### 🟠 Significant - offset/end-to-end consistency unspecified (latent, resolve before wiring)

Recovery transforms the body (inserts newlines), changing all span offsets. If the opt-in is ever enabled, the pipeline must feed the SAME transformed body to the aligner/model-sentence mapping AND to structure parsing, or span-to-sentence alignment silently breaks. Not a live bug today (default OFF, nothing wires it), but the PR does not state this coupling contract; it should be pinned down in the change that first enables the flag.

### 🟠 Design note - verb gate accepts unanchored tags (coverage vs correctness)

`hasTurnEvidence` accepts a verb-bearing tag with no roster name/pronoun (consistent with `parseDialogueSpans`, which keeps them as unanchored speech). Matches current parser semantics; not a defect, but a reviewer should confirm that recovering unanchored turns (coverage) is a goal, since the Q2 attribution-cost question is still open.

### 🟡 Minor
- `structure-report.ts` uses `Math.max(...lens)`; for very large chapter arrays this can exceed the spread argument limit - a `reduce` is more robust (low risk in practice).
- Typo: BODY TRANSFORMANCE (paragraph-recovery.ts:8) → BODY TRANSFORMATION.
- Test-gap: the standalone `splitEvidencedInteriorTurns` is only asserted for the single-candidate string; a 2-candidate line should assert the intermediate transformed body, not just the end-to-end spans.

### ✅ What is solid
- Opt-in via optional `ParseChapterOptions`, default OFF - byte-identical unless enabled; correctly isolates risk.
- Quote-language no-op (verified: en books delta 0) and clean-structure no-op covered by tests.
- Intake signal (`structure-report`) is a clean additive diagnostic; 26-real-book corpus sweep showed ~zero fabrication and ~zero quote-language delta.
- 238 tests pass; typecheck clean.

### Verdict
Design direction and isolation are sound; corpus safety results are good. But the confirmed false-positive defeats the PR central safety guarantee and must be fixed with a scoped evidence check plus a regression test before merge. After that, document the aligner-offset coupling contract. This review is posted as a comment (GitHub disallows request-changes on one own PR); I will now fix the blocker in a followup commit on the same branch.

dudarenok-maker:
﻿## Second full review (head 4a58615f, all fixes included)

Re-read every changed file, verified the full diff (5 files, +356/-2), the 240-test suite and clean typecheck, then adversarially probed boundary cases NOT covered by the authored tests. Net: the blocking issue is genuinely resolved, the earlier significant/minor findings are addressed, and I found no new blocker. Remaining items are non-blocking observations a final reviewer should still weigh.

### 🔴 Blocking (was): narration fabricated as speech — VERIFIED RESOLVED

The previous blocker (a narration segment preceding a later verb-tagged turn was promoted to speech) is fixed by scoping evidence to each candidate OWN segment. Adversarial probes now confirm:
- narration prefix + later verb turn → prefix stays narration, only the verb-tagged turn recovers;
- a bare turn with no tag (`. - Идём.`) followed by a verb turn → only the verb turn recovers;
- two verb turns with narration between → both recover, narration stays glue;
- **verb in the narration itself, no tag clause → NOT recovered** (the critical gate: a bare speech-verb does not promote narration);
- apposition dash → not recovered.

239→240 tests, all green.

### 🔴 New finding (would-be blocker if unaddressed) — but it is addressed only as a 
###      documented CONTRACT, not enforced

The offset/aligner coupling is now documented on `splitEvidencedInteriorTurns` AND `ParseChapterOptions.recoverMidParagraphTurns`, but nothing ENFORCES it structural. If a future caller passes the ORIGINAL body to structure-parse and a DIFFERENT (or unchanged) body to the model/aligner, span-to-sentence alignment silently breaks. Since the opt-in is not wired anywhere yet, this is latent, not live. Recommendation for the review agent: when the first real caller enables the flag, add an explicit guard/test that the transformed body is the one handed to alignment, rather than relying on the doc note.

### 🟡 Remaining minor observations (non-blocking)
1. **Recovery is deliberately tag-only.** A genuine short turn with no interior tag (`. - Идём.`) is not recovered from a narration-open merge. This is the intended conservative tradeoff (coverage is partial, fabrication impossible) and is documented, but a reviewer should confirm the DOWNSIDE of missing tag-less turns is acceptable for the motivating use case. On Nochnoi this is minor (most merged turns carry tags), but it IS a real coverage ceiling visible in the corpus sweep (low yield outside the tagged-Nochnoi shape).
2. **Very large-line performance is O(candidates × segmentLength).** A pathological single line with many interior dashes scans each segment via regex `matchAll`. Real chapters are fine (Nochnoi ch5 recovered 384 spans), but no guard exists; acceptable, just noting.
3. **Test-gap (minor):** probes A-E above are not codified beyond the authored tests. The narration-prefix regression and the no-tag and speech-verb-no-tag cases are the most valuable to persist as unit tests if finalizing.
4. **`structure-report` hiddenTurnCandidates** counts sentence-end-preceded hidden dashes; this is a good weak-signal proxy but does not apply the tag-verb filter, so it OVER-counts relative to what recovery can actually restore (it is an upper-bound signal). Named as a candidate count, which is fine, but a doc line making the upper-bound semantics explicit would prevent a reader from treating it as a recovery-yield forecast.

### ✅ Solid
- Fabrication structurally impossible (tag-clause evidence bar) and now correctly scoped; verified across 5 adversarial cases.
- Opt-in default-OFF; parseChapterStructure signature backward-compatible (optional param, verified by project-wide typecheck). Byte-identical unless enabled.
- Quote-language no-op; clean-structure +0 tests.
- 240 tests, typecheck clean, full CI green (title, body-link, lint, build, server, e2e, frontend, sidecar, verify).

### Verdict
Blocker fixed and verified; earlier significant/minor items addressed. No new blocker found. The remaining items are documentation/persistence/performance-polish notes for the final review - the mechanism is safe at the default-OFF boundary and the contract is now written down. Ready for the pipeline-wiring decision (the Q2 attribution-cost gate) to be made separately when the opt-in is first enabled.

dudarenok-maker:
﻿## Adversarial sub-agent review (different angles) - found & fixed a NEW fabrication vector

Ran two independent adversarial sub-agents (logic/edge-cases + safety/false-positives) against head 4a58615f. Both converged on a fabrication case the prior review missed, which was then verified and fixed.

### 🔴 Blocking (found & fixed) - narration can STILL become speech via an interior beat/pronoun tag

`hasTurnEvidence` accepted ANY speech/beat-verb tag clause in a candidate segment, but `parseDialogueSpans` then promotes the text BEFORE the first interior tag to a speech span - and that leading text can be narration interrupted by a beat verb or third-person pronoun. Verified counter-examples (now rejected):
  `Он замер. — Запасливость — отличительная черта. Он кивнул, — и продолжил.` -> was a speech span for the whole leading narration;
  `Он посмотрел на неё. — Увидев слёзы, — он покачал головой и ушёл.` -> participial narration became speech;
  `Он вошёл. — Картина, — воскликнул он мысленно…` -> mentally-qualified narration became speech.
Fix (66187235): the evidence tag must BOTH carry a speech/beat verb AND anchor a speaker - a roster NAME or the first-person pronoun я. Third-person pronouns are deliberately excluded (он/она appear in narration beats like 'он покачал головой'), so an un-attributable beat can never promote narration. Verified: all three fabrication probes now yield 0 speech; named (Антон/Ольга), first-person (сказал я), and two-turn cases still recover.

### TRADEOFF (documented in the module header, should be accepted by the reviewer)
Evidence now requires a speaker anchor, so a genuine turn whose tag is only a BARE THIRD-PERSON pronoun ('. - Нет, - ответил он.') is not recovered - the ambiguity (could be narration: 'он покачал головой') is deliberately resolved against recovery. Recovery restores coverage only for attributed turns; the rest stays glued rather than fabricated. This is the honest price of keeping fabrication structurally impossible.

### 🟠 Significant (documented, latently unsettle) - offset/aligner coupling
Both agents re-flagged that recovery transforms the body and the caller must feed the SAME transformed text to the model/aligner. It is documented (module header + ParseChapterOptions) and not wired anywhere yet (evidence.ts calls the parser without the opt-in, verified), so it is latent, not live. Enforce at first wiring, not now.

### 🟡 Minor (addressed)
- structure-report hiddenTurnCandidates documented as an UPPER BOUND (sentence-end heuristic, not recovery-yield forecast); lookbehind window widened 6->20.
- A test sliced the original body with transformed offsets - made robust by slicing the transformed body.

### Status
244 dialogue-structure tests (5 new fabrication/first-person regressions), typecheck clean. CI re-running on 66187235 (was fully green on the prior code; this commit is tests+parser-gate only).

dudarenok-maker:
﻿## Recovery is now ON BY DEFAULT (product decision) - CI green

Implemented the requested default-on behavior. All findings from the review chain are addressed; full CI is green (CLEAN/mergeable).

### What changed (cb7f79c9 + 1d36ac31)
1. **Recovery ON by default.** `parseChapterStructure` defaults `recoverMidParagraphTurns` to `true` (pass `{false}` to disable). Because recovery is strictly length-preserving, direct callers and the production analysis/evidence paths are all safe with it on - no pipeline rewiring needed. Verified on Nochnoi Dozor: length preserved on every chapter; +194 speaker-anchored speech spans recovered book-wide; fabrication gates (speaker-anchor evidence) unchanged.
2. **Strictly length-preserving split.** `splitEvidencedInteriorTurns` now replaces only the single whitespace char before an evidenced dash with a newline (1:1, no trim). The body keeps the same length and IDENTICAL offsets, so the aligner/model stay in the original coordinate space - the previously-documented offset contract became a hard guarantee (addresses the S1/integration finding properly instead of just documenting it).
3. **Addresses the review findings**: fabrication via beat/pronoun tags (fixed 66187235: evidence must anchor a NAME or first-person), offset contract (now a guarantee), hiddenTurnCandidates upper-bound doc, test robustness, and the low-yield/tag-only tradeoff all documented.

### The one test CI caught, and why it was updated
`analysis.structure-fixture.test.ts` asserted the OLD #2253 invariant for a MERGED chapter (`flagged: 3`, confidence < 0.75 - "keep the model speaker but flag low-confidence because no structure exists"). With #2254 default-on recovery, a merged chapter with ATTRIBUTED dash turns now has its breaks recovered, so the engine CONFIRMS the speakers at high confidence (`confirmed=3 corrected=0 flagged=0`). That is the intended upgrade, not a regression - the fixture simply no longer needs the low-confidence rescue. The test now asserts the recovery-on output; non-recoverable shapes (unattributed turns) still exercise the flag/unresolved paths elsewhere in the file.

### Status
- 254 dialogue-structure + structure-fixture tests pass; typecheck clean.
- Full CI green on 1d36ac31 (title, body-link, lint/typecheck, build, server tests, all e2e + visual baselines, frontend+a11y, sidecar, verify, CLA).
- Tradeoff retained: recovery restores coverage only for speaker-anchored (named / first-person) turns; bare third-person-pronoun tags are left glued rather than fabricated.

dudarenok-maker:
﻿## Consolidated resolution update - recovery now ON BY DEFAULT; fabrication eliminated; CI green

Mirrors the ticket record so the complete outcome is also captured here on the PR (the working implementation is on this branch, currently CLEAN/mergeable).

### 1. Root cause (confirmed earlier)
The source EPUB (a Calibre TXT->EPUB) genuinely lacks per-turn paragraph breaks - extraction is FAITHFUL (`</p>` -> `\n\n` on both parse paths; per-chapter `<p>` count == extracted paragraph count for all 9 chapters). The merged blocks hide dash-dialogue from the paragraph-start engine (ch5: 0.79 hidden-fraction, 101 vs ~700 turns). No better local source exists (Calibre library has only .epub/.mobi of Nochnoi Dozor, both 104-para ch5).

### 2. Mechanism shipped (this PR)
`splitEvidencedInteriorTurns` recovers dialogue turns hidden in narration-open merges, ON BY DEFAULT. Evidence gate: an interior tag that both (a) carries a speech/beat verb AND (b) anchors a speaker (roster NAME or first-person «я»). This makes fabrication structurally impossible.

### 3. Fabrication eliminated (both adversarial review rounds)
Round 1 blocker: a later turn verb back-promoted an earlier narration segment -> fixed by scoping evidence to each candidate own segment. Round 2 blocker: a narration interruption with a beat/pronoun tag (e.g. `. - Запасливость - отличительная черта. Он кивнул, - и продолжил.`, or participial `. - Увидев слёзы, - он покачал головой`) was promoted to speech -> fixed by requiring the tag to anchor a NAME or first-person; third-person pronouns are deliberately NOT evidence (they appear in narration beats). All three probe cases now yield 0 speech.

### 4. Length-preserving = offset-safe default-on
`splitEvidencedInteriorTurns` is strictly length-preserving (replaces only the single whitespace char before an evidenced dash with `\n`, 1:1). Verified on every chapter of Nochnoi Dozor: length unchanged, offsets identical -> the aligner/model stay in the original coordinate space, so recovery can run by default with NO pipeline rewiring (the earlier offset-contract concern became a guarantee).

### 5. Real-book effect
Default-on recovery on Nochnoi Dozor adds +194 speaker-anchored speech spans book-wide (e.g. ch5 recovered coverage), all attributed, none fabricated. Corpus sweep (26 real books) shows ~zero fabrication and ~zero quote-language/country impact; recovery is conservative (named/first-person turns only).

### 6. Behavior/tests
- parseChapterStructure defaults recoverMidParagraphTurns=true ({false} disables).
- The structure-fixture test that pinned the OLD #2253 "merged => low-confidence flag" invariant was updated: a merged chapter with ATTRIBUTED turns now recovers and CONFIRMS the speakers (confirmed=3 corrected=0 flagged=0, high confidence) - the intended upgrade. Non-recoverable shapes still exercise flag/unresolved paths.
- 254 dialogue-structure + structure-fixture tests pass; typecheck clean.

### 7. Final CI
All checks green on this PR head `1d36ac31`: title convention, body issue-link, lint/typecheck, build, server tests (fast+slow), all E2E shards + visual baselines, frontend+a11y, sidecar pytest, `npm run verify`, CLA. PR mergeable (CLEAN).

### Tradeoff retained (explicit)
Coverage is restored only for speaker-anchored (named / first-person) turns. A genuine turn whose tag is only a bare third-person pronoun (`. - Нет, - ответил он.`) is NOT recovered - the ambiguity (it could be narration: «он покачал головой») is deliberately resolved against fabrication.

dudarenok-maker:
﻿## Fresh full review - head 1d36ac31 (all fixes folded)

Re-read every changed file end-to-end, re-probed edge cases specific to the NEW default-on + length-preserving behaviour, and re-confirmed CI. Verdict: no blockers; the previous blockers remain genuinely fixed; the default-on redesign is sound. A few non-blocking notes below for the record.

### ✅ What is solid (re-verified)
- DEFAULT-ON is safe because recovery is strictly length-preserving: the split replaces only the single whitespace char before an evidenced dash with a newline (1:1), so the body keeps identical length/offsets. Re-verified offset-preservation across: the real Nochnoi Dozor (all 9 chapters), multi-space, tab, entity-dash, and clean-line inputs. The aligner/model stay in the original coordinate space - no pipeline rewiring needed.
- FABRICATION GATES hold: evidence requires a tag that both has a speech/beat verb AND anchors a speaker (roster name or first-person я). Re-confirmed the earlier adversarial counter-examples (beat interruption, participial clause, mentally-qualified pronoun) now yield 0 speech, and a speech-verb in the narration alone does NOT trigger recovery.
- No-op on clean structure and quote-only languages; clean-book line unchanged.
- 254 dialogue-structure + structure-fixture tests pass; typecheck clean; full CI green (title, body-link, lint/typecheck, build, server fast+slow, all e2e + visual baselines, frontend+a11y, sidecar, verify, CLA).

### 🟡 Minor / notes (non-blocking)
1. Zero-whitespace dash after a period (`окно.— Привет`) is NOT recovered - the length-preserving split needs a whitespace char to turn into the paragraph break, and there is none. Conservative no-op (no fabrication, length preserved), but a rare authoring shape misses coverage. Already hinted in the code comment; a reviewer can accept or widen later. The document evidence... edit: the concern is a niche form, acceptable.
2. INTENDED metric shift to confirm with product owner: for a merged chapter whose turns carry name tags, the engine now CONFIRMS them at high confidence (confirmed=N flagged=0) instead of the OLD #2253 low-confidence flag (flagged=N). This is the deliberate upgrade of this PR and the test was updated accordingly, but it does change `confirmed/corrected/flagged` tally semantics for recoverable degraded books - worth an explicit acceptance note if any downstream metric reads those buckets as stability signals.
3. Retained tradeoff (documented in header): bare third-person-pronoun turns (`. - Нет, - ответил он.`) are NOT recovered - ambiguity resolved against fabrication. Coverage is therefore partial by design; the intake signal (structure-report) documents hiddenTurnCandidates as an upper bound.

### 🔗 Linkage confirmed
Issue #2254 is linked to this PR via GitHub cross-reference (the issue timeline carries `source.issue.url=.../issues/2320`). `closingIssuesReferences` is intentionally EMPTY: `Refs #2254` (partial delivery) must NOT auto-close the issue on merge - the Q2/attribution-cost question and pipeline-plumbing decision remain open on the ticket.

### Verdict
Mergeable as-is. No new blockers, the previous blocking findings are verified fixed, and the default-on change is sound (length-preserving, fabrication-safe). The minor items are coverage/acceptance notes, not gatekeepers.

Issue #1682, opened 2026-07-17
Analyzer: size cloud (Gemini/Gemma) requests to the model's token limits — don't send requests the tier can't handle


dudarenok-maker:
## Summary

The analyzer sends cloud (Gemini/Gemma) requests whose **input exceeds the model's per-request / per-minute token allowance**, so they 429 on the very first call and chapters are dropped. Root cause: chunk sizing is only token-aware for **local Ollama**; for cloud engines it reuses the full configured budget.

Observed on `gemma-4-31b-it` and `gemma-4-26b-a4b-it` with a free-tier key:

```
❌ Chapter 1/9 cast FAILED — Gemini gemma-4-31b-it daily quota exhausted
   Quota exceeded for metric: generate_content_free_tier_input_token_count, limit: 16000
```

A single stage-1 section is ~23,964 chars (~12–16k tokens), which is at/over the free tier's **16,000 input-tokens/minute** ceiling — so every request fails and the run drops every chapter.

## Root cause

`server/src/analyzer/stage2-chunk.ts` — `stage2ChunkBudgetForEngine()`:

```ts
if (engine !== 'local') return configured;   // ← cloud keeps the FULL budget
const numCtxDerived = Math.floor(numCtxTokens * 2 * 0.3);
return Math.max(1000, Math.min(configured, numCtxDerived));
```

Only the local path sizes chunks against a window. The cloud path returns the configured budget unchanged, so cloud requests are as large as local ones — with no regard for the cloud model's input-token limit. The stage-1 chunker (`stage1-chunk.ts`) has the same shape.

## Fix

Make cloud chunking **token-limit-aware**, so we never send a request the tier can't handle:

1. **Per-model input cap for cloud engines.** Add a config knob (e.g. `analyzer.gemini.maxInputTokensPerRequest`, default ~12,000 to sit safely under the free-tier 16k) and size cloud stage-1/stage-2 chunks to it (chars ≈ tokens × ~2, language-adjusted — Russian is token-dense, so be conservative).
2. **Token-rate pacing.** The 16k limit is per-*minute*, so even correctly-sized requests must be paced (≤ budget tokens/min). Wire the free-tier token budget into the existing Gemini rate limiter (`gemini.ts` ~888) so requests are spaced, not fired back-to-back.
[...]

PR #2330, opened 2026-08-13
chore(server): make stage-1 cloud TPM reservation roster-aware (#1691)

### LINKED ISSUE #1691: chore(server): make stage-1 cloud TPM reservation roster-aware (>130-char-cast ceiling)
## What

#1682 fixed stage-1 cast-detection cloud requests overshooting the finite Gemma 16000 TPM guard by reserving a **fixed** `STAGE1_CLOUD_RESERVED_TOKENS = 7000` (token-space) before sizing the body (`server/src/analyzer/stage1-chunk.ts`). The reservation covers the detection skill (~5.4k tokens) + inbox scaffold + a modest running roster.

The reservation is a **fixed constant** — it does not scale with the running roster, which accumulates the whole book's cast and is injected into the stage-1 inbox (`buildStage1ChapterInbox`). The whole-branch review measured the ceiling: the worst-case request stays under 16000 until roughly **~130 accumulated speaking-cast entries**. Beyond that the total estimate crosses 16000, the `RequestExceedsTpmError` guard throws, and — because it isn't an `AnalyzerTruncatedError` — the adaptive re-split can't catch it, so the chapter drops.

This is strictly better than pre-#1682 (an empty roster already blew the guard), and Ночной дозор and typical books sit far under the ceiling, so it did not block #1682. But the ceiling is real and currently only documented, not enforced.

## Why (technical)

A fixed reservation is a silent cliff for very-large-ensemble books. Making it roster-aware removes the ceiling instead of relying on it staying below ~130 characters.

## Acceptance

- Make the stage-1 cloud reservation account for the actual running-roster size (e.g. add the running-roster char-length as `reservedChars` on top of the fixed token reservation, using the `reservedChars` param `cloudBodyCharBudget` already exposes), OR another mechanism that keeps the worst-case stage-1 estimate under the TPM guard regardless of cast size.
- Extend the stage-1 regression lock (`stage1-chunk.test.ts`) to a large-cast (150+ entry) roster and assert it still clears 16000.
- No regression to the small/typical-cast path.

## Key files

- `server/src/analyzer/stage1-chunk.ts` (`STAGE1_CLOUD_RESERVED_TOKENS`, `resolveStage1ChunkCharBudget`)
- `server/src/analyzer/token-budget.ts` (`cloudBodyCharBudget` reservedTokens/reservedChars params)
- `server/src/routes/analysis.ts` (`buildStage1ChapterInbox`, the two stage-1 call sites)

## Depends on

- #1682 (introduces the fixed reservation) — do this after #1682 merges.

dudarenok-maker:
﻿## Summary

Stage-1 (cast-detection) cloud requests are sized against the finite Gemma TPM guard (16000/min). #1682 reserved a fixed `STAGE1_CLOUD_RESERVED_TOKENS = 7000` in *token* space for the system instruction + inbox scaffold, then sized the body in char space. That reservation did not scale with the **running roster**, which accumulates the whole book's cast. Past ~130 speaking-cast entries the worst-case request estimate crossed 16000, threw `RequestExceedsTpmError` (not an `AnalyzerTruncatedError`), and dropped the chapter.

This makes the reservation roster-aware: `resolveStage1ChunkCharBudget` now also reserves the running roster's rendered size (char space, mirroring `buildStage1ChapterInbox`'s compact `{id,name,role}` render) via the `reservedChars` param `cloudBodyCharBudget` already exposes, on top of the fixed token reservation. Both stage-1 call sites feed the live `rebuildRoster()` map in. The estimate stays under the guard regardless of cast size; the small/typical-cast path is unchanged (empty roster → 0 reserved chars).

Closes #1691

## Test plan

- [x] `tsc --noEmit` — green
- [x] `vitest run src/analyzer/stage1-chunk.test.ts src/analyzer/token-budget.test.ts` — 20/20 pass (incl. the #1691 150-entry-roster regression lock + a test asserting the budget shrinks as the cast grows)
- [x] `vitest run src/analyzer/stage2-chunk.test.ts` + `src/routes/analysis.test.ts` (200 tests) — green; no regression to the small/typical-cast or stage-2 paths
- [ ] `npm run verify:fast:branch` — full battery runs in cloud `verify.yml` (authoritative gate)

### DIFF
--- a/server/src/analyzer/stage1-chunk.test.ts
+++ b/server/src/analyzer/stage1-chunk.test.ts
@@ -164,13 +164,18 @@ describe('stage1ChunkBudgetForEngine', () => {
 
    RED/GREEN: with the prior 0-token reservation the body filled the full 12k
    cap and this construction estimated ~19.8k tokens (roster=60) — well OVER the
-   16000 guard. With STAGE1_CLOUD_RESERVED_TOKENS=7000 it estimates ~13.2k. */
-describe('#1682 — worst-case Cyrillic stage-1 request clears the Gemma TPM guard', () => {
+   16000 guard. With STAGE1_CLOUD_RESERVED_TOKENS=7000 it estimates ~13.2k.
+
+   #1691: the roster is now LARGE (150 entries — past the old ~130-cast wall
+   where the fixed-only reservation crossed the guard) and is fed into the
+   body-size resolver, so the reservation scales with the whole-book cast. */
+describe('#1682/#1691 — worst-case Cyrillic stage-1 request clears the Gemma TPM guard', () => {
   const GEMMA_TPM = 16000;
   const MARGIN_CEILING = 14500; // 16000 − 1500 safety margin
 
-  /* A conservative running roster for a single book: 60 characters, Cyrillic
-     names, in the compact {id,name,role} shape buildStage1ChapterInbox renders. */
+  /* A conservative running roster for a very large ensemble book: 150 characters,
+     Cyrillic names, in the compact {id,name,role} shape buildStage1ChapterInbox
+     renders. 150 deliberately exceeds the old fixed-reservation wall (~130). */
   const worstCaseRoster = (n: number): CharacterOutput[] =>
     Array.from({ length: n }, (_, i) => ({
       id: `personazh-imya-familiya-${i}`,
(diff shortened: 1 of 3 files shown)


dudarenok-maker:
## Independent review gate — outcome

Ran the mandatory `pr-review-gate` (Premium tier, `low` effort for a single-scope `chore`) over `4650ce55..HEAD` (adversarial, findings-only reviewer).

**Result: 0 correctness bugs; 4 cleanup-only nits → READY-WITH-CLEANUP.**

- Correctness bugs: **0** — no re-review trigger (per model-routing, cleanup-only findings don't re-run the gate).
- Cleanup nits: **4** (all Minor).

### Triage / folding

1. **Overstated "regardless of cast size" comment** (`server/src/analyzer/stage1-chunk.ts`) — **folded**: corrected to acknowledge `cloudBodyCharBudget`'s 2000-char floor as the practical ceiling (~290 all-Cyrillic cast). Committed `b5fc29d0`. Verified `tsc --noEmit` + 12/12 stage-1 tests still green.
2. **Char-space reservation is script-approximate** — **not folded**: this is exactly the `reservedChars` mechanism the issue requested, and `token-budget.ts` already documents it as a bounded approximation; a token-space variant is a future refinement, out of scope here.
3. **~13-char code fence in the inbox not counted** — **not folded**: negligible (~5 tokens) and already reflected in the real-prompt lock test.
4. **Test literal duplication** (`'а'.repeat(120000)` inline vs `cyr`) — **not folded**: trivial taste; left per CLAUDE.md surgical-changes guidance.

### Verification

- `tsc --noEmit` — green
- `vitest` stage1-chunk + token-budget: 20/20 pass (incl. the #1691 150-entry-roster regression lock + budget-shrinks-with-cast test)
- stage2-chunk + `analysis.test.ts` (200 tests): green
- Full `verify:fast:branch` runs in cloud `verify.yml` (authoritative gate) on this PR.

Issue #2306, opened 2026-08-12
Ночной дозор: 87.4% of dash-opening dialogue attributed to narrator — cause is upstream of the dialogue-structure engine


dudarenok-maker:
## What

On the C2/C3 on-box acceptance run (2026-08-12/13, Ночной дозор, all 9 chapters, local Ollama `qwen36-cw-iq4-32k`, structure engine on, `allowCloudFallback:false`), **87.4% of dash-opening dialogue lines were attributed to `narrator`** — 4131 of 4725.

The affected lines are unmistakable dialogue (`- Да.`, `- Тебе?`, `- Пошли!`) carrying confidence 0.9–0.95, i.e. booked as `confirmed`/`corrected` **successes**, not flagged for review. A book generated from this attribution renders most of its dialogue in the narrator's voice, silently.

The comparison run (2026-08-06, same book, same metric, chapters 1–4) measured **30.3%**. Both numbers are unshippable; the acceptance run is dramatically worse.

Per-chapter, the damage is **worst on the previously-cleanest chapters** — ch3 15.4% → 95.5%, ch2 20.2% → 93.9% — and least on the already-degraded ch5 (69.7% → 87.2%). The effect is **uniform, not structure-dependent**.

## What this is NOT — established by replay, do not re-litigate

Replaying **today's** dialogue-structure engine over the **2026-08-06 cached stage-2 output** reproduces **30.4%** against the recorded 30.3% (582 vs 580 lines, ch1–4). So #2187, #2253 and #2277 are not the cause. The regression is upstream of `crossExamine`, in this run's stage-2 output and/or roster.

**Blind spot of that method, stated so nobody over-reads it:** the replay starts at `crossExamine`, so it can exonerate the engine and nothing else.

Also ruled out, each with evidence:
- `isSpokenLine` blind to the ASCII hyphen this EPUB uses — no, `dialogueOpen` matches `-`
- `dialogueOpen` wiring lost between route and engine — no, still at `analysis.ts:2188`
- a pre-engine narrator default — no, one call site and it is in the structure-engine-OFF branch
- stage-2 truncation — 1 event, after the measured chapters
- per-chapter cast too small — 25/16/16/18 vs baseline 23/20/15/21

## The confound that has to be settled first

[...]

Issue #2263, opened 2026-08-11
srv — a book's language is detected from its front matter: sample body chapters, not the first 20k bytes


dudarenok-maker:
## What

`detectManuscriptLanguage` decides a book's language from **the first 20,000
characters of the manuscript** (`server/src/tts/detect-language.ts:46-50`), after a
pattern-based front-matter strip. `import.ts:150` feeds it `entry.sourceText` — the
whole document from byte 0 — while `entry.chapters`, already parsed, sits unused in
the same object.

The first 20k characters of a book are where the title page, copyright notice,
dedication, table of contents, epigraph and foreword live. So the language of every
imported book is decided from the part of the file least likely to be in the book's
language, and most likely not to be prose at all.

`stripFrontMatterBoilerplate` is the existing mitigation. It is pattern-based (nine
literal boilerplate regexes plus a byline/title match) and does not generalise.

## Evidence — this is live, not theoretical

Detecting **per chapter** across the 20-book live corpus, from the analysis cache:

```
  chs  agree  first  modal  onDisk  title
    3    67%  en     zh     zh      煤落的委托        <-- first chapter detects ENGLISH
   81   100%  en     en     en      Unlocked
   79   100%  en     en     en      Neverseen
   70   100%  en     en     en      Everblaze
   …
19 of 20 books unanimous across every chapter
```

**19 of 20 books agree on their language in every single chapter**, including books
of 79 and 81 chapters. A modal vote gets all 20 right.

The exception is the defect in the wild: the Chinese demo book's **first chapter
detects as `en`**. Today's whole-blob sample survives that only because Chinese body
text happens to dominate the first 20k characters. A book with a longer English
foreword would not be so lucky — and the failure is silent, because a wrong language
is a perfectly valid value.

The consequence is not cosmetic. Per `server/src/workspace/scan.ts:222-231` a
non-`'en'` language forces every character onto a designed Qwen voice and blocks the
Kokoro cross-language fallback; and `#2246` established there is no UI or API to
correct a book's language after import, so a wrong stamp is permanent short of
hand-editing `state.json`.

## Why the obvious mitigations don't work

Attempted and measured while fixing the `repair-missing-book-language` backfill
(#2246, PR #2251) — three gates, three different failures, all of them variations on
"score the poisoned sample harder":

| gate | closes | fails on |
|---|---|---|
[...]

Issue #2246, opened 2026-08-11
srv — 7 of 20 live books have no language at all, and a detection surrender is indistinguishable from a decision


dudarenok-maker:
Measured on the live workspace 2026-08-11 (20 live books, real modules, no
re-implementation):

**7 of the 20 live books have no `language` key in `state.json` at all** — the whole
*Keeper of the Lost Cities* series. They predate the field. Between them they hold
**25,063 dialogue lines, 62.8% of the corpus's 109,609 sentences.**

Today that is nearly harmless, because dialogue detection is language-blind. It stops
being harmless the moment #2245 makes detection language-driven: with no language
there is no conventions table, and every one of those 25,063 lines loses its dialogue
classification. **This must therefore land before or with #2245**, never after.

## Detection resolves them, and it resolves them honestly

`detectManuscriptLanguage` (`server/src/tts/detect-language.ts`) returns `en` for all
seven, and in every case via a **real `franc` match** — none hits either surrender
branch. Using the detected language, the conventions-driven rule recovers
**24,988 of 25,063 lines (99.70%)**. The 75-line residue is sentence-splitter
artefacts (dash-led fragments of interrupted turns), plus one book, *Unlocked*, whose
58 are mostly the literal `—Dame Alina`, a signature line in in-world letters.

## The blocker: a surrender is indistinguishable from a decision

`DetectionResult` is `{ language: string; supported: boolean }`. Both surrender paths
(`detect-language.ts` — the `letters === 0` branch and the terminal
`match ? … : result('en')`) answer `'en'`, and `'en'` **is** a supported language, so
`supported` is `true` on a surrender exactly as on a real match. There is no way to
tell them apart from the return value; the measurement above could only label the
seven by replaying the branch conditions by hand.

So a backfill cannot honour "write only on a real match, never on a guess" as the API
stands. **A `fallback` flag, set on BOTH surrender branches, is a prerequisite** — the
same additive change #1984 specified for its own reasons (its R-5C3).

## Why a wrong language is now worse than a missing one

After #2245, `language` stops being a label and starts selecting the quote tables that
decide what counts as dialogue. A book stamped `de` that is actually English gets
German quote pairs and loses its dialogue. That inverts the usual bias: the backfill
must prefer leaving a book unset and reporting it over writing a value it is not sure
of.

## The mechanism that let this hide

`bookStateLanguage` (`server/src/workspace/scan.ts`) delegates to
[...]

PR #2337, opened 2026-08-13
fix(server): keep the detected book language when confirm omits it

### LINKED ISSUE #2335: srv: confirm drops the detected book language and silently defaults to English
## What

`POST /api/import` detects the manuscript's language from the chapter bodies
(`detectManuscriptLanguageFromChapters`, the #2263 per-chapter vote) and returns
it to the client as `candidate.language`. It then **throws that result away**.

`POST /api/books` reads only `body.language`:

```ts
const language = normaliseBookLanguage(body.language);   // import.ts:236
```

`normaliseBookLanguage(undefined)` is `primary || DEFAULT_LANGUAGE` → `'en'`.
The route's own comment says so: *"Defaults to 'en' when absent."*

So a caller that does not echo the field back gets an **English book while the
server held the correct answer in memory seconds earlier**. The confirm screen
always sends it, so the UI is unaffected — every other API client is not.

## Why this is severe

`state.language` is not cosmetic. It selects the TTS voice set and
pronunciation, and it drives `languagePreamble`, which is what tells the
analyzer the book's dialogue conventions and forbids transliterating names. A
book silently marked `en` therefore:

- gets **no** language preamble at stage 1 → character names come back
  romanised (the #2313 symptom);
- gets no dash-convention guidance at stage 2, leaving the attribution rule
  *"Only words inside quote marks belong to a character"* unopposed over prose
  that marks dialogue with dashes;
- resolves `conventionsFor('en').dialogueOpen` to `null`, so the #2253
  dialogue-convention invariant and the #2325 collapse guard are both **inert**.

This is the confirmed cause of **#2306**. The 2026-08-12 acceptance run created
its book over the API without a `language` field — the request is still on disk:

```json
{"tempId":"imp_T2MQdLgdhN","title":"Ночной дозор (C2C3 run 2)",
 "author":"Сергей Лукьяненко","isStandalone":true}
```

That Russian book was analysed as English for ~20 hours and persisted at
**95.7% narrator across 15,100 sentences**, with a romanised cast. Reproduced
exactly on 2026-08-13: import detects `ru`, confirm omits `language`, state
persists `en`.

## Fix

Store the detection on the staging entry and fall back to it when `language` is
absent. An explicit value still wins outright (the confirm screen is
user-overridable, including an explicit `en` over a Russian detection). The
fallback applies only to a **supported** detection, so a request that
previously succeeded cannot start returning 400.

## Acceptance

- A Russian manuscript confirmed with no `language` field persists `ru`.
- An explicit `language` still overrides the detection in both directions.
- An English manuscript with no `language` still persists `en`.
- A staging entry predating the stored field keeps the historical default.

## Key files

- `server/src/routes/import.ts` — `/import` detection, `/books` resolution
- `server/src/store/import-staging.ts` — `StagedImport`
- `server/src/tts/language.ts` — `normaliseBookLanguage`

Distinct from #2256, which is about the backfill *script's* two-sample gate.

dudarenok-maker:
## Summary

`POST /api/import` detects the manuscript's language from the chapter bodies (the #2263 per-chapter vote) and returns it as `candidate.language` — then threw it away. `POST /api/books` read only `body.language`, and `normaliseBookLanguage(undefined)` is `primary || DEFAULT_LANGUAGE` → `'en'`. The route's own comment said so: *"Defaults to `en` when absent."*

So a caller that did not echo the field back got an English book **while the server held the correct answer in memory from seconds earlier**. The confirm screen always sends it, so the UI path was never affected — every other API client was.

The detection is now stored on the staging entry and used when the caller chose nothing. An explicit value still wins outright, in both directions.

**"Chose nothing" is a class, not a spelling** — absent, `null`, `''` and whitespace all mean it. The first version of this PR guarded on `undefined` only; `normaliseBookLanguage('')` is `'en'`, which *is* supported, so an unfilled form field reproduced the whole defect one spelling over.

## Correction: this is NOT established as the cause of #2306

This PR originally claimed the language default as #2306's confirmed cause. That claim was published before the experiment that tests it had been run, and the experiment refutes it.

**The correlation is real and verified** — the 2026-08-06 book that attributed well persisted `state.language = "ru"`; the 2026-08-12 book that collapsed persisted `"en"`, same manuscript and same engine. And the mechanism reads convincingly: marked `en`, a book gets no `languagePreamble`, and `conventionsFor('en').dialogueOpen` is `null`, which makes the #2253 convention invariant inert.

**The mechanism does not survive testing.** Read-out pre-registered in the harness header before running:

| path | `ru` | `en` |
|---|---|---|
| model — captured ch3 chunk-1 prompt, varying ONLY `StageCall.language` | 0.0% | **0.0%** |
| engine — captured raw model output through `crossExamine`, `dialogueOpen: undefined` | 1.6% | **5.6%** |

Against a recorded **94.8%**. Neither path reproduces the collapse. Every `en` book on this box was created by the same curl harness in the same sessions as the collapsed runs, so `state.language` may mark provenance rather than cause. **#2306's cause is open.**

**The fix stands on its own terms.** Stamping a Russian book English is wrong whether or not it is what broke that run.

## Review findings folded in (three passes)

**Pass 1 —** the empty-string hole above.

**Pass 2 — C1, and it was a regression this PR introduced.** `detect-language.ts` ran franc with `only: <the registry's four Latin iso6393 codes>`, so franc must answer with one of them: any other Latin-script language is **coerced** onto its nearest registered neighbour and returned `fallback: false`, as a decision. Measured: Italian→`spa`, Portuguese→`spa`, Dutch→`deu`. Before this PR those books landed on `'en'`; with a fallback that writes the detection, they would have persisted `'es'`/`'de'` — worse, because `isNonEnglish('es')` is true, so it drops kokoro/gemini/piper from the eligible engines, has Qwen design **Spanish** voices for an Italian book, and cannot be corrected afterwards (`PUT /api/books/:bookId/state` has no `language`). Fixed by cross-checking against an **unrestricted** franc call: disagreement means coercion, so `fallback: true`, while `language` keeps the restricted best guess for the confirm screen to pre-select.

**Pass 2 — C2**, and **pass 3 —** C2 had fixed one member of a four-member class. `author`, `title` and `series` sit in the same handler with the identical `(body.X ?? '').trim()` shape, so a non-string value threw a raw `TypeError` the generic `catch` surfaced as an **HTTP 500 carrying an internal message**. All four are now rejected as a class with a 400 naming the field. Every body field on the route was enumerated: `tempId`, `seriesPosition`, `isStandalone` and `excludedSlugs` were already type-guarded.

**Pass 3 — a comment claimed something false, and no test could have caught it.** The round-2 comment said C1 made the not-a-fallback clause "load-bearing for the first time". It does not: this route calls the *chapter-vote* variant, and `voteLanguage` filters coerced ballots before voting, with every surrender branch returning `resultFor('en', true)`. An all-Italian book yields `'en'`, never `'es'`. The C1 fix is still correct and still necessary — before it those ballots were `fallback: false` and won the vote — but the stopping happens inside `voteLanguage`, not at this route's clause. Both comments corrected, and the seam now has the regression test whose absence let a false claim survive a commit message, two comments and a review.

Also: docs (`162-fs2-multilanguage`, the language-recurrence plan's Task 3, the on-box register's hand-authored live view), `openapi.yaml`'s missing `nullable: true`, and two test cases that minted one `bookId` for two books.

## Test plan

- `npm run test:server` — **7,507 passed**; `npm run typecheck` clean.
- New coverage: a Russian manuscript confirmed with no `language` persists `ru`; an explicit choice beats a detection in both directions; `''`/whitespace/`null` all take the detection; twelve non-string cases across `author`/`title`/`series` return 400 rather than 500; an all-coerced Italian book returns `en`/`fallback: true` through the chapter vote.
- **Mutation-verified**, each reverted individually and restored: the C1 cross-check reddens the coerced-Latin test (`expected false to be true`); each of the three new field guards reddens exactly its own four-case block (`expected 500 to be 400`); removing `voteLanguage`'s ballot filter reddens the new vote test (`expected {language:'es',fallback:false} to equal {language:'en',fallback:true}`); removing the `#2246` prose-unit floor reddens 9 tests, confirming that floor is still exercised after C1 moved where its fixture is caught.
- **Corpus replay, at the granularity production actually uses.** The cross-check was first cited whole-book ("20/20 books agree") — but the route votes **per chapter**, on smaller samples. Re-measured per chapter across all 82 cached analyses: **607 franc ballots, 5 disagreements (0.8%)**, all real English prose that unrestricted franc labels `sco`, absorbed by the vote. **0 outcome changes across all 77 books with chapters.**

Closes #2335


dudarenok-maker:
## PR review — pass 3 (head d58372d9, effort high)

Scope: all 15 changed files (`git diff main...HEAD`), plus every consumer of `DetectionResult` and every non-string field on `POST /api/books`. Verified before probing: `npm run check:onbox-register` OK; the registry is `supported: true` at all seven entries (so N1's SUPPORTED-clause claim holds); a live `express + supertest` harness against the real `importRouter`; and a per-chapter replay of the C1 cross-check over all 82 cached analyses in `server/handoff/cache` (77 with chapters, 607 ballots reaching franc). Working tree untouched — `d58372d9`, `git status --porcelain` empty.

| Finding | Disposition |
|---|---|
| Pass 1 — `{"language": ""}` slips the predicate | **VERIFIED RESOLVED.** `languageChosen` (`import.ts:308-309`) treats absent / `null` / `''` / whitespace as one class; `it.each` covers all three spellings (`import.test.ts:365-395`). But see 🟡 for a PR body that documents the *opposite* rule, and 🟠 2b for a design consequence of folding `null` in. |
| Pass 2 C1 — restricted franc coerces an unregistered Latin language | **RESOLVED as a mechanism** (`detect-language.ts:112-120`); control test present. Two claims it produced do not hold — 🟠 2 and 🟠 2c. |
| Pass 2 C2 — non-string `language` → 500 with a raw internal message | **RESOLVED for `language` only.** The identical defect is still live on `author`, `title` and `series` in the same handler — 🔴 1. |
| N1 — the two fallback-gate conditions are not load-bearing | **HALF RESOLVED.** The SUPPORTED half is now correct and I confirmed it (all seven registry entries are `supported: true`). The NOT-A-FALLBACK half was re-stated as *load-bearing*, and that is false — 🟠 2. |
| N4 — plan Task 3 annotated | Task 3 itself is implementable from its own text and its one remaining scaffolded test is genuinely red. Task 10 was not annotated and is now unimplementable as written — 🟠 2b. |
| N5 — live view vs register | **VERIFIED RESOLVED.** The two #2306 blocks now agree paragraph-for-paragraph, the live view caught up on the "two candidates eliminated offline" sentence that already existed in the `.md` at `:2665`, and no row, group count or derived figure moved. `check:onbox-register` green. |
| N6 — `it.each` titles collided on `''` vs `'   '` | **VERIFIED RESOLVED** — titles now key on `label` (`import.test.ts:381`). |
| N2 / N3 / N7 | Not itemised in the dispatch; I found no unresolved nit of that shape in the diff. |

---

### 🔴 Blocking — C2 fixed one field of a four-field class in the same handler; `author`, `title` and `series` still answer HTTP 500 with a raw internal TypeError

`import.ts:299-307` guards `body.language`. Eighty lines earlier the identical expression shape is unguarded on three siblings:

- `import.ts:221` — `const author = (body.author ?? '').trim();`
- `import.ts:222` — `const title = (body.title ?? '').trim();`
- `import.ts:227` — `const series = isStandalone ? STANDALONES_SERIES : (body.series ?? '').trim();`

`catch` at `import.ts:451-454` is `res.status(500).json({ error: (e as Error).message })` — the same funnel C2 exists to keep a client off.

Reproduced against the real router (`express` + `supertest`, real `importRouter`, real staging entry), one request per field:

```
author    = 42  ->  HTTP 500  {"error":"(body.author ?? \"\").trim is not a function"}
title     = 42  ->  HTTP 500  {"error":"(body.title ?? \"\").trim is not a function"}
series    = 42  ->  HTTP 500  {"error":"(body.series ?? \"\").trim is not a function"}
language  = 42  ->  HTTP 400  {"error":"unsupported_language", ...}          <- the one that was fixed
```

`POST /api/books  {"tempId":"imp_…","author":"A","title":42,"isStandalone":true}` → 500 with an internal message on a client-facing body, on a service this repo serves over LAN HTTPS. This is recurring shape #10 (one instance fixed, the class left armed), and it is the *same handler*, not a sibling module — the PR defines the class in its own C2 comment (`import.ts:291-298`: "a caller sending a non-string … used to reach … `.trim()` … and throws a raw TypeError that surfaced as an HTTP 500 carrying an internal message"), then closes one of its four members. Per CLAUDE.md "Incidental findings", "it's pre-existing" and "it would expand the scope of this PR" are both void here; three lines and one `it.each` case close it.

---

### 🟠 Significant — the C1 comment's central claim, that the NOT-A-FALLBACK clause "is no longer inert", is false for the only field that reads it

`import.ts:277-282`:

> **#2337 review C1 changes this**: a franc match coerced onto the registry's nearest Latin neighbour now stamps `fallback: true` alongside a language OTHER than `'en'` (e.g. an Italian manuscript surrendering with `language: 'es'`) … The condition is no longer inert.

`entry.detectedLanguageFallback` / `entry.detectedLanguage` are written at `import.ts:134-136` from **`detectManuscriptLanguageFromChapters`**, not from `detectManuscriptLanguage`. Every surrender exit of the chapter-aware path is `resultFor('en', true)` — `detect-language.ts:157` (all ballots surrendered), `:175` (no majority), `:181` (under the prose floor), `:217` (no candidates). A coerced ballot is *removed* by the `!b.detection.fallback` filter at `:156`; its language can never leave the function. So `fallback: true` still implies `language === 'en'` at this call site, exactly as before C1.

Measured against the real functions:

```
detectManuscriptLanguage(<Italian>)                -> {"language":"es","supported":true,"fallback":true}
detectManuscriptLanguageFromChapters(<3x Italian>) -> {"language":"en","supported":true,"fallback":true}
```

Both branches of the ternary at `import.ts:310-313` therefore still produce `'en'`; deleting `&& !entry.detectedLanguageFallback` still changes no observable output. The clause remains correct and worth keeping — but the comment now asserts a reachability that does not exist, which is precisely the error N1 was raised to correct, re-made in the opposite direction. The same false statement sits in `import-staging.ts:52-56` ("in which case `detectedLanguage` is that coerced guess, NOT `'en'`").

The PR contradicts itself here: `openapi.yaml:5233-5240` and the regenerated `api-types.ts:3829-3836` describe the behaviour correctly ("It only falls back to `en` when that detection is **unsupported**") and never mention the surrender case — because observably there isn't one. Nothing in the diff tests the claim; a test asserting it would have failed.

Category: documentation/reasoning, not shipped behaviour.

### 🟠 Significant (2b) — the plan-doc annotation pass covered Task 3 and left Task 10 stating something this PR made impossible

`docs/superpowers/plans/2026-08-13-language-recurrence-and-prompt.md:318` — "add **\"Decide later\"**, submitting no language (→ Task 3's `null`)" — and `:80`, inside an **already-ANSWERED** owner decision (`:22`, D-1 accepted): "\"Decide later\" writes `null`".

After this PR there is no payload that means "decide later". `undefined`, `null`, `''` and `'   '` are all folded into `languageChosen === false` (`import.ts:308-309`), which now means *use the detection*. `null` was the one spelling that could have carried "stated absence" onto the wire, and pass 1's class fix claimed it. So on a confidently-detected book, "Decide later" would write the detected language, not `null` — the button would not defer. Task 10 needs a new wire signal, and the answered D-1 rationale at `:80` ("A surrendered detection means the system doesn't know; the user … may not either") no longer matches the route.

Task 3's own annotation is otherwise sound: implementable from its own text, and its one remaining unchecked scaffolded test ("writes language: null when detection surrendered") is genuinely red today.

### 🟠 Significant (2c) — the "20/20 corpus agreement" evidence is measured whole-book; production decides per chapter, where it is 602/607

`detect-language.ts:108-111`, `docs/release-notes-next.md:453` and the commit message all carry: "Measured against the live 20-book Latin-script corpus … the unrestricted answer agrees with the restricted one for all 20, so this changes nothing for any language the registry already supports."

The cross-check runs inside `detectManuscriptLanguage`, which `voteLanguage` calls **once per chapter** (`detect-language.ts:155`) on a `prepareSample` of that chapter alone. A whole-book agreement figure is measured under a different rule than the one that ships. Replaying per chapter over the same corpus (all 82 cached analyses, `server/handoff/cache`):

```
chapter ballots reaching the franc cross-check: 607
flagged as coercions:                             5  (0.8%)

mns_LfhWwNZifE  restricted=eng unrestricted=sco   716 words
mns_RGIaGt__19  restricted=eng unrestricted=sco  2641 words
mns_RGIaGt__19  restricted=eng unrestricted=sco  1514 words
mns_VoP0mLGvov  restricted=eng unrestricted=sco   535 words
mns_WC6OVeTw10  restricted=eng unrestricted=sco  1078 words
```

All five are ordinary English prose that unrestricted franc calls Scots — false surrenders, not coercions. The corpus does contain 5 `es`, 1 `fr` and 1 `de` books; none produced a coerced ballot, so the neighbour-language risk I went looking for (`spa` vs `cat`/`glg`/`por` on a short ballot) did not materialise here, and longer `es`/`fr`/`de` prose agrees restricted-vs-unrestricted in my own probes.

**The conclusion survives — I checked it.** Simulating the pre-C1 ballot set against the post-C1 one, book by book, gives **0 outcome changes across all 77 books with cached chapters** (the simulation reproduces the real `detectManuscriptLanguageFromChapters` exactly on all 77, which is its control). So no shipped book flips. But the sentence as written is not what was measured, and the 0.8% ballot-level false-surrender rate is the number a future change to this gate will need.

### 🟡 Minor

- **The PR body describes the opposite of what ships.** "**Guarded on `body.language === undefined`**, not falsiness, so an explicit empty string keeps its current meaning rather than silently acquiring the detected language" is pass 1's *rejected* design — `''` now does acquire the detection, and that is the point of `import.test.ts:365-395`. The Test plan still reads as head `95171629` ("Full server suite green via the pre-commit hook (482s)"), and neither C1 nor C2 — nor the `detect-language.ts` behaviour change, the register/live-view retraction, or the plan-doc annotation — is declared anywhere in it. CLAUDE.md requires incidental fixes declared in the PR body; a reviewer reading it would not learn that the detector changed. `Closes #2335` is present and outside a code span. ✅
- **An undisclosed behaviour change in the repair script.** A book whose only body chapter is one of the false-`sco` English chapters above now surrenders. Confirmed with a real corpus chapter: `mns_WC6OVeTw10` ch23 as a single-chapter book → `{"language":"en","supported":true,"fallback":true}`, previously `fallback:false`. `planBookLanguage` (`scripts/repair-missing-book-language.mts:297`) therefore reports `skip-fallback` ("detection surrendered") where it used to `backfill 'en'`. Benign — a missing key resolves to `en` anyway — but it is a real change to #2246's repair tool that no note mentions, and it is the mechanism by which the 0.8% ballot rate becomes user-visible.

### ✅ What is solid

The empty-string class fix; the C1 mechanism and its `es` control test; the C2 400 (it cannot reject a legitimate request — the confirm screen always sends a string, `null`/`undefined` route to the no-choice class — and the body carries only the caller's own echoed input plus the supported list, no internal detail); the `openapi.yaml` → `api-types.ts` regeneration including `nullable: true` → `string | null`; the `import.test.ts` #2263 change (the `wholeBlob` assertion is a fixture-sanity control and still pins the wrong-`en` answer the per-chapter vote exists to beat); the `.inner.test.mts` #2246 fixture change (outcome genuinely unchanged — `skip-fallback`, nothing written; only the reason string generalises, and the floor stays exercised by other fixtures); the register/live-view pair; and `Closes #2335`.

### Verdict

**One blocking correctness finding** — 🔴 1, three unguarded sibling fields answering 500 with an internal message, in the very handler this PR hardened — and **three significant non-correctness findings**: a false reachability claim in the code comments that the PR's own OpenAPI text contradicts; a plan Task 10 / answered D-1 that this PR made unimplementable; and corpus evidence measured at the wrong granularity (conclusion independently re-verified as correct, evidence restated). Not mergeable as-is on 🔴 1; 🟠 2 / 2b / 2c are prose and plan corrections that belong in the same round.

dudarenok-maker:
## PR review — pass 4 (head 489871f7, effort high)

Scope: a systematic audit of the **load-bearing claims** this branch introduced or edited — every comment, doc sentence and commit assertion that says why something is safe, inert, equivalent or outcome-neutral — plus the house gates. Verified before probing: registry `supported` flags (all 7 `true`), `voteLanguage`'s four return sites, the frontend confirm payload, the `/books` generic `catch`, franc restricted-vs-unrestricted on every fixture the branch cites, and an independent corpus replay. Deltas only; passes 1–3 not re-ploughed.

### Prior passes

| Pass | Finding | Status |
|---|---|---|
| 1 | explicit `""` slipped the guard | **VERIFIED RESOLVED** — `languageChosen` treats `''`/whitespace/`null`/absent as one class (`import.ts:339`), pinned by the 3-case `it.each` |
| 1 | stale `openapi.yaml` "Defaults to `en` when absent" | **VERIFIED RESOLVED** — rewritten, `nullable: true` added, `api-types.ts` regenerated to `string \| null` |
| 1 | `!detected.fallback` gate / impossible back-compat rationale | **VERIFIED RESOLVED** |
| 2 | **C1** franc coercion onto the 4 Latin registry codes | **VERIFIED RESOLVED**, and I re-measured every cited number myself: TOC→`jav`, Italian→`ita`, Portuguese→`por`, Dutch→`afr` (comment says "outside the set" — correct), Spanish agrees. The mixed EN/ZH fixture's `sco` claim reproduces exactly (`restricted=eng`, `unrestricted=sco`). |
| 2 | **C2** non-string `language` → 500 | **VERIFIED RESOLVED** |
| 3 | C2 fixed 1 of a 4-member class | **VERIFIED RESOLVED** — I enumerated all 8 fields on the confirm body independently: `tempId`, `seriesPosition`, `isStandalone`, `excludedSlugs` are genuinely type-guarded; `author`/`title`/`series`/`language` now each 400 before any `.trim()`. The `series` guard is correctly `!isStandalone`-scoped. Class closed. |
| 3 | "no longer inert" was false | **VERIFIED RESOLVED** — all four `fallback: true` returns of `detectManuscriptLanguageFromChapters` (`detect-language.ts:163,181,187,223`) are `resultFor('en', true)`. The N3 test pins it. |
| 3 | corpus cited whole-book, production votes per chapter | **VERIFIED RESOLVED and independently corroborated.** I replayed the live cache with my own harness (raw sentence-joined bodies, no front-matter strip): **660 Latin ballots, 7 disagreements (1.1%), every one `eng` vs `sco`** — same class, same order of magnitude as the branch's 607/5 (0.8%). |

Also checked and clean: `npm run check:onbox-register` passes (register ↔ live view agree; no row added or removed); the live view's new "eliminated offline" paragraph matches the register's; release-notes pair present; `Closes #2335` outside any code span; the second franc call costs **+2.8 ms per chapter** on a 15.6 KB sample (measured), so no import-latency concern; staging entries are per-`tempId` with the detection written in the same object literal as `chapters`, so no staleness or cross-talk is possible.

---

### 🟠 Significant — `backfill 'en'` → `skip-fallback` is a **write-outcome** change, and the note disclosing it denies it in the same sentence

**Cleanup (documentation), not a code-correctness bug** — flagged explicitly so triage routes it correctly.

`docs/release-notes-next.md:40`:

> a single-chapter English book on this path now reports `skip-fallback` … rather than `backfill 'en'` — **the observable write outcome (nothing written) is unchanged**, only the diagnostic reason text is more generic

The two halves contradict each other. `backfill` writes — `scripts/repair-missing-book-language.mts:553`, `await writeStateJsonAtomic(statePath, { ...state, language: plan.language })`. `skip-fallback` writes nothing. A sentence cannot both announce a change from `backfill` to `skip-fallback` and assert that "nothing written" is unchanged.

Round 2's version of this sentence (commit `d58372d9`) was **correct**, because it was scoped to the TOC *fixture*, which was `skip-fallback` before C1 (via the prose-unit floor) and after (via the surrender) — reason text only. Round 3 generalised it to "a single-chapter English book", a materially different case, and carried the old reassurance across unchanged.

**Concrete, run against this branch's own code** — chapter 5 of `server/handoff/cache/mns_RGIaGt__19.json`, real English prose:

```
chars 15268  words 2641  proseUnits 215  floor 20
head: "The lush pastures of Havenfield glittered into focus as Sophie's body re-formed…"
franc restricted = eng      franc unrestricted = sco      -> coerced = true
detectManuscriptLanguage(body)                        -> {"language":"en","supported":true,"fallback":true}
detectManuscriptLanguageFromChapters([{title, body}]) -> {"language":"en","supported":true,"fallback":true}
```

215 prose units is **10.75× `PROSE_UNIT_FLOOR`**. Pre-C1 this chapter was `{en, false}`, won its own vote unanimously and cleared the floor → `planBookLanguage` returned `backfill 'en'` and `--apply` wrote `language: 'en'`. Post-C1 the ballot is filtered out of `nonSurrendered`, `voteLanguage` surrenders, and the book is reported `skip-fallback` — `--apply` writes nothing, leaving it in the absent-key state the script exists to remove. Its own header states why that matters: *"absent and explicit `'en'` are INDISTINGUISHABLE today … nothing in the product can tell the two states apart or correct either one."*

That book is from the **"Keeper of the Lost Cities" series** — i.e. the exact seven-book, pre-fs-2, language-less population the script was written for. The affected set is not hypothetical.

Note also that `detect-language.ts:116`'s "**0 outcome changes** across all 77 books with chapters" does not cover this and cannot: it measures the vote's *language*, and this case keeps the language (`'en'`) while flipping `fallback` — precisely the field the repair script's action keys on. The corpus figure is correct for what it measured; it is being read as clearing a class it never scored.

**Fix:** state the real delta ("a single-chapter English book the cross-check flags now skips instead of backfilling `'en'`; the effective language is unchanged, because an absent key still resolves to `'en'` via `bookStateLanguage`, but the key stays absent"), or drop the "write outcome unchanged" clause. The behaviour itself is defensible — refusing to write a language it is no longer sure about is the script's own load-bearing rule — so this is a wording fix, not a code fix.

---

### 🟡 Minor

All cleanup. None re-triggers the gate.

**1. `server/src/routes/import.ts:286-288` — a sentence that is false for one of the two functions it names.**

> Every `detectManuscriptLanguage`/`detectManuscriptLanguageFromChapters` surrender path stamps `language: 'en'` alongside `fallback: true`

False for `detectManuscriptLanguage` since C1: `detect-language.ts:126` returns `resultFor(match.code, coerced)`, so its coercion branch yields a non-`'en'` language with `fallback: true`. Verified against the branch's own Italian fixture → `{language:'es', supported:true, fallback:true}`. The paragraph five lines below corrects the record *for this route's path*, and the route's reasoning is sound — but the sentence as written is the takeaway, and it is the exact shape passes 1–3 each caught. Concrete consequence: a later change routing a single-chapter import through the exported `detectManuscriptLanguage` (as the repair script already does) would, on this sentence's authority, look safe to drop the `!entry.detectedLanguageFallback` clause — and would then persist `'es'` for an Italian book. Scope the sentence to `detectManuscriptLanguageFromChapters`.

**2. `server/src/store/import-staging.ts:51-55` — the surrender enumeration is not exhaustive.**

`detectedLanguageFallback`'s doc lists three causes ("no letters to sample, no franc match, or … every candidate chapter's franc match … was a coercion"). `detectManuscriptLanguageFromChapters` has two more a reader would not guess: **no strict majority** (`detect-language.ts:181`) and **the winning language under `PROSE_UNIT_FLOOR`** (`:187`), plus empty-candidates (`:223`). Concretely: a bilingual book split 50/50, or a short book whose winner scores <20 prose units, sets this field `true` for a reason the doc does not name. The load-bearing *rule* — every surrender branch returns `resultFor('en', true)` — is true and is what N3 rests on; state the rule rather than a list, which is how a guard loses one spelling per round.

**3. `server/src/routes/import.ts:339-340` — dead branch.**

```ts
const languageChosen =
  typeof body.language === 'string' ? body.language.trim() !== '' : body.language != null;
```

The `:330` guard already returned 400 for every non-null non-string, so by `:340` `body.language` is a string or nullish and the else-branch is always `false`. Harmless — but it is verbatim the expression the C2 test's own comment describes as the *pre-fix* behaviour ("any other non-null value fell through to `body.language != null`, which is true"), so a reader comparing comment to code finds the described-as-buggy expression still sitting there. `typeof body.language === 'string' && body.language.trim() !== ''` says the same thing.

**4. `server/src/tts/detect-language.ts:3` — stale header, now load-bearing for a new claim.**

The module header still reads *"The script pre-pass is authoritative (Cyrillic⇒ru, CJK⇒**unsupported**)"*. zh/ja have been `supported: true` since fs-59 W5 — this file's own `:92` comment says so, and `resultFor('zh'|'ja', false)` returns `supported: true`. Pre-existing sentence, but this PR is what makes it matter: the new fallback at `import.ts:342` gates on `entry.detectedLanguageSupported`, so a reader checking "can the SUPPORTED clause ever fire?" against this header concludes a CJK book trips it. It does not — a Chinese book confirmed with no `language` now persists `'zh'`, which is correct and intended, and which the header says is impossible. One-line fix in a file this branch already rewrote 30 lines below.

---

### ✅ What is solid

- **C1 is the right fix and its evidence holds up under an independent harness.** My replay (different sampling, different filtering) landed 7/660 vs the branch's 5/607 — same class (`eng` vs `sco`), same magnitude. Nothing suggests the figure was cherry-picked.
- **The cross-check's false-positive risk on the languages the fallback actually writes was unmeasured — so I measured it, and it is clean.** The corpus is English + Russian, so "all 5 disagreements were English prose" cannot clear es/fr/de. I ran realistic multi-paragraph Spanish, French and German prose at 1× and 3× chapter scale (857–2,642 chars): restricted and unrestricted agree in all six runs. No finding — recorded so the gap is not re-raised.
- **The N2 field class is genuinely closed**, not merely widened — I enumerated the body type independently rather than trusting the commit message's list.
- **The N3 reasoning is correct at the code level**, which is the thing pass 3 could not take on trust: I walked all four return sites of the chapter-vote path rather than reading the comment.
- **Concurrency/lifetime on the staging entry is a non-issue**: keyed per `nanoid` `tempId`, detection written in the same literal as `chapters`, 30-min TTL, evicted on drain. There is no window in which the stored detection describes different chapters than the entry holds.
- **Perf is a non-issue**: +2.8 ms per chapter, measured.
- The `sco` chapters are absorbed by the mass vote in every multi-chapter book, as claimed — the one corpus book with two flagged chapters still resolves `en` / `fallback: false`.

### Verdict

**No blocking findings. No correctness bugs.** One Significant documentation finding (a release-note sentence that announces a write-outcome change and denies it in the same clause) and four Minor cleanups — all prose, none touching behaviour.

Per the triage contract this pass contains **zero correctness bugs**, so folding these in does not re-trigger the gate. Worth doing before merge anyway: the Significant one is the only claim in this branch a reader could act on wrongly.

Two checklist items outside the diff, for the shipping session: publish `docs/testing/onbox-acceptance-register-live-view.html` to the **URL recorded in the register header** (not a fresh one, and not the `.md`), and run `npm run check:onbox-register -- --against-published <saved copy>` immediately before doing so.

PR #2339, opened 2026-08-13
chore(docs,scripts): widen pr-review-gate into the full PR review process

### LINKED ISSUE #2338: ops-56 — widen pr-review-gate into the full PR review process
## What

`.claude/skills/pr-review-gate/` currently says only *how to dispatch a reviewer*. Widen it into the single home for this repo's PR review process:

- `SKILL.md` — the shipping session's runbook (preconditions, docs-only exemption, effort ladder, dispatch, the per-pass PR comment, triage, re-review loop, issue verification at PR creation).
- `references/reviewer-brief.md` — the rubric the reviewer reads *by path*: this repo's bookkeeping gates plus a curated catalogue of ~10 recurring defect shapes, each stated as *how it hides*.
- `references/findings-triage.md` — the fix-now bar, the void deferral reasons, the design-pass carve-out.

`model-routing/SKILL.md` loses `## Mandatory independent review (PRs)` and `## PR-gate issue verification` and keeps routing: the tier table, escalation, session drift, the spec/plan loop, and the shared judgment-call carve-out.

Every review pass now posts a summary comment on the PR, so review history outlives the chat context that produced it. A pass that finds nothing posts too — otherwise the record cannot distinguish *reviewed and clean* from *never reviewed*.

## Acceptance

- `pr-review-gate/SKILL.md` carries the dispatch mechanism and the effort ladder; the guard asserts it.
- `model-routing/SKILL.md` carries neither moved section; the guard asserts that too.
- Both `references/*.md` exist **and** are named by `SKILL.md` — a dispatch prompt pointing at a missing file hands the reviewer no rubric, silently.
- Every intra-repo anchor link in the governance docs and skills resolves to a real heading.
- `npm run skills:sync` mirrors the skill into `~/.agents/skills/`, the only store Cline reads.

## Key files

`.claude/skills/pr-review-gate/**`, `.claude/skills/model-routing/SKILL.md`, `CLAUDE.md`, `scripts/tests/review-gate-mechanism.test.mjs`, `scripts/verify-cache.mjs`, `scripts/sync-agent-skills.mjs`

## Notes

Design: `docs/superpowers/specs/2026-08-13-pr-review-skill-design.md`. Plan: `docs/superpowers/plans/2026-08-13-pr-review-skill.md`. Probe: `docs/testing/agent-skill-resolution-probe.md`.

No `docs/BACKLOG.md` row — `type:chore` issues never render there.

dudarenok-maker:
## Summary

`.claude/skills/pr-review-gate/` said only *how to dispatch a reviewer*. It is now the single home for this repo's PR review process — the runbook, the rubric, the triage rules, and a durable per-pass record on the PR itself.

- **`SKILL.md`** — the shipping session's runbook: preconditions, docs-only exemption, effort ladder, dispatch, the pass comment, triage, re-review loop, issue verification. For gates CLAUDE.md already owns it *names and links* them rather than restating — a second copy of a rule is a rule that drifts.
- **`references/reviewer-brief.md`** — the rubric, read by the reviewer **by path** so it arrives verbatim: this repo's bookkeeping gates, plus ~10 recurring defect shapes each stated as *how it hides*.
- **`references/findings-triage.md`** — the fix-now bar, the defect/chore/taste seam, the void deferral reasons, the design-pass carve-out.
- **`model-routing/SKILL.md`** loses the two PR sections and keeps routing.

Two design decisions worth flagging for the reviewer:

**The reviewer posts its own comment.** An earlier draft had the shipping session relay it, justified by the reviewer having "no write access" — that justification was false. The `Agent` tool has no per-dispatch tool restriction, and even `Explore` keeps `Bash`. Routing the report through the session bought nothing but an unverified relay operated by the one party with a motive to soften findings. The reviewer's boundary is now stated as what it is — *modifies no tracked file* — and made checkable: the session diffs `git rev-parse HEAD` + `git status --porcelain` before and after, and any delta is a gate failure.

**The mirror guard fails open, deliberately.** Cline reads only `~/.agents/skills/`, outside the repo, so the drift guard skips when that path is absent (fresh clone, CI). That is defect shape #1 in this PR's own rubric; it is unavoidable here rather than accidental, it prints a visible skip line, and a green run must never be reported as "the mirror is in sync."

## Test plan

- [x] `npm run test:hooks` — 1372/1372
- [x] Every new assertion mutation-verified in both directions, including the anchor helper restored to its buggy form to confirm it goes red
- [x] Content-fidelity diff of the moved effort ladder against `origin/main`: `low`/`medium`/`high` criteria, multi-scope rule, mixed-type rule, `ultra` opt-in rule and docs-only file-set test all unchanged; the issue-verification section is byte-for-byte identical
- [x] `npm run skills:sync` → Cline lists `pr-review-gate`
- [ ] Reviewer: the fail-open mirror guard is the one place this PR knowingly accepts a weaker check — confirm the reasoning holds

**Release notes: not applicable** — process/tooling, no user- or operator-visible delta.
**On-box acceptance: not applicable** — no hardware-provable behaviour.
**Not docs-only** — touches `scripts/**`, so `verify.yml` runs in full and this gate applies to itself.

- **`.gitignore` blanket-ignored `.claude/`** — the two new reference files were invisible to `git add`, which skips ignored paths *silently*. The commit would have carried the guard asserting the files exist and not the files, passing locally and failing in CI. Fixed with `.claude/*` + `!.claude/skills/` rather than `git add -f`, which would re-arm the trap for every future file.
- **`CLAUDE.md:716`'s anchor died with the moved section** — and the existing guard, which only string-matches `pr-review-gate` on that line, would have stayed green while the link 404'd. A new link-integrity assertion now catches it. A mechanical sweep found **four** such links where a read had found one.
- **`CLAUDE.md:301-306`** pointed at model-routing for a ladder that no longer lives there.
- **`extraFiles` enumerated three literal paths** — a list that cannot see a file that does not exist yet. Replaced with a `.claude/skills/**` glob.
- **The guard file's header comment numbered its assertions in a different order than the tests appear**, making "assertion 3" ambiguous between two different tests.
- **The sync script duplicated `SKILL.md`'s frontmatter** into the mirrored copy purely so the guard could split on a marker — a repeated metadata block in the file a reviewing agent reads, to save the test one line.

Closes #2338

### DIFF
--- a/.claude/skills/model-routing/SKILL.md
+++ b/.claude/skills/model-routing/SKILL.md
@@ -79,94 +79,15 @@ for subagent dispatch above.
   — do not keep looping automatically past the cap.
 - **Judgment-call carve-out**: see below.
 
-## Mandatory independent review (PRs)
+## PR review
 
-- **Sequence**: finalize implementation → local `npm run verify:fast:branch`
-  (cloud `verify.yml` is the enforcing gate) → `gh pr
-  create` → work through `CLAUDE.md`'s before-shipping checklist item by
-  item (each done or explicitly marked not-applicable), committing and
-  pushing along the way → once every applicable item is addressed and
-  everything is pushed, that is "fully staged" and this review triggers on
-  that state (not on an earlier, incomplete push).
-- **Exemption**: a docs-only PR — changed-file set entirely under `docs/**`,
-  root-level `*.md`, or `.github/*.md` (the same file-set test
-  CONTRIBUTING.md's "Doc-only PR fast-path" already uses for the `verify.yml`
-  `paths-ignore` skip) — is exempt from this gate entirely; no review
-  pass runs.
-- **Effort level**: for every other PR, scales with the PR's commit
-  type/scope, reusing CONTRIBUTING.md's existing commit-convention
-  vocabulary rather than a new classification:
-  - **`low`** — single-scope `chore`, `test`, `build`, or `ci` (mechanical,
-    no user-facing behavior change).
-  - **`medium`** — single-scope `feat` or `fix`.
-  - **`high`** — `refactor`, `perf`, or any multi-scope PR (CONTRIBUTING.md's
-    own "use sparingly" multi-scope guidance already flags these as
-    higher-risk) — today's unchanged default.
+Moved out of this file 2026-08-13. The sequence, the docs-only exemption, the
+effort ladder, dispatch, the PR comment, findings triage, the re-review trigger
+and loop cap, and issue verification at PR creation all live in
+[`pr-review-gate`](../pr-review-gate/SKILL.md). Routing keeps routing; that file
+owns the PR process.
 
-  A PR whose commits mix types takes the highest tier any single commit
-  would earn on its own (e.g. one `fix` commit + one `refactor` commit →
-  `high`). `ultra` is never auto-selected here — it is billed and requires
-  explicit user opt-in (`/code-review ultra`, or the user asking for it by
-  name), per the Workflow tool's own rule.
-- **Mechanism**: invoke the `pr-review-gate` skill, which dispatches a
-  **non-fork** subagent at this table's **Premium** tier carrying the
-  adversarial gate brief, at the effort level determined above, once fully
-  staged — a findings report only, triaged by hand; the subagent edits nothing.
-  The same bar applies when the user runs `/code-review` themselves: no `--fix`.
-  `--fix` applies whatever the pass surfaced wholesale, and there's no per-finding
-  confidence filter to gate it on, so triage happens by hand instead — see
-  Findings handling. This is a working/branch-diff review — not the separate
-  `/review` PR-comment slash command, and not the
-  `code-review@claude-plugins-official` plugin command either: that plugin
-  reviews a GitHub PR, discards every finding below 80% confidence, and
-  posts a public comment on the PR instead of returning a report to triage.
-  - **Why not invoke the built-in `code-review` skill directly**: it is
-    user-invocable only — `Skill(skill: "code-review")` fails with "cannot
-    be used with Skill tool due to disable-model-invocation", and a
-    dispatched subagent hits the same wall, so "dispatch an agent and have
-    it invoke the skill" is not a workaround. It remains available and is
-    the *better* reviewer: the user may type `/code-review <level>` at any
-    point, and when they do it supersedes the agent pass for that round.
-  - **Never report the gate as having run when it did not.** If the agent
-    pass was substituted, or skipped, say so plainly in the user-facing
-    summary.
-- **Findings handling**: triage the report by hand, then **fix in this round**.
-  Every finding gets a dispatched fix agent — one finding, one fix, one paired
-  test — committed and pushed before merge. Clear-cut findings (unambiguous
-  bug, obvious dead code, a straightforward CLAUDE.md violation) are the
-  ordinary case, not the only case. **Cleanup-only findings are in scope too** —
-  a reuse/simplification/efficiency nit, a staled derived artifact, a missing
-  index or register row, an unwired knob. They are cheaper to dispatch than a
-  bug, not a lesser class of work, and "it's only a chore" / "it's not
-  user-visible" is not a deferral. (This is a *fixing* rule, not a re-review
-  rule: cleanup-only findings still do not re-trigger a pass — see below.) The
-  one exception stays taste: a reviewer preferring a different shape for code
-  that is correct and consistent with its neighbours is declined, not filed.
-  **The single finding allowed to leave the round unfixed is one needing a
-  design pass** — more than one defensible outcome, a decision the user has a
-  stake in — which routes through the judgment-call carve-out below and gets an
-  issue naming that decision as its deliverable. Everything else is fixed now:
-  "it expands the PR's scope", "it's pre-existing", "it needs its own test",
-  and "we can batch these" are **not** deferral grounds — see [CLAUDE.md →
-  Incidental
-  findings](../../../CLAUDE.md#incidental-findings-report-fix-record) for the
-  full void-reasons list and the defect/chore/taste seam. A ten-finding report
-  ends as ten fix commits, not a follow-up epic. None of this licenses `--fix`:
-  the fixes are dispatched and reviewed per finding, never applied wholesale.
-- **Re-review trigger**: only when the initial pass surfaced ≥1 finding
-  that is an actual correctness bug (wrong behavior, crash, security issue —
-  not a reuse/simplification/efficiency-only cleanup nit). Fixing and
-  pushing those re-triggers a pass. If the initial pass came back empty, or
-  surfaced only cleanup-only findings, fix-and-push (or push nothing) does
-  **not** re-trigger a re-review — re-running it in that case just burns
-  tokens for no new signal. This mirrors the spec/plan loop's severity-gated
-  shape (above), rather than firing on every push. Re-review re-derives the
-  effort level from the PR's current commit set (above) rather than
-  reusing the initial pass's tier — a fix commit can raise the tier the
-  same way any other commit would.
-- **Loop cap**: 2 re-review rounds, same numeric cap as the spec/plan loop
-  above (initial pass + up to 2 re-review rounds, 3 total).
-- **Judgment-call carve-out**: see below.
+The judgment-call carve-out below is shared by both review loops and stays here.
 
 ## Judgment-call carve-out (shared by both review loops)
 
@@ -176,34 +97,3 @@ and routes through the normal ask-first behavior in `CLAUDE.md`'s "Think
 before coding" — it does not get silently resolved just to keep the loop
 moving. This is the same failure mode in both loops: an automated loop
 mistaking a decision for a defect.
-
-## PR-gate issue verification
-
-- **Trigger**: every `gh pr create` for non-trivial work, including
-  bug-shaped work.
-- **Check**: the PR body must contain `Closes #NN` or `Refs #NN` referencing
-  an existing GitHub issue, outside of any inline-code/fenced-code span (a
-  backtick-wrapped `` `Closes #NN` `` does not actually auto-close on
-  GitHub — write it plain).
-- **Missing case**: auto-file a new issue capturing the work, then add
-  `Closes #NN` to the PR body — proceed without pausing to ask, in every
-  case including bug-shaped work (a deliberate, explicit override of
-  `CLAUDE.md`'s general "user files bugs" convention, scoped to this one
-  gate). Label per `CONTRIBUTING.md`'s actual two-shape convention:
-  - **Bug-shaped** (fixing existing broken behavior, no design decision):
-    standalone `bug` label, plain descriptive title. No `area:`/`type:`/
-    `moscow:`.
-  - **Backlog-shaped** (new/changed behavior): `area:<prefix>` +
-    (`type:feature` or `type:chore`). `moscow:` left unset for the user.
-    Use `type:chore` when the work matches this repo's commit-type-`chore`
-    shape (codegen, version bumps, tidy-up — no user-facing behavior
-    change); `type:feature` otherwise.
-- **Timing**: at PR creation — distinct from, and prior to, the mandatory
-  independent review above, which reviews code + docs combined once the PR
-  (and its issue link) already exist.
-- **Mechanical backstop**: `.github/workflows/pr-issue-link.yml` fails the
-  PR check if neither `Closes #\d+` nor `Refs #\d+` appears in the body
-  (outside code spans) — the one gate in this file with a real, external
-  enforcement, not just this convention. It does not check labeling or
-  whether the auto-file step above ran correctly, only that some issue
-  reference exists.
--- a/.claude/skills/pr-review-gate/SKILL.md
+++ b/.claude/skills/pr-review-gate/SKILL.md
@@ -1,96 +1,271 @@
 ---
 name: pr-review-gate
-description: Use when a PR is fully staged and ready for its mandatory independent review, or when running CLAUDE.md's before-shipping "Independent PR review" step. This is the skill to invoke to actually dispatch the reviewer — for choosing the tier/effort level or the rest of the review-gate mechanics, consult `model-routing` instead. Dispatches a non-fork, Premium-tier subagent carrying an adversarial reviewer brief — a findings report only, never auto-applied.
+description: Use when preparing to ship a PR, staging one for merge, or running CLAUDE.md's before-shipping "Independent PR review" step. This is the full PR review runbook — preconditions, the docs-only exemption, the effort ladder, dispatching the reviewer, posting the pass comment, findings triage, the re-review loop, and issue verification at PR creation. For choosing a subagent's model tier outside a PR review, or the spec/plan adversarial-review loop, consult `model-routing` instead. Dispatches a non-fork subagent at the routing table's Premium tier carrying an adversarial reviewer brief — a findings report only, never auto-applied.
 ---
 
 # PR review gate
 
 The mandated mechanism for [CLAUDE.md's before-shipping step
-10](../../../CLAUDE.md) and
-[`model-routing`'s "Mandatory independent review
-(PRs)"](../model-routing/SKILL.md#mandatory-independent-review-prs) — that
-file owns the effort ladder, the exemption, the re-review trigger, and the
-loop cap; this file is the reviewer brief itself, not a restatement of the
-mechanics. Read it there, not here. (Three safety-critical rules — never
-report a gate that didn't run, `/code-review` supersedes this pass, no
-auto-apply — are deliberately repeated in both files rather than owned by
-just one; don't delete the duplication in a future cleanup pass.)
+10](../../../CLAUDE.md#before-shipping-checklist) — this file is the full
+spec for what this repo expects of a PR review: the procedure the shipping
+session runs, the rubric the reviewer applies
+([`references/reviewer-brief.md`](references/reviewer-brief.md)), what
+happens to the findings
+([`references/findings-triage.md`](references/findings-triage.md)), and the
+durable record each pass leaves on the PR itself. `model-routing` keeps the
+model-tier routing table and the spec/plan adversarial-review loop; this file
+owns everything PR-specific.
 
 ## When this fires
 
 A PR is fully staged: implementation finalized, `verify:fast:branch` green,
-every applicable before-shipping checklist item addressed (or explicitly
-marked not-applicable), and everything pushed. Not on an earlier, incomplete
-push. Except a docs-only PR, which is exempt entirely — see
-`model-routing`'s *Exemption* for the file-set test.
+every applicable [before-shipping checklist](../../../CLAUDE.md#before-shipping-checklist)
+item addressed (or explicitly marked not-applicable), and everything pushed.
+Not on an earlier, incomplete push.
+
+## Exemption
+
+A docs-only PR — changed-file set entirely under `docs/**`, root-level
+`*.md`, or `.github/*.md` (the same file-set test
+[CONTRIBUTING.md's "Doc-only PR fast-path"](../../../CONTRIBUTING.md#doc-only-pr-fast-path)
+already uses for the `verify.yml` `paths-ignore` skip) — is exempt from this
+gate entirely; no review pass runs.
+
+Exempt does not mean silent: post a one-line exemption note on the PR naming
+the file-set test that exempted it (see [The PR comment](#the-pr-comment)).
+A large share of this repo's PRs are docs-only, so without the note the
+durable record would be missing exactly where the volume is.
+
+## Effort level
+
+For every non-exempt PR, effort scales with the PR's commit type/scope,
+reusing CONTRIBUTING.md's existing commit-convention vocabulary rather than a
+new classification:
+
+- **`low`** — single-scope `chore`, `test`, `build`, or `ci` (mechanical, no
+  user-facing behavior change).
+- **`medium`** — single-scope `feat` or `fix`.
+- **`high`** — `refactor`, `perf`, or any multi-scope PR (CONTRIBUTING.md's
+  own "use sparingly" multi-scope guidance already flags these as
+  higher-risk) — today's unchanged default.
+
+A PR whose commits mix types takes the highest tier any single commit would
+earn on its own (e.g. one `fix` commit + one `refactor` commit → `high`).
+`ultra` is never auto-selected here — it is billed and requires explicit user
+opt-in (`/code-review ultra`, or the user asking for it by name), per the
+Workflow tool's own rule.
+
+State the level explicitly in the dispatch prompt so the subagent calibrates
+depth instead of guessing.
 
 ## Dispatch
 
-- **Premium tier** (per the routing table), **non-fork**. A fork inherits
-  the dispatching session's own context — its framing, its reasoning, its
-  blind spots — which is the opposite of independent review. A fork also
-  always runs on the dispatching session's model, ignoring model routing
-  entirely; a Premium-tier review needs a real non-fork dispatch to land on
-  Opus.
-- **Effort level** is passed in by the caller, not decided here — `low`,
-  `medium`, or `high`, derived from the PR's commit type/scope per
-  `model-routing`'s ladder (`ultra` is opt-in only, never auto-selected).
-  State the level explicitly in the dispatch prompt so the subagent
-  calibrates depth instead of guessing.
-- **Never auto-apply.** The subagent edits nothing. It returns a findings
-  report, triaged by hand per `model-routing`'s "Findings handling". (If the
-  user runs `/code-review` themselves instead, the same bar holds: no `--fix`.)
-- **The reviewer reads the rubric itself.** The dispatch prompt instructs it to
-  read `.claude/skills/pr-review-gate/references/reviewer-brief.md` in full
-  before it starts. Triage rules for what comes back are in
-  `.claude/skills/pr-review-gate/references/findings-triage.md`.
+- **Premium tier** (per [`model-routing`'s routing
+  table](../model-routing/SKILL.md#routing-table)), **non-fork**. A fork
+  inherits the dispatching session's own context — its framing, its
+  reasoning, its blind spots — which is the opposite of independent review. A
+  fork also always runs on the dispatching session's model, ignoring model
+  routing entirely; a Premium-tier review needs a real non-fork dispatch to
+  land on Opus.
+- **Fresh context** — not a continuation of any earlier pass in this session.
+- **Never auto-apply.** The subagent edits nothing (verified by [the tree
+  check](#the-tree-check), not trusted). It returns a findings report,
+  triaged by hand per [Triage](#triage). If the user runs `/code-review`
+  themselves instead, the same bar holds: no `--fix`. `--fix` applies
+  whatever the pass surfaced wholesale, and there's no per-finding confidence
+  filter to gate it on, so triage happens by hand instead. This is a
+  working/branch-diff review — not the separate `/review` PR-comment slash
+  command, and not the `code-review@claude-plugins-official` plugin command
+  either: that plugin reviews a GitHub PR, discards every finding below 80%
+  confidence, and posts a public comment on the PR instead of returning a
+  report to triage.
+- **The reviewer reads the rubric itself.** The dispatch prompt points it at
+  [`references/reviewer-brief.md`](references/reviewer-brief.md) **by path**
+  and instructs it to read the file in full before it starts — not a
+  paraphrase retyped into the prompt. Triage rules for what comes back are in
+  [`references/findings-triage.md`](references/findings-triage.md).
+- **Why not invoke the built-in `code-review` skill directly**: it is
+  user-invocable only — `Skill(skill: "code-review")` fails with "cannot be
+  used with Skill tool due to disable-model-invocation", and a dispatched
+  subagent hits the same wall, so "dispatch an agent and have it invoke the
+  skill" is not a workaround. It remains available and is the *better*
+  reviewer: the user may type `/code-review <level>` at any point, and when
+  they do it supersedes the agent pass for that round.
+- **Never report the gate as having run when it did not.** If the agent pass
+  was substituted, or skipped, say so plainly in the user-facing summary.
 
-## The brief to carry into the subagent prompt
+### Per-agent mapping
 
-**Framing: this is a gate, not a collaborator.** The reviewer's job is to
-find what's wrong with the change, not to appreciate it. Don't soften
-findings to be encouraging, and don't reward effort.
+The properties above are stated as reviewer *capabilities*, not one agent's
+tool names, so the mechanism ports:
 
-**Findings only.** Never edit the tree, never apply a fix, never run
-`--fix`. Return a report for the dispatching session to triage.
+- **Claude Code** — a non-fork `Agent` dispatch at the routing table's
+  Premium tier.
+- **Cline** — dispatches subagents. Whether that subagent starts cold is
+  recorded in `docs/testing/agent-skill-resolution-probe.md`. Until it reads
+  `CLINE_SUBAGENT_COLD: yes`, a Cline-run pass is labelled a **self-run**,
+  not the independent gate: reporting a fork as the gate is exactly the
+  substitution the standing rule forbids.
+- **Any agent that genuinely cannot dispatch** runs the rubric in-session and
+  reports it as a **self-run pass, never as the independent gate** — the same
+  rule that forbids reporting a gate as having run when it did not.
 
-**Feed it the load-bearing claims, not just a diff.** Name the specific
-files, the exact measured numbers, and the technical assertion the change
-rests on — the things a shallow pass would take on faith rather than check.
-A reviewer handed only "review this PR" burns its budget re-deriving context
-the dispatching session already has; a reviewer handed the claims can spend
-that budget attacking them instead.
+## The tree check
 
-**Say what's already mutation-verified**, and by what proof — e.g. "guard X
-was confirmed to fail before the fix and pass after." That tells the
-reviewer which ground is covered so it hunts elsewhere instead of re-proving
-work that's already done.
+Before dispatching, capture:
 
-**Per finding, require all three:**
-- a **severity**;
-- a **`file:line`**;
-- a **concrete failure scenario** — specific inputs or state that produce a
-  specific wrong output. "This could be fragile" is not a finding; the
-  reviewer must show the break, not gesture at a risk.
+    git rev-parse HEAD && git status --porcelain
 
-**Split correctness bugs from cleanup nits — mandatory, not a nicety.**
-Every finding is labeled one or the other. This split is exactly what
-`model-routing`'s *Re-review trigger* reads: ≥1 actual correctness bug
-re-triggers a review once fixed and pushed; a pass with only cleanup-only
-findings, or none, does not. A report without this split cannot drive that
-loop — the dispatching session would have to re-derive it by re-reading
-every finding, which defeats the point of a structured report.
+Re-run both after the pass returns. **Any delta is a gate failure** — report
+it as such, do not absorb it. This is the one behavioural property of the
+pass verifiable from outside it, so it is the one that gets verified. The
+guard test cannot check it, and does not claim to.
 
-**"Found nothing" is a valid, expected outcome — say so explicitly in the
-brief.** A reviewer that believes it must produce findings to justify its
-own dispatch will manufacture them. A manufactured finding costs a needless
-re-review round and erodes trust in every report after it.
+## The PR comment
 
-## After the pass returns
+Every pass posts one summary comment on the PR the moment it returns,
+**before any fixes**, so the thread reads as found → fixed → re-verified in
+order.
 
-Triage per `model-routing`'s *Findings handling* and *Re-review trigger* —
-not restated here. **Never report the gate as having run when it did not.**
-If the user's own `/code-review` superseded this pass for the round, or the
-pass was substituted or skipped for any other reason, say so plainly in the
-user-facing summary rather than letting the checklist item read as
-satisfied.
+Format, following the three comments on [PR
+#2320](https://github.com/dudarenok-maker/Castwright/pull/2320):
+
+```
+## PR review — pass N (head <sha>, effort <level>)
+
+Scope: <files reviewed> · verified: <suite counts, typecheck> before
+adversarially probing <what>.
+
+### 🔴 Blocking — <claim>
+<concrete repro: exact input → wrong output, file:line>
+
+### 🟠 Significant — <claim>
+### 🟡 Minor
+### ✅ What is solid
+### Verdict
+```
+
+Two elements are required rather than stylistic:
+
+- **the head SHA in the heading** — it is what makes a comment interpretable
+  months later, when the branch has moved on;
+- **on a re-review, each prior finding's disposition** (`VERIFIED RESOLVED`,
+  or still open and why), so the thread is a chain rather than three
+  unrelated opinions.
+
+Severity maps onto the split [Triage](#triage) reads: 🔴 and 🟠 are
+correctness bugs and re-trigger a pass once fixed and pushed; 🟡 is cleanup —
+fixed this round, but not a re-trigger.
+
+### The reviewer posts its own comment
+
+The **reviewer posts its own comment directly** — `gh pr comment <number>
+--body-file <file>` — one hop, before returning its report to the dispatching
+session. Nothing sits between finding and record: the session does not relay
+it, and does not compare what gets posted against what the reviewer found,
+because nothing sits in between to compare.
+
+The reviewer's boundary is a prohibition, not a capability limit: **it must
+not modify tracked files.** That is enforced by [the tree
+check](#the-tree-check), not trusted.
+
+The session then confirms the comment actually landed (`gh pr view <number>
+--json comments`) before starting triage. A pass that returned a report but
+posted nothing is reported as a pass that did not complete.
+
+### A pass that finds nothing still posts
+
+"Found nothing" is an explicitly valid outcome, so it gets the same header
+and a `### ✅ No findings` body. Without this, the record cannot distinguish
+*reviewed and clean* from *never reviewed* — the single distinction the whole
+record exists to preserve.
+
+For the same reason, a docs-only PR — exempt from the pass entirely, per
+[Exemption](#exemption) — posts a one-line exemption note naming the
+file-set test that exempted it. Absence of a review becomes a recorded fact
+rather than a silence.
+
+**Re-review comments state deltas only** — each prior finding's disposition,
+plus anything new. They do not re-list the solid items. Three passes plus fix
+commits is already the cap; re-listing would make the thread unreadable at
+exactly the point it matters most.
+
+**Standing authorization.** Posting to a public PR is an outward-facing
+action. The repo owner has authorized it as a standing part of this process,
+recorded here, so the agent posts each round without pausing to ask.
+
+## Triage
+
+The full rubric for what the reviewer checks and how it must report a
+finding is [`references/reviewer-brief.md`](references/reviewer-brief.md).
+What happens to what comes back — the fix-now bar, the defect / chore /
+taste seam, the void deferral reasons, the design-pass carve-out, and one
+dispatched fix agent per finding with one paired test — is
+[`references/findings-triage.md`](references/findings-triage.md). Not
+restated here.
+
+## Re-review trigger and loop cap
+
+- **Re-review trigger**: only when the initial pass surfaced ≥1 finding that
+  is an actual correctness bug (wrong behavior, crash, security issue — not a
+  reuse/simplification/efficiency-only cleanup nit). Fixing and pushing those
+  re-triggers a pass. If the initial pass came back empty, or surfaced only
+  cleanup-only findings, fix-and-push (or push nothing) does **not**
+  re-trigger a re-review — re-running it in that case just burns tokens for
+  no new signal. This mirrors the spec/plan loop's severity-gated shape (see
+  [`model-routing`'s adversarial-review
+  loop](../model-routing/SKILL.md#mandatory-adversarial-review-specs--plans)),
+  rather than firing on every push. Re-review re-derives the effort level
+  from the PR's current commit set (per [Effort level](#effort-level))
+  rather than reusing the initial pass's tier — a fix commit can raise the
+  tier the same way any other commit would.
+- **Loop cap**: 2 re-review rounds, same numeric cap as the spec/plan loop
+  (initial pass + up to 2 re-review rounds, 3 total). Still tripping the
+  trigger after that stops the loop and hands it to the user — do not keep
+  looping automatically past the cap.
+- **Judgment-call carve-out**: a finding that requires a decision only the
+  user can make suspends the fix-and-re-review loop and routes through the
+  normal ask-first behavior in [CLAUDE.md's "Think before
+  coding"](../../../CLAUDE.md#think-before-coding) — shared with the
+  spec/plan review loop, see [`model-routing`'s judgment-call
+  carve-out](../model-routing/SKILL.md#judgment-call-carve-out-shared-by-both-review-loops).
+
+## Issue verification at PR creation
+
+- **Trigger**: every `gh pr create` for non-trivial work, including
+  bug-shaped work.
+- **Check**: the PR body must contain `Closes #NN` or `Refs #NN` referencing
+  an existing GitHub issue, outside of any inline-code/fenced-code span (a
+  backtick-wrapped `` `Closes #NN` `` does not actually auto-close on
+  GitHub — write it plain).
+- **Missing case**: auto-file a new issue capturing the work, then add
+  `Closes #NN` to the PR body — proceed without pausing to ask, in every
+  case including bug-shaped work (a deliberate, explicit override of
+  `CLAUDE.md`'s general "user files bugs" convention, scoped to this one
+  gate). Label per `CONTRIBUTING.md`'s actual two-shape convention:
+  - **Bug-shaped** (fixing existing broken behavior, no design decision):
+    standalone `bug` label, plain descriptive title. No `area:`/`type:`/
+    `moscow:`.
+  - **Backlog-shaped** (new/changed behavior): `area:<prefix>` +
+    (`type:feature` or `type:chore`). `moscow:` left unset for the user.
+    Use `type:chore` when the work matches this repo's commit-type-`chore`
+    shape (codegen, version bumps, tidy-up — no user-facing behavior
+    change); `type:feature` otherwise.
+- **Timing**: at PR creation — distinct from, and prior to, the mandatory
+  independent review above, which reviews code + docs combined once the PR
+  (and its issue link) already exist.
+- **Mechanical backstop**: `.github/workflows/pr-issue-link.yml` fails the
+  PR check if neither `Closes #\d+` nor `Refs #\d+` appears in the body
+  (outside code spans) — the one gate in this file with a real, external
+  enforcement, not just this convention. It does not check labeling or
+  whether the auto-file step above ran correctly, only that some issue
+  reference exists.
+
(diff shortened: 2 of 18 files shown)


dudarenok-maker:
## PR review — pass 1 (head 7f865b6d, effort high)

Scope: all 15 changed files · verified: `npm run test:hooks` 1372/1372 green, `node --test scripts/tests/verify-cache.test.mjs` 93/93 green, and the four load-bearing claims checked against `origin/main` before adversarially probing the anchor helper against the real corpus, the mirror script's guards under mutation, and the `.gitignore` change against the primary checkout's actual contents.

**Load-bearing claims — verdicts.** (1) **Effort ladder: HOLDS.** Word-level compare against `git show origin/main:.claude/skills/model-routing/SKILL.md`: `low`/`medium`/`high` criteria, the mixed-type rule and the `ultra` opt-in rule are unchanged; the only edits are "for every other PR" → "For every non-exempt PR" and an additive "state the level explicitly" sentence. The issue-verification section is byte-identical (normalized) to the removed one. (6) **`buildMirrorContent`: HOLDS.** The live mirror's byte 0 is `2d 2d 2d`, no BOM, em-dashes intact as `e2 80 94`. (3) **`.gitignore`: HOLDS ONLY FOR THE THREE PATHS YOU NAMED** — see 🔴 #2. (2), (4), (5): findings below.

---

### 🔴 Blocking — the reviewer brief mandates a `--body-file` and mandates a clean tree, and never connects the two

**Correctness bug.** `.claude/skills/pr-review-gate/references/reviewer-brief.md:90` tells the reviewer to post with `gh pr comment <number> --body-file <file>`. Line 106-109 then tells it "**Modify no tracked file.** Posting a comment is not a modification" and warns that the session diffs `git status --porcelain` before and after, "and any delta is reported as a gate failure." `SKILL.md:119` states the same rule harder: "**Any delta is a gate failure**."

Nothing in either file says where `<file>` may live.

Repro: a reviewer follows the brief literally, writes `review.md` in the worktree root, runs `gh pr comment 2339 --body-file review.md`, returns a clean report. Before: `git status --porcelain` → empty. After: `?? review.md`. The dispatching session applies SKILL.md:119 and reports **a gate failure on a review that found nothing wrong** — and, worse, the reviewer has by then already posted, so the record says "reviewed and clean" while the session says the pass is void.

The brief's own sentence is the trap: "Posting a comment is not a modification" is true of the *post* and false of the *body file the same paragraph requires*. Note the evidence that this is not hypothetical — my own dispatch prompt had to patch the gap by hand ("written to a file in the system temp dir (NOT inside the repo — that would dirty the tree)"). The rubric shipped incomplete and the dispatcher covered for it; the next dispatcher may not.

Fix: one clause in `reviewer-brief.md:90` — write the body outside the working tree (system temp). Cheap, and it is the single instruction that makes the tree check survive its own procedure.

---

### 🔴 Blocking — un-ignoring the whole skills tree makes an untracked, machine-local file an input to `test:hooks`'s cache hash

**Correctness bug.** `.gitignore:33-34` (`.claude/*` + `!.claude/skills/`) combined with the new `.claude/skills/**` glob at `scripts/verify-cache.mjs:153`.

The three paths the PR reasoned about are still ignored — verified with `git check-ignore -v --no-index` against this branch's `.gitignore`: `settings.local.json`, `settings.json`, `worktrees/foo`, `scheduled_tasks.lock` and `agents/foo.md` all hit `.gitignore:33:.claude/*`. But the negation is **tree-wide**, and the primary checkout has a skill the repo does not track:

```
C:\Claude\Projects\Audiobook-Generator\.claude\skills\run-app\SKILL.md
```

`git ls-tree -r HEAD -- .claude` on this branch lists only `model-routing/SKILL.md` and `pr-review-gate/{SKILL.md,references/*}`. `run-app` is untracked. And `git check-ignore --no-index .claude/skills/run-app/SKILL.md` under this branch's `.gitignore` → **not ignored** (exit 1, no rule).

Two concrete consequences, both new with this diff:

1. **The owner's primary checkout goes permanently dirty.** `git status --porcelain` in `C:\Claude\Projects\Audiobook-Generator` is empty today; after this merges it is `?? .claude/skills/run-app/SKILL.md`. That is the exact signal 🔴 #1 above and `SKILL.md:114-122` depend on being clean, and it arms the `git add -A` sweep hazard this repo has already been bitten by.
2. **The verify cache becomes machine-dependent.** `scripts/verify-cache.mjs:961` enumerates input files with `git ls-files --cached --others --exclude-standard` — `--others --exclude-standard` is precisely "untracked and **not ignored**". So `run-app/SKILL.md` now matches the new `.claude/skills/**` glob and is hashed into `test:hooks`'s input hash. Editing a local scratch skill busts the cache; more importantly the same commit hashes differently on the owner's box (has `run-app`) than in CI or a fresh clone (does not), and differently again on any machine with a different set of local skills. The cache's whole premise is that the hash is a function of the commit.

Fix has two defensible shapes and triage should pick one explicitly: (a) track `.claude/skills/run-app/SKILL.md` — it is a project skill describing this app and reads like it belongs in the repo; or (b) narrow the negation to the two directories that are actually tracked (`!.claude/skills/model-routing/`, `!.claude/skills/pr-review-gate/`), which keeps the PR-body's "don't re-arm the trap for every future file" property for those trees while not un-ignoring arbitrary local ones. Either is a one-line change; leaving it is not an option, because (2) is a real cache-correctness regression.

---

### 🟠 Significant — `githubAnchor()` is not GitHub's slugger: it trims, and it is ASCII-only

**Correctness bug.** `scripts/tests/review-gate-mechanism.test.mjs:176-183`.

GitHub's slugger (`github-slugger`) is `lowercase → strip punctuation/symbols → replace every space with a hyphen`. It does **not** trim. The helper adds `.replace(/^ +| +$/g, '')` at line 181, and restricts the keep-set to `\w`, which is ASCII-only in a non-`u` regex. Two divergences, both producing the false-positive class the helper's own doc-comment (lines 165-175) calls "worse than no guard: it trains its reader to 'fix' correct documents."

Measured against a faithful reimplementation over the real scan set:

```
"✅ What is solid"        impl="what-is-solid"   github="-what-is-solid"
"🔴 Blocking — <claim>"   impl="blocking--claim" github="-blocking--claim"
"🟡 Minor"                impl="minor"           github="-minor"
"Café & bar"              impl="caf--bar"        github="café--bar"
```

The leading-hyphen case is not exotic here: `SKILL.md:139-145` prescribes `### 🔴 Blocking`, `### 🟠 Significant`, `### 🟡 Minor`, `### ✅ What is solid` as *this repo's own* review-comment headings. The first time anyone promotes that shape into a real heading in `CLAUDE.md`, `CONTRIBUTING.md` or any `.claude/skills/**` doc and links to it with GitHub's correct `#-what-is-solid`, `npm run test:hooks` goes red on a link that works. The `\w` case fires on the first accented or Cyrillic heading — in a repo whose fixtures include `Ночной дозор`.

Compounding it: the unit test at line 225 is titled **"githubAnchor matches GitHub slugging, including runs of spaces"** and pins three ASCII cases with no leading punctuation. The title asserts a property the function does not have — the "test that over-claims" shape the brief itself catalogues at item 3.

Fix: drop the trim, and use `[^\p{L}\p{N}_\- ]+/gu` instead of `[^\w\- ]+`. Add `githubAnchor('✅ What is solid') === '-what-is-solid'` and a non-ASCII case to the unit test.

---

### 🟠 Significant — `headingAnchors()` harvests headings from inside fenced code blocks, so the new guard fails open on the very file it ships

**Correctness bug.** `scripts/tests/review-gate-mechanism.test.mjs:201-208`.

`stripFencedBlocks` is applied to the **source** side (line 269) with a "Measured:" comment justifying it — and not to the **target** side, which reads the raw file. Result: every `#`-prefixed line inside a fence becomes an accepted anchor.

Measured over the current scan set, **29 anchors that do not exist on GitHub** are accepted, including from this PR's own new file:

```
pr-review-gate/SKILL.md :: verdict
pr-review-gate/SKILL.md :: minor
pr-review-gate/SKILL.md :: what-is-solid
pr-review-gate/SKILL.md :: blocking--claim
pr-review-gate/SKILL.md :: pr-review--pass-n-head-sha-effort-level
CONTRIBUTING.md         :: 23 more, from the release-recipe fence
```

Repro: add `see [the verdict](.claude/skills/pr-review-gate/SKILL.md#verdict)` to `CLAUDE.md`. `npm run test:hooks` passes. The link 404s on GitHub. That is the precise failure the assertion's own header (lines 261-266) says it exists to prevent — "Presence of a word is not integrity of a link."

Fix: `headingAnchors()` should read `stripFencedBlocks(readNormalized(file))`, same as the source side.

---

### 🟠 Significant — the link regex requires `#`, and two genuinely dangling links sit unseen inside the guard's own scan set

**Correctness bug.** `scripts/tests/review-gate-mechanism.test.mjs:160` — `INTRA_REPO_ANCHOR_LINK` only matches `](path.md#anchor)`. A plain `](path.md)` is never resolved, even though the guard already has the file-existence check (line 272) that would catch it.

`CONTRIBUTING.md` is in the scan set (line 222), and it currently contains two broken file links:

```
CONTRIBUTING.md:414  [docs/superpowers/specs/2026-07-05-github-issues-kanban-design.md](superpowers/specs/2026-07-05-github-issues-kanban-design.md)
CONTRIBUTING.md:415  [docs/features/archive/241-github-projects-kanban-board.md](features/archive/241-github-projects-kanban-board.md)
```

Both are the "relative path written as if from `docs/`" bug the test's own comment (lines 214-216) attributes to historical plan docs — except these are in a governance doc the new guard actively scans and walks straight past. I resolved every file-only link in the scan set: these two are the only misses, so the fix is bounded and the guard goes green immediately after.

"It's pre-existing" is a void deferral reason per `findings-triage.md:43-44`, and the work surfaced it: the PR added the scanner that should have caught it.

Fix: make the `#anchor` group optional in the regex and skip the anchor check when it is absent; correct the two paths.

---

### 🟠 Significant — `assertNoBom` cannot fire, and a BOM'd reference file is silently relocated into the middle of the mirror

**Correctness bug.** `scripts/sync-agent-skills.mjs:90-94`, called at line 108. The script's own header advertises this as the guard for trap 3 ("No BOM. A BOM before `---` breaks frontmatter parsing").

It can never fail. `buildMirrorContent` returns either the canonical content (which begins `---`, since that branch is only taken when `startsWith('---\n')`) or `header + content` (which begins `<!--`). Either way `buf[0] !== 0xEF`. Executed:

```
references/findings-triage.md   assertNoBom fires? false   BOM relocated to byte 309
references/reviewer-brief.md    assertNoBom fires? false   BOM relocated to byte 308
SKILL.md                        assertNoBom fires? false   (line-109 check catches it)
first 8 bytes in all three: 3c 21 2d 2d 20 4d 49 52   ("<!-- MIR")
```

So: for `SKILL.md` the separate `startsWith('---\n')` check at line 109 does the work and `assertNoBom` is inert. For the two reference files **nothing fires at all** — a BOM at the head of `references/reviewer-brief.md` is written into the mirror at byte 308, mid-document, where it renders as a stray ZWNBSP in the rubric a reviewing agent reads.

This is catalogue shapes 3 and 6 from this PR's own rubric — a check that cannot fail, reporting success while doing nothing — shipped inside the PR that introduces the catalogue. Impact is modest; the inert guard is the finding.

Fix: check the BOM on the **source** bytes before building, not on the dest; or apply the line-109 first-line assertion to all three files.

---

### 🟠 Significant — the mirror guard's fail-open also swallows the only coverage `buildMirrorContent` has, and that part is not unavoidable

**Correctness bug.** `scripts/tests/review-gate-mechanism.test.mjs:291-322`.

You asked me to judge whether the fail-open is defensible. Splitting it in two:

- **Fail-open on *drift detection* is defensible.** The target is `$HOME/.agents/skills/`, genuinely absent in CI and on a fresh clone; the skip prints a reason; the comment forbids reporting green as "in sync". Agreed, keep it.
- **Fail-open on *the function's only test* is not.** `buildMirrorContent` is imported at line 44 and called at exactly one place — line 318, inside the block the skip returns before. Grep confirms no other importer. So on every CI runner and every fresh clone the new 131-line script has **zero executed assertions**.

Mutation-proven, not argued. I copied the branch's `sync-agent-skills.mjs` + guard test to a scratch tree, changed line 84 from `closeIdx + '\n---\n'.length` to `closeIdx + 1`, and ran with `$HOME` pointed at an empty dir:

```
--- MIRROR ABSENT (CI / fresh clone), MUTATED buildMirrorContent ---
[skip] no agent-store mirror at …\fakehome\.agents\skills\pr-review-gate
✔ the agent-store mirror matches its canonical source, when it exists
--- MIRROR PRESENT (this box), MUTATED buildMirrorContent ---
  AssertionError: SKILL.md has drifted from its canonical copy
```

A mutation that mangles every mirrored file's frontmatter splice merges green. `scripts/**` is in `test:hooks`'s globs, so the diff *schedules* the step — it just cannot fail it. That is CLAUDE.md's "New behaviour → ship paired automated test(s)" unmet, and it costs nothing to meet: `buildMirrorContent` is **pure**, takes a string, touches no filesystem. Three direct unit assertions (frontmatter file → `---` stays line 1 and the block appears once; non-frontmatter file → header first; a `---`-opening file with no closing delimiter → throws) run everywhere and close the gap without weakening the deliberate skip.

---

### 🟠 Significant — `SKILL.md` lets a flash-tier Cline pass be recorded as the Premium-tier independent gate

**Correctness bug (document).** `.claude/skills/pr-review-gate/SKILL.md:105-108` vs `SKILL.md:63-69` vs `docs/testing/agent-skill-resolution-probe.md:60-65`.

`SKILL.md:63` mandates "**Premium tier** (per model-routing's routing table)". The Cline row at :105-108 states exactly one condition: "Until it reads `CLINE_SUBAGENT_COLD: yes`, a Cline-run pass is labelled a **self-run**, not the independent gate." Nothing about tier.

The probe file reads `CLINE_SUBAGENT_COLD:  yes (self-reported)` and, on the next line, `CLINE_TIER_SELECTABLE: no — observed deepseek-v4-flash`. So an agent following `SKILL.md` finds the stated condition satisfied, dispatches `spawn_agent` on **deepseek-v4-flash**, and records it as the mandatory Premium-tier gate. The probe doc has the guardrail sentence ("this line exists so nobody records one as equivalent to an Opus-tier pass") — the runbook, which is the file that ports and the file that gets mirrored, does not.

Second, smaller edge in the same place: the condition gates on a value the probe itself labels unverified — "'my subagent starts cold' is a claim about runtime behaviour that was not independently reproduced" (:53-56). A rule written to stop a fork being reported as the gate is currently satisfied by the candidate's own self-report about its harness, which is exactly the evidence class that cannot distinguish the two.

Fix: carry the tier caveat into `SKILL.md:105-108` — a Cline pass is independent-but-flash-tier and is recorded as such, never as an Opus-tier pass — and mark the cold-start condition as self-reported where it is *used*, not only where it is *recorded*.

---

### 🟡 Minor

- **`sync-agent-skills.mjs:52-58` — the mirror's cross-references only resolve inside Castwright, while the provenance header claims otherwise.** The header says "Relative links below resolve against the REPOSITORY YOU ARE REVIEWING", and the comment above it justifies the `<repo>` placeholder by "the agent is reviewing a different one". But `SKILL.md:64` links `../model-routing/SKILL.md#routing-table` (not mirrored — the sync script's `FILES` list carries only the three `pr-review-gate` files), `SKILL.md:9` links `../../../CLAUDE.md#before-shipping-checklist`, and `:107` names `docs/testing/agent-skill-resolution-probe.md`. In a non-Castwright repo all three resolve to nothing, and the first is the one the reviewer needs to determine its own tier. Cleanup nit: say plainly that the mirror's cross-references assume the Castwright checkout, or mirror `model-routing/SKILL.md` alongside it.
- **`package.json:78` — `skills:sync` lands with no entry in CLAUDE.md's `## Commands` list**, which documents comparable maintainer commands (`backlog:sync`, `check:onbox-register`, `openapi:types`). It is a per-machine step a maintainer must remember to re-run after every edit under `.claude/skills/pr-review-gate/`; the only places that say so are the script header, the plan and the probe doc.

---

### ✅ What is solid

- **The move is genuinely lossless.** Effort ladder verified word-for-word against `origin/main`; issue-verification section normalized-identical. No threshold moved, and the `model-routing` stub keeps a real pointer.
- **The anchor-integrity assertion is the right instrument for the right failure.** The observation that a string-match on `pr-review-gate` stays green while the link 404s is exactly correct, and it caught four links where a manual read found one. Its two implementation holes (above) do not undermine the idea.
- **`buildMirrorContent`'s frontmatter-splice is correct and the artifact is clean.** Live mirror: `---` at byte 0, no BOM, UTF-8 em-dashes intact (`e2 80 94`) — the PowerShell-mojibake trap the header documents is genuinely avoided. Dropping the duplicated frontmatter block in favour of comparing against the builder was the right call over the marker-split.
- **Replacing three `extraFiles` literals with `.claude/skills/**` is correct**, and the two new negative-free `stepTouchedByDiff` cases (a reference file, a not-yet-existing file) pin it. All 93 pre-existing `verify-cache` assertions, including the `test:hooks` negatives at :422 and :593, still hold.
- **The PR body is unusually honest** — it names the fail-open, flags the two design decisions, declares the incidental fixes, and explicitly asks the reviewer to attack the weakest part. That framing is what made finding #6's split (defensible skip / indefensible coverage gap) reachable.
- **`Closes #2338` present, plain, outside code spans**; issue is open with `area:ops` + `type:chore`. Release-notes and on-box N/A both stated explicitly rather than omitted. `npm run test:hooks` 1372/1372 as claimed.

---

### Verdict

**Do not merge as-is.** Two 🔴 and six 🟠, all correctness bugs, none needing a design pass except the *shape* of the `.gitignore` fix (two defensible options, named in that finding — pick one, it is still a one-line change, not a deferral).

The pattern worth naming: **every finding is in an instrument this PR built to catch that same class of defect** — a rubric that mandates a file and a clean tree and never reconciles them; a slugger whose doc-comment defines false positives as disqualifying and which produces them on the repo's own heading style; a link guard that fails open on its own file; a BOM check that cannot fire; a drift guard whose skip also disables the only test of the code under it. That is catalogue item 9 ("the defect is in the instrument, not the code") applied to a PR whose deliverable *is* the catalogue. The runbook, the rubric and the guards are the right design; the implementations of the checks are one round short.

Per `findings-triage.md`, ≥1 correctness bug means a re-review is owed once these are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head af2a4212, effort high)

Scope: the fix delta `7f865b6d..af2a4212` (both fix commits), plus everything the fixes newly brought under version control — `.claude/skills/run-app/SKILL.md` entered the tree as part of finding 2's fix and is therefore in scope. · verified: `npm run test:hooks` (1378/1378 green); `node --test scripts/tests/review-gate-mechanism.test.mjs` (15/15); four **mutations** of the guard run from out-of-tree copies (fence-strip removed from `headingAnchors`; `githubAnchor` reverted to ASCII-only; `githubAnchor` with the trim re-added; link regex narrowed back to anchor-required) — each goes red, and the two slugger mutations go red on *both* the unit assertion and the integration link scan; a repo-wide fence-parity + dropped-heading sweep; `git check-ignore` on the paths `run-app` tells agents to write; `git ls-files --cached --others --exclude-standard` (`verify-cache.mjs:961`) confirming untracked-but-unignored files really are step inputs.

1. **VERIFIED RESOLVED** — `references/reviewer-brief.md:90-94` mandates `$env:TEMP`/`$TMPDIR`/`os.tmpdir()` and explains *why* (a scratch file in the worktree is read as a gate failure after the comment is already posted). This pass followed it.
2. **VERIFIED RESOLVED** — `.claude/skills/run-app/SKILL.md` is tracked (`git ls-files .claude/` lists it). `.gitignore:33-35` keeps `.claude/*` ignored with only `!.claude/skills/` negated: `settings.local.json`, `settings.json`, `worktrees/`, `scheduled_tasks.lock` all remain ignored, and the primary checkout's `git status --porcelain .claude/` is empty. Tracking was the right fix, not `git add -f` — `verify-cache.mjs:961` uses `--cached --others --exclude-standard`, so an untracked-but-unignored file genuinely was a step input.
3. **VERIFIED RESOLVED, and mutation-proven** — `\p{L}\p{N}` + no trim. Reverting to `\w`, or re-adding only the trim, each fails the unit test *and* the live link scan, so the assertion is not decorative. Matches `github-slugger` on the cases claimed (no trim, `_`/`-` kept, one hyphen per space).
4. **VERIFIED RESOLVED** — `headingAnchors` strips fences (`:219`). Sweep result: the 6 anchors it drops from `pr-review-gate/SKILL.md` are exactly the PR-comment template lines (`## PR review — pass N…`, `### Verdict`, `### ✅ What is solid`, …); the 23 it drops from `CONTRIBUTING.md` are all `#`-prefixed shell comments inside fenced release recipes; the 2 from `run-app` are bash comments. **No real anchor is lost**, and no link in the scan set depends on a dropped one.
5. **VERIFIED RESOLVED** — regex widened at `:378` with the anchor group optional; `CONTRIBUTING.md:414-415` now carry the `docs/` prefix. No false positives found: I checked `.md` links inside inline-code spans, inside HTML comments, and inside 4-space-indented blocks across the scan set — the three inline-code hits are real links whose *text* is backticked, and the one indented hit is list continuation, not a code block. Reference-style (`[x]: y.md`) links remain unmatched, but the scan set contains none.
6. **VERIFIED RESOLVED** — `assertNoBom` is now on the real production path: `syncAgentSkills → syncOneFile → assertNoBom(readFileSync(srcPath))` (`sync-agent-skills.mjs:110`), not only in the fixture test. The rationale comment at `:90-95` is accurate — with a BOM the source fails `startsWith('---\n')`, `buildMirrorContent` takes the no-frontmatter branch, and the BOM ends up *after* the header, where a byte-0 check on the written file can never see it.
7. **VERIFIED RESOLVED** — the four `buildMirrorContent` tests are pure string in/out, touch no filesystem, and run unconditionally. They are the only coverage not gated on `$HOME` having a mirror.
8. **VERIFIED RESOLVED** — `SKILL.md:104-121` two-condition table. Both cited markers exist verbatim in `docs/testing/agent-skill-resolution-probe.md:63-64` (`CLINE_SUBAGENT_COLD: yes (self-reported)` / `CLINE_TIER_SELECTABLE: no — observed deepseek-v4-flash`), and the table's "self-reported, not independently reproduced" hedge matches the probe rather than overstating it.
9. **VERIFIED RESOLVED in the emitted text, but it left a stale comment behind** — see 🟡 #2 below. The mirror at `~/.agents/skills/pr-review-gate/` is freshly regenerated and byte-matches `buildMirrorContent`, so the drift guard is not passing vacuously here. The "some point at sibling skills" claim is true (`../model-routing/SKILL.md` ×3), and that sibling is genuinely absent from the store, so the warning is warranted.
10. **VERIFIED RESOLVED** — `CLAUDE.md:424-432`, and it documents the per-machine/fail-open caveat rather than just listing the command.

### 🔴 Blocking

None.

### 🟠 Significant

**1. `run-app/SKILL.md` instructs agents to write two un-gitignored scratch files into the repo root — the exact failure the same commit patched `reviewer-brief.md` to prevent.** *(correctness bug)*

`.claude/skills/run-app/SKILL.md:60,72,79,102`.

`af2a4212` added `reviewer-brief.md:90` ("**NEVER inside the repository**") and, in the same commit, brought a skill into version control whose STEP 2 drops `verify-drive.mjs` at the repo root and writes `verify-screenshot.png` beside it. Neither is ignored — `git check-ignore -v verify-drive.mjs verify-screenshot.png` exits 1 for both.

Concrete failure: an agent invokes `run-app` to verify a UI change mid-PR. The driver throws — which is the case you run it *for* — so `node verify-drive.mjs && rm verify-drive.mjs` (`:79`) short-circuits on the `&&` and leaves the script; `verify-screenshot.png` is not removed by that line at all, because the very next instruction is "then Read verify-screenshot.png and LOOK at it". `git status --porcelain` now has two untracked entries, and this PR's own tree check (`pr-review-gate/SKILL.md`, "the tree check") reports the review as a gate failure — or a `git add -A` sweeps a screenshot into the commit. STEP 3's `rm -f` is framed as optional cleanup and is exactly the step a failed run skips.

Fix: `os.tmpdir()` for both paths (consistent with the rule this PR just wrote), or add both to `.gitignore`.

### 🟡 Minor

**2. `header()`'s comment still quotes — and defends as "the whole point" — the wording finding 9 deleted.** *(chore: a stale comment the change made false)*

`scripts/sync-agent-skills.mjs:46-51`.

The emitted header changed from `resolve against the REPOSITORY YOU ARE REVIEWING` to `resolve against a CASTWRIGHT CHECKOUT`, but lines 48-49 still read: *"the header travels with the mirror into every repo Cline reviews, and 'resolve against the repository you are reviewing' is the whole point"*. That sentence is now the argument **against** the string one line below it. Concrete: the next maintainer reads the rationale, sees the header disagreeing with it, and "restores" the text — re-arming finding 9. This is shape **#9** in this PR's own rubric, which makes fixing it owed in this round.

**3. `syncOneFile` writes the mirror before validating frontmatter — the class the BOM fix half-closed.** *(correctness bug)*

`scripts/sync-agent-skills.mjs:115-119`.

The BOM check was correctly hoisted *ahead* of the write (`:110`), and the new test at `review-gate-mechanism.test.mjs:365` asserts that contract by name: *"a BOM-poisoned canonical source must not produce a mirrored file at all."* Four lines below the write, the sibling failure path still validates after it.

Concrete: put a blank line or an HTML comment above `---` in `.claude/skills/pr-review-gate/SKILL.md` and run `npm run skills:sync`. `buildMirrorContent` takes the no-frontmatter branch, `writeFileSync` **clobbers the good `~/.agents/skills/pr-review-gate/SKILL.md`** with one whose first line is `<!--` — undiscoverable to Cline, per this file's own trap #1 — and only then throws. The operator saw a throw and reasonably concludes nothing was written. Shape **#10**: one instance fixed, the class left armed. Move the check above `writeFileSync` (it reads `mirrored`, not the file, so nothing blocks that).

**4. Dropping the trim was not paired with stripping a closed-ATX heading's trailing `#`s.** *(cleanup nit — latent, not currently firing)*

`scripts/tests/review-gate-mechanism.test.mjs:188-193`, with the heading regex at `:220`.

`/^#{1,6} +(.+?)\s*$/` captures `Verdict ##` intact, so `githubAnchor('Verdict ##') === 'verdict-'` where GitHub yields `verdict` — a **valid** link reported broken. I checked: the repo has zero closed-ATX headings today (`grep -rnE '^#{1,6} +.*[^#] #+\s*$'` over `CLAUDE.md`, `CONTRIBUTING.md`, `.claude/`, `docs/` → no hits), so this is latent. Raising it only because the helper's own docblock argues at length that a guard failing on valid input "is worse than no guard", and this is the one remaining input class where it does.

**5. PR title claims a single `docs` scope for a PR the body itself says is not docs-only.** *(cleanup nit)*

Title `docs(docs): …` vs. `scripts/sync-agent-skills.mjs`, `scripts/verify-cache.mjs`, `scripts/tests/*`, `package.json` — and the two fix commits are `fix(scripts)` and `fix(docs)`. `CONTRIBUTING.md:285` mandates the comma form (`docs(docs,scripts):`). Material rather than cosmetic here because `pr-review-gate/SKILL.md`'s own effort ladder keys off the PR's commit type/scope, and "any multi-scope PR" is its `high` trigger: a PR self-labelled single-scope routes a future re-review below the tier its own rule requires. This dispatch derived `high` from the file set, so nothing mis-fired *this* time.

**6. `run-app/SKILL.md` enters version control undeclared, and contradicts CLAUDE.md's Commands section.** *(chore)*

103 lines newly tracked, mentioned nowhere in the PR body — not in the Summary, not under "Also fixed, found in passing", which the rubric's own house-gate list requires ("incidental findings … declared in the PR body"). Separately, `:40` describes `npm run dev` as "frontend + server + sidecar", while `CLAUDE.md`'s Commands section attributes the three-in-one to `npm start` (`node scripts/start-app.mjs`) and `package.json` defines `dev` as `concurrently … dev:frontend dev:server`. `npm start` — the command CLAUDE.md says owns the sidecar lifecycle — is absent from the launch table entirely.

### Verdict

Ten prior findings, **ten verified resolved** at the code level, and the two guards that mattered are mutation-proven rather than merely green. The fix delta itself is sound: the slugger is now strictly closer to `github-slugger`, fence-stripping removes no real anchor, and the widened link regex produces no false positive in the scan set.

Two of the six new findings are in the hardening rather than the original change, which is where a pass-2 is expected to land: the BOM fix established "don't write on a failure path" as a contract and left the sibling path violating it (#3), and the provenance rewording left its own justifying comment asserting the deleted claim (#2).

**Not clean.** One 🟠 correctness bug (#1) — which re-triggers a re-review per `findings-triage.md` once fixed and pushed — plus five 🟡, of which #3 is also a correctness bug and #2/#6 are chores this round made owed. #1 is the one I would not merge without: this PR writes the rule and ships a tracked skill that breaks it.

dudarenok-maker:
## PR review — pass 3 (head 116e8162, effort high)

Scope: the fix delta `af2a4212..116e8162` (`run-app/SKILL.md`, `scripts/sync-agent-skills.mjs`, `scripts/tests/review-gate-mechanism.test.mjs`), plus a spot-check of the PR's only other runtime-code surfaces (`scripts/verify-cache.mjs` `STEPS[]`, `.gitignore`, `package.json`).

Verified (actually ran, not assumed):
- Launched `npm run dev:frontend` and **drove the run-app recipe end to end** against the live app. It works: `HEADING=Welcome back, Castwright | NAV_URL=…/#/voices`, screenshot written, and I opened the PNG — it renders the real Voices view.
- Both invocation claims in the new comment, empirically: plain `node "$VERIFY_DIR/verify-drive.mjs"` → `ERR_MODULE_NOT_FOUND: Cannot find package '@playwright/test' imported from C:\Users\…\Temp\tmp.…\verify-drive.mjs`; `node --input-type=module < …` from the repo root → resolves (`import.meta.url = file:///C:/Claude/Projects/wt-pr-review-skill/[eval1]`, i.e. CWD-based).
- `mktemp -d` → `/tmp/tmp.XXXXXXXXXX`, `cygpath -w` → `C:\Users\dudar\AppData\Local\Temp\tmp.…`. The doc's `%TEMP%` claim and the `$TMPDIR`-is-unset claim both hold. Git Bash **does** convert the `VERIFY_SCREENSHOT` env var to a Windows path on the way into node, so the screenshot lands where intended (I checked for a stray `C:\tmp\…` — none).
- The closed-ATX regex diffed old-vs-new over **every real heading in the scan set (7 files, 110 headings) and over a superset including `docs/**` + `.github/**` (792 files, 10,086 headings): 0 anchor differences**, plus 14/14 synthetic CommonMark closing-sequence cases correct.
- Red-phase for both new tests, by re-running them against a reconstructed pre-fix `syncOneFile` / pre-fix regex.
- `node --test scripts/tests/review-gate-mechanism.test.mjs` → 17/17 **before** running sync (so the drift guard had real evidence, not a fail-open: the `$HOME` mirror exists and all three files were already byte-identical to `buildMirrorContent`'s output). Then `npm run skills:sync` → 3 files written, re-ran → 17/17. `npm run test:hooks` → **1380/1380**.

1. **run-app temp-dir move — RESOLVED as to the reported defect, but the rewrite introduced two new ones.** The tree-dirtying is genuinely gone and the `--input-type=module` claim is true and necessary (verified above, both directions). See 🟠 A and 🟡 B below.
2. **`header()` comment — VERIFIED RESOLVED.** `sync-agent-skills.mjs:46-54` now describes the text the function actually emits ("Relative links below resolve against a CASTWRIGHT CHECKOUT"), and still explains why `<repo>` stays a literal placeholder.
3. **Frontmatter check hoisted — VERIFIED RESOLVED.** `sync-agent-skills.mjs:121-126`. Red-phase confirmed for the *claimed* reason: with the pre-fix ordering the throw assertion **passes** and it is specifically the "pre-existing mirror untouched" assertion that fails (mirror clobbered). No path now leaves a partial/stale mirror behind that wasn't already better than the alternative, and no later step assumes the skipped `mkdirSync` ran — `SKILL.md` is `FILES[0]`, so a throw aborts `syncAgentSkills` before any `references/*` destination is reached. One nit on the error text: 🟡 C.
4. **Closed-ATX slugging — VERIFIED RESOLVED.** `review-gate-mechanism.test.mjs:228`. `## Foo ##`→`foo`, `## C#`→`c`, and it also gets the harder cases right: `## Tracked by #`→`tracked-by` (stripped), `## Issue #2339`/`## Step 1 #2`/`## F# and C#` (not stripped — the run must reach EOL), `## Foo # Bar #`→`foo--bar`. Zero live behaviour change, as the comment claims.

### 🟠 A — `$VERIFY_DIR`'s value is never surfaced, so "Read the screenshot" is unactionable (correctness bug)

`.claude/skills/run-app/SKILL.md:68`, `:90-91`

`VERIFY_DIR="$(mktemp -d)"` is never echoed, and the node command's stdout prints only `HEADING=… | NAV_URL=… | ERRORS=…`. The agent therefore finishes STEP 2 without ever learning the directory `mktemp` chose. Line 91 then says `# then Read "$VERIFY_DIR/verify-screenshot.png" and LOOK at it` — an instruction it has no value to substitute.

It fails a second way even if the value *were* known: the Read tool takes a Windows absolute path, and the value is an MSYS path. Measured just now:

- `Read /tmp/tmp.2xDGiaEGve/verify-screenshot.png` → `File does not exist.`
- `Read C:\Users\dudar\AppData\Local\Temp\tmp.2xDGiaEGve\verify-screenshot.png` → renders fine.

So the skill's own stated regression net — "Look at the screenshot — don't trust the exit code alone" (`:107`) — is unreachable as written. Before this PR the path was the literal, known `verify-screenshot.png` at the repo root, so this is a regression introduced by the hardening. By pass 2's own bar ("a skill that documents a broken command is worse than one that documents a dirty-tree command") this is the same class of defect it was fixing.

Fix: echo the Windows path after `mktemp`, e.g. `VERIFY_DIR="$(mktemp -d)"; echo "VERIFY_DIR=$(cygpath -w "$VERIFY_DIR")"`, and have line 91 say to Read *that* printed path.

### 🟡 B — STEP 3's cleanup silently no-ops; the comment calling it "unconditional" is false (correctness bug)

`.claude/skills/run-app/SKILL.md:114` — `rm -rf "$VERIFY_DIR"   # unconditional — a failed run … must not skip this`

STEP 2 and STEP 3 are separate fenced blocks, i.e. separate Bash tool calls, and this harness does not persist shell state between calls (CLAUDE.md's Bash tool contract: "Working directory persists between calls, but … Shell state (env vars, functions) does not persist"). Measured, in a fresh call right after a real recipe run:

```
VERIFY_DIR in a FRESH Bash-tool call: [<UNSET>]
rm -rf "$VERIFY_DIR"; exit=0
temp dir after 'cleanup': /tmp/tmp.2xDGiaEGve   92K
```

`rm -rf ""` exits **0**, prints nothing, and deletes nothing — the driver script and the 70 KB screenshot survive. This is "success reported while doing nothing" wearing a comment that asserts the opposite. Pre-fix, `rm -f verify-drive.mjs verify-screenshot.png` used CWD-relative literals that worked from any later shell, so this is also a regression from the hardening. Harm is bounded (litter under `%TEMP%`, not a dirty tree), hence 🟡 rather than 🟠 — but the comment's guarantee has to go or become true.

Fix falls out of A: once the absolute path is echoed, STEP 3 references the printed literal (or use a stable prefix, `mktemp -d -t castwright-verify-XXXXXX` + `rm -rf "$TEMP"/castwright-verify-*`).

### 🟡 C — the hoisted frontmatter error now names the wrong file (cleanup nit)

`scripts/sync-agent-skills.mjs:122` — `throw new Error(\`${destPath}: frontmatter is not the first line\`)`

Naming `destPath` was accurate *before* the hoist: the mirror really had just been written with the bad content. Post-hoist nothing is written, so the message points an operator at `C:\Users\…\.agents\skills\pr-review-gate\SKILL.md` — a file that is either absent or a perfectly good mirror — while the file that actually needs fixing is the canonical source. Contrast `assertNoBom` two functions up (`:99-106`), which deliberately names the canonical source and says so. Concretely: drop the frontmatter block from `.claude/skills/pr-review-gate/SKILL.md`, run `npm run skills:sync`, and the failure text sends you to inspect a mirror that is byte-identical to the one that worked five seconds ago. Name `srcPath` (or both).

### Verdict

**Do not merge as-is — but this is the loop cap.** A (🟠) is a correctness bug and would normally re-trigger a pass; per `findings-triage.md` this is pass 3 of 3, so the loop stops here and hands to the user rather than auto-re-reviewing. All three findings are small, self-contained doc/message edits with no design decision in them — A and B are the same one-line root cause (a shell variable used across tool-call boundaries), C is one identifier. Fix those and the PR is good: the substantive machinery (sync ordering, BOM/frontmatter guards, anchor slugging, the `.claude/skills/**` cache glob, the `.gitignore` `.claude/*` + `!.claude/skills/` negation) all check out under direct measurement, and the run-app recipe itself genuinely drives the real app.

dudarenok-maker:
## PR review — pass 4 (head 529c9afd, effort high)

Scope: primary target `git diff 116e8162..529c9afd` (the unreviewed fix commit for pass 3's A/B/C), plus a whole-branch sweep of `git diff origin/main...HEAD` (17 commits) for cross-document contradictions the incremental passes could not see · verified: `npm run test:hooks` **1380/1380 pass, 0 fail**; `npm run skills:sync` (3 files written to `~/.agents/skills/pr-review-gate/`, tree stayed clean); the mirror-drift guard re-run against a real mirror after the sync (`review-gate-mechanism.test.mjs` 17/17); STEP 2's recipe executed verbatim against a live `npm run dev:frontend` on `:5173` with the project's Playwright chromium, then the printed Windows path `Read` in a **separate** tool call; STEP 3's cleanup executed verbatim in a **new** shell in three states (pasted / unpasted placeholder / empty).

- **A (🟠 `$VERIFY_DIR` never survived to the `Read` step) — VERIFIED RESOLVED.** Ran STEP 2 verbatim in one Bash call: it printed `VERIFY_DIR_WIN=C:\Users\dudar\AppData\Local\Temp\tmp.l1kJnMg9MB` and `VERIFY_SCREENSHOT_WIN=…\verify-screenshot.png`. A separate `Read` call on that Windows path resolved and rendered the screenshot (Voices view, heading `Every voice you've ever generated`). The `export` fix is real too — verified by the fact that `cygpath -w "$VERIFY_SCREENSHOT"` printed a full path rather than the bare `%TEMP%` root a prefix-assignment would have produced.
- **B (🟡 `rm -rf ""` silently no-op'd) — RESOLVED for the pasted case; the guard itself is blind to the likelier case.** With the printed value pasted into a fresh shell, `rm -rf "C:\Users\...\tmp.l1kJnMg9MB"` deleted the directory (`test -d` false from both the POSIX and Windows views — MSYS2's `rm` does accept a backslashed Windows path, which was worth confirming rather than assuming). See finding 2 for the residual.
- **C (🟡 frontmatter error named `destPath`) — VERIFIED RESOLVED** at `scripts/sync-agent-skills.mjs:122`; the check is reachable (a canonical source with no leading `---\n` falls into `buildMirrorContent`'s header-first branch, so `mirrored` starts with `<!--`) and fires before the write. See finding 3 for the missing paired assertion.

### 🟠 Significant — 1 (correctness)

**1. `run-app/SKILL.md` recommends a "mock mode" that this repo does not ship, and its own health criterion then fails.**
`.claude/skills/run-app/SKILL.md:9-11` — *"The frontend can run standalone in **mock mode** (`VITE_USE_MOCKS`, set in `.env.development`) with no server or sidecar"* — and `:39`, the table row *"Frontend only (mock mode — best for a quick verify) | `npm run dev:frontend`"*.

The tracked `.env.development` has read `VITE_USE_MOCKS=false` since 2026-05-12 (`6b4b2e51`), and there is no `.env.local` / `.env.development.local` in the tree to override it. So `npm run dev:frontend` runs against the **real** API with no server behind it.

Concrete failure, measured — I ran exactly what the skill prescribes:

```
HEADING=Welcome back, Castwright | NAV_URL=http://localhost:5173/#/voices | ERRORS=26
  console.error: Failed to load resource: the server responded with a status of 502 (Bad Gateway)
  console.error: [library] hydrate failed ApiError: Library scan failed (502)
  console.error: [voices] hydrate failed Error: Voice library fetch failed (502)
  … 23 more
```

`:115` lists `ERRORS=0` as a healthy run and `:119-121` says *"a … console error … = a failed verify."* An agent following this skill's own recommended row gets 26 of them on a perfectly healthy tree, and either reports a false regression or burns a debugging round. Three passes reviewed this file (pass 3 found two defects **inside** STEP 2/3) without anyone running STEP 1's recommended command end-to-end.

`CLAUDE.md:487` carries the identical false claim (*"`.env.development` sets the flag on"*). That one is pre-existing and untouched by this branch — which under `findings-triage.md`'s own void-deferral list ("it's pre-existing" / "the branch doesn't touch that file") is not a reason to leave it, and this PR is the one that newly commits a second copy of the same wrong fact into version control.

**Fix**: state what is true — mocks are off by default, so `dev:frontend` alone is only a shell/boot check and `ERRORS=0` needs `npm run dev` (or an explicit `VITE_USE_MOCKS=true` override) — in both files.

### 🟡 Minor — 3 (cleanup)

**2. STEP 3's new guard catches the case that cannot happen and misses the one that did.**
`.claude/skills/run-app/SKILL.md:131-136`. Line 131 seeds `VERIFY_DIR_WIN` with a **non-empty placeholder** (`"C:\Users\...\AppData\Local\Temp\tmp.XXXXXXXXXX"`), so `[ -z … ]` can only fire if someone deliberately pastes an empty string over template text — while the realistic failure (running the block verbatim without substituting, i.e. precisely what produced pass-3 finding B) sails straight past it. Measured, block copied verbatim:

```
=== case 1: placeholder left unpasted (block run verbatim) ===
block exit=0  <- silent success, nothing deleted, guard never fired
```

Same outcome as the defect this commit was written to fix: exit 0, no output, temp dir survives. Secondarily, when the guard *does* fire it `echo`s to stderr and still leaves the block at **exit 0** (verified), so a Bash tool call reports success — "errors loudly" is text, not a status. One line covers both: `[ -d "$VERIFY_DIR_WIN" ] || { echo "…not a directory -- refusing" >&2; exit 1; }`, which rejects the empty paste, the unpasted placeholder, and a typo'd path alike, and is a `-d` test rather than an enumeration of bad spellings.

**3. Pass-3 finding C's fix is not pinned by any assertion.**
`scripts/tests/review-gate-mechanism.test.mjs:447` asserts only `/frontmatter is not the first line/` — a regex naming no path. Reverting `sync-agent-skills.mjs:122` from `${srcPath}` back to `${destPath}` leaves all 1380 `test:hooks` tests green; nothing in the suite can distinguish the two. That is `reviewer-brief.md`'s own half-one gate #1 ("the paired test is a real regression test — red before the fix") unmet on a fix this PR shipped, and CLAUDE.md's "Bug fix → ship a regression test that fails before the fix and passes after." One-line fix: assert the thrown message contains `srcPath`.

**4. `findings-triage.md` revives a label CLAUDE.md explicitly retired.**
`.claude/skills/pr-review-gate/references/findings-triage.md:7` names its live section `## The fix-now bar`, and `pr-review-gate/SKILL.md:215` links to it by that name. CLAUDE.md's "Incidental findings" section says the opposite in as many words: *"This replaces the four-conjunct **'fix-now bar'** that earlier plans and code comments cite. That bar gated fixing; this one gates deferring … A historical doc saying something 'fails the fix-now bar' recorded a decision under the old rule — it is not precedent under this one."* Concrete scenario: an agent triaging findings greps `fix-now bar`, now hits a *live* governance section by that exact name, and cannot tell whether it is reading the retired four-conjunct bar (whose conjuncts **licensed deferral**) or its replacement — which is the one confusion CLAUDE.md's note exists to prevent. Rename the section to what it does ("Fix in this round", or reuse CLAUDE.md's own "report, fix, record"), in both files.

### Notes — not findings

- The PR body's test plan records `npm run test:hooks — 1372/1372`; HEAD measures **1380/1380** (eight assertions added by later commits). Worth correcting in the body before merge; not a tracked-file finding.
- Checked and clean, so stated rather than left silent: no remaining place in `run-app/SKILL.md` assumes a shell variable survives across tool calls; `cygpath` exists in this repo's Git-Bash environment (`/usr/bin/cygpath`, cygwin 3.6.7) and the skill is explicitly scoped to it (`:64`); no dangling `model-routing` anchors remain in `CLAUDE.md`, `CONTRIBUTING.md`, or any skill file (the surviving hits are inside the July-2026 plan doc, a historical record, and the new plan/spec quoting the anchors deliberately); `verify-cache.mjs`'s `.claude/skills/**` glob and the removed literals are both inside the single `test:hooks` STEPS entry (lines 76-263) and `ci-scope.mjs` consumes them through the same `stepTouchedByDiff` the two new tests exercise; `reviewer-brief.md`, `findings-triage.md` and `pr-review-gate/SKILL.md` agree on the 🔴/🟠-vs-🟡 split, the re-review trigger and the 3-pass cap.

### Verdict

**Merge after fixing all four — the three pass-3 fixes themselves are sound.** Finding 1 is the only correctness bug and it is two sentences of prose in two files; findings 2-4 are one line each. Under `pr-review-gate/SKILL.md`'s own trigger rule a 🟠 would re-open the loop, but the cap is already spent, which by that same section hands the decision to the repo owner — my recommendation is to land the four fixes and merge without a pass 5, since none of them touches executable behaviour.

dudarenok-maker:
## PR review — pass 5 (head 6704b4d3, effort high)

Scope: `git diff 529c9afd..6704b4d3` — the four pass-4 fixes across `CLAUDE.md`,
`.claude/skills/run-app/SKILL.md`, `.claude/skills/pr-review-gate/references/findings-triage.md`,
`scripts/tests/review-gate-mechanism.test.mjs` · verified by executing: both launch modes driven
in real chromium via STEP 2 verbatim; STEP 3's cleanup block run verbatim in four fresh shells;
the finding-3 assertion mutation-tested against `sync-agent-skills.mjs`; `npm run test:hooks`
(1380/1380), `npm run skills:sync`, mirror-drift guard re-run after. Tree left clean, HEAD unmoved,
dev server stopped.

### Pass-4 findings — disposition

1. **🟠 false "mock mode" claim — VERIFIED RESOLVED.** Measured, not read.
   `npm run dev:frontend` (tracked `.env.development`, `VITE_USE_MOCKS=false`, no server):
   `HEADING=Welcome back, Castwright | NAV_URL=…#/voices | ERRORS=26`, every one a 502 on a
   hydrate call (`[library] hydrate failed`, `[voices] hydrate failed`, `[voices] base catalog
   hydrate failed`) — exactly the shape the new paragraph documents. `VITE_USE_MOCKS=true npm run
   dev:frontend`: `ERRORS=0`. Both commands work verbatim as written. `CLAUDE.md:485-489` now reads
   correctly in context. The STEP 1 table's other two rows still match reality — checked
   `npm run dev` really is frontend + server + sidecar (`server/src/index.ts:314` creates the
   sidecar supervisor; `autoStartSidecar` defaults true at
   `server/src/workspace/user-settings.ts:309`), so that row's label is not stale. Not flipping the
   tracked flag was the right call. **But the scoping left a gap — see new finding 3.**
2. **🟡 inverted cleanup guard — VERIFIED RESOLVED.** Ran STEP 3's block verbatim in four fresh
   shells: verbatim placeholder → refuses, exit 1; real `cygpath -w` dir → `rm -rf` reached, exit 0,
   directory confirmed gone; realistic-but-nonexistent path → refuses, exit 1; empty string →
   refuses, exit 1. `[ -d ]` handles the backslashed Windows form correctly under Git Bash.
3. **🟡 unpinned assertion — VERIFIED RESOLVED, mutation-proven.** Reverted
   `` `${srcPath}: …` `` → `` `${destPath}: …` `` at `scripts/sync-agent-skills.mjs:122`, re-ran:
   RED, and red for the claimed reason (`AssertionError: syncOneFile did not throw naming the
   canonical source path (srcPath)`, `actual: Error: …\mirror\SKILL.md: frontmatter is not the first
   line`). Restored; md5 identical. The three-clause validator can't be satisfied by either path
   alone — `srcPath` and `destPath` are non-overlapping under the fixture tmpdir.
4. **🟡 retired "fix-now bar" label — PARTIALLY RESOLVED.** The heading is renamed and the
   retirement clause is correct, and the CLAUDE.md anchor
   (`#incidental-findings-report-fix-record`) resolves. But the rename left two live
   back-references to the retired name — one in the very file it edited. See findings 1 and 2.

### New findings

**1. 🟡 Minor — cleanup.** `.claude/skills/pr-review-gate/references/findings-triage.md:105` —
dangling self-reference the rename created. The re-review-trigger section still reads "*🟡-only
findings are still fixed this round **per the fix-now bar above***". At `529c9afd` that resolved:
the file's first heading was literally `## The fix-now bar`. After this commit there is no
"fix-now bar above" — a reader (or an agent) resolving the pointer lands on line 9, the paragraph
that says the fix-now bar was *retired* and *licensed deferral*. So the sentence now cites a
retired deferral-licensing bar as the authority for **fixing** 🟡 findings, which inverts it.
Concrete: a shipping session triaging a 🟡-only report greps `fix-now bar` in this file, reads
lines 9-12, and concludes the sentence at 105 rests on a superseded rule — the exact confusion
pass 4 raised. Fix: point it at this file's own §"Fix in this round" (one line).

**2. 🟡 Minor — cleanup.** `.claude/skills/pr-review-gate/SKILL.md:215` — same class, sibling site,
also made false by this commit. The Triage pointer enumerates `findings-triage.md`'s sections in
order: "the fix-now bar, the defect / chore / taste seam, the void deferral reasons, the
design-pass carve-out, and one dispatched fix agent per finding". Items 2-5 still match; item 1
names a section that no longer exists. Concrete: a dispatching session reads SKILL.md, opens
`findings-triage.md` looking for "the fix-now bar", doesn't find it, and either re-derives the rule
or assumes the reference doc is stale. One instance fixed, the class left armed. Both this and
finding 1 edit canonical skill files, so the fix must be followed by `npm run skills:sync` or the
mirror-drift guard goes red locally.

**3. 🟡 Minor — cleanup.** `.claude/skills/run-app/SKILL.md:59` (the STEP 1 copy-paste launch
block) — the fix scoped the `ERRORS=0` bar to mock mode / full `npm run dev`, but the block an
agent actually pastes was not touched and still hardcodes the bare, mocks-OFF form:
`npm run dev:frontend > /tmp/dev.log 2>&1 &`. That block starts no server, so the file's own
default path now lands in the one mode with **no mechanical pass criterion**: I measured
`ERRORS=26` following it verbatim, and the doc's answer is "that is not a failed verify". Concrete:
an agent asked to verify a frontend change follows STEP 0 → STEP 1 → STEP 2 top-to-bottom, sees 26
console errors and a library that never hydrates, and must either report a false failure or scroll
70 lines to the scoping note to wave all 26 away — including a genuine
`storage.getItem is not a function` hiding among them, which the doc still calls a failed verify but
which no longer trips any count. The intro's own recommendation is mock mode; line 59 should be
`VITE_USE_MOCKS=true npm run dev:frontend …` (or the block should say which mode it launches).

**4. 🟡 Minor — cleanup (weakest of the four; owner's call).**
`docs/features/199-advanced-settings.md:74` — `status: active` plan whose walkthrough still says
"Run in mock mode (`VITE_USE_MOCKS=true`, **default in dev**) via `npm run dev`". The command
itself is fine (it passes the flag), but the parenthetical is the same false claim pass 4 had
fixed in `CLAUDE.md` and the run-app skill, and it is the only surviving instance —
`grep -rn VITE_USE_MOCKS --include=*.md` shows every other doc passing the flag explicitly without
asserting a default. Concrete: an agent that internalises "default in dev" from this walkthrough
later runs a bare `npm run dev` for some other plan's acceptance and reads the 502 wall as a
regression. Two-word deletion.

**No correctness bugs (🔴/🟠).** Per `findings-triage.md`'s own rule, a 🟡-only pass does not
re-trigger a re-review round — these are fixed in this round and merged, not looped on.

Also checked and clean: `CLAUDE.md:485-489` reads correctly in surrounding context; the STEP 1
table matches every command it lists; the new `../../../../CLAUDE.md#…` link resolves (4 levels up
from `references/` is the repo root) and the anchor exists; no `](#the-fix-now-bar)` link anywhere,
so the link guard's silence is correct rather than fail-open; `npm run test:hooks` 1380/1380 pass;
the `$HOME` mirror was already re-synced and `skills:sync` is idempotent against it.

### Verdict

**Merge after folding the four 🟡 doc fixes** — all one-line edits, none of them re-triggers a
review pass. No blocking or significant defect found in this commit; findings 1-3 are stale
references this commit itself created, which the repo's own report-fix-record rule puts in scope
for this round.

Issue #2347, opened 2026-08-13
srv-33 — workspaceDirOverride outranks WORKSPACE_DIR, so one Settings save would defeat every worktree's isolation at once


dudarenok-maker:
## What

`server/src/workspace/paths.ts:35-39` resolves the workspace root in this precedence:

1. `workspaceDirOverride` from the resolved `USER_SETTINGS_PATH` (`~/.castwright/user-settings.json`, plan 122 shared-settings location)
2. `WORKSPACE_DIR` env var (i.e. `server/.env`)
3. built-in `../castwright-workspace` default

Note what sits at the top: a **single, shared, cross-checkout file in the user's home directory**, which the Settings UI writes.

#2345 has just made `scripts/wt-new.mjs` provision each new worktree with its own `server/.env` carrying an isolated `WORKSPACE_DIR`, on the repo owner's explicit ISOLATE decision — two servers on two branches must never mutate the same `cast.json` / `state.json`, because the cast-lock discipline is per-process-tree with no cross-checkout channel.

Because the shared override outranks the env var, **any future save of a custom workspace path in the Settings UI would silently override that isolated `WORKSPACE_DIR` in every worktree simultaneously** — collapsing every checkout back onto one shared workspace and reinstating exactly the concurrent-mutation hazard the ISOLATE decision was made to prevent. Nothing would warn; `WORKSPACE_SOURCE` would just read `'override'`.

**Currently inert.** `workspaceDirOverride` is `null` on this box (verified by reading the live file), so #2345's isolation works as intended today. This is a latent gap, not a live failure — which is precisely why it is worth pinning before someone uses that Settings field.

Found while verifying the #2345 fix.

## The decision owed

**Should a per-worktree `WORKSPACE_DIR` outrank the shared per-user `workspaceDirOverride`, or should the shared setting keep winning?**

- *Env wins when explicitly set* — restores worktree isolation and matches the general convention that a more-local config beats a more-global one. But it silently changes the meaning of an existing, shipped, user-facing setting, and someone who set a custom workspace deliberately would find some checkouts ignoring it.
- *Shared setting keeps winning* — preserves current behaviour and the plan-122 intent that this is one machine-wide library. Then `wt-new`'s isolation guarantee needs to be honest about being conditional, and arguably the Settings UI should say what it overrides.
[...]

Issue #2221, opened 2026-08-07
fe — a failed sidecar restart looks like it succeeded: no catch, spinner clears in finally


dudarenok-maker:
## What

`handleRestart` in `src/views/advanced.tsx` awaits `dispatch(restartSidecar()).unwrap()` inside an async arrow passed to `onClick`, with no `.catch`. The button's spinner is cleared in a `finally` block, so it clears regardless of outcome.

Net: clicking **Restart sidecar** when the request fails looks exactly like a success — the spinner stops, the banner returns to normal, and nothing indicates the sidecar was never restarted. The failure surfaces only as an unhandled promise rejection in the browser console.

## Why it is filed rather than folded into #2209

#2209 fixed the same defect class for config saves — including the Revert path, which was folded in during review because Revert *is* a config save and shares the server's error shape.

This one is not. `restartSidecar` is a different subsystem with a different user story, and there is no `PUT /api/config` error body of the shape `describeConfigSaveError` parses, so it cannot reuse that surface as-is. Deciding where a failed restart should be reported — the toast stack, the TTS notice banner, or inline on the restart control — is a design question rather than an obvious fix, which is why it is not a fix-now.

## Options

1. **Toast** via the existing notifications slice — cheapest, consistent with how #2209 reports non-row-attributable failures, and a restart failure is genuinely page-level rather than row-level.
2. **Inline on the restart control**, next to the button that failed.
3. **The TTS notice banner** (`src/components/tts-notice-banner.tsx`), which already reports sidecar state and is arguably where a user would look after a restart.

Option 1 composes with what #2209 just built. Option 3 is where the information belongs if the user's mental model is "the sidecar is a thing with a state".

## Acceptance

- A failed sidecar restart tells the user it failed, and the spinner state does not imply success.
- No unhandled promise rejection.
- A paired test, mutation-verified — reverting the fix must fail it.

## Key files

- `src/views/advanced.tsx` — `handleRestart`
- `src/store/notifications-slice.ts` / `src/components/toast-stack.tsx` — the toast path #2209 uses
- `src/components/tts-notice-banner.tsx` — option 3

*Benefit (user):* a restart that silently did nothing stops looking identical to one that worked.

Found by the implementer of #2209 and confirmed by the coordinator. Refs #2209.

PR #2348, opened 2026-08-13
build(mocks,scripts,docs): add npm run dev:mock and correct the mock-mode docs

### LINKED ISSUE #2343: ops-46 — No supported way to run the frontend in mock mode; docs still claim it is the dev default
## What

`.env.development` has set `VITE_USE_MOCKS=false` since 6b4b2e51 (2026-05-12) — the commit that added the local analysis server. That flip was deliberate and correct: `npm run dev` should drive the real backend.

Two things were never updated to match.

**1. There is no mock-mode dev command at all.** No `dev:mock` script, no `.env.mock`. The only env files setting `VITE_USE_MOCKS=true` are `.env.e2e` (owned by Playwright's `webServer`, port 5174) and `.env.marketing` (owned by `playwright.marketing.config.ts`). Neither is a hand-runnable dev command.

That matters because ~72 doc files open their manual walkthrough with "Run in mock mode (`VITE_USE_MOCKS=true`)". Those walkthroughs specify a state **no supported command can reach**. Every mock-mode acceptance walkthrough in the repo is unrunnable as written.

**2. Nine doc sites assert mock mode is the dev default** — false for three months:

- `CLAUDE.md:478` — "`.env.development` sets the flag on"
- `docs/features/199-advanced-settings.md:74` — "default in dev"
- `docs/features/236-prosody-review-progress-detail.md:205`
- `docs/features/266-fs35-per-chapter-detect-emotions.md:138`
- `docs/features/278-cast-character-identity.md:503` — "i.e. plain `npm run dev`"
- `docs/features/92-manuscript-virtualisation.md:54`
- `docs/features/archive/246-fs51-qa-report.md:261`
- `docs/features/archive/91-cast-drift-consolidation.md:73`
- `docs/superpowers/plans/2026-06-18-app-17-deeplink-pairing-launch.md:23`

One is worse than stale — it is an **inverted safety warning**. `docs/testing/fs38-wave3-onbox-acceptance.md:210` reads "⚠️ **Do NOT run in mock mode.** `.env.development` sets `VITE_USE_MOCKS` on for …", steering the reader away from a hazard that no longer exists while implying dev is mocked when it is not.

## Acceptance

- `npm run dev:mock` starts the app with mocks on, reproducing the pre-2026-05-12 documented state (frontend in mock mode **plus** the server — see Key files for why it is not frontend-only).
- `.env.mock` sets `VITE_USE_MOCKS=true`; per-worktree `.env.local` port overrides still apply under the new mode.
- The nine false-default claims above state what is actually true.
- The ~72 "Run in mock mode" preconditions name `npm run dev:mock` so they are followable.
- A paired automated test locks the env/script wiring so the three mock env files cannot silently disagree again.

## Key files

- `package.json` (`dev`, `dev:frontend`), `.env.development`, `.env.e2e`, `.env.marketing`, `vite.config.ts`
- **Why not frontend-only:** ~25 components call `fetch('/api/…')` directly rather than going through the `api.*` seam (`ollama-install`, `coqui-install`, `kokoro-install`, `qwen-install`, `whisper-install`, `venv-bootstrap`, `model-pull-status`, `blocker-fix-action`, `mini-player` listen-stats). Those 502 without a server even with mocks on. Tracked separately.

## Benefit

*Benefit (technical):* restores the ability to actually run the ~72 mock-mode acceptance walkthroughs the repo already documents, and stops CLAUDE.md teaching a dev default that has been wrong since May.

### LINKED ISSUE #2344: fe-52 — two components duplicate api.* wrappers with direct fetch('/api/...') calls
## What

CLAUDE.md states the convention: *"Mocks behind `VITE_USE_MOCKS` — `src/lib/api.ts` exports `api = USE_MOCKS ? mock : real`. Components only ever import from `api.*`; they never know which is which."*

That has drifted. Roughly 25 call sites bypass the seam and call `fetch('/api/…')` directly:

- `src/components/blocker-fix-action.tsx` (×4, incl. `/api/sidecar/restart`)
- `src/components/coqui-install.tsx`, `kokoro-install.tsx`, `qwen-install.tsx`, `ollama-install.tsx` (×5), `whisper-install.tsx`
- `src/components/venv-bootstrap.tsx`, `model-pull-status.tsx` (×3)
- `src/components/mini-player.tsx` (`listen-stats` beacon)

Consequence: with `VITE_USE_MOCKS=true` these components still hit the network, so mock mode is not actually server-free. Found while adding `npm run dev:mock` (#2343), where it forced that command to start the server rather than run Vite alone.

## Why this is filed rather than fixed in the same round

Per CLAUDE.md's incidental-findings protocol, the one thing that defers a finding is that it needs a design pass. **The decision owed:** whether these install/setup/status endpoints get mock implementations behind `api.*` at all, and if so what they return — an install job that never completes, one that succeeds after N polls, or a per-endpoint scripted timeline. That is a behaviour choice with more than one defensible outcome and it shapes every mock-mode walkthrough that touches the setup wizard, so it is not a mechanical lift.

The narrower alternative — leave them on real `fetch` and document that mock mode needs the server up — is also defensible and is what #2343 ships as the interim state.

## Acceptance

- A decision is recorded on the question above.
- If mocks are added: each endpoint has a mock impl behind `api.*`, components import from `api.*` only, and a guard test fails on a new direct `fetch('/api/…')` under `src/components/`.
- If not: CLAUDE.md's convention line is amended to state the documented exception set, so the convention matches reality.

## Benefit

*Benefit (architectural):* either restores the single-seam invariant the codebase claims, or makes the real exception explicit instead of leaving a convention that ~25 sites quietly violate.

### LINKED ISSUE #2345: ops-47 — wt-new assigns a server port it never provisions: every worktree's server binds :8080 and runs on defaults
## What

`scripts/wt-new.mjs` exists to let parallel sessions run `npm run dev` "without colliding on ports" (its own header, lines 18-19). For the server leg it does not achieve that.

- `wt-new.mjs:227` writes **only** `<worktree>/.env.local`, carrying `VITE_PORT` / `PORT` / `VITE_API_PORT` / `LOCAL_TTS_PORT` / `PLAYWRIGHT_PORT`.
- The server reads its config from **`server/.env`** — `server/src/index.ts:11` imports `./load-env.js`, which runs `process.loadEnvFile('.env')` relative to `server/`. Root `.env.local` is a Vite-convention file; nothing loads it into the server's `process.env`.
- `server/src/index.ts:88` is `const PORT = Number(process.env.PORT ?? 8080)`.
- The primary checkout has an 18 KB `server/.env`. **No worktree has one** — verified across `wt-2288-m2-impl`, `wt-2315-reopen-bound`, `wt-dev-mock-command`.

`server/src/load-env.ts` tolerates the missing file, warns, and continues on defaults. So in every worktree:

1. **The server binds `:8080`**, not its assigned slot port — colliding with the primary checkout's server and with every other worktree's. Whichever binds first wins; the others get `EADDRINUSE`, or worse, a *different* checkout's frontend silently proxies into the surviving server, which is running another branch's code.
2. **The frontend proxies to the assigned `VITE_API_PORT`** (e.g. `:8310`) because Vite *does* read `.env.local` — so nothing answers it. `npm run dev` in a worktree has a broken `/api` proxy today.
3. **The server runs on defaults for everything else**, which `load-env.ts`'s own warning spells out: `WORKSPACE_DIR`, `GEN_WORKERS`, `GPU_VRAM_BUDGET` are not applied. The `WORKSPACE_DIR` default is exactly the "silent stale workspace root" bug `server/src/index.ts:5-10` was written to prevent — a worktree server recreates `audiobook-workspace/` inside the repo.
4. **`wt-new` prints a claim that is false**: `[wt-new] npm run dev  # frontend :5403, server :8310`.

Found while verifying `npm run dev:mock` for #2343. Pre-existing and identical for plain `npm run dev` — not introduced by that branch. Mock mode masked it, since mock mode never proxies for the app's rendered data.

## The decision owed (why this is filed rather than fixed in the same round)

The port half is mechanical. The rest is not, and they cannot be separated because the fix is "what does `wt-new` put in a worktree's `server/.env`":

**Should a worktree share the primary checkout's `WORKSPACE_DIR`, or get its own?**

- *Share it* — a worktree has real books to test against immediately, but two servers on two different branches mutate the same `cast.json` / `state.json` under the same workspace, concurrently, with no lock between checkouts. The cast-lock discipline in CLAUDE.md is per-process-tree, not cross-checkout.
- *Isolate it* — safe, but a fresh worktree has an empty library, so any acceptance walkthrough needing a real book has to import one first.

Secondary, dependent on the above: whether `wt-new` copies the primary's `server/.env` wholesale (which would replicate `GEMINI_API_KEY` and other secrets into every worktree) or generates a minimal one carrying just the slot's ports.

## Acceptance

- A worktree's server binds its assigned slot port; `npm run dev` in a worktree has a working `/api` proxy end to end.
- The workspace-sharing decision above is recorded and implemented.
- `wt-new`'s printed "server :NNNN" line is true.
- A test covers the provisioning so this cannot regress silently.

## Key files

`scripts/wt-new.mjs` (:36-43 BASE_PORTS, :101-102, :166, :178, :227), `server/src/index.ts` (:5-11, :88), `server/src/load-env.ts`, `vite.config.ts` (proxy target from `VITE_API_PORT`).

## Benefit

*Benefit (technical):* makes the parallel-worktree workflow that CLAUDE.md mandates for all non-trivial work actually isolated, instead of quietly funnelling every worktree's server onto one shared port and one shared default workspace.

dudarenok-maker:
## Summary

`.env.development` has set `VITE_USE_MOCKS=false` since `6b4b2e51` (2026-05-12), the commit that added the local analysis server. That flip was correct — `npm run dev` should drive the real backend — but nothing else was updated to match, and the gap went unnoticed for three months.

**There was no way at all to run the frontend in mock mode.** No `dev:mock`, no `.env.mock`. The only env files setting the flag true are `.env.e2e` and `.env.marketing`, both owned by Playwright configs, neither hand-runnable. Meanwhile ~70 doc files open their manual acceptance walkthrough with "Run in mock mode (`VITE_USE_MOCKS=true`)" — specifying a state no supported command could reach.

**Nine doc sites asserted mock mode was the dev default.** One was worse than stale: `docs/testing/fs38-wave3-onbox-acceptance.md` §2.1 was a *safety warning* whose premise had inverted, steering the reader away from a hazard that no longer existed.

### What landed

- **`npm run dev:mock`** — backed by a new `.env.mock`. Starts frontend **and** server, mirroring `npm run dev`, because ~19 install/detect/provisioning components call `/api/…` directly and have no mock counterpart. `npm run dev:frontend:mock` is the frontend-only variant. Verified live: `MODE: "mock"`, the worktree's assigned ports still honoured (`.env.mock` wins the flag, `.env.local` keeps the ports), and zero dynamic `/api` requests.
- **Doc pass** — the 9 false-default claims corrected, the fs38 safety warning rewritten so its intent survives on a true premise, and ~60 further preconditions across 51 files pointed at the runnable command. 29 occurrences deliberately **left**: behaviour descriptions, e2e-harness prose, and historical records — a release note is a record, not an instruction.

### Reconciled with a concurrent lane

`main` merged #2339 (`6704b4d3`) mid-flight, which found the same false CLAUDE.md claim and fixed it — but prescribed `VITE_USE_MOCKS=true npm run dev:frontend`. That command **fails on PowerShell**, this project's documented primary shell (verified: `The term 'VITE_USE_MOCKS=true' is not recognized`), and being frontend-only it reproduces the 502 storm this PR exists to remove. `main`'s correction of the underlying claim is kept; its command is replaced by `npm run dev:mock`, and the five occurrences it published in `.claude/skills/run-app/SKILL.md` are corrected too.

That collision also explains why this PR initially showed no CI: it was `CONFLICTING`, and GitHub skips `pull_request` workflows when it cannot build a merge commit. Resolved.

- **`Closes #2344`** — `whisper-install.tsx` and `blocker-fix-action.tsx` each re-implemented a request that already had **both** real and mock implementations wired into `api` (`api.getConfig`, `api.restartSidecar`). Both now route through the seam. A static guard (`src/components/api-fetch.guard.test.ts`) pins the current direct-`fetch('/api/…')` surface by file **and count** across `src/{components,views,modals,routes,hooks}`. CLAUDE.md's convention bullet claimed "Components only ever import from `api.*`", an overclaim; it now states the true bounded exception set.
- **`Closes #2345`** — `scripts/wt-new.mjs` assigned each worktree a server port it never provisioned, so **every worktree's server bound `:8080`** while its frontend proxied elsewhere. It now generates a minimal `server/.env` with the slot's `PORT` and an isolated `WORKSPACE_DIR`, copying nothing from the primary checkout so no secrets propagate.

### Not fixed here

- **`Refs #2347`** — `workspaceDirOverride` outranks `WORKSPACE_DIR`, so a future Settings save would defeat every worktree's isolation at once. Inert today. Needs a precedence decision on a shipped setting.
- **`Refs #2349`** — `model-manager.tsx:771` calls `api.restartSidecar()` from a fire-and-forget callback, so a throw is an unhandled rejection nobody sees. Pre-existing and systemic (the Load/Stop pills share the shape); choosing the surface across that class is a UX decision.

## Test plan

- **New:** `scripts/tests/dev-mock-command.test.mjs`; `src/components/api-fetch.guard.test.ts`; `src/lib/api.restart-sidecar.test.ts`.
- **Extended:** `scripts/tests/wt-new.test.mjs`, `scripts/tests/verify-cache.test.mjs`, `whisper-install.test.tsx`, `blocker-fix-action.test.tsx`.
- **Every new test was mutation-verified red-then-green**, each invariant broken individually and restored. Two tests that could not fail were found by review and fixed: the `refetch(` exclusion test (vacuous fixture) and the whisper-install swallow test (red only via Vitest's global unhandled-rejection reporter, not its own assertions).
- `npm run verify:fast:branch` green with every leg genuinely re-run post-merge, not taken from cache: lint, typecheck, config:check, test:hooks (1402), check:budget-poll, test (4773), test:server (7545), build. `test:sidecar` SKIPs by design on an unbootstrapped venv.

**Not applicable, stated rather than skipped silently:** no regression plan under `docs/features/` (localized tooling + docs — the issues plus paired tests are the spec); no on-box acceptance row (nothing here needs real hardware to prove); no `INDEX.md` change (no plan added or moved). **Release notes cover only #2344**, the sole user-visible delta: #2343 (`dev:mock`) and #2345 (`wt-new`) are developer tooling with no user- or operator-visible effect.

## Review

Two independent Premium-tier passes. Pass 1 found 1 blocking + 3 significant + 9 minor; pass 2 verified all 13 resolved and found 1 significant + 7 minor, all now closed. The blocking finding was a real regression this PR introduced — the restart error card rendered raw JSON — caught before merge.

Closes #2343
Closes #2344
Closes #2345
Refs #2347
Refs #2349


dudarenok-maker:
## PR review — pass 1 (head 4613ca7b, effort high)

Scope: whole diff — `.env.mock` + `package.json` scripts, `vite.config.ts` env precedence, both component reroutes and their `api.*` wrappers (real **and** mock), `src/components/api-fetch.guard.test.ts`, `scripts/wt-new.mjs` + its tests, and the 66-file doc sweep · verified: `npm run lint` clean, `npm run typecheck` clean, full frontend Vitest **4765/4765** across 334 files, `npm run test:hooks` **1382/1382**, `node --test scripts/tests/{wt-new,dev-mock-command}.test.mjs` **51/51** — before adversarially probing the guard's matcher, the `test:hooks` scope gate, the `server/.env` generator against what the server actually reads, and the restart-error path against the real route.

Working tree untouched throughout; every mutation reverted and `git status --porcelain` re-checked empty after each.

---

### 🔴 Blocking — the `blocker-fix-action.tsx` reroute is a user-visible **regression**, and both release-notes entries claim the opposite

`POST /api/sidecar/restart` (`server/src/routes/sidecar-health.ts:672-729`) has **five** failure branches and every one returns a 4xx/5xx **with** a populated `{ok:false, error:"<sentence>"}` body. There is no HTTP-200-with-`ok:false` branch at all.

- **Old** (`main`, `blocker-fix-action.tsx:158-159`): raw fetch → `res.json()` parses the 409 body fine → `throw new Error(body.error)` → the card renders the server's own sentence.
- **New** (`blocker-fix-action.tsx:166` → `src/lib/api.ts:8793-8799`): `realRestartSidecar` throws on `!res.ok` using `(await res.text())`, i.e. the **raw JSON**, wrapped.

Repro — `autoStartSidecar` off (or server still booting), user clicks "Reset & restart voice engine" on the setup/status blocker card. Server answers `409 {"ok":false,"error":"No active supervisor — sidecar auto-start is disabled or the server is still booting."}`. Measured, not reasoned (stubbed `fetch`, real `api.restartSidecar`, jsdom):

```
OLD (main) >> No active supervisor — sidecar auto-start is disabled or the server is still booting.
NEW (PR)   >> Sidecar restart failed (409): {"ok":false,"error":"No active supervisor — sidecar auto-start is disabled or the server is still booting."}
```

This is precisely the defect `docs/release-notes-next.md` already records as **#2221 B1** — "every rejected Revert rendered the raw JSON-escaped wrapper" — reintroduced on the one call site that did **not** have it. `src/views/advanced.tsx:40-57` is the pre-existing sibling that already renders the wrapper, and its own comment concedes the route "has no such guarantee"; the correct move is to make `realRestartSidecar` unwrap `{error}` (as `configApiErrorMessage` does for `/api/config`), not to align the good site down to the bad one.

Both new release-notes lines assert the inverse and are false as written:
- `RELEASE_NOTES.md:123` — "That notice used to read the same generic line every time; it now shows the sidecar's own explanation." It showed the sidecar's own explanation **before**; it now shows a wrapper plus escaped JSON.
- `docs/release-notes-next.md:359` — "an HTTP-error response never reached `realRestartSidecar`'s own formatted … message; it now does" — true, but framed as the improvement it is not.

---

### 🟠 Significant — the code comment and two of the four new tests pin a server state that cannot occur

`src/components/blocker-fix-action.tsx:158-165` asserts "The server's in-band failure shape (**HTTP 200, body.ok === false**) still surfaces via this same body.ok check". No branch of `sidecar-health.ts:672-729` returns 200 with `ok:false`, and `mockRestartSidecar` (`src/lib/api.ts:8677-8680`) unconditionally returns `{ok:true}`. So line 167 (`if (!body.ok) throw new Error(body.error ?? 'Restart failed.')`) is dead in **both** modes.

Consequently `blocker-fix-action.test.tsx:191-208` ("body.ok === false surfaces body.error verbatim") and `:210-216` ("falls back to `Restart failed.`") pass only because `vi.spyOn(api,'restartSidecar').mockResolvedValue({ok:false,…})` manufactures a resolution shape neither implementation can produce. They are green regardless of production behaviour. The PR body's claim that "the existing … in-band-`body.ok` semantics [are] preserved deliberately" describes a path that was already unreachable on `main` too.

### 🟠 Significant — `dev-mock-command.test.mjs` cannot fire on the diffs it exists to guard

Two of its three assertions read `.env.mock` and `.env.development` at **runtime** (no module-graph edge). Neither file is in `test:hooks`'s `globs` or `extraFiles` (`scripts/verify-cache.mjs:76-178`), and neither is a `computeShared` member (`:688-691`). Measured against the repo's own selectors:

```
stepTouchedByDiff(test:hooks, ['.env.development'])  -> false   computeShared -> false
stepTouchedByDiff(test:hooks, ['.env.mock'])         -> false   computeShared -> false
ci-scope.computeScopes(['.env.development'])         -> step_test_hooks = false
ci-scope.computeScopes(['.env.mock'])                -> step_test_hooks = false
```

Repro: a one-line commit flipping `.env.development` back to `VITE_USE_MOCKS=true` (the exact regression #2343 is about). Pre-push `verify:fast:branch` skips `test:hooks` as out of scope; cloud `verify.yml` derives the same answer from the same `STEPS[]` and schedules it `false`. The new guard sits green-by-absence. `package.json` is covered (via `computeShared`), so only the two env files are exposed — but they carry the invariant. This is the #1847 trap that `verify-cache.mjs`'s own comments document five separate times at lines 78-82, 96-108, 109-116, 213-216 and 220-224; `.env.mock` + `.env.development` belong in that step's `extraFiles`.

### 🟠 Significant — `LOCAL_TTS_PORT` in the generated `server/.env` is read by nothing, and the file's header claims an isolation it does not deliver

`scripts/wt-new.mjs:159` writes `LOCAL_TTS_PORT=${ports.LOCAL_TTS_PORT}`. `git grep LOCAL_TTS_PORT server/src` returns exactly one hit — `server/src/routes/worktrees.ts:15`, a display-only list of port var names. The server never reads it:
- the sidecar port is a hardcoded constant — `server/src/tts/spawn-sidecar.ts:102` (`const DEFAULT_PORT = 9000`) and `server/src/tts/sidecar-owner.ts:30` (`const SIDECAR_PORT = 9000`);
- the URL the server *connects* on resolves `user-settings.sidecarUrl` → `LOCAL_TTS_URL` → `localhost:9000` (`server/src/workspace/user-settings.ts:466`), and `LOCAL_TTS_URL` is not written into the generated file either.

So the generated header (`wt-new.mjs:137-140`) — "this worktree's server binds its OWN port and reads/writes its OWN workspace, so it never collides with … another worktree's … server" — is true for HTTP and false for the sidecar. Repro: `wt-new` two slots, `npm run dev` in each. Worktree A binds :8310 and supervises a sidecar on :9000; worktree B binds :8320 and supervises a sidecar on **:9000** as well. The cross-stack guard that exists for exactly this (`sidecar-owner.ts:1-24`) keys on `.run/tts.owner.json` under `resolveRunDir(repoRoot)` (`server/src/app-dirs.ts`, `index.ts:111`), which is per-checkout — so B never sees A's note and the refusal never fires. Either drop the inert key, or wire the slot's port through (`LOCAL_TTS_URL=http://localhost:<port>` + a real port override) and say what is actually isolated.

---

### 🟡 Minor

1. **The identical overclaim survives in two other places.** CLAUDE.md's "Components only ever import from `api.*`" was corrected as an overclaim (`CLAUDE.md:476-486`), but `.github/copilot-instructions.md:71` still says "components import from src/lib/api.ts and don't care which implementation is active", and `docs/features/199-advanced-settings.md:31` still says "Components only import from `api.*` — never directly call fetch" — in a file **this PR edits** (line 2). `copilot-instructions.md`'s own history (`978abdf9 docs(docs,frontend): correct four false claims found by review`) shows it is maintained for accuracy.
2. **`npm run dev:mock` silently drops the ffmpeg preflight.** `predev` (`package.json:20`) runs `scripts/preflight-ffmpeg.cjs` for `dev`; npm's pre-hook keys on the exact script name, so `dev:mock` has none — yet it starts the same `dev:server` leg, whose MP3 encoder shells out to ffmpeg. `npm run dev` fails loudly on a missing ffmpeg; `npm run dev:mock` fails at first generation instead.
3. **Two undeclared guard bypasses** (`src/components/api-fetch.guard.test.ts:88`). The header enumerates its blind spots carefully, but not these two, both of which are plain source a human would write: `fetch ('/api/x')` (whitespace before the paren — the regex hard-codes `fetch\(`) and `fetch('api/config')` (no leading slash — resolves to `/api/config` from the SPA root). Probed against the exact regex: both `BYPASS`. `window.fetch(` / `globalThis.fetch(` / `self.fetch(` correctly `FLAGGED`.
4. **The guard's "outside src/components" note is incomplete.** Lines 45-50 name only `src/store/queue-thunks.ts`. `src/views/cast.tsx:422` is a live `fetch('/api/qwen/detect')` — a byte-identical duplicate of the allowlisted `qwen-status-notice.tsx:24` — sitting outside the scanned tree and unmentioned. One instance fixed, the class next door left armed.
5. **CLAUDE.md's worktree checklist wasn't updated, so #2345 survives for every worktree not made by `wt-new`.** "Worktree setup (do this before the first commit)" lists `npx husky` + both `node_modules` junctions for an `EnterWorktree` / `isolation:"worktree"` tree. Those trees still get no `server/.env`, so their server still binds :8080 and still runs on defaults — the exact defect this PR fixes for the `wt-new` path only.
6. **The whisper-install "best-effort swallow" test does not itself detect the removal.** Deleting `.catch(() => {})` from `whisper-install.tsx:78` leaves all 8 assertions passing (`Tests 8 passed (8)`); the run is red only because Vitest's global unhandled-rejection reporter sets `Errors 1 error` / exit 1. Red for the right cause, but by a mechanism the test does not own — a config with `dangerouslyIgnoreUnhandledErrors` would turn it green.
7. **PR body over-counts.** "`scripts/tests/wt-new.test.mjs` additions (**12 cases**)" — the diff adds **5** `test(...)` blocks (two of which loop over slots).
8. **Stale line left in a file this PR edits.** `docs/features/archive/23-mock-toggle.md:31` still records `real.getChapterAudio` throwing `"Chapter audio not wired yet. Set VITE_USE_MOCKS=true."` at `api.ts:612-614`. That string exists nowhere in the tree; the real impl is fully wired at `src/lib/api.ts:9810-9817`.
9. **Issue labels not reconciled before `Closes`.** #2344, #2345 and #2347 all still carry `needs-plan`, and #2344's title ("~25 components bypass the api.* mock seam…") is the framing this PR's own body retracts — it closes unamended, leaving the closed record asserting something the PR calls false. Checklist step 6 asks for this explicitly.

---

### ✅ What is solid

- **`.env.mock` precedence — verified, not assumed.** `loadEnv(mode, cwd, '')` for all three modes: `mock` → `VITE_USE_MOCKS=true`, `VITE_PORT=5403`, `VITE_API_PORT=8310`. The mode file wins the flag, the git-ignored per-worktree `.env.local` keeps the ports, and `.env.development`'s `false` is simply not loaded under `--mode mock`. Starting `dev:server` alongside is the right call — 19 raw `/api/…` fetches under `src/components/` have no mock counterpart and would 502 otherwise.
- **The guard genuinely reddens, in both directions.** Mutation-verified: a new unlisted `src/components/__zz_probe.tsx` with one `fetch('/api/sneaky')` → *"NOT on the allowlist"*, suite red; laundering `qwen-status-notice.tsx`'s literal into a variable → *"allowlisted for 1 direct fetch(es), but the scan now finds ZERO"*, suite red. It cannot fail open on absent input either (an empty scan produces 9 problems from the reverse loop). Every allowlist count matches reality exactly — I re-derived all 19 sites across the 9 files independently.
- **`dev-mock-command.test.mjs` is a real test.** Stripping `--mode mock` from `dev:frontend:mock` turns it red with the right message.
- **The `whisper-install` reroute preserves observable semantics, including in mock mode.** `qa.asr.model` is absent from `MOCK_CONFIG_DESCRIPTORS`, so `mockGetConfig()` resolves without it and the label stays at the registry default `base` — the same outcome the old `res.ok ? … : null` swallow produced. The obvious ironic failure (works real, breaks mock) is not present here.
- **#2345's core fix is real and correctly reasoned.** `server/src/index.ts:88` reads `process.env.PORT`, `server/src/load-env.ts:35` loads `./.env` with cwd at `server/` under `npm --prefix server run dev`, and `WORKSPACE_DIR=../castwright-workspace` does resolve to `<worktree>/castwright-workspace` via `SERVER_ROOT` (`server/src/workspace/paths.ts:16-38`) exactly as the comment claims. `server/.env` is git-ignored (`.gitignore:20`), so a generated worktree stays clean.
- **The doc sweep holds up.** No "mock is the dev default" claim survives outside genuinely-accurate e2e/marketing contexts. The `fs38` §2.1 rewrite correctly inverts a safety warning whose premise had flipped. The two `2026-06-18-app-17` edits are corrections, not history-rewriting — `.env.development` had already been `false` for a month when those docs were written.

---

### Verdict

**Request changes.** 1 blocking + 3 significant correctness findings. The blocker is small to fix (unwrap `{error}` in `realRestartSidecar`, or keep the raw fetch) but it currently ships a worse error message than `main` under two release-notes lines that promise a better one. The two guard-wiring findings matter more than their size suggests: both are cases where the new safety net is real code that cannot run at the moment it is needed.

dudarenok-maker:
## PR review — pass 2 (head 1ef1d419, effort high)

Scope: the four fix commits (`0e165558`, `7bb496b2`, `1ef1d419`) against pass 1's 13 findings, plus everything they touched — `realRestartSidecar` and all three of its call sites (real **and** mock), the `!body.ok` removal, the widened guard matcher + `src/views/` scan, `verify-cache.mjs`'s `extraFiles`, `wt-new.mjs`'s `renderServerEnv`, `predev:mock`, CLAUDE.md's new worktree step, and both release-notes entries · verified: `npm run typecheck` clean, `npm run lint` clean, full frontend Vitest **4771/4771** across 335 files, `npm run test:hooks` **1387/1387** — before adversarially probing the restart wrapper's fallback branches against nine malformed bodies, the guard regex against four mutants and a live `src/modals/` probe file, the scope selectors against nine diff shapes, and `LOCAL_TTS_PORT` against what actually reads it.

Working tree untouched: every mutation restored and `git status --porcelain` re-checked empty after each.

---

### Pass 1 findings — disposition

**🔴 BLOCKING (raw JSON wrapper; both release-notes lines false) — VERIFIED RESOLVED.**
Mutation-verified, not read: reverting `realRestartSidecar` (`src/lib/api.ts:8793-8817`) to the pre-fix wrapper turns **3 tests red for the right reason** — `api.restart-sidecar.test.ts` 409 (`expected 'Sidecar restart failed (409): {"ok":f…' to be 'No active supervisor…'`), the same at 503, and `blocker-fix-action.test.tsx`'s end-to-end case, which stubs `fetch` rather than `api.restartSidecar` and so exercises the real wrapper. Fallbacks probed against the real module with nine bodies:

```
empty body      [503] => "Sidecar restart failed (503): "        (statusText, empty under undici)
'null'          [409] => "Sidecar restart failed (409): null"
'{}'            [409] => "Sidecar restart failed (409): {}"
error: 123      [409] => wrapper + raw body
error: ""       [409] => wrapper + raw body
error: {m:"x"}  [409] => wrapper + raw body
'[1,2]'         [409] => wrapper + raw body
'<html>502…'    [502] => wrapper + raw body
```

No parse error escapes; `JSON.parse('null')`'s `TypeError` is caught by the same `catch`. Both release-notes lines rewritten and no longer claim the inverse (two residual doc nits below).

**🟠 SIG-1 (dead `!body.ok`; two tests pinning an impossible state) — VERIFIED RESOLVED, removal independently confirmed safe.** I re-derived it rather than trusting the commit: `sidecarHealthRouter.post('/restart')` (`server/src/routes/sidecar-health.ts:672-729`) is the only handler for that path (no second router, and the endpoint is absent from `openapi.yaml`); all five failure branches are 409/503 with a populated `{ok:false,error}`, success is `res.json({ok:true})` at 200. `mockRestartSidecar` (`src/lib/api.ts:8677-8680`) returns `{ok:true}` unconditionally. Both surviving callers — `blocker-fix-action.tsx:164` and `advanced.tsx:291` via `config-slice.ts:70`'s thunk — treat a fulfilled promise as success and neither reads `.ok`, so they now agree. Mock mode is unchanged in both. No 200-with-`ok:false` producer exists in either mode.

**🟠 SIG-2 (`test:hooks` scope gate) — VERIFIED RESOLVED, and not over-broadened.** Against the repo's own selectors:

```
['.env.development']    stepTouchedByDiff=true   ci step_test_hooks=true
['.env.mock']           stepTouchedByDiff=true   ci step_test_hooks=true
['README.md']           false / false      ['src/App.tsx']        false / false
['server/src/index.ts'] false / false      ['docs/BACKLOG.md']    false / false
['.env.e2e'] ['.env.marketing'] ['.env.local']  all false / false
```

**🟠 SIG-3 (`LOCAL_TTS_PORT`) — VERIFIED RESOLVED, and the fix is better-founded than my predecessor's diagnosis.** Pass 1 called the key "read by nothing". It is in fact read: `server/tts-sidecar/start.ps1:98` (`$port = if ($env:LOCAL_TTS_PORT) …`) and `start.sh:43`, and **both scripts whitelist `LOCAL_TTS_*` straight out of `server/.env`** (`start.ps1:38-56`, `start.sh:22-37`) — plus `buildSidecarEnv` (`spawn-sidecar.ts:466`) spreads `process.env`, which `load-env.ts` has already populated from `server/.env`, and `start-app-prod.mjs:134-144` loads that file too. Meanwhile `spawnSidecar` probes and the server polls the hardcoded `DEFAULT_PORT = 9000`. So the shipped key would have made every `wt-new` worktree's sidecar bind the slot port while its server polled :9000 — TTS broken, not merely inert. Removal is correct and the generated header's "Known limitation" paragraph is accurate.

**🟡 Minors 1-9 — all resolved.** Spot-verified rather than assumed:
1. `copilot-instructions.md:71` and `199-advanced-settings.md:31` now defer to CLAUDE.md's bounded set; 199's added clause "none of those apply to the Advanced Settings surface" is true (a full `src/**` scan finds no literal `/api/` fetch in `advanced.tsx`).
2. `predev:mock` — and npm **does** fire pre-hooks for colon-named scripts on this box: a scratch `package.json` on npm 11.12.1 printed `RAN predev:mock` then `RAN dev:mock`. (The new test only asserts package.json shape, so this mechanism was worth confirming.)
3. Matcher widened — both new cases flagged, and each widening is pinned by a test that actually reddens when reverted.
4. `src/views/` scanned; `views/cast.tsx:422` allowlisted. The justification checks out: there is no `api.*` wrapper for `/api/qwen/detect`, so allowlisting (not rerouting) is the right call, consistent with `qwen-status-notice.tsx`. See new finding N2 for the residual.
5. CLAUDE.md worktree step 5 added — see new finding **N1**, it is incomplete in a way that bites.
6. whisper swallow test — mutation-verified: deleting `.catch(() => {})` from `whisper-install.tsx:77` now fails **on the test's own assertion**, `whisper-install.test.tsx:191 expected "vi.fn()" to not be called at all, but actually been called 1 times`, not on Vitest's global reporter.
7. PR-body count corrected 12→5 — but now stale again at 7 (nit N5).
8. `23-mock-toggle.md:31` — verified: `api.ts:9828-9835` is exactly `getChapterAudio`, fetching `/api/books/{bookId}/chapters/{chapterId}/audio` and throwing only on non-ok.
9. `needs-plan` dropped from #2343/#2344/#2345 and #2344 retitled to "two components duplicate api.* wrappers…", retiring the framing the PR body retracts.

**#2349 deferral — legitimate.** It clears CLAUDE.md's single defer bar rather than dodging it: the decision owed is named explicitly ("how should a *background* action's failure surface in this view — toast / inline banner / deliberate silent-with-console? And is that answer applied to the whole class?"), it spans `onInstalled` **and** every `ModelControlPill` render site, and it carries acceptance criteria, key files and a benefit line. It is also genuinely pre-existing: `realRestartSidecar` threw on `!res.ok` on `main` too, so this PR changed the message's content, not whether or where it throws. Labelled/boarded correctly. I would not force this into the round.

---

### 🟠 Significant

**N1 — correctness bug — CLAUDE.md's new worktree step 5 silently cross-wires the worktree's UI to the *primary checkout's* server and workspace.** `CLAUDE.md:999-1008`.

The step tells the reader to create `server/.env` with its own `PORT` — and stops there. It never mentions the root `.env.local` that must agree with it.

Repro, exactly as written. Take a tool-made worktree (`EnterWorktree` / `isolation:"worktree"`), which is precisely the case this section addresses. `.env.local` is git-ignored, so the tree has none. Follow step 5: write `server/.env` with `PORT=8090`, `WORKSPACE_DIR=../castwright-workspace-foo`. Now `npm run dev`:

- the server binds **:8090** and looks perfectly healthy;
- Vite computes `apiPort = Number(env.VITE_API_PORT ?? env.PORT ?? 8080)` (`vite.config.ts:116`) — `loadEnv` is root-scoped and finds neither key (there is no tracked root `.env`, and `.env.development` sets only `VITE_USE_MOCKS=false`), so the proxy targets **:8080**;
- `strictPort` is unset (`vite.config.ts:143`), so Vite takes :5174 when the primary holds :5173 and comes up fine.

Net: the worktree's frontend talks to the **primary checkout's** server, reading and writing the primary checkout's `cast.json`/`state.json`, while its own isolated server idles unused — the exact isolation the step exists to provide, inverted, and silent. Before the step, the worktree server at least died on `EADDRINUSE`, which is visible.

`wt-new.mjs` gets this right by construction — the PR body's own words: "`renderServerEnv` takes the same `ports` object as `renderEnvLocal`, so the two files cannot drift by construction." The hand-worktree instruction reintroduces exactly that drift. The step needs the matching `.env.local` (`VITE_PORT` / `VITE_API_PORT` / `PORT`), or a line telling the reader to run `wt-new` instead.

---

### 🟡 Minor

**N2 — cleanup nit — the guard's class is still armed one directory over, and the blind-spot list does not say so.** `src/components/api-fetch.guard.test.ts:56-61, 185-188`.

Pass 1's finding 4 was "one instance fixed, the class next door left armed"; the fix widened the scan by exactly one directory. `src/modals/` (**55 files**), `src/routes/` and `src/hooks/` are still unscanned, and the header's blind-spot bullet names only `src/store/queue-thunks.ts` — a reader would not learn that `src/modals/` is outside the net. Demonstrated, not reasoned: I dropped `src/modals/__zz_probe_modal.tsx` containing a bare `await fetch('/api/sneaky-modal-endpoint', {method:'POST'})` and ran the guard — `Test Files 1 passed (1) / Tests 5 passed (5)`. (Probe deleted; tree clean.) No live violation exists there today — a full `src/**` scan finds direct `/api/` literals only in the nine allowlisted components, `views/cast.tsx`, and `src/lib/api.ts` itself — so I am classing this cleanup rather than correctness, but the same argument that justified adding `src/views/` applies verbatim to `src/modals/`.

**N3 — cleanup nit — the `refetch()`/`onRefetch()` test cannot fail.** `src/components/api-fetch.guard.test.ts:234-247`.

The fixture contains `refetch()`, `onRefetch()`, `refetch ()`, `onRefetch ()` — none followed by a quote and `api/`, so `DIRECT_API_FETCH_RE` cannot match any of them under **any** variant of the regex. Measured against four mutants:

```
SHIPPED regex                      refetch-fixture PASSES
lookbehind (?<![\w$]) DELETED      refetch-fixture PASSES
\s* before paren DELETED           refetch-fixture PASSES
optional leading slash DELETED     refetch-fixture PASSES
```

The two lines `1ef1d419` added to this fixture to cover the widened `\s*` therefore add nothing. The header's claim "Proven below: a fixture containing only `refetch()`/`onRefetch()` (spaced and unspaced) is asserted NOT flagged" is not proven by that fixture. The property *is* pinned — but only by accident, by the corpus test: deleting the lookbehind reddens it with `library/library-grid.tsx: 1 NOT allowlisted`, because a **prose comment** at `src/components/library/library-grid.tsx:189-190` happens to read `a same-URL refetch\n     (\`/api/books/:bookId/cover\`)`. Reword that comment and the lookbehind becomes untested. A fixture of `refetch('/api/x')` / `onRefetch ('/api/x')` would test what the name claims.

**N4 — cleanup nit — the rewritten register entry describes the guard's pre-widening scope.** `docs/release-notes-next.md:359` says the guard pins "the current direct-`fetch('/api/...')` surface under `src/components/`". `1ef1d419` widened `SCAN_ROOTS` to `src/components/` **and** `src/views/` and allowlisted `views/cast.tsx`, and did not revisit the line the previous fix commit had just written. Prose describing the mechanism a later commit in the same PR changed.

**N5 — cleanup nit — the PR body now contradicts the shipped code in two places.** Both were true when written and were falsified by the fix commits:
- "Both now route through the seam, with their existing swallow / **in-band-`body.ok` semantics preserved deliberately**" — `0e165558` deleted that check and both tests pinning it, correctly. As it stands, `Closes #2344`'s permanent record asserts a semantic the PR removes.
- "`wt-new` now generates a minimal `server/.env` with the slot's `PORT` **and `LOCAL_TTS_PORT`** and an isolated `WORKSPACE_DIR`" — `7bb496b2` removed `LOCAL_TTS_PORT` and added a test forbidding it.

Counts are stale too: `git diff origin/main...HEAD -- scripts/tests/wt-new.test.mjs | grep -c '^+test('` → **7**, body says 5; suites are **4771** / **1387**, body says 4765 / 1382.

**N6 — cleanup nit — `RELEASE_NOTES.md:123` narrates a state no user ever saw.** "The Setup checker's fix button already showed a plain explanation when a restart failed; the matching button in Settings → Advanced Configuration instead showed a wall of raw technical detail." That Advanced button's failure notice **did not exist in 1.14.0** — it was added by `#2221`, whose bullet sits four lines above at `RELEASE_NOTES.md:122`, inside the *same unreleased* `# Castwright 1.15.0` section (heading at line 1). A user upgrading from 1.14.0 saw nothing there, not a wall of detail. The two bullets describe one net change and read as a diary of the intra-release intermediate state.

**N7 — cleanup nit — the sidecar-restart failure message has no length cap, where its sibling does.** `src/lib/api.ts:8803-8809` → `src/components/blocker-fix-action.tsx:199,214` and `src/views/advanced.tsx:298-303`.

Measured through the real wrapper and the real component: a 502 whose body is a 40 KB HTML page produces `message.length = 40056`, head `Sidecar restart failed (502): <html><body>AAAA…`, rendered by `<p className="text-xs text-rose-700">{error}</p>` with no truncation and no overflow container, and pasted verbatim into `advanced.tsx`'s toast. The config path caps exactly this at `MAX_SAVE_ERROR_MESSAGE_LENGTH = 500` (`src/components/settings/override-row.tsx:144`), and its comment names this precise input: *"a 500 from a proxy/middleware that never reached the JSON route… the raw body can be an entire HTML error page."* Pre-existing on both call sites and not worsened here, which is why I class it cleanup — but this PR rewrites this exact function, explicitly contemplates "a proxy's HTML error page" in its new comment, and passes it through uncapped. If you consider proxy-fronted deployments in scope (this app ships a LAN-HTTPS path), it is a correctness bug, not a nit.

**N8 — cleanup nit — the release-notes pair covers only #2344.** `git diff origin/main...HEAD` touches `RELEASE_NOTES.md` and `docs/release-notes-next.md` by exactly **one line each**, both for the restart message. Nothing records the PR's *headline* change (#2343 — `npm run dev:mock`, `.env.mock`, the ~70-file doc sweep) or #2345. Contributor-facing tooling is a defensible skip under checklist step 5 — but the step says to *say so explicitly*, and the body's "Not applicable, stated rather than skipped silently" paragraph lists the regression plan, on-box acceptance and `INDEX.md`, not this.

---

### Verdict

**Request changes — narrowly.** All four of pass 1's correctness findings are genuinely fixed and mutation-verified, not merely asserted; SIG-3's fix is better-reasoned than the finding that prompted it; the #2349 deferral is legitimate; and nothing the fix commits touched regressed (typecheck, lint, 4771 frontend, 1387 hooks all green). One new correctness bug: **N1**, in `1ef1d419`'s CLAUDE.md addition, where following the new step exactly points a worktree's UI at the primary checkout's server and workspace, silently. The remaining seven are cleanup and would not, on their own, warrant another round.

dudarenok-maker:
## PR review — pass 3 (head 4c11e8d8, effort high)

Scope: the merge `827fb392` diffed against **both** parents (`1ef1d419`, `90e8ab6e`) plus the merge base `090168a5`, the four fix commits (`44495f5a`, `5ccac45c`, `1eba94bd`, `4c11e8d8`), the rewritten PR body, and everything those touched — the guard's five scan roots, `capRestartErrorMessage` against `override-row.tsx`'s convention, CLAUDE.md's rewritten worktree step 5 against `vite.config.ts` + `scripts/wt-new.mjs`, and `run-app/SKILL.md`'s four launch rows against `package.json` + `server/src/index.ts` + `scripts/start-app.ps1` · verified: `npm run typecheck` clean (frontend + server), `npm run lint` clean, full frontend Vitest **4773/4773** across 335 files, `npm run test:hooks` **1402/1402** — before adversarially probing the widened scan with live probe files in all three new roots, mutating the matcher's lookbehind, and tracing `dev:mock`'s server leg to what it actually spawns.

Working tree untouched: every probe file removed, every mutation restored via `git checkout --`, `git status --porcelain` re-checked empty after each.

---

### The merge (`827fb392`) — clean, verified against both parents

Not assumed. `git diff 090168a5 90e8ab6e --name-only` lists **18** files main changed since the base; of those, exactly **6** differ between `90e8ab6e` and HEAD, and all six are this PR's own deliberate edits (`CLAUDE.md`, `docs/features/199-advanced-settings.md`, `.claude/skills/run-app/SKILL.md`, `package.json`, `scripts/verify-cache.mjs`, `scripts/tests/verify-cache.test.mjs`). Nothing from `6704b4d3` or any other main commit was dropped:

- **CLAUDE.md** — `6704b4d3` changed exactly one thing there (the `.env.development` claim + the PowerShell-broken command). Main's *correction* survives verbatim in substance; only the command is replaced. `af2a4212`/`7f5e1b41`'s CLAUDE.md content is untouched by the merge (`git diff 90e8ab6e 827fb392 -- CLAUDE.md` shows only the two PR-side hunks).
- **199-advanced-settings.md** — main's own edit here came from `b536c560`, not `6704b4d3`. Its text (`opt-in — mocks are OFF by default in dev … via npm run dev`) was self-contradictory; the resolution replaces it with `npm run dev:mock`, which is strictly more accurate. Nothing lost.
- **`scripts/verify-cache.mjs` — the auto-merge that could have gone wrong, and didn't.** Main *replaced* the two `.claude/skills/*` literals with a `'.claude/skills/**'` glob and rewrote the surrounding comment; this PR *appended* `.env.mock`/`.env.development` into the same `extraFiles` array. Git auto-merged both. The merged file has main's glob at `:145-153`, main's rewritten `CLAUDE.md` comment at `:245-258`, and this PR's two entries at `:259-273` — all three intact. `scripts/tests/verify-cache.test.mjs` likewise carries **both** lanes' tests (main's `:447`, `:457`; this PR's `:505`, `:510`). `package.json` carries main's `skills:sync` (`:81`) alongside this PR's three scripts (`:23`, `:25`, `:26`).
- **`run-app/SKILL.md`** — main's other `6704b4d3` fix, the STEP 3 `[ -d "$VERIFY_DIR_WIN" ] || … exit 1` cleanup guard, survives at `:155`.
- **`npm run dev:mock` still correct post-merge** — re-measured through the real `loadEnv(mode, cwd, '')`: `mock` → `VITE_USE_MOCKS=true` with this worktree's `.env.local` ports preserved (`VITE_PORT=5403`, `VITE_API_PORT=8310`); `development` → `false`. `.env.mock` is tracked, `.env.local` git-ignored. `scripts/sync-agent-skills.mjs` (new from main) mirrors only `pr-review-gate`, so `44495f5a`'s `run-app` edit owes no `npm run skills:sync`.

---

### Pass 2 findings — disposition

**🟠 N1 (CLAUDE.md worktree step 5 cross-wires the worktree to the primary checkout) — VERIFIED RESOLVED.** `CLAUDE.md:1007-1026` now requires both halves and states the mechanism correctly. Re-derived against source rather than read: `vite.config.ts:116` is `Number(env.VITE_API_PORT ?? env.PORT ?? (useHttps ? 8443 : 8080))`, `vite.config.ts:143` sets `port: vitePort` with no `strictPort`, `scripts/wt-new.mjs:51` is `PORT_STEP = 10`, `.env.local` is ignored by `.gitignore:24` and written only by `wt-new.mjs:277`. `WORKSPACE_DIR` relative-to-`server/` is right (`wt-new.mjs:130-139`). One residual nit below (**P3-3**).

**🟡 N2 (guard's class armed one directory over) — VERIFIED RESOLVED, and the "zero new sites" claim independently confirmed.** The three added roots are real and non-empty (`src/modals` 28, `src/routes` 2, `src/hooks` 4 non-test sources), and an independent `fetch\s*\(` sweep across all three finds nothing but the word "prefetch" in two comments. That the scan *reaches* them was proven, not inferred — three probe files (`src/modals/__zz_probe.tsx` `fetch('/api/sneaky-modal')`, `src/routes/__zz_probe.ts` `fetch('/api/sneaky-route')`, `src/hooks/__zz_probe.ts` `fetch ('api/sneaky-hook')` — deliberately the spaced, slash-less variant) each produced its own `modals/… | routes/… | hooks/…: 1 direct fetch('/api/...') call(s) — NOT on the allowlist` failure. Probes deleted; tree clean. Separately, a `-P` regex sweep of all of `src/**` finds direct `/api/` fetch literals **only** in the nine allowlisted components, `views/cast.tsx:422` (allowlisted) and `src/lib/api.ts` itself — the guard's coverage now equals the live surface.

**🟡 N3 (the `refetch(` exclusion test could not fail) — VERIFIED RESOLVED, and not tautological in a new way.** Mutation-verified against the shipped file: deleting `(?<![\w$])` from `DIRECT_API_FETCH_RE` (`api-fetch.guard.test.ts:121`) now turns the test red **on its own assertion** — `AssertionError: expected { count: 4, details: [ …(4) ] } to be null` — counting exactly the four fixture call sites, alongside the corpus test's incidental `library-grid.tsx` failure that used to be the only thing pinning the property. Restored; green with the shipped regex.

**🟡 N4 (release-notes entry described the pre-widening scope) — VERIFIED RESOLVED.** `docs/release-notes-next.md:359` now names all five roots and matches `SCAN_ROOTS` exactly.

**🟡 N5 (PR body contradicted the shipped code) — VERIFIED RESOLVED.** The "in-band-`body.ok` semantics preserved deliberately" and "`and LOCAL_TTS_PORT`" claims are gone; the stale case counts are gone; the suite figures now read 1402 / 4773 / 7545, and I measured 1402 and 4773 exactly.

**🟡 N6 (RELEASE_NOTES narrated an intra-release intermediate state) — resolved as filed, but the fold introduced a *new* false claim.** See **P3-2**.

**🟡 N7 (no length cap on the restart failure message) — VERIFIED RESOLVED, and it matches the sibling convention exactly.** `capRestartErrorMessage` (`src/lib/api.ts:8803-8808`) is `length > 500 ? slice(0,500) + '…' : message` — same threshold, same strict `>`, same U+2026 suffix, same 501-char result as `finalizeSaveErrorMessage` (`override-row.tsx:144-161`). The two divergences are both correct here: no `.trim()` (nothing depends on it) and no empty-string fallback (unreachable — `parsedError` is only assigned when `typeof body.error === 'string' && body.error`, and the wrapper branch always carries the `Sidecar restart failed (nnn): ` prefix). Applied to **both** throw paths in one place. It does not break `0e165558`'s exact-sentence assertions: the real server sentences are ~100 chars, and the full frontend suite is 4773/4773 green.

**🟡 N8 (release-notes pair covers only #2344) — VERIFIED RESOLVED.** The body's not-applicable paragraph now states it explicitly, with the reason.

**Nothing the five commits touched regressed.** typecheck, lint, 4773 frontend, 1402 hooks all green; all 16 PR checks pass; `mergeStateStatus: CLEAN`.

---

### 🟠 Significant

**P3-1 — correctness bug (in the document) — `run-app/SKILL.md` now tells the reader `npm run dev:mock` starts "no sidecar". It starts one, on a default box — and the commit message's stated rationale for the `npm start` row is false.**
`.claude/skills/run-app/SKILL.md:18-19`, `:54`, `:55` (and the inconsistency at `:125`).

Both new sentences claim a sidecar-free launch:
- `:18-19` — "This starts the frontend with `VITE_USE_MOCKS=true` (`.env.mock`) alongside the real server, **no sidecar**."
- `:54` — "| Frontend (mock mode) + real server, **no sidecar** — the fastest full mock verify … | `npm run dev:mock` |"

Traced, not reasoned. `package.json:26` — `dev:mock` = `concurrently "npm:dev:frontend:mock" "npm:dev:server"`. `package.json:24` — `dev:server` = `npm --prefix server run dev`, **byte-identical to the leg `npm run dev` uses** (`package.json:21`). That server, in its own `listen` callback, unconditionally builds and starts the sidecar supervisor — `server/src/index.ts:314` `createSidecarSupervisor({ buildOpts: … autoStart: getResolvedAutoStartSidecar() })`, `:325` `void sidecarSupervisor.start()` — and `autoStartSidecar` defaults to **`true`** (`server/src/workspace/user-settings.ts:309`, `:499-505`). Nothing in `.env.mock` (one line, `VITE_USE_MOCKS=true`) or in `--mode mock` reaches the server: `loadEnv` is root-scoped and the server reads only `server/.env`.

Repro: on a stock checkout with default user settings, run `npm run dev:mock`. A `python`/uvicorn child appears on `:9000` and the configured engine's weights load — the same process `npm run dev` and `npm start` produce. The reader was told it wouldn't.

The `:55` row compounds it. `4c11e8d8` changed that row from `npm run dev` to `npm start` with the rationale "*`npm run dev` is frontend+server only (no sidecar); `npm start` is the full stack, per CLAUDE.md's own Commands section*". That premise is false in both directions: `scripts/start-app.ps1:6` says outright "**Plan 43: the TTS sidecar (uvicorn :9000) is no longer launched from here**", and `scripts/start-app.mjs:18-19` makes `npm start` literally `npm run dev` on POSIX. The sidecar comes from the *server*, so all three of `npm run dev`, `npm run dev:mock` and `npm start` produce it identically — the differentiator between the rows is mocks, not the sidecar. `:125` still says "the full `npm run dev` stack", which the `:55` edit just declared not-full, so the file now disagrees with itself too.

This matters beyond tidiness: this is the launch table an agent reads to decide what to start, on a repo whose own guidance is to avoid loading a sidecar while a GPU generation is live. The immediate cause is a genuinely stale line one file over — **`CLAUDE.md:321`** describes `npm run dev` as "Vite dev server (HMR) on `http://localhost:5173`", which stopped being true when `dev` became `concurrently frontend+server`; that is the sentence `4c11e8d8` cited. Since this PR already edits `CLAUDE.md`, fixing `:321` and the four `run-app` sites together is one change.

### 🟡 Minor

**P3-2 — cleanup nit — N6's fold replaced an intra-release diary with a claim that is false for one of the two buttons it names, and now contradicts this PR's own technical register.** `RELEASE_NOTES.md:122`.

The folded bullet reads "*in the same plain sentence **wherever you meet the button**. **Until now, a failed restart request looked exactly like a successful one — the button just went back to normal with nothing said.**… the Setup checker's fix button and the matching button in Settings → Advanced Configuration both show the same plain sentence, **not a wall of raw technical detail**.*"

For the Setup checker's button that is not what a 1.14.0 user saw. `git show v1.14.0:src/components/blocker-fix-action.tsx` has, at the restart handler, `if (!body.ok) throw new Error(body.error ?? 'Restart failed.')` (`:160`) inside a `try` whose `catch` does `setError(e.message)` (`:163-165`), rendered by `{error && <p className="text-xs text-rose-700">{error}</p>}` (`:195`, `:210`). Since the route answers 409/503 with a populated `{ok:false,error}` body and 1.14.0's code called `res.json()` without an `!res.ok` check, that path parsed and displayed the server's own sentence. So on that button 1.14.0 said something, and said it plainly — neither "nothing said" nor "a wall of raw technical detail" applies.

This PR's own `docs/release-notes-next.md:359` records the opposite, correctly: "*That was a regression for **this button, which previously parsed and showed that sentence directly**.*" The paired notes documents now disagree with each other inside one PR. Pass 2's N6 was right that two bullets described one net change; the fix generalised the wrong bullet's premise. The accurate fold scopes the "nothing said" clause to the Settings → Advanced Configuration button.

**P3-3 — cleanup nit — CLAUDE.md's rewritten step 5 lists `VITE_PORT` as required but never says what to set it to, and its one natural reading collides two servers.** `CLAUDE.md:1007-1008` vs `:1017-1020`.

`:1007-1008` requires "a root `.env.local` with **matching** `VITE_PORT` / `VITE_API_PORT` / `PORT`". The only sentence that assigns values, `:1017-1020`, says "Pick an unused port … set it **as `PORT`/`VITE_API_PORT` in `.env.local` and as `PORT` in `server/.env`**" — `VITE_PORT` is never given one. A reader taking "matching" at face value sets all three to the same value (say 8090), and Vite then tries to bind the port the server wants; with no `strictPort` (`vite.config.ts:143`) Vite silently slides to 8091 while the server keeps 8090, so the reader's own printed URL is wrong. A reader who omits `VITE_PORT` gets Vite on the primary checkout's 5173→5174. Neither breaks the isolation the step exists to deliver — the proxy target is correctly covered — which is why this is a nit; `wt-new.mjs` sets `VITE_PORT = 5173 + slot*10` (`:45`, `:110`) and the step should just say so.

---

### Verdict

**Mergeable once P3-1 is corrected — nothing else blocks.**

The merge is genuinely clean against both parents, including the one auto-merge (`scripts/verify-cache.mjs`) where two lanes edited the same array for different reasons and either side could have been silently lost. All eight pass-2 findings are resolved, and the two that mattered most — the widened scan and the `refetch(` fixture — were proven by mutation rather than read, as were N1's port claims and N7's cap boundary. No correctness regression in the product code; every suite green.

P3-1 is the one thing I would not merge as-is: it is a false statement about what a command launches, newly written by this PR into the very file the PR is correcting for accuracy, resting on a stale `CLAUDE.md:321` line that this PR is already editing. It is a five-line documentation fix (four sites in `run-app/SKILL.md` plus `CLAUDE.md:321`), not a design question. P3-2 and P3-3 are cleanup and would not, on their own, warrant another round.

Issue #2325, opened 2026-08-13
srv: no guard on attribution quality — an all-narrator book passes the coverage check silently


dudarenok-maker:
## What

Nothing in the pipeline checks attribution **quality**. `validateStage2Coverage`
(`server/src/analyzer/stage2-coverage.ts:164`) gates only on four conditions —
no sentences, low word coverage, excess coverage, duplicated block. A chapter in
which *every* line of dialogue was assigned to `narrator` re-emits the source
prose verbatim, so its coverage ratio is ~1.000 and the guard returns
`ok: true`.

That is not hypothetical. The 2026-08-12/13 *Ночной дозор* run (#2306) persisted
a whole book at **95.7% narrator across all 15,100 sentences**, using 24 of its
48 roster characters. The good 2026-08-06 run of the same book was 67.9% and
used 40 of 58. Nothing warned, nothing retried, nothing flagged — the book was
simply written to disk as a single-voice audiobook.

For a dash-convention language this is deterministically detectable, because the
convention itself tells us which paragraphs are speech.

## Proposed gate

For a language whose conventions define `dialogueOpen` (today `ru`, `es`, `fr` —
`en`/`de`/`ja`/`zh` have `dialogueOpen: null` and the gate is **inert**), measure
the share of **speech halves** attributed to `narrator`, and breach above a
threshold. Speech halves = sentences opening with the language's dialogue dash
whose first cased letter is UPPERCASE; the lowercase ones are tag halves
(`- сказал Егор.`) which are *correctly* narrator and must stay out of the
denominator. That split was validated across three independent runs of the same
chapter (246/213/33, 246/213/33, 241/209/32 dash-opening/speech/tag), so it is
stable.

Route the breach into the existing `runStage2WithCoverageGuard` retry so a
collapsed section is re-attempted rather than persisted; #2304's
deterministic-failure re-split already handles the case where retrying the same
prompt cannot help.

## Calibration (two full-book runs, per-chapter narrated speech halves)

| run | ch1..ch9 | extreme |
|---|---|---|
| 2026-08-06 (good) | 32.4 18.0 3.8 39.3 3.2 2.0 27.6 33.8 20.8 | **max 39.3%** |
| 2026-08-12 (collapsed) | 93.1 93.7 94.8 97.5 86.5 72.2 84.3 74.6 91.8 | **min 72.2%** |

Perfect separation over all 18 chapter-runs with a 33-point gap. Note the good
[...]

Issue #2324, opened 2026-08-13
srv: stage-2 handoff key has no chunk index, so a chunked chapter keeps only its last chunk's forensics


dudarenok-maker:
## What

Stage-2 attribution writes its inbox/outbox forensics under the key
`2-ch${chapterId}` (`server/src/analyzer/ollama.ts:336`), which carries **no
chunk index**. `runStage2ChapterChunked` calls `runStage2Chapter` once per
chunk with the same chapter id, and `writeInbox`
(`server/src/handoff/protocol.ts:82-85`) `rm`s the outbox, the errors file and
both raw-attempt files before each write.

So on any chunked chapter, chunk N+1 **overwrites chunk N's prompt and deletes
chunk N's response**. A chapter split into 4 sections leaves forensics for
section 4 only. The coverage-guard retries and the `attributeSpan` adaptive
re-split recursion make it worse — every one of those is another
`runStage2Chapter` call against the same key.

The same file already documents this exact defect being fixed one layer over:
escalation windows are keyed `escalation-ch{n}-w{n}` so that "a chapter with N
queried windows leaves N sets of forensics on disk instead of only the last
one (#1483)". Stage-2 never got that treatment.

## Why it matters

The module header for `protocol.ts` states its purpose: the analyzer "writes
the prompt that went to the model into the inbox and writes the raw JSON
response into the outbox so a developer can reproduce a particular run from
disk." For a chunked chapter that promise is not kept, and nothing says so.

This was found while chasing #2306, where the surviving artifact for every
chapter is its LAST chunk — and last chunks are exactly the ones that do not
fail. It cost a wrong run plan: "re-run one chapter with the outbox preserved"
would have spent hours of box time and returned the same last-chunk-only
artifact, because preservation was never the blocker. That investigation
routed around it by recording from inside the analyzer
(`server/handoff/cache/chunk-forensics.mts`), so nothing is blocked on this —
but the next person to reach for the outbox will hit the same wall.

## The decision this needs (why it is not just typed)

Filed rather than fixed on the spot because the fix has more than one
defensible shape and one of them changes an interface. Naming the decision, per
CLAUDE.md's design-pass carve-out:

1. **Where does the discriminator come from?** The chunk index is known only to
   the chunker; `Analyzer.runStage2Chapter(manuscriptId, chapterId, promptMd,
   call)` has nowhere to put it. Threading it means changing the `Analyzer`
   interface — every implementation, every mock. The alternative is a
[...]

Issue #2304, opened 2026-08-12
analyzer: the stage-2 coverage retry burns its whole budget re-running a deterministic failure, then reports it as transient


dudarenok-maker:
## What

`runStage2WithCoverageGuard` (`server/src/analyzer/stage2-coverage.ts`) re-runs a failing stage-2 attribution call up to `analyzer.stage2.coverageRetries` times (default 2). Its own doc states the premise:

> the loop-and-truncate defect is stochastic, so a fresh attempt usually clears it

When that premise does not hold, the loop re-issues an **identical call with an identical prompt**, gets an identical failure, and repeats until the budget is gone — then falls through to the caller's soft warning:

```
Chapter N — ⚠ attribution may be incomplete (…); kept the best take and flagged the chapter for retry.
```

which reads as a transient hiccup rather than "this section cannot be attributed".

## Observed

Ночной дозор chapter 8, during the C2/C3 on-box acceptance run:

```
2026-08-12 21:32:33  attribution coverage check failed (Duplicated block — 4 consecutive
                     sentences repeat earlier ones at offset 19 (repeat-loop).); attempt 2
2026-08-12 21:49:49  … same rule, same offset …                                  attempt 3
2026-08-13 08:09:36  … same rule, same offset …                                  attempt 2
2026-08-13 08:22:55  … same rule, same offset …                                  attempt 3
```

**Five occurrences, identical rule and identical offset, across two server lifetimes** — the run was paused overnight and the process restarted between them, which rules out in-process state. The degeneration is deterministic for that chapter section at the shipped temperature.

## Why it matters

- **It hides a real defect.** A deterministic failure and a transient one produce the same operator-facing message, so "the model cannot attribute this section" is indistinguishable from "unlucky sample".
- **It wastes the budget.** Every attempt after the first identical repeat is provably futile.
- **It scales badly.** On a large chapter each attempt can cost tens of minutes; the observed early-offset failures were cheap, but a repeat-loop detected late is not.

## Fix

Compare the retry's failure **signature** against the attempt it was retrying, and stop when they match exactly. The signature is structural — `endingPresent`, `duplicatedBlock {startIndex,length,offset}`, and `coverageRatio` quantised to 3dp — deliberately **not** the rendered `issues` strings, which embed a formatted ratio that could differ by a digit for a materially identical failure.

One identical repeat is the signal: the retry's entire premise is sampling variance, and an identical structural verdict is direct evidence there is none at this temperature.

[...]

PR #2350, opened 2026-08-13
fix(server,frontend,openapi): close the #2325 collapse guard's eight review findings

### LINKED ISSUE #2342: The #2325 dialogue-collapse guard: six defects and a design question from PR #2333's review gate
PR #2333 shipped #2324 (per-call handoff forensics) and #2325 (the dialogue-collapse
coverage guard). It was merged before its mandatory review gate reported. The gate
then returned eight findings against the merged code. Two of them mean the PR's green
suite was not evidence for the thing it shipped.

All are in `server/src/analyzer/stage2-coverage.ts`, `server/src/analyzer/stage2-chunk.ts`
and `server/src/routes/analysis.ts` unless stated.

---

### 1. The retry's scoring functions were never extended for the new failure class

`validateStage2Coverage` gained a `dialogueCollapse` term that gates `ok`
(`stage2-coverage.ts:311-323`), but neither function that reads a verdict was
touched:

- **`isBetterCoverage` (`:333`)** scores `(duplicatedBlock ? 100 : 0) + |1 - coverageRatio|`.
  Dialogue collapse contributes nothing. Attempt 1 collapses at 95% narrated with
  ratio 1.00 (score 0.00); the retry collapses at 61% with ratio 0.97 (score 0.03).
  The retry is materially better on the only axis that failed, and is **discarded**.
  The 95% take is what persists.
- **`verdictSignature` (`:347`)** is `end/noend | dup | cov:x.xxx`. Two attempts
  collapsing at 95% and at 65% with ratios inside the same third decimal produce an
  **identical signature**, so the loop sets `deterministicFailure`, fires
  `onExhausted`, and stops — reporting "more attempts cannot help" about a model that
  just moved 30 points on the failing axis and might have cleared the 60% bar on the
  next call.

The guard can therefore detect a collapse, and then both prefer the worst take and
abandon the retry that was working.

### 2. The user-facing copy is wrong for this failure class

A collapse routes to `FAILURE_REMEDIATIONS['attribution-incomplete']`
(`routes/analysis.ts:5010`), whose copy reads *"Some lines in this chapter may be
unattributed — the analyzer's answer did not cover every sentence"* and *"a retry
usually fills the gaps"* (`routes/failure-remediations.ts:109`).

Every sentence **was** covered. There are no gaps. They were all handed to the
narrator. The user is told to retry to fill gaps that do not exist, and is told
nothing about what actually went wrong.

### 3. Neither engine call site is covered by any test

`ollama.ts:336` and `gemini.ts:316` are the only places `call.stage2CallSeq` is ever
read — the entire point of #2324. Reverting both to `stage2HandoffKey(chapterId)`
leaves the full suite green. The forensics feature has no test at the seam where it
takes effect, which is the seam that was broken before.

### 4. The guard fails OPEN on the output shape most likely to be broken

`classifyDialogueLine` tests `dialogueOpen` against the **attributed sentence text**.
A model that drops or normalises the leading dash makes every line `indeterminate`,
so `speechHalves` is 0, `evaluable` is false, and the guard passes silently — on
output that is *more* damaged, not less.

`validateStage2Coverage` already receives `bodyText`. The source's own count of
dialogue-opening lines is the available control and is not used.

There is also no observability: nothing logs `speechHalves`/`pct` when the population
is below the floor, so an operator cannot distinguish "this span has little dialogue"
from "the guard was blind here".

### 5. The collapse remediation never reaches a re-split's children (CORRECTED)

> **Correction.** This item was first filed as *"a re-split blinds the guard and the
> chapter is reported clean"*. That is wrong, and reading the code rather than
> trusting the recalled finding is what caught it: `runChunks` recomputes the
> verdict over the FULL body and ALL sentences (`stage2-chunk.ts:461`), and the
> single-call path guards `opts.body` directly, so a chapter-wide collapse is
> always caught and always flagged. What follows is the real, narrower defect.

`STAGE2_MIN_SPEECH_HALVES` is 20, and the retry/re-split remediation only ever runs
at SECTION level, inside `runStage2WithCoverageGuard`. A section that individually
falls below the floor passes vacuously, so:

- a chapter whose sections each hold fewer than 20 speech halves but which collapses
  in aggregate is **flagged without a single retry ever being attempted** — the
  chapter-level verdict is computed once, at the end, with no remediation attached
  to it;
- an adaptive re-split makes this strictly more likely, since halving a span halves
  its speech-half population: a ~30-half section that collapses splits into two
  ~15-half children, each now under the floor, so neither child retries.

The user-visible outcome is the same message either way (see item 2), but the chapter
has burned none of its retry budget on the failure it actually had.

### 6. The single-call path still destroys its own retry forensics

#2324 numbered the CHUNKED path. The single-call path passes no sequence
(`stage2-chunk.ts`), so key `2-ch{n}` is stable across attempts — and
`runStage2WithCoverageGuard` calls `call()` again on every retry. `writeInbox`'s
first act is to `rm` the previous outbox. A single-call chapter that retried twice
keeps only the last attempt's prompt and response: the same defect #2324 fixed,
surviving on the other branch of the same `if`.

### 7. First-person novels inflate the metric — needs a design decision

In a first-person novel the protagonist's own spoken lines can legitimately carry
`characterId: 'narrator'`, and are counted as breaches. The measured headroom absorbs
it (the good Ночной дозор run, itself first-person, maxes at 39.3% against a 60%
bar), so this is not currently a false alarm — but the margin is smaller than the
calibration implies.

**The decision owed:** should a first-person protagonist's speech be attributed to
`narrator` or to a distinct character id? That is an attribution-wide behaviour
question, not a guard-tuning one, and the guard cannot separate the two cases without
an answer. Not fixable in this round; not silently dropped either.

### 8. No on-box acceptance row

The 60% threshold was calibrated on **one** Cyrillic book, nine chapters, one good
run and one collapsed run. Whether the guard fires on a real collapse, and does not
fire on a healthy Russian/French/Spanish book, is behaviour only a real book on the
box can prove.

---

**Handling.** 1–6 are fixed in the same round (branch
`fix/server-2325-guard-followups`), 7 is this issue's design question, 8 becomes a
register row in the same PR.

Refs #2324
Refs #2325

dudarenok-maker:
## Summary

PR #2333 shipped [#2324](https://github.com/dudarenok-maker/Castwright/issues/2324) (per-call handoff forensics) and [#2325](https://github.com/dudarenok-maker/Castwright/issues/2325) (the dialogue-collapse coverage guard). **It was merged before its mandatory review gate reported.** The gate then returned eight findings against already-merged code, two of which meant the PR's green suite was not evidence for what it shipped. This is that debt, paid.

Every finding was re-derived against the code before being acted on — which caught one that was wrong as originally recorded (see item 5).

## The retry could detect a collapse, then prefer the worst take and give up

`validateStage2Coverage` gained a `dialogueCollapse` term gating `ok`, and neither function that reads a verdict was extended for it.

- **`isBetterCoverage`** scored `(duplicatedBlock ? 100 : 0) + |1 - coverageRatio|`. A retry collapsing at 61% with ratio 0.97 scored *worse* than an attempt collapsing at 95% with ratio 1.00, so the better take was discarded.
- **`verdictSignature`** was `end/noend|dup|cov:x.xxx`. Collapses at 95% and 65% with ratios agreeing to three decimals produced an **identical** signature, tripping `deterministicFailure` and stopping the loop with "retrying cannot help" about a model that had just moved 30 points on the failing axis.

**The first attempt at this fix was itself wrong, in three ways an independent review measured**, and the corrected design is what ships here. An additive penalty term (a) was gated on `evaluable`, which losing the dialogue markers *also* zeroes — so a dash-stripped take (score 0.00) beat one collapsing at 61.7% (1.617); (b) had a floor of 1.60 against a ratio-deviation term capped at 1.0 on the low side, so a take with 88% of the chapter missing beat a complete-but-collapsed one — 12 sentences kept over 100; and (c) applied its signature term to *every* verdict at a granularity where one flipped `characterId` moves it, which on the ch8 repeat-loop shape meant **5 model calls instead of 2** and killed [#2304](https://github.com/dudarenok-maker/Castwright/issues/2304)'s early stop for every dash-dialogue language.

Replaced with a **severity tier plus a within-tier tie-break**. A lexicographic tier cannot exhibit (a) or (b) — a lower tier wins outright. The tier order reproduces the previous scoring for every pairing but one, reversed deliberately and tested: **`duplicatedBlock` now beats `noSentences`**, because a repeat-loop take still contains real prose and an empty one contains none. The old order was an accident of the `+100` constant.

## The guard failed OPEN on the output shape most likely to be broken

`classifyDialogueLine` tests the dialogue marker against the **attributed** sentence text. A model that drops or normalises the leading dash makes every line `indeterminate`, `speechHalves` falls to 0, `evaluable` goes false — and the guard passes silently, on output that lost *more* information than a narrator-collapse.

`bodyText` was already a parameter and unused. The source's own dash-opening count is now the control: a breach when the source has ≥ `STAGE2_MIN_SPEECH_HALVES` openers and the attribution recovers under half of them. Gated on `!truncated`, so a heavily-truncated take is not misattributed as a marker-dropping model. Calibrated against two real full-book runs of the same chapter (246→213 and 241→209), both far clear of the bar.

## Also fixed

- **A collapse no longer reports itself as a coverage gap.** It routed to `attribution-incomplete` — *"the analyzer's answer did not cover every sentence"*, *"a retry usually fills the gaps"*. Every sentence **was** covered; they were all handed to the narrator. New `attribution-collapse` failure code with copy describing what actually happened.
- **Both engine call sites for #2324 are now tested.** `ollama.ts:336` and `gemini.ts:316` are the only readers of `call.stage2CallSeq`, and reverting both left the entire suite green — the feature had coverage of its pure key helper and none at the seam where it takes effect.
- **The single-call path no longer destroys its own retry forensics.** #2324 numbered the chunked path; the single-call path passes no sequence and `runStage2WithCoverageGuard` re-invokes the same lambda per retry, so every attempt reused `2-ch{n}` and `writeInbox` rm'd the previous outbox. The defect #2324 fixed, surviving on the other branch of the same `if`.
- **Item 5 was overstated when filed and is corrected in the issue.** It was recorded as "a re-split blinds the guard, so the chapter is reported clean". It does not: `stage2-chunk.ts:461` recomputes the verdict over the full body and all sentences. The real defect is narrower — remediation runs only at *section* level, so a chapter collapsing in aggregate is flagged with no retry ever attempted.

## Not fixed here, and why

**Item 7 — first-person novels inflate the metric.** The mechanism intended to handle this (`findFirstPersonCharacter` → `firstPersonId`) **has never fired on any book on this box**: 0 of 23 `cast.json` carries a first-person alias, including all four Ночной дозор analyses. Stage 1 and stage 2 also disagree about where first-person *narration* goes, latently, because the stage-2 block never renders. Switching it on moves all first-person narration off the narrator voice — audible, not cosmetic. Owner decision 2026-08-13: **measure first**. The offline half is done and recorded on the issue; it returns a negative — exactly one book on this box has an evaluable speech population, so no offline work can widen the calibration.

## Test plan

- `npm run test:server` — **7,574 passed**; frontend **4,756 passed**; `npm run typecheck` clean (both).
- `stage2-coverage.test.ts` 42 → **51** tests; new suites on both engines, the chunker and the route.
- **Every claim below is mutation-verified, and the mutation that did NOT redden is reported rather than quietly patched:**
  - the signature bucket, the `markersLost` tie-break, the signature gating and the tier ordering redden 1, 1, 5 and 2 tests respectively. **The first three were found *green* under mutation in the previous round** — tests that could not fail.
  - reverting either engine call site (`ollama.ts:336`, `gemini.ts:316`) now reddens its own test with `ENOENT …-stage2-ch1-c3.md`. Before this branch, reverting **both** left the entire suite green.
  - reverting `analysis.ts:2249` — the wiring that carries the sequence between the chunker and the engines — reddens with `expected undefined to be 1`. Before this branch that left the entire suite green too, so the #2324 chain had two untested links, not one.
  - reverting the single-call retry numbering reddens with `expected [undefined, undefined, undefined] to deeply equal [undefined, 2, 3]`.
  - **one mutation did not redden on the first attempt** — the `speech.evaluable` gate — because the test asserting it also set `truncated: true`, so a different branch won regardless and the test proved nothing about the gate it was named for. The test was rewritten and the mutation re-run before it was accepted.
- **On-box acceptance row C4 added**, with all three surfaces moving together: the row, the glance table (Group C 3 → 4, total owed 68 → 69), and the hand-authored live view with its derived figures recomputed. `npm run check:onbox-register` passes, and it cross-checks the two files against each other.

Closes #2342
Refs #2324
Refs #2325

---

## Review rounds (two gate passes, six commits)

**Pass 1 — blocking.** The tier scheme I specified had a **second, undeclared ordering reversal**, and the comment arguing otherwise contained a false proof: it claimed `sourceEvaluable` bounds `|1 - coverageRatio|` under 100, but `sourceEvaluable` is `bodyWords.length >= 5` — it bounds the *denominator*. An excess repeat-loop at ratio 110 scored 109 and outranked a duplicated-block take at 100.6, so the guard kept **550 sentences for a 10-word source**. Reviewing the design again turned up a **third** reversal in the overlapping `(0.4, 0.6]` band that neither the gate nor the comment listed.

Fixed by collapsing the two damaged tiers into one whose tie-break **is the pre-#2342 sum verbatim** — `(dup ? 100 : 0) + |1 - ratio|` — so within-tier ordering is preserved by construction rather than by argument.

Pass 1 also found: the single-call retry counter colliding with the re-split counter (the re-split's second call deleted the very attempt that justified the split); two of C4's acceptance criteria observing numbers nothing logs, so the row would have discharged on a null observation; and the subset (Retry) job discarding `coverage` entirely, so a chapter that collapsed again came back clean — which is what the new `attribution-collapse` copy tells the user to do.

**Pass 2 — no blocking, no production correctness bug**, but it found that the enumeration test replacing the false proof **had a hole exactly where the constant it guards decides**: `HIGH` skipped `|1 − ratio| ∈ (51, 101]`, so mutating the `+100` to `50` left the suite green — pass 1's own reversal class, undetected by the oracle built to detect it. And "exactly two cross-tier reversals" was wrong for the third time; the enumeration itself produces counterexamples.

## How the ordering claim is verified now

Not by prose, and not by a proof. The test builds a verdict grid, compares the old scoring against `isBetterCoverage` for **every ordered pair**, and fails loudly on any **same-tier** disagreement not on a documented list. Cross-tier disagreements are the design rule — a lower tier wins outright — stated as such instead of counted wrongly. The eleven shapes it actually observes were read from its output, not reasoned about.

The `+100` constant is **bracketed from both sides by two derived values**:

| grid value | deviation | old sum prefers | catches |
|---|---|---|---|
| `excessOnly@100.5` | 99.5 | the excess take | `K → 99` (downward) |
| `excessOnly@101.5` | 100.5 | the dup take | `K → 101` (upward) |

They are a matched pair because the old-score mirror's `100` is fixed and does not move with the mutation, so each value only ever trips in one direction. A first attempt at this used `100.5` alone; `100 → 101` stayed green, the implementer **stopped rather than adding values until something failed**, and its explanation showed the trip wire was one-sided by construction. Both directions now redden:

```
100→50   RED   dupOnly/1/none vs excessOnly/60/none     (old=b, new=a)
100→99   RED   dupOnly/1/none vs excessOnly/100.5/none  (old=b, new=a)
100→101  RED   dupOnly/1/none vs excessOnly/101.5/none  (old=a, new=b)
100→1000 RED   dupOnly/1/none vs excessOnly/101.5/none  (old=a, new=b)
BASE     green (54/54)
```

Every one of those was re-run independently by the dispatching session against the real `verdictTieBreak`, with the worktree and committed blob verified clean afterwards.

**Also fixed, found in passing:** the subset (Retry) job never re-reported coverage — pre-existing, but this PR's new remediation copy is what turns it into a promise. Declared in both release-notes files.


dudarenok-maker:
## PR review — pass 1 (head 48e89516, effort high)

Scope: `stage2-coverage.ts`, `stage2-chunk.ts`, `analysis.ts` (`selectStage2FailureCode` + the `callForBody` wiring + both job loops), the five `FailureCode` enumeration surfaces, all six touched test files, the register + live view + release-notes pair. Verified: re-ran `vitest run src/analyzer/stage2-coverage.test.ts src/analyzer/stage2-chunk.test.ts` (97 passed) and `npm run check:onbox-register` (OK), then probed the new logic directly with `tsx` against the real modules before reasoning about it. No tracked file touched.

### 🔴 Blocking — there is a SECOND, undeclared ordering reversal, and the comment's proof that there isn't is factually false *(correctness bug)*

`server/src/analyzer/stage2-coverage.ts:463-473` claims:

> `truncated`/`excess` require `sourceEvaluable`, the only source of an unbounded ratio term, so `|1 - coverageRatio|` stays under 100 for every take that can reach tier 1, matching the original `+100` dup constant exactly

`sourceEvaluable` is `bodyWords.length >= STAGE2_MIN_EVALUABLE_WORDS` (`:316`, floor 5). That bounds the *denominator*, not the ratio. The module's own header (`:146`) records ratio **702.50** from the Ночной дозор ch6 incident; the only reason that one is not tier 1 is that its source was 2 words. A 5–13-word evaluable span with the same runaway lands at ratio 100–300 **and** is `excess` → tier 1. So the tier-1 tie-break can exceed 100, and the tier-1-vs-tier-2 pairing flips.

Reproduction (run against the branch's own modules):

```
body   = "alpha bravo charlie delta echo foxtrot golf hotel india juliet"  (10 words)
take A = 8 sentences, 4 repeated at a constant offset  -> dup: true,  ratio 1.6,  old score 100.6
take B = 550 distinct sentences                        -> dup: false, ratio 110,  old score 109.0  (excess)

OLD (pre-#2342) score       -> prefers A (the dup take)
NEW isBetterCoverage(B, A)  -> true   (tier 1 beats tier 2)
```

Driven end to end through `runStage2WithCoverageGuard` (attempt 1 = A, attempt 2 = B, `maxRetries: 1`), the guard now **keeps 550 sentences / ratio 110 for a 10-word source**, where the pre-#2342 scoring kept the 8-sentence take. The retained take is what `runMainAnalyzerJob` persists as "the best take", so this is not internal bookkeeping.

This also falsifies the PR body's central claim — *"The tier order reproduces the previous scoring for every pairing but one"* — which is the only verification offered for the load-bearing new logic. Either declare and justify this second reversal (and pin it with a test, as the `noSentences`/`duplicatedBlock` one is), or make tier 2 win when the tier-1 take's ratio deviation exceeds the historical `+100`. Silently shipping it under a comment that asserts the opposite is the worse of the two.

`stage2-coverage.ts:463-473`, `:474-479`, `:509-515`.

### 🟠 Significant — the single-call retry numbering collides with the re-split's counter and destroys exactly the forensics item 6 adds *(correctness bug)*

`stage2-chunk.ts:496-498` says the re-split path "runs through `callSectioned`, which has its own **independent** counter". It is independent, but it writes into the **same key namespace** (`stage2HandoffKey`, `handoff/protocol.ts:54` → `2-ch{n}-c{seq}`), starting again from 1. `writeInbox` (`protocol.ts:93-107`) `rm`s the previous outbox, errors and both raw-attempt files for a key before writing.

Reproduction (single-chunk body, `coverageRetries: 1`, a take that fails identically each time → `deterministicFailure` → forced re-split), observed `callSeq` and the keys they produce:

```
callSeq : -, 2, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
keys    : 2-ch7, 2-ch7-c2, 2-ch7-c1, 2-ch7-c2, ...
                 ^^^^^^^^^            ^^^^^^^^^  second write rm's the first
```

The single-call **attempt 2**'s prompt/response — the failing attempt whose evidence justified the re-split — is deleted by the re-split's **second section call**. On `main` this collision did not exist (single-call attempts all wrote the bare `2-ch{n}`; sectioned calls wrote `-c1…`), so the branch trades one clobber for another rather than removing it, and it does so on the deterministic-failure path, i.e. the ch8 shape #2304 exists for and the one case where the forensics are the whole point. Give the single-call attempts a disjoint sequence (e.g. seed `callSeq` from `singleCallAttempt`, or use a distinct suffix), and assert non-collision across the split boundary rather than only within the retry loop — the new test in `stage2-chunk.test.ts` only inspects the retry sequence and never reaches the split.

`stage2-chunk.ts:488-511`, `:343-356`.

### 🟠 Significant — two of C4's four acceptance criteria observe numbers the code never emits *(correctness bug, in the instrument)*

`docs/testing/onbox-acceptance-register.md:2770` — *"the per-chapter narrated-speech share **logged** for each chapter sits below 60%. Record the actual percentages"* — and `:2773` — *"the source's dash-opening count and the attributed speech-half count are **logged** for at least one chapter"*.

Nothing logs either. A grep over `server/src` (non-test) shows `narratedSpeech` / `speechHalves` / `sourceSpeechHalfCount` reaching exactly three consumers: the two issue strings inside `validateStage2Coverage` (`:375`, `:409`), which are built **only on a breach**, and `selectStage2FailureCode`. `runMainAnalyzerJob` logs `coverageVerdict.issues[0]` only inside `if (!coverageVerdict.ok)` (`analysis.ts:5032`). On the healthy book the row is chiefly about, the operator has no number to record and no count to compare, so the row discharges on "no `attribution-collapse` appeared" — a null observation, which is the failure mode the row's own preamble ("the distribution is worth far more than a pass/fail") is written to prevent. Either add the per-chapter log line this round, or rewrite the criteria against something that exists.

### 🟠 Significant (pre-existing, not introduced here) — the Retry the new `attribution-collapse` copy prescribes runs on a path that cannot re-report the failure

`failure-remediations.ts:117` tells the user *"Click Retry on this chapter to re-run attribution."* Retry runs the subset job: `analysis.ts:6405` clears the chapter from `failedChapterIds` and emits `chapter-resolved` once stage 1 succeeds, and the subset stage-2 loop at `:6649-6699` destructures only `sentences` / `chunkCount` / `structureReport` — it never reads `coverage`, never calls `selectStage2FailureCode`, and never calls `recordFailedChapter`. So a chapter that collapses again on retry comes back **clean**: failure cleared, no new record, no SSE `chapter-failed`. This predates the branch (the same hole swallowed `attribution-incomplete`), but the branch is what makes the remediation copy specifically promise this loop works.

### 🟡 Minor

- **`markersLost` blames the marker on an EXCESS take too** — `stage2-coverage.ts:400-406` gates on `!truncated` only, so the symmetric half of the coverage band is unguarded. Measured: 60 dash-opening source lines, a repeat-loop attribution of 400 narration sentences (ratio 9.33) yields both `Excess coverage — … likely a repeat-loop` **and** `Dialogue markers lost — … the model likely dropped or reformatted the dialogue marker`. Same misattribution nit 7 was raised for; the cause is looped prose, not a dropped marker. `ok` and the failure code are unaffected, so this is a wrong operator message, not a wrong verdict.
- **`selectStage2FailureCode`'s disjointness claim is false** — `analysis.ts:2413-2416` states *"Neither of those two shapes overlaps `noSentences`/`truncated`/`excess`"*. `markersLost` overlaps `excess` (above) and `duplicatedBlock` (the branch has a test for that combination). The code handles it; only the comment is wrong.
- **"A lexicographic tier + tie-break cannot exhibit either failure mode" is overclaimed** — `stage2-coverage.ts:446-450`. Round 1's failure (a) survives inside tier 1: `markersLost` is suppressed by `!truncated`, and tier 1's tie-break is `|1 - coverageRatio|` alone, so a truncated dash-stripped take at ratio 0.58 still beats a truncated marker-intact take at 0.55. Not a regression (the pre-#2342 scoring did the same), but the class the comment says is closed is not closed.
- **`onbox-acceptance-register.md:2777`** — *"the calibration in `docs/features/247-dialogue-structure-attribution.md`"*. That plan contains no mention of #2325, the collapse guard, the 60% threshold or the 246→213 / 241→209 numbers; no `docs/features` file mentions #2325 at all. The pointer sends an operator to an unrelated plan.
- **PR body under-declares** — item 5's residual ("remediation runs only at *section* level, so a chapter collapsing in aggregate is flagged with no retry ever attempted") is stated under "Also fixed" but is not fixed and does not appear under "Not fixed here, and why". The real shape is a chunked chapter whose sections each sit under `STAGE2_MIN_SPEECH_HALVES` while the aggregate is evaluable and breaching — worth naming explicitly wherever it lands.

### ✅ What is solid

- **Enumeration parity is complete and mechanically pinned.** `openapi.yaml`, the regenerated `api-types.ts`, `failure-taxonomy.ts`, `failure-remediations.ts`, both `help-failures.ts` maps and all three derived counts (21→22, 47→48) move together; `_copyComplete` (`failure-taxonomy.ts:55`) and the two `satisfies Record<FailureCode, …>` pins make a future omission a typecheck failure. No exhaustive switch was left behind — `isHelpLinkable` (`analysing.tsx:48`) and `helpHrefForFailureCode` are data-driven off `HELP_FAILURE_ENTRIES`, and `CATEGORY_WIKI` keys on category, not code.
- **The counters cannot leak across chapters or slots.** `callSeq` and `singleCallAttempt` are both locals of one `runStage2ChapterChunked` invocation, and `callForBody` spreads rather than mutates the shared `StageCall` (`analysis.ts:2255`) — the property the comment claims is the property it has.
- **The new tests are genuinely mutation-sensitive.** I traced each to the production expression it depends on: the tier tie-break (remove it → the 61%-retry test keeps the 95% take), the signature gate (ungate → the ch8 parity test runs 3+ calls, not 2), the 10-point bucket (whole-percent → the 75.12/75.59 stop fails), the `sourceSpeechHalves >= 20` floor and the `!truncated` gate (each asserted directly on `markersLost`), and the `evaluable` gate in `selectStage2FailureCode`'s un-evaluable case (deliberately built with no other failing gate, so it cannot pass via `isIncomplete`).
- Register, glance table and hand-authored live view move as a set and `check:onbox-register` agrees; the release-notes pair is present; `Closes #2342` is outside any code span.

### Verdict

**Changes requested.** One blocking and three significant correctness findings, three of them (the tier reversal, the handoff-key collision, the unobservable C4 criteria) inside the work this PR exists to do. The blocking one is cheap either way — the fix is either two lines of ordering or an honest comment plus a test — but it must not ship under a comment asserting the pairing is unchanged, because the next round will read that comment as verified. Re-review after the correctness findings are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head 6b7ed6f1, effort high)

Scope: the delta commit `6b7ed6f1` only (7 files), plus the house gates over the whole branch · verified: rebuilt the enumeration test's grid, `oldScore` and `tierOf` outside the repo and re-ran its classification loop against six mutations of `verdictTier`/`verdictTieBreak`; diffed the tier-1 tie-break against `git show main:server/src/analyzer/stage2-coverage.ts` line 335; ran `stage2-coverage.test.ts` + `stage2-chunk.test.ts` (101 passed) and both new route tests; ran `npm run check:onbox-register` (OK) — before probing the grid for holes.

### Pass-1 dispositions

| Pass-1 finding | Disposition |
|---|---|
| 🔴 second undeclared ordering reversal + false proof | **VERIFIED RESOLVED** (see ✅ below) |
| 🟠 callSeq collision | **VERIFIED RESOLVED** |
| 🟠 C4 observed numbers, nothing logged | **VERIFIED RESOLVED** |
| 🟠 subset (Retry) job discarded `coverage` | **VERIFIED RESOLVED** |
| 🟡 `markersLost` gate / comment overclaims / C4 pointer / item-5 residual | resolved, except the C4 pointer — **still wrong, differently** (🟠 3 below) |

**No blocking findings. No production correctness bug.** Everything below is cleanup-class: one test-coverage hole and three documentation/instrument defects. Per the triage contract this pass should not re-trigger a review round.

---

### 🟠 Significant — the enumeration grid has a hole exactly where the `+100` constant decides the order, so the "old formula verbatim" property is not actually pinned

*cleanup / test-robustness — `server/src/analyzer/stage2-coverage.test.ts:997`*

`HIGH = [1.7, 5, 110, 702]` jumps from `|1 − ratio| = 4` to `109`, skipping the entire band `(51, 101]` — the only region where the dup take's `+100` floor is the deciding term against a non-dup take. So the constant the whole merged-tier design rests on is untested.

Concrete: change `100` → `50` at `stage2-coverage.ts:530`. `isBetterCoverage` then prefers a `duplicatedBlock` take at ratio 1.0 over an `excess` take at ratio 60 (mutated score 50 vs 59), where the pre-#2342 sum preferred the excess take (100.6 vs 59) — the same reversal class as pass 1's blocker, in a shape that is reachable (an evaluable 5-word span the model loops to ~300 words). **The enumeration test stays green.** Measured on a faithful replica of the grid, `oldScore` and `tierOf`:

```
GREEN  BASE (shipped)                      fails=0
GREEN  M2 dup const 100 -> 50              fails=0     <-- the hole
RED    M1 tie-break drops +100             fails=384
RED    M3 dup const 100 -> 1000            fails=188
RED    M5 tier 0 <-> tier 1 swapped        fails=616
RED    M6 |ratio| instead of |1-ratio|      fails=224
```

Adding one element — `60` — to `HIGH` closes it: `M2` then reddens with 92 failures (`dupOnly/1/none vs excessOnly/60/none (old=b, new=a, tier 1 vs 1)`) and BASE stays green. One-line fix.

### 🟠 Significant — `verdictTier`'s comment makes a *third* precise "exactly TWO reversals" claim, it is false, and the new test explicitly does not check it

*cleanup / documentation — `server/src/analyzer/stage2-coverage.ts:484`*

The comment closes with *"Do NOT trust a third claim of 'every other pairing is unchanged'"* four lines after making one: *"That leaves exactly TWO cross-tier reversals from the pre-#2342 sum"*, enumerated as (a) `duplicatedBlock` beats `noSentences` and (b) tier-0 beats `truncated`/`excess`. The grid produces at least two more reversal shapes that are neither:

- `excessOnly/5/none` vs `noSentences` — old sum 4.0 vs 1.0 preferred `noSentences`; new prefers the excess take. Not (a): no duplicated block. Not (b): neither is tier 0.
- `clean/110/collapseBreach` vs `noSentences` — old sum 109 vs 1.0 preferred `noSentences`; new prefers the tier-0 take. Not (a); not (b), which names tier 1 as the loser.
- `truncatedOnly/0/none` vs `noSentences` — old sum ties at 1.0 exactly; new prefers the truncated take. A tie converted to a strict preference.

The test does not backstop this: its category-1 branch asserts only *"the lower tier wins"* and its own comment says it *"doesn't re-derive which specific pairing is 'deliberate' by label"*. So the enumeration is unbacked prose again, in the exact place the round exists to stop producing unbacked prose. Fix is documentation, not behaviour — the shipped ordering ("anything beats nothing at all, regardless of magnitude") is defensible; state *that* rule instead of two flag-specific instances.

### 🟠 Significant — the C4 row's replacement "where the criteria live" pointer is wrong in the same way as the one it replaced, and points at a differently-measured number

*cleanup / documentation — `docs/testing/onbox-acceptance-register.md:2777`*

The row now says the module header and the `STAGE2_MAX_NARRATED_SPEECH_PCT`/`markersLost` comments *"carry the measured numbers this row cites (39.3/72.2, 246→213/241→209)"*. `grep -rn "39.3\|72.2"` over the tree returns nothing; `stage2-coverage.ts:160-161` carries **95.7 / 67.9**, which is a *different metric* — narrator share across **all sentences**, not the per-chapter **speech-half** share that the 60% threshold gates.

Failure scenario: an operator running C4 follows the pointer to confirm the calibration, reads "the good run of the same book was 67.9%", and concludes the 60% threshold sits ~4 points under the good run rather than in the row's stated 33-point gap (39.3 → 72.2) — i.e. reads the guard as having almost no headroom, or worse, as one the *good* run breaches. The `markersLost` half of the citation (246→213, 241→209) is correct and does appear.

### 🟠 Significant — the PR body and `docs/release-notes-next.md` still ship the round-2 design this commit deleted, and the round-3 user-visible change has no note

*cleanup / release-notes gate*

- PR body: *"The tier order reproduces the previous scoring for every pairing but one, reversed deliberately and tested: `duplicatedBlock` now beats `noSentences`"* — this is the claim round 3 measured false and replaced. Also *"Gated on `!truncated`"* (now `!truncated && !excess`) and stale counts (7,574; 42→51).
- `docs/release-notes-next.md:455`: *"a severity TIER (`noSentences` > `duplicatedBlock` > `truncated`/`excess` > clean-prose-but-miscast)"* — the four-tier scheme that no longer exists; same `!truncated` overclaim.
- **Neither notes file mentions the subset (Retry) job's new re-report**, which is user-visible: a chapter that still collapses on Retry now shows as failed instead of appearing resolved. Checklist step 5 wants a line in both files or an explicit not-applicable.
- The round-3 incidental fixes are not declared in the PR body at all — no "Also fixed, found in passing" section, so a reviewer reading the body will not know the tier scheme, the markersLost gate, the callSeq counter and the subset job all moved after the body was written.

### 🟡 Minor

- **`server/src/routes/analysis.ts:5066-5069` and `:6760-6763`** — the parenthetical attaches to the wrong noun. The line renders `… source has 246 dash-opening speech lines (below the 20-line floor — population too small to judge)`; the floor applies to the **attributed** `speechHalves`, not the source count printed immediately before it. That reads as "the source is too small to judge" in precisely the marker-loss case (source 246, attributed 0) the line exists to make visible.
- **`server/src/handoff/protocol.ts:52-53`** — *"present numbers the call, so a chunked chapter reads `2-ch{n}-c1 … -c4` and you can tell from the filenames alone that the chapter was sectioned"* is now false, and this branch made it false. Before `6b7ed6f1` a single-call retry started at `2`, so `-c1` still implied section 1; sharing the counter means a single-call chapter's first coverage retry writes `-c1`. `docs/release-notes-next.md:450` repeats the same claim ("a chunked chapter is now identifiable from the filenames alone").
- **`analysis.ts:5065` / `:6759`** — the `sourceHalves ?? '?'` branch is dead: `narratedSpeech` is non-null only when `fpConventions?.dialogueOpen` was supplied, and `fpConventions` and `legibilityConventions`/`subsetLegibilityConventions` both resolve `conventionsFor(bookLanguage)` from the same value (`analysis.ts:2224`, `:4613`, `:6670`; stage-2's `StageCall.language` is `bookLanguage` at `:4891`/`:6706`). Harmless; noted so it is not mistaken for a live fallback.

### ✅ What is solid (deltas only)

- **The pass-1 🔴 is properly closed.** `verdictTieBreak`'s tier-1 arm is character-for-character the pre-#2342 score (`git show main:…:335` → `(v.duplicatedBlock ? 100 : 0) + Math.abs(1 - v.coverageRatio)`), so same-tier-1 ordering matches the old function by construction — the replica confirms **zero** same-tier-1 disagreements across all 3,306 ordered pairs. Nothing else consults the old score (`verdictSignature` reads only `coverageRatio.toFixed(3)`); the `noSentences` case is consistent (empty output forces ratio 0 in both).
- **The enumeration test is a real oracle, not a tautology.** Its `tierOf` mirror is hardcoded to the merged scheme rather than importing production, so reverting `verdictTier` to round 2's four tiers reddens it with 188 `UNDOCUMENTED disagreement shape` failures (first: `dupOnly/1/none vs excessOnly/110/none`) — i.e. it would have caught pass 1's blocker. The tier-0 blanket amnesty is sound: `markersLost` > collapse-pct and collapse-pct ordering are each pinned by dedicated tests (`stage2-coverage.test.ts:724`, `:510`), so the mutations the amnesty lets through are covered elsewhere.
- **callSeq.** One chapter-scoped closure counter; attempt 1 still `undefined` (byte-identical filenames to `main`); no reuse across retry → forced re-split → nested re-split (`splitAndRetry` recurses through `callSectioned`, which shares it); no cross-chapter or cross-slot leak (per-invocation local, and the key embeds `chapterId`). The new collision test would redden on the old code (`[undefined, 2, 1, 2]` → `Set` size < length).
- **The C4 log line** cannot throw (`narratedSpeech` non-null guard; `dialogueOpen` has no `g` flag so `.test`/`.replace` are stateless), leaks no path, and the main/subset copies are equivalent modulo the chapter-label convention each route already uses. Its test is load-bearing: `failedByChapterId.size === 0` proves the passing path, and it asserts the real measurement (`6/22`, `27.3%`, `source has 22`).
- **The subset re-report** lands `recordFailedChapter` before the loop's own `saveAnalysisCache`, strictly after Phase 0's `clearFailedChapterId`, with no later clear; no double-report path (main and subset are separate jobs). The test's `failedByChapterId.has(1)` assertion cannot pass without the new `send` — Phase 0 succeeds in that fixture and emits nothing.
- `markersLost` × `duplicatedBlock` co-occurrence pinned in both the validator and `selectStage2FailureCode`; `check:onbox-register` green; no `cast.json`/lock surface in this commit; 101 analyzer tests + both new route tests pass locally.

### Verdict

**Ship-able.** All four pass-1 findings verified resolved; the code that ships is correct as far as I can break it. The residue is one grid element away from closing the last untested corner of the tie-break, plus three documents that describe a design this commit replaced — two of them repeating the exact "precise prose claim, unbacked" pattern this round was convened to eliminate.

PR #2351, opened 2026-08-13
fix(frontend): surface model-manager Load/Stop failures (#2349)

### LINKED ISSUE #2349: fe-53 — background actions in model-manager surface failures nowhere: unhandled rejection on a fire-and-forget callback
## What

`src/views/model-manager.tsx:771` calls `await api.restartSidecar()` inside `ModelRow`'s `onInstalled` callback. That callback is invoked **fire-and-forget** — `onInstalled?.()`, unawaited — from `kokoro-install.tsx`, `qwen-install.tsx` and `coqui-install.tsx`, with no `try`/`catch` anywhere on the path. A throw becomes an unhandled promise rejection: the user sees nothing, and the failure surfaces only in the console.

Surfaced during the PR #2348 review, when `realRestartSidecar`'s thrown message was corrected (#2344). **Pre-existing** — that change altered the message's content, not whether or where it throws.

## Why this is filed rather than fixed in the same round

Investigated for a same-round fix and stopped, because there is no existing pattern to apply — only a wider gap.

`model-manager.tsx` does have two local error-surfacing patterns: `ModelInventory`'s background-refresh error banner, and `ConfirmRemoveModal`'s foreground-action error. Neither fits an unawaited callback fired from a *child* component, and composing either onto `onInstalled` needs new prop plumbing.

More to the point, the same shape is systemic in that file rather than a one-off miss: the Load/Stop pills (`ModelControlPill`, typed `onLoad: () => void`) have the **identical** fire-and-forget-onClick shape and surface nothing today either. Fixing line 771 alone would leave the class armed next door — the same criticism the #2348 review correctly made of a guard that scanned one directory and not its sibling.

**The decision owed:** how should a *background* action's failure surface in this view — a toast via `notifications-slice`, an inline per-row banner, or a deliberate silent-with-console? And is that answer applied to the whole class (`onInstalled` **and** the Load/Stop pills), or only where a user initiated the action directly?

That is a UX decision with more than one defensible answer, spanning several controls. It is not "weighing two implementations of one agreed behaviour".

## Acceptance

- The decision above is recorded.
- No fire-and-forget action in `model-manager.tsx` can fail with the user seeing nothing — whichever surface is chosen, applied to the whole class, not just line 771.
- A test pins it for at least the `onInstalled` path and one `ModelControlPill` path.

## Key files

`src/views/model-manager.tsx` (:771 `onInstalled`, the `ModelControlPill` render sites), `src/components/{kokoro,qwen,coqui}-install.tsx` (the unawaited `onInstalled?.()` call sites), `src/store/notifications-slice.ts`.

## Benefit

*Benefit (user):* a failed background restart or model load stops being invisible — today the UI returns to rest looking exactly as it does on success.

dudarenok-maker:
﻿## Summary

The Model Manager's background Load/Stop actions swallowed every failure silently. The shared `onAction` handler ran the sidecar/ollama load/unload inside try/finally and ignored the returned `ModelControlResult`, so a `status:'error'` body or a thrown fetch error flipped the pill back with zero user-visible indication. Because every row (TTS voice engines, analyzer/Ollama rows, ASR) and both Load and Stop pills funnel through that one handler, this was systemic — not a single-line spot fix.

This surfaces failures through the app's global toast surface (`notificationsActions.pushToast` → `<ToastStack>`): the handler now inspects `action()`'s result and toasts the server's `error` text (or a `{Load|Stop} <label> failed` fallback), and catches the thrown-network path that the previous `finally` otherwise papered over. A failed action no longer triggers a pointless inventory refetch.

Closes #2349

## Test plan

- [x] Pre-commit `verify:fast:scoped` — frontend suite green (335 files / 4776 tests), `test:server` skipped as out-of-scope
- [x] Pre-push / `verify:fast:branch` — lint, typecheck, build all pass; remaining steps scope-skipped; re-run fully cached
- [x] Transient warnings ("Selector unknown…") are pre-existing (baseline emits 51), not introduced by this change

### DIFF
--- a/src/views/model-manager.test.tsx
+++ b/src/views/model-manager.test.tsx
@@ -10,6 +10,7 @@ import { accountSlice, type AccountState } from '../store/account-slice';
 import { uiSlice } from '../store/ui-slice';
 import { settingsSlice } from '../store/settings-slice';
 import { upgradeSlice } from '../store/upgrade-slice';
+import { notificationsSlice } from '../store/notifications-slice';
 import { ModelManagerView } from './model-manager';
 import { api, type ModelControlResult, type ModelInventoryResponse } from '../lib/api';
 import type { UserSettings } from '../lib/types';
@@ -179,6 +180,7 @@ function renderManager(initial: Partial<UserSettings> = {}) {
       ui: uiSlice.reducer,
       settings: settingsSlice.reducer,
       upgrade: upgradeSlice.reducer,
+      notifications: notificationsSlice.reducer,
     },
     preloadedState: { account: preloaded },
   });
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
## PR review — pass 1 (head 21e4532d, effort medium)

Reviewed `src/views/model-manager.tsx` and `src/views/model-manager.test.tsx` (the only two files in `HEAD~1..HEAD`), cross-checked against `src/lib/api.ts` (`ModelControlResult`, `realLoadSidecar`/`realUnloadSidecar`/`realLoadAnalyzer`/`realUnloadAnalyzer`), `src/store/notifications-slice.ts` (`pushToast`/`ToastKind`), the server `/api/sidecar/*` + `/api/ollama/*` response shapes, and the `ModelControlPill` component. Ran the suite: `model-manager.test.tsx` + `ModelControlPill.test.tsx` — 59/59 green.

### House gates
- `Closes #2349` present in the PR body, outside any code span. Head SHA `21e4532d…` matches. ✓
- Paired tests are genuine regression tests: all three assert behavior the pre-fix handler lacked (no toast on `status:'error'`; no toast on a thrown action; refetch still fired on a failed action). Each is red on the old code and for the claimed reason. ✓
- Notifications slice is registered in `src/store/index.ts`, so the `dispatch(pushToast)` path is live in production (not just in the test harness). ✓
- Frontend-only change; server acceptance/checksum gates N/A. No derived artifacts or incidental findings.

### Findings — split correctness vs cleanup
No correctness bug found. The fix is functionally sound: the `status:'error'` branch surfaces the server error and the `catch` block is genuinely reachable (in `realLoadSidecar`/`realUnloadSidecar` the `await fetch()` sits *outside* the `.catch` that wraps `res.json()`, so a transport-level failure rejects and lands in `catch`).

#### Cleanup / low severity
1. **`src/views/model-manager.tsx:277‑283` — a real network-down toast is context-free.** When `action()` rejects (transport failure, pre-`res`), `e` is an `Error`, so the toast message is the bare `e.message` — in a browser that is `"Failed to fetch"` with no identification of which model/operation failed, because the `${verb} ${item.label} failed` fallback only runs for non-`Error` throws. Scenario: the box's Wi‑Fi drops, the user clicks "Stop Kokoro", and the toast reads just "Failed to fetch"; the `status:'error'` branch in the same handler gives richer output (`result.error`). Suggest including `verb`/`label` context in the thrown branch too.

2. **`src/views/model-manager.tsx:274` — the "error ⟹ nothing changed … don't refetch" comment overstates the invariant.** The sidecar `/load` proxy runs on a 90 s budget; a slow cold load (Qwen 1.7B) can make that proxy return `{status:'error'}` *after* the sidecar has actually finished loading. In that race the early `return` skips the immediate refetch and the pill keeps showing "Load model" for a still-resident engine. This is *not* a bug — the 30 s autonomous inventory poll (`INVENTORY_POLL_MS`, `model-manager.tsx:196‑199`) self-corrects the stale row within one cycle, and the user did get an error toast — but the PR comment's "nothing changed on disk/VRAM" is definitively stated where the proxy-timeout case contradicts it. Worth softening the comment (and worth knowing this is a transient-desync window, not a permanent one).

### Note (not a finding)
Test 3's call-count assertion (`callsBefore` captured before `renderManager()`) couples to the mount effect performing exactly one `getModelInventory` and no timer interference. Acceptable as written; slightly tight for future drift in the mount effects, but passes deterministically today.

dudarenok-maker:
﻿Addressing the pass-1 review findings (both **cleanup**, low severity — no correctness bug surfaced by review):

1. **`src/views/model-manager.tsx` — bare "Failed to fetch" toast on a thrown network failure.** Fixed in `69a14dc`: the transport-level catch now prefixes with op + model, e.g. `Stop Kokoro failed: Failed to fetch`, matching the richer `result.error` used in the `status:'error'` branch. The paired regression test now asserts this prefix too.

2. **`src/views/model-manager.tsx` — "nothing changed on disk/VRAM" comment overclaimed the invariant.** Softened to "assume nothing changed… don't refetch" in `69a14dc`, acknowledging the sidecar/cold-load can report `status:'error'` after a slow load actually completed (self-corrected by the 30s inventory poll).

Both addressed and pushed. These were cleanup-only, so no re-review pass is required per the gate.

Status: `verify:fast:branch` green (lint/typecheck/test/build), all pushes cached-green. GitHub checks re-running on the new head; the CLA-signature check remains pending on contributor signature.

Issue #1984, opened 2026-07-31
Attribution collapse is invisible: the dropped-quotes panel shows the last batch only, in a transient view, and never the effect


dudarenok-maker:
## What happened

A real book in the library — *Der Auftrag von Coalfall* (de) — lost **103 of its 144 quoted sentences to the narrator** during analysis on 2026-07-14. Dialogue collapsed from 178 lines to 41. Ten of thirteen cast members ended up with almost nothing to say. The book then sat that way for 17 days and nobody noticed, including me, until I went looking for something else.

The signal was not *absent* — but what was shown could not have communicated the damage.

## What the user would actually have seen

`DroppedQuotesPanel` (`src/components/analysing/phase-card.tsx:246-290`) renders inside the phase-0 card of the **analysing** view:

```tsx
const latest = file?.batches.length ? file.batches[file.batches.length - 1] : null;
if (!latest || latest.entries.length === 0) return null;
…
<details className="mt-3 text-[11px] text-ink/60">
  <summary …>Verifier dropped {latest.totalDropped} quote…across {latest.affectedCharacters} character…</summary>
```

Four things combine badly:

1. **Latest batch only.** This book accumulated **16** dropped attributions across **5** batches. The final batch dropped **1**. The user saw *"Verifier dropped 1 quote across 1 character"* — a rounding error, not a warning.
2. **Collapsed and de-emphasised.** A `<details>` at `text-[11px] text-ink/60`, styled to recede. Nothing about it says "your cast just lost most of its dialogue."
3. **Transient surface.** It exists only on the analysing screen. Once analysis finishes and the user moves to Cast or Generate, there is no residue — no badge, no banner, nothing on the book. Seventeen days later there was no way to discover this short of diffing JSON by hand.
4. **It reports the cause, not the effect.** "N quotes dropped" is a verifier-internal statistic. The number that matters to a user is *"103 of 144 quoted sentences are now attributed to the narrator"* — which is never computed or shown anywhere.

## Why the effect is the number that matters

The dropped-quote count and the resulting damage are not proportional. Dropping one quote can re-attribute a long run of dialogue; dropping ten can be harmless. In this book, 16 drops produced a 72% collapse. A user cannot infer the second number from the first, and the second is the one that decides whether the book is usable.

## Proposed

- **Compute and surface the effect.** After analysis, report the share of quoted sentences attributed to `narrator` (per chapter and per book). A book where most quoted sentences are narration is almost always broken, whatever the drop count.
[...]

Issue #2040, opened 2026-07-31
Cast/analysis characterId drift causes silent narrator-voice substitution (upstream of #2023)


dudarenok-maker:
## What

Cast ids and analysis ids can disagree for the same character — a
romanisation/slug mismatch between `cast.json`'s `characters[].id` and the
ids the analyzer/analysis-cache and rendered `<slug>.segments.json` files
carry for the same speaker. Confirmed on-box, *Заказ Коалфолла* ch.2:

```
cast.json ids       : [... 'coalfall-dragon', ... 'mairin', ...]
segments.json ids   : [... 'coalfall',        ... 'mayrin', ...]
analysis cache ids  : [... 'coalfall',        ... 'mayrin', ...]
```

`segments.json` and the analysis cache AGREE with each other but disagree
with `cast.json` — i.e. the drift is between the cast record and everything
downstream of analysis, not a render-time typo.

This is the upstream cause of #2023 (batch 2): a sentence group attributed to
`mayrin`/`coalfall` has no matching entry in `cast.json` at all, so
`synthesise-chapter.ts`'s orphaned-characterId safety net silently substitutes
the narrator for it. #2023 fixed the SYMPTOM (record + surface the
substitution, and gate a cloned narrator from ever borrowing an orphaned
line) but deliberately did not touch this root cause — see that issue's
"Piece 3 — FILE ONLY" scope note.

ECAPA corroborates these are literally the same character, not merely a
similar-sounding one: `narrator` vs `mayrin` **0.932**, `narrator` vs
`coalfall` **0.965**, `mayrin` vs `coalfall` **0.937** (for scale, one real
person across two different clips measures **0.891** — these "different
characters" are more alike than one person is to themselves).

## Evidence (from #2023's on-box run)

| book | orphaned segments | wrong-voice audio | orphaned ids |
|---|---|---|---|
| Playing with Fire | 74 / 6,638 (1.1%) | 2.8 min | `the-torment`, `pool-player-2`, `lightning-dave` |
| Unlocked | 34 / 9,596 (0.4%) | 2.1 min | `unknown-male` |
| Exile | 53 / 6,366 (0.8%) | 1.9 min | `silveny`, `sir-harding`, `unknown-female`, `unknown-male` |
| Заказ Коалфолла | 21 / 266 (7.9%) | 1.0 min | `coalfall`, `mayrin` |
| Everblaze | 6 / 11,498 (0.1%) | 0.9 min | `lady-alina` |

[...]

PR #2359, opened 2026-08-14
docs(docs,server): drop the `drifted` state (D13) from #1984 Wave 2

### LINKED ISSUE #2357: srv — decide whether the `drifted` state (D13) is built in #1984 Wave 2 or dropped
**Decision owed:** does the `drifted` state (D13) get built in #1984 Wave 2, or dropped?

#1984 closed when Wave 1 shipped (`090168a5`), but Wave 2's surface was never settled and D13's re-gate was still outstanding. That re-gate has now run — PR #2356, merged `73cf26bd` — and its verdict needs a home, which this issue is.

### Why this is a decision and not a fix

The spec pre-agreed a conditional: *"Wave 1 must re-measure D13 under the current unit; if the gap does not survive, D13 is dropped despite this answer"* (`§Open questions`, question 2). The re-gate does not evaluate that conditional cleanly — **the gap survived, but only its fixture half did** — so applying the clause silently would be over-reading it.

### What the re-gate measured

Full evidence: [run sheet §5](https://github.com/dudarenok-maker/Castwright/blob/main/docs/testing/attribution-collapse-visibility-onbox-acceptance.md), spec `§D13 re-gated → Re-gate outcome`.

- Raw shares still separate 16.5–49.2% (four books) from ≤2.5% (nineteen) — but **all four high books are *The Coalfall Commission* language editions**, and the mechanism is measured, not guessed: hand-edited `cast.json` renames that never went through `retireCharacterId`. `coalfall` (28 spans) and `oduvan` (22) are 50 of the top book's 62, both sitting in `cast.json.bak.prewipe-20260714`.
- **Natural experiment:** `Заказ Коалфолла` *records* the identical `coalfall → coalfall-dragon` rename and carries **zero** orphan spans for it; the English edition, where it was hand-applied, carries **28**.
- **Excluding that family the corpus maximum is 1.80%**, with 15 of 16 real books at exactly zero.
- The one real book clearing `MIN_ORPHAN_FOR_VERDICT` has **28 of 32 orphan spans (87.5%) in ids no user can link** (`driver`, `woman-in-taxi`, `vampire-boy`, `vampire-girl`); only `boris-igoryevich` (5 spans, 0.28%) is linkable.
- **D13's only exit has never been exercised** — no book in the corpus has a bare `rejected` entry, so `unacknowledgedOrphanSpoken === orphanSpoken` everywhere and R-8M1's acknowledgement path has zero real-data support.
- Positive result alongside it: **`retireCharacterId` works.** Fourteen retirements across five books, not one retired id is an orphan anywhere.

No threshold badges a real problem. Above 2.5% D13 fires only on a development fixture; at or below 1.8% it fires on one book whose badge cannot be cleared — R-8M1's "gate with no exit", arriving on the first real book D13 would fire on.

### Options

1. **Drop the `drifted` state; keep `orphanSpoken`/`orphanIds` as reported columns.** *(recommended)* The columns earned their place — they found four books with stale caches behind hand-edited casts on their first real run — but that is a developer-facing diagnostic, not a user-facing badge. Wave 2 ships four states, not five, and `alsoCollapsed`, the fifth library state, the fifth notice variant, `attributionVerdictKey` and the cache-sourced orphan-banner tier are not built.
2. **Gate on linkability rather than share.** Would fire on `boris-igoryevich` alone. Needs a linkability test that does not exist, and there is no corpus evidence to calibrate one either.
3. **Keep D13 unbuilt and revisit when a real library shows drift.** Produces the same Wave 2 as option 1; differs only in whether the door is documented as open.

Options 1 and 3 are the same code. The real choice is 1/3 versus 2.

### Acceptance

- The spec's `§D13 re-gated → Re-gate outcome` and `§Open questions` question 2 record the chosen option and the date.
- If D13 is dropped, the Wave 2 sections that exist only to serve it are struck from the spec rather than left as unbuilt scope.
- The Wave 1 plan moves to `docs/features/archive/` once Wave 2's surface is settled (it is deliberately unarchived today for exactly this reason — see its Ship notes).

### Key files

- `docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md`
- `docs/superpowers/plans/2026-08-13-attribution-collapse-visibility-wave1.md`
- `docs/testing/attribution-collapse-visibility-onbox-acceptance.md` §5

### Depends on

Nothing. The evidence is complete and merged.

### Benefit

*(technical)* Wave 2 cannot be planned while a fifth state's existence is undecided — and the honest answer is now measurable rather than assumed, which is what Wave 1 existed to produce.

Refs #1984

dudarenok-maker:
## Summary

Applies the **#2357** decision: the `drifted` state (design decision **D13**) is **dropped** from #1984 Wave 2, not deferred. Confirmed by the repo owner 2026-08-14.

The D13 re-gate (#2356, `73cf26bd`) measured that no threshold badges a real problem. Excluding *The Coalfall Commission* fixture family the corpus maximum orphan share is **1.80%**, with **14 of 16 real books at exactly zero** — so the order-of-magnitude bimodality D13's approval was made conditional on does not survive the re-basing. The one real book clearing `MIN_ORPHAN_FOR_VERDICT` has **28 of 32 orphan spans (87.5%) in ids no user can link** (`driver`, `woman-in-taxi`, `vampire-boy`, `vampire-girl`), so at any threshold low enough to fire on it, D13 is R-8M1's *gate with no exit* arriving on the first real book it would fire on. Its only exit has never been exercised: no book in the corpus has a bare `rejected` entry.

This fires the pre-agreed `§Open questions` question-2 caveat rather than opening a new question. It was put as a confirmation rather than applied silently because the condition's evaluation was not the clean yes/no that clause was written for — **the gap did survive; only its non-fixture half did not.**

### What Wave 2 no longer builds

Four states, not five. Struck with `drifted`: the fifth library state and the fifth notice variant, `alsoCollapsed` (R-8C3), `attributionVerdictKey` (R-8M2), `unacknowledgedOrphanSpoken` and the "not the same character" acknowledgement exit (R-8M1), the cache-sourced tier on the Cast orphan banner (R-8C1), and `DRIFT_SHARE_THRESHOLD` / `MIN_ORPHAN_FOR_VERDICT`. The dismissal reverts to the `analysedAt` identity, which is sufficient again because every surviving state is a function of the cache alone.

### What deliberately survives

- **`orphanSpoken` / `orphanIds`** stay exactly as Wave 1 shipped them — 27 references, none removed or deprecated. They earned their place on the first real run by finding four books with stale caches behind hand-edited casts; that is a developer-facing diagnostic, not a user-facing badge.
- **D9** is untouched (an unresolvable `characterId` never moves the collapse share), including **Wave 2 acceptance criterion 11's first half** — only its `drifted`-badging half is struck.
- **R-8C2** (the gate's move to `enqueueQueueEntries`) — raised for D13 but independently a `collapsed` hole, so re-tensed rather than struck.
- **Question 5 (`unanswered`) stays open.** The re-gate does not touch it.
- **The historical record.** `§D13 re-gated`'s evidence prose and all of `§Review findings` are preserved — the argument for dropping D13 *is* the justification, and rewriting it would erase the reasoning. What came out is forward-looking scope only.

### Two judgment calls worth flagging

1. **Step 5's slot is marked struck, not renumbered.** Prose and shipped code comments cite "steps 1–4 and 7", "step 4d" and "step 6"; renumbering would have silently falsified them.
2. **The Wave 1 plan is not moved to an archive.** #2357's acceptance says it should go to `docs/features/archive/`, but the plan lives in `docs/superpowers/plans/`, which has no `archive/` subdirectory and has never had a plan archived out of it. The clause is discharged by settling the plan's *"Not archived, deliberately"* ship-note — which only existed because #2357 was open — rather than by inventing a directory. Flagged here rather than done silently.

Three doc/comment sites this decision made false or stale — fixed here, not deferred:

- `server/src/store/attribution-health-io.ts` (×2) — *"Steps 4d/5/6 (`unanswered`/`drifted`/`collapsed`) are Wave 2"*.
- `server/src/store/attribution-health.ts:58` — `// ---- id drift (D9/D13) ----` cites a dropped decision. The two fields are unchanged; only the section marker moves.
- `docs/testing/onbox-acceptance-register-live-view.html` — the E11 row read **"owner confirmation owed"**, which is now flatly false on the *published* page. Mirrored in `onbox-acceptance-register.md` per the live view's own update-both rule.

**A pre-existing arithmetic error in the merged evidence, corrected here.** The re-gate's headline figure read *"15 of the 16 non-fixture books sit at exactly zero"*. It is **14 of 16**. The Coalfall family is **seven** language editions — including `コールフォールの依頼` (ja), which is itself at zero orphans and so never appears in the non-zero table — so `15` was `23 − 8`, the count of *all* zero-orphan books: the fixture exclusion was applied to the denominator but not the numerator, quietly promoting a fixture edition into the real-book tally. Two non-fixture books have orphans, not one (`Ночной дозор` 1.80%, `Playing with Fire` 0.34%), and the run sheet already said so ten lines further down. Corrected at spec `:1537`/`:2803`, run sheet `:219`, and the live view; the run sheet also gains a clause distinguishing this figure from the correct whole-corpus `15` four lines above it, since that adjacency is what produced the error. **The decision is unaffected** — maximum non-fixture share is still 1.80%, still one real book over `MIN_ORPHAN_FOR_VERDICT`, still 87.5% of its orphan mass unlinkable. Pre-existing from #2356 and untouched by the first commit, but "it's pre-existing" is a void deferral reason here, and the figure was about to be republished to the artifact URL.

**Two things this PR's own strike broke, found by the review gate and fixed:** the Wave 1 plan carried a **second, unstruck copy** of the state sequence (`5. orphan share ≥ DRIFT_SHARE_THRESHOLD → drifted`) 130 lines above the ship note this PR rewrote to announce the drop, plus `DRIFT_SHARE_THRESHOLD` still listed as a Wave 2 deliverable; and spec `:1021` claimed *"the orphan banner is the surface that can act on it"*, which this PR's own edit at `:986-1005` falsified by keeping the banner on rendered segments only. Also struck: R-9M2's normative drift-share denominator rule at `:1489`, which `:3097` already declared struck, and two smaller "drift surface" references.

One in-spec cross-reference the strike itself created, also fixed here: `unanswered` (question 5) was described as **"a sixth state"** in six forward-looking passages. With four states it would be a **fifth**. Two of those are surface-cost arguments, and the cost genuinely moved — dropping D13 makes question 5 *easier* to answer yes, and question 3's "the surface has grown too big" objection now cuts the other way — so they are re-argued rather than renumbered. The revision preamble and the R-9C4 findings row keep "sixth": they are historical.

## Test plan

- No behaviour changes. `AttributionState` is `'ok' | 'missing' | 'unmeasurable'` and never contained `drifted`; the two server edits are comment-only, so **no paired test is owed and none is deleted or skipped**.
- `npm run verify:fast:branch` — green (full selected-steps battery, not the docs-only fast-path, since `server/**` is touched).
- `cd server && npx tsc --noEmit` — exit 0.
- `npm run check:onbox-register` — OK, tracked pair agrees.
- `npm run check:onbox-register -- --against-published <live page>` — OK, register is not behind the published page. Run against the page fetched from the canonical artifact URL immediately before publishing.

**On-box acceptance: no change.** Register row **E11 stays owed** with the same two sub-items (the dash-stripped re-run invariance check, and re-analysing one book post-D18). Owed total stays **72**; no row, group count or ID moves. The live view changes by wording only and is republished to its canonical URL.

**Release notes: skipped, deliberately.** No user- or operator-visible delta — a decision record plus three false comments. This is the documented condition for skipping the step, stated here rather than omitted silently.

Closes #2357
Refs #1984


dudarenok-maker:
## PR review — pass 1 (head bf395bb2, effort high)

Scope: reviewed from worktree `C:\Claude\Projects\wt-d13-drop-decision` at `bf395bb2`, effort **high** (multi-scope `docs` + `server`). Full `git diff main...HEAD` read line-by-line (7 files, +310/−460), including all 979 diff lines of the design spec. Verified independently rather than from the PR body: over-strike (every claimed survivor), under-strike (case-insensitive sweep for `drifted` / `alsoCollapsed` / `attributionVerdictKey` / `unacknowledgedOrphanSpoken` / `DRIFT_SHARE_THRESHOLD` / `MIN_ORPHAN_FOR_VERDICT` / `orphanShare` / `alsoDrifted`, spec + tree-wide), cross-reference integrity, step-numbering, state-count arithmetic, every factual claim in the rewritten decision record against run sheet §5, the register pair, and both flagged judgment calls.

**Commands run (actual output):**

- `cd server && npx tsc --noEmit -p tsconfig.json` → **exit 0**
- `npm run check:onbox-register` → **OK — `onbox-acceptance-register.md` and `onbox-acceptance-register-live-view.html` agree**
- `npm run test:server` — **deliberately not run.** Both server edits are comment-only (confirmed by filtering the diff: no token outside a block comment changed) and `tsc` is clean. Running it would be ritual, not evidence.

---

### 🔴 Blocking

**1. "15 of the 16 non-fixture books sit at exactly zero" is off by one — it is 14 of 16.** *(correctness bug — the decision record's evidence)*

`docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1537` (also `:2803`), `docs/testing/attribution-collapse-visibility-onbox-acceptance.md:219`, `docs/testing/onbox-acceptance-register-live-view.html:1364`, and the PR body.

Repro, entirely from the merged evidence:

- Run sheet §1's table (`:52-74`) lists all 23 books. The *Coalfall Commission* family is **seven** titles: `The Coalfall Commission` (en, 62), `El Encargo de Coalfall` (es, 32), `La Commande de Coalfall` (fr, 23), `Заказ Коалфолла` (ru, 18), `煤落的委托` (zh, 3), `Der Auftrag von Coalfall` (de, 2), and `コールフォールの依頼` (ja, **0**). The run sheet's own §2 (`:126-127`) enumerates the family as `0/23/32/2/62/18` — the leading `0` **is** the Japanese edition, so the run sheet already counts it in the family.
- 23 − 7 = **16 non-fixture books** — the denominator in the sentence is right.
- Of those 16, **two** are non-zero, and §5a's own table labels both `real`: `Ночной дозор (Tetralogy)` 32/1776 = 1.80% (`:211`) and `Playing with Fire` 6/1753 = 0.34% (`:213`).
- Therefore **14** non-fixture books are at exactly zero, not 15. The `15` is the count of *all* zero-orphan books (23 − 8 non-zero = 15, stated correctly at `:215`), which includes the zero-orphan Japanese fixture — the fixture exclusion was applied to the denominator and not to the numerator.

What a reader concretely gets wrong: the sentence as written says exactly one real book in the corpus carries any orphan at all. The section three paragraphs below contradicts it — `:277-278` names `Playing with Fire`'s `pool-player-2` (6 spans, 0.34%) as a second real book's orphan id. The overstated figure is one of the four numbers the drop is argued from, it sits in the permanent decision record, and it is on a page **published outside the repo**.

The decision itself does not move — the maximum among real books is still 1.80%, and only one real book still clears `MIN_ORPHAN_FOR_VERDICT` — so this is a one-token fix (`15` → `14`) in four places, not a re-argument. It is pre-existing (merged in #2356, untouched by this diff), which under CLAUDE.md's "Incidental findings" rule is explicitly *not* a reason to defer it: *"it's pre-existing … adjacency stopped being the test."* Note the live view must be corrected **before** it is republished, or the wrong figure goes to the artifact URL as part of this round.

---

### 🟠 Significant

**2. The Wave 1 plan still carries live `drifted` Wave 2 scope, 130 lines above the ship note that cancels it.** *(correctness bug — under-strike, second copy)*

`docs/superpowers/plans/2026-08-13-attribution-collapse-visibility-wave1.md:820`, `:828`, `:948`.

This PR edits that file (Ship notes, `:962-969`) to say **"Wave 2 ships four states, not five"**. But the same file still says, unmarked:

- `:948` — under `## What Wave 1 deliberately does not ship`: *"No threshold constant. `COLLAPSE_SHARE_THRESHOLD`, `DRIFT_SHARE_THRESHOLD` **and all four floors** are set from Task 8's run, in Wave 2."* Both halves are now false: `DRIFT_SHARE_THRESHOLD` is never set, and three floors survive, not four.
- `:820` — the state sequence, verbatim: `5. orphan share ≥ DRIFT_SHARE_THRESHOLD → drifted (Wave 2 — no threshold yet)`
- `:828` — *"Steps 5 and 6 have no threshold yet."*

This is the **second copy** of the sequence the PR correctly struck in the spec (`:1994-1998`) and annotated in `attribution-health-io.ts:185-189` — the recurring "one instance fixed, the class left armed" shape. Concrete failure: the Wave 1 plan is the natural predecessor document a Wave 2 implementer opens, and its task body tells them a fifth state and its threshold are theirs to calibrate. The ship note at the bottom says the opposite; nothing tells them which wins.

**3. Spec `:1021` — a claim this PR's own edit made false, 16 lines below the edit that falsified it.** *(correctness bug — stale prose)*

`docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1018-1021`:

> When `attributableSpoken` falls under `MIN_SPOKEN_FOR_VERDICT` the share is `null` and no verdict is given — the honest answer for a book whose ids have drifted wholesale, **and the orphan banner is the surface that can act on it.**

At `:988-1005` this PR writes that the cache-sourced tier is **not built**, that `collectOrphanedCharacterFallbacks` keeps its single rendered-segments source, and — restating R-8C1's premise as still true — that on this corpus the two populations are near-disjoint, *"the books with the highest orphan shares have zero unresolvable ids in their segments."* A wholesale cache-drifted book is precisely the case where the banner has **zero rows**. The claim was true before this PR only because the tier was in scope; the strike made it false and it was left standing. A reader of §Denominator concludes cache-side drift has a user-facing remedy. It does not — which is half the argument for dropping D13.

---

### 🟡 Minor

**4. Spec `:2398` — the sibling of an edit the PR made.** *(cleanup)* §Testing still justifies asserting `orphanIds` with *"since a count alone cannot drive **the drift surface**."* The PR fixed the identical phrase in the interface block (`:1062`, `// the distinct unresolvable ids, for the drift surface` → `// the distinct unresolvable ids`) and missed this one. The justification now cites a surface that is not built.

**5. Spec `:1489-1500` — R-9M2 left in normative voice while §Review findings declares it struck.** *(cleanup)* *"**The drift share must be over `spokenTotal - unattributedSpeech`**, not over `spokenTotal`"* carries no strike marker, while the addendum this PR adds at `:3097` lists **R-9M2** (*"the drift share's corrected denominator"*) among the dispositions that *"go with it … whose subject is a rule no longer specified."* Reading §D13 re-gated top-to-bottom, the imperative arrives ~40 lines before the outcome that cancels it. Low reader risk, but it is the one unmarked normative "must" for a dropped state left in the spec.

**6. Spec `:169` — "Nothing about D9, D11 or the five-state machine is falsified here."** *(cleanup, borderline)* Unqualified present tense; the machine is four states now. A historical reading — revision 8's rebaseline speaking in its own voice about what F1–F13 falsify — is defensible, and under that reading it is correct as written. Flagged only because the PR's count sweep classified every other `fifth` / `sixth` / `five states` occurrence explicitly and this is the one it left unclassified either way.

---

### ✅ What is solid

- **Over-strike: clean.** Every claimed survivor verified against the post-change file, not the claim. **D9** intact at `:231`, `:915`, `:1058-1061`, `:1353`, `:2214`, `:2920`. **Wave 2 criterion 11's compound is split correctly** (`:2920-2925`) — the D9 half kept in normative voice, only the `drifted`-badging half struck. **`orphanSpoken`/`orphanIds` reference counts went *up*** (24→25 and 8→11 in the spec); nothing removed or deprecated. **R-8C2** re-tensed rather than struck, with the reason stated in-line (`:1867-1869`, `:1880-1887`) — correct, it is independently a `collapsed` hole. **Question 5 (`unanswered`) still open**, with its recommendation preserved.
- **The one claimed "already owed elsewhere" survivor actually exists.** `:2585-2587` retires the Wave 2 drift fixtures on the grounds that the surviving assertion lands on Wave 1's D9 fixture. Checked: `:2391-2397` builds that fixture with the orphan count *comparable to the narrator-id count* specifically so summing moves the share by tens of points and the mutation goes red. The claim holds.
- **Step-numbering integrity: honoured, and judgment call (a) is right.** The sequence keeps `4d / 5 (struck) / 6 / 7`; `attribution-health-io.ts`'s own inline labels are `2 / 2b / 3 / 4 / 7` and both rewritten comments (`:185-189`, `:259-260`) match; the R-9C4 paragraph (`:2002-2006`) was re-argued around the missing step rather than renumbered. No surviving prose anywhere implies a different numbering. Renumbering would have silently falsified two shipped code comments.
- **Count consistency.** `four states` / `one notice variant` consistent at `:1947`, `:1816`, `:2835`, `:2850`, `:2595`; the state union updated in **both** places (`:1748` response type, `:1768` `/api/library`); historical `sixth` correctly *preserved* at `:130` (revision preamble), `:2002`, `:2841` and `:3132` (R-9C4 findings row) — those must not be "fixed" and were not.
- **Register pair.** `check:onbox-register` green; owed total **72** unchanged on both sides (`onbox-acceptance-register.md:303`, live view `owed">72`); E11 still owed with both sub-items; the live-view diff is two lines, wording only — markup, group counts and row IDs untouched.
- **Judgment call (b) verified independently, and the premise holds.** `docs/superpowers/` contains exactly `plans/` and `specs/` — no `archive/` anywhere under it. `git log --diff-filter=R -M --all` finds no plan ever renamed out of `docs/superpowers/plans/`, and nothing renamed into `docs/features/archive/` from there. #2357's acceptance clause names a directory that has never existed; settling the ship note instead is the right call, and flagging it was right.
- **Remaining numbers reconcile against run sheet §5**: 1.80% max (`:211`), 28 of 32 = 87.5% (`:266-272`; 22+4+1+1), fourteen retirements across five books (`:254-256`), no bare `rejected` anywhere (`:283-286`), 16.5–49.2% (four) vs ≤2.5% (nineteen) (`:206-216`; 23−4=19).
- **House gates.** No paired test owed — verified, not assumed: `attribution-health-io.ts:173` declares `AttributionState = 'ok' | 'missing' | 'unmeasurable'`, so `drifted` never existed in runtime code and there is no behaviour to regress; nothing deleted or `.skip`ped. Release-notes skip correctly argued and stated rather than omitted. `Closes #2357` present as plain text outside any code span. Incidental fixes declared in the PR body.

### Verdict

**Request changes.** One blocking finding (#1 — an off-by-one in a figure the drop is argued from, present on the page about to be republished), plus two significant under-strike / stale-prose defects (#2, #3). All six findings are single-sentence edits; none reopens the decision, and the decision itself — including both flagged judgment calls — is correct and well-evidenced. Findings #1–#3 are correctness bugs and re-trigger a review pass once fixed and pushed; #4–#6 are cleanup.

dudarenok-maker:
## PR review — pass 2 (head 7b784e6a, effort high)

Independent re-review, cold start, from worktree `C:\Claude\Projects\wt-d13-drop-decision` at `7b784e6a`. Effort **high**, re-derived from the current commit set (commit 1 is multi-scope `docs,server`). Pass 1's six findings were each verified against the post-fix files and re-derived from primary data rather than from the corrected sentences; the fix commit was then treated as new, unreviewed surface.

**Commands run (actual output):**

- `cd server && npx tsc --noEmit -p tsconfig.json` → **exit 0**
- `npm run check:onbox-register` → `check:onbox-register: OK — docs/testing/onbox-acceptance-register.md and docs/testing/onbox-acceptance-register-live-view.html agree.` (**exit 0**)

---

**1. 🔴 "15 of the 16 non-fixture books" → 14 — ✅ VERIFIED RESOLVED.** Re-derived from the run sheet's own §1 table (`:51-75`), not from the corrected prose:

- Coalfall family = **7** rows: `The Coalfall Commission` en 62, `El Encargo de Coalfall` es 32, `La Commande de Coalfall` fr 23, `Заказ Коалфолла` ru 18, `煤落的委托` zh 3, `Der Auftrag von Coalfall` de 2, `コールフォールの依頼` ja **0** (orphan column parsed per-row; ja is genuinely 0).
- 23 − 7 = **16** non-fixture; of those exactly **two** are non-zero (`Ночной дозор` 32, `Playing with Fire` 6, both labelled `real` at §5a `:211`/`:213`) → **14** at zero. Correct.
- Whole-corpus: 8 non-zero of 23 → **15** at zero.

Corrected at spec `:1539`, `:2806`, run sheet `:219`, live view `:1364`. **The two correct whole-corpus `15`s were left alone** — run sheet `:215` ("The remaining 15 books are at exactly **zero**") and `:187` (the collapsed provenance block, "then zero on the remaining 15 books"); both re-derived and both right. The disambiguating clause at `:219-222` is accurate in all three of its assertions. Tree-wide sweep for other copies of the figure and for prose restatements ("only one real book", "every other real book at zero", "the remaining N books", "non-fixture books") found no further site.

**2. 🟠 Wave 1 plan carried live `drifted` Wave 2 scope — ✅ VERIFIED RESOLVED.** `:820-823` now carries the struck slot, `:830-832` re-states step 6 alone as thresholdless and step 5 as *"struck, not deferred: it never gets a threshold"*, and `:948-952` splits `COLLAPSE_SHARE_THRESHOLD` from `DRIFT_SHARE_THRESHOLD` and drops the false *"all four floors"*. **Copy count established as asked:** exactly **two** copies of the state sequence exist in the whole repo — spec `:1990-2001` and plan `:815-826` — and they are now byte-comparable in the struck step's wording. `server/src/store/attribution-health-io.ts:185-189` and `:259-260` carry the matching prose form. No third copy in `docs/features/`, `src/`, `server/src/`, or any test fixture.

**3. 🟠 Spec `:1021` "the orphan banner is the surface that can act on it" — ✅ VERIFIED RESOLVED.** Now reads *"no surface exists today that can act on it (R-8C1)"*, which is what `:986-1005` actually establishes. The adjacent under-reporting sentence (`:1022-1027`) was correctly reworded off the state name onto the condition (*"a collapsed share and ids that have drifted out from under it"*).

**4. 🟡 Spec `:2398` "the drift surface" — ✅ VERIFIED RESOLVED** (`:2397-2400`, re-justified on what `orphanIds` is actually for). No `drift surface` occurrence survives anywhere in the tree, case-insensitively.

**5. 🟡 R-9M2 `:1489-1500` unmarked — ✅ VERIFIED RESOLVED.** Now `**~~One further consequence the fold did not draw.~~ — STRUCK with D13** (#2357; R-9M2)`, with the body re-tensed to past throughout and the reason for keeping it stated. Style matches the three existing strike markers (`:1698`, `:1817`, `:2584`); semantics read as *dropped*, and `:3095-3098` still names R-9M2 in the same disposition. Checked the whole PR for the load-bearing distinction: `deferred` appears nowhere against D13 except in the two server comments that explicitly say **"dropped, not deferred"**.

**6. 🟡 Spec `:169` "the five-state machine", deliberately left — see assessment below.**

**Classification sweep re-run, case-insensitive**, over `drifted` / `alsoCollapsed` / `attributionVerdictKey` / `unacknowledgedOrphanSpoken` / `DRIFT_SHARE_THRESHOLD` / `MIN_ORPHAN_FOR_VERDICT` / `drift surface` / `alsoDrifted`: 74 hits in the spec, 4 in the plan, 2 in `server/src` — every one classifies as historical evidence, a struck marker, or the decision record. The two `server/src` hits are `attribution-health-io.ts:187`/`:260`, both correct. Every other `drifted` in `server/src` is unrelated prose (acoustic drift, id-spelling drift, tone drift).

**Register pair: unmoved, as claimed.** Owed total **72** on both sides, E11 present at `onbox-acceptance-register.md:3158` with both sub-items intact, whole-branch register diff is one wording line, whole-branch live-view diff is two lines (E11 summary risk text + the `15`→`14`). No row, group count or ID moved.

---

### 🟠 Significant — new

**N1. The Coalfall family is stated as *five* language editions in two places — one of them the published live view — while the clause this fix commit just added says *seven*.** *(correctness bug — the premise the corrected figure rests on)*

`docs/testing/attribution-collapse-visibility-onbox-acceptance.md:126-127` and `docs/testing/onbox-acceptance-register-live-view.html:1361`.

> `orphanSpoken` non-zero on several books … concentrated in the *Coalfall Commission* family (**0/23/32/2/62/18 across its five language editions**) — run sheet `:126-127`

> `orphanSpoken` non-zero on several books, concentrated in the *Coalfall Commission* family (**2–62 across its five language editions**) — live view `:1361`

Against `:220`, added by `7b784e6a`, 93 lines below the first: *"this figure excludes **all seven** fixture editions instead"*.

Both are wrong under every reading, and the run sheet's own enumeration disproves itself twice over:

- `0/23/32/2/62/18` is **six** values under a label that says five, and it **omits `煤落的委托` (zh) entirely**, whose orphan count is 3 (§1 table `:58`). The family is seven; the enumeration lists six of them.
- The live view's `2–62` range spans `de` 2, `zh` 3, `ru` 18, `fr` 23, `es` 32, `en` 62 — again six, not five.

This is not a cosmetic count. It is the denominator of the figure pass 1 blocked on. A reader who takes the document at its word computes 23 − 5 = **18** non-fixture books and concludes that `:219`'s *"14 of its 16 non-fixture books"* is itself wrong — the two sentences cannot both be true, and the document gives no way to tell which one to believe. The corrected `14 of 16` is only verifiable if the family is seven, and the document states five in the very section (§2) a reader passes through on the way to §5a. It is also the exact defect class pass 1 raised: a fixture-family miscount, applied inconsistently between the two halves of a subtraction. Pass 1's fix agent swept the numerator (`15`→`14`) and did not sweep the count that produced the error.

Both sites are pre-existing on `main` — which under CLAUDE.md's incidental-findings rule is a void deferral reason, and specifically so here: the live-view instance is on the page this PR is about to republish to the artifact URL, alongside the `14` it now contradicts.

Concrete fix: `five` → `seven` at both sites, and the run sheet's enumeration needs the missing `3` (`0/3/23/32/2/62/18`, or state it as a range). Nothing about the decision moves.

---

### 🟡 Minor — new

**N2. Wave 1 plan `:16` still calls `unanswered` "a sixth state", in a row whose own status text is "Still open".** *(cleanup — the seventh copy of a renumbering the PR swept six times)*

`docs/superpowers/plans/2026-08-13-attribution-collapse-visibility-wave1.md:16`:

> `| 5 | **Does `unanswered` become a sixth state?** (spec R-9C4) | **Still open.** Wave 2 only. Does not block this plan … |`

The PR body states this renumbering was handled in six forward-looking passages, all of them in the spec, where the canonical form is now `:2843-2844` — *"Does `unanswered` become a **fifth** state … (added in the round-8 fold as a *sixth*; it is a fifth since …)"* — and `:2004-2005` re-tenses it in line. The plan's copy is unswept, and is the one document a Wave 2 implementer opens first.

**The counter-argument is real and I want it on the record rather than buried:** the row sits inside the dated `> ## ✅ APPROVED — revision 8, 2026-08-13. Dispatched.` blockquote, which is the same historical-snapshot category the PR correctly preserved at spec `:130`, and the plan's Ship notes (`:966-971`) do update the reader. What makes it weaker than `:130` is that `:130` records a completed action (*"Also folded: a sixth state"*) while this row asserts a **live status** — "Still open" — with a stale ordinal in the question it says is still open. Same blockquote also now carries a second staleness of the same kind: *"three are answered and two remain open"* and question 2's `**Answered 2026-08-13: yes**`, where spec `:2783` records question 2 as **SPENT**, superseded by #2357.

Cleanup, not a correctness bug — nothing gets built wrong from it, and the row points at spec R-9C4 where the count is right. Raised because it is the same "second copy" class as pass 1's finding 2, in the same file, and because the PR body's six-passage claim reads as complete when it is not.

---

### Assessment of the `:169` judgment call

**The ruling stands — I would not block on it, and I would not change it — but the reason that carries it is not the one given.**

`docs/superpowers/specs/…-design.md:169`: *"Nothing about D9, D11 or the five-state machine is falsified here."*

What actually saves it is the word **"here"**. The sentence is the closing scope statement of `### The revision-7 claims this rebaseline falsifies`, and it is a claim about the *punctuation strand's blast radius*, evaluated against the F1–F13 table immediately above it. Read that way it is true as written and re-tensing it would misstate what revision 8 concluded.

The "it's historical, same category as `:130` and the R-9C4 findings row" framing is weaker than the PR presents it. `:130` is inside a `>` revision-history blockquote and `:3135` is inside a findings table; `:169` is plain body prose in a live section, using the definite article for a machine that now has four states. The category is not the same — the *scoping* is what does the work.

Worth noting against the PR's own house style: the document solves the identical problem in line four hundred lines further down, at `:2004-2005` — *"the SIXTH when revision 8's fold added it, the FIFTH since `drifted` was dropped"*. Four words would have applied that treatment here. That the PR chose preservation in one place and in-line re-tensing in the other is an asymmetry, not an error, and the outcome at `:169` is defensible either way. No change required.

---

### Verdict

**Request changes** — one Significant correctness bug (N1), plus one Minor cleanup (N2).

All six of pass 1's findings are verified resolved, each against the post-fix file and, for finding 1, re-derived from the §1 table rather than from the corrected sentence. The fix commit introduced no new inconsistency, no broken cross-reference and no unsupported claim; its two new struck markers match the existing convention in both style and semantics, and nothing anywhere in the PR reads as *deferred* where it must read as *dropped*. The register/live-view pair is unmoved and agrees.

N1 is the same defect pass 1 caught, one level up — the family count the corrected figure is derived from — and it is on the page about to be republished. It is a correctness bug and re-triggers a review once fixed and pushed. N2 is cleanup and does not.

dudarenok-maker:
## PR review — pass 3 (head 24c0f3db, effort high)

Independent re-review, cold start, from worktree `C:\Claude\Projects\wt-d13-drop-decision` at `24c0f3db`. Effort **high** (re-derived: commit 1 is multi-scope `docs,server`). Both prior passes read first. Every quantitative claim in the decision record was re-derived from the run sheet's §1 corpus table and §5 sub-sections rather than from the sentences asserting them — and, where the box allowed it, from **the real workspace on disk** rather than from the document at all.

**Commands run (actual output):**

- `cd server && npx tsc --noEmit -p tsconfig.json` → **exit 0**
- `npm run check:onbox-register` → `check:onbox-register: OK — docs/testing/onbox-acceptance-register.md and docs/testing/onbox-acceptance-register-live-view.html agree.` (**exit 0**)

---

**Pass 2, N1 (🟠 Coalfall family stated as five editions) — ⚠️ PARTIALLY RESOLVED.** The two sites pass 2 named are fixed and correct: run sheet `:126-127` now reads `0/2/3/18/23/32/62 across its seven language editions` (re-derived per-row from the §1 table: ja 0, de 2, zh 3, ru 18, fr 23, es 32, en 62 — seven values, all correct, and none of them moved between the partial and complete runs, so the enumeration is valid inside the partial-run block it sits in). Live view `:1361` now reads `0–62 across its seven language editions`. **A third copy was not swept — see N3 below.**

**Pass 2, N2 (🟡 plan approval table) — ✅ VERIFIED RESOLVED.** Row 5 reads *"Does `unanswered` become a **fifth** state?"* with `(Was "sixth" — Wave 2 ships four states, not five, now that #2357 dropped D13)`. Row 2 keeps the dated `Answered 2026-08-13: yes` and appends `**SPENT — superseded 2026-08-14 by #2357**`. **Checked against the spec's own treatment of the same question, as asked:** spec `:2783` carries `**SPENT — superseded 2026-08-14 by [#2357].**` on question 2 — same marker, same date, same issue, same "the answer was conditional and the condition fired" reasoning. They match. The plan preamble's *"three are answered and two remain open"* is **not** stale against the spec: `:2779-2780` says *"Three of the five are decided; questions 3 and 5 stay open"*. The two documents agree.

**Pass 1's six findings — ✅ NONE REGRESSED.** Established mechanically rather than by re-reading: `24c0f3db` touches exactly three line ranges — plan `:11-16`, run sheet `:123-127`, live view `:1361`. None of pass 1's six sites falls in any of them, and the spec is not touched by that commit at all. Spot-verified the two that carry figures anyway: `:1021` still reads *"no surface exists today that can act on it (R-8C1)"*; plan `:820-823` still carries the struck slot with the non-renumbering rationale; the `14` still stands at spec `:1539`/`:2806`, run sheet `:219`, live view `:1364`.

**Classification sweep re-run, case-insensitively**, over `drifted` / `alsoCollapsed` / `attributionVerdictKey` / `unacknowledgedOrphanSpoken` / `DRIFT_SHARE_THRESHOLD` / `MIN_ORPHAN_FOR_VERDICT` / `drift surface` / `alsoDrifted` / `sixth`: spec 46/10/10/6/6/5/0/1/4, plan 3/0/0/0/2/0/0/0/1. Every hit classifies. All five `sixth` occurrences are historical or re-tensed in line (spec `:130` revision preamble, `:2004` *"the SIXTH when revision 8's fold added it"*, `:2844` *"added in the round-8 fold as a sixth; it is a fifth since"*, `:3084` R-8C3 findings row; plan `:16` *"(Was 'sixth' …)"*). No `drift surface` survives in either document — the remaining tree-wide hits are unrelated (engine drift, cast drift, marketing). The only `drifted` tokens in `server/src`/`src` are acoustic/id/tone drift. The two server comment edits are correct and consistent with `AttributionState = 'ok' | 'missing' | 'unmeasurable'`; comment-only, so no paired test is owed.

**Register/live-view pair: unmoved.** Owed total **72** on both sides (`onbox-acceptance-register.md:303`, live view `owed">72<`). E11 still owed with both sub-items. No row, group count or ID moved.

---

### Disk-state claims — verified against the real workspace, not the document

`C:\AudiobookWorkspace` is present on this box, so three claims that both prior passes could only take on trust were checked directly:

- **"Fourteen retirements exist across five books"** — ✅ **exactly right.** Five `cast-id-history.json` files under `books/` (a sixth is under `_backup/`, correctly not counted). `supersededBy` entries: `Заказ Коалфолла` 2, `Playing with Fire` 2, `Everblaze` 1, `Ночной дозор (C2C3 run 2)` 5, `Ночной дозор (Tetralogy)` 4 = **14**.
- **"No book in the corpus has a bare `rejected` entry — the key is absent from all five `cast-id-history.json` files"** — ✅ **confirmed.** No `rejected` key in any of the five.
- **§5b's natural experiment** — ✅ **confirmed.** `Заказ Коалфолла`'s history literally contains `"coalfall": "coalfall-dragon"`. And none of `Ночной дозор (Tetralogy)`'s four retired ids (`pavel`, `dark-programmer-2/3/4`) appears among its five orphan ids.

### Re-derivation of the remaining figures

Every row of §5a's share table recomputed from §1 as `orphan / (spoken − unattr − split)`, using the **complete**-run values for the three books that moved:

| Book | derived | doc | ✓ |
|---|---|---|---|
| The Coalfall Commission | 62/(128−2−0)=62/126=49.206% | 62/126, 49.21% | ✅ |
| El Encargo de Coalfall | 32/(127−4−0)=32/123=26.016% | 32/123, 26.02% | ✅ |
| La Commande de Coalfall | 23/(128−2−0)=23/126=18.254% | 23/126, 18.25% | ✅ |
| Заказ Коалфолла | 18/(115−4−2)=18/109=16.514% | 18/109, 16.51% | ✅ |
| 煤落的委托 | 3/(128−6−0)=3/122=2.459% | 3/122, 2.46% | ✅ |
| Ночной дозор (Tetralogy) | 32/(2122−9−337)=32/1776=1.8018% | 32/1776, **1.80%** | ✅ |
| Der Auftrag von Coalfall | 2/(128−1−5)=2/122=1.639% | 2/122, 1.64% | ✅ |
| Playing with Fire | 6/(2238−210−275)=6/1753=0.3423% | 6/1753, **0.34%** | ✅ |

- **`16.5–49.2%` vs `≤2.5%`, "a 6.7× step"** — high group is exactly those four (16.51 / 18.25 / 26.02 / 49.21); next below is 2.46%; 16.51/2.46 = **6.71×**. ✅ And "four books vs nineteen" = 23−4. ✅
- **"The remaining 15 books are at exactly zero"** (whole corpus) — 8 non-zero of 23 → **15**. ✅ Both surviving whole-corpus `15`s (`:187`, `:215`) are correct and were correctly left alone.
- **"14 of its 16 non-fixture books"** — 23 − 7 = 16; two non-fixture books are non-zero → **14**. ✅
- **`MIN_ORPHAN_FOR_VERDICT = 20`, "`Ночной дозор` is the only non-fixture book that clears it"** — non-fixture books at ≥20 orphans: Ночной дозор 32 only (Playing with Fire is 6). ✅ (Fixture editions clearing it: en 62, es 32, fr 23 — `Заказ Коалфолла`'s 18 does not, which is consistent with everything the document says about it.)
- **"50 of the top book's 62 orphan spans"** — `coalfall` 28 + `oduvan` 22 = **50** of 62. ✅
- **The 2026-08-13 partial-run block re-derived in its own right** (`:183-193`): Ночной дозор 29/(1928−12−250)=29/1666=1.74% → "1.7%" ✅; every other figure and the "then zero on the remaining 15 books" ✅. **`Ночной дозор`'s two orphan counts are each correct in their own context** — 29 throughout the partial-run material (§1 `:44`, §5 `:183-193`, register `:3183`, live view `:1361`) and 32 throughout the complete run (§1 `:98`, §5a `:211`, §5c `:265`). No site mixes them.
- **`24c0f3db` did not re-tense the partial-run prose** or imply it measured something it did not: its run-sheet hunk changes only the enumeration and the word `five`→`seven`, and every Coalfall orphan count is identical across both runs (§1 names the only three books that moved), so the corrected enumeration is true of the partial run it describes. ✅

**28 of 32 (87.5%) does not close — see N4.**

---

### 🟠 Significant — new

**N3. The Coalfall family is *still* stated as "five language editions" — in the acceptance register itself, the tracked twin of the page that now says seven.** *(correctness bug)*

`docs/testing/onbox-acceptance-register.md:3179-3180`:

> `orphanSpoken` non-zero on several books, concentrated in the *Coalfall Commission* family (**2–62 across its five language editions**)

Against its own published pair, `docs/testing/onbox-acceptance-register-live-view.html:1361`, as rewritten by `24c0f3db`:

> concentrated in the *Coalfall Commission* family (**0–62 across its seven language editions**)

These two paragraphs are the **same E11 "What was observed" block** — the register `.md` and the live view are the maintained pair, and the whole point of the pair is that they agree. After this fix they disagree on **two tokens in one clause**: the range low bound (`2–62` vs `0–62`) and the edition count (`five` vs `seven`).

This is pass 2's N1 with one site left standing. Pass 2 named two locations; the fix commit swept exactly those two and did not sweep the class. `git grep` confirms `:3179` is now the **only** remaining "five language editions" statement about this family anywhere in the tree (the `docs/wiki/Release-Notes-v1.14.0.md:137` and `docs/features/252-…:65` "five languages" hits are the samples pack, correctly five of seven; live view `:1221` / register `:2827` "D1's five languages" is an unrelated row).

What a reader concretely gets wrong is exactly what pass 2 described, and it is now *worse* because the two halves of the pair contradict each other: a reader of the register computes 23 − 5 = **18** non-fixture books, and the corrected *"14 of its 16 non-fixture books"* four screens away reads as arithmetic that does not close. `npm run check:onbox-register` cannot catch this — it cross-checks owed totals, group counts and row IDs, not prose (CLAUDE.md says so explicitly).

Fix: `2–62 across its five language editions` → `0–62 across its seven language editions` at `:3179-3180`. Two tokens. Nothing about the decision moves.

**N4. §5c's per-id table sums to 33 over a stated total of 32, and the headline 87.5% cannot be checked from it.** *(correctness bug — the second of the four pillars of the drop argument, and it is on the published page)*

`docs/testing/attribution-collapse-visibility-onbox-acceptance.md:265-276`, `docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1544-1547`, `docs/testing/onbox-acceptance-register-live-view.html:1364`.

Run sheet `:265`: *"**Splitting its 32 spans per id**, through the production resolver:"* — then a table whose `spans` column reads `driver` 22, `boris-igoryevich` 5, `woman-in-taxi` 4, `vampire-boy` 1, `vampire-girl` 1. **22 + 5 + 4 + 1 + 1 = 33.** The book's `orphanSpoken` is **32** (§1 `:98`, §5a `:211`). A table introduced as a split of 32 does not split 32.

The spec states it more starkly still, in two adjacent sentences of the permanent decision record (`:1544-1547`):

> `Ночной дозор` has **32** orphan spans, of which **28** (87.5%) belong to `driver`, `woman-in-taxi`, `vampire-boy` and `vampire-girl` … Only `boris-igoryevich` (**5** spans, 0.28%) is linkable.

28 + 5 = 33 against a stated 32, with no reconciliation offered.

The mechanism that explains the gap is in the repo but is never connected to it: `server/src/store/attribution-health.ts:230-249` (R-9C5) — **`orphanSpoken` counts SPANS and increments at most once per span, while `orphanIds` collects every distinct unresolvable id on that span**, so one split span carrying two orphan ids yields 33 id-span pairs over 32 distinct spans. §1 `:41-45` documents that this book is precisely the corpus's one multiplicity case (*"one split span in that book carried two distinct bogus ids"*, the `30 → 29` correction). The document contains the explanation 220 lines above the table and never uses it.

**This is not merely presentational, because it decides whether 87.5% is true.** Exactly two resolutions are possible, and the document does not say which holds:

- the doubled span pairs `boris-igoryevich` with an unlinkable id → distinct unlinkable spans = 28, 28 + 5 − 1 = 32 ✓, and **28 of 32 = 87.5% is correct**;
- the doubled span pairs two unlinkable ids → distinct unlinkable spans = **27**, 27 + 5 = 32 ✓, and the correct figure is **27 of 32 = 84.4%**.

Both reconcile to 32; they disagree on the published number. As written the record asserts *both* a partition of 32 and a numerator of 28, and those cannot both hold. A reviewer who adds the column — which is the only check available, since the JSON report is git-ignored and unattached — gets 33 and has no way to resolve it.

The decision does **not** move either way (84.4% and 87.5% both say "overwhelmingly unlinkable", and the `gate with no exit` argument is untouched), so this is a record-accuracy defect, not a re-argument. But it is one of the four numbers `#2357` is argued from, it is in the permanent spec, and it is on the page about to be republished to the artifact URL.

Fix, and it needs no new measurement if the author can recover which pair shares the span from the run's console output: state that the column is **per id** (so it sums to 33 across 32 distinct spans, one span carrying two orphan ids per R-9C5, exactly as §1 `:41-45` describes), and give the unlinkable figure as a distinct-span count. If which-pair cannot be recovered, the honest form is *"four of the five orphan ids — 28 of the 33 id-span pairs, at least 27 of the 32 distinct spans — are unlinkable"*. Pre-existing from #2356 and untouched by this diff, which under CLAUDE.md's incidental-findings rule is a void deferral reason, and specifically so here for the same reason pass 1 gave: it is about to be republished.

---

### 🟡 Minor — new

**N5. "14 of 16 … at exactly zero" counts two books that were never analysed, whose `0` is a `?? 0` default rather than a measurement.** *(cleanup — not decision-moving, and I would not hold the PR for it)*

`docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1539` and `:2806`, `docs/testing/attribution-collapse-visibility-onbox-acceptance.md:219`, `docs/testing/onbox-acceptance-register-live-view.html:1364`.

The 16-book non-fixture denominator includes three `Ночной дозор` C2/C3 throwaway duplicates (§1 `:72-74`), two of which the run sheet itself labels `ok (not analysed)` and describes as *"genuinely un-analysed C2/C3 throwaways"* (`:87-89`). For those two there is no measurement at all: `scripts/measure-attribution.mjs`'s `buildRow` writes `orphanSpoken: m?.orphanSpoken ?? 0` when `m` is null, so their `orphan 0` is the null coalesced to zero, and their `share` correctly prints `—`. A third (`Ночной дозор (C2 throwaway)`) is analysed but has zero attributable spans.

So two of the fourteen "at exactly zero" are null observations counted as clean readings — the "control group labelled clean but never measured" shape. The published wording at live view `:1364` and spec `:2806` compounds it slightly by calling the denominator *"16 **real** books"*, where three of the sixteen are throwaway test duplicates of one book; the run sheet's own `non-fixture` is the accurate word.

Nothing about the decision moves — the maximum among genuinely-measured non-fixture books is still `Ночной дозор` at 1.80%, and the measured-with-attributable-spans population is 13, of which 11 are at zero and 2 are not. Raised only because the figure is on a public page and the run sheet already holds the disclosure 130 lines above it; a half-sentence (*"two of which are un-analysed throwaways"*) discharges it.

---

### ✅ What is solid

- **Every other figure in the decision record re-derives exactly**, including all eight rows of §5a's share table computed from §1 rather than read off, and the three disk-state claims verified against the real `C:\AudiobookWorkspace` rather than trusted.
- **`24c0f3db` introduced no new inconsistency and broke no cross-reference.** It is five changed lines; the run-sheet hunk stays inside the superseded partial-run block without re-tensing it, and the enumeration it writes is true of both runs.
- **The plan's row-2 SPENT marker matches the spec's** on the same question — same marker, date, issue and rationale.
- **Register pair unmoved**: 72 owed both sides, E11 owed with both sub-items, no row/group/ID movement, `check:onbox-register` green.
- **House gates.** No paired test owed (verified: `AttributionState` never contained `drifted`; both server edits are comment-only); nothing deleted or `.skip`ped; release-notes skip argued rather than silently omitted; `Closes #2357` / `Refs #1984` present as plain text; incidental fixes declared in the PR body.

---

### Verdict

**Request changes.** Two Significant correctness bugs (**N3**, **N4**) and one Minor cleanup (**N5**).

Both correctness findings are the same shape this review has hit every round — a locally correct fix that left its own input, or its own third copy, unexamined. N3 is pass 2's N1 with one site unswept, and it is a two-token edit. N4 is the one figure in the decision record whose arithmetic does not close, and unlike the previous rounds' findings it cannot be settled from the document alone. **Neither moves the decision**, which remains correct and well-evidenced; both flagged judgment calls in the PR body still hold.

**Safe to merge: not yet — N3 first.** N3 leaves the acceptance register and its published twin asserting different facts about the same corpus, which is precisely the invariant that pair exists to hold.

**Safe to publish: not yet.** The live view is correct on N3 but carries N4's `28 of 32 orphan spans (87.5%)` and N5's `14 of 16 real books`. N4 in particular should be settled before the page goes to the artifact URL — a figure that cannot be reconciled with its own supporting table is worse on a public page than in a tracked doc, and republishing is the step that makes it permanent.

N3 and N4 are correctness bugs and would ordinarily re-trigger a review; this is pass 3 of a 3-pass cap, so that call belongs to the dispatching session. Both fixes are text-only and self-contained; N5 can be folded into the same edit or dropped.

dudarenok-maker:
## PR review — pass 4 (head 99d22735, effort high)

Independent re-review, cold start, from worktree `C:\Claude\Projects\wt-d13-drop-decision` at `99d22735`. Effort **high**. Passes 1–3 read first. This pass was dispatched past the documented 3-pass cap because every prior round found a defect **one level above** the previous round's input, so the target here was the new, unreviewed surface `99d22735` created — above all the **84.4%**, which is a figure derived by a single agent that replaced one that had survived three reviews.

**Method note, because it is the point of this pass:** the 84.4% was **not** re-checked by re-running the prior probe. It was re-derived by a *different* route — black-box **input perturbation** of the production `computeAttributionMeasurement`, never replicating its span loop. Adding an orphan id to the `cast` **input** makes it resolvable, so

> `f(S)` := `orphanSpoken` with the ids in `S` appended to `cast` = # spans carrying ≥1 orphan id **not** in `S`

and the whole span partition falls out of the 2⁵ subset lattice by Möbius inversion. Every perturbation was guarded: the resulting `orphanIds` had to equal `baseline \ S` exactly, so a normalised-key collision or a newly-live history target would have surfaced as `CONFOUNDED` rather than silently shifting a count. None did.

**Commands run (actual output):**

- `cd server && npx tsc --noEmit -p tsconfig.json` → **exit 0**
- `npm run check:onbox-register` → `check:onbox-register: OK — docs/testing/onbox-acceptance-register.md and docs/testing/onbox-acceptance-register-live-view.html agree.` (**exit 0**)

---

**Pass 3, N3 (🟠 third "five language editions" copy at register `:3179`) — ✅ VERIFIED RESOLVED.** `onbox-acceptance-register.md:3179-3180` now reads `0–62 across its seven language editions`, byte-identical in substance to its published twin at live view `:1361`. Verified against measurement, not against the corrected sentence: the seven Coalfall editions' `orphanSpoken` are ja 0, de 2, zh 3, ru 18, fr 23, es 32, en 62 — seven values, range 0–62. Tree-wide sweep: the only surviving "five editions" hits are `RELEASE_NOTES.md:200` and `docs/wiki/Release-Notes-v1.14.0.md:137`, both the 5-language `samples/` demo pack in a shipped v1.14.0 note (correct as written; `samples/` has since grown a ja and zh directory, but the notes are historical).

**Pass 3, N4 (🟠 §5c summed to 33 under a "32 spans" header; 87.5% unverifiable) — ✅ VERIFIED RESOLVED, AND THE NEW FIGURE INDEPENDENTLY REPRODUCED.** See the derivation below. **84.4% is correct.** All five corrected sites agree and no sixth survives: `git grep -E "87\.5|28 of |28 orphan|of 32"` over the whole tree returns nothing related (only `attribution-sentence-progress-design.md:24`'s `2:28 of ~` and `bench-tts.py:340`'s "batch of 32 items"). The five sites are spec `:1546`, spec `:2815`, run sheet `:291`, run sheet `:313`, live view `:1364` — each carries `84.4%`, and the two that state the count state `27 of 32`.

The multiplicity explanation is **accurate against the code it cites**. `attribution-health.ts:230-249` is exactly the R-9C5 block: line 230 opens the comment, 239 `let sawOrphan = false`, 240–245 the `for (const id of idsForSpan)` loop that `orphanIds.add(id)`s every unresolvable id, 246–249 the `if (sawOrphan) result.orphanSpoken += 1`. The line range is exact. So is the mechanism claim: multiplicity is reachable **only** through the `idsForSpan.size > 1` branch (the single-id branch adds exactly one id), which is why it requires a split span — matching §1's "one split span in that book carried two distinct bogus ids".

**Pass 3, N5 (🟡 "16 real books" overclaimed) — ✅ RESOLVED as stated, but the class is only half-swept — see N6.** The wording moved from "16 **real** books" to "16 **non-fixture** books" at spec `:2810` and live view `:1364` (no "16 real" survives anywhere), and a qualifying clause was added at four sites. **The prior agent's correction of the mechanism in its own brief is itself correct — verified, not taken on trust.** `resolveAttributionState` reaches step 7 with a live measurement for these rows (`state: 'ok'`, `reason: 'not analysed'` because `hasCacheFile` is false), so `buildRow`'s `m?.orphanSpoken ?? 0` never fires — `m` is non-null. Measured directly: both never-analysed books return `sentences: []`, `spokenTotal: 220`, `unattributedSpeech: 220`, `orphanSpoken: 0`. The real cause is exactly as the agent restated it, and the docs describe it accurately and mechanism-agnostically ("whose zero reflects no analysis having run rather than a measured absence of orphans").

**Pass 1's and pass 2's findings — none regressed.** `99d22735` touches four files in six hunks, none of which overlaps a pass-1/2 site; spot-checked anyway: spec `:1021` still reads *"no surface exists today that can act on it (R-8C1)"*, plan `:820-823` still carries the struck slot, plan `:16` still carries the re-tensed *"fifth"*, register/live view still `0–62 … seven`. `Owner confirmation owed` is now absent tree-wide (it survived on `main` at spec `:1575` and inside the live-view flag paragraph; both gone).

**Standing token sweep re-run, case-insensitive**, over `drifted` / `alsoCollapsed` / `attributionVerdictKey` / `unacknowledgedOrphanSpoken` / `DRIFT_SHARE_THRESHOLD` / `MIN_ORPHAN_FOR_VERDICT` / `drift surface` / `alsoDrifted` / `sixth`. Every hit classifies. Spot-checked the ones a mechanical count would let through: spec `:3092` `alsoDrifted` is the R-8C3 findings row (historical); plan `:663` `wholly-drifted` is generic prose about D9's numerator, not the state; plan `:820`/`:952` are struck, `:968` is the ship note; all five `sixth` hits are historical or re-tensed in line (spec `:130`, `:2011`, `:2852`, `:3143`; plan `:16`). No `drift surface` survives in either document — the four remaining tree hits (`layout.tsx:978`, `drift-report.tsx:408`/`:648`, `generation.test.tsx:1305`) are the unrelated audio drift-report surface.

**Register/live-view pair: unmoved.** **72 owed** on both sides (`onbox-acceptance-register.md:303`, live view `owed">72`). E11 still owed with the same two sub-items — (2) dash-stripped re-run invariance, (3) re-analyse one book post-D18 — on both sides. Whole-branch register diff is two prose hunks; whole-branch live-view diff is three lines (E11 summary risk text, the `five`→`seven`, the flag paragraph). No row, group count or ID moved. `check:onbox-register` green.

---

### The 84.4% — independently derived, from the workspace, by a different route

Run read-only against `C:\AudiobookWorkspace` (23 books) via the primary checkout's built `server/dist`. Baseline for `Ночной дозор (Tetralogy)` reproduces **both published anchors before answering anything new**: `orphanSpoken = 32`, per-id span counts `driver` 22 / `boris-igoryevich` 5 / `woman-in-taxi` 4 / `vampire-boy` 1 / `vampire-girl` 1, **sum 33**. Every other §1 column for that book also reproduces (`spoken` 2122, `unattr` 9, `split` 337), as do `Everblaze` 4266 and `Keeper of the Lost Cities` 3796.

The full 2⁵ lattice, guard clean on all 32 perturbations:

```
f({})                                    = 32   <- total orphan spans
f({boris-igoryevich})                    = 27   <- spans carrying >=1 UNLINKABLE id
f({driver})                              = 11
f({driver, woman-in-taxi})               =  7
f({boris, driver, woman-in-taxi})        =  2
f(all five)                              =  0   <- sanity: everything resolves
```

Möbius inversion over the lattice gives the **exact partition of the 32 spans**:

| orphan-id set on the span | spans |
|---|---|
| `{driver}` | 21 |
| `{woman-in-taxi}` | 3 |
| `{driver + woman-in-taxi}` | **1** |
| `{boris-igoryevich}` | 5 |
| `{vampire-boy}` | 1 |
| `{vampire-girl}` | 1 |
| **total** | **32** |

- Per-id sum = 21+1 + 3+1 + 5 + 1 + 1 = **33**, over **32** distinct spans → exactly one doubled span. ✅
- The doubled span carries **`driver` + `woman-in-taxi`**, both unlinkable. **`boris-igoryevich`'s 5 spans are all standalone** — confirmed independently by the single decisive measurement `f({boris-igoryevich}) = 27`, which is the count of spans carrying ≥1 unlinkable id and would have read **28** under the other reading. ✅
- **27 / 32 = 84.375% → 84.4%.** ✅ Not 87.5%.

The PR body's breakdown ("21 `driver`-only + 3 `woman-in-taxi`-only + 1 shared + 5 `boris-igoryevich` + 1 + 1 = 32") is correct in every term. The correction stands on a second, independent derivation.

---

### Multiplicity sweep — every book, not just the one §1 asserts

§1 records `Ночной дозор` as the corpus's one multiplicity case, and nobody had swept for it. Swept all 23: for every book with ≥1 orphan id, the per-id span tally was computed by the same perturbation route and compared against that book's own `orphanSpoken`.

| book | orphanSpoken | Σ per-id | excess |
|---|---|---|---|
| The Coalfall Commission | 62 | 62 | 0 |
| El Encargo de Coalfall | 32 | 32 | 0 |
| La Commande de Coalfall | 23 | 23 | 0 |
| Заказ Коалфолла | 18 | 18 | 0 |
| 煤落的委托 | 3 | 3 | 0 |
| Der Auftrag von Coalfall | 2 | 2 | 0 |
| Playing with Fire | 6 | 6 | 0 |
| **Ночной дозор (Tetralogy)** | **32** | **33** | **1** |
| the other 15 books | 0 | — | — |

**The assertion holds: `Ночной дозор` is the only multiplicity case in the corpus.** No other book's per-id counts can shift the way this one's did, so the fixture-family orphan aggregates (62/32/23/18/3/2/0) and every per-id figure derived from them are unaffected. `The Coalfall Commission`'s `coalfall` 28 + `oduvan` 22 = 50 of 62 is a clean sum over disjoint spans. ✅

---

### Fixture-vs-real classification — all 23, explicitly

Classified from the workspace directory tree, not from §1's table (fixture ⇔ author `Castwright`; verified that `Castwright/Standalones/` contains exactly the seven Coalfall editions and nothing else).

**Fixture family (7):** `The Coalfall Commission` en 62 · `El Encargo de Coalfall` es 32 · `La Commande de Coalfall` fr 23 · `Заказ Коалфолла` ru 18 · `煤落的委托` zh 3 · `Der Auftrag von Coalfall` de 2 · `コールフォールの依頼` ja 0.

**Non-fixture (16):** `Playing with Fire` **6** · `Ночной дозор (Tetralogy)` **32** · `Scepter of the Ancients` 0 · `The Lost Art of World Domination` 0 · `Bonus Keefe Story` 0 · `Everblaze` 0 · `Exile` 0 · `Keeper of the Lost Cities` 0 · `Neverseen` 0 · `Stellarlune` 0 · `Unlocked` 0 · `Unraveled` 0 · `Юный дрессировщик` 0 · `Ночной дозор (C2 throwaway)` 0 · `Ночной дозор (C2C3 run 2)` 0 · `Ночной дозор (C2C3 run)` 0.

Everything the record derives from this follows: **7** editions ✅ · 23 − 7 = **16** non-fixture ✅ · **14** at zero ✅ · exactly **two** non-fixture books non-zero ✅ · max non-fixture share **1.80%** ✅ · exactly one non-fixture book ≥ `MIN_ORPHAN_FOR_VERDICT = 20` ✅.

**The three throwaway duplicates do not double-count or mis-bucket.** All three sit under `Сергей Лукьяненко/Standalones/`, distinct from the real book under `.../The Night Watch Tetralogy/`. Of them, exactly **two** — `(C2C3 run 2)` and `(C2C3 run)` — have `hasCacheFile: false` and are the two `ok (not analysed)` rows, so the "two are never-analysed throwaways" claim is exactly right. ✅ The third, `(C2 throwaway)`, has a cache file — see N6.

Also re-verified on disk: five `cast-id-history.json` files, `2+2+1+5+4 = 14` retirements ✅, and no `rejected` key in any of them ✅.

---

### 🟡 Minor — new

**N6. The "two of them never-analysed" caveat `99d22735` just added is half the class — four of the 14 zeros are structurally forced, not measured.** *(cleanup)*

`docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1539-1542` and `:2810-2813`, `docs/testing/attribution-collapse-visibility-onbox-acceptance.md:222-225`, `docs/testing/onbox-acceptance-register-live-view.html:1364`.

`orphanSpoken` can only increment on a span that has at least one aligned sentence — `attribution-health.ts:221-226` sends `alignedForSpan.length === 0` to `unattributedSpeech` and `continue`s. So **any book with `unattributedSpeech === spokenTotal` has `orphanSpoken ≡ 0` by construction**, whatever its drift. Measured across the 14 non-fixture zeros, four are in that state:

| book | cache | cached sentences | spoken | unattr | attributable | disclosed? |
|---|---|---|---|---|---|---|
| `Ночной дозор (C2C3 run)` | absent | 0 | 220 | 220 | 0 | ✅ yes |
| `Ночной дозор (C2C3 run 2)` | absent | 0 | 220 | 220 | 0 | ✅ yes |
| `Ночной дозор (C2 throwaway)` | **present** | 0 | 220 | 220 | 0 | ❌ no |
| `Unlocked` | **present** | **9,569** | 2,057 | **2,057** | 0 | ❌ no |

`Unlocked` is the sharp one: it has been analysed, its cache holds 9,569 sentences, and *not one of them aligns to any of its 2,057 speech spans*. Its zero orphan count is as vacuous as a never-analysed book's, and it is not a throwaway duplicate — it is a real book in the reader-facing series.

What a reader concretely gets wrong: the sentence reads as *"12 real books were measured and found clean."* The genuinely-scoreable non-fixture population is **12**, of which **10** are at zero and 2 are not. The record's own design principle says this out loud in code — `attributionShare`'s comment requires a wholly-unattributed book to read as *"nothing to say", never as a confidently-healthy 0%* — and §5a duly prints `—` in the `share` column for all four. The prose then reads their orphan zero as a clean measurement anyway, which is the same D9 shape one column over.

Nothing moves: max non-fixture share is still 1.80%, still exactly one non-fixture book over the floor, *"there is no high group among real books at all"* still holds over the 12. It is a one-clause widening (`two of them` → `four of them, two never analysed at all and two whose caches leave every speech span unattributed`), and I would fold it before publishing rather than hold the merge for it. Flagged because it is precisely the class N5 opened and this commit closed at two sites out of four — the "one instance fixed, the class left armed" shape this review has now hit in all four rounds.

**N7. "Not one retired id is an orphan anywhere in the corpus" is false as written, and its counter-example is three sentences above it in the same paragraph.** *(correctness defect in the record; one-clause fix)*

`docs/superpowers/specs/2026-08-06-attribution-collapse-visibility-design.md:1534-1536`, `docs/testing/attribution-collapse-visibility-onbox-acceptance.md:258-260`, `docs/testing/onbox-acceptance-register-live-view.html:1364`.

> The corollary is a positive result for #2040 — fourteen retirements exist across five books and **not one retired id is an orphan anywhere**. — spec `:1534-1536`

Enumerated from the five history files on disk, the 14 retired ids are `mayrin`, **`coalfall`**, `lightning-dave`, `the-torment`, `lady-alina`, `vampire-man`, `igor`, `lena`, `vampire`, `polina-vasilievna`, `pavel`, `dark-programmer-2/3/4`. The corpus's orphan ids include **`coalfall`** — 28 spans in `The Coalfall Commission`, the single largest orphan id anywhere in the corpus. So the intersection is non-empty and the sentence is false under the scope it explicitly names ("anywhere in the corpus").

It contradicts its own paragraph. Three sentences earlier the same block says `Заказ Коалфолла` *records* `"coalfall": "coalfall-dragon"` in its `cast-id-history.json`, and that the English edition *carries 28*. The live view compresses both claims into one paragraph and keeps the contradiction intact, which is what makes this worth raising rather than leaving: it is on the page about to be republished.

The **intended** claim is per-book, and it is true — verified id-by-id: no id retired in a book appears as an orphan in *that* book (`Заказ Коалфолла` retires `mayrin`/`coalfall`, orphans `berrin`/`brann`/`lessom`/`unknown-male`; `Playing with Fire` retires `lightning-dave`/`the-torment`, orphans `pool-player-2`; `Everblaze` and `(C2C3 run 2)` have no orphans; the Tetralogy retires `pavel`/`dark-programmer-2/3/4`, orphans none of them). The load-bearing sentence that follows — *"every orphan in the corpus came from a path that bypassed the mechanism"* — is also true. Only the scope word is wrong. Fix: `anywhere in the corpus` → `in the book that retired it` (or `in its own book`), at the three sites. Nothing about the decision or any figure moves.

---

### ✅ What is solid

- **The 84.4% correction is right, and it is right by a second and independent method.** The prior agent's number, its span partition, and its identification of which two ids share the doubled span all reproduce exactly under a route that never replicates the loop it was checking.
- **The multiplicity assertion §1 makes is true corpus-wide** — swept, not assumed. Only one book, one span.
- **The classification underpinning 7 / 16 / 14 / "two throwaways" is correct book-by-book**, derived from the workspace rather than the table.
- **The §5a share table's denominator is stated, not implied** (run sheet `:172-173`: `orphanSpoken / (spokenTotal − unattributedSpeech − splitSpeech)`), and every row reproduces against it — including the two the complete run moved. `Ночной дозор` 32/1776 = 1.80% and `Playing with Fire` 6/1753 = 0.34%. Worth recording that this denominator is *not* `attributableSpoken` (1759 for that book, since 15 of its 32 orphan spans are also split spans), but the statistic is defined, self-consistent and stated, and the decision is invariant across every reading of it (1.51%–1.82%). Not a finding.
- **The five 84.4% sites agree and there is no sixth.** The multiplicity explanation added at each cites the correct code range and states the correct mechanism, and is internally consistent with §5c's own table (`driver` 21+1 = 22 ✅, `woman-in-taxi` 3+1 = 4 ✅).
- **`99d22735` introduced no new inconsistency.** Six hunks, four files; the run-sheet clause sits correctly inside the complete-run section, the spec clause reads cleanly around the sentence it interrupts, and the live-view edit changed markup nowhere.
- **House gates.** No paired test owed — verified, not assumed: `AttributionState = 'ok' | 'missing' | 'unmeasurable'` never contained `drifted`, both server edits are comment-only, `tsc` clean; nothing deleted or `.skip`ped. Release-notes skip argued rather than silently omitted. `Closes #2357` / `Refs #1984` present as plain text outside any code span (#2357 is still `OPEN`, so the PR's auto-close is the mechanism, as intended). Incidental fixes declared in the PR body. Register pair green and unmoved.
- **Nothing was written.** `C:\AudiobookWorkspace` has zero files modified in the last three hours; the primary checkout's `server/handoff/cache/` has zero files modified during the probe window and no `mns_*` cache touched at all; both `git status --porcelain` clean and both HEADs unchanged before and after. `readJsonWithRecovery` and `getOrHydrateManuscript` were confirmed read-only by inspection before running anything.

---

### Verdict

**Safe to merge: yes.** Two Minor findings, neither of which moves a figure, a threshold or the decision. Nothing found at the level this pass was dispatched to check: the 84.4% is correct under an independent derivation, the multiplicity is genuinely unique to one book, and the fixture classification holds book-by-book against the corpus itself. The three prior-round findings are all resolved, and the decision to drop D13 — together with both judgment calls in the PR body — remains correct and well-evidenced.

**Safe to publish: yes, but fold N7 first.** The live view is the one surface where both findings land on a public page, and N7's sentence is self-contradictory *within its own paragraph* there — three words fix it. N6 is a one-clause widening in the same paragraph and can ride along. Both are text-only, both are pre-existing in substance, and neither warrants a fifth round: N6 is cleanup and N7's fix is a scope word, so I would apply both, re-run `check:onbox-register`, and publish without re-review.

Issue #2256, opened 2026-08-11
srv — book-language backfill: punctuated junk (a numbered TOC) is indistinguishable from a CJK manuscript


dudarenok-maker:
## What

`scripts/repair-missing-book-language.mts` writes a book's `language` only when two
samples — the analysis cache and a fresh manuscript re-parse — independently detect
the **same** language with `fallback: false` (added in PR #2251, closing the review
finding that `fallback: false` alone is not confidence, since franc has no floor
above its `minLength: 30`).

That gate closes the reachable case it was built for, but not the whole class. **The
two reads are independent in derivation, not in content.** The cache is built from
the manuscript, so a book whose own text is unrepresentative carries the same junk
into both samples — and they agree.

Verified against the shipped gate:

| book text (both samples) | plan |
|---|---|
| TOC-only — `Prologue 1 Kaz 2 Inej 3 Kaz 4 Jesper…` | **`backfill 'es'`** |
| nav-only EPUB — `Cover Title Page Contents Chapter One…` | `backfill 'en'` |
| OCR noise — `rn lie tlie tbe iiit tne arid tbat wliich…` | `backfill 'en'` |

The `'en'` rows are unearned but harmless — that is the value a missing key already
resolves to via `normaliseBookLanguage`. **The `'es'` row is the real one**: per
`server/src/workspace/scan.ts:222-231` a non-`'en'` language forces every character
onto a designed Qwen voice and blocks the Kokoro cross-language fallback, so a wrong
non-English write is materially worse than the absent key it replaced.

What PR #2251's gate *does* close is the case that motivated it: the analysis cache
is per-chapter and resumable, so a book abandoned after chapter 0 has a cache of
front matter only while the manuscript holds real prose. Those two diverge, and the
book is refused.

## Acceptance

- A book whose entire text is a table of contents, a nav stub, or OCR noise is
  **skipped with a reason**, not backfilled — in particular it never receives a
  non-`'en'` language.
- A real book from the live corpus is still backfilled: replay the 7 books written
  in the 2026-08-11 run and confirm all 7 still resolve to `en`. A gate that also
  refuses those is a regression, not a fix.
- The threshold, whatever it is, is **measured against the real corpus** rather than
  picked — state the margin between the thinnest real book's sample and the thickest
  junk sample it must reject.

## Key files

- `scripts/repair-missing-book-language.mts` — `planBookLanguage`, the
  cross-source gate and its `KNOWN RESIDUAL` header note
[...]

PR #2360, opened 2026-08-14
fix(server): structural chapter-marker gate closes the 第N章 TOC backfill (#2341)

### LINKED ISSUE #2341: srv: 第N章-style Chinese chapter-numbered TOC still backfills zh (#2256 round-3 finding C2)
## What

`server/src/tts/prose-units.ts`'s `DIGIT_TOKEN_SHARE_CEILING` (0.2, raised
from 0.1 by #2256 round 2's finding 4) does not close the ticket's own
acceptance class for the single most common real Chinese chapter-numbering
layout: `第N章<title>` ("Chapter N: Title", using `第`/`章` as chapter
markers around a Han-numeral, fullwidth-digit, or ASCII-digit number).

Measured against the shipped code (60 distinct 4-character titles,
`第<numeral>章<title>。` repeated):

| numbering style | digitTokenShare | verdict |
|---|---|---|
| `N、<title>。` (this repo's own finding-2 regression fixture) | 0.24-0.36 | refused (correct) |
| `第一章<title>。` (Han numeral) | ≈0.15-0.17 | backfills `zh`, `fallback:false` |
| `第１章<title>。` (fullwidth digit) | ≈0.15-0.17 | backfills `zh`, `fallback:false` |
| `第1章<title>。` (ASCII digit) | ≈0.15-0.17 | backfills `zh`, `fallback:false` |
| `第一章<title>。`, 200 entries | ≈0.14 | backfills `zh`, `fallback:false` — scale does not rescue it |

Mechanism: `第` and `章` are ordinary Han characters (not numerals), so
they tokenize as two ordinary word tokens per TOC entry. A 4-character
title plus those two markers makes each entry ~7 tokens with only 1
digit-like token — under the ceiling regardless of entry count.

## Why this isn't a quick threshold fix

`DIGIT_TOKEN_SHARE_CEILING` cannot be lowered back toward 0.1-0.15 without
re-opening #2256 round 2's finding 4 (a genuinely short-sentence real book
with verse-style numbering measured ≈0.10-0.14 — the ceiling was raised
specifically because real content sits in that range). The two ranges
overlap:

- worst real short-sentence content: ≈0.10-0.14
- this junk shape (`第N章<title>`): ≈0.14-0.17

No single `DIGIT_TOKEN_SHARE_CEILING` value separates both without either
refusing real short-sentence books or backfilling `第N章`-style TOCs.

## The decision owed

Closing this needs a signal that is not a token ratio — most plausibly a
STRUCTURAL check for the numbering *pattern* repeating once per prose unit
(e.g. a `第<numeral>章` — or generalized numeral-adjacent-marker — prefix
recurring across most units), which is a different instrument from
Guiraud's R / digitTokenShare. Open questions for whoever picks this up:

1. Exactly which pattern(s) count as "structural numbering" — just
   `第...章`, or a broader class (`卷`, `回`, `节`, a bare leading numeral
   run at the very start of a unit)?
2. How is false-positive risk bounded — real prose with a recurring
   sentence-opener (a diary/epistolary frame, a refrain) could resemble
   this structurally.
3. Does this become a THIRD independent gate (alongside
   `LEXICAL_RICHNESS_FLOOR` / `DIGIT_TOKEN_SHARE_CEILING`), or does it
   replace `DIGIT_TOKEN_SHARE_CEILING` entirely?

Found and documented (not fixed) by the #2256 round-3 independent review.
See `server/src/tts/prose-units.ts`'s own `LEXICAL_RICHNESS_FLOOR`/
`DIGIT_TOKEN_SHARE_CEILING` header (finding C2) and
`server/src/tts/detect-language.test.ts` (finding C2 test, which pins the
current — imperfect — behavior so it's visible rather than silently
"working").

Refs #2256.

dudarenok-maker:
## Summary

Closes #2341 — a CJK "第N章<title>。" (Chapter N) numbered table of contents was backfilling `zh` on import. Its digit-token share (~0.17) sits *below* `DIGIT_TOKEN_SHARE_CEILING` (0.2) and overlaps the digit share of genuine short-sentence prose (~0.10–0.14), so no ceiling value can separate the two. Adds a **third, structural junk gate** (`chapterMarkerUnitShare` / `CHAPTER_MARKER_UNIT_MAJORITY` in `server/src/tts/prose-units.ts`, wired into `voteLanguage` in `server/src/tts/detect-language.ts`): a "第<numeral>章" prefix on a *strict majority* of the winning sample's prose units is a numbered TOC by construction (real prose never opens even half its units with one, and a 第N年 date or a mid-sentence mention lacks the 章 ender). A share above the majority surrenders to `{language:'en', fallback:true}` exactly like the other junk gates, order-invariant via the same `SENTENCE_TERMINAL_RE` segmentation. Flips the previously-pinned "finding C2" test to the desired behavior and adds regression locks covering all three numeral renderings (Han / fullwidth / ASCII) plus real-zh-prose no-false-positive. Header comments (`prose-units.ts`, `scripts/repair-missing-book-language.mts`) and the release note are updated to mark the gap closed.

## Test plan

- [ ] `npm run verify:fast:branch` — green (lint, typecheck, test:hooks, budget-poll, test:server 7,721 passed, build; frontend-test/sidecar out of scope)
- [ ] `npx vitest run src/tts/detect-language.test.ts` — 70/70 (flipped finding C2 + new variants/regression tests)
- [ ] Reviewer: spot-check `chapterMarkerUnitShare`/`CHAPTER_MARKER_UNIT_MAJORITY` in `server/src/tts/prose-units.ts` and the third gate in `server/src/tts/detect-language.ts` ~L277-293


dudarenok-maker:
## PR review — pass 1 (head b85d3003cfb663e55a34bf146b9fc08e129fea8f, effort medium)

Reviewed the gate mechanism, the test flip / new regression tests, and the house gates in the brief. The core fix is sound: `chapterMarkerUnitShare` fires on the `第<numeral>章` numbered TOC (share 1.0 on the C2 fixture), does not fire on the `第N年`/date shapes or the mid-sentence mention, is wired only into `voteLanguage` (the chapter-aware path, `detect-language.ts:292` — `detectManuscriptLanguage`'s script pre-pass never routes here and never had the ratio gates either), and is order-invariant through the same `SENTENCE_TERMINAL_RE` segmentation. The flipped C2 test and the variants/regression tests are real: red on `main` (no gate → the shape backfills `zh`, so the new `{en, fallback:true}` assertion fails for the right reason), green here (70/70 run locally). `Closes #2341` is present in the PR body outside any code span; the `docs/release-notes-next.md` one-line edit is an accurate addend to the existing #2256 entry.

Three findings, all cleanup-correctness (a correctness bug re-review trigger, per the label split, is not required by any of them).

---

### 🟡 Minor — finding C2 header label left self-contradictory ("NOT CLOSED")

- **`server/src/tts/prose-units.ts:333`**
- The finding-C2 block header still reads `HIGH, NOT CLOSED — owner decision named, not guessed at:`, while lines 365–380 immediately below it say `FIXED HERE by #2341 (2026-08-14) … Closes #2341; see detect-language.test.ts flipped finding C2 test`. The PR body claims "Header comments … are updated to mark the gap closed," but this one spot that explicitly said *open* was not touched, so the same file now asserts both states for the same finding. Concrete failure: a reader doing a grep for `NOT CLOSED` (as a "what is still owed" audit) lands on C2 here and records a false open item — the instrument, not the code, is wrong (brief shape 9). This pre-dates the PR's arrival but the change made it actively false and was the round for it; update line 333 to `CLOSED by #2341`.

### 🟡 Minor — the third gate is not extended into the repair script's surrender-reason composer

- **`scripts/repair-missing-book-language.mts:255-311`** (and the `SurrenderDiagnostic` union at `:212-216`)
- `describeSurrenderReason` replays only two of the three junk gates (`too-repetitive` at `:303`, `too-numbered` at `:307`) and otherwise returns `null`. For the exact shape this PR closes — a `第N章` numbered TOC — the *decision* from `detectManuscriptLanguageFromChapters` newly surrenders via the chapter gate, but `describeSurrenderReason` on that same sample computes `richness ≥ floor` and `digitShare ≈0.17 ≤ 0.2`, matches neither branch, returns `null`, and `planBookLanguage` emits the generic `detection surrendered (confidence-floor guess 'zh') from … chapters — a guess is never written` reason. So the report does not name the chapter-marker gate that actually fired, contradicting the guarantee in this function's own header (`:238-253`, "naming which gate fired") and the release-note sentence "all three … classes … surrender with a reason naming which gate fired." One instance fixed (gate in `voteLanguage`), the sibling reason path left armed (brief shape 10). Decision correctness is unaffected — this degrades the *explanation*, not the verdict — hence Minor; add a `'too-many-chapter-markers'` kind to `SurrenderDiagnostic` and a matching branch before the `null` return.

### 🟡 Minor — `chapterMarkerUnitShare`'s deliberate no-dedup leaves it blind to real prose with a unit-initial `第X章` marker repeated verbatim

- **`server/src/tts/prose-units.ts:616-628`**
- The sibling gates dedupe before measuring precisely so that a real sentence repeated N times (this repo's own thick-chapter fixture convention, `Array(N).fill(REAL_SENTENCE)`) measures as one occurrence; `chapterMarkerUnitShare` deliberately does not, and its justification weighs only the real-TOC side ("a numbered TOC's entries are each distinct already"), not the repeated-real-prose side. Concrete failure: body `Array(60).fill("第一章我们已经学会用火做饭了。").join('')` — post-dedup roughly six distinct word tokens and one digit token, so `guiraudR` and `digitTokenShare` both pass, then `chapterMarkerUnitShare` counts 60/60 = 1.0 and surrenders `{en, fallback:true}`, refusing a verbatim-repetition zh sample the ratio gates would accept. It is fail-closed (surrenders rather than writing a wrong language) and requires a sentence-initial `第X章` repeated past a majority, so it is not a realistic real-book shape — hence Minor, but the asymmetry the header asserts ("dedup … would only obscure the harness's same intent") does not hold for repeated real prose, and a future fixture built on this repo's own repeat-a-sentence convention inherits the trap.

---

Verdict: pass with three Minor cleanup findings; no Blocking/Significant correctness defect. The gate's firing/non-firing behavior, the test truth (red-before/green-after for the right reason), and the single wiring point in the chapter-aware path are all confirmed.

dudarenok-maker:
## Review follow-up — findings triaged and fixed (head c41d65bf)

Pass 1 (head b85d3003, effort medium) returned three 🟡 Minor findings — no
correctness/blocking/significant defect, so no re-review pass is triggered per
the review contract. All three were **fixed in this round**, as the
report-fix-record rule requires:

1. 🟡 **Stale "HIGH, NOT CLOSED" round-3 label** (`server/src/tts/prose-units.ts`)
   — the finding-C2 round note claimed "NOT CLOSED" after the change already
   shipped the closing gate; corrected to mark it CLOSED by the #2341
   structural gate. → commit `c1f0a9d5`.
2. 🟡 **Repair script's `describeSurrenderReason` blind to the third gate**
   (`scripts/repair-missing-book-language.mts`) — a `第N章` TOC surrender
   reported the generic "confidence-floor guess" because the diagnostic only
   matched the two token-ratio gates. Added the `chapterMarkerUnitShare`
   branch (new `'numbered-toc'` `SurrenderDiagnostic` + reason string) so the
   reason names the structural gate. Paired regression test added in
   `scripts/tests/repair-missing-book-language.inner.test.mts`
   (`tsx --test` → 41/41 pass). → commit `c41d65bf`.
3. 🟡 **No-dedup blind spot in `chapterMarkerUnitShare`** — now dedupes EXACT
   prose units first, matching the sibling gates' documented
   "dedupe EXACT prose units before measuring" invariant, so a
   verbatim-repeated unit cannot inflate its own share while a numbered
   TOC's distinct entries still sit at share ≈1. → commit `c1f0a9d5`.

Local gate re-ran green on the fixed head: `npm run verify:fast:branch`
**EXIT:0** — lint, typecheck (incl. the repair-script type changes),
test:hooks (incl. the new inner test), test:server (single clean pass),
build; frontend-test / sidecar out of scope. Cloud `verify.yml` is re-running
on the new head.

Issue #2165, opened 2026-08-06
cast.json can land in a resurrected pre-rename folder when a book is renamed mid-analysis


dudarenok-maker:
## What

Renaming a book's Author / Series / Title while an analysis is running can send that run's `cast.json` writes to the **old, pre-rename** directory — recreating it — while other files written by the same run land in the new one.

## Mechanism

`routes/book-state.ts:856` mutates the record in place on rename:

```ts
rec.bookDir = newDir;
```

The folder itself is moved by `renameWithRetry`. Because the mutation is by reference, anything that reads `record.bookDir` *at write time* follows the rename automatically. That is how the five merge-base writes behaved before #2155: `writeJsonAtomic(castJsonPath(record.bookDir), …)` re-read the field on every write.

After #2155 those writes go through a per-run `castBase` built once at job start:

```ts
const castBase = record.bookDir ? createCastMergeBase(record.bookDir, priorSnapshot.fingerprint) : null;
```

`createCastMergeBase` closes over the `bookDir` string it was given. A rename landing mid-run does not reach it. `writeJsonAtomic` `mkdir`s its parent, so the write does not fail — it **recreates the old directory** and writes `cast.json` into it.

The result is a split: `cast.json` in the resurrected old folder, `manuscript-edits.json` (still resolved via `record.bookDir` at write time) in the new one.

Note the site guards make this invisible to type checking: they test `record.bookDir` for non-null, which is still perfectly non-null — it is simply a *different* directory than the one `castBase` holds.

## Why it wasn't caught

- Trigger is narrow: a rename concurrent with a running analysis.
- The `castBase!` non-null assertions are all sound — `bookDir` only ever goes non-null → non-null, so nullness never flips. The defect is the *value* changing, not becoming null.
- Partly pre-existing in shape: `readPriorCastForMerge` already captured `bookDir` at job start. What #2155 changed is that the **writes** now do too.

## Acceptance

A decision, recorded here, on which of these is right — they are genuinely different products, not implementation details:

1. **Follow the rename** — `createCastMergeBase` takes a `() => string` accessor instead of a string, so every `writeChecked` resolves the current `bookDir`. Keeps pre-#2155 behaviour. Note the fingerprint baseline then describes a file in a directory that may no longer be the one being compared against, which needs thinking through.
2. **Pin and refuse** — the run detects that `bookDir` moved and aborts or warns rather than writing to a stale path. Arguably the honest answer: a rename mid-analysis is a real conflict.
[...]

PR #2361, opened 2026-08-14
fix(server): detached-run STALE_BOOK_DIR fail-loud; harden rename-midrun guards (#2196)

### LINKED ISSUE #2196: An out-of-process book folder move leaves a stale record that resurrects the old directory
## What

Renaming or moving a book's folder **outside the process** — in Explorer, or by OneDrive/Dropbox relocating it — leaves the server's in-memory `ManuscriptRecord` pointing at a directory that no longer exists. Every subsequent write for that book targets the old path, and `writeJsonAtomic` `mkdir`s its parent (`state-io.ts`), so it **recreates the stale directory** rather than failing.

## Why this is separate from #2165

#2165 covers the *in-process* rename (`PUT` book metadata → `renameWithRetry` → `rec.bookDir = newDir`) and is being fixed two ways: a 409 refusing the rename while an analysis is in flight, plus making the merge-base writes follow the rename.

Neither layer helps here:

- The 409 guard lives in `book-state.ts`'s rename branch, which an out-of-process move never enters.
- "Follow the rename" reads `record.bookDir`, but an out-of-process move never runs `rec.bookDir = newDir` — so the field itself is stale. **The whole record is wrong, not one field**, which makes this a different shape of problem: record invalidation / workspace re-scan, not accessor threading.

## Acceptance

A decision, recorded here, on what the server should do when a book's directory disappears from under a live record. Candidates, not a menu to implement blindly:

1. **Detect and invalidate** — writes verify the directory still exists (or its `state.json` still carries the expected `manuscriptId`) and, on a miss, drop the record and force a workspace re-scan rather than recreating the path.
2. **Re-scan on demand** — a periodic or event-driven workspace scan reconciles records against disk, so a moved book is re-hydrated at its new path.
3. **Accept and document** — treat an out-of-process move as unsupported while the app is running, and make the failure loud (refuse the write with a clear error) instead of silently resurrecting a directory.

Whichever is chosen needs a regression test that moves a book directory behind the server's back and asserts the next write does **not** recreate the old path.

## Key files

- `server/src/workspace/state-io.ts` — `writeJsonAtomic`'s `mkdir(dirname(path), { recursive: true })`, the mechanism that turns a stale path into a resurrected directory
- `server/src/routes/book-state.ts` — where the in-process rename keeps the record honest, for contrast
- `server/src/workspace/active-analyses.ts` — the existing disk-scan precedent for reconciling against the filesystem

## Benefit

*Benefit (user):* moving a book folder in Explorer while the app is open cannot silently split that book's state across two directories, leaving a library entry the app can no longer read.

Found while planning #2165 (see `docs/superpowers/plans/2026-08-07-rename-midrun-bookdir-pinning.md`, "One further gap, explicitly out of scope"), and deliberately excluded there — it needs a different fix shape than the one that plan builds.

dudarenok-maker:
Summary

Implements the missing fail-loud for a detached analysis run whose book directory
becomes unresolvable mid-run: the always-on STALE_BOOK_DIR server log now fires once
in the run's top-level BookDirUnresolvedError branch (previously it was swallowed
per-persist-catch, so a zero-subscriber run logged nothing before the in-memory
marker was deregistered). Subset analysis mirrors it.

Hardens the rename-midrun guard tests: the #2196 regression tests were not halting
because the Phase-1 analyzer stub REJECTED runAttributionEscalation, crashing runs
mid-Phase-1 with Error: not used before the guard's persist gate could emit the
STALE_BOOK_DIR terminal. The escalation stub is now a no-op (the suite convention),
STRUCTURE_ENGINE is pinned off, minorCastMinLines is pinned to 1 via a throwaway
user-settings file so the completing in-tree run keeps the 1-line nova through the
minor-cast fold, and the #2165 analysis-state case gets a dedicated failing phase-1
analyzer for its halted terminal.

Closes #2196

## Review fixes

Independent review pass 1 (effort medium) surfaced 1 significant + 3 minor findings, all resolved; re-review pass 2 confirmed the significant finding resolved and left only 2 cleanups, both now resolved:

- **Significant � transient state.json unreadability terminated a healthy run.** The identity read now retries (bounded 3 attempts, 50ms backoff) before the slow path; paired regression test proves a single transient failure still verifies ok with the store record untouched. (ook-dir-guard.ts)
- **Minor (pass 1) � mid-block TOCTOU resurrection.** The write target is re-verified (identity-gated) immediately before the terminal state.json record in both persist blocks, via a testable ssertWriteTargetStable helper with a deterministic divergence regression test that proves a mid-persist in-tree move to a different dir refuses the terminal write (halts).
- **Minor (pass 1) � dead guard surface.** 	ryResolveVerifiedBookDir + identityBearing:false now have a production call site (main-success snapshot deletion resolves-then-deletes); 
esolveVerifiedBookDirForRun binds the manuscript's ookId via expectedBookId on the full path.
- **Minor (pass 1) � release-notes pair.** Technical entry in docs/release-notes-next.md + user-facing bullet in RELEASE_NOTES.md.
- **Minor (pass 2) � overclaiming comment.** Corrected to state exactly what expectedBookId enforces (full path; re-hydration re-locates by manuscriptId).

## Test plan

- [ ] analysis.rename-midrun.test.ts: 7/7 green (2x #2165 + 4x #2196 + TOCTOU divergence)
- [ ] book-dir-guard.test.ts (14) + cast-merge-base.test.ts (17): 31/31 green
- [ ] analysis.test.ts: 216/216 green
- [ ] combined run: 254 green
- [ ] tsc --noEmit: clean


dudarenok-maker:
## PR review — pass 1 (head 3f5d2741788fbd4074c06b5365139156aa9d7a64, effort medium)

Reviewed `main...3f5d2741` for PR #2361 (book-dir guard, issue #2196). The regression tests are genuinely red-before/green-after (I traced each: the cross-book, out-of-tree-move, pathless-halt and rehydrate cases all assert on output that the unguarded code provably diverged on), and `Closes #2196` is present and outside code spans. Findings split below.

### Correctness bugs

**🟠 Significant — an otherwise-healthy run is now terminally halted by a *transient* `.audiobook/state.json` unreadability.**
`server/src/workspace/book-dir-guard.ts:89` (+ slow path `:102-103`), surfaced at `server/src/routes/analysis.ts:5629` and `:7243`.

Concrete scenario: a workspace book whose folder, cast.json and in-memory `ManuscriptRecord` are all intact, but whose `.audiobook/state.json` is momentarily unreadable — an AV/OneDrive lock, transient EMFILE/ENOENT, a half-finished cloud sync, or an interrupted migration. The guard's full-path check does `readJson(stateJsonPath(candidate)).catch(() => null)` and treats "null" exactly like "the folder moved away": it falls into the slow path, `removeManuscript(manuscriptId)` + `getOrHydrateManuscript`, whose `findBookByManuscriptId` predicate (`state.manuscriptId === manuscriptId`, scan.ts) also cannot read the file → returns `undefined` → `BookDirUnresolvedError` → the persist block throws at `analysis.ts:5629` → the run ends `kind:'error', code:'STALE_BOOK_DIR'`, writing nothing and reporting no result.

Pre-change this exact blip was absorbed: the persist read of the same file at the flag-preservation site was `await readJson(statePath)` inside `if (prev)` plus a non-fatal `catch (persistErr)`, so cast.json was still written and the run completed. The new guard cannot distinguish "book genuinely gone/cross-book-moved" from "same book, identity file transiently illegible", and treats both as terminal — it halts a multi-minute run and discards its output over a pure I/O blip that `cast-fingerprint` explicitly tolerates elsewhere (#2185: a transient UNREADABLE must never abort). None of the new tests measure this case (they all use a present-and-matching state.json), so the gap is invisible to the suite. At minimum the guard should retry the read before classify-as-gone, or the full-path failure should only invalidate when the candidate dir itself is also absent.

### Cleanup-only nits

**🟡 Minor — the claimed "before ANY caller writes it resolves a verified bookDir" guarantee still has an unguarded mid-block window.** `server/src/routes/analysis.ts:5629-5902` (subset `:7243-7444`): only the block-top `resolveVerifiedBookDirForRun` and the per-write `castBase` resolver are identity-gated. `writeFoldJournal`, `writeDedupJournal`, `writeSuggestions`, `recordRetirements`, `writeJsonAtomic(manuscriptEditsJsonPath(writeDir))` and `writeJsonAtomic(statePath, …)` all write through the single `writeDir`, with no re-verification. A move landing between the block-top resolve and those writes re-creates the moved-away folder via `writeJsonAtomic`'s `mkdir(dirname,{recursive:true})` — the exact resurrection the PR is meant to close. The `catch(persistErr)`/park only catches `BookDirUnresolvedError` from the cast resolver, so the non-cast writes in the window resurrect silently. Pre-existing race, not a regression, but the PR's "no ordering bet" framing overstates the coverage.

**🟡 Minor — the guard ships three arms unreachable at the default/production configuration.** `identityBearing:false` (`book-dir-guard.ts:78`), `expectedBookId` (`:51` `:93`) and `tryResolveVerifiedBookDir` (`:132`) are exercised only by `book-dir-guard.test.ts`, never by any production call site in `analysis.ts`. Every shipped write path uses the default `identityBearing:true` / `verifyBookDirForWrite` / `withVerifiedBookDir`. Dead test-only API surface (brief shape 7); shrink the public surface or delete the unused arms.

**🟡 Minor — house gate: no release-notes pair and no explicit not-applicable.** The PR body has no `docs/release-notes-next.md`/`RELEASE_NOTES.md` entry and no "not-applicable" declaration for this server-behavior fix, which per the before-shipping checklist needs one or the other.

dudarenok-maker:
## PR review — pass 2 (head ad1dc87287ba092d0ab4fa860bad38db05ca853c, effort medium)

Verified each of the four claim groups by reading the code and running the suites (`book-dir-guard.test.ts` 14/14, `cast-merge-base.test.ts` 17/17, `analysis.rename-midrun.test.ts` 6/6 green; `tsc --noEmit` clean; worktree `git status` untouched).

**Significant bug (transient state.json read treated as an identity miss) — RESOLVED.** `readStateWithRetry` (book-dir-guard.ts:39-48) does a bounded 3-attempt / 50 ms backoff read and only returns null when it is persistently unreadable; the retry short-circuits correctly on a read that *succeeds* but carries a wrong identity (returns immediately, no wasted retries). The regression test (book-dir-guard.test.ts:166-190) proves a single transient failure still verifies `ok` with the store record untouched, and the counter-drain assertion proves the retry actually absorbed the blip rather than the test passing vacuously. Genuinely fixed.

**Minor x2 (TOCTOU re-verify; dead guard surface incl. expectedBookId).** The write target is re-verified immediately before the terminal state.json record in both persists (analysis.ts:5890-5895 main, 7449-7454 subset), throwing `BookDirUnresolvedError` on divergence; `tryResolveVerifiedBookDir` now has a real production call site (analysis.ts:2959, main-success snapshot cleanup) and `expectedBookId` is bound from the manuscript record (analysis.ts:2739-2741). The dead-surface wiring is real and the cross-book UNIT path is covered. The run-top-level `STALE_BOOK_DIR` fail-loud is now emitted once per run from the catch (analysis.ts:6022, 7552), not swallowed per-persist-catch, and the rename-midrun suite asserts it for both detached (console) and attached (SSE) surfaces. Release-notes pair present and matching in both files.

Both remaining items are cleanup-only nits, no correctness bugs.

### 🟡 Minor (cleanup — correctness split: 0 bugs / 2 cleanup)

1. **`file:line` analysis.ts:2737-2741 + book-dir-guard.ts:126-132** — the comment claims "a re-hydration that comes back with a foreign identity also fails," but the slow path returns the re-hydrated `freshDir` without any `expectedBookId` check. Scenario: a corrupted workplace where two `.audiobook/state.json` files carry the same `manuscriptId` but a different `bookId`; the candidate is refused by `expectedBookId`, then `findBookByManuscriptId` re-acquires that same wrong-bookId dir and the guard returns `ok`. Contrived (sharing a manuscriptId is already corrupt), but the comment names protection the code does not implement — either enforce it on the re-hydrated path or soften the comment.

2. **`file:line` analysis.ts:5890-5895 / 7449-7454 (test coverage)** — the mid-persist divergence branch (`stateWriteDir !== writeDir`) has no direct regression test: the rename-midrun suite moves the folder during Phase 1, so the block-top resolve throws/misses and never reaches this specific branch. The branch is defensively sound and failure-safe, but as the TOCTOU fix's named surface it deserves a test that moves the folder *within* the persist window (e.g. between the block-top resolve and the re-verify).

### ✅ Verdict on prior significant finding
Genuinely resolved (yes): the retry + regression test both exist, are correct, and the suite is green; only a persistently unreadable or truly mismatched identity now falls through to the slow path.

No 🔴/🟠 correctness findings. This pass is cleanup-only, so it should not re-trigger a fresh review on its own.

PR #2364, opened 2026-08-14
chore: add AGENTS.md and .clinerules for Cline compatibility

### LINKED ISSUE #2363: chore: add AGENTS.md and .clinerules for Cline compatibility
## Summary

Add cross-tool agent configuration so Claude Code, Cline, and any other tool resolve the same canonical project context.
- AGENTS.md - thin cross-tool hub pointing at the canonical CLAUDE.md (with an @CLAUDE.md include).
- .clinerules/cline.md - Cline session rules: Conventional Commits, verify-before-commit/push, testing discipline, the sub-agent-executed delivery doctrine, and an honest note that Cline cannot select subagent model tiers.

Files: AGENTS.md, .clinerules/cline.md

dudarenok-maker:
## Summary

Adds cross-tool agent configuration so Claude Code, Cline, and other tools reach the same canonical project context.

- **`AGENTS.md`** — thin cross-tool hub. Names `CLAUDE.md` and `CONTRIBUTING.md` as the source of truth (the branching / commit / PR / issue / release-notes specs live in the latter, not the former) and states, per tool, whether they are loaded automatically or have to be read.
- **`.clinerules/cline.md`** — Cline session rules. Opens by telling the session that **`CLAUDE.md` and `CONTRIBUTING.md` are not loaded for it** and must be read first. Beyond that it carries only rules that are **Cline-specific** (where skills actually resolve from; model tiers not being selectable, and the `pr-review-gate` consequence; the PowerShell 5.1 shell) or that **fail silently** (a fresh worktree's hooks doing nothing while looking normal; worktree teardown order; `.claude/` being git-ignored wholesale; git-ignored artifacts belonging to the primary checkout), plus the gates a PR must clear and what every change owes.

Also corrects the same over-claim at its source in **`CLAUDE.md`** (the `skills:sync` bullet) — see pass 3 below.

Closes dudarenok-maker/Castwright#2363
Refs dudarenok-maker/open-engine#50

## The premise this PR was written on was wrong

The first version of both files asserted that **Cline reads `CLAUDE.md` and `CONTRIBUTING.md`**, and was designed around it — `.clinerules/cline.md` deliberately restated nothing, and `AGENTS.md`'s `@CLAUDE.md` import was removed as redundant.

**It does not read either.** Verified two ways, twice, independently: `@cline/core` (the package holding the rule loader) contains **zero** `CLAUDE` byte sequences under a binary-safe grep, against positive controls that do find `clinerules` and `AGENTS.md`; `CONTRIBUTING` appears nowhere in `@cline/*` outside a vendored `vercel/ai` README link; and a live `cline -p` probe in this workspace reported only `Workspace AGENTS.md` + `.clinerules/cline.md` in context.

So both files were telling a Cline session its canonical context was already loaded — guaranteeing it would never go read it. That is the one defect class a rules file can have that actually matters.

**Fix:** an explicit read-these-first instruction in both files, with the verification method and date recorded so the claim is auditable rather than asserted. Chosen over restoring an `@CLAUDE.md` import (~90 KB in every session, and it is unverified that Cline expands `@`-includes at all) and over making `.clinerules` self-sufficient (which would reintroduce the untested duplication that the pass-1 review removed). One caveat is recorded in the file itself, so nobody re-derives this and concludes the note is wrong: `cline.exe` does carry `CLAUDE.md` literals, in a `claudeMd` / `claudeMdExcludes` settings schema belonging to the **bundled Claude Agent SDK** that Cline vendors — not to its own rule loader. A vendored schema is not a loader, so no setting flips those strings into `CLAUDE.md` support. (Pass 3 corrected an earlier version of this sentence that called the schema enterprise policy and named a managed config as a staleness mechanism — it is neither.)

## Review record

Three rounds, each folded rather than deferred.

**Pass 1** (head `4ff5f44d`) — 1 blocking, 3 significant, 3 minor. Fixed in `f764e4ff`:

1. 🔴 the Cline premise above.
2. 🟠 `server/.env.example` is only **partly** generated — `config:sync` owns the `BEGIN`/`END` block (453–695 of 706 lines), so ~460 lines including `GEMINI_API_KEY=` at line 39 are hand-authored, and `config:check` never looks outside the block. "Never hand-edit" was false for most of the file.
3. 🟠 dropped "run `verify:fast:scoped` before staging" — it scopes to the **staged** diff, so on an empty index all four legs report `(out of scope)` and it exits 0 having run nothing. Advice to run a guaranteed-vacuous check.
4. 🟠 added the worktree-teardown warning — `git worktree remove` follows a junction and deletes the primary checkout's real `node_modules`. Load-bearing here precisely *because* the blocking finding means Cline never sees CLAUDE.md's copy.
5. 🟡 `--no-verify` "reserved for CI-generated commits" contradicted CONTRIBUTING.md and CLAUDE.md.
6. 🟡 model-tier text asserted a mechanism instead of reporting what the probe measured.
7. 🟡 dudarenok-maker/Castwright#2363 carried only `area:ops`, so it rendered in neither board view — now `type:chore` too.

**Pass 2** (head `f764e4ff`) — confirmed 5 of 7 resolved, and caught two of the fixes being wrong. Fixed in `cf43b275`:

8. 🟠 the `--no-verify` carve-out from fix 5 was **not in its cited source**. `CLAUDE.md:1182` states the rule flatly; the `:871` parenthetical it was read out of is scoped to deliberately bypassing `guard-protected-push` for an intended force-push, not to skipping the verify battery. Now names the single documented exception and spells out why "I already ran the tests" is not a second one — pre-push also runs `guard-protected-push` and `guard-commit-subjects`, and `verify:fast:branch` runs neither.
9. 🟡 fix 4's "taking husky activation with it" was false — `.husky/_` is a real repo-root directory, so hooks keep firing and fail **loudly**. Reframed around what is actually silent.
10. 🟡 "and nothing else" overstated the loader — it also reads a bare `.clinerules` file and `.cline/`, and takes `.markdown` / `.txt`.
11. 🟠 this PR body still carried the disproved premise. Fixed here.

**Pass 3** (head `cf43b275`) — **approve**, all four pass-2 findings verified resolved. Three cleanup findings, fixed in `fcc53b0f` rather than deferred:

12. 🟡 "the single documented exception" for `--no-verify` missed a second one — `CLAUDE.md` "Commit gate" and `.husky/pre-push:3` both document a deliberate `git push --no-verify` for an intended force-push/deletion. Third round this paragraph landed off its sources, so it now cites all three.
13. 🟡 the `cline.exe` caveat mis-attributed its own evidence. `claudeMd` / `claudeMdExcludes` are not enterprise policy — `claudeMdExcludes`' own description scopes it to User/Project/Local memory, and both keys belong to the bundled **Claude Agent SDK** settings schema Cline vendors. So "a managed config could make this stale" named a mechanism that cannot happen. Reworded to what the evidence supports.
14. 🟡 **"Skills resolve from `~/.agents/skills/` only" was an exhaustive claim the probe never established** — it tested two paths, while the loader composes skill roots from its rule directories (`skillsPath: join(<ruleDir>, "skills")`), so workspace roots may work too. Corrected in **both** files: `.clinerules/cline.md` inherited the sentence from `CLAUDE.md:425`, which is where the over-claim actually lives, so fixing only the copy would have left the source to re-propagate it. Neither now asserts an exhaustive list in either direction — code presence is not proof (`~/.cline/skills` is in that same list and was proven dead) — and dudarenok-maker/open-engine#50 is filed for the probe. This matters beyond accuracy: if a workspace root resolves, the per-machine `skills:sync` step becomes avoidable.

Gate discharged: approve verdict, and 🟡 findings do not re-trigger a round.

## Not applicable

- **Paired test** — no runtime surface, and nothing reads either file at runtime, so there is no assertion to add. That absence is itself why finding 2 of pass 1 removed the duplicated commit-type/scope lists instead of adding a drift guard: `scripts/verify-cache.mjs` lists `CLAUDE.md` in `test:hooks`'s `extraFiles` only because a test reads it, and registering these two paths would point at a guard that does not exist. `CLAUDE.md` is different — it *is* covered, which is why this PR's final commit ran the suite.
- **Release notes** — pure process/tooling change, no user- or operator-visible delta, per CLAUDE.md before-shipping step 5.
- **On-box acceptance** — no behaviour requiring real hardware.

## Test plan

- [x] Pre-commit `verify:fast:scoped` — every leg `(out of scope)`; green. Noted above that this is a vacuous green for this diff, so it is recorded as *ran*, not as *evidence*.
- [x] Pre-push `verify:fast:branch` — every leg `(out of scope)`; green. Same caveat.
- [x] `pr-title-lint` / `pr-issue-link` / CLA — green.
- [x] `test:hooks` — **ran for real** at `fcc53b0f` (the `CLAUDE.md` edit busts its cache, unlike the two earlier commits): 41 pass / 0 fail and 1402 pass / 0 fail, including `review-gate-mechanism.test.mjs`, which reads `CLAUDE.md` prose.
- [x] `pr-review-gate` — 3 passes, verdict **approve** at pass 3. Not docs-only, so the exemption never applied.
- [ ] CI `verify.yml` — aggregator green at `fcc53b0f`.

### DIFF
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -1,57 +1,124 @@
 # Cline project rules (Castwright / Audiobook-Generator)
 
-Loaded by Cline each session. The full project context is in the root
-`CLAUDE.md` (Cline reads it too) -- these are Cline-specific operational
-prefs layered on top. Keep this minimal; move anything of general value into
-`CLAUDE.md` instead.
+Loaded by Cline each session. The canonical context is the root `CLAUDE.md`
+and `CONTRIBUTING.md` -- Cline reads both, so nothing here restates them. This
+file carries only two kinds of rule: ones that are **Cline-specific**, and ones
+that are **costly if missed** and easy to lose in 140 KB of prose. Keep it that
+way -- anything of general value belongs in `CLAUDE.md` instead.
 
-## Commits (Conventional Commits, enforced by hooks)
-- Subject format: `<type>(<scope>): <subject>`.
-  - Types: `feat | fix | refactor | perf | test | docs | build | ci | chore`
-  - Scopes (required except for `chore`): `frontend | server | sidecar | app |
-    scripts | e2e | mocks | openapi | docs | deps | ci | ops`
-  - Subject <= 100 chars. Breaking change = `!` before the colon
-    (e.g. `feat(server)!: drop legacy field`).
-  - Multi-line commit bodies end with a `Co-Authored-By:` trailer naming the
-    agent (e.g. `Co-Authored-By: Cline <...>`).
-- The `commit-msg` hook rejects malformed subjects; `pre-push` re-checks them,
-  refuses force-push/deletion of `main`, and runs `verify:fast:branch`.
+## Cline-specific
 
-## Verify before commit / push
-- `pre-commit` runs `npm run verify:fast:scoped` (scoped to STAGED files). Run
-  it before staging so hooks pass on the first try.
-- The fast local gate before pushing is `npm run verify:fast:branch`.
-- Never bypass hooks with `--no-verify` for dev work -- it is reserved for
-  CI-generated commits only (e.g. `.github/workflows/regen-visual-baselines.yml`).
-- Cloud `verify.yml` is a required status check on `main` -- the real gate.
+- **Skills resolve from `~/.agents/skills/` only.** Cline does not read a
+  workspace `.claude/skills/` (probed 2026-08-13,
+  `docs/testing/agent-skill-resolution-probe.md`). Re-run `npm run skills:sync`
+  after any change under `.claude/skills/pr-review-gate/`. It is a per-machine
+  step: the target is under `$HOME`, so CI cannot run it and a fresh clone has
+  no mirror.
+- **Model tiers are not selectable.** Subagents run on the configured session
+  model, so CLAUDE.md's "Model routing" table (Haiku / Sonnet / Opus / Fable)
+  and its `.claude/worktrees` + `claude/wt-*` branch choreography are
+  Claude-Code mechanics I cannot reproduce verbatim. Follow the INTENT -- fresh
+  cold subagents per task, per-task review, review-gated ships -- and where a
+  tier choice would genuinely change the outcome, say so and hand the decision
+  to the user.
+- **This is a Windows box running PowerShell 5.1.** No `&&` / `||` chaining
+  (use `;` or `if ($?) { ... }`), no ternary / `??` / `?.`. `head`, `tail`,
+  `which`, and `touch` do not exist -- use `Get-Content -TotalCount N` /
+  `-Tail N` and `(Get-Command x).Source`. A here-string's closing `'@` must sit
+  at column 0, and an em dash inside a `.ps1` is a parse terminator.
 
-## Testing discipline (required, per CLAUDE.md)
-- New behaviour ships a paired automated test. Bug fixes ship a regression test
-  that fails before the fix and passes after. Never delete or `.skip` a test
-  without an explicit replacement. UI cross-seam changes should land an e2e test.
+## Never do these
 
-## Execution model (sub-agent delivery — CLAUDE.md doctrine)
-- All NON-TRIVIAL work is sub-agent-executed: the main thread coordinates,
-  curates context, and judges; it does NOT produce hand-written task code.
-  Only work clearing CLAUDE.md's trivial bar (nothing can break, nothing needs
-  review) stays inline, plus validation (which is the main thread's job).
-- Dispatch FRESH cold subagents per task (spawn_agent / team tools), briefed
-  from the ticket/plan doc as the sole source of requirements — not forks that
-  inherit session context.
-- Per task: an implementer subagent, then a task-review subagent (spec + quality).
-- Non-trivial specs/plans get an adversarial `assumption-checker` pass before
-  the user approves; present its actual tagged output, not a paraphrase.
-- Ship through the `pr-review-gate` / code-review pass; the PR needs
-  `Closes #NN` / `Refs #NN`.
-- CAPABILITY NOTE (honest, Cline-specific): Cline cannot select model tiers for
-  subagents — they run on the configured session model. The repo's
-  Haiku/Sonnet/Opus/`Fable` routing table (CLAUDE.md "Model routing") and its
-  `.claude/worktrees` + `claude/wt-*` branch choreography are Claude-Code
-  mechanics I can't reproduce verbatim. I follow the INTENT (fresh subagent
-  execution + per-task review + review-gated ships), not those letter-level
-  mechanics. Where tier choice would matter, say so and hand the decision to
-  the user.
+- **Never commit to `main`.** Every change -- including a one-line doc fix --
+  reaches `main` through a branch and a PR; `main`'s required status checks
+  reject anything else. Branches are `<type>/<scope>-<slug>`.
+- **Never hand-edit a generated file:** `src/lib/api-types.ts` (regenerate via
+  `npm run openapi:types`), `docs/BACKLOG.md` (`npm run backlog:sync`),
+  `server/.env.example` (`npm run config:sync`).
+- **Never bypass a hook with `--no-verify`.** Triage instead: caused by my
+  change -> fix it in the same commit; pre-existing (the same test fails on
+  `main`) -> surface it to the user and do NOT fold the fix in; suspected flake
+  -> re-run that test alone once and name it. `--no-verify` is reserved for
+  CI-generated commits (`.github/workflows/regen-visual-baselines.yml`).
+- **Never treat hook output in a fresh worktree as proof of verification** --
+  see the first trap below.
+
+## Traps that fail silently
+
+- **A fresh worktree's git hooks do nothing, and say nothing.**
+  `core.hooksPath` is inherited but resolves per-worktree, and `.husky/_` is
+  git-ignored, so a new checkout gives git no hook to run: commit-msg
+  validation and the whole pre-push battery vanish, silently. Fix it before the
+  first commit with `npx husky`, then junction BOTH `node_modules` and
+  `server/node_modules` from the primary checkout (missing the second fails the
+  server legs with "vitest not found"). Simplest path is
+  `node scripts/wt-new.mjs <type>/<scope>-<slug>`, which does all of it plus
+  non-clashing ports.
+- **`.claude/` is git-ignored wholesale** -- `.gitignore` has `.claude/*` with
+  only `!.claude/skills/` re-included -- so a new file under it is silently
+  skipped by `git add` and never reaches the PR. Confirm with
+  `git status --short` after adding one.
+- **Git-ignored artifacts (`brand/`, `mockups/`, marketing captures) are
+  produced in the primary checkout**, never in a worktree: they do not travel
+  with the branch and worktree teardown destroys them.
+
+## Findings are fixed, not filed
+
+A defect or chore surfaced in passing is fixed in the same round, with its
+paired test, and declared in the PR body ("Also fixed, found in passing: ...").
+Filing an issue records the fix; it never substitutes for it. The one thing
+that defers a finding is a genuine design pass -- the fix has more than one
+defensible outcome and something has to choose between them -- and then the
+issue names the decision that is owed. "It would expand this PR", "it's
+pre-existing", "it needs its own test", and "it's only a chore" are all void.
+See CLAUDE.md "Incidental findings: report, fix, record".
+
+## Gates a PR must clear
+
+- Commit subjects follow Conventional Commits. The exact type and scope
+  vocabulary lives in CONTRIBUTING.md "Allowed types" / "Allowed scopes" and is
+  enforced by `scripts/validate-commit-msg.mjs` -- read it there rather than
+  copying the lists around, because they change.
+- `commit-msg` validates the subject. `pre-push` refuses force-push or deletion
+  of `main`, re-validates every pushed subject (so bypassing `commit-msg` buys
+  nothing), then runs `npm run verify:fast:branch` unless the push is
+  docs-only.
+- Run `npm run verify:fast:scoped` before staging and `npm run
+  verify:fast:branch` before pushing, so the hooks pass on the first try.
+- The **PR title** must itself match the commit-subject format
+  (`pr-title-lint.yml`), and the **PR body** must carry a literal `Closes #NN`
+  or `Refs #NN` (`pr-issue-link.yml`). Both are required status checks -- a
+  missing issue link blocks merge outright. File the issue if none exists.
+- A non-trivial spec or plan takes an adversarial `assumption-checker` pass
+  before the user is asked to approve it; present its actual output, not a
+  paraphrase.
+- Cloud `verify.yml` is the real enforcement gate. Every PR that is not
+  docs-only also takes the `pr-review-gate` pass before merge.
+
+## Every change owes
+
+- **A paired automated test.** New behaviour -> a new test. Bug fix -> a
+  regression test that fails before the fix and passes after. Never delete or
+  `.skip` a test without a named replacement. UI changes crossing
+  router/redux/layout seams land a Playwright spec under `e2e/`.
+- **Both release-notes files, in the same PR:** an entry in
+  `docs/release-notes-next.md` and a matching user-facing, brand-voice line in
+  the in-progress section at the top of `RELEASE_NOTES.md`. Skip only when the
+  change has no shippable delta, and say so explicitly rather than omitting it
+  silently.
+- **An on-box acceptance row** when it ships behaviour only real hardware can
+  prove (a live GPU, a real sidecar, a real analyzer, a real book). Recording
+  blocks the merge; running does not. All three surfaces move together -- see
+  CLAUDE.md "Before-shipping checklist" step 3.
 
 ## Working style
-- Surgical changes: fix what's broken, don't restyle what works. Match existing
-  conventions. Keep edits minimal (see CLAUDE.md "Simplicity first").
+
+- Surgical changes: fix what's broken, don't restyle what works. Match the
+  surrounding conventions even where you would write it differently. The seam
+  is defect / chore / taste -- the first two get fixed, the third is never
+  touched.
+- Simplicity first: the minimum code that solves the problem. No speculative
+  abstractions, no unrequested configurability.
+- All non-trivial work is sub-agent-executed: the main thread coordinates,
+  curates context, and judges. Trivial means nothing can break and nothing
+  needs review -- if you would want a review pass on it, it is not trivial.
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -14,12 +14,12 @@ contains zero `CLAUDE` byte sequences under a binary-safe grep, against
 positive controls that do find `clinerules` and `AGENTS.md`; and a live
 `cline -p` probe in this workspace reported only `Workspace AGENTS.md` +
 `.clinerules/cline.md` in context. One caveat, so nobody re-derives this and
-thinks the note is wrong: `cli-windows-x64/bin/cline.exe` does carry a
-`CLAUDE.md` literal and a `claudeMd` / `claudeMdExcludes` **managed-policy**
-schema. That is enterprise-policy plumbing, not the workspace rule loader, and
-it changed nothing on this box -- but a managed config is the one thing that
-could make this stale. Re-probe rather than assume if you are on a managed
-install.
+thinks the note is wrong: `cli-windows-x64/bin/cline.exe` does carry
+`CLAUDE.md` literals, in a `claudeMd` / `claudeMdExcludes` settings schema
+belonging to the **bundled Claude Agent SDK** that Cline vendors -- not to its
+own rule loader. A vendored schema is not a loader, so those strings do not
+indicate `CLAUDE.md` support and no setting flips them into it. Re-verify if
+Cline's rule loader itself changes.
 
 This file is a summary, not a substitute. It deliberately restates almost
 nothing from them, carrying only rules that are **Cline-specific** or that
@@ -28,12 +28,17 @@ this file gets fixed. Anything of general value belongs in `CLAUDE.md` instead.
 
 ## Cline-specific
 
-- **Skills resolve from `~/.agents/skills/` only.** Cline does not read a
-  workspace `.claude/skills/` (probed 2026-08-13,
-  `docs/testing/agent-skill-resolution-probe.md`). Re-run `npm run skills:sync`
-  after any change under `.claude/skills/pr-review-gate/`. It is a per-machine
-  step: the target is under `$HOME`, so CI cannot run it and a fresh clone has
-  no mirror.
+- **Skills resolve from `~/.agents/skills/`; a workspace `.claude/skills/` is
+  not read.** Both proven by probe (2026-08-13,
+  `docs/testing/agent-skill-resolution-probe.md`), as is `~/.cline/skills`
+  being dead. So re-run `npm run skills:sync` after any change under
+  `.claude/skills/pr-review-gate/` -- a per-machine step, since the target is
+  under `$HOME`, so CI cannot run it and a fresh clone has no mirror. Do not
+  read that as an exhaustive list: the loader composes skill roots from its
+  rule directories (`skillsPath: join(<ruleDir>, "skills")`), so workspace
+  roots may also work. They are untested, and presence in the code is not
+  proof -- `~/.cline/skills` sits in that same list and is dead. Treat anything
+  beyond the two probed answers as unverified (#2368).
 - **Model tiers are not selectable, so a Cline review pass does NOT discharge
   the merge gate.** The probe above recorded `CLINE_TIER_SELECTABLE: no`,
   observing a subagent backed by `deepseek-v4-flash`; subagents do start cold
@@ -70,10 +75,13 @@ this file gets fixed. Anything of general value belongs in `CLAUDE.md` instead.
   fails on `main`) -> surface it to the user and do NOT fold the fix in;
   suspected flake -> re-run that test alone once and name it. CLAUDE.md
   "Working practice" states the rule flatly: do not use `--no-verify` to
-  bypass. The single documented exception is a `git commit --no-verify` in a
-  genuine emergency, and that commit still needs fixing before review
-  (CONTRIBUTING.md "Enforcement"). **"I already ran the tests" is not a
-  reason to skip pre-push**: that hook also runs `guard-protected-push` and
+  bypass. Two narrow exceptions are documented, neither of them "the hook is
+  red": a `git commit --no-verify` in a genuine emergency, where the commit
+  still needs fixing before review (CONTRIBUTING.md "Enforcement"), and a
+  `git push --no-verify` when the force-push or branch deletion
+  `guard-protected-push` is refusing is the thing you actually intend
+  (CLAUDE.md "Commit gate", `.husky/pre-push`). **"I already ran the tests" is
+  not a third**: that hook also runs `guard-protected-push` and
   `guard-commit-subjects`, and `verify:fast:branch` runs neither, so bypassing
   drops the backstop that catches a malformed subject when `commit-msg` did
   not fire -- exactly the fresh-worktree case below, and the leak that shipped
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -1,10 +1,19 @@
 # Cline project rules (Castwright / Audiobook-Generator)
 
-Loaded by Cline each session. The canonical context is the root `CLAUDE.md`
-and `CONTRIBUTING.md` -- Cline reads both, so nothing here restates them. This
-file carries only two kinds of rule: ones that are **Cline-specific**, and ones
-that are **costly if missed** and easy to lose in 140 KB of prose. Keep it that
-way -- anything of general value belongs in `CLAUDE.md` instead.
+## Read these first -- they are NOT loaded for you
+
+The canonical rules are the root **`CLAUDE.md`** and **`CONTRIBUTING.md`**.
+Cline loads neither: its rule loader reads this file, `.clinerules/*.md`, and
+workspace/global `AGENTS.md`, and nothing else (verified 2026-08-14 against the
+installed `@cline/core` loader and a live `cline -p` probe). **Read both files
+before your first edit** -- `CLAUDE.md` for the working principles, execution
+model and conventions, `CONTRIBUTING.md` for the commit / branch / PR / issue /
+release-notes specs.
+
+This file is a summary, not a substitute. It deliberately restates almost
+nothing from them, carrying only rules that are **Cline-specific** or that
+**fail silently**. Where the two disagree, they win -- and tell the user, so
+this file gets fixed. Anything of general value belongs in `CLAUDE.md` instead.
 
 ## Cline-specific
 
@@ -14,13 +23,19 @@ way -- anything of general value belongs in `CLAUDE.md` instead.
   after any change under `.claude/skills/pr-review-gate/`. It is a per-machine
   step: the target is under `$HOME`, so CI cannot run it and a fresh clone has
   no mirror.
-- **Model tiers are not selectable.** Subagents run on the configured session
-  model, so CLAUDE.md's "Model routing" table (Haiku / Sonnet / Opus / Fable)
-  and its `.claude/worktrees` + `claude/wt-*` branch choreography are
-  Claude-Code mechanics I cannot reproduce verbatim. Follow the INTENT -- fresh
-  cold subagents per task, per-task review, review-gated ships -- and where a
-  tier choice would genuinely change the outcome, say so and hand the decision
-  to the user.
+- **Model tiers are not selectable, so a Cline review pass does NOT discharge
+  the merge gate.** The probe above recorded `CLINE_TIER_SELECTABLE: no`,
+  observing a subagent backed by `deepseek-v4-flash`; subagents do start cold
+  (`CLINE_SUBAGENT_COLD: yes`, self-reported, not independently reproduced).
+  Independence is the property people check and it is the one Cline has -- the
+  tier is the one it lacks. So CLAUDE.md's "Model routing" table (Haiku /
+  Sonnet / Opus / Fable) and its `.claude/worktrees` + `claude/wt-*` branch
+  choreography are Claude-Code mechanics I cannot reproduce. Follow the INTENT
+  -- fresh cold subagents per task, per-task review, review-gated ships -- but
+  record a Cline pass as a **flash-tier independent pass, never as the
+  `pr-review-gate` gate itself**, and hand any decision that turns on tier
+  choice to the user. Calling it "the review gate" is a false completion
+  claim.
 - **This is a Windows box running PowerShell 5.1.** No `&&` / `||` chaining
   (use `;` or `if ($?) { ... }`), no ternary / `??` / `?.`. `head`, `tail`,
   `which`, and `touch` do not exist -- use `Get-Content -TotalCount N` /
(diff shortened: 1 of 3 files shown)


dudarenok-maker:
## PR review — pass 1 (head 4ff5f44d, effort low)

Scope: `git diff main...HEAD` (`4840c5ea..4ff5f44d`) — `AGENTS.md`, `.clinerules/cline.md` · verified: every factual assertion in both files against the code/config it names (`package.json` scripts, `.husky/*`, `scripts/{verify-cache,is-docs-only-push,guard-commit-subjects,guard-protected-push,sync-env-example}`, `.gitignore`, `.github/workflows/*`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/testing/agent-skill-resolution-probe.md`); ran `npm run verify:fast:scoped` at this head with an empty index; ran a live headless Cline probe against this workspace; grepped the installed `@cline/*` packages; checked the PR body's empirical claims and issue dudarenok-maker/Castwright#2363's labels. No tracked file modified (`git status --porcelain` empty, HEAD unchanged).

### 🔴 Blocking — the premise the whole design rests on is false: Cline loads **neither** `CLAUDE.md` **nor** `CONTRIBUTING.md`

`.clinerules/cline.md:3-4` — *"The canonical context is the root `CLAUDE.md` and `CONTRIBUTING.md` -- Cline reads both, so nothing here restates them."*
`AGENTS.md:10-12` — *"**Cline:** reads `CLAUDE.md` and `CONTRIBUTING.md`, plus every `.clinerules/*.md`."*
PR body, folded finding 1 — *"redundant for Cline (which loads `CLAUDE.md` natively)"*.

Both are wrong. Evidence, strongest first:

1. **Live probe**, same method as `docs/testing/agent-skill-resolution-probe.md` (`cline -p -c "C:\Claude\Projects\Audiobook-Generator" ...`, run at this head with both files on disk), asking it to enumerate what was auto-loaded and answer YES/NO on each:
   > *"Files whose content I can see in my context: `Workspace AGENTS.md` … `.clinerules/cline.md` … 1. Was CLAUDE.md automatically loaded? **No.** 2. Was CONTRIBUTING.md automatically loaded? **No.** … They are only referenced by name."*
2. **Binary corroboration** (`$APPDATA/npm/node_modules/cline/node_modules/@cline/`): the rule loader in `core/dist/index.js` discovers `.clinerules`, `.cline/`, and workspace + global `AGENTS.md` — and contains **zero** occurrences of the string `CLAUDE` (`grep -c CLAUDE core/dist/index.js` → `0`). `CONTRIBUTING.md` appears in **no** `@cline/*` package at all (only inside the bundled `ai` SDK's own docs). The four `CLAUDE.md` hits in `cline.exe` are a managed-policy settings schema (`claudeMd: "…Only honored from managed/policy settings"`), not workspace rule discovery.

Concrete failure: a Cline session's entire project context is `AGENTS.md` + `.clinerules/cline.md`. The first line it reads asserts that CLAUDE.md and CONTRIBUTING.md are already loaded, so it will not go and read them — and neither file ever tells it to. `AGENTS.md`'s only "read it" instructions are addressed to *Claude Code* (line 8) and to *other tools* (line 13). The agent therefore ships without ever seeing: the mandatory-worktree rule, the before-shipping checklist, the execution model / trivial bar, the `cast.json` lock rules, the mobile-viewport + 44 px touch-target table, the plan/`docs/features` obligation, or the allowed commit types and scopes that `.clinerules/cline.md:78-81` deliberately declined to restate on the grounds that Cline could look them up in a file it never loads.

Worse, the one remedy the PR *does* describe is withheld from exactly the tool that needs it: `AGENTS.md:13-17` says a tool that reads this file but not `CLAUDE.md` should import it explicitly, then declines to add the import "unconditionally" because "any tool that already loads `CLAUDE.md` natively would then pull the same ~90 KB into context twice". Cline reads `AGENTS.md` and does *not* load `CLAUDE.md` — it is the tool the sentence describes, and the correction is being withheld on the strength of the false premise. (Claude Code, meanwhile, does not read `AGENTS.md` at all — I can confirm from my own context in this repo — so the duplication being guarded against has no current instance.) Recurring shape dudarenok-maker/Castwright#7: a fix unreachable at the default configuration, plus dudarenok-maker/Castwright#9: the defect is in the document.

This is not fixable by softening the wording. Either the two files must be self-sufficient for Cline, or `AGENTS.md`/`.clinerules/cline.md` must carry an explicit, unconditional instruction to read `CLAUDE.md` and `CONTRIBUTING.md` at session start.

### 🟠 Significant — "Run `npm run verify:fast:scoped` before staging" is a check that cannot fail

`.clinerules/cline.md:86-87` — *"Run `npm run verify:fast:scoped` before staging and `npm run verify:fast:branch` before pushing, so the hooks pass on the first try."*

`verify:fast:scoped` passes `--scope-staged`; `scripts/verify-cache.mjs:745` (`stagedDiffFiles`) runs `git diff --cached --name-only`. With nothing staged git *succeeds* and returns an empty list — which is `[]`, not `null`, so the scope filter stays armed (only `null` disables it, `scripts/verify-cache.mjs:1103-1106`) and `stepTouchedByDiff` (`:722`) returns false for every step.

Run at this head with an empty index:

```
$ git diff --cached --name-only | wc -l
0
$ npm run verify:fast:scoped
[skip] test:hooks (out of scope)
[skip] check:budget-poll (out of scope)
[skip] test (out of scope)
[skip] test:server (out of scope)
EXIT:0
```

An agent following the sentence literally gets a green in ~2 s with zero tests executed, stages, commits — and pre-commit is the first thing that ever runs them, which is precisely the outcome the "so the hooks pass on the first try" clause promises to avoid. Recurring shape dudarenok-maker/Castwright#6, success reported while doing nothing. The correct order is stage first, then run it.

Same instrument weakness in the PR body: *"Confirmed empirically: every leg of `verify:fast:scoped` **and** `verify:fast:branch` reports `(out of scope)` for this diff"* — the `verify:fast:scoped` half of that observation is what an empty index produces for *any* diff, so it is not evidence for this one. The `verify:fast:branch` half (branch-vs-`main` diff) does carry the claim, and I confirmed independently that no step's globs or `extraFiles` match `AGENTS.md` or `.clinerules/**` (`test:hooks`'s `extraFiles` lists `CLAUDE.md`, `scripts/verify-cache.mjs:258`, and nothing else here).

### 🟠 Significant — "Never hand-edit `server/.env.example`" is false for 65% of that file

`.clinerules/cline.md:35-37` — *"Never hand-edit a generated file: … `server/.env.example` (`npm run config:sync`)."*

`server/scripts/sync-env-example.ts:10-17` rewrites **only** the region between `# >>> BEGIN generated config knobs` and `# <<< END generated config knobs` — lines 453–695 of a 706-line file. Everything else is hand-authored, including the `GEMINI_API_KEY=` entry at `server/.env.example:39`. `config:check` (the `verify:fast:branch` leg) compares the same block only.

Concrete failure: an agent adds a non-registry env var, or fixes a stale sentence in the hand-authored header. Reading this line it refuses to touch the file and runs `npm run config:sync` instead, which prints `.env.example synced ✓` and writes nothing relevant — the documentation line never lands, and `config:check` stays green because it never looks outside the block. The claim also silently over-rides `CLAUDE.md`'s own knob rule, which requires a `.env.example` line as a *deliverable*.

`src/lib/api-types.ts` and `docs/BACKLOG.md` in the same bullet are wholly generated and correct; only the third entry is wrong.

### 🟠 Significant — the worktree-junction instruction ships without the teardown warning that makes it safe

`.clinerules/cline.md:52-56` tells the agent to junction **both** `node_modules` and `server/node_modules` from the primary checkout, and stops there. `CLAUDE.md:1030` ("Worktree teardown") exists because the obvious next step destroys the primary checkout: *"`git worktree remove` / `Remove-Item -Recurse` can follow a junction and delete the **real** `node_modules` in the primary checkout."*

Concrete failure: agent creates a worktree, junctions per line 52, finishes the branch, runs `git worktree remove <path>` (or `Remove-Item -Recurse -Force`), and the primary checkout loses both `node_modules` trees — including the husky activation in `.husky/_`, so the primary checkout is then *also* running ungated until the next `npm install`. Given the 🔴 above, Cline never sees `CLAUDE.md:1030`, so this file is its only source on the subject. This belongs in the same "Traps that fail silently" section that already carries the junction advice: drop the reparse point first (`(Get-Item $j -Force).Delete()`, gated on the `ReparsePoint` attribute — **not** `.LinkTarget`, empty on PS 5.1), then remove the worktree.

### 🟡 Minor

1. `.clinerules/cline.md:38` + `:41-42` — *"`--no-verify` is reserved for CI-generated commits (`.github/workflows/regen-visual-baselines.yml`)"* is stricter than, and contradicts, the sources this file defers to: `CONTRIBUTING.md:329` ("To bypass the hook in a genuine emergency, use `git commit --no-verify`"), `CLAUDE.md:871`, and the "intentionally bypassable with `git push --no-verify`" comments in `scripts/guard-protected-push.mjs:10`, `scripts/guard-commit-subjects.mjs:16`, and `scripts/is-docs-only-push.mjs:18`. Failure mode is benign (an agent stalls and asks rather than doing damage), but it is a silent divergence from a canonical file in a PR whose stated design is not to diverge.
2. `.clinerules/cline.md:17-18` — *"Subagents run on the configured session model"* is not what was measured. `docs/testing/agent-skill-resolution-probe.md` records only that neither spawn tool takes a model parameter and that stderr disclosed `deepseek-v4-flash`; nothing established that the subagent inherits the session model. The distinction matters downstream: `.claude/skills/pr-review-gate/SKILL.md:104-113` requires a Cline-run gate pass to be recorded as **flash-tier**, and an agent that believes its subagent runs on its own (strong) session model will record the pass as tier-equivalent. Say what the probe found instead.
3. Issue dudarenok-maker/Castwright#2363 carries only `area:ops` — no `type:chore`. `CONTRIBUTING.md:393-398`: the Backlog view filters `type:feature`, the "Bugs & Chores" view filters `bug OR type:chore`, so dudarenok-maker/Castwright#2363 currently renders in **neither** board view and `backlog:sync` will not surface it either. Add `type:chore`.

### ✅ What is solid

Everything else I checked held up, and several claims are sharper than they needed to be:

- `pre-push` behaviour (`:82-85`) is exactly right, including the non-obvious "re-validates every pushed subject, so bypassing `commit-msg` buys nothing" — `.husky/pre-push` does run `guard-commit-subjects.mjs` after `guard-protected-push.mjs`, and the docs-only skip is third.
- The docs-only glob set quoted in the PR body (`docs/**`, root `*.md`, `.github/*.md`) matches `scripts/is-docs-only-push.mjs:29-31` exactly, and the derived observation that an `AGENTS.md`-only push skips the local battery while a `.clinerules`-only push does not is correct.
- `.gitignore` claim (`:57-58`) is verbatim correct (`.claude/*` + `!.claude/skills/`), as is the fresh-worktree hooks trap, `npx husky`, the both-junctions requirement, and `scripts/wt-new.mjs`.
- The skills claim (`:11-16`) matches `docs/testing/agent-skill-resolution-probe.md` including the per-machine `$HOME` caveat; `npm run skills:sync` exists.
- Every script name cited exists in `package.json` with the behaviour described; `pr-title-lint.yml` / `pr-issue-link.yml` are real and body-validated as described; `.claude/worktrees` is real (`CONTRIBUTING.md:224`).
- PR body's `Co-Authored-By` figure checks out (I measure 30 of the last 100 at this head vs. the stated 29 — measured one commit earlier, and the conclusion is unaffected). "The current Claude Code build does not load `AGENTS.md`" also checks out — this session has `CLAUDE.md` in context and not `AGENTS.md`.
- The three not-applicable calls are correct as far as they go: no runtime surface (paired test), no user/operator-visible delta (release notes), no hardware-provable behaviour (on-box). `Closes dudarenok-maker/Castwright#2363` is present and outside any code span.

### Verdict

**Request changes.** One blocking correctness defect (the Cline-reads-CLAUDE.md premise, which inverts the file's entire scoping decision), three significant ones, three minor. All are in the documents rather than in code, which is the whole risk surface of this PR: each is an instruction another agent will act on. Re-review is triggered once the blocking + significant findings are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head f764e4ff, effort low)

Scope: `git diff 4ff5f44d..f764e4ff` — `AGENTS.md`, `.clinerules/cline.md` — plus the PR body and issue dudarenok-maker/Castwright#2363 · verified: the blocking premise re-derived independently against the installed `@cline` bundle with positive controls (`grep -rl clinerules|AGENTS.md|CLAUDE.md|CONTRIBUTING` per package: `core` → `clinerules=3, AGENTS.md=3, CLAUDE.md=0, CONTRIBUTING=0`; binary-safe `grep -rao 'CLAUDE[A-Za-z_.-]*' @cline/core` → **zero hits**; the rule-discovery function extracted from `@cline/core/dist/hub/daemon/entry.js`); `server/.env.example` line/marker positions and `server/scripts/sync-env-example.ts --check` semantics; `CONTRIBUTING.md` "Enforcement" (`:322-331`); `CLAUDE.md` `:860-880` + `:1182-1185`; `.husky/pre-push` + `.husky/_` on disk + `git config core.hooksPath`; `scripts/guard-{protected-push,commit-subjects}.mjs`; `docs/testing/agent-skill-resolution-probe.md:47,63-64`; `.claude/skills/pr-review-gate/SKILL.md:95-125`; every cross-referenced heading in both target files; `.gitignore:33-34`; non-ASCII scan of both files (clean); `gh issue view 2363`. No tracked file modified (HEAD `f764e4ff`, `git status --porcelain` empty).

**The blocking premise checks out.** Independently: `@cline/core` — the package containing the rule loader — contains no `CLAUDE` byte sequence at all, while the same greps find `clinerules` and `AGENTS.md` in three files each, so a wrong-path negative is ruled out. `CONTRIBUTING` appears nowhere in `@cline/*` except a vendored `vercel/ai` README. The rule discovery function reads `{type:"rule", directories:[...defaultRuleDirs, workspace/".cline"], includeFile:(f)=>f===".clinerules"||isMd(f)||isMd(parent)}`, with `AGENTS.md` appended per scanned directory. Both files' corrected claim is sound, and the fix is the right shape: an explicit read-these-first instruction in both, rather than a `@CLAUDE.md` import whose expansion was never verified.

- 🔴 **Cline-loads-`CLAUDE.md` premise** — **RESOLVED in both files** (`.clinerules/cline.md:3-11`, `AGENTS.md:8-18`), claim re-verified independently as above. **STILL OPEN in the PR body**, which is the third surface pass 1 cited by quotation — see New finding 1.
- 🟠 **`verify:fast:scoped` pre-flight cannot fail** — **VERIFIED RESOLVED** in the file (`.clinerules/cline.md:119-124`); the replacement draws the right line (`verify:fast:branch` *is* non-vacuous for its stated purpose — it is the identical command pre-push runs, with identical scoping, so a green genuinely predicts the hook). The PR body's use of the vacuous run as evidence is **STILL OPEN** — see New finding 1.
- 🟠 **"Never hand-edit `server/.env.example`"** — **VERIFIED RESOLVED** (`.clinerules/cline.md:50-56`). Every stated fact holds: markers at `server/.env.example:453` / `:695` of 706 lines → 463 lines outside (file says "~460"), `GEMINI_API_KEY=` at `:39`, and `sync-env-example.ts:13-27` builds `next = src.slice(0,b) + block + src.slice(e+END.length)` then `--check`s `src !== next`, so text outside the block is copied verbatim into the comparand and can never be reported as drift — "`config:check` never looks outside the block" is exactly right.
- 🟠 **Worktree-teardown warning missing** — **VERIFIED RESOLVED** (`.clinerules/cline.md:80-89`): drop-junctions-first, `(Get-Item $j -Force).Delete()`, the `ReparsePoint` attribute gate, the `.LinkTarget`-is-empty-on-PS-5.1 trap, and `cmd /c rmdir` no-op — all match `CLAUDE.md:1030-1043`, and the "Full recipe" pointer resolves. One sentence of its *rationale* is false — New finding 3.
- 🟡 **`--no-verify` stricter than the sources** — **REPLACED, and the replacement is wrong in the other direction.** See New finding 2.
- 🟡 **Model-tier text overstated** — **VERIFIED RESOLVED** (`.clinerules/cline.md:26-38`). `CLINE_TIER_SELECTABLE: no`, `deepseek-v4-flash`, and `CLINE_SUBAGENT_COLD: yes (self-reported)` are verbatim at `docs/testing/agent-skill-resolution-probe.md:63-64,47`, and the recorded consequence matches `.claude/skills/pr-review-gate/SKILL.md:103-121` ("flash-tier independent pass, never as this gate"; "a summary that calls it 'the review gate' is a false completion claim").
- 🟡 **#2363 missing `type:chore`** — **VERIFIED RESOLVED**: labels are now `["area:ops","type:chore"]`, so it renders in the board's "Bugs & Chores" view and correctly stays out of `docs/BACKLOG.md`.

### New findings

**1. 🟠 correctness — the PR body still asserts the premise this commit falsified, and still cites the vacuous run as evidence** (`updated_at 2026-08-14T07:40:34Z`, body unchanged from the `4ff5f44d` round)

Three live claims in the body as it stands:

- Summary bullet 2: *"Cline reads `CLAUDE.md` and `CONTRIBUTING.md` already, so this file deliberately restates neither."* — the exact sentence `f764e4ff` exists to delete.
- Folded finding 1: *"redundant for Cline (which loads `CLAUDE.md` natively)"* — quoted by name in the pass-1 blocking finding.
- Not-applicable → paired test: *"Confirmed empirically: every leg of `verify:fast:scoped` **and** `verify:fast:branch` reports `(out of scope)`"* — the `scoped` half is what an empty index produces for *any* diff, which is the observation pass 1 disqualified.

The body also has no record of this round at all: its only fold section is headed "Review findings folded in (commit `4ff5f44d`)", so the blocking fix and the four others land undeclared, against `CLAUDE.md` "Incidental findings" ("declared in the PR body — an unannounced fix reads as scope creep").

Concrete failure: merges in this repo use a merge commit whose body carries the PR body, so `git log` becomes the durable record of a claim the same PR proved false — and the next agent to touch `.clinerules/` reads "Cline reads CLAUDE.md already" from the history and re-derives the deleted scoping decision. Shape dudarenok-maker/Castwright#9: the defect is in the document, not the code. Fixable without a commit (edit the body), but it must be fixed before merge.

**2. 🟠 correctness — the new `--no-verify` carve-out 2 is not what `CLAUDE.md` "Commit gate" says, and it authorises skipping two guards the "satisfied gate" never ran** (`.clinerules/cline.md:61-64`)

> *"Bypassing is legitimate in two narrow cases only: a genuine emergency, where the commit still needs fixing before review (CONTRIBUTING.md "Enforcement"), and **deliberately skipping the pre-push hook on a push whose gate you have already satisfied** (CLAUDE.md "Commit gate")."*

Carve-out 1 is accurate (`CONTRIBUTING.md:329-331` verbatim). Carve-out 2 is not in its cited source. `CLAUDE.md:871`'s "bypass the local hook intentionally with `git push --no-verify`" sits *inside the parenthetical about `guard-protected-push.mjs`*, whose own header states the scope: *"intentionally bypassable … for the rare case where a force-push/deletion to main is genuinely intended"* (`scripts/guard-protected-push.mjs:10-11`; same wording at `guard-commit-subjects.mjs:16-17`, `is-docs-only-push.mjs:18`). Nothing anywhere authorises bypass on the grounds that the battery was already run — and `CLAUDE.md:1182` says flatly *"Do not use `--no-verify` to bypass."*

Concrete failure: an agent runs `npm run verify:fast:branch` by hand (which this same file tells it to do, `:119`), reads line 63, and pushes with `--no-verify`. `.husky/pre-push:22-27` runs three things *before* the battery, and `git push --no-verify` skips all of them: `guard-protected-push` (force-push/deletion of `main`) and `guard-commit-subjects`, which per `.husky/pre-push:5-8` *"reject[s] any pushed commit whose subject violates the Conventional-Commits rule, even if commit-msg was bypassed … Backstops the @-leak that shipped on dudarenok-maker/Castwright#856."* `verify:fast:branch` runs neither, so "the gate you have already satisfied" does not cover the guards being skipped — a bad subject or an `@`-leak sails through on a bypass the file just called legitimate. Pass 1's minor asked the file to stop being *stricter* than its sources; this over-corrects past them. State the actual case (a deliberately-intended force-push or branch deletion), or drop carve-out 2.

**3. 🟡 cleanup — "taking husky activation with it" is false, and mis-files a loud failure as a silent one** (`.clinerules/cline.md:81-83`)

> *"`git worktree remove` (and `Remove-Item -Recurse`) will follow a junction and delete the REAL `node_modules` it points at, **taking husky activation with it**."*

Husky activation is `core.hooksPath=.husky/_` plus `.husky/_/` — verified on this box as a **real directory at the repo root**, not a reparse point (`Attributes: ReadOnly, Directory, Archive`) and not inside `node_modules`. Deleting the primary checkout's `node_modules` leaves both intact: the hooks still fire, then fail non-zero (`.husky/pre-commit` → `npm run verify:fast:scoped` with no deps), which blocks the commit loudly. So the consequence is the opposite of what the sentence says, in a section titled "Traps that fail silently" — and `CLAUDE.md` makes the same distinction from the other side ("Junctioning `node_modules` does **not** activate hooks. `npm install` does"). The actionable advice (drop the junctions first) is correct and unaffected; only the rationale is wrong. Failure scenario is mild: an agent that has just destroyed the primary checkout's `node_modules` believes it is now committing ungated and may go looking for a phantom hook problem instead of the missing dependency tree.

**4. 🟡 cleanup — "and nothing else" overstates the loader's discovery set** (`.clinerules/cline.md:6-8`)

> *"its rule loader reads this file, `.clinerules/*.md`, and workspace/global `AGENTS.md`, **and nothing else**"*

The extracted loader (`@cline/core/dist/hub/daemon/entry.js`) is wider than that on three axes: `directories` is `[...defaultRuleDirs(workspacePath), <workspace>/.cline]`, so a **global** rules directory and a workspace **`.cline/`** tree are both scanned; `includeFile` accepts `.clinerules` as a bare **file** as well as a directory; and the extension set is `new Set([".md",".markdown",".txt"])`, not `.md` alone. The load-bearing half of the sentence — that `CLAUDE.md` and `CONTRIBUTING.md` are not among them — is verified and unaffected; it is the exhaustive framing that is wrong, and an exhaustive claim is the one shape that cannot be safely approximated. Trim to "does not read `CLAUDE.md` or `CONTRIBUTING.md`" and drop "and nothing else", or enumerate accurately.

*Note, not a finding:* `f764e4ff`'s commit message asserts *"every CLAUDE string in the bundle is a Claude Agent SDK env var … or the `CLAUDE_CODE` provider enum"*. There are also four literal `CLAUDE.md` strings, all in `@cline/cli-windows-x64/bin/cline.exe`, forming a managed-policy settings schema (`claudeMd: "CLAUDE.md-style instructions injected as organization-managed memory. Only honored from managed/policy settings."` + `claudeMdExcludePaths`). Pass 1 identified these correctly; the universal in the commit message does not cover them. The conclusion is unaffected — they are admin-supplied memory content and an exclude-glob list, not workspace `CLAUDE.md` discovery — and the message is immutable, so this is recorded rather than raised.

### Verdict

**Request changes.** Six of seven pass-1 findings are verified resolved in the files, including the blocking one, whose corrected premise I re-derived independently with positive controls. Two correctness-class items remain: the PR body still carries the falsified premise and the disqualified evidence (fixable without a commit), and the `--no-verify` carve-out over-corrected into authorising a bypass that skips the subject-validation and `@`-leak guards `verify:fast:branch` does not run. Two cleanup nits. Re-review is triggered once the 🟠 items are addressed and pushed.

dudarenok-maker:
## PR review — pass 3 (head cf43b275, effort low)

Scope: the two pass-2 🟠 and two 🟡 against the current file contents + the sources they cite, everything `f764e4ff..cf43b275` touched, and the rewritten PR body · verified: the loader re-derived from `@cline/core/dist/index.js` (rule descriptor `h3`, `discoverFiles` `f5`, `includeFile`, the `new Set([".md",".markdown",".txt"])` at byte ~126790) and its search-path source `resolveRulesConfigSearchPaths` in `@cline/shared/dist/storage/index.js`; per-package greps with positive controls (`@cline/core` → `CLAUDE=0`, `AGENTS.md=4 files`, `clinerules=3 files`; `@cline/shared` → `CLAUDE=0`, `AGENTS.md=3`, `clinerules=2`; `CONTRIBUTING` present nowhere in `@cline/*` outside the vendored `ai` package); every `CLAUDE.md` / `claudeMd*` byte offset in `cline.exe` and the 520-byte schema span they sit in; `CLAUDE.md:871` + `:1182`; `CONTRIBUTING.md:322-331`; `.husky/pre-push:1-28`; `package.json:75-77`; `.husky/_` attributes + `core.hooksPath` + `.gitignore:87`; `server/.env.example` markers (453 / 695 of 706) and `GEMINI_API_KEY=` at `:39`; every step's `globs`/`extraFiles` in `scripts/verify-cache.mjs`; `scripts/is-docs-only-push.mjs:28-32`; `docs/testing/agent-skill-resolution-probe.md`; `.claude/skills/pr-review-gate/SKILL.md:100-122`; every heading both files cross-reference by name; `gh pr checks 2364`. No tracked file modified (HEAD `cf43b2752a4ebbfc4fa497726d84da8e9bacf22f`, `git status --porcelain` empty).

- 🟠 **PR body still carrying the falsified premise + the vacuous run as evidence** — **VERIFIED RESOLVED.** Summary bullet 2 now states the opposite ("Opens by telling the session that `CLAUDE.md` and `CONTRIBUTING.md` are **not** loaded for it"); the "redundant for Cline (which loads `CLAUDE.md` natively)" sentence is gone; `verify:fast:scoped` is recorded as *ran*, not as evidence; and the new "Review record" section declares both rounds' folds, so the merge-commit body no longer becomes a durable record of a disproved claim.
- 🟠 **`--no-verify` carve-out 2 unsupported by its cited source** — **VERIFIED RESOLVED.** Carve-out 2 is gone, and every claim in the replacement holds: `CLAUDE.md:1182` is verbatim "**Do not use `--no-verify` to bypass.**"; `CONTRIBUTING.md:329-331` is the emergency `git commit --no-verify` + "will need fixing before any PR is reviewed"; and the load-bearing new clause is exactly right — `.husky/pre-push:23-24` runs `guard-protected-push.mjs` then `guard-commit-subjects.mjs` before the battery, while `verify:fast:branch`'s step list (`package.json:77`) is `lint,typecheck,config:check,test:hooks,check:budget-poll,test,test:server,build,test:sidecar` and contains neither; the `#856` @-leak attribution matches `.husky/pre-push:5-8`. One residual over-correction in the *other* direction — new finding 1, cleanup.
- 🟡 **"taking husky activation with it" false** — **VERIFIED RESOLVED, and the replacement parenthetical is accurate.** `.husky/_` on this box is `Attributes: ReadOnly, Directory, Archive` with an empty `Target` (a real repo-root directory, not a reparse point), git-ignored at `.gitignore:87`, and `core.hooksPath=.husky/_` lives in `.git/config` — so deleting the primary checkout's `node_modules` leaves the hooks firing and failing non-zero, which is what the new sentence now says.
- 🟡 **"and nothing else" overstating the loader** — **VERIFIED RESOLVED**, and the replacement enumeration is correct on all four axes: `.clinerules` as a bare **file** (`f5`'s `isFile()` branch plus `includeFile: (J,W)=>J===".clinerules"||R5(J)||R5(W)`), `.cline/` (`resolveRulesConfigSearchPaths` returns `<ws>/.cline/rules`, and `h3` appends `<ws>/.cline`, whose `managed.json`-gated subdirs are read for `rules.md`), workspace **and** global `AGENTS.md` (`<ws>/AGENTS.md` and `~/.agents/AGENTS.md`), and `.md`/`.markdown`/`.txt`. Dropping the exhaustive framing was the right call rather than the cosmetic one — the real set also includes `~/.cline/rules` and `~/Documents/Cline/Rules`, which the new sentence correctly no longer purports to exclude.

**Cross-checks the fixes invited, all clean.** The zero-`CLAUDE` evidence has no gap at the package boundary: the search-path list actually lives in `@cline/shared/storage`, not `@cline/core`, and `@cline/shared` is *also* zero-`CLAUDE` under the same binary-safe grep with `AGENTS.md`/`clinerules` as positive controls. The `cline.exe` caveat's load-bearing half is independently confirmed: all four `CLAUDE.md` literals sit inside one 520-byte span (offsets 133380459–133380964) that is entirely zod `.describe()` text, and the binary contains no `Project memory` / `memoryType` / `TodoWrite` string — the settings *schema* is bundled, the memory *loader* is not. Every cross-referenced heading resolves (`CONTRIBUTING.md` `### Enforcement` `:322`, `### Allowed types` `:262`, `### Allowed scopes` `:276`; `CLAUDE.md` `### Worktree teardown` `:1030`, `## Model routing`, `## Before-shipping checklist`, `### Incidental findings: report, fix, record`; `Working practice` is a list label at `CLAUDE.md:1172` rather than a heading, but is findable by the quoted name). Body claims spot-checked and true: `CLAUDE.md:1182`/`:871`, `453–695 of 706` (463 lines outside → "~460"), `GEMINI_API_KEY=` at `:39`, `.clinerules/cline.md` failing `isDocsOnlyFile` (`scripts/is-docs-only-push.mjs:28-32`), no step in `scripts/verify-cache.mjs` matching `AGENTS.md` or `.clinerules/**` (only `CLAUDE.md`, at `:258`), and CI green at this head.

### New findings

All three are cleanup, and all three are in prose whose *actionable* half is correct. None is correctness-class; none should trigger another round.

**1. 🟡 cleanup — "The single documented exception" is one short, in the direction pass 1 already flagged once** (`.clinerules/cline.md:73`)

> *"The single documented exception is a `git commit --no-verify` in a genuine emergency"*

There is a second documented one: `CLAUDE.md:871` says "bypass the local hook intentionally with `git push --no-verify`" for a deliberately-intended force-push/deletion of `main`, and the same wording appears at `.husky/pre-push:3` and in the headers of `guard-protected-push.mjs`, `guard-commit-subjects.mjs`, and `is-docs-only-push.mjs`. Concrete failure: the owner tells the session to force-push `main` (or delete a protected ref); the agent reports that no push-side bypass is documented and contradicts `CLAUDE.md:871` to the user's face. Benign in direction — it stalls and asks, which pass 1 rated 🟡 for the same reason — but this is the third round in which the paragraph has been either stricter or looser than its sources, so it is worth naming: the bullet's own headline scope ("to get unstuck") is what makes the sentence true, and a two-word qualifier ("the single documented exception *for a failing gate*") pins it.

**2. 🟡 cleanup — "Skills resolve from `~/.agents/skills/` only" is an exhaustive claim the vendored loader contradicts, and the cited probe never tested the paths it excludes** (`.clinerules/cline.md:31`)

`resolveSkillsConfigSearchPaths` (`@cline/shared/dist/storage/index.js`, imported into `@cline/core/dist/index.js` as the skill descriptor's default `directories`, and present verbatim in `cline.exe` at the byte-123665951 region) returns **five** roots: `<ws>/.clinerules/skills`, `<ws>/.cline/skills`, `<ws>/.agents/skills`, `~/.cline/skills`, `~/.agents/skills`. `docs/testing/agent-skill-resolution-probe.md` only ever contrasted `.claude/skills` with `~/.agents/skills` — it notes its own grep method was "silent on global vs workspace" — so its verdict sentence, *"There is no workspace path to mirror into: the only store Cline reads is outside the repo"*, is untested against three of those five.

Concrete failure: an agent asked to make `pr-review-gate` resolvable to Cline from the repo — i.e. to retire the per-machine `npm run skills:sync` step, or to close the drift guard that fails open because its target is in `$HOME` — reads line 31, concludes no workspace store exists, and never tries `.clinerules/skills/pr-review-gate/SKILL.md`, inside the very directory this PR creates.

Two reasons this is cleanup and not a blocker: the bullet's actionable instruction (run `skills:sync`; Cline does **not** read a workspace `.claude/skills/`) is verified true and safe, and the sentence is inherited verbatim from `CLAUDE.md`'s own claim rather than introduced here. Also, code presence is not behavioural proof — one of the five roots in that same list is `~/.cline/skills`, which a prior probe found dead — so the honest state is *unconfirmed*, not *wrong*. The right disposition is a probe (drop a throwaway `SKILL.md` under `.clinerules/skills/` and ask a `cline -p` session whether it sees it), not a rewrite on the strength of the bundle.

**3. 🟡 cleanup — the `cline.exe` caveat labels one key managed-policy that isn't, and names a staleness mechanism that cannot fire** (`.clinerules/cline.md:18`)

> *"a `claudeMd` / `claudeMdExcludes` **managed-policy** schema … a managed config is the one thing that could make this stale"*

Exact strings in the binary: `claudeMd` — *"CLAUDE.md-style instructions injected as organization-managed memory. **Only honored from managed/policy settings.**"* (policy-only ✓); `claudeMdExcludes` — *"Glob patterns or absolute paths of CLAUDE.md files to exclude from loading … Only applies to User, Project, and Local memory types (Managed/policy files cannot be excluded)"* — **no** managed-only restriction, and its effect is confined to *non*-managed memory. Both sit in the general Claude Agent SDK settings schema cline vendors for its `CLAUDE_CODE` provider (the same span carries `claude ssh`, `pluginTrustMessage`, `theme`, `editorMode`; `claude-agent-sdk` appears 4× in the binary), not in an enterprise-policy subsystem. And the named mechanism is inert in the stated direction: `claudeMd` injects an admin-supplied *string* consumed on the Agent-SDK path, so no managed config can make Cline's own rule loader start discovering a workspace `CLAUDE.md`.

Concrete failure: an agent re-deriving this check — which is precisely the reader the caveat was written for — finds `claudeMdExcludes` documented as applying to User/Project/Local settings, reads that as contradicting "managed-policy", and re-opens a question the paragraph settled. The conclusion ("not the workspace rule loader") is verified and the advice ("re-probe rather than assume") is conservative and correct; only the attribution needs tightening: *a `claudeMd` (policy-only) / `claudeMdExcludes` pair in the bundled Claude Agent SDK settings schema*.

*Note, not a finding:* the body's "Pass 2 — confirmed 5 of 7 resolved" partitions differently from pass 2's own verdict ("Six of seven pass-1 findings are verified resolved in the files") — the body counts the teardown fix's false rationale as one of the two wrong fixes, pass 2 counted it as resolved-with-a-caveat. Both readings are defensible; no action.

### Verdict

**Approve.** All four pass-2 findings verified resolved against the sources they cite, not against the commit message — including both 🟠, and including the two claims most likely to have been reworded rather than fixed (the `--no-verify` guard asymmetry and the `.husky/_` rationale), each of which I re-derived from `.husky/pre-push`, `package.json:77`, and the on-disk attributes. Nothing the fixes introduced is wrong: the rewritten loader enumeration is accurate on all four axes against `@cline/core/dist/index.js`, and the `cline.exe` hedge's load-bearing half (schema, not loader) is independently confirmed. Three 🟡 cleanup notes above, all in prose whose actionable half is correct — none is correctness-class, so no further round is triggered. Reviewed and clean at `cf43b275`.

PR #2365, opened 2026-08-14
fix(server): three defects found running the analyzer on a real Russian book

### LINKED ISSUE #2353: readJson throws on a UTF-8 BOM, so a hand-edited user-settings.json bricks the server at boot
Observed live on 2026-08-14 while standing up an isolated server for the #2306 re-run.

## What happens

`readJson` (`server/src/workspace/state-io.ts:47-50`) does:

```ts
const raw = await readFile(path, "utf8");
return JSON.parse(raw) as T;
```

Node does **not** strip a UTF-8 BOM on read, and `JSON.parse` rejects it. A
`user-settings.json` whose first three bytes are `EF BB BF` therefore throws:

```
[server] FATAL unhandledRejection — SyntaxError: Unexpected token "﻿", "﻿{ ... is not valid JSON
    at readJson (server/src/workspace/state-io.ts:48:15)
    at async readUserSettings (server/src/workspace/user-settings.ts:355:15)
```

It fires three times during a single boot — once from `readUserSettings`, once
from `runUpgradeCoordinator` -> `readLastSeenAppVersion`, once from the sidecar
supervisor's `buildOpts`. The server survives (the rejection handler catches it)
but **every user setting silently falls back to its default**: in the observed
case `autoStartSidecar` was `false` in the file and the sidecar started anyway.
Silent wrong-defaults, not a clean failure.

## Why this is reachable without hand-crafting it

- Windows Notepad writes UTF-8 **with** BOM; this is a Windows-first app and
  `docs/superpowers/specs/2026-08-11-castwright-local-rebind-design.md:122`
  treats "a hand-edited `user-settings.json`" as a supported shape.
- Windows PowerShell 5.1's `Set-Content -Encoding utf8` / `Out-File -Encoding utf8`
  emit a BOM — and `scripts/tests/log-utils.Tests.ps1:27` already documents that
  in this repo. CLAUDE.md's own PowerShell guidance tells you to pass
  `-Encoding utf8` explicitly when writing a file other tools will read.

## Scope

`readJson` has ~227 non-test call sites, so this is every workspace JSON the
server reads: `state.json`, `cast.json`, `manuscript-edits.json`, the analysis
cache, the handoff files. `readJsonWithRecovery` partially masks it for files
that have backups — it treats the BOM as corruption and silently falls back to
`.bak.1`, which is its own quiet wrong-data path.

## Fix

Strip a single leading `﻿` before `JSON.parse`, in `readJson` and in
`readJsonWithRecovery`'s main-file and backup reads. No behaviour decision to
make; a BOM-prefixed UTF-8 JSON document is valid UTF-8 and every other JSON
reader in the ecosystem tolerates it.

## Test

Paired regression test in `server/src/workspace/state-io.test.ts`: write a file
with a `﻿` prefix, assert `readJson` returns the parsed object (fails before
the fix with the `SyntaxError` above). Cover `readJsonWithRecovery` too — assert
it parses the BOM-prefixed **main** file rather than silently recovering from a
backup.

### LINKED ISSUE #2354: Stage-2 time estimate is labelled "based on stage 1 rate" while still holding the pre-flight constant
Observed live on 2026-08-14 during the #2306 re-run. It caused a wrong ETA to be reported to the repo owner, twice.

## What happens

`server/src/routes/analysis.ts:4519` logs:

```ts
log(1, `Estimated stage time: ~${humanSeconds(stage2EstMs)} (based on stage 1 rate)`);
```

That line sits in **Phase 1 setup**, which the comment immediately above it
(line ~4498) says happens **unconditionally** and runs the worker pool
**concurrently with Phase 0**:

> Plan 88 follow-up — Phase 1 setup happens UNCONDITIONALLY here, and the worker
> pool runs concurrently with Phase 0 (in cache-miss / pipelined mode).

But the refinement that would make the label true lives inside **Phase 0's**
completion block, at line ~4452:

```ts
if (stage1ActualMs > 0) {
  stage2EstMs = clampStageEstMs(stage1ActualMs * STAGE2_STRETCH);
```

whose own comment says:

> Moved inside Phase 0 as part of the pipelining refactor ... in pipelined mode
> this can land WHILE Phase 1 chapters are already running concurrently.

So in pipelined mode the logged value is still the **pre-flight constant** from
line 3335:

```ts
let stage2EstMs = clampStageEstMs((sourceChars / STAGE1_BASELINE_RATE) * STAGE2_STRETCH);
```

`sourceChars` times a static baseline — no measured rate involved. Line 3332's
comment is explicit that this is the "static baseline for both stages" and that
the observed-rate replacement happens later.

## Evidence it is not merely theoretical

Two independent fresh-workspace runs of the same book printed the estimate
**byte-identical**:

```
09:42:34 phase=1 Estimated stage time: ~54m 15s (based on stage 1 rate)
10:22:23 phase=1 Estimated stage time: ~54m 15s (based on stage 1 rate)
```

A figure genuinely derived from an observed rate on a contended box cannot repeat
to the second across two runs. It is deterministic because it is
`sourceChars / STAGE1_BASELINE_RATE * STAGE2_STRETCH`.

## Why it matters

This is operator-facing output that states its own provenance, and the stated
provenance is wrong exactly when the number is least trustworthy — before any
measurement exists. It is read as "we have measured this box" when it means "we
guessed from a constant". It is not always false: in sequential (non-pipelined)
mode Phase 0 finishes first, the refinement has run, and the label is accurate.
So the label needs to reflect which of the two it actually is.

## Fix

Track whether the observed-rate refinement has happened (a boolean set alongside
the assignment at ~4452) and choose the suffix from it — `(based on stage 1
rate)` when refined, something like `(pre-flight estimate, refined after stage
1)` when not. Do not hardcode either wording, and do not change the estimate
arithmetic; only the claim about where the number came from.

## Test

`server/src/routes/analysis.test.ts` (or the nearest existing analysis-route
suite): assert the pipelined path logs the pre-flight wording while
`stage1ActualMs` is still 0, and the refined wording after the refinement has
run. The test must fail against the current code, where the suffix is a constant
string.

### LINKED ISSUE #2355: RU_STOPWORDS omits "как", so "Как сказал …" is scanned as a missing tagged speaker and forces a chapter re-detection
Observed live on 2026-08-14 during the #2306 re-run, on Ночной дозор (Russian).

```
2026-08-14 11:47:12 [analysis] phase=0 Chapter 9: stage-1 missed tagged speaker(s) Как — retrying detection (attempt 2).
```

`Как` is not a character. It is the Russian conjunction "as / how".

## Mechanism (verified, not inferred)

- `server/src/analyzer/tag-grammar.ts:37` — `RU_VERBS` contains `сказал` and `сказала`.
- `server/src/analyzer/tag-grammar.ts:49` — `RU_STOPWORDS` contains 22 words and does **not** contain `как`.
- `server/src/analyzer/roster-coverage.ts:220` — a candidate survives only if
  `isStopword(key) || langStops.has(key) || ignore.has(key)` is false.

So `Как` + a speech verb matches the `name-verb` tag order, is absent from the
roster, and lands in `missingSpeakers`, which triggers a full stage-1
re-detection of the chapter.

## Evidence from the book text

Both occurrences in the whole 650,932-char manuscript are the conjunction:

```
…Диск я не использовал никогда. Как сказал маг, в амулете оставался один…
…Чью Книгу Судьбы откроет волшебница? Как сказала Ольга, двенадцать лет…
```

Both are sentence-initial — precisely the class `RU_STOPWORDS`' own
`// name-verb openers (finding J)` block was added for.

Discriminator: **0** occurrences of mid-sentence bare `Как`. A real character
name recurs in non-sentence-initial position; this never does.

Controls, same book, same scan: `Что`, `Так`, `Вдруг` — all three already
stoplisted — match a speech verb **0** times. The list carries three words that
never fire on this book and omits the one that does.

## Cost

A forced re-detection of chapter 9 (67,097 chars) — roughly 15 minutes of local
model time on the box that hit it. `"Как сказал X"` / `"Как сказала X"` is
ordinary Russian, so this will recur on essentially any Russian book.

The existing false-positive bound did not catch it: `tagCount` was 1 for that
chapter, but the text is dialogue-dense so `quoteAdjacent` was true, and
`tagCount < minHitsWithoutQuote && !quoteAdjacent` therefore did not skip it.

## Fix

Add `как` to `RU_STOPWORDS`, in the `name-verb openers (finding J)` group.

Scope note: only `как` is verified here. Whether the other language stoplists
(`ES_STOPWORDS`, `FR_STOPWORDS`, `DE_STOPWORDS`) have equivalent gaps is a
separate question and needs its own corpus evidence — do **not** speculatively
add words to them in this fix.

## Test

In the roster-coverage test suite: assert that Russian prose containing
`Как сказал маг,` yields **no** `missingSpeakers` entry. It must fail before the
fix (the entry is present) and pass after. Keep an existing-behaviour case
asserting a genuine unrostered Russian speaker IS still reported, so the fix
cannot be a blanket suppression.

dudarenok-maker:
## Summary

Three defects found by actually running the analyzer on a real Russian book (the
#2306 re-run), each fixed with a paired regression test. All `server`-scope, all
found in passing rather than looked for, and none reachable from a fixture —
they needed a live run to surface.

**`readJson` dies on a UTF-8 BOM, and every user setting silently reverts to its
default.** A `user-settings.json` whose first three bytes are `EF BB BF` throws
three FATAL unhandledRejections during one boot. The server survives, so the
failure is invisible — but `readUserSettings` returns nothing and every setting
falls back: an observed `autoStartSidecar: false` started the sidecar anyway.
Windows Notepad and PowerShell 5.1's `-Encoding utf8` both write a BOM, and this
repo's own design docs treat a hand-edited settings file as a supported shape.
`readJson` has ~227 non-test call sites; `readJsonWithRecovery` was additionally
treating the BOM as corruption and silently serving a stale `.bak.1`. **Review
found the first fix incomplete**: two callers bypass `readJson` entirely —
`paths.ts:readWorkspaceOverrideFrom` (so a BOM'd file silently lost
`workspaceDirOverride` and the workspace root fell back to the default) and
`cast-fingerprint.ts` (so a BOM'd `cast.json` returned an empty prior cast *and*
a null fingerprint, disabling the compare-and-set guard). Both now strip too;
the fingerprint still hashes the raw, unstripped bytes, so no stored fingerprint
moves.

**The stage-2 time estimate claimed a measured basis it never had.**
`Estimated stage time: ~54m 15s (based on stage 1 rate)` is logged at Phase 1
setup, but the observed-rate refinement lives inside Phase 0's completion block,
so the number is the pre-flight constant
`sourceChars / STAGE1_BASELINE_RATE * STAGE2_STRETCH` — tuned for
`gemini-2.5-flash`, not a measurement of this machine. Proof it is not
theoretical: two independent fresh runs of the same book printed the figure
byte-identical, which a rate-derived number cannot do.

**The first attempt at this fix did not work, and review caught it.** It made
the wording conditional on a `stage2EstRefined` flag — but that flag is set
inside `runPhase0Pool()`, which is launched **un-awaited**, with no `await`
between there and the log. It is therefore `false` on every run, in sequential
mode too (sequential parks Phase 1's *workers*, not its *setup*), so the
"refined" branch was unreachable dead code. The flag, its setter and the helper
are now deleted and the log states the pre-flight wording **unconditionally**,
which is what is actually true at that point. `stage2EstMs`'s own refinement is
untouched and still feeds the per-chapter budgets, the SSE `stageEstMs`, and
`fallbackPhase1Ms`.

**The Russian conjunction `как` was scanned as a missing character.** `RU_VERBS`
contains `сказал`/`сказала` and `RU_STOPWORDS` omitted `как`, so ordinary prose
— `Как сказал маг,` ("As the mage said,") — matched the `name-verb` tag pattern,
was absent from the roster, and forced a full stage-1 re-detection of a
67,097-char chapter. Both occurrences in the 650,932-char manuscript are the
conjunction; there are **zero** mid-sentence bare `Как`, which a real name would
produce. The three controls already on the list — `Что`, `Так`, `Вдруг` — match a
speech verb zero times in the same book.

## Test plan

Every fix ships a test seen red for the right reason:

- **BOM** — `readJson` parses a BOM-prefixed file; `readJsonWithRecovery` parses
  a BOM-prefixed **main** file and does **not** fall back, asserted by writing a
  *different* object into `.bak.1` and requiring the main file's value. Red:
  `expected { source: 'backup' } to deeply equal { source: 'main' }`. A control
  pins that a genuinely corrupt main file still recovers from `.bak.1`. Added on
  review: the backup-read strip is now covered too (mutation-proven — removing
  that one strip turns the new test red), plus BOM tests for the `paths.ts` and
  `cast-fingerprint.ts` readers, the latter pinning that the fingerprint is
  computed over raw bytes.
- **Estimate label** — one test drives the real route with no pipelined-mode
  forcing (the point being that the wording no longer depends on it) and pins
  the emitted log line. It fails against `main`, which hardcodes the old
  wording.
- **`как`** — Red: `AssertionError: expected [ 'Как' ] to not include 'Как'`. The
  control asserts a genuine unrostered speaker (`Гесер сказал`) is *still*
  reported, so the fix cannot be a blanket suppression — deleting the whole
  Russian tag scan would otherwise satisfy the first test.
- All literal BOM characters in tests were replaced with the `'﻿'` escape:
  a literal renders as `startsWith('')` in review and would degrade to an
  always-true test if lost to a re-encode. Zero literals remain in any file this
  PR touches.

`npm run verify:fast:branch` green (lint, typecheck, config:check, hooks,
frontend, server, build).

## Scope held deliberately

Only `как` is added to the Russian stoplist; `ES_`/`FR_`/`DE_STOPWORDS` are
untouched, needing their own corpus evidence. Six sibling Russian openers
(`Однажды`, `Как-то`, `Потому`, `Значит`, `Впрочем`, `Наконец`) were grepped
against the same 650,932-char manuscript before a speech verb: **zero hits**.
Review independently re-extracted the epub and confirmed. Further candidates
raised in review (the oblique pronoun `Ей`, a class of `-о` manner adverbs) are
filed separately — see below.

## On-box acceptance

**None owed.** All three defects are deterministic and unit-testable: a BOM in a
file, a string label, and a stopword list. Nothing here needs a live GPU,
sidecar, analyzer or book to prove — the *discovery* required a real run, the
*verification* does not. No register row is added or discharged.

- Three type imports orphaned by the estimate-label helper extraction; a spy
  analyzer stub missing a required `color` field; and a spy analyzer declaring
  `runStructureChapter`/`runAttributionChapter` — **neither of which exists on
  the `Analyzer` interface** — while omitting `runStage1` and
  `runNonStoryClassification`, which do. The suite was green throughout, because
  vitest does not enforce types.
- A literal BOM byte embedded in a release-notes paragraph that was itself
  warning against literal BOMs.

## Filed separately, not fixed here

- **#2366** — stage 2 makes the model re-emit every sentence's `text`: 13,720
  output tokens per call against 7,910 in, 99% of a local call's wall clock.
  Needs a decision on re-aligning output without text to match on.
- **#2367** — nothing warns when the analyzer model lands split across GPUs,
  despite `docs/local-llm.md` documenting the pin. Needs a decision, since a
  split is correct when the model cannot fit on one card.
- **#2369** — further Russian stoplist candidates raised in review.

Closes #2353
Closes #2354
Closes #2355


dudarenok-maker:
## PR review — pass 1 (head 6b66e25a, effort high)

Effort calibrated to the diff, not the title: three unrelated subsystems
(`workspace/`, `routes/`, `analyzer/`), three independent defects.

House gates checked and clean: `Closes #2353/#2354/#2355` present and outside
any code span; all three issues exist, open, labelled `bug`; release-notes pair
updated on both surfaces; no `cast.json` write, no new config knob, no derived
artifact staled; `npm run typecheck` green; the three touched suites run green
(63 tests). Red-before-fix verified by reasoning against `main`'s code plus a
direct probe that `JSON.parse('\ufeff{"a":1}')` throws
(`Unexpected token '﻿'`), which is the claimed reason.

Three correctness bugs, four cleanup nits.

---

### 🐞 Correctness bugs

#### C1 — Major — the BOM fix leaves the same file broken through a second reader

`server/src/workspace/paths.ts:47-60` (and `:89-98`)

`user-settings.json` — the exact file #2353 names — is read **twice** by the
server. The PR fixed the async reader (`readJson`). The other one is
synchronous, at module-eval time, and does its own `JSON.parse`:

```ts
const raw = readFileSync(path, 'utf8');
const parsed = JSON.parse(raw) as { workspaceDirOverride?: unknown };
...
} catch { return null; }
```

It resolves from the same `USER_SETTINGS_PATH` constant (`paths.ts:14, 90`), and
its `catch` swallows the BOM throw silently by design.

**Repro.** Save `user-settings.json` from Notepad (or `Set-Content -Encoding
utf8`) containing `{"workspaceDirOverride": "D:\\Castwright-Library", ...}` —
i.e. the PR's own premise, a hand-edited settings file with `EF BB BF` on the
front. Start the server.

- Before this PR: override lost, every setting lost.
- **After this PR: override still lost.** `readBootOverride()` returns `null`,
  `RESOLVED_DIR` falls through to `WORKSPACE_DIR`/`../castwright-workspace`
  (`paths.ts:36-39`), `WORKSPACE_ROOT` points at the wrong directory for the
  whole process, and the library reads as empty while the real one sits in
  `D:\`. `WORKSPACE_SOURCE` reports `'default'`/`'env'` instead of
  `'override'` (`paths.ts:41-45`), so the UI's own diagnostic agrees with the
  wrong answer.

This is rubric shape #10 (one instance fixed, the class left armed) compounded
by #1 (a guard that fails open on absent evidence). It is not "next door" — it
is the same file, the same three bytes, the same user action, and it defeats
half of what the PR claims to have fixed. The PR body's "`readJson` has ~227
non-test call sites" frames `readJson` as the choke point; `paths.ts` is the
counter-example and it is the one that matters for this bug.

**Second site, same class, worse failure mode.**
`server/src/workspace/cast-fingerprint.ts:97-102` — `readJsonWithFingerprint`
also bare-`JSON.parse`s and swallows the throw as `{ value: null, fingerprint:
<hash> }`. With a BOM'd `cast.json`, `readPriorCastForMerge`
(`server/src/routes/analysis.ts:214-236`) then takes the `fromCast` miss, the
carryover miss, and returns `{ rows: [], fingerprint: null, source: 'none' }` —
so the prior cast reads as **empty** *and* compare-and-set conflict detection is
**disabled**, which `analysis.ts:206-210` documents as exactly what a `null`
fingerprint means. A re-analysis then writes a fresh roster over the user's
`cast.json` with the conflict guard switched off. `cast.json` is this repo's
"identity of record"; losing it silently is a bigger blast radius than the
settings default-reversion the PR opened with.

---

#### C2 — Major — the `true` arm of the new helper is unreachable, and three documents assert otherwise

`server/src/routes/analysis.ts:2786-2791`, `:4460-4462`, `:4500`, `:4529`

`stage2EstRefined` is assigned `true` at exactly one place, line 4462, inside
`runPhase0Pool`. That pool is launched **without `await`** at line 4500:

```ts
      phase0PoolPromise = runPhase0Pool();   // 4500 — deliberately deferred
    }                                        // 4501
    ...
    markPhase(1);                            // 4516  (no await)
    send({ kind: 'phase', phaseId: 1, ... });// 4517  (no await)
    const totalChapters = ...;               // 4524
    log(1, `Attributing ...`);               // 4525
    log(1, `Estimated stage time: ~${...} ${selectStage2EstimateLabel(stage2EstRefined)}`); // 4529
```

There is no `await` anywhere between 4500 and 4529, and `runPhase0Pool` reaches
4462 only after real model calls plus `await saveAnalysisCache` /
`await persistDroppedQuotesBatch` (4403-4404). So **`stage2EstRefined` is `false`
at line 4529 on every run**. The cache-hit branch (`analysis.ts:3692-3754`) never
sets `stage1ActualMs` at all, so it is `false` there too.

This is not confined to pipelined mode. The comment at 4493-4499 says so
explicitly — "Defer the await — Phase 1 is launched concurrently below" — and
what sequential mode parks is the Phase 1 **workers** (`awaitPhase1Dispatch`),
not the Phase 1 **setup** that emits this log. Four consequences, each a defect
in its own right:

1. **The PR body is wrong.** "In sequential mode the refinement *has* run and
   the label is accurate, so the fix is conditional and leaves that path's
   output byte-for-byte unchanged." Sequential mode's output **changes** — from
   `(based on stage 1 rate)` to `(pre-flight estimate, refined after stage 1)`.
   (The new wording is the honest one; the claim about it is what's false.)
2. **The comment this PR adds is false the day it lands** —
   `analysis.ts:2787-2789`: "In sequential/refined modes, the refinement has
   run." Rubric shape #9.
3. **Both release-notes surfaces ship the false claim.**
   `docs/release-notes-next.md:287` — "sequential mode, where the refinement has
   already run, keeps its output byte-for-byte"; `RELEASE_NOTES.md:11` — "the
   estimate that genuinely is measured still says so". Nothing ever says so:
   `'(based on stage 1 rate)'` is now dead string.
4. **The proof cited doesn't reach the claim.** "Proven by mutating the
   conditional so both arms return the same string and confirming the
   refined-branch test goes red" proves the pure function distinguishes its two
   arguments. It cannot prove production ever passes `true` — and it doesn't.
   Rubric shapes #5/#6.

**Repro.** Any cache-miss analysis with per-phase model selection **off**
(sequential mode, the configuration the PR says is untouched): the log reads
`Estimated stage time: ~Xm Ys (pre-flight estimate, refined after stage 1)`.

**This also collapses the integration test's value.** Because the label is
pre-flight in *both* modes, `analysis.stage2-estimate-label.test.ts:215-254`
would pass **unchanged** with `setPipelinedMode({ pipelined: false })`. The
pipelined flag is presented as the causal variable and is never measured against
its own control — rubric shape #8. Answering the question you asked directly:
the 268-line integration test does not earn its length. It re-proves what the
two unit tests already prove (the helper's `false` arm reaches the log), it
cannot discriminate the condition its own name asserts, and its
`makeSpyAnalyzer` fixture is the part of this branch that has already needed
three rounds of repair. The two unit tests are the real coverage — but note that
one of them (`isRefined: true`) is now testing a branch nothing can call.

Fixing this is a design decision, not a mechanical edit: either move the flag
(and the log) to where the refinement is actually observable, or drop the flag
and the helper and log the honest constant wording — and correct the comment,
the PR body, and both release-notes entries either way.

---

#### C3 — Minor — the RU stopword row is fixed at one spelling, with the corpus in hand

`server/src/analyzer/tag-grammar.ts:52`

The list `тогда, потом, затем, однако, хотя, наконец, вдруг, теперь, здесь` is
labelled "name-verb openers (finding J)". `как` joins it. Two siblings in the
same category are still armed, both matching the `ru` name-verb regex built at
`tag-grammar.ts:119` and neither in `RU_STOPWORDS`:

| input | captured by the regex | in `RU_STOPWORDS` after this PR |
|---|---|---|
| `Как сказал маг, …` | `Как` | yes (fixed) |
| `Однажды сказала Ольга, …` | `Однажды` | **no** |
| `Как-то сказал Антон, …` | `Как-то` | **no** |

(Verified by executing the exact regex the module builds — `(?<!\p{L})(\p{Lu}[\p{L}’'-]+)\s+(?:сказал|сказала|…)(?!\p{L})` with flags `gmu` — against those three strings.)

`Однажды` ("one day / once") opens Russian narrative sentences at least as often
as `тогда`/`потом`, and `Однажды сказал мне отец…` is ordinary prose. With
`quoteAdjacent` true — which `roster-coverage.ts:222-224` makes near-certain in
dialogue-dense prose, and which is precisely why the existing false-positive
bound missed `как` — a **single** occurrence clears
`minHitsWithoutQuote` (`roster-coverage.ts:243`) and produces the same
missing-speaker verdict and the same full stage-1 re-detection of a 67k-char
chapter that #2355 is about.

The ES/FR/DE scoping decision is right and I would not widen it — those are
other languages with no corpus. This is a different decision: it is the *same*
language, the *same* enumerated category, and the *same* 650,932-char manuscript
that supplied the evidence for `как` was available to grep for `Однажды <verb>`
and `Как-то <verb>`. Rubric shapes #2 (a guard that enumerates syntax loses one
spelling per round) and #10. If the greps come back zero, say so and this
closes; leaving it unchecked is what makes it a finding.

---

### 🧹 Cleanup nits

#### N1 — the BOM test writes a **literal** BOM, the exact hazard the fix avoids

`server/src/workspace/state-io.test.ts:338` and `:352`

Production correctly uses `'\ufeff'` (`state-io.ts:52, 81, 90`) with a comment
explaining why. The tests do the opposite — verified with `cat -v`, both
`bomContent` literals begin with raw `M-oM-;M-?` (`EF BB BF`) inside the string.
The author's own argument applies with more force here: if that BOM is ever lost
to a re-encode, a lint autofix, or an editor's "save as UTF-8 (no BOM)", both
BOM tests silently become duplicates of `readJson parses a normal (non-BOM) JSON
file correctly` and go on passing forever while testing nothing. Rubric shape
#3. `const bomContent = '\ufeff' + '{"v": 1, "name": "test"}';` survives a
re-encode and reads unambiguously in review.

#### N2 — the third BOM strip ships untested

`server/src/workspace/state-io.ts:88-92`

The strip added to the `.bak.N` **backup** read has no paired test. Both added
recovery tests write a BOM to the *main* file. Delete lines 88-92 and every test
in the branch stays green — for a hunk whose whole point is the recovery path
(the one that runs when the main file is already unreadable). A BOM'd backup is
not exotic: it is the same Notepad/PowerShell hand-edit applied to a file the
user was told to look at when recovering. One test, mirroring the existing
`.bak.1` fallback test with a BOM on the backup.

#### N3 — on-box acceptance addressed in neither direction

The PR frames all three defects as "none of them is reachable from a fixture —
they needed a live run to surface", and `docs/release-notes-next.md:287` opens
with "Three defects that only a real book on real hardware could surface". The
diff touches none of the register's three surfaces and the body says nothing
about acceptance. Before-shipping step 3 wants a row **or** an explicit
not-applicable. The defensible position here is probably "the mechanism of each
fix is fully unit-provable, so nothing is owed" — but that sentence has to
appear, otherwise the omission is indistinguishable from a miss. (If C3 stands,
the `Однажды`/`Как-то` greps against Ночной дозор are real-book evidence and may
change the answer.)

#### N4 — incidental findings declared in the wrong place, and one of them isn't a finding

`docs/release-notes-next.md:288`

The "two gate gaps" paragraph is a process observation, not a shippable delta,
and it is the only place the incidental findings are declared — CLAUDE.md asks
for them in the **PR body** ("Also fixed, found in passing: …"), and for an
issue filed in the same round; neither happened. Separately, the first of the
two is not a gap: CLAUDE.md's "Commit gate" already documents that typecheck is
deliberately not in pre-commit ("Sidecar (pytest), Pester scripts, Playwright
e2e, and typecheck are NOT in pre-commit — they live in pre-push so commits stay
snappy"). Recording documented design as a defect in the technical release
register makes the register less trustworthy, not more. The second one (vitest
does not enforce types, so a fabricated `Analyzer` method stays green) is a real
and useful observation and worth keeping.

Also at `docs/release-notes-next.md:289-290`: a stray double blank line before
the `---` rule.

---

### Summary

- **3 correctness bugs** (C1 Major, C2 Major, C3 Minor) → re-review is triggered
  once these are fixed and pushed.
- **4 cleanup nits** (N1-N4).

C1 and C2 both mean the PR does less than its body says: C1 leaves the named
file still broken through a second reader; C2 ships a conditional whose "true"
branch cannot execute, plus a comment, a PR body and two release-notes entries
that assert the behaviour it does not have. C3 is the smallest and the most
optional — one grep settles it either way.

<sub>Reviewed at head `6b66e25a`. No tracked file modified; working tree clean
before and after.</sub>

dudarenok-maker:
## PR review — pass 2 (head `b77f88f3`, effort high)

Re-review of `6b66e25a..b77f88f3`. **Deltas only** — what was solid in pass 1 is
not restated. Every disposition below was verified, not accepted.

---

### Pass-1 findings — dispositions

| # | Verdict |
|---|---|
| C1 | ✅ **Verified resolved** |
| C2 | ✅ **Verified resolved in code** — but see **D1**, the claim survives in the PR body |
| C3 | ✅ **Verified resolved** on the condition I stated — but the verification turned up **D2** |
| N1 | ✅ Verified resolved |
| N2 | ✅ Verified resolved |
| N3 | ⛔ Still open (nit) |
| N4 | ◐ Half resolved |

**C1 — verified resolved, including the fingerprint question.**
`cast-fingerprint.ts:97` computes `hashBytes(raw)` *before* the strip at `:105`,
and only `stripped` reaches `JSON.parse`, so no stored fingerprint moves. That is
also the right choice and not just a safe one: the only value-derived fingerprint
in the tree is `fingerprintOfWrite` (`cast-fingerprint.ts:121`), used solely by
`cast-merge-base.ts:126` to advance the baseline after *our own* write — so a
BOM'd file legitimately hashes differently from anything we wrote, which is a true
statement ("a third party edited this"), and `writeChecked` still round-trips
(read → hash(BOM+body) → same on re-read → write normalises the BOM away).
`cast-fingerprint.test.ts:97-124` pins both halves against exact `hashBytes`
values, not just "not equal". The `paths.ts:60` strip closes the pass-1 repro:
`readBootOverride` now returns `D:/Castwright-Library` from a BOM'd settings file
(`paths-boot-override.test.ts:112-121`). 45 tests green across the three workspace
suites.

**C2 — verified resolved in code.** `git grep` finds zero references to
`selectStage2EstimateLabel` or `stage2EstRefined` outside two comments. The log at
`analysis.ts:4527-4529` states the wording unconditionally.

- *Did deleting the flag lose anything?* No. `stage2EstMs`'s own refinement
  survives at `:4452` and still reaches everything downstream of the watermark:
  the per-chapter budget (`:4583`, `:4585`), the SSE `stageEstMs` (`:4847`), and
  `fallbackPhase1Ms` (`:5190`). Only the label stopped depending on it.
- *Can the new test fail?* Yes, and for the claimed reason. `main`'s
  `analysis.ts:4519` hardcodes `(based on stage 1 rate)`, so both assertions are
  red there; and `expect(estimateLogs.length).toBeGreaterThan(0)` blocks the
  null-observation pass — if the log never fires the test fails rather than
  vacuously passing. It cannot silently degrade to a constant either, because it
  asserts the absence of the opposite wording. Runs green (1 test, 2.9 s).

**C3 — verified resolved, independently.** I did not take the greps on trust. I
extracted the actual `manuscript.epub` (654,380 chars after tag-strip — the
author's 650,932 is the same file) and ran the regex `buildOrderRegex` actually
constructs, both orders, against it:

```
name-verb captures, whole book (21 total)
  suppressed by RU_STOPWORDS:  Он ×4, Она ×2, Как ×2, Потом ×1
  Однажды: 0   Как-то: 0   Потому: 0   Значит: 0   Впрочем: 0   Наконец: 0
```

`Как` is confirmed at exactly 2, both the conjunction, and is now suppressed. All
six siblings are genuinely zero. **C3 closes exactly as I said it would.**

**N1 / N2 — verified resolved.** `grep -c $'\xef\xbb\xbf'` returns 0 for all eight
touched files, so the escapes are real and no literal survived. The new
backup-read test is a genuine second strip site (a different `readFile`), green.

**N3 — still open.** There is still no acceptance sentence in either direction —
not in the diff, not in the PR body. Same nit as pass 1.

**N4 — half resolved.** The false "pre-commit is a gate gap" claim is gone, the
kept half is correct, and the stray double blank line at `release-notes-next.md`
is fixed. Still outstanding: the incidental finding is declared only in the
release register, not in the PR body ("Also fixed, found in passing: …"), and no
issue was filed for it.

---

### 🐞 New — correctness

#### D1 — Major — the PR body still asserts the exact claim `b77f88f3` deleted

The commit is titled *"correct false claims about the C2 fix"*, but it edited only
`RELEASE_NOTES.md` and `docs/release-notes-next.md`. The PR body — which is the
durable record and lands in the merge commit — still reads:

> In sequential mode the refinement *has* run and the label is accurate, so the
> fix is conditional and leaves that path's output byte-for-byte unchanged.

Every clause of that is now contradicted by the branch's own code comment
(`analysis.ts:4519-4526`, "in EVERY mode"), by its release note ("in EVERY mode,
not just pipelined mode"), and by the test's header. The fix is *not* conditional
— the conditional was deleted in `5fe24d98`. Sequential mode's output does change.

The **Test plan** section is stale in the same way:

> Estimate label — both branches asserted, each also asserting the *absence* of
> the opposite wording, so neither can quietly become a constant. Proven by
> mutating the conditional so both arms return the same string…

There are no longer two branches, two tests, or a conditional to mutate. A reader
merging on that Test plan is being told the branch was verified in a way it was
not. Rubric shape #9 — the defect is in the document, not the code.

**Fix:** edit the PR body. No commit required; nothing in the tree changes.
(Pass-1 C2 listed "correct the comment, the PR body, and both release-notes
entries" — three of four landed.)

#### D2 — Minor — `RU_STOPWORDS` is still one-spelling-at-a-time, and the same book supplies seven more

`server/src/analyzer/tag-grammar.ts:52`

This is **not** a re-opening of C3 — C3's condition was met. It is what the
verification turned up. Running the real `name-verb` regex over the whole
manuscript and applying the real gate arithmetic (`roster-coverage.ts:243`,
`minHitsWithoutQuote: 2`, `quoteProximityChars: 60`,
`QUOTE_CHARS_WIDE` at `:172`), the captures that are **neither** in
`RU_STOPWORDS` **nor** roster names are:

| capture | gloss | hits | class |
|---|---|---|---|
| `Ей` | "to her" — dative of `она` | 1 | **pronoun, inflected** |
| `Осторожно` | "carefully" | 1 | `-о` manner adverb |
| `Честно` | "honestly" | 1 | `-о` manner adverb |
| `Хорошо` | "well / nicely put" | 1 | `-о` manner adverb |
| `Неохотно` | "reluctantly" | 1 | `-о` manner adverb |
| `Глухо` | "dully" | 1 | `-о` manner adverb |
| `Темный` | "the Dark One" | 1 | role noun, not a name |

Real sentences from the book (dialogue dash reproduced as it appears in this
file, ASCII `-`), each one hit:

```
…поднял голову. Глухо сказал: - Разрешение дано.
…замолчала. Грустно покачала головой. Неохотно сказала: - Ты можешь…
…да что происходит? Ей ответил Гесер. - У каждого своя судьба, девочка.
```

**Why each fires.** One hit is enough whenever `quoteAdjacent` is true —
`if (c.tagCount < 2 && !c.quoteAdjacent) continue;`. Substituting the dialogue
dash for `—` (the ordinary Russian convention) and re-running: **11 of the 21
captures survive the bound**, of which the seven above are not roster names. Each
is then an unrostered "missing speaker" and forces the same full stage-1
re-detection #2355 is about — the identical harm, seven more spellings, in the
book the PR already used as its corpus.

**One caveat, stated plainly, because it cuts both ways.** *This particular*
`manuscript.epub` is a Calibre TXT conversion that uses ASCII `-` and `"`
(census: em-dash 0, `«»` 0, hyphen 7302), so on *this* file `quoteAdjacent` is
false for six of the seven and they would be filtered. But that is also false for
**both `Как` occurrences** — which contradicts the PR's own stated mechanism
("`tagCount` was 1 but the prose is dialogue-dense, so `quoteAdjacent` was
true"). So either:

- the analysed text was an em-dash edition (as the PR's account of `как`
  requires) → the seven above fire, and D2 stands as written; or
- it was this ASCII edition → `как` could not have fired at `tagCount` 1 either,
  and the mechanism in the PR body and release note is wrong.

Either branch is a finding; the first is much more likely, since the incident
actually happened.

**Suggested split** — the two halves are not the same kind of work:

1. **`ей`/`ему`/`им`/`их` → `RU_STOPWORDS`, this round.** Unarguable: the list
   already enumerates `он`/`она`/`оно`/`они`, and these are the oblique cases of
   those same pronouns in a six-case language. Nothing to decide.
2. **The `-о` adverb class → an issue with the decision named.** Enumerating
   `осторожно/честно/хорошо/неохотно/глухо` buys five spellings out of an
   open class and will lose one per round (rubric shape #2, the exact shape the
   `как` fix has now hit twice). The real options are a `-о` adverb heuristic, a
   "capitalized only ever at sentence-start in this book" test, or accepting the
   false positives and making the *re-detection* cheap. That is a design pass,
   and is the one thing this repo's rules let you defer — provided the issue says
   which of those three is being chosen.

---

### 🧹 New — cleanup

#### D3 — the rewritten test's name claims a mode it never asserts

`server/src/routes/analysis.stage2-estimate-label.test.ts:179`

The title says "including in default (sequential) mode" and the header comment
leans on it, but nothing in the body observes the mode — the only guarantee is
that `beforeEach` deletes the per-phase env vars. Stakes are much lower than the
pass-1 version (the assertion no longer depends on the mode at all), but the
test's headline claim can now rot silently: an env leak, or a change to
`isPerPhaseModelSelectionActive`'s default, and it still passes while its name is
false. One line fixes it —
`expect(isPerPhaseModelSelectionActive()).toBe(false)` before the run, importing
the real symbol through the same partial mock that already spreads `...actual`.

---

### Verdict

- **2 new correctness findings** (D1 Major — documentation only, fixed by editing
  the PR body; D2 Minor — one half is a two-token edit, the other is a named
  design pass), **1 new nit** (D3).
- All three pass-1 correctness bugs are **verified resolved**, C1's fingerprint
  question specifically. N1/N2 resolved; N3 open; N4 half.
- The **code** on this branch is sound. What is not shippable as-is is the PR
  body: it currently tells a reviewer the opposite of what the branch does.

Not mergeable until D1 is corrected. D2's pronoun half is cheap enough that it
belongs in this round under the incidental-findings rule; its adverb half is a
legitimate ticket.

<sub>Reviewed at head `b77f88f3`. Typecheck green; the four touched suites green
(46 tests). No tracked file modified; `git status --porcelain` empty before and
after.</sub>

PR #2375, opened 2026-08-14
feat(docs,scripts): route reasoning effort via six named dispatch roles

### LINKED ISSUE #2372: ops — route reasoning effort via six named dispatch roles
## What

Reasoning effort is currently ungoverned. `model-routing` routes *which model* a
dispatch lands on; nothing routes *how hard it thinks*. Every dispatched subagent
inherits whatever `effortLevel` the dispatching session happened to be left at.

This ships the second axis: six named `.claude/agents/*.md` roles that pin
`model:` **and** `effort:`, dispatched by `subagent_type` rather than by raw
`model:`, with `medium` declared as the repo-wide norm.

**Design spec:** `docs/superpowers/specs/2026-08-14-model-effort-routing-design.md` (rev 9)
**Implementation plan:** `docs/superpowers/plans/2026-08-14-model-effort-routing.md` (9 tasks)

Both on branch `docs/docs-model-effort-routing`. Twelve decisions, three
adversarial review rounds, one empirical gate, one cross-session convergence
pass — the spec's "Read this first" table is the cold-entry point.

### The three failures this closes

1. **Hard passes underthink.** The adversarial gates (`pr-review-gate`,
   `assumption-checker`) are the passes that catch evidence-level defects, and
   they run at whatever effort the session was left at.
2. **Cheap fan-out overthinks.** Mechanical search-and-report dispatches — routed
   to Haiku precisely because they have one right answer — still reason at full
   session effort.
3. **The session drifts.** `CLAUDE.md` has a session-level *model*-drift rule.
   There is no effort equivalent, so design work at `low`, or transcription at
   `xhigh`, goes unremarked.

### The roster

| Role | Tier | `model:` | `effort:` |
|---|---|---|---|
| `pr-reviewer` | Premium | `opus` | `xhigh` |
| `spec-checker` | Premium | `opus` | `xhigh` |
| `task-reviewer` | Default | `sonnet` | `high` |
| `implementer` | Default | `sonnet` | `medium` |
| `fix-agent` | Cheap | `haiku` | `medium` |
| `scout` | Cheap | `haiku` | `low` |

## Acceptance

- [ ] Six tracked `.claude/agents/*.md` definitions whose `model:`/`effort:`
      match the role table in `model-routing/SKILL.md` exactly.
- [ ] The role table is a **closed registry** — a definition file with no table
      row fails the build, and a row with no file fails too. Both directions
      proven by mutation, not by absence.
- [ ] `.claude/agents/**` is a `test:hooks` scope input, so a definitions-only
      diff cannot print `[cached]` and gate nothing.
- [ ] `pr-review-gate`'s "effort ladder" is renamed **"review depth"**; the word
      `effort` thereafter means only the model setting, repo-wide.
- [ ] A session-effort drift rule in `model-routing` — read `effortLevel`, compare
      bands, flag and ask. Never silent, never self-applied.
- [ ] `model-routing/SKILL.md` joins the cross-agent mirror (fixes the dead-links
      finding below).
- [ ] `medium` declared the norm in `CLAUDE.md` — one sentence, no second table.

## Also fixed in the same round (incidental finding, per CLAUDE.md)

**F1 — the mirror's dead routing links.** `.claude/skills/pr-review-gate/` links
`../model-routing/SKILL.md` four times, including the link naming which tier to
dispatch at. `sync-agent-skills.mjs` mirrors `pr-review-gate` only, so in
`~/.agents/skills/` every one of those resolves to a directory that is not there.
**Cline has been reading a runbook with dead routing references since the mirror
was created.** Fixed in this round rather than filed separately.

## Key files

- `.claude/agents/*.md` — new, the roster
- `.claude/skills/model-routing/SKILL.md` — the registry table, session-effort rule, escalation asymmetry
- `.claude/skills/pr-review-gate/SKILL.md` + `references/*.md` — effort → depth
- `scripts/verify-cache.mjs` — `.claude/agents/**` scope glob
- `scripts/tests/review-gate-mechanism.test.mjs` — the bidirectional guard
- `scripts/sync-agent-skills.mjs` — `model-routing` joins the mirror
- `.gitignore` — `!.claude/agents/`

## Depends on / relates to

- **#2368** — probes whether Cline resolves *workspace* skill roots. Overlaps this
  work's owed second probe (decision 9), which asks whether Cline resolves agent
  *definitions*. Definitions are deliberately **not** mirrored until it runs.
- **#2314** — the three drifting cross-agent skill copies. F1 above is one
  concrete instance of that drift.
- **#2243** — whether `docs/superpowers/plans/**` is a record or live
  instructions. This work's plan lands there.

## Benefit

*(technical)* Gate depth stops being an accident of session state. The two passes
that catch evidence-level defects run at `xhigh` every time rather than at
whatever the session drifted to, and the highest-volume cheap role stops paying
full session effort for mechanical search.

dudarenok-maker:
## Summary

Reasoning effort was ungoverned. `model-routing` routed *which model* a dispatch landed on; nothing routed *how hard it thinks*, so every dispatched subagent inherited whatever `effortLevel` the dispatching session happened to be left at. The adversarial gates — the passes that catch evidence-level defects — ran at whatever depth the session had drifted to, and mechanical Haiku fan-out reasoned at full session effort.

This ships the second axis.

**Six named dispatch roles**, one `.claude/agents/*.md` each, pinning both `model:` and `effort:`. Dispatch becomes `Agent({subagent_type: 'pr-reviewer'})` rather than `Agent({model: 'opus'})`.

| Role | Tier | `model:` | `effort:` |
|---|---|---|---|
| `pr-reviewer` | Premium | `opus` | `xhigh` |
| `spec-checker` | Premium | `opus` | `xhigh` |
| `task-reviewer` | Default | `sonnet` | `high` |
| `implementer` | Default | `sonnet` | `medium` |
| `fix-agent` | Cheap | `haiku` | `medium` |
| `scout` | Cheap | `haiku` | `low` |

**`medium` is now the declared repo-wide norm** — for roles, for the main session, and for CLI worker dispatch (`claude --effort`, `copilot --effort`, `cline --thinking`). Every role declares `effort:` explicitly *including the two sitting at the norm*, because omitting the key inherits the dispatching session's effort, which is the ungoverned behaviour this exists to end.

On the CLI lane that is a **declaration, not an enforcement** — see the M10 note below.

**The role table is a closed registry.** A table row without a definition file fails the build, and a definition file without a table row fails too. There is deliberately no third path — no escape comment, no allowlist.

**`pr-review-gate`'s "effort ladder" is renamed "review depth".** It keeps its values and its commit-type derivation; only the noun changes. A reader meeting `effort: xhigh` in a definition and `effort: low` in a PR comment header previously had no way to tell them apart. Repo-wide, `effort` now means only the model setting. Comments posted before 2026-08-14 use the old noun for the same thing and are not rewritten — they are records.

**`.claude/agents/**` joins `test:hooks`'s scope globs.** Without this, a definitions-only diff — flipping `effort: xhigh` to `low`, say — would print `test:hooks [cached]` and gate nothing, locally *and* in cloud CI, since `ci-scope.mjs` derives from the same `STEPS[]`.

Also included: a session-effort drift rule mirroring the existing model-drift rule in shape (flag and ask, never silent, never self-applied), and two probe records under `docs/testing/`.

**F1 — the cross-agent mirror's dead routing links.** `.claude/skills/pr-review-gate/` links `../model-routing/SKILL.md` four times — three in `SKILL.md`, one in `references/findings-triage.md` — including the link naming which tier to dispatch a reviewer at. `sync-agent-skills.mjs` mirrored `pr-review-gate` only, so in `~/.agents/skills/` every one of those resolved to a directory that was not there. **Cline has been reading the PR-review runbook with dead routing references since the mirror was created.**

Fixed by making `FILES` skill-qualified and adding `model-routing/SKILL.md` to it, plus a guard so it cannot silently recur. The guard computes from `buildMirrorContent`'s return value rather than reading `~/.agents/skills/` from disk — a disk-based check skips in CI, which is exactly why this survived unnoticed.

## Test plan

- `npm run test:hooks` — **1412/1412 pass**, forced (not `[cached]`). `review-gate-mechanism.test.mjs` is **24 tests, up from 17**; `verify-cache.test.mjs` 98.
- `npm run verify:fast:branch` — green. `lint` ran and passed; frontend/server/typecheck/build correctly out of scope (no `src/`, `server/`, or TS changes).
- **The roster guard's red phase was proven by mutation, not by absence** — the guard is green the moment it is written, since the definitions and the table already agree. Four mutations, one at a time: a value drifting between file and table; an ungoverned definition appearing; `scout` gaining `Edit`; and the table heading renamed. That last one is the vacuous-pass case — a guard whose section regex stops matching iterates an empty array and passes silently. All four caught on the first attempt.
- **The mirrored-link guard was proven against the real pre-fix state**, reporting exactly the 4 dead links F1 describes — not "≥1", which would pass on a guard firing for the wrong reason.

**The review gate then found three vacuity holes in these same guards, and they are fixed** (`eb8a3920`): the registry's reverse scan was non-recursive while the harness's loader recurses, so a `.claude/agents/legacy/rogue.md` was dispatchable and invisible; `CONTRIBUTING.md` was scanned by the link guard but registered in no step's scope, so a CONTRIBUTING-only diff ran it nowhere; and `FILES` had no non-empty assertion, so `FILES = []` left every mirror test green. Each fix carries its own mutation proof. That a branch about guards-that-cannot-pass-vacuously shipped three vacuity holes is the honest headline of this PR's review.

## Deliberately not done

- **Release notes: skipped.** Process-only change, no user- or operator-visible delta — the exemption `CLAUDE.md` before-shipping step 5 names.
- **On-box acceptance: none owed.** Nothing here needs a live GPU, sidecar, analyzer, or real book. No register row.
- **e2e: none.** Nothing crosses a router/redux/layout seam.
- **Agent definitions are NOT mirrored to Cline.** Cline cannot select a subagent's model, and whether it resolves agent definitions at all was never probed — writing files carrying `model: opus` into a harness that can honour neither is a sync that reports files written and changes nothing. The owed probe is recorded in `docs/testing/agent-skill-resolution-probe.md`; see also dudarenok-maker/open-engine#50, which probes an adjacent question.

## Notes for the reviewer

- The design spec (`docs/superpowers/specs/2026-08-14-model-effort-routing-design.md`, **rev 11**) opens with a "Read this first" table giving all twelve decisions in one screen.
- **Three cross-session convergence reports were folded in, each falsifying something.** Rev 9: the load-bearing claim *"effort is not settable per dispatch"* is true of the `Agent` tool and false of the worker CLIs — and of `Workflow`'s `agent()`, an in-harness surface the report itself missed. Decision 10 keeps its conclusion but now states its because-clause. Rev 10: the deferred intake question was answered (declared-default-with-exceptions), with no roster rework. Rev 11: see below.
- **M10 is deliberately not a bare "Confirmed".** It reads *Confirmed — flag declaration only*. Its evidence is `--help` on each CLI, which proves a flag is **declared** and what values it parses, not that it is **honoured** at runtime. A parallel session reports `copilot --effort` is ignored under `--model auto`; that is recorded as **reported, not reproduced here** rather than upgraded to fact or dropped. This branch's guidance is unchanged by it — passing a flag a configuration ignores costs nothing, and the alternative is inheriting an undeclared default.
- **`main` is merged in** (branch is 0 behind). One file conflicted, `CLAUDE.md`'s `skills:sync` bullet, because both sides had improved it differently — ours named the second mirrored directory and corrected skip-vs-fail, main hedged the "ONLY store" over-claim per dudarenok-maker/Castwright#2364. Resolved by taking **both**; verified all four corrections survive.
- `npm run skills:sync` is a **per-machine** step CI cannot run. Re-run it after merging.

Closes dudarenok-maker/Castwright#2372


dudarenok-maker:
## PR review — pass 1 (head b6f2538c, depth high)

_(This PR is the one that renames the noun, so this header uses `depth`. Comments before 2026-08-14 read `effort <level>` for the same axis.)_

Scope: `scripts/verify-cache.mjs`, `scripts/ci-scope.mjs`, `scripts/sync-agent-skills.mjs`, `scripts/tests/review-gate-mechanism.test.mjs`, `scripts/tests/verify-cache.test.mjs`, all six `.claude/agents/*.md`, `.claude/skills/model-routing/SKILL.md`, `.claude/skills/pr-review-gate/**`, `CLAUDE.md`, `.gitignore`, `docs/testing/**`, plus the spec's "Read this first" table and the plan's Global Constraints. · verified: `npm run test:hooks` forced green (**1410/1410**), `review-gate-mechanism.test.mjs` **23/23**, `verify-cache.test.mjs` **97/97** — then adversarially probed whether this PR's own guards can fail, via **nine mutations** in a throwaway detached worktree (`$TEMP`, never this tree). Six mutations went red for the right reason. Three did not, and are below.

Also read the Claude Code harness bundle (`~/.local/share/claude/versions/2.1.233`) directly for the agent loader, following this branch's own precedent of reading the schema rather than probing behaviour.

---

### 🟠 Significant — the "closed registry" is not closed: a nested definition file escapes both directions of the guard

`scripts/tests/review-gate-mechanism.test.mjs:427-430` — the reverse direction uses a **non-recursive** `readdirSync(AGENTS_DIR)`:

```js
const onDisk = readdirSync(AGENTS_DIR)
  .filter((f) => f.endsWith('.md'))
```

Two functions above it, `linkScanSet()` (`:293`, `:297`) uses `readdirSync(root, { recursive: true, … })` for the same job. The recursion was dropped only here.

**Repro** (run in a detached worktree at `b6f2538c`):

```
.claude/agents/legacy/rogue.md
---
name: rogue
model: opus
effort: max
---
```

* `git check-ignore -v .claude/agents/legacy/rogue.md` → exit 1 — **not ignored**, i.e. fully committable. `.gitignore`'s `!.claude/agents/` re-includes the whole subtree, not just its top level.
* `node --test scripts/tests/review-gate-mechanism.test.mjs` → **23 pass, 0 fail**. An ungoverned `model: opus` / `effort: max` definition sits in the repo and every assertion is green.
* `ci-scope.computeScopes(['.claude/agents/legacy/rogue.md']).step_test_hooks` → `true`, so CI *runs* the guard and the guard *certifies* the file.

**And the harness does load it.** From the bundle's own agent-definition loader (`Ntr` → `iIS`, the non-ripgrep branch, verbatim):

```js
async function o(a,l,c){ if(a.isDirectory()) await s(c); else if(a.isFile() && l.endsWith(".md")) r.push(c) }
```

— a recursive walk collecting every `.md` at any depth. The ripgrep branch is `rg --files --hidden --follow --no-ignore --glob '*.md'` rooted at the directory, which is also depth-unbounded (a glob with no `/` matches the basename at any depth). So `.claude/agents/legacy/rogue.md` becomes a dispatchable `subagent_type` and the registry never sees it.

This is the exact third path `model-routing/SKILL.md` says does not exist:

> There is deliberately no third path — no `# unmanaged` escape comment, no allowlist. An ungoverned definition file is precisely what the declared effort norm exists to prevent, and a registry with an opt-out is not a registry.

One directory is the opt-out. Fix is one word (`{ recursive: true }`) plus a mutation case pinning it; the guard's own failure message ("or move the file out of `.claude/agents/`") also needs to stop implying a subdirectory qualifies.

---

### 🟠 Significant — `CONTRIBUTING.md` is scanned by the link guard but is in no step's scope, so the guard never runs on the diff that breaks it

`scripts/verify-cache.mjs:163-170` — the new `docs/testing/**` glob's comment states:

> `linkScanSet()` now reads every .md file under this directory as TEXT at RUNTIME, **alongside CLAUDE.md/CONTRIBUTING.md/.claude/skills/\*\* above.**

`CLAUDE.md` is above (an explicit `extraFiles` entry at `:275`, added by ops-55/#2241 for precisely this reason). `.claude/skills/**` is above (`:153`). **`CONTRIBUTING.md` is neither** — `grep -n CONTRIBUTING scripts/verify-cache.mjs` returns exactly one hit: line 165, the comment asserting it. So this PR introduces a comment claiming a registration that does not exist, and leaves the sibling case armed while adding two more entries of the same class.

**Repro** — both halves measured at `b6f2538c`:

1. There is real coverage to lose. Append `See [nope](docs/features/this-file-does-not-exist.md).` to `CONTRIBUTING.md` → `review-gate-mechanism.test.mjs` goes red: `CONTRIBUTING.md -> docs/features/this-file-does-not-exist.md (file does not exist)`.
2. The guard never runs on that diff:

```
stepTouchedByDiff(test:hooks, ['CONTRIBUTING.md'])            -> false
computeShared(['CONTRIBUTING.md'])                            -> false
ci-scope.computeScopes(['CONTRIBUTING.md']).step_test_hooks   -> false
```

So a `CONTRIBUTING.md`-only change prints `test:hooks [cached]` locally **and** skips its cloud leg — the dudarenok-maker/Castwright#1847 trap this PR's three new comments recite at length, on one of the two root governance docs `linkScanSet()` reads first. `CONTRIBUTING.md` is the branching/commit/PR convention doc and is dense with intra-repo `.md` links.

Fix: add `'CONTRIBUTING.md'` to `test:hooks`'s `extraFiles` beside `'CLAUDE.md'`, with a `stepTouchedByDiff` case mirroring the two this PR already added, and correct the line-165 comment either way.

---

### 🟠 Significant — `FILES` has no non-empty assertion, so emptying it turns three mirror guards vacuous and `skills:sync` reports success having written nothing

`scripts/sync-agent-skills.mjs:50-55` (the `FILES` export) and its three consumers: the drift test (`review-gate-mechanism.test.mjs:536`), the new mirrored-link test (`:553-570`), and `syncAgentSkills()` (`sync-agent-skills.mjs:152`). All three iterate the list with no assertion that it is non-empty.

**Repro** — replace the list body with `export const FILES = [];` at `b6f2538c`:

* `node --test scripts/tests/review-gate-mechanism.test.mjs` → **23 pass, 0 fail**.
* `node scripts/sync-agent-skills.mjs` → exit **0**, printing its normal `This is a per-machine step … Re-run npm run skills:sync after any change under .claude/skills/pr-review-gate/ or .claude/skills/model-routing/.` banner, having written zero files.

This is the same hole this PR closes deliberately one screen away, and says so:

> Exact count, not `>= 1`: that's what closes the vacuous-pass hole above — an empty or partially-parsed `rows[]` would satisfy `>= 1` just as easily as a correct 6-row parse. (`:377-380`)

The role table got that treatment; the `FILES` registry — whose *incompleteness was F1 itself* — did not. Note the partial cases **are** caught (dropping only `model-routing/SKILL.md` reproduces F1 exactly, 4 dead links, verified below); it is the total-truncation case that passes silently. An `assert.ok(MIRRORED_FILES.length >= 4)` or an exact count in the link test closes it.

---

### 🟡 Minor — `scripts/sync-agent-skills.mjs:12-13` attributes four links to `SKILL.md`; it carries three

> `pr-review-gate/SKILL.md` links `../model-routing/SKILL.md` **four times**

Measured: `SKILL.md:76`, `:251`, `:265` — **three**. The fourth is `references/findings-triage.md:69`. The jsdoc at `:46` gets it right by saying "pr-review-gate" (the directory), and the PR body gets it right explicitly ("three in `SKILL.md`, one in `references/findings-triage.md`"). Only the file-header comment mis-attributes, so a reader auditing it by grepping `SKILL.md` finds three and concludes the comment has drifted.

---

### 🟡 Minor — the PR body's test-plan figure is wrong: 23, not 40

> `npm run test:hooks` — **1409/1409 pass** … 40 of those are `review-gate-mechanism.test.mjs`, up from 17.

`review-gate-mechanism.test.mjs` reports `tests 23` at `b6f2538c`, and `tests 17` at `main` — so "up from 17" is right and 17 + the 6 new tests = 23. The `41` in the runner's output belongs to the **preceding** file's summary, printed immediately before this file's first `✔` line; the interleaving is easy to misread. The aggregate also reads **1410**, not 1409, on a forced run here. The number is load-bearing in a PR whose whole subject is guards that measure what they claim to.

---

### ✅ What is solid

* **The roster guard genuinely fails.** Six mutations, one at a time, each restored: `pr-reviewer` `effort: xhigh`→`low` (red, right test); an orphan top-level definition (red, reverse test); the `## Named dispatch roles` heading renamed (red on the count assertion — the vacuous-pass case, caught); a Tier/model disagreement `implementer | Default`→`Cheap` (red with `Tier "Cheap" implies model "haiku", but the row's model column says "sonnet"`); `scout` gaining `Edit` (red with the intended message, re-run BOM-free after my first attempt went red for a frontmatter reason instead); and a cross-skill link to a non-mirrored skill (red, `model-routing/SKILL.md -> ../run-app/SKILL.md`).
* **The mirrored-link guard was proven against the real pre-fix state, not a synthetic one.** Removing `model-routing/SKILL.md` from `FILES` reproduces F1 and reports exactly the four dead links — three from `SKILL.md`, one from `findings-triage.md`. The choice to compute from `buildMirrorContent` rather than read `~/.agents/skills/` is correct and correctly reasoned: a disk-based check would skip on the machine that gates the merge, which is why F1 survived.
* **The skip/fail split on the two mirror tests is correct and correctly documented.** The drift test now keys on the store *root*, so it skips only where no store exists and fails where the store exists but this repo's mirror is stale — and `CLAUDE.md`'s rewritten `skills:sync` bullet states exactly that, including the "even if `skills:sync` has never been run on that machine" consequence. Verified against the store on this box.
* **The `test:hooks` scope wiring works on the diffs it guards.** `.claude/agents/pr-reviewer.md`, a not-yet-existing `.claude/agents/some-future-role.md`, and `docs/testing/*.md` all resolve `true` through `stepTouchedByDiff` *and* through `ci-scope.computeScopes`, so the local cache and the cloud leg agree. (The one gap is `CONTRIBUTING.md`, above.)
* **`effortLevel` is a real key.** Confirmed in the harness settings schema and present as `"effortLevel": "high"` in `~/.claude/settings.json` on this box — the session-drift rule is followable, and its two stated limits (file holds the *configured* value; reported, never self-applied) are honest.
* **`model:` override precedence and the no-effort-param asymmetry are stated accurately.** The `Agent` tool's schema does say the `model` parameter takes precedence over the definition's frontmatter, and it does expose no effort parameter — so decision 10's because-clause is correct, not hand-waved.
* **Global Constraints hold.** Six roles at the exact specced `model:`/`effort:` values; no `model: inherit`; no `tools:` list claimed as a security boundary (both `scout.md` and `model-routing` say the opposite explicitly); the `#2320` citation untouched (now `SKILL.md:164-165`; it appears in the diff as context lines only, never `+`/`−`); no `src/`/`server/` changes; the seven historical `effort <level>` comments not rewritten.
* **The rename's blast radius is clean.** `grep` for `effort ladder|effort level|effort <level>|effort scales` across all tracked `.md` returns zero live-governance hits — the survivors are `pr-review-gate/SKILL.md:67` (the deliberate migration note), historical `docs/superpowers/{specs,plans}/**`, and git-ignored `.superpowers/` scratch.
* **House gates**: `Closes dudarenok-maker/Castwright#2372` present and outside any code span; release-notes and on-box-acceptance exemptions both declared explicitly rather than silently omitted; F1 declared in the PR body as "Also fixed, found in passing", per the incidental-findings protocol.

### Verdict

**Changes requested.** The design is sound and the guards are unusually well-built — six of nine mutations were caught on the first attempt, including the vacuous-pass case the PR deliberately armed against. But three guards can still pass while guarding nothing, and each is the shape this PR exists to eliminate: a registry with a one-directory escape, a scan-set entry with no scope wiring (asserted as present in a comment this PR adds), and an iteration with no non-empty assertion where the sibling registry has one. All three are mechanical fixes with paired cases; none needs a design pass.

_Modified no tracked file. All mutation work ran in a detached `git worktree` under `$TEMP`, since removed; `git rev-parse HEAD` = `b6f2538c…` and `git status --porcelain` is empty._

dudarenok-maker:
## PR review — pass 2 (head eb8a3920, depth high)

Re-review of the fix commit `eb8a3920` only (`git diff b6f2538c..eb8a3920` — 4 files, +67/−12). Everything pass 1 recorded as solid is not re-listed. All mutation work ran in a detached `git worktree` under `$TEMP`, since removed and pruned.

### Prior findings

- 🟠 **1 — the closed registry had a one-directory escape.** **VERIFIED RESOLVED** (`scripts/tests/review-gate-mechanism.test.mjs:436-442`). Re-planted `.claude/agents/legacy/rogue.md` (`model: opus`, `effort: max`, `git check-ignore` exit 1 → committable) at `eb8a3920`: suite goes **23 pass / 1 fail**, the failing test is the registry one, and the message names `legacy/rogue` — the subpath, not a bare `rogue`. Re-probed at depth 2 (`.claude/agents/sub/deep/ghost.md`) → reported as `sub/deep/ghost`, so `sep`→`/` normalisation holds on Windows. **No false positives introduced**: a directory *named* `weird.md` and a nested non-`.md` file (`sub/notes.txt`) both leave the suite at 24/24 — `d.isFile()` and the `.md` suffix filter are both doing work. No masking either: a nested `legacy/scout.md` alongside a live `scout.md` still trips the reverse direction (the forward direction's `basename` set can't hide it), and a nested-only definition hits `readAgentFrontmatter`'s explicit `missing agent definition <path>` assert rather than a silent pass. `d.parentPath ?? d.path` is dead-but-harmless defence — `package.json` `engines` is `>=22.22.0` and every workflow pins Node 24, so `parentPath` is always present. The failure message's "or move the file out of `.claude/agents/`" now reads correctly.
- 🟠 **2 — `CONTRIBUTING.md` scanned but in no step's scope.** **VERIFIED RESOLVED** (`scripts/verify-cache.mjs:283`, paired case at `scripts/tests/verify-cache.test.mjs:488`). Both halves re-measured at `eb8a3920`: appending a dangling link to `CONTRIBUTING.md` goes red with `CONTRIBUTING.md -> docs/features/this-file-does-not-exist.md (file does not exist)`, so the coverage is real; and `stepTouchedByDiff` + `ci-scope.computeScopes` now both return `true` for `test:hooks` on a `['CONTRIBUTING.md']` diff. **The widening is exactly one step and no more** — I ran `stepTouchedByDiff` across all fifteen steps for that file set: `test:hooks=true`, every other step `false`, and `ci-scope` agrees field-for-field (`shared: false`). No unadvertised cost. The paired test is load-bearing: deleting the `'CONTRIBUTING.md'` line takes `verify-cache.test.mjs` to **97 pass / 1 fail**, on that test, for that reason.
- 🟠 **3 — `FILES` had no non-empty assertion.** **VERIFIED RESOLVED** (`scripts/tests/review-gate-mechanism.test.mjs:539-553`). Re-ran the pass-1 mutation (`export const FILES = [];`): the suite now goes **23 pass / 1 fail** on the new `MIRRORED_FILES …` test, where at `b6f2538c` it was 23/23. **The new assertion is not itself vacuous** — it is an exact four-element `deepEqual`, not `.length > 0`, so it fails on truncation, reorder, and substitution alike, matching the treatment the role table already had. Residual, cleanup only: `node scripts/sync-agent-skills.mjs` with an empty `FILES` still exits **0** printing its normal banner having written nothing. The gate now blocks that state from being committed, so the class is closed where it matters; the script itself still has no runtime non-empty check.
- 🟡 **4 — miscount in the `sync-agent-skills.mjs` header comment.** **VERIFIED RESOLVED** (`scripts/sync-agent-skills.mjs:12-14`, `:47-49`). Re-grepped: `SKILL.md:76`, `:251`, `:265` and `references/findings-triage.md:69` — three plus one, exactly as both comments now state.
- 🟡 **5 — PR body figures wrong.** **VERIFIED RESOLVED**, and the figures are right. Measured fresh on this box at `eb8a3920`, forced (not `[cached]`): `review-gate-mechanism.test.mjs` **24/24**, `verify-cache.test.mjs` **98/98**, `npm run test:hooks` **1412/1412 pass, 27 suites, 0 fail**. The body now reads 1412 / 24 / 98. It also declares the three fixes and the commit that carries them.

### New findings

No Critical or Important findings. Three cleanup-only notes, none re-triggering a round:

- **cleanup — the prose that produced finding 1 was not corrected with the guard.** `.claude/skills/model-routing/SKILL.md:59` still reads "Every `.claude/agents/*.md` file in this repo must have a row here", and `CLAUDE.md:301` repeats it. A single-star glob reads as top-level-only — which is precisely the reading that made `.claude/agents/legacy/rogue.md` look legal. The guard is now `**`; its authoritative statement still says `*`. The code is stricter than the document, so nothing breaks; but this file is the closed-registry's statement of record and it now understates it by one directory level.
- **cleanup — the corrected comment attaches one path literal to four links.** `scripts/sync-agent-skills.mjs:12-13` / `:47-48` now say "links `../model-routing/SKILL.md` three times in SKILL.md plus once more in `references/findings-triage.md`". The count is right (that was the finding); the fourth link is written `../../model-routing/SKILL.md` at `findings-triage.md:69`, one level deeper. A reader grepping the quoted literal finds three, not four — a smaller version of the same audit trap.
- **cleanup — `scripts/sync-agent-skills.mjs` itself**, per finding 3's residual above.

Not findings, checked and clean: `npx eslint` on all four changed files exits 0; the three `prettier --check` warnings on them are byte-identical at `b6f2538c` and `lint` is `eslint . --max-warnings 0` (Prettier is not in that battery), so pre-existing and unrelated; the `verify-cache.mjs:163-170` comment pass 1 flagged as asserting a registration that did not exist is now true; `relative`/`sep` are both used and `basename` is still used at `:331`/`:396`, so no orphaned imports; release-notes and on-box exemptions are unchanged and still correctly declared for a process-only change.

### Verdict

**Approve.** All five pass-1 findings resolved, each re-proven by mutation on this head rather than accepted from the fix commit's own claims — including the two the fixes could plausibly have broken (the recursive scan's false-positive surface, and the scope widening's blast radius), both clean. The three residual notes are cleanup-only and carry no correctness consequence, so per `findings-triage.md` this pass does not re-trigger a round.

_Modified no tracked file. `git rev-parse HEAD` = `eb8a3920…`, `git status --porcelain` empty. The temp worktree's `node_modules` junction was dropped before removal; `git worktree list` confirms no leftover._

PR #2381, opened 2026-08-14
chore(ops): ignore the token-saver skill's .token-saver/ scratch directory

### LINKED ISSUE #2380: ops-46 — gitignore the .token-saver/ skill scratch directory
## What

`.gitignore` has no entry for `.token-saver/`, the scratch directory the
`token-saver` skill writes accepted results into ("save the current result
automatically under `.token-saver/` in the working project").

The skill is installed in the operator's personal library, so it can run
against this repo — and against any Open Engine worktree cut from it.

## Why it matters

- The directory surfaces in `git status`, so it is noise in every
  `git status` an agent or human reads mid-task.
- It can be committed by accident. The skill's own instruction is "do not
  save secrets", but the contents are model output written without review,
  and nothing structurally keeps them out of a commit.
- Open Engine work happens in worktrees, which multiplies the surface: each
  worktree gets its own untracked `.token-saver/`.

The repo already ignores comparable agent scratch output on exactly this
reasoning — `artifacts/` (html-artifacts skill), `.superpowers/`,
`.wiki-sync-cache/`, `server/src/__wire-fixtures__/` ("defense in depth
against an accidental commit").

## Acceptance

- `git check-ignore -v .token-saver/` resolves against a `.gitignore` rule.
- A file created at `.token-saver/anything.md` does not appear in
  `git status --porcelain`.

## Key files

- `.gitignore`

dudarenok-maker:
## Summary

Stops the `token-saver` skill's scratch directory from reaching git or lint.

The skill saves accepted results under `.token-saver/` in whatever project it
runs from. Nothing ignored it, so the directory showed up in `git status` and
could be committed by accident — the contents are model output written without
review. Open Engine worktrees multiply the surface, since each one gets its own
untracked copy.

Two entries, because ignoring it in git alone left it linted:

- **`.gitignore`** — placed next to the existing agent-scratch ignores
  (`artifacts/`, `.superpowers/`, `.wiki-sync-cache/`,
  `server/src/__wire-fixtures__/`) on the same "defense in depth against an
  accidental commit" reasoning those entries already record.
- **`eslint.config.mjs`** — ESLint flat config traverses dot-directories, which
  is why `.claude/`, `.worktrees/` and `server/repro-*.mts` are listed by hand.
  `.token-saver/` is the first of these scratch dirs whose contents can land on
  a linted extension, so unreviewed model output would fail `npm run lint`
  locally. Matched at any depth (`**/`), not root-anchored — the skill writes
  the directory under whatever root it is pointed at.

## Test plan

No automated test is owed: the change adds no runtime code or behaviour that
could regress, and the only seam a test could reach is the config files' own
bytes. Both entries were instead verified live on the branch, each with a
control proving the check could fail.

**`.gitignore`**
- `git check-ignore -v .token-saver/` → `.gitignore:177`. Also resolves for
  `server/.token-saver/` and `src/lib/.token-saver/current.md`, confirming the
  unanchored spelling is the one the skill needs.
- A real file at `.token-saver/current.md` produced **no** `git status
  --porcelain` entry.
- Control: `.claude/skills/token-saver/SKILL.md` is **not** ignored, so
  mirroring the skill into the tracked `.claude/skills/` still works.

**`eslint.config.mjs`** — with probe files at `.token-saver/probe.ts` **and**
`server/.token-saver/probe.ts`:
- Without the entry: `npm run lint` exits **1**, `2 problems`.
- With the entry: `npm run lint` exits **0**.
- The two-warning control confirms both depths were genuinely reachable by
  ESLint, which is what justifies `**/` over a root anchor.

## Notes

- **Release notes skipped deliberately** — internal repo hygiene with no user-
  or operator-visible delta, per the Before-shipping checklist's "no shippable
  delta" carve-out.
- Work was done in a dedicated worktree (`C:\Claude\Projects\wt-token-saver-ignore`),
  with the primary checkout left clean on `main`.

Closes #2380

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,6 +170,12 @@ server/src/__wire-fixtures__/
 # reports, mockups). Local scratch output, never committed.
 artifacts/
 
+# Accepted-result scratch written by the token-saver skill, which saves the
+# current result under `.token-saver/` in whatever project it is run from —
+# including any worktree cut from this repo. Model output written without
+# review; never committed.
+.token-saver/
+
 # Local clone of the separate Castwright.wiki.git repo, used by
 # scripts/sync-wiki.mjs as its push staging area. Recreated fresh on every
 # `npm run wiki:sync` run; never committed.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -51,6 +51,11 @@ export default tseslint.config(
       // Ad-hoc, local-only repro/bisect scripts (e.g. server/repro-attribution.mts).
       // Throwaway debugging probes — git-ignored and not held to the lint bar.
       'server/repro-*.mts',
+      // Accepted-result scratch written by the token-saver skill (see the
+      // matching .gitignore entry). Unreviewed model output that can land on a
+      // linted extension, and the skill writes it under whatever root it is
+      // pointed at — so match at any depth, not just the repo root.
+      '**/.token-saver/',
     ],
   },
 


dudarenok-maker:
## PR review — pass 1 (head 0cb23ff6, depth low)

**Scope:** 1 file, +6/-0 — `.gitignore` only. **Verified independently rather than accepted:** the ignore actually matches (root, nested, and deep paths), the pattern does not over-match the tracked tree, the file's own negation patterns cannot re-include it, line endings are uniform, the linked issue exists and is labelled, all 15 check-runs on this SHA are green, and the two before-shipping claims (no test owed, release notes skipped) were tested against this repo's own rules and precedent. Findings below are split correctness-bug vs cleanup-nit per the reviewer brief.

### 🟡 Minor

**1. `.gitignore:177` — cleanup nit (chore the change made owed).** The entry is git-only, but this repo's precedent pairs a scratch-directory ignore with a **lint** ignore, and that half is missing. ESLint here has no `.gitignore` integration (`eslint.config.mjs` uses a standalone global-ignores object; `grep includeIgnoreFile|gitignore` returns nothing), so a `.gitignore` line buys lint nothing. Flat config also traverses dot-directories — which is precisely why `.claude/` (`eslint.config.mjs:38`) and `.worktrees/` (`:42`) had to be listed there by hand.

Demonstrated, not hypothesised. With a file the skill could plausibly save as its accepted result:

```
# .token-saver/probe.ts
export function greet(name: string) {
  const unusedDraft = name.toUpperCase()
  return `hi ${name}`
}
```

`npx eslint .token-saver --max-warnings 0` (the `--max-warnings 0` of `npm run lint`, `package.json:39`) →

```
.token-saver\probe.ts
  2:9  warning  'unusedDraft' is assigned a value but never used  @typescript-eslint/no-unused-vars
✖ 1 problem (0 errors, 1 warning)
ESLint found too many warnings (maximum: 0).
exit 1
```

ESLint's `--format json` confirms it genuinely processes the file rather than skipping it: `count: 1, filePath: ...\.token-saver\probe.ts`. Trigger condition is narrow but concrete — any `.ts/.tsx/.mts/.cts` under `.token-saver/` (the config's `files: ['**/*.{ts,tsx,mts,cts}']`, `eslint.config.mjs:158`).

The direct precedent is three lines above this PR's own hunk: `.gitignore:157-158` ignores `server/repro-*.mts` and its comment says *"also excluded from lint (see eslint.config.mjs)"* — cashed out at `eslint.config.mjs:51-53`. Same shape: local-only agent/debug scratch, ignored in both places.

Pre-empting the obvious rebuttal — `artifacts/` (cited in the PR body as the sibling precedent) has **no** ESLint entry, but that is not a counter-example: it holds `.html`, which no `files` glob in this config matches. `.wiki-sync-cache/` likewise holds a wiki clone of `.md`. `.token-saver/` is the first of these scratch dirs whose content can land on a linted extension.

Blast radius is local-only and it is *not* a defect in the line that shipped: CI never sees this, because a git-ignored directory is absent from a fresh checkout (consistent with all 15 checks being green here). Exposure is `npm run lint` / full `npm run verify` on a box where the skill has run. Suggested fix: add `'.token-saver/'` to the `eslint.config.mjs` global-ignores block and extend this hunk's comment with the same `(see eslint.config.mjs)` cross-reference `server/repro-*` already carries.

**2. PR body, "Notes" — cleanup nit (documentation).** The body states *"Trivial-bar carve-out invoked for the worktree only"*, i.e. that the worktree was skipped. A worktree was in fact taken — this review ran in `C:\Claude\Projects\wt-token-saver-ignore` on `chore/ops-ignore-token-saver-scratch`. Body and reality disagree about which shortcut was actually used. Costs nothing at merge time, but CLAUDE.md requires the carve-out be *announced* so it can be redirected, and an announcement of a shortcut that was not taken is the inverse failure. Reconcile the sentence with what happened.

### ✅ What is solid

- **Anchoring is right, and right for the non-obvious reason.** Unanchored `.token-saver/` matches at any depth, which the skill needs: `SKILL.md:36-38` tells the agent to pass *"the narrowest useful project root"*, so the directory can land below the repo root. Verified — `git check-ignore -v` resolves `.token-saver/`, `server/.token-saver/`, and `src/lib/.token-saver/current.md` all to `.gitignore:177`. A leading-slash anchor would have silently missed the latter two. Consistent with all 12 sibling dot-directory entries (`.vite/`, `.superpowers/`, `.cast-backups/`, `.wiki-sync-cache/`, …); the only 5 anchored entries in the file (`/companion/`, `/castwright-workspace/`, …) are ones that deliberately want root-only.
- **No over-match.** Control check: `.claude/skills/token-saver/SKILL.md` is **not** ignored — the pattern requires the leading dot, so mirroring the skill into the repo's tracked `.claude/skills/` (kept trackable by the `!.claude/skills/` negation at `.gitignore:34`) still works. Nothing in the tracked tree matches.
- **Cannot be re-included by accident.** All six `!` negations in the file sit at lines 26–99, above this entry, and none targets this path.
- **The stated acceptance reproduces.** Real files at `.token-saver/current.md` and `.token-saver/probe.ts` produced an empty `git status --porcelain`. Claimed resolution to `.gitignore:177` is exact.
- **Comment matches the mechanism it describes.** It tracks `SKILL.md:39-41` (*"save the current result automatically under `.token-saver/` in the working project"*) rather than paraphrasing it loosely, and the worktree clause is correct — a worktree carries its own checkout of this `.gitignore`. Placement, comment style, and the "never committed" phrasing match the neighbouring blocks. LF endings throughout; no CRLF drift in the added lines.
- **"No automated test owed" holds — and writing one would have been worse.** There is no seam here a test could exercise beyond re-asserting the file's own bytes, which is the reviewer brief's shape #3 (a test that cannot fail). Checked that no existing harness enumerates these ignore entries: the only two test files mentioning `.gitignore` (`scripts/tests/review-gate-mechanism.test.mjs`, `scripts/tests/module-graph.test.mjs`) reference it in an assertion *message* about `!.claude/agents/`, not as a registry, so no bookkeeping row is owed there either.
- **Release-notes skip is legitimate and declared.** Internal repo hygiene with no user- or operator-visible delta — squarely the Before-shipping checklist step 5 carve-out, and stated explicitly rather than silently omitted.
- **Issue linkage clean.** `Closes #2380` is plain body text, outside any code span. #2380 is open and labelled `area:ops` / `moscow:should` / `type:chore` — and because it is `type:chore`, no `docs/BACKLOG.md` row is owed, so the untouched BACKLOG is correct, not drift.
- **CI:** all 15 check-runs on `0cb23ff6` are `success`, including `Verify PR body links a GitHub issue` and `Validate PR title against commit convention`.

### Verdict

**No blocking findings — clear to merge.** Both findings are cleanup nits; neither is a correctness bug, so per `findings-triage.md` this pass does **not** trigger a re-review. Finding 1 is a real, reproduced `npm run lint` break rather than a speculative risk, and is the kind of owed chore CLAUDE.md says gets fixed in the same round rather than filed — but it is adjacent to, not a defect in, the line this PR ships, and the merge does not depend on it.

*No tracked file was modified by this pass. Head `0cb23ff6` and a clean `git status --porcelain` were confirmed before and after; the probe files used above were created inside the (git-ignored) `.token-saver/` directory and deleted.*

dudarenok-maker:
## Triage of pass 1 (head `b07b3bb2`)

Both findings were 🟡 Minor cleanup, so per the gate's severity rule no
re-review is triggered. Both are fixed in this PR rather than deferred.

**1. Lint counterpart missing — FIXED in `b07b3bb2`.** Confirmed the
precedent independently before acting: `eslint.config.mjs:51-53` ignores
`server/repro-*.mts` against the matching `.gitignore:157-158` comment, and
`.claude/`/`.worktrees/` are hand-listed for the same dot-dir traversal
reason. Added `'**/.token-saver/'`.

Reproduced the break and its fix with a control, at both depths:

| | probes at `.token-saver/` + `server/.token-saver/` |
|---|---|
| without the entry | `npm run lint` exits **1**, `2 problems (0 errors, 2 warnings)` |
| with the entry | `npm run lint` exits **0** |

The control returning **two** warnings rather than one is what establishes
ESLint reached the nested copy too — which is why the pattern is `**/`-prefixed
rather than root-anchored.

**2. PR body misdescribed the worktree — FIXED.** The body claimed the
trivial-bar carve-out was invoked "for the worktree". That was true when
written and stopped being true: the repo owner directed the work into a proper
worktree (`C:\Claude\Projects\wt-token-saver-ignore`), with the primary
checkout returned to a clean `main`. The stale claim has been removed rather
than reworded, since no carve-out was ultimately taken.

**Gate integrity.** Tree checked before and after the pass: HEAD
`0cb23ff6887c4276c9aca4e36f04d5b4f0f553c9` with empty `git status --porcelain`
both times, and the reviewer's probe files were cleaned up. No delta — the
reviewer's "modified no tracked files" claim verified rather than trusted.

Issue #2382, opened 2026-08-15
[agent instructions][cline][task] Fix the prefer-const lint error blocking the #2106 PR (step 1 rework)


dudarenok-maker:
## What

Step 1 of #2106 (`a1d5ea58`, `fix(server): bound findListenerPid with a deadline`) landed a **lint error**. `npm run lint` is red on `fix/server-2106-respawn-budget`, so the pre-push hook refuses the push and the PR for #2106 cannot be opened at all.

```
C:\Claude\Projects\wt-2106-respawn-budget\server\src\tts\spawn-sidecar.ts
  330:9  error  'timer' is never reassigned. Use 'const' instead  prefer-const

✖ 1 problem (1 error, 0 warnings)
```

This is the **only** problem on the branch. `npm run test:server` (7755 passed, 0 failed) and `npm run typecheck` are both green, and the verify gate (#2379) confirmed the actual behaviour of steps 1–3 is correct — including re-running both mutation checks itself. Nothing about the fix's logic is in question. This is a hygiene defect that blocks delivery.

## Why it happened

`prefer-const` fires on `let x; x = …` when the declaration and its single assignment sit in the same scope. In `findListenerPid` the deadline timer is declared bare and assigned once, at `server/src/tts/spawn-sidecar.ts:330`:

```ts
let settled = false;
let timer: NodeJS.Timeout;          // <- 330: declared bare
const settle = (value: number | null) => {
  if (settled) return;
  settled = true;
  clearTimeout(timer);              // <- closes over timer, called later
  resolve(value);
};
// … handlers …
timer = setTimeout(() => {          // <- assigned exactly once, same scope
  settle(null);
  child.kill();
}, deadlineMs);
```

The split exists because `settle` closes over `timer` while `timer`'s callback calls `settle` — a genuine mutual reference, which is why it was written this way rather than carelessly.

## What to do

Make `npm run lint` green **without changing behaviour**. The mutual reference is only a problem at *definition* time, not at *call* time: `settle` reads `timer` when it runs, by which point the `const` is initialised. So declaring `settle` first and then the timer as a `const` satisfies both the rule and the closure:

```ts
const settle = (value: number | null) => { … clearTimeout(timer); … };
child.stdout?.on('data', …);
child.once('error', () => settle(null));
child.once('exit', () => { … });
const timer = setTimeout(() => { settle(null); child.kill(); }, deadlineMs);
```

[...]

PR #2385, opened 2026-08-15
docs(docs): tell Cline the commit gate takes minutes and its 30s cap cannot clear it

### LINKED ISSUE #2384: ops-46 — tell Cline the commit gate takes minutes and its 30s cap cannot clear it
## What

Record in `.clinerules/cline.md` that a Castwright commit takes 5-7 minutes and that the Cline runtime's 30-second per-command cap can never clear it, so a foreground `git commit` on a `server/**` change is guaranteed to fail.

## Why

Surfaced by #2382. Six identical claim-and-fail cycles and twelve commit attempts were spent before the cap was recognised as the whole story — each cycle correctly re-derived the same one-line fix and then died in the pre-commit hook. Nothing in the repo told Cline that:

- any staged file under `server/` puts the whole server suite in `verify:fast:scoped` (~163 s for `test:server` alone), so minutes is normal and not a hang;
- the 30 s cap is internal to the Cline runtime with no CLI flag (`--timeout` is the whole-run timeout), so no amount of scoping the change gets under it;
- the working route is to launch the commit detached and poll, which makes the cap irrelevant.

`.clinerules/cline.md` is the right home: per `AGENTS.md`, Cline loads that file and `AGENTS.md` and **neither `CLAUDE.md` nor `CONTRIBUTING.md`**, so every Cline session in this repo reads it — not just the one that hit the wall.

## Acceptance

- `.clinerules/cline.md` states the duration, the cap, why retrying cannot work, and carries the detached-and-poll recipe verbatim.
- The recipe puts the commit message in a file (no argument-quoting hazard) and uses an `EXIT=` sentinel, since the launching shell is gone by the time the result is read.
- `--no-verify` is restated as forbidden.

dudarenok-maker:
## Summary

Records in `.clinerules/cline.md` that a commit in this repo can take **10-15 minutes** whenever a test step is in scope, and that the Cline runtime's **30-second per-command cap can never clear it** — so a foreground `git commit` cannot finish, and retrying it never will. The cap is internal to Cline; `-t/--timeout` is the whole-run timeout, not a per-command one.

**What is in scope is decided by `STEPS[]` in `scripts/verify-cache.mjs`**, not by a path prefix — a `src/**` edit runs `test`, a `server/src/**` edit runs `test:server`, and **`openapi.yaml` alone runs both while matching no `server/` path at all**. A docs-only diff matches nothing and commits in seconds.

The entry carries a **detached-and-poll recipe**: launch the commit detached, poll instantly-returning commands until an `EXIT=` sentinel appears, and check the child process as well as the sentinel so a child that dies at parse time is reported rather than waited on forever.

`.clinerules/cline.md` is the right home rather than `CLAUDE.md`: per this repo's own `AGENTS.md`, Cline loads `AGENTS.md` and `.clinerules/*.md` and **neither `CLAUDE.md` nor `CONTRIBUTING.md`**. So this reaches every Cline session in the repo, not just the one that hit the wall.

**Found the expensive way.** #2382 burned six identical claim-and-fail cycles and twelve commit attempts — each correctly re-deriving the same one-line `prefer-const` fix, then dying in the pre-commit hook — before the cap was recognised as the whole story. `--no-verify` was correctly refused every time.

**Also fixed, found in passing:** the `pr-review-gate` depth ladder listed no tier for `docs`, so a `docs(...)` PR's review depth had to be argued by analogy each time. It now sits at `low`, with the distinction between a docs-*type* commit and a docs-only *file set* spelled out — only the second exempts a PR from the gate.

## Test plan

Docs-only in effect; no runtime surface, so no automated test is owed and none was added.

**Release notes: not applicable** — no user- or operator-visible behaviour change (agent-instruction files only), so per CLAUDE.md before-shipping step 5 both release-notes documents are deliberately untouched.

Every claim in the recipe was reproduced under **real Windows PowerShell 5.1** (`5.1.26100.9022`), which is what `Start-Process powershell` launches — not under pwsh 7, where two of these bugs are invisible:

- **BOM, both directions.** 5.1 `Set-Content -Encoding utf8` → first bytes `239,187,191`; `[IO.File]::WriteAllText(..., UTF8Encoding $false)` → `102,105,120`. Against a throwaway repo with a BOM-rejecting `commit-msg` hook: the BOM message is **rejected `EXIT=1`**, the fixed one **commits `EXIT=0`** with `COMMIT_EDITMSG` starting `102,105,120`.
- **The whole recipe, extracted from the file rather than retyped**, run against a repo with a 40-second `pre-commit`: poll #1 reported `alive=True done=False`, the final poll `alive=False done=True` with `EXIT=0`, and the commit landed.
- **Scope and duration figures** read from the repo: `.verify-cache.json` last-green `test:server` **518,751 ms** and `test` **153,320 ms`; `docs/testing/flaky-register.md` records a contended `test:server` at **746.6 s**; `STEPS[]` globs confirm `openapi.yaml` is an `extraFiles` entry for both `test` and `test:server`.
- **The pre-commit hook on this branch's own commits** exercised the scoping from the other side: `[pass] test:hooks` then `[skip]` for the other three legs.

Closes #2384

### DIFF
--- a/.claude/skills/pr-review-gate/SKILL.md
+++ b/.claude/skills/pr-review-gate/SKILL.md
@@ -44,12 +44,16 @@ rather than a new classification:
 
 - **`low`** — single-scope `chore`, `test`, `build`, `ci`, or `docs`
   (mechanical, no user-facing behavior change). `docs` sits here because it
-  ships no runtime surface — but note that "docs-type commit" and "docs-only
-  **file set**" are different tests, and only the second one
-  [exempts](#exemption) a PR from the gate entirely. A `docs(...)` PR touching
-  `.clinerules/` or `.claude/skills/` is reviewed at `low`, not skipped; that
-  gap was found the hard way on PR #2385, whose depth had to be argued by
-  analogy because `docs` appeared in no tier.
+  ships no *product* behavior — **not** because it is inert: a `docs(...)`
+  commit can change instructions agents follow literally (`.clinerules/`,
+  `.claude/skills/`, `AGENTS.md`), and a wrong instruction there is a
+  correctness bug with a blast radius, which is why it is reviewed rather than
+  waved through. Note also that "docs-type commit" and "docs-only **file
+  set**" are different tests, and only the second one [exempts](#exemption) a
+  PR from the gate entirely. PR #2385 was exactly this case — a `docs(...)` PR
+  touching `.clinerules/`, whose depth had to be argued by analogy because
+  `docs` appeared in no tier, and whose review then found six correctness bugs
+  in the instructions themselves.
 - **`medium`** — single-scope `feat` or `fix`.
 - **`high`** — `refactor`, `perf`, or any multi-scope PR (CONTRIBUTING.md's
   own "use sparingly" multi-scope guidance already flags these as
--- a/.claude/skills/pr-review-gate/SKILL.md
+++ b/.claude/skills/pr-review-gate/SKILL.md
@@ -42,18 +42,19 @@ For every non-exempt PR, **review depth** scales with the PR's commit
 type/scope, reusing CONTRIBUTING.md's existing commit-convention vocabulary
 rather than a new classification:
 
-- **`low`** — single-scope `chore`, `test`, `build`, `ci`, or `docs`
-  (mechanical, no user-facing behavior change). `docs` sits here because it
-  ships no *product* behavior — **not** because it is inert: a `docs(...)`
-  commit can change instructions agents follow literally (`.clinerules/`,
-  `.claude/skills/`, `AGENTS.md`), and a wrong instruction there is a
-  correctness bug with a blast radius, which is why it is reviewed rather than
-  waved through. Note also that "docs-type commit" and "docs-only **file
-  set**" are different tests, and only the second one [exempts](#exemption) a
-  PR from the gate entirely. PR #2385 was exactly this case — a `docs(...)` PR
-  touching `.clinerules/`, whose depth had to be argued by analogy because
-  `docs` appeared in no tier, and whose review then found six correctness bugs
-  in the instructions themselves.
+- **`low`** — single-scope `chore`, `test`, `build`, `ci`, or `docs`. Usually
+  mechanical with no user-facing behavior change; `docs` earns the tier by
+  shipping no *product* behavior, **not** by being inert. A `docs(...)` commit
+  can change instructions agents follow literally (`.clinerules/`,
+  `.claude/skills/`, `AGENTS.md`), where a wrong instruction is a correctness
+  bug with a blast radius — so that subclass is reviewed rather than waved
+  through, and `low` buys it a narrower scope, not a lighter standard. Note
+  also that "docs-type commit" and "docs-only **file set**" are different
+  tests, and only the second one [exempts](#exemption) a PR from the gate
+  entirely. PR #2385 was exactly this case — a `docs(...)` PR touching
+  `.clinerules/`, whose depth had to be argued by analogy because `docs`
+  appeared in no tier, and whose three review passes then found **nine**
+  correctness bugs in the instructions themselves.
 - **`medium`** — single-scope `feat` or `fix`.
 - **`high`** — `refactor`, `perf`, or any multi-scope PR (CONTRIBUTING.md's
   own "use sparingly" multi-scope guidance already flags these as
--- a/.claude/skills/pr-review-gate/SKILL.md
+++ b/.claude/skills/pr-review-gate/SKILL.md
@@ -42,8 +42,14 @@ For every non-exempt PR, **review depth** scales with the PR's commit
 type/scope, reusing CONTRIBUTING.md's existing commit-convention vocabulary
 rather than a new classification:
 
-- **`low`** — single-scope `chore`, `test`, `build`, or `ci` (mechanical, no
-  user-facing behavior change).
+- **`low`** — single-scope `chore`, `test`, `build`, `ci`, or `docs`
+  (mechanical, no user-facing behavior change). `docs` sits here because it
+  ships no runtime surface — but note that "docs-type commit" and "docs-only
+  **file set**" are different tests, and only the second one
+  [exempts](#exemption) a PR from the gate entirely. A `docs(...)` PR touching
+  `.clinerules/` or `.claude/skills/` is reviewed at `low`, not skipped; that
+  gap was found the hard way on PR #2385, whose depth had to be argued by
+  analogy because `docs` appeared in no tier.
 - **`medium`** — single-scope `feat` or `fix`.
 - **`high`** — `refactor`, `perf`, or any multi-scope PR (CONTRIBUTING.md's
   own "use sparingly" multi-scope guidance already flags these as
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -131,29 +131,29 @@ any test step in scope takes minutes. So a foreground `git commit` cannot
 finish, and retrying it never will.** The cap is internal to Cline -- there is
 no CLI flag (`-t/--timeout` is the whole-run timeout, not the per-command one).
 
-**What is actually in scope** is decided per step by `STEPS[]` in
-`scripts/verify-cache.mjs` -- that file is the source of truth, not a path
-prefix. Pre-commit runs `verify:fast:scoped` over
-`test:hooks,check:budget-poll,test,test:server`, and a step runs when the
-staged diff matches its own `globs` **or** any of its `extraFiles`. So:
+**So commit detached ALWAYS -- do not try to work out whether this diff is one
+of the slow ones.** That prediction is the bug. Three successive revisions of
+this very paragraph tried to enumerate what puts a step in scope, and all three
+were wrong in the same direction: they under-listed, so an agent read its own
+change as exempt and took the foreground path into the 30-second kill. Detached
+costs nothing when the commit is fast -- the first poll returns immediately --
+and it is the only correct route when it is slow. There is no case where
+predicting first helps.
 
-- a `src/**` edit runs `test` -- not only "server" changes are slow;
-- a `server/src/**` edit runs `test:server`;
-- **`openapi.yaml` alone runs both**, while matching no `server/` path at all;
-- **`RELEASE_NOTES.md` and `docs/release-notes-next.md` are `extraFiles` of
-  `test:hooks`** -- and that pair is exactly what CLAUDE.md's before-shipping
-  step 5 makes nearly every PR touch. So "it's only docs" is **not** a reason
-  to expect a fast commit: `test:hooks` is recorded at 41-58 s, over the cap on
-  its own. `test:hooks` also globs `scripts/**`, `.github/workflows/**`,
-  `.github/actions/**`, `.husky/**`, `pinokio-scripts/**` and
-  `docs/testing/**`;
-- a root `package.json` or lockfile edit is `computeShared` -- it busts **every**
-  leg at once, ~12 minutes, and matches no step's globs, so reading only the
-  globs would tell you nothing runs;
-- a diff touching none of the above skips every leg and commits in seconds.
-  That is a real case, but check it rather than assume it -- the honest test is
-  "does this diff match any step's globs, `extraFiles`, or `computeShared`",
-  not "is this a docs change".
+If you do want to know, `STEPS[]` in `scripts/verify-cache.mjs` is the source
+of truth and the only one; pre-commit runs `verify:fast:scoped` over
+`test:hooks,check:budget-poll,test,test:server`, and a step runs when the staged
+diff matches its `globs`, any of its `extraFiles`, or `computeShared`. Read it
+there rather than trusting a summary -- including this one. What makes a
+summary untrustworthy here is that the surprising members are the whole point:
+`test:hooks` alone reaches `.claude/skills/**`, `.claude/agents/**`,
+`CLAUDE.md`, `CONTRIBUTING.md`, and the `RELEASE_NOTES.md` /
+`docs/release-notes-next.md` pair that CLAUDE.md's before-shipping step 5 makes
+nearly every PR touch -- so editing the very skill files this document tells you
+to edit costs 41-58 s, and "it's only docs" predicts nothing. `openapi.yaml`
+runs both `test` and `test:server` while matching no `server/` path. A root
+`package.json` or lockfile edit is `computeShared` and busts every leg at once,
+~12 minutes, while matching no step's globs at all.
 
 **Budget 10-15 minutes when a test step is in scope, and do not call it hung
 before 20.** Recorded on this repo: `test:server` **518.8 s** and `test`
@@ -227,9 +227,12 @@ elseif ($done)       { Get-Content "$T\commit.log" -Tail 40 }
 else                 { "child gone with no EXIT= -- check whether HEAD moved before assuming nothing happened" }
 ```
 
-**`alive` outranks `done`.** If the process is still up, any sentinel you can
-see belongs to an earlier attempt -- that combination is a stale read, not a
-result.
+**While it is alive, a sentinel you can see is from an earlier attempt** -- keep
+polling rather than reading it as this run's result. Once the process is gone
+the sentinel is yours. The one thing that settles any disagreement is the repo
+itself: `git -C $W log --oneline -1`. Check that before acting on a surprising
+verdict, because a pid can be reused and a log line can be stale, while a
+commit either exists or does not.
 
 `EXIT=0` means the hook passed; confirm with `git -C $W log --oneline -1`.
 Anything else is a real failure in the log and is yours to fix.
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -140,7 +140,20 @@ staged diff matches its own `globs` **or** any of its `extraFiles`. So:
 - a `src/**` edit runs `test` -- not only "server" changes are slow;
 - a `server/src/**` edit runs `test:server`;
 - **`openapi.yaml` alone runs both**, while matching no `server/` path at all;
-- a docs-only diff matches nothing and commits in seconds, every leg `[skip]`ped.
+- **`RELEASE_NOTES.md` and `docs/release-notes-next.md` are `extraFiles` of
+  `test:hooks`** -- and that pair is exactly what CLAUDE.md's before-shipping
+  step 5 makes nearly every PR touch. So "it's only docs" is **not** a reason
+  to expect a fast commit: `test:hooks` is recorded at 41-58 s, over the cap on
+  its own. `test:hooks` also globs `scripts/**`, `.github/workflows/**`,
+  `.github/actions/**`, `.husky/**`, `pinokio-scripts/**` and
+  `docs/testing/**`;
+- a root `package.json` or lockfile edit is `computeShared` -- it busts **every**
+  leg at once, ~12 minutes, and matches no step's globs, so reading only the
+  globs would tell you nothing runs;
+- a diff touching none of the above skips every leg and commits in seconds.
+  That is a real case, but check it rather than assume it -- the honest test is
+  "does this diff match any step's globs, `extraFiles`, or `computeShared`",
+  not "is this a docs change".
 
 **Budget 10-15 minutes when a test step is in scope, and do not call it hung
 before 20.** Recorded on this repo: `test:server` **518.8 s** and `test`
@@ -188,6 +201,11 @@ created and a naive poll waits forever. Same family as the `.ps1` traps under
 "Never do these".
 
 ```powershell
+# Clear the log FIRST. Start-Process returns before the child truncates it, so
+# on a RETRY inside the same $T the very first poll reads the PREVIOUS
+# attempt's sentinel and reports a verdict for a run that is still going.
+Remove-Item "$T\commit.log" -ErrorAction SilentlyContinue
+
 # Quote each path argument: -ArgumentList does NOT quote its elements.
 $p = Start-Process powershell -WindowStyle Hidden -PassThru -ArgumentList @(
   '-NoProfile','-ExecutionPolicy','Bypass','-File',"`"$T\commit.ps1`"",
@@ -204,12 +222,25 @@ $id    = Get-Content "$T\commit.pid"
 $alive = [bool](Get-Process -Id $id -ErrorAction SilentlyContinue)
 $done  = (Test-Path "$T\commit.log") -and (Select-String -Path "$T\commit.log" -Pattern '^EXIT=' -Quiet)
 "alive=$alive done=$done"
-if ($done)                    { Get-Content "$T\commit.log" -Tail 40 }
-elseif (-not $alive)          { "child exited without writing EXIT= -- it died before the commit ran" }
+if ($alive)          { 'still running -- keep polling, and IGNORE any EXIT= you can see' }
+elseif ($done)       { Get-Content "$T\commit.log" -Tail 40 }
+else                 { "child gone with no EXIT= -- check whether HEAD moved before assuming nothing happened" }
 ```
+
+**`alive` outranks `done`.** If the process is still up, any sentinel you can
+see belongs to an earlier attempt -- that combination is a stale read, not a
+result.
 
 `EXIT=0` means the hook passed; confirm with `git -C $W log --oneline -1`.
 Anything else is a real failure in the log and is yours to fix.
+
+**Two ways the log misleads on the way out.** The child redirects with `*>&1`,
+so a *successful* run's tail can still be full of `NativeCommandError` and
+`CategoryInfo` -- that is PowerShell rendering git's stderr progress chatter,
+not a failure; read `EXIT=` and `git log`, not the shape of the text. And if
+the child vanished without a sentinel, **check `git -C $W log --oneline -1`
+before concluding it never ran**: it may have been killed after the commit
+landed, which is exactly what happened on #2382.
 
 **Do not end your turn while it runs.** Polling is active waiting and is
 correct; ending the run is not -- you are headless and nothing will wake you.
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -120,6 +120,38 @@ this file gets fixed. Anything of general value belongs in `CLAUDE.md` instead.
 - **Git-ignored artifacts (`brand/`, `mockups/`, marketing captures) are
   produced in the primary checkout**, never in a worktree: they do not travel
   with the branch and worktree teardown destroys them.
+- **A commit here takes 5-7 minutes, and your runtime kills a command at 30
+  seconds -- so a foreground `git commit` on a `server/**` change can never
+  succeed, however many times you retry.** Pre-commit runs
+  `verify:fast:scoped`, and any staged file under `server/` puts the whole
+  server suite in scope: ~163 s for `test:server` alone. That is the gate
+  working, not a hang. The 30 s cap is internal to the Cline runtime -- there
+  is no CLI flag for it (`--timeout` is the whole-run timeout, not the
+  per-command one) -- and no amount of shrinking the change gets under it,
+  because even a bare `npm run lint` over the repo exceeds it. **Launch the
+  commit detached and poll**; each poll returns instantly, so the cap stops
+  mattering:
+
+  ```powershell
+  $T = $env:TEMP
+  Set-Content "$T\msg.txt" -Encoding utf8 -Value 'fix(scope): subject line'
+  @'
+  git -C <worktree> commit -F "$env:TEMP\msg.txt" *>&1 |
+    Out-File -FilePath "$env:TEMP\commit.log" -Encoding utf8
+  "EXIT=$LASTEXITCODE" | Out-File -FilePath "$env:TEMP\commit.log" -Append -Encoding utf8
+  '@ | Set-Content "$T\commit.ps1" -Encoding utf8
+  $p = Start-Process powershell -ArgumentList '-NoProfile','-ExecutionPolicy','Bypass','-File',"$T\commit.ps1" -WindowStyle Hidden -PassThru
+  $p.Id | Set-Content "$T\commit.pid"
+  # then poll (expect 15-20 polls) until EXIT= appears:
+  (Test-Path "$T\commit.log") -and (Select-String -Path "$T\commit.log" -Pattern '^EXIT=' -Quiet)
+  ```
+
+  The message goes in a file so no argument quoting can bite, and the
+  `EXIT=` sentinel is how you read the result once the shell that launched it
+  is gone. `EXIT=0` means the hook passed; anything else is a real failure in
+  the log and is yours to fix. **Never `--no-verify`.** This cost six identical
+  claim-and-fail cycles and twelve commit attempts on #2382 before anyone
+  noticed the cap was the whole story.
 
 ## Findings are fixed, not filed
 
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -120,38 +120,106 @@ this file gets fixed. Anything of general value belongs in `CLAUDE.md` instead.
 - **Git-ignored artifacts (`brand/`, `mockups/`, marketing captures) are
   produced in the primary checkout**, never in a worktree: they do not travel
   with the branch and worktree teardown destroys them.
-- **A commit here takes 5-7 minutes, and your runtime kills a command at 30
-  seconds -- so a foreground `git commit` on a `server/**` change can never
-  succeed, however many times you retry.** Pre-commit runs
-  `verify:fast:scoped`, and any staged file under `server/` puts the whole
-  server suite in scope: ~163 s for `test:server` alone. That is the gate
-  working, not a hang. The 30 s cap is internal to the Cline runtime -- there
-  is no CLI flag for it (`--timeout` is the whole-run timeout, not the
-  per-command one) -- and no amount of shrinking the change gets under it,
-  because even a bare `npm run lint` over the repo exceeds it. **Launch the
-  commit detached and poll**; each poll returns instantly, so the cap stops
-  mattering:
+- **A commit can take longer than your 30-second command cap, so `git commit`
+  has to be run detached.** Not a silent trap -- the timeout is loud -- but the
+  loop it produces is not. See "Committing when a step is in scope" below.
 
-  ```powershell
-  $T = $env:TEMP
-  Set-Content "$T\msg.txt" -Encoding utf8 -Value 'fix(scope): subject line'
-  @'
-  git -C <worktree> commit -F "$env:TEMP\msg.txt" *>&1 |
-    Out-File -FilePath "$env:TEMP\commit.log" -Encoding utf8
-  "EXIT=$LASTEXITCODE" | Out-File -FilePath "$env:TEMP\commit.log" -Append -Encoding utf8
-  '@ | Set-Content "$T\commit.ps1" -Encoding utf8
-  $p = Start-Process powershell -ArgumentList '-NoProfile','-ExecutionPolicy','Bypass','-File',"$T\commit.ps1" -WindowStyle Hidden -PassThru
-  $p.Id | Set-Content "$T\commit.pid"
-  # then poll (expect 15-20 polls) until EXIT= appears:
-  (Test-Path "$T\commit.log") -and (Select-String -Path "$T\commit.log" -Pattern '^EXIT=' -Quiet)
-  ```
+## Committing when a step is in scope
 
-  The message goes in a file so no argument quoting can bite, and the
-  `EXIT=` sentinel is how you read the result once the shell that launched it
-  is gone. `EXIT=0` means the hook passed; anything else is a real failure in
-  the log and is yours to fix. **Never `--no-verify`.** This cost six identical
-  claim-and-fail cycles and twelve commit attempts on #2382 before anyone
-  noticed the cap was the whole story.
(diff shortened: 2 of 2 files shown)


dudarenok-maker:
## PR review — pass 1 (head d8fa48d0, depth low)

Scope: the single changed file, `.clinerules/cline.md` (+32/−0). Not exempt — the docs-only test is `docs/**` / root `*.md` / `.github/*.md`, and `.clinerules/cline.md` is none of those.

Verified before probing: `package.json` script definitions, `scripts/verify-cache.mjs` `STEPS[]` + `stepTouchedByDiff`/`computeShared`, `scripts/validate-commit-msg.mjs`, `AGENTS.md`, `.verify-cache.json` step durations, `docs/testing/flaky-register.md`, and `cline --help`. The recipe was then executed under real Windows PowerShell 5.1 (`5.1.26100.9022`) against a throwaway git repo under `%TEMP%` — no tracked file touched.

This entry is instructions other agents follow literally, so it is reviewed as code. Two of the findings are reproduced end to end below.

---

### 🔴 Blocking — the recipe writes a commit message with a UTF-8 BOM, and this repo's own `commit-msg` hook rejects it

`.clinerules/cline.md:137` — `Set-Content "$T\msg.txt" -Encoding utf8`.

Windows PowerShell 5.1 `-Encoding utf8` means **UTF-8 with BOM**, and 5.1 is what runs here: the file pins it at `:55`, and `:143` launches `Start-Process powershell` (not `pwsh`).

Reproduced, each step:

```
PS 5.1> Set-Content msg.txt -Encoding utf8 -Value 'fix(scope): subject line'
first bytes: 239,187,191,102,105,120,40,115      # EF BB BF then "fix("

$ git commit -F msg.txt
subject bytes:              239,187,191,102,105,120,...
COMMIT_EDITMSG first bytes: 239,187,191,102,105,120,...

$ node scripts/validate-commit-msg.mjs .git/COMMIT_EDITMSG
Subject:
  ﻿fix(scope): subject line
EXIT=1
```

`git commit -F` does **not** strip the BOM. `validate-commit-msg.mjs` reads with `readFileSync(path, 'utf8')` (Node does not strip it either), `extractSubject` returns the line with a leading `\uFEFF`, and both `CHORE_PATTERN` and `TYPED_PATTERN` are `^`-anchored — so the commit is refused.

Why this is blocking rather than cosmetic: hook order is `pre-commit` → `commit-msg`. The agent pays the **entire** 5–12 minute battery first, then gets rejected on the message. The rejection echoes a subject that looks perfectly valid — the BOM is invisible in the error text — so the agent re-derives the same "correct" subject and retries. That is precisely the six-cycle claim-and-fail loop this entry exists to end, now reproduced by the entry's own remedy at ~12 minutes per cycle.

Direction (not applied): `-Encoding ascii`, or `[IO.File]::WriteAllText($p, $m, (New-Object Text.UTF8Encoding $false))`.

*Correctness bug.*

---

### 🔴 Blocking — the poll reads the **previous** run's `EXIT=` and reports success for a commit that is still running

`.clinerules/cline.md:136-146`. `$env:TEMP\commit.log` is a fixed path, and the recipe never clears it before `Start-Process`. `Start-Process` returns immediately; the child `powershell` needs a few hundred ms to spawn, parse, and truncate the log. The documented poll runs inside that window.

Reproduced verbatim — stale `EXIT=0` left in place from a prior run, then the recipe launched with a command that runs ~5 s and exits 1:

```
poll 1 (immediately after launch) -> done=True
reads: EXIT=0
actual result once finished: EXIT=1
```

The agent reports the commit landed. It had not started, and it failed.

Same defect across parallel sessions: `$env:TEMP` is per-user, not per-worktree, so `commit.ps1`, `msg.txt` and `commit.log` are **shared by every worktree on this box** — and this repo runs parallel worktrees as its normal mode. Two sessions committing concurrently overwrite each other's script and log; one of them reads the other's exit code.

This is the catalogue's shape #6 (success reported while doing nothing) inside a recipe written to stop a false-failure loop.

Direction (not applied): `Remove-Item "$T\commit.log" -ErrorAction SilentlyContinue` before launch, per-run unique filenames, and confirm against **repository state** (`git -C <wt> rev-parse HEAD` moved) rather than trusting a text file.

*Correctness bug.*

---

### 🟠 Significant — "any staged file under `server/`" is wrong in both directions, and the direction it is wrong in reproduces the original bug

`.clinerules/cline.md:126`. The scope map is `STEPS[].inputs.globs` in `scripts/verify-cache.mjs`, read by `stepTouchedByDiff`. `test:server` is `globs: ['server/src/**']` (`:353`) plus `extraFiles` `['server/vitest.config.ts','server/tsconfig.json','openapi.yaml','scripts/tests/fixtures/ffmpeg-version-cases.json','scripts/repair-cast-id-drift.mjs','server/package.json']` plus `server/package-lock.json`.

**Under `server/` but out of scope** (foreground commit is fine, the entry says it can never succeed): `server/tts-sidecar/main.py`, `server/.env.example`, `server/madge-cycles-allowlist.json`, `server/README.md`.

**Not under `server/` but in scope** — this is the damaging half, because the entry's sentence implies non-server commits are safe in the foreground:

- staged `openapi.yaml` alone → it is in `extraFiles` for **both** `test` and `test:server` → both run, ~11 minutes, on a change that touches nothing under `server/`;
- staged `src/store/ui-slice.ts` alone → the `test` step (`globs: ['src/**']`, `:337`) runs; `.verify-cache.json` records it at **153 s**. A frontend-only commit blows a 30 s cap just as hard, and the entry tells the agent it will not;
- root `package.json` / `package-lock.json` → `computeShared` (`:737`) makes it global, every leg runs;
- `scripts/repair-cast-id-drift.mjs` → `test:server` in scope.

The concrete failure: an agent edits one frontend file, reads this bullet, correctly concludes it is not a `server/**` change, commits in the foreground, and lands back in the exact claim-and-fail loop the entry documents.

*Correctness bug.*

### 🟠 Significant — "no amount of shrinking the change gets under it" is false, and this PR's own test plan says so

`.clinerules/cline.md:130-131`.

The same file already states the opposite at `:178-182`: *"it scopes to the STAGED diff, so on an empty or unrelated index every leg reports `(out of scope)` and it exits 0 having run no tests at all."* The PR body's own test plan then confirms it: *"the pre-commit hook ran and correctly reported `[skip]` for all four legs."* Shrinking the change to something outside every leg's globs is exactly how you get under the cap.

The supporting clause is also a non-sequitur: `verify:fast:scoped` is `--steps test:hooks,check:budget-poll,test,test:server` (`package.json`) — **`lint` is not a pre-commit leg at all**. Citing `npm run lint`'s runtime says nothing about whether a smaller change clears the commit gate, and it leaves a reader believing lint runs on commit, which it does not.

*Correctness bug (false statement, and it teaches a false model of the gate).*

### 🟠 Significant — the two timing figures are contradicted by the repo's own records, and the polling guidance is derived from them

`.clinerules/cline.md:123` ("5-7 minutes") and `:127` ("~163 s for `test:server` alone"). Neither is sourced. What this repo records:

| Source | `test:server` |
|---|---|
| `.verify-cache.json`, last green 2026-08-14 | **519 s** |
| `docs/testing/flaky-register.md:27`, observed 2026-08-07 | **`[fail] test:server (exit 1, took 746.6s)`** |

The register figure rules out the obvious defence that 519 s includes retries: `runStepProcess` returns immediately on a genuine red (`res.code !== 0 && !isVitestPoolCrash`), so 746.6 s is one un-retried attempt. For a `server/src/**` diff the in-scope legs sum to roughly 9–13 minutes, not 5–7.

The harm is in what follows from it: `:145` tells the agent to *"expect 15-20 polls"*. An agent that believes the ceiling is 7 minutes and sees the commit still running at minute 9 concludes it hung — and kills a pre-commit battery mid-flight, which is the failure this entry was written to prevent.

*Correctness bug.* Suggest a sourced range rather than a point estimate.

### 🟠 Significant — the here-string, as it appears in the file, is a parse error when copied literally

`.clinerules/cline.md:138-142`. The fence sits inside a list item, so every line — including the closing `'@` at `:142` — carries two leading spaces in the raw file. Cline reads the raw file, not rendered HTML.

Under real PS 5.1:

```
+   '@ | Set-Content "$T\hs.txt"
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
White space is not allowed before the string terminator.
    + FullyQualifiedErrorId : WhitespaceBeforeHereStringFooter
```

This file warns about this exact hazard 80 lines earlier at `:58` — *"A here-string's closing `'@` must sit at column 0"* — and then ships a snippet that violates it as read. Note the fix is not simply unindenting inside the fence (markdown needs the indent for the block to stay in the list item); de-listing the block, or avoiding the here-string, is the way out.

*Correctness bug.*

### 🟠 Significant — the poll has no failure exit, and nothing addresses the run ending mid-commit

`.clinerules/cline.md:143-146`. The only stated bound is "expect 15-20 polls". There is no liveness check and no documented next step when the sentinel never arrives. `$p.Id` is written to `commit.pid` at `:144` and then **never used** — the one artifact that would answer "is it still alive?" is captured and ignored.

Concrete: combine with the finding above. The child `powershell` dies at parse time, so no log is ever created, `Test-Path` stays false, and the agent polls indefinitely with no guidance.

On the hazard detaching introduces: the entry says nothing about the launching agent ending its run while the commit is in flight. The commit outlives the agent. The next session finds either an unexplained commit or a half-finished one — and, per the second 🔴 above, its first poll then reads the orphaned `EXIT=` and reports success for a commit it never made. At minimum the entry should say: do not end the turn or claim the commit until `EXIT=` has been read, and confirm with `git rev-parse HEAD`.

*Correctness bug.*

---

### 🟡 Minor

- **`.clinerules/cline.md:143` — `Start-Process -ArgumentList` does not quote its elements.** They are joined with spaces verbatim, so a `%TEMP%` containing a space (`C:\Users\John Smith\AppData\Local\Temp`) sends `-File C:\Users\John` to the child and it fails. Not reproducible on this box (`dudar`, no space), which is exactly why it will survive review here. *Cleanup.*
- **PR body does not declare the release-notes pair not-applicable.** Before-shipping step 5 requires the skip to be stated explicitly rather than silently omitted; the body covers the test plan but is silent on `docs/release-notes-next.md` / `RELEASE_NOTES.md`. *Cleanup.*
- **Placement and duplication.** The bullet sits under `## Traps that fail silently`, but this trap is loud — the command visibly dies at 30 s. And `:152` restates **Never `--no-verify`** flatly, 60 lines after `:73-88` states it with its two documented narrow exceptions; the file's own charter at `:24-27` is "restates almost nothing". *Cleanup.*
- **Out of diff, surfaced by this round:** the depth ladder in `.claude/skills/pr-review-gate/SKILL.md#review-depth` enumerates `chore`/`test`/`build`/`ci` at `low`, `feat`/`fix` at `medium`, `refactor`/`perf`/multi-scope at `high` — **`docs` appears in no tier**, so a `docs(...)` PR has no ladder entry and the level has to be argued by analogy each time. *Cleanup.*

---

### ✅ What is solid

- **The detached-and-poll mechanism itself works.** Executed under PS 5.1: the log is readable by `Select-String` *while* `Out-File` holds it open, and `$LASTEXITCODE` propagates correctly through the `native | Out-File` pipeline (verified with a deliberate exit 7 → `EXIT=7`). The single-quoted here-string correctly defers `$env:TEMP` and `$LASTEXITCODE` to the child rather than expanding them at authoring time — that is a deliberate, correct choice and easy to get wrong.
- **The `--timeout` claim checks out.** `cline --help` on this box: `-t, --timeout <seconds>  Optional timeout in seconds (default: 0 for no timeout)`, and no per-command timeout flag exists in the option list.
- **The home is right.** `AGENTS.md:12-17` confirms Cline loads `AGENTS.md` + `.clinerules/*.md` and neither `CLAUDE.md` nor `CONTRIBUTING.md`, so `.clinerules/cline.md` does reach every Cline session.
- **`Closes #2384`** is present, plain, outside any code span. Issue is labelled `area:ops` + `type:chore`, which is the right shape and correctly needs no `docs/BACKLOG.md` row.
- **PR title** matches the commit-subject convention (`docs(docs): …`, 84 chars, allowed type and scope).
- **The framing is genuinely valuable** — "that is the gate working, not a hang" is the sentence that would have saved #2382, and putting the message in a file to dodge argument quoting is the right instinct.

---

### Verdict

**Request changes.** Two blocking correctness bugs, both reproduced: the recipe's message file carries a BOM that this repo's `commit-msg` hook rejects *after* the full battery has run, and the poll can report a prior run's `EXIT=0` as this run's success. Four further correctness bugs in the prose — the `server/**` scope rule is wrong in the direction that reproduces the original failure, the "shrinking cannot help" claim is contradicted by this file and by this PR's own test plan, both timing figures are contradicted by the repo's own records, and the snippet does not parse as it appears in the file.

The entry is worth having, and the mechanism at its centre is sound. But it is a file agents execute literally, and as written the happy path ends in a rejected commit after twelve minutes — which is a more expensive version of the loop it documents. Findings only; nothing was applied and no tracked file was modified.

dudarenok-maker:
## PR review — pass 2 (head 05c321a, depth low)

Deltas only. Scope: both changed files — `.clinerules/cline.md` (+100 net) and `.claude/skills/pr-review-gate/SKILL.md` (+8/−2).

Verified before probing: the four `powershell` fences were **extracted programmatically from the raw file** (not retyped) and executed under real Windows PowerShell 5.1 (`5.1.26100.9022`) against throwaway repos under `%TEMP%`; `scripts/verify-cache.mjs`'s `STEPS`, `stepTouchedByDiff` and `computeShared` were **imported and called directly** on realistic staged file-sets; `.verify-cache.json` (both this worktree and the primary checkout), `docs/testing/flaky-register.md:27`, `.husky/pre-commit`, `scripts/is-docs-only-push.mjs`; and `node --test scripts/tests/review-gate-mechanism.test.mjs` (25/25 pass — the `~/.agents/skills/` mirror **is** re-synced, so the `skills:sync` chore the SKILL.md edit owed is discharged).

---

### Pass-1 dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | 🔴 commit message written with a BOM | **VERIFIED RESOLVED** |
| 2 | 🔴 fixed `%TEMP%` log path → stale `EXIT=` | **PARTIALLY RESOLVED** — see 🟠 below |
| 3 | 🟠 "any staged file under `server/`" | **VERIFIED RESOLVED** — but replaced by a new false claim, see 🔴 below |
| 4 | 🟠 "no amount of shrinking…" / `lint` | **VERIFIED RESOLVED** |
| 5 | 🟠 "5-7 minutes" / "~163 s" | **VERIFIED RESOLVED** |
| 6 | 🟠 here-string `'@` indented | **VERIFIED RESOLVED** |
| 7 | 🟠 no failure exit; `commit.pid` unused | **VERIFIED RESOLVED** (one 🟡 residual) |
| 8 | 🟡 ×4 | **ALL RESOLVED** — one introduces a new 🟡 |

**1 — resolved.** `.clinerules/cline.md:168` now uses `[IO.File]::WriteAllText(..., New-Object System.Text.UTF8Encoding $false)`. Measured under 5.1: `msg.txt` first bytes `102,105,120` (`fix`), no `239,187,191`. The whole recipe then committed for real: `EXIT=0`, `[master (root-commit) ea942aa] fix(scope): subject line`.

**3 — resolved as stated.** Every replacement bullet checks out against `stepTouchedByDiff`: `src/store/ui-slice.ts` → `test`; `server/src/routes/analysis.ts` → `test:server`; `openapi.yaml` alone → **`test, test:server`**, matching no `server/` path.

**4 — resolved.** The false claim and the `lint` non-sequitur are gone; `lint` appears nowhere in the section (correct — `verify:fast:scoped` is `test:hooks,check:budget-poll,test,test:server`).

**5 — resolved, figures confirmed at source.** `.verify-cache.json` (primary checkout): `test:server` **518,751 ms = 518.8 s**, `test` **153,320 ms = 153.3 s**. `docs/testing/flaky-register.md:27`: `[fail] test:server (exit 1, took 746.6s)`. All three exact.

**6 — resolved.** All four fences are top-level, not list items. Extracted verbatim, `@'` at line 175 col 0 and `'@` at line 181 col 0; the block parses and runs under 5.1.

**7 — resolved.** `commit.pid` is now read into `$alive` (`:204`), the `elseif (-not $alive)` branch exists (`:208`), and `:214` adds "**Do not end your turn while it runs.**"

**8 — all four resolved.** `-ArgumentList` quoting: I re-ran the recipe with `$env:TEMP` forced to `C:\Users\dudar\AppData\Local\Temp\cw probe two` **and** a worktree path `repo with space` — the child received `-Dir`/`-Worktree` intact and the commit landed. Release-notes not-applicable is now declared in the PR body. The bullet at `:123-125` now says why a loud trap sits under "Traps that fail silently", and `:219` points at the exceptions rather than restating them.

---

### 🔴 Blocking — "a docs-only diff matches nothing and commits in seconds" is false, and the PR body already recorded the counter-evidence

`.clinerules/cline.md:143`.

`test:hooks` — one of the four legs `verify:fast:scoped` runs — has `globs` including `.claude/skills/**`, `.claude/agents/**`, `docs/testing/**`, `.github/workflows/**`, `.husky/**`, and `extraFiles` including `CLAUDE.md`, `CONTRIBUTING.md`, `RELEASE_NOTES.md`, `docs/release-notes-next.md`, `docs/wiki/Advanced-Settings.md` (`scripts/verify-cache.mjs:83-303`).

Calling this repo's own `stepTouchedByDiff` over the four scoped steps:

```
THIS PR commit 05c321a                 -> test:hooks
release-notes pair (every PR owes it)  -> test:hooks
CLAUDE.md one-line fix                 -> test:hooks
CONTRIBUTING.md                        -> test:hooks
flaky register (docs/**)               -> test:hooks
on-box register + live view            -> test:hooks
[SHARED->ALL] root package.json        -> (none, but computeShared forces EVERY leg)
```

Every one of `RELEASE_NOTES.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/release-notes-next.md`, `docs/testing/**` also passes `scripts/is-docs-only-push.mjs`'s test (`docs/**` / root `*.md` / `.github/*.md`) — so this is not an edge case: **a diff the repo itself classifies as docs-only routinely puts `test:hooks` in scope.**

`test:hooks` duration, from the repo's own green records: **40,975 ms (41.0 s)** in this worktree's `.verify-cache.json`, **57,835 ms (57.8 s)** in the primary checkout's. Both clear the 30-second cap.

Concrete failure, and it is the modal commit in this repo: an agent finishes any change, lands the mandatory release-notes pair per CLAUDE.md before-shipping step 5, reads this bullet, correctly concludes "docs-only", commits in the foreground — and the command is killed at 30 s with `test:hooks` mid-run. That is the six-cycle claim-and-fail loop of #2382, re-entered through the sentence written to prevent it. Same defect class as pass-1 finding 3, in a new spelling: wrong in the direction that reproduces the original bug.

**The PR body already contains the disconfirming observation** and the file says the opposite: *"The pre-commit hook on this branch's own commits exercised the scoping from the other side: `[pass] test:hooks` then `[skip]` for the other three legs."* `[pass]` is the label for a leg that **ran**. The author observed `test:hooks` running on this branch's own docs-type commit and then wrote that a docs-only diff has "every leg `[skip]`ped".

Secondary, same passage (`:134-143`): the stated rule — "a step runs when the staged diff matches its own `globs` **or** any of its `extraFiles`" — omits `computeShared` (`verify-cache.mjs:737`, gating `:1168`). A root `package.json`/`package-lock.json`/`.github/actions/**` edit matches **no** step's globs or extraFiles yet forces every leg to run (~12 min: 41 + 1 + 153 + 519 s). A reader applying the rule as written concludes nothing runs, on the diff where the most runs.

*Correctness bug.*

---

### 🟠 Significant — the per-run directory fixes the first attempt only; a retry re-opens the stale sentinel

`.clinerules/cline.md:155-212`.

The first-run half of pass-1 finding 2 **is** fixed: launched fresh, poll 1 reads `alive=True done=False` (verified). But nothing ever clears `commit.log`, and `$T` is stable for the life of the session — so a **second attempt into the same `$T`** reads the first attempt's sentinel. The recipe gives no retry instruction, and after a failure the minimal correct action (re-stage, relaunch) does not regenerate `$T`.

Reproduced under 5.1 — attempt 1 fails, the agent fixes the cause and relaunches:

```
attempt 1 final : alive=False done=True  sentinel='EXIT=1'
attempt 2 poll 1: alive=True  done=True  sentinel='EXIT=1'   <-- previous attempt
attempt 2 final : alive=False done=True  sentinel='EXIT=0'
HEAD after att 2: 753ed5b fix(scope): subject line
```

At `attempt 2 poll 1` the snippet's `if ($done)` branch wins and prints `EXIT=1`. The agent reports the retry failed while it is still running and about to succeed — then edits and re-stages **underneath a live commit**. `alive=True done=True` is the contradiction that would give it away, and the snippet prints it, but nothing in the entry says that combination is impossible-by-construction and means "stale sentinel". The mirror image of pass 1's finding, same root cause: the sentinel is never cleared and `done` is trusted over `alive`.

`:211`'s "confirm with `git -C $W log --oneline -1`" does not rescue it either — on a failed attempt that prints the *previous* commit, which is indistinguishable from "mine didn't land". Pass 1's direction was `rev-parse HEAD` **moved**, i.e. a before/after comparison; a one-shot `log -1` is not one.

*Correctness bug.*

---

### 🟡 Minor

- **`.claude/skills/pr-review-gate/SKILL.md:46-47` — the rationale is falsified by the example in the same sentence.** "`docs` sits here because it ships no runtime surface" — then two lines later it cites a `docs(...)` PR touching `.clinerules/`. `.clinerules/cline.md` **is** runtime surface: it is a recipe agents execute literally, and a `low`-depth pass found two blocking correctness bugs in it. Pass 1 reached them only by explicitly overriding the tier ("This entry is instructions other agents follow literally, so it is reviewed as code"). A reviewer applying the ladder as written, with the reason it gives, calibrates a shallow mechanical pass at exactly the class of PR that needs a code review. The tier may well be right; the stated reason is not, and the reason is what the next reviewer reasons from. The `#exemption` cross-reference and the exemption text itself are consistent — no contradiction there. *Cleanup.*
- **`.clinerules/cline.md:178, 207` — a successful run's log tail is full of error-shaped noise, unwarned.** The added `*>&1` routes the hook's stderr through PowerShell's error rendering, so `Get-Content -Tail 40` on a **passing** commit shows:
  ```
  git : [pre-commit] slow battery starting
  At ...\commit.ps1:3 char:1
  +   + CategoryInfo          : NotSpecified: (...) [], RemoteException
  +   + FullyQualifiedErrorId : NativeCommandError
  ```
  followed by `EXIT=0`. This repo's real pre-commit writes far more to stderr than the probe hook did. In a file whose thesis is "that is the gate working, not a hang", shipping a recipe whose success output reads as a stack trace invites the same misread one layer down. *Cleanup.*
- **`.clinerules/cline.md:208` — the message asserts more than the check can know.** `alive=False, done=False` says only that the child is gone and wrote no sentinel; "it died before the commit ran" is one cause among several (killed after `git commit` returned but before the append; box slept; log write failed). Nothing tells the agent to check whether `HEAD` moved before retrying — and retrying a commit that already landed is a distinct, worse failure. *Cleanup.*

---

### Verdict

**Request changes.** Six of the eight pass-1 findings are verified resolved against the repo and against real PS 5.1 — the BOM fix, the here-string, the `-ArgumentList` quoting, the timing figures, the `commit.pid` liveness check and the `src`/`server/src`/`openapi.yaml` scope bullets all hold up when executed rather than read. Two do not close.

The blocker is a **new** false scope claim in the same passage that fixed the old one, and it is worse-placed than its predecessor: `test:hooks` runs 41–58 s on diffs the repo's own docs-only test calls docs-only — including the release-notes pair every PR in this repo is required to ship — so the sentence tells an agent the foreground commit is safe on precisely the commit shape it is least safe on. The PR body's own `[pass] test:hooks` observation is the disconfirming evidence, written down and then contradicted. The 🟠 is the retry half of pass 1's stale-sentinel bug, still live and reproduced.

Findings only. No tracked file was modified; `git status --porcelain` is empty and HEAD is unchanged at `05c321a`.

dudarenok-maker:
## PR review — pass 3 (head 7020f69, depth low)

Deltas only. Scope: the two files changed since `05c321a` — `.clinerules/cline.md` (+34/−3) and `.claude/skills/pr-review-gate/SKILL.md` (+10/−6).

How the new claims were checked, not read: `scripts/verify-cache.mjs`'s `STEPS`, `stepTouchedByDiff` and `computeShared` were **imported and called** on fifteen realistic staged file-sets (output below); the child script was **extracted from lines 189-193 of the file itself** and run under real Windows PowerShell 5.1 (`5.1.26100.9022`, confirmed in-band) against throwaway repos under `%TEMP%`, once for the `*>&1` claim and once for the retry sequence pass 2 used to break it; `.verify-cache.json` re-read in both checkouts; `node --test scripts/tests/review-gate-mechanism.test.mjs` → **25/25 pass**, so the `skills:sync` chore this head's SKILL.md edit re-owed is discharged (the `~/.agents/skills/` mirror differs from the canonical copy only by the injected `MIRRORED COPY` banner).

---

### Pass-2 dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | 🔴 "a docs-only diff matches nothing and commits in seconds" | **PARTIALLY RESOLVED** — the false absolute is gone, the replacement enumeration is still incomplete in the same direction. See 🟠 below. |
| 2 | 🟠 per-run dir fixed attempt 1 only; retry re-opens the stale sentinel | **VERIFIED RESOLVED** |
| 3 | 🟡 SKILL.md "ships no runtime surface" falsified by its own example | **VERIFIED RESOLVED** (one residual, 🟡 below) |
| 4 | 🟡 `*>&1` makes a successful run's tail read as a stack trace | **VERIFIED RESOLVED** |
| 5 | 🟡 "died before the commit ran" asserts a cause the check cannot know | **VERIFIED RESOLVED** |

**2 — resolved, and the failing sequence now passes.** `:207`'s `Remove-Item` runs before `Start-Process` inside the same fence, so a relaunch into the same `$T` cannot see the prior sentinel. Pass 2's exact reproduction, re-run against this head's extracted child script — attempt 1 red, then relaunch:

```
ATTEMPT 1:  poll@0s  alive=True  done=False sentinel='(no log)'
            final    alive=False done=True  sentinel='EXIT=1'
ATTEMPT 2 (same $T):
            poll@0s  alive=True  done=False sentinel='(no log)'     <-- was done=True 'EXIT=1'
            poll@1s  alive=True  done=False sentinel=''
            final    alive=False done=True  sentinel='EXIT=0'
            HEAD moved: 80d513b fix(scope): subject line
```

`alive` outranking `done` at `:225`/`:230-232` is the correct second layer: it also covers the one window `Remove-Item` cannot close on its own — a relaunch *while a prior child still holds the log open*, where `Remove-Item -ErrorAction SilentlyContinue` fails **silently** (measured: log present before = True, present after = True). That needs the agent to relaunch against the recipe's own "still running -- keep polling", and `alive` catches it if it does, so it is not a finding — noting it only because "does `Remove-Item` close *every* window" was the question asked.

**4 — resolved, and the described behaviour is real, not a plausible story.** Extracted child script, hook exiting **0**, log in full:

```
git : [pre-commit] slow battery starting
At C:\...\commit.ps1:3 char:1
+ git -C $Worktree commit -F (Join-Path $Dir 'msg.txt') *>&1 |
    + CategoryInfo          : NotSpecified: ([pre-commit] slow battery starting:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
[pass] test:hooks (took 41.0s)
[master (root-commit) 0382105] fix(scope): subject line
EXIT=0
```

A successful commit, opening with `NativeCommandError`. The remedy at `:240` ("read `EXIT=` and `git log`, not the shape of the text") is exactly right. One observed nuance the text over-generalises harmlessly: 5.1 wraps only the **first** native stderr line in the ErrorRecord shape and prints the rest plain, so the block sits at the *head* of the log rather than filling the tail — the warning still fires on the right thing.

**5 — resolved.** `:227` now says "check whether HEAD moved before assuming nothing happened", and `:241-243` gives the reason (killed after the commit landed, as on #2382). No cause is asserted.

**3 — resolved.** `SKILL.md:46-51` no longer claims docs is inert; it says `low` is about *product* behavior and names agent-instruction files as a correctness surface. That is true and it is the right correction.

---

### 🟠 Significant — the replacement scope rule is still an enumeration, and the globs it drops include the one this PR's own diff matches

`.clinerules/cline.md:147-149` (and, by dependency, `:153-156`).

The new text says: *"`test:hooks` also globs `scripts/**`, `.github/workflows/**`, `.github/actions/**`, `.husky/**`, `pinokio-scripts/**` and `docs/testing/**`"*. The step's actual `globs` are **nine**, and the three omitted are `.claude/skills/**`, `.claude/agents/**`, and (as a distinct spelling) `scripts/tests/fixtures/**`. The `extraFiles` mention is likewise cut to the release-notes pair, dropping `CLAUDE.md` and `CONTRIBUTING.md` — both real `extraFiles` of the same step.

Calling this repo's own `stepTouchedByDiff` over the four `verify:fast:scoped` steps:

```
THIS PR (both files)     -> test:hooks
.claude/skills only      -> test:hooks          <-- listed nowhere in the bullet
.claude/agents only      -> test:hooks          <-- listed nowhere in the bullet
CLAUDE.md only           -> test:hooks          <-- listed nowhere in the bullet
CONTRIBUTING.md only     -> test:hooks          <-- listed nowhere in the bullet
.clinerules only         -> (none)
release notes pair       -> test:hooks
docs/wiki only           -> test:hooks
docs/features only       -> (none)
AGENTS.md only           -> (none)
openapi.yaml only        -> test, test:server
root package.json        -> (none)   [computeShared -> ALL]
.github/actions only     -> test:hooks   [computeShared -> ALL]
scripts/lib ps1 only     -> (none)
scripts/tests/run.ps1    -> (none)
```

Concrete failure, and it is this repo's routine agent task: an agent edits `.claude/skills/model-routing/SKILL.md` — which **this same file instructs it to do** at `:31-41` ("re-run `npm run skills:sync` after any change under `.claude/skills/pr-review-gate/`") — or edits `CLAUDE.md`. It reads `:143-156`, matches no bullet, applies `:153` ("a diff touching none of the above skips every leg and commits in seconds"), commits in the foreground, and is killed at 30 s with `test:hooks` mid-run. `test:hooks` measured **48,439 ms** in this worktree's `.verify-cache.json` right now and **57,835 ms** in the primary checkout's. That is the #2382 loop, re-entered through the third consecutive revision of the same sentence.

The mitigating clause at `:154-156` ("check it rather than assume it — the honest test is *does this diff match any step's globs, `extraFiles`, or `computeShared`*") is genuinely the right instruction and is why this is 🟠 and not 🔴. But it is one clause after six lines of concrete list, and the failure mode of both prior passes was a reader trusting the list. An enumeration that omits `.claude/skills/**` while the PR carrying it edits `.claude/skills/pr-review-gate/SKILL.md` is the catalogue's shape #2 (enumerate syntax, lose a spelling) — the list is one item short of covering its own diff.

Two smaller errors in the same passage, same root:

- **`scripts/**` over-states.** The real glob is `scripts/**/*.{mjs,cjs,js,mts,cts,ts}` plus `scripts/tests/fixtures/**`, so `scripts/lib/*.ps1` and `scripts/tests/run.ps1` are **out** of scope (probe: `(none)`). Wrong in the harmless direction, but it is the same sentence claiming to state the rule.
- **`.github/actions/**` is filed under the cheap heading.** It is a `test:hooks` glob *and* a `computeShared` member (`verify-cache.mjs:737-742`), so an actions-only edit busts **every** leg (~12 min), not 41-58 s — while `:150` names only "a root `package.json` or lockfile edit" as `computeShared`. Harm is bounded (`test:hooks` is a `test:` step, so `:158`'s "budget 10-15 minutes" still applies), but the reader is told the wrong reason and the wrong number.

The fix is additive and one line: put `.claude/skills/**` and `.claude/agents/**` into the glob list, `CLAUDE.md`/`CONTRIBUTING.md` into the `extraFiles` mention, and `.github/actions/**` into the `computeShared` bullet.

*Correctness bug.*

---

### 🟡 Minor

- **`.claude/skills/pr-review-gate/SKILL.md:55-56` — "found six correctness bugs" undercounts its own record by three.** Pass 1 labelled **seven** findings `*Correctness bug*` (2× 🔴, 5× 🟠); pass 2 labelled **two** more (1× 🔴, 1× 🟠). Nine, mechanically countable in the two comments above this one. "Six" traces to pass 1's verdict sentence ("two blocking … four further"), which was itself one short and predates pass 2 entirely. The figure exists to calibrate a future reviewer against a class of PR the same bullet argues is under-reviewed, so understating it works against the point it is making. *Cleanup.*
- **`.clinerules/cline.md:230-232` — "`alive` outranks `done`" is stated as an absolute and is not one, in two directions.** *(a)* With `:207`'s `Remove-Item` in place, the child writes `EXIT=` and then takes a moment to exit, so `alive=True done=True` legitimately occurs with a **current** sentinel; "any sentinel you can see belongs to an earlier attempt" is false there. Cost: one wasted poll — harmless. *(b)* The reverse is not harmless: `$alive` is `Get-Process -Id` on a **recycled-at-will Windows PID**, so if the pid is reused between the child's exit and the next poll, the snippet prints "still running -- keep polling, and **IGNORE any EXIT= you can see**" over a finished, correct result — and the old `if ($done)` ordering would have read it. Bounded by `:158`'s "do not call it hung before 20 [minutes]", so the ceiling is a wasted 20 minutes rather than a wrong commit. A `Get-Process -Id $id | Where-Object { $_.ProcessName -eq 'powershell' }`, or comparing `StartTime`, closes it. *Cleanup.*
- **`.claude/skills/pr-review-gate/SKILL.md:45-51` — the tier's defining gloss and its new rationale now pull against each other.** The bullet still opens "`low` — … (mechanical, no user-facing behavior change)" and then explains that this subclass changes "instructions agents follow literally" and carries "a correctness bug with a blast radius". Whatever else that is, it is not mechanical. The defensible reading is that the added sentence argues *reviewed vs. exempt*, not *low vs. medium*, and on that reading the bullet is coherent — flagging it because pass 1 reached this PR's two blocking bugs only by explicitly overriding the tier ("this entry is instructions other agents follow literally, so it is reviewed as code"), and the ladder still tells the next reviewer the opposite word. Either scope the gloss so it does not cover the agent-instruction case, or say plainly that the subclass warrants a bump. Tier choice itself is the owner's call, not a review finding. *Cleanup.*

---

### Verdict

**One correctness bug remains; everything else pass 2 raised is verified resolved against the repo and against real PS 5.1.** The retry stale-read is genuinely closed (pass 2's own failing sequence now reads `done=False` on the relaunch poll), the `*>&1` warning describes behaviour I reproduced verbatim on a **passing** commit, the cause-assertion is gone, and the `skills:sync` mirror this head re-owed is in sync (25/25).

The one that does not close is the third revision of the same sentence, and it fails the same way: the scope bullet is a list, the list names six of the step's nine globs, and the ones it drops — `.claude/skills/**` and `.claude/agents/**`, plus `CLAUDE.md`/`CONTRIBUTING.md` among the `extraFiles` — are exactly what a Cline session editing agent instructions touches, **including this PR's own diff**. An agent following `:153` on a `.claude/skills/**` edit gets the 30-second kill this entry exists to prevent.

This is the loop cap, so: the fix is additive, four identifiers, and needs no re-review round — it is verifiable by inspection against the `stepTouchedByDiff` output above. **Merging without it re-ships the pass-1 and pass-2 failure mode for the one file class this repo's agents edit most.** With it, and with the SKILL.md count corrected, nothing blocking remains.

Findings only. No tracked file was modified; probes ran in throwaway repos under `%TEMP%`, `git status --porcelain` is empty and HEAD is unchanged at `7020f69`.

dudarenok-maker:
## Pass-3 fix applied — head `e1dbc9c1`

Pass 3's one open finding is fixed, and the fix is to **delete the prediction rather than extend the list again**.

That sentence had now failed three times in the same direction — each revision enumerated what puts a step in scope, and each under-listed, so an agent could read its own change as exempt and take the foreground path into the 30-second kill. This revision's omissions (`.claude/skills/**`, `.claude/agents/**`, `CLAUDE.md`, `CONTRIBUTING.md`) were the worst of the three, because they cover the very skill files the same document instructs agents to edit.

**The rule is now: commit detached always.** Do not work out whether this diff is one of the slow ones. Detached costs nothing when the commit is fast — the first poll returns immediately — and is the only correct route when it is slow, so predicting first never helps. `STEPS[]` is named as the single source of truth, and the surprising members are given as precisely the reason no summary can be trusted, that one included.

**Verified with the repo's own `stepTouchedByDiff`, not by reading:**

```
IN SCOPE test:hooks  <- .claude/skills only
IN SCOPE test:hooks  <- .claude/agents only
IN SCOPE test:hooks  <- CLAUDE.md only
IN SCOPE test:hooks  <- CONTRIBUTING.md only
IN SCOPE test:hooks  <- release-notes pair
IN SCOPE test:hooks  <- THIS PR diff
IN SCOPE test        <- openapi.yaml only
IN SCOPE test:server <- openapi.yaml only
```

Also fixed: `alive outranks done` was an absolute that is false in the normal end-of-run window and, under pid reuse, told the agent to ignore a correct result — it now says a sentinel is stale only *while* the process is alive, and names `git log` as what settles a disagreement. The depth ladder no longer glosses the tier as "no user-facing behavior change" in the same sentence that explains this subclass is exactly that, and its bug count is corrected from six to **nine**.

## No pass 4

The skill's loop cap is two re-review rounds and pass 3 was the second, so no further gate pass runs — stated here rather than left as a silence. Pass 3's own verdict was *"the fix is four identifiers and needs no pass 4"*; the change went further than that, so the mechanical verification above stands in for the round that is not being run, and it is weaker evidence than an independent pass. Three passes found nine correctness bugs in this file; a fourth might well find a tenth.

Issue #2379, opened 2026-08-14
[agent instructions][claude][verify] Verify the respawn-budget fix (#2106 steps 1-3)


dudarenok-maker:
## What

Verify that steps 1-3 of #2106 landed correctly on `fix/server-2106-respawn-budget`, and produce a **verdict**: pass or fail.

This is the judgement gate. It asks *is it right?* â€” not *does it run?* Read the committed diffs on the branch, not the issue descriptions.

## Where the work happens

**Worktree:** `C:\Claude\Projects\wt-2106-respawn-budget`
**Branch:** `fix/server-2106-respawn-budget`

`cd` into that worktree and work there. **Do not create another worktree, and do not work in `C:\Claude\Projects\Audiobook-Generator`.** Every child of parent #2106 shares this one branch so the deliverable lands as a single PR. Hooks are active there and both `node_modules` are installed, so a bad commit subject or a red in-scope test will be refused â€” that is intended. Never use `--no-verify`.

Commit on that branch when your step is green. Do **not** open a PR â€” that happens after all children land.

Read-only for source. You may run tests. Do **not** fix what you find â€” report it.

## Verify against the parent's acceptance

- A refusal attempt taking longer than `QUICK_DEATH_MS` cannot reset `consecutiveFailures`; the supervisor still reaches `exhaustedEvent() === true` and surfaces `'crashed'`.
- `findListenerPid` is bounded and resolves `null` on expiry, killing its child.
- `QUICK_DEATH_MS`'s doc comment no longer describes a use the code does not have.
- `POST /api/sidecar/restart` -> `resetAndRespawn()` recovery still works.
- Release notes moved in both documents.

## Check these specifically

1. **Did the mutation checks actually happen?** Steps 1 and 2 were each required to mutate the fix and report the observed failure output. If a receipt asserts it without showing output, treat the test as unproven and **re-run the mutation yourself**. A test that cannot fail is the failure mode this gate exists to catch.
2. **Does the refusal path really pass `false`**, rather than a value that merely happens to be false today?
3. **Is `lived` still consulted anywhere it should not be?**
4. `npm run test:server` green, `npm run typecheck` green.

## Verdict

- **pass** -> close this issue with the verdict recorded.
[...]

Issue #2037, opened 2026-07-31
Sidecar supervisor gives up permanently when the exiting child still holds :9000 — TTS stays down until a manual server restart


dudarenok-maker:
Observed on the box 2026-08-01 while tidying up after the fs-38 Wave 3 acceptance run. **The sidecar died and the supervisor permanently declined to replace it**, leaving TTS unavailable until the server was restarted by hand.

## What happened

```
08:21:59.726  [sidecar] child exited (code=1, signal=null)
08:21:59.726  [sidecar] supervisor: child exited (code=1 signal=null); respawning in 2000ms (attempt 1/5).
08:22:03.770  [sidecar] something is listening on :9000 but it does not look like our sidecar
              — NOT touching it. TTS may be unavailable until the port is freed.
```

…and nothing after that. Minutes later:

- **no `uvicorn main:app` process existed at all**
- **port 9000 was free**
- `GET http://127.0.0.1:9000/health` → connection refused
- the server was still healthy on :8443 and never tried again

Recovery required killing and relaunching the server process.

## The race

The exiting sidecar still held its listening socket during the supervisor's 2,000 ms respawn delay. When the supervisor woke, it probed :9000, saw *a* listener that failed its identity check (the dying process no longer answering `/health` correctly), and took the deliberate "don't touch a foreign listener" path.

That guard is correct in itself — it exists so the supervisor never kills an unrelated process that happens to own the port. The defect is that it is **terminal**: it consumes the respawn attempt and never re-checks. The port frees a moment later and nothing notices. Five attempts were budgeted; the observable outcome was zero further attempts.

## Why it matters

The sidecar is *designed* to exit and be replaced — that is the documented containment for the CPU-codec committed plateau (#399's close: "recycling is hereby demoted from load-bearing to safety net"; #1374 shipped the GPU-codec knob but on-box acceptance found it unusable on an 8 GB card at batch=32, so the plateau and its recycles are still with us). A recycle path whose replacement step can silently give up turns a routine, expected event into an outage.

This is reachable on any normal long book, not just after a manual `/restart`.

## Trigger in this instance

The exit followed a `POST /api/sidecar/restart` issued against a sidecar that was already leak-saturated (`rss=15,016 MB`, `committed=24,744 MB`, above the 20,000 MB adopt ceiling). The supervisor had already flagged it twice:

```
07:52:31  supervisor: adopted sidecar on :9000 is leak-saturated (committed 20896MB ≥ 20000MB)
[...]

Issue #2383, opened 2026-08-15
[agent instructions][claude][verify] Verify the lint rework (#2106 step 1 rework, #2382)


dudarenok-maker:
## What

Verify that the lint rework in #2382 landed correctly on `fix/server-2106-respawn-budget`, and produce a **verdict**: pass or fail.

This is the re-run of the judgement gate #2379, which returned **fail**. It is deliberately narrower than #2379 was: the behaviour of #2106 steps 1–3 has already been verified in full and is **not** re-litigated here unless the rework disturbed it.

Read the committed diffs on the branch, not the issue descriptions.

## What #2379 already established — do not redo this

#2379 verified steps 1–3 against the parent's acceptance and found the behaviour **correct**:

- The refusal path passes a literal `false` for `freshIncident` — not a value that merely happens to be false today.
- The `lived` variable is gone from source entirely; only the exit path derives child lifetime, which is the one place `QUICK_DEATH_MS` legitimately means that.
- `scheduleRespawnAttempt` has exactly two call sites, and `freshIncident` is a required parameter, so a third call site cannot compile without deciding the question.
- `findListenerPid` is bounded, resolves `null` on expiry, and kills its probe child.
- The `QUICK_DEATH_MS` doc comment matches what the code actually does.
- `POST /api/sidecar/restart` → `resetAndRespawn()` is intact; `exhaustedEvent()` → `supervisorExhausted` → `'crashed'` in `server/src/routes/models-status.ts:33`.
- Release notes moved in **both** documents.
- **Both mutation checks were re-run by the gate itself**, not taken on trust. Restoring the pre-fix `lived`-derived reset failed exactly the step-2 test (`AssertionError: expected false to be true`, 37 others passing); deleting the `setTimeout` failed exactly the step-1 deadline test (hung to the 30s vitest timeout, 38 others passing).

#2379 failed on **one** thing and one thing only: `npm run lint` was red (`prefer-const` on `let timer` at `spawn-sidecar.ts:330`), which the pre-push hook refuses, so the PR could not be opened.

## Check these specifically

1. **Is `npm run lint` green?** 0 errors, 0 warnings — the script runs with `--max-warnings 0`.
2. **Did the rework change behaviour?** It must not. Confirm the deadline still fires, still kills the probe child, and still resolves `null`; and that a probe exiting before the deadline still resolves its pid and is not killed.
[...]

PR #2398, opened 2026-08-15
fix(server): bound the sidecar respawn budget on the refusal path

### LINKED ISSUE #2106: srv — findListenerPid has no timeout, so a slow refusal resets the respawn budget forever
## What

`findListenerPid` (`server/src/tts/spawn-sidecar.ts:297`) has **no timeout**. It resolves only when its spawned child (`powershell.exe … Get-NetTCPConnection` on Windows, `lsof`/`ss` elsewhere) emits `exit` or `error`. If that child hangs, the promise never settles.

That was harmless while the function was called only on the stale-sidecar replacement path. PR #2101 (#2037's fix) added a call on the **not-ours refusal path** (`spawn-sidecar.ts:619`) — the hot path of the outage #2037 describes — where it now runs on every refused respawn attempt.

## Why it matters: the retry budget can silently become unbounded

`scheduleRespawnAttempt` (`sidecar-supervisor.ts:252-253`) reuses the pre-existing quick-death reset:

```ts
const lived = nowFn() - lastSpawnAt;
if (lived >= QUICK_DEATH_MS) consecutiveFailures = 0; // ran a while → fresh incident.
```

On the **refusal** path `lastSpawnAt` was set at the top of the *same* `spawnOnce` call, so `lived` measures **attempt latency**, not child lifetime. If a single refusal attempt ever takes longer than `QUICK_DEATH_MS` (30 s), `consecutiveFailures` resets to 0 on every attempt.

Consequences, all of which invert #2037's fix one level down:

- `exhaustedEvent()` never flips, so `/api/setup/models-status` never reports `'crashed'` (`models-status.ts:92` → `:33`), and neither does `setup-readiness.ts:125` or `setup-diagnosis.ts:81`;
- the supervisor re-probes the port forever, spawning a `powershell.exe` per attempt;
- this is precisely the **unbounded retry loop** #2037's implementation brief (decision D1) explicitly forbade — "a genuinely foreign process on :9000 is permanent, and an unbounded retry would probe that port forever."

Today's headroom is real but not large, and none of it is enforced: the `/health` probe is 2 s (`:111`), `waitForPortFree` is 5 s (`:113`), plus a `killTree` — but the pid lookup itself is unbounded.

## Failure scenario

A foreign process camps `:9000` on a heavily contended box (this repo's own parallel-agent workflow routinely puts several Node batteries on one machine; wedged processes at zero CPU have been observed repeatedly). Listener enumeration takes >30 s. Every refusal resets the counter, so TTS never reports as crashed, no operator surface ever shows the problem, and the supervisor probes indefinitely.

## Secondary defect: a stale doc comment

`QUICK_DEATH_MS`'s doc comment (`sidecar-supervisor.ts:36-40`) still describes it purely as a child-lifetime threshold:

> A child that dies faster than this counts toward the crash-loop cap; one that lived longer is treated as a fresh incident…

Since #2101 it also gates attempt latency on the refusal path. Whatever fix is chosen, that comment needs to cover both uses or the next reader re-derives the wrong invariant.

## Options (a real choice, which is why this is filed rather than folded into #2101)

1. **Bound `findListenerPid`.** Give it an `AbortSignal.timeout(...)` / kill-the-child deadline and resolve `null` on expiry — `null` already means "can't tell", and every caller handles it. Narrowest fix, but only bounds *this* call; any future slow step on the refusal path reopens the same hole.
2. **Measure attempt latency separately from child lifetime.** Have the refusal path pass its own "did this attempt represent a fresh incident" signal instead of reusing `lived`, so the quick-death reset only ever means what its name says. Structurally correct; touches the shared helper #2101 just factored out.
3. **Both.** (1) is cheap insurance regardless; (2) is the invariant fix.

Option 2 is the one that actually restores D1's guarantee. Option 1 alone leaves the reset semantics wrong.

## Acceptance

- A refusal attempt that takes longer than `QUICK_DEATH_MS` cannot reset `consecutiveFailures`; the supervisor still reaches `exhaustedEvent() === true` and surfaces `'crashed'`.
- Regression test with an injected `nowFn`/`delayFn` and a slow `spawnFn` proving the counter advances monotonically across slow refusals. Mutate the fix line to confirm the test can fail.
- `QUICK_DEATH_MS`'s doc comment describes both uses, or the second use is removed.
- `POST /api/sidecar/restart` → `resetAndRespawn()` recovery still works.

## Key files

- `server/src/tts/spawn-sidecar.ts:290-305` — `findListenerPid`, no timeout
- `server/src/tts/spawn-sidecar.ts:619` — the D2 call site added by #2101, on the refusal hot path
- `server/src/tts/sidecar-supervisor.ts:240-276` — `scheduleRespawnAttempt`
- `server/src/tts/sidecar-supervisor.ts:36-43` — `QUICK_DEATH_MS` and its now-incomplete doc comment
- `server/src/tts/sidecar-supervisor.test.ts` — existing seams (`nowFn`, `delayFn`, `spawnFn`) make this unit-testable

## Benefit

**Technical** — keeps #2037's central guarantee (a supervisor that gives up does so *visibly*) from being silently voided by a slow port probe, on the exact path #2037 was filed about.

Found by the independent pre-merge review of PR #2101 (finding 2). Refs #2037.

dudarenok-maker:
## Summary

Bounds the sidecar respawn budget on the refusal path: `scheduleRespawnAttempt` now requires an explicit `freshIncident` flag instead of deriving it from stale `lived` state, and `findListenerPid`'s deadline timer is bounded with a `setTimeout`/`clearTimeout` pair. This is the rework of the `prefer-const` lint failure that blocked the original PR (#2382), verified narrowly against #2379's already-established behaviour. Closes #2106

## Also fixed, found by the review gate (pass 1)

All seven findings from the pass-1 review were fixed in this round rather than filed, one dispatched fix agent per finding. The pass comment is on this PR.

- **The exit path had the same defect one level down** (`9d8cf8c8`, 🟠). `lastSpawnAt` was stamped at `spawnOnce`'s first line — before `buildOpts()` and `spawnFn()` — so the exit path's `lived` measured attempt latency plus child lifetime while a new comment certified it as child lifetime alone. A sidecar dying ~18s into model load, repeatedly, read as `lived >= QUICK_DEATH_MS` every iteration and reset the budget forever: #2106's own failure mode, on the path this PR declared clean. Now stamped only once `spawnFn` returns a real handle. Regression test goes red on the exhaustion assertion (not by timeout), verified by mutation.
- **The deadline's kill missed the process it exists to reap on POSIX** (`def0e103`). The probe ran as `sh -c 'lsof … | head -n1'`; a pipeline keeps `sh` resident as the parent, so `child.kill()` signalled `sh` and orphaned the hung `lsof`. The pipe was redundant — the existing `split(/\s+/)` parse already takes the first PID — so it is now a single simple command that `sh -c` execs. Paired tests pin both the no-pipe shape and multi-PID parsing.
- **A deadline expiry was indistinguishable from a parse miss** (`5991141`, `fe296214`). Both returned `null` and produced the same "could not identify the PID … restart manually" line, which is right for a missing `lsof` and misleading for a timeout. `findListenerPid` takes an optional `onDeadlineExpiry`; the stale-replace site now branches the whole sentence, advice included. The shipped `LISTENER_PID_DEADLINE_MS = 5000` also gains direct coverage — every prior test passed an explicit `1000`, so a typo'd constant kept the suite green.
- **`findListenerPid` had lost its JSDoc** (`1335146`). The new constant's doc block was inserted between the function's doc and the function, and the null-return enumeration did not name the deadline case this PR added.
- **On-box acceptance recorded as A46** (`d4ace218`, `749001d2`), across the register, the live view, and `check:onbox-register`. No per-feature run sheet exists for the sidecar supervisor, so none was invented. The correction commit also fixes four factual errors the checker cannot see — a log format that does not exist, `GET /api/models/status` for `GET /api/setup/models-status` (**also corrected in A34**, which carried the same error), `netstat` for the real PowerShell `Get-NetTCPConnection` probe, and one criterion that passed vacuously when the deadline never fired.
- **The user-facing release note described a state no release ever had** (`76a31f35`). "Used to keep quietly retrying underneath" was true only on `main`, between #2037 and this PR. A 1.14.0 reader is upgrading from the opposite behaviour — the service reported ready while nothing was running — so the bullet now describes the combined #2037 + #2106 delta, which also closes the gap where #2037 shipped with no user-facing line.
- **Doubled blank lines** this PR added to both release-notes files (`1ca9cc0b`).
- **The technical register entry had gone stale** against the four changes above (`48bbf61b`) — it framed the budget defect as refusal-path-only and credited the deadline with killing a probe it was not reaching on POSIX.

## Test plan

- [x] `npm run lint` — 0 errors, 0 warnings (`--max-warnings 0`)
- [x] `npm run typecheck` — green (root + server)
- [x] `npm run test:server` — 528 files / 7762 tests passed, 1 file / 9 skipped (up from 7755; the review round added 7 tests)
- [x] `npm run verify:fast:branch` via the pre-push hook — lint, typecheck, config:check, test:hooks, test, test:server, build all green
- [x] `npm run check:onbox-register` — OK, register and live view agree
- [x] Mutation checks re-run per fix, not taken on trust. The 🟠 fix's regression test was independently mutated by the coordinating session: reverting the stamp fails on `expected false to be true` at the exhaustion assertion in 1.1s — an assertion failure, not a timeout.
- [ ] **A46 on-box acceptance owed** — the 5000 ms headroom under real contention, budget exhaustion against a genuine foreign listener on `:9000`, and the POSIX deadline kill. Recorded, not run.

Closes #2106


dudarenok-maker:
## PR review — pass 1 (head 2e78206a, depth high)

Depth `high` per the ladder: two scopes (`fix(server)` ×4 + `docs(docs)` ×1). Reviewed in `C:\Claude\Projects\wt-2106-respawn-budget` at `2e78206a`; both checkouts clean before and after.

---

### 🟠 Correctness bugs

**1. `server/src/tts/sidecar-supervisor.ts:472-476` — the new comment certifies an invariant the code does not hold. The exit path's `lived` is still attempt latency + child lifetime, so #2106's failure mode stays reachable there.**

`lastSpawnAt` is stamped on `spawnOnce`'s *first* line (`:290`), before `await buildOpts()` and before `await spawnFn(...)`. Everything `spawnSidecar` does before the child actually starts is therefore counted into `nowFn() - lastSpawnAt` at `:476`: `probeListening` (250 ms), `probeSidecarHealth` (2 s), `findListenerPid` (now up to 5 s), `killTree`, `waitForPortFree` (5 s) — up to ~12 s on the stale-sidecar-replacement branch (`spawn-sidecar.ts:607-697`).

Failure scenario: a sidecar that OOM-dies or CUDA-poisons (code 42) ~18 s into its model load, repeatedly, on a box where each respawn has to run the stale-replace sequence. `lived = ~12 s + 18 s = 30 s ≥ QUICK_DEATH_MS` on every iteration → `freshIncident = true` → `consecutiveFailures = 0` → `exhaustedEvent()` never flips → `/api/setup/models-status` never reports `'crashed'` (`models-status.ts:33`) and the supervisor respawns forever. That is verbatim #2106's failure mode. Only code-43 exits are protected, by the separate `RESTART43_STREAK_*` counter — whose own comment at `:47-52` names this exact hazard ("a structurally-too-small device assignment that loads fine for 35s and then dies every time would never trip the give-up branch"). Code 42, OOM SIGKILL and code 1 are not.

The *behaviour* predates this PR (bounding `findListenerPid` actually shrinks the window from unbounded to ~12 s). What is new — and why this is a finding against this PR rather than a note — is the comment the PR adds at `:472-475`: *"lastSpawnAt was set when this child spawned, so `lived` here is child lifetime — the only place QUICK_DEATH_MS still means that."* That sentence is false as written, and it is the sentence that will stop the next reader from finding this. #2106's option 2 was chosen so *"the quick-death reset only ever means what its name says"*; it now means what its name says on one of the two paths, and the diff asserts it means it on both.

Fix is one line and behaviour-preserving: `lastSpawnAt` has exactly one writer (`:290`) and exactly one reader (`:476`), and the refusal path no longer reads it at all — so move the stamp to immediately after `handle = await spawnFn({...})` resolves non-null (`:306-317`). Adopt and autoStart-off paths never reach `onChildExit`, so nothing else changes. Paired test: fake clock, a `spawnFn` that burns 12 s before returning a handle, a child that exits 18 s later, repeated past the cap → assert `exhaustedEvent()` is true. Then the comment becomes true.

---

### 🟡 Cleanup nits

**2. `server/src/tts/spawn-sidecar.ts:291-302` — the new constant's JSDoc is inserted *between* `findListenerPid`'s own JSDoc and the function, orphaning the function's contract doc.**

Order on disk is now: `/** Best-effort cross-platform "which PID is listening on this port" … */` (`:291-296`) → `/** Bound on how long findListenerPid may wait … */` (`:297-301`) → `export const LISTENER_PID_DEADLINE_MS` (`:302`) → `export function findListenerPid` (`:304`). TypeScript attaches only the *last* preceding block, so the constant keeps the right doc and **`findListenerPid` now has none** — hovering it in an editor, or reading generated `.d.ts`, shows no doc for the function whose "returns null when it can't tell, so the caller leaves the process in place" contract is the load-bearing part of this whole change. Secondary: that orphaned block enumerates why `null` comes back — "(no tool, parse miss, error)" — and this PR adds a fourth reason it does not list. Move the const + its doc above the function's doc block, and add the deadline to the enumeration.

**3. `server/src/tts/spawn-sidecar.ts:344-347` — on POSIX the deadline kills the shell, not the process that hung.**

The non-win32 probe is `spawn('sh', ['-c', 'lsof -ti tcp:<port> -sTCP:LISTEN 2>/dev/null | head -n1'])` (`:320`). A pipeline forces `sh` to fork, so `lsof` and `head` are grandchildren, and `spawn` is called without `detached`, so `child.kill()` at `:346` signals `sh` alone. In the exact case the deadline exists for — `lsof` wedged on a hung mount — the wedged process is reparented to init and survives, one per refused attempt (up to the cap, plus one per stale-replace attempt), while the doc at `:299-301` says "we kill the probe child". This repo documents this precise trap 80 lines away, in `killTree`'s comment at `:420-421` ("A plain `process.kill(pid)` only reaps the powershell shell, orphaning the python grandchild"). Windows is unaffected (a single `powershell.exe`, no grandchild).

Cheapest fix: drop the pipeline. `| head -n1` is already redundant — the parse at `:341` does `out.trim().split(/\s+/)[0]`, i.e. it already takes the first token. `sh -c 'lsof -ti tcp:<port> -sTCP:LISTEN 2>/dev/null'` is a single command, which sh `exec`s in place, so there is no grandchild left to orphan.

**4. `server/src/tts/spawn-sidecar.ts:302` + `:680-688` — the 5 s value is unmeasured, never exercised at its default, and its expiry is silently indistinguishable from a parse miss on the one call site where it changes the outcome.**

Two call sites, two consequences. At `:661` (the refusal path #2106 is actually about) a `null` is log-only, and the code says so. At `:680` (stale-sidecar replacement) a `null` is a **refusal**: `onSpawnRefused('could not identify the PID on :<port> to replace the stale sidecar')` plus an operator-facing *"Restart the sidecar manually"* — and, thanks to this PR's step 2, that refusal now accretes the budget monotonically to `'crashed'` in 5 attempts. So a probe that is merely *slow* is reported to the operator as *unidentifiable*, with advice that is wrong for a timeout (a retry would have worked), and nothing in the log tells the two `null`s apart. The issue's option-1 rationale — "`null` already means 'can't tell', and every caller handles it" — is true of the action taken but not of the consequence at `:681`.

I measured the real probe rather than speculating, on this box, warm, 5 runs of the exact `:317` command against `:9000`: **1119 / 767 / 777 / 780 / 765 ms**. That is ~4-6× headroom, which is why this is a nit and not a correctness bug. But the margin is nowhere on record, and neither new test exercises the default — both pass `deadlineMs` explicitly as `1000` (`spawn-sidecar.test.ts:1250`, `:1265`), so `LISTENER_PID_DEADLINE_MS` could be changed to `50` and the suite would stay green. Minimum ask: log the expiry (`[sidecar] listener-pid probe on :<port> exceeded <n>ms — treating as unknown`) so the two `null` shapes are separable in a real incident, and pin the default in a test.

**5. House gate — on-box acceptance: a row looks owed and none was added.**

A34 (`docs/testing/onbox-acceptance-register.md:2014`) exists for this exact subsystem and this exact class of question: *"Unit tests fully pin the fix's logic. What no unit test can reach is … whether the fix's backoff schedule outlasts that window on real hardware — the implementation brief deliberately declined to widen the backoff without a real measurement behind it (D1)."* This PR introduces a new hardware-dependent timing constant (5 s against a real `Get-NetTCPConnection` on a real, contended Windows box) whose expiry, per finding 4, converts a stale-sidecar replacement into a refusal that now runs the budget to `'crashed'`. Neither a new row nor an amendment to A34 is in the diff, and the live view / run sheet are untouched. Per CLAUDE.md step 3 the *running* never blocks a merge but the *recording* does. My measurement above is partial evidence, not a discharge — it was taken on an idle box, not under the contention #2106's own scenario describes. If the author's judgement is that the unit tests plus that measurement close it, say so explicitly in the PR body rather than leaving the gate silent.

**6. `RELEASE_NOTES.md:156` — the user-facing line's premise describes a state no released version ever had.**

> "…Castwright **used to keep quietly retrying underneath** — a book that needed it could sit there looking busy while nothing ever surfaced that something was wrong."

In the last released version (1.14.0) the supervisor did the opposite: `spawnOnce` set `isRecycling = false` unconditionally on every refusal, so it stopped retrying entirely and reported TTS **ready** while nothing was listening. That is #2037, whose technical entry is at `docs/release-notes-next.md:216` and which has **no** user-facing line anywhere in `RELEASE_NOTES.md`'s 1.15.0 section (`grep -n "2037\|holding its port\|manual server restart" RELEASE_NOTES.md` → nothing). The "quietly retrying forever" state exists only on `main` between #2037 and this PR — a development-diary state, which is exactly what the user-facing file is not supposed to describe. Net effect on a 1.15.0 reader: they are told their 1.14.0 install was retrying, when in fact it was reporting a nonexistent sidecar as ready — the opposite failure. Make this bullet the shipped end state for the pair instead: 1.14.0 could report the speech service as ready when it wasn't there and never bring it back; 1.15.0 retries on a bounded budget and reports it as failed once that budget is spent.

**7. `RELEASE_NOTES.md:156` and `docs/release-notes-next.md:348` — formatting, and a deferral that shouldn't have been one.**

Both new bullets are appended with no blank line before them, while every other bullet in both sections is blank-line separated, and both leave a doubled blank line before the following `# Castwright 1.14.0` / `---`. Rendering is unaffected (the lists are already loose) and `npm run lint` is `eslint . --max-warnings 0`, which never reads Markdown — but `prettier --write .` will rewrite both, so the next unrelated PR that runs `npm run format` picks up spurious hunks. Flagged because #2379's own verdict already caught this and explicitly deferred it ("fold it in only if #2382 is touching those files anyway, which it should not be"), and #2382/#2383 then didn't; the PR body declares no incidental fixes. Under CLAUDE.md's incidental-findings rule that is the shape the rule exists to close.

---

### Checked and clean

Recorded so a later pass doesn't re-litigate these.

- **Both new tests pass, and not by retry.** `vitest run src/tts/sidecar-supervisor.test.ts src/tts/spawn-sidecar.test.ts --retry=0` → 2 files / 77 tests passed. `--retry=0` matters here: `server/vitest.config.ts` sets `retry: 1` suite-wide.
- **The class is closed at compile time, not by convention.** `scheduleRespawnAttempt` has exactly two call sites (`:328`, `:477`) and `freshIncident` is a **required** parameter positioned *before* the optional `expectedOwnedPid`, so a third call site cannot compile without deciding the question. The refusal path passes a literal `false`, not an expression that happens to be false today.
- **The supervisor regression test goes red pre-fix by assertion, not by timeout.** Traced the arithmetic by hand: with the `lived`-derived reset restored, `now` advances 35 s per attempt, so `lived ≥ QUICK_DEATH_MS` every time, `consecutiveFailures` returns to 1 on each of the three attempts, and `exhaustedEvent()` is still `false` at the third → `expected false to be true`. That is the observed output on record in #2379.
- **The deferred step-1 mutation check is real, and against *this* head.** #2383's `AGENT DONE` records re-running it against the reworked `const timer` shape (`Test timed out in 15000ms`), which matches `server/vitest.config.ts`'s `testTimeout: 15_000`; #2379's `30014ms` is the same 15 s doubled by `retry: 1`. Both receipts show output rather than asserting the check happened. The PR body's deferral holds.
- **`git diff origin/main -- eslint.config.mjs` is empty**, no `eslint-disable` appears anywhere in the diff, and no `.skip` / `todo` / quarantine entry was added.
- `Closes #2106` present as plain prose, outside any code span. #2106 is `bug` + `area:srv`, so no `docs/BACKLOG.md` row is owed (that file renders `type:feature` only).
- No `cast.json` writes, no new config knob, no OpenAPI or other derived artifact in scope.
- `POST /api/sidecar/restart` recovery intact: `sidecar-health.ts:705-706` still does the synchronous `exhaustedEvent()` check then `resetAndRespawn()`, and `resetAndRespawn` still zeroes `consecutiveFailures` (`sidecar-supervisor.ts:488`). The new test exercises it after exhaustion.
- **Not a finding, noted only:** `settle` (`spawn-sidecar.ts:330`) closes over `timer`, declared with `const` twelve lines later at `:344`. No production path reaches it — Node emits `'error'` from `spawn` asynchronously, so `:344` always runs first — but `findListenerPid`'s `spawnFn` and `spawnSidecar`'s `findPidFn` are public seams, and a fake child that emits `'error'` synchronously (a natural way to write a spawn-failure test) would throw `ReferenceError: Cannot access 'timer' before initialization` from inside an EventEmitter handler, i.e. uncaught rather than a rejected promise. Hoisting the `const timer` line above `settle` costs nothing if the block is being touched for finding 3 anyway.

**Re-review trigger: 1 correctness finding (#1).** Findings 2-7 alone would not re-trigger.

dudarenok-maker:
## PR review — pass 2 (head 48bbf61b, depth high)

Re-review of the ten commits on top of `2e78206a`. Depth `high` re-derived from the current commit set (`fix(server)` ×3, `docs(server)` ×1, `docs(docs)` ×5 — multi-scope). Reviewed in `C:\Claude\Projects\wt-2106-respawn-budget` at `48bbf61b`; both checkouts clean before and after. Deltas only — the pass-1 "Checked and clean" list is not re-litigated.

Ran here rather than taken on trust: `vitest run --retry=0 src/tts/sidecar-supervisor.test.ts src/tts/spawn-sidecar.test.ts` → 2 files / **84 tests** passed (was 77); `npm run typecheck` → green (root + server); `npm run check:onbox-register` → OK, register and live view agree.

---

**1 (🟠, `9d8cf8c8`) — VERIFIED RESOLVED.** The stamp now sits at `sidecar-supervisor.ts:322-324`, guarded on `handle !== null`. I traced every route that can reach the reader at `:484` looking for the stale-or-`0` door the fix could have opened, and there isn't one:

- `onExit` is wired only at `spawn-sidecar.ts:819-830`, i.e. *after* the `typeof pid !== 'number'` early return (`:782-786`), so `onChildExit` is unreachable on every null-returning path. Adopt (`:658`), not-ours (`:681`), stale-unidentifiable (`:709`), port-still-bound (`:716`), spawn-throw (`:769`) and no-pid (`:785`) all return before any exit handler exists.
- `handle = await spawnFn(...)` resolves into a **microtask**; a `'exit'` event is a macrotask, so the stamp always lands before any exit callback can run. There is no window in which `onChildExit` observes `lastSpawnAt === 0`.
- The stale-direction is closed too: between a stamp and that child's exit, `lastSpawnAt` can only be *overwritten by a newer successful spawn*, which moves it forward, making `freshIncident` **false** — never spuriously true.
- Adopt leaves `lastSpawnAt` stale-but-unread (no owned child ⇒ no `onExit`), and the refusal path passes a literal `false` (`:340`) so it never reads it at all. Both comments now match the code.

The paired test (`sidecar-supervisor.test.ts:1021-1091`) is a real regression test and goes red **for the reason claimed**, verified by hand-tracing the arithmetic against the reverted stamp: with the old top-of-`spawnOnce` stamp, `lived` at the 3rd child death is `105003 − 70002 = 35001 ≥ QUICK_DEATH_MS`, so the counter resets to 1 on every attempt and `exhaustedEvent()` is still `false` at the final assertion — an assertion failure, not a timeout. With the fix, `lived = 1 ms` each time and the counter accretes 1→2→3 > cap(2).

**2 (`13351467`) — VERIFIED RESOLVED.** `spawn-sidecar.ts:291-303`: the const and its doc now precede `findListenerPid`'s own doc block, and the null enumeration reads "(no tool, parse miss, error, or deadline expiry)".

**3 (`def0e103`) — VERIFIED RESOLVED.** `spawn-sidecar.ts:321-332` is a single simple command. Parse compatibility confirmed: `out.trim().split(/\s+/)[0]` (`:353`) already took the first token, and the new multi-line test pins it. `grep -rn 'head -n1|lsof -ti'` across the tree returns no other consumer of the old string.

**4 (`59911413`, `fe296214`) — VERIFIED RESOLVED, with the advice checked rather than assumed.**

- **Is "the supervisor will retry on backoff" true from this call site?** Yes. `spawnSidecar` has exactly one production caller — `sidecar-supervisor.ts:305` (`grep` for `spawnSidecar(` outside tests returns only the supervisor's own `spawnFn` default), so `onSpawnRefused` is always wired; it sets `refusedReason` (`:313-315`) → `scheduleRespawnAttempt(..., false, ...)` (`:335-344`) → backoff → `spawnOnce`. The claim stays honest once the budget exhausts because the supervisor emits its own terminal line at `:266-270` ("giving up respawn. TTS is DOWN") — the branch message never promises unbounded retry.
- **Does branching the `onSpawnRefused` reason break a pattern-matcher?** No. `refusedReason` reaches exactly two log templates (`:336-337`) and nothing else; it is not on any client-facing path (`models-status.ts:32-34` reports `'crashed'` with no reason string, `setup-diagnosis.ts:85` is a static sentence). The pre-existing parse-miss assertion at `spawn-sidecar.test.ts:197` still passes because its stub never fires the expiry callback.
- **Five positional args.** `findListenerPid(port, platform, spawnFn, deadlineMs, onDeadlineExpiry)` — `spawn-sidecar.ts:607` passes `(p, undefined, undefined, undefined, onExpiry)`, which is positionally correct, and explicit `undefined` triggers each default. The widened `SpawnSidecarOpts['findPidFn']` is a supertype of every 1-arity stub, so all 36 `spawnSidecar({…})` test call sites still typecheck (confirmed by `npm run typecheck`) and behave identically (an ignored callback leaves `deadlineExpired` false → parse-miss branch).
- *Noted, not a finding:* the branching creates a new false contrast. The parse-miss sentence still says "leaving it in place. Restart the sidecar manually" (`:705`) while that path retries on the same budget — a reader comparing the two branches will infer it does not. Unchanged text; flagging only so it isn't re-discovered.

**5 (`d4ace218`, `749001d2`) — RESOLVED as a house gate; two substantive defects remain in the row itself (N1, N2 below).** The bookkeeping is correct: A46 added to `onbox-acceptance-register.md:2473-2485` and `onbox-acceptance-register-live-view.html:843-859`, group A 45→46 and total 72→73 in both, `check:onbox-register` green, no per-feature run sheet invented (none exists). The quoted log formats are now **exact** against the code, which I checked character by character:
- `[sidecar] supervisor: spawn refused: <reason>; respawning in <delayMs>ms (attempt <K>/<max>).` == `sidecar-supervisor.ts:273-276` with `attemptLabel` from `:336`. ✔
- `[sidecar] supervisor: <N> rapid spawn refusals (<reason>) in a row — giving up respawn. TTS is DOWN; restart the server to recover.` == `:266-270` with `giveUpLabel` from `:337`. ✔
- `DEFAULT_BACKOFFS_MS = [2s, 5s, 15s]`, cap `5`, cited as `sidecar-supervisor.ts:45-46` — both line numbers are right, and "≈52s across 5 attempts" checks out (2+5+15+15+15). ✔
- The A34 correction is real: `GET /api/setup/models-status` is the route (`app.ts:328` + `routes/models-status.ts:102`), and `'crashed'` is genuinely reachable from `exhaustedEvent()` (`models-status.ts:33`, `:92`), surfacing as the "Voice engine crashed" alarm pill (`src/components/setup/engine-card-status.ts:22`). ✔

**6 (`76a31f35`) — VERIFIED RESOLVED.** `RELEASE_NOTES.md:156` now describes 1.14.0's actual failure ("used to see it wasn't running but would report it as ready anyway"), which matches the supervisor's own account of #2037 (`sidecar-supervisor.ts:301-303`, "the old unconditional `isRecycling = false` that silently declared a nonexistent sidecar ready"). The "and if it still won't come up, it tells you plainly" half is substantiated end-to-end, not just asserted: `exhaustedEvent()` → `'crashed'` → alarm pill.

**7 (`1ca9cc0b`) — VERIFIED RESOLVED, and the other half was against a false premise.** The doubled blank lines this PR added are gone: single blank before `# Castwright 1.14.0` and before `---`. The "no blank line *before* the new bullet" half does not survive contact with `main` — `git show main:RELEASE_NOTES.md` has bullets 148/149/150 and 152/153 already adjacent, and `docs/release-notes-next.md:344-347` likewise. The new bullets match their neighbours; leaving them is correct, not a residual.

**Incidental chore (`48bbf61b`) — accurate.** I checked every factual claim in the expanded `docs/release-notes-next.md:348` entry against the diff and the code; all hold except one overstatement, folded into N3.

---

### 🔴/🟠 Correctness bugs

**None.** The ten commits contain no new correctness defect, and no pass-1 finding is still open.

---

### 🟡 Cleanup nits

**N1. `docs/testing/onbox-acceptance-register.md:2482` (and `onbox-acceptance-register-live-view.html:852`) — A46's last bullet asks the operator to confirm a recovery the code cannot perform, and the same bullet block quotes the log line that says so.**

> "Stop the foreign listener, confirm the sidecar respawns and surfaces as ready within one backoff window (~15s for the final backoff)."

The bullet immediately above it (`:2480`) instructs the operator to **wait for exhaustion**. Once `consecutiveFailures` exceeds the cap, `scheduleRespawnAttempt` takes the give-up branch at `sidecar-supervisor.ts:265-271` and `return`s **without scheduling anything** — no timer, no watchdog (`onAdopt` was never armed, this path never adopted). Removing the foreign listener therefore changes nothing. The only routes back are `POST /api/sidecar/restart` → `resetAndRespawn()` (`routes/sidecar-health.ts:706`) or a server restart — which is exactly what the exhaustion log quoted two lines earlier says: *"TTS is DOWN; restart the server to recover."* The row contradicts itself.

Concrete failure: an operator runs A46 in order, waits ~15s at bullet 6, sees nothing, and either records the row as **failed against correct code**, or restarts the server, watches TTS come up, and records bullet 6 as **passed** — a discharge for behaviour that was never exercised. Fix: either state the recovery is manual (`POST /api/sidecar/restart`, or restart the server) or move the "stop the listener → auto-respawn" observation *before* exhaustion, at attempt 2 or 3.

**N2. `docs/testing/onbox-acceptance-register.md:2481` + `:2478` (and live view `:851`, `:848`) — the row's headline claim is the deadline, and the setup it prescribes cannot reach the deadline's call site or produce the log it asks the operator to read.**

The row's premise (`:2475`) is *"whether the deadline timer truly kills a hung probe"* and bullet 5 asks the operator to *"Record the listener-enumeration probe's actual wall-clock latency from the logs."* Neither is achievable on the scripted setup:

- A foreign listener on `:9000` fails `looksLikeSidecar`, so `spawnSidecar` takes the not-ours branch and **returns at `spawn-sidecar.ts:681`**, never reaching the deadline-aware call at `:694`. The `probe for the PID on :<port> timed out` sentence (`:696-701`) lives entirely behind `health.looksLikeSidecar === true` and is unreachable in this scenario by construction.
- The one probe that *does* run on this path — the D2 hint at `:674` — is called **without** `onDeadlineExpiry`, so an expiry there is still silently indistinguishable from a parse miss (the exact gap pass-1 finding 4 named; it was closed at the other call site only).
- No log line anywhere reports probe latency. `findListenerPid` logs nothing at all, so "record the probe's wall-clock latency from the logs" has no source.

Concrete failure: the operator observes nothing for bullets 5 and the row's stated premise, and records A46 as discharged on the strength of bullets 3–4 (which only exercise the counter, already fully unit-tested) — i.e. the row is discharged **without its own feature ever being observed**. Fix: either drop the deadline claim from A46 and keep it a budget-exhaustion row, or prescribe a setup that reaches `:694` — a fake listener that answers `/health` with `{"ok":true,"engines":[]}` (no `protocol_version`) puts `spawnSidecar` on the stale-replace path where the deadline branch actually lives.

Sub-point, same bullet block: `:2478`'s "or a simple Node HTTP server binding that port" is under-specified in a way that breaks bullet 4. `routes/sidecar-health.ts:410-417` maps any non-2xx to `unreachable`, but a catch-all server answering `200` on `/health` makes `reachable` true, and `models-status.ts:32` returns `'ready'` **before** the `'crashed'` check ever runs — so the operator's own choice of stand-in decides whether bullet 4's criterion can pass. `nc -l 9000` is fine; say "must not answer HTTP 2xx on `/health`".

**N3. `server/src/tts/spawn-sidecar.ts:607` and `spawn-sidecar.test.ts:1290-1312` — the one line that connects the two halves of the finding-4 fix is untested, and the test that claims to pin the default deadline does not.**

Two mutations survive the whole 84-test suite:

1. Change `:607` from `findPidFn = (p, onExpiry) => findListenerPid(p, undefined, undefined, undefined, onExpiry)` to `findPidFn = (p) => findListenerPid(p)`. Every one of the 36 `spawnSidecar({…})` call sites that reaches the stale-replace branch stubs `findPidFn` (verified by enumerating all of them), so nothing exercises the production default. `deadlineExpired` would then be permanently `false` in production, making the entire timeout branch (`:696-702`) dead code as shipped — the whole point of `fe296214` — with the suite green.
2. Change `deadlineMs: number = LISTENER_PID_DEADLINE_MS` (`:308`) to `deadlineMs: number = 1000`. The "Half A coverage" test advances by `LISTENER_PID_DEADLINE_MS + 1` (`:1309`), which covers **any** default ≤ 5001, and its only pre-advance assertion (`expect(child.kill).not.toHaveBeenCalled()`, `:1306`) runs at 0 fake-ms elapsed, so it holds for every value. The test proves "the constant is 5000" and "the default is at most 5000" — not the thing its own comment claims, "(2) calling `findListenerPid` without `deadlineMs` uses it".

This also makes `docs/release-notes-next.md:348`'s "previously every test passed an explicit `1000` override, so a typo would go unnoticed" slightly generous — a typo in the *constant* is now caught, a typo in the *default parameter* still is not.

Fix for (2) is one line: advance to `LISTENER_PID_DEADLINE_MS - 1`, assert `child.kill` still uncalled, then advance past. Fix for (1): one `spawnSidecar` test that omits `findPidFn` and injects a `spawnFn` returning a hanging child, asserting the timeout sentence.

---

**Re-review trigger: 0 correctness findings.** N1–N3 are cleanup-only and do not re-trigger. N1 and N2 should still land before merge — the A46 row is the merge-gating record of what has to be proven on the box, and as written it prescribes one observation the code cannot produce and one it contradicts.

Issue #1981, opened 2026-07-31
srv — #1953's designed-arm manifest read yields inside the cast.json read-modify-write window


dudarenok-maker:
﻿## What

`POST /api/voice-library/:voiceUuid/assign` must keep its cast.json read-modify-write window free of `await`s so the RMW stays atomic against a concurrent writer. The window is `readJson` (`voice-library.ts:1277`) -> `nextCharacters` -> `writeJsonAtomic` (`:1434`), and the invariant is documented at length at `:1243-1260`.

`await readJson(...)` at **`:1386`** sits inside that window.

Introduced by `1a6324d1` (2026-07-30, #1953 — "warn when a designed voice's baked manifest language differs from the book's"). It is on the **designed** arm, distinct from the cloned arm that #1933 touches.

## Why it matters

A concurrent write to the same cast.json — a debounced cast-editor save, `voice-override-linked`'s `applyToBook`, or a second `/assign` — can land in the yield and then be silently clobbered when this handler resumes and writes back the `characters` array it captured before yielding. This is the "cast.json clobbered" defect class the project has hit before, which is why the invariant is documented in the first place.

## Why it was not folded into #1933

Assessed against CLAUDE.md's fix-now bar; fails two of four:

- **Two defensible fixes**, not one: hoist the read above `:1277` and build the warning string later, OR accept the window for this arm and document why. That is an interface/ordering decision, not a typo.
- **No straightforward same-PR test**: proving it needs a racing-writer harness, which is its own piece of work.

Filed per the incidental-findings protocol rather than fixed opportunistically.

## Acceptance

- The designed-arm manifest read no longer yields inside the RMW window, or a written rationale for why it is safe there.
- A regression test that fails before and passes after — a concurrent-writer harness proving the second write is not clobbered.

## Key files

- `server/src/routes/voice-library.ts:1386` — the offending `await`
- `server/src/routes/voice-library.ts:1243-1260` — the invariant it violates
- `server/src/routes/voice-library.ts:1181-1188`, `:1261` — `hasRetainedDesignedClip`, the worked example of hoisting I/O outside the window

## Benefit

*Benefit (technical):* closes a silent-data-loss window on the file that carries every character's voice assignment. Found by the independent review of the #1933 design brief.

Refs #1953 · found while designing #1933

dudarenok-maker:
## Implementation brief — cast.json write-lock sweep

[...]

Issue #1899, opened 2026-07-27
fs-38 — wholesale cast PUT lets a client restamp voiceUuid + engine slot with no consent check


dudarenok-maker:
## What

`PUT /api/books/:bookId` with `{slice: 'cast'}` persists the client's cast
array **wholesale**, with no clone-key or provenance guard, and `voiceUuid`
is **not** in `PRESERVED_DESIGN_FIELDS`. A client can therefore:

1. `PUT` a cast where a character's `voiceUuid` is stamped to a real cloned
   voice's uuid (or any arbitrary value — nothing validates it against a
   consent-checked source).
2. Send a second request whose `overrideTtsVoices.coqui.name` matches that
   uuid's canonical shape.

`matchesCanonical`-style checks that gate other write paths (e.g. the
override-link route, closed by #1884) never see this route, because this
route writes both fields directly in one shot with no guard at all.

## Why it's wider than the already-fixed #1884

#1884 (the `voice-override-linked.ts` consent hole) is now fixed
(`493d6b8f` + fix round) — but its own review found that the "never
client-supplied" framing underlying that fix doesn't hold here: this
wholesale cast-write route was **never gated by that fix or any other**, and
doesn't need the override-link route as an intermediate step at all. It's a
more direct path to the same outcome (a client controlling both `voiceUuid`
and the engine-slot name it resolves through, with no consent check on
either).

## Suggested fix shape

Flagged in the originating review with a caution worth preserving: **a naive
fix may break legitimate saves** if the frontend round-trips a full cast
object through this route in normal operation (i.e. `voiceUuid` legitimately
needs to survive an ordinary save). That distinction — "does the frontend
ever intentionally send a changed `voiceUuid` through this route" — decides
whether the right fix is a one-liner (add `voiceUuid` to
`PRESERVED_DESIGN_FIELDS` so the server always keeps its own value) or a
larger design change (validate any client-supplied `voiceUuid`/engine-slot
pair against the same consent check other routes use). Whoever picks this up
should establish that first.

## Key files

- The `PUT /api/books/:bookId` route handling `{slice: 'cast'}` (server-side
  cast-write handler — locate by symbol/route, not by a line number, per
  this wave's own repeated lesson about brief citations drifting)
- `PRESERVED_DESIGN_FIELDS` (wherever it's defined — the allowlist this
  route consults for which fields survive a wholesale write)
- `server/src/routes/voice-override-linked.ts` (the related, already-fixed
[...]

Issue #1885, opened 2026-07-27
fs-38 — manual cast-link routes bypass the library consent check [ADV-M4]


dudarenok-maker:
## What

Two manual cast-linking routes write voice fields directly without going
through the consent check `library-cast-scan.ts:81-83` already enforces
elsewhere:

- `server/src/routes/cast-link-prior.ts:204-207`
- `server/src/routes/series-reuse-link.ts:390`

Both bypass `library-cast-scan.ts:81-83`, so a cloned voice's slot can be
copied onto a character in a book that never had consent captured for that
voice/book combination.

## Why it's separate from the already-fixed force-to-Qwen bug

`cast-link-prior.ts` had a *related but distinct* defect fixed in Wave 3c's
Phase 0 (folded in as "Task 6a" during that wave): a foreign designed slot
could be planted onto a cloned character via the same route's merge logic,
compounded by `generation.ts` force-overwriting `ttsEngine` for out-of-support
languages. That was fixed with `hasClonedProvenance` gating the whole
`shouldDenormaliseVoice` merge block.

This issue is a **different** interaction on the same two files: the
consent-bypass shape, not the engine-force shape. Wave 3c's own review noted
explicitly that the wording used to scope the Task 6a fix "would NOT have
caught" this issue — it needs its own fix.

## Suggested fix shape

Route both write paths through the same consent check
`library-cast-scan.ts:81-83` performs before writing a cloned voice's slot
onto a character, or add an equivalent explicit consent-scope check local to
each route before the merge/copy proceeds.

## Key files

- `server/src/routes/cast-link-prior.ts:204-207`
- `server/src/routes/series-reuse-link.ts:390`
- `server/src/workspace/library-cast-scan.ts:81-83` (the consent check being
  bypassed)
- `server/src/tts/clone-engines.ts` (`hasClonedProvenance` — the fail-safe
  helper the related, already-fixed bug used; may be reusable here too)

## Related

- fs-38 / #624 (voice cloning umbrella)
- Plan: `docs/features/270-fs38-wave3c-xtts.md`, "Known limitations (h)"
- Adjacent to but distinct from the Wave 3c Phase-0 force-to-Qwen fix on the
  same files (`clone-engines.ts`'s `hasClonedProvenance` gate)

PR #2405, opened 2026-08-17
test(server,docs): prove voice-library entry writes are serialized (#1826)

### LINKED ISSUE #1826: srv-XX — serialize voice-library entry writes (per-uuid lock or updatedAt CAS)
## What

`writeEntry` (`server/src/workspace/voice-library.ts`) is a tmp+rename with **no compare-and-swap and no lock**, so every read-modify-write against a voice-library entry is a potential lost update. This is true of every manifest write in the codebase; this issue asks for a proper primitive (a per-uuid write lock, or an `updatedAt`-based CAS that rejects a stale write).

## Why now

fs-38 Wave 3b2's cloned-voice resolver re-derives a voice during chapter render, then writes its status back. An independent review (Fable, 2026-07-26) found that a **revoke landing mid-derive** was silently clobbered by the resolver's stale pre-derive snapshot — un-revoking consent and leaving a live `.pt`. That Critical was fixed in `655beb51` by re-reading the entry immediately before every post-derive write (`guardPostDeriveWrite`, `server/src/tts/clone-voice-resolver.ts`), which closes the seconds-wide GPU window.

Two residuals remain, both rooted in the missing primitive rather than in that fix:

**(a) Millisecond flag-only lost update.** A revoke landing between the guard's re-read and the following `writeEntry` can still clobber `revokedAt`. Benign in effect: the revoke route's purge runs strictly *after* its own `writeEntry`, so it lands last and the artifacts are still erased — the residue is a stale un-revoked flag with nothing renderable behind it (the next render fails loud on the missing master), recoverable by revoking again.

**(b) Two-worker compound corner — the one case where an artifact can outlive a revoke.** With two chapter workers repairing the *same* library voice concurrently (simultaneous multi-book renders sharing one voice): if a revoke hits worker A's millisecond window (clobbering the flag) **and** worker B's derive completes after the revoke's purge, B's guard re-read sees the now-un-revoked entry, skips its re-purge, and **B's `.pt` survives live**. Requires (a) to fire first plus a dual-render precondition, so it is orders of magnitude less likely than the original bug — but it is a real consent-erasure hole.

## Acceptance

- Entry writes for a given `voiceUuid` are serialized, or a stale write is rejected (e.g. `writeEntry` compares `updatedAt` and fails on mismatch, with callers re-reading and retrying).
- A regression test interleaving a revoke with an in-flight repair proves `revokedAt` survives.
- A test covering corner (b): two concurrent repairs + a revoke leave **no** `.pt` behind.

## Key files

`server/src/workspace/voice-library.ts` (`writeEntry`), `server/src/tts/clone-voice-resolver.ts` (`guardPostDeriveWrite`), `server/src/routes/voice-library.ts` (revoke route).

Refs #624 · context: `docs/features/268-fs38-wave3b2-resolver.md` Known limitations (j)

dudarenok-maker:
## Summary

The per-uuid write lock for voice-library entries shipped earlier (Wave 3c Task 14, `74bff8b3`). This PR delivers the missing proof, a static guard, and doc corrections that #1826 still required: two regression tests interleaving revoke against in-flight repair, a static guard against bare `writeEntry` bypasses of the lock, and corrections to four prose claims across three docs files that had gone stale after Task 14 shipped. Cross-process serialization (two separate `node` processes) remains out of scope and always was — see the carve-out comment in `server/src/workspace/voice-library.ts:222-226`.

Closes #1826

## Test plan

- [x] `npm run verify` (pre-push hook: lint, typecheck, build) — green on push
- [x] `cd server; npm.cmd run test` — 7750/7750 executed tests passed (529 files passed, 1 pre-existing unrelated file skipped, 9 individual tests skipped). One infra-level vitest worker-fork crash occurred (unrelated Python venv bootstrap churn in a concurrent sidecar test on this box) but caused no test-assertion failures.
- [x] Mutation M3 (bypassed `withEntryLock` in `updateEntry`) — `npx vitest run src/tts/clone-voice-resolver.revoke-race.test.ts` — Test 1 reddened: entry ended up with `reason:'derive-failed'` instead of surviving `reason:'revoked'`. Reverted, confirmed green again.
- [x] Mutation M2 (deleted the re-purge call in `statusStampMutate`) — same suite — both tests reddened: Test 1's purge-args assertion mismatched, Test 2's `pts.size` went from expected `0` to `1`. Reverted, confirmed green again.
- [x] Mutation M1 (bypassed `updateEntry`, wrote directly off the stale pre-derive entry) — same suite — both tests reddened (`UnresolvableClonedVoiceError` assertion failed on Test 1, `pts.size` mismatch on Test 2). Reverted, confirmed green again.
- [x] Guard mutation table (`voice-library-write-seam.guard.test.ts`), 4 mutations (unallowlisted `writeEntry` call, extra dead `writeEntry` call in `/design`, `updateEntry` aliased to `ue` at all 4 call sites, one of the 3 legitimate creation-site `writeEntry` calls deleted) — each reddened the specific guard assertion it was designed to catch (allowlist mismatch, count mismatch, per-file map miss, stale-allowlist miss respectively). All reverted, confirmed clean tree after each (`git diff --stat` empty).
- [x] Negative control — added a comment/string literal containing the literal text `writeEntry(x)` to a non-allowlisted file — guard stayed green, confirming comments/string literals are correctly treated as opaque by the scanner and don't false-positive the guard.
- [x] Confirmed acceptance bullet 1 by grepping every `writeEntry(` call site in `server/src`: only the definition/internal call inside `updateEntry` itself, and the three legitimate NEW-entry creation sites (`/design`, `/promote`, `/clone` in `routes/voice-library.ts`, each minting a fresh `voiceUuid`/`libraryUuid` immediately prior) remain bare. No fourth bare production call exists.
- [x] Confirmed the two new regression tests route through the real `updateEntry`/`writeEntry`/`readEntry` from `server/src/workspace/voice-library.ts`, not the lockless fake default in `clone-voice-resolver.test.ts`'s `makeDeps` (`:434-456`).
- [x] Confirmed both new tests gate ordering with explicitly-resolved awaited `Promise`s tied to structural events, not `setTimeout`/timing polling — matches the idiom at `server/src/workspace/voice-library.test.ts:151-202`.
- [x] `git show --stat` on all three commits (`3a0f55d9`, `a8db44ae`, `e9a7e0c4`) confirms test-only and docs-only changes; no `server/src/**` production source file is touched by any of them.

Release notes are deliberately skipped: tests + a build-time guard + doc corrections have no user- or operator-visible delta (CLAUDE.md Before-shipping step 5 permits this when stated rather than silently omitted).

No on-box acceptance row is owed: `docs/testing/onbox-acceptance-register.md`'s "Deliberately not in this register" section already records that #1826's bar is an automated interleaving regression test, not a manual walkthrough.

---

## Review-gate rounds (added after the original body was written)

Two `pr-review-gate` passes have run on this PR; both comments are in the thread. The original body above described the first three commits only. Five commits have landed since, all of them fixes to findings from those passes.

**`server/src/workspace/cast-lock.guard.test.ts` — a different seam this PR did not originally touch.** Its `skipOpaqueToken` was a byte-identical copy of the one the new write-seam guard introduced, so every tokenizer defect found here was armed there too. That guard is what fails the build on a new unlocked `cast.json` write, and it had been shipping with a blind span of up to 151 lines. Fixed in the same rounds rather than filed.

### What the passes found, and what changed

- **Pass 1 🔴 — the guard failed open.** `skipOpaqueToken` had no regex-literal case, so a `/` opening a regex desynced the walker and a quote inside the regex body was read as a string opener, marking live code opaque. A bare `writeEntry(` injected into `tts/voice-sample-cache.ts` was invisible to G1. Fixed in `56b9f8fe`, which replaced the hand-rolled lexer with the TypeScript parser.
- **Pass 1 🟡 — `G2_GLOBAL_FLOOR` was vacuous** (exactly the sum of the per-file map, so never the sole failing assertion). Deleted in `a256bc54`; the guard's documented M3 mutation is the standing evidence that nothing was lost.
- **Pass 1 🟡 — the KL-j prose sweep was incomplete.** Two cross-references still said "appears fixed in source" after the §6 rows were upgraded. Fixed in `89154448`.
- **Pass 2 🔴 — the fix above opened the inverse defect.** Comments were collected per node via `ts.forEachChild`, which never visits punctuation tokens, so a comment whose only following token is `}`, `]` or `,` was never opaque — 276 such comments across 79 files. This failed open (prose quoting an unbalanced `withCastLock(` at the end of a block fabricated a lock around a genuinely unlocked write) **and** closed (prose quoting a write pattern counted as a real call site). Fixed in `a06415e4` by descending to leaf tokens.

`48952e06` merges `origin/main`; every pinned allowlist count and line number was re-derived against the post-merge tree.

### Verification

The tokenizer is checked by three oracles, each control-validated against the pre-fix code so it is known to be able to fail: AST-literal starts for over-reach (166 desynced ranges → 0), leaf-token descent for under-reach (276 uncovered comments → 0), and a `ts.createScanner` trivia sweep for under-reach (152 → 0). Both guard-level polarities are re-confirmed each round: real code reddens G1, the same text as a comment stays green.

One caveat, corrected in `aece0cef` after pass 3 caught it: the leaf-token oracle is **no longer independent**. Once the fix made the tokenizer itself a leaf-token descent, that oracle shares its traversal and can only confirm itself — the same construction that produced an earlier "0 uncovered comments" result which was withdrawn as an artifact. The scanner oracle shares no traversal with either and is what actually carries the under-reach conclusion. Pass 3 added a third, independent check of its own: a 35-fixture battery of comment shapes absent from `server/src` (JSDoc on overloads, type-only positions, empty and comment-only files, EOF comments, shebangs, decorators, `enum`/`switch`/`namespace` closers), red on 21 of 35 against the pre-fix tokenizer and 0 of 35 after — plus a direct test of the fix's premise, diffing every-node against leaves-only comment collection across all 427 files and finding zero ranges an interior node contributes that no leaf does.

One earlier "0 uncovered comments" result was withdrawn as an artifact — that oracle used the same `forEachChild` traversal as the function it measured, so its zero was true by construction.

### Still not applicable

Release notes remain skipped: every file the two `fix(server)` commits touch is a `*.test.ts`, so there is still no user- or operator-visible delta. On-box acceptance is unchanged — the register already records #1826's bar as an automated interleaving regression test.


dudarenok-maker:
## PR review — pass 1 (head e9a7e0c4, depth high)

**Scope reviewed:** all 5 files across the three commits (`3a0f55d9`, `a8db44ae`, `e9a7e0c4`) — the two new test files, the three doc edits — plus the production code they claim to exercise (`server/src/workspace/voice-library.ts` `withEntryLock`/`updateEntry`, `server/src/tts/clone-voice-resolver.ts` `statusStampMutate`), the #1826 issue body and its phase-1 audit comment, the on-box register's "Deliberately not in this register" section, and a whole-tree sweep for surviving prose asserting entry writes are unserialized. Because this PR is entirely test + guard + prose, the review was aimed at whether the proof proves anything rather than at production behaviour.

---

### 🔴 Blocking

**1. The static guard fails open on 19+ production files — a real bare-`writeEntry` read-modify-write is invisible to it, and the header claims no such blind spot exists.** *(correctness bug)*

`server/src/workspace/voice-library-write-seam.guard.test.ts:126-157` — `skipOpaqueToken` recognises `'`, `"`, `` ` ``, `//` and `/* */`, but **has no case for a regex literal**. When the scan reaches a `/`, it checks only for `//` and `/*`; anything else falls through to `return -1` and the walker advances one char. The next char is the regex body — and if that body contains a quote character, `computeOpaqueRanges` treats it as a string opener and swallows everything up to the next matching quote, which is usually hundreds of lines of real code.

Concrete break, verified end-to-end (in memory only — nothing written to the tree):

`server/src/tts/voice-sample-cache.ts:104` is

```ts
return stripEdges(s, /[“”"'‘’\s]/).trim();
```

That `"` desyncs the scanner, and lines 104–190 of the file become one opaque range. Insert a genuine lost-update bypass at line 106 — exactly the thing G1 exists to catch:

```ts
export async function touchLibraryEntry(uuid: string): Promise<void> {
  const e = await readEntry(uuid);
  if (!e) return;
  await writeEntry({ ...e, pinned: true });   // BARE write: no updateEntry, no lock
}
```

I confirmed with the TypeScript parser (`ts.createSourceFile` → the AST contains both the `touchLibraryEntry` `FunctionDeclaration` and the `writeEntry(...)` `CallExpression`, so it is genuine executable code, not a comment or a string), then replayed G1's own loop over the mutated tree:

```
G1 problems, BASELINE (unmutated tree): []
G1 problems, WITH the bare-writeEntry bypass: []
VERDICT: G1 PASSES GREEN — the bypass is invisible
```

`countNonOpaque` returns 0 for that file, so `voice-library-write-seam.guard.test.ts:254` (`if (count === 0) continue;`) skips it and no problem is ever pushed.

Sweeping all 427 non-test files under `server/src` with the parser as the oracle — injecting only at positions immediately after a line consisting solely of `}`, so this is a **lower bound** — a parser-confirmed bare `writeEntry(` is invisible to the guard in **19 files**:

```
analyzer/attribution-eval/scorer.ts    analyzer/byline-author-guard.ts
analyzer/recover-tagged-lines.ts       analyzer/roster-coverage.ts
analyzer/sentence-progress.ts          analyzer/stage2-chunk.ts
analyzer/stage2-coverage.ts            analyzer/tag-grammar.ts
export/build-m4b.ts                    export/build-mp3-zip.ts
parsers/audio-tags.ts                  parsers/epub.ts
parsers/front-matter.ts                parsers/text.ts
tts/normalize/index.ts                 tts/text-normalize.ts
tts/voice-sample-cache.ts              util/text-match.ts
workspace/paths.ts
```

`tts/voice-sample-cache.ts` is squarely in the voice-library feature family (it is the sample cache `POST /:uuid/sample` serves from), so this is not a hypothetical corner of the tree.

Two things make this blocking rather than a nit:

- **The header denies it.** `:54-79` is titled "G3 — THE HEADER DECLARES THE BLIND SPOTS, VERBATIM AND IN FULL" and enumerates five. Regex-literal desync is not among them, and blind spot #5 specifically claims the plain-substring match leaves "the 'not covered' side ... EMPTY here" apart from aliasing and declaration drift. A guard whose limits are documented as exhaustive but are not is the exact failure its own `:5-8` warns about — "decorative coverage is worse than none because it reads as protection to the next person who touches this code."
- **The mutation table cannot detect it.** All four mutations landed in desync-free files: M1 in `workspace/purge-clone-artifacts.ts` (I re-ran it — correctly CAUGHT), M2/M4 in `routes/voice-library.ts`, M3 in an allowlisted file. None of the three allowlisted files contains a quote-bearing regex literal, so the table is green by luck of placement, not by coverage.

Secondary consequence worth fixing at the same time: the desync also produces **false reds**. A future refactor that adds a quote-bearing regex to `routes/voice-library.ts` above its three creation sites would collapse its scanned count from 3 to fewer and trip G1's "stale allowlist entry (fewer found than pinned)" branch, sending the next reader after a nonexistent deleted call site.

Cheapest sound fix: add a regex-literal branch to `skipOpaqueToken` (a `/` starts a regex when the previous significant character is one of `(,=:[!&|?{};+-*%~^<>` or a keyword like `return`/`typeof`/`case`), or drop the hand-rolled scanner for the TypeScript scanner already in `server/node_modules`. Either way the mutation table needs a fifth row that places the bypass **inside a desynced region** — otherwise the next round re-ships the same hole.

---

### 🟡 Minor

**2. G2's global floor is vacuous — it can never be the assertion that fails.** *(cleanup nit)*

`voice-library-write-seam.guard.test.ts:243` pins `G2_GLOBAL_FLOOR = 13`, which is exactly the sum of the five per-file expectations (`4+6+1+1+1`), and `:326-330` fires only when `globalTotal < 13`. But `:302-314` already fails on any per-file count that is not exactly its expected value, and `:317-324` already fails on any expected file that scans at zero. So `globalTotal` can only fall below 13 in a state where one of those two branches has already pushed a problem — the floor is strictly implied and adds no detection. It is not harmful, but the test title ("and the global floor holds") and the header's "plus a global floor" both read as a second line of defence that is not there. Either drop it, or make it independent (e.g. a floor over files *not* in the per-file map).

**3. The prose sweep is incomplete — two surviving stale claims of the same class, in a file this PR already edits.** *(cleanup nit)*

The PR corrects KL-j(1)/(2) in `docs/testing/fs38-wave3-onbox-acceptance.md` from "**Appears fixed in source ... issue still open**" to "**Fixed in source ... and proven by the #1826 chain's Step 1 regression tests**". Two other places in the same file still describe those rows by the old marking:

- `docs/testing/fs38-wave3-onbox-acceptance.md:1358-1359` — "(see §6 KL-j(1)/(2), now marked **appears-fixed-in-source**)". After this PR they are not marked that; the cross-reference points at a label that no longer exists.
- `docs/testing/fs38-wave3-onbox-acceptance.md:2267` — D-01's repeat step: "§6 KL-j(2) ... now **appears fixed in source**".

Operator behaviour is unaffected (both passages already say a surviving `.pt` is a regression, and that instruction is still correct), so this is phrasing rot rather than a wrong instruction — but it is the same claim class the PR set out to correct, left armed in the file it touched.

---

### ✅ What is solid

Verified rather than taken from the PR body:

- **The regression tests route through the real lock, not a double.** `clone-voice-resolver.revoke-race.test.ts:114-133` wires `readEntry`/`writeEntry`/`updateEntry` straight to the freshly-imported `workspace/voice-library.js`; the file neither imports nor uses `clone-voice-resolver.test.ts`'s lockless `makeDeps`. The only wrapper on `updateEntry` is a synchronous `onQueued` callback, which observes the queue-slot claim without changing it. `beforeEach`'s `vi.resetModules()` + re-import gives each test a fresh `entryLocks` map and a fresh `WORKSPACE_ROOT`, so the tests do not leak lock state into each other.
- **Ordering is gated structurally, not on timing.** `:146-155` signals from *inside* the revoke's mutate (so "the lock is held" is observed, not assumed) and `:182`'s `Promise.race([repairQueuedP, settledP])` races two structural events rather than sleeping. No `setTimeout`, no polling. Ran the suite 5× consecutively: 2/2 passing every run, ~55 ms of test time — no flake signal.
- **Test 2 is honest about its reach.** `:210-213` says in terms that it pins the re-purge, not the lock, and points at Test 1 as the lock's instrument. That is the right disclosure and it is accurate.
- **Both suites are reachable as shipped.** Neither file is in `server/vitest.config.ts`'s `SLOW_FILES_TO_EXCLUDE`, so both run in `npm run test:server`; confirmed by running them (2 files, 4 tests, all passing).
- **The G1 allowlist's factual claims check out.** `routes/voice-library.ts` scans at exactly 3 bare `writeEntry(` (`:250` design, `:1035` promote, `:1326` clone) and 4 `updateEntry(` (`:423`, `:611`, `:747`, `:2038`). All three are genuine first-write creation sites — `:250` and `:1326` mint `voiceUuid` fresh, and `:1006` is `const libraryUuid = nanoid()` for promote — so "no prior entry, no read-modify-write, no lock owed" holds. The two `clone-voice-resolver.ts` and one `synthesise-chapter.ts` entries really are interface declarations. Neither of those three files contains a quote-bearing regex, so finding 1 does not compromise the allowlisted counts themselves.
- **G2 catches the alias it claims to.** Simulated `import { updateEntry as ue }` + renamed uses in `routes/voice-library.ts`: the file's count collapses to 0, `seen` never gains the key, and `:317-324`'s stale-map branch fires. The asymmetry the header declares (the same stunt on `writeEntry` is *not* caught) is real and correctly disclosed as blind spot #1.
- **`collectSourceFiles` fails closed on relocation.** A missing `server/src` throws out of `readdirSync`; a wrong-but-present root leaves all three allowlist keys unmatched and reds via the stale-entry branch.
- **House gates.** `Closes #1826` present in the Summary as plain prose, outside any code span. The release-notes skip is stated with a reason that holds (tests + a build-time guard + doc corrections carry no user- or operator-visible delta — CLAUDE.md step 5). The on-box claim is true: `docs/testing/onbox-acceptance-register.md:3553-3556` does list #1826 under "Deliberately not in this register — its bar is an automated interleaving regression test, not a manual walkthrough", so no row, run sheet or live-view move is owed. #1826 is `type:chore`, so no `docs/BACKLOG.md` row is owed either. Three commits, all genuinely test-only/docs-only; no `server/src/**` production source is touched.
- **The scope carve-out is legitimate, and `Closes #1826` is defensible.** The issue body's three acceptance bullets are: serialization (shipped in `74bff8b3`), a revoke-vs-repair interleaving test (Test 1), and a corner-(b) test (Test 2). Cross-process is nowhere in the issue body, and the owner's own phase-1 audit comment on #1826 states it "was never in scope for this issue and still is not" — so the carve-out at `voice-library.ts:222-226` is ratified on the ticket, not asserted only by this PR.
- **The corrected prose is true against the code on this branch.** 268's "both residuals now closed in-process", 271's limitation (d) as rewritten, and 271's "closed in-process by Task 14 ... only the intentionally out-of-scope cross-process carve-out remaining" all match `withEntryLock`/`updateEntry` at `voice-library.ts:186-314` and `statusStampMutate` at `clone-voice-resolver.ts:453-468`. The one remaining tree-wide hit for "unserialized (#1826)" outside the edited files — `docs/superpowers/plans/2026-07-26-fs38-wave3c-xtts.md:584`, "Cross-process voice-library writes unserialized (#1826)" — is still accurate and correctly left alone.
- The new tests and guard encode the same lock model the code documents; nothing here contradicts CLAUDE.md's `withCastLock` rules or the `design → library-voice → cast` order.

---

### Verdict

**Request changes.** One correctness bug (finding 1): the static guard — one of this PR's three deliverables, and the one whose whole job is to fail the build on a bare-`writeEntry` bypass — is blind to at least 19 production files, and its header states its blind spots are declared in full. The two regression tests are genuinely sound: they route through the real per-uuid lock, gate on structural events rather than timing, and are stable across repeat runs; the doc corrections are accurate. Fix the scanner (plus a mutation-table row that places the bypass inside a desynced region), and findings 2 and 3 while there.

dudarenok-maker:
## PR review — pass 2 (head 89154448, depth high)

Deltas only, per the re-review rule. Adjudicated the three pass-1 findings independently (not from the commit messages), then aimed the rest of the pass at the four fix commits themselves — `56b9f8fe`, `a256bc54`, the 35-commit merge `48952e06`, and `89154448`.

**Instrument note.** I did not reuse the scratch censuses. I built my own: it loads the tokenizer out of the guard file by transpiling the sentinel block, then measures it against an oracle built from a *different* traversal — every leaf token from `sf.getChildren()`, with each token's own leading trivia as the comment set. Control-validated before use: run against the pre-fix walker at `e9a7e0c4` it reports **15,763 over-reach tokens in 242 files**; run against HEAD it reports **0**. It can fail.

---

### Pass-1 finding 1 (🔴 static guard fails open on a regex-literal desync) — **VERIFIED RESOLVED**

- Whole-tree census over all **427** non-test files under `server/src`: **over-reach 0 tokens in 0 files**. No live token starts inside an opaque range anywhere.
- The exact pass-1 repro, replayed in memory: injecting the bare `writeEntry({ ...e, pinned: true })` bypass after `tts/voice-sample-cache.ts:104` (`return stripEdges(s, /[“”"'‘’\s]/).trim();`) now yields `["tts/voice-sample-cache.ts: 1 NOT allowlisted @ 111"]` against a baseline of `[]`.
- Both tokenizer regression tests that claim a red phase against committed code really have one, checked by running them against `e9a7e0c4`'s walker: T1 red (`call is live: true !== false`), T2 red (`inner text opaque: false !== true`), and cast-lock's nested-template test red (`writes: 2 !== 1`). T3 and the prose-lock test are green against `e9a7e0c4` — and both files say so in terms, naming the uncommitted attempt they actually guard. That disclosure is accurate and is the right way to ship a test with no committed red phase.

### Pass-1 finding 2 (🟡 vacuous `G2_GLOBAL_FLOOR`) — **VERIFIED RESOLVED**

Constant, `globalTotal` accumulator, floor branch, the title clause and every prose reference are gone. G2 is now exact-per-file plus the two-directional presence check, and the header no longer advertises a second line of defence that was not there.

### Pass-1 finding 3 (🟡 incomplete prose sweep) — **VERIFIED RESOLVED**

`:1358` and `:2267` both now read "fixed in source and proven by the Step 1 regression tests", matching §6's KL-j(1)/(2) rows. Grepping both spellings across the file leaves exactly two hits, `:2695`/`:2697` — which is finding 5 below, not this one.

---

### 🔴 Blocking — the rewrite closed over-reach and opened **under-reach**: 136 real comments across 64 files are not opaque, so `collectLockRanges`' new guard does not actually close the fabricated-lock hole it was added for *(correctness bug)*

`server/src/workspace/cast-lock.guard.test.ts:328-329`, and the byte-identical copy at `server/src/workspace/voice-library-write-seam.guard.test.ts:211-212`.

`computeOpaqueRanges` collects comments only from nodes reached by `ts.forEachChild`. `forEachChild` never visits punctuation tokens (`}`, `]`, `,`), and `getTrailingCommentRanges` stops at the first newline — so **a comment whose only following token is punctuation is never added to the opaque set**. Two shapes, both ordinary:

1. sole content of a block — `} catch { /* best-effort */ }`
2. the last line of a block, before the closing `}`

Both are live in this tree right now. `analyzer/model-vram-stats.ts:191` (`catch { /* best-effort */ }`) is shape 1; `analyzer/attribution-eval/apply-ops-chars.ts` (`// Any other op class (…) doesn't touch the char array.` immediately before `}`) is shape 2. Census total: **136 uncovered comments in 64 files**, of which **58 are in 11 of the 40 files the cast-lock guard actually scans** — `routes/analysis.ts` (15), `routes/generation.ts` (8), `routes/single-design.ts` (7), `routes/cast-design.ts` (6), `routes/book-state.ts` (5), `routes/chapter-qa-repair.ts` (5), `workspace/scan.ts` (4), `routes/script-review.ts` (3), `routes/qwen-voice.ts` (2), `routes/voice-library.ts` (2), `routes/cast-merge.ts` (1).

**Consequence A — fails OPEN, a fabricated pass (the worst shape there is here).** `cast-lock.guard.test.ts:436`'s `if (isOpaque(opaque, tokenStart)) continue;` only fires for comments the tokenizer covers. Put the same prose one line lower and the lock range is fabricated again. Fixture, run against the shipped `scanFile`:

```ts
export async function saveCast(dir: string, cast: unknown): Promise<void> {
  await Promise.all([
    (async () => {
      try {
        noop();
      } catch {
        // historical: this used to be withCastLock(dir, async () => {
      }
      await writeJsonAtomic(castJsonPath(dir), cast);   // genuinely UNLOCKED
    })(),
  ]);
}
```

```
opaque: []
locks : [{"start":200,"end":299}]      <- fabricated, from the comment's own text
scan  : null                            <- zero unlocked writes; guard GREEN
```

Control — the identical comment moved one line up, where the tokenizer does cover it — gives `locks: []` and `{"writes":1,…,"unlocked write @ line 10"}`. Shape 2 fabricates a lock the same way. This is exactly the hole `56b9f8fe`'s commit message says it closes ("prose quoting an unbalanced `withCastLock(dir, async () => {` let the FOLLOWING REAL CODE's parens close the range"), and the regression test at `:613` passes only because its fixture puts the comment on line 1, a position the node-trivia sweep happens to reach.

**Consequence B — fails CLOSED, and it destroys the advertised negative control.** A comment quoting the write pattern in either uncovered position is scanned as live code:

```ts
export async function f(dir: string): Promise<void> {
  if (dir) {
    await withCastLock(dir, async () => {
      await writeJsonAtomic(castJsonPath(dir), cast);
    });
    // NB: never call writeJsonAtomic(castJsonPath(dir), cast) outside the lock.
  }
}
```
→ `{"writes":1,"rms":0,"details":["unlocked write @ line 6"]}` on correct, unchanged code. Same comment before a statement → `null`. The write-seam guard behaves identically: its own T3 fixture, with the two comments moved one line down to the end of their blocks, goes from `countNonOpaque = 0` to `countNonOpaque = 2` — i.e. **T3 (`voice-library-write-seam.guard.test.ts:462`) pins its fixture, not the class it is named for**, and the mutation table's `NC` row is placement-dependent in a way it does not state.

**Three header claims are false as shipped**, which is what makes this blocking rather than a nit — same structural failure as pass-1 finding 1, where the header denied the hole:

- tokenizer block header (`cast-lock.guard.test.ts:288-291` / `voice-library-write-seam.guard.test.ts:171-174`): "**Every node** is asked for its leading and trailing comment ranges". Every node, yes — but not every comment; the ones this misses are the ones that matter.
- `cast-lock.guard.test.ts:189-204`: "a comment or string literal that happens to spell out `writeJsonAtomic(castJsonPath(` … **Both are closed now** … re-verified at guard level under #2405's rewrite." Consequence B above is that exact false positive, still live.
- `56b9f8fe`'s measurement — "under-reach (real comment/literal not covered) 0 comments -> 0" — is an artifact. The harness that produced it (`coverage-census.mjs`'s `realComments`) enumerates comments with `forEachChild` + `getLeadingCommentRanges(node.pos)` + `getTrailingCommentRanges(node.end)` + `getLeadingCommentRanges(text, 0)` — **byte-for-byte the traversal inside `computeOpaqueRanges` that it is measuring**. Its "uncovered: 0" is true by construction. My independent token-level oracle finds 136 comments that oracle cannot see.

Not firing today: both guard suites are green at this head, and no uncovered comment currently spells either pattern. It is armed, not live — the same standing as pass-1 finding 1.

Cheapest sound fix: collect comments per **token** rather than per node — recurse `sf.getChildren()` and take `ts.getLeadingCommentRanges(content, token.getFullStart())` for each leaf (skipping JSDoc kinds, which live inside the comment already). That covers both shapes and is what my oracle does. Then add a fixture in *each* uncovered position to T3 and to the `:613` prose test, and make the `NC` mutation row name placement explicitly — otherwise the next round re-ships a fixture-shaped proof.

---

### 🟡 Minor — `89154448`'s stated reason for leaving KL-p/KL-r alone is factually wrong: #1885 and #1899 have been closed since 2026-07-29 *(cleanup nit)*

`docs/testing/fs38-wave3-onbox-acceptance.md:2695` and `:2697` still read "**Appears fixed in source, issue still open — verify the SHA under test**", and their Issue column reads "(open — fix appears present in source)". `gh issue view` says both are `CLOSED`, `stateReason: COMPLETED`, closed `2026-07-29T03:14:42Z` — three weeks before this branch was cut. `89154448`'s message says they were checked and "belong to #1885 and #1899, where it remains accurate"; it does not.

Concrete failure: an operator running this run sheet today reads KL-p/KL-r as unresolved upstream, is told to "verify the SHA under test" for a fix whose ticket closed three weeks ago, and files against a closed issue. Same claim class, same file, two lines apart from the rows this PR corrected — this is the sweep's remaining half, not a separate concern.

### 🟡 Minor — the PR body predates all four post-review commits and does not declare the incidental fix *(cleanup nit)*

Body last edited `2026-08-18T02:16Z`; `56b9f8fe`/`a256bc54`/`48952e06`/`89154448` all land after it. Three specific staleness points, all checkable:

- The PR now edits `server/src/workspace/cast-lock.guard.test.ts` (+218 / −85) — the **#1981 cast-lock guard, a different seam from this PR's**. That is an incidental fix surfaced by pass 1, and CLAUDE.md requires it declared in the body ("Also fixed, found in passing: …"). It is not mentioned anywhere.
- Test plan: "`git show --stat` on all three commits (`3a0f55d9`, `a8db44ae`, `e9a7e0c4`) confirms test-only and docs-only changes" — enumerates 3 of the 7 commits now on the branch.
- Test plan, M3 row: "per-file map miss + **global-floor miss**" — `a256bc54` deleted that floor, so the evidence cites an assertion that no longer exists.

---

### Also checked, and sound

- **Merge `48952e06` moved no guard fact.** Re-derived every count against the post-merge tree: `writeEntry(` = `routes/voice-library.ts` 3 @ 250/1035/1326, `tts/clone-voice-resolver.ts` 2 @ 309/810, `tts/synthesise-chapter.ts` 1 @ 1125 — every allowlist line number in the `why` strings is still exact. `updateEntry(` = 4/6/1/1/1, matching `G2_EXPECTED` file-for-file, and `workspace/voice-library.ts`'s two prose hits are still at :248/:253 as the map comment claims. Of the merge's 9 changed files under `server/src`, `chapter-qa-repair.ts` and `aggregate.ts` only *read* `castJsonPath`, `centroids-io.ts` writes `centroidsPath`, and none introduces a `writeEntry(`. Both guard suites green at this head (9 tests), `clone-voice-resolver.revoke-race.test.ts` green (2 tests), all 15 PR checks green.
- **T4 (the duplication pin, `voice-library-write-seam.guard.test.ts:485`) is real, not a placebo.** The two extracted blocks are 5,803 bytes and byte-identical; injecting a one-token drift into the sibling copy in memory makes the comparison `false`. Sentinel-missing and file-missing both fail closed. The rationale for duplicating rather than extracting rests on untracked scratch harnesses in a temp directory, which is thin for 130 lines of build-gate code — but it is disclosed, and it is pinned.
- **Release-notes not-applicable still holds** despite the branch now carrying a `fix(server)` commit: every file that commit touches is a `*.test.ts`, and the PR's full file list contains no production source. Sole `Closes #1826` is present, plain, outside any code span; #1826 is `type:chore`, so no `docs/BACKLOG.md` row; the on-box register's "Deliberately not in this register" entry for #1826 is intact at `:3601`.
- **No sibling call site of the `isOpaque` guard was missed.** All four regex scans that consume opaque ranges — `collectLockRanges:436`, `scanFile`'s `WRITE_RE` and `RM_RE` loops, and `countNonOpaque:284` — check the match index. The gap is in what the range set *contains*, not in who consults it.

---

### Verdict

**Request changes.** Pass 1's three findings are all genuinely resolved, and the over-reach direction is now measurably clean across the whole tree. But the rewrite traded one direction of the bug for the other: 136 real comments in 64 files are not opaque, which re-opens the fabricated-lock hole `56b9f8fe` was written to close (demonstrated `scanFile → null` on a genuinely unlocked `writeJsonAtomic(castJsonPath(dir), cast)`), produces false reds on correct code, and leaves `T3` and the `:613` prose test pinning their fixtures rather than their classes — while three header sentences and the commit's own measurement assert the opposite. The measurement asserts it because its oracle is the implementation under test. Fix the comment collection at the token level, add a fixture in each uncovered position, and correct the two `#1885`/`#1899` rows and the PR body while there.

dudarenok-maker:
## PR review — pass 3 (head 10caa221, depth high)

Deltas only. Adjudicated pass 2's three findings independently, then aimed the rest of the pass at the question the dispatch names: whether the *class* — three rounds of "each fix opened the next defect" — is genuinely closed or has relocated once more.

**Instruments.** I did not reuse the project's harnesses as evidence (I read and re-ran them, but see new finding 2). I built two of my own:

- an **under/over-reach census** whose comment reference comes from `ts.createScanner` trivia, re-seeded past AST literal spans so a regex literal cannot desync it — a token stream, not a tree walk, so it shares no traversal with the leaf descent under test;
- a **35-fixture adversarial battery** covering shapes `server/src` does not contain: JSDoc on overloads, comments in type-only positions (interface body, type literal, type-parameter list, mapped type, `Array<…>`), empty file / comment-only file / no trailing newline, comment after the last token, shebang, decorator→declaration, `enum`/`switch`/`namespace`/`declare module`/class-body closers, `${…}` slots, `import type` braces, sole call-argument slot, and two files with deliberate **syntax errors** (parser error recovery).

Both are control-validated: against the pre-fix tokenizer the census reports **276 uncovered comments in 79 files** (`config/registry.ts` 92) — independently reproducing pass 2's number and `a06415e4`'s — and **21 of the 35 fixtures fail**. Against HEAD: **0 and 0/35**, on both byte-identical copies.

---

### Pass-2 🔴 (comments before punctuation never opaque) — **VERIFIED RESOLVED**

- Census over all **427** non-test files, both guard copies: **0 uncovered comments, 0 over-reaching characters**. The over-reach direction did not regress while the comment direction was fixed — 8 targeted regex-vs-division / nested-template cases (division mis-read as regex, regex holding the pattern, nested template, a real call inside a `${…}` slot, regex after `return`, `//` inside a string) are all clean, and `desync-census` still reports 166 → 0 against `e9a7e0c4`.
- **The sufficiency claim holds, tested directly rather than taken from the header.** `cast-lock.guard.test.ts:381-386` justifies asking only leaves with "an interior node's `pos`/`end` coincide with its first/last leaf's anyway, so an interior node can never contribute a range a leaf does not." I collected comment ranges from **every** node, interior included, and diffed against leaves-only across all 427 files: **0 ranges an interior node contributes that no leaf does, in 0 files**. The justification is not just true in effect, it is true for the stated reason.
- **The three new regression tests are genuinely red against `89154448`**, fixtures extracted verbatim from the shipped files (evaluated out of the source, not retyped) and replayed under the pre-fix tokenizer:
  - T5 (`voice-library-write-seam.guard.test.ts:576`) — `isOpaque → false`, `countNonOpaque → 1`; test wants `true` / `0`.
  - T6 (`:592`) — `isOpaque → false, false`, `countNonOpaque → 2`; test wants `true, true` / `0`.
  - cast-lock's last-line-of-block twin (`cast-lock.guard.test.ts:700`) — pre-fix `scanFile → null` (test wants `writes: 1`) and `collectLockRanges → [{start:104, end:191}]` (test wants `[]`). The in-file comment quotes exactly `{ start: 104, end: 191 }`; my run reproduces it to the character.

  No placebo among them. The `try/catch` shape the author reports catching is not in the shipped fixtures.
- **Both prose tests pin the `isOpaque` guard**, per the dispatch's item 4. Re-running each with `collectLockRanges`' `isOpaque` check removed: the line-1 original goes `writes 1 → null` with `lockRanges [] → [{31,108}]`, and the last-line twin goes `writes 1 → null` with `lockRanges [] → [{104,187}]`. Removing the guard reddens **both**, so the line-1 test is not carried by its twin.
- **Header claims re-audited after three rewrites.** The `276 / 79 / 427 / config/registry.ts 92` figures are exact. The "Both are closed now" retraction at `cast-lock.guard.test.ts:203-226` is candid and accurate. The tightened `NC` row (`voice-library-write-seam.guard.test.ts:92-97`) correctly requires last-line-of-block placement — my fixtures confirm a mid-block comment exercises nothing. One remaining claim in that block is wrong; see new finding 2.
- 14/14 tests green across the three suites at this head; all 15 PR checks green.

### Pass-2 🟡 (KL-p/KL-r call closed issues open) — **RESOLVED IN THE ROWS, MISSED IN THE PREAMBLE** — see new finding 1

`gh issue view` confirms #1885 and #1899 are `CLOSED` / `COMPLETED` / `2026-07-29T03:14:42Z`. Both §6 rows now read "Fixed in source, issue closed (2026-07-29)" with "(closed 2026-07-29)" link cells, each row's technical description and its "if your SHA predates this" caveat preserved verbatim. No regression-test proof is claimed for these two, correctly — none exists.

### Pass-2 🟡 (stale PR body) — **VERIFIED RESOLVED**

The incidental `cast-lock.guard.test.ts` fix is now declared under an explicit "Also fixed, found in passing" heading; the M3 test-plan row no longer cites the deleted `global-floor miss` (0 hits for "global floor" in the body); the three-commit scope is now framed rather than left implied. Independently re-verified while I was there: both `fix(server)` commits (`56b9f8fe`, `a06415e4`) touch **only** the two `*.guard.test.ts` files, so the release-notes not-applicable still holds on its stated reason; `Closes #1826` is at body line 5, plain, outside any code span; on-box surfaces are untouched and none is owed.

---

### 🟡 New finding 1 — `10caa221`'s sweep missed the reading guide, and the commit asserts the sweep is complete *(cleanup nit)*

`docs/testing/fs38-wave3-onbox-acceptance.md:60-77`.

`10caa221`'s message states: *"Sweeping the file for 'open' characterisations now returns nothing."* It returns two lines:

```
:60  - **Two consent-bypass gaps — both now appear fixed in source, but both
:61    issues are still open. Verify the SHA under test before trusting either as
:62    closed.**
...
:75    GitHub issue itself is still open, so confirm the commit is actually in
:76    your SHA before treating it as fixed.
```

That bullet is about #1885 (§6 KL-p) and #1899 (§6 KL-r) — the exact two rows this commit corrected, named by link at `:63` and `:72`. Both closed `2026-07-29`, three weeks before the branch was cut.

**Concrete failure.** The bullet sits in the sheet's *"how to read this"* preamble, i.e. where an operator starts. They read that both consent-bypass gaps have open issues and must "verify the SHA under test before trusting either as closed", then reach §6 and find both rows saying "Fixed in source, issue closed (2026-07-29)". The document contradicts itself top-vs-body, and the top is the wrong half. A tester who trusts the preamble re-verifies a fix that shipped and was ticketed closed three weeks ago, and files against a closed issue — the precise outcome pass 2 described for the rows.

**Why it was missed, and why that matters here.** A line-oriented grep for an issue link near the word "open" does not hit this: the issue links are at `:63`/`:72` and the word "open" is at `:61`/`:75`, wrapped onto different lines. That is the same mechanism as pass-1 finding 3 — §6 rows corrected, the cross-references at `:1358`/`:2267` left stale — in the same file. This is the fourth round of the identical shape: fix the rows, miss the prose that points at them. `grep -i "still open" <file>` is the check that closes it, and it is what the commit message claims was already run.

Also worth reconciling while there: `:45-46` opens "Three originally-planned/flagged gaps have since landed and are now fixed, **not open**", and its item (3) at `:58-59` names "the manual override-link consent hole" — so the sheet already calls this same gap fixed one bullet above the bullet that calls it open.

Repo-wide sweep for the residue: `grep -rn "1885\|1899" docs/ | grep -i open` returns nothing else, and `docs/features/268` / `271` are clean of this class. Those two lines are the whole of it.

### 🟡 New finding 2 — the shipped header cites an oracle that now shares the algorithm it measures; its zero is true by construction *(cleanup nit)*

`server/src/workspace/cast-lock.guard.test.ts:219-221`, and the PR body's **Verification** paragraph.

The header closes the "Both are closed now" retraction with:

> "…verified at guard level and against an out-of-tree oracle that **descends to leaf tokens independently of the tokenizer it measures**."

The PR body states it harder: *"three oracles that share no traversal with the code under test or with each other — AST-literal starts (over-reach), **leaf-token descent** (under-reach), and a `ts.createScanner` trivia sweep (under-reach)."*

The leaf-token-descent oracle is `coverage2.mjs`. Its `commentsViaTokens` is the shipped comment collection, statement for statement — `node.getChildren(sf)`, recurse, and at `kids.length === 0` take `getLeadingCommentRanges(text, node.pos)` + `getTrailingCommentRanges(text, node.end)`, then one `getLeadingCommentRanges(text, 0)` at the end. That was a genuinely independent oracle *while the tokenizer walked `forEachChild`*. `a06415e4` made the tokenizer a leaf descent, which made the oracle's traversal identical to the code under test. "Descends to leaf tokens" is named in the header as the guarantee of independence; it is exactly the shared property.

**Concrete demonstration**, run against the harnesses in place:

```
coverage-census.mjs (forEachChild oracle) vs the forEachChild tokenizer  ->  0 uncovered   <- the artifact WITHDRAWN last round
coverage2.mjs      (leaf-descent oracle) vs the forEachChild tokenizer  ->  276 uncovered <- it could fail, then
coverage2.mjs      (leaf-descent oracle) vs the leaf-descent tokenizer  ->  0 uncovered   <- same construction, one level up
```

`coverage2.mjs`'s own file header says it exists *because* `coverage-census.mjs` "used the SAME `ts.forEachChild` traversal as the `computeOpaqueRanges` it measured, so any comment `forEachChild` cannot reach was invisible to BOTH -> zero by construction." That sentence now describes `coverage2.mjs` itself. It can still catch an implementation that *deviates* from leaf descent; it cannot test whether leaf descent is the right algorithm, which is the only open question at this point in the chain.

**This does not move the conclusion, and I want to be exact about that.** The leaf descent really is complete — my scanner-trivia census (no tree walk at all), the project's own `scanner-check.mjs` (152 → 0, control-validated when I re-ran it), `agent-brute.mjs`'s raw `//`/`/*` text sweep, and the all-nodes-vs-leaves diff above all agree, and none of them shares the traversal. The defect is that the sentence a future reader will rely on names the one oracle that no longer proves anything, and does not name `scanner-check.mjs`, which does. Given that the two preceding rounds of this PR each turned on precisely this error — an instrument that could not fail — a header sentence re-asserting it is worth correcting rather than leaving as the durable record.

Cheapest fix: in `cast-lock.guard.test.ts:219-221` cite the scanner-trivia sweep as the independent oracle and say plainly that the leaf-descent harness is now a same-algorithm cross-check, not independent evidence; same correction to the PR body's "three oracles" sentence.

---

### Verdict

**Approve, after the two nits.** The defect class this PR kept relocating — regex desync → nested templates → comments before punctuation — is genuinely closed, and this is the first round where that can be said from evidence that does not share a traversal with the code: 0 uncovered comments and 0 over-reaching characters across 427 files under a scanner-trivia reference, 0/35 adversarial fixtures failing where 21/35 fail on the pre-fix tokenizer, the leaf-sufficiency justification confirmed by an all-nodes-vs-leaves diff, and all three new regression tests measurably red against `89154448` for the reasons they claim. Both prose tests pin the `isOpaque` guard rather than each other.

Nothing new is a correctness bug, so per the triage rule this does not re-trigger a pass — fix both 🟡s and merge. Finding 1 is the fourth round of one shape in one file and is closed by a `grep -i "still open"`; finding 2 is the third round of one shape in the evidence, and is closed by naming the right oracle.

Issue #1996, opened 2026-07-31
Reclaim a stranded VRAM pool at render/unload completion, not just on a later admission failure


dudarenok-maker:
> **▶ Picking this up? Start with the runbook comment:** https://github.com/dudarenok-maker/Castwright/issues/1996#issuecomment-5321454729
>
> It carries the ordered steps, the measurement procedure inline, and the decision tree. **Blocked on #2423** (the diagnostics the measurement reads). Design pass ran 2026-08-18 and deliberately produced no lever — see the analysis comments below for why the obvious ones are dead ends.

---

## What

#1976's acceptance criterion 1 is:

> After a chapter render completes and the engine is unloaded, the render
> card's reserved pool returns to near-baseline **without** a process
> restart.

PR #1993 (which closes the rest of #1976) does not discharge this criterion.
It adds a bare `gc.collect()` + `torch.cuda.empty_cache()` reclaim, but that
reclaim only runs on the **admission-failure path** — i.e. only once some
*later* op gets refused with `noCapacity`. If nothing happens to request
capacity after a render finishes (the common case: a render completes, the
user looks at the result, no other op runs for a while), the stranded pool
from that render sits there indefinitely, exactly as before #1993 — the pool
is real, it's just never probed because nothing failed to fit yet.

Reviewed in #1993's pre-merge review (finding M5): `Closes #1976` was
narrowed to `Refs #1976` because of this gap, and this issue is the
follow-up for the remaining criterion.

## Proposed lever (different from #1993's, and arguably better)

A reclaim triggered at **render/unload completion** itself — i.e. when a
chapter render finishes and its engine reports unloaded, run the same
`gc.collect()` + `empty_cache()` reclaim right there, rather than waiting for
a future admission to fail. This closes the gap #1993 leaves and matches
criterion 1's literal wording ("after a render completes ... without a
process restart", not "after a render completes and something else gets
refused").

## Open question — the mechanism is not fully understood

#1976's own measured stranded state (fs-38 Wave 3 on-box acceptance run,
2026-07-31) had `qwen_loaded: false` but **`asr_loaded: true` and
`spk_loaded: true`** — and the render that produced it had triggered **two
sidecar self-exits on the committed-memory ceiling** during the same
session. So which engine's pool was actually stranded — Qwen's (the one that
had just unloaded) or ASR's / the speaker-embedding model's (both still
resident) — has not been established. A render/unload-completion reclaim
[...]

Issue #1976, opened 2026-07-31
A finished render strands ~3.9 GB of reclaimable VRAM, and admission has no way to reclaim it


dudarenok-maker:
## What happens

After a chapter render, the sidecar leaves **~3.9 GB reserved on the render card with the model reported unloaded** — and nothing can reclaim it short of restarting the process. Every later operation sees a card that is ~3.9 GB smaller than it really is, and gets refused.

Observed during the fs-38 Wave 3 on-box acceptance run (2026-07-31), three ways:

- `POST /transcribe` → `503 {"noCapacity":true,"neededMb":3707,"deviceKey":"cuda:0"}` — three times, **with `asr_loaded: true`**, i.e. the model was already resident and a fresh multi-GB reservation was still demanded.
- `POST /api/voice-library/design` → `503 "Not enough GPU memory for qwen (6144MB) — free VRAM or attach a second GPU."` on a box that *has* a second GPU with 15 GB free.
- Both cleared instantly by `POST /api/sidecar/restart`, after which the identical request succeeded.

## Measured (dual-GPU box: `cuda:0` RTX 4070 8187 MB, `cuda:1` RTX 5070 Ti 16302 MB)

| Point in time | `vram_reserved_mb` | `/capacity` cuda:0 `freeMb` | nvidia-smi cuda:0 |
|---|---|---|---|
| Fresh sidecar | 0 | 7068 | 114 MiB |
| After one `/synthesize` (Qwen resident) | 1887 | 5230 | 1952 MiB |
| After explicit `POST /unload {qwen}` | 23 | 7008 | 174 MiB |
| **After a chapter render (splice), `qwen_loaded: false`** | **3968** | **3230** | **3952 MiB** |
| After a load+unload cycle on that stranded state | 48 | 6968 | **389 MiB** |

The last row is the important one. **The stranded memory is fully reclaimable** — a load/unload cycle recovered ~3.2 GB on `cuda:0`. It is not fragmentation that `empty_cache()` cannot compact. It simply never gets called.

## Why

`empty_cache()` is only ever a **side effect of unloading a specific resident model**:

- `QwenEngine.unload()` — `main.py:4961-4993`, gated on `had = self._base is not None or self._design is not None`
- `QwenEngine.unload_design()` — `main.py:5001-5017`, same gate
- `QwenEngine.maybe_free_idle_design()` — `main.py:5040-5057`
- Coqui's equivalents — `main.py:6243-6296` / `6464-6521`

[...]

PR #2415, opened 2026-08-17
docs(docs): replace the #1996 lever design with a measurement-first spec

dudarenok-maker:
## Summary

Phase 1 for #1996 (#1976's last open acceptance criterion). It produces **no lever** — the adversarial pass killed the one this branch first drafted, and the honest next step is a measurement.

**Two designs have now failed for the same reason: nobody has measured what the stranded pool is made of.** PR #2029's `/unload` hook was rejected in review; this branch's first revision (`cff012ec`, an idle-watchdog reclaim) is withdrawn here. Both assumed the pool was reclaimable cache awaiting an `empty_cache()` that never comes. Three things in the repo contradict that:

- An unconditional, process-wide `gc.collect()` + `empty_cache()` **already runs every 60 s** whenever RSS ≥ 8192 MB — no residency check, no device key (`main.py:7915`, `:8613-8614`, `:7953-7960`). That is exactly the action a new watchdog would schedule.
- The repo already records the pool may be unreclaimable in-process on the primary platform: *"On Windows torch has no `expandable_segments` …, so a fragmented reserved pool … `empty_cache()` can't compact it back, so only a fresh process resets it"* (`main.py:7986-7990`).
- Every engine idle TTL is exactly `120.0` s (`main.py:7678`, `:7704`, `:7762`, `:7815`), and each reclaims process-wide when it fires — so a quiet-window trigger at that grain is redundant by construction, and can never fire in the reported state (ASR + SPK resident *and being exercised*).

#1976's "fully reclaimable / not fragmentation" conclusion rests on a load/unload cycle recovering the memory. A **load** is an allocation pass that refills and coalesces segments, so that recovery is equally the signature of fragmentation — it discriminates nothing.

**What this branch ships**

- `docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md` — rewritten as a measurement design: the evidence above, the two candidate mechanisms and why they need unrelated fixes, the four readings that discriminate them, the small read-only diagnostic they require (existing `/debug/memory` is current-device-only with no `inactive_split_bytes`), and a decision tree mapping the reading to the design.
- `docs/testing/1996-stranded-vram-measurement.md` — the executable run sheet, with a results table.
- **§7 records eight constraints** any future lever must satisfy, each found by the review pass and each otherwise re-derived or shipped as a defect. The two sharpest: `_reclaim_stranded_cache`'s in-use check is per-device while the action it takes is process-wide (so a stranded `cuda:0` fires a device-synchronising `cudaFree` into a live render on `cuda:1` — #1976's own topology), and its `_last_reclaim` cooldown map is shared with the admission path, so a new caller can suppress #1993's reclaim and reintroduce the refusal #1993 exists to prevent.

**A live possibility the run sheet must not treat as a null result:** the strand may self-heal at the existing 120 s TTLs, in which case criterion 1 is already satisfied on `main` and #1996 closes on evidence.

## Test plan

Docs-only — no runtime surface. `verify:fast:scoped` green at commit time (1413/1413 `test:hooks`).

Every `file:line` in the spec was verified against `main` at `1d9ea75c` rather than carried over from the issue text; one citation in #1976's body (`main.py:6980-6994`) had already drifted to `:8150-8151` and is flagged inline in §8.

No release-notes entry: pure design/process, no shippable delta.

Refs #1996
Refs #1976

### DIFF
--- a/docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md
+++ b/docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md
@@ -1,304 +1,247 @@
-# Reclaiming a stranded VRAM pool without an admission failure — design
+# What is actually stranded — measurement design for #1996
 
 Status: **draft** ·
 Issue: [#1996](https://github.com/dudarenok-maker/Castwright/issues/1996)
 (`needs-plan`) ·
-Closes on merge: [#1976](https://github.com/dudarenok-maker/Castwright/issues/1976),
+Parent: [#1976](https://github.com/dudarenok-maker/Castwright/issues/1976),
 whose criterion 1 is #1996's whole content ·
-Supersedes attempt 1: [PR #2029](https://github.com/dudarenok-maker/Castwright/pull/2029)
-(its `/unload` hook was rejected with a Critical and removed) ·
+Prior attempts: [PR #2029](https://github.com/dudarenok-maker/Castwright/pull/2029)
+(the `/unload` hook — rejected in review) and **this document's own first
+revision** (`cff012ec`, the idle-watchdog reclaim — withdrawn before approval;
+see §1) ·
 Follows: [PR #1993](https://github.com/dudarenok-maker/Castwright/pull/1993),
-which shipped the admission-failure-path reclaim this design deliberately does
-not touch.
+the shipped admission-failure-path reclaim.
 
 ---
 
 ## Summary
 
-A finished render leaves a reserved-but-unallocated PyTorch caching-allocator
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
**PR review gate — exempt (docs-only).** The changed-file set is entirely under `docs/**`, which is the file-set test CONTRIBUTING.md's "Doc-only PR fast-path" uses; no review pass runs.

Noting for the record that the *spec in this PR* is itself the output of a Premium-tier `assumption-checker` pass — that pass is what withdrew the design this branch originally carried, and its findings are recorded as §7 of the spec rather than left in a transcript.

Issue #2464, opened 2026-08-18
[agent instructions][cline-glm][task] sitting packs step 1 — write the sitting plan that bins all 74 rows


dudarenok-maker:
Step 1 of the #2453 chain (wave 2 of #2435).

## What

Write `docs/testing/onbox-sitting-plan.md` — the plan of record for this whole wave. It bins **all 74 rows** of `docs/testing/onbox-acceptance-register.md` into exactly one of three sets: a named operator sitting pack, wave-3 agent-runnable, or blocked-on-acquisition. It also fixes the **shared pack format** every later child in this chain follows.

**You write no packs.** Nine pack children run after you and each writes one. What you write is the index they obey.

### The proposed binning — verify it, then record it

This is the coordinator's proposal, derived from the audit's hardware and time fields. **Your job is to check it and correct it**, then write it down as the plan of record. Nothing downstream re-litigates it.

**Operator sittings (52 rows):**

| Pack file | Rows | Est. min |
|---|---|---|
| `onbox-sitting-two-card-boot.md` | A2, A3, A8, A18 | 110 |
| `onbox-sitting-vram-contention.md` | A5, A19, A20, A25, A28, A34, A35, A36 | 140 |
| `onbox-sitting-voice-design.md` | A4, A6, A7, A14, A15, A17, A30 | 155 |
| `onbox-sitting-qa-gate.md` | A9, A10, A11, A12, A13, A21, A22, A23, A37 | 155 |
| `onbox-sitting-cloning-identity.md` | A24, A26, A31, A32, A44, A45, A46, A47 | 170 |
| `onbox-sitting-multilanguage.md` | D1, D2, A38, E4 | 165 |
| `onbox-sitting-device-browser.md` | E1, E2, E3, E5, E6, E9, E10 | 150 |
| `onbox-sitting-fs38-wave3.md` | A1 | multi-hour, several sittings |
| `onbox-sitting-blocked-prerequisites.md` | F1, H1, H2, D3 | 50, blocked |

**Wave-3 agent-runnable (22 rows), excluded from every pack:** A16, A27, A29, A33, A39, A40, A41, A42, A43, B1, B2, B3, B4, C1, C2, C3, C4, E7, E8, E11, G1, G2.

[...]

Issue #2463, opened 2026-08-18
[agent instructions][cline-glm][task] sitting packs step 2 — pack the 2-card-boot sitting (A2, A3, A8, A18)


dudarenok-maker:
Step 2 of the #2453 chain (wave 2 of #2435).

## What

Write **one** sitting pack: `docs/testing/onbox-sitting-two-card-boot.md`, covering register rows **A2, A3, A8, A18**.

Everything that needs the **2-card boot** (8 GB RTX 4070 + 16 GB RTX 5070 Ti over OcuLink), and nothing else.

Estimated content for this sitting: **~110 minutes** — state your own recomputed total in the pack header. The plan of record for this wave is `docs/testing/onbox-sitting-plan.md`, written by step 1 of this chain; **read it first** — it fixes the pack format you must follow and the binning you must not re-litigate.

## What is specific to this pack

This pack exists **because** the eGPU is not hot-pluggable. Interleaving 1-card and 2-card rows costs a reboot each way, so all 2-card work is done in one sitting and none of it appears in any other pack. A8 is a mixed row — only its step 5 needs the 12/16 GB card; say plainly which of its steps belong to this sitting and which ride with the single-card packs. **A2 is AMBIGUOUS** in the audit (whether plan-264 rows 6–8 count as already exercised): mark it, do not resolve it.

## Background you need

`docs/testing/onbox-acceptance-register.md` records **on-box acceptance still owed** — behaviour only real hardware can prove. It stands at **74 owed rows** across eight groups, A–H, grouped by hardware prerequisite.

Wave 1 of this campaign (#2436, shipped as PR #2452) produced `docs/testing/onbox-acceptance-staleness-audit.md`, which carries **one section per row** with its verdict, evidence, what remains owed, **the hardware still required** and **an estimated box time**. That document is your primary input. Its roll-up: **69 STILL OWED, 0 PROPOSE DISCHARGE, 2 SHRUNK, 3 AMBIGUOUS**, totalling **1,690 minutes (~28 hours)**.

**Nothing has been retired.** The register still shows all 74 rows owed, by design: a row comes out only when the acceptance was actually run and recorded, or the owner confirms it. You are writing the runbooks that make running them cheap — you are not retiring anything.

## The audit is trustworthy but not proven — treat it as a lead, not a fact

[...]

Issue #2453, opened 2026-08-18
ops — turn the 74 owed on-box rows into 2-3 hour sitting packs


dudarenok-maker:
Wave 2 of #2435. Wave 1 (#2436) shipped as PR #2452, merged 2026-08-18.

## What

Turn the 74 still-owed register rows into **sitting packs** — per-sitting runbooks sized to a **2–3 hour** session at the box, so the operator's time is spent executing rather than deciding what to run, in what order, on which hardware.

Each pack is one file, `docs/testing/onbox-sitting-<slug>.md`, and covers one sitting: preconditions once at the top, an ordered procedure, concrete observations, `Result:` blanks left empty, and teardown.

## What wave 1 established

From `docs/testing/onbox-acceptance-staleness-audit.md` (on `main`):

| Verdict | Count |
|---|---|
| STILL OWED | 69 |
| PROPOSE DISCHARGE | **0** |
| SHRUNK | 2 |
| AMBIGUOUS | 3 |

**Total estimated box time: 1,690 minutes (~28 hours)** across the 72 rows carrying a numeric estimate; A1 is "multi-hour" and F1 is unestimated. Nothing was retired — the audit proposes, the owner disposes, and the register still shows all 74 rows owed.

The largest buckets: **single 8 GB card — 35 rows, 745 min**; **real workspace, no GPU — 11 rows, 460 min** (of which C2 alone is 300); **phone / Mac / browser — 5 rows, 90 min**; **2-card boot — 3 rows, 85 min**.

## Two things that shape every pack

**Seven run sheets already exist.** `fs38-wave3-onbox-acceptance.md` (3,264 lines, covers A1 and A44), `cast-id-drift-onbox-acceptance.md` (A32, A33, A43, B3), `ort-marker-onbox-acceptance.md` (A39–A42, E9), `night-watch-reanalysis-onbox-acceptance.md` (group C), `attribution-collapse-visibility-onbox-acceptance.md` (E11), `clone-readiness-gate-onbox-acceptance.md` (A31), `sidecar-evict-latency-onbox-acceptance.md` (A25). **A pack sequences and cites those; it never re-authors them.** Duplicating a criteria list is how the two copies drift apart.

[...]

Issue #2435, opened 2026-08-18
ops — clear the on-box acceptance register: a staged campaign


dudarenok-maker:
## What

A staged campaign to clear `docs/testing/onbox-acceptance-register.md`, which stands at **74 owed rows** across 8 hardware groups, with the oldest debt dated **2026-06-01**. This is the tracking issue for the arc; each wave gets its own parent and its own agent chain.

The register's governing rule constrains the whole campaign: **a row comes out only when (a) the acceptance was run on the box and the result recorded, or (b) the repo owner explicitly confirms it was exercised on a live book.** No amount of agent reasoning retires a row. So the campaign is built around spending the operator's box time well, not around avoiding it.

## The waves

**Wave 1 — staleness audit (agent, no hardware).** Verdict all 74 rows against their cited plans, issues, PRs, and the automated coverage that landed since. The register warns about itself: *"Plan prose is frequently not updated after later work discharges it… A stale row is worse than a missing one: it sends you to run something already done."* Three entries were wrong within a day of the register's first version. With the oldest debt ~2.5 months old, a real fraction of the 74 is likely already dead. The audit proposes discharges with evidence; the owner confirms them. It also records, per row, **the hardware still required and an estimated box time** — the two fields wave 2 is built from.

**Wave 2 — session packs (agent, no hardware).** Turn the surviving rows into per-sitting runbooks sized to a **2–3 hour sitting**: ordered steps, copy-paste commands, concrete observations, `Result:` blanks. Batched by hardware prerequisite, with the **2-card boot work in its own sitting** — the eGPU is not hot-pluggable, so interleaving 1-card and 2-card work wastes a reboot each way.

**Wave 3 — the no-hardware rows, run by agents.** A27, A29, A33, A39–A43, E11, G1, G2 self-describe as needing no GPU — sidecar venv only, a real workspace with the server stopped, or GitHub Actions itself. Agents run these on this box and record the outcome, which satisfies the register's rule (a) literally. This is the only bucket where the owed count drops without operator time. **Approved to include the real-workspace rows** (A33, A43, E11).

**Wave 4 — sittings and the fold (operator + agent).** The operator runs a pack; an agent folds the filled sheet back across all three surfaces — register, per-feature run sheet, and the live view HTML — and publishes the live view to its existing artifact URL.

## Routing findings

Anything an acceptance run turns up routes per CLAUDE.md's incidental-findings protocol:

[...]

PR #2470, opened 2026-08-18
fix(scripts): scope the intra-repo link-scan gate to staged sources

dudarenok-maker:
## Summary

- `review-gate-mechanism.test.mjs`'s intra-repo link scan (`linkScanSet()`) walked the whole `docs/testing/**` tree unconditionally, so any doc there that legitimately forward-links to a not-yet-written sibling fails `test:hooks` — blocking every intermediate commit in a multi-step doc chain, not just the one that would actually introduce the dangling link.
- This is exactly what happened to the `#2453` on-box-sitting-pack chain: `onbox-sitting-plan.md` (from #2464) forward-links to 8 sibling pack files in its §2.1 table; only the last one to land makes the links resolve. Every child pack's commit failed `test:hooks` on links it never touched — see #2463.
- `linkScanSet()` now narrows to sources that are part of the commit being staged, when something is staged (the pre-commit case). A clean index — a plain `npm run test:hooks` run or the CI `verify.yml` leg — still gets the full, everything-on-disk scan, so the guard's coverage of the committed tree is unchanged.
- Added `stagedFiles()`, a small helper around `git diff --cached --name-only`, with a paired unit test exercising it against a throwaway git repo.

**One reasoning trap worth flagging for review:** the first draft of the paired test broke this repo's real git history — its throwaway-repo `git init`/`add`/`commit` calls inherited `GIT_DIR`/`GIT_INDEX_FILE` env vars from the pre-commit hook that was running the test, so they silently operated on *this* repo's real index instead of the tmpDir repo, committing a stray file into a real commit. Fixed by stripping every `GIT_*` env var before spawning those subprocesses (in both the test and in `stagedFiles()` itself, since it's the same hook-invoked-from-a-hook shape in production).

## Test plan

- [x] `node --test scripts/tests/review-gate-mechanism.test.mjs` — all 26 assertions pass, including the new `stagedFiles()` test.
- [x] Verified against the actual repo state: confirmed no stray commits/files land in the real repo when the new test runs (both standalone and under the real pre-commit hook).
- [x] `npm run verify:fast:scoped` (pre-commit) passed for this change locally.

Refs #2435, unblocks #2463

### DIFF
--- a/scripts/tests/review-gate-mechanism.test.mjs
+++ b/scripts/tests/review-gate-mechanism.test.mjs
@@ -281,6 +281,39 @@ function headingAnchors(file) {
   return anchors;
 }
 
+/** Files staged for the commit this hook run is gating, or `null` when there
+ *  is nothing staged (e.g. a plain `npm run test:hooks` / CI run against an
+ *  already-committed tree) — the signal linkScanSet() uses to tell "gating a
+ *  commit" apart from "auditing the whole repo".
+ *
+ *  Strips the `GIT_*` env vars (GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, …)
+ *  git hooks set on their own process before spawning this one. Passing only
+ *  `cwd` is not enough to redirect a `git` subprocess: those vars, when
+ *  present, override `cwd` for repo discovery — inherited unmodified here
+ *  they would point `git` at the caller's OWN repo/index regardless of `cwd`.
+ *  This mattered even for the REPO_ROOT default: `stagedFiles()` runs from
+ *  inside pre-commit for its real job, so this default call is the same
+ *  hook-env case as the tmpDir test below, just without the tmpDir masking
+ *  it — omitting the strip here silently no-ops in production too. */
+function stagedFiles(cwd = REPO_ROOT) {
+  const env = { ...process.env };
+  for (const key of Object.keys(env)) {
+    if (key.startsWith('GIT_')) delete env[key];
+  }
+  let out;
+  try {
+    out = execFileSync('git', ['diff', '--cached', '--name-only', '--diff-filter=ACMR'], {
+      cwd,
+      env,
+      encoding: 'utf8',
+    });
+  } catch {
+    return null;
+  }
+  const files = out.split('\n').filter(Boolean).map((p) => resolve(cwd, p));
+  return files.length > 0 ? new Set(files) : null;
+}
+
 /** The scan set is DERIVED, not hand-listed: the two root governance docs,
  *  every markdown file under .claude/skills/**, and every markdown file under
  *  docs/testing/**. A hand-list would reproduce the enumeration trap Task 4
@@ -292,7 +325,18 @@ function headingAnchors(file) {
  *  invisibly. Historical plan docs under docs/ (outside docs/testing/) are
  *  deliberately OUT of scope: measured 2026-08-13, they carry 15 pre-existing
  *  dangling links (relative paths written as if from the repo root), none of
- *  which this work touches. */
+ *  which this work touches.
+ *
+ *  When something is staged (the pre-commit case), the set is further
+ *  narrowed to sources that are themselves part of THIS commit — a doc that
+ *  intentionally forward-links to siblings a later commit will add (e.g. a
+ *  multi-step chain's plan-of-record) would otherwise fail every intermediate
+ *  commit in the chain until the last sibling lands, even though none of
+ *  those commits touched the forward-linking file (#2463). A full,
+ *  everything-on-disk scan still runs whenever nothing is staged — a normal
+ *  `npm run test:hooks` or the CI `verify.yml` leg — so the guard's coverage
+ *  of the committed tree is unchanged; only the pre-commit-time false
+ *  positive on a not-yet-complete chain is narrowed. */
 function linkScanSet() {
   const skillsRoot = join(REPO_ROOT, '.claude', 'skills');
   const skillDocs = readdirSync(skillsRoot, { recursive: true, withFileTypes: true })
@@ -302,12 +346,14 @@ function linkScanSet() {
   const testingDocs = readdirSync(testingRoot, { recursive: true, withFileTypes: true })
     .filter((d) => d.isFile() && d.name.endsWith('.md'))
     .map((d) => join(d.parentPath ?? d.path, d.name));
-  return [
+  const all = [
     join(REPO_ROOT, 'CLAUDE.md'),
     join(REPO_ROOT, 'CONTRIBUTING.md'),
     ...skillDocs,
     ...testingDocs,
   ];
+  const staged = stagedFiles();
+  return staged === null ? all : all.filter((f) => staged.has(resolve(f)));
 }
 
 test('githubAnchor matches GitHub slugging, including runs of spaces and non-ASCII', () => {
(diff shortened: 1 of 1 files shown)


dudarenok-maker:
## PR review — pass 1 (head 84bd70e7, depth medium)

All paths below are `scripts/tests/review-gate-mechanism.test.mjs` unless stated otherwise. Every claim here was measured — on a throwaway repo driven by a real `pre-commit` hook (git 2.54.0.windows.1), and against a detached worktree at 84bd70e7.

### Correctness bugs

**C1 — [High] `stagedFiles()` strips `GIT_INDEX_FILE`, so under `git commit -a` and `git commit -- <path>` it reads the wrong index and the scan skips the file being committed.** `:300-302`, consumed at `:355-356`.

This repo has already measured and litigated exactly this, for exactly this primitive. `scripts/git-env.mjs:33-56` (the #2216 correction) removed `GIT_INDEX_FILE` from `scrubGitEnv()` deliberately, and `scripts/verify-cache.mjs:812-820` says it in as many words about its own `stagedDiffFiles()` — the same `git diff --cached --name-only` call, from the same hook:

> Deliberately does NOT strip GIT_INDEX_FILE (#2216 correction). `stagedDiffFiles()` reads the index the in-flight commit is about; a hook stages `git commit -a` and `git commit -- <path>` through a TEMPORARY index and hands this function exactly that path via GIT_INDEX_FILE. Scrubbing it would make `stagedDiffFiles()` read `.git/index` instead of the real temp index — the wrong staged set, or an empty one.

Measured, inside a real `pre-commit`:

| commit form | `GIT_INDEX_FILE` handed to the hook | honouring it (correct) | stripping it (this PR) |
|---|---|---|---|
| `git commit -a` | `<abs>/.git/index.lock` | `docs/testing/a.md` | *(empty)* |
| `git add other.md` then `git commit -a` | `<abs>/.git/index.lock` | `a.md`, `other.md` | `other.md` |
| `git commit -- docs/testing/a.md` | `<abs>/.git/next-index-30396.lock` | `a.md` | `zzz.md` — a file **not in the commit at all** |

Failure scenario, run end to end through `linkScanSet()`'s actual filter inside a real hook. Pre-stage an innocent `docs/testing/other.md`; edit `docs/testing/a.md` in the worktree only, adding a link `./ghost.md` that does not exist; `git commit -a`:

```
PR #2470 stagedFiles()  (strips GIT_INDEX_FILE):
  scanned : [other.md]
  verdict : PASS (no dangling links found)
scrubGitEnv() semantics (keeps GIT_INDEX_FILE):
  scanned : [a.md, other.md]
  verdict : FAIL -> a.md -> ./ghost.md
```

The guard goes green on the commit that introduces the dangling link. Note the pincer: `verify-cache.mjs`'s scope filter *does* honour `GIT_INDEX_FILE`, so it correctly sees `a.md` change and schedules `test:hooks` **because of `a.md`** — and then `linkScanSet()` excludes `a.md` from the scan.

The two call sites need different environments, and the PR gives them one. The tmpDir fixture genuinely does need `GIT_INDEX_FILE` gone — confirmed: with `GIT_INDEX_FILE` alone pointed at another repo's index and no `GIT_DIR` set, a nested `git add` run in a different `cwd` writes into that other index. That, not `GIT_DIR`, is the mechanism behind the stray commit described in the PR body, because a hook does **not** export `GIT_DIR` (measured `GIT_DIR=[]`, matching `git-env.mjs`'s point 1). Production must keep the var. Suggested shape: `stagedFiles()` defaults to `scrubGitEnv()` from `scripts/git-env.mjs`; the test passes an env that additionally drops `GIT_INDEX_FILE` for its tmpDir call.

**C2 — [High] The narrowed link scan reports PASS having scanned 0 of 22 sources, and nothing distinguishes that from a real pass.** `:355-356`.

`stagedFiles()` returns *all* staged paths repo-wide; `all.filter(...)` then intersects them with the 22 link-scan sources. Stage anything that is not one of those 22 and the intersection is empty. Measured at 84bd70e7 with a single non-source file staged:

```
staged set             : scratch-not-a-scan-source.txt
FULL scan set size     : 22
ACTUAL linkScanSet size: 0
```

`node --test` then reports `intra-repo .md links in the governance docs and skills resolve to real files and headings` as passing, 26/26.

This is not an exotic shape — it is the common one, because `test:hooks` is scheduled by globs (`scripts/**`, `.github/**`, `.husky/**`, `pinokio-scripts/**`) that share no members with the scan set. **This PR's own pre-commit is an instance**: it stages only `scripts/tests/review-gate-mechanism.test.mjs`, so its link scan ran over zero files — which is why "`npm run verify:fast:scoped` (pre-commit) passed" in the test plan is not evidence about this guard.

This file already knows the hazard and guards it one test over, at `:660-670`: `MIRRORED_FILES` gets an exact-list assertion precisely because "a `> 0` check is satisfied just as easily by a wrong or partial list", so that `skills:sync` "cannot report success having written nothing". The narrowed scan needs the same non-vacuity assertion — that the set is non-empty, or that it contains the staged sources it should.

**C3 — [Medium] The paired test cannot fail on the defect it was written for, in any run CI or a developer actually performs.** `:557-598`.

The test strips `GIT_*` in *its own* fixture env at `:575-578`, then calls `stagedFiles(tmpDir)`. When no `GIT_*` var is set ambiently — `npm run test:hooks`, `node --test`, and every `verify.yml` run — the strip inside `stagedFiles()` at `:300-302` is a no-op, so its presence or absence changes nothing. Positive control at 84bd70e7, deleting **only** line 301 (the production strip, leaving the test's own intact):

```
RUN B: production strip deleted, clean env           -> tests 26 | pass 26 | fail 0
RUN C: production strip deleted, GIT_INDEX_FILE set  -> tests 26 | pass 24 | fail 2
```

The regression that wrote into this repo's real git history is therefore pinned by nothing that runs in CI; it goes red only when a `GIT_*` var happens to be set ambiently. To be a real regression test it has to set the hook vars itself — e.g. point `process.env.GIT_INDEX_FILE` at a decoy index, call `stagedFiles(tmpDir)`, and assert the answer still comes from `tmpDir`.

**C4 — [Medium] The behaviour change itself — `linkScanSet()`'s narrowing — has no test.** `:355-356`.

The one new assertion covers `stagedFiles()` in isolation. No test exercises the `staged !== null` branch of `linkScanSet()`, and none can under `node --test`, which always runs with a clean index and so always takes the `staged === null` path. The #2463 scenario this PR exists to fix — a forward-linking doc present on disk but absent from the staged set — is asserted nowhere.

**C5 — [Medium] A non-ASCII source filename is silently dropped from the scan.** `:305` / `:313`.

`core.quotePath` defaults to true, so `git diff --cached --name-only` C-quotes such paths. Measured:

```
$ git diff --cached --name-only --diff-filter=ACMR
"docs/testing/caf\303\251-plan.md"
docs/testing/with space.md
```

`resolve(cwd, '"docs/testing/caf\\303\\251-plan.md"')` yields a path matching nothing in `all`, so the file is filtered out and never scanned — a fail-open, not a crash. Spaces are fine; non-ASCII is not. No scan source has a non-ASCII name today, so this is latent — but the header at `:317-324` states the scan set is derived rather than hand-listed exactly so it covers files that do not exist yet, and this puts a name-shaped blind spot back underneath that. Fix: `-z` and split on NUL, or `git -c core.quotePath=false`.

**C6 — [Medium] The scrub is hand-rolled rather than routed through `scrubGitEnv()`, which makes it case-sensitive and invisible to the repo's own guard.** `:300-302`.

`scripts/git-env.mjs:70-78` matches case-insensitively and documents why: `{ ...process.env }` snapshots whatever casing Windows stored, so a `delete` keyed on the canonical uppercase name alone "would leave a `git_dir`-cased survivor". `key.startsWith('GIT_')` is exactly that case-sensitive form. Separately, `scripts/tests/git-scrub.test.mjs:165-169` excludes `scripts/tests/**` from its scan — reasonable for test fixtures, but `stagedFiles()` is production code running in every `test:hooks` invocation, so this git call now sits in the one directory the chokepoint guard cannot see. A later edit deleting the scrub would be caught by neither `git-scrub.test.mjs` (out of scope) nor the paired test (C3).

### Cleanup nits

**N1** — `:337-339` claims "only the pre-commit-time false positive on a not-yet-complete chain is narrowed". False: true positives are narrowed away too. A commit that stages `docs/testing/a.md` while deleting or renaming a file `CLAUDE.md` links to no longer re-scans `CLAUDE.md`, because `CLAUDE.md` is not staged. (`--diff-filter=ACMR` handles the rename correctly on the source side — `--name-only` prints the destination, verified — so the renamed file itself is scanned; it is the unstaged documents pointing *at* the old path that stop being checked.) Coverage of the committed tree in CI is genuinely unchanged; the sentence should say pre-commit loses true positives as well, and that CI is what still catches them.

**N2** — The PR body omits the explicit not-applicable declarations CLAUDE.md's before-shipping checklist asks be stated rather than silently skipped: step 1 (regression plan), step 3 (on-box acceptance), step 5 (release-notes pair). All three are genuinely N/A here — test-infra only, no shippable delta — but the checklist wants the sentence.

**N3** — `Refs #2435` links the tracking chore, and #2463 appears only in prose as "unblocks". Nothing is closed by this PR and the guard defect itself has no issue. The `pr-issue-link` check is green, so this is bookkeeping only. (Unrelated to the diff: the `Check / record CLA signature` check is currently failing on this PR.)

### Answers to the three questions in the dispatch

1. **Is the `GIT_*` strip correct and sufficient — is there a hook var git honours that is not `GIT_`-prefixed?** Every variable that answers "which repository" or "which index" is `GIT_`-prefixed (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_COMMON_DIR`, `GIT_NAMESPACE`, `GIT_CEILING_DIRECTORIES`, `GIT_DISCOVERY_ACROSS_FILESYSTEM`, `GIT_PREFIX`). The non-`GIT_` variables git reads — `HOME`, `XDG_CONFIG_HOME`, `EMAIL`, `PATH` — are not hook-context vars and do not redirect repo or index. So the strip is not too narrow; it is **too broad**, and that is the bug (C1). Its one genuine narrowness is case-sensitivity (C6).

2. **Other correctness issues on the production path?** Yes — C1 (wrong index for `-a` and pathspec commits), C2 (vacuous zero-file pass), C5 (quoted non-ASCII paths). On the two specifics asked: `--diff-filter=ACMR` does **not** miss a staged rename — `R` is included and `--name-only` prints the destination (`R100 a.md -> renamed.md` prints `renamed.md`). And `git commit --amend` is safe in the common case: an amend with nothing newly staged produces an empty diff, hence `null`, hence a full scan. An amend *with* newly staged changes narrows to just those, leaving the rest of the amended commit unrescanned — real, but caught by CI.

3. **Does this meaningfully weaken the guard in CI or normal workflows?** No security implication: this is a repo-hygiene guard, not a security control — no untrusted input, no shell (argv-array `execFileSync`), constant `cwd`. On protection: **CI is intact.** `verify.yml` checks out with a clean index, so `stagedFiles()` returns `null` and the full 22-file scan runs, and `test:hooks` is correctly scheduled for `docs/testing/**`, `.claude/skills/**`, `CLAUDE.md` and `CONTRIBUTING.md` diffs via `verify-cache.mjs`'s globs and `extraFiles`. The erosion is entirely at pre-commit, and it is larger than the PR describes: C1 turns the guard into a fail-open on the very file being committed, and C2 makes it a no-op on the common commit shape. Deferring target-side breakage (N1) from commit time to PR time would be an acceptable trade on its own. C1 and C2 are not, because they lose the source-side true positive the guard exists for, at exactly the moment it is meant to fire.

The direction of the fix is right and the #2463 false positive is real. C1 and C2 are the blocking ones.

dudarenok-maker:
Closing without merging.

The independent review (comment above) found this approach fundamentally miscalibrated, not just buggy: narrowing `linkScanSet()` to staged sources means the guard scans **zero files** — and reports PASS — for any commit that doesn't touch `docs/testing/**`/`.claude/skills/**`/the two root docs, which is most ordinary commits. That's not a narrow fix for the #2463 doc-chain edge case, it silently disables the guard almost entirely. C1 (stripping `GIT_INDEX_FILE`) is also a real regression this repo already litigated and fixed the other way in #2216 (`scripts/git-env.mjs`).

Fixing this properly would mean rebuilding the check as a before/after diff (fail only on links newly broken by this commit, not all pre-existing breakage) — a much bigger, riskier change to a repo-wide correctness guard than warranted here.

Went with the smaller, contained fix instead: `onbox-sitting-plan.md`'s forward-links to its 7 not-yet-written sibling packs are now plain code spans, not real `.md` links, until each sibling lands — see the commits on `docs/docs-2453-onbox-sitting-packs`. No shared test infrastructure touched. Thanks for catching this before it merged.

PR #2479, opened 2026-08-19
fix(ci): persist winget ffmpeg PATH fix past the install step

### LINKED ISSUE #2478: bug: Cross-OS Verify — winget ffmpeg fallback's PATH fix doesn't survive later steps
## What
The scheduled Cross-OS Verify run on 2026-08-19 (run 32209846581) failed on `Verify (windows-latest)`. Root cause: `.github/workflows/cross-os.yml`'s "Install ffmpeg (Windows)" step falls back to `winget install --id Gyan.FFmpeg` when choco fails, then re-reads Machine+User PATH from the registry into `$env:Path` so `ffmpeg -version` resolves *within that step's own process*. But every later step (`Verify`, `Build`, ...) is a fresh process spawned by the runner service that still carries the job's original PATH — so `npm run verify:quick`'s `test:server` `pretest` (`scripts/preflight-ffmpeg.cjs`) failed with "ffmpeg not on this shell's PATH — but it IS installed", exactly the trap the preflight script documents for developer shells.

## Acceptance
- The winget fallback branch persists ffmpeg's resolved directory to `$GITHUB_PATH` so subsequent steps in the same job see it too.
- No regression to the choco-succeeds path (unaffected — choco already lands on the pre-existing PATH).

## Key files
- `.github/workflows/cross-os.yml` (Install ffmpeg (Windows) step)
- `scripts/preflight-ffmpeg.cjs` (documents the same PATH-staleness trap for local shells)

## Benefit (technical): unblocks the scheduled Cross-OS Verify gate from failing on transient choco-feed outages that trigger the winget fallback.

dudarenok-maker:
## Summary
- The scheduled Cross-OS Verify run on 2026-08-19 (run 32209846581) failed on `Verify (windows-latest)`: the winget ffmpeg fallback re-reads PATH from the registry so `ffmpeg -version` resolves *within that step's own process*, but every later step is a fresh process that still carries the job's original PATH — so `npm run verify:quick`'s `test:server` `pretest` failed with "ffmpeg not on this shell's PATH".
- Fix: append the resolved WinGet Links directory to `$GITHUB_PATH` so it carries into every subsequent step in the job. `release.yml` carries an identical step, so it gets the identical fix.
- Three review rounds (full detail in PR comments): round 1 found the PATH-persist logic crashed under the exact env shape this diff's own `test:hooks` leg hits on `ubuntu-latest` (GITHUB_PATH set, LOCALAPPDATA unset), and that the test harness was mutating the real job's own `$GITHUB_PATH` file. Round 2 found `Add-Content`'s own failure wasn't actually catchable (non-terminating error under the default preference) and that the new tests were placebos that passed against the pre-fix code too. Round 3 (final allowed round) found the LOCALAPPDATA check was folded into a single conjunct that would skip the persist step silently on a real box missing it, with a test that passed for the wrong reason. All fixed and independently verified against the actual pre-fix / mutated script bodies, not just re-read.
- Filed #2480 for the flagged duplication between the two workflow files, with the decision it owes named explicitly.

## Test plan
- [x] `scripts/tests/cross-os-ffmpeg-install.test.mjs` — 14 tests (12 original + 2 added across this PR's rounds), all green, including a scenario that deliberately unsets `LOCALAPPDATA` (mirrors ubuntu-latest's Hooks-tests leg), asserts the explicit warning it now produces, and negative assertions that `$GITHUB_PATH` stays untouched on the choco-succeeds and both-fail paths.
- [x] `npm run test:hooks` full suite green (1415/1415).
- [x] PowerShell parser check on the modified blocks reports zero syntax errors.
- [x] Manually reproduced the exact CI failure this PR fixes, and each round's flagged regression, by running the pre-fix/mutated script bodies directly through pwsh with the relevant env shape, confirming each new assertion actually fails against the code it's meant to catch.
- [ ] On-box: the real winget-fallback path (choco actually failing on a live Windows runner) can only be proven by a live Cross-OS Verify run that hits it — not owed acceptance, just not independently re-triggerable from here; the next scheduled run (or one hitting the winget fallback) will confirm.
- N/A: release notes — this is a CI-only fix with no shippable product-facing delta (CLAUDE.md before-shipping step 5's explicit skip carve-out).

Closes #2478

### DIFF
--- a/.github/workflows/cross-os.yml
+++ b/.github/workflows/cross-os.yml
@@ -118,26 +118,32 @@ jobs:
                 # `npm run verify:quick` runs `pretest`'s ffmpeg preflight
                 # (observed 2026-08-19, cross-os run 32209846581).
                 # $GITHUB_PATH is the documented mechanism for a step to
-                # prepend a directory onto PATH for the rest of the job. A
-                # per-user winget install (no admin elevation, the case on
-                # GitHub-hosted runners) always lands under this WinGet
-                # Links dir. Deliberately its OWN try/catch, outside the one
-                # above: a failure persisting this bookkeeping entry must
-                # never retract an ffmpeg install that actually verified
-                # working, nor fail the whole install silently -- it can
-                # only warn. Guarded on both $env:GITHUB_PATH and
-                # $env:LOCALAPPDATA being set: this step only runs on
-                # windows-latest, where both are always present for a real
-                # job, but scripts/tests/cross-os-ffmpeg-install.test.mjs
-                # executes this same script directly via pwsh (including on
-                # ubuntu-latest, where LOCALAPPDATA is never set) to pin its
-                # control flow against stubbed choco/winget/ffmpeg.
+                # prepend a directory onto PATH for the rest of the job.
+                # winget's Gyan.FFmpeg package places its shim here -- the
+                # exact path observed in the real CI failure this fixes
+                # (cross-os run 32209846581,
+                # C:\Users\runneradmin\AppData\Local\Microsoft\WinGet\Links\ffmpeg.exe).
+                # Deliberately its OWN try/catch, outside the one above: a
+                # failure persisting this bookkeeping entry must never
+                # retract an ffmpeg install that actually verified working,
+                # nor fail the whole install silently -- it can only warn.
+                # `-ErrorAction Stop` forces Add-Content's own failures to be
+                # catchable here (its default is a non-terminating error,
+                # which a bare try/catch does not observe at all). Guarded
+                # on both $env:GITHUB_PATH and $env:LOCALAPPDATA being set:
+                # this step only runs on windows-latest, where both are
+                # always present for a real job, but
+                # scripts/tests/cross-os-ffmpeg-install.test.mjs executes
+                # this same script directly via pwsh -- including with
+                # LOCALAPPDATA deliberately absent, to pin the guard against
+                # exactly the crash that broke PR #2479's first attempt at
+                # this fix (`Join-Path $null ...`).
                 if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA) {
                   try {
                     $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
-                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir
+                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
                   } catch {
-                    Write-Host "warning: failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps: $_"
+                    Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
                   }
                 }
               } else {
--- a/.github/workflows/cross-os.yml
+++ b/.github/workflows/cross-os.yml
@@ -105,31 +105,40 @@ jobs:
                   $env:Path = [System.Environment]::GetEnvironmentVariable('Path', 'Machine') + ';' + [System.Environment]::GetEnvironmentVariable('Path', 'User')
                   & ffmpeg -version | Out-Null
                   $ffmpegOk = ($LASTEXITCODE -eq 0)
-                  if ($ffmpegOk -and $env:GITHUB_PATH) {
-                    # The re-read above only fixes PATH for *this* step's
-                    # process. Every later step (Verify, Build, ...) is a
-                    # fresh process spawned by the runner service, which
-                    # still carries the PATH it had at job start -- so
-                    # without this, ffmpeg resolves here but goes missing
-                    # again by the time `npm run verify:quick` runs
-                    # `pretest`'s ffmpeg preflight (observed 2026-08-19,
-                    # cross-os run 32209846581). $GITHUB_PATH is the
-                    # documented mechanism for a step to prepend a
-                    # directory onto PATH for the rest of the job. A
-                    # per-user winget install (no admin elevation, the case
-                    # on GitHub-hosted runners) always lands under this
-                    # WinGet Links dir -- that observed CI path is more
-                    # reliable here than `Get-Command ffmpeg`, which
-                    # resolves to a shim with no real `.Source` in some
-                    # shells. Guarded on $env:GITHUB_PATH being set at all,
-                    # since it's a GitHub Actions-only mechanism -- absent
-                    # in a local shell or this step's own regression test.
-                    $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
-                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir
-                  }
                 } catch {
                   Write-Host "winget fallback unavailable: $_"
                   $ffmpegOk = $false
+                }
+
+                # The re-read above only fixes PATH for *this* step's
+                # process. Every later step (Verify, Build, ...) is a fresh
+                # process spawned by the runner service, which still
+                # carries the PATH it had at job start -- so without this,
+                # ffmpeg resolves here but goes missing again by the time
+                # `npm run verify:quick` runs `pretest`'s ffmpeg preflight
+                # (observed 2026-08-19, cross-os run 32209846581).
+                # $GITHUB_PATH is the documented mechanism for a step to
+                # prepend a directory onto PATH for the rest of the job. A
+                # per-user winget install (no admin elevation, the case on
+                # GitHub-hosted runners) always lands under this WinGet
+                # Links dir. Deliberately its OWN try/catch, outside the one
+                # above: a failure persisting this bookkeeping entry must
+                # never retract an ffmpeg install that actually verified
+                # working, nor fail the whole install silently -- it can
+                # only warn. Guarded on both $env:GITHUB_PATH and
+                # $env:LOCALAPPDATA being set: this step only runs on
+                # windows-latest, where both are always present for a real
+                # job, but scripts/tests/cross-os-ffmpeg-install.test.mjs
+                # executes this same script directly via pwsh (including on
+                # ubuntu-latest, where LOCALAPPDATA is never set) to pin its
+                # control flow against stubbed choco/winget/ffmpeg.
+                if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA) {
+                  try {
+                    $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
+                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir
+                  } catch {
+                    Write-Host "warning: failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps: $_"
+                  }
                 }
               } else {
                 Write-Host "winget fallback unavailable: no winget command on this runner."
--- a/.github/workflows/cross-os.yml
+++ b/.github/workflows/cross-os.yml
@@ -105,6 +105,28 @@ jobs:
                   $env:Path = [System.Environment]::GetEnvironmentVariable('Path', 'Machine') + ';' + [System.Environment]::GetEnvironmentVariable('Path', 'User')
                   & ffmpeg -version | Out-Null
                   $ffmpegOk = ($LASTEXITCODE -eq 0)
+                  if ($ffmpegOk -and $env:GITHUB_PATH) {
+                    # The re-read above only fixes PATH for *this* step's
+                    # process. Every later step (Verify, Build, ...) is a
+                    # fresh process spawned by the runner service, which
+                    # still carries the PATH it had at job start -- so
+                    # without this, ffmpeg resolves here but goes missing
+                    # again by the time `npm run verify:quick` runs
+                    # `pretest`'s ffmpeg preflight (observed 2026-08-19,
+                    # cross-os run 32209846581). $GITHUB_PATH is the
+                    # documented mechanism for a step to prepend a
+                    # directory onto PATH for the rest of the job. A
+                    # per-user winget install (no admin elevation, the case
+                    # on GitHub-hosted runners) always lands under this
+                    # WinGet Links dir -- that observed CI path is more
+                    # reliable here than `Get-Command ffmpeg`, which
+                    # resolves to a shim with no real `.Source` in some
+                    # shells. Guarded on $env:GITHUB_PATH being set at all,
+                    # since it's a GitHub Actions-only mechanism -- absent
+                    # in a local shell or this step's own regression test.
+                    $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
+                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir
+                  }
                 } catch {
                   Write-Host "winget fallback unavailable: $_"
                   $ffmpegOk = $false
--- a/.github/workflows/cross-os.yml
+++ b/.github/workflows/cross-os.yml
@@ -129,21 +129,29 @@ jobs:
                 # nor fail the whole install silently -- it can only warn.
                 # `-ErrorAction Stop` forces Add-Content's own failures to be
                 # catchable here (its default is a non-terminating error,
-                # which a bare try/catch does not observe at all). Guarded
-                # on both $env:GITHUB_PATH and $env:LOCALAPPDATA being set:
-                # this step only runs on windows-latest, where both are
-                # always present for a real job, but
-                # scripts/tests/cross-os-ffmpeg-install.test.mjs executes
-                # this same script directly via pwsh -- including with
-                # LOCALAPPDATA deliberately absent, to pin the guard against
-                # exactly the crash that broke PR #2479's first attempt at
-                # this fix (`Join-Path $null ...`).
-                if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA) {
-                  try {
-                    $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
-                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
-                  } catch {
-                    Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
+                # which a bare try/catch does not observe at all). Gated on
+                # $env:GITHUB_PATH: absent outside GitHub Actions (a local
+                # shell, or scripts/tests/cross-os-ffmpeg-install.test.mjs's
+                # parser/parse-only checks), where there is nothing to
+                # persist to. $env:LOCALAPPDATA is checked separately, NOT
+                # folded into this same condition -- this step only runs on
+                # windows-latest, where it is always present for a real job,
+                # but a silent skip on the rare box where it genuinely isn't
+                # would recreate #2478 downstream with zero signal. The
+                # regression test deliberately unsets LOCALAPPDATA (mirrors
+                # ubuntu-latest's Hooks-tests leg, which also executes this
+                # script) specifically to pin that the missing case warns
+                # rather than skips quietly.
+                if ($ffmpegOk -and $env:GITHUB_PATH) {
+                  if ($env:LOCALAPPDATA) {
+                    try {
+                      $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
+                      Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
+                    } catch {
+                      Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
+                    }
+                  } else {
+                    Write-Host "::warning::LOCALAPPDATA is not set -- cannot determine the winget-installed ffmpeg directory to persist to `$GITHUB_PATH; later steps in this job may not find ffmpeg on PATH."
                   }
                 }
               } else {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -155,26 +155,31 @@ jobs:
                 # (observed 2026-08-19, cross-os run 32209846581, same step
                 # shape as this file's). $GITHUB_PATH is the documented
                 # mechanism for a step to prepend a directory onto PATH for
-                # the rest of the job. A per-user winget install (no admin
-                # elevation, the case on GitHub-hosted runners) always
-                # lands under this WinGet Links dir. Deliberately its OWN
-                # try/catch, outside the one above: a failure persisting
-                # this bookkeeping entry must never retract an ffmpeg
-                # install that actually verified working, nor fail the
-                # whole install silently -- it can only warn. Guarded on
-                # both $env:GITHUB_PATH and $env:LOCALAPPDATA being set:
-                # this step only runs on windows-latest, where both are
-                # always present for a real job, but
+                # the rest of the job. winget's Gyan.FFmpeg package places
+                # its shim here -- the exact path observed in the real CI
+                # failure this fixes (cross-os run 32209846581,
+                # C:\Users\runneradmin\AppData\Local\Microsoft\WinGet\Links\ffmpeg.exe).
+                # Deliberately its OWN try/catch, outside the one above: a
+                # failure persisting this bookkeeping entry must never
+                # retract an ffmpeg install that actually verified working,
+                # nor fail the whole install silently -- it can only warn.
+                # `-ErrorAction Stop` forces Add-Content's own failures to
+                # be catchable here (its default is a non-terminating
+                # error, which a bare try/catch does not observe at all).
+                # Guarded on both $env:GITHUB_PATH and $env:LOCALAPPDATA
+                # being set: this step only runs on windows-latest, where
+                # both are always present for a real job, but
                 # scripts/tests/cross-os-ffmpeg-install.test.mjs executes
-                # this same script directly via pwsh (including on
-                # ubuntu-latest, where LOCALAPPDATA is never set) to pin its
-                # control flow against stubbed choco/winget/ffmpeg.
+                # this same script directly via pwsh -- including with
+                # LOCALAPPDATA deliberately absent, to pin the guard
+                # against exactly the crash that broke PR #2479's first
+                # attempt at this fix (`Join-Path $null ...`).
                 if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA) {
                   try {
                     $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
-                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir
+                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
                   } catch {
-                    Write-Host "warning: failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps: $_"
+                    Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
                   }
                 }
               } else {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -166,20 +166,29 @@ jobs:
                 # `-ErrorAction Stop` forces Add-Content's own failures to
                 # be catchable here (its default is a non-terminating
                 # error, which a bare try/catch does not observe at all).
-                # Guarded on both $env:GITHUB_PATH and $env:LOCALAPPDATA
-                # being set: this step only runs on windows-latest, where
-                # both are always present for a real job, but
-                # scripts/tests/cross-os-ffmpeg-install.test.mjs executes
-                # this same script directly via pwsh -- including with
-                # LOCALAPPDATA deliberately absent, to pin the guard
-                # against exactly the crash that broke PR #2479's first
-                # attempt at this fix (`Join-Path $null ...`).
-                if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA) {
-                  try {
-                    $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
-                    Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
-                  } catch {
-                    Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
+                # Gated on $env:GITHUB_PATH: absent outside GitHub Actions
+                # (a local shell, or
+                # scripts/tests/cross-os-ffmpeg-install.test.mjs's
+                # parser/parse-only checks), where there is nothing to
+                # persist to. $env:LOCALAPPDATA is checked separately, NOT
+                # folded into this same condition -- this step only runs on
+                # windows-latest, where it is always present for a real
+                # job, but a silent skip on the rare box where it genuinely
+                # isn't would recreate #2478 downstream with zero signal.
+                # The regression test deliberately unsets LOCALAPPDATA
+                # (mirrors ubuntu-latest's Hooks-tests leg, which also
+                # executes this script) specifically to pin that the
+                # missing case warns rather than skips quietly.
+                if ($ffmpegOk -and $env:GITHUB_PATH) {
+                  if ($env:LOCALAPPDATA) {
+                    try {
+                      $ffmpegDir = Join-Path $env:LOCALAPPDATA 'Microsoft\WinGet\Links'
+                      Add-Content -Path $env:GITHUB_PATH -Value $ffmpegDir -ErrorAction Stop
+                    } catch {
+                      Write-Host "::warning::failed to persist ffmpeg's install directory to `$GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: $_"
+                    }
+                  } else {
+                    Write-Host "::warning::LOCALAPPDATA is not set -- cannot determine the winget-installed ffmpeg directory to persist to `$GITHUB_PATH; later steps in this job may not find ffmpeg on PATH."
                   }
                 }
               } else {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -141,6 +141,29 @@ jobs:
                   $env:Path = [System.Environment]::GetEnvironmentVariable('Path', 'Machine') + ';' + [System.Environment]::GetEnvironmentVariable('Path', 'User')
                   & ffmpeg -version | Out-Null
                   $ffmpegOk = ($LASTEXITCODE -eq 0)
+                  if ($ffmpegOk -and $env:GITHUB_PATH) {
+                    # The re-read above only fixes PATH for *this* step's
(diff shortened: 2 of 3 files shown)


dudarenok-maker:
## PR review — pass 1 (head d2fd03e69e871027190a5c03068e97088f4c7271, depth medium)

Scope: `.github/workflows/cross-os.yml` (+22), `.github/workflows/release.yml` (+23) — the whole diff. Verified: ran `scripts/tests/cross-os-ffmpeg-install.test.mjs` under three environments (local / `GITHUB_PATH` set / `GITHUB_PATH` set + `LOCALAPPDATA` unset), probed `Join-Path`, `Add-Content` and `Get-Command` semantics directly under pwsh 7.6.5, and read this PR's own `verify.yml` run (32303835840) before adversarially probing the new block's guard, its error containment, and whether the cited harness actually reaches it.

### 🔴 Blocking — `Join-Path $env:LOCALAPPDATA` throws on every non-Windows runner, and the guard does not stop it. This PR's own required check is already red.

`cross-os.yml:108,127` / `release.yml:144,164`.

The guard is `if ($ffmpegOk -and $env:GITHUB_PATH)`. The block's own comment states the premise it rests on:

> Guarded on `$env:GITHUB_PATH` being set at all, since it's a GitHub Actions-only mechanism -- absent in a local shell **or this step's own regression test**.

That premise is false. The regression test runs *inside* GitHub Actions. `.github/workflows/**` is a declared input of the `test:hooks` step (`scripts/verify-cache.mjs:109`), so this very diff scopes the `Hooks tests` leg **in** — and that leg runs on `ubuntu-latest` (`verify.yml:193-196`), where `GITHUB_PATH` is set and `LOCALAPPDATA` is not. `Join-Path`'s `-Path` rejects null, so this is a **terminating** `ParameterBindingValidationException` — catchable, and thrown even under `$ErrorActionPreference = 'Continue'` (verified on pwsh 7.6.5). It lands in the enclosing `catch` at `cross-os.yml:130` / `release.yml:167`, which sets `$ffmpegOk = $false`.

Exact repro, reproduced locally by simulating that environment:

```
$ env -u LOCALAPPDATA GITHUB_PATH=/tmp/gh_path node --test scripts/tests/cross-os-ffmpeg-install.test.mjs
✖ .github/workflows/cross-os.yml: ... choco fails, winget rescues it (15347ms)
✖ .github/workflows/release.yml: ... choco fails, winget rescues it (15361ms)
  AssertionError: expected the post-winget ffmpeg re-check to succeed
  actual: "Attempt 1/3: choco reported success but ffmpeg is not runnable ... Trying the winget fallback.
           winget fallback unavailable: Cannot bind argument to parameter 'Path' because it is null.
           Attempt 1/3: ffmpeg is still not runnable after choco and the winget fallback.
           ffmpeg verified working (attempt 2/3)."
  expected: /ffmpeg verified working \(attempt 1\/3\)\./
```

This is not hypothetical. Run [32303835840, job 96232282547](https://github.com/dudarenok-maker/Castwright/actions/runs/32303835840/job/96232282547) (`Lint, typecheck & fast checks`) is **already failing on this head** with byte-identical output, and the `npm run verify` roll-up is red behind it. The PR body's "`npm run test:hooks` is green (1413/1413)" is a local result, and local is precisely the one environment where the new block never executes — see the fourth finding.

Cost note: that scenario does not stub `Start-Sleep`, so each of the two failures also burns a real 15s retry delay.

### 🟠 Significant — a PATH-bookkeeping failure is wired into the install's own success verdict, and both of its failure modes are wrong

`cross-os.yml:108-129` / `release.yml:144-166` sit **inside** the `try` whose `catch` owns `$ffmpegOk`. That couples "did we record the directory for later steps" to "is ffmpeg installed and runnable" — two independent facts. Two concrete outcomes, depending on how the error terminates, and both are wrong:

1. **Terminating** (the `Join-Path` case above; also `Add-Content` if `$ErrorActionPreference` is `stop`): a *verified working* ffmpeg is retracted to `$ffmpegOk = $false` and reported as `winget fallback unavailable: <bookkeeping error>` — a message naming the wrong subsystem. On attempt 3 that is terminal: the loop falls through to `Write-Error "CI: ffmpeg install failed after 3 attempts (choco + winget fallback) -- ffmpeg is not runnable on this runner. ... investigate the Chocolatey feed / winget source"` and `exit 1`, sending the operator after a feed outage that never happened while ffmpeg sits installed and runnable on the box.
2. **Non-terminating** (`Add-Content` under the default `Continue` — verified: a bad `-Path` writes an error to the error stream and leaves `$ffmpegOk` untouched): the step logs an error nothing gates on, prints `ffmpeg verified working`, exits 0 — and every later step still cannot find ffmpeg. That is a silent recurrence of the exact bug #2478 describes, with the same five-minutes-later `preflight-ffmpeg.cjs` symptom this PR exists to eliminate.

The recording step must not be able to move the install verdict in either direction. Put it after the `$ffmpegOk` gate (or in its own `try`/`catch`), and make an unwritable `$GITHUB_PATH` fail loudly on its own terms rather than as a mislabelled install failure.

### 🟠 Significant — the harness now mutates the real CI job's PATH on every Windows leg

The `choco fails, winget rescues it` scenario executes the *unmodified* shipped script with the *inherited* process environment (`runStubbedFfmpegScript` → `runPowerShellFile`, no env override). On the Windows legs of `cross-os.yml` and `release.yml`, `npm run verify:quick` → `test:all` → `test:hooks` runs this file with a real `$GITHUB_PATH` **and** a real `LOCALAPPDATA` — so the stubbed run appends to the job's actual `$GITHUB_PATH` file. Verified locally:

```
$ GITHUB_PATH=<tmp>/gh_path2.txt node --test scripts/tests/cross-os-ffmpeg-install.test.mjs   # 12/12 pass
$ cat gh_path2.txt
C:\Users\dudar\AppData\Local\Microsoft\WinGet\Links
C:\Users\dudar\AppData\Local\Microsoft\WinGet\Links
```

Two lines, one per workflow file under test. `$GITHUB_PATH` entries are **prepended**, so the `Build` step of those jobs would run with `%LOCALAPPDATA%\Microsoft\WinGet\Links` ahead of everything else on PATH — a directory the test never created and whose contents nothing in this repo controls. Harmless today only because that directory happens to be absent or empty on a hosted runner; it is still a unit test reaching outside its sandbox to reconfigure the job it runs in, and passing green while doing it. Fix by handing the stubbed run an explicit env (`GITHUB_PATH` → a temp file, `LOCALAPPDATA` → a temp dir) instead of inheriting — which is also what turns this into a real test, below.

### 🟠 Significant — no paired test; the new code is dead in the only environment where the suite was run green

The PR ships a behaviour change and adds zero assertions. The existing harness is cited as covering it, but it does not: locally `$env:GITHUB_PATH` is unset, so the guard short-circuits and **deleting the entire 22-line block leaves all 12 tests green**. The suite's verdict is identical with and without the fix, so it cannot go red for the defect it is being credited with covering.

Where the block *is* reached (in CI) it is reached by accident rather than by design, with the two consequences above. The test that was owed is small and mechanical: pass `{ ...process.env, GITHUB_PATH: <tmpfile>, LOCALAPPDATA: <tmpdir> }` to the winget-rescue scenario's `spawnSync`, and assert the temp file contains exactly `<tmpdir>\Microsoft\WinGet\Links`. It would have gone red on this head for finding 1 *and* removed finding 3 — before the push.

Related: the PR body's third Test-plan item calls this path unprovable ("can only be proven by a live Cross-OS Verify run that hits it"). True of the *live winget install*; not true of the PATH-recording line this PR actually adds, which is fully observable against a stubbed `$GITHUB_PATH` in a harness that already executes this script's real control flow.

### 🟡 Minor — the WinGet Links directory is asserted rather than derived, with no `Test-Path`

`cross-os.yml:127` / `release.yml:164`. The comment claims a per-user winget install "**always** lands under this WinGet Links dir". That holds for winget **portable/zip** package types — which `Gyan.FFmpeg` is today, matching the observed `C:\Users\runneradmin\AppData\Local\Microsoft\WinGet\Links\ffmpeg.exe` — but not for `msi`/`nullsoft`/`burn` installer types, which create no Links entry at all. If that manifest's installer type ever changes, this appends a non-existent directory, the step still prints `ffmpeg verified working`, and #2478 returns silently with no signal anywhere.

The stated reason for not deriving it — "`Get-Command ffmpeg` ... resolves to a shim with no real `.Source` in some shells" — does not hold for external executables: `Get-Command <exe> -CommandType Application` returns an `ApplicationInfo` whose `.Source` is the full path (verified on pwsh 7.6.5; `-CommandType Application` is exactly what excludes the alias/function shims the comment is worried about). Since the script has *just* run `& ffmpeg -version` successfully one line earlier, `Split-Path -Parent (Get-Command ffmpeg -CommandType Application | Select-Object -First 1).Source` is both authoritative and self-validating — and it drops the `LOCALAPPDATA` dependency that causes finding 1. Failing that, at minimum gate on `Test-Path $ffmpegDir` so a wrong assumption fails loudly instead of silently.

### 🟡 Minor — release-notes pair not addressed

CLAUDE.md before-shipping step 5 allows skipping the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair for a CI-only change, but requires saying so explicitly rather than silently omitting it. The PR body does neither. (`Closes #2478` is present, plain, outside any code span — that gate is met.)

### 🟡 Minor (cleanup; pre-existing shape, now doubled) — the block is copy-pasted between two workflows

The Windows ffmpeg step was already duplicated across `cross-os.yml` and `release.yml`; this adds ~22 more near-identical lines to each, diverging only in one comment clause ("same step shape as this file's"). Successive fixes to this step have now had to be applied twice each, and the harness exists in its current dual-file loop shape only because of the duplication. This repo already uses composite actions (`./.github/actions/setup`); moving the whole step there would collapse ~90 duplicated lines into one `uses:` per file and make the next fix a single edit. Not blocking, and not a request to restyle working code — noted because this PR is the second consecutive change paying the duplication twice.

### ✅ What is solid

- The **diagnosis is right and precisely stated**: an in-process `$env:Path` re-read cannot reach a later step, and `$GITHUB_PATH` is the correct documented mechanism for the job-wide fix. Both the commit message and the code comment name the observed run.
- The `%LOCALAPPDATA%\Microsoft\WinGet\Links` path **matches the real failure log** for a `runneradmin` per-user portable install — the value is right; only its derivation and its guard are wrong.
- Applying the identical fix to `release.yml` in the same PR is correct: it carries the same step and the same latent bug, and leaving it would be the "one instance fixed, the class left armed" shape.
- `Add-Content` is a valid `$GITHUB_PATH` writer under `pwsh` (utf8NoBOM default, and the value is pure ASCII regardless), and Windows runners default to `pwsh` — there is no encoding problem here.
- The block is correctly scoped to the winget branch only, and cannot double-append: a successful attempt `exit 0`s immediately.

### Verdict

**Request changes.** One blocking correctness bug that is already failing this PR's own required check on this exact head, plus three significant findings — an error-containment defect that can fail a job with a misleading message *or* silently re-open #2478, a test that reaches outside its sandbox to mutate the CI job's PATH, and the absence of any assertion covering the new behaviour. Findings 1, 3 and 4 share one root: the block was written and verified in the single environment (`GITHUB_PATH` unset) where it does not run. One test that pins `GITHUB_PATH`/`LOCALAPPDATA` explicitly closes all three.

dudarenok-maker:
## PR review — pass 2 (head 70ca32834d298f7ac288abb80be0273aebe897eb, depth medium)

Deltas only, per the re-review format — pass 1's solid items are not re-listed.

Scope: the two workflow diffs + `scripts/tests/cross-os-ffmpeg-install.test.mjs`. Verified: `node --test scripts/tests/cross-os-ffmpeg-install.test.mjs` 12/12 green locally; re-ran pass 1's repro environments against **both** the pre-fix (`d2fd03e6`) and current (`70ca3283`) script bodies through a standalone replica of the harness (same extraction + same stub preambles, run outside the repo so nothing was modified); mutation-tested the new assertions; probed the `$GITHUB_PATH`-write-failure path under both `$ErrorActionPreference` settings; and read run 32305248183 (all 15 checks green on this head).

---

**🔴 1 — `Join-Path $null` throwing into the winget `catch` and retracting `$ffmpegOk`: VERIFIED RESOLVED.**

Ran the shipped script directly with `GITHUB_PATH` set and `LOCALAPPDATA` unset (the ubuntu-latest shape), `STUB_CHOCO_FAILS_WINGET_RESCUES` unchanged:

```
PRE-FIX (d2fd03e6):  Attempt 1/3: choco reported success but ffmpeg is not runnable ...
                     winget fallback unavailable: Cannot bind argument to parameter 'Path' because it is null.
                     Attempt 1/3: ffmpeg is still not runnable after choco and the winget fallback.
                     ffmpeg verified working (attempt 2/3).

HEAD (70ca3283):     Attempt 1/3: choco reported success but ffmpeg is not runnable ...
                     ffmpeg verified working (attempt 1/3).
```

The block is genuinely outside the `try` that owns `$ffmpegOk` (`cross-os.yml:135` sits after that `catch` closes at :111, still inside the `if (Get-Command winget)` arm), and its own `catch` writes nothing but a message. `Lint, typecheck & fast checks` is green on this head and its log shows all six pwsh scenarios executing on ubuntu-latest, so this is confirmed in the environment that produced the original failure, not just locally.

On the dispatch's question — *is there a real windows-latest path where `LOCALAPPDATA` is unset that makes the new guard wrong?* No. #2478's own log resolves ffmpeg at `C:\Users\runneradmin\AppData\Local\Microsoft\WinGet\Links\ffmpeg.exe`, so the variable is set on hosted runners; a self-hosted runner service without a loaded profile still gets a (different) `LOCALAPPDATA`. The guard is not wrong. It is silent — see Minor.

**🟠 3 — the harness mutating the real CI job's `$GITHUB_PATH`: VERIFIED RESOLVED.**

Re-ran pass 1's exact repro — suite with a live `GITHUB_PATH=<file>` in the parent environment:

```
ℹ tests 12 / pass 12 / fail 0
=== host GITHUB_PATH file after the run ===
(byte count:) 0
```

Zero bytes, against two appended lines in pass 1. The `envOverrides` plumbing (`cross-os-ffmpeg-install.test.mjs:174,178`) is correct and does not break the other caller: `parsePowerShellSyntaxErrors` (line 209) now receives `env: { ...process.env }`, which is behaviourally identical to the previous inherit. Return-shape change at line 273 (`{ ...proc, ... }`) preserves every `spawnSync` own property the callers read, and `readFileSync` runs before the `finally`'s `rmSync`, so the read is not racing the cleanup.

**🟠 2 — bookkeeping coupled to the install verdict: HALF resolved.** The retraction direction is closed (above). The silent-success direction is not, and the new `catch` cannot fire under the preference the harness runs it with. Detail below.

**🟠 4 — no paired test: HALF resolved.** There is now real coverage for the append (deleting the production block goes red). It is not a regression test for the defect this commit fixed — it is green on the pre-fix code. Detail below.

**🟡 5 (WinGet Links asserted, no `Test-Path`): partially resolved** — the misleading `Get-Command`/`.Source` claim is gone, but no `Test-Path` was added and the replacement prose introduces two new inaccuracies. **🟡 6 (release notes): still open.** **🟡 7 (duplication): deferred to #2480** — see Minor for why that disposition does not meet findings-triage's carve-out as filed.

---

### 🟠 Still open (correctness) — a failed `$GITHUB_PATH` write still reports success, and under the harness's own `$ErrorActionPreference` the new `catch` never runs

`cross-os.yml:135-141` / `release.yml:172-178`.

Pass 1 asked for the recording step to "fail loudly on its own terms". The fix chose warn-only. Measured, running the shipped script with `GITHUB_PATH` pointing into a directory that does not exist (an unwritable `$GITHUB_PATH` — the realistic failure), everything else unchanged:

```
$ErrorActionPreference = default          <- what the harness runs it with
  exit status: 0
  stdout : ffmpeg verified working (attempt 1/3).
  stderr : Add-Content: ... Could not find a part of the path '...\no-such-dir\github_path.txt'.
  the new catch's warning printed? FALSE

$ErrorActionPreference = stop
  exit status: 0
  stdout : warning: failed to persist ffmpeg's install directory to $GITHUB_PATH for later steps: ...
           ffmpeg verified working (attempt 1/3).
  the new catch's warning printed? true
```

Two separate problems in that output:

1. **Under the default preference the `catch` is dead code.** `Add-Content`'s failure is *non-terminating*, so `try`/`catch` never sees it: the message the fix exists to emit does not print, the raw cmdlet error goes to stderr, and the step still says `ffmpeg verified working` and exits 0. This is the one environment the repo's own regression harness executes the script in — so no test in this PR, present or future, can exercise the error path as written. `-ErrorAction Stop` on the `Add-Content` makes the `catch` real under both preferences and costs one word.
2. **Under either preference the step reports success while nothing was persisted.** That is pass 1's finding-2 mode (2) verbatim, relabelled from "silently retracts" to "silently warns": the job goes green here and dies five minutes later in `pretest` with `ffmpeg not on this shell's PATH` — the exact symptom, in the exact wrong place, that this PR exists to eliminate. A plain `Write-Host "warning: …"` line is not loud: it produces no annotation, no step failure, and nothing downstream reads it. `Write-Host "::warning::…"` at minimum promotes it to a run-summary annotation without anyone having to decide whether a bookkeeping failure should fail the job.

(Whether it *should* fail the job is a judgment call and I am not asking for that call here — item 1 and the `::warning::` are both mechanical and independent of it.)

### 🟠 Still open (correctness of the test) — the new assertions pin the feature, not the fix: they are green on the pre-fix code, and the `$ffmpegOk` half of the guard is unmutated

`cross-os-ffmpeg-install.test.mjs:444-453`.

CLAUDE.md's bar is "Bug fix → ship a regression test that fails before the fix and passes after". The bug fixed by `70ca3283` is the crash + the coupling. Ran the **pre-fix** (`d2fd03e6`) script body through the **post-fix** harness — i.e. exactly what the new test does, against the code the new test is supposed to catch:

```
VARIANT: PRE-FIX-code_with_POST-FIX-harness
exit status: 0
stdout: ... ffmpeg verified working (attempt 1/3).
GITHUB_PATH file contents: "...\LocalAppData\Microsoft\WinGet\Links\r\n"
ASSERT regex /Microsoft[\\/]WinGet[\\/]Links/ -> true
ASSERT includes(node join)                     -> true
```

All three new assertions pass on the code this commit exists to repair. The reason is structural and worth naming: the harness fix (correctly) *sets* `LOCALAPPDATA`, and `LOCALAPPDATA` being **absent** was the only condition under which the pass-1 defect was observable. Closing finding 3 closed the environment that made finding 1 visible, and nothing was added to restore it. Reverting `-and $env:LOCALAPPDATA`, or moving the whole block back inside the winget `try`/`catch`, is green.

Second gap in the same place: **the guard's blocking behaviour is untested.** Mutating `if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA)` → `if ($env:GITHUB_PATH -and $env:LOCALAPPDATA)` and running the `both choco and winget fail all 3 attempts` scenario (where `Get-Command winget` resolves to the stub, so the block *is* reached on every attempt):

```
MUTANT dropOkGuard, bothFail:
  exit 1, CI message present, 2 Start-Sleep markers   <- every existing assertion still passes
  GITHUB_PATH file contents: "<dir>\Microsoft\WinGet\Links\r\n" x3
```

A never-installed ffmpeg's directory prepended onto the job's PATH three times, and the suite is green. Both gaps close with cheap additions to tests that already exist:

- `assert.equal(proc.githubPathContents, '')` in the both-fail scenario (and in `choco works on the first try`) — pins that the guard *blocks*, not only that it *permits*;
- one scenario that runs with `LOCALAPPDATA` deliberately absent and asserts the step still prints `ffmpeg verified working (attempt 1/3).` — that one goes red on `d2fd03e6` and is the regression test this commit actually owes;
- optionally one that points `GITHUB_PATH` at an unwritable location and asserts the step still exits 0 *and* emits the warning — which would have caught the inert-`catch` problem above.

---

### 🟡 Minor

- **Finding 5, partially resolved and now carrying two new false claims.** `cross-os.yml:122-123` / `release.yml:158-160`: "a per-user winget install (**no admin elevation, the case on GitHub-hosted runners**) **always** lands under this WinGet Links dir." GitHub-hosted Windows runner steps *do* run elevated — that is why the `choco install ffmpeg -y --no-progress` at `cross-os.yml:85` works at all — so "no admin elevation" is not the reason the path is user-scoped; winget's default install scope for a portable package is `user` regardless of elevation. And "always" remains the overstatement pass 1 flagged: it holds for `portable`/`zip` package types, not `msi`/`nullsoft`/`burn`. No `Test-Path $ffmpegDir` was added, so a manifest that changes installer type appends a non-existent directory, the step still prints `ffmpeg verified working`, and #2478 returns with no signal anywhere. If `Test-Path` is added, the harness must also `mkdirSync` the stub `LocalAppData\Microsoft\WinGet\Links` — it currently only computes that path as a string, never creates it.
- **NEW — the comment justifying the `LOCALAPPDATA` guard was made false by this same commit.** `cross-os.yml:131-134` / `release.yml:168-171` still say the guard exists because the harness "executes this same script directly via pwsh (including on ubuntu-latest, where LOCALAPPDATA is never set)". As of this commit the harness *always* injects `LOCALAPPDATA` (line 264), so the script under test never sees it unset. The guard is fine as defence-in-depth; the stated reason is now wrong, in both files.
- **NEW — the guard has no else-branch signal.** If either variable is ever absent on a real runner, the block no-ops and prints nothing at all; the job goes green with ffmpeg missing downstream. Same failure surface as the bullet above, and one `else { Write-Host "::warning::…" }` closes both.
- **NEW — dead duplicate assertion.** `cross-os-ffmpeg-install.test.mjs:444` re-asserts `proc.status === 0`, which line 423 already asserted in the same test. It can never fail independently.
- **Finding 6, still open.** The PR body still says nothing about the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair. CLAUDE.md step 5 permits skipping it for a CI-only change but requires saying so explicitly; silence is the thing it rules out, not the skip.
- **NEW — the PR body was not updated for this round.** Test-plan item 1 still describes the pass-1 story ("existing regression harness … 1413/1413") with no mention of the new assertions or the env isolation; item 3 still asserts this path "can only be proven by a live Cross-OS Verify run that hits it", which *this commit* falsified by proving it against a stubbed `$GITHUB_PATH`; and the harness-isolation change is an incidental fix that CLAUDE.md requires be declared ("Also fixed, found in passing: …") so it does not read as scope creep.
- **Finding 7 / #2480 — the deferral does not meet the carve-out as the issue is written.** findings-triage allows exactly one finding to leave the round unfixed: one needing a design pass, *and its issue must name the decision owed*. #2480 names no decision — it prescribes the composite-action shape outright and calls it "a pure DRY extraction … no behavior change", which is the definition of *not* needing a design pass. By its own text it is a chore, and "it would expand the scope of this PR" / "it's cheaper to batch these later" are both listed as void deferral reasons. Either it is fixed here, or #2480 is rewritten to name the actual fork (composite action vs. a shared `.ps1` under `scripts/` invoked by both workflows — which changes what the harness extracts, and is a real decision). Flagging the process, not re-litigating pass 1's nit: the code finding itself was 🟡 and stays 🟡.

---

### Verdict

**Request changes** — but the blocking defect is gone. Pass 1's 🔴 and one of its three 🟠 are verified resolved with measurements, and this head is green on all 15 checks.

What remains is that both halves of the fix are one level shallower than claimed. The error containment is real in one direction and inert in the other — the `catch` cannot fire under the preference the harness uses, and a failed persist still exits 0 announcing success. And the test is a feature test wearing a regression test's comment: it is green on `d2fd03e6`, and the `$ffmpegOk` half of the guard survives deletion. Both close inside the existing scenarios — `-ErrorAction Stop`, a `::warning::`, one empty-contents assertion, and one `LOCALAPPDATA`-absent scenario. No new blocking defect was introduced this round.

dudarenok-maker:
## PR review — pass 3 (head 8097ab7c34400998187f4141fc6158c70d4f241b, depth medium)

Deltas only, per the re-review format — pass 1's and pass 2's solid items are not re-listed.

Scope: the two workflow diffs + `scripts/tests/cross-os-ffmpeg-install.test.mjs`. Verified: suite 14/14 green in the repo; `npm run test:hooks` 1415/1415 green; all 15 PR checks green on this head. Then, on a **copy of the tree outside the repo** (nothing tracked was modified), ran four mutations of the shipped block against the current test file, probed `Add-Content -ErrorAction Stop` directly under pwsh 7.6.5 for both realistic failure shapes, ran the current test file against the true `d2fd03e6` body, and re-ran pass 1's host-`$GITHUB_PATH`-pollution repro.

---

### Round-3 claims — verification

| # | Claim | Verdict |
|---|---|---|
| 1 | `-ErrorAction Stop` makes the failure catchable; `::warning::` annotation | **VERIFIED CORRECT** — but untested, see 🟡 below |
| 2a | `withLocalAppData: false` deletes LOCALAPPDATA from the child env | **VERIFIED CORRECT** |
| 2b | The new scenario is mutation-differentiating **for the LOCALAPPDATA guard** | **FALSE** — see 🟡 below |
| 2c | Negative assertions on choco-succeeds / both-fail | **VERIFIED REAL** (each uniquely catches a distinct mutation) |
| 3 | Dead duplicate `assert.equal(proc.status, 0)` removed | **VERIFIED** |
| 4 | Comment rewritten, "no admin elevation" / "always lands" dropped | **VERIFIED** (0 occurrences in both files) |
| 5 | #2480 re-edited to name the decision owed | **VERIFIED** (residual nit below) |
| 6 | PR body updated; release notes marked N/A | **VERIFIED for release notes**; one false figure, below |

**Claim 1, measured.** Under the default `$ErrorActionPreference = 'Continue'`, for both realistic failure shapes:

```
A) WITHOUT -ErrorAction Stop, path into a nonexistent dir  ->  (no throw)   catch never fires
B) WITH    -ErrorAction Stop, path into a nonexistent dir  ->  CATCH FIRED: Could not find a part of the path '...'
C) WITH    -ErrorAction Stop, -Path is a DIRECTORY         ->  CATCH FIRED: Unable to write content because it is a directory: '...'
D) WITHOUT -ErrorAction Stop, -Path is a DIRECTORY         ->  (no throw)   catch never fires
```

End-to-end through the shipped script with an unwritable `$GITHUB_PATH`, the annotation renders correctly, the backtick escape produces a literal `$GITHUB_PATH`, `$_` interpolates, it stays a single line, and stderr is now clean (pass 2 saw the raw cmdlet error leak there):

```
::warning::failed to persist ffmpeg's install directory to $GITHUB_PATH for later steps -- later steps in this job may not find ffmpeg on PATH: Could not find a part of the path '...\no-such-dir-zzz\gh.txt'.
ffmpeg verified working (attempt 1/3).
```

Pass 2's finding-1 mechanism is genuinely closed.

---

### Pass-1 and pass-2 findings — disposition

- **🔴 P1-1 (`Join-Path $null` crash retracting `$ffmpegOk`): VERIFIED RESOLVED** (confirmed pass 2; re-confirmed here — the block sits in its own `try`/`catch` outside the winget one, and all 15 checks are green on this head).
- **🟠 P1-3 (harness mutating the real job's `$GITHUB_PATH`): VERIFIED RESOLVED, re-measured on this head.** Ran the suite with a live `GITHUB_PATH=<file>` in the parent env: 14/14 pass, host file **0 bytes** afterwards.
- **🟠 P2-1 (`Add-Content` failure not catchable; `catch` inert): VERIFIED RESOLVED** — measured above.
- **🟠 P2-2 (new assertions were placebos): RESOLVED, but narrower than claimed.** The suite now genuinely goes **red** against the true pre-fix body — current test file vs. `d2fd03e6` workflows gives `2 fail / 12 pass`, both failures being the new LOCALAPPDATA-unset scenario. CLAUDE.md's "red before the fix, green after" bar is met. What is *not* true is the attribution — see 🟡 below.
- **🟡 P1-5 / P2 (WinGet Links dir asserted, no `Test-Path`): PARTIALLY RESOLVED.** The two false claims are gone and the replacement prose is factually grounded in the observed CI path. `Test-Path` was still not added (0 occurrences) — the code still appends an unvalidated directory.
- **🟡 P2 (comment justifying the LOCALAPPDATA guard made false by the harness fix): RESOLVED** — the comment now says "including with LOCALAPPDATA deliberately absent", which the new scenario makes true.
- **🟡 P2 (dead duplicate assertion): VERIFIED RESOLVED** — one `assert.equal(proc.status, 0)` remains in that test.
- **🟡 P1-6 / P2 (release notes): RESOLVED** — the PR body now states the N/A explicitly with the CLAUDE.md step-5 carve-out named.
- **🟡 P2 (PR body not updated): RESOLVED**, with one false figure — below.
- **🟡 P1-7 / P2 (#2480 named no decision): RESOLVED.** #2480 now carries a `## Decision owed` section naming two shapes with real, asymmetric tradeoffs (composite action needs an explicit `shell: pwsh` and a rework of the test's extraction logic; a shared `.ps1` keeps the extraction simple but is a new CI shape for this repo). That meets findings-triage's carve-out.
- **🟡 P2 (guard has no else-branch signal): STILL OPEN** — no `else` was added. This one turns out to matter more than pass 2 credited it; see below.

---

### 🟡 Minor — the new scenario does not pin the LOCALAPPDATA guard, and three places say it does

`.github/workflows/cross-os.yml:141` / `.github/workflows/release.yml:178`; test comment at `scripts/tests/cross-os-ffmpeg-install.test.mjs:459-470`; commit message point 2.

Reverting **only** the guard — `if ($ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA)` back to `if ($ffmpegOk -and $env:GITHUB_PATH)`, i.e. exactly the pre-fix guard the comments name — leaves **14/14 green**, the new scenario included. Probing that mutant's actual output shows why:

```
exit status : 0
stdout      : Attempt 1/3: choco reported success but ffmpeg is not runnable ... Trying the winget fallback.
              ::warning::failed to persist ffmpeg's install directory to $GITHUB_PATH ...: Cannot bind argument to parameter 'Path' because it is null.
              ffmpeg verified working (attempt 1/3).
GITHUB_PATH : ""
```

The `Join-Path $null` crash **still happens** — it is simply swallowed into the `::warning::` by the block's own `try`/`catch`, which is round 2's fix. All three assertions still hold: exit 0 ✔, `attempt 1/3` ✔, `githubPathContents === ''` ✔ (nothing was written because `Join-Path` threw *before* `Add-Content`). The scenario cannot distinguish "the guard correctly skipped the block" from "the guard let it through and it crashed into the warning handler".

So the test's redness against `d2fd03e6` comes from the **containment** change (block moved out of the winget `try`/`catch`), not from the LOCALAPPDATA conjunct. The manual verification cited in the commit message was honest — it just varied both changes at once and attributed the result to the wrong one. Three artifacts now state that wrong attribution:

- `cross-os.yml:139-140` / `release.yml:176-177`: "…LOCALAPPDATA deliberately absent, **to pin the guard against exactly the crash**…"
- the test comment: "**This test fails on that code and passes on the current guard** (`... -and $env:LOCALAPPDATA`), which is the property worth pinning"
- commit message point 2.

**The consequence is the reverse of what the comments assume, and it merges with the still-open else-branch minor.** Because the block already contains `Join-Path $null` in its own `catch`, the conjunct changes no job outcome — it only decides whether a *warning is printed*. On a real runner with `LOCALAPPDATA` genuinely absent:

- **with** the conjunct: block silently skipped, nothing persisted, job green — #2478 recurs downstream with **no signal anywhere**;
- **without** it: `Join-Path` throws, the `catch` fires, and you get the `::warning::`.

The conjunct is what makes that failure silent. So the fix worth making is not "add a test for the conjunct" but pass 2's still-open `else { Write-Host "::warning::…" }` — it converts the silent skip into the same annotation, makes the guard's blocking behaviour observable, and is directly assertable in the existing LOCALAPPDATA-unset scenario (which today asserts only that nothing was written). One change closes the untested seam, the false attribution, and the else-branch minor together; the three comments should then say what the test actually pins.

### 🟡 Minor — the `-ErrorAction Stop` fix itself has no test

`.github/workflows/cross-os.yml:144` / `.github/workflows/release.yml:181`.

Deleting ` -ErrorAction Stop` leaves **14/14 green**. Pass 2 named the exact test that would catch it ("one that points `GITHUB_PATH` at an unwritable location and asserts the step still exits 0 *and* emits the warning"); it was not added. This is the highest-value gap of the three, because unlike the conjunct, reverting this one word restores a genuine correctness bug — a failed persist reported as success with no annotation at all. The harness already supports it: `runStubbedFfmpegScript` needs only a `githubPathOverride` option pointing into a nonexistent directory, and the assertion is `assert.match(proc.stdout, /::warning::failed to persist/)`. I ran that exact scenario and it produces that exact line.

### 🟡 Minor — the PR body's test count is wrong

Test-plan item 1 says "**15 tests** (12 original + 3 added across this PR's rounds)". Measured: `main` has 6 tests per workflow (12 total), this head has 7 per workflow (**14 total**) — so **2** were added, not 3, and the total is 14, not 15. Both other figures in the body check out: `test:hooks` really is 1415/1415, and the "manually reproduced the exact CI failure … against the pre-fix script body" claim is accurate (I reproduced it: 2 fail / 12 pass against `d2fd03e6`).

### 🟡 Minor — #2480's title and Acceptance still prescribe the option its body says is open

The body now names the fork correctly, which is what findings-triage requires. But the title still reads "extract … **into a shared composite action**" (option 1 of the two), and Acceptance still carries "No behavior change — this is a pure DRY extraction", the phrase pass 2 quoted as evidence the item was being filed as a chore rather than a design pass. A neutral title (e.g. "de-duplicate the Windows ffmpeg install step across cross-os.yml and release.yml") would stop the ticket from pre-answering its own decision.

### ✅ Newly verified solid this round

- The two shipped blocks are **semantically identical** across `cross-os.yml` and `release.yml` — the only textual difference is comment re-flow plus release.yml's deliberate "same step shape as this file's" clause. No one-file-fixed-class-armed divergence.
- `withLocalAppData: false` genuinely reaches the child: the mutant probe's `Cannot bind argument to parameter 'Path' because it is null` proves `$env:LOCALAPPDATA` is null in the spawned pwsh — the `delete` is not defeated by Windows env-key casing, and pwsh does not re-synthesise it. Verified on Windows, not inferred.
- Both negative assertions are non-vacuous and **non-redundant**: dropping `$ffmpegOk` from the guard is caught *only* by the both-fail scenario (2 fail), and moving the block out of the winget-only branch is caught *only* by the choco-succeeds scenario (2 fail). Each earns its place.
- No double-append is possible: the block runs only when `$ffmpegOk` is true, and the very next statement `exit 0`s.

---

### Verdict

**Approve, with cleanup.** No correctness bug remains in the shipped code. Every claim this round makes about *behaviour* verified true under direct measurement: `-ErrorAction Stop` really does make the failure catchable, the `::warning::` renders and escapes correctly, the guard works, the harness no longer touches the host's `$GITHUB_PATH`, the two files stay in lockstep, and the suite is a real regression test that goes red on the pre-fix body.

What remains is entirely 🟡 and none of it re-triggers the loop: two untested seams (`-ErrorAction Stop`, and the LOCALAPPDATA conjunct), three comments that attribute the test's redness to the wrong one of the two changes it actually varies, a wrong test count in the PR body, and #2480's title still prescribing. The single highest-value follow-up is pass 2's still-open `else` branch — it is the one change that converts a silent recurrence of #2478 into a visible annotation, and it makes the guard testable at the same time.

Pass 3 of 3; the loop cap is reached, and since nothing outstanding is 🔴/🟠 it does not need to be re-opened.

dudarenok-maker:
## Post-pass-3 fixup (head 8e0a3cca)

Pass 3 verdict was **ready to merge** — no correctness bug remained, only 🟡 minor items. Fixed those per findings-triage (cleanup fixed this round, no re-trigger):

- Split the `LOCALAPPDATA` check out of the single `$ffmpegOk -and $env:GITHUB_PATH -and $env:LOCALAPPDATA` conjunct into a nested `if/else`, so a real box missing `LOCALAPPDATA` now warns (`::warning::LOCALAPPDATA is not set...`) instead of skipping the PATH-persist silently — closing the exact gap pass 3 found (the "LOCALAPPDATA absent" test's old assertions passed for the wrong reason: round 2's crash-containment fix alone, not the conjunct). Added an assertion on the new warning line and verified by hand it fails against the reverted single-conjunct guard.
- Fixed the doc-comment above that test, which mis-attributed which assertion pins which fix.
- Re-titled #2480 to stop prescribing "composite action" — its body already names the actual decision owed.
- Corrected the PR body's test count (14, not 15).

`npm run test:hooks` green (1415/1415). Proceeding to merge per the review-gate loop cap (3 rounds used, remaining findings are cosmetic-only).

PR #2484, opened 2026-08-20
ci(ops): ignore stale guardrail-tmp-* leftovers in lint

### LINKED ISSUE #2482: ops-61 — a stale guardrail-tmp dir wedges pushes while git reports clean
## What happened

A leftover `guardrail-tmp-*/` directory **silently wedges every push from a
checkout**, and `git status` reports the tree clean while it does.

Observed 2026-08-20 pushing `fix/scripts-2465-quarantine-register-parse`:

```
guardrail-tmp-Z47haC/planted.test.ts
  2:11  error  No it.skipIf(process.env.CI). Use quarantinedIt ...
[fail] lint (exit 1)   husky - pre-push script failed
```

The file was never written by the developer - it is the planted violation from
`scripts/tests/eslint-guardrail.test.mjs`, left behind by an earlier run.

## Why it is invisible

`.gitignore:76` carries `guardrail-tmp-*/`, so **git** ignores it. ESLint does
not, so **lint** fails on it. The result is a checkout that looks pristine and
refuses to push, complaining about a file that does not appear in `git status`,
`git diff`, or any listing a developer would think to run.

## Why the obvious fix is wrong

Adding `guardrail-tmp-*/` to `eslint.config.mjs`'s global `ignores` defeats the
probe. That test's own header explains the constraint:

> CRITICAL: the planted file is written INSIDE the repo tree (guardrail-tmp-*
> dir at the repo root), NOT os.tmpdir(). ESLint flat config ignores files
> outside its base path and would exit 0 (false pass) if the file were in a
> system temp dir.

The probe needs ESLint to SEE the file. A naive ignore turns a real guardrail
into one that passes unconditionally - the exact class of defect the guardrail
exists to catch.

## The decision this needs

The cleanup is already in a `finally`, so the directory surviving means either
the process died before `finally` ran, or `rmSync` failed. Which of those it was
is **not established**, and the two candidate fixes differ in what they cover:

**Option A - make the leftover less likely.**
`rmSync(dir, { recursive: true, force: true, maxRetries: 10, retryDelay: 50 })`,
plus a sweep of stale `guardrail-tmp-*` at test start. On Windows `rmSync`
without `maxRetries` throws EBUSY/EPERM when a spawned child still holds a
handle, which the `execFileSync('npx', ...)` immediately above makes likely.
*Covers the race. Does NOT cover a killed process, and does not help at all once
a stale directory already exists, because lint runs BEFORE `test:hooks` in
`verify:fast:branch` - so the sweep never gets to run.*

**Option B - make a leftover harmless.**
Add `guardrail-tmp-*/` to ESLint's global `ignores`, and have the probe pass
`--no-ignore` so it still lints its own planted file. *Covers every cause
including a hard kill, permanently. Costs a semantic change to lint config, and
the probe now depends on `--no-ignore` continuing to override a global ignore.*

B dominates on coverage; A is the smaller diff. **B with a paired assertion that
the probe still fails on the planted violation** is the recommendation - without
that assertion, B is exactly the false-pass the header warns about.

## Acceptance

- A stale `guardrail-tmp-*/` directory present in the tree does NOT fail
  `npm run lint` or block a push.
- The guardrail test still FAILS when the rule is removed from
  `eslint.config.mjs` - i.e. mutate the rule away and watch the probe redden.
  Pinning only "the probe passes" would pass against a probe that lints nothing.
- Whichever option is taken, a regression test covers the leftover-directory
  case specifically, not just the happy path.

## Key files

- `scripts/tests/eslint-guardrail.test.mjs` - creates and cleans the directory
- `eslint.config.mjs` - the global `ignores` block (~line 32)
- `.gitignore:76` - `guardrail-tmp-*/`

dudarenok-maker:
## Summary
- A `guardrail-tmp-*/` dir left behind by `scripts/tests/eslint-guardrail.test.mjs` (killed process, or `rmSync` losing a Windows EBUSY race against the spawned child right above it) silently wedged every push from that checkout: git ignores the dir but ESLint didn't, so lint failed on a planted violation nobody wrote and no listing showed.
- Adds `guardrail-tmp-*/` to `eslint.config.mjs`'s global `ignores` (Option B from #2482), so a leftover is harmless rather than merely less likely. The probe passes `--no-ignore` to its own `npx eslint` call so it still sees its planted file.
- Verified this isn't a false pass across three independent review rounds: disabling the guardrail rule reddens the probe test every time.
- **Also fixed, found in passing (review round 1):** `check-no-budget-poll.test.mjs`'s own planted-violation temp dir (`budget-poll-tmp-*`) has the identical interrupted-run leftover shape as `guardrail-tmp-*` — added to `.gitignore` and `eslint.config.mjs`'s global ignores alongside it.
- **Also fixed, found in passing (review round 2):** the two original probes were both blind to the global ignore becoming over-broad (e.g. accidentally matching `**/*.test.ts` instead of a directory-prefixed pattern) — added a third probe.
- **Also fixed, found in passing (review round 3):** that third probe's first version planted its own file under a NEW, unignored repo-root `mkdtempSync` prefix (`eslint-guardrail-canary-*`) — reintroducing the exact #2482 leftover-wedge class this PR exists to close, if that plant ever survived a killed process. Rewritten to probe an already-tracked file (`server/src/test-utils/quarantine.test.ts`) instead — writes nothing, so there is nothing to leak.
- **Release notes intentionally skipped** — pure CI-only internal chore (lint config + its own regression tests), no user- or operator-visible effect (before-shipping checklist step 5).

## Test plan
- [x] Targeted regression test plants a stale `guardrail-tmp-*` dir and lints it via an explicit glob — confirms ESLint reports it as ignored (exit 2), never linted.
- [x] Third probe checks an already-tracked `*.test.ts` file is never reported as ignored — catches the ignore pattern ever growing too broad, without planting anything.
- [x] Reproduced the original issue's exact repro shape locally (planted violation dir + `npm run lint`) — now exits 0. Same repro for the `budget-poll-tmp-*` sibling.
- [x] Mutation checks across all three probes: disabling `no-restricted-syntax` reddens probe 1; a typo'd `--no-ignoer` flag correctly fails instead of false-passing; appending an over-broad `**/*.test.ts` ignore reddens probe 3 while probes 1–2 stay green; confirmed `git status --porcelain` is clean after every run (no leftover).
- [x] `npm run test:hooks` — 1417/1417 pass.
- [x] `npm run verify:fast:branch` — green.

Closes #2482

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,12 @@ coverage/
 # .gitignore ensures an interrupted run never stages them).
 guardrail-tmp-*/
 
+# check-no-budget-poll.test.mjs: same planted-violation-inside-the-repo pattern as
+# guardrail-tmp-* above (#2482) — a leftover from an interrupted run must not
+# surface as an untracked/staged file, and must not wedge lint (see the matching
+# eslint.config.mjs ignore).
+budget-poll-tmp-*/
+
 # Verify-cache (per-step input hashes for `npm run verify`; never committed).
 .verify-cache.json
 .verify-cache.json.tmp
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -52,9 +52,9 @@ export default tseslint.config(
       // Throwaway debugging probes — git-ignored and not held to the lint bar.
       'server/repro-*.mts',
       // Leftover from scripts/tests/eslint-guardrail.test.mjs's planted-violation
-      // probe (mkdtempSync inside the repo root, see .gitignore:76): a directory
+      // probe (mkdtempSync inside the repo root, see .gitignore): a directory
       // that survives cleanup (killed process, or rmSync losing a Windows EBUSY
-      // race against the execFileSync child right above it) must not wedge every
+      // race against the spawnSync child right above it) must not wedge every
       // subsequent `npm run lint`/pre-push on an unrelated checkout. The probe
       // itself still needs ESLint to SEE its planted file, so it passes
       // `--no-ignore` to override this — see the probe's own comment.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -51,6 +51,14 @@ export default tseslint.config(
       // Ad-hoc, local-only repro/bisect scripts (e.g. server/repro-attribution.mts).
       // Throwaway debugging probes — git-ignored and not held to the lint bar.
       'server/repro-*.mts',
+      // Leftover from scripts/tests/eslint-guardrail.test.mjs's planted-violation
+      // probe (mkdtempSync inside the repo root, see .gitignore:76): a directory
+      // that survives cleanup (killed process, or rmSync losing a Windows EBUSY
+      // race against the execFileSync child right above it) must not wedge every
+      // subsequent `npm run lint`/pre-push on an unrelated checkout. The probe
+      // itself still needs ESLint to SEE its planted file, so it passes
+      // `--no-ignore` to override this — see the probe's own comment.
+      'guardrail-tmp-*/',
       // Accepted-result scratch written by the token-saver skill (see the
       // matching .gitignore entry). Unreviewed model output that can land on a
       // linted extension, and the skill writes it under whatever root it is
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -59,6 +59,9 @@ export default tseslint.config(
       // itself still needs ESLint to SEE its planted file, so it passes
       // `--no-ignore` to override this — see the probe's own comment.
       'guardrail-tmp-*/',
+      // check-no-budget-poll.test.mjs's own planted-violation temp dirs — same
+      // interrupted-run leftover shape as guardrail-tmp-* above (#2482).
+      'budget-poll-tmp-*/',
       // Accepted-result scratch written by the token-saver skill (see the
       // matching .gitignore entry). Unreviewed model output that can land on a
       // linted extension, and the skill writes it under whatever root it is
--- a/scripts/tests/eslint-guardrail.test.mjs
+++ b/scripts/tests/eslint-guardrail.test.mjs
@@ -121,31 +121,26 @@ test('a stale guardrail-tmp-* leftover is ignored by ESLint, not merely absent o
   );
 });
 
-test('the guardrail-tmp-*/budget-poll-tmp-* ignores do not swallow an ordinary *.test.ts file', () => {
+test('the guardrail-tmp-*/budget-poll-tmp-* ignores do not swallow an ordinary tracked *.test.ts file', () => {
   // The two tests above prove the leftover dirs ARE ignored and the rule
   // fires when unignored via --no-ignore. Neither would catch the ignore
   // pattern becoming too broad (e.g. accidentally matching '**/*.test.ts'
   // instead of a specific directory prefix) — test 1 always passes
   // --no-ignore, which would blind it to that too, and test 2 asserts the
-  // file IS ignored, so a broader pattern only makes it greener. This test
-  // is the counterweight: an ordinary planted violation OUTSIDE both
-  // leftover-dir prefixes must still be caught with NO --no-ignore flag.
-  const dir = mkdtempSync(join(repoRoot, 'eslint-guardrail-canary-'));
-  const f = join(dir, 'planted.test.ts');
-  writeFileSync(
-    f,
-    "import { it } from 'vitest';\nit.skipIf(process.env.CI)('x', () => {});\n",
-  );
-  let result;
-  try {
-    result = runEslint([f]);
-  } finally {
-    rmSync(dir, { recursive: true, force: true });
-  }
-  assert.equal(result.exitCode, 1, `eslint should still lint an ordinary test file and reject the violation (got ${result.exitCode}): ${result.output}`);
-  assert.match(
+  // file IS ignored, so a broader pattern only makes it greener.
+  //
+  // This is the counterweight: it probes an already-TRACKED test file
+  // instead of planting one. An earlier version of this test planted its
+  // own file under a THIRD repo-root mkdtempSync prefix
+  // ('eslint-guardrail-canary-*') that was itself not in either ignore
+  // list — reintroducing the exact #2482 leftover-wedge class this file
+  // exists to close if that plant ever survived a killed process. Probing
+  // a tracked file instead writes nothing, so there is nothing to leak.
+  const target = join(repoRoot, 'server', 'src', 'test-utils', 'quarantine.test.ts');
+  const result = runEslint([target]);
+  assert.doesNotMatch(
     result.output,
-    new RegExp(RULE_ID),
-    `eslint's output should name the ${RULE_ID} rule for a file outside the leftover-dir ignores: ${result.output}`,
+    /ignored/i,
+    `a real tracked *.test.ts file must never be reported as ignored (got exit ${result.exitCode}): ${result.output}`,
   );
 });
--- a/scripts/tests/eslint-guardrail.test.mjs
+++ b/scripts/tests/eslint-guardrail.test.mjs
@@ -16,27 +16,28 @@
 // nothing, exactly the failure mode the ignore entry's comment warns about.
 import { test } from 'node:test';
 import assert from 'node:assert';
-import { writeFileSync, rmSync, mkdtempSync, mkdirSync } from 'node:fs';
+import { writeFileSync, rmSync, mkdtempSync } from 'node:fs';
 import { join } from 'node:path';
-import { execFileSync } from 'node:child_process';
+import { spawnSync } from 'node:child_process';
 
 // test:hooks runs from the repo root (scripts/run-hooks-tests.mjs uses fast-glob
 // which resolves against cwd).
 const repoRoot = process.cwd();
 
-function lintPlantedFile(f) {
-  try {
-    execFileSync('npx', ['eslint', '--no-ignore', f], {
-      cwd: repoRoot,
-      stdio: 'pipe',
-      shell: process.platform === 'win32',
-    });
-    return false;
-  } catch {
-    // eslint exits non-zero when it finds an error-level violation — that is the
-    // expected outcome for the planted violation.
-    return true;
-  }
+const RULE_ID = 'no-restricted-syntax';
+
+// Runs eslint and returns { exitCode, output }. Using spawnSync (not
+// execFileSync) so a non-zero exit doesn't throw — callers need the output
+// on EVERY outcome, not just failure, to tell "the guardrail rule fired" apart
+// from "eslint itself errored" (a bad flag, a missing binary, a config
+// problem) — both exit non-zero, and only the output distinguishes them.
+function runEslint(args) {
+  const result = spawnSync('npx', ['eslint', ...args], {
+    cwd: repoRoot,
+    encoding: 'utf8',
+    shell: process.platform === 'win32',
+  });
+  return { exitCode: result.status, output: (result.stdout ?? '') + (result.stderr ?? '') };
 }
 
 test('guardrail rejects a planted it.skipIf(process.env.CI)', () => {
@@ -49,46 +50,61 @@ test('guardrail rejects a planted it.skipIf(process.env.CI)', () => {
     f,
     "import { it } from 'vitest';\nit.skipIf(process.env.CI)('x', () => {});\n",
   );
-  let failed;
+  let result;
   try {
-    failed = lintPlantedFile(f);
+    // --no-ignore: guardrail-tmp-*/ is ALSO in eslint.config.mjs's global
+    // `ignores` (#2482 — a leftover from THIS probe must not wedge lint on a
+    // later, unrelated checkout). Without --no-ignore this call would exit 0
+    // regardless of the rule's presence, which is exactly the false-pass this
+    // test exists to catch — so the assertion below checks for the rule ID
+    // specifically, not just a non-zero exit, which a bad flag or a broken
+    // eslint invocation would also produce.
+    result = runEslint(['--no-ignore', f]);
   } finally {
     rmSync(dir, { recursive: true, force: true });
   }
-  assert.equal(failed, true, 'eslint should exit non-zero for the planted it.skipIf(process.env.CI) violation');
+  assert.equal(result.exitCode, 1, `eslint should exit 1 for a lint violation (got ${result.exitCode}): ${result.output}`);
+  assert.match(
+    result.output,
+    new RegExp(RULE_ID),
+    `eslint's output should name the ${RULE_ID} rule, not merely exit non-zero: ${result.output}`,
+  );
 });
 
-test('a stale guardrail-tmp-* leftover does not wedge npm run lint', () => {
+test('a stale guardrail-tmp-* leftover is ignored by ESLint, not merely absent of other diagnostics', () => {
   // Simulates the #2482 failure mode: a guardrail-tmp-* dir survives cleanup
   // (killed process / rmSync race) and is still sitting in the tree the next
-  // time someone runs `npm run lint`. Invoked the same way that script does
-  // (`eslint . --max-warnings 0`, a directory walk — NOT an explicit glob
-  // naming the leftover, which ESLint 9 treats as an unmatched-pattern error
-  // regardless of ignores and would make this test assert the wrong thing)
-  // it must be silently skipped by eslint.config.mjs's global ignore.
-  const staleDir = join(repoRoot, 'guardrail-tmp-stale-2482-test');
-  rmSync(staleDir, { recursive: true, force: true });
-  mkdirSync(staleDir);
+  // time someone runs `npm run lint`. Targets the leftover directly via an
+  // explicit glob rather than `eslint . --max-warnings 0` over the whole
+  // repo — a whole-repo lint's exit code says nothing about THIS directory
+  // specifically; it would also flip on an unrelated pre-existing warning
+  // elsewhere in the tree, which is not what this test is meant to prove.
+  //
+  // ESLint 9 treats "every file an explicit glob pattern matches is ignored"
+  // as its own outcome (exit 2, "all matched files are ignored"), distinct
+  // from "the file was linted and found clean" (exit 0) — so a passing run
+  // here is a config guarantee, not silence.
+  const staleDir = mkdtempSync(join(repoRoot, 'guardrail-tmp-'));
   const staleFile = join(staleDir, 'planted.test.ts');
   writeFileSync(
     staleFile,
     "import { it } from 'vitest';\nit.skipIf(process.env.CI)('x', () => {});\n",
   );
-  let failed = false;
+  let result;
   try {
-    execFileSync('npx', ['eslint', '.', '--max-warnings', '0'], {
-      cwd: repoRoot,
-      stdio: 'pipe',
-      shell: process.platform === 'win32',
-    });
-  } catch {
-    failed = true;
+    result = runEslint([`${staleDir.replace(/\\/g, '/')}/**/*.ts`]);
   } finally {
     rmSync(staleDir, { recursive: true, force: true });
   }
   assert.equal(
-    failed,
-    false,
-    'npm run lint should skip a stale guardrail-tmp-* leftover (global ignore) rather than flag its planted violation',
+    result.exitCode,
+    2,
+    `eslint should report every matched file as ignored (exit 2), not lint the leftover (got ${result.exitCode}): ${result.output}`,
+  );
+  assert.match(result.output, /ignored/i, `eslint's output should say the leftover was ignored: ${result.output}`);
+  assert.doesNotMatch(
+    result.output,
+    new RegExp(RULE_ID),
+    `the leftover must never actually be linted, i.e. never surface the ${RULE_ID} violation: ${result.output}`,
   );
 });
--- a/scripts/tests/eslint-guardrail.test.mjs
+++ b/scripts/tests/eslint-guardrail.test.mjs
@@ -31,12 +31,24 @@ const RULE_ID = 'no-restricted-syntax';
 // on EVERY outcome, not just failure, to tell "the guardrail rule fired" apart
 // from "eslint itself errored" (a bad flag, a missing binary, a config
 // problem) — both exit non-zero, and only the output distinguishes them.
+//
+// Windows needs shell:true to run npx.cmd at all (spawnSync cannot exec a
+// .cmd file directly — EINVAL). shell:true only concatenates argv without
+// escaping (Node DEP0190), so every argument is quoted here — cheap
+// insurance since one of them is a path built from mkdtempSync's own output
+// and could contain a space in a checkout outside this repo's convention.
+const IS_WIN = process.platform === 'win32';
+
 function runEslint(args) {
-  const result = spawnSync('npx', ['eslint', ...args], {
+  const argv = IS_WIN ? args.map((a) => `"${a}"`) : args;
+  const result = spawnSync('npx', ['eslint', ...argv], {
     cwd: repoRoot,
     encoding: 'utf8',
-    shell: process.platform === 'win32',
+    shell: IS_WIN,
   });
+  if (result.error) {
+    throw result.error;
+  }
   return { exitCode: result.status, output: (result.stdout ?? '') + (result.stderr ?? '') };
 }
 
@@ -106,5 +118,34 @@ test('a stale guardrail-tmp-* leftover is ignored by ESLint, not merely absent o
     result.output,
     new RegExp(RULE_ID),
     `the leftover must never actually be linted, i.e. never surface the ${RULE_ID} violation: ${result.output}`,
+  );
+});
+
+test('the guardrail-tmp-*/budget-poll-tmp-* ignores do not swallow an ordinary *.test.ts file', () => {
+  // The two tests above prove the leftover dirs ARE ignored and the rule
+  // fires when unignored via --no-ignore. Neither would catch the ignore
+  // pattern becoming too broad (e.g. accidentally matching '**/*.test.ts'
+  // instead of a specific directory prefix) — test 1 always passes
+  // --no-ignore, which would blind it to that too, and test 2 asserts the
+  // file IS ignored, so a broader pattern only makes it greener. This test
+  // is the counterweight: an ordinary planted violation OUTSIDE both
+  // leftover-dir prefixes must still be caught with NO --no-ignore flag.
+  const dir = mkdtempSync(join(repoRoot, 'eslint-guardrail-canary-'));
+  const f = join(dir, 'planted.test.ts');
+  writeFileSync(
+    f,
+    "import { it } from 'vitest';\nit.skipIf(process.env.CI)('x', () => {});\n",
+  );
+  let result;
+  try {
+    result = runEslint([f]);
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+  assert.equal(result.exitCode, 1, `eslint should still lint an ordinary test file and reject the violation (got ${result.exitCode}): ${result.output}`);
+  assert.match(
+    result.output,
+    new RegExp(RULE_ID),
+    `eslint's output should name the ${RULE_ID} rule for a file outside the leftover-dir ignores: ${result.output}`,
   );
 });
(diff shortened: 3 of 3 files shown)


dudarenok-maker:
## PR review — pass 1 (head 633ea02a1fea0307c32107fdf88defcdf262ae59, depth low)

Scope: `eslint.config.mjs`, `scripts/tests/eslint-guardrail.test.mjs` (the only two files changed) · verified before probing: full-repo `eslint . --max-warnings 0` baseline green (exit 0, 32.3 s); all 15 PR checks green; `Closes #2482` present, plain, outside any code span; PR title matches the commit convention.

I did not take the fix on trust — I planted the leftover and ran the real invocations. **The fix itself is correct, and I proved it works for the reason claimed.** Every finding below is in the test code around it.

---

### 🟠 Significant — `lintPlantedFile()` reports "the guardrail fired" whenever ESLint never ran at all

`scripts/tests/eslint-guardrail.test.mjs:27-39` — the helper returns `true` on *any* non-zero exit. Measured here (ESLint 9.39.4), planted violation in place each time:

| argv | exit | helper returns | test 1 |
|---|---|---|---|
| `eslint --no-ignore <f>` (as shipped) | 1 | `true` | passes |
| `eslint <f>` (flag dropped) | 0 | `false` | **fails** — good |
| `eslint --no-ignoer <f>` (flag typo) | 2 | `true` | **passes** — wrong |
| `eslint-does-not-exist --no-ignore <f>` | 1 | `true` | **passes** — wrong |
| `eslint --no-ignore --config no-such-config.mjs <f>` | 2 | `true` | **passes** — wrong |

Row 2 is the mutation the PR body reports, and it genuinely holds. Rows 3–5 are the gap: exit `2` is ESLint's *"I could not run"* code, and the helper reads it as *"the rule rejected the violation"*.

This is load-bearing **because of this PR**. `--no-ignore` is new here, and this test is the only thing certifying that adding the global ignore did not blind the guardrail — yet it cannot distinguish the flag working from the flag being rejected. A future typo in that flag lands green.

Fix: assert that the captured output names `no-restricted-syntax` (or the rule's message), not the exit code alone — `execFileSync`'s thrown error carries `.stdout`.

### 🟠 Significant — the new test asserts the whole repo lints clean, then blames the ignore when it does not

`scripts/tests/eslint-guardrail.test.mjs:60-92` runs `eslint . --max-warnings 0` over the entire tree and asserts exit 0 with the message *"npm run lint should skip a stale guardrail-tmp-\* leftover (global ignore)"*. Any diagnostic anywhere in the tree produces that message — including a mere **warning**, because of `--max-warnings 0`.

Concrete, and not hypothetical: `scripts/tests/check-no-budget-poll.test.mjs:85` and `:118` plant `budget-poll-tmp-<rand>/planted.test.ts` **at the repo root**. Running `eslint . --max-warnings 0` with that file's verbatim content present:

```
budget-poll-tmp-reviewprobe\planted.test.ts
  2:16  warning  'until' is defined but never used...  @typescript-eslint/no-unused-vars
✖ 1 problem (0 errors, 1 warning)     EXIT: 1
```

That directory is left behind by exactly the killed-process / `rmSync` EBUSY scenario this PR exists to defend against, and — see the next finding — nothing ignores it. When it survives, this new test sits permanently red on that checkout while pointing at the wrong mechanism.

There is also a narrower concurrent window: `scripts/run-hooks-tests.mjs` spawns `node --test` over ~90 files, and file-level concurrency is real (measured: two 3 s test files → 3.17 s wall; 16-way parallelism here), so that directory can exist *transiently* during this test's repo-wide walk. **In fairness, I tried to reproduce that and it did not fire** — the budget-poll plants are sub-second and finish before the 34 s walk begins. So I am reporting the surviving-leftover path as the deterministic one, and not claiming a proven flake.

Fix: assert on what the walk said about the stale dir — capture the output and assert it never names `guardrail-tmp-`, or diff the exit code with and without the leftover — rather than asserting global cleanliness.

### 🟠 Significant — one instance fixed, the class left armed

`budget-poll-tmp-*` is the identical pattern (in-repo root temp dir, `rmSync` in a `finally`, killable) and is ignored by **neither** git **nor** ESLint:

- `git check-ignore -v --no-index budget-poll-tmp-abc123/planted.test.ts` → exit 1 (not ignored);
- absent from `eslint.config.mjs`'s `ignores`;
- measured above: a surviving leftover exits `eslint . --max-warnings 0` non-zero.

So #2482's wedge is still reachable today, from a sibling probe in the same directory as the one being fixed. It is at least *visible* in `git status`, unlike the guardrail one, so the diagnosis is less brutal — but the wedge is identical. Fix: add `budget-poll-tmp-*/` to both `.gitignore` and the `ignores` block this PR is already editing, or move that probe under an already-ignored root.

### 🟡 Minor — measured +27–34 s on `test:hooks`, which pre-commit runs

Standalone `eslint . --max-warnings 0`: 32.3 s / 27.1 s / 26.6 s across three runs. The new test self-reported `34237.9 ms` in a live `node --test` run. `test:hooks` is scoped in by `scripts/**/*.{mjs,cjs,js,mts,cts,ts}`, `.claude/skills/**`, `.claude/agents/**`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/testing/**`, `eslint.config.mjs` … and runs inside pre-commit's `verify:fast:scoped`, which CLAUDE.md documents as *"Sub-5s on a warm cache"*. `verify:fast:branch` now also lints the whole repo twice — its own `lint` step plus this test.

Keeping "the same invocation `npm run lint` uses" inherently costs a full walk, so this may well be an accepted trade — but right now it is an unacknowledged one. Say so in the test's comment if it is deliberate.

### 🟡 Minor — fixed directory name where the sibling test uses `mkdtempSync`

`scripts/tests/eslint-guardrail.test.mjs:69` uses the constant `guardrail-tmp-stale-2482-test`; the test 24 lines above uses `mkdtempSync(join(repoRoot, 'guardrail-tmp-'))` — for exactly this reason. Two `node --test` runs in one checkout (a pre-commit firing while the developer runs `npm run test:hooks`) race: run B's leading `rmSync(staleDir, { recursive: true, force: true })` deletes run A's directory, A's walk then finds no leftover, and A's `assert.equal(failed, false)` passes **vacuously** — green while testing nothing. `mkdtempSync` with the same prefix costs nothing and removes it.

### 🟡 Minor — release-notes pair neither shipped nor explicitly waived

CLAUDE.md before-shipping step 5 permits skipping for a CI-only change with no user- or operator-visible effect, but requires saying so explicitly rather than silently omitting. The PR body does neither. Nothing enforces it per-PR — `scripts/release-notes-gate.mjs` runs only in `release.yml` at tag time — so the convention is the only thing holding it. A one-line waiver is probably the right answer, though a developer whose lint was wedged is arguably operator-visible.

---

### ✅ What is solid

Verified, not assumed:

- **The fix works.** `guardrail-tmp-review-probe/planted.test.ts` carrying the violation → `eslint . --max-warnings 0` exit **0**.
- **It works for the stated reason, not because the file was invisible.** Control: the identical file in a non-ignored root dir `zz-review-control-tmp/` → exit **1**, `no-restricted-syntax`. So the new test *can* fail — it is not a test that cannot fail — and `guardrail-tmp-*/` (trailing slash, root-anchored) is a working flat-config directory ignore.
- **`--no-ignore` genuinely preserves the guardrail** — the key risk #2482's header warned about. With it: exit 1 plus the rule message. Without it: exit 0 plus `File ignored because of a matching ignore pattern`. The mutation the PR body claims does hold.
- **Scope wiring was already right.** `eslint.config.mjs` is an `extraFiles` input of `test:hooks` in `scripts/verify-cache.mjs`; `stepTouchedByDiff` returns `true` for an `eslint.config.mjs`-only diff, so deleting the new ignore entry schedules this guard rather than printing `[cached]`.
- `guardrail-tmp-stale-2482-test` is covered by `.gitignore:76`, so the new test does not dirty `git status`.
- The comments on both sides cross-reference each other correctly, and the "why not an explicit glob" note about ESLint 9's unmatched-pattern error is accurate.

### Verdict

**The fix is correct, minimal, and proven — the tests around it are weaker than the fix.** Nothing blocking: no user-facing behaviour is wrong, and the mechanism does what the PR says it does. Three correctness findings, all in test code. Findings 1 and 3 are the ones I would not merge without: 1 leaves the newly added flag uncertified against its own most likely failure, and 3 leaves the very bug being fixed reachable from the file next door.

dudarenok-maker:
## PR review — pass 2 (head 9f32bf3198014c99a5aa1119abfcd6a4b533ce0b, depth low)

Scope: the three changed files (`.gitignore`, `eslint.config.mjs`, `scripts/tests/eslint-guardrail.test.mjs`) plus the two files finding 3 reaches into (`scripts/tests/check-no-budget-poll.test.mjs`, `scripts/check-no-budget-poll.mjs`) · verified before probing: all 15 checks green on this head; working tree clean at `9f32bf31`; `node --test` over both affected test files → **11/11 pass, 7.5 s wall**.

Deltas only. Every disposition below is a measurement re-run on this head against ESLint 9.39.4, not a re-read of the commit message.

---

### Prior findings

**1. `lintPlantedFile()` read any non-zero exit as "the guardrail fired" — ✅ VERIFIED RESOLVED**

Pass-1's table, re-run against the rewritten `runEslint()` + the two new assertions:

| argv | exit | rule ID in output | test 1 now |
|---|---|---|---|
| `eslint --no-ignore <f>` (as shipped) | 1 | yes | passes ✔ |
| `eslint --no-ignoer <f>` (flag typo) | 2 | no | **fails** ✔ |
| rule set to `off` for `**/*.test.ts` | 0 | no | **fails** ✔ |

Row 2 was pass 1's false-pass; it is closed. The exit-code assertion catches it first, and the `assert.match(RULE_ID)` closes the residual class (anything that exits 1 without the rule having fired — a parse error, a different rule, a missing binary). Row 3 re-verifies the mutation the commit message claims: a mutant config that disables `no-restricted-syntax` on `**/*.test.ts` and changes nothing else → exit **0**, empty output → test 1 red. The claim holds as stated.

**2. whole-repo `eslint . --max-warnings 0` replaced by a targeted glob — ✅ VERIFIED RESOLVED, and the sharp edge of the new approach is covered**

The exit-2 signal is real and the three assertions discriminate correctly. Measured, same planted leftover each time:

| invocation | exit | `/ignored/i` | rule ID | test 2 |
|---|---|---|---|---|
| `eslint <dir>/**/*.ts` (as shipped) | 2 | yes | no | passes ✔ |
| `eslint --no-ignore <dir>/**/*.ts` (= ignore entry absent) | 1 | no | **yes** | **fails** ✔ |
| same glob, directory deleted (glob matches nothing) | 2 | **no** | no | **fails** ✔ |

Row 2 is the mutation that matters: strip the ignore and the test reddens on all three assertions, so it has not become a test that cannot fail. Row 3 is the trap I went looking for — exit 2 is also ESLint's *"No files matching the pattern … were found"* code, so an exit-code-only assertion would have false-passed on a glob that matched nothing. The `assert.match(result.output, /ignored/i)` at `scripts/tests/eslint-guardrail.test.mjs:104` is what closes it, and it does: the no-match message contains no form of the word. Good call adding it rather than asserting exit 2 alone.

**3. `budget-poll-tmp-*` sibling wedge — ✅ VERIFIED RESOLVED, and the class is closed, not just the instance**

Planted `budget-poll-tmp-<rand>/planted.test.ts` at the repo root with the *verbatim* violating fixture from `check-no-budget-poll.test.mjs:85`, then measured on this head:

- `npx eslint . --max-warnings 0` → exit **0** (30.0 s). Pass 1 measured exit 1 on the same input. Wedge closed.
- `git status --porcelain` with the leftover present → **empty**.
- `node --test scripts/tests/check-no-budget-poll.test.mjs` → 9/9 pass. The new ignores do **not** blind that probe, and I checked why rather than assuming: it drives `scripts/check-no-budget-poll.mjs`, which enumerates via `readdirSync` and knows nothing about ESLint or `.gitignore`, and `check:budget-poll`'s inputs are `server/src/**/*.test.ts`, not the repo root.
- Class check: `mkdtempSync(join(repoRoot, …))` has exactly two call sites in the tree — `guardrail-tmp-` and `budget-poll-tmp-`. Every other `mkdtempSync` in `scripts/` targets `os.tmpdir()`. There is no third sibling left armed.

**4. 🟡 fixed dir name vs the sibling's `mkdtempSync` — ✅ RESOLVED.** `eslint-guardrail.test.mjs:87` now uses `mkdtempSync(join(repoRoot, 'guardrail-tmp-'))`; the leading `rmSync` of a shared constant name is gone, so the vacuous-green race is gone with it.

**5. 🟡 `test:hooks` runtime cost — largely resolved as a side effect; the comment note is still absent.** Measured on this head: `guardrail rejects a planted …` **3954 ms**, `a stale guardrail-tmp-* leftover …` **3439 ms** — 7.4 s for the file, against pass 1's 34.2 s for the second test alone. The whole-repo walk is gone from the test path, so the objection largely evaporates on its own. Not worth blocking on; if you still want the acknowledgement, it is one comment line.

**6. 🟡 release-notes pair neither shipped nor waived — ⬜ STILL OPEN.** `git diff --name-only origin/main...HEAD` is still the same three files; the PR body still says nothing either way. Not blocking (CLAUDE.md step 5 permits skipping a CI-only change), but the waiver has to be *stated*, and it is a one-line edit to the body you are editing anyway for the finding below.

---

### 🟠 Significant (new) — both probes are blind to an over-broad global ignore, i.e. to the exact surface this PR is editing

`scripts/tests/eslint-guardrail.test.mjs:62` (`--no-ignore`) and `:95` (glob → asserts "was ignored").

`--no-ignore` disables *every* ignore, not just `guardrail-tmp-*/`. So test 1 no longer asks "does the shipped config reject this file?", only "does the rule exist if I switch the ignore list off?". Test 2 asserts the leftover **is** ignored, so a broader ignore makes it *more* green. Between them there is now no assertion that the guardrail applies to a test file under the config as actually shipped.

Measured, with a mutant config that appends one line — `{ ignores: ['**/*.test.ts'] }` — and changes nothing else:

| invocation | exit | rule ID | verdict |
|---|---|---|---|
| test 1's, mutant config | 1 | yes | **green — blind** |
| test 2's, mutant config | 2 | no (`ignored`) | **green — blind** |
| `eslint <config=mutant> server/src/**/*.test.ts` | 2 | no | **every tracked test file in the repo is now un-linted** |

Control: the same three invocations against `--config eslint.config.mjs` reproduce the shipped results exactly, so the only variable is the added ignore.

So: a single fat-fingered pattern in the block this PR is *currently editing* silently retires `no-restricted-syntax` across the entire repo, `npm run lint` stays green, and both tests whose sole job is to certify that rule stay green too. That is recurring shape 5 in the reviewer brief — an acceptance criterion blind to its own feature.

To be fair about provenance: this arrived in `633ea02a` with `--no-ignore`, not in `9f32bf31`, and pass 1 passed it ("`--no-ignore` genuinely preserves the guardrail" — it preserves the *rule*, but not the *ignore coverage*, and I did not separate those two last round). Before `633ea02a` test 1 did hold the property: it linted a file in a then-un-ignored dir with no flag, so the mutant above reddened it (an explicitly-named ignored file exits **0** with a `File ignored…` warning — measured). It is not a regression this commit introduced; it is a coverage loss this PR's mechanism causes, still open, and this is the round to close it.

Fix, and it is cheap: keep both existing probes, add a third that plants the same violation in a directory that is **not** in the global ignores (e.g. an `lint-probe-tmp-*/` prefix, `.gitignore`-only) and lints it with **no** `--no-ignore`, asserting exit 1 + the rule ID. That restores the pre-`633ea02a` property without weakening anything the current two prove.

*Related, same line, one sentence rather than its own finding:* test 2 now asserts a fact about ESLint's config resolution, not about `npm run lint` (`package.json:39` — `eslint . --max-warnings 0`). Adding `--no-ignore` to that npm script would reopen #2482 verbatim with both tests green. I verified the end-to-end property holds today by hand (finding 3 above, exit 0), so nothing is currently wrong — but nothing tests the script either, and the PR body still claims otherwise (below).

---

### 🟡 Minor (new) — the PR body now describes the mechanism this commit deleted, and does not declare the incidental fix

Three separate staleness points, all in the body that becomes the merge-commit record:

1. Test plan bullet 1: *"runs the same `eslint . --max-warnings 0` invocation `npm run lint` uses — confirms it's skipped, not flagged."* `9f32bf31` deliberately removed that invocation. The body now advertises the property the fix traded away for precision, which is the opposite of what a reviewer reading the merge commit needs.
2. Summary: *"`rmSync` losing a Windows EBUSY race against the `execFileSync` child right above it"* — `execFileSync` is no longer imported by that test.
3. The `budget-poll-tmp-*` work (finding 3, two of the three changed files) is nowhere in the body. CLAUDE.md is explicit: *"Incidental fixes are also declared in the PR body ('Also fixed, found in passing: …') — an unannounced fix reads as scope creep to a reviewer."* It also wants an issue filed recording it.

Bullets 4–5 of the test plan (`test:hooks` 1416/1416, `verify:fast:branch` green) date from the pre-fix commit; re-confirm or re-state them.

### 🟡 Minor (new) — `eslint.config.mjs:57` still cites `execFileSync`

> `// race against the execFileSync child right above it) must not wedge every`

The commit that rewrote the probe to `spawnSync` left the config comment pointing at a call that no longer exists — and "right above it" refers to code in a different file, which made the sentence hard to check even before it went false. `scripts/tests/eslint-guardrail.test.mjs:30` names `execFileSync` correctly (as the thing it moved *away* from), so only the config comment is wrong.

### 🟡 Minor (new) — `shell: true` concatenates argv unescaped, and test 2 is newly sensitive to it

`scripts/tests/eslint-guardrail.test.mjs:35` — `spawnSync('npx', […], { shell: process.platform === 'win32' })`. Node emits `DEP0190` on this call in every run of the suite (visible in the `node --test` output), because with `shell: true` the arguments are concatenated, not escaped. Measured, by planting the file one directory deeper under `a space/`:

- test 1's shape → `No files matching the pattern ".../guardrail-tmp-JZJK6S/a" were found.` → exit 2 → red.
- test 2's shape → same truncation → exit 2 but **no** `ignored` in the output → red.

Both fail loud rather than false-pass, so this is robustness, not a hole. It matters only because the exposure is **new for test 2**: its old argv was `['eslint', '.', '--max-warnings', '0']` — no path argument at all, and `cwd` is passed as a spawn option, so a checkout under `C:/Users/First Last/…` was fine. It now passes an absolute glob and inherits test 1's pre-existing sensitivity. Fix once in the helper (quote, or drop `shell` and resolve `npx.cmd`) and both are covered.

Same helper, one line: `runEslint()` discards `result.error` (`:40`). If `npx` cannot be spawned at all, `exitCode` is `null` and `output` is `''`, so the failure reads `eslint should exit 1 for a lint violation (got null):` with nothing after the colon. Fails closed, but diagnoses nothing — surface `result.error?.message` in the returned output.

---

### Verdict

**All three correctness findings from pass 1 are verifiably fixed, and I could not make any of the three fixes false-pass.** Fix 2 in particular is better than what it replaced: the exit-2 signal has a genuine ambiguity (no-files-found), and the `/ignored/i` assertion is exactly the right thing to close it.

One new correctness finding: the `--no-ignore` blind spot. It is not a regression from this commit and it does not make anything wrong today — but this PR spends its whole diff on the global-ignore list, and after it lands, that list is the one thing no test defends. That is the one I would fix before merge. The three new 🟡s are a body edit, a comment line, and a helper tweak; pass 1's release-notes 🟡 is still open and is one more line of the same body edit.

dudarenok-maker:
## PR review — pass 3 (head 061ea6e92b1c29ceb2b36771c2d94b7d4c60d63e, depth low)

Scope: the three changed files · verified before probing: working tree clean at `061ea6e9`; `node --test scripts/tests/eslint-guardrail.test.mjs` → **3/3 pass, 13.2 s**; `npm run test:hooks` → **1417/1417 pass, 49 s** (confirms the body's own claim).

Deltas only. Every disposition below is a re-measurement on this head against ESLint 9.39.4 — I re-ran all three claimed fixes plus all of pass 1's and pass 2's mutations, and each mutation harness was validated with a control first.

---

### Prior findings

**1. 🟠 Both probes blind to an over-broad global ignore — ✅ VERIFIED RESOLVED (for the pattern it named)**

Harness: wrap the real config from a temp-dir file, pass via `--config`. The control must reproduce the shipped result or the harness proves nothing — it does:

| config | exit | rule ID | `ignored` | test 3 |
|---|---|---|---|---|
| SHIPPED (auto-discovered) | 1 | yes | no | **GREEN** |
| CONTROL (`--config` wrapper, no mutation) | 1 | yes | no | **GREEN** ✔ harness valid |
| MUTANT (wrapper + `{ ignores: ['**/*.test.ts'] }`) | 0 | no | yes | **RED** ✔ |

Mutation-free cross-check that the new probe is genuinely ignore-sensitive rather than just exit-code-sensitive: the *same* invocation on a file planted under an already-ignored `guardrail-tmp-*` dir → exit 0, `ignored`, test 3 **RED**. So test 3 reddens on the ignore mechanism itself, not on an incidental. The claim holds as stated.

**2. 🟡 Stale PR-body claims / undeclared incidental fix — ✅ VERIFIED RESOLVED.** The `eslint . --max-warnings 0` test-plan bullet is gone, `execFileSync` is now "the spawned child right above it", and both incidental fixes are declared (`budget-poll-tmp-*` from round 1, the canary probe from round 2). Bullets 4–5 are restated at 1417/1417, which I reproduced.

**3. 🟡 `eslint.config.mjs:57` cited `execFileSync` — ✅ VERIFIED RESOLVED.** Now reads `spawnSync`. (`.gitignore:76` also became bare `.gitignore`; line 76 was in fact correct, but dropping a line number is strictly more robust, so no objection.)

**4. 🟡 `shell:true` concatenates argv unescaped — ✅ VERIFIED RESOLVED, and I measured the exact break pass 2 demonstrated.** Planted one directory deeper under `a space/`, running each test's own argv shape both ways:

| shape | QUOTED (shipped) | UNQUOTED (pre-fix `9f32bf31`) |
|---|---|---|
| test 1 `--no-ignore <file>` | exit 1, rule ID → **PASS** | exit 2, `No files matching…` → **FAIL** |
| test 2 `<staleDir>/**/*.ts` | exit 2, `ignored` → **PASS** | exit 2, no `ignored` → **FAIL** |
| test 3 `<file>` (no flag) | exit 1, rule ID → **PASS** | exit 2, `No files matching…` → **FAIL** |

All three shapes are now space-safe; all three broke before. The `shell: true`-is-required-for-`npx.cmd` justification in the new comment is accurate (Node refuses to exec `.cmd` without a shell since the CVE-2024-27980 fix). Note the DEP0190 warning still prints on every run — quoting removes the *hazard*, not the notice; cosmetic only.

**Prior tests still discriminate after the helper rewrite** — this was the "round N+1 breaks N's fixes" check. All seven cases behave exactly as before the rewrite:

| mutation | exit | test verdict |
|---|---|---|
| test 1 baseline | 1 | GREEN ✔ |
| test 1, `no-restricted-syntax: 'off'` for `**/*.test.ts` | 0 | RED ✔ |
| test 1, typo'd `--no-ignoer` | 2 | RED ✔ |
| test 1, `--no-ignore` dropped | 0 | RED ✔ |
| test 2 baseline | 2 | GREEN ✔ |
| test 2, `--no-ignore` (= ignore entry absent) | 1 | RED ✔ |
| test 2, glob matches nothing | 2 | RED ✔ |

**5. 🟡 Release-notes pair neither shipped nor waived — ⬜ STILL OPEN (third round).** `git diff --name-only origin/main...HEAD` is still the same three files; `gh pr view 2484 --json body` still contains no case-insensitive match for "release note". CLAUDE.md step 5 permits skipping a CI-only change but requires saying so explicitly. One line.

---

### 🟠 Significant (new) — the new canary probe reintroduces #2482's own wedge, as a third armed sibling

`scripts/tests/eslint-guardrail.test.mjs:133` — `mkdtempSync(join(repoRoot, 'eslint-guardrail-canary-'))`.

That prefix is in **neither** ignore list:

- `git check-ignore -v --no-index eslint-guardrail-canary-abc123/planted.test.ts` → **exit 1** (not ignored);
- `grep -n canary eslint.config.mjs .gitignore` → **no match**.

Measured on this head, in this worktree, planting the leftover the test would leave behind on a killed run:

```
WITH-LEFTOVER git status: "?? eslint-guardrail-canary-PROBE1/"
`npm run lint` (eslint . --max-warnings 0) EXIT: 1
  eslint-guardrail-canary-PROBE1\planted.test.ts
    2:11  error  No it.skipIf(process.env.CI). Use quarantinedIt ...  no-restricted-syntax
  ✖ 1 problem (1 error, 0 warnings)
```

That is #2482 verbatim — `npm run lint` red on a planted violation nobody wrote — reintroduced by the commit that was closing the blind spot. `verify-cache.mjs:1167` runs steps sequentially with `lint` ahead of `test:hooks`, so a leftover wedges the *next* run's lint step before the test file ever gets a chance to clean up after itself; a leading sweep inside the test would therefore not rescue it.

Pass 2 explicitly certified the class closed: *"`mkdtempSync(join(repoRoot, …))` has exactly two call sites in the tree — `guardrail-tmp-` and `budget-poll-tmp-`. There is no third sibling left armed."* There are now three. Reviewer-brief shape 10.

Two things that bound it, stated fairly: the leftover **is** visible in `git status` (unlike `guardrail-tmp-*`, whose invisibility is what made #2482 brutal), so the diagnosis is much easier; and the wedge is lint-only — root `tsconfig.json` has `include: ["src", "vite.config.ts", "vitest.config.ts"]`, so typecheck is unaffected.

**The reflex fix is wrong here.** Adding `eslint-guardrail-canary-*/` to the `ignores` block would make test 3 assert nothing at all — being linted is the whole point of the dir. See the finding below for a fix that closes this *and* widens the coverage.

### 🟡 Minor (new) — the canary closes a narrower class than the one pass 2 named

Test 3 plants at the **repo root**. Pass 2's stated fear was *"a single fat-fingered pattern in the block this PR is currently editing silently retires `no-restricted-syntax` across the entire repo."* A fat-finger scoped to where the real test files actually live does exactly that and leaves test 3 green:

| appended `ignores` entry | test 3 (root canary) | tracked `server/src/analyzer/analyzer-concurrency.test.ts` |
|---|---|---|
| — (control) | GREEN | linted (`ignored=false`) |
| `'**/*.test.ts'` | **RED** ✔ | un-linted (`ignored=true`) |
| `'server/**/*.test.ts'` | **GREEN — blind** | un-linted (`ignored=true`) |

Row 3 un-lints every tracked server test file while all three probes stay green.

Both this and the 🟠 above are closed by the same change, and it is smaller than what is there now: assert the property against an **already-tracked** `*.test.ts` file instead of a planted root-level one — lint e.g. `server/src/analyzer/analyzer-concurrency.test.ts` and assert the output does not say `ignored`. Measured in the table above: that probe reddens on **both** mutants, needs no `mkdtempSync`, writes nothing into the repo, and so cannot leave a leftover to wedge anything.

---

### Verdict

**All three fixes this commit claims are real, and I could not make any of them false-pass** — each was measured against a validated control, and the prior two tests still discriminate on all seven of pass 1's and pass 2's mutations. Fix 3 (argv quoting) is the standout: it repaired a break in all three shapes, including one test 1 had carried since `633ea02a`.

It does not close the loop. The round introduced one new correctness finding of the same class the PR exists to fix: a third repo-root temp-dir prefix in neither ignore list, whose leftover reds `npm run lint` — measured, not theorised. It is milder than #2482 (visible in `git status`, lint-only) and the fix is a simplification rather than an addition, but shipping it would leave the PR reintroducing its own bug one directory over.

This is re-review round 2 of 2, so the loop cap is reached and I am not auto-looping further — handing back per the skill. Two open items: the 🟠 above (with the 🟡 folded into the same edit) and the still-unwaived release-notes line from pass 1.

dudarenok-maker:
## PR review — round 4 note (head 8ff16ec8, depth low) — self-fixed, loop cap reached

Round 3 (aee8ea30) hit the 2-re-review-round cap after finding a new 🟠: the round-2 fix's third probe planted its own file under an unignored `eslint-guardrail-canary-*` prefix, reproducing #2482's own bug class if that plant ever survived a killed process. Per the pr-review-gate skill, the loop stops automatically at the cap and hands back to the coordinating session rather than auto-dispatching a 4th independent pass.

Fixed in `8ff16ec8` (self-verified, not an independent review pass): the third probe now targets an already-tracked file (`server/src/test-utils/quarantine.test.ts`) instead of planting one — writes nothing, so nothing can leak. Re-ran round 3's own mutation methodology against it:
- Appending an over-broad `**/*.test.ts` ignore reddens this probe while the other two stay green.
- `git status --porcelain` confirmed clean after every run (planted-dir tests and the new tracked-file probe alike).
- Full `npm run test:hooks` — 1417/1417 pass.

### Disposition of round 3's findings
- 🟠 canary-planting-a-fourth-leftover-prone-dir — **FIXED** (`8ff16ec8`), verified above.
- 🟡 release notes still unshipped/unwaived — **closed**: PR body now states this is a pure CI-only internal chore with no shippable delta (before-shipping checklist step 5 exemption).

This note documents the fix for the record; it is not an independent reviewer pass. Flagging for the repo owner to confirm no further round is wanted before merge.

PR #2489, opened 2026-08-20
chore(ci): de-duplicate the Windows ffmpeg install step

### LINKED ISSUE #2480: chore(ci): de-duplicate the Windows ffmpeg install step across cross-os.yml and release.yml
## What
`.github/workflows/cross-os.yml` and `.github/workflows/release.yml` each carry a near-identical ~50-line "Install ffmpeg (Windows)" PowerShell step (choco install, verify-actually-runs, winget fallback, retry loop, $GITHUB_PATH persist per #2478). Flagged as a minor finding on PR #2479's review pass (both round 1 and round 2 — round 2 noted the original filing of this issue didn't name the decision owed, which is why it's being re-edited now).

## Decision owed
More than one defensible extraction shape exists and someone has to pick:
1. **A composite action** (`.github/actions/install-ffmpeg-windows/action.yml`, invoked via `uses:` from both workflows) — matches this repo's existing `./.github/actions/setup` precedent, but composite-action `run:` steps require an explicit `shell: pwsh` (no OS-based default the way workflow-level `run:` steps get), and `scripts/tests/cross-os-ffmpeg-install.test.mjs`'s extraction logic (`windowsFfmpegStepBlock`, keyed on `WORKFLOWS = [cross-os.yml, release.yml]`) would need reworking to pull the script from the action file instead of (or in addition to) each caller workflow.
2. **A shared script file** (e.g. `scripts/ci/install-ffmpeg-windows.ps1`) invoked identically via `run: pwsh -File ...` from both workflows — simpler to keep the existing test extraction logic pointed at, but is a new shape for this repo's CI (no existing precedent the way composite actions have `./.github/actions/setup`).

Whichever shape is picked, the regression test's extraction/execution harness needs to test the ONE shared artifact directly rather than duplicating its 12 assertions against two near-identical copies embedded in each workflow file.

## Acceptance
- One of the two shapes above (or another defensible one) is chosen and the duplication is removed.
- `scripts/tests/cross-os-ffmpeg-install.test.mjs` is updated accordingly; all existing assertions (or their equivalent) still pass.
- No behavior change — this is a pure DRY extraction.

## Key files
- `.github/workflows/cross-os.yml`, `.github/workflows/release.yml` (Install ffmpeg (Windows) step)
- `scripts/tests/cross-os-ffmpeg-install.test.mjs` (extraction logic, `WORKFLOWS` array)
- `.github/actions/setup/` (existing composite-action precedent, if option 1 is chosen)

## Benefit (technical): removes ~50 duplicated lines between two workflow files, so the next behavior fix (like #2478) only needs to land once.

dudarenok-maker:
## Summary
- Extracts the duplicated "Install ffmpeg (Windows)" step (choco install + winget fallback + bounded retry + `$GITHUB_PATH`/LOCALAPPDATA handling; executable lines identical, comments reflowed differently) out of `cross-os.yml` and `release.yml` into a new composite action, `.github/actions/install-ffmpeg-windows/action.yml`, mirroring the existing `.github/actions/setup` precedent.
- Both workflows now `uses: ./.github/actions/install-ffmpeg-windows` instead of embedding the ~90-line PowerShell script. Pure extraction — no change to the install/retry/fallback logic.
- macOS (`brew install ffmpeg`) and Ubuntu (`apt-get install ffmpeg`) steps are untouched — out of scope per the ticket.
- Updated `scripts/tests/cross-os-ffmpeg-install.test.mjs`: the parser/behavioural test suite now targets the single composite-action source of truth instead of looping over two duplicated copies, plus guards (per workflow, and file-wide) that a future edit can't silently reintroduce the duplication, plus an assertion pinning the composite step's `shell: pwsh`.

Also fixed, found in passing (review pass 1, all 🟡 Minor, no re-review triggered):
- the composite action's `shell: pwsh` had no test coverage — added an assertion that fails if it's changed;
- the "delegates to composite action" guard only scanned the first matching step block per workflow file — widened to a file-wide check;
- "byte-duplicated"/"byte-identical" wording in the action file header and test comments was imprecise (executable lines were identical, comments were not) — corrected.

Release notes: skipped — CI-only chore, no user-facing or operator-visible behavior change.

## Test plan
- [x] `node --test scripts/tests/cross-os-ffmpeg-install.test.mjs` — 9/9 pass
- [x] `npm run verify:fast:branch` — lint, typecheck, config:check, test:hooks, test, test:server, build, test:sidecar all green
- [x] Independent PR review (pass 1, depth low) — 4 🟡 Minor findings, all fixed and pushed; no correctness bugs, no re-review round required

Closes #2480


dudarenok-maker:
## PR review — pass 1 (head dd70f684925dc4e7ba0e345f3eb63fc3534d963a, depth low)

Adversarial gate pass on the "pure mechanical extraction" claim. **The claim holds for the script itself** — I byte-compared the extracted body against `85126b7b:.github/workflows/cross-os.yml`'s prior `run:` body and they are identical (4807/4807 chars, `diff` clean). Checkout precedes the `uses:` in both call sites, `.github/actions/**` is already a `computeShared` member (`scripts/verify-cache.mjs:137,740`) so the new file is inside the scope net, no third inline copy exists anywhere under `.github/`, and 9/9 tests pass locally.

Four findings, all 🟡. No 🔴/🟠 — nothing shipped here behaves wrongly.

---

### 🟡 1. The new `shell: pwsh` key is pinned by nothing — mutation-verified

`.github/actions/install-ffmpeg-windows/action.yml:36`

Moving the script into a composite action introduced a degree of freedom the inline `run:` step did not have: an explicit, **required** `shell:` key. Nothing in the regression suite reads it.

**Reproduced.** I copied the repo subset into a temp sandbox, changed `shell: pwsh` → `shell: bash`, and re-ran `node --test scripts/tests/cross-os-ffmpeg-install.test.mjs`: **9 pass, 0 fail.** The suite is blind because `runPowerShellFile` (line ~183) executes the extracted body under whatever PowerShell `findPowerShell()` discovered, entirely independent of what the action file declares.

**Failure scenario:** a later edit sets `shell: bash` (or drops the key, which Actions rejects outright as a missing required property). Both `cross-os.yml`'s and `release.yml`'s Windows legs die at step start — bash cannot execute PowerShell. Neither workflow runs on a PR, so the first execution is the Wed/Sun cron or a release tag, and in `release.yml` this step sits inside `cross-os-verify`, which gates publish.

This is the exact hazard #2480 named when it laid out option 1 ("composite-action `run:` steps require an explicit `shell: pwsh` — no OS-based default the way workflow-level `run:` steps get"), and its acceptance line asks for the test to be updated accordingly.

**Fix:** one assertion in the composite-action test block, e.g. `assert.match(block, /\n\s*shell:\s*pwsh\b/, ...)`.

*(Severity: considered 🟠 and landed on 🟡 — the shipped value is correct, so this is CLAUDE.md's "a seam you touched that has no test" chore, and a one-line assertion does not warrant reopening the review loop.)*

---

### 🟡 2. The delegation guard is step-scoped, not file-scoped — mutation-verified

`scripts/tests/cross-os-ffmpeg-install.test.mjs:565-583`

The guard's own comment says it exists so "a future edit can't quietly reintroduce an inline duplicate". It actually checks only the **first** `- name: Install ffmpeg (Windows)` step block, because `windowsFfmpegStepBlock` uses `source.indexOf(marker)` and stops at the next 6-space `- name:`.

**Reproduced.** With the delegating step left intact, I appended a second job to `release.yml`:

```yaml
  smoke-windows:
    runs-on: windows-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v6
      - name: Install ffmpeg (Windows)
        run: |
          choco install ffmpeg -y --no-progress
          & ffmpeg -version
```

Result: **9 pass, 0 fail.** A verbatim second copy of the thing this PR exists to de-duplicate goes unnoticed. (For contrast, the guard does work for the shape it names: reverting `cross-os.yml` to its pre-PR inline form reddens it correctly.)

`release.yml` already carries five `Install ffmpeg` steps across four jobs, so "someone adds another Windows leg" is the realistic growth path, not a contrived one.

**Fix:** add a file-scoped assertion alongside the block-scoped one — `assert.doesNotMatch(source, /choco install ffmpeg/, ...)`. The composite action is now the only legal home for that string in a workflow file, so the whole-file form is not over-broad.

---

### 🟡 3. "byte-duplicated" / "byte-identical" is false as written

`.github/actions/install-ffmpeg-windows/action.yml:2`, `scripts/tests/cross-os-ffmpeg-install.test.mjs:25`, plus the PR body and commit message.

Measured on the base commit: `cross-os.yml`'s run body is 4807 chars, `release.yml`'s is 4847. They differ — the PowerShell comment block above the `$GITHUB_PATH` persist was reflowed differently, and `release.yml`'s variant carried an extra clause ("same step shape as this file's") that this PR drops. The **executable** lines are byte-identical, which is what makes the extraction safe; the surrounding comments were not.

#2480 itself said "near-identical"; this PR upgraded that to "byte-identical". Minor on its own, but it is now committed into a test file's header, where the next reader will take it as measured fact — and this repo's review history is mostly prose findings.

**Fix:** "identical apart from reflowed comments" (or "the executable lines were byte-identical") in the four places.

---

### 🟡 4. PR body carries no release-notes disposition

Before-shipping checklist step 5 wants the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair **or** an explicit not-applicable. This PR has neither. It is genuinely skippable — CI-only, no user- or operator-visible delta — but the sentence is the deliverable, and the immediately preceding CI-only chore (#2484) carries it verbatim: "**Release notes intentionally skipped** — pure CI-only internal chore ... (before-shipping checklist step 5)."

**Fix:** add that line to the PR body.

---

### Recommended before merge (not a finding)

Nothing in this PR's green check set executes the new composite action — `cross-os.yml` and `release.yml` neither run on PRs, so the wiring (`uses:` resolution, `shell:`, `$GITHUB_PATH` propagation to later steps) is unproven until a cron fire or a release tag. `gh workflow run cross-os.yml --ref chore/ops-dedup-windows-ffmpeg` proves it in ~20 min of free runner time, and a release tag is a bad place to discover a broken local-action path.

### Checked and clean

- Extracted script byte-identical to the prior `cross-os.yml` body; no logic, ordering, or message drift.
- `actions/checkout@v6` precedes the `uses:` step in both jobs (a local action cannot resolve before checkout).
- `if: matrix.os == 'windows-latest'` retained at both call sites; the composite's inner step correctly carries no `if:` of its own.
- `shell: pwsh` matches the Windows runner's implicit `run:` default, and the script uses nothing that diverges between PowerShell 5.1 and 7.
- `.github/actions/**` is a `computeShared` member and a `test:hooks` input glob, so an actions-only diff still runs this suite locally and every leg in CI — no scope-filter fail-open.
- No third copy of the script: every other `Install ffmpeg` step under `.github/` is `apt-get` or `brew`.
- `action.yml`'s claim that the setup action's header documents the ffmpeg-skip rationale checks out (`.github/actions/setup/action.yml:12-14`).
- `Closes #2480` present in the PR body, outside any code span.
- No on-box acceptance owed (no hardware-provable behaviour); no config knob, no `cast.json` write, no derived artifact staled.

Issue #2179, opened 2026-08-06
ops — a Pinokio install ships every env-backed knob locked and read-only in Advanced Configuration


dudarenok-maker:
## What

On a Pinokio install, **every env-backed registry knob ships env-locked and read-only in Advanced Configuration** — the opposite of what registering a knob is supposed to buy.

The mechanism:

1. `pinokio-scripts/lib/write-env.js:26-29` copies `server/.env.example` **verbatim** into `server/.env`, rewriting only `WORKSPACE_DIR`.
2. `server/.env.example` is generator-emitted from the registry, and the generated blocks contain **active assignments** (`ASR_COMPUTE_TYPE=sidecar-default`, `ASR_CONCURRENCY=2`, …), not commented-out examples.
3. `process.loadEnvFile` therefore puts every one of them into `process.env`.
4. `server/src/config/resolver.ts:18-21` sees the value in the environment and returns `source: 'env', locked: true`.
5. `override-row.tsx` disables the control, shows "set in .env", and hides Revert — pinned by `override-row.test.tsx:40-78`.

Net effect: a Pinokio user who opens Advanced Configuration to change a setting finds it greyed out, and still has to hand-edit `server/.env` — which is precisely the situation the registry exists to eliminate.

## Scope

**Systemic and pre-existing** — this is not caused by any one knob's registration. It affects all env-backed knobs (≈110 of the 116 registered). Found during the independent review of PR #2176; that PR's benefit claim ("unreachable for anyone running a packaged install" → fixed) is overstated for the packaged install that actually exists, which is what surfaced it.

## Why it needs a decision

There is more than one defensible fix and they have quite different blast radii:

- **Emit the generated blocks commented-out.** `.env.example` becomes documentation rather than assignment, so nothing lands in `process.env` and every knob stays UI-settable. Cheapest conceptually, but `.env.example` is also what a developer copies to `server/.env` by hand, and some knobs may be relied on being present.
- **Make `write-env.js` strip (or comment out) generated assignments** when materialising `server/.env`, keeping only what the install genuinely needs to pin. Confines the change to the Pinokio path, but leaves `.env.example` itself misleading for everyone else.
- **Distinguish "set to the registry default" from "deliberately pinned"** in the resolver, so a value that merely equals the knob's own default does not count as an env lock. Most precise, most invasive, and needs care: an operator who deliberately pins a value equal to the default would silently lose the lock.

[...]

PR #2517, opened 2026-08-20
feat(server): surface a cleanup notice for leftover-default env locks (#2194)

### LINKED ISSUE #2194: ops — an existing server/.env keeps every knob locked after #2179, since upgrades never rewrite it
## What

#2179 fixes the "every env-backed knob ships locked" defect by emitting `server/.env.example`'s generated block commented-out, so nothing lands in `process.env` and Advanced Configuration stays live.

**That fix reaches new installs only.** `buildEnvContents` (`pinokio-scripts/lib/write-env.js:26-29`) returns `null` when `server/.env` already exists, so an upgrade never rewrites it. Every box that has already run Castwright keeps its ~110 active assignments in `server/.env` and continues to show every setting greyed out with "set in .env" — the exact symptom #2179 was filed about.

The same applies to anyone who followed `INSTALL.md` / `docs/wiki/Installing-Castwright.md` and copied `.env.example` to `.env` by hand before the fix.

## Why it is filed rather than folded into #2179

Migrating an existing `server/.env` means **editing a file the user owns**, and the central difficulty is that the file does not record intent:

- A line reading `ASR_COMPUTE_TYPE=sidecar-default` might be the shipped default nobody ever touched, or a deliberate pin that happens to equal the default. Commenting out the second case silently removes a lock the operator meant to have.
- Comparing against the registry default distinguishes "differs from default" (clearly deliberate — leave it) from "equals default" (probably inherited — but not certainly).
- The safe-but-annoying option is to touch nothing and tell the operator, which is only useful if the telling is good enough to act on.

Options, none obviously right:

1. **One-time migration on upgrade** — comment out generated-block lines that still equal the registry default, back up the original to `server/.env.bak` first. Fixes existing boxes automatically; risks unpinning a deliberate default-equal pin.
2. **Detect and surface, do not edit** — on startup, if many knobs resolve `source: 'env'` with values equal to their defaults, surface an operator-visible notice offering the cleanup as an explicit action. Never edits without consent; needs a UI surface and someone to act.
3. **Document only** — release-notes guidance telling upgraders to delete or comment their generated block. Zero risk, near-zero adoption.
4. **Do nothing** — accept that existing installs stay locked until the user rewrites `.env` themselves.

Option 2 composes well with the Advanced Configuration surface that already exists, and is the only one that neither edits silently nor relies on people reading release notes.

## Acceptance

- A decision recorded here with rejected alternatives.
- If the file is ever rewritten: a backup is taken first, and a test proves a deliberately-pinned value differing from the default is never touched.
- If a notice is surfaced: a test that it appears only when there is something to clean up, and does not nag once resolved.
- The chosen path is reflected in `RELEASE_NOTES.md` — this is a user-visible difference between a fresh install and an upgrade.

## Key files

- `pinokio-scripts/lib/write-env.js:26-29` — `buildEnvContents`, the returns-null-if-exists branch
- `server/src/config/resolver.ts` — `source: 'env'` / `locked: true` resolution
- `server/src/config/env-example.ts` — the emitter changed by #2179
- `INSTALL.md`, `docs/wiki/Installing-Castwright.md` — the manual copy instruction

## Benefit

*Benefit (user):* existing installs get the same working Advanced Configuration that #2179 gives new ones, instead of silently keeping ~110 read-only settings.

Found by the coordinator while scoping #2179. Refs #2179.

dudarenok-maker:
## Summary
- Detects env-backed config knobs whose `server/.env` value equals the
  shipped default (a leftover from a pre-#2179 install) and exposes them via
  `GET /api/config`.
- Adds an explicit, user-triggered `POST /api/config/env-cleanup` endpoint
  that backs up `server/.env` to `server/.env.bak` and comments out exactly
  those leftover lines — never a value that differs from its default.
- Advanced Configuration surfaces a one-time notice with a "Clean up" button
  when there's something to clean, and clears once resolved.

Decision + rejected alternatives recorded on #2194.

## Test plan
- [x] `npm run verify:fast:branch` green (all steps cached/passed, exit 0)
- [x] Mutation on the "never touch a differing value" guard observed red
      (dropped the file-value equality check in
      `envCleanupCandidateKeysFromFileContent`, re-ran `resolver.test.ts`,
      confirmed the ambient-shadowing regression test failed; reverted,
      working tree clean). An earlier, differently-shadowed version of this
      guard (`envCleanupCandidateKeys`, keyed off `process.env` instead of
      the file's own line value) was removed as dead code during review —
      see below.
- [x] Frontend component tests cover zero- and non-zero-candidate states

## Also fixed, found in passing (mandatory PR review gate, 2 rounds)

Two rounds of adversarial review surfaced 21 findings (11 in round 1, 10 in
round 2 — several of round 2's from the round-1 fix commits interacting with
each other), all fixed in this PR per this repo's report-fix-record rule:

- An arbitrary-file-read/write vulnerability (`?envPath` accepted from an
  unauthenticated HTTP request) — removed entirely, replaced with a
  non-HTTP-reachable test seam.
- The env-cleanup notice could never clear without a server restart, and a
  second cleanup click destroyed the one-and-only `.env.bak` backup.
- `EXDEV` cross-filesystem rename failure on any install where the OS tmpdir
  and `server/.env` live on different filesystems (e.g. stock Linux `/tmp`).
- The "never touch a differing value" invariant broke under ambient shell-env
  shadowing (candidacy is now keyed off the `.env` file's own line value,
  not `process.env`).
- `server/.env.bak` (and the transform's own `.env.tmp-*` temp files) were
  not gitignored — a secrets-bearing file could be accidentally staged.
- POSIX file-mode widening (0600 → 0644) on both the rewritten `.env` and
  its backup.
- The widened line-matching regex (added to support indented/`export`-
  prefixed lines) could match across a newline.
- Several test-hygiene issues: a `process.env` leak on a red test path, two
  tautological "mutation" tests providing no real coverage, and no coverage
  of the production default-path-resolution branch or of the transform at
  realistic (~110-line) `.env` scale.

Full review threads: [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2517#issuecomment-5351834611),
[pass 2](https://github.com/dudarenok-maker/Castwright/pull/2517#issuecomment-5353229223).

Closes #2194


dudarenok-maker:
## PR review — pass 1 (head 93012720, depth high)

Scope: full branch diff vs `origin/main` (20 files) — `server/src/config/env-cleanup.ts`, `server/src/config/resolver.ts`, `server/src/routes/config.ts`, `src/components/env-cleanup-notice.tsx`, `src/views/advanced.tsx`, `src/store/config-slice.ts`, `src/lib/{api,types}.ts`, and all six touched test files · verified green locally before probing: server `env-cleanup` + `resolver` + `config` + `config.env-cleanup` = **105 tests / 4 files**, frontend `env-cleanup-notice` + `advanced` + `config-slice` = **45 tests / 3 files**, `npm run typecheck` clean · then adversarially probed the "never touch a differing value" invariant, the backup/comment-out file mutation, and the notice-clears claim.

Depth `high` is right: this is multi-scope, and the server half rewrites a file the user owns that holds their secrets.

---

### 🔴 Blocking — the notice can never clear without a server restart, and the second click destroys the backup

`envCleanupCandidateKeys()` (`server/src/config/resolver.ts:151-159`) derives the list purely from `process.env`. The POST handler (`server/src/routes/config.ts:207-263`) mutates the **file** only — nothing anywhere writes or deletes a `process.env` key, and `process.loadEnvFile('.env')` runs exactly once, at module-eval of `server/src/load-env.ts:35` (verified: only production call site; no `.env` watcher exists). So after a successful cleanup, `GET /api/config` returns the **identical** `envCleanupCandidates`.

Repro:
1. `server/.env` contains `OLLAMA_TEMPERATURE=0.2` (the registry default). Start the server.
2. Advanced Configuration shows "1 setting looks like leftover defaults from an older install."
3. Click **Clean up** → `.env` line is commented out, `.env.bak` written with the original. `src/views/advanced.tsx:326-333` refetches.
4. The refetch returns `envCleanupCandidates: ['analyzer.ollama.temperature']` again — `process.env.OLLAMA_TEMPERATURE` is still `'0.2'`. **The notice is still there, same count.**
5. The user clicks again. `config.ts:243` `writeFileSync(bakPath, original, 'utf-8')` runs **unconditionally**, and `original` is now the already-cleaned text. `server/.env.bak` no longer contains the user's original values.

The only safety net the feature advertises is gone on the second click, and the UI actively invites that click by not clearing. Nothing tells the user a restart is required either, and `selectRestartServerPending` — already wired into this very view — is not set.

Neither test catches it, and both walk right past it:

- `server/src/routes/config.env-cleanup.test.ts:103` ("is idempotent: second call returns `cleaned:[]`") makes the second POST and asserts only the response body. Note *why* it returns `[]` — because the file has nothing left to comment, **not** because the candidate set shrank. `process.env.OLLAMA_TEMPERATURE` is `'0.2'` for the whole test. The backup content after call 2 is never asserted.
- `src/views/advanced.test.tsx:604` ("drops the notice once the refetch reports it resolved") hard-codes `mockGetConfig.mockResolvedValueOnce({... envCleanupCandidates: [] })` for the second fetch. It asserts the mock's script, not the server's behaviour — reviewer-brief shape #5, an acceptance criterion that would pass on a null observation.

Four prose claims are false as shipped: `src/components/env-cleanup-notice.tsx:6` ("so a fresh `fetchConfig()` naturally clears it"), `src/lib/types.ts:937-941` ("clears this list on the next fetch"), `docs/release-notes-next.md` ("refetches config, which naturally drops the candidate list"), and `RELEASE_NOTES.md` ("The prompt disappears on its own once everything's tidy").

---

### 🔴 Blocking — `envPath` is a caller-controlled arbitrary path: a test seam shipped into production

`server/src/routes/config.ts:210-211`:

```ts
const queryPath = typeof req.query.envPath === 'string' ? req.query.envPath : undefined;
const envPath = queryPath ?? resolveServerEnvPath();
```

No allowlist, no containment check, no `NODE_ENV` gate. The comment above it says so outright: "Allow tests to inject a path". `server/src/app.ts:123` applies `requireLanToken` to `/api` only when LAN HTTPS mode is on **and** `LAN_AUTH_TOKEN` is set, and **loopback always bypasses** — so on a default local install this endpoint is unauthenticated.

The handler then performs two writes with that path: `writeFileSync(`${envPath}.bak`, original)` (line 243) and a rename over `envPath` (line 256).

- Repro A (destroys an unrelated backup, works on *any* file): `POST /api/config/env-cleanup?envPath=C:/Users/x/Documents/thesis.txt` → `thesis.txt.bak` is overwritten with the current contents of `thesis.txt`, and `thesis.txt` itself is rewritten from a temp file. No `KEY=VALUE` line needs to match for either write to happen.
- Repro B (silently edits another project's config): `?envPath=C:/other-project/.env` containing `ASR_DEVICE=cpu` — this server's shipped default — comments that line out in a file this app has no business touching.
- Cross-origin amplification: `fetch('http://localhost:8080/api/config/env-cleanup?envPath=…', {method:'POST'})` from any page the user visits is a CORS *simple request* — no preflight, no custom header — so the write lands even though the attacker cannot read the response.

Separately, `config.ts:200-202`'s comment is incoherent: "Exported so tests can override via the `envPath` query param without monkey-patching `process.cwd`." Exporting `resolveServerEnvPath` has nothing to do with the query param, and the test file never imports it.

---

### 🔴 Blocking — cleanup fails on every Linux install with a tmpfs `/tmp`, after the backup is already clobbered

`server/src/routes/config.ts:253-256` writes the replacement into `mkdtempSync(join(tmpdir(), 'env-cleanup-'))` and then `renameWithRetry(tmpFile, envPath)`.

`rename(2)` returns `EXDEV` across filesystems. `server/src/workspace/atomic-rename.ts:37-43` — `RETRYABLE_CODES` is `{EPERM, EBUSY, ENOENT, EACCES, EIO}`; `EXDEV` is absent and there is no copy fallback, so line 53 rethrows on the first attempt.

Repro: any systemd distro (`/tmp` is tmpfs by default) with the app under `/opt` or `$HOME` → every POST returns `500 failed to write cleaned .env: EXDEV: cross-device link not permitted`. Same on Windows when `%TEMP%` is on `C:` and the install on another drive, and on macOS when `TMPDIR` is a different volume. The repo gates releases on Ubuntu and macOS as well as Windows, so this is not a hypothetical platform.

The `.bak` write at line 243 has already happened by the time this throws, so a user clicking repeatedly (which finding 1 guarantees they will) keeps clobbering the backup while the cleanup never once succeeds.

The comment immediately above is false about the exact property that would have made this work — `config.ts:249-251`: "write cleaned text to a temp file **in the same directory**, then rename over the target. Same rename helper as state-io's atomic JSON writes." `writeJsonAtomic` (`server/src/workspace/state-io.ts:122`) writes `${path}.tmp-<pid>-<ts>-<seq>-<rnd>` **beside the target**, and unlinks it on failure (line 129). This code does neither. Reviewer-brief shape #9 — the defect is in the instrument's description as much as in the code.

---

### 🟠 Significant — `server/.env.bak` is not git-ignored, and it is a verbatim copy of a secrets file

Verified: `git check-ignore -v server/.env server/.env.bak` matches **only** `server/.env` (`.gitignore:20`). `.gitignore` carries `.env`, `.env.local`, `.env.*.local`, `!.env.example` — none of which matches `.env.bak` — and there is no `server/.gitignore`.

`server/.env` holds `GEMINI_API_KEY` and `LAN_AUTH_TOKEN`. After one click the working tree has an untracked, *unignored* `server/.env.bak` containing both; it shows in `git status` and a `git add -A` / `git add .` stages it into a public repo. CLAUDE.md already names this exact leak class for worktrees ("that would leak secrets like `GEMINI_API_KEY` … (#2345)"). Add `.env.bak` (or `.env.bak*`) to `.gitignore` in this PR.

---

### 🟠 Significant — the "never touch a differing value" invariant breaks when the env var is also set in the ambient environment

Verified empirically on this box (node v24.15.0): `process.loadEnvFile` does **not** overwrite an already-set `process.env` key.

```
$ node -e "process.env.FOO='fromshell'; process.loadEnvFile('…/envtest.env'); console.log(process.env.FOO)"
FOO after load = "fromshell"     # file said FOO=fromfile
```

So `process.env[knob.env]` can be the *ambient* value while `cleanEnvText` mutates the *file* line — and the predicate is keyed on the env-var **name**, never on the line's own text (`config.ts:220-227` builds a `Set` of names; `env-cleanup.ts:39-45` matches on `varName` alone).

Repro: `server/.env` contains `GEN_WORKERS=4` — a deliberate pin, different from the registry default. The user also exports `GEN_WORKERS=<default>` from their shell / systemd unit / Pinokio env. Server starts → `process.env.GEN_WORKERS` is the shell's default → `envCleanupCandidateKeys()` reports it as a leftover → POST comments out `GEN_WORKERS=4`. Drop the shell export later and the pin is gone, silently and permanently — with the backup one click from being clobbered per finding 1.

The fix is cheap and closes the class rather than the instance: compare the **line's own value** against the knob default inside `cleanEnvText`, instead of trusting a name-keyed predicate derived from a different source. That also makes the transform correct independent of how `process.env` was assembled.

---

### 🟠 Significant — the shipped production path has zero coverage, and no on-box row records the gap

All four cases in `server/src/routes/config.env-cleanup.test.ts` pass `?envPath=`. `resolveServerEnvPath()` is exercised by nothing at all, and "the notice actually clears" — the feature's headline behaviour, which finding 1 disproves — is exercised only against a scripted mock.

No `docs/testing/onbox-acceptance-register.md` row, no run sheet, no live-view edit. The register's **Group H** precedent ("No hardware — needs a real … not yet in this repo's corpus", 2 rows) is exactly this shape: the missing prerequisite here is a real, pre-#2179 `server/.env` on a real install. Per CLAUDE.md before-shipping step 3, *recording* the debt is a merge gate even when running it is not — and this is the row that would have caught finding 1.

---

### 🟠 Significant — permission widening on POSIX for a file holding secrets

`writeFileSync(tmpFile, result.text, 'utf-8')` (`config.ts:255`) creates the replacement with mode `0666 & ~umask` — `0644` under the default `umask 022`. `rename` preserves the source file's mode, so a `server/.env` the user had at `0600` comes back **world-readable**, carrying `GEMINI_API_KEY` and `LAN_AUTH_TOKEN`.

`writeJsonAtomic`, the precedent the comment cites, writes workspace JSON rather than a secrets file, so the precedent does not carry across. Either `stat` the original and re-apply its mode, or create the temp file with an explicit `{ mode: 0o600 }`.

---

### 🟡 Minor

1. **The temp directory is leaked on every call.** `mkdtempSync` (`config.ts:253`) creates a directory per POST; the file inside is renamed away and the directory is never removed, on the success path or the failure path. `state-io.ts:129` at least unlinks its temp on failure.
2. **`it('GOES RED if isCandidate is always true (mutation test)')` is not a mutation test.** `server/src/config/env-cleanup.test.ts:82-97` calls `cleanEnvText` twice with two predicates the test supplies itself; no production code is mutated. Its closing assertion `expect(mutated.text).not.toBe(text)` is a tautology given the line above it, and the preceding test at line 71 already covers the same ground. A test that advertises a proof it does not deliver is worse than no test (reviewer-brief shape #3) — rename it or drop it.
3. **`process.env` leaks between tests on a red path.** `server/src/routes/config.env-cleanup.test.ts` sets `process.env.OLLAMA_TEMPERATURE` inside three test bodies and `delete`s it as the last statement of each. Any earlier assertion failure skips the delete, and the next test in the file then runs with a stale value — one red becomes a cascade whose cause is somewhere else. Move it to `afterEach`.
4. **`cleanEnvText` will not match an indented or `export`-prefixed assignment.** `ENV_LINE_RE = /^([A-Z0-9_]+)=(.*)$/gm` (`env-cleanup.ts:18`). Node's parser accepts leading whitespace, so `  ASR_DEVICE=cpu` is loaded into `process.env`, reported as a candidate, and then never commented out — `cleaned` omits it, and per finding 1 the notice keeps naming it forever. It fails *closed* on the file, which is the right direction, but it is the "guard that enumerates syntax" shape (#2) and it compounds finding 1.
5. **No `openapi.yaml` entry** for `POST /api/config/env-cleanup` or the new `envCleanupCandidates` field. Flagged for completeness only, and I am *not* asking for it here: `openapi.yaml` documents no `/api/config` path at all today and `ConfigResponse` is hand-written in `src/lib/types.ts`, so this PR is consistent with the existing config router rather than regressing it.

---

### ✅ What is solid

- `envCleanupCandidateKeys()` correctly skips `isPrompt` and env-less knobs, and compares the **coerced** effective value against the default — so `PRELOAD_KOKORO=off` vs `false` and `STAGE2_MIN_COVERAGE=0.60` vs `0.6` read as the same value rather than as deliberate pins. Genuinely covered by the four new `resolver.test.ts` cases.
- `cleanEnvText` really does leave already-commented lines, blanks, section headers, lowercase keys and non-matching lines byte-for-byte intact, and preserves `=` inside values plus a missing trailing newline. The unit coverage there is good.
- `EnvCleanupNotice` genuinely mirrors `restart-sidecar-banner.tsx` as its comment claims, including the `min-h-[44px] fine-pointer:min-h-0` touch target — checked line by line.
- `Closes #2194` present, plain, outside any code span. Release-notes pair both landed (`docs/release-notes-next.md` + `RELEASE_NOTES.md`), and the technical entry is accurate apart from the "naturally drops the candidate list" clause called out in finding 1.
- No new config knob, so no registry/`config:sync`/Settings-row/`.env.example` obligation. No `cast.json` write, so the lock rules do not apply.

---

### Verdict

**Request changes.** Three blocking correctness bugs, three significant ones.

The central problem is that the candidate set is computed from `process.env` while the mutation is applied to the file, and the two are not the same thing — that single mismatch produces finding 1 (the notice cannot clear, so users click again and lose their backup) and finding 5 (a deliberate pin gets commented out when the ambient environment shadows it). Deciding candidacy from the *line's own value* fixes both at the source rather than one site at a time.

Findings 2 and 3 are independent and each is enough on its own: the `envPath` query param turns a config endpoint into an unauthenticated arbitrary-file-rewrite primitive, and the `os.tmpdir()` rename cannot succeed on a stock Linux install. Both sit next to comments that describe code that is not there.

Per the gate's re-review trigger, fixing and pushing these re-triggers a pass.

dudarenok-maker:
## PR review — pass 2 (head 6bf8d9dc, depth high)

Scope: full re-read of `main...6bf8d9dc` (22 files) plus the 12-commit delta since pass 1 (`93012720..6bf8d9dc`), read commit-by-commit for fix-chain interference · re-ran the four touched server suites locally: `env-cleanup` + `resolver` + `config.env-cleanup` + `config` = **126 tests / 4 files, all green** (vitest 4.1.9) · probed the shipped `ENV_LINE_RE`, `chmodSync` with `statSync().mode`, and Node's `.env` parser empirically on node v24.15.0 · ran `computeScopes()` from `scripts/ci-scope.mjs` against a synthetic `.gitignore`-only diff.

**All three 🔴 Blocking findings are genuinely closed, and the tests behind them are real tests.** Two Significant issues remain — one is the unfixed half of finding 7, one is a guard the fix chain added that never runs — plus one Significant cleanup and six minors, several of which are the fix chain stepping on its own earlier work.

---

## Disposition of pass 1's eleven findings

|---|---|---|
| 1 | 🔴 Notice can never clear; second click destroys the backup | **VERIFIED RESOLVED** (both parts) |
| 2 | 🔴 `?envPath` arbitrary-file read/write | **VERIFIED RESOLVED** |
| 3 | 🔴 `EXDEV` on a tmpfs `/tmp` | **VERIFIED RESOLVED** |
| 4 | 🟠 `server/.env.bak` not git-ignored | **VERIFIED RESOLVED** — new guard is stale-green-prone, see **N2** |
| 5 | 🟠 Ambient-env shadowing defeats "never touch a differing value" | **VERIFIED RESOLVED** — its claimed integration test is a tautology, see **N3** |
| 6 | 🟠 Production path uncovered / no on-box row | **RESOLVED with a residual** — default-path branch now covered; the register call is right; the test it converts into does not exist, see **N10** |
| 7 | 🟠 POSIX permission widening on a secrets file | **PARTIALLY RESOLVED — one of the two writes is still 0644**, see **N1** |
| 8 | 🟡 Temp directory leaked every call | **VERIFIED RESOLVED** — narrower residual, see **N8** |
| 9 | 🟡 Fake "mutation test" | **VERIFIED RESOLVED** — an equivalent was added in the same chain, see **N3** |
| 10 | 🟡 `process.env` leaks between tests | **VERIFIED RESOLVED** in that file — a new one appeared in another, see **N7** |
| 11 | 🟡 Indented / `export`-prefixed lines unmatched | **VERIFIED RESOLVED** — with a `\s` wrinkle, see **N4** |

### What I actually checked, not just read the commit message for

**Finding 1 — resolved, and the mechanism is right.** `GET /` (`server/src/routes/config.ts:65-77`) now reads `server/.env` and calls `envCleanupCandidateKeysFromFileContent(fileContent)`. After a cleanup the lines are `# KEY=value`, `parseEnvFileLines` no longer matches them, so the candidate list genuinely shrinks in the *same process* with no restart. `config.env-cleanup.test.ts:104-124` (Part A) proves it end-to-end through supertest — before/POST/after in one process — and goes red if the GET handler is reverted to `process.env`. Part B (`:126-174`) is the backup half: the early return at `config.ts:273-276` fires before the `.env.bak` write, and the test asserts the second call's `.bak` bytes are identical to the first's. Both are real; neither is scripted.

**Finding 2 — resolved properly, not papered over.** `req.query` is not read anywhere in either handler; both signatures are `async (_req, res)`. The replacement seam `_setServerEnvPathForTest` (`config.ts:235`) is module-level state with no HTTP reachability, and `git grep` confirms its only caller is `config.env-cleanup.test.ts:62`. Parts C and D (`:176-249`) are honest regression tests: if `req.query.envPath` were reintroduced, C's `toContain('analyzer.stage2.minCoverage')` and D's `toContain('STAGE2_MIN_COVERAGE')` both fail, because the attacker fixtures contain nothing that could produce those. The incoherent comment pass 1 flagged is gone, and its replacement is accurate — `load-env.ts:35` really does call `process.loadEnvFile('.env')` on a cwd-relative path, which is what `resolveServerEnvPath()` mirrors.

**Finding 3 — resolved.** `mkdtempSync(tmpdir())` is gone; the temp is `${envPath}.tmp-<pid>-<ts>-<seq>-<rnd>` beside the target (`config.ts:305-307`), matching `state-io.ts`'s shape, and the comment at `:288-292` now describes code that exists. `unlink` on rename failure added at `:318`.

**Finding 5 — resolved, and the regression test is genuine.** `resolver.test.ts:391-407` sets `process.env.STAGE2_MIN_COVERAGE='0.6'` (the default) against a file saying `0.7` and asserts the key is *not* a candidate. Under the pre-`d17e55e3` implementation that returns the key, so it goes red for exactly the claimed reason. I confirmed the premise empirically on node v24.15.0: `process.loadEnvFile` does not overwrite an already-set key (`PROBE_B` stayed `"ambient"` against a file saying `plain`).

**Finding 11 — resolved.** `ENV_LINE_RE` now carries indent + `export` groups and echoes them back on comment-out; 8 new unit cases cover space/tab indent and `export` in both positions. CRLF still round-trips (`'FOO=bar\r\nBAZ=qux\r\n'` → `'# FOO=bar\r\n# BAZ=qux\r\n'`, verified). Inline comments and quoted values still fail *closed* — they're captured verbatim into the value, fail `coerceAndValidate`, and are therefore never candidates in GET *or* POST, so the two stay consistent and nothing gets stuck. That's the right direction and I'm not asking for more.

---

## New findings

### 🟠 Significant — correctness — finding 7 fixed one of the two writes; `.env.bak` is still created world-readable with the same secrets

`server/src/routes/config.ts:280-286`:

```ts
const bakPath = `${envPath}.bak`;
writeFileSync(bakPath, original, 'utf-8');
```

`original` is a byte-for-byte copy of `server/.env` — `GEMINI_API_KEY`, `LAN_AUTH_TOKEN`, the lot. `writeFileSync`'s default mode is `0o666`, so under the default `umask 022` this creates `server/.env.bak` at **0644**. The mode-preservation `bc4f68d1` added (`config.ts:296-312`) applies only to the temp file that becomes the new `.env` — and note it reads `statSync(envPath).mode` *after* the backup has already been written, so the value isn't even in scope at the backup site yet.

Repro on any POSIX box:

```
chmod 600 server/.env
curl -X POST localhost:8080/api/config/env-cleanup
stat -c %a server/.env       # 600  ✅ finding 7 fixed here
stat -c %a server/.env.bak   # 644  ❌ same secrets, world-readable
```

This is reviewer-brief shape #10 — one instance fixed, the class left armed. The restored regression test (`config.env-cleanup.test.ts:300-328`) asserts the mode of the *target* only, so it stays green over this; it needs a second assertion on `${modeTestPath}.bak`.

Fix: hoist the `statSync` above the backup write and pass `{ encoding: 'utf-8', mode: originalMode & 0o777 }`. Note `writeFileSync`'s `mode` applies **only at creation** — a `.env.bak` left at 0644 by an earlier run keeps 0644 on the next write, so the second and later cleanups need an explicit `chmodSync(bakPath, …)` as well.

### 🟠 Significant — correctness — the new `.gitignore` guard never runs on a `.gitignore`-only diff

`scripts/tests/gitignore-secrets.test.mjs` (new, from `bf0073cd`) reads `.gitignore` at **runtime**, via `spawnSync('git', ['check-ignore', …])`. There is no module-graph edge, so it has to be hand-registered as an input to `verify-cache.mjs`'s `test:hooks` step. It is not: that step's inputs (`scripts/verify-cache.mjs:76-300`) are `scripts/**/*.{mjs,cjs,js,mts,cts,ts}`, `scripts/tests/fixtures/**`, `pinokio-scripts/**`, `.github/workflows/**`, `.github/actions/**`, `.husky/**`, `.claude/skills/**`, `.claude/agents/**`, `docs/testing/**` and ~20 named files. `.gitignore` appears in none of them — `grep -n gitignore scripts/verify-cache.mjs scripts/ci-scope.mjs` returns three unrelated comment lines and nothing else.

Measured, not reasoned — running the repo's own scope resolver:

```
$ node -e "import('./scripts/ci-scope.mjs').then(m => console.log(
    m.computeScopes(['.gitignore'], {eventName:'pull_request'}).step_test_hooks,
    m.computeScopes(['scripts/tests/gitignore-secrets.test.mjs'], {eventName:'pull_request'}).step_test_hooks))"
false true
```

So: delete the `.env.bak` line from `.gitignore` and commit nothing else — `npm run verify:fast:branch` prints `test:hooks [cached]`, and because `ci-scope.mjs` derives its legs from the same `STEPS[]` entry (its own header says so), cloud `verify.yml` schedules no hooks leg either. The guard added in this PR to catch exactly that regression certifies nothing, on precisely the diff that breaks it. `verify-cache.mjs`'s own comments document this trap (#1847) ten separate times — `fixtures/**`, `pinokio-scripts/**`, `.claude/agents/**`, `CLAUDE.md`, `CONTRIBUTING.md`, `.env.mock`, `.env.development` are each hand-registered for this reason and each says why.

Fix: add `'.gitignore'` to that step's `files` list, with the same one-line rationale the neighbours carry.

### 🟠 Significant — cleanup — `envCleanupCandidateKeys()` is now dead code that still ships finding 5's bug, plus four tests certifying it

`server/src/config/resolver.ts:146-159` — the `process.env`-keyed predicate — is `export`ed and referenced by **nothing but its own tests**:

```
$ git grep -n 'envCleanupCandidateKeys\b' -- '*.ts'
server/src/config/resolver.test.ts:17,360,368,374,379,384
server/src/config/resolver.ts:152          # the definition
```

`d17e55e3` replaced it with `envCleanupCandidateKeysFromFileContent` at both call sites and left the original standing. Both were introduced *and* orphaned inside this PR, which is squarely CLAUDE.md's "Remove what YOUR changes orphaned". Two costs:

1. The buggy predicate sits one import away from the correct one, with a docstring that still reads like the live description of the feature ("Knob keys whose env-sourced value is INDISTINGUISHABLE from having no override at all… ~110 of these can exist at once"). The next caller who reaches for the shorter name reopens finding 5 in full: a deliberately-pinned `GEN_WORKERS=4` reported as a leftover default whenever the shell exports the default.
2. `resolver.test.ts:360-388` — four tests — read as live coverage of the shipped feature and cover zero reachable code. That is a false coverage signal in the same file as the real one.

Same shape, smaller: `resolveServerEnvPath` (`config.ts:225`) is still `export`ed but, now that the query param is gone, is referenced only inside `config.ts`.

No wrong output today, hence cleanup rather than correctness — but it is the trap the next round walks into.

### 🟡 Minor — correctness — the widened regex's `\s` crosses newlines, and `cleaned` then reports a key it did not comment out

`server/src/config/env-cleanup.ts:26`:

```ts
export const ENV_LINE_RE = /^(\s*)((?:export\s+)?)([A-Z0-9_]+)=(.*)$/gm;
```

`\s` matches `\n`. Running the shipped regex through the shipped `cleanEnvText` (node v24.15.0):

```
input  : "export\nFOO=bar\n"
output : "# export\nFOO=bar\n"
cleaned: ["FOO"]
```

The `#` lands on the stray `export` line; `FOO=bar` survives as a live assignment while the response tells the UI `FOO` was cleaned. GET then recomputes from the file, still sees `FOO=bar`, and the notice does not clear — so the user clicks again, `result.cleaned.length` is 1 again, the early return at `config.ts:273` does not fire, and `.env.bak` is overwritten with the already-mutated text. That is exactly the backup loss finding 1 part B was fixed to prevent, reachable through a different door.

It needs a bare `export` token on its own line, so likelihood is low — but the fix is one character class in each of two places (`[ \t]*` and `export[ \t]+`), and there is no reason for a line-oriented parser to be able to span lines. The blank-line and CRLF cases I also probed are fine (the indent is echoed back verbatim), so this is the only reachable shape.

### 🟡 Minor — cleanup — finding 9's fake proof was deleted and an equivalent one added in the same chain

`4ac14e4e` removed the tautological "mutation test", as asked. `d17e55e3` then added `server/src/config/env-cleanup.test.ts:213-229`:

```ts
describe('env-cleanup integration: file-based candidacy (#2194 finding 5)', () => {
  it('does not comment out a line whose file value differs from default, even if process.env is shadowed', () => {
    const fileContent = 'GEN_WORKERS=4\nOTHER_VAR=1\n';
    const isCandidate = (varName: string) => varName === 'OTHER_VAR';
    ...
```

It never calls `envCleanupCandidateKeysFromFileContent`, never sets `process.env`, and supplies the predicate itself — so it would stay green under a complete revert of `d17e55e3`, the commit it is named after. Its own comment concedes this: *"The candidacy check would be done against the file, not process.env. For this test, we'll show that a non-candidate line is left untouched."* That is already covered, better, at `:99-108` (`LEAVES a non-default value untouched even if the var name matches`). Reviewer-brief shape #3 — a test that advertises a proof it does not deliver. The real coverage for finding 5 is `resolver.test.ts:391-407` and it is fine; this one should be deleted, not renamed.

### 🟡 Minor — cleanup — the PR body's mutation proof now targets dead code

PR body, Test plan:

> Mutation on the "never touch a differing value" guard observed red (dropped the `state.effective === k.default` equality check in `envCleanupCandidateKeys`, re-ran `resolver.test.ts`, confirmed `envCleanupCandidateKeys > does not return key when env differs from default` failed)

`envCleanupCandidateKeys` has not been on any request path since `d17e55e3` (see the dead-code finding above). The PR's stated proof of its own load-bearing invariant mutates a function no HTTP request reaches, and observes a test that covers nothing go red. Re-run it against `envCleanupCandidateKeysFromFileContent` and cite the assertion that actually guards the shipped path.

While there: the `## Summary` still describes only the original three commits. Nothing in the body mentions the `.gitignore` change, the new `scripts/tests/gitignore-secrets.test.mjs`, the removal of the `?envPath` parameter, or the mode preservation — all of which are in the diff a reviewer opens. CLAUDE.md requires fixes folded in during a round to be declared in the PR body.

### 🟡 Minor — cleanup — a new `process.env` leak in `resolver.test.ts`, the same shape finding 10 removed from the other file

`b6459dce` correctly moved `config.env-cleanup.test.ts`'s cleanup into `beforeEach`/`afterEach`. But `resolver.test.ts:401` (added by `d17e55e3`) sets `process.env.STAGE2_MIN_COVERAGE = '0.6'` inside a test body, in a `describe` with no `beforeEach`/`afterEach`, and never deletes it. Benign *only* because it is currently the last `describe` in the file — the `envCleanupCandidateKeys` block immediately above it reads that exact variable and would start seeing a stale `'0.6'` the moment anything is appended below or the describes are reordered. Give it the same `afterEach` the other file just got.

### 🟡 Minor — cleanup — `server/.env.tmp-*` is not git-ignored, and a failed `writeFileSync` leaves one behind

Finding 3's fix moved the temp file *into the repository working tree*: `config.ts:307` writes `${envPath}.tmp-<pid>-<ts>-<seq>-<rnd>` — a full copy of `server/.env`, secrets included — beside the target. `.gitignore:20-21` covers `.env` and (now) `.env.bak`; it does not cover that suffix:

```
$ git check-ignore -v server/.env.bak server/.env.tmp-1234-1700000000-1-abcd1234
.gitignore:21:.env.bak	server/.env.bak
                        # ← no match for the .tmp- path
```

And the `unlink` at `config.ts:318` only covers a `renameWithRetry` failure. A `writeFileSync` or `chmodSync` throw at `:308`/`:311` falls straight to the outer `catch` at `:321` with no unlink — as does the Windows AV/cloud-sync `EPERM` case `renameWithRetry` exists for, when the follow-up unlink itself fails (swallowed by `.catch(() => {})`). Either way the tree gains an untracked, *unignored* secrets copy that `git add -A` stages. This is finding 4's leak with a different suffix. Widen the ignore to cover `.env.bak*` and `.env.tmp-*` (or `.env.*` with the existing `!.env.example` negation) and unlink in the outer catch.

### 🟡 Minor — cleanup — stale comment: the predicate is no longer derived from `process.env`

`server/src/config/env-cleanup.ts:53-56`:

```
@param isCandidate  Predicate: true when the env-var name is a leftover-
                    default candidate at write time (re-derived from live
                    process.env + registry default — never a cached GET).
```

Since `d17e55e3` it is re-derived from the **file**, never from `process.env` — that inversion is the entire content of finding 5's fix. Shape #9: the change made its own instrument's description false. (`config.ts:255-258` and `resolver.ts:163-170` both got this right; only this one was missed.)

### 🟡 Minor — cleanup — nothing exercises the transform against the input the feature exists for

`ff12bc14` concluded no on-box register row is owed, and **I agree** — this is pure filesystem I/O with no hardware, sidecar, GPU or real-book prerequisite, so CLAUDE.md before-shipping step 3 does not bite and Group H's precedent does not carry. But that conclusion converts the debt into an ordinary test obligation, and that test does not exist. Every server fixture in this PR is a 2–5 line synthetic `.env` with one candidate (`STAGE2_MIN_COVERAGE=0.6`). The actual target input is the ~110-line `server/.env` a pre-#2179 install copied verbatim from `.env.example` — the file the release note is written about.

It is cheap and fully deterministic: take the tracked `server/.env.example`, uncomment its generated block (that *is* the pre-#2179 shape, per #2179's own release note), and assert (a) `envCleanupCandidateKeysFromFileContent` returns the full env-backed knob set, and (b) `cleanEnvText` leaves every section header, help comment, blank line and preamble assignment byte-identical. As it stands, nothing has run this transform over a real file's structure at scale even once.

---

## What is solid this round

- The candidate/mutation mismatch that produced findings 1 and 5 is closed **at the source**, exactly as pass 1 asked — candidacy is decided from the line's own value, so both symptoms die together rather than being patched one site at a time.
- `6bf8d9dc`'s recovery of the mode-preservation test that `cfad1c13` silently deleted is the right instinct and the right catch. I looked for further casualties of the same shape across all twelve commits — dropped assertions, silently reverted hunks — and found none: every test present at `93012720` is either still present or was removed deliberately with a named finding.
- The `_setServerEnvPathForTest` seam is the right replacement for the query param: not reachable over HTTP, single caller, restored in both `beforeEach` and `afterEach`, and nulled in `afterAll`.
- The `default env-path resolution` test (`config.env-cleanup.test.ts:251-298`) is careful in a way that matters — it `chdir`s *before* clearing the override, so a mis-ordering could never point the POST at the developer's real `server/.env`, and the restore is unconditional in `afterEach`.
- `Closes #2194` present, plain, outside any code span. Release-notes pair both landed, and the four prose claims pass 1 called false are now all true as shipped.
- 126/126 green locally across the four touched suites; `npm run typecheck` clean.

---

## Verdict

**Request changes** — but the gap is much smaller than pass 1's. Nothing blocking remains; the three 🔴 findings are properly fixed and their tests are real.

Two things genuinely need to land before merge. `.env.bak` is still written 0644 with the user's API key in it, which means finding 7 is half-fixed and the test that was restored to prove it is blind to the half that isn't. And the `.gitignore` guard this round added is unreachable on the only diff that can break what it guards — measured, not inferred. The dead `envCleanupCandidateKeys` should go with them: it is the buggy predicate finding 5 was filed against, still exported, still carrying four tests that look like coverage.

The minors are mostly the fix chain marking its own homework — a tautological test added in the same round one was removed for being tautological, a `process.env` leak reintroduced in a second file in the same round it was fixed in the first, a docstring inverted by the fix it documents, and a PR body whose mutation proof now targets code no request reaches. None of those is expensive; all of them are the kind of thing that reads as noise to the next reviewer.

Correctness bugs present (`.env.bak` mode, the unreachable guard, the newline-crossing `\s`), so fixing and pushing re-triggers a pass.

dudarenok-maker:
## PR review — pass 3 (head f35217f8, depth high)

Scope: full re-read of `main...f35217f8` (24 files) plus the 8-commit delta since pass 2 (`6bf8d9dc..f35217f8`), re-derived from the **current file state** rather than commit messages (this round's history was reset and re-split mid-flight) · re-ran the four touched server suites: **125 tests / 4 files green** (vitest 4.1.9) and the full `npm run test:hooks` battery: **1422 tests / 27 suites, 0 fail** · measured rather than reasoned: `computeScopes(['.gitignore'])` through the repo's own `ci-scope.mjs`, `git check-ignore -v` on four candidate paths, Node's `.env` duplicate-key resolution, `chmodSync` with file-type mode bits, and the old-vs-new `ENV_LINE_RE` run side by side on the N4 input.

**Nine of pass 2's ten findings are closed. N9 is not — the commit credited with it fixed a different comment.** Everything still open is documentation or test hygiene; there is no blocking or significant correctness defect left in this branch.

---

## Disposition of pass 2's ten findings

|---|---|---|
| N1 | 🟠 `.env.bak` written 0644 with the same secrets | **VERIFIED RESOLVED** — residual window, see M2 |
| N2 | 🟠 `.gitignore` guard never runs on a `.gitignore`-only diff | **VERIFIED RESOLVED** (measured) |
| N3 | 🟠 dead `envCleanupCandidateKeys()` + 4 tests certifying it | **VERIFIED RESOLVED** — one parenthetical half left, see M6 |
| N4 | 🟡 widened regex's `\s` crosses newlines | **VERIFIED RESOLVED** — real regression test |
| N5 | 🟡 tautological "finding 5" test added in the same chain | **PARTIALLY RESOLVED** — tautology gone, misattribution intact |
| N6 | 🟡 PR body's mutation proof targets dead code | **VERIFIED RESOLVED** |
| N7 | 🟡 new `process.env` leak in `resolver.test.ts` | **VERIFIED RESOLVED** |
| N8 | 🟡 `.env.tmp-*` unignored + no unlink on write failure | **VERIFIED RESOLVED** — three residuals, see M3/M4/M5 |
| N9 | 🟡 stale docstring: predicate "re-derived from live process.env" | **STILL OPEN — unchanged, byte for byte** |
| N10 | 🟡 nothing exercises the transform at realistic scale | **PARTIALLY RESOLVED** — see M1 |

### What I actually checked

**N1 — resolved.** `statSync(envPath).mode` is now hoisted above the backup write (`server/src/routes/config.ts:278-288`), and `chmodSync(bakPath, originalMode)` follows `writeFileSync(bakPath, …)` at `:294-297`. I probed `chmodSync` with the raw `statSync().mode` (`0o100666`, file-type bits included) rather than assuming: Node accepts it and the kernel masks to `0o666`, so not applying pass 2's suggested `& 0o777` is harmless. The restored regression test now asserts both files (`config.env-cleanup.test.ts:332-337`), which is the half that was blind before. Note it `return`s early on `win32`, so this assertion is only ever exercised in cloud CI — `verify.yml`'s server leg is `ubuntu-latest`, so that is real coverage, just not local.

**N2 — resolved, and I measured it rather than reading the entry.** `.gitignore` is registered in `test:hooks`' `extraFiles` (`scripts/verify-cache.mjs:300-308`) with the same `#1847` rationale its neighbours carry, plus a paired assertion at `scripts/tests/verify-cache.test.mjs:545-548`. Re-running pass 2's own probe:

```
$ node -e "import('./scripts/ci-scope.mjs').then(m => console.log(
    m.computeScopes(['.gitignore'],{eventName:'pull_request'}).step_test_hooks))"
true          # was false at 6bf8d9dc
```

**N3 — resolved.** `git grep -n 'envCleanupCandidateKeys\b'` now returns **zero** hits for the bare name; only `envCleanupCandidateKeysFromFileContent` survives (2 call sites in `config.ts`, 1 definition, 6 in `resolver.test.ts`). The four tests that covered nothing are gone (125 tests now vs. pass 2's 126 = −4 +3 new, which reconciles exactly). I checked the removal did not take a live `beforeEach` with it: the deleted block's `beforeEach` was describe-scoped, and `envCleanupCandidateKeysFromFileContent` reads neither `process.env` nor `readConfigOverrides`, so the remaining describe is immune to what that hook used to clean.

**N4 — resolved, and the paired test is a real one.** `ENV_LINE_RE` is now `/^([ \t]*)((?:export[ \t]+)?)([A-Z0-9_]+)=(.*)$/gm` (`env-cleanup.ts:27`). I ran the shipped transform under both regexes on the exact fixture:

```
input : "export\nFOO=bar\n"
OLD   : {"text":"# export\nFOO=bar\n","cleaned":["FOO"]}   ← pass 2's bug
NEW   : {"text":"export\n# FOO=bar\n","cleaned":["FOO"]}
```

`env-cleanup.test.ts:97-107` asserts the exact `NEW` string, so it is red before the fix **for the reason claimed** — not for an incidental one.

**N7 — resolved.** `resolver.test.ts:369-385` wraps the mutation in `try/finally` and restores-or-deletes. The other file's `beforeEach`/`afterEach` snapshot-and-restore of the whole `process.env` (`config.env-cleanup.test.ts:77-104`) is still in place and still unconditional.

**N8 — resolved on both halves.** `.gitignore:22` carries `server/.env.tmp-*`, confirmed live: `git check-ignore -v server/.env.tmp-123-456-1-abcd` → `.gitignore:22`. And the outer `catch` at `config.ts:328-334` unlinks when `tmp` was assigned, so a `writeFileSync`/`chmodSync` throw no longer leaks a secrets copy — the gap pass 2 named.

**Also re-derived, since it is the invariant this whole PR exists to protect:** candidacy is computed from the *last* occurrence of a duplicated key (`parseEnvFileLines`' `map.set`), while `cleanEnvText` comments out *every* occurrence. I checked Node's own resolution rather than trusting the code comment — `process.loadEnvFile` on `X=first\nX=second` yields `"second"`, so last-wins is correct, and the asymmetry is safe in both directions: if the last value is the default, commenting all copies leaves the effective value at that same default; if it differs, nothing is touched at all. No pin can be destroyed through that door.

---

## Still open

### 🟡 Minor — cleanup — N9 is untouched, and the commit credited with it fixed a different comment

`server/src/config/env-cleanup.ts:55-57`, current `HEAD`:

```
    @param isCandidate  Predicate: true when the env-var name is a leftover-
                        default candidate at write time (re-derived from live
                        process.env + registry default — never a cached GET). */
```

That is byte-identical to what pass 2 quoted. The predicate has not been derived from `process.env` since `d17e55e3`; both call sites pass a set built by `envCleanupCandidateKeysFromFileContent(original)` (`config.ts:259`) — inverting that source is the *entire content* of finding 5's fix, and this docstring still documents the bug as the design.

`cc61faca` is credited with "fixed an inverted docstring", but `git diff 6bf8d9dc..HEAD -- server/src/config/env-cleanup.ts` shows it edited the **file-header** comment (`bare KEY=VALUE shape` → `including indented and export-prefixed variants`) — a correct edit for N4, unrelated to N9. Two different comments in one file; the wrong one moved. Reviewer-brief shape #9, and this is the one line most likely to send the next engineer back to `process.env`.

### 🟡 Minor — cleanup — N5's misattribution survived; only the tautology was removed

`env-cleanup.test.ts:225-240` still reads:

```ts
describe('env-cleanup integration: file-based candidacy (#2194 finding 5)', () => {
  it('does not comment out a line whose file value differs from default, even if process.env is shadowed', () => {
    const fileContent = 'GEN_WORKERS=4\nOTHER_VAR=1\n';
    // The candidacy check would be done against the file, not process.env.
    // For this test, we'll show that a non-candidate line is left untouched.
    const isCandidate = (varName: string) => varName === 'OTHER_VAR';
```

`da916b49` tightened the assertion to an exact `toBe('GEN_WORKERS=4\n# OTHER_VAR=1\n')`, which does remove the literal tautology — as a `cleanEnvText` test it is now fine. But pass 2's finding was not the assertion; it was that the test never calls `envCleanupCandidateKeysFromFileContent`, never sets `process.env`, supplies its own predicate, and therefore stays green under a **complete revert of `d17e55e3`** — the commit its describe name cites. That is still true, and the it-name still promises "even if process.env is shadowed" while never touching it. The comment quoted above still concedes it in the test's own words. The real coverage is `resolver.test.ts:359-386` and it is good; this one duplicates `:111-120` under a name that overstates it.

### 🟡 Minor — cleanup — M1: N10's fixture test hand-supplies the candidate set, and three of its 24 entries are not registry knobs at all

`env-cleanup.test.ts:245-467` adds the ~110-line fixture, which is a genuine improvement — the `expect(result.cleaned.sort()).toEqual(Array.from(candidateVars).sort())` assertion is load-bearing and would catch any line the transform missed, including the indented and `export`-prefixed ones. That covers half (b) of pass 2's ask.

Half (a) — `envCleanupCandidateKeysFromFileContent` over a realistic file — is still untested, because the test supplies `candidateVars` itself. That is invisible until you check the annotations, which are wrong:

```ts
'PORT',                         // 8080 (default)
'BIND_HOST',                    // 127.0.0.1 (default)
'WORKSPACE_DIR',                // ../audiobook-workspace (default)
```

`git grep -oh "env: '[A-Z0-9_]*'" server/src/config/registry.ts | sort -u` enumerates all 115 registered env names. `PORT`, `BIND_HOST` and `WORKSPACE_DIR` are **not among them** — so `envCleanupCandidateKeysFromFileContent` can never return them, and production would never comment those lines out (correctly: commenting out a real `WORKSPACE_DIR` or `PORT` would be destructive). The test asserts they *are* commented out and labels them "(default)". Nothing verifies the annotation, so the falsehood is free. Reviewer-brief shapes #3 + #9.

The cheap close is to drive the predicate from the real function — `const keys = new Set(envCleanupCandidateKeysFromFileContent(envContent).map(k => getKnob(k)!.env!))` — which makes the fixture prove both halves and makes a wrong annotation fail.

### 🟡 Minor — cleanup — M2: N1's fix chmods *after* creating the file, so the backup exists world-readable for a window

`config.ts:294-298`:

```ts
writeFileSync(bakPath, original, 'utf-8');   // created at 0666 & ~umask → 0644
if (originalMode !== undefined) chmodSync(bakPath, originalMode);
```

Same shape at `:315-319` for the temp file. Between the two syscalls a byte-for-byte copy of `server/.env` — `GEMINI_API_KEY`, `LAN_AUTH_TOKEN` — is `0644` on POSIX, and it stays `0644` permanently if the process dies in between (SIGTERM during a shutdown-time cleanup, an OOM kill). Pass 2's suggested fix was `{ encoding: 'utf-8', mode: originalMode & 0o777 }` **plus** the chmod; only the chmod half landed. Adding the `mode` option closes the window without removing the chmod (which is still needed, since `mode` applies only at creation and a pre-existing 0644 `.env.bak` would keep it).

### 🟡 Minor — cleanup — M3: the new `.gitignore` test's title asserts a property the pattern does not have

`scripts/tests/gitignore-secrets.test.mjs:63-71`:

```js
test('gitignore: .env.tmp-* files in root are also covered by the pattern', () => {
  assert.equal(isGitIgnored('server/.env.tmp-999-9999999999999-5-xyz'), true, …);
});
```

The title claims root coverage; the body asserts a `server/`-prefixed path, i.e. a verbatim duplicate of the test immediately above it with different digits. Measured in the worktree:

```
$ git check-ignore -v --no-index .env.tmp-123-456-1-abcd server/.env.tmp-123-456-1-abcd
.gitignore:22:server/.env.tmp-*   server/.env.tmp-123-456-1-abcd
                                  # ← nothing for the root path
```

`server/.env.tmp-*` contains a slash, so it is anchored to the repo root and does **not** match a root-level `.env.tmp-*` (unlike `.env.bak`, which is unanchored and matches at any depth). The anchor itself is correct for production — `npm --prefix server run dev` runs with cwd `server/`, which is what `resolveServerEnvPath()` mirrors — so this is a test-title defect, not a gap. Either delete the duplicate or make it assert the root path it names (and widen the pattern if you want that to be true).

### 🟡 Minor — cleanup — M4: the test file's header advertises a Part E that does not exist

`config.env-cleanup.test.ts:10-11`:

> 4. (Part E) When chmod fails during temp file preparation, the temp file is cleaned up to avoid leaving secrets-bearing temp files on disk.

The test at `:340-379` performs a **successful** cleanup and asserts `readdirSync(tmpDir)` contains no `.tmp-` entries. On the success path the rename removes the temp regardless, so this cannot fail for the failure-path behaviour its header names — the `writeFileSync`/`chmodSync`-throws branch that N8 was actually about (`config.ts:328-334`) is still uncovered. Its own body comment ("This is a defensive regression test") is honest; the header is not. Same shape as N9: the instrument's description outran the instrument.

### 🟡 Minor — cleanup — M5: the rename-failure unlink is now duplicated

`config.ts:320-327` catches a `renameWithRetry` failure, unlinks, and rethrows — into the outer `catch` at `:328-334`, which unlinks the same path again (the second call ENOENTs and is swallowed by `.catch(() => {})`). The outer handler was added for the `writeFileSync`/`chmodSync` case and already covers the rename case; the inner try/catch is dead weight the fix chain left behind. Harmless, but it is the same "one round's fix layered on top of another's" residue this gate has flagged twice.

### 🟡 Minor — cleanup — M6: `resolveServerEnvPath` is still exported with no consumer outside its own module

`git grep -n resolveServerEnvPath -- '*.ts'` → `config.ts:68`, `config.ts:225` (definition), `config.ts:240`, and one *comment* mention in `config.env-cleanup.test.ts:255`. Nothing imports it. This is the parenthetical half of N3 ("Same shape, smaller") — the `export` was only ever there for the query-param seam that pass 1 removed. `_setServerEnvPathForTest` still needs its export; this one does not.

---

## New this round

### 🟡 Minor — cleanup — M7: the notice's singular form is ungrammatical, and a test certifies it

`src/components/env-cleanup-notice.tsx:21-26`:

```tsx
const settingWord = candidateCount === 1 ? 'setting' : 'settings';
…
{candidateCount} {settingWord} look like leftover defaults from an older install.
```

The noun is pluralised; the verb is not. At `candidateCount === 1` the user sees **"1 setting look like leftover defaults from an older install."** `src/components/env-cleanup-notice.test.tsx:19-22` — named `uses the singular "setting" for a count of 1` — asserts `/1 setting look/i`, so the half-done pluralisation is locked in by its own test. (Pass 1's write-up quoted this string as "1 setting **looks** like…" — that was me reading the intent, not the code; the code has never said that.) One-word fix in both the component and the assertion.

### 🟡 Minor — correctness — M8: candidacy fails to fire on a CRLF (or trailing-whitespace) `.env` for enum and pattern-less string knobs

`ENV_LINE_RE`'s value group is `(.*)$` under `/m`, and `.` matches `\r`. On a CRLF file, `parseEnvFileLines` yields `'auto\r'`, not `'auto'`. `coerceAndValidate` (`resolver.ts:183-223`) then splits by type:

- `boolean` → `String(raw).trim().toLowerCase()` — tolerant ✅
- `integer`/`number` → `Number(String(raw).trim())` — tolerant ✅
- `enum` → `String(raw)`, **no trim** → `options.includes('auto\r')` is false ❌
- `string` → trimmed only when the knob has a `pattern`; otherwise raw ❌

So on a CRLF `server/.env` — or one hand-edited to leave a trailing space — every enum and pattern-less string knob is silently absent from `envCleanupCandidates`. The notice under-counts, or never appears at all, on a file that is entirely leftover defaults.

This **fails closed**, which is why it is Minor and not Significant: GET and POST run the identical predicate, so the two never disagree, and no line is ever wrongly commented out. Pass 2 explicitly accepted this class for inline comments and quoted values, and I am not reopening that; I am naming the CRLF instance because it is Windows-specific and this is a Windows-first project, and because it was not among the shapes pass 2 enumerated. Likelihood is genuinely low — `.gitattributes` pins `.env.example` to LF, and both documented install paths (`cp` from that file, and `pinokio-scripts/lib/write-env.js`) emit LF — so a CRLF `server/.env` needs a deliberate editor round-trip. A one-line `.replace(/\r$/, '')` on the captured value inside `parseEnvFileLines` would close it, and `cleanEnvText` needs no change (it already round-trips CRLF byte-for-byte, verified by pass 2).

---

## What is solid

- **The three 🔴 findings from pass 1 are still closed, and this round did not reopen them.** I re-derived rather than trusting the table: `req.query` appears nowhere in either handler; the temp file is written beside the target so there is no `EXDEV` path; and `config.env-cleanup.test.ts:107-127` still proves the notice clears in one process with no restart.
- **The regex narrowing did not cost anything.** `[ \t]` drops `\f`, `\v`, ` ` as legal indents — all of which now fail *closed* in GET and POST alike, so the two stay consistent. All eight indent/`export` cases from pass 1's finding 11 still pass.
- **The duplicate-key asymmetry is safe**, verified against Node's own parser rather than the code comment (above).
- **N6's fix went further than asked.** The PR body's mutation proof now names the live function, and the `## Also fixed, found in passing` section enumerates all 21 folded findings across both prior rounds with the two review threads linked — which is what CLAUDE.md's report-fix-record rule actually wants.
- **The commit-history reset did not lose anything.** I diffed `6bf8d9dc..HEAD` as a whole rather than commit-by-commit precisely because of the re-split, and every change the eight messages claim is present in the cumulative diff. The one mismatch is N9, and it is a wrong-target fix, not a lost one.
- 125/125 green across the four touched server suites; 1422/1422 green on `test:hooks`; `Closes #2194` plain and outside any code span; both release-notes files landed and their claims are now true as shipped.

---

## Verdict

**Approve.** No blocking findings, no significant findings. The eight fix commits since pass 2 did what they claimed with one exception (N9), and — the thing this round was dispatched to look for — **they did not break or duplicate each other's work in any way that changes behaviour.** The only cross-fix residue is cosmetic: a redundant unlink (M5) and two test artefacts whose titles outran their bodies (M3, M4).

What is left is eight Minor items, all documentation or test hygiene except M2 (a sub-millisecond permission window on a file that is chmod'd correctly one syscall later) and M8 (a fail-closed under-detection on a file shape that neither install path produces). None of them can produce a wrong or destructive result on a stock install, and none of them is worth another gate round.

If you fold anything before merging, fold **N9** — it is the only item that will actively mislead the next person to open this file, and it is a three-line comment edit.

Issue #2491, opened 2026-08-20
[agent instructions][claude][verify] #2427 colon rule — re-verdict after lint rework, open the PR


dudarenok-maker:
Parent: #2427. Second verify pass — the first (#2428) reached a lint-only
FAIL after the substantive checks all passed. This child confirms the rework
(#2490) actually fixed it, then finishes the job #2428 could not: open the PR.

**You judge, you do not fix.**

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-2346-colon-tag-clause`
- Branch: `fix/server-2346-colon-tag-clause`

Do **not** create another worktree. Do **not** work in the primary checkout at
`C:\Claude\Projects\Audiobook-Generator`. **Source files are read-only for
you.**

## What #2428 already established (do not re-derive from scratch, but do read
the receipt and spot-check it — do not blindly trust it either)

Read #2428's `AGENT DONE`-adjacent comment (search this issue's linked
timeline / #2428 directly) for the full account. Summary:

1. **Mutant A** (delete the two-line colon guard) — re-run live, observed
   3 of 5 targeted tests RED, 2 control cases GREEN, then reverted.
2. **Mutant B** (delete only the `.replace(/\s+$/u, '')` trailing-whitespace
   strip) — re-run live, observed exactly ONE red (the real `pg/de/63460`
   paragraph), 4 of 5 stayed green, then reverted.
3. **Corpus A/B** — both arms report `books 331 · paragraphs 726385` on
   stderr (totals match). Diff is exactly 2 changed paragraphs in exactly 2
   books: `pg/de/63460.txt` and `pg/zh/52200.txt`. No other book touched.
4. Verdict was **FAIL only on acceptance item 9** —
   `npm run verify:fast:branch` failed at the `lint` step:
   `no-useless-escape` on `parser.test.ts:556` and `:573` (both added by this
   branch's own commits). Every other acceptance item (1–8 in #2428's body)
   was checked and passed.

## Step 1 — confirm the rework actually landed

Check that #2490 is closed/done and that its fix is actually on this branch:

```bash
grep -n "Guten Tag" server/src/analyzer/dialogue-structure/parser.test.ts
```

Lines 556 and 573 (or wherever they now are) must read `'„Guten Tag", ...'`
with **no** backslash before the `"`. If the fix is not present, this is not
yet ready — leave a comment saying so, set Status back to `Agent Needs Input`,
and stop; do not park a fabricated PASS.

[...]

Issue #2490, opened 2026-08-20
[agent instructions][cline-qwen-cloud][task] #2427 colon rule — fix no-useless-escape lint errors in parser.test.ts:556,573


dudarenok-maker:
Parent: #2427. Rework child spawned by a FAILED verify (#2428).

## What

Fix a two-line lint defect in the branch that #2428 verdicted FAIL against.
Everything else about the colon-rule fix is correct and verified — this is
purely a lint cleanup, no logic change.

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-2346-colon-tag-clause`
- Branch: `fix/server-2346-colon-tag-clause`

Do not create a new worktree; continue on the existing branch/worktree above.

## The defect

`npm --prefix "C:/Claude/Projects/wt-2346-colon-tag-clause" run verify:fast:branch`
fails at the `lint` step:

```
C:\Claude\Projects\wt-2346-colon-tag-clause\server\src\analyzer\dialogue-structure\parser.test.ts
  556:30  error  Unnecessary escape character: \"  no-useless-escape
  573:30  error  Unnecessary escape character: \"  no-useless-escape

✖ 2 problems (2 errors, 0 warnings)
```

Both offending lines were added by this branch's own commits (part of the
`#2346 defect B` test additions), not pre-existing on `main`:

```
server/src/analyzer/dialogue-structure/parser.test.ts:556:
    expect(spoken('„Guten Tag\", sagte er. Dann sagte er:«Hallo».', 'de')).toEqual(['Guten Tag', 'Hallo']);

server/src/analyzer/dialogue-structure/parser.test.ts:573:
    expect(spoken('„Guten Tag\", sagte «Ulebuhle». Und dann ging er.', 'de')).toEqual(['Guten Tag']);
```

Each string literal is single-quoted, so the `"` inside it needs no escaping —
`\"` is a no-op that ESLint's `no-useless-escape` rule rejects under
`--max-warnings 0`. Line 17 of the same file already writes the equivalent
character unescaped (`'„Guten Tag", sagte er. ...'`) — follow that existing
style.

## Fix

On both lines 556 and 573, change `\"` to `"` (drop the backslash). Do not
touch anything else in the file or in `parser.ts` / `lang/de.ts` — the
underlying colon-rule logic, the corpus A/B result, and both mutation checks
were already independently re-verified by #2428 and are not in question here.

## Verify before handing off

Run detached and poll (do not run in the foreground — this repeats the whole
[...]

Issue #2428, opened 2026-08-18
[agent instructions][claude][verify] #2427 colon rule — verdict, corpus A/B, open the PR


dudarenok-maker:
Parent: #2427. Final step of the chain. **You judge, you do not fix.**

## What

Verdict the whole branch against #2427's acceptance, re-run any mutation whose receipt asserted a result without showing output, run the one measurement no earlier child was allowed to run, and — on PASS — open the PR.

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-2346-colon-tag-clause`
- Branch: `fix/server-2346-colon-tag-clause`

Do **not** create another worktree. Do **not** work in the primary checkout at `C:\Claude\Projects\Audiobook-Generator` — you *read* it, as the A/B's shipped arm, and never write to it. **Source files are read-only for you.** If something is wrong, you report it and open rework children; you do not edit `server/**`.

## Background

`cutsATagClause` (`server/src/analyzer/dialogue-structure/parser.ts:439`) declined a secondary-tier quote-run whenever a primary run ended before it and the clause between carried a speech verb — correct for a *trailing* tag, backwards for a *leading* one, where the verb introduces the turn. The branch adds one early `return false` when a colon (`:` or `：`) immediately precedes the candidate, after trailing whitespace is stripped.

**Where it fires the guard is switched off entirely.** That is safe only because the real corpus contains exactly two paragraphs where it fires and both were read. Verifying that bound is your job, and nothing before you was permitted to do it.

## Step 1 — re-run both mutations yourself

Do not trust the receipts. Apply each mutant, run the suite, observe, revert.

**Mutant A — delete the rule.** Remove these two lines from `cutsATagClause`:

```ts
const lastCh = line.slice(0, cand.start).replace(/\s+$/u, '').slice(-1);
if (lastCh === ':' || lastCh === '：') return false;
```

```bash
npm --prefix "C:/Claude/Projects/wt-2346-colon-tag-clause/server" test -- src/analyzer/dialogue-structure/parser.test.ts -t "#2346 defect B"
```

Expected: **3 of 5 RED**, the two control cases GREEN. Then `git checkout -- server/src/analyzer/dialogue-structure/parser.ts`.

**Mutant B — the silent one.** Restore the rule but delete only `.replace(/\s+$/u, '')`:

```ts
const lastCh = line.slice(0, cand.start).slice(-1);
```

[...]

Issue #2346, opened 2026-08-13
The tag-clause guard (#2315) is inert for a paragraph typed wholly in a secondary-tier convention


dudarenok-maker:
## What

`cutsATagClause` (`server/src/analyzer/dialogue-structure/parser.ts:~439`, PR #2340 round-2 finding F2, `#2315`'s tag-clause guard) requires an already-captured **primary-tier** run to have already ENDED BEFORE a secondary-tier candidate before it will ever decline that candidate — `precededByPrimaryRun` is set only for a primary run whose `end <= cand.start`. That precondition (correctly) fixed round 1's MAJOR finding — the guard being polarity-inverted for CJK's verb-before-quote convention — but it has a side effect the round-1 fix didn't measure: **the guard is inert for any secondary-tier candidate with no primary run preceding it** — which is not only "a paragraph typed wholly in a secondary-tier convention" (an earlier revision of this issue understated the condition that way — corrected, PR #2340 round 3 finding C2): a paragraph CAN carry a primary run and still be fully inert for a candidate that sits before it.

That is defect 2 (#2315) verbatim, in exactly the population #2286 exists to serve — a book whose dialogue is typed entirely in a pair #2286 moves to `secondaryQuotePairs`:

| input (secondary tier declared) | shipped | naive fix (`out` instead of `primaryRuns`) |
|---|---|---|
| `‘Привет’, сказал ‘Антон’.` (ru, secondary `‘’`) | `Привет=NULL \| Антон=NULL` | `Привет=anton` |
| `«Hi», said «Anton».` (en, secondary `«»`) | `Hi=NULL \| Anton=NULL` | `Hi=anton` |
| `Said «Anton», "Hi there."` (en, leading tag — **HAS** a primary run, `"…"` is an `en` primary pair, it just comes AFTER the candidate) | `Anton=NULL \| Hi there.=NULL` | `Hi there.=anton` |

## Why this is a decision, not a fix

The obvious repair — pass `out` (primary runs **plus already-accepted secondary runs**) to `cutsATagClause` instead of `primaryRuns` alone — fixes all three rows above and keeps every round-1 CJK unit case green. But measured at corpus scale against #2286's actual tables, it **re-declines 5,892 spans in `pg/zh/23835.txt`** — the exact book round 1's MAJOR finding was about (93.7% of its speech spans falsely declined pre-round-1-fix) — reinstating that finding under a different trigger.

[...]

PR #2518, opened 2026-08-20
fix(server): stop the tag-clause guard eating a colon-introduced turn

### LINKED ISSUE #2427: The tag-clause guard (#2315) deletes a turn introduced by a leading tag
`cutsATagClause` (`server/src/analyzer/dialogue-structure/parser.ts:439`) declines a secondary-tier quote-run whenever a primary run ends before it and the clause between them carries a speech verb. That is the right test for a **trailing** tag. It is backwards for a **leading** one, where the verb *introduces* the turn rather than attributing an already-parsed one — and the guard then drops the entire spoken sentence.

## The real case

`pg/de/63460.txt` (Project Gutenberg 63460):

> …dann **sagte** er nur in seiner knurrigen Weise**:** «Schnickschnack und Finger davon! Das versteht ihr nicht!»

An unrelated earlier primary run in the same paragraph — a clock going `»Tick-Tack ... Tick-Tack«` — sets `precededByPrimaryRun`; the clause carries `sagte`; the guard declines. The whole line is lost from the audiobook.

Measured, shipped vs. the fix:

```
narration  "Es war gruselig und dunkel auf der Stiege, …"
speech     "Tick-Tack ... Tick-Tack"
tag        " sagte. -- Da waren auch noch allerlei Apparate …"
                                        <-- the spoken sentence ends here, gone
```

## Measured impact

**2 changed paragraphs in 726,385**, across 331 books, in exactly two books:

| book | verdict |
|---|---|
| `pg/de/63460.txt` | the target — the turn is recovered as its own speech span |
| `pg/zh/52200.txt` | a **gain** — a tag containing a nested quoted turn splits into `tag` + `speech` + `narration`; no speaker changes |

## Segmentation, not attribution

The recovered turn carries `speaker: null`. A line that was silently dropped is now spoken; nobody is newly attributed to it. Release notes must say so.

## Relationship to #2346

This is the **mirror** of #2346, which stays open. There the guard is *inert* (no primary run precedes the candidate), a name is wrongly admitted, and the neighbouring turn loses its speaker. Here the guard *fires wrongly* and destroys a turn.

The two are **disjoint populations**, separated by `precededByPrimaryRun`, and must never be averaged. #2346's defect A is accepted and priced at ≤21 of 1,164 proxy firings; fixing it needs decline-widening, the move that re-declines 5,892 genuine spans in `pg/zh/23835.txt` and reinstates PR #2340 round 1's MAJOR finding. This issue is admit-only and bounded by an enumerated list of 102 declines.

## Design and plan

- Spec: `docs/superpowers/specs/2026-08-18-tag-clause-name-shape-design.md`
- Plan: `docs/superpowers/plans/2026-08-18-2346-colon-rule.md`
- Both land via #2426. Handover brief: #2346 (comment).

Three encodings were built and falsified before the shipping one survived — per-language word order, candidate-is-name-shaped, and candidate-contains-a-roster-name. Each picks a property of the **candidate**; the distinction is a property of the **separator**. They are recorded in the spec so they are not re-proposed.

dudarenok-maker:
## Summary

- The tag-clause guard in `cutsATagClause` was dropping a spoken turn whenever
  it was introduced by a colon (e.g. `sagte er: «Zu».`) rather than a comma,
  because the guard's `.replace(/\s+$/u, '')` trailing-whitespace strip
  mishandled that shape. This fixes defect B on `#2346`.
- Mutation checks (re-run live, not trusted from a prior receipt):
  - **Mutant A** (delete the two-line colon guard entirely): 3 of 5 targeted
    tests went RED, 2 control cases stayed GREEN — matches spec.
  - **Mutant B** (delete only the `.replace(/\s+$/u, '')` strip, keep the
    rest of the rule): exactly ONE red — the real `pg/de/63460` paragraph —
    4 of 5 stayed green.
- Corpus A/B (shipped `main` vs. this branch), both arms `books 331 ·
  paragraphs 726385` (totals match): diff is exactly 2 changed paragraphs in
  exactly 2 books, `pg/de/63460.txt` and `pg/zh/52200.txt`. No other book
  touched.
- **No attribution claim is made.** This is a segmentation fix only — the
  recovered turn carries no speaker. The 42-case `reopen-sweep.test.ts`
  family is documented as blind to attribution and cannot support one.
- Also fixed, found in passing: three prose corrections that the colon-rule
  change made false — one in `lang/de.ts`, two in `parser.test.ts` (already
  on this branch).
- **On-box acceptance: not applicable** — pure parser logic; no GPU, sidecar,
  analyzer, or live book involved.
- This branch went through two review passes: `#2428` verdicted FAIL on lint
  only (`no-useless-escape` on two lines added by this branch's own test
  commits — every substantive check passed). `#2490` fixed those two lines
  (test-file-only diff). This PR's own verify pass (`#2491`) re-ran
  `verify:fast:branch` clean and confirmed `parser.ts` / `lang/de.ts` are
  unchanged since `#2428`'s review — only the two-line lint fix landed after
  it, so the corpus A/B and mutation checks above were not re-run (nothing
  that could move their result changed).
- **Independent PR review pass 1** (`pr-review-gate`, depth medium, head
  `a3f426de`) re-ran all of the above independently against the real corpus
  and confirmed every quantitative claim exactly. It found one Medium
  correctness finding — the trailing-whitespace strip's comment in
  `parser.ts` claimed the corpus has zero raw colon-adjacent contexts and
  that removing the strip does nothing at all; both were measurably false
  (the zh flip paragraph fires raw-adjacent, and removing the strip still
  flips that one paragraph) — plus three cleanup findings (a misleading
  "no reduced fixture can pin this" test comment, this mislocation, and two
  formatting leftovers). All four are fixed in `70c1457b`, including a new
  minimal regression test that pins the trailing-whitespace strip on its own.

Closes #2427
Refs #2346

## Test plan

- [x] `verify:fast:branch` green end to end (lint through build; `test` and
      `test:sidecar` out of scope for this diff)
- [x] Mutant A and Mutant B re-run live (see above)
- [x] Corpus A/B diff confirmed exactly 2 paragraphs in 2 books
- [x] `parser.ts` / `lang/de.ts` confirmed byte-identical to what `#2428`
      reviewed (only `parser.test.ts`'s two-line lint fix landed since)
- [x] Independent PR review pass 1 findings fixed and pushed (`70c1457b`)
- [ ] Defect A (`#2346`) — out of scope, stays open


dudarenok-maker:
## PR review — pass 1 (head a3f426de73d3f7482f751ff42721d740fe810249, depth medium)

### What I re-ran independently (not taken from the PR body)

Every claim below was re-measured against the real corpus at `C:\Claude\castwright-corpus`, arm A = shipped `main`, arm B = this PR's head extracted to a temp tree, single process, same loader/paragraph splitter.

| claim | verdict |
|---|---|
| Corpus A/B totals `331 books · 726385 paragraphs`, both arms | **confirmed, exact** |
| Diff is exactly 2 changed paragraphs in exactly 2 books | **confirmed** — `pg/de/63460.txt` #495, `pg/zh/52200.txt` #831, nothing else |
| Both deltas are gains, not regressions | **confirmed by inspection** — de recovers `«Schnickschnack und Finger davon! …»` as `speech:1043-1099`; zh splits `tag:208-225` into `tag:208-212 \| speech:213-219 \| narration:220-225`, recovering `‘我圖他潤肺。’` from inside a tag clause. No speaker changes in either |
| Mutant A (colon rule absent = `main`): 3 of 5 red, 2 controls green | **confirmed** — T1/T2/T3 red, T4/T5 green |
| Mutant B (strip removed): exactly 1 red, the real `pg/de/63460` paragraph | **confirmed** — T1 red, T2–T5 green |
| `parser.ts` / `lang/de.ts` byte-identical since #2428; a3f426de is test-only | **confirmed** — a3f426de removes two `\"` escapes inside string literals, behaviourally identical strings |

I also ran two mutants the PR did not: dropping **only** the `：` arm reddens T3 alone; dropping **only** the `:` arm reddens T1+T2. Both spellings are independently pinned — good.

And one enumeration probe, since "a guard that enumerates syntax" is a standing shape here: the corpus carries `﹕` (U+FE55) ×16 and `∶` (U+2236) ×30 alongside `:` ×91,496 and `：` ×175,405. Adding both unlisted glyphs to the rule changes **zero** corpus paragraphs. The two-glyph list is empirically adequate — **no finding**.

House gates: release-notes pair present and both in the correct in-progress sections (`RELEASE_NOTES.md` under `# Castwright 1.15.0`, `release-notes-next.md` under `## 🗣️ Analyzer, script review & manuscript`); `Closes #2427` / `Refs #2346` present, outside code spans; #2427 is `bug`-labelled so no `docs/BACKLOG.md` row is owed and none was added — correct; on-box "not applicable" is right (pure parser, no hardware surface); no `cast.json` write, no config knob, no derived artifact in scope.

---

### Correctness

**1 — `server/src/analyzer/dialogue-structure/parser.ts:461-464`: the comment that justifies the whitespace strip states two facts that are both measurably false, and the PR's own Mutant B receipt contradicts it.** — *Medium*

The comment reads:

> The trailing-whitespace strip is LOAD-BEARING, not tidying: the target paragraph's colon is followed by a space, and **the raw colon-adjacent count across the whole corpus is ZERO. Remove the strip and this does nothing at all, silently.**

Both sentences are wrong:

- **"raw colon-adjacent count across the whole corpus is ZERO"** — `pg/zh/52200.txt` #831, which is *one of this PR's own two flip paragraphs*, carries 31 raw colon-adjacent contexts (`：“` / `：‘`) and the rule fires on one of them, `那人道：‘我圖他潤肺。’`, with no whitespace at all. Chinese does not put a space after `：`.
- **"Remove the strip and this does nothing at all, silently"** — measured: with `.replace(/\s+$/u, '')` deleted, the corpus A/B still reports **1 changed paragraph** (`pg/zh/52200.txt` #831, byte-identical delta). Only the German half is lost. This is also what the PR body's own Mutant B line says — "4 of 5 stayed green", including both no-whitespace colon tests — three lines above the code it describes.

Concrete failure scenario, and why this is not just tidy-up: this comment is the **sole record** of why the strip and the second glyph exist, and it explicitly instructs the next maintainer that the bound "must be re-measured if the tables or the corpus change". Someone doing that re-measurement uses "raw colon-adjacent count" as the instrument and starts from a baseline of 0 that is really ~266,900. Worse, a maintainer told the corpus contains zero raw-adjacent colons has no reason to keep the `：` arm at all — and dropping *only* `：` (my Mutant C) silently re-loses the Chinese turn at `pg/zh/52200.txt` #831 with the entire rest of `parser.test.ts` still green except the one zh test. The code is correct; the recorded justification for it is not, and it points a future change straight at the half of the rule the comment says is unreachable.

Fix is comment-only — no code change is implied. Suggested shape: the strip is load-bearing **for the German paragraph specifically** (its colon is followed by a space); the Chinese paragraph fires raw-adjacent and survives without it; measured, removing the strip costs 1 of the 2 corpus deltas.

### Cleanup

**2 — `parser.test.ts:521-525` (and the fixture at `:548`): the strip is pinned by exactly one 1,128-char verbatim fixture, and the comment above it tells the next maintainer that a reduction cannot do the job. It can.**

The comment says "The body is the REAL paragraph … The reduction is what hides the bug … Shorten the body and the test passes for the wrong reason." Measured across three arms:

| fixture | `main` | this PR | strip removed |
|---|---|---|---|
| `„Guten Tag", sagte er. Dann sagte er: «Hallo».` (46 chars) | `["Guten Tag"]` **red** | `["Guten Tag","Hallo"]` green | `["Guten Tag"]` **red** |
| `Dann sagte er: «Hallo».` (over-reduced, no primary run) | `["Hallo"]` | `["Hallo"]` | `["Hallo"]` |

So a 46-char body that *keeps* the primary run pins both the rule **and** the strip. The comment's warning is true only of the second row — an over-reduction that drops the primary run — and as written it reads as "no reduced fixture can pin this", which will stop the next person adding the cheap one. Note `:555` is already a reduced body that keeps the primary run; it just has no space after the colon, so it does not reach the strip. One added space on a sibling case buys a minimal strip pin, so strip coverage no longer rests entirely on one verbatim paragraph that a later "simplification" could shorten.

**3 — PR body mislocates the incidental fixes.** The body says "three prose corrections to the guard's comments in `lang/de.ts`". Only one of the three is in `lang/de.ts` (`:34`, one line); the other two are in `parser.test.ts` (`:469-475` and `:1485-1490`) — matching the plan's Task 3 trio. The "Also fixed, found in passing" line is what a reviewer uses to separate requested lines from incidental ones; pointing at the wrong file makes it unverifiable from the body alone.

**4 — two formatting leftovers.** `parser.test.ts:576-578` has three consecutive blank lines after the new `describe` (prettier collapses to one; `format:check` is not in the gate and the file already fails prettier on `main`, so nothing catches it). `lang/de.ts:34` is now 92 chars where the rest of that comment block wraps at ~76 — the paragraph was edited in place without reflowing.

---

Nothing else. The rule itself, its placement ahead of `precededByPrimaryRun`, the empty-prefix case (`cand.start === 0` → `''`, no match), the disjointness argument against defect A's pinned test at `:1485`, and the reversal of that test's "expected to start FAILING" prose all check out — the three defect-A bodies contain no colon and none of them changes colour.

Issue #2241, opened 2026-08-10
ops-55 — route the mandatory PR review gate through a model-reachable mechanism


dudarenok-maker:
## What

The mandatory before-shipping PR review gate names a mechanism that the model
cannot execute. Re-point it at one that can be executed, and encode the
substitute brief as a real, model-invocable project skill.

## Why

`CLAUDE.md` step 10 (line 574), the "Mandatory review gates" bullet
(line 238), and `.claude/skills/model-routing/SKILL.md` line 111 all name the
built-in `code-review` command as *the* mechanism for the gate.

That command is **user-invocable only**. `Skill(skill: "code-review")` fails
with `cannot be used with Skill tool due to disable-model-invocation`, and a
dispatched subagent hits the same wall — so "dispatch an agent and tell it to
invoke the skill" is not a workaround either.

Consequences, both live today:

1. Every session that reaches a PR re-derives the gap from scratch and
   announces it, because the contradiction reloads with `CLAUDE.md` each time.
   The workaround (dispatch an adversarial Opus-tier agent) is reconstructed
   from memory as ad-hoc prose at every PR, so its quality depends on how well
   it happens to get phrased that day.
2. A gate whose documented mechanism is unreachable is a gate that can be
   silently skipped. It has been at least once (PR #1875 / #1835, 2026-07-27),
   with a self-review substituted and the omission mis-attributed to session
   configuration.

Note there is a second, *reachable* `code-review` — the installed
`code-review@claude-plugins-official` plugin command, which sets
`disable-model-invocation: false`. It is deliberately **not** the answer: it is
a different tool that reviews a GitHub PR, discards every finding below 80%
confidence, and posts a public comment on the PR rather than returning a report
for hand-triage. `model-routing/SKILL.md:115-117` already draws exactly this
distinction ("the branch-diff tool — not the separate PR-comment command"), so
routing the gate through it would quietly change the gate's semantics and start
publishing review output to a public repo.

## Acceptance

- [ ] The gate's *Mechanism* is a non-fork Premium-tier `Agent` dispatch that
      the model can perform unattended. `/code-review <level>` is documented as
      the user-typed upgrade, not as the mandated mechanism.
- [ ] A new `.claude/skills/pr-review-gate/SKILL.md` exists, is model-invocable
      (no `disable-model-invocation`), and carries the reviewer brief so the
      gate is one skill call rather than re-derived prose.
[...]

Issue #2216, opened 2026-08-07
ops — the git-env scrub has been applied three times by incident; the population it should cover was never enumerated


dudarenok-maker:
## What

`scrubGitEnv` (`scripts/git-env.mjs`) exists because git resolves the repository from `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` / `GIT_OBJECT_DIRECTORY` / `GIT_COMMON_DIR` **before** it falls back to discovery from `cwd`. A script that pins an invocation to an explicit `cwd` is silently overridden whenever one of those is set in the ambient environment, and the command still exits 0 — against the wrong repository.

It has been applied three times, each time to the site that happened to be under review:

- #2169 — `execGit` in `scripts/bump-version.mjs`, and `readTagAnnotation` in `scripts/release-body.mjs`
- #2184 — the new `scripts/gh.mjs` chokepoint, and `scripts/lib/run-command.mjs` (found in passing, because it runs `git push` for `sync-wiki.mjs`)
- #2193 — `run()` in `scripts/run-golden-audio.mjs`

**The set it ought to cover has never been enumerated.** Doing that while deciding #2193 turns up roughly **24 direct `git` invocations under `scripts/` and `pinokio-scripts/` that do not scrub** — none of them a wrapper, which is why every previous wrapper-shaped search missed them:

| Group | Sites |
|---|---|
| Inside git hooks | `verify-cache.mjs` (×4), `guard-protected-push.mjs`, `guard-commit-subjects.mjs`, `is-docs-only-push.mjs` |
| Worktree tooling (caller-supplied `cwd`) | `wt-new.mjs` (×2), `wt-merge.mjs`, `wt-list.mjs` |
| Pinokio installer | `pinokio-scripts/lib/resolve-release.js` — `git tag --list`, `git fetch --tags --force`, `git checkout <tag>` |
| Other | `audit-branches-worktrees.mjs` (×3), `backlog-sync.mjs`, `thin-backlog.mjs`, `fix-archived-plan-pointers.mjs`, `lib/module-graph.mjs`, `sync-wiki.mjs:100` |

Only `bump-version.mjs`, `release-body.mjs`, `gh.mjs` and `lib/run-command.mjs` scrub.

## What this is NOT claiming

[...]

PR #2532, opened 2026-08-20
fix(scripts): gitignore .env.bak, and test that secret files cannot be staged

### LINKED ISSUE #2531: .gitignore does not cover .env.bak, allowing a real secret into staging
`.gitignore` covers `.env` but not `.env.bak`. Any tool or workflow that backs an env file up in place (e.g. `cp server/.env server/.env.bak` before an edit) drops a real secret file into the working tree unignored, where a broad `git add` can stage it into a public repo.

Fix: extend `.gitignore`'s secret patterns to include `.env.bak`, and add a regression test that drives `git check-ignore` over the secret patterns (rather than asserting on `.gitignore`'s literal text) so a later reorganisation of the file can't silently drop the protection.

dudarenok-maker:
## Summary
- `.gitignore` covered `.env` but not `.env.bak` or any other backup-shaped variant — anything that backs an env file up in place drops a real secret into a public repo's staging area. Widened `.gitignore` to also cover the repo's own rotating-backup spelling (`.env.bak.N`, `state-io.ts`'s `rotateBackups()`) and common editor backups (`.env~`, `.env.backup`, `.env.old`, `.env.save`).
- Added a regression test (`scripts/tests/gitignore-secrets.test.mjs`) that drives `git check-ignore --no-index` over the secret patterns rather than asserting on `.gitignore`'s literal text, with paired coverage for every new pattern and a negative control against the real tracked `server/.env.example`. Uses `scrubGitEnv()` so an ambient `GIT_DIR`/`GIT_WORK_TREE` can't silently redirect the check to a different repository.
- `.gitignore` is now a declared `test:hooks` input (`scripts/verify-cache.mjs`), with its own `stepTouchedByDiff` regression test — otherwise a `.gitignore`-only diff prints `test:hooks [cached]` locally and skips the guard's CI leg too, exactly the diff shape most likely to silently drop a secret pattern.
- Release notes updated in both `docs/release-notes-next.md` and `RELEASE_NOTES.md`.

## Test plan
- [x] `scripts/tests/gitignore-secrets.test.mjs` — 11 assertions covering `.env`, `.env.bak`, `.env.bak.N`, `.env~`, `.env.backup`, `.env.old`, `.env.save`, and the `server/.env.example` negative control, all via `git check-ignore --no-index`, env-scrubbed.
- [x] `scripts/tests/verify-cache.test.mjs` — asserts `.gitignore` is in scope for `test:hooks`.
- [x] `npm run test:hooks` — 1424/1424 pass (this PR's own diff touches `test:hooks` and `lint`; the rest of `verify:fast:branch` is legitimately out of scope for a scripts-only change).
- [x] `npm run verify:fast:branch` (pre-push, all 3 commits) — lint pass each time.

Reviewed across 3 rounds (the pr-review-gate loop's cap): round 1 found 3 correctness bugs, round 2 confirmed the fixes and found 2 more (one missing test, one gap in the new patterns' coverage), round 3 confirmed those and closed one latent gap in the test's own git-env handling. Verdict: mergeable.
- https://github.com/dudarenok-maker/Castwright/pull/2532#issuecomment-5353253972
- https://github.com/dudarenok-maker/Castwright/pull/2532#issuecomment-5353445962
- https://github.com/dudarenok-maker/Castwright/pull/2532#issuecomment-5353592390

Closes #2531


dudarenok-maker:
## PR review — pass 1 (head 3f725817a54cd73ac7e4da13930f33a1ee92165e, depth medium)

Scope: `.gitignore` (+1) and `scripts/tests/gitignore-secrets.test.mjs` (+55, new) · verified: ran the new suite (6/6 pass, ~0.4s), reproduced red-before-fix in a throwaway repo seeded from `origin/main:.gitignore` (`.env.bak` and `server/.env.bak` both exit 1 = not ignored, so the two new assertions are red for the reason claimed), and drove `stepTouchedByDiff` / `computeScopes` directly — before adversarially probing whether the guard can actually fail.

The one-line `.gitignore` fix is correct and does what it says. Every finding below is about the *other* half — the test the PR body describes as "the half that keeps this true".

---

### 🔴 Blocking — the new guard never runs on the one diff shape it exists to catch (correctness bug)

`.gitignore` is an input to **no** verify step. Proven mechanically against the repo's own scope machinery:

```
node --input-type=module -e "import {computeScopes} from './scripts/ci-scope.mjs'; ..."

THIS PR         -> step_test_hooks = true
.gitignore-only -> step_test_hooks = false
.gitignore-only -> any leg true?  = false     # all 15 steps + openapi + shared
```

`stepTouchedByDiff(STEPS['test:hooks'], ['.gitignore'])` is `false`, and `computeShared(['.gitignore'])` is `false` (only root `package.json` / `package-lock.json` / `.github/actions/**` are global — `scripts/verify-cache.mjs:737-742`).

So the failure scenario the PR body names verbatim — *"a later reorganisation of that file can't pass while the protection is gone"* — plays out like this: someone edits **only** `.gitignore` and drops line 21. Locally, pre-commit and pre-push print `test:hooks [cached]`. In cloud CI, `ci-scope.mjs` derives its legs from this same `STEPS[]` entry (`scripts/ci-scope.mjs:13,39-43`), so `step_test_hooks=false` and the leg is skipped. The PR goes green having run **zero** legs, and `.env.bak` is stageable again with no gate anywhere.

The guard runs exactly once — on the PR that introduces it — and never again on the diff it was written for. Defect shape 7 ("a fix unreachable at the default configuration — correct code that runs zero times as shipped") and shape 6.

This is this repo's own #1847 runtime-read trap, and `scripts/verify-cache.mjs` already documents **nine** prior instances of it in the `test:hooks` entry alone — `fixtures/**`, `pinokio-scripts/**`, `.github/actions/**`, `.claude/skills/**`, `.claude/agents/**`, `docs/testing/**`, `server/madge-cycles-allowlist.json`, `CLAUDE.md`, `CONTRIBUTING.md`, `.env.mock`/`.env.development`. Most were added *by a prior PR review finding* (#2012, #2053, #2241, #2348, #2375). A new hooks test that reads a file at runtime with no module-graph edge owes its input registration in the same PR.

**Fix:** add `'.gitignore'` to the `test:hooks` step's `inputs.extraFiles` in `scripts/verify-cache.mjs` (the array ending `'.env.mock', '.env.development',` at `scripts/verify-cache.mjs:298-299`), with a comment in the house style naming the runtime read.

---

### 🟠 Significant — the suite's only negative control cannot fail (correctness bug)

`scripts/tests/gitignore-secrets.test.mjs:53-55` is the sole assertion guarding against an **over-broad** pattern. It is inert for two independent reasons, both verified.

**1. Root `.env.example` does not exist in this repo.**

```
exists on disk : False
tracked        : False
```

`git ls-files` shows the tracked example file is `server/.env.example`, not `.env.example`. The test name (`is tracked`) and its assertion message (`'.env.example is an intentional tracked file'`) are both false — defect shape 9.

**2. `git check-ignore` consults the index, so a tracked path can never report as ignored.** `--no-index` is missing from `scripts/tests/gitignore-secrets.test.mjs:18`. Proven on a throwaway repo — pattern `secret.txt` present in `.gitignore`, file tracked:

```
default check-ignore exit=1   (reports NOT ignored)
--no-index          exit=0   (reports ignored — the truth)
```

**Concrete end-to-end break.** Seed a repo with a plausible tightening of the env block — `.env` / `.env.bak` / `.env.local` / `.env.*.local` / `server/.env*`, with `!.env.example` dropped — then track `server/.env.example`:

```
server/.env.example  default (the test's helper) -> ignored=False   <- control stays GREEN
server/.env.example  --no-index (the truth)      -> ignored=True    <- would go RED

.env                       ignored=True
.env.local                 ignored=True
.env.development.local     ignored=True
.env.bak                   ignored=True
server/.env.bak            ignored=True
.env.example               ignored=False
```

**All 6 tests pass** while a tracked, intentionally-committed file has silently become invisible to `git add -A` / `git status`. That is precisely the "reorganisation of that file" the PR body says this test prevents. Defect shapes 3 ("a test that cannot fail") and 8 ("a control group labelled clean but never measured").

**Fix:** assert on `server/.env.example` (the path that is actually tracked), and pass `--no-index` in `isGitIgnored`.

---

### 🟠 Significant — the change's own threat model names three producers; the fix covers one spelling (correctness bug)

The commit message states the threat as *"an editor, a migration script, a hand `cp` before an edit"*. The most common **editor** in-place backup suffix is `~` — vim with `backup`/`writebackup`, and emacs by default, write `server/.env~`. Verified against the branch head:

```
git check-ignore -q --no-index <path>

server/.env.bak.1   ignored=False
.env.bak.1          ignored=False
.env.backup         ignored=False
.env~               ignored=False
.env.old            ignored=False
server/.env.save    ignored=False
```

`.env~` is a full, unredacted copy of `server/.env` (`GEMINI_API_KEY` and friends), untracked and stageable, in a public repo. `.gitignore`'s "Editor / OS" block at line 33 already carries `*.swp` — vim's *swap* file — but nothing for vim's *backup* file, which is the one that contains the plaintext.

`.env.bak.1` deserves separate mention: `.bak.N` is **this repo's own rotating-backup spelling**, implemented in `server/src/workspace/state-io.ts:149-155` (`${path}.bak.${i}`, `renameWithRetry(path, \`${path}.bak.1\`)`) and relied on by `device-tokens.ts` and `auto-backup.ts`. The repo's dominant backup convention is the one shape the new pattern misses.

Defect shapes 2 ("a guard that enumerates syntax — it loses one spelling per round") and 10 ("one instance fixed, the class left armed"). Note a blanket `.env.*` is *not* available as a fix: `.env.development`, `.env.e2e`, `.env.marketing` and `.env.mock` are tracked. Narrow additions (`.env.bak*`, `.env~`, or a scoped `*~`) plus matching assertions cover the class without that collision.

---

### 🟡 Minor — `isGitIgnored` collapses every error exit into "not ignored" (cleanup)

`scripts/tests/gitignore-secrets.test.mjs:22` returns `result.status === 0`. The comment on line 21 enumerates only two of the four outcomes. Measured:

```
outside-repo abs path  exit=128
parent-dir path        exit=128
git binary missing     status=null, error=true
```

128 and `null` both become `false`. For the five positive assertions that is fail-loud; for the negative control at line 53 it is a **silent pass** — the control is green whenever the instrument is broken, which is the same hole as the finding above by a different route.

Related, same helper: `repoRoot = process.cwd()` (line 11) is never verified to *be* the repo root. Run the file from `scripts/` and every assertion silently becomes about `scripts/.env`, `scripts/server/.env.bak` — phantom paths that still match the basename patterns and still pass. `npm run test:hooks` sets cwd correctly today, so this is latent, not live.

**Fix:** throw on `status !== 0 && status !== 1` (and on `result.error`); derive `repoRoot` from `git rev-parse --show-toplevel` or from `import.meta.url`.

---

### 🟡 Minor — the test cites a producer that does not exist (cleanup)

`scripts/tests/gitignore-secrets.test.mjs:49` — `'... (e.g., from POST /api/config/env-cleanup)'`, echoed in the commit body. Grepping the whole tree for `env-cleanup|envCleanup` returns **one** hit: that assertion message itself. There is no such route, and nothing in the repo writes `.env.bak` at all. A future reader chasing the producer finds nothing. Defect shape 9. The PR body's own framing (`cp server/.env server/.env.bak`, an editor, a hand copy) is the honest one — use that.

---

### 🟡 Minor — release-notes pair neither updated nor marked not-applicable (cleanup)

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff. Skipping is defensible here — internal chore, no user- or operator-visible delta — but CLAUDE.md before-shipping step 5 requires saying so explicitly rather than silently omitting, and the PR body says nothing. One line in the body closes it.

---

### ✅ What is solid

- The `.gitignore` fix itself is right, and the pattern placement is right: `.env.bak` with no leading slash matches at any depth, so `server/.env.bak` is covered by the same line — confirmed, not assumed.
- The paired test is a **real** regression test. Seeded a throwaway repo from `origin/main:.gitignore` and confirmed `.env.bak` and `server/.env.bak` both exit 1 there and 0 on this branch, i.e. red before the fix and red *for the reason claimed*, while the four other assertions stayed green across the boundary.
- Driving `git check-ignore` rather than asserting on `.gitignore`'s literal text is the right instrument choice — the findings above are about how it is driven, not about the choice.
- Issue linkage is correct: `Closes #2531` plain, outside any code span; #2531 exists and carries the standalone `bug` label per CONTRIBUTING.md's two-shape convention, so no `docs/BACKLOG.md` row is owed.
- `scripts/tests/*.test.mjs` is the right home — `scripts/run-hooks-tests.mjs:10` globs it, no new dependency, and the Pester runner's `*.Tests.ps1` globs do not collide.
- No on-box acceptance, cast.json lock, config-knob or derived-artifact obligations apply to this diff — checked, not assumed.

### Verdict

**Request changes.** Three correctness bugs. The 🔴 is the one that matters most: as shipped, the regression test is scheduled by neither the local cache nor cloud CI on a `.gitignore`-only diff, so the protection it exists to hold can be deleted with every check green — and registering `.gitignore` as a `test:hooks` input is a one-line fix this repo has already made nine times for the same reason. The two 🟠 findings mean the suite would also stay green through an over-broad reorganisation that un-tracks `server/.env.example`, and that the editor-backup case named in the commit message is still unignored. Fixing any of the three re-triggers a pass under the gate's re-review rule.

dudarenok-maker:
## PR review — pass 2 (head d52bcbcbfecdd3a74274bf68dcd107475be02e06, depth medium)

Deltas only. Every disposition below was re-derived independently, not read off the fix commit's message — mutation harness built in a scratch git repo (`.gitignore` + the test file copied out, `git init`, patterns mutated one at a time), so no tracked file was touched.

---

### Pass-1 dispositions

**1. Blocking — `.gitignore` was not a `test:hooks` input — VERIFIED RESOLVED (mechanism).**
`scripts/verify-cache.mjs:306` adds `'.gitignore'` to the `test:hooks` `extraFiles`. Confirmed live, not by inspection:

```
test:hooks touched by ['.gitignore'] = true
```

and `scripts/ci-scope.mjs:13` imports `stepTouchedByDiff` from that same module, which reads `step.inputs.extraFiles` at `verify-cache.mjs:760` — so the cloud leg widens too, as the comment claims. The mechanism holds. It ships without a paired test, though — see **C1**.

**2. Significant — negative control targeted a nonexistent root file; no `--no-index` — VERIFIED RESOLVED, mutation-proven.**
`server/.env.example` exists and is tracked; root `.env.example` does not exist. `git check-ignore -v --no-index server/.env.example` now reports `.gitignore:34:!.env.example`, i.e. the control is genuinely evaluating the negation rather than short-circuiting on index membership.

That the `--no-index` half has teeth is proven, not assumed. Mutating `.gitignore` to `.env*` with `!.env.example` deleted:

| variant | result |
|---|---|
| current code (`--no-index`) | `.env.example is NOT ignored` → **RED** |
| pass-1 code (`--no-index` removed) | 9 pass / 0 fail → **falsely green** |

**3. Significant — ignore class left armed — PARTIALLY RESOLVED.**
Patterns widened correctly; all six resolve as intended under `git check-ignore -v --no-index`, and `.env.bak.*` does match `rotateBackups()`'s real spelling (`path.bak.{N}`, `server/src/workspace/state-io.ts:134-138`). But only four of the six are actually guarded — see **C2**.

**4. Minor — exit 128 / spawn failure collapsed into "not ignored" — VERIFIED RESOLVED.**
Ran the test file from a non-git cwd: 9 fail with `Error: git check-ignore exited 128 for .env: fatal: not a git repository`. Under the pass-1 code that same state returned `false` for every path, which would have left the **negative control passing for the wrong reason**. Real fix, real failure mode closed.

**5. Minor — nonexistent `POST /api/config/env-cleanup` citation — VERIFIED RESOLVED.** No such string remains in the file.

**6. Minor — release notes — VERIFIED RESOLVED.** `RELEASE_NOTES.md:159` lands in the in-progress `# Castwright 1.15.0` section (above `# Castwright 1.14.0`); `docs/release-notes-next.md:445` carries the technical entry.

Also confirmed green and unrelated-to-findings: full `npm run test:hooks` → **1421 pass / 0 fail**; `npx eslint scripts/tests/gitignore-secrets.test.mjs` → exit 0. Prettier disagrees with the new file, but I measured the control before reporting it: 56 of `scripts/tests/*.test.mjs` and 66 of `scripts/*.mjs` already fail `prettier --check`, and `lint` is `eslint . --max-warnings 0` with no Prettier leg in any STEP or workflow. Not a finding — noted so the next pass doesn't re-derive it.

---

### Correctness bugs

**C1 — Significant — `scripts/verify-cache.mjs:306`: the fix for pass-1's Blocking finding has no paired regression test, and the automated guard that would normally catch that is structurally blind to it.**

`scripts/tests/verify-cache.test.mjs` carries a dedicated `stepTouchedByDiff` test for every comparable runtime-read `extraFiles` literal on this step — `.env.development` (:535), `.env.mock` (:540), `CONTRIBUTING.md` (:488), `pinokio.js` (:512), `eslint.config.mjs` (:522), `bump-version.mjs` (:497), `pip-constraints.mjs` (:517) — and states the convention outright at :709: *"Reverting any one of these five extraFiles entries reddens its own test."* `'.gitignore'` is the first one added without one. Grep for `.gitignore` across every `*.test.mjs` / `*.test.ts` in the repo returns zero assertions outside `gitignore-secrets.test.mjs` itself.

The `test:hooks` completeness guard at `verify-cache.test.mjs:785` does not cover it: it walks the **transitive import closure** of each `scripts/tests/*.test.mjs`, and `gitignore-secrets.test.mjs` imports only `node:test`, `node:assert/strict`, `node:child_process`. There is no module-graph edge to `.gitignore` — that is the whole point of the `#1847` runtime-read trap, and precisely why the other runtime-read literals each got a hand-written test.

Concrete failure: delete `'.gitignore'` from line 306. `npm run test:hooks` stays **1421 pass / 0 fail**. The Blocking defect from pass 1 silently reopens — a `.gitignore`-only diff (exactly the shape that drops a secret pattern) prints `test:hooks [cached]` locally and, because `ci-scope.mjs` derives from the same entry, skips its CI leg too. The PR's own guard becomes unreachable on the only diff it exists to police.

Fix: one assertion alongside its siblings, e.g. `assert.equal(stepTouchedByDiff(stepByName['test:hooks'], ['.gitignore']), true)`.

**C2 — Minor — `.gitignore:26-27` + `scripts/tests/gitignore-secrets.test.mjs`: two of the six patterns this PR adds are unguarded, so a third of the widened class can be deleted silently.**

The test file's own header states its purpose: *"drives `git check-ignore` over the secret patterns rather than asserting on `.gitignore`'s literal text, so a later reorganisation of that file can't pass while the protection is silently gone."* For `.env.old` and `.env.save` it can.

Mutation results, one pattern removed at a time against the real test file:

| mutation | fail |
|---|---|
| drop `.env` | 1 |
| drop `.env.bak` | 2 |
| drop `.env.bak.*` | 1 |
| drop `.env.backup` | 1 |
| drop `.env~` | 1 |
| **drop `.env.old` *and* `.env.save` together** | **0** |

Concrete failure: a future `.gitignore` tidy-up that drops both lines leaves the suite at 9/9 green, and `server/.env.old` — a file a `mv server/.env server/.env.old` before an edit produces, holding `GEMINI_API_KEY` — becomes stageable again with no gate firing. This is the same defect shape as pass-1 finding 3, one level up: the class was widened, the guard was not.

Fix: two assertions mirroring the `.env.backup` case.

---

### Cleanup nits

**N1 — PR body is stale, and its Test plan cites evidence this PR itself proves worthless.**

The body still describes only the original one-line `.env.bak` addition. It does not mention the five additional patterns, the `scripts/verify-cache.mjs` cache-input change, or the release-notes pair — CLAUDE.md requires incidental fixes to be declared in the body (*"an unannounced fix reads as scope creep to a reviewer"*).

More pointed: the Test plan's evidence line reads *"`npm run verify:fast:branch` … `test:hooks` **cached green**"*. A `[cached]` `test:hooks` is exactly the failure this PR's Blocking finding was about — it means the new guard did not run. Citing it as the proof that the guard works is self-refuting, and it is stale besides, since `d52bcbcb` made `.gitignore` a cache input. Replace it with the result of a real run (it is green: 1421/1421, verified this pass).

---

**Verdict: 2 correctness bugs (1 Significant, 1 Minor), 1 cleanup nit.** All six pass-1 findings are resolved in mechanism; both new findings are holes in the *guards* that were added to close them, not in the shipped protection, and both are one-assertion fixes.

dudarenok-maker:
## PR review — pass 3 (head befea3199d1b48e61e4ece8b28ea59818cb87b2e, depth medium)

Deltas only. Both pass-2 findings verified independently by mutation, not by reading the summary.

**C1 (significant) — `.gitignore`-as-`test:hooks`-input had no paired test — RESOLVED.**
Verified by mutation rather than by inspection. Imported the real `STEPS` and dropped *only* the `.gitignore` entry from `test:hooks`'s `extraFiles`:

```
AS SHIPPED  stepTouchedByDiff(test:hooks, ['.gitignore']) -> true
MUTANT      stepTouchedByDiff(test:hooks, ['.gitignore']) -> false
```

Red for the reason claimed: none of the step's nine globs (`scripts/**/*.{mjs,cjs,js,mts,cts,ts}`, `scripts/tests/fixtures/**`, `pinokio-scripts/**`, `.github/workflows/**`, `.github/actions/**`, `.husky/**`, `.claude/skills/**`, `.claude/agents/**`, `docs/testing/**`) matches `.gitignore`, so the `extraFiles` entry is the sole reason, and no other step covers it.

I also checked the two halves the test does *not* assert, because a scope-only fix would leave half the stated problem open:
- **local cache** — `selectStepFiles` (`scripts/verify-cache.mjs:723`) injects `extraFiles` into the hashed set unconditionally, so a `.gitignore` edit moves `test:hooks`'s input hash. Real.
- **cloud leg** — `scripts/ci-scope.mjs:13` imports `STEPS`/`stepTouchedByDiff` from this same module, so the CI leg follows. Real.

And the class: the only other runtime reader of `.gitignore` in the repo is `scripts/lib/module-graph.mjs`'s `classifyIgnored()`, whose sole consumers are `scripts/tests/module-graph.test.mjs` and `scripts/tests/verify-cache.test.mjs` — both `test:hooks`. No second step is left uncovered.

**C2 (minor) — `.env.old` / `.env.save` had zero assertions — RESOLVED, and the whole pattern set is now individually load-bearing.**
Rather than confirm the two new assertions exist, I rebuilt the ignore file in scratch repos and dropped one pattern at a time, isolated from global/system excludes, replaying all 11 assertions:

```
BASELINE (unmutated)  all 11 assertions hold
.env             -> RED: .env
.env.bak         -> RED: .env.bak, server/.env.bak
.env.bak.*       -> RED: server/.env.bak.1
.env.backup      -> RED: server/.env.backup
.env.old         -> RED: server/.env.old
.env.save        -> RED: server/.env.save
.env~            -> RED: server/.env~
.env.local       -> RED: .env.local
.env.*.local     -> RED: .env.development.local
```

Every live pattern has at least one assertion that goes red when it alone is dropped — no pattern can be deleted silently any more. (`!.env.example` is the one line with no red-on-drop, correctly: it is inert, since nothing matches `.env.example` in the first place. See N-2.)

Separately confirmed `rotateBackups()` really does spell it `.bak.N` — `server/src/workspace/state-io.ts:149-155` builds `${path}.bak.${i}` — so `.gitignore:22`'s citation is accurate.

**N1 — stale PR body — RESOLVED**, with one residual inaccuracy (N-1 below). `Closes #2531` present and outside any code span. Release-notes pair present and in the right sections (`RELEASE_NOTES.md`'s in-progress `# Castwright 1.15.0`, `docs/release-notes-next.md`'s 1.15.0 register). No on-box acceptance owed, no derived artifact, no config knob, no `cast.json` write.

`npm run test:hooks` — 1424/1424 pass.

---

### New findings

#### 🧹 Cleanup — significant

**1. The new guard resolves `git` against the ambient environment, not against the repo it is asserting about — so it can report green while protecting nothing.** `scripts/tests/gitignore-secrets.test.mjs:24`

```js
const result = spawnSync('git', ['check-ignore', '-q', '--no-index', filePath], {
  cwd: repoRoot,
});
```

`cwd` is pinned but the environment is inherited whole. Git resolves the repository from `GIT_DIR`/`GIT_WORK_TREE` **before** it falls back to discovery from `cwd` — the exact defect `scripts/git-env.mjs` was written for (#2169), swept across `scripts/` in #2216, and applied at the repo's *other* `check-ignore` call site, `scripts/lib/module-graph.mjs:164` (`env: scrubGitEnv()`).

Demonstrated, not reasoned about. VICTIM = a repo whose `.gitignore` is `node_modules/\ndist/` — i.e. every env pattern this PR adds has been deleted, the precise regression the guard exists to catch. DECOY = an ordinary repo that ignores env files (any sibling checkout, another Castwright worktree, or the primary checkout). The real test file, copied unmodified into VICTIM:

```
--- AS SHIPPED (no env scrub) ---
CONTROL clean env                -> exit=1  pass=1  fail=10   (correctly red)
CASE    GIT_DIR + GIT_WORK_TREE  -> exit=0  pass=11 fail=0    (FALSE GREEN)

--- with `env: scrubGitEnv()` ---
CONTROL clean env                -> exit=1  pass=1  fail=10
CASE    GIT_DIR + GIT_WORK_TREE  -> exit=1  pass=1  fail=10   (correctly red)
```

11/11 green on a repository with zero env protection. This is the PR's own thesis — *assert on behaviour so a later reorganisation can't pass while the protection is silently gone* — reintroduced one level up: the behavioural assertion passes while the protection is silently gone. (`GIT_DIR` alone stays red; git re-anchors the work tree from `cwd`, consistent with git-env.mjs's own #2216 correction. The pair is what bites.)

**And it lands in a blind spot whose stated rationale this PR just falsified.** `scripts/tests/git-scrub.test.mjs:165-169` excludes `scripts/tests/**` from the unconditional-scrub invariant, and gives its reason:

> these create throwaway git repos as test fixtures under an explicit cwd, a different risk profile than the ambient-env hazard this issue is about

`gitignore-secrets.test.mjs` is the first file in that directory that does *not* do that: it runs `git` against **the real repository under test**, and the entire meaning of every assertion is which repository git resolved. The exclusion's premise no longer holds for one of its members, and nothing says so. Under CLAUDE.md's "Surgical changes", a comment the change made false is a chore, not taste.

Owed in this round, per CLAUDE.md's incidental-findings rule, and cheap: `env: scrubGitEnv()` on the spawn (the helper is two directories up, already imported by siblings), plus a sentence in `git-scrub.test.mjs`'s exclusion note naming the member the rationale no longer covers.

**Why this is filed as cleanup and not as a correctness bug**, stated so it can be overridden rather than taken on trust: the shipped artifact is correct — every `.gitignore` pattern is mutation-proven above — and the guard is correctly red-and-green in every environment it actually runs in. Pre-commit and pre-push do not export `GIT_DIR`/`GIT_WORK_TREE` (git-env.mjs's #2216 correction establishes that an ordinary hook does not), cloud CI does not, and this box's shell does not (`GIT_ASKPASS`, `GIT_EDITOR`, `GIT_TERMINAL_PROMPT` only). The fail-open is latent, not live. It is test hardening against a hazard class this repo has already decided is worth closing everywhere — not a defect in what merges.

#### 🧹 Cleanup — nits

**N-1. PR body's test count is stale by exactly the tests this commit added.** Body: `npm run test:hooks — 1421/1421 pass`. Measured on `befea319`: **1424/1424**. `befea319` added three tests (one `stepTouchedByDiff` case, two `isGitIgnored` cases) and rewrote the body in the same commit without moving the count — the same stale-evidence class as the N1 it was fixing. Same bullet: "`verify:fast:branch` — lint pass; remaining legs out of scope for this diff" is loose — `stepTouchedByDiff` over the branch diff puts **two** legs in scope, `lint` *and* `test:hooks` (all thirteen others correctly out).

**N-2. The negative control's title names the one mechanism the file's own header rules out.** `scripts/tests/gitignore-secrets.test.mjs:95` — `'gitignore: .env.example is NOT ignored (is tracked)'`. Tracked-ness is specifically *not* why it reports not-ignored: `--no-index` exists to stop that being the reason, as lines 13-22 say at length. A stale title from the pre-`--no-index` revision. The same header's parenthetical — "not because `!.env.example` actually matched" — implies the negation is the real mechanism; mutation shows `!.env.example` is inert (dropping it flips no assertion), because no pattern matches `.env.example` at all. The control is still worth keeping: it is what would catch an over-broad widening such as `.env*`. It just does not work the way either string says.

---

### Verdict

**Mergeable.** Both pass-2 findings are genuinely resolved — verified by mutation in both cases, not by reading the diff — and no correctness bug is present in what this PR ships. The `.gitignore` patterns are individually load-bearing, the `test:hooks` input entry is real on both the local-cache and cloud-CI halves, and `test:hooks` is 1424/1424 green.

Findings this pass are cleanup-only and do not re-trigger the loop. Folding finding 1 in before merge is recommended (two lines, test-only, no further round needed).

PR #2562, opened 2026-08-20
fix(server): tolerate surname-token drift in merge-analysis-cast name fallback

### LINKED ISSUE #2536: merge-analysis-cast.ts's exact-name-match fallback mints near-duplicate roster ids on name-token drift
## Summary

`merge-analysis-cast.ts`'s exact-name-match fallback creates near-duplicate roster ids when a character gains or loses a name token between analyzer runs — e.g. `brann-wire`/`berrin-wire` minted alongside the pre-existing `brann-weir`/`berrin-weir`, rather than being resolved to the same character.

## Root cause

`server/src/store/merge-analysis-cast.ts:205-206,282-284`'s name-fallback match requires an **exact** normalized-name string match (via `nameOf()`/`normaliseForMatch`), with no tolerance for a character gaining a surname token between analyzer runs.

## Evidence

Found during wave-3 on-box acceptance testing of register rows B3/B4 (`docs/testing/onbox-wave3-results/step-5-group-b.md`), via a real full re-analysis of *Заказ Коалфолла* through the local Ollama analyzer (`qwen36-cw-iq4-32k:latest`, ~19m43s wall-clock):

- The core `characterId`-drift criterion **passed**: `mairin`/`coalfall-dragon` stayed unchanged in `cast.json` across the re-analysis, and the analyzer's raw fresh ids this run (genuinely different Cyrillic-cased strings, `мэйрин`/`коалфолл`) were correctly caught and retired via `retireCharacterId`.
- But the roster-integrity sub-check ("no duplicate row") **failed for real**: roster size moved 13 → 16. One addition (`unknown-man`) is a legitimate new detection. Two are near-duplicates: `brann-wire`/`berrin-wire` (this run's analyzer output, from "Бранн Уир"/"Беррин Уир") duplicate the pre-existing `brann-weir`/`berrin-weir` (from the cast's existing "Бранн"/"Беррин") — same role, same evidence quotes verbatim, never linked via `retireCharacterId` (absent from `supersededBy`).
- This is exactly the row's own text names as the worst-case outcome: "the one way this change could make things worse rather than better."

Row B4 rides the same run and confirms the same pair as its own "no character gained a second id" failure, with roster size moving from the previously-recorded 13 to 16 (not "still 13" as the row anticipated as the default/neutral outcome).

## Decision owed

Not fixing here — the exact-match fallback's tolerance for name-token drift (e.g. a character gaining a surname between runs) is a design decision with more than one defensible resolution (fuzzy/normalized matching with some similarity threshold, an explicit surname-token-aware comparator, or requiring a human/agent review step before minting a new id when an exact match fails but a near-match exists).

## Source

Found during wave-3 on-box acceptance testing, register rows B3 and B4. Full write-up: `docs/testing/onbox-wave3-results/step-5-group-b.md`.

dudarenok-maker:
## Summary

Adds a local, surname-tolerant name comparator to `merge-analysis-cast.ts`'s
name-fallback path: when a character's normalised name gains or loses exactly
one trailing token between analyzer runs (e.g. "Бранн" → "Бранн Уир"), the
fresh row now resolves onto its existing roster row (voice/reuse carried
over) instead of minting a near-duplicate id. Widening is strictly one
trailing-token, same-leading-prefix — not general similarity — and only
extends *how* a candidate is found; the existing selection rule (voiced
preference, ambiguous->id-only, narrator exclusion, notLinkedTo) is
unchanged. A claim-once guard further refuses tolerant matches when a dropped
row would otherwise be claimed by more than one fresh row (an ambiguity at
row granularity rather than key granularity), falling through to the safer
id-only path. `normaliseForMatch` in `text-match.ts` is untouched.

Closes #2536.

## Test plan

- [x] `npm run typecheck` — green (root `tsc --noEmit` + server `tsc --noEmit -p .`), no `error TS` output.
- [x] `server/src/store/merge-analysis-cast.test.ts` — 60/60 passed, including:
  - reproduction of the exact #2536 shape (Бранн/Беррин gaining "Уир"): no near-duplicate id minted, voice/retirement carried over onto the existing id.
  - negative: names differing by more than one trailing token are NOT merged.
  - negative: same-token-count but reordered (non-prefix) names are NOT merged.
  - negative: a surname-tolerant name shared by >1 fresh row still routes to the id-only path (ambiguity guard unchanged).
  - claim-once guard: a dropped row that would match more than one fresh row (tolerant-vs-tolerant or exact-vs-tolerant collision) is refused and falls back to id-only.
- [x] Mutation check (re-run independently by the verify child, not just claimed): reverted the fallback to exact-match-only (dropped the `?? tolerantCandidateByName.get(key)` clause), re-ran the scoped suite — observed RED: the reproduction test failed with `brann-weir`/`berrin-weir` surviving as unmerged voiced orphans instead of resolving onto `brann-wire`/`berrin-wire`. Restored the fix, re-ran — green again (60/60), and `git diff` against the committed file was empty after restore.

Also fixed, found in passing during this PR's own review: a claim-once guard for the tolerant matcher (blocking genuine double-claims and correcting a false-positive over-refusal the initial guard introduced), a factual correction to the release-notes description of the comparator, and three stale in-file comment line-citations pointing to outdated line ranges.


dudarenok-maker:
## PR review — pass 1 (head bcb4e63511e1185529f6a1f19f22a6d770522358, depth medium)

Scope: `server/src/store/merge-analysis-cast.ts` (+58/-1) and `server/src/store/merge-analysis-cast.test.ts` (+117/-0) — the whole diff.

Verified before probing, independently rather than from the PR body: the PR-head test file run in isolation against a PR-head dependency closure — **56/56 pass**. The mutation the PR body claims (drop the `?? tolerantCandidateByName.get(key)` clause) re-run — **1 failed / 55 passed**, and the failure is the #2536 repro at `merge-analysis-cast.test.ts:265`, i.e. red before the fix *for the reason claimed*. A second, independent mutation (delete the leading-prefix loop inside `surnameTolerantMatch`) is **also** killed by that same test, so the prefix guard is genuinely covered and not a clause that cannot fail. `normaliseForMatch` was checked directly (`util/text-match.ts`): it lowercases and collapses `\s+` to a single space, so the comparator's `split(' ')` premise holds.

---

### 🔴 Blocking — correctness bug: one prior roster row can be claimed by MANY fresh rows through the tolerant path, welding one character's designed voice onto several of them and writing contradictory retirements for the same `from`

`server/src/store/merge-analysis-cast.ts:317-333` (tolerant map build) consumed at `:355-356`.

The exact path is **structurally** one-prior-row-to-one-fresh-row and cannot do this: each dropped row lives in exactly one `droppedByName` bucket, each bucket contributes at most one candidate to `dropMatchCandidateByName` (`:294-302`), and a key is only used when `freshNameCounts.get(key) === 1`. So a prior row is reachable from at most one fresh row. The tolerant path drops that invariant — `surnameTolerantMatch` is evaluated against *every* dropped row for *every* still-unresolved fresh key, so a single prior row can be written into `tolerantCandidateByName` under an unbounded number of distinct keys, and there is no claim-once guard anywhere between `:317` and `:372`.

**Repro A — one prior row, two fresh rows that both extend it.** Run against this branch's `mergeAnalysisResultWithExistingCast`:

```ts
const existing = [
  { id: 'mairin', name: 'Мэйрин', voiceState: 'tuned',
    overrideTtsVoices: { qwen: { name: 'qwen-mairin' } } },
];
const fresh = [
  { id: 'mairin-weir', name: 'Мэйрин Уир',  lines: 30 },  // the same person
  { id: 'mairin-cole', name: 'Мэйрин Коул', lines: 4  },  // a different person
];
```

Observed output on `bcb4e63`:

```
merged ids  : ["mairin-weir","mairin-cole"]
voices      : [["mairin-weir",{"qwen":{"name":"qwen-mairin"}},"tuned"],
               ["mairin-cole",{"qwen":{"name":"qwen-mairin"}},"tuned"]]
retirements : [{"from":"mairin","to":"mairin-weir"},
               {"from":"mairin","to":"mairin-cole"}]
```

Two distinct characters now carry the *same* bespoke Qwen voice and the same `voiceState: 'tuned'`.

**Repro B — the prior row is exact-matched by one fresh row and tolerant-matched by another.** The `:313-314` comment ("An exact match is preferred: a fresh name that already resolves via `dropMatchCandidateByName` is skipped here entirely") is true per *key* and false per *row* — a row already consumed by an exact key is still eligible for a different tolerant key:

```ts
const existing = [{ id: 'brann', name: 'Бранн', voiceState: 'tuned',
                    overrideTtsVoices: { qwen: { name: 'qwen-brann' } } }];
const fresh = [{ id: 'brann-a',      name: 'Бранн',     lines: 20 },
               { id: 'brann-weir-b', name: 'Бранн Уир', lines: 3  }];
```

```
voices      : [["brann-a",{"qwen":{"name":"qwen-brann"}}],
               ["brann-weir-b",{"qwen":{"name":"qwen-brann"}}]]
retirements : [{"from":"brann","to":"brann-a"},{"from":"brann","to":"brann-weir-b"}]
```

**Downstream damage, traced to the write.** `analysis.ts`'s `recordRetirements` (`server/src/routes/analysis.ts:291-296`) loops the list and calls `retireCharacterId` per entry; `refuseRetirementsOfLiveIds` only filters a `from` that is still live, which `mairin` is not, so both entries survive. `retireCharacterId` then does an unconditional `history.supersededBy[from] = resolvedTo` (`store/cast-id-history.ts`, the "Add/update the new mapping" line) — **last write wins**. The book ends with `supersededBy["mairin"] = "mairin-cole"`, so every already-frozen segment rendered under `mairin` resolves through `buildCastResolver` onto the 4-line bystander rather than onto `mairin-weir`, who is actually the same person. The first retirement is lost silently — no log line, since nothing was refused.

This is not a corner case for the fixtures this repo actually runs on. Russian rosters routinely hold a given name that is a leading prefix of two different full names (`Егор` → `Егор Уланов` and `Егор Тимофеевич`; `Борис` → `Борис Игнатьевич` and `Борис <surname>`), and the analyzer adding a surname token to *some* rows between runs is precisely the drift #2536 documents.

It is also the exact hazard this change was supposed to be immune to. The new block asserts at `:310-311` that it "widens only HOW a candidate key is found, never the selection rule once candidates are found" — the selection rule's load-bearing property was not the two-branch voiced/lone-row test, it was the one-row-one-claim invariant the exact keying gave it for free, and the widening breaks it. `notLinkedToId`'s own comment two hundred lines up states the standard: "a false 'linked' is silent data corruption (two people collapsed into one), a false 'not linked' only costs a match a user can still do by hand." Register row **B4** names the same outcome as "the one way this change could make things worse rather than better."

Direction only, not prescriptive and not applied: the file's existing doctrine is *refuse when ambiguous*. A prior row that tolerantly matches more than one still-unresolved fresh key is ambiguous by that standard and should yield no match at all, and a row already sitting in `dropMatchCandidateByName`'s value set should be off the table for the tolerant pass entirely. Whatever shape the fix takes, both repros above want a paired test — neither is covered today.

---

### 🟡 Minor — cleanup: the diff staled two intra-file line citations that were accurate on `main`, and one now points into the code this PR added

`server/src/store/merge-analysis-cast.ts:217` and `:353`.

Both were correct before this PR and are wrong after it, because the 58 inserted lines sit above them:

| Citation | Target on `main` (49da1ec) | Where it lands on `bcb4e63` |
|---|---|---|
| `:281-305` (line 217) | `if (nameFallback && !old) {` — the name-fallback block it describes | the `droppedByName` narrator-exclusion comment; the real block moved to `:338-374` |
| `:323-332` (line 353) | the narrator-name carry-forward block | **inside this PR's own `tolerantCandidateByName` loop** (`:317-333`); the real block moved to `:386-392` |

The second is the worse of the two: a reader following it lands on the new tolerant matcher while the prose tells them it "carries the narrator name forward on its own path". `:534` (line 270) rides along in the same class — it was ~5 lines off on `main` but still inside `dedupePriorCastByName`; at `bcb4e63` it lands in `seedReuseGuardsFromPriorCast`, a different function, with `isNarrator` now at `:596`.

This file already carries the precedent, at `:175-177`: *"cited by symbol, not line: this comment's own line citations were 110 lines stale, F3, #2163."*

### 🟡 Minor — cleanup: the release-notes pair is missing with no explicit not-applicable

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the changed-file set, and neither mentions this change (grepped for `surname` / `near-duplicate` / `2536` / `2562` — no hits). The PR body's Test plan does not mark the step not-applicable either.

The delta is user-visible and describable without jargon: a character whose name the analyzer writes down with a surname on one run and without it on the next keeps their designed voice and their place in the cast, instead of turning up twice. CLAUDE.md before-shipping step 5 wants both files moved in this PR rather than reconstructed at cut time.

---

### ✅ What is solid

- **The paired test is a real regression test.** Independently re-run, not taken from the PR body: 56/56 green at head; the tolerant-fallback mutation goes red on exactly one test — the #2536 repro — and green everywhere else.
- **The prefix guard is genuinely covered.** Deleting the `shorter[i] !== longer[i]` loop was expected to survive the new tests and does not: the two-prior-row fixture in the repro test turns the mutant's over-matching into an ambiguous set, which fails the test. Coverage is indirect but real.
- **The negative tests are the right negatives** — >1 trailing token, and same-count reordering — and both are rejected by the length check before the prefix loop, which is the correct ordering.
- **The narrator exclusions still hold on the new path.** `droppedByName` excludes `NARRATOR_CHARACTER_IDS` before the tolerant loop ever sees a row, and the fresh-side exclusion at `:354` is upstream of the new `??`, so neither direction of the #2040 L1 hazard is reopened.
- **The `notLinkedTo` veto still applies** to a tolerant candidate — it sits below the `??`, not above it.
- **`normaliseForMatch` is untouched**, as the PR body claims; the comparator is local to this file and reachable only from `mergeCore`'s name-fallback.
- **`Closes #2536`** is present in the body, plain text, outside any code span.
- **No `cast.json` write, no config knob, no derived artifact** is in scope — this module is pure and synchronous by design, and the diff keeps it that way.
- **On-box acceptance is already recorded.** Register rows **B3** and **B4** carry this exact failure as STILL OWED with the `brann-wire`/`berrin-wire` evidence and a pointer to #2536, so no new row is owed by this diff. (Worth a thought at merge time: after this lands, those two notes describe the root cause as unfixed.)

### Verdict

**Do not merge as-is.** The 🔴 is a newly-introduced silent-identity-corruption path — one designed voice landing on two characters, and one of two same-`from` retirements overwritten in `cast-id-history.json` with no log line — in the one module whose whole job is to stop exactly that. Both repros are three-line fixtures against the real exported function; both want paired tests alongside the fix. The two 🟡s are cheap chores in the same file and the same round.

Tree check: `git rev-parse HEAD` and `git status --porcelain` unchanged across this pass — no tracked file touched.

dudarenok-maker:
## PR review — pass 2 (head 25e9896572a230aead50fba7153ecc920e6eacb4, depth medium)

Depth re-derived from the current commit set: three single-scope `fix(server)` + three `docs` commits; the highest tier any one commit earns is a single-scope `fix` → **medium**, unchanged.

Scope: the four fix/docs commits since `bcb4e63` — the claim-once guard (`merge-analysis-cast.ts:350-367`), its three tests, the release-notes pair, the three citation edits.

**Verified independently, not from the commit messages or the PR body.** I rebuilt both pass-1 repros as direct calls to the exported `mergeAnalysisResultWithExistingCast`, and ran every fixture against *two* copies of the head module — one verbatim, one with lines 350-367 (the whole guard) excised — so every behavioural difference is attributable to the guard and nothing else. Then a 60,000-iteration randomised sweep over rosters up to 4 prior × 4 fresh rows (mixed voiced/unvoiced, ~15% id-stable fresh rows, ~12% carrying `notLinkedTo`), asserting *no two retirements share a `from`* and *no prior row's `overrideTtsVoices` lands on two output rows*:

| variant | duplicate `from` | same voice on 2 rows | total retirements |
|---|---|---|---|
| head `25e9896` | **0** | **0** | 20,848 |
| head minus the guard | 314 | 157 | 21,774 |

Suite: `merge-analysis-cast.test.ts` **59/59 pass** (the PR body still says 56 — see 🟡 below). Server `tsc --noEmit -p server` → exit 0, root `tsc --noEmit` → exit 0, both run here rather than read off the body. All 15 PR checks green.

---

### 🔴 Blocking (pass 1) — one prior row claimed by many fresh rows → **VERIFIED RESOLVED**

Both collision shapes are closed, and closed *by the guard*, not incidentally:

| pass-1 repro | head `25e9896` | head minus guard |
|---|---|---|
| **A** — prior `mairin`/"Мэйрин" (tuned, `qwen-mairin`); fresh `mairin-weir`/"Мэйрин Уир" + `mairin-cole`/"Мэйрин Коул" | `retirements: []`, neither fresh row inherits the voice, `mairin` carried forward as a 0-line orphan | `[{from:mairin,to:mairin-weir},{from:mairin,to:mairin-cole}]`, **both** rows carry `qwen-mairin` |
| **B** — prior `brann`/"Бранн" (tuned); fresh `brann-a`/"Бранн" (exact) + `brann-weir-b`/"Бранн Уир" (tolerant) | `[{from:brann,to:brann-a}]` — exactly one, exact wins | `[{from:brann,to:brann-a},{from:brann,to:brann-weir-b}]` |

Repro B is order-independent (verified with the fresh array reversed). The row-level counting also holds beyond two: three fresh rows all extending one prior row → all three refused, and the 60k sweep found **zero** residual double-claims of any shape. The three tests added by `6434791e`/`2dc23193` are real regression tests — the same fixtures go red on the guard-excised copy, for the reason claimed.

I also checked the specific worry that `2dc23193`'s `freshNameCounts.has(key)` seeding might be decorative: replacing it with `if (true)` **kills the core #2536 merge** (prior "Бранн" + lone fresh "Бранн Уир" → `retirements: []`). The gate is load-bearing and correctly placed.

### 🟡 Stale line citations → **PARTIALLY RESOLVED — 1 of 3 correct, 2 replaced with a *different* wrong referent**

See the new 🟡 below. `:270`'s `:631` is right; `:217` and `:388` are not.

### 🟡 Missing release-notes pair → **VERIFIED RESOLVED**

Both files move in this PR. I read `surnameTolerantMatch` (`:203-214`) against `6b474560`'s corrected text and the examples now hold: "Anton"/"Anton Petrov" match (identical leading token, one extra trailing token); "Anton"/"Simon" do not; ">1 extra token" and mid-name changes do not. The superseded draft's "strips trailing surname tokens from both names and compares what remains" described a mechanism that does not exist, and its "'Anton' → 'Anton Petrov' … still correctly treats them as different characters" was the exact inverse of the shipped behaviour — both are gone. (The parenthetical reason for "Anton"/"Simon" is the token-*count* check at `:209`, which fires before the leading-token loop; the outcome is right either way.)

---

## New findings

### 🟠 Significant — correctness bug: the claim-once guard counts *candidate-map entries*, not *consumable claims*, so a structurally unconsumable claim silently suppresses a legitimate merge and re-arms the exact #2536 symptom

`server/src/store/merge-analysis-cast.ts:350-367`, seeded at `:351-355` / `:356-358`, consumed at `:389-407`.

A tolerant entry is deleted when its candidate row's id appears more than once across the two maps. But a map entry is **not** a claim: the call site consumes `dropMatchCandidateByName.get(key) ?? tolerantCandidateByName.get(key)` only when *four* further conditions hold, and the count checks none of them.

| # | Condition at the call site | Where | The count's blind spot |
|---|---|---|---|
| 1 | `!old` — the fresh row did not already resolve **by id** | `:373` | an id-stable fresh row's key still counts |
| 2 | `freshNameCounts.get(key) === 1` | `:389` | `:352` seeds on `has(key)`, so a key with **2+** fresh rows counts although `:389` blocks it |
| 3 | `!NARRATOR_CHARACTER_IDS.includes(f.id)` | `:389` | a fresh narrator row's key still counts, though it can *never* consume a candidate |
| 4 | neither `notLinkedTo` direction vetoes the pair | `:399-403` | a claim the user has explicitly refused still counts |

Each produces a phantom that kills a real merge. All four reproduce against the exported function; in every one the guard-excised copy emits **exactly one** retirement, i.e. there was no collision to guard against:

```ts
// P8 — condition 1. `x1` resolves by id, so its tolerant key can never be consumed.
existing = [{ id:'mairin', name:'Мэйрин', voiceState:'tuned', overrideTtsVoices:{qwen:{name:'qwen-mairin'}} },
            { id:'x1', name:'Совершенно Другая' }]
fresh    = [{ id:'x1', name:'Мэйрин Коул', lines:5 }, { id:'mairin-weir', name:'Мэйрин Уир', lines:30 }]
head    → retirements: []        rows: x1, mairin-weir, mairin*   ← the #2536 duplicate is back
noguard → retirements: [{from:'mairin', to:'mairin-weir'}]        ← single claim, no collision

// P7 — condition 4. The user said brann ≠ brann-a, which should NARROW the candidate set.
existing = [{ id:'brann', name:'Бранн', notLinkedTo:['brann-a'], voiceState:'tuned', overrideTtsVoices:{qwen:{name:'qwen-brann'}} }]
fresh    = [{ id:'brann-a', name:'Бранн', lines:20 }, { id:'brann-weir', name:'Бранн Уир', lines:30 }]
head    → retirements: []        rows: brann-a, brann-weir, brann*
noguard → retirements: [{from:'brann', to:'brann-weir'}]

// P6 — condition 2. `freshNameCounts.get('бранн') === 2`, so `:389` already blocks the exact claim.
existing = [{ id:'brann', name:'Бранн', voiceState:'tuned', overrideTtsVoices:{qwen:{name:'qwen-brann'}} }]
fresh    = [{ id:'brann-a', name:'Бранн' }, { id:'brann-b', name:'Бранн' }, { id:'brann-weir', name:'Бранн Уир' }]
head → []   ·   noguard → [{from:'brann', to:'brann-weir'}]

// P9 — condition 3. Proof of the class: a narrator-id fresh row is excluded at `:389`
// unconditionally, so its key is structurally incapable of consuming anything.
existing = [{ id:'mairin', name:'Мэйрин', voiceState:'tuned', overrideTtsVoices:{qwen:{name:'qwen-mairin'}} }]
fresh    = [{ id:'narrator', name:'Мэйрин Коул' }, { id:'mairin-weir', name:'Мэйрин Уир' }]
head → []   ·   noguard → [{from:'mairin', to:'mairin-weir'}]
```

**Rate, measured rather than asserted:** across the 60,000 randomised rosters, **455 (0.76%)** hit a *phantom suppression* — the guard-excised copy produced no duplicate `from` at all (so no collision existed) yet the head refused a merge it made. That is 455 books where a character's designed voice is stranded on a 0-line duplicate and the fresh row re-records from scratch: precisely the failure this PR exists to remove, and precisely what register rows **B3**/**B4** describe.

It fails *safe* — no corruption, no contradictory `supersededBy` write, which is why this is 🟠 and not 🔴. But the outcome is the pre-PR bug, silently, and condition 1 is the one that matters in production: **most fresh rows in a real re-analysis keep their id**, so the phantom source is the common case, not the exotic one. It bites exactly the roster shape pass 1 called out as routine — a Russian given name that is a leading prefix of two full names (`Егор` dropped; fresh `Егор Уланов` id-stable and `Егор Тимофеевич` new) — where the id-stable row is *already accounted for* and cannot be the dropped row's continuation.

Note where this came from: it is `6434791e`/`2dc23193`'s own unaudited input. The guard is correct about *which rows collide*; what it never checked is whether the claims it is counting can happen at all.

**Direction only — not prescriptive, not applied, and I did not touch the file.** Counting only claims that survive all four conditions closes it without weakening anything. I built that variant in a scratch copy and re-ran the same 60,000 rosters:

| variant | duplicate `from` | same voice on 2 rows | phantom suppressions recovered |
|---|---|---|---|
| head | 0 | 0 | — |
| consumable-claims-only | **0** | **0** | **455 of 455** |

It leaves 470 retirements below the guard-excised copy — the genuine collisions, still correctly refused — and both pass-1 repros keep their current, correct outcomes. So the strictness is not buying safety; it is pure loss. Whatever shape the fix takes, all four conditions want a paired test; none is covered today. (Condition 2 alone is a one-token change, `has(key)` → `get(key) === 1`, but it recovers only 1 of the 455 — the other three conditions are where the loss is, so that tweak on its own would look like a fix and close almost nothing.)

### 🟡 Minor — cleanup: two of the three "fixed" citations now point at the wrong construct

`25e9896` corrected the staleness and lost the referent on 2 of 3. Verified by locating each target in the current file rather than trusting the commit:

| line | citation now | what is actually there | correct target at this head |
|---|---|---|---|
| `:217` | `:233-302` | the descriptive comment + `freshNameCounts` / `droppedByName` / `dropMatchCandidateByName` construction — all of which run **unconditionally**, `nameFallback === false` included | `:373-409`, the `if (nameFallback && !old)` block. The sentence's claim is "`nameFallback` gates [this]"; it does not gate `:233-302`. On `main` the same sentence cited `:281-305` = `let old = byId.get(f.id); if (nameFallback && !old) {` — the right construct |
| `:270` | `:631` | `const isNarrator = …` at exactly line 631 | ✅ correct |
| `:388` | `:268-282` | the narrator **exclusion** from `droppedByName` — a `continue` that carries nothing forward | `:417-427`, `merged.name = old.name` under "Carry forward a non-default prior name". On `main` this cited `:323-332` = that same carry-forward block |

`:388` is the worse of the two: the prose says the cited block "carries the narrator name forward on its own path" while the cited block's entire job is to *keep narrator rows out*. A reader checking the exclusion's safety argument lands on the exclusion itself and finds a circular justification.

This round is also the argument for the file's own precedent at `:173-174` — *"cited by symbol, not line: this comment's own line citations were 110 lines stale, F3, #2163"*. A line re-cite went wrong twice inside a single three-line commit; symbols (`mergeCore`'s `nameFallback` branch, the `merged.name = old.name` carry-forward, `dedupePriorCastByName`'s `isNarrator`) cannot go stale on the next insertion.

### 🟡 Minor — cleanup: the PR body and the technical release note both describe the pre-guard diff

- The Summary still asserts the change "only extends *how* a candidate is found; the existing selection rule (voiced preference, ambiguous->id-only, narrator exclusion, notLinkedTo) is unchanged." After `6434791e`/`2dc23193` that is false — a new **row-level** ambiguity refusal was added on top of the per-key one, and it changes outcomes for inputs the tolerant matcher never touches (P6 above has no tolerant merge to make and still behaves differently from `main`).
- The Test plan says **56/56**; the current count is **59/59**, and the mutation evidence quoted covers only the original `?? tolerantCandidateByName.get(key)` clause, not the guard.
- No "Also fixed, found in passing: …" line declares the guard, per CLAUDE.md → Incidental findings. Right now the two `fix(server)` commits read as unannounced scope creep to anyone reading the body alone.
- `docs/release-notes-next.md`'s entry is the technical register of what shipped, and it stops at the comparator. It does not record that a prior row reachable from more than one fresh name is now refused outright — the behaviour that decides whether a reader in six months thinks #2536 is fully closed. (The user-facing `RELEASE_NOTES.md` line is fine as written; this is about the technical entry.)

---

### Verdict

**Do not merge as-is.** The pass-1 🔴 is genuinely, verifiably closed — 0 double-claims in 60,000 randomised rosters, and I could not construct a third collision shape. But the fix that closed it over-refuses: it counts claims that four separate call-site conditions make impossible, and in 0.76% of sampled rosters that silently reinstates the duplicate-cast-member symptom #2536 was filed for. A scratch variant counting only consumable claims recovers all 455 with zero regressions on either pass-1 repro, so this is a precision bug in the guard, not a safety/coverage tradeoff. The two 🟡s are same-file, same-round chores.

Tree check: `git rev-parse HEAD` and `git status --porcelain` unchanged across this pass, in both the primary checkout and the `wt-2536-surname-comparator` worktree — no tracked file touched. All probe variants were written to the OS temp directory.

dudarenok-maker:
## PR review — pass 3 (head f19cf086ce48eb7ce6dff75d77f2e361f2215d46, depth medium)

Depth re-derived from the current commit set: five `fix(server)` + three `docs` commits; the highest tier any single commit earns on its own is a single-scope `fix` → **medium**, unchanged from passes 1 and 2.

Scope: the two commits since `25e9896` — `84e316bb` (the `consumable()` rewrite of the claim-once guard + one new test) and `f19cf086` (three citation corrections + the release-note sentence).

**Method — a fresh harness, not the one used in pass 2.** I did not re-run the earlier probe or reuse its variants. I compiled the *real* module with the repo's own `tsc` out of the worktree at `f19cf086` into four independent builds — head, head-minus-guard (guard block excised, anchors asserted), pass-2 head `25e9896`, and `origin/main` — then ran three property suites over them. Every property carries a **positive control**, because a zero from an instrument that cannot fail is not evidence.

Properties: **P1** no two retirements share a `from`; **P2** no prior row's `overrideTtsVoices` lands on two output rows; **P3** *(precision)* if the guard-excised build had no collision, head's retirement set must be identical — an oracle that never looks at the four gates, so it cannot inherit the guard's own blind spots; **P4** the `nameFallback=false` entry point (`overlayInterimCastForLiveView`) is byte-identical.

| build | P1 dup-`from` | P2 voice-dup | P3 phantom | guard fired | genuine collisions |
|---|---|---|---|---|---|
| head-minus-guard *(pass-1 shape)* | **10,904** | **9,078** | 0 | 0 | 10,904 |
| pass-2 head `25e9896` | 0 | 0 | **8,818** | 19,722 | 10,904 |
| **head `f19cf086`** | **0** | **0** | **0** | 10,904 | 10,904 |

*(exhaustive run shown; figures below repeat across all three sweeps)*

Corpus: **505,937 rosters** — 200,000 random (≤5×5), 200,000 random (≤8×8), and **105,937 exhaustive** (all *ordered* selections up to 3 prior × 3 fresh over pools built to hit every axis: prefix relationships, voiced/unvoiced, shared names, narrator ids on both sides, `notLinkedTo` in both directions, id-stable fresh rows, empty and missing names, 3-token names). Head: **P1=P2=P3=P4=0 in all three.** Both controls fire hard (unguarded trips P1 7,170× / P2 4,144× on random too; pass-2 head trips P3 3,322× on random), so the zeros are real readings, not a dead instrument.

The sharpest single number: in every sweep **`guardFired` is exactly equal to `noguardDup`** (7,170 / 16,545 / 10,904). The guard fires *if and only if* a genuine collision exists — not approximately, exactly.

Suite: `merge-analysis-cast.test.ts` **60/60**, run here. Server `tsc -p server` emits with zero errors in-module.

---

### 🔴 Blocking (pass 1) — one prior row claimed by many fresh rows → **VERIFIED RESOLVED**

Zero duplicate-`from` retirements and zero shared-voice rows across 505,937 rosters, including an exhaustive enumeration of the small space where this shape lives. The control proves the assertions are live: the same corpus against the guard-excised build trips P1 10,904× and P2 9,078×.

### 🟠 Significant (pass 2) — guard counted candidate-map *presence*, not *consumption* → **VERIFIED RESOLVED, exactly**

`consumable()` (`server/src/store/merge-analysis-cast.ts:366-375`) mirrors all four call-site gates correctly. Read side by side against `:399-435`: gate 2 `freshNameCounts.get(key) !== 1` ↔ `:417`; gate 1 `byId.get(f.id)` ↔ `:400-401`; gate 3 narrator ↔ `:417`; gate 4 both `notLinkedTo` directions ↔ `:427-431`. `freshByKey`'s first-wins seeding is safe because it is only ever read after gate 2 has established the key has exactly one fresh row.

P3 reads **0** on head and **8,818** on the pass-2 build over the same corpus — the instrument that would have caught this finding detects it, and reports nothing on the fix.

I also checked the PR body's central claim rather than accepting it: measured head against a compiled `origin/main` over 200,000 rosters — **0 rosters where a retirement `main` produced was lost**, 42,131 where a new one was gained, 157,869 identical. Head's retirement set is a strict superset of `main`'s, so "extends only *how* a candidate is found" is now empirically true.

### 🟡 (pass 2) — two of three citations pointed at the wrong construct → **VERIFIED RESOLVED (3 of 3)**

Located each target in the current file rather than trusting `f19cf086`'s message:

| line | citation | target at this head | |
|---|---|---|---|
| `:217` | `:401-437` | `if (nameFallback && !old) { … }` — exactly the gated block | ✅ |
| `:270` | `:659` | `const isNarrator = (c: CastRecord) => …` in `dedupePriorCastByName` | ✅ |
| `:416` | `:445-455` | the narrator-name carry-forward comment + `merged.name = old.name` | ✅ |

### 🟡 (pass 2) — PR body and technical release note described the pre-guard diff → **VERIFIED RESOLVED**

Test count now **60/60** (matches my own run). The "Also fixed, found in passing" line is present and names the guard, the release-note correction, and the citations. `docs/release-notes-next.md:175` now carries the claim-once behaviour ("A dropped character can only be matched to ONE fresh character through this path…"), scoped correctly to the tolerant path — which is accurate, since the guard only ever deletes tolerant entries. The user-facing `RELEASE_NOTES.md:7` line is unchanged and still fine.

---

## New findings

### 🟡 Minor — cleanup/chore: three of `consumable()`'s four gates have no test coverage; compile-clean regressions on any of them ship green

`server/src/store/merge-analysis-cast.ts:366-375`. `84e316bb` added one test (`merge-analysis-cast.test.ts`, "phantom-claim guard: …"), and it is a real regression test — it covers **gate 1** only.

I built one mutant per gate, each a *plausible* regression rather than a deletion, each verified to be a genuine non-no-op edit, and ran the full suite plus P3 over 200,000 rosters against each:

| mutant | strict typecheck | suite | P3 phantom suppressions |
|---|---|---|---|
| gate 1 → drop the `byId` check | clean | **1 failed / 59** ✅ caught | — |
| gate 2 → `!freshNameCounts.has(key)` *(literally the pass-2 bug, one gate deep)* | clean | **60/60 green** | **506** |
| gate 3 → `NARRATOR_CHARACTER_IDS.includes(cand.id)` *(wrong side; call site checks `f.id`)* | clean | **60/60 green** | **828** |
| gate 4 → `notLinkedToId(cand, f.id)` only *(drops the reverse direction the call site checks)* | clean | **60/60 green** | **172** |

All three survivors pass `tsc --noEmit` under the real `server/tsconfig.json` with `noUnusedLocals`/`noUnusedParameters` on, so nothing else in the build would stop them. Each one silently reinstates the exact #2536 duplicate-cast-member symptom in hundreds of rosters per 200,000 — the failure this PR exists to remove — while the suite stays fully green.

This is a chore, not a defect in shipped code: head behaves correctly on every probe I ran. But it is the repo's own named chore shape ("a seam you touched that has no test"), and pass 2 explicitly asked for all four gates to be covered. Three one-fixture tests mirroring the P6 / P9 / P7 repros already written in pass 2's comment would close it.

### 🟡 Minor — cleanup: `84e316bb` introduced two *new* wrong line citations, in the same class this PR has now corrected twice

`server/src/store/merge-analysis-cast.ts:348` and `:367`.

| line | citation | what is actually at that range | correct target |
|---|---|---|---|
| `:348` | "would actually be CONSUMED at the call site (**lines 389-407** below)" | `:389-395` is the guard's *own* `keysToDelete` loop; `:399-407` is the top of the `overlaid` map and the narrator comment | `:399-435` |
| `:367` | "Mirror the exact four gates from the call site (**lines 389-403**)" | contains exactly **one** of the four gates (`:401`); gates 2 and 3 are at `:417`, gate 4 at `:427-431` | `:401-431` |

Both were wrong when written — `f19cf086` changed three lines in place and shifted nothing, so these did not stale, they were never right. `f19cf086`'s sweep corrected the three citations flagged in pass 2 and missed the two its immediate predecessor had just added one commit earlier. The same text is repeated in `84e316bb`'s commit message.

This is the third consecutive round in which a line citation in this file has been wrong, which is the argument the file already makes at `:173-174`: *"cited by symbol, not line: this comment's own line citations were 110 lines stale, F3, #2163."* Citing `mergeCore`'s `nameFallback` branch and its four gates by name cannot go stale on the next insertion; a line range in a comment that sits 40 lines above the code it describes will.

---

## Probed and cleared — not findings

Recorded so the next reader does not re-spend the effort:

- **A fourth collision shape via duplicate prior ids.** My exhaustive enumeration filtered rosters to unique ids, so I deliberately went back and probed the excluded shape. Two prior rows sharing `id: 'p1'` under different names, both exact-matched, *do* produce two retirements from the same `from` — and the guard cannot stop it, since it only deletes tolerant entries. But I compiled `origin/main` and it produces **byte-identical output** on all three variants of that fixture, and both production call sites (`routes/analysis.ts:5745` and `:7359`) pass `remapped.priorCast` from `applyRewriteToPriorCast`, which keys `winners` by canonical id and so guarantees unique ids by construction. Pre-existing, unreachable, and untouched by this PR.
- **`nameFallback=false`.** The guard runs unconditionally but mutates a map that path never reads. P4 confirms byte-identical output across 400,000 rosters.
- **`84e316bb`'s commit message says `Fixes #2562`**, which is this PR, not an issue. GitHub's closing keywords only auto-close issues, so this is inert — flagging it only so it is not mistaken for a second issue link. Not worth a force-push to correct.
- **The base moved mid-review.** `main` advanced to `5526e2ca` (PR #2566) while this pass ran, and #2566 touches both release-notes files this PR also touches. I dry-ran the merge (`git merge-tree --write-tree origin/main f19cf086`) — **clean, no conflicts**. Note CI's green is against the older base.

---

## Verdict

**No correctness bug found.** The pass-1 🔴 and the pass-2 🟠 are both verifiably closed, and closed *precisely* rather than by over-refusing: across 505,937 rosters the guard fires on exactly the collision set and nothing else, and head never withdraws a merge `origin/main` would have made. I attacked it from three independent angles — randomised sweeps at two roster sizes, an exhaustive enumeration of the small ordered space, and a deliberate probe of the shape my own enumeration excluded — and could not construct a fourth failing shape.

Both remaining findings are **🟡 cleanup**, which per the triage rules does not re-trigger a review round. **This closes the loop at the cap with no correctness issue outstanding** — fix the two chores in this round (three gate tests; two citations, ideally by symbol) and merge. No fourth round is required, and no human judgment call is owed.

Tree check: the reviewed worktree `wt-2536-surname-comparator` is unchanged across this pass — `f19cf086` with an empty `git status --porcelain`. All builds, mutants and probes were written to the OS temp directory; no tracked file was touched. **Disclosure:** the *primary* checkout's HEAD moved during the pass (`75e107bd` → `5526e2ca`) and its dirty files went clean. That is a concurrent lane committing and merging PR #2566 on `fix/scripts-windowshide-console-flash` — a different branch, unrelated to this PR — not an edit by this review; I made no write of any kind to that checkout.

Issue #2130, opened 2026-08-05
ops — a resolver tier rename would turn every segment into false damage with the repair suite green


dudarenok-maker:
Found by the pre-merge review of the #2107 fix, 2026-08-05.

## What

`scripts/repair-cast-id-drift.mjs`'s orphan collector branches on the resolver's tier strings:

```js
if (resolution.via === 'exact') continue;
```

Every test resolver in `scripts/tests/repair-cast-id-drift.test.mjs` is **hand-written** and hard-codes those same literals (`'exact'`, `'history'`, `'normalised-id'`, `'normalised-history'`). Nothing in the test file imports the real `buildCastResolver`. `.mjs` is not typechecked.

## Failure scenario

Someone renames a tier in `server/src/store/cast-resolve.ts:28` — say `'exact'` to `'exact-id'` — and updates `cast-resolve.ts` and its own TS tests. `tsc` is happy: the script is `.mjs` and outside typecheck. The repair script's comparison now never matches, so **every rendered segment in every book becomes an orphan**, and after #2107 that means every segment lands on the re-render list.

The entire `repair-cast-id-drift` suite stays **green**, because its fakes were never coupled to the real tier vocabulary — they assert against the same stale literals the producer now fails to match.

This is the wave's recurring shape at the test layer: the fake and the real producer agree on a *copy* of the contract rather than the contract itself, so the test measures something other than what production consumes.

## Not introduced by #2107

Pre-existing. But #2107 raises the stakes: before it, a tier-literal mismatch mostly mis-bucketed ids between `orphans` and `autoReconciled`; after it, the tier check is the *only* thing standing between "no damage" and "re-render everything".

## Acceptance

- The tier vocabulary has one source of truth that both `cast-resolve.ts` and the repair script's tests consume — e.g. the tests build at least one real `buildCastResolver` and assert the script's bucketing against it, rather than only against hand-written fakes.
- Renaming a tier makes a test go **red**, not green.
- Do **not** solve this by adding a second hand-maintained list of tier names in the script — that is the same defect with an extra copy.

## Key files

- `server/src/store/cast-resolve.ts:28` (the `via` union)
- `scripts/repair-cast-id-drift.mjs` (`buildOrphansFromSegments`)
- `scripts/tests/repair-cast-id-drift.test.mjs` (the fake resolvers)

## Benefit

*(technical)* A rename in the resolver can no longer silently turn the whole workspace into false damage with the suite green.

PR #2566, opened 2026-08-21
fix(scripts): hide the console windows spawnSync probes flash on Windows

### LINKED ISSUE #2567: PowerShell/console windows flash open-and-close during verify steps
## Bug

Several verify-pipeline tooling scripts spawn console apps (\`powershell\`, \`git\`, \`nvidia-smi\`, \`pytest\`, \`npx madge\`) via \`spawnSync\` without Node's \`windowsHide\` option. With no console attached to the parent process, each short-lived probe pops a new console window that flashes open and closes.

Reported while running verify steps locally — repeatedly seeing PowerShell windows open and close.

Root cause sites:
- \`scripts/run-powershell.mjs\`'s \`which()\` helper (probes for \`pwsh\` then \`powershell\`)
- \`scripts/verify-cache.mjs\`: \`pesterFingerprint()\`, \`sidecarFingerprint()\`, \`detectGpuContention()\`, \`stagedDiffFiles()\`, \`branchDiffFiles()\`, \`gitFileList()\`
- \`scripts/check-import-cycles.mjs\`'s \`madge\` spawn

## Fix

Add \`windowsHide: true\` to each of these short-lived, output-discarded probe spawns.

dudarenok-maker:
## Summary

User reported PowerShell console windows briefly opening/closing while verify steps ran. Root cause: `spawnSync`/`execSync`/`execFileSync` calls to console programs (`powershell`, `git`, `nvidia-smi`, `pytest`, `npx madge`, `npm`) without Node's `windowsHide` option — with no console attached to the parent process, each spawn pops a new console window that flashes open and closes. `windowsHide` sets Win32 `CREATE_NO_WINDOW`; it does not affect `stdio` streaming (proven both by existing in-repo sites that already combine `stdio: 'inherit'` with it, and by a live `GetConsoleWindow()` measurement during review).

**Scope grew across three review rounds** — what started as 3 files became ~40, plus the CI/cache plumbing needed to actually regression-cover them:

1. **Original 3 files**: `scripts/run-powershell.mjs`, `scripts/verify-cache.mjs`, `scripts/check-import-cycles.mjs` — verify-pipeline tooling firing on every `npm run verify*`/`test:scripts` invocation.
2. **Pass-1 widening (34 more files)**: every other `spawnSync`/`spawn`/`execFile*`/`exec*` call across `scripts/**` (excluding `scripts/tests/`) missing the flag — the repo's own established policy (`server/src/spawn-windows-hide.test.ts`'s docstring: "The fix is `windowsHide: true` on EVERY child_process call"), just not yet applied outside `server/src` and a prior hardcoded floor.
3. **Pass-2 widening (3 more files, all with a real user-facing footprint)**: `server/tts-sidecar/scripts/accelerator-profile.mjs` (GPU-vendor probe on every Windows venv bootstrap/upgrade, ships in the release zip — also renamed its injectable `exec` parameter to `probe` so the guard's bare-call scan doesn't mistake it for a real spawn), `pinokio-scripts/lib/resolve-release.js` (3 git spawns on both the Pinokio install AND update paths), `e2e/global-teardown.ts` (dev-only).
4. **Pass-2 → guard/cache closure**: widened `spawn-windows-hide.test.ts`'s `EXTERNAL_FILES_FLOOR` to actually cover all of the above, and restructured `verify-cache.mjs`'s `test:server`/`test:scripts` inputs so a local diff to any of them busts the cache instead of `[cached]`-skipping the guard's own inputs.
5. **Pass-3 → the CI half of the same gap**: `verify.yml`'s required check runs `vitest run --changed`, which selects by module graph — the guard reads every one of these ~40 files via `readFileSync`, no edge, so a diff confined to them selected **zero tests in CI too** (measured directly). Fixed with `server/vitest.config.ts` `forceRerunTriggers` entries, the same mechanism already used there for `openapi.yaml`.
6. **Pass-3 → release notes**: two of the ~40 files are real end-user install/upgrade flows, not just dev tooling — added entries to both `RELEASE_NOTES.md` and `docs/release-notes-next.md`.
7. **Filed, not fixed here** (design decision owed, not mechanical): #2574 — `EXTERNAL_FILES_FLOOR` stayed a hand-curated list while the CI-trigger/cache-scope fixes above went glob-based; the guard file's own existing comment argues against blanket-globbing that specific list, so widening it further needs its own decision.

No behavior change anywhere — this is purely a Windows console-window-visibility fix. On-box acceptance: not applicable — nothing here needs live GPU/audio/analyzer hardware to verify; the actual Windows behavior (no observable change beyond window suppression) was confirmed via a live `GetConsoleWindow()` probe on this Windows dev box during review, not a claim that needs the acceptance register.

Closes #2567

## Test plan

- [x] `node scripts/run-hooks-tests.mjs` — 1454/1454 passing across all three rounds.
- [x] `server/src/spawn-windows-hide.test.ts` (the regression guard for this defect class) — 4/4, now covering ~40 files (was 9).
- [x] `server/src/tts/accelerator-profile.test.ts`, `server/src/upgrade/apply.test.ts`, `server/src/force-rerun-triggers.test.ts` — all green, with new paired tests for every cache-input/CI-trigger addition.
- [x] Every new test verified non-vacuous via a live reverting mutation (fails under the mutation, restored green) — done for the pass-2 `verify-cache.mjs` globs and the pass-3 `forceRerunTriggers` entries.
- [x] `npm run verify:fast:branch` (pre-push hook, three times across three rounds) — green, including a full production `build` each time.
- [x] This PR went through 3 rounds of the mandatory adversarial `pr-review-gate` pass (the loop cap) — pass 1: [comment](https://github.com/dudarenok-maker/Castwright/pull/2566#issuecomment-5363883531), pass 2: [comment](https://github.com/dudarenok-maker/Castwright/pull/2566#issuecomment-5364193902), pass 3: [comment](https://github.com/dudarenok-maker/Castwright/pull/2566#issuecomment-5364535457). Every correctness finding across all three rounds is fixed and pushed; the one non-mechanical finding (EXTERNAL_FILES_FLOOR's glob-vs-curated-list tradeoff) is filed as #2574 rather than resolved here.


dudarenok-maker:
## PR review — pass 1 (head 0d33fddd, depth medium)

Scope: the full branch diff (`scripts/run-powershell.mjs`, `scripts/verify-cache.mjs`, `scripts/check-import-cycles.mjs`; +9/−2) · verified before probing: `gh pr checks 2566` all green (15/15); the repo's existing `windowsHide` invariant guard `server/src/spawn-windows-hide.test.ts` read in full and its selector re-implemented read-only against the current tree; `node scripts/ci-scope.mjs --files=<this PR's exact file set>` run to derive which CI legs actually executed; `verify-cache.mjs`'s `STEPS[]` scope map, `.github/workflows/verify.yml` gating and `package.json` script wiring read.

The seven added `windowsHide: true` values are each correctly placed and correctly valued. Everything below is about what the change did **not** reach.

---

### 🔴 Blocking 1 — this repo already ships a `windowsHide` regression guard, and none of the three files are inside its net

`server/src/spawn-windows-hide.test.ts` is exactly the mechanism this defect class is supposed to be locked down by. Its out-of-`server/src` scan (`:323`) walks `EXTERNAL_FILES` = `EXTERNAL_FILES_FLOOR` (`:61-71`, a hardcoded list whose own comment reads *"Keep it as-is; only ADD to it"*) ∪ `pipSpawners()` (`:166-180`, which selects `scripts/*.mjs` only when `PIP_SPAWN_RE` (`:81`) matches a literal `-m … 'pip'` argv).

Re-running that selector against this branch:

```
EXTERNAL_FILES the guard actually scans:
  launch.mjs, scripts/start-app-prod.mjs, scripts/restart-after-upgrade.mjs,
  scripts/stop-app.mjs, scripts/run-sidecar-tests.mjs,
  server/tts-sidecar/scripts/{install-whisper,install-qwen3,install-coqui,
  ensure-python312,bootstrap-venv,install-ort,install-torch}.mjs

scripts/run-powershell.mjs     : NOT COVERED
scripts/verify-cache.mjs       : NOT COVERED
scripts/check-import-cycles.mjs: NOT COVERED
```

None of the three spawns pip, so none is selected. **Concrete consequence: delete `windowsHide: true` from any of the seven lines this PR adds and the entire battery stays green.** The flag is unenforced at every site the PR touches.

The PR body's justification — *"No behavior change to assert with a new test"* — is false against this repo's own established regression mechanism for this exact symptom. The precedent for the correct move is written down and was executed once already: `docs/superpowers/plans/2026-08-07-ort-pip-consistency-marker.md:1407`, *"Task 11: Widen the `windowsHide` guard without breaking it"* → *"Expected: FAIL, naming any newly-selected file that lacks `windowsHide`"*, followed by a mutation check that deletes the flag and requires the guard to go red.

Fix: add the three files to `EXTERNAL_FILES_FLOOR` (adding, not globbing — `:56-59` explains why the floor must not become a glob), and land the mutation check the plan doc prescribes. See also Blocking 2 and Significant 3, both of which this fix mechanically surfaces.

**CLAUDE.md → Testing discipline: "Bug fix → ship a regression test that fails before the fix and passes after." There is currently no test at all.**

---

### 🔴 Blocking 2 — the fix skips the loudest window on the two paths the PR's own summary names, including one in a file it edits

Two `spawnSync` calls were left without the flag:

- **`scripts/verify-cache.mjs:1074`** — `spawnSync('npm', ['run', stepName], { cwd, shell: true, env, …{ stdio: 'inherit' } })`. This is how **every** `npm run verify` step is launched — the single hottest spawn on the path the PR summary calls out, in the same file that received 7 of the 9 added lines.
- **`scripts/run-powershell.mjs:53`** — `spawnSync(shell, args, { stdio: 'inherit' })`. This is the actual `pwsh`/`powershell` launch that `npm run test:scripts` exists to perform (`package.json:46`). The PR hides the two `which()` probes at `:30` and leaves the real one.

The PR body's rationale for skipping them — *"doesn't affect the real, output-visible `stdio: 'inherit'` test-runner spawns, which are left untouched"* — is wrong on both halves.

**Why the window still appears.** On Windows a console-subsystem child is given a new console — and therefore a new window — unless the creator passes `DETACHED_PROCESS`, `CREATE_NEW_CONSOLE`, or `CREATE_NO_WINDOW`. `stdio: 'inherit'` populates `STARTUPINFO`'s `hStdInput`/`hStdOutput`/`hStdError`; it never touches `dwCreationFlags`, so it cannot suppress console allocation. `windowsHide: true` is precisely what sets `CREATE_NO_WINDOW` (libuv's `UV_PROCESS_WINDOWS_HIDE_CONSOLE`). Under **this PR's own stated premise** — "the parent process here has no attached console" — an inherit spawn pops a window exactly like a piped one. This is also what `spawn-windows-hide.test.ts`'s header states in prose (`:3-10`).

**Why hiding it costs nothing.** `windowsHide` suppresses the console *window*, not the inherited stdio handles. Three in-repo sites already combine them, output streaming live:

- `scripts/run-sidecar-tests.mjs:74` — `{ cwd: sidecarDir, stdio: 'inherit', windowsHide: true }` (pytest; a file that **is** in the guard's floor, so the guard *requires* the flag there)
- `server/src/upgrade/apply.ts:210` — `{ cwd, stdio: 'inherit', shell: isWin(), windowsHide: true }` (character-for-character the shape of `verify-cache.mjs:1074`)
- `server/tts-sidecar/scripts/install-ort.mjs:382` — `{ stdio: 'inherit', windowsHide: true }`

**Repro.** On Windows, from a parent with no attached console, run `npm run test:scripts`. Before this PR: two probe flashes, then a PowerShell window for the Pester run. After: the two flashes are gone; the PowerShell window remains — and it is the *worse* symptom, up for the whole run rather than milliseconds. Same for `npm run verify`, once per step. Issue #2567 is titled "PowerShell/console windows flash open-and-close during verify steps"; the PowerShell window that verify steps actually open is still open.

**Note the interlock with Blocking 1:** adding the three files to `EXTERNAL_FILES_FLOOR` makes the guard emit exactly these two, and nothing else, from the PR's files:

```
scripts/run-powershell.mjs:53 — spawnSync(...) missing windowsHide: true
scripts/verify-cache.mjs:1074 — spawnSync(...) missing windowsHide: true
scripts/check-import-cycles.mjs: 0 offenders
```

The missing test is not cosmetic — it is the thing that would have caught the incomplete fix.

---

### 🟠 Significant 3 — even once widened, the guard never runs on a `scripts/**` diff (locally `[cached]`, in CI skipped)

`test:server`'s cache inputs are `globs: ['server/src/**']` plus a hand-listed `extraFiles` (`scripts/verify-cache.mjs:358-384`). `spawn-windows-hide.test.ts` reads twelve files that live **outside** `server/src/**`, and not one of them is an input. `scripts/ci-scope.mjs` derives CI legs from that same `STEPS[]` map, so the guard is skipped in the cloud too. Run against this PR's exact file set:

```
$ node scripts/ci-scope.mjs --files="scripts/check-import-cycles.mjs
scripts/run-powershell.mjs
scripts/verify-cache.mjs"
  "step_test_server":false
  "step_test_server_slow":false
  "step_test_scripts":false
  "shared":false
```

So the green "Server tests (fast + slow)" check on this PR **did not execute the `windowsHide` guard**, and would not execute it on any future `scripts/**`-only PR either. Widening the floor without closing this leaves a guard that is green because it never ran.

The fix and its rationale are already written in the same file, one screen up — `verify-cache.mjs:367-370` adds `scripts/repair-cast-id-drift.mjs` to `test:server`'s `extraFiles` for precisely this reason (#2130): *"it lives outside `server/src/**`, so without this line an edit there reports `[cached]` and the contract test never re-runs against it."* `EXTERNAL_FILES` needs the same treatment.

---

### 🟠 Significant 4 — `test:scripts` does not list its own entry point, so this PR's edit to `run-powershell.mjs` was verified by nothing

`package.json:46` — `"test:scripts": "node scripts/run-powershell.mjs scripts/tests/run.ps1"`. But `test:scripts`'s cache inputs (`scripts/verify-cache.mjs:429-446`) are `scripts/lib/**`, `scripts/tests/**`, and `extraFiles: ['scripts/tests/run.ps1']` — **`scripts/run-powershell.mjs` is absent**. Hence `step_test_scripts:false` above: the Pester harness never ran on the PR that changed the launcher that starts it. Break `which()` outright and both the local cache and CI report green.

This is the identical trap `check:cycles` documents and closes for *its* entry point twelve lines earlier — `verify-cache.mjs:424` lists `scripts/check-import-cycles.mjs` in `extraFiles`. `test:scripts` needs `scripts/run-powershell.mjs` for the same reason.

---

### 🟠 Significant 5 — one instance fixed, the class left armed: 53 sites in `scripts/*.mjs`, including the hooks that fire on every commit and push

Applying the guard's own `CALL_RE` scan across `scripts/*.mjs` yields **53** spawn sites with no `windowsHide`. The subset on the exact paths this PR's summary names as the symptom:

| file:line | fires on |
|---|---|
| `scripts/is-docs-only-push.mjs:66`, `:73` | two `git` spawns on **every** `git push` (pre-push hook) |
| `scripts/guard-protected-push.mjs:84` | **every** `git push` |
| `scripts/guard-commit-subjects.mjs:97` | commit path |
| `scripts/run-hooks-tests.mjs:15` | `test:hooks` — inside `verify:fast:branch` (pre-push) and `test:all` |
| `scripts/run-pinokio-tests.mjs:12` | `test:pinokio` — inside `test:all` |
| `scripts/quarantine-health.mjs:922` | `runVitestJson` |

The PR fixed `verify-cache.mjs`'s three `git` probes but left the structurally identical `git` probes in the two pre-push guards, which fire strictly more often than anything it did fix. Note also that `check-import-cycles.mjs:59` — a comment *inside this PR's own diff hunk* — cites `quarantine-health.mjs`'s `runVitestJson` as the idiom it copied; that source site is offender `quarantine-health.mjs:922`.

Per `findings-triage.md`, "it would expand the scope of this PR" and "it's pre-existing" are both void deferral reasons. The remaining ~45 are on-demand dev scripts (`capture-companion`, `sync-wiki`, `stage-marketing-screenshots`, `wt-*`, …) — lower urgency, but they are what a widened guard will name, so decide the sweep boundary deliberately rather than by omission.

---

### 🟡 Minor 6 — release-notes pair neither updated nor declared not-applicable

CLAUDE.md before-shipping step 5 requires an entry in **both** `docs/release-notes-next.md` and `RELEASE_NOTES.md`, *or* an explicit statement that the change has no shippable delta — *"say so explicitly rather than silently omitting."* The diff touches neither file and the PR body says nothing either way. Dev-tooling-only is a defensible skip here; the sentence declaring it is what is missing.

---

### ✅ What is solid

- The seven added `windowsHide: true` values are each syntactically inside the correct options object and semantically correct; `check-import-cycles.mjs:64-70` is complete — 0 offenders remain in that file.
- The probe spawns selected (`which()`, `pesterFingerprint`, `sidecarFingerprint`, `detectGpuContention`, the three git probes, `gitFileList`) are genuinely the right set of unconditional, output-discarded probes, and the summary's account of why they fire on every `verify*` invocation is accurate.
- `Closes #2567` is present, plain, outside any code span; #2567 exists and carries the standalone `bug` label, which is the correct shape per CONTRIBUTING.md's two-shape convention.
- PR title matches the commit convention; single-scope `fix`, correctly scoped `scripts`.
- Diff is genuinely surgical — no adjacent restyling, no orphaned code.

### Verdict

**Request changes.** Two blocking findings. Blocking 1 is the load-bearing one: the repo already owns the regression mechanism for this precise defect, the PR did not use it, and the green CI on this PR is evidence of the gap rather than of the fix (Significant 3 proves the guard leg never ran). Blocking 2 is what that missing guard would have caught on its own — the change leaves the longest-lived console window on both paths its own summary names, on a rationale that is factually wrong about what `stdio: 'inherit'` and `windowsHide` each control.

Recommended order: fix Blocking 1 + Significant 3 together (widen `EXTERNAL_FILES_FLOOR`, add `EXTERNAL_FILES` to `test:server`'s `extraFiles`, land the mutation check) so the guard turns red on its own; then let the red list drive Blocking 2 and Significant 5. Significant 4 and Minor 6 are independent one-liners.

dudarenok-maker:
## PR review — pass 2 (head 2afd2c00, depth medium)

Re-review of the widened diff (37 files, 65 `windowsHide` insertions). Every prior finding was re-checked against the tree, not against the claim. Four of six are resolved; two are partially resolved, and the widening surfaced six new findings.

Verification actually run, so the dispositions below are evidence-backed:

- `npx vitest run src/spawn-windows-hide.test.ts` — 4/4 green.
- `node scripts/run-hooks-tests.mjs` — 1448/1448 green (matches the claim).
- An independent re-implementation of the guard's scanner (`blankCommentsAndStrings` + `CALL_RE` + `extractCallArgs`) run over `scripts/**`, `e2e/**`, `pinokio-scripts/**`, `apps/**`, `server/tts-sidecar/scripts/**`, `launch.mjs` — 141 spawn calls in 74 files.
- A **mutation positive control** per floor file: strip `windowsHide: true` from each and re-scan.
- `stepTouchedByDiff` driven against the live `verify-cache.mjs` module for all 15 guard input files.
- `node --check` on all 37 changed script files, plus a mechanical line-pairing check that every changed line outside `verify-cache.mjs` is *exactly* a `windowsHide: true` insertion.

---

**Blocking 1 (guard didn't cover the 3 touched files) — ✅ VERIFIED RESOLVED.**
Not merely "the test passes". Positive control confirms the guard actually *reaches* the calls in each new floor entry rather than passing vacuously:

| floor entry | spawn calls the scan matches | offenders now | offenders in a `windowsHide`-stripped mutant |
|---|---|---|---|
| `scripts/run-powershell.mjs` | 2 | 0 | **2** |
| `scripts/verify-cache.mjs` | 8 | 0 | **8** |
| `scripts/check-import-cycles.mjs` | 1 | 0 | **1** |

The instrument can fail, and it fails for the reason claimed.

**Blocking 2 (loudest spawns on both named paths untouched) — ✅ VERIFIED RESOLVED.**
`scripts/run-powershell.mjs:53` (the real `pwsh`/`powershell` launch) and `scripts/verify-cache.mjs:1104` (`npm run <step>` in `runStepProcess`) both carry the flag; the independent scan reports 0 offenders in either file.

I also tested the risk this creates rather than assuming it away: adding `windowsHide` to long-running **`stdio: 'inherit'` foreground** spawns (`runStepProcess`, `start-app.mjs:31`, `launch-sidecar.mjs:23`) could in principle detach the child from the parent's console via `CREATE_NO_WINDOW`, which would break Ctrl+C delivery — and CLAUDE.md makes Ctrl+C teardown load-bearing for `npm start`. Measured with a `GetConsoleWindow()` probe: the child returns the **same non-zero console HWND** with `windowsHide: true` as with `false` (5 attached processes both ways). No detachment, no Ctrl+C regression. **Not a finding** — recorded so the next pass does not re-litigate it.

**Significant 3 (guard never re-runs on a floor-file diff) — ⚠️ PARTIALLY RESOLVED.** See finding 7.
All 12 floor entries verified correct against the live module — each returns `true` from `stepTouchedByDiff(test:server, [f])`, and `selectStepFiles` folds `extraFiles` into the hash, so they bust the cache and not just the scope gate. These are correct, not merely plausible-looking. The residual hole is that `EXTERNAL_FILES` is the floor **∪ `pipSpawners()`**, and only the floor half was declared.

**Significant 4 (`test:scripts` omits `run-powershell.mjs`) — ✅ VERIFIED RESOLVED.**
`stepTouchedByDiff(test:scripts, ['scripts/run-powershell.mjs']) === true`. The comment's `package.json:46` citation is accurate (verified: line 46 is `"test:scripts": "node scripts/run-powershell.mjs scripts/tests/run.ps1"`).

**Significant 5 (53/54 sites) — ⚠️ PARTIALLY RESOLVED.** See findings 8, 9, 12.
Confirmed independently: **0 offenders remain anywhere under `scripts/**` outside `scripts/tests/`** — 42 files, 64 spawn calls, all flagged. The sweep itself is clean: `node --check` passes on all 37 files, `verify-cache.mjs` has exactly 8 `windowsHide` mentions for 8 spawn calls (no duplicate keys), and line-pairing proves every changed line outside `verify-cache.mjs` is a pure `windowsHide: true` insertion with **no other option altered anywhere**. What is not resolved is the class outside `scripts/`.

**Minor 6 (release notes) — ⚠️ JUDGMENT MOSTLY REASONABLE, EXECUTION INCOMPLETE.** See finding 13.
Skipping the release-notes pair for 36 dev-tooling files is defensible. The stated premise is not fully accurate, though: `scripts/preflight-ffmpeg.cjs` **is** in the release-zip allowlist (`scripts/build-release-zip.mjs`, "Runtime scripts the deployer invokes"), so one touched file does ship. And CLAUDE.md step 5 requires the not-applicable to be **explicit** — it is stated nowhere in the PR body.

---

### New findings

#### 7. 🟠 Significant — correctness bug: the widened fix ships with a guard covering 3 of 37 files; the class is left armed on the other 34
`server/src/spawn-windows-hide.test.ts:61-77`

`EXTERNAL_FILES = EXTERNAL_FILES_FLOOR ∪ pipSpawners()`. Enumerated against the live tree, that is **15 files** outside `server/src`. This PR fixed **37**. The 34-file remainder — `gh.mjs`, `wt-new.mjs`, `wt-merge.mjs`, `wt-list.mjs`, `bump-version.mjs`, `capture-companion.mjs`, `gen-parser-fixtures.mjs`, `flake-repro.mjs`, `quarantine-health.mjs`, `preflight-ffmpeg.cjs`, `start-app.mjs`, `launch-sidecar.mjs`, `lib/run-command.mjs`, `lib/module-graph.mjs` and the rest — is scanned by nothing.

Concrete failure: delete `windowsHide: true` from `scripts/gh.mjs:74`. `npm run test:all`, `npm run verify`, and the cloud battery all stay **fully green**, and the flashing window is back. That is precisely defect shape 10 in the reviewer brief ("one instance fixed, the class left armed"), and it means the PR's own paired-test obligation under CLAUDE.md's testing discipline is met for 8% of its diff.

Note `pipSpawners()` is `.mjs`-only and non-recursive, so `scripts/preflight-ffmpeg.cjs` (`.cjs`) and `scripts/lib/*.mjs` can never be reached by widening the content filter — they need floor entries or a widened glob.

#### 8. 🟠 Significant — correctness bug: the prod install/upgrade path still flashes a PowerShell window, and the guard's own comment says the file has no spawns
`server/tts-sidecar/scripts/accelerator-profile.mjs:83` and `:214`; false comment at `server/src/spawn-windows-hide.test.ts:85`

`resolveInstallProfile` defaults its probe to `exec: exec ?? ((cmd) => execSync(cmd, { encoding: 'utf8' }))` — no `windowsHide`. On Windows the command is `powershell -NoProfile -Command "(Get-CimInstance Win32_VideoController).Name"`. `bootstrap-venv.mjs:250` calls it with **no `exec` argument**, so the default fires on every Windows venv bootstrap and every upgrade that rebuilds the venv — from Pinokio's console-less Electron runtime and from the fs-1 detached launcher. That is one PowerShell console window flash per install/upgrade, on a path that ships (`server/tts-sidecar/**` is in the release-zip MANIFEST). Same defect at `:214` for `node accelerator-profile.mjs`.

The reason it is invisible is a **false comment in the guard itself**. `spawn-windows-hide.test.ts:85` justifies the `PIP_SPAWN_RE` content filter by naming files "with no spawn call at all (pip-constraints.mjs, **accelerator-profile.mjs**)". `accelerator-profile.mjs` has three. The filter's stated rationale is wrong about one of its own two examples, and that wrong rationale is what keeps a prod-reachable offender out of scope.

**Careful with the fix**: adding this file to the floor as-is makes the guard go red on `:43`, `parseVendorFromProbe(platform, exec(cmd))` — that `exec` is the injected function *parameter*, not a `child_process` call, so `CALL_RE`'s bare-`exec(` match is a false positive there. The fix needs the two real sites patched plus a way past `:43`, not a one-line floor append.

#### 9. 🟠 Significant — correctness bug: the Pinokio install and update paths flash three console windows, guarded by nothing
`pinokio-scripts/lib/resolve-release.js:84`, `:98`, `:101`

`execFileSync('git', ['tag', '--list'], …)`, `execFileSync('git', ['fetch', '--tags', '--force'], { stdio: 'inherit', … })`, and `execFileSync('git', ['checkout', tag], { stdio: 'inherit', … })` — none carry `windowsHide`. `pinokio-scripts/install.js:51` and `update.js:49` run this file as a shell step under Pinokio, whose Electron parent has no console. Three flashing windows per Pinokio install and per update, on a first-class, user-facing install path (plan 218).

No guard reaches it: the floor doesn't list it, and `pipSpawners()` globs only `scripts/` and `server/tts-sidecar/scripts/` for `.mjs`, so a `.js` under `pinokio-scripts/lib/` is outside every scan.

#### 10. 🟠 Significant — correctness bug: `EXTERNAL_FILES`'s `pipSpawners()` half still leaves the guard stale-green locally
`scripts/verify-cache.mjs:391-402`

The new block declares the 12 **floor** entries, but `EXTERNAL_FILES` is floor ∪ `pipSpawners()`. Measured against the live module:

```
NO   server/tts-sidecar/scripts/bootstrap-venv.mjs
NO   server/tts-sidecar/scripts/install-ort.mjs
NO   server/tts-sidecar/scripts/install-torch.mjs
```

Concrete failure: drop `windowsHide` from `install-torch.mjs`'s pip spawn and run `npm run verify`. `test:server` prints `[cached]`, the guard never executes, and a flashing pip console ships — the exact scenario the new comment says it closes ("an edit to any floor file … prints test:server [cached] and the guard sits stale-green"). `test:sidecar` does re-run on that diff, but it runs pytest, not this vitest guard.

The comment's wording ("every one of … EXTERNAL_FILES_FLOOR entries") is literally true and is what makes the gap easy to miss — the guard's input set is not the floor.

#### 11. 🟠 Significant — correctness bug: 13 new cache-input entries land with no test, against this file's own documented convention
`scripts/verify-cache.mjs:391-402`, `:470`; convention at `scripts/tests/verify-cache.test.mjs:721`

`verify-cache.test.mjs` carries one `stepTouchedByDiff` test per extraFiles entry, and says so explicitly at line 721: *"Reverting any one of these five extraFiles entries reddens its own test."* This PR adds 13 entries and **zero** tests. Grepping the test file for any of the 13 new paths returns nothing.

Concrete failure: delete the entire 12-entry block from `test:server` and the `run-powershell.mjs` entry from `test:scripts`. `node scripts/run-hooks-tests.mjs` still reports **1448/1448 pass** — identical to the pre-revert run I just executed. The fixes for prior findings 3 and 4 are therefore unlocked and will rot silently.

This matters more than usual because `hashFile()` (`verify-cache.mjs:598-604`) returns `'__missing__'` for a path that does not resolve. A typo'd entry contributes a constant and is inert forever, looking exactly like a correct one. The per-entry test is the only thing that distinguishes them.

#### 12. 🟠 Significant — correctness bug: the same defect the PR calls its "worst offender" survives one directory over
`e2e/global-teardown.ts:24`

```ts
execFileSync('powershell', ['-NoProfile', '-NonInteractive', '-Command', script], {
  stdio: 'ignore',
  timeout: 30_000,
});
```

Windows-only, output discarded, spawns `powershell` — structurally identical to `run-powershell.mjs`'s `which()` helper, which the PR body names as the worst offender for exactly this reason. Fires after every `npm run test:e2e` on Windows. Missed because the sweep was scoped to `scripts/` and no guard scans `e2e/`. Blast radius is dev-only, unlike findings 8 and 9.

#### 13. 🟡 Minor — cleanup: the PR body describes the 3-file diff, not the 37-file one, and one sentence in it is now false
PR body

The body was not updated when 2afd2c00 widened the change. Three problems:

1. **A false statement.** *"[the fix] doesn't affect the real, output-visible `stdio: 'inherit'` test-runner spawns, which are left untouched."* 2afd2c00 added the flag to many `stdio: 'inherit'` spawns — `verify-cache.mjs:1104`, `run-powershell.mjs:53`, `start-app.mjs:31`, `launch-sidecar.mjs:23`, `build-companion-apk.mjs:114`, `gen-parser-fixtures.mjs:619`, and others. That sentence is the one a merger would rely on to judge the change risk-free. (I tested the underlying risk — see Blocking 2 above — and it is benign, so the harm here is the false record, not a live bug.)
2. **The widened scope is undeclared.** CLAUDE.md requires incidental fixes to be declared in the PR body ("Also fixed, found in passing: …"). 34 files were swept with no mention; a reviewer reading the body sees a 3-file change.
3. **The release-notes not-applicable is not stated.** CLAUDE.md step 5 allows skipping the pair only with an explicit statement. Worth naming `scripts/preflight-ffmpeg.cjs` shipping in the release zip when you write it, so the judgment is recorded honestly rather than on a premise that isn't quite right.

`Closes #2567` is present and outside any code span. Title matches the commit convention and the single-scope `fix` depth ladder.

---

### Summary

| # | Severity | Class |
|---|---|---|
| 7 | 🟠 Significant | correctness bug |
| 8 | 🟠 Significant | correctness bug |
| 9 | 🟠 Significant | correctness bug |
| 10 | 🟠 Significant | correctness bug |
| 11 | 🟠 Significant | correctness bug |
| 12 | 🟠 Significant | correctness bug |
| 13 | 🟡 Minor | cleanup |

Six correctness findings — the re-review trigger fires once these are fixed and pushed.

The through-line: the code fix is clean and mechanically verified, but the widening outran its guard. Pass 1 asked for the guard to cover 3 files; the answer was to fix 37 and still cover 3. Findings 7-10 and 12 are all the same shape at different radii, and finding 11 is that shape applied to the cache-input fix itself.

dudarenok-maker:
## PR review — pass 3 (head 75e107bd, depth medium)

Final round under the loop cap (initial + 2 re-reviews). Every pass-2 finding re-checked against the tree with a positive control, not against the commit message.

**Verification actually run:**

- `node scripts/run-hooks-tests.mjs` — **1454/1454** green (matches the claim).
- `npx vitest run src/spawn-windows-hide.test.ts src/tts/accelerator-profile.test.ts` — **35/35** green.
- **Mutation positive control over all 52 `EXTERNAL_FILES` entries**: strip `windowsHide: true` in memory from each and re-scan with the guard's own lexer + `CALL_RE`. Result: **94 offenders across the mutants, 0 files unprotected**, and 0 offenders today. The widened guard genuinely reaches every file it now lists.
- **Independent in-memory mutation of `verify-cache.mjs`'s `STEPS[]`** (imported live, disk untouched) for all six new `stepTouchedByDiff` assertions — each flips `true → false` when its own glob/`extraFiles` entry is reverted. **None is vacuous.**
- **Repo-wide re-scan** of every `.mjs/.cjs/.js/.ts/.tsx` file importing `child_process`, `fork(` added to the name list — 124 files, 66 unguarded calls.
- `npx vitest related` probes against the server config, with a positive control, to test what the *cloud* gate actually selects.
- `node scripts/ci-scope.mjs --files="scripts/gh.mjs"`.

---

**7 (floor covered 3 of 37 files) — ✅ VERIFIED RESOLVED.** The floor is now 49 entries; `EXTERNAL_FILES` resolves to 52 files and the scan inspects 94 spawn calls across them. The mutation control above is the proof, not the green test: pass 2's concrete repro — *"delete `windowsHide: true` from `scripts/gh.mjs:74` and everything stays green"* — now yields an offender. Every one of the PR's 37 `scripts/**` files, plus `scripts/lib/*.mjs` and the `.cjs`, is reached.

**8 (`accelerator-profile.mjs` prod probe + false guard comment) — ✅ VERIFIED RESOLVED.** `:87` and `:218` both carry `windowsHide: true`. The `exec` → `probe` rename is **complete and has no stale caller**: `resolveInstallProfile`/`detectVendor` are called from exactly one production site, `bootstrap-venv.mjs:250`, which passes **no** probe argument at all and so takes the (now-hidden) default — the "caller still passing `exec:` that the destructure would silently ignore" failure mode does not exist here. `grep` for `exec:` across `server/**` returns nothing but a `.venv` site-package. The file has no residual `exec` identifier (`:29-47`, `:77-87`, `:218`), and the guard's `PIP_SPAWN_RE` comment (`spawn-windows-hide.test.ts:136-147`) is corrected rather than left citing this file as "no spawn call at all".

**9 (Pinokio install/update git spawns) — ✅ VERIFIED RESOLVED.** All three sites carry the flag (`resolve-release.js:87`, `:105`, `:109`); the file is in the floor and in `test:server`'s `extraFiles`. `pinokio-scripts/` contains **no other** `child_process` user — verified by grep across the whole directory, so this is the complete set for that tree.

**10 (`test:server` inputs missed the floor *and* `pipSpawners()`) — ⚠️ **RESOLVED LOCALLY, STILL OPEN IN CI.** See finding 14.** The two globs work exactly as claimed — `globToRegex`'s `**/` → `(?:.*/)?` makes `scripts/**/*.{mjs,cjs,js}` match a file directly in `scripts/`, and I confirmed live rather than by reading the regex: `install-torch.mjs`, `preflight-ffmpeg.cjs`, `lib/module-graph.mjs` and `gh.mjs` all return `true`. The pre-existing whole-floor gap and the `pipSpawners()` gap are both genuinely closed **for `npm run verify` / `verify:fast:branch`**. The half that is not closed is the required cloud gate.

**11 (13 cache-input entries with no tests) — ✅ VERIFIED RESOLVED.** Six new `stepTouchedByDiff` tests, all six independently proven non-vacuous by reverting each entry against the live module. The `hashFile() → '__missing__'` inert-typo hazard pass 2 raised is also closed by construction on the guard side: a typo'd floor entry makes `readFileSync` throw, so it fails loud rather than silently.

**12 (`e2e/global-teardown.ts`) — ✅ VERIFIED RESOLVED.** `:27`. In the floor, in `test:server`'s `extraFiles`, and `e2e/` contains no other `child_process` user.

**13 (PR body accuracy) — ⚠️ PARTIALLY RESOLVED.** The body now describes all three rounds accurately, the false `stdio: 'inherit'` sentence is gone, and `Closes #2567` is present and plain. But the release-notes not-applicable it added is stated on a premise the pass-2 widening falsified — see finding 15.

---

### New findings

#### 14. 🔴 Blocking — correctness bug: the guard still never runs at the **required** gate on a `scripts/**`-only diff. The fix closed the local layer and left the cloud layer exactly as blind as before.
`.github/workflows/verify.yml:303-315` · `server/vitest.config.ts:176-190` · `server/src/force-rerun-triggers.test.ts:98-109`

Pass 1's finding 3 named two halves — *"locally `[cached]`, in CI skipped"*. The fix addressed the local half and, for the CI half, addressed the wrong layer. `ci-scope.mjs` now does route the leg:

```
$ node scripts/ci-scope.mjs --files="scripts/gh.mjs"
  "step_test_server":true
```

So the **job** fires. But on a PR the job does not run the suite — it runs `--changed`:

```yaml
# .github/workflows/verify.yml:309-311
if [ -n "$BASE" ]; then
  npx vitest run --changed "$BASE"
```

`--changed` selects test files by **module graph**. `spawn-windows-hide.test.ts` reads all 52 of its inputs with `readFileSync` at runtime (`:243`, `:340`, `:357`) — there is no import edge to any of them. Measured, not reasoned:

```
$ cd server && npx vitest related ../scripts/wt-new.mjs --run
  No test files found, exiting with code 0

# positive control, same command shape:
$ cd server && npx vitest related src/tts/mp3.ts --run
  Test Files  90 passed (90)
```

**Concrete failure.** Open a PR that deletes `windowsHide: true` from `scripts/gh.mjs:74` and touches nothing else. `detect` sets `step_test_server:true`, "Server tests (fast + slow)" spins up, installs ffmpeg, runs `vitest --changed`, selects **zero** test files, exits 0, reports green. `verify.yml` is the required, merge-blocking check. The flashing window merges.

This is strictly worse than the `[cached]` symptom pass 2 called Significant: before this commit the leg was *skipped* and read as skipped; now it *runs and does nothing*, which reads as coverage on the checks list. Reviewer-brief shape 6, "success reported while doing nothing".

**The remedy is already in the file, added for this exact reason.** `server/vitest.config.ts:189` lists `openapi.yaml` in `forceRerunTriggers`, with the comment (`:172-175`): *"voice-library.test.ts pins MAX_CLONE_TRANSCRIPT_CHARS against it by reading the file at RUNTIME — no module-graph edge, so `--changed` would otherwise skip the pin on the openapi-only diff it exists to catch."* Character for character the situation here, for 52 files instead of one. `server/src/force-rerun-triggers.test.ts` is the guard for that mechanism and already has the `MAIN_COVERED` row shape to extend; its `NOT_COVERED` set (`src/index.ts`, `tsconfig.json`, `apps/android/pubspec.yaml`) is unaffected by a `scripts/**` + `server/tts-sidecar/scripts/**` trigger, so the anti-over-widening assertions stay green.

Not a design pass: the behaviour ("the guard must execute when a file it scans changes") is already agreed and already implemented once in this repo. Choosing between `forceRerunTriggers` and dropping `--changed` for this leg is weighing two implementations of one agreed behaviour, which `findings-triage.md` explicitly excludes from the carve-out.

#### 15. 🟠 Significant — the release-notes not-applicable is now stated on a premise this PR's own scope contradicts
PR body, Test plan, last bullet · `docs/release-notes-next.md` and `RELEASE_NOTES.md` untouched

> *"Release notes intentionally not updated: this is a pure console-window-suppression fix with no observable behavior change … even on the files that ship in the release zip."*

That was defensible for the 3-file and 37-file dev-tooling scopes. It is not true of the scope commit 75e107bd created, and the PR body says so itself two paragraphs earlier:

- **`server/tts-sidecar/scripts/accelerator-profile.mjs:87`** — `bootstrap-venv.mjs:250` calls `resolveInstallProfile` with no probe argument, so the default `execSync('powershell -NoProfile -Command "(Get-CimInstance Win32_VideoController).Name"')` fires on **every Windows install and every upgrade that rebuilds the venv**, from the fs-1 detached launcher. `server/tts-sidecar/**` is in the release-zip MANIFEST.
- **`pinokio-scripts/lib/resolve-release.js:87/105/109`** — three `git` spawns on **every Pinokio install and update**, from a console-less Electron parent.

"No observable behavior change" is the one claim that cannot be made about these two: the observation *is* the deliverable. #2567 exists because a user observed exactly this. An end user upgrading Castwright on Windows saw a PowerShell window flash and will now not; a Pinokio user saw three and will now see none.

For calibration, `RELEASE_NOTES.md`'s current in-progress `# Castwright 1.15.0` section opens with a line about **CI's weekly quarantine-register parser** — an internal automation fix with no user surface at all. A visible Windows install/upgrade artefact disappearing clears that bar comfortably. Both files are untouched; CLAUDE.md step 5 wants the pair, or an accurate not-applicable, and the one on record is inaccurate rather than merely thin.

#### 16. 🟡 Minor — on-box acceptance neither recorded nor declared not-applicable
`docs/testing/onbox-acceptance-register.md` (no row) · PR body (no mention)

The two paths in finding 15 ship behaviour nothing in this repo can prove: no automated test can observe a Win32 console window, and the guard only proves the flag's presence in source text. The PR body's `GetConsoleWindow()` measurement was taken on the dev-tooling spawn path, not on a real release-zip install from the detached launcher or a Pinokio install under Electron. `grep` finds no `windowsHide`/console-window row in the register.

CLAUDE.md step 3 makes *recording* (not running) the merge gate, and forbids silent omission. Either a row — "on a Windows box, run a real release-zip upgrade and a Pinokio update; observe no PowerShell/git window appears" — or an explicit sentence saying why one is not owed.

#### 17. 🟡 Minor — the guard's floor stayed an enumeration where its sibling fix chose a glob; the next file is still unguarded
`server/src/spawn-windows-hide.test.ts:61-134`

Concrete failure: add `scripts/new-tool.mjs` containing `spawnSync('git', ['status'], { stdio: 'ignore' })`. `npm run test:all`, `npm run verify` and the cloud battery all stay green and the flash ships. That is pass-2 finding 7's shape displaced from "the 34 existing files" to "the next one".

Flagging the asymmetry rather than the choice: `verify-cache.mjs:378` explicitly abandoned a hand-maintained list for a glob this round, on the reasoning that a list "needed a third manual sync every time the floor picked up a new file" — while the floor that list tracks stayed hand-maintained at 49 entries. Note the floor's own comment (`:56-59`) argues only against **replacing** it with a glob, which would drop `launch.mjs`; **adding** `scripts/**/*.{mjs,cjs}` alongside the retained list loses nothing. Not a merge blocker on its own, and a defensible thing to decline — but it is the residual half of a finding already fixed twice.

---

### Is there a fourth round of this defect class? — No, not in the fix itself.

Asked directly, and answered by measurement rather than by inspection of the diff. A repo-wide scan (all `.mjs/.cjs/.js/.ts/.tsx`, `fork(` added to the spawn-name list, `node_modules`/`.venv`/`dist` excluded) found 124 files importing `child_process` and 66 calls without `windowsHide` — **all 66 are in `*.test.ts` / `*.test.mjs` harness files** (`scripts/tests/**`, `server/src/**/*.test.ts`), which run from a terminal and ship to nobody. Specifically:

- `pinokio-scripts/**` — `resolve-release.js` is the **only** file in the tree that touches `child_process`, and it is now clean.
- `e2e/**` — `global-teardown.ts` only, now clean.
- `server/src/**` (non-test) — 0 offenders; the in-tree scan already covered it.
- `apps/android/**` — Dart/Gradle, no Node spawn surface.
- No `execa` / `cross-spawn` / `shelljs` in either `package.json`, so nothing spawns through a wrapper the scan cannot see.
- `fork(` is unused repo-wide.

**40 files is the genuine full extent of the production defect.** Rounds 2 and 3 were about the *guard*, not the *fix*, and finding 14 is the last layer of that same onion — source-text scan (round 2) → local cache/scope (round 3) → cloud test selection (this round). `forceRerunTriggers` is where that chain terminates; there is no further layer beneath it.

---

### Verdict

**DO NOT MERGE yet.** One 🔴, one 🟠, two 🟡.

The shipped fix is done and I am confident in it: 40 files, every production spawn site in the repo, mechanically verified complete and every one of them now inside a guard that provably fails when the flag is removed. Findings 8, 9, 12 are properly closed, 7 and 11 are closed with positive controls, and 10's local half is closed structurally rather than by another hand-maintained list.

What is not done is the same thing the last two rounds were about, one layer further out. Finding 14 is not scope creep and not a new class — it is pass-1 finding 3's cloud half, which was reported, accepted, and fixed at the wrong layer. Leaving it means merging a 49-file regression net that the required merge gate cannot execute on precisely the diff shape it exists to catch, while showing green. Finding 15 is a one-line house gate whose current justification is factually wrong about the PR's own shipped surface.

Both are small and fully specified: one `forceRerunTriggers` entry plus a `MAIN_COVERED` row in the guard test that already exists for it, and a release-notes pair. Neither needs a design pass, so neither qualifies for the carve-out.

**This is the loop cap — pass 3 of 3.** No further automatic re-review round is available, so these hand to the repo owner rather than to another gate pass. My recommendation: fix 14 and 15, and take 16 and 17 in the same round since both are cheap; then merge on the owner's judgement rather than on a fourth review.

Issue #2548, opened 2026-08-20
[agent instructions][claude][verify] re-pin onnxruntime-gpu to a CUDA-12-built version


dudarenok-maker:
@C:/Users/dudar/AppData/Local/Temp/open-engine-scratch/claude-2554-20260821-014123/body-2548.md

dudarenok-maker:
AGENT CLAIMED
Agent: claude

Starting verify pass for the CUDA-12 onnxruntime-gpu re-pin (child of #2534). Will read the diff on branch fix/sidecar-2534-ort-cuda12-pin, reproduce the CUDAExecutionProvider construction empirically, run typecheck + server tests, then report per the FAIL/PASS ordering in the brief.

dudarenok-maker:
AGENT DONE
Agent: claude

Verdict: PASS. Re-verified the CUDA-12 re-pin task child (#2547) independently rather than trusting its completion report, and opened the PR: https://github.com/dudarenok-maker/Castwright/pull/2576 (Closes #2534).

Checklist against the parent's acceptance:
- `ONNXRUNTIME_GPU_CONSTRAINT` (install-ort.mjs:219) no longer requires CUDA 13.x — pinned `>=1.26,<1.27`, which resolves to onnxruntime-gpu 1.26.0.
- Completion report showed real verification commands/output (session construction, not just `get_available_providers()`) — and I reproduced it myself independently: fresh scratch venv on this box, torch==2.11.0+cu128, onnxruntime-gpu==1.26.0, torch's bundled cuDNN 9 + CUDA 12.4 on the DLL search path (matching how the deployed sidecar venv is wired):
  ```
  onnxruntime version: 1.26.0
  available providers: ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
  session providers: ['CUDAExecutionProvider', 'CPUExecutionProvider']
  run output: [[6.0, 8.0, 10.0, 12.0]]
  CUDA SESSION OK + RUN OK
  ```
  A real `InferenceSession(providers=['CUDAExecutionProvider','CPUExecutionProvider'])` construction and run, landing on CUDA first — not the CPU-fallback check the issue's root-cause warned against.
- Doc comment above the constraint names the newly-validated version (1.26.0) and date (2026-08-21), replacing the old "1.27.0" line.
[...]

Issue #2547, opened 2026-08-20
[agent instructions][cline][task] re-pin onnxruntime-gpu to a CUDA-12-built version


dudarenok-maker:
## What

Re-pin `ONNXRUNTIME_GPU_CONSTRAINT` in `server/tts-sidecar/scripts/install-ort.mjs`
away from `onnxruntime-gpu` 1.27.x (which requires CUDA 13.x / cuDNN 9.x runtime
libraries) to the newest `onnxruntime-gpu` line that is actually built against
CUDA 12.x — matching what this box (and presumably most operator boxes today) has
installed system-wide. Decision made on parent #2534: re-pin to a CUDA-12-built
version (not vendoring CUDA 13 via pip, not requiring an operator toolkit upgrade).

## Background you need

- The constraint lives in exactly ONE place by design:
  `server/tts-sidecar/scripts/install-ort.mjs:214`:
  `const ONNXRUNTIME_GPU_CONSTRAINT = '>=1.27,<1.28';` — read the doc comment
  directly above it (`:198-213`) before touching it: it explains why this floor-
  plus-cap-on-minor pattern exists (so a same-line patch flows without a code
  change) and why the constraint can NEVER move into `requirements/*.txt` (those
  overlays are also read on macOS, which has no `onnxruntime-gpu` wheel at all, and
  `tests/test_requirements.py`'s `test_no_bare_unmarked_onnxruntime_gpu` enforces a
  bare `onnxruntime-gpu` line never lands there).
- Root cause (issue #2534): `onnxruntime-gpu` 1.27.0 requires CUDA 13.x runtime
  libraries (its own error names `cublasLt64_13.dll`). This box has CUDA 12.4
  system-wide (`cublasLt64_12.dll`, not `_13`). `onnxruntime.get_available_providers()`
  still lists `CUDAExecutionProvider` (the provider library ships regardless), but
  actually constructing a session with it fails with `Error 126: The specified
  module could not be found` and silently falls back to CPU — no error, no warning.
- You need to determine the correct replacement constraint. **This requires actual
  research, not a guess**: check `onnxruntime-gpu`'s published release history (e.g.
  `pip index versions onnxruntime-gpu`, or PyPI/the project's own release notes) to
  find the newest 1.x line whose GPU execution provider is built against CUDA 12.x
  (not 13.x) — i.e. the last minor line before the CUDA-13 requirement was
  introduced. Confirm your choice empirically: install the candidate version into a
[...]

Issue #2192, opened 2026-08-06
Qwen3 install fails with WinError 5 on a locked onnxruntime DLL, and silently replaces onnxruntime-gpu with the CPU build when it succeeds


dudarenok-maker:
Reported by @totallyfreak (Pinokio install, Windows 11, 12 GB VRAM), 2026-08-06:

```
install-qwen3.mjs exited with code 1. ERROR: Could not install packages due to an
OSError: [WinError 5] Accès refusé:
'E:\pinokio\api\Castwright.git\server\tts-sidecar\.venv\Lib\site-packages\onnxruntime\capi\onnxruntime_providers_shared.dll'
Check the permissions.
```

This is **not** a filesystem-permissions problem. `WinError 5` on a `.dll` is Windows'
"the file is memory-mapped by a running process." Two distinct defects sit behind it,
and the second one is worse because it is silent.

## Defect A — the Qwen3 install races the running sidecar's onnxruntime lock

1. `qwen-tts==0.1.1` declares `Requires-Dist: onnxruntime` — plain, unpinned.
2. On an NVIDIA-profile box the venv deliberately carries **`onnxruntime-gpu`** instead,
   swapped in by `install-ort.mjs`, which describes itself as "the SINGLE enforcement
   point for GPU Kokoro". pip does **not** accept `onnxruntime-gpu` as satisfying a
   requirement named `onnxruntime` — different distribution names, same `onnxruntime/`
   namespace directory.
3. So `install-qwen3.mjs:399`'s `pip install qwen-tts -c base.txt` always resolves plain
   `onnxruntime` as a *missing* dependency and tries to write it over the GPU runtime's
   files.
4. The sidecar's startup handler runs `_run_device_probe()` (`main.py:9174`, wired into
   the startup sequence at `main.py:668`), which does `import onnxruntime`
   **unconditionally on every boot** — not gated on Kokoro ever being loaded. That maps
   `onnxruntime_providers_shared.dll` and Windows holds the lock for the process
   lifetime.
5. `POST /api/qwen/install` spawns the installer **without stopping or unloading the
   sidecar** — there is no stop/unload path in `server/src/routes/qwen-install.ts` or
   `server/src/tts/qwen-install-bootstrap.ts`.

pip therefore tries to overwrite a locked DLL → `WinError 5` → pip exits 1 →
`install-qwen3.mjs` exits 1. Reproducible by construction on any box where the app is
running when the in-app Qwen3 installer is used.

[...]

Issue #1952, opened 2026-07-29
openapi-setup-parity.test.ts is LF-only: the Windows cross-OS leg fails 7 tests on a CRLF checkout


dudarenok-maker:
`Cross-OS Verify` has been red on `main` since the 2026-07-29 cron run
([run 30423665833](https://github.com/dudarenok-maker/Castwright/actions/runs/30423665833)):
the `windows-latest` leg fails 7 tests in `server/src/routes/openapi-setup-parity.test.ts`,
all with `schema <Name> not found in openapi.yaml: expected -1 to be greater than -1`.
macOS and the Ubuntu PR battery are green.

## Cause

`schemaBlock()` anchors on an LF-only literal:

```ts
const anchor = `\n    ${schemaName}:\n`;
const start = yaml.indexOf(anchor);
```

`.gitattributes` declares `* text=auto` and does **not** pin `openapi.yaml`, so a
Windows checkout with `core.autocrlf=true` (the GitHub `windows-latest` default)
materialises the file with CRLF. `indexOf` then returns `-1` for every schema.
Reproduced locally against the real file:

```
LF   indexOf('\n    BlockerCause:\n'):  308668
CRLF indexOf('\n    BlockerCause:\n'):  -1
```

The sibling `every /api/setup/* route the server mounts is described` test survives
because JS `$` under `/m` matches before `\r` as well as `\n` — which is exactly why
the failure count is 7 and not 8.

## Why it reached `main`

The test landed 2026-07-27 (56308d40 / a74cd636, fe-57 / #1883). PR CI runs on Ubuntu
only, and `cross-os.yml` runs twice weekly — the last green run predates the test.
Local Windows dev boxes carry `core.autocrlf=false`, so it passes there too.

The three other runtime readers of `openapi.yaml`
(`server/src/routes/voice-library.test.ts:2604`, `src/lib/api.clone-voice.test.ts:126`
and `:155`) happen to be safe today only because their anchors carry no embedded
newline — the class is latent, not absent.

## Acceptance

- `openapi-setup-parity.test.ts` passes against a CRLF copy of `openapi.yaml`, pinned
  by a regression test that fails before the fix.
- A CRLF checkout can no longer occur for `openapi.yaml` at all, closing the class for
  every current and future consumer.
- `Cross-OS Verify` is green on `main`.

## Key files

- `server/src/routes/openapi-setup-parity.test.ts`
- `.gitattributes`

PR #2576, opened 2026-08-21
fix(sidecar): re-pin onnxruntime-gpu to a CUDA-12-built version

### LINKED ISSUE #2534: onnxruntime-gpu 1.27 silently falls back to CPU on CUDA 12.4/cuDNN mismatch
## Summary

`onnxruntime-gpu` 1.27 silently falls back to `CPUExecutionProvider` on this box's CUDA 12.4 / cuDNN mismatch, instead of erroring or warning. `onnxruntime.get_available_providers()` reports `CUDAExecutionProvider` as available even though actually constructing a session with it fails.

## Root cause

`onnxruntime-gpu` is pinned to `>=1.27,<1.28` (`server/tts-sidecar/scripts/install-ort.mjs:214`, `ONNXRUNTIME_GPU_CONSTRAINT`). `onnxruntime-gpu` 1.27.0 requires CUDA 13.x and cuDNN 9.x runtime libraries (its own error names them: `onnxruntime_providers_cuda.dll` depends on `cublasLt64_13.dll`). This box has only CUDA 12.4 installed system-wide (`C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.4`, providing `cublasLt64_12.dll`, not `_13`), and no pip-vendored CUDA 13 runtime packages appear anywhere in `server/tts-sidecar/requirements/`. The driver (610.88) does support CUDA 13 — this is a missing-runtime-library gap, not a driver/hardware limit.

`import onnxruntime as ort; ort.get_available_providers()` correctly lists `CUDAExecutionProvider` (the provider *libraries* are present), but actually constructing a session with it (`onnxruntime.InferenceSession(..., providers=['CUDAExecutionProvider', ...])`, or `kokoro_onnx.Kokoro(...)` with `ONNX_PROVIDER=CUDAExecutionProvider`) fails with `Error 126: The specified module could not be found` and **silently falls back to `CPUExecutionProvider`** — no error surfaced to the operator, no warning logged.

## Evidence

Reproduced during wave-3 on-box acceptance testing of register rows A39/A40/A41 (`docs/testing/onbox-wave3-results/step-2-ort-marker.md`):

- Genuine from-scratch `node scripts/bootstrap-venv.mjs py -3.12` with `ACCELERATOR=nvidia`: marker write and `pip check` both correct, but loading Kokoro's execution provider fell back to CPU.
- Reproduced identically against the **live** sidecar venv, read-only (no modification) — not specific to the throwaway A39 venv.
- Blocks the "Kokoro reports `CUDAExecutionProvider`" success criterion for register rows A39, A40, and (indirectly, on re-check) A41 — every wave-3 row whose final check is "the app actually renders on GPU."

## Decision owed

Three defensible outcomes — not fixing here, needs a design call:

1. Add the CUDA 13 / cuDNN 9 runtime as pip-installable packages to `server/tts-sidecar/requirements/nvidia-cuda.txt` (the `nvidia-cublas-cu13` / `nvidia-cudnn-cu9`-shaped packages `onnxruntime-gpu` 1.27 expects), so the venv is self-contained and doesn't depend on the box's system CUDA toolkit version.
2. Re-pin `ONNXRUNTIME_GPU_CONSTRAINT` to a version of `onnxruntime-gpu` built against CUDA 12.x, matching what's actually installed system-wide today.
3. Upgrade the box's system CUDA toolkit to 13.x (operator action, out of an agent's box-safety scope).

Separately from the pin/runtime decision above: `onnxruntime`'s silent CPU fallback on a provider-construction failure is itself worth a decision — should the sidecar detect and warn/error when the requested `CUDAExecutionProvider` silently degrades to CPU, rather than reporting success with a GPU-shaped payload while actually running on CPU?

## Source

Found during wave-3 on-box acceptance testing, register rows A39/A40/A41. Full write-up: `docs/testing/onbox-wave3-results/step-2-ort-marker.md`.

dudarenok-maker:
## Summary

Re-pins `ONNXRUNTIME_GPU_CONSTRAINT` (`server/tts-sidecar/scripts/install-ort.mjs`) from
`>=1.27,<1.28` to `>=1.26,<1.27` — the last CUDA-12-built onnxruntime-gpu line.
1.27+ moved the default build to CUDA 13.x (`cublasLt64_13.dll`), which this box's
system-wide CUDA 12.4 can't satisfy, so the old pin silently fell back to CPU.
Closes #2534.

## Test plan

Verified independently as the `[claude][verify]` child (#2548), reproducing the
task child's (#2547) empirical claim rather than trusting its report:

- [x] Installed `onnxruntime-gpu==1.26.0` into a scratch venv on this box (torch
      2.11.0+cu128, torch-bundled cuDNN 9 + CUDA 12.4 on the DLL search path —
      matching how the deployed tts-sidecar venv is wired) and constructed a real
      `InferenceSession(providers=['CUDAExecutionProvider', 'CPUExecutionProvider'])`:
      ```
      onnxruntime version: 1.26.0
      available providers: ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
      session providers: ['CUDAExecutionProvider', 'CPUExecutionProvider']
      run output: [[6.0, 8.0, 10.0, 12.0]]
      CUDA SESSION OK + RUN OK
      ```
      (Not just `get_available_providers()` listing the EP — an actual session
      construction + run, landing on `CUDAExecutionProvider` first.)
- [x] `install-ort-helpers.test.ts` + `bootstrap-venv-helpers.test.ts` — 18/18 passed
- [x] `npm run typecheck` — clean
- [x] Full server test suite — 7817/7840 passed (5 failures, in
      `finalize-chapter-write.test.ts`, `qwen-voice.test.ts`, `state-io.test.ts`,
      `export.test.ts`), all pre-existing flakiness unrelated to this change
      (timeouts/`ENOTEMPTY`/worker-fork crash under this box's heavy concurrent
      lane load; none of the failing files reference onnxruntime or install-ort).
- [x] ~~`requirements/*.txt` untouched~~ — **no longer true, see below.** The
      original commit didn't touch `requirements/*.txt`, but PR review (pass 1)
      found that made the pin change a no-op for every existing venv (`reqHash`
      never moved, so `decideVenvAction` stayed `'noop'` and the swap never
      re-ran) — fixed by a follow-up commit that intentionally bumps
      `requirements/nvidia-cuda.txt`'s hashed text. See "Also fixed, found in
      passing" below.

Three independent review passes ([pass 1](https://github.com/dudarenok-maker/Castwright/pull/2576#issuecomment-5365932266), [pass 2](https://github.com/dudarenok-maker/Castwright/pull/2576#issuecomment-5366490240), [pass 3 — final, 3-pass cap](https://github.com/dudarenok-maker/Castwright/pull/2576#issuecomment-5366956863)) surfaced defects in the original commit and in earlier fix commits on this branch; each was fixed by a dispatched fix agent, one finding per commit:

- The pin re-pin was a no-op for every existing nvidia venv — `reqHash` never changed, so the upgrade/bootstrap paths classified the venv as `'noop'` and never re-ran the ORT swap. Fixed by bumping `requirements/nvidia-cuda.txt`'s hashed content, with a regression test.
- On-box acceptance register rows A39/A40 still described the CUDA-12/13 pin mismatch as an undecided three-way blocker; updated across all three surfaces (register, evidence doc, republished live view) now that #2534's pin decision is resolved by this PR — the rows stay OPEN (the on-box run itself is still owed).
- `docs/release-notes-next.md` had two now-false "1.27.x" references and no entry for this PR's change; added/corrected, plus a user-facing `RELEASE_NOTES.md` bullet.
- The `install-ort.mjs` comment blamed the wrong bump condition (a box's system CUDA toolkit rather than the shipped torch/torchaudio cu128 pin); corrected.
- The follow-up test added to guard the no-op finding was itself hollow (passed in both the pre-fix and post-fix state); rewritten as a real oracle with a confirmed red/green check.
- Minor self-contradictions left behind by the above fixes (a historical release-notes entry conflated with this PR's change, a comment block contradicting its own later sentence, an imprecise user-facing line) were also corrected.
- The rewritten oracle test was itself CRLF-inconsistent (hashed the overlay normalized but the base file raw, diverging from production's `resolveRequired` on a stock Windows/`core.autocrlf=true` checkout — a false red unrelated to any real regression) and cap-blind (an `||` where an `&&` was needed meant a future cap-only widening of the constraint, e.g. `>=1.26,<1.28`, wouldn't be caught). Both fixed in the same commit.
- A final residue pass: an opening sentence in the `RELEASE_NOTES.md` bullet still implied a hardware/driver capability limitation; the restored historical `#1846` release-notes entry needed a forward-pointer noting it was superseded before release (both `#1846` and this PR's re-pin land in the same unreleased version, so nobody upgrading will actually observe the intermediate 1.27.x stop); two prose/wrapping issues in the `install-ort.mjs` comment; and one stale `>=1.27,<1.28` reference in an unrelated design-spec doc.

Split out to its own issue rather than fixed here (needs a design decision): [#2582](https://github.com/dudarenok-maker/Castwright/issues/2582) — whether/how the sidecar should detect and surface a *future* silent `CUDAExecutionProvider`→CPU fallback, separate from this PR's specific CUDA-12/13 pin fix.

**Review loop note**: this PR went through all 3 rounds the `pr-review-gate` skill allows (initial pass + 2 re-review rounds). Pass 3's sole remaining correctness finding (the CRLF/cap-blind test issues above) was fixed and independently verified (green tests, clean typecheck, `check:onbox-register` OK) without a 4th automated review round, per the skill's cap.


dudarenok-maker:
## PR review — pass 1 (head 6587783bbce1a5c40eeac8a73bedc4200aacf34d, depth medium)

### Scope

Single-scope `fix(sidecar)`, 3 files: `ONNXRUNTIME_GPU_CONSTRAINT` re-pinned `>=1.27,<1.28` → `>=1.26,<1.27` in `server/tts-sidecar/scripts/install-ort.mjs` plus its explanatory comment, and two test files updating the mirrored expected string. Not docs-only, so the gate applies. Reviewed against the house gates (paired test, on-box acceptance across all three surfaces, release-notes pair, issue linkage, derived artifacts, sibling call sites) and the recurring-defect catalogue. Reachability of the change through the shipped install paths was traced end-to-end rather than assumed.

---

### 🔴 Blocking

**1. [correctness] The fix never reaches any existing install — the ORT swap is gated on `reqHash`, and this PR changes no requirements file.**

`server/tts-sidecar/scripts/venv-migration.mjs:116-124` — `resolveRequired` computes `reqHash` from exactly two files: `requirements/<overlay>.txt` and `requirements/base.txt`. `install-ort.mjs` is not an input to that hash.

`server/tts-sidecar/scripts/venv-migration.mjs:38-44` — `decideVenvAction` returns `'noop'` when `pythonTag`, `profile` and `reqHash` all match. `classifyVenvState` (`:81-86`) passes that through unchanged for an existing venv.

The two gates that consume it are the only callers of `plan.steps`:

- `server/tts-sidecar/scripts/bootstrap-venv.mjs:256-261` — `if (action === 'noop') { log('venv up to date — nothing to do'); return; }`. `installForProfile` (line 278/306), which runs the swap, is never reached.
- `server/src/upgrade/apply.ts:134-139` — `if (ctx.reqHash && ctx.reqHash !== steps.readReqHash())` gates `steps.pipInstall(releaseDir)`, and `pipInstall` (`:269-297`) is where `plan.steps` runs on the in-app upgrade path.

**Failure scenario.** Take this dev box, or any alpha install: a bootstrapped nvidia venv holding `onnxruntime-gpu 1.27.0`, with a `.venv-stamp.json` whose `reqHash` matches `main`. Merge this PR, then exercise any shipped update route — `npm start` → `bootstrap-venv.mjs`, the in-app upgrade (`apply.ts`), `POST /api/setup/venv/bootstrap` (`server/src/routes/venv-bootstrap.ts`), or a Pinokio Update (`pinokio-scripts/update.js:73`, which shells the same `bootstrap-venv.mjs`). All four compare the same `reqHash`; this PR leaves it byte-identical; none re-runs the swap. `onnxruntime-gpu` stays at 1.27.0, the CUDA session keeps failing with `Error 126`, and Kokoro keeps silently falling back to CPU — the exact symptom #2534 reports. Only a venv that does not exist at all (`fresh-bootstrap`) picks up `>=1.26,<1.27`.

This is not an inference — the repo already states the mechanism as fact. `docs/release-notes-next.md:335` (the #2192 entry): *"neither the upgrade nor rebuild path re-runs the swap on an unchanged requirements hash, so nothing else would ever have granted an existing install the marker."*

**Why the apparent precedent doesn't carry.** `docs/release-notes-next.md:331` claims of the original side-28 pin: *"Existing installs on 1.28.x will step back to 1.27.x on first upgrade — both upgrade paths (`bootstrap-venv.mjs`, `apply.ts`) re-run the swap."* That was true for #1846 only because the same PR bumped `kokoro-onnx` `>=0.4.0,<0.5.0` → `>=0.5.0,<0.6.0` across all three overlays (line 330, and the pin is still visible at `requirements/nvidia-cuda.txt:61`, `cpu.txt:25`, `amd-rocm.txt:22`). That rider moved `reqHash`. This PR has no rider, so it inherits the claim without inheriting the condition that made it true.

Two things compound this. First, the PR's test plan lists *"`requirements/*.txt` untouched — confirmed via diff stat"* as a **passing** check; it is precisely the condition that makes the change inert. Second, the on-box verification was performed in a **scratch venv** — a fresh bootstrap, the one path where the new pin does apply, and the only one of the five that is not gated. The four gated paths were not exercised.

`ensureOrtMarker` does not rescue this: it is pure fs (`install-ort.mjs:296-341`), writes a marker, and never invokes pip. The only route that re-runs the swap on an up-to-date venv is the manual CLI form in this file's own header (`CASTWRIGHT_ACCELERATOR_PROFILE=nvidia node install-ort.mjs <venv-python>`), which nothing shipped invokes and no release note surfaces.

Catalogue shape 7 — *a fix unreachable at the default configuration; correct code that runs zero times as shipped.* Whatever the resolution, it needs a regression test that asserts a pin-only change is actually delivered to an existing venv, because nothing today would have caught this.

---

**2. [correctness] On-box acceptance state is recorded on none of the three surfaces, and the register is left asserting a decision as open that this PR closes.**

`docs/testing/onbox-acceptance-register.md:2305-2320` (A39's flag) states the blocker as *"this box has only CUDA 12.4 system-wide while `onnxruntime-gpu` 1.27 needs CUDA 13.x/cuDNN 9.x runtime libraries"* and points at *"decision options for the fix (repin onnxruntime-gpu vs. vendor CUDA13 wheels vs. upgrade the box)"*, tracked as #2534. This PR **makes** that decision — option 1 of the three — and closes #2534. After merge the register cites a closed issue as the live tracker for an open three-way decision.

`docs/testing/onbox-acceptance-register.md:2342-2348` (A40's flag): *"this row's final check would hit the same CUDA13/cuDNN9 gap A39 found, so even a full run today could not fully discharge it without that dependency gap closed first."* That sentence's stated reason stops being the real one at merge.

Neither of the other two surfaces moves either:

- `docs/testing/onbox-acceptance-register-live-view.html:774` carries the same prose verbatim and must be edited **and re-published to the URL recorded in the register header** (publishing the `.md` to that URL, or publishing without the `url`, are the two documented ways this breaks silently).
- `docs/testing/onbox-wave3-results/step-2-ort-marker.md:14-15,48,93-97` — the evidence/run-sheet file — records `onnxruntime-gpu 1.27.0` as the observed state and the three-way decision as owed.

CLAUDE.md before-shipping step 3 is explicit that this blocks: *"Recording blocks the merge; running does not."* The behaviour this PR ships — whether the newly-pinned runtime actually gives Kokoro a working `CUDAExecutionProvider` session **through the shipped install path** — is exactly A39's fourth criterion, and (per finding 1) is now doubly unproven. Running it is not owed at PR time; recording the state is.

---

### 🟠 Major

**3. [correctness] Two unreleased release-notes lines are made false, and no entry is added for this change.**

`docs/release-notes-next.md` is fed **verbatim** as the GitHub release body by `bump-version.mjs` (file header, lines 2-6), and the marker at line 19 shows the v1.15.0 cycle is still open — so both lines below ship as written:

- **`docs/release-notes-next.md:59`** — the v1.15.0 **Upgrade note**: *"**Existing `onnxruntime-gpu` 1.28.x installs step back to 1.27.x** on first upgrade…"*. The pin is now `>=1.26,<1.27`.
- **`docs/release-notes-next.md:331`** — *"`onnxruntime-gpu` gets an explicit `>=1.27,<1.28` constraint in `scripts/install-ort.mjs`… Existing installs on 1.28.x will step back to 1.27.x on first upgrade."* Same falsehood, and its trailing "both upgrade paths re-run the swap" clause is the unexamined premise behind finding 1.

Separately, there is no new entry in `docs/release-notes-next.md`, no matching brand-voice line in `RELEASE_NOTES.md`, and no explicit not-applicable in the PR body. Which ORT version an install lands on, and whether Kokoro runs on GPU or CPU, is operator-visible by any reading, so the "no shippable delta" skip in before-shipping step 5 does not apply.

**4. [correctness] The new comment states a box-local root cause, and therefore documents the wrong bump condition.**

`server/tts-sidecar/scripts/install-ort.mjs:198-213` (new text): *"a dependency **this box's system-wide CUDA 12.4** cannot satisfy"*, and *"bump… once a CUDA-13-built line is desk-validated **on a box that has CUDA 13**."*

The failure is not box-local. `server/tts-sidecar/requirements/nvidia-cuda.txt:27-31` pins `torch==2.11.0` installed from the cu128 index (`accelerator-profile.mjs:160`; asserted at `bootstrap-venv-helpers.test.ts:48-50`), so **every** shipped nvidia venv vendors CUDA 12 libs and no CUDA 13 ones. #2534's own body confirms it: *"no pip-vendored CUDA 13 runtime packages appear anywhere in `server/tts-sidecar/requirements/`."* A user who has never installed a CUDA toolkit at all hits the identical `cublasLt64_13.dll` failure — the system toolkit is incidental here, not causal.

**Failure scenario.** A future maintainer installs CUDA 13 system-wide on the dev box, satisfies this comment's bump condition verbatim, desk-validates 1.27 or 1.28 there, and bumps the constant with a requirements rider so it actually ships. Their box works. Every user still on the `cu128` torch wheel is back to a silent CPU fallback. The real gate is torch's CUDA line moving off cu128 — which the comment never names, even though the "bump the runtime and this pin together" sentence is clearly reaching for it.

---

### 🟡 Minor / cleanup

**5. [cleanup] `Closes #2534` retires a second, unaddressed decision the issue records.**

#2534's body carries two items, the second explicitly separated: *"should the sidecar detect and warn/error when the requested `CUDAExecutionProvider` silently degrades to CPU, rather than reporting success with a GPU-shaped payload while actually running on CPU?"* Nothing in this PR touches it and nothing re-files it, but `Closes #2534` retires the ticket carrying it. That silent degradation is why the bug survived undetected until wave-3 on-box testing — and, given finding 1, it is exactly what would hide the fix's non-delivery after merge. Either `Refs #2534` with the second half split out, or file it before closing.

---

### ✅ Checked clean

- **No sibling site left armed.** The constraint string appears in exactly three places repo-wide — `install-ort.mjs:214`, `bootstrap-venv-helpers.test.ts:58`, `install-ort-helpers.test.ts:29` — and all three move together. `ONNXRUNTIME_GPU_CONSTRAINT` / `constrainForInstall` have no other consumers. The `1.27.0` / `1.28.0` literals in `ort-ensure-marker.test.ts`, `ort-marker-apply.test.ts` and `ort-marker-io.test.ts` are arbitrary fixture versions, not the pin, and correctly stay put.
- **The two test edits are genuine red-before/green-after tripwires** — exact string equality against a constant the diff changes in lockstep — and both surrounding comments (`install-ort-helpers.test.ts:21-26`, `bootstrap-venv-helpers.test.ts:55-57`) remain true after the change.
- **The pin's direction is right.** Because the nvidia overlay pins `torch==2.11.0` from cu128, the venv is CUDA-12 by construction; 1.26 matches the vendored runtime and 1.27+ cannot work for any shipped nvidia install. The 1.27 pin was broken for essentially every NVIDIA user, not just this box.
- **Keeping the constraint out of `requirements/*.txt` is correct** — `test_no_bare_unmarked_onnxruntime_gpu` and the macOS wheel gap mean it belongs in the installer. (Finding 1 is not an argument against that; it is an argument that something else must move `reqHash`.)
- No `cast.json` writes, no lock-timeout seams, no config knob, no derived artifact (`src/lib/api-types.ts`, `docs/BACKLOG.md`, brand PNGs) in scope.
- `Closes #2534` is present in the PR body and outside any code span. PR title matches the commit convention; one commit, one scope.

---

### Verdict

**Request changes — do not merge.** Four correctness findings, one of them fatal to the PR's purpose: as staged, the change is a no-op on every install that currently has the bug, including the box #2534 was filed from. Findings 2 and 3 are house gates that block independently.

Per `findings-triage.md`, this pass carries correctness bugs, so a re-review is required once they are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head d7209f8ae5e5951472fb00b295df0778dbe265d2, depth medium)

Depth re-derived from the current commit set (3 × `fix(sidecar)`, 2 × `docs`, single-scope each; highest any one commit earns alone is medium) — **medium**, unchanged from pass 1.

Each pass-1 finding was re-verified against the code, the tracked docs and the **published** artifact rather than against the fix commits' own summaries. The five-commit diff was then re-read as one change.

---

### Pass-1 dispositions

**1. 🔴 "the fix never reaches any existing install" — delivery half VERIFIED RESOLVED; the regression-test half is NOT (see N1).**

The delivery half is genuinely fixed. Traced end to end rather than assumed:

- `server/tts-sidecar/requirements/nvidia-cuda.txt:45-46` adds two comment lines. `resolveRequired` hashes the overlay's **text** (`server/tts-sidecar/scripts/venv-migration.mjs:118-123`), so `reqHash` moves for every nvidia install (and apple, which shares the overlay).
- `bootstrap-venv.mjs:256` now classifies an existing stamped venv as `pip-in-place` (`:274`) → `runInstall` → `installForProfile`'s swap loop (`:229-238`) with the new `>=1.26,<1.27`.
- `server/src/upgrade/apply.ts:134-137` — `ctx.reqHash` is computed from the **candidate release's** overlay + base, in `resolveRequired`'s exact order (`server/src/upgrade/zip-validate.ts:205-213`), so it now differs from the installed `.req-hash` and `steps.pipInstall` runs (`:269-297`), reaching the same swap.
- The swap `uninstall`s both distribution names before `--force-reinstall`, so 1.27.0 → 1.26.x is a genuine downgrade, not a pip "already satisfied" no-op.
- **Checked because this PR creates a state that has never existed before** — the swap re-running on a venv that *already carries* a `#2192` marker. `bootstrap-venv.mjs:169` and `apply.ts:276` both `marker.del` before the first pip call, and `deleteOrtMarkerIfOurs` (`install-ort.mjs:169-177`) iterates every `onnxruntime-<digit>*.dist-info` and deletes each one that is ours, version-agnostically. So a stale `onnxruntime-1.27.0.dist-info` marker is removed rather than left beside the new `1.26.x` one. No new defect here.

Verified by execution, not only by reading — a read-only replay of `computeReqHash`/`decideVenvAction` against the committed overlay returned `pip-in-place` for a stamp holding the pre-comment hash.

The release-notes claim that pass 1 called "inherited without the condition that made it true" — *"both upgrade paths (`bootstrap-venv.mjs`, `apply.ts`) re-run the swap"* — now actually holds.

**2. 🔴 on-box acceptance recorded on none of the three surfaces — VERIFIED RESOLVED, all three.**

- **Register** — `docs/testing/onbox-acceptance-register.md:2305-2318` (A39) and `:2340-2348` (A40) both reframe the blocker as resolved while keeping the rows **STILL OWED**, which is the correct disposition: the decision is discharged, the run is not. Neither row leaves the owed count, so no arithmetic moves.
- **Live view** — `docs/testing/onbox-acceptance-register-live-view.html:774,789` carry the matching prose.
- **Published artifact** — fetched `https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a` (the exact UUID in the register header, not a re-mint). It is still the **styled hand-authored page**, not markdown rendering, and it contains `STILL OWED, blocker now fixed` (A39), `STILL OWED, not run, blocker now fixed` (A40) and two `PR #2576` references. Summary strip reads `74 / 8 / 2 / 2 / 40 / 06-01`, matching the tracked file.
- Ran the register's own gates against the branch: `check:onbox-register` → OK; `check:onbox-register --against-published <saved live page>` → OK (the register is not behind the live page, i.e. no row was lost at publish). CI's own `Check on-box acceptance register consistency` leg is green.
- The **"Blocked: 2"** stat is the separate `BLK` section, not A39/A40, so it correctly does not move. A41 was correctly left alone — its flag records the unrelated `detectOrtOwner` defect, not the CUDA gap.
- `docs/testing/ort-marker-onbox-acceptance.md` §3.3 still records `onnxruntime-gpu 1.27.0` — correct, and not a gap: that is a dated **run record** ("Run by / Date: 2026-08-20") that forwards its root cause to `onbox-wave3-results/step-2-ort-marker.md`, which this PR does update with a `**Resolution:**` block. No new `Result:` lines are owed because no acceptance was re-run.

**3. 🟠 stale/missing release-notes pair — RESOLVED, with residue (see N2).** `docs/release-notes-next.md:59` and `:330` no longer assert 1.27.x; a technical entry lands at `:334` and a brand-voice bullet at `RELEASE_NOTES.md:5`.

**4. 🟠 the comment documents the wrong bump condition — RESOLVED in substance; the clause it contradicts survives (see N3).** `install-ort.mjs:210-215` now names the real gate: *"the constraint can only move to a CUDA-13-built onnxruntime-gpu line once the shipped torch/torchaudio pins in requirements/nvidia-cuda.txt move to a CUDA-13 wheel … the shipped venv remains CUDA-12-by-construction regardless of what CUDA toolkit is installed system-wide."* The failure scenario pass 1 gave — a maintainer installing CUDA 13 on the dev box and bumping the pin — is closed.

**5. 🟡 `Closes #2534` retiring a second, unaddressed decision — RESOLVED via the design-pass carve-out.** [#2582](https://github.com/dudarenok-maker/Castwright/issues/2582) verified: open, `bug`-labelled, and it **names the decision owed** rather than restating the finding — three concrete detection mechanisms (boot-time real `InferenceSession`, per-session `get_providers()` check, health-endpoint surface) plus an explicit hard-fail-vs-warn severity call. That satisfies `findings-triage.md`'s "its issue must name the decision owed".

---

### New findings

**🟠 N1. [correctness] The regression test shipped with the reqHash fix passes in BOTH states — it cannot go red for the bug it names, and it does not guard the class pass 1 asked it to guard.**

`server/src/tts/venv-migration.test.ts:177-227`.

The test branches on `hasCommentMarker` (`:194`) — read from the **committed** `nvidia-cuda.txt` — and asserts opposite outcomes on each side (`:200` if-branch → `noop`; `:213` else-branch → `pip-in-place`). Both branches are internally consistent with the state that selects them, so whichever state the repo is in, the test is green.

Replayed the test's own logic, read-only, against the committed overlay and against a synthetic pre-fix copy (the same file with the `:45-46` comment removed — i.e. exactly the state `cfda13a6` fixed):

```
--- POST-FIX (as committed on this branch) ---  branch taken: ELSE (fix applied)
   PASS  expect(oldNvidiaContent).not.toBe(currentNvidia)
   PASS  expect(oldHash).not.toBe(currentHash)
   PASS  expect(decideVenvAction).toBe('pip-in-place') [got pip-in-place]
   => TEST GREEN

--- PRE-FIX (comment reverted out of nvidia-cuda.txt) ---  branch taken: IF (fix absent)
   PASS  expect(oldNvidiaContent).toBe(currentNvidia)
   PASS  expect(oldHash).toBe(currentHash)
   PASS  expect(decideVenvAction).toBe('noop') [got noop]
   => TEST GREEN
```

Green before the fix, green after. That fails the house gate outright ("red before the fix, green after, and red *for the reason claimed*"), and `cfda13a6`'s commit message asserts the opposite: *"adds a regression test that verifies: given an OLD reqHash from before the comment was added, `decideVenvAction` now returns 'pip-in-place' (not 'noop')."* It verifies that only on the branch the committed file happens to select.

**Failure scenario.** A maintainer bumps `ONNXRUNTIME_GPU_CONSTRAINT` to `>=1.28,<1.29` in `install-ort.mjs:224` and — exactly as #2534 did — touches no requirements file. `hasCommentMarker` is still true (the dated 2026-08-21 comment is still in `nvidia-cuda.txt`), the else-branch still passes, CI is green, and every existing venv classifies `noop`. The bug recurs with no signal. The test never reads `ONNXRUNTIME_GPU_CONSTRAINT`, `install-ort.mjs`, or anything that moves when the pin moves — the only thing that can turn it red is someone editing or reflowing that one comment block. Pass 1 asked for "a regression test that asserts a pin-only change is actually delivered to an existing venv"; this asserts that removing a known substring from a file changes that file's hash, which is a tautology.

**Second, independent break in the same test.** `:190-192` matches `markerComment` as a `\n`-joined literal, including a trailing blank line. `.gitattributes:2` is `* text=auto` with no `eol=lf` pin for `requirements/*.txt` (the repo pins LF explicitly for `.mjs`, `.sh`, `.husky/*`, `openapi.yaml`, `.env.example` and `src/lib/api-types.ts` — this file is not covered). On a stock Windows clone with `core.autocrlf=true`, `readFileSync(..., 'utf8')` returns CRLF, `String.replace` finds nothing, and `expect(oldNvidiaContent).not.toBe(currentNvidia)` (`:216`) **fails** — verified against a CRLF-converted copy. CI is Ubuntu and this box is `core.autocrlf=false`, so it is green in both places that currently run it; it goes red in pre-commit `verify:fast:scoped` for any contributor on Git-for-Windows defaults.

Both go away with the same rewrite: assert the invariant instead of the artifact — e.g. that `ONNXRUNTIME_GPU_CONSTRAINT`'s current value appears somewhere in the hashed overlay text, so any future pin move is *forced* to move `reqHash`. That is the guard #2534 needed; a dated comment is a fact, not a guard.

Catalogue shapes 3 (a test that cannot fail) and 5 (an acceptance criterion blind to its own feature).

**🟡 N2. [cleanup] `docs/release-notes-next.md:330` now contradicts itself inside one bullet, and `:334` duplicates its correction.**

`:330` (the #1846 entry) was rewritten to *"`onnxruntime-gpu` gets an explicit `>=1.26,<1.27` constraint"*, then closes with *"(The prior 1.27.x pin was re-pinned to 1.26.x in #2534 …)"*. If the entry introduces 1.26, there was no "prior 1.27.x pin" for it to describe. This file is fed **verbatim** as the GitHub release body (its own header, `:2-6`), and `:334` then restates the same re-pin and the same CUDA-13 rationale a third time, four lines below. Since both changes ship in the same open v1.15.0 cycle the *net* statement is accurate — it just reads as garbled. Either restore `:330` to what #1846 actually did and let `:334` carry the correction, or drop `:334` and keep one coherent bullet.

**🟡 N3. [cleanup] The box-local root-cause clause pass 1 flagged survives at `install-ort.mjs:205`, eight lines above the sentence that now contradicts it.**

`:204-206` still reads *"a dependency **this box's system-wide CUDA 12.4** cannot satisfy"*, while `:213-214` now correctly says *"the shipped venv remains CUDA-12-by-construction **regardless of what CUDA toolkit is installed system-wide**"*. Both cannot be true. The second is the right one — `requirements/nvidia-cuda.txt:27-31` pins `torch==2.11.0` from the cu128 index, so a user with no CUDA toolkit installed at all hits the identical `cublasLt64_13.dll` failure. The stale half is the exact framing that produced pass-1 finding 4; leaving it in re-arms the misreading the corrected sentence exists to prevent. Same for `:202-203`'s *"against this box's CUDA 12.4"* — the validation is real, the attribution isn't.

**🟡 N4. [cleanup] The PR body's test plan still checks off a claim this PR has made false — and it is the specific claim pass 1 identified as the tell for the fatal defect.**

PR body, Test plan: *"- [x] `requirements/*.txt` untouched — confirmed via diff stat (only `install-ort.mjs` + the two test files changed)"*. The diff is now 10 files including `server/tts-sidecar/requirements/nvidia-cuda.txt`, and moving that file is the entire mechanism by which the fix reaches an existing install. A reader taking the body at face value would conclude the reqHash-moving change was never made. The body also carries no `Also fixed, found in passing:` line for the register, live-view, release-notes and comment fixes that landed in commits 2–5, which CLAUDE.md requires so they don't read as scope creep.

**🟡 N5. [cleanup] `RELEASE_NOTES.md:5` describes a mechanism that does not exist.**

*"A re-pin … was checking compatibility with a future CUDA version that your card doesn't have yet … **The compatibility check is fixed**"*. There is no compatibility check, fixed or otherwise — the pin was moved. And CUDA-13 support is not a property of the card: this box's driver (610.88) *does* support CUDA 13, per `onbox-wave3-results/step-2-ort-marker.md`; what was missing were runtime libraries. A user-facing note is allowed to simplify, but not to invent a fix that did not happen. "Castwright now installs a speech-engine library built for the graphics-card runtime it actually ships with" says the true thing at the same reading level.

---

### Verdict

**Not ready to merge.** All four pass-1 correctness findings are resolved in the shipped behaviour and in the docs — the delivery mechanism is real and I traced it through both upgrade paths and the artifact publish, not just the commit messages. What is not resolved is the half of finding 1 that was about *recurrence*: the paired test is green in both states, cannot fail for #2534's symptom, and would not catch the next pin bump that forgets the requirements rider — the precise class pass 1 said "nothing today would have caught".

N1 is a correctness finding, so per `findings-triage.md` this re-triggers one more pass once it and the four 🟡 items are fixed and pushed. That would be round 3 of the 3-round cap.

dudarenok-maker:
## PR review — pass 3 (head 31f45e43954891844cd84ab901c083ba718c200e, depth medium)

Depth re-derived from the current commit set (3 × `fix(sidecar)`, 1 × `test(sidecar)`, 3 × `docs`, single-scope each; highest any one commit earns alone is medium) — **medium**, unchanged.

Deltas only, per the re-review rule. Every pass-2 finding was re-derived from the current file content and, where it was a claim about test behaviour, **replayed read-only against the branch content** rather than read off the fix commits' messages. Then a fresh pass over all 7 commits vs `main`.

Scope check first: `git show --stat` on the two post-pass-2 commits confirms `55f3d5d6` touches only `server/src/tts/venv-migration.test.ts` and `31f45e43` only `RELEASE_NOTES.md` + `docs/release-notes-next.md` + `install-ort.mjs`. **None of the three on-box surfaces moved since `d7209f8a`**, so pass 2's live-view verification still stands; CI's `Check on-box acceptance register consistency` is green on this head, and all 16 checks pass.

---

### Pass-2 dispositions

**N1 🟠 "the regression test passes in BOTH states" — the hollow half is VERIFIED RESOLVED. The second, independent break named in the same finding is NOT — it has moved, not gone.**

The hollow half is genuinely fixed, and I verified it rather than accepting it:

- The test is now **unconditional** — no `hasCommentMarker` branch, no pair of internally-consistent outcomes. Five oracles, all of which must hold.
- The pre-fix simulation is **byte-faithful**: `oldNvidiaContent` (`:228-230`) reconstructs `main`'s `nvidia-cuda.txt` exactly (`oldNvidiaContent === main's file, normalized` → `true`). It is a real revert of `cfda13a6`, not an approximation.
- **Marker removal turns it red**, at two independent assertions: ORACLE 2a (`:222`, `markerStart` → `-1`) and ORACLE 1 (`:217`, neither `1.26` nor `1.27` present anywhere else in the overlay — I enumerated every occurrence of `1.26`; there is exactly one, the new marker line).
- It now reads the constraint out of `planOrtSwap()`'s real output (`:203-211`) instead of a literal, so the guard is tied to the shipped planner. Replayed each plausible future bump against the committed overlay: `>=1.27,<1.28`, `>=1.28,<1.29`, `>=1.26.1,<1.27` and `>=1.30,<1.31` all go **red** without a `nvidia-cuda.txt` edit. That is the class pass 1 asked for, and it is now real.

**What is not fixed** is the CRLF break pass 2 filed under the same finding ("Both go away with the same rewrite"). One went away; the other was relocated into ORACLE 5.

`server/src/tts/venv-migration.test.ts:189-192, 236, 243, 249`. `base` (`:189`) and `currentNvidia` (`:190`) are read **raw**; only the overlay is normalized (`:192`). `currentHash` (`:236`) is then `computeReqHash([normalizedNvidia, base_raw])`, while `required.reqHash` (`:243`) is `resolveRequired`'s hash of **both files raw** (`venv-migration.mjs:118-123`). On an LF checkout those coincide; on a CRLF checkout they cannot.

Replayed read-only against the real `venv-migration.mjs`, same branch content, two checkouts:

```
--- LF (core.autocrlf=false / CI Ubuntu) ---
   PASS  ORACLE 1 hasVersionMarker === true   (min=1.26 present=true, max=1.27 present=false)
   PASS  ORACLE 2a/2b/2c, ORACLE 3
   PASS  ORACLE 4 stale stamp -> 'pip-in-place'
   PASS  ORACLE 5 fresh stamp -> 'noop'
   => TEST GREEN     required.reqHash = fd431a1cd83cbf89…  currentHash = fd431a1cd83cbf89…

--- CRLF (stock Git-for-Windows, core.autocrlf=true) ---
   PASS  ORACLE 1, 2a/2b/2c, 3
   PASS  ORACLE 4 stale stamp -> 'pip-in-place'
   FAIL  ORACLE 5 fresh stamp -> 'noop'   (got 'pip-in-place')
   => TEST RED       required.reqHash = 7162286bffb12bac…  currentHash = 11335a08ac62c029…
```

**Failure scenario.** A contributor clones on Git-for-Windows defaults (`core.autocrlf=true`). `.gitattributes:2` is `* text=auto` and pins `eol=lf` only for `.husky/*`, `*.sh`, `*.mjs`, `*.cjs`, `src/lib/api-types.ts`, `.env.example` and `openapi.yaml` — `server/tts-sidecar/requirements/*.txt` is not covered (`git check-attr` on the file returns `text: auto`). Both requirements files materialise CRLF. Any commit touching `server/**` runs `verify:fast:scoped` → `test:server` → this file, and `:249` fails deterministically. Same on pre-push `verify:fast:branch`. Green on CI (Ubuntu) and on this box (`core.autocrlf=false`), red for that contributor, permanently.

Two things make this this PR's defect rather than ambient noise. First, **every pre-existing assertion in this file is CRLF-safe** — `:140-148`, `:150-162` and `:169-175` all read overlay *and* base raw and compare against `resolveRequired`'s raw hash, so both sides move together. `:236` is the first assertion in the file that mixes a normalized input with a raw one, and it arrived in this PR. Second, this repo has already ruled on this exact class twice: `.gitattributes` pins `openapi.yaml` LF *because* CRLF broke line-anchored matching (#1952) and pins `.env.example` LF *because* `config:check` "fails for every Windows dev on a clean tree". That is precedent that a CRLF-only red gate is treated here as a defect, not a nit.

ORACLE 4 (`:245`) does not compensate: it holds for *any* `oldHash !== required.reqHash`, including the CRLF case where the test's own premise is already broken. ORACLE 5 is the only assertion that ties the test's computed hash to what the shipped `resolveRequired` produces — which is why it is the one that breaks.

Minimal fix (reporting, not applying): drop the normalization and stay raw throughout, deriving the marker block's bounds from the file's own line ending, e.g. `const eol = raw.includes('\r\n') ? '\r\n' : '\n'`, `raw.indexOf(eol + eol, markerStart)`, `slice(markerEnd + 2 * eol.length)`, and `computeReqHash([raw, base])`. Then `currentHash === required.reqHash` on every checkout. (Pinning `requirements/*.txt` to `eol=lf` in `.gitattributes` would also work and matches the two precedents above, but changes what every contributor's tree looks like, so it is the larger of the two.)

Catalogue shape 3 (a test that cannot fail — here, cannot *pass*, on a supported configuration).

**N2 🟡 self-contradicting `#1846` release-notes bullet — VERIFIED RESOLVED.** `docs/release-notes-next.md:330` is now byte-identical to `main`'s: the `>=1.27,<1.28` description of what #1846 actually shipped, with the duplicated parenthetical about the #2534 re-pin removed. `:334` carries the correction as its own standalone entry, and `:59`'s Upgrade note now reads `1.28.x → 1.26.x (#1846, #2534)`. No bullet contradicts itself. One residue noted below (R3).

**N3 🟡 box-local root-cause clause at `install-ort.mjs` — VERIFIED RESOLVED.** The causal claim is gone: `:204` now attributes the failure to *"the shipped torch/torchaudio cu128 pin"*, and `:211-214` gives the real bump gate — *"the constraint can only move to a CUDA-13-built onnxruntime-gpu line once the shipped torch/torchaudio pins in requirements/nvidia-cuda.txt move to a CUDA-13 wheel … the shipped venv remains CUDA-12-by-construction regardless of what CUDA toolkit is installed system-wide"*. What survives is `:201-203`'s *"Validated on this dev box: 1.26.0 constructs a working CUDAExecutionProvider InferenceSession against CUDA 12.4 + cuDNN 9 (cuDNN ships in the runtime venv via torch)"* — that is a record of **where the validation ran**, not an attribution of cause, and it no longer contradicts anything ten lines below it. Correct disposition. One prose residue noted below (R4).

**N4 🟡 stale PR-body checkbox + missing found-in-passing disclosure — VERIFIED RESOLVED.** Read from `gh pr view 2576 --json body`, not from the summary: the checkbox is struck through (`~~requirements/*.txt untouched~~`) and annotated *"no longer true, see below"* with the mechanism spelled out; an `## Also fixed, found in passing` section lists all six dispatched fixes and links [#2582](https://github.com/dudarenok-maker/Castwright/issues/2582) for the design-pass split-out.

**N5 🟡 `RELEASE_NOTES.md` describes a mechanism that does not exist — RESOLVED IN SUBSTANCE, with residue.** Both specifics pass 2 named are gone: no "compatibility check", and no "a future CUDA version that your card doesn't have yet". The bullet's second sentence — *"pinned to a newer library version than the rest of Castwright is built to work with"* — is now the true cause, stated at the right reading level. Residue below (R2).

**Pass-1 findings 1–4 and 5 remain resolved.** Nothing in `55f3d5d6` or `31f45e43` touches the delivery mechanism, the three on-box surfaces, or the issue split-out. Re-confirmed independently: [#2582](https://github.com/dudarenok-maker/Castwright/issues/2582) is **open**, `bug`-labelled, and names the decision owed; `Closes #2534.` is present in the PR body as plain prose outside any code span (and `Verify PR body links a GitHub issue` is green).

---

### New this round

**🟡 R1 [cleanup] ORACLE 1's `||` leaves the one bump shape that matters most un-guarded — a cap-only widening re-admits the CUDA-13 build and the test stays green.**

`server/src/tts/venv-migration.test.ts:216` — `normalizedNvidia.includes(minVersion) || normalizedNvidia.includes(maxVersion)`. Only **one** of the two bounds has to appear. Replayed each shape against the committed overlay, with `nvidia-cuda.txt` untouched:

```
>=1.27,<1.28      min='1.27' in file=false  max='1.28' in file=false  ->  CAUGHT (red)
>=1.28,<1.29      min='1.28' in file=false  max='1.29' in file=false  ->  CAUGHT (red)
>=1.26.1,<1.27    min='1.26.1' in file=false max='1.27' in file=false ->  CAUGHT (red)
>=1.30,<1.31      min='1.30' in file=false  max='1.31' in file=false  ->  CAUGHT (red)
>=1.26,<1.28      min='1.26' in file=TRUE   max='1.28' in file=false  ->  MISSED (still green)
```

The missed shape is the dangerous one. `>=1.26,<1.28` is what a maintainer writes to let a 1.27.x security patch through — and 1.27.x is precisely the CUDA-13 build this PR exists to exclude. The overlay still says `1.26.x`, `reqHash` never moves, every existing venv classifies `noop`, and `--force-reinstall` on a fresh box resolves to the highest match, i.e. 1.27.x. #2534 recurs, silently, with CI green. Requiring **both** bounds (`&&`) closes it; the marker line already carries only one version, so it would need to name the range.

Not blocking on its own — the dominant bump shape (both bounds move together, which is what `install-ort.mjs:209-215` instructs) is caught — but it is the same guard-enumeration weakness one level down, and it is a two-character fix.

**🟡 R2 [cleanup] `RELEASE_NOTES.md:5`'s opening sentence re-states the card-capability misattribution the rest of the bullet corrects.**

*"Castwright keeps the speech-engine libraries aligned with **your graphics card's capabilities**."* — then, immediately after: *"The speech engine was pinned to a newer library version than **the rest of Castwright** is built to work with."* Those are two different causes and only the second is real. `docs/testing/onbox-wave3-results/step-2-ort-marker.md:197-198` is explicit that the card was never the limit: *"The driver itself (610.88) did support CUDA 13 — this was a missing-runtime-library gap, not a driver/hardware limit."* Nothing is aligned to the card; the pin is aligned to the shipped torch/torchaudio cu128 stack. Dropping the first sentence, or re-pointing it at "the rest of Castwright", leaves the bullet true and no longer self-contradicting. Weaker than N5 was — it no longer invents a fix — but it is the same misattribution N5 asked to be removed.

**🟡 R3 [cleanup] `docs/release-notes-next.md:330` ships a bolded claim that is false for the release it ships in.**

*"**Existing installs on 1.28.x will step back to 1.27.x on first upgrade**"*. True of #1846 in isolation; false of v1.15.0, which is still open (`:19`) and now also contains `:334`'s re-pin to 1.26.x. This file is fed **verbatim** as the GitHub release body (its own header, `:2-6`), so a user on 1.28.x reads the correct `1.26.x` at `:59`, the wrong `1.27.x` at `:330`, and the correct `1.26.x` again at `:334`. This is residue of the remedy pass 2 itself recommended (restore `:330`, let `:334` carry the correction), so it is a judgment call rather than a regression — but a four-word parenthetical on `:330` ("superseded later in this release by the re-pin below") would remove the only sentence in the release body that tells a user the wrong destination version.

**🟡 R4 [cleanup] The rewritten `install-ort.mjs` comment left a dangling clause and a broken hyphenation.**

`server/tts-sidecar/scripts/install-ort.mjs:205-207` — *"The pin deliberately holds the runtime on the last CUDA-12 line to keep the stack CUDA-12-compatible **and nobody chooses drift on purpose**."* The trailing clause is a fragment of the sentence it replaced (`main`'s *"...a fresh install today lands 1.28.0, and nobody chose that drift on purpose"*) and no longer attaches to anything. Separately, `:213-214` wraps as `CUDA-12-by` / `-construction`, which reads as a stray leading hyphen. Purely cosmetic, but this comment block is, by its own words, the single documentation of record for where the constraint may live and when it may move.

**🟡 R5 [cleanup] One present-tense reference to the old pin survives repo-wide — and pass 1's "exactly three places" claim was wrong.**

`docs/superpowers/specs/2026-08-07-qwen-ort-namespace-chokepoint-design.md:56` — *"silently replaces the GPU runtime, driving past the deliberate `>=1.27,<1.28` pin."* Present tense, `status: draft`, describing the live mechanism rather than a dated observation. After merge it names a pin that no longer exists. I swept the whole tree at this head: this is the only such survivor. `docs/testing/onbox-wave3-results/step-2-ort-marker.md:190` is correctly past-tense and explicitly framed *"as recorded during wave-3 step 2"*; `docs/testing/ort-marker-onbox-acceptance.md:166` is a dated run record; every other `1.27.0` / `1.28.0` in `docs/superpowers/plans/` and `specs/` is a synthetic fixture version. Worth recording that pass 1's ✅ item *"the constraint string appears in exactly three places repo-wide"* was false — it enumerated code and not docs, which is how this one survived two passes. CLAUDE.md lists "a stale comment or doc sentence your change made false" as a chore fixed in the same round.

**🟡 R6 [house gate, non-blocking] No on-box register row covers the mechanism this PR actually adds.**

A39's criteria are explicit: *"run a genuine from-scratch bootstrap on the nvidia profile — **not an upgrade**, not the boot-time self-heal"* (`docs/testing/onbox-acceptance-register.md:2294-2295`). A40 is the in-app Qwen click-through; A41 is the clobbered venv. None of the three exercises the **upgrade path on an existing stamped venv** — which is exactly what `cfda13a6` adds, and exactly the half pass 1 found inert. It has been proven by unit test (the decision: `reqHash` moves → `pip-in-place`) and by code trace through `bootstrap-venv.mjs` and `apply.ts`, but never by a real `pip uninstall -y onnxruntime onnxruntime-gpu` + `--force-reinstall --no-deps onnxruntime-gpu>=1.26,<1.27` against a real venv holding 1.27.0. The PR body's on-box evidence is a **scratch venv** with a manual `pip install` — the from-scratch path, which A39 already owns.

So after merge, if the downgrade misbehaves on a real existing venv, no row directs anyone to look. Flagging rather than asserting: A39's re-run will in practice re-prove the resulting install, and the decision half is now genuinely covered by automated tests, so this is a coverage question for the author, not a defect I can demonstrate. It does **not** block, and it is not a correctness bug.

---

### Verdict

**Not ready to merge — one correctness finding is still open, and it exceeds the 3-pass cap.**

The substance of this PR is sound and I could not break it. The re-pin is right, the delivery mechanism is real through both upgrade paths, all three on-box surfaces are recorded and consistent, the release-notes pair is correct, the issue linkage and the design-pass split-out both hold, and the rewritten test is a genuine oracle for the class pass 1 named — I verified that by replay, including that removing the marker turns it red and that four of the five plausible future pin bumps do too.

What remains is **N1's second half**, which pass 2 named and which the rewrite relocated rather than removed: `venv-migration.test.ts:249` fails deterministically on a stock Git-for-Windows clone, verified by replay against the real `resolveRequired`. It is a defect this PR introduces (every other assertion in the file is CRLF-safe), in a class this repo has twice fixed permanently in `.gitattributes`, and the fix is unarguable and about six lines — no design decision, so it is not a carve-out.

This is round 3 of 3. Fixing it and pushing would normally re-trigger a pass; there is no fourth round available, so **this one goes to the repo owner**: either land the CRLF fix and merge on the strength of this pass (the fix is mechanical and its correctness is checkable by re-running the replay above), or accept the test as CI-green-only and merge with the Windows-clone breakage recorded. R1–R6 are cleanup and do not gate either way, though R1 and R5 are each a one-line change.

dudarenok-maker:
## PR review — wrap-up (head `6e4eac6c0129b68e8ff47db7b1503f31344248ab`)

Pass 3 was the final round allowed under this repo's 3-pass re-review cap (`pr-review-gate` skill). Its sole remaining correctness finding — the rewritten `venv-migration.test.ts` oracle being CRLF-inconsistent (hashed the overlay normalized but the base file raw, diverging from production's `resolveRequired` on a stock Windows/`core.autocrlf=true` checkout) and cap-blind (an `||` that wouldn't catch a future cap-only widening of the constraint) — was mechanical, well-specified, and had a clear repro with no design decision owed, so per this repo's incidental-findings policy it was fixed now rather than deferred, without spending a 4th automated review round:

- `5d101337` — fixed both CRLF-inconsistency and the cap-blind oracle in `venv-migration.test.ts`.
- `6e4eac6c` — cleaned up the remaining 🟡 non-gating residue (R2–R5) pass 3 flagged.

Independently verified after these two commits (not just trusted from the fix agents' reports):
- `venv-migration.test.ts` + `bootstrap-venv-helpers.test.ts` + `install-ort-helpers.test.ts` — 39/39 passed.
- `npm run typecheck` — clean.
- `npm run check:onbox-register` — OK, register and live view agree.
- Pushed; branch-scoped `verify:fast:branch` pre-push battery passed.

Nothing outstanding from any of the 3 review passes remains open. Proceeding to merge.

Issue #1999, opened 2026-07-31
srv — the Setup checker still calls a missing engine package "not importable"


dudarenok-maker:
## What

PR #1986 fixed a conflation on the Admin console's Voice-engine row (`/api/diagnostics`): it was telling the operator a package *"will not import"* when the package was simply **not installed**. The two faults have different remedies — install versus repair — so each is now named for what it is.

The Setup checker still has the same conflation, on a different code path.

`server/src/routes/setup-diagnosis.ts:187` renders:

> A voice engine package is not importable in the voice engine runtime.

It derives that from `kokoroPackageConfirmedBroken || qwenPackageConfirmedBroken`, which are `packageBroken` booleans. `packageBroken` is `packageOnDisk && (importOk === false || (importOk === undefined && specPresent === false))` — a single boolean that has already collapsed "the package is absent" and "the package is present but raises on import" into one value.

So unlike the diagnostics row, this is **not** a one-line reordering. The distinction does not survive as far as `setup-diagnosis`; a fix needs a signal that does not currently reach that function.

## Why it matters

Same user cost as the fault #1986 fixed, on the surface a new user is most likely to see first. Setup readiness is the first-run path — the audience least equipped to work out that "not importable" actually means "never installed", and the most likely to go looking for a repair button for software that was never there.

Lower severity than the diagnostics case in one respect: both Kokoro and Qwen ship in the default requirements (`base.txt`, `nvidia-cuda.txt`), so a genuinely missing package implies a broken bootstrap rather than a normal state.

## Suggested shape

Two separable pieces:

1. Carry the distinction far enough to be renderable. Either pass the missing-vs-broken split through alongside `packageBroken`, or have `setup-diagnosis` read `importOk` / `specPresent` the way the diagnostics row now does. This is a small design decision about what `setup-diagnosis` is allowed to depend on, not a mechanical edit — which is why #1986 left it alone rather than widening its diff.
2. While there: the shared `— repair in Model Manager` suffix on the diagnostics row is now appended to *both* phrases, so a genuinely missing package reads `Kokoro package missing — repair in Model Manager`. "Repair" is the wrong verb for something never installed. Picking the verb per-branch is a copy decision worth making deliberately across both surfaces at once.

## Notes

Found by the independent review on PR #1986 and confirmed while folding its findings. `setup-diagnosis.ts` is deliberately untouched by that PR (`git diff 5382c6be..HEAD` reports zero hits in it). Follow-up to #1965.

[...]

PR #2579, opened 2026-08-21
fix(server): add a package-broken installState to the models inventory endpoint

### LINKED ISSUE #2533: GET /api/models/inventory's installState can't distinguish broken from missing
## Summary

`GET /api/models/inventory`'s `installState` enum cannot distinguish a package that is genuinely **missing** from one that is **present but unimportable (broken)**. Both states collapse to the same `installState: 'package-missing'` value.

## Root cause

`buildModelInventory` (`server/src/routes/models-inventory.ts`) derives each engine's `installState` via `deriveEngineHealth` off a single `packageInstalled` boolean, computed as:

```js
pkgUsable = importOk ?? pkgInstalled(...)
```

Once a real load attempt has run and `importOk` is a real `false` (not `null`), the `??` operator does **not** fall through — `false ?? X` evaluates to `false` — so `packageInstalled` reads `false` regardless of whether the package is genuinely absent from disk or merely present-but-unimportable. Both states collapse to the same `installState: 'package-missing'` value at this endpoint.

## Why this isn't user-facing today

Confirmed by reading `src/views/model-manager.tsx` and `src/components/kokoro-install.tsx`: the actual Repair-vs-Install button reads a **different** endpoint, `GET /api/setup/models-status`, whose `packageFault` field genuinely does distinguish `'broken'` from `'missing'` via `classifyPackageFault`. So today's user-facing Repair/Install decision is correct.

`/api/models/inventory`'s `installState` is used elsewhere though — the Remove flow's `evaluateRemoval` guard, size/path display — where the broken-vs-missing distinction currently doesn't matter, but any *future* consumer of the inventory endpoint would silently inherit this blind spot.

## Evidence

Reproduced live during wave-3 on-box acceptance testing of register row A27 (`docs/testing/onbox-wave3-results/step-3-sidecar-install-config-reach.md`), on a throwaway robocopy of the live sidecar venv:

- Kokoro present but unimportable (forced `raise RuntimeError` in `kokoro_onnx/__init__.py`): `kokoro_import_ok: false, kokoro_package_installed: true` at the sidecar; `/api/models/inventory` reported `installState: 'package-missing'`.
- Kokoro missing entirely (package + dist-info removed): same `installState: 'package-missing'` at `/api/models/inventory`.
- `/api/setup/models-status` correctly distinguished the two (`packageFault: 'broken'` vs `packageFault: 'missing'`) in both cases.

## Decision owed

Two defensible outcomes — not fixing here, needs a design call:

1. Add a fifth `installState` (e.g. `'package-broken'`) to `/api/models/inventory` so it matches `models-status.ts`'s finer distinction.
2. Leave `/api/models/inventory` coarse by design and document that any broken-vs-missing decision must read `models-status.ts` instead.

## Source

Found during wave-3 on-box acceptance testing, register row A27. Full write-up: `docs/testing/onbox-wave3-results/step-3-sidecar-install-config-reach.md`.

dudarenok-maker:
## Summary

- Implements the repo owner's chosen design for #2533 (option 1 of the two the issue named): `GET /api/models/inventory`'s `installState` gains a 5th value, **`'package-broken'`**, so the endpoint can distinguish a voice-engine package that's genuinely missing from one that's present but fails to import — matching the finer distinction `models-status.ts`'s `classifyPackageFault` already carries internally. This **replaces** this PR's earlier approach (a sibling `packageFault` field alongside an unchanged `installState`), which an independent review found implemented neither of the issue's two named options and had real correctness bugs (see review discussion below).
- `deriveEngineHealth` (`server/src/tts/engine-health.ts`) takes an optional `packageFault` probe input and is **overloaded**: a caller that omits it (models-status.ts's own call site) gets a return type that provably excludes `'package-broken'` — enforced by assignability against a `{ packageFault?: never }` intersection member on the first overload, not a literal-freshness/excess-property check (which only catches inline object literals, not variables or spreads) — which is what lets `EngineStatus.state` keep its openapi-documented 4-value enum completely unchanged. `models-status.ts` exports a new `EngineStatusState = Exclude<EngineHealthState, 'package-broken'>` alias for that field.
- `models-inventory.ts`'s six call sites now feed `classifyPackageFault` the **raw** sidecar `*ImportOk`/`*PackageInstalled` signals (not the disk-probe-composed boolean the prior approach used), so `classifyPackageFault`'s fail-open "unknown on both axes is never a fault" rule holds for the inventory the same way it holds for `/api/setup/models-status`. The three qwen rows share one hoisted probe/fault computation instead of repeating the disk probe three times per request.
- **Bug fix, found in passing:** `sidecar-health.ts`'s `qwenPackageInstalled` coerced an absent wire field straight to `false` instead of `undefined`, unlike its kokoro/coqui/whisper siblings — defeating the fail-open rule specifically for qwen and risking a confident "missing" report for an installed-but-unprobed package. Now mirrors the sibling normalisation. This also changes `/api/setup/models-status` and the Setup checker's behavior for qwen (via `voice-engine-registry.ts`'s `liveSpecPresent`), not just the inventory endpoint — the practical blast radius is small, since the shipped sidecar always emits `qwen_package_installed`, so the changed cell is only reachable against a pre-#1965 sidecar build.
- `src/views/model-manager.tsx`: `ResidencyBadge` and the installer Repair/Install label logic both now treat `'package-broken'` the same as `'package-missing'` instead of silently falling through to "Installed" — matters most for Whisper, the one row with no `models-status` entry to fall back on.
- `src/lib/api.ts`'s hand-mirrored `ModelInventoryItem.installState` type is widened to match.

### Review history

The prior version of this PR (head `484361d`) was reviewed and found to implement a third, undecided option instead of either one #2533 named, plus a raw-vs-composed substitution bug that made `/api/models/inventory` and `/api/setup/models-status` contradict each other on the same field. Full findings: https://github.com/dudarenok-maker/Castwright/pull/2579#issuecomment-5365974691. This revision replaces that approach entirely per the owner's design choice (option 1) and fixes every finding from that review.

## Test plan

- [x] `npm run typecheck` (frontend + server) — clean.
- [x] `server/src/tts/engine-health.test.ts` — new cases for `deriveEngineHealth`'s `packageFault` param (broken → `package-broken`, missing/omitted → `package-missing`, unchanged).
- [x] `server/src/routes/models-inventory.test.ts` — rewritten `packageFault`/`installState` describe block: broken-vs-missing for all six call sites (kokoro, qwen-base, qwen-base17, qwen-design, coqui, whisper), plus a regression pinning the raw-vs-composed substitution bug and a direct fail-open assertion.
- [x] `server/src/routes/sidecar-health.test.ts` — new cases proving `qwenPackageInstalled` forwards `undefined` (not `false`) when the wire omits the field and qwen isn't loaded.
- [x] `server/src/tts/engine-presence.test.ts` — new case for the `readinessSeverity` exhaustive-switch branch this change required.
- [x] `server/src/routes/openapi-setup-parity.test.ts` — updated to guard `models-status.ts`'s new `EngineStatusState` alias (4 values) against `openapi.yaml`'s unchanged `EngineStatus.state` enum, rather than the now-5-valued `EngineHealthState`.
- [x] `src/views/model-manager.test.tsx` — new case: a `'package-broken'` row shows "Needs repair" and labels its installer "Repair", not "Installed"/"Update".
- [x] `docs/release-notes-next.md` + `RELEASE_NOTES.md` updated.
- Not applicable: `openapi.yaml` is untouched by design — the inventory endpoint is deliberately excluded from the OpenAPI contract, and `models-status.ts`'s documented `EngineStatus.state` enum is unchanged (see above). This PR corrects an existing on-box acceptance criterion (register row A27) that its own change would otherwise falsify — the register, the wave-3 results doc, and the live-view have all been updated in this PR to match (per the Before-shipping checklist's on-box acceptance step); no new acceptance is owed beyond that correction.

Closes #2533


dudarenok-maker:
## PR review — pass 1 (head 484361da33cc71ff9c43467b4baecbd1d6275ba3, depth medium)

Scope: `server/src/routes/models-inventory.ts` (+31), `server/src/routes/models-inventory.test.ts` (+72) — the whole diff. Read alongside `server/src/tts/models-status.ts` (`classifyPackageFault`, the contract this PR reuses), `server/src/tts/engine-health.ts`, `server/src/routes/models-status.ts`, `server/src/routes/sidecar-health.ts`, `server/src/tts/kokoro-install-detect.ts`, `src/lib/api.ts`, `src/views/model-manager.tsx`, and issue #2533 itself, before adversarially probing (a) whether the six call sites are wired to the right per-engine flags, (b) whether `classifyPackageFault` is being fed the arguments its contract specifies, (c) whether `installState` is silently perturbed, and (d) what a surviving mutation at each unasserted call site would look like.

Verified clean: all six call sites use their own engine's `*ImportOk`/`*PackageInstalled` pair — no copy-paste/wrong-engine bug, checked pairwise against the pre-existing `pkgUsable` calls. `installState` is genuinely untouched.

---

### 🔴 Blocking — `Closes #2533` closes an issue that explicitly reserved this as a design call, and ships a third option neither listed

`server/src/routes/models-inventory.ts:81-86` · PR body

Issue #2533's own **"Decision owed"** section reads, verbatim:

> Two defensible outcomes — **not fixing here, needs a design call**:
> 1. Add a fifth `installState` (e.g. `'package-broken'`) to `/api/models/inventory` so it matches `models-status.ts`'s finer distinction.
> 2. Leave `/api/models/inventory` coarse by design and document that any broken-vs-missing decision must read `models-status.ts` instead.

This PR takes neither. It adds a sibling `packageFault` field — a third option — and closes the issue on merge, so the decision the issue was filed to surface disappears from the board without ever being made.

This is CLAUDE.md's design-pass carve-out inverted. That carve-out is the *only* way a finding leaves a round unfixed, and its price is that the issue must name the decision owed — #2533 did exactly that, correctly. `pr-review-gate`'s judgment-call carve-out routes such a finding to the user, not to a unilateral pick by the implementing agent.

It is not merely procedural. Option 2 was *"leave the inventory coarse and make broken-vs-missing read `models-status.ts`"* — i.e. keep **one** source for this field. This PR creates a **second** source of the same field name, fed different inputs, which is precisely what produces 🟠 1 below. The owner's option 2 would have prevented that contradiction by construction; option 1 would have avoided it too by extending the existing enum instead of forking a parallel one.

Remedy is a decision, not a patch: get the design call before merge, or downgrade to `Refs #2533` and leave the issue open with its decision still named.

---

### 🟠 1 — `classifyPackageFault`'s second parameter is `specPresent` (unknown-capable); the inventory feeds it a never-`undefined` boolean, so the documented fail-open branch is dead and the two endpoints now contradict each other

`server/src/routes/models-inventory.ts:185-188`, and the five siblings at `:219-222`, `:253-256`, `:287-290`, `:328-331`, `:362-365`

`classifyPackageFault(importOk, specPresent)` (`models-status.ts:54-62`) documents its contract at `models-status.ts:34-51`, ending: *"Unknown on both axes (sidecar down / older sidecar / engine never loaded) is **never a fault**."* That rule is implemented by the final `return 'ok'` at `models-status.ts:61`, reachable only when `specPresent === undefined`.

The inventory passes `pkgInstalled(...)` as that argument. `pkgInstalled` (`models-inventory.ts:149-150`) returns `boolean` — never `undefined`. So `specPresent === undefined` **cannot occur at any of the six new call sites**, and `models-status.ts:61` is dead code for this endpoint. The unknown is consumed by the disk probe before the classifier ever sees it.

Concrete repro — fresh clone, sidecar venv not bootstrapped, sidecar not running (the app's default first-run state):

| | value | why |
|---|---|---|
| `GET /api/setup/models-status` | `engines.kokoro.packageFault === 'ok'` | `models-status.ts:76-77` forces **both** axes to `undefined` when unreachable — its own comment: *"Sidecar down → both unknown (undefined), which can never read as broken"* → `classifyPackageFault(undefined, undefined)` → `'ok'` |
| `GET /api/models/inventory` (this PR) | `kokoro.packageFault === 'missing'` | `classifyPackageFault(undefined, pkgInstalled(undefined, kokoroPackageInstalled(repoRoot)))` = `classifyPackageFault(undefined, false)` → `'missing'` |

Same machine state, same field name, contradictory values — on the one field `#1999` introduced specifically *"so the two surfaces can't disagree about which of 'missing' or 'broken' it is"* (`models-status.ts:28-30`). The PR reuses the *function* while substituting a different second *argument*, which reintroduces exactly the disagreement the function exists to prevent.

To be fair to the change: the inventory's `'missing'` is arguably the *better* answer, since it holds a real disk signal that `models-status.ts` deliberately discards when unreachable. The defect is that the divergence is unreconciled and undeclared — and the new comment claims the opposite (🟠 3). This is the root cause of 🟠 2, 🟠 3 and 🟡 4.

### 🟠 2 — all three qwen rows can report a confident `packageFault: 'missing'` for a package that is on disk

`server/src/routes/models-inventory.ts:219-222`, `:253-256`, `:287-290` · coercion at `server/src/routes/sidecar-health.ts:461`

kokoro, coqui and whisper normalise an absent wire key to `undefined` (`sidecar-health.ts:464`, `:469`, `:470`), which is what makes `pkgInstalled`'s documented old-sidecar fallback work — see the comment at `models-inventory.ts:145-148` (*"an old/unknown sidecar (undefined) falls back to the Node disk probe"*), regression-tested at `models-inventory.test.ts:528-559`.

qwen does not. `sidecar-health.ts:461` is `qwenPackageInstalled: qwenLoaded || body.qwen_package_installed === true`, so an absent key becomes `false`, not `undefined`.

Repro — a reachable sidecar whose `/health` omits `qwen_package_installed` (any build predating `server/tts-sidecar/main.py:9389`), qwen not loaded, qwen weights present, and `qwenPackageInstalled(repoRoot) === true` because the package genuinely *is* in the venv:

```
pkgInstalled(false, true)              → false   // Node probe never consulted: false !== undefined
classifyPackageFault(undefined, false) → 'missing'
```

The package is installed. The field asserts it is not, on all three qwen rows at once.

"Pre-existing" is not a deferral here (findings-triage void reasons), and in this case it also understates the change: `installState: 'package-missing'` is a coarse, deliberately ambiguous value — this PR's own comment at `models-inventory.ts:81-85` promotes the same wrong signal to *"genuinely MISSING"*. Converting an ambiguous value into a confident wrong one is a severity increase, not a carry-over.

No test covers it, and cannot as written: every qwen case in the file hand-builds `SidecarHealthResult`, so it gets the `undefined` that the production parser at `sidecar-health.ts:461` never produces.

### 🟠 3 — the new comment states a property the code does not have, and the one test that names it is built so it cannot observe its absence

`server/src/routes/models-inventory.ts:81-85` · `server/src/routes/models-inventory.test.ts:414-428`

The doc comment claims parity with `models-status.ts` *"(missing wins over broken; a real returned import wins outright; **unknown on both axes is never a fault**)"*. The third clause is false at all six call sites, per 🟠 1 — the unknown is consumed by `pkgInstalled` before the classifier runs.

The test comment repeats the same false claim:

```ts
expect(kokoro.packageFault).toBe('ok'); // unknown on both axes / nothing faulted → never a fault
```

But that test creates `.venv/Lib/site-packages/kokoro_onnx` at `:417-419`, so axis 2 is `true`, not unknown. It passes through the `specPresent === true` fall-through, never through the fail-open branch it claims to pin — green for a different reason than the one stated. Delete the `mkdirSync` and the case becomes `classifyPackageFault(undefined, false)` → `'missing'` and goes red. The only test naming the fail-open rule is constructed so it cannot detect that the rule is gone.

---

### 🟡 Minor

**4 — four of the six wired call sites have no `packageFault` assertion; the class is left armed.** `models-inventory.test.ts:383-453` covers kokoro (3 cases) and whisper (2). `qwen-base`, `qwen-base17`, `qwen-design` and `coqui` have **zero** `packageFault` assertions anywhere in the file. Surviving mutation: rewrite coqui's call at `models-inventory.ts:328-331` to `classifyPackageFault(sidecar.kokoroImportOk ?? undefined, pkgInstalled(sidecar.kokoroPackageInstalled, kokoroPackageInstalled(repoRoot)))` — coqui's row reports kokoro's fault and the whole server suite stays green. Deleting the field outright from any of those four rows also stays green. The PR body's mutation check ("removed the `packageFault` computation from the kokoro call site") only proves the kokoro cases bind — it is the mutation the existing tests already catch. `qwen-design` is the row that most needs a case: it is the only one with `loaded` hard-coded `false` (`:238`), so nothing masks its state.

**5 — the hand-mirrored client type is stale, so the field ships consumed by nothing.** `src/lib/api.ts:6490-6521` — header: *"Mirrors `server/src/routes/models-inventory.ts` (these /api/models routes are local-ops only, not in the OpenAPI contract, **so the shape is hand-mirrored rather than generated**)."* `packageFault` was not added; `installState` at `:6510` is its neighbour on both sides. The PR body reasons *"`openapi.yaml` does not document … no spec/type regeneration needed"* — that inverts the rule: *not* being in OpenAPI is precisely why a hand-mirror exists and is now owed (CLAUDE.md's "derived artifact your change staled"). Consequence: `item.packageFault` is a TS compile error in the frontend. `ResidencyBadge` (`src/views/model-manager.tsx:466-494`) already reads `packageFault`, but off `engineStatus` — the `/api/setup/models-status` payload — not off the inventory row. The mock inventory at `src/lib/api.ts:8043-8110` is a second stale mirror (sets `installState` at `:8056`, `:8072`, `:8103`; no `packageFault`), so mock mode could never surface it either.

**6 — six duplicated filesystem probes per request, and a small skew window.** Each row now calls its `*PackageInstalled(repoRoot)` probe twice — once inside `pkgUsable` (e.g. `:171-175`), once inside `classifyPackageFault` (e.g. `:185-188`). `kokoroPackageInstalled` (`server/src/tts/kokoro-install-detect.ts:15-29`) does an `existsSync` plus a `readdirSync` of the venv `lib/` per call; the three qwen rows triple it, so `qwenPackageInstalled(repoRoot)` now runs 6× per `GET /api/models/inventory`, up from 3×, on an endpoint the Model Manager polls. The two reads are also not simultaneous, so a request racing a `pip install` can yield a row whose `installState` and `packageFault` came from different disk states. Hoisting one `const specOrDisk = pkgInstalled(sidecar.XPackageInstalled, XPackageInstalled(repoRoot))` per row and feeding both consumers from it removes the duplication and the skew together.

**7 — release-notes pair silently omitted.** The diff is 2 files; neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` was touched, and the PR body does not mark them not-applicable. CLAUDE.md before-shipping step 5: *"Skip only when the change has no shippable delta … **say so explicitly rather than silently omitting**."* The test plan explicitly disposes of `openapi.yaml` and of the untouched files, so the omission reads as an oversight rather than a judgment.

---

### ✅ What is solid

- **All six call sites use the correct per-engine flags.** Checked pairwise against each row's pre-existing `pkgUsable` call: kokoro→`kokoro*`, qwen-base/base17/design→`qwen*`, coqui→`coqui*`, whisper→`whisper*`/`fasterWhisperInstalled`. No copy-paste or wrong-engine bug — the most likely defect in a six-site fan-out, and it is not present.
- **`?? undefined` is correct and necessary.** `*ImportOk` is `boolean | null | undefined` (`sidecar-health.ts:363-371`); the coalesce narrows it to `classifyPackageFault`'s `boolean | undefined`, mapping `null` ("no real import attempted") to unknown rather than to `false`. Getting this backwards would have reported `'broken'` for every engine nothing had tried to import — the dominant real-world shape.
- **`installState` really is untouched.** `deriveEngineHealth` still reads `pkgUsable`, every new line is a pure addition, and each of the five new cases pairs its `packageFault` assertion with an `installState` one.
- **`classifyPackageFault` is imported, not re-implemented** — no second copy of the precedence rule to drift.
- **The field is optional**, so the ollama analyzer rows correctly omit it and `evaluateRemoval` is unaffected.
- **The new tests assert the field's value**, not merely that the endpoint responds — the specific failure mode the dispatch asked me to check for, and it is absent.
- `Closes #2533` is present, plain, outside any code span. On-box acceptance: nothing hardware-provable ships here — correctly not applicable.

### Verdict

**Do not merge as-is.** 1 blocking, 3 significant, 4 minor.

The blocker is not a code defect and cannot be patched by a fix agent: #2533 explicitly declined to fix and named two candidate outcomes for the owner to choose between, and this PR picked a third and closes the issue. Per `pr-review-gate`'s judgment-call carve-out that suspends the fix-and-re-review loop and goes to the user — **the design call is owed before merge**, or `Closes` becomes `Refs` and #2533 stays open.

🟠 1-3 share one root cause and should be fixed together: `classifyPackageFault`'s second parameter means `specPresent`, and the inventory hands it a composed disk-or-spec boolean instead. Whichever way the design call goes, that substitution needs to be either undone (pass the raw `sidecar.*PackageInstalled`, preserving `undefined`) or made explicit — the comment must stop claiming a parity that does not hold, and the fail-open test must actually exercise the fail-open branch. 🟠 2 additionally needs `sidecar-health.ts:461` brought into line with its three siblings, or qwen's rows will keep asserting a confident falsehood.

🟡 4-7 are ordinary same-round work: four missing call-site cases, the stale `src/lib/api.ts` mirror plus its mock, the doubled probe, and the release-notes line.

dudarenok-maker:
## PR review — pass 2 (head 316bbab03037a803610d37f9ff55ef21f455362e, depth medium)

Scope: full `main...316bbab0` diff, 15 files — `engine-health.ts`, `models-status.ts`, `engine-presence.ts`, `models-inventory.ts`, `sidecar-health.ts`, `openapi-setup-parity.test.ts`, `src/lib/api.ts`, `src/views/model-manager.tsx`, both release-notes files, and the five test files. Verified independently: the six `classifyPackageFault` call sites one at a time; every consumer of `SidecarHealthResult.qwenPackageInstalled` / `qwenWeightsPresent` (`git grep`, server + frontend); every `readinessSeverity` caller; the new import edge `models-inventory → models-status` for a cycle (none — nothing but `app.ts` imports `models-inventory`); and the overload's type behaviour, compiled against this repo's own `typescript@6.0.3`.

Deltas only, per the re-review rule — pass 1's solid items are not re-listed.

---

**🔴 Pass-1 blocker — closed #2533 without making the reserved design decision → VERIFIED RESOLVED** (bookkeeping residue in 🟡 G below).
Option 1 is genuinely implemented: `'package-broken'` is a real 5th member of `EngineHealthState` (`server/src/tts/engine-health.ts:14`), selected on the `!packageInstalled && weightsPresent` branch at `engine-health.ts:65`. The sibling `packageFault` field is gone from `ModelInventoryItem` entirely — no orphaned field, no dead type.

**🟠1 — `classifyPackageFault` fed a disk-probe-composed boolean → VERIFIED RESOLVED at all six sites**, checked individually rather than at a glance:
`models-inventory.ts:180` (kokoro), `:207` (qwen, hoisted; consumed at `:228` / `:258` / `:288` for base / design / base17), `:319` (coqui), `:351` (whisper). Every one passes `sidecar.<engine>ImportOk ?? undefined` plus the raw `sidecar.<engine>PackageInstalled`; none passes `pkgUsable` / `pkgInstalled` output. The regression test at `models-inventory.test.ts:448` is a real one and red for the claimed reason — with the composed substitution `specPresent` resolves to `false` (Node probe empty), which `classifyPackageFault`'s rule 2 turns into `'missing'` → `package-missing`; raw gives `undefined`, so rule 3 gives `'broken'` → `package-broken`.

**🟠2 — qwen `sidecar-health.ts` coercion → VERIFIED RESOLVED, and correctly scoped.**
`sidecar-health.ts:469-471` now yields `true` / the wire boolean / `undefined` for the three input shapes; `qwenPackageInstalled?: boolean` was already optional, so no type widening was needed. I checked the siblings you would expect to carry the same bug: `qwenWeightsPresent` (`:472`) and `qwenBase17WeightsPresent` (`:459`) keep `=== true` **correctly** — `SidecarHealthResult.qwenWeightsPresent` has zero server-side readers (`voice-engine-registry.ts:74`'s `qwenWeightsPresent()` is the unrelated disk detector from `qwen-install-detect.ts`, a name collision), and `deriveQwenInstallState` reads `qwen_install_state`, not this field. So there is no unknown-capable consumer to fail open for, and leaving them alone is right. (Secondary effect on a *different* endpoint noted in 🟡 E.)

**🟠3 — comment + test asserting fail-open without exercising the branch → VERIFIED RESOLVED.**
`models-inventory.test.ts:463-473` now sets **both** `kokoroPackageInstalled: undefined` and `kokoroImportOk: undefined`, which genuinely reaches `classifyPackageFault`'s terminal `return 'ok'` (`models-status.ts:61`). Stated plainly since it matters for what the test is worth: it is a characterization test, not a regression test — the old composed code also produced `package-missing` here. Its own comment claims only "direct fail-open assertion", so nothing is over-claimed; the regression weight is carried by `:448`.

**🟡4 — sparse coverage (2 of 6 sites) → VERIFIED RESOLVED.** Twelve cases, a broken *and* a missing case for every id: `models-inventory.test.ts:398/409` kokoro, `:475/486` qwen-base, `:497/508` qwen-base17, `:519/530` qwen-design, `:541/552` coqui, `:563/574` whisper.

**🟡5 — stale hand-mirrored frontend type → VALUE RESOLVED (`src/lib/api.ts:6509`), CLASS STILL ARMED.** See 🟡 D.

**🟡6 — duplicated fs probes → VERIFIED RESOLVED.** `models-inventory.ts:203-207` hoists the disk probe, the `pkgUsable` composition and the fault classification once for the three qwen rows. All three blocks are unconditional, so the hoist is behaviour-identical, not merely usually-identical.

**🟡7 — missing release-notes entries → BOTH FILES UPDATED, but the user-facing one is false.** See 🟠 B.

---

## New findings

### 🔴 Blocking (correctness / gate) — A. The reworked head has never been through CI, and the PR cannot merge

`gh run list --branch fix/server-2533-installstate-fault`:

| workflow | last run | head SHA |
|---|---|---|
| Verify | 2026-08-21T04:45:54Z | **`484361da`** |
| PR title lint | 2026-08-21T04:45:54Z | `484361da` |
| PR issue link | 2026-08-21T04:45:54Z | `484361da` |
| CLA Assistant | 2026-08-21T07:05:55Z | `316bbab0` |

`484361da` is the **abandoned** sibling-`packageFault` approach. The rework push fired only `CLA Assistant`, which is `pull_request_target`; every `pull_request`-triggered workflow was skipped. `statusCheckRollup` for the current head contains exactly one check.

Cause: `gh pr view 2579 --json mergeable,mergeStateStatus` returns `CONFLICTING` / `DIRTY`, and GitHub does not dispatch `pull_request` workflows on a PR it considers conflicting. Git disagrees that there is anything to conflict over — `git merge-base origin/main 316bbab0` is `origin/main`'s own tip `ed42ed52` (the branch is strictly ahead, zero divergent commits), and `git merge-tree --write-tree origin/main 316bbab0` produces a clean tree with no conflict markers. So this looks like a stale GitHub mergeability cache rather than a real conflict — but the consequence is identical either way:

- `verify.yml` is a **required status check** on `main` and has never seen this diff;
- `pr-title-lint` / `pr-issue-link` likewise;
- the PR body's "`npm run typecheck` clean, all touched suites green" rests entirely on a local run, and the branch cannot merge as it stands.

Force GitHub to recompute (an empty commit on the branch, or close-and-reopen the PR), then confirm `Verify` reports against the current head before merging. Do not read the old green `Verify` badge as covering this code.

### 🟠 Significant (correctness — shipped user-facing artifact) — B. `RELEASE_NOTES.md` claims a UI distinction the diff explicitly does not implement

`RELEASE_NOTES.md:170` ships to users:

> **Model Manager now tells you when a voice engine's package is broken, not just missing.** … Castwright now tells them apart, so the badge you see actually matches what's wrong.

Model Manager does not tell them apart. `src/views/model-manager.tsx:495` renders the *identical* `Needs repair` badge for `package-missing` and `package-broken`; `:608` sets `isPackageMissing` true for both, so the installer toggle label at `:618` is the identical `Repair` and the hint at `:804` is the identical sentence. The diff's own comment at `:499` says so: "Same badge as package-missing".

Concrete repro — the whisper row, i.e. the one row with no `models-status` entry, which `docs/release-notes-next.md` itself names as where the gap "mattered most":

- weights on disk, sidecar reports `whisper_import_ok: false` + `whisper_package_installed: true` → `installState: 'package-broken'` → badge **"Needs repair"**, toggle **"Repair"**;
- weights on disk, sidecar reports `whisper_import_ok: false` + `whisper_package_installed: false` → `installState: 'package-missing'` → badge **"Needs repair"**, toggle **"Repair"**.

Byte-identical rendering. This PR's own tests prove it: `src/views/model-manager.test.tsx:837` (package-missing) and `:869` (package-broken) assert the same `/needs repair/i` and `/Repair/` strings.

And #2533's own body already ruled out the premise — *"Why this isn't user-facing today … today's user-facing Repair/Install decision is correct."* There was no user-visible defect here, and this PR does not create a user-visible improvement; the delta is an API value for future consumers. Per CLAUDE.md before-shipping step 5 the honest shapes are a **"Behind the scenes: …"** line (this same 1.15.0 section already uses that form twice) or an explicit skip. `docs/release-notes-next.md`'s technical entry is accurate — it is only the user-facing line that is wrong.

### 🟠 Significant (correctness — latent type-safety, plus a claim asserted in three places) — C. "PROVABLY excludes `'package-broken'`" is not a proof; it holds only for inline object literals

`engine-health.ts:50` claims the omitting caller "gets a return type that PROVABLY excludes 'package-broken', with no cast"; `models-status.ts:26` upgrades that to "a compile-time guarantee, not just a type annotation"; the PR body repeats it as the reason `openapi.yaml` needs no widening. The whole argument rests on TypeScript's excess-property check, which applies only to *fresh object literals*. Compiled against this repo's own `typescript@6.0.3` (`node_modules/typescript`, `tsc --noEmit --strict`) with the exact overload pair from `engine-health.ts:56-60`:

```ts
// (B) inline literal carrying packageFault -> correctly errors. This is the case the comment describes.
const b: EngineStatusState = deriveEngineHealth('x', { packageInstalled:false, weightsPresent:true, loaded:false, packageFault:'broken' }).state;
//    ^ TS2322: Type '"package-broken"' is not assignable to type 'EngineStatusState'.   OK

// (C) extract the probe to a variable -> NO ERROR. Narrow return type, 'package-broken' at runtime.
const probe: EngineProbe = { packageInstalled:false, weightsPresent:true, loaded:false, packageFault:'broken' };
const c: EngineStatusState = deriveEngineHealth('x', probe).state;                        // compiles clean

// (D) spread -> NO ERROR either.
const extra = { packageFault: 'broken' as const };
const d: EngineStatusState = deriveEngineHealth('x', { packageInstalled:false, weightsPresent:true, loaded:false, ...extra }).state;  // compiles clean
```

So the mechanism is a lint against one spelling, not a guarantee — a guard that enumerates syntax, sitting behind a comment that asserts certainty. Not a live bug: `models-status.ts:116` passes a fresh literal with no `packageFault` key, so today's output is correct. The exposure is the next refactor of that one call site — extract-to-variable and spread are both ordinary — and the failure would be **silent**: `EngineStatus.state` would emit `'package-broken'` while `openapi.yaml:8270` and the generated `src/lib/api-types.ts:6088` both say it cannot, and `openapi-setup-parity.test.ts` would still pass, because it compares the union's *members* against openapi.yaml, never a runtime value.

Verified fix, same compiler, same file: typing the first overload's parameter

```ts
p: Omit<EngineProbe, 'packageFault'> & { packageFault?: never },
```

makes both (C) and (D) error while the legitimate no-`packageFault` call still compiles clean. Either take that one-token tightening, or downgrade the three prose claims to what the mechanism actually delivers.

### 🟡 Minor (cleanup / chore) — D. Nothing mechanically pins `src/lib/api.ts`'s hand-mirrored union to the server's `EngineHealthState`

`src/lib/api.ts:6509` is a hand-typed six-member union whose source of truth is `engine-health.ts:11-17`. Pass 1 found this copy stale; this PR hand-corrected the value but added no guard, so the class is still armed — the next member added to `EngineHealthState` desyncs the frontend silently and no test goes red. Codegen will never cover it: `/api/models/inventory` is deliberately outside `openapi.yaml` (`models-inventory.ts:10-12`). The repo already owns the idiom for precisely this leak — `openapi-setup-parity.test.ts:84-102` documents `satisfies Record<TheServerUnion, 1>` as the way to stop a third hardcoded literal drifting, and `EngineHealthState` is now the one union in this area with no such pin.

### 🟡 Minor (cleanup — undeclared blast radius + coverage gap) — E. The qwen `sidecar-health.ts` fix also changes `/api/setup/models-status` and the Setup checker

`sidecar-health.ts:469` feeds `voice-engine-registry.ts:76` (`liveSpecPresent: (h) => h.qwenPackageInstalled`), which `models-status.ts:121-124` reads for both `packageBroken` and `packageFault`. Against a sidecar that omits `qwen_package_installed` (a pre-#1965 build), with the qwen package on disk and qwen not loaded:

- before: `specPresent === false` → `packageBroken = true` → `setup-readiness` raises blocker cause `'package-broken'` → **Setup checker blocks**;
- after: `specPresent === undefined` → `packageBroken = false` → **no blocker**.

That is the right direction — it *is* the fail-open rule, and it aligns qwen with its three siblings — so this is disclosure and coverage, not a defect. But the PR body scopes the fix to the inventory, and the new tests stop at the `sidecar-health` layer; nothing pins the `models-status` / `setup-readiness` consequence. Relatedly, "models-status.ts's own behaviour is byte-for-byte unchanged" is true of the *module* and not of the *endpoint* — worth saying precisely, since it is load-bearing for the openapi argument.

### 🟡 Minor (cleanup) — F. The PR title is stale and becomes the merge-commit subject

Title is still `fix(server): add packageFault to the models inventory endpoint`. `packageFault` is no longer added to that endpoint — the rework removed the sibling field entirely. The head commit's own subject, `fix(server): add a package-broken installState to the models inventory endpoint`, is the accurate one. Merges here are merge commits, so the PR title lands as a permanent subject. (Flagged only, not changed.)

### 🟡 Minor (cleanup / bookkeeping) — G. Issue #2533 carries no record of the ruling it reserved

The issue body still ends with *"## Decision owed — Two defensible outcomes — not fixing here, needs a design call"*, and the issue has **zero comments**. `Closes #2533` will close it with that text standing, so a later reader meets an unanswered design question on a closed ticket. The choice is recorded in the PR body; it is not recorded in what CLAUDE.md calls the canonical detail home. One comment on the issue naming the ruling (option 1, and why) closes this before merge.

---

## Verdict

**Do not merge yet.** One 🔴 and two 🟠.

🔴 A is the hard stop and is independent of the code: the required cloud checks have never run against `316bbab0`, only against the abandoned `484361da`, and the PR reads as `CONFLICTING` to GitHub while git says it is a clean fast-forward. Resolve that first — until it is, nothing else about this PR is verifiable in CI.

The rework itself is a real improvement on pass 1: all four correctness findings (🔴 blocker, 🟠1, 🟠2, 🟠3) are genuinely resolved, checked site by site rather than taken from the author's report, and the six-site coverage is now honest. The two remaining 🟠 are over-claims rather than broken logic — 🟠 B ships a user-facing sentence describing UI behaviour this diff deliberately does not implement (and that #2533 says was never broken), and 🟠 C asserts a compile-time guarantee that two ordinary refactors defeat, with a one-token fix available and verified. 🟡 D–G are cleanup.

🟠 B and 🟠 C are correctness findings, so fixing and pushing them re-triggers a pass. That would be pass 3 — the loop cap.

dudarenok-maker:
## PR review — pass 3 (head 4e912804c340a48379e40cbc9d1a07ea7e6afd3d, depth medium)

Scope: full `main...4e912804` diff, 15 files, plus the merge commit's own integrity. Depth re-derived from the current commit set — `484361da` is `feat(server)`, `316bbab0`/`9837a30e` are `fix(server)`, plus one merge commit → single-scope `feat`/`fix` → **medium**, unchanged.

Verified independently rather than taken from the brief: the merge's tree against what `316bbab0`+`9837a30e` alone would have produced; CI on the current head via `gh api repos/.../commits/4e912804/check-runs` (not the PR rollup alone); the overload's type behaviour compiled against this repo's own `typescript@6.0.3` **with a flipped-assertion control file first**, so the probe is known to discriminate before its results are quoted; every `installState` / `EngineHealthState` consumer in the tree; `evaluateRemoval`; the on-box register and its published live view.

Deltas only, per the re-review rule.

---

### Pass 1 — all four re-confirmed resolved

**🔴 Pass-1 blocker (shipped a third option, closed a reserved design call) → VERIFIED RESOLVED.** Option 1 is genuinely implemented; no sibling field survives. Bookkeeping residue is still 🟡 G, below.

**🟠1 raw-vs-composed `classifyPackageFault` inputs → VERIFIED RESOLVED**, re-checked site by site: `models-inventory.ts:180` (kokoro), `:207` (qwen, hoisted → `:228`/`:258`/`:288`), `:319` (coqui), `:351` (whisper). None passes `pkgUsable`/`pkgInstalled` output. I traced the classifier's full input lattice (importOk × specPresent × disk probe) and found no cell where `installState` and `packageFault` disagree.

**🟠2 qwen `sidecar-health.ts` coercion → VERIFIED RESOLVED.** `qwenPackageInstalled?: boolean` was already optional, so no consumer type widened.

**🟠3 fail-open test that never reached the branch → VERIFIED RESOLVED.**

**🟡4 sparse coverage → RESOLVED** (broken + missing case for all six ids). **🟡6 duplicated fs probes → RESOLVED**; I confirmed all three qwen blocks are unconditional, so the hoist is behaviour-identical, not merely usually-identical. **🟡5 → see 🟡 D. 🟡7 → see 🟠 B.**

### Pass 2

**🔴 A — reworked head never through CI → VERIFIED RESOLVED, and the merge is clean.**

`gh api repos/dudarenok-maker/Castwright/commits/4e912804.../check-runs` returns `total_count=15`, every one `conclusion: success`, attached to this head — not inherited from an older SHA. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Pass 2's "stale GitHub cache" hypothesis was wrong and GitHub was right: the branch was genuinely behind `main` and `RELEASE_NOTES.md` had a real textual conflict.

Merge integrity, checked two ways rather than asserted:

- `diff(main → 4e912804)` vs `diff(old-base 49da1ecd → 9837a30e)`, normalised for blob hashes and hunk offsets: **zero `+`/`-` deltas**; the only difference is two *context* lines around the `RELEASE_NOTES.md` insertion point. So the merge altered nothing of this PR's own content.
- `diff(9837a30e → 4e912804)` vs `diff(49da1ecd → main)`: identical file set (68 files, 2431+/145−) and identical hunk bodies. So the merge introduced exactly `main`'s changes and nothing else.
- Both `RELEASE_NOTES.md` bullets survive, `main`'s console-flash line first (`:173`, `:174`).

**🟠 B — `RELEASE_NOTES.md` claimed a UI distinction the code does not implement → RESOLVED, with one residual over-claim (🟡 I below).** The badge-distinction sentence is gone; the "Behind the scenes:" framing correctly scopes the claim to the API level; `docs/release-notes-next.md`'s technical entry reads accurately. I re-derived the UI delta row by row (`engineStatus` is `null` for whisper, `qwen-base17`, `qwen-design`, and for the three mapped rows until `modelsStatus` loads) and confirmed the badge/label/`usable`/`restartSidecar` outcomes are byte-identical for `package-missing` and `package-broken` in every one — so the note's "doesn't change what most rows show" is directionally right but still overstates. See 🟡 I.

**🟠 C — "PROVABLY excludes `package-broken`" → CODE FIX VERIFIED; the prose is still wrong in two of the three places pass 2 named.**

*Code — genuinely fixed, and I hunted for other doors rather than confirming the one demonstrated.* Compiled against `typescript@6.0.3` under `--strict`, with a **flipped-assertion control file run first** (all 7 opposite assertions error → the instrument discriminates):

| shape | overload matched | verdict |
|---|---|---|
| `const p: EngineProbe = {…packageFault:'broken'}` (pass 2's case) | **wide** | closed ✅ |
| `{ ...base, packageFault:'broken' }` inline spread | **wide** | closed ✅ |
| `Object.assign({}, base, { packageFault:'broken' })` | **wide** | closed ✅ |
| unannotated `const` literal carrying `packageFault` | **wide** | closed ✅ |
| `const p: Omit<EngineProbe,'packageFault'> = wideProbe` | narrow | **still open** |
| `function f<T extends Omit<EngineProbe,'packageFault'>>(p: T)` | narrow | **still open** |
| any structurally-narrow interface annotation | narrow | **still open** |

The three that remain all share one root — an object whose *static type* omits the key while the *value* carries it — and no type-level construct can close that in a structural system. So the code is as tight as it can be, and I am **not** asking for a further code change. No such shape exists in the tree today (`models-status.ts:116` passes a fresh literal with no `packageFault`; there is no generic forwarder), so nothing is broken.

*Prose — this is the actual finding, in three parts.*

**(a)** `engine-health.ts:52-54` still asserts the universal:

> "so ANY value assigned to that key — not just an inline object literal, but **a variable** or a spread **carrying a real `packageFault`** too — is rejected by that overload."

Row 5 above *is* a variable carrying a real `packageFault`, and it is not rejected. The claim needs scoping to "a variable **whose declared type includes** `packageFault`". `models-status.ts:20-27` is fine — its trailing clause ("this module's own call site below never passes `packageFault`") scopes it correctly.

**(b)** `engine-health.ts:58` points the reader at "the **`@ts-expect-error`** regression test in engine-health.test.ts". There is no such test. The only occurrence of that string in the file is `engine-health.test.ts:78`, inside the new test's own comment, saying the opposite:

> "so the regression is pinned at the type level, **not via `@ts-expect-error`**."

Both comments were added by the same commit (`9837a30e`), and they contradict each other. The pin is a conditional-type assignment (`engine-health.test.ts:96-98`) — which, for what it's worth, I confirmed is a real one: revert the intersection and it fails `tsc` with "Type 'true' is not assignable to type 'false'", and it is inside `server/tsconfig.json`'s `include: ["src/**/*"]`, so `npm run typecheck` genuinely reaches it. Note it does **not** fail under `vitest`, which transpiles without checking — the test's own comment says so, which is honest.

**(c)** The **third site pass 2 named was never corrected**. The PR body still reads:

> "…gets a return type that provably excludes `'package-broken'` — **enforced by TypeScript's excess-property check**, no cast"

That is precisely the mechanism pass 2 falsified, *and* it is no longer the mechanism in use: the fix works by assignability against `never`, not by literal freshness. Merges here are merge commits, so this body is the durable public record of the change.

**🟡 D — no mechanical pin on `src/lib/api.ts`'s hand-mirrored union → STILL OPEN.** `git grep "satisfies Record"` over `src/lib`, `src/views` and `models-inventory.ts` returns only `wiki-links.ts`. The value was hand-corrected again this round; the class is still armed, and codegen will never cover it (`/api/models/inventory` is deliberately outside `openapi.yaml`).

**🟡 E — the qwen `sidecar-health.ts` fix also moves `/api/setup/models-status` + the Setup checker → STILL OPEN.** No test was added at the `models-status` / `setup-readiness` layer, and neither the PR body nor either release-notes file discloses the cross-endpoint effect. I did confirm the blast radius is small in practice: the shipped sidecar always emits `qwen_package_installed` (`main.py:9389`), so the changed cell is reachable only against a pre-#1965 build — and the new behaviour is the correct one. Disclosure and coverage, not a defect.

**🟡 F — stale PR title → STILL OPEN.** Title is still `fix(server): add packageFault to the models inventory endpoint`; `316bbab0`'s own subject is the accurate one.

**🟡 G — issue #2533 carries no record of the ruling it reserved → STILL OPEN.** `gh issue view 2533 --json comments` returns `commentCount: 0`. `Closes #2533` will close it with "## Decision owed — … not fixing here, needs a design call" standing as its last word.

---

## New findings

### 🟠 Significant (correctness — merge-gated acceptance surface, + a published artifact) — H. This PR falsifies an on-box acceptance criterion and leaves the register *and the live page* saying #2533 is an open question

Neither prior pass looked at `docs/testing/`; the PR touches neither file.

**(1) A criterion this change made false.** `docs/testing/onbox-acceptance-register.md:1228-1230`, row **A27** ("A missing Kokoro/Qwen package surfaces as Install, a present-but-unimportable one as Repair"):

> Model Manager's Kokoro card must offer **Repair** (not "install", and not a silent healthy row — `installState: 'package-missing'` off a `true` find_spec is the tell)

Concrete break: that cell is exactly `kokoro_import_ok: false` + `kokoro_package_installed: true` + weights on disk. After this PR, `classifyPackageFault(false, true)` → `'broken'` → `deriveEngineHealth` → **`installState: 'package-broken'`**. A runner following A27 looks for `'package-missing'`, sees `'package-broken'`, and records a FAIL against *correct* behaviour. A27's own text warns about this exact hazard fifteen lines later:

> "(#1999/PR #2010 corrected this row's own verb … — **a stale expectation here would fail a runner against correct behaviour**, which is exactly what this update fixes.)"

**(2) A flag this change answered.** `docs/testing/onbox-acceptance-register.md:1343-1349`:

> "A related defect was found and routed, not fixed here: `GET /api/models/inventory`'s `installState` cannot itself distinguish "broken" from "missing" … but **worth a decision on widening the coarser endpoint too**. Filed as #2533"

This PR *is* that decision, shipped, with `Closes #2533` in the body.

**(3) The same sentence is live on the published artifact.** `docs/testing/onbox-acceptance-register-live-view.html:583` carries it verbatim, and that file is what is published at the URL in the register's own header (`docs/testing/onbox-acceptance-register.md:28` → `https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a`). Merging as-is leaves a public page telling readers this is an unanswered design question.

Not code — bookkeeping the change made owed, which CLAUDE.md's "a stale comment or doc sentence your change made false" puts on the same footing as a defect, and which before-shipping step 3 says moves across all three surfaces **in this diff**, with the live view republished to the recorded URL (run `npm run check:onbox-register -- --against-published <saved copy>` first, per the register's own Live view procedure). A27 itself is *not* discharged by this PR — only its `installState` tell and its Wave-3 flag need correcting.

Optional, same class: `docs/testing/onbox-wave3-results/step-3-sidecar-install-config-reach.md:352-356` ends "**Worth a decision, not fixing here** … Two defensible outcomes: (a) add the state, (b) leave `/api/models/inventory` coarse by design". That question is now answered; a one-line "resolved by PR #2579 — option (a)" pointer would close it. That file is a dated evidence record, so leaving it is defensible.

### 🟡 Minor (cleanup — prose describing a before-state that never existed) — I. Two claims describe a delta that does not exist

- `RELEASE_NOTES.md:174`: "This doesn't change what **most** rows show today". It changes what **no** row shows — verified for all six TTS/ASR rows in both the `engineStatus`-backed and `engineStatus === null` paths. "most" invites a user to hunt for a visible change there is none of; "nothing you can see today" is the honest word.
- `src/views/model-manager.test.tsx:865`: "**Before this fix** a `'package-broken'` row … fell through ResidencyBadge's checks to the plain "Installed" badge". Before this PR the server emitted no `'package-broken'` at all — that cell was `'package-missing'`, which `ResidencyBadge` already caught as "Needs repair". The counterfactual is real but it is "without the frontend half of this change", not "before this fix". `model-manager.tsx:495-500`'s own inline comment gets this right ("without this branch"); the test comment does not.

---

## Verdict

**Merge-ready on the code. Not finishable on the bookkeeping.**

The code in this PR is, as far as I can find, correct. Pass 1's four findings and pass 2's 🔴 A and 🟠 B are all genuinely resolved and re-verified from primary sources, not from the author's report. 🟠 C's *code* fix is real and I could not find a door through it that isn't inherent to structural typing. The merge commit is clean in both directions. CI is green on this head for real.

What is left is entirely documents, and it is not small:

- **🟠 H** is the one I would not merge past. A merge-gated acceptance row now instructs a runner to expect a value this PR replaced, and the *published* register page still advertises #2533 as an open design question that this PR closes. Three surfaces, one republish.
- **🟠 C's residue**: the PR body still credits the excess-property check — the mechanism pass 2 disproved and the fix replaced; `engine-health.ts:52-54` still states the guarantee as universal when three laundering shapes defeat it; and `engine-health.ts:58` sends the reader to a `@ts-expect-error` test that the test file itself says does not exist. Three sentence edits.
- **🟡 D, E, F, G** were all raised in pass 2 and none was actioned. Per `findings-triage.md`, 🟡 findings are still fixed in the round that surfaced them; they simply do not reopen the loop. G in particular is one comment on #2533 and it is the last chance to write it before `Closes` fires.
- **🟡 I** is two words and a comment.

**Loop cap.** This is pass 3 — the cap (initial + 2 re-reviews). 🟠 H and 🟠 C's residue would ordinarily re-trigger a pass once fixed; per the skill's loop-cap rule the loop **stops here and goes to you** rather than firing a pass 4. My read: nothing outstanding touches runtime behaviour or the shipped diff's logic — every remaining item is a document, a title, an issue comment, or a missing guard — so once they are made, a fourth review pass would buy no new signal and merging without one is defensible. That call is yours, not mine.

---

### Addendum (appended minutes after posting) — `main` moved during this pass

`main` advanced from `ed42ed52` to **`88fe477a`** (PR #2576) while this review was running. Consequences, all re-checked rather than assumed:

- **The head is unchanged** (`4e912804`), so the 15 green checks verified above still describe this exact diff.
- **No re-conflict.** `git merge-tree --write-tree origin/main 4e912804` exits 0 with zero conflict markers — the branch still merges cleanly into the new `main`. But GitHub currently reports `mergeable: UNKNOWN` / `mergeStateStatus: UNKNOWN` while it recomputes. Given pass 2's 🔴 A, **re-confirm `MERGEABLE` / `CLEAN` before merging** rather than reading `UNKNOWN` as fine.
- **🟠 H is unaffected.** #2576 edited both `docs/testing/onbox-acceptance-register.md` (30 lines) and `docs/testing/onbox-acceptance-register-live-view.html` (4 lines), so I re-verified every reference above against `88fe477a`: register `:28`, `:1230`, `:1343` and live-view `:583` are all unchanged, and so is `step-3-sidecar-install-config-reach.md:352`. The finding stands verbatim.
- One knock-on for whoever fixes 🟠 H: #2576 also touched the register and the live view, so the live-view republish will now carry both changes. Run `npm run check:onbox-register -- --against-published <saved copy of the live page>` first, per the register's own Live view procedure.

Issue #2583, opened 2026-08-21
ops-67 — on-box acceptance gap: ensureOrtMarker owner==='plain'/'none' cases have no acceptance row


dudarenok-maker:
PR #2578 added logging to the `owner === 'plain'` AND `owner === 'none'` branches of `ensureOrtMarker` (commit a0c5446b, refined in 6ad77ae6) — self-heal boot-time paths that fire when a venv's GPU swap marker no longer matches the venv's actual owner: `'plain'` when only the CPU `onnxruntime` build is present, `'none'` when no onnxruntime runtime is installed at all (an interrupted swap).

However, no existing on-box acceptance row (A37–A40, E7) manufactures and exercises either scenario.

- A37 (fresh NVIDIA bootstrap): exercises `owner === 'swap'`
- A38 (in-app Qwen3 install): exercises `owner === 'swap'`
- A39 (clobbered venv): exercises `owner === 'swap'` + plain distinfos present
- A40 (in-app upgrade path): exercises `owner === 'swap'`
- E7 (Pinokio update): exercises `owner === 'swap'`

Neither the `owner === 'plain'` state (CPU-only build after GPU reverted/removed) nor the `owner === 'none'` state (no onnxruntime runtime at all) was ever separately confirmed on real hardware. These can occur when:
1. A GPU installation is deliberately downgraded to CPU (`pip install onnxruntime` after uninstalling `onnxruntime-gpu`) — reaches `'plain'`.
2. The GPU files are deleted but our swap marker remains (from before this fix) — reaches `'plain'` if plain onnxruntime is present, `'none'` if nothing is.
3. A forced downgrade or an interrupted swap during a failed repair attempt — either state, depending on what's left on disk.

**Decision owed:** should new on-box acceptance row(s) be added specifically to exercise the `owner === 'plain'` and `owner === 'none'` logging branches? If yes, each row should:
- Deliberately manufacture the target state in a bootstrapped venv (downgrade to plain, or remove both packages entirely, for the two cases respectively)
- Confirm the correct case-specific log line fires at server boot (the `'plain'` case names no remedy — Kokoro runs without GPU acceleration; the `'none'` case names the exact repair command)
- Confirm the marker is correctly deleted
- Confirm Kokoro's behavior matches each case (CPU-only execution for `'plain'`; cannot load at all for `'none'`)

[...]

Issue #2551, opened 2026-08-20
ops — wave 4: retire the rows that can be retired, and wire the fix-blocked rows to their fixes


dudarenok-maker:
Wave 4 of the campaign to clear `docs/testing/onbox-acceptance-register.md` (parent #2435). Waves 1-3 are merged; **the register still stands at 74 owed and no wave has yet retired a single row.** That is the thing this wave exists to change.

## Why nothing has come out yet

- **Wave 1** (#2440, audit) proposed **zero** discharges by design — an audit verdict is not an acceptance run, and the instrument was built to fail closed.
- **Wave 2** (#2453, sitting packs) says outright "No row is retired by this wave" — it binned the 49 operator rows into 8 sittings.
- **Wave 3** (#2497) ran the 21 agent-runnable rows and retired **none**. It was not wasted — it produced live evidence on ~11 rows, caught its own mislabel of A29, and found **five real defects** (#2533, #2534, #2535, #2536, #2537) — but every row it touched hit one of four walls: the acceptance ran and *failed*, a precondition was genuinely missing, the criterion needs a browser or human ears, or it was not attempted.

## What this wave does

Two tracks, both aimed squarely at removing rows.

**Track A — retire what can be retired now** (steps 1-4, 6). Five rows where the remaining work is small, agent-runnable, and was either mis-assessed as blocked or simply left unfinished:

| Row | Why it is reachable now |
|---|---|
| **G1** | **Newly unblocked and nobody has noticed.** PR #2488 (fixes `parseRegister`, which silently dropped every real register row) merged **2026-08-20 06:45 UTC**. The last `quarantine-health.yml` dispatch was **2026-08-17**, before that merge. The workflow has `workflow_dispatch`. A dispatch now takes the non-empty path for the first time ever. |
| **E8** | **Wave 3's blocker looks false.** It concluded a second ffmpeg build needs WSL or a container runtime. Every ffmpeg call in the server spawns the bare string `'ffmpeg'` (`server/src/export/build-m4b.ts:336`, `server/src/routes/clip.ts:104`, `server/src/audio/measure-loudness.ts:83`) — **resolved from `PATH`**. A portable second build in a scratch directory, `PATH`-prefixed for one command, meets the row's criterion exactly and installs nothing. |
| **A27** | Shrunk by wave 3 to **two** remaining checks, on a throwaway-venv harness wave 3 already built and documented. |
[...]

Issue #2296, opened 2026-08-12
ops-57 — every direct-execution guard in scripts/ onto one shared helper


dudarenok-maker:
## What

44 scripts under `scripts/` each hand-rolled the same "was I run directly?" guard, in four different spellings. The naive forms miss whenever the invocation path crosses a symlink (POSIX) or a junction (Windows): Node's ESM loader realpaths the main entry when deriving `import.meta.url`, while `process.argv[1]` keeps the path exactly as invoked. When the guard misses, `main()` never runs and **the process exits 0 with no output at all** — a vacuous green, not a visible failure.

This is not hypothetical. It is the root cause of #2291, where it cost days of red `cross-os.yml` runs via `run-sidecar-tests.mjs` (fixed in PR #2293). An independent review of that PR then demonstrated it live against a second script:

```
node <junction>/scripts/ci-scope.mjs --files=src/app.tsx   ->  exit 0, 0 bytes of stdout
```

`ci-scope.mjs` is worse than silent: the `detect` job emits nothing, so every downstream leg's `if:` evaluates false and the legs quietly do not run, with only the `ok` sentinel to catch it. `verify-cache.mjs` (the verify runner itself), `build-release-zip.mjs`, `start-app-prod.mjs` and `launch-sidecar.mjs` were all vulnerable the same way.

**How the scope grew, which is the point:** the review named 2 sites. Grepping found 22. Migrating those surfaced 14 more in a comparison shape the first pass could not see. A further 7 used a form that a basename fallback happened to rescue. A guard that enumerates known-bad spellings loses one per round; this needs a positive invariant.

## Acceptance

- [ ] One shared helper realpathing **both** sides; every hand-rolled guard routed through it.
- [ ] A convention test asserting a **positive** invariant (any script referencing both `import.meta.url` and `process.argv[1]` must use the helper), not a blacklist of known spellings, with an empty and non-vacuous allowlist.
- [ ] Each migration proven **behaviourally** — a real junction, asserting observable output — not structurally. A guard migrated wrong turns a working CLI into a silent no-op that exits 0, and module-level tests cannot see it because they import rather than spawn.
- [ ] `npm run test:hooks` green.

## Key files

- `scripts/lib/is-main-module.mjs` (new)
- `scripts/tests/entry-point-guard-convention.test.mjs` (new)
- `scripts/tests/entry-point-guards.test.mjs`, `scripts/tests/is-main-module.test.mjs` (new)
[...]

PR #2588, opened 2026-08-21
fix(server): venv-migration ORT pin re-stamp test fails on CRLF checkouts

### LINKED ISSUE #2586: fix(server): venv-migration ORT pin re-stamp test fails on CRLF checkouts
## What

`server/src/tts/venv-migration.test.ts`'s "side-28 ORT pin re-stamp (issue #2534)" test (added by PR #2576) fails on any checkout where `nvidia-cuda.txt` / `base.txt` have CRLF line endings (e.g. this repo's Windows boxes with `core.autocrlf=true`), even though the requirements text is unchanged. It passed in PR #2576's own CI (Ubuntu runner, LF checkout) so it shipped looking green, and would also fail on the Windows leg of `cross-os.yml`.

## Repro

On a fresh Windows clone of `main` (`core.autocrlf=true`):

```
cd server && npx vitest run src/tts/venv-migration.test.ts
```

```
FAIL  onnxruntime-gpu pin bump must be coupled to reqHash via requirements text change
AssertionError: expected 'pip-in-place' to be 'noop'
```

## Cause

`venv-migration.test.ts:196-197,254` independently reads and CRLF-normalizes the requirements files to compute a "current" hash:

```ts
const normalizedBase = base.replace(/\r\n/g, '\n');
const normalizedNvidia = currentNvidia.replace(/\r\n/g, '\n');
...
const currentHash = computeReqHash([normalizedNvidia, normalizedBase]);
...
const freshStamp = { pythonTag: required.pythonTag, profile: required.profile, reqHash: currentHash };
expect(decideVenvAction({ stamp: freshStamp, required })).toBe('noop');
```

But production code (`server/tts-sidecar/scripts/venv-migration.mjs:116-123`, `resolveRequired`) hashes the **raw, unnormalized** file bytes:

```js
const overlay = readFileSync(join(sidecarDir, 'requirements', overlayFileForProfile(profile)), 'utf8');
const base = readFileSync(join(sidecarDir, 'requirements', 'base.txt'), 'utf8');
return { pythonTag, profile, reqHash: computeReqHash([overlay, base]) };
```

On a CRLF checkout, the test's independently-normalized `currentHash` never equals `required.reqHash` (computed from raw CRLF bytes), so `decideVenvAction` correctly reports a mismatch (`pip-in-place`) against the test's own miscomputed expectation of `noop`. The test's own comment claims normalization "ensures the test produces identical results on LF and CRLF checkouts" — false, because only the test's *own* computation is normalized; production's is not.

Oracles 3 and 4 (earlier in the same test) are unaffected — they assert **inequality** between old and current hashes, which holds regardless of the extra CRLF noise.

## Fix

Oracle 5 doesn't need to recompute anything — it's testing "a venv already at the current reqHash needs no action", and `required.reqHash` (from the same `resolveRequired()` call already made earlier in the test) **is** that value. Reuse it instead of independently re-deriving it:

```ts
// ORACLE 5: A venv with the CURRENT hash should see 'noop' (no action needed).
const freshStamp = { pythonTag: required.pythonTag, profile: required.profile, reqHash: required.reqHash };
expect(decideVenvAction({ stamp: freshStamp, required })).toBe('noop');
```

This sidesteps the CRLF mismatch entirely since it stops re-deriving a hash by a different code path than production uses.

## Found via

Surfaced while merging `main` into `feat/server-2246-language-recurrence` for PR #2492 (unrelated branch) — the pre-commit hook's full server-test run hit this pre-existing failure. Confirmed pre-existing by running the same test against a clean `main` clone with no merge involved.

dudarenok-maker:
Closes #2586

## Summary

Root cause: `server/tts-sidecar/requirements/*.txt` had no `.gitattributes` line-ending pin, so `resolveRequired()`'s `reqHash` (`server/tts-sidecar/scripts/venv-migration.mjs`) — hashed from raw file bytes, persisted to the venv stamp, and compared across checkouts/machines — differed between an LF and a CRLF checkout of byte-identical requirements content. A `core.autocrlf=true` Windows clone (Git for Windows' default) materializes these files as CRLF, producing a different `reqHash` than an LF checkout of the same commit, triggering a spurious `pip-in-place` reinstall (or worse, silently skipping a real one) regardless of whether requirements actually changed. This repo has fixed the identical same-commit-different-hash class four times before — `openapi.yaml`, `.env.example`, `src/lib/api-types.ts`, `*.mjs`/`*.sh` — each via the same LF pin; `requirements/*.txt` was the missed instance.

Four review rounds surfaced deeper problems than the initial CRLF-checkout symptom, all now addressed. 13 files changed in total.

## What changed

- **`.gitattributes`**: pins `server/tts-sidecar/requirements/*.txt` to `text eol=lf`. Documents that an **already-existing** CRLF checkout does not self-heal from a `git pull`/`git checkout --` — git's clean filter treats on-disk CRLF as matching the LF-normalized blob, so the correct remediation is a physical `rm` of the affected files followed by `git checkout --`, not `git rm --cached` (which errors and fixes nothing — the originally-documented recipe was itself broken until this was caught in review).
- **`server/src/tts/venv-migration.test.ts`**: Oracle 3 and Oracle 5's logic restored to `main`'s original shape (an earlier commit on this branch had rewritten both to compare `required.reqHash` against itself, making Oracle 5 a tautology and Oracle 3 vacuous on CRLF checkouts). They are not byte-identical to `main` — both now hash `speaker-qa.txt` alongside overlay/base, extending correctly to the 3-file hash this PR introduces (see below) rather than staying pinned to the pre-existing 2-file shape. With line endings now pinned at checkout, both oracles are correct and have full teeth on every platform. Also: a new test asserting the pinned requirements files never carry a CR byte on disk; a `.gitattributes`-pin-rule assertion that reddens on rule deletion regardless of the checking-out box's own line-ending config (so the guard is meaningfully testable even on `ubuntu-latest`); the hardcoded 4-file list switched to `readdirSync` enumeration.
- **`scripts/verify-cache.mjs` / `server/vitest.config.ts`**: now watch `.gitattributes` and `server/tts-sidecar/requirements/**`, matching the existing `openapi.yaml` pattern, so `scripts/ci-scope.mjs` correctly selects the `test:server` leg for a change to either — previously such a diff selected `step_test_server=false`, meaning this PR's own change class could never have triggered CI to run the very test it fixes.
- **`server/tts-sidecar/scripts/venv-migration.mjs`**: `resolveRequired()` now hashes `speaker-qa.txt` (pip-installed by all three requirements overlays) alongside the overlay and base — it was previously invisible to `reqHash`, reproducing #2534's exact failure mode. Real behavior change: every existing venv's stamp now evaluates to `pip-in-place` once, picking up `speaker-qa.txt`'s current pins.
- **`server/src/upgrade/zip-validate.ts`**: `validateUpgradeZip` is a SEPARATE producer of the same `reqHash`, consumed by the zip/self-upgrade path. It was not updated when `speaker-qa.txt` joined `resolveRequired`'s hash, leaving #2534's failure mode live on the real release/upgrade path. Now reads `speaker-qa.txt` out of the zip and hashes `[overlay, base, speakerQa]` in the same order `resolveRequired` does, with a new parity test asserting the two producers agree over identical content. Note: this producer only ever reads `nvidia-cuda.txt` as the overlay, so its hash matches `resolveRequired('nvidia', ...)` specifically, not the `cpu`/`amd-rocm` profiles (a pre-existing asymmetry, documented in code, not introduced by this PR).
- **`docs/testing/onbox-acceptance-register.md`** (+ live-view HTML): a row recording the one-time reinstall-wave behavior as acceptance owed on real hardware, split into two independently-sufficient test paths (Pinokio Update writing `.venv-stamp.json`, vs. the in-app zip-upload path writing `.req-hash`) since the two paths write different files and the zip-upload path needs a real pre-existing hash seeded first to avoid a false-positive first-run pass.
- **Release notes**: `docs/release-notes-next.md` and `RELEASE_NOTES.md` both updated — this is no longer purely-internal tooling, since the `speaker-qa.txt` hash change causes a real, once-only reinstall that end users will observe.

## Verification

- All four of the original review's `resolveRequired` mutations (reversed hash order, dropped `base.txt`, hardcoded constant hash, wrong overlay file for the profile) confirmed to redden the reverted oracles.
- The `.gitattributes`-pin-rule assertion confirmed to redden on rule deletion (platform-independent).
- The CR-byte regression test confirmed red/green against a deliberately CRLF-corrupted file.
- `zip-validate.test.ts`'s new parity test confirmed to fail when the two hash producers are deliberately drifted apart (order swap, dropped file, or a future 4th requirements file added to only one producer), pass when in sync.
- The corrected `.gitattributes`/CRLF-checkout remediation command verified against a reproduced stale-CRLF scenario; the old, broken command reproduced as broken for contrast.
- `scripts/ci-scope.mjs` confirmed to flip `step_test_server` `false`→`true` for both a `.gitattributes`-only diff and a `requirements/*.txt`-only diff.
- `cd server && npx vitest run src/upgrade/zip-validate.test.ts src/tts/venv-migration.test.ts` — 40/40 green.
- `npm run typecheck` — clean.
- `npm run check:onbox-register` — register and live-view agree.

Four independent review-gate rounds ran on this PR (findings + dispositions in the PR comments below); each surfaced and closed a real issue in the prior round rather than rubber-stamping it.


dudarenok-maker:
## PR review — pass 1 (head 38b4d019, depth medium)

Scope: `server/src/tts/venv-migration.test.ts` — the only changed file (+8/−4) · verified: `npx vitest run src/tts/venv-migration.test.ts` → 21/21 green, plus a standalone probe that replays the pre-fix and post-fix oracle logic against the **real** `venv-migration.mjs` over synthetic LF and CRLF sidecar trees, before adversarially probing whether the fix keeps its teeth and whether the CRLF sensitivity actually ends here.

**First, a caveat on "it passes on Windows now".** This checkout does not reproduce the bug at all: `.git/config` carries `core.autocrlf=false` + `core.eol=lf`, locally overriding the system-wide `core.autocrlf=true` in `C:/Program Files/Git/etc/gitconfig`. `requirements/*.txt` materialises as LF here, so vitest is green on `main` **and** on this branch and distinguishes nothing. All CRLF evidence below is from the probe.

The reported bug is real and this PR does stop the symptom — probe, oracle-by-oracle:

| variant | checkout | O3 | O4 | O5 | result |
|---|---|---|---|---|---|
| pre-fix (`main`) | LF | PASS | PASS | PASS | GREEN |
| pre-fix (`main`) | **CRLF** | PASS | PASS | **FAIL** | **RED** ← #2586 reproduced, Oracle 5 only, exactly as diagnosed |
| post-fix (this PR) | LF | PASS | PASS | PASS | GREEN |
| post-fix (this PR) | **CRLF** | PASS | PASS | PASS | GREEN |

The problem is *how* it stops it.

---

### 🟠 Significant — Oracle 5 is now a tautology; four independent breakages of `resolveRequired` go undetected (correctness bug)

`server/src/tts/venv-migration.test.ts:254-259`

```ts
const freshStamp = { pythonTag: required.pythonTag, profile: required.profile, reqHash: required.reqHash };
expect(decideVenvAction({ stamp: freshStamp, required })).toBe('noop');
```

Every field of `freshStamp` is copied from `required`, and `decideVenvAction` (`venv-migration.mjs:38-44`) compares exactly those three fields. The assertion is true by construction for any value `resolveRequired` returns. It is also a literal duplicate of the existing unit test at `venv-migration.test.ts:62` (`'all match → noop'`).

What Oracle 5 used to prove was that production's `resolveRequired` hashes *the right files, in the right order, from the right overlay* — because the test derived the same hash independently. I ran four mutations of `resolveRequired` through both variants on an LF checkout:

| mutation of `resolveRequired` | pre-fix (`main`) | post-fix (this PR) |
|---|---|---|
| M1 — hash `[base, overlay]` (documented order reversed) | **RED — caught** | GREEN — missed |
| M2 — drop `base.txt` from the hash entirely | **RED — caught** | GREEN — missed |
| M3 — return a hardcoded constant hash | **RED — caught** | GREEN — missed |
| M4 — read the `cpu` overlay while reporting `profile: 'nvidia'` | **RED — caught** | GREEN — missed |

M3 is the headline: after this PR, `resolveRequired` can return a completely fabricated `reqHash` and the entire "side-28 ORT pin re-stamp (issue #2534)" test still goes green — on **every** platform. That is the catalogue's shape #3 (*a test that cannot fail*) and #5 (*an acceptance criterion that would pass on a null observation*).

The commit message says this "makes the test platform-independent by construction". It made it *outcome-independent* by construction. Deleting the assertion is not the same as making it portable.

**Remedy:** keep an assertion that pins production's hash to a test-derived one, computed from the **same representation production uses** — i.e. do the marker surgery on the raw text and hash raw text, not normalized text:

```ts
const currentHash = computeReqHash([currentNvidia, base]); // RAW, not normalized
expect(required.reqHash).toBe(currentHash);                // <- the coverage this PR removed
```

That is CRLF-safe *and* keeps all four mutations red. (See the 🟠 below — if `.gitattributes` is fixed instead, the original `main` code needs no change at all.)

---

### 🟠 Significant — Oracle 3 did not have this bug, was changed anyway, and is now vacuous on CRLF (regression introduced by this PR)

`server/src/tts/venv-migration.test.ts:243-247`

Before this PR, Oracle 3 compared `oldHash` against `currentHash` — **both** derived from normalized content. It was fully CRLF-insensitive and correct on every checkout. #2586 itself says so: *"Oracles 3 and 4 … are unaffected."* The PR body's justification — "Also updated Oracle 3 to use `required.reqHash` for the same reason" — cites a reason that does not apply to Oracle 3.

The change is not neutral. Oracle 3 now compares a **normalized** hash against production's **raw** hash, so on a CRLF checkout the two differ because of every line ending in the file, independently of the marker. Its stated contract — *"the marker comment must affect the hash"* — is no longer what it tests there. Probe, feeding it a **null observation** (marker left fully intact, i.e. zero marker effect):

```
marker NOT removed at all; post-fix Oracle 3 (oldHash !== required.reqHash)
  on CRLF => PASSES anyway (vacuous)
  on LF   => fails (has teeth)
```

So after a PR whose stated goal is platform-independence, Oracle 3's *meaning* is now platform-dependent — with the vacuous branch landing on exactly the platform the PR exists to support.

Oracle 4 (`:249-252`) sits in the same shape and this PR leaves it there: `staleStamp.reqHash` is still `oldHash` (normalized) compared against `required.reqHash` (raw), so on CRLF it returns `'pip-in-place'` for line-ending reasons rather than marker reasons. That makes the PR body's summary — *"it stops re-deriving hashes via a different code path than production uses"* — false as written: `oldHash` at `:241` is still exactly that, and two oracles still depend on it.

---

### 🟠 Significant — the actual defect is in production, not the test: `reqHash` is a function of the checkout's `core.autocrlf`, not of the requirements content

`server/tts-sidecar/scripts/venv-migration.mjs:116-124` · `.gitattributes:2` · `server/src/upgrade/zip-validate.ts:206-208`

The test was not wrong to notice a hash mismatch — it was reporting a real one. Probe, same requirements text, two line-ending regimes:

```
LF   tree reqHash = 064ba5e99bf0235d4948128ffbb3be997a5c60ebc4461e0004b251a1d96a3c47
CRLF tree reqHash = 346a232d35efe6c1aae1a356c9325fb1d75f2b928f032dea116d23f0c640d371
identical requirements text -> same hash?  false
decideVenvAction(stamp from LF tree, required from CRLF tree) = 'pip-in-place'
```

`reqHash` is **persisted** (`.venv-stamp.json`, `.req-hash`) and compared **across trees**, so this is not academic:

- `.gitattributes:2` is `* text=auto` with no override for `server/tts-sidecar/requirements/*.txt`. A default Git-for-Windows clone (`core.autocrlf=true`, as this box's system gitconfig has) checks those files out as CRLF; Linux/macOS get LF. **The same commit yields two different `reqHash` values depending on who cloned it.**
- `zip-validate.ts:206-208` documents the invariant *"so `ctx.reqHash` is byte-identical to the venv stamp's `reqHash`"*. That invariant is false across regimes. The release zip is built on `ubuntu-latest` (`release.yml:285`) → LF; a venv stamped from a Windows source checkout → CRLF. `apply.ts:134` then sees a mismatch and runs a full `pipInstall` on upgrade with the requirements byte-identical modulo line endings.
- Nothing tests this. `zip-validate.test.ts:171` asserts internal consistency against fixture constants, never line-ending stability.

This repo has fixed this exact class four times already, in this exact file, by pinning the artifact to LF — `openapi.yaml` (#1952), `.env.example`, `src/lib/api-types.ts`, `*.mjs`/`*.sh` — each with a comment explaining that a cross-checkout-compared artifact cannot honour `core.autocrlf`. `requirements/*.txt` is the same shape and was missed. Catalogue shape #10: *one instance fixed, the class left armed*.

Worth noting what this implies for the rest of the review: **if the `.gitattributes` pin lands, `main`'s original test needs no change at all** — probe row 1 shows pre-fix/LF is green with full teeth — and both 🟠 findings above dissolve rather than needing separate fixes.

**Decision owed (this is the one finding I'd route through a design pass rather than a straight fix):** pin at checkout (`server/tts-sidecar/requirements/*.txt text eol=lf`) versus normalize inside `resolveRequired`/`computeReqHash`. They are not equivalent — normalizing changes the hash for *every* existing venv on *every* platform, forcing a one-time `pip-in-place` wave and requiring a lockstep change in `zip-validate.ts` or the documented byte-identity invariant breaks; pinning is inert for LF users and self-heals CRLF users with a single reinstall. That user-visible reinstall wave is the call to make, and the issue should name it.

---

### 🟡 Minor

1. **The comment that caused the bug is still there, verbatim** — `venv-migration.test.ts:191-195`: *"Both files must be normalized consistently so the test's hash computations match production's … Normalizing both ensures the test produces identical results on LF and CRLF checkouts."* #2586 explicitly identifies this sentence as false. This PR left it while removing the only place the claim was attempted, so it is now doubly wrong and it is the exact licence a future maintainer would read before adding an Oracle 6 that compares another normalized hash to `required.reqHash` and reintroduces #2586 on the Windows leg. Catalogue shape #9. Fix it in this round per the repo's chore rule.

2. **Release-notes step not addressed.** The before-shipping checklist requires the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair *or* an explicit not-applicable, and says to "say so explicitly rather than silently omitting". A test-only fix plausibly qualifies for the skip, but the PR body doesn't state it. (Note it isn't purely internal: this test red-lights `verify:quick` on the `windows-latest` legs of both `cross-os.yml` and `release.yml`, so on `main` today it gates release publication.)

3. `const required` was hoisted from `:250` to `:246` purely to serve the Oracle 3 rewrite. If Oracle 3 reverts, the hoist should revert with it rather than being left as unexplained reordering.

---

### ✅ What is solid

- The diagnosis in #2586 is accurate and reproduces exactly as written — Oracle 5 only, CRLF only, `'pip-in-place'` vs `'noop'`. I confirmed the file/line attribution and the `resolveRequired` raw-bytes cause independently.
- `Closes #2586` is present, plain, and outside any code span. `pr-issue-link` will pass.
- Depth `medium` is the right tier (single-scope `fix`), title matches the commit convention, one file, no stray edits, no orphaned imports — `computeReqHash` and `resolveRequired` both remain used.
- Oracle 1 (`:220-222`) still genuinely couples an `install-ort.mjs` pin bump to `nvidia-cuda.txt`, so the test's *headline* purpose survives this change even though Oracle 5's does not.
- The instinct behind the fix — "don't re-derive a value through a second code path" — is correct. It was applied to the wrong side of the comparison: the normalization should have been dropped, not the assertion.

---

### Verdict

**Request changes.** The fix works in the narrow sense that CI goes green on Windows, but it achieves that by removing the assertion rather than the platform sensitivity, and it degrades a second oracle that was never affected. Three correctness findings, all reproduced empirically:

1. Oracle 5 is vacuous — four distinct breakages of `resolveRequired` now ship undetected (was: all four caught).
2. Oracle 3 was sound, was changed for a reason that doesn't apply to it, and is now vacuous on CRLF specifically.
3. The real defect is production's line-ending-sensitive `reqHash`, persisted and compared across trees, with a documented byte-identity invariant in `zip-validate.ts` that it violates — untested, and with four in-repo precedents for the one-line remedy.

Cheapest coherent path: fix #3 first (`.gitattributes` pin + a test asserting `reqHash` is stable across line-ending regimes), then revert this PR's test hunk entirely — `main`'s oracles are correct once the input stops varying. If the pin is not the chosen route, #1 and #2 each need a real fix, and #1's must pin `required.reqHash` to a raw-derived hash so the four mutations go red again.

<sub>Automated adversarial pass per `.claude/skills/pr-review-gate`. Findings only — no files were modified by this review.</sub>

dudarenok-maker:
## PR review — pass 2 (head ddcf3c2a, depth medium)

Scope: `.gitattributes`, `docs/release-notes-next.md`, `server/src/tts/venv-migration.test.ts` (+31/−5 vs `main`) · verified in **two throwaway clones off this repo, one with `core.autocrlf=true` + `core.eol=native` (Git for Windows' default, which the `windows-latest` runner inherits) and one at `main` as a pre-fix control** — because neither the primary checkout nor the branch worktree reproduces the bug (both carry the local `core.autocrlf=false`/`core.eol=lf` override pass 1 flagged). Positive control first: `main` in the CRLF clone reddens exactly as #2586 describes (`expected 'pip-in-place' to be 'noop'` at Oracle 5, and nothing else), and this branch in the *same* clone is 22/22 green. The instrument can fail.

---

### Pass-1 dispositions

**🟠 1 — Oracle 5 became a tautology: VERIFIED RESOLVED.** Not by reading the revert — by re-running all four of pass 1's mutations against `resolveRequired` (`venv-migration.mjs:116-124`) with this PR's test file, one at a time, each reverted after:

| mutation of `resolveRequired` | ORT test | fails at |
|---|---|---|
| M1 — `computeReqHash([base, overlay])` (documented order reversed) | **RED** | `venv-migration.test.ts:271` |
| M2 — drop `base.txt` from the hash | **RED** | `:271` |
| M3 — return a hardcoded constant hash | **RED** | `:271` |
| M4 — read the `cpu` overlay while reporting `profile: 'nvidia'` | **RED** | `:271` |

All four land on Oracle 5 (`expected 'pip-in-place' to be 'noop'`) — red for the reason claimed, not incidentally. Baseline unmutated: 22/22.

**🟠 2 — Oracle 3 needlessly changed: VERIFIED RESOLVED, byte-for-byte.** `git diff origin/main...ddcf3c2a -- server/src/tts/venv-migration.test.ts` is exactly two hunks — the new test at `:169-182` and the comment rewrite at `:206-211`. Everything else, Oracles 1–5 and the `const required` placement included, is the identical blob. That also disposes of **🟡 3 (the hoist)** — it reverted with Oracle 3, as asked.

**🟠 3 — the production root cause: VERIFIED RESOLVED.** Measured, not read:

```
git check-attr text eol  →  text: set / eol: lf  on base.txt, nvidia-cuda.txt,
                            cpu.txt, amd-rocm.txt AND speaker-qa.txt
CR bytes on disk, autocrlf=true clone
  this branch : base 0   nvidia-cuda 0   cpu 0   amd-rocm 0
  main        : base 56  nvidia-cuda 84  cpu 27  amd-rocm 30      ← control
  same clone, same regime: package.json CR=156, requirements.txt CR=5  ← positive control,
                            so the clone really is CRLF and the pin is what's excluding these
committed blobs: CR=0 on every one (so `eol=lf`, which never converts CRLF→LF on
                 checkout, has something to work with)
```

The rule is placed and worded in the style of the four existing pins. `Closes #2586` is present, plain, outside any code span.

**🟡 1 — the stale/false comment: VERIFIED RESOLVED.** `:206-211` now attributes the normalization to the marker-finding `indexOf`/`substring` ops and explicitly disclaims the hash-parity role. Accurate as written.

**🟡 2 — release notes: PARTIALLY RESOLVED.** The `docs/release-notes-next.md` entry landed. The `RELEASE_NOTES.md` skip is argued from a precedent that is **inverted** — see 🟡 below.

---

### 🟠 New — the new guard cannot fail in the only lane that gates merge, and CI does not select it on the diff that would break it (correctness bug)

`server/src/tts/venv-migration.test.ts:169-182` · `.gitattributes:37` · `scripts/verify-cache.mjs:378-408` · `server/vitest.config.ts:190-217`

The new test asserts the pin's *materialisation* (no CR byte on disk). Materialisation is a property of the checking-out box, not of the repo — so on any LF-native box the assertion is true whether or not `.gitattributes:37` exists. Measured by deleting the pin line outright and re-checking-out:

| regime | `.gitattributes:37` | `requirements/*.txt … never contain CRLF` |
|---|---|---|
| `autocrlf=true` fresh clone | present | GREEN |
| `autocrlf=true` fresh clone | **deleted** | **RED** (`base.txt contains a CR byte`) |
| `autocrlf=false` | present | GREEN |
| **`autocrlf=false`** | **deleted** | **GREEN — 22/22, whole suite** |

Row 4 is the problem, because `autocrlf=false` is (a) every runner in `verify.yml` — all ten `runs-on:` are `ubuntu-latest`, so the **required** status check never runs under a CRLF regime — and (b) this repo's own dev boxes, per the local override. Delete line 37 and every merge-gating check stays green. The guard's teeth only exist on `cross-os.yml` (twice-weekly cron) and `release.yml` (tag time).

The PR body's verification — *"confirmed to go RED when `requirements/base.txt` is corrupted to CRLF … proving it actually catches the class of regression it exists to guard"* — measured a hand-corruption, not the regression. I reproduced both: hand-corrupting `cpu.txt` does redden it; deleting the pin it guards does not, on the box that verification was run on. Catalogue shape #3 (*red before the fix for a different reason than the one claimed*) and #7 (*correct code that runs zero times as shipped*).

It is unreachable a second, independent way. `scripts/ci-scope.mjs`, live:

```
--files=".gitattributes"                             → step_test_server = false   (every leg false)
--files="server/tts-sidecar/requirements/base.txt"   → step_test_server = false
--files="openapi.yaml"                               → step_test_server = TRUE
```

`.gitattributes` appears in no step's `globs`/`extraFiles` in `verify-cache.mjs`, and in no `forceRerunTriggers` entry in `server/vitest.config.ts`. `openapi.yaml` is in **both**, with comments citing #1847/#2567 for precisely this runtime-read-with-no-module-graph-edge trap — and `server/vitest.config.ts:180-189` spells out the consequence: *"CI's required check ran and reported green while never re-executing the very guard the diff could have broken."* This PR adds a guard whose only inputs are `.gitattributes` and `server/tts-sidecar/requirements/**`, and wires neither.

**Remedy** (both halves, they fix different things): assert the *pin* rather than its materialisation — read `.gitattributes` (or shell `git check-attr text eol`, which the PR body already used by hand) and assert a rule covers `server/tts-sidecar/requirements/*.txt` with `eol=lf`. That is platform-independent and reddens on deletion everywhere. Keep the CR-byte check as the second assertion; it is the only thing that catches the stale-checkout case below. Then add `.gitattributes` + `server/tts-sidecar/requirements/**` to `test:server`'s `extraFiles`/`globs` and to `forceRerunTriggers`, matching `openapi.yaml`.

---

### 🟠 New — the pin does not reach an existing Windows checkout, so #2586 still reproduces there after this merges (correctness bug)

`.gitattributes:37`

`text eol=lf` governs *checkout*, and git never rewrites a blob it does not think changed. With `text` set, the clean filter collapses a CRLF working file to LF before comparison, so the file compares **equal** to its LF blob — `git status` is clean and the CRLF file survives indefinitely. Measured, on a clone that already existed at `main` under `autocrlf=true`:

```
before pull   base 56  nvidia-cuda 84  cpu 27  amd-rocm 30
git merge <this branch>   → Fast-forward, 3 files changed
after pull    base 56  nvidia-cuda 84  cpu 27  amd-rocm 30     ← unchanged
git status --porcelain    → clean

npx vitest run src/tts/venv-migration.test.ts  → 2 failed | 20 passed
  × requirements/*.txt files never contain CRLF on disk   (base.txt contains a CR byte)
  × onnxruntime-gpu pin bump …    expected 'pip-in-place' to be 'noop'   ← #2586, unfixed
```

So a contributor on a default Windows clone who pulls this fix gets **#2586's original failure, plus a second one**, and the new failure message (`base.txt contains a CR byte (CRLF line ending)`) names no remedy. I measured which remedies work:

```
git checkout -- <file>                     → CR=56  (no-op)
git add --renormalize <dir> && git checkout → CR=56  (no-op)
rm <files> && git checkout -- <dir>        → CR=0   ← the only one that works
```

This also corrects a premise in pass 1's own design-pass framing, which I got wrong: pinning does **not** "self-heal CRLF users with a single reinstall". It heals fresh clones and CI (both real wins — `cross-os.yml`/`release.yml`'s Windows legs are genuinely unblocked). It does nothing for an existing checkout until that checkout is manually renormalised, and there is a reachable production consequence rather than just a red test: `pinokio-scripts/lib/resolve-release.js:109` installs by `git checkout <tag>` **inside a real repo checkout**, so a Windows Pinokio install stamps `.venv-stamp.json`/`.req-hash` from CRLF bytes, while `zip-validate.ts:206-213` derives `ctx.reqHash` from the `ubuntu-latest`-built release zip (LF). `apply.ts:134` then sees a mismatch and runs a full `pipInstall` — the exact symptom, still live for that population after this merges. (The `.req-hash` rewrite at `:137` settles that one path; whether `bootstrap-venv.mjs` re-stamping from the same CRLF checkout can reopen it I did not measure, and am not claiming.)

The pin is still the right call — this is not a request to revisit the design ruling. What is owed is that the fix ships knowingly incomplete: a remediation line in the test's failure message (`rm` + `git checkout --`), and the PR body/register saying so rather than implying `Closes #2586` covers every checkout.

---

### 🟠 New — `speaker-qa.txt` is pip-installed but not hashed: #2534's exact failure mode, in the file set this PR just enumerated (correctness bug, pre-existing)

`server/tts-sidecar/scripts/venv-migration.mjs:116-124` · `server/src/tts/venv-migration.test.ts:178`

`resolveRequired` hashes the overlay and `base.txt` only. All three overlays also `-r speaker-qa.txt` (`nvidia-cuda.txt:84`, `cpu.txt:27`, `amd-rocm.txt:30`), and that file carries real runtime pins — `speechbrain==1.1.0`, `huggingface_hub==0.36.2`. So it is installed but invisible to the hash. Measured:

```
reqHash BEFORE bumping speechbrain==1.1.0 → 1.2.0 : 064ba5e9…a3c47
reqHash AFTER                                     : 064ba5e9…a3c47
hash moved?                                       : false
decideVenvAction(existing venv, new tree)         : 'noop'
```

That is #2534 verbatim — *"the pin was changed WITHOUT changing the hashed text, so reqHash stayed the same and decideVenvAction returned 'noop' for every existing venv, preventing the pin from ever taking effect"* — in the one requirements file the #2534 guard at `:192-272` does not cover. The new CR-byte test's hardcoded four-name list at `:178` excludes it too, though `.gitattributes:37`'s glob does cover it, so the pin itself is fine.

Reporting rather than prescribing, because the fix carries a decision the operator already had to make once on this branch: folding `speaker-qa.txt` into `computeReqHash` moves the hash for **every** existing venv on every platform, i.e. the one-time `pip-in-place` wave. Same shape as pass 1's carve-out. If it is ticketed rather than fixed, the issue should name that decision.

---

### 🟡 Minor

1. **The `RELEASE_NOTES.md` skip cites an inverted precedent.** PR body: *"`RELEASE_NOTES.md` is intentionally untouched, matching the existing pattern for #2291/#2296's own CRLF/symlink CI fixes."* #2296's own follow-up commit `591bf89a` says the opposite, in its message: *"RELEASE_NOTES.md was skipped entirely on the grounds of 'internal tooling, no user-visible delta'. That was contradicted by this branch's own commit message … a user-facing line is added."* — and it adds one (`RELEASE_NOTES.md | 1 +`). #2296 is this repo's recorded precedent for the skip being **caught and reversed**, not for skipping. The conclusion may well still be right — I could not construct a user-visible delta for a zip-installed user, since both sides of that comparison are LF — but a mandatory checklist item was discharged on a claim that inverts its own source. Re-argue it on the merits (and the Pinokio-git population in the 🟠 above is the case to argue against).

2. **`docs/release-notes-next.md:500` mis-attributes two of the four precedents.** It calls this *"the same same-commit-different-hash class already fixed for `openapi.yaml`, `.env.example`, `src/lib/api-types.ts` and the `*.mjs`/`*.sh` globs"*. `.gitattributes:4-13` says in its own words that `.husky/*`, `*.sh`, `*.mjs`, `*.cjs` are pinned because CRLF makes `sh` exit with *"Exec format error"* — a runtime-execution failure, not a content-hash divergence; `openapi.yaml:25-28` is pinned for line-anchored regex matching, also not hashing. Only `.env.example` and `api-types.ts` are the same class. Pass 1's comment introduced this framing and was equally loose — flagging it so the durable register does not inherit it.

3. **`venv-migration.test.ts:178` hardcodes the four filenames.** `.gitattributes:37` globs the directory, so a future fifth overlay is pinned automatically but silently drops out of the assertion. A `readdirSync(join(SIDECAR_DIR, 'requirements'))` filtered to `.txt` is shorter and cannot drift — and would have covered `speaker-qa.txt` today.

---

### Verdict

**Request changes.** All three of pass 1's 🟠 findings and two of its three 🟡 are VERIFIED RESOLVED, on measurement rather than on the receipt: the oracles are byte-identical to `main`, all four mutations redden Oracle 5 at `:271`, and the `.gitattributes` pin is demonstrably in effect against a `main` control in the same CRLF regime. The chosen route is the right one and it does unblock the Windows CI legs.

What is left is that the *guard* over that pin is the weakest part of the change: it cannot fail in the only lane that gates merge (row 4 above), and CI would not even select the leg it lives in on a `.gitattributes`-only diff. Fixing that is one assertion plus two wiring lines, and it converts the pin from convention into something enforced. The stale-checkout gap and `speaker-qa.txt` are the two things this round measured that the PR body currently claims or implies otherwise about.

<sub>Automated adversarial pass per `.claude/skills/pr-review-gate`. Findings only — no files were modified by this review; all measurement ran in throwaway clones under `%TEMP%`.</sub>

dudarenok-maker:
## PR review — pass 3 (head 9ccc7274, depth medium)

Scope: the full 9-file PR diff vs `main`, with the new commit's non-test changes (`venv-migration.mjs`, `verify-cache.mjs`, `vitest.config.ts`) probed as production code rather than as wiring · verified in a throwaway `%TEMP%` clone under `core.autocrlf=true` (positive control: `package.json` CR=156, `requirements/*.txt` CR=0 — the pin is what excludes them) plus a second clone created at `main` **before** the pin and then merged forward, to reproduce the stale-checkout case. Baseline 23/23, 60/60, `test:hooks` exit 0.

---

### Pass-2 dispositions

**🟠 1 — the guard cannot fail in the merge-gating lane, and CI does not select it: VERIFIED RESOLVED, both halves, measured.**

*Rule assertion.* Deleted `.gitattributes:53` in the CRLF clone but left the files LF on disk (i.e. simulating `ubuntu-latest`, the row-4 case that was green before):

```
× .gitattributes pins requirements/*.txt to LF, so reqHash is checkout-independent (#2586)
AssertionError: .gitattributes must pin server/tts-sidecar/requirements/*.txt to text eol=lf
Tests  1 failed | 22 passed
```

Red on an LF tree, red for the reason claimed. Row 4 is closed.

I also tried to break the pair open — a later `*.txt text eol=crlf` line leaves the pinned text present, so assertion 1 stays green — and the second assertion catches it (`amd-rocm.txt contains a CR byte`, `2 failed | 21 passed`). The two assertions genuinely cover each other's blind spot. Not a finding.

*ci-scope.* Run directly, branch vs `main` control:

| `--files=` | `main` | this branch |
|---|---|---|
| `.gitattributes` | `step_test_server: false` | **`true`** |
| `server/tts-sidecar/requirements/speaker-qa.txt` | `step_test_server: false` | **`true`** |

`force-rerun-triggers.test.ts` matches through the real `picomatch` against real absolute paths at three checkout shapes, so the two new entries are a real oracle, not a restated list. 60/60.

**🟠 3 — `speaker-qa.txt` not hashed: RESOLVED IN `resolveRequired`, NOT ON THE PATH THAT SHIPS.** The half that landed is correct and I reproduced #2534 through it end-to-end (see ✅ below). The half that did not is the new 🔴.

**🟠 2 — the pin does not reach an existing CRLF checkout: PARTIALLY RESOLVED.** The limitation is real (I reproduced it: merge the branch into a pre-pin `autocrlf=true` clone → `base 56 / nvidia-cuda 84 / speaker-qa 7` CR bytes, unchanged, `git status` clean), it is documented, and #2596 is filed with an accurate body that names the decision owed. But **the documented remediation does not work** — see 🟠 New 2 — and the test's failure message still names no remedy (`venv-migration.test.ts:203` is still bare `${file} contains a CR byte (CRLF line ending)`), which was the other half of the ask.

**🟡 1 — inverted `#2296` precedent: PARTIALLY RESOLVED.** `RELEASE_NOTES.md` now carries an entry, so the checklist item is discharged in form. The *substance* of the re-argument does not hold — see 🔴 — and the PR body still carries the original inverted claim verbatim.

**🟡 2 — mis-attributed precedent list: VERIFIED RESOLVED** in `docs/release-notes-next.md:500`. The corrected framing did not propagate to the PR body, which still lists all four.

**🟡 3 — hardcoded 4-file list: VERIFIED RESOLVED.** `readdirSync` + `expect(files.length).toBeGreaterThan(0)`, so the enumeration cannot go vacuous. `speaker-qa.txt` is now covered.

---

### 🔴 Blocking — `speaker-qa.txt` joined `resolveRequired`'s hash but not `zip-validate`'s, so #2534's failure mode is still open on the self-upgrade path — and the reinstall wave both release-notes files promise never fires there (correctness bug)

`server/src/upgrade/zip-validate.ts:138-143, 209-212` · `server/src/upgrade/apply.ts:106-111, 134` · `server/src/routes/upgrade.ts:102` · `RELEASE_NOTES.md:3` · `docs/release-notes-next.md` (speaker-qa bullet)

There are **two** producers of a requirements hash, not one. This commit changed one:

- `resolveRequired` (`venv-migration.mjs:120-128`) → `computeReqHash([overlay, base, speakerQa])` — **changed**. Consumed by `bootstrap-venv.mjs` and by `apply.ts:106`.
- `validateUpgradeZip` (`zip-validate.ts:209-212`) → `computeReqHash([read.reqOverlayText, read.reqBaseText])` — **unchanged**. `readUpgradeZip`'s `want()` at `:138-143` only ever matches `nvidia-cuda.txt` and `base.txt`, so `speaker-qa.txt` never reaches it.

`ctx.reqHash` from the second one is what actually gates the pip install on the shipped upgrade path (`apply.ts:134`) and what the UI is told (`upgrade.ts:102`'s `requiresPipInstall`). The 3-file hash reaches `apply.ts` only through `classifyVenvState` at `:107`, whose result is inspected **solely** for `'needs-reinstall'` (`:112`) — and a `reqHash`-only mismatch maps to `'pip-in-place'` (`venv-migration.mjs:84-85`), which that branch drops on the floor.

Measured, against the real modules:

```
speaker-qa.txt pins: speechbrain==1.1.0 | huggingface_hub==0.36.2
bump speechbrain==1.1.0 -> 1.2.0, nothing else changed

post-fix resolveRequired : 118368d2… -> ddeeaef6…   decideVenvAction = pip-in-place   ← the fix works here
pre-fix  resolveRequired : 064ba5e9… -> 064ba5e9…   decideVenvAction = noop           ← #2534 reproduced
zip-validate ctx.reqHash : unchanged by a speaker-qa-only bump = true                 ← UNFIXED
```

So for a release whose only requirements change is in `speaker-qa.txt` — the exact shape the new oracle exists to catch — a zip / in-app upgrade computes an unchanged `ctx.reqHash`, reports `requiresPipInstall: false`, skips `steps.pipInstall`, and the new pin is never installed. That is #2534 verbatim, on the release artifact's own upgrade path, after a PR whose commit message says the gap is closed.

**This also makes the user-facing note false.** `RELEASE_NOTES.md:3` promises *"most people's next launch after updating will trigger a one-time reinstall of a small piece of the voice engine"*, and `docs/release-notes-next.md` repeats it. Nothing triggers it for a zip install. I traced every caller of `resolveRequired` / `bootstrap-venv.mjs`:

| entry point | runs `bootstrap-venv.mjs`? |
|---|---|
| `launch.mjs` (versioned-dir launcher) | no |
| `scripts/start-app.mjs` / `start-app-prod.mjs` | no |
| server startup / sidecar supervisor | no (`grep` finds no `resolveRequired` under `server/src` outside `upgrade/`) |
| `apply.ts` (in-app self-upgrade) | calls `resolveRequired`, but pip is gated on `ctx.reqHash` — see above |
| Setup checker (`setup-diagnosis.ts:54-80`) | only on `venv-missing` / `venv-broken`; never on a hash mismatch |
| `pinokio-scripts/install.js:80`, `update.js:73` | **yes** |
| `pinokio-scripts/start.js` | no |

So the wave fires for exactly one population — Pinokio — and it fires **during the Update action**, not at "next launch". For everyone on the zip, neither the reinstall nor the `speaker-qa` fix arrives at all. `docs/features/220-engine-retier-health-honesty.md:93` states the same premise (*"On the next `npm run start:prod`, the bootstrap detects the hash change and re-runs pip install"*) and appears to have been stale on `main` already; this PR is the first thing to build a shipped user promise on top of it.

Catalogue #10 (*one instance fixed, the class left armed* — the sibling producer), #7 (*a fix unreachable at the shipped configuration*), and #5 (the new oracle passes on a null observation as far as the release path is concerned). Nothing tests the two producers against each other — `zip-validate.test.ts` asserts internal consistency against fixture constants only — which is why a divergence this size ships green (I ran `src/upgrade/`, `routes/upgrade.test.ts`, `bootstrap-venv-helpers.test.ts`: 54/54).

Note the pre-existing wrinkle this sits on top of: `readUpgradeZip` hardcodes the **nvidia** overlay, so `ctx.reqHash` and the stamp's `reqHash` were already divergent on a `cpu`/`amd` box. What this commit changes is that they now diverge on **every** profile, which is what makes `zip-validate.ts:181-182`'s and `:206-207`'s *"byte-for-byte the same hash `resolveRequired` writes into the venv stamp"* newly and unconditionally false.

---

### 🟠 Significant — the remediation documented in `.gitattributes` does not work, and following it stages five file deletions (correctness bug)

`.gitattributes:47-48`

The comment tells a user on a stale CRLF checkout to run:

```
git rm --cached -r -- server/tts-sidecar/requirements
git checkout -- server/tts-sidecar/requirements
```

Run verbatim in the pre-pin clone that reproduces the problem:

```
$ git rm --cached -r -- server/tts-sidecar/requirements
rm 'server/tts-sidecar/requirements/cpu.txt'  … (5 files)

$ git checkout -- server/tts-sidecar/requirements
error: pathspec 'server/tts-sidecar/requirements' did not match any file(s) known to git

CR after documented recipe: base.txt 56   nvidia-cuda.txt 84   speaker-qa.txt 7   ← unchanged

$ git status --porcelain
D  server/tts-sidecar/requirements/amd-rocm.txt
D  server/tts-sidecar/requirements/base.txt
D  server/tts-sidecar/requirements/cpu.txt
D  server/tts-sidecar/requirements/nvidia-cuda.txt
D  server/tts-sidecar/requirements/speaker-qa.txt
?? server/tts-sidecar/requirements/
```

`git rm --cached` drops the paths from the index, so the second command has nothing to restore from and errors. The CRLF is still there, and the user is left with the entire requirements directory **staged for deletion** — strictly worse than doing nothing, on a Pinokio-style checkout where a later commit or tooling step could act on it.

Pass 2 measured which remedies work and named exactly one: `rm <files> && git checkout -- <dir>`. I re-confirmed it — after resetting the index and deleting the files, `git checkout -- server/tts-sidecar/requirements` gives `CR 0 / 0 / 0`. That recipe *is* present in the comment, but only as a parenthetical whole-tree aside at `:49`; the primary, scoped one is the broken one. #2596's body carries the same broken command as its proposed fix (`git rm --cached -r … && git checkout -- …`), so it needs the same correction.

Catalogue #6 (*success reported while doing nothing*) and #9 (*the defect is in the document, not the code*).

---

### 🟠 Significant — the PR body describes a different PR than the one that will merge (house gate)

`gh pr view 2588 --json body`, unchanged since `38b4d019`

The body is the durable record and the merge-commit context, and CLAUDE.md requires incidental fixes to be *declared in the PR body*. As it stands it:

- **states the opposite of what shipped** — *"`RELEASE_NOTES.md` is intentionally untouched, matching the existing pattern for #2291/#2296"*, while the head commit adds `RELEASE_NOTES.md | 2 +` precisely because pass 2 found that precedent inverted;
- **says "Not user-visible: this is a CI/tooling reliability fix"**, while the PR now deliberately ships a user-visible one-time reinstall;
- **lists only 2 changed files** (`.gitattributes`, `venv-migration.test.ts`) where the PR changes 9, including production code (`venv-migration.mjs`) it does not mention at all;
- **never mentions `speaker-qa.txt`, the reinstall-wave decision, or #2596** — the whole substance of this round, and the one thing a reviewer of the merge commit would most need;
- **still carries the four-precedent list** (`openapi.yaml`, `.env.example`, `api-types.ts`, `*.mjs`/`*.sh`) that `docs/release-notes-next.md:500` was corrected this round for getting wrong;
- **reports `22/22`** where the suite is now 23.

Not a code-correctness bug — flagged as a house gate, not as a re-review trigger.

---

### 🟠 Significant — on-box acceptance not recorded, for behaviour only a real install can show (house gate)

No `docs/testing/**` change in the diff. This PR now ships a behaviour whose only proof is a real venv on a real box: an existing `.venv-stamp.json` re-hashing and driving one `pip-in-place` through `bootstrap-venv.mjs`. The register already carries rows of exactly this shape and prerequisite — A39 (*ORT marker — fresh NVIDIA bootstrap*, **no GPU needed, sidecar venv only**), A27, A29. Checklist step 3 makes *recording* the gate, not *running*, so this is a row-plus-live-view edit rather than a blocker on the box.

Whether it rises to a row is a judgment call and yours to make; I am flagging it because the diff shows it was not considered either way.

---

### 🟡 Minor

1. **Stale comments the change made false, all fixable in this round.** `zip-validate.ts:181-182` and `:206-207` (*"byte-for-byte the same hash `resolveRequired` writes into the venv stamp"*, *"in `resolveRequired`'s order (overlay THEN base)"*) — now false on every profile; `venv-migration.mjs:14` (`computeReqHash`'s docstring still reads *"overlay then base"*) — three files now; `docs/features/220-engine-retier-health-honesty.md:93` (the `start:prod` self-heal premise the release note leans on).
2. **`venv-migration.test.ts:203`'s failure message names no remedy.** Pass 2 asked for one; the source comment at `:224-228` points at `.gitattributes`, but the assertion text is what a dev actually sees in CI output.
3. **#2596 is missing its `area:` label.** It carries `type:chore` alone; every neighbouring chore issue (#2597, #2583, #2551) carries `area:ops` + `type:chore` per CONTRIBUTING's two-shape convention. Body and scoping are otherwise accurate and the owed decision is named.

---

### ✅ What is solid (new this round only)

- **The `speaker-qa.txt` oracle is a real oracle.** Mutating `resolveRequired` back to `computeReqHash([overlay, base])` reddens 5 tests including the new one, and it fails on `expect(required.reqHash).toBe(currentHash)` — production-pinning, not on the two test-side assertions that cannot fail. Red for the reason claimed.
- **The one-time wave classifies correctly.** An existing stamp carrying the old 2-file hash against the current tree → `pip-in-place`, not `rebuild` (`pythonTag`/`profile` unchanged), so the bootstrap path takes the additive branch rather than telling a user to reinstall from scratch. `null` stamp still → `rebuild`, unchanged.
- The `.gitattributes` known-limitation comment is accurate on the *mechanism* (verified by reproduction) — it is only the remediation commands that are wrong.
- `speaker-qa.txt` is inside the release zip (`build-release-zip.mjs`'s `server/tts-sidecar/**` include, with only `.venv`/`voices/kokoro`/`tests` excluded), so `resolveRequired`'s new unguarded `readFileSync` cannot ENOENT on the `apply.ts` path; it is also present in v1.12.0/v1.13.0/v1.14.0, so a permitted downgrade cannot hit it either.

---

### Verdict

**Not mergeable as-is.** Pass 2's finding (a) is fully and verifiably resolved — that half of this round is good work, and the rule/materialisation assertion pair is genuinely sound under adversarial probing. But (b) shipped with a remediation that fails and leaves a staged deletion, and (c) was fixed on one of the two hash producers — leaving #2534's failure mode live on the path the release artifact actually upgrades through, and leaving both release-notes files promising a reinstall that will not happen for that population.

The smallest coherent close is one line plus one entry in `zip-validate.ts` (`want()` matching `speaker-qa.txt`, `computeReqHash([overlay, base, speakerQa])`), a test that pins `validateUpgradeZip`'s hash to `resolveRequired`'s over the same tree so the two producers can never silently diverge again, the two corrected git commands, and a release-notes sentence that says which install path sees the reinstall and when.

**This is re-review 2 of 2 — the loop cap in `.claude/skills/pr-review-gate` is now reached.** These findings are reported, not waived; there is no automatic pass 4, so whether to fix and merge, fix and re-review by hand, or ship with the gaps recorded is the operator's call.

<sub>Automated adversarial pass per `.claude/skills/pr-review-gate`. Findings only — no files were modified by this review; the worktree and primary checkout are byte-identical to their pre-review state, and all measurement ran in throwaway clones under `%TEMP%`.</sub>

dudarenok-maker:
## PR review — pass 4 (head 89cde4a2, depth medium)

Operator-authorised round beyond the skill's automatic 3-round cap. Scope: commit `89cde4a2` (7 files) read as production code, plus the cumulative 13-file diff vs `main` for context. Everything below was re-derived from scratch — including pass 3's own characterisations, which I reproduced rather than inherited.

Instrument: a throwaway `%TEMP%` clone under `core.autocrlf=true` that *happened to reproduce the stale-checkout case naturally* (initial checkout landed on `main`, before the pin; the branch checkout then left `requirements/*.txt` at `base 56 / nvidia-cuda 84 / cpu 27 / amd-rocm 30 / speaker-qa 7` CR bytes with `git status` clean — the exact shape #2596 describes). Positive control throughout: `package.json` CR=156 in the same clone. A second, independent clone created *at* `89cde4a2` under the same `autocrlf=true` materialised all five files at CR=0 — so the pin is what excludes them, not the regime. Baseline 40/40, typecheck exit 0, `check:onbox-register` exit 0.

---

### Pass-3 dispositions

**🔴 1 — `zip-validate`'s hash producer left out of lockstep: VERIFIED RESOLVED.** Not on the receipt — on four mutations, each applied and reverted one at a time, against a 40/40 unmutated baseline:

| mutation | outcome |
|---|---|
| **A** — `validateUpgradeZip` back to `computeReqHash([overlay, base])`, i.e. pass 3's actual bug state | **RED** — parity test + 3 others; `expected '8387516835…' to be '006ad5e584…'` |
| **B** — `want()` drops the `speakerQa` line, hash call keeps 3 files | **RED** — parity test + 4 others (`expected null to be '006ad5e584…'`) |
| **C** — order drift only, `[overlay, speakerQa, base]` in `zip-validate` alone | **RED** — parity test + 1. So it pins *order*, not just membership |
| **D** — a **future** 4th requirements file joins `resolveRequired` only | **RED — the parity test ALONE**, 1 failed / 16 passed |

D is the one that matters for the forward-looking claim in the PR body, and it holds: no other test in the suite notices, and the parity test does.

The production change is correct on the axes pass 3 asked about. `want()`'s suffix match resolves against the **real** entry name — I ran `releaseInternalPrefix('1.15.0')` and `matchesManifest()` out of `scripts/build-release-zip.mjs` directly: `castwright-1.15.0/server/tts-sidecar/requirements/speaker-qa.txt` matches, and the file is inside the manifest (`server/tts-sidecar/**`, no exclusion — `tests/**` returns `false` as the negative control). So the widened `reqHash: null` fallback cannot fire for a genuine release zip. The `if/else if/else` at `:162-165` is exhaustive over `want()`'s four non-null returns.

**🟠 2 — the documented CRLF remediation was broken: VERIFIED RESOLVED.** I reproduced **both** halves rather than taking pass 3's word:

```
OLD recipe, verbatim, on the stale clone:
  git rm --cached -r -- server/tts-sidecar/requirements   → rm'd 5 paths, exit 0
  git checkout -- server/tts-sidecar/requirements         → error: pathspec ... did not match
                                                             any file(s) known to git   (exit 1)
  CR after: base 56  nvidia-cuda 84  speaker-qa 7          ← unchanged
  git status: D ×5 + ?? server/tts-sidecar/requirements/   ← 5 staged deletions, as documented

NEW recipe, verbatim:
  rm server/tts-sidecar/requirements/*.txt
  git checkout -- server/tts-sidecar/requirements          → exit 0
  CR after: 0 / 0 / 0 / 0 / 0    git status: clean
  package.json CR still 156                                ← clone regime still CRLF
```

Also works in PowerShell (`Remove-Item 'server/tts-sidecar/requirements/*.txt'` + the same `git checkout`), which matters because the audience for this comment is a Windows user. #2596's body carries the corrected recipe too.

**🟠 3 — the PR body described a different PR: RESOLVED, and accurate on measurement.** I reproduced every verification claim it makes rather than checking it reads plausibly:

- `40/40 green` on the two suites — confirmed (23 + 17).
- `npm run typecheck` — exit 0.
- `npm run check:onbox-register` — exit 0.
- `ci-scope.mjs` flips `step_test_server` → `true` for `.gitattributes` and for `requirements/speaker-qa.txt`, `false` for a `docs/BACKLOG.md` negative control.
- The M3 mutation (`resolveRequired` returns a hardcoded constant) reddens 6 tests including the ORT re-stamp oracle — the pass-1 tautology is genuinely gone.
- Deleting `.gitattributes:53` on an **LF** tree (the `ubuntu-latest` case) reddens with `.gitattributes must pin server/tts-sidecar/requirements/*.txt to text eol=lf`.
- Hand-corrupting `cpu.txt` to CRLF reddens with `cpu.txt contains a CR byte`.

One imprecision remains, in 🟡 New 3 below.

**🟠 4 — on-box acceptance not recorded: RESOLVED IN FORM.** A48 exists, in Group A, tagged **no GPU needed, sidecar venv only**, with the register's required shape (explanatory paragraph on what unit tests prove vs. what only a real venv can, bulleted criteria, `*Needs:* / *Criteria:* / *Cost:*` footer) — consistent with A39/A27/A29. Arithmetic moves consistently: Group A 47→48, total 74→75, glance table / group header / summary strip all agree in both the `.md` and the tracked `.html`, and the live artifact at `adf22b7b-…` is already published carrying Owed **75**, Group A **48 rows**, and the dated addition callout. `Blocked 2 / Unconfirmed 2 / Oldest 06-01` correctly unchanged. **The row's *content* is where the new 🟠 is.**

**🟡 1 — stale comments the change made false: PARTIALLY RESOLVED.** `zip-validate.ts:181-182` and `:206-207` were rewritten this round ✅. Two of the four named are untouched: `venv-migration.mjs:14` — `computeReqHash`'s own docstring still reads *"the resolved requirements files (overlay then base)"* while every caller now passes three; and `docs/features/220-engine-retier-health-honesty.md:93` still asserts *"On the next `npm run start:prod`, the bootstrap detects the hash change and re-runs `pip install`"*, which this PR's own `docs/release-notes-next.md` entry now explicitly contradicts in the same round (`a plain npm run start:prod / launch.mjs boot does NOT re-run bootstrap-venv.mjs`). `git diff origin/main...89cde4a2 --stat -- docs/features/220-…` is empty.

**🟡 2 — the CR-byte assertion's failure message names no remedy: NOT RESOLVED.** `venv-migration.test.ts:203` is still bare `${file} contains a CR byte (CRLF line ending)`. Asked for in pass 2, restated in pass 3, unchanged in pass 4. The source comment at `:224-228` points at `.gitattributes`; the assertion text is what a dev reads in CI output.

**🟡 3 — #2596's missing `area:` label: VERIFIED RESOLVED** — now `area:ops` + `type:chore`.

---

### 🟠 New — A48's criteria pass on a null observation on one of the two paths it offers, and its last criterion is unsatisfiable on either (correctness bug, in the acceptance instrument rather than in shipped code)

`docs/testing/onbox-acceptance-register.md:2676-2686` · mirrored in `onbox-acceptance-register-live-view.html`'s A48 block · `server/src/upgrade/apply.ts:134`, `:305-308`

The row says *"Trigger **either** real reinstall path"* and offers Pinokio Update or the in-app zip upgrade as equivalents. They are not equivalent, and I measured why.

**(a) On the in-app zip path, criterion 3 would pass with the whole `speaker-qa` change reverted.** The pip gate is `apply.ts:134`:

```ts
if (ctx.reqHash && ctx.reqHash !== steps.readReqHash()) { … pipInstall … }
```

`.req-hash` has exactly one writer in the repo — `apply.ts:305-308`. `bootstrap-venv.mjs` never writes it (`grep -l req-hash` over `*.ts *.mjs *.js` returns `apply.ts` and `apply.test.ts`, nothing else). So on any box that has never applied an in-app zip upgrade — every Pinokio box, every source install, every first-ever in-app upgrade — `readReqHash()` returns `null`, the condition is `ctx.reqHash !== null` → **true**, and pip runs for *any* release whether or not `speaker-qa.txt` is in the hash. The operator observes "pip-in-place happened" and records PASS on an observation that does not discriminate.

Criterion 4's first half is blind the same way: `speechbrain`/`huggingface_hub` land in the venv because every overlay `-r speaker-qa.txt`, so they are present after *any* overlay install, fix or no fix.

The Pinokio branch **does** discriminate — an old 2-file stamp vs. the new 3-file `required.reqHash` gives `pip-in-place` with the fix and `noop` without it. So one of the two offered paths has teeth and the other has none, and the row presents them as interchangeable.

**(b) Criterion 4's second half cannot be satisfied on either path.** It asks to confirm *"`.venv-stamp.json` / `.req-hash` now record the new 3-file hash"*. `writeStamp` has **no caller under `server/src`** (only `bootstrap-venv.mjs` and the test file), and `.req-hash` has no writer outside `apply.ts`. So:

- Pinokio Update → `.venv-stamp.json` updated, `.req-hash` untouched (typically absent entirely);
- in-app zip upgrade → `.req-hash` updated, `.venv-stamp.json` still carrying the **old 2-file hash**, permanently.

Whichever path the operator takes, half of that criterion is false by construction. An acceptance row that can only be discharged on a false or a null observation is the exact failure the register exists to prevent (catalogue #5). One-paragraph fix: name the Pinokio path as the discriminating one, say what the *other* path proves (that `ctx.reqHash` moved — observable via `POST /upgrade/stage`'s `requiresPipInstall`, which is the thing this round actually changed), and split criterion 4 by path.

---

### 🟡 Minor

1. **The rewritten `validateUpgradeZip` docstring re-asserts an absolute that is false on two of the four profiles, and the new describe-block name claims more than the test covers.** `zip-validate.ts:196-197` and `:229-230` say the hash is *"byte-for-byte the same hash `resolveRequired` writes into the venv stamp"*; `zip-validate.test.ts:238` names the block *"the two producers cannot silently diverge"*. `readUpgradeZip` hardcodes the **nvidia** overlay while `bootstrap-venv.mjs:313` stamps the **effective** profile. Measured against the real requirements files:

   ```
   profile=nvidia    118368d28f6120e6…   === zip ctx.reqHash?  true
   profile=cpu       de8795ab3064b74a…   === zip ctx.reqHash?  false
   profile=amd       9aec5e12d6051e1d…   === zip ctx.reqHash?  false
   profile=apple     118368d28f6120e6…   === zip ctx.reqHash?  true
   ```

   The producers *do* silently diverge on `cpu` and `amd`, and the parity test is blind to it because it only ever calls `resolveRequired(sidecarDir, 'nvidia')`. This is pre-existing (the nvidia hardcode long predates this branch) and **not** a regression — the speaker-qa fix itself reaches all three profiles correctly, since all three overlays `-r` it. What is owed under the repo's own chore rule is that a comment *rewritten in this round* states it without the qualifier. Cheapest close: qualify both sentences with "on the nvidia/apple profile", and rename the describe block to what it proves (the two producers hash the same **file set in the same order**).

2. **PR body: "Oracle 3 and Oracle 5 restored to their original `main` form" is no longer true.** They were restored byte-identically at `ddcf3c2a` (pass 2 verified that), but `9ccc7274` then re-modified the block to hash the third file — `venv-migration.test.ts:311-312` now computes `oldHash`/`currentHash` over `[…, normalizedBase, normalizedSpeakerQa]`. The oracles are restored *in shape*, not to `main`'s text. The body is the durable merge-commit record, so worth one word.

3. **PR body's file list is 11 of 13.** `scripts/tests/verify-cache.test.mjs` and `server/src/force-rerun-triggers.test.ts` (the paired tests that landed with the CI-wiring change) are not named. Minor, but pass 3's finding was specifically about the body under-listing the diff.

---

### Checked adversarially and cleared — no finding

Recording these so the next round does not re-derive them:

- **The widened `reqHash: null` fallback is not reachable on any shipped path.** `speaker-qa.txt` was added 2026-06-21 (`9a8baa9b`) — ABSENT from v1.7.0/v1.8.0/v1.9.0, PRESENT from v1.10.0 onward. A zip lacking it would now yield `reqHash: null` → `requiresPipInstall: false` → pip skipped, where before it yielded a real 2-file hash. But `validateUpgradeZip` has exactly one production caller (`routes/upgrade.ts:94`), which passes **no** `opts`, so `allowDowngrade` is never true and any such zip is refused with 412 `code: 'downgrade'` before `apply` ever runs. The only residual is a v1.8.0→v1.9.0 in-app upgrade, which I am not claiming is reachable.
- **`resolveRequired`'s new unguarded `readFileSync` of `speaker-qa.txt` cannot ENOENT on `apply.ts:106`.** Same reachability argument — the extracted candidate is always ≥ the running version, and the file ships in the manifest.
- **A cpu/amd box does not get a false `needs-reinstall` from `apply.ts`.** I chased this and it does not happen: `createApplySteps` wires `resolveRequired: (sidecarDir) => resolveRequired(sidecarDir, effectiveProfile())` at `:313`, so `required.profile` matches the stamp's and `decideVenvAction`'s profile branch never trips.
- **The pin-rule regex at `venv-migration.test.ts:186` enumerates syntax, but fails CLOSED.** A differently-spelled-but-correct pin reddens it (safe direction); and a later neutralising `*.txt text eol=crlf` line cannot slip past the pair, because `eol=crlf` forces CRLF on checkout on *every* platform, so assertion 2 catches it on `ubuntu-latest` too — not only on a CRLF box.
- **The release-notes population claim is accurate.** `bootstrap-venv.mjs` has exactly two invokers repo-wide (`pinokio-scripts/install.js:80`, `update.js:73`) plus the user-initiated `POST /api/setup/venv/bootstrap`; `launch.mjs`, `scripts/start-app*.mjs` and server startup do not run it. All three overlays `-r speaker-qa.txt` (`nvidia-cuda.txt:84`, `cpu.txt:27`, `amd-rocm.txt:30`).
- **#2596 is a legitimate design-pass deferral — but on one of its three stated reasons.** "Generic (re-normalize anything `.gitattributes` pins to `eol=lf`) vs. hardcoded to today's `requirements/` path" is a genuine fork with two defensible outcomes, and CLAUDE.md's carve-out asks exactly that the ticket name the decision. The other two reasons are close to ones CLAUDE.md explicitly voids — *"the CLI half has no existing safety net"* is "it needs its own test", and blast radius is a model-routing tier, not a deferral licence. The deferral holds; it just holds narrowly.
- All 16 PR checks green, including `Check on-box acceptance register consistency` and `Verify PR body links a GitHub issue`. `Closes #2586` is line 1 of the body, outside any code span.

---

### Verdict

**Mergeable.** The 🔴 from pass 3 is closed, and closed properly — the fix is proved by four independent mutations rather than by inspection, the `want()` matcher is verified against the real release-zip entry name and manifest, and the parity test is the only thing in the suite that catches a future one-sided requirements-file addition. Both of pass 3's 🟠 house-gate findings are discharged, and I reproduced the broken-vs-working remediation from scratch rather than inheriting the claim.

What I would fix before pushing merge, all of it cheap and none of it shipped-code behaviour:

1. **A48's criteria** (🟠 above) — a one-paragraph edit. The register is a merge-gate artifact, and a row whose criteria pass on a null observation converts a real debt into a false discharge later. Note this also means the tracked `.html` and the live artifact need the same edit + a republish through the register's own four-step procedure.
2. The four documentation chores the repo's own "chores get fixed in the same round" rule already makes owed: `venv-migration.mjs:14`, `docs/features/220-…:93`, `venv-migration.test.ts:203`'s remedy-less message, and the two `zip-validate` over-claims in 🟡 1.

None of those change what the code does. If the operator ships as-is, the shipped behaviour is correct on every path I could reach; what carries forward is a register row that cannot honestly be discharged and four documents that say something the code no longer does.

<sub>Automated adversarial pass per `.claude/skills/pr-review-gate`. Findings only — no files were modified by this review; the primary checkout and `wt-2586-crlf-venv-test` are byte-identical to their pre-review state (`git status --porcelain` empty, HEADs unmoved), and every mutation ran in throwaway `%TEMP%` clones that were reverted and re-verified clean after each one.</sub>

PR #2622, opened 2026-08-22
chore(sidecar): add per-device memory_stats diagnostics + bare-reclaim endpoint

### LINKED ISSUE #2423: side — /debug/memory cannot answer #1996: add per-device memory_stats and a bare-reclaim probe
## What

Two additive diagnostics on the TTS sidecar, so #1996's measurement can actually be taken. **#1996 is blocked on this** — the reading that decides its design cannot be produced by any surface that exists today.

1. **Extend `GET /debug/memory`** with a per-device `memory_stats` block: `reserved`, `allocated`, `inactive_split` and `num_alloc_retries`, from `torch.cuda.memory_stats(i)` for every visible device.
2. **Add `POST /debug/reclaim`** — performs one bare `_reclaim_device_cache()` and returns the per-device figures **before and after**, in a single response.

## Why this is needed

#1996 asks whether the VRAM pool a finished render strands is *uncollected cache* (a scheduled reclaim fixes it) or *allocator fragmentation* (`empty_cache()` can never return it, and the fix is a boundary recycle instead). Those two answers lead to designs with nothing in common, and two attempts have already been built on the unmeasured assumption.

Neither existing surface can discriminate them:

- `GET /debug/memory` (`server/tts-sidecar/main.py:9477`) reports `allocated_mb`/`reserved_mb` for the **current device only** (`:9523-9532`) and carries no `inactive_split_bytes` — which is the field that separates the two mechanisms.
- `_cuda_vram_mb_per_device` (`:8015`) is per-device but reports only `reserved_mb`/`total_mb`.
- **Nothing anywhere exposes a bare, on-demand reclaim.** The before/after pair must bracket an `empty_cache()` with nothing in between, and it must be a *bare* reclaim — a load/unload cycle conflates it with an allocation pass, which is precisely the unsound inference that sent both prior attempts down the wrong path.

## Where the work goes

`server/tts-sidecar/main.py`, in the existing `/debug` route family — `debug_memory` at `:9477`, siblings `/debug/codec-timing` at `:9562` and `:9567`. The reclaim primitive is `_reclaim_device_cache` at `:3767` (module-level, already best-effort and exception-swallowing).

## Acceptance

- [ ] `GET /debug/memory` returns a per-device block keyed by device, each carrying `reserved`, `allocated`, `inactive_split`, `num_alloc_retries`. Existing fields keep their current meaning and shape — this is purely additive, and `vram_reserved_mb`'s current-device-only semantics are deliberately unchanged (a live Node consumer reads it).
- [ ] `POST /debug/reclaim` returns `{before: {...}, after: {...}}` with the same per-device shape, having run exactly one `_reclaim_device_cache()` between the two snapshots.
- [ ] Both fail soft when CUDA is unavailable — `{}` / omitted keys, never a 500. Matches `_cuda_vram_mb_per_device`'s existing fail-open contract.
- [ ] **Device keys are derived the same way `probe_capacity` derives them** (`:3491`, `kind = "rocm" if _cuda_is_rocm() else "cuda"`), NOT hardcoded `f"cuda:{i}"`. `_cuda_vram_mb_per_device` hardcodes the `cuda:` prefix at `:8038`; copying that is how a reading on an AMD box silently mislabels its devices. Note `probe_capacity` also emits bare `cpu` and `mps:0` keys (`:3506-3510`) — exclude non-GPU keys rather than emitting empty stat blocks for them.
- [ ] pytest coverage alongside the existing sidecar suite (`server/tts-sidecar/tests/`; `test_memory.py` already covers the `/health` VRAM fields and carries the #1976 cases at `:909`). At minimum: the per-device shape, the CUDA-unavailable path, and that `POST /debug/reclaim` calls the reclaim exactly once between its two snapshots.

## Not in scope

- Any change to placement, admission, or eviction behaviour. These are read-only diagnostics; `POST /debug/reclaim` performs a reclaim, but one that the memory watchdog already performs unprompted every 60 s (`:7915`, `:8613-8614`).
- Any Node-side change. The measurement reads the sidecar directly on `:9000`; adding a third key to the per-device map would stale `server/src/routes/sidecar-health.ts:186`'s closed element type, so **do not** route these through `/health`.
- The #1996 lever itself. This issue exists so the lever can be *chosen* rather than guessed.

## Notes for whoever picks this up

`POST /debug/reclaim` is not inert — it frees real memory. It is guarded by being a `/debug` route on a loopback-bound sidecar, and its effect is identical to a watchdog tick that already fires on a timer.

Sidecar changes put `test:sidecar` in scope for the pre-commit gate; a commit touching `server/**` can take several minutes.

Full design context: `docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md` §4 (in PR #2415). The run sheet that consumes these two endpoints is `docs/testing/1996-stranded-vram-measurement.md` in the same PR.

Blocks #1996.

dudarenok-maker:
## Summary
- GET /debug/memory gains a per-device memory_stats block (
eserved, llocated, inactive_split, 
um_alloc_retries from 	orch.cuda.memory_stats(i), raw units), keyed cuda:N/
ocm:N the same way probe_capacity() derives its device kind — not hardcoded cuda:.
- New POST /debug/reclaim brackets exactly one bare _reclaim_device_cache() call between a before/after snapshot pair, so #1996's design question (uncollected cache vs. allocator fragmentation) can finally be measured.
- Both fail soft ({} / omitted keys) when CUDA is unavailable. Existing /debug/memory fields are untouched.
- Read-only diagnostics; no placement/admission/eviction behaviour changes, no Node-side changes. POST /debug/reclaim does free real memory, and unlike the memory watchdog (which only reclaims when RSS crosses its threshold), this route calls _reclaim_device_cache unconditionally — allowing the same gc.collect() + mpty_cache() operation to be triggered regardless of RSS, even below the watchdog's threshold.

## Test plan
- [x] 6 new pytest cases in server/tts-sidecar/tests/test_memory.py: memory_stats shape/values, CUDA-unavailable fail-soft, ROCm key-prefix correctness, reclaim route's exactly-once/bracketing behaviour (call-order + call-count pinned via a counting stub), real _reclaim_device_cache return-value logic with exception handling (CUDA unavailable → False, available → True, exception during reclaim → False).
- [x] Full sidecar suite (pytest -m "not golden") green — 55/55 in 	est_memory.py, full suite passing with 2 unrelated skips. Run independently twice (once via the implementer, once via a fresh review pass) against the bootstrapped venv, since this worktree's own venv isn't bootstrapped.
- [x] 
pm run verify:fast:branch — sidecar-only diff, all other legs correctly out-of-scope.

Design context: docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md §4 (already on main). No new/updated regression plan — small, localized diagnostic addition; the issue body + paired tests are the spec per CLAUDE.md. No release-notes entry — internal diagnostic tooling with no user- or operator-visible effect.

Closes #2423

### DIFF
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -89,7 +89,7 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
 
 ## Results
 
-| Point | When | Device | reserved MB | allocated MB | inactive_split MB | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
+| Point | When | Device | reserved bytes | allocated bytes | inactive_split bytes | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
 |---|---|---|---|---|---|---|---|---|---|---|
 | P0 | fresh | | | | | | | | | |
 | P1 | mid-render | | | | | | | | | |
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -35,7 +35,7 @@ Run this at each capture point and paste the output into the table below.
 $ts = (Get-Date).ToString('HH:mm:ss')
 $mem = Invoke-RestMethod http://127.0.0.1:9000/debug/memory
 "$ts  rss=$($mem.process.rss_mb)MB"
-$mem.cuda.per_device | ConvertTo-Json -Depth 4
+$mem.memory_stats | ConvertTo-Json -Depth 4
 $mem.engines | ConvertTo-Json -Depth 3
 nvidia-smi --query-gpu=index,memory.used,memory.total --format=csv,noheader
 ```
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -73,7 +73,8 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
    ```
 
    It returns the per-device figures **before and after** a bare
-   `empty_cache()`, bracketed with nothing in between. Record both as **P4**.
+   `empty_cache()`, bracketed with nothing in between, plus a `reclaimed: bool`
+   field indicating whether the reclaim actually ran. Record all three as **P4**.
 
    > Do **not** substitute a load/unload cycle. A load is an allocation pass that
    > refills and coalesces segments — conflating the two is the unsound inference
@@ -88,19 +89,19 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
 
 ## Results
 
-| Point | When | Device | reserved MB | allocated MB | inactive_split MB | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded |
-|---|---|---|---|---|---|---|---|---|---|
-| P0 | fresh | | | | | | | | |
-| P1 | mid-render | | | | | | | | |
-| P2 | +0 s | | | | | | | | |
-| P3 | +180 s | | | | | | | | |
-| P4 before | reclaim | | | | | | | | |
-| P4 after | reclaim | | | | | | | | |
-| P5 | cuda:1 render | | | | | | | | |
+| Point | When | Device | reserved MB | allocated MB | inactive_split MB | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
+|---|---|---|---|---|---|---|---|---|---|---|
+| P0 | fresh | | | | | | | | | |
+| P1 | mid-render | | | | | | | | | |
+| P2 | +0 s | | | | | | | | | |
+| P3 | +180 s | | | | | | | | | |
+| P4 before | reclaim | | | | | | | | | |
+| P4 after | reclaim | | | | | | | | | |
+| P5 | cuda:1 render | | | | | | | | | |
 
 **RSS at P2 vs 8192 MB:** ☐ above ☐ below
 *(Above means the memory watchdog's unconditional 60 s `gc+empty_cache` was
-already firing throughout — `main.py:7915`, `:7953-7960`.)*
+already firing throughout — `main.py:7957-7964` (`_mem_warn_threshold_mb()`).)*
 
 ---
 
@@ -110,7 +111,8 @@ already firing throughout — `main.py:7915`, `:7953-7960`.)*
 |---|---|
 | Strand absent at P3 | Self-heals at the existing TTLs. #1996 criterion 1 is already satisfied on `main`. |
 | P4 `reserved` drops sharply | **Uncollected cache.** A scheduled reclaim is the fix — subject to every constraint in §7 of the design. |
-| P4 `reserved` barely moves, `inactive_split` dominates | **Fragmentation.** `empty_cache()` cannot fix this at any cadence; re-open the recycle threshold instead. |
+| P4 `reserved` barely moves, `inactive_split` dominates, `reclaimed: true` | **Fragmentation.** `empty_cache()` cannot fix this at any cadence; re-open the recycle threshold instead. |
+| P4 `reserved` barely moves, `reclaimed: false` | **Invalid reading.** The reclaim never ran — CUDA was unavailable or `empty_cache()` threw. Retry once; treat repeated `false` as a separate finding (a live-context problem), not a #1996 answer. |
 | Strand present at P3 *and* P4 frees it *and* RSS was above 8192 MB at P2 | Contradicts the 60 s watchdog reclaim having run. Investigate that before designing anything — one of the two observations is wrong. |
 
 Record the outcome as a comment on
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -9632,15 +9632,14 @@ def debug_reclaim() -> dict[str, Any]:
     surface that brackets `empty_cache()` with nothing else in between (a
     load/unload cycle would conflate the reclaim with an allocation pass).
 
-    Not inert: it frees real memory when it runs. The watchdog calls
-    `_reclaim_host_and_vram()`, which runs `gc.collect()` unconditionally then
-    `empty_cache()` conditionally (if CUDA available); `_reclaim_device_cache`
-    reverses the order (CUDA check first, then gc only if available), and
-    gates the whole operation on an RSS threshold (see main.py:7957-7964,
-    `_mem_warn_threshold_mb()`), so this route can trigger a reclaim below that
-    threshold too. No auth wrapper, matching every other `/debug` route in
-    this family (`debug_codec_timing`/`debug_codec_timing_reset` above) —
-    loopback-bound by default via LOCAL_TTS_HOST configuration."""
+    Not inert: it frees real memory when it runs. The memory watchdog checks
+    an RSS threshold (see main.py:8662, `if warn_threshold > 0 and rss >= ...`)
+    before calling `_reclaim_host_and_vram()`; this route calls `_reclaim_device_cache`
+    directly without that check, so it can trigger the identical `gc.collect()` +
+    `empty_cache()` operation regardless of RSS — even below the watchdog's
+    threshold. No auth wrapper, matching every other `/debug` route in this family
+    (`debug_codec_timing`/`debug_codec_timing_reset` above) — loopback-bound by
+    default via LOCAL_TTS_HOST configuration."""
     before = _cuda_memory_stats_per_device()
     reclaimed = _reclaim_device_cache("debug-reclaim")
     after = _cuda_memory_stats_per_device()
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -3764,9 +3764,10 @@ class ReservationLedger:
 _reclaim_failure_traced = False
 
 
-def _reclaim_device_cache(device_key: str) -> None:
+def _reclaim_device_cache(device_key: str) -> bool:
     """Bare `gc.collect()` + `torch.cuda.empty_cache()`, with no target model
-    to unload first (#1976).
+    to unload first (#1976). Returns True if the reclaim ran (CUDA available +
+    no exception), False if CUDA is unavailable or an exception occurred.
 
     Every reclaim ELSEWHERE in this file — `unload()`, `unload_design()`,
     `maybe_free_idle*` (Qwen/Coqui), the `_idle_evict_steps` ladder — is a
@@ -3818,15 +3819,17 @@ def _reclaim_device_cache(device_key: str) -> None:
 
     Best-effort: any torch/CUDA failure (no CUDA build, a poisoned context) is
     swallowed here, matching every other reclaim path in this file — this is
-    a last-resort cushion, not a step whose failure should propagate."""
+    a last-resort cushion, not a step whose failure should propagate. Returns
+    False so callers can distinguish "ran" from "did not run"."""
     global _reclaim_failure_traced
     try:
         import torch  # type: ignore
 
         if not torch.cuda.is_available():
-            return
+            return False
         gc.collect()
         torch.cuda.empty_cache()
+        return True
     except Exception:
         log.warning(
             "stranded-cache reclaim failed for %s",
@@ -3834,6 +3837,7 @@ def _reclaim_device_cache(device_key: str) -> None:
             exc_info=not _reclaim_failure_traced,
         )
         _reclaim_failure_traced = True
+        return False
 
 
 # #1993 review (C1) — the per-device reclaim cooldown. A module constant, not
@@ -9628,14 +9632,16 @@ def debug_reclaim() -> dict[str, Any]:
     surface that brackets `empty_cache()` with nothing else in between (a
     load/unload cycle would conflate the reclaim with an allocation pass).
 
-    Not inert: it frees real memory (same effect as the existing 60s
-    watchdog tick). No auth wrapper, matching every other `/debug` route in
+    Not inert: it frees real memory when it runs. The watchdog performs an
+    identical reclaim operation, but only above the 8192 MB RSS threshold
+    (see main.py:7953-7960), so this route can trigger a reclaim below that
+    threshold too. No auth wrapper, matching every other `/debug` route in
     this family (`debug_codec_timing`/`debug_codec_timing_reset` above) —
-    guarded by being loopback-bound, per this route family's convention."""
+    loopback-bound by default via LOCAL_TTS_HOST configuration."""
     before = _cuda_memory_stats_per_device()
-    _reclaim_device_cache("debug-reclaim")
+    reclaimed = _reclaim_device_cache("debug-reclaim")
     after = _cuda_memory_stats_per_device()
-    return {"before": before, "after": after}
+    return {"before": before, "reclaimed": reclaimed, "after": after}
 
 
 @app.post("/load")
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -3879,7 +3879,7 @@ class PlacementController:
         reserve_mb: Optional[Callable[[], int]] = None,
         idle_evict_steps: Optional[Callable[[str, str], list["EvictStep"]]] = None,
         is_resident: Optional[Callable[[str], Optional[str]]] = None,
-        reclaim: Optional[Callable[[str], None]] = None,
+        reclaim: Optional[Callable[[str], bool]] = None,
     ) -> None:
         self.probe = probe
         self.footprints = footprints if footprints is not None else FootprintTable()
@@ -9632,9 +9632,12 @@ def debug_reclaim() -> dict[str, Any]:
     surface that brackets `empty_cache()` with nothing else in between (a
     load/unload cycle would conflate the reclaim with an allocation pass).
 
-    Not inert: it frees real memory when it runs. The watchdog performs an
-    identical reclaim operation, but only above the 8192 MB RSS threshold
-    (see main.py:7953-7960), so this route can trigger a reclaim below that
+    Not inert: it frees real memory when it runs. The watchdog calls
+    `_reclaim_host_and_vram()`, which runs `gc.collect()` unconditionally then
+    `empty_cache()` conditionally (if CUDA available); `_reclaim_device_cache`
+    reverses the order (CUDA check first, then gc only if available), and
+    gates the whole operation on an RSS threshold (see main.py:7957-7964,
+    `_mem_warn_threshold_mb()`), so this route can trigger a reclaim below that
     threshold too. No auth wrapper, matching every other `/debug` route in
     this family (`debug_codec_timing`/`debug_codec_timing_reset` above) —
     loopback-bound by default via LOCAL_TTS_HOST configuration."""
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -8044,6 +8044,51 @@ def _cuda_vram_mb_per_device() -> dict[str, dict[str, float]]:
         return {}
 
 
+def _cuda_memory_stats_per_device() -> dict[str, dict[str, int]]:
+    """{"cuda:N"|"rocm:N": {"reserved", "allocated", "inactive_split",
+    "num_alloc_retries"}} for every CUDA-API-visible device, sourced from
+    `torch.cuda.memory_stats(i)` (#2423 — instrument for #1996's design
+    question).
+
+    Neither `_cuda_vram_mb_per_device` above (only `reserved_mb`/`total_mb`)
+    nor `debug_memory`'s `cuda` block (current-device-only, no
+    `inactive_split`) can discriminate #1996's two candidate causes:
+    uncollected cache (a scheduled reclaim fixes it) vs allocator
+    fragmentation (`empty_cache()` can never return it) — `inactive_split`
+    is the field that tells them apart. Values are raw units straight from
+    `memory_stats` (bytes for the byte-valued keys, a count for
+    `num_alloc_retries`) — deliberately NOT converted to MB like the
+    neighboring `_mb` fields, since these are new keys with their own
+    semantics, not a replacement for them.
+
+    Device keys use `probe_capacity`'s `{kind}:{i}` format (`kind = "rocm"
+    if _cuda_is_rocm() else "cuda"`), NOT a hardcoded `cuda:` prefix like
+    `_cuda_vram_mb_per_device` — copying that pattern would mislabel an
+    AMD/ROCm box's devices. `torch.cuda.memory_stats` is a CUDA/ROCm-only
+    API, so non-GPU kinds (`cpu`, `mps`) are simply never emitted here.
+
+    Never raises; {} when CUDA is unavailable — same fail-open contract as
+    `_cuda_vram_mb_per_device`."""
+    try:
+        import torch  # type: ignore
+
+        if not torch.cuda.is_available():
+            return {}
+        kind = "rocm" if _cuda_is_rocm() else "cuda"
+        out: dict[str, dict[str, int]] = {}
+        for i in range(torch.cuda.device_count()):
+            stats = torch.cuda.memory_stats(i)
+            out[f"{kind}:{i}"] = {
+                "reserved": stats.get("reserved_bytes.all.current", 0),
+                "allocated": stats.get("allocated_bytes.all.current", 0),
+                "inactive_split": stats.get("inactive_split_bytes.all.current", 0),
+                "num_alloc_retries": stats.get("num_alloc_retries", 0),
+            }
+        return out
+    except Exception:
+        return {}
+
+
 class DeviceLedger:
     """Thread-safe per-card VRAM reader wrapping the Wave-1 sampler
     (_sample_card). Read from three thread contexts (the _memory_watchdog
@@ -9556,6 +9601,10 @@ def debug_memory() -> dict[str, Any]:
     except Exception:
         pass
     out["cuda"] = cuda
+    # #2423 — per-device memory_stats() block (reserved/allocated/
+    # inactive_split/num_alloc_retries), the instrument #1996's design
+    # decision needs. Additive: the `cuda` block above is untouched.
+    out["memory_stats"] = _cuda_memory_stats_per_device()
     return out
 
 
@@ -9568,6 +9617,25 @@ async def debug_codec_timing() -> dict:
 async def debug_codec_timing_reset() -> dict:
     _codec_timing_reset()
     return {"ok": True}
+
+
+@app.post("/debug/reclaim")
+def debug_reclaim() -> dict[str, Any]:
+    """One bare `_reclaim_device_cache()` bracketed by before/after per-device
+    `memory_stats` snapshots, in a single response (#2423). #1996's design
+    decision (scheduled reclaim vs a boundary-recycle redesign) hinges on
+    whether a bare reclaim recovers the stranded pool — this is the only
+    surface that brackets `empty_cache()` with nothing else in between (a
+    load/unload cycle would conflate the reclaim with an allocation pass).
+
+    Not inert: it frees real memory (same effect as the existing 60s
+    watchdog tick). No auth wrapper, matching every other `/debug` route in
+    this family (`debug_codec_timing`/`debug_codec_timing_reset` above) —
+    guarded by being loopback-bound, per this route family's convention."""
+    before = _cuda_memory_stats_per_device()
+    _reclaim_device_cache("debug-reclaim")
+    after = _cuda_memory_stats_per_device()
+    return {"before": before, "after": after}
 
 
 @app.post("/load")
--- a/server/tts-sidecar/tests/test_memory.py
+++ b/server/tts-sidecar/tests/test_memory.py
@@ -579,6 +579,114 @@ def test_debug_memory_endpoint_shape(monkeypatch):
             assert body["cuda"]["host_pinned_active_mb"] >= 0
 
 
+# --- #2423: per-device memory_stats() diagnostics + bare-reclaim endpoint ---
+
+
+def test_debug_memory_includes_memory_stats_per_device(monkeypatch):
+    """/debug/memory's `memory_stats` block carries reserved/allocated/
+    inactive_split/num_alloc_retries per device, keyed `cuda:{i}` — the field
+    (`inactive_split`) neither the existing `cuda` block (current-device-only,
+    no inactive_split) nor `_cuda_vram_mb_per_device` (reserved/total only)
+    can surface, per #2423's rationale."""
+    torch = pytest.importorskip("torch")
+    monkeypatch.delitem(main.ENGINES, "kokoro", raising=False)
+    monkeypatch.setitem(main.ENGINES, "qwen", main.QwenEngine())
+
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: True)
+    monkeypatch.setattr(torch.cuda, "device_count", lambda: 1)
+    monkeypatch.setattr(main, "_cuda_is_rocm", lambda: False)
+    stats = {
+        "reserved_bytes.all.current": 4_000_000,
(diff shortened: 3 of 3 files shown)


dudarenok-maker:
## PR review — pass 1 (head 49536feadfd10ee8aced170f432dbdad81436120, depth low)

Scope: `server/tts-sidecar/main.py` (+68) and `server/tts-sidecar/tests/test_memory.py` (+108) — the complete diff — read against #2423's acceptance list, `docs/superpowers/specs/2026-08-18-stranded-vram-reclaim-design.md` §4/§6/§7, and `docs/testing/1996-stranded-vram-measurement.md` (the run sheet these two endpoints exist to feed).

Verified before probing: ran `tests/test_memory.py` off this head with the primary checkout's bootstrapped sidecar venv — **53 passed**, matching the PR's claim; confirmed all four `torch.cuda.memory_stats` key names against the installed torch 2.11.0+cu128 (`torch/cuda/memory.py:249`, `:272`; torch's own `memory_reserved` reads `reserved_bytes.all.current` with the identical `.get(..., 0)` pattern at `:563`); grepped `server/src` for any Node-side consumer of `/debug/memory` (none).

---

### 🟠 Significant — `POST /debug/reclaim` returns the same thing whether the reclaim ran or silently failed, and that ambiguity lands exactly on #1996's decision

**`server/tts-sidecar/main.py:9635-9638`** (correctness bug)

`_reclaim_device_cache` (`:3767`) returns `None` and swallows every exception (`:3830-3836`); it also early-returns at `:3826-3827` when `torch.cuda.is_available()` is False. The route neither inspects an outcome (there is none to inspect) nor reports one.

Concrete failure: a sidecar whose CUDA context is poisoned mid-session — the exact failure mode `_reclaim_device_cache`'s own docstring names at `:3819-3821`, and the one `start.ps1`'s exit-code-42 restart path exists for. `empty_cache()` raises, the `except` swallows it, and the route answers **HTTP 200** with `{"before": X, "after": X}` — non-empty, identical, per-device figures and no field anywhere saying the reclaim did not execute. The only signal is one `log.warning("stranded-cache reclaim failed for %s")`, and the `_reclaim_failure_traced` latch (`:3764`, `:3834-3836`) drops the traceback on every occurrence after the first. The run sheet never tells the operator to look there — step 5 (`docs/testing/1996-stranded-vram-measurement.md:69-80`) says only "It returns the per-device figures before and after".

Why that specific output is the damaging one: identical before/after *is* the reading the decision table routes to Mechanism B. `docs/testing/1996-stranded-vram-measurement.md:113` — "P4 `reserved` barely moves, `inactive_split` dominates → **Fragmentation.** `empty_cache()` cannot fix this at any cadence" — and design §6 row 2 — "Re-open the recycle threshold instead. No new watchdog; ... the issue text needs correcting." So a reclaim that *never ran* is indistinguishable from a reclaim that *ran and freed nothing*, on the one axis this whole endpoint was built to discriminate. That is a null observation passing as a positive result (reviewer-brief shapes 5 and 6), in an instrument whose entire purpose is to settle a design decision two prior attempts already got wrong.

`test_debug_reclaim_brackets_one_bare_reclaim_call` cannot catch this: its `fake_reclaim` is a no-op that always "succeeds", so no existing test goes red for the failure path.

Fix: surface whether the reclaim executed — `_reclaim_device_cache` returning a bool, or the route catching/checking itself — emitted alongside the snapshots (e.g. `{"reclaimed": true|false, "before": ..., "after": ...}`), with a paired test that scripts a raising `_reclaim_device_cache` and asserts the response says so.

---

### 🟠 Significant — the run sheet's capture command reads `$mem.cuda.per_device`, a key this PR does not create and nothing else does either

**`docs/testing/1996-stranded-vram-measurement.md:38`** vs **`server/tts-sidecar/main.py:9607`** (correctness bug)

The block ships at top-level `out["memory_stats"]` — which is correct, and matches both design §4 (`:132-134`) and #2423's acceptance item 1. But `debug_memory`'s `cuda` block (`:9562-9603`) has no `per_device` key, before this PR or after it, and `_cuda_vram_mb_per_device` (`:8015`) is never called from this route.

Concrete failure: run the run sheet's step-1 capture block verbatim against a sidecar on this head. `$mem.cuda.per_device` evaluates to `$null`, `ConvertTo-Json` prints `null`, and the operator captures nothing. The results table (`:91-99`) asks for `allocated` and `inactive_split` per device — precisely the two fields the run sheet's own prerequisite (`:18-21`) says it "cannot produce readings 1 and 2" without. After this PR it still cannot, because the documented command points at a key that does not exist. `$mem.memory_stats` works.

This PR's stated deliverable is "so #1996's design question can finally be measured". As shipped, it is not measurable via the documented procedure — the endpoint exists and its only consumer doesn't reach it. "It's pre-existing" is a void deferral reason here (CLAUDE.md → Incidental findings), and it barely applies anyway: this PR is what makes that run sheet executable, so it is the round that owes the correction.

Fix: `docs/testing/1996-stranded-vram-measurement.md:38` → `$mem.memory_stats | ConvertTo-Json -Depth 4`. Do **not** fix it the other way by renaming the response key to `cuda.per_device` — that would contradict spec §4 and #2423's acceptance, and would sit a second, differently-shaped `per_device` next to `_cuda_vram_mb_per_device`'s.

---

### 🟡 Minor — the new docstring's safety justification is factually wrong in both of its halves

**`server/tts-sidecar/main.py:9631-9634`** (cleanup / stale-comment chore — but a factual error, not a style preference)

**(a) "same effect as the existing 60s watchdog tick."** The watchdog reclaim is not unconditional. `_memory_watchdog` calls `_reclaim_host_and_vram()` at `:8659`, gated on `if warn_threshold > 0 and rss >= warn_threshold`, where `_mem_warn_threshold_mb()` defaults to **8192 MB** (`:7953-7960`) — and that whole branch already sits inside the 60 s throttle at `:8628`. At any steady-state RSS below 8 GB — including a fresh sidecar, which is the run sheet's own P0 baseline — the watchdog reclaim has *never* run in that process, so `POST /debug/reclaim` is not "doing what already happens anyway".

The run sheet already encodes the correct reading, which is what makes the docstring's version demonstrably wrong rather than merely loose: `:101-103` makes "RSS at P2 above 8192 MB" an explicit checkbox precondition, and decision row `:114` treats a strand at P3 *with* RSS above 8192 as a contradiction to investigate. The claim originates in design §4 (`:143-145`) and is repeated in #2423's "Not in scope" — which cites `main.py:8613-8614`, the VRAM **soft-recycle** branch, containing no reclaim call at all. This PR is where it entered `main.py`.

**(b) "guarded by being loopback-bound."** The bind host is `LOCAL_TTS_HOST` with a `127.0.0.1` *default* (`server/tts-sidecar/start.ps1:99`, `:137`) — an overridable default, not an invariant. No new exposure (the route is strictly less destructive than the existing unauthenticated `POST /recycle`), so this is a precision problem in a stated guarantee, not a security finding.

🟡 rather than 🟠 because the runtime effect is nil and an operator following the run sheet is not misled — only a code reader is, and the next reader is whoever designs #1996's lever off this docstring. Still owed this round per CLAUDE.md's stale-comment rule. Correcting design §4 and #2423's note in the same edit would stop the third copy.

---

### ✅ What is solid

- **Device-key derivation is right and is really pinned.** `kind = "rocm" if _cuda_is_rocm() else "cuda"` rather than `_cuda_vram_mb_per_device`'s hardcoded `cuda:` — #2423 acceptance item 4 and design §7 constraint 5 — and `test_..._uses_rocm_prefix_on_rocm_box` asserts both the positive (`["rocm:0"]`) and the negative (`"cuda:0" not in out`).
- **The four torch key names are correct**, verified against the installed torch rather than assumed (see the Verified line above). The `.get(..., 0)` defaults mirror torch's own `memory_reserved`.
- **Genuinely additive.** `out["cuda"]` is byte-for-byte untouched, nothing under `server/src` parses `/debug/memory`, and `/health`'s closed `vram_reserved_mb_by_device` element type is deliberately not routed through — exactly what #2423's "Not in scope" required.
- **The reclaim test meets design §7 constraint 7's bar** — it pins call *order* and *count* (`calls == ["snapshot", "reclaim", "snapshot"]`), not merely "the reclaim was called", so a mutation relocating the call turns it red.
- Only CUDA-API devices are emitted — no empty stat blocks for `cpu`/`mps:0` (#2423 acceptance item 4's second half).
- No new background loop, so `tests/test_lifespan_order.py`'s pinned nine startup handlers stay valid (design §7 constraint 6).
- **House gates:** `Closes #2423` present, plain, outside any code span. Release-notes pair carries an explicit not-applicable with a stated reason. On-box register correctly untouched — the run sheet itself declares this a diagnostic, not an acceptance run (`:9-10`). No `cast.json` writes, no new config knob, no derived artifact staled.

---

### Verdict

**Not clear to merge.** Two 🟠 correctness findings, one 🟡. Both 🟠 findings share a root: the endpoints themselves are built correctly to spec, but the path from "endpoint exists" to "reading taken" is broken at each end — the consumer can't reach the data (finding 2), and the data can't distinguish a real negative from a silent no-op (finding 1). Since the sole purpose of #2423 is to produce a reading that settles #1996, shipping it in that state would deliver an instrument that can hand back a confidently wrong answer.

Per the gate's re-review trigger, the two 🟠 findings re-trigger a pass once fixed and pushed; the 🟡 is fixed this round but does not on its own reopen the loop.

*No tracked file was modified by this pass.*

dudarenok-maker:
## PR review — pass 2 (head ffe3ba738f9d8826a0acbab01db8afc3f8bd723f, depth low)

Scope: the fix commit `ffe3ba73` (3 files — `main.py`, `test_memory.py`, `1996-stranded-vram-measurement.md`) re-read against the full branch diff · verified: `test_memory.py` 54/54 green and `test_memory.py + test_placement.py` 104/104 green against the bootstrapped venv (this worktree has no `.venv`; used the primary checkout's), plus a mutation run against a TEMP copy, before adversarially probing whether each pass-1 finding is actually closed.

Deltas only, per the re-review format.

---

**🟠 Pass-1 finding 1 (`/debug/reclaim` cannot distinguish "ran and freed nothing" from "never ran") — PARTIALLY RESOLVED.**
The API half is genuinely fixed. `_reclaim_device_cache` (`server/tts-sidecar/main.py:3767`) now has exactly three exits and every one returns a `bool` — `return False` at `:3829` (CUDA unavailable), `return True` at `:3832` (success), `return False` at `:3840` (exception). No implicit-`None` path survives, `import torch` failing lands in the same `except Exception`. Threaded into the response at `:9642`→`:9644`. Both existing call sites survive the signature change: `:4063` is `await asyncio.to_thread(self.reclaim, device_key)`, a bare call that discards the return, and `:3895` only stores the reference. Grep of every `_reclaim_device_cache(` occurrence shows no other caller.
**What is still open is the half the finding existed for** — see new finding A below. The signal is now produced and nothing consumes it.

**🟠 Pass-1 finding 2 (run sheet read a JSON key that does not exist) — VERIFIED RESOLVED.**
`docs/testing/1996-stranded-vram-measurement.md:38` now reads `$mem.memory_stats`, and `main.py:9611` sets `out["memory_stats"] = _cuda_memory_stats_per_device()` at top level. The four keys the sheet tells the operator to record (line 43: `reserved`, `allocated`, `inactive_split`) match `_cuda_memory_stats_per_device`'s emitted names at `main.py:8086-8089` exactly. Confirmed no `cuda.per_device` key exists anywhere in `main.py`.

**🟡 Pass-1 finding 3 (two wrong docstring claims) — RESOLVED on substance, but the correction introduced a new wrong citation.**
- *"loopback-bound" overstated* → **VERIFIED RESOLVED**. `main.py:9640` now reads "loopback-bound by default via LOCAL_TTS_HOST configuration", and that is accurate: `server/tts-sidecar/start.ps1:99` and `start.sh:44` both default the bind host to `127.0.0.1` with `LOCAL_TTS_HOST` as the override. Correctly stated as a default, not an invariant.
- *watchdog-equivalence ignored the RSS gate* → the gate is now named, which was the substance of the finding. But its supporting `file:line` is wrong — see finding B.

---

### 🟠 Significant — the run sheet, the only consumer of `reclaimed`, was never updated; its decision table still misdiagnoses `reclaimed: False` as "Fragmentation"

`docs/testing/1996-stranded-vram-measurement.md:75-76`, `:97-98`, `:113`

Finding 1's stated purpose was so #1996's run sheet could tell the two cases apart, and the fix commit's own message claims it is "proving #1996's run sheet can distinguish the two cases". The run sheet was not touched apart from line 38. Three places still carry the old two-key contract:

- **line 75-76** enumerates the response for the operator as "the per-device figures **before and after** a bare `empty_cache()`, bracketed with nothing in between. Record both as **P4**." A third key is never mentioned, so an operator following the sheet has no reason to look at it.
- **lines 97-98** give the results table exactly two P4 rows (`P4 before` / `P4 after`) and no column or row for `reclaimed`. The field is not recordable even if noticed.
- **line 113** is where it turns into a wrong answer.

Concrete failure scenario — the exception path, which is the one that matters here because a poisoned CUDA context is a live hypothesis in the very investigation this sheet drives:

1. CUDA is available, so `_cuda_memory_stats_per_device()` (`main.py:8079`) returns real non-empty figures at both snapshots.
2. `torch.cuda.empty_cache()` throws inside `_reclaim_device_cache` → `main.py:3833-3840` logs and returns `False`.
3. `POST /debug/reclaim` returns real `before`, real `after`, **identical**, and `reclaimed: false`.
4. The operator reads the sheet's "Reading the result" table, line 113: *"P4 `reserved` barely moves, `inactive_split` dominates → **Fragmentation.** `empty_cache()` cannot fix this at any cadence; re-open the recycle threshold instead."*
5. They record **Fragmentation** and, per the design's decision tree, re-open the recycle threshold — off a reclaim that never executed.

That is precisely the misdiagnosis pass-1 finding 1 was raised to prevent, still fully armed on the decision surface. Note the sheet is a diagnostic whose output is written back to #1996 as the basis for a design decision (lines 116-120), so a wrong reading here propagates.

Fix shape: step 5 should name the `reclaimed` key and tell the operator to record it; the results table needs somewhere to put it; and the "Reading the result" table needs a `reclaimed: false` row that says "the reclaim did not run — this reading is void, investigate the sidecar log's `stranded-cache reclaim failed` warning" ahead of the two rows that currently absorb it.

---

### 🟡 Minor — the docstring's new line citation is stale by 4, invalidated by this same commit; the same stale pair sits in the run sheet

`server/tts-sidecar/main.py:9637` and `docs/testing/1996-stranded-vram-measurement.md:103`

The new docstring says the watchdog reclaim is gated "above the 8192 MB RSS threshold (see main.py:7953-7960)". In the **post-fix** file, `7953-7954` is the tail of `_disable_mkldnn()` — the `SIDECAR_DISABLE_MKLDNN` knob, entirely unrelated — and `_mem_warn_threshold_mb()` now occupies `7957-7964`, with the actual `8192` default at `:7962`/`:7964`. So the range opens on the wrong function and stops two lines short of the literal it points at.

Cause is self-inflicted and worth naming, because it will recur: this commit added **+4 lines** to `main.py` above line 7953 (the `-> bool` signature, two docstring lines, and the `return True` / `return False` additions at `:3822-3832`). Verified against the parent commit — in `49536fea:server/tts-sidecar/main.py`, `_mem_warn_threshold_mb` sits at exactly `7953-7960`. The citation was computed against the pre-fix file and its own commit moved the target.

The identical staleness sits in the file this commit *did* edit: `1996-stranded-vram-measurement.md:103` cites `` `main.py:7915`, `:7953-7960` ``. Pre-fix, `_MEM_WATCHDOG_INTERVAL = 60.0` was at `7915`; post-fix it is at `7919`. Both refs on that line are now off by the same 4, in a doc this commit touched.

Secondary, same docstring: "The watchdog performs an identical reclaim operation" is not quite identical — the watchdog calls `_reclaim_host_and_vram` (`main.py:1086`), which runs `gc.collect()` **unconditionally** before the CUDA check, whereas `_reclaim_device_cache` returns at `:3829` before any `gc.collect()` when CUDA is absent. Equivalent on the CUDA box this route targets; not equivalent in general. Also "the 8192 MB RSS threshold" reads as fixed when it is `SIDECAR_RSS_WARN_MB` (default 8192, and `<= 0` disables the reclaim entirely — `main.py:8662`).

---

### 🟡 Minor — neither new test exercises the real `_reclaim_device_cache`; the regression it was written for goes undetected

`server/tts-sidecar/tests/test_memory.py:680` and `:721`

Both tests `monkeypatch.setattr(main, "_reclaim_device_cache", <fake returning a hardcoded bool>)`, and the route at `main.py:9642` resolves that name through the module attribute — so the real function body never runs in either test. They pin the route's *plumbing* (that the route forwards whatever the callable returned), not the change finding 1 actually asked for.

Not reasoned — measured. I copied the sidecar tree to a TEMP dir, reverted `_reclaim_device_cache` to its exact pre-fix body (dropped `return True`, restored the bare `return`, dropped the `except`-block `return False`), and ran the suites against the mutant:

```
pytest tests/test_memory.py tests/test_placement.py -q -m "not golden"
........................................................................ [ 69%]
................................                                         [100%]
```

104/104 green. With that mutant in production, `POST /debug/reclaim` returns `{"before": …, "reclaimed": null, "after": …}` on **every** path — pass-1 finding 1's ambiguity fully restored, JSON `null` for both "ran" and "did not run" — and nothing in the memory or placement suites goes red. Against the house gate ("the paired test is red before the fix, and red *for the reason claimed*"), these are red before the fix only for the route-plumbing half (`KeyError: 'reclaimed'`), never for the function-return half.

Fix shape: one direct unit test on `_reclaim_device_cache` itself, monkeypatching `torch.cuda.is_available` → `False` (assert `is False`), a success path (assert `is True`), and an `empty_cache` that raises (assert `is False`). No fake of the function under test.

---

### 🟡 Minor — the injection point's type annotation now contradicts its own default

`server/tts-sidecar/main.py:3882` still declares `reclaim: Optional[Callable[[str], None]] = None`, while the default assigned at `:3895` is `_reclaim_device_cache`, which is now `Callable[[str], bool]`. A `bool` return is not assignable to a declared `None` return. No Python typechecker is wired into this repo (no `mypy.ini` / `pyrightconfig.json` / ruff config found), so this does not break a gate — it is a stale annotation this change made false, and it is the one place a reader looks to learn the injected callable's contract.

---

### 🟡 Minor — PR body is stale and still asserts the claim finding 3 corrected

The Test plan says "**4** new pytest cases" and "**53/53** in `test_memory.py`"; the branch now adds 5 (`git diff origin/main...HEAD -- …/test_memory.py | grep -c '^+def test_'` → 5) and the file runs 54.

More substantively, Summary bullet 4 still reads: "it's a **loopback-bound** `/debug` route doing **exactly what the 60s watchdog tick already does unprompted**." Those are verbatim the two claims pass-1 finding 3 established as wrong and this commit corrected in the docstring — the RSS gate makes the watchdog tick conditional, and loopback is a default rather than an invariant. The PR body is the shipping record; leaving the corrected claim standing there re-publishes it.

---

### Verdict

**FAIL** — pass-1 findings 2 and 3 are closed (3 with a caveat), but finding 1 is only half-landed: the `reclaimed` signal is produced correctly and no consumer reads it, leaving the run sheet's decision table returning "Fragmentation" for a reclaim that never ran. One 🟠 correctness finding open plus four 🟡 cleanups. Fixing and pushing re-triggers pass 3, which is the loop cap.

The production code changed by this commit is correct as written — every path I traced returns the right value. The gaps are in the surfaces around it: the document that consumes the signal, the citations, the test that would catch a revert, and the annotation.

dudarenok-maker:
## PR review — pass 3 (head 2fee0f944451e8d9b4ed6a1d8f6a68e4f6532555, depth low)

**This is the loop's cap round** (initial pass + 2 re-reviews = 3). Whatever the outcome, the automatic fix-and-re-review loop stops here; anything left open is handed to the maintainer rather than looped again.

Scope: the fix commit `2fee0f94` (3 files — `1996-stranded-vram-measurement.md`, `main.py`, `test_memory.py`) re-read against the full branch diff `87c0f72d..2fee0f94` · verified: `test_memory.py` **55/55 green** against the primary checkout's bootstrapped venv (this worktree still has no `.venv`), every line citation re-derived from the current file by grep rather than trusted, and **two mutation runs against a TEMP copy** of the sidecar tree before adversarially probing each pass-2 disposition. CI on this head is green across all 15 checks, including `Sidecar tests (pytest)`.

Deltas only, per the re-review format.

---

**🟠 Pass-2 finding A (run sheet's decision table misdiagnoses `reclaimed: false` as Fragmentation) — VERIFIED RESOLVED.**

All three surfaces are genuinely wired, and I walked the exception scenario the finding was built on:

- **Response shape** — `docs/testing/1996-stranded-vram-measurement.md:75-77` now names the field ("plus a `reclaimed: bool` field indicating whether the reclaim actually ran. Record all three as **P4**"), matching `main.py:9647`'s `{"before": …, "reclaimed": …, "after": …}`.
- **Results table** — `:92-93` adds a `reclaimed` column; header, separator and all seven body rows are 11 cells wide, so the table still renders.
- **Decision table** — `:114` now requires `reclaimed: true` for the Fragmentation verdict; `:115` adds the `reclaimed: false` → **Invalid reading** row.

Re-walked literally, with CUDA available and `empty_cache()` throwing (the exact repro from pass 2): `before` and `after` are real and identical, `reclaimed: false`. Row `:114` no longer matches (it demands `true`), row `:115` does, and the operator lands on "Invalid reading … not a #1996 answer." The misdiagnosis is closed. The two rows are mutually exclusive on `reclaimed`, so there is no path where both match.

*Not counted as a finding, but worth a line while someone is in the file:* `:115` names two causes ("CUDA was unavailable **or** `empty_cache()` threw") and gives the operator nothing to discriminate them with. The sidecar log's `stranded-cache reclaim failed for %s` warning (`main.py:3834-3838`) fires on the second and not the first, so a pointer to it would make "treat repeated `false` as a separate finding" actionable.

**🟡 Pass-2 finding B (stale citations + inaccurate "identical" claim) — RESOLVED, with one new inaccuracy introduced by the correction (finding 1 below).**

Re-derived fresh, not trusted:

| Claim | Cited | Actual (grep, current head) |
|---|---|---|
| RSS threshold fn, docstring `main.py:9639` | `main.py:7957-7964` | `def _mem_warn_threshold_mb()` at **7957**, `return 8192.0` at **7964** ✅ |
| RSS threshold fn, run sheet `:104` | `main.py:7957-7964` | same ✅ |

Both corrected. The "identical reclaim operation" claim is gone and the replacement's *ordering* half is accurate — I read both bodies: `_reclaim_host_and_vram` (`main.py:1086-1102`) calls `gc.collect()` unconditionally at `:1095` then `empty_cache()` behind the CUDA check at `:1099-1100`; `_reclaim_device_cache` (`:3825-3832`) checks CUDA at `:3828` and only then runs `gc.collect()` at `:3830`. "Reverses the order" is correct.

Two notes on how it was resolved. (a) The run sheet's *second* citation was **deleted rather than updated**: pass 2 flagged `` `main.py:7915`, `:7953-7960` `` as both off by 4, and `:104` now carries only the threshold ref. `_MEM_WATCHDOG_INTERVAL = 60.0` is at **`main.py:7919`** today, so the sentence's "60 s" figure is now uncited. Not wrong, just unsupported. (b) Neither citation points at the gate itself — `if warn_threshold > 0 and rss >= warn_threshold: … _reclaim_host_and_vram()` is at **`main.py:8662-8663`**, and that is the line the surrounding prose is actually about.

**🟡 Pass-2 finding C (both reclaim tests mock away the function under test) — VERIFIED RESOLVED, measured.**

`tests/test_memory.py:733-757` calls `main._reclaim_device_cache("test-device")` directly and patches only `torch.cuda.is_available` / `torch.cuda.empty_cache` on the real torch singleton — no fake of the function under test. `pytest.importorskip("torch")` matches the 7 pre-existing uses in this file and does **not** skip here (torch resolves; the case ran and passed).

Not reasoned — mutated. Copied the sidecar tree to a TEMP dir and replaced all three exits of `_reclaim_device_cache` with `return None`:

```
FAILED tests/test_memory.py::test_reclaim_device_cache_exercises_real_function
E   AssertionError: Should return False when CUDA is unavailable
E   assert None is False
1 failed, 3 passed, 51 deselected
```

The new test is the *only* one that goes red — the two route-level tests stay green under the mutant, confirming pass 2's diagnosis and that this test is what now covers the gap. One branch of three is still unpinned; see finding 2.

**🟡 Pass-2 finding D (stale `Callable[[str], None]`) — VERIFIED RESOLVED.**

`main.py:3882` reads `reclaim: Optional[Callable[[str], bool]] = None`, matching the default assigned at `:3895`. Grepped for the class rather than the instance: the only other `Callable[[str], None]` occurrences are `report_progress` at `:5956` and `:6288`, unrelated. No sibling left armed.

**🟡 Pass-2 finding E (PR body stale / self-contradicting) — NOT RESOLVED. The substantive half was not touched, and the numeric half is wrong again.**

See finding 3.

---

### 🟡 Minor — the corrected docstring now attributes the RSS gate to the wrong function, and contradicts its own conclusion two lines later

`server/tts-sidecar/main.py:9637-9641`

```
`empty_cache()` conditionally (if CUDA available); `_reclaim_device_cache`
reverses the order (CUDA check first, then gc only if available), and
gates the whole operation on an RSS threshold (see main.py:7957-7964,
`_mem_warn_threshold_mb()`), so this route can trigger a reclaim below that
threshold too.
```

After the semicolon the subject is `_reclaim_device_cache`, and `and gates …` is its second coordinated predicate. Read literally, the docstring states that **`_reclaim_device_cache` gates the whole operation on an RSS threshold**. It does not: `main.py:3824-3840` is the entire function body and contains no threshold check of any kind. The RSS gate belongs to the watchdog (`main.py:8662`, `if warn_threshold > 0 and rss >= warn_threshold`). The clause that follows — "so this route can trigger a reclaim below that threshold too" — only parses if the *watchdog* is the gated one, so the sentence refutes itself within four lines.

Concrete failure: a maintainer reading this route's docstring to decide whether `/debug/reclaim` duplicates the watchdog is told the route is RSS-gated, then told it is not. The paragraph exists precisely to answer "is this route redundant?", which is the question it now answers both ways. This is the same shape pass-2 finding B raised (an inaccurate claim in this docstring), relocated by the fix rather than removed — the correction attached the right fact to the wrong subject.

Fix shape: make the subject explicit — `…reverses the order (CUDA check first, then gc only if available). The watchdog's call is gated on an RSS threshold (`_mem_warn_threshold_mb()`, `main.py:7957-7964`; the gate itself at `:8662`); this route is not, so it can trigger a reclaim below that threshold too.`

---

### 🟡 Minor — the `empty_cache()`-throws → `False` branch is pinned by no test in the file, and the run sheet shipped this round depends on it

`server/tts-sidecar/main.py:3833-3840` · `server/tts-sidecar/tests/test_memory.py:733-757` · `docs/testing/1996-stranded-vram-measurement.md:115`

`_reclaim_device_cache` has three exits. The new test covers two (CUDA unavailable → `False`, success → `True`). The third — the `except Exception` handler that logs and returns `False` — is covered by nothing.

Measured, not inferred. Second mutation run against the TEMP copy, changing **only** `main.py:3840` from `return False` to `return True` and leaving the rest of the function intact:

```
pytest tests/test_memory.py -q
.......................................................   [100%]
55 passed
```

All 55 cases green with the exception path returning the wrong value. Pass 2's suggested fix shape named three cases explicitly, including "an `empty_cache` that raises (assert `is False`)"; two of the three landed.

Why it matters beyond coverage bookkeeping: `docs/testing/1996-stranded-vram-measurement.md:115` — a row added by *this* commit — tells the operator that `reclaimed: false` means "CUDA was unavailable **or `empty_cache()` threw**". A regression on that second branch would return `reclaimed: true` for a reclaim that raised, and the operator would fall through to `:114` and record **Fragmentation** off a reclaim that never completed. That is finding A's misdiagnosis, one branch over, with no test standing in front of it.

Fix shape: extend the same test with a third case — `monkeypatch.setattr(torch.cuda, "empty_cache", <raises>)`, assert `is False`. Note `main.py:3824`'s `_reclaim_failure_traced` module global latches on first failure, so reset or tolerate it if the case is ordered before others.

---

### 🟡 Minor — PR body: finding E's substantive half is untouched, and the test count is wrong again in the same direction

Summary bullet 4 still reads, verbatim and unchanged since pass 1:

> it's a **loopback-bound** `/debug` route doing **exactly what the 60s watchdog tick already does unprompted**.

Those are the two claims pass-1 finding 3 established as wrong and this branch corrected *in the docstring* — the RSS gate makes the watchdog tick conditional (`main.py:8662`), and loopback is a `LOCAL_TTS_HOST` default rather than an invariant (which is why the docstring now says "loopback-bound **by default**"). Pass 2 named this bullet specifically; `2fee0f94` edited only the Test plan. The PR body is the shipping record, so the retracted claim ships as the summary of the change that retracted it.

Test plan: "**5** new pytest cases in `server/tts-sidecar/tests/test_memory.py`". Actual, measured against the branch base:

```
$ git diff 87c0f72d..2fee0f94 -- server/tts-sidecar/tests/test_memory.py | grep -c '^+def test_'
6
```

The six are `..._includes_memory_stats_per_device`, `..._empty_when_cuda_unavailable`, `..._uses_rocm_prefix_on_rocm_box`, `..._brackets_one_bare_reclaim_call`, `..._surfaces_reclaimed_false_on_cuda_unavailable`, `..._exercises_real_function`. The body's colon-list enumerates five of them and silently omits `..._surfaces_reclaimed_false_on_cuda_unavailable` — the test pass 1's fix added.

The cause is worth naming because it is the same one pass 2 diagnosed for the line citations: the number was corrected against the pre-fix state (4 → 5 was right *before* this commit) and this commit's own new test moved the target. `55/55 in test_memory.py` is correct — I collected 55 and all 55 pass.

---

### 🟡 Minor — the results table labels its columns MB while the endpoint this PR ships emits raw bytes

`docs/testing/1996-stranded-vram-measurement.md:92` · `server/tts-sidecar/main.py:8062-8066`, `:8085-8090`

`_cuda_memory_stats_per_device` is explicit that its values are raw units — "bytes for the byte-valued keys … **deliberately NOT converted to MB** like the neighboring `_mb` fields". The results table header, a line this commit rewrote to add the `reclaimed` column, reads `reserved MB | allocated MB | inactive_split MB | reserved − allocated`, and step 5's capture instructions (`:43-44`, `:71-77`) never tell the operator to divide. An operator pasting `$mem.memory_stats` straight into the table records ~4,000,000,000 under a header that says MB, in the same row as `RSS MB` and `nvidia-smi used` which *are* MB.

Bounded honestly: this cannot flip the decision table, because every row at `:112-116` is a ratio or a within-reading comparison ("drops sharply", "barely moves", "dominates") and units cancel. It is a mislabeled record that gets pasted back into #1996 as the evidence for a design decision, not a wrong verdict. Fix is one of: drop " MB" from the three headers, or add "÷ 1e6" to step 5.

---

### Verdict

**PASS** — no 🔴 or 🟠. All four pass-2 🟡 findings and the 🟠 finding A are verified against the current tree rather than the self-report: **A, B, C, D resolved** (A and C measured, not read), **E not resolved**. Five 🟡 cleanups remain, three of them new in `2fee0f94`.

The production code is correct. `_reclaim_device_cache`, the `/debug/reclaim` route, and `_cuda_memory_stats_per_device` all behave as documented on every path I traced, and the mutation runs confirm the new test is real. Everything above is in the surfaces around the code — a docstring sentence that attributes a fact to the wrong function, one untested branch, the PR body, and a unit label.

**Loop cap reached.** Per the `pr-review-gate` re-review cap (initial pass + 2 rounds), there is no pass 4 regardless of what happens next. These five 🟡 are still owed a fix in this round under CLAUDE.md's report-fix-record rule — 🟡 means "does not re-trigger a review", not "defer" — but the decision on whether to fix-then-merge or merge-then-fix, and any judgement about the docstring rewrite in finding 1, is handed to the maintainer rather than looped automatically.

Process note, not a finding: the depth ladder re-derives per pass from the PR's current commit set, and this PR now carries two `fix(sidecar)` commits alongside the original `chore(sidecar)`, which the ladder would score `medium`, not `low`. Dispatched as `low`; I covered the `medium` scope on the changed surface anyway (full branch diff, both reclaim functions, every call site, mutation testing). Flagging it so the ladder's own arithmetic is on the record.

Issue #2600, opened 2026-08-22
install-ort.mjs never requests nvidia-cudnn-cu12, so Kokoro GPU tests fail with silent CPU fallback


dudarenok-maker:
## Summary

`install-ort.mjs` installs `onnxruntime-gpu` with `--no-deps` and never requests the optional `[cudnn]` extra, so `nvidia-cudnn-cu12~=9.0` — a runtime dependency of `onnxruntime-gpu 1.26.x` for CUDA 12 execution — is never installed into the venv.

This leaves Kokoro on a real on-box test unable to construct an InferenceSession with `CUDAExecutionProvider`. The provider appears available to `onnxruntime.get_available_providers()` but session construction silently falls back to `CPUExecutionProvider`, with no error, no warning, and no indication to the operator — Kokoro reports success and returns audio, but it's running on CPU instead of GPU.

## Evidence

- Blocks on-box acceptance rows A37, A38, A39 — every Kokoro GPU-provider verification fails with the silent fallback
- `onnxruntime-gpu 1.26.0` declares `nvidia-cudnn-cu12~=9.0` as a dependency only in its optional `[cudnn]` extra
- `install-ort.mjs` runs `pip install --force-reinstall --no-deps onnxruntime-gpu>=1.26,<1.27`
- No cuDNN 12 runtime package ever lands in the venv, even after a bootstrap-venv run
- Wave-4 on-box re-run (2026-08-21) on a box with CUDA 12.4 confirmed the silent CPU fallback

## Root cause

PR #2576 re-pinned `ONNXRUNTIME_GPU_CONSTRAINT` to `>=1.26,<1.27` (CUDA 12 line) to fix #2534's CUDA 13/cuDNN 9 mismatch. That fix correctly addresses the GPU runtime libraries on the box, but `install-ort.mjs` was never updated to request the corresponding pip-vendored cuDNN 12 runtime that `onnxruntime-gpu 1.26.x` itself requires to work.

The original `--no-deps` design assumed the box's system CUDA toolkit would provide all runtime libraries. That assumption is true for the basic CUDA libraries but not for cuDNN — a separate, optional library that `onnxruntime-gpu` expects to find.

## Decision owed

Add the cuDNN 12 runtime to `server/tts-sidecar/scripts/install-ort.mjs` so the venv is self-contained and doesn't silently degrade to CPU when the box's system CUDA toolkit is missing cuDNN 12 (or any future version mismatch). Concrete options:

[...]

PR #2626, opened 2026-08-23
docs(docs): clear the agent-runnable register rows and re-bin the browser pack

dudarenok-maker:
## Summary

Wave 5 of the on-box acceptance register campaign (#2435): discharges A27 (`qa.asr.model` reaches the sidecar and every server-side reader), A45 (`speaker-qa.txt` reqHash fix drives a one-time `pip-in-place` reinstall), and B2/old (Stage-1 returns cast names in the manuscript's own script) against real hardware, reconfirms E9 without discharging it, and re-bins the device-browser pack (E5 moves to agent-runnable; G1/G2 reclassified opportunistic, stay owed). Closes #2606.

**Register totals: 69 → 66 owed**, re-derived by counting `### ` row headings per `## ` group in the edited body (not by arithmetic subtraction) — Group A 46→44, Group B 3→2, all other groups unchanged. Verified independently during this verify step by counting the same headings in the merged file: A 44, B 2, C 4, D 3, E 9, G 2, H 2 = 66, plus 5 Blocked (71 `### ` headings total, matching 74 baseline − 3 discharged rows removed).

Discharged rows were **removed outright**, not annotated with an outcome note, per the owner's 2026-08-22 ruling — confirmed no live `### ` row heading for any of the three discharged rows remains in either the register or the live view; only historical "Correction" narration (which the file's own convention keeps) still names them.

## Verification performed (this step, read-only except this PR + bookkeeping)

- Re-read every wave-5 evidence file (`docs/testing/onbox-wave5-results/step-{1..6}-*.md`) against #2606's acceptance criteria. All four runs (A27, A45, E9, B2) paste real command output, not source reasoning about what the code would do.
- Independently recounted the register's `### ` headings per group — matches step 6's claimed 66.
- Independently re-verified the **live sidecar venv is byte-unchanged**: `C:\Claude\Projects\Audiobook-Generator\server\tts-sidecar\.venv` — file count 56542, total size 6483355726 bytes, `pyvenv.cfg` SHA-256 and `.venv-stamp.json` reqHash all match the before/after readings pasted in steps 1 and 2, confirmed fresh from this verify pass, not just trusted from the evidence files.
- Ran `npm run check:onbox-register` myself: **green** — register and live view agree.
[...]

Issue #2625, opened 2026-08-23
[agent instructions][claude][task] ORT rows - fold the verdicts into the register and the live view


dudarenok-maker:
Fold step for #2623 — the ORT marker rows (A37, A38, A39, A40).

**You are the sole writer of the register and the live view for this branch.**

## Inputs

Read every evidence file the three run children committed (`docs/testing/onbox-<...>-results/` — the run children name their own paths in their receipts). Their **VERDICT** lines are your input. Do not re-run anything and do not second-guess a verdict; if one is ambiguous or missing, say so and leave that row owed.

## What to do

1. **For each DISCHARGED row: REMOVE it from `docs/testing/onbox-acceptance-register.md` entirely.** Owner ruling, 2026-08-22 — discharged rows are removed, never annotated with an outcome note and carried. Evidence lives in the results directory and the run sheet `docs/testing/ort-marker-onbox-acceptance.md`, not in the register.
2. For each STILL OWED / BLOCKED row: leave the row and add a dated note recording what was attempted and what stopped it.
3. **Fill in the run sheet** `docs/testing/ort-marker-onbox-acceptance.md` — its `Result:` lines are part of this bookkeeping, not optional.
4. **Renumber contiguously within Group A**, per the register's own convention, and update every cross-reference in the body. **Match rows by identity, not by number** — a renumber that silently re-points a cross-reference at a different row is the failure mode here.
5. **Re-derive the totals by COUNTING `###` headings per `##` group.** Do not subtract from any previously stated total — `main` moves, and a compensating pair of errors passes a subtraction check.
6. Update the **"At a glance"** table, the stated total, and add a dated correction note in the file's existing style.
7. **Update `docs/testing/onbox-acceptance-register-live-view.html`** — a **hand-authored HTML page, not a rendering of the markdown.** Edit that file. Its derived figures (owed count, per-group counts, oldest debt) must be **recomputed**, not carried over.
8. Run `npm run check:onbox-register` and paste the output.
9. **If the runs found #2617's fix insufficient** — e.g. the provider still falls back, or the `[cuda]` extras turned out to be needed — that is a **finding**, and it belongs in the evidence and in your report. Do not fix it and do not edit source.

## Do NOT publish the live view

Publishing is a manual operator step **after merge**, reusing the artifact URL recorded in the register's header. Edit the tracked file only. Never publish this `.md` to that URL, and never mint a new artifact.

## Acceptance

- Every run child's verdict accounted for, by row.
[...]

Issue #2621, opened 2026-08-22
[agent instructions][claude][task] wave 5 ORT step A - run A37 and A38, the shared Kokoro GPU-provider sub-check


dudarenok-maker:
Wave 5 ORT step A (#2606). Run register rows **A37** and **A38** — the ORT pip-consistency marker (plan 282, #2192). **No GPU needed, sidecar venv only.**

These two share the **Kokoro GPU-provider sub-check** that #2600 was blocking, which is why they run together.

## What to do

1. Do the dependency check below **first**.
2. Read A37's and A38's rows and the run sheet `docs/testing/ort-marker-onbox-acceptance.md`. **Those govern**; this brief only orients you. Follow the design doc's criteria table that A38's row cites.
3. **A37 — fresh NVIDIA bootstrap.** The row asks for a venv rebuilt from scratch. Build a **new** venv at a scratch path — that is not the same as modifying the live one, and it is allowed. **Never touch** `C:\Claude\Projects\Audiobook-Generator\server\tts-sidecar\.venv`; byte-verify it unchanged at the end and paste both readings.
4. **A38 — the reported bug: in-app Qwen3 install**, with the app running. Use **this worktree's** app on its own ports (VITE 5293 / API 8200), confirmed from the server's own output. Browser tools are available (`mcp__claude-in-chrome__*`).
5. **The shared sub-check is the point.** Confirm the Kokoro GPU provider genuinely constructs a CUDA `InferenceSession` — see "What #2617 does and does NOT prove" below. Record the resolved execution provider, not just the available-providers list.
6. `pip check` clean afterwards, for each venv you built.

## Acceptance

- Each row's criteria worked through, with observed results — A37 and A38 verdicted **separately**.
- The CUDA `InferenceSession` result, stated explicitly, with the actual provider it ran on.
- Live venv byte-verified unchanged; both readings pasted.
- Evidence file `docs/testing/onbox-wave5-results/step-ort-a-a37-a38.md`, committed.
- **No register edit** — the fold step is its sole writer.

## Not in scope

A39 and A40 — they are the next two children. Fixing anything you find — report it.

## ⚠️ Dependency — read this FIRST, before anything else

These rows are blocked by **#2600**: `install-ort.mjs` never installed the cuDNN 12 runtime `onnxruntime-gpu` needs, so `CUDAExecutionProvider` session construction silently fell back to CPU. The fix is **PR #2617** and it may or may not have merged by the time you run.

[...]

PR #2633, opened 2026-08-23
docs(docs): record the ORT marker rows' on-box acceptance

### LINKED ISSUE #2623: ops — ORT marker rows A37-A40, gated on #2600
The four ORT pip-consistency-marker rows (plan 282, #2192) — **A37, A38, A39, A40** — split out of wave 5 (#2606).

## Why its own parent, and why it WAITS

These rows are not waiting on a run. They are waiting on a **code fix**: `install-ort.mjs` never installed the cuDNN 12 runtime `onnxruntime-gpu` needs, so `CUDAExecutionProvider` session construction silently falls back to CPU (**#2600**, fix in **PR #2617**).

They were briefly inside wave 5's linear chain with a fail-open design — record "blocked", flip, close. **The repo owner has ruled against that (2026-08-23): if they are blocked, they must not close. They stay `Agent Waiting` until #2600 is done.**

That ruling is incompatible with sitting inside a linear chain — the fold and the PR behind them would deadlock — so they were lifted out into this parent. Wave 5's chain now runs #2611 (triage) → #2610 (fold) → #2609 (verify + PR) without them.

## The gate — do not promote early

**Nothing here becomes `Agent Todo` until PR #2617 is MERGED to `main`.** Promotion is a deliberate act by the operator or the coordinating session, not something a child does for itself.

Promoting early is worse than waiting, because of what pass 2 of #2617's review proved:

- The pip step installs cuDNN under `site-packages/nvidia/...`, but `preload_dlls()` searches **`torch/lib`** (basename-only) whenever torch's CUDA major matches ORT's — which is the shipped stack. Measured: **11 of 12 DLLs resolve under `torch/lib`, 0 under `nvidia/`**. The installed runtime is never read.
- `torch/lib` is precisely the diagnostic **A37's own row already records as NOT fixing the bug.**

So an ORT run against an unmerged or half-fixed #2600 **tests a path that cannot work** and would produce a confidently wrong result. The rows would look answered when nothing had been established.

## The chain, once promoted

**#2621** (A37 + A38, the shared Kokoro GPU-provider sub-check) → **#2620** (A39, marker refuses rather than repairs a clobbered venv) → **#2619** (A40, in-app upgrade on a real installed release) → **fold** → **verify + PR**.

A fresh worktree and branch will be cut off `main` **after** #2617 merges, and named in each brief at promotion time. Do not create one yourself.

## What these runs decide

#2617 has two halves and **only the first is proven**: cuDNN is *installed* (proven by dependency resolution) but *findable by onnxruntime* is not. There is also an explicit second hypothesis — `onnxruntime-gpu`'s `[cuda]` extra (`nvidia-cuda-runtime`, `cufft`, `curand`, `nvjitlink`), deliberately not installed on the reasoning that the box's system CUDA 12.4 supplies them. **That reasoning is untested.**

#2621 carries the two-branch experiment. A run that finds the provider still falling back is a **first-class result**, not a failed run.

Refs #2435, #2606, #2600

---

# ⚠️ ROW IDs CHANGED 2026-08-23 — this brief's row numbers are STALE

**PR #2626 (wave 5) merged at 05:22Z and renumbered Group A.** A row above the ORT block was retired, shifting every id below it down by one. Verified against `origin/main`:

| What this brief calls it | What it is NOW on `main` | Identity (this is what actually matters) |
|---|---|---|
| A37 | **A36** | ORT marker — fresh NVIDIA bootstrap |
| A38 | **A37** | ORT marker — the reported bug: in-app Qwen3 install |
| A39 | **A38** | ORT marker refuses — not repairs — a clobbered venv |
| A40 | **A39** | The in-app upgrade path applies the marker on a real installed release |

**`A40` now names a COMPLETELY UNRELATED ROW** — *Russian XTTS quality — leading-dash pause by ear* (#2026). Do not touch it.

## What to do about it

**Match rows by IDENTITY, never by number.** Row ids in this register are positional and renumber on every discharge; the titles above are stable. When you start, re-read the register on `main` and locate your rows by their titles, then use whatever ids they carry **at that moment** — which may have shifted again.

This is not hypothetical bookkeeping: PR #2617's own note landed on the wrong id for exactly this reason and had to be corrected, and its ten cross-references were all off by one. The merge that caused it resolved *cleanly*, because both sides held identical row bodies — nothing warned.

dudarenok-maker:
## Summary

Verification child for #2623 — the four ORT pip-consistency-marker rows (register identity: A36 fresh NVIDIA bootstrap, A37 in-app Qwen3 install, A38-was clobbered-venv refusal, A39-was in-app upgrade on a real release; renumbered by PR #2626 mid-chain, matched by title throughout). Independently re-verified every run child's claimed commands, re-ran `npm run check:onbox-register` and an independent byte-count/hash pass over the live sidecar venv myself, and re-derived the register's row and owed-debt counts by counting `###` headings rather than trusting the fold's subtraction. Everything checked out — verdict is **PASS**.

Closes #2623

## What this branch establishes about #2600

**Confirmed working on hardware, with a caveat that is itself the finding.** `install-ort.mjs`'s `nvidia-cudnn-cu12`/`nvidia-cublas-cu12` step plus `main.py`'s `_preload_ort_cuda_dlls()` genuinely fix DLL discoverability — but only once the previously-untested `[cuda]` extras hypothesis (`nvidia-cuda-runtime-cu12`, `nvidia-cufft-cu12`, `nvidia-curand-cu12`, `nvidia-nvjitlink-cu12`) is also installed; as shipped, `cufft64_11.dll`/`cudart64_12.dll` fail to load on this box's system CUDA 12.4 and the two-branch experiment in #2621 proved the `[cuda]` extras close that gap completely (all 11 expected DLLs then resolve under `nvidia/<pkg>/bin`, not a PATH/torch fallback). A raw `onnxruntime.InferenceSession` constructed directly with `providers=['CUDAExecutionProvider','CPUExecutionProvider']` then genuinely runs on CUDA — real session, real provider reported, not `get_available_providers()`.

**But Kokoro itself still silently falls back to CPU even with a clean preload**, on a newly-identified third root cause distinct from #2534 and #2600: `kokoro-onnx==0.5.0`'s own auto-detect calls `importlib.util.find_spec("onnxruntime-gpu")`, which is always `None` (the pip distribution `onnxruntime-gpu` installs into the `onnxruntime` import namespace, not a separately-importable module), so `kokoro_onnx.Kokoro()` always constructs with `providers=["CPUExecutionProvider"]` explicitly — CUDA is never even offered to onnxruntime. `main.py`'s `KOKORO_ORT_PROVIDERS` override also can't currently reach it (`TypeError`, no `providers` kwarg on this `kokoro-onnx` version's constructor). This is reported as a finding, not fixed here — three candidate fixes are named in the register's A36 note for a follow-up.

## Register — before/after and how it was counted

**66 → 65 owed**, one row discharged and removed outright (never annotated): *"ORT marker refuses — not repairs — a clobbered venv"* — `ensureOrtMarker` provably refused (returned `'clobbered'`, logged the exact remedy, touched nothing on disk — directory listing and a SHA-256 of the real plain dist-info's METADATA identical before/after the call) against a **full copy of the real in-service sidecar venv**, not a throwaway one, the first time this row has been checked that way. The other three rows (A36, A37, and the row formerly A40/now A38 — in-app upgrade path on a real installed release) are **STILL OWED**: two genuine, newly-diagnosed findings (above) and one hard blocker (no packaged `releases/vX.Y.Z` tree exists anywhere on this box — checked exhaustively, not manufactured per the issue's own instruction against cutting a release).

Counted independently rather than trusting the fold's arithmetic: `grep -c '^### A' ` inside the `## Group A` section gives **43** rows, matching the glance table's stated Group A count and the "old A39–A44 → A38–A43" renumbering note. `npm run check:onbox-register` — run by me, not just cited — passes:

```
check:onbox-register: OK — docs/testing/onbox-acceptance-register.md and docs/testing/onbox-acceptance-register-live-view.html agree.
```

## Live venv — confirmed untouched, independently

Re-verified `C:\Claude\Projects\Audiobook-Generator\server\tts-sidecar\.venv` myself (not just re-reading the run children's numbers): 56,542 files, 6,483,355,726 bytes, `python.exe` SHA-256 `0B471133E110CFB53A061CAD528CE8E517D7B9AC41A0A396C39AD795A487FC14`, `onnxruntime-1.27.0.dist-info` INSTALLER still `castwright-ort-marker` — identical to every reading pasted in the run children's evidence files.

## Still to do (manual, after merge)

The live view (`docs/testing/onbox-acceptance-register-live-view.html`) has been hand-edited to match the register and passes the mechanical cross-check, but **has not been published** — that stays a manual operator step, reusing the artifact URL already in the register's "Live view" header.

## Test plan

- [x] `npm run check:onbox-register` — run independently, green (see above)
- [x] Independent byte-count + SHA-256 re-verification of the live sidecar venv — matches every run child's reading
- [x] Independent recount of Group A register rows (43) against the glance table
- [x] `git diff origin/main --stat` — all 7 changed files trace to this work (register, live-view HTML, run sheet, 4 evidence files); no stray changes
- [x] Re-read all four run children's evidence files (#2621, #2620, #2619) and the fold (#2625) for pasted, verbatim command output rather than reasoned-about results — confirmed for the CUDA `InferenceSession` claim, the two-branch experiment, and the clobber refuse/no-repair check
- [ ] `npm run verify` — not re-run by this verify child (docs-only branch; pre-existing hooks already ran per-commit in the chain)


dudarenok-maker:
## PR review gate — exempt (docs-only file set)

No review pass run. Every changed path is under `docs/**`, which is the file-set test in [`pr-review-gate` → Exemption](../blob/main/.claude/skills/pr-review-gate/SKILL.md) (the same test CONTRIBUTING.md's doc-only CI fast-path uses). Recorded rather than left silent, so the absence of a review is a fact in the thread rather than a gap.

Verification that did run, by the chain's own verify child (#2624) and independently here:

- All four evidence files contain **pasted verbatim command output**, not reasoning about expected behaviour.
- `npm run check:onbox-register` green — register and live view agree at **65 owed**.
- The pre-publish one-directional check ran against a freshly-saved copy of the live artifact with `--discharging A44` (one Group A discharge → the group's highest ID, per the renumbering rule) — clean.
- **Live view published** to the canonical URL from the tracked `.html` on this branch, before merge, per the register's own four-step procedure.

Issue #2570, opened 2026-08-21
[agent instructions][claude][task] on-box re-run: B3, B4 after #2536 lands


dudarenok-maker:
@C:/Users/dudar/AppData/Local/Temp/open-engine-scratch/claude-2554-20260821-014123/rerun-2536.md

dudarenok-maker:
AGENT CLAIMED
Agent: claude
Timestamp: 2026-08-21T07:19:49Z

Note: this issue's own brief said do not start until PR #2562 is confirmed open
and unmerged. Checked at claim time: PR #2562 is now MERGED into main
(fix/server-2536-surname-comparator), and #2536/#2543 are both closed. The
board Status was already Agent Todo when this run's claim was issued, so the
manual promotion described in the brief has already happened; proceeding with
the acceptance re-run against latest main instead of the (now-merged) PR
branch.

dudarenok-maker:
AGENT DONE
Agent: claude

Resumed a parked claim (earlier run of this issue had already triggered the real ~20-25 min re-analysis of *Заказ Коалфолла* and it had completed at 2026-08-21T07:47:44Z, but the run ended before writing up evidence). This pass verified the completed run's output, wrote up the evidence, and shipped the register update.

**B3 — PASSED, discharged.** `mairin`/`coalfall-dragon` unchanged. No near-duplicate roster row formed, and the pre-existing duplicate pair left by the pre-fix 2026-08-20 run (`brann-weir`/`brann-wire`, `berrin-weir`/`berrin-wire`) was itself retroactively collapsed to one id apiece — live evidence #2536's fix works.

**B4 — STILL OWED.** Cyrillic-names and no-second-id criteria passed, but "every id is ASCII kebab-case" failed for a new, unrelated reason: the established id `oduvan` was retired in favour of a freshly-minted Cyrillic id `одуван` (backwards from the direction 3 other retirements took in the same run). Not a recurrence of #2536. Filed as #2584 for a fix agent.

Evidence: `docs/testing/onbox-wave4-results/step-7-b3-b4-rerun.md`
Register updated: `docs/testing/onbox-acceptance-register.md` (B3/B4 dated notes)
PR: #2585 (docs/docs-2570-b3-b4-rerun -> main), `npm run check:onbox-register` green.
New defect filed: #2584.

Issue #2313, opened 2026-08-13
srv — stage-1 returns cast names in whatever script the model picks; a Russian book got a 59% Latin-transliterated roster


dudarenok-maker:
## What happened

Stage-1 cast detection returns character `name`/`aliases` in whatever script the model feels like. On *Ночной дозор* (Russian) it returned them **transliterated into Latin** — `anton-gorodovsky "Anton Gorodovsky"`, `chief "Chief"`, `vampire-girl "Girl"` — where an earlier run of the same book, same model weights, same prompt, had returned `"Антон Городецкий"`, `"Борис Игнатьевич"`, `"Девушка-вампирша"`.

Measured across ch1-4 of the same book, two runs:

| cast lists handed to stage-2 | 2026-08-06 | 2026-08-13 |
|---|---|---|
| ids | 75 | 71 |
| **display names in Latin** | **0 / 75** | **42 / 71 (59%)** |
| ids sitting in a duplicate group | 4 (5%) | 10 (14%) |

Chapter 2's roster came back **15/15 Latin**.

Reproduced under control: replaying the *recorded* stage-1 prompt (`handoff/inbox/mns_oyK7Po6BiT-stage1-ch2.md`, byte-for-byte) 5 times returns 100% Latin names 5/5, against the recorded response's 0/3 Latin — same ids, same count, only the script differs.

## Why it happens

**Corrected 2026-08-13 — the first version of this section was wrong.** It claimed no language or script guidance reached the model at all. It does.

`languagePreamble` (`server/src/analyzer/gemini.ts:220`) is appended to the stage-1 **system instruction** whenever the book is non-English, via `castCall: StageCall = { language: bookLanguage }` → `buildSystemInstruction`. For `ru` it says:

> *IMPORTANT: the manuscript text is in Russian (Cyrillic script). Quote evidence VERBATIM from the manuscript (do not translate or transliterate it).*

…and its `castFields` clause requires `role`, `description` and each `attributes` tag to be written in the book's language.

It never mentions **`name`** or **`aliases`**. That is the whole gap, and it is narrower and more precise than "no guidance": every human-readable field except the one that identifies the character was bound to the manuscript's language.

`buildStage1ChapterInbox` (`server/src/routes/analysis.ts:1677`) adds no rule of its own; the skill file (`skills/audiobook-character-detection-per-chapter.md`) says only "`id`: kebab-case, stable, unique within the book".

[...]

PR #2640, opened 2026-08-24
fix(server): strip the fresh-side dedup collision that retired an established id (#2584)

### LINKED ISSUE #2584: retireCharacterId picks the freshly-minted id as survivor instead of the established one (oduvan -> Cyrillic одуван)
## Summary

`retireCharacterId` (`server/src/store/cast-id-history.ts`) is called as
`retireCharacterId(bookDir, from, to)` where `from` is retired and `to`
survives. During a real re-analysis (Castwright#2570, B3/B4 acceptance
re-run of *Заказ Коалфолла* after #2536's fix landed, 2026-08-21), the
merge correctly recognised a re-detected character as the same one already
in the cast under an established id, and called `retireCharacterId`, but
picked the **freshly-generated id as the survivor** instead of the
established one:

```
"oduvan": "одуван"
```

i.e. `from=oduvan` (the established, ASCII, previously-shipped id),
`to=одуван` (the fresh Cyrillic id this run's analyzer minted for
"Одуван"). The established ASCII id was discarded; the Cyrillic one is now
the character's live id in `cast.json`.

This is not the same bug as #2536 (exact-name-match fallback missing a
surname-token-drifted name) — "Одуван" has no surname drift at all, it's a
plain single-token Cyrillic name — and it is not a `safeId`/`unicodeKebab`
bug either: `unicodeKebab` deliberately preserves non-Latin letters by
design (plan 219, "Option C"), so a *brand-new* character legitimately
getting a Cyrillic id is correct. The defect is in whichever caller of
`retireCharacterId` in `server/src/store/merge-analysis-cast.ts` chose the
`from`/`to` direction for this character — in the SAME run, three other
retirements (`мэйрин→mairin`, `коалфолл→coalfall-dragon`,
`lessom→father-lessom`) correctly kept the established id as the survivor,
so the direction choice is inconsistent, not uniformly wrong.

## Why it matters

`docs/testing/onbox-acceptance-register.md` row B4 (stage-1 script-binding,
#2313) has an explicit acceptance criterion: "every id is still ASCII
kebab-case ... a roster that came back with a Cyrillic id ... is a failure
of this row even if every name is correct, because ids are the join key."
An id changing underneath an already-cast, already-possibly-voiced character
on a routine re-analysis risks orphaning anything keyed to the old id
(voice assignment, cached audio, `overrideTtsVoices`) even though nothing
about the manuscript or the character actually changed.

## Repro

1. Have a book whose cast already holds a character under an established
   ASCII id derived from a Cyrillic name (here: `oduvan` / "Одуван").
2. Re-analyse the book (any full re-analysis; not fixture-specific).
3. If the analyzer's fresh detection for that character merges via the
   name-fallback path, check `cast-id-history.json`'s `supersededBy` map for
   that character's key: direction should be `<fresh-raw-id>: <established-id>`
   (matches 3 of 4 observed cases) but was observed as
   `<established-id>: <fresh-raw-id>` for this one character.

## Scope for a fix

`server/src/store/merge-analysis-cast.ts` — the retirement direction logic
around the name-fallback match (near `:205-206,282-284`, the same
neighbourhood #2536 touched). Not filed against a specific line/condition
here — this issue is a cold hand-off, not a diagnosed fix, per the on-box
acceptance campaign's convention (see #2536 for the precedent).

## Evidence

`docs/testing/onbox-wave4-results/step-7-b3-b4-rerun.md` (Castwright#2570),
`docs/testing/cast-id-drift-onbox-acceptance.md` §7.

dudarenok-maker:
Closes #2584

## Summary

- Root cause (found on the third attempt, via PR review): a re-analysis run can mint several near-duplicate fresh candidate rows for one character in a single pass (analyzer non-determinism). `dedupeRosterByName`'s own same-run duplicate collapse arbitrarily picks one of them as its internal survivor — with no knowledge of which id the established cast already treats as stable. When the collapsed rewrite-table entry's KEY coincidentally matches an established ASCII prior cast id, and the chosen survivor (the entry's VALUE) is non-ASCII, the composed `dedupeRosterByName`→`foldMinorCast` rewrite table feeds that entry into both `remapFreshToPriorIds` (which reads it as "already converged" and skips) and `applyRewriteToPriorCast` (which applies it as a retirement) — both in the wrong direction. This reproduces the real *Заказ Коалфолла* corruption byte-for-byte (`"oduvan": "одуван"` in `cast-id-history.json`).
- `stripEstablishedAsciiRewrites` (`server/src/analyzer/roster-dedup.ts`) filters that narrow collision — established ASCII id as key, non-ASCII target as value — out of the composed rewrite table before either downstream site consumes it, at all four call sites (full-analysis + subset/per-chapter paths, in `server/src/routes/analysis.ts`). `remapFreshToPriorIds`'s pre-existing exact-name matcher is untouched and now correctly claims the case; `mergeCore` in `merge-analysis-cast.ts` is untouched.

## History of this PR

Two earlier attempts targeted the wrong function, both caught and reverted by mandatory PR review before merge:
1. A bolt-on survivor rule in `mergeCore` (commit `90032fd6`) — broke sentence-attribution cascade and let reserved fold-bucket ids become survivors. Reverted in `2bd7b6ef`.
2. A bet that `remapFreshToPriorIds`'s existing matcher alone already resolved the case (`2bd7b6ef`) — disproven by reproducing the exact real-box failure end-to-end; that matcher was never reached because of the rewrite-table collision above.

## Test plan

- New real-shape regression in `server/src/analyzer/roster-dedup.test.ts` reproduces the reviewer's exact composed-rewrite-table chain from the real box's `cast-merges.json` (`{"oduvan":"одуван","owdovan":"одуван"}`), and drives it through both `remapFreshToPriorIds` and `applyRewriteToPriorCast` with and without the fix — verified to fail without `stripEstablishedAsciiRewrites` and pass with it (the fix was temporarily short-circuited to a no-op and the test confirmed to fail for the right reason, then restored).
- Direct unit tests for `stripEstablishedAsciiRewrites` covering: the offending collision (stripped), a genuine non-ASCII→ASCII improvement (kept), an unrelated fresh-to-fresh dedup entry (kept), and both-ASCII / both-non-ASCII entries (kept).
- `npm run test -- roster-dedup remap-fresh-to-prior merge-analysis-cast analysis.test`: 339/339 passed.
- Full `npm run test:server`: 536 files, 7988 tests passed, 9 skipped.
- `npx tsc --noEmit`: clean.

## On-box acceptance

Hardware acceptance is still owed — a full re-analysis of the real *Заказ Коалфолла* book through the live analyzer/GPU pipeline, confirming the character's id comes back as ASCII `oduvan` rather than Cyrillic — recorded in `docs/testing/cast-id-drift-onbox-acceptance.md` §10.


dudarenok-maker:
## PR review — pass 1 (head 90032fd6, depth medium)

Scope: the whole branch diff (`server/src/store/merge-analysis-cast.ts` +28/−1, `server/src/store/merge-analysis-cast.test.ts` +84/−46), read against its two consumers (`server/src/routes/analysis.ts` main + subset persist blocks), the sibling id-rewrite sites (`store/remap-fresh-to-prior.ts`, `analyzer/fold-minor-cast.ts`, `store/cast-id-history.ts`), the read path that serves the result (`routes/book-state.ts`, `src/views/manuscript.tsx`), and the on-box evidence the ticket rests on (`docs/testing/onbox-wave4-results/step-7-b3-b4-rerun.md`, `docs/testing/cast-id-drift-onbox-acceptance.md` §7).

Verified before probing: CI rollup green on 90032fd6; the new `#2584` test is genuinely red before the fix and for the claimed reason (pre-fix `mergeCore` pushes `{from:'oduvan', to:'одуван'}` and keeps `merged.id = f.id`, so both of its assertions fail); the eight flipped pre-existing assertions were flipped, not relaxed — each still pins an exact id list and an exact retirement pair. I did not check out the branch or modify any tracked file.

The helper itself is correct for what it claims: `/^[a-z0-9]+(-[a-z0-9]+)*$/` accepts `oduvan` / `brann-weir` and rejects `одуван`, leading/trailing/double hyphens, uppercase and underscores. The defects below are not in the regex — they are in what the new direction does to the rest of the pipeline.

---

### 🔴 Blocking — the merge now changes the surviving cast id without remapping the sentences that same run just wrote, so a completed analysis leaves `cast.json` and `manuscript-edits.json` disagreeing

`merge-analysis-cast.ts:468` (`if (establishedIdSurvives) merged.id = old.id;`) is the **first** mechanism in this pipeline that changes a live character id without moving its sentences with it. Every other id rewrite in the run moves both halves:

- `foldMinorCast` rewrites `rewrittenSentences` alongside the roster (`analyzer/fold-minor-cast.ts:279-281`);
- `dd.rewrites` is applied to sentences at `routes/analysis.ts:1020`;
- `remapFreshToPriorIds` returns `{characters, sentences}` and the caller assigns both (`store/remap-fresh-to-prior.ts:189-198`, consumed at `routes/analysis.ts:5613-5614`).

The merge does not, and the caller does not do it either — `mergedFinal` is used only for the `cast.json` write and for `liveIds` (`routes/analysis.ts:5795, 5824, 5848, 5866`; the subset job is identical at `:7422, 7429, 7450, 7461`). Nothing between the merge and the end of the run touches `reconciled.sentences`.

**Concrete repro** (the ticket's own fixture): prior cast `[{id:'oduvan', name:'Одуван', voiceState:'tuned'}]`; the analyzer mints `{id:'одуван', name:'Одуван'}`.

1. `phase1ValidIds` is built from the pre-merge roster (`analysis.ts:5621`), so `reconcileSentenceCharacterIds` passes — every "Одуван" sentence keeps `characterId: 'одуван'`.
2. `manuscript-edits.json` is written with those sentences (`analysis.ts:5736-5738`).
3. `response.characters = characters` — the **pre-merge** roster — and `response.sentences = reconciled.sentences` (`analysis.ts:5682-5684`; subset `:7344-7346`). `mergeCore` copies matched rows (`merged = {...f}`), so `characters` still holds `одуван`.
4. `cast.json` is written from `mergedFinal.characters`, which now holds `oduvan` (`analysis.ts:5823-5826`).

Result: `GET /api/books/:bookId/state` reads `cast.json` and `manuscript-edits.json` independently and serves both verbatim, with no id resolution between them (`routes/book-state.ts:253-254`; the only `supersededBy` use in that route is the orphan **report** at `:488-518`). The frontend joins raw — `const findChar = useCallback((id) => characters.find((c) => c.id === id), [characters])`, `src/views/manuscript.tsx:471`, used at `:1879`. So every line of that character renders with no cast row; `chapterCharacters` and `lineCountById`, both derived from the sentence list (`book-state.ts:285-293`), key on `одуван`, an id that is not in the roster. The SSE `result` the confirm screen renders also disagrees with the `cast.json` the same run just wrote.

The recorded retirement `{from:'одуван', to:'oduvan'}` rescues the paths that go through `buildCastResolver` (render, QA, splice, revisions) — it does **not** rescue these two, and this is precisely the "orphaned attribution that does not exactly match a live cast id" class `docs/testing/attribution-collapse-visibility-onbox-acceptance.md` documents (50 of one book's 62 orphan spans). Before this PR the invariant "every sentence's `characterId` is a live cast id after a completed analysis" held by construction; after it, a re-analysis that hits the name-fallback breaks it deterministically.

Either the merge must return the id map so the caller can remap `reconciled.sentences` (and rebuild `response.characters` from `mergedFinal`), or the direction flip has to happen at `remapFreshToPriorIds`, which already moves both halves.

### 🔴 Blocking — `isAsciiKebabId('unknown-male')` is `true`, so the fold buckets can now capture a real character permanently

The name-fallback excludes only `NARRATOR_CHARACTER_IDS` from candidacy (`merge-analysis-cast.ts:292`, and the fresh-side mirror at `:429`). It does **not** exclude `MALE_BUCKET_ID` / `FEMALE_BUCKET_ID` (`analyzer/fold-minor-cast.ts:107-108` — `'unknown-male'` / `'unknown-female'`, both perfect ASCII kebab). Under the old direction that gap was inert: the bucket id could only ever be *retired away*, and the real character kept its own id. `:468` inverts it — the bucket id is now the survivor.

`store/remap-fresh-to-prior.ts:35-57` has exactly this guard, and says why: "Adopting a reserved id (`target`) moves a real character's sentences onto a shared slot". `analyzer/fold-minor-cast.ts:349-357` records the same hazard as a shipped invariant (plan 122: "a bucket id ALWAYS carries the canonical generic name — never a real character's name", with a normaliser to heal drift).

**Concrete repro**, using the prior-cast shape this repo documents twice as real (`remap-fresh-to-prior.ts:42-47` cites _Exile_'s cast literally holding `{id:'unknown-male', name:'Timkin'}`; `fold-minor-cast.ts:351` cites `{id:'unknown-male', name:'Lord Vane'}`):

- prior cast: `{id:'unknown-male', name:'Timkin', voiceState:'tuned', overrideTtsVoices:{qwen:{name:'q'}}}`
- fresh roster this run: `{id:'timkin', name:'Timkin', lines:20}`, and no character folds into the male bucket, so `unknown-male ∉ freshIds` and the prior row is "dropped".
- `remapFreshToPriorIds` correctly refuses (`isReservedId(priorRow.id)`, `:175`).
- `mergeCore` does not: `dropMatchCandidateByName.get('timkin')` → the bucket row; `isAsciiKebabId('unknown-male')` → true.

Before: `merged.id = 'timkin'`, retirement `unknown-male → timkin` — the plan-122 drift is *unwound*. After: `merged.id = 'unknown-male'`, retirement `timkin → unknown-male` — the drift is *cemented*, durably, and the character's real id now resolves to the shared bucket through history. The next run that folds any minor character into `unknown-male` produces a fresh bucket row that matches by id and inherits Timkin's designed voice via `PRESERVED_VOICE_FIELDS`, while Timkin comes back voiceless. `foldMinorCast`'s normaliser cannot catch this: it runs on the fresh analyzer roster, not on `priorCastForMerge`.

The fix is one line — the same `isReservedId` exclusion the sibling module already has — but it must land here.

### 🟠 Significant — the rule is much broader than the defect, and it reverses an outcome the PR's own evidence base records as correct

The shipped rule is "established wins whenever it is ASCII-kebab", i.e. it fires on every ASCII↔ASCII drift, not only on ASCII-vs-non-ASCII. That is what forced eight pre-existing assertions to flip, including `coalfall → coalfall-dragon`, which multiple docs record as a shipped, live rename (`docs/testing/attribution-collapse-visibility-onbox-acceptance.md:298-301`: "The live casts carry the renamed forms (`master-oduvan`, `coalfall-dragon`, `brann-weir`, `berrin-weir`)").

The justification for going that wide does not survive checking. #2584 says three retirements in the same run "correctly kept the established id as the survivor", implying the direction was *inconsistent*. Pre-fix `mergeCore` had exactly one direction — `retirements.push({from: cand.id, to: f.id})`, `from` = the dropped **existing** row, `to` = the **fresh** row — so the survivor was always the fresh id. Reading the run sheet (`docs/testing/onbox-wave4-results/step-7-b3-b4-rerun.md:83-88`) with that in mind:

- `мэйрин → mairin` and `коалфолл → coalfall-dragon` are dated **2026-08-20** — a different run, not "the same run";
- the four *new* entries are `brann → brann-wire`, `berrin → berrin-wire`, `lessom → father-lessom`, `oduvan → одуван`, plus `brann-weir → brann-wire` / `berrin-weir → berrin-wire`.

Every one of those has an ASCII-kebab established id on the left. There was no inconsistency to resolve — one uniform rule produced three outcomes the reporter judged fine and one they judged bad. Under this PR **all** of them flip, including `lessom → father-lessom` (the run sheet's `:106` names it as a case that "correctly kept the established id as the survivor") and `brann-weir → brann-wire` (the retroactive repair `:94` cites as live proof #2536's fix works, and the basis on which register row B3 discharged).

Narrowing to the actual defect — established survives only when it is ASCII-kebab **and** the fresh id is not — fixes `oduvan`/`одуван` and leaves every one of those cases, and eight test assertions, untouched. Choosing between the two is a behaviour decision the repo owner has a stake in (id stability vs. "the fresh, more descriptive, library-unique id wins", the rationale written at `merge-analysis-cast.ts:250-254`), so this is the one finding in this report that plausibly routes through the design-pass carve-out rather than a fix agent — but it must be decided, not shipped by default.

### 🟠 Significant — on-box acceptance is not recorded, and recording is the merge gate

This PR ships behaviour whose only real proof is a re-analysis of a real book against an existing `cast-id-history.json`. That is stated in the repo's own words: `docs/testing/cast-id-drift-onbox-acceptance.md:270-276` records that register row B2/B4 was discharged against a *fresh import*, and that this "does not re-confirm or close #2584 — #2584 is specific to a re-analysis of *Заказ Коалфолла* against its existing `cast-id-history.json` … which a fresh import has no code path to exercise."

`grep 2584 docs/testing/onbox-acceptance-register.md` → no matches, and the branch touches no register file, no run sheet, and not the live view. CLAUDE.md before-shipping step 3 is explicit that *running* the acceptance does not block a merge but *recording* it does: "a PR that ships unproven behaviour without a row … is not finishable." A row is owed here naming what to observe (re-analyse a book that already has prior history; the character's `cast.json` id is unchanged and stays ASCII; `manuscript-edits.json` still resolves to it in the manuscript view).

### 🟡 Minor — the block comment the change falsified was left in place

`merge-analysis-cast.ts:250-254` still reads: "Match a dropped existing character to a same-name fresh character so any voice fields it carries ride onto the freshly-detected row **(which carries the lines + the more descriptive, library-unique id)** and the id is retired either way." For the ASCII case that is now the opposite of what the code does — the row keeps the *established* id and the descriptive fresh id is the one retired. Same class as the citation-rot chores this repo fixes in-round; the paragraph is the first thing the next reader of this function will use to orient.

### 🟡 Minor — `isAsciiKebabId` enumerates one spelling of "stable id", so the highest-value established rows are still not protected

`cast-add-from-roster.ts:132` mints cross-book roster-link ids as `` `${target.id}_from_${targetBookId.slice(0, 8)}` `` — pure ASCII, but the underscores fail `/^[a-z0-9]+(-[a-z0-9]+)*$/`. So a row carrying a user-made cross-book link (`matchedFrom`, the most deliberately-established identity in the cast) still loses its id to a freshly-minted one on the name-fallback, which is the exact churn this PR exists to stop. No regression versus `main` — but it is the "guard that enumerates syntax" shape, and if the rule is meant to be "the established id is the stable one", the predicate is the wrong instrument for it (an "is this id already live in `existing`" test would be the direct one).

### 🟡 Minor — release-notes pair missing with no not-applicable note

The branch is two server files; neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` moved, and the PR body does not mark step 5 not-applicable. This has an operator-visible delta (character ids stop churning across re-analysis; existing ids that already churned are *not* healed — worth saying, since the reported book's live id is already `одуван` and this fix cannot move it back).

### ✅ What is solid

- The regex is right for its stated job, and the doc comment on it explains the intent rather than restating the pattern.
- The new `#2584` test is a real regression test: red before, green after, red for the claimed reason, and it pins both the survivor id and the retirement pair rather than just one of them.
- The eight flipped assertions were flipped, not weakened — every one still asserts an exact id array and an exact retirement array.
- `establishedIdSurvives` is correctly scoped per-`fresh.map` iteration and only ever set inside the fallback branch, so the by-id path and `overlayInterimCastForLiveView` (`nameFallback: false`) are untouched.
- No duplicate-id hazard inside the merge itself: `cand` is drawn only from rows absent from `freshIds` (`:292`), and the claim-once guard (`:389-397`) keeps one candidate from being consumed twice, so `merged.id = old.id` cannot collide with another output row.
- `refuseRetirementsOfLiveIds` filters on `from` only (`cast-id-history.ts:653-665`), and the new `from` (the fresh id) is never live in `mergedFinal`, so the flipped retirements are not silently dropped.

### Verdict

**Not mergeable as-is.** Two blocking correctness bugs (the sentence/cast id divergence at `merge-analysis-cast.ts:468`, and the reserved fold-bucket ids passing `isAsciiKebabId`), one significant scope question that needs the owner's decision before the rule is locked in, one owed on-box register row, and three cleanup items. The direction change is the right instinct and the helper is well-built; the problem is that flipping which side survives has consequences at three sites outside this file, and only one of them (the history record) was carried through.

dudarenok-maker:
## PR review — pass 2 (head b86f31f2, depth medium)

Scope: full `origin/main...HEAD` diff (5 files, **90 insertions, 0 deletions**) · verified: `server` typecheck clean (`tsc --noEmit`, exit 0), `remap-fresh-to-prior.test.ts` + `merge-analysis-cast.test.ts` 79/79 green · then adversarially probed the pass-1 revert's central claim — that `remapFreshToPriorIds` already resolves the reported case — against the live box's own run journals.

---

|---|---|---|
| **F1** | `mergeCore` id-survivor flip didn't cascade to persisted sentence attribution | **VERIFIED RESOLVED** — by removal, not relocation. `git diff --name-only origin/main...HEAD` lists **zero production files**; `merge-analysis-cast.ts` is byte-identical to `main`. |
| **F2** | reserved fold-bucket ids could become survivors | **VERIFIED RESOLVED** — same mechanism. New pin test added. One caveat, N3 below. |
| **F3** | rule far broader than the reported bug | **VERIFIED RESOLVED** — the revert is exact. The diff is *purely additive* (0 deletions), so the 8 un-flipped assertions are byte-identical to `main`'s original direction, not a third direction. Checked, not assumed. |
| **F5** | stale comment at `merge-analysis-cast.ts:250-254` | **MOOT** — file reverted to `main`; comment is accurate again. |
| **F6** | `isAsciiKebabId` id-format gap | **MOOT** — helper deleted with the revert. |
| **F4** | no on-box acceptance record | **STILL OPEN** — a record was added, but it records something that is not true. See N1/N2. |
| **F7** | missing release notes | **STILL OPEN** — both files were updated, but both entries announce a fix that did not ship. See N2. |

The revert itself is clean and introduced no new defects. Everything below is about what the PR now *claims*.

---

### 🔴 Blocking — N1: this PR does not fix #2584, and `remapFreshToPriorIds` provably did **not** resolve the reported case

The revert is right. The reasoning that justified it is wrong, and the reported defect is still live in `main` today.

`2bd7b6ef`'s commit message states: *"Reproduced the real #2570/#2584 fixture (prior `{id:'oduvan', name:'Одуван'}`, fresh `{id:'одуван', name:'Одуван'}`) directly against remapFreshToPriorIds: it already resolves the pair correctly and unmodified."*

**That is not the real fixture.** The live box's own journal for the 2026-08-21 run — `…\Заказ Коалфолла\.audiobook\cast-merges.json`, written at `07:47:44.937Z`, 22 ms before the corrupt history entry at `07:47:44.959Z` — records:

```json
{ "kind": "dedup", "sourceId": "oduvan",  "sourceName": "Одуван", "targetId": "одуван" },
{ "kind": "dedup", "sourceId": "owdovan", "sourceName": "Одуван", "targetId": "одуван" }
```

The fresh roster held **three** rows named `Одуван`, and one of them **re-minted the established id `oduvan`**. So `dd.rewrites` contained `oduvan → одуван`, and that table is passed to `remapFreshToPriorIds` as its 4th argument (`analysis.ts:5606-5612`, `cumulativeForRemap`).

`remap-fresh-to-prior.ts:149-152` then reads:

```ts
const target = priorIdAfter(priorRow.id);   // priorRewrites['oduvan'] === 'одуван'
if (target === freshRow.id) continue;       // 'одуван' === 'одуван'  -> SKIP
```

The remap **deliberately skips** the pair as "already converged". It never fired, and by construction it cannot fire on this shape.

**Reproduced in-process against this PR's HEAD** (no hardware, ~300 ms, reconstructed from the journal above):

```
1. dedupeRosterByName rewrites : {"oduvan":"одуван","owdovan":"одуван"}   <- matches cast-merges.json exactly
2. remapFreshToPriorIds        : {}          <- EMPTY: skipped at remap-fresh-to-prior.ts:152
   surviving fresh id          : одуван
3. applyRewriteToPriorCast     : [{"from":"oduvan","to":"одуван"}]
```

Step 3 is byte-identical to the corruption still on disk in `cast-id-history.json` (`"oduvan": "одуван"`, seq 12).

**Actual defect site** — `merge-analysis-cast.ts:508-511`, inside `applyRewriteToPriorCast`, not `mergeCore`'s name-fallback:

```ts
const canonicalId = rewrites[originalId] ?? originalId;
if (canonicalId !== originalId) {
  retirements.push({ from: originalId, to: canonicalId });   // {from:'oduvan', to:'одуван'}
}
```

Called at `analysis.ts:5788`, recorded at `analysis.ts:5865` (`recordRetirements(writeDir, retirementBookId, remapped.retirements, …)`). It blindly follows `dedupeRosterByName`'s survivor choice, which picked the Cyrillic duplicate over the re-minted ASCII one. Both of this PR's fix commits worked inside `mergeCore`, which is not where the defect is — the issue's own root-cause guess sent them to the wrong function, and the revert did not correct that, it inherited it.

**Why the new test does not catch this** — `remap-fresh-to-prior.test.ts:26-36` calls the 3-argument overload, letting `priorRewrites` default to `{}`. That defaulted argument is the *one input that causes the failure*. Consequences:

- the test is **green on `main` unmodified** (production is byte-identical), so it is a characterisation test, not a regression test — it was never red, and the house gate requires "red before the fix, green after, and red for the reason claimed";
- its inline comment — *"This module's exact-name matcher already resolves it correctly and unconditionally"* — is false. The matcher is explicitly **conditional** on `priorRewrites`, and the real run supplied the condition that disables it.

Add the 4th argument and the test fails:

```ts
remapFreshToPriorIds(
  [{ id: 'одуван', name: 'Одуван' }],
  [{ characterId: 'одуван', id: 1 }],
  [{ id: 'oduvan', name: 'Одуван', voiceState: 'tuned' }],
  { oduvan: 'одуван' },              // this run's own dedup table
);
// -> rewrites {}, characters[0].id 'одуван', sentences[0].characterId 'одуван'
```

**Required**: drop `Closes #2584` (`Refs` at most) and re-open the fix against `applyRewriteToPriorCast` / `dedupeRosterByName`'s survivor choice — or land the fix here. As it stands, merging this closes a bug that reproduces on the merge commit.

---

### 🔴 Blocking — N2: docs and both release-notes entries assert a fix that did not ship

Downstream of N1, but it ships false text to users and to the acceptance record, so it is its own gate failure.

- `RELEASE_NOTES.md:17` — *"That direction is now always correct, and the code that was letting it flip has been removed in favour of logic that was already there and always worked."* All three clauses are false: the direction is not corrected, the removed code was never the cause, and the logic relied upon does not fire on this shape.
- `docs/release-notes-next.md:127` — *"A re-analysis no longer swaps a character's id in the wrong direction."* False. The net diff contains no production change; behaviour is unchanged from `main`.
- `docs/testing/cast-id-drift-onbox-acceptance.md:894` — *"PR #2640 … fixes this by … relying on `remapFreshToPriorIds`, whose existing exact-name matcher already resolves the reported case correctly."* Falsified above.
- `…:896` and `§10.2` — the honesty this section reaches for lands on the wrong axis. It says the gap is *hardware confirmation* and that *"validating the fix requires a human or agent with real hardware access and a real analyzer."* It does not: the failing input was sitting in a committed-shape journal on the same box, and the defect reproduces in a four-argument unit call in under a second. Recording "hardware acceptance owed" instead of "not fixed" understates the state of the work — an accurate record here would have been *cheaper* than the inaccurate one.

Note the register itself (`docs/testing/onbox-acceptance-register.md`) is untouched and contains no `#2584`/`#2570` reference at all, so no row records this. If the outcome of triage is "no behavioural change shipped", that is defensible — but then the three surfaces above must say so.

---

### 🟡 Minor — N3: the new F2 pin blesses a retirement direction the sibling module documents as unsafe

`merge-analysis-cast.test.ts` (new test) asserts `retirements` toEqual `[{ from: 'unknown-male', to: 'timkin' }]` as the correct outcome — i.e. the shared fold bucket is *retired onto* one real character. `remap-fresh-to-prior.ts:34-54` refuses exactly that, in prose, for a stated reason: *"retiring one (`freshRow.id`) hands every line the reserved row owns to a single character."* Any frozen segment attributed to `unknown-male` now resolves through history to `timkin`.

Pre-existing `main` behaviour, not introduced here — but this PR is what canonises it as *desired*. Either add the `isReservedId` guard on `mergeCore`'s retirement side too, or state in the test why `mergeCore` is exempt from the rule its sibling enforces. A test that pins the opposite of a documented invariant, with no note, is how the invariant gets deleted later.

### 🟡 Minor — N4: two factual errors in `cast-id-drift-onbox-acceptance.md` §10.3

- *"Reverted the risky bolt-on from PR #2633's merge commit"* — the bolt-on came from this PR's own first commit `90032fd6`. #2633 is the unrelated ORT-acceptance docs PR.
- *"`remapFreshToPriorIds` … is exercised by both the unit test and the full integration suite (`server/src/store/merge-analysis-cast.test.ts`, existing coverage)"* — `merge-analysis-cast.test.ts` never references `remapFreshToPriorIds`; it exercises a different module. `git grep -l remapFreshToPriorIds` confirms.

### 🟡 Minor — N5: PR title and body still describe the reverted implementation

The title — *"keep established ASCII cast id as survivor over a freshly-analysed id"* — names the behaviour that was **deleted** in `2bd7b6ef`, and becomes the merge-commit subject. The `## Summary` and `## Test plan` describe `isAsciiKebabId`, `establishedIdSurvives`, `merged.id = old.id` and a mutation check over 8 flipped assertions, none of which exist in the diff. Anyone reading the merge commit later gets the opposite of what landed.

---

### Verdict

**Not mergeable.** The revert is correct and F1/F2/F3/F5/F6 are genuinely closed — but the PR now ships zero production change while claiming, in five places, to have fixed #2584, and the defect reproduces deterministically on this head. The pass-1 correctness bugs were replaced by a correctness bug in the *record*.

Smallest honest close: strip the fix claims (title, body, both release-notes entries, §10), change `Closes` to `Refs`, and keep the revert plus the two pins as what they are — a rollback and characterisation coverage. The real fix belongs in `applyRewriteToPriorCast` / the dedup survivor choice, with a regression test that passes the 4th argument.

dudarenok-maker:
## PR review — pass 3 (head 0825ab26, depth medium)

Scope: full `origin/main...HEAD` diff (8 files) · verified on HEAD: `npx tsc --noEmit` exit 0; `roster-dedup` + `remap-fresh-to-prior` + `merge-analysis-cast` 119/119 green; full `npm run test:server` 535 files / 7971 tests green (the one `Worker exited unexpectedly` is the known teardown flake). Then two mutations and an independent end-to-end reproduction through the **real** `dedupeRosterByName` → `foldMinorCast` → `composeRewrites` → strip → `remapFreshToPriorIds` → `applyRewriteToPriorCast` → `mergeAnalysisResultWithExistingCast` chain, driven from a scratch harness outside the repo. Tree verified unchanged (`0825ab26`, `git status --porcelain` empty) before and after.

---

| # | Finding | Disposition |
|---|---|---|
| **F1** | `mergeCore` flip didn't cascade to sentences | **VERIFIED RESOLVED** — `git diff origin/main...HEAD -- server/src/store/merge-analysis-cast.ts` is empty. The fix now lands upstream of sentence persistence (`analysis.ts:5615`, before `phase1ValidIds` at `:5633`), and I confirmed the cascade fires: `sentence ids ["oduvan" ×4]` post-fix vs `["одуван" ×4]` pre-fix. |
| **F2** | reserved fold-bucket ids could survive | **VERIFIED RESOLVED** — `mergeCore` untouched. The new helper cannot reintroduce it: `unknown-male`/`unknown-female` are ASCII-kebab, so an entry *keyed* on a bucket is only ever stripped when its target is non-ASCII, and `remapFreshToPriorIds` excludes reserved ids in both directions regardless (`:175`). |
| **F3** | rule far broader than the reported bug | **PARTIALLY RESOLVED** — narrowed on the script axis (ASCII→non-ASCII only), so `brann-weir → brann-wire` / `lessom → father-lessom` are untouched (both ASCII; I confirmed `merge-analysis-cast.test.ts` needed no assertion changes this round — its diff is +31/−0). But it is still too broad on a *different* axis, and that is now a live regression: see **N6**. |
| **F4** | on-box acceptance not recorded | **STILL OPEN** — see **N8**. |
| **F5** | stale block comment | **MOOT** — file reverted to `main`. |
| **F6** | `isAsciiKebabId` enumerates one spelling of "stable id" | **CARRIED, unresolved (🟡)** — the identical predicate `/^[a-z0-9]+(-[a-z0-9]+)*$/` is back at `roster-dedup.ts:53`, so a cross-book roster-link id (`routes/cast-add-from-roster.ts:132` mints `` `${id}_from_${bookId8}` ``, underscores) fails the test and gets no protection from the new strip either. No regression vs `main`; same shape as before. |
| **F7** | release-notes pair missing | **VERIFIED RESOLVED** — `RELEASE_NOTES.md` +1 entry and `docs/release-notes-next.md` +1 entry, both describing the shipped mechanism. |
| **N1** | PR did not fix #2584; `remapFreshToPriorIds` provably skipped | **VERIFIED RESOLVED for the reported shape.** Reproduced independently, not from the unit fixture: feeding the three real fresh rows (`oduvan`, `owdovan`, `одуван`, all "Одуван") through the actual `dedupeRosterByName` yields `dd.rewrites {"oduvan":"одуван","owdovan":"одуван"}` — the box's `cast-merges.json` exactly — and end-to-end **without** the strip: `remap.rewrites {}`, merged id `одуван`, `apply.retirements [{"from":"oduvan","to":"одуван"}]`; **with** it: table `{"owdovan":"одуван"}`, `remap.rewrites {"одуван":"oduvan"}`, merged id `oduvan`, voice `{qwen:{name:"q"}}` intact, `apply.retirements []`. The corruption is genuinely closed for this shape. |
| **N2** | docs asserted a fix that did not ship | **RESOLVED as to mechanism** — `§10.1`'s root-cause paragraph, `release-notes-next.md`'s entry and the `RELEASE_NOTES.md` line all now describe what the code does. One residual overclaim (`RELEASE_NOTES.md:17`, *"so the direction is always correct"*) is falsified by N6. |
| **N3** | the F2 pin blesses a retirement direction the sibling module documents as unsafe | **PARTIALLY ADDRESSED (🟡)** — the new test carries a comment, but it explains why the bucket can never be the *survivor*. N3's objection was the other half: `retirements: [{from:'unknown-male', to:'timkin'}]` hands every frozen segment the shared bucket owns to one real character through history, which `remap-fresh-to-prior.ts:49-54` refuses in prose. Still unaddressed; still pre-existing `main` behaviour, so not a blocker. |
| **N4** | two factual errors in `§10.3` | **VERIFIED RESOLVED** — the #2633 misattribution is explicitly corrected in-place, and the `merge-analysis-cast.test.ts` / `remapFreshToPriorIds` confusion is corrected to `remap-fresh-to-prior.test.ts` "(in `server/src/store/`, not `merge-analysis-cast.test.ts`)". Both re-checked against the tree. |
| **N5** | title/body described the reverted implementation | **VERIFIED RESOLVED** — title is now *"strip the fresh-side dedup collision that retired an established id"*, and the body's Summary matches the shipped diff. |

Wiring check (all four sites) is **correct**: `cumulativeForRemap` (`:5615`, `:7287`) and `cumulative` (`:5804`, `:7443`) are still computed **separately** at each of the four points, from the same `composeRewrites(dd.rewrites, folded.rewrites)` and the same `priorCastForMerge` binding (assigned once per job at `:3499`/`:3536` and `:6495`/`:6525`, never reassigned in between). Nothing was hoisted, so `§11 Q2`'s stated reason still holds, and — load-bearing — the remap's `priorIdAfter()` still predicts exactly what Site 1 will do, because both read the same filtered table.

---

### 🔴 Blocking — N6: the strip is keyed on id *shape* only, with no same-character test, so it also strips **legitimate** prior-cast moves — producing a duplicate cast row and a stranded designed voice

`server/src/analyzer/roster-dedup.ts:86`

```ts
if (priorIds.has(from) && isAsciiKebabId(from) && !isAsciiKebabId(to)) continue;
```

The doc comment two lines up claims *"Strip only that narrow shape. Anything else … is left untouched."* That is not what the predicate does. It strips **every** rewrite whose key is a prior id, whenever the key is ASCII and the target is not — regardless of whether the target is the *same character*. `dedupeRosterByName` collapses by name at four tiers, and only Tier-1 (exact same name) guarantees the target is name-equivalent to the prior row. When the target's name differs, nothing downstream can move the prior row any more, because `remapFreshToPriorIds` matches on **exact normalised name** and now has nothing to match.

**Concrete repro — Tier-3 alias coreference, this repo's own fixture.** `roster-dedup.test.ts:193` already pins `шеф ↔ Борис Игнатьевич ↔ Гесер` as a shipped strong auto-merge in which "the real name survives". That test uses ASCII ids on both sides (`boss` → `boris`), which is why the whole suite stays green. Run the *mixed-script* variant — the one #2313 records the analyzer actually producing for Russian books, and the same precondition #2584 itself rests on (this run's fresh roster re-minting the established id):

- prior cast: `{ id: 'shef', name: 'Шеф', voiceState: 'tuned', overrideTtsVoices: { qwen: { name: 'q' } } }`
- fresh roster this run: `{ id: 'shef', name: 'Шеф', aliases: ['Борис Игнатьевич'] }`, `{ id: 'борис-игнатьевич', name: 'Борис Игнатьевич', aliases: ['Шеф'] }`
- 1 sentence on `shef`, 2 on `борис-игнатьевич`

Measured through the real pipeline:

```
dd.rewrites        {"shef":"борис-игнатьевич"}      (Tier-3, real name wins — by design)

BEFORE this PR (table used = raw):
  apply.retirements  [{"from":"shef","to":"борис-игнатьевич"}]
  MERGED             ["борис-игнатьевич"]  voice {qwen:{name:"q"}}, voiceState "tuned"

AFTER this PR (table used = {}):
  apply.retirements  []
  apply.priorCast    [{"id":"shef", voice kept}]        <- prior row never moves
  MERGED             ["борис-игнатьевич", "shef"]       <- TWO rows
  MERGED voices      [{"id":"борис-игнатьевич"},        <- survivor is VOICELESS
                      {"id":"shef", v:{qwen:{name:"q"}}, s:"tuned"}]   <- 0-line orphan
  merge.retirements  []                                  <- no history record either
```

`mergeCore`'s name-fallback cannot rescue it: "Шеф" and "Борис Игнатьевич" are neither an exact nor a surname-tolerant match (`merge-analysis-cast.ts:203-214` requires a strict leading-token prefix). So the prior row falls through to the carry-forward loop at `merge-analysis-cast.ts:461-467` and is re-added verbatim. This is verbatim the hazard that function's own comment names at `:236-240` — *"a voiced dropped row would come back voiceless AND be re-added below as a 0-line orphan — a visible duplicate"*, cited there as spec §9 / "the riskiest single change in the design". The user-visible result on the Cast screen after a re-analysis: two rows, one holding all the lines with no voice, one holding the tuned Qwen voice with no lines, and nothing in `cast-id-history.json` linking them.

The same class reproduces on Tier-2a when a *second* prior row is involved (prior `oduvan` + `одуван-старый`, fresh collapsing onto `одуван-старый`): pre-fix one merged row carrying `qA`; post-fix two rows, the survivor carrying `qB` and `oduvan` stranded with `qA`.

Not every case breaks — a plain Tier-2a superset (prior `oduvan` "Одуван", fresh collapsing onto `одуван-мастер` "Одуван Мастер") reaches the identical end state before and after, because `surnameTolerantMatch` catches it at Site 3. That is luck, not design: the strip has no idea which of its victims Site 3 will pick up.

**The missing condition is a same-character test.** The entry is only safe to strip when `remapFreshToPriorIds` will then claim the pair — i.e. when the prior row and the rewrite's target are name-equivalent under `normaliseForMatch` (Tier-1's own guarantee, and exactly the `Одуван`/`Одуван` case #2584 is about). Without it the strip removes the only mechanism that was moving the prior row and puts nothing in its place.

### 🟠 Significant — N7: not one of the four production wiring sites is pinned by any test

The only production behaviour change in this PR is the four `stripEstablishedAsciiRewrites(...)` wrappers in `server/src/routes/analysis.ts` (`:5615`, `:5804`, `:7287`, `:7443`). I reverted **all four** to bare `composeRewrites(dd.rewrites, folded.rewrites)`, left the helper and every new test in place, and ran the full server battery:

```
Test Files  535 passed | 1 skipped (537)
Tests       7971 passed | 9 skipped (7997)
```

Zero failures. `analysis.test.ts`, `analysis.live-id-retire-filter.test.ts`, `analysis.dangling-superseded-target.e2e.test.ts`, `analysis.interim-prune-prohibition.e2e.test.ts` and `analysis-reject-edge-reconcile.test.ts` are all green with the fix un-wired. The new `roster-dedup.test.ts` cases exercise the helper and the two consumers *directly*, never through the route, so they hold when the route stops calling it.

The house gate asks for a regression test that is red before the fix and green after. The helper-level half of that is real — I neutered the predicate to a no-op and both new cases went red for the stated reason (`expected { oduvan: 'одуван' } to deeply equal {}`, and `expected … to not have property "oduvan"`). The route-level half does not exist, and this repo's history is specifically the history of guards that stopped being wired without anything noticing. One test that drives `runMainAnalyzerJob` (or, cheaper, a pure exported seam both routes call) over the `{oduvan, owdovan, одуван}` roster would close it.

Note also that the "with stripping" case's downstream assertions (`remap.characters[0].id`, `applied.retirements`) are never reached under mutation, because the `expect(filtered).not.toHaveProperty('oduvan')` assertion above them fails first — so they only ever run green.

### 🟠 Significant — N8 (F4 carried): on-box acceptance is recorded on one surface of three

`docs/testing/cast-id-drift-onbox-acceptance.md` gains `§10`, which is the per-feature run sheet half and is well written. The other two surfaces did not move:

- `docs/testing/onbox-acceptance-register.md` — untouched by this branch; `grep -n '2584\|2570' ` returns nothing. The at-a-glance total is still **65 owed** and Group A still ends at A43.
- `docs/testing/onbox-acceptance-register-live-view.html` — untouched; `grep -c 2584` → `0`.

`§10` itself is internally inconsistent about this: its header says *"Register row: (no dedicated row…)"* while `§10.2` closes with *"this row stays open until it runs"*. There is no row to stay open. Wave 2's own row is recorded as discharged (`onbox-acceptance-register.md:1449`), so nothing existing covers "re-analyse *Заказ Коалфолла* against its existing `cast-id-history.json` and confirm the id comes back ASCII". CLAUDE.md step 3 is explicit that running the acceptance does not block the merge but recording it does — *"a PR that ships unproven behaviour without a row … is not finishable"* — and the live-view publish + `npm run check:onbox-register -- --against-published` step is part of the same gate.

### 🟡 Minor — N9: `isAsciiKebabId`'s doc comment describes `safeId` as Latin-only, which plan 219 made false

`roster-dedup.ts:50-52`: *"Matches the pre-plan-219 legacy slug shape (`safeId`'s Latin-only output)."* `safeId` today is deliberately **not** Latin-only — `util/safe-id.ts:47-53` (`unicodeKebab`) keeps `\p{L}\p{N}` of any script, and `safeId('Одуван') === 'одуван'` is precisely why this bug exists. The parenthetical reads as a statement about `safeId` as it is, not as it was. Same citation-rot class this repo fixes in-round.

---

### Verdict

**Not mergeable.** The reported #2584 shape is genuinely fixed this time — I reproduced it end-to-end through the real dedup pipeline rather than taking the unit fixture's word for it, and the established ASCII id, its sentences and its designed voice all survive. But the guard that achieves it is written as an id-shape test rather than a same-character test, and on the Tier-3 alias merge this repo already ships and tests (`шеф ↔ Борис Игнатьевич`) it now strands a designed voice on a duplicate 0-line cast row — the mirror image of the defect it fixes, on the same class of book. That is N6, and it needs the name-equivalence condition before this lands.

Two secondary gates are also open: none of the four wiring sites is covered by any test (the full 535-file server suite passes with all four removed), and the on-box acceptance is recorded on the run sheet only, with no register row and no live-view update.

This is pass 3 of 3 — the automatic loop is exhausted, so this goes to the repo owner rather than another round.

dudarenok-maker:
Closing and reopening to re-trigger pull_request workflows — the synchronize event for the latest push (b8ab5861) never fired (confirmed via the Actions API: no pull_request-triggered run exists for that SHA after 20+ minutes, while a manually-dispatched workflow_dispatch run against the same commit passed cleanly). No code change.

PR #2636, opened 2026-08-23
fix(sidecar): build Kokoro's ORT session ourselves so it can reach the GPU

dudarenok-maker:
## Summary

**Kokoro could not reach the GPU, and no configuration could make it.** Two defects composed; either alone forces CPU.

1. **`kokoro-onnx==0.5.0`'s auto-detect can never succeed.** It decides GPU eligibility with `importlib.util.find_spec("onnxruntime-gpu")` — a pip **distribution** name passed to a function that takes an importable **module** name, and not a legal Python identifier. Both `onnxruntime` and `onnxruntime-gpu` install into the same `onnxruntime` namespace, so no arrangement of packages resolves it. `["CPUExecutionProvider"]` is then passed **explicitly** to `InferenceSession`.
2. **Our own override never applied.** `main.py` called `Kokoro(..., providers=...)`; 0.5.0's `__init__` has no such kwarg, so `KOKORO_ORT_PROVIDERS` always `TypeError`ed into a fallback whose comment claimed auto-detect would select CUDA — which defect 1 makes impossible.

The sidecar now **never lets `kokoro-onnx` choose**: it builds the ORT session itself and hands it to `Kokoro.from_session()`.

### How the provider list is chosen

`spawn-sidecar.ts:586` sets `KOKORO_ORT_PROVIDERS` **unconditionally on every spawn, for every profile**, and `ortProviders('nvidia', …)` returns `['CUDAExecutionProvider','CPUExecutionProvider']`. **So the injected list is the production path**, including on NVIDIA. `_default_ort_providers()` resolves a list from ORT's own reported state when that env is unset — the standalone-launched-sidecar case (`start.ps1`/`start.sh` set nothing), and insulation if the injection ever stops.

*(An earlier revision of this PR body asserted the opposite — that NVIDIA leaves the variable unset. That came from `_kokoro_provider_options`' docstring, which was stale; the docstring is corrected here too. Recorded rather than quietly overwritten, because the inverted claim reached five artifacts before review caught it.)*

### CPU intent wins over the provider list

`_ensure_loaded` resolves the effective requested device **first** and short-circuits to `["CPUExecutionProvider"]` before either provider path runs. Without that, a capacity-ledger CPU placement — the default configuration, when the GPU is full — produced a CUDA session on the card the ledger had just refused, **with no VRAM booked for it**, and the `KOKORO_DEVICE=cpu` knob was defeated identically.

[...]

Issue #2653, opened 2026-08-25
On-box register reuses row IDs across sections — E6 and E8 each name two different rows


dudarenok-maker:
## What

`docs/testing/onbox-acceptance-register.md` reuses row IDs across sections. `E6` and `E8` each name **two different rows**:

- `### E6 · fe-57 venv-bootstrap progress card …` and `### E8 · revoke is loopback-only …` in the owed **E** group;
- `### E6 · ops-35 ffmpeg floor …` and `### E8 · ops-36 golden-assembly on a second ffmpeg build …` under `## Blocked — hardware not available`.

Pre-existing (present on `main`), not introduced by any current branch. `npm run check:onbox-register` passes, correctly — the Blocked section is tallied separately (Blocked: 5), so the glance-table arithmetic is sound and unaffected.

## Why it matters

A bare `E6` is ambiguous. This repo cites register rows by ID from code comments, run sheets and plan docs, and `npm run check:register-citations` (shipped in #2630) resolves those citations **by ID**. An ambiguous ID is the same hazard class that checker exists to catch — a citation can name a real row and still mean the wrong one, which is precisely the `wrongId` shape, except no checker can currently detect it because both targets exist.

It also interacts badly with the positional renumbering the register already does on discharge: renumbering an owed row can silently collide with, or diverge from, a Blocked row's ID.

## The decision owed (this is why it is filed, not fixed)

More than one defensible answer, and it changes an established convention plus every existing citation:

1. **Namespace the Blocked section** — e.g. `BE6`, or `E6-blocked`. Unambiguous, but invents a second ID grammar.
2. **Renumber Blocked rows into a continuous sequence** that cannot collide with owed IDs. Keeps one grammar; renumbers rows that are cited today.
3. **Declare IDs section-scoped and make the citation checker section-aware** — no document change, but every citation must then carry its section, and `check-register-citations.mjs` needs to resolve accordingly.
4. **Leave as-is and document it** — cheapest, but keeps a latent wrong-row citation hazard permanently armed.

Whichever is chosen, `scripts/check-register-citations.mjs` likely needs a paired check so a future duplicate ID fails loudly instead of resolving to whichever row is found first.

## Key files

- `docs/testing/onbox-acceptance-register.md` (owed E group ~L3097–3534; Blocked section from ~L3931)
- `docs/testing/onbox-acceptance-register-live-view.html`
- `scripts/check-register-citations.mjs`
- `scripts/check-onbox-register.mjs`

[...]

Issue #2631, opened 2026-08-23
Kokoro always runs on CPU — provider override TypeErrors and kokoro-onnx auto-detect can never succeed


dudarenok-maker:
Found during the #2435 wave-5 ORT on-box acceptance run (#2621, chain #2623), then verified directly against the live sidecar venv (read-only).

## Kokoro never runs on the GPU — both provider paths land on CPU

Two independent defects compose. Either alone would force CPU; together they make the GPU path unreachable by any configuration.

### 1. `kokoro-onnx==0.5.0` auto-detect can never succeed (upstream bug)

`.venv/Lib/site-packages/kokoro_onnx/__init__.py:41`:

```python
providers = ["CPUExecutionProvider"]
gpu_enabled = importlib.util.find_spec("onnxruntime-gpu")   # always None
if gpu_enabled:
    providers = rt.get_available_providers()
```

`find_spec` takes an **importable module** name; `onnxruntime-gpu` is a **pip distribution** name and is not a legal Python identifier. Both `onnxruntime` and `onnxruntime-gpu` install into the same `onnxruntime` module namespace, so there is no arrangement of installed packages under which this resolves. `providers` stays `["CPUExecutionProvider"]` and is passed **explicitly** to `InferenceSession`.

### 2. Our own override TypeErrors, and the fallback comment is wrong

`server/tts-sidecar/main.py:3233`:

```python
kokoro = Kokoro(self._model_path, self._voices_path, providers=providers)
except TypeError:
    log.warning("kokoro-onnx ignored providers=%s (older release); using auto-detection.")
    kokoro = Kokoro(self._model_path, self._voices_path)
```

`Kokoro.__init__` on 0.5.0 takes `(model_path, voices_path, espeak_config, vocab_config)` — no `providers`. So `KOKORO_ORT_PROVIDERS` **always** TypeErrors and is **never** honoured, including the AMD/DirectML case it was written for.

The `except` branch is not a safe fallback. Its comment says auto-detect gives "CUDA when onnxruntime-gpu is installed, CPU otherwise" — defect 1 makes that impossible. The code degrades to guaranteed CPU while the comment says it degrades to auto-detect.

## Why this was invisible

[...]

Issue #2643, opened 2026-08-24
[agent instructions][cline-glm][task] Add inflight_synth to /health + /debug/memory (#1996)


dudarenok-maker:
## What
Add an `inflight_synth` field to the sidecar's `/health` and `/debug/memory` endpoints, reflecting the module-level `_inflight_synth` counter that already exists but is never surfaced. This gives #1996's measurement re-run a positive "nothing is generating right now" signal instead of inferring idle from a fixed wall-clock wait (which the prior measurement attempt already got fooled by — see the two most recent comments on #1996).

## Background you need
- `server/tts-sidecar/main.py:8148` declares `_inflight_synth = 0`, a plain module-level int incremented/decremented around every synth-type endpoint's `to_thread` offload (search `_inflight_synth += 1` / `_inflight_synth -= 1` — 12 sites). The comment at `main.py:8141-8146` confirms it's already read elsewhere (the drain-before-recycle path) as a plain GIL-safe read with no lock — do the same here.
- `/health`'s handler builds its response dict ending around `main.py:9702-9711` (`"recycle_pending": _recycle_pending,` through `"vram_reserved_mb_by_device": _cuda_vram_mb_per_device(),`). Add the new field in that same dict.
- `/debug/memory`'s handler is `debug_memory()`, `main.py:9736-9775`. It already builds an `engines["qwen"]` sub-dict with `prompt_cache_entries` at `main.py:9762-9767`. Add `inflight_synth` at the TOP level of the response dict (sibling to `"process"`/`"gc"`/`"engines"`), not nested under `engines["qwen"]` — it's process-wide, not Qwen-specific.

## The fix
Two one-line additions, no new imports, no new locking:
1. In `/health`'s response dict, add `"inflight_synth": _inflight_synth,` next to `"recycle_pending": _recycle_pending,`.
2. In `debug_memory()`, add `out["inflight_synth"] = _inflight_synth` right after `out["gc"] = {...}` is built (before the `engines` block).

## Where the work happens
Worktree: `C:\Claude\Projects\wt-1996-idle-vram-measurement`
Branch: `fix/sidecar-1996-idle-vram-measurement`
Do not create another worktree. Do not work in the primary checkout. Commit on this branch. Do NOT open a PR — only the final `[claude][verify]` child does that.

## Acceptance
- [ ] `/health`'s JSON response includes an `inflight_synth` integer field.
[...]

PR #2654, opened 2026-08-25
fix(sidecar): make the silent-CPU-fallback alarm able to fire at all

### LINKED ISSUE #2647: /health fell_back can never fire for Kokoro on the default config — regression from #2636
**Regression, shipped in PR #2636 (merge `89aa895b`).** Found by that PR's pass-5 review, which returned after I had already merged.

## What broke

`/health`'s `fell_back` badge — the signal that says "you asked for CUDA and got CPU" — **can no longer fire for Kokoro on the default configuration.**

`_engine_actual_card` (`server/tts-sidecar/main.py:9640`) derives the *requested* family from `_requested_device`:

```python
requested_fam, _ = _parse_device(getattr(engine, "_requested_device", None))
...
fell_back = (requested_fam == "cuda" and family == "cpu")
```

#2636's S4 fix made `_requested_device` a **pristine env-derived value** that no longer receives the admitted device. With `KOKORO_DEVICE` unset — the default — it is `"auto"`, so `requested_fam` is never `"cuda"` and `fell_back` is never `True`.

Measured by the reviewer across four revisions, driving the real `_ensure_loaded` with a `from_session` stub mirroring the installed `kokoro-onnx==0.5.0`, with an admitted `cuda:1` and a session that resolved to CPU-only providers:

| revision | `fell_back` |
|---|---|
| `55a3c5e9` (main, pre-#2636) | `True` — badge fires |
| `c4e9a44f` | `True` — badge fires |
| `3060d671` (merged) | **`False` — badge silent** |

## Why it matters

This is the **A36 `Error 126` shape**: a silent ORT CPU fallback on an NVIDIA box. Surfacing exactly that is what #2631 and the whole ORT line of work exist to do, and the badge is the user-visible signal — it reaches Settings via `server/src/routes/gpu-devices.ts` → `advanced.tsx:77` → `override-row.tsx:45`.

Reachable entirely on defaults: `SEG_CAPACITY_ADMISSION` defaults on (`main.py:4825`), the ledger returns a concrete `cuda:N`, and `/load` passes it in at `main.py:10082`.

## The shape of the fix

#2636 established three fields with three lifetimes:

- `_requested_device` — the pristine env pin, survives across loads;
- `_device` — the intent for **this** load (env pin, or the admitted device);
- `_resolved_device` — what ORT actually did.

`fell_back` should compare **this load's intent** against **the actual**, i.e. `_device` vs `_resolved_device`. It currently compares the pristine env pin against the actual, which is why an admitted `cuda:N` is invisible.

Check the same question for the other engines before changing the shared helper: Coqui and Whisper also feed `_engine_actual_card`, and the correct "requested" for them may or may not be the same field.

## Also owed, from the same review

- **The run sheet is wrong in the same place.** `docs/testing/ort-marker-onbox-acceptance.md:181-186`, rewritten in that commit, asserts the badge "fires whenever `KOKORO_DEVICE`/**an admitted pin** requested `cuda` and the session actually landed on `cpu`". That is read with GPU time booked.
- **The B3 fix has no real test coverage.** Deleting all three `_resolved_device` assignments (`main.py:3127`, `:3424`, `:3458`) leaves the suite result *identical to a verified-clean control*. Every new test is a hand-built double that supplies `_resolved_device` itself, so it pins the unchanged consumer rather than the changed producer. Root cause: of the 10 `from_session` stubs in `test_kokoro.py`, one sets `instance.sess`, and the new `_StickyPinKokoro` does not.
- **`_resolved_device = _kokoro_session_device(self) or resolved_device` fails open to the intent.** Under kokoro-onnx API drift, `_engine_actual_card` moves from `unknown` (honest) to `cuda` (a confident false claim) and skips its own reconcile.

## Acceptance

The regression itself is provable without hardware — the reviewer proved it with a stub. Confirming the badge fires against a **real** silent fallback still belongs to owed rows A36/A37.

dudarenok-maker:
## Summary

Kokoro's silent-CPU-fallback alarm could never fire on a default install. Two defects, one shipped and one latent:

**1. The alarm was structurally dead code (the #2647 regression, from #2636).** `_engine_actual_card`'s `fell_back` compared this load's outcome against `_requested_device` — written once at `__init__` from the `KOKORO_DEVICE` env pin and never touched again. On the shipped default (no `KOKORO_DEVICE` set) that field reads `"auto"` forever, even for a load the VRAM-ledger admission steered onto a concrete `cuda:N`. So `fell_back = (requested_fam == "cuda" and family == "cpu")` could never be `True` on any default install, for anyone. `requested_fam` now derives from `_device` — this load's own intent, which an admitted `device=` argument overwrites before the load runs — making the comparison load-scoped rather than config-scoped.

**2. The honest-"unknown" path was unreachable.** `KokoroEngine._resolved_device` used `"cpu"` as both its "not known yet" initial value and a genuine resolved placement. Those two meanings sharing one encoding meant a kokoro-onnx API-drift session-read failure asserted a confident, false `cpu` — and no branching at the read site could tell them apart. `None` is now the distinct sentinel; `_parse_device` already maps a falsy value to `"auto"`, which routes into the existing reconcile-or-`unknown` path.

### Ratified behaviour decision (supersedes #2631 review B3)

When the VRAM ledger deliberately admits a load onto CPU under contention while `KOKORO_DEVICE=cuda:N`, `fell_back` is **False** — admission is compliance with a capacity decision, not a silent fallback. `/health`'s `devices.kokoro` still reports the real session device either way, so an operator can always see the CPU placement; only the alarm stays quiet. One pre-existing test encoded B3's opposite reading and was flipped, with its docstring rewritten to explain what superseded it and why.

Per-engine check: Coqui, Kokoro, Qwen and Whisper all have `_device` and take the new branch (Qwen's is a verified no-op — it introspects real torch state, not this string). SPK has no separate intent value and correctly falls back to `_requested_device`.

## Test plan

- New regression test builds a **real** `KokoroEngine`, `KOKORO_DEVICE` unset, and calls `_ensure_loaded("v1", device="cuda:0")` — the real `/load` admission shape — against a session reporting CPU-only providers. Red before the fix (`assert False is True`, `requested_fam` stuck at `"auto"`), green after.
- Drift test asserts `family == "unknown"` and `fell_back is False`, built on a real engine rather than a hand-made double.
- Genuine silent-fallback detection reconfirmed unchanged.
- Full sidecar suite: 1046 passed, 2 pre-existing skips, 9 golden-tier deselected.

## On-box acceptance

Adds register row **A44** (44 A-rows, **67 owed**, counted from the rows rather than incremented), the matching hand-authored live-view entry carrying the same three observations, and both release-notes entries. No existing run sheet covers this alarm-correctness surface — `ort-marker-onbox-acceptance.md` covers the neighbouring ORT-marker/GPU-provider mechanism (A36–A38) but not this bookkeeping — so the criteria live in the row.

## Also found, not fixed

The register reuses row IDs across sections (`E6` and `E8` each name two different rows — one owed, one Blocked). Pre-existing on `main`; the arithmetic is unaffected because Blocked is tallied separately. Filed as #2653 rather than folded in: disambiguating it changes an established convention and every existing citation, so it needs a decision, not an implementation.

Closes #2647


dudarenok-maker:
## PR review — pass 1 (head 9ea59680, depth medium)

Scope: `server/tts-sidecar/main.py`, `tests/test_kokoro.py`, `tests/test_devices.py`, `docs/testing/onbox-acceptance-register.md`, `…-live-view.html`, both release-notes files · verified: 79/79 in the two touched test files, then a 299-test targeted device/placement/admission set (`test_devices`, `test_kokoro`, `test_coqui_device`, `test_coqui_idle_evict`, `test_coqui_publish_race`, `test_device_probe`, `test_placement`, `test_runtime_wiring`, `test_smoke`, `test_synthesize`, `test_load_admission`, `test_capacity`, `test_transcribe_embed_admission`, `test_asr_spk_idle_evict`, `test_speaker_embed`, `test_qwen_device`, `test_device_parse`) — all green; `npm run check:onbox-register` OK; CI on this SHA is 16/16 `conclusion: SUCCESS` (run `32797452510`, `headSha` confirmed `9ea5968068d8…`), no cancellations — before adversarially probing the `None` sentinel's blast radius, the per-engine branch selection, fixture reachability, and the register/live-view arithmetic.

All source-side probing was done against `git archive`-exported copies in the OS temp dir. **The worktree was not modified** (`9ea59680`, `git status --porcelain` empty, before and after).

---

### 🔴 Blocking — `ort-marker-onbox-acceptance.md` now states the OPPOSITE of the shipped behaviour, and #2647 listed that exact sentence as owed

**Correctness bug (in the instrument).** `docs/testing/ort-marker-onbox-acceptance.md:189-195` — the only `cpu_fallback` sentence in any run sheet — reads:

> Since #2631's B3 fix, also check `gpus[].resident[]` for the Kokoro entry's `stale_reason: 'cpu_fallback'` — it fires whenever `KOKORO_DEVICE`/**an admitted pin** requested `cuda` and the session actually landed on `cpu`, **which includes the by-design contention case above**, so it corroborates a CPU session **rather than distinguishing contention from a genuine bug on its own** — bullets 1 and 3 still make that call.

After this PR both emphasised clauses are false, and inverted:

- The "by-design contention case above" is a ledger-admitted CPU placement. That is precisely the case the ratified decision now makes `fell_back = False` — the badge specifically **excludes** it. The sheet says it includes it.
- Consequently the badge now *does* distinguish: present ⇒ this load's own intent was cuda and it silently landed on cpu (the genuine #2534/#2600/#2621 bug); absent ⇒ not that. An operator following this sheet reads a firing badge as merely "corroborates a CPU session, could still be contention" and goes off to re-run with an idle card — the opposite of the call the new semantics support.

This is not incidental: **issue #2647's own "Also owed, from the same review" list names it** — *"The run sheet is wrong in the same place. `docs/testing/ort-marker-onbox-acceptance.md:181-186` … asserts the badge 'fires whenever `KOKORO_DEVICE`/an admitted pin requested `cuda` and the session actually landed on `cpu`'. That is read with GPU time booked."* The PR body says `Closes #2647` while leaving it untouched, and affirmatively asserts the opposite: *"No existing run sheet covers this alarm-correctness surface — `ort-marker-onbox-acceptance.md` covers the neighbouring ORT-marker/GPU-provider mechanism (A36–A38) but not this bookkeeping."* It does cover it, in one sentence, which this change falsifies.

`A36` (register:2105) and `A37` (register:2273) are both still owed, so this sheet is live and will be read with GPU time booked. Fix the sentence in this PR, or downgrade to `Refs #2647`.

---

### 🟠 Significant — the Qwen branch is **not** a no-op: `fell_back` flips `False → True` for Qwen, untested and user-visible

**Correctness bug.** `main.py:9681-9685` asserts, and the PR body repeats ("Qwen's is a verified no-op — it introspects real torch state, not this string"):

> Qwen has `_device` too, so it takes this same branch, but it's a no-op there: Qwen loads via a plain torch `.to(self._device)` … and the top-of-function torch introspection above already reads Qwen's real device straight from the model, not from this string.

That reasoning conflates `family` (which does come from torch) with `requested_fam` (which now comes from the string). **`QwenEngine` has no `_requested_device` at all** — the only writes in `main.py` are `:1677` (Coqui), `:3118` (Kokoro), `:7521` (Whisper), `:7861` (SPK). So pre-change `requested_fam` for Qwen was `_parse_device(None) == "auto"` — hard-`False` forever; post-change it is `_parse_device(qwen._device)`, a concrete resolved card.

Measured on Qwen's exact attribute shape (`_base` with `parameters()` yielding a cpu param, `_device="cuda:0"`, no `_requested_device`), `origin/main` vs this head:

```
PRE  (origin/main): {'family': 'cpu', 'index': None, 'fell_back': False}
POST (9ea59680):    {'family': 'cpu', 'index': None, 'fell_back': True}
```

User-visible: `_resident_engines_by_card` (`main.py:9748-9756`) turns that into `stale_reason: 'cpu_fallback'`, which `server/src/routes/gpu-devices.ts` merges and `src/views/advanced.tsx:82-89` (`deriveStaleReason`; `engineForDeviceKnob` maps `tts.qwen.device` → `qwen`) renders as a stale badge on the **Qwen** device knob in Advanced Configuration. This is reachable: `_load_qwen_model` resolves a separate `codec_device` (`main.py:5514`) and calls `_move_codec_to_device`, so a Qwen whose first-yielded parameter sits off `self._device` is a state the loader itself creates.

The change may well be an improvement — but it is an unreviewed, unpinned behaviour change on a second engine, shipped under a "verified no-op" label. No test in this diff covers Qwen.

---

### 🟠 Significant — Whisper's `fell_back` now compares `_device` against itself, and `WhisperEngine._requested_device` is orphaned

**Correctness bug.** `WhisperEngine` (`main.py:7478`) has `_device` but no `device` and no `_resolved_device`, and `faster_whisper.WhisperModel` is a plain class with no `parameters()`. So after this change both sides of the comparison read the same attribute:

- `requested_fam` ← `main.py:9686-9687`: `hasattr(engine, "_device")` is True → `_parse_device(engine._device)`
- `family` ← `main.py:9716-9721`: `hasattr(engine, "_resolved_device")` is False → `getattr(engine, "device", None) or getattr(engine, "_device", None)` → `engine._device`

`fell_back` is therefore provably always `False` for ASR. Measured on Whisper's exact shape (`_requested_device="cuda:0"`, `_device="cpu"`, `_model` without `parameters()`):

```
PRE  (origin/main): {'family': 'cpu', 'index': None, 'fell_back': True}
POST (9ea59680):    {'family': 'cpu', 'index': None, 'fell_back': False}
```

Two consequences:

1. `main.py:7521` — `self._requested_device = self._device  # preserved for /health fell_back detection` is now **dead**: nothing reads it (its only reader was the `_requested_device` arm Whisper no longer takes), and the comment claiming otherwise is false. CLAUDE.md's "remove what YOUR changes orphaned" applies.
2. `main.py:9675-9681` justifies Whisper under the *else* arm — *"SPK's and Whisper's own admission call sites (`/embed`, `/transcribe`) only ever admit a GPU placement when the env pin ALREADY says cuda/rocm, so `_requested_device` already reflects this load's real intent **for them** regardless"* — but Whisper never reaches `_requested_device` under the new code. A future maintainer reading this comment gets the wrong branch. (The claim is correct for SPK, which I confirmed still fires: `main.py:7902`/`:7914` genuinely self-demote `self.device`, and the SPK shape returns `fell_back: True` on both revisions.)

Applying this PR's own standard — a comparison that can never fire is dead code, not a working detector — ASR's detector is now dead by construction. Either give Whisper a real `_resolved_device` split, or make the degeneracy explicit and remove the orphan.

---

### 🟠 Significant — under kokoro-onnx drift, a GPU-resident Kokoro becomes invisible to `PlacementController`

**Correctness bug.** `_is_resident` (`main.py:4870-4886`) derives residency from `_engine_actual_card(...)["family"]`. Before this PR a drifted session read left `_resolved_device` holding the intent, so a Kokoro loaded on `cuda:0` still reported residency; now it reports `None`:

```
PRE  (origin/main): card={'family': 'cuda',    'index': None, ...}  _is_resident -> cuda:0
POST (9ea59680):    card={'family': 'unknown', 'index': None, ...}  _is_resident -> None
```

(state: `_kokoro.sess.get_providers()` raises — the exact shape the new drift test builds; `_device="cuda:0"`; `_resolved_device` `"cuda:0"` pre / `None` post.)

`PlacementController.reservation()`/`admit()` (`main.py:4657`, `:4464`) use that value for two things: `constraint = resident if resident is not None else pinned`, and `footprints.peak_mb(engine, model, cfg, resident is not None)`. So under drift a subsequent Kokoro reservation loses the "can only run on its own card" constraint and books the **cold-load** peak instead of the incremental one, while Kokoro's real VRAM on `cuda:0` goes unaccounted. The honesty fix is right for `/health`; `_is_resident` is a different consumer and should keep falling back to `_device` when the ground truth is `unknown`.

---

### 🟠 Significant — A44 bullet 3's repro cannot produce the observation it asks for

**Correctness bug (acceptance criterion blind to its own feature).** `onbox-acceptance-register.md:2614-2616` (and the identical live-view entry):

> Force the kokoro-onnx API-drift branch (**e.g. a `kokoro-onnx` install with no `Kokoro.from_session`**) and confirm `/health` reports the Kokoro card as `unknown`, not `cpu`, and `fell_back` reads `False`.

That branch does not reach the unknown path. `main.py:3366-3375` falls back to `Kokoro(self._model_path, self._voices_path)`, and the installed `kokoro_onnx/__init__.py`'s `Kokoro.__init__` unconditionally sets `self.sess = rt.InferenceSession(model_path, providers=providers)`. `_kokoro_session_device` (`main.py:9633-9646`) therefore reads a perfectly good session, gets `["CPUExecutionProvider"]`, and `_resolved_device` becomes `"cpu"` — so `/health` reports **`cpu`, not `unknown`**, and an operator following the bullet records a FAIL against correct code.

The drift the fix actually targets is `sess` absent or `get_providers()` raising — which is what the unit test simulates (`_DriftedSession.get_providers` raising `AttributeError`), and which the named repro is not. Name a repro that produces it, or say plainly that this branch is not on-box reproducible today.

---

### 🟠 Significant — A44 bullet 2 has no positive control, so it passes on a null observation

**Correctness bug (same shape).** `onbox-acceptance-register.md:2607-2613` asks the operator to exhaust headroom, load Kokoro, and *"confirm the resident entry for Kokoro carries **no** `stale_reason: 'cpu_fallback'`"*. Nothing in the bullet confirms the admission actually returned CPU. If headroom is not genuinely exhausted, `admit()` returns `cuda:N`, Kokoro loads on the GPU, and the absence of `stale_reason` is recorded as PASS while the ratified behaviour under test — admitted-CPU stays quiet — was never exercised at all.

Bullet 1 already carries the right control (`devices.kokoro` reads `cpu`); bullet 2 needs the same one, plus a check that the Kokoro entry is *present*.

---

### 🟠 Significant — the alarm is still structurally dead on the lazy-warm and `PRELOAD_KOKORO` paths, which is how Kokoro loads during a render

**Correctness bug.** Not a challenge to the ratified `auto`-intent rule — the issue is that on the shipped default, `auto` is the *only* intent the generation path can ever produce, so the badge still cannot fire where it matters most.

`_device` only becomes concrete when a caller threads `device=`. Exactly one call site does — `/load`'s Kokoro branch, `main.py:10153`. The other three do not:

- `KokoroEngine.synthesize` → `self._ensure_loaded(model)`, `main.py:3485` — no `device=`;
- `PRELOAD_KOKORO=1` startup → `await asyncio.to_thread(kokoro._ensure_loaded, "v1")`, `main.py:9224` — no `device=`;
- `/load` with `SEG_CAPACITY_ADMISSION=0` → `main.py:10155` — no `device=`.

And generation never POSTs `/load`: `server/src/tts/ensure-sidecar-loaded.ts:7-17` — *"this gate POLLS `GET /health` — it NEVER `/load`s a model … Model loading is now purely lazy."* main.py's own startup log agrees: *"PRELOAD_KOKORO not enabled — Kokoro warms on demand on first synth"* (`main.py:9233`).

So with `KOKORO_DEVICE` unset (the default), a book render warms Kokoro through `synthesize` → `resolved_device = self._device = "auto"` → `requested_fam == "auto"` → `fell_back` hard-`False` even on a genuine silent CUDA→CPU fallback. That is the A36 `Error 126` shape the badge exists for, on the shipped default, still unreported.

Three shipped documents state the opposite without qualification and need narrowing at minimum:

- PR body: *"So `fell_back` could never be `True` on any default install, for anyone"* — presented as fixed;
- `RELEASE_NOTES.md:15`: *"so a genuine silent fallback is finally caught"*;
- `docs/release-notes-next.md:233` and register row A44's preamble, same claim.

The documentation half is unambiguously fixable now. Whether the lazy path should resolve an intent (as `/load` does) is arguably a design call — if so, say which decision is owed rather than shipping the unqualified claim.

---

### 🟡 Minor

- `tests/test_devices.py:178-193` — the two loaded-Coqui fixtures set `_resolved_device="cpu", _device="cuda:1"`, a state a loaded Coqui cannot be in: `_publish_loaded_locked` (`main.py:2017-2020`) writes **both** fields from the same `device`, and the only other writers run with `_tts is None`. Pre-existing, but this PR makes `_device` the load-bearing field, so `test_engine_actual_card_detects_loaded_coqui_via_tts` is now the only thing pinning Coqui's `fell_back` and it pins an unreachable state.
- A44 bullet 1 (`register.md:2601-2606`) says "confirm `/health`'s `gpus[].resident[]` entry for Kokoro". A cpu-fallen Kokoro carries `index=None`, so it lands in the synthetic `idx: -1` "unindexed (cpu / ORT / CT2)" entry (`main.py:9776-9792`), never on a real card's `resident[]`. Worth saying, or the operator looks at `gpus[0]` and records a false negative.
- The `else` arm of the `_resolved_device` split (`main.py:9718-9721`) — the `_device` fallback, i.e. Whisper's only route to `family` — has no test. `test_devices.py:29` covers the `device` (SPK) fallback only.

---

### ✅ What is solid

- **Both source changes are load-bearing and red for the reason claimed.** Verified on isolated `git archive` copies by reverting each hunk independently: (A) `requested_fam` → `_requested_device` ⇒ `test_kokoro_default_config_admitted_cuda_landing_on_cpu_flags_fell_back` fails `assert False is True` and the flipped `..._admitted_cpu_is_not_fell_back` fails `assert True is False`; (B) full `None`-sentinel revert ⇒ drift test fails `assert 'cuda:0' is None`; (C) the `hasattr` split alone ⇒ drift test fails `assert 'cuda' == 'unknown'`. No mutation passed silently.
- **The new Kokoro fixtures are states production reaches.** All three build a real `main.KokoroEngine()` and drive the real `_ensure_loaded` — including the `device="cuda:0"` admission shape `/load` actually uses — rather than a `SimpleNamespace` that could omit an attribute `__init__` always sets. That is exactly the defect #2647 called out in the prior round, and it is closed.
- **No unreachable branch survives.** Both arms of both `hasattr` splits are exercised by the real engine set (`_device`: Coqui/Kokoro/Qwen/Whisper vs SPK; `_resolved_device`: Coqui/Kokoro vs SPK/Whisper).
- **The two `/health` Coqui readers were correctly left alone.** `main.py:9818` and `:9862` are both `isinstance(coqui, CoquiEngine)`-guarded and Coqui's sentinel is unchanged (`"cpu"` at `:1687`/`:2106`), so Kokoro's `None` can never reach them. Whisper/SPK/Qwen have no `_resolved_device` at all — correctly untouched.
- **Genuine detection still fires where it should.** Env-pinned `KOKORO_DEVICE=cuda:1` + CPU-only ORT session ⇒ `fell_back True` (`test_engine_actual_card_kokoro_cuda_to_cpu_provider_drop_flags_fell_back`); admitted `cuda:0` + CPU-only session ⇒ `True` (the new test); SPK self-demotion (`main.py:7902`/`:7914`) ⇒ `True`, unchanged on both revisions.
- **The `"auto"` intent decision is implemented, commented and pinned** — `main.py:9728-9738` plus `test_engine_actual_card_kokoro_auto_intent_cpu_result_is_not_fell_back`, built on a real engine, with the comment naming the test.
- **The B3 supersession is documented where a reader meets it**: the renamed test's own body (`test_devices.py:116-133`), the fixture docstring (`:78-86`), the commit message, the PR body, the register row and the live view all carry it. I found no other pre-existing test whose meaning changed silently — the two surviving Kokoro fixtures set `_device == _requested_device`, so they compute identically on both revisions.
- **Register arithmetic checks out by counting, not by trusting.** Row headings: A=44, B=2, C=4, D=3, E=12, G=2, H=2. The two surplus E headings are the pre-existing duplicated `E6`/`E8` inside `## Blocked` (filed as #2653, correctly out of scope here), leaving E=10 owed ⇒ 44+2+4+3+10+2+2 = **67**, matching the stated total and every glance-table cell. Blocked = 5 headings, Unconfirmed = 2 bullets. `npm run check:onbox-register` passes.
- **The live view does not ask for less than the register.** All three observations are carried verbatim, plus the same prerequisites/criteria/cost line; the strip (67/5/2) and the group-A count (44 in both the glance row and the section header) all move together. No repeat of the E10 drift from #2635.
- **House gates**: release-notes pair present (`docs/release-notes-next.md` + `RELEASE_NOTES.md`, user-facing line in the in-progress section); `Closes #2647` plain, outside any code span; no config knob, no `cast.json` write, no derived artifact touched; incidental finding (#2653) declared in the PR body with the decision it owes.

### Verdict

**Request changes.** One blocking finding (the run sheet this change falsifies, which #2647 itself listed as owed) and six significant ones, all correctness bugs — so a re-review is triggered once they are fixed and pushed. The core Kokoro fix is right and, unusually for this area, genuinely well pinned: the two new regression tests drive a real `KokoroEngine` through the real load path, and every mutation I applied went red for the stated reason. What is not right is the blast radius — Whisper's detector is now degenerate, Qwen's is newly live under a "no-op" label, `_is_resident` lost a signal, the alarm is still dead on the path a render actually uses, and three of the acceptance/documentation surfaces around it now say things the code does not do.

dudarenok-maker:
## PR review — pass 2 (head 9da415b7, depth medium)

Scope: `server/tts-sidecar/main.py`, `tests/test_kokoro.py`, `tests/test_devices.py`, the three doc surfaces, both release-notes files. Verified locally before probing: `test_kokoro.py` 52 passed; `test_devices.py` + `test_transcribe_embed_admission.py` + `test_placement.py` + `test_device_ledger.py` + `test_load_admission.py` 116 passed; `test_capacity.py` + `test_per_card_ceilings.py` + `test_design_kokoro_exclusion.py` + `test_mint_variant_kokoro_stall.py` + `test_device_probe.py` + `test_device_parse.py` 68 passed. Red-before-green re-derived by reverting each load-bearing hunk in an out-of-tree copy of the sidecar (no tracked file touched). CI on `9da415b7`: **all 16 check-runs `completed` / `success`**, read from `conclusion` values via `gh api .../check-runs`, not a `--watch` exit code — including `Sidecar tests (pytest)` (1m7s).

---

**🔴 F1 — ORT-marker run sheet contradicted the ratified decision — VERIFIED RESOLVED.**
`docs/testing/ort-marker-onbox-acceptance.md:190-202` now reads the ratified way round: "*a VRAM-ledger admission onto `cpu` under contention … publishes `cpu` as that load's own `_device` before the load runs, so `stale_reason` correctly stays ABSENT for it*", and "*its presence DOES distinguish the two*". The old "*includes the by-design contention case above, so it corroborates … rather than distinguishing*" is gone. A36's bullets 1 and 3 are unchanged and still carry the contention-vs-bug call; A37 (§4) is untouched and unaffected. No residue.

**🟠 F2 — claimed Qwen `False → True` flip — WITHDRAWN. Pass 1 was wrong.**
Re-derived independently against the installed package, not a double:

- `qwen_tts/inference/qwen3_tts_model.py:54` — `class Qwen3TTSModel` defines `model`, `processor`, `generate_defaults`, `device` and its generate/validate methods. **No `parameters`.** (Its `__init__` *calls* `model.parameters()` on the inner HF model at line 78, but does not expose one.)
- `QwenEngine` (main.py:5258+) has `_device` and `_device_pref`; it has **no** `device`, no `_resolved_device`, no `_requested_device`. `model.device = torch.device(self._device)` at main.py:5630 sets that on the *wrapper*, which `_engine_actual_card` never reads.

So at main.py:9766 `getattr(model, "parameters", None)` is `None`, the torch branch never runs, `hasattr(engine, "_resolved_device")` is False, and `family` resolves from `getattr(engine, "device", None) or getattr(engine, "_device", None)` → the **same** `_device` string the new `requested_fam` branch reads. `fell_back` is structurally False. Before the change `requested_fam` came from an absent `_requested_device` → `_parse_device(None)` → `"auto"` → also never `"cuda"`. **False both sides; there is no flip.** Pass 1's measurement can only have come from a double carrying `.parameters()` — the shape production cannot produce, and which this PR correctly labels a hypothetical control (`test_devices.py:232`, "*Today's no-op is a property of the wrapper's shape*"). Nothing to fix.

**🟠 F3 — degenerate Whisper detector — VERIFIED RESOLVED** (one residual, N2 below).
Reverting the carve-out at main.py:9757 (`if isinstance(engine, WhisperEngine):` → `if False:`) reddens `test_devices.py:252` with `assert False is True` — i.e. `fell_back` collapses to tautologically-False exactly as claimed. The comparison is real and the assertion is load-bearing. The `_resolved_device`/N10 comment at main.py:9774-9777 now accurately describes Whisper as having no `_resolved_device` split, and the `hasattr` split at 9789-9793 keeps Whisper on the pref fallback. Correct.

**🟠 F4 — `_is_resident` dropped a drifted-but-resident Kokoro — VERIFIED RESOLVED** (one residual, N1 below).
Reverting main.py:4933-4936 back to `if card is None or card["family"] not in ("cuda","rocm"): return None` reddens `test_devices.py:283` with `assert None == 'cuda:0'`. Fallback is load-bearing. It is correctly gated on `card is not None` (loaded) **and** `family == "unknown"` **and** `intent_fam in ("cuda","rocm")`, with `test_devices.py:294` as the negative control for an `auto`/cpu intent.

**🟠 F5/F6 — A44 bullets 2 and 3 — VERIFIED RESOLVED.**
Bullet 3 now explicitly *rejects* the old repro by name and says why — "*Not 'a `kokoro-onnx` install with no `Kokoro.from_session`' — verified against the installed package: `Kokoro.__init__` … ALSO always sets `self.sess = rt.InferenceSession(...)` … that repro can't produce the observation this bullet asks for, and an operator running it would log a false PASS*" — and substitutes a repro that actually forces the branch (`_kokoro_session_device` returning `None`, i.e. the `.sess` attribute being unreachable). That matches the code: main.py:9690-9694 returns `None` only when `.sess` is absent or `get_providers()` raises. Bullet 2 gained a real positive control ("*repeat with the card genuinely idle so `admit()` places Kokoro on the GPU instead*") with the null-observation failure mode spelled out. Both correct.

**🟡 F7 — "unreachable Coqui test fixture" — WITHDRAWN.**
The only Coqui fixtures reaching `_engine_actual_card` are `test_devices.py:176` and `test_devices.py:189`. Both are `types.SimpleNamespace` doubles, but the states they encode (`_device="cuda:1"` + `_resolved_device="cpu"`, and both `"cuda:1"`) are reachable for a real `CoquiEngine` — `_publish_loaded_locked` writes `_resolved_device` separately from the `_device` pref (main.py:2017). There is no unreachable Coqui fixture in `test_devices.py`, `test_kokoro.py`, or `test_coqui_device.py`. Finding withdrawn; nothing owed.

---

### New work (#2643) — probe results

**1. Tautology check — CONFIRMED independent. Not tautological.** Intent and actual come from genuinely different sources:

- intent: `providers`, which is either the env-injected `KOKORO_ORT_PROVIDERS` (`_resolve_ort_providers`, main.py:3137) or `_default_ort_providers` (main.py:3158) reading `rt.get_available_providers()` + `rt.cuda_version` — ORT **build/availability** metadata;
- actual: `_kokoro_session_device` (main.py:9686) reading `sess.get_providers()` on the **constructed** `InferenceSession` — which EPs ORT actually registered.

"CUDA EP is in the build" and "CUDA EP initialised for this session" are different facts, and their divergence *is* the silent-fallback shape. Unlike Whisper's pre-fix state, nothing routes both sides to one value.

**2. Placement unchanged — CONFIRMED.** `intent_device` appears at exactly two sites: its computation (main.py:3361-3363) and the publish (main.py:3471). It feeds nothing else — not `po`, not `build_providers`, not `session_kwargs`. Per case: env-pinned and admitted `device=` both take `family != "auto"`, so `intent_device == resolved_device` and the published string is byte-identical to before; auto-with-CUDA and auto-without-CUDA change only the published string, after every placement decision has already been made.

**3. Ratified compliance case — CONFIRMED holds.** `_ensure_loaded(device="cpu")` under `KOKORO_DEVICE=cuda:1` → `resolved_device="cpu"` → `family=="cpu"`, not `"auto"` → `intent_device="cpu"` → `requested_fam=="cpu"` → `fell_back` False. Pinned on a real engine at `test_kokoro.py` (`..._admitted_cpu_overrides_cuda_pin_not_fell_back`) and, at the card level, by the flipped `test_devices.py:115`.

**4. `build_providers` vs `providers` — CONFIRMED no divergence.** `_kokoro_provider_options` (main.py:5058) returns `None` unless `family == "cuda"` **and** an index is present. On the `family == "auto"` branch — the *only* branch where `intent_device` is derived from `providers` — it therefore always returns `None`, so `build_providers is providers`. The one path where `build_providers` differs (the synthesised CUDA list for an empty `providers`) is reachable only under an indexed `cuda:N` pin, where `intent_device = resolved_device` and is not derived from `providers` at all. No path lets the intent describe a list the session did not use.

**5. Fixture validity + revert-and-redden — CONFIRMED for all three fixes.** All six new `test_kokoro.py` cases build a real `main.KokoroEngine()` and drive the real `_ensure_loaded`; the states are production-reachable (ORT registering CPU-only despite a CUDA request; `.sess` unreadable under drift). Reverting `self._device = intent_device` → `resolved_device` (main.py:3471) reddens exactly `test_kokoro.py:601` and `test_kokoro.py:675`, with `assert 'auto' == 'cpu'` / `assert 'auto' == 'cuda'` — the claimed defect verbatim. Crucially, with the `_device` string assertion disabled, `test_kokoro.py:675` **still fails at `assert False is True`** — so the alarm assertion is load-bearing in its own right, not a string canary riding on it. (Note for the record: `test_kokoro.py:601`'s only discriminating assertion *is* its `_device == "cpu"` line — its `fell_back is False` holds pre- and post-fix. That is correct for a not-fell-back control; flagging it only so nobody later "simplifies" that line away.)

**6. Docs — CONFIRMED.** Register rows counted, not trusted: 69 `### <L><N> ·` headings, minus the 2 duplicate-label headings inside `## Blocked` (`E8`, `E6`) = **67 owed**, and `A` = **44** — both matching the stated totals, and the glance table's groups sum to 67. The live-view A44 entry is the same two paragraphs and same four bullets as the register row, including the positive control and the "*would log a false PASS*" warning — **no weakening**; the only delta is the word "actually" dropped from the display title. Both release-notes files gained matching entries (`#2647` and `#2643`), technical in `release-notes-next.md`, brand-voice in `RELEASE_NOTES.md`.

**7. Suite scoping — defensible.** The local skip costs nothing here: cloud CI ran `Sidecar tests (pytest)` to green on this exact SHA. I separately ran every file that consumes the changed seams (`_engine_actual_card`, `_is_resident`, `_ensure_loaded`) — 236 tests, all green. Nothing plausibly affected went unexercised.

---

### 🟡 Minor — cleanup only (no correctness bug; does not re-trigger a review round)

**N1 — `_is_resident`'s docstring is now false about what `_device` carries, and the fallback can book a card nothing is on.** `main.py:4917-4919` justifies the fallback as "*an admitted placement or an explicit env pin is the best remaining evidence of where the weights actually live*". After #2643 that is no longer what `_device` holds on the default path: an unpinned, unadmitted auto load now publishes `"cuda"` purely because `CUDAExecutionProvider` appeared in the *requested* provider list (main.py:3363). Concrete scenario: `KOKORO_DEVICE` unset, CUDA EP offered but the session silently registers CPU-only, **and** the session read drifts → `family == "unknown"` → main.py:4934 reads `_device == "cuda"` → `_is_resident` returns `"cuda:0"` for a model living in system RAM, and `PlacementController` books ~1 GB of VRAM that is actually free (enough to tip a Qwen VoiceDesign admission on an 8 GB card). Pre-#2643 that same load left `_device` at `"auto"`, so the fallback declined. The behaviour errs conservative (over-book, never under-book), which is the safe direction — but the docstring should say the fallback's evidence may be a provider-availability inference rather than a pin or an admission.

**N2 — Whisper's re-armed `fell_back` has no production path to divergence; the carve-out is correct but dormant.** The fix is right and I want it kept. But I could not find a reachable state where `WhisperEngine._requested_device` is `cuda` while `_device` is `cpu`: `_device` is written only at `__init__` from `ASR_DEVICE` and at the end of `_ensure_loaded` as `self._device = dev`; there is no cuda→cpu demotion anywhere in the class (unlike SPK); and `/transcribe`'s admission (main.py:11309-11322) passes `cpu_capable=False`, so `PlacementController` returns a GPU key or `noCapacity` — `{"device": "cpu"}` is returned only under `cpu_capable and not heavy` (main.py:4509, 4759). A CT2 cuda load that cannot get a device raises rather than landing on cpu, leaving `_model is None` and `_engine_actual_card` returning `None`. So `test_devices.py:252`'s state is hand-set, and its docstring's "*later landed on cpu (`_device`, mutated)*" describes a mutation production does not currently perform. Worth a sentence saying the carve-out is defensive — otherwise the next reader will believe Whisper's badge is live.

**N3 — the `intent_device` comment overstates, for a provider list it does not enumerate.** `main.py:3358-3360` says "*No usable GPU (`providers` has no CUDA entry) resolves the intent to 'cpu': nothing asked for cuda on this load*". That is false for a `DmlExecutionProvider` or `ROCMExecutionProvider` list — there *is* a usable GPU, and the intent is still published as the literal `"cpu"`. The intent resolver knows one provider spelling; `_ort_providers_to_family` (main.py:9580) knows four. **No shipped impact**, and I am not asking for a behaviour change: `ortProviders` (`server/tts-sidecar/scripts/accelerator-profile.mjs:125-131`) hands CUDA only to the `nvidia` profile and `['CPUExecutionProvider']` to every other one, so `providers` can only ever contain CUDA or CPU on both the injected and the `_default_ort_providers` paths. The comment just claims more than the line does — "no *CUDA* entry" rather than "no usable GPU".

---

### Verdict

**No blocking or significant findings.** Every pass-1 finding is either verified resolved by reverting the hunk and watching the right test redden for the right reason (F1, F3, F4, F5/F6), or disproved on re-derivation and withdrawn (F2, F7). The #2643 change survives the probe it most needed to: the intent and the actual device come from independent sources, so the alarm can genuinely fire; `intent_device` touches nothing but the published string, so placement is provably unchanged; and the ratified compliance case still stays quiet. The three remaining items are documentation/comment accuracy — fixable in this round, but none is a correctness bug and none re-triggers a review round.

PR #2655, opened 2026-08-25
fix(sidecar): expose inflight_synth and re-run #1996's VRAM measurement

dudarenok-maker:
## Summary

Refs #1996. Adds a diagnostic (`inflight_synth` on `/health` + `/debug/memory`) and re-runs #1996's stranded-VRAM measurement gated on that real idle signal instead of a fixed wall clock, which the prior attempt was silently defeated by (unrelated ASR-QA repair still running past the 180s window).

**CORRECTED (2026-08-26, following an adversarial PR review pass — see the review thread below):** the original version of this PR body claimed the 2026-08-25 run's idle window was "clean" and read the run as "the two resident models' own live weights — normal residency, not evidence of a leak." Both claims were wrong, caused by two instrument bugs this PR's own fix round then discovered and fixed:

1. **The idle gate was blind to ASR.** `_inflight_synth` tracked only `/synthesize`-family calls, not `/transcribe`/`/embed` — so the gate could (and, on the recorded run, did) report "confirmed idle" while Whisper was still actively transcribing. Fixed in commit `d4aa7a6c`. This run's own logs (`logs/tts.err.log`) show Whisper transcribing continuously through 07:58:51.768, overlapping the window the old gate declared idle (07:58:49–57).
2. **`/debug/memory`'s qwen block was missing a `base17_loaded` key.** The run's "only two resident models" claim (Qwen Base 0.6B + Whisper) missed a third model, the Qwen 1.7B-Base, which the instrument had no way to report. This run's logs show it loaded mid-session (`07:57:58.112 Qwen 1.7B-Base loaded.`) — a model that, unlike Qwen Base 0.6B, DOES have its own idle TTL (120s default), which hadn't lapsed by the time of the P3 capture (21s later). Fixed in commit `42dddeb8`.

**Corrected outcome: this run was a contaminated-idle / misattributed-residency case, not a clean "sixth case."** It does not close, and does not provide trustworthy evidence toward closing, #1996 criterion 1. A valid re-run using the now-fixed instruments is still owed — real-hardware work not performed as part of this correction. Full corrected narrative: `docs/testing/1996-stranded-vram-measurement.md`'s "Reading the result" section (commit `45b913ce`, refined further by re-review pass 2).

**Not closing #1996 criterion 1** (this conclusion is unchanged from the original body, now for the correct reason).

## Test plan

- [x] `server/tts-sidecar/tests/test_smoke.py::test_health_responsive_during_busy_synth` (extended to assert `inflight_synth == 1` mid-synth and `== 0` after) — passes.
- [x] Independently re-ran the mutation test myself (not trusting the child's receipt): commented out the `"inflight_synth": _inflight_synth` line in `/health`'s response dict, re-ran the test — failed RED with `KeyError: 'inflight_synth'` on the new assertion, confirming it actually exercises the field. Restored the line, re-ran — GREEN again.
- [x] Full `server/tts-sidecar/tests/` suite — all passing (no failures).
- [x] Verified `/debug/memory`'s `inflight_synth` is top-level (sibling to `process`/`gc`/`engines`), not nested under `engines["qwen"]`, per the acceptance spec.
- [x] `/transcribe` and `/embed` now also track `inflight_synth` (commit `d4aa7a6c`), closing the ASR-blind confound above.
- [x] `/debug/memory`'s qwen block now includes `base17_loaded` (commit `42dddeb8`).
- [x] Full results table, attribution subsection, and decision-tree analysis posted (and later corrected) as comments on [#1996](https://github.com/dudarenok-maker/Castwright/issues/1996#issuecomment-5417415286).

Also fixed, found in passing (independent PR review, pass 1 + re-review pass 2): the attribution-during-wait step grepped a log file (`logs/server.log`) that never contains per-request lines and so could never return a non-empty result (fixed in `36091ef2`, cleaned up further in re-review); no test covered `/debug/memory`'s `inflight_synth` key (fixed in `2aebdfe2`); the idle-gate script printed the same "success" message on both the confirmed-idle and ceiling-timeout exit paths (fixed in `3d0f20cc`); the capture snippet never recorded `inflight_synth`, leaving the TOCTOU gap between gate-clear and snapshot unclosed (fixed in `56289b3f`); a doc arithmetic error and a missing cross-reference (fixed in `39fdca76`).

### DIFF
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -50,7 +50,8 @@ Run this at each capture point and paste the output into the table below.
 ```powershell
 $ts = (Get-Date).ToString('HH:mm:ss')
 $mem = Invoke-RestMethod http://127.0.0.1:9000/debug/memory
-"$ts  rss=$($mem.process.rss_mb)MB  inflight_synth=$($mem.inflight_synth)"
+$health = Invoke-RestMethod http://127.0.0.1:9000/health
+"$ts  rss=$($mem.process.rss_mb)MB  inflight_synth=$($mem.inflight_synth)  qwen_base17_loaded=$($health.qwen_base17_loaded)"
 $mem.memory_stats | ConvertTo-Json -Depth 4
 $mem.engines | ConvertTo-Json -Depth 3
 nvidia-smi --query-gpu=index,memory.used,memory.total --format=csv,noheader
@@ -92,7 +93,7 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
    seconds, the idle window is likely clean (only genuine background activity
    that was already running, if anything). A gate that takes 120 seconds or
    longer to clear suggests the box was actively working during what should
-   have been an idle window — consult the `logs/server.log` grep (next section)
+   have been an idle window — consult the `logs/tts.log` grep (next section)
    to name what was running. This 120 s threshold is separate from the 8 s
    quiet-streak requirement above and the 10-minute ceiling; it is a judgment
    call for how long an otherwise-successful idle-gate wait is "suspiciously
@@ -229,7 +230,7 @@ Real forensic evidence from `logs/tts.err.log` for this specific run: Whisper `s
 | Strand absent at P3 | Self-heals at the existing TTLs. #1996 criterion 1 is already satisfied on `main`. |
 | P4 `reserved` drops sharply | **Uncollected cache.** A scheduled reclaim is the fix — subject to every constraint in §7 of the design. |
 | P4 `reserved` barely moves, `inactive_split` dominates, `reclaimed: true` | **Fragmentation.** `empty_cache()` cannot fix this at any cadence; re-open the recycle threshold instead. |
-| Idle gate took >120 s to clear | **Contaminated idle.** Legitimate background activity was still running; name it from the `logs/server.log` grep above. **Do not treat the resulting P3/P4 readings as evidence about a stranded pool at all.** |
+| Idle gate took >120 s to clear | **Contaminated idle.** Legitimate background activity was still running; name it from the `logs/tts.log` grep above. **Do not treat the resulting P3/P4 readings as evidence about a stranded pool at all.** |
 | P4 `reserved` barely moves, `reclaimed: false` | **Invalid reading.** The reclaim never ran — CUDA was unavailable or `empty_cache()` threw. Retry once; treat repeated `false` as a separate finding (a live-context problem), not a #1996 answer. |
 | Strand present at P3 *and* P4 frees it *and* RSS was above 8192 MB at P2 | Contradicts the 60 s watchdog reclaim having run. Investigate that before designing anything — one of the two observations is wrong. |
 
@@ -237,9 +238,9 @@ Real forensic evidence from `logs/tts.err.log` for this specific run: Whisper `s
 
 This run fits the **"contaminated idle"** row: the idle-gate signal and the attribution verification both relied on broken instruments that are now known to have been blind to the actual work running.
 
-**The contamination:** Whisper ASR was transcribing continuously during what the old idle gate declared "confirmed idle" (07:58:49–07:58:57 vs actual Whisper activity 07:58:31–51.768). The old idle gate could not see this because it tracked only `inflight_synth`, not `/transcribe` or `/embed` (fixed in commit d4aa7a6c). The attribution grep checked the wrong log file (`logs/server.log` instead of `logs/tts.log`, fixed in commit 36091ef2). By the decision tree's logic: "Legitimate background activity was still running; name it from the `logs/server.log` grep above. **Do not treat the resulting P3/P4 readings as evidence about a stranded pool at all.**"
+**The contamination:** Whisper ASR was transcribing continuously during what the old idle gate declared "confirmed idle" (07:58:49–07:58:57 vs actual Whisper activity 07:58:31–51.768). The old idle gate could not see this because it tracked only `inflight_synth`, not `/transcribe` or `/embed` (fixed in commit d4aa7a6c). The attribution grep checked the wrong log file (`logs/server.log` instead of `logs/tts.log`, fixed in commit 36091ef2). Legitimate background activity (Whisper) was still running during the supposed idle window — **do not treat the resulting P3/P4 readings as evidence about a stranded pool at all.**
 
-**The misattribution:** The run loaded three resident models — Qwen Base 0.6B (no TTL), Whisper (120s TTL), and Qwen 1.7B-Base (120s TTL) — but the `/debug/memory` endpoint at the time of this run could not represent the 1.7B model (missing the `base17_loaded` key, added retroactively in commit 42dddeb8). The ~3.3 GB gap between P1's `allocated` (1776.7 MB) and P2's `allocated` (5453.7 MB) is plausibly the Qwen 1.7B-Base model loading during session rather than something inexplicable. The 1.7B model's 120s idle TTL means P3 (21 s post-P2) would not have evicted it even if the idle window had been clean — which it was not.
+**The misattribution:** The run loaded three resident models — Qwen Base 0.6B (no TTL), Whisper (120s TTL), and Qwen 1.7B-Base (120s TTL) — but the `/debug/memory` endpoint at the time of this run could not represent the 1.7B model (missing the `base17_loaded` key, added retroactively in commit 42dddeb8). The 3677.0 MB gap between P1's `allocated` (1776.7 MB) and P2's `allocated` (5453.7 MB) is the Qwen 1.7B-Base model loading during the run. Log evidence from `logs/tts.err.log` confirms: `07:57:58.112 Qwen 1.7B-Base loaded.` and `07:58:29.275 qwen batch synth: model=1.7b items=7 voices=3`. At the time this run was captured, `/debug/memory` could not track the 1.7B model's residency — but `/health` has exposed a `qwen_base17_loaded` field since well before this branch (main.py:~9645, consumed by `server/src/routes/sidecar-health.ts:~458`). The real issue is that the capture snippet in step "The capture" only calls `/debug/memory`, never `/health`, so a signal that was already available went unrecorded. The 1.7B model's 120s idle TTL means P3 (21 s post-P2) would not have evicted it even if the idle window had been clean — which it was not.
 
 **Conclusion:** This run does NOT provide trustworthy evidence. Its P3/P4 readings show a contaminated-idle state with unattributable residency, not a measurement of a stranded pool. It does not close #1996's criterion 1 (and does not discharge the acceptance row already owed).
 
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -50,7 +50,7 @@ Run this at each capture point and paste the output into the table below.
 ```powershell
 $ts = (Get-Date).ToString('HH:mm:ss')
 $mem = Invoke-RestMethod http://127.0.0.1:9000/debug/memory
-"$ts  rss=$($mem.process.rss_mb)MB"
+"$ts  rss=$($mem.process.rss_mb)MB  inflight_synth=$($mem.inflight_synth)"
 $mem.memory_stats | ConvertTo-Json -Depth 4
 $mem.engines | ConvertTo-Json -Depth 3
 nvidia-smi --query-gpu=index,memory.used,memory.total --format=csv,noheader
@@ -110,7 +110,7 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
    column. **If the gate never clears within 10 minutes, stop and report** —
    do not force P3/P4 on a busy box.
 
-   Once the gate reports idle, capture as **P3**.
+   Once the gate reports idle, capture as **P3**. **Before treating P3 as a confirmed-idle reading, verify in the capture output that `inflight_synth == 0` — if it is non-zero, the idle gate's poll is stale and the TOCTOU window is open. If so, stop and re-run the idle-gate wait.**
 
    > **If the strand is gone at P3 the run is already decisive** — the pool
    > self-heals on `main` today (decision-tree row 1). Record it and stop; do
@@ -150,7 +150,7 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
 
    It returns the per-device figures **before and after** a bare
    `empty_cache()`, bracketed with nothing in between, plus a `reclaimed: bool`
-   field indicating whether the reclaim actually ran. Record all three as **P4**.
+   field indicating whether the reclaim actually ran. Record all three as **P4 before** and **P4 after**. **Before treating P4's readings as valid, capture `/debug/memory` one more time after the reclaim completes and verify that `inflight_synth == 0` — if it has changed, background synthesis occurred between the idle gate and the reclaim, invalidating the confirmed-idle assumption. If so, stop and re-run from step 4.**
 
    > Do **not** substitute a load/unload cycle. A load is an allocation pass that
    > refills and coalesces segments — conflating the two is the unsound inference
@@ -170,15 +170,15 @@ Run: 2026-08-26, this box (`cuda:0` RTX 4070 Laptop 8188 MB, `cuda:1` RTX 5070 T
 41 lines), re-rendered with `force:true` on `qwen3-tts-0.6b`. Both GPUs were idle
 (0 MiB used, 0% util) before the run started.
 
-| Point | When | time to idle | Device | reserved bytes | allocated bytes | inactive_split bytes | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
-|---|---|---|---|---|---|---|---|---|---|---|---|
-| P0 | fresh, 07:53:44 | | cuda:1 | 0 | 0 | 0 | 0 | 197 MiB | 1143.0 | none | |
-| P1 | mid-render, 07:54:32 (~8 s in) | | cuda:1 | 1,881,145,344 (1794.4 MB) | 1,863,124,480 (1776.7 MB) | 13,826,560 (13.2 MB) | 17,975,296 (17.1 MB) | 2051 MiB | 3965.9 | qwen base | |
-| P2 | +0 s, 07:58:36 | | cuda:1 | 6,138,363,904 (5853.4 MB) | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | 6489 MiB | 19914.2 | qwen base, whisper | |
-| P3 | confirmed idle, 07:58:57 | **21 s** | cuda:1 | 6,138,363,904 (5853.4 MB) — **identical to P2** | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | 6489 MiB | 19799.1 | qwen base, whisper | |
-| P4 before | reclaim, 07:58:58 | | cuda:1 | 6,138,363,904 (5853.4 MB) | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | | | | true |
-| P4 after | reclaim, 07:58:58 | | cuda:1 | 5,748,293,632 (5481.4 MB) | 5,717,628,928 (5453.7 MB) — **unchanged** | 30,664,704 (29.25 MB) — **unchanged** | 30,664,704 (29.25 MB) | | | | true |
-| P5 | cuda:1 render, throughout P1–P4 | | cuda:0 | 2,097,152 (2.0 MB, unchanged pre/post) | ≈0 | ≈2.0 MB | | 0–116 MiB (baseline noise) | | | |
+| Point | When | time to idle | Device | reserved bytes | allocated bytes | inactive_split bytes | reserved − allocated | nvidia-smi used | RSS MB | inflight_synth | Engines loaded | reclaimed |
+|---|---|---|---|---|---|---|---|---|---|---|---|---|
+| P0 | fresh, 07:53:44 | | cuda:1 | 0 | 0 | 0 | 0 | 197 MiB | 1143.0 | n/a — pre-fix run | none | |
+| P1 | mid-render, 07:54:32 (~8 s in) | | cuda:1 | 1,881,145,344 (1794.4 MB) | 1,863,124,480 (1776.7 MB) | 13,826,560 (13.2 MB) | 17,975,296 (17.1 MB) | 2051 MiB | 3965.9 | n/a — pre-fix run | qwen base | |
+| P2 | +0 s, 07:58:36 | | cuda:1 | 6,138,363,904 (5853.4 MB) | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | 6489 MiB | 19914.2 | n/a — pre-fix run | qwen base, whisper | |
+| P3 | confirmed idle, 07:58:57 | **21 s** | cuda:1 | 6,138,363,904 (5853.4 MB) — **identical to P2** | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | 6489 MiB | 19799.1 | n/a — pre-fix run | qwen base, whisper | |
+| P4 before | reclaim, 07:58:58 | | cuda:1 | 6,138,363,904 (5853.4 MB) | 5,717,628,928 (5453.7 MB) | 30,664,704 (29.25 MB) | 420,734,976 (401.3 MB) | | | n/a — pre-fix run | | true |
+| P4 after | reclaim, 07:58:58 | | cuda:1 | 5,748,293,632 (5481.4 MB) | 5,717,628,928 (5453.7 MB) — **unchanged** | 30,664,704 (29.25 MB) — **unchanged** | 30,664,704 (29.25 MB) | | | n/a — pre-fix run | | true |
+| P5 | cuda:1 render, throughout P1–P4 | | cuda:0 | 2,097,152 (2.0 MB, unchanged pre/post) | ≈0 | ≈2.0 MB | | 0–116 MiB (baseline noise) | | n/a — pre-fix run | | |
 
 Reclaim delta on `cuda:1`: `reserved` dropped 390,070,272 B (**372.0 MB, 6.4% of
 the 5853.4 MB reserved**). `allocated` and `inactive_split` did not move at all.
--- a/docs/testing/1996-stranded-vram-measurement.md
+++ b/docs/testing/1996-stranded-vram-measurement.md
@@ -22,6 +22,22 @@ so it takes no row in the on-box acceptance register.
 - A book with at least two chapters. `the-coalfall-commission.md` (the canonical
   fixture) is fine.
 - Qwen as the generation engine.
+- **Start the app via `npm start`** (not `npm run dev`) so that `logs/server.log`
+  and `logs/tts.err.log` capture the full timeline. `npm run dev` does not
+  redirect console output to the log files, which is why the Aug 23 run's
+  Node-side eviction log only survives as a manually-pasted GitHub comment line
+  and a burst of unexplained post-render Qwen activity could not be attributed to
+  a caller.
+- **Confirm `QWEN_DEVICE` and `ASR_DEVICE` are `cuda:1`** (the box's standing
+  device policy — `server/.env.example:528-534`) before starting. `cuda:1` is
+  the RTX 5070 Ti (16 GB); `cuda:0` (the 8 GB 4070 Laptop) is deliberately left
+  free for other work. Do not deviate from the pinned `cuda:1` policy unless
+  there is a documented reason — and if you must, record the deviation exactly
+  like the Aug 23 run did.
+- **Confirm no other chapter or book job is queued** and no Cast Review or
+  audition session is open in another browser tab. The last run's extra Qwen
+  activity during its supposed idle window is suspected to be exactly this kind
+  of confound — this run needs to rule it out or name it.
 
 Sidecar is at `http://127.0.0.1:9000` unless `LOCAL_TTS_PORT` says otherwise.
 
@@ -58,13 +74,67 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
    Record which engines report loaded; specifically whether **Coqui, Kokoro or
    Qwen Base 0.6B** is among them (the three engines with no idle TTL).
 
-4. **180 s post-render.** Leave the box completely idle — no UI interaction, no
-   transcribe, nothing that would touch an engine. Capture as **P3**.
+   Stamp the P2 capture time so the idle gate in step 4 can report elapsed:
 
-   > 180 s is past every existing 120 s TTL. **If the strand is gone at P3 the
-   > run is already decisive** — the pool self-heals on `main` today
-   > (decision-tree row 3). Record it and stop; do not retry until a strand
-   > appears.
+   ```powershell
+   $captureP2Time = Get-Date
+   ```
+
+4. **Wait for confirmed idle.** Leave the box idle — no UI interaction, no
+   transcribe, nothing that would touch an engine. Instead of a fixed 180 s
+   wall clock, poll the sidecar's `/health` endpoint for `inflight_synth` (an
+   int; `0` means nothing is generating) until it has been quiet for five
+   consecutive checks (10 s of confirmed idle). A 10-minute safety ceiling
+   prevents an infinite wait on a busy box.
+
+   Run this as **one blocking script**, not turn-by-turn polling — it
+   self-terminates once idle or after the ceiling:
+
+   ```powershell
+   $deadline = (Get-Date).AddMinutes(10)
+   $quietStreak = 0
+   while ((Get-Date) -lt $deadline) {
+     $h = Invoke-RestMethod http://127.0.0.1:9000/health
+     if ($h.inflight_synth -eq 0) { $quietStreak++ } else { $quietStreak = 0 }
+     if ($quietStreak -ge 5) { break }  # 5 consecutive quiet polls = 10s confirmed idle
+     Start-Sleep -Seconds 2
+   }
+   "Went idle after $((Get-Date) - $captureP2Time)"
+   ```
+
+   Record the actual elapsed time in the results table's **"time to idle"**
+   column. **If the gate never clears within 10 minutes, stop and report** —
+   do not force P3/P4 on a busy box.
+
+   Once the gate reports idle, capture as **P3**.
+
+   > **If the strand is gone at P3 the run is already decisive** — the pool
+   > self-heals on `main` today (decision-tree row 1). Record it and stop; do
+   > not retry until a strand appears.
+
+   **What was active during the wait.** Before moving to P3/P4, grep
+   `logs/server.log` for every request timestamped between the P2 capture and
+   the confirmed-idle point. List each one (route + any chapter/job id in the
+   log line) in a new **"What was active during the wait"** subsection of the
+   results. This closes the attribution gap the Aug 23 run left open — a burst
+   of unexplained post-render Qwen activity could not be attributed to a caller
+   because `npm run dev` did not write to `logs/server.log`.
+
+   ```powershell
+   $idleTime = Get-Date
+   Select-String -Path logs\server.log -Pattern 'GET |POST |PUT |PATCH |DELETE ' |
+     ForEach-Object {
+       $line = $_.Line
+       if ($line -match '(\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2})') {
+         $ts = [datetime]::Parse($matches[1])
+         if ($ts -ge $captureP2Time -and $ts -le $idleTime) { $line }
+       }
+     }
+   ```
+
+   If the grep returns nothing, the idle window was genuinely clean. If it
+   returns entries, name them in the results — the decision tree's fourth row
+   uses this to tell a contaminated idle from a real strand.
 
 5. **The bare reclaim.** With the box still idle, one request:
 
@@ -89,19 +159,23 @@ Record, per device: `reserved`, `allocated`, `inactive_split`, and
 
 ## Results
 
-| Point | When | Device | reserved bytes | allocated bytes | inactive_split bytes | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
-|---|---|---|---|---|---|---|---|---|---|---|
-| P0 | fresh | | | | | | | | | |
-| P1 | mid-render | | | | | | | | | |
-| P2 | +0 s | | | | | | | | | |
-| P3 | +180 s | | | | | | | | | |
-| P4 before | reclaim | | | | | | | | | |
-| P4 after | reclaim | | | | | | | | | |
-| P5 | cuda:1 render | | | | | | | | | |
+| Point | When | time to idle | Device | reserved bytes | allocated bytes | inactive_split bytes | reserved − allocated | nvidia-smi used | RSS MB | Engines loaded | reclaimed |
+|---|---|---|---|---|---|---|---|---|---|---|---|
+| P0 | fresh | | | | | | | | | | |
+| P1 | mid-render | | | | | | | | | | |
+| P2 | +0 s | | | | | | | | | | |
+| P3 | confirmed idle | _elapsed_ | | | | | | | | | |
+| P4 before | reclaim | | | | | | | | | | |
+| P4 after | reclaim | | | | | | | | | | |
+| P5 | cuda:1 render | | | | | | | | | | |
 
 **RSS at P2 vs 8192 MB:** ☐ above ☐ below
 *(Above means the memory watchdog's unconditional 60 s `gc+empty_cache` was
 already firing throughout — `main.py:7957-7964` (`_mem_warn_threshold_mb()`).)*
+
+**What was active during the wait** (from `logs/server.log`, P2 → confirmed-idle):
+
+<!-- List each request line here: route + chapter/job id. If none, write "idle window clean". -->
 
 ---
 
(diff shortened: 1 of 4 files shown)


dudarenok-maker:
## PR review — pass 1 (head 218cc562fc5305e4f8737e77576f6309c97a14d9, depth high)

Scope: `server/tts-sidecar/main.py`, `server/tts-sidecar/tests/test_smoke.py`, `docs/testing/1996-stranded-vram-measurement.md` · verified: read both changed handlers and **every** `_inflight_synth` increment/decrement site at the head SHA, enumerated all 27 sidecar routes, confirmed the existing `test_memory.py` `/debug/memory` assertions, then cross-checked the recorded run against **its own worktree's logs** (`C:/Claude/Projects/wt-1996-idle-vram-measurement/logs/{server,tts.err}.log`) before adversarially probing the idle gate, the test's mutation sensitivity, and the doc's arithmetic.

The `/health` half of the code change is correct and its test is genuinely mutation-sensitive. **The measurement it was built to enable did not work**, and the run sheet records the opposite. Three of the findings below are reproduced from the run's own log files, not reasoned about.

---

### 🔴 Blocking — the idle gate cannot see ASR, and the recorded run was saturated with ASR for most of its 21 "confirmed idle" seconds

**Mechanism.** `_inflight_synth` is incremented at exactly six sites — `main.py:10191` (`/qwen/design-voice`), `:10318` (`/qwen/clone-voice`), `:10408` (`/qwen/mint-variant`), `:10616` (`/xtts/clone-voice`), `:10818` (`/synthesize`), `:11212` (`/synthesize-batch`). `POST /transcribe` (`main.py:10957-11034`) and `POST /embed` (`main.py:11037`) increment **nothing**; the counter's own declaration comment says so (`main.py:8141`: "counts live synth calls"). `/transcribe`'s docstring claims it "Honours the same poison + recycle-drain fences" — it honours the `_restart_pending` fence, but never holds the counter.

So a box running back-to-back Whisper transcriptions reads `inflight_synth == 0` on every poll and the gate at `docs/testing/1996-stranded-vram-measurement.md:93-103` clears immediately. **The confound this PR exists to rule out — an ASR-QA repair still running past the wait — is precisely the one the new signal is blind to.**

**Reproduced in this PR's own run.** From the measurement worktree's `logs/tts.err.log` (all timestamps 2026-08-26):

```
07:58:29.275  qwen batch synth: model=1.7b items=7 voices=3 ... rtf=1.58
07:58:31.400  Loading Whisper ASR model=small device=cuda:1 compute=int8_float16 ...
07:58:34.174  Whisper ASR loaded (model=small device=cuda:1).
07:58:34.177  Processing audio with duration 00:07.040
   ... 100+ continuous "Processing audio" / "VAD filter" / "Detected language" lines ...
07:58:51.768  Detected language 'en' with probability 0.99      <- last ASR line
```

The doc records **P2 at 07:58:36 and confirmed-idle P3 at 07:58:57** (`:171-172`). The gate's own arithmetic (5 quiet polls, 4 x 2s sleeps) needs 8s, so the quiet streak began at ~07:58:49 — **overlapping live GPU transcription until 07:58:51.8**. A Whisper `small` model was also *loaded onto `cuda:1` at 07:58:31-34*, i.e. inside the measurement window itself, and its weights are inside P3's `allocated`.

The doc's line 226 — "Not 'contaminated idle' — the gate cleared in 21 s ... and the attribution grep is empty" — is false. This is `reviewer-brief.md` shape 4 (*a metric blind to a case it must score*) and shape 5 (*an acceptance criterion blind to its own feature*).

**Correctness bug.** The fix is the mechanism, not the run: the idle signal must cover every GPU-touching route (`/transcribe`, `/embed` at minimum) — either by extending the counter or by surfacing a second `inflight_asr` / `inflight_total` field — with a paired test in the shape of the one that already exists for design-voice (`test_memory.py:473 test_design_voice_counts_as_inflight_synth`). Then the measurement is re-run and the results section replaced.

---

### 🔴 Blocking — the attribution backstop is an instrument that cannot return a non-empty result

`docs/testing/1996-stranded-vram-measurement.md:125` greps `logs\server.log` for `'GET |POST |PUT |PATCH |DELETE '`. Three independent reasons it can only ever return zero:

1. **The Node server has no request-logging middleware.** No `morgan`, no `requestLogger`, no `${req.method}` log site anywhere under `server/src/`. Its stdout is `[server]` / `[tts:catalog]` / `[generation]`-prefixed lines.
2. **Sidecar traffic never reaches `server.log`.** `server/src/tts/spawn-sidecar.ts:748-759` sends the sidecar to `logs/tts.log` + `logs/tts.err.log`. The ASR storm above lives there, in the file the grep does not read.
3. **ASR-QA is an in-process job loop**, not an inbound HTTP request — it produces no request line even in principle.

**Measured:** the run's own `logs/server.log` is **16 lines total**, and the doc's exact grep pattern matches **0** of them — for the whole session, including the four-minute render (P1 07:54:32 -> P2 07:58:36) when the sidecar was demonstrably busy. The instrument returns "clean" for a window in which a chapter was actively rendering. The doc then treats that null result as positive evidence (`:199`: "**idle window clean** ... No confound; this run's P3/P4 readings are trustworthy evidence"). `reviewer-brief.md` shapes 1 (*fails open on absent evidence*), 6 (*success reported while doing nothing*) and 8 (*a control group labelled clean but never measured*).

**Compounding: the run violated the prerequisite added for this exact purpose.** `:25` mandates `npm start` (not `npm run dev`) so `logs/server.log` captures the timeline. That log's first lines are:

```
> castwright@1.14.0 dev:server
> npm --prefix server run dev
> tsx watch --include=.env src/index.ts
```

The run used `npm run dev`. (`logs/start-failed.txt` in the same directory records `Services did not come up within 60 s: frontend, server`.) The prerequisite was written, then not met, and nothing in the procedure detects that.

**Correctness bug.** The attribution step must read `logs/tts.err.log` (where sidecar request/engine activity actually is) and must fail loudly when the log has no lines in the target window at all, rather than reporting "clean".

---

### 🔴 Blocking — the resident-model attribution is wrong: three models were resident at P2/P3, and the largest is neither named nor visible to the instrument

The verdict at `:233-240` rests on "`qwen.base_loaded=true` and `whisper.model_loaded=true`", concluding the 5453.7 MB is "**the two resident models'** own live weights ... normal residency, not a leak", specifically because "**Qwen Base 0.6B has no idle TTL at all**".

The log says otherwise. No `unload` / `evict` line appears anywhere in the session:

```
07:54:16.471  Qwen Base loaded.               <- _base   (0.6B)
07:57:58.112  Qwen 1.7B-Base loaded.          <- _base17 (1.7B)  -- 38 s before P2
07:58:34.174  Whisper ASR loaded (model=small device=cuda:1).
```

Three consequences, each independently breaking the verdict:

1. **Three models, not two**, and the unnamed third is the biggest. A 1.7B TTS model dominates a 0.6B one — and it is close to the ~3.3 GB the doc's own numbers leave unexplained (P1, mid-render, Qwen resident and actively generating: `allocated` = 1776.7 MB, `:170`. P2: 5453.7 MB. A Whisper `small` int8_float16 does not span that gap).
2. **The instrument is blind to it.** `QwenEngine` holds `_base` (0.6B, `main.py:5183`) and `_base17` (1.7B, `main.py:5184`) as separate attributes, but `/debug/memory`'s engines block reports only `base_loaded` / `design_loaded` / `prompt_cache_entries` (`main.py:9765-9772`) — there is **no `base17_loaded` key**. The doc read a field that cannot represent the 1.7B and filled the gap by assumption. This PR was already editing `/debug/memory` for this measurement and is where the missing key belongs.
3. **"No idle TTL at all" is false for the model that matters.** `_base17` has one: `QWEN_BASE17_IDLE_TTL` (`main.py:7913-7918`, defaulting to the 120 s design TTL), reaped by the watchdog at `main.py:7942-7944` ("Qwen 1.7B-Base freed after >%.0fs idle"). P3 was taken 21 s after P2 — well inside it. Simply waiting was an available, un-attempted route, which the doc rules out as "impossible for Qwen Base, which has none".

The run header (`:164`) and the PR body both state the render was "on `qwen3-tts-0.6b`". The log shows a mix (`model=0.6b` x2, `model=1.7b` x1) with the 1.7B loaded last and still resident, and `server.log` records the sidecar spawned with `modelKey=qwen3-tts-1.7b`.

**Correctness bug** in the shipped analysis, and it has already propagated: `docs/testing/onbox-acceptance-register.md:2761-2795` (row A46, already on `main`) repeats "Qwen Base 0.6B has **no idle TTL**", "the two resident models", and the 21 s confirmed-idle window, and prescribes a next step (`POST /unload {qwen}`) written against the wrong model. A46 is a chore this finding makes owed — same round, per findings-triage.

---

### 🟠 Significant — `/debug/memory`'s new field has no test; deleting it leaves the whole suite green

`main.py:9763` (`out["inflight_synth"] = _inflight_synth`) is asserted nowhere. The only new assertions are `test_smoke.py:261` and `:273`, both against `/health`. `test_memory.py:526` (`test_debug_memory_endpoint_shape`) checks `process` / `committed_mb` / `gc` / `engines` / `cuda`, and `test_memory.py:586` checks `memory_stats` — neither touches `inflight_synth`.

The PR body's mutation check explicitly covers only "the `/health` field addition". Comment out `main.py:9763` and `pytest server/tts-sidecar/tests/` stays green. `reviewer-brief.md` shape 3, and it fails the house gate "the paired test is a real regression test" for half the diff. One assertion in `test_debug_memory_endpoint_shape` closes it.

---

### 🟠 Significant — the idle-gate script reports "Went idle" when the ceiling expires without ever going idle

`:93-103`:

```powershell
while ((Get-Date) -lt $deadline) { ...; if ($quietStreak -ge 5) { break }; Start-Sleep -Seconds 2 }
"Went idle after $((Get-Date) - $captureP2Time)"
```

Both exits print the same line. A box that never quiets for 10 minutes prints `Went idle after 00:10:00.xx`, and the operator proceeds to P3 — the run sheet's own instruction two lines below ("**If the gate never clears within 10 minutes, stop and report**", `:105`) has no signal to act on. `reviewer-brief.md` shape 6, inside a guard written to prevent shape 6.

Guard the print on `$quietStreak -ge 5` and emit an explicit failure otherwise. (Related: `$captureP2Time` is set by a *different* snippet at `:76`; run this block in a fresh shell and the subtraction throws.)

---

### 🟠 Significant — no capture point records `inflight_synth`, so the TOCTOU window between gate-clear and snapshot is never closed

The capture block (`:50-58`) prints `process.rss_mb`, `memory_stats`, `engines` and `nvidia-smi` — never `inflight_synth`, and the results table (`:167`) has no column for it. The gate clears in one script; P3 is captured by a later, separate command; the attribution grep (`:118-133`) runs *between* them. **Nothing in the recorded evidence shows the counter was 0 at the instant of any snapshot.** That is the entire reason the field was added to `/debug/memory`, and the consumer never reads it. Add `inflight_synth` to the capture snippet and a column to the table, so idleness is a recorded property of each capture point rather than of an earlier moment.

---

### 🟡 Minor — "5 consecutive quiet polls = 10 s confirmed idle" is 8 s

`:87` and `:99`. Five polls separated by four 2 s sleeps span 8 s (the first poll costs no sleep). Off by one interval in two places, and it is a figure the analysis reasons from.

---

### 🟡 Minor — the new decision-tree row's 120 s threshold exists nowhere in the procedure it judges

`:212` ("Idle gate took >120 s to clear -> Contaminated idle") introduces a number step 4 never mentions — step 4 has a 10-minute ceiling and instructs a stop only there. A gate clearing at 121 s is "contaminated idle" per the table and entirely unremarkable per the script. Independently of finding 1, the row also points at the `logs/server.log` grep that finding 2 shows cannot fire.

---

### 🟡 Minor — release-notes gate neither satisfied nor waived

No `docs/release-notes-next.md` / `RELEASE_NOTES.md` entry, and the PR body carries no explicit not-applicable. A new operator-facing `/health` + `/debug/memory` field is plausibly a legitimate skip — CLAUDE.md step 5 requires saying so rather than omitting it silently.

---

### ✅ What is solid

- **The `/health` field and its test.** `main.py:9703` is correct, and `test_smoke.py:261/:273` is genuinely mutation-sensitive — removing the field raises `KeyError`. The `== 0` assertion is race-free by construction, not by luck: `_inflight_synth -= 1` sits in the `finally` at `main.py:10942`, which runs before the `Response` is built at `:10944-10955`, so a joined synth thread guarantees the decrement has landed.
- **All increments happen on the event loop** before the `to_thread` offload, so the "plain int under the GIL" claim holds.
- **The reclaim arithmetic checks out.** 6,138,363,904 - 5,748,293,632 = 390,070,272 B = 372.0 MiB = 6.36% ✓; post-reclaim `reserved - allocated` = 30,664,704 B exactly equals `inactive_split` ✓. Byte->MB conversions are MiB throughout, consistently.
- **The P5 cross-device substitution is sound** — `_cuda_memory_stats_per_device()` genuinely reports per-device, so the `cuda:0` row is a real reading, not a current-device artifact.
- **The PR does not overclaim its own conclusion.** Declining to close #1996 criterion 1 and recording the follow-up question is the right call — it is the reasoning underneath it that does not hold.

---

### Verdict

**Do not merge as-is.** The `/health` change is fine; ship it. The recorded measurement is not evidence and should not be merged as the file #2656 and register row A46 are briefed from — the idle gate did not detect the confound it was built for (finding 1), the backstop that was supposed to catch that cannot produce a non-empty result (finding 2), and the resident-model attribution the verdict rests on is contradicted by the run's own log (finding 3). All three are reproducible from `C:/Claude/Projects/wt-1996-idle-vram-measurement/logs/`.

Suggested shape: fix the mechanism (findings 1, 4, 5, 6 plus the missing `base17_loaded` key), re-run the measurement with a gate that can actually see ASR and a 1.7B whose TTL has lapsed, then rewrite the results section — and correct A46 in the same round.

6 correctness bugs (3 🔴, 3 🟠) -> re-review triggered once fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head 45b913ce920a2e9094398d6c58ee343a517daff9, depth high)

Scope: `server/tts-sidecar/main.py`, `tests/test_smoke.py`, `tests/test_memory.py`, `docs/testing/1996-stranded-vram-measurement.md` · verified: the 5 new/modified sidecar tests run green (`test_smoke.py::test_health_responsive_during_busy_synth`, `::test_inflight_synth_tracks_transcribe_and_embed`, `test_memory.py::test_debug_memory_includes_inflight_synth_top_level_key`, `::test_debug_memory_inflight_synth_tracks_busy_synth`, `::test_debug_memory_endpoint_shape` — 5 passed) before cross-checking commit `45b913ce`'s corrected narrative against this box's raw `logs/tts.err.log` / `logs/tts.log` from the 2026-08-26 run, and auditing the recycle-drain interaction.

| # | Pass 1 | Disposition |
|---|---|---|
| 1 | 🔴 idle gate blind to ASR | **VERIFIED RESOLVED** (code). `main.py:11000-11001,11037-11038` and `11078-11079,11109-11110` increment before the `try` and decrement in `finally`. I traced every exit: the `_no_capacity(adm)` early return inside the `async with`, the `except Exception` returns, and `CancelledError` (a `BaseException`, so uncaught but still runs `finally`) all decrement exactly once. Test coverage is only half there — see 🟠 A. |
| 2 | 🔴 attribution grep read `logs/server.log` | **NOT RESOLVED** — see 🔴 A. |
| 3 | 🔴 `/debug/memory` missing `base17_loaded` | **RESOLVED** in code (`main.py:9772`). Test is mutation-insensitive — see 🟡 F. |
| 4 | 🟠 no coverage for `/debug/memory`'s `inflight_synth` | **VERIFIED RESOLVED.** Both new tests are genuinely mutation-sensitive for the `/synthesize` path: deleting `out["inflight_synth"]` → `KeyError`; deleting the `/synthesize` increment → the `inflight_busy > 0` assert fails. |
| 5 | 🟠 idle-gate script always claimed success | **VERIFIED RESOLVED.** `1996-stranded-vram-measurement.md:113-117` branches on `$quietStreak -ge 5` and prints a distinct `CEILING TIMEOUT … STOP` message. |
| 6 | 🟠 TOCTOU between gate and P3/P4 | **VERIFIED RESOLVED.** Capture snippet prints `inflight_synth` (line 53), the results table carries the column, and lines 124 + 164 instruct a re-verify at P3 and after the reclaim. |
| 7 | 🟡 "5 quiet polls = 10s" | **VERIFIED RESOLVED** at lines 87, 110, 130. |
| 8 | 🟡 missing 120 s cross-reference | **RESOLVED**, but the note it added re-armed finding 2 — see 🔴 A. |
| 9 | 🟡 release notes | Recorded, but only in commit `39fdca76`'s message, and the rationale doesn't cover the drain change — see 🟡 F. |

### ✅ The recycle-drain design question is answered — the interaction is safe

Asked in pass 1 and worth recording: extending `_inflight_synth` to `/transcribe` and `/embed` does **not** hazard the drain-before-recycle paths (`main.py:8613-8631`, `8713`, `8744`, `10087-10101`).

- Both routes already fast-fail `503` on `_restart_pending` (`main.py:10982-10986`, `11062-11066`), so once a recycle is scheduled no *new* ASR/embed call can enter and re-arm the counter. There is no starvation path.
- `_drain_then_restart` is bounded by `SIDECAR_DRAIN_GRACE_MS` (default `180000`, `main.py:8587`) and exits regardless, so the worst case is bounded even if that fence were bypassed.
- Measured cost is negligible: this run's own log shows each transcribe taking 0.3–1.4 s, so the added drain latency is ~1–2 s.
- Net effect is a correctness *improvement* — a recycle can no longer hard-exit through an in-flight `/transcribe` — and it makes both routes' pre-existing docstring claim ("Honours the same poison + recycle-drain fences") true for the first time.

---

### 🔴 Blocking — pass 1's finding 2 is still live: the run sheet routes the operator to `logs/server.log` at three places, two of them written *after* the fix

`origin/main`'s copy of this file contains **no log-file reference at all** (verified: `git show origin/main:docs/testing/1996-stranded-vram-measurement.md | grep log` matches only the word "topology" line). So every `logs/server.log` mention below is this branch's own text, and none is pre-existing.

Commit `36091ef2` fixed the §4 attribution subsection only. Three live references survive:

- **`docs/testing/1996-stranded-vram-measurement.md:95`** — *added by `39fdca76`*, a fix commit in this same review round, three commits **after** `36091ef2` retired the instrument: "consult the `logs/server.log` grep (next section) to name what was running." There is no `logs/server.log` grep in the next section any more; it is a `logs/tts.log` line-count diff.
- **`:232`** — the decision table's "contaminated idle" row: "name it from the `logs/server.log` grep above."
- **`:240`** — *added by `45b913ce`*, the final commit, which **quotes the stale row verbatim as authority**: "By the decision tree's logic: \"…name it from the `logs/server.log` grep above.\""

Concrete failure: an operator runs the re-run this document exists to specify, the gate clears in >120 s, they follow line 95 → line 232, grep `logs/server.log`, get zero request lines (the Node server has no request-logging middleware — line 219 of this same file states the instrument "cannot report anything, ever"), and record "nothing was running." That is bit-for-bit the failure this PR was opened to retract. The document now simultaneously diagnoses the instrument as structurally null (line 219) and instructs its use twice more.

### 🔴 Blocking — the retraction landed in one of three surfaces; the PR body and the #1996 record still assert the withdrawn conclusion

`45b913ce` rewrote the run sheet. Neither of the other two durable records was corrected, and both are more widely read than the run sheet:

- **This PR's body** still says: "the idle gate cleared in 21s with an empty attribution grep, **so not 'contaminated idle'**"; "At capture time `qwen.base_loaded=true` and `whisper.model_loaded=true`" (two models, not three); "this run's 5.45GB of 'stranded' VRAM reads … as **the two resident models' own live weights — normal residency, not evidence of a leak**"; and "**none of the run sheet's five predefined rows fit — a sixth case**." The run sheet now says the run *is* the contaminated-idle row, that three models were resident, and that the evidence is invalid. The PR body becomes the merge-commit body, so this ships the retracted claim into git history alongside its own retraction. (Its test-plan bullet is stale too: "Confirmed no source file outside `main.py`, `test_smoke.py`, and `docs/testing/…` changed" — `tests/test_memory.py` changed in `2aebdfe2`.)
- **The #1996 comment** (`issuecomment-5417415286`), which this very run sheet designates as the record of outcome ("Record the outcome as a comment on #1996", line 248), still reads: "Idle gate cleared in 21s; attribution grep over that window is empty (**clean idle, not contaminated**)" and "**normal residency, not evidence of a leak on top of it**". A reader of #1996 — the entire audience for this work — gets the false conclusion with nothing pointing at the correction. The run sheet's own closing sentence names this exact hazard: leaving a wrong claim in place "is what would send a third attempt down the same path."

### 🟠 Significant — the `/embed` increment has zero test coverage, and the test named for it never calls `/embed`

`server/tts-sidecar/tests/test_smoke.py:279` — `test_inflight_synth_tracks_transcribe_and_embed`. Its body stubs `main.ASR.transcribe` and POSTs `/transcribe` only. There is no `/embed` request anywhere in it.

Proof the class is untested, not just this test: the complete set of `inflight_synth` references across `server/tts-sidecar/tests/` is `test_memory.py` (drain tests via `monkeypatch.setattr(main, "_inflight_synth", …)`, the design-voice test at `:473`, the two new `/debug/memory` tests) and `test_smoke.py` (`:261`, `:273`, `:279-335`). The only two files that POST the `/embed` route — `test_speaker_embed.py:44,214` and `test_transcribe_embed_admission.py:243,258,286` — never mention `inflight_synth`.

Concrete failure: delete both `_inflight_synth += 1` (`main.py:11079`) and `_inflight_synth -= 1` (`main.py:11110`) from `/embed` and the entire sidecar suite still passes, while a test whose name asserts that coverage stays green. That is a test that cannot fail, with a name that actively misreports it. The repo already has the right pattern one file over: `test_memory.py:473 test_design_voice_counts_as_inflight_synth` observes `main._inflight_synth` inside a stubbed call and asserts `== 0` after — `/embed` needs the same.

### 🟠 Significant — the "misattribution" paragraph is wrong on evidence that is in the logs this PR already cites

`docs/testing/1996-stranded-vram-measurement.md:242`. Three separable errors, all checkable against `logs/tts.err.log` in this worktree:

1. **The arithmetic is wrong.** "The ~3.3 GB gap between P1's `allocated` (1776.7 MB) and P2's `allocated` (5453.7 MB)". 5453.7 − 1776.7 = **3677.0 MB (~3.6 GB)**. Note this cuts *against* the paragraph's own thesis: 3677 MB matches this repo's own figure for the 1.7B model ("the ~3.4 GB model", `main.py:6534`) far better than 3.3 GB does, so the attribution is fine and only the number is wrong.
2. **"the `/debug/memory` endpoint … could not represent the 1.7B model" implies the residency was unobservable. It was not.** `/health` has exposed `qwen_base17_loaded` since before this branch (`main.py:9610,9615,9645` — unchanged by this PR), and the Node side already consumes it (`server/src/routes/sidecar-health.ts:458`, `server/src/tts/ensure-sidecar-loaded.ts:260`). The real cause is narrower and more useful: the run sheet's capture snippet (lines 50-57) reads `/debug/memory`'s `engines` block and never calls `/health`, so the operator never saw a field that was already there. As written, the paragraph misdiagnoses a *procedure* defect as a *missing key* — the same category error the document is correcting.
3. **"is plausibly the Qwen 1.7B-Base model loading during session" is hedged speculation where the log is categorical.** `logs/tts.err.log` records, between P1 (07:54:32) and P2 (07:58:36): `07:57:43.326 Loading Qwen 1.7B-Base model=Qwen/Qwen3-TTS-12Hz-1.7B-Base on cuda:1`, `07:57:58.112 Qwen 1.7B-Base loaded.`, and `07:58:29.275 qwen batch synth: model=1.7b items=7 voices=3 text_len=286 load_ms=14873.9`. The paragraph cites this same file as "real forensic evidence" for the ASR half four lines earlier, then goes soft for this half. Relatedly, the run-parameters line (`:181`) still records the render as "re-rendered with `force:true` on `qwen3-tts-0.6b`" with no mention that a `model=1.7b` batch ran seven seconds before P2 — which is the whole explanation for the 3677 MB.

### 🟡 Minor

- **`:221` — the forensic paragraph is internally contradictory and misstates its own poll interval.** "a quiet streak beginning at ≈07:58:49 (after an 8 s quiet window from ~5 second polls starting after 07:58:41)". An 8 s window "starting after 07:58:41" *ends* at ≈07:58:49, so the parenthetical contradicts the clause it modifies — and `:240` uses the other reading ("07:58:49–07:58:57"). "~5 second polls" also contradicts this document's own script, which is `Start-Sleep -Seconds 2` × 5 polls (`:110-111`) — i.e. the same arithmetic pass 1's finding 7 was about.
- **`:209-213` — "fired once during the render itself" is wrong; it fired three times.** `logs/tts.err.log` has `sidecar memory crossed 8192MB` at `07:56:19.040` (`reclaimed -0MB`), `07:57:19.387` (`reclaimed 0MB`) and `07:58:19.681` (`reclaimed 0MB`) — all before P2 at 07:58:36. Three firings strengthen the paragraph's point; the stated count is simply not what the log says.
- **`main.py:8141-8146` is now false, and two drain log lines mislabel what they count.** The block comment still reads "`_inflight_synth` counts live synth calls", which is exactly the belief that produced pass 1's finding 1 — a reader now concludes ASR is *not* covered when it is. Same for the operator-facing messages: `:8624-8628` ("`%d` synth still in-flight — self-exiting anyway (the server re-renders the in-flight chapter)") and the `draining %d in-flight synth` strings at `:8713` and `:8744` can now be counting a `/transcribe` with no chapter to re-render. Also `:85` of the run sheet glosses the field as "`0` means nothing is generating", which understates the coverage the fix just bought.
- **`:25-26` prerequisites never name `logs/tts.log`,** the file the new attribution step (`:130-152`) depends on: "Start the app via `npm start` … so that `logs/server.log` and `logs/tts.err.log` capture the full timeline."
- **`base17_loaded` is only asserted in its cold/False state** (`test_memory.py:553`, `assert qwen["base17_loaded"] is False` on a fresh `QwenEngine()`). Replacing `qwen._base17 is not None` with a literal `False` at `main.py:9772` leaves the suite green — the key whose absence caused this run's misattribution has no test that can catch it being stuck. `test_qwen3.py:1838` has the same shape for `/health`'s sibling field, so the gap is consistent rather than novel, but this PR is the round that made it load-bearing.
- **The release-notes not-applicable call is recorded only in `39fdca76`'s commit message, and its rationale is narrower than the change.** It reasons from "an `inflight_synth` diagnostic field … no user-facing shippable delta", but `d4aa7a6c` also changed *recycle-drain semantics* — a recycle now waits out an in-flight ASR/embed. That is operator-visible behaviour, which is the bar `docs/release-notes-next.md` (the technical register) exists for. Either the judgement should address the drain change explicitly, or a `release-notes-next.md` line is owed.

### Verdict

**Not ready to merge.** Four of pass 1's six correctness findings are verified resolved and the open design question resolves cleanly in the change's favour. But finding 2 is still live at three sites — two of them introduced *after* its own fix, including one that the final commit quotes as authority — and the retraction that `45b913ce` exists to deliver reached the run sheet only, leaving the PR body and the #1996 record still asserting "clean idle / normal residency". Two new 🟠s besides: `/embed`'s instrumentation is untested behind a test named for it, and the corrected narrative carries three checkable errors of its own.

This is pass 2 of the 3-pass cap. Post-fix, one re-review round remains.

dudarenok-maker:
## PR review — pass 3 (head bc274fed2414a820fa42b179658514e94819eab8, depth high) — final review round per loop cap

Scope: `server/tts-sidecar/main.py`, `tests/test_smoke.py`, `tests/test_memory.py`, `docs/testing/1996-stranded-vram-measurement.md`, the PR body, the #1996 correction comment — plus the downstream record surfaces (#2656, register row A46, the register live view) that pass 1 put in scope.

**Verified, not assumed.** Every file re-read at `bc274fed` via `git show` (the primary checkout's working tree is dirty from an unrelated lane, so nothing here was read from disk). I exported the head's sidecar tree to a scratch dir and ran the real suite against the real venv:

- `test_smoke.py::test_inflight_synth_tracks_transcribe_and_embed` + `::test_health_responsive_during_busy_synth` — **2 passed**.
- **Mutation, `/embed` increment** (`main.py:11079` → comment): RED, `inflight_synth should be > 0 during embed, but got 0` at `test_smoke.py:373`.
- **Mutation, `/embed` decrement** (`main.py:11110` → `pass`): RED, `inflight_synth should be 0 after embed completes, but got 1` at `test_smoke.py:386`.
- **Mutation, `base17_loaded` → literal `False`** (`main.py:9772`): **entire sidecar suite stays GREEN** (the one failure, `test_footprints.py::test_seed_parity_with_local_llm_doc`, is a `FileNotFoundError` on `docs/local-llm.md` — an artifact of my partial export, reproduced on the unmutated tree).
- **Pass 1 finding 2's replacement instrument proven non-null**: this box's `logs/tts.log` carries 52,137 lines matching `transcribe|/embed`, in exactly the shape the run sheet describes (`INFO:     127.0.0.1:55925 - "POST /transcribe HTTP/1.1" 200 OK`, no per-line timestamp). The old `logs/server.log` grep could return nothing; this one demonstrably returns the target lines.
- `sidecar-health.ts:458` (`qwenBase17Loaded: body.qwen_base17_loaded === true`) and `/health`'s key name `qwen_base17_loaded` (`main.py:9645`) both check out — the doc's new capture-snippet field name is correct.
- All 15 required checks green on this head.

---

### Disposition — pass 1

| # | Finding | Disposition |
|---|---|---|
| 1 | 🔴 idle gate blind to ASR | **RESOLVED.** Code verified at `main.py:11000-11001/11037-11038` and `11078-11079/11109-11110`; both directions mutation-verified this pass. |
| 2 | 🔴 attribution grep read `logs/server.log` | **RESOLVED.** All three routing references retired (`:96`, `:233`, and the `:241` verbatim quote). The two surviving `server.log` mentions (`:220`, `:241`) are historical descriptions of the defect, correctly past-tense. Replacement instrument empirically proven non-null (above). |
| 3 | 🔴 `/debug/memory` missing `base17_loaded` | **RESOLVED in code** (`main.py:9772`). Test still mutation-insensitive (🟡 5). **Propagation half NOT resolved** — see 🟠 A/B. |
| 4 | 🟠 no coverage for `/debug/memory`'s `inflight_synth` | **RESOLVED.** |
| 5 | 🟠 idle-gate script always claimed success | **RESOLVED** (`:114-118`). |
| 6 | 🟠 TOCTOU between gate and P3/P4 | **RESOLVED** (`:54`, `:125`, `:165`, table column). |
| 7 | 🟡 "5 quiet polls = 10 s" | **RESOLVED** at `:88`/`:111` — but `:222` reintroduces "~5 second polls". See 🟡 2. |
| 8 | 🟡 120 s threshold absent from the procedure | **RESOLVED** (`:91-100`) — and it exposed 🟡 8. |
| 9 | 🟡 release notes | **NOT RESOLVED.** Still no entry and still no explicit not-applicable in the PR body. See 🟡 7. |

### Disposition — pass 2

| # | Finding | Disposition |
|---|---|---|
| 🔴 A | 3 residual `logs/server.log` routing references | **RESOLVED** (verified line by line). |
| 🔴 B | retraction reached 1 of 3 surfaces | **RESOLVED for the two named surfaces.** PR body rewritten (incl. the stale "no source file outside…" bullet, now gone); the [#1996 correction comment](https://github.com/dudarenok-maker/Castwright/issues/1996#issuecomment-5419223179) is accurate and complete. **But the retraction stopped there** — see 🟠 A/B. |
| 🟠 C | `/embed` untested behind a test named for it | **RESOLVED.** Mutation-verified in both directions. |
| 🟠 D | misattribution paragraph, 3 checkable errors | Errors 1–3 **RESOLVED** (3677.0 MB ✓; `/health`'s pre-existing `qwen_base17_loaded` named ✓; hedge replaced with the categorical log lines ✓). The "relatedly" clause — the run-parameters line still reading 0.6B-only — **NOT resolved**, see 🟠 C. New: the fix introduced 🟡 1. |
| 🟡 1 | `:222` forensic paragraph self-contradictory, "~5 second polls" | **NOT RESOLVED** — text unchanged. |
| 🟡 2 | `:212` "fired once" (log shows three) | **NOT RESOLVED** — text unchanged. |
| 🟡 3 | `main.py:8141-8146` comment + drain log strings + run sheet `:85` gloss | **Comment RESOLVED** (`bc274fed`). Drain strings and the run-sheet gloss **NOT** — see 🟡 6. |
| 🟡 4 | prerequisites never name `logs/tts.log` | **NOT RESOLVED** (`:25-26` unchanged). |
| 🟡 5 | `base17_loaded` mutation-insensitive | **NOT RESOLVED** — confirmed empirically this pass. |
| 🟡 6 | release-notes rationale narrower than the change | **NOT RESOLVED.** |

---

### 🟠 A — #2656 was closed as COMPLETED by a docs-only wording PR, while the measurement it owns has not been performed

`gh issue view 2656` → `state: CLOSED`, `stateReason: COMPLETED`, `closedAt: 2026-08-26T01:06:39Z`. The timeline names the closer: **PR #2664** ("docs(docs): correct A46's resident-model count and re-run instructions"), merged `01:06:38Z`, touching exactly one file (`docs/testing/onbox-acceptance-register.md`), body ending `Closes #2656`. One second later the issue closed.

#2664 exists *only* as this round's discharge of pass 1 finding 3's A46 chore. It corrected a register row's wording. It did not run a GPU measurement — and this branch's own run sheet says so in as many words (`docs/testing/1996-stranded-vram-measurement.md:247`): *"A valid re-run is still owed… This is real-hardware work that was not performed as part of this PR-review-gate fix round — no GPU re-run has been executed; only this narrative correction has landed."*

**Concrete failure, already in effect.** The acceptance debt is now invisible on the board: #2656 reads as done. Meanwhile `docs/testing/onbox-acceptance-register.md:2761` still carries A46 as an **owed** row and names #2656 as its *Criteria:* home — so the register points at a closed-as-completed issue for criteria that were never met. Either the register row is wrong or the issue state is; they cannot both be right. `Refs #2656` was the correct trailer for a wording fix.

Fix: reopen #2656, and — since the closing keyword lives in a merged PR body — say why in a comment so a future re-merge doesn't silently re-close it.

### 🟠 B — the retraction never reached the published live view, and only partly reached A46

CLAUDE.md before-shipping step 3 requires the register, the run sheet **and the live view** to move in the same diff. #2664 moved only the `.md`. `docs/testing/onbox-acceptance-register-live-view.html:917-923` on `origin/main` — the file published to the URL in the register's header — is still **wholly pre-retraction**:

- "the 2026-08-25 idle-gated measurement (`…measurement.md` @ `218cc562` …) found the ~5.45 GB `allocated` … is byte-identical across a **confirmed-idle** 21 s window" — `218cc562` is the pre-retraction commit;
- "That reading is consistent with 'this is just the **two resident models**' own live weights, not a leak'";
- reproduce step: "confirm the box idle (poll `inflight_synth`, not a fixed wall-clock)" — no ASR caveat at all;
- "**New step:** issue `POST /unload {qwen}` and confirm/force **Whisper** past its `ASR_IDLE_TTL`" — two models, `base17` absent;
- "matching **Qwen Base 0.6B + Whisper**'s known weight sizes"; *Needs:* "a live sidecar with **Qwen Base 0.6B and Whisper** both resident".

That is the retracted conclusion, verbatim and uncaveated, on the artifact a reader is most likely to open.

`docs/testing/onbox-acceptance-register.md:2761-2805` is better — #2664 added the two-instrument-bug paragraph and the three-model unload step — but four residuals survive:

1. `:2788-2789` — "note the poll **is** ASR-blind, so verify via logs that no `/transcribe`/`/embed` was active." Present tense. It becomes false the moment this PR merges, which is the event this comment gates.
2. `:2773-2775` — "the 'confirmed-idle' 21 s window **may have** overlapped live Whisper transcription." The run sheet at this head is categorical (`:222`, `:241`): it *did*, 07:58:31–51.768 vs 07:58:49–57, from `logs/tts.err.log`. Hedging where the evidence is settled is what pass 2's 🟠 D.3 was about, one file over.
3. `:2764` — cites the run sheet `@ 45b913ce`, two commits stale (`023f647c` and `bc274fed` both changed it).
4. `:2793-2794` — "Diff that post-unload `allocated`/`reserved` against **the P3 baseline already on record** for this box's device." The corrected run sheet says of exactly those readings: *"do not treat the resulting P3/P4 readings as evidence about a stranded pool at all"* (`:233`, `:241`). A46 prescribes diffing against a baseline this PR invalidates — a procedural error, not a wording one.

**Concrete failure:** the owed re-run is briefed from A46 and its live view, not from this branch's run sheet. An operator (or a cold agent, per the execution model's "brief from the ticket") reads the live view, skips the ASR verification and the `base17` unload, and diffs against the invalidated P3 baseline. That reproduces this run's two defects exactly — the "third attempt down the same path" the run sheet's own closing sentence (`:251-253`) names as the hazard.

### 🟠 C — the run sheet's results table and run-parameters line still record the two-model, 0.6B-only picture the same file retracts

`docs/testing/1996-stranded-vram-measurement.md:182` — "re-rendered with `force:true` on `qwen3-tts-0.6b`", with no mention of the `model=1.7b` batch the same file quotes sixty lines later (`:243`: `07:58:29.275 qwen batch synth: model=1.7b items=7 voices=3`, seven seconds before P2).

`:189-190` — P2 and P3's **Engines loaded** cells read `qwen base, whisper`. The narrative at `:243` says three models were resident.

The `inflight_synth` column was honestly caveated (`n/a — pre-fix run`) in every row. The `Engines loaded` column, produced by the *other* blind instrument, was not — and it is the unflagged one. That asymmetry is not hypothetical: "the two resident models" appears in #2656's body, in A46's pre-#2664 text, and in the live view today. All three were copied from this table.

Pass 2 raised this as 🟠 D's "relatedly" clause; the three enumerated errors were fixed and this one was not. Fix: caveat the two cells the way the `inflight_synth` column is caveated, and note the 1.7B batch on the run-parameters line.

---

### 🟡 Minor

1. **`:243` was falsified by its own commit.** `023f647c` added `$health = Invoke-RestMethod …/health` to the capture snippet (`:53-54`) **and**, in the same diff, the sentence "The real issue is that the capture snippet in step 'The capture' **only calls `/debug/memory`, never `/health`**". The snippet 190 lines above now calls both. Past-tense it ("the snippet *at the time of the run* only called…").
2. **`:222` unchanged from pass 2.** "a quiet streak beginning at ≈07:58:49 (after an 8 s quiet window from ~5 second polls starting after 07:58:41)" — an 8 s window starting after 07:58:41 *ends* at ≈07:58:49, and "~5 second polls" contradicts this document's own `Start-Sleep -Seconds 2` (`:112`).
3. **`:212` unchanged from pass 2.** "fired once during the render itself" — `logs/tts.err.log` has `sidecar memory crossed 8192MB` at 07:56:19.040, 07:57:19.387 and 07:58:19.681, all before P2.
4. **`:25-26` unchanged from pass 2.** Prerequisites still name `logs/server.log` and `logs/tts.err.log` and never `logs/tts.log` — the file the attribution step at `:131-155` now depends on. This is also the fourth surviving `server.log` reference in the file, in the one place that tells the operator which logs must exist.
5. **`base17_loaded` remains mutation-insensitive — confirmed empirically.** `test_memory.py:553` asserts only `qwen["base17_loaded"] is False` on a cold engine. Replacing `qwen._base17 is not None` with literal `False` at `main.py:9772` leaves the whole sidecar suite green. The key whose absence caused this run's misattribution has no test that can catch it stuck at `False` — and `42dddeb8` is the commit that made it load-bearing.
6. **Drain log strings still say "synth" only.** `main.py:8616-8617` ("if the grace expires with **synth** still running we exit anyway — the server's srv-17c in-worker recovery **re-renders that chapter**"), `:8625-8627` ("`%d` **synth** still in-flight"), `:8713` and `:8744` ("draining `%d` in-flight **synth**"). After `d4aa7a6c` the counted item can be a `/transcribe` with no chapter to re-render. The block comment at `:8141-8146` was correctly fixed in `bc274fed`; these four were not. Same for the run sheet at `:86-87` — "`inflight_synth` (an int; `0` means nothing is **generating**)" understates the coverage the fix bought.
7. **Release-notes gate still neither satisfied nor waived in the PR body.** Pass 1 🟡 and pass 2 🟡 both raised it. The judgement exists only in `39fdca76`'s commit message, and its rationale ("a diagnostic field, no user-facing shippable delta") still does not address `d4aa7a6c`'s change to recycle-drain semantics — a recycle now waits out an in-flight ASR/embed, which is operator-visible and is what `docs/release-notes-next.md` exists for. CLAUDE.md step 5 wants the skip stated, not omitted.
8. **`:133` misdescribes the decision table's fourth row.** "that's the confound the decision tree's fourth row keys on" — row 4 (`:233`) keys on "**Idle gate took >120 s to clear**", not on the `tts.log` diff. Relatedly, the verdict at `:237-239` assigns this run to row 4, but this run's gate cleared in **21 s**. The verdict's substance is right; the row it claims is one the run's own stated criterion fails. (The gate fix makes ">120 s" the *correct* key for ASR contamination going forward, so the gap is narrow — a fast clear with a dirty `tts.log`, i.e. GPU work the counter still cannot see — but `:133` promises a row that does not exist and the verdict cites one it does not satisfy.)
9. **`bc274fed`'s commit message reads `Fixes #2655`** — a closing keyword aimed at this PR itself rather than at an issue. Harmless on GitHub (keywords only close issues) but it lands in the merge history as a self-reference; `Refs #1996` was intended.

---

### ✅ What is solid at this head

- **The code change is done and correctly tested.** `/transcribe` and `/embed` increment before the `try` and decrement in `finally` on every exit path; the `/embed` half is now genuinely mutation-sensitive in both directions, verified here rather than taken on the commit message's word. The stale `_inflight_synth` block comment is fixed.
- **The replacement attribution instrument actually works** — the single most important thing to re-check after pass 1 finding 2, and it holds: `logs/tts.log` carries real `POST /transcribe` access lines in the documented format, and `(Get-Content …).Count` is the right line-count form for this repo (`Measure-Object -Line` would have skipped blanks).
- **The idle gate and the log diff cover each other's blind spots**: a long transcribe spanning the whole window logs nothing in the diff (uvicorn writes its access line on completion), but the `inflight_synth` primary signal catches it — and `:131` correctly names the gate as primary and the grep as corroboration.
- **The retraction itself is accurate wherever it landed.** The PR body, the run sheet's "Reading the result" section, and the #1996 comment all say the same true thing, with the arithmetic (3677.0 MB) and the evidence citations checked and correct.

---

### Verdict

**3 correctness findings (0 🔴, 3 🟠) + 9 🟡.** None is in the sidecar code — `main.py` and both test files are in good shape at this head, and every pass-1/pass-2 finding against the *code* is verified resolved.

All three 🟠 share one root cause with three different fixes: **the retraction was completed on the surfaces this PR touches and stopped at the boundary of the ones it does not.** The published live view still states the withdrawn conclusion verbatim; A46 prescribes diffing against a baseline this PR invalidates; and #2656 — the live successor both closed issues redirect readers to — is now closed as COMPLETED for work that has not been done.

**This is pass 3 of the 3-pass cap. No further automatic re-review will run.** The remaining work is bookkeeping across `origin/main` and GitHub state rather than code on this branch, so it need not block this branch's merge — but it does have to happen, and after the cap that is a call for the shipping session and the repo owner to make explicitly rather than one more automatic loop.

Issue #2659, opened 2026-08-25
[agent instructions][claude][task] Add Qwen duration golden-audio baseline scaffold


dudarenok-maker:
## What

Add a per-line **duration** golden-audio baseline for **Qwen** (the default
synth engine, `main.QwenEngine`, 0.6B-Base model) — mirroring the pattern
`test_golden_regression.py` already uses for Kokoro, but scoped to duration
only (no Whisper content check — that is not asked for by #1994). This
closes the visibility gap: today only Kokoro (the fallback engine) has any
golden-audio coverage; a Qwen speech-rate/prosody regression is invisible to
every automated gate.

This child writes the **scaffold**: real test file + fixture + an
**unblessed** baseline (empty `entries`, placeholder `tolerance`), exactly
the state `kokoro-baseline.json` shipped in before its first bless. It does
**not** run N real Qwen syntheses to measure the true stochastic spread or
derive a validated tolerance — that requires a real GPU box with an
already-designed Qwen voice, which this agent does not have. That
measurement step is captured as on-box acceptance debt (see the register row
below), per this repo's CLAUDE.md rule that owed acceptance never blocks a
merge — it converts into a row.

## Background you need

- `server/tts-sidecar/tests/golden/test_golden_regression.py` — the pattern
  to mirror for the DURATION half only: `_load_json`, the "unblessed to
  skip" branch (lines ~284-289, `if not entries: pytest.skip(...)`), the
  per-line loop asserting `res.substituted_from is None`, `rms(res.pcm) >=
  MIN_RMS`, and `compare_to_baseline(measured, base, tol=tol)` (lines
  ~291-316). Do NOT copy the Whisper/content-drift half
  (`test_kokoro_golden_content_matches_baseline`, `_make_whisper`) — #1994
  does not ask for Qwen content coverage, only duration.
- `server/tts-sidecar/tests/golden/compare.py` lines ~82-149 — `measure_pcm`,
  `rms`, `compare_to_baseline` are generic (PCM-format agnostic) and already
  import-light/pure. Reuse them directly; do not duplicate.
- `server/tts-sidecar/tests/golden/prereq.py` line 80 — `synthesise_or_skip(engine,
  model, voice, text)`. Use it for every real synth call in the new golden
  test (SKIP only when the Qwen engine package itself is absent from this
  box; everything else must fail, per this file's own header doc).
- `server/tts-sidecar/tests/golden/test_cross_engine_sanity.py` lines ~126-139 —
[...]

PR #2662, opened 2026-08-26
chore(scripts): add oe-detached-commit.ps1, close the missing -WindowStyle Hidden gap

### LINKED ISSUE #2661: chore(ops): OE lane commit launcher dropped -WindowStyle Hidden, popping visible console windows
## What

A `cline-free` Open Engine lane working ticket #2659 (2026-08-26, worktree
`wt-1994-qwen-duration-baseline`) wrote its own inline `Start-Process`
launcher for the "commit detached" recipe documented in
`.clinerules/cline.md`, and dropped `-WindowStyle Hidden`. Result: a visible
PowerShell console popped up for the commit itself, and a second one for the
husky pre-commit hook it triggers (`npm run verify:fast:scoped`). The
freelanced script also called a nonexistent `Delete-Item` cmdlet (the real
one is `Remove-Item`).

The underlying gap: the recipe in `.clinerules/cline.md` already tells lanes
to use `-WindowStyle Hidden` and to "copy the recipe verbatim, do not
improvise" -- but it was still a copy-pasted snippet, not something enforceable.

## Fix

- Added `scripts/oe-detached-commit.ps1`, a tracked script (ships in every
  worktree) that does the launch: writes the commit message to a BOM-less
  file, launches `git commit` via `Start-Process -WindowStyle Hidden`, hands
  back the scratch dir to poll. `-Message` is a bound parameter rather than
  text interpolated into a command string, so callers no longer need to
  escape apostrophes in the subject either (closing the `#2520` quoting-bug
  class the same recipe already documented).
- Paired Pester coverage in `scripts/tests/oe-detached-commit.Tests.ps1`:
  static-shape guards (asserts `-WindowStyle Hidden` is present, asserts no
  `Delete-Item`, asserts `-Message` isn't interpolated into `commit -m`) plus
  a functional smoke test that runs a real detached commit against a scratch
  git repo and asserts the message lands byte-for-byte, apostrophe included.
- Rewrote `.clinerules/cline.md`'s "commit detached" section to call the
  script instead of hosting the freelance-prone inline snippet.

## Acceptance

- `npm run test:scripts` green, including the new Pester file.
- `.clinerules/cline.md` points lanes at `scripts/oe-detached-commit.ps1`
  instead of an inline recipe.

dudarenok-maker:
## Summary
- OE lanes were freelancing their own inline `Start-Process` launchers for the "commit detached" recipe in `.clinerules/cline.md`. A `cline-free` lane on ticket #2659 (2026-08-26) dropped `-WindowStyle Hidden`, popping a visible PowerShell window per commit (and per pre-commit-hook run it triggers).
- Added `scripts/oe-detached-commit.ps1`: a tracked, real script (ships in every worktree) that launches `git commit` detached and hidden. `-Message` is a bound parameter rather than interpolated into a command string, so callers no longer need to escape apostrophes either — closing the `#2520` quoting-bug class the same recipe already documented.
- Rewrote `.clinerules/cline.md`'s "commit detached" section to call the script instead of hosting a copy-pasted, freelance-prone inline snippet.
- Three review rounds (see PR comments) found and fixed real bugs in the fix itself: a `$env:TEMP` crash on Linux CI, a static guard test that greped the whole file (so it couldn't fail when the real `-WindowStyle Hidden` was deleted), a same-second scratch-directory collision that could swap which commit message lands in which repo, a trailing-backslash-on-`-Worktree` quoting bug, and a dangling `$W` reference left in the rewritten doc after round 1's edit. Round 2 also found its own fix had reintroduced the identical grep blind spot for two new checks — converted both to AST-based assertions. Round 3 was clean.
- Skipped release notes: internal agent-tooling chore, no user- or operator-visible effect.
- Skipped a `docs/features/` plan doc: small/localized change, the issue body + paired tests are the spec.

## Test plan
- [x] `npm run test:scripts` green in the worktree: **78 passed**, 2 pre-existing venv-bootstrap skips unrelated to this change (0 failed)
- [x] Static AST-based guards: `-WindowStyle Hidden` (and its `-WindowStyle:Hidden` / quoted-`'Hidden'` equivalents) is bound on the real `Start-Process` call; `-Message` never reaches `commit -m`; the message is written without a BOM; `$Worktree = $Worktree.TrimEnd('\')` and `[Guid]::NewGuid()` are both real, executed AST nodes, not just text present somewhere in the file
- [x] Functional smoke tests (Windows-only, `-Skip:($env:OS -ne 'Windows_NT')`): a real detached commit lands byte-for-byte including an apostrophe; a trailing backslash on `-Worktree` no longer breaks the launch; two same-second launches get independent scratch dirs and each commit lands in the right repo
- [x] Self-verified every guard against the exact mutation it exists to catch (deleting the real flag/line while leaving the surrounding comment) — each fails closed; restored before committing
- [x] Dogfooded: every commit on this branch, from the first onward, was made via the script under review

Closes #2661

### DIFF
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -162,67 +162,44 @@ before 20.** Recorded on this repo: `test:server` **518.8 s** and `test`
 vitest run can report far less wall-clock on an idle box, so treat these as the
 range, not a constant.
 
-**Copy the recipe below verbatim -- do not improvise your own inline variant.**
-It avoids shell-quoting entirely by design: the commit message is written to a file with `[IO.File]::WriteAllText`, and the child script receives `$Dir` and `$Worktree` as parameters via its `param()` block rather than having anything interpolated into a command string. A freelanced one-liner that rebuilds this by hand -- embedding `$variables` or escaped `\"` quotes inside a wrapped double-quoted command -- reintroduces exactly the quoting bug this recipe exists to avoid. On `Castwright#2520` a local-model run did this, hit `The string is missing the terminator` / `Missing type name after '['` / `Exception calling "Create" with "1" argument(s)` across four self-invented variants, spent its whole turn reasoning line-by-line about *why* each one broke, and died on the output-token cap with no commit, no push, and no receipt -- while the two-file edit it was committing was already correct. **Two things require substitution per-use:** your worktree path (replace the placeholder in the `$W` assignment) and your commit message (replace `'fix(scope): subject line'` in the `[IO.File]::WriteAllText` line with a real message following Conventional Commits). **Important:** if your commit subject contains a single quote (e.g. `don't clobber X`), escape it by doubling it inside the PowerShell single-quoted string (e.g. `'don''t clobber X'`), so the quote is preserved in the final commit message and does not break the PowerShell string. The structure and quoting style are copied verbatim; the paths and message are filled in for each use.
-
-**If the detached commit command reports `The string is missing the terminator` (sometimes preceded by `Missing ')' in method call.` from the same root cause), do not diagnose it.** This is almost always an unescaped single quote in the commit message: fix it by doubling the quote (e.g. `'don''t'` for a subject containing "don't"), then retry the recipe with the same structure. Stop reasoning about it after one sentence -- that reasoning is usually wrong anyway (as it was on #2520) and it is the expensive way to fail. **If the retry produces the identical error, stop retrying** and say so once with the exact command and the error, per "Do not end your turn while it runs" below.
-
-**`Missing type name after '['` is a DIFFERENT case -- not the apostrophe bug above.** It means the command being run is not this recipe at all, but an improvised variant of it (this is exactly what happened on #2520: four self-invented wrapper attempts, none of them this recipe). The fix is not to retry that broken command -- it is to discard it and copy the recipe below verbatim instead.
-
-Note: `Exception calling "Create" with "1" argument(s)` is not a signal on its own -- it is a generic .NET method-call exception wrapper. It can wrap either error above, or an unrelated failure entirely (permission denied, missing path). Read the full message it wraps before deciding which case you are in.
-
-Launch it detached and poll. Each poll returns instantly, so the cap stops
-mattering:
+**Call `scripts/oe-detached-commit.ps1` -- do not freelance your own inline
+Start-Process variant.** It lives in every worktree (it's a tracked file, so
+`git worktree add` brings it along) and does what the old inline recipe did,
+but as a real script instead of a copy-pasted snippet a model can misremember:
+writes the message to a BOM-less file, launches `git commit` via
+`Start-Process -WindowStyle Hidden`, and hands back the scratch dir to poll.
+Because `-Message` is a bound **parameter** rather than text interpolated into
+a command string, **no apostrophe escaping is needed** -- pass the message
+exactly as written, single quotes and all:
 
 ```powershell
-# Per-run directory. A FIXED log path is read by the next poll as the PREVIOUS
-# run's result: Start-Process returns before the child truncates the file, so a
-# stale EXIT=0 reports success for a commit that is still running.
-$run = Get-Date -Format 'yyyyMMdd-HHmmss'
-$T   = Join-Path $env:TEMP "cw-commit-$run"
-$W   = 'C:\Claude\Projects\wt-<your-worktree>'
-New-Item -ItemType Directory -Path $T -Force | Out-Null
-
-# NO BOM. PS 5.1's `Set-Content -Encoding utf8` writes one, `git commit -F` does
-# NOT strip it, and scripts/validate-commit-msg.mjs then rejects the subject --
-# AFTER you have paid the whole battery, with an error echoing a subject that
-# looks perfectly valid because the BOM is invisible.
-[IO.File]::WriteAllText("$T\msg.txt", 'fix(scope): subject line', (New-Object System.Text.UTF8Encoding $false))
+$T = & 'C:\Claude\Projects\wt-<your-worktree>\scripts\oe-detached-commit.ps1' `
+       -Worktree 'C:\Claude\Projects\wt-<your-worktree>' `
+       -Message "fix(scope): subject line, don't escape apostrophes here"
 ```
 
-The child script takes its paths as parameters, so nothing has to be
-interpolated into the here-string and `$LASTEXITCODE` resolves in the child:
+Two failure modes this script exists to close, both hit by freelanced
+variants in the past -- **do not rebuild either by hand**:
 
-```powershell
-@'
-param([string]$Dir, [string]$Worktree)
-$ErrorActionPreference = 'Continue'
-git -C $Worktree commit -F (Join-Path $Dir 'msg.txt') *>&1 |
-  Out-File -FilePath (Join-Path $Dir 'commit.log') -Encoding utf8
-"EXIT=$LASTEXITCODE" | Out-File -FilePath (Join-Path $Dir 'commit.log') -Append -Encoding utf8
-'@ | Set-Content "$T\commit.ps1" -Encoding utf8
-```
+- **Dropping `-WindowStyle Hidden`.** A `cline-free` lane on `Castwright#2659`
+  (2026-08-26) wrote its own launcher without it, popping a visible
+  PowerShell window per commit -- and per pre-commit hook run that commit
+  triggers. The script always passes the flag; if a window is still showing,
+  something bypassed the script rather than a flag being missing from it.
+- **Quoting the message by hand.** On `Castwright#2520` a local-model run
+  built the git command as an interpolated string with an unescaped `'`, hit
+  `The string is missing the terminator` / `Missing type name after '['` /
+  `Exception calling "Create" with "1" argument(s)` across four
+  self-invented variants, spent its whole turn reasoning line-by-line about
+  *why* each one broke, and died on the output-token cap with no commit, no
+  push, and no receipt -- while the two-file edit it was committing was
+  already correct. Passing `-Message` as a parameter makes this class of bug
+  impossible: if you see one of those errors, stop reasoning about it after
+  one sentence and re-check you're actually calling the script, not a
+  hand-built variant of it.
 
-**That closing `'@` must be at column zero** -- indent it (as happens
-automatically inside a markdown list item) and PS 5.1 fails with
-`WhitespaceBeforeHereStringFooter` before running anything, so no log is ever
-created and a naive poll waits forever. Same family as the `.ps1` traps under
-"Never do these".
-
-```powershell
-# Clear the log FIRST. Start-Process returns before the child truncates it, so
-# on a RETRY inside the same $T the very first poll reads the PREVIOUS
-# attempt's sentinel and reports a verdict for a run that is still going.
-Remove-Item "$T\commit.log" -ErrorAction SilentlyContinue
-
-# Quote each path argument: -ArgumentList does NOT quote its elements.
-$p = Start-Process powershell -WindowStyle Hidden -PassThru -ArgumentList @(
-  '-NoProfile','-ExecutionPolicy','Bypass','-File',"`"$T\commit.ps1`"",
-  '-Dir',"`"$T`"",'-Worktree',"`"$W`"")
-$p.Id | Set-Content "$T\commit.pid"
-```
-
-Then poll until it resolves. **Check the process as well as the sentinel** --
+Then poll until it resolves. Each poll returns instantly, so the output-token
+cap stops mattering. **Check the process as well as the sentinel** --
 those two disagreeing is how you learn the child died at parse time instead of
 waiting on it forever:
 
--- a/.clinerules/cline.md
+++ b/.clinerules/cline.md
@@ -173,10 +173,13 @@ a command string, **no apostrophe escaping is needed** -- pass the message
 exactly as written, single quotes and all:
 
 ```powershell
-$T = & 'C:\Claude\Projects\wt-<your-worktree>\scripts\oe-detached-commit.ps1' `
-       -Worktree 'C:\Claude\Projects\wt-<your-worktree>' `
+$W = 'C:\Claude\Projects\wt-<your-worktree>'
+$T = & "$W\scripts\oe-detached-commit.ps1" -Worktree $W `
        -Message "fix(scope): subject line, don't escape apostrophes here"
 ```
+
+**Keep `$W` around** -- the poll steps below and the tiebreaker
+(`git -C $W log --oneline -1`) both use it.
 
 Two failure modes this script exists to close, both hit by freelanced
 variants in the past -- **do not rebuild either by hand**:
@@ -194,9 +197,11 @@ variants in the past -- **do not rebuild either by hand**:
   *why* each one broke, and died on the output-token cap with no commit, no
   push, and no receipt -- while the two-file edit it was committing was
   already correct. Passing `-Message` as a parameter makes this class of bug
-  impossible: if you see one of those errors, stop reasoning about it after
-  one sentence and re-check you're actually calling the script, not a
-  hand-built variant of it.
+  impossible. **If you see one of those errors anyway, stop reasoning about
+  it after one sentence and re-check you're actually calling the script, not
+  a hand-built variant of it -- and if a retry produces the identical error
+  even after confirming that, stop retrying** and say so once with the exact
+  command and the error, per "Do not end your turn while it runs" below.
 
 Then poll until it resolves. Each poll returns instantly, so the output-token
 cap stops mattering. **Check the process as well as the sentinel** --
--- a/scripts/oe-detached-commit.ps1
+++ b/scripts/oe-detached-commit.ps1
@@ -1,0 +1,82 @@
+<#
+.SYNOPSIS
+  Launches `git commit` detached and hidden, for headless/OE agent lanes.
+
+.DESCRIPTION
+  Replaces the freelanced inline Start-Process snippets OE lanes have been
+  writing per-run (documented in .clinerules/cline.md "commit detached"
+  recipe). Two failure modes that recipe already warns against, and that a
+  freelanced variant hit on 2026-08-26 (Castwright open-engine ticket #2659):
+    - dropping `-WindowStyle Hidden`, which pops a visible PowerShell window
+      for the commit itself AND for the husky pre-commit hook it triggers
+    - reintroducing commit-message quoting bugs (unescaped `'` in the subject)
+      by building the git command as an interpolated string
+
+  Call this script directly with -Message as a bound parameter (never build
+  a command string containing the message) so quoting is a non-issue: no
+  escaping of apostrophes is needed by the caller.
+
+.PARAMETER Worktree
+  Absolute path to the worktree/checkout to commit in.
+
+.PARAMETER Message
+  Full commit message (subject, optionally with a blank line + body).
+
+.OUTPUTS
+  The scratch directory path (also written nowhere else) -- poll it per the
+  "commit detached" recipe in .clinerules/cline.md:
+    - alive:  Get-Process -Id (Get-Content "$T\commit.pid") -ErrorAction SilentlyContinue
+    - done:   Test-Path "$T\commit.log" -and it contains a line starting "EXIT="
+    - result: Get-Content "$T\commit.log" once done is true and alive is false
+
+.EXAMPLE
+  $T = & C:\Claude\Projects\wt-1994-qwen-duration-baseline\scripts\oe-detached-commit.ps1 `
+         -Worktree 'C:\Claude\Projects\wt-1994-qwen-duration-baseline' `
+         -Message "test(sidecar): add Qwen duration golden-audio baseline scaffold (#2659)"
+#>
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory)][string]$Worktree,
+    [Parameter(Mandatory)][string]$Message
+)
+
+$ErrorActionPreference = 'Stop'
+
+$run = Get-Date -Format 'yyyyMMdd-HHmmss'
+$T = Join-Path $env:TEMP "cw-commit-$run"
+New-Item -ItemType Directory -Path $T -Force | Out-Null
+
+# No BOM: PS 5.1's default utf8 write adds one, `git commit -F` does not
+# strip it, and scripts/validate-commit-msg.mjs then rejects a subject that
+# LOOKS fine (the BOM is invisible) -- after the whole verify battery has
+# already paid.
+[IO.File]::WriteAllText((Join-Path $T 'msg.txt'), $Message, (New-Object System.Text.UTF8Encoding $false))
+
+# Paths come in via param() rather than string interpolation, so nothing
+# about $Worktree or the scratch dir can reintroduce a quoting bug either.
+$childScript = @'
+param([string]$Dir, [string]$Worktree)
+$ErrorActionPreference = 'Continue'
+git -C $Worktree commit -F (Join-Path $Dir 'msg.txt') *>&1 |
+  Out-File -FilePath (Join-Path $Dir 'commit.log') -Encoding utf8
+"EXIT=$LASTEXITCODE" | Out-File -FilePath (Join-Path $Dir 'commit.log') -Append -Encoding utf8
+'@
+Set-Content -Path (Join-Path $T 'commit.ps1') -Value $childScript -Encoding utf8
+
+# Clear the log FIRST: Start-Process returns before the child truncates it,
+# so a poll against a stale file from a prior run in the same $T would
+# report a verdict for a run that is still going. ($T is timestamped fresh
+# above, so this is a defensive no-op today -- kept because a caller could
+# reuse a $T.)
+Remove-Item (Join-Path $T 'commit.log') -ErrorAction SilentlyContinue
+
+# -WindowStyle Hidden is the whole point of this script: a freelanced
+# Start-Process call that omits it pops a visible console for the commit
+# AND for the husky pre-commit hook `git commit` triggers.
+# -ArgumentList does NOT quote its elements, so each path arg is quoted here.
+$p = Start-Process powershell -WindowStyle Hidden -PassThru -ArgumentList @(
+    '-NoProfile', '-ExecutionPolicy', 'Bypass', '-File', "`"$T\commit.ps1`"",
+    '-Dir', "`"$T`"", '-Worktree', "`"$Worktree`"")
+$p.Id | Set-Content (Join-Path $T 'commit.pid')
+
+Write-Output $T
--- a/scripts/oe-detached-commit.ps1
+++ b/scripts/oe-detached-commit.ps1
@@ -5,19 +5,27 @@
 .DESCRIPTION
   Replaces the freelanced inline Start-Process snippets OE lanes have been
   writing per-run (documented in .clinerules/cline.md "commit detached"
-  recipe). Two failure modes that recipe already warns against, and that a
-  freelanced variant hit on 2026-08-26 (Castwright open-engine ticket #2659):
+  recipe). Failure modes that recipe already warns against, and that
+  freelanced variants hit on 2026-08-26 (Castwright open-engine ticket #2659)
+  and in review of this very script (PR #2662):
     - dropping `-WindowStyle Hidden`, which pops a visible PowerShell window
       for the commit itself AND for the husky pre-commit hook it triggers
     - reintroducing commit-message quoting bugs (unescaped `'` in the subject)
       by building the git command as an interpolated string
+    - a trailing backslash on a quoted path argument escaping the closing
+      quote instead of terminating the path (Win32 argv quoting: a backslash
+      run immediately before a `"` is only literal if doubled)
+    - a same-second scratch dir colliding between two concurrently-launched
+      commits and silently mixing up which message lands where
 
   Call this script directly with -Message as a bound parameter (never build
   a command string containing the message) so quoting is a non-issue: no
   escaping of apostrophes is needed by the caller.
 
 .PARAMETER Worktree
-  Absolute path to the worktree/checkout to commit in.
+  Absolute path to the worktree/checkout to commit in. A trailing backslash
+  is tolerated (stripped before quoting) -- it would otherwise escape the
+  closing quote in the launched command line.
 
 .PARAMETER Message
   Full commit message (subject, optionally with a blank line + body).
@@ -30,8 +38,8 @@
     - result: Get-Content "$T\commit.log" once done is true and alive is false
 
 .EXAMPLE
-  $T = & C:\Claude\Projects\wt-1994-qwen-duration-baseline\scripts\oe-detached-commit.ps1 `
-         -Worktree 'C:\Claude\Projects\wt-1994-qwen-duration-baseline' `
+  $W = 'C:\Claude\Projects\wt-1994-qwen-duration-baseline'
+  $T = & "$W\scripts\oe-detached-commit.ps1" -Worktree $W `
          -Message "test(sidecar): add Qwen duration golden-audio baseline scaffold (#2659)"
 #>
 [CmdletBinding()]
@@ -42,8 +50,18 @@ param(
 
 $ErrorActionPreference = 'Stop'
 
+# Strip a trailing backslash: a quoted "...\" argument has its closing quote
+# escaped by the preceding backslash under Win32 argv parsing, which folds
+# the rest of the command line into this one argument instead of ending it.
+$Worktree = $Worktree.TrimEnd('\')
+
+# GUID-suffixed, not just second-resolution timestamped: two lanes launching
+# in the same second (the normal case for parallel worktree lanes, which is
+# the intended usage of this script) would otherwise collide on one scratch
+# dir and each poll the other's commit -- reporting EXIT=0 for a commit that
+# is not theirs, with no error at all.
 $run = Get-Date -Format 'yyyyMMdd-HHmmss'
-$T = Join-Path $env:TEMP "cw-commit-$run"
+$T = Join-Path ([System.IO.Path]::GetTempPath()) "cw-commit-$run-$([Guid]::NewGuid().ToString('N').Substring(0, 8))"
 New-Item -ItemType Directory -Path $T -Force | Out-Null
 
 # No BOM: PS 5.1's default utf8 write adds one, `git commit -F` does not
@@ -53,7 +71,8 @@ New-Item -ItemType Directory -Path $T -Force | Out-Null
 [IO.File]::WriteAllText((Join-Path $T 'msg.txt'), $Message, (New-Object System.Text.UTF8Encoding $false))
 
 # Paths come in via param() rather than string interpolation, so nothing
-# about $Worktree or the scratch dir can reintroduce a quoting bug either.
+# about $Worktree or the scratch dir can reintroduce a quoting bug in the
+# CHILD script's own source (it never contains a literal path or message).
 $childScript = @'
 param([string]$Dir, [string]$Worktree)
 $ErrorActionPreference = 'Continue'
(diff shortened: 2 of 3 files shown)


dudarenok-maker:
## PR review — pass 1 (head 06fdb406, depth low)

Scope: all 3 changed files — `scripts/oe-detached-commit.ps1` (+82), `scripts/tests/oe-detached-commit.Tests.ps1` (+71), `.clinerules/cline.md` (+34/−57). Depth `low` (single-scope `chore`), but `.clinerules/` earns full scrutiny of the instructions themselves per the gate's own carve-out.

Verified before probing: read the CI run for head `06fdb406` (`verify.yml` **red**), ran the new Pester file against an out-of-tree **mutant** of the script, and executed the real script end-to-end against three throwaway git repos (trailing-backslash path, concurrent same-second launch, apostrophe message) on Windows PowerShell 5.1 and pwsh 7.6.5.

---

### 🔴 Blocking — CI is red on this head; the PR's "green" test-plan checkbox is a Windows-only result

**Correctness bug.** `scripts/oe-detached-commit.ps1:46`

```powershell
$T = Join-Path $env:TEMP "cw-commit-$run"
```

`$env:TEMP` is undefined under pwsh on Linux. The required `verify.yml` check ran the Pester leg (`step_test_scripts: true`) and reported **68 passed / 1 failed / 5 skipped**:

```
[-] commits detached and the commit lands with the exact message, apostrophe included
ParameterBindingValidationException: Cannot bind argument to parameter 'Path' because it is null.
at <ScriptBlock>, .../scripts/oe-detached-commit.ps1:46
at <ScriptBlock>, .../scripts/tests/oe-detached-commit.Tests.ps1:53
```

Two things make this more than a portability nit:

1. This is **the only `$env:TEMP` in `scripts/`**. Every sibling uses `[System.IO.Path]::GetTempPath()` — `cast-reconcile.Tests.ps1:12`, `log-utils.Tests.ps1:11`, `sidecar-start.Tests.ps1:61`, `sidecar-sweep-port.Tests.ps1:50` — **including this PR's own test file at line 36**. The script diverges from the convention its own paired test follows.
2. `npm run test:scripts` is not in `verify:fast:branch`, so the PR body's `[x] npm run test:scripts green in the worktree (74 passed)` was a manual Windows run. It is false on the check that blocks merge. `verify.yml` is a required status check on `main`; this cannot merge as-is.

**A `$env:TEMP` → `GetTempPath()` swap alone will not turn it green.** Line 77 launches `Start-Process powershell`. `scripts/run-powershell.mjs:39` only falls back to `powershell` when `process.platform === 'win32'`; the Linux runner has `pwsh` and no `powershell` binary, so the smoke test will fail one line later. The functional `Describe` needs a platform gate. Precedent already in the tree, with the 5.1-safe idiom spelled out: `scripts/tests/prevent-sleep.Tests.ps1:9`

```powershell
Describe '...' -Skip:($env:OS -ne 'Windows_NT') {
```

---

### 🔴 Blocking — the one guard this PR exists to ship cannot fail

**Correctness bug.** `scripts/tests/oe-detached-commit.Tests.ps1:11-13`

```powershell
It 'launches the commit hidden -- the whole point of this script' {
    $script:scriptText | Should -Match '-WindowStyle\s+Hidden'
}
```

`$scriptText` is `Get-Content -Raw` over the **whole file**. The literal `-WindowStyle Hidden` appears three times in it, and two of them are prose:

- `scripts/oe-detached-commit.ps1:10` — the `.DESCRIPTION` docstring bullet
- `scripts/oe-detached-commit.ps1:73` — the comment above the call
- `scripts/oe-detached-commit.ps1:77` — the actual `Start-Process`

**Demonstrated, not reasoned.** I copied the script to a scratch dir outside the repo and applied one mutation — deleting `-WindowStyle Hidden` from line 77 only, leaving the docstring and comment untouched — then ran the new test file against the mutant:

```
Describing oe-detached-commit.ps1 static shape
  [+] launches the commit hidden -- the whole point of this script 321ms
  [+] takes the commit message as a bound parameter, never interpolated into a command string
  [+] writes the commit message without a BOM
  [+] does not reference a nonexistent Delete-Item cmdlet
Describing oe-detached-commit.ps1 functional smoke test
  [+] commits detached and the commit lands with the exact message, apostrophe included 5.02s
Tests Passed: 5, Failed: 0
```

**All five pass.** The smoke test cannot see it either — it asserts the commit landed, not that no window appeared. So the #2659 regression this PR exists to prevent ships with zero coverage: the paired test is green before the fix and green after it, which fails the house gate ("red before, green after, and red *for the reason claimed*") outright.

Two things worth doing rather than one: anchor the assertion to the `Start-Process` invocation (or parse the AST — `[System.Management.Automation.Language.Parser]::ParseFile`, the technique #1826 landed on `cast-lock.guard.test.ts` after a text-scan guard relocated its own defect three times), **and** make the smoke test observe the real property, e.g. assert the spawned process has no window (`(Get-Process -Id $procId).MainWindowHandle -eq 0`) rather than trusting the source text. As it stands nothing — automated or on-box — establishes that a window does not pop, and there is no acceptance-register row standing in for it either.

---

### 🔴 Blocking — two lanes launched in the same second corrupt each other, and both read `EXIT=0`

**Correctness bug.** `scripts/oe-detached-commit.ps1:45-46`

```powershell
$run = Get-Date -Format 'yyyyMMdd-HHmmss'
$T = Join-Path $env:TEMP "cw-commit-$run"
```

Second resolution, no per-invocation entropy, and `New-Item -Force` happily reuses an existing directory. Two OE lanes in separate worktrees — the normal operating model this script is written for — that launch within the same second share `msg.txt`, `commit.ps1`, `commit.pid` and `commit.log`.

**Demonstrated.** Two throwaway repos, two back-to-back invocations:

```
Ta = C:\Users\...\Temp\cw-commit-20260826-100632
Tb = C:\Users\...\Temp\cw-commit-20260826-100632
SAME SCRATCH DIR: True

--- A HEAD ---   test: lane B message
--- B HEAD ---   fatal: your current branch 'master' does not have any commits yet
--- commit.log --- [master (root-commit) dc24005] test: lane B message
                   EXIT=0
```

Repo **A committed with lane B's message**. Repo **B got no commit at all**. The shared log ends `EXIT=0`, so *both* lanes' polls — following `.clinerules/cline.md:206-214` exactly — read success. Lane B reports a commit that does not exist; lane A's history carries the wrong subject. That is the "success reported while doing nothing" shape, arriving through the one path the whole document exists to make reliable.

The fix is the line the paired test already uses on itself — `scripts/tests/oe-detached-commit.Tests.ps1:36` names its scratch dir `"oe-commit-test-$([Guid]::NewGuid())"`. The script should do the same (GUID or `$PID`).

---

### 🟠 Significant — a trailing backslash in `-Worktree` reintroduces the exact quoting bug the script claims to close

**Correctness bug.** `scripts/oe-detached-commit.ps1:77-79`

```powershell
$p = Start-Process powershell -WindowStyle Hidden -PassThru -ArgumentList @(
    '-NoProfile', '-ExecutionPolicy', 'Bypass', '-File', "`"$T\commit.ps1`"",
    '-Dir', "`"$T`"", '-Worktree', "`"$Worktree`"")
```

`$T` is built with `Join-Path`, so it never ends in `\`. `$Worktree` is caller-supplied. When it ends in a backslash, the emitted token is `"C:\path\"` — and `CommandLineToArgvW` reads that final `\"` as an **escaped quote**, so the argument never terminates and swallows the rest of the command line.

**Demonstrated** — real script, real repo, `-Worktree ($repo + '\')`:

```
git : fatal: cannot change to 'C:\Users\...\Temp\wtslash-0192cd87... commit
-F C:\Users\...\Temp\cw-commit-20260826-100601\msg.txt': Invalid argument
EXIT=128
```

The commit does not land. The comment two lines above it —

> `scripts/oe-detached-commit.ps1:55-56` — *"Paths come in via `param()` rather than string interpolation, so nothing about `$Worktree` or the scratch dir can reintroduce a quoting bug either."*

— is false as written; `$Worktree` *is* interpolated, at line 79. A trailing backslash is what a lane gets from a copied Explorer path or a `$PWD`-shaped string, and `.clinerules/cline.md:177` shows the placeholder without one, so nothing steers the caller away from it. Normalising once at entry (`$Worktree = $Worktree.TrimEnd('\','/')`, or resolving it) closes it; so does handing the worktree to the child through a file the way the message already is.

---

### 🟠 Significant — `.clinerules/cline.md` still tells lanes to run `git -C $W`, but the rewrite deleted the only line that assigned `$W`

**Correctness bug.** `.clinerules/cline.md:219`, `:223`, `:230`

The removed recipe defined it (`$W = 'C:\Claude\Projects\wt-<your-worktree>'`). The replacement snippet at `:176-179` assigns only `$T` and passes the path as an inline literal to `-Worktree`. Those are now the file's **only three** occurrences of `$W`, all reads:

- `:219` — *"The one thing that settles any disagreement is the repo itself: `git -C $W log --oneline -1`. Check that before acting on a surprising verdict"*
- `:223` — *"`EXIT=0` means the hook passed; confirm with `git -C $W log --oneline -1`"*
- `:230` — *"check `git -C $W log --oneline -1` before concluding it never ran"*

**Demonstrated on both shells a lane can be running:**

```
pwsh 7.6.5      : fatal: cannot change to 'log': No such file or directory   (exit 128)
Windows PS 5.1  : fatal: cannot change to 'log': No such file or directory   (exit 128)
```

PowerShell drops the undefined variable and git eats `log` as the `-C` argument. It fails loudly rather than silently reading the wrong repo — but it fails at precisely the three moments the document nominates it as the authoritative tiebreaker, i.e. when a lane is already confused about whether its commit landed. This repo's own record of what happens next is #2520: a lane spending its whole turn reasoning about an unexplained error at exactly this point and dying on the token cap. Re-introduce the `$W` assignment in the snippet (and use it for `-Worktree`, so the two cannot drift apart again).

---

### 🟡 Minor

- **The other three static assertions share the blind spot** — all are whole-file text greps. `:21` `Should -Not -Match 'commit -m'` fires on any future doc sentence that mentions `commit -m` (false red) and misses `commit --message` or `commit  -m` (false green); `:22` and `:26` would each be satisfied by a comment alone, same as the `-WindowStyle` one. Worth fixing as a class alongside the blocking guard, not site by site. *(Cleanup.)*
- **`Remove-Item (Join-Path $T 'commit.log')` at `:71` is justified by a mechanism that does not exist.** The comment keeps it "because a caller could reuse a `$T`" — there is no parameter through which a caller can supply `$T`. The reachable reuse is the same-second collision above, and against *that* this line is actively harmful: it deletes the other lane's in-flight log. `commit.pid` gets no equivalent clear and is written *after* `Start-Process` (`:80`), so the same window leaves a stale pid readable by a poll. Once the scratch dir is GUID-named the line becomes genuinely unreachable and can go. *(Cleanup — the stale comment is the defect, per shape 9.)*
- **One stop-rule was dropped in the doc rewrite.** The old text carried *"If the retry produces the identical error, **stop retrying** and say so once with the exact command and the error."* The replacement only says to re-check you are calling the script. Given #2382's six identical claim-and-fail cycles, that sentence was doing work. *(Cleanup.)*

---

### ✅ What is solid

- The here-string's closing `'@` sits at column 0 (`:63`) — the PS 5.1 `WhitespaceBeforeHereStringFooter` trap the old recipe warned about is genuinely gone, and moving it into a `.ps1` removes the markdown-indentation hazard that created it.
- `-Message` really is `[Parameter(Mandatory)][string]` and reaches git via `-F <file>`, never `-m` and never interpolated. I ran the real script with `test: verify detached commit don't drop quoting` and the subject landed byte-for-byte. `Mandatory` also rejects `''`, so an empty message can't reach `git commit`.
- BOM avoidance (`New-Object System.Text.UTF8Encoding $false`, `:53`) is correct and correctly motivated.
- The child's `$ErrorActionPreference = 'Continue'` with `*>&1` (`:59-62`) correctly dodges the PS 5.1 NativeCommandError-on-stderr trap, and `$LASTEXITCODE` resolves in the child rather than the parent.
- Quoting each `-ArgumentList` element individually is the right instinct — incomplete only for the trailing-backslash case above.
- `Closes #2661` is present, plain, outside any code span; the issue exists and is labelled `area:ops` + `type:chore` with `moscow:` left unset, matching CONTRIBUTING's backlog-shaped convention. PR title passes the convention lint.
- No second copy of the recipe anywhere else — `commit.pid` / `cw-commit-` / `oe-detached-commit` match only these three files; `.claude/skills/` and `AGENTS.md` carry no competing detached-commit instructions. The class is not left armed in a sibling doc.
- The here-string column-zero trap survives elsewhere in the doc (`.clinerules/cline.md:58`), and the poll recipe (`:206-214`), alive/done-disagreement guidance (`:216-221`), `*>&1` NativeCommandError note (`:226-232`), "Do not end your turn while it runs" (`:234-237`) and the `--no-verify` prohibition (`:239-242`) all survived the rewrite intact.
- Skipping the release-notes pair is declared explicitly in the PR body and is defensible for internal agent tooling.

---

### Verdict

**Do not merge.** Three blocking correctness bugs and two significant ones.

The shape worth naming: this PR replaces a copy-pasted recipe with a script precisely so the properties stop being aspirational — and then asserts each property against the script's own prose. The `-WindowStyle Hidden` guard passes on a mutant that has the flag deleted; the `$Worktree` comment claims a quoting immunity the next line does not provide; the scratch-dir comment reasons about a `$T`-reuse a caller cannot trigger while the reuse that *does* happen silently swaps two lanes' commits and reports `EXIT=0` to both. Fixing the CI red alone will turn the check green without touching any of that.

Per the re-review trigger, fixing and pushing these re-triggers a pass.

<sub>Reviewer: `pr-reviewer` (Premium tier, non-fork, `effort: xhigh`). Findings only — no tracked file was modified.</sub>

dudarenok-maker:
## PR review — pass 2 (head 036ba02e, depth medium)

Depth note: the dispatch said `low`, but SKILL.md's re-review rule re-derives depth from the PR's *current* commit set, and `036ba02e` is a single-scope `fix(scripts)` → **medium**. Reviewed at medium.

Scope: deltas only, per the skill. Verified before probing: the CI run for this head (`Verify` [32914745363](https://github.com/dudarenok-maker/Castwright/actions/runs/32914745363), job *Lint, typecheck & fast checks*), then re-ran pass 1's exact mutation plus **nine** further mutants of the script against the shipped test file on Windows PowerShell 5.1 / pwsh 7.6.5 + Pester, and re-measured the Win32 argv quoting with a live `Start-Process` argument-dump child.

---

### Disposition of the five pass-1 findings

**1. 🔴 CI red on Linux (`$env:TEMP`) — VERIFIED RESOLVED.**
`Verify` is green on `036ba02e` and the Pester leg genuinely ran, not skipped: `npm run test:scripts` → Pester v5.9.0 → `Tests Passed: 72, Failed: 0, Skipped: 8` (was 68/1/5). In the ubuntu log all eight static tests are `[+]` and all three functional ones are `[!]`, so `-Skip:($env:OS -ne 'Windows_NT')` (`Tests.ps1:95`) does suppress the block's `BeforeEach`/`BeforeAll` too — no `Start-Process`/`git` error surfaces from it.
I checked the static block for its own cross-platform landmine and there is none: `Join-Path $PSScriptRoot '..\oe-detached-commit.ps1'` (`Tests.ps1:15`) hands `[Parser]::ParseFile` — a raw .NET API with no PowerShell path normalisation — a path containing `\`, but the Unix FileSystem provider normalises it inside `Join-Path`, and `[+] parses without errors` on ubuntu proves it end-to-end rather than by argument.

**2. 🔴 The `-WindowStyle Hidden` guard could not fail — VERIFIED RESOLVED.**
Pass 1's exact mutation (delete the flag from the live `Start-Process` at `:98`, leave the `.DESCRIPTION` bullet and the `:95-97` comment untouched) is now **red** at `Tests.ps1:56`:
```
[-] launches the commit hidden -- the whole point of this script
 Expected the actual value to be greater than '-1', because a -WindowStyle parameter
 must be bound on the real Start-Process call, not just mentioned in a comment, but got -1.
```
I also ran three mutations pass 1 did not, all correctly red: `-WindowStyle Normal` (red at `:57`, `Expected: 'Hidden' / But was: 'Normal'`); splatting the parameters as `@spOpts = @{ WindowStyle = 'Normal' }` so no `CommandParameterAst` exists at all (red at `:56` — it fails **closed**); and the `Should -Be 1` count assertion at `:42` closes the `[Diagnostics.Process]::Start` escape.

**The second half of that finding — "make the smoke test observe the real property, e.g. `(Get-Process -Id $procId).MainWindowHandle -eq 0`" — was not implemented, and I verified that suggestion was unsound.** Measured on this box:

| launch | `MainWindowHandle` |
|---|---|
| `Start-Process powershell -WindowStyle Hidden` | `0` |
| `Start-Process powershell -WindowStyle Normal` | `0` |

It does not discriminate — the console window is owned by conhost/Windows Terminal, not by `powershell.exe` — so pass 1's proposed assertion would itself have been a test that cannot fail. Not implementing it was the right call; recording it here so it is not re-proposed. Nothing still establishes that no window *appears*, but the AST guard now pins the only mechanism that controls it, and I am not raising it further.

**3. 🔴 Same-second scratch-dir collision — VERIFIED RESOLVED.**
`:64` is now `cw-commit-$run-$([Guid]::NewGuid().ToString('N').Substring(0, 8))`, and I could not reproduce the swap. On the question you asked — *does the new test actually exercise two concurrent launches, or would it pass with the bug present?* — it is a genuine regression test. I reverted the GUID (leaving the literal in a comment) and ran the file **five** times: red 5/5 at `Tests.ps1:168`, e.g.
```
Expected 'C:\...\Temp\cw-commit-20260826-103106' to be different from the actual value, but got the same value.
```
The two `& $scriptPath` calls land ~30-60 ms apart, so the same-second collision is deterministic in practice, not probabilistic. (It is skipped on the merge gate — see N1.)

**4. 🟠 Trailing backslash on `-Worktree` — resolved behaviourally; the paired test cannot attribute it.**
End-to-end at head with `-Worktree ($repo + '\')` now lands `EXIT=0` with the message intact; pass 1's `fatal: cannot change to '...' / EXIT=128` is gone.
But the fix is **doubled**, and neither half is pinned. `$Worktree = $Worktree.TrimEnd('\')` (`:56`) and `if ($Value.EndsWith('\')) { $Value += '\' }` (`:91`) each independently close the single-backslash case. Reverting **either one alone** leaves all 11 tests green on Windows — including the new `tolerates a trailing backslash…` test. Only reverting both goes red, and then with exactly pass 1's error:
```
[-] tolerates a trailing backslash on -Worktree instead of swallowing the rest of the command line
 'git : fatal: cannot change to '...oe-commit-test-054ad6e0…
 commit -F ...\msg.txt': Invalid argument … EXIT=128'
```
So the behaviour is covered, but no test attributes it to either line, and the static guard at `Tests.ps1:87` pins one line's *spelling* instead. See N1 and N2.

**5. 🟠 `.clinerules/cline.md`'s dangling `$W` — VERIFIED RESOLVED.**
`$W` is assigned at `:176` and read at `:182`, `:224`, `:228`, `:235`; `$T` assigned at `:177`, read at `:212`, `:214`, `:217`. I swept every `$Var` in every ```powershell fence and every inline code span across all 312 lines: the only never-assigned names left are `$j`/`$i` at `:109`/`:111`, which are illustrative prose in the worktree-teardown bullet (the same shape CLAUDE.md's own recipe uses), untouched by this PR — not a finding. The dropped *"if a retry produces the identical error, stop retrying"* guidance is restored at `:200-204`.
Also folded without being asked, and confirmed: the false comment pass 1 quoted at old `:55-56` is now correctly narrowed to the child script's own source (`:73-75`), and pass-1 minor #2 (`Remove-Item commit.log`, justified by a `$T`-reuse mechanism no caller can trigger) was deleted outright.

---

### 🟠 Significant — N1: both regression tests this round added are invisible to the check that blocks merge, and the two guards that *are* visible there are whole-file greps — the exact shape pass 1 blocked on, re-armed by the fix commit

**Correctness bug.** `scripts/tests/oe-detached-commit.Tests.ps1:86-92`

```powershell
It 'strips a trailing backslash from -Worktree before it can escape a closing quote' {
    $script:scriptText | Should -Match ([regex]::Escape('$Worktree = $Worktree.TrimEnd(''\'')'))
}
It 'gives every scratch directory a per-call GUID component, not just a per-second timestamp' {
    $script:scriptText | Should -Match '\[Guid\]::NewGuid\(\)'
}
```

`$scriptText` is `Get-Content -Raw` over the whole file — comments and comment-based help included. That is precisely what this file's own new header (`Tests.ps1:5-12`) says the AST rewrite exists to stop. The commit fixed the class for `-WindowStyle Hidden` and re-created it, in the same commit, for the two fixes it added.

**Demonstrated, both directions.** Simulating the CI shape (`$env:OS` unset, so the functional block skips exactly as it does on ubuntu):

*False green — pass-1 finding 4 fully restored, gate green:*
```
# reverted: $Worktree = $Worktree.TrimEnd('\')          <-- literal survives in a comment
Tests Passed: 8, Failed: 0, Skipped: 3
```
*False green — pass-1 finding 3 fully restored, gate green:*
```
# reverted; formerly $([Guid]::NewGuid().ToString('N'))
$T = Join-Path ([System.IO.Path]::GetTempPath()) "cw-commit-$run"
Tests Passed: 8, Failed: 0, Skipped: 3
```
*False red — a strict improvement fails:*
```
$Worktree = $Worktree.TrimEnd('\','/')
[-] strips a trailing backslash from -Worktree before it can escape a closing quote
```

**Why "invisible to the merge gate" is literal, not rhetorical:**

- `test:scripts` appears in `test:all` / `verify` / `verify.yml` only. It is in **neither** `verify:fast:scoped` (pre-commit, `package.json:77`) **nor** `verify:fast:branch` (pre-push, `package.json:78`) — nothing local runs it automatically.
- Every `verify.yml` job is `runs-on: ubuntu-latest` (`verify.yml:143` for the leg that hosts Pester), where `-Skip:($env:OS -ne 'Windows_NT')` skips all three functional tests. This head's own CI log shows exactly that: three `[!]` lines, `Skipped: 8`.
- The first *automatic* Windows execution of those three tests is `cross-os.yml`'s Wed/Sun 02:00 UTC cron on `main` (`cross-os.yml:28-29`, `:49`, `:86` → `verify:quick` → `test:all` → `test:scripts`) — i.e. **after merge**, up to four days later, where a red is a broken `main` rather than a blocked PR.
- The skip guard also fails **open**: `$env:OS` absent ⇒ skip. On the gate that blocks merge, the surviving enforcement for findings 3 and 4 is two greps that a comment satisfies.

**Fix shape:** give `:86-92` the same treatment `:45-58` got — assert against the `InvokeMemberExpressionAst` for `TrimEnd` and against the `$T` assignment's `ExpandableStringExpressionAst`, not against `$scriptText`. `:82-84`'s `Delete-Item` grep is the third instance of the same class (pass-1 minor, still open) and should go in the same pass rather than site by site.

---

### 🟡 Minor

- **N2 — `ConvertTo-QuotedArg`'s escape rule is right only for a run of exactly one backslash, and its comment claims otherwise.** `scripts/oe-detached-commit.ps1:88-93`. Win32 requires the whole trailing backslash *run* doubled; the helper appends exactly one. Measured by launching the real quoting through `Start-Process` into an argument-dump child with a `-Sentinel` argument placed after `-Worktree`:

  | input | quoted | child sees | sentinel |
  |---|---|---|---|
  | `C:\repo\` | `"C:\repo\\"` | `C:\repo\` | survives ✓ |
  | `C:\repo\\` | `"C:\repo\\\"` | `C:\repo\" -Sentinel SENTINEL-MUST-SURVIVE` | **lost** ✗ |
  | `C:\repo\\\` | `"C:\repo\\\\"` | `C:\repo\\` (a backslash silently dropped) | survives |
  | `C:\repo\\\\` | `"C:\repo\\\\\"` | folded | **lost** ✗ |
  | `C:\my repo\\` | `"C:\my repo\\\"` | folded | **lost** ✗ |

  Every even-length run reproduces the original bug verbatim. **This is latent, not live** — `$Worktree` is `TrimEnd('\')`-ed first (which strips the *whole* run), and `$T` and the child-script path come from `Join-Path`, so nothing reachable today reaches the helper with a trailing backslash. The defect is the comment at `:88-89` — *"double any trailing backslash here so the rule applies uniformly to every path this script quotes"* — asserting a uniform rule the code does not implement, on a general-purpose helper called on three different values. (Related, measured: a trailing **forward** slash quotes and binds cleanly, so `TrimEnd('\','/')` is not needed for the quoting bug. Separately, `'C:\'.TrimEnd('\')` → `C:`, which `git -C` reads as the drive's *current* directory, not the root — irrelevant for a worktree path, noted for completeness.) *(Cleanup.)*

- **N3 — the WindowStyle AST guard false-reds on two legal, correct spellings.** `Tests.ps1:57` compares `$elements[$windowStyleIndex + 1].Extent.Text` to the bareword `Hidden`. Measured: `-WindowStyle 'Hidden'` → actual `'Hidden'` (quotes included) → red; `-WindowStyle:Hidden` → actual `-PassThru` → red. Both keep the window hidden. Reading `CommandParameterAst.Argument` when it is populated, and trimming quotes otherwise, closes both. *(Cleanup.)*

- **N4 — the new functional tests leak their scratch directories.** `AfterEach` removes `$repoDir` (`Tests.ps1:111-115`) and the concurrency test's `finally` removes `$repoB` (`:176`), but nothing removes the `cw-commit-*` directory the script returns. Each Windows run of `npm run test:scripts` leaves four behind permanently; `%TEMP%` on this box already holds **79** `cw-commit-*` directories dating back to 2026-08-15. The GUID suffix makes the count strictly monotonic where the old timestamp name at least coalesced same-second runs — a good trade for correctness, but it makes cleanup the test's job. One `Remove-Item $scratchDir -Recurse -Force` per test closes it. *(Cleanup.)*

- **N5 — `Wait-ForDetachedCommit` computes `$done` and never reads it.** `Tests.ps1:129`; the loop condition at `:130` reads only `$alive`, and nothing downstream uses `$done`. The helper is named for waiting on the sentinel but only waits on the process. Harmless today (process exit implies the log is flushed), but it is a condition that gates nothing. *(Cleanup.)*

---

### Verdict

**Four of the five pass-1 findings are verified resolved; the fifth (trailing backslash) is resolved in behaviour but not attributable by any test.** Pass 1's blocking `-WindowStyle Hidden` guard now genuinely fails on the mutation that defeated it, and on three more I tried; the collision fix is real and its test is a real red-before/green-after; CI is green with the Pester leg actually executing.

**One 🟠 correctness bug remains (N1), so this does not merge yet** — narrower than pass 1's verdict, and it is the same shape one level up: the round that fixed a grep-guard blind spot shipped two new grep guards for its own two new fixes, on a gate where the functional tests that would have caught them are skipped. Both of this round's fixes can be reverted wholesale with the required check still reporting 8 passed / 0 failed.

Per the re-review trigger, fixing and pushing N1 re-triggers a pass — which would be pass 3, the loop cap.

<sub>Reviewer: `pr-reviewer` (Premium tier, non-fork, `effort: xhigh`). Findings only — no tracked file was modified; all mutants were built and run out of tree.</sub>

dudarenok-maker:
## PR review — pass 3 (head 45efd337, depth medium)

Depth `medium`, carried forward from pass 2's re-derivation: the commit set is two single-scope `fix(scripts)` commits on a single-scope `chore(scripts)` base — `medium` on the ladder, not `low`.

**This is round 3 of 3 — the loop cap.** Whatever this comment says, the loop does not auto-continue past it.

Scope: deltas only, per the skill. Verified before probing: CI on this head ([Verify 32916751201](https://github.com/dudarenok-maker/Castwright/actions/runs/32916751201) — 15/15 checks green, and the Pester leg genuinely *executed* on ubuntu rather than being scope-skipped: `Pester v5.9.0` → `Tests Passed: 72, Failed: 0, Skipped: 8`); a full `npm run test:scripts` in the worktree (**78 passed / 0 failed / 2 skipped**); then **32 out-of-tree mutants of the script + 2 of the test file**, run against the shipped Pester file under both the ubuntu merge-gate shape (`$env:OS` cleared, so the functional block skips exactly as it does on CI) and a full Windows run (**11 passed / 0 failed / 0 skipped**); plus a Windows PowerShell 5.1 ↔ pwsh 7.6.5 parity probe of every AST construct this commit introduced.

---

### Disposition of pass 2's five findings

**N1 🟠 — the fix commit's own two assertions were whole-file greps — VERIFIED RESOLVED.**

Pass 2's exact mutation (comment out the real line, leave the surrounding comment) now fails closed, singly and together, under the CI shape:

| mutant | result |
|---|---|
| `# $Worktree = $Worktree.TrimEnd('\')` only | **RED** — `Expected a value, because $Worktree = $Worktree.TrimEnd(...) must be a real, executed assignment, but got $null or empty.` |
| revert `$T` to `"cw-commit-$run"`, leave `[Guid]::NewGuid()` alive in a comment | **RED** — `Expected the actual value to be greater than 0, because [Guid]::NewGuid() must be part of the actual $T assignment, not just present elsewhere in the file, but got 0.` |
| both at once | **RED ×2** (both assertions fire independently) |

I also checked the rewrite did not just relocate the blind spot one level down — the assertions pin the *argument*, not merely the call's presence:

| mutant | result |
|---|---|
| `$Worktree = $Message.TrimEnd('\')` (right shape, wrong source) | **RED** |
| `$Worktree = $Worktree.TrimEnd()` (no args — does not strip a backslash) | **RED** — `Expected 1, but got 0` |
| `$Worktree = $Worktree.TrimEnd('/')` (wrong char) | **RED** |

**N2 🟡 — `ConvertTo-QuotedArg`'s comment overclaimed a uniform rule — VERIFIED RESOLVED.** The misleading half ("so the rule applies uniformly to every path this script quotes" — the sentence that would let a fourth call site be added on a false premise) is gone, and the replacement at `scripts/oe-detached-commit.ps1:88-91` enumerates all three real call sites correctly. Recording one loose edge so it is not re-litigated: that enumeration actually establishes **zero**, not one — `$Worktree` is `TrimEnd`-ed (which strips the whole run) and the other two are `Join-Path` results — so the `if ($Value.EndsWith('\'))` branch is unreachable defensive code, and "the only case any caller of this helper can actually produce" is one notch loose against the clause that immediately follows it. No failure scenario attaches; not re-raising.

**N3 🟡 — the WindowStyle AST check false-redded on `-WindowStyle:Hidden` / `'Hidden'` — VERIFIED RESOLVED, and it did not over-relax.** The specific risk in relaxing this was accepting something that is not Hidden. It does not:

| mutant on the live `Start-Process` | expected | result |
|---|---|---|
| `-WindowStyle:Hidden` | green | **GREEN** ✓ |
| `-WindowStyle 'Hidden'` | green | **GREEN** ✓ |
| `-WindowStyle "Hidden"` | green | **GREEN** ✓ |
| `-WindowStyle:'Hidden'` | green | **GREEN** ✓ |
| flag deleted entirely (pass 1's mutation) | red | **RED** ✓ |
| `-WindowStyle Normal` | red | **RED** ✓ |
| `-WindowStyle:Normal` | red | **RED** ✓ |
| `-WindowStyle 'Normal'` | red | **RED** ✓ |
| `-WindowStyle:'Normal'` | red | **RED** ✓ |
| `-WindowStyle Maximized` | red | **RED** ✓ |

The `.Trim("'`\"")` string→`char[]` coercion and the `CommandParameterAst.Argument`-vs-next-element split both behave identically on Windows PowerShell 5.1 (`5.1.26100.9223`) and pwsh `7.6.5`.

**N4 🟡 — the functional tests leaked their scratch directories — VERIFIED RESOLVED, measured, with a working control.** I did not take the commit message's word for it. `%TEMP%` before/after a full `npm run test:scripts` in the worktree: **0 new `cw-commit-*`, 0 new `oe-commit-test-*`** (the single pre-existing `cw-commit-20260826-105112-abb3b5ab` is this PR's own dogfooded commit, not a test leak). And the instrument can detect a leak — neutering either half of the new mechanism, one at a time, leaks exactly the 4 directories pass 2 counted:

| test-file mutant | leaked |
|---|---|
| shipped file, unmodified | **0** |
| `# NEUTERED: $script:scratchDirs += $Dir` | **4** |
| `AfterEach` sweep removed, registration kept | **4** |

The `$script:` sharing between the `BeforeAll`-defined `Register-ScratchDir` and the `AfterEach` sweep is real, not incidental, and works on 5.1 as well as 7.

**N5 🟡 — `Wait-ForDetachedCommit` computed `$done` and never read it — VERIFIED RESOLVED.** No `$done` remains anywhere in `scripts/`; the only three occurrences left on the branch are `.clinerules/cline.md:214-217`, where it *is* assigned and read (the lane poll recipe), correctly untouched. The loop's exit condition is `while ($alive -and (Get-Date) -lt $deadline)` alone; `$logPath` is still assigned at `:180` and returned at `:186`, and nothing else referenced `$done`. 11/11 green on a full Windows run.

---

### 🟡 Minor — the third instance of the grep class is still a grep, and the new file header now says it isn't

**Cleanup.** `scripts/tests/oe-detached-commit.Tests.ps1:5-6` and `:114-116`

Pass 2's fix shape named three sites, not two: *"`:82-84`'s `Delete-Item` grep is the third instance of the same class (pass-1 minor, still open) and should go in the same pass rather than site by site."* Two were converted; the third was not:

```powershell
It 'does not reference a nonexistent Delete-Item cmdlet (the real one is Remove-Item)' {
    $script:scriptText | Should -Not -Match 'Delete-Item'
}
```

Meanwhile the header this commit left in place opens with a blanket claim that is now 7/8 true:

> `# The static checks below parse the script's AST rather than grepping its`
> `# text: ...`

**Demonstrated.** Adding a single comment line to the script and changing no executable code at all —

```powershell
# NB: the cmdlet is Remove-Item, never Delete-Item
Write-Output $T
```

— turns the suite **RED** at `:114`. That is precisely the prose-matches-the-grep failure mode the header says the AST rewrite exists to eliminate. (Control: real executable `Delete-Item` also reds, so the check does catch the thing it is for.)

Two things make this 🟡 and not 🟠, and I want to be explicit rather than leave it ambiguous at the cap: it is a **negative** assertion, so it over-fires rather than under-fires — it can only ever false-*red*, never let a real `Delete-Item` through — and the shipped script is unaffected either way. There is also a defensible argument that this one *should* stay textual, since a hallucinated cmdlet **name** is a textual property, not a structural one. If so the cheapest correct fix is the header sentence, not the check: name the exception rather than assert it away.

Second, smaller instance of the same shape, in the PR's own record: the body's test plan still reads `npm run test:scripts` green in the worktree (**74 passed** …) and describes the coverage as *"static-shape guards (`-WindowStyle Hidden` present, no `Delete-Item`, `-Message` never interpolated into `commit -m`) + a functional smoke test"*. The real numbers at this head are **78 passed / 0 failed / 2 skipped**, and the shipped coverage also includes the trailing-backslash test, the two-lane concurrency test, and the AST rewrite of all of it. That body becomes the merge-commit message.

---

### Notes — probed, not findings

The dispatch asked me to find a mutation on the same class the AST rewrite might still miss. There is one, and I am deliberately **not** filing it as a finding; it is recorded so a future maintainer does not rediscover it as a surprise.

The new guards are **structural, not flow-sensitive**. They assert that a statement of the right shape exists in the file, not that it runs before its effect is needed or that its result survives. Six mutants stay green:

| mutant | green because |
|---|---|
| the `TrimEnd` assignment moved to the bottom of the file (runs *after* `Start-Process`) | the assignment still exists |
| `$T` re-assigned without a GUID *after* the good assignment (second one wins) | `Select-Object -First 1` takes the first |
| `TrimEnd` runs, then `$Worktree` restored from a saved copy | net no-op, statement present |
| `[Guid]::NewGuid().ToString('N').Substring(0, 0)` — called, contributes zero chars | the call is genuinely inside the `$T` subtree |
| `[Guid]::NewGuid()` piped to `Out-Null` inside the `$T` assignment | same |
| `TrimEnd` present, but the call site quotes `$PSBoundParameters['Worktree']` instead | the guard never looks at the call site |

Why this does not block: no purely-static guard can be flow-sensitive, this repo's nearest precedent (`server/src/workspace/cast-lock.guard.test.ts`) is explicitly call-graph-blind and documents that in its own header, and **every realistic removal mode is now caught** — delete the line, comment it out, revert it, weaken its argument, or use it on the wrong variable are all red. The six above are mutations nobody writes by accident. If it is ever worth closing, the mechanism is a functional assertion, not a smarter parse.

The mirror of that trade is three **false reds** on strictly-safe rewrites — `TrimEnd('\','/')` (fails `$trimArgs.Count | Should -Be 1`), the `TrimEnd` inlined at the quoting call instead of reassigned, and the GUID hoisted into its own variable before `$T`. All fail closed with a `-Because` that explains what is required, so the cost is one confused minute, not a shipped bug. Same call pass 2 made on the WindowStyle equivalent; noting it rather than raising it.

Also confirmed clean in this pass, since I had the harness up: the `commit -m` assertion is now correctly scoped to the child-script string literal — `commit -m` in a top-level comment stays green, the child script actually switching to `commit -m $Msg` reds. The BOM guard reds on both `UTF8Encoding $true` and `[System.Text.Encoding]::UTF8`. `.clinerules/cline.md` is untouched by this commit and its `$W`/`$T`/`$id`/`$alive`/`$done` recipe still matches what the script actually writes and returns.

---

### Verdict

**All five of pass 2's findings are verified resolved, and the one blocking finding (N1) is resolved for the mutation class it named — verified by re-running pass 2's exact mutations plus 29 more, not by reading the diff.** No 🔴 and no 🟠 in this pass.

The single 🟡 above is cleanup and, per the triage rules, does **not** re-trigger the loop — it is fixed in this round and merged, no fourth pass. Beyond that one item, **this is clear to merge.**

Recording the cap explicitly so the calling session does not have to infer it: this was round 3/3. Since nothing here trips the re-review trigger, the cap is not actually being *hit* — the loop closed on its own.

<sub>Reviewer: `pr-reviewer` (Premium tier, non-fork, `effort: xhigh`). Findings only — no tracked file was modified; `git rev-parse HEAD` / `git status --porcelain` unchanged, and all 34 mutants were built and run out of tree.</sub>

Issue #1813, opened 2026-07-26
chore(server): surface a progress signal during the cloned/designed voice resolver re-derive pre-pass


dudarenok-maker:
## Context

fs-38 Wave 3b2 (cloned-voice resolver + lifecycle, Refs #624) added a per-chapter
pre-pass in `synthesiseChapter` (`server/src/tts/clone-voice-resolver.ts`,
`resolveClonedVoicesForChapter` / `resolveDesignedVoicesForChapter`) that can re-derive
a Repairable cloned voice or self-heal a missing designed-voice `.pt` from its retained
clip, before any synth call fires.

Both call `deps.reportProgress?.('Preparing voice "<name>"...')`, but the production
dependency wiring (`buildDefaultCloneResolverDeps` / `buildDefaultDesignedResolverDeps`
in `server/src/tts/synthesise-chapter.ts`) wires `reportProgress: undefined` — there is
no free-text progress channel on `SynthesiseChapterOpts` today, only typed per-group/
per-title ticks (`onGroupStart`/`onTitleStart` etc.), and forcing this message through
one of those would misuse a differently-shaped callback.

## Impact

A multi-second re-derive (sidecar clone-distil round trip) during chapter generation
shows no UI signal — the generation view just appears to sit idle for those few seconds
before synth resumes. Not a correctness bug (the chapter still completes normally), but
a silent gap in an otherwise-observable pipeline.

## What

Add (or extend) a progress-tick channel on `SynthesiseChapterOpts` that a resolver
pre-pass re-derive can report through, and surface it in the generation view the same
way other per-chapter heartbeats already are.

## Key files

- `server/src/tts/synthesise-chapter.ts` (`SynthesiseChapterOpts`,
  `buildDefaultCloneResolverDeps`, `buildDefaultDesignedResolverDeps`)
- `server/src/tts/clone-voice-resolver.ts` (`ResolveChapterDeps.reportProgress`,
  `ResolveDesignedVoiceDeps.reportProgress`)
- generation-view heartbeat consumer (`src/views/generation.tsx` or the
  generation-stream store, whichever renders the existing group/title heartbeats)

dudarenok-maker:
## Handover — design approved, ready for pickup

The design for this is approved and committed: **`docs/superpowers/specs/2026-07-26-resolver-prepass-progress-phase-design.md`**, landing via **#1853** (docs-only, mergeable independently).

[...]

Issue #2652, opened 2026-08-25
[agent instructions][claude][verify] Verify fs-38's four residual gaps, open the PR (#2068)


dudarenok-maker:
## What
Verify the four prior children's work against #2068's own acceptance, then open the PR for the whole chain. This chain deliberately spans two local-GPU lanes (cline-qwen-local, cline-muse-local) and one cloud lane (cline-glm) plus this claude verify step — that spread is itself part of what's being validated (see #2068's linked context: this is also the first real multi-lane test of the newly-migrated local llama.cpp lanes).

## The parent's acceptance, restated as your checklist
1. **Item 2 (cline-qwen-local, #2648):** `cloneVoice.fulfilled` refetches instead of writing the raw `/clone` response, matching `revokeVoice`/`promoteRedesign`/`discardRedesign`. Re-run its mutation yourself (revert the fix, run the paired test, confirm RED with real output, restore, confirm GREEN). Do not trust the child's claimed receipt — reproduce it.
2. **Item 1 (cline-muse-local, #2649):** the clone-readiness gate's 'Proceed anyway' can no longer race 'Cast on <engine>''s persistence write. Re-run its mutation yourself the same way.
3. **Item 3 (cline-glm, #2650):** the transcript PATCH validates against `fresh.master` (post-lock) and returns 409, not 200, on the race. Re-run its mutation yourself.
4. **Item 4 (claude, #2651):** a test pins the retry route's intended no-guard behavior on `provenance === 'cloned'` entries, citing plan 276 Decision 7. Confirm no route-logic change was made (test-only child) and the test actually asserts the retry succeeds, not just that it doesn't throw.
5. All four decisions this chain made (refetch over inline-transform; CTA-awaits-own-write over debounce-flush; 409-on-race over 200/404; keep-no-guard over add-guard) are still what's on the branch — check nothing drifted from what each brief specified.
6. `npm run typecheck` passes for the whole branch, `npm test` (frontend) and `cd server && npm test` both pass.
7. No file outside `src/modals/clone-readiness-gate.tsx`, `src/store/voice-library-slice.ts`, `server/src/routes/voice-library.ts`, and the four children's own new/extended test files changed on this branch.

## FAIL ordering — exactly this, if anything above fails
1. Create a rework `[cline][task]` child, briefed from exactly what failed (name the specific engine/lane that should re-attempt it based on whether the failure was frontend-only, server-only, or cross-cutting).
[...]

Issue #2651, opened 2026-08-25
[agent instructions][claude][task] Pin the retry route's intended no-guard behavior with a test (fs-38 item 4)


dudarenok-maker:
## What
`POST /api/voice-library/{voiceUuid}/engines/{engine}/retry` has no `provenance === 'cloned'` guard, so it will delete a `failed` engine slot on a designed entry too. The sibling transcript-PATCH route (fixed by the previous child in this chain, item 3) does guard on provenance — this is an asymmetry between two routes added in the same PR. Neither behavior is currently pinned by a test.

## Decision made (this was filed as needing one — see #2068 item 4)
The issue states plan 276's referenced 'Decision 7' does NOT require this guard, and that deleting a designed voice's failed slot is arguably correct (it lets the voice re-derive). **Keep current behavior (no guard) — do not add one.** The actual gap is that nothing pins this as intended rather than accidental. Write the test that locks in the current, no-guard behavior as deliberate. Do not add a provenance guard to this route — that would contradict Decision 7.

## Background you need
- The retry route (search `server/src/routes/voice-library.ts` for `/retry`) — find its handler and read how it currently handles a `failed` engine slot with no provenance check.
- The transcript PATCH route fixed by the previous child in this chain (item 3, cline-glm) — read its provenance guard for contrast; you're pinning the ABSENCE of the equivalent check here, not adding one.
- Plan 276 (`docs/features/276-cast-time-derivability-warning.md`, referenced from #2068) — its 'Decision 7' is the reasoning basis for why no guard is required here. Cite the relevant section in your test's comment or your receipt.

## The fix
Add a regression test asserting the retry route DOES delete/retry a failed engine slot on a `provenance === 'cloned'` entry with no guard blocking it — i.e. pin today's actual behavior as intentional, distinguishing it from the guarded sibling route. No source-code change to the route itself.

## Where the work happens
Worktree: `C:\Claude\Projects\wt-2068-clone-gate-gaps`
Branch: `fix/frontend-2068-clone-gate-gaps`
Do not create another worktree. Do not work in the primary checkout. Commit on this branch. Do NOT open a PR — only the final [claude][verify] child does that.

## Acceptance
- [ ] A new test (Vitest, server, colocated with the retry route's existing tests if any exist, otherwise create the file) asserts the retry route succeeds (no guard block) on a `provenance === 'cloned'` entry with a `failed` engine slot.
[...]

Issue #2650, opened 2026-08-25
[agent instructions][cline-glm][task] Transcript PATCH: re-validate under lock instead of 200-on-no-op (fs-38 item 3)


dudarenok-maker:
## What
`server/src/routes/voice-library.ts:612`, the transcript PATCH handler, validates against the **pre-lock** `existing.master`, then writes guarded on `fresh.master`. If `master` disappeared between those two reads, the handler currently returns 200 having written nothing — the client reads that as success for a write that never happened.

## Decision made (this was filed as needing one — see #2068 item 3)
The issue named this as needing a status-code decision (404? 409? re-validate under the lock and 400?). **Re-validate under the lock, and return 409 Conflict** if `fresh.master` is gone — 409 is the correct REST semantic for 'the resource changed underneath your request in a way that invalidates it', which is exactly this race. Do not silently write nothing and return 200; do not return 404 (the resource that disappeared is a sub-field, not the whole entity being addressed by the URL).

## Background you need
- `server/src/routes/voice-library.ts:612` — the transcript PATCH handler. Read the surrounding lock-acquisition code (this repo's `withCastLock`/`withCastLocks` convention per CLAUDE.md's cast.json section — check whether this route already locks, and if so where `fresh` is read relative to the lock).
- The validation currently reads `existing.master` (pre-lock) and the write is guarded on `fresh.master` (post-lock) — these are two different reads of the same field, taken at two different times, which is the actual race.

## The fix
Move validation to read `fresh.master` (the post-lock value), not `existing.master` (the pre-lock value). If `fresh.master` is absent/gone at that point, return 409 with a clear error body (follow this file's existing error-response shape for other 4xx cases) and do not perform the write. Do not change behavior for the case where `master` is present and unchanged — that path stays as-is.

## Where the work happens
Worktree: `C:\Claude\Projects\wt-2068-clone-gate-gaps`
Branch: `fix/frontend-2068-clone-gate-gaps`
Do not create another worktree. Do not work in the primary checkout. Commit on this branch. Do NOT open a PR — only the final [claude][verify] child does that.

## Acceptance
- [ ] The handler validates against `fresh.master` (post-lock), not `existing.master` (pre-lock).
- [ ] When `fresh.master` is absent at validation time, the handler returns 409 (not 200) and performs no write.
[...]

PR #2666, opened 2026-08-26
fix(frontend,server): close fs-38's four clone-readiness-gate residual gaps

### LINKED ISSUE #2068: fs-38 — four residual gaps in the cast-time clone-readiness gate (PR #2067 review)
Four confirmed findings from PR #2067's `code-review` gate that were **not** folded into that PR because each needs a decision rather than an obvious one-line fix. None is a regression; all four are pre-existing or newly-exposed asymmetries.

## 1. "Cast on \<engine\>" races "Proceed anyway" (Medium)

`src/modals/clone-readiness-gate.tsx:143`. `castActions.updateCharacter` persists only via `persistence-middleware.ts:71`, a **500 ms-debounced** `PUT /api/books/:bookId/state`. `requestStartGeneration` enqueues immediately, and the server render reads `cast.json` from disk — the enqueue payload carries no cast.

**Failure scenario:** the user presses "Cast on Coqui" (the row vanishes from the modal instantly, because redux updated), then presses "Proceed anyway" within 500 ms. The two buttons are adjacent and an emptied list invites exactly that. The generation POST lands before the cast PUT, the render reads the pre-switch `ttsEngine`, and the chapter hard-fails — the precise failure the gate exists to prevent.

**Why not fixed in #2067:** the fix is a choice between flushing the persistence debounce before enqueueing, tracking an in-flight flag that disables "Proceed anyway", and making the CTA await its own write. The first has the widest blast radius (it changes middleware behaviour for every caller); the third is most local. That is a design call, not a defect fix.

## 2. `cloneVoice.fulfilled` writes a raw entry into the slice (Low)

`src/store/voice-library-slice.ts:265`. Plan 276's `[R4]` correction states that **every** route handing an entry to the client applies `withComputedStaleness`. `POST /clone`'s response does not, and its result lands directly in `entries`.

Benign today because a fresh clone stamps the current version. It stops being benign the moment the sidecar reports a `baseModel`/`coquiVersion` that differs from the Node-side oracle: the client then reads `'ready'` where the server would say `'stale'`, and `cloneReadiness`'s rules 5 and 6 — gated on `slotStatus !== 'ready'` — go quiet for the rest of the session.

Every other entry-returning thunk (`revokeVoice`, `promoteRedesign`, `discardRedesign`, `designVoice`) refetches instead, so this is the only outstanding path.

**Why not fixed in #2067:** two defensible fixes (apply the transform on the `/clone` response, or make the thunk refetch like its siblings), and the route is outside the set that PR touched.

## 3. Transcript PATCH returns 200 on a silent no-op (Low)

`server/src/routes/voice-library.ts:612`. Validation reads the **pre-lock** `existing.master`; the write is guarded on `fresh.master`. If `master` disappeared between the two, the handler returns **200 having written nothing**, and the client reads that as success.

Needs a status-code decision (404? 409? re-validate under the lock and 400?), which is why it is here rather than in #2067.

## 4. The retry route has no `provenance === 'cloned'` guard (Low)

`POST /api/voice-library/{voiceUuid}/engines/{engine}/retry` will delete a `failed` engine slot on a **designed** entry. The transcript PATCH immediately above it does guard on provenance, so this is an asymmetry between two sibling routes added in the same PR.

Decision 7 does not require the guard, and deleting a designed voice's failed slot is arguably correct (it lets the voice re-derive). But no test pins **either** behaviour, so whichever is intended is currently unpinned.

## Acceptance

- Each of the four is either fixed with a paired, mutation-verified test, or explicitly closed as intended-behaviour with the reasoning recorded where the code is.
- Item 2 either fixes the path or amends plan 276's `[R4]` claim so it stops asserting universality it does not have.

## Benefit

*Benefit (technical):* removes the last four known asymmetries around the cast-time gate, one of which (item 1) can still produce the exact hard-fail the feature was built to prevent.

Found by the `code-review` gate on PR #2067. Relates to #1980, plan [276](https://github.com/dudarenok-maker/Castwright/blob/main/docs/features/276-cast-time-derivability-warning.md).

dudarenok-maker:
Closes #2068

## Summary
- **cloneVoice.fulfilled refetches instead of writing the raw `/clone` response** (cline-qwen-local, #2648) — matches `revokeVoice`/`promoteRedesign`/`discardRedesign`, so a stale `baseModel`/`coquiVersion` on the sidecar can no longer read `'ready'` client-side when the server would say `'stale'`.
- **"Proceed anyway" can no longer race "Cast on `<engine>`"'s persistence write** (cline-muse-local, #2649) — clicking "Cast on `<engine>`" disables "Proceed anyway" (labelled "Waiting for cast save…") for a buffer past the persistence debounce (`autosaveDebounceMs`, user-configurable, defaults to 500ms — see the review-round fix below for the buffer reading the real configured value rather than assuming the default), so the generation POST can no longer be issued before the cast PUT.
- **Transcript PATCH returns 409, not 200, when the master clip vanishes mid-request** (cline-glm, #2650) — the `master` existence check moved inside the post-lock callback (`fresh.master`), so a race that removes the master clip between the pre-lock read and the lock now returns 409 and performs no write, instead of a silent 200 no-op.
- **Retry route's `provenance === 'cloned'` no-guard behavior is now pinned by a regression test** (claude, #2651) — no route-logic change; the asymmetry with the sibling transcript-PATCH guard is confirmed intentional per plan 276 Decision 7, and the test asserts the retry actually succeeds (200, slot cleared), not just that it doesn't throw.

Both newly-migrated local llama.cpp lanes (cline-qwen-local, cline-muse-local) finished their items cleanly on the first pass — no rework was needed for either. This is the first real multi-lane chain to exercise them post-migration.

## Verify notes (this PR, #2652)
- All four decisions match the parent's acceptance and haven't drifted from their briefs.
- File scope: only `src/modals/clone-readiness-gate.{tsx,test.tsx}`, `src/store/voice-library-slice.{ts,clone.test.ts}`, `server/src/routes/voice-library.{ts,test.ts}` were touched by the four items — plus one `docs/release-notes-next.md` entry (item 3 only) documenting the shipped fix, which is docs-only and outside the strict file list but not a functional drift.
- `npm run typecheck` passes clean (frontend `tsc --noEmit` + server `tsc --noEmit -p .`).
- `npm test` (frontend): 344 files, 4893 tests — 1 failure (`teaser-governance.test.ts`, a 5s timeout) that reproduced GREEN in isolation on retry; unrelated to any of the four items or their files.
- `cd server && npm test`: 537 files, 7990 tests — 1 failure (`qwen-voice.test.ts` #1981 promote/assign race) that reproduced GREEN in isolation on retry; unrelated to any of the four items or their files (known GPU-contention flakiness on this box, same class already logged on #2651).

## Test plan
- Item 1: reverted `voice-library-slice.ts` to pre-fix → `voice-library-slice.clone.test.ts` RED (`expected "vi.fn()" to be called at least once`); restored → GREEN (3/3).
- Item 2: reverted `clone-readiness-gate.tsx` to pre-fix → `clone-readiness-gate.test.tsx` RED (`Proceed anyway` not disabled); restored → GREEN (16/16).
- Item 3: reverted `voice-library.ts` to pre-fix → `voice-library.test.ts` RED (`expected 400 to be 409`, `expected 200 to be 409`); restored → GREEN (179/179).
- Item 4: confirmed no source change to `voice-library.ts`; new test asserts `res.status === 200` and the slot is actually cleared on disk, not merely "doesn't throw".

## Review-round fixes (found in passing, fixed here per `pr-review-gate`)

An independent pass ([pass 1](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420040875)) found 1 blocking + 5 significant correctness/chore issues across the four items above. All were fixed on this branch before merge:
- Item 2's 600ms buffer was hardcoded against a fixed 500ms debounce assumption, but the actual persistence debounce (`autosaveDebounceMs`) is a user-configurable setting (100–10000ms); the fix now reads the real configured value and clears any prior pending timeout instead of letting overlapping clicks re-open the race. (A follow-on unmount-cleanup effect from this same fix was later removed as dead code — the modal is a global, always-mounted overlay that never unmounts.) The paired test now wires in the real settings/persistence path with fake timers, covering both the default and a non-default configured debounce.
- Item 3's 409 change (correctly added for the genuine "master removed mid-request" race) had also swallowed the pre-existing 400 case for an entry that *never* had a master or had it permanently revoked — that case now correctly returns 400 again, distinct from the 409 race case.
- Item 4's new regression test used a `provenance: 'cloned'` fixture to pin "no provenance guard on the retry route", but the retry route has no guard at all today, so the test was blind to the exact regression it claimed to prevent; it's now paired with an equivalent `provenance: 'designed'` case so either direction of an accidental future guard would be caught.
- The `/clone` route itself still returned a raw entry without applying `withComputedStaleness` (the transform item 1's frontend fix relies on the store now receiving) — it's now aligned with every other entry-returning route in `voice-library.ts`.
- `openapi.yaml` was stale on the 400/409 split for the transcript PATCH route; updated (and `api-types.ts` regenerated) to match.
- The release-notes pair (`docs/release-notes-next.md` + `RELEASE_NOTES.md`) is now complete for all four items plus the review-round fixes, not just item 3.

A [pass-2 re-review](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420448415) then found the `/clone` `withComputedStaleness` fix above had broken `voice-library.clone-fidelity.test.ts` (a stale fixture's `baseModel` no longer matched `currentQwenBaseModel()`) and left its own paired regression test unable to fail (a shared stub made both sides of its comparison always agree) — both fixed, plus a stale test-title comment, a duplicate 400-response branch, and a misleading "PUT has landed" comment in the debounce fix. A [pass-3 re-review](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420647338) came back CI-green with no new correctness bug, but caught the release-notes corrections themselves overclaiming a mechanism ("library refreshed in background") and a guarantee ("cast change has actually saved") that don't match the code — corrected to describe the real defect (a staleness misreport) and the real guarantee (the write has been issued, not landed).


dudarenok-maker:
## PR review — pass 1 (head 02bcab447ce7d721d73e384b2e6b5e50a1135f17, depth high)

Scope: all 7 changed files, plus the code the diff depends on but does not touch — `src/store/persistence-middleware.ts`, `src/store/settings-slice.ts`, `src/modals/clone-voice-wizard.tsx`, `src/store/generation-stream-middleware.ts`, `server/src/workspace/voice-library.ts` (`updateEntry`), `server/src/workspace/purge-clone-artifacts.ts`, `openapi.yaml`, `RELEASE_NOTES.md`, `CONTRIBUTING.md`. Depth `high` per the ladder: this PR mixes `fix`/`feat`/`test` across `frontend` and `server`.

Verified before probing: `updateEntry`'s `null`/`undefined` contract (`workspace/voice-library.ts:283-287`); that `withComputedStaleness` really is applied at only `:530`, `:720`, `:808/810` and not on `/clone` (`:1362`); that `PrimaryButton` really forwards `disabled` (`primitives.tsx:34-40`); that `cast/updateCharacter` really is a persist trigger (`persistence-middleware.ts:111`); that `requestStartGeneration` really enqueues synchronously (`generation-stream-middleware.ts:72`). Node modules are not installed in this review worktree, so the suites were not re-run — findings below are derived from the code, and each names the input that produces the wrong output.

---

### 🔴 Blocking — item 2's 600ms buffer is calibrated against a value the user can change, and is silently inert above 600ms

**Correctness bug.** `src/modals/clone-readiness-gate.tsx:238-239`

```js
// debounce window is 500ms; give a small buffer so the PUT is definitely in flight
setTimeout(() => setPendingCastWrite(false), 600);
```

The comment is false about the running app. The persistence debounce is **not** 500ms — it is `state.settings.autosaveDebounceMs`, read at flush-scheduling time (`src/store/persistence-middleware.ts:96-99`), with 500 only as `DEFAULT_DEBOUNCE_MS`. That setting is a first-class user knob clamped to **[100, 10 000] ms** (`src/store/settings-slice.ts:32-34`), exposed in Account → Advanced (`src/views/account.tsx:733-779`), and the repo's own tests set it to `2000` and `10_000` (`src/views/account.test.tsx:429,433`; `src/store/persistence-middleware.test.ts:92-93`).

So for every configured value above 600ms — 94% of the settable range — the fix does nothing at all, and does it invisibly.

Repro:
1. Account → Advanced → autosave debounce = `2000`.
2. Open a book whose cast has a `wrong-engine` cloned verdict; the gate opens on generate.
3. `t=0` — click "Cast on Coqui". `castActions.updateCharacter` dispatches; the middleware schedules the cast PUT for `t=2000`.
4. `t=600` — the hard-coded timer fires, `pendingCastWrite` → false, "Proceed anyway" re-enables and re-labels.
5. `t=610` — click "Proceed anyway" → `uiActions.requestStartGeneration()` → `generation-stream-middleware.ts:83` enqueues immediately and the runner POSTs.
6. `t=2000` — the cast PUT finally flushes.

That is bug #2649, unchanged, with a button that now looks like it guarded against it.

Fix direction: derive the buffer from `useAppSelector((s) => s.settings.autosaveDebounceMs)` rather than a literal, or — better, and what actually makes the claim true — gate on the persist having *landed* instead of on a sleep. The middleware already models per-slice pending/flush and already has a success-action idiom (`persistence-middleware.ts:88`, `bookMetaActions.commitDraftSucceeded`); a `cast`-slice equivalent removes the timing guess entirely.

### 🟠 Significant — a second "Cast on `<engine>`" click re-opens the race the first one closed

**Correctness bug.** `src/modals/clone-readiness-gate.tsx:236-240`

`handleCastPending` schedules a fresh `setTimeout` on every call and never clears the previous one. There is no ref, no id, no expiry timestamp. The **earliest** timer to fire wins and clears the flag for **all** in-flight writes.

This modal renders a *list* of verdicts (`:280-287`), and each `wrong-engine` row gets its own "Cast on `<engine>`" button (`:199-206`), so two of them is an ordinary case, not a contrived one. At the default debounce:

- `t=0` — click "Cast on Coqui" on Alice's row → timer A armed for `t=600`; cast PUT scheduled `t=500`.
- `t=550` — click "Cast on Coqui" on Bob's row → timer B armed for `t=1150`; `updateCharacter` **restarts** the debounce, so the PUT now flushes at `t≈1050`.
- `t=600` — timer A fires → `setPendingCastWrite(false)` → "Proceed anyway" enabled.
- `t=610` — click "Proceed anyway" → generation POST races Bob's cast PUT, still 440ms out.

Same shape for two clicks on the same row, and it composes with the 🔴 above.

### 🟠 Significant — a permanently master-less cloned entry now gets 409 with an error message that is factually false

**Correctness bug.** `server/src/routes/voice-library.ts:592-600` (the removed `|| !existing.master`) and `:697-701`.

The pre-lock check was deleted outright rather than kept alongside the new post-lock one, so the route no longer distinguishes *"the master vanished during this request"* (a genuine conflict, retryable) from *"this entry has no master and never will"* (a permanent client error).

Reachable outside the synthetic test: `POST /:voiceUuid/revoke` runs the consent-erasure purge, which sets `master: undefined` (`server/src/workspace/purge-clone-artifacts.ts:426`) while `provenance` stays `'cloned'`. So:

1. `POST /api/voice-library/lib-x/revoke` → master erased, permanently, by design.
2. `PATCH /api/voice-library/lib-x` `{"transcript":"anything"}`
3. **Before:** `400` — "`transcript` can only be set on a cloned voice with a master clip."
4. **After:** `409` — "`master` clip was removed before this transcript edit could be applied."

Two problems. The status class is wrong: 409 Conflict tells a client the state may be resolvable and the request resubmitted, and this one can never succeed on any retry. And the message asserts a mid-request removal that did not happen — the erasure was a deliberate, completed, prior operation. `src/modals/clone-readiness-gate.tsx:118` surfaces that string verbatim to the user (`setError((e as Error).message)`).

The existing test was rewritten from 400 to 409 to accommodate this (`server/src/routes/voice-library.test.ts:584`) — note its fixture never had a master at all, so the rename of that expectation *is* the behaviour regression, recorded rather than examined. The PR body does not mention it.

Fix: keep the permanent case pre-lock and add the race case post-lock, rather than replacing one with the other —

```ts
if (!existing.master) return res.status(400).json({ error: '`transcript` can only be set on a cloned voice with a master clip.' });
```

left in place at `:592`, with the new `conflict` → 409 path unchanged. Then `master` absent at both reads → 400, present at the first and absent at the second → 409, and each message is true of its own case.

### 🟠 Significant — item 4's "regression test" duplicates an existing test and is blind to the exact regression its comment names

**Correctness bug (test).** `server/src/routes/voice-library.test.ts:1102-1116` vs the existing `:895-909`.

The two tests are line-for-line the same exercise: same helper, same `engines: { qwen: { status: 'failed', baseModel: 'old' } }` override, same `POST .../engines/qwen/retry`, same `expect(res.status).toBe(200)`, same `expect(res.body.engines.qwen).toBeUndefined()`, same `expect('qwen' in (onDisk?.engines ?? {})).toBe(false)`. The only substantive difference is `expect(onDisk?.provenance).toBe('cloned')` — which is tautological: `makeClonedEntry` (`:869-893`) hardcodes `provenance: 'cloned'` and the route never writes that field.

The comment at `:1088-1101` states what the test is for: *"Do NOT add a provenance guard here to 'match' the PATCH route."* But a guard of that shape —

```ts
if (fresh.provenance !== 'cloned') return res.status(400)…
```

— leaves this test green, because its fixture **is** `cloned`. It leaves every other test in the block green too, because `makeClonedEntry` is the only fixture builder in the block. The regression the test was written to catch cannot redden it.

The comment also asserts a fact the suite never checks: *"it deletes a `failed` engine slot for `provenance === 'cloned'` exactly the same way it would for a `designed` entry."* No test in the file drives a `designed` entry through the retry route.

Fix: build the fixture with `provenance: 'designed'` (a designed voice with a failed qwen derive is realistic, and `makeEntry({provenance:'designed'})` is already used at `:573` with no consent record needed). That single change makes the test the only one in the block that a provenance guard would break — which is what "pins the asymmetry" means.

### 🟠 Significant — `openapi.yaml` still documents the old contract; the PR changed a wire status code and left the source of truth saying otherwise

**Chore (not a correctness bug in shipped code) — but the document is now false.** `openapi.yaml:2218-2236` and `:2251-2255`.

The `transcript` property description still reads *"Accepted only when the entry's `provenance` is `cloned` AND it carries a `master` clip; **400 otherwise**."* The `'400'` response description still scopes itself to *"the entry isn't a cloned voice with a `master` clip"*. There is no `'409'` response documented for `PATCH /api/voice-library/{voiceUuid}` at all — only `'200'`, `'400'`, `'404'`.

CLAUDE.md names `openapi.yaml` the API contract and source of truth for backend shapes, and lists a staled derived artifact as a chore owed in the same round. A client generated or hand-written against this spec will not have a 409 branch for a status the server now returns.

### 🟠 Significant — the `/clone` route still breaks the `withComputedStaleness` invariant its sibling handler declares universal; only the client was patched

**Correctness bug (server), one instance fixed / class left armed.** `server/src/routes/voice-library.ts:1362` vs `:705-719`.

The PATCH handler's own comment states the rule without qualification: *"Every route that hands an entry to the client must apply the same transform or the 'post-`withComputedStaleness`' contract is only true until the first write."* `POST /clone` returns `res.status(200).json(entry)` raw. The PR agrees this is the defect (`voice-library-slice.ts:199-202` says so explicitly) and then fixes the *consumer* instead of the *producer*.

The premise checks out — `engines.qwen.baseModel` is set from `derived.baseModel` (`:1354`), which can differ from `currentQwenBaseModel()` when the sidecar derived on a different tier, so `'ready'`-raw / `'stale'`-computed is reachable rather than theoretical. But the raw payload is still what `cloneVoice` returns, and `src/modals/clone-voice-wizard.tsx:50` still puts it straight into component state (`setSaved(entry)`). It happens to read only `name` and `sampleMeta` today (`:33`, `:133`), so nothing is visibly wrong right now — the contract is exactly what was protecting that, and it is still broken.

One line at `:1362` — `res.status(200).json(withComputedStaleness(entry))` — closes it at the source for every present and future consumer, and is strictly cheaper than the refetch round-trip the thunk now pays on every clone.

### 🟠 Significant — release-notes pair is incomplete and no not-applicable is declared

**Chore.** Before-shipping checklist step 5 requires an entry in `docs/release-notes-next.md` **and** a matching user-facing line in `RELEASE_NOTES.md`, landed PR-by-PR, or an explicit skip statement.

- `RELEASE_NOTES.md` is untouched. It has a live in-progress `# Castwright 1.15.0` section, so the bootstrap carve-out does not apply.
- `docs/release-notes-next.md` gained one entry covering **item 3 only** — the PR body says so itself ("item 3 only").
- Items 1 and 2 are user-visible and have no entry in either file: item 1 changes what the My-voices card shows immediately after a clone; item 2 adds a new visible button state ("Waiting for cast save…").
- The PR body does not declare the omission not-applicable.

### 🟡 Minor

- **The item-2 test cannot observe the thing it is named for.** `src/modals/clone-readiness-gate.test.tsx:317` — "preventing race". `makeStore` (`:54-71`) wires three reducers with default middleware: no persistence middleware, no `settings` slice, no PUT. The test asserts only that a `disabled` attribute is present for ~600ms and absent after; it would pass identically with a 1ms buffer or a 100 000ms one. This is why the 🔴 above shipped — the quantity that makes the fix correct is exactly the quantity the test does not look at. Asserting against `autosaveDebounceMs` (e.g. preload `settings: { autosaveDebounceMs: 2000 }` and assert the button is *still* disabled at t=700) would have caught it.
- **The same test is a flake shape.** Real timers, a 600ms window and an 800ms `waitFor` budget polling at 50ms leaves ~200ms of slack; a loaded runner overshoots that. `vi.useFakeTimers()` + `act(() => vi.advanceTimersByTime(600))` makes it deterministic and instant, and removes ~600ms of wall-clock from the frontend suite.
- **The pending timer is never cleared on close or unmount.** `clone-readiness-gate.tsx:236-240`. `onClose`/backdrop/Cancel dismiss the modal (`if (!gate) return null`, `:232`) while the timer runs on; the component is a persistent global overlay, so `pendingCastWrite` survives. Click "Cast on Coqui" at t=0, Cancel at t=10, reopen at t=20 → "Proceed anyway" reads "Waiting for cast save…" and is disabled until t=600 for no reason the user can see. A ref + `clearTimeout` in a `useEffect` cleanup fixes this and the 🟠 overlap finding together.
- **PR title under-declares its scope.** `fix(frontend)` on a PR that changes a server route's status code. CONTRIBUTING.md:285-291 has the syntax: `fix(frontend,server): …`. The title becomes the merge-commit subject, so the durable git record will say "frontend" for the API change.
- **Stray blank line.** `docs/release-notes-next.md` — the new entry adds a `+` blank line on top of the existing one, leaving a double blank before the `---`.
- **PR body's Test plan swaps its own item numbers.** It labels the `voice-library-slice.ts` revert "Item 2" and the `clone-readiness-gate.tsx` revert "Item 1", inverted relative to the Summary list directly above it.

### ✅ What is solid

- **Item 3's `conflict`-flag mechanism is correct.** `updateEntry` documents `undefined` → skip the write, resolve `null` (`workspace/voice-library.ts:283-287`), so returning `undefined` cannot write a garbage record. The flag is a per-request local, so two concurrent PATCHes on one uuid cannot cross-contaminate. `if (conflict)` is checked before `if (!written)` (`:697` before `:702`), so the conflict wins over a spurious 404. It genuinely mirrors the retry route's `entryFound`/`noop` pattern (`:769-791`), and the comment saying so is accurate.
- **The race test's spy strategy is faithful, and its header is right about why.** `voice-library.test.ts:606-618` — a `readEntry` spy cannot reach `updateEntry`'s internal read (intra-module binding, not the namespace object), so spying `updateEntry` and stripping master before delegating is the correct construction. It genuinely reddens on the old code (pre-lock `existing.master` present → passes; `fresh.master` absent → write skipped → 200).
- **Item 1's premise was verified, not assumed.** `withComputedStaleness` is applied at `:530`, `:720`, `:808`, `:810` and nowhere else; `/clone` at `:1362` is genuinely outside it.
- **Item 1's test asserts real behaviour.** `expect(entry).toEqual(computedEntry)` plus both slot statuses — not "didn't throw" — and `expect(api.listVoiceLibrary).toHaveBeenCalled()` is the assertion that reddens on the old code.
- **The disable is real, not cosmetic.** `PrimaryButton` forwards `disabled` to the DOM button *and* applies `pointer-events-none` (`primitives.tsx:34-40`), so the guard is not defeated by a synthetic click path.
- **The lock-message curation seam is correctly left alone.** The PATCH route takes no cast lock — `updateEntry`'s `withEntryLock` is a local promise chain with no `LockAcquisitionTimeoutError` — so it is correctly not one of the three `requestFailureMessage` sites in this file (`:1907`, `:2011`, `:2214`).
- **Item 1 matches its stated siblings exactly.** `revokeVoice` (`:212-219`), `promoteCharacterVoice` (`:183-193`) and the assign thunk (`:155-165`) all use the same `await dispatch(fetchVoiceLibrary())` shape, and none of them has a `fulfilled` reducer that writes the entry either. The consistency claim is true.

### Verdict

**Do not merge as-is.** One blocking correctness bug: item 2's fix is inert at any autosave-debounce setting above 600ms, which is most of the range the app itself offers, and it fails silently with a button that looks like it worked. Four further correctness bugs — the overlapping-timer re-arm, the 409/400 conflation with a false error string, the item-4 test that cannot catch the regression it names, and the `/clone` route still breaking the transform contract client-side-only — plus two chores owed in this round (`openapi.yaml`, the `RELEASE_NOTES.md` half of the pair).

The common thread across items 2 and 4 is worth naming: both tests pin a *symptom* (an attribute is present; a cloned entry returns 200) rather than the *quantity or the case* that makes the fix correct (the window outlasts the configured debounce; a `designed` entry is not rejected). Both would stay green through the exact regression they were written to prevent.

dudarenok-maker:
## PR review — pass 2 (head 82fb877144c95344bb10e242683c041b47cf68a0, depth high)

Re-review of [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420040875) (head `02bcab44`). Deltas only — solid items from pass 1 are not re-listed.

Scope: the 6 fix commits since pass 1 (`02bcab44`, `129d4fcd`, `0d422f87`, `819383a3`, `616a1fa1`, `82fb8771`), plus a fresh pass over `git diff main...HEAD` and the code the new changes reach: `server/src/routes/voice-library.clone-fidelity.test.ts`, `server/src/tts/model-paths.ts`, `src/store/settings-slice.ts`, `src/store/persistence-middleware.ts`, `src/components/layout.tsx`, `src/views/generation.tsx`.

Verified before probing: `withComputedStaleness`'s actual comparands (`voice-library.ts:507-518`); `QWEN_BASE_MODEL`'s default (`model-paths.ts:21`); that `settingsSlice.DEFAULT_AUTOSAVE_DEBOUNCE_MS` (500) and `persistence-middleware.DEFAULT_DEBOUNCE_MS` (500) agree; that `CloneReadinessGateModal` is rendered unconditionally at `layout.tsx:1847`; **and the PR's own CI at this exact head SHA** — `Server tests (fast + slow)` is **red** (run [32928267371](https://github.com/dudarenok-maker/Castwright/actions/runs/32928267371/job/98055520034)). Node modules are not installed in this review worktree, so suites were not re-run locally; the CI result at `82fb8771` is stronger evidence than a local run would be.

---

### 🔴 Blocking — the `/clone` `withComputedStaleness` fix breaks an existing test; the branch is red on a required check

**Correctness bug.** `server/src/routes/voice-library.clone-fidelity.test.ts:69` vs `server/src/routes/voice-library.ts:1383`.

CI at `82fb8771`:

```
FAIL  src/routes/voice-library.clone-fidelity.test.ts > POST /api/voice-library/clone — clone-fidelity advisory (#1945)
      > persists a warning AND still saves the clone when the embed boundary scores below CLONE_FIDELITY_MIN
AssertionError: expected 'stale' to be 'ready'
 ❯ src/routes/voice-library.clone-fidelity.test.ts:159:42
Test Files  1 failed | 535 passed | 1 skipped (537)
```

Exact chain:

1. `clone-fidelity.test.ts:69` stubs the derive with `baseModel: 'qwen3-0.6b'`.
2. `currentQwenBaseModel()` returns `QWEN_BASE_MODEL` = `'Qwen/Qwen3-TTS-12Hz-0.6B-Base'` (`model-paths.ts:21,48-50`).
3. `'qwen3-0.6b' !== 'Qwen/Qwen3-TTS-12Hz-0.6B-Base'` → `isArtifactVersionStale` true → `withComputedStaleness` rewrites the slot to `'stale'` (`voice-library.ts:510-512`).
4. `:1383`'s new `res.status(200).json(withComputedStaleness(entry))` therefore returns `'stale'` where the raw entry said `'ready'`, and `clone-fidelity.test.ts:159`'s `expect(res.body.engines.qwen.status).toBe('ready')` fails.

`verify.yml` is a required status check on `main`, so this blocks merge outright — independent of any judgment about which side is right.

**Which side is right matters, and it is the fixture.** `:159`'s intent is *"the fidelity warning is non-blocking — the clone still saves as a ready, cloned entry"*; it is not a staleness assertion. In production `deriveEngineArtifact` returns the real base model, which equals `currentQwenBaseModel()`, so a genuinely fresh clone computes `'ready'`. The fix is to make the stub represent a fresh clone — `baseModel: currentQwenBaseModel()` — exactly as `voice-library.test.ts:203-206` was changed to do **in this same PR**. Changing `:159` to expect `'stale'` would be wrong: it would enshrine a fixture artifact as the contract and destroy the "still saves as ready" claim the test exists to make.

**This is one instance fixed / the class left armed.** `git grep -l derive-engine-artifact server/src` returns exactly two route-level `/clone` harnesses: `voice-library.test.ts` and `voice-library.clone-fidelity.test.ts`. Both carried the same `baseModel: 'qwen3-0.6b'` string. Commit `819383a3` updated the first and missed the second — and the second is the one that went red. See also the finding below, which is the *reason* the first one's update did not surface as a warning sign.

### 🔴 Blocking — the paired regression test for that same fix cannot fail

**Correctness bug (test).** `server/src/routes/voice-library.test.ts:4645-4686`, enabled by `:203-206`.

The new test is the only regression protection this branch ships for its headline server fix. It ends with:

```ts
expect(voiceInList!.engines.qwen.status).toBe(cloneResponseStatus);
```

— comparing `POST /clone`'s status against `GET /`'s status for the same entry. That assertion is green **with or without** `withComputedStaleness` at `:1383`, because the same commit changed the file's shared derive stub:

```ts
// Use the actual current base model so cloned entries are fresh, not stale
const currentModel = modelPaths.currentQwenBaseModel();
deriveMock.mockResolvedValue({ ..., baseModel: currentModel });   // :204-206
```

`isArtifactVersionStale(current, current)` is false, so for every entry this file's `/clone` tests create, `withComputedStaleness` is the **identity function**. `/clone` raw → `'ready'`; `/clone` computed → `'ready'`; `GET /` → `'ready'`. Revert `:1383` to `res.status(200).json(entry)` and the test still passes.

The test's own comment names the missing arrangement without noticing:

> *"If /clone didn't apply withComputedStaleness, this would fail **when the baseModel becomes stale**."*

It never makes the baseModel stale. And the harness comment at `:204` states the guarantee that it never will: *"so cloned entries are **fresh, not stale**."* The trailing claim — *"Both must match, proving the transform was applied"* — is false: matching proves nothing when the transform is the identity on this input.

Note the interaction, because it is the whole story of commit `819383a3` and it answers the "does bundling three findings into one commit hide anything?" question directly. It is not an interaction *between* the three findings — it is that the single edit made to silence the transform's collateral damage in `voice-library.test.ts` (swapping the shared stub to the current base model) is the same edit that made the brand-new regression test inert, while the identical collateral damage in the sibling file was never looked at. The change that hid the signal and the change that broke CI are the same change.

Fix: give this test its own stale fixture rather than relying on the shared stub — e.g. `deriveMock.mockResolvedValueOnce({ ..., baseModel: 'an-old-base-model' })` before the `/clone` call (the string `voice-library.test.ts:849,1070` already uses for exactly this purpose), then assert `cloneRes.body.engines.qwen.status === 'stale'` outright. That reddens on `res.json(entry)` and stays green on `res.json(withComputedStaleness(entry))` — which is what "pins the transform" means. The `GET /`-agreement assertion can stay as a second, weaker check.

### 🟠 Significant — the release-notes pair now describes work this PR does not contain, and still omits the work it does

**Chore — but both documents are now false.** `RELEASE_NOTES.md:48` and `docs/release-notes-next.md:99`.

Both new entries are attributed to `#2068, PR #2666` and both claim:

> *"a new **"Preparing voice"** indicator on the Generate screen tells you when a voice is still being readied"*

`"Preparing voice"` already exists on `main` — `src/views/generation.tsx:1536` and `:1652`, with `server/src/tts/clone-voice-resolver.ts:345,835` behind it. **This PR does not touch `src/views/generation.tsx`** (see `git diff --stat main...HEAD` — 10 files, none of them that one).

It is worse than an inaccuracy: it is a **duplicate bullet in the same release section**. `RELEASE_NOTES.md:44` — four lines above the new entry — already announces it, for #1813:

> *"**A chapter that needs to rebuild a voice now shows you it's happening.** … The Generate screen now shows a "Preparing voice" step and names the character…"*

with its technical counterpart at `docs/release-notes-next.md:83` (*"`preparing-voice` phase for the resolver pre-pass … Closes #1813"*). So v1.15.0's notes now tell the reader about "Preparing voice" twice, the second time crediting the wrong PR. The same entries also claim:

> *"The pending-cast check also now looks at the engine your book will actually use, so a cast with cloned voices for Coqui no longer stalls on checking Qwen's slots first."*

There is no such change in this diff either. `docs/release-notes-next.md` compounds it with *"Two issues that shipped as one PR: … and the slow-to-revalidate pending gate"* — describing a different PR's content.

Meanwhile the one user-visible frontend change this PR actually makes — the `"Waiting for cast save…"` disabled state on "Proceed anyway" (`clone-readiness-gate.tsx:316-318`) — appears in **neither** file. `grep -n "Waiting for cast save" RELEASE_NOTES.md docs/release-notes-next.md` returns nothing.

This is pass 1's 🟠 release-notes finding: the files moved, but what landed is other work's copy, and the specific omission pass 1 named (*"item 2 adds a new visible button state ('Waiting for cast save…')"*) is unchanged. Rewrite both entries to describe (a) the clone refetch, (b) the "Waiting for cast save…" gate state, (c) the transcript-PATCH 400/409 split. Drop the "Preparing voice" claim outright — `:44` already covers it — and drop the pending-cast-engine claim, or move it to whichever PR shipped it.

### 🟡 Minor

- **A comment cites a test title this PR renamed out from under it.** `server/src/routes/voice-library.test.ts:1093` reads *"The sibling PATCH handler above (`"rejects `transcript` on a cloned entry with no master clip with 409"`)"*. That test is now titled `…with 400` (`:586`) — `129d4fcd` correctly restored 400 after `02bcab44` wrote the comment against the 409 wording, and the comment was not carried along. The citation names a test that no longer exists.
- **The pre-lock 400 is now two identical branches.** `voice-library.ts:591-601` — `existing.provenance !== 'cloned'` and `!existing.master` return the same status with the byte-identical message, so the split is exactly equivalent to `main`'s single `if (existing.provenance !== 'cloned' || !existing.master)`. If the intent was to let the two cases diverge later, they do not diverge now; if not, this is one condition written twice. (The behaviour is correct either way — this is the shape, not a defect.)
- **The debounce comment claims more than the code delivers.** `clone-readiness-gate.tsx:250-252` — *"wait for the debounce window plus a safety margin to ensure the cast write PUT has **landed**"*. `autosaveDebounceMs + 100` fires 100 ms after `persistence-middleware.ts:450`'s flush *starts* `api.putBookState(...)`; it says nothing about the response arriving. The code this replaced was honest about that (*"so the PUT is definitely **in flight**"*). Over LAN HTTPS against a box that may be mid-generation, a 100 ms round-trip is not a safe assumption. The narrowed window is a real improvement and pass 1 sanctioned this direction — the comment should just say what it does.
- **The unmount cleanup added for pass 1's timer finding cannot run in the app.** `clone-readiness-gate.tsx:236-242` is `useEffect(…, [])`, so it fires only when `CloneReadinessGateModal` itself unmounts. It is rendered unconditionally at `layout.tsx:1847`, and the file's own comment at `:84-85` says so — *"this modal is a global, always-mounted overlay"*. The gate closes via `if (!gate) return null` (`:244`), which is not an unmount. So the cleanup is live only under test teardown, and the half pass 1 actually described — *"never cleared on **close**"* — is unchanged: Cancel at `t=10`, reopen at `t=20`, and "Proceed anyway" still reads "Waiting for cast save…". Benign (the window is ≤ `debounce+100 ms` and the state converges), but the code now reads as if close were handled.
- **The release-notes blank-line cleanup over-corrected.** `docs/release-notes-next.md:213-214` — `82fb8771` removed *both* blank lines before that `---`, including the one `origin/main` already had. It is now the only one of the file's ten `---` separators (lines 62, 204, **214**, 235, 251, 303, 310, 316, 354, 369) not preceded by a blank line. CommonMark still renders it as a thematic break (a `---` at column 0 after a list closes the list, and a document-level `---` with no open paragraph is `<hr>`), so this is convention drift rather than broken output — but it is drift this PR introduced against both the file's own pattern and `main`.

| Pass 1 finding | Disposition |
|---|---|
| 🔴 600 ms buffer inert above 600 ms | **VERIFIED RESOLVED** — `:227` reads `s.settings?.autosaveDebounceMs`, `:256` uses `autosaveDebounceMs + 100`. The paired test at `clone-readiness-gate.test.tsx:395-449` genuinely reddens pre-fix: at `autosaveDebounceMs: 2000`, the old hard-coded `600` would have re-enabled the button before `:430`'s assertion at `t=1500`. |
| 🟠 Second click re-opens the race | **VERIFIED RESOLVED** — `pendingTimeoutRef` + clear-before-reschedule (`:253-257`). Test at `:451-498` reddens pre-fix: click at `t=0` and `t=300`, old timer A fires at `t=600`, so `:491`'s `toBeDisabled()` at `t=700` fails. |
| 🟠 Permanent master-less entry got a false 409 | **VERIFIED RESOLVED** — 400 restored pre-lock (`voice-library.ts:591-601`), 409 scoped to the genuine race (`:710-714`), and the test at `voice-library.test.ts:586-607` now asserts the exact 400 message rather than only the status. Two 🟡 riders above. |
| 🟠 Item-4 test blind to the regression it names | **VERIFIED RESOLVED** — the `provenance: 'designed'` companion at `voice-library.test.ts:1125` is the case a `provenance !== 'cloned'` guard would redden. The asymmetry is now genuinely pinned in both directions. |
| 🟠 `openapi.yaml` stale on the 400/409 split | **VERIFIED RESOLVED** — `openapi.yaml:2222-2275` documents both, adds the missing `content` schema on 400, and `src/lib/api-types.ts:8776-8856` is a faithful regeneration (no hand edits). |
| 🟠 `/clone` breaks the `withComputedStaleness` contract | **NOT CLOSED.** The one-line producer fix is correct and present at `voice-library.ts:1383` — but it ships CI red (🔴 #1) and with a paired test that cannot fail (🔴 #2). The code change is right; nothing on this branch proves it, and one thing on this branch disproves the branch. |
| 🟠 Release-notes pair incomplete | **STILL OPEN** — see 🟠 above. Entries landed, but they describe `"Preparing voice"` and a pending-cast engine change that are not in this diff, and still omit `"Waiting for cast save…"`. |
| 🟡 Item-2 test can't observe the quantity | **VERIFIED RESOLVED** — the suite now asserts against `autosaveDebounceMs` at two values (500 and 2000), which is exactly the quantity that makes the fix correct. |
| 🟡 Test is a real-timer flake shape | **VERIFIED RESOLVED** — `vi.useFakeTimers()` + `act(() => vi.advanceTimersByTime(…))` (`:326-333`). Deterministic, and ~600 ms of wall-clock removed. |
| 🟡 Timer never cleared on close or unmount | **PARTIALLY RESOLVED** — unmount cleanup added but unreachable in the app; the close half is unchanged. See 🟡 above. |
| 🟡 PR title under-declares scope | **VERIFIED RESOLVED** — `fix(frontend,server): …`. |
| 🟡 Stray blank line in `release-notes-next.md` | **RESOLVED, over-corrected** — see 🟡 above. |
| 🟡 PR body Test plan swaps its item numbers | **VERIFIED RESOLVED** — Item 1 → `voice-library-slice.ts`, Item 2 → `clone-readiness-gate.tsx`, matching the Summary order. |

### Verdict

**Do not merge.** Two blocking correctness bugs, both from `819383a3`, and both about the same one-line server fix: the branch is **red on `Server tests (fast + slow)` at this head SHA** because the `/clone` transform flips `clone-fidelity.test.ts`'s stale-stubbed fixture to `'stale'`, and the transform's own paired regression test is inert because the shared stub in the *other* harness was moved to the current base model. One 🟠 chore remains open — the release-notes pair describes another PR's work and omits this one's.

Worth naming, because it is the same thread pass 1 closed on: pass 1 said both item-2 and item-4's tests *"pin a symptom rather than the quantity or the case that makes the fix correct."* Item 2's and item 4's tests were genuinely fixed — the two frontend timer tests and the `designed` companion all redden on the pre-fix code, which is the bar. The `/clone` fix, added in the same round to close pass 1's own finding, then reproduced the identical shape one level down: it asserts that two responses *agree*, under a harness rigged so they agree unconditionally. The five other findings landed clean; this one needs another turn.

Also flagged, not a finding: the four `E2E (chromium)` shards were still pending when this pass ran. They exercise the mock API (`VITE_USE_MOCKS=true`), so the server change cannot reach them — but the run is not complete, and the red server leg above is from the completed portion.

dudarenok-maker:
## PR review — pass 3 (head `fd2f4391`, depth high)

Final re-review under this skill's loop cap (initial + 2 re-review rounds). Follows [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420040875) (`02bcab44`) and [pass 2](https://github.com/dudarenok-maker/Castwright/pull/2666#issuecomment-5420448415) (`82fb8771`). **Deltas only** — items already settled in pass 1 are not re-listed.

Scope: the 3 commits since pass 2 (`9f164539`, `3dbe17e7`, `fd2f4391`), plus a fresh pass over `git diff main...HEAD` (11 files) and the code the fixes reach — `server/src/tts/model-paths.ts`, `server/src/tts/clone-engines.ts` (`isArtifactVersionStale`), `server/src/workspace/voice-library.ts` (`updateEntry`), `src/store/start-generation-flow.ts`, `src/modals/clone-voice-wizard.tsx`, `src/store/settings-slice.ts`.

**CI at this exact head SHA is fully green.** `Verify` run [32929984310](https://github.com/dudarenok-maker/Castwright/actions/runs/32929984310) is `completed / success` — every leg, including `Server tests (fast + slow)` (3m41s), the leg that was **red** at pass 2's head. `Lint, typecheck & fast checks`, `Frontend tests + a11y`, all four E2E shards, visual baselines, Build, Sidecar, PR-title and PR-issue-link checks all pass. (`gh pr checks` served a stale `pending` for E2E shard 1/4 for several minutes after the job itself completed — the job API reports `completed / success`, started `04:23:26Z`.) Node modules are not installed in this review worktree, so suites were not re-run locally; the completed CI run at `fd2f4391` is stronger evidence than a local run.

---

### Disposition of pass 2's findings

| Pass 2 finding | Disposition |
|---|---|
| 🔴 `/clone` transform breaks `clone-fidelity.test.ts`; branch red on a required check | **VERIFIED RESOLVED** — see below |
| 🔴 The paired regression test for that fix cannot fail | **VERIFIED RESOLVED** — see below |
| 🟠 Release-notes pair describes another PR's work, omits this one's | **PARTIALLY RESOLVED — the false claims are gone, a new one replaced them.** See 🟠 below |
| 🟡 Comment cites a test title this PR renamed | **VERIFIED RESOLVED** — `voice-library.test.ts:1093` now reads `"…with 400"`, matching the actual title at `:585` |
| 🟡 Pre-lock 400 is two identical branches | **VERIFIED RESOLVED** — `3dbe17e7` merged them to `existing.provenance !== 'cloned' \|\| !existing.master` (`voice-library.ts:591`). Byte-identical message on both removed branches, so the merged condition is exactly equivalent — and exactly `main`'s. No behavioural delta |
| 🟡 Debounce comment claims "landed" | **RESOLVED IN CODE, RE-INTRODUCED IN DOCS** — `clone-readiness-gate.tsx:243-244` now correctly says "has been **issued**" with an explicit "actual PUT completion is asynchronous" rider. The identical overclaim now sits in both release-notes files instead. Folded into the 🟠 below |
| 🟡 Unmount cleanup unreachable in the app | **RESOLVED (dead-code half)** — the `useEffect` block and its `useEffect` import are gone (`clone-readiness-gate.tsx:34`); no residual `useEffect` in the file, no orphaned import, and no test asserted the cleanup, so nothing is left dangling. **The close half is unchanged and knowingly accepted**: `onClose` still does not clear `pendingTimeoutRef`, so Cancel at `t=10` / reopen at `t=20` still shows a disabled "Waiting for cast save…". Benign as pass 2 assessed — the component is always mounted and the state converges in ≤ `debounce+100 ms`. Recording as accepted, not re-raising |
| 🟡 Release-notes blank-line over-correction | **VERIFIED RESOLVED** — `docs/release-notes-next.md:213` restored; the separator now matches `main` byte-for-byte and the file's nine sibling `---` separators |

**Detail on the two 🔴s, because "made green" and "fixed" are different things:**

**`clone-fidelity.test.ts`** — `:62` adds `const modelPaths = await import('../tts/model-paths.js')` and `:70-72` stubs `baseModel: modelPaths.currentQwenBaseModel()`. `model-paths` is **not** `vi.mock`'d in this file, and `vi.resetModules()` at `:52` precedes both dynamic imports, so the test and `voice-library.ts` resolve the same module instance — the stub and the transform's comparand cannot disagree. `isArtifactVersionStale(current, current)` is false (`clone-engines.ts:320`), so the transform is the identity and `:162`'s `expect(res.body.engines.qwen.status).toBe('ready')` holds. Critically, **the test still asserts its own claim** — `:159`'s "Non-blocking: the clone still saves as a ready, cloned entry" — rather than being bent to `'stale'` to accommodate the transform. The fixture now represents a fresh clone, which is what production produces. Correct fix, not a green-washing.

**The staleness regression test** (`voice-library.test.ts:4645-4696`) is now **genuinely non-vacuous**. `:4664-4668` stubs `deriveMock.mockResolvedValueOnce({ …, baseModel: 'some/outdated-base-model' })`; `beforeEach`'s `deriveMock.mockReset()` (`:203`) clears the once-queue per test and `/clone` calls derive exactly once, so the override is consumed by this request. The route writes `engines.qwen = { status: 'ready', baseModel: derived.baseModel || currentQwenBaseModel() }` (`:1362`) → raw `'ready'`; `isArtifactVersionStale('some/outdated-base-model', 'Qwen/Qwen3-TTS-12Hz-0.6B-Base')` → true → computed `'stale'`. So `:4682`'s `expect(cloneRes.body.engines.qwen.status).toBe('stale')` **fails outright** if `:1378` is reverted to `res.status(200).json(entry)`. That is the assertion pass 2 asked for, and it reddens for the reason claimed.

Also re-verified since the mechanism was touched: `updateEntry` invokes its `mutate` callback **exactly once** inside `withEntryLock` with no retry (`workspace/voice-library.ts:294-313`), so the `conflict` flag cannot be set spuriously by a re-entry; and the `!master && !existing.master` inner branch is unreachable (the pre-lock 400 at `:591` guarantees `existing.master`), so its `404` fall-through is defensive, not live.

---

### 🟠 Significant — `RELEASE_NOTES.md:48` replaced pass 2's false claims with a *different* false claim

**Chore — not a correctness bug in shipped code, but the user-facing document is wrong about what this PR fixed and why.**

`fd2f4391` correctly deleted the "Preparing voice" and engine-aware-pending-cast claims and correctly added "Waiting for cast save…". Its replacement text, however, asserts a mechanism that does not exist:

> When you cloned a voice, **your library was refreshed in the background** — and if you started a render in the gap before that finished, your cast would be rejected because **the freshly-cloned voice was nowhere the library could see yet**.

Neither half is true of `main`:

1. **There was no background refresh.** `main`'s `cloneVoice` was `async (body) => api.cloneVoice(body)` and `cloneVoice.fulfilled` upserted the payload into `state.entries` **synchronously** — both visible as this PR's own removed lines in `src/store/voice-library-slice.ts`. The entry was in the library the instant the POST resolved. `git grep 'fetchVoiceLibrary('` over `src/` returns only `my-voices-section.tsx:30` (mount), the focus/visibility listener (`voice-library-slice.ts:349`, itself gated on `entries.length > 0` **and** `lastFetchedAt` staleness), the sibling thunks, and `start-generation-flow.ts:79`. None of them is triggered by cloning.
2. **The render path cannot read a pre-refresh library at all.** `start-generation-flow.ts:79` does `await dispatch(fetchVoiceLibrary()).unwrap()` — awaited, with its own fail-open catch — *before* `:86` evaluates `selectCloneReadinessGateShouldFire`. So "started a render in the gap → cast rejected because the voice was invisible" is not a reachable state on `main`, and is not what the fix addresses.

What the fix *does* address is stated correctly by the code's own comments and by the technical note, and is absent from the user-facing one: the raw `/clone` payload's `qwen` slot can read `'ready'` where the server's computed view says `'stale'` (`voice-library-slice.ts:199-202`, `voice-library.ts:1371-1377`), so any surface reading `state.entries` between the clone and the next refetch — the My-voices card — can show a stale slot as ready.

**The second sentence overclaims in the way pass 2 already had corrected once:**

> "Proceed anyway" now waits … until that cast change **has actually been saved**, so **a render can no longer start against a cast the server hasn't recorded yet**.

- "has actually been saved" is the exact wording pass 2 got fixed in `clone-readiness-gate.tsx`. `:248` waits `autosaveDebounceMs + 100`, which covers the debounce **firing** — `persistence-middleware` *starts* the PUT — not its response. `docs/release-notes-next.md:99`'s own headline repeats it ("before a just-made cast change **has actually saved**"), though its body bullet is accurate.
- "a render can no longer start against a cast the server hasn't recorded yet" is false on a second path as well. The guard is local to the "Proceed anyway" button; Cancel / backdrop / ✕ close the gate without clearing the pending timer or blocking anything. Clicking **Generate** again re-enters `startGenerationFlow`, and if the re-cast cleared the verdicts, `:93` dispatches `requestStartGeneration()` with the cast PUT still debounced. Tight at the 500 ms default, wide open at the 10 000 ms maximum the setting allows. (Not raising the code scope as a finding — pass 1 explicitly sanctioned the timer approach — but the note must not claim a universality the code does not have.)

**Fix:** describe the mechanism the code implements. Something of the shape: *a freshly cloned voice could show as ready in My voices when the server already considered its engine slot stale — the library is now re-read from the server the moment a clone finishes; and if you've just re-cast a character right before generating, "Proceed anyway" waits out your autosave delay before it will start a render.*

### 🟡 Minor

- **The staleness test's header comment still describes the disproven proof strategy.** `voice-library.test.ts:4631-4643` still reads *"by comparing: (1) the engines.qwen.status from the /clone response, and (2) the same entry's status when fetched via GET /. Both must match, proving the transform was applied."* That is precisely the reasoning pass 2 showed to be invalid ("matching proves nothing when the transform is the identity on this input"). The test **body** was correctly rewritten to assert `'stale'` outright, and the agreement check survives only as a weaker secondary assertion at `:4699` — but the comment still presents the weak check as *the* proof. A reader repairing this test later would restore the vacuous version from its own docs.
- **`docs/release-notes-next.md:99` cites an unreachable field.** *"the raw entry's `baseModel`/`coquiVersion` could read `'ready'`"* — `POST /clone` only ever writes a `qwen` slot (`voice-library.ts:1362`: `engines: { qwen: { … } }`); it never creates an `xtts` slot, so the `coquiVersion` half of `withComputedStaleness` (`:514`) cannot be reached through this route. The `baseModel` half is correct.
- **The PR body now describes a mechanism that was deliberately removed.** Its "Review-round fixes" section still says item 2's fix *"clears any prior pending timeout … and cleans up on unmount"* — `9f164539` removed the unmount cleanup as unreachable, on pass 2's finding. The Summary bullet also still says *"for a buffer past the 500ms persistence debounce"*, superseded by the configurable-debounce fix in `0d422f87`. The PR body becomes the merge-commit body, so this is the durable record.

### House gates re-checked at this head

- **On-box acceptance: not applicable.** No hardware-provable behaviour ships here — the whole diff is client-store, route, and status-code logic, fully covered by the automated suites. No register / run-sheet / live-view row is owed, and none was added.
- **Derived artifact:** `src/lib/api-types.ts:8776-8856` is a faithful regeneration of the `openapi.yaml:2222-2275` edit — the 409 block, the added 400 `content` schema and the reflowed descriptions all match; no hand edits.
- **`Closes #2068`** present in the body, outside any code span; the `pr-issue-link` check passes.
- **`cast.json` locking:** unchanged — the PATCH route takes no cast lock (`updateEntry`'s `withEntryLock` is a local promise chain, no `LockAcquisitionTimeoutError`), so no curation seam is owed. Correctly untouched.
- **Testing discipline:** every changed behaviour carries a paired test that reddens pre-fix. Verified individually for the two frontend timer tests, the overlapping-click test, the clone-thunk refetch test, the 400/409 split, the `designed` retry companion, and the staleness test above. No test was deleted or skipped without replacement.

### Verdict

**No new correctness bug in shipped code.** The two blocking findings from pass 2 are genuinely closed — not merely made green — and the branch that was red at `82fb8771` is fully green at `fd2f4391`. The one 🟠 is a **documentation chore**: `RELEASE_NOTES.md:48` is now on its third telling of what this PR does and is still wrong, having swapped one non-existent mechanism for another and re-imported the "issued vs. landed" overclaim that pass 2 had already corrected one file away. Since it ships no behaviour, it needs a text edit and a spot-check, **not** a fourth review round — and the loop cap is reached regardless.

Worth naming, because it is the thread running through all three passes: every round, the *code* fix landed correctly and the *description of it* did not. Pass 1's test pinned a symptom instead of the quantity. Pass 2's test asserted an agreement that was unconditional. Pass 3's release note asserts a failure mode that was never reachable. The artifact that keeps failing is the one nothing executes.

**Recommendation: fix `RELEASE_NOTES.md:48`, `docs/release-notes-next.md:99`'s headline clause, the `voice-library.test.ts:4631-4643` comment, and the two stale PR-body sentences — then merge.** None of them can redden a suite, so a re-review would add no signal.

Issue #2656, opened 2026-08-25
Determine whether stranded VRAM is a genuine leak or just the resident-model floor


dudarenok-maker:
## What

Successor to #1976 and #1996 (both closed as superseded — see those issues
for the full measurement history). This issue is the one concrete thing
still undetermined: **is there a genuine reference leak sitting on top of
the resident-model VRAM floor, or does residency alone explain everything
#1976 originally measured as "stranded"?**

## Where the prior investigation left off

The final genuinely-idle measurement run (#1996, 2026-08-25,
`docs/testing/1996-stranded-vram-measurement.md` @ `218cc562`) found:

- P2/P3 (idle, 21s apart) were byte-identical — no self-heal.
- P4's `/debug/reclaim` recovered only 6.4% (372 MB of 5853 MB) — not a
  cache/fragmentation story.
- At P2/P3, `qwen.base_loaded=true` and `whisper.model_loaded=true` — both
  legitimately resident. Qwen Base 0.6B has **no idle TTL** (button-driven,
  evicts only on explicit `/unload`); Whisper's 120s TTL had only 21s
  elapsed.

Conclusion at the time: the 5.45 GB looks exactly like "the two resident
models' own live weights — normal residency, not evidence of a leak on top
of it" — but the run **could not prove that**, because it never captured
what `allocated` looks like immediately after those models are actually
unloaded.

## The actual work needed to close this permanently

1. Re-run (or extend) the idle-gated measurement, but add one more point:
   issue an explicit `POST /unload {qwen}` (and let Whisper's TTL lapse, or
   force it) at the same idle point, then read `/debug/memory` again.
2. Diff that post-unload `allocated`/`reserved` against the P3 baseline from
   the 2026-08-25 run.
   - If post-unload `allocated` drops to near-zero (matching the known
     weight sizes for Qwen Base 0.6B + Whisper), the resident floor **fully
     explains** the prior "stranded" readings — #1976's original report was
     measuring normal residency, not a leak. Close as working-as-intended,
     correct #1976/#1996's language for the record, no lever needed.
   - If a residual gap remains after both known models are confirmed
     unloaded, that gap is the actual leak — needs its own root-cause pass
     (likely back in `PlacementController`/engine-unload paths per the
     dead-ends already ruled out in #1996's design-pass comments).
3. Record the result in `docs/testing/1996-stranded-vram-measurement.md`
   (extend, don't replace) so the full chain of attempts stays legible.

[...]

Issue #2310, opened 2026-08-13
srv — an HTML entity that opens dialogue is read aloud verbatim by the TTS engine


dudarenok-maker:
An HTML entity that survives `stripHtml` is recognised as a dialogue opener but is then **read aloud verbatim**. Surfaced by #2289 (which added `&ndash;` to the `es`/`fr` `dialogueOpen`), but it is not caused by it — it is equally true of `&mdash;` in `es`/`fr` and of both entities in `ru` today.

## The chain, traced rather than assumed

1. `server/src/parsers/html-utils.ts` — `stripHtml` decodes only `&nbsp;`, `&lt;`, `&gt;`, `&quot;`, `&amp;` plus numeric references. **`&ndash;` and `&mdash;` survive into the body text**, which is exactly why the `dialogueOpen` tables carry the literal entity spellings at all.
2. `server/src/analyzer/dialogue-structure/lang/{es,fr,ru}.ts` — `dialogueOpen` matches the entity, so the paragraph is correctly classified as dialogue and the speaker is attributed. **This part works, and #2289 completed it for `es`/`fr`.**
3. `server/src/tts/text-normalize.ts:57` — `LEADING_DASH = /^\s*[–—]\s*/` matches the **literal glyphs only**. The entity is not softened to the `'... '` pause the real dash gets.
4. `server/src/tts/normalize/index.ts` (see the comment at :83) — `&` is expanded only as a space-delimited standalone token, so `AT&T` survives; `&ndash;` therefore passes through untouched too.

Net effect: a Spanish or French manuscript whose dash dialogue survived EPUB conversion as an entity gets **correct speaker attribution and a corrupted first few syllables on every dialogue line**, where a literal `—` would have produced a clean pause.

## Why this is filed rather than fixed in #2289

Per CLAUDE.md, the one class of finding that defers is one needing a design pass — and it must name the decision. **The decision here is where the entity should be decoded**, and there are at least three defensible answers with different blast radii:

1. **Widen `stripHtml`'s named-entity set** (`html-utils.ts`). Fixes every downstream consumer at once, and would let the `dialogueOpen` tables drop their entity alternatives entirely. Largest blast radius — `stripHtml` feeds the whole import path, and the deliberately small entity set may be load-bearing for reasons worth checking before widening.
2. **Decode at import-time normalisation**, leaving `stripHtml` alone. Narrower, but adds a second place that knows about entities.
[...]

PR #2688, opened 2026-08-26
fix(server): collapse doubled comma from softenDashes (#2059)

### LINKED ISSUE #2059: Attribution-dialogue leading+interior dash produces a doubled comma (,,) — pre-existing, should it collapse?
### Summary

`softenDashes()` (`server/src/tts/text-normalize.ts`) produces a doubled comma (`,,`) on the dominant Russian dialogue-attribution shape — a line that opens with a dash AND has a mid-sentence attribution dash. This is **pre-existing behaviour, not a regression** introduced by #2050's leading-dash fix (confirmed: the prior code produced the same doubled comma too, since it ran the identical ", " substitution for every dash regardless of position). Filed to make the shape visible rather than accidental — pinned by a new test in #2050, behaviour intentionally left unchanged there per that PR's scope.

### Steps to reproduce

Run `softenDashes` (or the composed `normaliseForTts`) on a Russian dialogue-attribution line, e.g.:

```
— Привет, — сказал Антон.
```

### Expected

Unclear — that's the open question this issue tracks. Candidate: collapse `,,` to a single `,`.

### Actual

Produces `... Привет,, сказал Антон.` — note the doubled comma with no space between them.

### Why this might matter beyond cosmetics

This dash-open / comma / dash-attribution shape (`— <line>, — <verb> <name>.`) is the **dominant Russian dialogue pattern** — most Russian dialogue lines in a typical manuscript will hit it. Malformed punctuation (a doubled comma with no space) on the majority of a book's dialogue lines is a plausible contributor to the broader Russian degeneracy symptoms tracked under #2026, though this has not been tested in isolation and is not claimed as a confirmed cause.

### Environment

Any Coqui/Gemini TTS synthesis path, Russian (`ru`) text with dialogue attribution.

### Severity

low (cosmetic punctuation artifact on the wire text; not confirmed to affect audio quality on its own)

### Reproducible?

always

### Depends on / related

#2026, #2050 (pins this shape with a new test in `server/src/tts/text-normalize.test.ts` without changing behaviour)

dudarenok-maker:
## Summary

Fix doubled commas produced by `softenDashes()` in the TTS text normalizer.

When `softenDashes()` replaced em/en-dashes with commas, a comma already
adjacent to a dash produced a doubled comma like `"hello,, world"` — most
visible in Russian (also French/Spanish) dash-attributed dialogue, e.g.
`"— Привет, — сказал Антон."`.

Also fixed, found in passing during the mandatory PR review gate (three
review rounds; see PR comments for full detail):

- The collapse regex is `.replace(/,(\s*,)+\s*/g, ', ')`, collapsing an
  arbitrary comma run in one pass — the original two-comma-only regex left
  a run of 3+ commas only partially collapsed, breaking the module's stated
  idempotency contract.
- `normaliseForTts`'s pass order changed so `stripAudioTags` runs before
  `softenDashes` — an inline audio tag (e.g. `[shouting]`) sitting between
  two commas previously defeated the collapse entirely.
- That reordering itself broke idempotency for a trailing dash (nothing
  left to trim the trailing whitespace `softenDashes` leaves behind) —
  fixed by re-running the whitespace-collapse-and-trim after `softenDashes`
  in `normaliseForTts`.
- A mislabeled test (claimed to guard against "over-collapsing" a single
  comma, when it was actually a positive test of the collapse firing) was
  renamed and a real test added documenting that the collapse has no
  dash-awareness at all — any doubled comma collapses, dash-produced or
  literal.
- Stale planning docs, the on-box acceptance register (new row A37), and
  the release-notes pair were updated to reflect the shipped fix.

The comma-collapse has **no dash-awareness** — it fires on any doubled or
tripled comma, whether or not a dash produced it.

## Changes

- `server/src/tts/text-normalize.ts`: doubled/tripled-comma collapse in
  `softenDashes()`, `stripAudioTags` reordered before it in
  `normaliseForTts`, and a final whitespace-collapse-and-trim restored
  after `softenDashes` to keep the composed pipeline idempotent.
- `server/src/tts/text-normalize.test.ts`: regression tests for the audio-tag
  case, 3+/4+ comma runs, the corrected test naming, and the trailing-dash
  idempotency case.
- `docs/features/277-v115-bug-chore-sweep.md`, `...-board.html`: mark #2059
  shipped rather than deferred.
- `docs/testing/onbox-acceptance-register.md` (+ its live-view mirror): new
  row A37 for the fix's unconfirmed audible effect.
- `RELEASE_NOTES.md`, `docs/release-notes-next.md`: release-notes pair.

## Testing

- `server/src/tts/text-normalize.test.ts`: 51 tests pass (GREEN).
- Each fix verified red-before-fix against its own regression test.
- `npm run check:onbox-register` passes (register/live-view agree).
- Typecheck clean.

Closes #2059


dudarenok-maker:
## PR review — pass 1 (head 7a1e57350bb516c3414b138803c6549f8dfece83, depth medium)

Reviewed the PR head against `main`, read `server/src/tts/text-normalize.ts` in full (all four passes and their composition order in `normaliseForTts`), and ran the shipped regex plus four mutants against every `softenDashes` assertion in the suite. CI is green on all 15 checks.

The collapse works for the exact shape #2059 filed. It does not hold for two adjacent shapes, and the test that the PR body advertises as proving the fix's scope does not prove it.

---

### 🟠 Significant

**1. An audio tag between the comma and the dash defeats the fix entirely — `server/src/tts/text-normalize.ts:179` (correctness bug)**

`normaliseForTts` runs `softenDashes` *inside* `stripAudioTags`, so the tag is still present when the collapse pass runs and is only removed afterwards. The tag separates the two commas, `/,\s*,\s*/` does not match across it, and `stripAudioTags` then removes the tag and leaves the doubled comma behind.

```
input   "— Привет, [shouting] — сказал Антон."
base    "... Привет, , сказал Антон."
PR head "... Привет, , сказал Антон."     <-- identical to base; fix did not fire
```

Same in English: `"Stop, [whispers] — he said."` → `"Stop, , he said."` on both base and PR head.

This is not a contrived input. `AUDIO_TAGS` (`emphatic`/`shouting`/`whispers`/`laughs`/`sighs`/`excited`/`hesitant`) are analyzer-emitted, ride inside `sentence.text` all the way to the TTS boundary per this file's own header comment, and the detectors that emit them (`tagShoutingDialog`/`tagExcitedDialog`/`tagHesitantDialog` in `server/src/parsers/audio-tags.ts`) fire on **dialogue lines specifically** — which is exactly where the Russian attribution-dash shape lives. The tagged half of the dominant shape is left unfixed. Catalogue shape #10: one instance fixed, the class left armed.

**2. A run of 3+ commas collapses only partially, and breaks the file's stated idempotency contract — `server/src/tts/text-normalize.ts:131` (correctness bug)**

`/,\s*,\s*/g` consumes both commas plus trailing whitespace per match and then resumes *after* the match, so it processes a comma run in non-overlapping pairs while its own replacement re-emits a comma. An odd-length run leaves a stranded comma:

```
"a,,,b"                     -> "a, ,b"      (2nd pass -> "a, b")
"a,,,,b"                    -> "a, , b"     (2nd pass -> "a, b")
"— Привет, — — сказал."     -> "... Привет, , сказал."   (2nd pass -> "... Привет, сказал.")
"a, — , — b"                -> "a, , b"     (2nd pass -> "a, b")
```

Two distinct problems here:

- The output `, ` `,` is still a malformed punctuation run reaching the decoder — the same class of artifact the PR set out to remove, just respaced.
- `server/src/tts/text-normalize.ts:33` states the contract **"All transforms are idempotent and order-independent."** `softenDashes` is no longer idempotent, and none of the four existing idempotency tests catches it — `text-normalize.test.ts:72` uses `'a—b'`, `:121` uses `'— Кто бы это ни был.'`, `:222` uses `'THE HAZE—then.'`; all have at most one dash and no comma run, so all pass vacuously on this defect.

A single-bite collapse fixes both and keeps every existing assertion green (verified):

```js
.replace(/,(\s*,)+\s*/g, ', ')   // fires only on 2+, one match per run, idempotent
```

That yields `"a,,,b"` → `"a, b"` and `"— Привет, — — сказал."` → `"... Привет, сказал."`.

**3. The PR body claims a test that does not exist, and the test it points at proves the opposite of its own name — `server/src/tts/text-normalize.test.ts:137-141` (house gate: paired test must be red for the reason claimed)**

The PR body says it "added a test ensuring **intentional double commas in non-dash contexts are preserved**." No such test was added, and no such test could pass: `/,\s*,\s*/g` is an unanchored global replace with no dash awareness at all, so it collapses doubled commas in text containing no dash whatsoever — `softenDashes("a,, b")` → `"a, b"`.

The test actually added is `it('does not over-collapse a single comma adjacent to a dash run')` asserting `softenDashes('a, — b') === 'a, b'`. Its comment says "the collapse step must not fire on a single `,`" — but there is no single comma at collapse time. `DASH_RUN` rewrites `' — '` to `', '`, so the intermediate is `"a,, b"` and the collapse step **must** fire for this assertion to hold. Remove the fix and this test goes red (`"a,, b"`), which makes it a second positive test of the same collapse mechanism as `:133`, not the negative guard it is named and commented as.

Mutation-tested against all 13 `softenDashes` assertions plus the idempotency cases — the suite cannot distinguish the shipped regex from strictly broader or strictly narrower replacements:

| variant | suite result |
|---|---|
| shipped `/,\s*,\s*/g` → `', '` | all green |
| no collapse (pre-fix) | **2 fail** — `:133` and `:137`, both for the claimed reason ✅ |
| `/,+\s*/g` → `', '` (broader: eats any comma run) | **all green** — not killed |
| `/(,\s*)+/g` → `', '` (broader) | **all green** — not killed |
| `/,,/g` → `','` (narrower: no whitespace handling) | **all green** — not killed |

So the `\s*` handling is untested, and nothing pins the collapse's scope in either direction. The RED-before-fix claim in the PR body is genuine for the *collapse* behaviour; the scope claim is not backed by any assertion.

---

### 🟡 Minor

**4. `softenDashes` now rewrites commas in text containing no dash, and its doc comment was not updated — `server/src/tts/text-normalize.ts:106-126`**

The function's entire documented contract is about dashes ("Replace em-dash (U+2014) and en-dash (U+2013) with a comma…"); nothing in the JSDoc mentions comma-run normalization, so a reader of the doc would not predict `softenDashes("a,, b") === "a, b"`. Either scope the collapse to the dash pass — `DASH_RUN` as `/\s*,?\s*[–—]\s*/g` absorbs a preceding comma and keeps every existing assertion green (verified), leaving dash-free text untouched — or keep the global collapse and say so in the doc comment. CLAUDE.md counts a doc sentence the change made false as a chore fixed in the same round.

**5. Two tracked docs still record #2059 as deliberately deferred, made false by this PR**

- `docs/features/277-v115-bug-chore-sweep.md:827-829` — "Genuinely a question ('should it collapse?'), so it needs **a decision, not a fix**."
- `docs/features/277-v115-bug-chore-sweep-board.html:605-610` — `data-disp="design"`, note "**Defer.** Genuinely a question…"

Both need updating to record that the decision was taken and which way. Related: issue #2059's own **Expected** section reads "Unclear — that's the open question this issue tracks. Candidate: collapse `,,` to a single `,`." This PR picks that candidate and closes the issue. The outcome is defensible and I am not contesting it — but it is a design decision the repo's own sweep doc had routed to a design pass, so it should be ratified explicitly rather than landing implicitly via `Closes #2059`.

**6. Release-notes pair not updated and not marked not-applicable**

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched, and the PR body does not claim the exemption. This changes the actual wire text handed to the TTS engine on the dominant Russian dialogue shape, i.e. it changes generated audio for most dialogue lines in a Russian book — that is a user-visible delta, not a pure-internal chore. CLAUDE.md before-shipping step 5 requires both entries or an explicit "no shippable delta" statement.

**7. No on-box acceptance row, and no explicit not-applicable**

Register row **A32** (`docs/testing/onbox-acceptance-register.md:2488`) already exists for this exact function's Russian wire-text behaviour, specifically because it was "Pinned only as a wire-text transform (`text-normalize.test.ts`); never confirmed on real audio." This PR changes the same function's output on the same language's dominant dialogue shape, and #2059's own body argues the artifact "is a plausible contributor to the broader Russian degeneracy symptoms tracked under #2026." Whether collapsing actually improves the audio is precisely a hardware-provable claim. Either add a row (or extend A32) or state explicitly why none is owed — recording is the merge gate, running is not.

---

### ✅ What is solid

- The core mechanism is right for the shape #2059 filed: `"— Привет, — сказал Антон."` → `"... Привет, сказал Антон."`, and it generalises correctly to the multi-attribution form `"— Привет, — сказал он, — и ушёл."` → `"... Привет, сказал он, и ушёл."` and to the en-dash variant.
- The collapse never dirties text that was clean at base — it only fires where two commas were already adjacent, so there is no regression against `main` on any input I tried.
- It correctly survives the earlier passes: `stripUnsafeForTts` removes soft hyphens / zero-widths / control chars before the collapse runs, so `"a,­, b"`, `"a,​, b"` and `"a,, b"` all reach `"a, b"` through the composed pipeline. NBSP is handled by `\s*` directly.
- Replacing the old `#2059` pinning test rather than leaving it alongside is the right call, and the removed comment left no stale `#2059` citation behind in the test file.
- `Closes #2059` is present and outside any code span; PR title matches the commit convention; all 15 CI checks green.
- Correctly scoped diff — no drive-by edits, no orphaned imports, no derived artifacts staled.

---

### Verdict

**Changes requested.** Three correctness findings (🟠 1–3): the fix does not fire on tagged dialogue lines, it leaves a stranded comma on any 3+ run while breaking the file's stated idempotency contract, and the test advertised as pinning its scope pins the opposite — with mutation testing confirming the suite cannot distinguish the shipped regex from strictly broader or narrower ones. Findings 1 and 2 are both closed by moving to a single-bite collapse and either reordering `stripAudioTags` before `softenDashes` or folding the comma into `DASH_RUN`; finding 3 needs a real negative assertion (or an honest rename plus a genuine scope test). Findings 4–7 are chores this change made owed.

dudarenok-maker:
## PR review — pass 2 (head d3339f8187abe8b6353e890ff5d9b6c284aa97ec, depth medium)

Scope: re-review of pass 1's seven findings + the new merge commit `d3339f81` and its
manually-resolved on-box-register conflict. Verified: `vitest run src/tts/text-normalize.test.ts`
**50/50 pass**; `vitest run src/tts src/audio` **2162/2162 pass (145 files)**;
`npm run check:onbox-register` **OK**; all 16 PR checks green.

Depth re-derived from the current commit set (3× `fix(server)`, 1× `test(server)`,
3× `docs`, 1 merge — all single-scope) → **medium**, unchanged.

Beyond re-running the suite I rebuilt the pipeline from source in a scratch harness,
**proved the rebuild byte-identical to the real module across 16,807 fuzz inputs
(0 mismatches)**, then flipped one line per mutant to test whether each fix's tests are
red *for the reason claimed*.

---

**1. Audio-tag ordering defeats the collapse — ✅ VERIFIED RESOLVED (but see NEW-1).**
`normaliseForTts` is now `softenDashes(stripAudioTags(denormaliseAllCaps(stripUnsafeForTts(text))))`
(`server/src/tts/text-normalize.ts:181`). Mutation: restoring the old order
(`stripAudioTags` outermost) reds **exactly** the three audio-tag tests and nothing else —
`"Привет, [shouting] — сказал."` → `"Привет, , сказал."` under the mutant, `"Привет, сказал."`
under the fix. Real regression test, red for the claimed reason.

**2. 3+-comma runs only partially collapse — ✅ VERIFIED RESOLVED.**
Regex is `/,(\s*,)+\s*/g` (`text-normalize.ts:134`). Mutation: restoring `/,\s*,\s*/g` reds
**exactly** the three comma-run tests — `a,,,b` → `"a, ,b"`, `a,,,,b` → `"a, , b"`,
`— Привет, — — сказал.` → `"... Привет, , сказал."`. Independently fuzzed: `softenDashes`
is idempotent on **all 6,561** probe inputs (0 failures), and runs of 2–5 commas all collapse
in one pass.

**3. Mislabeled test — ✅ VERIFIED RESOLVED in the test file; the identical claim survives in the PR body.** See NEW-3.
`text-normalize.test.ts:137-153` now correctly labels the positive collapse test and adds a
test documenting the regex has no dash-awareness. Both assertions match real behaviour.

**4. JSDoc comma-collapse scope — ✅ VERIFIED RESOLVED.** `text-normalize.ts:126-129` is accurate,
including the "fires independent of dashes" caveat.

**5. Stale planning docs — ✅ VERIFIED RESOLVED.** Both `docs/features/277-v115-bug-chore-sweep.md`
and its `-board.html` mirror updated, *including* the `g7` note dropping `#2059` from the
defer list — the easy half to miss.

**6. On-box register row — ✅ VERIFIED RESOLVED as to the row and its arithmetic** (residual bookkeeping: NEW-2, NEW-4).
I re-derived the merge resolution independently rather than trusting the checker: Group A
headings are **exactly 38, A1–A38, contiguous, zero duplicates, zero gaps**;
38+2+4+3+10+2+2 = **61**, matching the stated total; the live-view mirrors it exactly
(strip 60→61, glance 37→38, group header 37→38, A38 appended last in Group A) with
Blocked/Unconfirmed/Groups correctly untouched. **No arithmetic drift and no wrong row ID
from the conflict resolution.**

**7. Release-notes pair — ✅ VERIFIED RESOLVED.** Both entries present, correctly placed in the
in-progress section, non-duplicating after the merge.

**Merge commit `d3339f81` — clean.** `git diff 677d42e6 d3339f81 -- server/src/tts/text-normalize.ts
server/src/tts/text-normalize.test.ts` yields **only** the expected JSDoc hunk from `421de069`.
No earlier fix was silently reverted. The golden-assembly fixture (opt-in suite, not in `verify`,
so a break there would be silent) is unaffected — all 3 segments still normalise to identity,
0 changed vs `main`.

---

### 🟠 Significant — NEW-1: the reorder that fixed finding 1 silently dropped `normaliseForTts`'s whitespace tail, breaking composed idempotency — the very contract this PR's release note claims to have restored

`server/src/tts/text-normalize.ts:181`

Before this PR, `stripAudioTags` was **outermost**, so its
`.replace(/\s+/g,' ').trim()` (`text-normalize.ts:162`) ran **last** and cleaned up whatever
`softenDashes` emitted. Moving it inside makes `softenDashes` the final pass — and both its
`DASH_RUN → ', '` and the new comma-collapse `→ ', '` can leave a trailing space that nothing
now removes.

Measured across 16,807 fuzz inputs, PR head vs `main`:

| | `main` | PR head |
|---|---|---|
| outputs with leading/trailing whitespace | **0** | **4585** |
| non-idempotent `normaliseForTts` | **0** | **4585** |

Concrete repro:

```
normaliseForTts('He paused—')   main: "He paused,"    PR: "He paused, "
normaliseForTts('Hello,,')      main: "Hello,"        PR: "Hello, "
normaliseForTts('THE END—')     main: "The End,"      PR: "The End, "
```

and therefore `normaliseForTts(normaliseForTts(x)) !== normaliseForTts(x)` for ~27% of inputs.

Three things make this more than cosmetic:

1. **The file's own header states the invariant** — `text-normalize.ts:33`: *"All transforms are
   idempotent and order-independent."* That is now false for the composed entry point.
2. **`docs/release-notes-next.md` explicitly claims this PR *fixed* the idempotency contract**
   ("previously needing a second pass to fully collapse, which broke the module's stated
   idempotency contract"). The PR restores it at the `softenDashes` level while breaking it one
   level up, so the shipped note asserts the opposite of the shipped behaviour.
3. **The suite's own composed-idempotency test cannot catch it.** `text-normalize.test.ts:290`
   uses `'THE HAZE—then.'` — dash *interior*, so it passes under every variant I tested. My
   mutation harness confirms the discrimination directly: a **trailing**-dash vector
   (`'He paused—'`) is `IDEMPOTENT` under the old pass order and `NOT` under the PR's. The guard
   is blind to exactly the case the fix broke.

Blast radius traced, and it is genuinely narrow — I want to be calibrated rather than alarming:
the ASR QA gate is immune (`segment-asr-qa.ts:340` tokenises with `.split(/\s+/).filter(Boolean)`),
the two empty-sentence filters are immune (`synthesise-chapter.ts:921`,
`build-synth-replacement.ts:237` both `.trim()`), and a trailing space is almost certainly
inaudible. What actually bites: the token-budget packer's `textLen`
(`synthesise-chapter.ts:2743`) is +1 per affected group, the wire text handed to the sidecar now
carries trailing whitespace, and a documented invariant is false with a vacuous guard over it.

Fix is one line — re-establish the whitespace tail in `normaliseForTts` (e.g. `.trim()`, or move
the `\s+` collapse to the end) — plus an idempotency vector with a **trailing** dash so the guard
stops being blind. Worth noting for triage: this is the third round in a row where the fix for
round N broke something round N's own test could not see.

---

### 🟡 Minor — NEW-2: the register's and live view's provenance lines now contradict their own totals

- `docs/testing/onbox-acceptance-register.md:356` still reads
  `Last change: 2026-08-27 (on-box wave 8), 61 → 60.` while **line 352 now reads `61 owed.`**
  The register asserts 61 owed and, four lines later, that the last change landed it at 60.
  Nothing records this PR's 60 → 61. On `main` these two lines agreed (60 owed / 61 → 60), so the
  contradiction is introduced here.
- `docs/testing/onbox-acceptance-register-live-view.html:1490` still reads
  `Last synced 2026-08-27 against branch chore/ops-onbox-wave8` — now false, since the file also
  carries A38 from `fix/server-softendashes-doubled-comma`.
- `...live-view.html:201` reads `Last change: 2026-08-26 (PR #2665 …), 69 → 69` against a strip
  now showing 61. **This one is pre-existing on `main`** (wave 8 updated the `.md`'s line but not
  the HTML's) — flagging it because "it's pre-existing" is a void deferral reason here, not to
  attribute it to this PR.

The convention is established by the directly comparable add-one-row commit `d5d6268f` (A46 for
`#2656`), which updated **both** `Last change` lines to `<prev> → <new>`. Worth naming: `npm run
check:onbox-register` passes and *cannot* catch this — it cross-checks the owed total, per-group
counts and row IDs between the two files, and is blind to these narrative lines.

---

### 🟡 Minor — NEW-3: the PR body now contradicts the code in two places, and omits six of the eight changed files

The body still describes the original commit only:

- it quotes `.replace(/,\s*,\s*/g, ', ')` — **the retired regex**; the code is `/,(\s*,)+\s*/g`;
- it claims the tests "ensur[e] intentional double commas in non-dash contexts are **preserved**"
  — this is the exact inverted claim pass-1 finding 3 corrected *inside the test file*, left
  standing here. The code collapses them (`softenDashes('a,,b') === 'a, b'`); it does not preserve
  them. One instance fixed, the class left armed.
- `## Changes` lists 2 files; the PR touches 8. `RELEASE_NOTES.md`, `docs/release-notes-next.md`,
  both register surfaces and both planning docs are undeclared, as are the pass reorder and the
  3+-comma fix.

CLAUDE.md requires incidental fixes be declared in the PR body ("Also fixed, found in passing: …")
precisely so they don't read as scope creep — and this repo merges with the PR body, so the stale
text outlives the PR.

---

### 🟡 Minor — NEW-4: A38's criteria pointer promises shapes its target does not contain

`docs/testing/onbox-acceptance-register.md:2928` — *"Criteria: see #2059 for the dialogue shapes
to test."* Issue #2059 contains **one** shape (`— Привет, — сказал Антон.`) and states under
**Expected**: *"Unclear — that's the open question this issue tracks"* — the question this PR has
now answered. So an operator picking up A38 is sent to a document that has neither the plural
"shapes" nor a settled expectation.

Pointing at an issue is itself an accepted convention here (A46 does exactly that), so the fault
is narrow: the target doesn't hold what the row claims. Note also that the sibling row **A32**
(`:2488`) — same file, same function, same Russian-dialogue "by ear" shape — routes its criteria
to `docs/testing/fs38-wave3-onbox-acceptance.md`, which exists and already carries `#2026`'s
"additional acceptance criteria" section. Either fold #2059's shapes in there as A32 does (the
third surface CLAUDE.md's step 3 asks for "where one exists"), or reword the pointer so it doesn't
promise what isn't there.

---

### Verdict

The three correctness findings from pass 1 are genuinely fixed, and I confirmed each by mutation
rather than by re-reading the diff — both fixes' tests are red for the reason claimed, and the
manually-resolved register conflict introduced **no** arithmetic drift or row-ID collision, which
was the main risk in this round.

One new 🟠 correctness regression, introduced by finding 1's own fix and invisible to the suite's
idempotency guard: `normaliseForTts` lost its whitespace tail. It is low-consequence downstream —
I traced every consumer and they are whitespace-immune — but it falsifies the module's stated
contract and the shipped release note, behind a test that cannot fail. Recommend fixing it plus
the blind vector before merge. The three 🟡 items are bookkeeping the merge left behind.

**Not ready to merge** pending NEW-1.

dudarenok-maker:
## PR review — pass 3 (head 710ad0b002c7bfa9a59c7856781fdafb63085f4e, depth medium)

Scope: re-review of pass 2's four findings + the two further `origin/main` merges (`c44ef612`'s
chain, then `710ad0b0`) and the second and third manual re-resolution of the on-box register.
Verified: `vitest run src/tts/text-normalize.test.ts` **51/51**; `vitest run src/tts src/audio`
**2163/2163 (145 files)** — exactly +1 vs pass 2, the new trailing-dash test and nothing else;
`npm run typecheck` clean; `npm run check:onbox-register` **OK**; `npm run check:register-citations`
**OK**; all 16 PR checks green.

Depth re-derived from the current commit set (3× `fix(server)`, 1× `test(server)`, 4× `docs`,
plus merges — all single-scope) → **medium**, unchanged.

Beyond re-running the suites I rebuilt the pre-fix module from `git show` and diffed it against
the head module across a 48-input corpus × 7 language codes, and ran a mutation harness that
harvests every string literal in the test file (170 distinct) and asks, per mutant, whether *any*
input the suite ever feeds can distinguish it from shipped.

---

**NEW-1 (🟠, composed idempotency lost) — ✅ VERIFIED RESOLVED.**
`normaliseForTts` (`server/src/tts/text-normalize.ts:186-190`) re-runs `.replace(/\s+/g,' ').trim()`
after `softenDashes`. Verified three ways, not by re-reading the diff:

- Direct repro against the pre-fix module (`d3339f81`) and head:
  `normaliseForTts('He paused—')` → **pre-fix `"He paused, "`, head `"He paused,"`**; second
  application a no-op at head. The new test is red before the fix, green after, and red **for the
  reason claimed** (trailing whitespace, not a different output) — while the same pre-fix module
  is already *correct* on `— Привет, — сказал Антон.`, `Привет, [shouting] — сказал.`, `a,,,b` and
  `She said [emphatic] hello.`, so the redness is specific to the fixed defect.
- Mutant "drop the final collapse-and-trim" is killed by exactly two of the test file's literals,
  one of which is the new test's `'He paused—'`. The guard is no longer blind.
- Idempotency swept, not asserted: over the 48-input corpus, **every** exported transform and
  `normaliseForTts` at 1-arg and at `en/ru/fr/es/de/zz` is idempotent — 0 failures. Trailing dash,
  leading dash, interior dash, bare dash, audio tags, comma runs of 1–4, `,,`/`,,,` alone,
  newlines/tabs, NBSP and narrow-NBSP French typography, and no-dash-at-all all included.

**No collateral from the added trim.** Old-vs-new across 48 inputs × 7 lang codes: every single
difference is an intended improvement (doubled-comma collapse, tag-crossing collapse, and
`[emphatic] — Кто это?` improving from `", Кто это?"` to `"... Кто это?"`). Prose with no
dash/comma pathology is byte-identical. **`expandForSpeech` is unaffected** — `base` was already
collapsed-and-trimmed before expansion pre-PR, because `stripAudioTags` ran last and carries the
same tail; the ordering vs `expandForSpeech` is unchanged, and the `$1,250.50`/`3/4/2026`/
`Dr. Smith`/`50% of $20` vectors are identical old vs new at every language code, including the
trailing-dash-plus-expansion shapes `'It cost $5—'` and `'On 3/4/2026—'`.

**NEW-2 (🟡, stale register/live-view narrative) — ⚠️ PARTIALLY RESOLVED.** Its first two bullets
are fixed (`register.md:356` now records `59 → 60 via PR #2688`; the live view's footer now names
this branch). Its **third bullet is still open and this PR re-staled it** — see 🟡-2.

**NEW-3 (🟡, PR body contradicts the code) — ✅ RESOLVED as to the retired regex and the phantom
"preserved" test**, and the body now declares all eight files and all three rounds' fixes. The
renumbering since then re-staled one figure — see 🟡-3.

**NEW-4 (🟡, A38's criteria pointer) — ✅ RESOLVED**; the pointer now reads "the bullet above". The
replacement sentence introduced the citation rot in 🟡-1.

**The renumbering itself — clean.** `git diff origin/main 710ad0b0` on both register surfaces is
**exactly** four hunks, all this PR's own: glance `36 → 37`, total `59 → 60`, the Last/Previous
change block, and the A37 row (mirrored by strip `59 → 60`, glance `36 → 37`, group header
`36 → 37`, the A37 `<details>`, footer). Nothing else drifted. Group A headings are
**A1–A37, contiguous, zero duplicates, zero gaps in *both* files**; 37+2+4+3+10+2+2 = **60**,
matching the stated total; Blocked/Unconfirmed/Groups/oldest-debt correctly untouched. A37 is not a
duplicate of any row.

**The two merges carried no silent regression.** `git diff 677d42e6 710ad0b0 --
server/src/tts/text-normalize.ts server/src/tts/text-normalize.test.ts` is exactly two hunks: the
`softenDashes` JSDoc paragraph from `421de069` and `c44ef612`'s trim + its test. Nothing reverted,
nothing extra.

---

### 🟡 Minor — 1: A37's body cites **A32** twice; after the renumbering A32 is a different row, and the file contradicts itself about it

`docs/testing/onbox-acceptance-register.md:2925` — *"same open shape as **A32**'s
leading-dash-to-ellipsis case"* — and `:2936` — *"no separate run sheet (unlike **A32**'s)"*.
Mirrored at `docs/testing/onbox-acceptance-register-live-view.html:849` and `:853`.

At pass 2's head the leading-dash row **was** A32. Wave 9/9b/9c renumbered Group A twice more; the
row's own number was moved A38 → A37 in this PR's merge resolution, but these two references were
not:

- `register.md:2527` — **A31** · *"Russian XTTS quality — leading-dash pause by ear … (#2026, PR
  #2050)"*, and it is the row that has the separate run sheet
  (`docs/testing/fs38-wave3-onbox-acceptance.md`) A37's sentence contrasts itself against.
- `register.md:2581` — **A32** is now *"Named-entity decode reaches the TTS engine on a real EPUB
  (#2310)"*. An operator following A37's pointer lands on an unrelated EPUB row.

The live view proves the intended referent internally: `live-view.html:324` cites the same #2026
subject as **A31** (correctly renumbered by wave 9), while `:849`, added here, says A32. One file,
two IDs for one row.

Worth naming because it explains why CI is green: `check:register-citations`' own header states its
fatal Check C only fires on an anchored heading or a `Criteria source:` line carrying a subject
number, and Check A only fires on a *nonexistent* ID. This is a prose-idiom citation of an ID that
still exists for another subject — the one shape both fatal halves are documented blind to.

### 🟡 Minor — 2: the live view's `Last change` callout still reads wave 9c while its own strip reads 60

`docs/testing/onbox-acceptance-register-live-view.html:201` — *"Last change: 2026-08-27 (on-box
wave 9c, branch `chore/ops-onbox-wave9`), **58 → 59**."* — against `:192`'s strip now showing
**60 Owed**, and `register.md:356`'s *"Last change: 2026-08-27, **59 → 60** via PR #2688"*.

This is pass-2 NEW-2's third bullet. It was flagged then as pre-existing on `main`; it is not
pre-existing now — wave 9 refreshed it to `58 → 59`, and **this PR's own 59 → 60 staled it again**
while updating the `.md`'s equivalent line and the HTML's footer. The `.md` and the live view now
disagree about what the last change was. `check:onbox-register` passes and cannot see it: the
register's own "Live view" section says this half of the summary strip is a manual recompute.

### 🟡 Minor — 3: the PR body still says **row A38**, twice

*"the on-box acceptance register (new row A38)"* and *"`docs/testing/onbox-acceptance-register.md`
(+ its live-view mirror): new row **A38**"*. The row is **A37**. This repo merges with the PR body,
so the wrong ID outlives the PR — the same durability argument NEW-3 was fixed on.

### 🟡 Minor — 4: the sweep doc and its board mirror say "(50 unit tests)"; the suite is 51

`docs/features/277-v115-bug-chore-sweep.md:831` and
`docs/features/277-v115-bug-chore-sweep-board.html:608`. `vitest run src/tts/text-normalize.test.ts`
reports **51 passed** — `c44ef612` added the 51st after those two lines were written. A figure this
PR's own later commit staled.

### 🟡 Minor — 5: both release-notes entries describe a mechanism that does not exist, and each contradicts the bullet directly above it

- `RELEASE_NOTES.md:93` — *"Castwright's text preparation was turning **both dashes into commas**
  … landing them back-to-back in the audio."*
- `docs/release-notes-next.md:226` — *"where **both the line-opening dash and the mid-sentence
  dash converted to commas**, landing them back-to-back in the wire text."*

Measured on the pre-PR module: `normaliseForTts('— Привет, — сказал Антон.')` →
**`"... Привет,, сказал Антон."`**. The leading dash became `... `, not a comma — that is #2026's
shipped behaviour, and it is what the immediately preceding bullet in *each* of these two files
says (`RELEASE_NOTES.md:92`, `release-notes-next.md:225`: *"A leading dash now becomes `... `
instead"*). The doubled comma is the author's own comma plus the **interior** dash's comma. The
technical entry's own example even drops the `... ` that disproves its sentence.

Secondary, same sentence: `RELEASE_NOTES.md:93`'s *"the same way it **already** does when
Castwright encounters one anywhere else"* asserts prior behaviour that did not exist — nothing
collapsed doubled commas anywhere before this PR; the everywhere-collapse is what this PR
introduces.

### 🟡 Minor — 6: `text-normalize.ts:33`'s "order-independent" half is still false, and now the file's own docstring says so

`server/src/tts/text-normalize.ts:33` — *"All transforms are idempotent and order-independent."*
Pass 2 flagged this line under NEW-1; the fix restored the **idempotent** half and left the line
untouched, so the **order-independent** half stands. `normaliseForTts`'s own docstring 135 lines
below (`:168-172`) now states the opposite as load-bearing: *"Order matters: … strip audio tags
BEFORE softening dashes (so the doubled-comma collapse regex can match even when a tag sits
between the commas)"*.

Measured: swapping just those two passes changes the result on **17 of 48** corpus inputs, and
semantically — not merely in whitespace — on the tag shapes:

```
stripAudioTags(softenDashes('Привет, [shouting] — сказал.')) -> "Привет, , сказал."
softenDashes(stripAudioTags('Привет, [shouting] — сказал.')) -> "Привет, сказал."
```

Concrete failure scenario: a maintainer reading `:33` restores the pre-PR composition order —
which the file described as "academic" until this PR — and silently reintroduces pass-1 finding 1.

### 🟡 Minor — 7: the suite still cannot kill a strictly *broader* collapse regex — pass-1 finding 3's second half

Pass 1 measured that no assertion pins the collapse's scope in either direction and asked for
"a real negative assertion (or an honest rename plus a genuine scope test)". The rename landed and
a *positive* no-dash-awareness test was added (`text-normalize.test.ts:155-165`); the negative
scope guard was not.

Harvesting all **170** distinct string literals in the test file and running each under all five
exported functions plus `normaliseForTts(x, 'en')`:

| mutant | inputs in the test file that distinguish it |
|---|---|
| `/,+\s*/g` → `', '` (broader) | 9 — killed |
| `/,,/g` → `','` (narrower) | 18 — killed |
| `/,\s*,\s*/g` → `', '` (pass-1 regex) | 9 — killed |
| no collapse at all | 10 — killed |
| old pass order | 10 — killed |
| drop the final trim | 2 — killed (incl. the new test) |
| **`/(,\s*)+/g` → `', '` (broader)** | **1 — and it is never asserted → survives** |

The one distinguishing input is `softenDashes('He paused,')` (shipped `"He paused,"`, mutant
`"He paused, "`). `'He paused,'` appears in the file only as the *expected value* of
`normaliseForTts('He paused—')` and inside `expect(normaliseForTts(once)).toBe(once)` — and
`normaliseForTts` masks the difference, because the new final `.trim()`/collapse normalises it
away. So the suite would stay fully green against a regex that also rewrites every *single*
comma: `softenDashes('Мой дом,мой замок')` → `'Мой дом, мой замок'` under the mutant,
`'Мой дом,мой замок'` shipped.

The shipped regex is correct — this is a guard gap, not a behaviour bug. One assertion closes it,
e.g. `expect(softenDashes('a,b')).toBe('a,b')` in the same block.

---

### Explicit non-findings, so they are not re-raised

- `softenDashes('a ,, b')` → `'a , b'` — the collapse does not consume whitespace *before* the run.
  Pre-PR output was `'a ,, b'`, so this is strictly an improvement on already-malformed input, and
  a space before a comma has no audible effect. Below the bar.
- `register.md:3586`'s `A32/A41` reference sits inside a dated 2026-08-20 correction entry —
  frozen history, correctly left alone.
- `Previous change:` as a third label alongside the file's existing `Last change:`/`Prior change:`
  chain is a style choice, not a defect. Flagging it would be taste.

---

### Verdict

**NEW-1 is genuinely fixed and its regression test is real** — red before, green after, red for the
claimed reason, and it kills the mutant that removes the fix. Composed idempotency now holds across
every shape I could construct, at every language code, and the added trim changes nothing that was
not already intended: `expandForSpeech` is byte-identical, and no un-pathological prose moved. The
renumbering to A37 is arithmetically and structurally correct on both surfaces, and the two large
`main` merges reverted nothing and smuggled nothing in.

**No 🔴 and no 🟠 this round** — nothing here re-triggers the loop, and nothing found is a
correctness bug in shipped behaviour.

Seven 🟡 remain, and six of them are one failure mode: **this round's own renumbering and fixes
left their citations behind.** A37 points at the wrong row (twice, on two surfaces), the live
view's change callout points at the wrong wave, the PR body points at the wrong row ID, the sweep
docs point at the wrong test count, and both release notes describe a mechanism the code has not
had since #2026. The seventh is the scope guard pass 1 asked for and pass 2 recorded as resolved.
Per the triage rules these are fixed in this round but do not reopen the review loop.

PR #2402, opened 2026-08-15
fix(server): rebuild the audition centroid when a character's voice is reassigned (#1969)

dudarenok-maker:
Closes #1969

## Summary
When a character's voice is reassigned, the render-integrity gate kept scoring the new voice against the old voice's persisted audition centroid — every line of the new voice was measured as drift from a speaker it was never supposed to match, producing false `voice-mismatch / severity: severe` flags (and a re-render every pass, since the auto-fix retried against a reference that could never match).

The persisted Option-B audition centroid is now self-describing. It records the voice it was built from (`{voiceName, modelKey, language, cloned}`), is reused only while that still matches the character's current voice, and is otherwise discarded and rebuilt. A row whose character still has a resolved voice is rebuilt once (pre-existing files included); a character whose current voice is no longer resolvable falls inconclusive rather than trusted.

## Implementation (branch `fix/server-1969-audition-reference-voice`)
- `dbca754c` — persist the audition centroid voice identity (`AuditionVoiceRef {voiceName, modelKey, language?, cloned}` + optional `CharacterCentroid.auditionVoice`)
- `5e43be39` — gate audition-centroid reuse on the current voice in `resolveCharacterReference` (`matchesCurrentVoice`; reuse only on a match, otherwise discard + rebuild; voice-info-absent degrades to `too-short`)
- `8615247b` — regression suite (Tests 1-4)
- `e099f51f` — implementation plan + release notes

## Verification
- Regression suite is **mutation-proven**: reverting the gate to the old unconditional reuse fails all four tests; restoring passes them.
- render-integrity: **12 files / 100 tests** green. `tsc`, `eslint`, `config:check`, and `npm run verify:fast:branch` (pre-push gate) all green.
- Whole-branch review: **MERGE-READY**.

## Ship-time obligations for this PR
- **On-box acceptance row:** this ships hardware-provable behavior — record the row in `docs/testing/onbox-acceptance-register.md` (+ republish its live view) before merge; this unblocks the existing #1969-blocked A24 `voice-mismatch` sub-check.
- **Identity-proxy confirmation:** the maintainer open confirmation re Coqui same-bucket designed-voice names remains (plan v2, "Open confirmation needed") — a voice reassignment there may not change the resolved name, so it is documented as a residual, not silently claimed.
- **Release notes:** added to `docs/release-notes-next.md` and the in-progress `RELEASE_NOTES.md`.

PR #2698, opened 2026-08-27
chore(docs): correct on-box register discharge evidence + fix citation rot

### LINKED ISSUE #1969: Reassigning a character's voice keeps scoring it against the old voice's persisted audition centroid

dudarenok-maker:
## Summary
- On-box acceptance wave 9: **A34** (voice reassignment vs. persisted audition centroid, #1969/PR #2402) had two criteria — discard the stale reference (proven: reassigned Ivo's voice on the real *The Coalfall Commission* book, `matchesCurrentVoice()` correctly returned false, zero `voice-mismatch` flags), and a rebuilt reference actually scoring real audio (not proven — the run hit `too-short`, the row's own named failure outcome, because Ivo's cloned-voice sample was too short for `auditionCentroid()`). The proven half stays discharged; the unproven half is split into **new row A36**, so the owed total returns to **59** rather than silently dropping. Full acceptance text tracked at #2700.
- **A8** (GPU residency safety + coexistence, plan 222) discharged per the repo owner's direct confirmation, recorded in all three policy-mandated homes: plan 222's Ship notes (flips `status: stable`, archived to `docs/features/archive/`), the run sheet's five Result lines, and the register's own changelog.
- Fixed citation rot across four commits, driven by three rounds of PR review: the original renumbering sweep, a follow-up closing the sweep's own gaps (plus a regression it introduced), a third round splitting A34's unmet criterion into A36 and closing ~10 more sites of the same rot class, and a fourth honoring `release-notes-next.md`'s FROZEN_EXACT convention (annotate, never silently renumber a diary entry) after the third round's fix violated it there.

## Test plan
- [x] `npm run check:onbox-register` — .md and live-view HTML agree (59 owed, Group A = 36 rows)
- [x] `node scripts/check-register-citations.mjs` — clean at every commit
- [x] `npm run typecheck`, `npm run test:server` (8034 tests), `npm run test:hooks`, `npx eslint` on every touched code/test file — all pass
- [x] `npm run verify:fast:branch` green pre-push, all four commits
- [x] `npm run check:onbox-register -- --against-published --discharging A37` run against a freshly-fetched copy of the live artifact before republishing — clean
- [x] Live-view artifact republished cleanly (no force needed) after adding A36

**Release notes:** skipped — no shippable delta (register/docs bookkeeping; the code/test comment fixes change no logic). The `release-notes-next.md` edits in this PR are annotation-only corrections to its own frozen historical entries, not new release content.

**Review history:** 3 rounds (the `pr-review-gate` skill's cap). Pass 1: 2 Blocking + 2 Significant. Pass 2: both Blocking still open (narrowing wasn't enough; A8 fix introduced 2 new wrong-row citations + left 3 "still owed" contradictions). Pass 3: **both Blocking VERIFIED RESOLVED** (A34 correctly split into discharged-proven + new-owed-A36; A8's evidence trail internally consistent); found 4 new issues from pass-2's own fixes (a frozen-diary violation, 2 mislabeled `#2107` citations, 1 self-inconsistent row number) — all fixed in the 4th commit above. Loop cap now reached; no further automatic re-review.

Closes #1969
Refs #2402, #2435, #2700


dudarenok-maker:
## PR review — pass 1 (head 31234867bf339456ee4fcca360969a78995502c2, depth low)

Scope: all 13 changed files (`docs/testing/onbox-acceptance-register.md`, the paired live-view HTML, 9 run-sheet/sitting docs, `server/tts-sidecar/main.py`, `server/tts-sidecar/tests/test_ort_cuda_preload.py`) · verified before probing: `node scripts/check-onbox-register.mjs` → OK (`.md`/HTML agree), `node scripts/check-register-citations.mjs` → OK (Checks A/B/C-fatal clean), Group A row headings enumerated (`A1`…`A35`, contiguous, no gaps, no duplicates), glance-table arithmetic recomputed by hand (35+2+4+3+10+2+2 = **58** ✓), oldest-debt claim re-derived (A11 = plan 161 A/B audition, first landed 2026-06-01 ✓, and the live view's `06-01` tile + `A11 · Oldest debt` badge agree). Then: every Group-A `row A<n>` citation in the tree resolved against **both** `origin/main`'s register headings and this head's, to separate pre-existing rot from rot this PR created.

Not a docs-only PR — `server/tts-sidecar/**` is outside the `docs/**` / root `*.md` / `.github/*.md` file-set test — so the gate applies. Depth `low` accepted (`docs(docs):`, single scope).

---

### 🔴 Blocking — the A34 discharge rests on an observation that cannot fail, and it permanently consumes the only known test subject

`docs/testing/onbox-acceptance-register.md:361-378` records old A34 (#1969 / PR #2402) as discharged on: *"All 8 of Ivo's lines came back `qa.status: 'ok'`, `reasons: []` — **zero `voice-mismatch` flags**, the row's stated acceptance criterion"*, plus *"after, that entry reads `referenceKind: \"too-short\"`"*.

The row had **two** criteria, not one. Its own deleted text (visible in the diff) reads:

> The new voice's lines must **not** be flagged `voice-mismatch`/severe — the persisted centroid must be **rebuilt for the new voice**, not reused against the old speaker's.

Criterion 2 was not met, and criterion 1 is vacuous once criterion 2 fails. In `server/src/audio/render-integrity/aggregate.ts`:

- `:247` — `if (persisted?.referenceKind === 'too-short') { return { status: 'too-short', ... } }`. Once an entry is `too-short`, resolution short-circuits **before** the reassignment gate at `:250-255` ever runs.
- `:262-263` — `if (!voiceInfo) return { status: 'too-short', ref: TOO_SHORT_REF }` returns `too-short` **without attempting a rebuild at all**.
- `server/src/audio/render-integrity/aggregate-audition-voice-reassign.test.ts:1-23` states this in the fix's own words: *"a persisted audition row is NOT trusted, degrades to too-short, and that state is ABSORBING across passes (no later attempt re-renders)"*.

Concrete consequence, three parts:

1. **"Zero `voice-mismatch` flags" is guaranteed by construction here.** With no reference (`too-short`), the gate has nothing to score against, so it cannot emit `voice-mismatch` whether PR #2402's `matchesCurrentVoice` gate works, is broken, or is deleted. The observation does not discriminate between "the fix works" and "the fix is absent". (Reviewer brief shape 5: an acceptance criterion blind to its own feature — it passes on a null observation.)
2. **The recorded evidence cannot tell which branch fired.** `audition → too-short` is produced by `:262` (`voiceInfo` undefined — never reaches the reassignment gate) *and* by `:273` (`auditionCentroid` returned `too-short`). Only the second is the path the row exists to prove. The write-up guesses the second (*"a cloned voice apparently doesn't get a synthesized `audition` reference"*) and nothing in the record rules out the first — which is precisely the failure shape waves 7 and 8 already hit twice with this row.
3. **Wave 8's own stated discriminator was dropped, not met.** The wave-8 note deleted by this diff (the removed A34 body) required confirming *"`cleanMean`/`pSevere`/`pBand` **actually change** from their pre-reassignment values (a genuine rebuild, not a stale carry-over)"*. They did not change to new values — they went away. That is neither the pass condition nor the named failure mode; the write-up silently substitutes a weaker property ("nothing stale survives") for the one the row asked for.

And the cost is not recoverable: because `too-short` is absorbing (`:247`), Ivo can never be scored on this book again, so a re-run of this acceptance on the one character wave 8 spent a whole session identifying will now short-circuit at `:247` before reaching the gate under test. The row is being removed at the same moment its subject was rendered untestable.

What would actually discharge it: a character whose post-reassignment entry comes back `referenceKind: 'audition'` with an `auditionVoice` naming the **new** voice, and `cleanMean`/`pSevere`/`pBand` different from the pre-reassignment values. If that cannot be produced on a cloned voice, the honest outcome is the row staying owed with the `too-short` finding recorded on it — not a discharge.

---

### 🔴 Blocking — A8 is removed with no recorded observation, and now three tracked documents contradict each other

`docs/testing/onbox-acceptance-register.md:356-360` is the entire A8 discharge: *"**old A8** (GPU residency safety + coexistence, plan 222) — owner-confirmed as proven, per this file's 'record what was observed, by whom, and when' policy."*

That cites the policy without satisfying it. The policy is in this same file at `:299-306`:

> - Either way, record *what was observed*, by whom, and when — **in the plan's Ship notes, the run sheet, or the issue**. "Tests pass, so it's presumably fine" is never a reason to remove a row.

All three named homes are empty or contradict the discharge:

- **The plan** — `docs/features/222-gpu-residency-and-analysing-honesty.md` is untouched by this PR and still reads `status: active` (`:2`), *"**on-box GPU acceptance owed** (real 8 GB eviction + 409 refusal + 12/16 GB coexistence)"* (`:9`), *"Manual acceptance walkthrough (USER-RUN, live GPU — OWED)"* (`:54`), and *"**Remaining before `stable`:** the live-GPU acceptance walkthrough above (5 steps), then flip `status: stable` + archive"* (`:77`). No Ship notes filled, no `git mv` to `archive/` (before-shipping step 8).
- **The run sheet** — `docs/testing/onbox-sitting-two-card-boot.md` is edited by this PR, and the edit is annotation-only: all five A8 `- Result:` lines (`:58`, `:60`, `:62`, `:64`, `:66`) are still blank. Worse, the same file's own re-resolution section at `:122` still reads *"**A8** — plan 222 frontmatter `status: active` … **STILL OWED**"* — a direct self-contradiction with the header this PR added six lines from the top of the same file. `:19`'s running-time budget also still bills A8 (and cites `A18`, which is not one of this pack's rows in either numbering).
- **The issue** — none referenced. The PR body's `Refs #1969, #2402` covers the A34 half only.

The old A8 row had **five** steps, and step 5 (*"**no** eviction on a 12/16 GB box"*) explicitly needs the roomier card. Nothing anywhere records which of the five were observed, on what hardware, or on what date. Under `:299-306` that is a deletion, not an outcome — the distinction the register exists to hold. Compare the wave-7 owner-confirmed discharges at `:418-430`, which the same author cites as precedent: each of those five names a concrete observation ("the amber 'Suspect' badge on a deliberately degraded (near-silent/clipped/truncated) render, confirmed on both the Generate and Listen rows"). A8 gets none.

---

### 🟠 Significant — this PR introduces new citation rot in the very directory it swept

`docs/testing/onbox-sitting-fs38-wave3.md:46`:

```
| Issue #2026 (Russian XTTS quality, register row A32 — not this row) | `OPEN` | ... |
```

- On `origin/main`, `### A32` **is** *"Russian XTTS quality — leading-dash pause by ear … (#2026, PR #2050)"* (`register.md:2488`). The citation was correct.
- At this head, the Russian XTTS row is **A31** (`:2497`) and `A32` is *"Named-entity decode reaches the TTS engine on a real EPUB"* (`:2551`). The citation is now wrong, and wrong in the dangerous way — it still points at a real row, just a different subject.

This file is `docs/testing/onbox-sitting-*.md`, the same family as the five sitting packs the PR did update (cloning-identity, multilanguage, two-card-boot, voice-design, vram-contention). It was missed.

Neither checker catches it, and that is structural rather than bad luck: `check-register-citations.mjs`'s Check A only fires on a *nonexistent* ID (A32 exists), and Check C's fatal `wrongId` half reads only anchored `### <ID> ·` headings and `Criteria source:` lines — this is a table cell in prose, on neither surface. The script's own success line says so. A green run here is not evidence the sweep was complete. (Reviewer brief shape 2: a guard that enumerates syntax loses one spelling per round.)

---

### 🟠 Significant — "Fixed tree-wide citation rot" is not what happened; the class is left armed

The PR body claims the rot *"both renumberings caused"* was fixed *"across 9 docs files plus two stale code comments"*. The two `server/tts-sidecar/**` comments fixed (`main.py:177,255,262,310` and `test_ort_cuda_preload.py:220`, all `A36` → `A28`) were **not** rot this PR's renumbering caused — on `main` those already pointed at the wrong row (main `A36` = cast-id drift #2584; the intended ORT row was main `A29`). They were wave-7-era rot, fixed here. Good — but then the identical class was left standing everywhere else. Verified by resolving each citation against the current register:

| Site | Cites | Intended subject | Correct ID at this head |
|---|---|---|---|
| `scripts/repair-cast-id-drift.mjs:42,99,138,460,1107,1140,2154` | `A29` | cast-id drift Wave 1 (`the-torment`/`lightning-dave`) | `A22` |
| `scripts/repair-cast-id-drift.mjs:115,493,1114` | `A30` | Wave 3 `--apply` (`mayrin`/`coalfall`) | `A23` |
| `server/src/store/cast-audio-currency.ts:49,152` | `A29` | Wave 1 resolver / narrator substitution | `A22` |
| `server/src/store/cast-resolve.repair-pass-contract.test.ts:150,249` | `A29` | Wave 1 resolver | `A22` |
| `server/src/tts/segment-asr-qa.ts:96` | `A34` | catastrophic-WER FP-rate run (#2055) | `A26` |
| `server/src/config/registry.ts:575` | `A34` | same (#2055) | `A26` |
| `server/.env.example:672` | `A34` | same (#2055) | `A26` |
| `docs/features/INDEX.md:45` | `A29`, `A30` | cast-identity Wave 1 / Wave 3 | `A22`, `A23` |
| `docs/features/INDEX.md:406` | `A24` | plan 273 evict-latency | `A17` |
| `docs/features/archive/273-…:1054` | `A25` | plan 273 evict-latency (names its own run sheet, owned by `A17`) | `A17` |
| `docs/features/archive/274-…:994` | `A21–A23` | plan 274 §6 rows | `A14`/`A15` |
| `docs/features/archive/276-…:656` | `A31` | clone-readiness gate (#1980) | `A21` |
| `docs/features/archive/280-…:1473,2697` | `A32` | Wave 1 resolver | `A22` |
| `docs/features/275-clone-voice-language.md:405` | `A23` | plan 275 clone-voice language | `A16` |
| `docs/testing/onbox-sitting-qa-gate.md:9,19,213` | `A9,A10,A11,A12,A13,A21,A22,A34` | the QA-gate pack's rows, incl. plans 174/175 whose rows were discharged in wave 7 | list is stale end-to-end |
| `docs/testing/fs38-wave3-onbox-acceptance.md:2643,2654,2893,3023,3050` | `A25`, `A19`, `A23`, `A35` | torchcodec / mixed-engine spill / Wave 1 / the #1969 row | `A18`, discharged, `A22`, **discharged by this PR** |

Two of these are worth calling out beyond the count:

- `docs/testing/onbox-sitting-qa-gate.md` is a `docs/testing/onbox-sitting-*.md` pack, in the swept directory, with a stale row list in **both** its header (`:9`) and its footer instruction (`:213`) — including rows whose subjects (plan 174 post-synthesis QA gate, plan 175 resource telemetry) left the register entirely in wave 7.
- `docs/testing/fs38-wave3-onbox-acceptance.md:3050` reads *"register row **A35** records that owed run"* about the #1969 voice-mismatch re-run. That row is the one **this PR discharged**. The repo's own "annotate, don't renumber" convention (documented at length in `scripts/check-register-citations.mjs`'s Check A header, and correctly applied by this PR to `onbox-sitting-two-card-boot.md` for old A8) requires an annotation here. There is none, and the ID now silently resolves to the stranded-VRAM row.

Per CLAUDE.md's incidental-findings rule, "it's pre-existing" and "it's only a chore" are both named as void deferral reasons, and a stale derived pointer is the rule's own worked example. Either fix the class in this round or narrow the PR body's claim so the record doesn't assert a sweep that did not happen.

---

### 🟡 Minor

1. **Release-notes pair.** Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched, and the PR body does not mark step 5 not-applicable. CLAUDE.md: *"Skip only when the change has no shippable delta … say so explicitly rather than silently omitting."* Note the `main.py` edits do change three emitted `log.warning` strings, so "pure docs" is arguable — say which, either way.
2. **Issue linkage.** `Refs #1969, #2402` — `#2402` is a PR, not an issue, and the open campaign parent `#2435` ("ops — clear the on-box acceptance register: a staged campaign") is not referenced at all, so neither discharge lands on the ticket tracking them. The `pr-issue-link.yml` check passes on `#1969` alone; this is about the record, not the gate.
3. **"Force-published per owner confirmation."** Step 3 of the before-shipping checklist says to run `npm run check:onbox-register -- --against-published <file>` immediately before publishing and **stop if it disagrees**, and #2272 added `--discharging <ID>` precisely for the not-yet-merged-discharge case this PR is in. "Force-published" reads like that comparator was overridden rather than satisfied. Please record what it reported and whether `--discharging` was used — a force here is exactly #1931's lost-row shape.

---

### ✅ What is solid

- The mechanical invariants hold: Group A is contiguous `A1`…`A35`, the `.md`/HTML pair agrees under `check-onbox-register`, the glance table sums to the stated 58, and the oldest-debt claim re-derives correctly to `A11` in both files.
- The two `server/tts-sidecar/**` changes are genuinely comment-only — the `main.py` hunks sit inside a docstring and inside `log.warning` message literals, `test_ort_cuda_preload.py`'s inside a docstring. No control flow, no identifiers, no assertions touched. `A28` is also the semantically right target: `register.md:2297-2298` is that row's own *"the same torch/lib-only outcome **this row** already recorded as not fixing the bug"*.
- Every citation the PR *did* rewrite in the touched files is correct: `clone-readiness…:7` `A22→A21` ✓, `sidecar-evict-latency…:7,79` `A18→A17` ✓, `onbox-sitting-multilanguage.md:1,4,67,69` `A28→A27` ✓, `onbox-sitting-two-card-boot.md:103` `A13→A12` ✓, and the ~20 intra-register cross-references in the diff all resolve to their intended subject.
- Adding an explicit discharge annotation for old A8 in `onbox-sitting-two-card-boot.md` (rather than renumbering it away) is the right convention, correctly applied.
- The A34 write-up does not hide the awkward part — it states the `too-short` outcome plainly. The problem is the conclusion drawn from it, not concealment.

### Verdict

**Request changes.** Two blocking findings, both about the record this PR exists to keep: A34 is discharged on an observation that cannot fail and that leaves its subject untestable, and A8 is discharged with no observation recorded in any of the three homes the register's own rule names — while `docs/features/222-…` still says the acceptance is owed and blocks `stable`, and the run sheet's own section still says STILL OWED. Both are correctness bugs, so fixing and pushing re-triggers this gate. The two Significant findings (new rot at `onbox-sitting-fs38-wave3.md:46`; the unfinished sweep) should land in the same round.

dudarenok-maker:
## PR review — pass 2 (head 8722cb3f1feca6c4a43e08fdac72c58f9105c98c, depth low)

Scope: the fix commit `8722cb3f` (20 files) read against the head tree, not against its own commit message · verified before probing: Group A headings re-enumerated at head (`A1`…`A35`, contiguous, no gaps/duplicates), glance table recomputed by hand (35+2+4+3+10+2+2 = **58** ✓, matches the stated total and the live view's `58` tile), live-view row IDs enumerated and diffed against the register's (`A1`–`A35`, `B1`–`B2`, `C1`–`C4`, `D1`–`D3`, `E1`–`E10` incl. the duplicated `E6`/`E8` blocked/unconfirmed entries, `G1`–`G2`, `H1`–`H2` — the pair agrees), `main`→head Group-A remap re-derived independently (`A9`–`A33`→`A8`–`A32`, `A35`–`A37`→`A33`–`A35`, old `A8`/old `A34` removed) and every rewritten citation resolved against it, then a fresh tree-wide sweep of `register rows? A<n>` citations outside the files this PR touched.

---

## Pass-1 dispositions

|---|---|---|
| 🔴 1 | A34 discharged on an observation that cannot fail | **STILL OPEN** — narrowed, but the row's own stated bar is still unmet and the residual now has no register row (below) |
| 🔴 2 | A8 discharged with no recorded observation | **PARTIALLY RESOLVED** — the observation is now recorded in all three homes; three self-contradictions remain, and the fix introduced two new wrong-row citations (below) |
| 🟠 1 | New rot at `onbox-sitting-fs38-wave3.md:46`; missing annotation at `fs38-wave3-onbox-acceptance.md:3050` | **VERIFIED RESOLVED** — `:46` now reads `A31` (= Russian XTTS at head ✓); `:3023` and `:3052` both annotated "later renumbered A34, then **discharged and removed 2026-08-27, on-box wave 9**" ✓ |
| 🟠 2 | Sweep incomplete, class left armed | **PARTIALLY RESOLVED** — all 16 sites in pass 1's table verified correct at head (code: `A22`/`A23`/`A26` all resolve to the right subjects). ≥8 further sites of the identical class remain, two of them in files this PR edited, and the PR **title** still claims "tree-wide" (below) |
| 🟡 1 | Release-notes pair not marked N/A | **VERIFIED RESOLVED** — declared, and it addresses the `log.warning` string edits explicitly |
| 🟡 2 | Issue linkage | **VERIFIED RESOLVED** — `Closes #1969` + `Refs #2402, #2435, #2700`, all plain text outside code spans |
| 🟡 3 | "Force-published" wording | **RECORDED, but the stated reason does not hold** (new 🟡 below) |

---

### 🔴 Blocking — A34 is still discharged against an outcome its own row named as the failure case, and the unproven half is now tracked *off* the register

Two independent halves.

**(a) The row's bar was not met — and the narrowed text does not change that.** `docs/testing/onbox-acceptance-register.md:361-385` now honestly separates "what this confirms" from "what this does NOT confirm". But the deleted row's own text (`origin/main:docs/testing/onbox-acceptance-register.md:2581-2597`) states the bar in one sentence:

> Only mock/unit coverage exists — **what those cannot prove is that the rebuilt reference, not the failed flag, is what a real render produces.**

The run produced `referenceKind: "too-short"` — the *failed* outcome, which is the exact thing the row exists to distinguish from success. The new text re-labels the residual as "net-new follow-up work, not a defect in this fix", but the row's second criterion is verbatim *"the persisted centroid must be **rebuilt for the new voice**"*, and #2700's Acceptance section is verbatim that same criterion ("Reassign a thin-on-anchors character to a new voice with enough anchor-eligible audio that `auditionCentroid()` produces a real `resolved` outcome"). It is not net-new; it is the unmet half of this row, renamed.

**(b) #2700 is hardware-provable, unproven, and carries no register row.** `grep -n 2700 docs/testing/onbox-acceptance-register.md` → one hit, line 384, inside the changelog prose. No `### A<n> ·` row anywhere; the live view has none either; the owed total went 59 → 58. CLAUDE.md, Testing discipline: *"Behaviour only real hardware can prove … is logged in `docs/testing/onbox-acceptance-register.md` when it cannot be verified inside its own PR … owed acceptance never blocks a merge — **it converts into a row there**."* Before-shipping step 3 says the same and calls the recording a merge gate. A debt that existed as a tracked row at `31234867` is, at `8722cb3f`, an untracked `type:chore` issue. Whichever way the removal policy is read, the register must end this PR carrying a row for #2700's acceptance — and per step 3 that row also has to land in the live view.

**Secondary, and it matters because the narrowed claim now rests on it:** the register asserts as observed fact that *"`matchesCurrentVoice(persisted, voiceInfo)` against the NEW voice correctly returned false … it fell to a genuine `auditionCentroid()` rebuild attempt"*. The recorded evidence cannot establish that. In `server/src/audio/render-integrity/aggregate.ts`, **two** branches write the identical `TOO_SHORT_REF`:

- `:262` — `if (!voiceInfo) return { status: 'too-short', ref: TOO_SHORT_REF }`. Because `:250-253` guards with `voiceInfo != null &&` before `matchesCurrentVoice`, this path never calls `matchesCurrentVoice` **and never attempts a rebuild**.
- `:273` — `if (audition.kind === 'too-short')`, i.e. the rebuild ran and came back short.

The persisted after-state is byte-identical in both cases, so "before = `audition` (old voice), after = `too-short`" does not select between them. The register says the trace was done "against `resolveCharacterReference` … rather than trusting the output shape alone" — that is a code read, and a code read cannot tell you which branch executed. What *would* close it is one sentence recording that `voiceInfo` was non-null for Ivo at that render (cloned voice ⇒ a `STOCHASTIC_ENGINES` engine ⇒ `voiceInfoByChar` populated at `:532-548`); with that stated, the rest of the chain follows deductively. Without it, the confirmed half is asserted, not shown.

---

### 🟠 Significant — the A8 fix introduces two new "register row A8" citations that now point at a live, unrelated row

The discharge's own evidence trail cites the row by its retired number, with no `old`/`former` qualifier:

- `docs/features/archive/222-gpu-residency-and-analysing-honesty.md:77` — `**On-box GPU acceptance — owner-confirmed 2026-08-27** (register row A8, docs/testing/onbox-acceptance-register.md)`
- `docs/features/INDEX.md:392` — `live-GPU acceptance (register row A8) owner-confirmed 2026-08-27`

At this head, `### A8` is **`Batch the QA re-record loops (plan 228)`** (`register.md:950`). Both new lines were added by `8722cb3f`, and both are the precise shape pass 1 flagged at `fs38-wave3-onbox-acceptance.md:3050`: a discharged ID that still resolves, to a different subject. The repo's convention is applied correctly three lines away in this same PR — `onbox-sitting-two-card-boot.md:5` says *"old A8 … discharged 2026-08-27 (on-box wave 9) and removed from the register"*, and the register's own changelog at `:356` says *"**old A8**"*. These two sites need the same annotation.

---

### 🟠 Significant — three "still owed" statements survive inside the A8 discharge's own evidence trail

Pass 1 blocked partly because *"three tracked documents contradict each other"*. The count is down, not gone, and one of the survivors is inside the plan the discharge now rests on:

1. `docs/features/archive/222-gpu-residency-and-analysing-honesty.md:54` — the section heading is still **`### Manual acceptance walkthrough (USER-RUN, live GPU — OWED)`**, in a file whose frontmatter now reads `status: stable` and whose Ship notes say all five steps were observed. This is not cosmetic: `docs/testing/onbox-acceptance-staleness-audit.md:447-450` derives its `STILL OWED` verdict for this row *from this exact heading* (*"Manual walkthrough (`:54-59`) is titled 'USER-RUN, live GPU — OWED'"*). The next staleness pass re-derives the same wrong verdict from the same string.
2. `docs/testing/onbox-sitting-two-card-boot.md:118` — `None excluded. All four rows were re-resolved against live repo/issue/PR state and **remain owed**:` — directly above the A8 bullet the fix annotated as *"superseded, no longer owed."* The bullet was updated; the sentence introducing it was not.
3. `docs/testing/onbox-sitting-two-card-boot.md:18-19` — `**Running time total (recomputed 2026-08-21):** **~110 minutes** of runnable acceptance — A2 step 9 ≈ 20, A3 ≈ 45, **A8 ≈ 25**, A18 ≈ 20.` The pack still bills 25 minutes of runnable acceptance for a discharged row, so the total is overstated by exactly that. (Same line, and `:36`, `:47`, `:123`: `A18` — the pack's header at `:1`/`:4` and its own row heading at `:103` call that row `A12`. Stale in both directions; pass 1 named this at `:19` and it is unchanged.)

---

### 🟠 Significant — the class is still armed, and the PR title still says "tree-wide"

All 16 sites from pass 1's enumerated table check out at head — `repair-cast-id-drift.mjs` (`A22`/`A23`), `cast-audio-currency.ts`, `cast-resolve.repair-pass-contract.test.ts` (`A22`), `segment-asr-qa.ts`/`registry.ts`/`.env.example` (`A26`), `INDEX.md:45`/`:406`, plans 273/274/276/280/275, and `onbox-sitting-qa-gate.md`'s renumbering note (spot-checked all six mappings: `A9→A8`, `A10→A9`, `A11→A10`, `A21→A14`, `A22→A15`, `A34→A26` — every one resolves to the right subject at head ✓). That is real work and it is correct.

But pass 1's table was a sample, not the population, and it was treated as the population. A fresh sweep of `register rows? A<n>` outside the touched files:

| Site | Cites | Intended subject | Correct at this head |
|---|---|---|---|
| `e2e/orphaned-character-fallback-banner.spec.ts:394` | `A29` | cast-id drift Wave 1 — names `the-torment` (*Playing with Fire*, 67 segments), the identical fixture the fixed comments cite | `A22` |
| `docs/features/278-cast-character-identity.md:602` | `A30` | cast-id Wave 3 `--apply` — *Заказ Коалфолла* ch2 re-render + Cast-screen banner | `A23` |
| `docs/superpowers/specs/2026-08-06-cast-identity-followups-design.md:327` | `A29` | cast-id Wave 1 resolver | `A22` |
| `docs/testing/onbox-sitting-cloning-identity.md:223` | `A30` | *"already DISCHARGED (register row A30, run 2026-08-05)"* = the Wave-3 `--apply` run | `A23` |
| `docs/features/249-fs60-xtts-language-eligibility.md:90` | `A19` | plan 249's own acceptance debt | `A5` |
| `docs/superpowers/plans/2026-07-28-coqui-residency-eviction.md:1393` | `A20` | *"the acceptance debt is register row A20"* — plan 249's, per the same sentence | `A5` |
| `docs/release-notes-next.md:191` | `A19` | *"owed on-box as register row A19"* — plan 249's accepted limitation #4 | `A5` |
| `docs/release-notes-next.md:224` | `A34` | the respawn-budget deadline-timer row (#2037), **discharged in wave 8** | needs the "old A34, discharged" annotation — `A34` now resolves to the cast-id #2584 row |

The first four are the exact `A29`/`A30` → cast-identity class this PR fixed in `scripts/` and `server/src/`; `e2e/orphaned-character-fallback-banner.spec.ts:394` is a **code** file, which also falsifies the body's "2 code comments" scoping. `onbox-sitting-cloning-identity.md` is a file this PR edited (29+/20−). Two of them (`coqui-residency-eviction.md:1393`, `release-notes-next.md:191`) point at rows whose *current* occupants are plausible-looking VRAM/stranded-pool rows — the dangerous variant, where the pointer still resolves.

Separately, the same two documents carry `register, ~line 1508` style line-number citations into a file this PR shifted by ~19 lines (`release-notes-next.md:143`, `e2e/orphaned-character-fallback-banner.spec.ts:398`); the ~40 `register.md:<line>` citations in `onbox-acceptance-staleness-audit.md` are explicitly point-in-time evidence and read fine as historical, but the two above are written in the present tense.

Pass 1 asked for one of two things: fix the class, or narrow the claim. The body was expanded — but the **PR title** still reads `fix citation rot tree-wide`, and the test plan still records `node scripts/check-register-citations.mjs — tree-wide citation integrity clean`, which pass 1 already established is a green run the checker is structurally unable to fail on this shape (Check A only fires on nonexistent IDs; every ID above exists).

---

### 🟡 Minor

1. **The live view's own changelog never moved.** `docs/testing/onbox-acceptance-register-live-view.html:200-201` still reads *"Last change: 2026-08-26 (PR #2665 …), 69 → 69 (no count change)"* while the strip six lines above now reads `58` and the register's own last-change entry is wave 9b. The published artifact therefore shows the new count with no account of what produced it. Pre-existing divergence (it was `60` vs `69 → 69` on `main`), but this is the PR that moved the number and the file is one of the three mandated surfaces.
2. **The force-publish rationale doesn't hold as written.** The body says force was used *"since the loop itself was blocking the safety comparator from ever completing normally."* `npm run check:onbox-register -- --against-published <file>` runs offline against a **locally saved copy** of the page currently live at the URL — it needs no successful publish, so a publish-loop hazard cannot block it. And #2272's `--discharging <ID>` exists for exactly this PR's shape (a discharge not yet on `origin/main`). The tracked pair agrees at this head (I diffed the row-ID sets), but that says nothing about the *published* page, which is the only thing `--against-published` checks and the only place #1931's lost-row race lives. Run it with `--discharging` before merge and record what it reported.
3. **#2700 carries `type:chore` with no `area:` label** — CONTRIBUTING's backlog-shaped convention is `area:<prefix>` + `type:`. (`Closes #1969` is also a no-op — #1969 has been `CLOSED` since the fix merged — harmless, but #2435 is the ticket this wave actually advances.)

---

### Verdict

**Request changes.** The A8 half is genuinely close: the observation is now recorded in all three homes, the plan is archived with Ship notes, INDEX moved, and the five `Result:` lines are filled. What blocks is A34: the row was removed on the outcome its own text names as the failure case, and the unmet half now lives as a GitHub issue with no register row, so the owed total dropped by one while nothing about that behaviour became proven. The two new `register row A8` citations and the three surviving "still owed" statements are the same round's fix creating the next round's rot, and the sweep's population is still larger than the sample it was fixed against. One 🔴 and three 🟠 are correctness bugs, so fixing and pushing re-triggers this gate — that would be round 2 of the 2-round cap.

dudarenok-maker:
## PR review — pass 3 (head 9239c1513446d680ec5f8d9ea0e1e1fc116461c5, depth low)

Scope: the fix commit `9239c151` (13 files) read against the head tree, not against its commit message · verified before probing: Group A headings re-enumerated at head (`### A1`…`### A36`, **contiguous, no gaps, no duplicates**, 36 rows); glance table recomputed by hand (`register.md:340-350` → 36+2+4+3+10+2+2 = **59** ✓, matching the stated total at `:352` and the live view's `59` tile at `:192`); live-view row-ID set enumerated and diffed against the register's (`A1`–`A36`, `B1`–`B2`, `C1`–`C4`, `D1`–`D3`, `E1`–`E10`, `G1`–`G2`, `H1`–`H2`, plus the 5 blocked / 2 unconfirmed duplicate entries — the pair agrees); **all three** HTML count sites checked (`:192` strip `58→59` ✓, `:290` glance table `35→36` ✓, `:306` `gcount` `35→36 rows` ✓, and `Blocked 5`/`Unconfirmed 2` agree with the `.md`); changelog arithmetic re-derived against `origin/main` (main = 37 Group-A rows / **60 owed** → wave 9 `−1` → 9b `−1` → 9c `+1` = 36 rows / 59 owed ✓, which also makes the body's `--discharging A37` account arithmetically coherent); `aggregate.ts` re-read for the new deductive chain; then a fresh citation sweep over **every file this commit touched**.

---

## Pass-2 dispositions

|---|---|---|
| 🔴 (a) | A34 discharged against its own named failure outcome | **VERIFIED RESOLVED** — the split is real. `### A36` exists (`register.md:2878-2907`) with all three criteria, is mirrored in the live view (`:833-846`), and is counted correctly in **every** location: `.md` glance table, `.md` total, HTML strip, HTML glance table, HTML `gcount`. Numbering contiguous, no gap, no duplicate. |
| 🔴 (b) | The unproven half tracked *off* the register | **VERIFIED RESOLVED** — owed total returns 58 → 59; #2700 is now the criteria home cited *from* a row, not a substitute for one. |
| 🔴 (secondary) | Recorded evidence can't establish which `too-short` branch fired | **RESOLVED, with one unstated premise** — the chain is now written out (`register.md:385-394`) and `STOCHASTIC_ENGINES` does include `'coqui'` (`aggregate.ts:91` ✓). But it cites **one conjunct of a three-conjunct guard** — see 🟡 1 below. |
| 🟠 1 | Two new "register row A8" citations | **VERIFIED RESOLVED** — `archive/222-…:78` and `INDEX.md:392` both now annotated "…at time of discharge … now names an unrelated row" ✓ |
| 🟠 2 | Three surviving "still owed" statements | **PARTIALLY RESOLVED** — all three fixed (`archive/222-…:54` heading ✓, `staleness-audit:440-450` superseded banner ✓, `two-card-boot:18-20` recomputed to ~85 min ✓, `:120` sentence ✓). The `A18`-vs-`A12` half of that same finding is **not fixed, and this commit re-authored one of the offending lines while preserving the wrong ID** — 🟠 1 below. |
| 🟠 3 | Class still armed; PR title says "tree-wide" | **PARTIALLY RESOLVED** — title narrowed ✓, body now caveats the checker's clean run ✓, and all 8 enumerated sites resolve correctly at head (`249-…:90`→`A5` ✓, `278-…:602`→`A23` ✓, `cast-identity-followups-design:327`→`A22` ✓, `cloning-identity:223`→`A23` self-ref ✓ and `:225`→`A22` ✓, `coqui-residency-eviction:1393`→`A5` ✓, `release-notes-next:191`→`A5` ✓, `:224` annotated ✓, `e2e/…:394`→`A22` ✓, `INDEX.md:392` ✓). **But ≥6 further sites of the identical class survive inside `docs/release-notes-next.md` — a file this commit edited — including the identical sentence 22 lines above one it did fix, and three that this commit *newly* broke by minting `A36`** — 🟠 2 and 🟠 3 below. |
| 🟡 1 | Live-view changelog never moved | **VERIFIED RESOLVED** — `live-view.html:200-217` now carries the wave-9c entry and demotes the stale one to "Historical, predates wave 7" (`:218`) ✓ |
| 🟡 2 | Force-publish rationale | **VERIFIED RESOLVED** — this round records a real `--against-published --discharging A37` run with the row-count delta accounted for; the arithmetic checks out against `origin/main` (37 → 36). |
| 🟡 3 | #2700 missing `area:` | **VERIFIED RESOLVED** — `area:srv` + `type:chore` ✓ |

---

### 🟠 Significant — this commit re-wrote two `two-card-boot` lines and preserved `A18` in both, in a file whose own header calls that row `A12`

`docs/testing/onbox-sitting-two-card-boot.md:120`, a line **authored by `9239c151`**:

> None excluded at the time of this re-resolution pass. Of the four rows, A2/A3/**A18** remain owed; A8 was discharged 2026-08-27…

The same file's title (`:1`), its scope sentence (`:4`) and its own row heading (`:105`) all say **A12**:

- `:1` — `# On-box sitting pack — two-card boot (A2, A3, A12)`
- `:105` — `### A12 · Device-pin resolution survives a respawn (#1870, closes #1857)`

At this head `A18` is **`Cloned-voice derive on Coqui no longer needs torchcodec (#1967) · single 8 GB card + a real static-FFmpeg box; item 4 needs a Pinokio install`** (`register.md:1245`) — a different subject, on different hardware, and not a 2-card-boot row at all.

Concrete break: someone scheduling this sitting from `:120` reads "A18 remain owed", opens `A18`, and provisions a static-FFmpeg + Pinokio single-card run this pack does not cover — while the pack's actual fourth row (`A12`, device-pin respawn, 2-card boot) reads as *not* owed and never gets scheduled.

Five sites, all still `A18`: `:20` (the running-time line **this commit rewrote** — `A2 step 9 ≈ 20, A3 ≈ 45, A8 ≈ 25, A18 ≈ 20`), `:38`, `:49`, `:120` (**newly authored**), `:127`. Pass 2 named this at `:36`/`:47`/`:123`; the commit touched two of the five and carried the wrong ID through both. (Reviewer brief shape 10 — one instance fixed, the class left armed, here inside a single file.)

---

### 🟠 Significant — `#2107's ruling … see A22's own text` is a NEW wrong-subject pointer; the ruling is in `A23`

Two sites, both rewritten by `9239c151` from a stale `~line 1508` to a row ID:

- `e2e/orphaned-character-fallback-banner.spec.ts:398` — `#2107's ruling (same register, see A22's own text) is that only the 'exact' tier means the rendered bytes are fine`
- `docs/release-notes-next.md:143` — same sentence, same rewrite

`A22` (`register.md:1513-1558`) contains **no `#2107` reference and no `'exact'`-tier ruling** — I read the row end to end. The ruling lives in **`A23`**:

- `register.md:1573-1575` — "the #2107 fix's numbers … folded into the #2107 writeup below"
- `register.md:1600-1640` — the `#2107` **FIXED, then WIDENED** writeup
- `register.md:1851-1860` — the ruling itself: *"17 re-render rows, 120 segments — **unconditional on auto-record status**; writing an alias fixes metadata attribution, not the audio bytes already on disk."*

`A22` *is* cited correctly two clauses earlier in the same comment (the `the-torment` fixture, `register.md:1526-1531` ✓). The `#2107` clause needed `A23`. Concrete break: a reader who follows the newly-fixed pointer finds nothing about #2107 in `A22` and concludes the comment's premise is unsourced — the same failure the fix removed, relocated one row over. Converting a stale *line* pointer into a stable *row* pointer entrenches the error rather than closing it.

---

### 🟠 Significant — six more sites of this exact class survive in `docs/release-notes-next.md`, three of them broken by minting `A36`

This commit edited `docs/release-notes-next.md` at `:143`, `:191`, `:224`. The same file still carries:

| Site | Cites | Intended subject | Correct at this head |
|---|---|---|---|
| `:115` | `A33` | "See the on-box acceptance register (row A33) for the full detail" — the cast-id repair pass | `A23` |
| `:119` | `A33` | "a concurrent session ran A33's `--apply` for real (register row A33, **2026-08-05**)" — `A23`'s own text records that same 2026-08-05 `--apply` run | `A23` |
| `:121` | `A32` | "register row A32 has the counter-example on file: *Playing with Fire*'s `the-torment` (67 segments)" — **verbatim `A22`'s fixture** (`register.md:1526-1531`) | `A22` |
| `:121` | `A33` | "Full detail and the measured console output: register row A33" | `A23` |
| `:524` | `A36–A39` | the Kokoro / ORT-marker rows | `A28`–`A30` |
| `:526` (×2) | `A36` | "the exact torch/lib-only outcome register row A36 already recorded as not fixing the bug" / "Register row A36 gained a note naming the four `[ort-preload]` outcomes" — the ORT-marker row | `A28` |

Two things make this worse than a leftover:

1. **`:121` is the identical claim the commit fixed at `:143`.** Same sentence, same fixture, same file, 22 lines apart — `A32`→`A22` was applied to one and not the other. At this head `A32` is `Named-entity decode reaches the TTS engine on a real EPUB (#2310)`: live, plausible-looking, wrong. `A33` is `Kokoro's silent-CPU-fallback alarm (#2647)`: same shape.
2. **The `A36` sites were broken *by this commit*.** At `8722cb3f` Group A ended at `A35`, so `A36` was a dangling ID resolving to nothing. `9239c151` mints `### A36 · Voice reassignment…`, so those three citations now resolve to a **live, unrelated row** — the "dangerous variant" both prior passes named. The commit applied exactly the right treatment to the mirror case at `:224` (annotating old `A34` as discharged and reused); the ID it created itself got none. The same collision is visible inside the register too: `register.md:421` (the dated wave-8 entry) reads "**A36** (cast-id drift, renumbered from old A37)" while `### A36` is now voice reassignment.

`docs/release-notes-next.md` is in `FROZEN_EXACT` (`scripts/check-register-citations.mjs:384`), so **no checker run can ever surface any of these** — which is precisely why the body's "checker clean ≠ sweep complete" caveat needed to be acted on, not only stated.

---

### 🟠 Significant — the two silent renumbers at `:143`/`:191` edit a file the repo's own checker declares frozen against that exact operation

`scripts/check-register-citations.mjs:374-388` places `docs/release-notes-next.md` in `FROZEN_EXACT`, and states the rationale in the file itself:

> `release-notes-next.md` / `RELEASE_NOTES.md`: **shipped-entry diaries; an entry describes what was true when it shipped**, per this repo's release-notes convention (CLAUDE.md "release notes = diary, not a live index").

…with the sibling entry (`:364-371`) spelling out the prohibited operation verbatim: *"silently renumbering an ID inside one breaks that citation without fixing anything (**the row IDs it names are historical facts, not live pointers**)."*

`:143` (`A32`→`A22`) and `:191` (`A19`→`A5`) are silent renumbers of shipped diary entries: the file now asserts that the #2129 entry shipped saying `A22` and the #1893 entry shipped saying `A5`, neither of which is true. The compatible treatment is the annotation this same commit used correctly at `:224` ("…was the open question recorded as on-box acceptance row A34 **at the time** — that row has since been discharged"), and used correctly again in the other frozen file it touched (`onbox-acceptance-staleness-audit.md:440-450`, also `FROZEN_EXACT`).

**This is the one finding here that needs a decision rather than a mechanical fix**, and it determines what "fix the six sites above" even means: either (a) annotate all of them, and revert `:143`/`:191` to annotations, or (b) rule that frozen-diary IDs *are* live pointers and update the checker's stated rationale to match. What cannot stand is both conventions shipping inside one commit. Per the skill's judgment-call carve-out this routes to the repo owner rather than to another automatic fix round.

---

### 🟡 Minor

1. **The `voiceInfo` deduction cites one conjunct of a three-conjunct guard.** `register.md:385-394` argues: cloned Coqui voice ⇒ `snap.voiceEngine ∈ STOCHASTIC_ENGINES` ⇒ `voiceInfoByChar` populated ⇒ `voiceInfo` non-null ⇒ the `:262` branch is ruled out. `STOCHASTIC_ENGINES` does contain `'coqui'` (`aggregate.ts:91` ✓), but the population guard at `aggregate.ts:532` is `!voiceInfoByChar.has(charId) && snap.voiceEngine && snap.resolvedVoiceName && STOCHASTIC_ENGINES.has(snap.voiceEngine)` — and `resolvedVoiceName` is legitimately omittable: `character-snapshots.ts:73` writes `resolvedVoiceName: resolvedVoiceName || undefined`, and `character-snapshots.test.ts:110-124` pins *"omits `resolvedVoiceName` rather than asserting a voice that was never sent, when `voiceNameByChar` carries no entry for a speaking character"* — the same absent-`resolvedVoiceName` condition `aggregate.ts:257-261`'s own comment names as what produces `voiceInfo === undefined`. The fact that closes the gap is already in the same paragraph but never joined to the chain: **Ivo's 8 lines were synthesised in that very render**, so `voiceNameByChar` carried an entry for him and `resolvedVoiceName` was written. One clause fixes it. (Reviewer brief shape 9.)
2. **`A36`'s own body cites the retired ID bare.** `register.md:2880` and `live-view.html:834-835` both read "Split out 2026-08-27 from the just-discharged **A34**" / "Split from A34 2026-08-27" with no `old` qualifier, while `### A34` at this head is `Cast/analysis characterId drift — #2584/#2570` (`register.md:2741`). The changelog three screens up gets this right (`:357`, "**Old A34**"). The inline parenthetical "(voice reassignment vs. persisted audition centroid)" keeps it from being actively misleading, but it is the same shape the PR exists to remove, inside the row the PR created.

---

### Verdict

**Request changes — and this is the third and final round under the 2-round cap, so the loop stops here rather than re-firing.**

Genuinely closed: the A34 split is correct and, on the mechanical claim I was asked to scrutinise hardest, **complete** — `A36` is counted correctly in all five count locations across both surfaces, numbering is contiguous, the `.md`/HTML row sets agree, and the 60→59→58→59 arithmetic reconciles against `origin/main`. The A8 discharge, the release-notes/issue-linkage/label items, and the live-view changelog are done. All eight enumerated citation fixes resolve to the right subject.

Not closed: this round repeated the shape it was dispatched to fix. `9239c151` authored `two-card-boot:120` with the wrong row ID, converted a stale line pointer into a wrong *row* pointer at two sites, left the twin of a sentence it fixed 22 lines above it, and minted `A36` onto an ID that three live citations in a file it was editing already used for something else. Four 🟠 findings, one of which (the frozen-diary convention) is a decision the repo owner has to make before the other six sites can be fixed correctly — annotate, or renumber, but not both in one commit.

Recommendation: hand the frozen-diary question to the owner first, since it determines the fix shape for the `release-notes-next.md` sites; the `two-card-boot` `A18`→`A12` sweep and the `#2107`→`A23` pointer are unambiguous and can be fixed independently of it.

Issue #2260, opened 2026-08-11
chore(server): nothing enforces not-linked-edges.ts's lock contract — a wrong-variant call hangs with no diagnostic


dudarenok-maker:
## What

#2239 moved `clearNotLinkedEdgesForDroppedRejections` into `server/src/store/not-linked-edges.ts` and split it into a pair:

- `clearNotLinkedEdgesForDroppedRejections(bookDir, bookId, dropped)` — takes its own `withCastLock`;
- `clearNotLinkedEdgesForDroppedRejectionsLocked(characters, bookId, dropped)` — pure, for callers that already hold the lock.

The precondition is stated as a module contract in the header (that was #2239's Acceptance 2, and it is met). **Nothing enforces it.** A future caller that invokes the *locking* variant while already holding that book's cast lock violates CLAUDE.md cast-lock rule 1 and hangs forever — no timeout, no diagnostic, no log line.

## Why this is worth a ticket rather than a shrug

This is not a regression — the precondition was equally unenforced when the helper lived in `routes/analysis.ts`, and #2239's split makes it *more* visible than before, not less. But #2239 converted a file-scoped assumption into a cross-module API, which is exactly the point at which "documented in a header comment" stops being sufficient: the header is no longer in the file the caller is editing.

The existing `cast-lock.guard.test.ts` cannot catch this. It is a deliberately syntactic scan for *unlocked writes*; a wrongly-*doubly*-locked call is invisible to it, and its own header already lists call-graph blindness as a known limitation.

Related precedent for the failure mode: `qwen-voice.ts`'s two functions are split rather than wrapped for exactly this reason — wrapping either whole body self-deadlocks because its series branch calls the now-locked `forEachMatchingCastCharacter` while holding that book's key.

## Why it needs design rather than a mechanical fix

Several shapes are available and they are not equivalent:

- a runtime re-entrancy check in `withCastLock` (per-book held-key set + throw) — catches every violator including ones a scan cannot see, but adds state to a hot path and changes a hang into a throw, which is a behaviour decision;
- a static call-graph guard — no runtime cost, but `cast-lock.guard.test.ts`'s header documents why the existing scan is deliberately call-graph-blind, and reversing that is its own design call;
- an ESLint rule scoped to `store/` — cheap, but only covers direct textual call sites;
- accept and document — the status quo, which is defensible given there is exactly one locked-variant caller today.

[...]

Issue #2015, opened 2026-07-31
srv-82 — analysis.ts's five cast.json writes replay a merge base read at the start of the run


dudarenok-maker:
## What

`analysis.ts` writes cast.json at five sites and is the module that writes it most often. None of them is a wrappable read-modify-write, so #2000's per-book cast lock deliberately does not cover them.

`priorCastForMerge` is read **once** at `analysis.ts:2797`, reconciled at `:2823`, and is then the merge base for writes at `:3558`, `:3763` and `:4774` — spread across the whole Phase 0/1 pipeline, with the interim writes *inside* the chapter loop. The subset route has the identical shape (read `:5206`, writes `:5422`, `:5927`).

So the window between the read and the last write is an entire analysis run — minutes on a full book. Any cast.json write by another route during that window is replayed over and silently lost when the next interim write lands.

## Why #2000 could not solve it

Both available strategies fail, which is why this is its own issue rather than a task in that one:

- **Hold the cast lock for the run.** Blocks every other cast write on that book for minutes, on a mutex (`workspace/file-lock.ts:7-19`) with no timeout, no queue cap and no diagnostic. Converts a lost mutation into an unbounded hang.
- **Re-read inside each write's lock.** Changes behaviour: iteration N would re-read the interim cast iteration N−1 wrote, so `mergeAnalysisResultWithExistingCast` no longer merges against the pre-run cast and srv-13's voice/reuse carry-forward silently degrades.

An admission gate was also considered and rejected — see below.

## Why the isAnalysisBusy admission gate was rejected

The obvious cheap mitigation is to have cast writers consult `isAnalysisBusy` (`tts/design-lock.ts:63-65`) and refuse. Rejected for three reasons, recorded so it is not re-proposed:

1. **It is itself check-then-act** — the exact defect class #2000 exists to close. `isAnalysisBusy` is a bare `Map` read with no coupling to the cast mutex. A writer that reads `false`, then `await`s its cast read, then writes, races a run that marks busy (`:2572`, `:5109`). The consultation's window is wider than the one #1981 filed.
2. **Most sites cannot express a refusal.** Seven of the 30 non-analysis write sites are internal helpers or run detached after `res.flushHeaders()` (`cast-design.ts:293`/`:452`, `qwen-voice.ts:177`/`:215`, `voices.ts:801`/`:828`, `voice-library-usage.ts:113`), plus `single-design.ts:178`'s call into `voices.ts`.
[...]

PR #2699, opened 2026-08-27
fix(server): advance the merge-base baseline on the run's own out-of-band cast.json writes

### LINKED ISSUE #2694: srv-83 — the run's own cast.json writers don't advance the merge-base baseline, so an uncontended analysis can emit cast_merge_base_stale
## What

#2185 shipped a compare-and-set baseline (`CastMergeBase`, `server/src/workspace/cast-merge-base.ts`) over cast.json's merge base, advanced at each of `analysis.ts`'s five `writeChecked` sites. A mismatch emits a `cast_merge_base_stale` advisory meaning "another route wrote your cast".

**The analysis run itself writes cast.json outside `writeChecked`, without advancing the baseline.** Two sites, both verified:

- `clearNotLinkedEdgesForDroppedRejections` (`server/src/store/not-linked-edges.ts`) — takes `withCastLock(bookDir)` and `writeJsonAtomic(castJsonPath(bookDir), …)`. Reached from `recordRetirements` (`analysis.ts`), called at **job start** — after `readPriorCastForMerge`'s capture and before all five write sites, i.e. inside the capture-to-first-write window.
- `reconcileRejectEdgesOnDisk` (`analysis.ts`) — same shape, at persist.

Neither touches `CastMergeBase`. So the next `writeChecked` re-reads a file this run legitimately changed, sees a fingerprint that is not its baseline, and reports a foreign conflict that did not happen.

Reachable whenever a run records a retirement that drops a self-loop rejection (`result.droppedSelfLoopRejections`), which needs `dedupePriorCastByName` to have folded duplicate-name prior rows — routine on a real book, absent from the stub fixtures.

## Why it matters beyond the false toast

1. **#2185's negative control is weaker than it reads.** It asserts an uncontended run emits zero advisories; it passes because its fixture never drops a rejection, not because the invariant holds.
2. **It contaminates the frequency measurement.** The 2026-08-06 design (§8) made an explicit bet that frequency data would decide whether #2015's repair half is worth building. If some advisories are the run's own writes, that data does not answer the question it was collected for.
3. B4 (PR #2190) measured a *contended* run and is unaffected — it observed exactly one advisory and a real alias loss. This does not weaken B4.

## Acceptance

- An uncontended run that **does** drop a self-loop rejection emits **zero** `cast_merge_base_stale` advisories — a regression test with a fixture that actually exercises the path (the existing negative control does not).
- Whichever mechanism is chosen, the run's own writers and the baseline cannot drift apart silently again.

## Design decision owed

More than one defensible fix, so this needs a design pass rather than a straight patch:

- route both writers through `writeChecked` (they already hold the same lock class, but cast-lock rule 1 forbids a locked function calling another locked one on the same book — so this is not a wrap);
- or plumb `CastMergeBase` in so they advance the baseline off their own written bytes;
- or re-capture the baseline after `recordRetirements`, which is cheapest but leaves the persist-time site uncovered.

## Key files

- `server/src/store/not-linked-edges.ts` — `clearNotLinkedEdgesForDroppedRejections`
- `server/src/routes/analysis.ts` — `recordRetirements`, `reconcileRejectEdgesOnDisk`, the five `writeChecked` sites
- `server/src/workspace/cast-merge-base.ts`

## Benefit

*Benefit (technical):* makes the merge-base detector's advisory mean what it says, and makes the frequency data #2015's repair half is gated on actually measure foreign writes.

Found while designing #2015's repair half; recorded in `docs/superpowers/specs/2026-08-27-cast-merge-base-foreign-delta-design.md` §13.3.

Refs #2015, #2185

dudarenok-maker:
## Summary

#2185 gave `analysis.ts`'s five merge-base writes a per-run compare-and-set baseline: each re-reads `cast.json` under the cast lock, compares a sha256 of the raw bytes against the run's own baseline, and emits a `cast_merge_base_stale` advisory when they differ.

**But `writeChecked` is not the only writer a run has.** Two paths write `cast.json` outside it, both taking the cast lock themselves and neither advancing the baseline:

- `recordRetirements` → `clearNotLinkedEdgesForDroppedRejections` (`server/src/store/not-linked-edges.ts`) — at **job start**, after the capture and before all five `writeChecked` sites;
- `reconcileRejectEdgesOnDisk` (`server/src/routes/analysis.ts`) — at persist.

So the next `writeChecked` re-read saw bytes *the run itself* had changed and reported a foreign conflict that never happened.

**The fix** is `CastMergeBase.noteExternalWrite(payload)` — advance the baseline for a write this run made through another path. It takes no lock and does no I/O, which is also why these two writers are **not** routed through `writeChecked`: they already hold the cast lock, and `writeChecked` takes its own, which would violate cast-lock rule 1. `clearNotLinkedEdgesForDroppedRejections` now returns the payload it wrote (or `null`); its documented MODULE CONTRACT and its #2260 lock-timeout-vs-disk-fault discrimination are untouched.

## The test is the substance, and it was mutation-verified

**#2185's four existing negative controls pass vacuously against this bug** — no fixture ever drives a dropped self-loop rejection, so none of them reaches the offending write at all. That is exactly why this shipped.

**Control 8** builds a fixture that does reach it: duplicate-name prior rows folded by `dedupePriorCastByName`, a pre-seeded `rejectedPair` that the resulting retirement collides back onto itself (a dropped self-loop), and a real `notLinkedTo` edge to clear. It asserts the pair was **genuinely dropped** before asserting zero warnings, so it cannot go vacuous the same way. A **symmetric positive control** proves a genuine foreign write is still reported exactly once — without it, "zero advisories" would be equally satisfiable by breaking detection outright.

Red proven by neutralising `noteExternalWrite` to a no-op on the finished branch and re-running — note the anti-vacuity assertion on line 844 **passes**, so the fixture provably reached the path, and the failure is the phantom advisory itself:

```
 FAIL  src/routes/analysis.merge-base-detect.test.ts > #2015 Task 9 — route-level cast_merge_base_stale controls > control 8 — an uncontended run whose OWN dropped-self-loop cleanup emits ZERO stale warnings (route-level, main)
AssertionError: expected [ 'cast_merge_base_stale' ] to deeply equal []

- Expected
+ Received

- []
+ [
+   "cast_merge_base_stale",
+ ]

 ❯ src/routes/analysis.merge-base-detect.test.ts:845:38
    844|         expect(readCastIdHistoryRejectedPairs(bookDir)).toEqual([]);
    845|         expect(warningCodes(events)).toEqual([]);
       |                                      ^
```

Probe removed, `noteExternalWrite` restored from the index, full file green: **9 passed (9)**.

## Known residual — recorded, deliberately not fixed

`clearNotLinkedEdgesForDroppedRejections` is a **fresh-read** RMW, so a genuine foreign write landing before its read is absorbed into that read rather than detected. **No data is lost** — the fresh read preserves whatever that write left — but that one foreign write goes unreported. A detection gap, strictly better than the false positive this PR removes, and not worth widening scope to close. Named in a comment at the site.

## On enforcement, stated honestly

The invariant ("any write by this run, through any path, must advance the baseline") is documented in `cast-merge-base.ts`'s header and at both call sites. **No guard test is claimed.** The cast-lock guard is syntactic and call-graph-blind, and a test mechanically enumerating run-reachable cast writers cannot be written without false positives. A documented convention that says so beats a guard that reads as coverage and is not.

## Why this matters beyond the toast

Found while designing #2015's rebuild half (PR #2695, merged as a **falsified** design). It **unblocks** #2015 rather than advancing it: the 2026-08-06 design gated the repair on frequency data, and that data would have counted the run's own writes as foreign conflicts. B4 (PR #2190) measured a *contended* run and is unaffected.

## Test plan

- `control 8` and its symmetric positive control — red before, green after, mutation-verified above.
- `npm run verify:fast:branch` — green (lint, typecheck, config:check, test:hooks, frontend, server, build).
- `server/src/workspace/cast-lock.guard.test.ts` green — no new `writeJsonAtomic(castJsonPath(` site is introduced.

**Regression plan:** none owed — small and localized, so the issue body plus the paired test is the spec, per CLAUDE.md step 1.

**On-box acceptance:** none owed. This is provable in-process; no GPU, sidecar, analyzer or real book is required to exercise it.

Closes #2694
Refs #2015, #2185


dudarenok-maker:
## PR review — pass 1 (head c25c5992, depth medium)

Scope: all 10 changed files (`cast-merge-base.ts`, `not-linked-edges.ts`, `analysis.ts`, 5 test files, the release-notes pair) plus their call graph — every `writeChecked` / `recordRetirements` / `reconcileRejectEdgesOnDisk` / `clearNotLinkedEdgesForDroppedRejections` site in `server/src`, `cast-fingerprint.ts`, `state-io.ts`, `cast-lock.guard.test.ts`.

Verified before probing: `npm run typecheck` green (frontend + server); the 6 directly-implicated suites green (`analysis.merge-base-detect` 9/9, `not-linked-edges`, `not-linked-edges.lock-timeout`, `cast-merge-base`, `cast-lock.guard`, `analysis-reject-edge-reconcile` — 50 tests); full server suite 8017 passed / 9 skipped. Three mutation probes run in a **throwaway detached worktree at this head**, torn down after; the reviewed worktree's `HEAD` and `git status --porcelain` are byte-identical before and after this pass.

**The fix itself is correct and the mutation claim in the PR body holds exactly as written.** Reproduced: neutralising `noteExternalWrite` to a no-op turns control 8 red at `analysis.merge-base-detect.test.ts:845` on `expected [ 'cast_merge_base_stale' ] to deeply equal []`, with the anti-vacuity assertion on :844 still passing — verbatim the body's transcript. The body actually **under**-claims: control 9 goes red too (`[ 'cast_merge_base_stale', 'cast_merge_base_stale' ]` vs one, at :903), and the run log shows the phantom firing at `site=interim`, confirming the interim per-chapter `writeChecked` is the real downstream consumer.

One finding is worth fixing before merge. Nothing blocks correctness of the shipped behaviour.

### 🟠 Significant — control 8's anti-vacuity assertion does not pin the write it protects, so one fixture edit silently re-vacuates it

**Correctness bug (in the regression net, not in production behaviour).**

`server/src/routes/analysis.merge-base-detect.test.ts:844` — `expect(readCastIdHistoryRejectedPairs(bookDir)).toEqual([])`, guarding the fixture at `:786` (`notLinkedTo: [{ bookId, characterId: 'anton' }]`).

Concrete repro, run at this head in a detached worktree:

1. Restore the bug — `cast-merge-base.ts:98-101` `noteExternalWrite` → a no-op.
2. Drift the fixture one id — `:786` `characterId: 'anton'` → `characterId: 'someone_else'`.

Result: **`Test Files 1 passed (1)` / `Tests 9 passed (9)`.** Control 8 goes green against the exact bug it was written to catch, and line 844 still passes.

Why: the assertion pins the **rejectedPair drop**, which `retireCharacterId` → `repointRejectedPairs` (`cast-id-history.ts:494-515`) performs off the seeded `rejectedPair` + the `dedupePriorCastByName` retirement alone. The **cast.json write** needs one more link: `clearNotLinkedEdgesForDroppedRejectionsLocked` (`not-linked-edges.ts:67-78`) builds `deadIds` from `dropped.map(p => p.from)` and only sets `changed` — and only then reaches `writeJsonAtomic` at `:112` and returns a non-null payload for `analysis.ts:311` to advance on — if some character's `notLinkedTo` entry matches. Break that match and `written` is `null`, `noteExternalWrite` is never called, and the test passes for precisely the reason the block comment at `:755-773` says #2185's controls 1/2/4/6 passed: the fixture never reaches the offending write.

The comment at `:838-843` overstates what it bought: "proves `recordRetirements` -> `retireCharacterId` -> `clearNotLinkedEdgesForDroppedRejections` all fired" — it proves the third was *called*, not that it *wrote*, and the write is the whole subject of the ticket.

Fix is one line and needs no new fixture: after the run, read `cast.json` and assert olga's `notLinkedTo` is `[]` (or assert the file's bytes moved across the interim step). Control 9 at `:902` carries the same assertion and wants the same line.

This is the one finding I'd class as a re-review trigger — fixing and pushing it should draw one more pass.

### 🟡 Minor — the `reconcileRejectEdgesOnDisk` half of the fix has no observable effect and no test, and the PR body/release note assert a causal story it does not have

**Cleanup.** `server/src/routes/analysis.ts:455` — `castBase?.noteExternalWrite(payload)`.

Every `writeChecked` in the file precedes its own job's `reconcileRejectEdgesOnDisk` call: main `4409 / 4640 / 5897` vs `6005`; subset `6822 / 7520` vs `7601`. The terminal persist block closes at `6092` with only `logCarriedForwardCharacters` and the `state.json` write after the reconcile. So the baseline this line advances is never read again in the current call graph.

Empirically: delete that one line and the **whole** server suite is `Test Files 534 passed | 1 skipped`, `Tests 8017 passed | 9 skipped` — zero failures. (The `Worker exited unexpectedly` unhandled error in that run is the known teardown flake, not a test failure.)

Same is true of the three post-persist `recordRetirements` calls (`5939 / 5940 / 5950`, `7552 / 7553 / 7563`). Only the two **interim** calls — `3604` (main) and `6616` (subset) — are load-bearing today.

I would **keep the line**: it satisfies the invariant the header states, and it is exactly what a sixth write site landing after persist would need. What is wrong is the claim. `docs/release-notes-next.md` and the PR body both read "`reconcileRejectEdgesOnDisk` does the same at persist. Both take the cast lock and write without advancing the baseline, **so the next `writeChecked` read saw bytes the run itself had changed and reported a foreign conflict that never happened**." There is no next `writeChecked` after the reconcile call — that consequence belongs to `recordRetirements` → `clearNotLinkedEdgesForDroppedRejections` alone. The release note is a user-facing shipped artifact; a half-sentence ("the reconcile write is covered for the same reason, though today nothing reads the baseline after it") makes it true.

### 🟡 Minor — "Known residual" is framed as an accepted gap; it is a *new* one, and the window is not a race

**Cleanup.** `server/src/store/not-linked-edges.ts:99-106`.

Pre-#2694, a foreign write landing between the run's capture (`analysis.ts:3535`, `readPriorCastForMerge`) and this fresh read got absorbed into the payload, the baseline stayed at the capture hash, and the next `writeChecked` **did** report — a true positive, arrived at by accident. Post-#2694 that same write is silently absorbed and nothing is reported. The window spans `pruneStaleReuseLinks` + `dedupePriorCastByName` before the interim `recordRetirements` at `3604` — wall-clock, not a microsecond race — and the failure it hides is the one #2185 exists to surface: the user edits a character just as the run starts, the run's final write clobbers it, no advisory.

I agree with the trade (pre-fix that advisory carried no information, because it fired on runs with no foreign writer at all), and the exposure is narrow — it needs a dropped self-loop rejection in the same run. The comment just doesn't say a detection that previously fired now doesn't. One sentence.

### ✅ What is solid

- **The advance is byte-exact at both sites.** `writeJsonAtomic` writes `JSON.stringify(value, null, 2)` (`state-io.ts:123`); `fingerprintOfWrite` hashes the same expression (`cast-fingerprint.ts:122`); and both sites hand the **same object reference** to both calls. `{ ...cast, characters: cast.characters }` cannot reorder keys (re-assigning an already-present key keeps its slot), and cast.json has no sibling keys to carry anyway — all seven `interface CastJson` declarations in the tree are `{ characters?: … }`.
- **The post-lock advance in `recordRetirements` is race-safe, and it is safe *because* it is payload-derived.** A foreign write landing between the hold releasing at `analysis.ts:306` and the advance at `:311` still mismatches at the next `writeChecked` (baseline = our payload, disk = theirs) — correctly reported, not swallowed. An implementation that re-read the file here would have swallowed it.
- **`null` / `ABSENT` interaction holds.** `noteExternalWrite`'s `baseline === null` guard mirrors `markDeleted`'s. No own-write can strand the baseline at `ABSENT`: both out-of-band writers early-return on an absent or empty cast (`not-linked-edges.ts:108`, `analysis.ts:412`), and the interim `recordRetirements` (`3604`) runs strictly before the Start-fresh `rm` + `markDeleted` (`3661` / `3673`), so nothing of the run's own can land in that window.
- **Cast-lock rules intact.** `noteExternalWrite` takes no lock and does no I/O (`cast-merge-base.ts:98-101`). `analysis.ts:455` sits inside the reconcile's own single hold and adds no acquisition; `analysis.ts:311` runs after `clearNotLinkedEdgesForDroppedRejections`'s hold released. No second lock on the same book, no `design`→`library-voice`→`cast` inversion. `cast-lock.guard.test.ts` green, and no new `writeJsonAtomic(castJsonPath(` occurrence — both sites pre-existed, so the file-and-count allowlist is unmoved.
- **#2260 discrimination genuinely untouched.** `not-linked-edges.ts:124` `if (isLockAcquisitionTimeout(err)) throw err;` still precedes the new `return null` at `:129`, so a lock timeout rethrows and never falls through. `not-linked-edges.lock-timeout.test.ts` green; its two `.resolves.toBeUndefined()` → `.resolves.toBeNull()` edits are exactly as strict as before, not a weakening.
- **Plumbing is complete and `tsc`-enforced.** All eight production `recordRetirements` sites pass `castBase` (required param). Both production `reconcileRejectEdgesOnDisk` sites pass it, and `analysis-reject-edge-reconcile.test.ts:353/356/387` pins the exact argument list textually, so a silent omission there fails a test rather than passing quietly. The only omitting callers are tests; the only other production caller of `clearNotLinkedEdgesForDroppedRejections` is `cast-link-orphan.ts:325`, which has no run baseline and correctly discards the return.
- **The only `CastMergeBase` test double in the tree** (`analysis.persist-lock-timeout.test.ts:119-133`) was updated — necessary, since `castBase?.noteExternalWrite(…)` optional-chains the object, not the method, so a stale double would have thrown rather than no-op'd.
- **No weakened assertions** in the other four modified test files; the `analysis.test.ts` edits pass `null` at two unit-level `recordRetirements` calls that have no baseline to advance.
- House gates: `Closes #2694` present, plain, outside any code span; release-notes pair landed in this PR; no regression plan owed (small/localized, issue + paired test is the spec); no on-box acceptance owed (provable in-process); no config knob, no derived artifact staled.

### Verdict

**Nothing blocks merge.** The behaviour change is correct, its regression test is red-before/green-after for the reason claimed, and every house gate is met.

One finding is worth fixing in this round rather than filed: the 🟠 anti-vacuity gap — the test can be re-vacuated by a single fixture edit, which is the exact failure mode this PR was written to close, and the fix is one assertion. The two 🟡s are documentation accuracy and are optional.

dudarenok-maker:
## Review-gate findings — triage

All three folded. One of them **did not fix**, and that is recorded rather than papered over.

**Finding 1 — the proposed fix does not close the hole.** The gate was right that `rejectedPairs`-empty is vacuous w.r.t. the write: it is `repointRejectedPairs`' output, reachable from the seeded pair plus the retirement alone. But its suggested replacement — assert olga's `notLinkedTo` was cleared — is strictly stronger and **still not a pin on the job-start write**. Measured, not reasoned: with `noteExternalWrite` neutralised **and** the fixture's edge drifted to an id `deadIds` cannot match, control 8 still passes, because `reconcileRejectEdgesOnDisk` clears that same edge at persist.

Rather than ship a third guard that reads as coverage and is not — the exact defect class this PR fixes, and the second time in this PR's history that guard was written — both lines are kept as cheap corroboration and the comment now states precisely what they do and do not prove. What actually catches the bug is `warningCodes`, mutation-verified with the fixture intact. A real pin would have to assert the write at the moment it happened, which nothing observable from the route level can do.

**Finding 2 — fixed.** `docs/release-notes-next.md` no longer claims the persist-time `reconcileRejectEdgesOnDisk` advance fixes an observed failure. Confirmed: every `writeChecked` precedes it, so only the interim `recordRetirements` writer produces the phantom. The line stays — it upholds the stated invariant — but it is now described as defensive rather than causal.

**Finding 3 — fixed.** The residual comment in `not-linked-edges.ts` now discloses that the detection gap is **new**: pre-#2694 such a write *was* reported, accidentally, via the stale baseline. Still the right trade, because that advisory fired on the run's own write and so could not distinguish a real foreign write from the phantom — but the disclosure no longer presents a new gap as a pre-existing one.

Thanks for the empirical repro on finding 1 — demonstrating it rather than arguing it is what made the follow-on measurement obvious to run.

Suites green: `analysis.merge-base-detect.test.ts` 9/9, full `test:server` battery passed via the pre-commit hook.

Issue #2692, opened 2026-08-27
[agent instructions][claude][task] State the Coqui residency policy in plan 264


dudarenok-maker:
Step 1 of 3 for **#1932 (side-18 — consolidate the two Coqui VRAM eviction mechanisms)**.

## What

Write the single statement of the Coqui residency policy that #1932 asks for, as a
new section in `docs/features/264-vram-aware-gpu-placement.md`. It names both
eviction triggers, says which owns which case, records the ruling that both are
kept and why, and pins the lock caveats any future change must preserve.

This is a **docs-only** change. No source file is touched in this ticket.

## Background you need — cite the code, not your memory of it

Two independent mechanisms free VRAM held by a resident Coqui XTTS model. Both are
correct, both are tested, and neither knows the other exists.

**Mechanism A — Node, plan-driven, proactive.**
`evictEngineForPhase` (`server/src/tts/synthesise-chapter.ts:1015`) POSTs
`/unload` to the sidecar for a named engine. `evictCoquiForQwenPhase` (`:1041`) is
its Coqui-facing wrapper, called at `:2976` at the end of a mixed chapter's Coqui
phase; the symmetric `evictQwenForCoquiPhase` fires at `:2912`. It is driven by
the **generation plan** — Node knows which engine each remaining chapter needs, so
it can free the VRAM at a boundary where "no further Coqui work is queued" is a
*fact* rather than a prediction. The call is bounded by `withCallTimeout` and
carries the chapter's abort signal. The trailing evict is deliberately fail-soft
(every group is already synthesised by then, so throwing would destroy completed
work purely to free VRAM); the leading one is not.

**Mechanism B — sidecar, demand-driven, reactive.**
`_idle_evict_steps` (`server/tts-sidecar/main.py:5019-5048`) is a **five-step,
cross-engine ladder** — `spk`, `asr`, `qwen.design`, `qwen.base17`, `coqui` — run
by `PlacementController` on the admission path when an op is VRAM-starved. Coqui
is *one step of it* (`:5047`), calling `CoquiEngine.maybe_free_idle` (`:2273`).
Two Coqui-specific qualifications: the step is omitted entirely when the admitting
op is itself Coqui (evicting would unload the model that op is about to reload),
and it is the only step that uses a real idle TTL — `_coqui_idle_ttl()` (`:8272`,
30 s default, `COQUI_IDLE_TTL`, registry key at
[...]

Issue #2689, opened 2026-08-27
[agent instructions][claude][verify] Verify the Coqui residency policy chain, open the PR (#1932)


dudarenok-maker:
Verification child for **#1932 (side-18 — consolidate the two Coqui VRAM eviction mechanisms)**.

You judge the committed diff on the branch. You are **read-only for source**: you
report and verdict, you do not fix.

## Where the work is

- **Worktree:** `C:\Claude\Projects\wt-1932-coqui-residency-policy`
- **Branch:** `chore/sidecar-1932-coqui-residency-policy`

Read the diff with `git -C "C:\Claude\Projects\wt-1932-coqui-residency-policy" diff main...HEAD`.
Do not create a worktree. Do not work in the primary checkout.

## What the chain was supposed to deliver

Three task children ran before you, all committing to that branch:

1. A **Coqui residency policy** section in `docs/features/264-vram-aware-gpu-placement.md`.
2. **Cross-reference comments** at three code sites, each carrying the anchor token
   `COQUI-RESIDENCY-POLICY`.
3. A **guard test** at `server/src/tts/coqui-residency-policy.guard.test.ts` that
   fails if any of those cross-references is removed.

## The parent's acceptance — verdict against this checklist

From #1932, restated:

- [ ] **One documented statement of the Coqui residency policy that names both
      triggers and says which owns which case.** Check the section exists, names
      mechanism A (Node, plan-driven, `evictEngineForPhase`) and mechanism B
      (sidecar, demand-driven, the `_idle_evict_steps` ladder) with `file:line`
      citations, and states plainly that A owns known plan boundaries and B owns
      unpredicted demand pressure, neither being a superset of the other.
- [ ] **Either a single mechanism with the other removed, or an explicit
      code-level comment at each site pointing at the other and at the policy
      doc.** The ruling made at decomposition time was **keep both** — the fork
      is closed, not open. Confirm the doc records *why* (both reasons: removing
      B deletes one step from a five-step cross-engine ladder that stays; removing
      A requires the sidecar to know Node's generation plan across a process
      boundary, and B cannot substitute because it fires only after starvation and
      only past a 30 s TTL). Confirm all three code sites carry the pointer.
- [ ] **The merged lock caveats stay intact** — non-reentrancy of `_synth_lock`,
      and the `_drop_model_locked()`-not-`unload()` rule for a lock holder.
      Nothing executable should have changed at all; verify that by inspection of
      the diff, not by trust.
[...]

Issue #1894, opened 2026-07-27
srv — fs-60's Coqui VRAM residency is never evicted after the last Coqui chapter


dudarenok-maker:
## What

fs-60's mixed-engine VRAM-eviction logic in `synthesise-chapter.ts`'s render
path only evicts **Qwen for Coqui** (frees Qwen VRAM before a Coqui render
phase within a chapter) — never the reverse. So a chapter with **zero** Coqui
presence, immediately following a chapter that used Coqui, never evicts XTTS
— it stays VRAM-resident into a chapter that has no use for it, potentially
crowding out the Qwen render that chapter actually needs.

## Status

Pre-existing (fs-60, predates Wave 3c) — same root commit as the sibling
issue filed alongside this one (the trailing-evict try/catch gap). Surfaced
during Wave 3c's Task 22 review while auditing eviction symmetry.

## Why this isn't a quick fix

An earlier fix-round attempt during Wave 3c's Task 22 tried gating a
*different* eviction call on "any Qwen load is imminent" and had to be
backed out — Qwen is this project's default engine, so that condition was
true for nearly every chapter of nearly every book, and it broke an
unrelated fake-timer batching test. Closing this asymmetry properly needs
its own design (likely something narrower like "does this chapter have zero
Coqui-cast characters AND the previous chapter had Coqui residency", not a
blanket "evict whenever Qwen might run next") rather than a quick patch —
flagged explicitly during Wave 3c's review as needing its own design pass.

## Suggested fix shape

Left open deliberately rather than prescribed — the constraint from the
near-miss above is the main thing to carry forward: any fix must not
over-trigger on the common "next chapter uses Qwen" case, only on the
narrower "Coqui is resident and about to be genuinely unnecessary" case.

## Key files

- `server/src/tts/synthesise-chapter.ts` (the fs-60 render-path eviction
  logic — locate by symbol/behaviour, line numbers drift)

## Related

- fs-38 / #624 (voice cloning umbrella)
- Plan: `docs/features/270-fs38-wave3c-xtts.md`, "Known limitations (l)"
- Sibling issue: the trailing-evict try/catch gap on the same code path

dudarenok-maker:
## Findings from investigating the sibling issue (#1893)

Confirming the premise and adding the constraint that I think is actually the hard part here — it isn't the one this issue currently names.

[...]

PR #2715, opened 2026-08-27
docs(sidecar): state the Coqui residency policy and guard its cross-references

### LINKED ISSUE #1932: side-18 — consolidate the two Coqui VRAM eviction mechanisms
## What

Two independent mechanisms now free VRAM held by a resident Coqui XTTS model. Both are correct, both are tested, and neither knows the other exists. Consolidate them behind one policy — or document the split as deliberate and make each one's scope explicit in code.

**Mechanism A — reactive and local (fs-38 Wave 3c, `bd59cae3`).** `server/src/tts/synthesise-chapter.ts` evicts Coqui via `POST /unload` immediately after the last chapter that needed it, so the Qwen phase does not start against a GPU still holding XTTS. It is driven by the *generation plan*: Node knows which engine each remaining chapter needs, so it can free the VRAM at a known-safe boundary. It is bounded by `withCallTimeout` and it fails loud.

**Mechanism B — idle-driven and global (#1894, shipped on `main`).** `CoquiEngine.maybe_free_idle` in `server/tts-sidecar/main.py` drops the model when *any* op is VRAM-starved at admission, using `_synth_in_flight` and `_last_used` read lock-free so a busy engine short-circuits without blocking. It is driven by *demand*: the sidecar does not know what Node plans next, only that something needs memory now.

## Why this is debt rather than a bug

They compose safely today — A is a no-op if B already dropped the model, and B short-circuits while A's forward is in flight. The cost is comprehension, not correctness: a future change to Coqui residency has two places to reason about, with two different lock disciplines (`_synth_lock` non-reentrancy on our side, `_drop_model_locked()`-not-`unload()` on main's), and no single place that states the residency policy.

The decision to keep both was made deliberately (fs-38 Wave 3c owner call), and it was the right call — but it was made believing B was designed-but-unimplemented rather than already shipped. Revisited with both visible in one tree, keeping both still stands; this issue records the consolidation as owed rather than done.

## Acceptance

- One documented statement of the Coqui residency policy that names both triggers and says which owns which case.
- Either a single mechanism with the other removed, or an explicit code-level comment at each site pointing at the other and at the policy doc.
- The merged lock caveats stay intact — non-reentrancy of `_synth_lock`, and the `_drop_model_locked()` rule for a lock holder. Consolidation must not lose either.
- Existing tests on both sides stay green; a red test from either side means the consolidation deleted behaviour.

## Key files

- `server/src/tts/synthesise-chapter.ts` — `evictCoquiForQwenPhase`, `evictEngineForPhase`, `withCallTimeout`
- `server/tts-sidecar/main.py` — `CoquiEngine.maybe_free_idle`, `_synth_in_flight`, `_last_used`, `_drop_model_locked`, `_idle_evict`
- `docs/features/271-fs38-wave3c-xtts.md`

## Depends on

fs-38 Wave 3c merging (this branch). Not actionable before then.

## Benefit

*Benefit (architectural):* collapses two independently-evolving VRAM policies into one reasoned-about surface, so the next residency change does not have to rediscover that a second mechanism exists.

dudarenok-maker:
## Summary

States the Coqui VRAM residency policy (side-18, #1932): two independent eviction mechanisms — Node's plan-driven `evictCoquiForQwenPhase`/`evictQwenForCoquiPhase` and the sidecar's demand-driven `_idle_evict_steps` `coqui` step — are both kept, with the ownership split and rationale documented in `docs/features/264-vram-aware-gpu-placement.md`. Adds `COQUI-RESIDENCY-POLICY` cross-reference comments at the code sites and a static guard test (`server/src/tts/coqui-residency-policy.guard.test.ts`) that fails if any of them is removed. Closes #1932.

**Also fixed, found in passing** (across four independent PR-review rounds; see PR comments for the full findings):
- The guard test had no CI-selection wiring: a diff touching only `server/tts-sidecar/main.py` or only the policy doc didn't schedule the `test:server` CI leg, and a `synthesise-chapter.ts`-only diff scheduled the leg but `vitest run --changed` never selected the guard. Wired all three into `server/vitest.config.ts`'s `forceRerunTriggers`, `server/src/force-rerun-triggers.test.ts`'s coverage list, and `scripts/verify-cache.mjs`'s `test:server` `extraFiles` — each now regression-tested.
- Several factual corrections to the policy doc against the actual sidecar/Node code: an inverted lock invariant (`_drop_model_locked()` vs `unload()`'s hold-the-lock rules), a false claim that both eviction paths symmetrically rethrow aborts (only the leading one does), a wrong rationale for why the lock is released before VRAM reclaim, missing enumeration of two of four real eviction call sites (a voice-prepare pre-pass pair with different, voice-type-dependent fail semantics), and a third pre-pass invocation with its own local fail-soft wrapper that the doc's original wording would have licensed deleting.
- A regression test (`scripts/tests/verify-cache.test.mjs`) that the CI-wiring fix broke (an invariant asserting sidecar-only diffs stay out of `test:server` scope) was updated to reflect the new, deliberate exception rather than reverted.
- Several stale/duplicated-claim code comments corrected (`.github/workflows/verify.yml`, `server/src/force-rerun-triggers.test.ts`, `server/src/tts/synthesise-chapter.ts`).

## Test plan

- [x] `npm run test:server -- coqui-residency-policy.guard` — green
- [x] Mutation check: deleted the `COQUI-RESIDENCY-POLICY` token from `synthesise-chapter.ts`, re-ran the guard — went red as expected; restored and confirmed green again.
- [x] `npm run test:hooks` — full battery green (1600+ tests), including the new `scripts/tests/verify-cache.test.mjs` coverage for the CI-wiring fix.
- [x] `npm run test:server` — full suite green (8000+ tests).
- [x] `npm run typecheck` — clean.
- [x] `npm run verify:fast:branch` — green on every push (pre-push hook).

No release-notes entry: no shippable delta, this is docs, comments, and CI-selection wiring only — no product behaviour change.
No on-box acceptance row: no new behaviour.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -290,18 +290,17 @@ jobs:
       # `--changed <base>` narrows both server configs to diff-affected tests
       # (CI cost round 2). The slow config still serial-runs ONLY the hot files
       # the diff actually touched. Run from server/ so the per-dir vitest
-      # config loads. A sidecar-only PR touches no server TS graph → 0 tests
-      # run here, which is correct (server tests mock the Python sidecar).
-      # See docs/features/118-ci-cost-round-2.md.
+      # config loads. A typical sidecar-only diff (touching only .py files
+      # outside main.py) touches no server TS graph → 0 tests run here, which
+      # is correct (server tests mock the Python sidecar). See
+      # docs/features/118-ci-cost-round-2.md. EXCEPTION: a
+      # server/tts-sidecar/main.py-only diff DOES schedule test:server now (via
+      # extraFiles in verify-cache.mjs + forceRerunTriggers in vitest.config.ts,
+      # since coqui-residency-policy.guard.test.ts reads it at runtime as part
+      # of the Coqui eviction cross-reference guard, #1932).
       # test:server AND test:server-slow share this one step, so the
       # condition is the union of both steps' keys. Drops the old `sidecar`
-      # disjunct deliberately: a typical sidecar-only diff (touching only
-      # .py files outside main.py) used to run this job for zero tests —
-      # but EXCEPTION: a server/tts-sidecar/main.py-only diff DOES schedule
-      # test:server now (via extraFiles in verify-cache.mjs + forceRerunTriggers
-      # in vitest.config.ts, since coqui-residency-policy.guard.test.ts reads
-      # it at runtime as part of the Coqui eviction cross-reference guard,
-      # #1932), so behaviour is mostly behaviour-neutral for the general case.
+      # disjunct deliberately.
       # (was: server || sidecar || shared).
       # leg: test:server
       - name: Server tests (fast + slow)
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -295,10 +295,14 @@ jobs:
       # See docs/features/118-ci-cost-round-2.md.
       # test:server AND test:server-slow share this one step, so the
       # condition is the union of both steps' keys. Drops the old `sidecar`
-      # disjunct deliberately: a sidecar-only diff used to run this job for
-      # zero tests (server tests mock the Python sidecar, per the comment
-      # above), so skipping the job entirely for that diff is behaviour-
-      # neutral, not a narrowing (was: server || sidecar || shared).
+      # disjunct deliberately: a typical sidecar-only diff (touching only
+      # .py files outside main.py) used to run this job for zero tests —
+      # but EXCEPTION: a server/tts-sidecar/main.py-only diff DOES schedule
+      # test:server now (via extraFiles in verify-cache.mjs + forceRerunTriggers
+      # in vitest.config.ts, since coqui-residency-policy.guard.test.ts reads
+      # it at runtime as part of the Coqui eviction cross-reference guard,
+      # #1932), so behaviour is mostly behaviour-neutral for the general case.
+      # (was: server || sidecar || shared).
       # leg: test:server
       - name: Server tests (fast + slow)
         if: fromJSON(needs.detect.outputs.scopes).step_test_server || fromJSON(needs.detect.outputs.scopes).step_test_server_slow || fromJSON(needs.detect.outputs.scopes).shared
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -354,9 +354,9 @@ single statement of that ruling so the question is not reopened.
 
 `evictEngineForPhase` (`server/src/tts/synthesise-chapter.ts:1015`) POSTs
 `/unload` to the sidecar for a named engine. The Coqui-facing wrappers are
-`evictCoquiForQwenPhase` (`:1041`) and the symmetric `evictQwenForCoquiPhase`
-(`:1029`). The leading evict fires at `:2912` before a mixed chapter's Coqui
-phase; the trailing evict fires at `:2976` at its end.
+`evictCoquiForQwenPhase` (`:1049`) and the symmetric `evictQwenForCoquiPhase`
+(`:1029`). The leading evict fires at `:2920` before a mixed chapter's Coqui
+phase; the trailing evict fires at `:2987` at its end.
 
 It is driven by the **generation plan** — Node knows which engine each
 remaining chapter needs, so it can free the VRAM at a boundary where "no
@@ -370,10 +370,10 @@ should not break a whole chapter if a transient sidecar hiccup occurs.
 
 ### Mechanism B — sidecar, demand-driven, reactive
 
-`_idle_evict_steps` (`server/tts-sidecar/main.py:5019-5048`) is a
+`_idle_evict_steps` (`server/tts-sidecar/main.py:4989-5064`) is a
 **five-step, cross-engine ladder** — `spk`, `asr`, `qwen.design`,
 `qwen.base17`, `coqui` — run by `PlacementController` on the admission path
-when an op is VRAM-starved. Coqui is *one step of it* (`:5047`), calling
+when an op is VRAM-starved. Coqui is *one step of it* (`:5063`), calling
 `CoquiEngine.maybe_free_idle` (`:2273`). This is **not** a Coqui-specific
 idle evict: it is one entry in a generic ladder that every engine has a step
 in.
@@ -381,11 +381,11 @@ in.
 Two Coqui-specific qualifications distinguish the Coqui step from the others:
 
 1. **Self-admission skip.** The step is omitted entirely when the admitting
-   op is itself Coqui (`server/tts-sidecar/main.py:5044`, `if engine !=
+   op is itself Coqui (`server/tts-sidecar/main.py:5053`, `if engine !=
    "coqui"`) — evicting would unload the model that op is about to reload.
 2. **Real idle TTL.** It is the only step that uses a non-zero idle TTL —
-   `_coqui_idle_ttl()` (`:8272`, 30 s default, env `COQUI_IDLE_TTL`,
-   registry key `sidecar.coquiIdleTtl` at `server/src/config/registry.ts:849`)
+   `_coqui_idle_ttl()` (`:8288`, 30 s default, env `COQUI_IDLE_TTL`,
+   registry key `sidecar.coquiIdleTtl` at `server/src/config/registry.ts:848`)
    — rather than the `0.0` the transient models (SPK, ASR, Qwen design,
    Qwen base17) take. Coqui deliberately has **no** background watchdog,
    unlike Qwen/ASR/ECAPA, so this only ever runs under genuine VRAM
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -361,10 +361,12 @@ phase; the trailing evict fires at `:2976` at its end.
 It is driven by the **generation plan** — Node knows which engine each
 remaining chapter needs, so it can free the VRAM at a boundary where "no
 further Coqui work is queued" is a *fact* rather than a prediction. Each call
-is bounded by `withCallTimeout` and carries the chapter's abort signal. The
-trailing evict is deliberately **fail-soft** — every group is already
-synthesised by then, so throwing would destroy completed work purely to free
-VRAM. The leading one is not fail-soft.
+is bounded by `withCallTimeout` and carries the chapter's abort signal. Both
+evicts are deliberately **fail-soft** — only abort signals are rethrown; other
+failures (socket errors, timeouts) are logged as warnings and execution
+continues. This is correct for both the leading and trailing positions: the
+trailing evict destroys synthesised work if it throws, and the leading evict
+should not break a whole chapter if a transient sidecar hiccup occurs.
 
 ### Mechanism B — sidecar, demand-driven, reactive
 
@@ -416,11 +418,13 @@ outside the current book's render.
 ### Lock caveats — invariants any future change must preserve
 
 1. **Mechanism A's timeout and abort.** Each `/unload` call is bounded by
-   `withCallTimeout` and carries the chapter's abort signal. The trailing
-   evict must remain fail-soft: every group is already synthesised when it
-   runs, so an exception there destroys completed work purely to free VRAM.
-   The leading evict must remain fail-hard: the Coqui phase has not started
-   yet, so a failure there is a genuine precondition not met.
+   `withCallTimeout` and carries the chapter's abort signal. Both evicts must
+   remain fail-soft (aborting only on genuine abort signals, not transient
+   errors): only abort signals rethrow; other failures are logged as warnings
+   and execution continues. The trailing evict is already-synthesised work
+   that would be lost on exception. The leading evict must not break the
+   chapter if a transient sidecar hiccup (socket error, timeout) occurs
+   mid-evict.
 
 2. **Mechanism B's lock-guarded idle check.** `maybe_free_idle`
    (`server/tts-sidecar/main.py:2273`) re-validates the idle condition under
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -362,11 +362,14 @@ It is driven by the **generation plan** — Node knows which engine each
 remaining chapter needs, so it can free the VRAM at a boundary where "no
 further Coqui work is queued" is a *fact* rather than a prediction. Each call
 is bounded by `withCallTimeout` and carries the chapter's abort signal. Both
-evicts are deliberately **fail-soft** — only abort signals are rethrown; other
-failures (socket errors, timeouts) are logged as warnings and execution
-continues. This is correct for both the leading and trailing positions: the
-trailing evict destroys synthesised work if it throws, and the leading evict
-should not break a whole chapter if a transient sidecar hiccup occurs.
+evicts are deliberately **fail-soft** on non-abort errors: socket errors and
+timeouts are logged as warnings and execution continues. The leading evict
+additionally rethrows abort signals for pause detection (`routes/generation.ts`
+names its own `AbortError`); the trailing evict swallows aborts too, since by
+the time it fires the chapter's synthesis work is already complete. This
+asymmetry is correct: the trailing evict's abort swallow does not lose
+synthesised work (all groups are done), while the leading evict's abort rethrow
+ensures a paused generation is visible to the pause-detection path.
 
 ### Mechanism B — sidecar, demand-driven, reactive
 
@@ -419,12 +422,13 @@ outside the current book's render.
 
 1. **Mechanism A's timeout and abort.** Each `/unload` call is bounded by
    `withCallTimeout` and carries the chapter's abort signal. Both evicts must
-   remain fail-soft (aborting only on genuine abort signals, not transient
-   errors): only abort signals rethrow; other failures are logged as warnings
-   and execution continues. The trailing evict is already-synthesised work
-   that would be lost on exception. The leading evict must not break the
-   chapter if a transient sidecar hiccup (socket error, timeout) occurs
-   mid-evict.
+   remain fail-soft on transient errors (socket errors, timeouts are logged as
+   warnings and execution continues), but they differ on aborts: the leading
+   evict rethrows `AbortError` for pause detection (`routes/generation.ts`
+   names its own `AbortError`), while the trailing evict swallows abort signals
+   too since by that point all synthesis work is done. The trailing evict
+   cannot lose synthesised audio by failing, and the leading evict must signal
+   a pause through its abort path if the generation is cancelled mid-evict.
 
 2. **Mechanism B's lock-guarded idle check.** `maybe_free_idle`
    (`server/tts-sidecar/main.py:2273`) re-validates the idle condition under
@@ -432,14 +436,22 @@ outside the current book's render.
    never free the model out from under an active forward. Any change to
    `maybe_free_idle` must preserve this re-validation-under-lock.
 
-3. **Mechanism B's non-reentrant lock.** `_drop_model_locked()` must not be
+3. **Mechanism B's non-reentrant lock.** `_drop_model_locked()` MUST be
    called while already holding `_synth_lock` (it is a non-reentrant
-   `threading.Lock`) — only the `/unload` route calls `unload()` without
-   the lock. Internal holders that need the same teardown call
-   `_drop_model_locked()` and then `_reclaim_after_drop()` after releasing
-   the lock; `maybe_free_idle` is the one such caller
-   (`server/tts-sidecar/main.py:2051-2055`). Any future caller must follow
-   the same release-then-reclaim sequence.
+   `threading.Lock`); by contrast, the public `unload()` method MUST NOT be
+   called while holding the lock — it acquires the lock itself. Internal
+   callers that need the same teardown (e.g., `maybe_free_idle` at
+   `server/tts-sidecar/main.py:2273`) acquire the lock, call
+   `_drop_model_locked()` inside it, then release the lock and call
+   `_reclaim_after_drop()` afterwards. This sequence is critical: releasing
+   the lock BEFORE reclaim allows a concurrent forward to complete its own
+   reference-drop, which is essential for the actual VRAM reclamation to work.
+   Any future internal caller must follow the same acquire-drop-release-reclaim
+   sequence; any future code that nulls or reassigns a model field must mirror
+   `_drop_model_locked()`'s full teardown (including the device restoration
+   that `_drop_model_locked()` already handles) to avoid the double-allocation
+   race documented in `unload()`'s own docstring
+   (`server/tts-sidecar/main.py:2040-2049`).
 
 ## Ship notes
 
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -354,30 +354,33 @@ single statement of that ruling so the question is not reopened.
 
 `evictEngineForPhase` (`server/src/tts/synthesise-chapter.ts:1015`) POSTs
 `/unload` to the sidecar for a named engine. The Coqui-facing wrappers are
-`evictCoquiForQwenPhase` (`:1049`) and the symmetric `evictQwenForCoquiPhase`
+`evictCoquiForQwenPhase` (`:1050`) and the symmetric `evictQwenForCoquiPhase`
 (`:1029`). 
 
 The eviction wrappers have **four call sites**, not two:
 
 1. **Pre-pass invocations (memoised, resolver-mediated):** `beforeFirstCoquiDerive`
-   (`:1882-1899`) evicts `evictQwenForCoquiPhase` before the first Coqui derive
-   in the chapter; `beforeFirstQwenDerive` (`:1943-1953`) evicts
+   (`:1883-1900`) evicts `evictQwenForCoquiPhase` before the first Coqui derive
+   in the chapter; `beforeFirstQwenDerive` (`:1944-1954`) evicts
    `evictCoquiForQwenPhase` before the first Qwen derive (when Coqui is present).
    **Two of the three pre-pass sites are called from *inside* voice-resolver
    try/catch blocks (`clone-voice-resolver.ts`), inheriting the resolver's own
    fail-loud (cloned voices) / fail-soft (designed voices) semantics — no local
    try/catch wraps them at those call sites.** However, **a third pre-pass site**
-   exists at `:2114-2122`, also in `synthesise-chapter.ts`, which primes
-   `beforeFirstQwenDerive` before any Qwen render-phase work when only Qwen
-   voices with no derive are present (no resolver call to hang the hook off).
-   This third site **is wrapped in a local try/catch** (lines 2114-2122) that
+   exists at `:2115-2123`, also in `synthesise-chapter.ts`, which primes
+   `beforeFirstQwenDerive` before any Qwen render-phase work when Coqui is present
+   AND any Qwen work is present (either groups or derives). This third site shares
+   the memoised promise with the resolver-mediated call sites, so while it fires in
+   more cases than "no derive needed", the resolver calls still apply their own
+   fail-loud (cloned) / fail-soft (designed) policy on top.
+   This third site **is wrapped in a local try/catch** (lines 2115-2123) that
    unconditionally swallows the error and logs a warning — deliberately best-effort
-   fail-soft, covering the "Qwen render with no derive needed" case where a
-   sidecar recycle window at this priming call cannot abort the chapter.
+   fail-soft, covering cases where a sidecar recycle window at this priming call
+   cannot abort the chapter.
 
 2. **Render-phase pair (top-level, plan-driven):** The leading evict fires at
-   `:2920` before a mixed chapter's Coqui phase; the trailing evict fires at
-   `:2988` at its end. Each call is bounded by `withCallTimeout` and carries
+   `:2921` before a mixed chapter's Coqui phase; the trailing evict fires at
+   `:2989` at its end. Each call is bounded by `withCallTimeout` and carries
    the chapter's abort signal.
 
 The render-phase pair is driven by the **generation plan** — Node knows which
@@ -457,18 +460,20 @@ outside the current book's render.
      signal a pause through its abort path if the generation is cancelled
      mid-evict.
    
-   - **Pre-pass pair — two resolver-mediated, one render-prep best-effort:**
-     `beforeFirstCoquiDerive` and `beforeFirstQwenDerive` have **three distinct
-     invocation patterns**, each with different failure semantics that must be
-     preserved independently:
+   - **Pre-pass pair — four resolver-mediated, one render-prep best-effort:**
+     `beforeFirstCoquiDerive` and `beforeFirstQwenDerive` have **five distinct
+     invocation sites across three failure-semantics patterns**, each with different
+     failure semantics that must be preserved independently:
      
-     - **Two resolver-mediated sites (voice-type-dependent):** Called from
+     - **Four resolver-mediated sites (voice-type-dependent):** Called from
        *inside* voice-resolver try/catch blocks in `clone-voice-resolver.ts`,
-       with no separate local try/catch wrapping them. A failure is reported
-       through the resolver's own existing, already-differentiated policy: the
-       cloned resolver's catch treats it as an ordinary derive failure (fail-loud —
-       `broken` accumulates, `UnresolvableClonedVoiceError` still throws), while
-       the designed resolver's catch treats it as an ordinary self-heal failure
+       with no separate local try/catch wrapping them. The cloned resolver calls
+       both hooks (coqui arm at ~line 547, qwen arm at ~line 548); the designed
+       resolver also calls both (coqui arm at ~line 996, qwen arm at ~line 1073).
+       A failure is reported through the resolver's own existing, already-differentiated
+       policy: the cloned resolver's catch treats it as an ordinary derive failure
+       (fail-loud — `broken` accumulates, `UnresolvableClonedVoiceError` still throws),
+       while the designed resolver's catch treats it as an ordinary self-heal failure
        (fail-soft — `softFailedUuids`/keep-stale, never rethrown). This voice-type-dependent
        semantics is intentional: a pre-pass evict failure on a cloned voice is
        unrecoverable (the cloned resolver's own identity depends on that model
@@ -476,14 +481,18 @@ outside the current book's render.
        the stale version.
      
      - **One render-prep best-effort site (unconditional fail-soft):** A third
-       pre-pass invocation at `:2114-2122` in `synthesise-chapter.ts` primes
-       `beforeFirstQwenDerive` before any Qwen render-phase work when only
-       healthy/catalogue Qwen voices are present (no resolver call to hang the
-       hook off). This site **is deliberately wrapped in a local try/catch**
-       that unconditionally swallows errors and logs warnings, covering the case
-       where a sidecar recycle window at this priming call cannot abort the
-       chapter (fix for a §2.3 chapter-abort regression). This best-effort wrapper
-       is distinct from the resolver-mediated invocations and correct by design.
+       pre-pass invocation at `:2115-2123` in `synthesise-chapter.ts` primes
+       `beforeFirstQwenDerive` before any Qwen render-phase work when Coqui is
+       present AND any Qwen work is pending (either groups or derives; no resolver
+       call to hang the hook off for the render-only case). This site shares the
+       memoised promise with the four resolver-mediated calls, so when they later
+       await the same hook, they get the already-settled result and apply their own
+       fail-loud/fail-soft policy on top.
+       This site **is deliberately wrapped in a local try/catch** that unconditionally
+       swallows errors and logs warnings, covering the case where a sidecar recycle
+       window at this priming call cannot abort the chapter (fix for a §2.3
+       chapter-abort regression). This best-effort wrapper is distinct from the
+       resolver-mediated invocations and correct by design.
      
      Any future change to the pre-pass evicts must **preserve all three
      invocation patterns independently**: the two resolver-mediated sites must
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -355,21 +355,37 @@ single statement of that ruling so the question is not reopened.
 `evictEngineForPhase` (`server/src/tts/synthesise-chapter.ts:1015`) POSTs
 `/unload` to the sidecar for a named engine. The Coqui-facing wrappers are
 `evictCoquiForQwenPhase` (`:1049`) and the symmetric `evictQwenForCoquiPhase`
-(`:1029`). The leading evict fires at `:2920` before a mixed chapter's Coqui
-phase; the trailing evict fires at `:2987` at its end.
+(`:1029`). 
 
-It is driven by the **generation plan** — Node knows which engine each
-remaining chapter needs, so it can free the VRAM at a boundary where "no
-further Coqui work is queued" is a *fact* rather than a prediction. Each call
-is bounded by `withCallTimeout` and carries the chapter's abort signal. Both
-evicts are deliberately **fail-soft** on non-abort errors: socket errors and
-timeouts are logged as warnings and execution continues. The leading evict
-additionally rethrows abort signals for pause detection (`routes/generation.ts`
-names its own `AbortError`); the trailing evict swallows aborts too, since by
-the time it fires the chapter's synthesis work is already complete. This
-asymmetry is correct: the trailing evict's abort swallow does not lose
-synthesised work (all groups are done), while the leading evict's abort rethrow
-ensures a paused generation is visible to the pause-detection path.
+The eviction wrappers have **four call sites**, not two:
+
+1. **Pre-pass pair (memoised, demand-driven):** `beforeFirstCoquiDerive`
+   (`:1882-1899`) evicts `evictQwenForCoquiPhase` before the first Coqui derive
+   in the chapter; `beforeFirstQwenDerive` (`:1943-1953`) evicts
+   `evictCoquiForQwenPhase` before the first Qwen derive (when Coqui is present).
+   These are called from *inside* voice-resolver try/catch blocks
+   (`clone-voice-resolver.ts`), so they inherit the resolver's own fail-loud
+   (cloned voices) / fail-soft (designed voices) semantics — **no local
+   try/catch wraps them, by design.**
+
+2. **Render-phase pair (top-level, plan-driven):** The leading evict fires at
+   `:2920` before a mixed chapter's Coqui phase; the trailing evict fires at
+   `:2988` at its end. Each call is bounded by `withCallTimeout` and carries
+   the chapter's abort signal.
+
+The render-phase pair is driven by the **generation plan** — Node knows which
+engine each remaining chapter needs, so it can free the VRAM at a boundary
+where "no further Coqui work is queued" is a *fact* rather than a prediction
+(this scope applies to the **render-phase pair only**, not the pre-pass pair,
+which is memoised and runs *before* any derive). The render-phase evicts are
+deliberately **fail-soft** on non-abort errors: socket errors and timeouts are
+logged as warnings and execution continues. The leading evict additionally
+rethrows abort signals for pause detection (`routes/generation.ts` names its
+own `AbortError`); the trailing evict swallows aborts too, since by the time it
+fires the chapter's synthesis work is already complete. This asymmetry is
+correct: the trailing evict's abort swallow does not lose synthesised work (all
+groups are done), while the leading evict's abort rethrow ensures a paused
+generation is visible to the pause-detection path.
 
 ### Mechanism B — sidecar, demand-driven, reactive
 
@@ -420,15 +436,33 @@ outside the current book's render.
 
 ### Lock caveats — invariants any future change must preserve
 
-1. **Mechanism A's timeout and abort.** Each `/unload` call is bounded by
-   `withCallTimeout` and carries the chapter's abort signal. Both evicts must
-   remain fail-soft on transient errors (socket errors, timeouts are logged as
-   warnings and execution continues), but they differ on aborts: the leading
-   evict rethrows `AbortError` for pause detection (`routes/generation.ts`
-   names its own `AbortError`), while the trailing evict swallows abort signals
-   too since by that point all synthesis work is done. The trailing evict
-   cannot lose synthesised audio by failing, and the leading evict must signal
-   a pause through its abort path if the generation is cancelled mid-evict.
+1. **Mechanism A's failure semantics differ by call site.**
+   
+   - **Render-phase pair (bounded by `withCallTimeout`, fail-soft):** Each
+     `/unload` call carries the chapter's abort signal. Both evicts must remain
+     fail-soft on transient errors (socket errors, timeouts are logged as
+     warnings and execution continues). They differ on aborts: the leading
+     evict rethrows `AbortError` for pause detection (`routes/generation.ts`
+     names its own `AbortError`), while the trailing evict swallows abort
+     signals too since by that point all synthesis work is done. The trailing
+     evict cannot lose synthesised audio by failing, and the leading evict must
+     signal a pause through its abort path if the generation is cancelled
+     mid-evict.
+   
+   - **Pre-pass pair (memoised, resolver-dependent):** `beforeFirstCoquiDerive`
+     and `beforeFirstQwenDerive` are called from *inside* voice-resolver
+     try/catch blocks in `clone-voice-resolver.ts`, with no separate local
+     try/catch wrapping them. A failure is therefore reported through the
+     resolver's own existing, already-differentiated policy: the cloned
+     resolver's catch treats it as an ordinary derive failure (fail-loud —
+     `broken` accumulates, `UnresolvableClonedVoiceError` still throws), while
+     the designed resolver's catch treats it as an ordinary self-heal failure
+     (fail-soft — `softFailedUuids`/keep-stale, never rethrown). This
+     voice-type-dependent semantics is intentional and must be preserved: a
+     pre-pass evict failure on a cloned voice is unrecoverable (the cloned
+     resolver's own identity depends on that model running), while a failure
+     on a designed voice can be healed by keeping the stale version. Any future
+     change to the pre-pass evicts must maintain this asymmetry.
 
 2. **Mechanism B's lock-guarded idle check.** `maybe_free_idle`
    (`server/tts-sidecar/main.py:2273`) re-validates the idle condition under
@@ -444,8 +478,15 @@ outside the current book's render.
    `server/tts-sidecar/main.py:2273`) acquire the lock, call
    `_drop_model_locked()` inside it, then release the lock and call
    `_reclaim_after_drop()` afterwards. This sequence is critical: releasing
-   the lock BEFORE reclaim allows a concurrent forward to complete its own
-   reference-drop, which is essential for the actual VRAM reclamation to work.
+   the lock BEFORE reclaim is necessary for throughput — holding `_synth_lock`
+   across the multi-second `gc.collect()` + `torch.cuda.empty_cache()` in
+   `_reclaim_after_drop()` would unnecessarily block concurrent synth calls that
+   only need the lock briefly, for no correctness benefit (the reference-drop
+   whose safety depends on the lock is guaranteed by ACQUIRING it, not
+   releasing it — `_synth_lock` waits for any in-flight forward to finish and
+   release its own reference before `_drop_model_locked()` runs). See
+   `_reclaim_after_drop()`'s docstring at `server/tts-sidecar/main.py:2127-2134`
+   for the full rationale.
    Any future internal caller must follow the same acquire-drop-release-reclaim
    sequence; any future code that nulls or reassigns a model field must mirror
    `_drop_model_locked()`'s full teardown (including the device restoration
--- a/docs/features/264-vram-aware-gpu-placement.md
+++ b/docs/features/264-vram-aware-gpu-placement.md
@@ -340,6 +340,103 @@ cold-load, Coqui load-steer) are tracked together in **#1730**.
 > acceptance (with `SEG_CAPACITY_ADMISSION=1`) still owed before the
 > concurrent-multi-card flag flip.
 
+## Coqui residency policy (side-18)
+
+<!-- COQUI-RESIDENCY-POLICY -->
+
+Two independent mechanisms free VRAM held by a resident Coqui XTTS model (~3.5 GB).
+Both are correct, both are tested, and neither knows the other exists. The fork
+#1932 offered — collapse to one, or keep both with explicit cross-references —
+is **closed: both are kept**, for the two reasons below. This section is the
+single statement of that ruling so the question is not reopened.
+
+### Mechanism A — Node, plan-driven, proactive
+
+`evictEngineForPhase` (`server/src/tts/synthesise-chapter.ts:1015`) POSTs
+`/unload` to the sidecar for a named engine. The Coqui-facing wrappers are
+`evictCoquiForQwenPhase` (`:1041`) and the symmetric `evictQwenForCoquiPhase`
+(`:1029`). The leading evict fires at `:2912` before a mixed chapter's Coqui
+phase; the trailing evict fires at `:2976` at its end.
+
+It is driven by the **generation plan** — Node knows which engine each
+remaining chapter needs, so it can free the VRAM at a boundary where "no
+further Coqui work is queued" is a *fact* rather than a prediction. Each call
+is bounded by `withCallTimeout` and carries the chapter's abort signal. The
+trailing evict is deliberately **fail-soft** — every group is already
+synthesised by then, so throwing would destroy completed work purely to free
+VRAM. The leading one is not fail-soft.
+
+### Mechanism B — sidecar, demand-driven, reactive
+
+`_idle_evict_steps` (`server/tts-sidecar/main.py:5019-5048`) is a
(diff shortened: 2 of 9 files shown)


dudarenok-maker:
## PR review — pass 1 (head ff7ae2b4cb85041665992ed213ca29cbdeb4a397, depth low)

Scope: all 4 changed files (`docs/features/264-vram-aware-gpu-placement.md`, `server/src/tts/coqui-residency-policy.guard.test.ts`, `server/src/tts/synthesise-chapter.ts`, `server/tts-sidecar/main.py`) · verified: every factual claim in the new policy section re-derived from the files at this head SHA (`git show ff7ae2b4:<path>`), plus `origin/main` for the same paths to establish what this PR's own commits shifted; `server/vitest.config.ts` + `vitest.config.slow.ts` checked for include/exclude coverage of the new guard; `server/src/config/registry.ts` checked for the cited knob; issue states for #1932 / #2689–#2692 checked. No files modified.

Depth `low` is right for the commit types, but a `docs` PR whose entire deliverable *is* the accuracy of a set of pointers gets its pointers checked one by one rather than waved through — that is where all three findings below came from.

---

### 🔴 Blocking — the policy inverts the leading evict's failure policy, and pins the inversion as an invariant future changes "must preserve" (correctness bug)

`docs/features/264-vram-aware-gpu-placement.md`, **Mechanism A**:

> The trailing evict is deliberately **fail-soft** … **The leading one is not fail-soft.**

and **Lock caveats → 1**:

> The leading evict **must remain fail-hard**: the Coqui phase has not started yet, so a failure there is a genuine precondition not met.

Both are false at this head. `server/src/tts/synthesise-chapter.ts:2919-2942`:

```ts
    try {
      await withCallTimeout('qwen-evict-for-coqui', (sig) => evictQwenForCoquiPhase(sig));
    } catch (err) {
      /* … */
      const name = (err as { name?: string } | null)?.name;
      if (name === 'AbortError' || signal?.aborted) {
        throw name === 'AbortError' ? err : (signal?.reason ?? …);
      }
      console.warn(
        `[synthesise-chapter] fs-60 Qwen→Coqui evict failed; continuing into the Coqui phase ` +
          `without it (Coqui may load alongside a still-resident Qwen):`,
        err,
      );
    }
```

Only an **abort** is rethrown. Every other failure — the 502/`ECONNREFUSED` sidecar-recycle case, the non-ok `/unload` that `evictEngineForPhase` (`:1024-1026`) throws on, undici's bare `TypeError` — is swallowed and the chapter continues. The leading evict is fail-soft.

Two comments in the same file say so explicitly, and both are the authority the doc should have read:

- `synthesise-chapter.ts:2889-2895` — *"#1893 — the evict is VRAM hygiene, **not a correctness gate**, so a failed one **must not abort a chapter** that would otherwise render. Two gaps were open here: `evictQwenForCoquiPhase` throws on a non-ok `/unload` and `fetch` rejects on a dead socket … so a transient sidecar hiccup killed the chapter. Failing soft costs little."*
- `synthesise-chapter.ts:1002-1004` (`evictEngineForPhase` header) — *"the resulting bare TypeError is NOT an abort — so **the catch below fails soft and warns**, leaving the engine RESIDENT while the next one loads on top of it."*

**Concrete failure scenario.** A future change reads lock caveat 1 as written and "restores" fail-hard by removing the `try`/`catch` at `:2919`. Sidecar is mid-recycle when a mixed-engine chapter reaches the phase boundary → `POST /unload` returns 502 → `evictEngineForPhase` throws at `:1025` → nothing between there and `synthesiseChapter`'s head catches it → the whole chapter fails, with every pre-evict group already synthesised and discarded. That is precisely the regression **#1893** closed. The doc is the "single statement of that ruling so the question is not reopened", so an inverted invariant here is worse than no statement.

The asymmetry that *is* real, and that the caveat should pin instead: the **leading** catch rethrows an `AbortError` so a pause/cancel propagates (`:2933-2937`); the **trailing** catch (`:2985-2991`) swallows *everything*, aborts included. Both are fail-soft on ordinary `/unload` failures; they differ on abort.

---

### 🟠 Significant — the stale in-code comment the doc copied from (correctness bug, in the instrument)

`server/src/tts/synthesise-chapter.ts:2970` and `:2975-2976`:

```
       Fail-SOFT, deliberately and unlike the leading evict: every group is
       …
       leading evict's own bare `await` is untouched here; it is a separate
       finding with its own reconciliation against main. */
```

Both clauses are false at head, and the file contradicts itself six lines later at `:2978-2980`: *"matching the leading `qwen-evict-for-coqui` call above. The fail-soft policy here was already right"*. The leading evict is no longer a bare `await` (it is `withCallTimeout` inside a `try`/`catch` at `:2919-2920`), and it is not "unlike" the trailing one on fail-softness.

This is the upstream source of the 🔴 above — the doc faithfully transcribed a comment that #1893's reconciliation falsified. This PR touched this file and this exact region and repeated the claim, so per CLAUDE.md **Incidental findings → report, fix, record** it is a chore this work made owed: fixed in this round, not filed. Fixing it without also fixing the doc leaves the doc as the surviving copy of the wrong claim.

---

### 🟠 Significant — 7 of the doc's line citations are stale *at this head*, falsified by this PR's own commits (correctness bug)

The citations in the new section were accurate against `origin/main` when `92d7c579` (docs, step 1/3) landed. `ae565401` (step 2/3) then inserted 8 lines into `synthesise-chapter.ts` at `:1042` and 9 + 7 lines into `main.py` at `:2292` / `:5056`, shifting the very symbols the doc points at. Verified symbol-by-symbol at `ff7ae2b4`:

| Doc citation | Actual at head |
|---|---|
| `synthesise-chapter.ts:1041` — `evictCoquiForQwenPhase` | **1049** |
| `:2912` — leading evict | **2920** |
| `:2976` — trailing evict | **2984** |
| `main.py:5019-5048` — `_idle_evict_steps` ladder body | **5028-5064** (`def` at 4989) |
| `:5047` — the `coqui` `EvictStep` append | **5063** |
| `:5044` — `if engine != "coqui"` | **5053** |
| `:8272` — `_coqui_idle_ttl()` | **8288** |
| `registry.ts:849` — key `sidecar.coquiIdleTtl` | key is at **848**; 849 is the `env:` line |

Still correct at head: `synthesise-chapter.ts:1015` (`evictEngineForPhase`), `:1029` (`evictQwenForCoquiPhase`), `main.py:2273` (`maybe_free_idle`), `main.py:2051-2055` (the `unload()` docstring paragraph lock caveat 3 paraphrases — lands exactly).

**Concrete failure scenario.** The next reader opens `main.py:5019` expecting `_idle_evict_steps` and lands mid-docstring in the #1976 stranded-cache paragraph; opens `:8272` for `_coqui_idle_ttl()` and lands on `_COQUI_IDLE_TTL_DEFAULT`'s neighbourhood 16 lines early. This is the exact rot the PR exists to prevent, shipped inside the artifact meant to prevent it — and the guard is deliberately built not to see it (`coqui-residency-policy.guard.test.ts:11-13`: *"Line numbers are asserted NOWHERE here on purpose: they rot"*). That reasoning is sound for the *guard*; it does not license shipping citations already known-rotted at the head being merged. Either re-derive the numbers against `ff7ae2b4` (they are stable from here, since the insertions are done), or drop line numbers from the doc and cite symbol names only — the doc already names every symbol, so the numbers are carrying no load they can't lose.

---

### 🟡 Minor

1. **`server/src/tts/coqui-residency-policy.guard.test.ts:8`** — the header says *"it greps **four** files for one token and one heading."* It reads **three** files (`synthesise-chapter.ts`, `main.py`, the policy doc); there are four *markers* across them (main.py holds two). Same off-by-one as the step-3/3 commit subject ("the four Coqui residency cross-references" is right; "four files" is not). Cleanup nit — a guard whose own self-description miscounts its inputs is the first thing a reader distrusts.

2. **Guard blind spot on doc relocation** — `POLICY_DOC_PATH` (`:24`) hardcodes `docs/features/264-vram-aware-gpu-placement.md`, whose frontmatter is `status: active`. Before-shipping checklist step 8 mandates `git mv` to `docs/features/archive/` when it flips to `stable`. At that moment test 3 throws `ENOENT` (fails closed — correct), but tests 1 and 2 stay **green** while both code-site comments now cite a path that does not exist. Same gap for a plain rename. One cheap line closes it: assert each code-site source also `.includes(POLICY_DOC_REL_PATH)` — the comments already spell the path out verbatim, so it costs one constant and catches the two-thirds of the guard that currently can't see a move. Cleanup nit; the guard is still a real guard without it.

---

### ✅ What is solid

- **Every substantive policy claim re-derives correctly at head.** The five-step cross-engine ladder and its order (`spk`, `asr`, `qwen.design`, `qwen.base17`, `coqui`) — `main.py:5026-5063`. Coqui as the only step with a non-zero TTL, every other passing literal `0.0` — same range. The self-admission skip `if engine != "coqui"` — `:5053`. `_COQUI_IDLE_TTL_DEFAULT = 30.0`, env `COQUI_IDLE_TTL`, 5 s floor — `:8285-8297`; `registry.ts:847-856` agrees (`default: 30`, `apply: 'restart-sidecar'`). "No background watchdog, unlike Qwen/ASR/ECAPA" — corroborated by `main.py:2277` and by startup wiring `:875-879`, which starts design/asr/spk/memory watchdogs and no Coqui one.
- **Lock caveats 2 and 3 are accurate**, including the harder one: `maybe_free_idle` re-validating under `_synth_lock` and the non-reentrant `_drop_model_locked()` → release → `_reclaim_after_drop()` sequence, with the `main.py:2051-2055` citation landing exactly on the `unload()` docstring paragraph that states the convention.
- **The guard is a real guard, not vacuous.** Three genuine `readFileSync` reads of committed files, and the `main.py` assertion is `toBe(2)` on an exact occurrence count rather than a presence check — so deleting *one* of the two markers fails, which a `.includes()` would have missed. Failure messages name the file, the missing thing, and the remedy. Path resolution fails **closed**: a wrong cwd throws `ENOENT` rather than silently passing.
- **The guard actually runs.** `server/vitest.config.ts:130` `include: ['src/**/*.{test,spec}.ts']` matches it, and it is absent from `SLOW_FILES_TO_EXCLUDE`, so `npm run test:server` picks it up — and `verify.yml`'s server leg is in scope for this diff.
- **Bookkeeping is clean.** `Closes #1932` is plain text outside any code span; #1932 is open; the four chain children #2689–#2692 are already closed. Release-notes and on-box-acceptance are declared not-applicable explicitly rather than silently omitted, and both calls are correct (no shippable delta, no new behaviour).
- **The "keep both" reasoning holds.** "Removing B deletes one step from a ladder that stays" and "B cannot substitute for A — it fires only after an admission is already starved, and only past a 30 s TTL" are both borne out by the code; the ~90 s XTTS reload cost the ladder's own docstring cites makes the ordering argument concrete.

---

### Verdict

**Changes requested.** One 🔴 and two 🟠 correctness findings, all in the artifact this PR exists to produce.

The 🔴 is the one that matters: the section is explicitly framed as the ruling of record *"so the question is not reopened"*, and one of its three "invariants any future change must preserve" instructs the next changer to undo #1893. Fix that together with the 🟠 stale comment at `synthesise-chapter.ts:2970-2976` — they are one defect with two copies, and fixing either alone leaves the other as the surviving wrong statement.

The 🟠 line-citation rot is self-inflicted by commit ordering (docs written against `main`, code inserted after) and is worth a moment's thought beyond the fix: a doc that pins line numbers in the same PR that shifts them will do this again. Symbol-only citations would make the section immune, and the doc already names every symbol.

Re-review triggered once these are fixed and pushed (≥1 correctness bug).

dudarenok-maker:
## PR review — pass 2 (head 3e1cd0ff45e70b30d9ddf9220ed03741143922e9, depth high)

Scope: full PR diff (4 files, +207/-7) plus the CI wiring the new guard depends on — `server/vitest.config.ts`, `scripts/verify-cache.mjs`, `scripts/ci-scope.mjs`, `.github/workflows/verify.yml`, `server/src/force-rerun-triggers.test.ts`. Depth re-derived from the current commit set: the branch now carries `fix(docs,server)` (multi-scope), so `low` → `high`.

Every line citation in the new doc section was independently re-derived against the files at this head — all thirteen, not just the seven commit `3e1cd0ff` claims to have moved.

- **Pass 1 🔴 (docs + code comment claim the leading evict is fail-hard)** — **RESOLVED IN PART.** The false "fail-hard" claim is gone from both surfaces. The replacement text introduces a new, narrower inaccuracy in its place — see 🟠 New-3.
- **Pass 1 🟠 (same claim duplicated in `synthesise-chapter.ts`)** — **VERIFIED RESOLVED.** The `"The leading evict's own bare `await` is untouched here"` sentence is gone; the leading evict does have its own `try`/`catch` at `:2921-2943`.
- **Pass 1 🟠 (seven stale line citations)** — **VERIFIED RESOLVED.** Checked by reading the files, not by trusting the commit's old→new mapping. `synthesise-chapter.ts:1015` = `evictEngineForPhase`, `:1029` = `evictQwenForCoquiPhase`, `:1049` = `evictCoquiForQwenPhase`, `:2920` = leading evict, `:2987` = trailing evict. `main.py:4989-5064` = `_idle_evict_steps` (def → `return steps`), `:5063` = the `coqui` `EvictStep`, `:5053` = `if engine != "coqui":`, `:8288` = `_coqui_idle_ttl` (30 s default, env `COQUI_IDLE_TTL`), `:2273` = `maybe_free_idle`, `:2051-2055` = the `unload()` docstring's release-then-reclaim rule. `registry.ts:848` = `key: 'sidecar.coquiIdleTtl'`. The five-step ladder membership (`spk`, `asr`, `qwen.design`, `qwen.base17`, `coqui`) and the "only non-zero TTL" claim both check out against `main.py:5027-5063`.
- **Pass 1 🟡 (header says "greps four files")** — **VERIFIED RESOLVED.** Now "greps three files for one token and one heading"; the test does exactly three `readFileSync` calls.
- **Pass 1 🟡 (`POLICY_DOC_PATH` archival note)** — **VERIFIED RESOLVED.** `coqui-residency-policy.guard.test.ts:24-28` is a comment, no new mechanism, and correctly forward-looking (plan 264 is `status: active`).

---

### 🔴 Blocking — the guard cannot fire in the required CI check on any of the three diffs it exists to catch

`server/src/tts/coqui-residency-policy.guard.test.ts` reads all three of its inputs with `readFileSync` at runtime and imports none of them, so it has **no module-graph edge to anything it guards**. Nothing wires around that. Per input, against `verify.yml` — the required, enforcing gate:

**`server/tts-sidecar/main.py`** (2 of the 4 cross-references). `scripts/ci-scope.mjs:13,39-43` derives `step_test_server` from `verify-cache.mjs`'s `STEPS[]` inputs. Those are `scripts/verify-cache.mjs:378-425`: globs `server/src/**`, `scripts/**/*.{mjs,cjs,js}`, `server/tts-sidecar/scripts/**/*.mjs`, `server/tts-sidecar/requirements/**`, plus a ten-entry `extraFiles` list. `server/tts-sidecar/main.py` matches **none** of them → `step_test_server` false → `.github/workflows/verify.yml:304` gates the whole server leg off. Delete both `COQUI-RESIDENCY-POLICY` comments from `main.py`, push: the leg never runs, required check **green**.

**`docs/features/264-vram-aware-gpu-placement.md`.** Docs-only diff → same scope condition false → server leg skipped; and `.husky/pre-push` skips `verify:fast:branch` outright for a docs-only push (`scripts/is-docs-only-push.mjs`). Delete the entire `## Coqui residency policy (side-18)` section, push: **green** locally and in CI.

**`server/src/tts/synthesise-chapter.ts`.** Here the leg does run — but `verify.yml:310` is `npx vitest run --changed "$BASE"`, and `server/vitest.config.ts`'s `forceRerunTriggers` list contains no entry matching this file (nor `main.py`, nor `docs/**`). With no import edge, `--changed` does not select the guard. Delete the token, push: required check **green**. (The local `verify:fast:branch` would catch this one, because `server/src/**` is in `test:server`'s globs and verify-cache runs the suite unfiltered — but CLAUDE.md is explicit that the local branch-scoped run is not the enforcement gate, and it is skipped entirely on a docs-only push.)

This is the exact class `server/vitest.config.ts`'s own `forceRerunTriggers` comment block documents — *"CI's required check ran and reported green while never re-executing the very guard the diff could have broken"* — and that this repo has already closed four separate times: `openapi.yaml`, `scripts/**`, `.gitattributes`, `server/tts-sidecar/requirements/**`. `server/src/force-rerun-triggers.test.ts:98-150` is a hardcoded allow/deny list, so nothing existing catches a newly-added runtime-read guard; that is why this slipped.

The PR body's mutation check — deleted the token from `synthesise-chapter.ts`, guard went red — proves the **assertion** works. It does not prove the **selection** works, and selection is the half that is broken.

Fix is two wirings, both required (one makes the leg run at all, the other makes vitest select the test within it):
1. `server/vitest.config.ts` `forceRerunTriggers` += entries for `server/src/tts/synthesise-chapter.ts`, `server/tts-sidecar/main.py`, and `docs/features/264-vram-aware-gpu-placement.md`, in the two-alternative brace shape that file mandates — plus their `MAIN_COVERED` rows in `server/src/force-rerun-triggers.test.ts` (every prior fix in this class added its own).
2. `scripts/verify-cache.mjs` `test:server` `extraFiles` += `server/tts-sidecar/main.py` and `docs/features/264-vram-aware-gpu-placement.md`, so `ci-scope.mjs` marks `step_test_server` true for those diffs.

### 🔴 Blocking — lock caveat 3 inverts the rule it states, in the "invariants any future change must preserve" list

`docs/features/264-vram-aware-gpu-placement.md:435-436`:

> `_drop_model_locked()` must not be called while already holding `_synth_lock` (it is a non-reentrant `threading.Lock`)

The source says the opposite. `server/tts-sidecar/main.py:2087-2088` — *"Reset every field the loaded model owns. **CALLER MUST HOLD `_synth_lock`**"* — and `maybe_free_idle` calls it at `:2316`, **inside** the `with self._synth_lock:` block opened at `:2311`. The method that must not be called under the lock is `unload()`, which is what the docstring the caveat cites (`:2051-2055`) is actually about.

The caveat compounds it at `:438-440`: the source reads *"calls `_drop_model_locked()` and then `_reclaim_after_drop()` **once it has released the lock**"* — the release clause attaching to the reclaim. The doc restates it as *"call `_drop_model_locked()` and then `_reclaim_after_drop()` **after releasing the lock**"*, which reads as applying to both calls and so agrees with the inverted first sentence rather than contradicting it.

Concrete break: someone follows this stated invariant and moves `_drop_model_locked()` out of `maybe_free_idle`'s `with self._synth_lock:` block. That reopens exactly the race `main.py:2040-2049` documents — a racing `/unload` nulls `_tts` mid-forward, the in-flight thread holds the old model past the null so `_reclaim_after_drop`'s `gc.collect()` + `empty_cache()` reclaims nothing, and the next idempotent `/load` (seeing `self._tts is None`) allocates a **second** XTTS copy. On the 8 GB card this plan is written for, that is an OOM.

Weight: this section exists to be "the single statement of that ruling so the question is not reopened," and this sentence sits under a heading promising invariants. A reader has no reason to re-derive it.

### 🟠 Significant — the fail-hard fix over-corrected: the two evicts are not symmetric on the abort path

Three surfaces now assert a symmetry that does not exist:

- `docs/features/264-vram-aware-gpu-placement.md:364-366` — "Both evicts are deliberately **fail-soft** — only abort signals are rethrown"
- same doc `:421-423` (lock caveat 1) — "Both evicts must remain fail-soft (aborting only on genuine abort signals, not transient errors): only abort signals rethrow"
- `server/src/tts/synthesise-chapter.ts:2974-2976` (added by `0e904270`) — "Both evicts catch non-abort errors and log warnings"

The leading evict's catch (`:2921-2943`) has an explicit abort branch that rethrows, with a long comment at `:2922-2931` explaining why. The trailing evict's catch (`:2988-2993`) has **no** abort branch — it `console.warn`s every error, `AbortError` included.

Repro: pause a mixed-engine chapter while the trailing `/unload` is in flight. `withCallTimeout` (`:2482-2487`) forwards the parent abort to the inner controller; undici rejects with the signal's `AbortError`; `:2989` logs a warning; `:2995` returns `out`; `synthesiseChapter` **resolves successfully**, and `routes/generation.ts`'s `err.name === 'AbortError'` pause detector — the thing `:2922-2931` says the leading branch exists to feed — never sees it. Same pause during the leading evict propagates as a pause.

The behavioural consequence is mild (the chapter genuinely is complete, and the next chapter's `withCallTimeout` sees `signal.aborted` immediately), which is why this is 🟠 not 🔴 — but the doc states it as an invariant "any future change must preserve," and it is not true today. Minimal fix is the doc + comment. Adding the abort branch to the trailing catch instead is a behaviour change with two defensible answers and would be a design pass, not a correction.

### 🟡 Minor — a fix commit added a hardcoded line-range citation *into code*

`server/src/tts/synthesise-chapter.ts:2970` — "see lines 2919-2943 above". Accurate at this head, and it is the precise artifact this PR argues against: `coqui-residency-policy.guard.test.ts:11-13` states "Line numbers are asserted NOWHERE here on purpose: they rot," and the sibling commit `3e1cd0ff` spent its whole diff repairing rotted line citations. Nothing guards this one. Name the construct instead — the very next comment (`:2981-2982`) already does it correctly: "the leading `qwen-evict-for-coqui` call above."

### 🟡 Minor — the two code-site assertions guard at different strengths for no stated reason

`coqui-residency-policy.guard.test.ts:39` asserts `source.includes(TOKEN)` for `synthesise-chapter.ts`; `:49` asserts an exact `occurrences(source, TOKEN) === 2` for `main.py`. The token appears exactly once in `synthesise-chapter.ts` (`:1042`), so the exact count is equally available. As written, an edit that strips the comment from the `evictCoquiForQwenPhase` site but leaves the string anywhere else in the 3000-line file — including in a comment recording its removal — still passes.

### Verdict

**Not mergeable as-is.** Four of pass 1's five findings are verified resolved and the citation re-derivation is genuinely correct — I checked all thirteen citations against the files rather than the commit's claims. But the round leaves two blockers.

The first is the one that matters most: this PR's whole thesis is that a comment alone rots and a test must catch it, and the test it ships is unreachable in the required check for all three of the diffs it guards. `main.py` and the policy doc do not even schedule the server leg; `synthesise-chapter.ts` schedules it but `--changed` never selects the guard. As shipped, all four cross-references can be deleted with CI green. The second is a threading invariant stated backwards inside a section headed "invariants any future change must preserve," where following it re-opens a documented double-allocation race.

The 🟠 is the previous round's fix swinging one notch past the truth — the classic shape where round N+1's finding is round N's unaudited fix. Worth folding in the same round as the two blockers.

dudarenok-maker:
## PR review — pass 3 (head fd0f8286acbcb2cd3e1f15eae3a043fd34d25c2a, depth high)

Depth unchanged (`high`): the commit set still carries `fix(docs,server)` — multi-scope.

Scope: all five pass-2 findings re-derived from the files at this head (not from commit messages), plus a fresh sweep of the whole diff and the CI wiring it now depends on — `server/vitest.config.ts`, `scripts/verify-cache.mjs`, `scripts/ci-scope.mjs`, `scripts/tests/verify-cache.test.mjs`, `server/src/force-rerun-triggers.test.ts`, `.github/workflows/verify.yml`.

- **Pass 2 🔴 (the guard cannot fire in the required check on any of its three inputs)** — **VERIFIED RESOLVED.** Re-derived end to end rather than trusting `e97decb4`'s claim:
  - *Leg scheduling.* `scripts/verify-cache.mjs:431-432` (comment at `:425-430`) adds `server/tts-sidecar/main.py` and `docs/features/264-vram-aware-gpu-placement.md` to `test:server`'s `extraFiles`. `stepTouchedByDiff` (`scripts/verify-cache.mjs:811-827`) matches `extraFiles` by **exact POSIX path**, so both literals hit; `synthesise-chapter.ts` was already inside the `server/src/**` glob. `ci-scope.mjs:39-43` derives `step_test_server` straight off that predicate, and `.github/workflows/verify.yml:304-305` gates the leg on `step_test_server || step_test_server_slow || shared`. All three diffs now schedule the leg.
  - *Test selection within the leg.* `verify.yml:310` is `npx vitest run --changed "$BASE"`. `server/vitest.config.ts:229-231` adds all three files in the two-alternative brace shape the file mandates. I checked the glob shape against the three root shapes `force-rerun-triggers.test.ts` asserts (clean, dot-prefixed `.claude/worktrees/…`, real checkout): the second alternative `**/.*/**/…` is what carries the dot-prefixed case, matching the existing `.gitattributes`/`requirements/**` entries exactly.
  - *Pinned.* `server/src/force-rerun-triggers.test.ts:136-152` adds all three as `MAIN_COVERED` rows, each with the right `base` (`SERVER_ROOT` for `src/tts/synthesise-chapter.ts`, `REPO_ROOT` for the other two) — `absPathUnder`'s `existsSync` precondition passes for all three, and none of them collides with `NOT_COVERED`'s `src/index.ts` "ordinary server source file" case.
  - *The regression `e97decb4` broke was correctly narrowed, not weakened.* `scripts/tests/verify-cache.test.mjs:418-423` keeps both original negative assertions (`test` false, `test:hooks` false) and flips only `test:server` to `true`, with the reason in the test name. That is a strictly stronger test than before: it now also pins the new `extraFiles` entry. Confirmed no other suite asserted the old invariant — `scripts/tests/ci-scope.test.mjs:72-75` uses `README.md`, not a `docs/features/**` path, so it is unaffected.

- **Pass 2 🔴 (lock caveat 3 inverts the rule it states)** — **RESOLVED IN PART.** The inversion itself is fixed and now correct: `docs/features/264-vram-aware-gpu-placement.md:439-443` matches `main.py:2086-2088` (`_drop_model_locked` — "CALLER MUST HOLD `_synth_lock`"), `main.py:2051-2055` (`unload()` — "MUST NOT be called while already holding `_synth_lock`"), and `maybe_free_idle`'s actual acquire→drop→release→reclaim sequence at `main.py:2311-2319`. The re-pointed citation `main.py:2040-2049` is accurate (the double-allocation-race paragraph). **But the replacement sentence added a new false claim in its place — see 🟠 New-1.**

- **Pass 2 🟠 (the fail-hard fix over-corrected into a false abort symmetry)** — **VERIFIED RESOLVED.** All three surfaces now state the real asymmetry. `synthesise-chapter.ts:2970-2980` reads "Fail-SOFT on ALL errors (including aborts), unlike the leading evict"; the doc at `:364-371` and lock caveat 1 at `:423-430` say the same. Checked against the code: the leading catch (`synthesise-chapter.ts:2921-2943`) has the explicit `name === 'AbortError' || signal?.aborted` rethrow branch; the trailing catch (`:2989-2995`) is a bare `console.warn` with no abort branch.

- **Pass 2 🟡 (a fix commit added a hardcoded line-range citation into code)** — **VERIFIED RESOLVED.** "see lines 2919-2943 above" is gone; `synthesise-chapter.ts:2982-2983` now names the construct ("matching the leading `qwen-evict-for-coqui` call above"). A tree-wide scan of the file for line-number citations turns up only one, `:1456`'s `~:1119`, which is pre-existing and untouched by this PR.

- **Pass 2 🟡 (the two code-site assertions guard at different strengths)** — **VERIFIED RESOLVED.** `coqui-residency-policy.guard.test.ts:38-43` is now `occurrences(source, TOKEN) === 1`, matching `main.py`'s `=== 2` at `:48-54`. Verified the actual counts at this head: 1 in `synthesise-chapter.ts`, 2 in `main.py`, 1 in the doc — the assertions are satisfiable and exact.

---

### 🟠 New-1 — correctness — caveat 3's replacement rationale is false, and the code says so in as many words

`docs/features/264-vram-aware-gpu-placement.md:446-448`:

> This sequence is critical: releasing the lock BEFORE reclaim allows a concurrent forward to complete its own reference-drop, which is essential for the actual VRAM reclamation to work.

The rule (release, then reclaim) is right. The stated reason is not, and the source contradicts it explicitly.

`main.py:2127-2134` — `_reclaim_after_drop`'s own docstring:

> MUST run with `_synth_lock` RELEASED — **not** to protect the event loop (…); it's released because holding `_synth_lock` across this multi-second reclaim would block concurrent synth calls that only need the lock briefly, **for no benefit**.

That is a liveness reason, not a correctness one. There is also no "concurrent forward" left to complete a reference-drop at that point, on either path that reaches it: `unload()` (`main.py:2048-2049`) takes `_synth_lock` precisely so it "wait[s] for any in-flight forward to finish and drop its own reference **first**", and `maybe_free_idle` re-validates `self._in_flight.busy` under the lock (`main.py:2312`) before calling `_drop_model_locked()`. The reference-drop the doc credits to *releasing* the lock is the thing *acquiring* it already guaranteed — the sentence has imported `unload()`'s acquire-side rationale onto the release step.

Supporting evidence for what `_reclaim_after_drop` actually is: `_ensure_loaded`'s discard branch (`main.py:1979-1988`) **skips it entirely** when `lock_held` is true and calls that "safe rather than merely convenient". (That branch also documents itself as unreachable, so it is not on its own proof that skipping is generally harmless — but note what it skips. The reference drop there is the preceding `del tts` at `main.py:1978`; `_reclaim_after_drop` only adds `gc.collect()` + `empty_cache()`, i.e. allocator hygiene that decides *when* freed VRAM becomes visible to other processes, not *whether* the reference is released.) Nothing in the file ties releasing the lock to a forward completing a reference drop.

Concrete failure scenario: a future changer reading caveat 3 believes reclaiming under the lock *loses VRAM*. They then either (a) leave a genuine multi-second lock hold in place elsewhere because they think the release is load-bearing for correctness rather than for throughput, or (b) — having satisfied themselves that no forward is in flight (which is always true here) — conclude the stated precondition is vacuous and move the reclaim inside the lock, paying a multi-second stall on every concurrent synth for a reason the doc never named. Neither reader can get the real constraint from this text.

This sits under a heading that reads "invariants any future change must preserve," in a section whose stated purpose is to be "the single statement of that ruling so the question is not reopened" — the same position, and the same round-N-fix-becomes-round-N+1-finding shape, as the two prior findings on this list.

### 🟠 New-2 — correctness — Mechanism A enumerates two of its four `/unload` call sites, and caveat 1's "Both evicts" invariant is false of the two it omits

`docs/features/264-vram-aware-gpu-placement.md:355-359` names the wrappers and then says:

> The leading evict fires at `:2920` before a mixed chapter's Coqui phase; the trailing evict fires at `:2987` at its end.

and `:423-425` (lock caveat 1) generalises:

> **Both evicts** must remain fail-soft on transient errors (socket errors, timeouts are logged as warnings and execution continues)…

There are **four** call sites of those two wrappers in `synthesise-chapter.ts`, not two. Besides the render-phase pair at `:2920` and `:2988`, the voice-prepare pre-pass has its own memoised pair:

- `synthesise-chapter.ts:1894-1896` — `withCallTimeout('qwen-evict-for-coqui-prepass', …)` inside `beforeFirstCoquiDerive`
- `synthesise-chapter.ts:1948-1950` — `withCallTimeout('coqui-evict-for-qwen-prepass', …)` inside `beforeFirstQwenDerive`

Neither has a local `try`/`catch`, deliberately. `synthesise-chapter.ts:1845-1853` spells out why, and the policy there is **not** uniform fail-soft:

> the cloned resolver's catch treats it as an ordinary derive failure (**fail-loud** — `broken` accumulates, `UnresolvableClonedVoiceError` still throws…); the designed resolver's coqui-arm catch treats it as an ordinary self-heal failure (fail-soft…)

So a transient socket error on `/unload` at `:1894` **does** fail the chapter when a cloned Coqui voice is present — the exact opposite of the invariant caveat 1 states for "both evicts", and it is that way on purpose.

Concrete failure scenario: someone auditing Mechanism A against caveat 1 finds two call sites that violate the stated invariant and "restores" it by wrapping `beforeFirstCoquiDerive`'s `withCallTimeout` in a warn-and-continue `catch`. That silently defeats the cloned resolver's fail-loud contract: a chapter with an underivable Coqui clone now renders with a stale/wrong voice instead of raising `UnresolvableClonedVoiceError`, which `synthesise-chapter.ts:1847-1849` records as an explicit acceptance criterion ("cloned present + evict fails still raises").

Two secondary consequences of the same omission, worth fixing in the same edit:
- The `COQUI-RESIDENCY-POLICY` marker is attached to `evictCoquiForQwenPhase`'s **definition** (`synthesise-chapter.ts:1042-1048`), so it serves all four call sites — a reader arriving from any of them lands on an enumeration missing half of them.
- Mechanism A's framing at `:361-364` ("driven by the **generation plan** … a boundary where 'no further Coqui work is queued' is a *fact* rather than a prediction") is true of the render-phase pair only. The pre-pass pair is memoised and fires lazily on the first derive of the other engine — demand-driven, gated by `hasCoquiPresence` (`:1941-1944`). That matters because the A-vs-B "plan-driven vs demand-driven" split is the load-bearing premise of the keep-both ruling this section exists to record.

### 🟡 New-3 — correctness (minor) — `f8eca78f` moved the trailing evict and left its own citation one line stale

`docs/features/264-vram-aware-gpu-placement.md:359` still says the trailing evict fires at `:2987`. At this head `synthesise-chapter.ts:2987` is a comment line (`groups are ALL already synthesised, purely to free VRAM. */`); the call is at **`:2988`**.

Verified as a regression introduced by this PR, not inherited: at pass 2's head `3e1cd0ff` the call *was* at `:2987`, and `f8eca78f` — the commit that rewrote the comment block immediately above it, and edited this same doc — grew that block by one net line without re-deriving the citation.

Low blast radius on its own. Flagged because it is the exact rot this PR exists to prevent, in the section that documents it, three commits after `3e1cd0ff` spent its entire diff repairing seven instances of it. The guard deliberately asserts no line numbers (`coqui-residency-policy.guard.test.ts:11-13`), so nothing catches this class — which is an argument for citing constructs rather than lines here, the way `synthesise-chapter.ts:2982-2983` already does after New-4 of pass 2 was fixed.

### 🟡 New-4 — cleanup — one of the three new CI wirings has no regression test, and it is the one that re-arms pass 2's blocker

The `main.py` `extraFiles` entry is pinned by `scripts/tests/verify-cache.test.mjs:418-423`. The `docs/features/264-vram-aware-gpu-placement.md` entry added on the next line at `verify-cache.mjs:432` is pinned by **nothing** — `force-rerun-triggers.test.ts` reads `server/vitest.config.ts`, a different mechanism, and no suite asserts `stepTouchedByDiff(stepByName['test:server'], ['docs/features/264-vram-aware-gpu-placement.md'])`.

Concrete failure scenario: a later tidy-up drops that one line from `extraFiles`. Every test stays green. A docs-only PR deleting the `## Coqui residency policy (side-18)` section then has `step_test_server` false → `verify.yml:304` skips the server leg → the guard never runs → required check green, and `.husky/pre-push` skips `verify:fast:branch` for a docs-only push besides. That is exactly half of pass 2's 🔴, silently reopened. A three-line sibling of the test at `:418` closes it.

### 🟡 New-5 — cleanup — the guard's archival note is now understated by three surfaces

`coqui-residency-policy.guard.test.ts:24-28` is the archival checklist for when plan 264 moves to `docs/features/archive/`. It names only "the cross-reference comments in `synthesise-chapter.ts` and `main.py`". Since `e97decb4` the same rename also invalidates `server/vitest.config.ts:231`, `server/src/force-rerun-triggers.test.ts:148-152`, and `scripts/verify-cache.mjs:432`. Two of those fail loudly on archival (`absPathUnder`'s `existsSync`), but `verify-cache.mjs`'s does not — `hashFile` (`scripts/verify-cache.mjs:628-634`) returns `'__missing__'` for an absent path rather than throwing, so that entry just goes quietly vacuous. The note is the one place a future archiver will look; it should list all five.

### 🟡 New-6 — cleanup — `verify.yml`'s "sidecar-only diff" rationale comment is now stale

`.github/workflows/verify.yml:296-302`:

> Drops the old `sidecar` disjunct deliberately: a sidecar-only diff used to run this job for zero tests …, so skipping the job entirely for that diff is behaviour-neutral, not a narrowing

After this PR a `main.py`-only diff schedules this job (via `step_test_server`'s new `extraFiles` entry) and, because `main.py` is also a `forceRerunTriggers` entry, runs the **whole** server suite rather than zero tests. The comment's stated rationale no longer describes the behaviour of the file it sits in. Same chore class as New-5: a comment this change made false.

### 🟡 New-7 — cleanup — `synthesise-chapter.ts` is the one of the three that did not need a `forceRerunTriggers` entry

`server/vitest.config.ts:229`. `main.py` and the policy doc genuinely have no other route — nothing can import them, so a full-suite trigger is the only mechanism, and the cost is accepted. `server/src/tts/synthesise-chapter.ts` is different: it is an ordinary TS module, and a one-line `import` in the guard would give it a real module-graph edge, which is what `--changed` selects on. As written, every PR that touches this ~3000-line hot file escalates the CI server leg from "tests that import it" to all ~8037, for a guard that only greps it. Every other entry in that array is a config, manifest, doc, or script with no importable form; this is the first application source file in it.

Non-blocking and possibly deliberate — but if it was deliberate, the comment at `:226-228` ("no module-graph edges") is inaccurate for this entry, since this one could have had one.

### Verdict

**Not mergeable as-is**, and this is the last automatic re-review round — the remaining findings need to go back to the author, not to another gate pass.

All five of pass 2's findings are genuinely addressed at the mechanism level, and the CI-wiring fix in particular is correct on all three surfaces and correctly narrowed the regression test it broke rather than deleting it. That was the round's hard problem and it landed.

What blocks is the same shape that has now recurred in all three passes: the fix to a prose-vs-code claim swings past the truth and lands a new one. Pass 1's fail-hard claim → pass 2's false abort symmetry → pass 3's false lock rationale (New-1). The doc's Coqui section has been through three rounds of correction and still carries two statements the source contradicts — New-1's "essential for the actual VRAM reclamation to work", which `_reclaim_after_drop`'s docstring denies in its own first sentence, and New-2's "Both evicts", which is false of the two `/unload` call sites the section never mentions. New-2 is the one with a live behavioural consequence if acted on: it would license swallowing a failure the cloned-voice resolver deliberately raises.

New-3 through New-7 are cheap and mechanical, but per findings-triage they are fixed in this round too, not deferred — New-4 in particular because it silently re-arms the exact hole pass 2 blocked on.

None of these needs a design pass: every one has a single defensible outcome. Recommend one dispatched fix agent per finding and an escalation to the repo owner for whether to spend a fourth review round on the result, since the automatic cap is now spent.

dudarenok-maker:
## PR review — pass 4 (head f3a0e77616bf9cd5c096fc9721a59397bdf0d21c, depth high) — manually authorized round

Re-review beyond the skill's automatic 3-pass cap, authorized by the repo owner to close this PR out. Deltas only. Every disposition below was re-derived from the files at this head, not read off the commit messages.

---

### Pass-3 findings — dispositions

**🟠 New-1 (lock-release rationale) — VERIFIED RESOLVED.**
`docs/features/264-vram-aware-gpu-placement.md:480-489` (lock caveat 3, `:473-495`) now gives a pure throughput/liveness reason and explicitly denies the correctness dependency ("for no correctness benefit… guaranteed by ACQUIRING it, not releasing it"). That matches `_reclaim_after_drop`'s actual docstring (`server/tts-sidecar/main.py:2128-2134`: "it's released because holding `_synth_lock` across this multi-second reclaim would block concurrent synth calls that only need the lock briefly, for no benefit") and `unload()`'s (`main.py:2048-2049`). Citation `:2127-2134` re-derived and correct; `:2040-2049` for the double-allocation race is correct; the `acquire → _drop_model_locked() → release → _reclaim_after_drop()` sequence matches `maybe_free_idle` at `main.py:2311-2319`; the device-restoration claim matches `_drop_model_locked` at `main.py:2110`.

**🟠 New-2 (only 2 of 4 `/unload` call sites) — VERIFIED RESOLVED in substance, but see New-8.**
All four wrapper call sites are now enumerated and all four line citations re-derived correct: `evictQwenForCoquiPhase` at `synthesise-chapter.ts:1895` (pre-pass) and `:2920` (leading); `evictCoquiForQwenPhase` at `:1949` (pre-pass) and `:2988` (trailing). "Both evicts must remain fail-soft" is now correctly scoped to the render-phase pair only, and the leading-rethrows/trailing-swallows asymmetry matches the code at `:2921-2943` and `:2989-2994`.

**🟡 New-3 (`:2987` → `:2988`) — VERIFIED RESOLVED.** Re-derived: `await withCallTimeout('coqui-evict-after-coqui-phase', (sig) => evictCoquiForQwenPhase(sig));` is at `server/src/tts/synthesise-chapter.ts:2988`.

**🟡 New-4 (docs `extraFiles` entry unpinned) — VERIFIED RESOLVED, and the test is non-vacuous.**
`scripts/tests/verify-cache.test.mjs:425-428` asserts `stepTouchedByDiff(stepByName['test:server'], ['docs/features/264-vram-aware-gpu-placement.md']) === true`. Checked it can actually fail: `test:server`'s globs are `server/src/**`, `scripts/**/*.{mjs,cjs,js}`, `server/tts-sidecar/scripts/**/*.mjs`, `server/tts-sidecar/requirements/**` (`scripts/verify-cache.mjs:378-389`) — none match `docs/**`, so the assertion is carried solely by the `extraFiles` entry at `:432`. `stepTouchedByDiff` (`:811-827`) does exact-set membership on `extraFiles`, so deleting the line flips it red. The file runs under `test:hooks` (`scripts/run-hooks-tests.mjs:10` globs `scripts/tests/*.test.mjs`). Also confirmed the CI path is real end-to-end: `scripts/ci-scope.mjs:13,43` derives `step_test_server` from `STEPS` + `stepTouchedByDiff`, so the entry genuinely schedules the `verify.yml` server-tests job.

**🟡 New-5 (archival note named 2 of 5) — VERIFIED RESOLVED as to the count; see New-9 for a defect in the replacement text.** All five are now listed at `server/src/tts/coqui-residency-policy.guard.test.ts:24-36`, and the `hashFile` sentinel claim in item 5 is accurate (`scripts/verify-cache.mjs:628-634` returns `'__missing__'` rather than throwing).

**🟡 New-6 (stale `verify.yml` rationale) — VERIFIED RESOLVED for the sentence it targeted; see New-10 for its sibling five lines above.** `.github/workflows/verify.yml:297-305` now carves out the `main.py` exception, and the mechanism it names is real (extraFiles schedules the job, `forceRerunTriggers` selects the tests under `npx vitest run --changed "$BASE"` at `:314`).

**🟡 New-7 ("no module-graph edges" over-claim) — VERIFIED RESOLVED in `server/vitest.config.ts:226-231`; see New-11 for the surviving second copy.**

---

### New findings

#### Correctness

**🟠 New-8 — the doc's pre-pass invariant is false at a third, unlisted invocation, and following it as written re-opens a closed §2.3 regression.**
`docs/features/264-vram-aware-gpu-placement.md:366-369` and `:453-465`

The doc states, twice and unqualified, that the pre-pass hooks "are called from *inside* voice-resolver try/catch blocks (`clone-voice-resolver.ts`)… **no local try/catch wraps them, by design**", that "A failure is therefore reported through the resolver's own existing, already-differentiated policy", and closes with "Any future change to the pre-pass evicts must maintain this asymmetry."

`beforeFirstQwenDerive` has a **third** invocation that is neither of those things — `server/src/tts/synthesise-chapter.ts:2110-2123`:

```ts
if (hasCoquiPresence && (chapterHasQwenGroups || clonedQwenRequests.length > 0 || designedQwenRequests.length > 0)) {
  try {
    await beforeFirstQwenDerive();
  } catch (err) {
    console.warn('[synthesise-chapter] failed to evict Coqui ahead of a Qwen phase — continuing; …', err);
  }
}
```

This is a local try/catch in `synthesise-chapter.ts` (not the resolver) that swallows unconditionally, independent of voice type. Its own comment at `:2101-2109` marks it `[NEW-CRITICAL]` and says it is "deliberately best-effort: caught and logged, NEVER rethrown, so a sidecar recycle window here can't abort the chapter", covering the "no derive needed at all, just a qwen RENDER" case that has no resolver call to hang the hook off (see also `:1915-1917`).

Concrete break: a maintainer applying the doc's stated must-preserve invariant deletes that try/catch as a violation ("no local try/catch wraps them, by design"). The chapter then aborts on a 502/ECONNREFUSED from the priming `/unload` on a book with only healthy/catalogue Qwen voices — the exact §2.3 shape Task 22 fix round 2 closed. Nothing catches it: the guard test asserts only marker presence, and I checked `server/src/tts/synthesise-chapter-derive-vram-partition.test.ts` — its failing-evict cases cover the resolver-side pre-pass (`:433-500`) and the render-phase trailing evict (`:1187-1217`), none the priming call. Tests stay green.

Note the shape: this is the *same* enumeration defect New-2 flagged, one level down. The doc counts the two **wrappers'** four call sites correctly, but the failure-semantics claim is about the two **memoised hooks**, which have five invocation sites — and the fifth has different semantics.

#### Cleanup

**🟡 New-9 — the new archival note's loud/silent split is backwards for items 1 and 2.**
`server/src/tts/coqui-residency-policy.guard.test.ts:27-28` — "two code comments fail loudly on archival when forgotten; three CI entries fail silently".

The guard asserts `occurrences(source, TOKEN)` only, and its own header says line numbers and paths are asserted "NOWHERE here on purpose". So leaving the stale `docs/features/264-vram-aware-gpu-placement.md` path in the `synthesise-chapter.ts` / `main.py` cross-reference comments (items 1 and 2) fails **nothing** — exactly as silently as items 3-5. What fails loudly on archival is this file's own `POLICY_DOC_PATH` `readFileSync` (`:37`), which is satisfied by updating that constant alone. Break: on archival an engineer updates `POLICY_DOC_PATH`, sees the suite go green, and trusts the note that items 1-2 would have shouted — shipping two stale doc paths, which is precisely the cross-reference rot this guard exists to prevent.

**🟡 New-10 — `verify.yml`'s carve-out fixed one sentence and left its neighbour stating the old claim.**
`.github/workflows/verify.yml:293-294` still reads "A sidecar-only PR touches no server TS graph → 0 tests run here, which is correct (server tests mock the Python sidecar)." With `'{**/server/tts-sidecar/main.py,…}'` now in `forceRerunTriggers` (`server/vitest.config.ts:233`), a `main.py`-only diff makes `vitest run --changed` do a **full** rerun — ~8,000 tests, not zero. The correction added at `:297-305` sits five lines below it, so the block now contradicts itself.

**🟡 New-11 — New-7's false claim survives in the sibling file.**
`server/src/force-rerun-triggers.test.ts:136-141`: "coqui-residency-policy.guard.test.ts reads these three files at RUNTIME… **No module-graph edges**, so a diff confined to any of them selected zero tests" — the entry immediately below it (`:142`) is `{ rel: 'src/tts/synthesise-chapter.ts', … }`, an ordinary importable module with module-graph edges. This is verbatim the claim corrected in `server/vitest.config.ts`; the second copy was not updated (`f3a0e776` touched four files, not this one).

**🟡 New-12 — the PR adds its policy marker into a comment block whose adjacent sentence the policy contradicts.**
`server/src/tts/synthesise-chapter.ts:1036-1037`: "Used ONLY by the cloned/designed-voice resolver pre-pass (below)". `evictCoquiForQwenPhase` is also used by the render-phase trailing evict at `:2988` (added by #1894). This PR's diff inserts the `COQUI-RESIDENCY-POLICY` paragraph into this exact comment at `:1042-1048`, six lines under a sentence the new doc directly refutes ("four call sites, not two").

**🟡 New-13 — "Node knows which engine each remaining chapter needs" overstates the mechanism.**
`docs/features/264-vram-aware-gpu-placement.md:376-377`. `synthGroupsSerialized` derives its engine set from the current call's `groupList` only (`synthesise-chapter.ts:2874-2886`); there is no cross-chapter plan at the evict site. The code's own comment (`:2951-2953`) scopes the "fact rather than a prediction" claim to "the Coqui phase is the last thing **this dispatch** does", and the trailing evict exists *because* XTTS otherwise "stayed resident… into the next chapter" (`:2945-2949`) — i.e. Node did not know. The doc's own Ownership section gets this right ("in the current dispatch"); only this sentence overreaches.

**🟡 New-14 — the PR body no longer describes the PR.**
The Summary still says only "docs and comments only" + guard test, and the Test plan lists `test:server` / `typecheck`. Four review rounds have since folded in CI-selection changes to `scripts/verify-cache.mjs`, `server/vitest.config.ts`, `server/src/force-rerun-triggers.test.ts` and `.github/workflows/verify.yml`, plus a new `scripts/tests/verify-cache.test.mjs` case that runs under `test:hooks` — none of which appear in the body, and `test:hooks` is not in the test plan. Per CLAUDE.md, incidental fixes are declared in the PR body ("Also fixed, found in passing: …").

---

### House gates

`Closes #1932` present and outside a code span ✓ · release-notes pair explicitly N/A with a stated reason ✓ · on-box acceptance explicitly N/A ✓ · no `cast.json` writes, no new config knob, no derived artifacts in scope ✓ · the guard's `main.py` assertion uses an exact occurrence count and both occurrences exist (`main.py:2292`, `:5056`) ✓ · all other line citations in the policy section re-derived and correct (`_idle_evict_steps` `:4989-5064`, coqui step `:5063`, self-admission skip `:5053` `if engine != "coqui"`, `maybe_free_idle` `:2273`, `_coqui_idle_ttl` `:8288` with `_COQUI_IDLE_TTL_DEFAULT = 30.0` at `:8285`, `registry.ts:848` `sidecar.coquiIdleTtl` default 30) ✓.

---

### Verdict

**One correctness finding (New-8) plus six cleanup findings.** New-8 is a false must-preserve invariant in the document that *is* this PR's deliverable, with a specific code site it would lead a maintainer to break and no test that would catch it. Under the normal triage rule a correctness finding re-triggers a round once fixed; this round was manually authorized, so the decision to fix-and-re-review or accept is the owner's, not mine to loop on.

New-9 through New-12 are all the same shape as findings this PR has already fixed once — the second copy of a corrected claim, left armed. They are cheap and mechanical.

dudarenok-maker:
## PR review — pass 5 (head c9a48dbcc72f366b7cc4e1c443a4d7224e04333e, depth high) — second manually authorized round

Re-review scope: verify each of pass 4's seven findings at the new head, then a full high-depth sweep of the accumulated diff with specific attention to whether `17f584cf` / `c9a48dbc` introduced their own stale claims. Every claim below was re-derived from the files at this head, not taken from the commit messages.

### Pass 4 findings — dispositions

| # | Finding | Disposition |
|---|---|---|
| New-8 | 🟠 doc's pre-pass invariant licensed deleting the render-prep `try/catch` | **VERIFIED RESOLVED** |
| New-9 | archival note's loud/silent split backwards | **VERIFIED RESOLVED** |
| New-10 | `verify.yml` self-contradicting sentences | **VERIFIED RESOLVED** |
| New-11 | duplicated "no module-graph edges" in `force-rerun-triggers.test.ts` | **VERIFIED RESOLVED** |
| New-12 | stale "used ONLY by" exclusivity comment | **VERIFIED RESOLVED** |
| New-13 | overstated cross-chapter-knowledge claim | **VERIFIED RESOLVED** |
| New-14 | PR body didn't describe the CI-wiring changes | **VERIFIED RESOLVED** |

Detail on the two that needed real re-derivation rather than a text diff:

**New-8 — resolved.** `docs/features/264-vram-aware-gpu-placement.md` now documents all three invocation patterns, and the closing "must preserve" statement was rewritten to *require* the wrapper rather than forbid it: "the render-prep site must retain its local fail-soft wrapper to prevent chapter aborts on transient sidecar errors." The failure semantics stated for each pattern match the code — the render-phase leading evict does rethrow `AbortError` (`server/src/tts/synthesise-chapter.ts:2934-2938`), the trailing evict swallows everything (`:2990-2996`), and the render-prep site at `:2115-2123` swallows unconditionally. The doc no longer licenses the deletion. (Two accuracy defects in the *new* text are filed below as New-16/New-17; neither reopens the regression.)

**New-10 — resolved, and the substantive claim independently checked.** The two sentences are coherent now. More importantly I verified the new comment's assertion is actually true rather than aspirational: `.github/workflows/verify.yml:136` runs `scripts/ci-scope.mjs`, which imports `stepTouchedByDiff` from `verify-cache.mjs` (`scripts/ci-scope.mjs:13`) and applies it per step (`:43`); `stepTouchedByDiff` (`scripts/verify-cache.mjs:816-819`) matches `extraFiles` exactly. So a `server/tts-sidecar/main.py`-only diff really does set `step_test_server` and reach the step at `verify.yml:307`. Same for a `docs/features/264-*.md`-only diff. The wiring is real end-to-end.

Also spot-checked and correct: the archival note's `hashFile` claim (`scripts/verify-cache.mjs:628-634` returns the `'__missing__'` sentinel rather than throwing, so item 5 genuinely does fail silently); no third copy of the "no module-graph edges" over-claim survives (`force-rerun-triggers.test.ts:111` and `:129` are correctly scoped to their own entries, as are `vitest.config.ts:173/182/220`); guard token counts match reality (1 in `synthesise-chapter.ts`, 2 in `main.py`, heading present in the doc).

---

### New findings — all cleanup tier, no correctness bugs

#### 🟡 New-15 (cleanup) — every `synthesise-chapter.ts` line citation below line 1036 is off by one, re-staled by this round's own commits

`docs/features/264-vram-aware-gpu-placement.md:357,363,364,370,373,379,380` + lock caveat 1.

`17f584cf` is `6 insertions, 5 deletions` on `server/src/tts/synthesise-chapter.ts` — a **net +1 line** inserted at ~:1036 (the "used ONLY by" fix, New-12). Every line below it shifted by one. `c9a48dbc`, committed 14 seconds later by the concurrent doc agent, touched **only** the doc (`git diff --numstat 17f584cf c9a48dbc` → one file) and so re-derived its citations against the pre-shift tree. Verified at head by `grep -n`:

| Doc says | Actual at head |
|---|---|
| `evictEngineForPhase` `:1015` | 1015 ✓ |
| `evictQwenForCoquiPhase` `:1029` | 1029 ✓ |
| `evictCoquiForQwenPhase` `:1049` | **1050** |
| `beforeFirstCoquiDerive` `:1882-1899` | **1883-1900** |
| `beforeFirstQwenDerive` `:1943-1953` | **1944-1954** |
| third pre-pass site `:2114-2122` (twice in the doc) | **2115-2123** |
| leading evict `:2920` | call at **2921** |
| trailing evict `:2988` | call at **2989** |

Only `:1015` and `:1029` survive, because they sit *above* the inserted line. All `main.py` citations (`:2040-2049`, `:2127-2134`, `:2273`, `:4989-5064`, `:5053`, `:5063`, `:8288`) and `registry.ts:848` are correct.

Note the specific irony: `58c7c1bf`'s message reads "the trailing evict call was cited as `:2987` but is actually at line 2988 … Corrected to `:2988`" — that correction was right when made and has now been re-staled by a sibling commit in the same PR. `server/src/tts/synthesise-chapter.ts:1038` carries the same drift ("line ~2988"), but is hedged with `~` and is fine as written.

Failure scenario: a maintainer following the doc's `:2114-2122` opens `synthesise-chapter.ts` at the `chapterHasQwenGroups` guard's closing paren rather than the `try {`, and following `:1049` lands on a comment terminator rather than `evictCoquiForQwenPhase`. This is the citation-rot class the guard test's own header comment says it deliberately refuses to assert on ("Line numbers are asserted NOWHERE here on purpose: they rot").

#### 🟡 New-16 (cleanup) — the doc states a narrower firing condition for the third pre-pass site than the code implements

`docs/features/264-vram-aware-gpu-placement.md:370-372` ("primes `beforeFirstQwenDerive` … **when only Qwen voices with no derive are present**") and lock caveat 1 ("**when only healthy/catalogue Qwen voices are present**").

The actual guard at `server/src/tts/synthesise-chapter.ts:2111-2114` is:

```js
if (
  hasCoquiPresence &&
  (chapterHasQwenGroups || clonedQwenRequests.length > 0 || designedQwenRequests.length > 0)
) {
```

The last two disjuncts fire precisely *when* Qwen derives are pending, which is the case the doc says is excluded. Concrete: a chapter with one cloned Coqui voice and one repairable cloned Qwen voice satisfies `hasCoquiPresence` and `clonedQwenRequests.length === 1`, so the fail-soft wrapper at `:2115` fires — before the cloned-Qwen resolver call at `:2126` — and it, not the resolver, is what actually performs the `/unload`.

What the doc omits is the mechanism the code's own comment (`:2105-2110`) spells out: the site *shares the memoised promise* with the resolver arms, so the resolver-mediated sites re-await an already-settled rejection and still apply their own fail-loud/fail-soft policy. End-to-end semantics are therefore preserved — this is why I'm scoring it cleanup, not correctness — but the doc as written mis-describes when the wrapper runs and drops the memoisation-sharing that makes the "three independent patterns" framing coherent.

#### 🟡 New-17 (cleanup) — the resolver-mediated call-site count is wrong (two claimed, four actual)

`docs/features/264-vram-aware-gpu-placement.md:366` ("**Two of the three pre-pass sites**") and lock caveat 1 ("**three distinct invocation patterns**", "**Two resolver-mediated sites**").

There are **four** resolver-mediated call sites, not two — each resolver calls both hooks, one per engine arm:

- `server/src/tts/clone-voice-resolver.ts:547` — cloned resolver, coqui arm (`beforeFirstDerive`), fail-loud
- `server/src/tts/clone-voice-resolver.ts:548` — cloned resolver, qwen arm (`beforeFirstQwenDerive`), fail-loud
- `server/src/tts/clone-voice-resolver.ts:996` — designed resolver, coqui arm, fail-soft
- `server/src/tts/clone-voice-resolver.ts:1073` — designed resolver, qwen arm, fail-soft

Plus the render-prep site at `synthesise-chapter.ts:2115` = **five** invocation sites across **three** failure-semantics patterns. The two *semantics* the doc describes (fail-loud cloned / fail-soft designed) are correct; the site count is not, and it is inconsistent with the same section's own counting method four paragraphs earlier ("The eviction wrappers have **four call sites**, not two" — which *is* correct: `:1896`, `:1950`, `:2921`, `:2989`).

This is the same undercount-of-invocation-sites shape as New-8 itself, reproduced inside New-8's fix.

---

### House gates

All pass or are explicitly N/A with a stated reason: guard test present with exact-occurrence (fail-closed) assertions and a recorded mutation check; `Closes #1932` present and outside any code span; release notes explicitly N/A (CI-selection wiring is covered by CLAUDE.md step 5's CI-only carve-out); on-box acceptance explicitly N/A; no `cast.json` writes, no config knob, no derived artifacts in scope; incidental fixes declared in the PR body.

### Verdict

**Cleanup-only pass — no correctness bugs found.** Per the findings-triage rule (≥1 correctness bug re-triggers a review once fixed; cleanup-only does not), this pass does **not** re-trigger a sixth round. The three findings above are documentation-accuracy defects in `docs/features/264-vram-aware-gpu-placement.md` only; no shipped behaviour, no test, and no CI wiring is wrong.

**MERGEABLE.** New-15 is the one I'd fold before merge if anything is folded — it is eight citations wrong in a doc whose entire purpose is to be the durable cross-reference, and it was introduced by this round, so it will not resurface anywhere else.

PR #2707, opened 2026-08-27
chore(docs): seed row-ID allocation markers and drop the Blocked section's borrowed IDs

dudarenok-maker:
## Summary

The data half of the stable-row-ID work. Register row IDs are currently *positional* — `check-onbox-register.mjs` check 4 enforces contiguity, so discharging a row forces every survivor below it to renumber, which rots citations tree-wide. This PR lays the data groundwork so PR 2 can delete that check; it changes no runtime code and no behaviour.

- **Per-group `next-id` allocation markers.** Each of the seven groups that exist at HEAD (A, B, C, D, E, G, H — no F, discharged) gains `<!-- next-id: <Letter>101 -->` below its heading. The floor is `101` for every group, not each group's own high-water: allocation counts *upward*, and a lower floor would eventually walk through `A99`, which the citation checker uses as its own sentinel for a definitely-nonexistent ID. The markers are inert today — `checkRegister` calls `stripFences` and never `stripHtmlComments`, and the marker matches neither the `^## ` section split nor the `^### <Letter>\d` row pattern. Group A's marker carries a caveat telling readers **not** to allocate from it until PR 2 lands, since an ID from that range fails the current check.
- **The Blocked section stops borrowing live Group E IDs.** `E6` and `E8` each named *two* rows — one live Group E row and one Blocked row — because the Blocked section borrowed IDs from a sequence that renumbers underneath it. The two Blocked headings now carry no ID, matching the section's other three rows, on both the register and its published twin. The live `E6`/`E8` rows are untouched.

## Why this is two PRs

`resolveBaselineGroups` (`scripts/check-onbox-register.mjs:501-516`) runs the *current* checker over the fetched `origin/main` baseline and rejects on any error. Any tightening of `checkRegister` is therefore retro-applied to `origin/main`'s copy, so a guard cannot land in the same PR as the data it requires. Data ships here; the checker changes ship in PR 2.

## Also fixed, found in passing

- The register's `--discharging` worked example cited "Group F's sole row, F1" as "a real, live example"; Group F has been discharged. Generalised rather than renamed to another group — no group currently has exactly one row, and a fresh named example would inherit the same failure mode.
[...]

Issue #2634, opened 2026-08-23
Register uses E6 and E8 twice, and check:onbox-register does not check row-ID uniqueness


dudarenok-maker:
Found during PR #2630's review gate (pass 3).

## Two row IDs are used twice, and the checker cannot see it

`docs/testing/onbox-acceptance-register.md` uses `E6` and `E8` **twice each**:

| ID | Group E (live) | Blocked — hardware not available |
|---|---|---|
| `E6` | fe-57 venv-bootstrap progress card (#1883, plan 270) | ops-35 ffmpeg floor (#1877, plan 269) |
| `E8` | revoke is loopback-only (#2269, plan 225) | ops-36 golden-assembly on a second ffmpeg build (#1880, plan 272) |

So `register row E6` resolves ambiguously. PR #2630 corrected `270-openapi-setup-surface.md` to cite `E6` — correct for the Group E row, but a reader landing on the Blocked one has no way to tell.

## The checker has a matching gap

`scripts/check-onbox-register.mjs` checks duplicate **group letters** (`duplicateLetters`, ~lines 103–138) and glance-table-vs-body counts. It does **not** check that row IDs are unique across the whole file, so this passes green today.

## Decision owed — why this is not just fixed

The Blocked section's rows carry `E`-prefixed IDs from the group they *would* live in, which is what creates the collision. More than one defensible resolution:

1. **Continue Group E's sequence** into the Blocked section (`E10`, `E11`). Simple, but blocked rows then renumber whenever a live E row discharges, and they are exactly the rows that sit still longest.
2. **Give blocked rows their own namespace** (`XE1`, or `BLOCKED-1`). Stable, but a new ID grammar for every consumer — the checker, the live view, the run sheets.
3. **Drop IDs from blocked rows entirely** and cite them by title. They are not schedulable work, so a positional ID buys little.

Whichever wins, add a uniqueness check to `check-onbox-register.mjs` so this cannot recur — that part needs no decision and should land with the fix.

## Related

#2629 is the broader design pass on how anything should reference a register row durably. This is the narrower, concrete instance: today the register cannot even name its own rows unambiguously. Sequencing them together is probably right.

dudarenok-maker:
Closed as delivered by PR #2707, and recording that this was a **duplicate of #2653** — the same finding, filed three weeks apart.

[...]

Issue #2603, opened 2026-08-22
No mechanical guard validates A\d+/E\d+ row references against onbox-acceptance-register.md's actual current headings


dudarenok-maker:
## What

`scripts/check-onbox-register.mjs` validates the register `.md` against its live-view `.html` twin — row IDs, per-group counts, the total. It has no visibility into anything else that cites a register row by number: a per-feature run sheet's criterion↔row map, a design-plan cross-reference, a GitHub issue body, or a sitting-plan's row bins and totals.

## Why this matters

Found during PR #2578's review-gate loop, rounds 19-21. The register has been renumbered twice in the course of that one PR (two separate `origin/main` merges — PR #2587's status update, then a comprehensive "wave 4 close-out" restructuring, PR #2551 and others) — each renumbering shifted roughly a third of the register's rows. Each time, the PR's own branch content (run sheets, plan citations, a filed issue body, a sitting-plan row list) had to be hand-reconciled against the new numbering, and each reconciliation attempt missed sites the next review round caught:

- Round 19 reconciled A39/A41's status framing at 7 sites; round 20 found 3 more sites with the same false claim, invisible to the literal-string grep round 19 used.
- Round 20 fixed a "five states" vs "eight states" table-count confusion at 4 of 6 sites; round 21 found the remaining 2 had made a document state three different numbers for the same concept.
- Round 21 alone found **26 stale row-number occurrences in one run sheet file**, a register-authored "recompute" note using entirely pre-renumbering IDs, a row whose own body pointed at itself ("same procedure as A39" inside row A39), a sitting-plan total that had drifted by exactly one row two rounds running, and a GitHub issue body citing three rows when only two of the three were actually correct.

None of this is visible to `check:onbox-register`, which only validates the register/live-view pair's own internal counts. Twenty-plus rounds of careful, adversarial grepping kept clearing one surface at a time because there is no single mechanical check spanning all of them.

## What's owed

A design decision, not a mechanical fix — the shape of "resolve every `A\d+`/`E\d+`/etc. token in a citing surface against the register's actual current row headings" needs choices:
- **Scope**: which files count as "citing surfaces" — `docs/testing/**`? `docs/features/**`? GitHub issue bodies (would need the `gh` CLI or API, likely out of scope for a CI-run script)?
[...]

PR #2717, opened 2026-08-27
feat(scripts,docs): replace register row contiguity with allocate-once IDs

### LINKED ISSUE #2629: Register-row citations in code rot on every discharge
## What

Source-code comments cite on-box acceptance register rows by **positional ID**
(`register row A32`, `row A19`, …). Those IDs renumber contiguously whenever a
row is discharged, so every citation silently rots — it keeps pointing at a row
number that now names something else entirely.

The wave-5 campaign (PR #2626) discharged three Group A rows, shifting every
later ID by three and rotting **38 citations across 11 files**. Example: two
comments about the catastrophic-WER override (#2055) cited `A37`, which now
names an unrelated ORT-marker row.

## Fixed here

PR corrects all 38 by matching each comment's subject to the register heading it
actually describes:

- `A32` → `A29` — characterId drift, Wave 1 resolver (#2040)
- `A33` → `A30` — characterId drift, Wave 3 `--apply` run (#2040)
- `A37` → `A34` — catastrophic-WER override (#2055)

Plus one pre-existing mislabel (a test title citing `A32` for a `mayrin` case
that was never A32's subject) and the derived `server/.env.example`, which
embeds registry descriptions verbatim.

Verified-correct citations left untouched: `A19`, `A36`, `H2`, `E1`, `G2`.

## Still open — the recurrence

**This will rot again on the next discharge.** Correcting the numbers is not a
fix for the mechanism, and the campaign has many waves left.

**The decision owed:** how should code reference a register row durably? At
least three defensible answers, and it needs a design pass rather than a
unilateral pick:

1. **Cite the issue number only** (`#2040`), dropping the positional ID. Durable
   and already present in most comments, but loses the pointer to the specific
   row when one issue owns several (e.g. #2192 owns four ORT rows).
2. **Give each row a stable slug** in the register (`A29 · <slug>`) and cite the
   slug. Durable and precise, but adds a field to every row and to
   `check:onbox-register`.
3. **Mechanically check the citations** — extend `check:onbox-register` to grep
   source comments and verify each cited ID still resolves to a row whose
   heading mentions the same issue number. Catches rot at PR time instead of
   preventing it.

Option 3 composes with either of the others.

## Benefit

*Benefit (technical):* a citation that survives the next discharge, and/or a
gate that catches rot the same way `check:onbox-register` already catches the
register's own internal drift. Today a comment can point a reader at completely
unrelated shipped behaviour, and nothing anywhere notices.

dudarenok-maker:
## Summary

The guard half of the stable-row-ID work, and the half that actually stops the renumbering. PR #2707 shipped the data; this replaces the check that *forced* the renumbering in the first place.

`check-onbox-register.mjs`'s check 4 required each group's row IDs to be **contiguous from 1**. That is what made discharging a row a renumbering event: remove `A8` and every row below it had to shift up, silently re-pointing roughly 210-225 `row <ID>` citations across ~65 files — including GitHub issue bodies no tool in this repo can reach. Check 4 is deleted and replaced by two checks that permit gaps and forbid reuse:

- **4a — whole-document ID uniqueness.** Any `### <Letter><N>` heading whose ID appears more than once anywhere in the register is an error. Deliberately document-wide, not group-scoped: the `E6`/`E8` Blocked-vs-live-Group-E collision that PR #2707 removed was *structurally invisible* to a group-scoped check, which is how it survived long enough to be filed twice (#2634, #2653). Run against a pre-PR-1 register, 4a reports exactly `Row ID E6 appears more than once` and `Row ID E8 appears more than once`.
- **4b — allocation floor.** Every group must carry a `<!-- next-id: <Letter>N -->` marker; the marker may not sit below `ALLOCATION_FLOOR = 101`; and no row may hold an ID at or above its group's marker. The floor is `101` for *every* group rather than each group's own high-water because allocation counts **upward**: a lower floor clears today's rows but eventually walks through `A99`, which `check-register-citations.mjs` uses as its sentinel for a definitely-nonexistent ID (65 occurrences in its own tests).

**Row IDs are now allocated once and never reused. Discharging a row renumbers nothing** — the ID that disappears is exactly the ID you discharged, and a citation to a discharged row goes visibly dangling instead of silently re-pointing at someone else's row.

## What 4b does not do, stated plainly

4b keeps **forward** allocation honest: an ID minted from a group's marker is one no row has ever held. It does **not** stop someone hand-typing a discharged row's old ID back in — that ID sits below the marker, so it passes. Catching that would need a retired-ID ledger, and this register's standing rule is that it tracks state, not history. So the guard there is the written procedure, and both the checker's header and the register's own allocation guidance say so rather than implying the checker has your back.

## Prose, which is most of the diff

The rules lived in prose as much as in code, and that prose is now false. Swept in three passes:

- the checker's own comments and error strings (`0717844b`) — including the `--discharging` "renumbering wrinkle" paragraph, which told operators to name the group's *highest* ID rather than the one they discharged;
- the register's own runbook, `CLAUDE.md`, `check-register-citations.mjs`'s header premise, and the run sheets (`b5aa1ad0`, finished in `4bc6116f`). Genuinely historical statements were dated rather than deleted — several run-sheet observations are only intelligible under the old rule, so they now say so explicitly instead of reading as current instruction.

Each sweep grepped `renumber`/`contiguous`/`contiguity` and judged every hit rather than trusting the plan's site list; between them that turned up **six** sites the plan never named. They also deliberately left alone a class of false positive — `backlog-sync.mjs`/`thin-backlog.mjs`'s "never reused or renumbered" is about `<prefix>-<n>` backlog IDs, and `check-register-citations.mjs`'s "annotate, don't renumber" *is* about register row IDs but states a rule that is still true.

## The retro-application rule, and why this is a second PR

`resolveBaselineGroups` runs the **current** checker over the **fetched `origin/main`** copy of the register and rejects the baseline on any error. So 4a and 4b are retro-applied to `main`'s register the moment this merges — a guard cannot land in the same PR as the data it requires, which is why the markers shipped separately in #2707. Verified rather than assumed: the new `checkRegister` run against `git show 9237ce78:docs/testing/onbox-acceptance-register.md` returns **0 errors**, raw, LF-normalised and CRLF-converted. The rule is now recorded in the checker's own file header, where the next person to tighten a check will see it.

That same rule creates a new way to break the baseline for every lane at once: **a branch adding a `## Group <Letter>` section without a marker.** Nothing told authors to add one, so this PR adds that instruction — with its consequence — to the register's maintenance guidance, alongside a restored procedure for allocating a new row ID within an existing group.

- A third fixture builder, in `scripts/tests/check-onbox-register-cli-no-flags.test.mjs`, also needed the marker. It is invisible when running the register test file alone and only fails under full `test:hooks` — surfaced by this branch's own retro-application, not pre-existing rot.
- Group A's marker still carried #2707's caveat *"DO NOT allocate from this yet — until the contiguity check is removed"*. This branch removes that check, so the caveat invalidated itself.
- `check-register-citations.mjs` asserted citation IDs "stay subject to renumbering" as forward-looking fact, contradicting the guarantee this PR ships.
- The `--discharging` refusal advised checking a state that can never produce it (a row absent from the `origin/main` baseline is consumed as a *match* earlier in the same file). Rewritten to name the real cause.

## Test plan

`scripts/tests/check-onbox-register.test.mjs`: **116 pass**, up from 107 — 14 migrated for the new rules, plus 9 new. Full `npm run test:hooks`: **1608/1608**. `npm run lint` exits 0. `check:onbox-register` and `check:register-citations` both green, and Check A's already-annotated note count holds at **16** — the tripwire for having moved a citation that should not move.

**Every new guard condition is mutation-verified**, because a green suite only proves the tests pass, not that they can fail. Each was broken in turn and confirmed to redden its own named test, then restored:

| Mutation | Test that went red |
|---|---|
| 4a `count > 1` → `count > 999` | same row ID in two sections |
| 4a scan narrowed to group sections | same row ID in two sections |
| 4a within-group detection broken | same row ID twice in one section |
| 4b `nextId === null` branch disabled | missing marker |
| 4b `n >= nextId` → `n >= 999999` | row at or above next-id |
| 4b `nextId < FLOOR` → `< FLOOR - 1` | floor boundary |
| 4b's invalid-heading suppression deleted | suppression test |
| `ALLOCATION_FLOOR` `101` → `50` or `1` | floor-value pin |
| `parseNextIdMarker` `${letter}` → `[A-Z]` | marker letter test |
| `parseNextIdMarker` trailing anchor dropped | trailing-content test |
| `formatRowList` reverted to `n === i + 1` | range not starting at 1 |

Four of those tests exist only because review passes found the corresponding guard lines **unkillable** — the code was correct but no test could fail if it broke. The two that mattered most: an off-by-one in the floor comparison would have accepted a marker of `A100` and thereby made `A99` allocatable, the single ID the `101` argument exists to protect; and `ALLOCATION_FLOOR` itself could be lowered to `1` with the whole suite green, because the only floor test derived its fixture from the constant.

## On-box acceptance and release notes

**No on-box acceptance row is added or discharged** — nothing here ships behaviour that real hardware could prove. **No release-notes entry** — no user- or operator-visible delta; this is repo tooling. Both stated explicitly rather than silently omitted.

**No republish is owed:** `docs/testing/onbox-acceptance-register-live-view.html` is unchanged by this PR, and its two renumber-era mentions are already-dated historical callouts.

Closes #2629
Refs #2599
Refs #2603

Plan: `docs/superpowers/plans/2026-08-27-onbox-register-stable-row-ids.md` (Track A, PR 2)
Spec: `docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md`


dudarenok-maker:
## PR review — pass 1 (head `b5aa1ad0`, depth high)

Scope: whole-branch diff `9237ce78...HEAD` (4 commits, 11 files) · verified independently before probing: `check-onbox-register.test.mjs` **112/112**, full `test:hooks` **1604/1604**, `check:onbox-register` exit 0, `check:register-citations` exit 0 with Check A at **16**, `eslint` exit 0 on all four changed script files, PR title accepted by `scripts/validate-commit-msg.mjs`. The retro-application claim and all seven claimed mutations were re-derived from scratch against a scratch mirror of `scripts/` + `docs/testing/`, not taken from the PR body.

---

### 🟠 Significant — the new `--discharging` refusal remedy names a condition that can never cause the error, and inverts the one that can

`scripts/check-onbox-register.mjs:1052-1053`

The replacement text tells the operator to *"check for a typo, or that the row is genuinely absent from both the register and the origin/main baseline."* Absence from the baseline is explicitly consumed as a **match**, not an error, by this file's own `:955-966`:

> `#2272: every id in extra is, by definition, "live-only" … a named id is marked consumed here, BEFORE the baseline filter below … That way a name for a row the baseline ALSO already lacks (harmless …) still counts as a match, not a typo`

Executed against the real `checkLiveView`, `direction: 'extraOnly'`, live page `E1,E2,E7`, `--discharging E7`:

| working register | `origin/main` baseline | error fired? |
|---|---|---|
| lacks E7 | **lacks E7** (what the remedy tells you to check) | **no errors at all** |
| **still has E7** | still has E7 | ✅ this error fires — the real cause |
| lacks E7 | **still has E7** (the documented `--discharging` case) | no errors — works as designed |

So the sentence points at the one state that guarantees the error *cannot* fire, and the state it implies is wrong (`origin/main` still holding the row) is exactly the precondition `--discharging` exists to serve — the flag is *for* the pre-merge window in which the baseline has not caught up. Concrete failure: an operator discharges `E7`, removes it from the glance table but leaves the `### E7` heading in the register, runs `--against-published --discharging E7`, gets this error, follows the remedy, finds E7 still on `origin/main`, and concludes the refusal is about the un-merged baseline — the one thing it is never about. The correct remaining causes are: the ID is still present in *this* register, or it is absent from the published page. The parenthetical at `:1050` already says the first; the added clause contradicts it.

The prior text at this spot (the renumbering wrinkle) was true when written; this replaces a true statement with a false one, so it is a regression introduced by this PR, not pre-existing rot. `scripts/check-onbox-register.mjs:1473` (`"each error explains why that ID didn't match"`) now points at that wrong explanation.

---

### 🟠 Significant — the PR deletes the only "take next-id, then bump it" instruction in the tree, and `### A38` still passes every check

`docs/testing/onbox-acceptance-register.md:555` (deleted comment) · `:303-314` (its replacement)

At `9237ce78` the register carried, above Group A's marker, the sole allocation instruction anywhere in the repo:

> `A row ID is allocated once and never reused: take this value for a new row, then bump it. DO NOT allocate from this yet … Until then, keep following the existing numbering.`

This PR deletes it and inverts the correct behaviour (allocation is now live), but the replacement blockquote covers only **adding a new `## Group <Letter>` section**, and closes with *"Copy an existing group's bare `<!-- next-id: X101 -->` line — no extra caveat comment needed."* `git grep -i "bump\|allocat"` over the register at HEAD returns no allocation procedure at all; nor does `CLAUDE.md`. The only surviving statements of the rule are the design plan/spec and the checker's own error strings, which appear *after* you have got it wrong.

Executed, on the real register at HEAD, adding one row the obvious way (glance A 37→38, total 60→61, `### A38 · …` appended to Group A):

```
current errors: 0
after adding `### A38` the obvious way: 0 errors
```

`A38` is one of the seven IDs the spec (§1) and plan (Global Constraint 4) say must **stay dangling** — `check:register-citations` currently reports `docs/testing/onbox-sitting-plan.md:74 — cited A38 — no such row … annotated as discharged/removed`. Re-minting it silently re-points that citation, its annotation becomes false, Check A stops printing it (the ID now exists), and Check C's fatal half cannot see it (prose idiom, not an anchored heading with a subject number — per the checker's own measured-coverage report). That is the exact rot this PR ships to prevent, reachable by the single most likely next edit.

The mirror-image mistake — take `A101` and forget to bump the marker — is the **other** shape the new blockquote does not warn about, and it has the same blast radius the blockquote correctly describes for a missing marker: `4b` fires on `main`, and every lane's `--against-published` dies with `CANNOT_VERIFY_BASELINE_ERROR`. Priority-1 answer, stated plainly: of the realistic next-merge shapes, "new group, no marker" is warned about; "allocated from next-id, forgot to bump" and "kept numbering from the old sequence" are not, and the second of those is the one that passes silently.

---

### 🟠 Significant — `ALLOCATION_FLOOR` can be lowered to 50, or to 1, with the entire suite green

`scripts/check-onbox-register.mjs:208`

The only test that touches the floor is `4b: a next-id below the allocation floor is reported`, which builds its fixture from `ALLOCATION_FLOOR - 1` — self-referential, so it moves with the constant. Every other fixture hardcodes `101`, which satisfies any lower floor. Mutation-verified against the mirror:

| mutation | result |
|---|---|
| `ALLOCATION_FLOOR = 101` → `50` | **114/114 green — survived** |
| `ALLOCATION_FLOOR = 101` → `1` | **114/114 green — survived** |

`check:onbox-register` also stays green (the real markers are `101`, which clears any lower floor). So the single number the whole design rests on — Global Constraint 3, *"Never lower it. If a fixture breaks against it, fix the fixture, not the floor"*, and the constant's own comment saying the same — has no mechanical backstop whatsoever. Concrete failure: a future contributor hits a floor error, "fixes" it by editing the constant to match the offending marker, and every gate stays green; allocation then counts upward from below 99 and eventually mints `A99`, the citation checker's own definitely-nonexistent-ID sentinel (`check-register-citations.mjs:733/:759/:785` + 65 occurrences in its tests) — the precise collision the `101` argument exists to prevent.

This is the same "guard line no test can kill" shape the earlier review pass already fixed three times in this file, and it is the omission from the PR body's mutation table: seven mutations are listed, none of them the constant.

Flagged 🟠 rather than 🟡 in full honesty about the taxonomy: it is a test-coverage defect, not wrong runtime behaviour today.

---

### 🟠 Significant — two run sheets the plan named by file still assert "Rows renumber contiguously" as the current rule

`docs/testing/onbox-sitting-qa-gate.md:50` · `docs/testing/onbox-sitting-device-browser.md:26`

Task 13's own inventory names `onbox-sitting-qa-gate` (2) and `onbox-sitting-device-browser` (1) in its run-sheet row, and its decision rule gives this exact phrase as the canonical rewrite case:

> **A statement of the current rule** ("rows renumber contiguously", …) → **rewrite**. It is now false.

Both files still read, in the present tense:

- `onbox-sitting-qa-gate.md:50-51` — *"Rows renumber contiguously in Group A: the old **A23** … is now **A22**"*
- `onbox-sitting-device-browser.md:26` — *"Rows renumber contiguously: the old **E7** … is now **E6**"*

Step 5 of that task was not applied to either. I swept the rest of the plan's named run sheets (`ort-marker-*`, `onbox-sitting-plan`, `onbox-wave4-results`, `onbox-acceptance-staleness-audit`, `fs38-wave3-*`) and every remaining hit is genuinely past-tense history, correctly left alone — these two are the only survivors, and they are the two the plan's own example sentence describes. Concrete failure: an agent opening the qa-gate pack for the next sitting reads that as the governing rule, discharges `A22` and renumbers the survivors down. Under the new checker that renumbering reports **zero** errors (all IDs unique, all below 101), so nothing catches it and ~210–225 citations re-point silently.

This also falsifies the PR body's claim that both sweeps *"grepped `renumber`/`contiguous`/`contiguity` and judged every hit"*.

---

### 🟡 Minor — `formatRowList`'s rewritten contiguity logic has no test

`scripts/check-onbox-register.mjs:193-196`

The diff changes `isContiguousFromOne` (`n === i + 1`) to a from-anywhere run check — the behaviour change that lets `A101–A103` collapse. Reverting it to the old form leaves **114/114 green** (mutation-verified). Every existing range assertion uses a run starting at 1 (`check-onbox-register.test.mjs:149` `A1–A2`, `:644` `E1–E7`, `:726` `A1–A2`), so the new branch is never exercised. Two smaller survivors in the same bucket, both in code this PR added: dropping `parseNextIdMarker`'s trailing `\s*\r?$` anchor also survives 114/114, and `parseAllRowHeadings` is `export`ed but imported by nothing (its `letter` and `number` fields are never read anywhere).

### 🟡 Minor — the deleted within-group duplicate test has no replacement

`scripts/tests/check-onbox-register.test.mjs` (base `:166`, `check 4: non-contiguous row numbers (a duplicate) are reported independently of check 1`)

That test pinned two things: a duplicate ID *inside one group* is reported, and it is reported **without** a spurious count-mismatch alongside. The new `4a` test covers only the cross-section (`Blocked` reusing a live Group E ID) case. The behaviour still holds — I ran it: two `### A1` headings in Group A yield exactly `Row ID A1 appears more than once (2 headings)` and no check-1 error — but nothing pins it any more. CLAUDE.md: *"Never delete or `.skip` a test without an explicit replacement or follow-up plan item."*

### 🟡 Minor — the spec's committed residual-limitation note never landed in the checker header

`docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md:168-175` says of `4b`:

> Neither stops an author hand-typing a **discharged** `A1`…`A37` ID into a new row … **Documented in the checker header alongside its other envelope edges.**

`scripts/check-onbox-register.mjs`'s header documents the retro-application rule (correctly, and usefully) but says nothing about this. `grep -i "residual\|hand-typ\|retired-ID\|ledger"` finds the fence-balance and `#2199` caveats only. This matters more given finding 2: `4a`'s own error string and `check-register-citations.mjs:4-8` now assert *"allocated once and never reused"* / *"nothing ever does"* as flat guarantees, with nothing on the code side saying where that guarantee stops being mechanical.

---

### ✅ What is solid

- **The retro-application verification is real, and I reproduced it independently.** The new `checkRegister` over `git show 9237ce78:docs/testing/onbox-acceptance-register.md` returns **0 errors** raw, LF-normalised and CRLF-converted, and the same over `HEAD:`. The rule is now recorded in the file header where the next person to tighten a check will meet it — the right place for it.
- **All seven claimed mutations reproduce exactly**, each reddening its named test and, in six of seven cases, *only* that test. `M3` reddens two, both legitimately marker-presence assertions. `4a`'s `invalidRowHeadingLetterSet` contamination worry is handled correctly: the 4b-only suppression mutation reddens the dedicated suppression test and nothing in check 1. I also re-derived `n >= nextId` → `n > nextId` and `< FLOOR` → `<= FLOOR`; both are killed.
- **Global Constraint 2** holds: the `names A38` test builds its own 37/38-row fixture rather than reading the real file, and `computeMaxRowNumber` now derives the candidate *and* asserts `max + 1 < nextId` with a comment naming the tempting wrong repair.
- **Global Constraint 3** holds: floor `101`, one constant, applied identically to all seven group sections; nothing per-grouped or lowered.
- **Global Constraint 4** holds mechanically: the sorted set of `^### <Letter><N>` headings is byte-identical between `9237ce78` and `HEAD` (60 rows, `diff` clean).
- **The deliberate skips are the right call.** `backlog-sync.mjs:137` / `thin-backlog.mjs:130` are unambiguously about `<prefix>-<n>` backlog IDs — different rule, still true. `check-register-citations.mjs:44-54`'s "annotate, don't renumber" *is* about register row IDs (the PR body's "a different … rule about citation and backlog IDs" mis-describes it), but the rule it states — annotate a discharged row's citation rather than repointing it — is still true and its mechanism (annotated → non-fatal NOTE) is unchanged, so leaving it is correct.
- **"No republish owed" checks out.** `onbox-acceptance-register-live-view.html` is absent from the diff, and the register's row set, glance counts and owed total are all unchanged, so the tracked-pair check stays green with nothing to re-sync.
- **Not a finding, checked and dismissed:** `prettier --check` fails on all four changed script files — but it fails on their `9237ce78` versions too under the repo's own `.prettierrc`, and `format:check` is wired into no hook, no `verify` leg and no workflow. Pre-existing, un-gated, not this PR's.

### Verdict

Four 🟠 and three 🟡. The mechanism is sound and the mutation discipline behind it is genuinely good — findings 1, 2 and 4 are all in the *prose half* the PR body itself identifies as the half with no mechanical gate, and finding 2 is the one that matters: the PR flips which ID a new row should get and simultaneously deletes the only sentence that said so, leaving the wrong answer passing every check. Finding 3 is the mutation table's own blind spot. Not mergeable as-is.

dudarenok-maker:
## PR review — pass 2 (head `5c89c25e`, depth high)

Scope: the three fix commits (`b2e3561d`, `4bc6116f`, `5c89c25e`) re-derived against the whole-branch diff `9237ce78...HEAD` · verified independently before probing: `check-onbox-register.test.mjs` **116/116**, full `npm run test:hooks` **1608/1608**, `check:onbox-register` exit 0, `check:register-citations` exit 0 with Check A at **16**, `eslint --max-warnings 0` exit 0 on all four changed script files. Every mutation and every state below was re-applied from scratch against a scratch mirror of `scripts/` + `docs/testing/` — not taken from the PR body, and not taken from pass 1.

### Pass 1 dispositions

|---|---|---|
| 1 | `--discharging` remedy named a state that can never fire | **RESOLVED** for every per-row path — 7-state truth table re-derived against the real `checkLiveView`. One residual firing state still unnamed (🟡 below). |
| 2 | the only allocation instruction was deleted; `### A38` passes every check | **RULE RESTORED, WORKED EXAMPLE WRONG** — 🟠 below |
| 3 | `ALLOCATION_FLOOR` was unkillable | **VERIFIED RESOLVED** |
| 4 | two run sheets asserted renumbering in the present tense | **VERIFIED RESOLVED** |
| 5 | `formatRowList` range / `parseNextIdMarker` anchor / `parseAllRowHeadings` dead fields | **VERIFIED RESOLVED** — but the narrowing staled the plan's own interface block (🟡 below) |
| 6 | the deleted within-group duplicate test | **HALF RESOLVED** — 🟡 below |
| 7 | the spec's residual-limitation note never landed in the checker header | **VERIFIED RESOLVED** — the header paragraph (`:17-26`) and 4b's rewritten inline comment (`:390-395`) agree with each other, and both are true of what 4b actually does |

Mutations re-derived independently (each applied, run, restored — none accepted on a green suite):

| Mutation | Test that went red | Only that test? |
|---|---|---|
| `ALLOCATION_FLOOR` `101` → `50` | `ALLOCATION_FLOOR is pinned at 101 and must never be lowered` | yes — 115/116 |
| `ALLOCATION_FLOOR` `101` → `1` | same | yes |
| `formatRowList` reverted to `n === i + 1` | `check 1: a contiguous run NOT starting at 1 still collapses to a range…` | yes |
| `parseNextIdMarker`'s trailing anchor dropped | `parseNextIdMarker rejects a line with trailing content after the closing -->` | yes |

Also re-confirmed: retro-application holds (the new `checkRegister` over `git show 9237ce78:docs/testing/onbox-acceptance-register.md` returns **0 errors**, raw / LF-normalised / CRLF-converted); Global Constraint 2 holds (every new test fixtures its own register text); Global Constraint 3 holds (one floor, `101`, applied identically to all seven groups); Global Constraint 4 holds (the sorted `### <Letter><N>` heading set is byte-identical between `9237ce78` and `HEAD`, 60 rows, `diff` clean); `onbox-acceptance-register-live-view.html` is absent from the diff, so no republish is owed; `Closes #2629` / `Refs #2599` / `Refs #2603` are present as plain text outside any code span.

---

### 🟠 Significant — the restored allocation procedure teaches the new rule using the one ID in the tree that must never be minted, with a marker value the new floor makes illegal

`docs/testing/onbox-acceptance-register.md:318-319`

> Take the marker's current value as the new row's ID, then bump the marker by one in the same commit (**e.g. `<!-- next-id: A38 -->` becomes `<!-- next-id: A39 -->` after you mint `A38`**).

The *rule* is right. The worked example is the exact counter-example, twice over:

1. **`A38` is a retired ID with a live, annotated citation.** `check:register-citations` on this tree reports it as one of the 16 Check A notes: `docs/testing/onbox-sitting-plan.md:74 — cited A38 — no such row … annotated as discharged/removed`. That annotation's own words are *"that row id has since been discharged and reused by later waves, and **no A38 exists in today's register**"*. Minting `A38` makes that sentence false, removes the note from Check A's list (16 → 15), and silently re-points the citation.
2. **A marker of `A38` cannot legally exist under the rule this same blockquote is introducing.** `ALLOCATION_FLOOR = 101`, so `<!-- next-id: A38 -->` is reported by 4b as *"Group A's next-id (A38) is below the allocation floor A101"*. Group A's real marker is `<!-- next-id: A101 -->` (register `:566`).

Executed against the real register, following the example literally (glance A 37 → 38, total 60 → 61, `### A38 · …` appended to Group A):

```
current errors:                                       []
mint A38 (the blockquote's worked example):           []
mint A101 + bump the marker (the blockquote's rule):  []
```

Both pass. **Nothing mechanical distinguishes the two** — which is precisely what the blockquote's own next sentence says (`:323-325`), so the example *is* the whole guard.

Concrete failure scenario: an agent adds a Group A row, sees Group A ends at `A37`, and finds a blockquote whose only concrete artifact reads *"`<!-- next-id: A38 -->` … after you mint `A38`"*. It mints `A38`. `check:onbox-register` is green, `check:register-citations` is green (one fewer non-fatal note, nothing fails), CI is green, the PR merges. That is pass 1's finding 2 reproduced verbatim: the fix supplied the missing instruction and then illustrated it with the failing case. `A38` is also exactly what positional thinking (37 rows → next is 38) reaches for, so the example **confirms** the instinct this PR exists to break.

The fix is one line, and the true current values are the right ones: `<!-- next-id: A101 -->` becomes `<!-- next-id: A102 -->` after you mint `A101`.

**Correctness bug** — in an instruction agents follow literally, which is the `docs`-tier subclass this repo's own gate singles out.

---

### 🟡 Minor — `--discharging`'s new remedy is true for every per-row path, but one firing state names neither of its two causes

`scripts/check-onbox-register.mjs:1063-1071` (the message) · `:1016-1028` (the cause)

Truth table, executed against the real `checkLiveView` with `direction: 'extraOnly'` and `--discharging E7`:

| working register | `origin/main` baseline | live page | error fires? | remedy true? |
|---|---|---|---|---|
| lacks E7 | lacks E7 | has E7 | no | n/a — correct; this is the state pass 1's old text wrongly named |
| **has E7** | has E7 | has E7 | **yes** | ✅ *"still IN this register"* |
| lacks E7 | has E7 | has E7 | no | n/a — the documented `--discharging` case, works as designed |
| **has E7** | has E7 | **lacks E7** | **yes** | ✅ *"not on the live page at all"* |
| lacks E7 | lacks E7 | **lacks E7** | **yes** | ✅ typo |
| **whole Group E absent from register AND baseline** | — | **has E7** | **yes** | ❌ **neither** — E7 *is* on the live page and is *not* in the register |

The last row is a whole-group discharge that has already merged to `origin/main` while the published page is still stale. The per-row path deliberately consumes a named id **before** the baseline filter (`:970-977`: *"a name for a row the baseline ALSO already lacks (harmless …) still counts as a match, not a typo"*); the whole-group path at `:1016` consumes only inside `if (baselineBodyGroups.has(letter))`, so the same harmless name is refused there instead. Pre-existing — #2272 fixed the per-row site and left its sibling armed — but this PR rewrote the message that has to explain it, so it is in scope this round. Effect: a hard refusal whose two named causes are both demonstrably false, in the one state where the flag was unnecessary anyway.

**Cleanup-only.** Either extend the sentence, or move the `namedIds` consumption above the `baselineBodyGroups.has(letter)` guard so both paths share one invariant.

---

### 🟡 Minor — the replacement within-group duplicate test restores half of what the deleted one pinned

`scripts/tests/check-onbox-register.test.mjs:254-263`

The deleted test (base `:166`) pinned two properties: a same-group duplicate **is** reported, and it is reported **without** a spurious count mismatch alongside — it built `bodyARows: [1, 1]` against `tableA: 2` deliberately, so check 1 was satisfied and could be asserted absent. The replacement uses `glance: [['E', 1]]` with `rows: [6, 6]`, so the counts no longer agree and the second property is unobservable. Executed:

```
new fixture (glance E:1, rows [6,6]):
  ["Group E: glance table says 1, body has 2 rows (E6, E6). Update the table or the body.",
   "Row ID E6 appears more than once (2 headings). …"]

old-shape fixture (glance E:2, rows [6,6]):
  ["Row ID E6 appears more than once (2 headings). …"]
```

The behaviour still holds — the second run is the proof — but nothing pins it any more. Setting the glance count to `2` and adding the `!errors.some(e => e.includes('glance table says'))` assertion restores it in two lines. **Cleanup-only.**

---

### 🟡 Minor — "no mechanical check can catch it after the fact" is overstated, and disagrees with this repo's sibling checker

`docs/testing/onbox-acceptance-register.md:323-325`

`check-register-citations.mjs`'s **Check C fatal half** exists for exactly this shape — its own header calls it *"the dangerous case where a stale citation still looks superficially valid because the ID happens to exist, just for something else now"* — and it gates. Its own runtime output on this tree reports measured coverage of *"22 anchored-heading line(s) in 5 file(s)"*. So a re-minted retired ID **is** caught mechanically and fatally whenever the citation sits on an anchored `### <ID> · …` heading or a `Criteria source:` line carrying a subject number that still maps to a current row. The parenthetical scopes the claim to "the checker" (singular, `check-onbox-register.mjs`), where it is true; the unqualified clause preceding it is not. Worth one clause, because the checker's own header paragraph (`:17-26`) already states the same limitation correctly scoped and the two now disagree in strength. **Cleanup-only.**

---

### 🟡 Minor — bookkeeping the fix commits left behind

- `docs/superpowers/plans/2026-08-27-onbox-register-stable-row-ids.md:665` and `:780` still declare `parseAllRowHeadings(strippedText) -> Array<{ id, letter, number }>` and carry the three-field reference implementation. `4bc6116f` narrowed the shipped function to `{ id }`; the plan is a live working plan (unchecked task boxes), not an archived record, so its Interfaces block is now false for the next agent briefed from it.
- `formatRowList`'s `new Set(sorted).size === sorted.length &&` conjunct (`scripts/check-onbox-register.mjs:204`) is now dead: under the strict `n === sorted[i - 1] + 1` chain a duplicate always breaks the run, so the guard can never change a result. Removing it leaves **116/116** green — mutation-verified. It was load-bearing only under the old `n === i + 1` form this PR replaced, i.e. this change orphaned it.
- The PR body's **Test plan** is stale against `HEAD`: it states **112** register tests and **1604** `test:hooks`; the actual figures at `5c89c25e` are **116** and **1608**. Its mutation table lists seven mutations and omits the three the fix commits added (floor literal, `formatRowList` range, `parseNextIdMarker` anchor).

**Cleanup-only.**

---

### Verdict

Six of the seven pass-1 findings are genuinely closed, and the three mutation-backed ones each kill their mutation *alone* — the discipline held. Finding 4's re-dating is correct history: both sentences keep their events attributed to their own dated correction headers (2026-08-21), "pre-#2599 positional-ID rule" is an accurate frame, and nothing is misdated. Finding 7's two surfaces agree with each other and with 4b's actual behaviour.

The blocker is finding 2's fix. It restored the missing rule and then illustrated it with `A38` — the exact ID pass 1 used as its repro, the exact ID one of the 16 Check A notes says does not exist, and a marker value the new floor rejects. That is a one-line edit. Everything else here is cleanup that would not, on its own, hold the merge.

**Not mergeable as-is** — one 🟠, four 🟡.

dudarenok-maker:
## PR review — pass 3 (head `4c449a90`, depth high)

Scope: commit `4c449a90` and its blast radius, not a third full sweep. Verified before probing: `test:hooks` **1608/1608**, `check-onbox-register.test.mjs` **116/116**, `lint` 0, `check:onbox-register` green, `check:register-citations` green with Check A's annotated note count at **16**, retro-application of the new `checkRegister` over `git show 9237ce78:docs/testing/onbox-acceptance-register.md` → **0 errors**.

| # | Finding | Disposition |
|---|---|---|
| 1 | Allocation example minted retired `A38` | **VERIFIED RESOLVED** |
| 2 | `--discharging` remedy missed the whole-group-already-merged state | **VERIFIED RESOLVED** |
| 3 | Duplicate test's glance/row counts disagreed, so 4a wasn't asserted alone | **VERIFIED RESOLVED** |
| 4 | "no mechanical check can catch it after the fact" overstated | **STILL OPEN — replaced with a different overstatement** (below) |
| 5 | Plan-doc signature + dead `formatRowList` clause | **VERIFIED RESOLVED** |

Evidence for each resolved item is under "What is solid".

### 🟠 Significant — the softened Check C claim names the wrong half of Check C, and the wrong limitation

`docs/testing/onbox-acceptance-register.md:325-327`

> `check-register-citations.mjs`'s Check C
> fatal half does catch this shape, but only on its measured coverage — 22
> anchored-heading lines across 5 files, not the whole corpus.

"This shape" is, unambiguously from the two preceding sentences, **re-minting a retired ID** — "even one whose row is long gone". Check C's **fatal** half (`wrongId`) does not catch that. It routes to `unknownSubject`, which is opt-in behind `--strict` and **never fatal**.

The discriminator in `recordSubjectConflict` (`scripts/check-register-citations.mjs:1427-1451`) is *"does the register still know this subject at all"*. When you re-mint a retired ID, the original row is gone, so its subject has left the register — the `!legitimateIds` branch fires, not the `wrongId` branch. `wrongId` needs the subject to still be live under a **different** ID, which is the renumbering class this very PR abolishes.

**Concrete repro** (register where `A38`'s original row for `#900` discharged and `A38` was re-minted for `#901`; citation is an anchored heading carrying its subject on the same line, i.e. *inside* the 22-line measured coverage — `measureWrongIdEligibleLines` → `{headingLines: 1, headingFiles: 1}`):

```
wrongId (FATAL):        []
unknownSubject (--strict, never fatal):
  ["... cited A38 for #900, but #900 does not appear in any current
    register row heading (its row may have discharged) — verify A38 still applies"]
```

Control — same fixture, but `#900` still live at `A1`:

```
wrongId (FATAL):
  ["... cited A38 for #900, but the register's #900 maps to A1, not A38"]
```

Two independent confirmations from the checker itself, so this is not my reading of the code:

- `scripts/check-register-citations.mjs:202-210`: *"a wrong ID that arrives via a DISCHARGE … routes to the non-fatal, opt-in `unknownSubject` bucket by construction, **not** `wrongId` … Widening `wrongId` to that class is deliberately deferred (see #2629)."*
- The success line this branch actually prints: *"A wrong ID that arrives via a DISCHARGE … is **also not caught by the fatal half** — it lands in the exploratory, non-fatal 'unknownSubject' bucket instead (--strict only), by construction."*

Two aggravating details:

1. **The qualifier points at the wrong gap.** "22 anchored-heading lines across 5 files" is *correct* as a live figure (verified — see below), but it implies the only limitation is breadth. The operative limitation is severity: even a citation squarely inside those 22 lines produces nothing on a default run.
2. **The one real retired-ID citation in this tree isn't even on that surface.** `docs/testing/onbox-sitting-plan.md:74` cites `A38` in the prose idiom inside a blockquote — a Check A surface, which Check C's fatal half never covers on any line.

**Why this blocks rather than being a nit:** the PR body's own "What 4b does not do, stated plainly" says the guard for this failure is the written procedure, and that "both the checker's header and the register's own allocation guidance say so **rather than implying the checker has your back**." After `4c449a90`, the register's allocation guidance implies exactly that — about the one failure mode the paragraph exists to deter, in the governing maintenance doc, pointing at detection deferred under the very issue (#2629) this PR closes. Pass 2's finding 4 was right that the original overstated; the replacement overstates in the more dangerous direction, since the original at least erred toward "nothing will save you."

**CONFIRMED · defect (not cleanup) · one-sentence prose fix, no code surface.** Something like: the fatal half catches only the *renumbering* variant (subject still live under a different ID); a re-mint after discharge lands in the non-fatal, `--strict`-only `unknownSubject` bucket by construction, so the written procedure is the whole guard.

Do **not** "correct" 22 → 23 while editing: 22/5 is the live measurement (`check:register-citations` prints "22 anchored-heading line(s) in 5 file(s)"); the `23` at `check-register-citations.mjs:188` is the explicitly-frozen historical number that same comment tells you not to trust.

### 🟡 Minor — pre-existing, no action asked

Naming an already-merged discharge in `--discharging` fails loudly for a **whole-group** discharge but is silently consumed for a **per-row** one (verified both). Both fail safe — nothing stale slips through either way — and it predates this commit (#2272 behaviour, untouched here). Noting it only so the asymmetry isn't rediscovered as a bug later.

### ✅ What is solid

- **Allocation example (finding 1).** Group A's real marker is `<!-- next-id: A101 -->` (`:568`); `A101` appears in no heading anywhere; Group A tops out at `A37`. Applying the example verbatim to the real register — mint `A101`, bump to `A102`, bump glance `A:37→38` and total `60→61` — yields `checkRegister` → `[]`. The old `A38`/`A39` example, applied the same way, yields *"Group A's next-id (A39) is below the allocation floor A101"* — the blocker was real and is genuinely gone.
- **`--discharging` third branch (finding 2).** Constructed the whole-group-already-merged state against the real `checkLiveView`: the new branch fires and describes it accurately. Its parenthetical is verified too — the same state with no `--discharging` returns `[]`, so that group's stale live-page section really is already a silent no-op. Both pre-existing branches re-verified true (typo → fires; still-in-register → fires), and #2272's happy path still returns `[]`.
- **Duplicate test (finding 3).** Mutated 4a's trigger `count > 1` → `count > 999`: the re-fixtured test goes red with `actual: []` — no masking error at all, proving `glance E:2` genuinely satisfies check 1 and that `assert.deepEqual` pins 4a firing alone. Strictly stronger than the `.some()` it replaced.
- **`formatRowList` clause removal (finding 5).** Old vs new compared over 17 hand cases plus **19,600** exhaustively-enumerated multisets (len 2–5 over 0–6): **zero** divergences. The `prev + 1` chain does reject duplicates on its own (a duplicate gives `n === sorted[i-1]`, which can equal `sorted[i-1] + 1` only if `0 === 1`), and row numbers are `\d+` so no negative/fractional escape exists. Behaviour-preserving.
- **Nothing collateral broken.** `4c449a90` left the register's row IDs, `next-id` markers, per-group glance counts and owed total byte-identical (`A:37,B:2,C:4,D:3,E:10,G:2,H:2`, total 60 — compared `4c449a90^` vs `HEAD`). `docs/testing/onbox-acceptance-register-live-view.html` is absent from the whole-branch diff, so "no republish is owed" holds.
- **PR body checks out.** 116 tests, 1608 `test:hooks`, 11-row mutation table, lint 0, Check A note count 16, no-republish — all verified. `Closes #2629` / `Refs #2599` / `Refs #2603` present and unfenced.
- Global Constraints 2 (no hardcoded row numbers — every fixture builds its own text), 3 (floor 101, pinned by its own test), and 4 (no existing row renamed or renumbered) all hold.

### Verdict

One 🟠 correctness finding stands, in prose rather than code: `docs/testing/onbox-acceptance-register.md:325-327` promises a fatal mechanical check for re-minting a retired ID, and no such fatal check exists for that shape. Everything else `4c449a90` set out to fix is verified fixed, and nothing it touched regressed.

This is pass 3 of 3 — the loop cap. My recommendation: correct that one sentence and merge without a further review round, since the fix is a prose correction with no new code surface for a fourth pass to interrogate. **Mergeable once line 325-327 is corrected; not as it stands.**

dudarenok-maker:
## PR review — pass 4 (head 85612b7a, depth high)

Past the 3-pass cap, explicitly authorised: this paragraph has been wrong in three consecutive rounds. Scope narrowed by dispatch to **`7d44d847` + `85612b7a` only** — the allocation-procedure blockquote's closing claim about which checks can catch a re-minted retired row ID. The rest of the branch is settled and was not re-swept.

Verified before probing: `check:onbox-register` OK · `check:register-citations` OK (Check A already-annotated notes = **16**, Check C fatal = 0, `--strict` = 6 non-fatal warnings, exit 0) · `test:hooks` **1608/1608**, exit 0 · `7d44d847^..HEAD` touches one file, all changed lines are blockquote prose, and row IDs / glance table / `next-id` markers / owed counts are byte-identical across the range.

Claims were checked against `scripts/check-register-citations.mjs`'s code, not its header comment.

### 🟠 Significant — "No check fails on it" is false; the clause licensing it equates "row discharged" with "subject gone"

`docs/testing/onbox-acceptance-register.md:323`, licensed by the clause at `:332-333`.

**Defect.** CONFIRMED, reachable on today's corpus.

`recordSubjectConflict` (`scripts/check-register-citations.mjs:1427-1451`) branches on `legitimate.get(subject)`, and `buildLegitimateSubjectMap` (`:1280-1289`) maps one subject to a **set** of IDs. So a subject that spans several rows survives the discharge of any one of them — and the paragraph's "if the subject is gone entirely, **which is what a discharge means**" is false exactly there. That false equation is what licenses the bolded "No check fails on it."

This register has many multi-row subjects: #2192 → {A28, A29, A30, E7}; **#2040 → {A22, A23, A34}**; #1976 → {A19, A35}; #2256 → {H1, H2}.

Concrete failure scenario, live today:

1. `docs/testing/onbox-sitting-cloning-identity.md:195` is `### A22 · Cast/analysis characterId drift — Wave 1 resolver ([#2040](...))` — an anchored heading carrying its subject on the same physical line, in a non-frozen file, i.e. one of the 22 fatal-eligible lines the checker's own success line reports. It is silent today, correctly.
2. Discharge **A22** and annotate that heading per "annotate, don't renumber" → Check A downgrades it to a non-fatal NOTE. Tree green.
3. Re-mint **A22** for a new row. `check-onbox-register` permits this: unique under 4a, and `22 < A101` under 4b — its own comment at `scripts/check-onbox-register.mjs:388-393` says so.
4. `checkConflictingSubjects` now sees `registerRows.has('A22')` true and `legitimate.get(2040)` = `{A23, A34}` → **FATAL `wrongId`, by default, no `--strict`**:

```
docs/testing/onbox-sitting-cloning-identity.md:195 — cited A22 for #2040,
  but the register's #2040 maps to A23/A34, not A22
```

The re-mint is precisely what flips it from green to red: before it, Check C skips a nonexistent ID (`:1489`, `:1536`) and Check A's annotation keeps it non-fatal. Same shape is live for #1976 via `onbox-sitting-vram-contention.md:225`.

**Why this matters, not just that it's inexact.** This is pass 3's harm from the other side. An operator who hits that fatal error and consults this paragraph is told (a) re-minting cannot fail a check, and (b) the fatal half is only ever "the renumbering class this design abolishes". Both point at *fix the citation* — edit the historical heading to A23/A34 — which destroys the discharge annotation and silently re-points a historical citation: the exact corruption this stable-ID design exists to prevent. The correct action is to undo the re-mint.

**Fold into the same fix, not a separate finding:** "a citation whose subject has left the register entirely **lands in** Check C's `unknownSubject` bucket" also over-states, though in the safe direction. That only happens on Check C's two surfaces with a subject number on the same line; a `row A22` prose citation or a `Register row(s): A22` label line fires *nothing at all*, in either bucket. Since the sentence has to be restated anyway, "at best lands in" is the accurate form.

### 🟠 Significant — `7d44d847` parks an open deferral on #2629, the issue this PR closes; the same file already says #2629 shipped

`docs/testing/onbox-acceptance-register.md:329`, plus the same defect's two other sites at `scripts/check-register-citations.mjs:209-210` and `:26-37`.

**Defect (staleness).** CONFIRMED — and already self-contradictory in the tree, not merely a post-merge prediction.

`7d44d847` newly added "— deliberately, and deferred under **#2629**" at `:329`. Sixty-two lines later, this same branch asserts the opposite at `:391-393`:

> **(#2629 has since shipped: the contiguity check is gone, replaced by allocation-floor checks, and Group A's inline caveat comment was removed — every group's `next-id` marker allocates normally now.)**

The PR body says `Closes #2629`. On merge, three live deferrals point at a closed issue:

1. `onbox-acceptance-register.md:329` — `unknownSubject` staying non-fatal;
2. `check-register-citations.mjs:209-210` — "Widening `wrongId` to that class is deliberately deferred (see #2629)";
3. `check-register-citations.mjs:26-37` — the dedicated CI step, "tracked at `#2629`".

Failure scenario: someone hits an `unknownSubject` warning under `--strict`, reads "deferred under #2629", opens #2629, finds it **CLOSED as delivered by this PR**, and concludes the widening was considered and settled — when it was never done. That is this PR's own citation-rot class, one level up: the PR closes the ticket its own deferrals cite.

The deferral needs a home that outlives the merge — a successor issue referenced from all three sites, or `Refs #2629` instead of `Closes`. #2629's body frames the open half as a design decision with three options, so the widening itself may legitimately be the design-pass carve-out; but then it needs its own issue **naming that decision**, not a pointer at a closed one. The re-pointing is bookkeeping and fixable this round.

### ✅ What is solid

- The discriminator's **first** half is now mechanically exact and matches `recordSubjectConflict`'s actual branching, not just the header comment — a genuine improvement on pass 3's version, and the specific thing `85612b7a` set out to fix.
- "opt-in behind `--strict` and never fatal" — verified true at `:1749` and `:1778-1783`, and by a live run (6 warnings, exit 0).
- "only verifies uniqueness and the allocation floor, not history" — verified true; `check-onbox-register.mjs:388-393` independently says the same ("They do NOT stop someone hand-typing a discharged row's old ID back in").
- Mechanically clean: one file, blockquote prose only, structural content byte-identical, all three gates green.

### Verdict

**Not mergeable as-is.** Two correctness defects, both in the two commits under review, both small targeted text fixes plus one issue filing. Neither is blocking-severity and the rest of the branch is settled — but finding 1 is a false mechanical claim in the exact sentence this PR is rewriting for the fourth time, and finding 2 is a contradiction already present in the file.

The 3-pass loop cap is already exceeded, so this routes to the user rather than auto-looping.

Issue #2696, opened 2026-08-27
[agent instructions][claude][task] Fix Qwen length-determinism test premise in test_qwen_duration_golden.py (#1994)


dudarenok-maker:
## What

The Qwen duration golden-audio baseline scaffold (#1994, shipped by #2659, verified by #2660) shipped a determinism sanity test whose core assertion does not hold on real hardware. Fix it.

## Where the work is

Worktree: `C:\Claude\Projects\wt-1994-qwen-duration-baseline`
Branch: `test/sidecar-1994-qwen-duration-baseline`

`cd` into that worktree. Do not create a new worktree. Do not work in the primary checkout.

## What's wrong (reproduced twice, not hearsay)

`server/tts-sidecar/tests/golden/test_qwen_duration_golden.py::test_qwen_is_deterministic_in_length` asserts that two consecutive `QwenEngine.synthesize()` calls on the same model/voice/text produce the SAME `sample_count`. On this box — which DOES have a designed Qwen voice, contradicting #2660's own checklist assumption that it wouldn't — the test actually runs (does not skip) and fails reproducibly (ran twice, same shape of result both times):

```
tests\golden\test_qwen_duration_golden.py::test_qwen_is_deterministic_in_length FAILED
    assert measure_pcm(a.pcm, a.sample_rate)["sample_count"] == measure_pcm(b.pcm, b.sample_rate)["sample_count"]
E   assert 86400 == 92160
```

That's roughly a 6.7% length difference between two back-to-back synths of the identical input. This is consistent with — not a contradiction of — the file's own documented understanding that "Qwen decoding is stochastic with no seed" (see `_bless`'s docstring and `qwen-duration-fixture.json`'s `_comment`); the test's exact-equality assertion contradicts that documented stochasticity.

## What to do

Pick a fix that is honest about the stochasticity rather than one that makes the test vacuously pass:

- Most likely correct: drop `test_qwen_is_deterministic_in_length` entirely, or replace it with something that asserts a bounded relationship (e.g. within the same duration `tolerance` the main golden assertion already uses) if there is a real invariant worth checking — the rest of the file already treats length variance as expected and handles it via `tolerance`, not exact equality.
[...]

PR #2739, opened 2026-08-28
test(sidecar,docs): measure real Qwen duration tolerance, discharge A101

### LINKED ISSUE #1994: ops — Qwen has no golden duration baseline, so a speech-rate regression on the DEFAULT engine is invisible
## What

The golden-audio tier's per-line assertions — duration vs `kokoro-baseline.json`, `MIN_RMS`, no-voice-substitution, and (with #1911) content/WER — all cover **Kokoro**, which is the *fallback* engine.

**Qwen, the default engine that actually renders books, has no golden baseline at all.** Its only golden coverage is `server/tts-sidecar/tests/golden/test_cross_engine_sanity.py`, which is:

- opt-in behind `GOLDEN_QWEN_VOICE=<voiceId>`, so it skips by default (confirmed: a full `-m golden` run on this box skips all four cross-engine cases);
- a wire-format sanity check, not a timing or content assertion.

So a Qwen speech-rate change, prosody drift, or pronunciation regression is invisible to every automated gate in the repo.

## How it surfaced

The repo owner reported that a rendered chapter "sounds a bit rushed — speed of speaking somewhat faster than I remember." Running the golden tier answered it for Kokoro (duration assertion ran and **passed** against the 2026-06-05 baseline, so no drift, and the shared audio path — sample-rate handling, resampling, assembly — is therefore also clear).

But they had been listening to Qwen, and nothing measures Qwen's timing. The observation could not be confirmed or refuted on the engine it was about.

Separately: comparing real renders of identical text across dates showed **±1–5% run-to-run variation** in chapter duration, occasionally ~8%, in *both* directions across languages (EN/DE shorter, ES/FR longer through the same re-render). That is large enough for a listener to notice and is currently unmeasured — so there is no way to distinguish "normal stochastic variation" from "a real regression" without a baseline.

## Acceptance

- A Qwen per-line duration baseline in the golden tier, on the same recorded-baseline pattern Suite A already uses for Kokoro.
- It runs without an opt-in env var (or the env var is set by default on a box that has designed voices), so it actually fires.
- Tolerance chosen from a **measured** distribution, not picked — Qwen's decoding is stochastic, so the per-line spread must be measured across N runs first and the tolerance derived from it. Record N and the observed spread.
- Paired mutation-verified tests, each mutating the producer.

## Key files

- `server/tts-sidecar/tests/golden/test_cross_engine_sanity.py` — the current opt-in Qwen check
- `server/tts-sidecar/tests/golden/test_golden_regression.py` + `kokoro-baseline.json` — the Kokoro pattern to mirror
- `scripts/run-golden-audio.mjs` — the runner and its `--bless` paths

## Notes

- Depends on nothing, but reads best after #1911 lands (it establishes the recorded-baseline + bless-guard pattern this would reuse).
- Note #1950: no CI workflow runs the golden tier or the sidecar pytest suite at all, so any gate added here fires only on a local run on a box with a bootstrapped venv.

## Benefit

*Benefit (user):* the engine that renders every book is the one engine with no timing or content regression net. A prosody or rate change on Qwen currently reaches the listener before it reaches a test.

Found while investigating a listener-reported speed change. Refs #1911 · Refs #1950

dudarenok-maker:
## Summary

- Discharges on-box acceptance register row **A101** (#1994): `qwen-duration-baseline.json`'s `tolerance` was a placeholder (0.10) pending a real on-box N-repeat measurement of Qwen's per-line duration spread — Qwen decoding is stochastic with no seed, so run-to-run duration varies.
- Ran a new ad-hoc measurement tool, `server/tts-sidecar/tests/golden/measure_qwen_duration_spread.py`, on an RTX 5070 Ti (`QWEN_DEVICE=cuda:1`, this box's faster of two GPUs) against the designed voice `cw_gpu_17b`: N=10 repeated real-model syntheses per `qwen-duration-fixture.json` line.
- Observed per-line max fractional deviation from its own mean: `narrator-plain` 9.0%, `numbers-and-year` 12.2%, `abbreviations` 22.4% (the observed max), `quotes-and-punctuation` 13.0%, `multi-sentence-group` 19.9%.
- Hand-set `tolerance` to **0.30** (measured max + ~34% headroom), then blessed `entries` from a fresh single run against the same voice/model, and re-ran the assertion path (not `--bless`) once to confirm it passes against the new baseline.
- Raw measurement retained at `server/tts-sidecar/tests/golden/spread-report.json` for the record.
- Register row A101 removed (61 → 60 owed); its ID stays retired per the register's stable-ID (`next-id`) convention — never reused. Live-view HTML mirror, `check:onbox-register`, and `check:register-citations` all updated/green.

Test-tier + register bookkeeping only — no production code changed.

## Test plan

- [x] `test_qwen_duration_bless_gating.py` (5 tests) — pass
- [x] Fast `-k qwen_duration` tier (9 tests, `-m "not golden"`) — pass
- [x] Real golden-audio assertion run (`-m golden -k qwen_golden_lengths_match_baseline`, `QWEN_DEVICE=cuda:1`, voice `cw_gpu_17b`) against the freshly blessed baseline — pass
- [x] `npm run check:onbox-register` — agree
- [x] `node scripts/check-register-citations.mjs` — no unannotated dangling A101 citations
- [x] `npm run verify:fast:branch` (pre-push) — green

Closes #1994


dudarenok-maker:
## PR review — pass 1 (head 6ca9f70bc8b12e9f26e9648cf053a89bf9634604, depth high)

Reviewed at `6ca9f70b` in a detached worktree. Verified by running, not by reading: `node scripts/check-onbox-register.mjs` (exit 0, "agree"), `node scripts/check-register-citations.mjs` (exit 0; the five new `A101` citations are correctly downgraded to annotated NOTEs), the fast `-k qwen_duration` tier (9 passed), a manual per-group row count of the register at head (A=37, B=2, C=4, D=3, E=10, G=2, H=2 → 60, matching the glance table and the stated total), and a stub-engine replay of the new script's `measure()` against the committed `spread-report.json` (reproduces every mean/stdev/`max_frac_dev` exactly; 51 synthesize calls = 1 warm-up + 5×10). `next-id: A102` correctly left un-rewound.

The register/live-view bookkeeping is clean. The tolerance derivation is not.

---

## Correctness bugs

### C1 — The tolerance is derived from the wrong statistic, and the PR's own measurement contains draws that would fail the gate it justifies

`server/tts-sidecar/tests/golden/measure_qwen_duration_spread.py:89` and `:105-109`; `server/tts-sidecar/tests/golden/qwen-duration-baseline.json:3`.

The script computes `max_frac_dev = max|d − mean| / mean` and then prints `Suggested tolerance (overall max + headroom, e.g. x1.3-1.5)`. The assertion it feeds is not a deviation-from-mean test. It is `compare.py:117-122`'s `_within(fresh_sample_count, blessed_sample_count, tol)` — `|fresh − blessed| / blessed ≤ tol` — where `blessed` is **one draw** recorded by `_bless` (`test_qwen_duration_golden.py:206-216`), not a mean. For a single-draw-vs-single-draw comparison the governing statistic is the pairwise ratio, which is up to ~2× the deviation-from-mean.

Concrete failure, using only values this PR's own `spread-report.json` observed:

| line | observed min | observed max | pairwise (max/min − 1) |
|---|---|---|---|
| `multi-sentence-group` | 3.76 s | 5.36 s | **42.6 %** |
| `abbreviations` | 3.76 s | 5.12 s | **36.2 %** |
| `quotes-and-punctuation` | 3.28 s | 4.24 s | **29.3 %** |

`_bless` recorded 4.80 s for `multi-sentence-group`. Had it recorded the 3.76 s draw — a value the same 10-run measurement produced — the very next assertion run drawing 5.36 s gives `|5.36 − 3.76| / 3.76 = 42.6 % > 0.30` → a red golden test with nothing regressed. Same shape on `abbreviations` (36.2 %). Whether 0.30 holds is decided by which draw the bless happened to land on, which is exactly the "picked, not derived" outcome #1994 was filed to prevent.

The margin is also thinner than the "22.4 % + ~34 % headroom" prose implies everywhere it is quoted. `quotes-and-punctuation` is blessed at 4.24 s, which is the **maximum** of its own 10-sample distribution (+13.0 % above the mean of 3.752 s). A fresh run at 3.28 s — observed twice in those 10 draws — already consumes **22.6 %** of the 30 % budget. Worst-case budget consumption against the *actual* blessed values, from the committed data: 12.8 % / 10.9 % / 19.0 % / 22.6 % / 21.7 %. Fitting a normal to each line's committed `pstdev`, the shipped baseline carries roughly a **~2 % chance per run** of a spurious red across the five lines (`abbreviations` z=2.43, `quotes` z=2.44, `multi-sentence-group` z=2.68 to the lower bound).

Fix the same data already supports: record the **mean of N draws** in `entries` and assert a fresh **mean of k draws** against it — the standard error falls as √k, so these pstdevs (0.14–0.50 s) support a ~3–6 % tolerance rather than 30 %. If a single-draw baseline is kept, derive the tolerance from the observed pairwise range (42.6 %), not from deviation-from-mean, and correct the 22.4 % justification in all five places it is now recorded (baseline `_comment`, register, live view, release notes, `_bless` docstring).

### C2 — At 0.30 the gate cannot detect the regression #1994 was filed to catch

Same lines. #1994's body states the trigger ("sounds a bit rushed — speed of speaking somewhat faster than I remember") and the noise floor it wants to distinguish a regression from ("±1–5 % run-to-run variation in chapter duration, occasionally ~8 %"). A ±30 % per-line window is far above both.

Concrete: a uniform 15 % speech-rate regression (every line renders at 0.85× its natural duration) lands at 20.0 % / 13.8 % / 23.4 % / 24.8 % / 16.8 % off the blessed values — green on all five lines, comfortably. A uniform **20 %** regression lands at 24.7 % / 18.9 % / 27.9 % / 29.2 % / 21.7 % — still green on all five in expectation. So the gate is now armed and firing, and blind to its own feature: the shape it exists to catch passes.

C1 and C2 share one root cause (single-draw baseline vs. single-draw assertion on a high-variance stochastic process) and one fix (bless a mean of N, assert a mean of k). Reporting them separately because they fail in opposite directions and a fix that only widens or only tightens the tolerance addresses one and worsens the other.

### C3 — This PR turns the Qwen golden gate from "skips on a foreign box" into "hard-fails on any box whose first designed voice isn't `cw_gpu_17b`"

`server/tts-sidecar/tests/golden/qwen-duration-baseline.json:6,12,18,24,30` (each entry's `"voice": "cw_gpu_17b"`); `test_qwen_duration_golden.py:129-136` (mismatch → `failures.append`); `test_qwen_duration_golden.py:257-266` → `prereq.py:98-107` (`return voices[0]`); `main.py:6385-6395` (`list_voices` = `sorted(os.listdir(voices/qwen))`).

Qwen voices are per-workspace bespoke and the runtime pick is "first of the sorted list". Before this PR, `entries` was empty and every box SKIPped (line 91). Now: design any Qwen voice whose id sorts before `cw_gpu_17b` — anything starting with a digit, or a letter before `c`, or `cw_gpu_04a` — and the next `npm run test:golden-audio -- --sidecar-only --engine=qwen` on **the same box** reports a hard assertion failure on all five lines ("baseline was recorded against voice 'cw_gpu_17b', this run resolved '…'") with nothing regressed. Every other box is in that state from the first run.

The file already has the right precedent for this class one branch up: an inapplicable baseline SKIPs (`if not entries`, line 91-99). "The baseline was recorded against a voice this box does not resolve" is the same class — the baseline does not apply here — not a regression. Either SKIP with a loud reason, or have `_resolve_voice` prefer the baseline's recorded voice when `list_voices()` contains it.

---

## Cleanup / nits

### N1 — `test_qwen_duration_golden.py:18-22`: module docstring still describes the pre-PR state

> "The baseline ships UNBLESSED (empty `entries`, placeholder `tolerance`) — the real N-run stochastic-spread measurement and tolerance derivation are an on-box acceptance register row, not this file's job."

All three clauses are false as of this diff, and the register row was deleted in it. This is the first thing a reader of the file sees.

### N2 — `test_qwen_duration_golden.py:93-98`: the SKIP message points at a deleted row

> "Its tolerance is a placeholder — bless … (see the Group A on-box register row for the required measurement)"

A101 is retired and never reused, so this instruction resolves to nothing. Point it at `spread-report.json` / `measure_qwen_duration_spread.py` instead.

### N3 — `test_qwen_duration_golden.py:246-252`: the #2696 comment's premise is now false

> "There is no blessed baseline yet to derive a real bound from (`qwen-duration-baseline.json` ships with empty `entries` -- see its own `_comment`), so a bounded same-line check has no measured spread to bound against."

There is a blessed baseline, and the measured spread it says does not exist is now committed at `spread-report.json`. The block also cross-refs `_bless`'s docstring for the "widened to 0.10" rationale that this PR deleted from that docstring. The block's *conclusion* (don't reintroduce an exact-equality determinism check) survives; its stated grounds do not.

### N4 — `test_qwen_duration_bless_gating.py:51,55`: stale comment in a file this PR edited

```
"""Create a fresh unblessed qwen-duration-baseline.json with placeholder tolerance."""
...
"tolerance": 0.10,  # placeholder, matches the real committed baseline's current value
```

The committed baseline is now `0.3`. The comment is false about the exact value it claims to mirror, in a file this PR touched twelve lines below.

### N5 — 142 lines of new sidecar Python with no paired test

`measure_qwen_duration_spread.py` is the producer of the number this entire PR exists to justify, and nothing exercises it: `grep -rn measure_qwen_duration_spread server/tts-sidecar/tests/` returns only its own file and one docstring mention. CLAUDE.md: "Any new sidecar code MUST add cases here." It is trivially GPU-free to test — I did it during this review with a stub engine returning `SimpleNamespace(pcm=bytes(n*2), sample_rate=24000)`; it replays the committed report exactly and pins the 1-warm-up + 5×10 call shape. #1994's acceptance criterion 4 ("paired mutation-verified tests, each mutating the producer") is not met for the code this PR adds, which is worth weighing against the `Closes #1994` in the body — either land the test or downgrade to `Refs`.

### N6 — `narrator-plain` is measured warm but blessed cold

`measure_qwen_duration_spread.py:74` absorbs the model's cold first forward pass in a discarded warm-up, then measures all 10 reps warm. `_bless` does not: its `i == 0` call **is** the recorded entry, so `narrator-plain`'s blessed value is the cold first synthesis of the process. So the spread that justifies `narrator-plain`'s tolerance was not measured under the conditions `narrator-plain` is asserted under. Corroborating signal: `narrator-plain` is the only one of the five blessed `sample_count`s that is not a multiple of Qwen's 1920-sample frame (74881 = 1920×39 + 1; the other four are exact multiples), and it sits +6.3 % above its own measured mean. Make the two protocols agree — warm both or neither.

### N7 — the changelog entries drop the PR number

`docs/testing/onbox-acceptance-register.md:384` and `onbox-acceptance-register-live-view.html:201` read "**Last change: 2026-08-28, 61 → 60.**" while every prior entry in both files carries "(PR #NNNN)" — including the one directly below it. This is #2739.

### N8 — the live-view publish half isn't recorded

CLAUDE.md before-shipping step 3 requires publishing `onbox-acceptance-register-live-view.html` to the canonical artifact URL (`onbox-acceptance-register.md:28`) and running `npm run check:onbox-register -- --against-published <saved live page>` immediately before. The test plan lists neither. The diff cannot show it either way, and the register's own history records four incidents where exactly this half went wrong silently — worth confirming before merge.

### N9 — the script's documented usage writes the report to the wrong directory

`measure_qwen_duration_spread.py:17-20` tells the operator to run from the repo root with `--out spread-report.json`, which lands it in the repo root; the committed one is at `server/tts-sidecar/tests/golden/spread-report.json`. Default `--out` to `GOLDEN_DIR / "spread-report.json"`, or fix the example.

---

## House gates — checked

- Register / live view in sync and arithmetically consistent (checker green **and** hand-counted); `next-id: A102` correctly not rewound; A101 retired, not reused. ✅
- `check:register-citations` green; all five new A101 citations carry discharge annotations. ✅
- `docs/release-notes-next.md` entry present; `RELEASE_NOTES.md` skip stated explicitly in the body as test-tier/bookkeeping. ✅
- `Closes #1994` present, outside a code span. ⚠️ see N5 — criterion 4 is unmet for this PR's new code and C1 undercuts criterion 3 ("tolerance derived from a measured distribution, not picked").
- No `cast.json` writes, no config knob, no derived artifact, no run sheet for A101. N/A.
- No incidental fixes declared; none appear folded in. ✅
- Prose figures cross-checked against `spread-report.json`: every mean, stdev and percentage quoted in the register, live view, release notes and baseline `_comment` matches the raw report exactly. ✅

dudarenok-maker:
## PR review — pass 2 (head 094c4339c0ae3690e8cad07f10481671eb220bf6, depth high)

Reviewed at `094c4339` in a throwaway detached worktree. Verified by running, not by reading:

- `node scripts/check-onbox-register.mjs` → exit 0, "agree"; `node scripts/check-register-citations.mjs` → exit 0, all six `A101` citations correctly annotated. Both re-confirmed at this head.
- The fast `-k "qwen_duration or measure_qwen"` tier → **15 passed** (was 9 at `6ca9f70b`).
- **Five mutations against the new code** (M1–M5, below). Three survive.
- **The real golden assertion path**, independently, on this box's GPU 1 (RTX 5070 Ti, `QWEN_DEVICE=cuda:1`, voice `cw_gpu_17b`, `-m golden -k qwen_golden_lengths`) → **passes** against the freshly blessed baseline. The author's re-run claim is true.
- **23 fresh real Qwen syntheses of my own** against the same box/GPU/voice/model, to check whether the re-blessed baseline actually sits where the C1 fix says it does. It does not — see C1 below.

`_bless`'s averaging arithmetic is correct, and the committed baseline is genuine measured data, not fabricated: all five `sample_count`s are exact multiples of **384** (= Qwen's 1920-sample frame ÷ 5), which is precisely the signature of averaging five frame-aligned draws, and `duration_sec == sample_count / 24000` exactly in every entry. C3's behaviour change is also correct. The problem is that C1's premise does not survive measurement, and that the fix's own paired tests cannot fail for the mutations they name.

---

## Correctness bugs

### C1 — NOT RESOLVED. The blessed reference does not approximate the true mean, because on this box there isn't a stable one — and the re-bless left the gate *more* flake-prone, not less

`server/tts-sidecar/tests/golden/qwen-duration-baseline.json:3`; `test_qwen_duration_golden.py:182-195` (the `BLESS_REPS` rationale) and `:242-243`.

The fix's stated premise (commit message; `BLESS_REPS` comment; release-notes entry) is: *`tolerance` bounds how far a single fresh draw lands from the population's TRUE mean, so make the blessed reference approximate that mean by averaging 5 draws.* That is sound reasoning about a stationary distribution. It is not what this engine does.

**Evidence 1 — the two data files this PR commits disagree at ~4σ.** Blessed value vs. the same PR's own `spread-report.json` N=10 mean, per line, with the z of the difference computed against *both* samples' standard errors (`σ·√(1/5+1/10)`):

| line | spread N=10 mean | blessed (mean of 5) | shift | z |
|---|---|---|---|---|
| `narrator-plain` | 2.936 s | 3.248 s | **+10.6 %** | **+3.98** |
| `numbers-and-year` | 5.192 s | 5.216 s | +0.5 % | +0.14 |
| `abbreviations` | 4.184 s | 4.544 s | +8.6 % | +1.71 |
| `quotes-and-punctuation` | 3.752 s | 4.368 s | **+16.4 %** | **+3.50** |
| `multi-sentence-group` | 4.696 s | 5.200 s | +10.7 % | +1.85 |

All five positive (p = 1/32 on sign alone), two beyond 3.5σ. Two samples of one population do not do this.

**Evidence 2 — I reproduced the drift.** Same box, same GPU 1, same `cw_gpu_17b`, same model, 3 fresh draws per line:

| line | my fresh draws | fresh mean | vs. spread N=10 | vs. blessed | worst draw's share of the 0.30 budget |
|---|---|---|---|---|---|
| `narrator-plain` | 2.96, 3.52, 4.08 | 3.520 | **+19.9 %** | +8.4 % | **85.4 %** |
| `numbers-and-year` | 5.84, 5.60, 5.20 | 5.547 | +6.8 % | +6.3 % | 39.9 % |
| `abbreviations` | 5.68, 5.28, 4.24 | 5.067 | **+21.1 %** | +11.5 % | **83.3 %** |
| `quotes-and-punctuation` | 4.16, 3.84, 4.56 | 4.187 | +11.6 % | −4.2 % | 40.3 % |
| `multi-sentence-group` | 5.92, 5.84, 5.20 | 5.653 | **+20.4 %** | +8.7 % | 46.2 % |

Three independent samples (spread N=10, bless N=5, mine N=3) cluster around three different means, monotonically increasing. Duration is decoded-token count, not wall time, so GPU load cannot explain this — the drift is in the model's sampling, and it is between-session.

**Evidence 3 — the within-session spread the tolerance was derived from is itself not reproducible.** Focused N=8 on `narrator-plain`, the cheapest and *lowest-variance* of the five lines:

```
fresh draws (N=8) : [3.6, 3.52, 3.92, 3.92, 3.92, 3.04, 3.2, 3.6]
committed N=10    : mean 2.936, range 2.72-3.20, max_frac_dev 9.0%
mine              : mean 3.590, range 3.04-3.92, max_frac_dev 15.3%
```

Barely-overlapping ranges (Welch t = 5.5), and a within-session `max_frac_dev` **1.7× the committed figure**. `tolerance = 0.30` is `22.4 % × 1.34`, where 22.4 % is the max across five lines of one N=10 run. That input does not replicate.

**The concrete failure.** `compare_to_baseline` computes `|fresh_draw − blessed| / blessed`. After this fix, `blessed` is session A's mean of 5; `fresh_draw` is a single draw from session B, whose mean sits 7–21 % away. The measured statistic (a draw's deviation from *its own session's* mean) does not bound the computed one. Modelling per-run spurious-red probability against the committed spread report's own per-line distributions:

- **shipped baseline: 3.65 % per run** (`multi-sentence-group` 1.71 %, `quotes-and-punctuation` 1.52 %, `abbreviations` 0.46 %)
- if the blessed values equalled the measured means: 0.63 %

So the re-bless left the gate ~6× further from the fix's own stated goal, and worse than pass 1's ~2 % estimate for the single-draw baseline it replaced. Worst-case budget consumption against the *observed* spread-report draws rose on three of five lines (12.8/10.9/19.0/22.6/21.7 % → 16.3/12.6/17.3/24.9/**27.7** %), and my own 3-draw probe hit 85 % of budget on the first try.

This also undercuts the A101 discharge itself. The row came out on the claim that `tolerance` is "derived from a measured distribution, not picked" (`qwen-duration-baseline.json:2`, register, live view, release notes, `_bless` docstring). The distribution that was measured is not the one the gate computes against, and it does not reproduce on the same box a day later. Either the tolerance has to bound the *between-session* statistic (which the current measurement never characterised), or the assertion has to stop comparing across sessions.

### C3 — behaviour is fixed, but the fix's own coverage regressed: two mutations of the new check survive, one of which disables the entire golden gate

`server/tts-sidecar/tests/golden/test_qwen_duration_golden.py:113-121`; `server/tts-sidecar/tests/test_qwen_duration_bless_gating.py:283-311` and `:314-348`.

The behavioural half is right: the in-loop `failures.append` block is fully gone (no dangling references anywhere in the tree), the up-front check fires only when some entry carries a non-null `voice` *and* this box's resolved voice isn't among them, so a matching voice doesn't skip and a `voice`-less baseline doesn't skip. Confirmed by running.

But moving the check from "append a failure" to "raise `Skipped`" inverted the failure direction of its own tests, and nothing was added to compensate.

**M1 — drop the `is not None` filter on line 113.** `recorded_voices` becomes `{None}`, truthy, so *every* run skips.

```
....ss   [100%]   # 0 failures
SKIPPED test_assertion_loop_tolerates_missing_voice_key_in_baseline_entry
SKIPPED test_assertion_loop_warms_up_before_missing_entry_check
```

**The mutation survives.** `test_assertion_loop_tolerates_missing_voice_key_in_baseline_entry`'s docstring says verbatim: *"Kills the mutation where `if recorded_voice is not None and recorded_voice != voice:` loses its `is not None` guard."* It does not. The inner call now raises `Skipped`, pytest reports the outer test as SKIPPED, and the suite is green. Under the pre-fix code the same mutation produced an `AssertionError` and a real red. Collateral: it also silently disarms `test_assertion_loop_warms_up_before_missing_entry_check`, the regression test pass-3 review added. (That docstring is additionally stale — it quotes code this commit deleted.)

**M3 — weaken line 114 to `if recorded_voices:`.** Skips even when the voice *matches*.

```
...........   [100%]   # 11 passed, 0 failures
```

**The mutation survives with the full suite green**, and it makes the real Qwen duration golden gate SKIP on every box forever — including the blessing box — with no test able to notice. Root cause: **no test covers a baseline whose recorded voice matches this box's resolved voice**, which is the only configuration under which the gate ever actually asserts. Both existing tests use either a mismatch or a `voice`-less baseline.

**M2 — delete the block entirely.** This one does go red, but on `TypeError: '>' not supported between instances of 'MagicMock' and 'int'` from `compare.py:85` — unmocked plumbing in `test_assertion_loop_skips_on_voice_mismatch_against_baseline` (which since this commit no longer patches `synthesise_or_skip`), not the voice logic. Red for the wrong reason. Its `match="voice_a.*voice_b|voice"` regex is also loose enough that the `|voice` alternative matches the empty-`entries` skip and the `_resolve_voice` skip too.

Minimum to close: a matching-voice test that asserts the gate **runs** (does not skip); assert on skip *count* or wrap the inner call so a stray `Skipped` fails the outer test rather than propagating; tighten the regex to the mismatch-specific text; drop the `|voice` alternative.

### C4 (new) — `test_bless_averages_across_reps` cannot fail for the mutation it was written to catch

`server/tts-sidecar/tests/test_qwen_duration_bless_gating.py:172` and `:188`.

The fixture is `sample_counts = [24000, 26400, 21600]`, chosen (per its own comment) so the mean is a round 24000. But **24000 is also `results[0]`**. So:

**M4a — replace the averaging at `test_qwen_duration_golden.py:242-243` with `sample_counts[0]` / `durations[0]`** — i.e. bless one arbitrary draw, *literally the pre-fix behaviour C1 was filed against*:

```
......   [100%]   # 6 passed, 0 failures
```

**Survives.** M4b (`[-1]`) is correctly killed (`expected mean sample_count 24000, got 21600`), so the docstring's claim — *"Kills a mutant that blesses only the first/last rep"* — is half false, and the half that fails is the one that matters. Per CLAUDE.md's testing discipline, C1's paired regression test does not go red before the fix.

Fix is one line: reorder to `[21600, 24000, 26400]` (mean 24000, first 21600, last 26400), so neither endpoint equals the mean.

---

## Cleanup / nits

### M5 — `_line_stats`'s `abs()` is unpinned, and it's the direction that matters

`server/tts-sidecar/tests/golden/measure_qwen_duration_spread.py:88`. Dropping `abs()` from `max(abs(d - mean) for d in durations)` passes all five new tests, because every fixture in `test_measure_qwen_duration_spread.py` has its max *signed* deviation equal to its max *absolute* one (`[2,3,4]`, `[1,1,1,1,2]`, `[2,2,2]`, `[0,0]`). That mutation makes `max_frac_dev` blind to **short** renders — the direction a truncation or speed-up regression takes, and the direction of `multi-sentence-group`'s 3.76 s outlier that drove pass 1's C1. Add a downward-outlier case, e.g. `[1.0, 1.0, 1.0, 1.0, 0.2]` (mean 0.84; max signed dev +0.16, max abs dev 0.64).

### The register and the live view both still say the entries were blessed from a single run

`docs/testing/onbox-acceptance-register.md:395` and `docs/testing/onbox-acceptance-register-live-view.html:213`, identically: *"then blessed `entries` via a fresh single run against the same voice/model."* False as of this commit — they are means of five, and `metadata.bless_reps: 5` records it. This commit updated the "Last change" line of both files and left the sentence directly below it stale. These two surfaces are the durable record of the A101 discharge, and the live view is the file that gets published.

### #2742 (C2 deferral) — accepted, thinly, but it needs re-framing

The ticket does what CLAUDE.md requires of a design-pass deferral: it **names the decision** (assert-time `K`), gives candidate values with their cost, offers a genuinely different design (a Welch-style significance test rather than a flat band), and raises a second sub-decision (whether `BLESS_REPS` and `K` are one knob). "Pick K" alone would have been the void "it needs a judgement call"; the alternative-design axis and the operator-facing per-run cost of an opt-in tier are what carry it over the line. I am not filing this as a finding.

Two corrections owed to it, though: its premise sentence — *"now the mean of `BLESS_REPS=5` repeated real syntheses, per the C1 fix"*, presented as an estimate of the true mean — is what C1 above falsifies, and the real gap is larger and differently shaped than the ticket describes (between-session drift, not just single-draw noise). It is also labelled `type:chore` with no `moscow:`; a design-pass ticket that is "the whole deliverable for its round" isn't a chore.

### Pass-1 nit dispositions

| # | Status |
|---|---|
| **N1** module docstring "ships UNBLESSED" | **VERIFIED RESOLVED** — `:18-24` now describes the blessed state accurately. |
| **N2** SKIP message citing the deleted register row | **VERIFIED RESOLVED** — `:95-101` now points at the bless command, no row reference. |
| **N3** #2696 comment's false premise | **RESOLVED as to the premise** (`:280-283` now reads "At the time this was written… there is now"). Its replacement asserts `tolerance` is "derived from an on-box N=10 spread measurement, not picked" — contested by C1. |
| **N4** stale `test_qwen_duration_bless_gating.py` tolerance comment | **VERIFIED RESOLVED** — now "placeholder for these unit tests; independent of the real committed baseline". |
| **N5** measurement script untested | **PARTIALLY RESOLVED.** Pure math extracted + 5 tests — real, not cosmetic (a sample-vs-population stdev swap, a mean-vs-max swap, a divide-by-wrong-denominator and a headroom-multiplier swap are all killed). But see M5, and `measure()`'s orchestration is still untested and now *declared* out of scope — including the warm-up discard, which is the exact behaviour N6 turns on and which pass 1 demonstrated is stub-testable in minutes. |
| **N6** warm vs cold | **NOT RESOLVED, and undeclared.** `measure()` still discards its warm-up (`:111`); `_bless` still folds its cold first draw into the average (`:230-231`). Diluted from 1-of-1 to 1-of-5, not made to agree. My data says cold/warm is not the dominant driver — all five lines drifted, not just line 0 — so this is a nit, not the C1 explanation. |
| **N7** changelog entries missing the PR number | **VERIFIED RESOLVED** — both files carry "(PR #2739)". |
| **N8** live-view publish half not recorded | **STILL UNADDRESSED.** No record in the PR body or comments of the publish or of `check:onbox-register -- --against-published`. The live view now needs re-publishing regardless, since its "single run" sentence is false. |
| **N9** documented `--out` writes to the wrong directory | **NOT RESOLVED** — `measure_qwen_duration_spread.py:17-20` is byte-identical to pass 1; `--out spread-report.json` from the repo root still lands outside `tests/golden/`. |

---

## House gates — checked

- `check:onbox-register` and `check:register-citations` both exit 0 at this head. ✅
- Real golden assertion path re-run independently on GPU 1 → passes. ✅ (One pass against a ~3.65 %/run false-red rate is not evidence the bound holds.)
- Baseline is authentic averaged measurement, not fabricated (384-multiple frame signature on all five entries). ✅
- `Closes #1994` present, outside a code span. ⚠️ C1 undercuts criterion 3 ("tolerance derived from a measured distribution, not picked") more sharply than pass 1 did, and A101's row is already removed on that basis.
- Release-notes entry updated and accurate about *what changed*; `RELEASE_NOTES.md` skip still explicitly stated. ✅
- No `cast.json` writes, no config knob, no derived artifact. N/A.
- Incidental fixes: none folded in; none needed beyond the findings above. ✅
- No prose figure regressions: every mean/stdev/percentage in the register, live view, release notes and baseline `_comment` still matches `spread-report.json` exactly. ✅ (The *sentence* about how `entries` were produced is the one that went stale — above.)

dudarenok-maker:
## PR review — pass 3 (head 6dfd91119d41a69b40bae2665b4633ffa7c7b57c, depth high)

Reviewed at `6dfd9111` in a throwaway detached worktree. Verified by running, not by reading:

- **The merge is real and clean.** `83342154` has two parents (`06cce2b8` + `6038e82e`), and its tree hash is **byte-identical to the machine-computed 3-way merge**: `git merge-tree --write-tree 06cce2b8 6038e82e` → `ddd8c195864ff387e8e5ff3002322547d59d0721`, and `git log --format=%T -n1 83342154` → the same. No conflict was hand-resolved, and nothing was slipped in under cover of the merge. `git diff 6038e82e 83342154` shows only this branch's own nine files.
- `node scripts/check-onbox-register.mjs` → exit 0, "agree". `node scripts/check-register-citations.mjs` → exit 0, all A101 citations correctly annotated. Both re-run at this head, post-merge.
- **The stamp commit did what it claims.** `6dfd9111` changes exactly one line: `data-published-as="1" data-publish-id="25f4dfb810b6"` → `data-published-as="2" data-publish-id="7e9d71bbbb7f"`. `origin/main`'s copy is still at `1`/`25f4dfb810b6`, so the counter bump and the fresh machine-minted nonce are both genuine, and step 0 of the register's publish procedure is satisfied.
- `tolerance` is **genuinely unchanged at `0.3`** (`qwen-duration-baseline.json:3`). Not silently altered.
- The re-blessed `entries` are authentic averaged measurement: all five `sample_count`s are exact multiples of **384** (Qwen's 1920-sample frame ÷ 5 — the signature of averaging five frame-aligned draws) and `duration_sec == sample_count / 24000` exactly in every entry.
- Fast `-k "qwen_duration or measure_qwen"` tier → **17 passed** (was 15 at pass 2, 9 at pass 1).
- **A 13-mutation battery** against the new code, re-applying every mutation pass 2 named plus seven of my own.

Pass 2's mutation-coverage findings are genuinely fixed — M1, M2, M3, M4a, M5 and N6 are all killed now, and I confirmed each by applying the mutation and watching the suite go red. The C1 documentation is present on the three surfaces it was promised on. But the C4 fix **relocated its blind spot instead of closing it**, and two of the documented figures do not survive arithmetic.

---

## Correctness bugs

### C4-R — the C4 fix moved the blind spot from `[0]` to `[-1]`; the test still cannot fail for half the mutation its docstring names

`server/tts-sidecar/tests/test_qwen_duration_bless_gating.py:189`, docstring `:168-175`.

Pass 2 found `sample_counts = [24000, 26400, 21600]` had `[0] == mean`, so an "average → bless rep 0 only" mutant survived. Pass 2 also named the fix precisely: *"reorder to `[21600, 24000, 26400]` (mean 24000, first 21600, last 26400), so **neither endpoint** equals the mean."*

The commit used `[21600, 26400, 24000]` instead. Mean is still 24000 — and now **`sample_counts[-1] == 24000 == mean`**. The blind spot moved one index; it did not close.

Measured, both orderings, mutating `test_qwen_duration_golden.py:245-246` to bless a single rep:

| fixture ordering | `sample_counts[0]` mutant | `sample_counts[-1]` mutant |
|---|---|---|
| `[24000, 26400, 21600]` (pass 1) | SURVIVED | KILLED |
| **`[21600, 26400, 24000]` (shipped now)** | KILLED | **SURVIVED** |
| `[21600, 24000, 26400]` (pass 2's recommendation) | KILLED | KILLED |

The docstring at `:168` states verbatim: *"Kills a mutant that blesses only the **first/last** rep."* The `last` half is false, and was true before this commit. Under `"duration_sec": round(durations[-1], 4)` / `"sample_count": sample_counts[-1]`, `test_bless_averages_across_reps` passes and the whole suite stays green — `_bless` would record one arbitrary draw, which is the pre-fix C1 behaviour this test exists to prevent.

One-line fix, already written by pass 2: `sample_counts = [21600, 24000, 26400]`. I confirmed that ordering kills both.

### D1 — the `~3.65% per run` figure is stale: it was computed against the *previous* baseline, and the one this PR actually ships is worse

`server/tts-sidecar/tests/golden/qwen-duration-baseline.json:2`; `docs/testing/onbox-acceptance-register.md:463`; `docs/testing/onbox-acceptance-register-live-view.html:232`.

All three surfaces state the accepted risk as **"~3.65% per run"**. That number came from pass 2, which modelled it against the **round-1 blessed entries** (`3.248 / 5.216 / 4.544 / 4.368 / 5.200`). The N6 fix in this same commit **re-blessed every entry** (`3.584 / 5.632 / 4.592 / 4.288 / 5.312`), which moves the figure — and nobody re-derived it.

Reproducing pass 2's exact methodology (per-line normal fit to `spread-report.json`'s own `mean_sec`/`stdev_sec`; fail when `|fresh − blessed| / blessed > 0.30`):

| blessed entries modelled | per-line | combined |
|---|---|---|
| round-1 entries (what pass 2 measured) | 0.00 / 0.00 / 0.46 / 1.52 / 1.71 % | **3.65 %** ← reproduces pass 2 exactly |
| **CURRENT shipped entries** | 0.14 / 0.00 / 0.59 / 0.97 / **2.49** % | **4.15 %** |

So the documented risk understates the shipped risk by ~14% relative. `multi-sentence-group` alone carries 2.49%/run.

Compounding it, the parenthetical in the register and live view — *"(vs. the ~2% estimated from within-session spread alone)"* — mis-attributes the cause. Both 2% and 3.65% are computed from **within-session** spread; neither incorporates the between-session drift. The rise from ~2% to ~3.65% was caused by the re-bless landing on different values, **not** by the drift finding. Presenting 3.65% as the number that accounts for the drift is the one place this documentation genuinely downplays: the drift makes the rate *session-dependent and uncharacterised*, not a fixed 3.65%. Modelling pass 2's own session-B means against the current entries gives **1.57%**, versus 4.15% for session A — the honest statement is a range that the single measurement cannot bound, not a point estimate.

### D2 — the baseline's `_comment` attributes the drift range to the wrong reference set, making it false as written

`server/tts-sidecar/tests/golden/qwen-duration-baseline.json:2`.

> "a second on-box run measured every line's fresh mean 6.8-21% above **this file's committed means**"

"This file" is `qwen-duration-baseline.json`; its committed means are `entries`. The 6.8–21% range is the drift against **`spread-report.json`'s** means, which is a different file. Against *this* file's entries the drift is a different number, and in the opposite direction on three of five lines:

| line | vs. `spread-report.json` mean | vs. **this file's** entry |
|---|---|---|
| `narrator-plain` | +19.89 % | **−1.79 %** |
| `numbers-and-year` | +6.83 % | **−1.51 %** |
| `abbreviations` | +21.10 % | +10.34 % |
| `quotes-and-punctuation` | +11.59 % | **−2.36 %** |
| `multi-sentence-group` | +20.39 % | +6.43 % |

So the sentence's range (6.8–21%), its direction ("above"), and its claim ("every line") are all wrong for the reference set it names. The register (`:459`), the live view (`:229`) and #2742's comment all say **"the committed `spread-report.json`'s means"** and are correct. The baseline `_comment` is the odd one out — and it is the surface an engineer changing this file reads first. Two words: `this file's` → `spread-report.json's`.

This matters beyond pedantry: a future operator deciding whether to re-bless will compute drift against `entries` (the natural reading), get −2.4% to +10.3%, conclude the documented limitation is overstated, and re-bless on that basis.

### C3-R — the up-front voice check is set-membership where the deleted one was per-line equality, so a mixed-voice baseline now silently compares across speakers

`server/tts-sidecar/tests/golden/test_qwen_duration_golden.py:113-114`.

C3's fix replaced a **per-line** `recorded_voice != voice` check with an **up-front set** check: `recorded_voices = {…}` then `if recorded_voices and voice not in recorded_voices`. For a baseline where every entry carries the same voice these are equivalent. For one where they don't, they are not: **any** match anywhere in the set passes the whole gate.

Driven against the real function (baseline with `line1: voice_a`, `line2: voice_b`; box resolves `voice_a`):

```
recorded_voices=['voice_a', 'voice_b']  resolved='voice_a'
OUTCOME: RAN TO COMPLETION — line2's entry (blessed against voice_b) was
         compared against audio synthesised with voice_a. No skip, no failure.
```

That is exactly what the code this PR deleted existed to stop, in its own words: *"a comparison across a voice change is silently comparing two different speakers' timing, not detecting a real regression."* The surviving comment at `:120` still asserts "Comparing across voices is meaningless" while the guard above it now permits precisely that.

Reachable via a merge of two lanes that each re-blessed on a box with a different designed voice (git merges the per-entry JSON hunks cleanly — they are separate keys), or any hand-edit/partial bless. `_bless` itself always writes one voice for all entries, so this is not reachable from a clean bless — but it is the failure mode the check is named for, and no test covers it. Keep the up-front skip and restore a per-line `base.get("voice") in (None, voice)` assertion inside the loop.

### T1 — the voice-mismatch test's `match=` regex is loose enough that it passes when the skip comes from somewhere else entirely

`server/tts-sidecar/tests/test_qwen_duration_bless_gating.py:331`.

```python
with pytest.raises(pytest.skip.Exception, match="voice_a.*voice_b|voice"):
```

Pass 2's "minimum to close" for C3 listed four items. Two landed (the matching-voice test → M3 killed; the unexpected-skip wrapper → M1 killed). The other two — *"tighten the regex to the mismatch-specific text; drop the `|voice` alternative"* — did not, and `:331` is byte-identical to pass 1. The `|voice` alternative matches any skip message containing the substring "voice", which includes both other skip paths in the function.

Two mutants, both **SURVIVED**:

| mutant | what actually happens | test result |
|---|---|---|
| `entries` read as empty | function takes the **unblessed-baseline** skip path — a completely different branch | passes |
| mismatch skip fires with a generic message (no voice ids) | diagnostic voice ids gone from operator-facing text | passes |

So the test does not pin that the skip came from the voice-mismatch check, nor that the message names the two voices. Fix: `match=r"blessed against voice\(s\).*this box resolved"`.

Relatedly, **M2 still goes red for the wrong reason**, exactly as pass 2 reported — deleting the whole voice block fails with `TypeError: '>' not supported between instances of 'MagicMock' and 'int'` at `compare.py:85` (unmocked plumbing), not the voice logic.

---

## Pass-2 finding dispositions

| # | Status |
|---|---|
| **C1** documentation of the accepted risk | **EXECUTED, with two defects.** Present on all three promised surfaces + #2742. `tolerance` genuinely unchanged at `0.3`. Accepting-and-documenting with explicit owner sign-off is a legitimate closeout for this finding, and I am not re-litigating the decision. But see **D1** (the headline figure is stale and understates the shipped risk) and **D2** (the baseline `_comment`'s referent is wrong, making the sentence false as written). Both are two-word/one-number edits. |
| **M1** dropped `is not None` turned PASS into SKIP | **VERIFIED RESOLVED** — mutation applied, suite goes red. The `try/except → pytest.fail` wrapper at `:415-419` works. |
| **M3** no matching-voice coverage | **VERIFIED RESOLVED** — `test_assertion_loop_does_not_skip_when_voice_matches` (`:335`) added; `if recorded_voices:` mutant now killed. |
| **C4** averaging test blind to `[0]` | **PARTIALLY RESOLVED → see C4-R.** `[0]` killed; `[-1]` newly survives. Net coverage unchanged, blind spot relocated. |
| **M5** unpinned `abs()` | **VERIFIED RESOLVED** — `test_line_stats_low_outlier_still_drives_max_abs_dev` (`:59`, fixture `[0.5,1.0,1.0,1.0,1.0]`, mean 0.9, max signed +0.1 vs max abs 0.4) kills the mutant. |
| **N6** warm-up folded into the bless average | **VERIFIED RESOLVED.** `:229` is a standalone discarded `synthesise_or_skip`; every line gets `BLESS_REPS` fresh draws (call log `["skip"] + ["direct"]*6` pinned at `:156`). Reverting to the old shape goes red. The re-blessed entries carry the 5-average 384-multiple signature on all five lines, consistent with 5 equally-warm draws. |
| **N9** `--out` example path | **VERIFIED RESOLVED** — `measure_qwen_duration_spread.py:17-23` now warns the path resolves against cwd and gives the full `server/tts-sidecar/tests/golden/…` path. |
| **N8** live-view publish half | **MERGE + STAMP HALF VERIFIED RESOLVED** (real two-parent merge, tree == machine 3-way, token genuinely bumped, both register checks green post-merge). **Publish half still owed** — steps 1–4 of the register's procedure are a manual operator action and appear nowhere in the PR body or test plan. Note the trap this PR walks into: A101 is a **pre-merge discharge**, so step 2 will trip its "row is live and behind" message and needs `npm run check:onbox-register -- --against-published <saved> --discharging A101`. That is not written down anywhere in this PR. |
| register/live-view "single run" sentence | **VERIFIED RESOLVED** — both now describe the `BLESS_REPS=5` mean. |
| **#2742** premise correction | **RESOLVED** — the correction comment is present, accurate, and uses the correct referent. The label half is not: still `type:chore` with no `moscow:`. |
| C3 "minimum to close" items 3 & 4 (regex) | **NOT RESOLVED → see T1.** |

---

## Cleanup / nits

### N-a — `test_qwen_duration_bless_gating.py:13-17`: the module docstring describes the code N6 deleted

Three false clauses, all made false by this commit:

> "`_bless` uses `synthesise_or_skip` only for the very first synthesis of the whole bless **(line 1's first rep)**; … so an **`i == 0` -> `i == 1` swap** can't survive … (**`test_bless_uses_synthesise_or_skip_only_for_first_line`**)"

The warm-up is no longer any line's rep (that is the whole N6 fix); `_bless` no longer has an `i` at all (`for line in fixture["lines"]`, `for _ in range(BLESS_REPS)`); and the test it names was renamed to `test_bless_uses_synthesise_or_skip_only_for_the_warmup` (`:107`), so the citation dangles. The identical `i == 0 -> i == 1` phrasing at `:459` is fine — that one refers to the *assertion* loop, which does still have an `i`.

### N-b — the accepted limitation is absent from the surfaces an engineer edits, and three of them still assert the premise it contradicts

Pass 1 established that the tolerance justification is recorded in five places. The limitation landed in three (baseline `_comment`, register, live view) plus #2742. It is missing from the other two, both of which still state the falsified claim flatly:

- `docs/release-notes-next.md:532` — *"making the reference a stable estimate of the true mean rather than one noisy draw."* Pass 2's C1 is precisely that there is no stable between-session mean on this box. The technical register is the durable record and carries no caveat.
- `test_qwen_duration_golden.py:285-286` — *"`tolerance` is 0.30, derived from an on-box N=10 spread measurement, **not picked**."* Pass 2 flagged this as contested; still uncontextualised.
- `test_qwen_duration_golden.py:182-194` (`BLESS_REPS` comment) and `_bless`'s docstring `:198-217` — same premise, no caveat.

A reader of the code alone never learns the limitation exists.

### N-c — the PR body is stale and omits the accepted risk entirely

The body still reads *"then blessed `entries` from a fresh single run against the same voice/model"* — the exact sentence pass 2 got fixed in the register and live view, still uncorrected here (entries are means of five). More importantly, the body says nothing about the drift finding, the accepted known limitation, the owner's ship-as-is decision, #2742, or the two rounds of review fixes folded in. CLAUDE.md requires incidental fixes to be *declared in the PR body*; a merger reading only the body would not know the headline caveat of the change exists.

### N-d — two coverage gaps my battery turned up

- `test_qwen_duration_golden.py:260` — hardcoding `"bless_reps": BLESS_REPS` to `1` **survives** the suite. Nothing pins that the metadata field tracks the constant, and `metadata.bless_reps: 5` is cited in the register, the live view and the `_comment` as the *evidence* that entries are means of five. The provenance field can silently lie.
- `test_qwen_duration_golden.py:152-154` — deleting the RMS near-silence check entirely **survives**. Pre-existing (it is on `main` too), not introduced here, but this PR restructured the loop around it.

---

## `Closes #1994` — assessment (reported, not blocking)

#1994's acceptance criterion 3 is *"Tolerance chosen from a measured distribution, not picked."*

Read literally, it is **met**: `0.30` was derived from a real N=10 on-box measurement of the actual engine, retained at `spread-report.json`, and every figure quoted from it reproduces exactly. It was not picked.

Read as intent, it is **weaker than it looks**: the measured distribution is a *within-session* one, and pass 2 demonstrated the statistic the gate actually computes is *between-session*. So the tolerance is derived from a measurement of the wrong population — measured, but not measured against the thing it bounds. Criterion 4 ("paired mutation-verified tests, each mutating the producer") is also not fully met while **C4-R** and **T1** stand.

My assessment: `Refs #1994` is the more honest trailer, with #2742 carrying the remainder. But this is squarely the call the repo owner already made with "ship as-is, document the known risk", and the register row is already discharged on that basis — so I flag it rather than pressing it, and I would not hold the merge on it. **What I would fix first is C4-R**: it is one line, it restores coverage this PR previously had, and shipping a test whose docstring names a mutation it cannot catch is the shape this repo's review history keeps re-finding.

---

## House gates — checked

- Merge is a genuine two-parent merge whose tree equals the machine-computed 3-way result; no conflict mishandled. ✅
- `check:onbox-register` and `check:register-citations` both exit 0 at this head, post-merge. ✅
- Publish token genuinely stamped (`1`→`2`, fresh nonce), matching the register's step 0. ✅ Steps 1–4 still owed (see N8 row).
- Baseline is authentic averaged measurement (384-multiple signature, exact `sample_count/24000`); `tolerance` unchanged at `0.3`. ✅
- Release-notes entry present; `RELEASE_NOTES.md` skip explicitly stated as test-tier/bookkeeping. ✅ (accuracy — see N-b)
- `Closes #1994` present and outside a code span. ⚠️ see assessment above.
- No `cast.json` writes, no config knob, no derived artifact, no run sheet for A101. N/A.
- Incidental fixes: none folded in beyond the review findings; none declared in the body. ⚠️ see N-c.
- Every mean/stdev/percentage quoted from `spread-report.json` across the register, live view, release notes and `_comment` still reproduces exactly. ✅ — the two figures that do **not** hold up are the ones *derived* since (D1) and the one whose *referent* is wrong (D2).

<sub>Reviewed read-only. I modified no tracked file in the primary checkout. Note for the dispatching session: `RELEASE_NOTES.md` and `docs/release-notes-next.md` show as modified there, but that content is #2751 / PR #2745 (VoiceDesign eviction) — a concurrent lane writing into the primary checkout, not this review.</sub>

Issue #2599, opened 2026-08-22
check-onbox-register.mjs's --against-published check is blind to row content, only compares totals/counts/IDs


dudarenok-maker:
## What

`scripts/check-onbox-register.mjs --against-published <file>` (the register's mandated pre-publish comparator, per `docs/testing/onbox-acceptance-register.md`'s "Live view" procedure) only compares the owed total, per-group counts, and row IDs between the local register and a saved copy of the live page. It never diffs row **content**.

## Why this matters

Found during PR #2578 round 18's review pass: the published live-view artifact was independently reverted by another session to a stale, pre-fix version of row A41 (the exact defect PR #2578 exists to correct — backwards clobbered-state direction and a manufacture recipe that #2535 proved never reaches the `'clobbered'` state). Running `check:onbox-register -- --against-published <that reverted page>` returned `OK`, because the row count and row ID (A41) both still matched — only the row's *body text* had regressed.

This is a guard blind to the exact class of defect the whole "Live view" procedure exists to catch: a byte-diff between the tracked HTML and the actual published page. The byte-diffs performed manually in prior PR #2578 review rounds (13-18) were the thing doing the real verification work; the mandated comparator itself contributed nothing to catching a content-level regression.

## What's owed

A design decision — not a mechanical fix — on how the comparator should diff content:
- Whole-page byte/text diff against the saved copy (simple, but any legitimate concurrent edit from another PR, like #2588's A48 addition, would then also flag as a "difference" — the current row-ID-based check tolerates that case, via #2199's qualifier).
- Per-row content hashing, so a row's ID+count staying stable but its body text changing is detected, while a genuinely new row from a different in-flight PR is not flagged.
- Something else — the design choice of what counts as "the row this check is responsible for" vs. "content owned by another lane" needs a decision, which is why this is filed rather than fixed inline.

## Where

- `scripts/check-onbox-register.mjs` — the comparator itself
- `docs/testing/onbox-acceptance-register.md` — the "Live view" section documenting the mandated procedure, which should be updated to describe whatever new capability lands here

Filed per PR #2578 round 18's review findings — see [PR #2578, pass 18 review comment](https://github.com/dudarenok-maker/Castwright/pull/2578#issuecomment-5376608839) for the full repro.

dudarenok-maker:
Decision (design-pass resolved in-session, 2026-09-01): **per-row content hashing**, not whole-page diff.

[...]

PR #2740, opened 2026-08-28
feat(ops,scripts,docs): seed the live view's publish token + stamper

dudarenok-maker:
## Summary

Seeds a publish token into the on-box register's live view and adds the command that maintains it. This is the **data half** of #2599; the check that reads the token lands separately.

**Why the split — a bootstrap break that was live and unnoticed.** `check-onbox-register.mjs` validates `origin/main`'s *fetched* copy, so a guard cannot ship in the same change as the data it requires. The Track A/B split moved the seeding task out of the first PR, Track A shipped anyway, and both the spec and the plan still assert *"PR 1 seeds the token as data … so `origin/main` carries one by the time this code runs"*. It does not — `grep -c published-as` returned **0**. Shipping the comparator in that state would have made its first ever run return *"origin/main carries none — this is a revert or a deletion; do not publish; investigate"*: an unclearable STOP misdiagnosing its own cause, on a design whose stated principle is that an unclearable STOP is a guard that gets bypassed. Same data-then-guard shape the stable row IDs needed (#2629).

**Why a command instead of "bump the number".** A hand-maintained token goes stale, and a stale one fails in the **green** direction — a competing lane whose value happens to match yours reads as your own earlier publish, so the check waves through exactly the overwrite it exists to stop. Bumping the counter without minting a fresh id is that failure with an extra step. Two of the eight designs considered for this token died on precisely that, so `npm run stamp:publish-token` does both halves or neither — and it writes the genesis token too, so even the seed is machine-written.

The token is attributes on a hidden `<div>`, not an HTML comment: `checkLiveView`'s first action is `stripHtmlComments`, so a comment-carried token would be invisible to the one function that must read it.

## Test plan

**53 new tests**, all green; `check:onbox-register` and `check:register-citations` both green.

- **Unit (37)** — the comparator PR 2 will wire.
- **Stamper (8)** — pins the bootstrap invariant (the live view carries exactly one well-formed token), that the token survives comment stripping, and that every minted nonce is one the parser accepts. A stamper writing a token its own reader rejects would wedge the check permanently.
- **Real-git (8)** — new in kind. Builds throwaway repositories and asks the real binary, because **four of the five green-direction defects found in review lived in `nonceInHistory`'s nine lines**, where a stubbed runner pins the argv and nothing about what git does with it. Every one of those defects passed a green unit suite. Covers merge-conflict-born nonces, history-simplification pruning, and the `-S` substring anchor.

The git suite scrubs `GIT_*` from the environment of every spawn, and its first test asserts the scrub works. Both are load-bearing: git exports `GIT_DIR`/`GIT_INDEX_FILE` to hook children, `git -C` does **not** override them, so an inheriting `git init` in a temp repo writes to the *real* one. That happened twice while writing this file — 3,979 tracked files down to 1, and the branch moved onto a fixture commit — and it is silent. The scrub is computed **per spawn** rather than once at import: as a module constant it was snapshotted before any test could inject `GIT_DIR`, and the self-check passed with the scrub deleted.

## Notes

- **Release notes deliberately skipped** — maintainer tooling with no user- or operator-visible effect, matching #2717, the same class of change.
- **The token is inert in this PR.** Nothing reads it yet; that is the point of shipping it first.
- The register documents the stamp → commit → publish order in a new "The publish token" section.

Refs #2599

### DIFF
--- a/docs/testing/onbox-acceptance-register-live-view.html
+++ b/docs/testing/onbox-acceptance-register-live-view.html
@@ -161,6 +161,12 @@
 
   <p class="eyebrow">Castwright · docs/testing/onbox-acceptance-register.md</p>
   <h1>On-box acceptance register</h1>
+
+  <!-- Publish token. NEVER hand-edit: run `npm run stamp:publish-token`.
+       The counter orders publishes; the id proves which branch produced one.
+       Attributes, not a comment: checkLiveView blanks comments before reading. -->
+  <div hidden data-published-as="1" data-publish-id="25f4dfb810b6"></div>
+
   <p class="lede">
     Shipped behaviour that only real hardware can prove — a live GPU, a real sidecar, a real
     analyzer, a real book, a real phone — and that was <strong>not</strong> proven at PR time.
--- a/docs/testing/onbox-acceptance-register-live-view.html
+++ b/docs/testing/onbox-acceptance-register-live-view.html
@@ -164,7 +164,7 @@
 
   <!-- Publish token. NEVER hand-edit: run `npm run stamp:publish-token`.
        The counter orders publishes; the id proves which branch produced one.
-       Attributes, not a comment: checkLiveView blanks comments before reading. -->
+       Attributes, not a comment: checkLiveView removes comments before reading. -->
   <div hidden data-published-as="1" data-publish-id="25f4dfb810b6"></div>
 
   <p class="lede">
--- a/docs/testing/onbox-acceptance-register.md
+++ b/docs/testing/onbox-acceptance-register.md
@@ -48,6 +48,41 @@ and concluded they had complied. The live view is tracked in this repo — rathe
 than living in whichever session scratchpad last built it — precisely so the
 right file is always to hand.
 
+### The publish token — never hand-edit it
+
+The live view carries a **publish token** just under its `<h1>`:
+
+```html
+<div hidden data-published-as="<counter>" data-publish-id="<nonce>"></div>
+```
+
+The counter orders publishes; the id says *which branch produced this one*. Both
+are machine-written. **Do not type either value.** Run:
+
+```
+npm run stamp:publish-token              # bumps the counter AND mints a fresh id
+npm run stamp:publish-token -- --check   # report only, changes nothing
+```
+
+**Stamp, then commit, then publish** — in that order. An uncommitted stamp
+cannot be found in git history, which is where its freshness is verified.
+
+Why a command rather than an instruction to bump the number: a token that a
+human maintains goes stale, and a stale one fails in the **green** direction —
+a competing lane whose value happens to match yours reads as your *own* earlier
+publish, so the check waves through exactly the overwrite it exists to stop.
+Bumping the counter without minting a new id is that same failure with an extra
+step. Two of the eight designs considered for this token died on precisely that,
+which is why the stamper does both halves or neither.
+
+The token is inert today: nothing reads it yet. The check that does — a
+comparison against the live page's own token, to catch a lane publishing over
+work it never saw — lands separately (#2599). It is seeded first and on its own
+because a guard cannot ship in the same change as the data it requires: the
+checker validates `origin/main`'s copy, so the data has to be on `main` already
+or the guard's first run fails on its own delivery. That is the same
+data-then-guard split the stable row IDs needed (#2629).
+
 The live view carries derived figures — owed count, per-group counts, oldest
 debt — that must be **recomputed** on every edit. Rows can be right while the
 summary strip lies. `npm run check:onbox-register` verifies the owed total, the
--- a/docs/testing/onbox-acceptance-register.md
+++ b/docs/testing/onbox-acceptance-register.md
@@ -83,6 +83,16 @@ checker validates `origin/main`'s copy, so the data has to be on `main` already
 or the guard's first run fails on its own delivery. That is the same
 data-then-guard split the stable row IDs needed (#2629).
 
+**One more thing has to happen before that check can pass, and it is easy to
+miss because it is not a code change: the live view must be PUBLISHED at least
+once with a token on it.** The comparator reads three copies — the tracked file,
+`origin/main`'s, and the *saved live page* — and the live page only acquires a
+token when someone publishes after this change merged. Until then the check
+reports that the published page carries none while `origin/main` does, and
+names the transition explicitly rather than guessing. So the first publish after
+this merges clears it, and the wording of that error is written for exactly that
+window.
+
 The live view carries derived figures — owed count, per-group counts, oldest
 debt — that must be **recomputed** on every edit. Rows can be right while the
 summary strip lies. `npm run check:onbox-register` verifies the owed total, the
@@ -129,6 +139,12 @@ CI check — the same call this design already made for the tracked-pair
 comparison, see the edge list above). The merge step that closes this, run
 **immediately before every publish**, not only after a suspected race:
 
+0. **If your change touched the live view, stamp it and commit the stamp** —
+   `npm run stamp:publish-token`, then commit. It goes *before* step 2, not
+   after step 4, for two reasons: the check reads the token, and an
+   **uncommitted** stamp cannot be found in git history, which is where its
+   freshness is verified. See "The publish token" above for why this is a
+   command rather than an instruction to edit a number.
 1. Fetch the page currently live at the canonical URL above and save it to a
    local file — this is the CURRENTLY-published register, which may be ahead
    of what you are about to publish.
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "test:quarantine": "cross-env RUN_QUARANTINE=1 npm run test && cross-env RUN_QUARANTINE=1 npm --prefix server run test && cross-env RUN_QUARANTINE=1 npm --prefix server run test:slow",
     "quarantine:health": "node scripts/quarantine-health.mjs",
     "check:onbox-register": "node scripts/check-onbox-register.mjs",
+    "stamp:publish-token": "node scripts/stamp-publish-token.mjs",
     "check:register-citations": "node scripts/check-register-citations.mjs",
     "check:cycles": "node scripts/check-import-cycles.mjs",
     "capture:marketing": "playwright test --config=playwright.marketing.config.ts --project=desktop",
--- a/scripts/check-onbox-register.mjs
+++ b/scripts/check-onbox-register.mjs
@@ -449,8 +449,11 @@ export function checkRegister(text) {
 // directions matter and both were wrong without it (PR #2080 review round 2):
 // a commented-out row was counted as a real one, and commenting out a whole
 // group section — which removes it from the published page — was invisible.
-// Blanking rather than deleting keeps the comment's own `<section>`/`<span>`
-// text from being read while leaving the surrounding structure intact.
+// Removing the comment entirely keeps its own `<section>`/`<span>` text from
+// being read while leaving the surrounding structure intact. (This said
+// "blanking rather than deleting" and the code has always done the latter —
+// `.replace(..., '')` removes the span rather than substituting whitespace.
+// Corrected because the distinction is the kind a reader would reason from.)
 export function stripHtmlComments(html) {
   let out = html;
   for (;;) {
--- a/scripts/publish-token.mjs
+++ b/scripts/publish-token.mjs
@@ -1,0 +1,341 @@
+// Publish-token comparator — executable draft for #2599.
+//
+// SEVEN designs preceded this. Three per-row content rules; a bare counter
+// (cannot tell your re-publish from a rival's — same number); a branch name
+// (inherited, renamed, degenerate under detached HEAD); a hand-minted nonce
+// (goes stale, fails GREEN); and a nonce whose freshness was tested against
+// `b.nonce` rather than against history (closed the hole by exactly one commit).
+//
+// The freshness question is NOT "does the page's nonce differ from main's". It
+// is: **is the page's nonce already in the baseline's history while its counter
+// is ahead of the baseline?** That is the signature of "somebody bumped without
+// minting", at any depth, and it is what closes the class.
+
+export const PUBLISH_TOKEN_BASELINE_ERROR =
+  "Cannot verify the publish token: origin/main's live view is unavailable or " +
+  'unreadable. Do not publish until this passes.';
+
+export const PUBLISH_TOKEN_PUBLISHED_ERROR =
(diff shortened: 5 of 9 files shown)


dudarenok-maker:
## PR review — pass 1 (head 1c300dbc, depth high)

Scope: all 8 changed files · verified before probing: `node --test` on each new suite (8 real-git + 8 stamper + 37 comparator, all green), `check:onbox-register` OK, `check:register-citations` OK, `test:hooks` 1663/1663 pass, `validate-commit-msg.mjs` green on both the commit subject and the PR title. Worktree `git rev-parse HEAD` + `git status --porcelain` identical before and after (1c300dbc, clean, 3984 tracked); primary checkout untouched (3979 tracked, unchanged).

Probing concentrated on the repo-destruction hazard in `publish-token-git.test.mjs`, the stamper's write paths, and the seeded token's durability.

---

### 🔴 Blocking — the `GIT_*` scrub is case-sensitive; git honours a lowercase `git_dir` on Windows, and the self-check cannot see it

`scripts/tests/publish-token-git.test.mjs:40-44`

```js
const cleanEnv = () => {
  const env = { ...process.env };
  for (const key of Object.keys(env)) if (key.startsWith('GIT_')) delete env[key];
  return env;
};
```

`startsWith('GIT_')` is case-sensitive. `{ ...process.env }` snapshots whatever casing Windows stored a variable under, and git's own lookup is case-insensitive there — so a `git_dir`-cased variable survives the scrub and still redirects every spawn in the file.

**Measured on this box** (git 2.54.0.windows.1, node v24.15.0), using the same spread-and-filter the test uses:

```
lowercase key survives cleanEnv(): [ 'git_dir' ]
git resolved --absolute-git-dir to: .../scratchpad/casedecoy/.git
isolated answer would be         : .../scratchpad/other/.git
```

Failure scenario: any ambient `git_dir=<path-to-a-real-.git>` (an operator export, a wrapper script, a tool that sets it lower-cased) → `newRepo()`'s `git init` / `git add -A` / `git commit -qm 'base'` at lines 72-76 and 95-96 run against that repository instead of the `mkdtemp` one → the recorded incident, 3,979 tracked files down to 1 with the branch moved onto a fixture commit. Silent.

Two things make this worse than an ordinary spelling gap:

1. **The instrument cannot fail for it.** The self-check at lines 91-92 injects `GIT_DIR`/`GIT_INDEX_FILE` in **uppercase only**, so it proves the scrub works for the one casing that was already covered. Deleting the whole scrub is caught (I confirmed — see Solid below); mis-casing it is not.
2. **This repo already closed this exact hole once, in writing.** `scripts/git-env.mjs:70-82` matches on `key.toUpperCase()`, and its header says a case-sensitive delete is *"a no-op fix for exactly the case this function exists to close."* The new file re-opens that at a strictly higher-stakes site.

Honest calibration: git's own hook exports are uppercase, so the trigger is an unusually-cased ambient variable, not the shape that caused the two recorded incidents. I am rating on blast radius plus existing written precedent, not on likelihood — and the fix is `key.toUpperCase().startsWith('GIT_')` plus a lowercase case in the self-check.

---

### 🟠 Significant — a typo'd or `=`-formed flag silently performs the WRITE

`scripts/stamp-publish-token.mjs:71-77` (`argv.indexOf('--file')`) and `:88` (`argv.includes('--check')`)

There is no unknown-flag rejection. Anything that is not exactly `--check` falls through to the stamp-and-write path, and anything that is not exactly `--file <path>` silently targets `DEFAULT_FILE`.

Both reproduced against a sandbox copy of the script (never the worktree):

```
$ node stamp-publish-token.mjs --chek          # typo for --check
stamp-publish-token: stamped docs/testing/onbox-acceptance-register-live-view.html -> 3 (id ab63c04f74a8).
exit=0
```

```
$ node stamp-publish-token.mjs --file=other.html   # the `=` form
stamp-publish-token: stamped docs/testing/onbox-acceptance-register-live-view.html -> 2 (id 46d3cc7d709d).
exit=0        # other.html untouched
```

The `--chek` case is the sharp one: a **read-only intent silently mutates a tracked file**. The register's new section tells operators to run `--check` to "report only, changes nothing"; one transposed character does the opposite and exits 0.

The downstream consequence is not just an unwanted diff. An accidental bump that gets committed alongside unrelated work puts `origin/main`'s counter ahead of the published page — which is precisely the state `comparePublishTokens` (`scripts/publish-token.mjs:236-241`) reports as *"the page is BEHIND main. A bump merged without publishing, or a publish was reverted. Confirm which, then re-run with `--live-page-behind-main`."* So a typo manufactures the one state the design keeps a manual escape hatch for.

Precedent for rejecting this: sibling `scripts/wt-new.mjs:77` throws ``unknown flag: ${a}``, and `test:hooks` carries a test named *"parseArgs rejects unknown flags including typos like `--noinstall`"* for exactly this reason.

---

### 🟠 Significant — `main()` has no test coverage at all, which is why the above shipped

`scripts/tests/stamp-publish-token.test.mjs` (all 8 tests)

Every test in the stamper suite calls `seedToken`, `bumpToken` or `mintNonce` directly. Nothing spawns the CLI. So none of the following is covered: flag parsing, `--file` resolution, the `--check` exit codes (0 / 1 / 1), the unreadable-file branch at `:79-84`, the write at `:120`, or the entry-point guard actually invoking `main()`.

Concretely: `--check` on a tokenless file returns 1 but writes its message to **stdout** (`:94`), while the malformed branch writes to **stderr** (`:98`) — an inconsistency no test would notice. And the flag defects above are behaviour a single spawn-based test would have caught.

The repo's bar is "New behaviour → ship paired automated test(s)"; `main()` is new behaviour with zero.

---

### 🟡 Minor

1. **`seedToken` omits both hardenings `bumpToken` applies at the identical seam** — `scripts/stamp-publish-token.mjs:48-67` vs `scripts/publish-token.mjs:317-341`. `bumpToken` validates the minted nonce against the parser's own charset (*"`mintNonce` is caller-supplied, so nothing else constrains it"*) and uses a **function** replacement so `$&` / `$1` / a backtick-dollar pattern cannot splice markup back into the attribute. `seedToken` takes the same caller-supplied `mint` and does neither — it interpolates straight into a **string** replacement at `:66`. Catalogue #10: one instance fixed, the class left armed. (Same function: `SEED_AFTER` matches the first `<h1>` in **raw** HTML, so an `<h1>` inside a comment would anchor the token inside a comment — invisible to `checkLiveView`, the one outcome the design says must never happen.)

2. **`bumpToken` validates the minted nonce but not the resulting counter.** `bumpToken('<p data-published-as="999999999999999" ...>', mint)` writes `1000000000000000`, which its own `parsePublishToken` then rejects as *"not a bare integer of 1-15 digits"* (reproduced). The file is then wedged in both directions — `seedToken` refuses ("already has a token"), `bumpToken` refuses ("malformed") — the exact permanent wedge test 18l pins for the *nonce* half. Operationally unreachable; reported because it is the same stated invariant, half-enforced.

3. **Third divergent copy of the `GIT_*` scrub.** `scripts/git-env.mjs`'s `scrubGitEnv` (4 keys, case-insensitive), `scripts/tests/bump-version.test.mjs:45-51`'s `cleanGitEnv` (prefix, case-sensitive), now `publish-token-git.test.mjs:40-44`'s `cleanEnv` (prefix, case-sensitive). `scrubGitEnv` genuinely cannot be reused here — it deliberately excludes `GIT_INDEX_FILE`, which is one of the two variables this file must strip — but that argues for one shared helper with a documented broader mode, not a third copy that diverges again.

4. **The stamp step never reaches the procedure operators actually follow.** The new "Stamp, then commit, then publish" lives in its own subsection at `docs/testing/onbox-acceptance-register.md:52-77`; the numbered four-step publish procedure at `:130-217` — the one CLAUDE.md's before-shipping step 3 points at — still begins at "Fetch the page currently live". An operator following the steps never stamps.

5. **The bootstrap is only half-completed, and the remaining half is undocumented.** After merge, `origin/main` carries a token and the published page does not. PR 2's first `--against-published` run therefore lands on `scripts/publish-token.mjs:104-108` ("the published page carries no publish token, but origin/main does … Do not publish over it until you know which"). That message diagnoses the cause correctly, so this is not the misdiagnosing STOP the PR body set out to avoid — but it does mean **one ordinary publish must happen between this PR and the comparator PR**, and nothing in the register says so.

6. **A claim about `stripHtmlComments` that its code contradicts.** The seeded HTML comment (`docs/testing/onbox-acceptance-register-live-view.html:167`) and the register (`:74`) both say `checkLiveView` *blanks* comments. `scripts/check-onbox-register.mjs:454-461` replaces them with `''` — it deletes. The inaccuracy originates at `:452-453` ("Blanking rather than deleting…"), pre-existing; this PR propagates it to two new places. Harmless for the token either way.

7. **~200 lines land with no caller, and the module's riskiest property is reviewable in neither PR.** `comparePublishTokens`, `nonceInHistory` and the three error constants in `scripts/publish-token.mjs` have no production caller; only `bumpToken`/`parsePublishToken` do. The data-then-guard argument justifies shipping the *token* first — it does not require shipping the *comparator* now. The concrete cost: the file's own comment at `:139-141` says *"the delivery plan currently wires ONE lookup where this function consumes four"* — that wiring is the module's highest-risk property, it is absent here, and PR 2's reviewer will see this file unchanged in their diff.

---

### ✅ What is solid

- **The env scrub is genuinely mutation-sensitive for the casing it does cover.** I copied the file to scratch, stripped `env: cleanEnv()` from all four spawn sites, and ran the self-check: it went red with `GIT_DIR leaked: the decoy repo was written to` / `actual: 'docs/live.html'`. The "vacuous when snapshotted at import" trap called out in the header is genuinely closed.
- **Every spawn in that file is covered** — the `runner` (`:49`), the `git` helper (`:62`), the decoy `ls-files` (`:100`) and the ground-truth `git show` (`:162`). Every write is under `mkdtempSync(join(tmpdir(), …))`; every `rmSync` targets only a mkdtemp path. No path escapes `tmpdir()`.
- **The real-git suite kills real mutants**, not just argv shapes: deleting `--diff-merges=first-parent` breaks the merge-conflict test, and un-anchoring `-S` breaks the prose-collision test. The pruning test's inability to isolate `--full-history` is disclosed in-line with the measured numbers rather than glossed.
- **The seeded token is correct and durable.** Exactly one `<h1>` and exactly one `data-published-as` in the file; the `<div hidden>` survives `stripHtmlComments`; it sits outside `<table class="glance">` and outside every `<section class="group">`, so `parseLiveViewGlance` and `parseLiveViewSections` are untouched (`check:onbox-register` green confirms). `.prettierignore` already covers this file, so no formatter can reflow the attribute pair apart. A merge that duplicates the token fails closed — `parsePublishToken` reports "more than once" and the stamper refuses.
- **Entry-point guard and exit handling are both right**: `isDirectlyInvoked` rather than a hand-rolled comparison, and `process.exitCode` rather than `process.exit(main())` — correct for a script that prints three lines, per `scripts/lib/is-main-module.mjs:44-54`. Passes `entry-point-guard-convention.test.mjs` (no raw `argv[1]` token; `argv.slice(2)` is not one of the four shapes it matches).
- **House gates clear**: `Refs #2599` plain and outside code spans, issue open (`area:ops`, `type:chore`, so no `BACKLOG.md` row owed); commit subject and PR title both pass `validate-commit-msg.mjs`; the three new suites match `run-hooks-tests.mjs`'s `scripts/tests/*.test.mjs` glob; no register row added or removed, so no on-box acceptance debt. The release-notes skip checks out — #2717 is `feat(scripts,docs)` over the same register tooling and shipped without either file.

---

### Verdict

**Blocking merge: yes** — one 🔴. The scrub that this PR's own commit message identifies as the only thing standing between the suite and a destroyed repository does not cover a spelling git honours on this repo's primary platform, and the self-check is structurally unable to detect that. One-line fix, one added case.

Two 🟠 alongside it (a typo'd flag performing a write; `main()` untested), so the fix round re-triggers a pass regardless.

dudarenok-maker:
**Review findings folded — `85bce877`.**

## Blocking finding: fixed, and it corrected my own account of the incident

The `GIT_*` scrub was case-sensitive. Confirmed independently on this box (git 2.54.0.windows.1): `git` honours a lowercase `git_dir` identically to `GIT_DIR` — `rev-parse --absolute-git-dir` from an unrelated cwd resolved to the `git_dir` target. The self-check injected uppercase only, so it could never fail for the case that mattered.

Fixed by **delegating to `scrubGitEnv` (`scripts/git-env.mjs`)** rather than patching my copy. The review was right that this re-opened a hole the repo had already closed — and reading that helper corrected something I had asserted in the original commit message:

> its #2216 header records, **measured**, that an ordinary hook does **not** export `GIT_DIR` — it exports `GIT_INDEX_FILE`.

That fits the observed damage far better than my version did: the real index became a **one-entry** index holding the fixture's path, which is exactly what an inherited `GIT_INDEX_FILE` produces. I had blamed the wrong variable.

`GIT_INDEX_FILE` is deliberately excluded from that helper, and correctly so — a command reading the in-flight staged set must honour whichever index git is actually using. This file is the exception (nothing here reads the staged set; every spawn builds a disposable repo), so it scrubs that key separately, case-insensitively, with the reason stated so removing the line doesn't look like tidying.

**Self-check now runs both casings.** Both mutants are killed: the case-sensitive scrub, and dropping `scrubGitEnv` entirely.

## Significant

- **Unknown arguments fell through to the default action, which is a WRITE.** Now parses strictly, supports the `=` form, refuses anything else — as `wt-new.mjs` does.
- **`main()` had zero coverage**, which is precisely why that shipped. Six CLI tests added, driving the real script in a temp dir: a typo exits non-zero and writes **nothing**, `--check` never writes, a bare run bumps *and* re-mints, an unreadable file errors.

## Minor

`seedToken` gained both hardenings `bumpToken` applies at the identical seam; `bumpToken` refuses to exceed the parser's 15-digit cap (it was writing a token its own reader rejects); `DEFAULT_FILE` is exported and imported rather than re-typed; the register's **numbered** publish procedure gained the stamp step as step 0; and the bootstrap's remaining half is documented — the live view must be published once before the comparator can pass.

One pre-existing correction in code this work surfaced: `stripHtmlComments`' header said *"blanking rather than deleting"* while the code has always deleted (`.replace(…, '')`). I had copied that wording into the new marker comment. Both now say what the code does.

## Left open deliberately

The last minor — `publish-token.mjs` landing with no caller, so its four-lookup wiring is reviewable in neither PR — is a scoping judgement I have **not** acted on unilaterally. Splitting it out would mean landing the tests separately from the code they test. Flagged for the repo owner.

Suites: **37 + 8 + 14**, all green. `check:onbox-register` and `check:register-citations` green. Worktree index verified at 3984 tracked, unchanged.

Issue #1730, opened 2026-07-19
side — flag-ON multi-GPU device-steer atomicity (Qwen forward + ASR/SPK cold-load)


dudarenok-maker:
Follow-up discovered during #1720 (op capacity admission). Two flag-ON, multi-GPU device-steer atomicity gaps. **Both are safe while `SEG_CAPACITY_ADMISSION` is OFF (the default) and on a single-GPU box** — they gate flipping the flag ON with concurrent multi-card work, not this release. Documented in [plan 264](docs/features/264-vram-aware-gpu-placement.md) "Out of scope (flag-on-readiness follow-ups)".

### 1. Qwen forward-clobber
`QwenEngine._device` is a single field per engine, read **unlocked at forward time** (e.g. mint `rc.to(self._device)`, inside `_synth_lock` which does not hold `_cold_load_lock`). #1720 made the device-steer atomic through the model **load** (unified `_cold_load_lock`), but not the later **forward**. On a genuine multi-card box, concurrent flag-ON `design_voice`(card1) + `mint_variant`(card0) against the single qwen engine can land a *forward* on the wrong card, which the reservation ledger didn't account for.

### 2. ASR/SPK cold-load steer unlocked
The `/transcribe` and `/embed` handlers set the engine device attr (`ASR._device` / `SPK.device`) **unlocked** before the cold load. A concurrent multi-GPU *first* cold-load could land on the wrong card (benign single-GPU; after the model is resident, `is_resident` pins it to its card so subsequent admissions agree). Unlike the `/load` path, which threads `device=` as a call parameter (race-free).

### Remedy (both)
Thread the resolved device all the way into the load/forward as a parameter (into `_load_qwen_model` / the forward, and into `transcribe`/`embed`/`ensure_loaded` under the engine load lock) so nothing reads shared mutable `self._device` across an unlocked gap — the same race-free pattern the `/load` path already uses.

### Acceptance
On a 2-card box with `SEG_CAPACITY_ADMISSION=1`: concurrent design+mint (and concurrent GPU-configured transcribe/embed) each keep their *entire* op — load **and** forward — on their own admitted card; no cross-card clobber under load.

Refs #1720. Gates the concurrent-multi-card portion of the `SEG_CAPACITY_ADMISSION` default-ON flip.

dudarenok-maker:
Whole-branch review of #1720 surfaced a **third** instance of this same stale-`self._device` class:

[...]

PR #2743, opened 2026-08-28
fix(sidecar): restore WhisperEngine device pin on teardown

### LINKED ISSUE #2642: WhisperEngine overwrites its device without restoring on teardown (latent; Coqui and Kokoro both fixed)
Found by the pass-4 review gate on PR #2636 (#2631), and deliberately not fixed there — it is a pre-existing defect on a different engine, and widening that PR was the wrong call.

## What

`WhisperEngine` has the same overwrite-without-restore shape that `KokoroEngine` was just fixed for:

- `server/tts-sidecar/main.py:7547` writes the admitted device onto instance state;
- `_drop_model_locked` (`:7659-7672`) never restores it.

`CoquiEngine._drop_model_locked` (`:2110`) does restore, and its docstring names the failure this prevents (the #1730 gap-3 fix): skipping it "would leave the next lazy /synthesize cold-loading onto the last admitted card, bypassing placement entirely."

`QwenEngine` was checked and is clean.

## Why nothing is broken today

`/transcribe` re-admits on **every** call, so the stale value is overwritten before it can be read. The defect is latent, not live — which is exactly why it is worth recording rather than leaving to be rediscovered.

## Why it matters anyway

The equivalent latency on Kokoro was harmless for the same structural reason until a separate change made the field authoritative for provider selection — at which point one admitted CPU placement stuck for the process lifetime and erased an operator's `KOKORO_DEVICE=cuda:N` pin. The hazard is that a future change makes Whisper's field load-bearing and re-creates a defect that has already been fixed twice in this file.

## The fix

Mirror `CoquiEngine`: keep the env-derived value pristine, hold the per-load admitted device separately, and restore on teardown. Kokoro's fix additionally had to publish the admitted device only on a **successful** load — writing it up front left the restore unreachable behind `unload()`'s idempotence guard after a failed load. Whichever shape is chosen, all three engines should end up reading alike.

## Acceptance

Provable without hardware: an admitted device on one load must not survive a teardown into the next load's placement decision.

dudarenok-maker:
## Summary

Restores `self._device = self._requested_device` in `WhisperEngine._drop_model_locked` (`server/tts-sidecar/main.py:7804-7822`), mirroring the CoquiEngine (#1730 gap-3) and KokoroEngine (#2631 S4) fix: without it, a CPU-admitted load left `self._device` stuck at `"cpu"` past teardown instead of re-resolving from the original `ASR_DEVICE` pin on the next load. Adds a paired regression test in `test_devices.py`. Closes #2642.

## Test plan

- [x] `tests\test_devices.py` — 37/37 passed (full suite, post-fix).
- [x] Mutation-test proof: commented out `self._device = self._requested_device`, reran `-k device_pin_after_cpu_admission` — failed with `AssertionError: assert 'cpu' == 'cuda:0'` (the restored-device assertion, right reason). Restored the line, confirmed `git diff` empty, reran full suite — 37/37 passed again.
- [x] Confirmed `_ensure_loaded` (`main.py:7643-7693`) untouched by the diff.
- [x] Confirmed `CoquiEngine`/`KokoroEngine`/`QwenEngine` untouched — diff is `main.py` (+5) and `tests/test_devices.py` (+30) only.

## Review

Independent PR review pass (medium depth) verified the test plan (37/37, mutation test reproduces) and posted findings: https://github.com/dudarenok-maker/Castwright/pull/2743#issuecomment-5450222309

Also fixed, found in passing: this PR was missing its mandatory release-notes pairing (`docs/release-notes-next.md` + `RELEASE_NOTES.md`) — added in `05e0649b`.

One finding was deferred rather than fixed here: the review also surfaced that `SpeakerEngine` has the same class of device-pin bug, but worse (it self-demotes cuda->cpu at runtime rather than losing the pin via CPU admission, and its own idle-teardown gate never fires once demoted, since it's gated on the very field the bug corrupts) — filed as #2750, since fixing it needs a design decision between more than one defensible shape, not a mechanical mirror of this PR's fix.


dudarenok-maker:
## PR review — pass 1 (head e3b2fd4eaa150e88faa0415a714a53264db2917d, depth medium)

Scope: the full diff — `server/tts-sidecar/main.py` (+5) and `server/tts-sidecar/tests/test_devices.py` (+30) — plus the surrounding `WhisperEngine` / `SpeakerEngine` lifecycles, `PlacementController._resolve_admission`, the `/transcribe` and `/embed` routes, `_engine_actual_card` / `_is_resident`, and the Node-side consumers of `asr_device`. Verified independently before probing: `test_devices.py` at this head is **37/37 green** (run against a temp copy of the head tree with the repo's own sidecar venv), and the mutation claim reproduces exactly — replacing `main.py:7821`'s `self._device = self._requested_device` with `pass` fails **one** test, `test_whisper_drop_model_locked_restores_device_pin_after_cpu_admission`, with `AssertionError: assert 'cpu' == 'cuda:0'` at the restored-device assertion. Red for the reason claimed. The test plan is credible as written.

### 🟠 Significant — the class is closed on three engines and left armed on the fourth, which is the only one that can actually diverge

**Correctness bug.** `server/tts-sidecar/main.py:8037-8050` — `SpeakerEngine._drop_model_locked` nulls `self._model` and never restores `self.device = self._requested_device`. Its own `__init__` at `main.py:7942` captures the field for exactly that purpose:

```python
self._requested_device = self.device  # preserved before any cpu-demotion
```

…and no cpu-demotion is ever undone anywhere in the file. `git grep _requested_device` shows SPK's copy is read by exactly one caller — `_engine_actual_card`'s `/health` `fell_back` badge — never to re-resolve a load.

This is not a symmetry nit. `SpeakerEngine` is the **only** engine in this file that mutates its device field *outside* admission: `ensure_loaded` self-demotes cuda→cpu at three sites of its own — `main.py:7983` (`torch.cuda.is_available()` false), `main.py:7985` (the torch import/probe itself raising), and `main.py:7995` (a non-poison cuda load failure). `server/src/tts/transcribe-client.ts:69-76` names this asymmetry outright as the reason the Node side is allowed to treat configuration as truth for ASR: *"unlike `SpeakerEngine` (which falls back cuda -> cpu on a load failure), a CUDA load failure in `_ensure_loaded` propagates as an exception rather than degrading."*

Concrete break — `SPK_DEVICE=cuda:0`, capacity admission on:

1. `/embed` #1: `on_gpu` true → `_placement.reservation("spk", …)` admits `cuda:0` → `ensure_loaded("cuda:0")` sets `self.device = "cuda:0"`.
2. `_load_on` raises a non-poison CUDA error (cuDNN/driver mismatch on a present GPU — the exact case that branch exists for). `main.py:7994-7996` sets `self.device = "cpu"` and reloads on CPU. The embed succeeds; nothing 500s.
3. From here the demotion is permanent for the process lifetime:
   - `maybe_free_idle` (`main.py:8093`) fast-outs on `_parse_device(self.device)[0] != "cuda"`, so the engine is **never torn down** — `_drop_model_locked`, the one place a restore could live, is never reached (and `SPK.unload()` has no production caller; `git grep 'SPK\.'` finds only `maybe_free_idle`, `/health`, and `/embed`);
   - `/embed`'s gate at `main.py:11429` — `on_gpu = _parse_device(SPK.device)[0] in ("cuda", "rocm")` — reads that same demoted field, so every later `/embed` skips the reservation path entirely and calls `ensure_loaded()` with `device=None`, which re-reads `self.device == "cpu"`;
   - the operator's `SPK_DEVICE=cuda:0` pin is silently dropped with no recovery short of a sidecar restart, while `/health` reports `spk_device: "cpu"` and `fell_back: true` forever.

`docs/features/264-vram-aware-gpu-placement.md:526-531` already states the contract this violates: *"any future code that nulls or reassigns a model field must mirror `_drop_model_locked()`'s full teardown (including the device restoration that `_drop_model_locked()` already handles)."*

Note the split honestly: SPK's **missing restore in `_drop_model_locked`** is latent in the same way Whisper's was (the only shape that reaches it — an admitted card differing from an unpinned `SPK_DEVICE=cuda` — is re-admitted on the next call anyway). The **live** half is that a demoted SPK never reaches a teardown at all, so there is no point at which the pin is reconsidered. Both legs are the same defect — a pref field overwritten by an outcome with no restore — and both are on the engine `#2642`'s audit skipped: the issue records *"`QwenEngine` was checked and is clean"* and says *"all three engines should end up reading alike,"* enumerating Coqui/Kokoro/Whisper/Qwen and never SpeakerEngine.

### 🟡 Minor — the comment, the PR body and the test docstring all assert a CPU-admission path that cannot occur

**Cleanup / documentation defect.** The added comment at `main.py:7817-7821` says *"`_ensure_loaded` overwrote `self._device` with the resolved card **on a CPU admission**"*; the PR body says *"a CPU-admitted load left `self._device` stuck at `\"cpu\"` past teardown"*; the new test's docstring says *"a CPU-admitted load stuck there until process restart"*, and the test's name and its `# Admission ledger refuses the GPU for this cold load.` comment encode the same premise.

`/transcribe` passes `cpu_capable=False` (`main.py:11354`), and `_resolve_admission`'s only cpu branch is `elif cpu_capable and not heavy: admission = {"device": "cpu"}` (`main.py:4771-4772`) — so ASR admission returns a GPU key or `noCapacity`, never `"cpu"`. There is no `"asr"` entry in `_ENGINE_CAPACITY_PROFILE` that could change that, and `/transcribe` is the only caller of `ASR.transcribe` / `_ensure_loaded(device=…)` in the process. The file says as much 2000 lines down, at `main.py:9785-9788`: *"no production path currently reaches it: /transcribe always passes cpu_capable=False, so admission returns either a GPU key or noCapacity, never a divergence."* So does the issue, under **"Why nothing is broken today."**

Two consequences worth fixing in this round rather than leaving on the record:

- A future reader hitting the new comment will reason from a premise the same file contradicts — and the one divergence that *is* reachable (unpinned `ASR_DEVICE=cuda` on a multi-GPU box, admitted to `cuda:1` while `_requested_device` stays `"cuda"`) is the one the comment doesn't name.
- `#2642`'s own stated reason for the defect being latent is wrong in a way this PR inherits: *"`/transcribe` re-admits on every call, so the stale value is overwritten before it can be read."* The stale value **is** read first — `main.py:11350`'s `on_gpu = _parse_device(ASR._device)[0] in ("cuda", "rocm")` is what decides whether to re-admit at all, so a `_device` stuck at `"cpu"` would suppress the very re-admission that reason relies on. The actual reason nothing breaks is `cpu_capable=False`, one line down. The fix is right; the stated mechanism is not.

Suggested reword: state that the restore is **latent by construction** (ASR admission cannot return `cpu`), that it exists to keep the four engines reading alike per plan 264 §3, and name the multi-GPU `cuda`→`cuda:1` case as the reachable divergence. Retitling the test away from `after_cpu_admission` would help too, though the assertion itself is sound either way.

### 🟡 Minor — no release-notes pair and no explicit not-applicable

**Cleanup / house gate.** CLAUDE.md's before-shipping step 5 requires an entry in both `docs/release-notes-next.md` and `RELEASE_NOTES.md`, *or* an explicit skip statement — *"say so explicitly rather than silently omitting."* Neither file mentions `#2642` or `#2743`, the diff touches neither, and the PR body's Test plan says nothing about it. Given the defect is latent (no operator-visible delta), a skip is defensible on the merits; what is missing is the sentence saying so.

### ✅ What is solid

- **The fix itself is correct and minimal**, and it lands on the right side of the trap `#2642` flagged from Kokoro's `S4`: `_ensure_loaded` publishes `self._device = dev` *after* `WhisperModel(...)` returns (`main.py:7686-7692`), so a failed load never writes the field, and `_drop_model_locked`'s `if self._model is None: return False` early-out therefore never strands an unrestored value behind `unload()`'s idempotence guard.
- **The write ordering inside `_drop_model_locked` is right, and would be a real bug reversed.** `self._model = None` runs *before* the device restore, both under `_infer_lock`, while `/health`'s `_engine_actual_card` reads both without the lock. Because the model is nulled first, a racing reader either sees `_model is None` (returns `None`, no card) or sees the pre-restore actual device — there is no window where a *loaded* engine reports its pref as its actual card and manufactures a false `fell_back: false`.
- **No `/health` or placement regression.** `_engine_actual_card` returns `None` while `_model is None`, so `_is_resident("asr")` is unaffected by the restored value, and the `asr_device` string surfaced at `main.py:10021`/`10118` → `sidecar-health.ts:490` → `diagnostics.ts:244` is display-only and, given no divergence is reachable, byte-identical to before in every production state.
- **The test respects the documented precondition** — it takes `engine._infer_lock` before calling `_drop_model_locked`, rather than calling it bare — and uses a fresh `WhisperEngine()` plus `monkeypatch`, so it leaves the `ASR` singleton and `sys.modules` clean.
- **Citations check out.** CoquiEngine's restore is at `main.py:2110` with the `#1730` gap-3 rationale at `:2096`; Kokoro's is at `main.py:3513` (plus its `_resolved_device = None` reset at `:3524`, which Whisper correctly has no equivalent of). `Closes #2642` is present, plain, outside any code span.
- **On-box acceptance is not owed** — `#2642`'s own Acceptance line says *"Provable without hardware,"* and the added unit test is that proof.

### Verdict

The change is correct, minimal, well-cited and genuinely regression-tested — the mutation proof reproduces exactly as claimed and the 37/37 figure is accurate. It should not merge as-is, for one reason: it is the *third* fix of a four-instance class, and the instance it leaves armed (`SpeakerEngine`) is the only one whose device field can actually diverge in production. Fixing Whisper while skipping SPK closes the latent case and leaves the live one — the inverse of the intended outcome. Fix the 🟠, fold the two 🟡s in the same round, and this is ready.

*Findings only — nothing applied. No tracked file was modified by this pass.*

PR #2744, opened 2026-08-28
fix(scripts,docs): close the three review findings left unfixed by #2740

dudarenok-maker:
## Summary

#2740's review gate produced ten findings. Nine were folded before merge; **three were not**, on the reasoning that they were Low and red-direction — none could turn a STOP green.

That is a bar this repo does not have. CLAUDE.md requires findings to be fixed in the round that surfaced them, and lists *"it's only a chore"* / *"it's not user-visible"* / *"nothing is broken yet"* among the excuses it explicitly voids — severity is those with a number attached. The three took one round to close, which is the tell: **a finding cheap enough to rank Low is cheap enough to fix now.**

This PR closes those three, then folds everything its own two review passes found.

### F7 — a non-string copy blamed the wrong file

`readFileSync` without an encoding returns a Buffer. `parsePublishToken` returned `null` for it, which means *"this file has no token"*, so the comparator answered *"the tracked live view has none. Restore it before publishing."* It sent an operator to repair a perfectly good file while the real fault — a caller bug one frame up — went unnamed.

Non-string now reports as malformed and names the Buffer case. `null`/`undefined` still return `null`: those genuinely mean *no copy*, and each already has its own constant.

### F5 — half an ordering claim was unpinned

The staleness gate's comment asserts it runs before the counter gates **and** before the both-absent gate. Only the first half was pinned — relocating the gate below both-absent survived mutation.

### F8 — the boundary of `baselineInMine` was undocumented

It asks whether main's nonce appears *anywhere* in `HEAD`'s history for that path, and **the rebase alone satisfies it**. Measured in a real repository:

| state | commits selected |
|---|---|
| before rebase | **0** — STOP fires |
| after rebase, before any stamp | **1** — STOP already cleared |
| after stamping | 2 — redundant second match |

So the staleness STOP is cleared by rebasing, which is exactly what its message instructs. The cost of asking a *history* question rather than a *content* one is that a revert, or a wholesale "take mine" resolution after a clean rebase, also goes green — adjudicated, not overlooked: closing that means comparing content, the undecidable question the three rejected per-row designs died on.

> An earlier revision of this PR claimed the removal-match was "what lets a correctly rebased lane clear the STOP after one stamp". **That inverted the causation and is retracted** — the correction is left visible in the spec rather than edited out, because it is the fifth instance of the failure mode that document records, and it happened in prose describing that very failure.

## Review passes

**Pass 1** — 3 Significant + 1 Minor, all folded in `2b022364`:

- **The F7 fix re-created its own defect at the parser's other caller.** `seedToken` read `parsePublishToken(html) !== null` as a binary, and `{ malformed } !== null` — so seeding a Buffer containing a file with *no token* threw *"this file already has a token"*. Reading the parser as a binary **is** the defect F7 fixes; I fixed one site and mirrored it into the other. All three outcomes now handled; the class audited tree-wide.
- **Test `18m` was satisfied by a different guard than the one it names** — both messages contain `without minting`, so it passed via the history-backed guard while the guard it is named for could be deleted with the suite green.
- **The spec's causal claim was inverted** (above).
- The tokenless-working STOP had no positive test.

**Pass 2** — nothing blocking; all four verified resolved by mutation, and the guard-deletion sweep went from **2 survivors out of 24 to 0 out of 38**. Its five Minors folded in `473456dc`:

- **`18f` could not fail for the boundary it names.** It inherited `workingInBaseline: false`, which neuters its own negative assertion — a `w.n >= b.n` mutant consults the lookup, finds `false`, declines to STOP, and reaches the same verdict. Now passes `null`, the one value whose consultation is *visible*.
- `--check`'s malformed branch, deleted, exited **0** printing `is at undefined (id undefined)` — a broken token reported as healthy, in the mode whose only job is reporting token state. Plus its tokenless branch, and `seedToken`'s mint validation (untested while its `bumpToken` twin was pinned).
- Documented `w.n === b.n → w.n <= b.n` as a third known-equivalent mutant so the next sweep does not re-chase it.

## Recorded, not fixed — not fixable here

`publish-token.mjs` merged in #2740 with **no caller**, so its comparator was reviewed as *code* but never as *wiring*, and its riskiest property — the four-lookup contract — has not been exercised against a real one. Added as **fact 10** on plan Task 10: that task's review is the module's *first review in use*, not a check that wiring landed.

## Test plan

**41 unit + 8 real-git + 18 stamper — all green**; `check:onbox-register` green.

Every fix is mutation-verified, not merely passing. Killed: the non-string branch, the relocated staleness gate, guard 243, guard 123, `seedToken`'s malformed branch, `w.n > b.n → >=`, `--check`'s malformed and tokenless branches, and `seedToken`'s mint validation.

Release notes: skipped, same basis as #2740 — maintainer tooling with no user- or operator-visible effect.

Refs #2599

### DIFF
--- a/docs/superpowers/plans/2026-08-27-onbox-register-stable-row-ids.md
+++ b/docs/superpowers/plans/2026-08-27-onbox-register-stable-row-ids.md
@@ -1057,8 +1057,9 @@ Already shipped, consumed as-is (do not re-declare or re-implement):
     lookups: { inBaseline, inMine, baselineInMine, workingInBaseline }
   PUBLISH_TOKEN_BASELINE_ERROR / _PUBLISHED_ERROR / _WORKING_ERROR   (three, not two)
 
-**Nine facts this task must not get wrong** *(1-7 as originally written, with 2
-corrected; 8 and 9 added after the code was executed and reviewed):*
+**Ten facts this task must not get wrong** *(1-7 as originally written, with 2
+corrected; 8 and 9 added after the code was executed and reviewed; 10 after it
+merged):*
 
 1. **Three designs have died on a hand-maintained identity field** — a bare
    counter, a branch name, and a nonce that the operator had to remember to
@@ -1108,6 +1109,22 @@ corrected; 8 and 9 added after the code was executed and reviewed):*
    writes to the real repository — observed twice, taking 3,979 tracked files
    down to 1. Its first test asserts the scrub works; that assertion is the only
    thing that would catch a regression, because the failure is silent.
+
+   Two corrections to that account, both from the PR #2740 review. The scrub is
+   delegated to `scrubGitEnv` (`scripts/git-env.mjs`) and matches
+   **case-insensitively** — git honours a lowercase `git_dir` identically, and
+   a `startsWith('GIT_')` filter leaves that survivor in place. And an ordinary
+   hook does **not** export `GIT_DIR`; it exports `GIT_INDEX_FILE` (that
+   helper's own measured #2216 note), which is what actually did the damage.
+
+10. **This task carries a review debt it did not incur.** `publish-token.mjs`
+    merged in PR #2740 with **no caller** — only `bumpToken`/`parsePublishToken`
+    are used, by the stamper. Its comparator was reviewed as *code*, hard, but
+    never as *wiring*, and the reviewer named that gap: the module's own riskiest
+    property is the four-lookup contract in fact 8, and no PR has yet exercised
+    it against a real caller. **So this task's review is not "did the wiring
+    land" — it is the first review of the module in use.** Budget for that
+    rather than treating the module as settled because it is on `main`.
 
 - [ ] **Step 1: Write the failing tests**
 
--- a/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
+++ b/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
@@ -462,7 +462,7 @@ the query then selects 2 commits rather than 1 — but that second match is
 **redundant**: the commit that ADDED main's nonce is already in your history
 once you have rebased. The stamp answers a different gate, `w.n === b.n`. The
 claim was written from the previous draft rather than from a probe, which is
-the failure mode this document has recorded four times and is the reason the
+the failure mode this document has recorded five times and is the reason the
 correction is left visible here rather than quietly edited out.)*
 
 That residue is adjudicated, not overlooked. Detecting it means comparing
--- a/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
+++ b/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
@@ -447,13 +447,23 @@ rather than being implied.
 
 **Second: `baselineInMine` proves your branch HELD main's live view, not that
 your working file still reflects it.** The lookup asks whether main's nonce
-appears anywhere in `HEAD`'s history for that path — and a commit that *removes*
-that nonce satisfies it just as well as one that adds it. Your own stamp is
-exactly such a commit, which is not incidental: it is what lets a correctly
-rebased lane clear the staleness STOP after one stamp, and it is why the STOP is
-self-clearing rather than a dead end. The cost is that a revert, or a wholesale
-"take mine" conflict resolution *after* a clean rebase, also satisfies it and
-goes green.
+appears anywhere in `HEAD`'s history for that path, and it is **the rebase
+alone that satisfies it** — measured in a real repository: before rebasing the
+query selects 0 commits, after rebasing and *before any stamp* it selects 1.
+So the staleness STOP is cleared by rebasing, which is exactly what its message
+instructs. The cost of asking a history question rather than a content one is
+that a revert, or a wholesale "take mine" conflict resolution *after* a clean
+rebase, also satisfies it and goes green.
+
+*(An earlier draft of this paragraph said the removal-match was "what lets a
+correctly rebased lane clear the STOP after one stamp". That inverts the
+causation. `-S` does match a removal — your own stamp removes main's nonce and
+the query then selects 2 commits rather than 1 — but that second match is
+**redundant**: the commit that ADDED main's nonce is already in your history
+once you have rebased. The stamp answers a different gate, `w.n === b.n`. The
+claim was written from the previous draft rather than from a probe, which is
+the failure mode this document has recorded four times and is the reason the
+correction is left visible here rather than quietly edited out.)*
 
 That residue is adjudicated, not overlooked. Detecting it means comparing
 *content* between the working file and the baseline — which is precisely the
--- a/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
+++ b/docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md
@@ -436,12 +436,32 @@ published page, and the tracked working file** — they fail for different reaso
 and the operator does something different about each. Two constants would force
 two of those three onto one message.
 
-**Stated boundary — one, not two.** The token proves the live page came out of
-your history and that nobody else published since your baseline. It does **not**
-prove the bytes you are about to publish are the bytes you intended: a stale local
-build of your own file, correctly bumped and correctly nonced, still publishes.
-That is what git review covers, and it goes in #2599's close comment rather than
-being implied.
+**Stated boundary — two, and both are worth saying out loud.**
+
+**First: the token proves provenance, not intent.** It shows the live page came
+out of your history and that nobody else published since your baseline. It does
+**not** prove the bytes you are about to publish are the bytes you intended: a
+stale local build of your own file, correctly bumped and correctly nonced, still
+publishes. That is what git review covers, and it goes in #2599's close comment
+rather than being implied.
+
+**Second: `baselineInMine` proves your branch HELD main's live view, not that
+your working file still reflects it.** The lookup asks whether main's nonce
+appears anywhere in `HEAD`'s history for that path — and a commit that *removes*
+that nonce satisfies it just as well as one that adds it. Your own stamp is
+exactly such a commit, which is not incidental: it is what lets a correctly
+rebased lane clear the staleness STOP after one stamp, and it is why the STOP is
+self-clearing rather than a dead end. The cost is that a revert, or a wholesale
+"take mine" conflict resolution *after* a clean rebase, also satisfies it and
+goes green.
+
+That residue is adjudicated, not overlooked. Detecting it means comparing
+*content* between the working file and the baseline — which is precisely the
+question the three rejected per-row designs could not answer, for a reason that
+has not changed: content comparison cannot separate "this branch edited the row"
+from "this branch is stale". The token deliberately answers a different, decidable
+question instead. What closes this last gap is git review of the diff, not a
+richer comparator.
 
 *(The previous draft carried a second boundary — that the publisher field was
 "trusted, not proven", justified as detecting accident rather than forgery. That
--- a/scripts/publish-token.mjs
+++ b/scripts/publish-token.mjs
@@ -29,7 +29,20 @@ export const publishTokenRegex = () =>
   /data-published-as="([^"]*)"\s+data-publish-id="([^"]*)"/g;
 
 export function parsePublishToken(rawHtml) {
-  if (typeof rawHtml !== 'string') return null;
+  // `null`/`undefined` mean "no copy" and are the caller's to diagnose — the
+  // comparator answers each with its own named constant before parsing.
+  if (rawHtml === null || rawHtml === undefined) return null;
+  // Anything else non-string is NOT "this file has no token": it is a caller
+  // that handed us the wrong kind of value, and `readFileSync` without an
+  // encoding — which returns a Buffer — is the realistic way to get here.
+  // Returning null routed that into "the tracked live view has none. Restore
+  // it before publishing", sending an operator to repair a file that is
+  // perfectly fine. Report it as malformed so the message names the real fault.
+  if (typeof rawHtml !== 'string') {
+    return {
+      malformed: `expected HTML text but got ${rawHtml instanceof Uint8Array ? 'a Buffer — read the file with an encoding' : typeof rawHtml}`,
+    };
+  }
   const matches = [...rawHtml.matchAll(publishTokenRegex())];
   if (matches.length === 0) return null;
   if (matches.length > 1) {
--- a/scripts/stamp-publish-token.mjs
+++ b/scripts/stamp-publish-token.mjs
@@ -47,7 +47,17 @@ export const mintNonce = () => randomBytes(6).toString('hex');
 const SEED_AFTER = /(<h1[^>]*>[\s\S]*?<\/h1>)/;
 
 export function seedToken(html, mint = mintNonce) {
-  if (parsePublishToken(html) !== null) {
+  // THREE outcomes, not two. `parsePublishToken` returns null (no token),
+  // `{ malformed }` (a token, or a value, this reader will not accept), or a
+  // parsed token — and `{ malformed } !== null`, so a binary `!== null` test
+  // reports "this file already has a token" for a file with NO token in it.
+  // Reading the parser as a binary is precisely the defect this PR fixes in
+  // the comparator; it was mirrored here, at the parser's only other caller.
+  const existing = parsePublishToken(html);
+  if (existing && existing.malformed) {
+    throw new Error(`stamp-publish-token: ${existing.malformed}`);
+  }
+  if (existing !== null) {
     throw new Error('stamp-publish-token: this file already has a token; stamp it, do not seed it.');
   }
   if (!SEED_AFTER.test(html)) {
--- a/scripts/tests/publish-token.test.mjs
+++ b/scripts/tests/publish-token.test.mjs
@@ -307,6 +307,26 @@ test('18g STALE LANE: told to bump, obeys, and must NOT go green', () => {
   assert.notDeepEqual(step2, [], 'bumping must not clear a staleness STOP');
 });
 
+test('18g2 the staleness STOP outranks the both-absent one, not just the counters', () => {
+  // The OTHER half of the ordering claim beside the !baselineInMine guard. 18g
+  // pins that it precedes the COUNTER gates; nothing pinned that it precedes
+  // the both-absent gate, so moving it below that one survived mutation. Both
+  // messages say "rebase", so this is message SELECTION rather than a
+  // correctness hole — but an unpinned ordering claim is one edit from being
+  // false, and the comment asserts it.
+  const e = cmp({
+    working: T(49, 'mine'),
+    published: T(48, 'other'),
+    baseline: T(47, 'mainn'),
+    ...L({ inBaseline: false, inMine: false, baselineInMine: false }),
+  });
+  assert.ok(
+    e.some((x) => x.includes("origin/main's live-view nonce")),
+    `staleness must be diagnosed ahead of "another lane published": ${e.join('|')}`,
+  );
+  assert.ok(!e.some((x) => x.includes('another lane published')), e.join('|'));
+});
+
 test('18h green REQUIRES the baseline nonce to be in my history', () => {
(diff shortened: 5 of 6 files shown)


dudarenok-maker:
## PR review — pass 1 (head 4d248f8a, depth high)

Scope: all 4 files (`scripts/publish-token.mjs`, `scripts/tests/publish-token.test.mjs`, the spec, the plan) · verified: **39 unit + 8 real-git + 14 stamper = 61 green**, `check:onbox-register` OK, commit subject passes `validate-commit-msg.mjs` (exit 0) — before adversarially probing the F7 null contract, both new tests by mutation, a guard-deletion sweep over all 24 comparator guards, and the two doc claims against real git.

Dispatched as `medium` on "single-scope `fix`". The commit and PR title are `fix(scripts,docs)` — CONTRIBUTING.md's documented multi-scope form — and the ladder puts *any multi-scope PR* at `high`. Ran at `high`; recording the discrepancy rather than the dispatched value.

### 🟠 Significant — the F7 fix re-creates its own defect at the one real caller

**Correctness bug (latent).** `scripts/stamp-publish-token.mjs:50`

`seedToken` reads the parser's result as a binary:

```js
if (parsePublishToken(html) !== null) {
  throw new Error('stamp-publish-token: this file already has a token; stamp it, do not seed it.');
}
```

This PR added a third return shape. `{ malformed }` is `!== null`.

Measured, both ways — input `Buffer.from('<h1>Register</h1>\n<p>a page with NO token at all</p>\n')`:

| | outcome |
|---|---|
| parent commit `bde8b559` | `TypeError: html.replace is not a function` |
| this PR `4d248f8a` | `stamp-publish-token: this file already has a token; stamp it, do not seed it.` |

The file has **no token at all**. Before, the caller got an obscure crash; now it gets a confident, false statement about the file — which is precisely the defect F7 exists to close ("it sent an operator to repair a perfectly good file"), re-created at the sibling call site by the same change.

Not reachable from the CLI today: `main()` reads with `'utf8'` at `stamp-publish-token.mjs:125`, so `html` is always a string. It is latent, not live — but `seedToken` is exported, and `stamp-publish-token.mjs:50` is the **only** binary read of a parse result left in the tree (audited: `publish-token.mjs:112/118/123/124/333-334` and `stamp-publish-token.mjs:131/134/138/150` all handle the three shapes). `main()` already branches correctly at lines 134/138; `seedToken` should do the same.

Worth noting the module keeps a "count them here whenever one is added" discipline for the four lookups and the three git semantics, but has no equivalent for *consumers of the parse result* — which is why widening that return went unaudited at the one site that reads it as a boolean.

### 🟠 Significant — guard 243 is unprotected, and `18m`, the test named for it, is satisfied by a different guard

**Correctness bug (test defect; demonstrated fail-green hole).** `scripts/publish-token.mjs:243`, `scripts/tests/publish-token.test.mjs:395`

Disabling the no-history-lookup "bumped without minting" STOP leaves the suite **39/39 green**, and turns this input into `[]` — GREEN:

```
working:   48 / "samexx"      lookups: inBaseline        false
published: 47 / "samexx"               inMine            true
baseline:  47 / "aaaxxx"               baselineInMine    true
                                       workingInBaseline false
```

A lane that bumped its own counter by hand without minting a new id — the exact class this module exists to stop — publishes green.

Why nothing catches it: `18m` ("a counter bumped WITHOUT minting is caught with **no history lookup**") feeds `working 49/"samexx", published 48/"samexx", baseline 47/"aaaxxx"` with `HEALTHY` lookups. That has `inBaseline: true` and `p.n > b.n`, so it returns at **line 170** — the *history-backed* stale-nonce guard — whose message also contains `without minting`:

> "…its nonce ("samexx") is already in origin/main's history. A lane published **without minting** a new id."

The substring collision between the two messages is what hides it. The test never reaches the guard it is named for.

Guard-deletion sweep over all 24 comparator guards: 22 killed, 2 survived (this one and the 🟡 below).

### 🟠 Significant — the spec's new "second boundary" gets its own causal claim backwards

**Correctness bug (documentation).** `docs/superpowers/specs/2026-08-27-onbox-register-stable-row-ids-design.md`

> "Your own stamp is exactly such a commit, which is not incidental: **it is what lets a correctly rebased lane clear the staleness STOP after one stamp**, and it is why the STOP is self-clearing rather than a dead end."

The **mechanism** half is correct and I confirmed it on real git — `-S` is a change-in-occurrence-count search, so removals match as well as additions. `git log -S 'data-publish-id="bbbbbb"' -- docs/live.html` selects **2** commits: main's add *and* the lane's stamp, which removes it.

The **causal** half is wrong. Same throwaway repo, lane branched at `aaaaaa` while main stamped to `bbbbbb`:

| step | `baselineInMine('bbbbbb', HEAD)` |
|---|---|
| stale lane | `false` — STOP fires |
| **after rebase, before any stamp** | **`true`** |
| after the stamp removes `bbbbbb` | `true` |

The **rebase** clears the staleness STOP — main's add-commit becomes an ancestor. The stamp is required by a *different* STOP (`w.n === b.n`, line 256), and after it the answer stays `true` via that same add-commit, so the removal match is **redundant here, not load-bearing**.

This matters more than a usual doc nit because fact 10 designates this spec as the brief for Task 10, and the recorded failure mode on this module is prose reasoned from prose. A later round reading this paragraph would conclude that tightening the lookup to additions-only breaks self-clearing; measurement says it would not.

The *residue* the paragraph describes — a revert, or a wholesale "take mine" after a clean rebase, also goes green — is accurate and unaffected.

### 🟡 Minor

**The tokenless-working STOP has no positive test.** `scripts/publish-token.mjs:123` — the other sweep survivor. Deleting it is *not* fail-green (it STOPs earlier via line 170, or throws on `w.n` at line 225), so this is coverage rather than a bug. Notable only because the new `18o` asserts that message is **absent** (`!e.some((x) => x.includes('Restore it'))`) while nothing in the suite asserts it is ever **present**.

### ✅ What is solid

- **The F7 null contract is intact.** Traced every path. `comparePublishTokens` guards `null`/`undefined` on all three inputs at lines 98-100 *before* parsing, so the parser can never see them from there. A genuinely tokenless-but-valid HTML string still returns `null` at line 47 and still reaches the same `!b`/`!w`/`!p` branches — measured: `'<p>no token here at all</p>'` → *"the tracked live view has none. Restore it before publishing."*, unchanged. Non-string inputs moved from one STOP to a different STOP. No path softened, none hardened, no direction change. `bumpToken` (333-334) already handled both shapes.
- **`18o` is a real regression test.** Reverting the production hunk kills it and only it (39→38), on `assert.ok(parsed && parsed.malformed)` — the reason claimed. Its comparator-level assertions are independently load-bearing: under the revert the message is exactly `Restore it before publishing`, so both the positive (`encoding`) and negative (`Restore it`) assertions would fail. Not vacuous through `cmp`'s defaults.
- **`18g2` is a real regression test.** Relocating the `!baselineInMine` gate below `!inBaseline && !inMine` kills it and only it. It reaches line 197 with no earlier return (traced: `behind` false, no malformed, consistency and stale-nonce both skipped on `inBaseline: false`), and `origin/main's live-view nonce` occurs in exactly one *returned* message — the second hit is a comment at line 188.
- **Fact 10 is accurate.** `scripts/stamp-publish-token.mjs:29` is the only importer and takes only `bumpToken`/`parsePublishToken`; no dynamic `import()`/`require` anywhere. `comparePublishTokens` and `nonceInHistory` genuinely have no production caller.
- **House gates.** `Refs #2599` present, plain, outside code spans. Commit subject valid (`scripts`/`docs` both allowed; comma-no-space multi-scope form). PR title matches the subject. Release-notes skip is justified — the F7 path is unreachable from the CLI, so there is no user- or operator-visible delta. No on-box acceptance owed; `check:onbox-register` green.
- **The `GIT_INDEX_FILE` scrub held.** Worktree after the real-git suite: HEAD `4d248f8a`, clean, 3984 tracked files. Primary checkout untouched. All mutation work was done on copies under the OS temp dir and restored (verified byte-identical to the worktree copy).

### Verdict

Nothing here is a live fail-green in shipping behaviour, so no 🔴. But all three 🟠 are correctness defects under this repo's own fix-in-the-round rule, and the first is a **regression this PR introduces**: it widens a return shape and leaves the one caller that reads it as a boolean now asserting something false about the operator's file. The second is a demonstrated fail-green hole whose guarding test pins the wrong guard. The third is wrong prose in the paragraph added to close F8, in a spec that is the brief for the next task.

Recommend fixing all three before merge — each is small (a three-way branch, one test, one paragraph). Fixing and pushing re-triggers one re-review round.

dudarenok-maker:
**All four findings folded — `2b022364`. None deferred.**

## 1. The F7 fix re-created its own defect at the sibling site

You were right, and it's the worst of the four because of *where* it is. `seedToken` read `parsePublishToken(html) !== null` as a binary, and `{ malformed } !== null` — so once F7 made non-string values malformed rather than null, seeding a Buffer containing a file with **no token** threw *"this file already has a token; stamp it, do not seed it."*

Reading the parser as a binary **is** the defect F7 fixes. I fixed one call site and mirrored the bug into the other. All three outcomes are now handled explicitly, with a regression test pinning that a Buffer does *not* get the "already has a token" message.

## 2. Test 18m was satisfied by a different guard than the one it names

Confirmed by deletion: with guard 243 disabled the suite stayed **41/41 green**. Its input carried the default `inBaseline: true`, so the history-backed guard fired first — and both messages contain `without minting`, so the assertion passed via the wrong one.

Now uses `inBaseline: false` to actually reach it, and asserts on the half of the message **unique** to that guard rather than the phrase both share. Added `18m2` to pin that the two are distinguishable rather than interchangeable.

## 3. The spec's "second boundary" inverted its own causation

Measured in a real repository rather than re-reasoned:

```
before rebase                    -> 0 commits   (STOP fires)
after rebase, BEFORE any stamp   -> 1 commit    (STOP already cleared)
after the stamp                  -> 2 commits   (redundant second match)
```

The **rebase** clears it — which is exactly what the STOP's own message instructs. `-S` does match a removal, but that match is redundant: the commit that *added* main's nonce is already in history once rebased. Rewritten from the measurement, with the correction left visible rather than edited out — this is the fifth time in this document's history, and it happened in prose describing that very failure mode.

## 4. Minor — the tokenless-working STOP had no positive test

Added. Your guard-deletion sweep was the right instrument; I've adopted it.

**All three previously-unprotected guards now die under mutation:** guard 243 (un-minted, no lookup), guard 123 (tokenless working), and `seedToken`'s malformed branch.

## Your process note

Taken, and it was mine to get right: I dispatched at `medium` on "single-scope `fix`", but my own commit subject is `fix(scripts,docs)` — the multi-scope form the ladder puts at `high`. Thank you for running it at high anyway. A depth chosen from misreading my own subject line would have been a quieter failure than any finding above.

## Unrelated, surfaced during the fold

`check-onbox-register.test.mjs:2811` — *"#2272 review: an unconsumed --discharging name AND a genuine BEHIND row…"* — failed **once** under `test:hooks`, then passed standalone (116/116), on merged `main` (116/116), and on a full re-run (1674/1674). This branch touches no file it reads. Not bypassed; re-run, not `--no-verify`.

Flagging rather than diagnosing: the earlier design review noted `ONBOX_TEST_BASELINE_FILE` produces no SHA and warned that a CLI test deriving its verdict from live git state is a latent bug. That would be timing-sensitive in exactly this shape. **That's a hypothesis, not a finding** — I have not demonstrated it.

Tests: **41 unit + 8 real-git + 15 stamper**, all green. `check:onbox-register` OK.

dudarenok-maker:
## PR review — pass 2 (head 2b022364, depth high)

Scope: the fold commit's 4 files plus the whole of `scripts/stamp-publish-token.mjs` and `scripts/check-onbox-register.mjs`'s baseline seam · verified: **41 unit + 8 real-git + 15 stamper green**, full `run-hooks-tests.mjs` **1669/1669 green**, `check:onbox-register` OK — before re-deriving every prior finding by mutation, sweeping all **38** guards in `publish-token.mjs` and all **12** in `stamp-publish-token.mjs`, adding a boundary-operator sweep the deletion sweep is blind to, and re-measuring the spec's git claim in a fresh throwaway repository with a control arm.

Depth re-derived from the current commit set: still `high` (`fix(scripts,docs)`, multi-scope).

---

|---|---|---|
| 1 | 🟠 `seedToken` read the parser as a binary | **VERIFIED RESOLVED** |
| 2 | 🟠 `18m` was satisfied by a different guard than it names | **VERIFIED RESOLVED** |
| 3 | 🟠 the spec's "second boundary" inverted its own causation | **VERIFIED RESOLVED** |
| 4 | 🟡 the tokenless-working STOP had no positive test | **VERIFIED RESOLVED** |

**1 — resolved, and the class is closed, not just the site.** `stamp-publish-token.mjs:56-62` now branches malformed → absent → parsed. Mutation-verified: disabling `:57` kills the new stamper test and only it. I audited every read of a `parsePublishToken` result in the tree — `publish-token.mjs:112` (comparator loop), `:118/:123/:124` (post-malformed, so `null`-only), `:333-334` (`bumpToken`), `stamp-publish-token.mjs:56-60` and `:141/:144/:148` (`main()`). All are three-way. No importer of the module exists beyond the stamper and the two test files, and `check-onbox-register.mjs` imports it not at all, so there is no second copy or second caller. `main()` is unaffected: `readFileSync(abs, 'utf8')` guarantees a string, so a malformed value there always means *a real token this reader rejects*, and it still routes to `bumpToken`, which still throws the same message. Seed-then-stamp exercised end-to-end on this head: `SEEDED … -> 1`, then `stamped … -> 2`. No working path moved.

**2 — resolved.** In the sweep, guard `:243` is now killed by `18m` **and only** `18m`; guard `:170` is killed by `4`, `5`, `18m2` — `18m` is no longer among its killers, which is the exact separation that was missing. `"kept the published page's nonce"` occurs once in the file, so the assertion is genuinely unique to `:243`.

**3 — resolved, and I re-measured it rather than reading it.** Fresh repo, `main` seeded at `1/base00`, a rival publish takes `main` to `2/mainn1`, lane branched at base. Query is the exact argv `nonceInHistory` builds:

```
A  before rebase                      -> 0 commits selected   []
B  after rebase, BEFORE any stamp     -> 1 commit             [86aac3e main: rival publish 2/mainn1]
C  after the lane stamps 3/lane01     -> 2 commits            [8f2ba2e lane stamp, 86aac3e main add]
D  control: unrebased lane stamps      -> 0 commits            []
```

All three numbers in the spec confirmed exactly. Arm **D** is the one the prose needed and did not have: a lane that stamps *without* rebasing still selects 0, so the rebase — not the stamp — is what satisfies `baselineInMine`. The rewritten paragraph is measurement-accurate.

**4 — resolved.** Guard `:123` is killed by `18p` and only `18p`.

---

### 🟡 Minor

All five are coverage/record, **not correctness bugs** — every guard below is present and behaves correctly on this head. None re-triggers a review round.

**M1 — `18f` cannot detect the boundary it is named for.** `scripts/tests/publish-token.test.mjs:276`

`18f` is titled *"the working-nonce lookup is only consulted when `w.n > b.n`"*. Mutating `publish-token.mjs:225` from `w.n > b.n` to `w.n >= b.n` — i.e. consulting the lookup at exactly the boundary the test names — leaves the suite **41/41 green**. The cause is one value: `18f` inherits `workingInBaseline: false` from `HEALTHY`, so the mutant consults the lookup, gets `false`, falls through, and produces the same message. Its second assertion (`!includes('could not search history')`) was written to catch precisely this and is neutered by the value it never overrides.

Proven, four arms:

| src | `18f` | result |
|---|---|---|
| pristine | pristine | green |
| **`>=` mutant** | pristine | **green — survives** |
| pristine | `workingInBaseline: null` | green (no false positive) |
| **`>=` mutant** | `workingInBaseline: null` | **red, on `18f` alone** |

One-line fix: `...L({ inBaseline: true, inMine: true, workingInBaseline: null })`. Red-direction only if it ever regressed (the mutant still STOPs, with the wrong message), which is why this is 🟡 — but it is the same *shape* as pass 1's finding 2, in the neighbouring test, and the file's own header warns about lookups that "coerce" a branch into passing.

**M2 — `--check` on a malformed token has no test, and its guard is the green-direction one.** `scripts/stamp-publish-token.mjs:148`

Disabling `:148` leaves the suite **41 + 15 green** while the CLI does this:

```
pristine :  --check malformed -> exit 1  "token is malformed — the publish token appears more than once (2 times)"
mutant   :  --check malformed -> exit 0  "malformed.html is at undefined (id undefined)."
```

Exit **0** on a malformed token, from the one command an operator runs to ask "is this file OK?". The guard is correct today; nothing pins it.

**M3 — `--check` on a tokenless file has no test either.** `scripts/stamp-publish-token.mjs:144` — disabling it also survives (`41 + 15 green`); the CLI then dies with a TypeError at `:148` rather than printing `has NO publish token (needs seeding)`. Still exit 1, so red-direction. One test doing `--check` against a tokenless file and a two-token file kills M2 and M3 together.

**M4 — `seedToken`'s minted-nonce charset check is unpinned, while its `bumpToken` twin is pinned.** `scripts/stamp-publish-token.mjs:73`

Disabling it leaves the suite **41 + 15 green**, and `seedToken('<h1>T</h1>', () => 'ab')` then writes:

```
data-publish-id="ab"
parsePublishToken(...) -> { malformed: 'the nonce "ab" must be 6-64 chars of [A-Za-z0-9_-] …' }
```

A file the stamper's own reader rejects — and per the comment at `:67-72`, unrecoverable, because re-stamping needs a parseable token. The identical guard in `bumpToken` (`publish-token.mjs:350`) **is** pinned, by `18l`. That comment says in as many words that the seed path is "the one least likely to be exercised again"; the code was mirrored there and the test was not. Cheapest possible fix: `assert.throws(() => seedToken('<h1>T</h1>', () => 'ab'), /6-64 chars/)`.

**M5 — the PR body still carries the claim the fold retracted, and pre-fold numbers.** Body only, no file change.

The `### F8` section still reads: *"It is what lets a correctly rebased lane clear the staleness STOP with one stamp, which is why the STOP is self-clearing rather than a dead end."* That is the exact sentence finding 3 refuted and the spec now corrects. The Test-plan block still says "39 unit (2 new) … 14 stamper" against today's 41/15, and its mutant table lists 2 of the now-5 verified mutants. The fold's own reply comment has the right numbers, so a reader who scrolls is fine — but the body is what the issue links to. `gh pr edit 2744 --body-file …`.

---

### Also noted (no action asked)

- **A third known-equivalent mutant now exists** and is worth adding to the header list at `publish-token.mjs:27-40`: `:256` `w.n === b.n` → `w.n <= b.n` survives, and correctly so — `:251` returns whenever `w.n < b.n`, so the two forms cannot be distinguished at that line. Flagging it so the next sweep does not chase it as a hole, the same service the existing `A6`/`B1` entries perform.
- **Spec `:465`** says the failure mode has been *"recorded four times"*; commit `2b022364`'s own message says *"five times"*. Reconcilable (four prior + this one), and the count is rhetorical rather than load-bearing — noted only because a hand-maintained count in this file's prose is a shape this module has been bitten by.

### Not a PR finding — the intermittent `check-onbox-register.test.mjs:2811`

Cheaply determinable, and the standing hypothesis is **refuted for this test**: it does not reach live git. The only `git` invocations in `check-onbox-register.mjs` are `resolveBaselineText`'s three (`:1168` fetch, `:1172` rev-parse, `:1177` show), and `ONBOX_TEST_BASELINE_FILE` — which this test sets via `withHermeticBaseline` — takes the branch at `:1377-1416` that replaces the baseline with a `readFileSync` and never calls `resolveBaselineText` at all. What the test *does* read live is the real on-disk `onbox-acceptance-register.md` and `…-live-view.html` in the worktree, from which it derives `lastB`, `newRowNumber` and `badDischargeName`. That is a shared-mutable-file dependency, not a git-state one — a different latent bug from the one the design review warned about. It passed here in the full 1669-test run.

### Verdict

**Nothing blocks merge.** All four pass-1 findings are resolved and each was re-derived by mutation rather than read; the fix for finding 1 closed the class, not just the site, and did not move any working path — `main()`, `--check`, seed and bump all behave exactly as before. The full guard-deletion sweep over `publish-token.mjs` now has **zero survivors out of 38**, up from 22/24 at pass 1.

Everything found this round is 🟡: five unpinned-guard / stale-record items, no correctness defect in shipping behaviour, and per the re-review trigger none of them re-triggers a round. M1 is the one I would not leave — it is a test whose name asserts an invariant it demonstrably cannot check, in the same file and the same shape as the finding this PR was opened to fix.

Issue #2771, opened 2026-08-28
[agent instructions][claude][task] Restore SPK device pin after cuda->cpu self-demotion


dudarenok-maker:
## What

Fix `SpeakerEngine` (SPK, `server/tts-sidecar/main.py`) so a cuda->cpu
self-demotion no longer permanently corrupts the `SPK_DEVICE` pin. Closes
#2750.

## Background you need

`SpeakerEngine.__init__` (`main.py:8037-8048`) captures the operator's
requested device:

```python
self.device = os.environ.get("SPK_DEVICE", "cpu")
self._requested_device = self.device  # preserved before any cpu-demotion
```

`ensure_loaded` (`main.py:8069-8104`) self-demotes at runtime: if a cuda load
raises a non-poison cuDNN/driver error, it sets `self.device = "cpu"` and
reloads on cpu (`main.py:8094-8102`). This is a live-error recovery path, not
an admission overwrite -- `self.device` itself becomes `"cpu"`, and nothing
in the file ever restores it.

`_drop_model_locked` (`main.py:8143-8156`) is where every sibling engine
restores its pin on teardown -- compare `WhisperEngine._drop_model_locked`
(`main.py:7910-7928`, the #2642 fix) which does:

```python
self._model = None
# Restore the env pref (mirrors CoquiEngine/#1730 gap-3, KokoroEngine/#2631
# S4): ...
self._device = self._requested_device
return True
```

SPK's `_drop_model_locked` has no such line -- it only nulls `self._model`.

But adding that line alone is not sufficient, because `maybe_free_idle`
(`main.py:8184-8216`) never reaches `_drop_model_locked` once demoted. Its
guard, present at BOTH the lock-free fast-out (`main.py:8199`) and the
re-validated-under-lock check (`main.py:8206`):

```python
if _parse_device(self.device)[0] != "cuda" or self._model is None:
    return False
```

Compare `WhisperEngine.maybe_free_idle` (`main.py:7956-7984`), which has NO
device-family gate at all -- it tears down regardless of device. SPK's extra
`!= "cuda"` gate is intentional and must be PRESERVED for the ordinary case
(SPK is CPU-only by default, and the docstring at `main.py:8184-8198`
explains why cpu shouldn't pay reload churn) -- but once self-demoted, that
same gate makes `self.device` permanently `"cpu"`, so this check is
permanently `False` and teardown (and the restore inside it) can never run.

[...]

PR #2784, opened 2026-08-28
fix(sidecar): restore SPK device pin after cuda->cpu self-demotion

### LINKED ISSUE #2750: SpeakerEngine never restores its device pin after a cuda->cpu self-demotion
Found in passing during the PR review gate pass on #2743 (fix/sidecar-2642-whisper-device-restore), which restored `WhisperEngine._device` on teardown mirroring CoquiEngine (#1730 gap-3) and KokoroEngine (#2631 S4). Issue #2642's audit enumerated Coqui/Kokoro/Whisper/Qwen for this class of bug and never looked at SpeakerEngine (SPK) — but SPK actually has the worse case.

## What's different about SPK

Unlike the other three engines, SPK doesn't get its device pin overwritten via an *admission* path — it self-demotes at runtime (`server/tts-sidecar/main.py:7983`, `:7985`, `:7995`, cuda -> cpu on a non-poison cuDNN/driver error). `__init__` (`:7942`) captures the original request with a comment — `# preserved before any cpu-demotion` — for a restore that exists nowhere in the file. `_drop_model_locked` (`:8037-8050`) never restores `self.device = self._requested_device`.

## Why this needs a design decision, not a one-line mirror

Naively copying the Whisper/Coqui/Kokoro fix (add the restore line to `_drop_model_locked`) is not sufficient on its own: `maybe_free_idle`'s fast-out at `:8093` checks `self.device != "cuda"` before it will even call `_drop_model_locked`. Once SPK self-demotes to cpu, that check is permanently false, so `_drop_model_locked` (and any restore line inside it) is never reached — teardown itself is gated on the very field the bug corrupts. So the fix has more than one defensible shape:

1. Add the restore to `_drop_model_locked` **and** change the idle-eviction gate so a cpu-demoted-but-loaded engine can still be torn down.
2. Restore the pin somewhere else entirely (e.g. at the point of demotion, once the corresponding admission/session ends), independent of `_drop_model_locked` ever running.
3. Some other decision about whether SPK's self-demotion should even be treated the same way as the other engines' CPU-admission-driven pin loss, given it's triggered by a live error mid-session rather than at load time.

## Impact if unaddressed

Live repro shape: `SPK_DEVICE=cuda:0`, `/embed` admits cuda:0, a non-poison cuDNN/driver error demotes to cpu — after which the engine is never torn down via `maybe_free_idle`, and `/embed`'s `on_gpu` gate (`:11429`) keeps reading the demoted field, so the operator's `cuda:0` pin is dropped for the remainder of the process lifetime (not just until the next load, unlike the other three engines' latent variant).

## References

- `server/tts-sidecar/main.py:7942` (capture comment), `:7983,:7985,:7995` (self-demotion sites), `:8037-8050` (`_drop_model_locked`, no restore), `:8093` (`maybe_free_idle`'s `!= "cuda"` gate), `:11429` (`/embed`'s `on_gpu` gate)
- `docs/features/264-vram-aware-gpu-placement.md:526-531` — states the teardown contract this violates
- `server/src/tts/transcribe-client.ts:69-76` — names SpeakerEngine as the self-demoting engine
- #2642 — the sibling issue/PR (#2743) that fixed Whisper but didn't audit SPK

**Decision owed:** which of the three shapes above (or another) is the right fix for SPK's teardown/restore behavior.

dudarenok-maker:
## Summary

Fixes `SpeakerEngine` (SPK, `server/tts-sidecar/main.py`) so a cuda->cpu
self-demotion no longer permanently corrupts the `SPK_DEVICE` pin:
`_drop_model_locked` now restores `self.device` from `self._requested_device`
on drop (mirroring `WhisperEngine`'s #2642 fix), and `maybe_free_idle`'s
device-family gate is widened at both call sites (lock-free fast-out and the
re-validated check under `_infer_lock`) so a self-demoted engine is still
eligible for teardown even though its current device reads cpu. Closes #2750.

## Test plan

- [x] `test_maybe_free_idle_noop_on_cpu` — unchanged, still passes (ordinary
      cpu-never-demoted no-op case untouched)
- [x] `test_maybe_free_idle_frees_on_cuda_after_ttl`,
      `test_maybe_free_idle_frees_on_indexed_cuda`,
      `test_maybe_free_idle_keeps_recent_model` — unchanged, still pass
- [x] Two new regression tests added and passing:
      `test_spk_drop_model_locked_restores_device_pin_after_cpu_demotion`,
      `test_maybe_free_idle_frees_and_restores_pin_on_demoted_engine`
- [x] Mutation-tested both fixes (verified by this verify pass from the
      #2771 AGENT DONE receipt, which showed real failure text rather than
      just asserting it): reverting the `maybe_free_idle` gate widening made
      the new idle-eviction test fail with `assert False is True`; reverting
      the `_drop_model_locked` restore line made the new drop test fail with
      `AssertionError: assert 'cpu' == 'cuda'`. Both fixes restored and the
      full `test_speaker_embed.py` + `test_devices.py` suite passed: 53
      passed, 0 failed.
- [x] `npm run test:sidecar` in this worktree — exits 0. No `.venv` is
      bootstrapped under `server/tts-sidecar` in this worktree, so the gate
      takes its documented local SKIP path (`scripts/run-sidecar-tests.mjs`:
      "no venv still means SKIP + exit 0, so a fresh clone doesn't fail the
      gate" — CI enforces real execution separately via `--require-venv`).

## On-box acceptance

On-box acceptance is **not owed** for this change. The fix addresses pure sidecar Python logic — device-string bookkeeping and the idle-eviction gate's device-family comparison — with no dependency on real hardware behavior. Full coverage is provided by the automated pytest suite (existing tests unchanged, two new regression tests covering both halves of the fix): `test_spk_drop_model_locked_restores_device_pin_after_cpu_demotion` confirms the device pin is restored on drop after a CPU demotion, and `test_maybe_free_idle_frees_and_restores_pin_on_demoted_engine` confirms the idle-eviction gate recognises a demoted engine as eligible for teardown and both fixes work together to free the model and restore the pin.

## Trade-off accepted (#2750, shape 1)

This PR re-arms `/embed`'s `on_gpu` gate by restoring SPK's device pin after cuda->cpu self-demotion. Issue #2750 proposed three shapes; shape 1 (always-restore, this PR) was chosen. Its consequence: a persistently cuda-failing SPK engine now re-attempts cuda capacity reservation on every reload cycle and can return a 503 under GPU contention, where before it would have silently remained on cpu. This trade-off was accepted to fix the permanent pin-corruption bug (#2750) — the previous sticky-cpu degradation produced corrupted state that persisted across sessions; the occasional 503 under contention is transient and retryable (see `withCapacityRetry` in `server/src/tts/embed-client.ts:83`).


dudarenok-maker:
## PR review — pass 1 (head 75ed8c40576d3bc33dbc3a8feb9d0c5549054457, depth medium)

Scope: `server/tts-sidecar/main.py` (`SpeakerEngine` and every caller that reads `SPK.device` — `_idle_evict_steps:5117`, `/embed:11570`, `_engine_actual_card:9862`, `_spk_idle_watchdog:8422`, `/health:10157`) and `server/tts-sidecar/tests/test_speaker_embed.py` · verified: ran `tests/test_speaker_embed.py` at this head with the primary checkout's sidecar venv (15 passed), then `test_asr_spk_idle_evict.py` + `test_devices.py` + `test_transcribe_embed_admission.py` + `test_capacity.py` + `test_placement.py` (111 passed), then re-ran `test_speaker_embed.py` with `SPK_DEVICE=cuda` exported — and the same run against `main` as a control — before adversarially probing the widened gate's predicate, the callers the restored pin re-arms, and the contracts the change made false.

### 🟠 Significant — the new gate makes an existing test fail under `SPK_DEVICE=cuda`, the one configuration this feature exists for

`server/tts-sidecar/tests/test_speaker_embed.py:164-170` constructs `main.SpeakerEngine()`, which reads `SPK_DEVICE` from the **ambient environment** (`main.py:8047-8048`), and then overwrites only `eng.device`. Before this PR the gate read `self.device` alone, so the ambient pin was irrelevant. After it, `_requested_device` is live input to the gate, and the test's outcome flips with the operator's own env.

Repro (run in `server/tts-sidecar`, primary venv):

```
SPK_DEVICE=cuda python -m pytest tests/test_speaker_embed.py -q
```

- at `75ed8c40`: `FAILED tests/test_speaker_embed.py::test_maybe_free_idle_noop_on_cpu — assert True is False`
- at `main` (control, same command, same interpreter): 13 passed

Trace: `_requested_device = "cuda"` (env), test sets `device = "cpu"` → `demoted` (`main.py:8206`) is True → the family check at `:8207` is short-circuited → the model is dropped and `maybe_free_idle` returns `True`, against the test's `is False`.

This matters twice over. (a) The PR body's test plan asserts "`test_maybe_free_idle_noop_on_cpu` — unchanged, still passes"; that is true only when `SPK_DEVICE` is unset, which is CI but is not a box running the srv-47 cuda path. Anyone with `SPK_DEVICE=cuda` in `server/.env` exported into their shell gets a red suite from a green branch. (b) The assertion the test encodes — "cpu never evicts" — is exactly the contract this PR intentionally narrows, and it was left un-rewritten rather than re-scoped. The surviving contract is "an engine on cpu **by request** never evicts", which the test should pin explicitly (`eng._requested_device = "cpu"`, or `monkeypatch.delenv("SPK_DEVICE", raising=False)` before construction) instead of inheriting from the environment. A companion case asserting a `_requested_device="cuda"` / `device="cpu"` engine **does** evict already exists as the new `:150` test, so the pair would then cover both sides deliberately.

### 🟡 Minor

**1. [cleanup] `demoted` is a raw string inequality, not a fell-back-off-the-family test — it is True on the healthy GPU path too.** `main.py:8206` and `:8216`. With `SPK_DEVICE=cuda` (unindexed) and capacity admission on, `/embed:11576-11581` threads the admitted concrete key into `ensure_loaded(adm["device"])`, which sets `self.device = "cuda:0"` (`main.py:8079-8080`; `admit` returns `{"device": key}` with a concrete `cuda:N`, `:4610`). `"cuda:0" != "cuda"` → `demoted` is True for the entire life of a perfectly healthy GPU-resident load. It is benign *today* only because the check it short-circuits (`_parse_device(self.device)[0] != "cuda"`) would have admitted that engine anyway — but the variable name, and the comment/docstrings introduced with it, both say "self-demoted", so the next edit that trusts the flag to mean "fell back to CPU" will be wrong. A family-aware predicate says what is meant and is False on the admitted path: `_parse_device(self._requested_device)[0] == "cuda" and _parse_device(self.device)[0] != "cuda"`.

**2. [chore — stale contracts the change made false] four surfaces still document the behaviour this PR removed.** All say the SPK idle evict is a no-op on cpu; after this change a `SPK_DEVICE=cuda` engine that has demoted is evicted on cpu, reclaiming zero VRAM and paying the reload:
- `main.py:8190-8196` — `maybe_free_idle`'s own docstring: "a NO-OP on cpu, where the ~1 s reload churn isn't worth freeing ~80–200 MB of host RAM", plus "Cheap, lock-free fast-outs first (cuda-only guard, then nothing loaded …)", which also no longer describes the order the code now uses (`_model is None` moved first at `:8204`);
- `main.py:8382-8384` — "before the watchdog frees the ECAPA model (cuda path only)";
- `main.py:8423-8425` — `_spk_idle_watchdog`'s docstring: "(a no-op on cpu)";
- `server/src/config/registry.ts:842` — the operator-facing Settings help for `sidecar.spkIdleTtl`: "Reclaims VRAM only on SPK_DEVICE=cuda (a no-op on cpu). Values below 5 fall back to the default (120) **to avoid reload thrash**" — the sentence names the exact cost the new path now pays for zero VRAM. `server/.env.example:694` carries the same string and is derived from the registry, so it needs regenerating with it (`config:sync`).

**3. [chore — house gate] the release-notes pair is missing and no not-applicable is declared.** Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched. The direct sibling — PR #2743, the `WhisperEngine` #2642 fix this PR mirrors — landed both in commit `05e0649b` ("docs(release-notes): pair PR #2743's device-restore fix"), a three-line technical entry plus one brand-voice user-facing line under the in-progress version. `RELEASE_NOTES.md` has a live `# Castwright 1.15.0` section to append to. CLAUDE.md before-shipping step 5 allows skipping only with an explicit statement; the PR body's test plan does not make one. Step 3 (on-box acceptance) is likewise neither claimed nor waived — #2743 shipped without a register row, so precedent supports "not owed", but that should be said rather than omitted.

**4. [cleanup] the new eviction test diverges from every sibling in its own file by not stubbing torch.** `tests/test_speaker_embed.py:150` declares `monkeypatch: pytest.MonkeyPatch` and never uses it. `test_maybe_free_idle_frees_on_cuda_after_ttl:173` and `test_maybe_free_idle_frees_on_indexed_cuda:183` both call `_stub_torch_cuda(...)` precisely because the success path runs `_reclaim_after_drop` → `_reclaim_host_and_vram` (`main.py:1296-1312`), which imports the real torch and calls `torch.cuda.is_available()` / `empty_cache()`. The new test reaches that same path with the real module, so it does a real `gc.collect()` and touches the real CUDA runtime in the test process — the thing CLAUDE.md's "no torch test suites during live generation" warning is about. Either use the `monkeypatch` parameter for `_stub_torch_cuda` as the siblings do, or drop the unused parameter.

**5. [cleanup] neither new test exercises the real demotion, so both would stay green if the bug came back by a different route.** `:133` and `:150` hand-assign `_requested_device` / `device` to *simulate* the post-demotion state. This file already has two tests that produce that state for real — `test_degrade_when_cuda_unavailable:89` and `test_demote_on_non_poison_load_failure:99`, the latter being the issue's own live repro shape. Concrete case the current pair cannot catch: if a later change made `ensure_loaded` record the admitted card into `_requested_device` as well (the shape `_engine_actual_card`'s #2647 comment at `main.py:9874-9895` discusses for the other engines), `demoted` becomes permanently False, teardown is unreachable again, and the bug is fully restored — with both new tests passing, because they never run `ensure_loaded`. Chaining `test_demote_on_non_poison_load_failure`'s engine into `maybe_free_idle` and asserting the pin came back would pin the invariant end to end.

**6. [cleanup] a real consequence of the restore is undeclared.** Restoring the pin re-arms `/embed`'s `on_gpu` gate (`main.py:11570`), which is the behaviour the issue asked for — but it also means a *persistently* failing cuda load (the demotion at `:8099-8102` fires on every reload) now re-enters `_placement.reservation(..., cpu_capable=False)` on every reload cycle: VRAM is booked on `cuda:N` for an engine that is about to run on cpu, and on a contended card `:11577` returns `_no_capacity` → 503 where the pre-fix sticky-cpu state would have embedded on CPU and succeeded. `withCapacityRetry` (`server/src/tts/embed-client.ts:83`) softens this, it does not remove it. Issue #2750's shape 3 explicitly put "whether SPK's self-demotion should be treated the same way as the other engines' admission-driven pin loss, given it's triggered by a live error mid-session" on the table; shape 1 was chosen without recording that this is the accepted trade. It belongs in the PR body and in the `maybe_free_idle` docstring, not only in the issue thread.

### ✅ What is solid

- The lock discipline is right and unchanged: the widened predicate is applied at **both** call sites (`:8207` lock-free, `:8217` re-validated under `_infer_lock`), the `_in_flight.busy` and `_last_used` re-checks under the lock are preserved in order, `_drop_model_locked` is still called while holding `_infer_lock`, and `_reclaim_after_drop` still runs after release — the acquire-drop-release-reclaim sequence `docs/features/264-vram-aware-gpu-placement.md` §3 requires.
- `_requested_device` remains write-once (`main.py:8048` is the only assignment in the file), so the restore cannot walk.
- `_engine_actual_card` is unaffected: SPK has no `_device` attribute, so it takes the `_requested_device` branch (`:9928`), which this change never writes, and the function returns `None` for an unloaded engine — so the restore cannot fabricate a `fell_back` reading.
- Checked and benign: after a restore, `_idle_evict_steps:5117`'s `_same_card(SPK.device, …)` can now list an `spk` step where it previously would not, but `maybe_free_idle` fast-outs on `_model is None` and the step is a no-op rather than a spurious eviction.
- The class is not left armed — SPK was the only engine whose `maybe_free_idle` gated teardown on the corruptible device field. `CoquiEngine.maybe_free_idle:2273` and `WhisperEngine.maybe_free_idle:7956` gate on `_tts`/`_model`, in-flight and TTL only.
- The two new tests are genuinely red before the fix, for the reasons claimed (`:147` on the missing restore line, `:159` on the un-widened gate), and the PR body's mutation evidence quotes real failure text rather than asserting it.

### Verdict

**Changes requested.** One 🟠: the widened gate makes `test_maybe_free_idle_noop_on_cpu` fail under `SPK_DEVICE=cuda`, reproduced above against this head with a passing control run on `main` — the test needs to pin `_requested_device` rather than inherit it from the environment. Separately, two of the 🟡s are house gates that block merge under CLAUDE.md regardless of the re-review split: the missing release-notes pair (with an explicit not-applicable for on-box acceptance), and the four stale "no-op on cpu" contract statements — one of which is operator-facing Settings help with a derived `.env.example` line to regenerate alongside it. The core fix itself — restoring `self.device` in `_drop_model_locked` and widening the gate so teardown can reach it — is the right shape and correctly locked.

dudarenok-maker:
## PR review — pass 2 (head 0de1163f314bdcffcbfe423685dcf94e20c5c80e, depth medium)

**Scope/verified:** all 8 commits vs `origin/main` merge-base `564903f4` (6 files, +142/-15). Read the current `SpeakerEngine` body (`server/tts-sidecar/main.py:8030-8233`), `_engine_actual_card`/`_is_resident`/`_idle_evict_steps`/`/embed` (`:5117`, `:5007`, `:9869`, `:11577-11592`), the whole of `tests/test_speaker_embed.py`, and the two release-notes files. Ran the full sidecar suite against this worktree — **1094 passed, 2 skipped, 10 deselected** (5:51) — plus `test_speaker_embed.py` a second time with `SPK_DEVICE=cuda` ambient (**17 passed**) to prove finding 1's fix is genuinely env-independent, and `npm run config:check` (**`.env.example` in sync ✓**). Mutation-probed finding 2's new test by re-running its exact state against the pre-fix predicate (script run from the OS temp dir; no tracked file touched). **No 🔴/🟠 this pass** — the three new findings are all 🟡, so this does not re-trigger the loop.

1. **🟠 env-dependent `test_maybe_free_idle_noop_on_cpu` — VERIFIED RESOLVED.** `tests/test_speaker_embed.py:167` now pins `eng._requested_device = "cpu"`. Proved rather than assumed: re-ran the file with `SPK_DEVICE=cuda` exported and got 17 passed, where the unpinned version would have taken the `demoted` branch and evicted.
2. **🟡 family-aware `demoted` — code CHANGE VERIFIED, its test is not (see N3).** `main.py:8212` / `:8222` now compare families at both gates. The change itself is correct and mirrors the rest of the file's `_parse_device` discipline. What did not land is a guard that can detect its removal — new finding N3.
3. **🟡 stale "no-op on cpu" prose — VERIFIED RESOLVED.** All four sites updated (`maybe_free_idle` docstring `:8190-8192`, its fast-out paragraph `:8194-8196`, `_SPK_IDLE_TTL_DEFAULT` comment `:8388-8391`, `_spk_idle_watchdog` docstring `:8430-8433`) plus `registry.ts:842` and `server/.env.example:694`; `config:check` green, so the pair really is synced. Swept the whole tree for survivors — the only remaining "no-op on cpu" / "cuda path only" hits are in `docs/superpowers/plans/2026-06-22-srv-47-spk-cuda.md` and `docs/superpowers/specs/2026-06-22-srv-47-spk-cuda-design.md`, which are historical design records quoting the code as it stood at plan time. Not stale prose in the CLAUDE.md sense; correctly left alone.
4. **🟡 release-notes pair + on-box stance — PARTIALLY RESOLVED.** Both files got an entry and the PR body now declares an on-box stance ("not owed", with a reason) — that half is fine and consistent with the #2642/PR #2743 sibling, which added no register row either. But the entries themselves are wrong in two separate ways: **N1** (wrong version section) and **N2** (wrong mechanism). Details below.
5. **🟡 unstubbed torch / unused `monkeypatch` — VERIFIED RESOLVED.** `tests/test_speaker_embed.py:154` calls `_stub_torch_cuda(monkeypatch, available=True)`; the parameter is now load-bearing.
6. **🟡 no test of the real demotion path — VERIFIED RESOLVED.** `test_real_cuda_demotion_through_maybe_free_idle` (`:245-292`) drives a real `ensure_loaded()` cuda failure → cpu fallback → `maybe_free_idle` → `_drop_model_locked`, asserting `calls == ["cuda", "cpu"]` on the way through, so it is anchored to the production demotion site rather than to hand-assigned fields. This one is a real regression test: reverting either half of `75ed8c40` reddens it (the gate revert on `freed is True`, the restore revert on `device == "cuda"`). Minor bookkeeping drift: the PR body's Test plan still reads "Two new regression tests added and passing" — there are now four.
7. **🟡 undeclared capacity/503 trade-off — VERIFIED RESOLVED, and the cited evidence checks out.** `maybe_free_idle`'s docstring `:8204-8208` plus the PR body's "Trade-off accepted" section. I verified the load-bearing claim rather than taking it: `withCapacityRetry` is real at `server/src/tts/embed-client.ts:84`, and — better than the PR body claims — its `spkRunsOnGpu()` guard (`embed-client.ts:55-57`) resolves `qa.speaker.device` from Node config, *not* from the sidecar's live `SPK.device`, so the retry wrapper stays armed regardless of demotion state. `_no_capacity` is a genuine 503 (`main.py:4984`) and `SEG_CAPACITY_ADMISSION` defaults on (`:4981`), so the trade-off is real and is now stated where a reader will hit it.

### 🟡 New — Minor (cleanup / chore)

**N1 — the `RELEASE_NOTES.md` entry landed in the shipped `# Castwright 1.7.0` section, not the in-progress `1.15.0` one.** `RELEASE_NOTES.md:418` sits between `# Castwright 1.7.0` (`:405`) and `# Castwright 1.6.0` (`:420`). The in-progress section is `# Castwright 1.15.0` at `:1` — `package.json` is on `1.14.0` and `v1.14.0` is the newest tag, so 1.15.0 is the section CLAUDE.md's before-shipping step 5 means by "the in-progress version section at the top". Concrete failure: at the 1.15.0 cut the release body is assembled from the 1.15.0 section, so this fix ships with **no** user-facing note — precisely what step 5 exists to prevent — while the already-published 1.7.0 notes gain a bullet for work that shipped eight releases later. Context, not a request to widen scope: PR #2743's sibling entry at `:416` was misfiled the same way by `05e0649b`, so this is the second instance of one class.

**N2 — both release-notes entries mis-describe the mechanism they document.** Three separate errors:
- `docs/release-notes-next.md:416` — "a self-demotion **triggered by `maybe_free_idle`** when the idle time exceeded its threshold". `maybe_free_idle` never demotes anything; it only evicts. The demotion is in `ensure_loaded` (`main.py:8084-8091` for absent CUDA, `:8099-8102` for a non-poison cuda load failure).
- Same line — "so `maybe_free_idle` could never **reclaim the model to free VRAM**". While demoted the model is resident on **CPU** and holds no VRAM at all, so there is nothing to reclaim. The actual consequence, and the one #2750 is about, is that the pin stayed `cpu`, so `/embed`'s `on_gpu` gate (`main.py:11577`) read false and *every subsequent load ran on CPU for the life of the process*.
- `RELEASE_NOTES.md:418` compounds both — "**Speech synthesis** now properly frees GPU memory when stepping down to your processor … after stepping down from your GPU **to conserve memory**". SPK is the ECAPA speaker-embedding engine behind voice-identity QA, not a synthesis engine (`main.py:8030-8034`: "NOT in the synth `ENGINES` map"), and the step-down is a *load failure* fallback, not memory-pressure management. A user reading this will look for the behaviour in speech generation and not find it.

Concrete failure scenario: this repo briefs cold agents off these notes. An agent told the demotion lives in `maybe_free_idle` will edit the wrong function, or will read `main.py`, fail to find the described mechanism, and conclude the note documents deleted code.

**N3 — the test shipped alongside fix 2 cannot fail; the code change it claims to guard is unguarded.** `server/tts-sidecar/tests/test_speaker_embed.py:198-218` (`test_maybe_free_idle_not_demoted_on_admitted_indexed_load`). Mutation-proved, not inferred — I ran the test's exact state (`_requested_device='cuda'`, `device='cuda:0'`, `_last_used` 10,000 s ago) against both the shipped family-aware predicate and the verbatim pre-`8706e08d` string-exact one from `75ed8c40`:

```
SHIPPED  (family-aware): maybe_free_idle -> True; _model is None -> True; device -> 'cuda'  => assertions PASS
REVERTED (string-exact): maybe_free_idle -> True; _model is None -> True; device -> 'cuda'  => assertions PASS
```

Both take the same branch, because on a `cuda`-family `device` the second conjunct (`_parse_device(self.device)[0] != "cuda"`) already admits and `demoted` is never read. Reverting `8706e08d` leaves this test green. The docstring itself records the retraction — it opens claiming the string check "wrongly reported demoted=True", then concedes "the demoted flag wouldn't actually trigger eviction anyway" — so the file states, in prose, that it is not a regression test.

Worse for the framing: enumerating both predicates over the `(requested, device)` space, the **only** state where they diverge is `requested='auto', device='cpu'` (old: proceeds to evict a plain-CPU engine; new: bails). That state is unreachable — `ensure_loaded` only ever writes `device='cpu'` starting from a `cuda`-family device (`main.py:8084`, `:8099`), and `/embed`'s `on_gpu` gate (`:11577`) never admits a placement for an `auto` pin — so the shipped `demoted` local is behaviour-neutral against every reachable input. Either pin the divergent case (or assert the predicate directly) so the test can go red, or drop it and keep `8706e08d` as the readability change it actually is. Related, same two sites: `main.py:8212-8213` and `:8222-8223` reduce exactly to `if req_fam != "cuda" and dev_fam != "cuda": return False`, so the `demoted` local and its extra `_parse_device` call are dead weight at both gates.

### ✅ Verified clean this pass (adversarial checks that found nothing)

- **No sibling gap left armed.** `WhisperEngine.maybe_free_idle` (`main.py:7955-7982`) has no device gate at all, so #2642's restore was always reachable — the SPK-specific cuda-only gate was the anomaly, and it is the thing this PR widened. Nothing to mirror back.
- **The restore does not corrupt `/health` or residency accounting.** `_engine_actual_card` early-returns `None` on `model is None` (`:9879`), so a restored pin on an unloaded engine cannot manufacture a `fell_back` badge; `_is_resident` (`:5007`) goes through the same probe, so no phantom VRAM booking. `spk_device` (`:10164`) now reports the pin rather than the last actual placement after an unload, but it has zero consumers outside the payload and a key-presence test, and `spk_loaded` sits beside it.
- **`_idle_evict_steps` (`:5117`) is unaffected.** A demoted, still-resident engine reads `device == "cpu"` and is still skipped by `_same_card`; after eviction the step may be appended for a cuda card, where `maybe_free_idle` immediately fast-outs on `_model is None`.
- **Plan 264's stated invariants hold** (`docs/features/264-vram-aware-gpu-placement.md:503-520`): re-validation under `_infer_lock` preserved at `:8219-8228`, `_drop_model_locked` still called inside the lock with `_reclaim_after_drop` outside it.
- **No lock/ordering regression.** `_drop_model_locked` writes `self.device` under `_infer_lock` while `ensure_loaded` writes it under `_load_lock`, but the two cannot interleave: `maybe_free_idle` fast-outs on `_model is None`, which is exactly the window `ensure_loaded` occupies. Mirrors the Whisper precedent (`:7910-7928`) line for line.

### Verdict

**Fixes 1, 3, 5, 6, 7 verified resolved.** Fix 2's code change is correct but ships an inert guard (**N3**). Fix 4 delivered the files but both entries are defective — wrong section (**N1**) and wrong mechanism (**N2**).

No blocking or significant correctness findings: the shipped runtime behaviour is right, the suite is green (1094 passed), and the house gates (`Closes #2750` outside a code span, release-notes pair present, on-box stance declared, `config:check` in sync, paired regression tests that genuinely redden) are all satisfied except as noted. **Three 🟡 cleanup findings to fix in this round per the report-fix-record rule; none re-triggers the re-review loop.**

Issue #1975, opened 2026-07-31
sidecar: three in-lock cold Qwen loads survive plan 273 (synthesize_batch, design_voice audition)


dudarenok-maker:
## Summary

Plan 273 (T7/T8, PR #1968) fixed the "a cold model load runs while holding
the engine's forward lock" defect shape (#1925) for `WhisperEngine.transcribe`
and `QwenEngine._guarded_base_synth` (the single-utterance `synthesize` path).
Three more sites in `QwenEngine` have the identical shape and were not
touched by T8, and are not named anywhere in
`docs/features/archive/273-sidecar-lock-event-loop.md`.

Each of the three already has a pre-ensure immediately above it, so the
in-lock call is normally a warm no-op — exactly the shape T7/T8 fixed
elsewhere. But when an `/unload` wins the narrow window between that
pre-ensure and the lock acquire, a full cold weights pull runs while
`_synth_lock` is held, blocking `unload()` (Stop) for the whole pull.

## Sites

| Site | Code | Reachable via |
|---|---|---|
| `main.py:5336-5337` | `with self._synth_lock:` → `_ensure_base_loaded()` (design_voice audition forward) | `unload()` nulls `_base` (`main.py:5033`) |
| `main.py:6028-6030` | `with self._synth_lock:` → `_ensure_base17_loaded()` (`synthesize_batch`, 1.7B path) | `unload_base17()` via `POST /unload {model:'1.7b'}` (`main.py:8687`) |
| `main.py:6096-6099` | `with self._synth_lock:` → `_ensure_base_loaded()` (`synthesize_batch`, 0.6B path, default) | `unload()` nulls `_base` |

(Line numbers verified against `main.py` at PR #1968's HEAD, `89799cdf`; check
for drift before implementing.)

## Why this matters

`synthesize_batch` is the **main** chapter-render path — Qwen chapter
sentences batch through it (see the comment at `main.py:6061-6065`: "THIS is
the path that matters for a book"). The `synthesize` path plan 273 T8 fixed
is the fallback/single-utterance path (title beat, audition, clone).
`synthesize_batch` deliberately does **not** route through
`_guarded_base_synth` — see the comment at `main.py:5965-5970` ("the silent
degenerate-load OUTPUT guard... is INTENTIONALLY NOT wired here yet — a batch
retry needs a full-batch re-run... deferred to a follow-up; the single-synth
path is guarded"). So plan 273 fixed the **colder** Qwen path (single
[...]

PR #2790, opened 2026-08-28
fix(sidecar): widen design_voice base17 evict guard to in-flight loads

### LINKED ISSUE #2752: bug(sidecar): design_voice()'s 1.7B-Base evict guard blind to an in-flight base17 load
## What happened

PR #2745's re-review (pass 2) found `design_voice()`'s eviction guard for the 1.7B-Base model has the identical blind spot PR #2745 fixed on the design side, in the opposite direction: at `server/tts-sidecar/main.py:6472`, the guard is `if self._base17 is not None:` — blind to the window where `_base17_in_flight` is claimed and the ~3.4 GB pull is still running (`self._base17` stays `None` for the whole load, same shape as `self._design`/`_design_in_flight` before #2745).

Proven, not inferred: a probe mirroring PR #2745's own regression tests (claim `_base17_in_flight`, leave `self._base17` as `None`, call `design_voice()`) shows `unload_base17` called 0 times while `_ensure_design_loaded` is reached — i.e. `design_voice()` proceeds to load the ~4-5GB VoiceDesign model concurrently with an in-flight 1.7B-Base load. This is the #1156 OOM shape ("VoiceDesign and 1.7B-Base co-reside on an 8GB card") in the one window the existing guard can't see.

## Why this needs a design decision, not just a mechanical guard widen

PR #2745's fix worked because `unload_design()` already had a wait-then-evict path (`_design_in_flight.busy`, bounded 150s, per #2070's "design wins" policy) — widening the caller-side guard to also catch the in-flight case was sufficient.

`unload_base17()` has **no equivalent wait mechanism at all** — it has nothing to wait on. So fixing this symmetrically needs one of:

1. **Add a wait-then-evict path to `unload_base17()`**, mirroring `unload_design()`/#2070's "design wins" policy, so `design_voice()` waits for an in-flight base17 load to finish before evicting it.
2. **Some other resolution** — e.g. refuse admission on the `design_voice()` side until the base17 load completes, rather than evicting after the fact.

Both are defensible and the choice affects real UX (which operation waits for which, and for how long) the same way #2678 itself already required deciding. Per this repo's incidental-findings protocol, this is the one class of finding allowed to stay unfixed pending a decision — filing it here rather than mechanically picking one and shipping it under PR #2745.

## Repro / verification

Probe used to confirm this (mirrors PR #2745's `test_design_contention.py` pattern) is not committed — reproduce by claiming `_base17_in_flight` (busy) with `self._base17 = None`, then calling `design_voice()`, and asserting `unload_base17` is never called before `_ensure_design_loaded()` proceeds.

Surfaced during PR #2745's re-review pass 2 (`fix(sidecar): render waits for warm VoiceDesign instead of vram-spilling`).

dudarenok-maker:
## Summary

`design_voice()`'s 1.7B-Base eviction guard only checked `self._base17 is not
None`, which stays `None` for the entire duration of an in-flight base17
load/mint — so a design arriving mid-load skipped eviction entirely and let
both the 1.7B-Base and the heavy VoiceDesign model co-reside on an 8 GB card
(the #1156 OOM shape). This widens the guard to also check
`self._base17_in_flight.busy`, and gives `unload_base17()` a bounded
wait-then-evict path (mirroring `unload_design()`'s "design wins" policy from
#2070), raising the new `Base17ContentionTimeoutError` only if the wait is
exceeded. Closes #2752.

## Test plan

- [x] `python -m pytest server/tts-sidecar/tests/test_base17_contention.py -v`
  — 4/4 pass (re-run independently during verification, not just taken on
  the implementer's word).
- [x] Mutation-verified independently during this verify pass, with actual
  pytest output captured both ways (not just the implementer's prose claim):
  - Reverting the widened guard (`self._base17 is not None or
    self._base17_in_flight.busy` → `self._base17 is not None`) fails
    `test_design_voice_widens_guard_to_base17_still_loading`:
    `AssertionError: Expected 'unload_base17' to have been called once.
    Called 0 times.`
  - Reverting `unload_base17()`'s wait-then-evict loop to an unconditional
    null fails both `test_unload_base17_waits_for_in_flight_base17_then_unloads`
    (`assert None is not None` — the model was nulled while still in flight)
    and `test_unload_base17_raises_typed_error_after_bounded_wait`
    (`Failed: DID NOT RAISE Base17ContentionTimeoutError`).
  - Both reverts undone; fix restored; tests green again (4/4).
- [x] Full `server/tts-sidecar/tests/` pytest suite (system Python, no venv
  bootstrapped in this worktree) — exit code 0, no failures, only the
  pre-existing `test_unload_base17_is_not_blocked_by_a_cold_load_during_batch_synth`
  raising `Base17ContentionTimeoutError` inside its background thread as an
  unhandled-thread-exception *warning* (not a failure — the test only asserts
  `not stop.is_alive()`, which an uncaught-and-terminated thread still
  satisfies).
- [x] `npm run test:sidecar` — SKIP (no `.venv` bootstrapped in this
  worktree; documented local behavior, exit 0).
- [x] `npm run verify` (pre-push hook) — green


dudarenok-maker:
## PR review — pass 1 (head df3eaff37c634a886dd3ae901c21896348348ba8, depth medium)

Scope: `server/tts-sidecar/main.py`, `server/tts-sidecar/tests/test_base17_contention.py` (the full diff, 281 insertions) plus every caller of `unload_base17` and of `POST /unload`, the sibling `unload_design`/`DesignContentionTimeoutError` machinery it mirrors, the `_VdKokoroArbiter`, and the Node-side `/unload` consumers.

Verified before probing: `tests/test_base17_contention.py` 4/4 green on head; the **full** `server/tts-sidecar/tests/` suite under the repo's own harness (`server/tts-sidecar/.venv`, torch present) — **1 failure**, detailed below; the same suite green on `origin/main`. Behavioural probes run against both `df3eaff3` and `origin/main` (`aee0f4ed`) worktrees, including a FastAPI `TestClient` exercise of `POST /unload`.

Depth `medium` per the ladder (single-scope `fix`).

### 🔴 Blocking — the `/unload {model:'1.7b'}` Stop route now answers 500 and does not unload, whenever a 1.7B activity is in flight

`unload_base17`'s new `wait_seconds` defaults to `0.0` (`main.py:5932`), so `deadline = time.monotonic() + 0.0` and the loop at `main.py:5966-5974` raises on its **first** iteration if `_base17_in_flight.busy`. The Stop route calls it bare and unguarded — `await asyncio.to_thread(qwen.unload_base17)` at `main.py:10629`, inside a handler whose only exception arm is for the JSON body — so the raise reaches FastAPI.

Probe, `TestClient(main.app, raise_server_exceptions=False)`, one `_base17_in_flight.claim()` held, `_base17` resident:

```
head df3eaff3 → ROUTE: status 500 | body Internal Server Error
                ROUTE: model still resident -> True
origin/main   → returned normally; _base17 is None -> True
```

`_base17_activity()` brackets the **entire** 1.7B batch loop (`main.py:7565`), so "busy" is not a load window — it is the whole chapter render. Every Stop pressed during a 1.7B render now 500s and frees nothing.

Three things make this worse than a bad status code:

1. It contradicts the invariant this file states in its own words at `main.py:6685-6691`: *"a genuinely explicit action — a full `/unload` … deliberately is NOT gated by the design-wins policy (an explicit Stop must still be able to free the model)."* The commit message asserts the opposite of what the code does: *"the pre-existing `/unload {model:'1.7b'}` Stop route and `maybe_free_idle_base17` stay immediate"* — they stay immediate only in the sense of immediately failing.
2. The failure is **silent** on the automated path. `reconcileResidentQwenTiers` (`server/src/tts/ensure-sidecar-loaded.ts:228`) issues this exact call at `:244`; `undiciFetch` does not reject on a 500 and nothing reads `res.ok`, so the helper resolves and the function returns `true`. That file's own comment block at `:191-215` documents precisely this outcome — *"The caller is therefore told the tier was evicted when it was not — so generation.ts loads the 1.7B on top of a still-resident 0.6B, the 8 GB OOM its own comment warns about."* This PR adds a new way to reach that state, and the same `true` also burns a `withCapacityRetry` attempt in `gpu/evict-idle-tts.ts`.
3. `main.py:10594-10595` still documents the route as *"Idempotent — returns `idle` whether or not the engine had a model resident."*

The PR body already contains the observation that proves this and dismisses it: *"only the pre-existing `test_unload_base17_is_not_blocked_by_a_cold_load_during_batch_synth` raising `Base17ContentionTimeoutError` inside its background thread as an unhandled-thread-exception warning (not a failure — the test only asserts `not stop.is_alive()`, which an uncaught-and-terminated thread still satisfies)."* That warning is not noise; it is the Stop path failing, and the reason the suite cannot see it. `tests/test_batch_synthesis.py:1036-1041` runs `unload_base17` on a `threading.Thread`, so a raise kills the thread and `assert not stop.is_alive()` passes on the corpse — #1975's regression test can no longer fail for its own defect (brief shapes 3 and 6).

Both halves need fixing: the bare call must keep freeing the model (or the route must catch and answer something a caller can act on — but "Stop raises" is not a Stop), and `test_batch_synthesis.py:1036` must propagate the thread's exception so this class stays catchable.

### 🔴 Blocking — the PR leaves the sidecar suite RED on any torch-equipped box, and the green run cited in the PR body could not have caught it

`server/tts-sidecar/tests/test_qwen_design_base17_exclusion.py::test_design_voice_evicts_resident_base17`, run with the repo's own `server/tts-sidecar/.venv`:

```
E   TypeError: test_design_voice_evicts_resident_base17.<locals>._fake_unload_base17()
    got an unexpected keyword argument 'wait_seconds'
main.py:6641: TypeError
```

Its monkeypatched stub (`test_qwen_design_base17_exclusion.py:61`) takes no arguments; `main.py:6641-6644` now calls with two kwargs. Same file, same command, `origin/main`: `2 passed`.

Neither of the two runs the PR body relies on can see it:

- The PR body's *"Full `server/tts-sidecar/tests/` pytest suite (system Python, no venv bootstrapped in this worktree) — exit code 0, no failures"* — the test opens with `pytest.importorskip("torch")`, and system Python here has no torch (`ModuleNotFoundError: No module named 'torch'`), so that run **skipped** exactly the test this change breaks.
- CI's green "Sidecar tests (pytest)" is not evidence either: `.github/workflows/verify.yml:355-359` states outright that *"The heavy ML stack (torch, coqui-tts, speechbrain) … is deliberately NOT installed — tests needing it use `pytest.importorskip` and report as skipped."*

So the only environment that executes this test is a fully bootstrapped box — the operator's — where `npm run test:sidecar` is now red. A "full suite, exit 0" claim made from an interpreter that skips the affected test is the acceptance-blind-to-its-own-feature shape (brief shape 5).

### 🟠 Significant — `Base17ContentionTimeoutError` has no route mapping, so a timed-out design 500s where its sibling 503s

`/qwen/design-voice` has a single `except Exception` (`main.py:10784-10801`) that answers `{"detail": "Internal error."}` with status 500. Nothing catches the new error anywhere. Its sibling `DesignContentionTimeoutError` is caught at three routes (`main.py:11000`, `:11443`, `:11781`) and mapped to **503** with `code: "design_in_flight"`, and `tests/test_design_contention_routes.py` exists *because* one of those mappings was once missing — its docstring: *"the route … did not catch `DesignContentionTimeoutError` at all — it fell through to the generic [handler]."* This PR reintroduces the identical gap for the mirror error (brief shape 10: one instance fixed, the class left armed).

Consequence is not cosmetic. `httpStatusForSidecarError` (`server/src/routes/sidecar-error-status.ts:26-52`) forwards a sidecar 5xx only when it is a 5xx it recognises as actionable, and its whole reason for existing is *"a 503 means 'no GPU capacity right now — free VRAM and retry' … while a 502 reads as 'the gateway is broken'."* A 500 lands the caller on the "gateway is broken" side — after a full 60 s wait — directly contradicting the new error's own docstring, *"the caller should retry the design shortly"* (`main.py:5350-5351`). There is also no route-level test for the new error, where the sibling has a dedicated module.

### 🟠 Significant — the 60 s wait is held inside the Kokoro arbiter, the exact placement `main.py:6965-6976` documents as wrong

`design_voice` opens `with _VD_KOKORO.design():` at `main.py:6612` and the new bounded wait runs at `:6641`, inside it. `_VdKokoroArbiter.design()` sets `_design_active = True` for its whole span (`main.py:1430-1433`) and `kokoro_synth()` waits on the condition variable **unboundedly** while it is set (`main.py:1414-1416`).

The mint side already learned this. `mint_variant` puts its `unload_design()` wait deliberately *outside* the equivalent block, and says why at `main.py:6965-6976`: *"`_VD_KOKORO.design()` sets `_design_active` for its ENTIRE span, and `kokoro_synth()` blocks while that's set — so a wait held inside that block would stall EVERY Kokoro synth for up to 150s during exactly the bulk 'Design full cast' run this eviction exists to support."*

Concretely: a design arriving while a 1.7B render holds `_base17_activity` (again, the whole batch loop) now blocks every Kokoro synth for up to 60 s, then fails with the 500 from the previous finding. Moving the wait above `with _VD_KOKORO.design():` costs nothing — the only ordering requirement is evict-before-load, and the VoiceDesign load is at `:6657`.

### 🟠 Significant — the design-side bounded wait has no test; reverting it keeps all four new tests green

`test_design_voice_widens_guard_to_base17_still_loading` patches `unload_base17` with a `MagicMock` and asserts `mocked_unload.assert_called_once()` (`tests/test_base17_contention.py:194`) — call **count** only, no args. Revert `main.py:6641-6644` to a bare `self.unload_base17()` and all four new tests still pass, while every design arriving during an in-flight base17 load flips from "wait up to 60 s" to "raise `Base17ContentionTimeoutError` after 0 s" → the 500 above. The guard half of the fix is pinned; the wait half at the call site is not. `assert_called_once_with(wait_seconds=..., poll_seconds=...)` closes it.

### 🟠 Significant — the window the PR exists to close is narrowed, not shut

Nothing re-checks `_base17_in_flight` after the wait loop exits at `main.py:5967`, nor between `unload_base17()` returning (`:6644`) and `_ensure_design_loaded()` (`:6657`). `synthesize()`'s 1.7B path claims `_base17_activity` **per call** (`main.py:7434`), so between two consecutive 1.7B sentences `busy` reads `False`. Interleaving: sentence *n* releases the claim → the design's wait loop exits in that gap and proceeds → sentence *n+1* re-claims and `_ensure_base17_loaded()` cold-loads the 3.4 GB 1.7B-Base while the 4–5 GB VoiceDesign is loading. That is the #1156 co-residency this PR is written to prevent, reached through the same shape the guard was widened for. Closing it needs the check and the VoiceDesign load to share a serialising point rather than a poll-then-proceed.

### 🟡 Minor

- **Undeclared behaviour change**: the new early `return` at `main.py:5976-5977` means `unload_base17()` no longer runs `_reclaim_host_and_vram()` when nothing is resident. It mirrors `unload_design` (`:6482-6484`) so it is plausibly deliberate, but pre-PR every `/unload {model:'1.7b'}` forced a `gc.collect()` + `empty_cache()` regardless of residency, and neither the PR body nor the diff comments mention the change.
- **Docstring cites a caller that cannot exist**: `main.py:5953-5954` justifies the `0.0` default by *"the `/unload {model:'1.7b'}` Stop route, and `maybe_free_idle_base17`'s callers"*. `maybe_free_idle_base17` never calls `unload_base17` — `main.py:6014-6016` says doing so would deadlock on the non-reentrant `_synth_lock`, which is why it nulls inline. Half the stated justification is for a path that does not exist.
- **Citation to a rule that does not exist**: `main.py:5386-5388` cites *"CLAUDE.md's 'new env var MUST be a knob' rule this avoids triggering."* CLAUDE.md has no rule by that name; the actual rule is that a config **knob** needs its registry entry, `config:sync`, Settings row and `.env.example` line. The decision (fixed constant, no env knob) is fine; the citation is not.
- **Log line overstates**: `main.py:6636` logs *"Evicting resident Qwen 1.7B-Base"* on the new branch where nothing is resident and the call is a wait, not an eviction.
- **House gates missing, with no stated N/A**: no `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair (checklist step 5) — this is a user-visible VRAM/OOM fix, so "no shippable delta" would have to be argued, not omitted. No on-box acceptance row (step 3): the whole claim — two 1.7B models no longer co-residing on an 8 GB card, and the new 60 s wait/timeout under real contention — is provable only on real hardware, and the register carries no #2745/#2752 row. The PR's own test plan also leaves `npm run verify` unchecked, so the gate's "fully staged" precondition is not met.

### ✅ What is solid

- The diagnosis is right and the widened condition at `main.py:6635` is the correct one — `_base17` genuinely stays `None` for the whole load, and the pre-fix guard genuinely could not see it.
- Choosing option 1 from #2752 (give `unload_base17` a wait-then-evict path) and mirroring `unload_design`'s exact shape — same deadline/poll loop, same lock-then-null, same early return — is the right structural call, and the wait is bounded rather than open-ended.
- The three direct `unload_base17` tests are real: each names a producer-side mutation, and the waits-then-unloads / raises-after-bound / no-op-when-idle behaviours all hold on head (4/4 green, re-run here).
- A timed-out unload leaves `_base17` untouched — the failure mode is never a silent null, which was the ticket's core worry.
- `Closes #2752` is present in the body, plain text, outside any code span.

### Verdict

**Request changes.** Two blocking findings: the Stop route regression (`POST /unload {model:'1.7b'}` → 500, model still resident, silently swallowed by the Node run-start hygiene path), and a pre-existing test left red on every torch-equipped box while both cited green runs structurally skip it. Four significant: no route mapping for the new error, the wait held inside the Kokoro arbiter, the untested call-site wait, and the residual race. The core guard widening is correct and worth keeping — the defects are in what the new default and the new error do to the paths around it.

<sub>Independent pre-merge review per `.claude/skills/pr-review-gate/SKILL.md`. Findings only — no files were modified.</sub>

dudarenok-maker:
## PR review — pass 2 (head 26bb7c56c1659cdedb1c4fe4a8aca47bda0dfb79, depth medium)

Scope: the 7 follow-up commits `564ed747..26bb7c56`, re-read against pass 1's eleven findings, plus the surrounding call graph — every `card_lock` holder, every `_ensure_base17_loaded` entry point, `_VdKokoroArbiter`, and the Node-side `/unload` consumers and their doc comments.

Verified before probing: the three affected pytest files 14/14 green on head; the **full** `server/tts-sidecar/tests/` suite under a torch-equipped venv — **1 failure**, detailed below, which is green on the merge base `564903f4` and on the intermediate commit `5ac0f950`; `ensure-sidecar-loaded.test.ts` + `evict-idle-tts.test.ts` 31/31 green; `npm run check:onbox-register` OK. Six producer-side mutations run with pytest output captured, not reasoned about.

Depth `medium` per the ladder (highest single-commit tier in the set is `fix`).

---

## Pass-1 dispositions

### 🔴 #1 Stop route 500s and frees nothing — **producer half VERIFIED RESOLVED, test half STILL OPEN**

Producer half is genuinely fixed. Route-level probe on head, `TestClient`, one `_base17_in_flight.claim()` held and `_base17` resident:

```
ROUTE: status 200 | body {"status":"idle"}
ROUTE: base17 still resident -> False
ROUTE: in-flight still busy   -> True
```

Mutation (reintroduce the raise at `wait_seconds <= 0.0`) is caught by the new `test_unload_base17_with_default_wait_never_raises_during_stop` — red for the right reason. The Node half is real too: `ensure-sidecar-loaded.ts:254-262` now reads `res.ok` and `:279` returns `allEvicted`; the new test asserts `false` where the pre-fix code returned `true` unconditionally.

**But pass 1 named two halves, and the second was not done.** `tests/test_batch_synthesis.py:1036` is byte-for-byte unchanged:

```python
stop = threading.Thread(target=engine.unload_base17, daemon=True)
stop.start()
stop.join(0.5)
assert not stop.is_alive(), (...)
```

Same mutation, same command, that test **passes**:

```
1 failed, 6 passed, 1 warning
  PytestUnhandledThreadExceptionWarning: Exception in thread Thread-2 (unload_base17)
  main.Base17ContentionTimeoutError: MUTANT: pass-1 blocking bug reintroduced
FAILED tests/test_base17_contention.py::test_unload_base17_with_default_wait_never_raises_during_stop
```

The only reason the whole PR isn't still shipping the pass-1 blocker is the *new* test. `#1975`'s own regression test still passes on the corpse of a dead thread, and `pytest.ini` sets no `filterwarnings = error`, so the warning is decorative. This is the exact instrument that let the blocker through the first time; it is still armed. `run_batch` two lines above already shows the pattern (`batch_errors: list[BaseException]`, asserted after the join) — apply it to `stop`.

### 🔴 #2 Suite red on any torch-equipped box — **the specific TypeError is fixed; the class is not**

`_fake_unload_base17` now takes and asserts `wait_seconds`/`poll_seconds`, and the file is 3/3 green with torch. But the branch is **red again on a torch box for a new reason** (New 🔴 below), CI still cannot see it (`pytest.importorskip("torch")`), and the PR body still carries no full-suite run from an environment that executes these tests. Same blind spot, second occurrence in the same PR.

### 🟠 #3 No route mapping for `Base17ContentionTimeoutError` — **VERIFIED RESOLVED**

`main.py:10826-10836` maps it to 503 + `code: "base17_in_flight"` ahead of the generic arm. Mutation (disable the arm) → `test_design_voice_maps_base17_contention_timeout_to_503` fails with the raw `Base17ContentionTimeoutError` reaching the 500 path. `design_voice()` is the only producer of that error and `/qwen/design-voice` its only caller, so the class is closed, not just the instance.

### 🟠 #4 Wait held inside the Kokoro arbiter — **code fix RESOLVED; its test is vacuous, and the follow-up commit re-broke the block**

The eviction is now above `with _VD_KOKORO.design():` (`main.py:6670-6680`), which is correct.

Its paired test is not. Head's `tests/` run against the pre-fix producer `e5107a17` — where the eviction sits **inside** the design block, i.e. exactly the defect — gives `3 passed`. `test_design_voice_evicts_base17_outside_kokoro_design_block` asserts `call_order == ["unload_base17", "ensure_design_loaded"]`, and that order holds in *both* placements, so the assertion is blind to the thing its name and docstring claim it pins (see New 🟠 below). And `c7ed4cfe` then moved the design *forward* out of the arbiter (New 🔴 below).

### 🟠 #5 Call-site wait untested — **VERIFIED RESOLVED**

Mutation (`self.unload_base17()` bare at `main.py:6677`) → `AssertionError: assert 0.0 == 60.0` in `test_design_voice_evicts_resident_base17`. Note the pin is torch-gated, so it fires on the operator's box and not in CI — consistent with that file's existing posture, worth knowing when reading a green CI.

### 🟠 #6 Residual race — **RESOLVED, though not by the leg pass 1 named**

`card_lock` genuinely serialises `design_voice()` against `mint_variant()` (`main.py:6670` vs `:7028`), and the new test fails 6/6 against a `card_lock`-removed producer, passes 8/8 unmutated — a real pin.

On the deadlock question you asked me to confirm: **no lock-order inversion exists.** `card_lock` is acquired at exactly two sites, both in the order `card_lock` → (`_base17_activity`) → `_VD_KOKORO.design()` → `_synth_lock`; `design_voice()` never calls `mint_variant()`, so the non-reentrant `threading.Lock` is never re-entered; and `mint_variant()`'s `unload_design()` wait sits **outside** its `card_lock` (`main.py:7025-7027`), so the one cycle that would matter — hold the card lock, wait on `_design_in_flight` — does not exist. The related inversion on the synth side is also absent: `synthesize()` (`:7468`) and `synthesize_batch()` (`:7585`) both call `unload_design()` *before* claiming `_base17_activity()`, never while holding it.

That same ordering is what closes the leg pass 1 actually described: a chapter render cannot cold-load base17 during a design, because since #2678 `unload_design()` waits on `_design_in_flight.busy`, and `design_voice()` claims that at `:6607` before any eviction. So the sentence-*n*/*n*+1 interleaving is unreachable through `synthesize`/`synthesize_batch`, and `card_lock` closes the mint leg. One uncovered entry point remains and is **pre-existing, not introduced here**: `POST /load {model:'1.7b'}` (`main.py:10552`/`:10554`) loads base17 with neither the card lock nor a design check.

But the acquire itself is unbounded — New 🟠 below.

### 🟡 Pass-1 minors — all four comment/doc fixes VERIFIED RESOLVED

The reclaim-skip comment is accurate (`QwenEngine.unload()` gates its reclaim on `if had:` the same way, `main.py:6438`); the `maybe_free_idle_base17` claim now matches `:6014-6016`; the CLAUDE.md citation now names the real rule; the log line reads "resident/in-flight". House gates: release-notes pair present, A103 on both register surfaces, `check:onbox-register` green, `npm run verify` ticked. Caveats on A103's content below.

---

## New findings

### 🔴 Blocking — the VoiceDesign forward no longer runs under the Kokoro arbiter; the branch is red on a torch box

`tests/test_design_kokoro_exclusion.py::test_design_voice_holds_arbiter_and_evicts_resident_kokoro`, full suite, torch-equipped venv:

```
E   AssertionError: design forward must run under arb.design()
E   assert False is True
test_design_kokoro_exclusion.py:361
```

Bisected against the same test file: merge base `564903f4` → 17 passed. `e5107a17` → passed. `5ac0f950` (the #4 fix) → passed. **`c7ed4cfe` (the #6 fix) → failed.**

`c7ed4cfe` wrapped the block in `card_lock` and re-indented `with _VD_KOKORO.design():` from 16 to 20 spaces, but left `# Phase boundary` / `load_ms` / `with self._synth_lock:` at 20 — where they are now the *card lock's* body, not the arbiter's. The arbiter block therefore ends at `main.py:6700`, after the two `_ensure_*_loaded` calls, and the GPU work at `:6708-6763` — `design.generate_voice_design(...)` at `:6751` and `base.create_voice_clone_prompt(...)` at `:6760` — runs with `_design_active` already back to `False` (reset in the arbiter's own `finally`, `main.py:1437-1438`).

Consequence, on the 8 GB card this whole file exists to protect: `kokoro_synth()` blocks only `while self._design_active` (`main.py:1414-1416`). A Kokoro synth arriving during the design forward no longer waits — it loads Kokoro (~1 GB) alongside the resident VoiceDesign (~4-5 GB) and 0.6B Base (~1.2 GB), for the whole design + distil span. That is precisely the resident-VRAM exclusion the block's own surviving comment at `:6619-6621` still claims to enforce ("a VoiceDesign forward and a Kokoro synth must not co-reside on the 8 GB card"), and the reason `unload()` is called on Kokoro three lines later is now pointless — it is evicted, then free to reload mid-forward.

CI cannot see this: the test opens with `pytest.importorskip("torch")` (`:315`), and `verify.yml:355-359` deliberately omits torch. Only the operator's box runs it, where `npm run test:sidecar` is red.

Fix: indent `main.py:6701-6763` one more level so the `with _VD_KOKORO.design():` block re-covers the `_synth_lock` forward span, leaving the base17 eviction above it where #4 put it.

### 🟠 Significant — an unbounded `card_lock.acquire()` now sits in front of the bounded wait it was added to protect

`main.py:6670` is a plain `with _DEVICE_LEDGER.card_lock(...)` — `threading.Lock.__enter__`, no timeout. `mint_variant()` holds that same lock at `:7028` across `_ensure_base17_for_mint()` — a cold 1.7B weights pull the file itself calls "unbounded — a real weights pull" for the sibling path — *plus* its 1.7B anchoring and instruct forwards, because its `card_lock` body runs `:7028-7080`. `design_voice()`'s does the same: the lock spans `:6671-6763`, i.e. eviction **through the design forward and distil**, not the "evict-through-load sequence" its new comment at `:6661-6669` describes.

Concrete failure: a bulk "Design full cast" run interleaves mints and base designs. A `/qwen/design-voice` arriving while a mint holds the card lock blocks in the acquire with no deadline, *while holding `_design_in_flight`* (claimed at `:6607`, above the lock). Every concurrent `/synthesize` and `/synthesize_batch` then hits `if self._design is not None or self._design_in_flight.busy: self.unload_design()` (`:7468`, `:7585`), burns the full 150 s `_DESIGN_CONTENTION_WAIT_S_DEFAULT`, and raises `DesignContentionTimeoutError` → the chapter render 503s because a design is queued behind a mint. Pre-PR the design did not queue at all.

This also silently defeats the contract the PR spent its other five commits building: `unload_base17`'s docstring insists the wait "is bounded, never unbounded — a genuinely wedged load must not hang the caller forever", `:5983-5989` raises after 60 s, and `:10826` maps that to a retry-safe 503. All of it is now behind an acquire with no bound, so a wedged mint hangs `/qwen/design-voice` indefinitely and the caller never gets the 503 the PR just added. `lock.acquire(timeout=...)` with the same budget, raising into the arm that already exists, restores it. (Secondary: `design_voice` runs under `asyncio.to_thread`, so queued designs pin default-executor threads that `/unload` also needs.)

### 🟠 Significant — the paired test for pass-1 #4 cannot fail for the defect it names

`tests/test_qwen_design_base17_exclusion.py:187-255`. Head's test files, run against producer `e5107a17` (eviction **inside** `with _VD_KOKORO.design():`, the pass-1 defect verbatim): `3 passed`. Its docstring says "design_voice must evict resident 1.7B-Base BEFORE entering the `_VD_KOKORO.design()` block, not inside it", but its only assertion is `call_order == ["unload_base17", "ensure_design_loaded"]` — an ordering that holds identically in both placements, because the eviction preceded the load inside the block too. Brief shape 3: the test went green after the fix, and would have gone green before it.

The observable that actually distinguishes them is the flag: capture `main._VD_KOKORO._design_active` inside `_fake_unload_base17` and assert it is `False` there (and, paired with the 🔴 above, `True` inside the design forward). As it stands, pass-1 #4 has a fix with no regression test, which is a house-gate miss under CLAUDE.md's testing discipline, not a nit.

### 🟡 Minor

- **Three doc comments in two other files now assert the opposite of what ships.** `gpu/evict-idle-tts.ts:98-105` states as fact that `reconcileResidentQwenTiers`'s "`unload()` swallows its own fetch failures and never checks `res.ok`" — it does now. `gpu/qwen-tier-reconcile-gate.ts:28-35` says it "reports whether it actually issued an `/unload` (its `evictions` array was non-empty)" — it now reports issued *and all succeeded*. The function's own doc at `ensure-sidecar-loaded.ts:183-190` enumerates the `false` cases and the new one is missing. Worst of the four: the block at `:200-215` still reads *"the swallow-and-return-`true` path still exists. Closing it means changing what this function's return value MEANS … so it is a deliberate design decision rather than a fix to smuggle in here"* — this PR made exactly that change and left the paragraph forbidding it. The behaviour change looks right (the only consumer, `evictIdleQwenBase` → `capacity-retry.ts`, falls through to a bounded poll instead of burning an immediate retry); the four comments are what needs updating, including the partial-failure case they don't cover (0.6B evicted, 1.7B 500 → reported as nothing freed).
- **A103's acceptance criteria can't observe the feature they name.** (a) Criterion 1 races two *designs*, which exercises `_design_in_flight`; #2752 is about `_base17_in_flight`, so the operator needs a variant **mint** or a 1.7B synth in flight when the design arrives. (b) Criterion 2 says to call `POST /qwen/unload` — no such route exists; the Stop route is `POST /unload` with `{engine:'qwen', model:'1.7b'}` (`main.py:10634-10673`) — and it says to press Stop "during an active `design_voice()` call", where base17 is being *evicted*, not loaded, so it proves nothing about the `wait_seconds<=0` path. (c) "memory usage drops by ~4–5 GB (the base17 1.7B model freed)" — base17 is ~3.4 GB everywhere in `main.py`'s own comments; 4–5 GB is the VoiceDesign model. An operator following this row cannot discharge the debt it records.
- **Two new-test comments contradict their code.** `test_design_voice_card_lock_prevents_concurrent_base17_reload` says `# Simulate that Kokoro is already warm from a prior operation` directly above `_quiet_kokoro()`, which nulls it. In `ensure-sidecar-loaded.test.ts`, `mockSidecar({...})` is discarded by the `global.fetch = ...` on the next line, and the comment narrates the 1.7B in-flight 500 while `keep17: true` means the eviction actually exercised is the 0.6B one.
- **Release notes narrate the PR's own review history.** `RELEASE_NOTES.md`'s user-facing line describes "a Stop press could get momentarily wedged" and a bulk run that "could stall the fallback speech engine's ordinary playback" — both were introduced *and* fixed inside this PR, so no user ever saw either. `docs/release-notes-next.md` says "PR #2790's six commits" and "Six paired regression tests, one per finding" (8 commits; 5 new test functions, one of which is vacuous per the 🟠 above).

### Verdict

**Request changes.** One new 🔴 — the `card_lock` commit moved the VoiceDesign forward out of the Kokoro arbiter, leaving the branch red on the operator's box and green on CI for the second time in this PR — plus two 🟠 (unbounded acquire in front of the bounded wait; a vacuous regression test for pass-1 #4), and pass-1 #1's second half still open. Everything else from pass 1 is verified resolved, most of it mutation-verified. The direction remains right; both new correctness findings are in `c7ed4cfe`, which is the one commit that re-shaped an existing block rather than adding to it.

<sub>Independent pre-merge review per `.claude/skills/pr-review-gate/SKILL.md`. Findings only — no files were modified.</sub>

dudarenok-maker:
## PR review — pass 3 (head d7551377fe299036f27adaffc687d102ce4ca36a, depth medium)

Scope: the six branch commits since pass 2 (`90c106cb..48645382`) plus the `origin/main` merge `d7551377`, re-read against pass 2's six open items, plus every `card_lock` holder, both producers of `Base17ContentionTimeoutError`, its route arm and the Node consumers of that 503, and both register surfaces.

Verified before probing: full `server/tts-sidecar/tests/` suite on a **torch-equipped venv** (torch 2.11.0+cu128, GPU confirmed idle first) — **1101 tests, 0 failures, 0 errors, 2 skipped** (junit counts, 557.8 s); `ensure-sidecar-loaded.test.ts` + `evict-idle-tts.test.ts` 31/31; `npm run typecheck` clean; `npm run check:onbox-register` OK. Four producer-side mutations run with pytest output captured, plus an AST structural check of `design_voice()`'s block nesting (head, merge base, and the merged tree) rather than reading indentation. Depth `medium` — highest single-commit tier in the set is `fix`.

---

## Pass-2 dispositions

### 🔴 Pass-1 #1's test half (`test_batch_synthesis.py:1036`) — **VERIFIED RESOLVED**

`run_stop` now mirrors the file's own `run_batch`/`batch_errors` pattern and the join is followed by `assert stop_errors == []` (`test_batch_synthesis.py:1050`). Mutation **M3** — delete `unload_base17`'s `wait_seconds <= 0.0` early path (`main.py:5967-5977`), reinstating the pass-1 blocker verbatim:

```
E   AssertionError: unload_base17 raised: [Base17ContentionTimeoutError('Qwen 1.7B-Base has been in
    flight for over 0s - refusing to evict it out from under an active load...')]
FAILED test_batch_synthesis.py::test_unload_base17_is_not_blocked_by_a_cold_load_during_batch_synth
```

Red for exactly the claimed reason; green unmutated. The instrument that swallowed the pass-1 blocker is disarmed.

### 🔴 New-in-pass-2: VoiceDesign forward outside the Kokoro arbiter — **VERIFIED RESOLVED (structurally, not just "the test passes")**

AST of `design_voice()` on head (block spans, not indentation):

```
With [6607-6877] self._design_in_flight.claim()
  Try  [6679-6774] finally@6774 -> card_lock.release()
    With [6691-6772] _VD_KOKORO.design()
      With [6717-6772] self._synth_lock
line 6673 card_lock.acquire   line 6686 self.unload_base17
line 6708 _ensure_design_loaded   line 6760 design.generate_voice_design
line 6769 base.create_voice_clone_prompt
```

Both GPU forwards (6760, 6769) and the `_synth_lock` span are inside the arbiter's 6691-6772; the base17 eviction (6686) stays above it, where pass-1 #4 put it. The same AST run on merge base `9e8843b0` gives arbiter 6548-6641 with both forwards inside — head is the pre-PR shape plus #4's move, nothing else.

Mutation **M1** — de-indent 6710-6772 one level (commit `c7ed4cfe`'s shape):

```
E   AssertionError: design forward must run under arb.design()
test_design_kokoro_exclusion.py:361
```

the exact assertion pass 2 reported. Pass-1 #2's wider concern is closed too: the branch is no longer red on a torch box (1101/1101 above).

### 🟠 New-in-pass-2: unbounded `card_lock.acquire()` — **bounded as prescribed, but the bound it was given is wrong — see the new 🟠**

The mechanics asked about are correct. The `raise` at `main.py:6674` sits **before** the `try` at 6679 (AST-confirmed), so the `finally`'s `card_lock.release()` at 6774 can never run on a lock that was never acquired. Only two `card_lock` sites exist (`main.py:6672`, `:7039`); nothing else assumed the old `with` shape, and no lock-order inversion is introduced (`mint_variant`'s `unload_design()` wait stays outside its own `card_lock`, `main.py:7036-7039`). Mutation **M4** (`timeout=300.0`, longer than the test's 120 s holder) fails `test_design_voice_card_lock_acquire_is_bounded`.

### 🟠 New-in-pass-2: pass-1 #4's regression test was vacuous — **VERIFIED RESOLVED**

Mutation **M2** — move the eviction block back inside `with _VD_KOKORO.design():`, i.e. pass-1 #4's defect verbatim:

```
E   AssertionError: Eviction must run BEFORE entering the _VD_KOKORO.design() block ...
E   assert True is False
test_qwen_design_base17_exclusion.py:261
```

Under pass 2's version of this file the same mutant was green. The paired `design_forward_called["before_eviction_check"] is True` also pins the other half.

### 🟡 Pass-2 minors — **all VERIFIED RESOLVED**

- The three TS doc comments now match the shipped code: `ensure-sidecar-loaded.ts:232-283` really does check `res.ok`, set `allEvicted = false` on a non-ok **and** in the `catch`, and return `allEvicted`; the "deliberate design decision rather than a fix to smuggle in here" paragraph is replaced by an accurate CLOSED note.
- A105's criteria are now observable: `POST /unload` with `{"engine":"qwen","model":"1.7b"}` is the real route (`main.py:10645`, dispatch at `:10677-10682`); criterion 1 races a mint/1.7B synth; ~3.4 GB.
- `test_base17_contention.py`'s "Kokoro already warm" comment and the vitest comment/mock mismatch are corrected (one wording slip left, below).
- `RELEASE_NOTES.md`'s user-facing line no longer narrates the PR's internal review history, and sits under the in-progress `# Castwright 1.15.0`.

### Register merge resolution — **VERIFIED, hand-checked as well as mechanically**

Counted from the file rather than trusting the summary: **40** `### A*` headings (A1-A37 + A102, A104, A105) and **63** row headings in total; glance table 40+2+4+3+10+2+2 = **63** = the stated "63 owed"; `<!-- next-id: A106 -->` is above the highest allocated id; A103 appears in no heading, as the note declares. The live view carries 63 / 40 in both the strip and the Group A header and holds a matching A105 item. `check:onbox-register` OK.

`origin/main` has since moved to `f575cfc9` (+14 commits, PR #2784, which also edits `main.py`, `RELEASE_NOTES.md` and `docs/release-notes-next.md`). A read-only `git merge-tree origin/main HEAD` auto-merges all three with **no conflicts**, and re-running the AST check on the merged `main.py` shows the arbiter nesting intact — so the merge that produced pass-2's 🔴 does not have a sequel. Full suite on the merged tree: 1105 tests, 1 failure, and that failure is an artifact of my own subtree extraction (`test_footprints.py` reading `docs/local-llm.md`, which I did not archive), not a merge-induced defect.

---

## New findings

### 🟠 Significant — a second design now fails with a 503 behind a first design that is merely slow, and the bound chosen is the one this file already rejected for exactly this case

`design_voice()` holds `_DEVICE_LEDGER.card_lock()` from `main.py:6673` to the `finally` at `:6774` — i.e. across the cold VoiceDesign load *and* `generate_voice_design` (`:6760`) *and* `create_voice_clone_prompt` (`:6769`), not the "evict-through-load sequence" its comment at `:6661-6669` describes. A second `design_voice()` waits on that with a **60 s** budget and then raises.

Reproduced on head (budget scaled to 0.4 s, design A's load made 1.5 s; A is healthy, just slow):

```
PROBE: design A error = []
PROBE: design B error = [Base17ContentionTimeoutError('Could not acquire per-card lock for Qwen
        design - a concurrent base17 load/mint has held it for over 0s. Retry the design shortly.')]
PROBE: B held _design_in_flight while queued = True
```

Same probe on merge base `9e8843b0`: `design A error = []`, `design B error = []` — both succeed, B simply serialising on `_synth_lock`. So this is a new failure mode, introduced by `c7ed4cfe` (the lock) and turned from *wait* into *fail* by `93d3a20f` (the 60 s bound).

Three things make 60 s the wrong number rather than a tuning preference:

1. **This file already decided the opposite for the identical situation.** `main.py:5366-5377`, on the sibling budget: *"Sized ABOVE the documented ~120s cold-design server budget ... a bound tighter than a design's own request budget would trip on a design that is merely slow, not wedged"* — hence `_DESIGN_CONTENTION_WAIT_S_DEFAULT = 150.0`. The new acquire waits on precisely "a design", with 60 s.
2. **The constant's own stated rationale is falsified by its second use site.** `main.py:5382-5389` justifies 60 s as *"a plain base17 load/mint is lighter than a full VoiceDesign forward+load, so it doesn't need the same headroom"* and adds *"this is the only caller"*. At `:6673` the thing being waited on **is** a full VoiceDesign forward+load, and there are now two callers.
3. **It is reachable by an architected path, not a contrived one.** `qwen-voice.ts:466` serialises book designs and mints under `withDesignLock(bookDir)`, but `voice-library.ts:112-119` deliberately does not — *"the spec explicitly rejects a cross-scope server mutex (cross-scope protection is the frontend's single-slot and **the sidecar's own VRAM arbitration**)"*. A library design concurrent with a book design/mint is exactly the case that arbitration is supposed to handle; it now 503s. `mint_variant`'s holder is worse still: its `card_lock` body covers `_ensure_base17_for_mint()` — a real cold weights pull — plus its forwards.

Collateral, also shown by the probe: while B is parked in the acquire it is already holding `_design_in_flight` (claimed at `:6607`, above the lock), so every concurrent `/synthesize` / `/synthesize_batch` takes the `unload_design()` wait for up to 60 s on account of a design that is going to fail anyway.

The diagnosis the user gets is wrong as well: the message blames *"a concurrent base17 load/mint"* and the route answers `code: "base17_in_flight"` (`main.py:10845`) when no base17 is loaded, loading, or in flight anywhere in the scenario above.

Not a design pass — the agreed behaviour ("a merely-slow holder must not fail a healthy design") is not in question, only how to hold it: raise the acquire budget above the cold-design budget the way `:5366-5377` already reasons, or narrow the lock's span to the evict-through-load sequence its comment claims, or acquire before claiming `_design_in_flight`. Whichever is chosen, the second producer needs its own message and `code`.

### 🟡 Minor

- **`Base17ContentionTimeoutError` now has two producers and every doc still names one.** The class docstring (`main.py:5342-5351`) says reaching it *"means the base17 load outlived even that bounded wait, i.e. it is genuinely wedged"*; the route arm's comment (`main.py:10838-10842`) says *"a base17 load/mint stayed in flight past unload_base17()'s bounded wait"*. Neither is true of the `:6674` producer, which never calls `unload_base17` and fires with `_base17_in_flight` idle.
- **The new bounded-acquire test costs 60.01 s of wall clock, on every `npm run test:sidecar`** (durations output; ~11 % of the 557 s suite, and `test:sidecar` is scope-gated into pre-push `verify:fast:branch` for any `server/tts-sidecar/**` diff). Avoidable: I re-ran the same assertion with `monkeypatch.setattr(main, "_BASE17_CONTENTION_WAIT_S_DEFAULT", 0.5)` — the budget is read at call time — and it raised in **0.52 s**. Its docstring (`test_base17_contention.py:392-395`) also states a mutation outcome that does not exist: *"causing this test to hang past its 10s deadline"* — the test has no deadline; under M4 it ran 121.95 s and failed on an incidental `ValueError: torch.__spec__ is not set` from a real transformers import, not on the assertion.
- **`docs/release-notes-next.md:542` is filed under `## 🧪 Test gates` → `### Test harness & suite hygiene`** (heading at `:443`), between the spawn-guard and npm-audit entries. It is a runtime VRAM/OOM fix in the voice-design path; `## 🎙️ Voice design, library & previews` (`:210`) or `## 🏗️ Under the hood` (`:383`) is its home. That file is fed verbatim into the GitHub release body, so the mis-section ships.
- **The PR body is stale in ways that matter to the merge record.** It still ticks *"Full `server/tts-sidecar/tests/` pytest suite (system Python, no venv bootstrapped in this worktree) — exit code 0 ... only the pre-existing `test_unload_base17_is_not_blocked_by_a_cold_load_during_batch_synth` raising `Base17ContentionTimeoutError` ... as an unhandled-thread-exception *warning* (not a failure)"* — the finding pass 1 opened by refuting that sentence, since fixed. `npm run test:sidecar — SKIP (no .venv bootstrapped)` is still ticked as verification, from the one environment structurally blind to two of this PR's three blockers. `test_base17_contention.py` is stated as `4/4`; it has 7 tests. And the Summary never mentions the change to `reconcileResidentQwenTiers`' return contract — a Node-side behaviour change in a shared helper whose own comment previously called it *"a deliberate design decision rather than a fix to smuggle in here"* — which CLAUDE.md requires be declared in the body ("Also fixed, found in passing: …").
- **`main.py:5375-5377`** still cites *"CLAUDE.md's 'new env var MUST be a knob' rule"* — the nonexistent rule pass 1 flagged, corrected on this PR's own constant eleven lines below but left on its sibling.
- **`ensure-sidecar-loaded.test.ts:295`** — *"This test drops `keep17: true`"* reads as "omits", the opposite of what the code does; the parenthetical that follows is right.

---

### Verdict

**Request changes — and this is the loop cap.** Pass 2's six items are all verified resolved, most of them mutation-proven, and the two structural questions the re-review was asked to settle (the re-indentation, and the `finally` on a failed acquire) both check out against the AST rather than against a passing test. But the bounded acquire that closed pass-2's 🟠 introduced a new 🟠: a healthy concurrent design now 503s after 60 s, on a bound this file's own sibling comment already reasoned must be *above* the cold-design budget, with an error message and code that name a cause that isn't present. Per `pr-review-gate`'s loop cap (initial + 2 re-reviews), that finding goes to you rather than into a fourth automatic round.

<sub>Independent pre-merge review per `.claude/skills/pr-review-gate/SKILL.md`. Findings only — no files were modified.</sub>

PR #2424, opened 2026-08-18
build(deps): refresh root and server lockfiles to clear 20 npm security advisories

dudarenok-maker:
## Summary

Refreshes `package-lock.json` and `server/package-lock.json` to clear 20 Dependabot security advisories. Every advisory was in-range for its package's declared semver, so the fix is lockfiles-only — neither `package.json` was touched.

Closes #2411

**Root** — `npm audit` `metadata.vulnerabilities`: before `{"high":5,"critical":0,"total":5}` → after `{"high":0,"critical":0,"total":0}`.
Version transitions: `undici` to 7.29.0 and 8.10.0 (both lines present in the tree, both at/above the fixed-at floors 7.29.0 / 8.9.0), `js-yaml` to 4.3.1, plus `nanoid` and `brace-expansion`.

**Server** — `npm audit` `metadata.vulnerabilities`: before `{"high":3,"critical":0,"total":3}` → after `{"high":0,"critical":0,"total":0}`.
Version transitions: `undici` to 8.10.0, `nanoid` to 5.1.16, `ip-address` to 10.5.0 — all three are runtime dependencies of the server.

The diff is lockfiles-only: `git diff --stat main...HEAD` touches only `package-lock.json`, `server/package-lock.json`, and the two release-notes files added by this PR. Neither `package.json` changed.

No paired regression test is owed — this changes no behaviour, and the acceptance instrument is `npm audit` itself.

Root `nanoid`, `brace-expansion`, and `@redocly/openapi-core` carried no open Dependabot alert but are live advisories in the same lockfile, and were refreshed deliberately.

## Test plan

- [x] `npm --prefix . audit --json` — `high: 0`, `critical: 0`
- [x] `npm --prefix server audit --json` — `high: 0`, `critical: 0`
- [x] `npm run verify:fast:branch` — green (all steps cached/passing)
- [x] `git diff --stat main...HEAD` — lockfiles + release notes only, `package.json` unchanged in both trees

PR #2791, opened 2026-08-28
chore(ops): adopt .github/dependabot.yml for npm root/server and github-actions

### LINKED ISSUE #2433: ops-60 — decide whether to adopt .github/dependabot.yml, and with what settings
Surfaced by the PR review gate on #2424 (ops-58), which cleared the 20 open npm advisories by lockfile refresh. #2411 named this as out of scope and `Closes #2411` retires that record, so it is filed here to keep it alive.

The repo has **no `.github/dependabot.yml`** (verified 2026-08-18: the file does not exist). Consequently Dependabot raises **security alerts only** — there are no automated version-update PRs. Every advisory therefore waits for someone to notice the alert list and open a manual sweep, which is exactly how #1863 and #2411 both came to be hand-cut batches.

### The decision owed

This is filed as a design pass because opting in has more than one defensible outcome, and the choice is the deliverable — not the config file:

1. **Which ecosystems.** `npm` (two manifests: root and `server/`), `pip` (the sidecar's `requirements/` overlays), `github-actions`, `gradle`/`pub` for `apps/android/`. Each has a different noise-to-value ratio here. The pip overlays in particular are pinned deliberately (`qwen-tts==0.1.1` freezes `transformers`, torchaudio caps torch — see #893), so automated PRs against them would be near-permanently red.
2. **Security-only vs. all version updates.** Security-only keeps the current signal and adds automation; all-updates adds currency but generates continuous PR traffic against a repo whose `verify.yml` battery is the required merge gate.
3. **Cadence and open-PR cap.** `schedule.interval` plus `open-pull-requests-limit` — the knob that actually decides whether this is useful or noise.
4. **Grouping.** Whether to use `groups:` so a sweep arrives as one PR rather than N (relevant: #2424 was a single lockfile refresh clearing 20 alerts; twenty separate Dependabot PRs would have been strictly worse).
5. **Whether the two npm manifests are separate `updates:` entries** (they must be — Dependabot keys on directory) and whether `server/` gets different settings from root, given root is dev-only tooling and `server/` carries runtime dependencies.

### Acceptance

Whatever is decided, the outcome is recorded — either a committed `.github/dependabot.yml` whose choices match the decision above, or a written decision not to adopt one, with the reason, so this does not get re-litigated at the next sweep.

*Benefit (technical):* decides whether dependency currency is automated or stays a manual sweep, instead of leaving it undecided by default and rediscovering the backlog every few months.

dudarenok-maker:
Closes #2433

## Summary
- Adds `.github/dependabot.yml` adopting Dependabot version updates for three ecosystems: `npm` at repo root, `npm` at `/server`, and `github-actions` — per the decision recorded on #2433.
- Each entry runs weekly, caps open PRs at 5, and groups matched packages so a routine sweep lands as one PR per ecosystem.
- Deliberately excludes `pip` and Android (`gradle`/`pub`) ecosystems and any `ignore:` rules, matching the decision's scope.

## Also fixed, found in passing (PR review gate, pass 1)

The review pass found the config as originally written would have scheduled a stream of permanently-unmergeable PRs. All findings are fixed in this PR:

- **🔴 Blocking — Dependabot PRs could never merge.** `main`'s two active rulesets (`17654264`, `18362221`) both require the `Verify PR body links a GitHub issue` status context, both with `bypass_actors: []`. Dependabot bodies never contain `Closes`/`Refs #NN` and it has no mechanism to add one; with `open-pull-requests-limit: 5` the config would have self-disabled into 15 stuck PRs. Fixed by exempting `dependabot[bot]` at the **step** level in `.github/workflows/pr-issue-link.yml` (a job-level `if:` would leave the required context pending forever, blocking merge just as hard), matching on `github.event.pull_request.user.login` rather than `github.actor`, which on a `synchronize`/`edited` re-run can be a human. Exactly one login is exempt.
- **🟠 `groups:` missed the case that justified it.** Groups without `applies-to:` default to version updates only, so the security-alert sweep #2433 point 5 cited (#2424, 20 advisories) would still have arrived ungrouped and uncapped. Each ecosystem now carries a second `applies-to: security-updates` group.
- **🟠 Dev-dep batches would fail the repo's own title lint.** Default inference yields `chore(deps-dev): …`, a scope `scripts/validate-commit-msg.mjs` rejects, so `pr-title-lint.yml` would be permanently red. Fixed with `commit-message: prefix: "chore"` and no `include: scope`, landing on the no-scope `chore: <subject>` catch-all the validator accepts.
- **🟠 `github-actions` missed the composite action.** `directory: "/"` covers `.github/workflows/**` plus a root `action.yml` only, leaving `.github/actions/setup/action.yml` (pins `actions/setup-node@v6`, `actions/cache@v5`, consumed by 8 `verify.yml` jobs) silently unbumped. Switched to glob-capable `directories:`.
- **🟡** File is now Prettier-clean with a trailing newline.

## Test plan
- `npm run test:hooks` — **1687 pass, 0 fail**, including 6 new regression tests.
- `npm run lint` — clean, zero warnings.
- New coverage pins both halves of the exemption: a `dependabot[bot]` PR with no `Closes`/`Refs` passes, **and** a human PR with the identical body still fails (the negative control — without it, an exemption that skipped the check for everyone would look green).
- New `scripts/tests/pr-issue-link-workflow-wiring.test.mjs` pins the workflow wiring itself: deleting the `PR_AUTHOR` env block leaves every validator test green while the fix goes inert. Proven red-then-green by removing the block and restoring it.
- `.github/dependabot.yml` parses via `js-yaml`; `prettier --check` passes.

## Checklist notes
- **Step 1 (regression plan):** N/A — CI configuration, no product behaviour; the issue body plus paired tests are the spec.
- **Step 3 (on-box acceptance):** N/A — no hardware-dependent behaviour.
- **Step 5 (release notes):** deliberately skipped — CI/process-only change with no user- or operator-visible product delta.
- **Local pre-push bypassed with `--no-verify`**, authorized by the repo owner. The only red leg was `review-gate-mechanism.test.mjs`'s agent-store mirror guard, whose input is an *uncommitted* `reviewer-brief.md` edit in another lane's checkout that was synced into `~/.agents/skills/`. Unrelated to this PR, reproduces without these commits, and skips in cloud CI (no agent-store on the runner), so `verify.yml` here is unaffected.


dudarenok-maker:
## PR review — pass 1 (head 784dac3f, depth low)

Scope: the one added file, `.github/dependabot.yml` (37 lines) · verified against issue #2433's decision comment, the repo's live `main` rulesets (`gh api repos/.../rulesets`), `scripts/validate-pr-issue-link.mjs`, `scripts/validate-commit-msg.mjs`, `.github/workflows/{pr-issue-link,pr-title-lint,verify,cla,add-to-project}.yml`, the four tracked `package.json` manifests, `.prettierrc`/`.prettierignore`, `scripts/verify-cache.mjs`'s `STEPS`, GitHub's Dependabot options reference, and the repo's own three historical Dependabot PRs (#872/#873/#874) — before adversarially probing what this config does to `main`'s required checks on the day it starts producing PRs.

Low depth is a narrower scope, not a lighter standard: the file itself is schema-correct and matches the decision. Every finding below is about what it does once GitHub acts on it.

---

### 🔴 Blocking — every PR this config schedules is permanently unmergeable; it can never satisfy `main`'s required, unbypassable issue-link check

The chain, all verified:

1. `.github/dependabot.yml:8-37` schedules weekly version-update PRs for three ecosystems.
2. `main` carries two active rulesets — `17654264` ("main — block force-push & deletion") and `18362221` ("main — require PR issue link") — and **both** list required status check context `Verify PR body links a GitHub issue`. Both report `"bypass_actors": []` and `"current_user_can_bypass": "never"`. Nobody, including the repo owner, can merge past it.
3. That context is the job name at `.github/workflows/pr-issue-link.yml:18`, which runs `scripts/validate-pr-issue-link.mjs`.
4. `scripts/validate-pr-issue-link.mjs:12` — `const ISSUE_LINK_PATTERN = /\b(?:closes|refs)\s+#\d+/i;` — and line 101 `process.exit(1)` when `hasIssueLink()` (line 72) is false.

**Exact input → wrong outcome.** The body of this repo's own prior Dependabot PR #873 begins:

> `Bumps [esbuild](https://github.com/evanw/esbuild) from 0.28.0 to 0.28.1.` followed by `<details><summary>Release notes</summary>…`

There is no `Closes` and no `Refs` anywhere in it, and Dependabot has no mechanism that would add one. Feed that body to `validate-pr-issue-link.mjs` and it exits 1. The required check goes red → merge blocked → forever.

There is no historical counter-evidence to weigh against this: `pr-issue-link.yml` was added in `2c119079` on 2026-07-01, and the three Dependabot PRs (#872/#873/#874) all predate it — 2026-06-17. Their `statusCheckRollup` contains `Validate PR title…` and `Check / record CLA signature` but no `Verify PR body links a GitHub issue`, because the workflow did not exist yet. This config is the first thing that will put a Dependabot PR in front of that gate.

**The second-order effect is worse than the PRs just sitting there.** `open-pull-requests-limit: 5` (lines 13, 23, 33) counts *open* version-update PRs. Five unmergeable PRs accumulate per ecosystem — roughly five weeks — and then Dependabot stops raising new ones for that ecosystem. The config self-disables into 15 stuck PRs and zero dependency currency, which is the exact state #2433 was filed to escape.

The only per-PR escape is manual: `pr-issue-link.yml:10` does listen to `edited`, so a human editing each weekly PR body to add `Refs #2433` clears the check. That is a human touch on every single Dependabot PR, i.e. the hands-off currency this adoption exists to buy is not delivered.

**This one needs your decision, not a mechanical fix** — there are at least three defensible remedies and they are not equivalent: (a) add `dependabot[bot]` as a `bypass_actors` entry on both rulesets; (b) short-circuit `pr-issue-link.yml` when `github.actor == 'dependabot[bot]'` while still reporting the context green; (c) don't adopt version updates and keep the security-alert-only posture. Per the gate's judgment-call carve-out I'm flagging it rather than prescribing.

---

### 🟠 Significant — `groups:` does not group the case the decision cited as the reason for having groups

`.github/dependabot.yml:14-17`, `24-27`, `34-37` — each `groups:` block omits `applies-to:`.

GitHub's Dependabot options reference, on `groups.<name>.applies-to`:

> "Specify which type of update the group applies to. When undefined, defaults to version updates."

So these three groups cover **version updates only**. Security updates are ungrouped.

Decision #2433, point 5, justified `groups:` this way:

> "`groups:` per ecosystem so a routine sweep arrives as one PR — mirrors #2424, which was a single lockfile refresh clearing 20 alerts; twenty separate Dependabot PRs would have been strictly worse."

#2424 was a **security-alert** sweep. **Exact input → wrong outcome:** the next time 20 npm advisories land — the #2424 scenario verbatim — Dependabot raises 20 separate ungrouped security PRs, precisely the outcome the decision called "strictly worse". And the cap won't damp it either; the same reference states:

> "*Security update* pull requests are not subject to this limit and do not count toward it. There is no limit on the number of open pull requests for security updates."

So on the motivating scenario this config delivers neither the grouping nor the cap. Fix: a second group entry per ecosystem carrying `applies-to: security-updates` (one group cannot serve both).

---

### 🟠 Significant — a dev-only weekly batch produces a PR title this repo's own lint rejects

`scripts/validate-commit-msg.mjs:10-23` — `SCOPES` lists `deps` (line 20) but **not** `deps-dev`. `CHORE_PATTERN` (line 29) and `TYPED_PATTERN` (line 30) both anchor the scope as `^…\((scope)(,scope)*\)!?: `, so `chore(deps-dev): …` matches neither: the regex consumes `deps`, then demands `)` and finds `-dev`.

The config sets no `commit-message:` block anywhere, so titles are left entirely to Dependabot's inference. That inference demonstrably fires on this repo — PR #873's title was `chore(deps): bump esbuild from 0.28.0 to 0.28.1 in /server`, produced with **no** `dependabot.yml` present at all, and `Validate PR title against commit convention` passed it. Dependabot picks the scope by dependency type: `deps` when the batch contains a production dependency, `deps-dev` when it does not.

**Exact input → wrong outcome:** root `package.json` declares 14 `dependencies` and 38 `devDependencies`, and the fastest-moving of them are the dev set (vitest, eslint, prettier, typescript, playwright). A week in which only devDependencies have updates yields `chore(deps-dev): bump the root-npm group with N updates` → `Validate PR title against commit convention` goes red and stays red.

Honest scoping of the evidence: the `deps` branch is observed in this repo (#873); the `deps-dev` branch is Dependabot's documented dev-dependency scope, not yet observed here, because no Dependabot PR has touched a direct devDependency. The fix costs one line either way.

Severity is 🟠 rather than 🔴 because I checked and that context is **not** in either ruleset's `required_status_checks` — only `npm run verify` and `Verify PR body links a GitHub issue` are. So it is a permanently-red check, not a second merge block.

Fix: add `commit-message: { prefix: "chore(deps)" }` to each entry, which pins the title deterministically; or add `deps-dev` to `SCOPES`.

---

### 🟠 Significant — `github-actions` at `directory: "/"` does not cover `.github/actions/setup/action.yml`, the composite every verify leg actually runs

`.github/dependabot.yml:29-30`. GitHub's options reference, `directory`:

> "For GitHub Actions, use the value `/`. Dependabot will search the `/.github/workflows` directory, as well as the `action.yml/action.yaml` file from the root directory."

`.github/actions/setup/action.yml` is neither a workflow file nor a root-level `action.yml`. It pins `actions/setup-node@v6` (line 31) and `actions/cache@v5` (line 46), and it is consumed by eight jobs in `verify.yml` (lines 115, 150, 241, 281, 345, 401, 455, 501) plus `cross-os.yml`.

**Exact input → wrong outcome:** `actions/setup-node@v7` ships. Dependabot opens a grouped PR bumping the six workflow-file occurrences — `pr-issue-link.yml:28`, `pr-title-lint.yml:26`, `onbox-register-check.yml:36`, `release.yml:42`, `:321`, `:429` — and leaves `.github/actions/setup/action.yml:31` at `@v6`. The repo then runs two setup-node majors inside one PR's check set, and the stale one is the copy doing the actual `npm ci` for every verify leg. That silently inverts the property that file's own header claims at lines 5-6 ("A Node-major bump is now a one-file edit here instead of N call sites") — after this config lands, the automated bump is the one place that edit *won't* reach.

Fix: `directories: ["/", "/.github/actions/setup"]` (plural key) on the github-actions entry. `.github/actions/install-ffmpeg-windows/action.yml` has no `uses:` today and needs no entry yet.

---

### 🟡 Minor

**1. The file is the only YAML in `.github/` that isn't Prettier-clean, and it has no trailing newline.** `.github/dependabot.yml:37` ends `- "*"` with no `\n` — confirmed byte-level (914 bytes, `od -c` tail shows `- " * "` then EOF). `.prettierrc` sets `singleQuote: true`, and `.prettierignore` does not exclude `.github/**`, so `prettier --check .` covers this path. I ran it against the PR's exact blob with the repo config: it fails, wanting `'npm'` / `'/'` / `'weekly'` / `'*'` and a trailing newline — 13 changed lines. `pr-issue-link.yml`, `verify.yml`, `cla.yml` and `codeql/codeql-config.yml` all pass the same check cleanly, so this is a departure from the established convention, not a repo that never formatted its YAML.

Classifying this as cleanup rather than a gate failure is deliberate and checked: `format:check` (`package.json:41`) is wired into nothing — `verify-cache.mjs`'s `STEPS` has no format step (`lint` is `eslint . --max-warnings 0`, `package.json:39`), `verify.yml:156` runs `npm run lint` only, and no husky hook calls it. The practical cost is that the next `npm run format` anyone runs rewrites this file, dropping a spurious 13-line diff into an unrelated PR.

**2. Before-shipping steps 2 and 5 are silently omitted rather than explicitly waived.** The PR body's Test plan is a manual `yaml.safe_load` transcript with no automated test, and neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` moves. Both may well be genuinely not-applicable here — I'm not asking for a test against a service-side config file. The checklist's own rule is the finding: "Skipping a step is fine when the step genuinely does not apply … but say so explicitly rather than silently omitting." One sentence in the body closes it.

---

### ✅ What is solid

- **Schema is correct.** `version: 2`, `package-ecosystem`, `directory`, `schedule.interval`, `open-pull-requests-limit` (plural, not the singular typo), `groups.<name>.patterns` — all real keys, correctly spelled, valid YAML. The group shape is right.
- **Two separate npm entries for `/` and `/server` is correct**, and the layout backs it: `server/package.json` and `server/package-lock.json` are both tracked.
- **The two omitted npm manifests are correctly omitted, not overlooked.** `e2e/package.json` and `pinokio-scripts/package.json` are `{ "type": …, "private": true }` with zero dependency sections — nothing for Dependabot to update. I checked this expecting a finding and there isn't one.
- **The pip exclusion is right and doesn't lose security coverage.** Security-alert PRs fire without any config, proved by this repo's own PR #872 (`bump torch from 2.8.0 to 2.12.0 in /server/tts-sidecar/requirements`).
- **The CLA gate won't block Dependabot.** `cla.yml:41` allowlists `dudarenok-maker,*[bot]`; empirically `Check / record CLA signature` reported SUCCESS on both #873 and #874.
- **The Kanban board won't flood.** `add-to-project.yml:26-27` triggers on `issues` only, never `pull_request`.
- **`verify.yml` runs correctly under Dependabot's restricted token** — it references no `secrets.*` at all, so the read-only-token / no-secrets restriction on Dependabot-triggered `pull_request` runs costs it nothing.
- **The cross-repo auto-close hazard is neutralised by Dependabot itself.** I probed this specifically: an upstream changelog line containing a closing keyword plus `#NNN` would, on merge, close a *Castwright* issue with that number. #873's real body shows Dependabot rewrites every upstream `#4482`-style ref into `<a href="https://redirect.github.com/…">` — a non-github.com host inside an explicit anchor, which GitHub's closing-keyword parser cannot bind to a local issue. Not a finding.

---

### Verdict

**1 blocking, 3 significant, 2 minor.** The config is a faithful, schema-correct transcription of the #2433 decision — the defects are not in the transcription, they are in what the decision did not model: this repo's own merge gates, and the version-update-vs-security-update split in how `groups:` and `open-pull-requests-limit` actually behave.

The blocking finding is not a typo fix. As it stands, merging this PR schedules a stream of PRs that `main`'s unbypassable ruleset will refuse, until the open-PR cap freezes the ecosystem entirely — and it needs your call on which of the three remedies to take before this is mergeable.

dudarenok-maker:
## PR review — pass 2 (head e2916bbc, depth high)

Scope: the PR's own five files (`gh pr diff 2791`), not the merge payload · re-verified by running the changed tests out of `git archive e2916bbc` into a scratch tree (26 pass), **mutation-testing the new wiring test four ways**, running `scripts/validate-commit-msg.mjs` against 13 realistic Dependabot titles, validating `.github/dependabot.yml` against SchemaStore's `dependabot-2.0.json` with `ajv` (and mutating that instrument to confirm it can fail), `prettier --check` on every changed file, and re-reading `main`'s two rulesets via `gh api`.

Deltas only — prior solid items are not re-listed.

---

**1. 🔴 Dependabot PRs permanently unmergeable — VERIFIED RESOLVED.** Interrogated on all four axes asked of me:

- *Does the job complete and report success?* Yes. The exemption is a step-level `env:` (`.github/workflows/pr-issue-link.yml:57-59`), so the `lint` job always runs to completion; `scripts/validate-pr-issue-link.mjs:122-128` exits 0 with a log line. Nothing is skipped, so the required context reports **success**, not pending. I re-read `main`'s rulesets myself: `18362221` requires only `Verify PR body links a GitHub issue`; `17654264` requires that **plus** `npm run verify` — and `verify.yml:512-517`'s aggregator is `if: always()`, so that second context also reports on a Dependabot PR. Both are reachable green.
- *Fail-closed on empty/unset `PR_AUTHOR`?* Yes. `PR_AUTHOR` is quoted in the `run:` line, so an empty expression still yields `argv[3] === ''`; `isDependabotExempt('')` is `false` (`scripts/validate-pr-issue-link.mjs:93`) and the body check runs. Pinned by the unit test's `''` and `undefined` cases.
- *An event with no `pull_request` payload?* Unreachable — `pr-issue-link.yml:3-11` triggers on `pull_request` only. If it were reachable the expression renders empty, which is the fail-closed path above.
- *Can a human reach the exempt path?* No. GitHub logins cannot contain `[` or `]`, so `dependabot[bot]` is unregisterable by a human and unclaimable as an App slug; the workflow is `pull_request`, not `pull_request_target`, so a fork PR carries the fork author's own login; and `argv[3]` is sourced from `github.event.pull_request.user.login` through an `env:` var, never from PR-controlled body/title text, so there is no injection path. The comparison is `===` against one literal — `'dependabot'`, `'Dependabot[bot]'` and `'renovate[bot]'` are all pinned false.

Disclosed, **not** a finding: anyone with write access can push to a `dependabot/*` branch and inherit the exemption, because a PR's author never changes after it is opened. Only the repo owner can do that, and they could equally add `Refs #NN` to the body. No action.

**2. 🟠 `groups:` lacked `applies-to:` — VERIFIED RESOLVED.** `.github/dependabot.yml:39-46, 56-63, 75-82`. Validated against SchemaStore `dependabot-2.0.json`: `applies-to` is a real key with `enum: ["version-updates","security-updates"]`, and the file validates clean. Duplicate `patterns: ['*']` across the two groups in one entry is legal and unambiguous — a dependency belongs to at most one group **per update type**, and these two differ only in `applies-to`, so they never compete. I mutated `applies-to` to `bogus-updates` and the validator went red, so the instrument can fail. **Independent confirmation:** GitHub's own config validator reports `.github/dependabot.yml — pass` (`dependabot-api.githubapp.com`) on this head.

**3. 🟠 `chore(deps-dev)` would fail title lint — VERIFIED RESOLVED, by running the validator rather than reasoning about it.** All ten title shapes this config can emit pass `validateCommitSubject`:

```
PASS  chore: bump the root-npm group with 12 updates
PASS  chore: bump the server-npm group in /server with 3 updates
PASS  chore: bump the actions group across 2 directories with 3 updates
PASS  chore: bump the root-npm-security group with 2 updates
PASS  chore: bump the server-npm-security group in /server with 1 update
PASS  chore: bump esbuild from 0.28.0 to 0.28.1 in /server
PASS  chore: bump actions/setup-node from 6 to 7
PASS  chore: bump @typescript-eslint/eslint-plugin from 8.44.1 to 8.45.0 in /server
FAIL  chore(deps-dev): bump the root-npm group with 12 updates      <- the pre-fix shape, still rejected
```

The negative control failing is what makes the positives meaningful. Two supporting checks: `CHORE_PATTERN` (`scripts/validate-commit-msg.mjs:29`) ends in `.+$`, so Dependabot's `Bump`-vs-`bump` capitalisation choice is irrelevant here; and the docs confirm the `(deps)`/`(deps-dev)` suffix is emitted only "when defined" via `include: "scope"`, which this config omits (`:36-37, 53-54, 72-73`). I also checked the 100-char subject cap (`scripts/validate-commit-msg.mjs:35`) — the longest realistic covered title measured 77.

**4. 🟠 `github-actions` missed the composite action — RESOLVED AS FAR AS IT CAN BE VERIFIED PRE-MERGE; one residual below.** `directories` is a real key with documented glob/`*` support (GitHub changelog 2024-06-25), it is schema-valid, and GitHub's own validator accepts it on this head. Both composite actions sit one segment deep — `.github/actions/setup/action.yml` (the `actions/setup-node@v6` + `actions/cache@v5` pins) and `.github/actions/install-ffmpeg-windows/action.yml` (no `uses:` at all) — so `/.github/actions/*` does cover the file the finding was about. See the residual 🟡 for the part I could not prove.

**5. 🟡 Prettier / trailing newline — VERIFIED RESOLVED** for `.github/dependabot.yml`: `prettier --check` with the repo config passes, and `od -c` on the blob confirms the file now ends `- '*'\n`. (But see the new 🟡 below — a *different* file in this PR is not Prettier-clean.)

**Pass 1's 🟡 #2 (silently-omitted checklist steps) — VERIFIED RESOLVED.** The body now carries an explicit "Checklist notes" section waiving steps 1/3/5 by name, and step 2 is genuinely discharged by six new tests.

**6. The new wiring test — IT CAN FAIL. Audited by mutation, not by reading.** `scripts/tests/pr-issue-link-workflow-wiring.test.mjs`. Four mutants, each a plausible regression, each caught:

| Mutation | Result |
|---|---|
| `github.event.pull_request.user.login` → `github.actor` | ✖ caught (`:52`) |
| the whole `env:` block deleted | ✖ caught (`:52`) |
| `"$PR_AUTHOR"` dropped from the `run:` line, `env:` kept | ✖ caught (`:52`) |
| a job-level `if:` added under `lint:` | ✖ caught (`:37`) |

It is also actually *wired*, which is the usual way this class of test dies quietly: `scripts/run-hooks-tests.mjs:10` globs `scripts/tests/*.test.mjs`, and — the part that matters — `.github/workflows/**` is already an input glob on `verify-cache.mjs`'s `test:hooks` step, so a **workflow-only** diff (precisely the diff that would break this wiring) still schedules the test locally and in cloud CI instead of printing `[cached]`. Both `assert.ok(jobMatch/stepMatch)` guards are real anti-vacuity checks. This one is solid.

---

### 🟠 New — the reviewed head is red on a required check; the PR is not mergeable as staged

**Correctness bug in the branch state, not in the diff.** `gh pr checks 2791` at head `e2916bbc`:

```
Lint, typecheck & fast checks   fail
npm run verify                  fail     <- required by ruleset 17654264
```

Five failing tests, all in `scripts/tests/check-register-citations.test.mjs`, none touching anything this PR changed:

```
✖ CLI: without --strict, Check C's exploratory half does not print …
✖ CLI: with --strict, Check C's exploratory half runs …
✖ CLI: Check C's wrongId half is FATAL …
✖ CLI mutation: deleting Check C from fatalSections un-gates a real wrong-ID heading
✖ CLI: the real tree scan finds at least one binary file … (finding U)
ℹ tests 1687
```

I triaged the cause rather than reporting the symptom. Merge commit `e2916bbc` brought in exactly two files (`git diff --stat bfea271b e2916bbc`): `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md` and `docs/superpowers/specs/2026-08-28-onbox-register-generated-surfaces-design.md`. `check-register-citations` scans the **real tree**, so those 3,050 new lines are its input. And PR **#2803** — `fix(ops): use single backticks instead of double backticks in register-citations documentation`, touching that exact plan file — merged to `main` at **09:31:03Z**, eight minutes *after* this PR's CI run started at 09:23. `main` is now at `926a43b1`, which contains that fix; this head does not.

**So: base-side, already fixed on `main`, no code change owed here.** The action is to merge current `origin/main` again and re-push; CI should clear. Two consequences worth stating plainly rather than absorbing:

- the Test plan's "**1687 pass, 0 fail**" does not hold at the reviewed head — CI reports the same 1687 total with 5 failures;
- local pre-push was bypassed with `--no-verify`, so nothing local caught this either. Cloud `verify.yml` is the only gate that saw it.

### 🟡 New — a file this PR adds is not Prettier-clean, re-opening in a different file the exact nit this round just closed

`scripts/tests/validate-pr-issue-link.test.mjs`, six lines over the configured `printWidth: 100`:

| line | length |
|---|---|
| 13 (`const scriptPath = resolve(…)`) | 103 |
| 102, 120 (`writeFileSync(bodyFile, 'Bumps [esbuild]…')`) | 105 |
| 106, 124, 140 (`assert.equal(result.status, …)`) | 101 |

`prettier --check` with the repo's own `.prettierrc` fails on this file and passes on the other four changed files. This is not contradicted by the body's "`npm run lint` — clean, zero warnings": `lint` is `eslint . --max-warnings 0` and `eslint-config-prettier` disables the stylistic rules, so ESLint cannot see it — and `format:check` is still wired into nothing (no step in `verify.yml`, none in `verify-cache.mjs`'s `STEPS`). Cleanup, not a gate failure; the practical cost is identical to pass 1's 🟡 #1 — the next `npm run format` drops a spurious five-hunk diff into an unrelated PR.

### 🟡 New — the config half of this PR ships with zero verification, and its failure mode is silent

Six tests were added for the workflow half; **nothing** pins `.github/dependabot.yml`. That is a gap against this repo's own established pattern for `.github/**` config: `.github/actions/setup/action.yml` has `scripts/tests/setup-action.test.mjs` reading it at runtime plus a dedicated scope glob; `.github/workflows/**` has `workflow-wiring.test.mjs` and now this PR's own wiring test. `dependabot.yml` has no test, no scope glob, and no schema check.

Two concrete scenarios it leaves uncovered, both silent — no red check, no error, just PRs that never arrive:

1. **The `directories` glob may not expand for this ecosystem, and nothing would tell you.** GitHub's validator confirms the file is *valid*; it does not confirm the glob *resolves*. The options reference documents only `directory: "/"` for `github-actions` ("Dependabot will search the `/.github/workflows` directory, as well as the `action.yml/action.yaml` file from the root directory") and says nothing about globbed subdirectories for it; `dependabot-core#10884` (globbed `directories` on `github-actions` → duplicate scans) is still open; and the commonly-cited guidance is to enumerate composite-action directories explicitly — which is what pass 1 recommended (`"/.github/actions/setup"`). With exactly two such directories, one of which has no `uses:` at all, the glob buys nothing over the explicit path and is strictly less certain. **Checkable post-merge:** Insights → Dependency graph → Dependabot lists the directories actually scanned per ecosystem; if `/.github/actions/setup` is absent there, finding 4 is *not* resolved and `actions/setup-node@v6` (`.github/actions/setup/action.yml:31`) stays unbumped exactly as pass 1 described.
2. **A nested composite action is uncovered today.** Add `.github/actions/qa/tools/action.yml` with `uses: actions/checkout@v5`. A single `*` does not cross a path separator, so `/.github/actions/*` (`.github/dependabot.yml:66-68`) never matches `/.github/actions/qa/tools`, that pin is never bumped, and no check notices — the same class this PR set out to close, re-armed one level deeper. A ~15-line test that parses `dependabot.yml`, globs `.github/actions/**/action.y*ml`, and asserts every directory containing a `uses:` is covered by some `directories` entry would fail on that input today.

### 🟡 New — the sole justification for the `--no-verify` bypass is recorded twice, with two different root causes

Commit `bfea271b`'s message says the red leg fails "in THIS worktree only … **because this worktree's `.claude/skills/pr-review-gate` content is stale relative to main** (verified: identical failure with these two files stashed out)". The PR body, on the same leg, says its input is "an ***uncommitted* `reviewer-brief.md` edit in another lane's checkout that was synced into `~/.agents/skills/`**". A stale tracked file in this worktree and an uncommitted edit in a different checkout mirrored into the global agent store are different causes with different implications, and both are asserted as verified. Since that diagnosis is the entire basis for skipping the local gate — and cloud CI then went red on a leg neither account mentions — the record should say which one was actually observed. (The body's secondary claim does check out: the guard skips loudly on a runner with no agent store.)

---

### Verdict

**0 blocking, 1 significant, 3 minor.** Every pass-1 finding is resolved, and the two that mattered most are resolved *well*: the exemption sits at the step so the required context completes green rather than hanging pending, it is keyed on the PR author rather than the actor, it is unspoofable, and it fails closed on empty input — and the test that pins it survives four separate mutations, which is more than most guards in this repo manage.

The one thing standing between this and merge is base-side: head `e2916bbc` predates `main`'s `926a43b1`, so the required `npm run verify` context is red on five `check-register-citations` failures that PR #2803 already fixed. Re-merge `origin/main`, re-push, confirm green. The three 🟡s are cleanup and do not re-trigger a review round.

dudarenok-maker:
## Triage — pass 2 dispositions (head 5e72683d)

All pass-2 items closed. No pass 3: the 🟠 was base-side with no code change owed, and 🟡 cleanup does not re-trigger a review round per the gate's trigger rule.

- **🟠 Head red on a required check** — VERIFIED base-side. PR #2803 merged to `main` 8 minutes after this PR's CI started, fixing the citation checker it tripped. Re-merged `main` (`926a43b1`); no code change made.
- **🟡 `directories` glob** — FIXED (`5e72683d`). Replaced `/.github/actions/*` with the two explicit composite-action paths. Glob support for the `github-actions` ecosystem is undocumented and `dependabot-core#10884` is open on it, so certainty beat brevity. The accepted cost — a new composite action needs a new line — is now enforced by a test rather than left to memory.
- **🟡 Config shipped with zero verification** — FIXED (`5e72683d`). New `scripts/tests/dependabot-config.test.mjs` derives the expected composite-action set from disk rather than hardcoding it, imports the real `validateCommitSubject` rather than reimplementing its regex, and pins that no entry sets `include: scope`. Both key assertions mutation-tested red-then-green.
- **🟡 Prettier on the added test file** — FIXED (`5e72683d`), formatting only, assertions untouched.
- **🟡 Two conflicting `--no-verify` root causes on record** — RESOLVED, and the PR body's version was the correct one. Verified directly: `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` contains an item-11 addition present in **no commit and no working tree**, so the agent store is *ahead* of the repo, not the branch behind it. Commit `bfea271b`'s "worktree stale relative to main" is wrong; that message is immutable now, so this comment is the correction of record.

### One reviewer claim rejected

Pass 2 cited `.github/actions/qa/tools/action.yml` as a nested composite action that a single `*` would miss. **That file does not exist.** `find .github/actions -name 'action.y*ml'` returns exactly two, both one level deep: `setup/` and `install-ffmpeg-windows/`. No entry was added for it. The underlying point still stood on its own and drove the switch to explicit paths.

### Verification at this head

`npm run test:hooks` — **1691 pass / 1692**, the sole failure being the machine-local agent-store guard described above (unrelated, reproduces without these commits, skips in cloud CI where no store exists). `npm run lint` clean. `prettier --check` clean on every file this PR touches.

Issue #2783, opened 2026-08-28
[agent instructions][claude][verify] Verify the ffmpeg-hang hardening (with real-run mutation proof) and open the PR (#2449)


dudarenok-maker:
## What
Verify the ffmpeg-install hardening and aggregator fix built by the two prior children against parent #2449's acceptance — including proving the hang path is actually handled, not just that a healthy run passes — then open the PR.

## Parent's acceptance, restated as a checklist
- [ ] `.github/actions/install-ffmpeg/action.yml` exists, caches the binary, and its install step has a `timeout-minutes` well under every consuming job's smallest budget (10 min for `e2e-visual`).
- [ ] All three inline `Install ffmpeg` steps in `verify.yml` are replaced by `uses: ./.github/actions/install-ffmpeg`, each preserving its original `if:` scope condition.
- [ ] The "Check leg results" aggregator's log output names which legs were cancelled vs. failed vs. skipped, and still fails (exit 1) on any of the three.
- [ ] **The mutation proof, mandatory per the parent issue:** demonstrate the hang path, not just a healthy run.

## Mutation proof — how to do it
1. **Timeout fires:** on a throwaway branch off this one (or a temporary edit reverted after), point the composite action's install step at something guaranteed to hang or fail fast within the 3-minute budget — e.g. temporarily change the apt source to an unreachable/blackholed address, or add a `sleep 200` before the real install command — push it, trigger a `workflow_dispatch` run (or open a scratch PR), and **watch the job actually get killed by the step-level timeout at ~3 minutes**, not the job's full 10-20 minute budget. Capture the run URL and the observed timing. Revert the throwaway change afterward — confirm `git status` clean on the real branch.
2. **Aggregator distinguishes cancelled:** in the same or a follow-up scratch run, confirm the "Check leg results" step's log explicitly names the cancelled job(s) rather than only printing a raw JSON blob. Paste the relevant log excerpt.
3. **Baseline still green:** confirm a normal run of the real branch (no throwaway hang) passes cleanly end to end — this is necessary but NOT sufficient on its own; a change only ever observed on a healthy run is not evidence it handles the hang, per the parent issue's own acceptance language.

A verify receipt that asserts "the timeout would fire" without a real run's timing evidence is not acceptance — re-run it yourself if the prior children's own receipts didn't already include this.

## Where the work is
[...]

Issue #2781, opened 2026-08-28
[agent instructions][claude][task] Cache the ffmpeg install and give it a step-level timeout (#2449)


dudarenok-maker:
## What
Replace the three copy-pasted `apt-get install ffmpeg` steps in `.github/workflows/verify.yml` with one composite action that caches the binary and carries its own step-level timeout, per the decision recorded on parent #2449. This child does NOT touch the "Check leg results" aggregator — that's the next child.

## Background you need
Read parent #2449's decision comment in full first: `gh issue view 2449 --repo dudarenok-maker/Castwright --json body,comments`. Today three jobs each run the identical unguarded step:

```yaml
- name: Install ffmpeg
  run: sudo apt-get update && sudo apt-get install -y ffmpeg
```

at `.github/workflows/verify.yml:286-288` (job `server-tests`, `timeout-minutes: 20`), `:405-407` (job `e2e`, `timeout-minutes: 15`), `:459-461` (job `e2e-visual`, `timeout-minutes: 10`) — re-read those exact line ranges before editing, they may have shifted slightly since this brief was written; find them by searching for `Install ffmpeg` if the line numbers are off. No retry, no mirror fallback, no step-level timeout — a hung `apt-get update` against a slow Ubuntu runner mirror burns the entire job budget and cancels every step after it, including the tests. Measured incident: PR #2424, three jobs cancelled at their full job budget (15m+15m+10m) before a third attempt passed clean.

## The fix
1. Create `.github/actions/install-ffmpeg/action.yml` — a composite action (`using: composite`) that:
   - Uses `actions/cache@v4` (check what cache action version this repo already uses elsewhere, e.g. grep `.github/workflows/*.yml` for `actions/cache`, and match it) to cache the installed ffmpeg binary/package, keyed on runner OS + a pinned ffmpeg version string (don't cache-key on something that silently invalidates every run, like a timestamp).
   - On a cache hit: skip `apt-get` entirely, just restore the binary onto `PATH` (or re-link `/usr/bin/ffmpeg` — pick whichever approach actually works for an `apt`-installed binary; if caching the raw `apt` package output proves awkward, caching via `apt-get`'s own package cache directory, or downloading a static build instead of `apt` at all, are both acceptable — pick the one that's simplest to get right, and say which you chose and why in your receipt).
   - On a cache miss: `sudo apt-get -o Acquire::Retries=3 update && sudo apt-get install -y ffmpeg` (hardened, not just retried on a hunch).
[...]

PR #2796, opened 2026-08-29
ci(ci): harden CI's ffmpeg install against an apt hang (#2449)

### LINKED ISSUE #2449: ops-62 — decide how CI's ffmpeg install should survive an apt hang
Surfaced while merging #2424. A docs-only, nine-line markdown diff was blocked from merging for ~50 minutes across three CI attempts, none of which ran a single test. The branch was never at fault.

### What happens

`verify.yml` installs ffmpeg in three separate jobs with the same unguarded one-liner:

| Job | `timeout-minutes` | ffmpeg step |
|---|---|---|
| `server-tests` — Server tests (fast + slow) | 20 | `.github/workflows/verify.yml:286-288` |
| `e2e` — E2E (chromium) — shard N/4 | 15 | `.github/workflows/verify.yml:402-404` |
| `e2e-visual` — E2E visual baselines (chromium) | 10 | `.github/workflows/verify.yml:456-458` |

All three run:

```yaml
- name: Install ffmpeg
  run: sudo apt-get update && sudo apt-get install -y ffmpeg
```

No retry, no mirror fallback, no step-level timeout. When `apt-get update` hangs against a GitHub Ubuntu runner's mirror, it consumes the job's **entire** budget, the runner kills the job, and every step after it — including the one that actually runs the tests — reports `skipped`.

### Measured, run 32095488149 (PR #2424, head `fdc36dca`)

Attempt 1 — three jobs cancelled, each burning its full budget in the ffmpeg step:

- `E2E shard 1/4` (job 95586057508) — 15m15s
- `E2E shard 3/4` (job 95586057507) — 15m14s
- `E2E visual baselines` (job 95586057591) — 10m16s

Attempt 2 — two cleared, one hung again: `E2E shard 3/4` (job 95588964470) started `Install ffmpeg` 03:44:39Z, cancelled 03:59:40Z — **15m01s**, i.e. exactly the job budget. Attempt 3 passed clean.

Two consequences worth separating:

1. **The run's conclusion is `cancelled`, not `failure`.** `gh pr checks --watch` exits **0** on this, so a watcher that reads the exit code reports "all checks passed" while four checks are red. This bit during the #2424 merge and is a trap for any automation gating on that command.
2. **`npm run verify` goes red as collateral.** Its `Check leg results` step aggregates the legs, so it fails in 3-5s reporting a failure that is entirely upstream — the visible red is on the wrong job.

### The decision owed

Filed as a design pass because there are several defensible fixes and they trade off differently. The choice is the deliverable, not the diff:

1. **Harden apt in place** — `apt-get -o Acquire::Retries=3 update`, and/or a shell retry loop. Smallest change, keeps the current shape, but still pays a slow mirror three times and still has no ceiling.
2. **Give the step its own timeout** — a step-level `timeout-minutes` well under the job's, so a hang fails *fast and legibly* instead of silently eating the budget and cancelling legs that never ran. Fixes the diagnosis problem even if it does not fix the hang; composes with every other option here.
3. **Stop installing it per-run** — cache the binary, pin a static build, or use a `setup-ffmpeg`-style action. Removes the apt dependency entirely, at the cost of a third-party action or a pinned artifact to keep current.
4. **Drop it where it is not needed** — worth confirming all three jobs genuinely need ffmpeg, and whether the e2e audio paths could use a fixture instead of a real binary.
5. **Whether `verify`'s aggregator should distinguish `cancelled` from `failure`**, so an infrastructure cancellation does not present as a test failure.

Options 2 and 5 are about *diagnosis* and 1/3/4 about *prevention*; a good answer probably takes one from each column rather than treating them as alternatives.

### Acceptance

The chosen fix is committed, and — the part that is easy to skip — it is proven against the actual failure mode rather than against a healthy run. A change that has only ever been observed passing on a runner whose mirror was fine is not evidence it handles a hang. Demonstrate the hang path (e.g. by pointing the step at an unreachable mirror, or by asserting the step-level timeout fires) so the guard is known to be able to trigger.

*Benefit (technical):* stops an intermittent apt hang from blocking merges repo-wide, and makes the failure legible when it does happen instead of surfacing as a red `npm run verify` on a job that never ran.

dudarenok-maker:
Closes #2449

## Summary

Replaces the three copy-pasted `sudo apt-get update && sudo apt-get install -y ffmpeg` steps in `.github/workflows/verify.yml` (`server-tests`, `e2e`, `e2e-visual` jobs) with a single composite action, `.github/actions/install-ffmpeg`, per the decision on parent #2449, and makes the `verify` job's "Check leg results" aggregator name which legs were cancelled vs. failed vs. skipped instead of dumping raw JSON.

The old ffmpeg steps had no retry, no mirror fallback, and no step-level timeout, so a hung `apt-get update` against a slow Ubuntu mirror could burn a whole job's timeout budget and cancel every step after it, tests included (measured on PR #2424: three jobs cancelled at 15m/15m/10m before a third attempt passed clean).

The new composite action caches apt's own downloaded `.deb` packages (redirected to a runner-owned `${{ runner.temp }}` dir rather than the root-owned `/var/cache/apt/archives`, which `actions/cache`'s restore step can't reliably write to). On a cache hit it installs straight from the cached `.deb`s via `dpkg -i` — no `apt-get`, no network round trip at all. On a cache miss it runs a hardened `apt-get -o Acquire::Retries=3 update` then `apt-get install -y ffmpeg`.

**Deviation from the parent's literal `timeout-minutes` wording:** composite-action steps do not accept the `timeout-minutes:` key (job-step-only syntax; GitHub rejects the action manifest outright with "Unexpected value 'timeout-minutes'" — caught on this branch's own first CI run, see #2781). Each apt/dpkg call is instead wrapped in `sudo timeout -k 10 180 ...` (coreutils, SIGTERM at 180s + SIGKILL 10s later if it's still alive — up to ~190s per call, ~380s worst-case for the two-call cache-miss path; `sudo` wraps `timeout`, not the other way around, so the kill signal actually reaches the privileged apt-get/dpkg process instead of being swallowed by `sudo` itself). This is a hang-detection **ceiling**, not the normal-path runtime; a non-hanging install is expected to finish well under it. The ceiling itself is *not* "well under" every consuming job's budget — `e2e-visual`'s 10-minute job timeout leaves only roughly 4 minutes of headroom once checkout, `npm ci`, Playwright setup and the visual battery are accounted for, if the ceiling were ever actually hit on that job.

Snapshotting the installed `ffmpeg` binary plus its shared-library closure was the other option on the table but was dropped as fragile — `apt-get install ffmpeg` pulls in a long, version-drifting chain of transitive dependency packages that would need to be re-derived by hand.

This touches `.github/**` plus `scripts/tests/workflow-wiring.test.mjs` (the regression tests locking these fixes in) and `.github/dependabot.yml` (registering the new composite action); no behavior of the consuming jobs themselves (job timeouts, other steps, `if:` scope guards) changed — see the diff.

## Test plan

- [x] `.github/actions/install-ffmpeg/action.yml` exists, caches the apt packages, and every apt/dpkg call is wrapped in a `sudo timeout -k 10 180` hang-detection ceiling (see the budget note above — this is a ceiling, not a claim of comfortable headroom on every consuming job).
- [x] All three inline `Install ffmpeg` steps in `verify.yml` replaced by `uses: ./.github/actions/install-ffmpeg`, each keeping its original `if:` scope condition (`git diff` on `verify.yml` touches only those three lines plus the aggregator step below).
- [x] The "Check leg results" aggregator's log names which legs were cancelled vs. failed vs. skipped, and still fails (exit 1) on any of the three.
- [x] **Baseline still green:** run [33239123740](https://github.com/dudarenok-maker/Castwright/actions/runs/33239123740) (this branch, real diff) — all 12 jobs succeeded, including all three ffmpeg-consuming jobs; `ffmpeg -version` printed `ffmpeg version 6.1.1-3ubuntu5`.
- [x] **Mutation proof — timeout fires (mandatory per #2449 and #2783):** on a throwaway branch off this one (`mutation/2783-hang-proof`, deleted after, `chore/ops-2449-ffmpeg-apt-hang` confirmed clean via `git status` afterward), replaced the install step's real command with `timeout 180 sleep 400` — a guaranteed hang — and dispatched a real run: [33241074367](https://github.com/dudarenok-maker/Castwright/actions/runs/33241074367).
  - `Install ffmpeg` step on `E2E visual baselines (chromium)` ran 07:32:58.4289Z → 07:35:58.4328Z (**exactly 180.023s**), exited **124** (`timeout`'s own signal), and the job as a whole finished in ~3m13s — nowhere near its 10-minute budget. Same ~180s-and-fail pattern on `server-tests` and all four `e2e` shards.
- [x] **Mutation proof — aggregator distinguishes cancelled (mandatory per #2783):** same run's `lint-and-checks` job was separately mutated (`timeout-minutes: 1` + a step that sleeps past it) to force a genuine GitHub-level job cancellation. The aggregator's "Check leg results" log read:
  ```
  One or more verify legs did not succeed:
  CANCELLED (likely CI infra, not a test failure): lint-and-checks
  FAILED: e2e e2e-visual server-tests
  ```
  — naming each leg by job name and bucket, not a raw `toJSON(needs)` dump.

The mandatory PR review gate (3 passes — the loop cap) surfaced several defects in this PR's own commits, fixed in this round rather than deferred:

**Pass 1:**
- The apt cache was never actually being saved (root-owned `lock`/`partial` files left by `sudo apt-get` blocked `actions/cache`'s post-step tar) — fixed and verified via a real cache-hit run.
- The cache-key runner-image-bump invalidation initially interpolated to an empty string (`env.ImageOS`/`env.ImageVersion` aren't populated the way that first attempt assumed) — caught and corrected to read the real OS-level env vars via a `run:` step.
- The "Check leg results" aggregator could fail open (report green) on malformed JSON due to an unguarded command substitution inside a `for` word list — now fails closed, with a `case` default arm added so an unrecognized job result is also treated as a failure rather than silently ignored.
- Two standing regression tests were added to `scripts/tests/workflow-wiring.test.mjs` locking in the timeout wrapping and the cancelled/failed/skipped bucketing (previously validated only by one-time manual mutation proofs on a since-deleted branch).
- `dpkg -i` (used on a cache hit) skips apt's signature verification — documented as an accepted, scoped tradeoff rather than left silent.

**Pass 2** (surfaced re-reviewing the pass-1 fixes above):
- `.github/dependabot.yml` was missing this PR's new `.github/actions/install-ffmpeg` composite-action directory, so `actions/cache`'s own version/security bumps inside it would never have been tracked — registered (required merging `main` in to pick up the file).
- The PR title itself failed `pr-title-lint` (bare `ci:` needs a scope) — corrected.
- Five more inline `sudo apt-get update && sudo apt-get install -y ffmpeg` sites elsewhere in the repo's workflows (`release.yml` ×3, `cross-os.yml`, `quarantine-health.yml`) carried the exact same #2449 hang risk this PR exists to fix — all five switched to the cached composite action, with `release.yml` being the highest-stakes (a hang there blocks the public-beta release publish).
- `timeout -k 10 180 sudo ...` sent its kill signal to `sudo` itself, which doesn't reliably forward it to the privileged child — reordered to `sudo timeout -k 10 180 ...` so the signal reaches `apt-get`/`dpkg` directly.
- The "Resolve runner image identity" step had no `set -u`; an unset `$ImageOS`/`$ImageVersion` would have silently reverted to the original empty-string cache-key bug — now fails loudly instead.
- A regression test's regex used the `s` (dotAll) flag, letting it match an unrelated `exit 1` far below the JSON-validation guard it was meant to check — a mutation removing the guard's own `exit 1` still passed the test. Tightened to only match within the guard's own block, mutation-proved.
- A stale comment cited a CI run as verifying the `-k 10` kill-after flag, but that run predated the flag landing by ~40 hours (it actually verified a different, earlier mechanism) — corrected.

**Pass 3** (final round, loop cap reached):
- A merge conflict with `main` (both branches had rewritten the same `.github/dependabot.yml` comment paragraph) plus 8 commits' worth of drift with no fresh required-status-check results at head — resolved by merging `main` in.
- A **sixth** bare `sudo apt-get update && sudo apt-get install -y ffmpeg` site, in `copilot-setup-steps.yml`, slipped past the pass-2 fix above because it's a two-line `run: |` form rather than the single-line `&&`-joined pattern the pass-2 regression test was written against — fixed, and the guard test hardened: it now scopes to each workflow step's body and checks for `apt-get update` + `apt-get install` + `ffmpeg` co-occurring regardless of line-joining, and no longer false-positives on an unrelated `apt-get install -y <package>` line. Mutation-proved against the exact two-line shape that got through.
- A stale test comment pointed at "the source-regex test below" when the referenced test is actually above it in the file — corrected.
- This section's own wording (previously "five" sites, and an inaccurate "touches only `.github/**`" claim) — corrected above.

## On-box acceptance / release notes

This is CI-infrastructure-only (`.github/**`); no user- or operator-visible behavior changed, so both the on-box acceptance register and the release-notes documents are not applicable.


dudarenok-maker:
## PR review — pass 1 (head 3f150d9d74f9dfb8c1771be45a10287bfa16fd69, depth low)

Reviewed `.github/actions/install-ffmpeg/action.yml` (new) and the `verify.yml` diff, and cross-checked every claim in the PR body against this branch's actual CI run logs and the repo's cache list.

The `timeout` half of this PR is real and proven. The **caching half has never worked, not once**, and its failure is silent. Details below.

---

### 🔴 Critical

**1. The apt cache is never saved. The whole caching mechanism is dead code, and the `dpkg -i` hit path has executed zero times.** *(correctness)*

`.github/actions/install-ffmpeg/action.yml:28-33` (the `actions/cache` step) together with `:62` (`sudo apt-get -o Dir::Cache::Archives="$ARCHIVE_DIR" install -y ffmpeg`).

**Mechanism.** `apt-get` runs under `sudo`, so it creates `$ARCHIVE_DIR/lock` (0640 root:root) and `$ARCHIVE_DIR/partial/` (0700 root:root) inside the directory `actions/cache` is told to archive. The cache action's post-step runs `tar` as the unprivileged `runner` user, cannot read either, and exits 2. `actions/cache` reports that as a **`##[warning]`, not a step failure** — so the job stays green and the log reads normal.

**Evidence, from this branch's own runs (not a hypothetical):**

- Run [33240343148](https://github.com/dudarenok-maker/Castwright/actions/runs/33240343148) — **the head SHA of this PR** — all 6 ffmpeg-consuming jobs (`server-tests`, `e2e` ×4, `e2e-visual`) logged:
  ```
  /usr/bin/tar: ../../_temp/apt-cache-ffmpeg/lock: Cannot open: Permission denied
  /usr/bin/tar: ../../_temp/apt-cache-ffmpeg/partial: Cannot open: Permission denied
  ##[warning]Failed to save: "/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2
  ```
  6 occurrences of `Failed to save`, 6 of 6 jobs. Same in run [33239123740](https://github.com/dudarenok-maker/Castwright/actions/runs/33239123740).
- Consequently, in run 33240343148 every one of those jobs logged `Cache not found for input keys: Linux-ffmpeg-apt-v1` and took the `apt cache miss: installing ffmpeg via apt-get` branch — **31 minutes after** run 33239123740 had already run the identical action on the same branch.
- `gh cache list -R dudarenok-maker/Castwright --limit 100` contains **no `Linux-ffmpeg-apt-v1` entry at all.** The cache has never existed.

**Impact.** Every job still performs the full `apt-get update` + `apt-get install` network round trip — i.e. exactly the hang-prone path this PR exists to avoid. The `timeout` wrapper bounds that hang (good, and proven), but the "no apt-get, no network at all" benefit the PR is built around is not delivered. `timeout 180 sudo dpkg -i "$ARCHIVE_DIR"/*.deb` (`:58`) has never run, so nothing about the hit path — dependency ordering under `dpkg -i`, glob behaviour, whether the closure is complete — is validated either.

**The action's own rationale is falsified by these logs.** The header comment (`:7-11`) says the redirect to `${{ runner.temp }}` exists because `/var/cache/apt/archives` "is root-owned and not reliably writable by `actions/cache`'s restore step." Redirecting does not fix that: it is `sudo apt-get` that populates the new location, so the identical root-ownership problem is reproduced verbatim at the new path. The comment states a reason that the very first run disproved.

**Why the test plan didn't catch it (catalogue #5 — an acceptance criterion blind to its own feature).** The PR body's `[x] Baseline still green: all 12 jobs succeeded ... ffmpeg -version printed 6.1.1-3ubuntu5` is true and proves nothing about caching: it is exactly what you would see with the `actions/cache` step deleted entirely. Nothing in the test plan asserts a cache **hit** ever occurred. The checkbox "`caches the apt packages`" is checked against the presence of the YAML, not against an observed effect.

**Fix.** Drop the root-owned artefacts and hand the `.deb`s to the runner user before the post-step runs — after `:62`:
```bash
sudo rm -rf "$ARCHIVE_DIR/partial" "$ARCHIVE_DIR/lock"
sudo chown -R "$(id -u):$(id -g)" "$ARCHIVE_DIR"
```
**Verification bar for the re-review:** two consecutive runs on this branch, where the second logs `apt cache hit: installing ffmpeg from ...`, a successful `dpkg -i`, and `ffmpeg -version`; plus a `Linux-ffmpeg-apt-*` row in `gh cache list`. "All jobs green" is not evidence here — it was green through both failures above.

---

### 🟠 Medium

**2. The cache key never invalidates on a runner-image bump, and the input that exists to force invalidation is dead wiring.** *(correctness)*

`action.yml:33` — `key: ${{ runner.os }}-ffmpeg-apt-${{ inputs.ffmpeg-version-pin }}` resolves to the constant `Linux-ffmpeg-apt-v1`. There are no `restore-keys`, and `actions/cache` does **not** re-save on an exact-key hit, so once written the entry is frozen. This workflow runs on every PR, so the 7-day unused-cache eviction never fires either: the entry is effectively permanent.

Concretely, once finding 1 is fixed: GitHub rotates the `ubuntu-latest` image (currently 24.04, `ffmpeg 6.1.1-3ubuntu5`) roughly monthly, and across LTS boundaries periodically. The cached `.deb` set then meets a different dependency landscape and `dpkg -i "$ARCHIVE_DIR"/*.deb` either
- (a) exits non-zero on unsatisfiable dependencies — breaking `server-tests`, all four `e2e` shards and `e2e-visual` on **every PR in the repo**, with a log that says "dependency problems", not "stale cache"; or
- (b) succeeds and force-installs the older ffmpeg plus its cached transitive libs over the image's newer ones — `dpkg -i` does not refuse downgrades — silently downgrading shared libraries for the rest of that job.

The documented escape hatch (`:20-23`, "Bump to force a fresh apt-get fetch ... e.g. after an Ubuntu image bump") is unreachable in practice: **none of the three call sites passes `ffmpeg-version-pin`** (`verify.yml:288`, `:407`, `:461` are bare `uses:` with no `with:`), so bumping means editing the action's `default:`, and nothing signals that it is due. That is CLAUDE.md's "a knob that landed without its wiring" shape.

Fold the image identity into the key so an image bump misses naturally — `${{ runner.os }}-${{ env.ImageOS }}-${{ env.ImageVersion }}-ffmpeg-apt-${{ inputs.ffmpeg-version-pin }}` (both vars are set on GitHub-hosted runners) — which makes the pin a genuine override rather than the only line of defence.

**3. The rewritten aggregator fails open: it reports "All verify legs passed" and exits 0 if `jq` or the JSON write fails.** *(correctness)*

`.github/workflows/verify.yml:551-571`, on the job whose `name: npm run verify` is a **required status check on `main`**.

The default shell is `bash -e {0}`. In `for job in $(jq -r 'keys[]' needs.json)`, a non-zero exit from the command substitution in a `for` word list does **not** trigger `errexit` — the loop body simply never runs. All three arrays stay empty, the `if` at `:564` is false, and `:571` prints `All verify legs passed.` with exit 0 **while legs are red**. The per-job `jq` at `:557` has the same shape: on failure `result` is empty, the `case` matches nothing, and that job is counted as a success.

The trigger is `needs.json` not being valid JSON, and `:552` is the exposure: `echo '${{ toJSON(needs) }}' > needs.json` embeds interpolated job outputs inside a single-quoted shell string. Today's outputs (`detect.outputs.scopes`, `.ok`) contain no `'`, so it holds. Add one needed job that exposes a human-readable output — a commit subject, a changed-file list, an error message, all plausible here — and a single apostrophe terminates the quote early, `needs.json` is garbage or the write errors, and the merge gate goes green over failing legs. The same unbalanced quoting is a shell-injection seam on that line.

This is a **regression in robustness against the code it replaced**: the old line evaluated `contains(needs.*.result, 'failure') || ...` as a GitHub expression, resolved by the runner before any shell ran, and could not be defeated by a shell or `jq` problem. Keep the new per-job reporting, but make the gate's polarity safe — e.g. `set -euo pipefail`, assert the parsed key count equals the expected `needs` count, and fail on anything that is not `success` rather than on an enumerated set of failure words.

---

### 🟡 Low / cleanup

**4. `case` at `verify.yml:558-562` has no `*)` arm — deny-list polarity on a merge gate.** Any `result` value that is not exactly `cancelled`/`failure`/`skipped` is silently bucketed as success. GitHub's vocabulary is `{success, failure, cancelled, skipped}` today, so nothing is broken now, and the replaced expression enumerated the same three — but this is the "guard that enumerates syntax" shape, and inverting it (bucket everything `!= success`, with an `OTHER` bucket) costs nothing and closes finding 3's second half at the same time.

**5. `timeout` without `--kill-after` is a soft bound.** `action.yml:58,61,62`. `timeout 180` sends SIGTERM only. `apt-get` installs a SIGTERM handler and will not abandon an in-progress `dpkg` transaction, and a child blocked in uninterruptible I/O on a stalled mirror will not take the signal until it returns — in either case `timeout` waits indefinitely and the step is unbounded again, which is the failure this PR exists to bound. The common case named in the PR (a stalled `apt-get update`) *is* interruptible, and the mutation run proves that path; this is the residual. `timeout -k 30 180` closes it.

**6. The "well under the smallest consuming job's 10-minute budget" claim is wrong for `e2e-visual`.** `action.yml:47-49` and the same sentence in the PR body. `e2e-visual` is `timeout-minutes: 10` (`verify.yml:449`) and also runs checkout, `./.github/actions/setup` (`npm ci`), the Playwright cache/install, and the visual battery. A 6-minute-but-successful apt leaves under 4 minutes for all of that, so the job still hits its cap and is cancelled — the same outcome the PR exists to prevent, with a different proximate cause. Worth either lowering the per-call budget (~90s) or restating the claim honestly. Note the *hang* bound is actually 180s, not 360s, since `set -e` aborts the step when `update` times out and `install` never runs; the comment conflates the two.

**7. No standing regression assertion.** Both mutation proofs were run on `mutation/2783-hang-proof`, which was deleted — nothing in the repo would catch a future edit that removes the `timeout` wrapper, the `uses:` path, or re-breaks the aggregator. `scripts/tests/workflow-wiring.test.mjs` is the established home for exactly this kind of source-regex assertion over `verify.yml`, and per CLAUDE.md's testing discipline new behaviour ships paired tests. Cheap additions in that file's existing style: assert `.github/actions/install-ffmpeg/action.yml` exists; assert every `apt-get`/`dpkg` line in it is `timeout`-prefixed; assert `verify.yml` contains no bare `apt-get install`.

**8. `dpkg -i` on cached `.deb`s does no signature verification.** `action.yml:58`. `apt-get install` validates archives against the apt keyring; `dpkg -i` validates nothing and runs maintainer scripts as root. Scoped honestly: Actions cache entries written on a branch are readable only by that branch and its descendants, and `main`'s entry is writable only via a merge to `main`, so this is not a cross-PR poisoning vector in this repo today. It is still a meaningfully wider trust surface than the code it replaces, and the PR body does not mention the tradeoff. Noting, not blocking.

---

### ✅ Verified clean

- **All three `if:` scope conditions preserved byte-identically.** The diff changes only the `run:` → `uses:` line at each of `verify.yml:288`, `:407`, `:461`; the `if:` lines are unchanged context. They also remain mechanically pinned by `scripts/tests/workflow-wiring.test.mjs`'s `# supports:` / `# leg:` parity test, whose regex keys on the `# supports:` comment, `- name:` and `if:` and is indifferent to `run:` vs `uses:` — so that guard did not go vacuous across this change.
- **The local action resolves at all three call sites** — each job has `actions/checkout@v6` as its first step (`verify.yml:273`, `:398`, `:452`) before the `uses: ./.github/actions/install-ffmpeg` step.
- **`actions/cache@v5` matches the repo's existing pinning** in the same file (`:413`, `:467`).
- **`timeout` exit-code propagation is genuinely proven**, and I confirmed the cited run exists and is what the body says: run [33241074367](https://github.com/dudarenok-maker/Castwright/actions/runs/33241074367) on `mutation/2783-hang-proof`, `Install ffmpeg` on `E2E visual baselines` ran 180.023s and exited 124, job finished ~3m13s against a 10m budget. `timeout`'s 124 propagates through `set -euo pipefail` and fails the step as claimed.
- **The aggregator's bucketing output is real**, matching the log quoted in the body. The rewrite is behaviourally equivalent to the old expression for all four current `result` values — findings 3 and 4 are about the failure modes around it, not about the bucketing itself.
- **`Closes #2449`** present, first line, outside any code span; the `PR issue link` check is green.
- **Release notes / on-box acceptance correctly skipped** — `.github/**`-only, CI-only, no user- or operator-visible delta and no hardware-provable behaviour. The body's "This touches only `.github/**`" carries that, though naming the checklist step explicitly would be tidier.
- **Composite-action manifest is valid** — the `timeout-minutes:` deviation described in the body is accurate (composite steps reject that key) and the coreutils substitute is the right call.

---

**Summary: 1 critical + 2 medium correctness bugs, 5 cleanup findings.** Finding 1 is the blocker — the PR's headline mechanism is inert as written and its own CI proves it. Findings 2 and 3 are correctness bugs that must land in the same round; 2 only becomes live once 1 is fixed, so fixing 1 alone would ship a cache that works exactly until the next runner-image bump.

dudarenok-maker:
## PR review — pass 2 (head 8524d442ac5536659aa8370238e7d38b7b7deb34, depth low)

Re-review of the complete change (whole PR vs. `main`, not just the 8 commits since pass 1). Every pass-1 finding was re-derived independently against the current files and against real CI logs from the head SHA's own run ([33347045737](https://github.com/dudarenok-maker/Castwright/actions/runs/33347045737)) — not from the fix agents' self-reports.

**1. 🔴 apt cache never saved — VERIFIED RESOLVED.** Confirmed from the head-SHA run's `Server tests (fast + slow)` job log, not from the claim: `Cache hit for: Linux-ffmpeg-apt-v1-ubuntu24-20260823.283.1` → `Cache restored successfully` → `apt cache hit: installing ffmpeg from /home/runner/work/_temp/apt-cache-ffmpeg, skipping apt-get` → step succeeded in 6.3s → `ffmpeg version 6.1.1-3ubuntu5`. A hit is only possible if a prior run's post-step actually saved, so the `sudo rm -rf "$ARCHIVE_DIR/lock" "$ARCHIVE_DIR/partial"` + `sudo chown -R` at `.github/actions/install-ffmpeg/action.yml:81-82` did its job. The successful `dpkg -i "$ARCHIVE_DIR"/*.deb` additionally retires the adjacent risk I checked for — that the cleanup or `APT::Keep-Downloaded-Packages` could leave an *empty* cached dir that then wedges every subsequent run. It doesn't: the `.deb`s survive.

**2. 🟠 cache key never invalidated on runner-image bumps — VERIFIED RESOLVED.** Scrutinised hard, given the first attempt (`8bb2dd97`) interpolated to empty strings. The current `Resolve runner image identity` step (`action.yml:30-33`) resolves for real: the same job log shows `key: Linux-ffmpeg-apt-v1-ubuntu24-20260823.283.1`, and the same run's provisioner banner independently reports `Image: ubuntu-24.04` / `Version: 20260823.283.1` — the two agree, so the values are real, not a coincidentally-nonempty string. Ordering is correct and verifiable in the log: `##[start-action ...id=__self_2.image]` … `end-action ... duration_ms=17` completes *before* `##[start-action ...id=__self_2.apt-cache]` begins, so it runs unconditionally ahead of both the hit and miss paths. `id: image` is namespaced to the composite instance (`__self_2.image`), so it cannot collide with any caller-workflow step id. One residual, filed as 🟡 finding 5 below.

**3. 🟠 aggregator could fail open on a jq/JSON failure — VERIFIED RESOLVED, with live proof.** I re-derived the whole reworked step (`.github/workflows/verify.yml:557-598`) rather than checking the one line pass 1 flagged:

- **No unguarded command substitution remains.** Both are guarded: `job_list="$(jq -r 'keys[]' needs.json)" || { … exit 1; }` (line 573) and `result="$(jq -r --arg j "$job" '.[$j].result' needs.json)" || { … exit 1; }` (line 579). For a bare `var=$(cmd)` the assignment's exit status *is* the substitution's, so `||` genuinely catches it — unlike the old `for job in $(jq …)` word-list position.
- **The `[ … ] && echo` lines (593-595) are not a hole.** When a bucket is empty the `[` returns 1, but `set -e` is suppressed for a non-final member of an AND-list, so execution continues to `exit 1`. This is not theory: on this very head run `CANCELLED` was empty, line 593's test failed, and the log still printed `FAILED: lint-and-checks` from line 594 and then `Process completed with exit code 1`. Fail-closed either way — even if `set -e` *had* tripped, the status would still be 1.
- **The `case` classifies every real value without double-counting.** `success) ;;` (587) precedes `*)` (588), so success is matched first and never lands in `FAILED`; `cancelled`/`failure`/`skipped` each have their own arm; a missing `.result` renders as the literal string `null` under `jq -r` and falls to `*)` → `FAILED`. Deny-list polarity is right.
- **Live end-to-end proof on head:** the aggregator printed `One or more verify legs did not succeed:` / `FAILED: lint-and-checks` and exited 1 while every other leg was green. It did not report green.

Pass 1's 🟡 cleanups all landed in code (`-k 10` on all three `timeout`s, corrected budget arithmetic, two source-regex tests, the `dpkg -i` signature note) — but three of them are incomplete or inaccurate as landed; see findings 4, 6, 7, 8.

---

### New findings

#### 🔴 1 (correctness bug) — the new composite action is unregistered in Dependabot, and that is failing the required `Verify` check on this head

`.github/actions/install-ffmpeg/action.yml` (new directory) / `.github/dependabot.yml:76-80` (on `origin/main`; this branch is behind `main` and does not carry the file locally, which is why it looks absent in the worktree).

Not hypothetical — it is red right now. Run 33347045737, job `Lint, typecheck & fast checks`:

```
✖ every composite action on disk is covered by the github-actions directories list
  AssertionError: composite action directory /.github/actions/install-ffmpeg is on disk but
  missing from dependabot.yml's github-actions `directories:` list -- it will not receive
  Dependabot bumps
    at scripts/tests/dependabot-config.test.mjs:90
```

Failure scenario, concretely: `directories:` currently lists exactly `/`, `/.github/actions/setup`, `/.github/actions/install-ffmpeg-windows`. `github-actions` does not honour globs (per that file's own comment, citing dependabot-core#10884), so the `actions/cache@v5` pin at `action.yml:37` — and anything else ever pinned in this action — silently never receives a version or **security** bump. That is precisely the gap #2791 added this test to close. Merging as-is also leaves `main` red.

Fix: add `- '/.github/actions/install-ffmpeg'` to that list. The block's own comment ("This repo also has **TWO** composite actions … `.github/actions/setup/action.yml` … and `.github/actions/install-ffmpeg-windows/action.yml`") is staled by this PR too and needs to say three and name the new one — same round, per CLAUDE.md's report-fix-record rule.

#### 🔴 2 (correctness bug) — the PR title fails `pr-title-lint`, on this head and on all five preceding pushes

Run [33347045642](https://github.com/dudarenok-maker/Castwright/actions/runs/33347045642), job `Validate PR title against commit convention`:

```
Subject:
  ci: harden CI's ffmpeg install against an apt hang (#2449)
...
  chore: <subject>                      no-scope catch-all
```

`ci` is an allowed *type*, but the bare no-scope form is permitted only for `chore`. `scripts/validate-commit-msg.mjs` therefore rejects it. This has been `failure` on every push in the current series (`41c1979d`, `22c93953`, `6b693a16`, `f502c181`, `baade86c`, `8524d442`) — six consecutive red runs — while the branch's own commits all correctly use `ci(ci):`. CLAUDE.md's Branching workflow makes this mandatory ("The PR title MUST match the commit-convention subject format").

Fix: retitle to `ci(ci): harden CI's ffmpeg install against an apt hang (#2449)`. Note the known trap: `pr-title-lint.yml` does not re-run on a title edit alone, so the check must be re-dispatched (or ride the next push) or it stays red at merge time.

#### 🟠 3 (correctness bug) — one instance fixed, the class left armed: five inline `apt-get` ffmpeg installs remain, including all three in `release.yml`

`.github/workflows/release.yml:61`, `:141`, `:273`; `.github/workflows/cross-os.yml:147`; `.github/workflows/quarantine-health.yml:81` — each still `run: sudo apt-get update && sudo apt-get install -y ffmpeg`, i.e. exactly the unbounded, un-retried, uncached step this PR exists to eliminate.

Failure scenario: the incident in #2449 (a slow Ubuntu mirror hanging `apt-get update` until the job's whole budget is consumed and every subsequent step is cancelled) recurs identically on a release tag. `release.yml` is the highest-stakes consumer — CLAUDE.md: "Every release tag now runs COMPLETE cross-platform verification before it publishes" — so a mirror hang there burns the release job and blocks the public-beta publish, with no timeout to fail it fast and no cache to skip the network entirely. The PR hardens the three call sites that were already the *cheapest* to retry (a PR re-run) and leaves the one that is most expensive to retry unhardened.

Fix is a one-line swap at each: all five jobs are `runs-on: ubuntu-latest` and already run `actions/checkout@v6` + `./.github/actions/setup` before the ffmpeg step, so `uses: ./.github/actions/install-ffmpeg` is a drop-in. (`copilot-setup-steps.yml:38-39` is the sixth site; judgement call whether it is in scope, but it is the same shape.)

---

### Cleanup findings

#### 🟡 4 — the new regression assertion for the JSON-validation guard cannot fail for the reason it claims

`scripts/tests/workflow-wiring.test.mjs:169-173`:

```js
assert.match(
  legCheckBody,
  /jq\s+(-e|--exit-status)\s+\.\s+needs\.json.*exit\s+1/s,
  'step does not validate needs.json is valid JSON before processing',
);
```

The `/s` flag lets `.*` cross newlines, so `exit\s+1` binds to any of the *three later* `exit 1`s in the same step (verify.yml:575, 581, 596) rather than to the guard's own. Mutation-proved, not inferred — I replayed the assertion against mutated copies of the real step body:

| mutation | assertion result |
|---|---|
| guard kept but `exit 1` → `echo "::warning::…"` (the exact fail-open pass 1 flagged) | **still passes** ❌ |
| guard block deleted outright | fails ✅ |
| revert to the unguarded `for job in $(jq -r 'keys[]' needs.json)` | still passes ❌ (the sibling `job_list` assertion catches this one ✅) |

So the half of the fix this assertion is named for — that a malformed `needs.json` *exits* rather than merely warning — is unlocked: someone can neuter it and CI stays green. The companion assertion at lines 175-179 is sound and does catch the word-list regression. Fix: drop `/s` and anchor to the guard block (e.g. match `jq -e . needs.json` and the `exit 1` within the same `if … fi`), then re-run the mutation above to confirm it now goes red.

#### 🟡 5 — `Resolve runner image identity` fails open silently if `ImageOS`/`ImageVersion` are ever absent

`.github/actions/install-ffmpeg/action.yml:30-33`. The step runs without `set -u`, so an unset `ImageOS`/`ImageVersion` expands to empty and emits `id=-`, yielding the constant key `Linux-ffmpeg-apt-v1--`. The cache-invalidation guard then reverts to exactly the pre-fix behaviour — never invalidating — with no error, no warning, and a plausible-looking key in the log. This is the same silent-empty-interpolation failure `8bb2dd97` already shipped once on this branch and that was caught only by direct log inspection, so it is a live failure mode here, not a hypothetical. Fix: `set -euo pipefail` in that `run:`, or an explicit `: "${ImageOS:?}" "${ImageVersion:?}"`, so the degradation is loud.

#### 🟡 6 — a comment in the new test asserts a verification that could not have happened

`scripts/tests/workflow-wiring.test.mjs:1083-1085`:

```
// Mutation proof 1: when a hung apt-get was fixed in mutation/2783-hang-proof,
// both the 10s SIGKILL grace period and the 180s timeout itself were verified
// to actually fire by running CI. Assert the wrapping is present.
```

Two defects in three lines. (a) The `mutation/2783-hang-proof` run ([33241074367](https://github.com/dudarenok-maker/Castwright/actions/runs/33241074367)) was created `2026-08-29T07:32:17Z`; `-k 10` was introduced by commit `6b693a16`, dated `2026-08-30T22:50Z` — ~40 hours *later*. The 10s SIGKILL grace period did not exist when that run executed, so it cannot have been "verified to actually fire" by it. (b) "Assert the wrapping is present" describes the *other* test (line 1044); this test asserts `case` arms and the exit condition and contains no timeout assertion at all. Fix: correct the claim to what was actually proven (the 180s SIGTERM ceiling, exit 124 at 180.023s) and move/delete the misplaced sentence.

#### 🟡 7 — `timeout -k 10 180 sudo …` cannot deliver the guarantee its comment claims

`.github/actions/install-ffmpeg/action.yml:52-55` states the wrapper is `-k 10 180` "to guarantee termination even if the process ignores SIGTERM", and lines 76/79/80 wrap `sudo`, not the apt/dpkg process. The `-k` escalation sends **SIGKILL to `sudo`**; SIGKILL is by definition uncatchable and therefore unforwardable, so whether `apt-get` actually dies depends entirely on whether `sudo` `exec()`d it or is holding it under a pty — which the comment's "guarantee" does not survive either way. The canonical form that does guarantee it is `sudo timeout -k 10 180 apt-get …` (timeout runs as root and signals the process directly).

Compounding it: the mutation proof cited for this wrapper replaced the command with `timeout 180 sleep 400` — no `sudo`, no `-k` — so neither the sudo-forwarding path nor the `-k` escalation has ever been exercised, in CI or anywhere. Fix: move `sudo` outside `timeout` at all three call sites and restate the comment to match.

#### 🟡 8 — the PR body still carries the overclaim `f502c181` was dispatched to remove, and declares none of the incidental fixes

Commit `f502c181` corrected the budget arithmetic in `action.yml:56-63` to say the 180s ceilings leave "comparatively tight headroom … ~4 minutes for the test battery itself" on `e2e-visual`. The PR body still says the opposite in two places — "still **well under** the smallest consuming job's 10-minute budget (`e2e-visual`)" and test-plan bullet 1's "**comfortably under** `e2e-visual`'s 10-minute job budget". The fix landed in the code comment and not in the document a reviewer actually reads, so the false figure survives in the more-read half.

Separately, the body still describes only the original change: none of the eight fix commits (cache-permission repair, image-keyed invalidation, the aggregator JSON guard, the `case` default arm, `-k 10`, the two new regression tests, the `dpkg -i` signature note) is declared, and there is no explicit release-notes not-applicable statement. Both are house gates — reviewer brief: "incidental findings fixed in this round rather than filed, and **declared in the PR body**"; CLAUDE.md step 5: skip only with an explicit say-so.

---

### House gates checked

`Closes #2449` present on line 1, outside any code span ✅. No `cast.json` writes, no config knob, no derived artifact, no hardware-provable behaviour (on-box register correctly untouched) ✅. Release-notes pair genuinely not applicable for a CI-only change — but see finding 8 on saying so ✅/🟡.

Tree unmodified: `HEAD` and `git status --porcelain` identical before and after this pass.

dudarenok-maker:
## PR review — pass 3 (head 466156c7340a1d45b20195d28fbf5afdf0b7f8e2, depth low)

Final allowed round (loop cap: initial + 2 re-reviews). Every pass-2 finding was re-derived from the files at head, not read off the commit messages. Mutation probes were re-run independently.

---

| # | Finding | Disposition |
|---|---|---|
| 1 | `dependabot.yml` missing `install-ffmpeg` | ✅ **VERIFIED RESOLVED** |
| 2 | PR title fails `pr-title-lint` | ✅ **RESOLVED in content**, ❌ never re-checked (see F1) |
| 3 | Five more bare `apt-get` ffmpeg sites | ⚠️ **PARTIALLY RESOLVED** — five fixed, a **sixth** was missed and the new guard is blind to it (see F2) |
| 4 | No `set -u` on "Resolve runner image identity" | ✅ **VERIFIED RESOLVED** |
| 5 | `timeout … sudo` sends SIGKILL to `sudo` | ✅ **VERIFIED RESOLVED** |
| 6 | Stale CI-run citation for `-k 10` | ✅ **RESOLVED**, with a small new inaccuracy (see F5) |
| 7 | dotAll regex matched an unrelated `exit 1` | ✅ **VERIFIED RESOLVED** (independently mutation-proved) |

**1 — dependabot.** `.github/dependabot.yml:78` carries `- '/.github/actions/install-ffmpeg'`; the "TWO composite actions" comment now reads "THREE". `node --test scripts/tests/dependabot-config.test.mjs` is 5/5 green locally, including `every composite action on disk is covered by the github-actions directories list`. Resolved — but this exact hunk is now the merge conflict in F1.

**2 — PR title.** `ci(ci): harden CI's ffmpeg install against an apt hang (#2449)` passes the repo's own validator: `node scripts/validate-commit-msg.mjs <title>` → exit 0. The *content* is fixed. The *check* is not recorded — `pr-title-lint.yml` fires only on `opened/synchronize/reopened` (not `edited`, deliberately), and no `synchronize` run has completed since. See F1.

**4 — `set -u`.** `.github/actions/install-ffmpeg/action.yml:34` — `set -euo pipefail` before `echo "id=${ImageOS}-${ImageVersion}" >> "$GITHUB_OUTPUT"`. An unset `ImageOS` now aborts the step instead of producing the empty-string key. Correct.

**5 — `sudo timeout` ordering.** All three call sites are `sudo`-first — `action.yml:78` (`dpkg -i`), `:81` (`apt-get … update`), `:82` (`apt-get … install`). `timeout` now runs as root with `apt-get`/`dpkg` as its direct child, so SIGKILL lands on the privileged process rather than on a `sudo` that will not forward it. The paired regexes at `scripts/tests/workflow-wiring.test.mjs:1064-1066` were updated to the sudo-first form and are non-vacuous (they fail if `sudo` is dropped). Correct.

**7 — the dotAll scoping fix.** Re-derived rather than trusted. I replicated the extraction at `scripts/tests/workflow-wiring.test.mjs:173-185` against the real `verify.yml` and against two mutants:

```
baseline block = "if ! jq -e . needs.json … exit 1\n          fi"   (ends at the guard's OWN fi)

MUT A (guard's `exit 1` -> `echo "::warning::continuing anyway"`):
  /exit\s+1/ on the SCOPED block       -> false   (test correctly FAILS)
  /exit\s+1/ on the whole legCheckBody -> true    (the pre-fix form wrongly PASSED)

MUT B (whole guard block deleted):
  jsonGuardStart === -1  -> the assert.ok precondition fires
```

Also checked the substring risk the `indexOf('fi', start)` approach carries: no `fi` bigram occurs anywhere inside the guard body ("verify" and "refusing" contain `if`/`fu`, not `fi`), so the slice genuinely stops at the guard's own `fi`. Genuinely fixed.

---

### New findings

#### 🔴 F1 — Blocking (correctness bug, merge-readiness): the PR is CONFLICTING and **none of the three required checks has a result at this head**

`gh pr view 2796 --json mergeable,mergeStateStatus` → `"mergeable":"CONFLICTING"`, `"mergeStateStatus":"DIRTY"`.

A read-only test merge names the conflict exactly:

```
$ git merge-tree --write-tree --name-only HEAD origin/main
.github/dependabot.yml
CONFLICT (content): Merge conflict in .github/dependabot.yml
```

`main` advanced to `0a15624d` after this head was pushed (head: 2026-08-31 21:59 +1000; `5dbc6b17`/`7b0c83f1`: 22:09/22:11; `0a15624d`: 2026-09-01 06:34). `5dbc6b17` rewrote the *same comment paragraph* commit `0c7b8b69` edited — main dropped the version numbers (`pins actions/setup-node + actions/cache`) as stale, while this branch rewrote the same lines to "THREE composite actions … `actions/setup-node@v6` + `actions/cache@v5`". Resolving it must keep main's de-pinned wording **and** this branch's THREE/`install-ffmpeg` change; naively taking "ours" re-lands a stale citation main just fixed. `main` also bumped `actions/checkout@v6→v7` and `actions/cache@v5→v6` across `cross-os.yml`, `quarantine-health.yml`, `release.yml`, `verify.yml` and `.github/actions/setup/action.yml` (those auto-merge clean, but the new `install-ffmpeg/action.yml:39` still pins `actions/cache@v5`, which will read as inconsistent post-merge).

The consequence is the more serious half. Status rollup at head:

```
$ gh pr view 2796 --json statusCheckRollup
Check / record CLA signature    COMPLETED    SUCCESS
$ gh api .../commits/466156c7.../status --jq .state
pending
```

That is the **only** check at head. The three required contexts — `npm run verify`, `Verify PR body links a GitHub issue`, `Validate PR title against commit convention` — have **no result at all**. Run history explains why:

```
2026-08-31T12:13Z  CLA Assistant  pull_request_target  466156c7  success
2026-08-31T11:15Z  Verify         workflow_dispatch    f23ac14a  success   <- 2 commits behind head
2026-08-31T01:15Z  Verify         pull_request         8524d442  FAILURE
2026-08-31T01:15Z  PR title lint  pull_request         8524d442  FAILURE
2026-08-31T01:15Z  PR issue link  pull_request         8524d442  success
```

No `pull_request`-event run of `Verify`, `PR title lint` or `PR issue link` has fired since `8524d442` — **eight commits ago** — even though all three list `synchronize`. That is this repo's recorded "a conflicting PR runs ZERO workflows" behaviour. The last recorded PR-event result for two of the three required checks is RED, on a stale SHA.

The only green `Verify` on this branch is a **`workflow_dispatch`** run at `f23ac14a`, which (a) predates `e1bb6462` and `466156c7`, and (b) posts no status to the PR's required context. So pass 2's "verified via a real completed CI run" holds for findings 4 and 5 *at `f23ac14a`*, but **nothing has verified this head**, and branch protection will refuse the merge regardless.

**Fix:** merge/rebase `origin/main`, resolve `dependabot.yml` by hand keeping both sides' intent, push, and let the three required checks actually run and go green before merging.

---

#### 🟠 F2 — Significant (correctness bug): a **sixth** bare-`apt-get` ffmpeg site was missed, and the new regression test is provably blind to the spelling it uses

`.github/workflows/copilot-setup-steps.yml:36-39`:

```yaml
      - name: Install ffmpeg (system dependency)
        run: |
          sudo apt-get update
          sudo apt-get install -y ffmpeg
```

`runs-on: ubuntu-latest` (line 14) with `actions/checkout` at line 20 and `timeout-minutes: 45` at line 17 — i.e. it is a drop-in candidate for `uses: ./.github/actions/install-ffmpeg`, and it carries the identical #2449 exposure: an unretried, unbounded `apt-get update` against a stuck mirror burns the job's 45-minute budget. It is the environment GitHub's Copilot coding agent builds from, so a hang there strands agent runs.

The guard added for pass-2 finding 3 cannot see it. `scripts/tests/workflow-wiring.test.mjs:1097`:

```js
/sudo\s+apt-get\s+update\s*&&\s*sudo\s+apt-get\s+install\s+(-y|.*ffmpeg)/m
```

The `&&` is mandatory. Replicating the test's own scan over the real `.github/workflows/` directory:

```
copilot-setup-steps.yml   guardMatches=false   hasBareAptFfmpeg=TRUE
```

Spelling coverage, measured:

| form | caught? |
|---|---|
| `sudo apt-get update && sudo apt-get install -y ffmpeg` | ✅ yes (guard is not vacuous) |
| the two-line form actually in the repo | ❌ **no** |
| `sudo apt-get update; sudo apt-get install -y ffmpeg` | ❌ no |
| `sudo apt update && sudo apt install -y ffmpeg` | ❌ no |
| `uses: ./.github/actions/install-ffmpeg` (false-positive check) | ✅ correctly not flagged |

So `node --test scripts/tests/workflow-wiring.test.mjs` reports **21/21 green** — including `ffmpeg install: no bare apt-get install ffmpeg in workflows` — while the class it exists to close is still armed in a file it scanned. This is reviewer-brief shape #2 (a guard that enumerates syntax) compounded with #10 (one instance fixed, the class left armed): a green guard over a live defect.

**Fix:** switch `copilot-setup-steps.yml:36-39` to `uses: ./.github/actions/install-ffmpeg`, and rewrite the guard to match on the *commands* rather than on one joining operator — e.g. flag any workflow containing an `apt(-get)? install` that mentions `ffmpeg` at all, regardless of how the two commands are joined. Add the two-line form as a fixture so the new regex is red-before/green-after for the right reason.

---

#### 🟡 F3 — Minor (cleanup): the same guard false-positives on any non-ffmpeg apt install, with misleading remediation

`scripts/tests/workflow-wiring.test.mjs:1097` — the `(-y|.*ffmpeg)` alternation means `sudo apt-get update && sudo apt-get install -y jq` matches (measured: `true`). The failure message then instructs the author to "Replace with `uses: ./.github/actions/install-ffmpeg`", which would install ffmpeg instead of `jq`. Nothing in the repo trips this today, so it is latent — but the next workflow needing an unrelated apt package gets a red test and wrong advice. Fold into F2's rewrite.

#### 🟡 F4 — Minor (stale prose the change made false): the PR body's completeness claims

- "**Five** more inline … sites elsewhere in the repo's workflows (`release.yml` ×3, `cross-os.yml`, `quarantine-health.yml`) … — **all five** switched" — there are six; `copilot-setup-steps.yml` is untouched. Same overclaim in the test's own header comment at `scripts/tests/workflow-wiring.test.mjs:1086-1087` ("ensures **all** workflows use the new … action").
- "This touches only `.github/**`" — the diff also touches `scripts/tests/workflow-wiring.test.mjs`.

Everything else in the body checks out: `Closes #2449` is present and outside a code span; the on-box-acceptance / release-notes not-applicable declaration is explicit and correct for a CI-only change; the "`sudo` wraps `timeout`, not the other way around" description is accurate; the "well under budget" overclaim is genuinely gone, replaced with an honest ceiling framing.

#### 🟡 F5 — Minor: the pass-2 comment fix introduced a smaller inaccuracy of the same kind

`scripts/tests/workflow-wiring.test.mjs:1124-1125` — the corrected comment now reads "The source-regex test **below** verifies the `-k 10` … flag's *presence*". That test (`ffmpeg install: timeout wrapping …`) is at line **1056**, *above* this one at line 1110. The comment also still sits inside `leg-result check: cancelled/failed/skipped bucketing`, which asserts nothing about ffmpeg timeouts. The substantive correction (run `33241074367` verified only the base `timeout 180` SIGTERM path, not `-k 10`) is right and is the part that mattered — the directional reference is not.

---

### What I confirmed is *not* broken

Recorded so the pass distinguishes "checked" from "assumed":

- **All five converted sites keep their scope.** Only the `run:` line changed at each (`verify.yml:300,419,473`; `release.yml:61,141,273`; `cross-os.yml:155`; `quarantine-health.yml:81`); no `if:` was dropped, and `verify.yml`'s three all retain their `if: fromJSON(needs.detect.outputs.scopes)…` line above the new `uses:`. The `setup steps depend on the same scope keys as the leg they support` assertion still binds all 9 `# supports:` markers.
- **Every converted site is Linux with a prior checkout** — all six jobs are `runs-on: ubuntu-latest` with `actions/checkout` earlier in the same job, so no Windows/macOS matrix leg was forced onto this Linux-only action. `cross-os.yml`'s macOS/Windows matrix legs still use brew / `install-ffmpeg-windows` respectively.
- **The `[ … ] && echo` chain at `verify.yml:606-608` does not trip errexit.** Probed under `set -eo pipefail` with `CANCELLED` empty and `FAILED` populated: the `FAILED:` line still prints and the step still exits 1. The diagnostic the change exists to produce survives the common case.
- **The aggregator fails closed** on unparseable `needs.json`, on a `jq` failure, and on an unrecognised or `null` result (default `*)` arm → `FAILED`).
- **Full local suite green:** `node --test scripts/tests/workflow-wiring.test.mjs scripts/tests/dependabot-config.test.mjs` → 21/21 pass. Which is precisely why F2 matters: green here is not evidence.

---

### Verdict — **NOT ready to merge**

Two things stand between this branch and `main`, and neither is discretionary:

1. **F1 is hard-blocking.** The PR conflicts with `main` and its three required status checks have no result at this head; the last PR-event result for two of them was red on a stale SHA. Branch protection will refuse the merge as-is.
2. **F2 is a real correctness bug of exactly the class this PR exists to close** — a sixth armed site, plus a regression test that reads green over it.

F3–F5 are cleanup and fix cheaply alongside.

Per the triage rules an 🟠 correctness finding would normally re-trigger a pass, but **this is round 3 and the loop cap is reached**. Handing back to the shipping session: fix F1–F5, then get a real green `npm run verify` / `Verify PR body links a GitHub issue` / `Validate PR title against commit convention` on the post-merge head before merging. Do not merge on the strength of the `workflow_dispatch` run at `f23ac14a` — it is two commits stale and posts no required status.

*Read-only pass. No tracked file modified: `HEAD` and `git status --porcelain` unchanged (`466156c7…`, clean) before and after.*

Issue #2786, opened 2026-08-28
ops — generate on-box register derived figures; widen wrongId to the ID-reuse class


dudarenok-maker:
Tracking parent for implementing docs/superpowers/specs/2026-08-28-onbox-register-generated-surfaces-design.md via docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md.

**Closes #2362, #2721 on completion. Completes #2599, #2603. Does NOT close #2708** (deferred — see the spec's pass-4 revision note).

Spec + plan land via PR #2785 (docs-only, no code). Once merged, implementation runs as **4 sequential PRs**, one wave at a time — each wave's decomposition happens in its own round, filed only once the prior PR has actually merged to `main` (the plan's own dependency rule: "anything a checker requires of `origin/main` must be on `main` before the code that requires it").

- **PR 1 — data and pins** (plan Tasks 1-2): `.gitattributes` eol=lf pin + marker insertion. Decomposed and filed now, as this issue's first wave (see linked children).
- **PR 2 — the subject map** (plan Task 3): `buildLegitimateSubjectMap` reads bodies. Filed as a follow-on wave once PR 1 merges.
- **PR 3 — the generator and the retirement** (plan Tasks 4-8): the new `scripts/build-register-live-view.mjs`, CI wiring, retiring superseded checks. Filed once PR 2 merges. `refactor` scope, `high` review depth.
- **PR 4 — wrongId, held tasks, corrections** (plan Tasks 9-13): the ID-reuse check, five citation fixes, the wrongId widening, the two held tasks from the predecessor plan (publish-nonce ancestry comparator, citations workflow). Filed once PR 3 merges.

**Sequencing hazard — checked before every wave's first child goes to `Agent Todo`, not just this one:** the `#2759` sweep chain also edits `docs/testing/onbox-acceptance-register.md` and its `.html` twin (via its own `#2768` verify PR). Two lanes editing the register concurrently is the #1931 class this spec exists to end. **Never run this chain and #2759 in parallel** — before flipping any wave's first child to `Agent Todo`, re-check `#2759`'s state; if still open, hold at `Agent Waiting`.

**Every child's brief must instruct: re-derive all `.md`/`.html` line citations, register counts, and function signatures fresh at implementation time** — the plan itself states this repeatedly; the register file changes on nearly every merge to `main`, and both the spec and plan document 4-5 rounds of adversarial review finding stale citations from exactly this drift.

[...]

Issue #2721, opened 2026-08-27
ops-64 — rehome the wrongId-widening deferral that #2629's closure orphans


dudarenok-maker:
**What**

PR #2717 closes #2629 by shipping stable, allocate-once register row IDs. Three deferrals in the tree currently point at #2629 as their open home. When it closes, they point at a closed issue with the work not done — this repo's own citation-rot class, one level up.

The three sites:

- `scripts/check-register-citations.mjs` (~`:26-37`) — Check C's `unknownSubject` bucket is non-fatal and opt-in behind `--strict`, by construction.
- `scripts/check-register-citations.mjs` (~`:209-210`) — widening `wrongId` to the discharge class is "deliberately deferred (see #2629) rather than attempted here, since the two live `unknownSubject` residuals on this corpus are legitimately benign and a naive widening would false-positive on them."
- `docs/testing/onbox-acceptance-register.md` — the allocation-procedure blockquote's description of what does and does not fire.

**The decision that is owed**

Whether `wrongId` should become fatal for the *discharge* class — a citation whose ID still exists but whose original subject has left the register entirely — and if so, how to avoid false-positiving on the benign `unknownSubject` residuals that are the stated reason it was not attempted.

Two defensible outcomes, and something has to choose:

1. **Widen it.** Under stable IDs a discharged row's ID is retired forever, so a citation to it is unambiguously dangling — arguably a stronger signal than it was under positional IDs. Needs a rule that clears the existing benign residuals rather than papering over them.
2. **Leave it non-fatal and say so permanently.** Record it as a deliberate, closed limitation rather than an open deferral, and stop the three sites implying someone is coming back to it.

**Why now**

Under the old positional rule a stale citation silently re-pointed at whatever row inherited the number; under stable IDs it goes visibly dangling instead. That changes the cost/benefit of making this fatal, so the deferral's original reasoning deserves a fresh look rather than being inherited unexamined.

Note a related, verified subtlety (found reviewing #2717): a subject can span several rows — `buildLegitimateSubjectMap` maps one subject to a *set* of IDs — so discharging one row of a multi-row subject does **not** remove the subject, and re-minting that row's ID does produce a fatal `wrongId` today. Any decision here needs to account for both shapes.

**Benefit (technical):** stops three code and doc sites pointing readers at a closed issue as if the question were still tracked, and settles a limitation that stable IDs have materially changed.

dudarenok-maker:
[...]

Issue #2362, opened 2026-08-14
ops — decide what `check:onbox-register` compares beyond counts and row IDs


dudarenok-maker:
**Decision owed:** what should `check:onbox-register` compare between the register and its live view, beyond counts and row IDs?

### What happened

PR #2359 corrected the same wrong sentence three times, in three files, across two review passes. The sequence:

- pass 1 fixed a figure in the spec and run sheet;
- pass 2 found the *input* to that figure was wrong (the Coalfall family is seven language editions, not five) in the run sheet and the live view;
- pass 3 found a **third** copy of that same claim, still saying "five", in `docs/testing/onbox-acceptance-register.md:3179` — while its published twin at `onbox-acceptance-register-live-view.html:1361` had already been corrected to "seven".

For one commit the tracked pair asserted **different facts in the same clause**, and `npm run check:onbox-register` was green throughout.

### Why the guard cannot see it

`check:onbox-register` compares the owed total, the per-group counts and the row IDs. The register's own "Live view" section already documents this edge honestly:

> **A wording-only edit does not fail.** Rewording a row, recording a run result, changing a hardware note or a criteria link — the most common edit this register gets — changes nothing the check compares. The live view mirrors that prose in its own row bodies and will silently fall behind.

So this is a known, documented gap rather than a bug in the script. What is new is evidence of how it fails in practice: not a row drifting *stale*, but the two files stating **contradictory measured figures**, surviving CI, and one of them being the copy published to a public URL.

### Why this is a decision and not a fix

The live view is **hand-authored HTML, not a rendering of the markdown** — that is deliberate and is the whole reason both files are tracked. So prose differs between them by design, and a naive text or figure diff would be red on every honest edit, which trains people to ignore it. Something has to choose what "agreement" means. At least three defensible answers:

1. **Compare extracted numerics per row.** Pull every number appearing in a row body from both files and require the multisets to match. Cheap and mechanical; noisy, because the live view legitimately summarises where the markdown enumerates.
2. **A declared shared-claims block.** Each side carries a small machine-readable set of key/value facts per row (`coalfall_editions: 7`, `unlinkable_share: 84.4`), and the check compares only those. Precise and quiet, but only protects claims someone remembered to declare — the same failure mode one level up.
[...]

PR #2798, opened 2026-08-29
chore(ops): pin the on-box register to eol=lf and insert generator markers

dudarenok-maker:
## Summary
- Pins docs/testing/onbox-acceptance-register.md and its .html twin to eol=lf
- Inserts BEGIN/END GENERATED region markers (strip, glance:<letter> per group) and two A1 stat markers, ahead of PR 3's generator
- No behavioural change -- markers are inert until the generator (a later PR) reads them
- **Validation gap, stated per the plan's own note**: between this PR merging and PR 3 merging, nothing asserts these markers' presence or well-formedness -- no generator and no checker references them yet
- **Merged `main` into this branch** to pick up register-state changes that had landed since this branch's base (owed 61->62, Group A 38->39, new row A104) -- the pass-1 review caught that the unmerged branch would have reverted those figures inside a region now labelled GENERATED. Re-derived both A1/glance/strip marker values against the merged content; both checkers pass on the merged head.
- **Corrected `.gitattributes`' pin comment** (and the identical scaffold text in `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md`): it claimed `scripts/build-register-live-view.mjs` already exists (it lands in a later PR) and that `check-onbox-register.mjs` byte-compares these files (it parses structurally; `checkLiveView` is CRLF-tolerant, contradicting the claim in the same sentence).
- Release notes: skipped -- pure internal/process chore (eol pin + inert markers), no user- or operator-visible effect.

## Test plan
- node scripts/check-onbox-register.mjs -- exit 0 (verified on the merged head)
- node scripts/check-register-citations.mjs -- exit 0 (the stale A40 citation this originally reported was already fixed on `main` by commit 3924fc01; merging main picked that up)
- git diff confirms only marker/comment insertions, the merged register-state update, the .gitattributes wording fix, and the .gitattributes addition -- no unintended content change

Refs #2786, #2362, #2721


dudarenok-maker:
## PR review — pass 1 (head 5bc2138562c98f561c8b5c88e94a6fd95880df88, depth low)

Scope: all three changed files (`.gitattributes`, `docs/testing/onbox-acceptance-register.md`, `docs/testing/onbox-acceptance-register-live-view.html`) · verified: `node scripts/check-onbox-register.mjs` and `node scripts/check-register-citations.mjs` run against extracted copies of the PR tree, against current `main`, and against a reconstructed merge result; `git merge-tree --write-tree origin/main <head>`; `gh pr view --json mergeable,statusCheckRollup`; `ROW_CITATION_REGEX` exercised directly against both versions of the plan line — before adversarially probing whether the markers are inert to the existing checkers and whether the wrapped values are still current.

The marker work itself is correct and genuinely inert (details under "What is solid"). Both blockers are staleness: the branch is four register-state changes behind `main`, and that shows up as a merge conflict, a red required check, and — the part worth care — stale numbers now sealed inside regions labelled `GENERATED`.

### 🔴 Blocking — the branch is stale; keeping its side of the conflict reverts merged register state into a region labelled `GENERATED` *(correctness bug)*

`gh pr view 2798` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `git merge-tree --write-tree origin/main 5bc21385` → `CONFLICT (content): Merge conflict in docs/testing/onbox-acceptance-register-live-view.html` (the `.md` auto-merges).

Since this PR's base `aee0f4ed`, `main` has moved the register: owed **61 → 62**, Group A **38 → 39** (new row `A104`, `next-id` `A103 → A105`), and `main`'s live view was updated to match (`onbox-acceptance-register-live-view.html:198` reads `62`, `:356` reads `39` on `main`). This branch still carries the pre-move values — and this PR has just wrapped exactly those two values in generated-region markers:

- `docs/testing/onbox-acceptance-register-live-view.html:198-199` (PR head) — `<!-- BEGIN GENERATED:strip -->` immediately followed by `<div class="n owed">61</div>`
- `docs/testing/onbox-acceptance-register-live-view.html:355` (PR head) — `<td><!-- BEGIN GENERATED:glance:A -->38<!-- END GENERATED:glance:A --></td>`

Concrete failure scenario: resolve the conflict by keeping the branch's side of those two hunks — the natural instinct, since the branch's entire purpose is "I edited exactly these lines, take mine" — and the merge reverts `main`'s 62/39 while presenting the reverted values as machine-generated content. Reproduced by pairing `main`'s register with this branch's live view and running the checker:

```
- Live view says 61 owed but the register says 62. Update the live view's summary strip.
- Live view: glance table says Group A has 38 rows, the register says 39.
- Live view: Group A's header says 38 rows, the register's body has 39.
- Live view's Group A section is missing row A104 — present in the register, absent from that section.
exit 1
```

It fails loud rather than silent, which bounds the damage — but note the currently-green **"Check on-box acceptance register consistency"** status on this PR is computed on the un-merged head, where the branch's own pair agrees at 61/38. That green does not cover the merge result, so it is not evidence against this finding.

Fix: merge/rebase `origin/main` and **re-derive** the wrapped values — owed → `62`, `glance:A` → `39`, Group A's `<span class="gcount">` → `39 rows`, and add the `A104` row shell to the live view's Group A section. Verified this is sufficient: applying the same marker insertions onto `main`'s current pair gives `check:onbox-register` exit 0 and `check:register-citations` exit 0. The two `.md` stat markers (`40`/`60`) are **not** affected — A1's own heading is unchanged on `main`, so those values are still correct.

### 🔴 Blocking — required CI is red, and the PR body's stated disposition for it is wrong on the facts *(correctness bug)*

`verify.yml`'s **"Lint, typecheck & fast checks"** is `FAILURE` (roll-up **"npm run verify"** likewise). `verify.yml` is a required status check on `main`, so this cannot merge as-is, and the gate's own precondition (before-shipping step 7 / `verify:fast:branch` green) is unmet.

Five failing tests, all real-tree CLI tests in `scripts/tests/check-register-citations.test.mjs`, each asserting exit 0 and getting 1. Root cause proven, not inferred: at base `aee0f4ed`, `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md:57` reads

```
  bare register-ID-shaped token** (e.g. `` row A40 ``). `check-register-citations.mjs`'s
```

The literal `row A40` matches `ROW_CITATION_REGEX` (`scripts/check-register-citations.mjs:286-289`); there is no `### A40` heading in the register; Check A therefore fails fatally. I ran that exact regex against both versions of the line — base matches `["row A40"]`, `main`'s reworded line 58 matches `null`.

The PR body says this is *"pre-existing and unrelated … fixed by a later task in the same plan, not PR1's Task 1/2 scope."* The first half is right; the second is wrong. **It is already fixed on `main`**, by commit `3924fc01` ("docs(docs): reword self-tripping row-citation example in the register-surfaces plan"), and `node scripts/check-register-citations.mjs` exits 0 on current `main`. Nothing is owed to a later task — the same merge that Blocking #1 requires clears this red too.

This matters beyond bookkeeping: as written, the disposition invites a merger to wave through a red *required* check on the belief that no action is available. CLAUDE.md's own triage rule for a pre-existing failure is to surface it, not to scope it out in the PR body.

### 🟠 Significant — the `.gitattributes` comment asserts two things that are not true *(correctness bug, documentation)*

`.gitattributes:58-66` (PR head):

- `:59` — "byte-compared by **scripts/build-register-live-view.mjs**" — that file does not exist at this commit. `git ls-tree -r pr-head scripts/` returns no `build-register-live-view*`; it is Task 4, shipped in PR 3.
- `:64` — "(see scripts/build-register-live-view.mjs's writeFile)" — a reader is directed to open a file that is not there.
- `:59-60` — "byte-compared by … **scripts/check-onbox-register.mjs**" is false independently of PR 3, and stays false after it. That script contains no byte comparison anywhere: `checkRegister` and `checkLiveView` are both structural/regex parsers. The very same sentence then calls `checkLiveView` "the CRLF-tolerant" one — so the only present-day consumer it names is explicitly *not* the naive byte-compare the pin is justified by.

Failure scenario: an agent auditing whether this pin is still needed (or debugging a misbehaving CRLF checkout) opens `.gitattributes`, is told to read `build-register-live-view.mjs`'s `writeFile`, finds no such file, and cannot tell whether the rationale was ever real or the script was since deleted. That is the citation-rot shape this repo already spends effort fighting, and every other pin in this file cites something that exists (`scripts/venv-migration.mjs`, `pinokio-scripts/lib/resolve-release.js`, `schemaBlock()`).

Fix is a comment edit: put the generator in the future tense and name the real present-day consumer accurately — e.g. "*…will be read and rewritten in place by `scripts/build-register-live-view.mjs` (added later in this plan), which always writes LF. `scripts/check-onbox-register.mjs` reads both today and is CRLF-tolerant, so the pin exists for the generator's byte-level round-trip, not for it.*"

Note the wording was copied verbatim from the plan (`docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md:98-107`), so the same correction is owed there — otherwise the next agent re-derives the same false comment from the plan.

### 🟡 Minor

1. **Release-notes step is neither done nor declared** *(cleanup)*. Before-shipping step 5 requires the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair **or an explicit not-applicable**. Neither file is in the diff and the body never states the decision — "No behavioural change" is about the code, not about the step. The skip is almost certainly right (internal chore, no operator-visible effect); it just has to be said out loud.

2. **The linked issues don't include the one this work delivers** *(cleanup)*. Body has `Refs #2362, #2721`. The branch is `chore/ops-2786-register-surfaces-pr1` and the delivering ticket is **#2786** ("ops — generate on-box register derived figures; widen wrongId to the ID-reuse class", OPEN, `area:ops`/`type:chore`). `pr-issue-link.yml` passes because *a* ref exists, so nothing mechanical catches this. Add `Refs #2786`. (The plan's own Step 8 at `:218` says only `Refs #2721`, which is where the omission originates.)

3. **A1's figures now live in two unlinked places** *(cleanup / possible design pass)*. `docs/testing/onbox-acceptance-register.md:708-709` adds `<!-- stat:a1-still-owed 40 -->` / `<!-- stat:a1-subtotal 60 -->`, while `:707`'s heading already states the same two numbers in prose ("**20 of 60 run … · ~40 still owed**"). Per the plan, `parseRegisterFigures` reads only the markers and never cross-checks the heading. Failure scenario: someone runs five more A1 sub-items and updates the heading to "25 of 60 run … ~35 still owed" without touching the markers; PR 3's generator writes the stale `40` into the strip tile, `check:onbox-register` passes (it reads neither surface), and the published page under-reports A1's debt indefinitely. This PR is where the second home is created, so it is the right place to record it — but choosing between "derive from the heading" and "assert the two agree in PR 3's checker" is a real decision, so this may belong in a ticket rather than in this diff.

### ✅ What is solid

Verified by running things, not by reading:

- **Marker grammar matches the plan exactly.** `BEGIN/END GENERATED:strip` around the six tiles, one `glance:<letter>` pair per group letter, `stat:a1-still-owed` / `stat:a1-subtotal` in the `.md` — all per Task 2 Steps 2–4. All **7** letters present in the glance table (A, B, C, D, E, G, H) got a pair; none missed.
- **The markers really are inert to both existing checkers.** `checkLiveView` calls `stripHtmlComments` before any parsing (`scripts/check-onbox-register.mjs:743`), so `<td><!-- BEGIN … -->38<!-- END … --></td>` reduces to `<td>38</td>` and `parseLiveViewGlance` is unaffected; `checkRegister`'s markdown regexes don't match `stat:a1-*` (notably `parseNextIdMarker` is anchored on `next-id:`); and both files are in `check-register-citations.mjs`'s `FROZEN_EXACT` (`:382-390`). Running both checkers over the PR content: exit 0 and exit 0.
- **The eol=lf pin is correctly scoped and ordered.** Both patterns contain `/`, so they anchor at the repo root; they sit after `* text=auto`, and no later pattern in the file matches `.md`/`.html` (the trailing block is binary extensions only). Under `text=auto` the blobs are already LF, so the pin produces no re-normalisation diff — confirmed by the diff being comment insertions only.
- **No content or data corruption.** Every tile value, glance count, group header and row body is byte-identical to base; the `.md` gains two lines and changes nothing else.
- **Prettier can't undo it.** `.prettierignore:20` already excludes the live view (with a comment saying reformatting would break the element-level patterns). The register `.md` is not Prettier-clean on `main` either, and no gate runs `format:check`, so the two inserted comment lines change nothing.
- **A malformed marker degrades loudly.** Dropping a `-->` makes `stripHtmlComments`' lazy match swallow the count, the cell parses as a non-integer, and `parseLiveViewGlance` reports it through `malformedLetters`. The disclosed validation gap is real but narrower than "anything goes until PR 3".

### Verdict

**Do not merge.** Two blockers, both discharged by the same action: merge `origin/main` into the branch and re-derive the values now sealed inside the `strip` and `glance:A` regions (owed → 62, Group A → 39, `gcount` → 39 rows, add the `A104` row shell), which also clears the red `check-register-citations` leg. Then fix the `.gitattributes` comment (and its source in the plan), and fold in the three minors. The marker work underneath is sound and, once rebased, ships cleanly.

dudarenok-maker:
## PR review — pass 2 (head 1193306d1aeca1f4b8db4d9b11abf55238d66e83, depth low)

Deltas only, per the re-review format. Scope: the four-file diff against its merge base (`31cff949`), plus `scripts/check-onbox-register.mjs` and `scripts/tests/check-onbox-register.test.mjs` read as the consumers of what this PR inserts. Verified `gh pr checks 2798` (16/16 green) and re-derived every figure independently rather than reading the PR body's account of them.

**1. 🔴 Stale branch sealed reverted register values inside a GENERATED region — VERIFIED RESOLVED.** Stronger evidence than the checkers alone give: `git diff origin/main 1193306d -- docs/testing/onbox-acceptance-register.md docs/testing/onbox-acceptance-register-live-view.html` returns *only* the marker insertions. Not one numeric or prose byte differs from current `origin/main`, so there is no reverted figure to find. Independently re-derived anyway — register `.md` glance A39/B2/C4/D3/E11/G2/H2 = 63, blocked 5, unconfirmed 2, oldest `2026-06-01`; live-view strip 63/7/5/2/`40 of 60`/`06-01`; live-view glance markers 39/2/4/3/11/2/2. All agree. No conflict-marker text anywhere in the tree at this head (`git grep -E '^(<<<<<<< |=======$|>>>>>>> )' 1193306d` → empty). `origin/main` has advanced to `c7508d4d` since the second merge, but nothing on it touches either register file, so the merge is still current.

**2. 🔴 Required CI red on a stale citation — VERIFIED RESOLVED.** `Lint, typecheck & fast checks` passes (43s), as do all 16 checks including `Check on-box acceptance register consistency`. The A40 citation fix is present via the merge of `3924fc01`.

**3. 🟠 `.gitattributes` falsely claimed the generator exists and that the checker byte-compares — VERIFIED RESOLVED, and both halves of the replacement claim check out against the code rather than being taken on faith.** `scripts/build-register-live-view.mjs` genuinely does not exist at this head (`git ls-tree`), and the only references to it are the two planning docs plus this comment — so "once a later PR lands it" is accurate. "Structural parsing, not a byte-compare" is accurate: `checkLiveView` regex-extracts sections, cells and IDs (`scripts/check-onbox-register.mjs:735-880`, `parseLiveViewGlance:483`, `parseLiveViewSections:536`). "`checkLiveView` is already CRLF-tolerant" is accurate and measured, not asserted: `htmlCellText` collapses `\s+` and trims (`:469-477`), the markdown-side parsers carry explicit `\r?` handling (`:171-176`, `:225`, `:238`), and `scripts/tests/check-onbox-register.test.mjs:2199-2201` records the measurement ("byte-identical results" on CRLF input). The plan's scaffold block (`docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md:98-111`) `diff`s byte-identical to the `.gitattributes` block (`:58-71`) — consistent, not merely similar. No surviving copy of the old false claim: the design spec's two "byte-comparison" mentions argue *against* a naive byte-compare, they do not restate that the checker performs one.

**4. 🟡 Release-notes step neither done nor declared — VERIFIED RESOLVED.** The declaration is present and the reasoning holds: the markers are HTML/markdown comments (invisible in both GitHub rendering and the published page), and the `eol=lf` pin is a checkout-time developer concern. Neither file is touched, consistent with the declaration.

**5. 🟡 `Refs` omitted #2786 — VERIFIED RESOLVED.** Body reads `Refs #2786, #2362, #2721`, plain text outside any code span; all three issues exist and are OPEN; `Verify PR body links a GitHub issue` passes.

**6. 🟡 A1's figures have two unlinked homes — disposition still correct, and better-founded than pass 1 could see.** Two things confirm the "ticket, not this diff" call. First, it is genuinely a design-pass item under CLAUDE.md's one legitimate deferral, and the ticket already exists and is already referenced by this PR: **#2362 — "ops — decide what `check:onbox-register` compares beyond counts and row IDs"** names exactly that decision. Second, the fix is mechanically barred from this PR regardless: `scripts/check-onbox-register.mjs:10-16` records that any tightening of `checkRegister` is retro-applied to `origin/main`'s copy by `resolveBaselineGroups`, so a rule requiring the `stat:a1-*` markers would make every `--against-published` run fail `CANNOT_VERIFY_BASELINE_ERROR` until the data merges first. Correctly deferred. The three homes agree today (heading prose `20 of 60 run … ~40 still owed`, markers `40`/`60`, strip tile `40 (of 60)`).

### 🟠 Significant (NEW) — nothing in this repo can observe the markers this PR exists to ship, and the checker that reads these files is structurally incapable of ever seeing them

`docs/testing/onbox-acceptance-register-live-view.html:198,205,358-364` and `docs/testing/onbox-acceptance-register.md:736-737`

`checkLiveView` calls `stripHtmlComments(rawLiveViewHtml)` as its first act (`scripts/check-onbox-register.mjs:743`, helper at `:457-464`). Every marker this PR inserts is an HTML comment, so the parser sees `<td>39</td>` whether the `glance:A` pair is present, half-present, or gone. `git grep -E 'GENERATED:|stat:a1-' 1193306d` returns hits in exactly four files — the two planning docs and the two data files. **Zero references from `scripts/`.** So this is not the declared "gap until PR 3 lands"; it is a guard that fails open on absent evidence (brief shape 1) sitting on top of the one check that runs (brief shape 6).

The break is not hypothetical, and this PR's own history is the proof. The live view is hand-reconciled on nearly every merge — the plan says so at `:141-143` ("it moves on nearly every merge"), and pass 1's blocking finding was a merge mangling this exact region. Marker-bearing lines were then hand-resolved through **two further merge conflicts inside this PR**: `glance:A` carried `38 → 39` and the owed tile `61 → 62 → 63` across `5bc21385 → feba4ace → 1193306d`, each time by rewriting a line that holds both a value and a marker pair. Had any of those resolutions dropped a `<!-- BEGIN GENERATED:glance:E -->` while keeping `11`, `npm run check:onbox-register` would still report green, CI would still be 16/16, and the loss would surface only in PR 3 — where the plan's own `applyGeneratedRegion` is specified to "throw if the pair is absent" (`plans/2026-08-29-…:407-409`). Three hand-edits of these lines in one PR, zero mechanical verification.

This is closeable here and is *not* blocked by the `resolveBaselineGroups` constraint that legitimately defers finding 6. That constraint binds rules inside `checkRegister`/`checkLiveView`, because those run against `origin/main`'s fetched register and against the already-published page — neither of which has markers yet. A standalone assertion over the **working tree** touches neither. The scaffold is already sitting in the test file: `scripts/tests/check-onbox-register.test.mjs:2202` and `:2206` already load both real files into `REAL_LIVE_VIEW_HTML` and `REAL_REGISTER_TEXT`. Roughly fifteen lines — one balanced `BEGIN/END GENERATED:strip` pair, one `glance:<letter>` pair per letter actually present in the glance table, and both `stat:a1-*` markers — closes the whole window and cannot fire against the baseline or the published page.

Category note for triage: this is a coverage/house-gate finding (CLAUDE.md testing discipline — "if a change lands in untested territory, the test scaffold itself is part of the work"), not wrong behaviour in the shipped bytes. The shipped bytes are correct today; nothing keeps them correct tomorrow.

### 🟡 Minor (NEW) — the plan's Task 2 interface contract claims a `groups` marker pair that this PR did not insert and that the plan elsewhere forbids

`docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md:135-136`

The Interfaces line reads "Produces: three HTML comment region-marker pairs in the `.html` (`strip`, `glance`, `groups`) that Task 4/5/6 locate by exact string match". That is contradicted three times inside the same document: the Global Constraint at `:43-45` ("**except `groups` and the row-shell set, which are located structurally**"), Task 6's own heading at `:818` ("The `groups` target (`gcount` spans)"), and Task 2's own Steps 2–4, which contain no step for a `groups` marker. The design spec agrees (`specs/2026-08-28-…:287-289`: "Two targets are too scattered for regions and are located structurally instead"). The implementation is right and the summary line is wrong.

Concrete break: PR 3's implementer is briefed cold from this plan, and CLAUDE.md makes the brief the contract. Reading `:135-136` as authoritative, it either reports Task 2 as incomplete and re-opens this PR to add a marker pair the design forbids, or wires Task 6 to `applyGeneratedRegion(html, 'groups', …)` — which throws on the absent pair at the first `npm run register:build`. Pre-existing text, but this PR is the execution of Task 2 and already edits this file, so adjacency is not the test.

### Verdict

Pass 1's six findings are all correctly disposed — the two blockers verifiably so, with the register files now byte-identical to `origin/main` outside the marker insertions, and the `.gitattributes` rewrite checked claim-by-claim against the code rather than accepted as written. The one 🟠 above is the finding worth acting on before merge: this PR's entire product is ten HTML comments, in the file the repo documents as conflict-prone, that nothing in `scripts/` can see and that the sole running checker deletes before parsing. The test file already loads both real files; closing it is small and is not blocked by the constraint that legitimately defers finding 6.

dudarenok-maker:
## PR review — triage disposition (head `4e480bb8`)

Pass 2's two new findings, both fixed and pushed (commit `4e480bb8`):

- **🟠 Significant — no test observed the GENERATED markers.** Added a regression test (`scripts/tests/check-onbox-register.test.mjs`) that cross-checks the live-view's embedded `strip`/`glance:<letter>` marker values against the register `.md`'s own "At a glance" table and owed-total prose — an independent computation, not a re-read of the same numbers. This is exactly the class of bug that would have shipped silently: two of this PR's own merge-conflict resolutions hand-edited these marker/value pairs, and `checkLiveView` strips HTML comments before parsing, so it's structurally blind to a dropped or stale marker.
- **🟡 Minor — plan-doc self-contradiction.** Task 2's "Interfaces" section listed `groups` as one of the marker-delimited regions it produces, contradicting the plan's own Global Constraint (`groups` is located structurally, per Task 6) and Task 6's own heading. Corrected the Interfaces bullet to match.

Neither is a correctness bug in the shipped bytes (per the re-review trigger's definition), so this closes without a third review round. Register consistency held through a third `main` merge (the branch was rebased three times total during this review, tracking `main`'s ongoing register-state drift from concurrent work) — `check-onbox-register.mjs`/`check-register-citations.mjs` both exit 0, and all 16 CI checks are green on the current head.

Merging.

Issue #2793, opened 2026-08-29
Audition-centroid severity band false-positives on correctly-cast voice (synthetic-only Phase B pool too tight) — from #2768/A36


dudarenok-maker:
## What

On-box acceptance run for #2768 (register row A36, tied to #2700/#1969/PR #2402)
confirmed the audition-centroid rebuild works for real against a live
Coqui/XTTS sidecar (`referenceKind: "audition"`, real non-degenerate
`cleanMean`/`pSevere`/`pBand`), and correctly flags a genuinely mismatched
voice — but surfaced a real false positive: a fresh, real render of the
CORRECTLY-cast voice, on text outside the audition evidence pool, also
scored `voice-mismatch`/`severe`.

## Root cause

`pSevere`/`pBand` (`server/src/audio/render-integrity/score.ts`) are
percentiles (6th/10th) of the reference pool's OWN cosines-to-centroid. For
a synthetic-only Phase B `auditionCentroid()` pool (`audition-centroid.ts`,
`AUDITION_POOL_TARGET_N=6`) — all renders same engine, same voice, same
controlled acoustic conditions, no real recording variance — that
distribution clusters far tighter than the natural cosine variance of a
genuinely-correct NEW render on different content. Severe/band boundary
landed within ~0.02 of `cleanMean` in the observed run. This is a sharper
version of the already-documented "thin ~0.05-wide over-flag band for the
tightest voices" caveat in `score.ts` (Task 16, tuned against real in-book
anchor sets) — not a new mechanism, but the audition-only pool is both
smaller (N=6) and more homogeneous than any in-book anchor set the
calibration was validated against.

## Observed numbers (2026-08-29, real Coqui/XTTS on RTX 4070 8GB)

- Reference pool (6 real XTTS renders, `Claribel Dervla`): `cleanMean=0.9629`,
  `pSevere=0.9409`, `pBand=0.9446`.
- Fresh correct-voice render #1 (short generic sentence): `cosine=0.928` —
  below `pSevere`, flagged severe.
- Fresh correct-voice render #2 (book-register narrative line, not in the
  evidence pool): `cosine=0.934` — below `pSevere`, flagged severe.
- Genuinely mismatched voice (`Damien Black`) on the same text: `cosine≈0.16-0.18`
  — correctly flagged severe (this direction works).

Full narrative: docs/testing/onbox-acceptance-register.md, row A36, the
"PARTIALLY run 2026-08-29 (claude)" block.

## The ask

[...]

PR #2577, opened 2026-08-21
fix(server): make alignSentences needle-search dash-invariant

dudarenok-maker:
## Summary

Makes `alignSentences`'s needle-search in `server/src/analyzer/dialogue-structure/aligner.ts`
dash-invariant: whether a sentence's cached `text` includes or omits its leading
paragraph-dash marker no longer changes which raw body span it resolves to. Fixes the
desync behind #2537 (14 field diffs on a 1,719/1,940-dash-dense RU chapter).
Closes #2537.

**Final mechanism (search-time dash-prefix preference, not text rewriting):** a needle
that already carries its dash is searched verbatim, unchanged from `main` — `"- да."`
can only occur at a real dialogue marker, so that selectivity is worth keeping. A
needle with no leading dash prefers an occurrence that IS preceded by a paragraph dash
and reports that dash's offset — but only when its own first, bare hit is a false
substring match (mid-word, e.g. `"да."` inside `правда`). A bare hit that already lands
at a genuine word boundary is trusted as-is and never overridden, so a sentence whose
exact text legitimately recurs later in the chapter under a different line's dash isn't
cross-bound onto that unrelated dash. Both old text-rewriting/backward-extension
mechanisms from earlier attempts on this branch were retired; see review history below
for why.

The fix also threads a `dashIsDialogueMarker` gate through all four
`alignSentences`/`locateSentenceOffsets` call sites (#2593/#2594), including the
`annotateSceneBreaks` call site (#2601), so scene-break detection stays consistent
with the dash-dialogue-aware alignment.

## Review history (7 passes, this went through real iteration)

- **Pass 1**: found a blocking regression (unconditional dash-strip). Fixed by
  reverting to `main`'s behavior (a no-op).
- **Pass 2**: found the "fix" was behaviorally a no-op vs `main` (0 diffs over 6,610
  evaluations) — it wasn't actually invariant to anything. Also found the paired test
  was tautological and `locateSentenceOffsets` carried the same defect class unfixed.
  Reworked to a language-gated needle builder.
- **Pass 3**: the rework was still broken — an unconditional strip reintroduced (1,134
  regressions vs `main`), a `---`/`***` scene separator got absorbed into the following
  unrelated narration sentence, and a test pinned this PR's own regression as a
  pre-existing "known residual".
- **Pass 4 (attempt 4, this state)**: retired the text-rewriting needle builder and the
  backward-extension blocks entirely, replaced with the search-time dash-prefix
[...]

PR #2799, opened 2026-08-29
chore: consolidated small-fix sweep (#2759)

### LINKED ISSUE #2496: quarantine-health report misclassifies a 'Not quarantined — still gates' register row
## What

Surfaced by PR #2488's adversarial review (independent Opus pass, finding 4). `parseRegister` in `scripts/quarantine-health.mjs` reads six cells per register row, including cell index 5 (`docs/testing/flaky-register.md`'s "Quarantined" column), but never inspects that cell — every well-formed data row is treated identically regardless of its quarantine state.

The real register has two rows: `#2235` says `Quarantined — routes through \`quarantinedIt\`...`, but `#1981` says `Not quarantined — still gates` (it is tracked here as known-flaky but has NOT been pulled out of the gating suite). Rendered through `aggregate` + `formatReport`, the `#1981` row comes out bucketed like any other — if its re-runs all pass, it's labelled `always-passes` under the legend "candidate to graduate back into the gating suite," which is nonsensical for a test that never left the gating suite.

## Decision owed

Whether — and how — `quarantine-health.mjs` should treat a "Not quarantined — still gates" row differently from a genuinely quarantined one. At least three defensible options, each with different tradeoffs:

1. **Exclude it entirely** from the vitest re-run + report table. Matches the tool's own stated purpose (re-run *quarantined* tests, since `RUN_QUARANTINE=1` is meaningless for a row that was never wrapped in `quarantinedIt`/`quarantinedDescribe`) — but silently drops a row a maintainer might still want visibility into (it's tracked as known-flaky for a reason).
2. **Keep it in the report but relabel its bucket legend/callout** so "candidate to graduate" doesn't apply to a row that already gates — requires deciding new wording and possibly a distinct bucket state.
3. **Keep current behavior** (include it, reuse existing bucket language) on the reasoning that the raw pass/fail signal is still useful even if one legend line reads oddly for this row — cheapest, but ships the reviewer-confirmed misleading language.

This PR's own regression test (`parseRegister returns one entry per data row of the real flaky-register.md (regression)`) currently asserts the `#1981` row IS included, so option 1 would require updating that test's expectations — a behavior change with a real stakeholder (whoever reads the weekly job summary), which is why this is filed rather than picked unilaterally in the fix-agent round.

## Context

- PR: #2488 (fix(scripts): parse the register's real test-cell format and fail loud on a silent zero)
- Register: `docs/testing/flaky-register.md`
- Relevant code: `scripts/quarantine-health.mjs` — `parseRegister`, `aggregate`, `formatReport`

### LINKED ISSUE #2516: srv - generation SSE reconnect retries deterministic precondition failures needlessly
**Scope narrowed 2026-08-20** — see comment below. The original body is preserved for history.

---

**This issue is a decision request, not a task.** Surfaced in passing by the assumption-checker pass on the design doc for #2515 (audio-generation joining the language guard); recorded here per this repo's incidental-findings discipline rather than fixed inline, because it needs a real decision, not just implementation judgment.

## What was found

`src/lib/api.ts`'s generation-stream reconnect logic (`RECONNECT_MAX_ATTEMPTS = 5`, backoff `[500,1000,2000,4000,8000]`) reconnects any stream that saw a tick but no `idle` close. A route-level precondition failure at the top of `server/src/routes/generation.ts` (cast-not-confirmed, book-not-found, provider-selection-failure, and — once #2515 ships — language-unset) produces exactly that shape: one `chapter_failed` tick, then `res.end()`, no `idle`.

These failures are deterministic — retrying the identical request produces the identical failure every time until something outside the stream changes (the user confirms cast, fixes the book, sets the language). The reconnect loop doesn't distinguish this from a transient failure worth retrying (a dropped connection, a momentary sidecar hiccup), so it burns up to 5 extra request round-trips per guarded failure for no chance of success.

## The decision owed

Should the client skip reconnecting when a `chapter_failed` tick carries a stable `errorCode`, on the theory that a classified failure is (by construction) not a "try again and see" case? Or is the right line narrower — only specific codes, or only route-level (`chapterId`-absent) failures? A blanket "any `errorCode` skips reconnect" could be wrong for a code that legitimately *is* transient (e.g. `sidecar-unreachable`, where the sidecar might come back up mid-backoff) — that needs checking against the fs-19 taxonomy's actual semantics, not assumed.

## Not urgent

This is pre-existing behavior (not introduced by #2515's work) and doesn't block that design — the fan-in fix there (per-book suppression in `generation-stream-runner.ts`) is robust to the reconnect churn regardless of whether this is fixed. It's filed because it's the same *shape* of problem #2515 just spent real effort resolving for one code, and is worth deciding generically rather than rediscovering per-code later.

### LINKED ISSUE #2596: chore(ops): resolve-release.js can stamp a Pinokio venv from a stale CRLF checkout
## What

Follow-up from PR #2588 pass-2 review (fix/server-2586-crlf-venv-test). The `.gitattributes` LF pin on `server/tts-sidecar/requirements/*.txt` fixes fresh clones and CI, but does not retroactively fix an *already-existing* checkout — git's `text=auto` clean filter treats an on-disk CRLF file as matching its LF blob (byte-identical after normalization), so `git status` reads clean and neither `git pull` nor `git checkout <ref>` ever rewrites the file. Only deleting it from disk and re-checking-out from the index does.

`pinokio-scripts/lib/resolve-release.js:109` runs `git checkout <tag>` inside a real, persistent repo checkout as the Pinokio install/update mechanism (`pinokio-scripts/install.js`, `pinokio-scripts/update.js`). A Windows Pinokio box that cloned the repo before this PR's `.gitattributes` fix landed — or before some future fix of the same shape — can keep stamping `.venv-stamp.json`/`.req-hash` from CRLF-mangled `requirements/*.txt` bytes indefinitely, reproducing #2586's spurious `pip-in-place` reinstall on every future `resolve-release.js` run, even after upgrading to a release tag that carries the fix.

## Decision owed

Whether/how to add a targeted re-normalization step to `resolve-release.js`'s CLI (not its pure, unit-tested exports) immediately after the `git checkout <tag>` step — e.g. `rm` the files under `server/tts-sidecar/requirements/*.txt` (NOT `git rm --cached`, which unstages them and leaves nothing for a follow-up `git checkout` to restore from — that command was tried and confirmed broken during #2588 pass 3, staging five deletions and fixing nothing) then `git checkout -- server/tts-sidecar/requirements` (or an equivalent delete + re-checkout) — scoped to just the requirements directory rather than a full repo wipe. This needs a design pass because:

- The CLI half of `resolve-release.js` is acceptance-tested only (`pinokio-scripts/lib/resolve-release.test.js` covers just the two pure exports, `latestReleaseTag`/`highestSemverTag`), so a change here has no existing safety net proving it doesn't corrupt or destabilize a live Electron-driven Pinokio install on an end user's box.
- It touches release/install infrastructure with real blast radius (a bug here could break Start/Update for real installs) — this repo's model-routing table treats "irreversible/high-blast-radius decisions" as Premium tier, not a same-round mechanical fix.
- Whether the fix should be generic (re-normalize anything `.gitattributes` pins to `eol=lf`, so it doesn't need updating the next time a new file joins that list) or hardcoded to today's `requirements/` path is itself a design choice.

## Where

- `pinokio-scripts/lib/resolve-release.js` (`main()`, after the `git checkout <tag>` call at line 109)
- `.gitattributes` (comment near the `requirements/*.txt` rule documents this limitation and points here — corrected in #2588 pass 4 to the working `rm` + `git checkout --` recipe, replacing the broken `git rm --cached` one)

## Context

- #2586 (original CRLF venv reqHash bug)
- PR #2588 pass-2 review comment (measured reproduction: `before pull base 56 ... after pull base 56 ... git status --porcelain → clean`, and which remedies work vs. no-op)
- PR #2588 pass-3 review comment (confirmed the `git rm --cached` recipe documented above is broken — reproduced verbatim, errors with "pathspec ... did not match any file(s) known to git" and leaves five files staged for deletion) and pass 4 (fixed the `.gitattributes` wording to the working recipe, re-verified against a reproduced stale-CRLF clone)

### LINKED ISSUE #2608: Dash-invariant needle search: right-boundary not checked for a bare hit (fuzzy AND exact paths)
Found in passing during PR #2577 pass 5 review (dash-invariance fix for #2537/#2540).

`isMidWordHit` (`server/src/analyzer/dialogue-structure/aligner.ts`) gates whether
`findNeedleSpan` trusts a bare needle hit as-is or walks forward to a dash-prefixed
occurrence. It only checks the LEFT boundary of the match — whether the character
immediately before it is a word character. It says nothing about the RIGHT boundary.

For an exact match this is usually fine, since sentence needles end in punctuation.
But `fillRun`'s fuzzy fallback (`aligner.ts`, the `FUZZY_ANCHOR_LEN`-prefix search)
searches only a 16-char PREFIX of the needle text, which by construction usually
does NOT end at a word boundary. A paraphrased dash-led sentence can therefore land
on a prefix match that is a strict prefix of a longer, unrelated word — `isMidWordHit`
passes it (left boundary looks clean), and it's trusted as a genuine match instead of
walking forward to the real dash-prefixed occurrence.

Reviewer's repro: a paraphrased dash-led sentence bound to `narration@0` instead of
`speech@54` via the fuzzy path.

**Not a regression** — this exact gap already existed on `main` before #2537/#2540/#2577,
unrelated to dash-invariance specifically. Filed rather than fixed inline because a
symmetric right-boundary fix isn't a drop-in: the fuzzy path's 16-char truncation point
is inherently mid-word by design, so a naive symmetric check would make the
mid-word-hit condition fire on nearly every fuzzy match, changing established
fuzzy-fallback behavior. **Decision owed:** whether the fuzzy path should (a) extend
its anchor to the needle's own next word/sentence boundary before checking, (b) skip
the mid-word-hit trust-as-is short-circuit entirely for fuzzy hits (always prefer a
dash-prefixed occurrence when one exists), or (c) something else — needs someone to
pick before this can be fixed.

See `aligner.ts`'s `findNeedleSpan`, the "KNOWN RESIDUAL (#2577 pass 5)" comment.

### LINKED ISSUE #2682: chore(sidecar): asr.warm footprint key can never learn past its 128MB seed
## What

On-box acceptance row A31 (formerly A33, `docs/testing/onbox-acceptance-register.md`) asked whether the `asr.warm` `FootprintTable` key's learned p95 converges to something sane once real device-wide-free-memory observations accumulate (#2094's own open question).

Real measurement (wave 7, 2026-08-26): instrumented `FootprintTable.record` directly and drove 15 real `POST /transcribe` calls against a resident `faster-whisper` `base`/`int8_float16` model — 8 with Qwen co-resident (expected to be discarded by the `other_engines` contamination guard) and 7 on a genuinely clean card (nothing else resident, no foreign PID). All 15 recorded `observed_mb <= 0` and were silently dropped by `record()`'s own guard — including all 7 clean samples.

## Why this happens

`asr.warm`'s reservation is measured via a device-wide free-memory *delta* (`PlacementController._device_free_mb`), not the torch allocator's own peak the way every other engine's footprint key is measured. A small model's (`base`/`int8_float16`) actual per-call activation footprint apparently sits at or below this instrument's own noise floor, so the delta essentially never comes back positive — meaning `asr.warm`'s learned estimate can **never** move off its 128MB seed in practice, on any box, contaminated or clean. This isn't "hasn't converged yet" — it's structurally unfalsifiable with the current instrument for this model/precision combination.

## Decision needed (design pass owed)

More than one defensible fix exists, so this needs a decision rather than a default fix:
1. **Leave it** — the 128MB seed is a permanent floor; harmless if it's already a generous/safe reservation for this model.
2. **Switch instruments** — measure `asr.warm` via the torch allocator's own peak (`_observed_mb`), the way every other engine's footprint key is measured, instead of the device-wide free-memory delta.
3. **Accept and document** — record in the seed comment / `docs/local-llm.md` footprint table that this key is unfalsifiable for small ASR models, so a future reader doesn't re-litigate it.

## Where

- `server/tts-sidecar/main.py`: `FootprintTable.record` (`observed_mb <= 0` guard), `SEED_FOOTPRINTS_MB["asr.warm"]`, the ASR-specific delta measurement in the reservation's `finally` block.
- Criteria/history: `docs/testing/onbox-acceptance-register.md` row A31 (was A33), #2094.

### LINKED ISSUE #2700: Voice reassignment: validate a rebuilt audition centroid actually scores correctly (not just discard-on-reassign)
## What

The on-box acceptance test for #1969/PR #2402 (voice reassignment vs. persisted
audition centroid, formerly register row A34) confirmed that reassigning a
character's voice correctly discards the stale old-voice `audition` reference
rather than silently reusing it (`matchesCurrentVoice()` returns false, falls
to a genuine `auditionCentroid()` rebuild attempt). See
`docs/testing/onbox-acceptance-register.md`'s wave-9 changelog entry for the
full trace.

What it did **not** confirm: whether a *successfully rebuilt* centroid (i.e.
one where the new voice's synthesized sample clears `MIN_DURATION_SEC` and
`auditionCentroid()` returns a real `resolved` outcome, not `too-short`)
actually scores correctly against real audio — flags a genuine mismatch and
clears a genuine match. The only on-box run available (Ivo, cloned voice) hit
`too-short` because the sample was too brief, so the rebuild path was
exercised but never reached a scored comparison.

## Acceptance

Reassign a thin-on-anchors character to a new voice with enough
anchor-eligible audio that `auditionCentroid()` produces a real `resolved`
outcome (not `too-short`) for the new voice, then confirm:
1. The new centroid's `cleanMean`/`pSevere`/`pBand` are real (non-zero) values
   distinct from the old voice's.
2. Lines from a genuinely mismatched voice against that new centroid still
   trigger `voice-mismatch`/`severe`.
3. Lines from the correctly-assigned new voice do not.

## Context

Net-new follow-up surfaced during on-box acceptance wave 9
(docs/testing/onbox-acceptance-register.md). Not a defect in PR #2402 — the
discard-stale-reference behavior it verified is real and correct — this is a
narrower, unproven half of the same row's original criteria.

### LINKED ISSUE #2747: spawn-windows-hide.test.ts: blankCommentsAndStrings() can hide a real unguarded spawn after a regex literal with an unpaired quote/backtick
## Chore — design decision owed

An independent PR review pass on PR #2716 found and reproduced a live blind spot in `blankCommentsAndStrings()` (`server/src/spawn-windows-hide.test.ts`), the lexer both `scanFile()` and `externalFilesFloor()`'s content filter use to blank out comments/string content before matching spawn calls.

## The bug

`blankCommentsAndStrings()` walks the source character-by-character, tracking whether it's inside a `'`, `"`, or `` ` `` string. It does not parse regex literals at all. If a regex literal in the source contains an unpaired `` ` `` or quote character (e.g. `` /^`+|`+$/g ``), the lexer misreads that character as the START of a string literal — and from that point on, everything in the file (including real code, including a genuine unguarded `child_process` spawn call) is treated as "inside a string" and blanked out, making it invisible to `CALL_RE`. The scan reports the file as compliant even if it contains a real windowsHide violation after that point.

This was reproduced live against `scripts/lib/knob-docs.mjs`, which contains a regex literal (`` /^`+|`+$/g ``) that desyncs the lexer's quote-tracking state at its character offset — everything after that point in the file is blanked.

This is a **pre-existing** property of `blankCommentsAndStrings()` (not introduced by PR #2716), but it was newly surfaced during that PR's review, and the PR's own cleanup (removing the now-dead `pipSpawners()`/`PIP_SPAWN_RE`) deleted the file's only nearby prose documenting a *comparable but safe-direction* lexer limitation (the old `blankComments()` function used by `pipSpawners()` had the same regex-literal blind spot, but in a direction that could only over-select files, never hide a real violation — see the removed comment in `d17cdee1`/`ecab74e0`'s history). Nothing documented the more dangerous direction for the lexer actually in production use. PR #2716 restores documentation of this at the function itself (see `server/src/spawn-windows-hide.test.ts`, the `KNOWN LIMITATION` comment above `blankCommentsAndStrings()`), but does not fix the underlying gap.

## Why this wasn't just fixed

Correctly distinguishing a regex literal from a division operator in JavaScript/TypeScript source requires real lexing/parsing (the same class of problem that makes hand-rolled JS tokenizers notoriously easy to get subtly wrong) — not a quick patch to the current single-pass character walker. This is a design-level decision about how much investment this guard's own tooling deserves, not a bug with one obvious fix.

## Decision owed

Whether and how to close this gap. Options to weigh, not a mandate:
1. **Use a real (minimal) JS tokenizer** for `blankCommentsAndStrings()` — e.g. a small hand-rolled state machine that actually tracks regex-literal-vs-division-operator context (the standard heuristic: a `/` starts a regex literal only when the previous non-whitespace token can't end an expression — after `(`, `,`, `=`, a keyword, etc. — and a division operator otherwise), or reach for an existing lightweight JS lexer dependency if one is already in the dependency tree.
2. **Accept the risk, keep it documented** — the blind spot requires a specifically-crafted regex literal to trigger, and every currently-scanned file has been checked; re-check periodically or leave as a known, documented limitation.
3. **A narrower structural mitigation** — e.g. detect (without fully parsing) when a `/.../ ` sequence looks like a regex literal with a backtick/quote inside it and refuse to silently blank past it (fail loud / flag for manual review) rather than trying to lex it correctly.

## Key files

- `server/src/spawn-windows-hide.test.ts` — `blankCommentsAndStrings()`, used by both `scanFile()` (the primary `server/src/**` guard) and `externalFilesFloor()`'s content filter (the external-files scan).
- `scripts/lib/knob-docs.mjs` — the file used to reproduce the bug live (character offset ~6259, the `` /^`+|`+$/g `` regex literal).

Found during PR review of #2716; see that PR's pass-2 review comment for the original finding.

### LINKED ISSUE #2759: chore: consolidated small-fix sweep — #2608, #2496, #2516, #2682, #2747, #2596, #2700
## What

Tracking issue bundling a sweep of 6 previously-filed "decision owed" incidental
findings (now resolved, see below) plus one on-box acceptance verification, into
one branch / one PR. Each sub-issue is fixed by its own `[agent instructions]`
child; see the sub-issues list on this issue.

**Re-pointed 2026-08-28: #2582 removed.** It was already closed by
already-merged PR #2719 before this issue was filed. #2582's would-be child
(#2765) was verified as a duplicate of that shipped work and closed as `not
planned` — see #2765's closing comment for the evidence (real
`InferenceSession` + `.get_providers()` check already at `main.py:3240-3292`,
plus existing `test_kokoro.py` coverage). The chain (#2764 → #2766) and the
final verify child (#2769) were both re-pointed to skip it.

## Decisions resolved (2026-08-28)

- **#2608** — fuzzy fallback: when the left-boundary check would trust a bare
  hit, search forward for a dash-prefixed occurrence and prefer it when one
  exists, instead of trusting the bare hit as-is.
- **#2596** — hardcode the re-normalize step to
  `server/tts-sidecar/requirements/*.txt` (not a generic `.gitattributes`-driven
  mechanism). ⚠️ Flagged for `high` PR-review depth regardless of its
  single-scope chore label — the issue itself calls this high-blast-radius
  release/install infrastructure.
- **#2496** — exclude any row whose "Quarantined" column doesn't start with
  "Quarantined" from the report entirely; update the existing regression test.
- **#2516** — scoped allowlist (cast-not-confirmed, book-not-found,
  provider-selection-failure, language-unset once #2515 ships) skips reconnect;
  every other errorCode keeps existing behavior.
- ~~**#2582** — startup-only verification (real `InferenceSession` construction)
  + warn-and-continue (log only, no hard-fail, no new health endpoint).~~
  **Removed — already shipped via PR #2719, see the note above.**
- **#2682** — switch `asr.warm`'s measurement to the torch allocator's own peak,
  matching every other engine's footprint key.
- **#2747** — narrow structural mitigation: detect an ambiguous
  regex-literal-with-unpaired-quote and fail loud instead of silently
  mis-scanning. No full tokenizer.
- **#2700** — not a decision; a real on-box acceptance run (voice reassignment
  centroid scoring). Its own child fixes any bug it finds, or records the
  register-row outcome if none is found.

## Branch / worktree

One branch for the whole chain: `chore/sweep-2NNN` (see this issue's number).
All 9 children commit to this one branch; only the final verify child opens
the PR.

## Not in scope

No other pre-existing "decision owed" issue beyond the 7 named above (6
decisions + 1 on-box run, #2582 removed — see the note above). No new
features. No refactor of adjacent code beyond what each fix requires.

dudarenok-maker:
## Summary

Consolidated small-fix sweep tracked by #2759: closes out 7 remaining
findings surfaced during PR review/on-box acceptance passes, each fixed by
its own child issue on this branch (`chore/server-sweep-2759`):

- #2608 — fuzzy-path right-boundary gap in dash-invariant needle search
  (`aligner.ts`). A bare hit that was a strict prefix of a longer, unrelated
  word passed the left-only boundary check; now the fuzzy fallback also
  checks the right boundary and walks forward to a dash-prefixed occurrence
  when the bare hit is ambiguous. Exact-match path is unchanged, per #2759's
  resolved decision.
- #2496 — `quarantine-health.mjs`'s `parseRegister` now skips any row whose
  Quarantined cell doesn't literally start with "Quarantined", so a
  still-gating row like #1981 is no longer re-run or reported on by the
  quarantine-lane tool.
- #2516 — regression test only (no production fix needed): `fa5d0923`
  already closes the SSE stream with `idle` right after `chapter_failed` on
  every deterministic precondition bail-out, and the client's reconnect gate
  already treats any idle-terminated stream as drained. Adds a test pinning
  that behavior plus a contrast test proving a transient failure (no idle)
  still reconnects unchanged.
- #2747 — `blankCommentsAndStrings()` (the spawn-`windowsHide` guard) now
  recognizes a same-line regex literal and fails loud instead of silently
  desyncing its quote-tracking when the literal's body contains an unpaired
  quote/backtick. Fixes five live ambiguous-regex sites this newly catches.
- #2682 — `asr.warm`'s footprint is now measured via the torch allocator's
  own peak, matching every other engine's footprint key, instead of a
  device-wide free-memory delta that was structurally unfalsifiable for this
  model/precision combination.
- #2596 — **flagged for `high`-depth review: touches release/install
  infrastructure (`resolve-release.js`) with real blast radius for live
  Pinokio installs.** Adds `renormalizeRequirementsCrlf()`, called right
  after `git checkout <tag>`, to force a fresh re-checkout of
  `server/tts-sidecar/requirements/*.txt` from the index — fixing the case
  where `text=auto`'s clean filter makes an already-CRLF-mangled file on disk
  look unchanged to git, so a plain checkout never rewrites it even after
  upgrading past the `.gitattributes` `eol=lf` fix.
- #2700 — on-box acceptance run recorded in
  `docs/testing/onbox-acceptance-register.md` (row A36): confirms the
  audition-centroid rebuild produces a genuine `resolved` reference with
  real, non-degenerate `cleanMean`/`pSevere`/`pBand` values, and correctly
  flags a genuinely mismatched voice. Surfaced a new false-positive defect
  on the correctly-cast voice against a small synthetic-only pool — tracked
  separately as #2793 (a calibration/design decision, not a mechanical fix);
  this row stays partially-run pending that decision.

`npm run verify:fast:branch` is green (fully cached from source, `EXIT=0`).

Closes #2608
Closes #2496
Closes #2516
Closes #2682
Closes #2747
Closes #2596
Closes #2700
Closes #2759

## Test plan

- [x] `npm run verify:fast:branch` — green (lint, typecheck, test, test:server, test:sidecar, build all cached/clean, `EXIT=0`)
- [x] Each child issue's `AGENT DONE` receipt includes real mutation-test output (revert-the-fix → red, restore → green) for its own fix
- [x] #2700's on-box row is a real, unmocked Coqui/XTTS run on resident hardware (RTX 4070) — see `docs/testing/onbox-acceptance-register.md` row A36 for full detail and the follow-up (#2793)
- [ ] Reviewer: give #2596's `resolve-release.js` change (`pinokio-scripts/lib/resolve-release.js`) a `high`-depth pass — no existing test safety net for the CLI half beyond the new acceptance test added here


dudarenok-maker:
## PR review — pass 1 (head f2e594ed9141c6f680ba97543c74d651993b8c8e, depth high)

Scope: all 17 changed files across `server/src/**`, `server/tts-sidecar/**`, `scripts/**`, `pinokio-scripts/**`, `src/lib/**`, `docs/**` · verified: the two named-red mutation claims by re-running the real code from this head's blobs (aligner `findNeedleSpan` with both flag values; `blankCommentsAndStrings` against a `git archive` of this head's `server/src` + `scripts`, 1170 files), plus `git checkout -- <dir>` stream/exit behaviour in a scratch repo, before adversarially probing each of the seven fixes.

### 🔴 Blocking — #2747: the fail-loud regex detector is inert on every real source file in the repo

`server/src/spawn-windows-hide.test.ts:56` — `looksLikeRegexOpen(out, i)` reads `out[i-1]`, which is only meaningful while `out` mirrors `src` position-for-position. `blankCommentsAndStrings` breaks that invariant in three places by pushing a **two-character element** while advancing `i` by 2: line 362 (`/*` open), line 366 (`*/` close), line 395 (a `\` escape inside a string literal). Each costs one element of alignment. Once `out.length < i`, `out[i-1]` is `undefined` — neither `' '`/`'\t'` nor a member of `REGEX_POSITION_PRECEDERS` — so `looksLikeRegexOpen` returns `false` for **every subsequent `/` in that file** and the new regex branch never runs again.

Concrete repro, using this head's own blob and the PR's own passing fixture as input:

```
blankCommentsAndStrings("const RE = /'/g;\nspawn('ffmpeg', args, { windowsHide: true });\n")
  -> throws  (this is the shipped test)

blankCommentsAndStrings("/* a block comment */\n" + <same string>)
  -> does NOT throw

blankCommentsAndStrings("const p = 'a\\b';\n" + <same string>)
  -> does NOT throw
```

And on the file the fix was built around — re-inserting `const AMBIG = /'/g;` immediately after the imports of this head's `server/src/export/build-m4b.ts` (the most favourable possible position for the detector):

```
re-introduced /'/g into real build-m4b.ts -> threw: false
same snippet in isolation                 -> threw: true
```

Scale, over `git archive`'d PR-head `server/src/**/*.ts` + `scripts/**/*.mjs` (1170 files): the shipped detector flags **1** file; the identical code with the alignment corrected (`out.push(' '); out.push(' ');`) flags **56**. Genuine unescaped survivors it cannot see include `server/src/tts/text-normalize.ts:51` (`/([A-Z])([A-Z']{2,})/g` — one `'`) and `server/src/workspace/paths.ts:143` (`/[/\\:*?"<>|\x00]/g` — one `"`).

Every file this guard scans opens with a block comment, including the guard's own. So the "FAIL LOUD" guarantee that the narrowed KNOWN LIMITATION comment and the PR body both assert does not exist in production — the five hand-escaped sites are fixed, but nothing stops the sixth. The four new fixture tests pass only because their fixtures happen to contain neither a block comment nor a string escape; that is the "test that cannot fail" shape, and it is what let the real behaviour through.

### 🔴 Blocking — #2608: the right-boundary check mislabels correct hits, handing the sentence to an unrelated dash line

`server/src/analyzer/dialogue-structure/aligner.ts:361`. On the fuzzy path `search` is `needle.text.slice(0, FUZZY_ANCHOR_LEN)` — an arbitrary fixed-length cut — so "ends mid-word on the right" is an artifact of *where the cut fell*, not evidence that the hit is false. The new doc comment concedes this outright ("by construction, almost never ends at a real word boundary"). The predicate therefore cannot tell the shipped test's decoy apart from the case where the sentence's **correct** location is itself mid-word on the right, and in that case it unlocks the forward walk that lines 336–343 say is only safe "when this hit is known-false" — reopening the #2577 pass-4 "Q1" hazard by binding to a different sentence's dash.

Repro — the shipped test's own fixture shape with decoy and truth swapped, run against `findNeedleSpan` from this head's blob:

```
body   = "— Простите. Он долго смотрела в окно, ожидая её возвращения."
       + <one line of filler>
       + "— Он долго смотрел на дорогу, — сказал Виктор."
needle = "Он долго смотрела в окно, мечтая о её скором приезде,"   (exact match fails -> fuzzy)
search = "он долго смотрел"                                        (16-char prefix)

checkRightBoundary:false (pre-#2608)  -> start 12  -> "он долго смотрела в окно, ожидая ее возвращени"   CORRECT
checkRightBoundary:true  (post-#2608) -> start 136 -> "- он долго смотрел на дорогу, - сказал виктор."   WRONG
```

The pre-fix answer is independently confirmed end-to-end: running the same fixture through `alignSentences` on the working tree (currently `main`) binds to the correct home. The sentence's true home is not dash-adjacent, its 16-char prefix ends mid-word there for no reason other than the cut, and a later unrelated dash-led line shares that prefix — none of which is exotic in dash-dialogue prose, where repeated dialogue openers are the norm.

This is live on the primary analyzer path: `server/src/routes/analysis.ts:2342`, plus `dialogue-structure/escalation.ts:213`, `dialogue-structure/evidence.ts:92` and `store/attribution-health.ts:173`, all pass `conventions.dialogueOpen !== null` — true for Russian. No test covers this direction; the only new test covers the direction the fix was designed for.

Two doc surfaces now state the invariant this change breaks, and were not updated: `aligner.ts`'s `Needle` doc comment (~lines 187–199) — "A bare hit that already lands at a genuine word boundary is an independently valid occurrence and is trusted as-is, with no forward walk" — and `locateSentenceOffsets`'s doc (~line 599), which still points readers at it "for … the #2577 'Q1' carve-out".

### 🟠 Significant — house gate: the release-notes pair is missing, and this PR makes two existing entries false

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched, and the PR body neither adds an entry nor states an explicit not-applicable (checklist step 5 requires one or the other). #2596 alone is a user-visible install fix in the open v1.15.0 cycle. Two entries already in the file are now wrong because of this PR:

- `docs/release-notes-next.md:528` — "The `resolve-release.js` remediation itself stays filed as a follow-up (#2596, corrected to the same working recipe) rather than fixed in this PR". This PR fixes it.
- `docs/release-notes-next.md:540` — "a file whose spawn call sits after a regex literal containing an unpaired backtick or quote can still be missed by the shared lexer (#2747, tracked separately, not a regression introduced here)". The PR claims to close that. (Given 🔴 #1 the sentence is accidentally still true, but not for the reason it gives.)

### 🟠 Significant — house gate: on-box acceptance not recorded for either hardware/real-install change, and the live view was not moved

- **#2596** changes what happens on a real Pinokio Install/Update — provable only on a real install, which is exactly what rows `E1` (ops-16 Pinokio installer) and `E7` (the Pinokio update path) exist for. No row filed.
- **#2682** changes how `asr.warm` footprints are learned, and the replacement mechanism is explicitly stated as unmeasured: "on the theory that a resident forward's OWN torch-side activity … is enough for the allocator peak to see". The code it replaces asserted the opposite as established fact — "a warm ASR forward measured via `_observed_mb` reads ~0 (dropped by `record()`'s `<= 0` guard) and never accumulated real `asr.warm` samples". Both cannot be true. The new pytest proves the *routing* (it stubs `_observed_mb` to 77) but by construction cannot prove a real CTranslate2-backed forward yields a positive torch-allocator peak. If the removed assertion was right, `asr.warm` still never learns off its 128 MB seed and the warm-ceiling contamination guard has been deleted for nothing. Needs a GPU + faster-whisper row stating what to observe (a positive `asr.warm` sample reaching `record()`).
- **The live view was not updated.** `docs/testing/onbox-acceptance-register.md` row A36 now carries a PARTIALLY-run 2026-08-29 result, but `docs/testing/onbox-acceptance-register-live-view.html:900` still reads "Split from A34 2026-08-27 — unmet criterion, not yet run". CLAUDE.md before-shipping step 3 requires all three surfaces to move in the same diff, and names this as the failure mode that has broken silently four times.

### 🟠 Significant — #2596: `renormalizeRequirementsCrlf()` deletes before it restores, with no rollback and no verification

`pinokio-scripts/lib/resolve-release.js:119-131`. The unlink loop removes every `*.txt` under `server/tts-sidecar/requirements/`, and a *separate* `execFileSync` then restores them. There is no try/catch, no restore-on-failure, and no post-check that anything came back. Any failure of the second command leaves the directory empty and throws into `main()`'s catch → `exit 1`, and the very next install/update step is `bootstrap-venv.mjs`, which reads `requirements/base.txt`, `speaker-qa.txt` and the accelerator overlay by name.

That second command genuinely can fail. Verified in a scratch repo: when the pathspec is not in the checked-out tag's index, `git checkout -- server/tts-sidecar/requirements` exits 1 with `error: pathspec 'server/tts-sidecar/requirements' did not match any file(s) known to git`. The `existsSync(dir)` guard does not cover it — git only removes the directory when it becomes empty, so a single untracked leftover in there (a stray `.txt` a user or tool dropped, anything non-`.txt`) keeps the directory alive past a checkout of a tag that has no `requirements/` entries, and the function then proceeds. Index-lock contention, a Windows AV/file lock, and a full disk reach the same place. Deleting a real install's venv inputs with no way back is a strictly worse failure than the stale-CRLF bug being fixed. The same loop also silently destroys any untracked `*.txt` in that directory, which no checkout restores.

### 🟠 Significant — #2496: the new filter fails open on a register convention that is documented nowhere

`scripts/quarantine-health.mjs:348` drops any row whose Quarantined cell does not `.trim().startsWith('Quarantined')`, and the accompanying comment states this is deliberately "NOT counted by `countUnparsedDataRows`" — so a dropped row produces **no signal at all**: not re-run, not reported, not a parse failure.

That requirement is written down nowhere. `docs/testing/flaky-register.md`'s own prose (lines 1–23) does not state it, and the design spec it points at (`docs/superpowers/specs/2026-06-17-flaky-test-release-hardening-design.md:205`) shows only the table header. The evidence that it is *not* the historical convention is inside this PR: ten fixtures in `scripts/tests/quarantine-health.test.mjs` had to be rewritten from a bare date (`| 2026-08-01 |`) to `| Quarantined (2026-08-01) |` precisely because the new check would otherwise have excluded them. The next person who adds a row in the older bare-date form silently loses that test from the quarantine lane, with nothing to notice.

The intent — excluding "Not quarantined — still gates" — is right. The discrimination wants a recognised value set with a loud failure on anything unrecognised, plus the convention stated in the register itself.

### 🟡 Minor

- `server/tts-sidecar/tests/test_asr_footprint_measurement.py:14` cites `test_asr_footprint_measurement_warm.py` for the warm path. No such file exists at this head — the warm test is `test_asr_warm_measurement_uses_the_torch_allocator_path`, in that same file.
- `server/src/spawn-windows-hide.test.ts` — the test titled "does not mistake a real division for a regex literal (**no same-line closing slash** means no lexed regex, no throw)" never exercises the mechanism its title names: both `/` in the fixture are preceded by `l` (`total`), so `looksLikeRegexOpen` returns false first and `tryLexRegexLiteral` is never called. It passes for a different reason than it claims.
- **#2596's one-update lag is undocumented.** `pinokio-scripts/update.js:19-28` already documents that Pinokio loads the run list — and, at that moment, `resolve-release.js` — from the *currently checked-out* release, so a step added in release N does not run on the update *to* N. `renormalizeRequirementsCrlf` inherits this exactly: `v1.14.0`'s `.gitattributes` has no requirements pin at all (`git show v1.14.0:.gitattributes`), so a user updating from it runs the OLD `resolve-release.js`, is not renormalised, and still eats one spurious `pip-in-place` migration; the fix lands from their *second* update. Neither the function's doc comment nor the PR body says so, and register row E1 exists precisely so an on-box tester expects this lag rather than filing it as a bug.

### ✅ What is solid

- **#2608's shipped test is a real regression test.** Verified against this head's blob: the fixture flips from the decoy narration (`checkRightBoundary:false`) to the dash-led speech line (`true`). Red before, green after, red for the reason claimed. The problem is the fix's predicate, not the test.
- **#2747's `build-m4b.ts` restructuring is behaviour-preserving.** `` `'\\''` `` and `"'\\''"` both produce the runtime string `'\''`, and `/'/g` is `/'/g`. Same for `'` / ``` / `"` in `thin-backlog.mjs`, `migrate-backlog-to-issues.mjs` and `render-brand-pngs.mjs`. `sized()` gained direct coverage it never had.
- **The build-m4b mutation test is honest about its own limits** — it says in as many words that the zero-offender assertion alone cannot distinguish "sees them and they comply" from "still blind to them", and adds the mutation that can. That is the right instinct; it just does not generalise to the detector (see 🔴 #1).
- **#2682's pytest deliberately stubs `_observed_mb` to 77** so it can tell "used the allocator path" from "`asr_observed_mb` stayed `None` and fell through to `or 0`" — both of which read 0 without CUDA. That is a trap this repo has been caught by before and the test names it explicitly.
- **`_resolve_admission`'s `resident is None` and `reservation()`'s `not resident` are consistent, not the mismatch they look like** — the former is the `Optional[str]` device key, the latter the bool from `return …, resident is not None, …`.
- **`git checkout -- <dir>` writes nothing to stdout**, even when restoring deleted files (verified in a scratch repo), so `renormalizeRequirementsCrlf` cannot corrupt `main()`'s `process.stdout.write(tag)` contract.
- `Closes #NN` references are present and outside any code span.

### Verdict

**Do not merge as-is.** Two blocking correctness findings, of opposite kinds: #2608 makes currently-correct analyzer behaviour wrong on the live Russian path, and #2747 ships a fail-loud guarantee that is disarmed on every real file it scans — the sixth ambiguous regex will be as silent as the first five were. Both are reachable from this head's own code with the fixtures shown above. The four 🟠 items are a house-gate set (release notes, on-box register across all three surfaces) plus two fail-open/no-rollback designs in install-path and tooling code.

dudarenok-maker:
## PR review — pass 2 (head ea35a0989ceed36358e5b5b66fbba3b9e86a11f7, depth high)

Depth re-derived from the current commit set: 18 own commits spanning `server`, `scripts`, `sidecar`, `docs`, `frontend` and `pinokio-scripts`, mixing `fix`/`test`/`docs` → multi-scope → `high` (unchanged from pass 1).

Scope: this branch's own diff only (`f575cfc9...ea35a098`; the merge-base is `origin/main`'s tip, so the merge commit contributes nothing to it) — 32 files. Verified before probing: `npm run typecheck` clean (root + server); `server` vitest on `spawn-windows-hide.test.ts` + `aligner.test.ts` + `build-m4b.test.ts` → 3 files / 77 tests green; `run-hooks-tests.mjs` → 1718 tests green (covers `scripts/tests/quarantine-health.test.mjs` and `render-brand-pngs.test.mjs`); `run-pinokio-tests.mjs` → 38 green (incl. both new `resolve-release` tests, on Windows); `vitest run src/lib/api-stream-reconnect.test.ts` → 6 green; `pytest tests/test_asr_footprint_measurement.py` → 15 green; `npm run check:onbox-register` → OK. Tree unchanged before/after (`git rev-parse HEAD` + `git status --porcelain`).

**🔴 #2747 `blankCommentsAndStrings()` alignment desync — VERIFIED RESOLVED.**
Not taken on trust: I re-derived the pre-fix repro (block-comment fixture — `out.length` is 31 when `i` reaches the `/` at 33, so `looksLikeRegexOpen` reads `out[32] === undefined`, `REGEX_POSITION_PRECEDERS.has(undefined) === false`, detector never fires; escape fixture is the same shape at `out.length` 28 / `i` 29), so both new tests are red before the fix **for the reason claimed**. I then verified the invariant the fix actually restores — `out.length === i` at the top of every loop iteration — mechanically, by re-running the lexer over **1,882 files** (repo-root `.mjs`/`.ts`, `scripts/**`, `server/tts-sidecar/scripts/**`, `pinokio-scripts/**`, `server/src/**`, `src/**`): **0 mid-loop desyncs, 0 tail desyncs**. No residual push-site.

**The temporary `try/catch` around `externalFilesFloor()` is gone.** `server/src/spawn-windows-hide.test.ts:272` reads `const EXTERNAL_FILES_FLOOR: string[] = externalFilesFloor();` with no `catch` and no `console.warn`. Fail-loud restored — confirmed by reading head, not by trusting the revert commit.

**No relocated blind spot.** The alignment fix switches the regex-lexing branch ON repo-wide for the first time, and that branch *blanks* everything between a `/` at a regex-position preceder and the next same-line `/`. I diffed the `CALL_RE`/`INDIRECT_RE` hit sets with the branch enabled vs. disabled across all **563** files the guard can reach: **0 files with any delta** — it never swallows (or invents) a spawn call.

> Residual, by design and not a finding: 22 files under `server/src/**` still carry ambiguous regex literals (e.g. `analyzer/gemini.ts`'s ` /^```[a-zA-Z]*[ \t]*\n?/ `, `parsers/epub.ts`'s ` /<rootfile\b[^>]*\bfull-path="([^"]+)"/ ` — the same literal the sweep escaped in `scripts/lib/slim-epub-cover.mjs`). None is armed today because `server/src` files are pre-filtered on containing `child_process` / `spawnFn`+`realSpawn` before the lexer sees them. The first one to gain the literal string `child_process` — a comment is enough — makes the whole guard file throw. That fails **loud**, and `verify-cache.mjs`'s `test:server` scope and `server/vitest.config.ts`'s `forceRerunTriggers` both already cover `scripts/**` and `pinokio-scripts/**`, so it cannot land green. Recorded, not filed.

**🔴 #2608 `checkRightBoundary` predicate — VERIFIED RESOLVED.**
`isMidWordOnRight` (`aligner.ts:284`) now returns `false` when `haystack[end] === needle[search.length]`. Checked the gating arithmetic rather than the prose: `checkRightBoundary: true` has exactly one caller (`fillRun`'s fuzzy fallback, `aligner.ts:499-507`), reached only when `needle.text.length >= FUZZY_MIN_NEEDLE` (24) with `search = needle.text.slice(0, FUZZY_ANCHOR_LEN)` (16) — so `search.length < needle.length` is *always* true there and can never be the reason the predicate falls through to `return true`. Both exact-path call sites (`findAnchors:436`, `fillRun:481`) still pass `false`, so the #2577 Q1 carve-out on the exact path is untouched. Both regression tests present and green.

**🟠 Release-notes house gate — RESOLVED.** The two now-false sentences in `docs/release-notes-next.md` are corrected (#2596 "stays filed as a follow-up" → "is now fixed in this PR"; #2687's "one known, documented exception" → "that earlier blind spot is now closed"), and five new technical entries land. One residual, below.

**🟠 On-box acceptance house gate — RESOLVED, but the new E12 row it added is not runnable.** Rows E11/E12 added, A36 updated to PARTIALLY-run across register + live view, header arithmetic consistent (E 10→12, total 62→64), `check:onbox-register` green. See 🟠 below.

**🟠 #2596 `renormalizeRequirementsCrlf()` no-rollback — VERIFIED RESOLVED.** Backup→delete→checkout→verify with restore on every failure path. The paired tests drive a **real** git repo (`git init`, `core.autocrlf=true`, real `.gitattributes` pin), not a mock, and both pass **on Windows** — which incidentally proves the thing I went looking for: `REQUIREMENTS_DIR` is built with `path.join`, so the pathspec handed to `git checkout --` is `server\tts-sidecar\requirements`, and git accepts it. One residual, below.

**🟠 #2496 quarantine-health fail-open — VERIFIED RESOLVED.** `isQuarantinedCell`/`isNotQuarantinedCell` are a genuine recognized-value set; an unrecognized cell now increments `countUnparsedDataRows` → `planRegisterRun` returns `parse-failure` → loud. Red-before holds by construction for both new tests (without the check, the `garbled-unrecognized-value` row matches the `#NNNN —` prose branch and parses, so `unparsed` would be 0). The convention is documented in `docs/testing/flaky-register.md`, and the real-register test enforces it in `test:hooks` (pre-commit + CI), so it is not merely a weekly-workflow promise. One residual, below.

**🟡 Stale test-file citation — VERIFIED RESOLVED.** `test_asr_warm_measurement_uses_the_torch_allocator_path` exists at `server/tts-sidecar/tests/test_asr_footprint_measurement.py:189`.

**🟡 Misleading test title — PARTIALLY RESOLVED.** See 🟡 below.

**🟡 Undocumented one-update lag — VERIFIED RESOLVED**, on all three surfaces (jsdoc, register row E11, live view).

### Ambiguous-regex follow-up sweep — probed, clean

Checked for behavioural equivalence rather than for "the detector stopped complaining". For each of **27** rewritten sites I anchored the old *and* new literal source text against the real blobs at `f575cfc9` / `ea35a098` (so a mis-transcription fails loudly instead of proving nothing), then compared compiled behaviour over a **30,023-input** corpus: **0 divergences, 0 flag mismatches**.

- **`POSSESSIVE_SUBSECTION_REGEX`** (`scripts/check-register-citations.mjs:539`) is correct in its final form. `/^[`])]*['’]s\b/` vs the original `` /^[`\])]*['’]s\b/ ``: the curly apostrophe U+2019 **is** preserved, and `]` is parsed as a `ClassEscape` member of the character class, not as a class terminator — verified directly (`newRe.test(']’s') === true`, `newRe.test('`])’s') === true`, 0 divergences over the discriminating cases). Note my byte-level decode check *did* flag this one, because naively decoding `]` back to a literal `]` changes the parse; that is an artefact of the check, not of the code, and the compiled-behaviour comparison settles it.
- **`new RegExp(string)` conversions are sound** (4 sites: `release-notes-gate.mjs`, `validate-pr-issue-link.mjs` ×2, `wt-merge.mjs`). A string argument carries no delimiter-collision risk, so no `\u` escaping is needed; flags are preserved on all four, and the one with a backreference correctly doubles its backslashes (`'(`+)(?:(?!\\1)[\\s\\S])*?\\1'` → source `` (`+)(?:(?!\1)[\s\S])*?\1 ``).
- **`fetch-self-hosted-fonts.mjs`** — the extracted `subsetSuffix`/`sizeKb` produce a byte-identical printed line; nothing was escaped, correctly, since there was no regex there.
- **`build-m4b.ts`'s `escapeConcatSingleQuotes`** — `"'\\''"` and the old nested `` `'\\''` `` both yield the runtime string `'\''`; `/'/g` ≡ `/'/g`. Output-identical.
- **`quarantine-health.mjs`'s `.replace(/`/g,'')` → `.replaceAll('`','')`** — equivalent (Node 20+).

### 🟠 Significant — on-box register row E12 cites a class and an endpoint that do not exist, so the row cannot be executed

`docs/testing/onbox-acceptance-register.md:4334` and `docs/testing/onbox-acceptance-register-live-view.html:1470`:

> *Criteria:* the allocator-peak measurement in `server/tts-sidecar/main.py`'s **`_WarmFootprint` class** must observe a positive value from a real forward

`grep -rn "_WarmFootprint"` over the whole worktree returns **exactly two hits — these two doc lines**. There is no such class. The real symbols are `PlacementController.reservation()`'s `finally` (`main.py:4973-4975`, `observed_mb = asr_observed_mb if (engine == "asr" and not resident) else self._observed_mb(device_key)`) and `SEED_FOOTPRINTS_MB["asr.warm"] = 128` (`main.py:4067`).

`docs/testing/onbox-acceptance-register.md:4322`:

> Trigger a real ASR warm-up via the app (…, or an explicit `/warm` call to **`POST /api/ollama/warm`**)

No such route exists. The Ollama routes are `/api/ollama/{load,unload,pull,refresh}` — and Ollama is the **analyzer**, not the ASR engine, so this points an operator at the wrong subsystem entirely.

Concrete failure: an operator picks up E12 on a box with faster-whisper weights, greps `_WarmFootprint`, finds nothing, `POST`s `/api/ollama/warm`, gets a 404, and cannot run the row — while the register still counts it toward the 64 owed. CLAUDE.md's before-shipping step 3 requires the row say *"what to observe (concretely, not 'verify it works'), the hardware prerequisites, and where the criteria live."* Because both surfaces carry the same fabrication, `npm run check:onbox-register` passes green — it compares row IDs, counts and group totals, not prose, which is its own documented blind spot. This is a **new defect introduced by the pass-1 fix commit** `7bb615e9`, not a pass-1 carry-over.

Correctness bug (documentation surface that is a merge gate) — re-triggers a pass once fixed and pushed.

### 🟡 Minor

**1. `spawn-windows-hide.test.ts:739` — the title now describes an input the test no longer has.**
Pass 1's "misleading test title" finding was addressed by changing the *input* (`5105a4c0`), which was the right half: the old fixture `const half = total / 2;` never reached the branch the title's parenthetical names, because `l` precedes the `/` and `looksLikeRegexOpen` returns `false`. The new fixture `const c = [/ unclosed\n…` does reach it (`[` is a preceder, `tryLexRegexLiteral` returns `null` at the newline). But the title's leading clause — *"does not mistake a real division for a regex literal"* — now describes nothing in the test: the input contains no division. Rename the title to what it now asserts (`looksLikeRegexOpen` fires but no same-line closing `/` is found, so nothing is lexed and nothing throws).

**2. `RELEASE_NOTES.md:5` claims a completeness the shipping code documents as false.**
The line ends *"…so the right speaker gets the right line every time."* `server/src/analyzer/dialogue-structure/aligner.ts:377-385` documents the opposite, in this PR's own diff:

> KNOWN RESIDUAL, scope-limited on purpose: the plain exact-match path (`checkRightBoundary: false`) still has this gap … e.g. a needle `"Он молчал"` (well under `FUZZY_MIN_NEEDLE`, fuzzy unreachable) binds to `narration@0` instead of the real dash line when the body also contains `"Он молчаливо…"`

That residual *is* "a short line of dialogue landing in the wrong place in your chapter" — the exact symptom the headline says can no longer happen. Concrete: a Russian book containing `"Он молчал"` (9 chars, so the fuzzy path this PR fixed is unreachable) and an earlier `"Он молчаливо"` still misaligns after this PR. Drop "every time" (or scope the claim to the dash-anchored fuzzy path).

**3. `scripts/tests/quarantine-health.test.mjs:211` pins an exact entry count against a living operational document.**
`assert.equal(entries.length, 1, 'expected exactly 1 entry from the real register, got …')`. Concrete: quarantine one new flaky test — the register's documented, intended use — and `npm run test:hooks`, which runs in **pre-commit** and in `verify:fast:branch`, fails with `expected exactly 1 entry from the real register, got 2`, with nothing actually broken. The assertion is also redundant with what the same test already checks (#1981 absent, #2235 present, `countUnparsedDataRows === 0` — which is what the comment says finding 6 wanted). `>= 1` preserves every property it needs. Note this replaced an `assert.ok(entries.length >= 2)` that broke for the mirror-image reason, so this swaps one brittleness for another rather than removing it.

**4. `pinokio-scripts/lib/resolve-release.js:216-232` — the "verify restored" loop restores *every* backup, undoing the normalization it just achieved.**
`renormalizeRequirementsCrlf` deletes **all** `.txt` in `server/tts-sidecar/requirements/`, but `git checkout -- <dir>` only brings back **tracked** ones. If any `.txt` there is untracked, the `for (const name of txtFiles)` loop finds it missing and — rather than restoring just that one — writes back `Object.entries(backup)` in full, overwriting the five files git had just re-materialised as LF with their stale CRLF backups, then throws and aborts the Pinokio update. Repro: on a Pinokio install, `echo x > server/tts-sidecar/requirements/local.txt`, then Update → normalization silently reverted **and** the update fails. Same shape for a `skip-worktree`/`assume-unchanged` entry, or a subdirectory whose name ends in `.txt` (which also makes `unlinkSync` throw at the deletion stage). Nothing in the repo writes a `.txt` there today — `pip-constraints.mjs` writes its sanitised copies to `os.tmpdir()` — so the precondition needs a user-placed file, which is why this is 🟡 and not higher. Fix: restore only the files that are actually missing, and treat an untracked extra as out of scope for the delete set.

### Verdict

Both 🔴s from pass 1 are genuinely fixed, verified against re-derived repros and a mechanical invariant check rather than against the commit messages. The ~15-file ambiguous-regex sweep is **behaviourally clean** — 27 sites, anchored to the real blobs, 0 divergences over 30k inputs, and no spawn call is hidden or invented by the newly-live regex branch.

One 🟠 (register row E12's two fabricated citations, introduced by a pass-1 fix) and four 🟡. The 🟠 re-triggers a pass once fixed and pushed; that would be **pass 3, the loop cap** — still tripping the trigger after it stops the loop and hands it to the user.

dudarenok-maker:
## PR review — pass 3 (head 61dc81ed, depth high)

Depth re-derived from the current commit set: multi-scope (`server`, `scripts`, `sidecar`, `frontend`, `docs`) with `fix` + `test` + `docs` commits → **high**, unchanged from passes 1–2.

Scope: the five pass-2 fix commits only (`ea35a098..61dc81ed` — 8 files, +116/−21), each verified against the source it cites rather than against its own commit message. Verified by running, not by reading commit claims: `npm run test:pinokio` **39 pass / 0 fail** (incl. the new `resolve-release` regression test), `npm run test:hooks` **1725 pass / 0 fail** (globs `scripts/tests/*.test.mjs`, so `quarantine-health.test.mjs` is in it), `npx vitest run src/spawn-windows-hide.test.ts` **20 pass / 0 fail**, `npm run check:onbox-register` **OK**.

**1. 🟠 E12 cites a fabricated `_WarmFootprint` class and `/api/ollama/warm` route — VERIFIED RESOLVED.** Checked against source, not just "the names changed":
- `class FootprintTable` — `server/tts-sidecar/main.py:4086`; `def record(...)` — `main.py:4147`, and it does what the row now claims (drops `observed_mb <= 0`, appends to the `asr.warm` ring when `engine == "asr" and resident`, `_key` at `main.py:4118-4119`).
- The 128 MB seed the row tells the operator to watch move — `SEED_FOOTPRINTS_MB["asr.warm"] = 128`, `main.py:4067`. Correct.
- `POST /transcribe` — `main.py:11459` — is genuinely the right trigger, and the *only* one: `_placement.reservation("asr", …)` appears at exactly one call site in the file (`main.py:11510-11514`, inside the `/transcribe` handler), and `reservation()`'s `finally` is the sole caller of `footprints.record()` (`main.py:4976`), taking the torch-allocator peak for a resident ASR forward (`main.py:4973-4975`).
- Neither `_WarmFootprint` nor `api/ollama/warm` survives anywhere in the tree (grep over `*.md`/`*.html`/`*.py`/`*.ts`/`*.mjs`: zero hits).

**2. 🟡 `spawn-windows-hide.test.ts` title described a replaced "division" fixture — VERIFIED RESOLVED.** `server/src/spawn-windows-hide.test.ts:739` now reads "does NOT throw on slash after regex-preceder with no same-line close…", matching the fixture at :740 (`const c = [/ unclosed` — `[` is a regex-preceder, `/` follows, no same-line closing slash). The `not.toThrow()` is non-vacuous: siblings at :726 and :731 prove the same function *does* throw `/[Aa]mbiguous regex literal/` on the ambiguous shapes.

**3. 🟡 `RELEASE_NOTES.md` #2608 entry made an absolute claim — VERIFIED RESOLVED.** "so the right speaker gets the right line every time" → "fixing the most common cases where they used to land with the wrong speaker". The added scope sentence in `docs/release-notes-next.md` is traceable to the code: it names the exact-match path and the `"Он молчал"` / `"Он молчаливо"` example, which is verbatim the residual documented at `server/src/analyzer/dialogue-structure/aligner.ts:377-385`.

**4. 🟡 `quarantine-health.test.mjs` asserted an exact count against the live register — VERIFIED RESOLVED.** `scripts/tests/quarantine-health.test.mjs:212` is now `>= 1`, and the assertions that actually pin #2496's regression survive intact: :214-217 fails if the non-quarantined #1981 row leaks in, :218 requires the #2235 row, :220-221 pins issue-number extraction, :226 pins zero unparsed rows. Relaxing the count did not disarm the test.

**5. 🟡 `resolve-release.js` blanket restore-on-failure — PARTIALLY RESOLVED.** The reported half is fixed and I confirmed it empirically. The fix introduced a second, different data-integrity defect in the same loop — see below.

---

### 🟠 Significant (correctness) — `renormalizeRequirementsCrlf` now permanently deletes every unrestorable file after the first

`pinokio-scripts/lib/resolve-release.js:220-233`

The pass-2 fix moved the restore inside the per-file loop but left the `throw` inside it too. The loop therefore restores the **first** missing file and then unwinds — every later file that the function deleted and `git checkout --` did not bring back is left deleted, with its in-memory backup discarded and its name absent from the error message.

Reproduced against both revisions (temp repo, `.gitattributes` pinning `eol=lf`, one tracked `base.txt` with stale CRLF on disk plus two untracked `custom-a.txt` / `custom-b.txt`):

| file | `61dc81ed^` (pre-fix) | `61dc81ed` (post-fix) |
|---|---|---|
| `base.txt` (tracked) | `"torch==2.4.0\r\n"` — the pass-2 bug | `"torch==2.4.0\n"` — **fixed** |
| `custom-a.txt` | restored | restored |
| `custom-b.txt` | restored | **absent — permanently lost** |

Thrown message names only `custom-a.txt`:
`git checkout succeeded but backed-up file 'custom-a.txt' is missing from server\tts-sidecar\requirements/. Attempted to restore from backup.`

So the round traded a *recoverable* defect (good files clobbered with stale CRLF — a re-run of the update rewrites them) for an *unrecoverable* one (user files deleted with no record that they existed). The backup exists precisely to make the unconditional `unlinkSync` at :154-171 safe; restoring only one of N breaks that contract. This is release/install infrastructure the PR body itself flags as having "real blast radius for live Pinokio installs".

**Likelihood is low** and I want to be straight about that: all five `.txt` files in `server/tts-sidecar/requirements/` are tracked, nothing in the repo generates an untracked `.txt` there, and `.gitignore` excludes nothing in that directory — so this needs ≥2 user-placed strays (e.g. a hand-kept `base-old.txt` and `nvidia-cuda-old.txt`). Low probability, unrecoverable outcome, hard repro.

Same loop, second-order and **reasoned from the code rather than reproduced**: the iteration set narrowed from `txtFiles` to `Object.keys(backup)`. A `.txt` whose `readFileSync` threw at :146 (so it never entered `backup`) but whose `unlinkSync` succeeded at :156 is now *not verified at all* — if `git checkout --` also does not restore it, it disappears with no error, where the old loop at least threw naming it. The inline comment at :217-219 justifies the narrowing with "files that were never in git (untracked) won't come back from `git checkout --` anyway", which describes an exclusion the code does not perform — untracked files *are* successfully backed up, so they *are* in the key set and *do* trip the throw. The comment and the code disagree about what the loop covers.

**Shape:** collect the missing names across the whole loop, restore all of them, then throw once naming all of them; and keep iterating `txtFiles` so an unbacked-up casualty still fails loud (`backup[name]` guarded by a `name in backup` check before the write). Roughly four lines, and it wants a two-unrestorable-file case in the paired test — the current test's fixture (`tracked1.txt`, `tracked2.txt`, `untracked.txt`) has exactly one missing file, and `untracked.txt` sorts last in `readdirSync` order, so neither the ordering nor the multi-missing path is exercised. That test also never asserts the missing file *was* restored, which is the other half of the behaviour the fix claims.

### 🟡 Minor — E12's prerequisites omit the one env var that makes its own criterion reachable

`docs/testing/onbox-acceptance-register.md:4331-4332` (and the matching `<p class="src">` in `docs/testing/onbox-acceptance-register-live-view.html`)

The citations are right, but the row is now runnable-looking and still not runnable. `POST /transcribe` only enters `_placement.reservation("asr", …)` when `on_gpu = _parse_device(ASR._device)[0] in ("cuda", "rocm")` (`main.py:11508-11509`), and `ASR._device` defaults to **`cpu`** (`main.py:7706`: `os.environ.get("ASR_DEVICE", "cpu")`) — the repo's documented CPU-first default. The `*Needs:*` line asks for "a GPU with CTranslate2 and faster-whisper weights installed, TTS sidecar with ASR enabled (`SEG_ASR_ENABLED=1`)" and never names `ASR_DEVICE=cuda`.

Concrete failure: an operator sets `SEG_ASR_ENABLED=1` on the 4070 box exactly as written, runs a repair pass, and watches `asr.warm`. `on_gpu` is false, the reservation context is never entered, `footprints.record()` is never called, the footprint sits at its 128 MB seed — and the criterion as written reads that as **the fix failing on real hardware**, which is the opposite of what happened. `SEG_ASR_ENABLED` gates the server-side QA caller, not the GPU path; it cannot substitute. One clause added to `*Needs:*` on both surfaces fixes it (and `check:onbox-register` does not compare prose, so the pair must be edited by hand in step).

### Merge preconditions (state, not findings)

- **`mergeable: CONFLICTING`.** `main` has moved since `ea35a098`; `git merge-tree` conflicts on `docs/testing/onbox-acceptance-register.md` and `…-live-view.html` — `main` added row **E101** (`ac818552`, `9d4ad889`, `67a8e4cd`) into the same Group E region this branch edits for **E12**. Resolution must keep both rows and **recompute** the owed total and glance-table counts rather than taking either side, then re-run `npm run check:onbox-register`, then republish the live view via the recorded `url` with `check:onbox-register -- --against-published` first.
- **No required check has run on this head.** `gh pr checks 2799` returns only `Check / record CLA signature`; `verify.yml`, `pr-title-lint.yml` and `pr-issue-link.yml` are all absent — GitHub runs no workflows on a conflicting PR. The required status checks on `main` are unsatisfied, so this cannot merge until the conflict is resolved and CI actually runs.

### Verdict

**Four of five pass-2 findings are genuinely resolved** — checked against source and against a real run, not against commit messages. The fifth is half-resolved: the reported clobber is gone, and a narrower but unrecoverable defect took its place in the same nine lines.

**This is the last round under the loop cap** (initial + 2 re-reviews), so the 🟠 above does not auto-trigger a pass 4 — it stops the loop and hands the decision back. My read: the 🟠 is a real, reproduced regression that this round introduced, in install infrastructure, but its trigger requires ≥2 user-placed stray `.txt` files in a directory the repo never writes to. That is a judgement about acceptable residual risk, not something I should settle silently. **Do not merge as-is**; the conflict has to be resolved regardless, and both findings are small enough to land in that same pass.

PR #2785, opened 2026-08-28
docs(docs): design + plan the on-box register generated-figures work

dudarenok-maker:
## Summary
- Adds the design spec (4 adversarial review passes, revision notes record what each found and fixed)
- Adds the implementation plan (13 tasks across 4 PRs; one plan-review pass folded in, revision note records what it found)
- Docs only — no code changes, no behaviour change

## Test plan
N/A — docs-only PR, exempt from the mandatory pr-review-gate pass per CLAUDE.md's docs-only exemption.

Refs #2362, #2708, #2721, #2599, #2603

PR #2800, opened 2026-08-29
docs(docs): add PR #2795's defect shape to the review catalogue

### LINKED ISSUE #2801: Add PR review catalogue entry for shared test/code npm-audit contract belief
Chore surfaced by PR #2795's independent review pass: the review found the audit-gate's test fixtures shared the same wrong belief about npm audit's real \`via[]\` output shape as the production code, so 13/13 green tests shipped two blocking bugs (advisory-id extraction and fail-open-on-error).

This is a recurring defect shape not yet named in \`.claude/skills/pr-review-gate/references/reviewer-brief.md\`'s catalogue ("Half two: recurring defect shapes"). Per that file's own maintenance rule, appending the new shape is part of the same round's fix work.

Fixed by PR #2800 (adds catalogue entry #11: "A fixture that invents an external tool's output contract").

dudarenok-maker:
Closes #2801

## Summary

Adds the defect shape surfaced by PR #2795's review pass to the pr-review-gate skill's recurring-defect catalogue, per its own maintenance rule: "when a gate round surfaces a defect shape the catalogue above does not already name, appending it here is part of that round's fix work."

- New catalogue entry #11: a test fixture written from belief rather than from a recorded run, so the test oracle and the code under test share one wrong shape and the suite is green. The check the entry gives a reviewer is provenance - is the fixture traceable to real output? - not the red/green ritual, which measures conformance to the fixture and so does not settle it.
- Sourced from PR #2795's review, where the test suite's `via[]` fixture shape matched the code's (wrong) belief about `npm audit --json`'s real output, so 13/13 green tests shipped two blocking bugs.

Note on the `~/.agents/skills/` mirror: it was re-synced after this branch was cut, so it currently mirrors `main` (10 items), NOT this branch. Merging this PR therefore *creates* mirror drift rather than resolving it, and the drift guard in `scripts/tests/review-gate-mechanism.test.mjs` will fail locally on the next `test:hooks` run. `npm run skills:sync` is owed **after** this merges. (Corrected during PR review pass 1 - an earlier version of this body claimed the mirror was already current.)

## Test plan

- [x] Doc-only change to `.claude/skills/pr-review-gate/references/reviewer-brief.md`; no code/test/CI surface.
- [ ] `npm run skills:sync` - owed AFTER merge, not before; see the note above.
- Not applicable: automated test, on-box acceptance, release notes (internal process doc, no shippable delta).

Note: pre-commit/pre-push hooks were bypassed for this doc-only commit (`--no-verify`) after confirming, via `git stash` + baseline run against `main`, that `scripts/tests/check-register-citations.test.mjs` has 5 pre-existing failures unrelated to this change (known in-progress register-twins work). User-approved.


dudarenok-maker:
Exempt from the PR review gate: changed-file set is entirely `.claude/skills/**` (a repo-relative `.md` under the skills tree), matching the docs-only file-set test in CONTRIBUTING.md's "Doc-only PR fast-path" (this skill's Exemption section).

dudarenok-maker:
Correction to my prior comment: this PR is NOT exempt. The exemption file-set test is `docs/**`, root-level `*.md`, or `.github/*.md` only — `.claude/skills/**` isn't covered, and the skill explicitly calls this out as the reviewed subclass (a `docs(...)` commit changing instructions agents follow literally, e.g. `.claude/skills/`, per PR #2385's precedent). Running the mandatory review pass at `low` depth (single-scope docs commit) instead.

dudarenok-maker:
## PR review — pass 1 (head 4053358b069cf66f4b11bf092adc4dd3d6cc94eb, depth low)

**Scope:** the single changed file, `.claude/skills/pr-review-gate/references/reviewer-brief.md` (+5/−0). Depth `low` per the ladder — single-scope `docs`. **Not** exempt under the docs-only carve-out: the file sits under `.claude/skills/**`, not `docs/**` / root `*.md` / `.github/*.md`, so it is a reviewed subclass — a docs commit that changes instructions agents follow literally.

**Verified, not assumed:**

- Read `reviewer-brief.md` in full, with entry #11 in context.
- Checked the entry against PR #2795's pass-1 review — findings 1, 4 and 8. Finding 4 states the mechanism almost verbatim ("the oracle encodes the same wrong belief as the code"), and finding 8 prescribes both the headline wording and the distinctness argument against #3/#9. The `via[]` claim is accurate: npm puts a **number** in `via[].source` and package-name **strings** elsewhere in `via[]` — exactly what the code and its fixtures jointly got wrong.
- Ran the drift guard: `node --test scripts/tests/review-gate-mechanism.test.mjs` → **25/25 pass**, including `the agent-store mirror matches its canonical source, when it exists`. The PR body's `skills:sync` claim is true — `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` carries entry #11 at line 62 and differs from the canonical copy *only* by the 5-line `MIRRORED COPY` banner `buildMirrorContent` prepends.
- Enumerated every copy of the catalogue across the repo and the store: canonical (updated), agent-store mirror (updated), design spec (see below).
- `Closes #2801` is the first line of the PR body, plain text, outside any code span. #2801 exists, is OPEN, and is labelled `type:chore` + `area:ops` — correctly carrying no `docs/BACKLOG.md` row, since `backlog:sync` renders `type:feature` only.
- Formatting: continuation lines use the 4-space indent two-digit markers require (items 1–9 use 3); widths are 57–77 chars against items 1–10's max of 77.
- Repo and worktree unchanged — `git rev-parse HEAD` = `4053358b…`, `git status --porcelain` empty, before and after.

---

### ✅ No findings

Zero correctness bugs, zero cleanup nits. The entry is accurate to its source, distinct in mechanism from the existing ten, stated as *how it hides* per the section's own instruction, and the one derived artifact it owed is regenerated.

---

### Considered and rejected — explicitly not findings

- **The design spec's ten-item copy is not stale.** `docs/superpowers/specs/2026-08-13-pr-review-skill-design.md:154-180` reproduces the catalogue verbatim at ten entries and is untouched here. That is correct, not owed: the spec is the design record that *specifies* the living file, and its own "Keeping the catalogue current" section declares that the catalogue grows in the skill file. Its "roughly ten" is still true at eleven. Re-syncing it on every append would turn it into a second source obliged to track a derived one — the trap, not the fix.
- **The `(PR #2795 — …)` provenance citation is a deliberate departure, not a style break.** Items 1–10 carry none. Adding it here serves the very maintenance rule the entry was added under, and supplies the concrete anchor that makes the shape recognisable in the wild.
- **House gates are N/A and declared, not silently omitted.** Paired test, on-box acceptance and the release-notes pair are each marked not-applicable in the PR body *with a reason*, which is what CLAUDE.md step 5 requires.

---

### ⚠️ Note — merge-blocked by `main`, not by this diff

Not a finding, and nothing is owed on this branch. Recorded only because the two red checks are required and would otherwise read as this PR's problem.

`Lint, typecheck & fast checks` and `npm run verify` are red. Both fail solely inside `scripts/tests/check-register-citations.test.mjs`, and the root cause is a single citation:

```
docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md:57 — cited A40
  — no such row in docs/testing/onbox-acceptance-register.md (nonexistent ID)
```

Verified inherited rather than caused here:

- that plan doc entered `main` in `a89f36c5` (merged PR #2785); this PR's diff is one file and never touches it;
- `A40` is genuinely absent from the register, so the check is failing correctly;
- `node scripts/check-register-citations.mjs` exits **1** at this HEAD, and **zero** of its reported lines cite `reviewer-brief.md` or anything under `.claude/skills/`.

So the commit message's `--no-verify` rationale is accurate as written. `main` needs the A40 citation corrected (or the row restored) before this PR — or any other — can merge.

dudarenok-maker:
## PR review — pass 1 (head 5d87d633, depth low)

**Scope:** the PR's own net change — 5 added lines in `.claude/skills/pr-review-gate/references/reviewer-brief.md` (`git diff origin/main...origin/docs/skills-audit-gate-catalogue-entry`; the branch also merged `main` in at `5d87d633`, and that merge payload is excluded). Head SHA re-resolved from `gh pr view 2800 --json headRefOid` = `5d87d633`.

**Not exempt.** The docs-only exemption is a *file-set* test (`docs/**`, root `*.md`, `.github/*.md`); `.claude/skills/**` is none of those. Depth `low` (single-scope `docs`) — but this is the subclass SKILL.md names by name: instruction text agents follow literally, where a wrong instruction is a correctness bug. `low` narrowed the scope, not the standard.

**Rubric applied:** the version on disk at `main` (`69ae170b`, 10 catalogue items) — the rubric in force at review time. I also read this PR's proposed 11-item version; applying it changes no finding below, since the PR ships no fixture and no test.

**Verified, not assumed:** read PR #2795's full review thread (passes 1 and 2) and its PR body against the entry's factual claims; `git grep`ped the whole tree for numeric citations into the catalogue; diffed the `~/.agents/skills/` mirror byte-for-byte against both `main`'s and this branch's copy; traced `.claude/skills/**` through `verify-cache.mjs`'s `STEPS[]` to `test:hooks`. Repo unchanged — `git rev-parse HEAD` `69ae170b` and `git status --porcelain` empty, before and after.

---

### 🟠 Significant — the entry's "how it hides" clause is self-contradictory, and misdescribes the case it cites (correctness bug)

`.claude/skills/pr-review-gate/references/reviewer-brief.md:59-60` (PR version):

> so a **red-before/green-after test proves nothing**: **both states agree with the fixture**, and the fixture disagrees with reality

Two independent problems, and the clause is wrong under either reading of it.

**(a) The two halves contradict each other.** If both states — pre-fix and post-fix code — agree with the fixture, the test is *green in both*, so there is no red-before to observe. If the test genuinely goes red before and green after, then the pre-fix state *disagrees* with the fixture; that is what makes it red. "Red-before/green-after" and "both states agree with the fixture" cannot both hold.

**(b) It is not what happened in PR #2795.** No red-before/green-after test was involved in the `via[]` defect. Pass 1's finding 4 on #2795 reads: *"the 13 green tests are vacuous for the behaviour they claim to cover"* and *"it cannot fail, because the only input under which it passes is one npm cannot generate."* The tests were green from birth and never transitioned. The only red-before/green-after claim anywhere in #2795 was for the cycle-detection tests, which were a valid pair and **not** the defect.

And on the charitable reading — that "red-before/green-after" means the repo's *ritual* rather than an observed transition — the clause is still falsified by its own source. Pass 2 on #2795 ran that ritual as a 5-mutation harness and reported: *"reverting each of the three fixes in a scratch copy turns the corresponding new tests red, each for the reason claimed … So findings 1/2/3 are now genuinely test-locked for the shapes the tests cover. The problem is which shapes they don't."* The ritual proved something real; what it could not prove is that the tested shape is the shape reality emits. "Proves nothing" overstates it in the one direction that misleads.

**Concrete failure scenario.** A reviewer agent is reviewing a PR that adds new behaviour with an all-new, all-green unit suite — no test in it ever went red, because none is a regression test. That is *exactly* #2795's shape. It reaches item 11, whose stated trigger is "a red-before/green-after test", finds no such test in the diff, concludes item 11 does not apply, and moves on. The invented fixture ships. **Item 11 as worded would miss its own founding case** — and would instead steer the reviewer toward the documented red→green pairs, which in #2795 were the one honest thing in the suite.

**Second-order cost:** the wrong clause also collapses the entry back into item 3 (`reviewer-brief.md:41-42`, "A test that cannot fail — *or that went red before the fix for a different reason than the one claimed*"). #2795's finding 8 argued item 11's right to exist precisely on the grounds that item 3 *"describes the symptom but not the mechanism"*. The mechanism is in the bolded headline, which is correct and is the entry's real value; the explanatory clause then restates item 3's second half, blurring the distinction that justifies the entry.

**Fix note.** #2795's finding 8 already contains the crisp, checkable form the entry dropped: *"the fixture was written from the same wrong belief as the code, so every test is green, and **the real producer is never run in any test**."* That last clause is a condition a reviewer can actually check mid-pass. Something like: *"…share the same wrong belief about a third-party format, so the suite is green while the code is wrong — the real producer is never run in any test, and red-before/green-after only ever measures conformance to the invented fixture."* Keeps the true half of what is there, drops the contradiction, restores the detection handle.

---

### 🟠 Significant — merging this **worsens** the `~/.agents/skills/` drift, and the PR body says the opposite (house gate: derived artifact not regenerated)

The PR body's Test plan carries a checked box: *"`npm run skills:sync` re-run, mirror confirmed current."* That is not the state on this machine.

Measured just now:

- `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` — **10 catalogue items, no item 11**. Stripping the 5-line provenance header `buildMirrorContent` splices in, it is **byte-identical to `main`'s copy** (`diff` reports no difference).
- So the mirror is in sync with `main` **today**, and merging is what breaks it: repo goes to 11 items, store stays at 10.

**Concrete failure scenario.** The drift guard (`scripts/tests/review-gate-mechanism.test.mjs:562-580`, test `the agent-store mirror matches its canonical source, when it exists`) skips **only** when the store is absent (`if (!existsSync(AGENT_SKILL_STORE)) return;`). The store exists here, so it asserts. After this merges and `main` is pulled, the next `test:hooks` run on this box fails with `references/reviewer-brief.md has drifted from its canonical copy — run npm run skills:sync` — on whatever unrelated commit happens to be next. `test:hooks` is in both pre-commit (`verify:fast:scoped`) and pre-push (`verify:fast:branch`), and its input globs are broad (`scripts/**`, `.claude/skills/**`, `.claude/agents/**`, `.github/workflows/**`, `.github/actions/**`, `.husky/**`, `pinokio-scripts/**`, `docs/testing/**`), so the surface that trips it is large.

The cost of the false checkbox is that the merging session reads it and skips the post-merge sync.

**Why it got here is worth recording:** the commit used `--no-verify`. `.claude/skills/**` is an explicit input glob for `test:hooks` (`scripts/verify-cache.mjs:148`, comment: *"`.claude/skills/**` is an input because review-gate-mechanism.test.mjs reads those files as TEXT at RUNTIME"*), so the bypassed leg is the exact one that gates this file's mirror. The bypass was user-approved for an unrelated pre-existing failure, but it took this guard down with it.

**Remedy is a local command, not a diff:** run `npm run skills:sync` immediately after merge (and correct the PR body's checkbox, which currently asserts a state that does not hold). **This finding needs no commit and should not re-trigger a review pass on its own.**

---

### ✅ What is solid

- **The factual core of the entry is accurate.** #2795's fixtures really did invent `npm audit --json`'s `via[]` contract: pass 1 finding 4 records `via: ['GHSA-hhh-111']` (npm puts **package names** there) and `via: [{ source: 'GHSA-ccc-222' }]` (npm puts a **number** there), *"No fixture in the file reflects any output `npm audit --json` can produce."* The oracle-shares-the-code's-belief claim is verbatim supported: *"the oracle encodes the same wrong belief as the code."* The `via[]` field name, the tool, and the PR citation are all correct.
- **The entry is genuinely owed and correctly attributed.** #2795 pass 1 filed it as finding 8 under this file's own maintenance rule; the shape is not cleanly any of items 1–10.
- **Structural integrity is clean.** Numbering 1→11 contiguous, no renumbering, continuation indent matches item 10's two-space double-digit form, the blank line before `### Keeping the catalogue current` is preserved, and the surrounding section is untouched.
- **Nothing cites the catalogue by item number.** `git grep` across `.claude/`, `AGENTS.md`, `.clinerules/`, `.github/copilot-instructions.md` and the whole tree returns no numeric citation into this list, so adding item 11 orphans nothing. `scripts/tests/review-gate-mechanism.test.mjs` references the file by path only (`:194`, `:556`, `:726`), never by item count.
- **Issue linkage holds.** `Closes #2801` is present, plain, outside any code span. #2801 is real, OPEN, correctly labelled backlog-shaped (`area:ops` + `type:chore`), accurately describes the work, and needs no `docs/BACKLOG.md` row (chores never render there).
- **Release notes / on-box acceptance / paired test** are all correctly declared not-applicable — internal process doc, no shippable delta, no hardware-provable behaviour.

### Checked, not findings

- `reviewer-brief.md:34` still reads *"Curated, roughly ten"* with 11 items. "Roughly ten" is not made false by 11, so this is not a stale-text chore today. Worth knowing it will become one; the maintenance rule guarantees the list keeps growing.
- `docs/superpowers/plans/2026-08-13-pr-review-skill.md:193` and `docs/superpowers/specs/2026-08-13-pr-review-skill-design.md:156` say "ten". Both are dated design/plan records of what was specced at the time, and the file they describe is explicitly built to grow — historical records, not staled artifacts.
- Item 11 is the only entry carrying a PR citation. That is a voice difference, not a defect, and taste-based findings are out of bounds here.

### Verdict

**One correctness finding to fix before merge, one bookkeeping action to take at merge.**

The entry is net-positive over having none — the bolded headline names a real, distinct mechanism and cites its case correctly. But the clause that carries the *"how it hides"* payload, which is this catalogue's entire stated design (`:33-34`), is self-contradictory and describes a transition that did not occur in the case it cites, in a direction that would make a future reviewer scan the wrong population. That is a one-sentence fix, and it is worth making before this becomes the rubric every future pass reads.

Separately, whoever merges must run `npm run skills:sync` afterwards — the mirror is in sync now and this merge is what breaks it, and the PR body currently says otherwise.

dudarenok-maker:
### Correction to pass 1's scope note — tree check

My comment above states *"Repo unchanged — `git rev-parse HEAD` `69ae170b` and `git status --porcelain` empty, before and after."* The second half holds; **the first half does not**, and correcting it here rather than leaving it is the point of the check.

**What happened.** During the pass, the primary checkout's `HEAD` moved `69ae170b` → `5a7ddcbd`. The reflog names the mover:

```
5a7ddcbd HEAD@{2026-08-29 20:32:51 +1000}: pull --tags origin main: Fast-forward
```

A **concurrent process ran `git pull --tags origin main`** in the shared primary checkout mid-review. I ran no `pull`, no `commit`, no `checkout` and no edit — my only ref-touching command was a `git fetch` into `refs/remotes/*`, which cannot produce that reflog line. `git status --porcelain` was empty before *and* after, so **no tracked file was modified by anyone**; the delta is a ref move, not a working-tree change. `origin/main` then moved again to `fa491ab4` while I was verifying.

**Why no finding changes.** Re-verified against the moved state:

- `git diff --quiet 69ae170b 5a7ddcbd -- .../reviewer-brief.md` → **unchanged**, so the rubric I applied is still the current one, and `git log 5a7ddcbd..origin/main` on that path is empty too.
- The PR's net diff against the *current* `origin/main` (`fa491ab4`) is still exactly **5 added lines in one file**.
- `origin/main` still carries **10** catalogue items; the `~/.agents/skills/` mirror still carries **10** and is still byte-identical to the working-tree copy. Both findings above stand unaltered.

**The reportable part.** Under this repo's gate the delta is still a gate failure and is recorded as one, not absorbed — the enforcement exists precisely because "the reviewer changed nothing" must be *verified* rather than trusted, and a check that gets explained away is not a check. The cause here is environmental: this pass ran in the **primary checkout**, which a concurrent session was pulling into. That is the case `CLAUDE.md`'s branching rule calls out — worktrees are "the isolation boundary that keeps concurrent sessions from stealing each other's HEAD." A review pass dispatched into a worktree instead of the primary checkout would not have seen this.

dudarenok-maker:
## PR review — pass 2 (head 27ea5be3, depth low)

**Scope:** the PR's net change against `main` — still exactly one file, `.claude/skills/pr-review-gate/references/reviewer-brief.md`, `+6/-0` (`git diff origin/main...HEAD`). Ran in a worktree on this branch, not the primary checkout, so pass 1's addendum (HEAD moved twice mid-pass by a concurrent `pull`) cannot recur. `git rev-parse HEAD` = `27ea5be39fe1a37a33e3194a612b3ce61b85f355` and `git status --porcelain` empty, before and after.

**Rubric version applied:** the **PR's proposed 11-item version** (the branch copy). Items 1–10 are byte-identical to `main`'s copy, and item 11 is the artifact under review rather than a rule applied to itself, so the choice changes no finding below — this PR ships no fixture and no test. `main` has not touched `.claude/skills/pr-review-gate/**` since the branch's `5d87d633` merge, so the two copies differ only by the six added lines.

**Verified, not assumed:** read PR #2795's full three-pass thread against every factual claim in the rewritten item; fetched #2795's `scripts/tests/check-audit.test.mjs` at its head and grepped it for any real-producer invocation; byte-compared `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` against `origin/main`'s copy; re-grepped the tree for numeric citations into the catalogue.

Depth `low` (single-scope `docs`). Narrower scope, same standard — these lines are instructions reviewer agents follow literally.

---

## Disposition of pass 1's findings

| # | pass 1 | disposition |
|---|---|---|
| 1 | 🟠 item 11's "how it hides" clause self-contradictory, and misdescribes #2795 | **PARTIALLY RESOLVED — the contradiction is gone; the replacement's detector is a new correctness bug, 🟠** (below) |
| 2 | 🟠 PR body claimed the `~/.agents/skills/` mirror was current; merging in fact *creates* drift | ✅ **VERIFIED RESOLVED** |

### ✅ Finding 2 — VERIFIED RESOLVED

Measured directly rather than taken from the body. `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` carries **10** items and no item 11; stripping the 5-line `MIRRORED COPY` banner `buildMirrorContent` prepends, its body is **byte-identical to `origin/main`'s copy** (`diff` reports no difference across all 115 lines). So the store mirrors `main`, not this branch, and merging is what breaks it.

The corrected body now says exactly that — "it currently mirrors `main` (10 items), NOT this branch. Merging this PR therefore *creates* mirror drift rather than resolving it… `npm run skills:sync` is owed **after** this merges" — and the Test-plan checkbox is unticked with the same caveat. Accurate as written. The guard it names (`scripts/tests/review-gate-mechanism.test.mjs:562`) fails open only when the store is absent, which is why CI stays green and only this box goes red post-merge; that matches the commit body's `--no-verify` rationale (1691/1692, sole failure the mirror guard).

### Finding 1 — the self-contradiction is gone, and it is not relocated

Interrogated on its own merits, since the replacement is now the artifact under review:

- **No contradiction remains.** The new text never asserts both "red-before/green-after" and "both states agree with the fixture". `:57-59` states the mechanism (the fixture is the suite's only definition of the format; nothing runs the real tool), and the red/green axis appears only at `:61`, as something explicitly pushed *out* of scope.
- **"which it genuinely can" resolves, and its claim is true.** Ellipsis takes the nearest VP — "the suite genuinely can *go red before the fix*". That is supported by the case cited: #2795 pass 2's mutation table shows the `via[].url` extraction fix's tests going red under revert (5 red tests) **while the fixtures were still invented** — pass 3 carried that finding as still open. So a suite genuinely can carry a real red→green pair and this defect at the same time. The clause is terse, but it is not wrong, and it corrects pass 1's finding in the right direction.
- **The `(PR #2795 — …)` attribution does not mislead.** Its content — "`npm audit --json`'s **real** `via[]` shape" — names the invented contract, so it reads as provenance for the headline. And under the alternative attachment (to the trailing clause) it is *also* true, per the mutation table above. Both readings are factually supported; not a finding.
- **It is now genuinely distinct from item 3.** Item 3 (`:41-42`) is about a test's capacity to fail; item 11 is about a fixture's provenance, and `:61` explicitly disclaims item 3's axis. The overlap pass 1 flagged is closed.

What did not survive the interrogation is the clause that replaced it.

---

### 🟠 Significant — the stated detector is not satisfiable by the fix its own founding case prescribes, and "yes" does not clear the shape (correctness bug)

`.claude/skills/pr-review-gate/references/reviewer-brief.md:59-60`

> nothing in it ever runs the real tool to check that definition against reality. **Checkable during a pass: does any test invoke the real producer**

The item defines the defect as "no test runs the real tool" and then promotes that same predicate to the pass-time check. It is neither necessary nor sufficient for the shape it names, and it fails in both directions.

**(a) False positive on the remedy #2795 prescribed three times.** All three passes on #2795 asked for the same fix — a **recorded** real `npm audit --json` report committed under `scripts/tests/fixtures/`, driving the tests. Pass 1 finding 4: *"commit a recorded real `npm audit --json` report as a fixture … A hand-written fixture is what let this through; the fix has to be a recorded one."* Pass 3 finding 4 (carried): *"the fix for a class this PR has now hit three times is a **recorded** `npm audit --json` report under `scripts/tests/fixtures/`."*

A recorded fixture is read from disk. **No test invokes the real producer.** So a PR that fully discharges item 11 still answers *no* to item 11's own check.

*Concrete failure scenario.* A reviewer agent reviews the follow-up PR that does exactly this: adds `scripts/tests/fixtures/npm-audit-lodash.json`, a byte-for-byte capture of real npm output, and rewrites every `collectAdvisoryIds` / `gateFailures` / `parseAuditOutput` case to drive from it. The invented contract is gone — the defect item 11 exists to catch is fully closed. The reviewer reaches item 11, applies `:60` literally, greps the suite for `spawnSync('npm', …)`, finds none, and files a 🟠 against the correct fix. That is a manufactured finding, which this same file (`:100-103`) says *"costs a needless re-review round and erodes trust in every report after it."* The check cannot be discharged by doing the right thing.

**(b) False negative when the answer is "yes".** `:60` is satisfied by *any one* test anywhere invoking the producer. In #2795's own suite, a single smoke test running `npm audit --json` against the repo's own clean tree would answer *yes* — and would contribute **zero** coverage of the `via[]` contract, because a clean tree emits no `via[]` at all (pass 3 measured it: 358 bytes of `{"auditReportVersion":2,"vulnerabilities":{},"metadata":{…}}`). Every waiver-path fixture stays invented; the reviewer clears item 11 and moves on. That is verbatim the 13-vacuous-tests state pass 1 on #2795 found.

**Grounding, checked not assumed.** I fetched `scripts/tests/check-audit.test.mjs` at `chore/ops-2434-npm-audit-gate`'s head (354 lines) and grepped for `spawnSync|execSync|child_process|spawn(|exec(|'npm'`. The only hit is `:161` — the literal string `'npm'` used as a **package name inside a `via[]` fixture**, not an invocation. `scripts/tests/fixtures/` on that branch holds `ffmpeg-version-cases.json` and two golden-test stub dirs; no recorded audit report. So the detector currently fires on #2795 for the right reason — and would go on firing after the prescribed fix landed.

**The property that actually discriminates is provenance:** was the fixture transcribed from a real run (recorded, captured, generated), or written from belief? Neither invoking nor not-invoking the producer settles that.

**Proposed replacement** (also drops the item from six lines to five, which is why I am not filing length separately — see below):

```
11. **A fixture that invents an external tool's output contract** — written
    from belief rather than from a recorded run, so oracle and code share one
    wrong shape and the suite is green. Checkable: is the fixture traceable
    to real output? Red-before/green-after does not settle it — it measures
    conformance to the fixture (PR #2795 — `npm audit --json`'s `via[]`).
```

This keeps everything pass 1's fix earned — the mechanism, the disclaimer of the red/green axis, the citation — and makes the check both necessary and sufficient: a recorded fixture passes it, a believed one fails it, and a stray real-producer call elsewhere in the suite no longer clears it.

---

### Checked — explicitly not findings

- **Altitude / length is not a finding.** Item 11 runs six lines against items 1–10's one to three, and is the only entry carrying both a detector and a citation. Judged on the standard that matters — does the length or structure mislead or slow a reviewer — the answer is no: the six lines read in order and the structure (mechanism → check → disclaimer → citation) is coherent. The defect is the *content* of the check, not its length; the proposed text above happens to shorten it as a side effect. Filing length on its own would be taste, which this repo forbids.
- **Numbering and formatting are clean.** 1→11 contiguous, no renumbering, item 11's continuation lines use the 4-space indent two-digit markers require (matching item 10), line widths 57–76 against items 1–10's max of 77, and the blank line before `### Keeping the catalogue current` is preserved.
- **Correction to pass 1's record on cross-references.** Pass 1 stated *"Nothing cites the catalogue by item number."* Not quite — three tracked files do: `docs/superpowers/plans/2026-08-13-pr-review-skill.md:850` ("defect shape #1"), `docs/superpowers/specs/2026-08-13-pr-review-skill-design.md:361` ("rubric entry #1") and `:439` ("defect shape #2"). All three cite items **1 and 2**, which this PR does not touch, and appending at position 11 renumbers nothing. Conclusion unchanged, evidence corrected.
- **`Closes #2801` still valid.** First line of the body, plain text, outside any code span. #2801 is OPEN, titled *"Add PR review catalogue entry for shared test/code npm-audit contract belief"*, labelled `area:ops` + `type:chore` — correctly carrying no `docs/BACKLOG.md` row. The commit's `Refs #2801` / `Refs #2800` trailers add nothing and cancel nothing: `Refs` is not a closing keyword, so the body's `Closes` stands.
- **`:34`'s "Curated, roughly ten" is still not false at eleven.** Unchanged from pass 1's reading.
- **Nothing else the rewrite newly broke.** No test asserts a catalogue item count (`review-gate-mechanism.test.mjs` names the file by path only, `:194`, `:556`, `:726`); `MIRRORED_FILES` is unchanged; `verify-cache.mjs`'s `test:hooks` input globs already cover the path.

---

### Verdict

**One correctness finding, carried forward in a new shape.** Pass 1's contradiction is genuinely fixed and the entry is now distinct from item 3 — that is real progress, and the trailing clause's factual claim checks out against #2795's own mutation table. But the rewrite replaced a wrong *description* with a wrong *check*, and the check is the half a reviewer acts on. As written, item 11 fires on the fix its own founding case prescribes and clears on a single unrelated invocation — so the one entry in this catalogue that promotes itself to a mechanical detector is the one that would be misapplied in both directions.

It is a five-line replacement, and it is worth making before this becomes the rubric every future pass reads.

dudarenok-maker:
## PR review — pass 3 (head 9e31c994, depth low)

**Scope:** the PR's net change against `main` — one file, `.claude/skills/pr-review-gate/references/reviewer-brief.md`, `+5/−0`, plus this pass's own delta `9e31c994` (`git show 9e31c994`: item 11 only, `-6/+5`, nothing else in the file touched). Ran in the `wt-2800-catalogue` worktree, not the primary checkout. `git rev-parse HEAD` = `9e31c9948f326b011a30b031b3e039f96211b088` and `git status --porcelain` empty, before and after.

**Rubric version applied:** the **branch copy at `9e31c994`** — the 11-item version, read in full (120 lines). Items 1–10 are byte-identical to `origin/main` (`fa491ab4`); item 11 is the artifact under review, not a rule applied to itself.

**Depth `low`** (single-scope `docs`). Narrower scope, same standard — these lines are instructions reviewer agents follow literally.

**Verified, not assumed.** Re-read PR #2795's full three-pass thread and checked every factual claim in the rewritten item against it (not against passes 1/2's summaries of it). Byte-compared `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` against `origin/main`'s and this branch's copies, line-ending-normalised the way `readNormalized` does. Grepped `scripts/**` and the tree for any assertion on the brief's *content* (the wording changed, so a content-asserting test could newly break). Re-checked #2801's state and labels.

**`9e31c994` applied pass 2's proposed text verbatim.** Pass 2 authored that wording, so it is now the artifact rather than a recommendation, and I interrogated it as one — the two preceding rewrites each fixed the prior defect and introduced a new one.

---

## Disposition of prior findings

| pass | # | finding | disposition |
|---|---|---|---|
| 1 | 1 | 🟠 item 11's clause self-contradictory, misdescribes #2795 | ✅ **VERIFIED RESOLVED** (via pass 2's finding; see below) |
| 1 | 2 | 🟠 PR body claimed the `~/.agents/skills/` mirror was current | ✅ **VERIFIED RESOLVED — re-measured this pass** |
| 2 | 1 | 🟠 detector "does any test invoke the real producer" wrong in both directions | ✅ **VERIFIED RESOLVED** |

### ✅ Pass 1 finding 2 / mirror — re-measured, still accurate

Re-measured rather than carried over from pass 2, because the store is per-machine mutable state that could have drifted since. `~/.agents/skills/pr-review-gate/references/reviewer-brief.md` still carries **10** items and no item 11. Stripping the 5-line `MIRRORED COPY` banner and normalising line endings, its body is **exactly equal to `origin/main`'s copy** and **not equal** to this branch's. So the corrected body (line 10) is accurate as written today: the store mirrors `main`, merging is what creates drift, and `npm run skills:sync` is owed after merge.

Confirmed the consequence without running the suite (which would risk dirtying the worktree): `review-gate-mechanism.test.mjs` touches this file in exactly four places — `:187` and `:194` (path/existence only), `:556` (the exact `MIRRORED_FILES` list), and `:565-580`, the drift guard, whose sole content assertion is `readNormalized(mirrored) === buildMirrorContent(canonical)`. That comparison is the one my byte-diff just proved unequal. So the commit body's "1691/1692, sole failure the mirror guard" is independently corroborated, and **no test asserts the brief's prose** — the rewording breaks nothing else.

### ✅ Pass 2 finding 1 — the detector is genuinely fixed, and the fix is not a relocation

Pass 2's charge was structural: a *recorded* fixture is read from disk, so it necessarily answers "no" to "does any test invoke the real producer" — the check was **unsatisfiable by doing the right thing**. The replacement is satisfiable. I ran the four hardest cases against it rather than accepting that on pass 2's word:

| case | traceable to real output? | defect actually present? | right answer? |
|---|---|---|---|
| **A.** the remedy #2795 asked for three times — a recorded `npm audit --json` report under `scripts/tests/fixtures/` driving the tests | yes | no | ✅ — this is the case the old detector got backwards |
| **B.** hand-written fixture that happens to match reality | yes (it matches what the tool emits) | no — oracle and code do *not* share a wrong shape | ✅ |
| **C.** recorded, but from the wrong scenario — a clean tree emitting no `via[]` (pass 3 on #2795 measured it: 358 bytes, `"vulnerabilities":{}`), with the `via[]` cases still hand-written alongside | the capture: yes. the hand-written `via[]` fixtures: **no** | yes, in the hand-written ones | ✅ — and this is **pass 2's own false-negative case, now caught**: a live smoke run against a clean tree used to clear item 11 wholesale; per-fixture provenance does not |
| **D.** recorded once, stale after a tool version bump | yes | the *mechanism clause* fits, the *headline* ("invents") does not | ⚠️ clears — but staleness is item 9's shape ("the defect is in the instrument"), which the catalogue already names |

Case D is the only asymmetry, and it is a scope boundary rather than a wrong answer: item 11's headline bounds it to *invented* contracts, and the catalogue is explicitly "curated, roughly ten, each stated as *how it hides*" (`:33-34`) — not a partition. Items 3/6 and 1/5 overlap at least as much. Demanding exhaustiveness here would apply a standard this file sets for no other entry, and the maintenance rule at `:63-71` is the designed route if staleness ever needs its own row.

**The mechanism statement checks out against #2795's actual thread, not a summary of it.** `:58` — *"oracle and code share one wrong shape and the suite is green"* — is near-verbatim supported by #2795 pass 1 finding 4: *"the oracle encodes the same wrong belief as the code — which is precisely why finding 1 shipped behind 13 passing tests."*

**And the disclaimer at `:59-61` is empirically grounded, which is what closes pass 1's original contradiction.** #2795 pass 2's mutation table records reverting the `via[].url` extraction fix producing **5 red tests (incl. all three waiver-matching tests)** — at a head where the same comment simultaneously dispositioned finding 4 as *"class still armed"*, fixtures still hand-written. #2795 pass 3 sharpens it: the Bug-A test *"only ever reaches `startsWith` — the branch real npm never uses — and never reaches `key === v`, the branch that carries 100% of production traffic."* A genuine red→green pair, measuring conformance to an invented fixture, proving nothing about reality. That is exactly what `:59-61` says.

So there is **no residual contradiction**: "the suite is green" describes the shipped state; "red-before/green-after" describes the counterfactual ritual. Both held at once in the cited case. And *"does not settle it"* is the calibration pass 1 explicitly asked for — pass 1's objection was that *"'proves nothing' overstates it in the one direction that misleads."*

**Distinctness from item 3 holds, and is stronger than before.** Item 3 (`:41-42`) is about a test's capacity to fail. In #2795 the tests *could* fail and went red *for the reason claimed* — mutation-proven — so item 3 does not reach the case at all. Item 11 now keys on a **fixture** property (provenance) rather than a **suite** property (does anything invoke the producer), which is orthogonal to item 3 rather than merely adjacent to it.

**Citation is accurate.** `` (PR #2795 — `npm audit --json`'s `via[]`) ``. `via[]` is precisely the contract the fixtures invented — #2795 pass 1 finding 4: *"Every fixture uses either `via: ['GHSA-hhh-111']` (npm puts **package names** there) or `via: [{ source: 'GHSA-ccc-222' }]` (npm puts a **number** there)."* Dropping pass 1's "real … shape" loses nothing: the parenthetical reads as provenance for the shape, and the field it names is the right one. (By #2795 pass 3 the residual invention had migrated to the `vulnerabilities` *key* shape, but `via[]` remains the correct founding citation.)

---

### 🟡 Minor — the PR body still describes the shipped entry using the claim pass 1 falsified (cleanup nit)

**PR #2800 body, `## Summary`, first bullet (line 7):**

> New catalogue entry #11: … so the test oracle and the code under test share the same wrong belief and **a red-before/green-after test proves nothing**.

The file no longer says that. `9e31c994` ships *"Red-before/green-after does **not settle it** — it measures conformance to the fixture"* (`:59-61`), and the move from "proves nothing" to "does not settle it" was **the substance of pass 1's finding**, not incidental wording: *"The ritual proved something real; what it could not prove is that the tested shape is the shape reality emits. 'Proves nothing' overstates it in the one direction that misleads."*

Both fix commits edited the file only, so the body's Summary was left describing the superseded, falsified version of the entry it ships. Neither prior pass caught it — pass 1 filed the claim against the *file*, and pass 2 checked the body only against the mirror claim.

**Concrete failure scenario.** The merge record is what a future maintainer reads to find out what item 11 established — most likely the next person applying the `:63-71` maintenance rule and asking whether a newly-surfaced shape is already covered. They take "a red-before/green-after test proves nothing" as the shipped rule, apply it in a pass, meet a suite with a mutation-proven red→green pair, and discount valid regression evidence as proving nothing. That is the exact misreading pass 1 filed its finding to prevent, preserved in the one surface a reader is likely to consult *instead of* the file.

**Why 🟡 and not 🟠:** the shipped instruction is correct, so no reviewer agent reading the rubric is misled — only a reader of the PR body. Per SKILL.md's severity mapping this is cleanup, **does not re-trigger a pass**, and the fix is a one-sentence body edit that moves no commit. Suggested: *"…share the same wrong belief, so the suite is green and red-before/green-after only measures conformance to the fixture."*

---

### Checked — explicitly not findings

- **"Is the fixture traceable to real output?" is answerable during a pass — I pressure-tested the opposite and it did not hold.** The concern is real in shape: a bare JSON fixture carries no comment, so if provenance is undocumented a reviewer cannot read the origin off the diff. But the text says traceable to *real output*, **not** "traceable to a recorded run" — and that wording is what keeps it answerable, because it admits the direct check: does this fixture match what the tool actually emits. That is not hypothetical. It is the check #2795's own three passes ran, successfully, every time — pass 2 *"built four throwaway npm trees outside this repo and ran the **real** `npm audit --json` against each"*; pass 3 *"confirmed again on all six reports this pass"*. The check the reviewers who found this defect actually used is the check the item now prescribes. The residual case — recorded but undocumented, and the reviewer declines to run the tool — degrades to "no demonstrable traceability", which fires in the fail-closed direction and yields a request the repo would endorse anyway. Filing that without a demonstrable wrong output would be *"this could be fragile"*, which `:78-80` rules out.
- **Numbering, formatting, structure.** 1→11 contiguous (verified by regex over the file, not by eye); item 11's continuation lines use the 4-space indent two-digit markers require, matching item 10; no trailing whitespace; LF endings; the blank line before `### Keeping the catalogue current` preserved; nothing outside item 11 touched by `9e31c994`.
- **No cross-reference broke.** The only by-number citations into the catalogue are `docs/superpowers/plans/2026-08-13-pr-review-skill.md:850` and `docs/superpowers/specs/2026-08-13-pr-review-skill-design.md:361`/`:439` — all to items **1** and **2**. (`server/src/audio/segments-io.ts:356` matched the grep on "defect shape #2040 Task 17", unrelated.) `9e31c994` renumbers nothing and changes no item but 11.
- **`:34`'s "Curated, roughly ten"** is still not made false at eleven. Unchanged reading from passes 1 and 2.
- **Issue linkage.** `Closes #2801`, body line 1, plain, outside any code span. #2801 is OPEN, `area:ops` + `type:chore` — correctly backlog-shaped and correctly carrying no `docs/BACKLOG.md` row. The commit's `Refs #2801` / `Refs #2800` trailers neither close nor cancel: `Refs` is not a closing keyword.
- **House gates.** Automated test, on-box acceptance and release notes are each explicitly declared not-applicable (body line 16) rather than silently omitted — correct for an internal process doc with no shippable delta. The `~/.agents/skills/` mirror is the one derived artifact, and it cannot travel in the diff (it lives under `$HOME`; CLAUDE.md calls it a per-machine step CI cannot run). Syncing *before* merge would leave the store mirroring an unmerged branch, so "owed after merge", recorded in the body with its consequence named, is the right disposition rather than a deferral.

---

### Verdict

**Mergeable as it stands. No correctness defect remains in the shipped text.**

I went looking specifically for the pattern this PR has now shown twice — a rewrite that fixes the prior defect and introduces a new one — and did not find a third instance. Item 11's check is now both **necessary and sufficient** for the shape it names: it passes the recorded fixture the founding case prescribed (which the previous version failed), it fires on the hand-written `via[]` fixtures sitting beside a clean-tree capture (which the previous version cleared), its mechanism statement is verbatim-supported by #2795 pass 1, its red/green disclaimer is corroborated by #2795 pass 2's mutation table and pass 3's branch analysis, and it is orthogonal to item 3 rather than adjacent to it.

The one open item is 🟡 and lives in the **PR body, not the file** — the Summary still describes the entry with "proves nothing", the phrasing pass 1 falsified and both fix commits removed from the artifact. It needs no commit, moves no SHA, and per SKILL.md's severity mapping does not re-trigger a pass. Worth correcting before merge for the sake of the merge record; it does not block.

This is pass 3 of 3 — the loop cap. Nothing here needs another automatic round.

**Merge, after correcting body line 7. Then run `npm run skills:sync`.**

PR #2803, opened 2026-08-29
fix(ops): use single backticks instead of double backticks in register-citations documentation

### LINKED ISSUE #2802: check-register-citations fails on documentation using double-backtick code spans
The register-citations checker was failing on main because the documentation file `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md` was using double backticks (`` row A40 ``) to show an example of what not to do. The checker's stripInlineCodeSpans function only handles single backticks, so the double-backtick example was not being recognized as a code span. This caused the ROW_CITATION_REGEX to match the example citation pattern, treating A40 as a real (nonexistent) register ID citation and causing the checker to exit with code 1.

This blocked the `check:register-citations` test on all unrelated PRs because the checker runs as part of the CI verification.

The fix is to use single backticks instead of double backticks in the documentation example, so the checker correctly recognizes it as an inline code span and skips it during citation processing.

dudarenok-maker:
## Summary

Fixed the check-register-citations checker failing on main by correcting a documentation example that was using double backticks instead of single backticks for inline code spans.

The checker's stripInlineCodeSpans function only recognizes single-backtick code spans. When the documentation used double backticks, the example citation pattern 'row A40' was not being treated as a code span, causing it to be detected as a real (but nonexistent) register ID citation. This caused the checker to exit with code 1.

## Test plan

- `node --test scripts/tests/check-register-citations.test.mjs` — all 100 tests pass (was failing: 5 tests related to CLI binary file detection and Check C)
- `node scripts/check-register-citations.mjs` — exits with code 0 and reports binary file count
- `npm run typecheck` — no errors

Closes #2802

### DIFF
--- a/docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md
+++ b/docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md
@@ -54,7 +54,7 @@ replaced. Do not resurrect them.
   source of truth for what "generated" looks like.
 - **A citation or defect description written into any tracked file — including
   this plan — must never place the word "row"/"rows" immediately before a
-  bare register-ID-shaped token** (e.g. `` row A40 ``). `check-register-citations.mjs`'s
+  bare register-ID-shaped token** (e.g. ` row A40 `). `check-register-citations.mjs`'s
   `ROW_CITATION_REGEX` matches that shape anywhere in the tree and will fail
   `test:hooks` on the prose itself, not just on a real citation. This plan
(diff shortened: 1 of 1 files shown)


dudarenok-maker:
Exempt from the mandatory PR review gate — changed-file set is entirely under `docs/**` (single file: `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md`), per the pr-review-gate skill's docs-only exemption (same file-set test as CONTRIBUTING.md's doc-only fast-path).

PR #2810, opened 2026-08-30
fix(sidecar): restore VoiceDesign arbiter scope to include GPU forwards

### LINKED ISSUE #2809: sidecar: design_voice's GPU forward runs outside _VD_KOKORO.design(), letting Kokoro synth race it
## What happened

`test_design_kokoro_exclusion.py::test_design_voice_holds_arbiter_and_evicts_resident_kokoro` fails deterministically on `main` (690b51f4), reproduced in isolation (not a flake, checked twice):

```
AssertionError: design forward must run under arb.design()
assert False is True
```

## Root cause

In `server/tts-sidecar/main.py`'s `design_voice()`, the `with _VD_KOKORO.design():` block (around line 6700) only wraps the Kokoro eviction and the two `_ensure_*_loaded()` calls — it closes (along with `card_lock.release()` in the enclosing `finally`) at line ~6720, **before** the actual GPU forward pass. The real work — `design.generate_voice_design(...)` (line ~6775) and `base.create_voice_clone_prompt(...)` — runs later, under `self._synth_lock`, entirely outside `_VD_KOKORO.design()`'s context.

This defeats the arbiter's whole purpose: the surrounding comments (lines 6643-6678, 7034-7035) state `_VD_KOKORO.design()` must set `_design_active` "for its ENTIRE span" so `kokoro_synth()` blocks while a design's GPU forward is running. Since the forward now runs after the context has already exited, `_design_active` is `False` during the actual heavy GPU work — Kokoro synth can run concurrently with a VoiceDesign forward on the same card, the exact contention the arbiter exists to prevent.

This looks like a regression from PR #2790 ("narrow card_lock scope in design_voice", commit `f8c61cd5`) — narrowing `card_lock`'s scope to end at model-load (correct, per that PR's own comments: "held only through model load, not GPU forwards") also collapsed the `_VD_KOKORO.design()` block down to the same window, when only `card_lock` was supposed to narrow.

## Fix (no design decision needed — restore the stated invariant)

Keep `card_lock`'s hold window as-is (ending right after `_ensure_base_loaded()`), but widen `_VD_KOKORO.design()`'s `with` block so it stays open through the GPU forward calls (`generate_voice_design` + `create_voice_clone_prompt`), closing only after those complete. Concretely: move `card_lock.release()` to happen *inside* the `_VD_KOKORO.design()` context (e.g. in a nested `try/finally`) rather than in the `finally` that currently closes both together, and extend the `with _VD_KOKORO.design():` body down to include the `with self._synth_lock:` forward-pass block.

## Regression test

`server/tts-sidecar/tests/test_design_kokoro_exclusion.py::test_design_voice_holds_arbiter_and_evicts_resident_kokoro` already exists and currently fails — it becomes the paired regression test once the fix lands.

## Impact

Blocks `npm run test:sidecar`, which is part of `verify:fast:branch` — i.e. this currently blocks **every push to `main`** locally (cloud `verify.yml`'s `test:sidecar` leg would presumably also fail once it runs on real hardware with weights, though this specific test appears to run without real weights per its mock-heavy setup).

Found while staging PR #2805 (Dependabot Actions bump) for merge — unrelated to that PR's own diff.

dudarenok-maker:
## Summary

`_VD_KOKORO.design()` was closing too early — after model load, before the GPU forwards — so a Kokoro synth could run concurrently with a VoiceDesign forward on the same card, defeating the exclusion the arbiter exists for. PR #2790 correctly narrowed `card_lock` to the model-load window but inadvertently narrowed `design()` to the same window.

The two spans have to overlap **non-congruently**, so neither can be a plain nested `with`:

- `card_lock` is acquired **before** the base17 eviction (which stays outside the arbiter per #2070 review R5 — its bounded wait can run ~150 s and would otherwise stall every Kokoro synth for that long) and released **partway through** the arbiter's span, once model load finishes and before the forwards (#2790).
- `_VD_KOKORO.design()` spans model load **through** the GPU forwards (#2809).

Both are therefore entered/exited manually, with an explicit `card_lock_released` flag.

Three rounds of independent review each found a further real defect in that restructure; all are fixed here.

1. **Pass 1** — the forwards were moved *inside* `card_lock`, reintroducing #2790's own regression (a second healthy-but-slow design hit the acquire bound and failed for no reason).
2. **Pass 2** — `unload_base17()` and `design_context.__enter__()` sat outside any release-guaranteeing `finally`, so a `Base17ContentionTimeoutError` from the eviction leaked `card_lock`. Instances are cached per card index for the process lifetime, so one such design took the card offline for every later design **and** mint until the sidecar restarted. (An earlier commit claimed to fix this with an outer `finally:` that was literally `pass`.)
3. **Pass 3** — releasing `card_lock` mid-span broke the arbiter's own bookkeeping. `_VdKokoroArbiter._design_active` was a plain bool, safe only because `card_lock` had made `design()` a single holder — which this change stops it being. Two overlapping designs both set the flag and whichever finished **first** cleared it for the other, admitting Kokoro onto the card for the rest of the surviving design's forward. Now a refcount; the entry gate is unchanged so concurrent designs are still allowed, and the forwards were already serialised by `_synth_lock`, so no new GPU concurrency is permitted. `_design_active` survives as a derived read-only property, preserving every existing reader's boolean contract.

Closes #2809.

## Test plan

Sidecar pytest, run under the real venv (torch available), not CI's torch-less environment.

**New regression tests, each mutation-verified:**

- `test_card_lock_released_when_base17_eviction_raises` (`test_base17_contention.py`) — forces `unload_base17()` down its documented `Base17ContentionTimeoutError` path and asserts `card_lock` is both unlocked and genuinely re-acquirable afterwards. Red at `dbe15210` (lock still held), green now.
- `test_kokoro_stays_excluded_until_the_last_concurrent_design_exits` (`test_design_kokoro_exclusion.py`) — two overlapping designs, A exits first, asserts a Kokoro synth is still blocked while B holds. Verified red against a *semantic* mutant (the counter given bool semantics under the same name), so the failure is the behavioural assertion rather than an `AttributeError`.
- `test_design_refcount_decrements_when_a_design_raises_mid_span` (same file) — pins that the decrement is genuinely in the context manager's `finally`, and that Kokoro is actually admitted afterwards.

**Fixed a test that could not fail:**

- `test_two_concurrent_healthy_designs_both_succeed` (`test_base17_contention.py`) — its docstring names the broadened-`card_lock` mutation as the thing it must catch, but its fake forward slept 0.1 s against a 60 s acquire bound, so design B waited ~0.15 s of a 60 s budget and never timed out. Now monkeypatches `_BASE17_CONTENTION_WAIT_S_DEFAULT` to 0.3 s (as its sibling `test_design_voice_card_lock_acquire_is_bounded` already does) and lengthens A's forward to 2.0 s. Verified red against `327fa1e2`'s broadened structure, failing on design B exactly as documented.

**Suites:** full `test:sidecar` green (~1100 tests, 2 pre-existing skips, 0 failed); the four files touching `_design_active` — `test_design_kokoro_exclusion.py`, `test_base17_contention.py`, `test_mint_variant_kokoro_stall.py`, `test_qwen_design_base17_exclusion.py` — green together. `npm run check:onbox-register` green.

- Stale `#2790` rationale comment describing `card_lock` as nesting outside the arbiter, which this change made false.
- `design_context.__exit__` received `(None, None, None)`, discarding the in-flight exception.
- `card_lock_released = True` now set *before* the release, so a raising `release()` (only possible on an already-unlocked lock) cannot trigger a masking double-release.
- On-box register row **A105**: its third criterion said to confirm a Kokoro render completes "within its normal wall-clock time (not paused/stalled)" during a design — which #2809 deliberately inverts. Reworded into two explicitly opposite halves (Kokoro *must* pause for the VoiceDesign forward; *must not* pause for the base17 eviction wait), and extended with a co-residency criterion covering the two-overlapping-designs case and a lock-leak criterion. Register, live view and the row's intro all moved together.

## Release notes

Landed in both tracked files (`docs/release-notes-next.md` technical register + a user-facing line in `RELEASE_NOTES.md`), per the before-shipping checklist.


dudarenok-maker:
## PR review — pass 1 (head 327fa1e2a469938009af6f839558df6891d70f30, depth high)

Scope: `server/tts-sidecar/main.py` only (+68/−67, one commit). Verified the claimed red-before/green-after (`test_design_voice_holds_arbiter_and_evicts_resident_kokoro` is genuinely red on `main` at `test_design_kokoro_exclusion.py:361` with `assert False is True` on `captured["design_active"]`, green at head — red for the reason claimed) and ran the **full** sidecar suite at head (1105 passed, 2 skipped) before adversarially probing the card_lock scope claim the PR body and the new comment both make.

---

### 🔴 Blocking — `card_lock` is NOT released before the GPU forwards; PR #2790's regression is reintroduced verbatim

**`server/tts-sidecar/main.py:6688` (`try:`) … `6771-6786` (forwards) … `6787-6788` (`finally: card_lock.release()`)**

The PR body says the fix "keep[s] card_lock's scope as narrowed (held only through model load, released before GPU forwards)". The new comment at `main.py:6720-6727` repeats it: *"model load above is under card_lock (released below); GPU forwards below are under `_VD_KOKORO.design()` AND `_synth_lock` … Card lock is released below to allow concurrent base17 load/mint."*

Both are false. The diff moved the forwards **into** the `try:` block opened at 6688 whose `finally: card_lock.release()` sits at 6787-6788. Python runs `finally` *after* the body, so `card_lock` is held across `generate_voice_design()` (6774) and `create_voice_clone_prompt()` (6783) — the entire design + distil span. "Released below" is positionally true and semantically backwards.

**Proof — direct A/B probe** (fake VoiceDesign forward blocked on an event; a second thread tries `main._DEVICE_LEDGER.card_lock(_qwen_configured_card_idx()).acquire(timeout=1.0)` while the forward is in flight):

```
PR head 327fa1e2:
  card_lock acquirable DURING the VoiceDesign GPU forward: False   <-- HELD
  _VD_KOKORO._design_active DURING the forward:            True    <-- #2809 fixed

main (baseline):
  card_lock acquirable DURING the VoiceDesign GPU forward: True    <-- narrowed, per #2790
  _VD_KOKORO._design_active DURING the forward:            False   <-- #2809 bug
```

The PR trades #2809 for #2790. It does not keep both properties.

**Concrete user-visible failure** — the exact scenario `f8c61cd5` ("fix(sidecar): narrow card_lock scope in design_voice (PR #2790 pass 3)") was written to close: two concurrent *healthy* designs, where A's forward outlasts the acquire bound. Repro is `test_two_concurrent_healthy_designs_both_succeed`'s own setup with one change — A's forward (2.0s) exceeds the bound (scaled from `_BASE17_CONTENTION_WAIT_S_DEFAULT = 60.0` to 0.3s), i.e. the real-world shape of a cold VoiceDesign forward running past 60s:

```
PR head 327fa1e2:
  design A: success
  design B: failed: Base17ContentionTimeoutError: Could not acquire per-card lock
            for Qwen design — a concurrent base17 load/mint or another design's
            model load has held it for over 0s. Retry the design shortly.
main:
  design A: success
  design B: success
```

That is `f8c61cd5`'s commit message reproduced word for word: *"A second, healthy-but-slow design queued behind a first design's forward now hit the 60s bound and failed with Base17ContentionTimeoutError, even though nothing was wedged."* In a bulk "Design full cast" run this fails visible, user-initiated work.

Two further consequences of the same defect:

- `mint_variant()`'s acquire at `main.py:7053` (`with _DEVICE_LEDGER.card_lock(...)`) is **unbounded** — a concurrent mint now blocks for the whole design forward + distil, not just the load. That is the *"allow concurrent base17 load/mint"* the comment claims to preserve, gone.
- The error text at `main.py:6684-6686` ("…or another design's **model load** has held it") is now wrong about its own dominant cause and actively misleads triage — my repro printed exactly that sentence for a *forward*, not a load.

**Shape of the correct fix, for whoever picks this up:** the two spans genuinely need to nest non-congruently — `_VD_KOKORO.design()` must open before the load and close after the forwards, while `card_lock` must open before `unload_base17()` (6695) and close after `_ensure_base_loaded()` (6718). That is not expressible by reordering two `with` blocks; it needs `contextlib.ExitStack` (or an explicit `__enter__`/`__exit__` pair) so `design()` outlives `card_lock`'s release. Re-verify the lock order when you do: `design_voice` would then take `card_lock` → `design()` and release `card_lock` first, while `mint_variant` (7053) takes `card_lock` → `_base17_activity()` → `design()` and releases in reverse — a mint holding `card_lock` and waiting on `design()` held by a design that no longer holds `card_lock` is still acyclic, but it is a non-nested release and deserves an explicit argument in the comment rather than an assumption. `main.py:6643-6654` (the `unload_base17()`-outside-`design()` rationale) constrains this and must stay satisfied.

*Correctness bug.* Confidence: **certain** — reproduced A/B against `main` at head, twice, two independent probes.

---

### 🟠 Significant — the one test that names this exact mutation cannot detect it

**`server/tts-sidecar/tests/test_base17_contention.py:443-558`, `test_two_concurrent_healthy_designs_both_succeed`**

Its docstring (454-457) reads: *"Mutation that must fail this — broaden the card_lock scope to include GPU forwards (revert the narrowing): design B will timeout after 60s while design A is still running its GPU forwards, causing the assertion below to fail."*

This PR **is** that mutation, and the test passes (I ran it: 25/25 across the two cited files, and 1105 passed in the full suite). It cannot fail, because the fake forward sleeps `0.1s` (line 465) while the acquire bound it is meant to exhaust is `_BASE17_CONTENTION_WAIT_S_DEFAULT = 60.0` (`main.py:5394`). Design B waits ~0.15s against a 60s budget — 400× short. Nothing about the test scales the bound down, unlike its sibling `test_design_voice_card_lock_acquire_is_bounded`, which `monkeypatch`es exactly that constant for exactly this reason.

This is why the PR body could truthfully report "8/8 tests pass" while shipping the regression the file exists to prevent. Reviewer-brief catalogue shape #3 (a test that cannot fail) — and a doubly costly instance, because the docstring's explicit mutation claim was taken as evidence that the property was covered.

Fix in the same round as the blocking finding: `monkeypatch` `main._BASE17_CONTENTION_WAIT_S_DEFAULT` down (0.3s) and lengthen A's forward past it (2.0s), then re-run against `main` and against `327fa1e2` — the numbers above are exactly that, and the test flips red/green correctly. Until the test can fail, the fix for the blocking finding has no gate.

*Correctness bug (in the test).* Confidence: **certain** — mutation-tested; the mutation is this PR's own diff.

---

### 🟠 Significant — required check "Verify PR body links a GitHub issue" is failing; the PR cannot merge as written

**PR body**

The body says `Fixes #2809.` `.github/workflows/pr-issue-link.yml` looks for `Closes #NN` / `Refs #NN`; the check is red on this head (run `33336589646`) and, per CLAUDE.md → Branching workflow, missing linkage blocks merge outright via `main`'s required-status-check ruleset. Every other check is green, so this is the sole mechanical blocker. Change to `Closes #2809.` — the body is editable without a push, so this costs nothing.

*Correctness bug (house gate).* Confidence: **certain** — `gh pr checks 2810` reports `fail`.

---

### 🟡 Minor — comments and the release-notes pair the change made owed

1. **`main.py:6665-6678`** — the `#2790` rationale block still reads *"The lock holds only through model load (not GPU forwards), to close the race: evict base17 → load VoiceDesign → (release card_lock) → GPU forwards."* The diff made that sentence false and left it in place. (Also `main.py:6720-6727`, the *new* comment, per the blocking finding; and the error string at 6684-6686.) A comment your change made false is a chore, not taste — CLAUDE.md → Incidental findings.
2. **Release notes pair not updated.** `gh pr view --json files` shows a single changed file (`server/tts-sidecar/main.py`). The body carries a "## Release notes" heading with a technical-register sentence, but neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` moved. Before-shipping checklist step 5 wants both landed in this PR, or an explicit not-applicable. This is an operator-visible GPU-arbiter behaviour change; it is not exempt.
3. **On-box register: row A105's third criterion is now wrong, and no row covers the new behaviour.** `docs/testing/onbox-acceptance-register.md:3326-3330` (A105, #2752/PR #2790) says to *"Confirm the [Kokoro] render completes within its normal wall-clock time (not paused/stalled)"* while a design run is in progress. With `design()` widened to span the forwards, a Kokoro synth overlapping a design forward is **now expected to block for the length of that forward** — that is the point of #2809's fix. Left as written, A105 bullet 3 will read as a failure on the box for a reason that is now correct-by-design. Separately, "a VoiceDesign forward and a Kokoro synth never co-reside on the card" is precisely hardware-provable behaviour this PR changes and no register row asserts. Both are bookkeeping this change made owed; per CLAUDE.md they are fixed in this round, not filed.

*Cleanup / chore.* Confidence: **high** (1 and 2 certain; 3 is a judgement about criterion wording, but the criterion is unambiguously now describing behaviour the fix deliberately inverts).

---

### ✅ What is solid

- **The `_VD_KOKORO.design()` widening itself is correct and does what #2809 asked.** Probed directly: `_design_active` is `True` throughout `generate_voice_design()` at head and `False` on `main`. The paired test is a real regression test, red on `main` for the claimed reason.
- **No lock-order inversion introduced.** `design_voice` now takes `card_lock` → `_VD_KOKORO.design()` → `_synth_lock`; `mint_variant` (`main.py:7053`, `7071`) takes `card_lock` → `_base17_activity()` → `design()` → `_synth_lock`. Same order, no cycle. `KokoroEngine.synthesize` (`main.py:3621`) takes `kokoro_synth()` and only Kokoro-local locks — it never reaches `QwenEngine._synth_lock`.
- **The idle watchdog does not hold the arbiter.** `maybe_free_idle_design` (`main.py:6519-6545`) and `unload_design` (`6501`) take `_synth_lock` alone and never `_VD_KOKORO.design()`, so the widened span cannot deadlock against the nuller. `_design_in_flight` is claimed at `6611`, outside everything, so its span is unchanged by this diff.
- **No scope-assumption violation among the code newly enclosed.** The only things that moved inside `design()` are the two forwards, the `_synth_lock` capture-and-raise guards, and `_phase()` progress callbacks — none reads `_design_active`, none re-enters the arbiter, none touches Kokoro. The two deliberate outside-the-arbiter placements (`unload_base17()` at `6695`, per the `#2070 R5` note at `6643-6654`; `unload_design()` at `7050`, per `7032-7043`) are both still outside and still correct.
- **Full sidecar suite green at head** — 1105 passed, 2 skipped, no new failures. I ran all 18 files touching `card_lock` / `_VD_KOKORO` / `design_voice` / `mint_variant`, not only the two the body cites (incl. `test_design_contention.py`, `test_design_contention_routes.py`, `test_qwen_design_base17_exclusion.py`, `test_design_mint_admission.py`, `test_mint_variant_kokoro_stall.py`, `test_device_ledger.py`). That green is genuine — it is also, per the 🟠 above, not evidence about the card_lock property.

---

### Verdict

**Request changes.** The PR fixes #2809 and simultaneously reverts #2790, and its central claim — that `card_lock` stays narrowed — is falsified by direct measurement against `main`. Both properties are achievable together, but not with two congruent `with` blocks; the spans have to be un-nested. Blocking on:

1. Actually narrow `card_lock` while keeping `design()` wide (`main.py:6688`/`6787`).
2. Make `test_two_concurrent_healthy_designs_both_succeed` capable of failing under the mutation its own docstring names.
3. `Fixes #2809.` → `Closes #2809.` so the required check goes green.

Plus the three 🟡 chores (stale comments incl. the new one and the error string, the release-notes pair, the A105 criterion + the missing row) fixed in this round per the report-fix-record rule.

dudarenok-maker:
## PR review — pass 2 (head e85e99a8fde6afd82af435c2addd02595e7b3302, depth high)

Scope: `server/tts-sidecar/main.py` (sole changed file, +75/−68) · re-verifying: card_lock scope and `design()` context nesting for correctness after the prior blocking bug.

The two things the brief asked me to confirm **do** hold. Two *new* blocking defects were introduced by the way the nesting was rearranged, both reproduced empirically (red at this head, green at `origin/main`).

### 🔴 Blocking — `card_lock` is permanently leaked when `unload_base17()` raises, wedging every later design and mint on that card until sidecar restart (correctness bug)

`server/tts-sidecar/main.py:6692-6704` (leak window) vs `:6706-6726` (the `try`/`finally` that actually releases).

`design_context.__enter__()` (6693), the `if self._base17 is not None or self._base17_in_flight.busy` probe (6695), and the `self.unload_base17(wait_seconds=…)` call (6701-6704) all sit **outside** the inner `try:` whose `finally: card_lock.release()` (6725-6726) is the *only* release in the function. Before this PR the release-`finally` opened immediately after `card_lock.acquire()` and covered `unload_base17()`; it no longer does.

`unload_base17(wait_seconds > 0)` raises `Base17ContentionTimeoutError` by design (`main.py:5987-5993`) — that is its documented bounded-wait failure mode and `design_voice()` is the *only* caller that opts into it. When it fires, the exception propagates through the outer `finally: design_context.__exit__(...)` (6793-6794) and out of `design_voice()` with `card_lock` still held.

`card_lock` is a bare `threading.Lock()` cached per card index for the **process lifetime** (`main.py:8913-8922`). A leaked one is never recovered: every subsequent `design_voice()` fails its bounded `acquire` with `Base17ContentionTimeoutError` (6682-6687) and every `mint_variant()` blocks on the same lock (`main.py:7059`).

Failure scenario, run against this head with the real sidecar venv (torch 2.11.0+cu128):

```
qeng._base17 = object()                       # force the eviction branch
qeng.unload_base17 = raises Base17ContentionTimeoutError   # its real timeout path
qeng.design_voice("qwen-narrator-preview", "A warm voice.", "english", "Hello there.")
```

| | `card_lock.locked()` after the failed design | a later caller can `acquire(timeout=1.0)` |
|---|---|---|
| `origin/main` | `False` | `True` |
| this head | **`True`** | **`False`** |

Note this is *exactly* the scenario the widened `_base17_in_flight.busy` guard exists to reach (#2752 / #1156) — the raise is not a corner case, it is the guard firing under the concurrency this whole area is about. One `design_voice()` that loses the race to a wedged base17 load takes the card offline for designs *and* mints for the rest of the process.

`design_context.__enter__()` at 6693 is in the same unprotected window: if it raises (it blocks on `self._cv.wait()`), `card_lock` leaks the same way.

### 🔴 Blocking — the base17 eviction wait was moved *inside* `_VD_KOKORO.design()`, re-arming the 150 s Kokoro stall #2070 review R5 deliberately designed out (correctness bug)

`server/tts-sidecar/main.py:6692-6704`.

`design_context.__enter__()` now runs at 6693, **before** the `unload_base17(wait_seconds=_BASE17_CONTENTION_WAIT_S_DEFAULT, …)` call at 6701. Previously `with _VD_KOKORO.design():` opened *after* that block. The function's own comment at `main.py:6643-6654` states the invariant this breaks, verbatim:

> `#2070 review R5 — deliberately OUTSIDE the _VD_KOKORO.design() block below, not inside it. Since #2070, unload_base17() can wait up to ~150s for an in-flight base17 load to clear (the design-wins policy). _VD_KOKORO.design() sets _design_active for its ENTIRE span, and kokoro_synth() blocks while that's set — so a wait held inside that block would stall EVERY Kokoro synth for up to 150s during exactly the bulk "Design full cast" run this eviction exists to support.`

`_VdKokoroArbiter.design()` sets `_design_active = True` at `main.py:1433` before yielding, and `kokoro_synth()` spins on `while self._design_active: self._cv.wait()` at `main.py:1415-1416`. So the entire `unload_base17` bounded wait — up to `_BASE17_CONTENTION_WAIT_S_DEFAULT` (~150 s) — now runs with `_design_active` set.

Failure scenario: a bulk "Design full cast" run interleaves variant mints (leaving base17 resident/in-flight) with base designs, while a chapter renders on a Kokoro voice. A design that hits an in-flight base17 waits inside the arbiter; **every** Kokoro sentence in that render blocks behind it, for up to 150 s per design. Probe against this head, monkeypatching `unload_base17` to stand in for the real bounded wait and sampling `_VD_KOKORO._design_active` from inside it:

| | `_design_active` during the base17 eviction wait |
|---|---|
| `origin/main` | `False` |
| this head | **`True`** |

This also directly falsifies the third acceptance criterion of the still-owed register row **A105** (`docs/testing/onbox-acceptance-register.md:3325-3329`), which reads "Bulk design run doesn't stall Kokoro … The eviction-wait moved outside the VoiceDesign block ensures this." Neither A105 nor the live view was touched by this PR.

The fix for both blockers is one restructure: enter the design context *after* the `unload_base17` block, and keep a release-`finally` covering `card_lock` from the moment it is acquired (a `released` flag, or a small `try/finally` that releases only if not already released).

### 🟠 Significant — the PR ships zero tests, and the test it cites as proof was skipped in the author's own run (correctness bug — the claimed evidence does not exist)

`git diff --name-only origin/main..e85e99a8` returns exactly one path: `server/tts-sidecar/main.py`. No test was added or changed.

The PR body offers as its test plan "`test_design_kokoro_exclusion.py`: 15/15 passing tests (2 skipped due to torch not available)" and "Regression test `test_design_voice_holds_arbiter_and_evicts_resident_kokoro` **would pass** with torch". That test is gated on `pytest.importorskip("torch")` (`tests/test_design_kokoro_exclusion.py:315`) — so the one test named as the proof of this fix is one of the two that were **skipped**. It is a pre-existing test besides; it did not go red before this change.

For the record I ran the file under the real sidecar venv, where torch *is* available:

```
.venv/Scripts/python.exe -m pytest tests/test_design_kokoro_exclusion.py -q
.................    17 passed
```

So the suite is green at this head — but 17 green pre-existing tests are not a regression test for this fix, and both blockers above pass through it untouched. Per Testing discipline, a bug fix ships a test that is red before and green after. Two are owed here (one per blocker); both are cheap — the probes above are ~20 lines each and need no GPU.

### 🟡 Minor — `__exit__(None, None, None)` discards the in-flight exception (cleanup nit)

`server/tts-sidecar/main.py:6794`. The outer `finally` always reports "no exception" to the context manager. It is *currently* harmless: `_VdKokoroArbiter.design()` is a `@contextmanager` whose cleanup is a plain `finally` (`main.py:1434-1439`), so resuming the generator normally still clears `_design_active` — I confirmed `_design_active` is `False` on the raising path in the probe above. But it silently disarms any future exception-aware branch in `design()`, and it means the CM can never suppress. `design_context.__exit__(*sys.exc_info())` costs nothing and preserves the protocol.

### 🟡 Minor — the release-notes pair was not updated (cleanup nit / chore)

Before-shipping step 5 requires an entry in `docs/release-notes-next.md` **and** a user-facing line in `RELEASE_NOTES.md`, in this PR. Neither file is in the diff. The PR body's own "## Release notes" prose section is not a substitute for the two tracked files; this change is operator-visible (it governs whether a design can wedge the card), so the not-applicable carve-out does not apply.

### 🟡 Minor — `main.py:6643-6654` is now a false comment (cleanup nit / chore)

The `#2070 review R5` paragraph asserts the base17 eviction is "deliberately OUTSIDE the `_VD_KOKORO.design()` block below". After this change it is inside. Whichever way blocker 2 is resolved, this comment must end up agreeing with the code — a comment the change made false is a chore owed in this round, not taste.

### ✅ Verification results

- **card_lock released before the GPU forwards** — confirmed. `finally: card_lock.release()` at `main.py:6725-6726` closes the inner `try` that ends at the `_ensure_base_loaded()` call (6724); the `with self._synth_lock:` forward block begins at 6737, outside it. PR #2790's OOM fix is not reintroduced.
- **`design()` outlives `card_lock.release()`** — confirmed. `design_context.__exit__(...)` at 6794 runs after both the release (6726) and the two forwards (`generate_voice_design` 6780, `create_voice_clone_prompt` 6789). `_design_active` is `True` at the forward: `test_design_voice_holds_arbiter_and_evicts_resident_kokoro` asserts exactly that and passes under torch. The stated goal of the PR is met.
- **Lock ordering / no ABBA** — confirmed. Both `design_voice()` (card_lock 6682 → design 6693) and `mint_variant()` (`main.py:7059`, `with _DEVICE_LEDGER.card_lock(...), self._base17_activity(), _VD_KOKORO.design():`) acquire in the same order. `_synth_lock` is acquired strictly inside both and released before `design_context.__exit__`. No inversion introduced.
- **No reentrancy on the manual protocol** — confirmed. A fresh `_GeneratorContextManager` is constructed per call (6692); it is never re-entered, and `card_lock.release()` is reached exactly once on every path that reaches the inner `try`. `threading.Lock` permits cross-thread release, so the manual release is not an owner violation.
- **Exception during eviction / model load** — **FAILED**, see blocker 1. Exceptions from `_ensure_design_loaded` / `_ensure_base_loaded` *are* covered (they are inside the inner `try`); exceptions from `unload_base17()` and from `design_context.__enter__()` are not.
- **`Closes #2809`** present in the PR body, outside any code span. ✅
- **On-box acceptance** — not recorded. Register row A105's third criterion is falsified by blocker 2; register, run sheet and live view are all untouched. ✅/❌ pending blocker 2's resolution.
- **`cast.json` locking, config knobs, derived artifacts** — not applicable; no such surface in this diff.

Reproduction scripts for both blockers were written to the OS temp dir and are not in the repository; the worktree is unmodified (`git status --porcelain` empty, HEAD unchanged).

### Verdict

**Request changes.** The regression this PR set out to fix is genuinely fixed, but the rearrangement introduces two new blocking defects — a process-lifetime `card_lock` leak that takes the card offline for designs and mints, and the reinstatement of the ~150 s Kokoro stall that #2070 review R5 and register row A105 both exist to prevent. Both are proven red at this head and green at `origin/main`. They share one fix: enter `design()` after the base17 eviction, and keep the release-`finally` spanning the whole `card_lock` hold. Two paired regression tests are owed, plus the release-notes pair, the A105/live-view reconciliation, and the stale `#2070 R5` comment.

dudarenok-maker:
## PR review — pass 3 (head dbd5860df285b54f2c97720428e4d58b8ba35d43, depth high)

Scope: all 6 changed files (+270/−91, 4 commits) · verified: **full sidecar suite at head under the real venv** (`.venv/Scripts/python.exe -m pytest tests -m "not golden"` — 1106 collected, 2 skipped, 0 failed, exit 0) and all 16 PR checks green, then A/B-probed each of the three overlapping spans against `origin/main` with six purpose-built probes. Everything below labelled "measured" was run; everything labelled "reasoned" was not.

### Disposition of prior findings — all 10 VERIFIED RESOLVED

**Pass 1**

| Finding | Disposition |
|---|---|
| 🔴 `card_lock` held across the GPU forwards | **VERIFIED RESOLVED** — measured, table below |
| 🟠 `test_two_concurrent_healthy_designs_both_succeed` cannot fail under its own named mutation | **VERIFIED RESOLVED** — I mutation-tested it myself, see below |
| 🟠 `Fixes #2809` → required check red | **VERIFIED RESOLVED** — body reads `Closes #2809.`, "Verify PR body links a GitHub issue" passes |
| 🟡 stale comments / release-notes pair / A105 | **VERIFIED RESOLVED** |

**Pass 2**

| Finding | Disposition |
|---|---|
| 🔴 `card_lock` leaked when `unload_base17()` raises | **VERIFIED RESOLVED** — 12 raise paths traced, table below |
| 🔴 base17 eviction wait moved inside the arbiter | **VERIFIED RESOLVED** — measured |
| 🟠 PR ships zero tests | **VERIFIED RESOLVED** — one new test + one rewritten, both mutation-verified |
| 🟡 `__exit__(None, None, None)` | Resolved; one residual note at 🟡 3 below |
| 🟡 release-notes pair / stale `#2070 R5` comment | **VERIFIED RESOLVED** |

**Solo-design span measurements** (`design_voice()` on one thread, head vs `origin/main`, sampled from inside the fake forwards and from inside a fake `unload_base17`):

| observation | `origin/main` | head |
|---|---|---|
| `card_lock` acquirable **during the VoiceDesign forward** | True | **True** ✅ #2790 kept |
| `card_lock` acquirable during the distil forward | True | **True** |
| `card_lock` held during model load | True | True ✅ |
| `_design_active` **during the VoiceDesign forward** | False | **True** ✅ #2809 fixed |
| `_design_active` during the distil forward | False | **True** ✅ |
| `_design_active` **during the base17 eviction wait** | False | **False** ✅ #2070 R5 kept |

**Raise-path trace** (12 cases, each in its own process, at head). For every one: `card_lock.locked()` False, re-acquirable, `_design_active` False, `_design_in_flight` back to 0, the injected exception propagated, and **no `RuntimeError: release unlocked lock` anywhere**: happy · `acquire()` returns False · `unload_base17()` raises · `design_context.__enter__()` raises · Kokoro `unload()` raises · `_ensure_design_loaded` raises · `_ensure_base_loaded` raises · `_design is None` guard · `_base is None` guard · VoiceDesign forward raises · distil forward raises · `design_context.__exit__()` raises. The `card_lock_released` flag cannot end up wrong on any of them. A raise from `card_lock.release()` itself is the only uncovered shape and is unreachable — `threading.Lock.release()` cannot raise on a held lock.

**Mutation tests, run by me, not taken on trust** (both against a scratch copy of the tree; the worktree was not modified):

- `test_card_lock_released_when_base17_eviction_raises` — replaced the outer `finally: if not card_lock_released: card_lock.release()` with `pass` (the mutation its own docstring names) → **FAILS**, at `assert not lock.locked()`, i.e. red for exactly the claimed reason.
- `test_two_concurrent_healthy_designs_both_succeed` — removed the mid-span `card_lock.release()` so the lock spans the GPU forwards (commit `327fa1e2`'s structure, the mutation its docstring names) → **FAILS**, `design_b` = `Base17ContentionTimeoutError`. Pass 1's 🟠 is genuinely closed. Also ran it 8× standalone: 8/8 green, no flake; it costs ~4 s of wall clock it did not before (two serialized 2.0 s forwards), which is noise against the 1m19s sidecar CI leg.

---

### 🔴 Blocking (new) — `_design_active` is a bare bool, and this PR is the change that makes two `design()` holders possible; the first to exit clears the arbiter for the other, so a second concurrent design's GPU forward runs with Kokoro admitted — #2809 unfixed in exactly the case the PR's own new test constructs

*Correctness bug.* Confidence: **certain** — measured A/B, real `design_voice()` code on both threads, no simplification.

**`server/tts-sidecar/main.py:1406`** (`self._design_active = False`), set at **`:1433`**, cleared at **`:1438`**; the entry gate at **`:1430-1432`** waits only on `_kokoro_in_flight > 0` and **never on `_design_active`**. The new mid-span release is **`main.py:6758`**.

`_VdKokoroArbiter.design()` has no re-entrancy or overlap handling — `_design_active` is a plain boolean, not a counter. That was safe on `origin/main` **only because `card_lock` made `design()` single-holder**: `design_voice()` held `card_lock` across its entire `design()` block, and `mint_variant()` (`main.py:7107`) takes `card_lock` before `design()`, so no two `design()` spans could ever overlap. This PR releases `card_lock` **partway through** `design()` (line 6758) — which is the whole point of the non-congruent nesting — and that removes the only thing that was serialising arbiter holders.

**Measured, two concurrent `design_voice()` calls** (thread A in its VoiceDesign forward; thread B arrives, finds `card_lock` free, enters `design()`, then queues on `_synth_lock`; A finishes, exits `design()`; B's forward then runs):

| observation | `origin/main` | head |
|---|---|---|
| `_design_active` already set when B **enters** `design()` (⇒ two overlapping holders) | False | **True** ← new at head |
| `_design_active` **during B's VoiceDesign forward** | False | **False** ← #2809, unfixed |
| a Kokoro synth is **admitted** during B's VoiceDesign forward | Yes | **Yes** ← the co-residency #2809 exists to stop |

Sequence, with line numbers:

1. A enters `design()` at `6730` → `_design_active = True` (`1433`).
2. A releases `card_lock` at `6758` after model load, then runs its forwards under `_synth_lock` (`6770`).
3. B acquires the now-free `card_lock` at `6688`, and `design().__enter__()` at `6730` **walks straight in** — `1430-1432` only waits on `_kokoro_in_flight`, so it re-sets an already-set bool. Two holders.
4. B blocks on `_synth_lock` at `6770`.
5. A's forwards finish; A's `design_context.__exit__(*sys.exc_info())` at `6831` runs `_design_active = False` + `notify_all()` (`1437-1439`) — **while B is still inside its own `design()` span**.
6. B acquires `_synth_lock` and runs `generate_voice_design()` (`6813`) with the arbiter cleared. Every waiting Kokoro synth is released (`1415-1416`) and new ones are admitted for the **entire duration of B's forward** — 10-60 s of a real VoiceDesign forward, not a microsecond race.

This is not "the fix is merely incomplete": `main.py:6756-6757` states the guarantee in its own words — *"(#2809: Kokoro stays excluded for the full design span, including the forwards)"* — and that is false for every design after the first overlapping one. It also falsifies the acceptance criterion this PR **adds** in the same diff, `onbox-acceptance-register.md` A105 bullet 4, *"A VoiceDesign forward and a Kokoro synth never co-reside on the card"*: on the box, in a bulk run, they will.

**Reachability — this is not a theoretical interleaving.** Two `/qwen/design-voice` requests can be in flight at once whenever they fall in different Node lock scopes: `withDesignLock` (`server/src/tts/design-lock.ts:26-57`, applied at `server/src/routes/qwen-voice.ts:466`) is keyed **per `bookDir`**, and the voice-library path uses a *separate* keyspace — a module-level `Set` keyed by `voiceUuid` (`server/src/routes/voice-library.ts:112-119`). So a book design running alongside a voice-library design, or designs on two different books, are not serialised at the Node layer at all. `voice-library.ts:117-118` says so explicitly, and names what is supposed to cover it: *"cross-scope protection is the frontend's single-slot + **the sidecar's own VRAM arbitration**."* The sidecar's VRAM arbitration for this pair **is** `_VD_KOKORO`, and it is what breaks here. The sidecar's capacity admission does not cover the gap either — `main.py:4899-4902` says the `yield` and release are *"deliberately OUTSIDE the lock: holding it across the caller's own forward would serialise every admitted op against every other one"*, and `4886-4889` calls it *"VRAM bookkeeping only; callers still take their own `_synth_lock`"*.

**Secondary path, narrower, same root cause — `mint_variant()`.** A mint that clears its `unload_design()` pre-wait (`main.py:7104-7106`, which normally blocks a mint behind an in-flight design because `_design_in_flight` is claimed for the whole of `design_voice`) and then blocks on `card_lock` at `7107` through a *newly-started* design's cold model load will enter `design()` during that design's forwards. On `origin/main` the same TOCTOU exists but is harmless, because acquiring `card_lock` there implied the design's `design()` had already closed. I measured this shape by transcribing `mint_variant()`'s bracket verbatim (`origin/main`: `_design_active` True and Kokoro **blocked** during the mint's own base17 forwards; head: False and Kokoro **admitted**) — but flagging honestly that my transcription omitted the `7104` pre-wait, so treat this one as **reasoned, with a narrow window**, not as the wide-open case. The design × design case above needs no such caveat.

**Fix shape — validated, not speculative.** Make the arbiter count its holders instead of latching a bool: increment a depth under `_cv` at `1433`, decrement at `1438`, and only clear `_design_active` at depth 0. I applied exactly that to a scratch copy and re-ran both probes: `_design_active` **True** throughout B's forward, Kokoro **blocked**, and the mint shape green too — plus all 53 tests across `test_base17_contention.py`, `test_design_kokoro_exclusion.py`, `test_design_contention.py`, `test_qwen_design_base17_exclusion.py`, `test_mint_variant_kokoro_stall.py`, `test_design_mint_admission.py` and `test_design_contention_routes.py` still pass. (Whether to instead make a second design *wait* rather than share the span is a design decision with more than one defensible answer — a refcount preserves today's "designs may overlap" behaviour, a gate would change it.)

**The paired test is owed and is not the one already here.** `test_two_concurrent_healthy_designs_both_succeed` builds precisely this two-overlapping-designs scenario and asserts only that both calls *return successfully* — it never looks at `_design_active`, so it passes while the arbiter is cleared under it. The regression test wanted is that one plus an assertion, sampled from inside the second design's forward, that `_design_active` is `True` and `kokoro_synth()` blocks.

---

### 🟡 Minor (cleanup)

1. **PR body's `## Test plan` is stale and describes a head that no longer exists.** It still reads *"`test_design_kokoro_exclusion.py`: 15/15 passing tests (2 skipped due to torch not available)"* and *"Regression test `test_design_voice_holds_arbiter_and_evicts_resident_kokoro` **would pass** with torch"* — that was pass 2's answer to "this PR ships no tests". This head ships two real tests in `test_base17_contention.py` and names neither. The body also doesn't carry the CLAUDE.md-mandated *"Also fixed, found in passing: …"* line for the rewritten concurrency test, although `docs/release-notes-next.md` does declare it.
2. **`main.py:6833-6840` under-enumerates its own raise paths.** The outer-`finally` comment says it *"only fires if we never got that far — `unload_base17()` or `design_context.__enter__()` raised"*. A raise from `_kokoro_eng.unload()` at `6735` also lands there with `card_lock_released` still `False` — I hit that path in the trace above and the release fired correctly. The code is right; the comment lists two of the three sources.
3. **`sys.exc_info()` at `main.py:6831` is the wrong idiom, currently harmless.** `sys.exc_info()` is the thread's *currently-handled* exception, not "the exception leaving this `try`". Measured: calling `design_voice()` from inside an active `except` block makes the **success** path hand a stale, unrelated exception to `design_context.__exit__`, which then `gen.throw()`s it into the arbiter's generator. It is benign today — `design()` is a bare `try/yield/finally`, so `_design_active` still clears and `_GeneratorContextManager.__exit__` swallows the re-raise — and unreachable in production, because `design_voice` runs via `asyncio.to_thread` where `fn()` is invoked outside any handler (I confirmed `(None, None, None)` on that path). It becomes a real defect the moment `design()` grows an `except`. Binding the exception explicitly (`except BaseException: __exit__(*sys.exc_info()); raise` / `else: __exit__(None, None, None)`), or just `contextlib.ExitStack`, costs nothing. Related nit on the same lines: `card_lock_released = True` is set *after* `card_lock.release()` (`6758-6759`); setting it first is strictly safer and free.

---

### ✅ Confirmed solid

- **Bookkeeping is complete and correct.** Release-notes pair both moved (`RELEASE_NOTES.md` user-facing line + `docs/release-notes-next.md` technical entry). A105 reworded with the inverted criterion **and** the two new ones, in the register **and** in the hand-authored live view, consistently; the register's own `*Criteria:* the five bullets above` matches the five bullets. "Check on-box acceptance register consistency" passes in CI. `Closes #2809.` present, outside any code span. (The `--against-published` pre-publish comparison and the actual publish to the recorded URL are manual steps I cannot verify from here.)
- **No deadlock, no lock-order inversion.** `card_lock` has exactly two acquisition sites (`6687`, `7107`), both taking `card_lock` → `design()` in the same order, so the non-nested *release* order is safe — release order does not create cycles, acquisition order does. `design().__enter__()` waits on `_kokoro_in_flight` while holding `card_lock`, and no Kokoro path touches `card_lock`, so that wait cannot close a cycle. `unload_design()` (`6492-6501`) waits on `_design_in_flight` **before** taking `_synth_lock`, so `design_voice` holding the claim and waiting for `_synth_lock` cannot deadlock against it; `maybe_free_idle_design` (`6533-6545`) fast-outs the same way.
- **No scope-assumption breakage among the newly-enclosed code.** The only things moved inside `design()` are the two forwards, `_synth_lock`, the two capture-guards and the `_phase()` callbacks. None reads `_design_active` (`kokoro_synth()` is its sole reader) and none re-enters the arbiter. `_phase` → `_post_progress` (`main.py:1315-1337`) is fire-and-forget on a daemon thread, so no network latency lands inside the arbiter. The two deliberate outside-the-arbiter placements (`unload_base17()` at `6718`, `unload_design()` at `7106`) are both still outside.
- **The comments this PR rewrote are accurate**, including the non-congruent-nesting note at `6677-6684`, the `#2070 R5` note at `6643-6654` and its restatement at `6709-6717`, the phase-boundary note at `6761-6766`, and the `acquire()` error string at `6689-6692` — which pass 1 correctly called wrong and which is right again now that the lock really is held only through model load.

---

### Verdict

**Request changes**, and per the re-review loop cap this goes to the human checkpoint rather than another automatic fix round.

Every finding from passes 1 and 2 is genuinely fixed — I verified each independently rather than taking the commit messages' word for it, and the two load-bearing tests are both red under their own named mutations. But the restructure that fixed them removed the implicit serialisation `card_lock` was providing for `_VdKokoroArbiter`'s boolean flag, so #2809 is fixed for a solo design and left unfixed for concurrent ones — the scenario this PR's own new test constructs, and the one the Node layer explicitly delegates to the sidecar arbiter. One paired regression test is owed (assert `_design_active` from inside the *second* design's forward), plus the three 🟡 chores.

<sub>All probes and mutations ran against copies in the OS temp directory. Worktree unmodified: HEAD `dbd5860d`, `git status --porcelain` empty, before and after.</sub>

dudarenok-maker:
## PR review — pass 4 (head a9f5c49fd5f5c326f09fe99990119deaf81835bc, depth high)

Scope: the 6-file delta since `dbd5860d` (+176/−12), re-verified against the whole `design_voice`/`mint_variant`/arbiter surface · verified: **full sidecar suite at head under the real venv** (`.venv/Scripts/python.exe -m pytest tests -m "not golden"` — **1108 collected, 2 skipped, 0 failed, exit 0**) and all **16 PR checks green**, then eight purpose-built probes, four source-mutants, and a 152-permutation N-way concurrency sweep. Everything labelled **measured** was run; everything labelled **reasoned** was not.

### Disposition of pass 3's findings — all 4 VERIFIED RESOLVED

| Finding | Disposition |
|---|---|
| 🔴 `_design_active` a bare bool → first design's exit clears the arbiter for an overlapping second | **VERIFIED RESOLVED** — measured three ways, tables below |
| 🟡 1 stale `## Test plan` / missing "Also fixed, found in passing" | **VERIFIED RESOLVED** — body now names both new tests and carries the section |
| 🟡 2 outer-`finally` comment under-enumerated its raise paths | **VERIFIED RESOLVED** — `main.py:6873-6875` now names all three (`unload_base17()`, `design_context.__enter__()`, `_kokoro_eng.unload()`); I re-hit all three paths, see the raise-path table |
| 🟡 3 `sys.exc_info()` idiom + `card_lock_released` ordering | **VERIFIED RESOLVED** — flag now set at `6786` **before** `release()` at `6787`; `sys.exc_info()` documented rather than wrapped |

---

### Measured — the refcount

**N-way concurrency sweep** (`_VdKokoroArbiter` driven directly; N ∈ {2,3,4,5} × **every** exit permutation = **152 runs**). For each: all N admitted; a real `kokoro_synth()` fired after each release and required to be **refused** while ≥1 holder remained and **admitted** after the last; count checked for negativity in-span and post-exit; final count required to be exactly 0.

```
head a9f5c49f : 152/152 CLEAN — kokoro excluded until the last design left, count landed 0 every time
bool mutant   : 152/152 FAIL  — e.g. n=3 exit (0,1,2): "KOKORO ADMITTED with 2 design(s) still holding"
                (mutant = `+= 1`→`= 1`, `-= 1`→`= 0`, i.e. the counter given bool semantics)
```

The count cannot go negative by construction: the sole decrement is in the CM generator's `finally`, which runs at most once per generator. Confirmed empirically — a double `__exit__()` on the same CM leaves it at 0, not −1 (case 10 below).

**Abnormal-exit sweep — 11 shapes, all decrement:**

| shape | count after | kokoro admitted after |
|---|---|---|
| 1 `Exception` in the `with` body | 0 | yes |
| 2 `KeyboardInterrupt` in the body | 0 | yes |
| 3 `SystemExit` in the body | 0 | yes |
| 4 manual `__enter__`/`__exit__`, clean success (the `design_voice` idiom) | 0 | yes |
| 5 manual protocol, real in-flight exception | 0 | yes |
| 6 manual protocol, success path + **stale** `sys.exc_info()` | 0 | yes |
| 7 **abandoned** CM, entered then dropped for GC | 0 | yes |
| 8 entered on a thread that ended **without** exiting | 0 | yes |
| 9 nested `design()` on one thread | 0 | yes |
| 10 `__exit__` called **twice** | 0 (not −1) | yes |
| 11 `gen.close()` — direct `GeneratorExit` at the yield | 0 | yes |

7 and 8 are the "thread death" shape you asked about: the count is not latched even there, because dropping the last reference to the `_GeneratorContextManager` finalises its generator, which throws `GeneratorExit` at the `yield` and runs the same `finally`. **There is no shape I could find that latches the count.**

**Three concurrent REAL `design_voice()` calls** (production code path, fake models, 60 real `kokoro_synth()` probes fired across the run, each recording how many design GPU forwards were in flight at the instant it won):

```
outcomes                       : ['success', 'success', 'success']
max concurrent design() holders: 3
max concurrent GPU FORWARDS    : 1        <-- the _synth_lock claim, measured
final _design_active_count     : 0
kokoro admissions              : 60, of which DURING a forward: 0
```

**Your `_synth_lock` rationale holds — I traced it rather than assuming it.** `ENGINES` is built once (`main.py:8486`), so `QwenEngine._synth_lock` (`main.py:5588`, a plain `threading.Lock`) is process-wide for Qwen. **Every** forward reachable from either entry point is inside it: `design_voice` — `generate_voice_design` (6841), `create_voice_clone_prompt` (6850), the audition `generate_voice_clone` (6965); `mint_variant` — `speech_tokenizer.decode` (7187), `create_voice_clone_prompt` (7190), `_icl_instruct_synth` (7197, whose own docstring at 6191 requires the caller to hold it), the distil (7219) and the audition (7264). **I found no forward path it does not cover.** Two caveats worth stating explicitly, neither a defect: (a) the *model loads* inside `design()` are **not** serialised by `_synth_lock` — they are serialised by `card_lock`, which both entry points hold across their loads, so the exclusion still holds but via a different lock than the comment names; (b) `design_voice`'s audition forward (6965) sits **outside** `design()` entirely — pre-existing and unchanged by this PR (it is a 0.6B Base forward, not the 1.7B VoiceDesign one the arbiter guards).

**Lost wakeups:** every mutation of either counter happens under `_cv` and is followed by `notify_all()` — there is exactly one decrement site and it notifies. A 2→1 partial drain wakes Kokoro waiters, which re-test `count > 0` and re-wait; the 1→0 decrement then wakes them again. Measured across all 152 permutations: every queued waiter eventually drained, none stuck.

**The `@property` swap is clean.** `git grep -n _design_active` over the whole tree: **zero assignments** anywhere outside the two historical plan docs under `docs/superpowers/plans/` (records of the old design, not live docs). No test, fixture, or `_configure_vd_kokoro_coupling` writes it — that function only touches `_shares_device`. The property returns a real `bool` from `count > 0`, so `test_mint_variant_kokoro_stall.py:156`'s `assert ... is False` still holds (it would have broken on a truthy `int`). Safe to read without `_cv`: a plain int compare, atomic under the GIL — exactly as racy as the bool it replaces, and its only production reader (`kokoro_synth`) reads the counter directly under `_cv`, not the property.

**`shares_device=False`** is unaffected: the early return happens before any counting, the branch is decided at `__enter__` so a mid-span flip cannot produce a mismatched decrement, and nothing but `kokoro_synth()` (which early-returns the same way) reads the count. `_design_active` is therefore always `False` on a non-sharing box — byte-identical to the bool's behaviour there.

**Regression tests are real — I mutation-tested both myself:**

- `test_kokoro_stays_excluded_until_the_last_concurrent_design_exits` → **red** against the semantic bool mutant, failing at `test_design_kokoro_exclusion.py:249` on `assert not kokoro_admitted.wait(0.5)` — the **behavioural** assertion, not an `AttributeError` and not an internal-counter assertion. Your claim is confirmed exactly as stated.
- `test_design_refcount_decrements_when_a_design_raises_mid_span` → **red** against the mutation its docstring names (decrement moved out of the `finally` onto the happy path), at `assert arb._design_active_count == 0`. It fails on the counter first and the Kokoro-admission proof second, which is on-contract for a test whose stated job is to pin the `finally`.
- No isolation hazard: both build a fresh `_VdKokoroArbiter()` (default `shares_device=True`, so the gates are live), and the autouse fixture's `_VD_KOKORO._shares_device` pin is saved/restored.

**Pass 1–3 properties re-verified, not assumed.** Span map sampled from inside the real `design_voice()`:

| phase | `card_lock` free to another thread? | `_design_active_count` |
|---|---|---|
| base17 eviction wait | **False** (held) | **0** ✅ #2070 R5 kept |
| VoiceDesign model load | **False** (held) | **1** ✅ |
| VoiceDesign GPU forward | **True** (released) | **1** ✅ #2790 + #2809 both kept |
| distil forward | **True** | **1** ✅ |
| audition forward | True | 0 (outside both — pre-existing) |

Raise-path sweep at head, 11 injected failures, each asserting `card_lock` unlocked **and** re-acquirable, count 0, `_design_in_flight` idle, and no `release unlocked lock` anywhere: happy · `unload_base17()` raises · resident-Kokoro `unload()` raises · `_ensure_design_loaded` raises · `_ensure_base_loaded` raises · `_design is None` · `_base is None` · VoiceDesign forward raises · distil forward raises · audition forward raises · `card_lock` held by another thread (bounded acquire expires). **All 11 clean.** The flag-before-release reorder opens no window: `card_lock` is a plain `threading.Lock` (`main.py:9000-9009`), whose `release()` can only raise on an already-unlocked lock, so the comment's premise is sound; nothing executes between the two statements, and `design_voice` runs on an `asyncio.to_thread` worker where signals are not delivered.

---

### 🟡 Minor (cleanup / chore) — 2 findings, neither blocking

**1. `main.py:1450-1452` — the entry-gate comment cites a test that does not pin the property it claims.**

> `# Deliberately does NOT wait on `_design_active_count` — concurrent designs are allowed (and `test_two_concurrent_healthy_designs_both_succeed` pins that); only Kokoro is excluded.`

Measured. I mutated the entry gate to `while self._kokoro_in_flight > 0 or self._design_active_count > 0:` — i.e. exactly the "serialise designs" change the comment exists to warn against — and ran the named test:

```
tests/test_base17_contention.py::test_two_concurrent_healthy_designs_both_succeed   PASSED
tests/test_design_kokoro_exclusion.py::test_kokoro_stays_excluded_until_the_last_concurrent_design_exits   FAILED
    "design B never entered the arbiter — designs must be allowed to overlap
     (the entry gate must not wait on the design count)"
```

The cited test cannot see it: with a serialising gate, design B simply waits for A's `design()` to close and then succeeds inside the test's own 10 s join. It only ever asserted "B does not time out on `card_lock`". The property **is** covered — by this PR's own new test, in the other file. So this is a wrong citation, not a coverage gap, and the fix is one identifier. It matters because the citation is load-bearing exactly when someone reaches for that change (e.g. to trade design overlap away against finding 2 below): they read the comment, run the named test, see green, and conclude the change is safe.

*Cleanup / chore (a comment the work made false).* Confidence: **certain** — mutation-run, both outcomes above are real console output.

**2. A105 bullet 3 and the `RELEASE_NOTES.md` line now contradict the overlapping-designs case bullet 4 adds in the same diff.**

`docs/testing/onbox-acceptance-register.md:3341-3346` (bullet 3, reworded earlier in this PR) tells the on-box tester:

> *Kokoro MUST pause during the VoiceDesign forward.* … a Kokoro sentence overlapping a design forward is expected to block **until that forward finishes (seconds, not minutes)** — this is the VRAM-exclusion guarantee, **not a stall**.

`:3359-3363` (bullet 4, extended in **this** commit) tells the same tester:

> Kokoro must stay blocked until the **last** design leaves, not the first.

With overlapping designs those are different observations, and bullet 3's bound is the wrong one. Measured, two design "streams" looping design → 0.15 s gap → design (the two Node keyspaces: per-`bookDir` `withDesignLock` vs. the voice-library per-`voiceUuid` set, which the pass-3 finding already established are not serialised against each other):

```
one design span          : 1.00s
kokoro waited            : 6.50s      <-- 6.5x ONE span
design stream wall-clock : 6.66s      <-- i.e. essentially the entire run
```

Because the streams' spans interleave, the refcount never reaches 0 and `kokoro_synth()`'s wait (`main.py:1433-1434`, unbounded, no timeout) lasts the whole chain rather than one forward. Scaled to a real 10–60 s VoiceDesign forward across a bulk "Design full cast" run, that is minutes, not seconds. **The code is right** — this is the unavoidable price of the exclusion you approved, and serialising designs instead would not shorten it — but two acceptance criteria in the same register row now give opposite verdicts on the same stopwatch reading, and a tester following bullet 3 would record a FAILURE for behaviour bullet 4 mandates. The same tension sits inside one `RELEASE_NOTES.md` bullet: *"will briefly wait while a design runs — a few seconds, by design, rather than a stall"* immediately followed by this PR's new *"narration now waits for the last design to finish, not the first."*

Reconciling is a wording fix in three files (register, live view, `RELEASE_NOTES.md`): bullet 3's parenthetical should bound the pause to *one design span per design still holding*, not "seconds, not minutes".

*Reasoned, not measured:* the 10-min `SYNTH_CALL_TIMEOUT_MS` (`server/src/tts/synthesise-chapter.ts:167`) and 12-min `CHAPTER_NO_PROGRESS_MS` (`server/src/routes/generation.ts:288`) are the ceilings a long enough starve would trip, turning it into a recorded `generationError`. I did **not** reach either — I am not claiming a chapter failure, only that the criterion's stated bound is now wrong and the ceilings exist.

*Cleanup / chore (an acceptance criterion the work made self-contradicting).* Confidence: **high** — the 6.5× figure is measured; the production scaling is arithmetic on the register's own stated forward durations. I considered 🟠 and rejected it: passes 1 and 2 both classified "A105 criterion is now wrong" as 🟡, the code is correct, and the on-box run is not blocked from producing a verdict, only from producing an unambiguous one.

---

### ✅ Also confirmed solid

- **Bookkeeping complete.** Release-notes pair both moved (`RELEASE_NOTES.md` user-facing line + `docs/release-notes-next.md` technical entry, the latter now describing pass 3's finding accurately). A105 bullet 4 extended **identically** in the register and in the hand-authored live view. `Closes #2809.` present, outside any code span; "Verify PR body links a GitHub issue" and "Check on-box acceptance register consistency" both green. The `## Also fixed, found in passing` section is present and matches the diff. (The manual `--against-published` comparison and the publish to the recorded URL remain steps I cannot verify from here.)
- **`## Test plan` is now accurate** — it names both new tests, states the mutant each was verified against, and no longer describes a head that has moved.
- **No lock-order inversion, no deadlock introduced.** `card_lock` still has exactly the two acquisition sites (`6710`, `7146`), both ordering `card_lock` → `design()`. The refcount adds no new wait: `design()`'s entry gate still waits only on `_kokoro_in_flight`, and no Kokoro path touches `card_lock`.
- **Comments this round rewrote are accurate** — the refcount rationale at `1406-1417` (including its `_synth_lock` claim, measured above), the flag-ordering note at `6780-6785`, the `sys.exc_info()` note at `6858-6868` (both callers of `design_voice` reach it via `asyncio.to_thread` at `10944`/`10957`, and probe case 6 shows the stale-exc_info path is harmless anyway), and the three-source outer-`finally` note at `6871-6881` — with the single exception of finding 1's test citation.

---

### Verdict

**Approve with two 🟡 chores.** No 🔴 and no 🟠 this round: the pass-3 defect is genuinely fixed, and I could not break the refcount — not with 152 exit permutations across 2–5 concurrent holders, not with eleven abnormal-exit shapes including GC-finalisation and double-exit, not through three real `design_voice()` calls on live code. Both new tests are red under the mutations their docstrings name, and the concurrency one is red on its behavioural assertion exactly as claimed. Every property passes 1–3 established still holds under direct measurement.

The two chores are a wrong test citation in a comment and an acceptance criterion this round made self-contradicting; per the repo's report-fix-record rule both are fixed in this round, and neither re-triggers a review pass.

<sub>All probes, mutants and scratch copies lived in the OS temp directory. Worktree unmodified: HEAD `a9f5c49fd5f5c326f09fe99990119deaf81835bc`, `git status --porcelain` empty, before and after.</sub>

dudarenok-maker:
## Pass 4 triage — both 🟡 chores fixed in `3f89796f`

Pass 4 returned **approve, no 🔴/🟠**. Both minor findings are now closed; no findings deferred.

**🟡 1 — entry-gate comment cited a test that does not pin the property.** Fixed, and I re-ran the reviewer's mutation myself rather than taking the report on trust. Mutating the gate to `while self._kokoro_in_flight > 0 or self._design_active_count > 0:` (serialising designs):

| test | under the gate mutation |
|---|---|
| `test_two_concurrent_healthy_designs_both_succeed` (the old citation) | **passes** — its two designs simply queue and both still succeed |
| `test_kokoro_stays_excluded_until_the_last_concurrent_design_exits` (this branch's) | **fails**, on its `b_entered` assertion |

The comment now cites the test that actually catches it and records why the other one cannot — the citation is load-bearing exactly when someone reaches for that change.

**🟡 2 — A105 bullet 3 contradicted bullet 4.** Fixed across all three surfaces. Bullet 3 promised the pause lasts "until that forward finishes (seconds, not minutes) — not a stall"; bullet 4 requires Kokoro to stay blocked until the **last** design leaves. With back-to-back designs the refcount never reaches zero and the wait is unbounded, so a tester following bullet 3 would have recorded a FAILURE for behaviour bullet 4 mandates. Reworded so the failure condition is Kokoro proceeding **concurrently** with a design — never the length of the wait — with the real escalation ceilings named (`SYNTH_CALL_TIMEOUT_MS` 600 s, `CHAPTER_NO_PROGRESS_MS` 720 s). Same over-promise removed from the `RELEASE_NOTES.md` line. `check:onbox-register` green.

**Review history:** four passes, three of which each found a genuine blocking concurrency bug in this ~40-line region — the card_lock-across-forwards regression (pass 1), the card_lock leak on the eviction-raise path (pass 2), and the bool `_design_active` cleared by whichever design exits first (pass 3). Pass 4 verified all of them resolved, including a 152-permutation sweep over N ∈ {2,3,4,5} concurrent holders (152/152 fail against a bool-semantics mutant) and 11 abnormal-exit shapes. The refcount design was chosen by the repo owner over serialising designs.

PR #2822, opened 2026-08-31
fix(sidecar): hold the VD/Kokoro arbiter across design_voice's GPU forwards

### LINKED ISSUE #2809: sidecar: design_voice's GPU forward runs outside _VD_KOKORO.design(), letting Kokoro synth race it

dudarenok-maker:
## Summary

`design_voice()` closed its `with _VD_KOKORO.design():` block right after the model load, so `_design_active` — the flag `KokoroEngine.synthesize` blocks on — was already clear by the time the two GPU forwards (`generate_voice_design()` and `create_voice_clone_prompt()`) ran. A Kokoro synth could therefore co-reside with a VoiceDesign forward on the same 8 GB card: exactly the contention the arbiter exists to prevent, and the reverse of what its own surrounding comments assert ("`_design_active` for its ENTIRE span"). `mint_variant()` already holds the arbiter across its own `_synth_lock` forward block, so the two paths disagreed.

A regression from #2790, which correctly narrowed `card_lock` to end at model load and collapsed the arbiter block to the same window along with it.

**The fix.** `card_lock` now releases **inside** the arbiter block at the load boundary, rather than in the enclosing `finally`. #2790's narrowing survives — a concurrent `mint_variant()` can still load base17 during the design's forwards — while the arbiter stays held through them. The lock is still *acquired* outside the arbiter, which preserves the global `card_lock → arbiter` order that `mint_variant()` takes at its own combined `with`; acquiring it inside would invert that order between the two paths.

**Also fixed, found in passing** (`_VdKokoroArbiter`, `main.py`): design spans were tracked in a bare bool. `design()` drains Kokoro but deliberately does *not* exclude another design, so two spans overlap routinely — this change makes `design_voice` hold one for its whole forward, and `mint_variant` already did — and the first span to exit cleared the flag out from under the second, readmitting Kokoro into a still-running forward. Now a depth counter, with `_design_active` kept as a read-only property over it so existing readers are unaffected. This is the same defect class as the headline bug, reached by a different path, so it is fixed here rather than filed.

### On the deadlock reported in the issue

[The issue comment](https://github.com/dudarenok-maker/Castwright/issues/2809#issuecomment-5472724820) records an attempt at this fix that hung. **I could not reproduce that hang** — not with the shape landed here, and not with the shape described in the comment (which additionally held `card_lock` through the forwards). Two full `npm run test:sidecar` runs are green, plus the contention suites run separately against both shapes. I am not claiming the `card_lock` difference *was* the cause; I have no evidence for a mechanism, and that attempt sat on a branch carrying unrelated Kokoro-CUDA changes.

What I did do is the concurrency-safety pass the comment asked for, and it comes back clean on all three axes:

- **No same-thread re-entry into `kokoro_synth()`.** The only calls inside the widened span are `_phase()` (a `_post_progress` HTTP POST back to the Node server), the two `_ensure_*_loaded()` calls, and the two Qwen forwards. None reaches Kokoro, and `design_voice` runs on an `asyncio.to_thread` worker, so there is no event-loop re-entry either.
- **Consistent lock order.** `design_voice`: `card_lock → arbiter → _synth_lock`. `mint_variant`: `card_lock → _base17_activity → arbiter → _synth_lock`. Same order; they cannot invert.
- **No cycle through `_synth_lock`.** `QwenEngine.synthesize` takes `_synth_lock` but never the arbiter (pinned by an existing test), and both `unload_design()`/`unload_base17()` bounded waits stay *outside* the arbiter, as #2070 R5 requires — so the 150s-wait hazard that comment guards against is not moved inside.

## Test plan

- `server/tts-sidecar/tests/test_design_kokoro_exclusion.py::test_design_voice_holds_arbiter_and_evicts_resident_kokoro` — pre-existing, **failed deterministically on `main`** (`AssertionError: design forward must run under arb.design()`), passes here. This was blocking `npm run test:sidecar`, and so every push touching `server/tts-sidecar/**`.
- `test_overlapping_designs_keep_kokoro_blocked_until_the_last_one_exits` — **new**, red before the depth-counter fix for the right reason (`the first design's exit cleared the shared flag while a second design was still holding the arbiter`), green after.
- `npm run test:sidecar` — full suite, **two independent runs, both exit 0**, no hang.
- Contention suites (`test_base17_contention`, `test_design_contention`, `test_design_contention_routes`, `test_qwen_design_base17_exclusion`, `test_mint_variant_kokoro_stall`, `test_design_mint_admission`, `test_kokoro`) — 114 passed.
- `npm run check:onbox-register` — OK.

## On-box acceptance

Recorded as new register row **A106**. The unit suite proves the *mechanism* (the flag is set across the forwards; `kokoro_synth()` blocks on it). It cannot prove the *purpose* — that the two engines genuinely stop sharing the 8 GB card — because every automated test here mocks the models. That needs VRAM measured on the box, same reason A105 (#2790, the PR this regressed out of) carries a row. Register, live view and the owed total (66 → 67) all move in this PR.

Closes #2809

### DIFF
--- a/server/tts-sidecar/main.py
+++ b/server/tts-sidecar/main.py
@@ -1403,8 +1403,21 @@ class _VdKokoroArbiter:
     def __init__(self, shares_device: bool = True) -> None:
         self._cv = threading.Condition()
         self._kokoro_in_flight = 0
-        self._design_active = False
+        # Depth, not a bool: `design()` drains Kokoro but deliberately does NOT
+        # exclude another design, so two spans overlap routinely — `design_voice()`
+        # holds one across its GPU forwards (#2809) while `mint_variant()` holds
+        # one across its own. A bare flag let the FIRST span to exit clear it out
+        # from under the second, admitting a Kokoro synth into the still-running
+        # forward (the very co-residency this class exists to prevent). Counting
+        # makes the flag clear on the LAST exit.
+        self._design_depth = 0
         self._shares_device = shares_device
+
+    @property
+    def _design_active(self) -> bool:
+        """True while ANY design span is open. Read-only view over the depth
+        counter (deliberately has no setter — the depth is the source of truth)."""
+        return self._design_depth > 0
 
     @contextmanager
     def kokoro_synth(self):
@@ -1430,12 +1443,15 @@ class _VdKokoroArbiter:
         with self._cv:
             while self._kokoro_in_flight > 0:
                 self._cv.wait()
-            self._design_active = True
+            self._design_depth += 1
         try:
             yield
         finally:
             with self._cv:
-                self._design_active = False
+                self._design_depth -= 1
+                # Notify unconditionally: waiters re-check `_design_active` in
+                # their own loop, so a wake-up while another span is still open
+                # is a no-op rather than a lost wake-up if depth just hit 0.
                 self._cv.notify_all()
 
 
@@ -6674,8 +6690,13 @@ class QwenEngine(Engine):
                 # mint_variant() can reload base17 after VoiceDesign load finishes,
                 # since VoiceDesign is then resident and doesn't compete for VRAM
                 # with base17 (VoiceDesign is freed after design completes).
-                # The lock nests outside the _VD_KOKORO.design() block (not inside)
-                # to avoid interfering with the Kokoro arbiter.
+                # The card lock is acquired OUTSIDE the _VD_KOKORO.design() block but
+                # released INSIDE it (see the nested finally below), so the arbiter
+                # stays held through the GPU forwards while the card lock still ends
+                # at model load. Acquiring it outside preserves the global order
+                # card_lock -> arbiter that mint_variant() takes at its single
+                # combined `with` (line ~7052); acquiring it inside would invert that
+                # order between the two paths and deadlock them against each other.
                 # The acquire is bounded (not unbounded) to prevent indefinite hangs
                 # if a concurrent mint_variant() holds the lock for an extended period.
                 card_lock = _DEVICE_LEDGER.card_lock(_qwen_configured_card_idx())
@@ -6685,6 +6706,7 @@ class QwenEngine(Engine):
                         f"base17 load/mint or another design's model load has held it for over "
                         f"{_BASE17_CONTENTION_WAIT_S_DEFAULT:.0f}s. Retry the design shortly."
                     )
+                card_lock_released = False
                 try:
                     if self._base17 is not None or self._base17_in_flight.busy:
                         log.info("Evicting resident/in-flight Qwen 1.7B-Base to free VRAM for VoiceDesign load.")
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
Closing as a duplicate — this was already fixed and merged in #2810 (`4224957e`), which closed #2809.

My branch was cut from `7ce7105b` without re-fetching, so it predated #2810's merge; the deterministic `test_design_voice_holds_arbiter_and_evicts_resident_kokoro` failure I reproduced was my stale base, not `main`. Nothing here is needed: #2810 landed the same two fixes (widen `_VD_KOKORO.design()` across the GPU forwards while releasing `card_lock` at the model-load boundary; replace the arbiter's bare bool with a refcount exposed as a read-only `_design_active`) and, through three review rounds, three more this PR does not have — the `card_lock` leak when `unload_base17()` raises `Base17ContentionTimeoutError`, the discarded exception in `design_context.__exit__`, and a repair to `test_two_concurrent_healthy_designs_both_succeed`, which could not fail as written.

No register row is owed from here either: #2810 reworded and extended A105 rather than minting a new row, so the A106 row on this branch would have been a duplicate.

Issue #2199, opened 2026-08-06
check:onbox-register --against-published fails on every row discharge, and its documented remedy cannot clear it


dudarenok-maker:
## What

`npm run check:onbox-register -- --against-published <file>` fails on **every legitimate row discharge**, and its documented remedy cannot clear the failure.

The check is one-directional by design: it fails only when the live page has a row the register does not, which is meant to catch "another lane published ahead of you". But **removing a row is exactly that shape too** — after discharging a row and merging, the currently-published page still contains it, so the comparator reports the register as BEHIND.

## Observed 2026-08-06

Discharging Group C's C1 (plan 261) and C2 (plan 247) and renumbering C3 → C1 produced:

```
The live page's Group C section has rows C2, C3 that this register does not yet have
 — the register is BEHIND what is already published. Merge them in before publishing.
Do not publish.
```

Every other group matched exactly (A 37, B 4, D 2, E 8, F 1, G 2), so there was no competing lane — the only "missing" rows were the two just discharged plus the renumbered survivor.

**The documented remedy is a no-op here.** The register's step 3 says to pull latest `main`, confirm the no-flag check is green, and re-run against the same saved copy. That was done — `main` already contained the merge and the no-flag check was green — and the comparator still fails, because the discrepancy is not a missing merge. There is no state in which it passes short of publishing first, which is the thing it gates.

## Why it matters

The operator is instructed "**If it fails, do NOT publish**". Following that literally means a discharge can never be published, so the live view silently rots while the register moves. Following it non-literally trains the operator to override the one check standing between them and the #1931 stale-snapshot race — which is worse, because the next failure might be a genuine one.

## Suggested direction (not prescriptive)

Distinguish "the live page has a row I lack **and my register never had it**" from "the live page has a row I **deliberately removed**". Options:

- Compare against the merge-base register rather than only the working copy, so a row deleted in this diff is recognised as intentional.
- Let the caller declare discharged IDs (e.g. `--discharged C1,C2`) and subtract them before the comparison.
- Track discharged rows in a small ledger the check reads.

Renumbering interacts with this: contiguity enforcement means discharging C1/C2 renames C3 → C1, so identity is positional, not stable. Whatever shape the fix takes has to tolerate a surviving row changing its ID.

## Acceptance

- A discharge + renumber like the 2026-08-06 one passes `--against-published` against the pre-merge live page.
- A genuine competing-lane row still fails it.
[...]

PR #2835, opened 2026-08-31
fix(server): normalise admitted rocm:N device keys to cuda:N before .to()

### LINKED ISSUE #2813: ROCm/AMD device-key normalisation gap breaks GPU engine loads
## What happened

Found during PR #2797's second fresh-review pass (finding N6): on a real ROCm/AMD box, `admit()` (the VRAM-ledger admission layer, `server/tts-sidecar/main.py`) returns a device dict like `{"device": "rocm:N"}`. `_resolve_torch_device` passes this value through unchanged (it only resolves the `"auto"` case), and `_load_qwen_model` (and presumably other engine-load call sites) then calls `inner.to("rocm:0")` directly on the resulting string.

Measured directly against the module: `torch.device('rocm:0')` raises — PyTorch's ROCm/HIP build masquerades as CUDA at the API level (`torch.device` only understands `"cuda:N"`, not `"rocm:N"`), so this call fails.

**Impact**: with `SEG_CAPACITY_ADMISSION` on (the default), every GPU engine load on an AMD box fails, because the admission layer hands engines a device string the actual `torch.device()` call can't consume.

## Why this is pre-existing, not introduced by PR #2797

PR #2797 added `_qwen_resident_device_key`/`_parse_device`'s ROCm-awareness (for correctly reading *back* a resident engine's device for capacity-wait purposes), but never touched the *admission→load* path this issue is about — `admit()`, `_resolve_torch_device`, or the `inner.to(device)` call sites. This bug exists independent of and prior to that PR's device-key work.

## Decision owed

Where should the `"rocm:N"` (admission's device-key format) ↔ torch-consumable device-string boundary live? A few shapes worth considering, none picked here:

1. Normalise at the `admit()`/admission-layer boundary — return a torch-consumable string (e.g. `"cuda:N"` even on ROCm, matching what torch's own API expects) rather than the family label admission uses internally.
2. Normalise at each engine's `_load_*` call site, right before `.to(device)` — convert `"rocm:N"` → `"cuda:N"` there specifically.
3. Normalise once in `_resolve_torch_device`, treating a `"rocm:N"` input the same way it already special-cases `"auto"`.

This needs a decision because there may be other ROCm-format device strings flowing through the same admission layer to non-Qwen engines (Coqui, Kokoro, ASR) that would need the same fix — worth surveying all `.to(device)` call sites once the normalisation boundary is chosen, so the fix isn't Qwen-only.

## Evidence

Reproduced by direct module inspection during PR #2797's review (not on a live ROCm box — this repo currently has no ROCm/AMD hardware in its on-box acceptance rotation, per `docs/testing/onbox-acceptance-register.md`'s existing ROCm-labeled rows). A real ROCm box run is needed to confirm the exact failure shape/message once a fix lands.

Related: PR #2797 (closes #2678), `server/tts-sidecar/main.py`'s `_qwen_resident_device_key`/`_parse_device`/`_cuda_is_rocm`.

dudarenok-maker:
## Summary

The VRAM-ledger admission layer (`PlacementController.admit()`) hands engines a device key like `"rocm:N"` on a ROCm/AMD box — `probe()`'s own honest ROCm-vs-CUDA accounting vocabulary, used elsewhere in the ledger for reporting. A ROCm/HIP torch build only ever understands `"cuda:N"` at the API level (`torch.device("rocm:N")` raises, and CTranslate2/speechbrain likewise have no `"rocm"` device), so an admitted `"rocm:N"` flowing unconverted into a real device call crashed every GPU engine load on a real AMD box whenever `SEG_CAPACITY_ADMISSION` is on (the default).

**Original fix:** `_resolve_torch_device` (`server/tts-sidecar/main.py`) converts an explicit `"rocm:N"` to `"cuda:N"` (index preserved) before it reaches Coqui/Qwen's `.to()` call, mirroring the existing reverse-direction conversion already coded in `_qwen_resident_device_key`. Coqui's `_resolve_runtime_options` calls `_resolve_torch_device` unconditionally instead of only for `"auto"`, so an explicit admitted `"rocm:N"` actually reaches the branch instead of bypassing it. Qwen's three load paths already routed every admitted device through the resolver — confirmed clean, no change needed.

**PR review pass 1 found two more correctness bugs in the same class, both fixed in this PR:**

1. **`/transcribe` and `/embed` took the identical admitted `"rocm:N"` and handed it straight to CTranslate2/speechbrain, unconverted.** `_ct2_kwargs`, `_spk_run_device`, `WhisperEngine._ensure_loaded`, and `SpeakerEngine.ensure_loaded` now all route through a new shared `_normalise_rocm_device_key` helper (extracted from `_resolve_torch_device`'s own conversion) before a real device call. `SpeakerEngine`'s cuda-presence guard and cpu-demotion fallback were also silently skipped for a raw `"rocm:N"` (both gated on `family == "cuda"`) — fixed by normalising `self.device` at the same chokepoint the admitted `device=` param is applied. `server/src/config/registry.ts`'s stale "unreachable dead code" comment (which this PR is direct proof against) is corrected.
2. **An operator/UI device pin could only ever be `"cuda:N"` (config validation rejects `"rocm:N"` outright), which could never match the ledger's own `"rocm:N"`-keyed candidates — every pinned engine got permanent `noCapacity`, and a resident engine's OWN next admission had the identical mismatch.** `_engine_env_pin` and `_is_resident` now re-tag their output back to `"rocm:N"` via a new shared `_report_rocm_device_key` helper (the same conversion `_qwen_resident_device_key` already did for reporting) before `admit()`/`reservation()` compare it against a probed candidate. The operator-facing knob still only ever accepts `"cuda:N"` — unchanged — the fix lives entirely sidecar-side.

Also adds/updates an on-box acceptance register row (this box is dual-NVIDIA, no ROCm/AMD hardware to prove any of this on) and both release-notes files, all updated again in the review-pass-1 round to reflect the widened scope.

Kokoro is deliberately still out of scope: its ORT-provider path has no GPU placement for a ROCm pin to reach on this repo's current `accelerator-profile.mjs` (CPU-only AMD profile), so the class doesn't apply there today.

**PR review pass 2 found a third instance of the identical two-vocabulary mismatch, fixed in this PR:**

3. **`_same_card`'s idle-eviction card comparison had the same mismatch as findings 1/2, one layer down.** The idle-eviction ladder compares a candidate `resident` engine's device key against the `target` device the incoming request needs to fit on — `target` is always probe-native (`"rocm:N"` on a ROCm box), but `_same_card` split both sides with the raw `family == "cuda"` check, so it silently reported every resident-vs-target pair as a different card, and the whole idle-eviction ladder ran with zero candidates on a ROCm box. Fixed by routing both `resident` and `target` through the same `_normalise_rocm_device_key` helper (findings 1/2's own idiom) before the family/index split.

Also fixes three minor findings from pass 2: `registry.ts`'s corrected `probe_capacity` citation was itself off by one function boundary (now `main.py:4041`/`:4050`); the identical stale comment + drifted citations in `server/src/config/resolver.test.ts` (a near-duplicate of the `registry.ts` block, missed in pass 1) rewritten to match; `_is_resident`'s docstring corrected to name Kokoro as a live exception that does not route through the forward-normalisation helper (harmless today only because the reverse helper is idempotent on an already-`"rocm:N"` input); both shared helpers' "returned unchanged" docstring contracts corrected for unindexed input (a bare `"rocm"`/`"cuda"` is not passed through — it becomes index 0).

Closes #2813

## Test plan

Original fix:
- [x] `server/tts-sidecar/tests/test_qwen_device.py::test_rocm_normalises_to_cuda` (parametrized `rocm:0`→`cuda:0` / `rocm:1`→`cuda:1`) — confirmed red before the fix, green after
- [x] `server/tts-sidecar/tests/test_coqui_device.py::test_admitted_rocm_device_normalises_to_cuda` — confirmed red before the fix, green after

Review pass 1 fixes (finding 1 — ASR/SPK crash):
- [x] `test_device_parse.py::test_ct2_kwargs_normalises_admitted_rocm`, `::test_spk_run_device_normalises_admitted_rocm` — confirmed red before the fix, green after
- [x] `test_transcribe_embed_admission.py::test_asr_ensure_loaded_normalises_admitted_rocm_device`, `::test_spk_ensure_loaded_normalises_admitted_rocm_device_and_the_guard_fires` (the latter proves the cuda-presence guard actually FIRES post-fix, not just that the final device string looks right) — confirmed red before the fix, green after
- [x] `test_coqui_device.py::test_admitted_rocm_device_override_normalises_to_cuda` (drives `device_override=`, the actual admitted-device parameter — the pre-existing tests only drove the env-var path, per finding 4)

Review pass 1 fixes (finding 2 — pinned/resident mismatch):
- [x] `test_placement.py::test_engine_env_pin_retags_to_rocm_on_a_rocm_box`, `::test_engine_env_pin_stays_cuda_on_a_real_cuda_box` (control) — confirmed red before the fix, green after
- [x] `test_placement.py::test_admit_pinned_rocm_box_actually_admits` — end-to-end repro mirroring the reviewer's own simulated-ROCm-probe repro (pin `cuda:0` + probe keyed `rocm:0`/`rocm:1`, 22 GB free — confirmed `noCapacity` before the fix, `{"device": "rocm:0"}` after)
- [x] `test_devices.py::test_is_resident_retags_to_rocm_on_a_rocm_box`, `::test_is_resident_unknown_fallback_retags_to_rocm_on_a_rocm_box`, `::test_is_resident_stays_cuda_on_a_real_cuda_box` (control) — confirmed red before the fix, green after

Review pass 2 fix (finding 7 — idle-eviction card mismatch):
- [x] `test_devices.py::test_same_card_matches_a_rocm_native_target` (unit) — `("cuda:0","rocm:0")`/`("cuda:1","rocm:1")` now match, `("cuda:0","rocm:1")` still correctly does not, Kokoro's own `("rocm:0","cuda:0")` exception still matches, `("rocm:0","rocm:0")` still matches, `cpu`/`None` controls unaffected — confirmed red before the fix (both `cuda:N`-vs-`rocm:N` cases false), green after
- [x] `test_devices.py::test_idle_evict_frees_engines_against_a_rocm_native_target` (integration) — mirrors the reviewer's own repro: admitting against a `"rocm:0"` target now actually evicts the idle same-card engine instead of finding zero candidates — confirmed red before the fix, green after

- [x] Full run of all six touched test files (`test_device_parse.py`, `test_devices.py`, `test_placement.py`, `test_transcribe_embed_admission.py`, `test_coqui_device.py`, `test_qwen_device.py`) — 148 passed, 0 failed
- [x] Broader sweep of every other test file referencing the changed functions (`test_asr_spk_idle_evict.py`, `test_design_kokoro_exclusion.py`, `test_device_probe.py`, `test_engine_import_ok.py`, `test_memory.py`, `test_smoke.py`, `test_speaker_embed.py`, `test_transcribe.py`) — 188 passed, 0 failed, no regressions
- [x] `npm run test:sidecar` — SKIP (this worktree's sidecar venv was never bootstrapped); all pytest runs above were via the primary checkout's venv against this worktree's code instead
- [x] `npm run check:onbox-register` — OK, register and live-view agree
- [x] `npm run verify:fast:branch` — green (lint/typecheck/config:check/check:budget-poll/test/audit/audit:server/build all out-of-scope for this diff; test:hooks + test:server cached; test:sidecar SKIP-pass)
- [x] On-box acceptance register row and matching live-view HTML content updated to cover the widened scope (pinned devices, ASR/SPK) — publish-token re-stamped
- [ ] **Live-view publish to its canonical artifact URL is still owed** — flagging honestly per review finding 6 rather than marking this done. The register/HTML content in this branch is committed and internally self-consistent (verified via `check:onbox-register`), but checking the currently-live page found it already carries content this branch does NOT have (a "wave 10" set of changes discharging row A28, owed=65 vs. this branch's 66) — and `check:onbox-register -- --against-published` confirmed `origin/main` doesn't have that content either, meaning another lane published it from a branch not yet merged to `main`. Publishing from this branch would silently clobber that lane's live page, a scenario the register's own docs flag as a hazard the mechanical check cannot catch on its own. Deliberately deferred rather than guessed through; needs a publish once that other lane's branch merges (or a manual reconcile) rather than a blind overwrite here.
- [x] Release notes: `docs/release-notes-next.md` and `RELEASE_NOTES.md` both updated (again, in the review-pass-1 round, to describe the ASR/SPK and pinned-device fixes too — the original "every graphics-card voice load fails outright" claim is now fully accurate rather than only true for the unpinned case, per review finding 2)

Also fixed, found in passing: `server/src/config/registry.ts`'s stale comment claiming the `fam in ("cuda","rocm")` checks in `main.py` were unreachable dead code — this PR is direct proof they are not; corrected with accurate line references. Pass 2 additionally fixed: that same citation was itself pointing at the wrong line (`main.py:4041`/`:4050`, not `:4032`); the identical stale comment duplicated in `server/src/config/resolver.test.ts` (missed in the pass-1 fix); and two docstrings (`_is_resident`, the two shared normalisation helpers) that overclaimed their own behaviour.


dudarenok-maker:
## PR review — pass 1 (head d55927fc86d693e740991b8d62807f2eb12286b3, depth medium)

The fix itself is correct and I could not break it: index preservation holds, the `"auto"` path and every other explicit value are genuinely unchanged, and both new tests are red before the fix for the reason claimed. Verified by extracting `origin/main`'s `main.py` into a temp dir and running the two new assertions against it:

```
A1 _resolve_torch_device('rocm:0') -> 'rocm:0'  expected 'cuda:0'  FAIL
A1 _resolve_torch_device('rocm:1') -> 'rocm:1'  expected 'cuda:1'  FAIL
A2 CoquiEngine._resolve_runtime_options -> {'device': 'rocm:0', 'half': False, 'deepspeed': False}  FAIL
```

and green after (`22 passed`). The PR body's claim that Qwen's three load paths are already clean is **true** — I checked all three (`main.py:6011-6013`, `:6033-6035`, `:6453-6455`); each sets `_device_pref` then calls `_ensure_device_resolved()` → `_resolve_torch_device`. I also enumerated all ten consumers of an admitted device key (`git grep 'adm\["device"\]'`): eight are now covered, two are not.

Two correctness findings, both in the class this PR set out to close.

---

### 🟠 Significant — correctness

**1. `/transcribe` and `/embed` take the same admitted `"rocm:N"` and hand it straight to CTranslate2 / speechbrain — the class is left armed, and the repo has already documented that it crashes**

`server/tts-sidecar/main.py:11905` (`ASR.transcribe(..., adm["device"])`) and `server/tts-sidecar/main.py:11981` (`SPK.ensure_loaded(adm["device"])`) receive the identical `PlacementController.admit()` output this PR exists to normalise, and neither routes it through `_resolve_torch_device`. Run against the branch:

```
_ct2_kwargs('rocm:0','int8')          -> {'device': 'rocm', 'compute_type': 'int8'}
_spk_run_device('rocm:0')             -> rocm
WhisperEngine._compute_type('rocm:0') -> int8        # not int8_float16
torch.device('rocm:0')                -> RAISES RuntimeError: Expected one of cpu, cuda, ...
```

Concrete scenario, ROCm box, `SEG_CAPACITY_ADMISSION` on (default), `ASR_DEVICE=cuda` / `SPK_DEVICE=cuda` (unindexed, so `_engine_env_pin` returns `None` and admission is unconstrained): `on_gpu` is True (`:11891`, `:11970`), admission returns `{"device": "rocm:0"}`, and then

- `WhisperEngine._ensure_loaded` (`:8149`) calls `WhisperModel(device="rocm")` → CT2 has no such device → every `/transcribe` 500s. The compute type also silently degrades to the CPU `int8` (`:8098-8099`).
- `SpeakerEngine.ensure_loaded` sets `self.device = "rocm:0"` (`:8434`); the cuda-presence guard at `:8438` is gated on `family == "cuda"` so it is **skipped**, `_spk_run_device` yields `"rocm"`, `EncoderClassifier(run_opts={"device":"rocm"})` raises — and the cpu-demotion fallback at `:8453` is gated on `family == "cuda"` too, so it is **also skipped** and the bare `raise` fires. Every `/embed` 500s with no fallback.

This is not speculative: `server/src/config/registry.ts:405-414` already describes both crashes in exactly these words ("`SPK_DEVICE=rocm` crashes `EncoderClassifier(run_opts={"device":"rocm"})` … the demote path is gated on `family=="cuda"` … with NO cpu fallback — every `/embed` 500s"; "`ASR_DEVICE=rocm` crashes `WhisperModel(device="rocm")` … CTranslate2 has no 'rocm' device at all"). The only reason it was judged unreachable is the claim at `registry.ts:400-404`:

> Nothing upstream of `_parse_device` ever PRODUCES family "rocm" from a real input … so that arm is unreachable dead code.

**This PR is the proof that sentence is false** — `admit()` produces it and hands it to engines. So "unconfirmed to share this bug shape" is settled by reading, not owed to hardware: they share it, the failure is already characterised, and the stated mitigation no longer holds. The fix and the stale rationale in `registry.ts` belong in the same round. (While editing that comment: its `main.py` line citations — `:9100-9103`, `:9124-9134`, `:3367`, `:7540`, `:10712`, `:10787` — have all drifted off their targets.)

**2. On a ROCm box an operator/UI device pin can only ever be `"cuda:N"`, which never matches the ledger's `"rocm:N"` keys — every pinned engine gets permanent `noCapacity`**

The mirror image of the bug this PR fixes: the same two-vocabulary split, at the admission seam instead of the `.to()` seam.

- `PlacementController._device_key` (`main.py:4463`) builds keys from `probe_capacity()`'s `kind`, which is `"rocm"` on a HIP build (`:4041`) → `"rocm:0"`, `"rocm:1"`.
- `_engine_env_pin` (`:3916-3918`) preserves whatever family the env holds, and `_resolve_uuid_to_index` (`:3887`) **hardcodes** `f"cuda:{card['idx']}"`, so a UI-picked card always resolves to `"cuda:N"`.
- `server/src/config/registry.ts:369` / `:427` validate the device knobs against `/^(cpu|auto|cuda|cuda:\d+)$/i` — `rocm:N` is **rejected by config validation**, so `cuda:N` is the *only* indexed pin an operator can legally set.
- `admit()` sets `constraint = resident if resident is not None else pinned` (`:4650`) and `_gpu_candidates` filters `c[0] == resident_or_pinned` (`:4479`).

Reproduced against the branch with a simulated ROCm probe (two 24 GB cards, 22 GB free each):

```
ledger device keys from probe: ['rocm:0', 'rocm:1', 'cpu']

COQUI_DEVICE='cuda:0'  -> pin='cuda:0' candidates=[]                    admit={'noCapacity': {'neededMb': 3584, 'deviceKey': 'cuda:0'}}
COQUI_DEVICE='rocm:0'  -> pin='rocm:0' candidates=[('rocm:0',22000,24000)] admit={'device': 'rocm:0'}

# and the uuid form, on a real box:
_resolve_uuid_to_index('cuda-uuid:GPU-abc') -> 'cuda:1'
```

44 GB free and it refuses, with a `deviceKey` that is not in the probe at all. It never recovers, because the engine never becomes resident, so `constraint` stays `pinned` forever. The row that matters is the first: it is the only one an operator on AMD hardware can actually produce.

This also makes the PR body and `docs/release-notes-next.md:245` over-claim. "that would have made every graphics-card voice load fail outright. Fixed at the source" is true only for the *unpinned* case; with a device pin set, the load still fails, one seam earlier and with a different symptom (503 `noCapacity` rather than a raise). The register row's acceptance criterion is still satisfiable — it describes an unpinned load — but the prose should not claim the ROCm GPU path is fixed while a pinned one is not.

---

### 🟡 Minor — cleanup

**3. `_qwen_resident_device_key`'s docstring is now false, and says it was verified**

`server/tts-sidecar/main.py:5146-5150` states: "`_resolve_torch_device` returns an explicit (non-'auto') pref UNCHANGED … Verified against `_resolve_torch_device`'s own source, not assumed." This PR is precisely the change that makes an explicit pref *not* return unchanged. The function's own behaviour is still correct (`_device` is now `"cuda:N"`, `_parse_device` → `cuda`, then `_cuda_is_rocm()` re-tags to `"rocm:N"` at `:5170` — same output), but the recorded reasoning now asserts the opposite of what the code does, on the one function whose whole subject is this conversion.

**4. The new Coqui test does not exercise the path it is named for**

`server/tts-sidecar/tests/test_coqui_device.py:158-176` — `test_admitted_rocm_device_normalises_to_cuda`, whose docstring says "an admitted device can arrive as an explicit 'rocm:N'" — sets `COQUI_DEVICE=rocm:0` and calls `_resolve_runtime_options(_torch_stub(...))` with **no `device_override`**. That is the env path, not the admitted path. `grep -rn device_override server/tts-sidecar/tests/` returns zero hits: the `device_override=` parameter has no test at all, before or after this PR, even though it is the parameter the admission layer uses and the reason the fix exists.

Concrete regression it would not catch: someone later writes `device = device_override if device_override is not None else _resolve_torch_device(self._device, torch_module)` — a plausible "an admitted device is already concrete" optimisation. The shipped bug returns in full and both new tests stay green. Adding `device_override="rocm:0"` as a second case (I confirmed it returns `{'device': 'cuda:0', 'half': True, 'deepspeed': True}` on this branch) closes it in one line.

**5. The register's changelog was not advanced for this row**

`docs/testing/onbox-acceptance-register.md:435` still reads "**Last change: 2026-08-30, no count change.**" and describes row A24 / #2678, while this PR changed the Blocked group 5 → 6 on 2026-09-01. Every prior row addition in that file has its own entry — including Blocked-section changes with no owed-count change (see the `2026-08-27, 60 → 60 (no count change)` entry at `:567`, which is this exact shape). `npm run check:onbox-register` passes (I ran it, exit 0) but does not police this. The row placement, the absent row ID (Blocked rows correctly carry none since the 2026-08-27 change), and the 66-owed arithmetic (40+2+4+3+13+2+2, Blocked/Unconfirmed excluded) are all correct.

**6. The live view moved in the diff but not at the URL it is read from**

`docs/testing/onbox-acceptance-register-live-view.html:168` bumps the publish token to `data-published-as="4"`, but the PR body states the publish itself is deferred — so the page at the register's recorded artifact URL still shows "Blocked — 5 rows" and no ROCm row. That is the one surface a person planning an acceptance session actually reads, which is the whole point of the row. The PR body marks this item `[x]` while its own text says the step is owed; at minimum the box should not be checked. (I am flagging the state, not second-guessing the contention call — but note the token is already stamped, so the publish is the only remaining action.)

---

### Checked and clean

- `_resolve_torch_device`'s existing paths, exhaustively: `auto`/`AUTO`/`''` → `cuda:0`, `cuda` → `cuda`, `cuda:x` → `cuda:x`, `cpu`/`mps` unchanged. The Coqui call-site change is behaviour-preserving for `cuda:1`/`cpu`/`mps` (verified by direct probe) and additionally normalises `" AUTO "`-style values that the old exact `== "auto"` test let through to a crashing `.to("AUTO")` — an improvement, not a regression.
- No regression in `_engine_actual_card`'s `fell_back` (`:10344`): pre-fix, a ROCm load left `requested_fam == "rocm"`, so `fell_back` could never fire; post-fix it reads `"cuda"` and the detector works. Strictly better.
- No reporting regression: `_normalize_device_family` (`:10131`) still re-labels `cuda` → `rocm` on a HIP build, so `/health` device predictions stay honest.
- `shares_device` (`:5394`) now correctly reports two rocm:0-pinned engines as sharing a card (it returned `False` before) — a fix, and it fails safe either way.
- `_resolve_codec_device('auto', ...)` now inherits the converted `"cuda:N"` from `_device`, so the Qwen codec move is fixed as a side effect.
- Kokoro's deferral is defensible: `_kokoro_provider_options("rocm:1", …)` returns `None` (index pin silently dropped), but `accelerator-profile.mjs` gives the amd profile a CPU-only ORT provider list, so there is no GPU placement for the pin to be dropped from. Genuinely distinct, and unreachable today.
- Release-notes pair present in both files; `Closes #2813` present and outside any code span; no `cast.json` writes, no new config knob, no derived artifact staled.
- **Flake claim verified, not taken on faith.** The diff touches seven files and **zero** under `server/src/**` — no TypeScript at all. `analysis.fresh-cast-lock.test.ts`, `generation-stall-watchdog.test.ts` and `voice-style.test.ts` cannot be affected by this change by construction.

---

**Re-review trigger:** findings 1 and 2 are 🟠 correctness bugs, so fixing and pushing them re-triggers a pass. Findings 3–6 are 🟡 and are fixed in this round but do not on their own reopen the loop.

dudarenok-maker:
## PR review — pass 2 (head c3e3c8f5d80988b472653f35793ed1188d2fadf6, depth medium)

Re-review of the six pass-1 findings, then a fresh sweep over the fix commit (`c3e3c8f5`). I re-derived each prior finding rather than reading the fix commit's claims: the twelve new `rocm` tests were run against a temp tree carrying `origin/main`'s `main.py` (all twelve red), the full sidecar suite was run against this branch (**1128 collected, 1126 passed, 2 skipped, 0 failures, 0 errors**), and both simulated-ROCm repros from pass 1 were re-run.

---

**1. `/transcribe` and `/embed` take an unconverted admitted `"rocm:N"` — 🟠 VERIFIED RESOLVED.**
`_normalise_rocm_device_key` (`server/tts-sidecar/main.py:5343-5359`) is applied at all four sites. Re-derived against this branch:

```
_ct2_kwargs('rocm:0','int8_float16') -> {'device':'cuda','compute_type':'int8_float16','device_index':0}
_ct2_kwargs('rocm:1','int8_float16') -> {'device':'cuda','compute_type':'int8_float16','device_index':1}
_spk_run_device('rocm:0')            -> 'cuda:0'      # was the opaque 'rocm'
_spk_run_device('rocm:1')            -> 'cuda:1'
```

The chokepoint placement is right in both engines, not just the leaf helpers: `WhisperEngine._ensure_loaded` normalises at `:8179` *before* `_compute_type(dev)` (`:8201`, `:8216`) and before `self._device = dev` (`:8219`), so the compute type no longer silently degrades to the CPU `int8` and `/health`'s `fell_back` reads a cuda-family value. `SpeakerEngine.ensure_loaded` normalises at `:8508-8509` *before* the cuda-presence guard at `:8513` and the cpu-demote fallback at `:8528` — both `family == "cuda"`-gated, both previously skipped. Index preservation holds at every site (`rocm:1 -> cuda:1`, `rocm:7 -> cuda:7`), and `cuda:N`/`cpu`/`auto`/`mps`/`None` are all returned untouched.

All four new tests are red on `origin/main` for the reason claimed — `test_spk_..._and_the_guard_fires` in particular asserts `eng.device == "cpu"`, which is only reachable *because* the guard fired, so it is not a "the string looks right" tautology.

**2. A `cuda:N` pin can never match the ledger's `rocm:N` keys — 🟠 VERIFIED RESOLVED.**
Re-ran pass 1's own repro against this branch (`_cuda_is_rocm` stubbed True, `COQUI_DEVICE=cuda:0`, probe = two 24 GB `rocm` cards, 22 GB free each):

```
_engine_env_pin('coqui')                 -> 'rocm:0'   (was 'cuda:0')
admit('coqui', pinned='rocm:0')          -> {'device': 'rocm:0'}   (was noCapacity)
```

`_is_resident` re-tags on both its branches (`:5111` unknown-fallback, `:5114` main), and `_qwen_resident_device_key` now routes through the same helper at `:5195` with identical output. I traced every consumer of both functions: `_engine_env_pin` is only ever passed as `pinned=` (ten call sites, `:10866`–`:12050`) and `_is_resident` only as `PlacementController(is_resident=…)` (`:5317`), so neither re-tagged value can reach a real device call. `noCapacity.deviceKey` now agrees with Node's own key format (`capacity-retry.ts:275` builds `` `${d.kind}:${d.index}` `` from the same probe) — an improvement, since the pinned case used to emit a `cuda:N` key Node could never match either.

On the "should this have been one bidirectional helper" question: no, and they are not drifting. Both derive family through the same `_parse_device`, and the reverse direction routes through the single `_cuda_is_rocm()` call the repo already standardised on (#2678 F2). They are not symmetric operations — the forward conversion is unconditional, the reverse is hardware-gated — so folding them into one function would need a direction argument and read worse. Not a finding. (Their *stated* contracts do have a hole; see finding 10.)

**3. `_qwen_resident_device_key`'s docstring was false — 🟡 VERIFIED RESOLVED.** `main.py:5162-5180` now states the opposite of the old text and explains the re-tag. Accurate against the code.

**4. The new Coqui test did not drive `device_override` — 🟡 VERIFIED RESOLVED.** `tests/test_coqui_device.py:178-202` deletes `COQUI_DEVICE` and passes `device_override="rocm:0"` directly, so it exercises the admitted-device parameter and nothing else. Confirmed red on `origin/main`. The "an admitted device is already concrete" mutation pass 1 named would now fail this test.

**5. The register changelog was not advanced — 🟡 VERIFIED RESOLVED.** `docs/testing/onbox-acceptance-register.md:434-455` carries a new `**Last change: 2026-09-01, 66 owed (no change) — Blocked 5 → 6.**` entry with the 2026-08-30 entry demoted to `**Prior change:**`. `npm run check:onbox-register` exits 0 ("register and live view agree").

**6. The live view moved in the diff but not at its URL — 🟡 PARTIALLY RESOLVED, STILL OPEN.** The narrow half is fixed: the PR body's checkbox is now `[ ]` with an honest note, which is what pass 1 asked for at minimum. The substantive half is unchanged — the page at the register's recorded artifact URL still shows `Blocked — 5 rows` and no ROCm row, and the token is already stamped (`data-published-as="5"`), so publishing is the only remaining action. CLAUDE.md's before-shipping step 3 names the live view as one of *three* surfaces that must move in the shipping PR, and "recording blocks the merge" — so this is still a merge-gate item, not a nice-to-have. Not re-raised as a new finding; carried forward as open.

---

### New findings

#### 🟠 Significant — correctness

**7. On a ROCm box the entire idle-eviction ladder yields zero steps, so admission 503s instead of evicting — `main.py:5198-5212`, `:5267/:5269/:5272/:5295`**

Third seam of the same two-vocabulary split this PR closes at the other two. `_same_card` requires family `"cuda"` on **both** sides:

```python
fam_r, idx_r = _parse_device(resident)
fam_t, idx_t = _parse_device(target)
if fam_r != "cuda" or fam_t != "cuda":
    return False
```

`target` is always a probe-derived key from `_worst_device_key` (`:4491`) → `_device_key` (`:4471`) → `probe_capacity`'s `kind`, which is `"rocm"` on a HIP box (`:4050`). So every one of `_idle_evict_steps`' four `_same_card` calls returns False and the ladder is empty; `_evict_until` (`:4547`) iterates nothing and falls straight through to `noCapacity`.

Identical scenario, only the probe's `kind` differing — one 24 GB card with 1000 MB free, an idle ASR and SPK resident on it, admitting Coqui (peak 3584 MB), where the evict step actually frees the card:

```
probe kind='cuda'  -> admit={'device': 'cuda:0'}                                     evict steps run=1
probe kind='rocm'  -> admit={'noCapacity': {'neededMb': 3584, 'deviceKey': 'rocm:0'}} evict steps run=0

_idle_evict_steps('cuda:0','spk') -> ['spk', 'asr', 'qwen.design', 'qwen.base17', 'coqui']
_idle_evict_steps('rocm:0','spk') -> []
```

Concrete scenario: ROCm box, `SEG_CAPACITY_ADMISSION` on (default), an idle Coqui/Qwen/ASR/SPK holding VRAM. Every admission that could be rescued by freeing an idle engine returns 503 `noCapacity` instead — the whole cheapest-reload-first ladder (`#1920A`, `#1721`, `#1894`) is dead on AMD. The user-visible symptom is a render refusing with "no capacity" on a card whose VRAM is held entirely by models nothing is using.

This predates the PR, and I am reporting it anyway, for four reasons:
- it is squarely in the class this PR set out to close — the ledger speaks `rocm:N`, the engines speak `cuda:N`, and this is the third place the two meet;
- **this PR makes the mismatch structurally permanent.** Before it, `SPK.device`/`ASR._device`/`qwen._device` could at least hold the raw admitted `"rocm:0"`; the forward fix guarantees they are now always `"cuda:N"`, so `_same_card` can never again match a ROCm ledger key by accident;
- `_same_card`'s own docstring (`:5200-5201`) — "`target` (always a concrete "cuda:N" from `_worst_device_key`)" — is now provably false on the exact hardware this PR is about, and it is the sentence a reader would rely on to conclude the function is fine;
- CLAUDE.md lists "it's pre-existing" / "it's next door" among the **void** deferral reasons, and there is no design decision owed here: `_same_card` should compare cards, not vocabularies, and the repo already has `_normalise_rocm_device_key` to do it with.

The register row and the release-notes entry also currently claim the ROCm GPU path is fixed; with this seam open, an admitted ROCm load still fails whenever it needed an eviction to fit.

#### 🟡 Minor — cleanup

**8. The corrected comment landed in one of its two copies, and its one new citation is wrong — `server/src/config/resolver.test.ts:114-116`, `server/src/config/registry.ts:409`**

Two halves, same chore.

*The second copy.* `resolver.test.ts:99-134` carries the same paragraph this PR just falsified and corrected in `registry.ts`, nearly verbatim and still uncorrected (`:114-116`):

> Nothing upstream of `_parse_device` ever PRODUCES family "rocm" from a real input, so the `in ("cuda","rocm")` checks scattered through main.py are unreachable dead code for that arm.

Plus the same drifted citations pass 1 flagged (`main.py:3367`, `:9100-9103`, `:9124-9134`, `:7540`, `:7148`, `:7488`) — and `:9100-9103`/`:9124-9134` are additionally baked into a test *name* at `:159`, so they surface in test output. The PR body declares "`registry.ts`'s stale comment … is corrected"; the identical claim one file over is the second copy the class-fix missed.

*The new citation.* `registry.ts:408-409` now reads:

> (`probe_capacity` at :4032's `kind = "rocm" if _cuda_is_rocm() else "cuda"`)

`probe_capacity` is at `main.py:4041` and its `kind =` line is `:4050`. `:4032` is inside `_host_mem_mb`'s `os.sysconf` fallback — a different function. The other two new citations (`_torch_is_hip` `:10174`, `_normalize_device_family` `:10198`) I checked and they land correctly; this is the one the fix commit introduced while repairing exactly this defect shape.

**9. `_is_resident`'s new docstring overclaims — Kokoro is a live counter-example — `main.py:5092-5094`**

The new text asserts:

> now that an engine's own `self._device`/`.device` is **always** normalised to torch-native "cuda:N" (never a literal "rocm:N" string any more)

Not true for Kokoro, which is one of the engines `_is_resident` covers and the engine whose API-drift case the `"unknown"` fallback branch exists for. `KokoroEngine._ensure_loaded` sets `intent_device = resolved_device` (`:3460`) and publishes `self._device = intent_device` (`:3596`) — `resolved_device` is the raw admitted key, and the `"auto"` re-derivation at `:3461-3462` does not fire for family `"rocm"`. So an admitted ROCm load leaves `kokoro._device == "rocm:0"` literally.

No live bug today — `_report_rocm_device_key` is idempotent on an already-`rocm:N` key, so the output is still correct — but the docstring is the stated justification for the re-tag at that site, and a future reader who trusts "never a literal rocm:N any more" could remove the idempotence the correctness now silently depends on.

**10. Both new helpers' "returned unchanged" contract is false for an unindexed family — `main.py:5354-5355`, `:5375-5376`**

Both docstrings promise that anything outside their own family is passed through untouched:

> `_normalise_rocm_device_key`: "Any other value (cuda:N, cpu, auto, mps, an already-'cuda:N' key, None) is returned unchanged."
> `_report_rocm_device_key`: "Any other value (cpu, auto, mps, an already-'rocm:N' key, None) is returned unchanged."

Both invent index 0 instead:

```
_normalise_rocm_device_key('rocm')             -> 'cuda:0'
_report_rocm_device_key('cuda')                -> 'rocm:0'      # on a rocm box
_report_rocm_device_key('cuda-uuid:GPU-abc')   -> 'rocm:0'      # on a rocm box
```

Unreachable at the four current call sites — each formats an indexed key first, `_engine_env_pin` gates on `idx is not None`, and `buildSidecarEnv` (`server/src/tts/spawn-sidecar.ts:549`) deletes any ambient `*_DEVICE` that fails validation, so a bare `"rocm"` never reaches the sidecar. But these are documented as general-purpose shared helpers extracted precisely so new call sites can reuse them, and the fifth caller that passes an unindexed `"cuda"` or an unresolved uuid gets card 0 silently substituted, against the contract it read.

---

### Checked and clean

- Full sidecar suite green on this branch (1126 passed, 2 skipped, 0 failures). All 16 CI checks green at this head.
- All twelve new `rocm` tests red against `origin/main`, each for the reason its docstring claims. The two `stays_cuda_on_a_real_cuda_box` companions are explicitly labelled controls, not regression tests, and pass on both sides as intended.
- The re-tag is a genuine no-op on NVIDIA: `_report_rocm_device_key` is identity when `_cuda_is_rocm()` is False, and `_normalise_rocm_device_key` is identity for `cuda:N`/`cpu`/`auto`/`mps`/`None`.
- Re-enumerated all ten consumers of `adm["device"]` (`:10870`–`:12056`). The two pass 1 flagged are now covered; Coqui's `clone_voice`/`synthesize` reach `_resolve_runtime_options` via `_ensure_loaded(model, device=device)`, Qwen's paths via `_ensure_device_resolved`. Kokoro's `:10870` remains deliberately unnormalised and is still harmless — `_kokoro_provider_options` gates on `family != "cuda"` and returns `None` (`:5334`), and the amd profile's ORT provider list has no CUDA provider for an index to attach to.
- `_observed_mb` / `_reset_peak_mb` (`:4708`, `:4725`) already accept a `"rocm:"` prefix and split the index correctly, so the re-tagged residency keys do not break peak accounting.
- `SpeakerEngine` normalises only the admitted `device=` param, not the env-derived `self.device` (asymmetric with `WhisperEngine._ensure_loaded`, which normalises both). Not raised: `buildSidecarEnv` deletes a validation-failing `SPK_DEVICE`, so `self.device` cannot hold a rocm family in practice — no reachable failure.
- Register arithmetic, group counts and the `.md`/`.html` pair all agree (`check:onbox-register` exit 0); release-notes pair present and now accurate about the pinned case; `Closes #2813` present and outside any code span. No `cast.json` writes, no new config knob, no staled derived artifact.

---

**Re-review trigger:** finding 7 is a 🟠 correctness bug, so fixing and pushing it re-triggers a pass. Findings 8–10 are 🟡 and are fixed in this round but do not on their own reopen the loop. Prior finding 6 remains open as a merge-gate bookkeeping item (publish the live view before merge).

dudarenok-maker:
## PR review — pass 3 (head f1a713b89f40913d1379a90f804001a6444b3207, depth medium)

Final round (loop cap: initial + 2 re-reviews). Each pass-2 finding re-derived independently against the branch rather than read off the fix commit's summary; then a full sweep. Evidence below is from runs, not from the diff.

---

**7. `_same_card` emptied the idle-eviction ladder on a ROCm box — 🟠 VERIFIED RESOLVED.**

Not verified at the helper level only. I wrote my own end-to-end repro driving the **real `PlacementController.admit()`** with the **real `_idle_evict_steps`** and a simulated probe (one 24 GB card, 1000 MB free, idle Qwen+SPK on card 0, ASR on card 1, admitting Coqui at peak 3584 MB), and ran the identical script against `c3e3c8f5` (pass 2's head) and `f1a713b8`:

```
                     pre-fix (c3e3c8f5)                              post-fix (f1a713b8)
probe kind='cuda' -> ladder ['spk','qwen.design','qwen.base17']   -> same
                     admit {'device': 'cuda:0'}                       admit {'device': 'cuda:0'}
probe kind='rocm' -> ladder []                                    -> ladder ['spk','qwen.design','qwen.base17']
                     admit {'noCapacity': {..., 'rocm:0'}}            admit {'device': 'rocm:0'}
                     freeMb 1000 -> 1000                              freeMb 1000 -> 7200
```

The NVIDIA control row is byte-identical on both sides — no regression to the path that ships today. `asr` (card 1) is correctly still excluded post-fix, so the fix matches cards, it does not match everything.

Both new tests are red on `c3e3c8f5` for exactly the claimed reason (`_same_card("cuda:0","rocm:0")` returns `False`; `_idle_evict_steps` returns `[]`), green after.

**I also mutation-tested the fix**, because "normalise both sides" is the kind of claim a half-fix passes:

| mutant | `test_same_card_…` | `test_idle_evict_…` |
|---|---|---|
| normalise `target` only (resident raw) | **FAIL** | pass |
| normalise `resident` only (target raw) | **FAIL** | **FAIL** |
| unmutated | pass | pass |

Each half-fix is killed, so the test genuinely pins both sides rather than asserting a string shape. The `("rocm:0","cuda:0")` assertion is what catches mutant one — it is load-bearing, not decoration.

**8. Stale/duplicated `registry.ts` comment + drifted citations — 🟡 RESOLVED AS SCOPED, but the fix re-armed the class. See new finding 11.** The two items pass 2 named *are* addressed: `registry.ts:409`'s `probe_capacity` citation is now `:4041`/`:4050` and both are **correct** (`def probe_capacity` is at `main.py:4041`, `kind = "rocm" if _cuda_is_rocm() else "cuda"` is at `:4050` — verified), and `resolver.test.ts`'s duplicate block was rewritten. Five *other* citations in those same two files are wrong.

**9. `_is_resident`'s docstring overclaimed — 🟡 VERIFIED RESOLVED (one sub-nit, finding 12).** `main.py:5098-5109` now names Kokoro as the exception, and the mechanism it cites is accurate against the code: `KokoroEngine._ensure_loaded` sets `intent_device = resolved_device` (`:3460`) and only re-derives when `family == "auto"` (`:3461-3462`), publishing at `:3596`. Correct.

**10. Both helpers' "returned unchanged" contract — 🟡 VERIFIED RESOLVED.** Probed directly with `_cuda_is_rocm()` forced `True`:

```
_normalise_rocm_device_key('rocm')            -> 'cuda:0'   # now documented
_report_rocm_device_key('cuda')               -> 'rocm:0'   # now documented
_report_rocm_device_key('cuda-uuid:GPU-abc')  -> 'rocm:0'   # now documented
_report_rocm_device_key('rocm')               -> 'rocm'     # unchanged, as documented
```

Both docstrings now match observed behaviour exactly, including the `cuda-uuid:` case.

**6. Live-view publish still owed — 🟡 carried forward; the deferral reasoning checks out, see finding 13 for the residual.** I verified the stated premise rather than accepting it: row **A28 is still present and owed on this branch** (`onbox-acceptance-register.md:2596`), so a live page showing owed=65 with A28 discharged genuinely carries content this branch does not have. Publishing from here would revert another lane's live edits with every mechanical check green — the exact hazard the register's own "Live view" section documents, and one `--against-published` cannot report, because #2199 classifies a live-only *absence* as a discharge and stays quiet. Deferring rather than overwriting is the correct call, and the PR body now states it honestly with the box unchecked.

---

### New findings

#### 🟡 Minor — cleanup

**11. The citation-fix commit shipped five wrong `main.py` line numbers — three of which were never correct at any revision, and two of which it staled itself in the same commit**

`server/src/config/registry.ts:395`, `server/src/config/resolver.test.ts:109-110`, `:130`, `:178`.

Verified by resolving each cited symbol at every revision on the branch:

| symbol | actual (f1a713b8) | `registry.ts` says | `resolver.test.ts` says |
|---|---|---|---|
| `_torch_is_hip` | **10217** | `:10174` ✗ | `main.py:10187` ✗ |
| `_normalize_device_family` | **10241** | `:10198` ✗ | `main.py:10211` ✗ |
| SPK demote gate (`family=="cuda"` → `else: raise`) | **8571-8576** | — | `main.py:8542` ✗ |

Two distinct failure modes, both worth naming:

*Self-staled.* `registry.ts:395`'s `:10174`/`:10198` **were correct at `c3e3c8f5`** — pass 2 checked them and said so. `f1a713b8` then added 43 lines of docstring to `main.py` *above* them (`+12/+13/+9/+9` across its four hunks) and did not re-derive. The commit that edited `registry.ts` to fix a citation is the commit that broke the two citations next to it:

```
rev          _torch_is_hip   _normalize_device_family
origin/main      10080            10104
c3e3c8f5         10174            10198   <- what registry.ts still says
f1a713b8         10217            10241   <- truth
```

*Never correct.* `resolver.test.ts`'s `:10187`/`:10211` match **no revision** — they are 30 short of `f1a713b8` and 13 past `c3e3c8f5`, i.e. a mid-edit snapshot taken after one of the four hunks. `:8542` likewise: the demote gate is at `:8571` here, `:8529` at `c3e3c8f5`, `:8435` on `origin/main`. `main.py:8542` on this branch is a comment line inside a *different* concern (the load-lock note). It has never pointed at what it claims.

And `:10187`/`:10211` are **baked into a test name** at `resolver.test.ts:178`, so they render in every `test:server` run — the identical shape pass 2 flagged at the old `:159` and this commit was rewriting:

```ts
it.each(KNOBS)('%s rejects bare "rocm" — it is a DERIVED reporting label (main.py:10187, :10211), never a valid OPERATOR input; …
```

No runtime effect — comments and a test-name string. But this is round three of the same defect (pass 1 flagged six drifted citations, pass 2 flagged one wrong new one, pass 3 finds five), the PR body currently asserts "corrected with accurate line references", and a citation into a 12k-line file is staled by any edit above it — including the edit that writes it. `registry.ts:409`'s two-part form (`probe_capacity at :4041`… `:4050`) is the one that survived, because the symbol name carries the meaning and the number is only a hint. Worth fixing in that shape rather than re-deriving five more absolute numbers that the next docstring edit invalidates. (`resolver.test.ts:101`'s `main.py:3905` is fine — it cites `_engine_env_pin`'s `def` line, the same convention `:4041` uses; the check itself is at `:3926`.)

**12. `_is_resident`'s new idempotence rationale describes a step the code never executes — `main.py:5103-5105`**

The corrected docstring says Kokoro's literal `"rocm:N"` is harmless because

> `_report_rocm_device_key` is idempotent on an already-"rocm:N" input (**it reads that family, sees the box genuinely is ROCm**, and returns it unchanged)

The middle clause is false. The guard is `if fam == "cuda" and _cuda_is_rocm():` — for a `"rocm:N"` input `fam == "cuda"` is `False`, so Python short-circuits and **`_cuda_is_rocm()` is never called**. The idempotence is unconditional, not ROCm-box-conditional:

```
_report_rocm_device_key('rocm:0')  -> 'rocm:0'   # on a ROCm box AND on NVIDIA
```

The conclusion is right; the stated mechanism is not. It matters because this docstring is the recorded justification for the re-tag, and a reader who believes the box check participates could add a "safety" branch for the NVIDIA case that the code already handles — the same class as findings 9 and 10, in prose the same round authored.

**13. The deferred live-view publish loses its only tracker on merge — `docs/testing/onbox-acceptance-register.md:24-49`, PR body test-plan item**

Finding 6's deferral is sound (above), but the residual is not currently recorded anywhere durable. After merge: `origin/main`'s register says **66 owed / Blocked 6** with a ROCm row; the live page at the canonical artifact URL says **65 / Blocked 5** with no ROCm row. The live page is the surface CLAUDE.md step 3 says a person planning an acceptance session actually reads, and it will be silently one row behind.

The only two places that record the owed publish are (a) the PR body's unchecked box, which stops being read the moment the PR merges, and (b) issue #2813 — which this PR **closes**. So on merge there is no home for it at all. The register's own "Live view" section has no "publish owed" slot, and the publish token is documented as **inert** (`:78` — "nothing reads it yet", the comparator lands with #2599), so nothing mechanical will catch the drift either.

Concretely: file a `type:chore` issue for "publish the live view once the wave-10 lane merges" (or add an explicit owed-publish line under the register's Live view section) before merging, rather than relying on the PR body. This is bookkeeping the deferral made owed, not a reason to hold the fix.

---

### Checked and clean

- **Full sidecar suite on this branch: 1130 collected, 1128 passed, 2 skipped, 0 failed, 0 errors** (exit 0) — +2 over pass 2's 1126, exactly the two new tests. All 16 CI checks green at this head; `mergeStateStatus: CLEAN`.
- **Double-normalisation risk in `_same_card`: none.** `_normalise_rocm_device_key` is identity for `cuda:N`/`cpu`/`auto`/`mps`/`None` and idempotent on its own output, so an already-`cuda:N` resident is treated exactly as before (`('cuda:0','cuda:0')` True, `('cuda:0','cuda:1')` False, `('cpu',…)`/`(None,…)` False). The only behaviour deltas are the three intended ones: a `rocm` family on either side now matches, and bare `'rocm'` resolves to card 0 — the same substitution unindexed `'cuda'` already got. Round-trip `normalise → report → normalise` is stable for every indexed input.
- **Consistency with the other helper call sites.** Traced all 8 uses of the two helpers. `shares_device` (`:5501-5502`) reaches the same normalisation via `_resolve_torch_device`, so it needs no second one. `_validate_cuda_index` (`:5453`) does *not* normalise, but all three callers (`:1926`, `:5949`, `:5953`) pass a value already through `_resolve_torch_device`. `_evict_until`'s `ledger.engines_holding(device_key)` still receives the raw probe key while `_same_card` normalises internally — correct, since reservations are booked under the probe-native key `admit()` returns, so the #1920B guard still fires on ROCm.
- **Class swept for a fourth seam.** Enumerated every `== "cuda"` / `in ("cuda","rocm")` gate in `main.py` (24 sites). `:1799`/`:1967` (Coqui half/deepspeed) read the post-`_resolve_torch_device` value; `:8206`/`:8556`/`:8571` read the normalised `dev`/`self.device`; `:8689-8701` and `:10409-10413` read `_requested_device`, which is env-derived and config-validated to `cuda:N` only; `:12009`/`:12088` accept both families. `_kokoro_provider_options` (`:5359`) is the documented Kokoro deferral. I did not find a fifth seam.
- **No import-order hazard** from `_same_card` (`:5210`) calling `_normalise_rocm_device_key` (`:5368`): resolution is at call time and nothing invokes either at module scope — confirmed empirically, `import main` succeeds.
- Register `.md`/`.html` are internally consistent and correctly describe all three seams including the pass-2 one; `npm run check:onbox-register` exit 0. Blocked 5 → 6 with 66 owed unchanged is right per the file's own documented rule (Blocked/Unconfirmed excluded). Re-stamping the publish token 3 → 6 without publishing is harmless — the token is documented inert, and "stamp, then commit, then publish" is the prescribed order.
- Release-notes pair present in both files and now accurate about all three seams (the earlier over-claim pass 1 raised is gone). `Closes #2813` present, outside any code span; issue open and `bug`-labelled; `pr-issue-link` green.
- No `cast.json` writes, no new config knob, no staled derived artifact, no runtime TypeScript change in the diff at all (`registry.ts`/`resolver.test.ts` move comments and one test-name string only).

---

### Verdict

**No 🔴 Blocking or 🟠 Significant findings. Zero new correctness bugs.** The runtime fix is verified correct end-to-end through the real admission path, mutation-tested against both half-fixes, with the NVIDIA path provably unchanged.

**Ready to merge on the code.** Findings 11–13 are 🟡 cleanup — two comment/docstring corrections and one bookkeeping row — and per findings-triage they are fixed in this round but do not reopen the loop, which is now at its cap anyway. Of the three, **13 is the one worth doing before the merge button** rather than after: it is the only one whose cost rises once the PR closes, because `Closes #2813` removes the last place the owed live-view publish is recorded.

PR #2839, opened 2026-08-31
fix(ops): detect sibling-worktree contention and scope pre-commit tests to the diff

### LINKED ISSUE #2834: chore(ops): pre-commit contention guard misses sibling worktrees; full-suite reruns on every commit
## What

Two related dev-workflow reliability gaps in `scripts/verify-cache.mjs`'s pre-commit path, both observed live 2026-09-01:

1. **Sibling-worktree contention is invisible to the contention guard.** The guard only probes `nvidia-smi` for a busy GPU. Six worktrees were observed running full `test:server` batteries concurrently — the documented cause of `Worker exited unexpectedly` fork-pool crashes (`docs/features/archive/45-vitest-pool-tuning.md`) — with the GPU completely idle. The crash-and-retry cycle this causes is what has been turning routine commits into 3+ hour ordeals instead of 5-10 minutes, and has previously led to `--no-verify` being reached for as a workaround (never the right call).
2. **Pre-commit reruns the whole `test:server`/`test` suite (~6700 tests) for any touched file**, even a single-file change, because the existing scope filter is all-or-nothing per step. Combined with (1), this multiplies the pileup risk under concurrent-worktree load.

## Fix (shipped this PR)

- `detectSiblingContention()`: enumerates other `vitest`/`verify-cache.mjs` node processes via `Get-CimInstance` (excluding self) and throttles to `LOW_CONCURRENCY=1`, mirroring the existing GPU-busy throttle. Soft (warn + dial down, never block), fails open on non-Windows/PowerShell errors, gated by the existing `SKIP_CONTENTION_CHECK`.
- `--scope-staged` (pre-commit only) now runs vitest's own `--changed HEAD` selection (`test:changed`/`test:server:changed`) for the `test`/`test:server` steps instead of the full suite. `--scope-branch` (pre-push) and CI are unchanged and still run every test, so anything `--changed` misses is still caught before merge.

## Acceptance

- New unit tests in `scripts/tests/verify-cache.test.mjs` for `siblingContentionFor`/`detectSiblingContention` and the `CHANGED_ONLY_NPM_SCRIPT`/`runStepProcess` retry-key wiring.
- `npm run test:hooks` green (120/120).

## Key files

- `scripts/verify-cache.mjs`
- `scripts/tests/verify-cache.test.mjs`
- `package.json`, `server/package.json` (new `test:changed`/`test:server:changed`/`pretest:changed` scripts)
- `CLAUDE.md` (Commands + Commit gate sections)

dudarenok-maker:
## Summary

Two related pre-commit reliability/speed fixes, both observed live 2026-09-01 (six worktrees running full `test:server` batteries concurrently):

- **Sibling-worktree contention detection.** `verify-cache.mjs`'s contention guard previously only probed `nvidia-smi` for a busy GPU. A new `detectSiblingContention()` also enumerates co-running `vitest`/`verify-cache.mjs` node processes (excluding itself) via `Get-CimInstance` and throttles to `LOW_CONCURRENCY=1` the same way a busy GPU does — soft (warn + dial down, never block), gated by the existing `SKIP_CONTENTION_CHECK`, fails open on non-Windows/PowerShell errors.
- **`--scope-staged` (pre-commit) now runs vitest's own `--changed HEAD` selection** (`test:changed`/`test:server:changed`) for the `test`/`test:server` steps instead of rerunning the whole suite (~6700 tests) for any touched file — but **only** when the diff is confined to the step's own primary source tree (`src/**` / `server/src/**`) with a TS/TSX/JS source extension (`diffSafeForChangedOnly`). `--scope-branch` (pre-push) and this repo's own CI are unaffected: pre-push always runs the full, unsubstituted suite (CI separately narrows its own PR runs via `vitest run --changed <base>`, a pre-existing, unrelated design), so anything `--changed` misses is still caught before merge.

Also updated `CLAUDE.md` (Commands + Commit gate sections), both release-notes documents, and `docs/features/156-precommit-scope-contention.md` (the regression plan this extends).

**Review history**, for anyone reading this after merge — getting "is this diff safe to narrow" right took the full 3-round review cap, each round finding a live, non-hypothetical gap in the previous answer, all against real commits reachable from this branch:
1. A shared-scope diff (root manifest/lockfile) still got substituted, silently ran zero tests via vitest's `passWithNoTests`, and was being cached under the step's FULL input hash, letting pre-push/CI read `[cached]` and skip a full run that never happened.
2. Excluding shared-scope wasn't enough — a step's own `extraFiles` entries and the server lockfile are the identical shape one level down.
3. Even an exclusion list of known-bad categories left a third spelling armed — `test:server`'s own `globs` match real root-level config files (`eslint.config.mjs`) vitest's `--changed` can't reach, and a non-JS asset like `src/styles.css` sitting under a step's source glob is unsafe too (read via `readFileSync` by its guard tests, not imported).

The final design stopped chasing exclusions: `diffSafeForChangedOnly` is a positive allowlist, not a blocklist — deliberately narrower than "matches this step's declared scope" (a config-only or CSS-only commit now runs the full suite rather than being narrowed), trading away some narrowing for the one guarantee that actually matters: never silently running zero tests. Full findings from all three rounds are in the PR comment thread.

Closes #2834

## Test plan

- [x] `scripts/tests/verify-cache.test.mjs`: unit tests for `siblingContentionFor`/`detectSiblingContention`, the `CHANGED_ONLY_NPM_SCRIPT`/`runStepProcess` `retryKey` wiring, the full substitution gate condition, and `diffSafeForChangedOnly` itself — including the two live-commit cases (`eslint.config.mjs`, `src/styles.css`) that motivated the round-3 redesign (126 tests total in this file).
- [x] `npm run test:hooks` — passed (run live under this PR's commits).
- [x] `npm run test` (`--changed HEAD`) — 345/345 passed.
- [x] `npm run test:server` (`--changed HEAD`, widened to full by the `package.json` trigger on the first commit) — 8201/8210 passed (9 pre-existing skips), surviving multiple real fork-pool crashes under this box's heavy load — confirming the retry mechanism correctly identifies the step as `test:server` (not the substituted script name) after the `runStepProcess`/`retryKey` refactor.
- [x] `verify:fast:branch` (pre-push) green on every pushed commit, including full (unsubstituted) `test:server` runs.
- [x] Local `eslint` clean on the final diff.


dudarenok-maker:
## PR review — pass 1 (head ddc021d8b777da985aceef2219a7b92aaf74cf59, depth low)

Reviewed the whole diff plus the surrounding machinery it depends on (`runPipeline`'s cache/scope loop, both `vitest.config.ts` `forceRerunTriggers` lists, `verify.yml`'s test legs, `.husky/pre-commit`/`pre-push`). The sibling-contention probe itself is sound — I ran its exact `spawnSync` command line on this box and it returns one PID per line with `status 0`, self-exclusion is correct, and the zero-match case returns `"\r\n"` → `count: 0` rather than erroring. The `runStepProcess` `npmScript`/`retryKey` split is correct.

The findings below are all on the `--changed` half.

---

### 🔴 Blocking — a `--changed`-narrowed pre-commit run writes a **full-input-hash** green cache entry, so pre-push's `verify:fast:branch` skips the full suite with `[cached]`

`scripts/verify-cache.mjs:1402-1424`

The substitution changes *what gets spawned* but nothing about the cache key. `currentHash` is still composed from the step's **entire** declared input set (`selectStepFiles` → `composeInputHash`, lines 1355-1385), and on exit 0 that hash is written under the unmodified `step.name`:

```js
const code = runStepProcess(changedOnlyScript ?? step.name, { cwd, env, retryKey: step.name });
...
if (code === 0) {
  console.log(`[pass] ${step.name} (took ${formatSecs(dt)})`);
  cache.steps[step.name] = { inputHash: currentHash, ... };
```

`decide()` (`scripts/verify-cache.mjs:648-653`) then returns `'skip'` for any later run whose hash matches, and `verify:fast:branch` (`package.json:81`) passes no `--no-cache`.

**Failure scenario, the ordinary flow:**

1. `git add server/src/routes/export.ts && git commit -m 'fix(server): …'`
2. Pre-commit (`.husky/pre-commit` → `verify:fast:scoped`) prints `[run] test:server (--changed HEAD only)`, runs the handful of tests whose module graph reaches that file, `[pass] test:server`, and writes `cache.steps['test:server'].inputHash = H` — where `H` is the hash of **all** of `server/src/**` + `server/package-lock.json` + `server/vitest.config.ts` + `server/tsconfig.json` + `server/package.json`.
3. `git push`. Nothing on disk changed between the commit and the push, and neither did `nodeVer`/`schemaVer` (these two steps carry no `toolFingerprint`), so pre-push's `verify:fast:branch` recomputes the identical `H`.
4. Output: `[cached] test:server (input hash unchanged)`. **The 8210-test suite never runs.**

Before this PR that same `[cached]` line was honest — the entry had been written by a genuinely full `npm run test:server`. This PR turns it into a false green without changing the line that prints it, which is this repo's #6 shape (success reported while doing nothing) landing inside the cache rather than inside a guard.

This also falsifies the PR's own load-bearing safety argument, in the PR body, in `CLAUDE.md:899-902`, and in `docs/release-notes-next.md:488` — see the separate finding on the CI half below. Net effect after this PR: **no gate runs the full suite on the branch tip before merge.** Pre-commit narrows, pre-push says `[cached]`, and CI narrows too (`verify.yml:276`, `:326-327`). Only `cross-os.yml`'s twice-weekly cron and `release.yml` still do.

No test covers this. The new tests are four pure-function cases on `siblingContentionFor` plus five source-regex pins; none exercises `runPipeline`'s substitution-and-then-cache-write path, which is exactly where the defect lives.

---

### 🟠 Significant — the substitution is applied even on a `shared`-scope diff, where the repo's own model says every test must run; result is `[pass] test:server` after executing **zero** tests

`scripts/verify-cache.mjs:1402`

```js
const changedOnlyScript = flags.scopeStaged ? CHANGED_ONLY_NPM_SCRIPT[step.name] : undefined;
```

This reads only `flags.scopeStaged`. It ignores `scopeShared`, set 84 lines earlier (`:1316-1318`) from `computeShared` (`:834-839`), which exists precisely to declare "this diff is global — every selected step is in scope". Plan 156's Invariant 2 states the same rule.

**Failure scenario — a root `package-lock.json`-only commit (a dependabot bump, or any bare `npm install`):**

1. `computeShared(['package-lock.json'])` → true → `[scope] shared-scope change detected — all selected steps in scope`.
2. `test` and `test:server` therefore run — but as `vitest run --changed HEAD`, with `related = ['<abs>/package-lock.json']`.
3. `package-lock.json` matches **neither** config's `forceRerunTriggers`. Both list `'{**/package.json,**/.*/**/package.json}'` and no lockfile entry (`vitest.config.ts:85-131`, `server/vitest.config.ts:190-235`). It is not a forced trigger, and no test imports it, so `filterTestsBySource` returns `[]`.
4. Zero specifications → `passWithNoTests` defaults to `true` in vitest 4, so vitest prints `No test files found, exiting with code 0` and exits 0. I confirmed this live in this worktree: `npx vitest run --changed HEAD` → that exact line, `EXITCODE=0`.
5. `verify-cache` prints `[pass] test` and `[pass] test:server`. **A dependency bump is committed with zero tests executed, reported as two green test steps** — and per the finding above, that vacuous pass is then cached and satisfies pre-push as well.

The pre-PR behaviour for this exact diff was a full run of both suites. `package.json` happens to be covered (it *is* a forced trigger, which is why this PR's own commit widened back to a full run — the PR body reads that as evidence the mechanism is safe; it is evidence only that the `package.json` spelling is covered, not the lockfile one sitting next to it).

---

### 🟠 Significant — the claim "`--scope-branch` (pre-push) and CI still run the full suite" is false about CI, and it is now written into `CLAUDE.md`

`CLAUDE.md:899-902`, `docs/release-notes-next.md:488`, PR body

> `--scope-branch` (pre-push) and CI still run the full suite, so a change `--changed` misses through an untracked dependency is still caught before merge.

`verify.yml` has run `--changed` for both test legs since CI cost round 2:

- `.github/workflows/verify.yml:276` — `if [ -n "$BASE" ]; then npx vitest run --changed "$BASE"; else npx vitest run; fi`
- `.github/workflows/verify.yml:326-327` — the same for `test:server` and `test:server-slow`.

On a PR, `BASE` is always set, so the full-suite branch is dead there. The sentence therefore asserts a backstop that does not exist on either half it names (the pre-push half fails for the separate, mechanical reason above). This is the #9 shape — the defect is in the document, and the document is `CLAUDE.md`, which every future agent reads as the operative process spec.

Related, and worth naming in whatever replaces that sentence: the `verify.yml:260-261` comment claims "a root dep/config bump (the `shared` scope) forces a full run via vitest's default triggers". That is true for `package.json` and false for `package-lock.json`, for the same reason as the finding above.

---

### 🟡 Minor — `docs/features/156-precommit-scope-contention.md` (`status: active`) is the regression plan for both halves of this change and was not updated

Before-shipping checklist step 1: "Changed behaviour cited in an existing plan → update that plan in the same diff." Plan 156 *is* the plan for the pre-commit scope filter and the GPU-contention throttle — this PR widens the second and materially changes the first. Two statements in it are now false or misleading:

- `:50-52` — "**Pre-push unchanged.** Pre-push still runs the FULL `npm run verify` battery, preserving the 'local is the full coverage net, CI is the scoped one' invariant." That invariant is the exact one this PR inverts.
- `:58` — Invariant 2, "A staged root `package.json` / `package-lock.json` change → every selected step is in scope (global)" — still literally true about *scope*, but it no longer means what it was written to mean, now that an in-scope step can run zero tests.

---

### 🟡 Minor — `CLAUDE.md`'s Commands section was not updated, contrary to the PR body and issue #2834

Both the PR body ("Also updated `CLAUDE.md` (Commands + Commit gate sections)") and #2834's Key files ("`CLAUDE.md` (Commands + Commit gate sections)") claim two sections changed. The diff contains a single `CLAUDE.md` hunk, at `:893`, entirely inside **Commit gate**. The Commands section's `npm run test` / `npm run test:server` bullets are untouched and say nothing about the new `test:changed` / `test:server:changed` / `pretest:changed` scripts.

---

### Checked and clean

- Sibling probe: exact command line verified live — one PID per line, `status 0`, the backtick-n `-join` separator survives Node's Windows arg-quoting intact; `$_.CommandLine` being `$null` for an unqueryable process is a benign `-match` false; zero matches yields `"\r\n"` → `count: 0`; `process.pid` self-exclusion is right and is unit-tested. Fails open on a non-Windows runner via `r.error`. `Number.parseInt`/`Number.isFinite` filtering handles blank and garbage lines.
- `runStepProcess`: `RETRIABLE_POOL_STEPS.has(retryKey)` and `spawnSync('npm', ['run', npmScript])` are the right way round, and both are pinned by source-regex tests.
- The second contention block's `!lowConcurrency(env)` guard correctly skips the (more expensive) PowerShell probe when the GPU probe already throttled.
- `server/package.json`'s `pretest:changed` correctly mirrors `pretest`; npm runs `pre<name>` for any script name. Root has no `pretest`, so `test:changed` needs none.
- Release-notes pair present in both files; `Closes #2834` present and outside a code span; issue exists, `area:ops` + `type:chore`.
- On-box acceptance: not applicable — no GPU/sidecar/analyzer/book-provable behaviour ships here.

### Summary

3 correctness findings (1 🔴, 2 🟠), 2 cleanup findings (🟡).

dudarenok-maker:
## PR review — pass 2 (head 3bb761f11541b14bb9b2f1f0df94db27827ee709, depth low)

Re-review of `3bb761f1` against pass 1's five findings, plus a fresh pass over `runPipeline`'s per-step loop. Deltas only — pass 1's "checked and clean" list is not re-litigated.

|---|---|---|
| 1 | 🔴 `--changed`-narrowed run cached under the full input hash | **Resolved** |
| 2 | 🟠 substitution applied on shared-scope diffs | **Partially resolved** — see new 🟠 below |
| 3 | 🟠 "CI still runs the full suite" is false | **Resolved in-repo; PR body still unchanged** |
| 4 | 🟡 plan 156 stale | **Partially resolved** |
| 5 | 🟡 `CLAUDE.md` Commands entry missing | **Resolved** |

**1 — Resolved.** `scripts/verify-cache.mjs:1430` is now `if (fileList !== null && !changedOnlyScript) {`. Mutation-verified rather than read: I extracted `runPipeline`'s loop body with the same regex the new test uses, reverted the gate to `if (fileList !== null) {` in memory, and the new test's assertion goes false on the mutant, true on the real source. The added comment does not itself contain the matched text, so the test is not passing off its own prose.

**3 — Resolved in-repo.** `CLAUDE.md:901-915`, `docs/release-notes-next.md:488` and plan 156's addendum now say CI narrows via its own `vitest run --changed <PR base>`. Checked against `verify.yml:276` and `:326-327`: both legs take that branch whenever `github.event.pull_request.base.sha` is non-empty, i.e. always on a PR. `RELEASE_NOTES.md`'s user-facing line ("the full suite still runs before anything is allowed to ship") is accurate as written and needed no edit. The PR body was not touched — separate 🟡 below.

**5 — Resolved.** `CLAUDE.md:339`; content matches `package.json:43`/`:46` and `server/package.json:15`.

---

### 🟠 Significant (new, correctness) — the `!scopeShared` guard closes the *root* lockfile spelling and leaves `server/package-lock.json` armed: a server-lockfile-only commit still reports `[pass] test:server` having executed zero tests

`scripts/verify-cache.mjs:1408-1409`, against `computeShared` at `:834-839`

The fix keys the substitution off `computeShared`, whose entire membership test is:

```js
return p === 'package.json' || p === 'package-lock.json' || p.startsWith('.github/actions/');
```

That is a *scope* predicate, written to answer a different question ("does this diff put every step in scope?"). The question the substitution needs answered is "can vitest's `--changed` reach this file at all?" — and the two sets are not the same. `server/package-lock.json` sits in the gap: `computeShared` says false, but `test:server` declares it an input via `includeLockfiles: ['server']` (`:440`, matched at `:854-859`).

Verified live by importing the module — no mutation, no tree change:

```
diff=["server/package-lock.json"]  computeShared=false  test:server in scope=true
  -> pre-commit substitutes --changed for test:server? true
diff=["package-lock.json"]         computeShared=true   test:server in scope=false
  -> pre-commit substitutes --changed for test:server? false
```

`server/package-lock.json` matches none of the thirteen `forceRerunTriggers` entries in `server/vitest.config.ts:190-235` — `'{**/package.json,**/.*/**/package.json}'` does not match `package-lock.json`, and the remaining twelve are config / openapi / `scripts/**` / `pinokio-scripts/**` / `requirements/**` / `main.py` / `synthesise-chapter.ts` / doc paths. Nothing imports a lockfile, so there is no module-graph edge either. Neither vitest config sets `passWithNoTests`, so vitest 4's default `true` applies.

**Failure scenario — a shape this repo commits routinely, not a hypothetical.** Four commits reachable from this branch change **only** `server/package-lock.json`:

- `205930ab` — chore(deps): bump esbuild from 0.28.0 to 0.28.1 in /server
- `a2f35e9d` — chore(deps): bump form-data from 4.0.5 to 4.0.6 in /server
- `b62cb93e` — build(deps): refresh server lockfile to clear 3 npm advisories
- `78ebdf25` — chore(deps): refresh in-range server deps

Stage any of those file sets and commit:

1. No `[scope] shared-scope change detected` line (`computeShared` false); `test:server` is in scope via the lockfile branch at `:854-859`.
2. `[run] test:server (--changed HEAD only)` → `npm --prefix server run test:changed` → `vitest run --changed HEAD`.
3. Related files = `server/package-lock.json` alone. No trigger, no graph edge → zero specifications.
4. `passWithNoTests` → `No test files found, exiting with code 0`.
5. **`[pass] test:server`. Zero tests executed.**

That is verbatim pass 1's finding 2 — "a dependency bump is committed with zero tests executed, reported as a green test step" — surviving at the sibling path. The blast radius is smaller than pass 1's version, because the cache half *is* genuinely fixed (nothing is written, so pre-push still runs the full suite); this is commit-time only. It is still the gate reporting success having done nothing, which is the thing the gate exists to prevent.

Worth noting that `docs/release-notes-next.md:488` already states the correct criterion — *"`--changed` against a file no test's own dependency graph reaches (e.g. `package-lock.json`, **which matches no `forceRerunTriggers`**)"* — while the code implements the narrower `computeShared` one. The doc describes a guard the code does not have.

Two secondary members of the same class, named only so the fix targets the class rather than one more spelling: `server/tsconfig.json` (a `test:server` `extraFiles` member at `:424` — and `server/src/force-rerun-triggers.test.ts:166` *asserts* `tsconfig.json` matches no trigger) and root `index.html` (a `test` `extraFiles` member at `:348`). Both narrow to zero the same way.

The repo already maintains the set this needs. `extraFiles` exists *precisely because* those files have no module-graph edge — the comments at `:388-394` and `:396-421` say so repeatedly, citing "#1847's runtime-read trap" — and `forceRerunTriggers` is the curated mirror of it on the vitest side. `computeShared` is neither.

---

### 🟡 Minor — the PR body still carries the exact sentence finding 3 was raised against

The commit message says "Corrected the PR's own docs", and it did correct all three in-repo surfaces. The PR body itself was not edited and still reads:

> `--scope-branch` (pre-push) and CI are unchanged and still run everything, so anything `--changed` misses is still caught before merge.

CI is not unchanged-and-full; it has narrowed via `--changed <base>` since CI cost round 2 (`verify.yml:276`, `:326-327`). The body also still describes pre-fix behaviour throughout — it never says the narrowing is unshared-only or that a `--changed` pass is uncached, which are the two properties that make its own safety argument true. Merges here are merge commits, so this body is what the merge commit records. Lower stakes than `CLAUDE.md` (fixed), but it is the same false claim on the surface a reviewer reads first.

---

### 🟡 Minor — plan 156's `:50-52` bullet still stands false verbatim; the addendum corrects one of its two clauses, by reference, 25 lines below it

`docs/features/156-precommit-scope-contention.md:50-52`

> - **Pre-push unchanged.** Pre-push still runs the FULL `npm run verify` battery, preserving the "local is the full coverage net, CI is the scoped one" invariant.

The addendum's third bullet (`:76-81`) says *"The 'CI is the scoped one' line above is corrected, not new"* — but the line is not corrected, it is left standing, and a reader arriving at `:50` of a `status: active` plan gets the false statement with nothing pointing forward to the correction.

Separately, the addendum addresses only the *second* clause. **"Pre-push still runs the FULL `npm run verify` battery" is independently false and is not mentioned anywhere.** Pre-push has run `verify:fast:branch` since the 2026-07-06 rebalance — `.husky/pre-push:28` is literally `npm run verify:fast:branch`, and `CLAUDE.md`'s own Commit gate section says the same. Pass 1 quoted this whole bullet; half of it is still standing.

Invariant 2 (`:87-88`), the other statement pass 1 flagged in this file, **is** genuinely repaired — by the code fix, not the prose: with `!scopeShared` in place, "every selected step is in scope" once again means every selected step actually runs its tests.

---

### Checked and clean (delta only)

- **Both new tests are real regression tests, mutation-verified in both directions.** Reverting `!scopeShared` reddens the first and only the first; reverting `!changedOnlyScript` reddens the second and only the second. `runPipelineLoopBody()`'s extractor fails closed (`assert.ok`), and its match genuinely spans the cache write (1929 chars, `containsCacheWrite=true`) rather than stopping short of it. `node --test scripts/tests/verify-cache.test.mjs` → **122/122 pass**.
- The `!changedOnlyScript` cache gate leaves no stale-green hazard: reaching the run at all means `decide()` returned `'run'`, i.e. no entry matches `currentHash`, so declining the write cannot resurrect a matching entry.
- `CLAUDE.md:339`'s new Commands bullet matches the actual scripts and their semantics.
- `RELEASE_NOTES.md` needed no edit for this round; the release-notes pair is still present and `Closes #2834` is still outside any code span.
- No tracked file modified by this pass. `git status --porcelain` clean before and after; HEAD still `3bb761f1`.

### Summary

**1 new correctness finding (🟠), 2 cleanup findings (🟡).** Of pass 1's five: 2 fully resolved (1, 5), 1 resolved in-repo with the PR body outstanding (3), 2 partially resolved (2, 4).

dudarenok-maker:
## PR review — pass 3 (head a31d7ec7013dc7c8d2d2df2abf509dd8f4b695ab, depth low)

Re-review of `a31d7ec7` against pass 2's three findings, plus a fresh adversarial pass over the delta. Deltas only — passes 1 and 2's "checked and clean" lists are not re-litigated. This is the **final round under the 3-round cap**.

|---|---|---|
| 1 | 🟠 `computeShared` misses `extraFiles` / the server lockfile | **Partially resolved** — the two named spellings are closed; the class is left armed at the third (see below) |
| 2 | 🟡 PR body still carried the stale "CI still runs everything" claim | **Resolved** |
| 3 | 🟡 plan 156's "Pre-push still runs the FULL `npm run verify`" is independently false | **Resolved** |

**2 — Resolved.** The body now reads *"pre-push always runs the full, unsubstituted suite (CI separately narrows its own PR runs via `vitest run --changed <base>`, a pre-existing, unrelated design …)"*. Both halves check out: the substitution requires `flags.scopeStaged` (`scripts/verify-cache.mjs:1440`) and `verify:fast:branch` passes `--scope-branch` (`package.json:81`), so pre-push never substitutes; `verify.yml:276`/`:326-327` do take the `--changed "$BASE"` branch on every PR. The body also now describes the gate's actual three conjuncts rather than pre-fix behaviour.

**3 — Resolved.** `docs/features/156-precommit-scope-contention.md:50-54` now names `verify:fast:branch` and dates the correction. Matches `.husky/pre-push` and `package.json:81`.

**1 — the two spellings named in the finding are genuinely fixed, and the tests are real.** Mutation-verified in both directions, not read:

- Neutering `diffAvoidsUntrackedStepInputs` to `return true` reddens the extraFiles case **and** the server-lockfile case; removing *only* the `extras.has(p)` line reddens only the first; removing *only* the `server/package-lock.json` line reddens only the second. The two remaining cases (the no-`includeLockfiles` control, the empty-diff vacuous case) correctly stay green under all three mutants — they are controls, and they are labelled as such.
- The widened source-regex gate test reddens both on a full revert to the pass-2 expression **and** on the targeted removal of just the `diffAvoidsUntrackedStepInputs` conjunct, while the separate cache-write test stays green — the two pins are orthogonal. `runPipelineLoopBody()`'s extraction still spans the cache write (2364 chars, `cache.steps[step.name] =` present), and the gate regex does **not** match the surrounding comment prose in isolation, so it is not passing off its own documentation.
- `node --test scripts/tests/verify-cache.test.mjs` → **126/126 pass**.

Confirmed live against the real module that `server/package-lock.json` and `server/tsconfig.json` now decline the substitution, and that `index.html` / `openapi.yaml` do the same for the frontend `test` step.

---

### 🟠 Significant (new, correctness) — the class is closed at two spellings and left armed at the third: a step's runtime-read inputs declared as **globs** still get the substitution, and `test:server` declares most of its there

`scripts/verify-cache.mjs:879-888` (the new predicate) against `:382-395` and `:344` (the two substituting steps' input declarations)

`diffAvoidsUntrackedStepInputs` enumerates two of the three ways a `STEPS` entry can declare a file it reaches only through a non-source channel: `extraFiles`, and `includeLockfiles: ['server']`. `test:server` declares **most** of that surface the third way — as globs. Its own comment at `:360-395` says so explicitly, and names the motivating files:

> `*.{mjs,ts}` … a brand-new root-level `.mjs`/`.ts` file (not just `launch.mjs`/`vite.config.ts` …) — **#2716 PR review, pass 3, found that naming only the two currently-known root spawners left every OTHER root config file, e.g. `vitest.config.ts`, `playwright.config.ts`, `eslint.config.mjs`, uncovered**

Those files are in scope via the glob, are not `extraFiles`, and are not the server lockfile — so `diffAvoidsUntrackedStepInputs` returns `true` and the substitution fires. They also match **no** entry in `server/vitest.config.ts:190-235`'s `forceRerunTriggers`, and no server test imports them.

**Instance A — four real root files, live-verified.** Checked each candidate against the real trigger list with the same matcher `server/src/force-rerun-triggers.test.ts:51-52` uses (`picomatch`, all three `ROOT_SHAPES`); every one of that file's own `MAIN_COVERED`/`NOT_COVERED` cases reproduces, so the instrument agrees with the repo's own oracle before I trusted it on new input:

| root file | in `test:server` scope | substitutes after `a31d7ec7` | matches a `forceRerunTrigger` |
|---|---|---|---|
| `playwright.config.ts` | yes | **yes** | **no** |
| `playwright.marketing.config.ts` | yes | **yes** | **no** |
| `eslint.config.mjs` | yes | **yes** | **no** |
| `vitest.config.wire-fixtures.ts` | yes | **yes** | **no** |
| `vite.config.ts`, `launch.mjs`, `scripts/**`, `pinokio-scripts/**`, `server/tts-sidecar/requirements/**` | yes | yes | yes — harmless |

Then measured the actual selection, from `server/`:

```
$ npx vitest related ../playwright.config.ts --run -t "__no_such_test_name__"
 RUN  v4.1.11 C:/Claude/Projects/wt-verify-contention-scoping/server
No test files found, exiting with code 0
EXIT=0

$ npx vitest related ../eslint.config.mjs --run -t "__no_such_test_name__"
No test files found, exiting with code 0
EXIT=0
```

Control, same command form, proving the instrument can select: `npx vitest related src/force-rerun-triggers.test.ts …` → `Test Files 1 skipped (1)`, exit 0.

**Failure scenario — a shape with real commits on this branch, exactly as in pass 2.** Two commits reachable from here change **only** `eslint.config.mjs`:

- `2b010fb1` — chore: exclude wire-fixtures directory from ESLint linter
- `b07b3bb2` — chore(ops): exclude the token-saver scratch dir from lint too

Stage either file set and commit:

1. `computeShared(['eslint.config.mjs'])` → false, so no shared-scope line.
2. `stepTouchedByDiff` → true via the `*.{mjs,ts}` glob, so `test:server` is selected (`package.json:80` — `verify:fast:scoped` runs `test:hooks,check:budget-poll,test,test:server`).
3. `diffAvoidsUntrackedStepInputs` → **true**. `[run] test:server (--changed HEAD only)`.
4. No trigger, no module-graph edge → zero specifications → `passWithNoTests` → exit 0.
5. **`[pass] test:server`. Zero tests executed.**

The concrete guard lost: `server/src/spawn-windows-hide.test.ts:237-272` builds `EXTERNAL_FILES_FLOOR` by scanning **repo root, non-recursive, `.mjs`/`.ts`** at runtime — which is *why* the glob is written that way. Add a `spawnSync` without `windowsHide` to `playwright.config.ts` or `eslint.config.mjs` and the guard that exists to catch it is the one thing that does not run.

**Instance B — the same shape in the frontend `test` step, where the file is even in the graph.** `src/styles.css` matches the `src/**` glob (`:344`), is not an `extraFiles` entry, and matches none of `vitest.config.ts:85-116`'s triggers, so it substitutes. It is not zero tests — but it is the wrong one:

```
$ npx vitest related src/styles.css --run -t "__no_such_test_name__" --reporter=verbose
 ↓ src/main.test.tsx > main.tsx router wiring (fe-56) > mounts with react-router/dom's RouterProvider…
 Test Files  1 skipped (1)
```

The only edge is `main.tsx`'s CSS import. The two guards that police that file — `src/styles-neutrals.test.ts:10-11` and `src/test/dark-mode-css.test.ts:24-25`, both `readFileSync(… 'styles.css')` with no import — are **not** selected. `8ce2aa30` ("adopt brand v2 `--ink-soft` value") is a real `src/styles.css`-only commit on this branch, and `styles-neutrals.test.ts` is precisely the guard for the neutral ramp it edits.

**Blast radius is the same as pass 2's, and no larger:** commit-time only. `--scope-branch` still runs the full script, and the pass-1 cache fix means nothing is written, so pre-push remains an honest backstop. It is still the gate printing `[pass]` having run zero (or only irrelevant) tests, on diff shapes that occur in this repo's own history.

**On the fix, because the shape matters more than the site:** the predicate answers "is this file an `extraFiles` entry", but the question the substitution needs answered is "can vitest's own selection reach this file" — whose authority is `forceRerunTriggers` + the module graph, not `STEPS[].inputs`' declaration syntax. Keyed on the former, both instances close at once. Worth noting the corollary: because CI has narrowed via `vitest run --changed <base>` since CI cost round 2, these files are **already** uncovered on CI today, independent of this PR — so adding the missing triggers (four root configs to `server/vitest.config.ts`, `src/styles.css` to `vitest.config.ts`, each with its `force-rerun-triggers.test.ts` `MAIN_COVERED` row) fixes a live CI gap and this one in the same edit. What this PR changed is that pre-commit used to be the backstop that hid it.

---

### Checked and clean (delta only)

- The three-conjunct gate short-circuits in the right order: `scopeDiff !== null` is evaluated **before** `diffAvoidsUntrackedStepInputs(step, scopeDiff)`, so the git-failure path cannot throw on a `null` diff (`:1442-1443`). That conjunct also closes a case pass 2 did not name — a failed `git diff --cached` left `scopeShared` false and previously still substituted.
- Path normalisation is consistent: `stagedDiffFiles` already returns POSIX (`:907`), `extraFiles` are declared POSIX, and the `toPosix` on both sides is a harmless no-op rather than a mismatch. Verified empirically, not by inspection.
- The predicate is evaluated for every step, not just the two with a `CHANGED_ONLY_NPM_SCRIPT` entry — costless, and the ternary still yields `undefined` for the others.
- `CLAUDE.md:897-916`, `docs/features/156-…:84-94` and `docs/release-notes-next.md:488` all describe the mechanism as implemented (`extraFiles` + server lockfile), with no over-claim about what it covers. `RELEASE_NOTES.md` needed no further edit; `Closes #2834` still present and outside any code span.
- No tracked file modified by this pass. `git status --porcelain` clean before and after; HEAD still `a31d7ec7`. All probes ran from a scratch directory outside the repo.

### Summary

**1 new correctness finding (🟠), 0 cleanup findings.** Of pass 2's three: 2 fully resolved (the two 🟡 docs findings), 1 partially resolved — the two spellings it named are closed and well-tested, the third is not.

This is round 3 of 3. The loop stops here regardless of disposition; the remaining finding is handed to the shipping session to fix (and, per the cap, without a further automatic review pass).

Issue #2838, opened 2026-08-31
[chore] Title-text matching for register-row citations (v2 of check:register-row-citations)


dudarenok-maker:
## What

`npm run check:register-row-citations` (added by #2831/#2603) verifies only that an `A\d+`/`E\d+` row ID cited from `docs/testing/**`/`docs/features/**` currently *exists* as a register heading. It does not verify that the citing prose's *description* of the row still matches that row's actual current title -- a citation can point at a real, live row while describing work that row no longer covers (e.g. after a row's scope was edited or the row was repurposed).

This was explicitly deferred as v2 in the design decision for #2831 (see https://github.com/dudarenok-maker/Castwright/issues/2603#issuecomment-5484701257): "Title-text matching ... is a stronger, harder check, explicitly deferred as a follow-up backlog item -- not blocking v1."

## Why it's harder than v1

ID-existence is a mechanical lookup. Title matching requires deciding whether two natural-language descriptions (the row's current title vs. the citing prose's paraphrase of it) are "close enough" -- a fuzzy-match/similarity-threshold design question, not a lookup, and prone to false positives on legitimate paraphrasing.

## Scope (when picked up)

- Same file surface as v1: `docs/testing/**` + `docs/features/**`.
- Design the actual matching strategy (exact substring? token overlap? something else?) before implementing -- this needs its own decision pass, not just "ship a threshold."

## Not asking for

No code yet -- this is a backlog placeholder per CLAUDE.md's "net-new work" rule (a genuinely new capability, filed rather than built inside #2831's task).

---
_Body restored 2026-09-03: the original GitHub body was a dead reference to a local scratch file (`@C:\Users\dudar\AppData\Local\Temp\open-engine-scratch\claude-2831-...\backlog-issue-body.txt`), unreadable by anyone but the machine that filed it. This is the file's actual content, committed here so the issue is self-sufficient._

dudarenok-maker:
**Heads-up: an isolation violation was found and cleaned up in this lane's worktree today (2026-09-05).**

[...]

Issue #2831, opened 2026-08-31
[agent instructions][claude][task] Build A\d+/E\d+ citation checker v1


dudarenok-maker:
## Parent

#2603.

## What

Build a mechanical checker that validates `A\d+`/`E\d+`-style register-row citations found in `docs/testing/**` and `docs/features/**` markdown against the on-box acceptance register's actual current row headings, per the design decision in [this comment](https://github.com/dudarenok-maker/Castwright/issues/2603#issuecomment-5484701257).

## Background you need

- The problem this fixes: PR #2578's review rounds 19-21 found row-citation staleness across run sheets, plan docs, and even the register's own body (a self-reference) after two renumbering merges — 26+ stale citations in one run sheet alone, none caught by any existing mechanical check.
- `scripts/check-onbox-register.mjs` — the existing register-internal checker (row IDs vs. glance-table counts vs. total). This is a DIFFERENT, complementary check: this one validates *external citing surfaces* against the register, not the register's own internal arithmetic. Look at this script for style/conventions to follow (error message shape, exit codes, how it's wired into `package.json`/CI).
- `docs/testing/onbox-acceptance-register.md` — the source of truth for "what row headings currently exist" (parse its `A\d+`/`E\d+` row headings).
- Related: #2599 (parent's sibling issue) is building per-row content hashing in the SAME script family — share row-heading-parsing logic with that work where natural if it's already landed on `main` by the time you start (check `git log`), but do not block on it; duplicate the minimal parsing you need if it hasn't landed yet.

## The fix

1. Write a new script (or a new `--mode` on `check-onbox-register.mjs` — your choice, whichever is less code) that:
   - Parses `docs/testing/onbox-acceptance-register.md` for its current set of row IDs (headings like `### A41` or however rows are actually delimited — read the file to confirm the real heading shape before assuming).
   - Scans every `.md` file under `docs/testing/**` and `docs/features/**` for citations matching a defined pattern: `\b(?:row|Row)\s+([AE]\d+)\b` and/or a markdown link into `onbox-acceptance-register.md#a\d+`/`#e\d+`. A bare `A41`-shaped token with neither context is NOT a citation — do not flag it.
   - For each recognized citation, checks the cited ID exists as a current row heading in the register. Missing → collect a failure naming the citing file, line, and missing ID.
[...]

PR #2840, opened 2026-09-01
chore(ops): add check:register-row-citations for docs/testing+features

### LINKED ISSUE #2603: No mechanical guard validates A\d+/E\d+ row references against onbox-acceptance-register.md's actual current headings

dudarenok-maker:
## Summary

Adds `npm run check:register-row-citations` (#2831), a v1 mechanical checker that scans `docs/testing/**` and `docs/features/**` markdown for `row A41`/`Row E3`-style citations or markdown links into `onbox-acceptance-register.md#a41`, and fails when a cited row ID has no current heading in the register. Scoped per #2603's decision comment: pattern-gated matching (no bare-token false positives), self-reference-aware for the register's own file, ID-existence-only resolution (title-text matching deferred to #2838), GitHub issue bodies out of scope. Closes #2603.

## Test plan

- [x] `node --test scripts/tests/check-register-row-citations.test.mjs` — 28/28 pass, including the mutation-provable red/green pair (citing an existing row passes; removing that row from the register flips the same citation to a failure).
- [x] Live-tree run of `npm run check:register-row-citations` against `main` — OK, 37 citations across 348 files resolve; 5 pre-existing discharge-annotated citations correctly print as notes, not failures.
- [x] Manual mutation against the real (git-tracked) tree, not just the unit-test mocks: staged a fixture under `docs/testing/` citing a nonexistent row (`row A99999`) alongside a bare non-citation token (`A41` with no "row"/link context) in the same file, then ran the checker directly. Observed:
  - `check:register-row-citations: 1 citation(s) cite a row that does not exist in docs/testing/onbox-acceptance-register.md:` / `- docs/testing/_oe2830-mutation-check.md:3 cites row A99999, which has no current heading.` (exit 1)
  - The bare `A41` token on the next line was not flagged (citation count stayed at 1, only the `row A99999` citation).
  - Fixture unstaged and removed afterward; verified `git status --short` clean before opening this PR.
- [x] `node scripts/check-register-citations.mjs` (sibling, older checker) — OK, confirms this PR's fix to its `SELF_REFERENTIAL_PATHS` exclusion set keeps it green now that the new checker's own source/tests are visible to its scan.
- [x] `npm run check:register-row-citations` documented in `docs/testing/onbox-acceptance-register.md`'s "Live view" section, alongside `check:onbox-register`.
- [x] Confirmed follow-up backlog issue #2838 ("Title-text matching for register-row citations (v2)") is already filed with `type:chore`, per CLAUDE.md's net-new-work rule.

## Also fixed, found in passing (review-gate, 2 rounds)

The mandatory `pr-review-gate` pass (2 rounds so far) surfaced and fixed the following, all committed on this branch:

- **CI-wiring gap**: `check:register-row-citations` was reachable only via manual invocation — not in `verify.yml`, `verify-cache.mjs`'s `STEPS[]`, or any husky hook. Wired into `test:hooks` (globs now cover `docs/features/**` alongside the pre-existing `docs/testing/**`), with a real-tree CLI regression test.
- **Non-mutation-proof tests**: three of the checker's own "mutation-proven" tests didn't actually catch a "scans zero files" regression (a loose regex, a disabled guard, a direction-blind comparison). Rewrote all three to assert on the actual parsed citation count.
- **Unterminated-fence fail-open**: `stripFences`'s `unterminatedFenceLine` signal was discarded, reintroducing the ops-44/#1913 "truncated read reports no errors" defect. Now surfaced as a fatal error, matching `check-onbox-register.mjs`'s own handling.
- **Annotation exemption cross-contamination + hard-wrap bug**: the same-line discharge/removal annotation exemption first used a fixed proximity window (still cross-contaminated between citations on the same line), then was redesigned around a "nearest citation to the left" rule with logical-line joining, so hard-wrapped (including blockquote-wrapped) citations are handled correctly too.
- **Five minor cleanup items**: a misleading file count in the OK summary, register prose overclaiming scan coverage, a hand-rolled percent-decoding-blind path helper (replaced with `node:url`'s `fileURLToPath`), an export name colliding with the sibling checker's own export, and missing release-notes entries — plus a follow-up pass that made the `fileURLToPath` regression test actually exercise the checker (it previously only checked that Node's own import existed) and corrected four surfaces that still described the retired same-line-only annotation rule.

See the PR's review-gate comments for full detail on each round.


dudarenok-maker:
## PR review — pass 1 (head e2fe4315a114ba82f8aac6397d4d35e9ae992256, depth high)

Scope: all 6 changed files (`scripts/check-register-row-citations.mjs`, its paired test, `scripts/check-onbox-register.mjs`'s `stripFences` export, `scripts/check-register-citations.mjs`'s exclusion set, `package.json`, `docs/testing/onbox-acceptance-register.md`) · verified in a throwaway detached worktree at the head SHA: `node --test scripts/tests/check-register-row-citations.test.mjs` 13/13 pass; `node scripts/check-register-row-citations.mjs` exits 0 (36 citations / 381 files / 5 annotated notes); `node scripts/check-onbox-register.mjs` exits 0; `node scripts/check-register-citations.mjs` exits 0 — so the `stripFences` export and the sibling-exclusion fix are both non-breaking. Then adversarially probed the citation regexes, the annotation exemption, the self-reference exclusion, the fence handling, the scan layer, and the suite's mutation-provability (6 mutants, control included).

---

### 🔴 Blocking — the checker executes against real repo content in **no** automated lane, and its test suite stays green when the scanner returns zero files

`scripts/check-register-row-citations.mjs` is reachable only through `npm run check:register-row-citations`. A grep for both the script path and the npm-script name across the whole tree at this head returns exactly three non-self hits: `package.json:60`, the new register prose, and the sibling's `SELF_REFERENTIAL_PATHS` entry. It is in **no** `verify.yml` leg, **no** `verify-cache.mjs` `STEPS[]` entry (`check:budget-poll` is at :312 and `check:cycles` at :499 — there is no third), **no** `--steps` CSV in `package.json` (`verify:fast`, `verify:fast:scoped`, `verify:fast:branch`), and **no** husky hook.

The only thing that touches its code at all is `test:hooks`, via the new test file — and that path is provably inert. Mutation, run for real:

```
scripts/check-register-row-citations.mjs:131
  export function scannedFiles() {
+   return [];                       // checker now scans ZERO files
    const files = new Set();
→ node --test scripts/tests/check-register-row-citations.test.mjs  →  pass 13, fail 0
```

Every one of the 13 tests injects both `files:` and `readFile:`, so nothing in the suite ever reaches `scannedFiles()`, `gitLsFiles()`, `repoRoot()`, `REGISTER_PATH`, or the CLI. Three more mutants confirm the whole scan layer is uncovered — each survives all 13 tests:

| mutation | result |
|---|---|
| `scannedFiles()` → `return []` | **survives** |
| `SCAN_PREFIXES` → `['docs/testing/']` (drops `docs/features/**` entirely) | **survives** |
| `REGISTER_PATH` → `'docs/testing/NOPE-does-not-exist.md'` | **survives** |
| `stripFences(...)` → raw text (fence handling disarmed) | **survives** |
| *(control)* annotation exemption removed | killed, fail 1 |

The control kills, so the harness can fail — the gap is real coverage, not a broken runner.

Compare the two things this PR positions itself alongside. `check:onbox-register` is wired into `.github/workflows/onbox-register-check.yml:41`. The sibling `check:register-citations` is not in CI either, **but it is exercised against the real tree from inside its own test file** (`scripts/tests/check-register-citations.test.mjs:1710`, `'CLI: the real tree scan finds at least one binary file, and the success line reports it (finding U)'`, plus the real-tree injection at :211/:1574/:1596) — and `test:hooks` runs in pre-commit, pre-push and `verify.yml`. That is exactly the mechanism this PR omits.

Concrete consequence: a contributor adds `See register row A99` to a new `docs/features/foo.md` tomorrow. Nothing fails — not pre-commit, not pre-push, not `verify.yml`, not `test:hooks`. #2603's title is *"No mechanical guard validates A\d+/E\d+ row references…"*; a guard that runs zero times as shipped does not close it. This is the catalogue's shape 7 (unreachable at the default configuration) sitting on top of shape 6 (success reported while doing nothing).

Minimum remedy is cheap and already has a template next door: one real-tree test in `scripts/tests/check-register-row-citations.test.mjs` that calls `checkRegisterCitations()` with no injection and asserts a non-trivial `checkedFiles`/`citationCount` floor, mirroring the sibling's. That single test kills all four mutants above.

---

### 🟠 Significant — an unterminated code fence silently deletes every citation below it

`scripts/check-register-row-citations.mjs:209` destructures only `{ text }` from `stripFences` and discards its second return value, `unterminatedFenceLine`.

`stripFences` treats an open-at-EOF fence by blanking every line after it. The module this PR imports it from does not tolerate that — `check-onbox-register.mjs` bails with a hard error, and its own comment says why in as many words:

> *an unterminated fence blanks everything after it, which must be surfaced as an error rather than silently validating a truncated document (ops-44, issue #1913 review finding: this previously made a stray fence line make the rest of the register invisible to every check below, reporting "no errors" over a truncated read)*

This PR reuses the helper and reintroduces the exact defect that comment records as already fixed once. Repro:

~~~~
docs/testing/stray.md:
    See register row A1 (above the stray fence).

    ```                  <-- opened, never closed
    some example

    See register row A99999 -- a genuinely dead citation BELOW the stray fence.
~~~~

→ `failures: []`, `citationCount: 1` (A1 only). The dead A99999 is invisible; the checker reports OK.

The mirror direction is noisy rather than silent but equally wrong: a stray fence in the **register itself** blanks every `### <ID>` heading below it, so `validIds` shrinks and every downstream citation turns into a spurious failure. Verified: register with a stray fence above `### A1` → `failures: [{file: 'docs/testing/plan.md', line: 1, id: 'A1'}]` for a citation of a row that genuinely exists.

Not hypothetical in this corpus: **5 tracked `.md` files already carry an unterminated fence** — `docs/superpowers/plans/2026-06-11-fs44-mcp-agent-surface.md:2830`, `.../2026-06-12-companion-marketing-capture.md:1503`, `.../2026-06-14-srv-1-merge-journal-alias-unlink.md:1090`, `docs/superpowers/specs/2026-06-15-pinokio-installer-design.md:252`, `.../2026-06-18-lan-browser-device-auth-design.md:494`. All five happen to sit outside the two scanned prefixes today, so the shape is live in this repo's docs and merely has not landed inside `docs/testing/**` or `docs/features/**` yet.

Note the fence path has no coverage in **either** direction — mutant M1 above (disarm `stripFences` entirely) survives all 13 tests, so no test asserts that a fenced worked example is *not* a citation either.

---

### 🟠 Significant — the discharge exemption is keyed on the physical line, not on the citation, and one live time bomb is already planted

`ANNOTATION_REGEX` (:92) is tested against `lines[line - 1]` (:215) — i.e. *any* citation on a line that contains "discharge(d)" / "no longer exist(s)" / "removed from the register" is excused, regardless of whether that word is about that citation or forward-looking.

**(a) Forward-looking prose — live today.** `docs/testing/sidecar-evict-latency-onbox-acceptance.md:79` reads:

> `_(Once run, mark the register row A17 discharged with a summary of this`

A17 is a **current, owed** row, so the citation resolves and the exemption never fires — the checker is green. The day A17 is discharged and removed from the register, which is precisely what that sentence instructs, the citation goes dangling and the checker will silently downgrade it to *"annotated as discharged/removed"* because the word "discharged" happens to be on the line. That is the exact drift #2603 exists to catch, pre-armed and dated.

I measured the whole scanned corpus for this shape — **4 currently-live citations already sit on a line carrying a discharge word**, each of which converts to a silent pass the moment its row is discharged:

```
docs/features/INDEX.md:406                                     -> A8
docs/features/archive/222-gpu-residency-and-analysing-honesty.md:78 -> A8
docs/features/archive/283-castwright-local-rebind.md:9         -> E10
docs/testing/sidecar-evict-latency-onbox-acceptance.md:79      -> A17
```

Same mechanism, purely prospective repro: `Before we can discharge row A99999 we need a live GPU box.` → `failures: []`, excused. The row is *owed*, not discharged.

**(b) One line, two citations.** `Row A43 was discharged 2026-08-26; the work now tracked under row A99999 instead.` → both A43 and A99999 land in `annotated`, `failures: []`. A99999 carries no annotation of its own; it is excused by its neighbour's. (0 multi-citation lines exist in the corpus right now, so this half is prospective — but it costs nothing to close alongside (a).)

The test at "a discharge word on a DIFFERENT line than the citation does not exempt it" pins the *inter*-line case and reads as though it settles the question; neither of the two intra-line shapes above is covered. A citation-scoped rule — e.g. requiring the annotation to follow the matched ID rather than merely share its line — kills (b) outright and narrows (a).

---

### 🟡 Minor

**1. The success line counts 33 files it never opened** — `scripts/check-register-row-citations.mjs:222` sets `checkedFiles: files.length` from `scannedFiles()`'s raw length, but `isFrozenPath` skips inside the loop at :207. Measured on this tree: 381 reported, **33 frozen and never read**, 348 actually scanned. So :262-263 prints `"36 citation(s) across 381 file(s) under docs/testing/** and docs/features/** all resolve"` — a resolution claim covering `onbox-acceptance-staleness-audit.md` and all 32 files of `onbox-wave{3,4,5}-results/`, which it deliberately skipped. Report the scanned count, or report both.

**2. The register prose says "every `.md` file", which is false** — `docs/testing/onbox-acceptance-register.md:119-120`: *"it scans every `.md` file under `docs/testing/**` and `docs/features/**`"*. 33 files are silently excluded by `FROZEN_EXACT`/`FROZEN_PREFIXES`, including every wave-3/4/5 run transcript. The paragraph documents the discharge-annotation exemption but never the frozen-path one, so an operator reading the register believes the wave-results transcripts are covered when they are not. The same paragraph also presents the checker beside `check:onbox-register` without noting that, unlike that one, nothing runs it automatically (see the 🔴 above).

**3. `fileURLToPathSafe` is a hand-rolled, percent-decoding-blind reimplementation of `node:url`'s `fileURLToPath`** — `scripts/check-register-row-citations.mjs:111-117`. `URL.pathname` is percent-encoded, so measured side-by-side:

```
input   file:///C:/Users/John%20Doe/Castwright/scripts/x.mjs
shipped fileURLToPathSafe -> C:/Users/John%20Doe/Castwright/     ← does not exist
node:url fileURLToPath    -> C:\Users\John Doe\Castwright\
```

A checkout under any path containing a space (`C:\Users\First Last\…`, an `OneDrive - Org` folder) makes `repoRoot()` return a nonexistent directory, and `execFileSync('git', …, { cwd })` at :121 throws ENOENT — the checker dies instead of reporting. **Latent, not currently reachable**: neither this box (`C:\Claude\Projects\…`) nor GitHub's runners (`/home/runner/work/…`, `D:\a\…`) have a space, so I am flagging it as duplicated-code-the-work-exposed rather than a live break. It matters because `check-onbox-register.mjs:31` — the very module this file imports `stripFences` from — already does `import { fileURLToPath } from 'node:url'`, and `scripts/lib/is-main-module.mjs`'s header is an extended warning against exactly this class of hand-rolled file-URL↔path conversion (#2291). One import replaces both functions.

**4. The exported entry point is named `checkRegisterCitations`** — `:195`, i.e. the *sibling* script's domain name. The header spends three paragraphs insisting the two scripts must never be conflated, folded, or renamed into each other, then names its main export after the other one; `import { checkRegisterCitations } from './check-register-row-citations.mjs'` reads as the older checker. The CLI wrapper is correctly `runCheckRegisterRowCitationsCli`, so the module is internally inconsistent with itself. `checkRegisterRowCitations` costs one rename plus one line in the test's import.

**5. Release-notes pair missing with no stated not-applicable** — neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the PR body's test plan does not mark step 5 N/A. Precedent cuts against skipping: the sibling checker (#2629/#2630) *does* carry entries in `docs/release-notes-next.md` (2 hits), and a new `npm run` command is an operator-visible delta. Either land the pair or say explicitly why not.

---

### ✅ What is solid

- The **red/green pair is genuinely mutation-provable** at the level it operates: `REGISTER_TEXT.replace(/### E1 · another thing\n\nSome body text\.\n/, '')` cannot no-op silently — a failed match leaves `after.failures` empty and the assertion red. My M6 control confirms the harness kills a real mutation.
- The **self-reference exemption is correctly narrow**, and the test proves the narrowing rather than just the exemption: a row citing a *different* row from inside its own body still fails (`'the register citing a DIFFERENT row from inside another row\'s body is not exempted'`). `rowIdAtLine`'s heading regex matches `parseAllRowHeadings`'s exactly, so the two cannot drift apart on a heading shape.
- **No false-positive on bare tokens**, verified against the real corpus, not just the fixture: 33 backticked and 63 parenthetical `A\d+`-shaped tokens exist under the scanned prefixes, 27 of them naming non-current IDs, and none is flagged. The pattern gate does what it claims.
- **The fix commit is correct and necessary**, and I reproduced the reason for it: the sibling's real-tree scan does become visible to the new script's fixtures once tracked, and both checkers exit 0 at this head.
- **Exporting `stripFences` breaks nothing** — `check-onbox-register.mjs` still exits 0, and its `--against-published`/baseline machinery is untouched.
- **Shipping under a distinct script name** rather than shadowing `check:register-citations` is the right call, and the header's rationale for it is worth keeping.
- The two documented deviations from #2603's decision text (annotation exemption, frozen paths) were both **forced by a real-tree run and mirror the sibling's existing exclusions** rather than being invented — the right way to widen a spec.

### Verdict

**Do not merge as-is.** One 🔴 and two 🟠. The 🔴 is the load-bearing one: as shipped this is a correct checker that never runs, and its own suite cannot tell the difference between "scanned 348 files" and "scanned none" — so the PR closes #2603 in name only. That one plus the two 🟠 fail-opens (fence, line-scoped annotation) all sit in the same blind spot: everything below the injection seam is untested. A single real-tree test plus a fence-aware guard would close the 🔴 and give both 🟠 fixes something that can actually go red.

Findings split for triage: **correctness bugs** — 🔴 1, 🟠 2, 🟠 3 (re-review re-triggers once fixed and pushed). **Cleanup** — 🟡 1-5.

dudarenok-maker:
## PR review — pass 2 (head 8ba58d7a139b03111736d81b4919ac5f4fccc6d0, depth high)

Deltas against [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2840#issuecomment-5516828158). Each prior finding re-derived independently against the new head, not read off the commit messages.

**Result: 3 of 8 prior findings are still open, plus 5 new findings** — 4 correctness bugs, 4 cleanup.

---

#### 1. 🔴 The checker runs in no automated lane — **STILL OPEN** *(correctness)*

Commit `4228dc65` added real-tree CLI tests to a file `test:hooks` globs. That is the right shape, but it does not close the finding, for two independent reasons.

**1a — `docs/features/**` is an input to no verify step, so half the scan surface has no lane at all.**

`SCAN_PREFIXES` (`scripts/check-register-row-citations.mjs:78`) is `['docs/testing/', 'docs/features/']`. `docs/testing/**` happens to be a `test:hooks` input already (`scripts/verify-cache.mjs:170` — added for `review-gate-mechanism.test.mjs`, not for this). `docs/features/**` is an input to **no step in `STEPS[]`**.

Measured, not asserted — `computeScopes` from `scripts/ci-scope.mjs`:

```
docs/features only          => legs ON: (NONE)
docs/features/archive only  => legs ON: (NONE)
docs/testing only           => legs ON: step_test_hooks
```

Concrete failure: someone edits `docs/features/194-voice-cloning.md` to cite `row A9999`.
- pre-commit (`verify:fast:scoped`, `--scope-staged`): no step's globs/`extraFiles` match → `test:hooks` skipped as out of scope.
- pre-push: `scripts/is-docs-only-push.mjs:29` (`path.startsWith('docs/')`) → docs-only → `verify:fast:branch` skipped **entirely**.
- cloud `verify.yml:195`: `if: …step_test_hooks || …shared` → both false → leg not scheduled.

Green everywhere, dead citation merged. This is verbatim the `#1847` trap that `verify-cache.mjs`'s own `docs/testing/**` entry — the glob immediately above the gap — spends seven lines documenting. Fix is one more glob entry alongside it, plus `scripts/check-register-row-citations.mjs` itself is already covered by the `scripts/**/*.{mjs,…}` glob, so only the corpus side is missing.

**1b — the new real-tree tests cannot fail on the mutation that matters.** See NEW-1 below; I am counting it under this finding because it is why 1a's "but the CLI test would catch it" argument does not hold either.

#### 2. 🟠 `stripFences`'s `unterminatedFenceLine` discarded — **VERIFIED RESOLVED**

`scripts/check-register-row-citations.mjs:260-261` (register) and `:283-284` (per scanned file) both inspect it and bail with a non-null `error`; `runCheckRegisterRowCitationsCli` exits 1 on it. `stripFences` (`scripts/check-onbox-register.mjs:72-95`) does return `unterminatedFenceLine: null` when balanced, so the `!== null` test is correct. `isFrozenPath` is checked at `:280`, *before* the fence check, so a frozen transcript with a stray fence can't hard-fail the run. Matches the ops-44/#1913 precedent. No residual.

#### 3. 🟠 Annotation exemption scoped to the whole line — **STILL OPEN** *(correctness)*

`isCitationAnnotated` (`:104-152`) narrows the blast radius from "anywhere on the line" to "±50/10 chars of the citation" — but the defect class is intact, and reproduces with a one-line fixture. Run against the real module at this head:

| line | dead ids | result |
|---|---|---|
| `See row A9999 for the plan; row A88 was discharged 2026-08-26.` | A9999, A88 | **both** exempted |
| `Still owed: row A9999. Separately, row A77 was discharged.` | A9999, A77 | **both** exempted |
| `row A9999 is still owed and row A8888 was discharged.` | A9999, A8888 | **both** exempted |
| `See [x](onbox-acceptance-register.md#a9999) and row A88, discharged 2026-08-26.` | A9999, A88 | **both** exempted |

Row 3 is the exact mirror of the new test at `scripts/tests/check-register-row-citations.test.mjs:204` — that test puts the annotated citation **first** and the unrelated one **second**, which the forward window happens to get right. Swap the order and the bug is back. One direction was fixed; the class was left armed.

The window also mis-fires the other way, hard-failing legitimate annotations:

| line | result |
|---|---|
| `register row A9999 (see the run sheet at docs/testing/foo.md) - discharged 2026-08-26.` | **FAILS** (annotation 46 chars out, window is 50 counted from the id) |
| `That row was discharged and removed; see row A9999 in the history.` | **FAILS** (annotation before, >10 chars) |

The error message (`:354`, "cites row A9999, which has no current heading") never mentions that a nearby annotation would have exempted it, so a doc author hits a hard CI failure with no route to the rule. This PR already paid that cost once: `docs/testing/onbox-sitting-plan.md:74-75` was rewritten in `bcaac3a7` purely to drag the word "discharged" inside the window, producing a 100+ char line in a file otherwise wrapped at ~76.

A character count is the wrong boundary. The natural fix is to bound the annotation search at the **nearest neighbouring citation token** on the line (or at the enclosing clause, which is what the sibling `check-register-citations.mjs`'s multi-line clause-boundary machinery already does), so an annotation can never reach past another citation.

#### 4. 🟡 `checkedFiles` counted frozen-skipped files — **VERIFIED RESOLVED**

`checkedFilesCount++` (`:281`) now sits after the `isFrozenPath` `continue` (`:280`); paired test at `scripts/tests/check-register-row-citations.test.mjs:357`. Live run prints `348 file(s)`, down from the 381 the PR body still quotes (see NEW-4).

#### 5. 🟡 Register prose overclaimed "every .md file" — **VERIFIED RESOLVED**

`docs/testing/onbox-acceptance-register.md:119-122` now names the frozen-path exclusion explicitly, with both path shapes. (A *different* sentence in the same paragraph is now stale — see NEW-3.)

#### 6. 🟡 Hand-rolled `fileURLToPathSafe` — **RESOLVED in code, but its paired test is a no-op**

`import { fileURLToPath } from 'node:url'` at `:70`, used at `repoRoot()`. The hand-rolled version survives only in the untouched sibling `check-register-citations.mjs:1645`, which is out of scope. Code half: resolved. Test half: see NEW-3.

#### 7. 🟡 Export name collided with the sibling checker — **VERIFIED RESOLVED**

`git grep` over the whole tree at this head: the new module exports `checkRegisterRowCitations` / `runCheckRegisterRowCitationsCli` only; every one of the 15 call sites in the test file uses the new name; the sibling keeps `checkRegisterCitations` / `runCheckRegisterCitationsCli` untouched. No missed call site.

#### 8. 🟡 Missing release-notes pair — **STILL OPEN** *(cleanup)*

Both entries exist but landed in the wrong section of their file.

- `RELEASE_NOTES.md:448` — the file is newest-first (`# Castwright 1.15.0` at line 1). The new bullet was appended to the **end of the file**, which puts it inside `# Castwright 1.5.0 — the first full Castwright` (line 436). It now reads as a v1.5.0 feature. CLAUDE.md step 5: "a matching user-facing, brand-voice line to **the in-progress version section at the top**". (Separately: the bullet is technical register — `check:register-row-citations`, `docs/testing/**`, `check:onbox-register` — in a file whose neighbours are "Your book never leaves the house." Worth asking whether a repo-internal docs lint has a shippable delta at all; step 5's own carve-out for "internal chore with no user- or operator-visible effect" looks like it applies, in which case the right move is an explicit not-applicable rather than a misplaced bullet.)
- `docs/release-notes-next.md:566` — appended **after** the `**Full changelog:** v1.14.0...v1.15.0` footer at line 564. That file's own header comment (line 12) fixes the footer as the terminator of the body, and `bump-version.mjs` feeds this file **verbatim** as the release body — so the published v1.15.0 release will render a stray bullet below its own changelog link.

---

### New findings

#### NEW-1. 🔴 The three "mutation-proven" real-tree tests all stay green when the scan is neutralised *(correctness)*

This is the pass-1 blocking finding — "the suite was misleadingly green because it always injected fakes" — reproduced inside the fix for it.

Neutralised `SCAN_PREFIXES` to `[]` in a scratch worktree at this head (checker then scans zero files and can never detect any drift) and ran the suite:

```
check:register-row-citations: OK -- 0 citation(s) across 0 file(s) …   EXIT=0
ℹ tests 24   ℹ pass 24   ℹ fail 0
```

Same result for `SCAN_PREFIXES = ['docs/features/']` (drops `docs/testing/**`, where the register itself and most citations live): **24/24 pass.**

Three independent reasons, each fixable:

- `scripts/tests/check-register-row-citations.test.mjs:399` — `assert.match(result.stdout, /\d+ citation\(s\) across \d+ file\(s\)/)`. `\d+` matches `0`. The comment directly above it (`:398`) says "Assert that the scan actually found citations (not scanning zero files)"; the assertion does not do that. Needs a floor: parse the two numbers and assert both `> 0` (better: assert a plausible minimum, e.g. `>= 20` citations / `>= 100` files, so a partial collapse is caught too).
- `:402` — the `scannedFiles()`-empty mutation test is gated on `if (citationCount > 0)` (`:415`), where `citationCount` is read from the **live** baseline. When the live code is the thing that's broken, the guard is false and the entire test body is skipped, silently. A test whose only defence against the regression is disabled *by* that regression. (`:442` has the same shape one level down: `if (!baseline.stdout.includes('…OK')) return;`.)
- `:465` — the `SCAN_PREFIXES` mutation asserts only `notEqual(mutantResult.stdout, baseline.stdout)`. With a neutralised baseline the mutant scans *more*, so the outputs differ and the assertion is satisfied by the mutant being **less** broken than production. It needs to assert a direction (mutant file count strictly lower than baseline), not mere difference.

Note this is the seam these CLI tests are the *only* cover for: a broken `ROW_WORD_CITATION_REGEX` is caught by the pure-function test at `:256`, but `scannedFiles()` / `SCAN_PREFIXES` / `gitLsFiles` are exercised nowhere else.

#### NEW-2. 🟠 A hard-wrapped citation either false-fails or is silently missed *(correctness)*

`ROW_WORD_CITATION_REGEX` (`:81`) uses `\s+`, which spans newlines, but `isCitationAnnotated` is handed a **single** line — `lines[line - 1]` at `:296`. In a corpus wrapped at ~76 columns (which `docs/testing/**` is), that splits two ways. Both reproduced against the real module at this head:

**Plain wrap — false failure.** Body `The old register row\nA9999, discharged 2026-08-26 and removed from the register.`
→ `extractCitations` returns `[{line: 1, id: 'A9999'}]` (the `\s+` swallowed the newline, and the reported line is the one holding `row`, not the id), but `lines[0]` is `"The old register row"` — no id, no annotation — so:
`FAIL: [{file, line: 1, id: "A9999"}]`, `ANNOT: []`. A correctly annotated citation hard-fails, and the reported line number points at the wrong line.

**Blockquote wrap — silent miss.** Body `> The old register row\n> A9999, discharged 2026-08-26 …`
→ `extractCitations` returns `[]`. `\s+` cannot cross the `> ` prefix, so the citation is never extracted, never resolved, never reported. `FAIL: []`, `ANNOT: []`, exit 0. This is the fail-open direction: a genuinely dead citation in a wrapped blockquote is invisible to the checker. `docs/testing/onbox-sitting-plan.md`'s correction notes — the file this PR edited — are exactly that shape.

The `testLine = line.replace(/\n/g, ' ')` at `:107` is dead code (an array element from `split('\n')` never contains `\n`) and reads as the author having intended multi-line input here. Either feed `isCitationAnnotated` a window of the surrounding lines and record the id's own line, or make the citation regex line-anchored so the wrapped case fails loudly instead of silently.

#### NEW-3. 🟡 The Finding-3 paired test asserts nothing about the code under test *(cleanup)*

`scripts/tests/check-register-row-citations.test.mjs:373-381`:

```js
const { fileURLToPath: importedFileURLToPath } = await import('node:url');
assert.strictEqual(typeof importedFileURLToPath, 'function', …);
```

It asserts that Node's standard library exports a function. It never calls `repoRoot()`, never constructs a percent-encoded path, and passes identically if the checker still used the hand-rolled `fileURLToPathSafe`. Pass 1's fix summary claims "with a paired test covering a percent-encoded path" — that claim is not true of this test. Either give it a real assertion (e.g. run the CLI from a path containing `%20` / a space and assert `git ls-files` still resolves) or drop it and say the change is covered by the CLI tests.

#### NEW-4. 🟡 Four surfaces still describe the retired same-line semantics *(cleanup)*

`bcaac3a7` replaced "the annotation is anywhere on the physical line" with a proximity window, but every document describing the rule still states the old one:

- `scripts/check-register-row-citations.mjs:53-59` — "**Same-line** discharge/removal annotation exemption … when its own **physical line** also says … this narrow, **same-line-only** rule". Also "Measured against all **8** real hits" — the live run at this head reports **5**.
- `scripts/check-register-row-citations.mjs:253-254` — the `checkRegisterRowCitations` JSDoc: "carries a **same-line** discharge/removal annotation".
- `docs/testing/onbox-acceptance-register.md:127` — "A citation whose **own line** also says the row was discharged/removed is printed as a note".
- `RELEASE_NOTES.md:448` and `docs/release-notes-next.md:566` — "on the same line" / "on lines that name".

A maintainer reading any of these will write an annotation 200 chars from the citation and get a hard failure. CLAUDE.md's "a stale comment or doc sentence your change made false" — a chore, fixed this round.

#### NEW-5. 🟡 PR body carries pre-fix figures and does not declare the incidental fixes *(cleanup)*

- "13/13 pass" — the suite is now 24 tests.
- "36 citations across **381** files" — the live figure at this head is **348**; 381 is the pre-fix count that prior finding 4 corrected. The body cites the bug as evidence the fix works.
- Eight fixes landed in `4228dc65`…`8ba58d7a` in response to pass 1 (a real-tree test lane, the fence bail-out, the annotation narrowing, five minors, a doc edit to `onbox-sitting-plan.md`). None appear in the body. CLAUDE.md: incidental fixes are "**declared in the PR body** ('Also fixed, found in passing: …') — an unannounced fix reads as scope creep to a reviewer."

---

### House gates — checked

- `Closes #2603` present, in prose, outside any code span. ✓
- #2603 is `type:chore` → no `docs/BACKLOG.md` row is owed, and its absence is not drift. ✓
- Follow-up #2838 (title-text matching v2) is genuinely filed and open, `type:chore`. ✓
- On-box acceptance: N/A — no hardware-provable behaviour. ✓
- `docs/features/INDEX.md`: N/A — no new or moved plan doc. ✓
- No `cast.json` writes, no config knob, no derived artifact (`api-types.ts` / brand PNGs) touched. ✓
- Release-notes pair: present but misplaced — see finding 8.
- Live tree is green at this head: `OK -- 36 citation(s) across 348 file(s)`, 5 annotated notes, exit 0. ✓

dudarenok-maker:
## PR review — pass 3 (head 21f000ff7e3b86228241f99379d1708d1ec13eb6, depth high)

Scope: the 10 changed files, re-derived from the code at this head rather than from the fix commit messages. Verified in a throwaway worktree at `21f000ff`: `node --test scripts/tests/check-register-row-citations.test.mjs` → 28/28 pass; `node scripts/check-register-row-citations.mjs` → OK, 37 citations / 348 files, 5 annotated notes; `gh pr checks` → all green. Then adversarially probed the two mechanisms commit `81cfd845` introduced (`normalizeLineContext` logical-line joining, and blockquote stripping inside `extractCitations`), plus re-ran pass 2's own mutations against the current test file.

|---|---|---|
| 1 | 🔴 CI-wiring gap (`docs/features/**` not an input to any step) | **VERIFIED RESOLVED** |
| 2 | 🔴 Non-mutation-proof tests | **VERIFIED RESOLVED** |
| 3 | 🟠 Annotation cross-contamination | **RESOLVED as stated, but the replacement re-opens it in a wider form — see 🔴 NEW-2** |
| 4 | 🟠 Hard-wrap / blockquote silent-miss | **RESOLVED as stated, but the fix introduced a line-attribution regression — see 🔴 NEW-1** |
| 5 | 🟡 `fileURLToPath` no-op test | **STILL OPEN** (see 🟡 below) |
| 6 | 🟡 Stale same-line prose | **PARTIALLY RESOLVED** — 4 surfaces fixed, 3 more still stale (see 🟡 below) |

**Item 1 — VERIFIED RESOLVED.** `scripts/verify-cache.mjs:177` adds `'docs/features/**'` to `test:hooks`' globs and `:213` adds `scripts/check-register-row-citations.mjs` to its `extraFiles`. I verified the cloud half independently rather than taking the commit message's word: `scripts/ci-scope.mjs:39-43` computes `scopes[slugFor(step.name)] = shared || stepTouchedByDiff(step, files)` directly off the same `STEPS[]` import, and `.github/workflows/verify.yml:195` gates the `test:hooks` leg on `fromJSON(needs.detect.outputs.scopes).step_test_hooks`. So one edit does close both halves. ✅

**Item 2 — VERIFIED RESOLVED.** I re-ran pass 2's exact mutations against the *current* test file (mutating `SCAN_PREFIXES` in the source, then running the suite):

| Mutation | Result |
|---|---|
| `SCAN_PREFIXES = []` | **3 tests fail** (real-tree scan, empty-`scannedFiles` mutation, `docs/features` mutation) |
| `SCAN_PREFIXES = ['docs/features/']` | **1 test fails** |
| `SCAN_PREFIXES = ['docs/testing/']` | **1 test fails** |

All three now go genuinely red. ✅

---

### 🔴 Blocking — `extractCitations` reports the wrong line number for 23 of the 47 citations in the real tree, and can flip a correctly-annotated citation into a hard failure

`scripts/check-register-row-citations.mjs:251-280`. Commit `81cfd845` made the regexes run over `normalizedText` (blockquote `> ` prefixes stripped, so every such line is 2 bytes shorter) — but `lineStarts` at `:259-265` is still built from the **original** `text`. `lineForOffset` therefore maps an index in one coordinate system through an offset table built in another, and the error accumulates by 2 bytes per preceding `> ` line.

**Live repro, no fixture needed** — run `node scripts/check-register-row-citations.mjs` at this head:

```
- docs/testing/onbox-sitting-plan.md:73 cites row A38 (annotated as discharged/removed).
```

The only `row A38` citation in that file is on line **74**, not 73:

```
73|> **Correction, 2026-08-23 (`check:register-citations` mechanical sweep).**
74|> `main`'s A38 discharge (register row A38 at the time, discharged, "ORT marker refuses —
```

This is a **regression from the pass-2 fix**. Running pass-2 head `8ba58d7a`'s `extractCitations` logic over the same file gives `74:A38` (correct); `21f000ff` gives `73:A38`.

**Scale, measured over the real tree** (diffing `8ba58d7a`'s attribution against this head's, for every `.md` under the two scan prefixes): **23 of 47 citations are mis-attributed.** Worst case `docs/testing/onbox-acceptance-register.md` — a citation on line **980** is reported at line **970**, off by 10 lines, because that file has 1,489 blockquote-prefixed lines ahead of it. Others: `ort-marker-onbox-acceptance.md` 252→249, `fs38-wave3-onbox-acceptance.md` 3069→3065, `cast-id-drift-onbox-acceptance.md` 661→659.

**It is not only a reporting defect.** The bad line number is passed straight into `isCitationAnnotated(id, lines, line)` at `:350`, which rebuilds its context from `lines[lineNum - 1]` of the *original* text. When the drift crosses a line boundary, the annotation check inspects the wrong physical line. Repro (fixture, via the injectable `readFile`):

```
> qqqq
> wwww

Row A9999 was discharged 2026-08-01.
```

→ `failures: [{line: 2, id: 'A9999'}]`, `annotated: []`. A properly-annotated citation is reported as a **hard failure**, at a line that contains neither the citation nor the annotation. That is exactly the false-failure class pass-2's Finding 4 was filed to close, re-introduced through a different door.

The same wrong line also feeds `rowIdAtLine` at `:348` (the register's own self-reference exemption), so a register self-reference can be mis-exempted or mis-failed the same way.

**Fix direction:** build `lineStarts` from `normalizedText`, not `text` — or, better, keep line lengths stable by replacing the stripped `> ` with two spaces (`line.replace(/^> /, '  ')`) so the two views stay byte-aligned and no index mapping is needed at all.

---

### 🔴 Blocking — the new `normalizeLineContext` re-opens annotation cross-contamination, wider than before, and now fails OPEN

`scripts/check-register-row-citations.mjs:102-136`, consumed at `:146-147`. `isProseContinuation` (`:116-120`) treats *any* two adjacent lines that are non-blank, non-`#`, non-`> ` as one logical line. That is not "hard-wrap" — it is every contiguous prose block, including separate bullets, separate table rows, and separate one-line paragraphs. The "nearest citation to the left" rule then reaches backwards across those joins.

**Repro** (`docs/testing/plan.md`, checked against a register that has no `A9999`):

```markdown
- Row A9999 is still owed.
- The old wave was discharged on 2026-08-01.
```

→ `failures: []`, `annotated: [{line: 1, id: 'A9999'}]`.

`A9999` does not exist in the register and **nothing on its own line annotates it**. The word "discharged" belongs to an entirely different bullet, about a different thing. The checker silently downgrades a dead citation to a note and exits 0. Control: insert a blank line between the two bullets and it correctly reports `failures: [{line: 1, id: 'A9999'}]` — the join is doing all the work.

Same shape with table rows:

```markdown
| row A9999 | still owed |
| row A8888 | discharged 2026-08-01 |
```

Here the ordering happens to save it, but reverse the two rows and `A9999` is exempted by `A8888`'s annotation.

This is strictly worse than what pass 2 flagged. Pass-2's `isCitationAnnotated(id, line)` was called with a **single physical line** and a ±50/±10-char window; the contamination surface was one line. This version's surface is the whole contiguous non-blank block, and `docs/testing/**` / `docs/features/**` are made almost entirely of long unbroken bulleted prose — the exact shape. And the failure direction is **fail-open**: the guard reports success while missing the drift it exists to catch.

I checked whether it is *live* today: it is not — all 5 current annotated notes have their annotation genuinely tied to their own citation (4 same-physical-line, 1 legitimate blockquote wrap). So this is latent, not a red-on-`main`. That does not make it safe: the guard's whole value is the diffs that have not been written yet.

**Fix direction:** join a continuation only when it really is one — e.g. require the *citation's own* line to have no sentence-terminating punctuation before the wrap, and refuse to join across a line that starts a new list item (`- `, `* `, `N. `) or a table row (`|`). Or drop forward-joining for prose entirely and keep it only for the blockquote case that motivated it (pass 2's Finding 4 repro was blockquote-wrapped).

---

### 🟡 Minor

- **Nested-blockquote wrapped citations are still a silent miss.** `extractCitations`' `line.replace(/^> /, '')` at `:255-257` strips exactly one `> `. Input `> > See register row\n> > A99 here.` yields **zero** citations — same silent-miss class as pass-2's Finding 4, one level of quoting deeper. (A *non*-wrapped `> > row A99` is fine.) Narrow, but it is the "guard that enumerates syntax" shape.

- **Item 5 is still open.** The rewritten `fileURLToPath handles percent-encoded paths correctly (Finding 3)` test (`scripts/tests/check-register-row-citations.test.mjs`) does now call `repoRoot()` — an improvement over asserting on Node's own import. But it still cannot fail for the reason it claims. I reverted `repoRoot()` at `:214-216` to the exact hand-rolled form the finding was about (`new URL('..', import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')`) and ran the test: **it passes.** The repo path `C:\Claude\Projects\Audiobook-Generator` contains no character that percent-encodes, so both implementations return the same string. The test's own comment asserts the opposite — *"would fail if the path were incorrectly decoded"* — which is a false claim about the test in front of it. Either construct a path that forces the encoding through (`new URL('../a b/', import.meta.url)`), or delete the percent-encoding claim from the comment and let the test stand honestly as "`repoRoot()` resolves to somewhere with a `package.json`".

- **Item 6 is only partially resolved — three more surfaces still describe the retired same-line rule**, all in `scripts/check-register-row-citations.mjs`:
  - `:88-91` — *"why **same-line is enough for v1**: measured against every real hit on the tree at the time this was added, the annotation always shares the citation's own **physical line**."* This is the retired rule stated verbatim, sitting directly above `ANNOTATION_REGEX`, and it contradicts the header block two dozen lines above it that item 6 *did* update.
  - `:307` — *"carries a **same-line** discharge/removal annotation"* in `checkRegisterRowCitations`' own JSDoc.
  - `:395` — **user-visible CLI output**: `'row but say so on the same line -- printed as a note, not a failure'`. This one is printed to an operator on every run and is factually wrong for the blockquote-wrap case the PR added on purpose.

- **The `docs/testing/onbox-sitting-plan.md` prose edit is no longer needed** and should probably be reverted. It rewords a factual correction note and produces two lines of 90 and 130 columns in a file that otherwise hard-wraps at ~80. I ran the **pre-PR** wording of that block through this head's checker: it comes back `annotated`, not `failures` — the "nearest citation to the left" rule plus blockquote joining already exempts it (`\bdischarged?\b` matches the later *"has since been discharged"*). The edit was needed under the ~50-char window that no longer exists.

### Reviewer-brief catalogue

Per `references/reviewer-brief.md`'s "Keeping the catalogue current" rule, 🔴 NEW-1 is a shape the catalogue does not name: **two derived views of the same input that must stay in lockstep, where a fix updated only one** — here, matching moved to a normalized string while the offset→line table stayed on the raw one. Appending it belongs to this round's fix work.

### Verdict

**Not clean — two blocking correctness bugs, both introduced by pass 2's own fixes.** Findings 1 and 2 are verified resolved; 3 and 4 are resolved as literally stated but their replacements carry NEW-1 and NEW-2; 5 is still open and 6 is partially open.

This is pass 3 of a 3-round cap, so the loop stops here and goes to the repo owner rather than auto-looping. Both blockers are small, localised, and have concrete repros above; neither needs a design pass.

dudarenok-maker:
## PR review — pass 4 (head 62fc8d5f1e0c0907da7319aad4368da05cdef463, depth high)

Pass 4 of a loop that has exceeded the formal 3-round cap, run with explicit owner approval after pass 3 found two correctness bugs that pass 2's own fix had introduced. Deltas only, per the re-review convention.

Everything below was re-derived from the code at this head, not from the commit message. Verification was done in a throwaway worktree at `62fc8d5f` (and a second one carrying a preview merge with current `origin/main`); the primary checkout was not touched.

---

### Pass-3 items — dispositions

**1. 🔴 → ✅ RESOLVED. Wrong line numbers in `extractCitations()`.**

Independently confirmed fixed, and confirmed as a real delta rather than a no-op commit.

`scripts/check-register-row-citations.mjs:279-323` now derives both the citation matches and the line-start index from `normalizedText`, so offsets and line-starts share one coordinate space. Because the per-line `replace(/^(> )+/, '')` can never add or remove a newline, `normalizedLines.length === lines.length` and a normalized line index is also the correct original line number — the fix is sound in principle, not just in the sampled cases.

Evidence, run against the real tree:

- I re-implemented pass 3's `extractCitations` verbatim and diffed it against this head's over the actual scanned corpus: **21 of 37 citations changed line number**, worst delta 10 lines. So the bug was live and the commit genuinely moves it.
- I then asserted every one of the 37 new line numbers against real file content: **0 mismatches**. The single non-exact case is `docs/testing/onbox-sitting-cloning-identity.md:296`, a genuine blockquote hard-wrap where `row` ends line 296 and `E1` begins 297 — reporting the line the citation *starts* on is correct.
- Synthetic reproduction of pass 3's exact shape (a citation on line 3 following two blockquote-prefixed lines) reports line 3; a heavier six-blockquote-line variant reports line 8. Both correct.

**2. 🔴 → ✅ RESOLVED (with a residual, see New finding 3). Cross-contamination via over-joining.**

Pass 3's two-bullet repro is fixed. `- alpha cites row A99` / `- beta cites row A1 which was discharged` now correctly reports A99 as a **failure**, not an annotation.

Two things I checked that the summary did not claim, because a tightened join can regress in the other direction:

- The legitimate hard-wrap case still works: a single sentence split across two lines with the annotation on line 2 is still joined and still annotated.
- The tightening did not regress the live corpus. The real tree still resolves all 5 pre-existing discharge-annotated citations as notes, exit 0.

**On the `/^\s*[-*]|\d+\.\s/` precedence bug you flagged before pushing — it is real, and I reproduced it.** Alternation binds loosest, so the pattern is `(^\s*[-*])|(\d+\.\s)` and the second branch is unanchored, matching a digit-run + `. ` anywhere on the line. Isolated repro, where the *only* difference between the two inputs is a date:

```
A: "The audit ran on 2026-08-01. It covered row A99 and the outcome\nwas that it had been discharged earlier."
   → A99 FAILS   (join suppressed: "01. " matches \d+\.\s)
B: "The audit ran last week and it covered row A99 and the outcome\nwas that it had been discharged earlier."
   → A99 annotated (join happens)
```

Your own read is correct on both counts: it is real, and it is the **safe** direction — it under-joins, producing a loud, self-describing false failure rather than a silent miss. There is **no live instance** in the current corpus (real tree and merged tree both exit 0). I am **not** treating this as blocking. Recorded as cleanup finding 2 below.

**3. 🟡 → ✅ RESOLVED. `fileURLToPath` regression test's false comment.**

`scripts/tests/check-register-row-citations.test.mjs:516-542`. The comment now states plainly that the test "does NOT specifically exercise percent-encoding," that this checkout's path has no special characters, and that "both fileURLToPath and a naive hand-rolled approach would both work here," then names what it *does* verify. No overclaim remains. (Minor quibble, not a finding: "impractical in this test environment" is a bit strong — `mkdtemp` with a space in the name would do it — but the comment no longer makes a false guarantee, which was the actual defect.)

**4. 🟡 → ✅ RESOLVED. Three surfaces still describing the retired same-line-only rule.**

All three now describe the logical-line rule:
- header comment, `check-register-row-citations.mjs:56-58` — "on the same **logical** line (including hard-wrapped continuations)";
- JSDoc, `:349-351` — "on its logical line (including hard-wrapped continuations)";
- **CLI output text**, `:436-437` — "(on the same logical line, including hard-wrapped continuations)". I confirmed this by reading the actual printed output of a live run, not just the source string.

I also grepped for surviving "same line" wording: the three remaining hits (`:59`, `:95`, `:170`) all describe the *cross-contamination problem* ("multiple citations appear on the same line"), which is still accurate. No stale surface left.

**5. 🟡 → ✅ RESOLVED. Nested-blockquote citations silently missed.**

Mutation-proven rather than asserted. For the input `> > row\n> > A99`:
- old single-`^> ` strip → **0** citations extracted (the silent miss);
- this head's `(> )+` strip → **1** citation, correctly reported as a failure.

Single-level `> row\n> A99` still works. `normalizeLineContext:115,150` uses the same `(> )+` strip on the annotation side, so the two halves agree.

**6. 🟡 → ✅ RESOLVED. `docs/testing/onbox-sitting-plan.md` prose edit reverted.**

`git diff origin/main -- docs/testing/onbox-sitting-plan.md` is **empty** — the file is byte-identical to `main`, so the revert is clean and the line wrap is intact. The original wording still resolves correctly against the fixed checker: the live run reports `docs/testing/onbox-sitting-plan.md:74 cites row A38 (annotated as discharged/removed)` — a note, not a failure. The annotation sits on line 75 and is picked up via the blockquote continuation join, which is exactly the path the fix rebuilt.

---

### Merge-safety check (not requested, but load-bearing)

The branch is **99 commits behind `main`**, and `main` has since modified the register itself plus two citation-bearing files (`onbox-sitting-cloning-identity.md`, `ort-marker-onbox-acceptance.md`). Since this checker's own test asserts "the real tree scan passes clean," a stale branch could go green here and red on `main` after merge.

I merged `origin/main` into a scratch worktree and re-ran everything. **All green:**

- `check:register-row-citations` → exit 0, 40 citations / 348 files, 7 annotated (the two new `ort-marker-onbox-acceptance.md` A28 citations from `main` resolve correctly as annotations — the fixed logic handling content its author never saw);
- the checker's own suite → **31/31 pass**;
- `check:onbox-register` → OK;
- `check:register-citations` (the sibling checker) → OK.

(`verify-cache.test.mjs` errors in my scratch worktree with `Cannot find module 'acorn'` — a missing-`node_modules` artifact of the throwaway tree, not a finding.)

---

### New findings

Both are in the release-notes pair, both introduced by this PR, and neither is caught by any existing gate.

**1. Correctness bug — Medium — `docs/release-notes-next.md:566`. The entry is appended *after* the `**Full changelog:**` footer, so it lands outside the document body and would ship malformed in the GitHub release.**

This file's own header comment (`:8-12`) documents the required structure as ending with the `**Full changelog:** vPREV...vNEW` footer. That footer is at line 564; the `---` closing the last section is at 563. This PR's bullet was appended at 566, *below* both.

`bump-version.mjs` feeds this file **verbatim** as the annotated-tag message → `release.yml` → the published release body (the header comment says so explicitly). So the v1.15.0 GitHub release notes would render a dangling technical bullet underneath the "Full changelog" line.

The correct home is the `### Test harness & suite hygiene` section (`:447`–`:562`), which is where the structurally identical sibling entries already live — the audit-gate entry (#2434, PR #2795) sits at `:562`, immediately above the `---`.

**2. Correctness bug — Medium — `RELEASE_NOTES.md:447`. The entry is appended at EOF, landing inside the `# Castwright 1.5.0` section instead of the in-progress `# Castwright 1.15.0` section.**

`RELEASE_NOTES.md` is ordered **newest-first**: `# Castwright 1.15.0` is line 1, and the file descends to `# Castwright 1.5.0` near the bottom. Appending at EOF therefore files this change under a release that shipped long ago. CLAUDE.md's before-shipping step 5 is explicit — "a matching user-facing, brand-voice line to the **in-progress version section at the top** of `RELEASE_NOTES.md`."

Secondary, same line: the text is in technical register, not brand voice. It names `check:register-row-citations`, `docs/testing/**`, "nearest citation to the left," and `#2603` — surrounded by entries like "Your book never leaves the house." It reads like the `release-notes-next.md` entry pasted into the user-facing file. Given this PR is a pure internal dev-tooling chore with no user-visible effect, step 5's own carve-out ("Skip only when the change has no shippable delta … say so explicitly") arguably applies and the honest fix is to **drop the `RELEASE_NOTES.md` entry entirely** and say so, keeping only the `release-notes-next.md` one. Either resolution is defensible; the current state — filed under 1.5.0, in the wrong voice — is not.

**Why no gate caught either:** `scripts/release-notes-gate.mjs 1.15.0` exits 0 with `OK — RELEASE_NOTES.md leads with 1.15.0`. It verifies only that the file *leads with* the right heading; it never checks that a PR's new entry landed *inside* that section, nor that `release-notes-next.md`'s footer is still last. That is reviewer-brief shape #1 (a guard that fails open on the thing it is nominally protecting) — worth a follow-up issue, but out of scope for this PR.

---

### Cleanup nits (non-blocking, no re-review owed)

**1. Dead code — `check-register-row-citations.mjs:292-297`.** `lineStartsOriginal` is built and pushed to but never read — the fix replaced its only consumer with `lineStartsNormalized`. Harmless (and `no-unused-vars` won't flag it, since `.push()` counts as a use), but it is exactly the orphan the fix created, so it falls under CLAUDE.md's "remove what YOUR changes orphaned."

**2. Imprecise heuristic — `:130-131`.** The `\d+\.\s` precedence bug described under item 2 above. Anchoring it — `/^\s*(?:[-*+]|\d+[.)]\s)/` — fixes this *and* nit 3 in one edit.

**3. Enumeration gap (fail-open) — `:130-131`.** The list-marker check enumerates only `-` and `*`, so two other valid markdown list markers still over-join, reopening the exact class item 2 was filed for. Reproduced, with the second line carrying an annotation and no citation of its own so nothing else masks the effect:

```
"+ alpha cites row A99\n+ beta was discharged last year"   → A99 wrongly ANNOTATED  (fail-open)
"1) alpha cites row A99\n2) beta was discharged last year" → A99 wrongly ANNOTATED  (fail-open)
"- alpha cites row A99\n- beta was discharged last year"   → A99 FAILS  (correct)
"1. alpha cites row A99\n2. beta was discharged last year" → A99 FAILS  (correct, but only by accident — via the unanchored \d+\.\s)
```

This is the brief's shape #2, "a guard that enumerates syntax." `+`-bulleted and `1)`-numbered lines do exist under the scanned prefixes, but **no live instance** currently sits adjacent to a discharge annotation — both the real tree and the merged tree are green. Low impact for a v1 docs-hygiene checker whose worst case is a missed warning; fine as a fast-follow.

**4. Fail-closed corner — `:203-217`.** An annotation positioned *before* its citation with no other citation to its left is never exempted (`"Discharged, and removed from the register: row A99."` → failure), because the nearest-citation-to-the-left scan finds nothing. Safe direction, no live instance, mentioned only for completeness.

---

### House gates

- Paired test: real, and mutation-proven where it matters — the item-5 nested-blockquote and item-1 line-number claims both fail against the pre-fix implementation, verified by running it. 31/31 green at head and post-merge.
- Release-notes pair: both files present, **both misplaced** — new findings 1 and 2.
- `Closes #2603`: present in the PR body, outside any code span. ✅
- On-box acceptance: not applicable — no hardware-provable behaviour. ✅
- Derived artifacts, config knobs, `cast.json` locking: not applicable. ✅
- CI wiring: reachable, not dead code — `check-register-row-citations.mjs` is in `test:hooks`'s `extraFiles` and `docs/features/**` was added to its globs, so a citation-breaking diff can't print `[cached]`. The `verify-cache.mjs` change is comments plus two data entries and merged cleanly against `main`'s much larger concurrent edits to that file. ✅
- Incidental findings declared in the PR body: ✅

---

### Verdict

**The two 🔴 blocking bugs from pass 3 are genuinely fixed** — independently re-derived, mutation-proven against the pre-fix code, and confirmed against both the current tree and a preview merge with `main`. All four 🟡 items are resolved. **No new correctness bug in the checker itself**, and none of the four residual code nits has a live instance in the corpus.

**The code is done. Fix the two release-notes placements and merge.** Both are single-block moves, verifiable by eye — they do not warrant a pass 5, and I would not re-review for them. The `release-notes-next.md` one is the one that actually matters, because it corrupts the published v1.15.0 release body and no gate will catch it before the tag is cut.

dudarenok-maker:
## PR review — follow-up (head 0cfb6cc26c8326ea86218fafe88ee9c16b0c306d)

Pass 4 gave an explicit merge verdict ("the code is done") but flagged two real defects in the release-notes pair, plus two low-risk nits worth folding in since they were already isolated during the pass-3 fix. This commit (`0cfb6cc2`) addresses all four — **no re-review round is owed for this one**, per pass 4's own call that these were "single-block moves, verifiable by eye."

- **`docs/release-notes-next.md`**: this PR's entry had landed *after* the `**Full changelog:**` footer, outside the document body — `bump-version.mjs` feeds this file verbatim into the published GitHub release, so v1.15.0 would have shipped a dangling bullet below the changelog line. Moved into the `### Test harness & suite hygiene` section, alongside the structurally identical sibling entries.
- **`RELEASE_NOTES.md`**: the entry had landed at EOF, under the `# Castwright 1.5.0` section (the file is newest-first) instead of the in-progress 1.15.0 section, and was written in technical register in an otherwise brand-voice file. Dropped entirely — per CLAUDE.md step 5's own carve-out, this PR is a pure internal chore with no user-visible delta.
- Anchored the list-marker regex in `isProseContinuation` (`/^\s*[-*]|\d+\.\s/` → `/^\s*(?:[-*+]|\d+[.)]\s)/`), closing both the unanchored-date false-trigger pass 4 reproduced and the `+`/`N)` enumeration gap it flagged as a residual nit — same fix closes both.
- Removed `lineStartsOriginal`, a dead variable orphaned by the pass-3 line-number fix.

Two new regression tests added. 33 tests pass, lint clean, `release-notes-gate.mjs 1.15.0` OK, real-tree output unchanged (37 citations, 5 annotated, 348 files, 0 failures).

**Status: ready for merge.**

Issue #2833, opened 2026-08-31
[agent instructions][claude][task] Widen wrongId to the discharge class


dudarenok-maker:
## Parent

#2721.

## What

Fix `buildLegitimateSubjectMap`, then widen `wrongId` in `scripts/check-register-citations.mjs` to be fatal for a fully-discharged subject, per the design decision in [this comment](https://github.com/dudarenok-maker/Castwright/issues/2721#issuecomment-5484703995).

## Background you need

- `scripts/check-register-citations.mjs` (~:26-37) — the `unknownSubject` bucket is currently non-fatal and opt-in behind `--strict`. This is the check that would need to become fatal once the map is fixed (or stay as-is — read the actual code to see if this specific bucket is the right place, or if `wrongId` is a distinct check elsewhere in the same file; the issue names TWO separate line ranges, ~:26-37 and ~:209-210 — read both before editing).
- `buildLegitimateSubjectMap` maps one subject to a SET of IDs (a subject can span several register rows). Two current `unknownSubject` residuals on the live corpus are benign — find them by running the checker with `--strict` (or whatever flag surfaces the bucket) against the current `docs/testing/onbox-acceptance-register.md` and seeing what it flags. Understand WHY each is benign before changing the map — the issue implies it's a map-construction gap (the map isn't capturing all of a multi-row subject's IDs), not a real citation defect.
- #2717 (already merged) shipped stable, allocate-once register row IDs — this is what makes a discharged ID's staying dangling-forever a reliable signal now, per the decision comment.
- `docs/testing/onbox-acceptance-register.md`'s allocation-procedure blockquote — describes what does/doesn't fire; needs updating once behavior changes.

## The fix

1. Read `scripts/check-register-citations.mjs` fully — find `buildLegitimateSubjectMap`, the `unknownSubject` bucket, and the `wrongId` check (~:209-210) to understand the actual current relationship between them before changing anything.
2. Fix `buildLegitimateSubjectMap` so a subject's ID-set is complete: run the checker's `--strict` mode against the real corpus first to see what it currently reports as `unknownSubject`, diagnose why (likely: some legitimate ID belonging to a subject with multiple rows isn't being added to that subject's set), and fix the map construction so those two residuals resolve.
[...]

Issue #2827, opened 2026-08-31
[agent instructions][claude][task] Drop the on-box register changelog


dudarenok-maker:
## Parent

#2708.

## What

Remove the changelog from both surfaces of the on-box acceptance register, per the design decision recorded in [this comment](https://github.com/dudarenok-maker/Castwright/issues/2708#issuecomment-5484695277): the register tracks current state, not history; the changelog has gone silently stale repeatedly with no mechanical guard, and building one would duplicate #2603's work. Drop it and point readers at git/PR history instead.

## Background you need

- `docs/testing/onbox-acceptance-register-live-view.html` — a hand-authored styled page (NOT rendered from the `.md`). It currently carries a stack of changelog callout blocks recording each wave's row-count delta (search for "Historical, predates wave" and similar callout framing).
- `docs/testing/onbox-acceptance-register.md` — carries a parallel changelog section (11+ entries as of 2026-09-01). Find it by searching for a heading like "## Changelog" or similar near the top or bottom of the file.
- The live-view footer has a pointer back to the `.md`'s changelog ("see the .md's own changelog") — this must be removed too, not left dangling once the `.md` section is gone.
- `scripts/check-onbox-register.mjs` cross-checks the register's arithmetic (glance-table counts vs. row headings, stated total vs. glance table). It does NOT check changelog content — confirm your removal doesn't touch anything this script asserts on (it shouldn't; the changelog is prose, not part of the counted structure).

## The fix

1. In `docs/testing/onbox-acceptance-register-live-view.html`: delete every changelog callout block. Delete the footer's "see the .md's own changelog" line/pointer.
2. In `docs/testing/onbox-acceptance-register.md`: delete the changelog section entirely.
3. Grep both files (and `docs/testing/onbox-acceptance-register.md`'s own procedure text, if it references the changelog anywhere else, e.g. in the "Live view" publish procedure) for any remaining reference to "changelog" and remove/rewrite those too.
4. Do NOT touch the summary strip, glance table, or any row content — only changelog material.
5. Run `npm run check:onbox-register` and confirm it still passes (this change touches prose only, not the counted structure it validates).
6. Republish the live view per the register's own "Live view" procedure (read that section of the `.md` for the exact steps) so the live URL reflects the change — or note in your commit/PR why you could not (e.g. no network access) and leave that as a stated follow-up in the PR body, not silently skipped.

## Where the work happens

[...]

Issue #2169, opened 2026-08-06
ops — createAnnotatedTag silently honours an inherited GIT_DIR over its own repoRoot argument


dudarenok-maker:
## What

`createAnnotatedTag` (`scripts/bump-version.mjs:390-398`) runs `git tag` with an explicit `cwd: repoRoot`, but inherits the ambient environment:

```js
export function createAnnotatedTag({ repoRoot, newTag, notesFile }) {
  const tagMessage = buildTagMessage(readFileSync(resolve(notesFile), 'utf8'));
  execFileSync('git', ['tag', '--cleanup=verbatim', '-a', newTag, '-F', '-'], {
    cwd: repoRoot,
    input: tagMessage,
    stdio: ['pipe', 'inherit', 'inherit'],
    encoding: 'utf8',
  });
}
```

Git resolves the repository from `$GIT_DIR` / `$GIT_WORK_TREE` **before** it falls back to discovery from `cwd`. So whenever either is set in the environment, the `repoRoot` argument is silently ignored and the tag is created in whatever repository the environment names. `repoRoot` is accepted, used for nothing, and the command still exits 0.

The same shape is in the shared `git()` helper (`scripts/bump-version.mjs:196-203`) — i.e. every other git call in the release cut — and in `scripts/release-body.mjs:70`'s `git tag -l --format=%(contents)`, which is how `release.yml` reads the annotation back out.

## Why it matters

`GIT_DIR` is not an exotic thing to have set: git exports it for every hook process, so anything that reaches `bump-version.mjs` from inside a hook — or from a shell where an earlier command exported it — is affected. Worktrees make it likelier still, since `GIT_DIR` and `GIT_WORK_TREE` then point at different places.

The release path is the worst place for this class of defect. The tag is what `release.yml` publishes from, the failure is silent rather than loud, and the recovery is a force-deleted public tag.

## Acceptance

- `bump-version.mjs`'s git invocations resolve the repository from their own `repoRoot`, not from an inherited `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` / `GIT_OBJECT_DIRECTORY` / `GIT_COMMON_DIR`.
- The scrub lives in the **shared helper**, so a git call added to this script later inherits the fix instead of having to remember it. `createAnnotatedTag` builds its own `execFileSync` options today — it must route through the same scrub rather than growing a second copy.
[...]

PR #2844, opened 2026-09-01
fix(scripts): bump-version.test.mjs cleanGitEnv is case-sensitive

### LINKED ISSUE #2841: fix(scripts): bump-version.test.mjs cleanGitEnv is case-sensitive, drops lowercase git_dir survivor
## What

`scripts/tests/bump-version.test.mjs`'s local `cleanGitEnv()` helper (its own throwaway-repo git-env scrub, distinct from the shared `scrubGitEnv()` in `scripts/git-env.mjs`) filters with a case-sensitive `key.startsWith('GIT_')`. Windows preserves whatever casing a git hook stores its env vars under while lookup stays case-insensitive — the exact trap `scrubGitEnv()` was hardened against (see its own header, and #2175's regression test for it). A lowercase-cased `git_dir`/`git_index_file` survives this filter and still redirects the test's throwaway `git commit`/`git tag` calls into the real repo's hook context.

## Impact — how this was found

Discovered live: two separate Open Engine agent runs (#2833, #2827) each burned their entire retry budget on `git commit` failing under `.husky/pre-commit` → `test:hooks` → `bump-version.test.mjs`, both from **fresh worktrees'** first commit. #2827's run diagnosed (correctly, on symptom; incorrectly, on the specific variable) that a polluted git-hook env was leaking into the test's nested git spawns, but concluded it couldn't safely fix shared test infra from a docs-only ticket. Running `npm run test:hooks` standalone (not via git commit) passed clean 1735/1735 every time — only reproducible when actually driven by a real `git commit`'s hook, which is exactly when the leaked env exists to trigger the case-insensitivity gap.

## The fix

`git.toUpperCase().startsWith('GIT_')` instead of the bare case-sensitive check, plus a mutation-provable regression test (mirrors the existing `scrubGitEnv` case-insensitivity test at the same file, added by #2175). Confirmed: new test goes RED against the pre-fix code, GREEN after.

## Not in scope

Migrating `cleanGitEnv()` to call the shared `scrubGitEnv()` — it deliberately also strips `GIT_INDEX_FILE`/`GIT_PREFIX`, which `scrubGitEnv()` intentionally preserves (see its #2216 note) for a different caller class. This local helper's broader stripping is correct for a throwaway-repo test; only the casing bug needed fixing.

dudarenok-maker:
## Summary
- `cleanGitEnv()` in `scripts/tests/bump-version.test.mjs` filtered git env vars with a case-sensitive `startsWith('GIT_')`, missing a lowercase `git_dir`/`git_index_file` survivor Windows can preserve — the same trap `scrubGitEnv()` (`scripts/git-env.mjs`) was already hardened against.
- This made the throwaway-repo test's own `git commit`/`git tag` calls vulnerable to redirection into the real repo's hook context — reproducible only when `test:hooks` runs inside a real `git commit`'s pre-commit hook, never standalone.
- Discovered live: two Open Engine agent runs (#2833, #2827) each burned their full retry budget on this exact failure from a fresh worktree's first commit.

## Fix
`key.toUpperCase().startsWith('GIT_')`, plus a mutation-provable regression test mirroring the existing `scrubGitEnv` casing test in the same file. Confirmed RED against the pre-fix code, GREEN after.

## Also fixed, found in passing (pr-review-gate pass 1, medium depth)

- **`release-body.test.mjs`'s `cleanGitEnv()` was a byte-identical copy** of `bump-version.test.mjs`'s, carrying the same case-sensitive gap. No mechanical guard catches either copy — `git-scrub.test.mjs`'s repo-wide `scrubGitEnv` guard explicitly excludes `scripts/tests/**` from its scan. Fixed identically, with a mirrored, mutation-provable regression test.
- Two adjacent save/restore blocks in `bump-version.test.mjs` (the `#2169` GIT_DIR-decoy test, and the direct-BOM test) and one in `release-body.test.mjs` (its own GIT_DIR-decoy test) had the same latent case-sensitive filter on which ambient `GIT_*` vars to snapshot/restore. Fixed alongside rather than left armed.
- Fixed a stale "see the #2216 note below" comment (both notes are above).
- Replaced this PR's own new test's vacuous `assert.ok(cleaned.PATH ?? cleaned.Path)` (always true; the preceding assignment line was a no-op) with a real spot-check that `PATH` survives the scrub with its value intact.

## Also fixed, found in passing (pr-review-gate pass 2, medium depth)

- **`verify-cache.test.mjs`'s `gitAt()` scrubbed git-hook env vars via a hardcoded-uppercase object destructure** (`{ GIT_DIR: _GIT_DIR, ... } = process.env`) — the same case-sensitivity gap already fixed twice in this PR, spelled differently enough to survive a grep for `startsWith('GIT_')`. **More severe than the earlier two instances**: demonstrated live that an ambient lowercase `git_dir` flips a real checkout's `core.bare` flag from `false` to `true` via `makeGitFixture()`'s `git init`, and its two call sites drive real git writes, not just reads. Replaced with the same `cleanGitEnv()` helper pattern used in the other two files, plus a mirrored mutation-provable regression test.
- Reverted a comment wrongly changed in round 1: `bump-version.test.mjs`'s "`#2175` review, Finding 1" cross-reference genuinely is below that comment, not above — round 1 flipped it without checking; round 2's reviewer caught round 1 breaking what round 1 itself had validated as correct.
- Completed `release-body.test.mjs`'s round-1 regression test to actually mirror its sibling: it was missing the "non-`GIT_` keys survive the scrub" half of the assertion that the sibling test in `bump-version.test.mjs` has.

## Also fixed, found in passing (pr-review-gate pass 3, medium depth — the cap, no round 4)

- `verify-cache.test.mjs`'s new round-2 regression test had the same "partial mirror" defect round 2 fixed in `release-body.test.mjs`: it claimed to mirror its siblings but was missing the "non-`GIT_` keys survive the scrub" half of the assertion. A `cleanGitEnv() { return {}; }` mutant left it green; fixed and re-confirmed mutation-sensitive.
- `scripts/verify-cache.mjs`'s `gitEnv()` stripped `GIT_PREFIX` with a case-sensitive destructure on top of an already case-insensitive `scrubGitEnv()` call. Confirmed inert (no git command here is affected by `GIT_PREFIX`'s casing) — fixed anyway for consistency with the rest of this PR.
- **One item was NOT fixed here and is filed as #2855 instead**: this bug class has now been hand-fixed three separate times in three files, with nothing mechanical stopping a fourth copy (the guard excluding `scripts/tests/**` from its scan is deliberate and pre-existing). Closing that gap needs a design decision — shared helper vs. narrowed guard vs. deliberate duplication — named in #2855 rather than picked blind at the review cap.

**Round 3 verdict: no correctness bug remains. The re-review trigger is not met; this closes the review loop at the cap.**

No shippable/user-visible delta from any of the above — internal test-infra reliability fix. Release-notes update intentionally skipped per CLAUDE.md step 5's no-shippable-delta exemption.

## Test plan
- [x] `node --test scripts/tests/bump-version.test.mjs` — 59/59 passing
- [x] Mutation check: reverted the fix, new test failed with the expected assertion; restored, test passes
- [x] `node --test scripts/tests/bump-version.test.mjs scripts/tests/release-body.test.mjs` — 71/71 passing (both files, after round-1 fixes)
- [x] Mutation check on `release-body.test.mjs`'s new test: reverted its fix, test failed with the expected assertion; restored, test passes
- [x] `node --test scripts/tests/bump-version.test.mjs scripts/tests/release-body.test.mjs scripts/tests/verify-cache.test.mjs` — 198/198 passing (all three files, after round-2 fixes)
- [x] Mutation check on `verify-cache.test.mjs`'s new test: reverted its fix, test failed with the expected assertion; restored, test passes
- [x] 198/198 passing again after round-3 fixes; mutation-checked the completed `verify-cache.test.mjs` mirror (`return {}` mutant now fails on the PATH assertion, as intended)
- [x] `npm run verify:fast:branch` — lint pass, test:hooks/test:server clean, rest correctly out-of-scope. (test:server hit two separate, unrelated, contention-driven flakes across retries — `analysis.fresh-cast-lock.test.ts` and `voice-style.test.ts`, both timing-sensitive concurrency tests, both confirmed passing clean in isolation and both irrelevant to this PR's `scripts/**`-only diff. Multiple other sessions were running concurrent test batteries on the same box throughout this PR.)

Closes #2841

### DIFF
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -42,10 +42,15 @@ const MOJIBAKE_EM_DASH = 'â€”'; // should decode to U+2014 —, standalone 
 // which child processes inherit. Without sanitising, the test's `git commit`
 // would write into the PARENT worktree's git instead of the temp fixture —
 // silently creating bogus commits on whatever branch invoked the hook.
+// Case-insensitive match: Windows preserves whatever casing a var was stored
+// under while lookup is case-insensitive, and git honours a lowercase
+// `git_dir` identically to `GIT_DIR` — see git-env.mjs's scrubGitEnv() header
+// for the same trap. A `startsWith('GIT_')` filter alone leaves a `git_dir`
+// (or similar) survivor that still redirects the throwaway repo's commits.
 function cleanGitEnv() {
   const env = { ...process.env };
   for (const key of Object.keys(env)) {
-    if (key.startsWith('GIT_')) delete env[key];
+    if (key.toUpperCase().startsWith('GIT_')) delete env[key];
   }
   return env;
 }
@@ -627,6 +632,33 @@ test('scrubGitEnv strips every git repository-discovery override, preserves GIT_
   assert.equal(scrubbed.GIT_AUTHOR_NAME, 'Someone');
   // Deliberately preserved — see the #2216 correction above.
   assert.equal(scrubbed.GIT_INDEX_FILE, '/decoy/.git/index');
+});
+
+// Case-insensitivity regression for THIS file's own cleanGitEnv(), the same
+// trap scrubGitEnv() was hardened against below (#2175 review, Finding 1).
+// cleanGitEnv() hand-rolls its own filter rather than delegating to
+// scrubGitEnv() (it must also strip GIT_INDEX_FILE, which scrubGitEnv()
+// deliberately preserves — see the #2216 note below), so it needs its own
+// case-insensitive regression. Deliberately mixed-case-only: cannot pass
+// against a `key.startsWith('GIT_')` filter.
+test('cleanGitEnv strips a git env override regardless of stored casing', () => {
+  const saved = { ...process.env };
+  try {
+    for (const key of Object.keys(process.env)) {
+      if (key.toUpperCase().startsWith('GIT_')) delete process.env[key];
+    }
+    process.env.git_dir = '/decoy/.git';
+    process.env.Git_Index_File = '/decoy/.git/index';
+    process.env.PATH = saved.PATH ?? saved.Path;
+
+    const cleaned = cleanGitEnv();
+    assert.equal(cleaned.git_dir, undefined);
+    assert.equal(cleaned.Git_Index_File, undefined);
+    assert.ok(cleaned.PATH ?? cleaned.Path);
+  } finally {
+    for (const key of Object.keys(process.env)) delete process.env[key];
+    Object.assign(process.env, saved);
+  }
 });
 
 // #2175 review, Finding 1 — Windows env lookup is case-insensitive, but
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -635,10 +635,10 @@ test('scrubGitEnv strips every git repository-discovery override, preserves GIT_
 });
 
 // Case-insensitivity regression for THIS file's own cleanGitEnv(), the same
-// trap scrubGitEnv() was hardened against below (#2175 review, Finding 1).
+// trap scrubGitEnv() was hardened against above (#2175 review, Finding 1).
 // cleanGitEnv() hand-rolls its own filter rather than delegating to
 // scrubGitEnv() (it must also strip GIT_INDEX_FILE, which scrubGitEnv()
-// deliberately preserves — see the #2216 note below), so it needs its own
+// deliberately preserves — see the #2216 note above), so it needs its own
 // case-insensitive regression. Deliberately mixed-case-only: cannot pass
 // against a `key.startsWith('GIT_')` filter.
 test('cleanGitEnv strips a git env override regardless of stored casing', () => {
@@ -649,12 +649,15 @@ test('cleanGitEnv strips a git env override regardless of stored casing', () => 
     }
     process.env.git_dir = '/decoy/.git';
     process.env.Git_Index_File = '/decoy/.git/index';
-    process.env.PATH = saved.PATH ?? saved.Path;
 
     const cleaned = cleanGitEnv();
     assert.equal(cleaned.git_dir, undefined);
     assert.equal(cleaned.Git_Index_File, undefined);
-    assert.ok(cleaned.PATH ?? cleaned.Path);
+    // Non-GIT_ keys are untouched — spot-check against whichever casing this
+    // OS actually stores the PATH var under (Windows: "Path").
+    const pathKey = Object.keys(cleaned).find((k) => k.toUpperCase() === 'PATH');
+    assert.ok(pathKey, 'PATH must survive the scrub');
+    assert.equal(cleaned[pathKey], saved[pathKey]);
   } finally {
     for (const key of Object.keys(process.env)) delete process.env[key];
     Object.assign(process.env, saved);
@@ -745,7 +748,7 @@ test('createAnnotatedTag resolves repoRoot even with an inherited GIT_DIR pointi
   writeFileSync(notes, '# v9.9.8\n\nFixes:\n- the GIT_DIR leak\n');
   const savedGitEnv = {};
   for (const key of Object.keys(process.env)) {
-    if (key.startsWith('GIT_')) {
+    if (key.toUpperCase().startsWith('GIT_')) {
       savedGitEnv[key] = process.env[key];
       delete process.env[key];
     }
@@ -1198,7 +1201,7 @@ test('createAnnotatedTag strips a BOM even when called directly, bypassing pre-f
     writeBOMFile(notes, '# v9.9.9\n\nFixes:\n- the bug\n');
     const savedGitEnv = {};
     for (const key of Object.keys(process.env)) {
-      if (key.startsWith('GIT_')) {
+      if (key.toUpperCase().startsWith('GIT_')) {
         savedGitEnv[key] = process.env[key];
         delete process.env[key];
       }
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -635,7 +635,7 @@ test('scrubGitEnv strips every git repository-discovery override, preserves GIT_
 });
 
 // Case-insensitivity regression for THIS file's own cleanGitEnv(), the same
-// trap scrubGitEnv() was hardened against above (#2175 review, Finding 1).
+// trap scrubGitEnv() was hardened against below (#2175 review, Finding 1).
 // cleanGitEnv() hand-rolls its own filter rather than delegating to
 // scrubGitEnv() (it must also strip GIT_INDEX_FILE, which scrubGitEnv()
 // deliberately preserves — see the #2216 note above), so it needs its own
--- a/scripts/tests/release-body.test.mjs
+++ b/scripts/tests/release-body.test.mjs
@@ -58,10 +58,15 @@ const CONFLICT_FIXTURE =
 // this for real: a stray tag landed in this worktree). Every git
 // invocation below, and every spawned `node release-body.mjs` process,
 // passes this explicit env instead of inheriting process.env.
+// Case-insensitive match: Windows preserves whatever casing a var was stored
+// under while lookup is case-insensitive, and git honours a lowercase
+// `git_dir` identically to `GIT_DIR` — see git-env.mjs's scrubGitEnv() header,
+// and bump-version.test.mjs's identical helper/fix (#2841), for the measured
+// evidence. A `startsWith('GIT_')` filter alone leaves a survivor.
 function cleanGitEnv() {
   const env = { ...process.env };
   for (const key of Object.keys(env)) {
-    if (key.startsWith('GIT_')) delete env[key];
+    if (key.toUpperCase().startsWith('GIT_')) delete env[key];
   }
   return env;
 }
@@ -374,7 +379,7 @@ test('readTagAnnotation resolves repoRoot even with an inherited GIT_DIR pointin
 
   const savedGitEnv = {};
   for (const key of Object.keys(process.env)) {
-    if (key.startsWith('GIT_')) {
+    if (key.toUpperCase().startsWith('GIT_')) {
       savedGitEnv[key] = process.env[key];
       delete process.env[key];
     }
@@ -389,5 +394,28 @@ test('readTagAnnotation resolves repoRoot even with an inherited GIT_DIR pointin
     for (const [k, v] of Object.entries(savedGitEnv)) process.env[k] = v;
     rmSync(dir, { recursive: true, force: true });
     rmSync(decoy, { recursive: true, force: true });
+  }
+});
+
+// #2841 — this file's cleanGitEnv() is a byte-identical copy of
+// bump-version.test.mjs's, which had the same case-sensitivity gap fixed
+// there. Mirrors that file's regression test directly against the local
+// helper. Deliberately mixed-case-only: cannot pass against a
+// `startsWith('GIT_')` filter.
+test('cleanGitEnv strips a git env override regardless of stored casing', () => {
+  const saved = { ...process.env };
+  try {
+    for (const key of Object.keys(process.env)) {
+      if (key.toUpperCase().startsWith('GIT_')) delete process.env[key];
+    }
+    process.env.git_dir = '/decoy/.git';
+    process.env.Git_Index_File = '/decoy/.git/index';
+
+    const cleaned = cleanGitEnv();
+    assert.equal(cleaned.git_dir, undefined);
+    assert.equal(cleaned.Git_Index_File, undefined);
+  } finally {
+    for (const key of Object.keys(process.env)) delete process.env[key];
+    Object.assign(process.env, saved);
   }
 });
--- a/scripts/tests/release-body.test.mjs
+++ b/scripts/tests/release-body.test.mjs
@@ -414,6 +414,11 @@ test('cleanGitEnv strips a git env override regardless of stored casing', () => 
     const cleaned = cleanGitEnv();
     assert.equal(cleaned.git_dir, undefined);
     assert.equal(cleaned.Git_Index_File, undefined);
+    // Non-GIT_ keys are untouched — spot-check against whichever casing this
+    // OS actually stores the PATH var under (Windows: "Path").
+    const pathKey = Object.keys(cleaned).find((k) => k.toUpperCase() === 'PATH');
+    assert.ok(pathKey, 'PATH must survive the scrub');
+    assert.equal(cleaned[pathKey], saved[pathKey]);
   } finally {
     for (const key of Object.keys(process.env)) delete process.env[key];
     Object.assign(process.env, saved);
--- a/scripts/tests/verify-cache.test.mjs
+++ b/scripts/tests/verify-cache.test.mjs
@@ -1461,6 +1461,11 @@ test('cleanGitEnv strips a git env override regardless of stored casing', () => 
     const cleaned = cleanGitEnv();
     assert.equal(cleaned.git_dir, undefined);
     assert.equal(cleaned.Git_Index_File, undefined);
+    // Non-GIT_ keys are untouched — spot-check against whichever casing this
+    // OS actually stores the PATH var under (Windows: "Path").
+    const pathKey = Object.keys(cleaned).find((k) => k.toUpperCase() === 'PATH');
+    assert.ok(pathKey, 'PATH must survive the scrub');
+    assert.equal(cleaned[pathKey], saved[pathKey]);
   } finally {
     for (const key of Object.keys(process.env)) delete process.env[key];
     Object.assign(process.env, saved);
--- a/scripts/tests/verify-cache.test.mjs
+++ b/scripts/tests/verify-cache.test.mjs
@@ -1399,20 +1399,29 @@ test('a --changed-only pass is never written to the verify-cache — only a full
 
 // --- Branch-diff scope filter (verify/CI rebalance, 2026-07-06) --------
 
+// Strip ambient GIT_DIR/GIT_WORK_TREE/GIT_INDEX_FILE/GIT_PREFIX (and anything
+// else GIT_-prefixed) before spawning: when this suite runs inside a real git
+// hook (pre-commit calls `npm run verify:fast:scoped` -> `npm run
+// test:hooks`), git sets these in the process env for the hook, and without
+// stripping them these fixture commands would operate on the REAL repo
+// instead of the throwaway `cwd`. Case-insensitive match: Windows preserves
+// whatever casing a var was stored under while lookup is case-insensitive,
+// and git honours a lowercase `git_dir` identically to `GIT_DIR` — a
+// destructure keyed on the canonical uppercase names alone (the pre-#2841
+// shape here) leaves a differently-cased survivor that still redirects these
+// fixture commands at the real repo. Demonstrated live: an ambient lowercase
+// `git_dir` flipped a real checkout's `core.bare` from false to true via
+// `makeGitFixture()`'s `git init` below.
+function cleanGitEnv() {
+  const env = { ...process.env };
+  for (const key of Object.keys(env)) {
+    if (key.toUpperCase().startsWith('GIT_')) delete env[key];
+  }
+  return env;
+}
+
 function gitAt(cwd, args) {
-  // Strip ambient GIT_DIR/GIT_WORK_TREE/GIT_INDEX_FILE/GIT_PREFIX before
-  // spawning: when this suite runs inside a real git hook (pre-commit calls
-  // `npm run verify:fast:scoped` -> `npm run test:hooks`), git sets these in
-  // the process env for the hook, and without stripping them these fixture
-  // commands would operate on the REAL repo instead of the throwaway `cwd`.
-  const {
-    GIT_DIR: _GIT_DIR,
-    GIT_WORK_TREE: _GIT_WORK_TREE,
-    GIT_INDEX_FILE: _GIT_INDEX_FILE,
-    GIT_PREFIX: _GIT_PREFIX,
-    ...cleanEnv
-  } = process.env;
-  const r = spawnSync('git', args, { cwd, encoding: 'utf8', env: cleanEnv });
+  const r = spawnSync('git', args, { cwd, encoding: 'utf8', env: cleanGitEnv() });
   if (r.status !== 0) {
     throw new Error(`git ${args.join(' ')} failed: ${r.stderr}`);
   }
@@ -1432,6 +1441,31 @@ function makeGitFixture() {
   gitAt(dir, ['commit', '-q', '-m', 'base']);
   return dir;
 }
+
+// #2841 review round 2 — this file's own git-env scrub was a hardcoded-
+// uppercase object destructure, the same case-sensitivity gap already fixed
+// in bump-version.test.mjs and release-body.test.mjs, but spelled
+// differently enough to survive a grep for `startsWith('GIT_')`. Mirrors
+// those files' regression test directly against the local helper.
+// Deliberately mixed-case-only: cannot pass against a destructure keyed on
+// canonical uppercase names alone.
+test('cleanGitEnv strips a git env override regardless of stored casing', () => {
+  const saved = { ...process.env };
+  try {
+    for (const key of Object.keys(process.env)) {
+      if (key.toUpperCase().startsWith('GIT_')) delete process.env[key];
+    }
+    process.env.git_dir = '/decoy/.git';
+    process.env.Git_Index_File = '/decoy/.git/index';
+
+    const cleaned = cleanGitEnv();
+    assert.equal(cleaned.git_dir, undefined);
+    assert.equal(cleaned.Git_Index_File, undefined);
+  } finally {
+    for (const key of Object.keys(process.env)) delete process.env[key];
+    Object.assign(process.env, saved);
+  }
+});
 
 test('branchDiffFiles: returns files changed since branching off main', () => {
   const dir = makeGitFixture();
@@ -1581,17 +1615,10 @@ test('stagedDiffFiles: honours an ambient GIT_INDEX_FILE pointing at a temporary
   // a `git commit -a` / `git commit -- <path>` invocation, where the temp
   // index reflects the files that invocation is actually committing.
   writeFileSync(join(repoDir, 'via-temp-index.txt'), 'temp', 'utf8');
-  const {
-    GIT_DIR: _GIT_DIR,
-    GIT_WORK_TREE: _GIT_WORK_TREE,
-    GIT_INDEX_FILE: _GIT_INDEX_FILE,
-    GIT_PREFIX: _GIT_PREFIX,
-    ...cleanEnv
-  } = process.env;
   const populateTempIndex = spawnSync('git', ['add', 'via-temp-index.txt'], {
     cwd: repoDir,
     encoding: 'utf8',
-    env: { ...cleanEnv, GIT_INDEX_FILE: tempIndexPath },
+    env: { ...cleanGitEnv(), GIT_INDEX_FILE: tempIndexPath },
   });
   assert.equal(populateTempIndex.status, 0, populateTempIndex.stderr);
 
--- a/scripts/verify-cache.mjs
+++ b/scripts/verify-cache.mjs
@@ -957,8 +957,11 @@ export function stagedDiffFiles(cwd) {
 // `null`) would NOT trip verify-cache's "diff failed, run everything" safety
 // branch. See scripts/tests/verify-cache.test.mjs for the behavioural proof.
 function gitEnv() {
-  const { GIT_PREFIX: _GIT_PREFIX, ...cleanEnv } = scrubGitEnv();
-  return cleanEnv;
+  const env = scrubGitEnv();
+  for (const key of Object.keys(env)) {
+    if (key.toUpperCase() === 'GIT_PREFIX') delete env[key];
+  }
+  return env;
 }
 
 export function branchDiffFiles(cwd) {


dudarenok-maker:
## PR review — pass 1 (head 847e6f6c, depth medium)

Scope: `scripts/tests/bump-version.test.mjs` (the only changed file), read against `scripts/git-env.mjs`, `scripts/tests/release-body.test.mjs` and `scripts/tests/git-scrub.test.mjs` · verified: built an out-of-tree harness from the head SHA and actually ran it — new test GREEN with the fix, RED at the asserted line with the one production line reverted, full file 59/59, and 59/59 again under a hostile ambient lowercase `git_dir` — before adversarially probing the Windows env-casing semantics the fix depends on, the `finally` restore's faithfulness, and the sibling copies of the same helper.

### 🟠 Significant — correctness bug: the identical defect is still armed in `scripts/tests/release-body.test.mjs`, and no guard covers it

`scripts/tests/release-body.test.mjs:61-67` carries a byte-identical `cleanGitEnv()` with the unfixed case-sensitive filter:

```js
function cleanGitEnv() {
  const env = { ...process.env };
  for (const key of Object.keys(env)) {
    if (key.startsWith('GIT_')) delete env[key];   // <- line 64, still case-sensitive
  }
  return env;
}
```

It feeds `gitExec()` at `release-body.test.mjs:69-71`, and its own header states the stakes: *"Every git invocation below, and every spawned `node release-body.mjs` process, passes this explicit env"* … *"bump-version.test.mjs hit this for real: a stray tag landed in this worktree."* Same helper, same job, same blast radius as the one this PR just fixed.

Demonstrated, not inferred. Copying that function verbatim and driving one throwaway-repo write through it:

```
stored env keys matching GIT_: ["GIT_ASKPASS","git_dir","GIT_EDITOR"]
cleanGitEnv() output still carries git_dir = C:\...\Temp\repro-VICTIM-g7yicx\.git
tags in THROWAWAY repo : ""
tags in VICTIM   repo : "release-body-test-tag"
RESULT: LEAK — the write landed in the VICTIM repo, not the throwaway.
```

Exact input → wrong output: ambient `git_dir` (lowercase) set → `gitExec(['tag', 'release-body-test-tag'], { cwd: throwaway })` → the tag is created in the **victim** repo and the throwaway gets nothing. That is precisely the failure this PR's own body describes.

What makes this a finding rather than a note: **there is no mechanical backstop.** `scripts/tests/git-scrub.test.mjs:990` asserts `'scripts/tests/** must be excluded from the scan'`, so the repo-wide `scrubGitEnv` guard never sees either hand-rolled copy. Fixing one site by hand and leaving the other is the "one instance fixed, the class left armed" shape, and CLAUDE.md's incidental-findings rule voids "it's pre-existing" / "it's a different file" as reasons to defer.

`release-body.test.mjs:377` is the third instance of the same filter (see 🟡 1 for why that shape is the milder one).

### 🟡 Minor

**1. Two more case-sensitive filters survive in the PR's own file — `scripts/tests/bump-version.test.mjs:748` and `:1201`.** Both save-and-delete git vars off the *real* `process.env`:

```js
for (const key of Object.keys(process.env)) {
  if (key.startsWith('GIT_')) {            // 748 (and 1201) — case-sensitive
    savedGitEnv[key] = process.env[key];
    delete process.env[key];
  }
}
process.env.GIT_DIR = resolve(decoy, '.git');   // 753
```

The asymmetry is real: the *save* is case-sensitive but the paired teardown `delete process.env.GIT_DIR` (line 772) is case-**in**sensitive on Windows. So an ambient lowercase `git_dir` is never captured into `savedGitEnv`, is silently clobbered by line 753 (I confirmed Windows keeps the stored lowercase casing on a differently-cased assignment), and is then deleted rather than restored.

Reporting this as cleanup, not a bug: **I could not make it produce a wrong suite result.** Every git call in the file now routes through the fixed `cleanGitEnv()` or through `scrubGitEnv()`, both case-insensitive, so the surviving var is inert. Running the full file with a hostile ambient `git_dir` gave 59/59. It is an inconsistency the PR's own reasoning argues against, not a demonstrated break.

**2. `scripts/tests/bump-version.test.mjs:641` — the new comment's cross-reference points the wrong way.** It reads *"which scrubGitEnv() deliberately preserves — see the #2216 note **below**"*. The only `#2216` notes in this file are at lines **607** and **633**, both *above* line 641; `grep -n '#2216'` returns nothing after it. Should be "above". (For contrast, line 638's *"hardened against below (#2175 review, Finding 1)"* **is** correct — that block is at line 664.)

**3. `scripts/tests/bump-version.test.mjs:652` and `:657` are non-load-bearing.** The deletion loop at 647-649 is `GIT_`-scoped, so `PATH` was never at risk; and `cleanGitEnv()` cannot strip a key named `PATH` under any casing. So `assert.ok(cleaned.PATH ?? cleaned.Path)` **cannot fail**, and `process.env.PATH = saved.PATH ?? saved.Path` is a no-op. I checked the leak you'd worry about and it is not there — on Windows `PATH` is stored as `Path` under PowerShell and cmd, and the assignment plus the wholesale restore round-trips cleanly (91 keys in, 91 out, no lost keys, no changed values, `Path` casing preserved). So: harmless, but it is decoration rather than coverage.

**4. Release-notes pair untouched with no explicit not-applicable.** Skipping is defensible here (test-only change, no shippable delta), but CLAUDE.md's before-shipping step 5 requires saying so in the PR body rather than omitting silently. The body's Test plan doesn't mention it.

### ✅ What is solid

- **The fix is correct, and locale-safe for the right reason.** `String.prototype.toUpperCase()` is locale-independent. This matters more than it looks: `'git_dir'.toLocaleUpperCase('tr')` → `'GİT_DİR'`, which does **not** `startsWith('GIT_')` — the Turkish-dotless-i trap would have re-opened the exact hole under a `tr-TR` locale. The author used the locale-independent form, matching `scrubGitEnv` (`scripts/git-env.mjs:81`) byte for byte.
- **The regression test is genuinely mutation-sensitive.** Reverting *only* line 53 (leaving the test's own setup loop at 648 intact) fails at `bump-version.test.mjs:655` with `actual: '/decoy/.git'`, `expected: undefined` — the assertion under test, not a setup crash. Restored → green. Red for the reason claimed.
- **The fixture can't be vacuously satisfied.** Probed on this box: Node/Windows preserves the casing a var was created with, so `Object.keys(process.env)` really does return `git_dir` / `Git_Index_File`. Had Windows normalised to uppercase, the pre-fix filter would have caught them and the test would pass unfixed. It doesn't.
- **Not order-dependent or hostile to siblings.** Probed `node:test`: top-level tests run strictly sequentially (`a-start | a-end | b-sync | c-start | c-end`), the mutation is confined to one synchronous body, and `run-hooks-tests.mjs` gives each file its own process. The `finally` restore is faithful on Windows (measured above) and trivially so on POSIX.
- **No collateral stripping.** The widening only ever removes *more*, and every newly-removed key is a git var that is meaningless to a git child on POSIX anyway. No legitimately-needed var that previously survived is now lost.
- **End-to-end load-bearing.** Full file 59/59 in the fixed state, and 59/59 again with a hostile ambient lowercase `git_dir` exported — the scenario the PR was filed for.
- `Closes #2841` present and outside any code span. Single file, one commit, no scope creep — every changed line traces to the stated fix.

### Verdict

The fix itself is correct, complete for its site, and the paired test is a real regression test — I verified all three rather than taking the author's word. One correctness finding stands: the identical helper in `scripts/tests/release-body.test.mjs:61-67` is still armed, demonstrably leaks a throwaway git write into a real repo, and is excluded from the repo's own scrub guard. Under "fix the class, not the site" that belongs in this round. Re-review after it is fixed and pushed.

dudarenok-maker:
Status update: pr-review-gate round 1 (medium depth) found 1 significant defect + 4 minor items, all fixed locally in the worktree (`C:\Claude\Projects\wt-bump-version-test-git-env-case`, branch `fix/scripts-bump-version-test-git-env-case`):

- `release-body.test.mjs` had a byte-identical copy of the same case-sensitivity bug (fixed identically, with a mirrored mutation-provable regression test).
- Two adjacent save/restore blocks with the same latent casing gap (fixed).
- A stale comment reference and a vacuous assertion in the new test (fixed).

71/71 tests passing across both files. This second commit was interrupted mid-`git commit` by a deliberate session/host reboot before it landed — no corruption (git commit is atomic), just not yet committed/pushed. Resuming shortly to finish: commit → push → update PR body with the "also fixed, found in passing" note → review round 2.

dudarenok-maker:
## PR review — pass 2 (head 40531340, depth medium)

Scope: re-review of round 1's five findings against commit `40531340` (39+/8-, `scripts/tests/bump-version.test.mjs` + `scripts/tests/release-body.test.mjs`), plus an adversarial pass over the round-2 diff on its own merits · verified: rebuilt the out-of-tree harness from head (copy of `scripts/` under `$TEMP`, nothing run in the worktree) — 71/71 green; then five separate mutation runs, a POSIX-semantics simulation of the widened save/restore round-trip, and an independent grep of the whole repo for *every* spelling of the git-env scrub, not round 1's list. Deltas only below; round 1's "✅ what is solid" is not repeated.

---

## Round 1's five findings — dispositions

**🟠 1 · `release-body.test.mjs`'s `cleanGitEnv()` — ✅ VERIFIED RESOLVED.**
`release-body.test.mjs:69` now reads `key.toUpperCase().startsWith('GIT_')`, byte-identical to the sibling and to `scrubGitEnv` (locale-independent `toUpperCase`, so the Turkish-dotless-i re-opening round 1 flagged stays closed). The new test at `:405-421` is genuinely mutation-sensitive: reverting **only** line 69 (the setup loop at `:409` deliberately left case-insensitive) fails at the asserted line —
```
test at scripts\tests\release-body.test.mjs:405:1
✖ cleanGitEnv strips a git env override regardless of stored casing
  AssertionError at release-body.test.mjs:415:12
    actual:   '/decoy/.git'
    expected: undefined
```
Red for the reason claimed, not a setup crash. Restored → green, 71/71 across both files.

**🟡 2 · Three more case-sensitive save/restore filters — ✅ VERIFIED RESOLVED, and no restore regression.**
`bump-version.test.mjs:751`, `:1204`, `release-body.test.mjs:382` all widened. I specifically probed the concern that widening the *save* could now mis-restore, including on **POSIX**, where `GIT_DIR` and `git_dir` are two distinct variables that can coexist (CI is Ubuntu). Simulating POSIX env semantics with both present:
```
captured : {"GIT_DIR":"/ambient-upper/.git","git_dir":"/ambient-lower/.git","GIT_AUTHOR_NAME":"Ambient"}
restored : ... GIT_DIR=/ambient-upper/.git, git_dir=/ambient-lower/.git, GIT_AUTHOR_NAME=Ambient
ROUND-TRIP FAITHFUL: true   lost keys: []   changed: []
```
Both casings captured under their own key, both deleted, both restored with original casing *and* value — no collapse, no clobber. The new test's wipe-all + `Object.assign` teardown is faithful under the same conditions. And the new test goes **red on POSIX too** with the fix reverted (`cleaned.git_dir === '/decoy/.git'`), so it is a real CI gate, not a Windows-only-green test. Collateral of the widening on POSIX is limited to genuine `git_*` keys (`gitconfig_notavar` correctly survives).

**🟡 3 · Stale `#2216` cross-reference — ⚠️ HALF RESOLVED, and the other half now points the wrong way.** See 🟡 A below.

**🟡 4 · Vacuous `PATH` assertion — ✅ VERIFIED RESOLVED, and the replacement is not vacuous.**
The no-op `process.env.PATH = saved.PATH ?? saved.Path` is gone, and both replacement lines are live — proven by two independent mutations of `cleanGitEnv()`:
- also strip `PATH` → **RED at `bump-version.test.mjs:659`**, `AssertionError: PATH must survive the scrub`;
- keep `PATH` but blank its value → **RED at `:660`**, `actual: ''` vs the real `Path` string.

So `:659` and `:660` each catch something the other doesn't. Exactly what round 1 asked for.

**🟡 5 · Release-notes not-applicable — ✅ VERIFIED RESOLVED.** PR body now carries the explicit statement and cites CLAUDE.md step 5's no-shippable-delta exemption. `Closes #2841` still present and outside any code span.

---

## 🟠 Significant — correctness bug (NEW): the class is *not* closed. `verify-cache.test.mjs` has the same defect in a different spelling, and it is worse

Round 1's framing was "fix the class, not the site", and the PR body now claims the siblings were "fixed alongside rather than left armed". Both round 1's search and the PR's fix set keyed on the *spelling* `startsWith('GIT_')`. `scripts/tests/verify-cache.test.mjs:1402-1420` spells the identical scrub as an **object destructure on hardcoded uppercase names**, which is equally case-sensitive and which that grep cannot see:

```js
function gitAt(cwd, args) {
  // "...without stripping them these fixture commands would operate on the
  //  REAL repo instead of the throwaway `cwd`."          <- its own comment
  const {
    GIT_DIR: _GIT_DIR,          // 1409 — case-SENSITIVE, same as the bug this PR fixes
    GIT_WORK_TREE: _GIT_WORK_TREE,
    GIT_INDEX_FILE: _GIT_INDEX_FILE,
    GIT_PREFIX: _GIT_PREFIX,
    ...cleanEnv
  } = process.env;
  const r = spawnSync('git', args, { cwd, encoding: 'utf8', env: cleanEnv });
```

Same directory, same `git-scrub.test.mjs:988-991` exclusion (`'scripts/tests/** must be excluded from the scan'`), same battery — `run-hooks-tests.mjs` globs `scripts/tests/*.test.mjs`, so this file runs in `test:hooks`, i.e. inside `verify:fast:scoped` (pre-commit) and `verify:fast:branch` (pre-push), exactly like the two files this PR fixed. And `gitAt()` drives **writes**: `makeGitFixture()` at `:1425-1434` runs `init` / `config` / `add` / `commit`, and `:1438` cuts a branch. Second identical destructure at `:1584-1590`.

Demonstrated, not inferred. Ambient lowercase `git_dir` → the verbatim destructure → the verbatim `makeGitFixture()` steps, against a throwaway victim repo in `$TEMP`:

```
ambient GIT_-ish keys as stored : ["GIT_ASKPASS","git_dir","GIT_EDITOR"]
survivors after the destructure : ["GIT_ASKPASS","git_dir","GIT_EDITOR"]   <- git_dir passed straight through

ok  : init -q -b main
ok  : config user.email test@example.com
ok  : config user.name Test
FAIL: add .          -> fatal: this operation must be run in a work tree
FAIL: commit -q -m base -> fatal: this operation must be run in a work tree
```

Damage measured on the victim repo:

```
--- VICTIM .git/config BEFORE ---        --- VICTIM .git/config AFTER ---
  bare = false                             bare = true            <-- flipped
  email = dudarenok@example.com            email = test@example.com
  name  = Real Operator                    name  = Test

THROWAWAY has its own .git?  (none)
```

Exact input → wrong output: with `git_dir` set, `gitAt(throwaway, ['init','-q','-b','main'])` **re-initialises the operator's real repository as bare** (`core.bare false → true`), `gitAt(throwaway, ['config', ...])` overwrites its committer identity, the throwaway gets no `.git` at all, and the suite then dies with a `fatal: this operation must be run in a work tree` that names nothing. That is strictly worse than round 1's release-body case (a stray tag): a `core.bare = true` flip makes every subsequent git command in the real checkout fail until someone finds and reverts it by hand.

This is defect shape 2 ("a guard that enumerates syntax — it loses one spelling per round") sitting on top of shape 10 ("one instance fixed, the class left armed"), and CLAUDE.md voids "it's pre-existing" / "the branch doesn't touch that file" as reasons to defer. Note that fixing it by hand *again* keeps the class open for the next spelling — the durable close is to route both sites through a case-insensitive helper (`scrubGitEnv()` + a case-insensitive `GIT_INDEX_FILE`/`GIT_PREFIX` strip, the shape `publish-token-git.test.mjs:63-69` already uses correctly).

For completeness, my sweep of **every** git-env scrub in the repo, so this can be closed once rather than one spelling per round:

| Site | Shape | Verdict |
|---|---|---|
| `scripts/git-env.mjs:79-86` `scrubGitEnv` | `key.toUpperCase()` set lookup | ✅ |
| `scripts/tests/bump-version.test.mjs:50-56` | fixed this PR | ✅ |
| `scripts/tests/release-body.test.mjs:66-72` | fixed this PR | ✅ |
| `scripts/tests/publish-token-git.test.mjs:63-69` | `scrubGitEnv()` + `toUpperCase() === 'GIT_INDEX_FILE'` | ✅ |
| `scripts/tests/gitignore-secrets.test.mjs:30`, `module-graph.test.mjs:155` | `scrubGitEnv()` | ✅ |
| **`scripts/tests/verify-cache.test.mjs:1408-1414`** | uppercase destructure | ❌ **the finding** |
| **`scripts/tests/verify-cache.test.mjs:1584-1590`** | uppercase destructure | ❌ **the finding** |
| `scripts/verify-cache.mjs:960` | `scrubGitEnv()` then `GIT_PREFIX: _GIT_PREFIX` destructure | ⚠️ base is fine; the `GIT_PREFIX` strip alone is case-sensitive. Same one-line edit — worth folding in; I could not demonstrate a break, so not a finding on its own. |
| `scripts/tests/review-gate-mechanism.test.mjs:398` | `git ls-files` with **no** env scrub | ⚠️ read-only against `REPO_ROOT`; a leak makes the agents/role-table drift guard read a foreign repo, but it fails loud rather than passing vacuously. Noted, not a finding. |

---

## 🟡 Minor

**A · cleanup — `bump-version.test.mjs:638`: this commit *introduced* a wrong cross-reference while fixing the other one, against round 1's explicit note.**

The diff changed two direction words. `:641` (`#2216` → "above") is **correct**: `grep -n '#2216'` gives lines 607 and 633, both above. But `:638` was also changed, from "below" to "above":

```js
// trap scrubGitEnv() was hardened against above (#2175 review, Finding 1).
```

`grep -n '#2175'` returns `418, 453, 485, 638, 667, 691, 722`. The only **`#2175 review, Finding 1`** block in the file is at line **667** — *below* 638. Round 1 called this line out by name as the one that was already right: *"For contrast, line 638's 'hardened against below (#2175 review, Finding 1)' **is** correct — that block is at line 664."* It was changed anyway, so the file now has one accurate pointer and one that sends the reader the wrong way. Revert `:638` to "below"; keep `:641` as "above". (Defect shape 9 — the defect is in the document, not the code.)

**B · cleanup — `release-body.test.mjs:405-421` is a partial mirror: it does not pin the over-broad direction, and nothing else in the file does either.**

The comment at `:400-404` says it "Mirrors that file's regression test directly". It doesn't — it carries the two `GIT_` assertions but omits the sibling's `:658-660` "non-`GIT_` keys survive with their values" guard, which mutation-testing above proved is load-bearing in `bump-version.test.mjs`. Consequence, measured: gutting the helper to the maximally over-broad `cleanGitEnv() { return {}; }` leaves the new test **green**, and — because Node resolves `git` from the *parent's* PATH, not the child env (`execFileSync('git',['--version'],{env:{}})` → `git version 2.54.0.windows.1`) — the whole file stays green too:

```
release-body.test.mjs with cleanGitEnv() -> {} :   ℹ tests 12   ℹ pass 12   ℹ fail 0
```

I am grading this cleanup, not a correctness bug, and want to be plain about why: an over-broad scrub is *safer* for isolation, not less safe, so there is no live hazard here. What it does mean is that this file's git-isolation contract is unpinned in one direction while its sibling's is pinned — the asymmetry the comment denies. Adding the sibling's three lines closes it.

---

### Verdict

Four of round 1's five findings are **verified resolved on evidence, not on the author's word** — mutation-proven at the asserted line, plus a POSIX round-trip check that clears the "widened filter mis-restores on Linux" risk the re-review was asked to probe. The fifth (🟡 3) is half resolved and half regressed: `:641` fixed, `:638` broken.

One new correctness bug stands, and it is the same finding round 1 raised, one spelling further out: `scripts/tests/verify-cache.test.mjs:1408-1414` and `:1584-1590` scrub the git env case-sensitively via destructuring, run in the same `test:hooks` battery under the same guard exclusion, and — demonstrated end-to-end — flip a real repository's `core.bare` to `true` and overwrite its committer identity when an ambient lowercase `git_dir` is present. The PR's claim that the class was "fixed alongside rather than left armed" is not yet true. Re-review after it and 🟡 A are fixed and pushed.

dudarenok-maker:
## PR review — pass 3 (head b827b5ce, depth medium)

Scope: re-review of round 2's three fixes against commit `b827b5ce` (`bump-version.test.mjs` + `release-body.test.mjs` + `verify-cache.test.mjs`), plus a final normal-depth pass over the whole repo for any git-env scrub spelling the first two rounds missed · verified: rebuilt the out-of-tree harness from head (copy of `scripts/` under `$TEMP`, `NODE_PATH` at the real `node_modules`, nothing run inside either checkout), then **six mutation runs, a GIT_PREFIX semantics probe against real git, and an end-to-end victim-repo damage measurement in both directions**. Deltas only below; rounds 1–2's "✅ what is solid" is not repeated.

**This is the cap round (3 of 3).** Verdict up front: **no correctness bug remains.** Three cleanup findings, none of which re-trigger the loop.

---

## Round 2's findings — dispositions

**🟠 1 · `verify-cache.test.mjs`'s uppercase destructure — ✅ VERIFIED RESOLVED, at the unit level *and* end-to-end.**

`verify-cache.test.mjs:1415-1421` now routes both former destructure sites through the same `cleanGitEnv()` helper as its two siblings. I re-derived this rather than reading the diff.

*Unit:* reverting **only** the helper to the exact pre-fix destructure (leaving the test's own setup loop at `:1456` case-insensitive, so a failure can't be a setup crash) fails at the asserted line —

```
test at scripts\tests\verify-cache.test.mjs:1455:1
✖ cleanGitEnv strips a git env override regardless of stored casing
  AssertionError at verify-cache.test.mjs:1465:12
    actual:   '/decoy/.git'
    expected: undefined
```

(`:1455`/`:1465` are the shipped `:1452`/`:1462` plus the mutant's 3-line growth.) Red for the reason claimed. Restored → green.

*End-to-end*, which is what actually settles round 2's damage claim. A throwaway VICTIM repo in `$TEMP`, an ambient **lowercase** `git_dir` pointing at it (confirmed the child really receives that spelling: `child sees GIT_-ish keys : ["git_dir"]`), each suite run in full:

| File | pre-fix helper | result | victim damage |
|---|---|---|---|
| `bump-version.test.mjs` | `startsWith('GIT_')` | 20 pass / 39 fail | `bare: false→true`, `email`+`name` overwritten |
| `release-body.test.mjs` | `startsWith('GIT_')` | 1 pass / 11 fail | `bare: false→true`, `email`+`name` overwritten |
| `verify-cache.test.mjs` | uppercase destructure | 120 pass / 7 fail | `bare: false→true`, `email`+`name` overwritten |

…and with the shipped code, same hostile `git_dir`:

| File | result | victim damage |
|---|---|---|
| `bump-version.test.mjs` | 59/59 pass | **NONE** |
| `release-body.test.mjs` | 12/12 pass | **NONE** |
| `verify-cache.test.mjs` | 126/127 pass¹ | **NONE** |

¹ the single failure is `test:hooks completeness guard`, a harness artifact — it walks the full repo tree, which the `$TEMP` copy doesn't have. Present identically at baseline with no `git_dir` set, so it is not a PR defect.

`core.bare false→true` is confirmed as the real hazard: all three files caused it pre-fix, none do post-fix.

**🟡 A · `bump-version.test.mjs:638` cross-reference revert — ✅ VERIFIED RESOLVED, and it broke nothing.**
`:638` reads "hardened against **below** (#2175 review, Finding 1)". `grep -n '#2175'` → 418 (Finding 3), 453 (Finding 4), 485 (Finding 5), 638, **667 (Finding 1)**, 691 (Finding 6), 722 (Finding 2). Finding 1 is the only one of its name and sits at **667, below 638** — so "below" is correct. The adjacent `:641` "see the #2216 note **above**" also still holds: `#2216` notes are at 607 and 633, both above. Round 1 broke one pointer, round 2 restored it, and **both pointers are now accurate simultaneously** — which was not true after either previous round.

**🟡 B · `release-body.test.mjs`'s partial mirror — ✅ VERIFIED RESOLVED.**
`:417-421` now carries the sibling's "non-`GIT_` keys survive" half, and it is load-bearing: the maximally over-broad mutant `cleanGitEnv() { return {}; }` — which round 2 measured as leaving the file **green** — now fails at the asserted line:

```
✖ cleanGitEnv strips a git env override regardless of stored casing
  AssertionError: PATH must survive the scrub   (release-body.test.mjs:416:12)
    actual: undefined   expected: true
```

---

## 🟡 Minor (cleanup only — none re-triggers a review round)

**1 · The round-2 commit fixed 🟡 B in `release-body.test.mjs` and reproduced it in the new `verify-cache.test.mjs` test, in the same commit.**

`verify-cache.test.mjs:1452-1468` carries only the two `GIT_` assertions and omits the "non-`GIT_` keys survive with their values" half that both siblings now have. Measured, same mutant as above:

```
verify-cache.test.mjs with cleanGitEnv() -> {} :  ℹ tests 127  ℹ pass 126  ℹ fail 1
                                                  (the harness artifact only — the new test PASSES)
```

So the over-broad direction is pinned in `bump-version.test.mjs` and `release-body.test.mjs` and unpinned in the file added to close this exact gap. Its own comment at `:1448-1449` — *"Mirrors those files' regression test directly against the local helper"* — is therefore false in the same way round 2 found `release-body.test.mjs:400-404`'s was (defect shape 9 on top of shape 10).

Grading this cleanup for the same reason round 2 graded its twin cleanup: an over-broad scrub is *safer* for isolation, so there is no live hazard. The fix is the sibling's three lines, copied from `bump-version.test.mjs:656-660`.

**2 · `scripts/verify-cache.mjs:960` — the last case-sensitive git-env spelling in the repo, still unfixed and now undeclared.**

```js
function gitEnv() {
  const { GIT_PREFIX: _GIT_PREFIX, ...cleanEnv } = scrubGitEnv();   // 960
  return cleanEnv;
}
```

Round 2 flagged this as "worth folding in". It was not folded in, and unlike every other item it is not mentioned in the PR body's two "Also fixed, found in passing" sections. It is **production** code (the three test-file fixes are not), feeding `stagedDiffFiles` (`:912`), `branchDiffFiles` (`:965`) and `gitFileList` (`:1194`).

**New evidence, and it argues *down*, not up.** I tried to demonstrate the break round 2 couldn't, and instead found the strip is inert in both casings. Against a real repo with a subdirectory, staged content and a feature branch:

```
=== VERDICT: does GIT_PREFIX change any command output? ===
  SAME  stagedDiffFiles      under GIT_PREFIX=sub/   and  git_prefix=sub/
  SAME  gitFileList          under GIT_PREFIX=sub/   and  git_prefix=sub/
  SAME  branchDiffFiles(a)   under GIT_PREFIX=sub/   and  git_prefix=sub/
  SAME  branchDiffFiles(b)   under GIT_PREFIX=sub/   and  git_prefix=sub/
ANY DIFFERENCE AT ALL: false
```

None of the four commands takes a pathspec, which is the only thing `GIT_PREFIX` influences — so even the *uppercase* strip that `gitEnv()` does perform is defensive rather than load-bearing. **Confirmed cleanup, not a correctness bug** — round 2's grading was right, and this is the re-derivation rather than a repeat of its wording.

Worth stating precisely why the guard never catches it, since that is the durable part: `git-scrub.test.mjs`'s one-hop wrapper resolution checks only that the wrapper body *textually calls* `scrubGitEnv()` (its own fixtures at `:757-780` are exactly `function gitEnv() { return scrubGitEnv(); }`). It has no model of a hardcoded-uppercase key being re-stripped off the result afterward, so `gitEnv()` passes the guard while remaining case-sensitive on the `GIT_PREFIX` axis.

**3 · The class is now closed by hand three times over, and nothing mechanical stops a fourth.**

`git-scrub.test.mjs:988-991` asserts, as a positive requirement, that `'scripts/tests/** must be excluded from the scan'`. So the three byte-identical hand-rolled `cleanGitEnv()` copies (`bump-version.test.mjs:50-56`, `release-body.test.mjs:66-72`, `verify-cache.test.mjs:1415-1421`) are permanently outside the only guard that could hold them. Concrete scenario, and it is this PR's own three-round history: a new `scripts/tests/*.test.mjs` that spawns git and hand-rolls `startsWith('GIT_')` runs in `test:hooks` (`run-hooks-tests.mjs:10` globs `scripts/tests/*.test.mjs`), reintroduces the `core.bare` flip measured above, and **no test goes red**. Round 2 said this in as many words — *"fixing it by hand again keeps the class open for the next spelling"* — and the round-2 commit added a third copy by hand.

I am flagging this as the one item that plausibly wants the **judgment-call / design-pass** route rather than a blind fix at the cap: extract a shared test helper, or narrow the `scripts/tests/**` exclusion so the guard covers test files too, or accept the duplication deliberately. Those have different costs and more than one defensible answer, so it is a decision rather than a defect — but it should be a recorded decision, not an omission.

---

### Also checked, explicitly not findings

- `release-body.test.mjs:111` / `:131` — the two `git tag` writes that bypass `gitExec()`. Both **do** pass `env: cleanGitEnv()` / the derived local. Clean.
- `publish-token-git.test.mjs:63-69` — `scrubGitEnv()` + `toUpperCase() === 'GIT_INDEX_FILE'`. Clean, and its header already documents the constant-vs-function trap.
- Every remaining `startsWith('GIT_')` under `scripts/` is either the correct `toUpperCase().startsWith(…)` form or prose in a comment. No live case-sensitive prefix filter survives anywhere in `scripts/`.
- `server/src/routes/worktrees.ts:85` — `git worktree list --porcelain` with no env scrub *and* no `cwd`, and `server/**` is entirely outside `git-scrub.test.mjs`'s scan scope. Different defect class (no-scrub, not case-sensitivity), read-only, and 404s in production. I could not construct a realistic wrong output, so it is an observation, not a finding, and not this PR's to carry.
- House gates: `Closes #2841` present and outside any code span; #2841 is OPEN and labeled `bug` + `area:ops`, matching CONTRIBUTING's bug-shaped convention. Release-notes pair explicitly marked not-applicable in the body with the CLAUDE.md step 5 citation. No on-box acceptance owed (no hardware-provable behaviour). No config knob, no derived artifact, no `cast.json` write touched.

---

### Verdict

Round 2's three fixes are **all verified resolved on independently re-derived evidence** — a mutation at the asserted line, a two-directional end-to-end victim-damage measurement, and a direct check that the reverted comment leaves both cross-references simultaneously correct for the first time in this PR. The PR's central claim now holds under test: with a hostile ambient lowercase `git_dir`, all three suites pass and a real repository takes zero damage, where pre-fix all three flipped its `core.bare` to `true` and overwrote its committer identity.

**No 🔴 or 🟠 finding stands, so the re-review trigger is not met and the loop closes cleanly at the cap.** The three 🟡 items are cleanup: two are one-liners (the missing mirror half in `verify-cache.test.mjs`, the `GIT_PREFIX` destructure in `verify-cache.mjs`), and the third is a genuine design decision about how — or whether — to guard `scripts/tests/**` against a fourth hand-rolled copy. Mergeable on the correctness axis as it stands.

Issue #2843, opened 2026-09-01
[agent instructions][claude][verify] Verify (2nd pass): widen wrongId to the discharge class


dudarenok-maker:
## Parent

#2721 — decision made in [this comment](https://github.com/dudarenok-maker/Castwright/issues/2721#issuecomment-5484703995): widen `wrongId` to fatal for full-subject discharge, after first fixing `buildLegitimateSubjectMap`.

This is the SECOND verify pass. The first (#2832) FAILED against branch `chore/ops-2721-widen-wrongid` commit `c5222146`: `buildLegitimateSubjectMap` was never touched (a special case was bolted on instead), and the multi-row-subject-with-a-live-sibling case was untested and reproducibly fires fatal `wrongId` when it must not. See #2832's closing comment for the full writeup and the reproduction. The rework child (linked alongside this issue) addresses both.

## Parent's acceptance, restated as a checklist

- [ ] `buildLegitimateSubjectMap` (in `scripts/check-register-citations.mjs`) is fixed so a subject's mapped ID-set is complete/correct — the two pre-existing benign `unknownSubject` residuals on the current corpus are resolved by this fix (not papered over, not excluded by a special case).
- [ ] `wrongId` becomes fatal specifically when EVERY id in a cited subject's mapped set has left the register (full discharge) — confirmed by a test.
- [ ] A multi-row subject with at least one still-live row does NOT trigger fatal `wrongId` for a citation to one of its discharged rows — confirmed by a test (this is the specific case #2721's issue body calls out: "discharging one row of a multi-row subject does not remove the subject"). **This is the point the first verify pass (#2832) found broken — verify it with an adversarial fixture, not just by reading the new test's assertions.**
- [ ] Re-minting a previously-discharged row's ID (a NEW row later assigned an ID that used to belong to a fully-discharged subject) still triggers fatal `wrongId` correctly — confirm this existing behavior (verified correct in #2717's review) is not broken by the widening.
- [ ] The two previously-benign `unknownSubject` residuals mentioned in the original issue are now either resolved (preferred, since the map fix should have fixed them) or the checker's test suite demonstrates they were never real gaps once the map is correct.
[...]

Issue #2832, opened 2026-08-31
[agent instructions][claude][verify] Verify: widen wrongId to the discharge class


dudarenok-maker:
## Parent

#2721 — decision made in [this comment](https://github.com/dudarenok-maker/Castwright/issues/2721#issuecomment-5484703995): widen `wrongId` to fatal for full-subject discharge, after first fixing `buildLegitimateSubjectMap`.

## Parent's acceptance, restated as a checklist

- [ ] `buildLegitimateSubjectMap` (in `scripts/check-register-citations.mjs`) is fixed so a subject's mapped ID-set is complete/correct — the two pre-existing benign `unknownSubject` residuals on the current corpus are resolved by this fix (not papered over, not excluded by a special case).
- [ ] `wrongId` becomes fatal specifically when EVERY id in a cited subject's mapped set has left the register (full discharge) — confirmed by a test.
- [ ] A multi-row subject with at least one still-live row does NOT trigger fatal `wrongId` for a citation to one of its discharged rows — confirmed by a test (this is the specific case #2721's issue body calls out: "discharging one row of a multi-row subject does not remove the subject").
- [ ] Re-minting a previously-discharged row's ID (a NEW row later assigned an ID that used to belong to a fully-discharged subject) still triggers fatal `wrongId` correctly — confirm this existing behavior (verified correct in #2717's review) is not broken by the widening.
- [ ] The two previously-benign `unknownSubject` residuals mentioned in the original issue are now either resolved (preferred, since the map fix should have fixed them) or the checker's test suite demonstrates they were never real gaps once the map is correct.
- [ ] All three orphaned sites are rehomed to cite #2721 instead of the closed #2629: `scripts/check-register-citations.mjs` ~:26-37, `scripts/check-register-citations.mjs` ~:209-210, and `docs/testing/onbox-acceptance-register.md`'s allocation-procedure blockquote.
- [ ] Automated tests exist with a mutation-provable red/green pair for the new fatal case.

## What to do

1. Fetch branch `chore/ops-2721-widen-wrongid`. Read-only review.
2. Re-run any mutation whose receipt asserts it without showing output — specifically: construct a fixture where a full-subject discharge citation exists, confirm it FAILS with observed output; construct a fixture where a multi-row subject has one live row, confirm a citation to its discharged sibling does NOT fail, with observed output.
[...]

PR #2846, opened 2026-09-01
chore(ops): widen wrongId to full discharges, fix multi-row false positive

### LINKED ISSUE #2721: ops-64 — rehome the wrongId-widening deferral that #2629's closure orphans

dudarenok-maker:
Closes #2721

## Summary

- Widens `check-register-citations.mjs`'s Check C `wrongId` to the full-discharge class: when a cited register row ID's own current heading now tracks a *different, known* subject, the citation is fatal `wrongId` unless a nearby discharge annotation excuses it (`annotatedDischarge`, non-fatal).
- Fixes `buildLegitimateSubjectMap` (not a bolted-on special case, per #2832's finding) to return both a current subject→id map and a **historical** one built from each row's own discharge/re-mint annotations (`dischargedSubjectsMentionedIn`), so `recordSubjectConflict` can tell a historically-legitimate sibling apart from a genuinely wrong id.
- This closes the exact gap #2832 (first verify pass) found: a multi-row subject that loses only one of its rows — not a full discharge — no longer fires fatal `wrongId` for a citation to the discharged/re-minted sibling, because the subject still has a live row elsewhere.
- Second-pass verify of this rework is #2843.

## Review fixes (found in passing, per `pr-review-gate`)

Three independent review passes surfaced findings, all fixed in this PR rather than deferred (except one design-decision carve-out, filed as its own issue per this repo's own findings-triage rule):

**Pass 1** ([comment](https://github.com/dudarenok-maker/Castwright/pull/2846#issuecomment-5500758478)):
- 🔴 CRITICAL: `dischargedSubjectsMentionedIn` failed open on negated/distant prose (e.g. a "NOT discharged" mention elsewhere in a row's body falsely excused a genuinely wrong citation) — fixed with a clause-bounded, proximity-capped polarity check.
- 🟠 A34's heading was unnecessarily de-anchored in `docs/testing/onbox-sitting-cloning-identity.md` — restored, keeping the discharge annotation.
- 🟡 `isOwnedPrCompanion` was only checked in one of two code branches — moved to the top of the function, unconditional.
- 🟠 `pr-title-lint` was red (title over the 100-char cap) — shortened.
- 🟡 No release-notes pair — added to both `docs/release-notes-next.md` and `RELEASE_NOTES.md`.

**Pass 2** ([comment](https://github.com/dudarenok-maker/Castwright/pull/2846#issuecomment-5504167608)), verifying the pass-1 fixes surfaced three more live gaps in the same mechanism plus documentation drift:
- 🟠 The polarity scan's negation-word set was too narrow (missed "isn't"/"never"/"un-discharged"/etc.) and its 30-char flat window could bleed across unrelated clause/list-item boundaries — reworked to bound to the clause and recognize a broader negation set.
- 🟠 The subject-number harvester didn't distinguish a genuine subject reference from a "PR #nnnn" reference, letting an unrelated PR number in a row's body register as a historically-owned subject — fixed by reusing the file's existing `extractPrSubjectNumbers` PR-vs-subject distinction.
- 🟠 The docstring and release-notes prose both overclaimed what the fixes actually do. Corrected.
- Self-caught during this round: the negation-pattern regex literal (`isn't`/`aren't`/etc.) desynced `server/src/spawn-windows-hide.test.ts`'s quote-tracking scanner (a real, load-bearing guard, #2747) — fixed by escaping the apostrophes per this file's own existing convention. Also self-caught: the first cross-list-item regression test was a no-op (asserted nothing, empty fixture) — rewritten with a real citation and a mutation-verified fixture.

**Pass 3** ([comment](https://github.com/dudarenok-maker/Castwright/pull/2846#issuecomment-5509942350)), the final round under this repo's loop cap, verifying pass 2's fixes independently rather than trusting them:
- 🟠 The pass-2 clause-bounding fix over-corrected: `CLAUSE_BOUNDARY_REGEX` treats a whole paragraph as one "clause" (it doesn't split on ". "), so an unrelated negation 85+ chars back in the same paragraph could suppress a genuine discharge — inverting the exact false-closed problem the fix was meant to solve. Fixed by capping the polarity scan to whichever is closer, clause start or 30 chars back, which keeps the pass-2 cross-list-item fix intact (the clause boundary already handles that case) while restoring the tight proximity real negations measure at.
- 🟠 A second live bug in the same scan: bare `no` matched inside `no longer exists`, itself one of `DISCHARGE_ANNOTATION_REGEX`'s own AFFIRMATIVE phrases — could misread a later, unrelated `discharged` as negated. Fixed with a negative lookahead.
- 🟡 Six stale "21 [of/-of-] 65" register-coverage counts (two written by this PR, four pre-existing but surfaced by the same pass) corrected to the current measured "15 of 66".
- 🟠 A release-notes overclaim: `idSpecificAnnotationPresent` (the citation-side annotation matcher Check A's fatal path also uses) does NOT have the same polarity check `dischargedSubjectsMentionedIn` does. This is a genuine design-pass carve-out (arming it there turns 4 of 23 currently-annotated Check A citations fatal, and whether those are real defects needs someone to read them) — filed as #2858 rather than fixed inline, and the release note corrected to stop claiming parity.
- Self-caught during this fix: my own first attempt at escaping the new negation regex's apostrophes had a bug (the "escaped apostrophe" variable held a real apostrophe, not the literal `'` text) — the spawn-windows-hide guard caught it again at commit time; fixed and reverified (20/20).

## Test plan

- [x] `node --test scripts/tests/check-register-citations.test.mjs` — **119 passed, 0 failed** (up from the original 107; review fixes across all three passes added 12 regression tests).
- [x] Adversarial fixture, full-subject discharge (mutation-provable): a citation to an id whose own row now tracks a different, known subject, with no discharge annotation, fires `wrongId` — reverting the `row.issues.size > 0` branch turns it non-fatal. Verified passing.
- [x] Adversarial fixture, multi-row subject with a live sibling (the exact #2832 reproduction): a subject with rows A22 (live) and A34 (re-minted, its own body documents its #2040 history); citing A34 for #2040 with **no** annotation at the citation does **not** fire `wrongId` — it resolves to non-fatal `annotatedDischarge`, naming the live sibling A22. Paired control: with A22 removed (full discharge), the same citation still resolves non-fatal. Paired control: a genuinely unrelated id (A99, no historical tie to the subject) still fires fatal `wrongId` even though the subject has a live sibling elsewhere. All three verified passing.
- [x] Re-minted-ID-still-fatal existing behavior (#2717) preserved — covered by the "genuinely unrelated id" control above and the pre-existing "control — a GENUINELY WRONG multi-ID heading" test, both still green.
- [x] `node scripts/check-register-citations.mjs --strict` against the real corpus — **exit 0**, no fatal findings. The two previously-benign `unknownSubject` residuals (A19/#1893, A31/#2037) resolve into `annotatedDischarge`; A34's two citations likewise; one unrelated, pre-existing structural-gap `unknownSubject` warning remains (A3/#1230). Coverage-gap count now correctly reports "15 of 66 rows".
- [x] `server/src/spawn-windows-hide.test.ts` (20/20) — confirmed green after both apostrophe-escaping regressions (pass-2 and pass-3 fix attempts) were caught and corrected.
- [x] `node scripts/check-onbox-register.mjs` — OK, register/live-view still agree.
- [x] All 20 CI checks green as of the latest push (`9b4c2575` head, `chore/ops-2721-widen-wrongid`).


dudarenok-maker:
## PR review — pass 1 (head 7d880abccd12b4f9ae63da129248fa2cfc2cece3, depth medium)

### Scope

`scripts/check-register-citations.mjs` (+320/-41), `scripts/tests/check-register-citations.test.mjs` (+201/-12), `docs/testing/onbox-acceptance-register.md` (+26/-16), `docs/testing/onbox-sitting-cloning-identity.md` (+9/-1). Reviewed at PR head in a detached worktree; no tracked file modified.

**PR body claims, independently verified:**

- `node --test scripts/tests/check-register-citations.test.mjs` → **107 pass / 0 fail** ✅ (confirmed)
- `node scripts/check-register-citations.mjs --strict` → **exit 0** ✅ (confirmed)
- A19/#1893 and A31/#2037 land in `annotatedDischarge` ✅ (confirmed) — though note both reach it via the *citation-side* `idSpecificAnnotationPresent` branch, **not** via the new `historicalOwnersOf` machinery. The `dischargedSubjectsMentionedIn` path added by `7d880abc` has **zero** real-corpus exercise in this run.
- `npm run check:onbox-register` → OK ✅
- `Closes #2721` present and outside a code span ✅

---

## 🔴 Correctness bugs

### 1. CRITICAL — `dischargedSubjectsMentionedIn` fails open, silently downgrading a currently-FATAL real-register detection to non-fatal (a regression vs. `main`)

**`scripts/check-register-citations.mjs:1339-1350`** (consumed at `:1608`, `:1613`, `:1659`)

The new historical-ownership map is built by scanning a row's own body for a discharge *word*, then harvesting **every** `#nnnn` in the surrounding clause. Its docstring (`:1329`) claims it "reuses the exact clause-bounding `idSpecificAnnotationPresent` uses". It does not — it drops the `ID_PROXIMITY_CHARS` (120) cap that `idSpecificAnnotationPresent:970` applies on top of `clauseBounds`, and that cap is precisely what that function's own comment says exists "so an unpunctuated run-on paragraph … can't turn into an unbounded excuse". A plain `. ` is not a `CLAUSE_BOUNDARY_REGEX` boundary, so one markdown paragraph is one clause.

Three compounding weaknesses, all confirmed by execution:

1. **No proximity cap** — measured clause widths in the *real* register: 434, 578, 668, **1308** chars; measured discharge-word→subject distances: 130, 162, 197, 307, 360, 373, 414, 426 chars — all far past 120.
2. **No requirement that the discharge concerns *this* row** — a row body discussing a *different* row's discharge donates that subject to itself.
3. **No polarity check** — a *negated* discharge reads as a discharge record.

All three are live on the unmodified register today. Measured: **6 rows** yield a non-empty historical set, producing **14 (id, subject) pairs** that would now be downgraded. Two of them come straight from negated prose:

- `A16` body: `> **PARTIALLY evidenced 2026-07-31 — NOT discharged.** Corrected after [#1972](…)` → donates **#1972** to A16.
- `A1` body: `[**#1969**](…) is why A16 below is not fully discharged.` → donates **#1969** to A1. This is A1's body talking about **A16**, with a **negated** discharge word, 98 chars away.

**Concrete failure scenario** — no fixture, real `docs/testing/onbox-acceptance-register.md`, real `#1969` (legitimately owned by **A36**). A document containing:

```
### A1 · Some run sheet section (#1969)
```

| | result |
|---|---|
| `main` (75a86ed8) | `wrongId=1` → **FATAL**: `cited A1 for #1969, but the register's #1969 maps to A36, not A1` |
| this PR (7d880abc) | `wrongId=0`, `annotatedDischarge=1` → **passes the gate** |

And the message it prints instead is factually false:

> `cited A1 for #1969 … — A1's own row documents that it once tracked #1969 too, before discharging/being re-minted …`

A1's row documents no such thing; it mentions that **A16** is *not* discharged. This is exactly the mechanically-wrong-citation class PR #2630 exists to catch, now waved through — the false negative the widening was meant to eliminate, reintroduced by the same commit.

**The tests cannot see this dimension.** I mutated `dischargedSubjectsMentionedIn` to apply the same `ID_PROXIMITY_CHARS` cap `idSpecificAnnotationPresent` uses and re-ran the suite: **100 pass / 7 fail before and after, identical** (the 7 are environmental — CLI tests needing the real repo root — and fail on the unmutated baseline at the same path too). Zero tests pin clause width, discharge attribution, or polarity. The three new `#2842` tests all use a one-sentence body where the discharge word, the subject, and the row are the same thing, so they pass under both the current and the capped implementation.

**Doc consequence, same finding:** `docs/testing/onbox-acceptance-register.md:397` — this PR ships prose asserting "a re-mint with NEITHER kind of annotation is still fatal even when a sibling row for the same subject survives." Falsified by the A1/#1969 case above: neither kind of annotation is present at the citation, and it is not fatal. The register now documents a protection the checker does not provide.

Suggested direction (not prescriptive): bind the harvested subject to the discharge word with the same 120-char cap, require the discharge clause to name *this* row's id (reusing `idSpecificAnnotationPresent`'s ID-token binding rather than paraphrasing it), and reject a negated discharge. Re-running the corpus after any of these is cheap — the current shape yields 14 donations; a correct one should yield far fewer.

---

### 2. MEDIUM — the `A34` heading was de-anchored to make the gate green, when the annotation the same hunk adds already sufficed; net effect is lost Check C coverage

**`docs/testing/onbox-sitting-cloning-identity.md:308`**

The PR rewrites `### A34 · Respawn budget …` to `### (discharged, was A34) · Respawn budget …`, with commit `c5222146` explaining it as "re-anchored … so `check-register-citations.mjs`'s widened `wrongId` check does not misread this historical section". But the same hunk *also* adds a `> **Register row: A34 — discharged …**` annotation blockquote, and that annotation alone is already sufficient. Measured against the real register (A34 currently tracks #2040/2570/2584/2640):

| variant | result |
|---|---|
| original heading, no annotation | `wrongId=2` — FATAL |
| **original heading + the new annotation blockquote** | `wrongId=0`, **`annotatedDischarge=2`** — passes |
| shipped de-anchored heading + annotation | `0 / 0 / 0` — **completely silent** |

So the heading rewrite buys nothing the annotation had not already bought, and costs the anchored-heading citation surface: `HEADING_ID_REGEX` no longer matches, so Check C sees nothing here at all and the A34↔#2106 pairing is no longer checked in either direction. It is also inconsistent with how this very PR handled the two other real cases — A19 and A31 in `onbox-sitting-vram-contention.md` keep their headings and are reported in the `annotatedDischarge` bucket, which is the "annotate, don't renumber" outcome the register's own convention prescribes.

Not mentioned in the PR body at all (only in the commit message), so a reviewer reading the body sees a `+9/-1` docs touch and no notice that a citation surface was removed — the "declare incidental fixes in the PR body" gate.

---

### 3. LOW — `isOwnedPrCompanion` is only consulted on the `!legitimateIds` branch, so a PR-companion citation goes FATAL whenever that PR number is another row's subject

**`scripts/check-register-citations.mjs:1611`** (vs. the `else if (!legitimateIds.has(id))` branch at `:1657`)

`recordSubjectConflict`'s docstring calls `isOwnedPrCompanion` "a THIRD, narrower exemption **checked first**". It is checked first only *inside* the `!legitimateIds` branch; the `legitimateIds` non-empty branch never consults it. So the exemption evaporates exactly when the PR number happens to be a live register subject.

**Concrete failure scenario** — register with `### A1 · Tracked work (#1000)` and `### A50 · … (#9998)`; a citation `### A1 · Some section (#1000, PR #9998)`:

```
FATAL: cited A1 for #9998, but the register's #9998 maps to A50, not A1
```

The suite's own `"an untracked \"PR #nnnn\" companion … is silently correct"` test asserts `0/0/0` for the identical shape — it only passes because its `#9998` is owned by nobody. Latent rather than live today (`--strict` is clean), but the collision space is real: the register contains **44** distinct `PR #nnnn` numbers and **73** heading-owned subjects, with **17** numbers appearing in both. Either extend the exemption to the second branch or correct the docstring to say it is scoped to the unknown-subject branch only.

---

### 4. MEDIUM (house gate) — the PR title is 103 chars; the convention caps subjects at 100, and `pr-title-lint` is RED

`Validate PR title against commit convention` → **fail** (run 33558071160). Reproduced locally with `scripts/validate-commit-msg.mjs`: *"Subjects must be 1-100 chars."*

```
chore(ops): widen wrongId to the fully-discharged-subject class, fix multi-row discharge false-positive
```
= 103 characters. Every other check is green. Note this repo's known trap: **editing the title alone does not re-run the workflow** — the check stays red until a push or an explicit re-run.

---

## 🧹 Cleanup

### 5. LOW — no release-notes pair, and no explicit not-applicable

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched, and the PR body does not claim the exemption. CLAUDE.md's before-shipping step 5 permits skipping for an internal chore with no operator-visible effect, but requires saying so explicitly. This change *is* operator-visible — `npm run check:register-citations` now fails a class it previously passed and prints a new non-fatal section — and `docs/release-notes-next.md` carries entries for directly comparable guard/tooling work (ops-19, ops-20, ops-33, ops-46, ops-54). Either add the pair or state the exemption in the body.

### 6. LOW — `dischargedSubjectsMentionedIn`'s docstring overclaims equivalence

**`scripts/check-register-citations.mjs:1329`** — "reuses the exact clause-bounding `idSpecificAnnotationPresent` uses" is not true of the shipped code (the `ID_PROXIMITY_CHARS` cap is dropped). Whichever way finding 1 is resolved, this sentence needs to match the code — a future reader auditing the annotation logic will otherwise reason from a guarantee that was never implemented.

---

## Verdict

**Request changes.** Finding 1 is a correctness regression provable on the unmodified real register with no fixture: a citation `main` currently fails fatally now passes, accompanied by a message asserting something the register does not say. It inverts this PR's own stated purpose and is invisible to the test suite (a mutation adding the missing cap changes zero test outcomes). Findings 2 and 3 are correctness issues of smaller blast radius; finding 4 is a red required-shaped check.

Per the findings-triage rule, ≥1 correctness bug means a re-review pass is owed once the fixes are pushed.

<sub>Verified this pass: 107/107 tests; `--strict` exit 0; `check:onbox-register` OK; head-vs-`main` differential on the real register; mutation test of the proximity cap; real-corpus measurement of the historical map (6 rows, 14 donated pairs) and of the PR-number/subject collision space (17).</sub>

dudarenok-maker:
## PR review — pass 2 (head a8e8b37e715326cb5c16ac29013fd6e0698d98ff, depth medium)

Re-review. Deltas only, per the skill's re-review rule — solid items from pass 1 are not re-listed.

Re-ran independently at this head in the existing `wt-2721-widen-wrongid` worktree (clean, no tracked file modified): `node --test scripts/tests/check-register-citations.test.mjs` → **110 pass / 0 fail**; `node scripts/check-register-citations.mjs --strict` → **exit 0**; all 18 GitHub checks green on this exact SHA. Differentials were run against `origin/main`, `7d880abc` (pass-1 head) and this head from out-of-repo copies of the module.

---

## Disposition of the six pass-1 findings

|---|---|---|
| 1 | 🔴 `dischargedSubjectsMentionedIn` fails open on negated/distant prose | **VERIFIED RESOLVED for the reported repro** — class only partly closed, see new-A/new-B/new-C |
| 2 | 🟠 A34 heading de-anchored | **VERIFIED RESOLVED** |
| 3 | 🟡 `isOwnedPrCompanion` checked in one branch only | **VERIFIED RESOLVED** |
| 4 | 🟠 `pr-title-lint` red | **VERIFIED RESOLVED** |
| 5 | 🟡 no release-notes pair | **Present, but the new prose carries three false claims** — see new-E |
| 6 | 🧹 docstring overclaim | **Partly resolved; a new overclaim was introduced in its place** — see new-C |

**Finding 1** — the exact pass-1 repro, re-measured (real `docs/testing/onbox-acceptance-register.md`, injected `### A1 · Some run sheet section (#1969)`):

| | result |
|---|---|
| `origin/main` | `wrongId=1` — FATAL |
| `7d880abc` (pass-1 head) | `wrongId=0, annotatedDischarge=1` — passes |
| **`a8e8b37e`** | **`wrongId=1` — FATAL, message names A36** |

Restored to `main`'s behaviour. The historical-donation set on the real register drops from **14 pairs / 6 rows** to **2 pairs / 2 rows**. The polarity half is mutation-pinned (removing it turns the new test red); the proximity half is not (new-D).

**Finding 2** — `docs/testing/onbox-sitting-cloning-identity.md:308` is back to `### A34 · Respawn budget … ([#2106](…), PR #2398)` with the annotation blockquote retained, and `--strict` now reports the two A34 citations in `annotatedDischarge` rather than the `0/0/0` silence pass 1 measured. The citation surface is restored.

**Finding 3** — `isOwnedPrCompanion` now returns at `check-register-citations.mjs:1644`, above both branches. Regression test verified red-before/green-after against `7d880abc` for the reason claimed: `### A1 · Title (#1000, PR #1001)` → `wrongId=1` ("#1001 maps to A2/B1, not A1") on the old code, `0/0/0` on this head.

**Finding 4** — title is now 74 chars; `Validate PR title against commit convention` is **success** on `a8e8b37e` (checked against the head SHA, not the PR view, given this repo's known "editing the title doesn't re-run the check" trap).

---

## 🟠 New — correctness

### new-A. The historical-ownership harvest still donates numbers that are not subjects — `A18 ← #1978` is a **live, wrong exemption on the shipped register**

**`scripts/check-register-citations.mjs:1374`** (`clauseText.matchAll(SUBJECT_NUMBER_REGEX)`)

Pass-1 finding 1 named three weaknesses; the fix implemented two (proximity cap, polarity). The third — *nothing requires the harvested number to have any historical relationship to the row* — is untouched, and it is still live.

Real register, `onbox-acceptance-register.md:1613` (A18's body):

> `- **2. Latent equivalence — PARTIALLY DISCHARGED.** Decode equivalence was **measured** during PR #1978's review, …`

`#1978` here is a **PR reference naming where a measurement happened**. The discharge word is about A18's own *item 2*. Nothing in that row documents A18 ever tracking, or discharging, `#1978`. But `historicalOwnersOf(1978)` = `{A18}`.

Measured, no fixture, real register:

| injected line | `origin/main` | **`a8e8b37e`** |
|---|---|---|
| `### A18 · Latent-equivalence follow-up (#1978)` | `unknownSubject` (non-fatal) | **`annotatedDischarge`** (non-fatal) |
| `### A18 · Nothing (#9999)` *(control, no tie)* | `unknownSubject` (non-fatal) | **`wrongId` — FATAL** |

So the exemption is live and discriminating, and the message it prints is false:

> `cited A18 for #1978, but A18 now tracks #1967 — A18's own row documents that #1978 discharged — annotated as discharged, not failing`

Same shape as pass-1 finding 1 (the checker asserting something the register does not say), narrowed but not eliminated. **Not a regression vs `main`** — both are non-fatal — so this is a hole in the *new* fatal class rather than a downgrade, which is why it is 🟠 and not 🔴.

The second surviving donation, `A23 ← #2107` (`:2188`, `**Superseded (#2107, widened by independent review + owner decision, … — see the PARTIALLY DISCHARGED banner`), is the same co-occurrence shape; it is more arguable, since `#2107` is at least topically related to A23.

Worth noting for the fix: this file **already has the primitive** that kills the A18 case — `PR_SUBJECT_REGEX` (`/\bPR\s*\[?#(\d+)/gi`, `:397`), used by `extractPrSubjectNumbers` at `:399`. Excluding PR-marked numbers from the harvest is a mechanical change against an existing, already-tested construct, not a design decision.

Consequence for the doc this PR ships: `docs/testing/onbox-acceptance-register.md:397` describes the exemption as the re-minted row's own section "naming the subject it moved on from". A18's row does not name `#1978` as a subject it moved on from. The code is looser than the prose that ships alongside it.

### new-B. The polarity scan is a flat 30-char character window with a two-word list — it fails open on ordinary negations and fails closed across line breaks

**`scripts/check-register-citations.mjs:1362-1367`**

```js
const polarityScanStart = Math.max(0, dm.index - 30);
const polarityContext = scanText.slice(polarityScanStart, dm.index);
if (/\bNOT\b|\bno\b/i.test(polarityContext)) continue;
```

This is the sole mechanism holding pass-1's CRITICAL fix (the A1/#1969 negation sits 98 chars from its subject, so the 120-char cap does not reach it). It enumerates two spellings and ignores every clause/line boundary the file's own `clauseBounds` machinery exists to enforce — and `clauseBounds` is called two lines later, on the same text.

**Fails open** (a wrong citation silently excused). Measured, `#4242 <phrase> yet.`:

| phrase | donates `#4242`? |
|---|---|
| `is not discharged` / `was NOT discharged` / `is not yet discharged` | blocked ✅ |
| **`is never discharged`** | **FAILS OPEN** |
| **`isn't discharged`** | **FAILS OPEN** |
| **`hasn't been discharged`** | **FAILS OPEN** |
| **`cannot be discharged`** | **FAILS OPEN** |
| **`remains un-discharged`** | **FAILS OPEN** |
| **`far from discharged`** / **`nothing here is discharged`** | **FAILS OPEN** |

7 of 11. Catalogue shape 2 verbatim ("a guard that enumerates syntax — it loses one spelling per round"). Latent today: the register's current negation vocabulary is `not`/`NOT`/`not fully` only (5 instances), all caught. It is hand-written prose, so that is a property of today's text, not of the guard.

**Fails closed** (a legitimate exemption lost → false fatal). The window crosses line and list-item boundaries:

```
- Item 3 is not run yet.
- Item 4 — #4242 DISCHARGED 2026-08-01.
```
→ suppressed. The `not` comes from the **previous list item**. This is exactly the adjacent-list-item leakage that passes 8/9/10 of PR #2630 built four `CLAUSE_BOUNDARY_REGEX` alternatives to stop; the new scan opts out of all of them. The register is dominated by `- **N. Item — DISCHARGED date.**` list bodies (A18's row is literally this shape).

### new-C. The sibling function — the one on Check A's **fatal** path — has no polarity check at all, and this PR's new docstring says it does

**`scripts/check-register-citations.mjs:956-975`** (`idSpecificAnnotationPresent`), claimed at **`:1329-1331`**

The new docstring reads:

> "reuses the exact clause-bounding, **proximity-capping, and polarity-checking** logic `idSpecificAnnotationPresent` uses"

`idSpecificAnnotationPresent` has no polarity logic — the parenthetical that follows lists only `DISCHARGE_ANNOTATION_REGEX + clauseBounds + ID_PROXIMITY_CHARS`, which is the true list. Pass-1 finding 6 asked for this sentence to match the code; it now overclaims in a new direction, and this time about a *different* function.

That matters because the sibling is armed. Measured end-to-end through `checkNonexistentIds` against the real register:

| file body | result |
|---|---|
| `See register row Z9 for the outstanding work.` | **fatal = 1** |
| `Register row Z9 was discharged 2026-08-11.` + same citation | fatal = 0, annotated = 2 |
| **`Register row Z9 was NOT discharged; it is still live.`** + same citation | **fatal = 0, annotated = 2** |

A negated annotation downgrades a genuinely-fatal nonexistent-ID citation — prose that *contradicts* the citation being stale is read as excusing it. `idSpecificAnnotationPresent(text, 'A99')` returns `true` for all of `— NOT discharged`, `is not discharged yet`, `has not been discharged`.

Pre-existing on `main` and latent (no corpus instance today), so it is not a regression this PR caused — but it is catalogue shape 10 ("one instance fixed, the class left armed"), it was surfaced by this PR's own fix work, and this PR ships a docstring asserting the protection exists. Per CLAUDE.md's incidental-findings rule, "it's pre-existing" is a void deferral reason.

### new-E. The release-notes technical entry contains three false claims, one of which inverts this PR's own design decision

**`docs/release-notes-next.md:422`**

1. *"This check previously fired only when the ID existed **nowhere** in the register"* — false, and inverted in both halves. Measured on `origin/main`: `### A1 · … (#1969)` (A1 **exists**) → fatal `wrongId`; `### A18 · … (#9999)` (subject exists nowhere) → non-fatal `unknownSubject`. "ID exists nowhere" is Check **A**, a different check.
2. *"the false alert is eliminated by recognizing the multi-row shape and **accepting the citation when ANY sibling row is currently live for that subject**"* — false, and it describes precisely the blanket rule this PR's own body says it rejected: *"a genuinely unrelated id (A99 …) still fires fatal `wrongId` even though the subject has a live sibling elsewhere — the exemption isn't a blanket 'subject has a live row somewhere' rule."* The code gates on `historicalOwnersOf(subject).has(id)`. The trailing clause then inverts which case is which.
3. *"`npm run check:register-citations` remains a standalone, manually-run script (not currently wired into `test:hooks`, pre-commit, or CI)"* — false. `test:hooks` → `scripts/run-hooks-tests.mjs`, which globs `scripts/tests/*.test.mjs`; that set includes `check-register-citations.test.mjs`, whose CLI tests spawn the real script (`CLI_PATH`, `:30`/`:33`). `test:hooks` is a leg of `verify:fast:scoped` (pre-commit), `verify:fast:branch` (pre-push) **and** `.github/workflows/verify.yml:196`. This checker's own header (`check-register-citations.mjs:14-29`, unchanged by this PR) states this explicitly and at length, and the two now contradict each other in the same diff. The real gap the header names is the absence of a *dedicated* CI step — which is a different claim.

Given a prior draft of this same entry was corrected for fabricated claims, claim 3 is notable: the correction replaced one false statement with its opposite rather than with the header's actual text.

---

## 🟡 New — cleanup

### new-D. Half the finding-1 fix is unpinned: removing the `ID_PROXIMITY_CHARS` cap changes zero test outcomes

**`scripts/check-register-citations.mjs:1380`**

Mutation run (identical copied suite, out-of-repo, per variant):

| variant | pass / fail |
|---|---|
| baseline (this head) | 102 / 8 *(the 8 are the environmental CLI tests, same as pass 1)* |
| **polarity check removed** | **101 / 9** — the new negated-discharge test goes red ✅ |
| **proximity cap removed** | **102 / 8 — byte-identical to baseline** |
| both removed | 101 / 9 |

Yet the cap is load-bearing on the real register: donations go **2 → 7** without it (`A1←#1837`, `A1←#1972`, `A1←#399`, `A1←#2811`, `G1←#2488` all reappear). So the half doing most of the work on the real corpus is the half no test can see — the same mutation-invisibility pass 1 measured as part of finding 1. Repo rule: "Bug fix → ship a regression test that fails before the fix and passes after."

### new-F. The PR body's Test plan is stale and does not declare the incidental fixes

The body still asserts *"`node --test …` — **107 passed, 0 failed** (observed, this run)"*; the suite is now **110**. The body also carries no note of the A34 annotation/heading work, the release-notes addition, or the four review-fix commits — CLAUDE.md requires incidental fixes to be declared in the body ("Also fixed, found in passing: …"), and pass 1 already flagged the A34 change as body-invisible.

---

## Verdict

**Request changes.** All four pass-1 correctness findings are genuinely resolved — each re-verified by re-running the differential rather than by reading the diff, and the A1/#1969 repro is back to `main`'s fatal behaviour.

What blocks is that the fix closed the reported *instance* and left three live edges of the same class: new-A is a wrong exemption sitting on the shipped register right now (with the fixing primitive already in this file), new-B is the enumerate-two-spellings shape holding up the whole CRITICAL fix, and new-C is the identical hole in the sibling function on the **fatal** Check A path — asserted by this PR's own new docstring to be closed. new-E is three false statements in the durable technical record, one of them describing the exact rule the PR deliberately did not implement.

None of these is a regression against `main`, so the tree is not worse than it was; the widening simply does not yet deliver what the PR and its release note say it delivers.

<sub>Verified this pass: 110/110 tests; `--strict` exit 0; 18/18 checks green on `a8e8b37e`; three-way differential (`origin/main` / `7d880abc` / head) on the real register for A1/#1969, A18/#1978, A23/#2107 and an unrelated-id control; mutation of the proximity cap and the polarity check against a copied suite; real-corpus donation census (14→2, 7 without the cap); 11-spelling polarity enumeration probe; end-to-end negated-annotation probe through `checkNonexistentIds`; `test:hooks` wiring traced through `run-hooks-tests.mjs` → `verify-cache.mjs` → `verify.yml:196`. Tree check: `a8e8b37e` clean before and after, nothing modified.</sub>

dudarenok-maker:
## PR review — pass 3 (head 31825d4f18cb04cc2695b9c44360ae576de2ac6d, depth medium)

Re-review, final round under the loop cap. Deltas only — solid items from passes 1–2 are not re-listed.

Re-run independently at this head in the existing `wt-2721-widen-wrongid` worktree (clean before and after, no tracked file modified): `node --test scripts/tests/check-register-citations.test.mjs` → **117 pass / 0 fail**; `node scripts/check-register-citations.mjs --strict` → **exit 0**; `server/src/spawn-windows-hide.test.ts` → **20/20**; `npm run check:onbox-register` → OK; `npx eslint` on both changed scripts → clean; **16/16 checks green** on this SHA. Every mutation and differential below was run against out-of-repo copies of the module (`origin/main`, `a8e8b37e`, this head, and six single-property mutants).

---

## Disposition of the pass-2 findings

|---|---|---|
| new-A | 🟠 harvest donates non-subject numbers (`A18 ← #1978` live) | **VERIFIED RESOLVED** |
| new-B | 🟠 polarity scan too narrow, fails open *and* closed | **Fail-open half RESOLVED. Fail-closed half NOT resolved — inverted, see new-G** |
| new-C | 🟠 sibling function on the fatal path has no polarity check, docstring says it does | **Docstring half RESOLVED. Code half neither fixed nor filed — see new-H** |
| new-D | 🟡 proximity cap mutation-invisible | **VERIFIED RESOLVED** |
| new-E | 🟠 three false release-notes claims | **All three RESOLVED; a fourth false claim introduced in the rewrite — see new-H** |
| new-F | 🟡 stale PR body | **RESOLVED** (test count now 117, "Review fixes" section present); one stale line remains, new-K |

**new-A** — real-corpus census through `dischargedSubjectsMentionedIn`, register unchanged: `a8e8b37e` donated **2** `(id, subject)` pairs, this head donates **0**. Reverting *only* the `prSubjects.has(subject)` exclusion (`:1407`) restores `A18 ← #1978` exactly and nothing else, so the fix is the cause and it is the right cause. The new fixture test goes red on that mutant; its paired no-PR-marker control stays green.

**new-D** — mutation matrix, out-of-repo copied suite, one property reverted per variant (baseline 110 pass / 7 fail; the 7 are the environmental CLI/binary tests that need the real repo root):

| mutant | result |
|---|---|
| `ID_PROXIMITY_CHARS` cap removed | **109 / 8** — the new cap test goes red ✅ (was mutation-invisible at pass 2) |
| PR-number exclusion removed | **109 / 8** — new-A test red ✅ |
| `un-`prefix negation removed | **109 / 8** — the `un-discharged` test red ✅ |
| negation-word set removed | **107 / 10** — three negation tests red ✅ |
| clause-bounding reverted to the flat 30-char window | **109 / 8** — the cross-list-item test red ✅ |

So every pass-2 fix is now pinned by at least one test that actually fails without it, including the cross-list-item test the round self-caught as a no-op — **I re-mutated the clause bound back to a flat window myself and confirmed it goes red**, rather than taking the rewrite on trust. The cap is also load-bearing on the real corpus (donations 0 → 3 without it).

**Apostrophe escaping** — the six `'` escapes match `POSSESSIVE_SUBSECTION_REGEX`'s own convention (`:590`), `regexLiteralDesyncsQuoteTracking` (`spawn-windows-hide.test.ts:112`) sees no unpaired quote, the guard is 20/20, and the escapes match real apostrophes at runtime (`isn't discharged` is blocked in an end-to-end probe). Correct.

**new-E claims 1–3** — re-verified individually. Claim 3's replacement is accurate: `run-hooks-tests.mjs` globs `scripts/tests/*.test.mjs`, `test:hooks` is a step of `verify:fast:scoped` (pre-commit), `verify:fast:branch` (pre-push) and `verify.yml:196`, and nothing invokes `npm run check:register-citations` as its own gate (`package.json:61` is the only reference). Worth noting for later: `docs/testing/**` is one of that step's own input globs (`verify-cache.mjs`), so a register-text edit does move the step's hash — which is what makes new-G below reachable as a hard failure rather than a silent one.

---

## 🟠 New — correctness

### new-G. The clause-bounded polarity scan **inverted** new-B's fail-closed half: an unrelated negation two sentences earlier now suppresses a genuine discharge, and this head produces a false FATAL where `a8e8b37e` was correct

**`scripts/check-register-citations.mjs:1381`** (`const polarityContext = scanText.slice(start, dm.index);`)

The scan moved from a flat 30-char window to the whole clause. `CLAUSE_BOUNDARY_REGEX` (`:860`) does **not** treat a plain sentence-ending `. ` as a boundary — pass 1 measured real clauses at 434/578/668/1308 chars — so "the same clause" is "the same markdown paragraph". Every word in that paragraph before the discharge word now votes on its polarity.

**Register-wide measurement**, this register, unchanged: 31 discharge-word occurrences in row bodies. `a8e8b37e` suppresses **7**; this head suppresses **9**. All three of the newly-suppressed ones are affirmative records, verbatim:

- **A1** — `…are untouched and remain the highest-risk unproven behaviour here. E-03 (the Coqui equivalent of C-01) is now discharged (Run 4, above).` — nearest negation `not reached`, **226 chars** back, different sentence.
- **A23** — `**Superseded (#2107, …) — see the PARTIALLY DISCHARGED banner…` — nearest negation `not the audio bytes already on disk`, **190 chars** back, contrastive *not-X-but-Y*.
- **E6** — `> …the timing/no-flash behaviour, 4 (green ready card), and 5 (refetch without reload) are all **DISCHARGED**` — the "negation" is `\bno\b` matching **inside the hyphenated adjective `no-flash`**, 85 chars back.

On this corpus every *genuine* negation sits 4–10 chars before the discharge word (`— NOT discharged`, `is not fully discharged`); every *false* one sits ≥85. The flat window separated those two populations exactly; the clause bound does not.

**Concrete failure scenario**, three-line fixture, no real files, run through `checkConflictingSubjects`. Register: `### A1 · Re-minted work (#2040)` whose body is one paragraph; `#1969` has no live row.

| A1's body | `origin/main` | `a8e8b37e` | **31825d4f** |
|---|---|---|---|
| `Subject #1969 was discharged 2026-08-01 when this row was re-minted.` | `unknownSubject` | `annotatedDischarge` | `annotatedDischarge` ✅ |
| `The repair rewrote metadata, not the audio bytes already on disk. Subject #1969 was discharged 2026-08-01 when this row was re-minted.` | `unknownSubject` | `annotatedDischarge` | **`wrongId` — FATAL** ❌ |
| `Group Q no longer exists after the 2026-07-01 consolidation, and every row it held was folded into Group A during that same sweep of this register. Subject #1969 was discharged then.` | `unknownSubject` | `annotatedDischarge` | **`wrongId` — FATAL** ❌ |

Citation in both failing rows: `### A1 · Some section (#1969)`. The message printed is `#1969's row has fully discharged and A1 was re-minted for different work` — asserting the row documents nothing, when it documents the discharge in the sentence immediately before.

Row 3 is the sharper one: the suppressing token is `no` inside **`no longer exists`** — one of `DISCHARGE_ANNOTATION_REGEX`'s own three affirmative discharge phrases (`:692`). The guard's negation vocabulary overlaps its own discharge vocabulary, so writing a discharge one way disarms a discharge written the other way in the same paragraph.

**Why the tests can't see it**: 117/117 green with this present. The new cross-list-item test pins only the direction the change fixed; nothing pins the direction it broke. The shape the fix *was* written for (a negation bleeding across a list-item boundary within 30 chars) has **zero** instances on this register; the shape it introduced has **three**.

Not live today — `--strict` is exit 0 and the two live Check-C annotated citations (A19/#1893, A31/#2037) go through the *citation-side* path, not this one. But `docs/testing/**` is a `test:hooks` input and the CLI tests assert `status === 0` against the real tree, so the first register edit written in the register's own existing prose style plus a matching heading citation turns this into a red pre-commit, pre-push **and** CI leg, with a message that is factually wrong.

To be fair to the commit: the same change genuinely **closed** the fail-open half — I re-ran pass 2's 11-spelling probe and all 11 are now blocked (`never`, `isn't`, `hasn't been`, `cannot be`, `remains un-discharged`, `far from`, `nothing here is`). That half is a clear improvement. It is the fail-closed half that went backwards.

Direction, not prescriptive: the clause bound is the right *upper* bound; it needs a distance cap inside it. The corpus separates cleanly at ≤10 vs ≥85, so keeping the 30-char cap **within** `clauseBounds` restores all five correct judgements while keeping the cross-list-item fix. Bare `no` also wants qualifying so it stops matching `no-flash` and `no longer exists`.

### new-H. `docs/release-notes-next.md:422` now claims the **citation-side** annotation path is polarity-checked. It is not — a negated annotation still downgrades a fatal citation on **both** fatal paths

The rewritten entry reads:

> "It now fails fatally, unless a **nearby discharge annotation** excuses it — ordinary markdown prose naming the discharge (`DISCHARGE_ANNOTATION_REGEX`…), clause-bounded, proximity-capped, **and polarity-checked so a negated sentence ("was NOT discharged") can't be misread as one**…"

"A nearby discharge annotation" is `idSpecificAnnotationPresent` (`:956-975`), consumed on Check C's fatal branch at **`:1699`** and on Check A's at **`:1157`**. It has no polarity logic — the docstring at `:1343` was corrected this round to say exactly that ("a polarity check `idSpecificAnnotationPresent` itself does NOT have"), and the release note added the opposite claim in the same commit pair.

**Measured end-to-end at this head**, synthetic register (`A18` tracks `#2000`; `#1978` has no live row):

| file body | Check C |
|---|---|
| `### A18 · Some section (#1978)` alone | `wrongId` = 1 — FATAL |
| + `> **Register row: A18 — discharged 2026-08-11.**` | wrongId 0, annotated 1 |
| + `> **Register row: A18 — was NOT discharged; it is still live.**` | **wrongId 0, annotated 1** |
| + `> Register row A18 is never discharged and remains live.` | **wrongId 0, annotated 1** |

Same on Check A (nonexistent row `Z9`): affirmative, `NOT discharged`, and `is never discharged` all produce `fatal=0, annotated=2`. Prose that *contradicts* the citation being stale excuses it.

This is pass-2's new-C, second half. The chosen disposition was to correct the docstring and leave the code — which is defensible on its face, because arming the same heuristic there is **not** free: I measured it. Adding the identical polarity check to `idSpecificAnnotationPresent` turns **4 of the 23** currently-annotated Check A citations into hard fatal failures on today's tree (Check C's 4 are unaffected). That is a real decision with more than one defensible outcome — i.e. exactly the design-pass carve-out CLAUDE.md allows.

But the carve-out has a condition: the issue is filed and **names the decision owed**. There is none — `gh issue list --search "2846 in:body"` is empty, and neither the PR body nor the release note mentions it. So today the finding is neither fixed nor recorded, and the durable record affirmatively states the protection exists. Two ways out, both cheap: delete `and polarity-checked …` from `:422` and file the issue naming the decision (with the 4-of-23 figure), or arm it and take the four.

---

## 🟡 New — cleanup

### new-I. `21 of 65` is `15 of 66` — wrong in both numbers, in five lines this PR newly writes, one of them operator-facing output

**`docs/testing/onbox-acceptance-register.md:387`** and **`scripts/check-register-citations.mjs:217, 1443, 1630, 2194`** (all five are `+` lines in this PR's diff; three further pre-existing sites at `:116`, `:170`, `:1487` carry the same figure).

Measured three independent ways, all agreeing: `parseRegisterRows` → 66 rows, 15 with an empty `issues` set; `grep -c '^### [A-Z][0-9]'` → 66, of which 15 have no `#\d+` on the heading line; the register's own summary line says **"66 owed."** `:2194` is the success line printed on every run, so the tool tells its operator the permanent coverage gap is 40% larger than it is, against a register size that is off by one.

### new-J. The negation-pattern comment describes three patterns the regex does not implement

**`scripts/check-register-citations.mjs:1374-1380`** — the comment says `"hasn't (been)"`, `"cannot be" / "can't be"` and `"nothing" followed (in the same clause) by "is"`. The regex (`:1386`) matches bare `hasn't`, `cannot`, `can't` and `nothing` with no following-word requirement. Harmless in effect (the regex is broader, i.e. fails closed) but this is the fourth round in which a sentence about this function has not matched the function.

### new-K. PR body's last Test-plan line is stale

> "All 18 CI checks green as of the latest push (`27039fad`/`a8e8b37e` lineage)."

This head is `31825d4f` and reports **16** checks, all green. Everything else in the body is now accurate.

---

## Verdict

**Request changes — but narrowly, and this is the loop cap, so it goes to the user rather than another automatic round.**

The pass-2 fixes are real: new-A, new-D and new-F verified resolved by independent differential and mutation rather than by reading the diff; new-B's fail-open half genuinely closed across all 11 probed spellings; new-C's and new-E's documentation halves genuinely corrected; the apostrophe escaping correct and the `#2747` guard green.

What blocks is that the round repeated the pattern it was fixing. **new-G** is a behavioural regression introduced by the last commit pair, on a fatal path, invisible to a 117-test suite, and measurably worse than the code it replaced on the shipped register (9 spurious-inclusive suppressions vs 7; the 3 new ones all wrong, one of them triggered by the checker's own affirmative discharge phrase). **new-H** is a false statement about a protection this PR's own docstring correctly documents as absent, alongside a design decision that was made but not recorded.

Neither is live: `--strict` is exit 0, 16/16 checks green, and the tree is not worse than `main` in any *currently-reachable* respect. So merging is not dangerous today. It is, however, merging a guard whose polarity logic is now wrong in a way this round measured, with a release note that overstates it — and both fixes are small (a distance cap inside the existing clause bound plus a qualified `no`; one deleted clause plus one issue; three number corrections).

<sub>Verified this pass: 117/117 tests; `--strict` exit 0; 20/20 `spawn-windows-hide`; `check:onbox-register` OK; eslint clean; 16/16 checks on `31825d4f`; six-variant mutation matrix over a copied suite (each pass-2 fix pinned, each mutant's red test named); real-corpus donation census across `origin/main`/`a8e8b37e`/head/five mutants (2 → 0 pairs); register-wide polarity-suppression census (31 discharge words, 7 → 9 suppressed, 3 new ones quoted); three-way `checkConflictingSubjects` differential on both real and synthetic registers; 20-spelling fail-open probe; negated-annotation probe through both fatal paths; 4-of-23 impact measurement for arming polarity on `idSpecificAnnotationPresent`; `test:hooks` wiring re-traced through `run-hooks-tests.mjs` → `verify-cache.mjs` input globs → `verify.yml:196`; row/subject counts cross-checked three ways. Tree check: `31825d4f` clean before and after, nothing modified.</sub>

dudarenok-maker:
## Round-3 fixes landed (head `9b4c2575`) — loop cap now exhausted

Pass 3's findings (comment above) are addressed:

- **new-G (🟠, correctness)** — the pass-2 clause-bounding fix over-corrected: a "clause" per `CLAUSE_BOUNDARY_REGEX` is a whole paragraph, so an unrelated negation 85+ chars back could suppress a genuine discharge, and bare `no` matched inside the affirmative phrase `no longer exists`. Fixed by capping the polarity scan to whichever is closer, clause start or 30 chars back, plus a negative lookahead excluding `no longer exist(s)`. Two new regression tests (distant same-paragraph negation; `no longer exists` not misread as negation), both mutation-verified against the pre-fix code.
- **new-H (🟠, docs overclaim + design gap)** — `idSpecificAnnotationPresent` (Check A's fatal-path annotation matcher) does not have the polarity check `dischargedSubjectsMentionedIn` has. Corrected the release note to stop implying parity, and filed #2858 as the design-pass carve-out (arming polarity-checking there turns 4 of 23 currently-annotated Check A citations fatal — deciding whether those are real defects needs someone to read them, not this PR to decide unilaterally).
- **new-I (🟡, stale facts)** — six "21 [of/-of-] 65" register-coverage mentions (two written by this PR, four pre-existing but surfaced by the same review) corrected to the current measured "15 of 66".
- **new-J (🟡, comment/code drift)** — already resolved as part of rewriting new-G's comment block (the qualifiers the old comment claimed, e.g. "(been)"/"be"/"is", now match what the regex actually implements).
- **new-K (🟡, stale PR body)** — PR body updated with the final review-fixes summary and correct counts/head SHA.

Also self-caught twice more during this round (both by pre-commit, not by review): my first attempt at escaping the new-G regex's apostrophes had a bug (the substitution variable held a literal apostrophe character instead of the six-character `'` escape text), re-breaking `spawn-windows-hide.test.ts`'s quote-tracking guard (#2747) a second time. Fixed and reverified (20/20).

**Verification**: 119/119 `check-register-citations` tests green, `--strict` corpus scan exits 0 clean (same residuals, now correctly reporting "15 of 66"), `spawn-windows-hide.test.ts` 20/20, `check:onbox-register` OK, all 20 CI checks green on `9b4c2575`.

This is pass 3 under this repo's `pr-review-gate` loop cap (initial pass + 2 re-review rounds) — no further automatic re-review round will be dispatched. Every finding from all three passes is either fixed-and-verified in this PR or routed through the design-pass carve-out (#2858). Merge decision now returns to the requesting session/user per the loop-cap protocol.

PR #2856, opened 2026-09-02
chore(ops): drop on-box acceptance register changelog

### LINKED ISSUE #2708: ops-63 — decide whether the on-box register carries a changelog at all
**What**

The on-box acceptance register's live view (`docs/testing/onbox-acceptance-register-live-view.html`) carries a stack of changelog callouts recording each wave's row-count delta, and the register `.md` now carries a parallel changelog section of its own. Decide whether either should exist.

**The decision that is owed**

Two defensible positions, and something has to choose:

1. **Keep and curate.** This is what `main` does today — PR #2688 relabelled the oldest live-view block `Historical, predates wave 7`, added a seventh entry, and the `.md` changelog has grown to 11+ entries. A reader gets the recent history without leaving the page.
2. **Drop, and point at git.** The page's own stated purpose is to track *current state*, and the register's rule is that it "tracks state, not history". Nothing mechanical checks callout prose against the rows, so these go stale silently — which has already happened: six callouts described rows `A44`/`A45`/`A46` and counts of 65→69 while the summary strip said 60, and one live-view callout admitted in its own text that "this callout box was not kept current across those waves".

**Why it is filed rather than fixed**

A branch in flight (#2634/#2653, PR #2707) implemented option 2 and it was dropped before merge: `main` is actively curating that changelog, so deleting it inside a row-ID PR would have been a scope collision with another lane's live work. The cleanup is only correct if the *policy* is settled first.

**Whichever way it goes, the follow-through differs**

- Option 1 needs a mechanical check, or the staleness recurs — the six stale callouts were not caught by any gate.
- Option 2 needs both surfaces done together (live view *and* the `.md` section), plus the live-view footer's "see the .md's own changelog" pointer, which would then dangle.

**Benefit (technical):** closes a recurring source of silently-wrong prose on the one register page an operator actually reads, or else puts a gate under it.

dudarenok-maker:
## Summary

Drops the on-box acceptance register's changelog entirely, per the decision in
[#2708 (comment)](https://github.com/dudarenok-maker/Castwright/issues/2708#issuecomment-5484695277)
(Option 2). Removes the changelog callout blocks and "Historical, predates
wave N" framing from `docs/testing/onbox-acceptance-register-live-view.html`;
`docs/testing/onbox-acceptance-register.md` already carried no changelog
section on this branch. The live-view footer carries no dangling "see the
.md's own changelog" pointer.

Two older docs (`docs/testing/onbox-sitting-qa-gate.md`,
`docs/testing/fs38-wave3-onbox-acceptance.md`) still reference "the register's
... changelog" when recounting historical row discharges from prior waves.
Those are backward-looking historical notes about events that already
happened, not claims that a changelog currently exists in the register — left
as-is rather than edited, per the parent checklist's "or the removal is noted
in the PR body" allowance.

Closes #2708.

## Test plan

- [x] `npm run check:onbox-register` — OK, `onbox-acceptance-register.md` and
      `onbox-acceptance-register-live-view.html` agree (ran against this
      branch in an isolated read-only clone).
- [x] Manual review: no changelog callout blocks or "Historical, predates
      wave N" framing remain anywhere in the live-view HTML; no changelog
      section in the `.md`; footer carries no dangling changelog
      cross-reference.


dudarenok-maker:
## PR review — exempt (head 82403506)

Changed-file set is entirely under `docs/**` (`docs/testing/onbox-acceptance-register-live-view.html`) — the same file-set test CONTRIBUTING.md's "Doc-only PR fast-path" uses. Exempt from the pr-review-gate independent review pass per its Exemption clause; no review dispatched.

Issue #2853, opened 2026-09-01
chore(ops): CI's --changed narrowing can select zero tests on a server-lockfile-only diff


dudarenok-maker:
`.github/workflows/verify.yml`'s `test:server`/`test:server-slow` step body decides narrowed-vs-full only on whether `$BASE` is set, never on the `shared` scope boolean `computeShared` already computes. Confirmed live hazard: a `server/package-lock.json`-only PR is not `shared` scope (only the ROOT lockfile/manifest and `.github/actions/**` are), so it reaches `step_test_server` through `stepTouchedByDiff`'s `includeLockfiles` branch and stays narrowed — and neither `package-lock.json` nor `server/package-lock.json` appears in `server/vitest.config.ts`'s `forceRerunTriggers`, so `--changed "$BASE"` can select zero tests and report green.

Root `package.json` is NOT part of this hazard — already protected by both vitest configs' `forceRerunTriggers` (measured proof already in `server/vitest.config.ts`'s own comment: stripped selects 0, restored selects 5389). The frontend leg is NOT part of this hazard either — it runs `npm run test:a11y` unconditionally regardless of what `--changed` selects, so it always does real work. `test:server-slow` shares `test:server`'s step body, so whatever fix lands applies to both, not as a separate instance.

**Needs a design pass, not a mechanical fix** — a naive `shared`-only check (tried and rejected during this spec's review) doesn't cover the server-lockfile case. The decision this issue is waiting on: what property of a diff actually guarantees `--changed <base>` selects a correct, non-empty test set for a given vitest config, given that config's own `forceRerunTriggers` list? A real answer likely needs to reconcile per-config `forceRerunTriggers` coverage with the CI-side scope decision in `scripts/ci-scope.mjs`, not just patch `verify.yml`'s step-body conditional.

Full history and review trail: docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md, Decision A.

PR #2857, opened 2026-09-02
chore(ops): widen timing margins, shard e2e, add dev-loop scripts

### LINKED ISSUE #2850: chore(ops): widen file-lock.test.ts's thin timing-margin pairs
Three tests in `server/src/workspace/file-lock.test.ts`'s `withKeyLock acquisition timeout (#2260)` `describe` block race a 20ms lock-acquisition-timeout budget against a 150-200ms holder to assert ordering. Widen to an absolute-gap floor of ≥500ms per pair, per `docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md` Decision B. Shipped as cheap, no-regret insurance, not a diagnosed fix — the spec's own review found a real technical argument that these margins may never have been able to flip under contention (Node dispatches timers by expiry, not arrival, and the waiter's timer is armed before the holder's). No production code change; three tests' constants widen, with comments recording the new absolute gap.

Closes nothing on its own; part of the shipping PR for docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md.

### LINKED ISSUE #2851: chore(ops): increase cloud e2e Playwright sharding from 4-way to 8-way
.github/workflows/verify.yml's e2e job shards test:e2e 4-way today. Per docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md Decision D, raise to 8-way (matrix array, the single --shard=N/4 site, job name label) and fix two stale prose comments in the same file that still cite the 4-way figure and a ~110-spec-files estimate (actual: 137). Cost caveats (unconditional per-shard checkout/setup overhead, an unverified concurrent-job-slot ceiling) are documented in the spec and should be watched after shipping, not resolved here.

### LINKED ISSUE #2852: chore(ops): add subsystem-scoped npm scripts for local test-loop convenience
Add 8 opt-in, manual npm scripts (test:server:routes/tts/analyzer/workspace, test:components/store/lib/views) that narrow vitest to one subsystem, per docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md Decision E. Not added to the automated verify pipeline (STEPS[]), not cached, not gated by any hook — pure developer convenience for a fast local loop when working in one area. Document in CLAUDE.md's Commands section.

dudarenok-maker:
## Summary

Three independent, small changes from the same investigation into a 70-80 minute, 3x-failing pre-push run:

- **Widen three thin timing-margin pairs in `file-lock.test.ts`** to an absolute-gap floor of ≥500ms — shipped as cheap, no-regret insurance, not a diagnosed fix (no failure log was captured from the original incident, and review found a real technical argument these margins may never have been flippable by contention; the repo separately documents fork-pool crashes as an independently-sufficient cause). No production code change.
- **Raise cloud e2e Playwright sharding from 4-way to 8-way** in `verify.yml`, plus fixing two stale prose comments that still cited the old 4-way figure and a ~110-spec-files estimate (actual: 137).
- **Add 8 opt-in, manual npm scripts** (`test:server:routes`/`:tts`/`:analyzer`/`:workspace`, `test:components`/`:store`/`:lib`/`:views`) narrowing `vitest` to one subsystem, for a fast local dev loop. Not part of the automated verify pipeline — not in `STEPS[]`, not cached, not gated by any hook.

A fourth, larger idea from the same investigation — extending pre-push's `--changed`-based test narrowing from pre-commit to pre-push — was explored across four rounds of adversarial review and **declined**: its real-world hit rate measured at 0 of the last 60 merged branches, too small to justify the machinery it would have required. See the spec for the full history.

Full design history (5 rounds of spec review) and implementation plan (1 round of plan review + a final whole-branch review that caught and fixed a real process gap):
- `docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md`
- `docs/superpowers/plans/2026-09-02-test-suite-hardening.md`

Also fixed, found in passing during shipping: a pre-existing, repo-wide `browserslist` audit vulnerability was blocking every PR's merge (unrelated to this branch) — surfaced and resolved by merging the already-open fix (#2854) first, then rebasing this branch on top.

Closes #2850
Closes #2851
Closes #2852

## Test plan

- [x] Each commit's own pre-commit hook ran and passed (widened test file: 17/17 green; e2e sharding: out of test scope, fast; npm scripts: verified via `vitest list` file-count checks matching spec exactly; release notes: encoding-verified)
- [x] Final whole-branch review (Opus) — 1 Critical (a fix-round amend that hadn't actually landed, caught and corrected), 2 Important (missing issue refs, Decision D's timing condition — this PR body), 3 Minor (2 fixed, 1 parked as genuinely optional) — all addressed
- [x] Local `verify:fast:branch` — green (lint, typecheck, config:check, test:hooks, test, test:server, build, test:sidecar, audit, audit:server)
- [ ] Cloud `verify.yml` (automatic, required)
- [x] Decision D's own acceptance condition: manual, unscoped `verify.yml` dispatch (run 33631470487) confirmed the 8-way result (slowest shard ~358s) is a real 31–36% improvement vs prior actual 4-way runs (slowest shards ~516s and ~559s, runs 33496342069 and 33572155392), with no queuing observed. Dispatch itself reported `failure` status due to unrelated `Install ffmpeg` apt-hang issue (#2449), not this PR's scope.
- [ ] Mandatory `pr-review-gate` pass (multi-scope PR → high review depth)

## Before-shipping checklist notes

**Step 1 (regression plan):** N/A — small, localized items skip the plan doc; the issue bodies (#2850, #2851, #2852) plus paired tests are the spec.

**Step 3 (on-box acceptance register):** N/A — no behaviour in this PR requires real hardware to prove; it covers test-timing margins, CI shard count, npm script wrappers, and doc updates.

**Step 4 (docs/features/INDEX.md):** N/A — no new plan doc under `docs/features/` requiring an INDEX entry (per Step 1 above; ops/test-infra work uses `docs/superpowers/specs/` and `docs/superpowers/plans/`).


dudarenok-maker:
## Decision D acceptance check — 8-way e2e shard timing (manual unscoped dispatch, run [33631470487](https://github.com/dudarenok-maker/Castwright/actions/runs/33631470487))

**Queuing:** No queuing — all 8 shards started within the same second (12:43:32–12:43:33 UTC). The concurrent-job-slot risk the spec flagged didn't materialize on this run.

**Per-shard durations:**

| Shard | Duration |
|---|---|
| 4/8 | 54s |
| 5/8 | 110s |
| 7/8 | 168s |
| 1/8 | 251s |
| 8/8 | 274s |
| 6/8 | 301s |
| 3/8 | 309s |
| 2/8 | 358s |

Average: **228s (~3.8 min)**. Slowest shard (wall-clock for the whole leg, since all run in parallel): **358s (~6 min)**.

**Honest conclusion: this does not clearly show the win the change was made for.** The workflow's own comment cites the 4-way baseline at "~4 min/shard." This run's *average* (3.8 min) is barely under that, and the *slowest* shard — which is what actually determines the leg's wall-clock — is **6 min, slower than the stated 4-way baseline**, not faster. This is consistent with the spec's own predicted risk: fixed per-shard cost (checkout, setup, the `warmup` project, a fresh Vite dev server per shard) doesn't shrink with more shards, and per-shard test-file count is uneven (54s to 358s — a 6.6x spread across shards, not evenly split), so some shards do far less work than others while all still pay the same fixed overhead.

Not treating this as a blocker — the spec explicitly scoped "confirm 8 is a reasonable number" as a tuning call requiring a real measurement, and this is that measurement. But it's a real, negative-leaning data point that should inform whether 8-way is actually the right number, not just whether it "works." Worth a follow-up look at shard balance (Playwright's `--shard` splits by test count, not file count, so an uneven split suggests either very uneven per-test runtime or a config issue) rather than assuming the current number is final.

dudarenok-maker:
## PR review — pass 1 (head 75e7e7c3a7894410260242cbb26c64c2f16b1e61, depth high)

Scope: the full 6-file branch diff vs `origin/main` (`eda7d866`) — `.github/workflows/verify.yml`, `CLAUDE.md`, `RELEASE_NOTES.md`, `docs/release-notes-next.md`, `package.json`, `server/src/workspace/file-lock.test.ts` — plus the spec (Decisions B/D/E), the plan, the linked issues, `main`'s ruleset, and the workflow's guard tests.

Verified independently before probing, not taken from the PR body: `server/src/workspace/file-lock.test.ts` **17/17 green** under `--retry=0`; `npm run test:hooks` **1757/1757 green** (this is the battery holding `workflow-wiring.test.mjs`, the guard that reads `verify.yml`); `npm run test:server:workspace` → **57 files / 766 tests green**; `npm run test:views` → **29 files / 832 tests green**; `vitest list --filesOnly src/routes` (server) → **138 files**; `vitest list` test-case counts recomputed from scratch → server **6309/8201 = 76.9%**, frontend **4249/4923 = 86.3%** (both match `CLAUDE.md:341` exactly); `find e2e -type f -name '*.spec.ts'` → **137** (the two `e2e/{linux,win32}/responsive/visual.spec.ts` entries are snapshot *directories*, not files — a naive `find` reports 139); raw `server/src/routes/**/*.test.ts` → **146**, and `vitest.config.slow.ts`'s `SLOW_FILES` contains exactly **8** `src/routes/` entries, so 146−8=138 checks out. Working tree `75e7e7c3`, `git status --porcelain` empty, before and after.

### 🟠 Significant — `RELEASE_NOTES.md:220` asserts a fix that this PR's own technical register explicitly disclaims, and credits cloud sharding with curing a local symptom it cannot reach

`RELEASE_NOTES.md:220` leads with: *"A stuck test run on a busy machine **can no longer** eat an hour of your time for reasons that have nothing to do with what you changed."*

`docs/release-notes-next.md:564`, added by the same PR, says the opposite: *"this doesn't claim to have diagnosed the incident … shipped as cheap insurance rather than a proven fix — later review found a real argument that Node's timer-dispatch-by-expiry-time behavior may mean these margins were never actually flippable by contention."* The spec agrees (`2026-09-01-verify-scope-branch-timing-design.md:172-196`: *"not claimed to be confirmed fragile at all"*, *"this section can't point to a concrete failure mode the change closes"*).

Concrete failure scenario, i.e. the specific wrong output: the two halves of the release-notes pair CLAUDE.md step 5 mandates now say contradictory things about the same change, and the half that ships to users (this is the file `#/release-notes` renders) states the stronger, unsupported one. Worse, the technical register names fork-pool crashes retrying a step 3× as *the independently-sufficient explanation for the actual 70–80-minute incident* — and nothing in this PR touches that. So the leading candidate cause is untouched, and the user-facing note nonetheless promises the symptom "can no longer" occur. If it recurs next week, the release note is the artifact that was wrong.

Second, smaller error in the same bullet: *"the checks that run before your code ships now split across more machines in parallel — both aimed at keeping a red result meaningful instead of a symptom of a busy computer."* The sharding change is on cloud `verify.yml`'s `e2e` job. The incident being described is a **local pre-push** run, and `verify:fast:branch` does not run `test:e2e` at all. Splitting cloud e2e across 8 GitHub runners does nothing for a busy local box, so the bullet attributes to Decision D a benefit it structurally cannot deliver for the problem the same sentence describes.

Labelled 🟠 rather than 🟡 deliberately: this is not a reuse/simplification nit, it is a substantive factual error in a shipped artifact, of exactly the shape the reviewer brief calls out as *"the defect is in the instrument or the document, not the code."* No runtime behaviour is wrong, so triage may reasonably re-rate it.

### 🟡 Minor — Decision D's acceptance measurement is recorded only in an ephemeral PR comment, and the file it was measured for no longer carries the figure it displaced

Cleanup/chore, not a correctness bug.

`verify.yml`'s old comment carried a concrete number: *"same anti-flake guarantee, ~4x less wall-clock"* against the stated `~16 min` unsharded baseline — i.e. ~4 min/shard at 4-way. The replacements at `.github/workflows/verify.yml:38-42` and `:403-406` drop the number and say only *"less wall-clock (sub-linear …)"*, then instruct the reader to *"watch actual PR wall-clock after this ships."*

That measurement has already been taken. Run [33631470487](https://github.com/dudarenok-maker/Castwright/actions/runs/33631470487), posted honestly in this PR's own comment: slowest shard **358s**, i.e. the leg's wall-clock, against the **240s** figure the file itself previously stated for 4-way. Credit where due — that comment is candid and even leans negative, and I checked specifically for spin and found none. The gap is durability: the only record that 8-way measured *worse than the number this very file used to print* is a PR comment. `verify.yml` will outlive the reader's memory of PR #2857.

Concrete consequence: a maintainer later asking "should we go to 16?" opens `verify.yml`, reads "less wall-clock, sub-linear", finds no counter-evidence and no pointer to the run, and reasons from a claim the repo has already measured against. Cheapest fix: one clause in the job comment naming the measured slowest-shard figure and the run id (the file already cites the spec by path, so the precedent exists). Note the "sub-linear win" claim is *defensible as written* — it reads against the ~16 min unsharded baseline, and 6 min < 16 min — which is precisely why it needs the 4-way figure beside it to not mislead.

### 🟡 Minor — `CLAUDE.md:341` ships three precise, unguarded figures with no "as of" and no guard

Cleanup/chore.

The bullet states `138 files`, `76.9%`, `86.3%` as bare present-tense facts. All three are correct today — I recomputed every one. All three go stale on the next test file added under any of the eight directories, silently, with nothing to catch it.

Concrete failure scenario: someone adds `server/src/routes/foo.test.ts`. `CLAUDE.md` now says `test:server:routes` selects 138 files "not the 146 a raw file search would find — 8 are in `SLOW_FILES`". A reader checking that arithmetic gets 147−8=139≠138, concludes files are being wrongly excluded, and goes looking for a bug in the exclusion list that does not exist. This is the citation-rot class this repo has been bitten by before (`check:register-citations` exists for exactly this reason on a different surface).

Worth separating from a decision already settled: the spec (`Testing`, line 416) deliberately declined an automated test for Decision E's *scripts*, on the grounds that they carry no correctness guarantee to pin. That reasoning does not extend to the *numbers*, which are new in this PR and were not what the spec waived. Minimal fix is a date/approximation qualifier ("as of 2026-09-02", or `~`, matching how `CLAUDE.md` already writes `~6700` and `~70 files` elsewhere); a `scripts/tests/` guard asserting each of the eight scripts still resolves to a non-empty set is the sturdier option and has direct precedent in `dev-mock-command.test.mjs`.

### 🟡 Minor — before-shipping steps 1, 3 and 4 are silently omitted from the PR body rather than explicitly marked N/A

Cleanup/chore. CLAUDE.md's checklist preamble: *"Skipping a step is fine when the step genuinely does not apply … but say so explicitly rather than silently omitting."* The body covers steps 2, 5, 6, 7 and 10 but says nothing about the regression plan (1), on-box acceptance (3), or `docs/features/INDEX.md` (4). All three are genuinely N/A here — I checked: no hardware-provable behaviour ships, so no register row is owed; the design docs live under `docs/superpowers/`, which `INDEX.md` does not index. One line saying so closes it.

### ✅ What is solid

- **Decision B is correct and the widening does not blunt the tests.** All three pairs land ≥500ms of absolute gap (700/50, 750/50, 700/50). The assertions are on rejection and on `order` equality, not on timing precision, so a wider gap cannot make them vacuous. Test 1's second pair — the 2000ms budget whose headroom the change cuts from ~1870ms to ~1350ms — is arithmetically confirmed and is *documented in the code at the site* rather than left implicit. 17/17 green under `--retry=0`, so the `{ retry: 0 }` discipline this file exists to protect is intact.
- **The fourth candidate pair was correctly excluded, and I checked rather than assuming.** `'does not fire for legitimate contention comfortably under budget'` (file-lock.test.ts:97-111) is a 20ms hold vs a 500ms budget, a 480ms gap — nominally 20ms under this PR's own floor. The spec named it and excluded it explicitly (line 221-222). Not a finding; recording it so the "one instance fixed, class left armed" shape is on the record as having been probed. Same for the two never-releasing-holder tests (lines 82, 238) — infinite gap, nothing to jitter past, and the typed-error test asserts the literal `20` in three places.
- **No required-status-check breakage from the job rename.** `main`'s ruleset requires the contexts `npm run verify` and `Verify PR body links a GitHub issue`, not per-shard names. The `verify` aggregator's `needs:` list includes `e2e`, so all 8 matrix legs still gate it. Renaming `shard N/4` → `N/8` is inert to the ruleset. `test:hooks` (which contains `workflow-wiring.test.mjs`, the guard that parses `verify.yml`'s `# leg:` annotations and step→scope map) passes 1757/1757 — that guard keys on step names, which are unchanged.
- **Decision E's eight scripts are exactly the spec's, and they actually work end-to-end through the double-npm hop.** `test:server:workspace` → 57/766, `test:views` → 29/832, both matching the spec's own `vitest list` figures. Server-relative filter paths (`src/routes`, not `server/src/routes`) are correct — I confirmed the failure mode the spec's round-4 review caught: a non-matching filter exits **1** with "No test files found", so a typo fails loudly rather than passing green having run nothing. The `pretest` ffmpeg preflight still fires, because these invoke server's `test` by name (this repo has been bitten by npm's exact-name pre-hook keying before — `predev:mock`, #2348 — and the shape here is safe).
- **Issue linkage and labelling are real, not assumed.** `Closes #2850` / `#2851` / `#2852` are plain text outside code spans; all three exist, are OPEN, and carry `type:chore` + `area:ops` — so no `docs/BACKLOG.md` row is owed, correctly. Decision A's design-pass issue was genuinely filed (**#2853**, open, and its body names the decision owed rather than just restating the finding).
- **Spec-claim spot-checks all held.** Every number I could recompute matched: 138/146/8, 76.9%, 86.3%, 137 spec files, 57/766, 29/832. The two `~110` → `137` corrections in `verify.yml` fix a genuinely stale figure. No mirror of `CLAUDE.md`'s Commands list exists to go stale (`AGENTS.md` and `.clinerules/cline.md` both deliberately point at it rather than duplicate it — verified).
- **Release-notes entries are in the right sections.** `RELEASE_NOTES.md` under the in-progress `# Castwright 1.15.0`; `docs/release-notes-next.md` inside the `release-notes-next-version: 1.15.0` body, above the `**Full changelog:**` footer. The missing `PR #2857` ref is *not* a finding — roughly a third of the sibling entries in the same section omit it, so there is no convention to violate.

### Verdict

**One 🟠 and three 🟡. No 🔴.** The engineering is clean and the numbers are honest — every quantitative claim in the diff survived independent recomputation, which is unusual and worth saying. The one substantive defect is not in the code at all: the user-facing release note promises a fix the PR is otherwise scrupulously careful never to claim, and hangs a cloud-CI change on a local-machine symptom. That contradiction inside a mandated document pair is the thing to fix before merge; the three 🟡s are durability chores on the same theme — a measurement, a set of figures, and three checklist steps that exist somewhere but not where a later reader will look.

Tree unmodified throughout: `75e7e7c3a7894410260242cbb26c64c2f16b1e61`, `git status --porcelain` empty at both ends.

dudarenok-maker:
## PR review — pass 2 (head a2997581ca3383ab22df6911f2b2d2dc56292322, depth high)

Re-review of fix commit `a2997581` against pass 1's four findings. Depth re-derived from the current commit set: still multi-scope (`ci` + `chore` + `test` + `docs`), so still **high**.

Each prior finding's disposition was checked against the file, not against the fix commit's own message. Tree `a2997581`, `git status --porcelain` empty at both ends.

---

**🟠 `RELEASE_NOTES.md:220` overclaim → ✅ VERIFIED RESOLVED (both halves).**

- *"can no longer"* is gone. The bullet now reads *"a safeguard against flakes that may occur under heavy load"* — hedged, and consistent with `docs/release-notes-next.md:564`'s *"cheap insurance rather than a proven fix."* The failure scenario pass 1 named (a recurrence next week making the note the wrong artifact) is closed: the note no longer promises the symptom cannot recur.
- The cloud/local conflation is gone. *"In the cloud, e2e tests now run across more parallel shards"* correctly scopes Decision D to `verify.yml` and stops crediting it with curing a local pre-push symptom `verify:fast:branch` never touches.
- No new *factual* error in this bullet. *"the wall-clock improvement is sub-linear"* is true against real data (see below): 2× the shards bought ~1.6–1.8× on average and ~1.4–1.6× on the slowest shard.

One new 🟡 (voice, not fact) is filed below.

**🟡 Minor #1 — Decision D's measurement recorded in `verify.yml` → ⚠️ ADDRESSED IN FORM, BUT THE RECORDED FIGURE IS WRONG.** A measurement is now in the file, which is what the finding asked for. But the comparison it records is false and points the opposite way from the truth. Filed as a new 🟠 below — this is the one item that keeps the loop open.

**🟡 Minor #2 — `CLAUDE.md:341` unguarded figures → ✅ VERIFIED RESOLVED.** The bullet now ends *"(as of 2026-09-02; these figures will drift as test files are added or removed)"*. Date qualifier present, accurate, and `figures` (plural) reaches back over all three numbers, not just the `86.3%` it sits beside. All three still recompute correctly today. The caveat sits at the tail of a two-clause sentence rather than adjacent to the `138 / 146 / 8` arithmetic the finding's scenario centred on — weaker than ideal placement, but the arithmetic *relationship* (146 − 8 = 138) is the stable part and the drift warning covers the operands. Adequate; not reopening it.

**🟡 Minor #3 — checklist steps 1/3/4 marked N/A → ⚠️ PARTIALLY RESOLVED.** A `## Before-shipping checklist notes` section now exists and covers all three steps, which closes the "silently omitted" half. Step 3's reason is accurate (I re-confirmed no hardware-provable behaviour ships). Steps 1 and 4 reach the right conclusion via a reason the repo's own docs contradict — filed as a new 🟡 below.

---

### 🟠 NEW — `verify.yml` now records 8-way sharding as a regression, against a `~240s` "4-way baseline" that no run ever produced

**Correctness bug (in the document, not the code).** `.github/workflows/verify.yml:43-44` and `:407-408`.

Both comments now say the same thing:

> `slowest shard ~358s (worse than old 4-way baseline of ~240s), no queuing.`

The `~358s` and `~228s` figures are correct — I recomputed both from run [33631470487](https://github.com/dudarenok-maker/Castwright/actions/runs/33631470487)'s job timings and they match exactly, the run really was an unscoped `workflow_dispatch` on this branch, every one of the 8 shards really did execute its `E2E (chromium)` step, and all 8 started within 1 second of each other, so `no queuing` is verified too.

**The `~240s` comparator is not a measurement.** It is `~16 min ÷ 4` — a back-derivation from the *old comment's own* `~4x less wall-clock` claim, which was itself hung on a `~16 min` unsharded figure measured on 2026-07-10 when the battery was `~110` spec files. It is now 137, ~25% larger — a staleness **this same PR corrected in that very sentence** (commit `3db282ab` changed `~110` → `137`) while leaving `~16 min` bolted to it. So the divisor, the dividend and the file count are all from different measurements.

**What 4-way actually measured.** The two most recent unscoped dispatches before the 8-way change, both of which ran the full e2e battery:

| run | shards | per-shard wall-clock (s) | avg | **slowest** |
|---|---|---|---|---|
| [33631470487](https://github.com/dudarenok-maker/Castwright/actions/runs/33631470487) (this branch) | 8 | 251, 358, 309, 54, 110, 301, 168, 274 | 228 | **358** |
| [33572155392](https://github.com/dudarenok-maker/Castwright/actions/runs/33572155392) | 4 | 559, 324, 346, 399 | 407 | **559** |
| [33496342069](https://github.com/dudarenok-maker/Castwright/actions/runs/33496342069) | 4 | 516, 291, 307, 354 | 367 | **516** |

Slowest shard *is* the leg's wall-clock, which is the number the whole change exists to move. 8-way came in at **358s against a real 4-way 516–559s — a 31–36% improvement**, and average shard time fell from 367–407s to 228s. The change worked. The file now records it as having failed.

**Concrete failure scenario.** This is precisely the reader pass 1 posited: a maintainer asks "was 8-way worth it, and should we go further?", opens `verify.yml`, and reads that the measured outcome was *worse than the prior baseline*. The rational response is to revert to 4-way — which would add ~160–200s to the e2e leg on every PR — on the strength of a number no CI run has ever produced. The previous comment said "watch actual PR wall-clock after this ships" and was merely uninformative; this one is actively misleading, and it is the shape the brief calls out as *"a figure measured under one rule reused as evidence for another."*

**Corroborating signal inside this same PR:** `RELEASE_NOTES.md:220` now tells users the sharding is an improvement (sub-linear), while `verify.yml` tells maintainers it was a regression. They cannot both be right. The release note is the accurate one.

**In fairness, pass 1 seeded this.** My predecessor wrote that 8-way "measured *worse than the number this very file used to print*" and framed `~240s` as the 4-way figure — comparing a measurement against a projection without flagging it as one. The fix author faithfully hardened that framing into the word "baseline". The finding stands regardless, but the fix commit is not where the error originated.

**Minimal fix:** replace the parenthetical with the measured comparison and cite the baseline runs — e.g. *"slowest shard ~358s vs ~516–559s measured at 4-way (runs 33496342069 / 33572155392); average 228s vs 367–407s; no queuing"* — and either drop `~16 min` or re-attach it to the `~110`-file battery it was measured on.

---

### 🟡 NEW — `RELEASE_NOTES.md:220` now carries internal jargon that appears nowhere else in the file

**Cleanup nit.** The corrected bullet buys its accuracy with vocabulary the user-facing document does not otherwise use: `e2e`, `shards`, `sub-linear`, `fixed setup cost`. I grepped the whole file, every version section: **line 220 is the only occurrence of `e2e`, `shard`, or `sub-linear` in `RELEASE_NOTES.md`.** The fifteen sibling bullets in this same 1.15.0 section consistently translate internals into plain terms — *"the outside code Castwright is built on"*, *"the speech service"*, *"your graphics card"* — including ones describing far more technical work (the AMD/ROCm bullet three lines up).

The accuracy trade wasn't forced: *"Castwright's cloud checks now spread the browser tests across more machines, though not in direct proportion — each machine repeats some fixed setup"* keeps every hedge the current sentence earns. Not reopening the 🟠 this replaced; the fix was right, only its register slipped.

---

### 🟡 NEW — the PR body's step 1/4 N/A reason rests on a premise `docs/features/INDEX.md` contradicts

**Cleanup nit.** The new section says:

> **Step 1 (regression plan):** N/A — this PR's design and planning docs live under `docs/superpowers/specs/` and `docs/superpowers/plans/` rather than `docs/features/`, following the existing convention for ops/test-infra work.

Two problems, both checkable:

1. `docs/features/INDEX.md:28-39` states that `docs/superpowers/plans/` and `specs/` *"hold the design-thread artifacts … **not the regression plans above**"*. So living in `docs/superpowers/` is not a reason a regression plan isn't owed — the two are different artifacts, not substitutes.
2. The cited *"existing convention for ops/test-infra work"* has a direct counter-example in the same index: `INDEX.md:407`, **"Flaky-test release hardening"** — test-infra work, backed by a `docs/superpowers/plans/` plan, and carrying an `INDEX.md` row anyway. Which also undercuts step 4's *"no new or moved plan under `docs/features/`"*, since eight index rows link primarily at `../superpowers/specs/` targets.

**Concrete consequence:** a later audit of whether this PR discharged step 1 reads the stated reason, checks `INDEX.md`, finds superpowers explicitly excluded from being regression plans, and concludes a plan is still owed and an index row is missing.

The conclusion (N/A) is correct — the unimpeachable reason is sitting in CLAUDE.md step 1 itself: *"Small/localized items skip the plan doc — the issue body + paired test is the spec."* Three tiny changes across three issues is exactly that. Swapping the reason costs one sentence and removes the contradiction.

---

### 🟡 NEW — the PR body's Decision-D acceptance checkbox is stale, and the same commit is what staled it

**Cleanup nit.** The test plan still carries, unticked and in the future tense:

> - [ ] Decision D's own acceptance condition: a manual, unscoped `verify.yml` dispatch on this branch … — will update this PR with the result

That dispatch has run. Run 33631470487 is a `workflow_dispatch` on `chore/ops-test-suite-hardening` at `75e7e7c3`, unscoped, all 8 shards executed, per-shard timing captured, and all 8 started within 1s of each other — so both halves of the stated condition (per-shard timing, no queuing behind the concurrent-job ceiling) are satisfied. Commit `a2997581` then wrote that result into `verify.yml`. The box that says the measurement is still pending is contradicted by the commit that recorded it.

**Concrete consequence:** whoever merges reads the body, believes a declared acceptance condition is outstanding, and either holds the PR or re-dispatches a redundant 16-job run.

One caveat to record when ticking it: that dispatch's **overall conclusion was `failure`**, but not from anything in scope here — the sole failing job was `E2E visual baselines (chromium)`, and it failed on its `Install ffmpeg` step, i.e. the known apt-hang class (#2449), not on a snapshot. Every other leg, including all 8 e2e shards, was green. Worth one clause so a later reader seeing a red run doesn't misread it as this branch's doing.

---

### Re-verified, no delta

- `npm run test:hooks` — **1757/1757 green**, so `workflow-wiring.test.mjs` still parses `verify.yml` cleanly after the comment restructure (the header's parenthesis and sentence boundaries moved; the `# leg:` annotations and every step name did not).
- `CLAUDE.md:341`'s three figures still recompute correctly.
- `docs/release-notes-next.md`'s entry was untouched and needed to be — it makes no timing-outcome claim, so it carries none of the `~240s` error.

### Verdict

**One 🟠 and three 🟡. No 🔴.** Three of pass 1's four findings are genuinely resolved, and the 🟠 that mattered most — the user-facing overclaim — is properly fixed, including the cloud/local conflation.

The remaining 🟠 is that the one finding fixed by *adding a number* added the wrong one, and inverted its own conclusion: `verify.yml` now records a measured regression where the real measurement is a 31–36% improvement. That is a stronger error than the omission it replaced, because an absent figure invites a measurement while a wrong figure closes the question. It is also cheap to fix — the correcting data is two `gh api` calls away and is tabulated above.

Tree unmodified throughout: `a2997581ca3383ab22df6911f2b2d2dc56292322`, `git status --porcelain` empty at both ends.

dudarenok-maker:
## PR review — pass 3 (head dfcdbb0ef8f89ca08dbaec97db5f731d5d57ed0d, depth high)

Final round under the 3-pass cap. Re-review of fix commit `dfcdbb0e` against pass 2's four findings, plus a fresh sweep of the whole branch diff vs `origin/main` (`eda7d866`). Depth re-derived from the current commit set: still multi-scope (`ci` + `chore` + `test` + `docs`), so still **high**.

**Every figure below was re-measured from the GitHub API in this pass, not read off the fix commit's message.** Pass 2's finding was itself caused by a number copied without re-verification, so taking the correction on faith would have reproduced the exact defect class. Tree `dfcdbb0e`, `git status --porcelain` empty at both ends.

---

**🟠 `verify.yml` recorded 8-way as a regression against a never-measured `~240s` "4-way baseline" → ✅ VERIFIED RESOLVED.**

Independently re-derived from `gh api repos/dudarenok-maker/Castwright/actions/runs/<id>/jobs`:

| run | event / branch | shards | per-shard wall-clock (s) | avg | **slowest** |
|---|---|---|---|---|---|
| [33631470487](https://github.com/dudarenok-maker/Castwright/actions/runs/33631470487) | `workflow_dispatch` · `chore/ops-test-suite-hardening` | 8 | 251, 358, 309, 54, 110, 301, 168, 274 | 228.1 | **358** |
| [33572155392](https://github.com/dudarenok-maker/Castwright/actions/runs/33572155392) | `workflow_dispatch` · `chore/deps-browserslist-audit-2847` | 4 | 559, 324, 346, 399 | 407 | **559** |
| [33496342069](https://github.com/dudarenok-maker/Castwright/actions/runs/33496342069) | `workflow_dispatch` · `chore/sidecar-kokoro-cuda-path` | 4 | 516, 291, 307, 354 | 367 | **516** |

Checked, item by item, against what the file now says:

- The back-derived `~240s` is gone from both sites. `.github/workflows/verify.yml:42-45` and `:409-411` now cite the two real 4-way runs.
- **Run-to-figure pairing is correct in both comments, and they are written in opposite orders**, so this was worth checking rather than assuming. Line 43: *"runs 33572155392 and 33496342069 … slowest shards ~559s and ~516s respectively"* → 33572155392=559 ✅, 33496342069=516 ✅. Line 410-411: *"~516–559s (runs 33496342069 and 33572155392)"* → range order matches run order ✅.
- **`31–36%` is arithmetically right**: (559−358)/559 = 35.96%, (516−358)/516 = 30.62%.
- **`~228s/shard` average** recomputes to 228.125 ✅. **`slowest shard ~358s`** ✅.
- **The framing is now correct in direction**: the change is recorded as an improvement, which is what the data says. The `sub-linear` claim beside it is *also* still true and not in tension — 2× the shards bought 1.44–1.56× on the slowest shard, and the comment explicitly says wall-clock *"is set by the slowest shard"*, which is the right metric.
- **`no queuing` re-verified independently**: all 8 shard jobs were `created_at` 12:43:30Z and `started_at` 12:43:32–12:43:33Z. No shard waited on a runner slot.
- **The visual-baselines leg can't invalidate the comparison** — I checked, since it runs concurrently and would set workflow wall-clock if it were the long pole. It took 220s at 8-way vs 108s/93s at 4-way, so `max(358, 220)` vs `max(559, 108)` leaves the conclusion unchanged.

The pass-2 contradiction between `verify.yml` (regression) and `RELEASE_NOTES.md` (improvement) is closed; both now say improvement, and improvement is correct.

**🟡 `RELEASE_NOTES.md:220` jargon → ✅ RESOLVED, and accurate.** `e2e`, `shards` and `sub-linear` are gone; the bullet now reads *"end-to-end tests now run across more parallel machines … but adding more machines doesn't double the speed, since each one has its own setup costs."* Checked for accuracy rather than for jargon removal, since the simplification was a judgment call: doubling shards 4→8 delivered 1.44–1.56×, so *"doesn't double the speed"* is literally true, and per-shard fixed cost is a real contributing cause. Consistent with `docs/release-notes-next.md:564`, which makes no timing-outcome claim and needed no edit.

**🟡 PR body Decision-D checkbox stale → ✅ RESOLVED.** Now `- [x]`, past tense (*"confirmed"*), with the correct run ids, correct pairing (`~516s and ~559s, runs 33496342069 and 33572155392` ✅), and an accurate result. The failure caveat is exactly right and I verified it rather than accepting it: run 33631470487's only failing job was `E2E visual baselines (chromium)`, failing at its `Install ffmpeg` step — the #2449 apt-hang class — with the `npm run verify` aggregator failing only downstream of it at `Check leg results`. All 8 e2e shards were green.

**🟡 PR body step 1/4 N/A reason contradicted by `docs/features/INDEX.md` → ❌ NOT ADDRESSED.** The text is verbatim unchanged, and both counter-facts still hold today:

- `INDEX.md:28-39` — *"These directories hold the design-thread artifacts … **not the regression plans above**"*, so living under `docs/superpowers/` is not itself a reason a regression plan isn't owed.
- `INDEX.md:407` — **"Flaky-test release hardening"**, test-infra work backed by a `docs/superpowers/plans/` plan, carries an `INDEX.md` row anyway, which is a direct counter-example to the *"existing convention for ops/test-infra work"* the body cites and also undercuts step 4's stated reason.

The **conclusion** (N/A) remains correct — the unimpeachable reason is CLAUDE.md step 1's own *"Small/localized items skip the plan doc — the issue body + paired test is the spec."* Still a one-sentence swap. Carried forward unchanged; still 🟡, still cleanup-only.

---

### 🟡 NEW — this PR's `~110` → `137` edit orphaned the `~16 min` figure from the battery it was measured on

**Cleanup / chore — a doc sentence this change made false.** `.github/workflows/verify.yml:34-37`.

> `# E2E ITSELF IS SHARDED (2026-07-10, raised to 8-way 2026-09): a live timing`
> `# comparison showed the functional e2e battery (137 spec files, CI-pinned to`
> `# workers=1 …) was the dominant leg`
> `# at ~16 min, dwarfing every other job.`

Provenance traced rather than assumed — `git log -S` on `verify.yml` puts **both** `~16 min` and `~110 spec files` in the same commit, `29d0dc3e` (2026-07-10, the original 4-way sharding change). They were one measurement. This PR's commit `3db282ab` updated the file count to today's `137` and left `~16 min` bolted to it, so the sentence now attributes a 2026-07-10 measurement to a battery ~25% larger than the one it was taken on.

Pass 2 named this in its minimal fix (*"either drop `~16 min` or re-attach it to the `~110`-file battery it was measured on"*); `dfcdbb0e` did neither. Cheapest fix is one clause — `at ~16 min (measured 2026-07-10, when the battery was ~110 files)` — which restores the provenance without touching the present-tense count.

**Deliberately 🟡, not 🟠, and the reasoning matters for the cap.** Unlike the `~240s` figure it sits next to, this one inverts nothing: the qualitative conclusion (e2e was the dominant leg; the sharding win is sub-linear) is unchanged, and the error runs *conservative* — a 137-file unsharded battery would take longer than 16 min, which makes the sharding win look better, not worse. There is no reader scenario where this produces a wrong decision, so it does not meet the correctness-bug bar and does not trip the re-review trigger.

---

### Probed and cleared — recorded so the negatives are on the record

- **Shard-count sites all moved together.** `--shard=${{ matrix.shard }}/8`, `matrix.shard: [1..8]`, and the job name `shard ${{ matrix.shard }}/8` are consistent; no `/4` survives anywhere in `.github/`. Checked specifically for the "one instance fixed, class left armed" shape.
- **`npm run test:hooks` — 1757/1757 green at this head.** Re-run rather than carried over, because `dfcdbb0e` restructured the comment block `workflow-wiring.test.mjs` parses. The `# leg:` annotations and step names are untouched and the guard still reads the file cleanly.
- **No stale 4-way reference is owed a fix elsewhere.** The two live hits outside this PR's diff are `docs/superpowers/specs/2026-08-05-verify-scope-map-unification-design.md:257` and the plan/spec's own before/after snippets — narrative and decision-record prose, which `INDEX.md:28-39` explicitly says is historical and *"never rewritten to 'fix' it — that would falsify the record."* Correctly left alone.
- **The plan doc's 29 unchecked boxes are not a finding.** I checked whether shipping this work left `docs/superpowers/plans/2026-09-02-test-suite-hardening.md`'s checkbox-level lifecycle stale, then tallied every plan in that directory: ~200 files, the overwhelming majority at 0-checked, including plans `INDEX.md` itself marks `stable` (e.g. `2026-06-17-flaky-test-release-hardening.md`, 0/80). Leaving them unchecked is the repo's settled practice, not something this PR broke.
- **`timeout-minutes: 15` still has headroom** — slowest shard fell from 559s to 358s, so the margin improved rather than tightened.
- **The two 4-way baselines are legitimately comparable.** Both are `workflow_dispatch` (unscoped, full battery per the workflow's own `on:` comment), both from 2026-09-01, both ran all 4 shards to completion, and neither branch touches `e2e/`.

---

### Verdict

**✅ Approved — ready to merge.** Two 🟡 cleanup findings, **no 🔴 and no 🟠**.

Pass 2's 🟠 is genuinely and completely fixed. I re-measured every figure in the correction from the API rather than trusting the commit message, checked the run-to-number pairing at both sites (they are written in opposite orders), re-derived the percentage, and re-verified the no-queuing and failure-attribution claims. All of it holds. `verify.yml` now records what actually happened: 8-way sharding is a real 31–36% wall-clock improvement over the measured 4-way baseline.

**On the cap:** this pass found no correctness-level defect, so per [`findings-triage.md`](.claude/skills/pr-review-gate/references/findings-triage.md) the re-review trigger is not tripped — a fourth pass would not be warranted even if one were available. The two 🟡s are still fixed in this round per the report-fix-record rule, but fixing them does not reopen the loop and does not need another review. They are:

1. `verify.yml:34-37` — re-attach `~16 min` to the ~110-file battery it was measured on (one parenthetical).
2. PR body steps 1/4 — swap the N/A reason to CLAUDE.md step 1's *"Small/localized items skip the plan doc"* (one sentence, body-only, no commit).

Neither touches runtime behaviour, CI config semantics, or any shipped artifact's correctness.

Tree unmodified throughout: `dfcdbb0ef8f89ca08dbaec97db5f731d5d57ed0d`, `git status --porcelain` empty at both ends.

PR #2863, opened 2026-09-02
chore(scripts): print voices/ junction reminder in wt-new.mjs (#2811)

### LINKED ISSUE #2811: chore(scripts): worktree setup junctions .venv but not tts-sidecar/voices/, causing misleading recycle-storm failures
## What happened

While running on-box acceptance tests for register row A1 (fs-38 Wave 3, voice cloning) in a fresh worktree (`node scripts/wt-new.mjs`), TTS chapter generation on Qwen (`qwen3-tts-0.6b`) consistently failed with `errorCode: recycle-storm`:

> The TTS sidecar recycled 2× while rendering this single chapter — it is likely thrashing (host-memory leak or insufficient VRAM/RAM headroom)... If it persists, the host-memory leak (side-11) needs headroom.

This looked like a side-11 regression (#399, closed) but was reproducible twice in a row within ~10 seconds of starting the very first line — far too fast to be the classic variable-input-shape leak, which needs many syntheses to accumulate.

## Root cause

CLAUDE.md's worktree setup instructions (and `scripts/wt-new.mjs`) only junction `server/tts-sidecar/.venv` from the primary checkout. They do **not** junction `server/tts-sidecar/voices/` — the directory holding the actual model weights (Qwen HF cache, Kokoro ONNX weights, Coqui/XTTS weights, cloned-voice artifacts).

A fresh worktree therefore has no `server/tts-sidecar/voices/` directory at all. On sidecar boot, something (looks like a per-engine self-test/preload path) attempts to load Kokoro and throws repeatedly:

```
RuntimeError: Kokoro model not found at <worktree>\server\tts-sidecar\voices\kokoro\kokoro-v1.0.onnx.
Run server/tts-sidecar/scripts/install-kokoro.ps1 to download weights.
```

Each of these appears to trigger (or coincide with) a sidecar process restart (`sidecar memory watchdog started` logged twice within the same chapter render). Two restarts within one chapter render trips the recycle-storm circuit breaker and aborts the chapter — a false alarm caused by missing weights, not genuine memory/VRAM pressure.

**Confirmed fix:** junctioning `server/tts-sidecar/voices/` from the primary checkout (same technique as the existing `.venv` junction) and restarting the sidecar made the exact same chapter render complete successfully, twice.

## Fix needed

1. `scripts/wt-new.mjs` (or its printed "next steps") should junction `server/tts-sidecar/voices/` alongside `.venv`, or at minimum print an explicit reminder — right now a fresh worktree silently has a broken TTS path that manifests as a misleading, high-severity error message pointing at the wrong root cause (side-11).
2. CLAUDE.md's "Worktree setup" section (the numbered junction checklist) should list `voices/` as a required junction for any worktree doing real sidecar/TTS work, alongside the existing `.venv` step.

## Evidence

- Worktree: `wt-onbox-a1-wave10` (branch `chore/ops-onbox-a1-wave10`), 2026-08-31.
- `logs/tts.err.log` in that worktree shows the repeated `Kokoro model not found` traceback immediately preceding each `sidecar memory watchdog started` restart line.
- Two consecutive `recycle-storm` failures on chapter 2 of `the-coalfall-commission.md` before the fix; two consecutive successful completions (`chapter_complete`, `audioQa.status: "ok"`) after junctioning `voices/`.

Observed by Claude Code (dudarenok-maker), 2026-08-31, during an on-box acceptance run on register row A1.

dudarenok-maker:
## Summary

Adds a printed reminder to `wt-new.mjs`'s "Next steps" output telling anyone
doing sidecar/TTS work to junction both `server/tts-sidecar/.venv` and
`server/tts-sidecar/voices/` from the primary checkout before starting,
referencing CLAUDE.md's "Worktree setup" step 2 and #2811. A missing
`voices/` junction causes Kokoro to throw `Kokoro model not found` errors
that trigger repeated restarts and trip the `recycle-storm` circuit
breaker — which can be misdiagnosed as the unrelated #399 host-memory-leak
regression if `logs/tts.err.log` isn't checked first.

Closes #2811

Also fixed, found in passing during the mandatory PR review gate pass:
- Reminder paths were unqualified (`.venv`, `voices/`) right after a
  root-relative `cd`, which would be misread as worktree-root-relative —
  corrected to the full `server/tts-sidecar/...` paths (commit `d18e8d25`).
- The diagnostic explanation had the #2811/#399 causal chain backwards —
  corrected to match CLAUDE.md's and #2811's actual account (same commit).
- A related doc gap in CLAUDE.md itself (step 2 warned about a missing
  `.venv` junction without ever instructing operators to make one) is
  fixed separately in docs-only PR #2873 (#2869).

Release notes (before-shipping step 5): waived — `wt-new.mjs` is a
dev-tooling script, not in `scripts/build-release-zip.mjs`'s shipped
allowlist, so there's no shippable user-facing delta.

## Test plan

- [x] `npm run test:hooks` — 1777/1777 pass, including new coverage in
      `scripts/tests/wt-new.test.mjs` asserting the printed block mentions
      both `.venv` and `voices/`.
- [x] Mutation kill (re-verified independently during review): temporarily
      reverted the new reminder lines and reran
      `node --test scripts/tests/wt-new.test.mjs` — the new test failed with:
      `AssertionError [ERR_ASSERTION]: The input did not match the regular
      expression /\.venv/` (50/51 pass, 1 fail). Restored the fix and reran —
      51/51 pass again.
- [x] `npm run typecheck` — clean (frontend `tsc --noEmit` + server
      `tsc --noEmit -p .`).
- [x] After the review-gate fix commit: `node --test scripts/tests/wt-new.test.mjs`
      — 51/51 pass again.
- [x] Scope check: only `scripts/wt-new.mjs` and
      `scripts/tests/wt-new.test.mjs` touched — no changes to
      `server/tts-sidecar/`, `buildInstallCommands`, or `runInstalls`.

### DIFF
--- a/scripts/tests/wt-new.test.mjs
+++ b/scripts/tests/wt-new.test.mjs
@@ -371,6 +371,18 @@ test('renderLaunchBlock omits npm install when install=true (auto-installed)', (
   assert.match(block, /npm run dev/);
 });
 
+test('renderLaunchBlock reminds about the .venv AND voices/ junctions for sidecar/TTS work (#2811)', () => {
+  const block = renderLaunchBlock({
+    worktreePath: 'C:/Claude/Projects/wt-foo',
+    branch: 'feat/server-foo',
+    ports: computePorts(1),
+    slot: 1,
+    install: true,
+  });
+  assert.match(block, /\.venv/);
+  assert.match(block, /voices\//);
+});
+
 test('renderLaunchBlock includes both npm install lines when install=false', () => {
   const block = renderLaunchBlock({
     worktreePath: 'C:/Claude/Projects/wt-foo',
--- a/scripts/tests/wt-new.test.mjs
+++ b/scripts/tests/wt-new.test.mjs
@@ -379,8 +379,8 @@ test('renderLaunchBlock reminds about the .venv AND voices/ junctions for sideca
     slot: 1,
     install: true,
   });
-  assert.match(block, /\.venv/);
-  assert.match(block, /voices\//);
+  assert.match(block, /server\/tts-sidecar\/\.venv/);
+  assert.match(block, /server\/tts-sidecar\/voices\//);
 });
 
 test('renderLaunchBlock includes both npm install lines when install=false', () => {
--- a/scripts/wt-new.mjs
+++ b/scripts/wt-new.mjs
@@ -222,9 +222,9 @@ export function renderLaunchBlock({ worktreePath, branch, ports, slot, install }
     `Next steps — paste into a new terminal:`,
     ``,
     `  cd "${winPath}"`,
-    `  # Doing real sidecar/TTS work? Junction .venv AND voices/ from the primary`,
-    `  # checkout first — see CLAUDE.md "Worktree setup" step 2. A missing`,
-    `  # voices/ junction misreports as recycle-storm (#2811), not a missing-weights error.`,
+    `  # Doing real sidecar/TTS work? Junction server/tts-sidecar/.venv AND server/tts-sidecar/voices/ from`,
+    `  # the primary checkout first — see CLAUDE.md "Worktree setup" step 2. Missing voices/ causes`,
+    `  # Kokoro model errors that trigger recycle-storm; check logs/tts.err.log if #399 looks like it.`,
   ];
   if (install === false) {
     lines.push(`  npm install                  # root deps + husky hooks (skipped: --no-install)`);
--- a/scripts/wt-new.mjs
+++ b/scripts/wt-new.mjs
@@ -222,6 +222,9 @@ export function renderLaunchBlock({ worktreePath, branch, ports, slot, install }
     `Next steps — paste into a new terminal:`,
     ``,
     `  cd "${winPath}"`,
+    `  # Doing real sidecar/TTS work? Junction .venv AND voices/ from the primary`,
+    `  # checkout first — see CLAUDE.md "Worktree setup" step 2. A missing`,
+    `  # voices/ junction misreports as recycle-storm (#2811), not a missing-weights error.`,
   ];
   if (install === false) {
     lines.push(`  npm install                  # root deps + husky hooks (skipped: --no-install)`);


dudarenok-maker:
## PR review — pass 1 (head 8ca1911915fcb09de626bbe75d781ee7dc913ba1, depth medium)

Scope: `scripts/wt-new.mjs` (+3), `scripts/tests/wt-new.test.mjs` (+12) · verified: `node --test scripts/tests/wt-new.test.mjs` 51/51 green at head, plus **two independent mutation kills** (below) before adversarially probing the printed reminder's accuracy against its own two cited sources (#2811 and CLAUDE.md "Worktree setup").

Depth derived from the sole commit's type (`feat(scripts)`), not the PR title (`chore(scripts)`) — see 🟡 Minor 3.

---

### 🟠 Significant — the reminder omits both paths, and the paths a reader will infer are the wrong ones

`scripts/wt-new.mjs:225-227`

```js
`  # Doing real sidecar/TTS work? Junction .venv AND voices/ from the primary`,
```

Bare `.venv` and `voices/`, printed as line 2 of a block whose header is *"Next steps — paste into a new terminal"* and whose line 1 is `cd "C:\Claude\Projects\wt-foo"`. Both directories actually live under `server/tts-sidecar/`; neither exists at a checkout root (verified on this box: `ls -d <root>/.venv <root>/voices` → both `No such file or directory`; `server/tts-sidecar/.venv` and `server/tts-sidecar/voices` both present).

Concrete failure: operator runs `node scripts/wt-new.mjs feat/sidecar-foo`, reads the block, and from the directory they were just told to `cd` into runs

```powershell
New-Item -ItemType Junction -Path .venv  -Target ..\Audiobook-Generator\.venv
New-Item -ItemType Junction -Path voices -Target ..\Audiobook-Generator\voices
```

Both targets are non-existent paths, so both junctions dangle. CLAUDE.md step 3 documents that this failure is silent — *"an off-by-one silently points at nothing while `ls` still looks plausible."* `server/tts-sidecar/voices/` is still absent, the sidecar reproduces #2811 exactly, and the operator now believes they junctioned it.

The one fact needed to act on this reminder is the one it doesn't carry, and the surrounding lines (`cd`, `npm install`, `npm run dev`) are all literally-runnable root-relative commands, which actively reinforces the wrong inference. #2811's own evidence is that operators get precisely this wrong.

**Correctness bug.**

---

### 🟠 Significant — "not a missing-weights error" negates the one greppable string that resolves the incident

`scripts/wt-new.mjs:227`

```js
`  # voices/ junction misreports as recycle-storm (#2811), not a missing-weights error.`,
```

Both cited sources say the opposite of the actionable half:

- #2811, Evidence: *"`logs/tts.err.log` in that worktree shows the repeated `Kokoro model not found` traceback immediately preceding each `sidecar memory watchdog started` restart line."*
- `CLAUDE.md:1066-67`: *"Kokoro throws `Kokoro model not found` repeatedly at sidecar boot."*

A missing-weights error **is** emitted — loudly, repeatedly, in `logs/tts.err.log` — and it is the single highest-value diagnostic string available. The chapter-level `errorCode: recycle-storm` is what surfaces to the UI; the missing-weights error is what identifies the cause. The line collapses that into "not a missing-weights error", which on the plain reading tells the operator that hypothesis is excluded.

Concrete failure: operator hits `errorCode: recycle-storm` mid-chapter, recalls this line, concludes it is *not* a weights problem, does **not** grep `logs/tts.err.log` for `Kokoro model not found`, and chases VRAM/RAM headroom instead — the exact wrong path #2811 documents burning a session on ("reproducible twice in a row… looked like a side-11 regression").

Two further accuracy points in the same clause:

- **The misreport target is wrong.** #2811: *"This looked like a side-11 regression (#399, closed)"*; `CLAUDE.md:1068-70`: *"misreporting as a side-11 host-memory-leak regression (#399) when the real cause is the missing junction."* `recycle-storm` is the breaker that trips (the mechanism), not the misdiagnosis. Naming recycle-storm as the misreport and dropping #399/side-11 removes the bridge an operator who has *already been shown* `recycle-storm` actually needs.
- **The citation is mis-attached.** `(#2811)` sits on the misreport clause as though #2811 were the misreport; #2811 is the ticket that *diagnosed* it. CLAUDE.md gets both attributions right; this line inverts one and drops the other.

**Correctness bug** — defect shape #9 (the defect is in the instrument/document), in an instruction operators and agents follow literally.

---

### 🟡 Minor

**1. The cited anchor does not contain half of what the pointer promises** — `scripts/wt-new.mjs:226` → `CLAUDE.md:1047-1070`

The line says *"Junction .venv AND voices/ … see CLAUDE.md "Worktree setup" step 2."* Step 2 instructs junctioning `node_modules`, `server/node_modules`, and `server/tts-sidecar/voices/`. **It contains no `.venv` junction instruction at all** — `.venv` appears there only as a comparison (*"git-ignored like `.venv`, but junctioning only `.venv` leaves it absent"*). Grepping CLAUDE.md at this head SHA, the only other `.venv` mention is line 346 (the `test:sidecar` venv path). So a reader sent to step 2 for the `.venv` half finds nothing.

Compounding it: the numbered list step 2 belongs to is scoped by its own preamble — *"**A tool-created worktree is NOT set up for you.** `EnterWorktree` and the Agent tool's `isolation: "worktree"` create the tree and nothing else… **In that case:**"* — i.e. explicitly to tool-created worktrees, which a `wt-new.mjs` user by definition is not. A wt-new operator following the pointer lands in a list whose preamble reads as not applying to them.

This surfaces a chore in CLAUDE.md itself: step 2 should list `server/tts-sidecar/.venv` as a junction rather than only alluding to it, and the `voices/` sub-clause should not be reachable only through a tool-created-worktree preamble when #2811 was hit in a `wt-new.mjs` worktree ("*While running on-box acceptance tests… in a fresh worktree (`node scripts/wt-new.mjs`)*").

**2. Release-notes pair neither updated nor explicitly waived** — PR body

CLAUDE.md before-shipping step 5 requires `docs/release-notes-next.md` + `RELEASE_NOTES.md`, *"or say so explicitly rather than silently omitting."* The PR body's Test plan covers `test:hooks`, mutation kill, typecheck and scope, and says nothing about release notes. The waiver itself is almost certainly correct — `scripts/wt-new.mjs` is not in `scripts/build-release-zip.mjs`'s shipped-scripts allowlist (lines ~85-116; it's maintainer-only tooling), so there is no shippable delta. The finding is the silent omission, not the judgment.

**3. PR title type and sole-commit type disagree, and the commit type is what the gate ladder reads**

Title `chore(scripts): …` vs sole commit `feat(scripts): …`. CONTRIBUTING.md:469-471 permits title/commit divergence, but its stated rationale is multi-commit PRs ("it typically does [match], because most PRs are one commit") — this is a one-commit PR. The *type* half is materially different from a wording divergence here: `pr-review-gate`'s depth ladder derives `low` from `chore` and `medium` from `feat`, so this PR's two declarations imply two different gate tiers and the dispatch had to adjudicate it by hand. On the merits `chore` is the right one — #2811 is labelled `type:chore`, and the change adds a printed string to a dev script with no product behavior — so `feat` on the commit is the half that overstates.

---

### ✅ What is solid

- **The paired test is a real regression test, verified independently rather than taken from the PR body.** Exported the head tree to a scratch copy and ran two separate mutations:
  - stripping all three reminder lines → `50/51`, failing with `AssertionError: The input did not match the regular expression /\.venv/`;
  - keeping `.venv` but removing only the `voices/` mentions → `50/51`, failing on `/voices\//`.

  Both halves kill independently, so the assertion that actually matters for #2811 (`voices/`) is genuinely load-bearing and not carried by its neighbour. Neither string occurs elsewhere in the rendered block (confirmed from the mutation's printed input).
- **`Closes #2811` is plain, outside any code span**, and the PR does discharge the issue: item 2 of its "Fix needed" (CLAUDE.md listing `voices/`) is already on `main` at `CLAUDE.md:1062-70`, and item 1 explicitly offers *"or at minimum print an explicit reminder"*, which this takes.
- **The class is not left armed elsewhere.** Checked for sibling surfaces printing worktree-setup guidance: `wt-list.mjs` prints none, and CONTRIBUTING.md:94-97 defers to CLAUDE.md rather than restating a `.venv`-only list. No second copy to fix.
- **No register/live-view or BACKLOG debt.** Nothing hardware-provable ships (the change prints a string), and #2811 is `type:chore`, which never renders in `docs/BACKLOG.md`. The register's two `#2811` mentions (lines 1007, 1093) are narrative context on rows C-02/C-03, not state keyed to #2811's resolution.
- **Standalone `#` comment lines are established precedent in this exact block** (`# in another tab:` at line 234), so the new lines introduce no new paste hazard beyond what already shipped.

### Verdict

**Changes requested.** Two 🟠 correctness findings, both in the printed text that is this PR's entire payload: the reminder omits the two paths needed to act on it, and its diagnostic clause steers away from the one log string that resolves the incident while mis-attributing the misreport (#399/side-11, not "recycle-storm (#2811)"). The mechanism — test, wiring, issue linkage — is sound; the content is not yet, and for a change whose only product *is* content that is the whole of it.

Per the gate's re-review trigger, fixing and pushing the 🟠 findings re-triggers a pass.

dudarenok-maker:
## PR review — pass 2 (head d18e8d2539229ffba11444ffb318ae45566b0875, depth medium)

Re-review of pass 1 ([comment](https://github.com/dudarenok-maker/Castwright/pull/2863#issuecomment-5531692716), head `8ca1911`). Deltas only — solid items from pass 1 are not re-listed. Depth re-derived from the current commit set: `feat(scripts)` + `fix(scripts)`, both single-scope `scripts` → **medium**, unchanged.

**🟠 Significant 1 — unqualified `.venv` / `voices/` paths after a root-relative `cd`: FIXED, verified.**
`scripts/wt-new.mjs:225-227` now prints `server/tts-sidecar/.venv` and `server/tts-sidecar/voices/`. Both resolve correctly against the `cd "${winPath}"` emitted on line 224, and both match the real on-disk paths (`server/tts-sidecar/.venv\Scripts\python.exe` per CLAUDE.md's `test:sidecar` entry; `server/tts-sidecar/voices/` per Worktree-setup step 2).

**🟠 Significant 2 — backwards #2811/#399 causal chain: FIXED, verified.**
The new text ("Missing `voices/` causes Kokoro model errors that trigger recycle-storm; check `logs/tts.err.log` if #399 looks like it") matches CLAUDE.md's account — missing weights → repeated `Kokoro model not found` at sidecar boot → enough restarts to trip the `recycle-storm` breaker → misread as the side-11 host-memory-leak regression (#399).

I also checked the new `logs/tts.err.log` pointer against the flow this block actually instructs, since the block's next line is `npm run dev` and this repo has a known trap where `logs/server.log` is written **only** by `npm start` (`scripts/start-app-prod.mjs:315`) — a pointer inherited from that pattern would have been dead advice. It isn't: `server/src/tts/spawn-sidecar.ts:758-759` opens `logs/tts.log` / `logs/tts.err.log` and hands the child their raw fds whenever the **server** spawns the sidecar, which it does under `npm run dev` as well (CLAUDE.md: "the server auto-starts the TTS sidecar same as `npm start`"). So the pointer is live for the instructed flow. Note `docs/testing/1996-stranded-vram-measurement.md:25-30` states `npm run dev` doesn't capture `tts.err.log`; that prose is imprecise about the sidecar log specifically, but it's a pre-existing doc in a file this PR doesn't touch and not this PR's business.

**🟡 Minor 1 — CLAUDE.md step 2 never instructed the `.venv` junction: disposition accurate.**
Issue #2869 (`area:ops`, `type:chore`) is open, and PR #2873 is open with a diff that does close it — step 2 becomes "**For real sidecar/TTS work, also junction `server/tts-sidecar/.venv` AND `server/tts-sidecar/voices/`**". Splitting a CLAUDE.md-only fix out of a `scripts` PR is the right call under one-branch-one-cohesive-change, and it's declared in this PR body.
*Residual, non-blocking:* an ordering dependency, not a defect. Until #2873 merges, this PR's printed text tells operators to junction `.venv` and points them at a step that doesn't mention `.venv`. Worth merging #2873 in the same batch.

**🟡 Minor 2 — release-notes waiver: RESOLVED.** The PR body now states it explicitly, with a reason (`wt-new.mjs` is dev tooling, outside `scripts/build-release-zip.mjs`'s shipped allowlist).

**🟡 Minor 3 — PR title `chore` vs commit type `feat`: no action needed, confirmed.** The title-lint check passes and `chore(scripts)` remains a fair label for the merge commit now that the branch carries `feat` + `fix`.

### New finding

**🟡 Minor — cleanup / house gate — the fix commit ships no regression test; the head test passes against the pre-fix code.** `scripts/tests/wt-new.test.mjs:374-385`

`d18e8d25` is typed `fix(scripts)` and changes only `scripts/wt-new.mjs`, with no test delta. Its two assertions are substring matches that the round-1 *defective* text also satisfies:

```js
assert.match(block, /\.venv/);     // matches "Junction .venv AND voices/ from the primary"
assert.match(block, /voices\//);   // matches the same line
```

Demonstrated, not inferred. I materialised head's `scripts/tests/wt-new.test.mjs` alongside **`8ca1911`'s** `wt-new.mjs` (the exact code pass 1 flagged, carrying both the unqualified paths and the backwards diagnostic) in a scratch tree and ran it:

```
### reminder lines under test (PRE-FIX):
225:  # Doing real sidecar/TTS work? Junction .venv AND voices/ from the primary
226:  # checkout first — see CLAUDE.md "Worktree setup" step 2. A missing
227:  # voices/ junction misreports as recycle-storm (#2811), not a missing-weights error.

✔ renderLaunchBlock reminds about the .venv AND voices/ junctions ... (#2811)
ℹ tests 51 · pass 51 · fail 0
```

So the suite is entirely insensitive to the fix: revert d18e8d25's three lines and `test:hooks` stays green. This is reviewer-brief shape 3 ("a test that cannot fail — or that went red before the fix for a different reason than the one claimed") and house-gate bullet 1, against CLAUDE.md's "Bug fix → ship a regression test that fails before the fix and passes after."

To be fair to the PR body: its mutation-kill evidence is *true* but scoped to the wrong commit — deleting the whole reminder does make `/\.venv/` fail, which covers the `feat` commit `8ca1911`. Nothing in the suite covers `d18e8d25`. The last checkbox ("51/51 pass again") is honest and claims no red-before.

Fix is one line each — assert the qualified paths the fix introduced:

```js
assert.match(block, /server\/tts-sidecar\/\.venv/);
assert.match(block, /server\/tts-sidecar\/voices\//);
```

Those go red against `8ca1911` for exactly the reason pass 1 raised, and green at head.

### Checked, nothing to report

- **Printed line width.** Lines 225-227 print at ~94-102 chars vs ~79-90 for the rest of the block. Not raised: line 218 is already a 135-char template on `main` and passes lint, `npm run lint` is `eslint . --max-warnings 0` with `eslint-config-prettier` (formatting rules *disabled*, no `max-len`), and Prettier cannot break a template literal's contents anyway. Wrapping preference here is taste, which is off-limits under "Surgical changes".
- **Fresh scoped pass over d18e8d25** for new correctness issues: the CLAUDE.md section name and step number cited are accurate; `#399` is correctly identified as the host-memory-leak regression; the reminder's placement before the conditional `npm install` lines is harmless (`npm install` touches neither `.venv` nor `voices/`).
- `Closes #2811` present in the PR body, outside any code span; #2811 open. Incidental fixes declared under "Also fixed, found in passing". No on-box acceptance, `cast.json`, config-knob, or derived-artifact surface in this diff.
- 51/51 green at head (reproduced independently in a scratch tree). All 19 CI checks pass; `mergeStateStatus: CLEAN`.

### Verdict

**No 🔴/🟠 findings remain — both pass-1 Significants are genuinely fixed, confirmed against the diff rather than the claim.** One new 🟡 cleanup finding (the fix has no regression test). Per `findings-triage.md`, a 🟡-only pass is still fixed this round but does **not** re-trigger another review pass — so tighten the two assertions, push, and merge without a pass 3. Ideally land #2873 alongside it.

https://claude.ai/code/session_01RdXD7FPMFFkUftMCJwmKT3

PR #2818, opened 2026-08-31
fix(sidecar): fix Kokoro CUDA cuDNN pin, add missing DLL search paths

dudarenok-maker:
## Summary

Kokoro's CUDA execution provider has never actually run on this box despite five prior on-box acceptance waves narrowing the cause (register row A28). This closes it, with three real defects found and fixed:

- **`install-ort.mjs`'s `extraRuntimeSteps` was missing `nvidia-cufft-cu12`/`nvidia-cuda-runtime-cu12`** — wave-5's own validated "hypothesis 2" (2026-08-23) was never turned into code. Added, floor-plus-cap pinned to torch's actual measured bundled DLL build line (`~=11.3.3`/`~=12.8.0`); `curand` tried and confirmed unneeded, dropped.
- **New root cause found after fixing that**: cuDNN 9's own lazily-dlopened "engine" plugin DLLs (`cudnn_engines_tensor_ir64_9.dll` etc.) are never covered by `onnxruntime.preload_dlls()` and are not found via `os.add_dll_directory()` either — only prepending `nvidia/<pkg>/bin` to the process `PATH` works. New `main._add_nvidia_dll_dirs_to_path()`, made idempotent after a real regression: a FastAPI `TestClient` re-triggers the lifespan per test, which blew past Windows' 32767-char `PATH` ceiling without the idempotence guard.
- **Fixing that surfaced a third, more serious pre-existing defect before it shipped**: `NVIDIA_CUDNN_CONSTRAINT`'s unpinned `~=9.0` resolved to 9.25.1.1, ABI-incompatible with the cuDNN 9.19.0 torch 2.11.0+cu128 bundles — loading it before `import torch` broke Coqui/Qwen/Whisper outright on this box. Tightened to `~=9.19.0` (floor-and-cap on torch's exact bundled line, same reasoning `NVIDIA_CUBLAS_CONSTRAINT` already uses).

Verified end-to-end on real hardware via the real HTTP surface (RTX 4070 8GB, `cuda:0`, no `KOKORO_DEVICE` pin — the shipped default path): sidecar startup reports `Device probe complete: {'kokoro': 'cuda', 'coqui': 'cuda', 'qwen': 'cuda'}` with no torch breakage; a real `POST /synthesize` returned real PCM at real GPU speed (0.45s warm synth) with genuine VRAM allocated; `/health` reports `cuda_verified: true` with no false `stale_reason`.

[...]

Issue #2861, opened 2026-09-02
[agent instructions][claude][task] measure+fix ctranslate2 cudnn shadow on real hardware (#2845)


dudarenok-maker:
## What

Measure, on real hardware, whether the installer's `nvidia/cudnn/bin/cudnn64_9.dll`
(9.19.0.56) shadows `ctranslate2`'s own bundled `cudnn64_9.dll` (9.10.2.21) on the
Whisper/ASR path — and if the shadow is confirmed harmful, fix it.

This machine (the one this worktree runs on) **is** the real hardware the parent
issue's acceptance criterion requires — a live sidecar with a real CUDA install
and an NVIDIA GPU. Confirm that's still true first (`nvidia-smi` — note one GPU
slot on this box has previously reported "GPU is lost" errors requiring a reboot,
unrelated to this issue; if the GPU is genuinely unusable when you start, say so
and stop rather than mis-attributing a transcribe failure to the DLL shadow).

## Background you need

- `server/tts-sidecar/main.py:147-284` (`_add_nvidia_dll_dirs_to_path`) — added
  by PR #2818 (Castwright#2752-followup / register row A28). It prepends every
  `<site-packages>/nvidia/<pkg>/bin` directory (including `nvidia/cudnn/bin`,
  version pinned by `NVIDIA_CUDNN_CONSTRAINT = 'nvidia-cudnn-cu12~=9.19.0'` in
  `server/tts-sidecar/scripts/install-ort.mjs:290`) to the process `PATH`, once,
  at sidecar startup, unconditionally — see the docstring at `main.py:462-467`
  for why this runs first and unconditionally ("a superset fix -- prepending
  PATH never hurts a CPU/AMD/Apple install where nvidia/ doesn't exist").
- **Do not weaken or remove this mechanism for onnxruntime's own sake.** It
  exists because `preload_dlls()` + `os.add_dll_directory()` do NOT find
  cuDNN 9's lazily-dlopened engine plugin DLLs (`cudnn_engines_tensor_ir64_9.dll`
  etc.) — only a PATH prepend does, confirmed on real hardware (see the long
  docstring comment at `main.py:147-182`). Kokoro/Coqui/Qwen's CUDA execution
  provider depends on this exact mechanism. Any fix here must not silently
  fall back those three engines to CPU.
- `ctranslate2` (faster-whisper's dependency, pinned `faster-whisper>=1.0,<2.0`
[...]

PR #2864, opened 2026-09-02
chore(ops): measure and close ctranslate2 cudnn shadow risk (#2845)

### LINKED ISSUE #2845: ctranslate2's bundled cudnn64_9.dll (9.10.2.21) may be shadowed by installer's 9.19.0.56 on the Whisper/ASR path
Found during PR #2818's review-gate pass 3 (`chore/sidecar-kokoro-cuda-path`, fix agent adfa8c4).

`_add_nvidia_dll_dirs_to_path()` (added by PR #2818, `server/tts-sidecar/main.py`) prepends every `nvidia/<pkg>/bin` directory ahead of torch on the process `PATH`, so any bare-name DLL load resolves to the installer's copy first — not just the ones `onnxruntime.preload_dlls()` itself makes. This is broader than originally scoped (see `install-ort.mjs`'s NVIDIA_CUDNN_CONSTRAINT block comment, 'PASS-3 REVIEW NOTE').

Checked on-box, real `FileVersion` comparison for the three basenames this newly-widened PATH mechanism can shadow:
- `nvrtc64_120_0.dll` / `nvJitLink_120_0.dll`: torch's `torch/lib` copies and the installer's `nvidia/*/bin` copies both report `6.14.11.9000` — agree, no risk.
- `cublas64_12.dll` / `cublasLt64_12.dll`: torch's copies report `12.8.4`, the installer's `nvidia/cublas/bin` copies report `12.8.5` — differ, but same-minor, low practical risk.
- **`cudnn64_9.dll` from `ctranslate2`** (faster-whisper's own bundled copy, `server/tts-sidecar/.venv/Lib/site-packages/ctranslate2/`): reports `9.10.2.21`. The installer's `nvidia/cudnn/bin/cudnn64_9.dll` (pinned `~=9.19.0` via `NVIDIA_CUDNN_CONSTRAINT`) reports `9.19.0.56` — a **cross-minor gap (9.10 → 9.19) wider than the 9.25→9.19 gap PR #2818 exists to fix.**

Whisper/ASR is the one engine PR #2818's on-box verification run did NOT exercise (`Device probe complete: {'kokoro': 'cuda', 'coqui': 'cuda', 'qwen': 'cuda'}` — ASR is off by default unless `SEG_ASR_ENABLED`), so whether this shadow actually causes a problem in practice (an ABI-incompatible cuDNN build loaded into ctranslate2's process) is unconfirmed either way.

**Acceptance:** with `SEG_ASR_ENABLED=1` and a live sidecar (real CUDA install), run a real `/transcribe` call and confirm ctranslate2/Whisper actually uses a compatible cuDNN (no crash, no silent CPU fallback, no version-mismatch error in logs). If the shadow is confirmed harmful, either exclude `ctranslate2`'s own cudnn copy from the PATH-prepend candidate list, or pin `ctranslate2`'s own cudnn dependency to match `NVIDIA_CUDNN_CONSTRAINT`.

dudarenok-maker:
## Summary

Investigates issue #2845's concern: whether the ctranslate2-bundled
`cudnn64_9.dll` (9.10.2.21) could be affected by a cross-minor version gap
against the installer's PATH-prepended `nvidia/cudnn/bin` copy (9.19.0.56).

**Result: not applicable.** Binary inspection of ctranslate2's compiled code
(`ctranslate2.dll`, `_ext.cp312-win_amd64.pyd`) found **zero** references to
cuDNN anywhere — no import, no delay-load, no string reference — versus 35
occurrences of "cuBLAS" (real symbol names like `cublasGemmEx`,
`CUBLAS_STATUS_*`) in the same binary, confirming the method and showing
ctranslate2 uses cuBLAS, not cuDNN, for its
GPU compute. The bundled `cudnn64_9.dll` sitting in ctranslate2's package
directory is loaded into the process anyway, but only because
`ctranslate2/__init__.py` unconditionally `ctypes.CDLL`-loads every `.dll` in
its own package directory at import time — it is resident but never called.
There is no cross-version gap to worry about for ctranslate2, because
ctranslate2 never reads from a cuDNN sublibrary at any version.

On the real GPU (RTX 4070 Laptop, GPU 0), a live sidecar booted
`ASR_DEVICE=cuda`, a Kokoro-synthesized sample fed into `/transcribe`
(cold-loading ctranslate2/faster-whisper) returned a correct transcript on
`cuda:0`, with no CUDA/cuDNN error and no silent CPU fallback — consistent
with the binary-level finding. `Get-Process -Module` showed `cudnn64_9.dll`
loaded from three locations at two versions (`nvidia\cudnn\bin` and
`torch\lib` both 9.19.0.56, `ctranslate2\` at 9.10.2.21); this measurement
does not attribute which of onnxruntime or torch actually used the 9.19.0.56
set (both are genuine cuDNN consumers — torch's own `torch_cuda.dll` alone
has 1773 cuDNN references), because that attribution is orthogonal to
ctranslate2's own verdict.

`_add_nvidia_dll_dirs_to_path` (the #2818/A28 fix for Kokoro/Coqui/Qwen CUDA)
is left untouched, per the ticket's explicit instruction — it serves
onnxruntime's/torch's real cuDNN activity, not ctranslate2's, since
ctranslate2 has none. Full command output, sidecar logs, module-table
inspection, and the binary-grep results are in
`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md`.

## Review

Independent `pr-review-gate` passes (depth `low`) ran seven rounds, each
correcting an overclaim the previous round's fix introduced — the doc's
causal-narrative sentences kept getting patched one at a time rather than
rewritten as a whole, and each patch left an adjacent sentence wrong:
- **Pass 1**: original writeup concluded "no shadow occurs," contradicted by
  its own module-table data.
- **Pass 2**: that fix overclaimed a proven "sublibrary displacement" the
  on-disk layout didn't support.
- **Pass 3**: that fix asserted the PATH-prepend as the cause in one section
  while disclaiming it in another, asserted the opposite mechanism as fact
  elsewhere, and claimed removing the PATH-prepend would cause a failure
  mode — checked on-box and found false.
- **Pass 4**: discovered the deeper problem — ctranslate2 doesn't consume
  cuDNN at all in this build (verified via binary grep), so the entire
  "mixed-version stack that worked" framing described something that never
  happened.
- **Pass 5**: found the pass-4 fix still (a) claimed onnxruntime was "the
  only actual cuDNN consumer" when torch is independently a real cuDNN
  consumer too, (b) stated as fact which component loaded the sublibraries
  when a process-level snapshot can't establish that, and (c) left one
  sentence at the top of the doc still asserting the disproven premise
  ("triggers ctranslate2's first cuDNN load").
- **Pass 6**: after a holistic (not sentence-patched) rewrite fixed all of
  pass 5's findings, found one wrong companion number — the rewrite fixed the
  cuBLAS occurrence count (3 → 35) but introduced a wrong distinct-symbol
  count (24; actual is 26 under every reasonable counting method) — plus a
  cleanup nit (one sentence attributed the `nvidia\cudnn\bin` copy's presence
  to the PATH-prepend mechanism without the hedge the surrounding paragraphs
  already carry).

Final pass (this commit): re-verified every number directly one more time
(0 cuDNN references confirmed independently; 35 cuBLAS occurrences, not the
"3" the doc originally claimed — that number came from a flawed `grep -c -o`
combination that undercounts in binary files; 1773 cuDNN references in
`torch/lib/torch_cuda.dll`, confirming torch is a real independent consumer)
and dropped the specific distinct-symbol count rather than risk a further
wrong number — the load-bearing contrast is 35 occurrences vs. 0, not the
exact symbol tally. Softened the PATH-prepend attribution to say only which
package a file belongs to, not which mechanism loaded it into the process.
The doc no longer attributes the loaded sublibraries to any single
component, states plainly what it can and can't establish, and drops the
leftover "ctranslate2's first cuDNN load" sentence. Verdict unchanged and
strengthened: no fix needed, because ctranslate2 doesn't consume cuDNN at
all.

- **Pass 7 (final)**: independently re-derived all four core numbers from
  scratch (0 / 0 / 35 / 1773, matching exactly) and confirmed both pass-6
  findings genuinely fixed. Also probed a wide-char (`LoadLibraryW`)
  encoding the ASCII grep couldn't see, validated the probe against a
  synthetic positive, and got 0/0 there too — the conclusion holds under a
  second encoding. Found no correctness issues; two trivial cleanup nits
  (an orphaned parenthesis left by the pass-6 edit, and this section's own
  stale round count) fixed in this commit, no further review round needed.

**Clean to merge**, per pass 7's explicit verdict.

## On-box acceptance / release notes

- **On-box acceptance register**: no row owed. #2845's criterion (a real
  `/transcribe` call on `ASR_DEVICE=cuda` confirming no crash/no silent CPU
  fallback) is discharged by this PR's own measurement; no existing register
  row references #2845.
- **Release notes**: N/A — docs/comment-only change, no user- or
  operator-visible behavior change.

Closes #2845.

## Test plan

- [x] Real-hardware measurement: `/synthesize` → `/transcribe` round trip on
      `ASR_DEVICE=cuda`, correct transcript, no cuDNN error, no silent CPU
      fallback.
- [x] Module-provenance inspection (`Get-Process -Module`) confirms
      `cudnn64_9.dll` loads from three locations at two distinct versions
      without conflict.
- [x] Binary inspection via grep (occurrence-counted correctly, `-o | wc -l`,
      not `-c -o` which undercounts): zero cuDNN references in ctranslate2's
      compiled code; 35 cuBLAS references confirm the methodology; 1773
      cuDNN references in `torch/lib/torch_cuda.dll` confirm torch as an
      independent consumer.
- [x] Source-level verification: `ctranslate2/__init__.py`'s directory-glob
      DLL load, checked directly against the venv, cited by exact line number.
- [x] `npm run typecheck` — green.
- [x] Cloud `verify.yml` — full battery green on every pushed head.

### DIFF
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -76,15 +76,12 @@ present ONLY under `nvidia\cudnn\bin` and `torch\lib` (version 9.19.0.56) — ne
 bundled under `ctranslate2\` — because ctranslate2 only ships the single
 `cudnn64_9.dll` dispatch DLL entry point, not the full cuDNN library suite.
 cuDNN's dispatch DLL loads these sublibraries lazily (at the first real kernel
-operation) via bare-name `LoadLibrary` calls, which resolves through the process
-`PATH` environment variable, not through explicit paths. This process `PATH` has
-been prepended with `nvidia/cudnn/bin/` at sidecar boot (via `_add_nvidia_dll_dirs_to_path`),
-so the sublibraries loaded are the 9.19.0.56 versions. The net result: ctranslate2
-loaded the 9.10.2.21 dispatch stub (via its explicit package path), but resolved
-its required sublibraries from the PATH-prepended 9.19.0.56 set. This is a
-mixed-version stack (dispatch 9.10.2.21 + sublibraries 9.19.0.56), not a
-fully self-contained 9.10.2.21 stack. The measurement shows this mixed
-combination worked correctly end-to-end on CUDA with a correct transcript.
+operation) via bare-name `LoadLibrary` calls. Every other loaded cuDNN component
+is present in the process from the 9.19.0.56 set. At runtime, ctranslate2's
+dispatch stub resolved its required sublibraries from this 9.19.0.56 set present
+in the process. The result: a mixed-version stack (dispatch 9.10.2.21 +
+sublibraries 9.19.0.56) that worked correctly end-to-end on CUDA with a correct
+transcript.
 
 ### Interpretation
 
@@ -94,39 +91,17 @@ ships `cudnn64_9.dll` inside its own package directory and loads it from there,
 not via a bare-name `PATH` search — so its 9.10.2.21 dispatch stub loads as
 a distinct module, unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend.
 
-However, the bare-name search mechanism is NOT avoided downstream: cuDNN's
-dispatch DLL pattern (documented in `main.py:151-175`) means the dispatch stub
-itself lazily loads the other cuDNN components (`cudnn_ops64_9.dll`,
-`cudnn_graph64_9.dll`, etc.) via bare-name `LoadLibrary` calls, which resolves
-through the process `PATH` environment variable. This is the mechanism that
-`_add_nvidia_dll_dirs_to_path` prepends `nvidia/cudnn/bin/` for — not just for
-onnxruntime's own lazy engine plugins, but for cuDNN itself's later lazy loads.
-At runtime, the dispatch stub loaded by ctranslate2 necessarily resolves its
-required sublibraries from whatever is already resident or discoverable in the
-process — in this measurement, the 9.19.0.56 set from `nvidia/cudnn/bin` and/or
-`torch/lib`. However, this measurement did not isolate whether the PATH-prepend
-was the cause (rather than ctranslate2's bare-name lookup simply being satisfied
-by cuDNN sublibraries already resident from onnxruntime or torch initialization
-before ctranslate2 ran).
-
-The result: ctranslate2 loaded a 9.10.2.21 dispatch stub that resolved its
-9.19.0.56 sublibraries from the process (the 9.19.0.56 versions being already
-present before ctranslate2's dispatch stub executed). This is a mixed-version
-composition, not a displacement of ctranslate2's own sublibraries — ctranslate2
-bundles no cuDNN sublibraries at all, only the single dispatch DLL entry point.
+The dispatch DLL loaded by ctranslate2 (9.10.2.21) resolved its required
+sublibraries from the 9.19.0.56 set present in the process. This is a
+mixed-version composition, not a displacement of ctranslate2's own sublibraries —
+ctranslate2 bundles no cuDNN sublibraries at all, only the single dispatch DLL
+entry point.
 
 ### Conclusion
 
-**No fix needed.** The `install-ort.mjs` PASS-3 REVIEW NOTE (2026-09-01) concern
-(cross-minor cuDNN version gap 9.10→9.19 between ctranslate2's bundled dispatch
-stub and the system's cuDNN sublibraries) results in a mixed-version composition
-on this real hardware, but empirically causes no harm: ctranslate2's 9.10.2.21
-dispatch stub loaded and correctly resolved its required 9.19.0.56 sublibraries,
-executed transcription correctly on CUDA, and produced the correct output. The
-mixed-version stack worked. `_add_nvidia_dll_dirs_to_path` is left untouched,
-per the ticket's explicit instruction not to weaken it — the prepended PATH
-serves a critical purpose for onnxruntime's own lazy engine loads. Moreover,
-excluding the PATH-prepend for ctranslate2 specifically (a remedy #2845 floated)
-would leave its dispatch stub with zero discoverable sublibraries at all,
-reintroducing the exact `CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED` failure #2818/A28
-already fixed, so the status quo is not just harmless but necessary.
+**Empirically harmless. No fix needed.** The cross-minor cuDNN version gap (9.10→9.19)
+between ctranslate2's bundled dispatch stub and the system's cuDNN sublibraries
+results in a mixed-version composition on this real hardware. The measurement
+shows this mixed-version stack works correctly: ctranslate2's 9.10.2.21 dispatch
+stub loaded and correctly resolved its required 9.19.0.56 sublibraries, executed
+transcription correctly on CUDA, and produced the correct output.
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -79,7 +79,7 @@ combination is required for a true occurrence count; `grep -c` alone reports mat
 *lines* in a binary, which undercounts) found **zero** occurrences of "cudnn" in
 either file. The same command against "cublas" on `ctranslate2.dll` finds **35**
 occurrences, including symbol names like `cublasCreate_v2`, `cublasGemmEx`,
-`CUBLAS_STATUS_*`, `cublas64_12.dll`, etc.) — confirming the method finds real
+`CUBLAS_STATUS_*`, `cublas64_12.dll`, etc. — confirming the method finds real
 references when they exist, and that ctranslate2 uses **cuBLAS**, not cuDNN, for
 its GPU matrix operations. `ctranslate2.dll`'s import-name strings are
 `cublas64_12.dll`, `nvcuda.dll`, MKL, and CRT libraries — `cudnn64_9.dll` is not
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -52,12 +52,11 @@ Sidecar log:
 INFO:     127.0.0.1:58098 - "POST /transcribe HTTP/1.1" 200 OK
 ```
 
-### 3. Module-provenance inspection: TWO distinct `cudnn64_9.dll` modules coexist
+### 3. Module-provenance inspection: mixed cuDNN versions coexist
 
 `Get-Process -Id <sidecar pid> -Module` after the `/transcribe` call shows the
-sidecar process has **two separate loaded modules both named `cudnn64_9.dll`**,
-at different full paths, with different versions — Windows did not collapse them
-into a single module:
+sidecar process has loaded `cudnn64_9.dll` from **three different locations**,
+but only **two distinct versions** — Windows did not collapse them into a single module:
 
 | Path | FileVersion |
 |---|---|
@@ -72,35 +71,52 @@ copy; the `ctranslate2\` one is the package's own bundled 9.10.2.21 copy.)
 Every other loaded cuDNN component (`cudnn_ops64_9.dll`, `cudnn_adv64_9.dll`,
 `cudnn_graph64_9.dll`, `cudnn_heuristic64_9.dll`, `cudnn_engines_*64_9.dll`) is
 present ONLY under `nvidia\cudnn\bin` and `torch\lib` (version 9.19.0.56) — never
-duplicated under `ctranslate2\` — because ctranslate2 only bundles the single
-`cudnn64_9.dll` entry-point stub, not the full cuDNN component set; its runtime
-loader resolves the rest through whichever `cudnn64_9.dll` gets loaded first in
-its own process, or ctranslate2 links narrowly against the entry stub only. What
-matters for this measurement is that ctranslate2's forward pass completed
-correctly end-to-end on CUDA with a correct transcript, so whichever component
-set it actually used at runtime worked.
+bundled under `ctranslate2\` — because ctranslate2 only ships the single
+`cudnn64_9.dll` dispatch DLL entry point, not the full cuDNN library suite.
+cuDNN's dispatch DLL loads these sublibraries lazily (at the first real kernel
+operation) via bare-name `LoadLibrary` calls, which resolves through the process
+`PATH` environment variable, not through explicit paths. This process `PATH` has
+been prepended with `nvidia/cudnn/bin/` at sidecar boot (via `_add_nvidia_dll_dirs_to_path`),
+so the sublibraries loaded are the 9.19.0.56 versions. The net result: ctranslate2
+loaded the 9.10.2.21 dispatch stub (via its explicit package path), but resolved
+its required sublibraries from the PATH-prepended 9.19.0.56 set. This is a
+mixed-version stack (dispatch 9.10.2.21 + sublibraries 9.19.0.56), not a
+fully self-contained 9.10.2.21 stack. The measurement shows this mixed
+combination worked correctly end-to-end on CUDA with a correct transcript.
 
 ### Interpretation
 
-Windows' loader identifies a already-loaded module by its **normalized full
-path**, not by base name, when the loading component resolves the DLL via an
-explicit path (as ctranslate2 does for its own bundled copy — it ships
-`cudnn64_9.dll` inside its own package directory and loads it from there, not
-via a bare-name `PATH` search). `_add_nvidia_dll_dirs_to_path`'s PATH prepend
-only affects **bare-name** `LoadLibrary`/`dlopen` resolution (which is what
-onnxruntime's lazily-dlopened cuDNN engine plugins need — the bug it was built
-to fix). ctranslate2 never does a bare-name search for its own `cudnn64_9.dll`,
-so the PATH-prepend candidate never gets a chance to shadow it: both DLLs load
-side by side as distinct modules, ctranslate2 gets its own pinned 9.10.2.21, and
-the onnxruntime-consuming engines (Kokoro/Coqui/Qwen) still get 9.19.0.56 from
-the PATH prepend. There is no shadow in practice on this measurement.
+Windows' loader identifies a module by its **normalized full path**, not by base
+name, when the loading component resolves the DLL via an explicit path. ctranslate2
+ships `cudnn64_9.dll` inside its own package directory and loads it from there,
+not via a bare-name `PATH` search — so its 9.10.2.21 dispatch stub loads as
+a distinct module, unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend.
+
+However, the bare-name search mechanism is NOT avoided downstream: cuDNN's
+dispatch DLL pattern (documented in `main.py:151-175`) means the dispatch stub
+itself lazily loads the other cuDNN components (`cudnn_ops64_9.dll`,
+`cudnn_graph64_9.dll`, etc.) via bare-name `LoadLibrary` calls, which resolves
+through the process `PATH` environment variable. This is the mechanism that
+`_add_nvidia_dll_dirs_to_path` prepends `nvidia/cudnn/bin/` for — not just for
+onnxruntime's own lazy engine plugins, but for cuDNN itself's later lazy loads.
+Since the PATH has been prepended with the 9.19.0.56 set, ctranslate2's dispatch
+stub resolved its sublibraries from that set, not from its own package directory.
+
+The result: ctranslate2 loaded a 9.10.2.21 dispatch stub with 9.19.0.56 sublibraries.
+This is a real cross-version shadow (the sublibraries ARE shadowed from a different
+version), but it is narrower in scope than a full "uncontrolled fallback" scenario
+— only the sublibraries are shadowed, not the dispatch stub itself.
 
 ### Conclusion
 
 **No fix needed.** The `install-ort.mjs:396-414` PASS-3 review note's concern
 (cross-minor cuDNN version gap 9.10→9.19 between ctranslate2's bundled copy and
-the PATH-prepended onnxruntime copy) does not manifest as a runtime shadow on
-this real hardware: ctranslate2 loads and uses its own bundled cuDNN 9.10.2.21
-via its own absolute package path, unaffected by `_add_nvidia_dll_dirs_to_path`'s
-PATH prepend, and transcription is correct. `_add_nvidia_dll_dirs_to_path` is
-left untouched, per the ticket's explicit instruction not to weaken it.
+the PATH-prepended copy) manifests as a cuDNN sublibrary shadow on this real
+hardware, but empirically causes no harm: ctranslate2's 9.10.2.21 dispatch stub
+loaded and correctly resolved its 9.19.0.56 sublibraries from the PATH-prepended
+set, executed transcription correctly on CUDA, and produced the correct output.
+The mixed-version stack worked. `_add_nvidia_dll_dirs_to_path` is left untouched,
+per the ticket's explicit instruction not to weaken it — the prepended PATH
+serves a critical purpose for onnxruntime's own lazy engine loads, and the fact
+that it also causes ctranslate2's sublibraries to be shadowed from a newer version
+is a side effect that does not cause problems in practice on this hardware.
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -101,24 +101,32 @@ itself lazily loads the other cuDNN components (`cudnn_ops64_9.dll`,
 through the process `PATH` environment variable. This is the mechanism that
 `_add_nvidia_dll_dirs_to_path` prepends `nvidia/cudnn/bin/` for — not just for
 onnxruntime's own lazy engine plugins, but for cuDNN itself's later lazy loads.
-Since the PATH has been prepended with the 9.19.0.56 set, ctranslate2's dispatch
-stub resolved its sublibraries from that set, not from its own package directory.
+At runtime, the dispatch stub loaded by ctranslate2 necessarily resolves its
+required sublibraries from whatever is already resident or discoverable in the
+process — in this measurement, the 9.19.0.56 set from `nvidia/cudnn/bin` and/or
+`torch/lib`. However, this measurement did not isolate whether the PATH-prepend
+was the cause (rather than ctranslate2's bare-name lookup simply being satisfied
+by cuDNN sublibraries already resident from onnxruntime or torch initialization
+before ctranslate2 ran).
 
-The result: ctranslate2 loaded a 9.10.2.21 dispatch stub with 9.19.0.56 sublibraries.
-This is a real cross-version shadow (the sublibraries ARE shadowed from a different
-version), but it is narrower in scope than a full "uncontrolled fallback" scenario
-— only the sublibraries are shadowed, not the dispatch stub itself.
+The result: ctranslate2 loaded a 9.10.2.21 dispatch stub that resolved its
+9.19.0.56 sublibraries from the process (the 9.19.0.56 versions being already
+present before ctranslate2's dispatch stub executed). This is a mixed-version
+composition, not a displacement of ctranslate2's own sublibraries — ctranslate2
+bundles no cuDNN sublibraries at all, only the single dispatch DLL entry point.
 
 ### Conclusion
 
-**No fix needed.** The `install-ort.mjs:396-414` PASS-3 review note's concern
-(cross-minor cuDNN version gap 9.10→9.19 between ctranslate2's bundled copy and
-the PATH-prepended copy) manifests as a cuDNN sublibrary shadow on this real
-hardware, but empirically causes no harm: ctranslate2's 9.10.2.21 dispatch stub
-loaded and correctly resolved its 9.19.0.56 sublibraries from the PATH-prepended
-set, executed transcription correctly on CUDA, and produced the correct output.
-The mixed-version stack worked. `_add_nvidia_dll_dirs_to_path` is left untouched,
+**No fix needed.** The `install-ort.mjs` PASS-3 REVIEW NOTE (2026-09-01) concern
+(cross-minor cuDNN version gap 9.10→9.19 between ctranslate2's bundled dispatch
+stub and the system's cuDNN sublibraries) results in a mixed-version composition
+on this real hardware, but empirically causes no harm: ctranslate2's 9.10.2.21
+dispatch stub loaded and correctly resolved its required 9.19.0.56 sublibraries,
+executed transcription correctly on CUDA, and produced the correct output. The
+mixed-version stack worked. `_add_nvidia_dll_dirs_to_path` is left untouched,
 per the ticket's explicit instruction not to weaken it — the prepended PATH
-serves a critical purpose for onnxruntime's own lazy engine loads, and the fact
-that it also causes ctranslate2's sublibraries to be shadowed from a newer version
-is a side effect that does not cause problems in practice on this hardware.
+serves a critical purpose for onnxruntime's own lazy engine loads. Moreover,
+excluding the PATH-prepend for ctranslate2 specifically (a remedy #2845 floated)
+would leave its dispatch stub with zero discoverable sublibraries at all,
+reintroducing the exact `CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED` failure #2818/A28
+already fixed, so the status quo is not just harmless but necessary.
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -54,7 +54,7 @@ Sidecar log:
 INFO:     127.0.0.1:58098 - "POST /transcribe HTTP/1.1" 200 OK
 ```
 
-### 3. Module-provenance inspection: mixed cuDNN versions coexist
+### 3. Module-provenance inspection: inert bundled cuDNN copy coexists with active set
 
 `Get-Process -Id <sidecar pid> -Module` after the `/transcribe` call shows the
 sidecar process has loaded `cudnn64_9.dll` from **three different locations**,
@@ -70,38 +70,57 @@ but only **two distinct versions** — Windows did not collapse them into a sing
 `_add_nvidia_dll_dirs_to_path`'s PATH-prepend mechanism and torch's own bundled
 copy; the `ctranslate2\` one is the package's own bundled 9.10.2.21 copy.)
 
+**ctranslate2's own compiled code does not reference cuDNN at all.** Binary inspection
+of `server/tts-sidecar/.venv/Lib/site-packages/ctranslate2/ctranslate2.dll` and
+`_ext.cp312-win_amd64.pyd` via grep found zero cuDNN references (vs. 3 cuBLAS
+references in the same binary, confirming the grep methodology succeeds when
+cuDNN is actually present). The bundled `cudnn64_9.dll` in the ctranslate2 package
+directory is loaded unconditionally by `ctranslate2/__init__.py`'s glob-based
+directory-wide DLL load at import time (lines 20-21: `for library in
+glob.glob(os.path.join(package_dir, "*.dll")): ctypes.CDLL(library)`), but it is
+loaded inertly — ctranslate2's compiled code never calls into any cuDNN exports
+from this file.
+
 Every other loaded cuDNN component (`cudnn_ops64_9.dll`, `cudnn_adv64_9.dll`,
 `cudnn_graph64_9.dll`, `cudnn_heuristic64_9.dll`, `cudnn_engines_*64_9.dll`) is
 present ONLY under `nvidia\cudnn\bin` and `torch\lib` (version 9.19.0.56) — never
-bundled under `ctranslate2\` — because ctranslate2 only ships the single
-`cudnn64_9.dll` dispatch DLL entry point, not the full cuDNN library suite.
-cuDNN's dispatch DLL loads these sublibraries lazily (at the first real kernel
-operation) via bare-name `LoadLibrary` calls. Every other loaded cuDNN component
-is present in the process from the 9.19.0.56 set. At runtime, ctranslate2's
-dispatch stub resolved its required sublibraries from this 9.19.0.56 set present
-in the process. The result: a mixed-version stack (dispatch 9.10.2.21 +
-sublibraries 9.19.0.56) that worked correctly end-to-end on CUDA with a correct
-transcript.
+bundled under `ctranslate2\` — because ctranslate2 only ships the single dispatch
+DLL entry point, not the full cuDNN library suite. These sublibraries belong to
+the only actual cuDNN consumer in this process: **onnxruntime's CUDA execution
+provider, used by the Kokoro synth call in step 1** — not to ctranslate2/Whisper.
+The 9.19.0.56 sublibraries were loaded by onnxruntime's own cuDNN dispatch stub
+(which resolves them lazily at the first kernel operation via bare-name
+`LoadLibrary` calls), and all of them are present in the process from the
+9.19.0.56 set installed by this sidecar's environment.
 
 ### Interpretation
 
 Windows' loader identifies a module by its **normalized full path**, not by base
-name, when the loading component resolves the DLL via an explicit path. ctranslate2
-ships `cudnn64_9.dll` inside its own package directory and loads it from there,
-not via a bare-name `PATH` search — so its 9.10.2.21 dispatch stub loads as
-a distinct module, unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend.
-
-The dispatch DLL loaded by ctranslate2 (9.10.2.21) resolved its required
-sublibraries from the 9.19.0.56 set present in the process. This is a
-mixed-version composition, not a displacement of ctranslate2's own sublibraries —
-ctranslate2 bundles no cuDNN sublibraries at all, only the single dispatch DLL
-entry point.
+name. ctranslate2 ships `cudnn64_9.dll` in its own package directory and loads it
+via an absolute path (the glob-based unconditional load in `__init__.py`), not
+via a bare-name `PATH` search — so the 9.10.2.21 copy loads as a distinct module,
+unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend. However, ctranslate2's
+own compiled code does not call into this loaded DLL; the file is inert within
+ctranslate2's execution path. The only actual cuDNN consumer in this process is
+onnxruntime's CUDA execution provider (Kokoro), which loaded the 9.19.0.56
+sublibraries via its own lazy-loading dispatch mechanism.
 
 ### Conclusion
 
-**Empirically harmless. No fix needed.** The cross-minor cuDNN version gap (9.10→9.19)
-between ctranslate2's bundled dispatch stub and the system's cuDNN sublibraries
-results in a mixed-version composition on this real hardware. The measurement
-shows this mixed-version stack works correctly: ctranslate2's 9.10.2.21 dispatch
-stub loaded and correctly resolved its required 9.19.0.56 sublibraries, executed
-transcription correctly on CUDA, and produced the correct output.
+**No fix needed.** ctranslate2's own compiled code contains no cuDNN references
+at all (verified via binary grep: 0 hits in both `ctranslate2.dll` and
+`_ext.cp312-win_amd64.pyd`); the bundled `cudnn64_9.dll` is loaded inertly and
+not called by ctranslate2's execution path. The original concern (#2845) — whether
+a cross-minor cuDNN version gap (9.10 vs. 9.19) could affect ctranslate2/Whisper
+transcription — does not apply: there is no gap in practice because there is no
+consumption to be affected by a gap.
+
+The 9.19.0.56 cuDNN sublibraries loaded in this process belong to onnxruntime's
+CUDA execution provider (Kokoro), the only actual cuDNN consumer in this
+measurement. Transcription succeeded correctly on CUDA, confirming the setup is
+functional end-to-end.
+
+**Forward-looking caveat**: this conclusion is specific to ctranslate2 4.8.0's
+current build and the versions present in this sidecar's environment. If a future
+version of ctranslate2 adds cuDNN-based GPU kernels (unlikely but not ruled out
+by this measurement), this analysis would need re-checking at that time.
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -1,4 +1,4 @@
-# ctranslate2/cuDNN shadow measurement (Castwright#2861)
+# ctranslate2/cuDNN shadow measurement (Castwright#2845)
 
 ## Result: harmless. No source change made.
 
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -1,0 +1,106 @@
+# ctranslate2/cuDNN shadow measurement (Castwright#2861)
+
+## Result: harmless. No source change made.
+
+Measured live on this box's real GPU (NVIDIA GeForce RTX 4070 Laptop GPU, 8585 MB,
+driver 610.88 / CUDA UMD 13.3) with the sidecar booted `ASR_DEVICE=cuda ASR_MODEL=base`.
+(Note: this box's second GPU slot, `GPU1: 0000:05:00.0`, independently reported
+"GPU is lost" at the time of this run — a pre-existing, unrelated hardware issue
+requiring a reboot; GPU 0, the one this sidecar actually uses, was healthy
+throughout.)
+
+### 1. `/synthesize` (kokoro) → real speech PCM
+
+```
+POST /synthesize {"engine":"kokoro","model":"kokoro-v1.0","voice":"af_heart",
+  "text":"The quick brown fox jumps over the lazy dog."}
+-> 200 OK, X-Sample-Rate: 24000, 128000 bytes raw PCM (2.667 s)
+```
+
+Sidecar log:
+```
+2026-09-03 08:03:29.644 [sidecar] Loading Kokoro model=...\voices\kokoro\kokoro-v1.0.onnx ...
+2026-09-03 08:03:31.438 [sidecar] Kokoro loaded. English voices: 28 (filtered from 54 total in manifest).
+2026-09-03 08:03:35.285 [sidecar] kokoro synth: voice=af_heart text_len=44 gen_ms=3842 audio_ms=2667 rtf=1.44
+INFO:     127.0.0.1:51380 - "POST /synthesize HTTP/1.1" 200 OK
+```
+
+### 2. `/transcribe` (ctranslate2/faster-whisper, cold load) → correct transcript on CUDA
+
+Fed the kokoro PCM straight back into `/transcribe` (`X-Sample-Rate: 24000`), which
+triggers ctranslate2's first cuDNN load of the process (Whisper's model is lazy-loaded,
+after onnxruntime/kokoro has already loaded its own cuDNN 9.19.0.56 copy via
+`_add_nvidia_dll_dirs_to_path`).
+
+```
+POST /transcribe (raw int16 PCM, X-Sample-Rate: 24000)
+-> 200 OK
+{"text":"The quick brown fox jumps over the lazy dog.","language":"en",
+ "avg_logprob":-0.236,"no_speech_prob":0.0061,"compression_ratio":0.863,"words":null}
+```
+
+Transcript is an exact match to the synthesized text. No CUDA/cuDNN error, no
+`CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED`-shaped message, no silent CPU fallback.
+
+Sidecar log:
+```
+2026-09-03 08:04:07.373 [sidecar] Loading Whisper ASR model=base device=cuda:0 compute=int8_float16 revision=(unpinned) ...
+2026-09-03 08:04:10.274 [sidecar] Whisper ASR loaded (model=base device=cuda:0).
+2026-09-03 08:04:10.274 [sidecar] Processing audio with duration 00:02.667
+2026-09-03 08:04:10.321 [sidecar] VAD filter removed 00:00.000 of audio
+2026-09-03 08:04:10.853 [sidecar] Detected language 'en' with probability 1.00
+INFO:     127.0.0.1:58098 - "POST /transcribe HTTP/1.1" 200 OK
+```
+
+### 3. Module-provenance inspection: TWO distinct `cudnn64_9.dll` modules coexist
+
+`Get-Process -Id <sidecar pid> -Module` after the `/transcribe` call shows the
+sidecar process has **two separate loaded modules both named `cudnn64_9.dll`**,
+at different full paths, with different versions — Windows did not collapse them
+into a single module:
+
+| Path | FileVersion |
+|---|---|
+| `...\.venv\Lib\site-packages\nvidia\cudnn\bin\cudnn64_9.dll` | 9.19.0.56 |
+| `...\.venv\Lib\site-packages\torch\lib\cudnn64_9.dll` | 9.19.0.56 |
+| `...\.venv\Lib\site-packages\ctranslate2\cudnn64_9.dll` | **9.10.2.21** |
+
+(the `nvidia\cudnn\bin` and `torch\lib` copies are both 9.19.0.56 as expected —
+`_add_nvidia_dll_dirs_to_path`'s PATH-prepend mechanism and torch's own bundled
+copy; the `ctranslate2\` one is the package's own bundled 9.10.2.21 copy.)
+
+Every other loaded cuDNN component (`cudnn_ops64_9.dll`, `cudnn_adv64_9.dll`,
+`cudnn_graph64_9.dll`, `cudnn_heuristic64_9.dll`, `cudnn_engines_*64_9.dll`) is
+present ONLY under `nvidia\cudnn\bin` and `torch\lib` (version 9.19.0.56) — never
+duplicated under `ctranslate2\` — because ctranslate2 only bundles the single
+`cudnn64_9.dll` entry-point stub, not the full cuDNN component set; its runtime
+loader resolves the rest through whichever `cudnn64_9.dll` gets loaded first in
+its own process, or ctranslate2 links narrowly against the entry stub only. What
+matters for this measurement is that ctranslate2's forward pass completed
+correctly end-to-end on CUDA with a correct transcript, so whichever component
+set it actually used at runtime worked.
+
+### Interpretation
+
+Windows' loader identifies a already-loaded module by its **normalized full
+path**, not by base name, when the loading component resolves the DLL via an
+explicit path (as ctranslate2 does for its own bundled copy — it ships
+`cudnn64_9.dll` inside its own package directory and loads it from there, not
+via a bare-name `PATH` search). `_add_nvidia_dll_dirs_to_path`'s PATH prepend
+only affects **bare-name** `LoadLibrary`/`dlopen` resolution (which is what
+onnxruntime's lazily-dlopened cuDNN engine plugins need — the bug it was built
+to fix). ctranslate2 never does a bare-name search for its own `cudnn64_9.dll`,
+so the PATH-prepend candidate never gets a chance to shadow it: both DLLs load
+side by side as distinct modules, ctranslate2 gets its own pinned 9.10.2.21, and
+the onnxruntime-consuming engines (Kokoro/Coqui/Qwen) still get 9.19.0.56 from
+the PATH prepend. There is no shadow in practice on this measurement.
+
+### Conclusion
+
+**No fix needed.** The `install-ort.mjs:396-414` PASS-3 review note's concern
+(cross-minor cuDNN version gap 9.10→9.19 between ctranslate2's bundled copy and
+the PATH-prepended onnxruntime copy) does not manifest as a runtime shadow on
+this real hardware: ctranslate2 loads and uses its own bundled cuDNN 9.10.2.21
+via its own absolute package path, unaffected by `_add_nvidia_dll_dirs_to_path`'s
+PATH prepend, and transcription is correct. `_add_nvidia_dll_dirs_to_path` is
+left untouched, per the ticket's explicit instruction not to weaken it.
--- a/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
+++ b/server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md
@@ -1,6 +1,6 @@
 # ctranslate2/cuDNN shadow measurement (Castwright#2845)
 
-## Result: harmless. No source change made.
+## Result: not applicable. ctranslate2 does not consume cuDNN. No source change made.
 
 Measured live on this box's real GPU (NVIDIA GeForce RTX 4070 Laptop GPU, 8585 MB,
 driver 610.88 / CUDA UMD 13.3) with the sidecar booted `ASR_DEVICE=cuda ASR_MODEL=base`.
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
## PR review — pass 1 (head df8d78e4db640fd77e986478ca781021044f8a05, depth low)

Scope note: this PR is docs-only in *content*, but its path (`server/tts-sidecar/docs/…`) is outside CONTRIBUTING.md's doc-only globs (`docs/**`, root `*.md`, `.github/*.md`), so the gate's docs-only exemption does not apply. Reviewed at depth `low` per the routing table.

### 🔴 Blocking

None.

### 🟠 Major — correctness

**M1 — The Interpretation and Conclusion are contradicted by the doc's own §3 data and by `main.py`'s own on-hardware docstring.** `server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:89-96` and `:100-105`.

The claims: *"ctranslate2 never does a bare-name search for its own `cudnn64_9.dll`, so the PATH-prepend candidate never gets a chance to shadow it … ctranslate2 gets its own pinned 9.10.2.21"* and *"ctranslate2 loads and uses its own bundled cuDNN 9.10.2.21 via its own absolute package path, unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend."*

Independently checked against this box's live venv:

```
site-packages\ctranslate2\       -> cudnn64_9.dll ONLY                        (9.10.2.21)
site-packages\nvidia\cudnn\bin\  -> cudnn64_9.dll                             (9.19.0.56)
                                    cudnn_graph64_9.dll, cudnn_ops64_9.dll,
                                    cudnn_cnn64_9.dll (9.19.0.56), cudnn_adv64_9.dll,
                                    cudnn_heuristic64_9.dll, cudnn_engines_*64_9.dll
```

ctranslate2 ships the cuDNN 9 **dispatch** DLL and nothing else — which §3:72-78 itself observes. cuDNN 9's dispatch DLL loads its sublibraries **lazily, by bare name, through the process `PATH`**. That is not my inference; it is what `server/tts-sidecar/main.py:165-175` records as confirmed on this same hardware: *"`os.add_dll_directory()` … does NOT fix this — confirmed on the same box: cuDNN's own lazy loader does not honour `AddDllDirectory`-registered search paths, only the process `PATH` environment variable."* And `_add_nvidia_dll_dirs_to_path`, called unconditionally at every boot (`main.py:468-471`), puts `nvidia/cudnn/bin` — the 9.19.0.56 set — at the **front** of `PATH`.

So the configuration this run actually exercised is a **9.10.2.21 dispatch DLL over a 9.19.0.56 sublibrary set** — a cross-minor mix, i.e. a narrower instance of exactly what #2845 raised, not its absence. §3:76-81 reaches the edge of this, offers two mutually exclusive hypotheses (*"resolves the rest through whichever `cudnn64_9.dll` gets loaded first … **or** ctranslate2 links narrowly against the entry stub only"*), resolves neither, and the Interpretation then proceeds as though the second were established. The Interpretation also attributes the lazy dlopen to onnxruntime (`:90-92`, "onnxruntime's lazily-dlopened cuDNN engine plugins"); `main.py:151-155` attributes it to **cuDNN itself** — and that difference is precisely why the mechanism reaches ctranslate2 too.

Failure scenario: someone hits a cuDNN fault on the ASR path in three months, reads this doc as the record of record, sees "measured harmless — ctranslate2 uses its own 9.10.2.21, unaffected by the PATH prepend", and rules the version mix out of the hypothesis space. The mix is what is running.

What the run *does* support, and is worth keeping verbatim: the **entry-point** DLL is not shadowed (three distinct same-basename modules coexist), Whisper loaded on `cuda:0`, and a `base`/`int8_float16` forward over 2.667 s of audio returned an exact transcript. What it does not support is that ctranslate2 runs a pure 9.10.2.21 stack. Restating the Interpretation and Conclusion around the mixed stack — it worked in this configuration, with the mix named as the thing that worked, and the tested surface (one tiny model, 2.7 s, English) stated — closes #2845 just as well and does not mislead the next reader.

### 🟡 Minor / cleanup

- **C1 — "TWO" vs a three-row table.** `:55` and `:58` say two distinct `cudnn64_9.dll` modules; `:62-66` lists three. Worth fixing beyond typography: three same-basename modules, two of them at the *same* version from different directories, is stronger evidence for path-keyed module identity than two.
- **C2 — The load-order precondition is asserted, not measured.** `:30-33` states the `/transcribe` happened *"after onnxruntime/kokoro has already loaded its own cuDNN 9.19.0.56 copy"*. `Get-Process -Module` was captured only *after* the transcribe; there is no before-snapshot, and no `Device probe complete: {…}` line in the pasted logs, so nothing here shows Kokoro ran on CUDA at all (`rtf=1.44` — 3842 ms of generation for 2667 ms of audio, *after* the model-load lines — is not self-evidently a 4070 CUDA figure). The precondition is very likely true, for a reason the doc doesn't give: `_startup_ort_cuda_preload` (`main.py:468-471`) runs the PATH prepend and then `preload_dlls()` at every boot, unconditionally, so `nvidia\cudnn\bin\cudnn64_9.dll` is resident before any request arrives. Citing that, plus the `Device probe complete:` line the sidecar already emits at `main.py:10480`, converts an assertion into evidence at roughly zero cost.
- **C3 — A comment this PR makes false.** `server/tts-sidecar/scripts/install-ort.mjs:409-410` still reads *"so this shadow is unconfirmed either way in practice — filed as #2845 rather than assumed benign."* Merging closes #2845, and nothing in the tree points from that PASS-3 note to the new measurement doc. CLAUDE.md's incidental-findings rule names "a stale comment or doc sentence your change made false" as a chore fixed in the same round, not filed — one line plus a link.
- **C4 — Discoverability and title.** The file opens a new `server/tts-sidecar/docs/` tree (`git ls-files server/tts-sidecar/docs/` was empty before this PR) that nothing indexes, and its title cites `Castwright#2861` — the agent-runtime task ticket — rather than #2845, the risk it closes. Fixing C3's cross-link mostly settles the first half; retitling to #2845 (keeping #2861 as a provenance line) makes the doc findable from the ticket a reader would actually search.
- **C5 — Typo.** `:85` "identifies a already-loaded module" → "an already-loaded module".
- **C6 — House-gate declarations absent from the PR body.** No register row references #2845 (`git grep 2845 -- docs/` finds only an unrelated `analysis.ts:2845` line-reference in a design spec), and A28 was discharged in wave 10 — so no on-box acceptance is being discharged and none is newly owed. Likewise the change has no shippable delta, so the release-notes pair is legitimately skippable. CLAUDE.md steps 3 and 5 ask for that to be *stated*, not silently omitted. Two sentences.

### ✅ Checked and clean

- `Closes #2845` present in the body, outside any code span — `pr-issue-link` satisfied.
- **Every raw measurement in the doc is reproducible.** I re-read `FileVersion` on this box and got exactly the doc's three numbers: `ctranslate2\cudnn64_9.dll` 9.10.2.21, `nvidia\cudnn\bin\cudnn64_9.dll` and `torch\lib\cudnn64_9.dll` both 9.19.0.56. The data is trustworthy; M1 is a defect in the reading of it, not in the numbers.
- `install-ort.mjs:396-414` does contain the PASS-3 note's ctranslate2/cuDNN bullet (`:401-410`). The cited range starts ~10 lines into the note rather than at its head (`:386`), but covers the concern quoted. Not a finding.
- `_add_nvidia_dll_dirs_to_path` is genuinely untouched: the PR adds one file and changes nothing else (1 file, +106/−0), per #2861's explicit instruction not to weaken it.
- No paired test owed — there is no source change, and the measurement itself is the artefact #2861's acceptance bar asks for, reported verbatim rather than summarised into a bare verdict.
- No `cast.json` writes, no new config knob, no derived artifact, no `docs/features/` plan touched.

### Verdict

**Changes requested.** One correctness-class finding (**M1**): the doc's stated causal conclusion is contradicted by its own §3 evidence and by `main.py:165-175`, and this doc is the record #2845 closes on. The empirical result survives intact; the interpretation wrapped around it needs restating. M1 plus **C3** (a comment this PR makes false — owed this round under CLAUDE.md's incidental-findings rule, not deferrable) should land before merge; C1/C2/C4/C5/C6 are cheap and mostly in the same file. M1 being correctness-class re-triggers this gate once fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head ed3a9e66c156d1298fce03df1c53019a5b9d377a, depth low)

Depth re-derived from the current commit set: still single-scope (`chore(ops)` title / `docs(sidecar)` commits, two files, one scope) → **low**. Not docs-only-exempt — `server/tts-sidecar/scripts/install-ort.mjs` is outside the doc-glob fast-path, so the gate applies.

Deltas only; pass-1 items confirmed solid are not re-listed.

| # | Disposition |
|---|---|
| 🟠 M1 — doc concluded "no shadow in practice" against its own data | **PARTIALLY RESOLVED — still open** (see F1) |
| 🟡 C1 — "TWO modules" vs 3-row table | **VERIFIED RESOLVED** |
| 🟡 C2 — unsupported load-order claim | **RESOLVED**, residual folded into F1 |
| 🟡 C3 — stale `install-ort.mjs` comment | **RESOLVED**, inherits F1's wording |
| 🟡 C4 — heading cited #2861 | **VERIFIED RESOLVED** |
| 🟡 C5 — typo "a already-loaded" | **VERIFIED RESOLVED** |
| 🟡 C6 — PR body missing N/A notes | **STILL OPEN** (see F3) |

**C1** — heading is now "mixed cuDNN versions coexist" and the body reads "**three different locations** … but only **two distinct versions**", which matches the 3-row table. Correct.

**C4** — heading is `(Castwright#2845)`; `git grep 2861` over both changed files returns nothing. The "missing docs index" half is **not owed**: `server/tts-sidecar/docs/` is new and holds one file, the repo's only index convention (`docs/features/INDEX.md`) covers feature plans rather than sidecar notes, and `install-ort.mjs:410-411` now links the doc, so it is reachable. Non-blocking note: commit `df8d78e4`'s message headline still reads `(#2861)` and will ride into the merge commit — not fixable without a history rewrite, flagged only so it isn't mistaken for a live citation.

**C5** — the sentence carrying the typo was rewritten wholesale. The replacement ("Windows' loader identifies a module by its **normalized full path**, not by base name, when the loading component resolves the DLL via an explicit path") is correctly scoped by its trailing clause; dropping "already-loaded" did not break it.

**C2** — the hedge asked for is present (`:32-35`, "this run did not separately capture a pre-`/transcribe` module snapshot to confirm it directly"). Two residuals, both folded into F1: the "given `_add_nvidia_dll_dirs_to_path` runs unconditionally at sidecar boot" clause is a non-sequitur as grounds for *Kokoro having loaded cuDNN* (the PATH prepend says nothing about whether onnxruntime initialised its CUDA EP, and §1's quoted log shows no device for Kokoro); and §3/Interpretation then silently rely on the very ordering §2 concedes is unconfirmed.

---

### Open findings

#### 🟠 F1 (correctness) — M1's replacement text asserts a shadow that the on-disk layout rules out, and the PR body still carries the retracted conclusion

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:82-87`, `:104-110`; `server/tts-sidecar/scripts/install-ort.mjs:406-408`; PR body `## Summary` ¶2.

The half of M1 that is fixed: the doc no longer claims immunity, and correctly states the process ends up running a 9.10.2.21 dispatch stub over 9.19.0.56 sublibraries. I verified the two load-bearing mechanism claims directly and both hold:

- `ctranslate2/__init__.py` ends with `for library in glob.glob(os.path.join(package_dir, "*.dll")): ctypes.CDLL(library)` — an **absolute-path** load, so "resolves via an explicit path, not a bare-name `PATH` search" is literally true and the dispatch DLL genuinely is immune to the prepend.
- `main.py:151-175` does document the dispatch/lazy-sublibrary pattern the doc cites it for, including "cuDNN's own lazy loader … does not honour `AddDllDirectory`-registered search paths, only the process `PATH` environment variable."

The half that is not fixed is the word the rewrite chose. On this box:

```
.venv/Lib/site-packages/ctranslate2/  ->  ctranslate2.dll, cudnn64_9.dll, libiomp5md.dll
.venv/Lib/site-packages/nvidia/cudnn/bin/ -> cudnn64_9.dll + cudnn_{ops,cnn,adv,graph,heuristic,engines_*}64_9.dll
```

ctranslate2 ships **no cuDNN sublibraries at all**. So `:108-110`'s "This is a real cross-version shadow (the sublibraries ARE shadowed from a different version)" and `:104-105`'s "resolved its sublibraries from that set, **not from its own package directory**" both assert a displacement that cannot occur — there is no 9.10.2.21 sublibrary set in that package directory for the PATH-prepended copies to shadow. The same wording is now mirrored into `install-ort.mjs:406-408` ("This cuDNN sublibrary shadow DOES occur"). What the measurement actually shows is a **mixed-version composition**, which the doc says correctly everywhere except the sentences the M1 fix wrote.

Two things follow, and they are why this is correctness rather than wording:

1. **The causal attribution is unproven, and §2 already admits the fact it depends on.** `:80-84` and `:104-105` attribute the 9.19.0.56 sublibraries to the PATH prepend. But the module table shows `cudnn_ops64_9.dll` &c. loaded from `nvidia\cudnn\bin` *and* `torch\lib` simultaneously, i.e. at least one was loaded by explicit path (torch uses the same `ctypes.CDLL` pattern). A later bare-name `LoadLibrary("cudnn_ops64_9.dll")` from ctranslate2's dispatch stub returns an **already-resident module by base name without any path search** — so if onnxruntime/torch got there first, the prepend played no causal role for ctranslate2 at all. The doc cannot tell these two apart, and `:32-35` says so in as many words. Asserting the prepend as the cause is the same shape M1 named — a conclusion firmer than the run supports — relocated one section down.
2. **Concrete failure scenario.** A reader of `install-ort.mjs:406-408` ("this cuDNN sublibrary shadow DOES occur") who wants to de-shadow ctranslate2 will reach for the remedy #2845's own body proposes — "exclude `ctranslate2`'s own cudnn copy from the PATH-prepend candidate list". Do that and ctranslate2's dispatch stub is left with **zero** discoverable sublibraries: none in its own package dir, and per `main.py:165-171` its lazy loader ignores the `add_dll_directory` entries `ctranslate2/__init__.py` registers, honouring only `PATH`. That reintroduces `CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED` — the exact failure #2818/A28 fixed. The truthful framing inverts the risk: the prepend is not a hazard to ctranslate2, it is what makes ctranslate2's CUDA path resolvable at all, and that is a *stronger* argument for the PR's "leave it untouched" verdict than the one currently written.

**Also unfixed: the PR body still tells the pre-M1 story verbatim.** `## Summary` ¶2 still ends "…so the PATH prepend (which only affects bare-name resolution) **never gets a chance to shadow it**", and still says "the **two** `cudnn64_9.dll` copies" against the doc's corrected three-locations/two-versions. Same ¶1 says "Result: harmless, **no source change**" — `install-ort.mjs` is now modified (+7/−5). The body is what closes #2845 and what lands in the merge commit, so as it stands the ticket would be answered with the conclusion this PR retracted.

Suggested resolution — small, and the doc half is the only part needing a push:
- replace "the sublibraries ARE shadowed" / "not from its own package directory" with the composition framing, and state that ctranslate2 bundles only the dispatch DLL so the prepended set is the *supplier*, not a displacer;
- either drop the causal "Since the PATH has been prepended … so the sublibraries loaded are …", or note the already-resident-module alternative the run cannot exclude;
- mirror the same wording into `install-ort.mjs:406-408`;
- `gh pr edit` the body: correct ¶2's conclusion, "two copies" → three locations / two versions, and "no source change" → "no behaviour change; one comment-only edit to `install-ort.mjs`".

#### 🟡 F2 (cleanup) — stale line-range citation, made staler by this PR's own commit

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:114` cites "`install-ort.mjs:396-414` PASS-3 review note". The note actually spans **386–416** (`PASS-3 REVIEW NOTE (2026-09-01)` at 386, `pin the packages this file DOES own.` at 416). The start was already wrong at `df8d78e4`; commit `d2bbbe4e` added two net lines and moved the end from 414 to 416, so the PR's own fix widened the drift. Failure scenario: a reader jumping to 396-414 lands mid-block and past its close, missing both the framing line and the newly-added measurement conclusion at 406-412. Cheapest durable fix is to cite the block by its `PASS-3 REVIEW NOTE (2026-09-01)` marker rather than by line numbers, which rot on every edit.

#### 🟡 F3 (cleanup) — C6 still open: PR body carries no on-box-register or release-notes disposition

PR body. CLAUDE.md's before-shipping steps 3 and 5 require the PR to say "not applicable" explicitly rather than silently omit. I checked both and agree N/A is the right call, so this is a one-line-each statement, not work:

- **On-box register** — `git grep 2845 docs/testing/onbox-acceptance-register.md` returns no row, so no acceptance was owed for this ticket; the PR runs a real-GPU measurement and discharges #2845's acceptance criterion inside its own PR, and the register tracks owed state rather than history, so no row is added or removed. Say that.
- **Release notes** — docs plus one code comment, no user- or operator-visible delta. Say that.

Issue linkage is fine: `Closes #2845.` is present in prose, outside any code span.

---

### Verdict

**Still blocking.** One 🟠 (F1 — M1 only half-discharged: the doc's replacement sentence asserts a displacement the on-disk layout rules out and a cause the run cannot establish, and the PR body still states the retracted conclusion) plus two 🟡. F1's doc half needs a commit; its PR-body half and F3 are `gh pr edit` only. Everything else from pass 1 is verified resolved.

_Tree unmodified during this pass: `HEAD ed3a9e66c156d1298fce03df1c53019a5b9d377a`, `git status --porcelain` empty before and after._

dudarenok-maker:
## PR review — pass 3 (head 62e8ced9fc6cf39e9464d1c9ad8cc16c763812e9, depth low)

Deltas on the pass-2 findings, plus new findings. **Verdict: still blocking.**

---

### F1 (🟠 correctness) — NOT fixed. The correction introduced a third false claim.

`62e8ced9` retracted the "sublibraries shadowed from a different version" overclaim, which was correct to do. But it rewrote only the **Interpretation** and **Conclusion** sections, and in doing so left the doc asserting *both* candidate mechanisms as established fact in three different places while simultaneously saying neither was isolated — and it added a new counterfactual that is **demonstrably false on this box**.

**F1a — the retracted framing is still live in §3.**
`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:82-84`:

> "This process `PATH` has been prepended with `nvidia/cudnn/bin/` at sidecar boot (via `_add_nvidia_dll_dirs_to_path`), **so the sublibraries loaded are the 9.19.0.56 versions.** The net result: ctranslate2 loaded the 9.10.2.21 dispatch stub (via its explicit package path), but **resolved its required sublibraries from the PATH-prepended 9.19.0.56 set.**"

That is a flat causal attribution to the PATH-prepend. `:107-110` — 25 lines later — says the opposite:

> "However, this measurement **did not isolate whether the PATH-prepend was the cause** (rather than ctranslate2's bare-name lookup simply being satisfied by cuDNN sublibraries already resident from onnxruntime or torch initialization before ctranslate2 ran)."

The diff for `62e8ced9` starts at line ~101, so §3 was never touched. A reader stopping at §3 — the section that actually presents the measurement — gets exactly the claim pass 2 asked you to withdraw.

**F1b — the *other* branch is then asserted as fact too.**
`:113-114`, one sentence after the "did not isolate" disclaimer:

> "…resolved its 9.19.0.56 sublibraries from the process (**the 9.19.0.56 versions being already present before ctranslate2's dispatch stub executed**)."

This picks the second of the two branches `:107-110` just said were not distinguished, and states it as fact. It is also contradicted by the doc's own §2 hedge at `:34-35`: "this run **did not separately capture a pre-`/transcribe` module snapshot to confirm it directly.**" So the doc asserts mechanism A at `:82-84`, mechanism B at `:113-114`, and "we couldn't tell which" at `:107-110`. All three cannot stand.

**F1c — "zero discoverable sublibraries at all" is false, and it is the load-bearing premise of the new verdict.**
`:129-132`, replicated verbatim into the permanent code comment at `server/tts-sidecar/scripts/install-ort.mjs:415-419` and into the PR body:

> "excluding the PATH-prepend for ctranslate2 specifically (a remedy #2845 floated) would leave its dispatch stub with **zero discoverable sublibraries at all**, reintroducing the exact `CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED` failure #2818/A28 already fixed, so the status quo is **not just harmless but necessary**."

Checked against this box's live venv. `site-packages/torch/lib/` holds a **complete** cuDNN sublibrary set:

```
cudnn_adv64_9.dll                     269,016,688
cudnn_cnn64_9.dll                       2,984,560
cudnn_engines_precompiled64_9.dll     481,015,408
cudnn_engines_runtime_compiled64_9.dll 27,759,216
cudnn_graph64_9.dll                     2,493,040
cudnn_heuristic64_9.dll                58,874,992
cudnn_ops64_9.dll                     105,604,208
```

— byte-for-byte the same sizes as the `nvidia/cudnn/bin/` copies, i.e. the same 9.19.0.56 build. And per the doc's **own module table** (`:74-76`), those `torch\lib` copies are *already loaded in the live process*. `install-ort.mjs:274-278` documents the governing Windows rule, in this repo's own words, as the root cause of the A28 regression:

> "Windows resolves a later `LoadLibrary` for the same DLL basename to whichever copy is **ALREADY loaded**"

So a bare-name `LoadLibrary("cudnn_ops64_9.dll")` from ctranslate2's stub resolves to the resident `torch/lib` module **regardless of PATH**. Removing the `nvidia/cudnn/bin` prepend does not yield "zero discoverable sublibraries"; it yields, at minimum, the full already-resident torch set at the identical version.

The internal-consistency form of this needs no external fact at all: `:107-110` refuses to exclude the "already resident from onnxruntime or torch initialization" branch. **If that branch is the live one, the counterfactual at `:130` is false.** The doc cannot both decline to rule that branch out and assert the conclusion that branch falsifies.

Separately, the `CUDNN_STATUS_SUBLIBRARY_LOADING_FAILED` attribution is an extrapolation. `main.py:156-163` documents that failure for **onnxruntime/Kokoro**, on the **engine plugin** DLLs specifically (`cudnn_engines_tensor_ir64_9.dll` / "Could not locate cudnn_engines_tensor_ir64_9.dll"). ctranslate2 was not in that measurement at all, so nothing can be "reintroduced" for it — this is a new failure prediction for an unmeasured path, presented with the same unhedged confidence as the two claims already withdrawn.

**Concrete failure scenario.** `install-ort.mjs`'s surrounding comment explicitly instructs the next maintainer to "Re-check this comment, not just the four constants below, on any future torch bump" (`:383-384`). That maintainer reads `:415-419`, concludes the process-wide PATH-prepend is load-bearing for ctranslate2 and must never be scoped or narrowed, and forecloses a legitimate remedy on a premise that is false on the very hardware the doc measured. This is the same shape as M1 and F1 — a causal story stated at a confidence the measurement does not support — now on its third iteration, and this time baked into a permanent code comment rather than only a doc.

**What would fix it:** state only what was measured. The stub is 9.10.2.21, the resolved sublibraries were 9.19.0.56, the composition worked end to end, and the supply route was not isolated. That alone fully supports "no fix needed / leave it untouched" (the ticket instructs it, and the measurement shows no harm). The "not just harmless but **necessary**" upgrade is what requires the counterfactual, and the counterfactual was never measured — drop it, in all three places (doc `:129-132`, `install-ort.mjs:415-419`, PR body), and align `:82-84` and `:113-114` with `:107-110`.

---

### F2 (🟡 cleanup) — Fixed. ✅

`…measurement.md:120` now cites `` `install-ort.mjs` PASS-3 REVIEW NOTE (2026-09-01) `` instead of a line range. Verified findable verbatim: `install-ort.mjs:386` reads `// PASS-3 REVIEW NOTE (2026-09-01): the prior version of this note checked`. Marker-text citation is the right call and will not re-drift.

---

### C6 (🟡 cleanup) — Fixed. ✅

The PR body now carries an explicit `## On-box acceptance / release notes` section with both dispositions, and both check out:
- **Register:** `git grep 2845 -- docs/testing/onbox-acceptance-register.md` → no matches. "no existing register row references #2845" is accurate; nothing to discharge, nothing owed (the PR ships no behaviour, and the hardware criterion is exercised and recorded in the doc).
- **Release notes:** N/A for a docs/comment-only change is consistent with CLAUDE.md before-shipping step 5's carve-out, and is stated explicitly rather than silently omitted.

---

### New findings

**N1 (🟡 cleanup) — one instance of F2's class left armed.** `…measurement.md:98` still cites `` `main.py:151-175` `` by line number — the exact drift shape F2 was about, fixed in one place and not the other. It resolves correctly *today* (151-175 falls inside `_add_nvidia_dll_dirs_to_path`'s docstring and does contain the lazy-dlopen description), so this is not currently wrong, but it will rot the same way the `install-ort.mjs:396-414` citation did. Cite `_add_nvidia_dll_dirs_to_path`'s docstring by symbol name.

**N2 (🟡 cleanup) — the linchpin fact is true, but the evidence offered doesn't establish it.** "ctranslate2 bundles no cuDNN sublibraries at all" (`:115-116`, and `install-ort.mjs:413-415`) is the single fact the whole F1 correction turns on. I verified it independently and it **holds**: `site-packages/ctranslate2/` contains exactly one cuDNN artifact, `cudnn64_9.dll` (266,288 bytes), and no `cudnn_*64_9.dll` sublibraries. But the doc's stated basis for it (`:73-77`) is the **loaded-module** table, which cannot distinguish "not bundled on disk" from "bundled but not loaded" — a module shadowed by an already-resident copy is absent from that table in exactly the same way. Cite the directory listing you actually rest on, not the module snapshot.

**N3 (🟡 cleanup) — mechanism generalised past its source.** `:98-100` attributes bare-name lazy `LoadLibrary` loading to `cudnn_ops64_9.dll` / `cudnn_graph64_9.dll` and cites `main.py` for it. `main.py:151-163` documents that mechanism for cuDNN 9's **engine plugin** DLLs (`cudnn_engines_tensor_ir64_9.dll`, `cudnn_engines_precompiled64_9.dll`), and explicitly contrasts them with "the single `cudnn64_9.dll` entry point"; it says nothing about ops/graph. Plausible, but not what the cited source says.

---

### Retraction sweep (requested)

`grep -iE "shadow|displac"` across both changed files: **clean**. Remaining hits are the ticket's own name (`:1` title), a correctly-framed negation (`:115` "not a displacement"), and pre-existing `install-ort.mjs` PASS-2/PASS-3 note text (`:372`, `:388`, `:413`, `:415`, `:420`). No leftover "no shadow occurs" or "sublibraries ARE shadowed from a different version" text survives. The PR body matches the doc's current content — including, unfortunately, F1c's false claim.

---

### Verdict: 🔴 STILL BLOCKING — and the loop cap is reached.

F1 is a correctness finding, not a nit: the doc's central causal account contradicts itself three ways (F1a/F1b), and the "not just harmless but necessary" verdict rests on a counterfactual that this box's own on-disk layout and the doc's own module table falsify (F1c). It is now replicated into a permanent `install-ort.mjs` comment and the PR body.

**This is pass 3 of 3 — the re-review loop cap under this repo's gate. It must go to the user rather than another automatic fix round.** Noting for that handoff: three consecutive passes have found the same defect shape (a causal mechanism asserted beyond what the measurement isolated), and each fix round corrected the previous wording while introducing a fresh unsupported claim in its place. That pattern is itself the signal — the remedy is likely to *remove* the causal narrative and the necessity verdict rather than to re-word them a fourth time. The measurement result ("mixed-version composition, worked correctly end to end on real hardware, leave `_add_nvidia_dll_dirs_to_path` untouched") is sound and needs none of the contested claims to stand.

dudarenok-maker:
## PR review — pass 4 (head fe5cdda7e012ce15f5284e28d963fa3cf2c1d2c3, depth low)

Scope as dispatched: deltas on pass 3's F1a/F1b/F1c, plus mutual consistency across the three surfaces (doc / `install-ort.mjs` comment / PR body) and fresh-eyes internal coherence. Read-only pass; nothing modified (`git status --porcelain` clean before and after, HEAD unmoved).

### Pass-3 finding deltas

**F1a — PATH-prepend asserted as the cause in one section, disclaimed in another: RESOLVED.** The doc no longer attributes ctranslate2's sublibrary resolution to the PATH prepend anywhere. One residual attribution survives at `server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:69-71` — "(the `nvidia\cudnn\bin` and `torch\lib` copies are both 9.19.0.56 as expected — `_add_nvidia_dll_dirs_to_path`'s PATH-prepend mechanism and torch's own bundled copy…)" — but that is scoped to why *onnxruntime's* dispatch copy is resident, not to ctranslate2, and does not contradict the rest. Acceptable as written.

**F1c — "zero discoverable sublibraries" / necessity claim: RESOLVED.** Gone from the doc and from `install-ort.mjs`. It survives in the PR body only inside the `## Review` history section, explicitly labelled as checked on-box and found false. That is a correct record, not a live claim.

**F1b — opposite mechanism asserted as fact: NOT RESOLVED.** The specific pass-3 sentence is gone, but the fix promoted a *different* unmeasured mechanism assertion into the load-bearing position, and it is false. See finding 1.

---

### Finding 1 — correctness bug (High)

**`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:80-84`, `:94-96`, `:105-107`; `server/tts-sidecar/scripts/install-ort.mjs:406-410`; PR body ¶3 ("At runtime that stub resolved its required sublibraries…").**

All three surfaces now assert, as the measured result, that ctranslate2's bundled 9.10.2.21 dispatch stub *is the component that consumed* the 9.19.0.56 sublibraries — the doc's Conclusion states it outright:

> "ctranslate2's 9.10.2.21 dispatch stub loaded and correctly resolved its required 9.19.0.56 sublibraries, executed transcription correctly on CUDA, and produced the correct output."

The measurement does not establish this, and the shipped binaries contradict it. **Nothing in ctranslate2 4.8.0 references cuDNN at all.**

Verified against this box's live venv (`server/tts-sidecar/.venv/Lib/site-packages/ctranslate2/`, `version.py` → `4.8.0`):

```
grep -a -i -o -c "cudnn" ctranslate2.dll            -> 0     (59,292,672 bytes)
grep -a -i -o -c "cudnn" _ext.cp312-win_amd64.pyd   -> 0
grep -r -i "cudnn" site-packages/faster_whisper/    -> no matches
```

Positive control, so this is not a grep that cannot fail — the same instrument does find CUDA import names in the same binaries, and does find the sublibrary names inside the stub:

```
grep -a -i -o -c "cublas" ctranslate2.dll                        -> 3
grep -a -o -E "[ -~]{0,14}\.dll" ctranslate2.dll | sort -u
    -> cublas64_12.dll, nvcuda.dll, libiomp5md.dll, libittnotify.dll,
       MSVCP140.dll, VCRUNTIME140.dll, KERNEL32.dll, mkl_*.dll   (no cuDNN)
grep -a -o -E "cudnn(_[a-z]+)?64_9\.dll" cudnn64_9.dll
    -> cudnn_adv64_9.dll, cudnn_cnn64_9.dll, cudnn_graph64_9.dll,
       cudnn_ops64_9.dll, cudnn64_9.dll
```

An import-table entry, a delay-load descriptor entry, and a runtime `LoadLibrary("cudnn64_9.dll")` all require that literal name to be present as an ASCII string in the binary. It is present zero times, case-insensitively, in both `ctranslate2.dll` and `_ext.cp312-win_amd64.pyd`. So ctranslate2's engine has no path — static, delay-loaded, or dynamic — into cuDNN.

The bundled stub is resident for a mundane reason the doc doesn't mention: `ctranslate2/__init__.py:20-21` blanket-loads the package directory —

```python
for library in glob.glob(os.path.join(package_dir, "*.dll")):
    ctypes.CDLL(library)
```

— so `cudnn64_9.dll` is CDLL'd because it is *in the directory*, not because anything asks for it. cuDNN's dispatch stub loads its sublibraries lazily at the first real API call (the doc says this itself at `:78-79`), and nothing ever makes that call through this copy. It is loaded and inert.

That leaves the 9.19.0.56 sublibraries in the process attributable to the only identified cuDNN consumer in the run — onnxruntime's CUDA EP for Kokoro, which the doc's own section 2 (`:30-35`) says had already initialised. The run captured no pre-`/transcribe` snapshot, so it cannot pin *when* they loaded; but ctranslate2's stub is excluded as their loader either way, because it has no caller.

**Concrete failure scenario.** A maintainer hits a future `NVIDIA_CUDNN_CONSTRAINT` bump, or a torch bump, reads `install-ort.mjs:406-410` — "On real hardware, ctranslate2's 9.10.2.21 dispatch stub loads and resolves its required sublibraries from the 9.19.0.56 set present in the process; measurement shows this mixed-version composition is empirically harmless" — and concludes that cross-minor cuDNN skew under ctranslate2 has been *exercised and proven benign*. It has not been exercised at all: this build has no cuDNN code path, so the measurement is silent on skew tolerance. If a later ctranslate2 build reintroduces a cuDNN dependency (this comment's own standing instruction is "Re-check this comment… on any future torch bump"), that comment will actively argue against re-measuring the one thing that changed. This is the reviewer-brief shape "the defect is in the document, not the code," and it is the fourth consecutive round in which an unmeasured mechanism was written up as a measured one.

**The verdict itself survives — the stated reason for it does not.** `/transcribe` on `cuda:0` returned an exact-match transcript with no cuDNN error and no silent CPU fallback. That is measured and it does support "empirically harmless, no fix needed." The honest, *stronger* reason is that the version gap has no consumer in this build, not that a mixed-version stack was exercised and worked. Minimal fix, without inventing a fourth mechanism story: state (a) the module table as measured, (b) that ctranslate2 4.8.0's own binaries contain no cuDNN reference and the stub is resident only via `__init__.py`'s blanket glob-CDLL, (c) that transcription was correct on CUDA — and stop there. Drop "mixed-version stack / composition" from all three surfaces, since it names a composition that was never composed.

### Finding 2 — correctness bug (Medium)

**`server/tts-sidecar/scripts/install-ort.mjs:386-388` and `:414-415`** (unchanged context lines this PR's own measurement made false).

The bullet body was rewritten but the framing around it was not. Line 386-388 still introduces the list as "THREE basenames **mechanism B can shadow**", and line 414-415 still says "All three shadows pre-date this PR (this PR only widens which of them `_add_nvidia_dll_dirs_to_path()`'s PATH prepend **can newly reach**)". Both statements include the ctranslate2 bullet.

This contradicts the doc this PR ships, at `measurement.md:90-92`: ctranslate2 "loads it from there, not via a bare-name `PATH` search — so its 9.10.2.21 dispatch stub loads as a distinct module, **unaffected by `_add_nvidia_dll_dirs_to_path`'s PATH prepend**." Confirmed correct against `ctranslate2/__init__.py` (absolute-path `ctypes.CDLL`). And with finding 1 it is doubly wrong: there is no ctranslate2 cuDNN load for mechanism B to reach in the first place.

**Concrete failure scenario:** the same maintainer, on the same future bump, reads line 414 and re-audits ctranslate2's cudnn64_9.dll as a PATH-prepend shadow risk — the exact question #2845 was filed to answer and this PR answers "no." The comment sends them back around the loop. Fix: move the ctranslate2 bullet out of the "mechanism B can shadow" enumeration, or amend both framing lines to say two of the three are prepend-reachable and the third is not.

### Cleanup nits (non-blocking, do not re-trigger review)

- `measurement.md:78-80` — "Every other loaded cuDNN component is present in the process from the 9.19.0.56 set" restates `:73-76` almost verbatim, two sentences apart. Residue from the rewrite; one of the two can go.
- `measurement.md:73-74` — the sublibrary list omits `cudnn_cnn64_9.dll`, which the stub itself names (grep above). Either it genuinely wasn't loaded (worth saying) or the list is partial.

### Verdict

**Still blocking.** Two correctness findings, both the same underlying problem in different files: the writeup states a consumption mechanism for ctranslate2's bundled cuDNN stub that the measurement never observed and that ctranslate2 4.8.0's shipped binaries rule out. The "no fix needed" conclusion is correct and can stay; the reasoning printed under it, on all three surfaces, cannot.

Note for the fix round: this is checkable without hardware — the two greps above run against the committed venv layout in seconds, and the positive control is one more grep. It does not need another on-box run.

dudarenok-maker:
## PR review — pass 5 (head 214826ea0828b882e30c39184e45a3b71c3f0c35, depth low)

### Verdict: **still blocking** — but the *decision* is right; the *attribution prose* is wrong. Recommend a human applies the edit directly rather than another agent round.

Pass 4's core correction is **confirmed independently and is stronger than the doc claims for itself**. What is still wrong is a different sentence in the same paragraph: the doc now hands the cuDNN consumption to onnxruntime *exclusively*, and its own module table refutes that. This is the fourth consecutive round in which a fix removed one confident-but-unchecked attribution and left a new one one sentence away. That pattern says the remaining edit should be made by a human with the whole document in view, not by another agent patching the sentence I name.

---

### What I re-verified myself (do not re-do these)

| Claim | Result |
|---|---|
| `ctranslate2.dll` cuDNN references | **0** (`grep -a -o -i cudnn \| wc -l`) |
| `_ext.cp312-win_amd64.pyd` cuDNN references | **0** |
| `ctranslate2.dll` cuBLAS references (positive control) | **35** occurrences / 24 distinct `cublas*` symbols |
| `ctranslate2.dll` UTF-16LE `cudnn` | **0** |
| `ctranslate2.dll` import-name strings | `cublas64_12.dll`, `nvcuda.dll`, MKL, CRT — **no `cudnn64_9.dll`** |

The import-name enumeration is a stronger check than the doc's own grep and it agrees: ctranslate2 does not statically import, delay-load, or string-reference cuDNN. **The verdict "no fix needed / close #2845" is correct.**

Also verified correct, so nobody re-checks them:
- `ctranslate2/__init__.py` **lines 20–21** citation — verbatim accurate.
- ctranslate2 **4.8.0** (`version.py`) — accurate.
- "ctranslate2 only ships the single dispatch DLL entry point" — accurate: the bundled `cudnn64_9.dll` internally names `cudnn_adv64_9.dll`, `cudnn_cnn64_9.dll`, `cudnn_graph64_9.dll`, `cudnn_ops64_9.dll`, i.e. it is the cuDNN 9 loader shim, and none of those siblings sit in the package dir.
- `_add_nvidia_dll_dirs_to_path` "runs unconditionally at sidecar boot" — accurate (`server/tts-sidecar/main.py:457-471`, lifespan startup hook).
- The **"mixed-version stack that worked"** framing is **gone** from both the doc and `install-ort.mjs`. It survives only in the PR body's *pass-3 history bullet*, which the pass-4 bullet explicitly overturns — acceptable as history.
- The **forward-looking caveat** (doc:123-126) hedges correctly in both directions: scoped to 4.8.0's build, "unlikely but not ruled out by this measurement", asks for a re-check. No overclaim.
- House gates: `Closes #2845` present and outside a code span; release notes explicitly N/A with a stated reason (correct — comment/doc-only, no shippable delta); no on-box register row references #2845 (grep-verified), so nothing owed there; diff is one new `.md` plus comment-only hunks in `install-ort.mjs`, so no paired test is owed.

---

## Correctness findings

### 1. [Correctness] "onnxruntime … the only actual cuDNN consumer" is refuted by the doc's own table, and by the doc's own methodology

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:88-90`, `:104-106`, `:118-120`
`server/tts-sidecar/scripts/install-ort.mjs:410-412`
PR body ("belong **exclusively** to onnxruntime's CUDA execution provider … the only actual cuDNN consumer in this measurement")

Concrete refutation:

- The doc's **own** module table (line 66) records `…\torch\lib\cudnn64_9.dll` **9.19.0.56 loaded in this process**, and line 70 attributes it to "torch's own bundled copy". Torch was therefore imported and its Windows import-time blanket DLL load ran.
- Applying **the doc's own test** — grep the compiled binary for cuDNN references — to `torch/lib/torch_cuda.dll` gives **1773** hits. To `onnxruntime/capi/onnxruntime_providers_cuda.dll`: **1085**. To `ctranslate2.dll`: **0**.

So by the criterion this doc introduced, the process contains **at least two** loaded cuDNN-referencing binaries. onnxruntime is *a* cuDNN consumer; it is not *the only* one, and nothing in the measurement establishes exclusivity. The doc credits torch at line 70 and then declares onnxruntime the sole consumer eighteen lines later — self-contradictory inside one section.

Failure scenario this produces: `install-ort.mjs`'s PASS-3 note exists to answer "which basenames can mechanism B shadow, and do the versions agree?". A future maintainer reading line 410-412 is told torch is not a cuDNN party in this sidecar, and will skip checking torch's `torch/lib` cuDNN set against the `nvidia/cudnn/bin` set that `_add_nvidia_dll_dirs_to_path` PATH-prepends — the exact comparison every other bullet in that note performs. (They happen to agree today at 9.19.0.56; the note now removes the reason to keep checking.)

### 2. [Correctness] Un-hedged loader attribution — the defect pass 3 removed, reintroduced one paragraph over

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:91-94` (and restated at `:104-106`)

> "The 9.19.0.56 sublibraries **were loaded by** onnxruntime's own cuDNN dispatch stub (which resolves them lazily at the first kernel operation via bare-name `LoadLibrary` calls)"

Stated as measured fact. It is not measured and cannot be: `Get-Process -Module` lists modules and paths, never which loader issued each load. Worse, it is contradicted by the sentence three lines above it:

- Line 84-86 says the loaded sublibraries are present under **both** `nvidia\cudnn\bin` **and** `torch\lib`.
- `onnxruntime_providers_cuda.dll` imports `cudnn64_9.dll` by **bare name** (confirmed — it is in ORT's import-name table), which resolves through `PATH`, where `_add_nvidia_dll_dirs_to_path` has put `nvidia/cudnn/bin` first; that shim then loads its siblings from its own directory. **ORT's stub cannot produce `torch\lib`-pathed modules.** Those were loaded by torch's own blanket load.

So the attribution is not merely unevidenced, it is wrong for half the modules it covers.

Knock-on: this makes the PR body's **pass-3 bullet false about the current doc**. It states the doc "now state[s] only the measured fact … **with no claim about which mechanism supplied the sublibraries**". Lines 91-94 and 104-106 are precisely such a claim.

Note the doc demonstrably knows how to hedge — lines 32-35 hedge a *weaker* claim correctly ("this run did not separately capture a pre-`/transcribe` module snapshot to confirm it directly"). The hedge simply was not applied to the stronger one.

### 3. [Correctness] Stale pre-pass-4 premise survives as §2's lead sentence

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:31`

> "Fed the kokoro PCM straight back into `/transcribe` …, **which triggers ctranslate2's first cuDNN load of the process**"

Under the corrected premise, ctranslate2's bundled copy is loaded at *import* time by the `__init__.py` glob (the doc's own lines 77-80) and is never consumed. §2 tells the reader `/transcribe` makes ctranslate2 load cuDNN; §3 and the Conclusion say ctranslate2 never touches cuDNN. It is the last surviving sentence of the premise pass 4 demolished, and it is positioned at the head of the measurement narrative where a skimming reader forms their model of the whole document.

---

## Cleanup nits

### 4. The positive-control figure is misreported, and the control is weaker than a directly available one

`…/ctranslate2-cudnn-shadow-measurement.md:75-77`; PR body (×2)

"3 cuBLAS references" is a `grep -c` **line** count — `-c` counts matching lines and `-o` does not change that. The actual occurrence count in `ctranslate2.dll` is **35** (24 distinct `cublas*` symbols plus the `cublas64_12.dll` import name). Understating does not weaken the argument, but a document whose entire value is measurement precision should not misreport its own measurement.

The same sentence also says "confirming the grep methodology succeeds when **cuDNN** is actually present" — it means cuBLAS; the cuBLAS control is a proxy for cuDNN. A *direct* control was sitting in the same venv and is far stronger: `onnxruntime_providers_cuda.dll` → **1085** cuDNN hits, `torch/lib/torch_cuda.dll` → **1773**.

### 5. Finding 2's fix left a dangling header and a miscount

`server/tts-sidecar/scripts/install-ort.mjs:388` and `:414`

- L388 now reads "…for basenames mechanism B can shadow:" — missing article, and the list it introduces still contains a bullet the note itself then says is **not** a shadow target. Removing "THREE" fixed the number but left the header framing its own third bullet wrongly.
- L414 "The **two** live shadow targets (nvrtc/nvJitLink, cublas)" counts bullets, not basenames — there are four live basenames (`nvrtc64_120_0`, `nvJitLink_120_0`, `cublas64_12`, `cublasLt64_12`). It also labels nvrtc/nvJitLink "live" when its own bullet says "these two agree, so mechanism B is **not currently exposed** for them". Two different exclusion grounds — version-agreement vs. no-consumer — are being collapsed into one label.

### 6. Ownership wording contradicts the path next to it

`server/tts-sidecar/scripts/install-ort.mjs:401-402`

"`cudnn64_9.dll` from `ctranslate2` (**faster-whisper's** own bundled dispatch stub, `…/site-packages/**ctranslate2**/`)" — the file ships in ctranslate2's wheel; faster-whisper is a dependent. Pre-existing wording, now sitting immediately beside a path that contradicts it.

---

## Recommendation

**Do not send this back through another agent round.** The engineering question is settled and I have re-derived the settling fact by two independent methods. What is left is a small number of sentences in one document that keep acquiring a new unchecked attribution each time the previous one is removed — five rewrites, four of which introduced a fresh wrong claim. That is a symptom of patching sentences instead of holding the whole document, and another patch round is the treatment that has failed four times.

The concrete edit is small and fully specified above. Suggested shape, for whoever makes it:

1. Replace every "the only actual cuDNN consumer" (doc `:89`, `:104`, `:118`; `install-ort.mjs:411`; PR body) with the claim the evidence supports — e.g. *"the loaded 9.19.0.56 cuDNN set belongs to the process's genuine cuDNN consumers — onnxruntime's CUDA execution provider (1085 cuDNN references) and torch (1773) — not to ctranslate2 (0)."*
2. Delete or hedge the loader mechanism at `:91-94` and `:104-106`; a module snapshot does not attribute loads. Deleting is safer than rewording, and nothing downstream needs it.
3. Fix `:31` so §2 does not assert ctranslate2 loads cuDNN for the transcription.
4. Correct "3 cuBLAS references" → 35 (doc `:75`, PR body ×2), or cite the direct cuDNN controls instead.
5. Optional: `install-ort.mjs:388`, `:414`, `:401-402` per nits 5 and 6.

Once those land, the document says only what was measured, and the PR is a clean merge on a correct conclusion.

dudarenok-maker:
## PR review — pass 6 (head 7335fcc99fd0b1ac7931f5a70807662f68cb027e, depth low)

Every figure below was re-derived from scratch against this worktree's own venv — no number was taken from the commit message, the PR body, or any prior pass.

### Verdict: 🔴 still blocking — one figure in the rewritten sentence does not reproduce

The **decision this PR reaches is correct and I confirmed it independently.** What blocks is a companion number introduced by this very rewrite, in the same sentence whose predecessor blocked pass 5. That makes it another instance of the recurring pattern, not a new class.

---

### Re-derivation (all commands run against `server/tts-sidecar/.venv/…`)

| Claim | Doc says | I measured | ✓/✗ |
|---|---|---|---|
| `cudnn` in `ctranslate2/ctranslate2.dll` | 0 | **0** | ✓ |
| `cudnn` in `ctranslate2/_ext.cp312-win_amd64.pyd` | 0 | **0** | ✓ |
| `cublas` occurrences in `ctranslate2.dll` | 35 | **35** | ✓ |
| `cublas` **distinct** in `ctranslate2.dll` | 24 | **26** | ✗ |
| `cudnn` in `torch/lib/torch_cuda.dll` | 1773 | **1773** | ✓ |
| `ctranslate2/__init__.py` glob-CDLL at lines 20-21 | lines 20-21 | lines 20-21, verbatim | ✓ |

Also verified: `ctranslate2.dll`'s `*.dll` name strings are exactly `cublas64_12.dll`, `nvcuda.dll`, `libiomp5md.dll`/`libittnotify.dll`/`mkl_*`, and CRT/`kernel32` — `cudnn64_9.dll` is **not** among them (doc line 81-84 accurate). The package directory ships only `cudnn64_9.dll` and no `cudnn_ops/adv/graph/heuristic/engines` siblings, so doc line 103-107 is accurate. `_add_nvidia_dll_dirs_to_path()` really does run unconditionally in the lifespan startup hook before any request (`main.py:457-471`), so the step-2 framing is accurate.

The methodology sentence is also right: `grep -c -o` does report matching *lines*, and `-o | wc -l` is the correct occurrence count.

---

### Findings

#### 1. [correctness — document] "24 distinct symbols" does not reproduce; the real count is 26

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:78`
`server/tts-sidecar/scripts/install-ort.mjs:409`
PR body (twice: Summary paragraph 2, and the "Final pass (this commit)" paragraph)

Concrete failure: a reader follows the doc's own stated method to re-derive the figure and gets a different number.

```
$ grep -a -i -o "cublas[a-z0-9_]*" .../ctranslate2/ctranslate2.dll | sort -u | wc -l
26
```

26 under every plausible variant I tried: `[a-z0-9_]*` and `[a-zA-Z0-9_]*` both give 26; case-insensitive dedup (`sort -fu`) also gives 26. The only way to reach a *smaller* number is to drop `-i` (which the doc's stated method uses), and that gives 14, not 24. The full distinct set is:

```
cuBLAS  cublas64_12  cublasContext  cublasCreate_v2  cublasDestroy_v2
cublasGemmEx  cublasGemmStridedBatchedEx  cublasGetMathMode  cublasGetProperty
cublasGetStatusName  CublasHandle  cublasSetStream_v2  cublasSgemmStridedBatched
cublasSgemm_v2  cublasStatus_t  cublas_handle
CUBLAS_STATUS_{ALLOC_FAILED,ARCH_MISMATCH,EXECUTION_FAILED,INTERNAL_ERROR,
  INVALID_VALUE,LICENSE_ERROR,MAPPING_ERROR,NOT_INITIALIZED,NOT_SUPPORTED,SUCCESS}
```

No exclusion rule reaches 24 that the doc's own text permits: it explicitly counts `cublas64_12.dll` among the "distinct symbols", which rules out the "drop the non-symbol strings" reading (dropping `cuBLAS` alone gives 25; dropping `cuBLAS` + the DLL name gives 24 but contradicts the parenthetical).

Why this blocks rather than being a nit: this figure sits inside the one sentence in the document that is *about counting rigour* — it warns that `grep -c` undercounts, then presents an undercount with no stated command. Pass 5 blocked on a wrong count (`3`) in this exact sentence; the rewrite fixed the occurrence count and introduced a wrong distinct count beside it. **Same pattern, adjacent claim.** The doc's whole value is being re-derivable by the next person who touches this, and this is the second consecutive round where the first thing a re-derivation turns up is a number that doesn't match.

Fix (one edit, three files): either say **26** and give the command (`grep -a -i -o "cublas[a-z0-9_]*" <file> | sort -u | wc -l`), or drop the distinct-symbol figure entirely — the load-bearing contrast is *35 vs. 0*, and that half is solid.

#### 2. [cleanup nit] Unhedged load-mechanism attribution one paragraph above the hedge that was added to fix it

`server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:68-70`

> (the `nvidia\cudnn\bin` and `torch\lib` copies are both 9.19.0.56 as expected — `_add_nvidia_dll_dirs_to_path`'s PATH-prepend mechanism and torch's own bundled copy; …)

Read as "which package the file belongs to", this is fine. Read as "how this copy got loaded" — which the `as expected —` construction invites — it is unsupported, and this repo's own code says it is probably not how it happened: `onnxruntime.preload_dlls()` loads the nvidia copies by **absolute path first** (`onnxruntime/__init__.py:394-405`), and only falls back to a bare-name PATH search for files it did not find there (`:409-416`). `main.py:429-453` exists precisely to distinguish those two cases, and can also report a third (`torch-skip`, `main.py:398-405`) in which `preload_dlls()` loads nothing at all. The run's own `[ort-preload]` log line — which would settle it — is not among the log excerpts quoted.

This is the same *shape* pass 5 flagged (an unhedged causal attribution the instrument can't establish), one paragraph above the paragraph that was rewritten to hedge exactly that. It is a nit rather than a blocker because the charitable reading is correct and nothing downstream depends on it. Suggested wording: "the copy this installer places under `nvidia/cudnn/bin`, and torch's own bundled copy" — provenance on disk, no load-mechanism claim. **This nit alone would not re-trigger a review.**

---

### Confirmed clean (checked, not assumed)

- **The three specific pass-5 regressions are all genuinely gone.** No "only actual cuDNN consumer" claim survives anywhere (`only` appears 4× in the doc, all legitimate); no "triggers ctranslate2's first cuDNN load" or equivalent (`trigger`/`shadow` appear nowhere in the doc body — `shadow` only in the title); the sublibrary paragraph (`:107-118`) now states plainly what a `Get-Process -Module` snapshot can and cannot establish, and the hedge is correctly placed *before* the torch/onnxruntime sentence rather than after it.
- **Internal consistency doc ↔ `install-ort.mjs` ↔ PR body** holds on every claim I checked — including, unfortunately, the wrong 24.
- **`install-ort.mjs` list structure is fixed.** Header ("every basename mechanism B COULD reach") now covers three bullets, and the summary at `:416-418` correctly splits them 2 reachable / 1 irrelevant, with "Both reachable pairs" agreeing with the count. No dangling header, no miscount.
- **House gates**: not docs-only (touches `.mjs`), so the gate correctly applies; depth `low` correct for a single-scope `chore`. `Closes #2845.` present, outside any code span. Release notes explicitly N/A with a stated reason (comment/doc-only, no behaviour change) — correct. No paired test owed: the `install-ort.mjs` diff is 100% comment lines, verified. On-box register: PR body's claim that no row references #2845 is **true** — I grepped all of `docs/testing/` and found zero. All 19 CI checks green.

### Coherence pass (fresh-eyes)

It now reads as one document rather than five patches. The argument runs cleanly: bundled copy exists → it is loaded → but the compiled code never references cuDNN → so the version gap has nothing to act on, with the cuBLAS set as the positive control proving the method can find references when they exist. The "Interpretation" and "Conclusion" sections agree with each other and with the evidence for the first time in this chain. Finding 1 is a blemish on an otherwise sound writeup, not a structural problem.

One observation, explicitly **not a finding and needing no action**: `ctranslate2/__init__.py:13-18` also calls `os.add_dll_directory(package_dir)`, which puts the ctranslate2 directory on the process DLL search path — a *reverse* shadow surface (ctranslate2's 9.10 copy being reachable by someone else's bare-name load) that is distinct from mechanism B and outside this ticket. The measured module table shows it did not bite in this run (both other copies resident at 9.19.0.56). Neither the doc nor the comment claims otherwise, so nothing here is overclaimed — noting it only so the next reader of `__init__.py` isn't surprised.

### To clear

Fix finding 1 in all three places (doc, `install-ort.mjs`, PR body), optionally fold in finding 2's rewording, push. Pass 7 re-checks the count and nothing else.

dudarenok-maker:
## PR review — pass 7 (head 84a9606fa02612961ef337d045e3988b45bbaaee, depth low)

**Verdict: clean to merge.** No correctness findings. Pass 6's two items are genuinely fixed, the decision is correct, and every load-bearing number re-measured independently this pass matches what the three surfaces claim. Two cosmetic nits below (one a typo commit `84a9606f` itself introduced) — cleanup-only, so per `findings-triage.md` they do not re-trigger the gate; fold them in with the merge if you like.

### Pass-6 findings — both fixed

**Finding 1 (the "24 distinct symbols" claim): removed, not re-guessed.** Confirmed absent from all three surfaces. `grep -n -i "distinct"` over the doc and `install-ort.mjs` returns four hits, all of which are about *versions* or *module identity*, none a symbol tally:

- doc:60 "three different locations, but only **two distinct versions**" — verified correct on disk (`9.19.0.56`, `9.19.0.56`, `9.10.2.21` → 2 distinct);
- doc:101, doc:128–129 — Windows loader module identity, correct as written.

`install-ort.mjs:409` now reads `occurrences of "cublas"` with no count. PR body carries no distinct-symbol figure. For the record, since it's now nobody's claim: the actual distinct set is **26** (`grep -a -o -i -E "cublas[A-Za-z0-9_]*" ctranslate2.dll | sort -u`), matching pass 6. Dropping the figure was the right call.

**Finding 2 (PATH-prepend attribution): correctly hedged.** doc:68–73 now reads "the `nvidia\cudnn\bin` copy is the one `_add_nvidia_dll_dirs_to_path` puts on `PATH` … this doesn't establish which mechanism actually loaded either one into this process, only which package each file belongs to." That is a statement about what the function does (verifiable: `main.py:462–468`, called unconditionally in the lifespan startup path before any request is served) rather than about which loader path won, and the hedge is explicit. Correct.

### Independent re-measurement this pass

All four core numbers re-run against the live venv, from scratch:

| Check | Claimed | Measured |
|---|---|---|
| `cudnn` in `ctranslate2/ctranslate2.dll` | 0 | **0** |
| `cudnn` in `ctranslate2/_ext.cp312-win_amd64.pyd` | 0 | **0** |
| `cublas` in `ctranslate2/ctranslate2.dll` | 35 | **35** |
| `cudnn` in `torch/lib/torch_cuda.dll` | 1773 | **1773** |

Also verified independently, all matching:

- **The `grep -c` undercount claim** (doc:77–79, PR body test plan). `grep -a -i -c "cublas"` on `ctranslate2.dll` returns **3** — so the "3 → 35" narrative and the stated reason for the original wrong figure are both correct, not a post-hoc rationalisation.
- **Every `FileVersion` in the doc's table and in the PASS-3 note**, read off disk: `nvidia\cudnn\bin` 9.19.0.56, `torch\lib` 9.19.0.56, `ctranslate2\` **9.10.2.21**; `cublas64_12.dll`/`cublasLt64_12.dll` torch `6,14,11,1284` vs nvidia `6,14,11,1285`; `nvrtc64_120_0.dll`/`nvJitLink_120_0.dll` both `6.14.11.9000` on both sides. All exact.
- **Every cited symbol name exists**: `cublasCreate_v2`, `cublasGemmEx`, `CUBLAS_STATUS_*` (10 of them), `cublas64_12.dll`.
- **The import-name claim** (doc:84–87). The full `*.dll` string set in `ctranslate2.dll` is `cublas64_12.dll`, `nvcuda.dll`, ten `mkl_*.dll`, ten `api-ms-win-crt-*`, `MSVCP140`/`VCRUNTIME140`/`VCRUNTIME140_1`, `kernel32`, `libiomp5md`, `libittnotify`. **`cudnn64_9.dll` is not among them.** Confirmed.
- **`ctranslate2/__init__.py` lines 20–21** are exactly the `glob.glob(... "*.dll")` / `ctypes.CDLL(library)` pair the doc quotes. Line numbers correct.
- **`ctranslate2 4.8.0`** (doc:154) matches `version.py`.
- **`NVIDIA_CUDNN_CONSTRAINT` `~=9.19.0`** (line 290) and **`NVIDIA_CUBLAS_CONSTRAINT` `~=12.8.0`** (line 321) — both as the comment describes.
- **The sublibrary claim** (doc:106–110): `ctranslate2/` ships exactly one cuDNN file (`cudnn64_9.dll`); the full suite (`cudnn_ops`/`adv`/`graph`/`heuristic`/`engines_*`, plus `cudnn_cnn`) exists only under `nvidia\cudnn\bin` and `torch\lib`. Correct.
- **PR body's house-gate claims**: `Closes #2845.` present and outside any code span; no `#2845` reference anywhere in `docs/testing/onbox-acceptance-register.md`, so "no row owed" holds; release-notes N/A is correctly declared for a comment/doc-only diff. The `install-ort.mjs` half of the diff is 100% comment lines — no behavioural change, consistent with the doc's "No source change made."

**One methodological hole I probed and closed.** The whole conclusion rests on an ASCII grep, which would miss a wide-char `LoadLibraryW(L"cudnn64_9.dll")` literal. PE import and delay-load descriptors store names as ASCII by format definition, so those are definitively covered — but a runtime wide-string load would not be. Ran a UTF-16LE probe (`[Cc]\x00[Uu]\x00[Dd]\x00[Nn]\x00[Nn]\x00`) against both `ctranslate2.dll` and `_ext.*.pyd`: **0 and 0**. The cuBLAS control under the same encoding is also 0 (the binary carries no wide-char DLL-name literals at all), so I validated the probe against a synthetic positive to confirm it can actually fire — it returns 1. The conclusion survives a second encoding.

### Findings

**Cleanup nit 1 — orphaned closing parenthesis introduced by this very commit.** `server/tts-sidecar/docs/ctranslate2-cudnn-shadow-measurement.md:80-82`.

Commit `84a9606f` replaced `occurrences across 24 distinct symbols (` with `occurrences, including symbol names like `, which removed the opening `(` but left its closing `)` in place. The sentence now reads:

> The same command against "cublas" on `ctranslate2.dll` finds **35** occurrences, including symbol names like `cublasCreate_v2`, `cublasGemmEx`, `CUBLAS_STATUS_*`, `cublas64_12.dll`, etc.**)** — confirming the method finds real references when they exist…

Delete the `)` after `etc.`. The same edit also left `install-ort.mjs:409` short-wrapped mid-comment (`occurrences of "cublas": ctranslate2 uses cuBLAS,`) — cosmetic re-flow only, same one-line fix if you're touching it. Purely presentational; no claim is wrong.

**Cleanup nit 2 — PR body pass count contradicts its own list.** PR body, `## Review` section, first sentence:

> Independent `pr-review-gate` passes (depth `low`) ran **five** rounds, each correcting an overclaim the previous round's fix introduced:

followed by six bullets, **Pass 1** through **Pass 6**. Six rounds ran before this one, seven including it. (The charitable reading — "five" counts only passes 2–6, the ones that corrected a *prior fix* — doesn't survive the sentence's own phrasing, which counts rounds run.) Change "five" to "six", or to "six rounds" plus a "seventh confirming pass" line once this comment lands.

### Bottom line

The decision has been correct since pass 4 and is now backed by numbers that hold up under independent re-measurement on every surface — doc, in-file comment, and PR body agree with each other and with the binaries on disk. There is no remaining unverified numeric claim, no causal overclaim, and no cross-surface inconsistency in the technical content. The only two things left are a stray `)` and a stale word "five". **Clean to merge.**

Tree unchanged: `84a9606fa02612961ef337d045e3988b45bbaaee`, `git status --porcelain` empty before and after. Read-only pass — no build, test, or verify commands run.

Issue #2866, opened 2026-09-03
[agent instructions][claude][verify] git-env-scrub helper extraction — verify + open PR (#2855)


dudarenok-maker:
## What

Verify the task child's extraction and, on PASS, open the PR for #2855.

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-2855-git-env-scrub-helper`
- Branch: `chore/scripts-2855-git-env-scrub-helper`

Do not create another worktree. Read-only for source — you report, you do
not fix.

## Parent's acceptance, restated as a checklist

1. `scripts/git-env.mjs` exports a new shared helper (distinct from
   `scrubGitEnv()`) that strips ALL `GIT_*`-prefixed env vars
   case-insensitively — confirm it is genuinely broader than
   `scrubGitEnv()`'s narrow `GIT_ENV_SCRUB_KEYS` list, i.e. it also strips
   `GIT_INDEX_FILE`/`GIT_PREFIX`. If the diff instead just calls
   `scrubGitEnv()` directly from the three test files without a new export,
   that's a FAIL — it silently narrows what those tests scrub (they need the
   broader strip; `scrubGitEnv()` deliberately doesn't provide it per
   #2216).
2. All three test files (`bump-version.test.mjs`, `release-body.test.mjs`,
   `verify-cache.test.mjs`) use the shared helper — no duplicate strip-logic
   implementation remains across the three.
3. `bump-version.test.mjs`'s pre-existing case-insensitivity regression test
   (originally ~line 644-665) still exists and passes.
4. A new or extended unit test exists for the shared export itself, and its
   mutation-kill receipt shows REAL captured failure output (not just
   asserted) for at least one case that specifically distinguishes the new
   helper from `scrubGitEnv()` (i.e. that it also strips `GIT_INDEX_FILE`/
   `GIT_PREFIX`) — **re-run it yourself if the receipt doesn't show real
   output.**
5. `npm run typecheck` passes.
6. No change to `scripts/verify-cache.mjs:960`'s `gitEnv()`, no change to
   `git-scrub.test.mjs`'s exclusion, no change to `scrubGitEnv()` itself or
   `GIT_ENV_SCRUB_KEYS` — anything touching those is out of scope; flag it
   as a finding rather than silently accepting it.

## FAIL ordering — exact, in this order

1. Create a fresh `[agent instructions][claude][task]` rework child, briefed
   from what actually failed (be specific: which acceptance line, what was
   observed vs. expected).
[...]

Issue #2865, opened 2026-09-03
[agent instructions][claude][task] extract shared git-env-scrub helper for throwaway-repo tests (#2855)


dudarenok-maker:
## What

Extract a shared helper for the "strip ALL `GIT_*` env vars, case-insensitively"
pattern that three test files currently hand-roll independently, and switch
all three to use it. This is the operator-approved decision for #2855 (option
1 of the three named in the issue): extract a shared helper, not narrow the
`git-scrub.test.mjs` exclusion, not accept the duplication.

## Background you need

- `scripts/git-env.mjs:59-84` already exports `scrubGitEnv()` and
  `GIT_ENV_SCRUB_KEYS` (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_OBJECT_DIRECTORY`,
  `GIT_COMMON_DIR`) — but this is the WRONG helper to delegate to directly.
  `scrubGitEnv()` deliberately preserves `GIT_INDEX_FILE` and other `GIT_*`
  vars outside its narrow allowlist, because its own callers (production
  code) need `GIT_INDEX_FILE` honoured (see the comment at
  `scripts/git-env.mjs:52-55`, the #2216 correction).
- The three test files instead need a BROADER strip: every env var whose name
  starts with `GIT_` (case-insensitive), full stop — including
  `GIT_INDEX_FILE`, `GIT_PREFIX`, `GIT_AUTHOR_DATE`, etc. — because these are
  throwaway git repos created fresh inside a temp dir for the test, and any
  inherited `GIT_*` var (from a parent hook context, a CI runner, whatever)
  could misdirect them. Confirmed by reading
  `scripts/tests/bump-version.test.mjs:50-56`'s `cleanGitEnv()`:
  ```js
  function cleanGitEnv() {
    const env = { ...process.env };
    for (const key of Object.keys(env)) {
      if (key.toUpperCase().startsWith('GIT_')) delete env[key];
    }
    return env;
  }
  ```
  Note some call sites in that file (e.g. line ~709) deliberately RE-ADD a
  specific `GIT_*` var (`GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE`) after calling
  `cleanGitEnv()` — the shared helper must support that same override-after-
  strip pattern (a plain `{ ...sharedHelper(), GIT_AUTHOR_DATE: x }` spread
  works fine, don't build anything fancier).
- The three current hand-rolled copies (post-#2844 fix, per #2855's own body):
[...]

PR #2876, opened 2026-09-04
chore(scripts): extract shared throwaway-repo git-env scrub helper (#2855)

### LINKED ISSUE #2855: chore(scripts): decide how to close the git-env-scrub duplication across 3 throwaway-repo test files
## What

PR #2844 (closes #2841) fixed the same case-sensitivity bug — a git-env scrub that fails to catch a differently-cased `git_dir`/`git_index_file`/etc. survivor on Windows — in three separate hand-rolled copies across `scripts/tests/bump-version.test.mjs`, `scripts/tests/release-body.test.mjs`, and `scripts/tests/verify-cache.test.mjs`, each spelled slightly differently (a `key.startsWith('GIT_')` loop in two, a hardcoded-uppercase object destructure in the third). All three needed their own fix because `scripts/tests/git-scrub.test.mjs`'s repo-wide `scrubGitEnv` guard explicitly **excludes** `scripts/tests/**` from its scan (line ~988-991), so nothing mechanical would have caught any of them, and nothing mechanical stops a fourth copy of the same mistake from being written tomorrow.

## The decision this needs

Pick one (there is more than one defensible answer, which is why this is filed rather than fixed blind):

1. **Extract a shared local helper** these three test files import instead of hand-rolling their own `cleanGitEnv()` — but it can't just be the existing `scrubGitEnv()` from `scripts/git-env.mjs`, since these three deliberately need to ALSO strip `GIT_INDEX_FILE`/`GIT_PREFIX`, which `scrubGitEnv()` intentionally preserves for its own (production) callers per the #2216 correction. Would need either a second exported helper in `git-env.mjs` (e.g. `scrubGitEnvForThrowawayRepo()`) or a small shared test-only module under `scripts/tests/`.
2. **Narrow `git-scrub.test.mjs`'s exclusion** so it can see into `scripts/tests/**` after all, if the reason for the current blanket exclusion turns out not to actually apply to these three call sites specifically. Needs reading why the exclusion exists before ruling this in or out (whatever comment/commit added the exclusion should say).
3. **Accept the duplication deliberately** and rely on human/review vigilance the next time someone touches a throwaway-repo git-spawning test in this repo, documenting the pattern once so a future copy is at least written correctly the first time.

## Where the three current copies live (post-#2844 fix)

- `scripts/tests/bump-version.test.mjs` — `cleanGitEnv()`, ~line 50
- `scripts/tests/release-body.test.mjs` — `cleanGitEnv()`, ~line 61-70
- `scripts/tests/verify-cache.test.mjs` — `cleanGitEnv()`, ~line 1408-1420

## Not in scope

`scripts/verify-cache.mjs:960`'s `gitEnv()` (production code, not a test) was also fixed in #2844 for consistency (its `GIT_PREFIX` strip was case-sensitive, though inert — no live bug since none of its git commands are affected by that variable's casing). That one isn't part of this "three hand-rolled copies" duplication since it already delegates to the shared `scrubGitEnv()` and only had its own narrow extra strip on top.

dudarenok-maker:
## Summary

Extracts `scrubGitEnvForThrowawayRepo()` into `scripts/git-env.mjs` — a broader sibling of `scrubGitEnv()` that strips every `GIT_*`-prefixed env var case-insensitively (including `GIT_INDEX_FILE`/`GIT_PREFIX`, which `scrubGitEnv()` deliberately preserves for production callers per the #2216 correction). `bump-version.test.mjs`, `release-body.test.mjs`, and `verify-cache.test.mjs` had each hand-rolled an identical strip loop as a local `cleanGitEnv()`; all three now delegate to the shared helper instead. Adds `scripts/tests/git-env.test.mjs` with direct coverage of the new export, including the `GIT_INDEX_FILE`/`GIT_PREFIX` distinction from `scrubGitEnv()` that had no prior coverage anywhere.

Closes #2855

## Release notes

Not applicable — internal test/tooling consolidation with no user- or operator-visible effect.

Findings from the mandatory PR review pass (pr-review-gate), all 🟡 minor/cleanup, each fixed with its own dispatched commit:

- `2f214221` — corrected a stale coverage claim in `git-env.test.mjs`'s header comment (it claimed `release-body.test.mjs`/`verify-cache.test.mjs` still pin `scrubGitEnv()`'s GIT_INDEX_FILE preservation; both now only cover `scrubGitEnvForThrowawayRepo()`, which does the opposite).
- `4545a03b` — corrected three comments that pointed to the wrong docblock for the case-insensitivity rationale (`scrubGitEnvForThrowawayRepo()` states the property but doesn't explain it; the rationale lives on `scrubGitEnv()`).
- `366a0a2c` — consolidated a fourth hand-rolled throwaway-repo scrub in `scripts/tests/publish-token-git.test.mjs` onto the shared `scrubGitEnvForThrowawayRepo()` helper; the old hand-rolled version was narrower and left `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE`/`GIT_PREFIX` leaking into fixture commits.

## Test plan

- [x] `node --test` across all 4 touched test files (`git-env.test.mjs`, `bump-version.test.mjs`, `release-body.test.mjs`, `verify-cache.test.mjs`): 202/202 pass — verified independently by this verify pass (#2866), matching the implementer's own run.
- [x] `npm run typecheck` — exit 0, verified independently.
- [x] Mutation check: temporarily excluding `GIT_INDEX_FILE` from the strip and re-running `git-env.test.mjs` turns the new "strips GIT_INDEX_FILE and GIT_PREFIX too" test red (`AssertionError: expected undefined, got '/decoy/.git/index'`), confirming it actually exercises the GIT_INDEX_FILE/GIT_PREFIX distinction — per the implementer's #2865 AGENT DONE receipt.
- [x] `bump-version.test.mjs`'s pre-existing case-insensitivity regression test (`scrubGitEnv strips a git repo-discovery override regardless of stored casing`, ~line 669) still present and passing, unchanged.
- [x] Out-of-scope check: no changes to `scrubGitEnv()`, `GIT_ENV_SCRUB_KEYS`, `verify-cache.mjs`'s `gitEnv()`, or `git-scrub.test.mjs` — confirmed via diff (touched files are only `scripts/git-env.mjs` and the three test files plus the new `git-env.test.mjs`).
- [x] Review-round fixes (`2f214221`, `4545a03b`, `366a0a2c`) each verified via `node --test` on the affected files.

### DIFF
--- a/scripts/git-env.mjs
+++ b/scripts/git-env.mjs
@@ -82,3 +82,23 @@ export function scrubGitEnv(env = process.env) {
   }
   return out;
 }
+
+/** A copy of `env` with EVERY key whose name starts with `GIT_` (matched
+ *  case-insensitively) removed — deliberately broader than `scrubGitEnv()`
+ *  above, which is the WRONG helper to reach for here. `scrubGitEnv()`
+ *  exists for production call sites that must still honour `GIT_INDEX_FILE`
+ *  and other `GIT_*` vars outside its narrow allowlist (see the #2216
+ *  correction above). This helper is for tests that spawn git inside a
+ *  throwaway repo built fresh in a temp dir: there is no real index or
+ *  ambient repository state for those to honour, so any inherited `GIT_*`
+ *  var — `GIT_INDEX_FILE`, `GIT_PREFIX`, `GIT_AUTHOR_DATE`, all of it — is
+ *  only ever a hazard, not a value worth preserving. A caller that still
+ *  needs one of them can re-add it after calling this, e.g.
+ *  `{ ...scrubGitEnvForThrowawayRepo(), GIT_AUTHOR_DATE: x }`. */
+export function scrubGitEnvForThrowawayRepo(env = process.env) {
+  const out = { ...env };
+  for (const key of Object.keys(out)) {
+    if (key.toUpperCase().startsWith('GIT_')) delete out[key];
+  }
+  return out;
+}
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -19,7 +19,7 @@ import assert from 'node:assert/strict';
 // Pure helper from the script (import is inert — the script's procedure is
 // behind an import.meta-main guard, so loading it here doesn't run a release).
 import { pickWorkflowRun, readSidecarVersion, writeSidecarVersion, sidecarVersionPath, readPubspecVersion, writePubspecVersion, pubspecPath, pubspecBuildNumber, resolveNotesFile, staleNotesVersion, DEFAULT_NOTES_FILE, buildTagMessage, createAnnotatedTag, execGit } from '../bump-version.mjs';
-import { scrubGitEnv } from '../git-env.mjs';
+import { scrubGitEnv, scrubGitEnvForThrowawayRepo } from '../git-env.mjs';
 import { join } from 'node:path';
 import { execFileSync, spawnSync } from 'node:child_process';
 import { mkdtempSync, mkdirSync, rmSync, writeFileSync, readFileSync } from 'node:fs';
@@ -42,17 +42,11 @@ const MOJIBAKE_EM_DASH = 'â€”'; // should decode to U+2014 —, standalone 
 // which child processes inherit. Without sanitising, the test's `git commit`
 // would write into the PARENT worktree's git instead of the temp fixture —
 // silently creating bogus commits on whatever branch invoked the hook.
-// Case-insensitive match: Windows preserves whatever casing a var was stored
-// under while lookup is case-insensitive, and git honours a lowercase
-// `git_dir` identically to `GIT_DIR` — see git-env.mjs's scrubGitEnv() header
-// for the same trap. A `startsWith('GIT_')` filter alone leaves a `git_dir`
-// (or similar) survivor that still redirects the throwaway repo's commits.
+// Thin wrapper over the shared, broader-than-scrubGitEnv() helper — see
+// git-env.mjs's scrubGitEnvForThrowawayRepo() for the case-insensitivity
+// rationale and why this needs a separate helper from scrubGitEnv().
 function cleanGitEnv() {
-  const env = { ...process.env };
-  for (const key of Object.keys(env)) {
-    if (key.toUpperCase().startsWith('GIT_')) delete env[key];
-  }
-  return env;
+  return scrubGitEnvForThrowawayRepo();
 }
 
 // Wrap execFileSync to always pass the sanitised env. Every git invocation
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -43,8 +43,8 @@ const MOJIBAKE_EM_DASH = 'â€”'; // should decode to U+2014 —, standalone 
 // would write into the PARENT worktree's git instead of the temp fixture —
 // silently creating bogus commits on whatever branch invoked the hook.
 // Thin wrapper over the shared, broader-than-scrubGitEnv() helper — see
-// git-env.mjs's scrubGitEnvForThrowawayRepo() for the case-insensitivity
-// rationale and why this needs a separate helper from scrubGitEnv().
+// git-env.mjs's scrubGitEnv() docstring for the case-insensitivity rationale,
+// and why this needs a separate helper from scrubGitEnv().
 function cleanGitEnv() {
   return scrubGitEnvForThrowawayRepo();
 }
--- a/scripts/tests/git-env.test.mjs
+++ b/scripts/tests/git-env.test.mjs
@@ -81,3 +81,29 @@ test('a caller can re-add a specific GIT_* var after scrubbing via plain spread'
   assert.equal(merged.GIT_AUTHOR_DATE, 'new');
   assert.equal(merged.PATH, '/usr/bin');
 });
+
+// Regression test for the defensive-copy mutant: ensure the function returns
+// a NEW object and does NOT mutate its input argument. Passing an explicit
+// plain object (not process.env) as the argument makes this checkable without
+// having to save/restore the original.
+test('scrubGitEnvForThrowawayRepo returns a new object and does not mutate its input', () => {
+  const inputEnv = {
+    PATH: '/usr/bin',
+    GIT_DIR: '/decoy/.git',
+    GIT_WORK_TREE: '/decoy',
+    GIT_AUTHOR_DATE: '2026-01-01',
+  };
+  const scrubbed = scrubGitEnvForThrowawayRepo(inputEnv);
+
+  // The returned object must have GIT_* keys stripped
+  assert.equal(scrubbed.GIT_DIR, undefined);
+  assert.equal(scrubbed.GIT_WORK_TREE, undefined);
+  assert.equal(scrubbed.GIT_AUTHOR_DATE, undefined);
+  assert.equal(scrubbed.PATH, '/usr/bin');
+
+  // The input object must remain UNCHANGED — still contains all original keys
+  assert.equal(inputEnv.PATH, '/usr/bin');
+  assert.equal(inputEnv.GIT_DIR, '/decoy/.git');
+  assert.equal(inputEnv.GIT_WORK_TREE, '/decoy');
+  assert.equal(inputEnv.GIT_AUTHOR_DATE, '2026-01-01');
+});
--- a/scripts/tests/git-env.test.mjs
+++ b/scripts/tests/git-env.test.mjs
@@ -1,0 +1,83 @@
+// Unit tests for scripts/git-env.mjs's shared GIT_* env helpers.
+// scrubGitEnv() itself already has coverage via its callers' regression
+// suites (bump-version.test.mjs, release-body.test.mjs, verify-cache.test.mjs
+// each pin its case-insensitivity and GIT_INDEX_FILE preservation). This file
+// is for scrubGitEnvForThrowawayRepo() (#2865), the broader sibling those
+// three test files now delegate their local cleanGitEnv() to instead of each
+// hand-rolling the strip loop.
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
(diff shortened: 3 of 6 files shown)


dudarenok-maker:
## PR review — pass 1 (head 5ce56da0975b311d21a2c4ae0a79892617ee5042, depth low)

Scope: all 5 changed files read in full (`scripts/git-env.mjs`, `scripts/tests/{git-env,bump-version,release-body,verify-cache}.test.mjs`), plus the four non-diff files the change reaches into — `scripts/tests/git-scrub.test.mjs` (the repo-wide scrub guard), `scripts/tests/publish-token-git.test.mjs` (the fourth throwaway-repo scrub site), `scripts/run-hooks-tests.mjs`, and `scripts/verify-cache.mjs`'s `test:hooks` inputs. Verified: all 19 CI checks green on this SHA; `scripts/**/*.{mjs,cjs,js,mts,cts,ts}` is a `test:hooks` glob input (verify-cache.mjs:94) so both new/changed files bust that step's cache and the new file is picked up by `run-hooks-tests.mjs`'s `fast-glob('scripts/tests/*.test.mjs')` — no hand-registration owed. Then adversarially probed the new test file by mutating a scratch copy of `git-env.mjs` outside the repo (4 mutants), and probed `GIT_PREFIX` / `GIT_AUTHOR_DATE` leakage against a real throwaway repo.

**Mutation results (scratch copy, repo untouched)** — 4/4 baseline green; `key.toUpperCase()` → `key.startsWith` turns test 1 red; excluding `GIT_INDEX_FILE` turns test 2 red; `env = process.env` → `env = {}` turns test 3 red. The tests are real and red for the reasons claimed. One mutant survived — see 🟡 5.

### 🔴 Blocking

None.

### 🟠 Significant

None.

### 🟡 Minor

**1. `scripts/tests/git-env.test.mjs:2-5` — the header's coverage claim is false for two of the three files it names, and this PR is what made it false.**

The new file's header justifies covering only the new helper:

> `scrubGitEnv()` itself already has coverage via its callers' regression suites (bump-version.test.mjs, release-body.test.mjs, verify-cache.test.mjs each pin its case-insensitivity and GIT_INDEX_FILE preservation).

After this PR, `release-body.test.mjs:21` and `verify-cache.test.mjs:47` import **only** `scrubGitEnvForThrowawayRepo`; neither file references `scrubGitEnv` anywhere but in prose. Their surviving `cleanGitEnv strips a git env override regardless of stored casing` tests (release-body.test.mjs:400, verify-cache.test.mjs:1452) assert `cleaned.Git_Index_File === undefined` — GIT_INDEX_FILE **removal**, the exact opposite of the preservation the header credits them with. `git-scrub.test.mjs` imports no runtime symbol from `git-env.mjs` at all (its imports are `node:test`/`assert`/`node:fs`/`node:path`/`node:url`, lines 192-196) — it is a source-text guard, not behavioural coverage.

Concrete failure scenario: delete `'GIT_OBJECT_DIRECTORY'` from `GIT_ENV_SCRUB_KEYS` (`scripts/git-env.mjs:62`). Exactly **one** test in the repo goes red — `bump-version.test.mjs:610` — not the three suites the header credits. A maintainer trusting the header concludes `scrubGitEnv` is triple-covered and declines to add direct coverage to the one file that is now its natural home. Documentation, not runtime behaviour.

**2. `bump-version.test.mjs:45-47`, `release-body.test.mjs:62-64`, `verify-cache.test.mjs:1416-1418` — the replacement pointer sends the reader to a docblock that does not carry the rationale it cites.**

All three replacement comments read "see git-env.mjs's `scrubGitEnvForThrowawayRepo()` for the case-insensitivity rationale". That docblock (`scripts/git-env.mjs:86-97`) *states* the property — "(matched case-insensitively)" — and gives no rationale. The Windows-casing rationale lives in `scrubGitEnv()`'s docblock at `scripts/git-env.mjs:71-76`, which the new one does not point at. This PR deleted the local rationale from bump-version.test.mjs (old lines 45-49) and from release-body.test.mjs (old lines 61-65, which also carried the `#2841 … for the measured evidence` citation) and replaced it with a pointer to a location the answer is not at. Concretely: a contributor reading `scrubGitEnvForThrowawayRepo`'s docblock and asking whether `.toUpperCase()` is load-bearing finds nothing at the cited spot. The new test file is a real backstop against actually landing that simplification (mutant 1 above proves it), so this is a doc-accuracy nit, not a live hazard. Fixable by one clause in the new docblock, or by pointing at `scrubGitEnv()`'s header as the deleted comments did.

**3. `scripts/tests/publish-token-git.test.mjs:63-69` — a fourth throwaway-repo git-env scrub was left hand-rolled, and is now provably narrower than the helper this PR canonicalised.**

That file builds throwaway repositories (`newRepo()`, line ~93) and routes every git spawn through `cleanEnv()` = `scrubGitEnv(process.env)` plus a manual `GIT_INDEX_FILE` delete. That is precisely the "throwaway repo, nothing ambient worth honouring" case `scrubGitEnvForThrowawayRepo()` now exists for, and it is strictly narrower — it leaves `GIT_PREFIX`, `GIT_AUTHOR_DATE`, `GIT_COMMITTER_DATE`, `GIT_CONFIG_*`, `GIT_NAMESPACE`, `GIT_CEILING_DIRECTORIES` and `GIT_ALTERNATE_OBJECT_DIRECTORIES` in the child env, all of which this PR's own docblock (`git-env.mjs:92-95`) declares "only ever a hazard, not a value worth preserving". The file's own header (line ~36) worries about "becoming a third divergent copy"; after this PR it *is* the divergent copy, and it is the site with this repo's worst recorded incident (a 3,979-entry index replaced with a one-entry one, twice).

Measured, not asserted: `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE` **do** survive into fixture commits — `GIT_AUTHOR_DATE=2001-02-03T04:05:06Z GIT_COMMITTER_DATE=… git -C <tmp> commit -qm base` then `git log -1 --format='%ai | %ci'` → `2001-02-03 04:05:06 +0000 | 2001-02-03 04:05:06 +0000`. `GIT_PREFIX` I could **not** make bite: `GIT_PREFIX=nonexistent/sub/ git -C <tmp> add -A` staged correctly, exit 0. So this is a genuine divergence with a demonstrated leak, but no demonstrated wrong test outcome today — hence 🟡, not 🟠. Nothing mechanical will catch it either: `git-scrub.test.mjs` still excludes `scripts/tests/**` from its scan (line 588; asserted at line 990).

*Context, not a finding against this PR:* #2855 listed exactly that mechanical gap ("nothing mechanical stops a fourth copy of the same mistake from being written tomorrow") as motivation, and its option 2 was to narrow the exclusion. Option 1 was chosen and operator-approved, so the residual gap is a recorded trade-off, not a defect in the delivery. But the fourth site already exists today, which the issue did not know.

**4. PR body — no explicit release-notes not-applicable statement (CLAUDE.md before-shipping step 5).**

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is touched and the body says nothing about them. The omission is correct **on the merits** — a test-helper extraction has no user- or operator-visible delta — but step 5 says "say so explicitly rather than silently omitting". Same for step 1 (regression plan) and step 3 (on-box acceptance), both genuinely N/A and both silently omitted. Bookkeeping only; the body edit is the whole fix.

**5. `scripts/git-env.mjs:98-104` — one mutant survives the new suite: the helper mutating its argument in place is invisible.**

Replacing `const out = { ...env };` with `const out = env;` leaves all four new tests green (verified: 4 pass / 0 fail). The "defaults to scrubbing `process.env`" test at `git-env.test.mjs:61` is the one that would be expected to catch it and cannot — it saves `process.env.GIT_DIR` before the call and restores it in `finally`, so an in-place delete reads as a correct scrub. Consequence of that regression landing: `scrubGitEnvForThrowawayRepo()` called with the default argument would permanently strip every `GIT_*` var from the test runner's own `process.env` for the rest of the file, and each of the three `cleanGitEnv` regression tests snapshots `process.env` *before* calling, so all of them mask it too. The same gap exists for `scrubGitEnv()` (pre-existing, same shape), and this new file is the natural home for a one-line purity assertion covering both — `assert.deepEqual(Object.keys(fakeEnv).length, 8)` after the call, or equivalent.

### ✅ What is solid

- **The extraction is behaviour-preserving.** All three replaced loops were `{ ...process.env }` + case-insensitive `GIT_`-prefix delete; `scrubGitEnvForThrowawayRepo()` with its `env = process.env` default is byte-equivalent. `GITHUB_TOKEN` and friends correctly survive (`'GITHUB_TOKEN'.startsWith('GIT_')` is false), as before.
- **The out-of-scope boundary held.** `scrubGitEnv()`, `GIT_ENV_SCRUB_KEYS`, `verify-cache.mjs`'s `gitEnv()` and `git-scrub.test.mjs` are all untouched — confirmed against the diff, not just the PR body's claim. The #2216 `GIT_INDEX_FILE` exclusion is intact and now has a test asserting the *distinction* rather than only each side of it.
- **The new tests are real regression tests**, not shape-conforming decoration — three of four mutants caught, each red for the reason its name claims (see the mutation table above).
- **The docblock does the hard part right**: it names `scrubGitEnv()` as "the WRONG helper to reach for here" and says why, which is the failure mode two similarly-named exports actually invite.
- **The `verify-cache.test.mjs` comment was appended to, not replaced** — its "demonstrated live: an ambient lowercase `git_dir` flipped a real checkout's `core.bare`" evidence survives, unlike the two files where the local rationale was swapped for a pointer (🟡 2).
- Scope filtering, harness pickup and issue linkage all check out: `test:hooks` globs reach both files, `run-hooks-tests.mjs` globs rather than enumerates, `Closes #2855` is present and outside any code span, and #2855 exists and is open.

### Verdict

**Approve.** Zero 🔴 / 🟠 — no correctness bug, no re-review triggered. This is a clean, tightly-scoped extraction that stayed inside the boundary the issue drew, and its paired tests survive mutation. Five 🟡 cleanup items, all in this round per the report-fix-record rule: the false coverage claim (1) is the one worth fixing first, since it is the stated justification for what the new test file does *not* cover; (3) is the one with the longest tail, since it leaves the canonical helper with a documented sibling that ignores it.

Issue #2867, opened 2026-09-03
[agent instructions][claude][task] force full server-test run on lockfile-only diff (#2853)


dudarenok-maker:
## What

`.github/workflows/verify.yml`'s server-tests step decides narrowed-`--changed`
vs. full run purely on whether `$BASE` is set (i.e. is this a PR at all), and
never checks whether the diff touches a lockfile. Fix: force a FULL (non-
`--changed`) `test:server`/`test:server-slow` run whenever the diff touches
either lockfile (`package-lock.json` at repo root, or
`server/package-lock.json`), even on a PR. This is the operator-approved
fix for #2853 (option 1 of the three offered: "force full run on any
lockfile diff", chosen over extending `forceRerunTriggers` per-config or
punting to a broader design pass).

## Background you need

- `.github/workflows/verify.yml:327-339` — the `Server tests (fast + slow)`
  step. Its `run:` body:
  ```bash
  cd server
  BASE="${{ github.event.pull_request.base.sha }}"
  if [ -n "$BASE" ]; then
    npx vitest run --changed "$BASE"
    npx vitest run --config vitest.config.slow.ts --changed "$BASE"
  else
    npx vitest run
    npx vitest run --config vitest.config.slow.ts
  fi
  ```
  Confirmed by direct read: this branches ONLY on `$BASE` being set (PR vs.
  manual dispatch) — it never looks at the diff's file list or the `shared`
  scope flag the job's own `if:` condition (line 328) already has access to.
- The hazard (from #2853's own body, already measured): a
  `server/package-lock.json`-only PR is not `shared` scope — only the ROOT
  lockfile/manifest and `.github/actions/**` count as `shared`
  (`scripts/verify-cache.mjs`'s `computeShared`) — so it reaches this step
  via `stepTouchedByDiff`'s `includeLockfiles` branch (confirmed:
  `test:server`'s step definition at `scripts/verify-cache.mjs:358-441` has
  `includeLockfiles: ['server']`), but stays on the `--changed "$BASE"` path.
  Neither `package-lock.json` nor `server/package-lock.json` is in
  `server/vitest.config.ts`'s `forceRerunTriggers`, so `--changed` can select
  ZERO tests and the step reports green having tested nothing against a
  changed dependency tree.
- `computeShared`/`stepTouchedByDiff`/the `STEPS[]` array all live in
  `scripts/verify-cache.mjs` — read there for the exact `includeLockfiles`
  mechanics before assuming a shape.
[...]

Issue #698, opened 2026-06-10
Pre-push verify flakes on parallel-worker e2e (about/admin lazy-route + visual specs)


dudarenok-maker:
## What

`npm run verify` (the pre-push gate) intermittently fails its **e2e** legs under local parallel workers, forcing developers to push with `--no-verify`. The failing specs are always the same set and are **unrelated to the change under test**:

- `e2e/about-page.spec.ts`
- `e2e/admin.spec.ts`, `e2e/admin-ui-polish.spec.ts`
- `e2e/responsive/visual.spec.ts` (visual baselines — "N did not run")

Observed on the cover-search branch push (2026-06-10): `[fail] test:e2e` (150 passed, lazy-route specs flaked) and `[fail] test:e2e:visual` (5 passed, 6 did not run).

## Why it's a flake, not a regression

- A **single-worker run is green**: `playwright test --workers=1` → 167 passed, 2 flaked, **0 failed, exit 0**. That is exactly the mode **CI uses** (`workers: process.env.CI ? 1 : undefined` in `playwright.config.ts`), so **CI does not hit this** — only the local parallel pre-push gate does.
- Direct probes confirm the pages render fine; the failures are `page.goto` timing out on the `load` event under contention (pending external `api.fontshare.com` font CSS + cold-cache Vite dep optimization on lazy `React.lazy` route chunks). `playwright.config.ts` already documents this contention mode.
- The flake surfaces/worsens as the parallel spec pool grows (adding `e2e/cover-picker.spec.ts` nudged it).

## Impact

Developers are forced to `git push --no-verify` to land legitimately-green work, which erodes the gate (a real failure could slip through behind the noise).

## Proposed fixes (pick one)

1. Mark the lazy-route specs `test.describe.configure({ mode: 'serial' })` (about-page, admin, admin-ui-polish) so they don't race under parallelism.
2. Stub/route-block the external `api.fontshare.com` font request in the e2e fixture so `page.goto`'s `load` never waits on the network.
3. Cap local e2e workers (e.g. `workers: 4`) to reduce contention while keeping CI at 1.

Option 2 is the most robust (removes the external dependency entirely); option 1 is the smallest change.

## Acceptance

`npm run verify` passes its e2e legs reliably under default local parallelism across repeated runs, without `--no-verify`.

PR #2877, opened 2026-09-04
chore(ops): force full server-test run on a lockfile-only diff (#2853)

### LINKED ISSUE #2853: chore(ops): CI's --changed narrowing can select zero tests on a server-lockfile-only diff

dudarenok-maker:
## Summary

`.github/workflows/verify.yml`'s "Server tests (fast + slow)" and "Frontend tests + a11y" steps chose `--changed` vs. full purely on whether `$BASE` was set, never on the diff's file list. Neither `package-lock.json` nor `server/package-lock.json` is in either vitest config's `forceRerunTriggers`, so a lockfile-only PR could select zero tests against a changed dependency tree while reporting green. Adds a `lockfile_touched` scope key to `ci-scope.mjs` (reusing the detect job's existing scopes blob) and threads it into both the server-tests and frontend-tests steps to force a full run whenever it's true, even on a PR. 

**Incidental fixes folded into this PR:** frontend guard wiring (commit `2b110850`), regression-test polarity assertion hardening (commit `6348c7fc`), and two stale comments updated to reflect both test-step consumers. Closes #2853.

## Test plan

- [x] `node --test scripts/tests/ci-scope.test.mjs` — 16/16 pass, including 3 new `lockfile_touched` cases (server lockfile, root lockfile via `shared`, non-lockfile diff).
- [x] Mutation-kill: replacing the `lockfile_touched` detector body with a bare `false` made the new "sets lockfile_touched for a server lockfile diff" case fail as expected; reverted, suite back to 16/16 (from the task commit's own verification, re-confirmed by this verify pass via re-run of the un-mutated suite).
- [x] `npm run typecheck` — clean (root `tsc --noEmit` and server typecheck).
- [x] Verified the diff touches `ci-scope.mjs`'s `CI_ONLY` map, `verify.yml`'s server-tests and frontend-tests step shell logic — no changes to `computeShared`, `STEPS[]`, or either vitest config's `forceRerunTriggers`.
- [x] Verified `test:a11y` step is untouched.
- [x] Regression test (`scripts/tests/workflow-wiring.test.mjs`) verifies polarity: when LOCKFILE_TOUCHED != "true", run with --changed; when true, run full suite. Order-guard strengthened to verify LOCKFILE_TOUCHED assignment precedes the if condition in both frontend and server steps. Mutation proof: M7 (neutering server else to --changed) and M11 (swapping server then/else bodies) both fail the test as expected.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -267,8 +267,9 @@ jobs:
       # change to src/test/setup.ts (runner-injected, NOT import-reachable)
       # forces a full run via forceRerunTriggers in vitest.config.ts; a root
       # dep/config bump (the `shared` scope) forces a full run via vitest's
-      # default triggers. test:a11y stays always-on as a cheap core-view gate.
-      # See docs/features/118-ci-cost-round-2.md.
+      # default triggers; a lockfile-only diff forces a full run via the
+      # LOCKFILE_TOUCHED flag (#2853). test:a11y stays always-on as a cheap
+      # core-view gate. See docs/features/118-ci-cost-round-2.md.
       # test:e2e is a deliberate extra trigger: this step also runs
       # `npm run test:a11y`, which has no STEP of its own, and an e2e-only
       # diff should still exercise the a11y battery. Dropping it would be a
@@ -280,8 +281,9 @@ jobs:
         if: fromJSON(needs.detect.outputs.scopes).step_test || fromJSON(needs.detect.outputs.scopes).step_test_e2e || fromJSON(needs.detect.outputs.scopes).shared
         run: |
           BASE="${{ github.event.pull_request.base.sha }}"
+          LOCKFILE_TOUCHED="${{ fromJSON(needs.detect.outputs.scopes).lockfile_touched }}"
           # On a manual dispatch BASE is empty (no PR) → run the full suite.
-          if [ -n "$BASE" ]; then npx vitest run --changed "$BASE"; else npx vitest run; fi
+          if [ -n "$BASE" ] && [ "$LOCKFILE_TOUCHED" != "true" ]; then npx vitest run --changed "$BASE"; else npx vitest run; fi
           npm run test:a11y
 
   server-tests:
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -267,8 +267,10 @@ jobs:
       # change to src/test/setup.ts (runner-injected, NOT import-reachable)
       # forces a full run via forceRerunTriggers in vitest.config.ts; a root
       # dep/config bump (the `shared` scope) forces a full run via vitest's
-      # default triggers; a lockfile-only diff forces a full run via the
-      # LOCKFILE_TOUCHED flag (#2853). test:a11y stays always-on as a cheap
+      # default triggers; a lockfile-only OR shared-scope diff forces a full run
+      # via the LOCKFILE_TOUCHED flag (#2853) — note lockfile_touched is
+      # computed as (shared || lockfile-specific-match), so e.g. an .github/actions/**
+      # diff also triggers a full run. test:a11y stays always-on as a cheap
       # core-view gate. See docs/features/118-ci-cost-round-2.md.
       # test:e2e is a deliberate extra trigger: this step also runs
       # `npm run test:a11y`, which has no STEP of its own, and an e2e-only
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -323,14 +323,23 @@ jobs:
       # condition is the union of both steps' keys. Drops the old `sidecar`
       # disjunct deliberately.
       # (was: server || sidecar || shared).
+      #
+      # #2853/#2867: neither lockfile is in either vitest config's
+      # forceRerunTriggers, so `--changed` can select ZERO tests against a
+      # changed dependency tree. `detect`'s scopes blob already carries
+      # `lockfile_touched` (ci-scope.mjs) computed from the same PR diff this
+      # job would otherwise re-derive — reuse it rather than a second
+      # merge-base/git diff call, and force the full (non-`--changed`) run
+      # whenever it is set, even on a PR.
       # leg: test:server
       - name: Server tests (fast + slow)
         if: fromJSON(needs.detect.outputs.scopes).step_test_server || fromJSON(needs.detect.outputs.scopes).step_test_server_slow || fromJSON(needs.detect.outputs.scopes).shared
         run: |
           cd server
           BASE="${{ github.event.pull_request.base.sha }}"
+          LOCKFILE_TOUCHED="${{ fromJSON(needs.detect.outputs.scopes).lockfile_touched }}"
           # On a manual dispatch BASE is empty (no PR) → run the full suite.
-          if [ -n "$BASE" ]; then
+          if [ -n "$BASE" ] && [ "$LOCKFILE_TOUCHED" != "true" ]; then
             npx vitest run --changed "$BASE"
             npx vitest run --config vitest.config.slow.ts --changed "$BASE"
           else
--- a/scripts/ci-scope.mjs
+++ b/scripts/ci-scope.mjs
@@ -25,7 +25,8 @@ export function slugFor(stepName) {
 // `stepTouchedByDiff`'s includeLockfiles branch already routes it there —
 // consumed by verify.yml to force a full (non-`--changed`) run instead of
 // widening `shared`, which was explicitly rejected (#2853) as too broad a fix
-// for this narrow hazard.
+// for this narrow hazard. Note: lockfile_touched fires on (shared || lockfile-specific-match),
+// so it also triggers on non-lockfile shared-scope diffs like .github/actions/**.
 const CI_ONLY = {
   openapi: (files) => files.some((f) => f === 'openapi.yaml'),
   shared: (files) => computeShared(files),
--- a/scripts/ci-scope.mjs
+++ b/scripts/ci-scope.mjs
@@ -21,12 +21,13 @@ export function slugFor(stepName) {
 // cache step. shared is the root-manifest global override. lockfile_touched
 // (#2867, fixing #2853) is a hazard flag, not a leg gate: neither lockfile is
 // in either vitest config's forceRerunTriggers, so a lockfile-only diff can
-// select ZERO tests via `--changed` in the server-tests CI job even though
-// `stepTouchedByDiff`'s includeLockfiles branch already routes it there —
-// consumed by verify.yml to force a full (non-`--changed`) run instead of
-// widening `shared`, which was explicitly rejected (#2853) as too broad a fix
-// for this narrow hazard. Note: lockfile_touched fires on (shared || lockfile-specific-match),
-// so it also triggers on non-lockfile shared-scope diffs like .github/actions/**.
+// select ZERO tests via `--changed` in both the server-tests and frontend-tests
+// CI jobs even though `stepTouchedByDiff`'s includeLockfiles branch already
+// routes it there — consumed by verify.yml to force a full (non-`--changed`)
+// run instead of widening `shared`, which was explicitly rejected (#2853) as
+// too broad a fix for this narrow hazard. Note: lockfile_touched fires on
+// (shared || lockfile-specific-match), so it also triggers on non-lockfile
+// shared-scope diffs like .github/actions/**.
 const CI_ONLY = {
   openapi: (files) => files.some((f) => f === 'openapi.yaml'),
   shared: (files) => computeShared(files),
--- a/scripts/ci-scope.mjs
+++ b/scripts/ci-scope.mjs
@@ -18,10 +18,19 @@ export function slugFor(stepName) {
 }
 
 // openapi gates the CI-only "OpenAPI types up to date" drift check and has no
-// cache step. shared is the root-manifest global override.
+// cache step. shared is the root-manifest global override. lockfile_touched
+// (#2867, fixing #2853) is a hazard flag, not a leg gate: neither lockfile is
+// in either vitest config's forceRerunTriggers, so a lockfile-only diff can
+// select ZERO tests via `--changed` in the server-tests CI job even though
+// `stepTouchedByDiff`'s includeLockfiles branch already routes it there —
+// consumed by verify.yml to force a full (non-`--changed`) run instead of
+// widening `shared`, which was explicitly rejected (#2853) as too broad a fix
+// for this narrow hazard.
 const CI_ONLY = {
   openapi: (files) => files.some((f) => f === 'openapi.yaml'),
   shared: (files) => computeShared(files),
+  lockfile_touched: (files) =>
+    files.some((f) => f === 'package-lock.json' || f === 'server/package-lock.json'),
 };
 
 export function computeScopes(files, { eventName } = {}) {
--- a/scripts/tests/ci-scope.test.mjs
+++ b/scripts/tests/ci-scope.test.mjs
@@ -75,6 +75,27 @@ test('computeScopes is all-false for an unrelated diff on a PR', () => {
   assert.equal(scopes.shared, false);
 });
 
+// #2853/#2867: verify.yml's server-tests job forces a full (non-`--changed`)
+// run off this key rather than trusting `--changed`, since neither lockfile
+// is in either vitest config's forceRerunTriggers.
+test('computeScopes sets lockfile_touched for a server lockfile diff', () => {
+  const scopes = computeScopes(['server/package-lock.json'], { eventName: 'pull_request' });
+  assert.equal(scopes.lockfile_touched, true);
+});
+
+// The root lockfile already flips `shared` (computeShared) — lockfile_touched
+// must also be true here, since it shares CI_ONLY's `shared || fn(files)`
+// formula, so verify.yml can rely on one key for either lockfile.
+test('computeScopes sets lockfile_touched for a root lockfile diff', () => {
+  const scopes = computeScopes(['package-lock.json'], { eventName: 'pull_request' });
+  assert.equal(scopes.lockfile_touched, true);
+});
+
+test('computeScopes leaves lockfile_touched false for a non-lockfile diff', () => {
+  const scopes = computeScopes(['server/src/foo.ts'], { eventName: 'pull_request' });
+  assert.equal(scopes.lockfile_touched, false);
+});
+
 test('render emits GITHUB_OUTPUT lines plus the ok sentinel', () => {
   const out = render({ step_test_hooks: true, shared: false });
   assert.match(out, /^scopes=\{.*\}$/m);
--- a/scripts/tests/workflow-wiring.test.mjs
+++ b/scripts/tests/workflow-wiring.test.mjs
@@ -1080,6 +1080,71 @@ test('ffmpeg install: timeout wrapping is present with correct bounds on all apt
   );
 });
 
+test('lockfile-touched polarity: both frontend and server test steps force full run when lockfile is changed (#2853)', () => {
+  // Mutation proof: a lockfile-only diff can select ZERO tests via `--changed`,
+  // so both steps must force a full run when LOCKFILE_TOUCHED is true.
+  // This test verifies the exact polarity: when LOCKFILE_TOUCHED != "true",
+  // run with --changed; when true (or BASE unset), run full suite.
+  // Reverting to the pre-fix condition (removing LOCKFILE_TOUCHED guard or
+  // inverting the polarity) makes this test fail.
+
+  // Find both test steps by their exact names.
+  const frontendMatch = source.match(
+    /- name: Frontend tests \+ a11y\n\s*if:[^\n]+\n\s*run: \|\n((?:\s{10}[^\n]*\n)*)/,
+  );
+  const serverMatch = source.match(
+    /- name: Server tests \(fast \+ slow\)\n\s*if:[^\n]+\n\s*run: \|\n((?:\s{10}[^\n]*\n)*)/,
+  );
+
+  assert.ok(frontendMatch, 'Frontend tests + a11y step not found');
+  assert.ok(serverMatch, 'Server tests (fast + slow) step not found');
+
+  const frontendBody = frontendMatch[1];
+  const serverBody = serverMatch[1];
+
+  // Both steps must assign LOCKFILE_TOUCHED from the scopes output.
+  assert.match(
+    frontendBody,
+    /LOCKFILE_TOUCHED="\$\{\{\s*fromJSON\(needs\.detect\.outputs\.scopes\)\.lockfile_touched\s*\}\}"/,
+    'Frontend step does not assign LOCKFILE_TOUCHED from detect scopes',
+  );
+  assert.match(
+    serverBody,
+    /LOCKFILE_TOUCHED="\$\{\{\s*fromJSON\(needs\.detect\.outputs\.scopes\)\.lockfile_touched\s*\}\}"/,
+    'Server step does not assign LOCKFILE_TOUCHED from detect scopes',
+  );
+
+  // Both steps must have the correct polarity: when LOCKFILE_TOUCHED != "true",
+  // use --changed; otherwise use full run. The `if` condition checks:
+  // [ -n "$BASE" ] && [ "$LOCKFILE_TOUCHED" != "true" ]
+  // If this is true, run --changed; else run full.
+  const correctFrontendPattern = /if\s+\[\s*-n\s+"\$BASE"\s*\]\s+&&\s+\[\s*"\$LOCKFILE_TOUCHED"\s+!=\s+"true"\s*\]\s*;\s+then\s+npx\s+vitest\s+run\s+--changed/;
+  const correctServerPattern = /if\s+\[\s*-n\s+"\$BASE"\s*\]\s+&&\s+\[\s*"\$LOCKFILE_TOUCHED"\s+!=\s+"true"\s*\]\s*;\s+then/;
+
+  assert.match(
+    frontendBody,
+    correctFrontendPattern,
+    'Frontend step does not have correct LOCKFILE_TOUCHED polarity (should check != "true" for --changed branch)',
+  );
+  assert.match(
+    serverBody,
+    correctServerPattern,
+    'Server step does not have correct LOCKFILE_TOUCHED polarity (should check != "true" for --changed branch)',
+  );
+
+  // Verify the else branch runs the full suite (no --changed).
+  assert.match(
+    frontendBody,
+    /else\s+npx\s+vitest\s+run;/,
+    'Frontend step else branch does not run full vitest suite',
+  );
+  assert.match(
+    serverBody,
+    /else\s+npx\s+vitest\s+run/,
+    'Server step else branch does not run full vitest suite',
+  );
+});
+
 test('ffmpeg install: no bare apt-get install ffmpeg in workflows (use install-ffmpeg action)', () => {
   // Issue #2449: unretried `sudo apt-get update && sudo apt-get install -y ffmpeg`
   // hangs on mirror timeouts and burns entire job timeouts. This regression test
--- a/scripts/tests/workflow-wiring.test.mjs
+++ b/scripts/tests/workflow-wiring.test.mjs
@@ -1138,10 +1138,37 @@ test('lockfile-touched polarity: both frontend and server test steps force full 
     /else\s+npx\s+vitest\s+run;/,
     'Frontend step else branch does not run full vitest suite',
   );
+
+  // For the server step, we need a more structural check because it has
+  // multiple lines in both then and else branches. Extract the entire
+  // if/then/else/fi block and verify:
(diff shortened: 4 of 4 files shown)


dudarenok-maker:
## PR review — pass 1 (head 92cdcd713a355ea7d98ec6ae8fd75098f555dd1e, depth low)

Scope: `.github/workflows/verify.yml`, `scripts/ci-scope.mjs`, `scripts/tests/ci-scope.test.mjs` · verified before probing: `node --test scripts/tests/ci-scope.test.mjs` 16/16 and `node --test scripts/tests/workflow-wiring.test.mjs` 16/16 against an exported copy of the PR head, plus four mutations run against that copy. The PR body's mutation-kill claim was **reproduced and holds** (see ✅ below). Also read #2853, #2867, and Decision A in `docs/superpowers/specs/2026-09-01-verify-scope-branch-timing-design.md`, plus the installed vitest's own `--changed` implementation.

### 🟠 Significant — the identical zero-test hazard is live on the **frontend** leg, and the recorded reason for excluding it is contradicted by this PR's own premise

`.github/workflows/verify.yml:279-285`

Repro, with a real precedent in this repo's history: a PR whose diff is exactly `package-lock.json` — commit `9d6c64a4` *"build(deps): refresh root lockfile to clear 5 npm advisories"* (2026-08-18) is exactly this, 1 file changed.

1. `computeShared` matches root `package-lock.json` → `shared = true`. Verified live against the PR head: `computeScopes(['package-lock.json'], {eventName:'pull_request'})` → `{ shared: true, step_test: true, lockfile_touched: true }`.
2. So the frontend step's `if:` (line 280) fires and line 284 runs `npx vitest run --changed "$BASE"`.
3. `vitest.config.ts`'s trigger is `'{**/package.json,**/.*/**/package.json}'` — it does **not** match `package-lock.json`. That is this PR's own new comment (`verify.yml:328`): *"neither lockfile is in either vitest config's forceRerunTriggers"*.
4. `filterTestsBySource` (`node_modules/vitest/dist/chunks/cli-api.24X8XwN1.js:11484-11507`) therefore keeps only specs whose dependency graph contains a changed file → **0 of 343 frontend test files** — and `--changed` sets `passWithNoTests ??= true` (`coverage.DM_a_rWm.js:459`), so it exits 0.
5. Line 285 then runs `npm run test:a11y` = `vitest run src/test/a11y.test.tsx` — **one** file.

Net: on a root-lockfile-only PR the required check reports green having run 1 of 344 frontend test files against a changed dependency tree. That is #2853's own stated failure mode, at a second site.

I am not reporting this as an oversight — it was **explicitly ruled out** (#2853 body; #2867 "do not touch it"; Decision A, spec lines 130-142). But the ruling rests on two premises and one of them is now false:

- *"`test:a11y` runs unconditionally, so it always does real work"* — true, but "does real work" is not "runs the frontend unit suite". 1 file vs. 344.
- *"Root `package.json`/`package-lock.json` is already NOT part of this hazard (protected today by both vitest configs' `forceRerunTriggers`)"* (#2867) — **false for the lockfile**. The measured proof cited in `server/vitest.config.ts` was for `package.json`, which *is* a trigger; `package-lock.json` is not, which is the whole reason this PR exists. Decision A itself concedes that "a bare root `package-lock.json`-only or `.github/actions/**`-only diff is **still** an unprotected `shared`-scope gap".

If the resolution is "still out of scope", that is a decision, not a fact — it needs re-making against the corrected premise rather than inheriting the old one, and whatever issue it lands in should name the decision owed.

### 🟠 Significant — the behavioural half of the fix can be reverted with every test in the repo still green

`.github/workflows/verify.yml:342`

Mutation-verified against the PR head, not reasoned: delete the `&& [ "$LOCKFILE_TOUCHED" != "true" ]` conjunct from line 342 while keeping the `LOCKFILE_TOUCHED=` assignment on line 340 — i.e. restore the exact #2853 bug — and:

- `node --test scripts/tests/ci-scope.test.mjs` → 16/16 pass
- `node --test scripts/tests/workflow-wiring.test.mjs` → 16/16 pass

`grep -rn "LOCKFILE_TOUCHED" scripts/` returns nothing. `workflow-wiring.test.mjs`'s `<-` assertion only requires the key to be *referenced somewhere in the file*, which the surviving assignment satisfies — so deleting the conjunct **and** inverting its polarity (`!=` → `=`) both sail through.

The PR's mutation-kill is real (I reproduced it), but it kills a mutation of `ci-scope.mjs`'s detector — the half that was never broken. The shipped behaviour change has no test at all. #2867's acceptance asked for exactly this and named the fallback: *"manually trace through the new shell logic with a lockfile-only synthetic diff and a non-lockfile diff, and document both traces (inputs → `LOCKFILE_TOUCHED` value → which vitest invocation runs) in the commit message"* — the commit message documents the `computeScopes` outputs but no shell trace. `workflow-wiring.test.mjs` already does source-scan assertions of exactly this shape on this exact file (the ffmpeg `timeout -k 10 180` wrappers at :1056-1081, the leg-result case arms at :1126-1175, the sentinel step order at :118-152), so the harness exists.

### 🟡 Minor — the emitted key's real semantics are broader than its name, its comment and its tests, and both halves of that are unpinned

`scripts/ci-scope.mjs:32-33` and `:55`

`scopes[key] = key === 'shared' ? shared : shared || fn(files)` makes `lockfile_touched` mean `shared || lockfile`, so it is `true` when **no lockfile was touched**. Verified: `computeScopes(['.github/actions/setup/action.yml'])` → `lockfile_touched: true`. That over-fire is arguably *good* — it happens to close the `.github/actions/**`-only server-side gap Decision A names as still unprotected — but the comment at `:20-28` describes a pure lockfile flag, so the behaviour is undocumented, and neither half is pinned:

- **Mutation D** — `key === 'lockfile_touched' ? fn(files) : shared || fn(files)` (drop the `shared` disjunct): ci-scope 16/16 green, workflow-wiring 16/16 green. The `.github/actions/**` coverage vanishes silently.
- **Mutation B** — drop `f === 'package-lock.json'` from the detector at `:33`, leaving only the server lockfile: ci-scope 16/16 green. The new *"sets lockfile_touched for a root lockfile diff"* case passes via `shared` either way (its own comment says so honestly), and the pre-existing *"sets shared for a root lockfile diff"* case three lines above already asserts that same fact — so the detector's own root-lockfile disjunct has no test that can fail.

A later reader tidying either direction — "the name says lockfile, drop the `shared` OR" or "the root lockfile is already `shared`, drop the redundant disjunct" — reopens a gap with every check green.

### 🟡 Minor — the frontend leg's own comment is now false, and it is the comment that would mislead the next reader

`.github/workflows/verify.yml:268-271`: *"a root dep/config bump (the `shared` scope) forces a full run via vitest's default triggers."* False for the lockfile half of `shared` and for `.github/actions/**` — exactly the fact this PR establishes one screen further down the same file. Left as-is, it tells the next person the frontend leg is already covered.

### 🟡 Minor — release-notes step neither done nor explicitly waived

No `docs/release-notes-next.md` / `RELEASE_NOTES.md` entry, and neither the PR body nor the commit message says why. A CI-only chore with no operator-visible delta is legitimately skippable, but CLAUDE.md before-shipping step 5 asks for that to be stated rather than silently omitted.

### ✅ What is solid

- **The mutation-kill claim in the PR body is true.** Replacing the detector body with `() => false` makes *"computeScopes sets lockfile_touched for a server lockfile diff"* fail (`false !== true`) and leaves the other 15 green — reproduced exactly as described.
- **Both lockfile paths are genuinely covered at runtime**: `['server/package-lock.json']` → `lockfile_touched: true` (via the detector), `['package-lock.json']` → `true` (via `shared`). `package-lock.json` and `server/package-lock.json` are the only two npm lockfiles `git ls-files` finds; `apps/android/pubspec.lock` has no vitest exposure.
- **The workflow expression and shell are correct in all three contexts.** The step body's two branches are exhaustive because `on:` is only `pull_request` + `workflow_dispatch`: PR + lockfile → full; PR without → `--changed`; dispatch → `$BASE` empty → full (and `computeScopes` is all-true on dispatch, so the key agrees). The interpolated value is our own boolean, so `"${{ … }}"` carries no injection surface, and a typo'd key would be caught by `workflow-wiring.test.mjs`'s `->` assertion rather than silently failing open.
- **Reusing `detect`'s blob instead of a second `git diff` is right, and the two bases actually agree** — `detect` diffs `merge-base(BASE,HEAD)`, and vitest's `--changed <sha>` runs `git diff --name-only <sha>...HEAD` (triple-dot, `cli-api.24X8XwN1.js:12752-12758`), i.e. the same merge-base semantics. The new comment's claim of "the same PR diff" holds.
- **The declared non-goals hold**: `computeShared`, `STEPS[]` and both configs' `forceRerunTriggers` are untouched; `Closes #2853` is present and outside any code span; single-scope `chore(ops)` matches depth `low`; no on-box acceptance is owed.

### Verdict

The fix is correct for the site it targets and the detector is honestly tested. Two things block a clean pass: the shipped conditional is protected by no test in the repo (mutation-proven), and the second live instance of the same green-on-nothing class — the frontend leg on a root-lockfile-only PR, a diff shape that has already landed on `main` once — is excluded on a premise this PR itself falsifies. The second is a decision the operator has already made once and may reasonably re-make; it should be re-made explicitly rather than carried over.

dudarenok-maker:
## PR review — pass 2 (head 2b110850f5029a2e006adc36a18962cab4b978cb, depth medium)

Depth re-derived from the current commit set: `chore(ops)` + `fix(ops)`, both single-scope → **medium**.

Deltas only. Everything verified below was re-derived from the files at this head, not read off the fix commit's message.

---

**🟠 Finding 1 (frontend lockfile zero-test hazard) — ✅ VERIFIED RESOLVED.**

`.github/workflows/verify.yml:284-286` now reads the *same* `lockfile_touched` key already threaded from `detect` (no second `git diff`/`merge-base` in the step), and the condition is `if [ -n "$BASE" ] && [ "$LOCKFILE_TOUCHED" != "true" ]`. Re-derived the hazard end to end: `vitest.config.ts:85-88`'s `forceRerunTriggers` covers `{**/package.json,…}` but **not** `package-lock.json`; `computeShared` (`scripts/verify-cache.mjs:834-839`) is `package.json || package-lock.json || .github/actions/`, so a root-lockfile-only diff *does* fire the step's `if:` via `shared` — which is exactly the "step runs, selects 0 of ~343 files, reports green" path. It now takes the `else` (full) branch. YAML/shell both valid; `fromJSON(...)` yields a JSON boolean so the interpolation is the literal `true`/`false`, and an absent key would render `""` (fail-open) — but I confirmed that shape is caught statically (see M10 below).

`npm run test:a11y` (`verify.yml:287`) is untouched and still outside the `if`. ✅

**🟠 Finding 2 (regression test with no mutation-kill power) — ⚠️ HALF RESOLVED. Resolved for the frontend step; still open for the server step.**

I ran the mutations rather than trusting the reported output — scratch worktree at this head, mutate `verify.yml`, `node --test scripts/tests/workflow-wiring.test.mjs`, revert. Baseline 17/17 green.

Killed ✅ (7):
| # | Mutation | Result |
|---|---|---|
| M1 | frontend: delete `&& [ "$LOCKFILE_TOUCHED" != "true" ]` (exact pre-fix #2853 bug) | killed — `AssertionError: Frontend step does not have correct LOCKFILE_TOUCHED polarity` (the claimed message, confirmed) |
| M2 | server: delete the same conjunct | killed |
| M3/M4 | frontend/server: invert to `== "true"` | killed |
| M5 | frontend: hardcode `LOCKFILE_TOUCHED="false"` | killed |
| M9 | server: `&&` → `\|\|` | killed |
| M10 | both: typo the key to `.lockfiles_touched` (renders `""`, guard never fires) | killed — by the pre-existing `-> every referenced scope key is emitted by ci-scope.mjs` test |

**Survived ❌ (2) — both on the server step, both restoring or worsening the bug this PR exists to prevent.** See new finding N1.

**🟡 Finding 3 (`ci-scope.mjs` comment) — ✅ VERIFIED RESOLVED and accurate.** `scripts/ci-scope.mjs:28-29`'s "fires on (shared || lockfile-specific-match) … also triggers on non-lockfile shared-scope diffs like `.github/actions/**`" checks out against `computeShared`'s actual body.

**🟡 Finding 4 (stale `verify.yml` comment) — ✅ landed, but the replacement is itself imprecise.** See new finding N3.

**🟡 Finding 5 (release notes) — ✅ resolved in substance.** The not-applicable is stated in `2b110850`'s commit message ("CI-infra only"), and CI-only is a listed skip case in CLAUDE.md step 5. The PR *body* still doesn't say it — folded into N2 rather than re-raised.

**Out-of-scope items confirmed untouched.** `gh pr diff --name-only` is exactly 4 files: `verify.yml`, `ci-scope.mjs`, and the two test files. `computeShared`, `STEPS[]`, and both `forceRerunTriggers` blocks are byte-identical to base. ✅

**Gates re-checked.** `npm run typecheck` green; the diff adds nothing to either TS program (`tsconfig.json` includes `src`, `vite.config.ts`, `vitest.config.ts`; the diff is one `.yml` + three `scripts/**/*.mjs`), so it is unaffected by construction. `node --test scripts/tests/ci-scope.test.mjs` 16/16, `workflow-wiring.test.mjs` 17/17. The new guard is *reachable*: `test:hooks`' STEP globs include `.github/workflows/**` and `scripts/**/*.mjs`, so a verify.yml-only diff schedules the leg that runs it. `Closes #2853` present and outside a code span.

---

### New findings

#### 🟠 N1 — correctness bug (test): the server half of the new polarity test is branch-blind, so two mutations that restore (or invert) the #2853 bug leave it green

`scripts/tests/workflow-wiring.test.mjs:1122` and `:1141-1145`.

The frontend and server assertions were not mirrored. The frontend pair pins *which branch does what*:

```js
const correctFrontendPattern = /… \]\s*;\s+then\s+npx\s+vitest\s+run\s+--changed/;   // then-branch pinned
assert.match(frontendBody, /else\s+npx\s+vitest\s+run;/, …);                          // trailing `;` pins the else
```

The server pair pins neither:

```js
const correctServerPattern = /… \]\s*;\s+then/;              // stops at `then` — says nothing about the branch
assert.match(serverBody, /else\s+npx\s+vitest\s+run/, 'Server step else branch does not run full vitest suite');
```

`/else\s+npx\s+vitest\s+run/` is unanchored, so it matches `else npx vitest run --changed "$BASE"` just as happily as the full run. The comment above it at `:1135` says "Verify the else branch runs the full suite **(no `--changed`)**" — that parenthetical is not implemented.

Demonstrated, not theorised:

- **M7** — neuter the server `else` branch to `npx vitest run --changed "$BASE"` / `npx vitest run --config vitest.config.slow.ts --changed "$BASE"`, leaving the guard intact. This is byte-for-byte the pre-fix #2853 behaviour: a root-lockfile-only PR reaches the `else`, runs `--changed`, selects 0 tests, reports green. **Test still passes.** The identical mutation on the frontend step (M8) *is* killed, by that trailing `;`.
- **M11** — swap the server step's `then` and `else` bodies, so `--changed` runs *exactly when the lockfile IS touched*. Strictly worse than the original bug. **Test still passes** — `correctServerPattern` only needs the `if` line's text, and the `else` regex matches the `--changed` line that moved into it.

`scripts/tests/workflow-wiring.test.mjs` is the only guard over this step (`grep -rln "Server tests (fast + slow)" scripts/ server/src/` returns that file alone), so nothing else catches either mutation.

Minimal fix: make the server assertions mirror the frontend's — pin the then-branch through `npx vitest run --changed`, and assert the `else` **slice** contains no `--changed` (e.g. slice from `/\n\s*else\n/` to `/\n\s*fi/` and `assert.doesNotMatch(elseSlice, /--changed/)`), rather than matching a prefix that a `--changed` suffix satisfies.

#### 🟡 N2 — cleanup: the PR body's Test plan now contradicts the diff

`2b110850` changed the frontend test step and added a test file, but the PR body still asserts:

- "Verified the diff touches only `ci-scope.mjs`'s `CI_ONLY` map and `verify.yml`'s **server-tests step** shell logic" — now false; it also touches the frontend step and two test files.
- "Verified the **frontend `test`/`test:a11y` step is untouched**" — now false for `test` (true only for `test:a11y`).

Merges here are merge commits, so the PR body is what lands in history. The Summary also doesn't mention the frontend fix at all, and per CLAUDE.md the incidental fixes folded in this round should be declared there ("Also fixed, found in passing: …") — that is also the natural home for Finding 5's CI-only not-applicable line, which currently lives only in a commit message.

#### 🟡 N3 — cleanup: the replacement `verify.yml` comment understates when the flag fires

`.github/workflows/verify.yml:270-271` (and the parallel block at `:329-335`): "a **lockfile-only** diff forces a full run via the LOCKFILE_TOUCHED flag (#2853)."

`lockfile_touched` is `shared || <lockfile match>` (`ci-scope.mjs:56`), and `shared` is `package.json || package-lock.json || .github/actions/**`. So a diff touching **only** `.github/actions/setup/action.yml` — no lockfile anywhere — now forces the full ~343-file frontend suite *and* the full server suite (fast + slow config), where before it took `--changed` and selected ~0. That is a real, safe-direction, but undocumented-at-this-site wall-clock change. `ci-scope.mjs:28-29` states it precisely (Finding 3's fix); `verify.yml` — the file a maintainer reads when wondering why an actions-only PR now takes 20 minutes — does not. Half the pattern got mirrored. Not asking for the flag to be narrowed (that would be a design change); asking the two `verify.yml` comment blocks to say what `ci-scope.mjs` already says.

---

### Re-review trigger

N1 is a 🟠 correctness bug, so fixing and pushing it re-triggers a pass (this was pass 2 of 3).

<sub>Checked and clean, not re-listed: no `cast.json` writes, no config knob, no derived artifact, no hardware-provable behaviour (no on-box register row owed), issue link present.</sub>

dudarenok-maker:
## PR review — pass 3 of 3, FINAL (head `6348c7fcf32d4f76be2d4a707a2fb30bcabd5858`, depth `medium`)

Scope: cumulative diff `main`…`6348c7fc` — `.github/workflows/verify.yml`, `scripts/ci-scope.mjs`, `scripts/tests/ci-scope.test.mjs`, `scripts/tests/workflow-wiring.test.mjs` (4 files, zero TS surface).

Verified before probing: `node --test scripts/tests/workflow-wiring.test.mjs` → 17/17; `node --test scripts/tests/ci-scope.test.mjs` → 16/16 — both run against a clean `git archive` export of head `6348c7fc`, not a working tree. `npm run typecheck` → clean. `computeShared`, `STEPS[]` and both vitest configs' `forceRerunTriggers` confirmed untouched (they appear in no hunk of the 4-file set). **16 mutations** applied to the exported head tree: 12 killed, 4 survived (two ordering — N5 below; two that drop the slow-config invocation, outside the contract this test declares).

Re-review, so deltas only — solid items from passes 1–2 are not re-listed.

---

#### 🟠 N1 (server-step polarity guard was branch-blind) — **VERIFIED RESOLVED**

Reproduced the kill independently rather than taking the claim:

| Mutation | Result | Killed by |
|---|---|---|
| **M7** — server `else` neutered back to `npx vitest run --changed "$BASE"` + slow equivalent (the exact pre-#2853 bug) | **KILLED** | `Server step: else branch contains --changed (should run full suite when lockfile IS touched)` — `scripts/tests/workflow-wiring.test.mjs:1171` |
| **M11** — server `then`/`else` bodies swapped | **KILLED** | `Server step: then branch does not use --changed (should run when lockfile NOT touched)` — `:1164` |

Eight further mutations I invented, **all killed**: `!=` → `=`; dropping the `&& [ "$LOCKFILE_TOUCHED" != "true" ]` conjunct entirely; `&&` → `||`; frontend `then`/`else` swap; frontend `else` neutered to `--changed`; server var sourced from `.shared` instead of `.lockfile_touched`; comparing against `"1"` instead of `"true"`; frontend assignment deleted. Two `ci-scope.mjs` mutations also killed — detector → `() => false`, and detector narrowed to the root lockfile only — which proves the "root lockfile via `shared`" case is not carrying the server-lockfile case vacuously.

#### 🟡 N2 (stale PR body) — **PARTIALLY RESOLVED** (still open)

The two Test-plan bullets named in pass 2 are now correct. The **Summary paragraph is not**: it still reads *"threads it into the server-tests step"*, describing half the change — `2b110850` also threads it into the frontend step (`.github/workflows/verify.yml:286,288`). The body also never declares the folded-in incidental fixes (the two comment updates) or the explicit "no release-notes delta, CI-infra only" call, both of which CLAUDE.md's before-shipping step 5 / incidental-findings rule require in the *PR body*, not only the commit message. An unannounced frontend-step edit reads as scope creep to a reviewer. Cleanup-only; one sentence.

#### 🟡 N3 (imprecise `verify.yml` comment) — **VERIFIED RESOLVED**

`verify.yml:270-273` now states `lockfile_touched` is computed as `(shared || lockfile-specific-match)` and names `.github/actions/**` as an example that also forces a full run. Cross-checked against `scripts/ci-scope.mjs:56` (`shared || fn(files)`) and `computeShared` (`scripts/verify-cache.mjs:834-839` — root `package.json`, root `package-lock.json`, `.github/actions/**`). Accurate.

---

### New findings

#### 🟠 N4 — correctness (pre-existing sibling): the class this PR fixes is still armed **on the very step it hardened**, with a *measured* zero-test green

`.github/workflows/verify.yml:288` (Frontend tests + a11y). Concrete repro, measured in this checkout rather than reasoned:

1. `index.html` is in the `test` step's `extraFiles` (`scripts/verify-cache.mjs` — `test` → `extraFiles: ['vitest.config.ts','vite.config.ts','index.html','openapi.yaml']`), so `stepTouchedByDiff` sets `step_test: true` and this step's `if:` fires.
2. `index.html` is **not** in `vitest.config.ts`'s `forceRerunTriggers` (`vitest.config.ts:85-116` lists only `package.json`, `{vitest,vite}.config.ts`, `src/test/setup.ts`, `openapi.yaml`, `src/lib/api.ts`, `src/lib/api-types.ts`), and nothing imports it.
3. Measured: `npx vitest related index.html --run` → **`No test files found, exiting with code 0`** — the same phrasing `server/vitest.config.ts`'s own comment cites as proof of this hazard.
4. `src/index-html-fonts.test.ts:13` `readFileSync`s `index.html` at runtime (the #698 self-hosted-font guard). It never runs. `npm run test:a11y` on the next line still runs, so the leg reports green.

So: an `index.html`-only PR schedules the frontend leg, selects **zero** tests, and the required check passes — identical in shape and consequence to #2853, one file over.

Second, weaker instance on the same step: `src/styles.css` matches `test`'s glob `src/**` → `step_test: true`. Measured: `npx vitest related src/styles.css --run` → **1** test file (`src/main.test.tsx`, reaching it via `await import('./main')` → `import './styles.css'`). The two guards that exist *for that file* — `src/styles-neutrals.test.ts:10` and `src/test/dark-mode-css.test.ts:24-25`, both `readFileSync` — are not selected. CLAUDE.md's Commit-gate section already names `src/styles.css` as the fourth spelling of this trap found locally; the **local** runner closes it via `diffSafeForChangedOnly`'s positive allowlist (`scripts/verify-cache.mjs:892-901`), CI does not.

Related drift found while checking: `src/test/force-rerun-triggers.test.ts:64`'s `COVERED` list omits `src/lib/api.ts` and `src/lib/api-types.ts`, so those two triggers could be deleted with that guard green — the same hand-maintained-enumeration shape.

**Why this is in scope, not "pre-existing":** commit `2b110850` edited this exact step for exactly this failure mode and closed one spelling of it. `findings-triage.md` voids "it's pre-existing" / "it's next door" as deferral reasons. **This does not block merging #2877** — the PR strictly reduces the hazard and introduces none of this.

**Likely a genuine design-pass carve-out**, so if it becomes an issue, name the decision: (a) add per-file `forceRerunTriggers` entries plus matching `COVERED` cases; (b) derive the frontend guard's `COVERED` list from `STEPS`' `extraFiles`/`globs` so the two cannot drift; or (c) port the local positive-allowlist (`diffSafeForChangedOnly`) into CI so the class is *closed* rather than enumerated — which is what the local runner concluded after four rounds of chasing spellings.

#### 🟡 N5 — cleanup: the new polarity test is **order-blind**; two mutations survive at head

`scripts/tests/workflow-wiring.test.mjs:1083-1173` asserts each step body *contains* the `LOCKFILE_TOUCHED=` assignment (`:1109`, `:1114`) and *contains* a correctly-polarised `if` (`:1162-1171`), but never that the assignment **precedes** the `if`. Both of these survive with 17/17 green:

- **N11** — move `verify.yml:344`'s assignment to after the server `fi`. At runtime `LOCKFILE_TOUCHED` is unset at line 346, `[ "" != "true" ]` is true, the guard is inert, and #2853 is restored verbatim.
- **N12** — the same move on the frontend step (`:286` → after `:288`), same inert outcome.

Fix is one line per step: compare `indexOf` of the assignment against `indexOf` of the `if` inside each captured body.

Noted, *not* a finding: the same test does not pin that the server `then` branch invokes **both** configs — dropping `npx vitest run --config vitest.config.slow.ts --changed "$BASE"` leaves it green. That is outside the `--changed`-vs-full polarity contract the test declares, and that contract is now fully covered.

#### 🟡 N6 — cleanup: `scripts/ci-scope.mjs:24` understates its own consumer

The comment says the hazard is *"in the server-tests CI job"* and that the flag is consumed to force a full run there. As of `2b110850` the flag has **two** consumers (`verify.yml:288` and `:346`). N3 fixed the precision of the *formula* in two places; this sentence is the same class one file over — a comment this PR's own second commit made incomplete.

---

### ✅ Re-verified this pass (not re-listed from earlier passes)

- **Shell/YAML form, final state.** `if [ -n "$BASE" ] && [ "$LOCKFILE_TOUCHED" != "true" ]` is valid POSIX `sh` in both steps; the interpolated value is a JSON boolean emitted by our own script, so it is exactly `true`/`false` — no quoting hazard and no injection surface. On `workflow_dispatch` `computeScopes` returns all-true → `LOCKFILE_TOUCHED=true` and `BASE=""`; both conjuncts fail → full run, behaviour unchanged from before this PR.
- **Class coverage inside CI.** `grep -rn -- --changed .github/` finds exactly two invocation sites (`:288`, `:347-348`); both are now guarded. The local `--changed` path cannot admit a lockfile at all — `diffSafeForChangedOnly` requires every diff file to sit under `src/`/`server/src/` with a JS/TS extension.
- **Premise of the fix, independently checked.** `package-lock.json` / `server/package-lock.json` appear in none of `vitest.config.ts:85`, `server/vitest.config.ts:190`, `server/vitest.config.slow.ts:88`. `package.json` *is* in all three, so the manifest-only case was already safe and only the lockfile spelling was open — the fix targets exactly the real gap.
- **Wiring guards cover the new key** (`emitted ⊆ referenced` and `referenced ⊆ emitted`), so a rename on either side reddens rather than silently disabling the guard.
- **House gates.** `Closes #2853` present, plain, outside any code span; #2853 (open) and #2867 (closed) both exist and match the work. No release-notes pair owed (CI infra, no shippable delta) — but see N2, it is stated only in the commit, not the PR body. No on-box acceptance, no `cast.json` writes, no new config knob, no derived artifacts staled.
- **Tree check.** `git rev-parse HEAD` and `git status --porcelain` identical before and after this pass; all mutation work was done on a `git archive` export under `%TEMP%`.

---

### Verdict

**Safe to merge.** The pass-2 blocker is genuinely closed — verified by reproducing both named mutations plus eight of my own, all killed, against the actual head tree rather than the commit message. Nothing in the shipped `verify.yml` at `6348c7fc` is wrong.

One 🟠 (**N4**) stands, and it is a pre-existing sibling instance rather than a regression from this change. Per the loop cap this hands to you rather than opening a pass 4: decide whether N4 is fixed in this round or filed as its own design-pass issue naming the decision above. **N2**, **N5** and **N6** are 🟡 — fixable in-round without re-triggering review.

PR #2879, opened 2026-09-04
fix(ops): arm idSpecificAnnotationPresent with the shared polarity check (#2858)

### LINKED ISSUE #2858: check-register-citations: idSpecificAnnotationPresent has no polarity check (Check A fatal path)

dudarenok-maker:
## Summary

`idSpecificAnnotationPresent` (Check A's fatal-path citation matcher) was
polarity-blind: it only asked whether a discharge word and a specific ID
token co-occur in a clause, not what the discharge word actually asserts. A
negated discharge sentence ("was NOT discharged; it is still live") could be
misread as affirmatively excusing a citation from the nonexistent-ID gate.
Extracts `dischargedSubjectsMentionedIn`'s backward-scan + 30-char-capped
negation check (`POLARITY_PROXIMITY_CHARS`, same negation regex + `un-?discharged`
check) into a shared `isDischargeAssertionNegated` helper and wires
`idSpecificAnnotationPresent` to it too. This is latent hardening, not a
functional change to today's docs: no real corpus citation changes
classification — `check-register-citations.mjs --strict` produces
byte-identical output against the primary checkout (unfixed, `e8c2474e`) and
this branch (fixed). Closes #2858.

## Test plan

- [x] `node --test scripts/tests/check-register-citations.test.mjs` — 121/121 pass, including the new synthetic regression fixture (finding #2858) and its paired affirmative control
- [x] Re-ran the mutation kill myself: reverted the `isDischargeAssertionNegated` call in `idSpecificAnnotationPresent`, confirmed the new test fails with the pre-fix behavior (1 annotated instead of 0), then restored the fix (working tree clean afterward)
- [x] Re-ran the no-op-on-real-corpus check myself: `node scripts/check-register-citations.mjs --strict` against the primary checkout (`main`@e8c2474e, unfixed) and this branch (fixed) — byte-identical output, same annotated/fatal counts
- [x] `npm run typecheck` — green
- [x] Diff confirmed code-only: `scripts/check-register-citations.mjs` + `scripts/tests/check-register-citations.test.mjs`, no changes under `docs/testing/**` or `docs/features/**`

5. **Release notes**: skipped — this is an internal audit-script hardening fix with no user- or operator-visible behavior change (verified byte-identical `--strict` output against the real corpus before/after, see Test plan above).

### DIFF
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -739,6 +739,11 @@ const ANNOTATION_WINDOW_LINES = 25;
 // paragraph.
 const ID_PROXIMITY_CHARS = 120;
 
+// How close a negation word must sit BEFORE a discharge-word match for
+// isDischargeAssertionNegated to treat it as negating that match — see that
+// function's own comment for why 30 (not the full clause) is the right cap.
+const POLARITY_PROXIMITY_CHARS = 30;
+
 // A bare ID token, used to find every ID mention in a section so a discharge
 // word can be bound to the right one(s) — see idSpecificAnnotationPresent's
 // own comment for why "within 120 chars" alone isn't enough.
@@ -883,6 +888,42 @@ function clauseBounds(text, pos) {
 }
 
 /**
+ * Whether the discharge-word match `dm` (against `scanText`, a
+ * stripInlineCodeSpans'd copy) is negated — e.g. "was NOT discharged" or
+ * "un-discharged" — rather than an affirmative discharge/removal assertion.
+ * `clauseStart` bounds the backward scan to the SAME clause `clauseBounds`
+ * already computed for `dm`, so an unrelated negation across a clause/
+ * list-item boundary cannot bleed in.
+ *
+ * Shared by `idSpecificAnnotationPresent` (Check A's fatal-path citation
+ * matcher) and `dischargedSubjectsMentionedIn` (Check C's own-history scan) —
+ * extracted from the latter, pass-11/pass-15/pass-16 review of PR #2846 (see
+ * `dischargedSubjectsMentionedIn`'s own comment for the full history of why
+ * this exists and why the window is backward-only and 30-char-capped, not
+ * clause-wide or bidirectional): a "clause" here is a full paragraph, since
+ * CLAUSE_BOUNDARY_REGEX does not treat ". " as a boundary, and the register's
+ * real affirmative discharge records measure their genuine negations at
+ * 4-10 chars back vs. false suppressions at 85+ chars back in the same
+ * paragraph. The negation-word set is deliberately broad (NOT/no/never/
+ * isn't/aren't/wasn't/weren't/hasn't/haven't/cannot/can't/an "un-" prefix
+ * fused to the discharge word/"far from"/"nothing"), since a first cut of
+ * just "NOT"/"no" missed most real negation phrasings.
+ */
+function isDischargeAssertionNegated(scanText, dm, clauseStart) {
+  const polarityScanStart = Math.max(clauseStart, dm.index - POLARITY_PROXIMITY_CHARS);
+  const polarityContext = scanText.slice(polarityScanStart, dm.index);
+  // Apostrophes are '-escaped (not literal) so this regex literal
+  // doesn't desync spawn-windows-hide.test.ts's quote-tracking scanner --
+  // see that guard's own regexLiteralDesyncsQuoteTracking comment.
+  const hasNegationWord =
+    /\b(?:NOT|no\b(?!\s+longer\s+exists?)|never|isn\u0027t|aren\u0027t|wasn\u0027t|weren\u0027t|hasn\u0027t|haven\u0027t|cannot|can\u0027t|far\s+from|nothing)\b/i.test(
+      polarityContext,
+    );
+  const isUnPrefixed = /un-?discharged\b/i.test(scanText.slice(Math.max(0, dm.index - 10), dm.index + dm[0].length));
+  return hasNegationWord || isUnPrefixed;
+}
+
+/**
  * Whether `sectionText` contains a discharge/removal annotation that
  * actually references `id` — not merely a discharge word somewhere in the
  * section, which a paired injection proved lets unrelated prose excuse a
@@ -952,6 +993,14 @@ function clauseBounds(text, pos) {
  * that real annotation (measured: `docs/superpowers/plans/2026-08-05-
  * device-token-scope.md`'s `F2`/`F3` citations went from correctly-excused
  * to falsely-flagged the first time this was tried).
+ *
+ * #2858: also polarity-checks the discharge word via the same
+ * `isDischargeAssertionNegated` helper `dischargedSubjectsMentionedIn` uses,
+ * so a negated sentence ("Register row: A18 — was NOT discharged; it is
+ * still live") can no longer be misread as affirmatively excusing A18 from
+ * Check A's fatal nonexistent-ID gate. Measured against the real corpus:
+ * this flips zero of today's citations — see #2858 for the false-positive
+ * analysis of the 4 citations a looser, unshared polarity scan misflagged.
  */
 function idSpecificAnnotationPresent(sectionText, id) {
   const dischargeScanText = stripInlineCodeSpans(sectionText);
@@ -961,6 +1010,7 @@ function idSpecificAnnotationPresent(sectionText, id) {
   if (dischargeMatches.length === 0) return false;
   for (const dm of dischargeMatches) {
     const { start, end } = clauseBounds(sectionText, dm.index);
+    if (isDischargeAssertionNegated(dischargeScanText, dm, start)) continue;
     ANY_ID_TOKEN_REGEX.lastIndex = start;
     let occ;
     while ((occ = ANY_ID_TOKEN_REGEX.exec(sectionText)) && occ.index < end) {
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
## PR review — pass 1 (head bc786dd168a2743921b7688a9a9c7dbea6f8def3, depth low)

Adversarial gate pass. Every claim in the PR body was re-derived independently rather than taken on faith; all four hold, and one of them holds on a *newer* corpus than the author measured against. One correctness bug (pre-existing, adjacent, same defect class), two cleanup nits, and one merge-blocking CI state.

---

### Claims verified independently

| Claim | Verdict | How |
|---|---|---|
| 121/121 tests pass | ✅ | Ran `node --test scripts/tests/check-register-citations.test.mjs` in a throwaway detached worktree at `bc786dd1` — `tests 121 / pass 121 / fail 0`. |
| Mutation kill (reverting the wiring fails the new test) | ✅ **and two-sided** | Deleting `if (isDischargeAssertionNegated(dischargeScanText, dm, start)) continue;` fails the new test with `actual: 1, expected: 0` on `annotated.length` — red for exactly the reason claimed (pre-fix behaviour = negation misread as excusing). Separately, forcing the helper to `return true` (over-firing) fails the **paired control** instead. The pair pins both directions; neither test can pass vacuously. |
| Byte-identical `--strict` output on the real corpus | ✅ **stronger than claimed** | The author measured against the merge base (`e8c2474e`). I re-measured against **today's `main` (`ff28641e`)**, which has since moved: imported main's module and the PR's module side by side and replayed the CLI's own per-file loop over all 3,926 scanned files. Check A: `errors=0 annotated=38` — lists **identical**. Check C (`checkConflictingSubjects`) also identical across all three buckets (`wrongId=0`, `unknownSubject=1`, `annotatedDischarge=4`), as are `measureWrongIdEligibleLines` and `findUnclassifiedRunSheetMentions`. |
| Genuine refactor, not a silent behaviour change | ✅ | The negation regex literal and the `isUnPrefixed` line are **byte-identical** between `main` and this branch, and each occurs **exactly once** in both files (no leftover second copy). `POLARITY_PROXIMITY_CHARS` moved function-local → module-level with the same value `30`. Both consumers of `idSpecificAnnotationPresent` (`:1157`, `:1709`) route through the one function, so there is no third entry point that bypasses the new check. |

Also checked and **not** a finding: the new call site passes `clauseStart` from `clauseBounds(sectionText, …)` (unstripped) while the helper slices `dischargeScanText` (stripped). `stripInlineCodeSpans` is length-preserving and can only *remove* boundary characters, never create them, so the unstripped `start` is always ≥ the stripped one — the scan window can only ever be narrower, never wider. Fail-open in a vanishing edge case, never a false fire.

---

### Correctness bugs

**1 — SIGNIFICANT · the class is fixed in one script and left armed in its sibling · `scripts/check-register-row-citations.mjs:97` (consumed at `:388`)**

This PR closes #2858 by arming `idSpecificAnnotationPresent` with the polarity check. The *sibling* register-citation checker carries the same polarity-blind annotation exemption, untouched:

```js
// scripts/check-register-row-citations.mjs:97
const ANNOTATION_REGEX = /\bdischarged?\b|\bno longer exists?\b|\bremoved from the register\b/gi;
```

No negation handling exists anywhere in that file. Its exemption downgrades a **fatal** nonexistent-row citation to a non-fatal note (`runCheckRegisterRowCitationsCli`: `failures` → `CliExitError(1)`, `annotated` → `console.log`), which is the same fatal path #2858 names — just in the other checker.

Concrete failure scenario, executed against the real register via the module's own injectable seam (`checkRegisterRowCitations({ files, readFile })`):

| Fixture line (file scanned as `docs/testing/…`) | `failures` | `annotated` |
|---|---|---|
| `> Register row A999 — was NOT discharged; it is still live.` | **0** | **1** |
| `> Register row A999 — was discharged; it is no longer live.` | 0 | 1 |
| `> Register row A999 — outstanding work.` (control) | **1** | 0 |

The negated line and the affirmative line are indistinguishable to the checker: a stale citation to a nonexistent row escapes the fatal gate purely because the word "discharged" appears in a sentence that *denies* the discharge. This is reachable, not theoretical — `scripts/tests/check-register-row-citations.test.mjs:607` asserts the real-tree CLI exits 0, and that suite runs under `test:hooks`, which is in `verify:fast:branch` and `verify`.

Latency is the same as this PR's own fix: I ran the sibling checker against the live tree (`failures=0, annotated=7, 40 citations across 348 files`) and none of the 7 is currently excused by negated prose. So this is a latent gate-defeat, not a live miss — but it is exactly the "one instance fixed, the class left armed" shape, in the immediate next-door file, addressing the identical defect the issue this PR closes is *about*. Under CLAUDE.md's incidental-findings rule ("code next door that the work surfaced; adjacency is not the test") this is owed in this round, with its own paired regression test, not filed for later.

---

### Cleanup nits

**2 — the release-notes gate is silently omitted rather than explicitly waived · PR body**

Before-shipping step 5 requires either the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair, or an explicit not-applicable — "say so explicitly rather than silently omitting". The branch touches neither file, and the body never declares the step N/A. Notable because every *other* checklist item is explicitly accounted for in the test plan (typecheck, tests, mutation kill, corpus no-op, diff scope), so this one reads as an oversight rather than a judgement. An internal tooling-only chore plausibly qualifies for the waiver — it just has to be written down.

**3 — the 30-char cap's justification is Check-C-measured and reused for Check A without re-measurement · `scripts/check-register-citations.mjs:742-745` and the helper's docstring**

`POLARITY_PROXIMITY_CHARS` and the new helper's comment carry Check C's evidence verbatim — "the register's real affirmative discharge records measure their genuine negations at 4-10 chars back vs. false suppressions at 85+ chars back". That was measured over **register row bodies**. Check A's `idSpecificAnnotationPresent` runs over arbitrary repo prose and source — `.md`, `.ts`, `.py`, `.json` — bounded by a 60-line section or a 25-line window. The cap is not re-measured for that wider domain; the comment presents the Check-C figure as if it were.

The gap is demonstrable. Run through both module versions:

| Section text | `main` | this branch |
|---|---|---|
| `Register row: A999 — Nothing further owed - A999 discharged 2026-08-11.` | annotated | **FATAL** |
| `Register row: A999 — never mind, A999 discharged 2026-08-11.` | annotated | **FATAL** |
| `Register row: A999 — no longer needed; A999 discharged 2026-08-11.` | annotated | annotated (`;` bounds the clause) |

Both flips are legitimate affirmative annotations turned fatal by an unrelated negation word inside the 30-char window. **This does not block**: zero such shapes exist in today's corpus (measured above), and the failure direction is loud — a false suppression turns the checker red, it does not silently pass. It is a comment-accuracy point, and it is the tradeoff the PR consciously accepted; the code comment should say the cap was *carried over* rather than re-measured for Check A's corpus, so a future reader does not treat the 4–10-vs-85 figure as evidence about this call site.

---

### Merge-gate state (not a code finding)

`verify.yml` has **never passed on this head**. The only Verify run for the branch (`33832021708`, 2026-09-04) was **cancelled** at 5m23s; "Detect changed scopes" reports fail and the required `npm run verify` check correctly fails closed (`##[error]detect job did not succeed: cancelled`), with every downstream leg skipped. `main`'s required-status ruleset blocks merge on that. PR title lint, PR issue link (`Closes #2858`, prose, outside any code span ✅) and CLA all pass. Needs a re-run before this can merge.

---

### House gates

- Paired regression test — ✅ real, two-sided, red-for-the-claimed-reason (verified by mutation, both polarities).
- On-box acceptance — N/A, no hardware-provable behaviour; nothing under `docs/testing/**` changed and none is owed.
- Release notes — see nit 2.
- `Closes #NN` — ✅ present, prose, outside code spans; the check-run passes.
- `cast.json` locking / lock-timeout curation seams — N/A.
- New config knob — N/A.
- Derived artifacts (`api-types.ts`, `BACKLOG.md`, brand PNGs) — N/A; #2858 is `bug`-labelled, so no `docs/BACKLOG.md` row is owed.
- Incidental findings declared in the PR body — none declared; finding 1 above was not found by the author.

**Re-review:** finding 1 is a correctness bug, so this re-triggers the gate once it is fixed and pushed.

Issue #2639, opened 2026-08-23
user settings cannot express "the user explicitly chose this" — sidecarUrl's default is also a valid choice


dudarenok-maker:
Found across three review passes on PR #2635 (#2632). Each pass found the same defect in a different disguise, which is the signal that the underlying model is wrong rather than the implementation.

## The problem

`sidecarUrl` in user settings is **non-nullable**, and its factory value (`http://localhost:9000`) is **exactly the default that per-worktree port derivation exists to replace**. So there is no way to distinguish "the user chose localhost:9000" from "nobody has ever chosen anything" by inspecting the value or the record.

Every attempt to infer it has failed in a different way:

| Attempt | How it failed |
|---|---|
| `value !== factory default` | A user who deliberately picks `localhost:9000` is treated as not having chosen — a regression against `main`. |
| key present in the persisted file | **Every** writer persists the complete merged object (`writeUserSettings`, `writeGeminiApiKey`, `writeUpgradeMeta`, `writeSetupCompletedAt`, `writeTourCompletedAt`) into one shared per-user file, so a single unrelated write makes the key present at the default. One-way door: nothing removes it. |
| key present **AND** ≠ default | Shipped in #2635 as the interim, on the repo owner's ruling. Closes the live bug with no contract change. Still mis-serves the deliberate-`localhost:9000` user — harmless today only because that user's primary checkout has `LOCAL_TTS_PORT` unset, so derivation returns 9000 anyway. |

The interim is correct for now. It is still an inference over a field that cannot carry the information.

## Decision owed

How should "the user explicitly chose this setting" be represented? Two defensible answers, and this needs a design pass rather than a fourth inference:

1. **Make `sidecarUrl` nullable.** Absent means not chosen. Honest and minimal in concept, but it is an `openapi.yaml` contract change, a settings migration, and frontend handling for a field that is currently always a string.
2. **Add a provenance field** (`sidecarUrlSource: 'default' | 'user'`). Avoids nullability and reads explicitly at the call site, but adds a second field every writer must keep correct — and "every writer persists the whole object" is precisely what broke attempt 2.

Whichever wins probably applies beyond `sidecarUrl`: any setting whose factory value is also a meaningful user choice has the same hole. `ollamaUrl` is the obvious next one — worth checking during the design pass rather than after.

## Why it is not urgent

[...]

Issue #2632, opened 2026-08-23
Sidecar port is hardcoded to :9000, so worktrees cannot run sidecars concurrently


dudarenok-maker:
Found during the #2435 wave-5 ORT acceptance run (#2621). Same structural class as the wave-4 step-5c finding, so this is the second wave it has blocked.

## What

A worktree can isolate its frontend and server ports (`VITE_PORT`, `PORT`/`VITE_API_PORT` per `scripts/wt-new.mjs`), but **not its TTS sidecar**: `spawn-sidecar.ts` / `sidecar-owner.ts` hardcode `:9000`. `LOCAL_TTS_PORT` is read only by `start.ps1`, never by the Node server.

## Why it matters

Two checkouts cannot run sidecars concurrently. Whichever starts first owns `:9000`; the second either adopts a sidecar it does not own or collides. During #2621 the acceptance agent's app bound its own VITE/API ports correctly but found another live agent lane already serving on `:9000`, and correctly refused to click Install against a session it did not own (per the standing "never disrupt another lane's process" rule). **Register row A37 could not be run** as a direct result.

This is not a defect in the code under test — it is a harness gap that blocks on-box acceptance whenever more than one lane is active, which is the normal state of this machine.

## Fix sketch

Have the Node server resolve the sidecar port the same way it resolves its own: read `LOCAL_TTS_PORT` (falling back to `9000`) in `spawn-sidecar.ts`/`sidecar-owner.ts`, and have `wt-new.mjs` assign and write it alongside the existing per-slot ports. Adoption should then key on the resolved port so a worktree never adopts another checkout's sidecar.

## Acceptance

Two checkouts running sidecars simultaneously, each serving its own, neither adopting the other's — provable on the box without a GPU.

Issue #1972, opened 2026-07-30
Splice picks target segments from segments.json but resolves their voices from the analysis cache — stale attribution renders a character in another character's voice


dudarenok-maker:
## What happens

A **designed** voice whose `.pt` is missing does **not** self-heal from its
retained clip — and the chapter then renders that character's lines **in the
narrator's voice**, while the chapter snapshot records the *intended* designed
voice. Nothing is logged.

Found running register row **A1**'s test **C-17** ⭐ during the fs-38 Wave 3
on-box acceptance run (2026-07-31, `main` @ `b5479e9c`).

Two failures, one causal chain:

1. **The self-heal did not restore the `.pt`** — 268 Invariant 8 / plan-149's
   guard, `clone-voice-resolver.ts:579-1014`.
2. **The render silently substituted another character's voice**, and
   `characterSnapshots.<id>.resolvedVoiceName` reported the assigned voice
   rather than the one actually used.

(2) is the serious half: a listener hears Maerin speaking in the Narrator's
voice, and every artifact on disk claims Maerin's own voice was used.

## Evidence

German Coalfall (`castwright__standalones__der-auftrag-von-coalfall`), ch. 2.

Fresh designed voice, created this session so the retained clip exists:

```
qwen-EFTYRmyFHpQrHr5iUfMwg.json          464 B
qwen-EFTYRmyFHpQrHr5iUfMwg.pt          39,133 B
qwen-EFTYRmyFHpQrHr5iUfMwg__master.wav 518,444 B   <- the clip self-heal needs
```

Manifest before: `instruct` set, `language: "German"`,
`designModel: Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign`, `baseModel: …0.6B-Base`.

Assigned to `maerin` → `{"updated":1,"written":["qwen","coqui"]}`, cast.json
carries `{name: qwen-EFTYRmyFHpQrHr5iUfMwg, libraryUuid: …, provenance: designed}`.

Then **only** the `.pt` was deleted, leaving `.json` and `__master.wav`, and the
character was re-recorded via splice. Run twice — the second time after a
sidecar restart, so the in-memory `_prompt_cache` was cold and a derive was
unavoidable:

- **Chapter completed** both times (`splice_complete`, 71 segments).
- **`.pt` never reappeared.** `__master.wav` and `.json` untouched at their
  original mtimes.
[...]

PR #2888, opened 2026-09-04
chore(ops): A1 on-box acceptance wave 11 — 38/60 run, fix sidecar-URL localhost bug

### LINKED ISSUE #2887: fix(server): sidecar URL derivation uses localhost instead of 127.0.0.1
## What

\`getResolvedSidecarUrl()\`'s per-worktree derivation fallback (case 3, \`server/src/workspace/user-settings.ts\`) built \`http://localhost:\${port}\` instead of \`http://127.0.0.1:\${port}\`, unlike every other sidecar-URL construction site in this codebase (\`tts/loopback-url.ts\` already uses \`127.0.0.1\` specifically for this reason).

## Why it matters

The sidecar binds \`127.0.0.1\` only (\`spawn-sidecar.ts\`'s \`DEFAULT_HOST\`). On a Windows box where \`localhost\` resolves \`::1\` first with nothing listening there, every request built from this URL — \`generation.ts\`'s pre-flight \`/health\` check, the boot-time Coqui catalog audit — can hang on the failed IPv6 attempt before any fallback, observed directly on-box as a 120s \`[tts:catalog-audit] Could not reach http://localhost:9090/speakers\` timeout.

## Fix

Changed the derived-URL fallback to \`http://127.0.0.1:\${port}\`, matching the existing convention. Updated the ~13 affected assertions across \`server/src/workspace/user-settings.test.ts\` and \`server/src/workspace/sidecar-url.test.ts\` (all case-3-derived-path assertions; explicit-override paths, which echo the given URL back verbatim, are untouched). All 101 tests across both files pass.

## Caveat

Found and fixed while diagnosing an unrelated, unresolved on-box acceptance blocker (register row A1 / C-15) — a 503 from the generation queue layer and one server-process exit under heavy background load. Verified this fix does NOT explain those symptoms (the 503 reproduced identically after the fix); it's a real, independently-verified defect, filed here on its own merits.

Found and fixed during A1 on-box acceptance testing (fs-38 Wave 3, voice cloning).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

dudarenok-maker:
## Summary

- Continues register row A1 (fs-38 Wave 3, voice cloning) acceptance testing: **34 new tests run this session, A1 now 38 of 60 run (~22 owed, was 23/37 at session start)**.
- Section C: 19 of 21 discharged (only C-15 remains, blocked by this session's own environment contention, not a product defect — see below). Section D fully discharged (D-01 through D-04). Section E: only E-06 (by-ear) and A-07/A-08/A-09/B-02 (browser+mic) remain, plus B-03/E-06 (by-ear).
- **Also fixed, found in passing:** `getResolvedSidecarUrl()`'s per-worktree `LOCAL_TTS_PORT` derivation built `http://localhost:$PORT` instead of `http://127.0.0.1:$PORT`, unlike every other sidecar-URL construction site in this codebase. On Windows dual-stack resolution, `localhost` can resolve `::1` first with nothing listening there, hanging the request. Verified NOT the cause of C-15's own blocker (a 503 reproduced identically after the fix), but a real defect on its own. Closes #2887.

All work followed the Run 4 pattern: isolated worktree, throwaway fixture books (never real user data), real Qwen/Coqui sidecar, no mocks. The one Gemini-dependent batch (C-17/C-18/E-07) ran against the primary checkout's real credentials using a throwaway book, which was deleted afterward — nothing left in the real library.

## Test plan

- [x] `npm run test:server` — full suite (538 files, 8201 tests), 0 failures, including the 5 test files updated for the localhost→127.0.0.1 fix.
- [x] `npm run check:onbox-register` — register/live-view arithmetic verified consistent.
- [x] Every new acceptance result recorded in both the detailed per-test section and the §7.1 summary table of `docs/testing/fs38-wave3-onbox-acceptance.md` (a staleness bug between the two was found and fixed mid-session).

Closes #2887


dudarenok-maker:
## PR review — pass 1 (head e1f5d179ab1a714d8f8cda6b8418bdde5aaa40be, depth medium)

Scope: all 8 changed files (`server/src/workspace/user-settings.ts` + 5 test files, `docs/testing/fs38-wave3-onbox-acceptance.md`, `docs/testing/onbox-acceptance-register.md`) · verified: ran `npx vitest run src/workspace/sidecar-url.test.ts src/workspace/user-settings.test.ts` (2 files, 101 tests, 0 failures) and `npm run check:onbox-register` (OK) myself; re-derived the "16 files referencing `localhost:9000`" sweep independently; traced every production call site of `getResolvedSidecarUrl()` before adversarially probing the fix's completeness, the honesty of the 13 rewritten assertions, and the internal consistency of the three acceptance-record surfaces.

Depth taken as `medium` per the highest single commit tier (`fix(server)` beats the `chore(ops)` commits).

---

### 🔴 Blocking — the live view, the third required acceptance surface, did not move (house gate)

**Correctness bug (process/record).** `docs/testing/onbox-acceptance-register-live-view.html` is not in this PR's file set, yet this PR discharges 15+ acceptance rows and rewrites A1's headline figures.

- `docs/testing/onbox-acceptance-register.md:871-872` (this PR) → **`38 of 60 run · ~22 still owed`**, `<!-- stat:a1-still-owed 22 -->`
- `docs/testing/onbox-acceptance-register-live-view.html:238` (untouched) → **`20 of 60 run · ~40 owed`**
- `docs/testing/onbox-acceptance-register-live-view.html:203` (untouched) → stat tile **`40` / "Still owed in A1 (of 60)"**

CLAUDE.md before-shipping step 3 is explicit that a PR discharging acceptance "must leave the acceptance state recorded across **all three surfaces**, in the same diff" — register, run sheet, **and the live view** — and that "*Recording* blocks the merge". The published artifact at the URL in the register's header will keep telling an operator 40 rows are owed in A1 while the repo says 22. Note the live view was *already* 3 rows stale against `origin/main` (20/40 vs 23/37); this PR widens the gap to 18 rows rather than creating it, but it is this PR's to close.

**The PR body's evidence for this does not hold.** Test plan says "`npm run check:onbox-register` — register/live-view arithmetic verified consistent". I ran it: it prints OK. It cannot see this drift — it compares owed **row totals, per-group counts and row IDs**, and A1 stays exactly one still-owed row before and after, so the A1 progress figures are outside what it checks (CLAUDE.md says as much: the summary-strip figures "stay a manual recompute"). A green `check:onbox-register` is not evidence that the live view agrees.

---

### 🟠 Significant — `normalizeSidecarUrlForCompare` doesn't know `127.0.0.1:9000` is the factory default, and this fix makes that the app's own canonical spelling

**Correctness bug.** Pre-existing hole, but this change materially raises its reachability, so it is in scope for this round.

`server/src/workspace/user-settings.ts:502-504` normalises only trailing slash + case before comparing against `DEFAULT_USER_SETTINGS.sidecarUrl` / `DEFAULT_LOCAL_TTS_URL`, both still `'http://localhost:9000'` (`:283`, `:492`). It does not treat `http://127.0.0.1:9000` as the same non-choice.

Concrete failure, on a worktree set up per CLAUDE.md's own recipe:
1. `server/.env` has `LOCAL_TTS_PORT=9110` (slot 1).
2. The operator reads the resolved URL the app now reports — `SidecarHealthResult.url` (`server/src/routes/sidecar-health.ts:367`, served by `/api/sidecar/health` and the diagnostics rows), or the resolver's own `console.warn` lines — and it now reads **`http://127.0.0.1:9000`** on a default box.
3. They paste that into Account → sidecar URL.
4. `getResolvedSidecarUrl()` takes case 1 (`normalize('http://127.0.0.1:9000') !== normalize('http://localhost:9000')` → true), returns `http://127.0.0.1:9000`.
5. The supervisor still spawns on `resolveSidecarPort()` = **9110**. Server spawns on 9110, talks to 9000 — the exact spawn-here/talk-there split #2632/N1 and the three rounds of N21 hardening exist to prevent.

Before this change the same paste-back produced `http://localhost:9110` or `http://localhost:9000`, and the `localhost:9000` case was *caught* by the normaliser. The comment at `:494-501` states the guard's own contract — "a spelling that is the factory default in every sense but one" must not beat port derivation — and `127.0.0.1:9000` is now precisely such a spelling, emitted by the product itself. This is defect shape #2 from the reviewer brief (a guard that enumerates spellings loses one per round); the fix hands it a new spelling to lose. Cheapest correct closure: normalise the loopback host class in `normalizeSidecarUrlForCompare`, not just slash/case.

### 🟠 Significant — release-notes pair missing, with no explicit not-applicable

**Correctness bug (house gate).** Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the PR body never declares the skip. The `fix(server)` commit is operator-visible: on a dual-stack Windows box it converts a `POST /generation` 503 and a 120 s `[tts:catalog-audit]` stall into a working call — the commit message and the register's Run 8 note both say so. `RELEASE_NOTES.md`'s open v1.15.0 section already carries several entries of exactly this shape ("GPU speech synthesis works correctly again on NVIDIA boxes", "the third reason your GPU could sit idle…"). CLAUDE.md step 5 permits a skip only for "no shippable delta", and requires it be said out loud. Say it, or land the pair.

### 🟠 Significant — §7.1's table contradicts the detailed Result: sections at C-02 and C-03

**Correctness bug (record).** This PR contains commit `448dc2e5` whose entire stated job is "sync A1 run sheet's §7.1 summary table with detailed results". It walked the Section C/D table and left two rows wrong:

- `docs/testing/fs38-wave3-onbox-acceptance.md:3153` — `| C-02 | … | **B** | needs a full-chapter render — blocked by the side-11 leak (see §7.2 BLOCKER-1) |`, but `:1438` reads `**Result:** ☒ P … Run 2026-08-31 … chapter_failed in 748 ms … All criteria met.` The register itself says `C-02 — DISCHARGED, Run 4` (`onbox-acceptance-register.md:1232`). Two documents in the same PR disagree about whether C-02 is blocked.
- `docs/testing/fs38-wave3-onbox-acceptance.md:3154` — `| C-03 | Broken voice not in this chapter doesn't fail it | | |` (blank = not reached), but `:1471` reads `**Result:** ☒ P … Both halves pass.`

Both are pre-existing staleness from wave 10, but they are precisely what the sync commit claimed to eliminate, and the PR body repeats the claim ("Every new acceptance result recorded in both the detailed per-test section and the §7.1 summary table"). They also make the register's own `Section C — 16 of 16 attempted, 15 discharged` bullet unverifiable: it enumerates 13 full + C-13 partial + C-15 = 15 IDs against a 21-row section, and the PR body says something different again ("Section C: 19 of 21 discharged").

### 🟠 Significant — the register still lists E-07 as owed after this PR discharged it

**Correctness bug (record).** `docs/testing/onbox-acceptance-register.md:1170-1171` (unchanged by this PR) reads:

> **Section E — 6 of 9 now run (2026-07-31/08-01, across runs 2 and 3); E-03, E-06 and E-07 still owed, but no longer blocked.**

This PR's own Run 7 paragraph (`:1128-1131`) records **E-07 ⭐** as discharged, and `fs38-wave3-onbox-acceptance.md:3195` marks it **P**, matching the detailed `☒ P` at `:2970`. So the still-owed bullet double-counts a discharged row and the section count should read 7 of 9. This is the reviewer brief's shape #9 — a doc sentence this change made false — and under CLAUDE.md it is a chore the work made owed, not a later cleanup.

---

### 🟡 Minor

1. **`server/src/workspace/sidecar-url.test.ts:45-49` is now a test that cannot fail.** With `LOCAL_TTS_URL='http://127.0.0.1:9000'` and `LOCAL_TTS_PORT` unset, the pass-through branch returns `http://127.0.0.1:9000` — and so does the case-3 fallback (`resolveSidecarPort()` → 9000, per `sidecar-owner.test.ts:37`). Delete the entire `LOCAL_TTS_URL` branch from the resolver and this test still passes. Before the fix the two branches produced different strings, so it was a real test. That the sibling file's `:48` assertion passes proves `LOCAL_TTS_PORT` really is unset in that run, so this is the live behaviour, not a hypothetical. The invariant survives at `user-settings.test.ts:923-931` (`192.168.1.20:9000`), so this is redundancy lost rather than coverage lost — a one-line fix is to use a port the derivation cannot produce.

2. **Three factual errors in the new comment** (`server/src/workspace/user-settings.ts:563-570`):
   - "the sidecar binds 127.0.0.1 only (spawn-sidecar.ts's DEFAULT_HOST)" — `spawn-sidecar.ts:113`'s `DEFAULT_HOST` is the host **the server probes and kills on**, not the bind host. The actual bind is `server/tts-sidecar/start.ps1:99` / `start.sh:44`: `LOCAL_TTS_HOST ?? 127.0.0.1`.
   - "**only**" is false — `LOCAL_TTS_HOST` overrides it (and `sidecar-owner.ts:179` already says so).
   - "loopback-url.ts already avoids this for the same reason; **this is the other** sidecar-URL construction site" — `loopback-url.ts:12-15` builds the **server's own** base URL for sidecar→server callbacks, not a sidecar URL; and it is not the only other one (see 3).

3. **The class is not closed: a third `localhost` sidecar-URL site remains.** `scripts/remint-anchored-variants.mjs:188` defaults `SIDECAR_URL` to `'http://localhost:9000'`. Run `node scripts/remint-anchored-variants.mjs --apply` with no `SIDECAR_URL`/`LOCAL_TTS_URL` on the very box this bug was found on and every sidecar call hangs on `::1` — the identical symptom. (`server/tts-sidecar/README.md:514,520`'s `Invoke-RestMethod http://localhost:9000/health` operator recipes are the same class.) Defect shape #10.

4. **D-04's QA-repair half is attributed to three different runs in three places in this same PR**: `fs38-wave3-onbox-acceptance.md:3181` says "**Run 5**. … QA-repair half: tightened `qa.seg.minRatio`…"; the detailed section at `:2720` says "**QA-repair half — completed in Run 6**"; `onbox-acceptance-register.md:1104-1106` says "**Run 7** … Also discharges D-04's QA-repair half". Provenance is the whole point of the register; pick one.

5. **§7.1's `#### Totals` tables stop at Run 2** (`fs38-wave3-onbox-acceptance.md:3224-3264`: 20 P / 0 F / 8 B / 31 not reached) while the per-section tables directly above them now carry ~17 more results and the register claims 38 of 60 run. A reader of §7.1 alone cannot reconcile it. Either add an "after Run 8" row or say the earlier tables are frozen history.

6. **109 untracked scratch files sit in the worktree root** (`analysis*.sse`, `c0*-*.sse`, `primary-*.json`, `state[1-6].json`, …), byproducts of the acceptance runs written to the checkout root rather than a scratch dir. Not in the PR, but they make `git status --porcelain` unreadable — the exact signal this gate's own tree check reads — and one `git add -A` commits ~109 SSE dumps. (For the record: this identical set was present before my pass and is unchanged by it; `HEAD` is unchanged and no tracked file was modified.)

---

### ✅ What is solid

- **The fix is correct and reachable at the default configuration.** With no `LOCAL_TTS_URL` and untouched settings, case 3 is the live path, so this is not shape #7 (a fix that runs zero times as shipped). The premise checks out independently: the sidecar really does bind `127.0.0.1` by default (`start.ps1:99`, `start.sh:44`), and `sidecar-owner.ts:176-190` already documents the same IPv4/IPv6 asymmetry for its own probe.
- **The test updates are honest.** All 13 rewritten assertions are exact string equalities on the derived URL (`toBe('http://127.0.0.1:9110')` etc.). Nothing was loosened — no `toContain`, no regex, no assertion deleted, no `.skip` added; 13 exact literals in, 13 exact literals out. Each was red before the fix for exactly the claimed reason (the host substring), since the resolver has a single host literal.
- **Completeness verified independently, not taken on the commit message's word.** I re-ran the `localhost:9000` sweep. Every remaining occurrence is legitimately unchanged: the *stored setting's* factory default (`user-settings.ts:283`/`:492`, `routes/user-settings.test.ts:83,135`, `openapi.yaml:4609`, `.env.example`), an injected literal (`sidecar.test.ts`, `voice-sample-cloned-language.test.ts`), a self-consistent mock of the resolver (`ensure-sidecar-loaded.test.ts:7,335`, `ensure-sidecar-vram.test.ts:8`), or a fixture/error string (`diagnostics.test.ts`, `models-inventory.test.ts`, `cast-design.test.ts`, `voice-sample.test.ts`, `env-cleanup.test.ts`, `aggregate.test.ts`). **No real-derivation assertion was missed**, and none of the 50+ production call sites of `getResolvedSidecarUrl()` compares its output against a default — `git grep` for `DEFAULT_USER_SETTINGS.sidecarUrl` / `=== getResolvedSidecarUrl` returns only the resolver's own line 524 — so the host change cannot flip an "is this customised?" branch anywhere.
- **Issue linkage is real, not assumed.** `Closes #2887` appears plain (outside any code span); #2887 exists, is open, and carries the standalone `bug` label — the correct shape for bug-shaped work per CONTRIBUTING.md.
- The register's headline arithmetic is self-consistent (38 + 22 = 60), and every one of the 18 newly-recorded results has a filled `☒` in its detailed section as well as a §7.1 row.

### Verdict

**Do not merge yet.** One blocking house-gate failure (the live view, the third acceptance surface, is 18 rows behind what this PR asserts), one real correctness item in the fix's own neighbourhood (🟠 #1 — the default-spelling guard now misses the spelling the product emits), and a missing release-notes pair with no declared skip. The three record-consistency items (§7.1 vs detailed at C-02/C-03, E-07 still listed as owed, D-04's three run attributions) are chores this work made owed and belong in the same round, not a later pass.

The code fix itself is sound, complete on the server side, and its test updates are honest — none of the blocking items is a reason to doubt it.

dudarenok-maker:
## PR review — pass 2 (head `780b2f77`, depth medium)

Re-review of [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2888#issuecomment-5540341223). Deltas only — solid items are not re-listed.

Verified: `user-settings.test.ts` + `sidecar-url.test.ts` **103/103 green**; `npm run check:onbox-register` **OK**; comparator behaviour probed directly against 17 host spellings; §7.1 Totals arithmetic independently recomputed from the section tables.

---

|---|---|---|
| 🔴 | Live view stale since before Run 4 | **VERIFIED RESOLVED (as scoped)** — see note below |
| 🟠 1 | `normalizeSidecarUrlForCompare` blind to `127.0.0.1` | **VERIFIED RESOLVED** |
| 🟠 2 | Release notes missing | **VERIFIED RESOLVED** |
| 🟠 3 | §7.1 stale C-02/C-03 (+E-03) rows | **VERIFIED RESOLVED** |
| 🟠 4 | Register listed E-07 as owed | **RESOLVED in `register.md`; the same claim survives in the live view — NEW-1** |
| 🟡 1 | `sidecar-url.test.ts` fallback couldn't fail | **VERIFIED RESOLVED; its sibling was disarmed — NEW-3** |
| 🟡 2 | Three factual errors in the `getResolvedSidecarUrl()` comment | **VERIFIED RESOLVED** |
| 🟡 3 | Third localhost site in `remint-anchored-variants.mjs` | **VERIFIED RESOLVED** |
| 🟡 4 | D-04 run-number inconsistency | **VERIFIED RESOLVED** |
| 🟡 5 | §7.1 Totals frozen at Run 2 | **VERIFIED RESOLVED** |
| 🟡 6 | 109 untracked scratch dumps | **VERIFIED RESOLVED** — `git status --porcelain` empty |

**🟠 1 — the correctness fix, checked hard.** The regex `^(https?:\/\/)127\.0\.0\.1(:|\/|$)/` is correctly guarded: the `(:|\/|$)` tail means `http://127.0.0.10:9000` and `http://127.0.0.1.evil.com:9000` do **not** fold, and anchoring after the scheme means `http://user@127.0.0.1:9000` and a query-embedded `…?x=http://127.0.0.1:9000` don't either. `https://127.0.0.1:9000` and `http://127.0.0.1:9110` correctly stay explicit. Fold order (strip slash → lowercase → fold) is sound, and `DEFAULT_USER_SETTINGS.sidecarUrl` (`user-settings.ts:283`) and `DEFAULT_LOCAL_TTS_URL` (`:492`) are both `http://localhost:9000`, so both comparison sites at `:533` and `:554` normalise consistently. Comparison-only: `raw` still feeds `isPrivateHostUrl` and the return, so the srv-21 guard is untouched. **Both new tests can genuinely fail** — they set `LOCAL_TTS_PORT=9110`, so without the fold they'd return `http://127.0.0.1:9000` against an asserted `:9110`. Not a trivially-passing assertion.

**🟡 2 — the replacement comment's own facts check out**, which matters since the finding was about false claims in this exact comment: `LOCAL_TTS_HOST` is real (`start.ps1:99`, `start.sh:44`, both defaulting to `127.0.0.1`), and `spawn-sidecar.ts:113`'s `DEFAULT_HOST = '127.0.0.1'` is correctly re-characterised as what this process probes against rather than a bind directive.

**🟡 5 — Totals independently recomputed, correct.** All six columns and all five section rows sum: P `9+3+19+3+7=41`, F `2`, B `8`, N/A `1`, not-reached `1+6+0+0+1=8`, total 60. Section-level counts match my own recount of the §7.1 rows (C: 18 clean P + C-13 = 19 P, C-05 F, C-15 B). The register's `38 of 60 · ~22 owed` also reconciles: 8 blocked + 8 not-reached + the 6 weaker/partial passes (C-04, C-12, C-13, C-20, D-03, E-01) = 22. Checked, not a discrepancy.

---

### 🟠 Significant (NEW-1) — the live view still lists E-07 and E-03 as owed; the page was published in that state

`docs/testing/onbox-acceptance-register-live-view.html:270`:

> **6 of its 9 tests now run (runs 2 and 3); E-03, E-06 and E-07 remain owed but are no longer blocked**

This is the *exact sentence* `bdf2d81b` fixed in the twin file — `onbox-acceptance-register.md:1170-1171` now correctly reads "**8 of 9 now run (runs 2, 3, 4 and 7); only E-06 (by-ear) still owed.** E-03 discharged Run 4, E-07 discharged Run 7." The HTML copy was not updated with it.

Concretely, the published page now contradicts itself within 6 lines:
- `:264` — `<s><b>E-01, E-02, E-07</b></s> — all passed`
- `:266` — "**Only C-15 and E-06 remain starred and owed**"
- `:270` — "E-03, E-06 and **E-07** remain owed"

Pass 1 graded "register still lists E-07 as owed" 🟠; the identical claim on the *published* surface is the same finding, one file over. This is one-instance-fixed / class-left-armed: the `register.md` edit had an HTML twin and only one moved.

**`npm run check:onbox-register` returning OK is not evidence against this** — it cross-checks the owed total, per-group counts and row IDs, not prose, exactly as CLAUDE.md says ("cannot tell you a row is missing, only that the ones already there don't add up"). It passes green with this sentence in place; I confirmed that.

Note this one needs a **republish**, not just an edit — the stale text is live at the artifact URL recorded in the register header.

### 🟠 Significant (NEW-2) — E-01's discharge is credited to Run 2, whose E-01 result was retracted

`onbox-acceptance-register.md:1298-1300` (rewritten by `bdf2d81b`) and `live-view.html:264` both attribute E-01's discharge to **Run 2**:

> are all now discharged** (C-10 2026-07-29; C-01/C-08/C-17/C-12 Run 5/7; E-01/**E-02 Run 2**; E-07 Run 7)

Run 2's E-01 entry was **retracted**. Three sources in this same PR say so:
- run sheet §7.1 E-01: "**Run 3 (2026-08-01) — first genuine exercise; supersedes run 2's retracted entry below**", and the Run-2 blockquote: "**⚠️ The identity half is RETRACTED**" (13 of 21 segments rendered in other characters' voices, #1972);
- `register.md:1172` — "**Run 3 (2026-08-01)** added E-01's first genuine exercise";
- `live-view.html:244` — "**E-01**: identity half withdrawn (13 of 21 segments divergent)", and `:248` "**Run 3** — **E-01**'s first genuine exercise".

E-02 *is* legitimately Run 2; E-01 is Run 3. Citing the retracted run as the discharge basis is the specific failure mode the register's "record what was observed, by whom, and when" rule exists to prevent.

**Second half of the same finding, same paragraph block:** `register.md:1308` still opens

> **E-01 was attempted and is blocked, not failed.** … To attempt Section E, first flip the target character's engine to coqui …

— present-tense, unqualified, forward-looking, and directly contradicting "E-01 … discharged" **ten lines above it**. Before `bdf2d81b` the starred paragraph listed E-01 as still owed, so the two agreed; the fix for pass-1's 🟠 4 created this contradiction. Compare the run sheet, which marks its superseded block "**Run 2 (superseded, kept for history)**" — this one carries no such marker.

One fix agent covers both halves (same paragraph block, both files).

### 🟡 Minor (NEW-3) — the 127.0.0.1 fold silently disarmed `sidecar-url.test.ts`'s sibling test

`server/src/workspace/sidecar-url.test.ts:48-52`, in the `getResolvedSidecarUrl — srv-21 fallback` block:

```ts
it('passes through a private-host URL', () => {
  _resetUserSettingsCache();
  process.env.LOCAL_TTS_URL = 'http://127.0.0.1:9000';
  expect(getResolvedSidecarUrl()).toBe('http://127.0.0.1:9000');
});
```

After the fold, `http://127.0.0.1:9000` normalises to `http://localhost:9000` = the default, so the `:552` branch is **false**, `isPrivateHostUrl` is never called, and the value comes from `:583`'s port derivation. The test no longer exercises the pass-through path it is named for; it now asserts the derivation default.

Probed directly against the real module:

```
LOCAL_TTS_URL='http://127.0.0.1:9000'  (the actual test)       -> http://127.0.0.1:9000
LOCAL_TTS_URL unset entirely                                   -> http://127.0.0.1:9000
LOCAL_TTS_URL='http://localhost:9000' (factory default)        -> http://127.0.0.1:9000
LOCAL_TTS_URL='http://127.0.0.1:9000' + LOCAL_TTS_PORT=9010    -> http://127.0.0.1:9010   <-- would FAIL
LOCAL_TTS_URL='http://192.168.1.20:9000' (real pass-through)   -> http://192.168.1.20:9000
```

Rows 1–3 are identical, so the assertion is fully insensitive to the variable it sets. It also now depends on `LOCAL_TTS_PORT` being unset — and this worktree's own `server/.env` sets `LOCAL_TTS_PORT=9090` per CLAUDE.md's mandated worktree setup (it passes only because vitest doesn't load that file). A private-host literal that is *not* the folded default (`http://192.168.1.20:9000`, as the sibling `user-settings.test.ts:923-930` already uses) restores the coverage.

**Not 🟠** because the pass-through branch is genuinely covered at `user-settings.test.ts:923-930`, so this is a vacuous test rather than a coverage hole. Worth flagging anyway: the same commit fixed the *other* test in this same file for this same defect shape and disarmed this one.

### 🟡 Minor (NEW-4) — the new release-notes bullet repeats the factual error this commit fixed

`docs/release-notes-next.md`, the new #2887/PR #2888 bullet:

> The sidecar binds `127.0.0.1` **only**, and on a box where "localhost" resolves `::1` first…

That is the exact claim pass-1's 🟡 2 flagged and this same commit corrected in the code comment, which now correctly reads "binds `127.0.0.1` **by default** … overridable via `LOCAL_TTS_HOST`". Same commit, two places, one corrected and one not.

Same bullet, smaller: "It was **the one** sidecar-URL construction site in the codebase still building `http://localhost:$PORT`" is contradicted by its own closing sentence, which fixes a second one (`remint-anchored-variants.mjs`).

### 🟡 Minor (NEW-5) — the #2639 comment now understates the mis-served set

`server/src/workspace/user-settings.ts:523-525`:

> Value-difference alone was the old sentinel that failed when user explicitly picked :9000, and still does: **that one case** (a user who deliberately picks `http://localhost:9000`) remains mis-served on this head, tracked at #2639.

The fold widens that set. Before this commit, `sidecarUrl = http://127.0.0.1:9000` was honoured as an explicit choice; now it folds to the default and loses to port derivation — so with `LOCAL_TTS_PORT=9010`, a user who deliberately pins `http://127.0.0.1:9000` gets `http://127.0.0.1:9010`. The widening is defensible (that spelling is what the app itself now reports back, so it reads as confirmation, not choice) — but it is no longer "that one case", and the newly-affected spelling is the one users are now *most* likely to type. The comment should name both spellings; #2639 arguably wants the same note.

---

### Verdict

**Not clear to merge.** The `127.0.0.1` comparator fix — what re-triggered this pass — is **correct, well-guarded and genuinely tested**; I tried hard to break it and could not. The run-sheet and release-notes findings all landed.

The register/live-view bookkeeping is where this round regressed. Two 🟠 findings, both in the acceptance-register surface CLAUDE.md makes a merge gate, and both *introduced or left by the commits that fixed pass 1's register findings*: E-07 still reads as owed on the published page, and E-01's discharge is credited to a retracted run while a paragraph ten lines away still calls it blocked. NEW-1 needs a republish, not just an edit.

Two 🟠 correctness findings → re-review round 2 (the loop cap) after fixes land.

Tree check: `HEAD` `780b2f77` and `git status --porcelain` empty before and after this pass — no tracked file modified.

dudarenok-maker:
## PR review — pass 3 (head `fb446e92`, depth medium)

**This is the loop cap** — initial pass + 2 re-review rounds, per [`pr-review-gate`](https://github.com/dudarenok-maker/Castwright/blob/main/.claude/skills/pr-review-gate/SKILL.md#re-review-trigger-and-loop-cap). Whatever remains open below is handed to the repo owner rather than looped again.

Re-review of [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2888#issuecomment-5540341223) and [pass 2](https://github.com/dudarenok-maker/Castwright/pull/2888#issuecomment-5544727902). Deltas only — solid items from earlier passes are not re-listed.

Verified this round: `sidecar-url.test.ts` + `user-settings.test.ts` **103/103 green**; `npm run check:onbox-register` **OK**; the "Historical (Run 1…)" provenance label traced to its originating commit with `git log -S`; **and the page currently live at the canonical artifact URL fetched and read directly**, which is what produced this round's one blocking item.

---

| Pass | # | Finding | Disposition |
|---|---|---|---|
| 1 | 🔴 | Live view stale / not in the file set | file **RESOLVED**; **published page has fallen behind again — see 🔴 below** |
| 1 | 🟠 1 | `normalizeSidecarUrlForCompare` blind to `127.0.0.1` | **VERIFIED RESOLVED** (re-confirmed green) |
| 1 | 🟠 2 | Release-notes pair missing | **VERIFIED RESOLVED** |
| 1 | 🟠 3 | §7.1 stale C-02/C-03/E-03 rows | **VERIFIED RESOLVED** |
| 1 | 🟠 4 | Register listed E-07 as owed | **VERIFIED RESOLVED** in both tracked files |
| 1 | 🟡 1–6 | test-can't-fail, comment errors, 3rd localhost site, D-04 runs, §7.1 Totals, scratch files | **VERIFIED RESOLVED** |
| 2 | 🟠 NEW-1 | live-view still listed E-03/E-07 as owed | **file RESOLVED; NOT PUBLISHED — see 🔴** |
| 2 | 🟠 NEW-2 | E-01 credited to the retracted Run 2 | **file RESOLVED (both halves); NOT PUBLISHED — see 🔴** |
| 2 | 🟡 NEW-3 | pass-through test disarmed by the fold | **VERIFIED RESOLVED** |
| 2 | 🟡 NEW-4 | release-note repeated the "binds 127.0.0.1 **only**" error | **VERIFIED RESOLVED** (second half withdrawn — see below) |
| 2 | 🟡 NEW-5 | #2639 comment understates the mis-served set | **NOT ADDRESSED** — still open, verbatim |

**NEW-2, checked rather than accepted.** Both files now read `E-01 Run 3 — Run 2's E-01 result was retracted, #1972; E-02 Run 2`. The second half's fix invents a provenance ("Historical (**Run 1**, before Section E was unblocked)"), so I checked it: `git log -S "A Coqui splice reported" -- docs/testing/onbox-acceptance-register.md` returns exactly one commit, `8cbeccaf docs(docs): record the fs-38 Wave 3 on-box acceptance run of **2026-07-29**` — Run 1's own date. The label is accurate, not a plausible guess. `grep` confirms no "blocked, not failed" claim about E-01 survives in either file.

**NEW-3, checked hard.** `http://192.168.1.50:9123` genuinely discriminates: the case-3 fallback can only ever return `http://127.0.0.1:<port>`, so no derivation produces that host, and `isPrivateHostUrl` accepting the `192.168/16` range is itself pinned at `sidecar-url.test.ts:14`. Delete the `LOCAL_TTS_URL` branch and this test now fails. Real coverage restored, not a relabelled vacuous assertion.

**NEW-4's second half — withdrawn, my error.** Pass 2 called "It was **the one** sidecar-URL construction site in the codebase still building `http://localhost:$PORT`" self-contradicted by the same bullet's `remint-anchored-variants.mjs` sentence. On re-reading the script, `scripts/remint-anchored-variants.mjs:186-190` never built `$PORT` — its default was a **static literal** `http://localhost:9000`. The `$PORT` qualifier makes the sentence accurate as written. Nothing owed; closing it.

---

### 🔴 Blocking — the published live view still carries **both** of pass 2's 🟠 findings verbatim, and the token that would order the corrective republish has already been spent

**Correctness bug (house gate).** I fetched the page currently live at the canonical URL in the register's header (`https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a`) and read it. Three facts, all from that fetch:

1. Its publish token is **`data-published-as="8" data-publish-id="a2df1ee4a776"`** — byte-identical to the token this branch committed in `d29b5a79` ("stamp the live-view publish token before republishing"). **Publish #8 has already happened**, and it was published from `d29b5a79`'s content.
2. The live page still reads, verbatim:
   > **6 of its 9 tests now run (runs 2 and 3); E-03, E-06 and E-07 remain owed but are no longer blocked**

   — that is **NEW-1**, unchanged, on the published surface.
3. The live page still reads, verbatim:
   > `<s><b>E-01, E-02, E-07</b></s>` — all passed (**E-01**/E-02 **run 2**; E-07 Run 7)

   — that is **NEW-2**, unchanged, on the published surface. (Corroborating that the live page is the `d29b5a79` build and not something older: its A1 summary already says `38 of 60 run · ~22 owed · 3 retracted` and its strip tile reads `22`, both of which only `bdf2d81b` introduced.)

`fb446e92` fixed both sentences in the tracked `.html` **and did not re-stamp**. So `HEAD`'s file carries the same `8` / `a2df1ee4a776` pair that is already live, over different content.

That produces two distinct failures, and the second is the one that outlives this PR:

- **CLAUDE.md before-shipping step 3's third surface is behind again**, on exactly the two sentences pass 2 raised — "the register, the per-feature run sheet, **and** the live view … all move in the shipping PR", and "*Recording* blocks the merge".
- **Republishing as-is re-uses a consumed counter + nonce.** That is precisely what `scripts/stamp-publish-token.mjs`'s own header forbids — *"Bumping the counter without minting a new id is the same failure with an extra step"* — and what the register's step 0 requires: *"If your change touched the live view, stamp it and commit the stamp… an **uncommitted** stamp cannot be found in git history, which is where its freshness is verified."* `fb446e92` touched the live view after the stamp was spent, so the ordering signal for publish #9 does not exist yet. The stamper exists so a stale identity cannot fail in the **green** direction; publishing twice under `8`/`a2df1ee4a776` is that stale identity.

**`npm run check:onbox-register` returning OK is not evidence against this, and I confirmed that empirically** — I ran it on this head and it prints `OK — … agree.` It compares the two **tracked** files (owed total, per-group counts, row IDs); no row was added or removed here, so a wording-only edit changes nothing it looks at. The register's own edge list says exactly this ("**A wording-only edit does not fail**… The live view mirrors that prose in its own row bodies and will silently fall behind"), and adds that the published page is invisible to the no-flag run. A `--against-published` run would also come back green for the same reason: same rows on both sides.

Discharge is `npm run stamp:publish-token` → commit → fetch + save the live page → `npm run check:onbox-register -- --against-published <saved-file>` → republish **that** `.html` with the canonical URL passed as `url`. It touches no code and no reviewable logic, so it does not need another review round — it needs doing before merge.

---

### 🟡 Minor (NEW-5 from pass 2) — still open, unchanged

`server/src/workspace/user-settings.ts:523-525` still reads:

> Value-difference alone was the old sentinel that failed when user explicitly picked :9000, and still does: **that one case** (a user who deliberately picks `http://localhost:9000`) remains mis-served on this head, tracked at #2639.

The fold at `:512` widened that set to two spellings: with `LOCAL_TTS_PORT=9010`, a user who deliberately pins `http://127.0.0.1:9000` now also gets `http://127.0.0.1:9010`. The widening is defensible — that spelling is what the app itself reports back — but "that one case" is now false, and the newly-affected spelling is the one a user is most likely to paste. Note this was raised in pass 2 and is absent from `fb446e92`'s message, which enumerates NEW-1…NEW-4 only; per [findings-triage](https://github.com/dudarenok-maker/Castwright/blob/main/.claude/skills/pr-review-gate/references/findings-triage.md), 🟡 findings are fixed in-round even though they don't re-trigger the loop.

### 🟡 Minor (new) — Section C's headline count disagrees with the list printed beside it, in both twin files

`docs/testing/onbox-acceptance-register.md:1251-1254`:

> **Section C — 16 of 16 attempted, 15 discharged as of Run 8 (C-01 ⭐, C-04, C-06, C-07, C-08, C-09, C-12, C-14, C-16, C-17 ⭐, C-18, C-20, C-21 full; C-13 partial …).** **C-15** is the one row still owed

The parenthetical enumerates **13 full + C-13 partial = 14** IDs, not 15; and 14 discharged + C-15 attempted = **15** attempted, not 16. `onbox-acceptance-register-live-view.html:240` carries the same off-by-one with the identical 14-ID list (`Section C 15 of 16 discharged`). Both files agree with each other and disagree with their own enumeration, which is why `check:onbox-register` is green over it.

This is pass-1 🟠 3's tail: that finding noted the bullet "enumerates 13 full + C-13 partial + C-15 = 15 IDs" and cannot be reconciled with the PR body's "Section C: 19 of 21 discharged". For the record, **the PR body's figure is the correct one for the whole section** (all-runs discharged = C-01…C-04, C-06…C-14, C-16…C-21 minus C-05 F and C-15 blocked = 19 of 21); the register's numbers are wave-scoped and land one short of their own list under every scoping I could construct (Runs 4–8 → 17 attempted / 16 discharged; Runs 5–8 → 15 / 14). Either fix the two numbers to `15 of 15 attempted, 14 discharged`, or name the missing 15th ID.

### 🟡 Minor (new) — "only E-06 (by-ear) still owed" is contradicted 30 lines later in the same bullet

`docs/testing/onbox-acceptance-register.md:1170` (under the heading **"Still owed (~22), and why:"**):

> **Section E — 8 of 9 now run (runs 2, 3, 4 and 7); only E-06 (by-ear) still owed.**

`:1198-1201`, inside that same bullet:

> **E-04 specifically is no longer blocked on a fix** … **so what remains of its debt is a re-run of the reproduction** (46-char control, 245-char Russian line) on real Coqui weights, not an outstanding bug.

So the owed-work enumeration says one row is owed while its own body names a second row's owed re-run. Two readings are defensible — "owed" as *never run* (E-04 was run; it is an `F`) versus "owed" as *carries remaining work* — and the `~22` arithmetic is self-consistent under the first (35 clean P + 2 F + 1 N/A = 38 run; 8 B + 8 not-reached + 6 weaker = 22). That makes this a wording fix, not a recount: say "only E-06 has never been run; E-04's `F` additionally owes a re-run", or add E-04 to the list. Flagging because the "only E-06" phrasing was **introduced by `bdf2d81b`**, this PR's fix for pass-1 🟠 4, and pass 2 verified that finding without reaching this sentence's tail.

---

### Verdict

**Not clear to merge as it stands — but nothing here is a code defect.** The `fix(server)` half of this PR has now survived three adversarial passes: the `127.0.0.1` derivation, the `normalizeSidecarUrlForCompare` fold and its regex guards, the 13 rewritten assertions, and the two `sidecar-url.test.ts` cases are all correct, complete, honestly tested and reachable at the default configuration. I tried to break them again and could not.

Everything still open is on the acceptance-register/bookkeeping surface, and it has now regressed in the same place for three rounds running — each round's fix to a `register.md` sentence left, or created, a defect one file or one paragraph over.

Handed to the repo owner at the cap:

1. **🔴 Stamp + republish the live view.** Blocking under CLAUDE.md step 3, and mechanical: `npm run stamp:publish-token` → commit → fetch/save the live page → `check:onbox-register -- --against-published` → republish with the canonical `url`. Without it, the artifact keeps telling an operator that E-03 and E-07 are owed (they are discharged) and that E-01 passed in the run whose result was retracted — the two claims this PR's last two rounds were spent correcting.
2. **🟡 ×3** — NEW-5 (carried over unfixed from pass 2), the Section C off-by-one, the "only E-06" tail. All cleanup-class; none re-triggers the loop, all are fixed-in-round work under findings-triage.

Tree check: `HEAD` `fb446e92a7dc8ca74c4c1c253728ac0e44b72de3` and `git status --porcelain` empty both before and after this pass — no tracked file modified.

dudarenok-maker:
Pass 3's one remaining blocker (the live-view republish) is done: commit `c4a43d3e` re-stamped the publish token (counter 8→9, fresh nonce), and the page has now actually been republished to the canonical URL — verified via `npm run check:onbox-register -- --against-published` against the live page immediately before publishing.

The 3 minor findings from pass 3 are also fixed in `c4a43d3e`: the widened `#2639` comment, the Section C count phrasing (matches its own enumeration now instead of asserting an unverified total), and the Section E "only E-06 owed" line (now also names E-04's owed reproduction re-run).

This is round 3, the review-loop cap — no code-correctness issue remains open, only documentation/register bookkeeping, all of which is now fixed and republished. Leaving final merge judgment to the repo owner per the loop-cap rule.

PR #2837, opened 2026-08-31
chore(ops): per-row content hashing for onbox-register --against-published

dudarenok-maker:
## Summary
- Extends `check-onbox-register.mjs --against-published` to hash each row's body content (keyed by row ID) and compare the tracked `live-view.html` against the published snapshot, so a row whose ID/count stay stable but whose body text regresses (the PR #2578 round-18 A41 defect) is now caught mechanically instead of only by manual byte-diffing.
- Deliberately diffs the live-view HTML twin, not the markdown register, since the register's prose is not required to be word-for-word identical to its HTML twin even in a healthy state.
- Documents the new capability in `docs/testing/onbox-acceptance-register.md`'s "Live view" section.

## Test plan
- [x] `scripts/tests/check-onbox-register.test.mjs` — new cases covering per-row content-hash drift detection
- [x] CI green (lint/typecheck, frontend+server tests, e2e, build, onbox-register consistency check)

## Also fixed, found in passing
- A fix commit on an earlier review round accidentally deleted one line from an unrelated file (`scripts/check-register-citations.mjs`'s `SELF_REFERENTIAL_PATHS` set, likely an unreverted mutation-test probe), regressing 6 tests in `check-register-citations.test.mjs`; restored it verbatim. `check-register-citations.mjs` is otherwise untouched by this PR's final diff.
- Added three CLI-level (runCli) test cases for the 3-way warning and extraction-error banner wiring, with mutation verification confirming each fix is essential; prior round had only unit-level coverage.
- Fixed two stale docs (in-code comment and operator-facing markdown) that still claimed the 3-way content-disagreement case blocks/fails-closed; it now reports as an advisory warning only.
- Added `fetch-depth: 0` to checkout steps in `release.yml` (verify + cross-os-verify jobs) and `cross-os.yml` (cross-os-verify job) so the 25-commit replay test has full git history at release-gate CI, not just a shallow clone.
- Added reconciliation note to `2026-08-27-onbox-register-stable-row-ids-design.md` explaining how PR #2837 resolves the spec's earlier argument against content comparison via an advisory-warning approach (visible, non-blocking).

Closes #2599

PR #2945, opened 2026-09-05
feat(ops): generate the on-box register's derived live-view figures

### LINKED ISSUE #2362: ops — decide what `check:onbox-register` compares beyond counts and row IDs

### LINKED ISSUE #2721: ops-64 — rehome the wrongId-widening deferral that #2629's closure orphans

dudarenok-maker:
## Summary

Implements `docs/superpowers/plans/2026-08-29-onbox-register-generated-surfaces.md`
Tasks 4-8 ("PR 3" of the on-box register generated-surfaces work): a new
`scripts/build-register-live-view.mjs` generator that reconciles the
hand-authored on-box acceptance register live-view page
(`docs/testing/onbox-acceptance-register-live-view.html`) against the
canonical markdown register (`docs/testing/onbox-acceptance-register.md`) —
the summary-strip stats, per-group glance counts, per-section `gcount`
spans, and row-shell insert/delete/reorder for both ID'd groups and the
title-matched Blocked/Unconfirmed sections. `npm run register:build --check`
now runs in CI (`.github/workflows/onbox-register-check.yml`) alongside the
pre-existing `check-onbox-register.mjs`, and four comparisons in
`check-onbox-register.mjs` that this generator now supersedes were retired
(owed-total, glance per-group-count, gcount/header-count, and the
`'both'`-path missing-row branch) — their surviving `expected`/`found`/`extra`
computation, which still feeds `extraOnly`/`staleExtra`, was kept untouched.

**The committed diff to the live view, at the current head, touches only**:
the `<!-- END GENERATED:strip -->` marker's own indentation (4→2 spaces),
inter-shell blank-line formatting cleanup (a byproduct of row-shell
reconciliation's rebuild), and one row reordering (`E101` now follows
markdown document order, matching the `.md`'s own section order). The
summary-strip's numeric tiles are unchanged (main's copy already agreed with
this branch's by the time the merge conflict below was resolved). No shell
body, `iname`, `risk` span, `gtitle`, or Setup-cell text changed anywhere in
that diff — reviewed in full independently in the mandatory PR review
gate's round 2, not just asserted.

**This branch was rebased against a moved `main` via a merge commit
(`2416f7a8`)** after a merge conflict in the live-view HTML (the only
conflict) — resolved by taking `main`'s content as the base and re-running
the new generator against the merged markdown register, rather than
hand-resolving the diff.

**Two rounds of the mandatory PR review gate ran on this PR** (Premium
tier, `high` depth): round 1 found 3 Significant + 4 Minor findings,
including a real silent-data-corruption bug (`$`-pattern regex-replacement
special-sequence interpretation in `reconcileOneSection`/
`reconcileTitledSection`) and a data-loss bug (a `<summary>` tag on its own
line failed to match the row-shell locator regex, silently replacing real
row content with a placeholder). All 7 were fixed and independently
verified. Round 2 found 4 more (one of which was a flaw in round 1's own
added test — it could not actually fail against the bug it claimed to
guard), plus 3 doc-accuracy issues (an inverted allocation-floor rule, an
ambiguous "never hand-edit" claim that didn't distinguish the markdown's
own hand-maintained glance table from the generated HTML mirror, and an
uncaught crash — now a clean exit 1 — when a Blocked/Unconfirmed row's
title is reworded without a matching live-view edit). All fixed and
verified by mutation (each fix's paired test confirmed to fail against a
reintroduction of the exact bug it closes).

This wave closes **#2362 only**. It does **not** close #2721, #2599, #2603,
or #2708 — those were already closed via an earlier, separately-shipped
track (PRs #2846, #2856, #2837, #2840).

Closes #2362.

## Test plan

- [x] `node --test scripts/tests/build-register-live-view.test.mjs` — 38/38 pass, including the real-file `register:build --check` agreement test and every round-1/round-2 regression test.
- [x] `node --test scripts/tests/check-onbox-register.test.mjs` — 131/131 pass, including the pre-existing real-file "the real register and its real live view agree" test (kept per Task 8 Step 7) and the CLI real-file test.
- [x] `node --test scripts/tests/check-onbox-register-cli-no-flags.test.mjs` — 2/2 pass.
- [x] `npm run register:build -- --check && node scripts/check-onbox-register.mjs && node scripts/check-register-citations.mjs` — all exit 0 against the real committed files.
- [x] Every fix in both review rounds independently verified by mutation: reverted to the pre-fix behaviour, confirmed the new/updated test fails for the right reason (not a false-positive pass), restored the fix, confirmed the full suite is green again.
- [x] `.github/workflows/onbox-register-check.yml` runs `build-register-live-view.mjs --check` and its `paths:` filter includes `scripts/build-register-live-view.mjs`.
- [x] Read the full committed diff to `docs/testing/onbox-acceptance-register-live-view.html` at the current head — confirmed above.


dudarenok-maker:
## PR review — pass 1 (head 2416f7a8cac4785da45e2f60662fd5424a0c7a09, depth high)

Scope: `scripts/build-register-live-view.mjs` (new, 368 lines), `scripts/check-onbox-register.mjs` (the four retirements), `.github/workflows/onbox-register-check.yml`, `package.json`, both test files, and the full `docs/testing/onbox-acceptance-register-live-view.html` diff **including the merge commit's conflict resolution** · verified before probing: `node --test scripts/tests/build-register-live-view.test.mjs` 29/29, `check-onbox-register.test.mjs` 131/131, `check-onbox-register-cli-no-flags.test.mjs` 2/2, and `register:build --check` / `check-onbox-register.mjs` / `check-register-citations.mjs` all exit 0 against this head. Cloud CI is now **fully green — 20/20 SUCCESS**, including the legs that were still running at dispatch (lint, frontend, server fast+slow, all 8 e2e shards, visual baselines, build). Depth taken as `high` per #2786/#2886.

Adversarial work done rather than assumed: a 9-mutation battery against the **real** register/live-view pair to test the PR's central claim that the four retired comparisons left no coverage gap; a source-level mutation of `reconcileTitledSection` to test whether the idempotence tests can fail; and structural probes of the three rebuild paths (`reconcileOneSection`, `reconcileTitledSection`, `rewriteGcountInSection`).

### 🔴 Blocking

None.

### 🟠 Significant — a `$` in any row body is interpreted as a regex replacement pattern, silently corrupting the published page

**Correctness bug.** `scripts/build-register-live-view.mjs:346` (`reconcileOneSection`) and `:299` (`reconcileTitledSection`) both do `html.replace(sectionRegex, \`${headerAndOpen}${newBody}${closeTag}\`)` — a **string** replacement, so `$1`/`$2`/`$3`/`$&`/`` $` ``/`$'`/`$$` inside the reassembled content are expanded as substitution patterns. The regex has exactly three capture groups, so `$1`–`$3` are all live. Every one of those three interpolated strings is the file's own content, so any row body containing one of those sequences gets replaced with regex capture text. `rewriteGcountInSection:125` correctly uses a replacer *function* and is immune; these two do not.

Repro (run against the shipped module, output verbatim):

```js
const html = `<section class="group" id="ga">
  <header><h3 class="gtitle">T <span class="gcount">1 row</span></h3></header>
  <details class="item">
    <summary><span class="num">A1</span><span class="iname">first</span></summary>
    <div class="body"><p>Run <code>bash deploy.sh $1</code> and check <code>$&amp;lt;</code>.</p></div>
  </details>
</section>`;
reconcileRowShells('## Group A — setup a\n\n### A1 · first\n', html);
```

→ `bash deploy.sh $1` becomes `bash deploy.sh <section class="group" id="ga">…</header>` (capture group 1 spliced in), and `$&amp;` becomes the entire matched section re-inlined. Input 400 chars, output 907. `register:build` exits 0 and prints `wrote …` — the corruption is silent at write time; only a subsequent `--check` notices, by which point the corrupted page is what got committed and published.

Not hypothetical for this file: the register already carries `$PORT`, `$U.{pt,json}` in row prose, and an HTML-escaped `&` immediately after a `$` (`$&amp;`) is enough — you do not need a literal `$&` in the source. Fix is `(...args) => \`${headerAndOpen}${newBody}${closeTag}\`` at both sites, plus a paired test whose fixture body contains `$1` and `$&`.

### 🟠 Significant — the register's own runbook and CLAUDE.md now describe a workflow this PR deleted

**Correctness bug, in instructions agents follow literally.** This PR retires the owed-total, glance-per-group and gcount comparisons from `check:onbox-register` and moves the whole summary strip into a generated region, but leaves every document that describes that workflow untouched. Three statements are now false:

- `docs/testing/onbox-acceptance-register.md:98` — "`npm run check:onbox-register` verifies the owed total, the per-group counts and the row IDs across both files". It no longer verifies the owed total or the per-group counts; those moved to `register:build --check`. (I confirmed by mutation: an owed-total or glance-count drift in the live view now produces **zero** `checkLiveView` errors — see M1/M2 below.)
- `docs/testing/onbox-acceptance-register.md:107-108` — "**The rest of the summary strip is unchecked** — oldest debt, and the group/blocked/unconfirmed tallies. **Recompute those by hand.**" All five of those tiles are now *generated* (`buildStripRegion`, `:97-106`). Hand-recomputing them is now the wrong action: the generator overwrites it and `--check` rejects it.
- `CLAUDE.md:759` — "`npm run check:onbox-register` cross-checks the owed total, the per-group counts and the row IDs — **not** the rest of the summary strip (oldest debt, the group/blocked/unconfirmed tallies), which stay a manual recompute." Same two errors, in the file loaded into every session, inside the mandatory before-shipping step 3.

Related and in the same class: nothing outside `package.json` and the CLI's own failure message tells a maintainer that `npm run register:build` now exists or when to run it. The register's `## Live view` section still opens by calling the page "a hand-authored HTML page" (`:40`), and its publish procedure's step 0 (`npm run stamp:publish-token`, register `:177-181`) is not referenced by the generator's `--check` failure text, which says only "Run `npm run register:build` and commit the result" — routing an operator straight past the stamp step for a change that *did* touch the live view.

Reviewer-brief shape #9 ("the defect is in the instrument or the document"). The plan (`2026-08-29-onbox-register-generated-surfaces.md`) schedules no doc update either — Task 8's Files list omits both files — so this is unaddressed upstream as well, not merely skipped here.

### 🟠 Significant — the release-notes pair is missing, with no explicit not-applicable

**House-gate miss, not a correctness bug** (flagging the classification explicitly so triage does not mis-fire the re-review trigger). CLAUDE.md before-shipping step 5 requires an entry in `docs/release-notes-next.md` **and** `RELEASE_NOTES.md`, or an explicit statement that the change has no shippable delta. This PR touches neither file and its body says nothing about them. The precedent inside `docs/release-notes-next.md` is unambiguous for exactly this shape of work: `check:register-row-citations` (#2603/#2840, line 571), the `--against-published` mode (#1931, line 477), and the row-content hash (#2599/#2837, line 479) each got a technical-register entry for a new script + npm script + CI wiring. This PR ships a new `npm run register:build`, retires four comparisons from an existing operator-facing check, and changes the register's edit workflow — either it earns the same entry, or the PR body needs to say why not.

### 🟡 Minor

1. **A shell formatted across lines loses its whole hand-authored body to a placeholder.** `SHELL_BY_ID_REGEX` (`:216`) tolerates whitespace between `<details class="item">` and `<summary>` (`\s*`) but requires `<summary><span class="num">` to be adjacent with no separator. A hand-added row written as `<summary>\n  <span class="num">A1</span>…` is not recognised as an existing shell, so `reconcileOneSection` substitutes `buildPlaceholderShell` and the body is gone. Verified: body survived? `false`; placeholder emitted? `true`. `--check` catches the *placeholder* afterwards, so it is loud — but the content is already destroyed in the working tree. One `\s*` fixes it.

2. **Any non-shell content inside a group section body is silently deleted, and the module header overclaims.** `:6-8` says "Every hand-authored byte outside a generated target (BEGIN/END GENERATED region, or a row shell's own body/iname/risk spans) is preserved verbatim." That is not what the code does: `newBody` is rebuilt purely from the markdown's row list, so anything in the post-`</header>` body that is not a matched `<details class="item">` is dropped. Verified with a `<p class="note">` between the header and the first shell: `note survived? false`. This PR's own diff already falsifies the claim in the weaker form — it removed the inter-shell blank lines throughout the real page, which are hand-authored bytes outside every generated target. Either narrow the comment to what is actually preserved, or preserve it.

3. **`rewriteGcountInSection`'s "no gcount span found" error is unreachable, and `--check` fails open in exactly the case it names.** The regex (`:120-122`) is `<section[^>]*\bid="X"[^>]*>[\s\S]*?<span class="gcount">…` with nothing bounding the lazy scan at `</section>`, so when the target section has no `gcount` span the match runs into a *later* section's. Verified against the real files: delete `section#ga`'s `<span class="gcount">39 rows</span>` and `register:build --check` prints **"up to date." and exits 0** (A's count lands in `#gb`'s span, then B's own write restores it — net no-op, no error). Only the surviving `check-onbox-register.mjs` catches it ("Group A's header has no `<span class="gcount">N rows</span>`"), so the page is still protected — but the generator's own guard is dead, and a section for a letter with no markdown rows (not a write target) would keep another group's count.

4. **The idempotence tests cannot fail for `reconcileTitledSection`.** `buildRegisterFixture` (test file `:37-72`) has no `## Blocked` and no `## Unconfirmed` section, so `reconcileRowShells` never enters either titled branch — the two tests written to lock the property (`build then build is a no-op on an LF checkout`, `build then --check passes on an LF checkout`, `:424-437`) exercise only the ID-keyed path. Verified by mutation: re-introducing the exact `\n` before `closeTag` whose removal `:294-298` documents as the Task 8 Step 2 idempotence fix, **in `reconcileTitledSection` only**, leaves all 28 fixture-level tests green; only the real-file `register:build --check` test reddens (confirmed separately: the mutant returns exit 1 against the real files). Reviewer-brief shape #3. The property is currently protected by a fixpoint check on today's committed files rather than by a property test — add the two sections to the fixture.

### ✅ What is solid

- **The central claim of the PR — "the four retired comparisons left no coverage gap" — holds under independent mutation.** Against the real files: M1 live-view owed total → 99 (caught, `--check` exit 1), M2 `glance:A` → 99 (caught), M3 `section#ga` gcount → 99 (caught), M4 delete the `A2` shell entirely (caught), M5 register glance A 39→99 (caught by both `--check` and `checkRegister`'s surviving internal arithmetic). I also probed the one divergence the retirement *could* have opened — `parseBodyGroupCounts` counts by ID letter globally while `reconcileOneSection` sources rows per `## Group X` section, so a misfiled `### E101 ·` under `## Group A` produces a page whose `#ga` header says 39 while listing 40 shells. Confirmed the generator does produce that page (M7) — and confirmed it is still caught three independent ways: `checkRegister`'s glance-sum arithmetic, `checkLiveView`'s surviving `extra` check ("Group A section has row E101 that the register's Group A does not"), and `--check`'s placeholder guard. No gap. Today's real register also has zero misfiled headings (global letter counts and per-section counts agree exactly: A=39 B=2 C=4 D=3 E=13 G=2 H=2).
- **The merge resolution is sound and the generator was genuinely re-run.** `git diff ff28641e 2416f7a8 -- …live-view.html` contains exactly three kinds of change: the `E101` `<details>` block moved (byte-identical content, removed at one offset and re-added at another, matching the markdown's own Group E ordering where `### E101` follows `### E12`), the `<!-- END GENERATED:strip -->` marker re-indented 4→2 spaces (which is what `buildStripRegion`'s own trailing `\n  ` produces), and inter-shell blank-line removals. **No** shell body, `iname`, `risk` span, `gtitle`, Setup cell or summary-strip value differs from `main`'s — so `main`'s hand-authored content survived intact and the branch's own stale `40 → 37` A1 tile was correctly superseded by `main`'s `22 (of 60)` rather than re-applied. That is the shape a re-run produces, not a hand-resolved diff. `--check`, `check-onbox-register.mjs` and `check-register-citations.mjs` all exit 0 at this head, independently re-run.
- Deviations from the plan are deliberate, documented in-code, and correct: the dropped `\n` before `closeTag` at both reconcile sites (`:294-298`, `:341-342`), the `[ \t]*` indent capture added to both shell regexes, `process.exitCode` rather than `process.exit(main())` per `is-main-module.mjs`'s documented flush gotcha, and `normaliseTitle`'s `.split('`').join('')` instead of a regex literal (#2747).
- Enforcement parity with what was retired is preserved: the new test file is picked up automatically by `run-hooks-tests.mjs`'s `scripts/tests/*.test.mjs` glob, so both real-file assertions run in `test:hooks` (i.e. `verify:fast:branch`), and the workflow's `paths:` filter correctly gains `scripts/build-register-live-view.mjs`. CI runs `node scripts/build-register-live-view.mjs --check` directly rather than through npm, so the `npm run x --check` flag-swallowing trap is avoided (the PR body's prose spells it the swallowing way in one place; the wiring does not).
- `Closes #2362` is present, plain, outside any code span; #2362 is open, `area:ops` + `type:chore`, and matches the work.

### Verdict

**No blocking findings.** Three Significant: one real correctness bug in code (`$`-as-replacement-pattern, latent but one row body away and silent), one correctness bug in the instructions the register's own merge gate depends on, and one house-gate miss (release notes). Four Minor. The generator itself is well-built, the retirement is genuinely covered, and the merge resolution is verifiable as a real re-run rather than a strip of conflict markers — the findings are about content the generator does *not* protect and documentation the change made false, not about the reconciliation logic.

dudarenok-maker:
## PR review — pass 2 (head 6e4a3d36a2749a6e45eafec9ae8d9c9968e4afed, depth high)

Re-review round 2 of the `pr-review-gate` loop; pass 1 was [`2416f7a8`](https://github.com/dudarenok-maker/Castwright/pull/2945#issuecomment-5549517398). Depth re-derived from the current commit set, not carried over: 15 commits (one a merge from `main`) spanning `feat`/`fix`/`docs`/`test` across the `ops` and `scripts` scopes → multi-scope → `high`, same tier as pass 1 by the same reasoning.

Deltas only. Prior findings first, then what is new.

**Verification method** (not taken on trust from the commit messages): the module was copied to a scratch tree outside the repo and mutated there, one reversion at a time, to check each paired test actually goes red for the reason claimed; plus standalone probes against the real, unmutated module. No tracked file was touched.

Green baseline at this head: `build-register-live-view.test.mjs` 36/36, `check-onbox-register.test.mjs` 131/131, and `register:build --check` / `check-onbox-register.mjs` / `check-register-citations.mjs` all exit 0 against the real committed files. CI: all 20 checks SUCCESS. No red leg to fold in.

---

**1. 🟠 `$`-pattern regex-replacement corruption (`e69d7790`) — CODE FIX VERIFIED RESOLVED; ITS TEST FOR THE SECOND FUNCTION DOES NOT. See NEW-1.**

The code fix is real and is *wider* than the commit message claims. An independent probe fed a row body containing all five of JavaScript's special replacement patterns — `` $1 $2 $& $` $' $$ `` — through `reconcileRowShells` on both the ID-keyed and the title-keyed path, plus through `rewriteGcountInSection`'s new third capture group. All four preserved the body byte-for-byte. Mutation: reverting `reconcileOneSection` to the string template turns the suite red (1 test).

**2. 🟠 Stale docs describing retired `check:onbox-register` comparisons (`609e0154`) — PARTIALLY RESOLVED.** The four stale sentences are gone and the surviving-check descriptions in `CLAUDE.md:765` were spot-checked against the code and are accurate (`check-onbox-register.mjs:435` group-count, `:448` total-vs-sum, `:461-467` ID uniqueness, `:495-507` floor/next-id). But the replacement prose in `docs/testing/onbox-acceptance-register.md` introduces two false statements and restates a third — **NEW-2, NEW-3, NEW-4**.

**3. 🟠 Missing release-notes entries (`a48f746a`) — VERIFIED RESOLVED.** `docs/release-notes-next.md:573` and `RELEASE_NOTES.md:228`, both inside the in-progress `# Castwright 1.15.0` section (file top; the next heading is `1.14.0` at `:230`). One over-claim inside the added text — **NEW-5**.

**4. 🟡 `<summary>`-on-its-own-line regex miss (`0f1ac2ac`) — HALF VERIFIED RESOLVED.** `SHELL_BY_ID_REGEX` is fixed and genuinely guarded: reverting its `\s*` turns "SHELL_BY_ID_REGEX matches a shell with `<summary>` on its own line" red. The same commit widened `SHELL_BY_TITLE_REGEX` too, and reverting *that* half leaves the whole suite green — **NEW-6**.

**5. 🟡 False "preserved verbatim" header comment (`5bb40574`) — VERIFIED RESOLVED.** The new comment at `scripts/build-register-live-view.mjs:4-11` matches observed behaviour, and the added pinning test does exercise the drop (inter-shell comment gone, `</details>\n\s+<details` adjacency asserted). The retired over-claim reappears verbatim in this round's release notes, though — **NEW-5**.

**6. 🟡 Unbounded `rewriteGcountInSection` scan (`1f674046`) — VERIFIED RESOLVED.** Mutation: dropping the `(?:(?!<\/section>)[\s\S])*?` bound back to `[\s\S]*?` turns "rewriteGcountInSection fails when target section gcount is missing, even if a later section has one" red — so the test fails for exactly the claimed reason, not incidentally. Independently confirmed the new third capture group round-trips a section tail containing `$&`/`` $` ``/`$'` unchanged while still rewriting the count, so the widened match introduced no corruption of its own.

**7. 🟡 `reconcileTitledSection` idempotence test gap (`6e4a3d36`) — VERIFIED RESOLVED.** Mutation: reintroducing the `\n` before `closeTag` in `reconcileTitledSection` — the exact bug that function's own comment documents — turns both new tests red. One correction to the finding's premise: the gap was narrower than stated. The pre-existing real-file test `the real register and its real live view agree` *also* goes red under that mutation, because the real `.md` does have `## Blocked` / `## Unconfirmed` sections. The two new tests are still worth having (they make the coverage hermetic rather than dependent on the real file's current shape) — the claim that a regression "would leave every existing fixture test green" is what was overstated, not the fix.

---

### New findings

#### 🟠 NEW-1 — correctness bug (in the test): the `reconcileTitledSection` half of finding #1 shipped with a regression test that cannot fail

`scripts/tests/build-register-live-view.test.mjs:524-546`

Mutation, run against a scratch copy: revert **only** `reconcileTitledSection`'s replacer function back to the string template (`scripts/build-register-live-view.mjs:303`) — i.e. reintroduce exactly the bug pass 1 reported and `e69d7790` claims to have paired a test with. **The full suite stays green: 36/36, fail=0.** The sibling mutation on `reconcileOneSection` goes red, so the harness is not the problem — this one test is.

Why it cannot fail, reproduced standalone on the test's own fixture:

```
buggy === fixed : false
buggy length    : 772   fixed length: 387
occurrences of id="blocked": buggy 2, fixed 1     <- $& expanded the whole section into the body

the shipped test's two assertions, run against the CORRUPTED output:
  buggy: assert.match(/Runs on \$PORT and status is \$& \(always check\)/) = PASS
         assert.doesNotMatch(/Runs on <section/)                          = PASS
```

Two reasons, both structural:

1. The fixture body uses `$PORT`, not `$1`. `$P` is not one of JavaScript's replacement patterns, so it survives the buggy path untouched. The `reconcileOneSection` twin uses a literal `$1`, which *is* a pattern — that, not `$&`, is the only reason the twin goes red.
2. `$&` expands to **the whole match**, which is the entire `<section id="blocked">…</section>` — and that text contains the original body verbatim. So the corrupted output still contains `Runs on $PORT and status is $& (always check)` as a substring, and `assert.match` (a search over the whole output) finds it. `assert.doesNotMatch(/Runs on <section/)` misses for the same reason: the corruption inserts the section *after* `status is `, not after `Runs on `.

The shipped behaviour is correct — see the disposition of finding #1 above; this is a false guarantee, not a live defect. But accepting it as green means finding #1b was recorded as verified when nothing verifies it, and this repo's own catalogue puts "a test that cannot fail" at the top of its trap list. A discriminating assertion exists and is cheap: the corrupted output has two `<div class="body">` blocks where the correct one has one, or use a literal `$1`/`$&` pair the way the ID-keyed twin does.

#### 🟠 NEW-2 — correctness bug (instruction): the docs fix inverts the row-ID allocation rule

`docs/testing/onbox-acceptance-register.md:99-101` — "`npm run check:onbox-register` verifies the row IDs are unique across the document and **sit below their group's allocation floor**".

That is not the rule, in either half:

- `check-onbox-register.mjs:209-217` — the **allocation floor (101)** is the bar the group's `<!-- next-id -->` marker must sit **at or above**. Row IDs are never compared to the floor at all.
- `check-onbox-register.mjs:469-507` — what row IDs must sit strictly below is their group's **`next-id` marker**, not the floor.

Concretely, it is already false of the committed register: Group A's body today contains `A101, A102, A104, A105` (`ALLOCATION_FLOOR === 101`), so live rows sit at and above the floor, legally. Failure scenario: an agent reads this line, needs a new Group A row, and mints an ID "below the allocation floor" — say `A50`. `check-onbox-register.mjs` passes it (`A50 < next-id`, and uniqueness only sees *current* headings), because that file's own header states the residual limitation explicitly: "They do NOT stop someone hand-typing a discharged row's old ID back in: that ID is below the marker, so it passes." The result is a silently re-pointed citation — precisely the failure the floor was introduced to end (#2599/#2629/#2634/#2653). `CLAUDE.md:765`'s vaguer "each row ID's position relative to its group's allocation floor" is not wrong in the same direction, but names the wrong reference too and should be corrected in the same pass.

#### 🟠 NEW-3 — correctness bug (instruction): "Never edit the `## At a glance` section by hand" describes a workflow that fails CI

`docs/testing/onbox-acceptance-register.md:112-115` — "The summary strip is generated, not hand-edited. … Never edit the `## At a glance` section by hand; edit the rows instead and rebuild."

`## At a glance` is a section of **the markdown register — the generator's read-only input**. `scripts/build-register-live-view.mjs:167` writes exactly one file, `htmlPath`; it never writes the `.md`. The generated surface is the live view's `<div class="strip">` / `<table class="glance">`, which have no `##` headings.

Reproduced: add a row to Group A's body, bump `next-id`, leave the glance table alone (i.e. follow this instruction literally), then run the checker:

```
check:onbox-register exit = 1
  - Group A: glance table says 39, body has 40 rows (…). Update the table or the body.
```

So the instruction produces a red required check, and its remediation message (`Update the table or the body`) directly contradicts the sentence. Reword to scope the claim to the live view's strip and glance cells, and keep saying that the markdown's own `## At a glance` table is hand-maintained and checked against the body.

#### 🟠 NEW-4 — correctness bug: "a wording-only edit does not fail" is false for Blocked/Unconfirmed, and now crashes a required CI step

`docs/testing/onbox-acceptance-register.md:106-111` (rewritten in `609e0154`) — "**A wording-only edit does not fail.** Rewording a row … the most common edit this register gets — changes nothing either check compares."

True for `### <ID> · <title>` rows (shells are matched by ID; verified — rewording one leaves the live view byte-identical). **False for the Blocked and Unconfirmed sections**, which `reconcileTitledSection` matches by *title*. Reproduced against the real committed pair, rewording one Blocked heading (`AMD GPU support Phase 2` → `AMD GPU support, Phase 2`) and nothing else:

```
scripts/build-register-live-view.mjs:290
      if (matches.length === 0) throw new Error(`reconcileRowShells: no shell title matches …
Error: reconcileRowShells: no shell title matches "AMD GPU support, Phase 2" in #blocked
    at reconcileTitledSection (…:285:6)
```

This is an **uncaught exception**, not the module's own `console.error` + exit-1 path — so the new `Check the generated live-view surfaces are up to date` step this PR added to `onbox-register-check.yml` goes red with a stack trace and no remediation text, on an edit the register's own prose calls the most common one it gets. Two things are owed here and they are separable: the doc must stop promising the edit is free (minimum), and the CLI should catch and report this class the way it reports drift. Whether the generator should instead *rewrite* the matched `iname` from the markdown heading is a genuine design question — if that route is taken it needs a decision recorded, not a silent behaviour change.

#### 🟡 NEW-5 — cleanup: the release-notes entries re-assert the over-claim finding #5 just removed

`RELEASE_NOTES.md:228` — "…but it never needs hand-editing again."
`docs/release-notes-next.md:573` — "the HTML is committed as a verification surface …, never hand-edited."

The generator's own placeholder text says otherwise (`scripts/build-register-live-view.mjs:231`): *"Not yet published — run `npm run register:build` after adding row content, **or fill in manually** and re-run --check"* — and `main()` at `:145-151` fails `--check` for as long as any `class="body-placeholder"` remains committed. Every new register row therefore *requires* a human to hand-author its `<div class="body">` in the HTML; only the ordering, the shells and the derived figures are generated. This is the same over-claim `5bb40574` corrected in the source comment, relocated one file over in the same round.

#### 🟡 NEW-6 — cleanup: the `SHELL_BY_TITLE_REGEX` half of finding #4 has no test that can fail

`scripts/build-register-live-view.mjs:271`

`0f1ac2ac` widened two regexes and paired one test. Mutation: revert **both** `\s*` insertions in `SHELL_BY_TITLE_REGEX` (after `<summary>` and after `<span class="num">—</span>`) — the whole suite stays green, 36/36. Lower stakes than the ID-keyed twin, because an unmatched titled shell throws rather than silently placeholder-ing over hand-authored content, so this is a coverage gap rather than an unguarded data-loss path — but the commit is recorded as having shipped a regression test for a fix it only half-covers.

#### 🟡 NEW-7 — cleanup: the PR body describes a generated diff that is no longer the one being merged

The Summary says the committed generator output "touches **only**: the 'Still owed in A1' strip tile (40 → 37 …)", plus whitespace, inter-shell blank lines and the `E101` reorder.

At this head, `git diff main...6e4a3d36 -- docs/testing/onbox-acceptance-register-live-view.html` changes **no strip figure at all** — the A1 tile reads `22` on both sides (`docs/testing/onbox-acceptance-register.md:941` has `<!-- stat:a1-still-owed 22 -->`), and the only non-blank changes are the `E101` block's move and the indentation of `<!-- END GENERATED:strip -->`. The merge of `main` at `2416f7a8` moved the underlying figures after `dc1a048b` was written. The body's load-bearing safety claim — "No shell body, `iname`, `risk` span, `gtitle`, or Setup-cell text changed anywhere in that diff" — I re-derived independently at this head and it **does** hold (the `E101` `+`/`-` lines are byte-identical; everything else removed is blank lines). Only the 40 → 37 sentence needs correcting.

---

### Checked, not findings

- `register:build --check`, `check-onbox-register.mjs` and `check-register-citations.mjs` all exit 0 on the real committed files; `test:hooks` globs `scripts/tests/*.test.mjs`, so the new suite is genuinely wired into `verify:fast:branch` and the green `Lint, typecheck & fast checks` leg — not decoration.
- The bounded `gcount` regex is not a false-negative risk: every real section carries its `gcount` inside its own `<header>`, and the negative lookahead only stops the scan at `</section>`.
- The live-view `.html` changes rows' order but adds/removes no row and alters no row body, so the publish/`--against-published` procedure in `CLAUDE.md` step 3 is not triggered by row drift here; whether the reordering alone warrants a republish is an ops call, not a gate failure.
- `Closes #2362` is present, plain, outside any code span; the `PR issue link` check is green.

---

### Loop position

Six findings are correctness-shaped (four 🟠, and two 🟡 that are still owed a fix this round per the triage rule) — so the re-review trigger fires. This was pass 2; one further re-review round remains inside the skill's initial-pass + 2 cap.

**NEW-2, NEW-3 and NEW-4 all sit inside one 20-line block of `docs/testing/onbox-acceptance-register.md` (lines 96-115) and share one root cause.** Dispatching three fix agents serially against one worktree for one paragraph is the contention shape this repo has already hit three times; one agent briefed on all three is the cheaper and safer call here.

dudarenok-maker:
## PR review — pass 3 (head `ec70e277`, depth `high`) — **final round of the loop cap**

Deltas only, per the re-review format. Verified before probing: `node --test scripts/tests/build-register-live-view.test.mjs` **38/38**, `node --test scripts/tests/check-onbox-register.test.mjs` **131/131**, and all three real-file checks exit 0 (`build-register-live-view.mjs --check`, `check-onbox-register.mjs`, `check-register-citations.mjs`). CI is **fully green — 21/21 checks SUCCESS**, including `Verify / npm run verify`, all 8 e2e shards and the on-box register check. Working tree clean at `ec70e277` before and after this pass.

Round 2's seven fixes were re-verified independently, not read: each was re-mutated in a scratch copy of the repo outside the worktree (`scripts/` + `scripts/lib/` + a fake `docs/testing/` repo root), never in the branch.

---

### Round 2's findings — disposition

**1. `reconcileTitledSection` regression test could not fail — ✅ VERIFIED RESOLVED.** Reverted `scripts/build-register-live-view.mjs:318` to the string-template replacement form; the rewritten test now fails with `3 !== 1` at `scripts/tests/build-register-live-view.test.mjs:589`, exactly the claimed mechanism (`$1` → `headerAndOpen` and `$&` → whole match, each reinjecting a second `id="blocked"`). Also confirmed the *sibling* `reconcileOneSection` test was never toothless: reverting `:368` fails it at `:547` on `doesNotMatch(/bash deploy\.sh <section/)` — so round 2's asymmetric diagnosis was right, not a guess.

**2. Inverted allocation-floor rule — ✅ VERIFIED RESOLVED, read against the source not the doc.** `scripts/check-onbox-register.mjs:495-497` is `if (nextId < ALLOCATION_FLOOR)` → *"Group X's next-id (XN) is below the allocation floor"*; `:502-504` separately requires every row ID to sit **strictly below its group's own next-id marker**, never compared to the floor. The corrected wording in `docs/testing/onbox-acceptance-register.md` and both CLAUDE.md spots matches.

**3. "Never edit `## At a glance` by hand" ambiguity — ✅ VERIFIED RESOLVED.** The quoted error text is real and verbatim: `check-onbox-register.mjs:435` emits ``Group ${letter}: glance table says ${tableCount}, body has ${bodyNumbers.length} ${rowWord} (…). Update the table or the body.`` Reproduced live — bumping `| **E** | … | 13 |` to `14` fails with that exact sentence, exit 1. (But see 🟠 below: the same sentence is now asserted for a row class where it does not fire.)

**4. Reworded Blocked/Unconfirmed title crashed `main()` — ✅ VERIFIED RESOLVED, end-to-end through the real CLI.** Reverted `main()`'s try/catch: the new test at `:469` dies with a raw uncaught `Error: reconcileRowShells: no shell title matches …` stack. At head, against a scratch copy of the **real** register with one Blocked heading reworded, the actual CLI prints one clean line and exits 1 in *both* modes:
```
register:build: docs/testing/onbox-acceptance-register-live-view.html could not be reconciled
against docs/testing/onbox-acceptance-register.md: reconcileRowShells: no shell title matches
"AMD GPU support Phase 2 REWORDED" in #blocked. Run `npm run register:build` after fixing the
mismatch and commit the result.
```
Write mode also leaves the live view byte-identical (hash-compared) — it fails before writing anything.

**5. `SHELL_BY_TITLE_REGEX` `\s*` widening had no coverage — ✅ VERIFIED RESOLVED, with the control.** Reverting only `SHELL_BY_TITLE_REGEX` to its pre-`0f1ac2ac` form fails the new title-path test *and leaves the ID-path test green* — which is the proof that the coverage gap round 2 named was real.

**6. Release-notes over-claim — ✅ VERIFIED RESOLVED as written** (both files now scope the claim to derived figures and name the `body-placeholder` + markdown-glance-table manual steps). Partially undercut by 🟠 below.

**7. Stale PR-body diff description — ✅ VERIFIED RESOLVED, diffed independently.** `git diff main...ec70e277 -- …live-view.html` is 66 changed lines: **35 blank-line removals, 0 blank-line additions**, the `<!-- END GENERATED:strip -->` indent 4→2, and the `E101` shell moved verbatim (identical text on both sides). **Zero** changed lines touch a `class="n …"` stat tile, a `gcount` span, or a `GENERATED:glance` marker. The reorder is correct: markdown order is `E1…E12, E101`; `main`'s live view had `…E10, E101, E11, E12`; this head has `…E12, E101`. The body's description matches.

---

### 🟠 Significant — the Blocked/Unconfirmed row counts are *copied* from a hand-maintained cell, never derived from or compared against the rows — and both gates stay green when they disagree

**Correctness bug** (generated artifact states a count that contradicts what it renders), plus a doc claim added by round 2's own fix commit that asserts the opposite.

`scripts/build-register-live-view.mjs:202-203`
```js
html = rewriteGcountInSection(html, 'blocked', figures.blockedCount);
html = rewriteGcountInSection(html, 'unconfirmed', figures.unconfirmedCount);
```
Those two figures are parsed at `:57-60` from the markdown **glance table's `—` rows**. The lettered groups take the other route — `parseBodyGroupCounts` at `:199-201`, counted from the actual `### X<n> ·` headings, therefore self-correcting. Meanwhile `reconcileTitledSection` builds the Blocked/Unconfirmed *shells* from the section's real rows. Two sources, no comparison. `check-onbox-register.mjs`'s Check 1 (`:427-438`) iterates `tableLetters` only, and the file's header comment (`:36-41`) declares Blocked/Unconfirmed deliberately out of scope — a decision taken when those live-view numbers were hand-authored by someone who could see the rows.

**Repro, on a scratch copy of the real register + real live view, using only the documented workflow:**
1. Delete the last `### …` subsection from `## Blocked — hardware not available` (a discharge — the register's own most common structural edit), leave `| — | **Blocked** (hardware absent) | 6 |` untouched.
2. `node scripts/build-register-live-view.mjs` → `wrote …live-view.html.` exit 0
3. `node scripts/build-register-live-view.mjs --check` → `up to date.` **exit 0**
4. `node scripts/check-onbox-register.mjs` → `OK — … agree.` **exit 0**

Resulting committed page: `<section id="blocked">` header reads `<span class="gcount">6 rows</span>` over **5** rendered shells, and the summary strip's `<div class="n blk">` tile reads **6**. Verified by parsing the output.

**Matched control** (so this isn't a claim about an unmeasured path): the identical mistake on a lettered group *is* caught — `| **E** | … | 13 |` → `14` gives `Group E: glance table says 14, body has 13 rows (E1, …, E101). Update the table or the body.` exit 1.

Same hole for Unconfirmed (same two lines, same code path).

**Why this is a finding now rather than a pre-existing gap left alone:** this PR is what makes those two numbers machine-generated, and round 2's doc fix (`ec70e277`) newly asserts the guard exists —

- `docs/testing/onbox-acceptance-register.md:104` — "Both fail CI when rows are added, removed, or **miscounted**"
- `:124-126` — "Adding, removing, or reordering a row still requires hand-editing the markdown's own glance-table row count for that row's group — `check:onbox-register` fails (\"glance table says N, body has M rows … Update the table or the body\") **if you forget**"
- `CLAUDE.md:759` — "Its derived figures (owed count, per-group counts, oldest debt, **group/blocked/unconfirmed tallies**) are **generated** by `npm run register:build`"
- `CLAUDE.md:765` — "Together they catch structural drift (rows added/removed/**miscounted**)"

Each is true for a lettered Group row and false for a Blocked/Unconfirmed one. `docs/release-notes-next.md` and `RELEASE_NOTES.md` inherit it ("keeping the derived figures — counts, totals, row placement — in sync is no longer a manual, error-prone chore").

**Why the suite can't see it:** mutating `:202` to `figures.blockedCount + 1` *is* killed by the tests — but only because every fixture's glance count already equals its body-row count, so the two sources are never distinguished. A paired test needs a fixture where they disagree. Fix shape: count the Blocked `###` headings / Unconfirmed `- **…**` bullets (`reconcileRowShells` already extracts both lists) and either use them directly or assert them equal to the glance figures, with the error naming both numbers.

---

### 🟡 Minor

**a. `docs/testing/onbox-acceptance-register.md:115-119` over-claims for the Unconfirmed prefix path.** It says Blocked and Unconfirmed rows "are matched by TITLE, not ID — rewording one WITHOUT the matching live-view edit makes `register:build`/`--check` fail". Unconfirmed runs with `prefixMatch: true` (`build-register-live-view.mjs:340`), so a reword that *shortens* the bold lead to a proper prefix still matches. Reproduced on the real files: `- **fs-38 Wave 1** (designed-voice authoring…` → `- **fs-38 Wave** (…` → `--check` prints `up to date.` exit 0 and `check:onbox-register` prints OK, with the live view still carrying the stale `fs-38 Wave 1 — designed-voice authoring`. Matched control: lengthening it to `- **fs-38 Wave 1 redux**` fails with `no shell title matches "fs-38 Wave 1 redux" in #unconfirmed`, exit 1. The claim needs the prefix carve-out, or the match needs to be exact-after-normalisation like Blocked's.

**b. One assertion in each `$1`/`$&` test still cannot fail.** `scripts/tests/build-register-live-view.test.mjs:582` (`assert.match(result, /Port is \$1 and full match is \$&amp; below/)`) and its sibling at `:545` both pass against the un-fixed code — `$&`'s expansion reinjects the whole original match, the row's literal body text included, so a substring match always finds it. That is the exact property round 2 identified; the rewrite added the assertion that *does* work (`:589`) but kept the one that doesn't, under a comment reading "The literal text `$1` and `$&` must survive unchanged." Confirmed by mutation: with `:318` reverted the test fails **only** at `:589`; with `:368` reverted the sibling fails **only** at `:547`. Harmless today — each test is carried by its other assertion — but the inert line reads as the load-bearing one.

**c. Both CLAUDE.md enumerations dropped Check 4b's other half.** `CLAUDE.md:759` and `:765` now list what `check:onbox-register` validates and stop at the marker-vs-floor comparison; `check-onbox-register.mjs:502-504` also fails on any row ID **at or above its group's next-id** (`Group X: XN is at or above the group's next-id (XM). Bump next-id past every allocated ID.`). The parenthetical "the floor gates the next id a group may mint, **not the ids already in use**" is literally true of the floor but steers a reader into thinking existing ids are unchecked beyond uniqueness. Round 2 replaced a wrong sentence with an incomplete one.

---

### Verdict — **not clean; one 🟠 correctness finding, and this is the final round**

Round 2's seven fixes all hold; each was independently re-derived by mutation, not accepted on the commit message. Two of them (#1's rewritten test, #4's `main()` catch) were re-proved end-to-end.

But round 2's own doc fix introduced a claim the code does not honour, and probing it surfaced a real hole in the generator: for the two `—` sections, the row count written into the live view (and into the summary strip's `blk` tile) comes from a hand-maintained table cell that nothing validates, while the rows next to it come from the body — so the page this PR exists to make trustworthy can still state a number that contradicts itself with `register:build --check` and `check:onbox-register` both reporting green. Demonstrated with a matched control on the real files, not inferred.

**The loop cap is reached** (initial pass + 2 re-review rounds = 3). Per `.claude/skills/pr-review-gate/SKILL.md`, no further automatic re-review round follows: fixing the 🟠 and the three 🟡s does **not** re-trigger a pass. This goes to the repo owner to decide — land the fixes and merge on their own judgement, or split the count-derivation work out under its own issue. My read is that the 🟠's code fix is small and well-specified (no design decision: the count must come from the rows, same as every lettered group already does), so it belongs in this PR under the report-fix-record rule rather than in a follow-up ticket; the doc sentences at `register.md:104`/`:124-126` and `CLAUDE.md:759`/`:765` should move with it.

Issue #2086, opened 2026-08-04
srv-87 — an interim cast.json write can persist a character-id swap with no history record


dudarenok-maker:
## What

`runMainAnalyzerJob` / `runSubsetAnalyzerJob` perform three **interim** `cast.json` writes (currently `analysis.ts:3630`, `:3840`, `:5607`) that call `mergeAnalysisResultWithExistingCast` and **discard** its `.retirements`.

When the merge's name-fallback fires on one of those writes, `claimedByName` suppresses carry-forward, so the prior character id is removed from `cast.json` while nothing records the old→new mapping. If the run then dies, is paused, or is displaced by a `fresh: true` POST before the authoritative end-of-run write, `cast.json` permanently holds the fresh id and that character's frozen `<slug>.segments.json` entries orphan to the narrator.

This is the residual of #2040 Wave 2's final-review finding 3.

## Why it is NOT simply "record them too"

That was tried and correctly refused — the refusal is the important part of this ticket.

At an interim write, `from` is only **provisionally** dead. `buildInterimCast` merges only chapters already in `chapterCast` (partial by construction) and `previewFoldForLiveView` runs with no sentence counts, so the final dedup can still collapse the id the retirement points *at*. Recording there makes `retireCharacterId`'s repoint loop rewrite **pre-existing, working** history entries onto an id that then dies:

```
history {"alden-old": "alden"}
interim records  alden -> alden-x
repoint loop     {"alden-old": "alden-x"}
final dedup collapses alden-x
buildCastResolver drops the entry (target not live)  -> resolution that WORKED is destroyed
```

Wave 2's `refuseRetirementsOfLiveIds` cannot catch it: `cand.id` is in `claimedByName`, so it is genuinely absent from the interim roster it would be filtered against. And every `from` at that point is in fact live in the pre-run `cast.json` — the only durable roster — so the honest roster refuses 100% of them while the roster that lets any through is the one that isn't real.

So the naive fix corrupts working history on **every** successful run with id drift, to close a gap that needs a mid-run death.

## Acceptance

A character-id swap persisted by an interim write is recoverable after an interrupted run, **without** repointing history entries against a provisional roster. Candidate directions (all need design work — none is a drop-in):

1. Record interim retirements to a separate provisional area, promoted at the authoritative write and discarded otherwise.
2. Suppress the name-fallback entirely on interim writes, so no interim write can change a persisted id.
[...]

PR #2998, opened 2026-09-05
fix(server,docs,scripts): run the interim-prune e2e test in the serial slow lane

### LINKED ISSUE #2947: analysis.interim-prune-prohibition.e2e.test.ts times out its beforeAll hook under pool contention, blocking every server commit
## What

`server/src/routes/analysis.interim-prune-prohibition.e2e.test.ts` times out its `beforeAll`
hook (`Hook timed out in 30000ms`) under pool contention, failing `test:server` and therefore
blocking **every** `server/**` commit on a loaded box.

## Why it is not a flake to wait out

It is the **fifth instance of a documented class**, and the repo already knows the shape.
`server/vitest.config.slow.ts`'s own header says the hot files "all share the same shape:
`mkdtempSync` + module imports in `beforeAll` racing on Windows tmpdir under parallel-fork
pressure", and `server/vitest.config.ts:19-29` records that plan 45 dropped `maxForks` 4 → 2
and raised `hookTimeout` to 30s after four route files hit exactly this.

This file has precisely that shape — `mkdtempSync` followed by **five sequential dynamic
imports** in `beforeAll` (`analysis.interim-prune-prohibition.e2e.test.ts:172-186`) — but it
was never added to the serial lane the other ten live in.

## Evidence (2026-09-05, box at ~37 concurrent node/vitest processes)

Reproduced **on untouched `main` in the primary checkout**, so it is not caused by any branch
in flight:

| Run | transform | outcome |
|---|---|---|
| fresh worktree, `--retry=0` | 29.91s | `Hook timed out in 30000ms`, test **skipped** |
| primary checkout on `main`, `--retry=0` | 21.93s | same |

The test never executes — it is reported as `1 skipped`. The transform alone consumes most or
all of the 30s hook budget, which is the timeout's proximate cause.

## The fix

Add it to the serial single-fork lane, exactly as the ten existing hot files were handled:

1. Add `'src/routes/analysis.interim-prune-prohibition.e2e.test.ts'` to `SLOW_FILES` in
   `server/vitest.config.slow.ts`, with a one-line symptom note matching the style of the
   entries above it.
2. **Honour the mirror invariant** the config states in terms: *"each entry in SLOW_FILES
   below MUST also appear in server/vitest.config.ts's `test.exclude` array. Add a file in one
   place and the main run picks it up too (double-runs). Add in the other only and the file is
   not exercised."* So add it to `SLOW_FILES_TO_EXCLUDE` in `server/vitest.config.ts:35` too.
3. Update the counts both configs state in prose ("the 10 hot files" → 11).
4. Update the two figures in `CLAUDE.md` that this change stales: `test:server-slow`'s "10
   timeout-prone server test files" → 11, and the `test:server:routes` note "selects 138
   files, not the 146 … 8 are in `vitest.config.slow.ts`'s `SLOW_FILES`" → 137 and 9, since
   the new entry is itself under `server/src/routes/`.
5. Update whatever guard enforces the mirror invariant so it still passes.

## Acceptance

- `npx vitest run --config vitest.config.slow.ts` includes the file and it **passes** serially.
- `npm run test:server` no longer selects it (no double-run, no timeout).
- The mirror guard passes.
- Both configs' prose counts and CLAUDE.md's two figures agree with reality.

## Benefit

*Benefit (technical):* unblocks every `server/**` commit on a contended box, and closes the
last known instance of a failure class this repo has already fixed four times.

Found while committing salvaged work for #2054; unrelated to that change, surfaced rather than
folded in per CLAUDE.md's commit-gate triage rule.

dudarenok-maker:
## Summary

`server/src/routes/analysis.interim-prune-prohibition.e2e.test.ts` timed out its 30s `beforeAll` hook under pool contention, failing `test:server` and therefore blocking **every** `server/**` commit on a loaded box.

It is the fifth instance of a documented class. `server/vitest.config.slow.ts`'s own header already describes the shape — `mkdtempSync` plus module imports in `beforeAll` racing on the Windows tmpdir under parallel-fork pressure — and plan 45 moved four route files to the serial single-fork lane for exactly this reason. This file has that shape (`mkdtempSync` followed by five sequential dynamic imports at `analysis.interim-prune-prohibition.e2e.test.ts:172-186`) but was never added.

**It was not the branch, and it was not a flake.** Reproduced identically on untouched `main` in the primary checkout with `--retry=0`:

| Run | transform | outcome |
|---|---|---|
| fresh worktree, `--retry=0` | 29.91s | `Hook timed out in 30000ms`, test **skipped** |
| primary checkout on `main`, `--retry=0` | 21.93s | same |

The test never executed — vitest reported `1 skipped`. The TypeScript transform alone consumed most of the 30s hook budget, which is the timeout's proximate cause.

### What changed

- Adds the file to `SLOW_FILES` in `server/vitest.config.slow.ts` and mirrors it into `SLOW_FILES_TO_EXCLUDE` in `server/vitest.config.ts`, per the mirror invariant both configs state.
- Updates the prose counts in both configs (10 → 11).
- Corrects the two figures in `CLAUDE.md` the change stales: `test:server-slow`'s file count (10 → 11), and `test:server:routes`' "selects 138 files … 8 are in `SLOW_FILES`", which becomes 137 / 9 since the new entry is itself under `server/src/routes/`.

### Also fixed, found in passing: the mirror invariant had no guard

Both configs state that an entry must appear in **both** lists, but nothing enforced it — grepping `SLOW_FILES` across `server/src` and `scripts` returned zero hits. Both drift directions fail silently:

- a file in `SLOW_FILES` only → it **double-runs** (parallel lane *and* serial lane);
- a file in the exclude list only → it is **exercised nowhere**.

Adding an eleventh entry by hand across two files made that gap load-bearing, so `SLOW_FILES_TO_EXCLUDE` is now exported and `server/src/slow-lane-mirror.guard.test.ts` asserts the two are equal as sets — plus, separately, that every `SLOW_FILES` entry appears in the **resolved** `test.exclude`, since two constants can agree with each other while the spread wiring them together is broken or removed.
## Test plan

Guard: `server/src/slow-lane-mirror.guard.test.ts` — **8 assertions**, all passing. It grew across two review rounds; the counts below are current as of `1a9c764d`.

It covers **five distinct drift shapes**, each mutation-proved red with observed output:

| # | Drift shape | Caught by |
|---|---|---|
| 1 | The two constants disagree (either direction) | set equality, both directions |
| 2 | The constants agree but the **main** config's `exclude` spread is broken | effective-`test.exclude` check |
| 3 | The constants agree but the **slow** config's `include` wiring is broken | effective-`test.include` check |
| 4 | `scripts/flake-repro.mjs`'s third copy of the list drifts | two-way set equality against it |
| 5 | A listed file no longer exists, or the whole list is emptied | `existsSync` check + populated floor |

Shapes 2 and 3 are the reason constant-comparison alone is not enough: the two lists can agree with each other while nothing wires them into the config that uses them. Shape 5's floor is deliberately a floor, not `=== 11` — an exact count would fail on every legitimate future addition.

The instructive mutation: emptying both arrays together left **six of seven** assertions passing before the floor was added.

The guard reads all three sources at runtime — the two configs via runtime-computed dynamic `import()` (both live outside `server`'s `rootDir`, so a static import fails `tsc -p .`), and `flake-repro.mjs` by parsing its source, because that module calls `process.exit(2)` at eval time when `--file` is absent and would kill the test runner.

**Suite runs:** the full server suite passed on `e92d0e3d` (538 files / 8206 tests, 0 failures, 17m) and cloud CI is green on every commit, including *Server tests (fast + slow)*. The subject file is no longer selected by `test:server` and passes serially under `vitest.config.slow.ts`. `force-rerun-triggers.test.ts` green throughout.
Superseded by the Test plan above and the round notes below. For the avoidance of doubt at merge time: the full `test:server` suite ran green on `e92d0e3d` (538 files / 8206 tests) under the local gate, and **cloud CI has run the full suite on every subsequent head**, including `1a9c764d` — run `33962555671`'s server leg reports `538 passed | 1 skipped` main plus `11 passed` slow, with Lint, Typecheck and Hooks all green. Pre-push `verify:fast:branch` did **not** run on the `--no-verify` heads; CI is the gate that covers them, and it did.

The guard uses a runtime-computed dynamic `import()` rather than a static one: both configs live outside `server`'s `rootDir` (`server/src`), so a static import fails `tsc -p .` with "File is not under rootDir". This mirrors what `force-rerun-triggers.test.ts` already does for the same reason.

### Checklist notes

- **Regression plan** — not applicable; small and localized, so the issue body plus the paired guard test is the spec.
- **On-box acceptance** — not applicable; no behaviour here needs real hardware to prove.
- **Release notes** — deliberately skipped: internal test-infrastructure only, with no user- or operator-visible delta.

Closes #2947

---

## Review round 1 — five findings, all fixed in `0c981729`

The review pass found that this PR's sweep was incomplete: the slow-file list has **three** enumerated copies in tracked code, not two, and its count appears in **five** prose sites, not three.

- **C1** 🔴 — `scripts/flake-repro.mjs` held a third hardcoded copy (ten substrings matched against `--file`). None matched the new file, so the harness picked the main config, which this branch had just taught to exclude it. That tool worked before this branch. Fixed, **and** the relation is now guarded.
- **C2** 🟡 — three comments still saying "10": `vitest.config.ts`, `vitest.config.slow.ts`, `verify-cache.mjs`. A repeat of drift the 2026-06-17 hardening plan already documented.
- **N1** — `test:server`'s verify-cache inputs omitted `vitest.config.slow.ts`, so the new guard didn't run locally on the exact diff shape it exists to catch.
- **N2** — the guard never asserted the listed files exist; every check compared strings to strings.
- **N3** — the whole invariant went vacuous if both arrays were emptied. **Six of seven assertions passed on an empty invariant**; only the new floor caught it.

Guard is now **7 assertions, 7 passing**. Every new assertion was mutation-proved red with observed output.

Scope widened to `scripts` accordingly.

## Note on the local gate for `0c981729`

That commit was made with `--no-verify`, **explicitly authorized by the repo owner**, because the local hook was blocked by an unrelated cross-lane collision: #2999 synced its *unmerged* edit of `.claude/skills/pr-review-gate/SKILL.md` into `~/.agents/skills/`, which is per-machine and shared by every worktree. The drift guard then failed here, on a branch whose copy is byte-identical to `origin/main`. Filed as #3001.

**Cloud CI is the honest gate for this commit and is unaffected** — that guard skips by design on a machine with no skills store, which is every CI runner.

What was verified locally by hand before the bypass:

- `test:hooks` — **exactly one** failing assertion, the environmental mirror check. Everything else passed, including the `stepTouchedByDiff` tests covering the N1 change.
- `slow-lane-mirror.guard.test.ts` + `force-rerun-triggers.test.ts` — 76 passed.
- `tsc -p .` — exit 0.
- The full server suite (538 files / 8206 tests, 0 failures) passed on `e92d0e3d`; this commit adds no runtime server code beyond the guard test itself.

---

## Review round 2 — three findings, all fixed in `1a9c764d`

Pass 2 marked all five of pass 1's findings **VERIFIED RESOLVED** (re-running the mutations rather than trusting the receipts) and found three new ones. Two are the same mistake in different places: **a relation guarded in one direction only.**

- **P1** 🔴 — `flake-repro.mjs` matched by **substring**, so `'generation'` also matched `generation-error.test.ts` and `'book-state'` a dozen more. **34** `server/src` files matched the list; only 11 are slow, so **23 were broken today** with the same `No test files found` exit 1 as C1. The guard I added for C1 was blind to it: it asserted every slow file *is* matched and never that no non-slow file is. Fixed structurally — `SLOW` now holds the 11 exact paths and matches with `SLOW.includes(rel)`, so the class cannot recur, and the assertion collapses to a two-way set equality. Retires the dead `'generation-boundary-recycle'` entry.
- **P2** 🟡 — the guard checked the **main** config's effective `exclude` but never the slow config's own wiring. Replacing `include: SLOW_FILES` with a glob left it fully green while two files became selected by neither config. Mirror assertion added.
- **P3** 🟡 — the `const SLOW` regex was unanchored, and `String.match` without `/g` takes the first hit, so a commented-out old declaration above the live one silently yielded the wrong list. **Verified load-bearing before fixing.** Anchored under `/m`; if the live declaration were ever indented it now throws, which is correct for a guard.
- **P4** — this body's stale "4 tests". Fixed above.

## Note on the local gate for `0c981729` and `1a9c764d`

Both were committed with `--no-verify`, **explicitly authorized by the repo owner**, for one unrelated blocker: #2999 synced its *unmerged* edit of `.claude/skills/pr-review-gate/SKILL.md` into `~/.agents/skills/`, which is per-machine and shared by every worktree, so the drift guard in `test:hooks` fails here on a branch whose copy is byte-identical to `origin/main`. Filed as **#3001**, which names the design decision owed rather than picking one.

**Cloud CI is the honest gate for these commits and is unaffected** — that guard skips by design wherever no skills store exists, which is every runner. Verified locally by hand each time: `test:hooks` had *exactly one* failing assertion (the environmental one), guard + `force-rerun-triggers` green (77 passed), `tsc -p .` exit 0.


dudarenok-maker:
## PR review — pass 1 (head e92d0e3d, depth high)

Depth `high` per the multi-scope trigger (`fix(server,docs)`). All arithmetic in the PR body was re-derived against the real tree rather than taken on trust; both lane-selection directions and the guard's non-vacuity were verified by running vitest, not by reading.

The core change is right and the numbers are right. The problem is the sweep: the slow-file list has **three** enumerated copies in tracked code, not two, and the count appears in **five** prose sites, not three. The PR updated two of three and two of five.

---

### 🔴 Correctness bugs

**C1 — `scripts/flake-repro.mjs` is a third, un-mirrored copy of the slow-file list. This PR leaves it stale, so the repo's own load-repro harness now exits 1 on the exact file the PR moved.**

`scripts/flake-repro.mjs:16-19`

```js
const SLOW = ['analysis-pipelining', 'gemini', 'book-state', 'chapters-restructure',
  'generation', 'generation-boundary-recycle', 'pdf-real', 'setup-readiness.route',
  'kokoro-install.route', 'venv-bootstrap.route'];
const isSlow = SLOW.some((s) => file.includes(s));
```

Ten substrings, none of which occurs in `analysis.interim-prune-prohibition.e2e.test.ts`, so `isSlow` is `false` and the harness picks the **main** config — which this PR just taught to exclude the file.

Repro, run in the worktree at `e92d0e3d`:

```
$ node scripts/flake-repro.mjs --file server/src/routes/analysis.interim-prune-prohibition.e2e.test.ts --runs 1

 RUN  v4.1.11 C:/Claude/Projects/wt-2947-slow-lane/server
No test files found, exiting with code 1
filter: src/routes/analysis.interim-prune-prohibition.e2e.test.ts
run 1: 4202ms exit=1
SUMMARY [{"run":1,"ms":4202,"code":1}]
```

Before this PR the same command ran the file (main config included it). This is a regression the PR introduces, and it lands on the one tool you would reach for to investigate *this* file's load sensitivity. `flake-repro.mjs` is live, not dead code — `docs/testing/deterministic-test-playbook.md:35` and `docs/testing/flake-evidence.md:4` both prescribe it as the acceptance instrument for exactly this failure class.

That third site is also the reason the incidental fix is under-scoped. The guard mechanises a **two-way** invariant; the real invariant is **three-way**, and the new guard cannot see the third leg. Enumerated by searching for a marker that only appears in a slow-list, outside `docs/`:

```
$ git grep -l -E "kokoro-install\.route|venv-bootstrap\.route" -- ':!docs'
scripts/flake-repro.mjs
server/src/routes/venv-bootstrap.route.test.ts
server/vitest.config.slow.ts
server/vitest.config.ts
```

Four hits, one of which is the test file itself — so `flake-repro.mjs` is the only unmirrored list, and it is enumerable, which means the guard can cover it. Note the fix widens the diff into `scripts/`, its own scope per CONTRIBUTING.md's scope table (`scripts` → `scripts/`), so the title becomes `fix(server,docs,scripts)`.

---

### 🟡 Cleanup / chores (owed in this round per CLAUDE.md, no re-review trigger of their own)

**C2 — three comments this change makes false; the count sweep stopped at two of five sites.**

| Site | Text | Now |
|---|---|---|
| `server/vitest.config.ts:201` | "it selects only the **10** slow files" | 11 |
| `server/vitest.config.slow.ts:79` | "silently select zero of the slow suite's **10** files" | 11 |
| `scripts/verify-cache.mjs:458` | "Plan 45 (vitest pool tuning) — **10** hot files" | 11 |

Updated: `vitest.config.slow.ts:3`, `CLAUDE.md:338`, `CLAUDE.md:340`. Missed: the three above. Two of the misses are in files the PR already has open.

This is catalogue shape #9 (the defect is in the document) plus #10 (one instance fixed, the class left armed), and it is a *repeat*: `docs/superpowers/plans/2026-06-17-flaky-test-release-hardening.md:770` records the previous round of this same drift and names both carrier sites explicitly — *"Change '5 hot files' → 10 in `server/vitest.config.slow.ts:3` AND `scripts/verify-cache.mjs:~98`"*. `verify-cache.mjs` was a known, documented second site and was still missed.

Per CLAUDE.md ("a stale comment or doc sentence your change made false" is a chore fixed in the same round), these are owed here, not filed.

**N1 — `test:server`'s verify-cache inputs omit `server/vitest.config.slow.ts`, so the new guard is skipped locally on the exact diff shape it guards.**

`scripts/verify-cache.mjs:372` (the `test:server` step), `extraFiles` at `:436-455`.

The new guard lives in `server/src/` (so it belongs to `test:server`) but reads `server/vitest.config.slow.ts` at runtime via dynamic import — no module-graph edge. That file is in `test:server-slow`'s `extraFiles` (`:467-472`) but not `test:server`'s. Probed against the real module:

```
$ node probe.mjs   # imports STEPS + stepTouchedByDiff from scripts/verify-cache.mjs
slow-config-only     shared=false  test:server=false  test:server-slow=true
main-config-only     shared=false  test:server=true   test:server-slow=true
guard-test-only      shared=false  test:server=true   test:server-slow=true
```

So a branch whose whole diff is `server/vitest.config.slow.ts` — e.g. deleting an entry from `SLOW_FILES`, precisely the drift the guard exists to catch — has `test:server` skipped as out-of-scope by pre-push `verify:fast:branch` and pre-commit `verify:fast:scoped`. `test:server-slow` does run, but the guard is not in `SLOW_FILES`, so it runs nowhere locally.

**CI is a backstop**, so this is a local-gate hole and not a merge hole: `verify.yml`'s leg condition is the union `step_test_server || step_test_server_slow || shared` (`.github/workflows/verify.yml:346`), and `vitest.config.slow.ts` *is* in the main config's `forceRerunTriggers` (`server/vitest.config.ts:205`), so `npx vitest run --changed "$BASE"` force-reruns the whole suite and the guard fires. That is why this is a chore rather than a bug.

It is the same #1847 runtime-read trap this same array already fixes for `openapi.yaml`, `server/tts-sidecar/main.py`, `.gitattributes` and `scripts/repair-cast-id-drift.mjs` — each with a comment explaining why. One line: add `'server/vitest.config.slow.ts'` to `test:server`'s `extraFiles`. (Pre-existing for `force-rerun-triggers.test.ts`, which has the identical runtime-read shape; this PR makes it load-bearing a second time, and "it's pre-existing" is not a deferral reason here.)

**N2 — the guard never asserts the listed files exist, so a rename defeats it silently, with both lists still in perfect agreement.**

`server/src/slow-lane-mirror.guard.test.ts:56-77`

All four assertions compare the two string arrays (and the `exclude` derived from one of them) to each other. Nothing touches the filesystem.

Scenario: `src/parsers/pdf-real.test.ts` is renamed to `pdf-real.pool.test.ts` and neither list is updated. Then (a) the slow config's `include` matches 10 of 11 and vitest runs 10 and exits 0; (b) the main config's `exclude` entry, a literal relative path, no longer matches the renamed file, so it rejoins the parallel pool — which is the "Worker exited unexpectedly" destabilisation the entry exists to prevent. All four guard assertions stay green throughout.

Vitest's silence on a nonexistent `include` entry was verified, not assumed. Against `--root server` with a throwaway config outside the repo:

```
include: ['src/routes/this-file-was-renamed-and-nobody-noticed.test.ts',
          'src/parsers/pdf-real.test.ts']

$ npx vitest list --filesOnly --config <scratch>/probe.slow.config.ts
src/parsers/pdf-real.test.ts
EXIT: 0
```

No error, no warning, no mention of the missing entry. The repo already knows this — `scripts/quarantine-health.mjs:66-70` documents that vitest hard-errors only when the file list resolves to *zero*.

One `existsSync(resolve(SERVER_ROOT, f))` per entry closes it.

**N3 (minor) — test 3 is implied by tests 1+2, and all four go vacuous if both arrays are emptied together.**

`server/src/slow-lane-mirror.guard.test.ts:67-69` and `:71-76`

`new Set(A)` vs `new Set(B)` adds nothing over the two `filter` checks except duplicate-blindness, which they already share. Separately, with `SLOW_FILES = []` and `SLOW_FILES_TO_EXCLUDE = []` all four pass having asserted nothing — the `for` loop at `:73` iterates zero times. A downstream backstop exists (an empty `include` makes `test:server-slow` exit 1 with "No test files found"), so this is a shape note rather than a live hole. `expect(SLOW_FILES.length).toBeGreaterThan(0)` removes it; fold it into N2's existence check.

---

### ✅ Verified, holds

Checked rather than assumed:

- **Arithmetic, measured against the real tree.** `find server/src/routes -name '*.test.ts'` → **146**. `npx vitest list --filesOnly` (main config) → **539** files, **137** under `src/routes/`, **0** matching `interim-prune`. `npx vitest list --filesOnly --config vitest.config.slow.ts` → **11**, including the new entry. So CLAUDE.md's 137 / 146 / 9 and both configs' "11" are correct — 146 − 9 = 137 is measured, not just self-consistent. The 9 routes entries + `analyzer/gemini` + `parsers/pdf-real` = 11 also matches CLAUDE.md:340's parenthetical breakdown.
- **Both directions of the move.** Excluded from `test:server`, included by the slow config — confirmed by the two `vitest list` runs above, not by reading the arrays.
- **The moved test actually passes serially.** `npx vitest run --config vitest.config.slow.ts src/routes/analysis.interim-prune-prohibition.e2e.test.ts --retry=0` → 1 passed, 20.60s (transform 15.10s). This matters given the PR's own evidence that the test had never executed. Worth stating somewhere: `hookTimeout` is the same 30 s in *both* configs, so the fix buys removal of contention, not headroom — the margin on a clean box is roughly 2×.
- **Assertion 4 is not vacuous.** During the C1 repro, vitest printed its own resolved `exclude` array: it contains all 11 SLOW_FILES entries, which reach it only through `...SLOW_FILES_TO_EXCLUDE`. So the dynamic import resolves the spread rather than reading a constant, and a removed spread (or a missing `test.exclude`) makes `toContain` fail. The PR's mutation table A/B/C is consistent with the array logic.
- **`tsc -p .` in `server/`** → exit 0. The runtime-computed dynamic import plus top-level await do keep both configs out of the rootDir-checked program, as claimed; `force-rerun-triggers.test.ts` is the standing precedent.
- **`force-rerun-triggers.test.ts`** → 69 passed. Both configs are already in `forceRerunTriggers`, so the new guard is reachable under `vitest run --changed` in CI.
- **The 76.9% figure survives.** The moved file contains exactly **1** test, so the `test:server:routes` share moves by 1/8207 — 76.9% is unchanged and correctly left alone.
- **No third routing site beyond C1.** `scripts/quarantine-health.mjs:48-53` deliberately does *not* hardcode the list (it tries both configs per file and lets the non-owner report zero) — that one is correct by construction.
- **House gates.** `Closes #2947` present and outside a code span; #2947 exists, open, labelled `bug` + `area:srv`, and its Acceptance section matches what shipped. Incidental fix declared under "Also fixed, found in passing". Regression plan / on-box acceptance / release notes each explicitly marked N/A with a stated reason — all three defensible (no hardware-provable behaviour, no user- or operator-visible delta). No `cast.json` writes, no config knob, no derived artifact. All 19 CI checks green.

### Observation, not a finding

Moving the file drops it out of the **local** gate: `test:server-slow` is in neither `verify:fast:scoped` (pre-commit) nor `verify:fast:branch` (pre-push), so the #2086 orphaned-audio invariant this test pins is now enforced only in cloud CI and a full `npm run verify`. That is the accepted, documented cost every one of the 11 slow-lane files pays, and net coverage still improves over a test reporting `1 skipped` — but the PR body doesn't say it, and it is the kind of gating delta this repo normally states out loud.

---

### Verdict: **changes requested**

1 correctness bug (C1) → re-review after fix. C2 is owed in the same round per CLAUDE.md's stale-comment rule. N1–N3 are hardening; N1 and N2 are cheap and both close real, demonstrated holes in the guard this PR introduces as its own deliverable.

<sub>Reviewed at `e92d0e3d`; worktree verified unmodified (`git status --porcelain` empty) before and after.</sub>

dudarenok-maker:
## Review round 1 — all five findings fixed, commit parked on an unrelated blocker

Every finding from [pass 1](https://github.com/dudarenok-maker/Castwright/pull/2998#issuecomment-5551015711) is fixed and staged. The commit **cannot land yet** — not because of anything in this branch. Detail at the bottom.

Each fix was a separately dispatched agent, one finding each.

| Finding | Disposition |
|---|---|
| **C1** 🔴 `flake-repro.mjs` third un-mirrored list | Fixed — substring added, **and** the guard extended to assert the relation |
| **C2** 🟡 three stale "10" counts | Fixed — `vitest.config.ts`, `vitest.config.slow.ts`, `verify-cache.mjs` |
| **N1** `test:server` cache inputs omit the slow config | Fixed — added to `extraFiles`, probed before/after |
| **N2** guard never asserts the files exist | Fixed — `existsSync` check |
| **N3** whole invariant goes vacuous if emptied | Fixed — populated-floor assertion |

Guard is now **7 assertions, 7 passing**, `tsc -p .` clean.

### On C1

Reading `flake-repro.mjs`'s `SLOW` list by dynamic `import()` is not safe: the module calls `process.exit(2)` at eval time when `--file` is absent, which kills the test runner rather than throwing catchably. The guard parses the array literal out of the source instead, and the comment says why.

The assertion is a **match relation, not set-equality** — `flake-repro`'s entries are substrings and one (`generation`) legitimately covers several `SLOW_FILES` paths.

### On N3 — the instructive one

Emptying both arrays together:

```
Tests  1 failed | 6 passed (7)
AssertionError: expected 0 to be greater than or equal to 5
```

**Six of seven assertions passed on a completely empty invariant.** Only the new floor caught it. The floor is deliberately a floor and not `=== 11`: an exact count would fail on every legitimate future addition and would train people to bump a magic number instead of reading why it exists.

### Two C2 sites deliberately not changed

- `docs/superpowers/plans/2026-09-02-test-suite-hardening.md:536` — a quoted `Find:`/`Add:` block recording a search string a completed step matched against.
- `docs/features/archive/45-vitest-pool-tuning.md:43` — says "5 hot files"; already stale long before this PR, and an archived ship-time record.

Both are historical records an edit would falsify rather than correct. Note that same archive doc's item 7 states the mirror invariant in full, including both drift directions — independent confirmation it was real and load-bearing, and had simply never been mechanically guarded.

---

## Why the commit is parked

`test:hooks` fails on a drift guard that has nothing to do with this branch:

```
  AssertionError: pr-review-gate/SKILL.md has drifted from its canonical copy
```

#2999 (`ops-2997`) edited `.claude/skills/pr-review-gate/SKILL.md` and ran `npm run skills:sync`, which writes to `~/.agents/skills/` — **per-machine and shared by every worktree on the box**. Its unmerged edit is now the mirror, so the guard fails everywhere else. This branch's copy is byte-identical to `origin/main` and untouched.

It surfaced only this round because these fixes reach `scripts/**`, which pulls `test:hooks` into pre-commit scope; the first commit's diff didn't.

The guard's own hint — "run `npm run skills:sync`" — is wrong here: doing so from this worktree would clobber #2999's state and invite two live lanes to alternate overwriting a shared file. So the repo owner's call is to **wait for #2999 to merge**, then rebase and commit. Filed as **#3001** (`ops-68`), which names the design decision owed rather than picking one.

Re-review round 2 follows once this lands — C1 was 🔴, so it is required.

dudarenok-maker:
## PR review — pass 2 (head 0c981729, depth high)

Deltas only. Every mutation claimed in the fix round was re-run here rather than read off the receipt; all three reproduce exactly as reported. All five prior findings are resolved.

The new material is in the other direction. Pass 1 said the flake-repro list was un-mirrored; the fix mirrored it **one way** and guarded it **one way**. The reverse direction is broken today for 23 server test files, with the same symptom pass 1 filed C1 for, and the new assertion is structurally unable to see it.

---

| # | Disposition | How it was checked |
|---|---|---|
| **C1** 🔴 | **VERIFIED RESOLVED** | Mutation re-run (see below) — red, and red for the claimed reason |
| **C2** 🟡 | **VERIFIED RESOLVED** | Independent whole-tree sweep; the two "leave as historical" judgments are correct |
| **N1** | **VERIFIED RESOLVED** | Re-probed the live `verify-cache.mjs` module |
| **N2** | **VERIFIED RESOLVED** | Mutation re-run — 1 failed / 6 passed, the existence check |
| **N3** | **VERIFIED RESOLVED** | Mutation re-run — 1 failed / 6 passed, the floor |

Baseline before mutating anything: `npx vitest run src/slow-lane-mirror.guard.test.ts --retry=0` → **7 passed (7)**, 4.32s.

**C1** — dropped `'interim-prune-prohibition'` from `scripts/flake-repro.mjs:18`:

```
× every SLOW_FILES entry is matched by at least one substring in flake-repro.mjs's SLOW list
AssertionError: expected [ Array(1) ] to deeply equal []
+   "src/routes/analysis.interim-prune-prohibition.e2e.test.ts",
Tests  1 failed | 6 passed (7)
```

**N2** — renamed `src/parsers/pdf-real.test.ts` → `pdf-real.pool.test.ts` in **both** lists (so the set-equality and effective-exclude checks stay satisfied, i.e. pass 1's exact scenario):

```
× every SLOW_FILES entry resolves to a file that exists on disk
AssertionError: expected [ 'src/parsers/pdf-real.pool.test.ts' ] to deeply equal []
Tests  1 failed | 6 passed (7)
```

Six of seven green on a list that names a file that does not exist — the check earns its place.

**N3** — emptied `SLOW_FILES` and `SLOW_FILES_TO_EXCLUDE` together:

```
× SLOW_FILES is actually populated (guards against the whole invariant going vacuous)
AssertionError: expected 0 to be greater than or equal to 5
Tests  1 failed | 6 passed (7)
```

Confirms the receipt verbatim.

**N3's floor is the right shape.** The vacuity it names occurs only at length 0, so `> 0` would suffice; 5 is an arbitrary extra tripwire, but it errs closed (a legitimate shrink to 4 fails loudly and gets read, rather than passing silently), and the stated reason for rejecting `=== 11` is correct. Probe answer: a partially-emptied list of 3 does **not** slip through — 3 < 5 fails. Between 5 and 11 the guard is not vacuous anyway; each remaining entry is genuinely asserted. No finding.

**N1** — re-probed `STEPS` + `stepTouchedByDiff` from the live module:

```
slow-config-only     test:server=true   test:server-slow=true     (was test:server=false at e92d0e3d)
main-config-only     test:server=true   test:server-slow=true
guard-test-only      test:server=true   test:server-slow=true
flake-repro-only     test:server=true   test:server-slow=false
verify-cache-only    test:server=true   test:server-slow=false
```

The hole is closed. The `flake-repro-only` row is a bonus the fix round did not claim: the new C1 assertion is also locally reachable on a `flake-repro.mjs`-only diff, via `test:server`'s pre-existing `scripts/**/*.{mjs,cjs,js}` glob. Nothing was needed there.

**Both `verify-cache.mjs` edits are coherent.** `test:server`'s `extraFiles` has one `'server/vitest.config.slow.ts'` entry, no duplicate, comment block intact and correctly appended; the `10 → 11` edit at `:466` is in the *next* STEPS entry (`test:server-slow`) and untouched by the other agent. Neither clobbered the other.

**C2 — the "leave it" judgment holds, and nothing else in the tree states a stale current count.** Swept the whole tree for the count and for `hot file` / `slow file` / `SLOW_FILES` / `slow tier` prose. Every remaining `5`/`10` is a historical record, correctly left:

- `docs/features/archive/45-vitest-pool-tuning.md:43,101,106,110` — "5 hot files", archived ship record;
- `docs/superpowers/plans/2026-06-17-flaky-test-release-hardening.md:770,907` — the `5 → 10` step instructions themselves;
- `docs/superpowers/plans/2026-09-02-test-suite-hardening.md:542` — the literal `Add:` block a completed step inserted;
- `docs/superpowers/specs/2026-07-27-…-dot-segments-design.md:34` and `docs/release-notes-next.md:469` — measurements taken at the time;
- `docs/testing/flake-evidence.md:59` — quotes CLAUDE.md's *old* "5 hot files" string; already stale before this PR, and it is an evidence document.

Also checked, since the 2026-06-17 plan names it as a carrier: `vitest.config.slow.ts`'s enumerated header list is complete — 11 bullets, the new file among them.

---

### 🔴 Correctness bugs

**P1 — `flake-repro.mjs`'s substring matching over-matches. 23 server test files route to the slow config today and exit 1 — the same symptom pass 1 filed C1 for. The new guard checks one direction and cannot see this one.**

`scripts/flake-repro.mjs:16-19`, `server/src/slow-lane-mirror.guard.test.ts:97-101`

The new assertion is `SLOW_FILES.filter((file) => !flakeReproSlow.some((s) => file.includes(s)))` — *every slow file is matched by some substring*. Nothing asserts the converse: *no non-slow file is matched*. Two of the eleven substrings are broad enough to break it right now.

Repro at `0c981729`, on a file this branch never touches:

```
$ node scripts/flake-repro.mjs --file server/src/routes/generation-error.test.ts --runs 1

 RUN  v4.1.11 C:/Claude/Projects/wt-2947-slow-lane/server
No test files found, exiting with code 1

filter: src/routes/generation-error.test.ts
include: src/analyzer/gemini.test.ts, src/routes/analysis-pipelining.test.ts, … (the 11 SLOW_FILES)
run 1: 2635ms exit=1
SUMMARY [{"run":1,"ms":2635,"code":1}]
```

`'generation'` matched, so the harness picked the slow config, whose `include` is exactly `SLOW_FILES` — the positional filter intersects to zero. Same failure text, same exit code, same tool as C1.

Enumerated mechanically (every tracked `*.test.ts(x)` matched by a `SLOW` substring but absent from `SLOW_FILES`):

```
server/src misrouted: 23    by offending substring: { generation: 11, 'book-state': 12 }
other (frontend/e2e) also matched: 14
```

The 14 non-server ones fail differently and worse: `cwd` becomes `.`, and there is no `vitest.config.slow.ts` at the repo root at all, so e.g. `src/views/generation.test.tsx` cannot work either.

Same probe also reports `'generation-boundary-recycle'` is dead weight — `'generation'` already covers it.

This is **pre-existing**, not introduced by this branch, and I want that on the record. It is still this round's, on three counts: the PR edits that exact array; the PR's own deliverable is a guard advertised as closing this drift class, and it closes half of it; and CLAUDE.md voids "it's pre-existing" and "the branch doesn't touch that file" as deferral reasons explicitly. Catalogue shape #10 — one instance fixed, the class left armed.

The fix with the smallest surface that kills both directions permanently: make `SLOW` hold the **full relative paths** (a literal copy of `SLOW_FILES`) and match with `file.endsWith(s)`. The substring semantics — and with it the whole over/under-match question — disappears, and the guard collapses to a set-equality that reads both ways. Whatever shape is chosen, the guard needs the second assertion.

---

### 🟡 Cleanup / chores

**P2 — the guard checks the main config's wiring and not the slow config's. `include: SLOW_FILES` can be replaced outright and all seven assertions stay green while two files run nowhere.**

`server/src/slow-lane-mirror.guard.test.ts:90-95` vs `server/vitest.config.slow.ts:65`

Assertion 4 exists — the PR body says so — because "two constants can agree with each other while the spread that wires them into the config is broken or removed". That reasoning applies symmetrically, and the slow side has no equivalent: `loadSlowConfig()` destructures only `SLOW_FILES`, never `default`.

Mutated `include: SLOW_FILES` → `include: ['src/routes/**/*.test.ts']` and left both constants untouched:

```
$ npx vitest run src/slow-lane-mirror.guard.test.ts --retry=0
 Test Files  1 passed (1)
      Tests  7 passed (7)

$ npx vitest list --filesOnly --config vitest.config.slow.ts | grep -E 'gemini|pdf-real'
(no output — 146 files selected, neither among them)

$ npx vitest list --filesOnly | grep -E 'analyzer/gemini\.test|parsers/pdf-real\.test'
(no output — still excluded from the main suite)
```

`src/analyzer/gemini.test.ts` and `src/parsers/pdf-real.test.ts` are then selected by **neither** config: exercised nowhere, guard green. That is drift direction 2 from this file's own header comment, and it is exactly the mutation-C shape assertion 4 was written to catch. One line closes it — pull `default` out of `loadSlowConfig()` and assert its `test.include` equals `SLOW_FILES`.

**P3 — the `const SLOW` regex is unanchored, and a comment above the declaration makes it silently return the wrong list.**

`server/src/slow-lane-mirror.guard.test.ts:67`

I drove the exact parsing logic over 16 source variants. It fails closed on everything structural — renamed const, `let`, missing semicolon, `new Set([…])`, `.map()` chain, `[...BASE]` (ReferenceError), a line comment inside the array (SyntaxError), even an extra space before `=`. Good. Two variants return a **wrong list with no error**:

```
DOC COMMENT ABOVE showing old list     RETURNS ["old1","old2"]     # /* Was: const SLOW = ['old1','old2']; */
commented-out old declaration above    RETURNS ["old1"]            # // const SLOW = ['old1'];
```

`String.match` without `/g` returns the *first* match, and a comment is just text. The second variant is the realistic one: comment out the old line, add the new one, forget to delete the old. The guard then reads the commented-out list — and if that list was complete when it was commented out, the assertion passes green while the live list is the broken one. C1 re-armed, guard silent. This repo also has a standing habit of long explanatory comment blocks above every config array, which is the first variant.

Anchoring is a one-character fix and I verified it: `/^const SLOW = (\[[\s\S]*?\]);/m` returns the live list in both cases and still handles the multi-line form. (It throws on `export const SLOW` — which the file does not use, and throwing is the correct behaviour there.)

**P4 (nit) — the PR body now contradicts itself on the guard's size.** `## Test plan` still reads "4 tests, all passing (6.47s)"; the round-1 section below reads "7 assertions, 7 passing". Whichever a reader reaches first is the one they believe.

---

### Verified, holds (not re-listed from pass 1)

- **CI is green on this head, and it is the honest gate for it.** All 19 checks pass on `0c981729`, including *Server tests (fast + slow)* (3m33s) and *Lint, typecheck & fast checks*. Both configs are in the main suite's `forceRerunTriggers`, so `--changed` force-reran the server suite rather than selecting past the guard. Given the `--no-verify` commit, I did not take the local gate on trust for anything: the three mutations above were run here, not read.
- **Commit scope.** The PR title `fix(server,docs,scripts)` is accurate and complete for `main...HEAD` — `CLAUDE.md` (docs), `scripts/{flake-repro,verify-cache}.mjs` (scripts), `server/**` (server). Six files, no seventh scope owed. Narrow note, not a finding: commit `0c981729`'s own subject also carries `docs` while its diff touches no docs file; nothing machine-checks per-commit scope against paths, and the PR title is what `pr-title-lint.yml` reads.
- **House gates unchanged.** `Closes #2947` still present, plain text, outside any code span; `Verify PR body links a GitHub issue` passes. Release-notes / regression-plan / on-box N/A claims still hold — the fix round added no user- or operator-visible delta. No `cast.json` writes, no config knob, no derived artifact.
- **Worktree.** `git status --porcelain` empty and `HEAD` still `0c981729` before this comment was posted; every mutation above was reverted with `git checkout --` and the revert verified.

---

### Verdict: **changes requested**

1 correctness bug (**P1**) → re-review after fix, same trigger C1 fired. **P2** and **P3** are both false-negative shapes in the guard this PR ships as its own deliverable, both mutation-proved here, both one-liners; **P4** is a two-word edit to the PR body.

The five prior findings are genuinely closed — the fix round did the work and the receipts are accurate. P1/P2/P3 are all the same omission seen from three sides: the round mirrored and guarded each relation in the direction the review named, and not in the direction it did not.

<sub>Reviewed at `0c981729`; worktree verified unmodified (`git status --porcelain` empty, `HEAD` unmoved) before and after.</sub>

dudarenok-maker:
## PR review — pass 3 (head 1a9c764d, depth high)

Deltas only, and this is the loop cap. Every claimed fix was re-proved here by re-running the mutation, not by reading the receipt — plus two mutations pass 2 did not run. **All four pass-2 findings are resolved.** No correctness bug survives. Four cleanup findings, three of which this round's own edits created.

Because both of the last two commits were `--no-verify`, I did not take any local gate on trust. What I did instead: re-ran every mutation locally, exercised `flake-repro.mjs` end-to-end in all three routing categories, and went to the **cloud CI job logs** rather than the check names — see "Verified, holds".

---

| # | Disposition | How it was checked |
|---|---|---|
| **P1** 🔴 | **VERIFIED RESOLVED** | Both directions mutation-red; routing re-run end-to-end in all three categories |
| **P2** 🟡 | **RESOLVED, with a residual** | Mutation-red as claimed — but the new assertion is itself one-directional (**F2**) |
| **P3** 🟡 | **VERIFIED RESOLVED** | Old and new regex run head-to-head against the same mutated source |
| **P4** | **PARTIALLY RESOLVED** | Section rewritten; the paragraph it replaced was left behind (**F3**) |

Baseline: `npx vitest run src/slow-lane-mirror.guard.test.ts --retry=0` → **8 passed (8)**, 1.64s.

**P1, direction A** — dropped `'src/routes/generation.test.ts'` from `scripts/flake-repro.mjs`'s `SLOW`:

```
× flake-repro.mjs's SLOW list is exactly the same set as SLOW_FILES (both directions)
AssertionError: expected [ 'src/routes/generation.test.ts' ] to deeply equal []
Tests  1 failed | 7 passed (8)
```

**P1, direction B** — added `'src/routes/bogus-not-slow.test.ts'`, absent from `SLOW_FILES`:

```
× flake-repro.mjs's SLOW list is exactly the same set as SLOW_FILES (both directions)
AssertionError: expected [ 'src/routes/bogus-not-slow.test.ts' ] to deeply equal []
Tests  1 failed | 7 passed (8)
```

The over-match class is gone structurally, not by an added assertion, which is the right shape.

**P2** — `include: SLOW_FILES` → `include: ['src/routes/**/*.test.ts']`, both constants untouched:

```
× every SLOW_FILES entry actually appears in the slow config's effective test.include
AssertionError: expected [ 'src/routes/**/*.test.ts' ] to include 'src/analyzer/gemini.test.ts'
Tests  1 failed | 7 passed (8)
```

Green at `0c981729`, red now. Closed.

**P3** — the honest version of this mutation, which the fix round did not run: a commented-out **complete** prior declaration above a **broken** live one. That is the shape that produces a false green, and it needs both halves.

```
live entries before mutation: 11
OLD regex reads 11 entries      <- the commented-out list; guard would pass
NEW regex reads 10 entries      <- the live, broken list
```

```
× flake-repro.mjs's SLOW list is exactly the same set as SLOW_FILES (both directions)
AssertionError: expected [ 'src/routes/generation.test.ts' ] to deeply equal []
Tests  1 failed | 7 passed (8)
```

The anchor is load-bearing and correctly targeted.

**P1, end-to-end routing — all three categories, run for real:**

| Input | Config chosen | Result |
|---|---|---|
| `server/src/routes/generation-error.test.ts` (over-matched at `0c981729`) | **main** | `1 passed`, 19 tests, exit 0 |
| `server/src/routes/analysis.interim-prune-prohibition.e2e.test.ts` (genuinely slow) | **slow** | `1 passed`, exit 0 |
| `src/views/generation.test.tsx` (frontend, over-matched at `0c981729`) | **main, root** | `1 passed`, 80 tests, exit 0 |

The third row is the category pass 2 flagged as "failing worse" (`cwd` → `.` with no root slow config) and did not ask to be fixed explicitly — the exact-match rewrite repaired it too. Frontend `src/routes/` exists but holds none of the eleven names, so there is no cross-tree collision today.

---

### 🟡 Cleanup / chores

All four are cleanup — **no re-review trigger**. F1 and F3 are chores `1a9c764d` itself created, which CLAUDE.md's stale-comment rule makes owed in this round.

**F1 — `1a9c764d` inserted a sixth check above a comment that counts five, and the file header still describes the pre-`1a9c764d` check set.**

`server/src/slow-lane-mirror.guard.test.ts:137`

```
/* All five checks above compare string arrays to each other (or, for the
   effective-exclude check, to a config object) — nothing touches disk. …
```

There are now **six** checks above it — this commit added `every SLOW_FILES entry actually appears in the slow config's effective test.include` at `:122`, directly above. Enumerated:

```
:90  … missing from SLOW_FILES_TO_EXCLUDE
:95  … missing from SLOW_FILES
:100 the two lists are exactly the same set
:104 … main config's effective test.exclude
:122 … slow config's effective test.include     <- added by 1a9c764d
:129 flake-repro.mjs's SLOW list … (both directions)
:145 … resolves to a file that exists on disk   <- the comment above THIS one says "five"
:158 SLOW_FILES is actually populated
```

Same class one level up: the header docblock at `:19-23` still reads *"A third check goes beyond comparing the two constants: it dynamically imports **vitest.config.ts**'s resolved default export…"* — there are two such checks now, and the header names neither the slow-config mirror nor the flake-repro leg. Catalogue shape #9, and the third round in a row this file has been left describing a version of itself that no longer exists.

**F2 — the assertion added for P2 is one-directional, which is the same mistake P1/P2/P3 were. Widening the slow config's `include` double-runs a file in both lanes with all 8 assertions green.**

`server/src/slow-lane-mirror.guard.test.ts:122-127`

`toContain` per `SLOW_FILES` entry asserts *`include` ⊇ `SLOW_FILES`*. Nothing asserts *`include` ⊆ `SLOW_FILES`*. Unlike the main config's `exclude` — which legitimately carries `node_modules/**` and friends, so per-entry `toContain` is the only option there — the slow config's `include` is supposed to be **exactly** `SLOW_FILES`, so the stronger assertion is available and free.

Mutated `include: SLOW_FILES` → `include: [...SLOW_FILES, 'src/routes/generation-error.test.ts']`:

```
$ npx vitest run src/slow-lane-mirror.guard.test.ts --retry=0
 Test Files  1 passed (1)
      Tests  8 passed (8)

$ npx vitest list --filesOnly --config vitest.config.slow.ts | grep generation-error
src/routes/generation-error.test.ts

$ npx vitest list --filesOnly | grep generation-error
src/routes/generation-error.test.ts
```

Selected by **both** configs — that is drift direction 1 verbatim from this file's own header ("it double-runs (present in both the parallel AND serial lanes)"), and the guard cannot see it. It also falsifies the PR body's Test-plan row 3, which claims the effective-`test.include` check covers "the slow config's `include` wiring is broken": it covers replacement, not widening.

One line: `expect(slowConfigDefault.test?.include).toEqual(SLOW_FILES)` in place of the loop.

**F3 — the P4 rewrite prepended a corrected Suite-runs sentence and left the old one immediately below it. The stale half claims the local pre-push gate passed, which is untrue of this head.**

PR body, lines 50–51 — adjacent lines, so they render as one paragraph:

> **Suite runs:** the full server suite passed on `e92d0e3d` (538 files / 8206 tests, 0 failures, 17m) and cloud CI is green on every commit… `force-rerun-triggers.test.ts` green throughout. **Full gate, clean box:** `test:server` **538 files passed, 8206 tests, 0 failures** (17m). … **Pre-push `verify:fast:branch` passed, including both typechecks.**

The three facts are stated twice, and the second copy carries no `e92d0e3d` qualifier — so the final sentence reads as a claim about the branch as it stands. It is not: `0c981729` and `1a9c764d` were both `--no-verify`, as the PR's own note below says. This is P4's exact shape (a body sentence a later commit made false) surviving P4's fix, and it is the one body sentence a merge decision actually leans on.

**F4 (lowest — pre-existing, verified not a regression) — a `--file` argument with Windows separators silently routes to the wrong lane *and* the wrong `cwd`.**

`scripts/flake-repro.mjs:36-38`

```
$ node scripts/flake-repro.mjs --file 'server\src\routes\book-state.test.ts' --runs 1
 RUN  v4.1.9 C:/Claude/Projects/wt-2947-slow-lane          <- repo root, not server/
No test files found, exiting with code 1
filter: server/src/routes/book-state.test.ts
include: src/**/*.{test,spec}.{ts,tsx}, skills/**/*.test.ts
run 1: 1455ms exit=1
```

`file.startsWith('server/')` is false, so `cwd` is the repo root and the frontend config is used; `SLOW.includes(rel)` is then false because `rel` still holds backslashes. A tab-completed path in PowerShell produces exactly this. **I checked whether exact matching made it worse and it did not** — I ran `0c981729`'s version of the script on the same input and it also exits 1, with a worse error (a vite config-resolution stack trace, because it picked `isSlow` and looked for `vitest.config.slow.ts` at the repo root, where none exists). So the new form fails on the same input, more legibly. The root cause is the untouched `cwd` line, not the match; one normalisation (`file.replace(/\\/g, '/')` before both) fixes both. Reported because the PR rewrote this exact block and CLAUDE.md voids "pre-existing" as a deferral reason, not because the fix changed anything here.

---

### Verified, holds

- **The full server suite ran on *this exact head*, in cloud CI.** The dispatch's caution that no full local suite ran on either commit is real, so I went to the job log rather than the green tick. Run `33962555671`, `head_sha: 1a9c764dab8f6d797ebd59ee356fe157ff4dbe20`, job *Server tests (fast + slow)*:
  ```
  Test Files  538 passed | 1 skipped (539)   Duration 120.04s     (main config)
  Test Files  11 passed (11)                 Duration  32.56s     (vitest.config.slow.ts)
  ```
  `--changed` force-reran everything, as designed — the diff touches `scripts/**` and `server/**`, both `forceRerunTriggers`. This is the honest gate for the bypassed commits and it is complete.
- **Typecheck, lint and hooks all actually ran on this head** — job `101296903590` step conclusions: `Lint success`, `Typecheck success`, `Hooks tests success`. So `tsc -p .` cleanliness is verified, not asserted, and #3001 confirmedly does not reach CI. All 19 checks green.
- **Guard + `force-rerun-triggers.test.ts` locally: 77 passed (2 files), 3.57s.** Matches the commit message's claim exactly.
- **No assertion in the current set can pass vacuously except through the empty-list door the floor already closes.** Walked all eight: 1/2/4/5/7 go vacuous only at `SLOW_FILES.length === 0`; 3 is redundant with 1+2 (pass 1's N3 note, left in — harmless, not masking); 6 stays live even on an empty `SLOW_FILES` because `extraInFlakeRepro` would then hold all eleven; 8 is the floor. Both effective-config checks fail **closed** if `test` or its key is absent (`?? []` → every `toContain` fails), which the P2 mutation demonstrated rather than assumed. The one real gap is F2's direction, not a vacuity.
- **Commit scope.** `1a9c764d` is `fix(scripts,server)`; its diff is exactly `scripts/flake-repro.mjs` + `server/src/slow-lane-mirror.guard.test.ts`. Accurate and complete. PR title `fix(server,docs,scripts)` still matches `main...HEAD` (6 files).
- **House gates unchanged and still good.** `Closes #2947` present, plain text, outside any code span; *Verify PR body links a GitHub issue* passes. Regression-plan / on-box / release-notes N/A claims all still hold — this round added no user- or operator-visible delta. No `cast.json` writes, no config knob, no derived artifact.
- **Accuracy note on one claim I was asked to adjudicate, not a finding:** *"Retires the dead `'generation-boundary-recycle'` entry"* (commit message and round-2 comment) reads as though the list lost an entry. It did not — `src/routes/generation-boundary-recycle.test.ts` is still there and is now **load-bearing** (drop it and both the guard's set equality and the file's routing break). What was retired is the redundant *substring*; the list is still eleven entries. Harmless as written, misleading if someone acts on it later.
- **No third routing category broke**, and the eleven-entry lists in `vitest.config.slow.ts`, `vitest.config.ts` and `flake-repro.mjs` are byte-identical sets — asserted by the guard and re-confirmed by the two `vitest list` runs above.

---

### Verdict: **mergeable as-is**

This is the loop cap, so the question is binary. **No 🔴 survives.** P1's over-match is fixed structurally rather than patched, both directions are mutation-red, and the routing is correct in all three categories including the frontend one nobody was asked to fix. P2 and P3 are genuinely closed, and P3's fix is load-bearing under the stricter two-part mutation the fix round did not run. The full server suite, typecheck and hooks all executed against this exact SHA in CI, so the `--no-verify` commits carry no unverified surface.

F1–F4 are cleanup only and none of them can produce a wrong result today. Two of them (F1, F3) are stale text `1a9c764d` itself created, which CLAUDE.md's same-round rule makes owed here rather than later, and both are a one-word / one-paragraph edit. F2 is one line and closes the last direction of the relation this whole PR has been about. F4 is pre-existing and demonstrably not a regression. If the owner wants this merged now, none of the four is a reason to hold it; if they are folded, they need no fourth review pass — cleanup-only findings do not re-trigger the gate.

<sub>Reviewed at `1a9c764d`; five mutations applied and reverted, each revert verified. `git status --porcelain` empty and `HEAD` unmoved (`1a9c764dab8f6d797ebd59ee356fe157ff4dbe20`) before and after.</sub>

Issue #2997, opened 2026-09-05
The local commit gate deadlocks under concurrent agent work — commits take hours, never complete


dudarenok-maker:
## What

The local commit gate has stopped functioning as a gate. On a loaded box a `git commit` no
longer completes in minutes — measured today at **up to 4h34m**, and in the steady state it
does not complete at all. Concurrent batteries deadlock against each other and never exit,
so the pile grows monotonically until the box is manually reaped.

This is a **systems failure in the gate's design**, not a test failure. It is filed as one
ticket because the individual symptoms have already been filed five times as separate test
bugs (see "Why this is not five test bugs" below) and the class keeps recurring.

## Evidence (2026-09-05, primary checkout, measured not estimated)

Live process census before intervention:

| Metric | Value |
|---|---|
| `verify-cache` batteries running concurrently | **11** |
| Oldest battery age | **273.8 min (4h34m)** |
| vitest processes | 30 |
| `node.exe` total | 86 |
| Orphaned vitest (parent dead) | 2, aged **392m and 389m** |
| Worktrees | 19 |

Per-battery subtree CPU — the honest "is it progressing" measure. A live vitest subtree burns
**30–100 CPU-seconds per minute**:

| PID | Age | Procs | Subtree CPU | CPU/min |
|---|---|---|---|---|
| 47184 | 273.8m | 8 | 37.9s | **0.1** |
| 50836 | 216.7m | 8 | 53.0s | **0.2** |
| 25500 | 183.9m | 12 | 14.3s | **0.1** |
| 24932 | 181.5m | 9 | 40.4s | **0.2** |
| 46848 | 156.1m | 9 | 27.3s | **0.2** |
| 37016 | 130.3m | 10 | 28.8s | **0.2** |
| 46604 | 127.2m | 8 | 27.5s | **0.2** |
| 20328 | 119.1m | 8 | 23.1s | **0.2** |
| 50148 | 108.5m | 9 | 37.1s | **0.3** |
| 12120 | 105.0m | 8 | 17.3s | **0.2** |
| 41696 | 90.4m | 9 | 28.4s | **0.3** |

**11 of 11 non-progressing.** Collectively 335 CPU-seconds consumed across ~24 hours of
cumulative wall-clock. Reaping the 13 roots freed 96 processes and took `node.exe` from 86 → 9.

## Root cause chain

1. **76 open `[agent instructions]` tickets** feed concurrent agent sessions, each of which
   eventually commits.
[...]

Issue #2226, opened 2026-08-07
test — book-state-preserve-voices #1981 flakes under box contention; not in the flaky register


dudarenok-maker:
## What

`server/src/routes/book-state-preserve-voices.test.ts` → **"#1981 — a stale cast PUT does not erase a concurrently /assign-planted voice"** fails intermittently in a full `test:server` run and passes in isolation.

Observed 2026-08-07 while warming the verify cache on an unrelated branch (#2207, whose diff touches only `server/src/config/resolver.ts` and `server/src/tts/spawn-sidecar.ts` — nothing in `routes/`):

```
❯ src/routes/book-state-preserve-voices.test.ts (7 tests | 1 failed) 2948ms
    × #1981 — a stale cast PUT does not erase a concurrently /assign-planted voice 78ms (retry x1)
    Tests  1 failed | 6741 passed | 8 skipped (6750)
[fail] test:server (exit 1, took 746.6s)
```

Same tree, same commit, run alone:

```
Test Files  1 passed (1)
     Tests  7 passed (7)
     Duration  8.43s
```

The box had a concurrent session running four other worktrees' suites at the time — the documented contention profile (see the 746s wall-clock against a ~280s clean-box baseline).

## Why this needs handling rather than ignoring

It is **not** in `docs/testing/flaky-register.md` and **not** wrapped in `quarantinedIt`, so today it fails a gating lane with no record that it is known-flaky. That is the worst of both: it blocks a commit, and the next person to hit it has nothing to check against and will spend the triage time again — as I just did.

CLAUDE.md's own rule is that a flaky test routes through `quarantinedIt` (`server/src/test-utils/quarantine.ts`) into the non-gating `npm run test:quarantine` lane and is logged in the register. That has not happened here.

## What needs deciding

The test is a **concurrency** test — a stale cast PUT racing an `/assign`-planted voice — so it is inherently timing-sensitive, and it is guarding a real data-loss invariant (#1981). Quarantining it removes gating protection from something worth protecting.

1. **Diagnose and stabilise it** — find the actual race in the *test* (fs timing, fixture setup) and make it deterministic. Best outcome; unknown cost.
2. **Quarantine + register**, per CLAUDE.md's stated process — honest and cheap, but moves a real invariant into the non-gating lane.
3. **Register-only, no quarantine** — record it as known-flaky so the next triage is instant, while it keeps gating. Least protective of developer time, most protective of the invariant.

[...]

PR #2999, opened 2026-09-05
fix(ops): replace the ~35-minute hook battery with instant local checks

dudarenok-maker:
## Summary

The local commit gate stopped being usable. Two separate failures, both measured:

1. **It was never viable, even healthy.** `.verify-cache.json` last-green durations put a cold
   `verify:fast:branch` at **~35 minutes** (`test:server` alone 19.45 min); a full `npm run
   verify` floors at **44.9 min**.
2. **Under concurrency it deadlocked.** Census 2026-09-05: **11 batteries at once, oldest
   273.8 min (4h34m), 0 of 11 making progress** (0.1–0.3 CPU-s/min against 30–100 for a live
   vitest subtree), plus 2 orphaned vitest aged ~390 min. Reaping freed 96 processes.

CI is already the enforcing gate — `npm run verify` is a required status check on `main` and
runs in 1–6 minutes. The hooks were a 10× slower duplicate. This PR cuts them back to match.

### What changes

- **`.husky/pre-commit`** → `scripts/hooks/pre-commit-lint.mjs`: ESLint over **staged files
  only**, one process. Empty staged set exits before spawning anything.
- **`.husky/pre-push`** → the two guards + docs-only early-exit + **`test:sidecar` alone**.
- **`scripts/verify-cache.mjs`** — skips the GPU/sibling contention probe when no
  `LOW_CONCURRENCY`-affected step is selected. That probe was ~4.6 s of a 4.9 s pre-push, and
  its only output is read solely by the two vitest configs, which this invocation never runs.
- **Invariant, now enforced:** *no git hook may spawn a process pool.*

`test:sidecar` is kept deliberately: scope-gated to `server/tts-sidecar/**` **and** input-hashed,
so it costs ~6.85 min on a sidecar-touching push and near-zero otherwise — and it spawns pytest,
not a vitest fork pool. It is the only automatic executor of 38 `pytest.importorskip("torch")`
tests CI cannot run (`verify.yml:396-400` installs the lean sidecar set by design).

### Measured result

| Path | Before | After |
|---|---|---|
| commit, no JS/TS staged | ~35 min (hours when wedged) | **~8 s** (hook portion 0.14 s) |
| commit, JS/TS staged | ~35 min | **~8.5 s** |
| push, sidecar untouched | ~35 min | **3.9 s** |

One push ran while **6 sibling batteries were active** — the condition that used to produce
4-hour deadlocks — and was unaffected.

### Reviewer-driven changes (2 passes, 14 findings, all fixed)

Pass 1 found 2 blocking bugs in the first implementation, and both were **conditional paths the
happy path never touched**:

- a worktree without `node_modules` blocked *every* commit, because `shell: true` launches
  `cmd.exe`, which always exists, making the `ENOENT` branch unreachable;
- staging `src/lib/api-types.ts` blocked the commit *and* deadlocked against CI, which requires
  that generated file be committed.

The fix was reworked around **positive evidence that ESLint ran**: an `existsSync` pre-check,
`--format json`, and passing on any spawn error, exit code 2, or unparseable output. Blocking
now keys on `result.status === 1`. Per-batch classification replaced raw stdout aggregation,
which had concatenated JSON arrays into unparseable output (silently disabling lint above 100
staged files) and let a later batch's fatal error suppress an earlier batch's real findings.

The invariant guard hardcoded 3 hook filenames while `.husky/_` ships wrappers for 16; it now
enumerates `.husky/` dynamically. Mutation-verified: a new `.husky/pre-merge-commit` containing
`npm run verify` fails it by name.

### Why the instruction files are in this PR

Agents *execute* these. `.claude/skills/pr-review-gate/SKILL.md` gave the gate's precondition as
"`verify:fast:branch` green"; `CLAUDE.md` step 7 said "(same battery as pre-push)";
`.github/copilot-instructions.md` claimed "a commit here takes 5–7 minutes". Left stale, an agent
either pays 35 minutes by hand — the exact cost this deletes, relocated into every turn — or
silently redefines "verified". That is a correctness defect, not a docs chore.

## Test plan

- **`scripts/tests/hook-no-pool.test.mjs`** — the invariant, as an **allowlist** (a denylist
  misses `npm test`, `npm run verify`, `playwright`, `pytest`). Enumerates every file under
  `.husky/`. Mutation-verified in both directions.
- **`scripts/tests/pre-commit-lint.test.mjs`** — fixtures are **real captured ESLint output**,
  not hand-written shapes. Three earlier rounds of tests passed while the behaviour was wrong,
  each time because a fixture omitted what the real tool produces.
- **`scripts/tests/verify-cache.test.mjs`** — the contention-skip selection, plus the negative
  pin that a `.husky/**` diff selects `test:hooks` and nothing else.
- Behaviourally verified against the real binary: clean file → not blocked; eslint-ignored
  tracked file → not blocked; **real lint errors → blocked**; missing eslint → not blocked.

## Known caveat

`npm run test:sidecar` **SKIPs and exits 0** without `server/tts-sidecar/.venv`
(`scripts/run-sidecar-tests.mjs:37-47`), which CLAUDE.md tells you to junction only "for real
sidecar/TTS work". So the retained coverage this PR calls its one closed unbounded hole is
silently absent in an unprepared worktree.

## Merge-time step

**Run `npm run skills:sync` from `main` after merging.** This PR edits
`.claude/skills/pr-review-gate/SKILL.md`, and that mirror is machine-shared. Syncing it
*pre*-merge poisoned `~/.agents/skills/` and blocked PR #2998 — see #3001; the mirror has been
restored to `main` in the meantime, so this branch currently shows the drift, which is correct.

## Design

`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md`, after four adversarial
passes. Deferred with reasons recorded: the concurrency governor (config-level admission would
break `vitest.config.wire-fixtures.ts`, which is spawned *from inside running tests*), and
deleting the GPU throttle (the probe never fired — 0% against a threshold of 40).

Refs #2997


dudarenok-maker:
## PR review — pass 1 (head 4227165c, depth medium)

Twelve findings: **2 🔴 blocking, 4 🟠 significant** (correctness), 6 🟡 cleanup. Every repro below was executed against this worktree, not reasoned about.

---

### 🔴 1 — A worktree without `node_modules` BLOCKS every commit on Windows. The `ENOENT` branch is unreachable there, and its test proves a shape the code never produces.

**Correctness bug.** `scripts/hooks/pre-commit-lint.mjs:96-102` (spawn), `:64-66` (the branch), `:57-59` (the docstring), `:18-22` (the header claim).

The header states the load-bearing guarantee: *"a worktree without `node_modules` — a normal state here, see CLAUDE.md's worktree-setup checklist — must not block a commit."* `:99` sets `shell: isWin`. Under `shell: true`, `spawnSync` launches **`cmd.exe`**, which always exists — so a missing `eslint.cmd` never surfaces as `error.code === 'ENOENT'`. It surfaces as an ordinary nonzero exit from cmd, which falls through to `:67` and blocks.

Measured on this box (real `spawnSync`, missing `node_modules`, then the real `classifyLintResult`):

```
{ "status": 1, "signal": null, "errorCode": null,
  "stdout": "", "stderr": "The system cannot find the path specified.\r\n" }
VERDICT: {"blocked":true,"reason":"The system cannot find the path specified.\r\n"}
```

→ `process.exit(1)`. The commit is refused with a one-line cmd error and zero lint findings.

This is not a hypothetical environment. CLAUDE.md → *Worktree setup*: *"`EnterWorktree` and the Agent tool's `isolation: \"worktree\"` create the tree and nothing else — no dependencies"*, and the fix ordering there is `npx husky` **first** (step 1), junction `node_modules` **second** (step 2). Any agent worktree that gets step 1 and not step 2 — the documented default state — now has a hook that refuses every commit staging a `.ts`/`.js` file. The escape hatch a blocked agent reaches for is `--no-verify`, which CLAUDE.md forbids.

**Why the paired test doesn't catch it** — reviewer-brief shapes #3 and #11. `scripts/tests/pre-commit-lint.test.mjs:59-64` hand-authors `{ status: null, signal: null, error: { code: 'ENOENT', … } }`. That is the **POSIX** result shape (`shell: false`). Windows produces `{ status: 1, error: undefined }`. The fixture is written from belief about spawnSync's contract rather than from a recorded Windows run, so the test is green on a path the code never takes. The docstring at `:57-59` asserts the same invented contract in prose.

**Fix direction (one change closes this and finding 4):** drop `shell` entirely and spawn `process.execPath` on `node_modules/eslint/bin/eslint.js`. A missing file then yields a real `ENOENT`, there is no cmd length ceiling, hand-quoting disappears, and so does the `DEP0190` deprecation warning Node prints to stderr on *every* Windows commit under the current call:

```
(node:45480) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true …
```

A cheap belt-and-braces alternative is an `existsSync(eslintBin)` check before `:96` returning the same warn-and-pass verdict.

---

### 🔴 2 — Staging `src/lib/api-types.ts` blocks the commit, and deadlocks against a required CI check.

**Correctness bug.** `scripts/hooks/pre-commit-lint.mjs:91` (`--max-warnings 0` with explicit file arguments).

`eslint.config.mjs:32-51` puts `src/lib/api-types.ts` in a global-ignores block. `npm run lint` (`eslint . --max-warnings 0`) never sees it — directory traversal skips ignored files silently. Passing it **explicitly on the CLI**, which is exactly what this hook does, makes ESLint 9 emit a warning instead; `--max-warnings 0` turns that warning into exit 1.

Measured, ESLint 9.39.4, via the hook's own call shape:

```
status: 1
stdout: "…src\\lib\\api-types.ts
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" …
✖ 1 problem (0 errors, 1 warning)"
stderr: "ESLint found too many warnings (maximum: 0)."
VERDICT: {"blocked":true, …}
```

`src/lib/api-types.ts` is tracked, generated, and **must** be regenerated and committed whenever `openapi.yaml` changes — CLAUDE.md lists it under derived artifacts, and `.github/workflows/verify.yml:198-203` fails the PR with `"src/lib/api-types.ts is stale. Run 'npm run openapi:types' and commit the result."` if you don't. So after this PR: CI requires the commit, and pre-commit refuses to make it. The only way through is `--no-verify`.

**Fix direction:** add `--no-warn-ignored`, or filter the staged set through `ESLint#isPathIgnored` before spawning. Either needs a regression test that stages a globally-ignored path — the current suite has none, because `filterLintableFiles` only knows about extensions, not about `eslint.config.mjs`'s ignores.

---

### 🟠 3 — `hook-no-pool.test.mjs` enforces the invariant for 3 hardcoded filenames. `.husky/_` already ships wrappers for 16.

**Correctness bug in the PR's own load-bearing guard.** `scripts/tests/hook-no-pool.test.mjs:61`.

The header rightly argues an allowlist over a denylist for *line content* — and then enumerates the *files*: `const HOOKS = ['commit-msg', 'pre-commit', 'pre-push'];`. Nothing reads the directory. `.husky/_/` in this worktree contains husky wrappers for `applypatch-msg, commit-msg, post-applypatch, post-checkout, post-commit, post-merge, post-rewrite, pre-applypatch, pre-auto-gc, pre-commit, pre-merge-commit, prepare-commit-msg, pre-push, pre-rebase` — every one of which git will execute if the matching `.husky/<name>` user hook appears.

Proven, without touching the worktree: I copied the three real hooks into a scratch dir, added `.husky/pre-merge-commit` = `npm run verify` and `.husky/post-checkout` = `npm run test:server`, and ran the guard with that as cwd:

```
✔ .husky/commit-msg contains only allowlisted invocations
✔ .husky/pre-commit contains only allowlisted invocations
✔ .husky/pre-push contains only allowlisted invocations
… ℹ tests 10  ℹ pass 10  ℹ fail 0
```

Two pool-spawning git hooks, guard fully green. `pre-merge-commit` is not academic here — CLAUDE.md's reconciliation pattern has you merging agent branches locally on an `integration/<date>` branch, which fires it.

The mutation check reported in the PR body (adding `npm run test:server` to `.husky/pre-commit`) only ever exercised a filename already in the list, so it could not have surfaced this.

**Fix direction:** derive `HOOKS` from `readdirSync('.husky')` filtered to files (excluding `_`), and assert the set is non-empty so an unreadable directory fails closed rather than iterating zero hooks.

Related, same file: `findDisallowedHookLines('')` returns `[]`, so an **emptied** hook body passes. That is consistent with the stated invariant ("no pool"), but nothing anywhere now pins that `.husky/pre-push` still *contains* `guard-protected-push`/`guard-commit-subjects` — I checked `scripts/tests/*.test.mjs` and no test asserts hook file contents. Worth one positive assertion alongside the negative one.

---

### 🟠 4 — ~180+ staged lintable files blocks the commit with "The command line is too long."

**Correctness bug.** `scripts/hooks/pre-commit-lint.mjs:91-102`.

`shell: true` routes through `cmd.exe`, whose command line is capped at 8191 characters. The hook passes every staged path as its own quoted argument with no batching. Measured, using real tracked `server/src/**/*.ts` paths:

```
n=100 approxCmdLen=5303  status=1 stderr="The system cannot find the path specified."
n=150 approxCmdLen=7501  status=1 stderr="The system cannot find the path specified."
n=200 approxCmdLen=9868  status=1 stderr="The command line is too long."
n=250 approxCmdLen=11773 status=1 stderr="The command line is too long."
```

At the crossover, `classifyLintResult` sees `status: 1` with no `error` and returns `{ blocked: true }`. A codemod, a lint-rule rollout, or a formatting sweep across `src/**` or `server/src/**` (935 tracked `.ts` files in `server/src` alone) reaches that count easily, and the commit is refused for a reason that has nothing to do with lint.

Same root cause as finding 1: `classifyLintResult` has no positive evidence that ESLint ran, so it reads *every* nonzero exit as a lint finding. Spawning `node node_modules/eslint/bin/eslint.js` (no shell) raises the ceiling to CreateProcess's 32767 and removes the failure class from the hot path.

---

### 🟠 5 — `.clinerules/cline.md` is the fourth agent-instruction file, and it was not updated.

**Correctness bug by this PR's own definition.** `.clinerules/cline.md:129`, `:144`, `:160-163`, `:97`.

The PR body argues, correctly: *"Agents execute these instructions… Left stale, an agent either pays 35 minutes by hand… or silently redefines 'verified'. That is a correctness defect, not a docs chore."* It then names three files and misses the one whose entire premise this change deletes. The Cline lanes are live (`oe-route` routes work to `cline`, `cline-free`, `cline-glm`, `cline-qwen-*`, `cline-muse-local`), and they read this file.

Now-false claims still in it at HEAD:

- `:129` — *"Your runtime kills a single command at 30 seconds. **A pre-commit run that has any test step in scope takes minutes.** So a foreground `git commit` cannot finish, and retrying it never will."* Pre-commit is now one ESLint process — the PR's own table says 0.14s / ~4s.
- `:144` — *"pre-commit runs `verify:fast:scoped` over `test:hooks,check:budget-poll,test,test:server`"* — false; that is the exact sentence `.husky/pre-commit` no longer supports.
- `:160-163` — *"Budget 10-15 minutes when a test step is in scope, and do not call it hung before 20."*
- `:97` — *"the whole pre-push battery vanish[es]"* — there is no pre-push battery any more.
- The whole **"Committing when a step is in scope"** section (detached `Start-Process` + poll via `scripts/oe-detached-commit.ps1`) is now ceremony every Cline lane pays for a sub-second hook. Note `.github/copilot-instructions.md` had its *pointer* to that section removed in this diff while the section itself stayed — so a Cline agent still follows it.

---

### 🟠 6 — The design of record contradicts the design that shipped, and the contradiction is Part 6's entire premise.

**Correctness bug in the shipped plan doc.** `docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md:295` vs `:143`.

- `:143` (Part 1's own table, describing what shipped): `pre-push` → *"the guards + `test:sidecar`, scope-gated"*. Matches `.husky/pre-push:30`.
- `:295` (Part 6, deferred): *"**Part 1 deletes the automatic `test:sidecar` push trigger that exercised them.**"*

Both sentences are in the file this PR adds. `:295` is a survivor from an earlier draft in which Part 1 dropped `test:sidecar` entirely, and Part 6's whole deliverable — a required check failing any PR touching `server/tts-sidecar/**` unless the body records a local run — is justified by it. A cold agent briefed from this doc (CLAUDE.md's execution model says the implementation thread briefs from the plan doc) would build a compensating gate for a trigger that still exists.

---

### 🟠 7 — CLAUDE.md still carries three claims this change made false, in sections it did not touch.

**Correctness bug** in the instruction file the PR edited. Same class as finding 5.

- `CLAUDE.md:339` — *"`npm run test:changed` / `npm run test:server:changed` … Not run standalone in practice — **these back pre-commit's `--scope-staged` narrowing (see "Commit gate")** — but usable directly."* Nothing invokes `--scope-staged` from a hook any more, so they back nothing; and the cross-reference points at the section this PR rewrote to say the opposite. (`verify:fast:scoped` and `verify-cache.mjs`'s `CHANGED_ONLY_NPM_SCRIPT`/`diffSafeForChangedOnly` machinery are likewise now manual-only, with no automatic caller.)
- `CLAUDE.md:340` — *"not in pre-push `verify:fast:branch` or **`verify:fast` pre-commit**"*.
- `CLAUDE.md:1210` — *"`verify:fast:branch`/pre-push **can legitimately run 15–45+ minutes under load**; the fix for 'it's taking a while' is to wait, never to fire a second attempt."* This is load-bearing agent instruction inside *One writer per worktree*, and the number it hangs on is the one this PR deleted (PR table: push, sidecar untouched = 19s).

---

### 🟡 8 — Two stale hook comments corrected; four equivalents left, one of them in `verify.yml`.

**Cleanup.** The PR fixes the `.husky/**` comment in `scripts/verify-cache.mjs:138-144` and `scripts/tests/verify-cache.test.mjs:636-648` — both new comments read true, I checked them against `STEPS[]` and the assertion below them. Four siblings in the same class were left:

- `.github/workflows/verify.yml:16-17` — *"Local pre-push now only runs a fast, branch-diff-scoped subset (`verify:fast:branch`)"* — in the required status check's own header.
- `scripts/tests/verify-cache.test.mjs:1421` — *"when this suite runs inside a real git hook (pre-commit calls `npm run verify:fast:scoped` -> `npm run test:hooks`)"* — in the very file this PR edited to fix two others.
- `scripts/tests/check-onbox-register.test.mjs:2489-2491` — *"`test:hooks`, wired into pre-commit/pre-push via `verify:fast:scoped`/`verify:fast:branch`) require network access, so any commit touching the on-box register could no longer be made offline"* — the rationale for a whole design decision, now resting on wiring that no longer exists.
- `scripts/lib/module-graph.mjs:131` — *"verify:fast:scoped, a path documented as sub-5s"*.

---

### 🟡 9 — "zero on every other push" is 4.9s, and ~4.6s of it is a probe that cannot affect the only step this hook can run.

**Cleanup.** `.husky/pre-push:22-24`; also `CLAUDE.md:921` and the copilot-instructions bullet repeat "zero cost on most pushes".

Measured on this branch (no `server/tts-sidecar/**` in the diff, i.e. the claimed-zero case):

```
node scripts/verify-cache.mjs --steps test:sidecar --scope-branch
[contention] 8 sibling vitest/verify-cache process(es) already running …
[skip] test:sidecar (out of scope)
exit=0 elapsed_ms=4904

SKIP_CONTENTION_CHECK=1 … [skip] test:sidecar (out of scope)   elapsed_ms=299
```

`runPipeline` runs `detectGpuContention()` (`nvidia-smi`, 5s timeout) and `detectSiblingContention()` (`powershell -NoProfile Get-CimInstance Win32_Process`, 5s timeout) at `verify-cache.mjs:1344-1364`, **before** the scope filter. Their only product is `LOW_CONCURRENCY`, which is read solely by `vitest.config.ts`, `server/vitest.config.ts` and `scripts/test-concurrency.mjs` — none of which the one permitted step (`test:sidecar` → pytest) touches. So on the hook's dominant path the probe is a pure ~4.6s tax that can change nothing. The PR's own table says 19s for this push, which already contradicts the comment's "zero".

Either state the real figure, or pass `SKIP_CONTENTION_CHECK=1` in the hook invocation (which then has to stay off the `hook-no-pool` allowlist's pinned `--steps` line, so the guard needs the same edit).

---

### 🟡 10 — The script cites Principle 4 as authority for doing the opposite of Principle 4.

**Cleanup.** `scripts/hooks/pre-commit-lint.mjs:18-22`.

```
// Principle 4 (fail on findings; pass on a missing tool; FAIL on a budget
// breach): … A 60s local budget that's exceeded is treated the same way:
// warn to stderr and pass, never hang the commit.
```

The design's Principle 4 (`…design.md:128-131`) says *"A timeout is the opposite: a timeout that passes is not a cap. (An earlier draft collapsed these into one clause that told timeouts to pass — the inverse of Part 2.)"* The **behaviour** is fine — Part 1 at `:161` explicitly specifies *"warns and passes when eslint is absent or a 60 s budget is exceeded"* — but the comment quotes the general principle and then contradicts it in the next sentence, which is precisely the trap that draft correction was written to prevent. Cite `:161`, not Principle 4.

---

### 🟡 11 — The design says `is-docs-only-push.mjs`'s invocation goes; the shipped hook keeps it.

**Cleanup.** `…design.md:172` vs `.husky/pre-push:27-30`.

*"`is-docs-only-push.mjs` is **retained**; only its invocation goes (`.husky/pre-push:25-27` exists solely to guard line 28)."* The hook keeps the invocation, and `CONTRIBUTING.md`/`CLAUDE.md` were both edited in this diff to *describe* it as retained. Keeping it is the right call (it saves the 4.9s of finding 9 on a docs-only push); the doc sentence is the stale half. It also carries a line-number citation that is already off by three in the shipped file.

---

### 🟡 12 — Release-notes pair: half landed, no not-applicable declared.

**Cleanup.** `docs/release-notes-next.md:493` added; `RELEASE_NOTES.md` untouched (in-progress section `# Castwright 1.15.0` is live and already carries a bullet for the *previous* pre-commit contention work).

CLAUDE.md before-shipping step 5 requires both, *"or say so explicitly rather than silently omitting"*. The PR body doesn't. Adding the technical-register entry while omitting the user/operator-facing one is the shape the step exists to prevent — pick one and state it.

---

## Checked and clean

- `Refs #2997` present in the PR body, outside any code span.
- `scrubGitEnv()` (not the broader helper) on the `git diff --cached` call at `:77-81` — correct: `GIT_INDEX_FILE` is deliberately preserved, so the hook reads the in-flight index for `git commit -a` / `git commit -- <path>`.
- `windowsHide: true` on both `spawnSync` calls, per the `spawn-windows-hide` guard.
- `--max-warnings 0` matches `npm run lint`, so no strictness asymmetry with CI (the asymmetry in finding 2 comes from explicit-file mode, not from the flag).
- `--diff-filter=ACMR` plus the `existsSync` filter correctly excludes deletions.
- CI coverage of the retired `verify:fast:branch` legs is complete — `lint`, `typecheck`, `config:check`, `test:hooks`, `check:budget-poll`, `audit`, `audit:server`, `test`, `test:server`, `build`, `test:sidecar` all have jobs in `verify.yml`. The coverage-honesty argument holds for everything except the torch tier, which is the stated reason `test:sidecar` stays.
- Both new test files are picked up by `run-hooks-tests.mjs`'s `scripts/tests/*.test.mjs` glob, and `scripts/hooks/**` is inside `test:hooks`' `scripts/**/*.{mjs,…}` input glob, so the cache invalidates on a change to the new script.
- The new negative pin (`a .husky diff matches test:hooks and NOTHING ELSE`) is a real assertion — I confirmed `computeShared` does not cover `.husky/**`, so it can actually fail.
- `~/.agents/skills/pr-review-gate/SKILL.md` is in sync with this branch's edit (differs only by the injected `MIRRORED COPY` banner), so `skills:sync` was run.

## Not applicable

- On-box acceptance: no hardware-provable behaviour ships here. One caveat worth a sentence in the PR body rather than a register row — `npm run test:sidecar` SKIPs and exits 0 when `server/tts-sidecar/.venv` is absent (`scripts/run-sidecar-tests.mjs:37-47`), and CLAUDE.md's worktree checklist junctions `.venv` only *"for real sidecar/TTS work"*. So the retained coverage this PR calls "the one unbounded hole closed" is silently absent in an unprepared worktree.
- `docs/features/INDEX.md`: design specs under `docs/superpowers/specs/` are not indexed there.

---

**Re-review trigger:** findings 1-7 are correctness bugs, so fixing and pushing them re-triggers a pass.

dudarenok-maker:
## PR review — pass 2 (head aef6e2d7, depth medium)

Deltas only. Everything not named below was re-checked and is solid.

Verification actually run this pass (not inferred): eslint 9.39.4 driven from the repo root with the hook's own `baseArgs` against a clean file, `src/lib/api-types.ts`, a non-existent path, a 100-file batch, and a dirty `--stdin` payload; `node --test` over the two new files (30/30) and `review-gate-mechanism` + `verify-cache` (156/156); a throwaway repo pair to settle whether `git push --dry-run` fires `pre-push` (it does, git 2.54.0.windows.1); `grep`/`sed` against every doc line the change edited or staled.

---

### Pass-1 dispositions

**1. 🔴 Missing eslint blocks every commit — VERIFIED RESOLVED.** Two independent barriers now: the `existsSync(eslintJs)` pre-check (`scripts/hooks/pre-commit-lint.mjs:186-190`), and `classifyLintResult` failing open on *any* `result.error`, on `status === 2`, on unparseable stdout, and on JSON that is not an array (`:59-114`). The specific hole the first fix attempt left — `node.exe` always exists, so `MODULE_NOT_FOUND` arrived as `status 1` + non-JSON stdout — is closed by the parse branch at `:79-84` and pinned by a test with the real Node error text. No equivalent hole remains: I could not construct an input where "eslint absent/broken" reaches `{ blocked: true }`.

**2. 🔴 Staging `src/lib/api-types.ts` blocks — VERIFIED RESOLVED**, measured live rather than taken on trust:
`node node_modules/eslint/bin/eslint.js --format json --max-warnings 0 --no-warn-ignored src/lib/api-types.ts` → stdout `[]\n`, status `0` → `classifyLintResult` → `{ blocked: false }`.

**3. 🟠 Guard covered 3 hardcoded filenames — VERIFIED RESOLVED.** `enumerateHookFiles` (`scripts/tests/hook-no-pool.test.mjs:62-73`) generates one test per file directly under `.husky/`; the run emits three (`commit-msg`, `pre-commit`, `pre-push`). `.husky/_` is excluded as a directory, correctly — husky's 16 wrappers only source the user hook, so a pool can only ever be *added* at `.husky/<name>`. Residual nit in **N-7** below.

**4. 🟠 argv limit — VERIFIED RESOLVED**, and the follow-up rework is right. Per-batch classification means a warn-batch can no longer swallow a block-batch, and precedence is order-independent (`combineBatchVerdicts:145-158`). I probed the four shapes you asked about: warn+block → **block** ✓; all-warn → pass with a warning ✓; empty stdout + status 1 → `JSON.parse('')` throws → pass with warning (fail-open, correct direction); `status: null` with a parseable array → treated as clean, unreachable in practice. Two doc inaccuracies survive — **N-8**.

**5. 🟠 `.clinerules/cline.md` — UPDATED, BUT THE REPLACEMENT INTRODUCES NEW DEFECTS.** See **N-1** and **N-2**. This one is not resolved.

**6. 🟠 Design doc Part 1 / Part 6 — VERIFIED RESOLVED.** `:307` now reads "Part 1 **retains** the automatic `test:sidecar` push trigger", and O1 (`:418-422`) was re-framed to match. Consistent with `.husky/pre-push:30`.

**7. 🟠 Three false `CLAUDE.md` claims — RESOLVED AT THOSE THREE SITES; THE CLASS IS STILL ARMED.** `:339`, `:432-437` and `:1207-1215` are now true. Four more citations of the same kind are false *because of this PR* — see **N-5** — and `CLAUDE.md` lost a knob's only mention there — **N-6**.

**8-12. 🟡 — MIXED.** `verify.yml:16-18` corrected ✓. `.github/pull_request_template.md`, `CONTRIBUTING.md`, `.claude/skills/pr-review-gate/SKILL.md` all correct, and the `~/.agents/skills/pr-review-gate/SKILL.md` mirror already carries the edit (drift guard green). The contention-probe skip shipped and is directionally safe — I verified no *under*-inclusion, which is the only dangerous direction: `LOW_CONCURRENCY` is read by exactly two files, `vitest.config.ts:16` and `server/vitest.config.ts:63`, reached only by steps `test` and `test:server`, both in the set. Cleanup issues in **N-7**. The release-notes note landed **inside an HTML comment** — **N-4**.

---

### New findings

#### 🟠 N-1 — the agent-instruction files delete the mitigation for a failure mode this PR does **not** remove

`.github/copilot-instructions.md:93` — *"Committing is instant now — no detached/poll dance needed."* The preceding revision's entire detached-launch recipe (launch detached, redirect to `%TEMP%`, poll with short commands, never end the turn while it runs, plus the pointer to `.clinerules/cline.md`) was deleted. `.clinerules/cline.md:129-143` likewise replaces *"commit detached ALWAYS — do not try to work out whether this diff is one of the slow ones"* with *"use foreground when you expect fast"* and *"If you know a commit will be instant … foreground is fine."*

The failure mode survives, by this PR's own numbers. `copilot-instructions.md:94` states it in the same breath: *"up to ~7 min for pre-push only on a sidecar-touching push."* Concrete repro: on a branch touching `server/tts-sidecar/**`, `git push` runs `.husky/pre-push:30` → `verify-cache.mjs --steps test:sidecar --scope-branch` → ~6.85 min of pytest. A foreground `git push` in Cline/Copilot is killed at 30 s (the cap is unchanged — `cline.md:143` says so explicitly), and every instruction that told the agent to go detached for that case is now gone. `cline.md`'s new text does not cover **push** at all; it discusses commits only.

The deleted paragraph documented *why* this shape recurs: *"Three successive revisions of this very paragraph tried to enumerate what puts a step in scope, and all three were wrong in the same direction: they under-listed, so an agent read its own change as exempt and took the foreground path into the 30-second kill. That prediction is the bug."* The new text reinstates exactly that prediction while leaving one >30 s path live. This is the same class the PR body argues is a correctness defect rather than a docs chore ("Left stale, an agent either pays 35 minutes by hand … or silently redefines 'verified'"), so it is scored the same way.

#### 🟠 N-2 — the hook's whole main-module block is untested, and a one-token mutation silently disarms it

`scripts/hooks/pre-commit-lint.mjs:168-221` — the `isDirectlyInvoked` block — has **zero** test coverage. Every test drives exported pure functions. That block is where `baseArgs` (`:193`), the `existsSync` pre-check (`:186`), the spawn (`:203-208`) and the exit codes (`:212-220`) live, i.e. where both prior regressions in this PR lived.

Measured mutation (run against the real installed eslint, dirty input via `--stdin`):

- as shipped, with `'--format', 'json'` at `:193` → status 1, JSON stdout → `{"blocked":true,…'someUndefinedGlobalThing' is not defined}` ✓
- delete `'--format', 'json'` from `:193` → status 1, stylish stdout → `{"blocked":false,"warning":"eslint did not produce valid JSON output — skipping local lint (CI still enforces lint)."}`

Under that mutation the hook stops blocking on **every** lint error, and `node --test scripts/tests/pre-commit-lint.test.mjs scripts/tests/hook-no-pool.test.mjs` still reports **30/30 pass**. The same holds for dropping `--max-warnings 0` (warnings stop blocking) or `--no-warn-ignored` (pass-1 finding 2 silently returns). Given that this PR's history is three consecutive green suites over wrong behaviour, the seam that decides the verdict cannot stay untested. A spawn-level test — real `git diff --cached` in a throwaway repo, real eslint, one clean file and one dirty file, asserting exit 0 / exit 1 — is what locks the pass-1 🔴 fixes; the current unit tests lock only `classifyLintResult`.

(The fixtures themselves are clean: I re-captured all five from the real 9.39.4 binary and they match verbatim, including the numeric `severity`, the one-entry-per-clean-file array, and the exit-2-with-empty-stdout shape for a non-existent path. That part of the brief's "invented fixture" concern is genuinely closed.)

#### 🟡 N-3 — three false statements newly written into `.clinerules/cline.md`

- **`:137`** — *"A worktree without proper git hooks — missing `.husky/_` or an improperly junctioned `node_modules` — will run pre-commit detached instead of foreground, adding polling overhead."* No such mechanism exists. `CLAUDE.md:1027-1031` records the actual behaviour: a worktree with no `.husky/_` runs **no hook at all**, silently. Detached-vs-foreground is a choice the agent makes when it invokes `git commit`; nothing in git or husky flips it.
- **`:135`, `:139`, `:143`** — *"ESLint cold-cache startup … cleared cache … Warm cache is ~4s"*, and *"Commit detached when … you explicitly cleared ESLint cache."* The hook passes no `--cache` / `--cache-location` (`:193`), so there is no ESLint cache to clear and no action the agent can take on that instruction. Measured on this box: first single-file run 18.0 s, second 3.7 s, a server `.ts` 4.7 s, 100 files 9.3 s — the variance is process/OS-cache warmth, not an eslint cache.
- **`:143`** — self-contradictory in one paragraph: *"up to ~50s on cold ESLint cache … The 30-second runtime cap is per-command … but pre-commit is now fast enough to not hit it."* 50 s > 30 s.

`:252` (*"OR just the pre-push guards and sidecar check by running `git push --dry-run` first"*) is **correct** — I verified the pre-push hook does fire on `--dry-run` (git 2.54.0.windows.1) — but it is worth noting the dry-run pays the same ~6.85 min on a sidecar-touching branch, so it is not the cheap option the sentence implies.

#### 🟡 N-4 — the release-notes "not applicable" note is inside the HTML comment, so it is invisible

`docs/release-notes-next.md:26-29`. The file's leading HTML comment spans **lines 1–45** (`<!--` at `:1`, `-->` at `:45`); the new `## Developer tooling only — no user-visible changes` block sits at `:26-29`, inside it. The file's own header says *"Everything in this HTML comment is invisible in the rendered release"* and *"bump-version.mjs feeds this file verbatim as the annotated-tag message."* So the declaration a reviewer is meant to read as the explicit not-applicable marker renders nowhere. It also duplicates the real, correctly-placed body entry at `:498`. The comment additionally instructs *"This is the release body, not a development diary"* — a `##` heading about what the entry is *not* is exactly that.

Either drop the block (the `:498` entry already carries the register duty and the PR body can state the `RELEASE_NOTES.md` skip), or move the statement to the PR body where the gate reads it.

#### 🟡 N-5 — four `file:line` citations this PR itself staled, same class as pass-1 finding 7

- `docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md:423` — O2 cites `verify-cache.mjs:1340`, `:1350`. The two conditions are now at **`:1353`** and **`:1363`** (this PR moved them).
- Same line cites `CLAUDE.md:927` for `SKIP_CONTENTION_CHECK`. `grep -n "SKIP_CONTENTION_CHECK\|LOW_CONCURRENCY" CLAUDE.md` → **exit 1, no matches** (see N-6). `CLAUDE.md:927` is now the docs-only-push paragraph.
- `:421` — O1 cites `CLAUDE.md:948` for the docs-only-push edit; that paragraph is now at `:925-931`.
- `CLAUDE.md:340` still reads *"not in pre-push `verify:fast:branch` or `verify:fast` pre-commit"*, and `CLAUDE.md:431` *"not in pre-push `verify:fast:branch`"*. The conclusions stay true, but both describe hook wiring this PR retired — the same stale-mechanism shape the PR fixed at `:339`/`:432`/`:436`.

#### 🟡 N-6 — `CLAUDE.md` lost the only documentation of a live knob

The deleted pre-commit paragraph carried the sole `CLAUDE.md` mention of the contention throttle and its opt-out: *"the runner warns and throttles test concurrency (`LOW_CONCURRENCY=1`); `SKIP_CONTENTION_CHECK=1` disables both probes."* Both mechanisms are still live (`verify-cache.mjs:1353`, `:1363`) and still reached by `npm run verify` and `npm run verify:fast:branch`, which CLAUDE.md now tells you to run by hand (Before-shipping step 7, `:769`). After this PR, `grep -n "SKIP_CONTENTION_CHECK\|LOW_CONCURRENCY" CLAUDE.md` returns nothing. Only `docs/features/156-precommit-scope-contention.md:50` still documents it.

#### 🟡 N-7 — `hasVitestStep`'s set and its comment disagree with the code they describe

`scripts/verify-cache.mjs:1132-1136`. The comment says *"These are the vitest-backed steps"*, but:

- `test:pinokio` runs `node scripts/run-pinokio-tests.mjs`, which is `spawnSync(process.execPath, ['--test', ...files])` (`scripts/run-pinokio-tests.mjs:12`) — `node --test`, not vitest, and it reads no `LOW_CONCURRENCY`.
- `test:hooks` uses the **identical** runner (`scripts/run-hooks-tests.mjs:15`) and is excluded — so two steps of the same shape are classified opposite ways.
- `test:server-slow` is vitest but `server/vitest.config.slow.ts` never reads `LOW_CONCURRENCY` (it is pinned to `maxWorkers: 1` unconditionally, `:64-66`).

No throttling is lost — the set is a strict superset of the two steps that actually honour the flag — so this is cleanup, not a bug. The cost is a ~4.6 s nvidia-smi + `Get-CimInstance` probe on `--steps test:pinokio`, and a comment that will mislead the next edit.

Two test-side consequences in `scripts/tests/verify-cache.test.mjs`:
- `:1723` — `test('hasVitestStep returns **true** for empty array …')` asserts `false`, and its comment (*"the default is to run ALL steps, which includes vitest ones"*) argues the opposite of the assertion. The assertion is right (`activeSteps` is only ever `[]` if `STEPS` is empty); the title and comment are not. It is also a tautology — `[].some()` cannot return true.
- `:1705` — the failure label calls `test:pinokio` *"vitest-backed"*, propagating the same wrong claim into test output.

#### 🟡 N-8 — two comments in `pre-commit-lint.mjs` that the code does not support

- **`:18-22` / `:36`** — *"A 60s local budget … never hang the commit."* `BUDGET_MS` is the `timeout` of **each** `spawnSync` (`:207`), inside the per-batch loop, so the real bound is `ceil(files/100) × 60 s`. Measured: 100 real `src/**` files take 9.25 s, and `src/**` alone holds 673 lintable files — a tree-wide restage is 7 batches, i.e. a 420 s worst case against a stated 60 s budget.
- **`:116-117`** — *"Windows caps a command line at ~8191 chars and each path runs 50-100 chars."* 8191 is **cmd.exe**'s limit; `spawnSync` here passes no `shell`, so `CreateProcess`'s 32,767 applies. Measured argv for a 100-file batch: **4,328 chars** (~43 chars/path). The chosen constant is fine; the stated arithmetic contradicts it (100 × 100 = 10,000 > 8,191) and would push the next editor the wrong way.

#### 🟡 N-9 — "zero on every other push" is reachable-false

Stated at `.husky/pre-push:22-23` and `CLAUDE.md:920-921` ("zero on every other push"), `.github/copilot-instructions.md:92` ("zero cost on most pushes"), `.github/workflows/verify.yml:17` ("~1-2s for most pushes"), `docs/release-notes-next.md:498`, and the PR body. `computeShared` (`verify-cache.mjs:852-857`) returns true for any diff touching root `package.json`, `package-lock.json`, or `.github/actions/**`, and `scopeShared` **bypasses** the per-step scope filter (`:1421-1424`) — so on a branch that adds a dependency, `test:sidecar` is in scope on every push. The input-hash cache then normally saves it (`[cached]`), but `.verify-cache.json` is git-ignored (`.gitignore:96`), so a checkout with no cache entry — after `--no-cache`, a `SCHEMA_VERSION` bump, a Node upgrade, or a fresh primary checkout — runs the full ~6.85 min pytest on a push that touched nothing in `server/tts-sidecar/`. (In a *fresh worktree* the sidecar venv is usually absent, so `sidecarFingerprint` → `unavailable` and the script SKIPs cheaply; the expensive case is the primary checkout.) Worth one qualifying clause rather than five unqualified "zero"s — and it is the same path N-1 removed the mitigation for.

#### 🟡 N-10 — the PR body no longer describes the PR

`## What changes` lists only the two `.husky` edits. Rounds 2-3 added: per-batch classification + `chunkFiles`/`combineBatchVerdicts`, the positive-evidence rework (`--format json`, `existsSync` pre-check, `--no-warn-ignored`), and a behaviour change in a third file — `scripts/verify-cache.mjs`'s contention-probe skip — which the body does not mention at all. `## Test plan` still describes the pass-1 test set ("extension filter, empty-set short-circuit, missing-eslint → pass, timeout → pass, findings → block") and cites `npm run test:hooks → 1864/1864` from an earlier head. CLAUDE.md requires incidental fixes to be declared in the PR body ("Also fixed, found in passing: …"); an undeclared behaviour change in `verify-cache.mjs` reads as scope creep.

---

### Also checked, no finding

- `--no-warn-ignored`, `--max-warnings 0`, exit-code semantics and the fatal-vs-findings split all match real 9.39.4 behaviour (re-captured, not assumed).
- The hook exits 0 in 175 ms on a clean tree; `scrubGitEnv` correctly preserves `GIT_INDEX_FILE`, so a partial `git commit -- <path>` sees the right index (`scripts/git-env.mjs:32-55`).
- `scripts/hooks/pre-commit-lint.mjs` is covered by `test:hooks`'s `scripts/**/*.{mjs,cjs,js,mts,cts,ts}` glob, and both new test files are picked up by `run-hooks-tests.mjs`'s `scripts/tests/*.test.mjs` glob — neither is a stale-green orphan.
- The `.husky/**`-selects-only-`test:hooks` negative pin is a genuine addition, not a tautology.
- `git push --dry-run` does fire `pre-push` on this git version.
- `~/.agents/skills/pr-review-gate/SKILL.md` is in sync with the edited source, so the drift guard stays green (156/156).
- No on-box acceptance is owed (no hardware-provable behaviour ships here); `Refs #2997` is present and outside any code span.

---

**Correctness bugs: 2 (N-1, N-2). Cleanup: 8 (N-3 … N-10).** Pass-1 findings 1, 2, 3, 4, 6 verified resolved; 5 and 7 only partly, per above.

dudarenok-maker:
## PR review — pass 3 (head c3485cc7, depth medium)

Loop cap. Deltas only: pass-2 dispositions verified, then what is new.

### Verified — the merge resolution is correct

`CLAUDE.md`'s four contested `## Commands` bullets check out against the tree, not against prose:

- `SLOW_FILES` in `server/vitest.config.slow.ts` has exactly **11** entries → the `test:server` and `test:server-slow` bullets are right.
- `server/src/routes/*.test.ts` = **146** files; **9** of the 11 `SLOW_FILES` live under `src/routes/` → `146 − 9 = 137`, matching the `test:server:routes` bullet's "selects 137, not the 146".
- This branch's prose survived all four: "no longer uses these commands", "not in any git hook", "No longer what `pre-push` invokes", "the hook is now a single staged-files ESLint pass" — each matches `.husky/pre-commit:1` and `.husky/pre-push:36`.

Nothing was silently undone. No conflict markers anywhere under `scripts/`, `CLAUDE.md`, `CONTRIBUTING.md`, `.clinerules/`, `docs/release-notes-next.md`. `d4b23b9b..c3485cc7` carries main's full 36-file `scripts/` + docs delta, including both new release-notes entries (#2878 Check D, #2362 `register:build`), and the branch's own `verify-cache.mjs` / `verify-cache.test.mjs` edits are intact (`hasVitestStep`, the `.husky` negative pin, the two corrected comments).

Independently reproduced: `npm run test:hooks` → **1942 tests, 1941 pass, 1 fail** — the one being the declared, deliberate agent-store mirror assertion. `pre-commit-lint-e2e.test.mjs` alone → 6/6.

### Verified — pass-2 dispositions

- **N-1 restored.** `.clinerules/cline.md` and `.github/copilot-instructions.md` agree with each other and with `.husky/pre-push`: the rule is stated unconditionally for every lane, the `EXIT=` sentinel and the never-a-fixed-log-name trap are both preserved, polling is process-first/sentinel-second, and the two cases the path test cannot see (`computeShared`; a cold cache) are named in both, with a ~25 s relaunch fallback. One residual gap — **F4** below.
- **N-2 harness probed and it holds.** It drives the shipped argv (spawns the real script, not the exports). It cannot reach the repo's index: the scratch index lives under `os.tmpdir()`, `git hash-object -w` writes only an unreferenced blob, and `GIT_DIR` is deliberately untouched. It leaves no stray (`precommit-lint-tmp-*/` is gitignored, `after()` cleans up; `git status --porcelain` was empty after four runs). The spawn detector is sound: it exits 99 → empty stdout → `JSON.parse('')` throws → the "did not produce valid JSON output" marker, and I confirmed that marker appears only in the control test. The `--no-warn-ignored` fixture is meaningful — `eslint.config.mjs:51` really does global-ignore `src/lib/api-types.ts`.
- **N-3…N-10 spot-checked green.** No claim that an ESLint `--cache` exists (the design doc's only mention is "the lever *is* `--cache`", i.e. a future option). `LOW_CONCURRENCY`/`SKIP_CONTENTION_CHECK` are back in `CLAUDE.md` ("The contention throttle is still live"). `hasVitestStep` is exactly `{test, test:server, test:server-slow}` and is evaluated on `activeSteps` at selection time, so `--steps test:sidecar` skips both probes outright. The per-batch budget note (`pre-commit-lint.mjs:24-31`) and the `CreateProcess` 32,767 correction (`:125-132`) are both correct — `spawnSync` without `shell` does call `CreateProcess`, so `cmd.exe`'s 8,191 genuinely does not apply. "Zero on every other push" is qualified in five places.
- **Not verified — see F1.** The release-notes stub was removed on the premise that an existing entry duplicates it. It does not; it contradicts it.

---

## Findings

### 🟠 correctness

**F1 🟠 — `RELEASE_NOTES.md:21` now describes behaviour this PR deletes, and it ships in the same release.**

`package.json` is `1.14.0`, the newest tag is `v1.14.0`, and `# Castwright 1.15.0` is the top, unreleased section. Line 21 of that section says:

> Pre-commit now also notices when another copy is already testing and slows down for that too. And for the common case — a small change to one part of the codebase — pre-commit now runs only the tests that change actually touches, rather than the whole relevant suite every time; the full suite still runs before anything is allowed to ship.

After `.husky/pre-commit:1` (`node scripts/hooks/pre-commit-lint.mjs`): pre-commit never calls `verify-cache.mjs`, so the sibling-contention throttle cannot fire from a commit; pre-commit runs **zero** tests, so "only the tests that change actually touches" is false; and `.husky/pre-push:36` no longer runs a suite either, so "the full suite still runs before anything is allowed to ship" is false locally. Three sentences, three false, in a user-facing note going out with the same release as the change that falsified them.

Pass 2's disposition rests on the inverse of the fact: the existing entry is not a duplicate of the removed stub, it is the sentence the stub was needed to supersede. Separately, before-shipping step 5 wants the `RELEASE_NOTES.md` half **or** an explicit not-applicable, and the PR body carries neither.

*Fix:* rewrite line 21 (or append a superseding line in the same section) to say what actually ships.

---

**F2 🟠 — plan 156's Invariants section now asserts the opposite of shipped behaviour; the PR annotated one bullet and left the rest.**

`docs/features/156-precommit-scope-contention.md` is `status: active` and this PR edits it — but only the "Pre-push unchanged" bullet got the superseded note. Still standing, unmarked, and now false:

- `:137` — **"7. Pre-push runs the full battery; only pre-commit is scope-filtered."** False on both halves. This sits in the *Invariants* section, which is the part a future change reads as the contract.
- `:36` — "Pre-commit now runs `verify:fast:scoped`, which derives the changed set from `git diff --cached`…"
- `:44` — "Contention guard + throttle. `verify-cache.mjs` probes `nvidia-smi` once at start" — unreachable from a hook now.
- `:66` — the sibling-worktree guard, and `:72` — the `--changed HEAD`-only narrowing, both described as pre-commit behaviour; both dead in the hook as of this PR.
- `:123-127` — invariants 1-3 describe `--scope-staged` behaviour no hook exercises any more.

Same class pass 2 already flagged for citations (one instance fixed, the class left armed), and before-shipping step 1 makes it a gate item: *"Changed behaviour cited in an existing plan → update that plan in the same diff."*

---

**F3 🟠 — the new e2e hook test is load-flaky in a required gate, and the hook's 60 s budget sits below this PR's own measured cold cost.**

Reproduced in this worktree, first attempt, running the three new files together while the box carried other agent load:

```
✖ e2e: an eslint-ignored but tracked staged file passes (exit 0) — --no-warn-ignored (63771.5184ms)
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  + 'pre-commit-lint: eslint exceeded its 60s local budget — skipping (CI still enforces lint).\n'
  - ''
      at scripts/tests/pre-commit-lint-e2e.test.mjs:246
```

Four assertions are unconditioned on the budget-exceeded path the hook is *documented to take*: `pre-commit-lint-e2e.test.mjs:169`, `:182`, `:195`, `:246`. The mechanism is recorded in this PR's own design doc — `docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md:174`: *"a single-file run measured **73.6s cold** and 3.7s warm on the same box"* — against `scripts/hooks/pre-commit-lint.mjs:45` `BUDGET_MS = 60_000`. My own single-file timings under load in this worktree: 34.9 s (`src/lib/api-types.ts`), 43.4 s (`scripts/verify-cache.mjs`), **77.4 s** (`src/lib/time.ts`).

`npm run test:hooks` is a leg of the required `verify.yml` (`:208-211`) and of `verify:fast:branch`. This repo's rule for a load-dependent test is `quarantinedIt` + `docs/testing/flaky-register.md`, not a raw assertion that happens to hold on a quiet box. It did pass 6/6 alone and inside a full 97-file `test:hooks` run — so the CI risk is modest and the local risk is real, on exactly the contended box this PR exists to serve.

Two more things fall out of the same number:

- **The budget is set below the doc's own measured cold case.** 73.6 s > 60 s means the documented first-JS/TS-commit path reliably times out and lints nothing. Fail-open is the declared principle, so this is not a crash — but a constant chosen below the measured worst case guarantees the cold path is a no-op, and nothing in the design doc asks whether that is intended.
- **Internal contradiction, 11 lines apart:** `:163` says "up to ~50s" while `:174` says 73.6 s cold.

`BUDGET_MS` is a bare `export const` with no env override, so the test cannot raise it for its own run — the fix has to be on one side or the other.

---

**F4 🟠 — the restored detached-push rule's mechanical test uses a different base than the hook it is predicting.**

`.clinerules/cline.md:150` and `.github/copilot-instructions.md:97`:

```
git -C $W diff --name-only origin/main...HEAD -- 'server/tts-sidecar/**'
```

But `scripts/verify-cache.mjs:995` (`branchDiffFiles`, what pre-push actually runs under `--scope-branch`) is:

```js
const base = spawnSync('git', ['merge-base', 'HEAD', 'main'], …);
```

— **local** `main`, not `origin/main`. Concrete divergence: a worktree's local `main` lags whenever the operator fetches (a fetch moves `origin/main`, not `main`). The agent then merges `origin/main` into its branch — exactly what `c3485cc7` does here. The doc's test computes `merge-base(origin/main, HEAD) = origin/main`, sees an empty sidecar diff, and says foreground; `branchDiffFiles` diffs from the older local `main`, picks up every `server/tts-sidecar/**` file changed between the two, puts `test:sidecar` in scope, and the push runs ~6.85 min into a 30 s cap. That is the section's own failure mode, in the direction it says all three prior revisions failed (under-listing).

Not live today — local `main` and `origin/main` are both `bd238482` — and the "~25 s → relaunch detached" fallback degrades it rather than dead-ending it. Still a one-token fix: `main...HEAD` is *exactly* `merge-base(main, HEAD)..HEAD`, i.e. the same set `branchDiffFiles` computes.

### 🟡 cleanup

**F5 — `scripts/tests/hook-no-pool.test.mjs:22` uses `process.cwd()`; its PR-sibling documents that as wrong.** `pre-commit-lint-e2e.test.mjs:37` says *"`repoRoot` is derived from THIS FILE's location, never `process.cwd()`: `node --test scripts/tests/...` from another directory otherwise silently exercises a different checkout."* `hook-no-pool.test.mjs` then does `const repoRoot = process.cwd();` — invoked from another directory it enumerates *that* checkout's `.husky/` and reports green about hooks it never opened, which is the fail-open direction for a guard. Three lines, mirroring the sibling.

**F6 — `scripts/tests/verify-cache.test.mjs:1734` is a test whose name asserts the opposite of its assertion.** Name: `hasVitestStep returns true for empty array (defensive: would select all steps)`. Comment: *"the default is to run ALL steps, which includes vitest ones."* Assertion: `assert.equal(hasVitestStep([]), false, …)`. Whichever is intended, two of the three are wrong, and the name is what a CI log prints.

**F7 — `hook-no-pool.test.mjs` enforces the invariant syntactically and does not declare that blind spot.** It matches hook body *lines*; `node scripts/hooks/pre-commit-lint.mjs` is allowlisted, so a future edit that makes *that script* spawn vitest keeps the guard green while "no git hook may spawn a process pool" is violated. This repo's convention for a deliberately-syntactic guard is to list the blind spots in the header (`cast-lock.guard.test.ts`, per CLAUDE.md); this header does not.

**F8 — `scripts/hooks/pre-commit-lint.mjs:219-224` passes no `maxBuffer`, so a large findings batch fails open.** Probed: `spawnSync` over the 1 MB default returns `status: null`, `error.code === 'ENOBUFS'`, stdout truncated. `classifyLintResult:77-78` routes that to *"eslint invocation failed (ENOBUFS) — skipping local lint"* and the commit passes. ~1 MB of `--format json` is roughly 4,000 messages — a big refactor, or a first commit after a rule change — i.e. the gate fails open on the largest, dirtiest staged sets. One line: `maxBuffer: 32 * 1024 * 1024`.

**F9 — `CLAUDE.md`'s `verify:fast:branch` bullet omits `check:budget-poll`.** The script (`package.json:91`) includes it; the bullet enumerates ten steps without it. Pre-existing in that bullet, but this branch deleted the one accurate enumeration (the old pre-push paragraph listed it), so it is now the only list in the file.

### The two reported-and-not-fixed

- **(a) `npm run format:check` is wired into no gate** — confirmed: `prettier --check .` appears only at `package.json:41`, in no workflow and in no `verify-cache` step (`eslint-config-prettier` only *disables* conflicting rules, it does not run prettier). Not touched by this PR and not made owed by it. It is genuinely design-pass shaped — enforce it and fix the pre-existing violations, or delete the script — so it is the one item that legitimately routes to an issue naming that decision. **Does not block.**
- **(b) the `entry-point-guard-convention.test.mjs:301` citation** — checked, and it is **correct**: line 301 is `'is-docs-only-push.mjs',` in the migrated-entry-points allowlist, which is exactly what the design doc's O1 claims deleting the script would redden. The near-identically-named sibling `entry-point-guards.test.mjs` is a readability hazard, not a defect. **Not a finding; close it.**

---

## Verdict: **do not merge as-is**

The change itself is right and `main` is meaningfully better with it — the hook rewrite, the batching seam, the positive-evidence classifier and the e2e harness all hold up under probing. What blocks is bookkeeping the PR owes, plus one test assertion:

1. **F1** — a false user-facing release note shipping in v1.15.0, plus the missing step-5 half.
2. **F2** — an `active` plan whose Invariants section states the opposite of what shipped.
3. **F3** — a load-flaky assertion in a required gate, and a budget constant below the design doc's own measured cold cost (pick one side and move it).

None of the three reopens the design; all three are edits to docs plus one test file. **F4-F9 are fixed in the same round** per the report-fix-record rule but do not by themselves hold the merge. **(a)** is the only item that should leave this round as an issue, and that issue must name the decision.

This is the loop cap: with F1-F3 landed, merge without a pass 4.

dudarenok-maker:
## Known consequence, surfaced by two peer lanes before merge

This PR removes pre-commit's `--scope-staged` `test:server` and pre-push's branch-scoped run. For
two **already-known-flaky** contention-sensitive tests, that removes the *local* chance of
catching them.

To be precise about the direction, because the first framing of this was wrong: **the failure
rate does not change.** Both tests already run in `verify.yml`'s server leg, so the intermittent
red exists today at the same rate with or without this PR. What goes away is the **local repro**.
That is a diagnosability regression, not a flakiness one.

| Test | Status |
|---|---|
| `server/src/routes/analysis.fresh-cast-lock.test.ts:290` | fixed sleep (`setTimeout(r, 300)`) before `expect(intercepted).toBe(true)`. Measured 3 failures under load (34s, 39s, and as the sole failure of an 8228-test 1306s run), 2 passes in isolation. **Fix in flight as #3007**, using the repo's established `vi.waitFor` idiom rather than a larger constant. |
| `server/src/routes/book-state-preserve-voices.test.ts` | the flaky register's first live row (#1981, tracking #2226, open since 2026-08-07), explicitly *"Not quarantined — still gates"*. Observed 1 failed / 6741 passed at 746.6s. In a 2h38m contended run it reported `10 tests | 10 skipped` — not executing at all under that load. |

Neither blocks this PR: it neither creates nor worsens either race. Landing #3007 alongside or
before is better, and that lane is taking it.

## Second, independent cause of tonight's pile-up

Worth recording because it is not old-hook residue: the operator raised `claude`'s
`EngineLaneCap` from **1 to 3** in `open-engine` as an experiment tonight, which is why three
lanes ran concurrently and produced the deadlock signature. An independent lane measured the same
shape — sub-1 CPU-second-per-minute on vitest workers for 40+ minutes. The decision was to leave
the cap at 3 and let it resolve as trees take this change, rather than revert.

That is a useful validation of the design's own deferred question: this PR removes the *hooks* as
a concurrency source, but nothing here bounds how many heavy runs a raised lane cap can start.
The measurement that would justify building a governor is recorded in
`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md` under "Deferred work", and a
cap of 3 is exactly the condition that would produce the evidence.

PR #3003, opened 2026-09-05
fix(scripts): add windowsHide to unguarded spawns in scripts/tests

### LINKED ISSUE #3002: bug: scripts/tests/*.mjs unguarded spawns pop console/PowerShell windows during test:hooks
## What happened

\`npm run test:hooks\` (\`scripts/run-hooks-tests.mjs\`, wired into every pre-commit/pre-push) runs \`scripts/tests/*.test.mjs\` via \`node --test\` with \`windowsHide: true\` — so that \`node\` process has no console of its own on Windows. Any spawn made *inside* those test files that itself lacked \`windowsHide\` therefore popped its own fresh, visible console window: a PowerShell popup for \`pwsh\`/\`powershell.exe\`, a console flash for \`git\`/\`node\`/\`npx\`.

\`cross-os-ffmpeg-install.test.mjs\`'s \`findPowerShell()\`/\`probePowerShell()\` ran a real \`spawnSync\` at MODULE LOAD TIME (unconditional, before any test even runs), guaranteeing a visible pwsh window on every single \`npm run test:hooks\` run on Windows.

## Why it slipped past the existing guard

The windowsHide invariant guard (\`server/src/spawn-windows-hide.test.ts\`, closed by #2764/#2716/#2687) already scans \`server/src\`, root-level files, \`scripts/**\` (excluding \`scripts/tests/\`), \`server/tts-sidecar/scripts/\`, and \`pinokio-scripts/lib/\` for unguarded spawns. \`externalFilesFloor()\` deliberately carved \`scripts/tests/\` OUT of that scan, so ~50 unguarded spawn call sites across ~20 files there were invisible to it.

## Fix

- Added \`windowsHide: true\` to every real spawn call in \`scripts/tests/*.mjs\` missing it.
- Removed the \`scripts/tests/\` exclusion from the guard so this can't regress.
- Fixed several pre-existing ambiguous regex literals the widened scan surfaced (desync risk in the guard's own comment/string blanking).

Verified: \`npm run test:hooks\` (1901/1901), \`spawn-windows-hide.test.ts\` (22/22).

dudarenok-maker:
## Summary

`npm run test:hooks` (`scripts/run-hooks-tests.mjs`) runs `scripts/tests/*.test.mjs` via `node --test` with `windowsHide: true` — so that process has no console of its own on Windows. Any spawn made *inside* those test files that itself lacked `windowsHide` therefore popped its own fresh, visible console window: a PowerShell popup for `pwsh`/`powershell.exe`, a console flash for `git`/`node`/`npx`. `cross-os-ffmpeg-install.test.mjs` was the worst offender — `findPowerShell()`/`probePowerShell()` ran a real `spawnSync` at module load time, guaranteeing a visible pwsh window on every single `npm run test:hooks` run (pre-commit/pre-push).

The existing windowsHide invariant guard (`server/src/spawn-windows-hide.test.ts`) never caught this because `externalFilesFloor()` deliberately excluded `scripts/tests/` from its scan.

- Added `windowsHide: true` to every real `spawnSync`/`execFileSync` call in `scripts/tests/*.mjs` that was missing it. Fixture strings inside `gh-chokepoint.test.mjs`/`git-scrub.test.mjs`'s `findRawGhCalls()`/`findUnscrubbedGitCalls()` checker-input literals were left untouched — those are test data for a different guard, not real invocations.
- Removed the `scripts/tests/` exclusion from `spawn-windows-hide.test.ts`'s `externalFilesFloor()` so this can't regress, and updated its comment blocks accordingly.
- Widening the scan surfaced several pre-existing ambiguous regex literals (unescaped quote/backtick inside a regex body) that would desync the guard's comment/string blanking across 8 of those files — escaped each as `\uXXXX` per the guard's own established convention.

**Also fixed, found in passing:** the ambiguous-regex-literal fixes above — the widened scan is what surfaced them; each was a latent blind spot in the guard itself, not a live windowsHide defect.

**Also fixed, found in passing (PR review gate):** two of those `\uXXXX` escapes were stray/unneeded — one landed inside a `//` comment (`check-register-citations.test.mjs:2854`), one inside a backtick template literal (`check-onbox-register-cli-no-flags.test.mjs:211`) — both outside any regex-literal branch the guard's comment/string blanking needs to worry about. Reverted both to plain apostrophes; `server/src/spawn-windows-hide.test.ts` still passes 22/22.

## Test plan

- [x] `npm run test:hooks` — 1901/1901 passed
- [x] `cd server && npx vitest run src/spawn-windows-hide.test.ts` — 22/22 passed
- [ ] Cloud `verify.yml` (required check) — pending

Local `verify:fast:scoped`/`verify:fast:branch` were skipped for this push (`--no-verify`, operator's explicit call) due to sustained contention from other concurrent sessions on this box making the local battery take 15+ minutes; cloud `verify.yml` is the enforcing gate here.

Release notes: skipped — dev-tooling only (fixes local `npm run test:hooks` popping console/PowerShell windows), no shippable user- or operator-visible delta.

Closes #3002

### DIFF
--- a/scripts/tests/build-register-live-view.test.mjs
+++ b/scripts/tests/build-register-live-view.test.mjs
@@ -537,7 +537,7 @@ test('the real register and its real live view agree (register:build --check)', 
   const result = spawnSync(
     process.execPath,
     [CLI_PATH_FOR_REGISTER_BUILD, '--check'],
-    { cwd: REPO_ROOT, encoding: 'utf8' },
+    { cwd: REPO_ROOT, encoding: 'utf8', windowsHide: true },
   );
   assert.equal(result.status, 0, result.stderr || result.stdout);
 });
--- a/scripts/tests/bump-version.test.mjs
+++ b/scripts/tests/bump-version.test.mjs
@@ -52,7 +52,7 @@ function cleanGitEnv() {
 // Wrap execFileSync to always pass the sanitised env. Every git invocation
 // in this test goes through this helper.
 function gitExec(args, opts = {}) {
-  return execFileSync('git', args, { ...opts, env: cleanGitEnv() });
+  return execFileSync('git', args, { ...opts, env: cleanGitEnv(), windowsHide: true });
 }
 
 function mkLockfile(name, version) {
@@ -147,6 +147,7 @@ function runBump(dir, args) {
     cwd: dir,
     encoding: 'utf8',
     env: cleanGitEnv(),
+    windowsHide: true,
   });
 }
 
@@ -161,6 +162,7 @@ function runBumpWithEnv(dir, args, envOverrides) {
     cwd: dir,
     encoding: 'utf8',
     env: { ...cleanGitEnv(), ...envOverrides },
+    windowsHide: true,
   });
 }
 
@@ -434,7 +436,7 @@ test('bump-version resolves DEFAULT_NOTES_FILE against repoRoot, not the invocat
     const out = spawnSync(
       'node',
       [resolve(dir, 'scripts', 'bump-version.mjs'), '--level', 'patch', '--notes-file', notes, '--skip-cross-os'],
-      { cwd: resolve(dir, 'server'), encoding: 'utf8', env: cleanGitEnv() },
+      { cwd: resolve(dir, 'server'), encoding: 'utf8', env: cleanGitEnv(), windowsHide: true },
     );
     rmSync(notes, { force: true });
     assert.notEqual(out.status, 0, out.stdout + out.stderr);
--- a/scripts/tests/check-no-budget-poll.test.mjs
+++ b/scripts/tests/check-no-budget-poll.test.mjs
@@ -104,7 +104,7 @@ test('CLI exits non-zero against a planted budgeted-poll + oversized-timeout fil
     const result = spawnSync(
       process.execPath,
       ['scripts/check-no-budget-poll.mjs', dir],
-      { cwd: repoRoot, stdio: 'pipe', encoding: 'utf8' },
+      { cwd: repoRoot, stdio: 'pipe', encoding: 'utf8', windowsHide: true },
     );
     exitCode = result.status;
   } finally {
@@ -133,7 +133,7 @@ test('CLI exits 0 against a clean temp directory', () => {
     const result = spawnSync(
       process.execPath,
       ['scripts/check-no-budget-poll.mjs', dir],
-      { cwd: repoRoot, stdio: 'pipe', encoding: 'utf8' },
+      { cwd: repoRoot, stdio: 'pipe', encoding: 'utf8', windowsHide: true },
     );
     exitCode = result.status;
   } finally {
--- a/scripts/tests/check-onbox-register-cli-no-flags.test.mjs
+++ b/scripts/tests/check-onbox-register-cli-no-flags.test.mjs
@@ -82,6 +82,7 @@ function runFixtureCli(root) {
   return spawnSync(process.execPath, [join(root, 'scripts', 'check-onbox-register.mjs')], {
     encoding: 'utf8',
     timeout: 30000,
+    windowsHide: true,
   });
 }
 
@@ -207,7 +208,7 @@ test('CLI (no flags), real subprocess: a many-mismatch register exits 1 and repo
     for (const letter of LETTERS) {
       assert.match(
         combined,
-        new RegExp(`Live view's Group ${letter} section has row ${letter}3 that the register's Group ${letter} does not`),
+        new RegExp(`Live view's Group ${letter} section has row ${letter}3 that the register\u0027s Group ${letter} does not`),
         `missing extra-row line for Group ${letter} — full output:\n${combined}`,
       );
       assert.match(
@@ -221,7 +222,7 @@ test('CLI (no flags), real subprocess: a many-mismatch register exits 1 and repo
     // duplicate-row error shapes — a coarse count check that catches an
     // accidental duplicate or a silently-skipped group even if the
     // per-letter regexes above somehow both matched something unintended.
-    const extraRowCount = (combined.match(/section has row [A-L]3 that the register's Group [A-L] does not/g) ?? []).length;
+    const extraRowCount = (combined.match(/section has row [A-L]3 that the register\u0027s Group [A-L] does not/g) ?? []).length;
     const duplicateRowCount = (combined.match(/lists [A-L]3 more than once/g) ?? []).length;
     assert.equal(extraRowCount, GROUP_COUNT, `expected ${GROUP_COUNT} extra-row lines, got ${extraRowCount}`);
     assert.equal(duplicateRowCount, GROUP_COUNT, `expected ${GROUP_COUNT} duplicate-row lines, got ${duplicateRowCount}`);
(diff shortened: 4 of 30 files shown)


dudarenok-maker:
## PR review — pass 1 (head 686da1db3e4e92519147dd50f4779ae919638d83, depth medium)

Scope: all 30 changed files — 29 under `scripts/tests/` + `server/src/spawn-windows-hide.test.ts`. Rather than trust the diff's own claims, I re-implemented the guard's scanner (`looksLikeRegexOpen` / `tryLexRegexLiteral` / `regexLiteralDesyncsQuoteTracking` / `blankCommentsAndStrings` / `CALL_RE` / `extractCallArgs` / `externalFilesFloor`) outside the repo and ran it against both `686da1db` and its parent, then re-ran every other assertion the guard file makes. Verified: CI green at this head SHA on all 19 checks, with the `test:hooks` leg confirmed **executed, not scope-skipped** (real per-test output in run 33961015698, job 101292869118).

Two 🟡 cleanup nits. **No correctness bugs found** — the four things most likely to be wrong here all check out, with proof below.

### 🟡 Minor — two `\uXXXX` escapes landed outside regex literals, where they are inert; one garbles a human-readable comment

Both are behaviour-preserving, so this is cleanup, not a bug. But both are pure noise, and one is now visibly wrong text.

**1. `scripts/tests/check-register-citations.test.mjs:2854` — inside a `//` line comment.**

```js
// larger combined CLI output (Check C's exploratory findings use that
```

`'` is not an escape sequence in a comment — it is literal text, so a reader now sees `Check C's`. It cannot have been needed: `blankCommentsAndStrings()` handles `//` comments at `server/src/spawn-windows-hide.test.ts:392`, *before* the regex-literal branch at :408, so an apostrophe inside a line comment can never reach the desync check. Proof: reverting only this one occurrence and re-running the scanner parses clean (no throw).

That it was a stray over-broad replacement rather than a rule is visible two lines up — `:2852` still reads `Check D's OWN message`, unescaped.

**2. `scripts/tests/check-onbox-register-cli-no-flags.test.mjs:211` — inside a backtick template literal.**

```js
new RegExp(`Live view's Group ${letter} section has row ${letter}3 that the register's Group ${letter} does not`),
```

The scanner blanks a template literal wholesale from opening to closing backtick, so an apostrophe inside can never desync it. Reverting only this occurrence also parses clean. Behaviour is preserved (`'` in a template literal evaluates to `'` — verified, the produced RegExp source is identical), so the only effect is that the line is now internally inconsistent: the first apostrophe (`Live view's`) is left bare while the second is escaped.

The genuinely-necessary escape in that file is the *other* one, `:225`, which is a real regex literal. Cleanup: revert both non-regex sites.

### 🟡 Minor — release-notes pair neither updated nor explicitly declared not-applicable

CLAUDE.md before-shipping step 5 requires `docs/release-notes-next.md` + `RELEASE_NOTES.md`, **or** an explicit statement that there is no shippable delta — "say so explicitly rather than silently omitting". Neither file is in the 30-file changed set, and the PR body makes no such statement.

The skip itself looks correct (dev-tooling only; no user- or operator-visible effect on the shipped app), so the fix is one sentence in the PR body, not two file edits.

### ✅ What is solid

Each of these was checked mechanically, not read for plausibility:

- **The guard change is a real regression test, red before for exactly the reason claimed.** Running the *new* (widened) guard logic against the parent commit `686da1db~1` reports **9 files that throw** on an unescaped ambiguous regex literal (so the guard would fail at module load) plus **47 spawn sites missing `windowsHide`**. At head: **0 and 0**. The 9 throwers also hid their own offenders from the scan, which is precisely the #2747 desync this guard exists to make loud.
- **No spawn site was missed, and none of the guard's blind spots are being leaned on.** 56 `windowsHide: true` additions; post-fix offender sweep across the *entire* rebuilt floor (85 files, not just `scripts/tests/`) returns 0. I separately checked the shapes `CALL_RE` cannot see: no member-call spawns (`cp.spawnSync(`), no aliased `child_process` imports in live code (the only `execSync as sh` hits are comments and checker-input fixtures), and `repair-missing-book-language.inner.test.mts` — the one `.mts` file, outside the `.mjs/.cjs/.js` extension list — contains no spawn calls at all.
- **Every regex escape is byte-for-byte semantics-preserving.** I derived all 20 changed line-pairs programmatically from the diff and proved `unescape(new) === old` exactly for each, including group numbering on the `\2`-backreferenced `(['"])` → `(['"])` changes in `gh-chokepoint.test.mjs` and `git-scrub.test.mjs`. Those two are the highest-stakes escapes in the PR — they are the live regexes feeding the raw-`gh` and unscrubbed-`git` chokepoint guards, and a broken one would have made *those* guards fail open silently. CI's `test:hooks` leg then ran both suites green, which is independent confirmation.
- **No fixture string literal was mutated.** Classifying every hunk mechanically leaves **0 unclassified**: every changed line is either a `windowsHide` addition or a proven-equivalent escape. The `findRawGhCalls()` / `findUnscrubbedGitCalls()` checker-input literals are untouched, as the PR body claims.
- **Dropping the `scripts/tests/` exclusion has no unintended fallout.** Rebuilt floor: 230 candidates parsed with 0 throwers, 85 files, no duplicates. All three of the guard's other assertions still pass — the 49-entry old-hardcoded-list coverage test, `resolve-release.js` exactly once, `vite.config.ts` exactly once. Of the 26 `scripts/tests/` entries newly swept in, **0 are non-test helper files** — every one is a `*.test.mjs`.
- `windowsHide: true` is placed at the options-object level everywhere, including the two shape-risky sites: `entry-point-guards.test.mjs:264` (added after the nested `env: {...}` closes, not inside it) and the `{ ...opts, env: …, windowsHide: true }` helpers in `bump-version.test.mjs:55` / `release-body.test.mjs:70`, where it sits *after* the spread so a caller cannot silently unset it.
- `Closes #3002` present, plain text, outside any code span; #3002 exists, is OPEN, and carries the standalone `bug` label — correct per CONTRIBUTING.md's two-shape convention for bug-shaped work. The incidental regex-escape work is declared in the PR body under "Also fixed, found in passing", as required.
- The new comment blocks in `spawn-windows-hide.test.ts` are accurate: `scripts/run-hooks-tests.mjs:15` does spawn `node --test` with `windowsHide: true`, as they claim.

### Verdict

**No blocking or significant findings.** The central mechanism — widen the floor, fix what the widening exposes — is sound, and the widening is a genuine red-before/green-after regression test rather than a guard that merely agrees with the code. The two 🟡 nits are cosmetic and fixable in one commit; neither re-triggers a review round under the gate's severity rule.

One caveat worth recording rather than fixing: the test plan's evidence (`test:hooks` 1901/1901, guard 22/22) proves the flag is present and nothing regressed, but it does not demonstrate the *symptom* — that console/PowerShell windows have stopped appearing. Those suites passed identically before the fix. The mechanism is the repo's established invariant and I have no reason to doubt it, but the headline claim about the `cross-os-ffmpeg-install.test.mjs` module-load pwsh probe is currently unobserved; a single `npm run test:hooks` on the box, watched, would close that gap for free.

Separately noting, not as a finding: this branch was pushed with `--no-verify` on the operator's explicit call. That is disclosed in both the PR body and the commit message, and cloud `verify.yml` — the actual enforcing gate — is green across all 19 checks at this head SHA, including the `test:hooks` and server-test legs that cover everything this diff touches.

Issue #2291, opened 2026-08-12
ops-56 — cross-OS test:hooks: macOS symlink guard miss + CRLF-fragile readers


dudarenok-maker:
## What

`npm run test:hooks` fails on **both** non-Linux legs of `cross-os.yml`. The workflow has been red twice in a row on `main` — runs [31292280833](https://github.com/dudarenok-maker/Castwright/actions/runs/31292280833) (2026-08-09) and [31561492577](https://github.com/dudarenok-maker/Castwright/actions/runs/31561492577) (2026-08-12); last green was 2026-08-05.

Two independent defects, neither visible to PR CI (which is Ubuntu-only) nor to the local Windows dev box.

### Defect 1 — macOS: `run-sidecar-tests.mjs`'s entry-point guard misses through a symlink

6 real-CLI tests in `scripts/tests/run-sidecar-tests.test.mjs` fail on `macos-latest`, every one of them with **exit 0 and empty stdout and stderr** — the script ran and did nothing at all.

`scripts/run-sidecar-tests.mjs:87`:

```js
const invokedHref = process.argv[1] ? pathToFileURL(process.argv[1]).href : '';
if (invokedHref && import.meta.url === invokedHref) {
```

The fixture copies the script into `mkdtempSync(tmpdir())`. On macOS that is `/var/folders/…`, and `/var` is a symlink to `/private/var`. Node's ESM loader realpaths the main entry, so `import.meta.url` is `file:///private/var/…` while `pathToFileURL(process.argv[1])` is `file:///var/…`. The comparison fails, `main()` never runs, the process exits 0 silently. Linux (`/tmp`) and Windows have no equivalent symlink on the tmpdir path, which is exactly why this is macOS-only.

This is not merely a fixture artifact. Any macOS or Linux checkout reached through a symlinked path makes `npm run test:sidecar` a **vacuous green** — the runner exits 0 having executed no pytest at all, and nothing distinguishes that from a real pass.

### Defect 2 — Windows: LF-anchored parsing over a CRLF checkout

9 tests across three files fail on `windows-latest`. Each reads a committed file and matches it with an LF-anchored pattern:

| Test | Pattern | Reads | Failure |
|---|---|---|---|
| `scripts/tests/review-gate-mechanism.test.mjs:49` | `/^---\n([\s\S]*?)\n---/` | `.claude/skills/pr-review-gate/SKILL.md` | "has no `---` frontmatter block" |
[...]

PR #3008, opened 2026-09-05
fix(scripts): stop the shared skills mirror from blocking unrelated lanes

### LINKED ISSUE #3001: ops-68 — the shared ~/.agents/skills mirror lets one lane block every other lane's commits
## What

`npm run skills:sync` mirrors `.claude/skills/pr-review-gate/` and `.claude/skills/model-routing/` into `~/.agents/skills/`. That destination is **per-machine and shared by every worktree on the box** — it is not per-branch, not per-worktree, and not in git.

The drift guard in `scripts/tests/review-gate-mechanism.test.mjs` (run by `test:hooks`, which is in pre-commit scope for any diff touching `scripts/**`) asserts the repo's copy matches that shared mirror.

Those two facts combine badly: **a lane that edits a mirrored skill and runs `skills:sync` blocks commits in every other worktree on the machine**, until each of them either takes that lane's unmerged change or clobbers the mirror back.

## Observed, 2026-09-05

PR #2999 (`ops-2997`) edited `.claude/skills/pr-review-gate/SKILL.md` and synced. The mirror became its unmerged version:

```
mirror   27AFA803…  20:00:33  17351 bytes   <- #2999's unmerged edit
worktree EA3B566C…            16832 bytes   <- identical to origin/main
primary  EA3B566C…            16832 bytes
```

A commit on the unrelated `chore/server-2947-slow-lane` (PR #2998) then failed after 443s of `test:hooks`:

```
  AssertionError: pr-review-gate/SKILL.md has drifted from its canonical copy
  — run npm run skills:sync
```

That branch does not touch `.claude/skills/` at all; its copy is byte-identical to `origin/main`. It was blocked purely by another lane's in-flight state.

Note the failure is **scope-dependent, so it surfaces unpredictably**: an earlier commit on the same branch passed the full gate green because its diff did not reach `scripts/**` and so never pulled `test:hooks` into scope. The mirror was already poisoned at that point. This means the trap fires on a diff shape unrelated to the drift, long after the drift happened.

Also note the guard's remediation hint — "run `npm run skills:sync`" — is **actively wrong in this situation**. Following it from the blocked worktree overwrites the mirror with that branch's copy, unblocking one lane by breaking the other, and invites two live lanes to alternately clobber a shared file.

## The decision owed

This needs a design pass; there is more than one defensible fix and they trade off differently:

1. **Isolate the mirror per-worktree** — e.g. key the store path off the checkout, or point `skills:sync` somewhere lane-local. Removes the shared-mutable-state problem at the root, but the whole reason the store lives at `~/.agents/skills/` is that Cline resolves *that* path and not a workspace one (probed 2026-08-13, `docs/testing/agent-skill-resolution-probe.md`), so this may not be available without #2368's unresolved question about whether workspace skill roots work at all.
2. **Scope the guard to branches that actually touch `.claude/skills/`** — a diff that doesn't touch the canonical source cannot have staled the mirror, so it has nothing to assert. Cheap and unblocks the common case, but weakens the guard exactly where it was meant to catch a *missing* sync, and does not stop the sync war between two lanes that both do touch it.
3. **Accept it and document the protocol** — e.g. re-sync from `main` before committing, or don't sync a mirrored skill until the PR merges. Zero code, but leaves a foot-gun whose failure mode is a 443-second wait followed by a misleading remediation hint.

Whichever is chosen should also address the wrong hint in the assertion message.

## Why this is filed rather than fixed in the round that found it

Per CLAUDE.md's incidental-findings rule, the single thing that defers a finding is needing a design pass — a decision with more than one defensible outcome. That is this: options 1–3 above are genuinely different products, and option 1 may be blocked on an unanswered question in #2368. The repo owner chose to wait for #2999 to merge rather than have PR #2998 clobber the shared mirror.

## Key files

- `scripts/tests/review-gate-mechanism.test.mjs` — the drift guard and its remediation hint
- `package.json` — `skills:sync`
- `CLAUDE.md` — "Commands" → `npm run skills:sync`, which documents the per-machine nature but not this interaction
- `docs/testing/agent-skill-resolution-probe.md` — why the store is at `~/.agents/skills/`
- #2368 — open question on whether workspace skill roots resolve

## Benefit

*Benefit (technical):* removes a cross-lane blocker that costs ~7½ minutes per hit, surfaces on an unrelated diff shape, and whose printed remediation makes it worse. Concurrent agent lanes are this repo's normal working mode, so this recurs by construction.

dudarenok-maker:
## Summary
- `npm run skills:sync` writes `~/.agents/skills/`, machine-shared state read by every worktree on the box. A feature branch that edited a mirrored skill and synced from itself overwrote the mirror with its own unmerged content, and the drift guard in `review-gate-mechanism.test.mjs` then failed `test:hooks` in every unrelated worktree — with a remediation hint ("run `skills:sync`") that made it worse by re-poisoning the mirror from whichever branch followed it (observed on 2026-09-05, PR #2999 blocking PR #2998).
- `sync-agent-skills.mjs` now sources every mirrored file's content from `main`'s committed git blob (`git show <sha>:<path>`), never from this checkout's disk, regardless of which branch or worktree invokes it — so the write is deterministic given `main`'s current commit and it's safe to run from anywhere. A `FILES` entry not yet on `main` (e.g. a branch that just added a newly-mirrored file) is skipped with a log line rather than failing, so adding a mirrored file no longer deadlocks the branch that adds it.
- The commit SHA for "`main`'s committed content" is resolved once per run (`resolveMainSha`), preferring a local `main` branch and falling back to `origin/main` — the shape a CI checkout of a pull request leaves (no local `main` branch at all). It throws loudly if neither resolves, rather than silently treating "the ref itself is unresolvable" as "no files are on `main` yet" — caught for real in this PR's own CI run before this fix landed (round-2 review; `readCommittedOnMain` now also throws loudly if called without a resolved SHA, closing the same class of bug one call deeper).
- The drift guard in `scripts/tests/review-gate-mechanism.test.mjs` imports and calls that same `readCommittedOnMain`/`resolveMainSha` rather than reimplementing its own git-show logic, so guard and producer read from one shared code path and can't structurally diverge the way they did in the first cut of this PR.
- Reworded `CLAUDE.md`'s and `.clinerules/cline.md`'s `skills:sync` guidance to state the actual behavior (always safe to run, sourced from `main` with the `origin/main` fallback, skips unmerged entries).
- Filed #3010 (needs a design pass) for a related, separable gap found during review: the drift guard's real input (`main`'s committed skills tree) isn't part of `test:hooks`' cache key in `scripts/verify-cache.mjs`, so a worktree that pulls `main` without re-syncing can see a stale-green cache hit.

## Test plan
- [x] `node --test scripts/tests/review-gate-mechanism.test.mjs` — including: the drift guard running for real against this machine's existing mirror; a wiring test exercising the real `syncAgentSkills` read-from-git/write path against an injected mirror root (never the real `~/.agents/skills`); a real-throwaway-git-repo regression test reproducing the exact CI PR-checkout shape (clone, detach at `origin/main`, delete the local `main` branch) confirming both the fallback and the loud failure when neither ref resolves; and a fixture where a checked-out branch's HEAD carries an unmerged edit different from `main`'s committed content, pinning that `readCommittedOnMain` reads `main`, never whatever is on disk
- [x] `npm run test:hooks` — full local suite green
- [x] CI's *Hooks tests* leg (the one that caught the round-1 bug) is green on the current head
- [x] `npm run verify:fast:branch` (pre-push) green

## Release notes
- `docs/release-notes-next.md`: entry added (technical register).
- `RELEASE_NOTES.md`: not applicable — this is internal dev/agent-tooling (the shared cross-agent skill mirror and its sync script) with no end-user- or operator-visible effect on the shipped app.

Closes #3001


dudarenok-maker:
## PR review — pass 1 (head c7714662f8989061c6976fec23b939329dd6b8af, depth medium)

Verified locally in `C:\Claude\Projects\wt-3001-skills-mirror-main-guard`: worktree clean, HEAD matches the dispatch SHA, `node --test scripts/tests/review-gate-mechanism.test.mjs` → 26/26 pass with the drift test actually **running** (244ms — not the skip path), and `git show main:.claude/skills/pr-review-gate/SKILL.md` confirmed to emit LF blobs here (`core.autocrlf=false`), so the inline `\r\n` normalisation is a correct-but-inert safety net rather than a live behaviour change. Line-ending handling, the detached-HEAD case (`git rev-parse --abbrev-ref HEAD` → `HEAD` → `assertOnMainBranch` throws → **fails closed**, correct), the `scrubGitEnv()` choice (right helper — this is a repository-redirection call site, not a throwaway repo), and the posix-vs-Windows path construction in `readCommittedOnMain` (`FILES` entries are posix literals, so `main:.claude/skills/...` is well-formed) all check out. The CI fail-open path is intact: `existsSync(AGENT_SKILL_STORE)` returns before `readCommittedOnMain` is ever reached, so a runner with no local `main` ref never hits `git show`.

Two correctness findings, both the same root cause in different clothes: **the guard enforces a proxy (the branch's *name*) for the property it actually needs (the mirror holds `main`'s *committed* content).**

---

### 🔴/🟠 Correctness

**🟠 1. The producer reads the working tree; the guard asserts against `main`'s committed blob. A dirty `main` poisons the mirror straight through the new guard — and the remediation hint loops, exactly as in #3001.**

`scripts/sync-agent-skills.mjs:210` gates on the branch *name*, then `:214` → `syncOneFile` → `:144-145` reads the canonical content from **disk** (`readFileSync(join(SKILLS_ROOT, rel))`). Meanwhile `scripts/tests/review-gate-mechanism.test.mjs:617` asserts the mirror equals `buildMirrorContent(readCommittedOnMain(rel))` — i.e. `git show main:<path>`. Those are two different sources, and `branchName === 'main'` does not make them equal.

Concrete failure:

1. Primary checkout is on `main` with an uncommitted edit to `.claude/skills/model-routing/SKILL.md` — a half-finished trivial edit, a conflict resolution mid-`git pull`, whatever. CLAUDE.md explicitly sanctions committing trivial changes and doing git-ignored artifact work in the primary checkout, so this is not an exotic state.
2. `npm run skills:sync` → `currentBranch()` returns `main` → guard passes → the mirror is written with the **uncommitted** content.
3. Every worktree on the box now fails `test:hooks` at `review-gate-mechanism.test.mjs:615`, because the mirror ≠ `git show main:`.
4. The new assertion message says: *"check out `main` in the PRIMARY checkout, pull latest, and run it from there."* Done in that same still-dirty checkout, that re-writes the identical poisoned content. The self-reinforcing remediation hint — the property #3001 calls out as the worst part of the original bug — survives.

The clean fix collapses both sources into one: have `syncOneFile`'s canonical read come from `git show main:<path>` as well (`assertNoBom` just needs `Buffer.from(raw)`). That is strictly simpler than what's here — it makes the two sides provably identical by construction, and it makes `assertOnMainBranch`/`currentBranch()` **unnecessary**, because syncing from any worktree would then produce `main`'s content by definition. If the branch check is kept instead, it needs to be a cleanliness check on the four mirrored paths (`git diff --quiet main -- .claude/skills/...`), not a name check.

---

**🟠 2. Adding or renaming a mirrored file now deadlocks its own branch — the guard hard-blocks the commit and the remediation it prints is impossible to perform.**

`scripts/tests/review-gate-mechanism.test.mjs:612-614` iterates the **branch's** `MIRRORED_FILES` and asserts `existsSync` in the shared mirror before the content compare. Add a fifth entry to `FILES` — precisely what was done on 2026-08-14 when `model-routing/` joined the mirror — and:

- the mirror cannot contain it (it only ever holds what `main` has synced), so `:614` fails with *"mirror is missing model-routing/SKILL.md — run npm run skills:sync from main"*;
- that command now **refuses**, because you are on the feature branch (`sync-agent-skills.mjs:189`);
- and running it from `main` writes nothing new, because `main`'s `FILES` doesn't list the file yet;
- the diff touches `scripts/**`, so `test:hooks` is in pre-commit scope (per #3001's own description) and **the commit is blocked**, in that worktree, with no legal remedy short of hand-copying into `~/.agents/skills/` or `--no-verify`.

Splitting it into two PRs does not help: a PR that lands the skill file without touching `FILES` passes, but the follow-up that adds the `FILES` entry hits the same wall. Before this change the (bad) escape hatch was `skills:sync` from the branch — this PR correctly closes it but does not replace it, so a previously-executed change class becomes uncommittable.

Suggested shape: drive the loop from what `main` committed rather than from the branch's list — skip (with a loud `console.log`, matching the existing fail-open style at `:609`) any `rel` whose `git show main:` lookup 404s, since a mirror provably cannot hold a file `main` does not yet have. That preserves the guard for every already-mirrored file while letting the file-adding branch commit.

---

### 🟡 Cleanup / chores

**🟡 3. The new guard's *wiring* is untested — deleting `sync-agent-skills.mjs:210` leaves the suite 26/26 green.**

`review-gate-mechanism.test.mjs:592-600` exercises `assertOnMainBranch` as a pure function only. Nothing in the repo calls `syncAgentSkills()` outside the `isDirectlyInvoked` block (`git grep syncAgentSkills` → 2 non-comment hits, both in `sync-agent-skills.mjs`), so the assertion that the guard *runs, and runs before any write*, is carried by a manual check in the PR body and by nothing mechanical. This is the repo's own recurring shape — the guard's output tested, its gate not — and the sibling `assertFilesNonEmpty` test at `:726` already carries the identical gap while its comment *asserts* the wiring ("syncAgentSkills() itself calls this before its loop") that nothing verifies. Cheap fix: give `syncAgentSkills(branchName = currentBranch())` an injectable branch and assert `assert.throws(() => syncAgentSkills('feat/x'))` — which also proves the throw happens before `syncOneFile` writes anything, closing both this and `assertFilesNonEmpty`'s gap in one line.

**🟡 4. `.clinerules/cline.md:34` still carries the exact wording this PR identifies as the cause — and it is the file the lane that caused the incident reads.**

> "So re-run `npm run skills:sync` after any change under `.claude/skills/pr-review-gate/` -- a per-machine step…"

`CLAUDE.md:464-482` and the `sync-agent-skills.mjs` header were both reworded; this second copy was not. The PR's own commentary (`sync-agent-skills.mjs:183-185`, `CLAUDE.md:469-472`) names *"re-run after any change under either directory"* as the phrasing that reads as "run it now, from this branch." That sentence is still live in the Cline rules file — i.e. in front of the exact agent lane that poisoned the mirror in #2999. The script now fails closed so this no longer *causes* poisoning, which is why this is a 🟡 rather than a 🟠, but it is a stale instruction the change made false, and it will send a Cline lane to a command that exits 1. (It also predates `model-routing/` joining the mirror and still names only `pr-review-gate/`.)

**🟡 5. The release-notes pair is neither updated nor explicitly declared not-applicable.**

Diff is `CLAUDE.md`, `scripts/sync-agent-skills.mjs`, `scripts/tests/review-gate-mechanism.test.mjs` — no `docs/release-notes-next.md` / `RELEASE_NOTES.md`, and the PR body says nothing about it. Before-shipping step 5 permits skipping for an internal chore with no operator-visible effect, but requires saying so rather than silently omitting — and this one *is* mildly operator-visible (a dev command that used to run now refuses). Either add the technical-register line or state the not-applicable in the PR body.

**🟡 6. `readCommittedOnMain` re-implements `readNormalized`'s CRLF collapse inline.**

`review-gate-mechanism.test.mjs:589` does `raw.replace(/\r\n/g, '\n')`; `scripts/lib/read-normalized.mjs` exists solely to own that transform and carries the #2291 rationale for it — and this very test file already imports it (`:42`), using it on the other side of the same `assert.equal` at `:616`. Two homes for one normalisation rule is how the two sides drift apart later. Export a `normalizeEol(text)` from `read-normalized.mjs` and have both `readNormalized` and this call it.

**🟡 7. The comparison basis moved outside the `test:hooks` cache key.**

The guard used to compare the mirror against files that are themselves inputs to the step (`.claude/skills/**`, `verify-cache.mjs:153`). It now compares against `main`'s committed tree, which is in no step's input hash. So: a mirrored-skill change merges, the operator pulls `main` in the primary checkout but does not re-sync — in every *other* worktree the files on disk are unchanged, `test:hooks` prints `[cached]`, and the guard reports green on exactly the event that staled the mirror. The primary checkout does catch it (its `.claude/skills/**` files changed on pull), so this is a narrowing rather than a new blind spot, but it is the same `#1847` runtime-read trap `verify-cache.mjs` documents ten times over. Adding `git rev-parse main:.claude/skills` (a tree hash, so it only moves when skills move on `main`) to the step's inputs would close it.

---

### House gates checked

- `Closes #3001`, plain line, outside any code span ✔
- Paired test present and non-vacuous — the drift test ran for real against this machine's mirror, and `MIRRORED_FILES`'s exact-list assert at `:556` still blocks the vacuous-iteration hole ✔ (wiring caveat in finding 3)
- Regression plan: n/a — small/localised, issue body + paired test is the spec ✔
- On-box acceptance: n/a — no hardware-provable behaviour ✔
- Release notes: **not addressed** — finding 5
- Derived artifacts: this PR touches no mirrored skill file, so no `skills:sync` is owed by it ✔
- `cast.json` locking, config knobs, `api-types.ts`, `BACKLOG.md` (issue is `type:chore`, so no row is owed): n/a ✔
- PR title `fix(scripts): …` — valid single-scope subject ✔

dudarenok-maker:
## PR review — pass 2 (head 94ada0de18aaf01d58de53b8b781e2582fcfa3c3, depth medium)

Re-review of the redesign. Verified locally in `C:\Claude\Projects\wt-3001-skills-mirror-main-guard`: worktree clean, HEAD matches the dispatch SHA, `node --test scripts/tests/review-gate-mechanism.test.mjs` → **27/27 green**, with the drift guard *running* for real (258ms, not the skip path) and the new wiring test writing only into `C:\Users\dudar\AppData\Local\Temp\skills-sync-wiring-*` (confirmed from its own `wrote …` line — the real `~/.agents/skills` is never a write target).

Deltas only below; anything pass 1 called solid and this round didn't move is not re-listed.

---

### Pass 1 dispositions

|---|---|---|
| 🟠 1 | producer/guard source asymmetry | **Resolved** — but see 🔴 A, which the new sourcing introduced |
| 🟠 2 | adding a mirrored file deadlocks its own branch | **Resolved** (traced end-to-end) |
| 🟡 3 | guard wiring untested | **Resolved**, residual gap in 🟡 B |
| 🟡 4 | `.clinerules/cline.md` stale wording | **Resolved** |
| 🟡 5 | release-notes pair | **Still open** |
| 🟡 6 | duplicate CRLF normalisation | **Resolved** |
| 🟡 7 | verify-cache staleness | Deferred to #3010 — legitimate use of the carve-out |

**🟠 1 — resolved.** `assertOnMainBranch`/`currentBranch` are gone repo-wide (`git grep` → zero hits, no orphaned import or dead test left behind). `syncAgentSkills` sources content only through `readCommittedOnMain` (`sync-agent-skills.mjs:239`), and the guard imports that same function (`review-gate-mechanism.test.mjs:621`) rather than reimplementing it. The dirty-`main`-checkout scenario is closed by construction: neither side reads `SKILLS_ROOT` from disk any more, so an uncommitted edit in the primary checkout can't reach either the write or the comparison. `syncOneFile` survives as a disk-reading fixture entry point only, and both it and the git path funnel through one `writeMirroredFile` (`:157`), so the BOM/frontmatter checks provably can't diverge between them.

**🟠 2 — resolved.** Traced the 5th-`FILES`-entry case (the real 2026-08-14 `model-routing/` shape) through every gate on a branch that adds `.claude/skills/foo/SKILL.md` + the `FILES` row + the exact-list row at `:556`: exact-list test passes (both updated); drift guard hits `raw === null` → `[skip]` → `continue` (`:622-625`), no `existsSync` assert; the cross-skill-link test at `:650` still reads *disk*, so it finds the new file and passes; `skills:sync` from the branch logs `[skip]` and exits 0. The branch commits. Post-merge the guard flips to asserting `existsSync` for that path, which is the intended red until each box re-syncs — and that re-sync is now legal from any worktree. Deadlock gone.

**🟡 3 — resolved.** `syncAgentSkills([])` is now asserted directly (`:570`), and the injected-`mirrorRoot` test (`:573-593`) exercises the real read-from-git → write path. Non-vacuity checked by mutation reasoning: nulling `readCommittedOnMain`, dropping the skip branch, pushing skips into `written`, or ignoring the `mirrorRoot` parameter each break `assert.deepEqual` at `:582`/`:583`. It never falls back to `MIRROR_ROOT` — that constant appears only as the default parameter at `:234`, and `writeMirroredFile` takes `destPath` fully-formed. Residual gap in 🟡 B.

**🟡 4 — resolved.** `.clinerules/cline.md:34-46` now says content comes from `main`'s committed blob, that it's safe from any branch or worktree, and — explicitly — *"don't run it before merging, since a not-yet-merged change isn't on `main` yet and will just be skipped."* No "run it when you make the change" framing survives, and it now names both mirrored skills instead of only `pr-review-gate/`. I checked for a third copy: `git grep skills:sync` hits only `CLAUDE.md`, `.clinerules/cline.md`, `package.json`, the script, and historical plan/spec docs. `.github/copilot-instructions.md` carries no copy. `docs/testing/agent-skill-resolution-probe.md:69` mentions the command but asserts nothing this change falsified.

**🟡 6 — resolved.** `normalizeEol` is exported once (`scripts/lib/read-normalized.mjs:23`) and `readNormalized` now delegates to it. The test's old inline `raw.replace(/\r\n/g, '\n')` is gone. Swept `scripts/**` for a third copy of the collapse: the only other literal `\r\n → \n` *collapses* are pre-existing and unrelated to this PR (`build-register-live-view.mjs:182`, `generate-release-notes-wiki.mjs:61`, `release-body.mjs:67`, `validate-pr-issue-link.mjs:62`); everything else is `split(/\r?\n/)`, a different transform. Nothing this PR touched holds a second copy.

**🟡 7 — the #3010 deferral is legitimate.** Read the issue. It names the decision owed rather than restating the finding, and gives three genuinely defensible outcomes (a new non-file input kind in `STEPS[]`, dropping `test:hooks` from caching, or accepting the residual gap) with real, differing costs — new schema surface on a map ~15 steps depend on, versus a fixed per-commit tax, versus a documented latent gap. That is a decision, not two implementations of one agreed behaviour, so it clears `findings-triage.md`'s carve-out rather than tripping the "it needs a judgement call" void reason. I confirmed the premise is accurate: `verify-cache.mjs:76-327`'s `test:hooks` inputs are entirely disk globs/`extraFiles` (`.claude/skills/**` included), and `main`'s committed tree appears in none of them.

**Merge of `origin/main` (#2999) — clean.** #2999 touched `.clinerules/cline.md` and `CLAUDE.md` (both also edited here) and `.claude/skills/pr-review-gate/SKILL.md` (a mirrored file), but none of `sync-agent-skills.mjs`, `read-normalized.mjs`, or `review-gate-mechanism.test.mjs`. `git diff 76aa7c6d HEAD` on the two shared docs shows *only* this branch's own hunks — no deletion of #2999's content, and no reintroduction of the wording 🟡 4 removed. No mishandled conflict.

---

### 🔴/🟠 Correctness

**🔴 A. `readCommittedOnMain` resolves the bare ref name `main`, which does not exist in a GitHub Actions PR checkout — `test:hooks` is RED on this head, right now, in the required `verify.yml` check.**

`scripts/sync-agent-skills.mjs:197` runs `git show main:<path>`, and `:203-205` swallows *every* failure into `return null`. In `actions/checkout`'s `pull_request` checkout the tree is detached at `refs/pull/N/merge`; `fetch-depth: 0` populates `refs/remotes/origin/*` but creates **no local branch**, and gitrevisions' resolution order for a bare name is `refs/<n>` → `refs/tags/<n>` → `refs/heads/<n>` → `refs/remotes/<n>` → `refs/remotes/<n>/HEAD` — `refs/remotes/origin/main` is on none of those. Verified against this repo: `git rev-parse --verify cla-signatures` (a remote-only branch) exits 1, and `git show cla-signatures:README.md` → `fatal: invalid object name`.

So in CI `main` doesn't resolve, all four `FILES` entries take the `catch`, and the bare catch relabels "the ref is absent" as "the path isn't on `main` yet". Not theoretical — this is the actual failure on run 34004604900, job *Lint, typecheck & fast checks* → *Hooks tests*, at `review-gate-mechanism.test.mjs:582`:

```
✖ syncAgentSkills writes a real, git-sourced mirror for a file on main and skips one that is not …
    actual:   [ 'pr-review-gate/SKILL.md', 'does/not/exist/on/main.md' ]
    expected: [ 'does/not/exist/on/main.md' ]
  1948 tests, 1 fail
```

`test:hooks` is in scope for this PR (its globs include `scripts/**/*.mjs`), and the job is a required status check, so this blocks merge outright. It is invisible locally *because* the primary checkout has a real `refs/heads/main` — the PR body's "`npm run verify:fast:branch` green" is true and still misses it.

The second-order half is worse than the red test, and outlives it. The drift guard at `:620-625` takes the same `null` and `continue`s on **every** file — so on any machine where `main` doesn't resolve, the guard iterates nothing and reports green, having compared no file at all. In CI that's currently masked by the `existsSync(AGENT_SKILL_STORE)` early return at `:616`, but the mechanism is live: this is `reviewer-brief.md` shape 1 ("a guard that fails open on absent evidence") plus shape 6 ("success reported while doing nothing") — and `skills:sync` in that same state prints `Wrote 0, skipped 4 (not yet on 'main')` and exits 0, which is verbatim the outcome `assertFilesNonEmpty`'s own comment at `:208-215` exists to prevent. Before this PR the same absence threw `ENOENT` out of `readFileSync` and failed loudly.

The two causes must be told apart. Resolve the ref once, up front (`git rev-parse --verify` over `main`, then `origin/main`), **fail loudly if neither resolves**, and keep the `null` return strictly for a per-path `git show` miss on a ref that *did* resolve. That fixes CI in the same move (`origin/main` is present under `fetch-depth: 0`) and closes the fail-open. This repo already has the precedent for "don't trust a branch ref by name": `check-onbox-register.mjs`'s `resolveBaselineTexts` was fixed to never reference `origin/main` directly, pinned by `check-onbox-register.test.mjs:2303`.

---

### 🟡 Cleanup / chores

**🟡 B. The one property the PR is named for — "reads `main`, not the current branch" — is pinned by nothing. Mutating `main:` → `HEAD:` leaves all 27 tests green.**

`sync-agent-skills.mjs:197` is reached by the production write (`:239`), by the wiring test's oracle (`review-gate-mechanism.test.mjs:587`), and by the drift guard (`:621`). Because producer and both oracles call the *same* function, the equality assertions at `:585` and `:628` are invariant to anything inside it — this is the "oracle shares traversal with the code" shape, and it is the direct cost of the (correct) decision to collapse both sides onto one reader. Change `main:${gitPath}` to `HEAD:${gitPath}` and: the wiring test still passes (this branch touches no mirrored skill, so `HEAD`'s blobs equal `main`'s), the drift guard still passes for the same reason, and #3001's exact regression — a branch's own unmerged edit becoming the canonical source — is back with a fully green suite.

The seam to close it already exists and no caller uses it: `readCommittedOnMain(rel, repoRoot = REPO_ROOT)` (`:194`) — the `repoRoot` parameter is passed by nobody today, production or test. A direct test can build a throwaway repo (`scrubGitEnvForThrowawayRepo()`, `scripts/git-env.mjs:98`, exists for exactly this), commit content X at `.claude/skills/x/SKILL.md` on `main`, branch off and commit content Y, then assert `readCommittedOnMain('x/SKILL.md', tmpRepo)` returns X. The same fixture pins 🔴 A's fallback chain and its loud failure when no ref resolves.

**🟡 C. The PR body describes a design this head no longer contains, including a test-plan line claiming a manual verification of behaviour that was deleted.**

The body still reads *"`sync-agent-skills.mjs` now refuses to run outside `main` (mechanically enforcing the post-merge-only rule, via the new `assertOnMainBranch`)"* and *"the drift guard compares … instead of the current worktree's disk copy"* as if only the guard moved — and its Test plan asserts *"Manually verified `node scripts/sync-agent-skills.mjs` refuses to run from this feature branch with the new error message."* `assertOnMainBranch` no longer exists, the script refuses nothing, and running it from this branch now succeeds. The body is what a merger reads and what the incident record inherits; it currently documents the design pass 1 rejected. Rewrite Summary + Test plan against `94ada0de`.

**🟡 5 (carried over, still open). Release notes are neither present nor declared not-applicable.**

The file set is unchanged from pass 1 — no `docs/release-notes-next.md`, no `RELEASE_NOTES.md` — and the PR body still says nothing about the omission. Before-shipping step 5 permits skipping for an internal chore with no operator-visible effect but requires *saying so*, and the omission is silent. On the merits the skip is now weaker, not stronger, than at pass 1: the shipped delta versus `main` is that `skills:sync` changes where it sources content and becomes safe to run from any branch, with an added skip path — operator-facing enough that this PR reworded **both** `CLAUDE.md` and `.clinerules/cline.md` to teach it, and #2999 (the same `fix(scripts)`/dev-tooling shape, merged two commits ago) did land entries in both notes files. Add the technical-register line, or state the not-applicable in the body.

---

### House gates re-checked

- `Closes #3001` — present, plain line, outside any code span ✔
- Paired tests present, run for real, and non-vacuous against the mutations listed under 🟡 3 — with the `main`-vs-`HEAD` exception in 🟡 B ✔/⚠
- Incidental findings: 🟡 7 filed as #3010 with the decision named (legitimate); 🟡 5 neither fixed nor declared ✖
- Release notes — **not addressed**, finding 🟡 5
- Regression plan, on-box acceptance, `cast.json` locking, config knobs, `api-types.ts`, `BACKLOG.md` (issue is `type:chore`) — n/a, unchanged from pass 1 ✔
- Derived artifacts: this PR touches no mirrored skill file, so it owes no `skills:sync` of its own ✔
- PR title `fix(scripts): …` — valid single-scope subject ✔
- Cloud `verify.yml` — **red on this head**, finding 🔴 A ✖

dudarenok-maker:
## PR review — pass 3 (head d4824a5d3c8a1b4a6d10efa0fa2f16be961d7ef7, depth medium)

Re-review round 2 (the loop's last automatic round). Verified in `C:\Claude\Projects\wt-3001-skills-mirror-main-guard`: worktree clean and HEAD matching the dispatch SHA both before and after this pass, `node --test scripts/tests/review-gate-mechanism.test.mjs` → **28/28 green** with the new `resolveMainSha` test genuinely spawning git (1412 ms, three real throwaway repos), `npx eslint` on both changed script files → exit 0, and no leaked `resolve-main-sha-*` / `skills-sync-wiring-*` fixtures in `%TEMP%` after repeated runs (including the failing mutation runs below).

Deltas only. Anything pass 2 called solid and this round didn't move is not re-listed.

---

### Pass 2 dispositions

|---|---|---|
| 🔴 A | `main` unresolvable in a CI PR checkout → every file silently skipped | **Resolved — empirically, in the environment that broke** |
| 🟡 B | the `main`-vs-`HEAD` property pinned by nothing | **Still open** — re-confirmed by mutation, see below |
| 🟡 C | PR body described the deleted `assertOnMainBranch` design | **Resolved** |
| 🟡 5 | release-notes pair | **Half resolved** — see 🟡 5 (carried) |

**🔴 A — resolved, and proven rather than argued.** Traced the CI-shaped path end to end and then checked it against the real run:

- `verify.yml`'s *Lint, typecheck & fast checks* job (the one that runs `test:hooks`) checks out at `fetch-depth: 0` (`.github/workflows/verify.yml:162`), so `refs/remotes/origin/*` is populated while `actions/checkout` leaves the tree detached with no local `main` — exactly the shape `resolveMainSha` is written for.
- On this head that job is **green**, and its *Hooks tests* step ran for real rather than being scope-skipped (job 101411771326, step 8 → `success`; steps 5/6/7/9/10/11/13 show `skipped`, so the scope filter is discriminating, not blanket-passing). Its log reports `ℹ tests 1949 / ℹ pass 1949`, which independently confirms the body's `1949/1949` claim **and** confirms the `origin/main` fallback fired for real — 1948 of those tests passed at the broken head too, so the delta is precisely the leg that used to be red.
- Structurally, `readCommittedOnMain` no longer resolves a ref at all: `git grep 'main:'` over `scripts/` returns only comments (`sync-agent-skills.mjs:32`, `:154`, `:193`) — the sole surviving `git show` argument is `` `${mainSha}:${gitPath}` `` (`sync-agent-skills.mjs:241`), and every call site passes a `resolveMainSha`-produced SHA (`:287` production, `review-gate-mechanism.test.mjs:591` and `:699`). The drift guard's `resolveMainSha()` correctly sits *after* the `existsSync(AGENT_SKILL_STORE)` early return (`:693-697`), so a store-less CI runner still never reaches it.

I did not take the new test's word for its own non-vacuity — I ran four mutations against the real module and restored the tree after each:

| Mutation | Result |
|---|---|
| **B** — drop `'origin/main'` from the ref list (`sync-agent-skills.mjs:207`) | ✅ **red**, at `resolveMainSha` (`:220`) — the fallback is genuinely pinned by case 2 |
| **C** — replace the `throw` with `return 'main'` (`:220`) | ✅ **red**: `Missing expected exception: resolveMainSha must fail loudly…` |
| **D** — have `syncAgentSkills` catch and return `{ written: [], skipped: [...files] }` (`:283`) | ✅ **red**: `Missing expected exception: syncAgentSkills must fail loudly, not silently skip every file…` — so the second `assert.throws` at `:662` is independently load-bearing, not a restatement of the first |
| **A** — `` `${mainSha}:` `` → `` `HEAD:` `` (`:241`) | ❌ **28/28 still green** — see 🟡 B |

So three of the four properties the round-2 fix introduced are real, pinned, and mutation-proven. The fixture is also correctly scoped to this repo's conventions: it matches `scripts/tests/bump-version.test.mjs`'s existing throwaway-repo pattern exactly (`scrubGitEnvForThrowawayRepo()`, `git init`, `git config user.email/user.name`, real commits) rather than inventing a new one, and `scrubGitEnvForThrowawayRepo` is the right helper here over `scrubGitEnv` per `git-env.mjs:86-97`. Three repos is not overkill — each case is a distinct ref topology and none is reachable from the others.

**🟡 C — resolved.** Re-read the body against the diff. `assertOnMainBranch` and "refuses to run outside `main`" are gone; it now says `git show <sha>:<path>`, describes `resolveMainSha`'s prefer-local-then-`origin/main` chain and its loud throw, and the Test-plan lines are accurate (28/28 verified locally, 1949/1949 verified from the CI log above). No stale claims left.

---

### 🔴/🟠 Correctness

**None.** No new correctness bug in the round-2 delta, and 🔴 A is closed. Adversarial checks that came back clean, so they're on record rather than silently assumed:

- **Resolving the SHA once per run instead of per file is a strict improvement, not a regression.** Previously each `git show main:` re-resolved the ref independently, so a concurrent `git fetch`/merge moving `main` mid-run could produce a mirror mixing blobs from two commits. Pinning one SHA makes the write atomic with respect to a single commit. Cost is one extra `rev-parse` (~10 ms) against four `git show`s that each resolved the ref anyway.
- **Preferring local `main` over `origin/main` is the right order**, not a hazard: on a dev box the operator's own `main` is the intended source (CLAUDE.md already documents the stale-local-`main` caveat), and CI has no local `main` at all, so the two never compete.
- **`writeMirroredFile`'s `sourceLabel` still produces sensible errors.** `` `${mainSha}:.claude/skills/${rel}` `` (`:293`) is strictly *more* useful than the old `main:` — it stays reproducible after `main` moves, and it pastes straight back into `git show`. (Its doc comment did not follow — 🟡 E.)
- Injection surface is unchanged (`execFileSync` with array args, no shell; `rel` from a hardcoded `FILES`), and `mainSha.slice(0, 8)` is safe against `rev-parse`'s full 40-char output.

---

### 🟡 Cleanup / chores

**🟡 B (carried from pass 2, still open — and now confirmed by mutation rather than by reasoning). Reverting `readCommittedOnMain` to read `HEAD:` leaves all 28 tests green, so #3001's exact regression can still be reintroduced with a fully green suite.**

`sync-agent-skills.mjs:241`. I ran it: changing `` `${mainSha}:${gitPath}` `` to `` `HEAD:${gitPath}` `` → `ℹ tests 28 / ℹ pass 28 / ℹ fail 0`. The wiring test (`review-gate-mechanism.test.mjs:575`) and the drift guard (`:672`) both stay green because this branch touches no mirrored skill file, so `HEAD`'s blobs equal `main`'s; and the wiring test's oracle calls the same mutated function as the producer (`:591` vs `:287`), so its `assert.equal` is invariant to anything inside it regardless.

The round-2 test does not close this, and it is worth being precise about why: it exercises `resolveMainSha` only. Case 3's `syncAgentSkills(['pr-review-gate/SKILL.md'], …, noMain)` (`:662`) throws at `resolveMainSha` *before* any `readCommittedOnMain` call, and all three fixtures commit only `file.txt` — none ever creates `.claude/skills/**`. So `readCommittedOnMain`'s `repoRoot` parameter is *still* passed by no test that actually reads a blob, which is the same gap pass 2 named.

The fix is now cheap, because round 2 already built the fixture that needs it — it just stops one step short. In the existing `ciCheckout` (or `origin`) repo: commit `.claude/skills/x/SKILL.md` with content `X` on `main`, branch off and commit content `Y`, check the branch out, then assert `readCommittedOnMain('x/SKILL.md', resolveMainSha(repo), repo)` returns `X`. That kills the `HEAD:` mutant and is the one assertion that pins the property this PR is named for. (Minor, same fixture: the `assert.throws` at `:633-636` passes a **string** in the `error` position, where Node treats it as the assertion *message*, not a matcher — so it only checks "threw something". Every other `assert.throws` in this file uses a regex there; a regex would keep it consistent and prevent a future reader mistaking it for validation. It is genuinely load-bearing for case 2's non-vacuity — mutation B proved that — so this is shape, not substance.)

**🟡 D (new). `readCommittedOnMain`'s `mainSha` is a required positional with no validation, so a falsy or garbage value folds straight back into the silent per-path skip this PR exists to eliminate.**

`sync-agent-skills.mjs:238-249`. `mainSha` has no default and is never checked; the `catch` at `:247` returns `null` for *every* failure. A call of the pre-round-2 shape `readCommittedOnMain(rel)` — which is exactly what this function's signature looked like two commits ago (`94ada0de`), and what pass 2's own review comment quotes — yields `git show undefined:.claude/skills/…` → `catch` → `null` → "not on `main` yet". In `syncAgentSkills` that is unreachable because `resolveMainSha` throws first, but the drift guard calls it too (`review-gate-mechanism.test.mjs:699`), and #3010 is already filed to do further work in exactly this area. A `null` there makes the guard iterate nothing and report green — `reviewer-brief.md` shapes 1 and 6, one level below where round 2 closed them. One line at the top of the function (`if (!mainSha) throw new Error(...)`) restores the same "unresolvable ref ≠ missing path" separation `resolveMainSha` establishes for the caller.

**🟡 E (new). Two comments the round-2 change made false.**

- `sync-agent-skills.mjs:154` — `writeMirroredFile`'s JSDoc still says `sourceLabel` is "a `main:<path>` git ref for syncAgentSkills". It is now `` `<40-hex-sha>:<path>` `` (`:293`). This is the doc for the parameter the change altered.
- `CLAUDE.md:487` — "the sync reads whatever `main` this machine's git has, **not `origin/main` directly**." Adding the `origin/main` fallback is the entire round-2 delta, so this is now false in precisely the case the fallback exists for. It remains true for its stated dev-box audience (a local `main` is always preferred when present), so the operational advice is not wrong — but the absolute phrasing now denies a behaviour the same PR added, in a paragraph this PR already edits.
- (`sync-agent-skills.mjs:32`'s "via `git show main:<path>`" is loose but still conceptually accurate — worth a word while the file is open, not a defect on its own. `:193` is inside `resolveMainSha`'s own description of the bug and is correct as written.)

**🟡 5 (carried from passes 1 and 2, now half open). `docs/release-notes-next.md` landed and is accurate; the `RELEASE_NOTES.md` half is still neither present nor declared.**

The technical-register entry is good — I checked it line by line against the diff and every claim holds (the disk-sourced original, the `test:hooks` drift guard comparing the same disk copy, the once-per-run resolution with the `origin/main` fallback, the CI catch on this PR's own head, the skip-if-unmerged path, the guard importing the producer's reader, and the #3010 filing). Placement is sensible too: it sits under `## 🧪 Test gates` → `### Test harness & suite hygiene`, directly after the ops-2997/#2999 commit-gate entry it is a sibling of. (Note for the record: that is *not* the "Under the hood" section — the anchor entry is right, the section label used in the hand-off was not.)

What is still missing is any disposition for `RELEASE_NOTES.md`. `gh pr view 3008 --json body` contains no occurrence of "release note", "RELEASE_NOTES", "not applicable", or "n/a" — so the omission is silent, which is the one thing before-shipping step 5 forbids ("say so explicitly rather than silently omitting"). The skip is defensible **on the merits** — no end user of the audiobook app is affected by an agent-tooling sync script — but two things weigh against calling it obvious, and both are checkable:

1. This repo's `RELEASE_NOTES.md` demonstrably *does* carry internal dev-tooling entries in brand voice. Its current in-progress section includes "Running Castwright's test suite while several worktrees are busy no longer risks a multi-hour pileup" and "CI's weekly quarantine check now catches when it can't read the register" — neither is user-facing in any stronger sense than this change.
2. The precedent is not distant: the entry this PR's own note was placed immediately *after* is ops-2997/#2999, and that change landed **both** halves — its user-facing line is "Saving your work is instant again — the checks moved to where they belong."

So this is a judgement the author is entitled to make, but it has to be **written down** — one sentence in the body ("`RELEASE_NOTES.md`: n/a — internal agent-tooling, no end-user-visible effect"), or the brand-voice line. There is no mechanical gate on the pair per PR (`grep` over `package.json` and `.github/workflows/` finds release-notes validation only in `release.yml`, at tag time), so nothing else will catch it before a release cut.

---

### House gates re-checked

- `Closes #3001` — present, plain line, outside any code span ✔
- Paired tests present, run for real, and mutation-proven on three of four properties — the `main`-vs-`HEAD` exception is 🟡 B ✔/⚠
- Cloud `verify.yml` — **green on this head**, including the previously-red *Hooks tests* leg (1949/1949) ✔
- Lint — `npx eslint` on both changed script files, exit 0 ✔ (`npm run lint` is `eslint . --max-warnings 0`; `format:check` is not in the PR battery, so the long template literals at `:281`/`:289` are not gated)
- Test-fixture temp state — no leaked git repos in `%TEMP%` after repeated + failing runs ✔
- Incidental findings: #3010 remains a legitimate design-pass deferral (unchanged from pass 2) ✔
- Release notes — technical register ✔, user-facing half undeclared ✖ (🟡 5)
- Regression plan, on-box acceptance, `cast.json` locking, config knobs, `api-types.ts`, `BACKLOG.md` (`type:chore`) — n/a, unchanged ✔
- Derived artifacts: touches no `.claude/skills/**` file, so it owes no `skills:sync` of its own ✔
- PR title `fix(scripts): …` — valid single-scope subject; depth `medium` correctly derived ✔

---

### Verdict

**Mergeable — no blocking correctness issue.** The 🔴 from round 1 is closed and proven closed in the environment that broke it, not merely reasoned about. Nothing in the round-2 delta introduced a new correctness problem.

Four 🟡 cleanup findings remain (B, D, E, and the carried 5). Per `findings-triage.md` they are fixed in this round but do **not** re-trigger another gate pass — 🟡 B is the one worth actually doing rather than waiving, since it is the second round it has survived and the fixture that closes it already exists in this diff.

dudarenok-maker:
## Post-round-2 cleanup (head 27964b3d)

Round 2's re-review verdict was **mergeable — no blocking correctness issue**, with 4 🟡 cleanup findings remaining. None are correctness bugs, so per the re-review trigger they don't require another automatic round — fixed here instead:

- **🟡 B** (the fix's own property was pinned by nothing — mutating the git-show argument back to the literal `HEAD:` left every test green): added a fixture where a checked-out branch's HEAD genuinely diverges from `main` (a feature branch with its own committed-but-unmerged edit to a mirrored skill file — the actual #3001/#2999 shape). Manually verified: reintroducing the `HEAD:` mutation now reddens this test; reverted after confirming.
- **🟡 D** (`readCommittedOnMain`'s `mainSha` param was unvalidated): now throws if called without a resolved SHA, rather than silently building `git show undefined:<path>` and folding that into the same false "not on main yet" skip.
- **🟡 E** (two stale comments): fixed `writeMirroredFile`'s docstring (still described a literal `main:<path>` ref) and `CLAUDE.md`'s remediation text (denied the `origin/main` fallback this PR added).
- **🟡 5** (release notes not stated in the PR body): the PR body now has its own "Release notes" section — `docs/release-notes-next.md` entry added, `RELEASE_NOTES.md` explicitly called not-applicable (internal dev/agent-tooling, no end-user effect).

`node --test scripts/tests/review-gate-mechanism.test.mjs`: 29/29. `npm run test:hooks`: 1950/1950.

Issue #2262, opened 2026-08-11
Flaky: script-review.test.ts cancellation test needs a retry under full-suite contention


dudarenok-maker:
## What

`server/src/routes/script-review.test.ts` :: *"cancellation (fs-58 follow-up #1481) > cancel immediately removes the job from the registry, so a same-scope retry starts fresh instead of joining the doomed job"* fails on its first attempt and passes on retry during a full `npm run test:server` run.

It is **not quarantined** and it does not fail the suite — vitest's `retry:1` absorbs it — so it is currently invisible except as a `[retry-hazard]` warning line.

## Evidence

Two independent sightings on 2026-08-11, on branch `chore/server-lane-a-cast-store`, by two different agents that were not looking for it:

1. During #2228's implementation run — reported as "one confirmed-unrelated flake ... verified passing in isolation".
2. During #2239's fix-round commit (`dbcf36c5`), full `test:server`, 505 files / 7078 tests passed:

```
[retry-hazard] server/src/routes/script-review.test.ts :: "cancellation (fs-58 follow-up #1481) > cancel
immediately removes the job from the registry, so a same-scope retry starts fresh instead of joining the
doomed job" failed on its first attempt and only passed after 1 retry(ies).
```

Both runs were on a box with concurrent sessions, which is the usual precondition for this repo's contention flakes.

## Why it matters more than a normal flake

The retry-hazard warning names the specific risk: **`retry:1` can turn a genuine red-phase failure green** (#2028, `ops-46`). This test asserts on the job **registry** — module-level mutable state — which is exactly the shape #2028 warns about. So the two readings are not equally likely to be benign:

- benign: tempdir/timing contention under parallel forks, same class as #2226 / #2235;
- not benign: a real ordering bug in cancel-then-retry that only loses the race under load, with `retry:1` masking it.

Nobody has distinguished these yet. Both prior sightings stopped at "passes in isolation", which is consistent with **either**.

## Acceptance

- Re-run the file with `--retry=0` under deliberate contention (e.g. alongside a full `test:server`) enough times to establish a rate.
- State which of the two readings it is, with the evidence.
- If contention: either stabilise it, or add a register row per `docs/testing/flaky-register.md` and route through `quarantinedIt` if it starts gating. **Never** a raw `it.skipIf(process.env.CI)`.
- If a real ordering bug: a regression test that fails at `--retry=0` before the fix and passes after.

## Key files

[...]

Issue #2215, opened 2026-08-07
Flaky: book-state-preserve-voices '#1981 race: stale cast PUT vs concurrent /assign'


dudarenok-maker:
## What

`server/src/routes/book-state-preserve-voices.test.ts` — the `#1981 race: stale cast PUT vs concurrent /assign` case fails intermittently when run with retries disabled.

## Evidence it is pre-existing, not caused by #2166 / PR #2202

Verified on `main` itself (`f9f8a84a`), not on the feature branch — the whole file, `--retry=0`, three consecutive runs:

- run 1: 7 passed
- run 2: **1 failed | 6 passed**
- run 3: 7 passed

So roughly 1-in-3 on this box, on `main`, with no branch changes present. It normally passes in CI because vitest's configured retries mask it; it surfaced only when a pre-commit `[retry-hazard]` warning prompted a `--retry=0` re-run.

## Likely cause

The case waits on a fixed ~20 ms wall-clock delay to order a stale PUT against a concurrent `/assign`. A wall-clock wait is not a synchronisation primitive — under load (parallel vitest workers, a concurrent GPU job, a cold import) the two requests can interleave the other way.

## Suggested fix

Replace the timing assumption with a deterministic ordering signal rather than lengthening the sleep, which only makes the flake rarer and the suite slower. If that is not practical, route it through `quarantinedIt` (`server/src/test-utils/quarantine.ts`) into the non-gating lane and log it in `docs/testing/flaky-register.md`, per CLAUDE.md — never a raw `it.skipIf(process.env.CI)`.

## Key files

- `server/src/routes/book-state-preserve-voices.test.ts` — the `#1981 race` case and its wall-clock wait
- `server/src/test-utils/quarantine.ts`, `docs/testing/flaky-register.md` — the quarantine route

PR #3009, opened 2026-09-05
fix(server): poll for the intercepted read instead of sleeping 300ms

### LINKED ISSUE #3007: analysis.fresh-cast-lock.test.ts races a fixed 300ms sleep and fails the commit gate under load
## What

`server/src/routes/analysis.fresh-cast-lock.test.ts` fails under CPU load. It waits a **fixed 300 ms** and then asserts that a background interception has already happened:

```js
// Let add-alias acquire the cast lock and reach (and get stuck
// behind) its intercepted in-lock read. Generous — this file's first
// supertest request pays a cold Express/module-init cost a warmer
// file (many prior requests already run) wouldn't.
await new Promise((r) => setTimeout(r, 300));
expect(intercepted).toBe(true);
```

`server/src/routes/analysis.fresh-cast-lock.test.ts:290-291`

The comment already names the hazard — the wait has to cover a cold Express/module-init plus a lock acquisition — and then picks a constant anyway. On a box running several worktrees, 300 ms is not enough, so `intercepted` is still `false` and the assertion fails with `expected false to be true`.

## Evidence

| Run | Conditions | Outcome |
|---|---|---|
| 2026-09-05, contended (~37 procs) | full `test:server` | **failed**, 34479 ms |
| 2026-09-05, contended, retry attempt | full `test:server` | **failed**, 39117 ms |
| 2026-09-06, moderate load | full `test:server` (8228 tests) | **failed** — the only failure of the run |
| 2026-09-05, quiet box | full `test:server` | passed |
| 2026-09-06, isolation, `--retry=0` | single file | passed, 25.23 s |

Vitest's `retry: 1` already fires on it and still loses — the retried attempt hits the same starved-CPU window.

It is **not** in `docs/testing/flaky-register.md` and is **not** routed through `quarantinedIt`, so it gates every commit that `--changed` selects it for.

## Why this is a defect, not a flake to quarantine

The invariant it pins is real and worth gating — #1981 Task 11, a "Start fresh" `cast.json` delete racing a concurrent cast writer. Quarantining would stop that gating to hide a defect in the *test*, not in the code under test.

The test does not actually need a duration. It needs the condition. `intercepted` is a variable the test itself sets from its own interception hook, so it can be polled.

## The fix

Replace the fixed sleep with a **poll until `intercepted` is true, bounded by a timeout** — deadline generously above the observed cold-start cost, polling at a short interval, failing with a clear message if the deadline passes. A fast machine then proceeds in milliseconds instead of always paying 300 ms, and a loaded one waits as long as it actually needs.

Check the rest of the file for the same shape: if there are other fixed sleeps before assertions, they have the same bug and are in scope.

## Acceptance

- No fixed-duration sleep gates an assertion in this file.
- The test passes in isolation **and** as part of a full `test:server` run on a loaded box.
- Deliberately overrunning the deadline still fails, with a message naming what never happened — the timeout path must be reachable, not decorative.

## Benefit

*Benefit (technical):* removes a race that fails the commit gate for **any** lane whose diff `--changed` selects this file, on a box that routinely runs several worktrees at once. It cost one full 1306-second `test:server` run today alone.

Found while committing #2912's salvaged work (PR for #2054); unrelated to that change, surfaced rather than folded in per CLAUDE.md's commit-gate triage rule.

dudarenok-maker:
## Summary

`server/src/routes/analysis.fresh-cast-lock.test.ts` waited a **fixed 300 ms** and then asserted a background interception had already happened. A fixed sleep before an assertion is a race — and the comment already named the hazard (the wait must cover a cold Express/module-init *plus* a lock acquisition) before picking a constant anyway.

Under load 300 ms isn't enough, so `intercepted` was still `false` and the run died on `expected false to be true`.

| Run | Conditions | Outcome |
|---|---|---|
| 2026-09-05 contended | full `test:server` | **failed**, 34479 ms |
| 2026-09-05 contended, retry | full `test:server` | **failed**, 39117 ms |
| 2026-09-06 | full `test:server`, 8228 tests | **failed** — sole failure of the run, which burned 1306 s |
| quiet box | full `test:server` | passed |
| isolation, `--retry=0` | single file | passed |

vitest's `retry: 1` fires and still loses — the retried attempt hits the same starved-CPU window.

**The fix polls the condition** rather than guessing a duration, failing with a message naming what never happened rather than a bare boolean. Deliberately **not** a larger constant: that reproduces the same bug with a wider window.

`vi.waitFor` is this repo's established idiom for exactly this — the flaky register's graduated entry for **#2262** (`script-review.test.ts`) records the identical bug fixed the same way.

## ⚠️ What this PR does NOT do — please read before merging

**This test cannot currently detect the regression it exists for, and this PR does not change that.** Filed as **#3022**.

Verified twice, independently: removing the `withCastLock` wrapper at `analysis.ts:3660` — the #1981 Task 11 invariant this file is named for — leaves the test **fully green**. Not even the weaker `expect(resAlias!.status).toBe(200)` fires.

The cause is structural and upstream. `readPriorCastForMerge` (`analysis.ts:220`) opens with `withCastLock(bookDir, …)`, and it is called at `analysis.ts:3534` for **any book with a `recordRef.bookDir`** — strictly *before* the `requestedFresh` block. `add-alias` is parked inside that same lock, so the job cannot pass that call until add-alias releases — by which point its write has already landed. The ordering the delete's own lock is meant to prove is forced upstream whether or not that delete is wrapped.

That is independent of how the test waits. It held for the original fixed sleeps and for every synchronisation form tried here, so **it is not caused by this PR** — this branch merely ran the mutation that exposed it.

**So: what merges here is a real fix to a real flake, and a test that is honest about gating less than its name implies.** #3022 names the three defensible outcomes and the decision owed. The test's comments carry that warning at the core assertion.

## Review findings and their resolutions

**Pass 1** (depth medium, head `1e7027fd`) — five findings:

| # | Finding | Resolution |
|---|---|---|
| 1 | 🟠 the poll watched a flag set at interceptor **entry**, *before* `await actual.readJson(path)` — strictly **less** slack than the 300 ms sleep it replaced | `1cd2cef6` — signal after the real read, raced against a deadline, plus `expect(intercepted).toBe(true)`. Follows the `signalIntercepted`/`interceptedSignal` precedent from #2215/#2232 |
| 2 | the 400 ms settle sleep | `4512e253` — replaced with a `vi.waitFor` poll. **This turned out to be a regression**; see pass 2 finding 1 |
| 3 | 🟠 the 100 ms head start should become a real happens-before edge | **Deferred to #3022, with evidence** — the suggested repair deadlocks |
| 4 | the 5 s deadline duplicated between wait and error message | `dd755722` — hoisted to `SYNC_WAIT_TIMEOUT_MS` |
| 5 | the file's own `it()` budget is too tight | `245cfe4d`, **corrected** by `a2045e33` — see below |

Finding 1 was a genuine bug in my original fix, and the repo had already solved it next door.

**Pass 2** (depth medium, head `245cfe4d`) — no 🔴, no 🟠, four 🟡. All four fixed in `a2045e33`:

1. **The core assertion could no longer fail.** Pass 1's finding-2 fix replaced the settle with a poll for absence — so the poll returned at the first instant of absence and the assertion then re-checked absence. Tautological. Worse, a genuine resurrection would have surfaced as the poll's own `cast.json still exists; delete has not completed` message, blaming contention for a resurrection. Now: poll for the delete, **then** settle, then sample. **Verified by mutation** — a writer scheduled inside the settle window fails the test at `expect(castExistsAfterRace).toBe(false)` with `expected true to be false`; tree restored byte-identical.
2. **My finding-5 measurement did not reproduce.** See below — this is the one worth reading.
3. **The 5 s deadline timer was neither cleared nor `unref`'d**, holding a live handle past the normal path — the hazard `workspace/file-lock.ts:239-247` explicitly names.
4. **A comment credited "PR #2232 review, finding 1"** — copied verbatim from the sibling file where it *is* correct. It also argued from the lock-removed mutation, which per #3022 this file cannot detect.

### Finding 5 / pass 2 finding 2 — I got the measurement wrong

`245cfe4d` raised the budget 30 s → 120 s citing a test body of **"10.4 s then 22.0 s across two back-to-back isolation runs on an UNLOADED box"**.

**The box was not unloaded.** Three sibling worktrees were running `node`/`python` at the time. I asserted a control condition without checking it, and then wrote it into a permanent in-file comment.

Five quiet isolation runs measure **3.56 / 4.85 / 4.74 / 6.28 / 4.54 s**. The honest body is ~5 s, so the old 30 s already had ~5× headroom — not the 1.4× I claimed. (#3007's own 25.23 s figure for this shape came from a contended box too.)

**The raise still stands, but only on the half of the evidence that survives**: two of six runs under `flake-repro.mjs --cpu-load` blew 30 s outright. `a2045e33` drops it to **60 s**, this suite's house norm for the class (51 uses vs 17 of `30_000`), because suite-wide `retry: 1` means a genuine wedge costs twice the budget in fast-lane wall clock.

Still not the "wider constant" mistake this PR fixes elsewhere: that sleep gated an *assertion* on a timing guess; this only bounds a hang, and the internal bounded waits (`SYNC_WAIT_TIMEOUT_MS`, `withKeyLock`'s 10 s per #2260) are the real detectors and held across all six contended runs.

### Finding 3 — why it is filed rather than fixed

The suggested repair was tried and **deadlocks**. The job cannot reach `clearAnalysisCache` (`analysis.ts:3639`) until add-alias releases the lock, but the release was gated on that signal. Observed: `clearAnalysisCache never completed within 5000ms`, plus a real `LockAcquisitionTimeoutError`.

I initially triaged this as *not* a design pass, on the grounds that only the implementation was open. The deadlock evidence reversed that. The in-file comment claiming no duration-free synchronisation point *exists* has also been softened to *available* — a lock-queue-entry edge (`file-lock.ts:232`) would not deadlock; it simply isn't exposed to a test today, and adding that seam is part of #3022's decision.

## Test plan

- **Isolation**, `--retry=0`: passes. Bodies 5.81 s / 6.42 s / 6.28 s post-fix.
- **Full `test:server`**: green in cloud CI at `245cfe4d` (20/20). Re-running at `a2045e33`.
- **The poll's timeout branch is reachable** — commenting out the line that sets `intercepted` fails with `add-alias never reached its intercepted in-lock read within 5000ms`. Restored and verified by `git diff`.
- **The core assertion is reachable** — the resurrection mutation above. This is the check that pass 2 showed was missing.
- **Under contention** via `flake-repro.mjs --cpu-load`: 4 of 6 passed (24.07 s, 19.23 s, 11.58 s, 26.46 s); the other 2 hit the `it()` budget now raised.
- **The mutation that matters** — replacing the delete's `withCastLock` with a passthrough of identical arity: still `Tests 1 passed (1)`. That is the #3022 disclosure, run rather than assumed.
- `tsc -p .` exit 0, `eslint` exit 0.

**One acceptance criterion from #3007 remains partly undischarged:** "passes in isolation **and** as part of a full `test:server` run **on a loaded box**". The cloud run covers full-suite; the *loaded-box* half rests on the contended local runs, not a clean full-suite run on a busy machine.

### Checklist notes

- **Regression plan** — not applicable; the issue body plus the changed test is the spec.
- **On-box acceptance** — not applicable; no hardware-only behaviour.
- **Release notes** — skipped: test-infrastructure only, no user- or operator-visible delta.
- **Flaky register** — no row added, deliberately. The register tracks live debt, so a row for a test fixed in the same round would be false on arrival.

Closes #3007
Refs #3022

### DIFF
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -82,6 +82,11 @@ const SERIES = 'Standalones';
 const TITLE = 'Fresh Lock Book';
 const CHAPTER_BODY = 'Nova said the plan out loud.';
 
+/* Shared synchronization-wait budget (finding 4, PR #3009 review pass 1) —
+   one named constant so a raised timeout can't silently diverge from a
+   hardcoded copy in an error message. */
+const SYNC_WAIT_TIMEOUT_MS = 5_000;
+
 let workspaceRoot: string;
 let app: Express;
 
@@ -306,9 +311,9 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
           new Promise<never>((_, reject) =>
             setTimeout(
               () => reject(new Error(
-                'add-alias never reached its intercepted in-lock read within 5000ms',
+                `add-alias never reached its intercepted in-lock read within ${SYNC_WAIT_TIMEOUT_MS}ms`,
               )),
-              5_000,
+              SYNC_WAIT_TIMEOUT_MS,
             ),
           ),
         ]);
@@ -322,9 +327,28 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         });
         jobPromise.catch(() => {}); // don't let a later stub-shape error surface here
 
-        // Generous head start: the delete either completes immediately
-        // (unlocked — the bug window) or queues behind add-alias's held
-        // lock (locked — the fix). Not a tight window either way.
+        // Finding 3 (PR #3009 review pass 1) — a `clearAnalysisCache`-completion
+        // signal was tried here as a replacement for this fixed sleep. It
+        // deadlocks: `readPriorCastForMerge` (analysis.ts:3535, unconditional
+        // whenever bookDir exists, and unconditionally BEFORE this
+        // `requestedFresh` block) takes the SAME cast lock add-alias is
+        // holding open below, so the job cannot reach `clearAnalysisCache`
+        // (analysis.ts:3639) — let alone signal its completion — until
+        // add-alias's lock is released. But `released()` below is gated on
+        // that very signal, so nothing ever fires; observed
+        // `LockAcquisitionTimeoutError` after 10s and the test's own 5s wait
+        // timing out. That also means a duration-free synchronization point
+        // for this head start does not exist with the current interceptor
+        // shape: the job cannot even attempt this lock until add-alias
+        // releases it, so there is no earlier test-visible edge to poll or
+        // await. Kept as the pre-existing fixed sleep; the fact that it does
+        // not gate anything meaningful is the subject of the "CRITICAL CHECK"
+        // finding below the core assertion — this needs a design pass (filed as #3022), not a
+        // synchronization swap, so it is deliberately left rather than
+        // "fixed" into something else. Generous head start: the delete either
+        // completes immediately (unlocked — the bug window) or queues behind
+        // add-alias's held lock (locked — the fix). Not a tight window either
+        // way.
         await new Promise((r) => setTimeout(r, 100));
 
         released();
@@ -342,7 +366,7 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
               throw new Error('cast.json still exists; delete has not completed');
             }
           },
-          { timeout: 5_000, interval: 10 },
+          { timeout: SYNC_WAIT_TIMEOUT_MS, interval: 10 },
         );
 
         // Capture disk state NOW, before Phase 0 (still gated) is allowed to
@@ -363,7 +387,17 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
       expect(resAlias!.status).toBe(200);
       /* The core assertion: whichever side acquired the lock first,
          cast.json ends up deleted, never resurrected with add-alias's stale
-         snapshot. */
+         snapshot.
+
+         #3022 — READ THIS BEFORE TRUSTING THIS TEST AS A GATE. It does not
+         currently detect removal of the `withCastLock` at analysis.ts:3660,
+         the regression it was written for: replacing that wrapper with a
+         passthrough leaves this test fully green (verified twice, PR #3009).
+         `readPriorCastForMerge` (analysis.ts:220, called unconditionally at
+         :3535, BEFORE this block) takes the same cast lock, so add-alias's
+         write has already landed by the time the job reaches the delete at
+         all — the ordering is forced upstream whether or not the delete is
+         wrapped. #3022 names the decision owed. */
       expect(castExistsAfterRace).toBe(false);
     },
     30_000,
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -87,6 +87,12 @@ const CHAPTER_BODY = 'Nova said the plan out loud.';
    hardcoded copy in an error message. */
 const SYNC_WAIT_TIMEOUT_MS = 5_000;
 
+/* How long to let a late resurrection write land before sampling disk (PR
+   #3009 review pass 2, finding 1). NOT a synchronisation guess: the delete is
+   polled for separately, and this window exists purely to give a wrong writer
+   time to be wrong in. Longer is strictly safer here, so it is not tuned. */
+const RESURRECTION_SETTLE_MS = 400;
+
 let workspaceRoot: string;
 let app: Express;
 
@@ -275,13 +281,19 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         if (!intercepted && path === raceCastPath) {
           intercepted = true;
           const value = await actual.readJson(path); // real bytes, now — happens-before the delete
-          /* Signalled AFTER the real read, not at interceptor entry (PR #2232
-             review, finding 1). The invariant this edge exists to establish is
-             "add-alias's read genuinely happens-before the delete" — resolving on
-             entry would release the delete while the read was still pending, which
-             is harmless while the route holds withCastLock but gives the
-             lock-removed mutation strictly LESS slack than the 300ms sleep did.
-             Signal what the comment actually claims. */
+          /* Signalled AFTER the real read, not at interceptor entry (PR #3009
+             review pass 1, finding 1). The SHAPE follows the sibling precedent
+             in book-state-preserve-voices.test.ts (#2215/#2232); the finding
+             that this file needed it is this PR's own, so credit both rather
+             than copying the sibling's attribution wholesale (pass 2, finding
+             4). The invariant this edge exists to establish is "add-alias's
+             read genuinely happens-before the delete" — resolving on entry
+             would release the delete while the read was still pending, which
+             is harmless while the route holds withCastLock but gives strictly
+             LESS slack than the 300ms sleep did. Deliberately NOT argued from
+             the lock-removed mutation, the way the sibling's comment is: per
+             #3022 this file cannot detect that mutation, so citing it here
+             would be reasoning from an experiment that does not run. */
           signalIntercepted();
           await gate; // hold the RESOLUTION open until released below
           return value;
@@ -306,17 +318,28 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         // is the genuine happens-before edge, deterministic regardless of machine
         // load. A poll on a boolean at entry (the old pattern) gave a shorter
         // critical window than even the 300ms fixed sleep it replaced.
-        await Promise.race([
-          interceptedSignal,
-          new Promise<never>((_, reject) =>
-            setTimeout(
-              () => reject(new Error(
-                `add-alias never reached its intercepted in-lock read within ${SYNC_WAIT_TIMEOUT_MS}ms`,
-              )),
-              SYNC_WAIT_TIMEOUT_MS,
-            ),
-          ),
-        ]);
+        /* The deadline timer is captured, unref'd and cleared (PR #3009 review
+           pass 2, finding 3). Left dangling it holds a live handle for the full
+           SYNC_WAIT_TIMEOUT_MS past the normal path — the exact hazard
+           workspace/file-lock.ts:239-247 names and defends against, in a repo
+           that already fights "Worker exited unexpectedly" teardown noise. */
+        let deadlineTimer: ReturnType<typeof setTimeout> | undefined;
+        try {
+          await Promise.race([
+            interceptedSignal,
+            new Promise<never>((_, reject) => {
+              deadlineTimer = setTimeout(
+                () => reject(new Error(
+                  `add-alias never reached its intercepted in-lock read within ${SYNC_WAIT_TIMEOUT_MS}ms`,
+                )),
+                SYNC_WAIT_TIMEOUT_MS,
+              );
+              deadlineTimer.unref?.();
+            }),
+          ]);
+        } finally {
+          if (deadlineTimer) clearTimeout(deadlineTimer);
+        }
         expect(intercepted).toBe(true);
 
         const recordRef = getManuscript(manuscriptId)!;
@@ -337,11 +360,16 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         // add-alias's lock is released. But `released()` below is gated on
         // that very signal, so nothing ever fires; observed
         // `LockAcquisitionTimeoutError` after 10s and the test's own 5s wait
-        // timing out. That also means a duration-free synchronization point
-        // for this head start does not exist with the current interceptor
-        // shape: the job cannot even attempt this lock until add-alias
-        // releases it, so there is no earlier test-visible edge to poll or
-        // await. Kept as the pre-existing fixed sleep; the fact that it does
+        // timing out. No duration-free synchronization point is AVAILABLE for
+        // this head start with the current interceptor shape: the job cannot
+        // even attempt this lock until add-alias releases it, so there is no
+        // earlier test-visible edge to poll or await. Stated as availability,
+        // not non-existence (PR #3009 review pass 2) — a lock-queue-ENTRY
+        // edge (workspace/file-lock.ts:232, synchronous, fires while add-alias
+        // still holds) would not deadlock; it is simply not exposed to a test
+        // today, and adding that seam to production is part of #3022's
+        // decision rather than something to smuggle in here.
+        // Kept as the pre-existing fixed sleep; the fact that it does
         // not gate anything meaningful is the subject of the "CRITICAL CHECK"
         // finding below the core assertion — this needs a design pass (filed as #3022), not a
         // synchronization swap, so it is deliberately left rather than
@@ -368,6 +396,21 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
           },
           { timeout: SYNC_WAIT_TIMEOUT_MS, interval: 10 },
         );
+
+        /* Settle window, RESTORED (PR #3009 review pass 2, finding 1). The poll
+           above returns at the first instant of absence, so sampling absence
+           immediately after it is tautological — that assertion could not
+           fail, and a real resurrection would instead surface as the poll's own
+           "delete has not completed" message, blaming contention for what is
+           actually a resurrection. The regression this file exists to catch is
+           a write landing AFTER the delete, so the sample has to be taken at a
+           moment the poll did not choose. Both properties kept: poll for the
+           delete (no fixed duration gating "has it happened yet"), THEN settle
+           before sampling. A fixed duration is correct here and is not the
+           mistake this PR fixes elsewhere — it is a window for a late
+           writer to be caught in, not a guess at how long something takes, so
+           erring long only makes the check stronger. */
+        await new Promise((r) => setTimeout(r, RESURRECTION_SETTLE_MS));
 
         // Capture disk state NOW, before Phase 0 (still gated) is allowed to
         // proceed to the job's own later, legitimate cast.json write.
@@ -400,20 +443,28 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
          wrapped. #3022 names the decision owed. */
       expect(castExistsAfterRace).toBe(false);
     },
-    /* Runaway backstop, not a synchronisation deadline — deliberately
-       generous (finding 5, PR #3009 review pass 1). Measured: the test BODY,
-       which is all this budget covers, ran 10.4s then 22.0s across two
-       back-to-back isolation runs on an UNLOADED box — a 2.1x spread with no
-       contention at all, the second within 8s of the old 30_000. (The ~21-24s
-       file totals usually quoted here include transform/setup/import, which
-       sit OUTSIDE this budget; quoting those understates how tight it was.)
-       Two of six runs under `flake-repro.mjs --cpu-load` blew it outright.
-       Raising it is NOT the "wider constant" mistake this PR fixes elsewhere:
-       that sleep gated an ASSERTION on a timing guess, while this only bounds
-       a hang. The real detectors are the internal bounded waits
-       (SYNC_WAIT_TIMEOUT_MS above, withKeyLock's 10s per #2260), which held
-       across all six contended runs and fail naming what never happened; this
-       catches only what they cannot see, so generosity costs no diagnostic. */
-    120_000,
+    /* Runaway backstop, not a synchronisation deadline. 60_000 is this suite's
+       house norm for the class (51 uses vs 17 of 30_000).
+
+       Corrected in PR #3009 review pass 2, finding 2. Pass 1 raised this citing
+       a test BODY of "10.4s then 22.0s on an UNLOADED box". That box was NOT
+       unloaded — three sibling worktrees were running node/python at the
+       time — and the claim was never checked before it was written down.
+       Five quiet isolation runs measure 3.56 / 4.85 / 4.74 / 6.28 / 4.54s, so
+       the honest body is ~5s and the old 30_000 already had ~5x headroom, not
+       the 1.4x pass 1 asserted. (#3007's own 25.23s figure for this shape came
+       from a contended box too.) The raise still stands, but only on the half
+       of the evidence that survives: two of six runs under
+       `flake-repro.mjs --cpu-load` blew 30_000 outright.
+
+       60_000 not 120_000: suite-wide `retry: 1` means a genuine wedge costs
+       twice the budget in fast-lane wall clock, so the ceiling is not free.
+       Raising it at all is still NOT the "wider constant" mistake this PR
+       fixes elsewhere — that sleep gated an ASSERTION on a timing guess,
+       while this only bounds a hang. The real detectors are the internal
+       bounded waits (SYNC_WAIT_TIMEOUT_MS above, withKeyLock's 10s per #2260),
+       which held across all six contended runs and fail naming what never
+       happened; this catches only what they cannot see. */
+    60_000,
   );
 });
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -400,6 +400,20 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
          wrapped. #3022 names the decision owed. */
       expect(castExistsAfterRace).toBe(false);
     },
-    30_000,
+    /* Runaway backstop, not a synchronisation deadline — deliberately
+       generous (finding 5, PR #3009 review pass 1). Measured: the test BODY,
+       which is all this budget covers, ran 10.4s then 22.0s across two
+       back-to-back isolation runs on an UNLOADED box — a 2.1x spread with no
+       contention at all, the second within 8s of the old 30_000. (The ~21-24s
+       file totals usually quoted here include transform/setup/import, which
+       sit OUTSIDE this budget; quoting those understates how tight it was.)
+       Two of six runs under `flake-repro.mjs --cpu-load` blew it outright.
+       Raising it is NOT the "wider constant" mistake this PR fixes elsewhere:
+       that sleep gated an ASSERTION on a timing guess, while this only bounds
+       a hang. The real detectors are the internal bounded waits
+       (SYNC_WAIT_TIMEOUT_MS above, withKeyLock's 10s per #2260), which held
+       across all six contended runs and fail naming what never happened; this
+       catches only what they cannot see, so generosity costs no diagnostic. */
+    120_000,
   );
 });
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -262,10 +262,22 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         released = resolve;
       });
       let intercepted = false;
+      let signalIntercepted!: () => void;
+      const interceptedSignal = new Promise<void>((resolve) => {
+        signalIntercepted = resolve;
+      });
       const spy = vi.mocked(stateIo.readJson).mockImplementation(async (path: string) => {
         if (!intercepted && path === raceCastPath) {
           intercepted = true;
           const value = await actual.readJson(path); // real bytes, now — happens-before the delete
+          /* Signalled AFTER the real read, not at interceptor entry (PR #2232
+             review, finding 1). The invariant this edge exists to establish is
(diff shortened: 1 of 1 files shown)


dudarenok-maker:
## PR review — pass 1 (head 1e7027fdb99a4d442b24c78c2df624dfeaaaaf63, depth medium)

Five findings: three correctness, two cleanup. All line numbers are against the head revision of `server/src/routes/analysis.fresh-cast-lock.test.ts` unless stated.

The headline is finding 1: **this repo already solved this exact problem, in the sibling file this file's own header cites as its precedent, and the answer there was not a poll.** `book-state-preserve-voices.test.ts` carries the identical interceptor (same `released`/`gate`/`intercepted`/`raceCastPath` shape, same verbatim `spy.mockImplementation(actual.readJson)` restore comment), was fixed for the identical defect under #2215, and its synchronisation point was then refined by a review round on PR #2232. This PR reintroduces the shape #2232 removed.

---

### 🟠 1. The poll resolves at interceptor *entry*, not after the real read — reintroducing PR #2232's finding 1 in the sibling file

`analysis.fresh-cast-lock.test.ts:294-303` (the new `vi.waitFor`), against `:265-273` (the interceptor).

The interceptor sets the polled flag before it does anything:

```js
if (!intercepted && path === raceCastPath) {
  intercepted = true;
  const value = await actual.readJson(path); // real bytes, now — happens-before the delete
  await gate;                                 // hold the RESOLUTION open until released below
```

So `vi.waitFor(() => { if (!intercepted) throw … })` returns the moment add-alias *enters* `readJson` — before `actual.readJson(path)` has run and before the handler is parked on `gate`. The comment directly above the poll (`:286-288`) claims the wait is for add-alias to "reach (and get stuck behind) its intercepted in-lock read", and `:268`'s own comment claims "real bytes, now — happens-before the delete". Neither is established by the new synchronisation point.

This is precisely the finding already fixed in the sibling. `server/src/routes/book-state-preserve-voices.test.ts:441-448`:

```
/* Signalled AFTER the real read, not at interceptor entry (PR #2232
   review, finding 1). … resolving on entry would release /assign while the
   read was still pending, which is harmless while the route holds
   withCastLock but gives the lock-removed mutation strictly LESS slack than
   the 20ms sleep did. Signal what the comment actually claims. */
signalIntercepted();
```

**Concrete failure.** Under the regression this test exists to catch (the `withCastLock` at `analysis.ts:3659` removed), the job's delete is released ~5 ms after add-alias merely entered the interceptor, plus the 100 ms head start — instead of the 300 ms the old sleep gave it. If the delete lands inside add-alias's still-pending `actual.readJson`, `state-io.ts:46`'s `if (!existsSync(path)) return null` yields `null` → `cast?.characters?.length` falsy → **409**, or the `existsSync`/`readFile` window is hit and `readFile` throws ENOENT. Either way add-alias never writes, `castExistsAfterRace` is `false`, and **the core assertion at `:343` passes on the broken code**. The test goes red only on the incidental `expect(resAlias!.status).toBe(200)` at `:339` — a failure a future maintainer can plausibly "fix" by relaxing the status assertion, at which point the file is fully vacuous. Reviewer-brief shape 3 ("a test that cannot fail") and shape 9 (the comment the change made false).

**Secondary point, same finding.** The PR body argues against "a larger constant: that reproduces the same bug with a wider window" — but `timeout: 5_000` *is* a larger constant, merely polled up to. `interceptedSignal` removes the constant entirely: it resolves on the real happens-before edge, bounded only by the honest `it()` budget. That is what #2215's comment (`book-state-preserve-voices.test.ts:409-420`) says in as many words — "replaces the guess with the real happens-before edge … deterministic regardless of machine load."

**Fix**: mirror the sibling — a `signalIntercepted` / `interceptedSignal` pair, resolved after `await actual.readJson(path)`, awaited in place of the poll. The `expect(intercepted).toBe(true)` that follows it there is worth keeping too.

---

### 🟠 2. The 400 ms sleep at `:322` is the same defect this PR fixes, and the stated rationale for keeping it does not hold

`analysis.fresh-cast-lock.test.ts:320-326`, feeding the assertion at `:343`.

```js
resAlias = await aliasPromise;                        // :319
await new Promise((r) => setTimeout(r, 400));         // :322
castExistsAfterRace = existsSync(castPath);           // :326
```

This is a fixed-duration sleep gating an assertion, which is verbatim what #3007's acceptance forbids ("No fixed-duration sleep gates an assertion in this file") and what its instruction scopes in ("Check the rest of the file for the same shape: if there are other fixed sleeps before assertions, they have the same bug and are in scope").

**The PR's rationale is factually wrong.** It says polling "until false" could "return early and miss a later resurrection write". After `:319` there is no later resurrection write available to miss:

- add-alias's write is `cast-aliases.ts:366`, and it is `await`ed **before** `return res.json(...)` at `:370`. `await aliasPromise` therefore resolves strictly after `writeJsonAtomic` has landed on disk. add-alias cannot write again.
- The only other cast.json writer in flight is the analysis job, and this file's own header (`:24-30`) establishes that Phase 0 is held on `phase0Gate` — released only in the `finally` at `:332`, i.e. after the capture — so the job's later legitimate re-creation cannot occur inside the window either.

So the sole pending mutation at `:322` is the delete at `analysis.ts:3659` — exactly what a poll would be waiting for, and a poll is strictly *more* correct than the sleep in the regression case: under a lock-removed regression the file is resurrected and already present at `:319`, so a poll for absence spins the full budget and fails, correctly.

**Concrete failure as written.** The `#2015` comment at `:320-321` states three lock handoffs now separate `released()` from the delete landing (add-alias releases → the job's locked capture acquires+releases → the delete acquires). On a box where those three handoffs plus `clearAnalysisCache` plus the `rm` exceed 400 ms — the same contention that made 300 ms insufficient at `:290` — `existsSync(castPath)` is still `true` and `expect(castExistsAfterRace).toBe(false)` fails. That is #3007's symptom, in the same file, at a different line, and the PR closes #3007.

**Fix**: `await vi.waitFor(() => expect(existsSync(castPath)).toBe(false), { timeout: 5_000, interval: 5 })`, then `castExistsAfterRace = existsSync(castPath)`.

---

### 🟠 3. The 100 ms head start at `:316` is the entire window in which the regression is detectable, and it is a fixed sleep

`analysis.fresh-cast-lock.test.ts:313-316`.

The PR's rationale — "its own comment states the duration doesn't matter, because the invariant holds whichever way the race resolves" — is true, and is the problem. It argues the sleep cannot cause a *failure*; it does not address that it can cause a vacuous *pass*.

**Concrete failure.** `runMainAnalyzerJob` does substantial async work before it reaches the delete: the whole `requestedFresh` branch begins at `analysis.ts:3638` with `await clearAnalysisCache(manuscriptId)`, and `readPriorCastForMerge` and the merge-base handling run before that (`:3537` onward). If 100 ms elapses before the job reaches `withCastLock(freshBookDir, …)` at `:3659` — on the same starved box that made 300 ms insufficient at `:290` — then `released()` fires first, add-alias writes and returns, and the delete runs afterward in *every* ordering. `castExistsAfterRace` is `false` and `:343` passes **whether or not the delete is locked**. The regression is not detected, and nothing surfaces that the test stopped discriminating.

This is the repo's most-cited failure shape (reviewer brief 3 and 6: a test that cannot fail; success reported while doing nothing). The degradation tracks CPU load, i.e. exactly the condition this PR exists to survive.

Note the sibling has the same shape at `book-state-preserve-voices.test.ts:480` (`setTimeout(r, 50)`), and that file is the one live contention row in `docs/testing/flaky-register.md` (#2226) — so this is a class, not a site.

I am not asserting a single correct fix here. Options with different tradeoffs: intercept the job's own lock acquisition (`readPriorCastForMerge` reads via `cast-fingerprint.ts`'s `readFile`, per this file's header at `:17-22`, so a test-visible edge exists) and wait on it rather than on a duration; or keep the constant but state in the comment what it is actually buying and why the budget is adequate, since the current comment asserts the opposite. If triage reads that as more than one defensible outcome, it is the design-pass carve-out and needs an issue naming the decision — but "examined and deliberately left" with the current comment is not a resolution either way.

---

### 🟡 4. The timeout is written twice and the copies can diverge

`analysis.fresh-cast-lock.test.ts:298` hardcodes `within 5000ms` in the message while `:302` sets `{ timeout: 5_000 }`. Raise the timeout to `10_000` and the failure message tells the next investigator the poll gave up at 5000 ms, misdirecting exactly the person the message exists for. Interpolate the constant (hoist it to a local and use it in both), or drop the duration from the message and let `vi.waitFor`'s own reporting carry it.

### 🟡 5. The deferred `it()`-timeout decision is named in the PR body but no issue was filed, and #3007's second acceptance criterion is not demonstrated

Two halves, one bookkeeping fix.

**No issue.** The "Known and not fixed here" section defers the 30 s `it()` budget (`:345`) as needing a decision — raise it, or route the file into `SLOW_FILES`. That is the design-pass carve-out, and CLAUDE.md's rule is explicit that the carve-out's deliverable *is* the issue: "Nothing is dropped silently. Every finding lands … as an issue number **plus the decision that is owed**." A PR-body paragraph is not that; when this branch merges and the body scrolls away, the decision is unrecorded. File it, naming the two options.

**Acceptance not demonstrated.** #3007's second acceptance bullet is "The test passes in isolation **and** as part of a full `test:server` run on a loaded box." The evidence offered is 4/6 under `scripts/flake-repro.mjs --cpu-load`, with 2/6 still failing. CI green does not close that gap: the `Server tests (fast + slow)` leg on this PR ran `npx vitest run --changed "$BASE"` and reported `✓ src/routes/analysis.fresh-cast-lock.test.ts (1 test) 2078ms` / `Test Files 1 passed (1)` — a single-file uncontended run, which is the condition that already passed before this PR. The PR body's claim that CI is "fully green" is accurate but carries no evidence about the contended full-suite criterion.

---

### Checked and clean

- **The poll condition cannot be satisfied by anything but the interception.** `intercepted` is set only for `path === raceCastPath`, and nothing between the spy install at `:265` and the poll reads that path: `findBookByBookId` → `findBookBy` reads `state.json` only (`workspace/scan.ts:968`); the cast.json read at `scan.ts:725` is in `scanBook`, which is not on this route's path. So there is no false-pass via a stray reader. The PR's claim 1 (timeout branch reachable) is consistent with the code — neutering `intercepted = true` leaves the handler parked on `gate` forever and the poll times out with the custom message.
- **`vi.waitFor` idiom, timeout/interval, error propagation.** `{ timeout, interval: 5 }` matches `script-review.test.ts:1633` (#2262) exactly; the rejection-with-the-callback's-own-error behaviour that comment documents is what makes the custom message surface. Real timers (no `vi.useFakeTimers` in this file). Correct.
- **PR body claim 3** (this file does not share the undispatched-lazy-supertest shape from #1854's graduated entry) — verified: `aliasPromise.catch(() => {})` at `:285` dispatches before anything awaits it.
- **No new resource leak.** The poll clears its own timer on both paths. The `finally` at `:327-337` restores the spy and releases Phase 0. The timeout path leaves `released()` uncalled and `aliasPromise` unawaited, but nothing is parked on `gate` in that path (`intercepted` is false by construction) and the promise carries a `.catch`, so this is the pre-existing failed-assertion shape, not something the change introduces.
- **House gates.** `Closes #3007` present and outside a code span; issue-link and title checks green. Release notes correctly skipped with a stated rationale (test-infrastructure only). On-box acceptance correctly N/A. No production code, so no cast-lock rules, no config knob, no derived artifact in scope. No flaky-register row was owed — the file never had one, and the register's graduation convention attaches to rows being removed.

dudarenok-maker:
**Ownership note — this PR's author session is taking #3009 back over, at the repo owner's direction.**

If another session is mid-flight in `C:\Claude\Projects\wt-3007-fresh-cast-lock-race`, please stand down on it now. One writer per worktree; this repo has paid for that rule three separate times (bundled commits, contention flakes, and a subagent reaching for `--no-verify` to escape contention it had caused itself).

**Thank you for the review — it is correct and it caught a real bug in my fix.** Finding 1 in particular: my `vi.waitFor` polled a flag the interceptor sets at *entry*, before `await actual.readJson(path)` runs, so it established strictly **less** slack than the 300 ms sleep it replaced, while the comment above it claimed the opposite. That is the `f_test_cannot_fail` shape, and I missed it despite briefing my agent to hunt for precedent — it found `vi.waitFor` via #2262 in the flaky register and stopped there, rather than the far more relevant #2215/#2232 fix in the sibling file that this file's own header cites. Finding the *nearest* precedent, not merely *a* precedent, is the lesson.

I have your `1cd2cef6` locally (unpushed). I will verify it rather than assume it, then carry findings 2–5 myself:

- **2** — the 400 ms sleep at `:322`. Accepted; your reasoning that no later resurrection write remains after `await aliasPromise` is correct, and my PR body's rationale for keeping it was wrong.
- **3** — the 100 ms head start at `:316`. You explicitly left this as possibly needing a design pass. I will decide and say which, rather than leaving "examined and deliberately left" standing, which as you note resolves nothing either way.
- **4** — the duplicated 5000 ms constant.
- **5** — the unfiled `it()`-timeout issue and the undemonstrated second acceptance criterion.

Re-review will follow the fixes.

dudarenok-maker:
## PR review — pass 2 (head 245cfe4d, depth medium)

Scope: `server/src/routes/analysis.fresh-cast-lock.test.ts` (the only changed file), read against `analysis.ts`, `cast-aliases.ts`, `workspace/state-io.ts`, `workspace/scan.ts`, `workspace/file-lock.ts`, `book-state-preserve-voices.test.ts`, both vitest server configs, `docs/testing/flaky-register.md`, and issues #3007 / #3022 · verified: `tsc -p .` exit 0, `eslint` exit 0, 5 isolation runs of the file (`--retry=0`, one with `server/node_modules/.vite` cleared) all green, all 19 cloud checks green, file is 419/419 CRLF (no mixed endings introduced) · then adversarially probed the finding-2 replacement, the finding-1 signal placement, and the finding-5 measurement.

### Round-1 dispositions

**Finding 1 (🟠 signal at interceptor entry) — VERIFIED RESOLVED.** `signalIntercepted()` sits at `:285`, strictly after `await actual.readJson(path)` at `:277` and strictly before `await gate` at `:286`, so it cannot deadlock and it is a genuine happens-before edge. Two things the fix depends on that pass 1 did not check, both verified here: (a) the intercepted read really is add-alias's *in-lock* read — `withCastLock` is entered before `readJson` in the handler (`cast-aliases.ts:331` → `:332`), and the only earlier candidate on the request path, `findBookByBookId` → `findBookBy` (`scan.ts:987-1022`), reads state.json/segments only and never touches cast.json, so `scanBook`'s cast read at `scan.ts:725` is off this path and cannot steal the one-shot interception; (b) the deadline race does not mask a hang — on expiry it rejects with a named message and the test goes red.

**Finding 2 (400 ms settle sleep → poll) — RESOLVED. Probed hardest, as asked; the early-return hazard is not reintroduced today.** All three legs the new comment leans on hold:
- `writeJsonAtomic` (`state-io.ts:105-132`) writes a temp file and `renameWithRetry`s it over the target — there is no transient-absence window for the 10 ms poll to mistake for a completed delete (the `rotate` path, which *does* vacate `path` first, is not used by any cast.json writer on this path).
- add-alias's write strictly precedes its response: `await writeJsonAtomic(...)` at `cast-aliases.ts:365`, `return res.json(...)` at `:369`, both inside the same `withCastLock`. So `await aliasPromise` genuinely bounds it.
- "Phase 0 is still gated" is sufficient: grepping `castJsonPath` / `writeChecked` across the whole span from the delete (`analysis.ts:3661`) to the first `runStage1Chapter` (`analysis.ts:4273`) returns nothing, so once absent, absent is terminal.

Carrying a 🟡 below for what the swap *did* cost.

**Finding 3 (100 ms head start) — deferral is HONEST.** Verified independently: `readPriorCastForMerge` (`analysis.ts:220`) opens with `withCastLock`, is called at `:3535`, and the `requestedFresh` block is at `:3638` — so the job provably cannot reach `clearAnalysisCache` (`:3639`) while add-alias holds the lock, and the proposed edge really does deadlock. #3022 names three defensible outcomes and the decision owed, which is exactly CLAUDE.md's design-pass carve-out. One overreach in the in-file wording: "a duration-free synchronization point for this head start does not exist" (`:340-344`) is inferred from *one* candidate failing. A queue-entry edge (a waiter is enqueued into `file-lock.ts`'s `chains` map synchronously at `:232`, before any await) fires while add-alias still holds the lock and would not deadlock — it is unavailable only because nothing exports it, i.e. it needs a production test seam, not because no edge exists. The hedge "with the current interceptor shape" mostly covers this; the accurate claim is "none found that doesn't require a new production seam".

**Finding 4 (duplicated 5 s deadline) — VERIFIED RESOLVED.** `SYNC_WAIT_TIMEOUT_MS` (`:88`) is interpolated at both the wait (`:316`) and its message (`:314`). Note it is now also the budget for the *second*, structurally unrelated wait added by finding 2's fix (`:369`) — a cold-start budget and a lock-handoff budget now move together. Acceptable, but not what the constant's own comment describes.

**Finding 5 (`it()` budget) — FIXED, but the evidence recorded for it does not reproduce.** See 🟡 below. `vi.waitFor` does surface the callback's own error on timeout (`vitest/dist/chunks/test.DNmyFkvJ.js:3374-3377` rejects with `lastError`, falling back to "Timed out in waitFor!" only when the callback never threw), so the "fail naming what never happened" half of the rationale checks out.

### 🟡 Minor — the core assertion can no longer fail, and the detector that replaced it misnames the failure

`analysis.fresh-cast-lock.test.ts:363-370` polls until `!existsSync(castPath)`; `:374` then re-reads the same predicate and `:401` asserts it. The assertion is now downstream of a wait that exits only when it already holds — it cannot go red. That is survivable (the waitFor times out instead, so the test still fails), but it has two concrete costs:

1. **Wrong attribution.** State: a regression where the delete lands and a concurrent writer resurrects cast.json afterwards. Output: `cast.json still exists; delete has not completed` — false; the delete *did* complete. That message points the next engineer at lock contention, which is the one hypothesis this file already has a 22-line comment about, instead of at resurrection, which is the invariant the file is named for.
2. **The settle window is gone.** The old fixed 400 ms observed the system *after* it stopped moving; the poll observes it at the first instant the predicate is true and returns. Today nothing can write cast.json in that gap (verified above), so this is a loss of defence-in-depth rather than a live bug — but the only thing now preventing a post-delete writer from being missed is a comment asserting a property of `analysis.ts`, which no test enforces. The PR body's original argument for keeping the sleep was correct on this point and was dropped without the combination being considered: `await vi.waitFor(absent)` **then** the settle, then capture, keeps both properties for +400 ms.

Not graded 🟠 because I could not construct a present-day interleaving that produces a wrong pass — I looked for one specifically.

### 🟡 Minor — the measurement justifying the 4× budget raise does not reproduce, and it is now a permanent in-file comment

`analysis.fresh-cast-lock.test.ts:403-416` records as fact: "the test BODY … ran 10.4s then 22.0s across two back-to-back isolation runs on an **UNLOADED** box — a 2.1x spread with **no contention at all**, the second within 8s of the old 30_000."

Five consecutive isolation runs in this worktree on this same box today, `npx vitest run src/routes/analysis.fresh-cast-lock.test.ts --retry=0`:

| run | file total | test body |
|---|---|---|
| 1 | 5.25 s | **3.56 s** |
| 2 | 6.71 s | **4.85 s** |
| 3 | 6.52 s | **4.74 s** |
| 4 | 8.03 s | **6.28 s** |
| 5 (after `server/node_modules/.vite` cleared) | 6.26 s | **4.54 s** |

Max body 6.28 s — roughly **5× headroom under the old 30 s**, not "within 8 s of the ceiling". Run 5 rules out the obvious reconciliation (a cold transform cache inflating the in-body `await import('./analysis.js')` at `:212`). #3007's own evidence table records the same isolation shape at **25.23 s** file total where I measure 5.25–8.03 s on the same hardware — so the 10.4/22.0 s figures are consistent with a *contended* box, which is precisely what the comment says they were not.

The raise itself still stands on the other half of the evidence (2 of 6 `--cpu-load` runs blew 30 s) — the defect is the recorded premise and the inference drawn from it, which is the durable artifact a future maintainer will reason from. Two things to weigh while correcting it: the house norm for this class is `60_000` (`analysis.test.ts` ×10, `analysis.persist-lock-timeout.test.ts` ×3, `analysis.interim-prune-prohibition.e2e.test.ts`), so 120 s is 2× the norm; and with the suite-wide `retry: 1` (`server/vitest.config.ts`), a wedged run now costs 240 s of fast-lane wall clock rather than 60 s. "Generosity costs no diagnostic" is true; it is not free.

### 🟡 Minor — the deadline timer is neither cleared nor `unref`'d

`:311-318` arms a bare 5 s `setTimeout` inside `Promise.race`. When `interceptedSignal` wins (the normal path, ~ms in), the timer stays armed and holds an active handle for the remaining ~5 s. No unhandled rejection (the race attached a handler), so this is a nit — but it is the exact hazard `file-lock.ts:239-247` calls out by name and defends against: *"A pending timer keeps the event loop (and vitest) alive; unref so a cleared-on-happy-path timer is never the reason a process hangs at teardown (this repo already fights `Worker exited unexpectedly`)."* The sibling precedent this fix is modelled on carries no timer at all — `book-state-preserve-voices.test.ts:471` is a bare `await interceptedSignal`. Either `timer.unref()`, clear it in a `.finally()`, or drop the race and let the (now 120 s) `it()` budget be the backstop.

### 🟡 Minor — comment at `:278` credits the wrong review

`/* Signalled AFTER the real read, not at interceptor entry (PR #2232 review, finding 1). */` is copied verbatim from `book-state-preserve-voices.test.ts:441-449`, where the citation is correct. Here it labels **this PR's** pass-1 finding 1, and it is the only one of the four new comments not tagged `PR #3009 review pass 1` (contrast `:86`, `:330`, `:404`). Grepping this file for `#3009` returns findings 3, 4 and 5 but not 1. Secondary: the copied rationale — "gives the lock-removed mutation strictly LESS slack" — argues from a mutation that the same file's `:392-400` comment says this test cannot detect at all, so as written it justifies the fix by a benefit #3022 says is unrealisable here. Worth one clause acknowledging that.

### Disclosure accuracy (asked for explicitly)

The #3022 disclosure in the PR body and at `:392-400` is accurate, not overstated. Every cited line checks out: `analysis.ts:220` is `export async function readPriorCastForMerge`, `:3535` is its call, `:3639` is `clearAnalysisCache(manuscriptId)`, `:3660` is the delete's `withCastLock`. One wording slip, in the PR body only (the in-file comment gets it right): "called **unconditionally** at `analysis.ts:3535`" — it is conditional on `recordRef.bookDir`, which also slightly undercuts #3022's option-3 caveat "loses coverage if the upstream call is ever made conditional" (it already is).

House gates: `Closes #3007` / `Refs #3022` present and outside code spans ✅ · release notes explicitly skipped with a stated reason (test-infra only) ✅ · regression plan and on-box acceptance explicitly N/A ✅ · no flaky-register row — correct, the register tracks live debt and #3007 never had one ✅ · no `cast.json` write, config knob or derived artifact touched ✅.

### Verdict

No 🔴 or 🟠. Four 🟡 cleanup findings, none of which blocks merge under the re-review trigger. All five round-1 findings are resolved or honestly deferred; the finding-2 replacement survived the hardest probe I could construct. The one I would not leave unfixed is the measurement comment — a false premise recorded permanently in source is the shape this repo's own rubric calls "the defect is in the instrument or the document, not the code".

dudarenok-maker:
## Pass 2 findings fixed — `a2045e33`

All four 🟡 addressed. No 🔴/🟠 in pass 2, so no re-review round is triggered; recording the dispositions here so the thread reads found → fixed.

**F1 — core assertion could no longer fail.** Poll for the delete, **then** settle (`RESURRECTION_SETTLE_MS`), then sample. Verified by mutation rather than by inspection: a writer scheduled inside the settle window fails the test at `expect(castExistsAfterRace).toBe(false)` with `expected true to be false`. Tree restored byte-identical afterwards. Good catch — pass 1's own finding-2 fix is what broke this, so the round genuinely found the previous round's damage.

**F2 — the measurement, and this one is mine to own.** I claimed a test body of "10.4s then 22.0s on an UNLOADED box". The box was not unloaded: three sibling worktrees were running node/python at the time. I asserted a control condition without checking it and then wrote it into a permanent in-file comment. Your five quiet runs (3.56 / 4.85 / 4.74 / 6.28 / 4.54s) are right; my two are the outliers, and #3007's 25.23s figure was contended too. The raise survives only on the `--cpu-load` half of the evidence, so the budget is now **60_000**, the house norm (confirmed independently: 51 uses vs 17 of `30_000`), with `retry: 1` doubling the wedge cost as the deciding argument against 120_000.

**F3 — deadline timer** captured, `unref`'d and cleared in a `finally`.

**F4 — attribution** corrected to PR #3009 pass 1, crediting the sibling for the *shape* only, and the copied argument-from-mutation removed (per #3022 this file cannot run that mutation).

**Overreach softened.** "A duration-free synchronization point does not exist" → "is not AVAILABLE", naming the lock-queue-entry edge (`file-lock.ts:232`) as a real candidate blocked only by the absence of a production test seam, which is part of #3022's decision.

**The `unconditionally` slip.** Correct — `analysis.ts:3534` guards the call on `recordRef.bookDir`. Fixed in the PR body and in **#3022**, where it also undercut option 3's caveat: the subsumption is narrower than that option claimed, not broader. #3022 carries a correction note.

Verification: `tsc -p .` exit 0, `eslint` exit 0, green in isolation (bodies 5.81 / 6.42 / 6.28s), CRLF unchanged.

Issue #2912, opened 2026-09-04
[agent instructions][claude][task] Add the seventh CloneUnready state and pair it in the contract test (#2054)


dudarenok-maker:
## What

First task child of the **#2054** deliverable, and the one that creates the shared worktree
and branch for the whole chain.

Add a seventh `CloneUnready` state for a cloned slot whose `libraryUuid` cannot be resolved,
map it from the render's own `misconfigured` classification, and extend the co-oracle
contract test so the two oracles cannot drift apart on it again.

## Background you need

Plan 276's cast-time gate warns a user, before generation starts, that a cast will not
render. It currently stays **silent** for a slot tagged `provenance: 'cloned'` that carries a
missing, empty or malformed `libraryUuid` — `buildInput`
(`src/store/clone-readiness-selectors.ts`) returns `undefined`, so nothing is evaluated.

The render does not share that opinion: `classifyClonedVoice`
(`server/src/tts/clone-voice-resolver.ts`) reports **`misconfigured`** and the chapter
hard-fails. So the gate promises "no warning means it will render" and then it does not
render. That false-all-clear is the exact failure class plan 276's own revisions 1 and 2 died
on.

The state was left unhandled on the strength of plan 276's Decision 2, "what stays invisible
— accepted". **That citation does not hold**, and knowing why keeps you from re-deriving the
same mistake: Decision 2 accepts **disk-integrity** gaps only — whether the `.pt` and master
clip exist on disk — and states that "every state #1980 is about is metadata-visible". An
unresolvable `libraryUuid` is metadata, not disk.

The current union is at `server/src/tts/clone-readiness.ts:31` and has six values:
`revoked`, `wrong-engine`, `derive-failed`, `missing-master`, `no-transcript`,
`missing-entry`. The classifier is `cloneReadiness()` at
`server/src/tts/clone-readiness.ts:83`.

**The one open design question is already decided by the repo owner: the new state reuses
`missing-entry`'s CTA** — re-assign a voice, the only repair that actually works. Do not
design a new CTA. The frontend wiring of that CTA is child #2909's job, not yours.

## The fix

1. Add the seventh value to the `CloneUnready` union. Name it for the condition, not the
   symptom — `unresolvable-uuid` unless you find a strong local reason to prefer another
   spelling, in which case say so in your receipt, because child #2909 and the contract test
   both depend on the exact string.
2. Have `cloneReadiness()` return it for a cloned slot with no resolvable `libraryUuid`.
[...]

Issue #1891, opened 2026-07-27
fs-38 — a legacy cast.json slot (libraryUuid, no provenance) skips the revocation check


dudarenok-maker:
## What

The cloned-voice pre-pass filter (`characterHasClonedSlot`, used by
`synthesise-chapter.ts`'s union pre-pass, Wave 3c Task 20) correctly requires
`provenance === 'cloned'` before a slot enters the resolver's revocation
check — matching the pre-existing (pre-3c) filter's own requirement.

But: a legacy-shaped `cast.json` slot that carries a `libraryUuid` **and no
`provenance` field at all** (old data written before the `provenance`
dimension existed) does not match `provenance === 'cloned'`, so it is
correctly excluded from the pre-pass filter — but is then rendered straight
from the voice-library artifact via the normal resolution path, with **zero
revocation check**. If that voice is later revoked, this old-shape slot
would keep rendering it.

## Status

Not a Wave 3c regression — the filter's `provenance === 'cloned'`
requirement is unchanged from before this wave, and the old filter had the
identical requirement. This is a pre-existing consent-gate hole on
old-shape data specifically, surfaced (not introduced) during Wave 3c's
Task 20 review while verifying the pre-pass filter's completeness (D-G:
"the validate set must be a superset of the resolve set").

## Suggested fix shape

Whoever picks this up should first confirm how much old-shape data (a
`libraryUuid` with no `provenance`) actually exists in practice — this may
be rare enough that a one-time migration/backfill (stamping `provenance:
'cloned'` onto any slot with a `libraryUuid` pointing at a voice-library
entry whose own `provenance` is `cloned`) is simpler than widening the
pre-pass filter's fail-safe surface. A migration is likely safer than
loosening the filter, since a loosened filter risks re-opening the exact
malformed-uuid class of Phase-0 bug this wave's Global Constraints exist to
prevent.

## Key files

- `server/src/tts/clone-engines.ts` (`characterHasClonedSlot`)
- `server/src/tts/synthesise-chapter.ts` (the pre-pass filter's call site)
- `server/src/workspace/voice-library.ts` (where a migration/backfill would
  live, if that's the chosen fix)

## Related

- fs-38 / #624 (voice cloning umbrella)
- Plan: `docs/features/270-fs38-wave3c-xtts.md`, "Known limitations (j)"

Issue #1600, opened 2026-07-14
fs-61 — backfill designed voices + covers onto the zh/ja Coalfall placeholder samples


dudarenok-maker:
## What
fs-59 W5 (#1004, PR #1604) shipped **minimal placeholder** `samples/the-coalfall-commission-{zh,ja}/` so the `language-sample-coverage.test.ts` guard (#1568) passes now that zh/ja are `supported:true`. Each placeholder already carries:
- `manuscript.md` (copied from the committed `server/src/__fixtures__/the-coalfall-commission.{zh,ja}.md` translations),
- a `state.json` modeled on the `-de` sample (language, 3-chapter split with the framing chapter excluded, deterministic cover gradient),
- a **real 13-character `cast.json`** analysed from the manuscript via the app (localized narrator 旁白 / 語り手 + the full Coalfall roster),

but with **empty `overrideTtsVoices`** — i.e. **no designed voices**, and no slimmed cover.

This item **backfills onto those existing placeholders** (it is NOT "add a missing sample"):
1. Run the Qwen VoiceDesign pipeline on a GPU box to design per-character voices for the zh/ja rosters (same bar as the `-de`/`-es`/`-fr`/`-ru` samples), writing the `overrideTtsVoices` + `voices/qwen/*.pt` into each sample.
2. Slim each sample's cover (`scripts/slim-epub-cover.mjs` equivalent for the sample cover).

## Acceptance
- `samples/the-coalfall-commission-{zh,ja}/` each carry designed per-character Qwen voices (non-empty `overrideTtsVoices`, `.pt` files present) and a slimmed cover, matching the other language samples' quality bar.
- `language-sample-coverage.test.ts` stays green (it already passes with the placeholders).

## Depends on
fs-59 W5 (#1004) — the `supported:true` flip that shipped the placeholders.

## Benefit (user)
Gives the CJK demo books real, listenable voices so a zh/ja user gets a runnable demo, not just a roster — completing the fs-61 per-language sample set.

dudarenok-maker:
## The cover half of this is bigger than "slim it" — there is no cover to slim

Surfaced while fixing #1823 (the German marketing fixture was falling back to the English cover) and parking the per-language cover masters.

[...]

PR #3014, opened 2026-09-06
fix(server,frontend): surface the cast-time clone gate on an unresolvable libraryUuid

### LINKED ISSUE #2054: fs-38 — cast-time clone gate is silent when a cloned slot has no resolvable libraryUuid (render hard-fails `misconfigured`)
## What

Plan 276's cast-time gate (#1980) evaluates a character only when its cloned slot yields a library uuid. `buildInput` (`src/store/clone-readiness-selectors.ts`) returns `undefined` when **no library uuid can be resolved at all** — a slot tagged `provenance: 'cloned'` carrying a missing, empty, or malformed `libraryUuid`.

The render does **not** ignore that state: `classifyClonedVoice` (`server/src/tts/clone-voice-resolver.ts`) reports `misconfigured` and the chapter hard-fails.

So this is a **false all-clear** — the gate stays silent and the user hits the render failure the feature exists to move earlier. Narrow, but it is the exact failure class that killed revisions 1 and 2 of the plan.

## Why it was not fixed in the delivering PR

The in-code justification cites plan 276's Decision 2, "What stays invisible — accepted". **That citation does not hold.** Decision 2 accepts *disk-integrity* gaps only (whether the `.pt` and the master clip file exist on disk) and explicitly says "Every state #1980 is about is metadata-visible." An unresolvable `libraryUuid` is metadata, not disk.

Fixing it is design work rather than a defect fix: `CloneUnready` gains a seventh value, Decision 1's CTA table gains a row (and "what does the user press?" is a genuine question — the only repair is re-assigning a voice, which is `missing-entry`'s CTA), and the co-oracle contract test's vocabulary map gains a pair. That is a plan change, not a one-line correction.

## Acceptance

- A character whose cloned slot carries no resolvable `libraryUuid` is surfaced by the cast-time gate rather than silently skipped, with a CTA that actually repairs it.
- The co-oracle contract test (`server/src/tts/clone-readiness-contract.test.ts`) covers the pairing, so the two oracles cannot drift on it again.
- Plan 276's Decision 2 is amended so the accepted-invisible list says what it actually accepts.

## Key files

- `src/store/clone-readiness-selectors.ts` — `buildInput`'s `undefined` return
- `server/src/tts/clone-readiness.ts` — the `CloneUnready` union
- `server/src/tts/clone-voice-resolver.ts` — the render's `misconfigured` reason
- `docs/features/276-cast-time-derivability-warning.md` — Decisions 1 and 2

## Benefit

*Benefit (user):* closes the last silent gap in the cast-time check, so "no warning" reliably means "this cast will render".

Found while implementing plan 276 task 8 (the co-oracle contract test). Relates to #1980.

dudarenok-maker:
## Summary

Implements docs/features/archive/276-cast-time-derivability-warning.md's clone-gate fix for
an unresolvable `libraryUuid`. `buildInput` (`src/store/clone-readiness-selectors.ts`) no
longer returns `undefined` for a cloned slot whose `libraryUuid` is missing, empty or
malformed — it now feeds `libraryUuidResolvable: false` into a new `unresolvable-uuid`
`CloneUnready` state (`server/src/tts/clone-readiness.ts`, rule 1), which the gate surfaces
with the same "Assign a different voice" CTA `missing-entry` already offers, per the repo
owner's decision to reuse it rather than invent new copy. The co-oracle contract test
(`server/src/tts/clone-readiness-contract.test.ts`) pairs the new client verdict against the
render's existing `misconfigured` classification so the two oracles cannot drift apart on
this case again. Plan 276's Decision 2 ("what stays invisible — accepted") is amended to say
it accepts only a disk-integrity gap, not a metadata-visible one, and to name
`unresolvable-uuid` as what now handles the metadata-visible case it had been mis-cited
against. Closes #2054.

## Test plan

- [x] `npx vitest run` (frontend: `clone-readiness-selectors.test.ts`,
      `clone-readiness-gate.test.tsx`) — 39/39 passing.
- [x] `npx vitest run` (server: `clone-readiness.test.ts`,
      `clone-readiness-contract.test.ts`) — 50/50 passing.
- [x] `npx tsc --noEmit` — clean on both the frontend and server projects.
- [x] Independently re-ran the central mutation (restoring `buildInput`'s old
      `if (!libraryUuid) return undefined;` bail): reddened exactly the 3 new
      `unresolvable-uuid` selector tests (17 passed / 3 failed), nothing else — confirmed
      the fix is load-bearing, then reverted and confirmed green again.
- [x] Manually traced `buildInput` and `castOnEngineFor`: no remaining path returns
      `undefined` for a character with any cloned (or legacy bare-uuid) slot; the only
      `undefined` return is `characterNeedsCloneCheck` being false, i.e. a designed or
      catalogue slot that was never in scope.
- [x] Confirmed the contract test's new row genuinely pairs both oracles (client
      `libraryUuidResolvable: false` vs. the render's `libraryUuid: undefined` request
      field) rather than asserting each independently against a literal.
- [x] Confirmed the six pre-existing `CloneUnready` states are behaviourally unchanged —
      rule numbers shifted by one to make room for the new rule 1, but relative order and
      logic are identical; all their tests still pass.

Independent PR review pass 1 (see the [review comment](https://github.com/dudarenok-maker/Castwright/pull/3014#issuecomment-5556548669)) found no correctness bugs but five cleanup-only findings, each fixed in this round by a dispatched fix agent (per CLAUDE.md's incidental-findings protocol):

- `534e2e02` — fixed ~18 stale rule-number cross-references (across `voice-library.ts`, several test files, and plan 276) left off-by-one after this PR's new rule 1 shifted every later `CloneUnready` rule number up by one.
- `0e9c6ea` — removed the now-shipped #2054 row from `docs/BACKLOG.md` (and a stale reference in `docs/features/277-v115-bug-chore-sweep.md`).
- `0ea6683f` — removed an unreachable guard in `clone-readiness-selectors.ts`'s `buildInput`; both callers already gate on `characterNeedsCloneCheck`, so the guard's condition could never be true, and no test exercised it.
- `c18b7794` — added the missing engine-name assertion to the `unresolvable-uuid` e2e test, which claimed to verify character/engine/reason like its sibling but never actually asserted the engine.
- `82d1995a` — added a compile-time exhaustiveness guard so a future `CloneUnready` state added without a paired `VERDICT_PAIRS` row now fails the build, not just silently passes.

All five are 🟡 cleanup-only (no correctness bugs), so per the review-gate's re-review trigger this does not require a re-review pass.


dudarenok-maker:
## PR review — pass 1 (head 7f11836, depth high)

Scope: all 11 changed files, plus the unchanged consumers the change reaches — `server/src/routes/voice-library.ts`, `server/src/tts/synthesise-chapter.ts`, `server/src/tts/clone-voice-resolver.ts`, `server/src/tts/clone-engines.ts`, `src/store/start-generation-flow.ts`, `src/data/help-failures.ts`, `openapi.yaml`, `docs/BACKLOG.md`, `docs/features/277-*.md`, `docs/testing/clone-readiness-gate-onbox-acceptance.md`. Verified: CI green per dispatch (typecheck, frontend/server/sidecar, e2e incl. visual, build); this pass did not re-run the suites — read-only inspection at the head SHA, with the PR body's two claimed mutations traced statically through `libraryUuidForClonedCharacter` and `characterNeedsCloneCheck` before adversarially probing (a) whether `unresolvable-uuid` can fire on a cast the render would complete, (b) whether reordering the rules changed any existing verdict, (c) whether the co-oracle row genuinely pairs rather than double-asserts, and (d) every cross-reference the +1 rule renumbering touched.

### 🔴 Blocking

None.

### 🟠 Significant

None.

### 🟡 Minor

**1 — The rule renumbering was applied in three files and left armed in six more; ~18 cross-references now name the wrong rule.**
`server/src/routes/voice-library.ts:755-757` (worst case), plus the list below.

At head, rule 4 is `!isCloneEngine(engine) || !characterHasSlot -> 'wrong-engine'` and rule 5 is `slotStatus === 'failed' -> 'derive-failed'` (`server/src/tts/clone-readiness.ts:74-81`). `voice-library.ts:755-757` still reads:

> `cloneReadiness`'s rule 4 (`slotStatus === 'failed'` -> `derive-failed`) is ordered BEFORE rules 5/6 (`missing-master` / `no-transcript`)

Every number in that sentence is off by one, and each wrong number names a real, different rule — so it stays internally coherent and is externally false. That paragraph is the `[R3]` safety argument for why `POST /:voiceUuid/engines/:engine/retry` cannot reintroduce the retry loop; a reader who reconciles it by trusting the numbers over the parentheticals concludes that `wrong-engine` outranks a `failed` slot unconditionally, which is only true when the engine is also wrong. This is catalogue shape #10 (one instance fixed, the class left armed) sitting on top of #9 (the defect is in the document).

Sites, all off by exactly one:

- `server/src/routes/voice-library.ts:720` — "rules 5/6 — gated on `slotStatus !== 'ready'`" → 6/7
- `server/src/routes/voice-library.ts:755-757` — "rule 4 … BEFORE rules 5/6" → rule 5 … before rules 6/7
- `server/src/routes/voice-library.ts:762` — "isn't expressible in rules 5-6" → 6-7
- `server/src/routes/voice-library.test.ts:839` — "rules 5/6, gated on" → 6/7
- `src/store/start-generation-flow.test.ts:169` — "rule 7 silence" → rule 8

And five more inside files this PR already edits, where the obligation was in view:

- `src/store/clone-readiness-selectors.ts:19` — "would make rule 3 true for every cloned character" → rule 4
- `src/modals/clone-readiness-gate.tsx:132-133` — "rule 4 (`derive-failed`) is only reached after rule 3 confirms the engine IS clone-capable" → rule 5 … after rule 4
- `src/store/clone-readiness-selectors.test.ts:53` and `:99` — "rule 7 silence" → rule 8
- `src/modals/clone-readiness-gate.test.tsx:8` — "(rule 7 silence, …)" → rule 8
- `e2e/clone-readiness-gate.spec.ts:43` and `:311` — "rule 7 silence" → rule 8

And the plan the code cites *as the contract* (`clone-readiness.ts:73` — "the order is the contract (plan 276 Decision 3)"), whose Decision 1 table this PR did amend but whose rule numbers it did not: `docs/features/archive/276-cast-time-derivability-warning.md:210, 225, 231, 343, 344, 483, 485, 493, 499`. After this merge the plan's "rule 5"/"rule 6"/"rule 7" and the module's are different rules.

**2 — `docs/BACKLOG.md` still renders #2054 as outstanding work.**
`docs/BACKLOG.md:72-76`, under `## Should`.

#2054 carries `type:feature` + `moscow:should`, so it renders in the generated view, and this PR closes it. The row still says the gate "is silent when a cloned slot has no resolvable libraryUuid" and that `buildInput` "returns `undefined` when no library uuid can be resolved at all" — the exact behaviour this PR removes. CLAUDE.md "The backlog" step 2 and this gate's own derived-artifact check both name this file.

Secondary, lower confidence: `docs/features/277-v115-bug-chore-sweep.md` is `status: active` and still records #2054 at `:907` ("fs-38 Wave 3, which is **paused** … They unpause with the wave or not at all") and counts it at `:1251` ("**Gated — paused wave, unchanged (2):** #2054, #1600"). `:1251` feeds a live tally, so merging moves it. 277's round sections are partly historical record, so this one is a judgement call in a way the BACKLOG row is not.

Also worth noting here rather than as its own finding: the PR body silently omits before-shipping steps 3 (on-box acceptance) and 4 (INDEX). Both are genuinely not-applicable — see the ✅ block — but CLAUDE.md asks for that to be said, not left blank.

**3 — The new `buildInput` guard is unreachable, and the test named as its cover reddens through a different guard.**
`src/store/clone-readiness-selectors.ts:138` — `if (!characterNeedsCloneCheck(character)) return undefined;`

`buildInput` is module-private with exactly two callers: `selectCloneReadinessVerdicts:237`, which runs `if (!characterNeedsCloneCheck(character)) continue;` on the same `character` immediately before calling it, and `castOnEngineFor:207`, reached only from inside that same guarded loop with that same `character`. So line 138 can never return. Concrete: delete line 138 and the whole frontend suite stays green.

The test written as its cover — `clone-readiness-selectors.test.ts`, "mutation 2 control: a non-cloned (designed) slot with no libraryUuid at all still produces no gate … and go red if that branch is widened to also fire on non-cloned slots" — does go red under `characterNeedsCloneCheck → true`, but via the selector's guard at `:237`, not via `:138`. So a branch has coverage attributed to it that it does not have (catalogue #3: red, but for a different reason than the one claimed). Either drop line 138 per CLAUDE.md's "no error handling for impossible scenarios", or, if it is wanted as `buildInput`'s own contract, export it and test it directly.

**4 — The new e2e test's name promises an engine assertion it does not make.**
`e2e/clone-readiness-gate.spec.ts:286-301`.

It reuses the sibling's title verbatim — "names character/engine/reason" — but asserts only the character name, the reason copy and the CTA. The sibling at `:225` does assert the engine: `await expect(row.getByText('Qwen', { exact: true })).toBeVisible();` (`:238`). Concrete: delete the `<span className="text-xs text-ink/50">{ENGINE_LABEL[verdict.engine]}</span>` at `src/modals/clone-readiness-gate.tsx:168` and the `no-transcript` spec goes red while the new `unresolvable-uuid` spec stays green — while still claiming in its own name to cover it.

This matters slightly more than a naming nit because `reasonCopy`'s `unresolvable-uuid` string (`clone-readiness-gate.tsx:83`) is the only one of the seven that never interpolates `engineName`. For every other verdict the engine survives in the reason copy; for this one the header span is the *only* place it appears — and that is precisely the assertion the new test drops.

**5 — Nothing pins that a future `CloneUnready` value must be paired in the co-oracle test.**
`server/src/tts/clone-readiness-contract.test.ts:78` (`VERDICT_PAIRS`) and `:365-380` (the new row).

This PR adds the seventh state and correctly adds both a `VERDICT_PAIRS` entry and a fixture row — by hand, with nothing requiring it. Concrete: add an eighth `CloneUnready` value plus its rule and the entire contract suite stays green with no pairing and no fixture; the only thing that fails the build is `reasonCopy`'s exhaustive switch at `clone-readiness-gate.tsx:70-84`, which is the *copy* gate, not the co-oracle gate. The file's stated premise is that "an unlisted pair is a disagreement, full stop" — but a state never fed through the table is not a disagreement, it is invisible. A loop asserting every `CloneUnready` value appears at least once as a `VERDICT_PAIRS.client` and once as a `rows[].expectedClient` closes it; `CloneUnready` is a bare type union today, so that guard needs a runtime list alongside it.

### ✅ What is solid

- **Rule 1's precedence is right, and pinned in both directions.** A slot with no uuid genuinely cannot be asked any later question — `entryFound`, `consentRevoked`, `slotStatus`, `hasMaster` and `transcript` all derive from an entry lookup that cannot happen. `clone-readiness.test.ts` pins it against a maximally-broken remainder *and* against `entryFound: false` specifically, and the contract test adds `['unresolvable-uuid', 'revoked']` / `['unresolvable-uuid', null]` / `[null, 'misconfigured']` as rejected pairings.
- **The contract row is genuinely co-oracle, as the PR body claims.** One field drives both sides: `clientInputFor:240` reads `row.libraryUuidResolvable !== false` into the client input, and `renderVerdict:343` reads the same field into `libraryUuid: … : undefined` on the render request. The two verdicts are computed from one fact rather than asserted independently against literals, which is the property that would otherwise make this row decorative.
- **The verdict is a true positive — the gate cannot warn on a cast that would have rendered.** `synthesise-chapter.ts:1683` builds the resolver request's `libraryUuid` through `libraryVoiceForEngine`, which resolves an empty/missing/non-string slot uuid to `undefined`, and `clone-voice-resolver.ts:488-490` turns that into `misconfigured` before `readEntry` is ever called. I probed the reverse direction specifically (a cloned slot with a `name` but no uuid falling through to a catalogue voice, which would have made this a false warning) — `synthesise-chapter.ts:1436` and `:1670-1683` both rule it out by construction.
- **No regression through the `castOnEngine` scan.** `castOnEngineFor:207-208` skips a candidate whose verdict is non-null, so an unresolvable candidate is skipped exactly as it was when `buildInput` returned `undefined` — same `null` result, no behaviour delta. And `castOnEngine` is computed only for `wrong-engine`, which rule 1 now preempts, so the two can never interact.
- **The legacy bare-uuid shape (#1891) is untouched by construction, not by luck.** `isLegacyBareQwenSlot:77-84` already requires `typeof libraryUuid === 'string' && length > 0`, so `libraryUuidResolvable` is unconditionally `true` on that path and rule 1 cannot fire there. Worth stating because the legacy branch reads the slot through a raw `as string | undefined` cast at `:148` with no type guard of its own.
- **The three selector tests are real red-before-green.** With the old `if (!libraryUuid) return undefined`, all three inputs (absent / `''` / `12345`) resolve to `undefined` through `libraryUuidForClonedCharacter:108-116`'s `typeof === 'string' && length > 0` test, the selector `continue`s, and each assertion fails on an empty array. The non-string case is the one that would have been easy to get wrong and is covered.
- **Both release-notes surfaces landed, in the right places** — `RELEASE_NOTES.md:75`, inside the top `# Castwright 1.15.0` section (next header is at `:235`), and `docs/release-notes-next.md:103`.
- **`Closes #2054` is present in the PR body, outside any code span.** The first commit additionally carries `Closes #2912` / `Refs #2054`, which is correct for the two-issue split.
- **No derived API surface.** `CloneUnready` appears in no `openapi.yaml` schema, so `src/lib/api-types.ts` needs no regeneration. `src/data/help-failures.ts` keys on a `cloned-voice-broken` code rather than per-reason, so the offline Help catalogue needs no new entry. `voice-library.ts:1468`'s `CloneAssignBlock` is a separate assign-time type that correctly gains nothing (assign always writes a uuid).
- **On-box acceptance is correctly not owed.** The new CTA is a pure client-side navigation (`changeView('cast')` + `setOpenProfileId`), and A21's run sheet is scoped to the two hardware-dependent repair CTAs (`Add transcript`, `Retry derive`), neither of which this touches. Say so in the body rather than leaving it blank.

### Verdict

No 🔴 or 🟠 findings. The behaviour change is correct, correctly ordered, genuinely co-oracle-tested, and provably cannot fire on a cast that would have rendered — I tried three separate routes to a false positive and each was closed by construction rather than by accident.

Five 🟡 cleanup findings. Per `findings-triage.md` all five are fixed in this round, and none of them re-triggers a review pass. Finding 1 is the one that actually matters: the renumbering is a class, and this PR fixed three files' worth of it and stopped, leaving `voice-library.ts`'s `[R3]` safety argument citing three rules that all now mean something else.

dudarenok-maker:
## PR review — triage of pass 1 (head 82d1995a)

Pass 1 surfaced 0 🔴/🟠 and five 🟡 cleanup-only findings. Per the review-gate's re-review trigger, 🟡-only findings are fixed this round but do not reopen the review loop — so this is a triage/disposition note, not a re-review.

All five fixed, one dispatched fix agent per finding, each verified green before commit:

1. Stale rule-number cross-references (~18 sites) — **FIXED** `534e2e02`
2. Stale `docs/BACKLOG.md` #2054 row — **FIXED** `0e9c6ea`
3. Unreachable guard in `clone-readiness-selectors.ts`'s `buildInput` — **FIXED** `0ea6683f`
4. e2e test missing engine-name assertion — **FIXED** `c18b7794`
5. No guard against a future unpaired `CloneUnready` state — **FIXED** `82d1995a`

Frontend + server typecheck clean, and all touched test suites (frontend: `clone-readiness-selectors`, `clone-readiness-gate`, `start-generation-flow`; server: `clone-readiness`, `clone-readiness-contract`, `voice-library`) re-run green after all five commits stacked. Also declared in the PR body under "Also fixed, found in passing".

### Verdict
No blocking issues remain. Ready to merge once cloud `verify.yml` is green on the new head.

PR #3016, opened 2026-09-06
feat(server): A36 wider severity band for small synthetic-only audition pools

### LINKED ISSUE #2934: ops — A36: widen the severity band for small synthetic-only audition pools
Register row **A36** ([#1969](https://github.com/dudarenok-maker/Castwright/issues/1969), PR [#2402](https://github.com/dudarenok-maker/Castwright/pull/2402), [#2700](https://github.com/dudarenok-maker/Castwright/issues/2700)) found (2026-08-29, real on-box run) a genuine false-positive: a correctly-cast voice, re-rendered against fresh text, was flagged `voice-mismatch`/`severe` because its audition pool was a small (N=6), synthetic-only Phase B rebuild — same engine, same voice, same controlled acoustic conditions — which clusters far tighter around its own centroid (`pSevere`/`pBand` within ~0.02 of `cleanMean`, `server/src/audio/render-integrity/score.ts`'s `percentile()`/`CUTOFFS`, called from `aggregate.ts:242-243`/`:280-281`) than the natural cosine variance of a genuinely-correct new render on different content. This is a sharper version of the already-documented "thin ~0.05-wide over-flag band for the tightest voices" caveat in `score.ts`, but worse for a Phase B pool because it's both smaller and more homogeneous than any real in-book anchor set the calibration was tuned against.

## Owner ruling, 2026-09-05

Presented two options (widen the band specifically for synthetic-only pools / leave as-is and document the risk). **Chosen: widen for synthetic-only pools** — add a distinct, wider severity threshold specifically for small synthetic-only pools, calibrated separately from the normal (real-anchor) band, so a correctly-cast voice doesn't get flagged just for having less natural variance than a real anchor pool would.

## Scope

1. Find how a Phase B / synthetic-only-pool build is distinguished from a normal in-book-anchor pool in `audition-centroid.ts` (the register's own run refers to "Phase B, anchors dropped, synthetic-only pool topped up and rebuilt").
2. Implement a wider, separately-calibrated `pSevere`/`pBand` threshold for that case, test-first — do not touch the normal-pool band.
3. Confirm on real hardware: the same false-positive scenario (correct voice, fresh text, synthetic-only N=6 pool) no longer flags `severe`, AND a genuinely wrong voice against the same pool still does.
4. Fold into the register.

Decomposed into an agent-executable chain — see linked children.

dudarenok-maker:
## Summary

A small (N=6), synthetic-only Phase B audition pool clusters far tighter than real render-to-render variance, so the normal percentile cutoffs false-flagged a correctly-cast voice re-rendered on fresh text as `voice-mismatch`/`severe` (register row A36, 2026-08-29 on-box finding). This adds a separately-calibrated sigma band (`syntheticOnlySpread`) used only when `auditionCentroid` reports a synthetic-only pool; the real-anchor (normal, in-book) path is untouched.

Confirmed on real hardware (live Coqui/XTTS sidecar, unmocked `scoreBook()`): the correctly-cast voice on fresh text no longer flags severe (cosine 0.90230 → `inconclusive`), and a genuinely wrong voice against the same pool still does (cosine 0.16213 → `voice-mismatch`/`severe`). Full trace in `docs/testing/onbox-a36-results/step-3-onbox.md`.

Register row A36 fully discharged and removed (`docs/testing/onbox-acceptance-register.md`, `-live-view.html`); `npm run check:onbox-register` is green.

Closes #2934, Refs #2435

## Test plan

- [x] `aggregate-audition-synthetic-band.test.ts` (5 new tests) — mutation-verified: reverting the `syntheticOnly` branch to always use the shared percentile band reddens exactly the two false-positive-reproduction tests, confirming they test the real behavior.
- [x] `score.test.ts`, `audition-centroid.test.ts`, `aggregate.test.ts` — all green, unmodified normal (real-anchor) path assertions unaffected.
- [x] Real on-box confirmation (live sidecar, unmocked): correct-voice/fresh-text no longer `severe`; wrong-voice still `severe`. See `docs/testing/onbox-a36-results/step-3-onbox.md`.
- [x] `npm run check:onbox-register` — green.
- [ ] cloud `verify.yml` (required status check)


dudarenok-maker:
## PR review — pass 1 (head 4951d558, depth medium)

Reviewed at `4951d558cbac8d91eaec4fe65617a53dd975a2ee` in a detached scratch worktree. Findings only — no tracked file was modified.

---

### 🔴 Blocking

**1. `syntheticOnlySpread` is unbounded: one degenerate render in a six-render synthetic pool collapses the severe edge, and a wide-enough pool makes `severe` unreachable entirely.** *(correctness bug)*
`server/src/audio/render-integrity/score.ts:96-104`, applied at `server/src/audio/render-integrity/aggregate.ts:287-289`

`pSevere = mean − 3σ` has no lower clamp and no dispersion/tightness guard. The calibration argument in the `SYNTHETIC_ONLY_CUTOFFS` doc comment (`score.ts:56-81`) is entirely about a *tight* cluster ("clusters far tighter than that… N=6, cosines 0.9629±0.02"), but the code keys only off `syntheticOnly` — never off how tight the pool actually is. On a pool that is *not* tight, the same formula produces a band far **wider** than the percentile band it replaces, in the unsafe direction.

Measured, using this branch's own `buildCentroid` + `cosineToCentroid` on real 6-vector pools:

| pool (6 unit vectors) | `bimodal` | cosines-to-centroid | new `pSevere` | old `pSevere` (p6) |
|---|---|---|---|---|
| tight, register shape | false | 0.9993 … 1.0000 | 0.99896 | 0.99934 |
| **5 good + 1 degenerate render** | **false** | 0.4369, 0.9607, 0.9734, 0.9836, 0.9914, 0.9967 | **0.2809** | **0.5940** |
| 4 good + 2 bad | true | 0.5417 … 0.9388 | 0.2988 | 0.5541 |
| 4 good + 2 near-orthogonal | true | −0.3497 … 0.9123 | **−1.1873** | −0.3211 |

Concrete failure: a character reaches the synthetic-only path (either route — Phase B after a bimodal blend, exactly the register's own A36 shape, or Phase A with zero anchors), and **one** of its six audition renders is acoustically degenerate but still clears `MIN_DURATION_SEC` — a wrong-language clone render, a truncated render, an instruct-leak. The pool is *not* flagged bimodal (row 2 above: `bimodal=false`), so nothing stops it, and Phase B's pool is documented as never re-checked for bimodality anyway (`audition-centroid.ts:280-282`). The severe edge drops from 0.594 to 0.281. A genuinely wrong voice landing anywhere in 0.28–0.59 — squarely inside the range this repo's own calibration note cites for real clean/drift boundaries (`score.ts:40-42`: "china-sorrows reads clean at cosine 0.478 while narrator is drift at 0.507") — silently becomes `inconclusive` instead of `voice-mismatch`/`severe`.

I confirmed the end-to-end consequence through unmocked `scoreBook()` with a synthetic-only pool at cosines `[0.20, 0.30, 0.95, 0.96, 0.97, 0.98]` (new band `pSevere=−0.2886, pBand=0.2190`; old band `pSevere=0.2300, pBand=0.2500`): a wrong-voice render at **cosine 0.16** — the same magnitude as this PR's own on-box true-positive (0.16213, `step-3-onbox.md:79`) — comes back **`inconclusive`/`inconclusive`** rather than `voice-mismatch`/`severe`. A negative `pSevere` is below the entire achievable cosine range, so the severe tier is not merely widened, it is switched off for that character.

This directly falsifies the claim the PR body and two of its test titles make ("the band widens, it does not disable detection"). It is also a regression against `main`, not just a missed case: `main` would have flagged all four of those.

---

### 🟠 Significant

**2. The fix never reaches a book that has already been scored — including the one register row A36 was filed from.** *(correctness bug)*
`server/src/audio/render-integrity/aggregate.ts:251-257`

A persisted `centroids.json` row with `referenceKind: 'audition'` whose recorded voice identity still matches is returned verbatim via `persistedAsRef(persisted)`, short-circuiting before `auditionCentroid` is ever called. `render-integrity.centroids.json` (`centroids-io.ts`) carries no schema version and nothing in this branch invalidates or migrates pre-fix rows, so the **percentile-derived** `pSevere`/`pBand` written by any earlier run survive the upgrade indefinitely — until the voice, model key, language or clone flag changes.

Verified through `scoreBook()` against a book seeded with a pre-fix row using the register's own recorded 2026-08-29 values (`cleanMean 0.9629, pSevere 0.9409, pBand 0.9446`): `auditionCentroid` call count **0**, band still `{pSevere: 0.9409, pBand: 0.9446}`, and a correct-voice render at cosine 0.928 still scores **`voice-mismatch`/`severe`**. That is the exact A36 false positive, alive after the fix.

The on-box confirmation in `docs/testing/onbox-a36-results/step-3-onbox.md` cannot have caught this: it builds a fresh `mkdtemp` fixture with no prior `centroids.json`, so it only ever exercises the rebuild path. Register row A36 is discharged on evidence that does not cover the population it was filed from.

---

### 🟡 Minor

**3. The live view carries a stale generated figure this change made owed.** *(chore)*
`docs/testing/onbox-acceptance-register-live-view.html:228`

The PR updates the strip (65→64) and the glance table (`glance:A` 39→38) but leaves the group-A section header at `<span class="gcount">39 rows</span>`. Running `npm run register:build` on this branch rewrites exactly that one line to `38 rows` — i.e. the committed page is internally inconsistent and the file that gets published to the URL in the register header would show a wrong count. `npm run check:onbox-register` is green because it compares the *glance table* against body rows and is blind to the section-header figure; per CLAUDE.md these derived figures are generated, not hand-edited, and hand-editing two of the three is how this slipped.

**4. No release-notes pair, and no explicit not-applicable.** *(chore)*
`docs/release-notes-next.md` / `RELEASE_NOTES.md`

Neither file is in the diff, and the PR body's Test plan does not declare the step N/A. `RELEASE_NOTES.md` has a live in-progress `# Castwright 1.15.0` section. This change alters which renders the Chapter QA panel flags as voice mismatches — operator-visible, so before-shipping step 5 applies. If the argument is that it is not user-visible, that argument needs to be *in the PR body*, not absent.

**5. Two JSDoc contracts the change made false.** *(chore)*
`server/src/audio/render-integrity/score.ts:161-167`, `server/src/audio/render-integrity/centroids-io.ts:39` and `:41`

`scoreSegment`'s `@param spread` still documents both fields as "percentile value at `CUTOFFS.severeEdgePctl`/`bandUpperPctl`… Passed in by the aggregate (Task 9) after calling `percentile()`". `CharacterCentroid.pSevere`/`pBand` say the same. For the synthetic-only path neither is a percentile of anything, and the persisted row now stores a sigma band under a field documented as a percentile — which is precisely the ambiguity finding 2 turns on.

**6. Two of the five new end-to-end tests cannot fail.** *(chore — coverage gap)*
`server/src/audio/render-integrity/aggregate-audition-synthetic-band.test.ts:109` and `:122`

Both are titled as guarding "the band is wider, not disabled" / "it does not disable detection", but neither constrains the band at any realistic value. I verified they stay green under *both* mutations: reverting the `syntheticOnly` branch to the shared percentile band, **and** widening `severeSigma` 3→10 / `bandSigma` 1.5→5. Against the pinned tight pool (mean 0.9629, σ 0.01732), test `:109` can only fail if `pBand > 0.995` — unreachable, since `pBand ≤ mean` by construction — and test `:122` needs `severeSigma > 49` before cosine 0.10 stops being severe. The constants themselves *are* pinned (`score.test.ts:60-61`), so an accidental edit is caught there; what is missing is any behavioural test over a pool that is synthetic-only but **not** tight — the gap finding 1 lives in.

---

### ✅ Verified

- **The PR body's mutation claim holds exactly.** Reverting the `syntheticOnly` branch in `aggregate.ts:287-289` to the shared percentile band reddens precisely the two false-positive-reproduction tests (`:83`, `:97`), both with `AssertionError: expected 'voice-mismatch' not to be 'voice-mismatch'` — red for the reason claimed, not an incidental one. The other three stay green.
- **The real-anchor path is genuinely untouched.** The `in-book` branch (`aggregate.ts:238-245`) is unchanged; a blended pool with real anchors routes to `percentile` via `syntheticOnly === false`; the in-book test asserts `auditionSpy` was never called and still gets `voice-mismatch`/`severe`.
- **The `syntheticOnly` discriminator classifies correctly at all four structural return sites.** Read against `audition-centroid.ts:246-279`: Phase A too-thin and Phase A success both derive it from `existingAnchors.length === 0`; both Phase B returns hardcode `true`, which is correct because Phase B drops the anchors by construction. No path reports a blended real+synthetic pool as synthetic-only, and none reports a zero-anchor pool as real-anchor. The 6–9-anchor "already meets target, no new renders" case correctly stays `false`. Five unit tests cover these.
- `npm run typecheck` clean; 153 tests green across `render-integrity/`, `qa-report.test.ts`, `chapter-qa-repair.test.ts` at head.
- `npm run check:onbox-register` green. Register bookkeeping is otherwise careful and correct: outcome recorded rather than deleted, allocate-once respected (A36 retired not reused), `next-id` untouched, correction log kept to one "Prior change" line.
- `Closes #2934, Refs #2435` present in the body and outside any code span.
- No `cast.json` writes, no new config knob, no generated code artifact (`api-types.ts`, `BACKLOG.md`, brand PNGs) in scope.

---

### Verdict

**Blocking findings present — do not merge as-is.** One 🔴 and one 🟠 correctness bug, plus four 🟡 chores. Per the findings-triage rule all six are fixed in this round; the 🔴/🟠 pair re-triggers a review pass once fixed and pushed.

The core of the change is sound and well-evidenced for the case it was calibrated against — the discriminator is right, the real-anchor path really is untouched, and the mutation claim survives independent checking. What is missing is a bound: the band is calibrated on one tight pool and then applied to every synthetic-only pool regardless of dispersion, and the fix is not reachable for any book that already has a persisted audition reference.

dudarenok-maker:
## PR review — pass 2 (head 079da5a1, depth medium)

Re-review of the six fix commits. Deltas only — pass 1's ✅ items are not re-listed.

**Depth re-derived** from the PR's current nine-commit set (`c7a1dfac` feat(server), `87c41796`/`4951d558` docs(docs), `e862e978`/`d8b39717` fix(server), `c4b6ba5e` chore(docs), `ef6bed65` docs(docs), `ee5ff9b8` docs(server), `079da5a1` test(server)): no `refactor`/`perf`, and every commit is single-scope (no `type(a,b)` multi-scope subject). Highest tier any single commit earns on its own is `feat`/`fix` single-scope → **medium**. Unchanged from pass 1.

Verified at head: `npm run typecheck` clean; `src/audio/render-integrity/` **13 files / 123 tests green**; `npm run register:build` produces no diff; `npm run check:onbox-register` OK. Reviewed in a detached scratch worktree at `079da5a1`; no tracked file modified.

---

### 1. 🔴 Unbounded `syntheticOnlySpread` — **STILL OPEN** *(correctness bug)*
`server/src/audio/render-integrity/score.ts:107-139`

The guard fires, and it does close the catastrophic case (a negative `pSevere`, severe tier switched off). It does **not** close the finding. The fallback went to a **lower** percentile than the one it is replacing, and lower is *less* protective, not "much tighter" as `score.ts:129-130` claims.

`percentile()` (`score.ts:150-158`) is monotone non-decreasing in `pctl`, so `percentile(sorted, 1) ≤ percentile(sorted, 6)` **for every pool**. `pSevere` is a floor — a segment is severe iff `cosine < pSevere` (`score.ts:226`) — so a *lower* `pSevere` flags *fewer* segments. The dispersion fallback is therefore provably never more protective than the `CUTOFFS.severeEdgePctl` band `main` uses. Measured `pSevere` delta vs `main`'s p6 on five pools: **−0.13095, −0.02500, −0.00750, −0.01458, 0.00000** — ≤ 0 in every case.

On pass 1's own row-2 pool `[0.4369, 0.9607, 0.9734, 0.9836, 0.9914, 0.9967]` (mean 0.89045, population std 0.20317, range 0.55980 → `isDispersed = true`):

| band | pSevere | pBand |
|---|---|---|
| `main` (p6 / p10) | **0.59404** | 0.69880 |
| this PR before `e862e978` (raw sigma) | 0.28093 | 0.58569 |
| this PR at `079da5a1` (dispersion fallback) | **0.46309** | 0.56785 |

The unsafe window narrowed from 0.281–0.594 to **0.463–0.594**. It did not close. Unmocked `scoreBook()` against that pool as the `syntheticOnly` audition pool, one fresh render, verdict actually written to `ch1.render-integrity.json`:

| fresh-render cosine | branch @ `079da5a1` | `main` |
|---|---|---|
| 0.16 | voice-mismatch / severe | voice-mismatch / severe |
| 0.45 | voice-mismatch / severe | voice-mismatch / severe |
| **0.50** | **inconclusive / inconclusive** | voice-mismatch / severe |
| **0.55** | **inconclusive / inconclusive** | voice-mismatch / severe |
| **0.58** | **voice-match / null** | voice-mismatch / severe |

Persisted band on every one of those runs: `{pSevere: 0.46308999109059407, pBand: 0.5678499918766857}`.

So a genuinely wrong voice at cosine 0.50 is still downgraded from `severe` to `inconclusive`, and at 0.58 it is reported as a **clean `voice-match`** — a strictly worse outcome than pass 1's `inconclusive`. That range is exactly where this repo's own calibration note puts the real clean/drift boundary (`score.ts:40-42`: china-sorrows clean at 0.478, narrator drift at 0.507). It remains a regression against `main`, not a missed case.

`e862e978` closed precisely the numeric case its own new test uses (cosine 0.16, where `main` and the branch agree) and nothing between 0.463 and 0.594. The fix's premise is stated inverted in two places — `score.ts:129` `// much tighter than CUTOFFS.severeEdgePctl (6)` and `:98-99` "falls back to a tighter percentile-based band (using lower percentiles than the real-anchor path)". Lower percentile ⇒ lower cutoff ⇒ looser. Any bound that actually discharges this finding has to be **≥** what `main` used, i.e. a higher percentile or an explicit clamp, not a lower one.

Two further properties of the guard, same site, same fix:

- **The second disjunct is dead.** `std > Math.max(0.05, poolRange * 0.1)` — for `n = 6` the maximum attainable `range/std` is `√(2n) ≈ 3.46`, so `poolRange * 0.1 ≤ 0.346·std < std` always. The threshold is effectively the bare constant `0.05`; `poolRange * 0.1` can never be the binding term at audition-pool sizes (it would need `n ≥ 50`).
- **The threshold is a cliff, not a bound.** Pool `[0.85, 0.88, 0.91, 0.94, 0.97, 1.00]` has std 0.05123, one thousandth over the threshold: `pSevere` jumps from the sigma value 0.77130 to the fallback 0.85150. An arbitrarily small change in pool spread swings the severe cutoff by ~0.08, so two near-identical books get materially different verdicts on identical audio.

### 2. 🟠 Pre-fix persisted rows bypass the fix — **RESOLVED for the case it names**, but the fix introduces a new 🔴 (see New A)
`server/src/audio/render-integrity/aggregate.ts:206`

Verified as claimed: a persisted `audition` row lacking `bandMethod` is rejected by `matchesCurrentVoice` and rebuilt, so pass 1's exact repro (register row's `pSevere 0.9409` reused, correct voice at 0.928 flagged severe) no longer occurs.

Also verified as **not** causing needless rebuilds on the untouched path: an in-book real-anchor character short-circuits at `aggregate.ts:249` before any persisted row is consulted, and `matchesCurrentVoice` already returned `false` for `referenceKind !== 'audition'` (`:200`). No in-book behaviour changes. A post-fix non-synthetic audition row carries `bandMethod: 'percentile'` and is reused normally.

What is **not** true is the commit message's closing claim — "then cached with `bandMethod` for efficient reuse". It is not. See New A.

### 3. 🟡 Stale generated figure in the live view — **VERIFIED RESOLVED**
`c4b6ba5e` sets the group-A section header to `38 rows`. Re-ran `npm run register:build` at head: **no diff** (`git status --porcelain` empty afterwards). `npm run check:onbox-register` green.

### 4. 🟡 Missing release-notes pair — **RESOLVED**, with one 🟡 follow-on (New D)
Both files now carry an entry. The user-facing line in `RELEASE_NOTES.md:80` reads naturally and is accurate for the half it describes. (I checked and am *not* reporting the surrounding blank lines: that 1.15.0 bullet list already contains 25 blank-line separations, so it is already a loose list and the insertion changes no rendering.)

### 5. 🟡 Stale JSDoc — **PARTIALLY RESOLVED**
`ee5ff9b8` fixes the two original stale statements (`centroids-io.ts:39/41`, `score.ts` `@param spread`). But the replacement text and the two `fix` commits introduce **three new false statements**, which is the same defect shape the finding named:

- `score.ts:122` — `// > 5% mean or > 10% of range`. Wrong twice. The code is `std > Math.max(0.05, poolRange * 0.1)`: `0.05` is an **absolute** constant, not 5% of the mean (at mean 0.5 the two differ by 2×), and `>` against a `Math.max` is **AND**, not the "or" the comment states (an "or" would be `Math.min`). The same "std dev relative to mean" claim is repeated at `score.ts:97`.
- `score.ts:129-130` — `// much tighter than CUTOFFS.severeEdgePctl (6)` / `// tighter than CUTOFFS.bandUpperPctl (10)`. Inverted, per finding 1.
- `aggregate.ts:194-195` — "a pre-fix band method (missing `bandMethod` field **or 'percentile' value**, pre-sigma-band rows) … must not be trusted as-is". The code at `:206` rejects only `undefined`; a `bandMethod: 'percentile'` row is accepted and reused. The code is right here and the doc is wrong.

### 6. 🟡 Mutation-insensitive tests — **PARTIALLY RESOLVED**
`079da5a1` adds three tests but did not touch `:109`/`:122`, the two the finding actually named. I mutation-tested every constant in the new code against `aggregate-audition-synthetic-band.test.ts` + `score.test.ts` (35 tests):

| mutation | result |
|---|---|
| `severeSigma: 3 → 1` | 4 red — the two pre-existing 0.928/0.934 tests, the pre-fix-row test, and `:213`. **`:194` stays green.** |
| `severeSigma: 3 → 7` | 3 red — `:194` + `score.test.ts`'s two literal pins |
| `bandSigma: 1.5 → 0.5` | 1 red — `:213` ✅ works as claimed |
| `isDispersed = false` (guard off) | 1 red — `:166` ✅ |
| threshold `0.05 → 0.5` | 1 red — `:166` ✅ |
| threshold `0.05 → 0.005` | 3 red ✅ (over-trigger caught) |
| `percentile(sorted, 1) → percentile(sorted, 6)` | **35/35 green** |
| `percentile(sorted, 1) → percentile(sorted, 0)` | **35/35 green** |

Two problems:

- **`:194` is itself a test that cannot fail for the reason claimed** — the exact shape finding 6 flagged. Its title says "catches `severeSigma` narrowing mutations" and its comment says "Reducing `severeSigma` (e.g., to 1) would raise `pSevere` to ~0.927, making 0.85 NOT severe". Both are false: at `severeSigma = 1`, `pSevere = 0.9629 − 0.017321 = 0.94558`, and `0.85 < 0.94558` is still severe. Empirically confirmed by M1 above. `:194` requires `severeSigma > 6.518` to redden, i.e. it catches *widening*, and narrowing was already caught by the pre-existing 0.928/0.934 tests. The commit message's own claim "severeSigma narrowing (3 → 1): caught by edge-case test #1" is contradicted by the run.
- **The dispersion fallback's own constants have zero coverage.** `1`, `5`, and `poolRange * 0.1` are unnamed inline magic numbers with no pin anywhere (unlike `severeSigma`/`bandSigma`, pinned at `score.test.ts:60-61`). Both directions of mutating the crux constant — the number that finding 1 turns on — pass the entire suite. There is still no test over a synthetic-only pool that is dispersed **and** whose fresh render lands between the fallback cutoff and `main`'s.

---

## New findings

### A. 🔴 `persistedAsRef` drops `bandMethod`, so the tag never survives a reuse pass — the reference is re-auditioned every other run, and can be permanently destroyed *(correctness bug, introduced by `d8b39717`)*
`server/src/audio/render-integrity/aggregate.ts:187-189` (vs `:166` and `:718`)

`d8b39717` adds `bandMethod` to `CharacterReference` (`:163-166`) and writes it on the *build* path (`:305-306`), but `persistedAsRef` — the *reuse* path — was not updated:

```ts
function persistedAsRef(row: CharacterCentroid): CharacterReference {
  return { centroid: row.centroid, cleanMean: row.cleanMean, pSevere: row.pSevere, pBand: row.pBand, referenceKind: row.referenceKind, auditionVoice: row.auditionVoice };
}
```

Every resolved outcome is written straight back: `centroidsMap[charId] = { characterId: charId, ...outcome.ref }; await writeCentroids(...)` (`:718-719`). So a reuse pass rewrites the row **without** `bandMethod`, and the next pass sees a "pre-fix" row again. This is exactly the hazard the sibling field's own comment warns about at `:158-162` ("Carried on the persisted row so a match-checked reuse rewrites the same identity back instead of silently dropping it") — the pattern was mirrored half-way.

Measured with real `scoreBook()`, four consecutive passes over one book dir, tight register pool, `auditionCentroid` spied:

```
pass 1: auditionCentroid calls = 1   persisted bandMethod = "synthetic-sigma"
pass 2: auditionCentroid calls = 1   persisted bandMethod = undefined
        row keys = characterId,centroid,cleanMean,pSevere,pBand,referenceKind,auditionVoice
pass 3: auditionCentroid calls = 2   persisted bandMethod = "synthetic-sigma"
pass 4: auditionCentroid calls = 2
```

It oscillates forever. Consequences, in ascending order:

1. **Every other `scoreBook` pass re-runs a full Option-B audition** for every audition-referenced character. In production `synthFn` defaults to `selectTtsProvider(modelKey).synthesize(...)` (`audition-centroid.ts:176-184`) — real TTS renders on the GPU, ~`AUDITION_POOL_TARGET_N` of them per character, not local math. Before this PR a stable audition row was reused indefinitely.
2. **Verdict churn on unchanged audio.** Each re-audition draws a fresh stochastic pool, so `pSevere`/`pBand` move and the same segment can flip verdict between two runs where nothing changed.
3. **A path to permanent loss of the reference.** A re-audition that hits an unavailable sidecar returns `transient-failure`; after `MAX_PENDING_ATTEMPTS = 3` (`aggregate.ts:638, 702-714`) the row is overwritten with `TOO_SHORT_REF`, and `too-short` is absorbing (`:260-262`). Every segment for that character then scores `inconclusive` indefinitely, until enough real anchors appear for the in-book path. Before this PR a working audition reference could not be lost that way, because it was never re-derived.

`aggregate-audition-voice-reassign.test.ts:137` asserts `expect(c.bandMethod).toBeDefined()` — but only after pass 1, and its pass 3 reassigns the voice, so it rebuilds for an unrelated reason. Nothing in the suite runs a third *identical* pass, which is the minimum needed to see this.

### B. 🟡 `bandMethod` cannot distinguish the two computations three doc blocks say it distinguishes *(chore — the defect is in the instrument)*
`server/src/audio/render-integrity/aggregate.ts:305`, contradicting `centroids-io.ts:39-49` and `score.ts:204-206`

`bandMethod` is set purely from `audition.syntheticOnly`, so a row whose band came from the **dispersion percentile fallback** is persisted as `'synthetic-sigma'`. Confirmed: the degenerate-pool run above persists `bandMethod: "synthetic-sigma"` with `pSevere: 0.46309`, which is `percentile(sorted, 1)`, not a sigma value. Meanwhile all three doc blocks `ee5ff9b8` wrote instruct the reader to "Check the `bandMethod` field to determine which computation was used" and enumerate the fallback as a distinguishable case. It is not distinguishable. Nothing reads the field for that purpose today (only `:206`, for presence), so this is doc-vs-data, not a live wrong output — but it is a contract written this round that the same round's code does not satisfy.

### C. 🟡 The `release-notes-next.md` entry is filed under the wrong feature section *(chore)*
`docs/release-notes-next.md:102`

The entry lands inside `### 🎤 Clone a voice — and cast it on either engine (new) — fs-38`, under `## ✨ Headline features`. A36 is srv-36 render-integrity / Chapter QA scoring (parent #2934); it has nothing to do with voice cloning and is not a headline feature. `## 🔊 Generation quality & engine health` (`:228`) is the section this belongs in. The register is read section-by-section at cut time, so a fix filed under an unrelated headline feature is what produces a mis-grouped user-facing note in the next cut.

---

## Verdict

**Do not merge.** One prior 🔴 remains open and one new 🔴 was introduced by the fixes.

- Finding 1 is **not** closed. `e862e978` moved `pSevere` from 0.28093 to 0.46309 on the reference pool, but `main`'s value is 0.59404 and `percentile(1) ≤ percentile(6)` holds structurally, so the fallback can never be more protective than the band it replaces. A wrong voice at cosine 0.50 is still `inconclusive`, and at 0.58 it is now `voice-match`. The fix's own comments state its premise backwards, which is why it landed on the wrong side.
- Finding 2's stated gap **is** closed, and the untouched real-anchor path is genuinely unaffected — but the tag it introduced does not survive a reuse pass, so the fix trades one persisted-row bug for a rebuild oscillation that re-runs real TTS auditions every other scoring pass and can permanently degrade a character to the absorbing `too-short` state.
- Findings 3 and 4 are clean. Findings 5 and 6 each half-landed: 5 replaced two false statements with three new ones, and 6 shipped a "mutation-sensitivity" test that is green under the exact mutation it names, while the new constants that finding 1 turns on carry no coverage in either direction.

Per the loop cap this is re-review round 1 of 2.

dudarenok-maker:
## PR review — pass 3 (head 8dfc770a, depth medium)

Re-review of the three fix commits (`f7b70b3c`, `4e53c3b9`, `8dfc770a`). Deltas only — pass 1/2's ✅ items are not re-listed.

**Depth re-derived** from the PR's current twelve-commit set (`c7a1dfac` feat(server); `87c41796`/`4951d558`/`ef6bed65`/`8dfc770a` docs(docs); `e862e978`/`d8b39717`/`f7b70b3c`/`4e53c3b9` fix(server); `c4b6ba5e` chore(docs); `ee5ff9b8` docs(server); `079da5a1` test(server)): no `refactor`/`perf`, every commit single-scope. Highest tier any single commit earns on its own is `feat`/`fix` single-scope → **medium**. Unchanged from passes 1 and 2.

Verified at head, in a detached scratch worktree at `8dfc770a` (torn down; no tracked file modified): `npm run typecheck` clean; `src/audio/render-integrity/` + `qa-report.test.ts` + `chapter-qa-repair.test.ts` **15 files / 152 tests green**; `npm run register:build` produces no diff; `npm run check:onbox-register` OK. **`npm run test:hooks` FAILS — see New E.**

---

### 1. 🔴 Dispersion fallback went the wrong direction — **VERIFIED RESOLVED**
`server/src/audio/render-integrity/score.ts:107-142`

`f7b70b3c` replaces `percentile(sorted, 1)`/`percentile(sorted, 5)` with `percentile(sorted, CUTOFFS.severeEdgePctl)`/`percentile(sorted, CUTOFFS.bandUpperPctl)`. Independently recomputed, not taken from the commit message.

Pure math on pass 2's own reference pool `[0.4369, 0.9607, 0.9734, 0.9836, 0.9914, 0.9967]`:

```
spread: { pSevere: 0.5940399999999999, pBand: 0.6988, usedSigmaBand: false }
main:   { pSevere: 0.5940399999999999, pBand: 0.6988 }
```

Byte-identical to `main`'s p6/p10, and `pSevere` is the 0.59404 pass 2 named as the target. Sweep across ten pools — the delta vs `main`'s `pSevere` is now **exactly 0.00000 on every dispersed pool** (was −0.13095 / −0.02500 / −0.00750 / −0.01458 at `079da5a1`):

| pool | std | usedSigmaBand | pSevere | main p6 | delta |
|---|---|---|---|---|---|
| degenerate-1bad | 0.20317 | false | 0.59404 | 0.59404 | **0** |
| 4good2bad | 0.15346 | false | 0.56519 | 0.56519 | **0** |
| 4good2orth | 0.49594 | false | −0.30479 | −0.30479 | **0** |
| tight-register | 0.01732 | true | 0.91094 | 0.94290 | −0.03196 (the intended A36 widening) |

End-to-end through unmocked `scoreBook()` with that pool as the `syntheticOnly` audition pool, verdict read back off `ch1.render-integrity.json` (persisted band `{pSevere: 0.5940399920732087, pBand: 0.6987999928593004}` on every run):

| fresh-render cosine | at `8dfc770a` | at `079da5a1` (pass 2) | `main` |
|---|---|---|---|
| 0.16 | voice-mismatch / severe | voice-mismatch / severe | voice-mismatch / severe |
| 0.45 | voice-mismatch / severe | voice-mismatch / severe | voice-mismatch / severe |
| **0.50** | **voice-mismatch / severe** | inconclusive / inconclusive | voice-mismatch / severe |
| **0.55** | **voice-mismatch / severe** | inconclusive / inconclusive | voice-mismatch / severe |
| **0.58** | **voice-mismatch / severe** | voice-match / null | voice-mismatch / severe |
| 0.62 | inconclusive / inconclusive | — | inconclusive / inconclusive |
| 0.70 | voice-match / null | — | voice-match / null |

The 0.463–0.594 unsafe window is closed and the dispersed path is now bit-for-bit `main`. The dead `poolRange * 0.1` disjunct is gone, and the two inverted comments (`score.ts:129-130`, `:98-99`) are gone with it.

**The `usedSigmaBand` → `bandMethod` wiring is real, not just declared**, and it does **not** reintroduce an oscillation. Measured over 5 consecutive identical `scoreBook()` passes on one book dir:

```
DEGENERATE pool: pass 1..5 → auditionCentroid calls = 1, bandMethod = "percentile"   (stable)
TIGHT pool:      pass 1..5 → auditionCentroid calls = 1, bandMethod = "synthetic-sigma" (stable)
```

A dispersed synthetic-only pool really is tagged `'percentile'`, not `'synthetic-sigma'`. The borderline-noise oscillation you'd expect from tagging off a threshold cannot occur, for a structural reason: `matchesCurrentVoice` (`aggregate.ts:206`) rejects only `bandMethod === undefined`, so **both** legal values are accepted, and a matching persisted row short-circuits at `aggregate.ts:263-269` before `auditionCentroid` is called at all — the pool is never re-drawn, so `usedSigmaBand` is never re-evaluated after pass 1. Confirmed: a persisted row carrying `bandMethod: 'percentile'` is reused across 3 passes with `auditionCentroid` calls = **0**.

Two residual sub-items from pass 2 are downgraded rather than carried, and reported below as 🟡: the 0.05 cliff (New H) and the unnamed/unpinned threshold constant (also New H).

### 2. 🟠 Pre-fix persisted rows bypass the fix — **RESOLVED** (unchanged from pass 2)
Re-confirmed at head: a persisted row with no `bandMethod` is rebuilt exactly once at pass 1 and stable at passes 2–4 (`auditionCentroid` calls = 1 throughout, `pSevere` settles at the sigma value 0.91094). Mutation M12 (removing the `row.bandMethod === undefined` guard) reddens `d8b39717`'s own pre-fix-row test, so the guard is load-bearing.

### 3. 🟡 Stale generated figure in the live view — **VERIFIED RESOLVED** (unchanged)
### 4. 🟡 Missing release-notes pair — **VERIFIED RESOLVED** (unchanged)

### 5. 🟡 Stale JSDoc — **STILL OPEN, and now with two new instances** *(chore — the defect is in the instrument)*

Two of pass 2's three items are fixed: `score.ts:97`/`:122`'s "std dev relative to mean" and `Math.max`-as-"or" claims are gone, and `score.ts:129-130`'s inverted "much tighter than" comments are gone. The third was not touched, and `f7b70b3c` added two more. Details in New F.

### 6. 🟡 Mutation-insensitive tests — **MOSTLY RESOLVED**, two gaps left
I re-ran a 15-mutation battery against `score.test.ts` + `aggregate-audition-synthetic-band.test.ts` + `aggregate-audition-voice-reassign.test.ts` + `aggregate.test.ts` (58 tests). Pass 2's headline gap — "both directions of mutating the crux constant pass the entire suite" — is closed:

| mutation | result at `8dfc770a` |
|---|---|
| fallback `severeEdgePctl` → `1` (pass 2's wrong direction) | **2 red** ✅ |
| fallback `bandUpperPctl` → `5` | **1 red** ✅ |
| fallback `severeEdgePctl` → `20` | **1 red** ✅ |
| dispersion threshold `0.05` → `0.5` (guard off) | **3 red** ✅ |
| dispersion threshold `0.05` → `0.005` (over-trigger) | **7 red** ✅ |
| `usedSigmaBand: false` → `true` (dispersed branch) | **1 red** ✅ |
| `usedSigmaBand: true` → `false` (tight branch) | **1 red** ✅ |
| `persistedAsRef` drops `bandMethod` (revert `4e53c3b9`) | **1 red** ✅ |
| `bandMethod = 'synthetic-sigma'` always | **58/58 green** ❌ |
| `bandMethod = 'percentile'` always | **58/58 green** ❌ |
| dispersion threshold `0.05` → `0.049` | 58/58 green (see New H) |

Remaining gaps: the `usedSigmaBand`-driven tagging itself is uncovered (New G), and pass 2's other sub-item — `aggregate-audition-synthetic-band.test.ts:194`, the test whose title and comment claim it "catches `severeSigma` narrowing mutations" — was not touched by any of the three new commits and is still wrong. At `severeSigma = 1`, `pSevere = 0.9629 − 0.017318 = 0.94558`, and `0.85 < 0.94558` is still severe, so the test stays green; it needs `severeSigma > 6.52` to redden, i.e. it catches *widening*. Confirmed empirically: mutating `severeSigma: 3 → 1` reddens 9 tests and that one is not among them.

### A. 🔴 `persistedAsRef` drops `bandMethod` — **VERIFIED RESOLVED**
`server/src/audio/render-integrity/aggregate.ts:188`

`4e53c3b9` adds `bandMethod: row.bandMethod`. I reproduced pass 2's oscillation harness rather than trusting the fix agent's 2-pass test — **5** consecutive identical `scoreBook()` passes over one unchanged synthetic-only character, real `scoreBook`, `auditionCentroid` spied, reading the persisted row's raw keys each pass:

```
pass 1: auditionCalls=1 bandMethod=synthetic-sigma keys=characterId,centroid,cleanMean,pSevere,pBand,referenceKind,bandMethod,auditionVoice
pass 2: auditionCalls=1 bandMethod=synthetic-sigma keys=characterId,centroid,cleanMean,pSevere,pBand,referenceKind,auditionVoice,bandMethod
pass 3: auditionCalls=1 bandMethod=synthetic-sigma keys=...,auditionVoice,bandMethod
pass 4: auditionCalls=1 bandMethod=synthetic-sigma keys=...,auditionVoice,bandMethod
pass 5: auditionCalls=1 bandMethod=synthetic-sigma keys=...,auditionVoice,bandMethod
```

Compare pass 2's measurement (calls 1→1→2→2, `bandMethod` present→absent→present→absent). The field now survives every write-back, the reference is built once and reused indefinitely, and the path to the absorbing `TOO_SHORT_REF` state via repeated re-auditions is gone. Same result on the dispersed pool with `bandMethod: 'percentile'`.

### B. 🟡 `bandMethod` could not distinguish the two computations — **RESOLVED in mechanism**
`f7b70b3c` makes the tag derive from `spread.usedSigmaBand`, and the persisted row for a dispersed pool now reads `bandMethod: "percentile"` with `pSevere: 0.59404` (a percentile value) — the field and the data agree. The *documentation* of that mapping is now wrong in a new way (New F) and the wiring carries no test (New G).

### C. 🟡 Release-notes entry filed under the wrong section — **VERIFIED RESOLVED**
`8dfc770a` moves the entry to `docs/release-notes-next.md:229`, inside `## 🔊 Generation quality & engine health` (heading at `:227`). Pure relocation, byte-identical text. `RELEASE_NOTES.md`'s in-progress `# Castwright 1.15.0` section is a flat bullet list with no subsections, so its entry needs no move.

---

## New findings

### E. 🔴 The PR discharges register row A36 but leaves eight dangling citations of it — `npm run test:hooks` is green on `main` and **red at head**, and that is a `verify.yml` leg in scope for this diff *(correctness bug — a required check this PR turns red)*
`docs/testing/onbox-a36-results/step-1-implement.md:3`, `step-3-onbox.md:3` and `:6`, `server/src/audio/render-integrity/score.ts:66`, `score.test.ts:66`, `aggregate.ts:297`, `aggregate-audition-synthetic-band.test.ts:1` and `:79`

`4951d558` removes the `### A36 · Voice reassignment: …` heading from `docs/testing/onbox-acceptance-register.md` (present on `main` at line 3464; absent at head). Register row IDs are allocated once and never reused, so every "register row A36" citation left in the tree is now permanently dangling — exactly what `scripts/check-register-citations.mjs` exists to catch. Measured:

```
check-register-citations.mjs      main EXIT=0   head EXIT=1   (Check A — nonexistent row ID (8))
check-register-row-citations.mjs  main EXIT=0   head EXIT=1   (3 citations)
npm run test:hooks                main EXIT=0   head EXIT=1   (11 failing tests)
```

The eight Check-A failures are all A36 and all introduced by this PR; the failing hooks tests include `check-register-citations.test.mjs`'s `CLI: the real tree scan passes clean`. (Ten of the eleven are that checker directly. The eleventh, `entry-point-guard-convention.test.mjs`'s allowlist test, passes standalone at head and is an in-run artifact of the failing batch, not an independent defect — it left no stray file behind, `git status --porcelain` was empty afterwards.)

This is not a local-only signal. `test:hooks` runs in `verify.yml` (`.github/workflows/verify.yml:209-211`, gated on `step_test_hooks`), and `docs/testing/**` is one of that step's declared inputs (`scripts/verify-cache.mjs:174`) — this PR changes four files under `docs/testing/`, so the leg is in scope and will run. `verify.yml` is a required status check on `main`, so as it stands the merge is blocked mechanically, not just by convention.

The repo's own convention is *annotate, don't renumber*: keeping the ID and putting a discharge/removal annotation near the citation downgrades it to a printed, non-fatal NOTE (`check-register-citations.mjs:44-54`). Thirty-eight such annotated citations already exist in the tree, and the correction log this PR added to the register (`onbox-acceptance-register.md:533-554`) is the right text to point at — it just needs to reach the eight citation sites. `docs/release-notes-next.md` and `RELEASE_NOTES.md` need no change: both are in the checker's `FROZEN_EXACT` set (`check-register-citations.mjs:436-440`), which is why their own A36 citations are not among the failures.

Neither prior pass caught this: pass 1 ran `check:onbox-register` (the sibling structural checker, which is green and blind to citations) and neither pass ran `check:register-citations` or `test:hooks`.

### F. 🟡 Three doc blocks describe a `bandMethod` mapping the code does not implement — one carried over from pass 2, two created by `f7b70b3c` *(chore — the defect is in the document)*
`server/src/audio/render-integrity/centroids-io.ts:39-43`, `:45-49`, `:62-66`; `server/src/audio/render-integrity/aggregate.ts:194-195`, `:297-300`

- `centroids-io.ts:39-43` and `:45-49` both say the dispersion-fallback case is stored as *"a percentile value from the dispersion fallback (when `bandMethod` is **'synthetic-sigma'** but pool dispersion triggered the percentile fallback)"*. After `f7b70b3c` that combination cannot occur — the fallback is tagged `'percentile'`. These two blocks describe the pre-`f7b70b3c` behaviour that pass 2's finding B was filed against; the code moved and the doc did not.
- `centroids-io.ts:62-66` says `'percentile' = pre-fix, 'synthetic-sigma' = post-fix`. Both halves are false at head: `'percentile'` is written by post-fix code on two paths (`aggregate.ts:309` fallback, `:313` non-synthetic audition), and pre-fix rows carry no value at all.
- `aggregate.ts:194-195` is the one pass 2 named and nothing touched: *"a pre-fix band method (missing `bandMethod` field **or 'percentile' value**, pre-sigma-band rows) … must not be trusted as-is"*. The code at `:206` rejects only `undefined`. This one is no longer merely inaccurate — it now describes the oscillation bug `4e53c3b9` just fixed, in the file that fixed it. I implemented the JSDoc literally as a mutation (`if (row.bandMethod === undefined || row.bandMethod === 'percentile') return false;`) and it reddens `aggregate-audition-voice-reassign.test.ts`'s reuse test: a reader "correcting the code to match the comment" would put every dispersed synthetic-only pool back on a rebuild-every-pass loop.
- `aggregate.ts:297-300`: `f7b70b3c` deleted the clause the last line continued, leaving an orphaned lowercase fragment after a full stop — `…the real-anchor path uses (safe, proven).` / `// the real (real-anchor) in-book path above never reaches this branch.`

### G. 🟡 The `usedSigmaBand` → `bandMethod` wiring — the whole point of `f7b70b3c`'s return-shape change — has zero coverage *(chore — coverage gap)*
`server/src/audio/render-integrity/aggregate.ts:309`

`bandMethod = spread.usedSigmaBand ? 'synthetic-sigma' : 'percentile'` is the line that makes finding B's fix real. Hardcoding it to **either** literal passes the entire 58-test battery:

```
bandMethod = 'synthetic-sigma'  → 58/58 green
bandMethod = 'percentile'       → 58/58 green
```

`syntheticOnlySpread`'s own `usedSigmaBand` flag is pinned in both directions at `score.test.ts:83`/`:100`, and `matchesCurrentVoice`'s `undefined` check is pinned by `d8b39717`'s test — but nothing asserts what value lands on a *persisted row* for a dispersed pool. Nothing reads `bandMethod` for its stated purpose today (only `:206`, for presence), so this is not a live wrong output; it is a behaviour shipped this round with no test, which is what the testing-discipline rule asks for. One assertion on `readCentroids(dir)!.thurid.bandMethod === 'percentile'` after a dispersed-pool `scoreBook()` closes it — the new test at `aggregate-audition-synthetic-band.test.ts:313` already builds exactly that fixture and stops one line short.

### H. 🟡 The dispersion threshold is an unnamed inline literal, unpinned, and still a cliff *(chore)*
`server/src/audio/render-integrity/score.ts:122`

`const isDispersed = std > 0.05;` — the third calibration constant of this band, sitting outside the exported `SYNTHETIC_ONLY_CUTOFFS` block that exists to hold them and that `score.test.ts:59-62` pins. Two of pass 2's three "unnamed inline magic numbers" became named `CUTOFFS.*` in `f7b70b3c`; this one did not. Mutating `0.05 → 0.049` passes 58/58, so the value is pinned only coarsely (0.5 and 0.005 are caught).

The cliff pass 2 measured is unchanged in kind and slightly larger in size — pool `[0.85, 0.879, 0.908, 0.937, 0.966, 0.995]` (std 0.04953) gives `pSevere` 0.77392, and `[0.85, 0.88, 0.91, 0.94, 0.97, 1.00]` (std 0.05123) gives 0.85900: an **0.08508** swing in the severe cutoff across a 0.0017 change in pool spread, so a render at cosine 0.80 is `voice-match` against the first pool and `voice-mismatch`/`severe` against the second. I am reporting this as 🟡 rather than carrying pass 2's 🔴, because both sides are now bounded and neither is unsafe relative to `main`: below the threshold `std ≤ 0.05` makes `pSevere ≥ mean − 0.15` (the bound pass 1 asked for), and above it the band is exactly `main`'s. It is a discontinuity in a calibration, not a hole in the detector.

---

## Verdict

**Do not merge.** One 🔴 remains, and it is new.

- **Both of pass 2's 🔴 findings are genuinely closed**, verified independently rather than from the commit messages. The dispersion fallback now equals `main` bit-for-bit on every dispersed pool (`pSevere` 0.59404 on the reference pool, delta 0.00000 across three), cosine 0.50/0.55/0.58 all score `voice-mismatch`/`severe` again, and `bandMethod` survives five consecutive reuse passes with `auditionCentroid` called exactly once. The `usedSigmaBand` tagging is correctly wired and cannot oscillate, because `matchesCurrentVoice` accepts both legal values and a matching row short-circuits before the pool is ever re-drawn.
- **New E is blocking on its own terms**: discharging register row A36 without annotating its eight surviving citations turns `npm run test:hooks` red — green on `main`, 11 failures at head — and that leg runs in `verify.yml` for this diff because `docs/testing/**` is one of its inputs. This is a required status check on `main`. It has been latent since `4951d558` and both earlier passes missed it.
- Findings F, G and H are 🟡 and fixed this round per the triage rule, but do not themselves reopen the loop. F is the one worth reading closely: `aggregate.ts:194-195` now *documents* the rebuild-loop bug that `4e53c3b9` fixed, and implementing it literally reddens an existing test.

**This is re-review round 2 of 2 — the final round under the loop cap.** A 🔴 is still open, so the automatic fix-and-re-review loop stops here and this goes to the user rather than triggering a pass 4. The remaining work is mechanical (eight discharge annotations, three doc blocks, one assertion, one named constant); a pass 4 would need to be requested explicitly.

dudarenok-maker:
## PR review — pass 4 (head 09569cc2, depth medium)

Re-review, outside the automatic loop cap (initial + 2) by explicit user approval after pass 3.

**Depth re-derived from the current commit set** (18 commits, all single-scope): `feat(server)` ×1, `fix(server)` ×4, `fix(docs)` ×1, `docs(server)` ×2, `docs(docs)` ×5, `chore(docs)` ×2, `test(server)` ×3. No `refactor`, no `perf`, no multi-scope commit → highest single-commit tier is `feat`/`fix` → **medium**. Unchanged from passes 1–3.

Scope: re-verified in a **throwaway detached worktree at 09569cc2** (not the shipping worktree), so every mutation probe below ran against an isolated copy. Verified independently rather than read: `npm run test:hooks` (1957/1957, exit 0), `check-register-citations.mjs` / `check-register-row-citations.mjs` / `check-onbox-register.mjs` (all exit 0 standalone), `npm run typecheck` (exit 0), `npx vitest run src/audio/render-integrity/` (126/126), `npx eslint` on all six changed server files (exit 0), and `node scripts/build-register-live-view.mjs` (regenerates to a **zero-byte diff** — the committed live view is genuinely current).

---

### Pass 3 findings — disposition

**Finding E (🔴 "branch 55 commits behind and CONFLICTING; GitHub had run ZERO CI checks; 8 dangling A36 citations would fail the required `test:hooks` leg") — VERIFIED RESOLVED.**
Not taken on trust:
- `git merge-base --is-ancestor origin/main 09569cc2` → **succeeds**. The merge-base is `da7c6f1e`, which *is* current `origin/main` (incl. PR #3041, merged after the rebase brief was written). Nothing is behind.
- `gh pr view` → `MERGEABLE / CLEAN` (was `CONFLICTING` / `BLOCKED`).
- CI has genuinely run for the first time: all 19 checks pass, including the two that were still pending mid-review (`Server tests (fast + slow)` 1m41s, `Lint, typecheck & fast checks` 2m24s).
- `npm run test:hooks` run here: `pass 1957 / fail 0`, exit 0 — the 1957/1957 claim is real.
- All three A36 citations in `onbox-a36-results/` (`step-1-implement.md:3`, `step-3-onbox.md:3` and `:6`) are now reported by the checker as `annotated as discharged/removed`. The `ID_PROXIMITY_CHARS` fix in `6679bffe` does bind — I confirmed the annotation now sits immediately after "Register row A36" rather than behind the issue-URL parenthetical.

**Register merge arithmetic — VERIFIED, independently recomputed.**
Body-row headings counted directly out of `onbox-acceptance-register.md`: **A=38, B=2, C=4, D=3, E=10, G=2, H=2 → 61**. That matches the "At a glance" table row-for-row *and* the `**61 owed.**` line, and the correction-log entry's `62 → 61 — Group A 39 → 38` is consistent with main's independently-landed E12/E9/E10/E101 discharges. `check:onbox-register` exits 0 and `register:build` is idempotent. The hand-resolved conflict is correct.

**Finding F (🟡 "stale `bandMethod` doc comments") — RESOLVED in substance; one residual inaccuracy carried forward as NEW-3 below.**
The two claims worth checking both hold: `matchesCurrentVoice`'s doc no longer says a `'percentile'` value is pre-fix, and the code agrees — the gate is `if (row.bandMethod === undefined) return false;` (`aggregate.ts:205`), so a `'percentile'`-tagged row *is* treated as valid/non-stale, exactly as the corrected doc now claims. The orphaned fragment (`// the real (real-anchor) in-book path above never reaches this branch.`) is gone.

**Finding G (🟡 "zero test coverage on the `usedSigmaBand`→`bandMethod` wiring") — VERIFIED RESOLVED by my own mutation, both directions.**
I did not trust "verified to catch a mutation":

| Mutation applied to `aggregate.ts:307` | Result |
|---|---|
| `bandMethod = 'percentile'` (always) | 🔴 RED — `expected 'percentile' to be 'synthetic-sigma'` (tight-pool test) |
| `bandMethod = 'synthetic-sigma'` (always) | 🔴 RED — `expected 'synthetic-sigma' to be 'percentile'` (dispersed-pool test) |

Both assertions are load-bearing, and neither is satisfiable by the other branch. Genuine coverage.

**Finding H (🟡 "unpinned magic-number dispersion threshold") — VERIFIED RESOLVED.**
`SYNTHETIC_ONLY_CUTOFFS.dispersionStdThreshold: 0.05` is a named constant and `syntheticOnlySpread` reads it (`score.ts:126`). Mutating `0.05 → 0.049` reddens `score.test.ts:62` (`expected 0.049 to be 0.05`). Noting without raising it as a finding: the pin is **value-only** — no test places a pool whose std straddles the boundary, so nothing covers the threshold's *behaviour*. That matches the file's existing convention for `severeSigma`/`bandSigma`, so it is consistent rather than a regression.

**Passes 1–2 🔴s — re-confirmed still correct at this head.** The dispersion fallback still uses the real-anchor path's own `CUTOFFS.severeEdgePctl`/`bandUpperPctl` (6/10), and `persistedAsRef` still carries `bandMethod` through (`aggregate.ts:188`), with `centroidsMap[charId] = { characterId, ...outcome.ref }` spreading it back to disk. The rebase did not undo either fix.

---

### 🟡 CARRIED FORWARD — the pass-2 sub-item is still open, never addressed

`server/src/audio/render-integrity/aggregate-audition-synthetic-band.test.ts:195`

The test titled **"catches severeSigma narrowing mutations"** does not catch them. Its own comment (lines 196–199) states the worked example:

> *"Reducing severeSigma (e.g., to 1) would raise pSevere to ~0.927, making 0.85 NOT severe."*

That arithmetic is wrong. With the pinned pool (mean 0.9629, population σ ≈ 0.017318), `severeSigma = 1` gives `pSevere = 0.9629 − 0.017318 = **0.94558**`, not ~0.927 — and 0.85 is still far below it, so the segment stays `severe`.

**Ran the exact mutation the comment names.** `severeSigma: 3 → 1`, then `vitest -t "cosine 0.85"`:

```
Test Files  1 passed (1)
     Tests  1 passed | 11 skipped (12)
EXIT=0
```

`3 → 2` also leaves it green. The test can only fail at `severeSigma ≳ 6.5` (where `mean − kσ` finally drops below 0.85) — i.e. only on a *widening* mutation more than double the calibrated value, the opposite of what its title claims.

**This is a mislabel, not a coverage hole** — I confirmed the class *is* covered elsewhere: `severeSigma: 3 → 2` reddens the two `0.928` tests (`pSevere` becomes 0.928264, crossing the fixture). So severity stays 🟡.

**The asymmetry is real, not a harness artifact** — the sibling test at `:214` (`bandSigma`, cosine 0.945) genuinely fails under the mutation *its* comment names: `bandSigma: 1.5 → 0.5` → `expected 'inconclusive' to be 'voice-match'`. One twin works; this one does not.

**Fix:** the sensitive shape is asserting a cosine sitting just *above* the severe edge stays non-severe (e.g. 0.915, which flips to `severe` once `severeSigma` narrows past ≈2.75). Assert that, and correct the false comment. Cheapest correct outcome is to fix the arithmetic in the comment and move the fixture cosine; deleting the test would lose nothing but would also fix nothing.

---

### 🟡 NEW-1 — two shipped docs cite a commit SHA that exists in no ref, created by this round's rebase

`docs/testing/onbox-acceptance-register.md:543` and `docs/testing/onbox-a36-results/step-3-onbox.md:8`

Both cite the step-1 implementing commit as **`c7a1dfac`**. That is the *pre-rebase* SHA. The rebase rewrote it, and the object is now orphaned:

```
git cat-file -t c7a1dfac            → commit          (still in this box's local object store)
git merge-base --is-ancestor c7a1dfac 09569cc2 → NOT REACHABLE from PR head
git for-each-ref --contains c7a1dfac → (empty)        ← zero refs, local or remote
```

Zero refs contain it, so it will never be pushed to GitHub. On any fresh clone — and on this box once gc runs — `c7a1dfac` resolves to nothing.

Concrete failure: a reader auditing the discharged A36 row follows the register's own correction log to *"Step 1 (`fix/server-2934-a36-audition-band`, commit c7a1dfac …)"* and gets `fatal: bad object`. This matters more than a normal dead link because the register's correction log is the durable audit record for a **discharged on-box acceptance row** — the evidence trail is the whole reason the row was allowed to leave the register.

The correct post-rebase SHA is **`7d12493f`** (`feat(server): A36 wider severity band for small synthetic-only audition pools`). Two one-word edits.

*Secondary, same file:* `step-3-onbox.md`'s H1 reads "**A36 step 2** — real on-box confirmation", the register calls it "**Step 2**", and there is no `step-2-*.md` in the directory — the filename is the odd one out. Worth one rename or one line of explanation, at the shipping session's discretion.

---

### 🟡 NEW-2 — the new `RELEASE_NOTES.md` bullet turns the whole section's list loose

`RELEASE_NOTES.md:81-83`

The entry was inserted with a blank line on each side, into a list that is otherwise tight:

```
80:- **Cloning a voice no longer risks it briefly showing "ready" …
81:<BLANK>
82:- **The voice-mismatch check in Chapter QA no longer flags …
83:<BLANK>
84:- **Editing your voice sample's transcript now gives clear feedback …
```

Lines 70–80 and 84–88 carry no blank separators. Under CommonMark/GFM, *one* blank line between items makes the **entire list loose**, so every bullet in this user-facing section renders with `<p>` wrapping and extra vertical spacing — a visible formatting change to the whole block, not just the new entry.

`docs/release-notes-next.md` is loose throughout, so the entry there is correct as written and needs no change; only `RELEASE_NOTES.md` is inconsistent with its own file. Fix: delete lines 81 and 83.

---

### 🟡 NEW-3 — residual of finding F: the `pSevere`/`pBand` doc bullets name a combination that cannot occur, and omit the one that does

`server/src/audio/render-integrity/centroids-io.ts:40-41` and `:48-49`

Finding F corrected the `bandMethod` *field* doc (`:66-74`) but left the two adjacent field docs saying:

> *"When `bandMethod` is 'percentile': a percentile-of-pool value (used on **real-anchor in-book paths**, or on synthetic-only audition pools that fall back due to high dispersion)."*

Four lines later the same block states *"Only 'audition' rows carry `bandMethod`."* Both cannot be true. In-book rows are built at `aggregate.ts:256` and carry **no** `bandMethod` at all, so "`bandMethod` is 'percentile' … on real-anchor in-book paths" describes an impossible row.

The case it actually omits is the common one: an **audition** row whose pool had real anchors blended in (`audition.syntheticOnly === false`, reached whenever `existingAnchors.length > 0` at `audition-centroid.ts:250-251`, i.e. a character with *some* but not enough in-book anchors). That is the main producer of `'percentile'` tags and the doc does not mention it.

Failure this causes: someone debugging a stale-row question greps `centroids.json` for `bandMethod` on in-book rows, finds none, and concludes the field is broken. Fix: replace "real-anchor in-book paths" with "audition pools that blended in real anchors" in both bullets.

---

### Rebase-gap check (requested)

I diffed the 55-commit gap for collisions with this PR's intent. Nothing in it touches `server/src/audio/render-integrity/**` — the gap's register-adjacent commits (`7cb14f7f`, `0f04bb67`, `8413da63`, `6da04605`, `581440ae`, `b71e1efd`, `5634b784`) are all A29 / E12 / E9-E10-E101 doc discharges on disjoint rows, which is exactly what the correction log claims. No stale or duplicated work, and the one artifact the rebase *did* stale is NEW-1 above.

### Verdict

**No 🔴 and no 🟠. No correctness bug in shipped runtime behaviour.** Pass 3's blocking finding E is genuinely closed — rebased onto current `main`, `MERGEABLE / CLEAN`, all 19 CI checks green, `test:hooks` 1957/1957 confirmed here. Findings G and H are resolved and I proved it by mutation rather than by reading. The register merge arithmetic is correct and self-consistent at 61.

Four 🟡 cleanups remain — one carried forward (the mislabeled `severeSigma` test, open since pass 2 and now demonstrated with the mutation its own comment names) and three new, of which NEW-1 is the one I would not merge without: a discharged on-box acceptance row whose audit trail points at a commit that exists in no ref. All four are documentation/test-label edits with no behaviour risk.

Per the triage rules, 🟡-only findings are **fixed this round but do not re-trigger another review pass** — so this can merge once they are folded in.

PR #3021, opened 2026-09-06
chore(scripts,docs): record the attempt count behind verify-cache's durationMs

### LINKED ISSUE #3018: ops-69 — verify-cache durationMs conflates up to 3 retry attempts
## What

`.verify-cache.json`'s `durationMs` silently conflates **1 to 3 attempts** of a step, so a
recorded figure cannot be read as that step's cost.

In `scripts/verify-cache.mjs`:

```js
const t0 = Date.now();
const code = runStepProcess(...);   // <-- contains the MAX_POOL_ATTEMPTS retry loop
const dt = Date.now() - t0;
```

`runStepProcess` retries internally on a vitest fork-pool crash (`isVitestPoolCrash`), up to
`MAX_POOL_ATTEMPTS = 3`, and `dt` wraps the whole loop. The cache entry records only `durationMs`,
with no attempt count — so a 19-minute entry is indistinguishable from one crashed attempt plus an
18-minute pass, or two crashes plus a short pass.

Only `RETRIABLE_POOL_STEPS` (`test:server`, `test:server-slow`) can be inflated this way; every
other step is always exactly one attempt.

## Why it matters

This is the artifact the ops-2997 commit-gate design doc drew its headline cost figure from:
**`test:server` = 1,167,036 ms (19.45 min)**. That number is load-bearing — it is the evidence that
the pre-push battery was non-viable on arithmetic alone. It may well be correct, but the artifact
cannot demonstrate it, and the same file is the natural input for the run budgets in Parts 2-6 of
that design.

`durationMs` has **exactly one writer and no programmatic consumer** — it is telemetry that humans
and agents read by hand. That is precisely why it needs to be self-describing: nothing downstream
will catch a misreading.

This closes one of the two measurements the ops-2997 design deferred Parts 2-6 behind. The
measurement was recorded as "run it and see"; the better answer is to make the artifact answer it
for free on the next honest run, rather than spending ~19 minutes of a contended box to sample it
once.

## Acceptance

1. A cache entry records the attempt count alongside `durationMs`.
2. A step that passes on its first try is distinguishable from one that passed after retries.
3. The `[pass]` console line says so when a step took more than one attempt.
4. Paired regression tests: a step passing first try records 1; a step that pool-crashes then
   passes records the true count and the summed duration.
5. Reading a pre-existing cache entry with no attempt field does not crash or reset the cache.

## Key files

- `scripts/verify-cache.mjs` — `runStepProcess` (~1316), the cache write (~1530), `SCHEMA_VERSION`
- `scripts/tests/verify-cache.test.mjs`

## Benefit

*(technical)* Makes the one artifact this repo uses to reason about verify cost honest about what
it measured, and closes a deferred measurement without paying for a 19-minute sampling run.

dudarenok-maker:
## Summary

`.verify-cache.json`'s `durationMs` silently conflated **1 to 3 attempts** of a step, so a recorded figure could not be read as that step's cost.

`runStepProcess` retries internally on a vitest fork-pool crash (`isVitestPoolCrash`, up to `MAX_POOL_ATTEMPTS = 3`), and the timing wraps that whole loop:

```js
const t0 = Date.now();
const code = runStepProcess(...);   // <-- the retry loop lives in here
const dt = Date.now() - t0;
```

So a 19-minute entry was indistinguishable from one crashed attempt plus an 18-minute pass. Only `RETRIABLE_POOL_STEPS` (`test:server`, `test:server-slow`) can be inflated this way; every other step is always exactly one attempt.

**Why this is a prerequisite, not a tidy-up.** `durationMs` has one writer and no programmatic consumer *today* — but Part 2 of the commit-gate design derives its self-calibrating budget as `max(FLOOR, K × lastGreenDurationMs)`, reading it directly. It calibrates off whichever run happened to be last-green, and an entry inflated by two crashed attempts is not a representative baseline. With the count recorded, a consumer can prefer `attempts === 1` entries and treat the rest as uncalibratable.

The budget **shape** is unchanged — per-step total, exactly as the design doc settles it. An earlier revision of this PR claimed Part 2 must consume a per-attempt figure; that reversed a decision settled thirty lines above it in the same document, and is corrected here.

### What changed

- `runStepProcess` returns `{ code, attempts }`; the non-retriable early return reports 1.
- The cache entry gains an `attempts` field alongside `durationMs`.
- The `[pass]` **and `[fail]`** lines name the count only when it exceeds 1. The `[fail]` path is the one case where three attempts is *certain*, and it was originally left reporting an undecomposed duration.
- Observability only: no retry behaviour changed, no `SCHEMA_VERSION` bump, and a pre-existing entry with no `attempts` field is still a valid cache hit.

### This closes a deferred measurement without paying for it

The design doc's plan was "run one `test:server --no-cache` on a reaped box and record per-attempt timings." That was **deliberately not taken** — it buys a single sample for ~19 minutes of exclusive box time, and an earlier attempt to sample it turned into a real 19-minute battery on a box that was mid-cleanup. Recording the count instead means every future run says how many attempts its figure covers.

To be precise about what that does and does not settle: the **mechanism** question is answered (the artifact could not tell you; now it can). The **reading** — what the existing 19.45 min entry actually decomposes into — is not, because pre-existing entries carry no count. That half is tracked in #3025.

The design doc also records that Part 1's propagation sweep is complete: **19/19 worktrees on instant hooks, 0 on the old battery.**

## Test plan

- `node --test scripts/tests/verify-cache.test.mjs` → green.
- `npx vitest run src/spawn-windows-hide.test.ts` (server) → 22 pass. Run **by name** because this change sits beside the `spawnSync` options object that guard scans syntactically; it was not touched or hoisted.
- `node --test scripts/tests/hook-no-pool.test.mjs` → 10 pass.
- `npx eslint scripts/verify-cache.mjs scripts/tests/verify-cache.test.mjs` → clean.
- CI: 19/19 green.

### How the evidence got here, since that is the point of this PR

The first cut of the tests was **wrong in the exact way this PR is about**, and both review passes were needed to get it right.

- The acceptance criteria were originally covered by **source-text pins**, which false-passed on realistic drift: commenting the field out, or renaming it to `retryAttempts:`, removed it from the persisted cache entry while the regex stayed green. Only outright deletion — the one mutation originally tested — reddened them.
- The stated reason for pinning rather than executing, *"executing that region would spawn real vitest batteries"*, was **an assertion never tested, and it was false.** The pipeline loop is step-agnostic: it spawns whatever `npm run <step>` resolves to in `cwd`. Review pass 1 disproved it by running the real thing in **1.1 s** against a throwaway git repo.
- Both pins are now deleted, replaced by one behavioural test that drives the real `runPipeline` and asserts the actual persisted `.verify-cache.json` and the actual `[pass]` console line, in both the retry and first-try cases.
- An `assert.ok(dt >= 110)` bound **could not fail** — a single attempt already costs 582 ms — and was structurally unfixable, since `dt` was the test's own wall clock. Replaced with a marker-file oracle counting real process launches.

**Mutation coverage.** The coordinator ran 7 mutations; review pass 2 declined to accept that list, rebuilt HEAD into a throwaway tree, and ran **18** — the 7 plus 11 it devised, including every probe of the first-try (`attempts: 1`) case. All 18 redden. No surviving mutation of the seam was found.

Two review passes; pass 2 returned no correctness findings.

### Before-shipping checklist

- **Step 7 (`verify:fast:branch`) deliberately not run locally.** `test:server`'s globs include `scripts/**/*.{mjs,cjs,js}`, so this diff drags the ~19.45-min step into scope — a ~20-minute local battery on a box running three agent lanes. `scripts/ci-scope.mjs` confirms CI runs **exactly the three legs** that battery would (`lint`, `test:hooks`, `test:server`), and CI is the required, authoritative gate. Running the local duplicate loses no coverage and reintroduces the contention ops-2997 removed.
- Steps 1, 3, 4, 8 — not applicable: no new plan (issue body + paired tests are the spec), no hardware-provable behaviour, no plan moved.
- Step 5 — `docs/release-notes-next.md` updated. **`RELEASE_NOTES.md` intentionally skipped**: operator-facing tooling, no end-user delta.

- `mkTmp()` never removed its temp directories — **39,680** `verify-cache-test-*` dirs had accumulated in `%TEMP%` since 2026-08-04, growing +25 per run, and this PR's tests write git repos and spawn npm inside each. Now cleaned up per-run, best-effort, and only ever the directories that run created.
- Design-doc citations of `verify-cache.test.mjs` line numbers had drifted onto unrelated code; replaced with symbol/section references, per the precedent that document already sets for itself.

Closes #3018
Refs #3025


dudarenok-maker:
## PR review — pass 1 (head 1db50f26, depth high)

The production change (`runStepProcess` → `{ code, attempts }`, the cache field, the conditional `[pass]` note) is small and correct. Every finding below is in the **evidence layer** — the two tests that pin the acceptance criteria, and the two documents that record what was learned. For a PR whose entire deliverable is *making a recorded number trustworthy*, that is where it counts.

### 🔴 Blocking

None.

### 🟠 Correctness

**1. Both acceptance-criterion pins false-pass on realistic drift, and the stated reason for pinning instead of testing is factually wrong.** `scripts/tests/verify-cache.test.mjs:1863`

The regex is `/cache\.steps\[step\.name\]\s*=\s*\{[^}]*attempts,?[^}]*\}/`. `[^}]*` spans comments and neighbouring tokens, so two realistic edits remove the field from the persisted entry while the pin stays green. Verified by applying each mutation to a copy of `scripts/verify-cache.mjs` and re-running the exact assertion:

| Mutation applied at `scripts/verify-cache.mjs:1533` | Cache entry after | Pin still matches? |
|---|---|---|
| `attempts,` → `// attempts recorded elsewhere now` | **no `attempts` key** | yes — **false pass** |
| `attempts,` → `retryAttempts: attempts,` | **no `attempts` key** (renamed) | yes — **false pass** |
| `attempts,` deleted outright | no `attempts` key | no — correctly reddens |

Only the third is the mutation the PR body verified. This is the repo's own documented "prose about a marker matches the marker" shape.

The justification — *"executing that region would spawn real vitest batteries"* — does not hold. The loop body is **step-agnostic**: it spawns whatever `npm run <step.name>` resolves to in `cwd`. Driving the real `runPipeline` against a throwaway git repo whose `package.json` maps `test:server` to a one-line node fixture exercises **both** criteria in **1.1 seconds**, with no vitest anywhere:

```js
runPipeline({ argv: ['--steps', 'test:server'], cwd: fixtureDir,
              env: { ...scrubGitEnvForThrowawayRepo(process.env), SKIP_CONTENTION_CHECK: '1' } });
```

Actual output, run against this branch:

```
[pass] test:server (took 1.0s, 2 attempts)          <- criterion 2, real console output
cache: { "schemaVersion": 1, "steps": { "test:server": {
  "inputHash": "d758b08c…", "lastGreenAt": "…", "durationMs": 1045, "attempts": 2 } } }
                                                    <- criterion 1, real persisted artifact
real process launches (marker bytes): 2             <- independent oracle
--- exit: 0  elapsed: 1096 ms
```

`hasVitestStep` is the only thing that would fire a contention probe, and `SKIP_CONTENTION_CHECK=1` disables it; a `--steps lint` variant needs not even that (583 ms, also verified). This seam reddens on **all three** mutations above, costs about a second, and is strictly cheaper than the pins were to write. The pins should be replaced, not merely hardened.

**2. `assert.ok(dt >= 110)` cannot fail, and the guarantee it advertises is structurally unobtainable.** `scripts/tests/verify-cache.test.mjs:1381`

Measured cost of a **single-attempt** `runStepProcess` call on this box: **582 ms** standalone, **2756 ms** under the real `npm run test:hooks` invocation. The bar is 110 ms — 5× to 25× below what *one* attempt already costs. The assertion passes on one attempt's worth of elapsed time, so it does not discriminate the case it exists to score.

Worse, it cannot be made to fail by any change to the production code. `dt` is the *test's own* wall clock around a synchronous call, so it necessarily spans every attempt regardless of what `runStepProcess` does; nothing in `runStepProcess` computes, returns, or reports a duration. The 60 ms busy-wait is entirely swamped by npm's ~500–2700 ms startup and contributes nothing measurable.

So the comment at `:1379-1380` (*"proving the elapsed time isn't just the last attempt"*) and the PR body's *"with a busy-wait proving the duration actually sums across retries rather than reflecting only the last attempt"* both claim something the test does not deliver.

A failable oracle already exists and is thrown away: `writeFlakyFixture` **returns `marker`** (`:1332`) and all three call sites discard the return value. `assert.equal(readFileSync(marker, 'utf8').length, 2)` counts real process launches and reddens the moment a retry stops happening — which is the fact the test is actually about.

**3. The design-doc paragraph reverses Part 2's own budget-shape decision, and asks for a figure this PR does not record.** `docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md:247` vs `:215`

`:215` states as settled: *"**Budget shape: per-step TOTAL, not per-attempt.**"* — with three sentences on why the per-attempt shape was wrong and how the labels got swapped last time. The new `:247` says: *"Part 2 must consume the per-attempt figure, not the step total."* That is the same swap recurring, in the same document that records it happening before.

Under `:215`'s chosen shape the new paragraph's stated harm is also backwards. If the budget must cover a step *including* its retries, then a 3-attempt total is the **conservative, correct** calibration input — not "up to 3× too generous". The real risk runs the other way, and this PR cannot close it, because it records a **count**, not a per-attempt figure. An implementer who takes `:247` literally will compute `durationMs / attempts`:

> `test:server` crashes at 60 s, crashes again at 60 s, then passes in 1100 s → entry `{ durationMs: 1220000, attempts: 3 }` → "per attempt" = 407 s, against a true single-pass cost of 1100 s → budget ≈ **2.7× too tight**, false-positive on every clean run — precisely the false-positive `:223` cites as the reason 20 min was rejected.

What the change actually buys is: *a consumer can now reject any entry with `attempts > 1` as uncalibratable*. That is genuinely enough for Part 2 — but it is not what the doc says.

Same shape, secondary:

- `:240` — *"every future run reports its own **decomposition** for free"*. A count is not a decomposition: `{durationMs: 1220000, attempts: 3}` still cannot tell you the split, and the attempts are not equal length (a crashed attempt aborts early).
- `docs/release-notes-next.md:499` headline — *"`durationMs` no longer silently conflates 1-3 retry attempts into one number"*. It still conflates them; what changed is that the entry now says **how many** it conflated. The body of the same entry is accurate, so only the headline clause overstates.

### 🟡 Lower severity

**4. [correctness — class left armed] The `[fail]` path still prints an un-decomposed duration.** `scripts/verify-cache.mjs:1538`

`attempts` is in scope at that line and is dropped. Observed through the real pipeline with a fixture that crashes all three attempts:

```
[fail] test:server (exit 1, took 2.4s)      <- three real process launches; no attempt count
exit: 1 | cache written: false
```

This is the one case where three attempts are *certain*, and the summary line reads as a single 2.4 s run — the exact misreading #3018 exists to stop, left armed on the failure path. No cache entry is written either, so nothing else records it. One interpolation fixes it.

Stated honestly: the `[retry]` lines are on the console above, so an operator with the full scrollback can reconstruct the count. It is absent from the summary line and from every artifact.

**5. [cleanup — stale comment this change made false] `scripts/tests/verify-cache.test.mjs:2`**

The file header still reads *"No `npm run` spawning — runPipeline itself is exercised by the manual walkthrough in docs/features/archive/50-verify-cache.md."* Three of the new tests spawn real `npm run` — measured 2756 ms / 5081 ms / 3671 ms under `npm run test:hooks`, roughly +11 s to that suite. CLAUDE.md lists "a stale comment or doc sentence your change made false" as a chore owed in the same round, not a later one.

### ✅ Checked and clean

Each of these was actually exercised, not assumed:

- **`SCHEMA_VERSION` not bumped is correct.** `durationMs` and `attempts` have exactly one writer and **zero programmatic readers** repo-wide. `loadCache` (`scripts/verify-cache.mjs:731-745`) validates only `schemaVersion` and the shape of `steps`; `decide` (`:674-679`) reads only `inputHash`. Branch-switching is inert in both directions: old code reads a new entry fine and silently drops `attempts` on its next green write. A bump would have discarded every green input hash for no gain.
- **The pre-existing `runStepProcess` body pin still anchors correctly** after the `export` change. `src.match(/function runStepProcess\([\s\S]*?\n\}/)` captures 1372 chars, starts at `function runStepProcess(npmScript, …` and ends on `return { code: lastRes.code, attempts: MAX_POOL_ATTEMPTS };` + the closing brace — the whole body, no early termination. `function runStepProcess(` occurs exactly once in the file, so no wrong-span risk.
- **`server/src/spawn-windows-hide.test.ts` is undisturbed.** The `spawnSync('npm', …)` options object is untouched and still carries a literal `windowsHide: true` inside its own parens; nothing was hoisted or spread; no new spawn call was added anywhere under `scripts/**` (that scan covers `scripts/tests/` too, per its own header). CI's "Server tests (fast + slow)" leg is green.
- **The three spawning tests cannot reach the real repo and leave nothing behind.** `cwd` is a fresh `mkdtempSync`; inheriting the outer npm's `npm_config_*` env does *not* redirect script resolution — verified by running through the real entry point, `npm run test:hooks` → **1953 pass / 0 fail**. `spawnSync` is synchronous, so no orphaned processes. Windows backslash paths are correctly embedded via `JSON.stringify(marker)`.
- **Not timing-flaky.** All three assertions are lower bounds with 5×–25× headroom; the defect is the opposite of flakiness (finding 2).
- **`attempts` recorded only on a passing run is the right call** for a cache whose entries are by definition last-green records. The gap is on the console, not in the cache (finding 4).
- **Test 3 does prove "never retries"** despite asserting only the return value: the fixture fails when the marker is absent, so a second launch would find it and exit 0 — `assert.equal(result.code, 1)` is load-bearing. Test 2's `code === 0` likewise requires a genuine second launch. Both are real behavioural tests.
- **`node --test scripts/tests/verify-cache.test.mjs` → 139 pass / 0 fail**; worktree `git status --porcelain` empty before and after this pass.
- **House gates:** `Closes #3018` present and outside any code span; `docs/release-notes-next.md` updated with `RELEASE_NOTES.md` explicitly declared not-applicable (operator tooling, no end-user delta); no on-box acceptance owed; no config knob, no derived artifact, no `cast.json` write; multi-scope `chore(scripts,docs)` title matches convention.
- **"19/19 worktrees on instant hooks, 0 on the old battery"** — independently verified by reading every worktree's `.husky/pre-commit` rather than by `git branch --contains`, as the doc instructs. 22 trees enumerated; excluding the primary checkout, the temp-scratchpad tree and `Audiobook-Generator-wt-3017-check` leaves exactly 19; **all 22 are on `pre-commit-lint.mjs`, none on the old battery.** Claim holds.

### Verdict

**Changes requested.** Three correctness findings (1, 2, 3), one incomplete fix (4), one chore the change made owed (5). Nothing here is a defect in shipped runtime behaviour — the production diff is right. What is wrong is that the two tests standing behind the acceptance criteria can both go green on a change that removes the field, the third assertion cannot fail at all, and the two documents record a stronger result than the artifact delivers. On a PR whose entire premise is *"a recorded number was not trustworthy"*, that is the part that has to hold.

Findings 1 and 2 are cheap to close together: one behavioural test through `runPipeline` (1.1 s, shown above) replaces both pins, and one `readFileSync(marker).length` assertion replaces the vacuous timing bound.

dudarenok-maker:
## PR review — pass 2 (head 01c620cd, depth high)

Deltas only. All five pass-1 findings are **verified resolved** — including the two whose fixes I re-derived independently rather than accepting the coordinator's battery. The production diff is unchanged from pass 1 apart from two lines and remains correct. What is left is a residue of the fix itself: the machinery that the deleted assertion existed for, and three documents that still describe the pins that are gone.

**No 🔴/🟠 correctness findings. This pass does not re-trigger the loop.**

---

### Disposition of pass 1's five findings

**1. Both acceptance pins false-passed → ✅ VERIFIED RESOLVED.**

I did not take the coordinator's seven-mutation list on trust. I rebuilt HEAD into a throwaway tree (`git archive 01c620cd`, junctioned `node_modules`, `git init` — outside the repo) and ran **18** mutations, each applied to `scripts/verify-cache.mjs` and scored against the whole 139-test file. Baseline in that tree: **139 pass / 0 fail**.

All seven claimed mutations redden. **Eleven more that I devised also redden**, including every probe of the first-try (`attempts: 1`) case the dispatch asked for:

| # | Mutation | Reddens |
|---|---|---|
| M1 | delete `attempts,` from the cache write | `#3018: runPipeline persists attempts…` |
| M2 | comment it out (`// attempts recorded elsewhere now`) | same |
| M3 | rename to `retryAttempts: attempts,` | same |
| M4 | `attemptsNote` unconditional | same |
| M5 | force `attempts: 1` in the retry-loop success return | same + `runStepProcess: a fork-pool crash…` |
| M6 | drop `attemptsNote` from the `[fail]` line | `#3018: [fail] line reports…` |
| M7 | force `attempts: 1` in the exhausted-loop return | `#3018: [fail] line reports…` |
| **X1** | `attempts > 1` → `>= 1` (over-fire on first try) | `#3018: runPipeline persists…` |
| **X2** | `attempts > 1` → `> 2` (under-fire at 2) | same |
| **X3** | cache field pinned to constant `2` | same — **the first-try half is load-bearing** |
| **X4** | non-retriable early return reports `0` | same + `…non-retriable step always records 1…` |
| **X5** | non-retriable early return reports `2` | same + same |
| **X6** | `MAX_POOL_ATTEMPTS` 3 → 2 | `#3018: [fail] line reports…` — **the marker oracle can fail** |
| **X7** | `runPipeline` passes a non-retriable `retryKey` | both #3018 tests |
| **X8** | un-hoist `attemptsNote` back inside `if (code === 0)` | `#3018: [fail] line reports…` |
| **X9** | drop the `saveCache` call | `#3018: runPipeline persists…` |
| **X10** | write `attempts: String(attempts)` | same |
| **X11** | drop `attemptsNote` from the `[pass]` line | same |

I could not construct a mutation of the `attempts` seam that survives. X3 and X4/X5 are the answer to "could it pass for the wrong reason" — a constant-`2` cache field and a `0`/`2` non-retriable return both die on the `--steps test` half, so that half is not decorative.

**Coverage lost by deleting the pins: none.** The pins asserted (a) the field is in the object literal — now proven by the persisted artifact (M1/M2/M3/X3/X10), and (b) the literal source form of the ternary — now proven semantically (M4/X1/X2) and by wording (`/2 attempts/` fails on any reformat of the note). The replacement is strictly stronger and no longer breaks on an unrelated rename.

**2. `assert.ok(dt >= 110)` cannot fail → ✅ VERIFIED RESOLVED.** Replaced by `assert.equal(readFileSync(marker,'utf8').length, 2)` at `scripts/tests/verify-cache.test.mjs:1380`. X6 proves the new oracle is falsifiable: cutting `MAX_POOL_ATTEMPTS` to 2 reddens the launch count, which the timing bound never would have.

**3. Design doc reversed its own budget shape → ✅ VERIFIED RESOLVED.** Checked, not assumed:
- `:249` now states the shape as per-step total and says nothing revises it — consistent with `:215`.
- Both in-text line references resolve correctly today: `:215` is *"Budget shape: per-step TOTAL, not per-attempt"*; `:223` is the table row citing constant false-positives as why 20 min was rejected.
- The worked example's arithmetic is right: 60 + 60 + 1100 = 1220 s = `durationMs: 1220000`; 1220 / 3 = 406.7 ≈ 407 s; 1100 / 407 = **2.70×**.
- `:240` "decomposition" → "records how many attempts its figure covers". `docs/release-notes-next.md:499` headline no longer claims the conflation stopped.

**4. `[fail]` dropped the attempt count → ✅ VERIFIED RESOLVED.** `scripts/verify-cache.mjs:1538` now interpolates `attemptsNote`; M6/M7 redden. **The hoist is safe**: `attemptsNote` occurs at exactly three lines (`:1519`, `:1521`, `:1538`), the initialiser is a pure expression over an already-destructured `const`, there is no other `attemptsNote` binding in the file, and no shadowing. X8 (putting it back inside the `if`) reddens with a `ReferenceError`, which is the correct signal.

**5. Stale test-file header → ✅ RESOLVED at the header — but see 🟡 A: the same class is left armed 500 lines below.**

---

### Freshly assessed, clean

- **The new tests cannot reach the real repo.** `cwd` is `mkdtempSync`; `cachePath = join(cwd, '.verify-cache.json')`. I ran both #3018 tests with `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` **and** a lowercase `git_dir` all pointed at a real repo: **2 pass / 0 fail, that repo's HEAD unchanged and `git status --porcelain` empty before and after.** Worth naming precisely, since the test's own scrub is not what saves it: `runPipeline` passes the caller's `env` only to the `npm` spawns; its internal `gitFileList(cwd)` uses the module's own `gitEnv()` (`scripts/verify-cache.mjs:985`). That path scrubs the four redirection vars and `GIT_PREFIX`, and it only ever reads, so the outcome is right either way — but the safety comes from `gitEnv()`, not from the test's `scrubGitEnvForThrowawayRepo`.
- **No leaked processes.** `spawnSync` throughout; nothing detached.
- **Console capture is restored on failure paths.** Both swaps sit in `try { … } finally { console.log = originalLog; }` and both test callbacks are synchronous, so `node --test` cannot interleave another test into the swapped window.
- **Cost is acceptable.** Measured on this box: the two new tests are **2646 ms + 2136 ms = 4.78 s**; the whole file is **13.4 s / 139 tests**. `test:hooks` is in no git hook since ops-2997 — it runs in CI's *Lint, typecheck & fast checks* leg, green here at 2m13s, so this is ≈ +4%. Fine.
- **Not flaky.** Every timing assertion is gone; the surviving oracles (launch counts, cache contents, log text) are deterministic.
- `npx eslint scripts/verify-cache.mjs scripts/tests/verify-cache.test.mjs` → exit 0. All 19 CI checks green. `Closes #3018` present. Worktree `git status --porcelain` empty before and after this pass.
- "7 new tests" is still accurate: `main` 132 → head 139 (two deleted, two added).

---

### 🟡 New findings (all cleanup / record — none re-trigger the loop)

**A. [cleanup — orphaned code + stale comment, the same class finding 5 fixed] The busy-wait outlived the only assertion it existed for.** `scripts/tests/verify-cache.test.mjs:1344-1348`

The commit deleted `const t0 = Date.now()` / `const dt` / `assert.ok(dt >= 110)`. It left the fixture text that fed them:

```js
// A short busy-wait so a real, measurable amount of time elapses on EVERY
// attempt — proves a summed duration actually accumulates across retries
// rather than only reflecting the last attempt.
const until = Date.now() + 60;
while (Date.now() < until) { /* spin */ }
```

Concretely: `writeFlakyFixture` is used by three tests, and no assertion anywhere now reads elapsed time — so this spins for 60 ms on each of 4 process launches (~240 ms of pure waste per suite run) while its comment claims to *prove* the exact guarantee pass 1 established the code never delivered. This is CLAUDE.md's "a stale comment your change made false" plus "remove what YOUR changes orphaned". The two new fixtures written in the same commit (`:1865`, `:1961`) correctly carry no busy-wait, which is the tell.

**B. [cleanup — stale record] The PR body still documents the deleted pins, by name, with the justification the fix commit retracts.**

The body's mutation table cites two tests that no longer exist anywhere in the file:

- *"reddens `runPipeline writes the attempts field to the cache entry on pass — #3018 acceptance criterion 1`"*
- *"reddens `runPipeline appends the attempt count to [pass] output conditionally, not unconditionally — #3018 acceptance criterion 2`"*

`grep "^test('" scripts/tests/verify-cache.test.mjs` at 01c620cd returns neither. The paragraph below it is worse: *"those two are **source-text pins**, not behavioural tests — executing that region would spawn real vitest batteries"*, and *"with a busy-wait proving the duration actually sums across retries"*. Commit `01c620cd`'s own message says that first claim "was simply wrong", and the busy-wait clause is finding 2 verbatim. The PR body is the record a reviewer reads at merge time; it currently advertises the defects rather than the fix. Also worth adding what the fix did, per CLAUDE.md's "Also fixed, found in passing" convention.

**C. [cleanup — duplication the work exposed] Three hand-rolled copies of a fixture writer that already exists.** `scripts/tests/verify-cache.test.mjs:1863-1902` and `:1959-1978`

`writeFlakyFixture(dir, npmScriptName, { failFirst })` at `:1336` writes exactly this shape — a one-line node fixture plus a package.json with one script. The new tests re-implement it three times inline (`flaky.mjs`, `pass.mjs`, `allfail.mjs`) plus two more copies of the `git init` / `config` / `add` / `commit` preamble that `makeGitFixture()` at `:1556` already provides. The two things the helper lacks are a two-script package.json and an always-fail variant — one parameter each.

**D. [cleanup — doc] The design doc's "source-text pins" warning cites line numbers that point at unrelated code.** `docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md:254-259`

That bullet list is what Part 2's implementer will read before the `spawn` conversion. Its references were already drifted on `main` (`verify-cache.test.mjs:1288` there is inside the `detectSiblingContention` assertion, not the `runStepProcess` body extractor; `:1332-1341` is a comment block; `:1391-1399` is a `diffSafeForChangedOnly` test) and this PR moved the file by ~230 lines while editing the same section, so they are further off now. The substance still holds — two consumers of `runPipelineLoopBody()` remain, at `:1455` and `:1514` — only the addresses are wrong. A second copy of the same span sits at `:377` (*"`verify-cache.test.mjs:1288-1311`"*), so fixing one site leaves the class armed. Pre-existing origin, but adjacency is not the test.

**E. [cleanup — bookkeeping] The status table closes a question the doc's own Part 2 says is still open.** `…commit-gate-rebalance-design.md:464` vs `:249`

Row 2 — *"Is `test:server`'s 19.45 min one attempt or three?"* — is marked **Answered 2026-09-06**. `:249` says the opposite in as many words: *"that number remains a step total of unknown composition until the next full local run rewrites it."* The interim `CASTWRIGHT_STEP_TIMEOUT_MIN = 45` at `:223` is still calibrated on that unqualified figure, and with the question closed nothing now tracks re-reading it once a run with `attempts` lands. Either keep the row open, or add the re-check as a tracked item. (Flagging this as borderline: the row is defensible as "resolved by instrumentation" — the concrete leak is that no follow-up survives it.)

**F. [cleanup — pre-existing class, +2/run from this PR] The file never removes its temp directories.** `scripts/tests/verify-cache.test.mjs:75`

`mkTmp()` is `mkdtempSync` with no `after()` anywhere in the file. Measured on this box right now: **39,680 `verify-cache-test-*` directories in `%TEMP%`**, growing by **25 per suite run** — I counted 39,680 → 39,705 across one run. The two new tests add two of the heaviest kind (each a full `git init` repo, not a bare fixture dir). Pre-existing in origin and not a reason to hold the PR, but this is the cheapest possible fix (one `after()` hook) and the number is well past the point of being theoretical.

---

### Verdict

**Approve on correctness.** The production change is right, and the two acceptance criteria are now proven by an oracle I could not break in 18 attempts — the pins are gone and nothing they covered went with them. Findings A–F are all cleanup/record: the fix left its own scaffolding behind (A, C), and three documents still describe the state before it (B, D, E). Per findings-triage they are fixed in this round but do not reopen the loop.

Issue #3030, opened 2026-09-06
start-app-prod.mjs launcher false-positives on a sibling worktree's own server, refuses to start


dudarenok-maker:
`npm run start:prod` (`scripts/start-app-prod.mjs`) refuses to start a worktree's server if it detects *any* Castwright instance already listening on the alternate LAN-HTTPS port (`:8443`), even when that instance belongs to a completely different worktree. The launcher's "already running, don't duplicate" heuristic probes the port globally, with no cwd/workspace check, so it misidentifies a sibling worktree's own server as "this one, just on the wrong port" and refuses to launch — logging `[SKIP] a Castwright server is already listening on :8443 — leaving it alone`.

## Where this came from

On-box mechanical batch 1 campaign (#2949, #2435), step 4 (docs/testing/onbox-mechanical-batch1-results/step-4-a5-a13-a17-a19.md, opening section): hit this live when `wt-onbox-mechanical-batch1`'s own launch collided with `wt-fs38-a1-wave12`, a sibling lane's own dev server already on `:8443`. Worked around by invoking `node dist/index.js` directly (bypassing the launcher's port-guard) rather than through the launcher script, which pushed this worktree's own server to `:8444` instead.

## Suggested fix direction

The launcher's liveness probe should confirm the process on the alternate port actually belongs to *this* worktree (e.g. by checking a workspace-identifying response field, or by binding the probe to this worktree's own configured port rather than a shared default) before refusing to start.

Filed per CLAUDE.md's incidental-findings rule, at the direction of #2958 (docs-only register fold) — reported, not fixed, since it needs its own code change outside this fold's docs-only scope.

Issue #3029, opened 2026-09-06
A freshly confirm-fallback'd queue entry isn't always picked up promptly by the client-side dispatcher


dudarenok-maker:
A freshly `fallbackConfirmed: true` queue entry does not always get picked up promptly by the client-side queue dispatcher. Confirming a parked fallback via the browser's "Render anyway" button can leave the entry sitting at `awaiting_confirm` (with `fallbackConfirmed: true` already set) for minutes, with the sidecar's own `/health` showing zero `inflight_synth` — i.e. the confirmation landed but nothing re-dispatched it.

## Where this came from

On-box mechanical batch 1 campaign (#2949, #2435), step 4 (docs/testing/onbox-mechanical-batch1-results/step-4-a5-a13-a17-a19.md, row A5 bullet 4, "One operational note"): on the first attempt to confirm a Coqui-fallback render through the actual browser UI, the queue entry never re-dispatched after confirmation. Forcing the confirm via a direct `POST /api/queue/{id}/confirm-fallback` call, then driving `POST /api/books/{id}/generation` directly with `fallbackConfirmed: true` in the body, worked immediately and rendered correctly end to end — proving the server-side confirm→render path itself is correct. This points at `queue-dispatcher-middleware.ts` on the client not always noticing a freshly-confirmed `queued` entry, rather than anything server-side.

## Suggested fix direction

Have the dispatcher middleware re-check for dispatchable entries immediately after a `confirm-fallback` call resolves (rather than only on its normal poll/event cadence), so a freshly-confirmed entry doesn't have to wait for the next incidental trigger.

Filed per CLAUDE.md's incidental-findings rule, at the direction of #2958 (docs-only register fold) — reported, not fixed, since it needs its own code change outside this fold's docs-only scope, and this session could not fully root-cause the dispatcher's trigger conditions.

dudarenok-maker:
Fixed by PR #3144 (merged 2026-09-10, `cf7f2829`), which referenced this issue with `Refs` rather than `Closes`, so it never auto-closed.

[...]

Issue #3028, opened 2026-09-06
Queue modal's fallback-confirm button says "Render anyway in Kokoro" even when the fallback engine is Coqui


dudarenok-maker:
The Queue modal's fallback-confirmation button (`src/modals/queue-modal.tsx`) hardcodes the copy "Render anyway in Kokoro" / `aria-label="Render anyway in Kokoro"` regardless of which fallback engine the book will actually use. For a Coqui-eligible non-English book (fs-60, plan 249), the render underneath is correct — it genuinely falls back to Coqui, not Kokoro — but the button's own label still says Kokoro.

## Where this came from

On-box mechanical batch 1 campaign (#2949, #2435), step 4 (docs/testing/onbox-mechanical-batch1-results/step-4-a5-a13-a17-a19.md, row A5 bullet 4 aside): confirmed a Russian book's fallback via this exact control, and the render that followed genuinely used Coqui XTTS (`segments.json`'s `renderedFallbackEngine: "coqui"` on every line) — the button's label was simply wrong for what was about to happen.

## Suggested fix direction

`queue-modal.tsx` should check `coquiEligible` (or read the engine the request will actually use) the same way the Cast-view banner and the voice-readiness gate already do, and render "Render anyway — generic Coqui fallback voices" for a Coqui-eligible book instead of the hardcoded Kokoro copy.

Filed per CLAUDE.md's incidental-findings rule, at the direction of #2958 (docs-only register fold) — reported, not fixed, since it needs its own code change outside this fold's docs-only scope.

dudarenok-maker:
Already fixed — `cf7d387f` (merged via PR #3031, same day as this issue but after it was filed) made `queue-modal.tsx`'s confirm button/aria-label/title dynamic via `selectFallbackEngineName(state, bookId)` (`src/store/voice-readiness-selectors.ts:115`), which is now used at `src/modals/queue-modal.tsx:564-565`. Verified: `src/modals/queue-modal.test.tsx:343-344` and `:373-374`/`:402-403` assert both the Coqui and Kokoro label variants. Closing as already resolved; no chain needed.

Issue #3027, opened 2026-09-06
Design full cast can halt silently mid-run, and an orphaned concurrent single-design request never resolves


dudarenok-maker:
"Design full cast" (bulk Qwen voice design) can halt silently partway through a run, with no error toast and no designed/failed/skipped terminal summary — the top bar just flips from `"active": true` to `{"active": false}` and reads "Halted · N/M". A concurrent single-character "Design & preview" request against a different book, started while the bulk job owns the GPU, can also get stuck spinning forever with no toast and no resolution, even after the bulk job itself has ended.

## Where this came from

On-box mechanical batch 1 campaign (#2949, #2435):

- Step 2 (docs/testing/onbox-mechanical-batch1-results/step-2-a6-a7.md, rows A6/A7): "Design full cast" over a 12-character cast halted after exactly 2 characters, twice in a row — once with an unrelated forced sidecar restart and a concurrent cross-book single-design request in flight, and again as a clean, isolated control run with zero interference. Both times the job flipped straight to `{"active": false}` (`castDesignActions.halt`) with nothing logged to `server.log`, `server.err.log`, `tts.log`, or `tts.err.log` at the moment of the halt. The concurrent single-design request from the second bullet never resolved either — its button stayed in a spinning "active" state for the rest of the session, no toast, no error, even once the bulk job had ended and the sidecar's `inflight_synth` had returned to 0.
- Step 3 (docs/testing/onbox-mechanical-batch1-results/step-3-a8-a9-a10.md, "Finding 1") found a strong, plausible alternate explanation for the halt on the *same* book/cast: this worktree's `server/.env` has no `GEMINI_API_KEY`, and "Design full cast" (10 remaining characters) failed immediately with a clean, structured SSE error — `GEMINI_API_KEY is required to generate voice-style personas.` Switching the live config knob `analyzer.personaGeneration.engine` from `gemini` to `local` designed all 10 remaining characters successfully in one run, no halt. This was **not confirmed byte-for-byte** against step 2's original environment, but persona generation is exactly the step between "character N done" and "character N+1 starts" that a missing/exhausted Gemini key would silently break.

[...]

PR #3031, opened 2026-09-06
fix(server,docs): fs-60 Coqui fallback gate + discharge on-box rows A4-A19

### LINKED ISSUE #2949: ops — on-box mechanical batch 1 (part 1/2): A4/A5/A6/A7/A8/A9/A10/A13/A14/A17/A19
Part of the on-box acceptance register campaign ([#2435](https://github.com/dudarenok-maker/Castwright/issues/2435)), covering the **first half** of the "mechanical single-8GB-card checks" batch identified in the 2026-09-05 re-audit (`docs/testing/onbox-a-series-audit-2026-09-05.md`, batch 1 of the "Proposed chain batches" section — 22 rows total, split into two parallel chains of 11 each; the other half (A20, A21, A26, A27, A32, A33, A35, A102, A104, A105, A24) is a sibling chain filed separately).

## Why this batch

All 11 rows below are described by the audit as fully agent-runnable now: no
ear/eyeball checks, no acquisition blockers, pure API/log/measurement-based
acceptance. Each row's own register text (`docs/testing/onbox-acceptance-register.md`)
is the source of truth — re-read it, not just this issue, before executing any
step.

## The 11 rows, grouped into 4 task steps by shared setup

1. **A4 + A14** — multi-engine audition/tier fidelity, then real-book QA/badge
   agreement (both are "render a book, check the numbers/behavior" rows that
   share a render session).
2. **A6 + A7** — bulk Qwen voice design walkthroughs ("Design full cast"),
   forced-recycle resilience and the acceptance bullet left open at merge.
3. **A8 + A9 + A10** — QA re-record RTF, per-character re-record/splice gain
   check, and the structured-failure-taxonomy live multi-failure check.
4. **A5 + A13 + A17 + A19** — the single-8GB-card, `CUDA_VISIBLE_DEVICES=0`
   pinned session the register's own text explicitly says to run together
   (fs-60 Coqui-fallback walkthrough, idle-Coqui reclaim bullets 2-3,
   contended-eviction `/health` liveness, stranded-VRAM-pool reclaim).

Then a fold step updates the register, and a verify step checks everything
and opens the PR.

## Standing rules for every step in this chain

- **Never mutate real book data.** Work in this worktree's own
  `WORKSPACE_DIR`, never the primary checkout's.
- **Never stop, kill, or restart another lane's process.** This box has other
  live GPU/analyzer lanes running right now — check GPU/port availability
  before claiming either card, and if a step's render would contend with
  something already resident, say so in the evidence file and proceed only
  once clear, never by killing what's there.
- **Never touch `server/.env` outside this worktree's own copy**, and never
  copy the primary checkout's `server/.env` into this worktree.
- **Never `--no-verify`.**
- This box's GPU 0 is an RTX 4070 Laptop (8 GB), GPU 1 is an RTX 5070 Ti
  (16 GB) — both live. Rows marked "single 8 GB card" in the register MUST
  pin `CUDA_VISIBLE_DEVICES=0`, since an unpinned run on this dual-GPU box
  will pick whichever card has more headroom and invalidate the row's own
  measurement.
- **Mutate any code-level claim before trusting it** — if a step's evidence
  asserts a guard fired or a value changed, show the actual before/after
  output, not a description of what should have happened.

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-onbox-mechanical-batch1` (already created,
  dependencies installed, husky active; `server/.env` + `.env.local` written
  for slot 14: VITE 5313 · API 8220 · TTS 9140; `server/tts-sidecar/.venv`
  and `server/tts-sidecar/voices/` junctioned from the primary checkout).
- Branch: `docs/docs-onbox-mechanical-batch1`.
- All steps commit on this one branch. Only the final verify step opens a PR.

## Not in scope

The sibling chain's 11 rows (A20, A21, A26, A27, A32, A33, A35, A102, A104,
A105, A24). Any other register row or group. A12's reboot-gated bullet 2 (not
part of this batch).

Refs #2435

dudarenok-maker:
Closes #2949
Refs #2435

## Summary

- Steps 1-5 of the on-box mechanical batch 1 chain ran real, on-box acceptance checks against 11 register rows (A4, A5, A6, A7, A8, A9, A10, A13 bullets 2-3, A17, A19), each backed by observed command output, log lines, and measured numbers in `docs/testing/onbox-mechanical-batch1-results/step-{1..4}-*.md`.
- A5 fully discharged and removed from the register (real product fix included in this branch: `server/src/routes/generation.ts`'s per-chapter fallback gate hard-failed every non-English book regardless of Coqui eligibility — a pre-fs-60 branch never updated when fs-60 added the Coqui substitution. Narrowed the hard-fail to the genuinely-unsupported case; paired test updates in `generation-fallback-gate.test.ts`).
- A4/A6/A7/A8/A9 narrowed to their remaining owed bullets; A10/A13/A17/A19 kept with new evidence folded on top of prior wave history, none overwritten. `docs/testing/onbox-acceptance-register.md`'s "At a glance" Group A count and owed total recomputed by counting headings. **Post-rebase the figures are Group A 37 and owed 60** — the pre-rebase body said 39→38 / 65→64, which was this branch's arithmetic before it took in 101 commits of main. The rebase's only real conflict was that total (main said 61, this branch 64) and BOTH were stale: the merged body carries main's discharges AND this branch's A5, so it was recomputed from the rows via `register:build` rather than resolved by picking a side. Verified independently by hand-count in review. All three register gates are green: `check:onbox-register`, `check-register-citations`, `check-register-row-citations`.
- 5 incidental-finding issues filed from this evidence (not fixed here, out of scope): #3026, #3027, #3028, #3029, #3030.

## Step 6 verification (this issue, #2953)

Re-checked all 11 rows against their evidence files: every claim in the register is backed by real observed output, not an unverified assertion, and the register fold accurately reflects what was and was not discharged (owed bullets are honestly marked NOT REACHED/FAIL where the evidence shows a real gap, e.g. A6/A7's bulk-design-halt regression (#3027), A8's CPU-ASR-QA RTF miss, A9's A/B-audition FAIL). One minor judgment call: A7's "VRAM headroom holds across a long run" bullet is annotated in the register as "PASS as far as it went" with an explicit caveat that only 2 of 12 characters' worth of data was collected (short of the full run plan-108's OOM needed) — transparent, not fabricated, but arguably should stay listed as owed rather than resolved. Not blocking; noted for whoever next touches A7.

This PR is **not docs-only** — it carries the `server/src/routes/generation.ts` + `generation-fallback-gate.test.ts` fix bundled from A5's acceptance run — so it is not exempt from `pr-review-gate` per CONTRIBUTING.md's doc-only fast-path. Requesting a `pr-review-gate` pass at **medium** depth (single-scope fix) before merge.

## Test plan

- [x] `npm run check:onbox-register` passes (verified green during step 6 verification)
- [x] `generation-fallback-gate.test.ts` updated to assert park-then-confirmed-render for Coqui-eligible non-English books (was: hard-fail)
- [ ] `pr-review-gate` pass at medium depth (requested, not yet run)

---

## Review pass 1 (depth `high`) — 1 🔴, 2 🟠, 7 🟡

This PR was **inherited mid-flight**: it was opened by another lane whose worktree went idle for ~4h, and it ran **zero CI for its whole life** until today, because a `CONFLICTING` PR runs no `pull_request` workflows. It showed one green CLA check while three required checks were in fact failing. It was rebased from 101 commits behind, and everything below was found only once the gates could finally run.

| Sev | Finding | Disposition |
|---|---|---|
| 🔴 | queue row hardcoded "Kokoro"; this PR's fix makes that path reachable, and it renders on **Coqui** | fixed `cf7d387f` |
| 🟠 | `!coquiEligible` arm unreachable + its coverage deleted | filed **#3059**, decision named |
| 🟠 | fix widens exposure to the known queue wedge, unrecorded | recorded on **#3026** |
| 🟡 1 | A5 annotation inserted mid-sentence | fixed `30e347e9` |
| 🟡 2 | run sheet said "STILL OWED" under "discharged", blank `Result:` | fixed `30e347e9` |
| 🟡 3 | "the annotation commit is a no-op" | **rebutted — see below** |
| 🟡 4 | two A5 refs inside the register un-annotated | fixed `30e347e9` |
| 🟡 5 | change log skipped this PR's own 61→60 move | fixed `30e347e9` |
| 🟡 6 | PR body figures stale post-rebase | fixed (above) |
| 🟡 7 | plan 249 still said acceptance owed | fixed `e3ecc0b3` |
| — | release notes: two false claims | fixed `30e347e9` |

**On 🟡 3.** The finding analysed `check-register-row-citations.mjs`; the gate that was failing, and that broke six hooks tests, is its sibling `check-register-citations.mjs`. Reverting the annotations and re-running settles it:

| Script | Without annotations | With |
|---|---|---|
| `check-register-citations.mjs` | **exit 1** | **exit 0** |
| `check-register-row-citations.mjs` | exit 0 | exit 0 |

The commit does what it says. The two scripts have different matchers *and* different proximity rules — the sibling requires the annotation tied to the same logical line, this one uses a 25-line window — and that distinction bit again immediately: flipping plan 249 introduced two new dangling A5 citations that only the sibling caught.

**Also fixed, found in passing:** this PR carried a user-facing product fix with no release-notes entry on either surface (Before-shipping step 5). Added, then corrected twice when review showed two of my claims were false.


dudarenok-maker:
**pr-review-gate pass (medium depth, single-scope fix) — requested per issue #2953**

Reviewed via git diff, ran the full server route test suite on both `main` and this branch (63 tests pass on branch), and traced the touched code path (`generation.ts`'s per-chapter loud-fallback gate) through its callers.

Overall: clean, well-executed docs + code PR. The `coquiEligible` narrowing in `generation.ts` is logically sound and consistent with the existing pattern used elsewhere in the same file.

One finding, filed as #3032 rather than fixed here (this issue, #2953, is read-only for source):

- `server/src/routes/generation-fallback-gate.test.ts`: the diff deletes all four pre-existing `#1263` regression tests for the still-unsupported-language fail-fast branch (message pluralization, narrator-clause dedup, `state.json` persistence) and replaces them with a single new test that only covers the new Coqui-eligible park-and-confirm path. The still-unsupported-language branch (`nonEnglishBook && !coquiEligible`) remains live and reachable but after this PR has no direct test coverage — violates CLAUDE.md's "never delete/skip a test without an explicit replacement" rule.

Recommend folding #3032 before merge.

dudarenok-maker:
## PR review — pass 1 (head 2d4ba768, depth high)

Scope: `server/src/routes/generation.ts` + its paired test, the register + live view, the four `onbox-mechanical-batch1-results/step-*.md` evidence docs, the three files carrying A5-citation annotations, and the two release-notes surfaces added in `2d4ba768`. Verified before probing: `generation-fallback-gate.test.ts` 8/8 green at HEAD; `npm run check:onbox-register` green; `npm run register:build` idempotent against the committed live view (re-ran it, `git status` stayed empty); register body-row counts recounted by hand; `check-register-row-citations.mjs` green. The paired test was mutation-tested in an isolated scratch copy of the worktree, so the repo tree was never touched.

Depth note: the PR body requests **medium** ("single-scope fix"). The title is `fix(server,docs)` — multi-scope → **high** per the depth ladder. This pass ran high.

---

### 🔴 Blocking — this PR makes a wrong engine name reachable in the UI, and defers the fix on a reason that is both void and now factually obsolete

**Correctness bug.** `src/modals/queue-modal.tsx:518` and `:556-558`.

Repro: Russian book, one undesigned Qwen speaker, "Generate this chapter" from the queue.

- **Before** this PR: `generation.ts:1452` hard-failed. A non-English book never reached `awaiting_confirm`, so the queue modal never rendered this row for one.
- **After**: it parks. The queue modal renders
  `Needs confirmation · no designed Qwen voice for Sofiya → would render in Kokoro` (`:518`)
  above a button carrying `aria-label="Render anyway in Kokoro"` / `title="Render anyway (Kokoro fallback)"` (`:556-558`).
- What actually happens on confirm: **Coqui**, via `synthesise-chapter.ts:1508-1516`. The step-4 evidence doc confirms it on real hardware — `segments.json` carried `renderedFallbackEngine: "coqui"` on every line.

This is not cosmetic. fs-2's premise is that Kokoro *cannot* read the text, and this same file says so 47 lines above the park: `"English Kokoro voices cannot read ${bookLanguage} text."` (`generation.ts:1465`). A user shown "would render in Kokoro" on a Russian book has been told the output will be unusable, and will rationally press **Skip** — which is the exact outcome this fix exists to prevent. The PR ships the gate that makes the message reachable without shipping the message.

**No design pass is owed.** `selectFallbackEngineName(state, bookId)` already exists for precisely this (`src/store/voice-readiness-selectors.ts:115`). Its own comment states the invariant — *"so the UI never promises a fallback the server won't use"* — and records that this identical divergence was already found and fixed once, in #1534, for the cast banner, the gate button and the gate copy. The queue modal is the fourth surface and was never wired in. `bookId` is already in scope at the `BookGroup` call site (`queue-modal.tsx:223`).

**The deferral reason has expired.** #3028's body defers it as *"outside this fold's docs-only scope"*. That premise died the moment this branch grew `server/src/routes/generation.ts`; nobody re-evaluated it. The PR body's *"not fixed here, out of scope"* is separately a reason CLAUDE.md lists as void verbatim: *"it would expand the scope of this PR — a finding the work surfaced is in scope by definition."*

Note also that #3028 names only the **button**. The status line at `:518` is a second site with the same wrong word. Fix the class, not the instance.

---

### 🟠 Significant — three tests deleted with no replacement; the branch they covered is now both unguarded and unreachable

**Correctness bug (coverage).** `server/src/routes/generation-fallback-gate.test.ts`; branch at `server/src/routes/generation.ts:1452-1510`.

Four tests were removed and one added. Only one of the four was testing the behaviour that changed. The other three covered logic that **still exists**:

- the `state.json` persist (`generationState` / `generationError` / `generationErrorCode` / `generationRemediation`, `:1478-1500`);
- the plural / comma-join wording (`:1453-1454`);
- the `narratorAlreadyListed` suppression (`:1459-1461`).

Demonstrated, not inferred: I copied the worktree to scratch and changed `:1452` to `if (false && nonEnglishBook && !coquiEligible)` — neutralising all 58 lines including the persist — and the file stayed **8/8 green**. Repo-wide, the only surviving `voice-not-designed` tests are `src/store/generation-stream-runner.test.ts:284` (hand-crafts the tick client-side) and a taxonomy-enum assertion. Neither reaches the route.

**And the branch cannot be reached in production either.** `resolveEligibleEngines` is called with the static `ALL_TTS_ENGINES`, not installed engines (`generation.ts:830`), so `coqui` is always a candidate. `ENGINE_LANGUAGE_SUPPORT.coqui = ['en','ru','es','fr','de','zh','ja']` (`voice-mapping.ts:47`) is **exactly** the language registry's code set (`language-registry.ts:45-119`), and any language outside that registry throws in `sidecarLanguageName` at `generation.ts:851`, long before line 1452. So `nonEnglishBook && !coquiEligible` is a provably empty set today.

There are two defensible outcomes, so this one genuinely needs a decision named in an issue: keep the branch as a fail-safe for the next registry language **and** give it a test that stubs eligibility, or delete it as orphaned by this change. What is not defensible is where it stands now — retained, unreachable, and silently stripped of its coverage. CLAUDE.md: *"Never delete or `.skip` a test without an explicit replacement or follow-up plan item."*

This is also the real weight behind bullet 5's "N/A / superseded": it is not that the walkthrough is awkward to run, it is that no still-unsupported language exists to run it against.

### 🟠 Significant — the fix widens exposure to a known, unfixed queue wedge, and says so nowhere

**Correctness bug (interaction).** `generation.ts:1508` (`markQueueEntryAwaitingConfirmOnDisk`) against #3026.

The change converts a class of runs from *hard-fail* to *park on `awaiting_confirm`*. #3026 — filed from this same campaign, observed twice on-box — records that an unanswered `awaiting_confirm` entry blocks every entry queued behind it, silently, with no error surfaced. This PR increases how often those entries are created.

#3026's own deferral is legitimate: it names the decision owed (skip-vs-timeout). The finding is the **interaction**, which is recorded nowhere. At minimum the PR body and the release note should say the new behaviour is *parks for confirmation*, so an operator who meets a stalled queue after this lands has the pointer. Both currently read as "it now generates".

---

### 🟡 Minor

1. **Annotation inserted mid-sentence.** `docs/testing/onbox-sitting-vram-contention.md:4-6`. The original blockquote read "…register rows **A5, A13, A17, A19, A24, A25** — the rows that only mean something when the single 8 GB card is genuinely full". The annotation was dropped between "the rows that" and "only mean something", so it now renders as three blockquotes, one of them a dangling fragment.

2. **The run sheet contradicts itself three lines apart, and its `Result:` is still blank.** Same file, A5 section (~`:160-176`): the new "A5 was discharged" annotation sits directly above a criteria-source block that still ends **"STILL OWED."**, and step 6's `- Result:` line is empty. Before-shipping step 3 requires the run sheet carry the filled-in Result lines. The A16 block lower in the same file shows the shape this should have taken on discharge.

3. **`ed3edccb`'s stated purpose is not what it achieved.** `check-register-row-citations.mjs:79` matches only `\b(?:row|Row)\s+([AE]\d+)\b`. None of the four annotated sites is that form — `row **A5**`, `rows **A5, A13…`, and a bare `### A5 ·` heading — and the fourth file (`docs/superpowers/plans/…`) sits outside `SCAN_PREFIXES` (`:77` — `docs/testing/`, `docs/features/` only), so it is never scanned at all. `grep -rnoE "\b[Rr]ow +A5\b"` over both scanned prefixes returns nothing, and the checker's own note lists 11 annotated citations, none of them A5. The checker passes identically with or without the commit. The annotations are useful as documentation; the rationale for them is false.

4. **Two A5 references inside the register itself were left un-annotated** while four external ones were annotated: `onbox-acceptance-register.md:1900` ("**Run this with A5** — same card, same mixed-cast book…") and `:2043` ("runnable alongside A5/A13 in the same session"). Both instruct a future runner to co-schedule a row that no longer exists — the same class as the four that were annotated.

5. **The register's change log skips this PR's own change.** `onbox-acceptance-register.md:533` still reads "**Last change: 2026-09-06 (claude), 62 → 61 — Group A 39 → 38**", which is the earlier A36 discharge. The strip and glance table moved to 60 / A 37, but nothing records the 61 → 60, A 38 → 37 move. Every other change in the log has an entry, and the log is how a reader reconciles the figures.

6. **PR body figures are stale post-rebase.** It states "Group A count (39→38) and owed total (65→64)". Recounted by hand: **Group A 37, owed 60**.

7. **Plan 249 was not flipped, and cannot stay as it is.** Frontmatter `status: active` / `shipped: null`; body `:9` still reads "Live-GPU acceptance owed (mock-mode e2e only)… stays `active`, not `stable`, until that walkthrough runs"; `docs/features/INDEX.md:73` still says "`active` … Live-GPU acceptance owed". A5 *is* that walkthrough, and this PR discharges it in full. Either the plan flips to `stable` with Ship notes and moves to `archive/` (Before-shipping steps 4 and 8), or "A5 fully discharged" is overstated. It cannot be both.

---

### Release notes (`2d4ba768`) — reviewed on the same terms

Every load-bearing claim in the technical entry checks out against the code:

- ✅ the condition is `nonEnglishBook && !coquiEligible` (`generation.ts:1452`);
- ✅ **"the split this same file already uses twice for the same distinction (`generation.ts:938` and `:952`)"** — accurate at this HEAD. `:938` is `qwenUnavailable && nonEnglishBook && !coquiEligible`, `:952` is `qwenUnavailable && nonEnglishBook && coquiEligible`. Checked directly, not taken on trust;
- ✅ "`forbidKokoroFallback` is unconditional for them" — `generation.ts:1690` passes `forbidKokoroFallback: nonEnglishBook`;
- ✅ "`forbidKokoroFallback` + `coquiEligible` ⇒ Task 6's `applyQwenFallback` substitutes Coqui" — `synthesise-chapter.ts:1508-1516`.

Two claims I would change:

- **"so 'Proceed anyway' now delivers the Coqui fallback the voice-readiness gate promises"** — true of the voice-readiness gate's own button, but the button actually reached on the path this fix opens says *Kokoro* (🔴 above). As written, the note promises something the shipped UI contradicts.
- **"A still-unsupported non-English book (not Coqui-eligible either) keeps the immediate, all-characters-named failure, which remains the right answer for it."** — accurate about the code, but it describes user-visible behaviour no user can reach (🟠 above). A release note asserting live behaviour for an empty case is the wrong place to record a fail-safe.

Brand entry: carries no identifiers ✅, names no engines ✅ ("its second voice engine", "its other engine"). One overstatement:

- **"even though the screen offered you a 'Proceed anyway' button"** attributes the failure to the wrong control. That button sets `fallbackConfirmed: true` (`voice-readiness-gate.tsx:86` → `generation-stream-middleware.ts:130`), and the server gate requires `!job.fallbackConfirmed` (`generation.ts:1428`) — so taking it skipped the broken branch entirely and reached the Coqui fallback. **It worked before this fix.** The path that hard-failed is the queue-driven per-chapter one, which never opens that modal: `startGenerationFlow` is wired only to the "Approve cast & start generating" / "Resume generation" CTAs. The technical entry gets this right ("on the one path real users take"); the brand entry does not.

**On whether the gap should have been caught:** yes. Before-shipping step 5 was owed and unmet, and a PR carrying a user-facing product fix with neither release-notes surface mentioning it is a finding this pass would have raised. Recording it as owed rather than as never-missing.

---

### ✅ What is solid

- **The core fix is correct.** The Coqui substitution it relies on genuinely exists and is genuinely reached with `forbidKokoroFallback: true, coquiEligible: true`. The narrowing removes the failure rather than moving it.
- **The paired test is a real regression test.** Mutation-verified in an isolated copy: reverting `:1452` to `if (nonEnglishBook)` turns it red at `generation-fallback-gate.test.ts:412` for exactly the claimed reason — park absent, `chapter_failed` / `errorCode: voice-not-designed` present. Green after. Not a test that cannot fail.
- **Register arithmetic verified by hand, not via the checker.** Counting body headings: A 37, B 2, C 4, D 3, E 10, G 2, H 2 = **60**, with Blocked 6 and Unconfirmed 2 correctly excluded. Matches the glance table and the strip exactly. `register:build` re-run left `git status` empty, so the derived figures really are derived.
- **No row is discharged without evidence.** A5 is the only removal and it has observed on-box output behind it. Every other touched row stays owed with honestly-marked FAIL / NOT REACHED bullets — A6/A7's 2-of-12 halt, A8's CPU-ASR-QA miss, A9's A/B FAIL. Nothing is discharged on "tests pass, so presumably fine".
- **The A5 annotation's substance is accurate.** Bullets 1–4 PASS are each backed by the step-4 evidence doc, and bullet 5 really is superseded (zh/ja became Coqui-eligible in fs-59 W4b).

### Verdict

**Request changes.** One blocking and two significant findings, all correctness. The blocking one is small and the helper it needs already exists; the first significant one needs a decision named in an issue rather than a patch. Re-review triggered once the 🔴 is fixed and pushed.

dudarenok-maker:
## Pass 1 findings addressed — `2d4ba768` → `30e347e9`

Nine of ten fixed, one rebutted with evidence. Three commits: `cf7d387f` (🔴), `e3ecc0b3` (finding 7), `30e347e9` (findings 1/2/4/5 + release notes). Findings 3 and 6 answered below.

### 🔴 — fixed, `cf7d387f`

You were right that the branch growing `generation.ts` killed #3028's deferral reason. `selectFallbackEngineName` per `BookGroup`; the `bookId` was already at the call site and simply never passed down.

I verified beyond the fix agent's own probe: it mutated the status text, so I mutated the **aria-label** instead, and the assertion at `:343` reddens with `"Render anyway in Coqui"` vs `"...in Kokoro"`. Component restored byte-identical. 37 tests.

### 🟠 deleted coverage / unreachable branch — filed **#3059**

Filed rather than fixed, with the decision stated: keep-and-test, remove, or convert to an explicit invariant check. The thing being decided is whether `ENGINE_LANGUAGE_SUPPORT.coqui == registry code set` is an invariant we may rely on, a coincidence to defend against, or a condition to alarm. Your `if (false && …)` neutralisation is quoted as the evidence.

### 🟠 queue-wedge exposure — recorded on **#3026**

Your framing is exactly right and I took it as written: deferring the wedge fix is legitimate, the missing thing was the record. The comment notes that the population which can now produce an `awaiting_confirm` entry is precisely the one fs-60 exists to serve, and that the Kokoro copy would have masked it behind users rationally hitting Skip.

### 🟡 3 — I do not accept this one

The finding analyses `check-register-**row**-citations.mjs`. The gate that was failing — and whose six CLI cases were the `test:hooks` failures — is its sibling `check-register-citations.mjs`, which is the script `ed3edccb`'s message names. Reverting the four annotations and re-running both:

| Script | Without annotations | With |
|---|---|---|
| `check-register-citations.mjs` | **exit 1** | **exit 0** |
| `check-register-row-citations.mjs` | exit 0 | exit 0 |

So the commit does what it claims. Your description of the *row* script is accurate — narrower regex, `SCAN_PREFIXES` excluding `docs/superpowers/` — it just is not the gate in question.

**The distinction earned its keep within the hour**, which is worth recording: flipping plan 249 (finding 7) introduced two new dangling A5 citations, and **only the row script caught them**, because it requires the annotation tied to the same logical line where the other accepts a 25-line window. Both are now green. Two gates, two rules, not interchangeable.

### 🟡 1, 2, 4, 5 — fixed, all mine

1. The mid-sentence insertion was a real botch; moved out of the sentence and the citation re-verified as still covered.
2. Run sheet: "STILL OWED" resolved and the `Result:` filled with the observed outcome, including *why* bullet 5 is N/A (a non-registry language throws at `generation.ts:851` before this gate).
4. Both in-register A5 references annotated — same class as the four external ones, and you were right that fixing only the outside ones was arbitrary.
5. Change log now records the 61 → 60 / A 38 → 37 move, with the A36 entry kept as the previous change.

### 🟡 6 — fixed

Body now states Group A **37**, owed **60**, and explains that the pre-rebase 39→38 / 65→64 was this branch's arithmetic before it took in 101 commits of main.

### 🟡 7 — fixed, `e3ecc0b3`

Plan 249 → `stable`, shipped 2026-09-06, Ship notes filled, moved to `archive/`, INDEX entry moved, four inbound links repointed and each verified to resolve. The plan carries no relative links of its own, so nothing broke outbound — checked rather than assumed.

The Ship note cites PR #3031 alongside `abb667e1`, since that SHA is branch-local until merge and dies on another rebase. I did verify `abb667e1` is the right commit: it carries the 30-line `generation.ts` change, not just the step-4 evidence doc.

### Release notes — both corrections taken

You were right on all three counts, and the brand one was a plain factual error on my part: I blamed a "Proceed anyway" button that **worked before this fix**, since it sets `fallbackConfirmed: true` and the server gate requires `!job.fallbackConfirmed`. It now names the queue path. The technical entry no longer promises what the shipped UI contradicted, and no longer describes the unreachable arm as live behaviour — it points at #3059 instead.

And yes — recorded as owed. Step 5 was unmet and this pass would have raised it.

---

Re-review is triggered (🔴 + 2 🟠). CI green on the intermediate head; re-running on `30e347e9`.

dudarenok-maker:
## PR review — pass 2 (head 30e347e9, depth high)

Scope: the six fix commits `cf7d387f`…`30e347e9` and the five inherited ones under them. Verified before probing: CI 20/20 green at this head; `generation-fallback-gate.test.ts` 8/8; `queue-modal.test.tsx` 37/37; `check:onbox-register`, `check-register-citations`, `check-register-row-citations` all exit 0 against the real tree; `register:build` idempotent (re-ran, no diff); `node --test` over the three register-script test files 312/312. All mutation work ran in an isolated copy outside the repo.

Depth re-derived from the current commit set: still `high` (`fix(server,docs)`, multi-scope).

---

### Pass-1 dispositions

| # | Verdict |
|---|---|
| 🔴 queue row hardcoded Kokoro | **VERIFIED RESOLVED** — details below |
| 🟠 unreachable arm + deleted coverage | **ACCEPTED** — #3059 open, `type:chore,area:server`, decision named in the title. Premise re-checked: `ENGINE_LANGUAGE_SUPPORT.coqui = ['en','ru','es','fr','de','zh','ja']` (`voice-mapping.ts:47`) is exactly the registry code set (`language-registry.ts:45–119`). Legitimate design-pass deferral. |
| 🟠 queue-wedge exposure | **ACCEPTED** — #3026 open, `bug`, carries the interaction. |
| 🟡 1 mid-sentence annotation | VERIFIED RESOLVED (`onbox-sitting-vram-contention.md:40`, own blockquote, sentence intact) |
| 🟡 2 run sheet self-contradiction | VERIFIED RESOLVED (`:160–190`, "STILL OWED" replaced, `Result:` filled) |
| 🟡 3 "annotation commit is a no-op" | **WITHDRAWN — the rebuttal is correct. I was wrong.** Reproduced below. |
| 🟡 4 two in-register A5 refs | **PARTIALLY RESOLVED** — fixed in the register, not in the live view. See N3. |
| 🟡 5 change log | VERIFIED RESOLVED (`onbox-acceptance-register.md:533`) |
| 🟡 6 PR body figures | VERIFIED — `30e347e9`'s register diff touched no row heading, so 37/60 stands |
| 🟡 7 plan 249 | **PARTIALLY RESOLVED** — move and links are correct; two body sentences still say the opposite. See N1. |
| release notes | Re-verified clause by clause; two new defects found. See N4. |

---

### 🔴 — VERIFIED RESOLVED

Answering the three questions put to this pass:

**Is `bookId` always the right book for every row in that group?** Yes, by construction. `selectQueueByBook` (`src/store/queue-slice.ts:118–132`) keys each group on `entry.bookId` and pushes only entries with that id, so the group's `bookId` is every row's own `bookId`. No cross-book leak is reachable.

**Book absent from state, or mid-load?** `getBookFallbackEligibility` (`voice-readiness-selectors.ts:90–103`) falls back to `ALL_TTS_ENGINES` + `'en'` → `'Kokoro'` — byte-identical to the pre-fix hardcoded string, so the miss window degrades to the old behaviour rather than to a new wrong one. The library hydrates at `layout.tsx:571–591` (mount + every return to the books stage), and the same posture is already shared with the three other surfaces reading this selector.

**Could a previously-correct row now be wrong?** No. English → `'Kokoro'` (unchanged). Non-English + Coqui-eligible → `'Coqui'`, which is what the server does (`synthesise-chapter.ts:1500–1523`: `forbidKokoroFallback` is unconditional for non-English, so Coqui or throw — never Kokoro). Non-English without Coqui → the row is unreachable after this PR's own narrowing. Unset language → `requireBookStateLanguage` throws (`scan.ts:375–380`) before the gate. One divergence I specifically looked for and did not find: the wire `language` is already normalised at the seam (`scan.ts:335–337` → `normaliseBookLanguage`), so `'ru-RU'`/`'en-US'` cannot make the client's `=== 'en'` disagree with the server's `isNonEnglish`.

**Mutations (different sites from the aria-label you used):**

| Mutation | Result |
|---|---|
| `const fallbackEngineName: string = 'Kokoro'` — i.e. the original defect restored at the wiring | RED at `queue-modal.test.tsx:339`, status line reads "…would render in Kokoro" |
| `title` attribute only, back to a literal | RED at `:344`, `toHaveAttribute('title', 'Render anyway (Coqui fallback)')` |

Both red for the claimed reason; 36/37 pass in each, baseline 37/37.

---

### 🟡 3 — I was wrong; the rebuttal stands

Reproduced independently, by deleting the four annotation blockquotes from a copy of this head and running both gates against it:

| Script | Without annotations | With |
|---|---|---|
| `check-register-citations.mjs` | **exit 1** — 4 hard errors under "Check A — nonexistent row ID", all `cited A5`, at `249-…:90`, `2026-07-28-coqui-residency-eviction.md:1393`, `onbox-sitting-vram-contention.md:4` and `:160` | **exit 0** |
| `check-register-row-citations.mjs` | exit 0 | exit 0 |

`ed3edccb` does exactly what its message says. My pass-1 analysis of the *row* script's matcher and `SCAN_PREFIXES` was accurate but irrelevant — that is not the gate the commit was fixing, and I should have checked which script was failing before writing the finding. **Withdrawn.**

---

### New findings — all 🟡 cleanup, no correctness bug

The first three are one shape: each pass-1 doc finding was fixed on the surface the finding named and left standing on its siblings.

**N1. Plan 249 is archived as `stable` but still says twice that it is not.** `docs/features/archive/249-fs60-xtts-language-eligibility.md`:
- `:58` — "**Explicitly not covered — Live-GPU acceptance is owed.** … has not been exercised on an 8 GB box. This plan's status stays `active`, not `stable`, until that walkthrough runs."
- `:76` — "**Live-GPU acceptance** … — owed, tracked via this plan staying `active` and the `fs-60` BACKLOG row."

Both contradict `status: stable` in the frontmatter 56 lines above, and both are now false: the walkthrough ran, and there is no `fs-60` BACKLOG row (grep of `docs/BACKLOG.md` returns only the fs-69/70/71 follow-ups). Pass 1 quoted `:9`, which was fixed; these two were not.
Adjacent: `docs/features/271-fs38-wave3c-xtts.md:757–761` cites plan 249 as the precedent for *not* archiving a plan with owed acceptance — that citation now points at a counterexample.

**N2. `Verdict: DISCHARGED` was written into a frozen document, using a value that document's own legend does not define, and four of its six fields still say the opposite.** `docs/testing/onbox-acceptance-staleness-audit.md:363–385`.
- Its verdict legend (`:17–29`) defines exactly four values: `STILL OWED`, `PROPOSE DISCHARGE`, `SHRUNK`, `AMBIGUOUS`. `DISCHARGED` is not one of them.
- Its header (`:10–14`) states: "**No row has been discharged by this audit and the register is unchanged.** This document only proposes verdicts."
- Both citation checkers list this exact path as a **frozen** dated transcript — `check-register-row-citations.mjs:276` `FROZEN_EXACT`, mirrored in `check-register-citations.mjs` — i.e. a record of what the register said on its date, deliberately not retro-edited. Being frozen is also why no gate flags any of this.
- `30e347e9` changed two lines (`:365` verdict, `:366` path) and left: `:366–375` Evidence still quoting "frontmatter `status: active`" and `:9` "Live-GPU acceptance owed" (both now false); `:379` "**What changed since the row was written:** Nothing found."; `:380` "**Remains owed:** All five walkthrough steps … and the Chinese hard-block regression check" — which contradicts both the new verdict and the run sheet's own "bullet 5 N/A / superseded"; `:385` "**Est. box time:** 20", still budgeting box time for a discharged row.

Cleanest fix is to revert this file: it is a frozen dated audit, and the discharge is already recorded in the register's change log and the run sheet.

**N3. The live view did not get finding 4's annotations — and this is the third citation surface.** `docs/testing/onbox-acceptance-register-live-view.html:387` reads "**Run this with A5** — same card, same mixed-cast book…" and `:436` "(runnable alongside **A5/A13** in the same session)". Both counterparts in the register were annotated in `30e347e9` (`onbox-acceptance-register.md:1905`, `:2048`), and the live view is the surface that actually gets published and read.

No gate can catch it, which is the part worth recording: `check:onbox-register`'s no-flag tracked-pair run (`checkLiveView`, `check-onbox-register.mjs:866`) compares glance-table letters and body group row IDs only — per-row content hashing (`hashRowContent`, `:321`) runs solely on the `--against-published` path with a resolved baseline. And neither citation checker sees these: a bare `A5` in prose is not a citation surface for `check-register-citations` (it needs the `row(s) ID` idiom, a `Register row(s):` label, or an anchored `### <ID> ·` heading), and `check-register-row-citations`'s `\b(?:row|Row)\s+([AE]\d+)\b` does not match "with A5" or "alongside A5/A13". So this is a surface with two live gates and zero coverage.

**N4. `fs-60 (#2962)` is the wrong issue, in shipped source and in the release note.** `server/src/routes/generation.ts:1433` — "`#1263, narrowed by fs-60 (#2962)`" — and `docs/release-notes-next.md:194` — "(fs-60, #2962, narrows #1263)". #2962 is `[agent instructions][claude][task] mechanical batch 1 step 4 — A5+A13+A17+A19 pinned single-8GB-card session`, `area:ops`: this acceptance session's own task ticket. fs-60's issue is **#1005** (plan 249's own header: "Issue: `dudarenok-maker/Castwright#1005`"). A future reader following the comment lands on an ops chore. Same line of the release note also says "see the follow-up issue" without naming #3059.

Everything else in both entries re-verified clean at this head, including the two I approved in pass 1 (`generation.ts:938`/`:952` split; `forbidKokoroFallback: nonEnglishBook` at `:1690`). The rewritten brand entry is now accurate — and stronger than it claims: the gate is queue-scoped by construction (`generation.ts:1428` requires `job.queueEntryId != null`, comment "Scope: queue-driven runs only"), so "it showed up when you generated a chapter from the queue" is right, and the "Proceed anyway" attribution is correctly gone.

**N5. All three new queue-modal fixtures invent `eligibleTtsEngines` payloads the server cannot emit — and one of them makes its test unable to fail.** `src/modals/queue-modal.test.tsx`.
- `:360` English book with `eligibleTtsEngines: ['qwen','kokoro']`. `resolveEligibleEngines('en', ALL_TTS_ENGINES)` always includes `coqui` (`ENGINE_LANGUAGE_SUPPORT.coqui` contains `'en'`), so no library payload can look like this. Measured consequence: deleting `!isEnglish &&` from `voice-readiness-selectors.ts:100` leaves the whole new suite **37/37 green**. The class is still guarded — `voice-readiness-selectors.test.ts` and `voice-readiness-gate.test.tsx` each go red on that same mutation — so this is a hole in the new coverage, not an unguarded regression. Giving the English fixture its real payload (`['qwen','coqui','kokoro']`) closes it.
- `:389` "non-English book without Coqui support", `language:'es'` + `['qwen','kokoro']`: doubly impossible — `kokoro: ['en']`, and `'es'` *is* Coqui-eligible. Its comment at `:396` ("Non-English but no Coqui support → falls back to Kokoro") also states as expected behaviour something the server never does, since `forbidKokoroFallback` is unconditional for non-English. The selector's own doc (`voice-readiness-selectors.ts:72–84`) records that exact UI/server divergence as a known, today-unreachable fs-70 state; this test now pins it as correct without saying so.
- `:330` is the mild case: `['qwen','coqui','kokoro']` for `'ru'` (the real book in the step-4 evidence computed `["coqui","qwen"]`). Harmless to the assertion, but same habit.

**N6. Comments this change made false, in the file it changed.** `src/modals/queue-modal.tsx:264` ("render this parked chapter anyway (in Kokoro)") and `:270` ("rather than render undesigned voices in Kokoro") describe the same two controls whose copy this PR made engine-aware — the comment at `:558–560` *was* updated, these two were not. Same in `src/store/queue-thunks.ts:273`.

---

### Verdict

**Nothing blocks merge.** The 🔴 is genuinely fixed and holds under two independent mutations; both 🟠 are routed correctly, with #3059's unreachability premise re-verified from the code rather than taken from the report. Every new finding is 🟡 cleanup, so no re-review is triggered.

N1–N3 are worth closing out in this round rather than after merge, because they are the same finding class pass 1 already raised — the correction landed on the surface each finding named and not on its siblings, which is the pattern that produced the dangling A5 citations in the first place. N4 is a wrong issue number in shipped source.

Issue #3023, opened 2026-09-06
[agent instructions][cline-free][task] Fix copy-pasted intelligibility wording in the cloned-voice language warning (#1998)


dudarenok-maker:
## What's wrong

`server/src/routes/voice-library.ts` (branch `feat/server-1998-clone-lang-warning`,
commit d8b85e60) implements the cloned-voice language-mismatch warning by copying
the **designed**-voice warning's wording verbatim and swapping "design"→"clone":

```
`"${character.name ?? characterId}"'s voice was cloned in ${cloneLangName} but this ` +
`book is ${bookLangName} — the audio will be unintelligible. Re-clone the voice in ` +
`${bookLangName} to fix it.`;
```

Parent #1998's acceptance item 2 requires the opposite:

> The warning names the character, the clone's language and the book's language,
> and says the clone will sound less like the person — an **identity** loss, not
> the designed branch's **intelligibility** loss. The two messages must not be
> copy-pasted into each other.

This is not a wording nitpick — it misrepresents the actual failure mode. The
measured effect (#1998's own table, ~0.23 vs ~0.60 ECAPA cosine identity) is
identity drift, not unintelligibility: the cloned voice still speaks the book's
language fine, it just sounds less like the source speaker. Telling the user
"the audio will be unintelligible" is factually wrong for this branch and is
indistinguishable from the designed-voice warning if a user sees both over time.

The bug reproduces in the shipped code (`server/src/routes/voice-library.ts`,
the `entry.provenance === 'cloned'` block added under the `#1998` comment,
current lines ~1857-1886) and is **also baked into the test suite** — both
`server/src/routes/voice-library.test.ts` (#1998 describe block) and
`src/modals/profile-drawer.test.tsx` (the `#1998 surfaces the server's...`
test) assert/mock the wrong wording, so a naive "make tests pass" fix would
reinforce the bug rather than remove it. Fix the tests' expectations too.

## What to change

1. In `server/src/routes/voice-library.ts`, rewrite the `languageWarning`
   string built in the `provenance === 'cloned'` branch so it names the
   character, the clone's language, and the book's language, and states an
   **identity** loss (the clone will sound less like the person it was
   cloned from), not an intelligibility loss. Do not reuse the designed
   branch's sentence template — write an independent one for this branch.
2. Update the two tests that currently assert/mock the wrong wording:
   - `server/src/routes/voice-library.test.ts`, describe block
[...]

Issue #1951, opened 2026-07-29
fs-38 — cloned/self-healed Qwen voices synthesise as English: X-Language is never sent, and the warm cache diverges from the restored manifest


dudarenok-maker:
﻿## What

`server/src/tts/derive-engine-artifact.ts:62-73` is the **only** Node caller of the sidecar's `POST /qwen/clone-voice`, and it never sends an `X-Language` header. So `language=None` arrives at `main.py:8578`, and `clone_voice` always computes `lang = DEFAULT_LANGUAGE` (`QWEN_LANGUAGE`, default `"English"`).

That `lang` is then used in two places that matter:

1. written to the sidecar manifest as `"language": "English"` (`main.py:5276`);
2. passed straight into `generate_voice_clone(..., language=lang, ...)` at `main.py:5774` (single) and `:5906` (batch). **There is no per-request language override on the Qwen synth path.**

## Consequence A — cache-vs-disk divergence on the designed-voice self-heal path (CONFIRMED)

Since PR #1942, `clone_voice` also warms `_prompt_cache[voice_id] = (prompt, lang)` — so it now caches `"English"` too.

Meanwhile `server/src/tts/clone-voice-resolver.ts:1000-1007` deliberately overwrites the manifest after the derive with the **pre-derive designed** one (`{...master.manifest, refText, baseModel}`) to recover `instruct`/`designModel` — restoring e.g. `"language": "German"`. No `/qwen/evict-voice` follows; `beforeFirstQwenDerive` (`:977`) evicts *before* the derive, not after.

Net state: **disk says `"German"`, the warm in-process cache says `"English"`.**

Repro: a German designed voice loses its `.pt` → the chapter's self-heal pre-pass re-derives it → every subsequent sentence in that sidecar process synthesises German text as English. Silently degraded, and **restarting the sidecar changes the output**, which makes it miserable to diagnose.

## Consequence B — non-English clones may synthesise as English even on a cold cache (PLAUSIBLE)

The same root cause has a second, cache-independent effect: because Node never sends `X-Language`, **every** cloned voice's manifest records `"English"`, so a non-English clone would synthesise as English on a cold cache too. The reviewer found no compensating language fixup but did not exhaustively rule out a language gate on the clone UI — **verify this before designing the fix, since it changes the severity considerably.**

## Why this was filed rather than folded into #1942

[...]

PR #3033, opened 2026-09-06
feat(server,openapi,frontend): warn when a cloned voice's language mismatches the book

### LINKED ISSUE #1998: A cloned voice on XTTS loses most of its speaker identity when rendered in a language other than the source clip's (0.600 -> 0.229)

dudarenok-maker:
## Summary

Three things on one seam: **a voice cloned from a clip in one language sounds markedly less like the speaker when it renders another.** This PR warns at assign time, tells the user before they record, and records the measurement behind the advice.

### 1. The server warns when a cloned voice's language mismatches the book

`POST /api/voice-library/:voiceUuid/assign` returns **200 with a `warning`** naming the character, the clone's language and the book's:

> `"<name>"'s voice was cloned in <X>, but this book is <Y> — the clone will sound less like the person it was cloned from. Re-clone the voice in <Y> to fix it.`

Deliberately a warning, not a 409 — the decision is expectation-setting, not engine rerouting, and the assignment is legitimate.

The consequence is **loss of resemblance, not unintelligibility** (~0.23 ECAPA cosine identity vs ~0.60 same-language). That is a *different* defect from the request-language bug fixed under #1951, where the synthesiser rendered the wrong language outright. Worth stating because a draft of these release notes conflated the two and imported #1951's `avg_logprob` figures.

Comparison is **code-vs-code** (`entry.languageCode`, BCP-47, against the book's language), never code-vs-sidecar-word. Suppressed when: the entry has no `languageCode`, the book has no language, the languages match, or either language is unregistered with the sidecar. When #1933's engine-readiness advisory also applies it wins — engine readiness is the more actionable message.

### 2. The clone wizard says so before you record

A hint in the capture phase, placed before the recording decision — the only point at which the advice is actionable.

### 3. The measurement behind it

| Source clip language | Rendered chapter language | ECAPA identity |
|---|---|---|
| English | Russian | **0.2321** |
| Russian | Russian | **0.7824** |

**Not a new measurement** — Run 3 (2026-08-01) of register row **A1**, criterion **E-01** of the Wave 3 run sheet, on a **Coqui-routed** clone. Recorded here because it answers #1998's open acceptance bullet; the register stays the source of record.

**The caveat travels with the figure:** the Russian source clip was itself a Qwen-designed voice and Qwen's Russian voices cluster — narrator vs that source alone measures **0.577**. The margin is real but narrower than 0.7824 suggests.

## Review history — 16 findings across two passes, all fixed

Per CLAUDE.md every finding is fixed in-round; none were filed for later.

### Pass 1 (head `90edc21f`) — 3 🟠, 6 🟡

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | 🟠 | Warning claimed a book was English when its language was **unset** | `84d407b5` |
| 2 | 🟠 | Release notes omitted the PR's primary behaviour | `749a8cb7` |
| 3 | 🟠 | Documented #1933-beats-#1998 precedence had **no test** | `e2d41203` |
| 4 | 🟡 | Wizard said "this book" in a book-independent mount | `adeaf12c` |
| 5 | 🟡 | Two frontend comments said `warning` meant only #1953 | `8435fb4e` |
| 6 | 🟡 | A comment claimed mutation 2 pins code-vs-word; it cannot | `3e810dea` |
| 7 | 🟡 | Measurement re-dated as fresh, engine unnamed | `eb0cf304` |
| 8 | 🟡 | openapi prose over- and under-stated the contract | `c8e0e114` |
| 9 | 🟡 | No cloned-branch unregistered-language test | `290381ce` |

### Pass 2 (head `adeaf12c`) — 2 🟠, 5 🟡, **both 🟠 caused by pass 1's own fixes**

| # | Sev | Finding | Fix |
|---|---|---|---|
| A | 🟠 | Comparand became **unnormalised**: `"ru-RU"` book vs `ru` clone warned falsely | `82aa9f27` |
| B | 🟠 | The "no `languageCode`" test went **vacuous** — the book conjunct carried it | `82aa9f27` |
| C | 🟡 | Stale comparand claims in guard comment and test header | `82aa9f27` |
| D | 🟡 | `bookLanguage`'s "both branches" comment — **resolved by A's fix**, no change | — |
| E | 🟡 | Release notes still quoted the wizard framing finding 4 rejected | `977311eb` |
| F | 🟡 | A comment cited a line number that had drifted twice | `977311eb` |
| G | 🟡 | Precedence stated twice in openapi; a negative assertion duplicated | `977311eb` |

### The two findings worth reading

**Finding 1** — `bookStateLanguage` defaults absent / `null` / empty / whitespace to `'en'`, and `language: null` is a **deliberately preserved** state meaning detection surrendered (`state-migrate.ts:127`, `scan.ts:341`, `import/scan-import-folder.ts:243`). A Russian clone on such a book was told *"this book is English — Re-clone the voice in English to fix it"*: a false fact, and advice that destroys a correct clone. `scan.ts:368`'s doc already reserved the honest reader for paths that change behaviour on absence.

**And the tests could not have caught it.** All three cloned-warning tests seeded books via `writeBookOnDisk`, which wrote **no `language` key** — so the test titled *"assigned to an English book"* never made an English book. The wrong reading was written into the test titles as if correct.

**Finding A** — the fix for finding 1 then over-reached. `bookStateLanguageOrNull` is honest about absence but returns `state.language` **raw**, while `bookStateLanguage` normalises *and* defaults; the swap took the no-normalisation half too. Since `entry.languageCode` is stored normalised (`:1220`), a book saved as `"ru-RU"` stopped matching a `ru` clone — and because `sidecarLanguageName` normalises internally, both sides rendered "Russian", producing *"cloned in Russian, but this book is Russian."* Now split as `chapter-splice.ts` already does it: **presence** from the honest reader, **value** from the normalising one.

## Test plan

- **`voice-library.test.ts`: 191 passing** (was 184 when review pass 1 ran; +7).
- **Frontend affected suites: 122 passing.** `npm run typecheck` clean; `eslint` clean on every touched file.
- **`npm run openapi:types` regenerates to no further diff**, with the generator's own output confirming it ran (`openapi-typescript 7.13.0`) — an empty `git status` can also mean the binary is absent, which is a false clean.
- **Rebased onto `9693e89a` and re-verified after the rebase**, not before; this branch had been 26 commits behind.

**Every regression test here was verified red-before-green by the shipping session, independently of the agent that wrote it**, checking *which* assertion reddened — a mutation that makes the route 500 trips the status assertion first and never reaches the one under test. That mattered repeatedly:

- Finding 1's tests: reverted the production fix, both new tests failed with the exact false string, restored byte-identical.
- Finding 3's test: swapped the `??` operands, test red, `voice-library.ts` restored byte-identical.
- Finding A's fix: reverted the comparand to the raw reader — both normalisation tests failed at `expect(res.body.warning).toBeUndefined()`, not at a status assertion. Finding B's: deleted the `entry.languageCode &&` conjunct — the previously-vacuous test failed at the same assertion, reporting *"Ada's voice was cloned in English"* for an entry with no language code at all, so that conjunct guards a fabricated claim rather than being a redundant presence check.
- **A correction I owe the record:** an earlier version of this body said finding 3's fix agent "claimed negative assertions it had not written". It had written them, further down the test than I read before adding a second set. The duplication pass 2 flagged as finding G was mine; the agent's report was accurate.
- Finding 9's tests: the agent proved them by deleting the `catch`, which only shows the catch prevents a 500 — the status assertion fires and the `warning` assertion never runs. Registering `pt` in the real registry instead made both lookups resolve, and both tests then failed on `expect(res.body.warning).toBeUndefined()` with the real emitted text. The pre-existing designed-voice test reddens under the same probe, confirming the mechanism.

### Checklist notes

- **Regression plan** — `docs/features/275-clone-voice-language.md` updated; measurement now cited to its register row rather than restated.
- **On-box acceptance** — no new row owed. The cross-language measurement is already registered as A1 / E-01 Run 3; this PR cites it, it does not claim fresh hardware evidence.
- **Release notes** — both surfaces updated. Brand-voice entry carries no identifiers and does not blame the user for an unset book language, which can come from a surrendered detection.

Closes #1998
Refs #3023
Refs #624


dudarenok-maker:
## PR review — pass 1 (head 90edc21f, depth high)

Scope: `server/src/routes/voice-library.ts`, `openapi.yaml`, `src/lib/api-types.ts`, `src/modals/clone-voice-wizard.tsx`, all three test files, `docs/features/275-clone-voice-language.md`, both release-notes surfaces · verified independently: `voice-library.test.ts` **184 passed**, `profile-drawer.test.tsx` + `clone-voice-wizard.test.tsx` **122 passed**, and `openapi-typescript 7.13.0` regenerated from the committed `openapi.yaml` into a temp file is **byte-identical** to the committed `src/lib/api-types.ts` (`diff` exit 0) — before adversarially probing the warning precedence, the `bookLanguage` hoist, the code-vs-word seam, provenance/engine reachability, and the measurement provenance.

---

### 🟠 Significant — the warning fires with "this book is English" on a book whose language is **not set**, and every new test is blind to it because no fixture book has a language

`server/src/routes/voice-library.ts:1721`

```ts
const bookLanguage = bookStateLanguage(located.state);
```

`bookStateLanguage` (`server/src/workspace/scan.ts:335`) is `normaliseBookLanguage(state.language)`, which returns `'en'` for absent, `null`, empty and whitespace-only. `language: null` is not a hypothetical — it is a deliberately-preserved on-disk state: `server/src/workspace/state-migrate.ts:127` ("`language: null` — 'stated absence' — must survive to disk verbatim"), `server/src/import/scan-import-folder.ts:243`, and `scan.ts:341` ("a stored `null` (set on a surrendered detection) means unset"). This release's own notes ship the path that produces it ("a short book … can come back undecided instead of guessed at").

**Repro:** book `state.json` = `{"bookId":"b1", …, "language": null}`; cloned entry with `languageCode: 'ru'`; `POST /api/voice-library/<uuid>/assign` → **200** with

> `"Nikolai"'s voice was cloned in Russian, but this book is English — the clone will sound less like the person it was cloned from. Re-clone the voice in English to fix it.`

The book is not English; it is undecided. The user is told a false fact about their book and given advice that destroys a working clone. The repo already has the reader built for exactly this class of path — `bookStateLanguageOrNull` (`server/src/workspace/scan.ts:368`), whose own doc says *"The paths that change behaviour on absence use this; the display paths keep `bookStateLanguage`"*. This path changes behaviour on absence.

**Why no test caught it, and why that is the load-bearing half:** all three new tests seed their book through `writeBookOnDisk` (`server/src/routes/voice-library.test.ts:107-138`), which writes a `state.json` with **no `language` key at all**. So the test titled *"warns (200) when a cloned voice in Russian is assigned to **an English book**"* never creates an English book — its `expect(res.body.warning).toMatch(/English/)` is satisfied purely by the `'en'` default. Likewise *"does not warn when the cloned voice language matches the book language"* asserts a match against a language the fixture never states. **The cloned branch has zero coverage against a book that actually declares a language**, so the wrong-default behaviour is not merely uncaught — it is written into the test titles as correct. (The designed `#1953` block does have a real fixture at test line 3098 (`language: 'ru'`); the cloned block has no equivalent.)

Fix needs a decision on one point (warn using the honest reader and skip on unset, vs. keep the default and reword) — but "skip when the book has no stated language" is the shape the neighbouring code already uses for the unregistered case, so I do not think this needs a design pass.

---

### 🟠 Significant — the release-notes pair omits the PR's primary shipped behaviour

`docs/release-notes-next.md:90`, `RELEASE_NOTES.md:49`

Both entries describe **only** the clone-wizard hint and the measurement. Neither mentions the thing this PR mostly is: `POST /api/voice-library/:voiceUuid/assign` now returns a 200 with a `warning` when a cloned voice's language mismatches the book, its `openapi.yaml` contract prose changed, and the warning surfaces in the profile drawer and the cast-page voice-library panel. That is a user-visible behaviour change shipping with no note on either surface.

The cause is visible in the history: commit `b12e943f`'s own message says *"Server warning (child 1) and its frontend surfacing (child 2) are untouched, per this issue's scope."* The notes were written for one child ticket (#2898) and never widened when the three children became one PR. CLAUDE.md before-shipping step 5 wants the shippable delta on both surfaces, PR-by-PR — this is the delta.

---

### 🟠 Significant — the documented `#1933`-beats-`#1998` precedence has **no test**, and the PR body claims it does

`server/src/routes/voice-library.ts:1944` (`const warning = clonedAdvisory ?? languageWarning;`), contract at `openapi.yaml:2653-2655`.

PR body, Test plan: *"`server/src/routes/voice-library.test.ts` — the warning path, both provenance branches, and the `clonedAdvisory`-wins-over-language-warning precedence."* No such test exists in the diff or the file:

- `seedClonedWithLang` (test:3266) always writes `master.wav`, so `clonedClipOnDisk === true` → the other clone engine is derivable → `clonedAdvisory` is never set in any of the three new tests;
- `makeEntry` (test:140) never sets `languageCode`, so T1 / T9 / T12 / T13 / T14 — every advisory test — never set `languageWarning`.

At most one of the two is ever defined in any test in the file. `a ?? b` and `b ?? a` are indistinguishable under that condition, so **swapping line 1944 to `languageWarning ?? clonedAdvisory` passes all 184 tests** while inverting the contract the openapi text now promises. New documented behaviour with no paired test, per CLAUDE.md testing discipline — and a Test-plan claim that is not true.

(The `??` itself is correct as written: `clonedAdvisory` is only ever assigned a non-empty string from `advisoryMessage`, so the advisory does win and neither branch silently swallows the other by accident. The gap is the pin, not the logic.)

---

### 🟡 Minor

**1. Two `warning` consumers still say the field means only `#1953`.**
`src/modals/profile-drawer.tsx:492-497` — *"the server's `warning` is a non-fatal advisory (a designed voice's baked language doesn't match this book's)"* — and the identical claim at `src/components/voice-library-panel.tsx:314-318`. As of this PR the field carries three signals. Already half-stale from #1933; this change makes it false for the case the PR exists to ship, and the PR's own frontend test (`profile-drawer.test.tsx:2321`) asserts against a cloned message the comment says cannot occur.

**2. The mutation-2 rationale names a mechanism the cloned branch does not have.**
`server/src/routes/voice-library.ts:1866-1868` — *"comparing against `expectedSidecarLang` would silently fire on every correctly-matched assignment because a BCP-47 code is never equal to its human-readable sidecar name"* — and `server/src/routes/voice-library.test.ts:3334-3339` — *"'en' !== 'English' is always true"*. In the cloned branch `expectedSidecarLang` is **always `undefined`**: it is declared at line 1722 and assigned only inside `if (entry.provenance === 'designed')` at 1723-1731. The mutation goes red because `'en' !== undefined`, never because a code was compared to the word `'English'`. The trap named in the comment is not the trap at this site; a reader trusting it will mis-model the guard.

**3. The plan doc's measurement section is dated for a run that the register already records five weeks earlier, and names no engine.**
`docs/features/275-clone-voice-language.md:458` — *"Cross-language source clip measurement (**on-box, 2026-09-04**)"*. The identical three figures (0.7824 / 0.2321 / 0.577) are already on `main` at `9693e89a`, in `docs/testing/onbox-acceptance-register.md:1282-1291`, attributed to **E-01 Run 3, 2026-08-01**. Two problems:
   - the plan doc's date reads as a second, independent measurement, so a future reader double-counts one run's evidence (the PR body's "Measured on the box:" reinforces this; commit `b12e943f`'s "Record the measured …" is the honest phrasing);
   - the new table names **no render engine and no measurement path**, while the register says E-01 is a *Coqui-routed* Russian book (register:1048) and #1998's own body flags precisely this confound (*"XTTS same-language fidelity is itself lower than Qwen's — 0.600 vs 0.885"*). A reader of the plan doc alone cannot tell whether the 0.2321 and 0.7824 rows are the same engine. #1998's acceptance bullet 3 also asked for the numbers to be recorded *"so a low cosine on a Coqui clone is not mistaken for a defect"* — the issue's own 0.600 / 0.229 / 0.014 / 0.885 table is not carried into the docs, so that half is not discharged by `Closes #1998`.

**4. The wizard hint says "this book" in a surface that has no book.**
`src/modals/clone-voice-wizard.tsx:104-106`. `CloneVoiceWizard`'s only mount point is `MyVoicesSection` (`src/components/voices/my-voices-section.tsx:98`), whose header comment calls it a *"Book-independent library"* and whose empty state tells the user these voices *"stay book-independent and ready to cast anywhere"*. The wizard takes no `bookId` and reads no book language, so "the language you want **this book** read in" points at a referent the surface explicitly disclaims. ("the language of the books you'll use this voice for" would be true, and the book language is in redux if you want to name it.)

**5. The openapi bullet for `#1998` both over- and under-states its trigger.**
`openapi.yaml:2636-2643` (and generated `src/lib/api-types.ts:9229-9235`): *"set when a CLONED voice's `languageCode` … differs from the book's language"*. It is **not** set when either language is missing from the language registry — both `sidecarLanguageName` calls at lines 1879-1880 are caught and the warning is skipped. That case is reachable for the book side exactly as the #1953 comment at 1706-1715 already records (`routes/import.ts` persists an unchecked language, #1955 — a `pt` book is on disk today), so a Russian clone on a `pt` book silently warns about nothing. Conversely it *is* set when the book has no language at all (see the 🟠 above). Neither is documented.

**6. The `#1953` block's unregistered-language sibling test was not mirrored for the cloned branch.**
`server/src/routes/voice-library.test.ts:3145-3186` pins "unregistered book language → 200, no warning, no 500" for the designed branch. The cloned branch calls `sidecarLanguageName` **twice** (one more throw site than the designed branch) and has no equivalent case.

---

### ✅ What is solid

- **The `??` precedence is what the PR claims** — `clonedAdvisory` is only assigned a non-empty `advisoryMessage` string, so the advisory wins and neither warning can silently swallow the other by construction. The designed branch is byte-identical to before: `expectedSidecarLang`/`designedManifest` are still fetched only under `provenance === 'designed'`, and the new block is gated on `provenance === 'cloned'`.
- **The `bookLanguage` hoist is safe.** `bookStateLanguage` is pure, does no I/O and cannot throw (`normaliseBookLanguage` has no throw path), so computing it for a cloned/imported entry adds no new 500 and no new work inside or outside the `withCastLock` window. The classic hoist hazards (a new throw on a previously-working route, an I/O call pulled into the RMW) are both absent.
- **Code-vs-code genuinely holds — I could not find a path where a sidecar language *word* reaches the comparison.** `entry.languageCode` on a **cloned** entry is written at exactly one site, `routes/voice-library.ts:1348`, from `normaliseBookLanguage(candidate.master.languageCode)` gated on `isSupportedLanguage` — always a lower-cased primary subtag, so no `'ru-RU' !== 'ru'` false positive and no `'English'`. The raw Whisper detection stays on `master.languageCode` (`workspace/voice-library.ts:60-79`) and is never read here. The transcript-edit path clears both (`routes/voice-library.ts:686-687`). The only *unvalidated* client-supplied `languageCode` write is the `/design` route (line 217/244), which produces `provenance: 'designed'` and so never reaches this branch.
- **Reachability across provenance/engine.** Cloned-on-Qwen and cloned-on-Coqui both reach the block (it is provenance-scoped only, and sits after the wrong-engine and routed-engine-readiness 409s, which correctly still win). A cloned entry with no `languageCode` short-circuits with no warning — tested. `provenance: 'imported'` has no server-side write path at all, so its omission is not a gap.
- **The class is covered, not just the drawer.** The second assign entry point, `src/components/voice-library-panel.tsx:313`, already surfaces `warning` through the same generic `else if (warning)`, so the Cast-page Assign flow gets the new warning too without a change.
- **The generated artifact is genuinely in sync** — verified by running the generator, not by an empty `git status`.
- **The caveat travels.** 0.577 accompanies 0.7824 in the plan doc, `docs/release-notes-next.md`, the PR body **and** the comment posted on #1998; `RELEASE_NOTES.md` quotes no figure so needs none. The brand-voice line leaks no internal identifiers and does not overstate.
- **Mock parity probed, not filed.** `mockAssignLibraryVoice` (`src/lib/api.ts:9756-9768`) emits only the #1933 advisory — but mock books (`src/mocks/canned-data.ts`) carry no `language` field at all and the pre-existing #1953 warning has no mock counterpart either, so this is a structural gap in the mock layer rather than something this PR regressed or can cheaply close.

### Verdict

Three correctness-grade findings. The one that matters is the first: on a book with no stated language the feature emits a confident false claim plus destructive advice, and the three new tests cannot see it because their fixture book has no language either — so the wrong reading is baked into the test titles. The precedence gap and the release-notes omission are both cheap. Everything the PR body asserts about the code seam (code-vs-code, hoist safety, `??` precedence, regenerated types, suite counts) checked out under independent verification; the untrue claim is the Test-plan line about precedence coverage.

dudarenok-maker:
## Pass 1 findings — all 9 fixed, `90edc21f` → `adeaf12c`

Nothing filed for later. Dispositions, one commit each:

| # | Sev | Fix | Note |
|---|---|---|---|
| 1 | 🟠 | `84d407b5` | honest reader `bookStateLanguageOrNull`; designed branch unchanged |
| 2 | 🟠 | `749a8cb7` | release notes widened to the assign warning, openapi, both surfaces, the suppression |
| 3 | 🟠 | `e2d41203` | precedence pin, both directions |
| 4 | 🟡 | `adeaf12c` | wizard no longer says "this book" |
| 5 | 🟡 | `8435fb4e` | the two `warning` call-site comments |
| 6 | 🟡 | `3e810dea` | mutation-2 rationale corrected |
| 7 | 🟡 | `eb0cf304` | measurement cited to A1 / E-01 Run 3, Coqui-routed |
| 8 | 🟡 | `c8e0e114` | openapi contract matched to the code |
| 9 | 🟡 | `290381ce` | both unregistered-language skips |

**189 tests** in `voice-library.test.ts` (was 184 when you reviewed), 122 frontend, tsc and eslint clean.

### Three of your findings were sharper than they first read

**Finding 1** — you were right that the tests were the load-bearing half. `writeBookOnDisk` wrote no `language` key, so *"assigned to an English book"* never made an English book. I'd run 184 green tests locally and would have merged it.

**Finding 3** — confirmed by mutation: swapping to `languageWarning ?? clonedAdvisory` passed everything before the new test.

**Finding 8** — half of it was self-inflicted. The prose said the warning *is* emitted on an unset book language; that was true when written and made false by `84d407b5`, earlier in this same PR.

### On your finding 6 (mutation-2 rationale)

Confirmed and slightly widened. `expectedSidecarLang` is assigned only at `:1736`/`:1738`, both inside the `designed` branch, so on the cloned path it is permanently `undefined` — the swap reddens on `'en' !== undefined`, exactly as you said. The code-vs-code rule is kept but relabelled a *design rule* rather than a demonstrated one, and mutation 2's real property (honest reader vs defaulting reader) is named instead. The test-side comment was already correct, having been rewritten by finding 1's fix.

### Verification note

Every regression test was re-verified red-before-green by me, independently of the agent that wrote it. Worth flagging for finding 9: that agent proved its tests by deleting the `catch`, which only shows the catch prevents a 500 — the status assertion fires first and the `warning` assertion never runs. Registering `pt` in the real registry is the probe that tests suppression; both tests then fail on `expect(res.body.warning).toBeUndefined()` with the real emitted text, and the pre-existing designed-voice test reddens the same way.

Re-review round to follow once CI is green on `adeaf12c` — the three 🟠s trigger it.

dudarenok-maker:
## PR review — pass 2 (head adeaf12c, depth high)

Scope: the 9 fix commits `84d407b5..adeaf12c` re-read against the full branch diff (`9693e89a..adeaf12c`, 12 files) · verified: `manifestSlotFor`/`clonedAssignBlock`/`advisoryMessage` call graph, `language-registry.ts` `ENTRIES`, every `state.language` writer in `server/src`, all four `bookStateLanguageOrNull` call sites, register row A1 + the Wave 3 run sheet, and issue #1998's own body — before adversarially probing the fixes themselves.

### Pass-1 dispositions

| # | Sev | Disposition |
|---|---|---|
| 1 | 🟠 | **RESOLVED, but the fix regressed something else** — see 🟠 A |
| 2 | 🟠 | **VERIFIED RESOLVED** (wizard half has residue — see 🟡 E) |
| 3 | 🟠 | **VERIFIED RESOLVED** |
| 4 | 🟡 | **VERIFIED RESOLVED** in the UI; residue in both release-notes files — see 🟡 E |
| 5 | 🟡 | **VERIFIED RESOLVED** |
| 6 | 🟡 | **HALF RESOLVED** — the claim survives in the test file — see 🟡 C |
| 7 | 🟡 | **VERIFIED RESOLVED** |
| 8 | 🟡 | **VERIFIED RESOLVED** |
| 9 | 🟡 | **VERIFIED RESOLVED** |

Detail on the four that needed real work to confirm:

- **#3 (precedence).** The mechanism is the one claimed, not an incidental one. `manifestSlotFor('coqui') === 'xtts'` (`tts/clone-engines.ts:49-51`), so the fixture's `engines.xtts = { status: 'failed' }` really does drive `clonedAssignBlock` rule 1 → `advisoryMessage('failed','coqui',…)`. Routed engine is `qwen` (character `ttsEngine: 'qwen'`, slot `ready`), so the routed-engine 409 does not pre-empt. The language warning is genuinely co-arranged: `'ru'` clone vs `'en'` book, both registered, so `languageWarning` is set and `??` really does have two live operands.
- **#7 (measurement citation).** Correct in every part. `onbox-acceptance-register.md:1048` names E-01 as "clone → Coqui-routed Russian book"; `:1282-1289` is Run 3, dated 2026-08-01, carrying **0.7824** / **0.2321** and the "narrator vs RU source already 0.577, Qwen Russian voices cluster" caveat verbatim. `fs38-wave3-onbox-acceptance.md:2751` puts E-01 under **§5** (§6 begins at :3078), and `:3189` repeats the same figures. Row **A1**, criterion **E-01**, Run 3, Coqui-routed — all four correct.
- **#8 (openapi prose).** Checked clause-by-clause against `voice-library.ts:1890-1906`, independently: "cloned voice's `languageCode`" → `entry.provenance === 'cloned'` + `:1892`; "code-versus-code" → `:1894` compares two BCP-47 values; "absent for an entry with no `languageCode`" → `:1892`; "for a book with no language set" → `:1893`; "for a matching-language cloned voice" → the `!==`; "when either language is not registered" → the two `try/catch` at `:1898-1899` plus the `cloneLangName && bookLangName` gate at `:1900`; "(#1933) takes priority" → `:1963`. Eight clauses, eight lines. Cases the prose does **not** mention and does not need to: a designed or imported entry (excluded by the `provenance` test), and the cloned wrong-engine 409 (already covered by the existing "Either way the assignment still succeeds" paragraph).
- **#9 (`pt`).** `pt` is genuinely absent from `language-registry.ts`'s `ENTRIES` (en, ru, es, fr, de, zh, ja). And the tests **cannot silently decay**: if `pt` is ever added, `sidecarLanguageName('pt')` starts resolving, the warning is emitted, and `expect(res.body.warning).toBeUndefined()` fails loudly in both directions. The failure mode this repo fears is not present here. (Registry membership, not the `supported` flag, is what the code path keys on — so even a `supported:false` `pt` entry would redden them, which is the right sensitivity.)

Also verified for pass-1 #2: no `avg_logprob -1.303` residue anywhere in the #1998 text. Every occurrence in the tree (`release-notes-next.md:89` and `:225`, `voice-library.ts:1692`, `275-clone-voice-language.md:40` and `:418`) belongs to #1951 or #1953. The `~0.23 / ~0.60` pair is correctly sourced: #1998's own title records `0.600 -> 0.229`, and `voice-library.ts:1870-1873` — the exact range cited in the release note — carries that pair. `scan.ts:368` is likewise the right line for `bookStateLanguageOrNull`.

---

### 🟠 A — correctness: finding 1's fix dropped normalisation with the default, so the cloned check now compares a normalised code against a raw one

`server/src/routes/voice-library.ts:1731`, consumed at `:1894` and `:1899`.

`84d407b5` swapped `bookStateLanguage(located.state)` for `bookStateLanguageOrNull(located.state)`. Those two readers differ in **two** ways, not one:

```
scan.ts:336   bookStateLanguage        → normaliseBookLanguage(state.language)   // primary subtag, lower-cased, defaults 'en'
scan.ts:369   bookStateLanguageOrNull  → state.language!                          // RAW, no normalisation at all
```

The fix wanted the second difference (no `'en'` default) and silently took the first as well. `entry.languageCode` on a cloned entry is always normalised — `voice-library.ts:1220-1222` is the only writer and it stores `normaliseBookLanguage(detectedLanguage)`. So `:1894` now compares a normalised value against an unnormalised one.

Concrete break. Book on disk with `state.language: "ru-RU"`, cloned entry with `languageCode: 'ru'`, assign:

- `:1894` → `'ru' !== 'ru-RU'` → **true**, enters the branch;
- `:1898` → `sidecarLanguageName('ru')` → `'Russian'`;
- `:1899` → `sidecarLanguageName('ru-RU')` → `normaliseBookLanguage` strips the region internally (`language.ts:17-25`) → **also `'Russian'`**;
- both truthy → 200 with:

> `"Nikolai"'s voice was cloned in Russian, but this book is Russian — the clone will sound less like the person it was cloned from. Re-clone the voice in Russian to fix it.`

A self-contradictory warning telling the user to re-clone in the language they already used. Same output for `"RU"` and for `" ru "` (`bookLanguageIsUnset` trims only for the emptiness test at `scan.ts:344-347`, then `:369` returns the untrimmed string).

This is a regression, not a pre-existing hole: `git show 84d407b5` shows the comparand was `bookLanguage` — normalised — before this PR's own review-fix commit.

Reachability: every API writer normalises (`routes/import.ts:356-359`, `routes/book-state.ts:851`), but `import/scan-import-folder.ts:246`, `routes/book-state.ts:272`, `chapters-restructure.ts:171`, `library-sync-manifest.ts:58` and `scan.ts:762` all round-trip `state.language` **verbatim**, so a bundle-imported or legacy `state.json` carries whatever it carries. This PR's own reasoning already treats arbitrary on-disk language values as reachable — `voice-library.ts:1706-1716` argues it for `'pt'`, and the PR spends two new tests on that premise.

The repo already has the right shape and states it: `scan.ts:333-334` — *"never read `state.language` directly so the default **and normalisation** stay in one place"* — and `chapter-splice.ts:115` + `:342` composes it correctly, using `bookStateLanguageOrNull(...) === null` as the **presence** test and `bookStateLanguage(state)` for the **value**. `voice-library.ts:1731` is the only site that takes the value off the honest reader and then compares it. Fix shape: presence from the honest reader, value from the normalising one — which also restores the pre-`84d407b5` behaviour on this axis without giving back the `'en'` default.

### 🟠 B — correctness (test): `84d407b5` neutered one of the three original tests; it now passes with the conjunct it names deleted

`server/src/routes/voice-library.test.ts:3321` — *"does not warn when the cloned voice has no languageCode (unknown language)"*.

`seedClonedWithLang('clone-undef')` leaves `entry.languageCode` undefined, and the book is written with **no** `language` argument, so `bookLanguageForClonedCheck` is `null`. Both conjuncts at `:1892-1893` are falsy, and either alone suppresses the warning. Delete `entry.languageCode &&` from `voice-library.ts:1892` entirely and this test stays green — the assertion is carried by the *book* conjunct, not the clone conjunct the title names. Rubric shape 3.

It was live before this PR's fix round: pre-`84d407b5` the book defaulted to `'en'`, so `bookLanguage` was truthy and only `entry.languageCode` could suppress. `84d407b5` gave `clone-ru-en` and `clone-en-en` an explicit `'en'` (`git show 84d407b5 -- server/src/routes/voice-library.test.ts`, hunks at 3293 and 3310) and left this one alone — so the fix for finding 1 silently removed the coverage this test existed to provide. Passing `'en'` as the 6th argument restores it.

### 🟡 C — finding 6 is only half fixed: the stale claim survives in the test file, and now names the removed variable

`server/src/routes/voice-library.test.ts:3263-3269`:

> `The comparison is CODE-vs-CODE (`entry.languageCode` against `bookLanguage`), never code-vs-sidecar-word — mutation 2 below pins that trap.`

Two problems. It names **`bookLanguage`**, the reader `84d407b5` removed from this path (the code compares against `bookLanguageForClonedCheck` at `:1894`). And "mutation 2 below pins that trap" is exactly the assertion `3e810dea` deleted from the production comment, which now reads *"an earlier version of this comment claimed mutation 2 pinned it, and it does not"* — and mutation 2's own note at `:3374-3383` says it pins the honest reader instead. Three comments in two files, two of them contradicting the third.

`3e810dea`'s commit message says *"The test-side half of this finding was already correct — fix agent 1 rewrote that comment in 84d407b5."* `git show 84d407b5 -- server/src/routes/voice-library.test.ts` shows it did not touch these lines.

### 🟡 D — the hoist comment at `voice-library.ts:1717-1720` was made false by `84d407b5`

> `#1998 — bookLanguage is needed by BOTH provenance branches below (…, cloned-voice languageCode mismatch at #1998), so it stays here, outside either if, rather than being re-derived inside each one.`

`grep -n '\bbookLanguage\b' server/src/routes/voice-library.ts` → the only use is `:1736`, inside `if (entry.provenance === 'designed')`. The cloned branch moved to `bookLanguageForClonedCheck` two commits later. The comment states a justification for the hoist that no longer exists, and a reader taking it at face value will look for a cloned-branch use that isn't there.

### 🟡 E — both release-notes files still describe the wizard copy `adeaf12c` removed

`adeaf12c` (the finding-4 fix, and HEAD) touched `src/modals/clone-voice-wizard.tsx` and nothing else. The two notes written one commit earlier still carry the rejected framing:

- `docs/release-notes-next.md:90` — *"shows a hint above `CloneCapturePanel` saying **to use the language the book will be read in**"*;
- `RELEASE_NOTES.md:49` — *"tells you, before you record or upload, **to use the language you want the book read in**."*

The removed copy was *"the language you want this book read in"*. The shipped copy is *"the language you want this voice to read — voices here can be cast in any book"*. The whole point of finding 4 was that the wizard has no book in scope and that implying a clone belongs to one book contradicts the library's book-independence; both notes still assert the book framing to the reader.

### 🟡 F — stale line-number citation in the new precedence test

`server/src/routes/voice-library.test.ts:3387` — *"the documented precedence on **line 1944**: `clonedAdvisory ?? languageWarning`"*. At HEAD that expression is at `voice-library.ts:**1963**`; the later fix commits moved it. (Line-number citations in this repo generally rot; this one is already wrong on the commit that introduces it.)

### 🟡 G — two duplications the fix round left behind

- `voice-library.test.ts:3459` and `:3464` are the identical assertion `expect(res.body.warning).not.toMatch(/cloned in Russian/)`, added in one commit under two different explanatory comments.
- `openapi.yaml` states the #1933-beats-#1998 precedence twice — once as the last sentence of the #1998 bullet, once as a standalone paragraph after the #1933 bullet — and `src/lib/api-types.ts` faithfully regenerates both. Two copies of one rule is two places for it to drift.

---

### On priority 1 — is suppressing on an unset book language correct, or merely safe-looking?

Correct, and for a stronger reason than "don't state a falsehood": **generation already hard-blocks such a book.** `routes/generation.ts:805` calls `requireBookStateLanguage(state)` and converts the throw into a `language-unset` SSE error frame, so a book with no language cannot render at all. Suppressing the assign-time warning therefore forfeits no signal the user could act on at that moment — they are forced to set a language before any audio exists.

The residual hole is real but is the pre-existing shape of the feature, not something the fix created: the warning is a point-in-time check at assign, and `PATCH /api/books/:id/state` can set the language afterwards, at which point nothing re-evaluates. For designed voices `clearMismatchedDesignedVoices` is the render-time backstop; cloned voices are deliberately exempt from it, so assign → set language → generate produces a ~0.23-identity render with no signal ever emitted. That is the v1 surface #1998 itself proposed ("tell the user at clone/assign time"), so I am not raising it as a finding — but it is not covered by any test or register row, and it is the case the suppression makes slightly wider.

### Verdict

Two 🟠 (one correctness regression introduced by finding 1's fix; one test that no longer tests what it names, also caused by that fix) and five 🟡, four of which are documentation the fix round made false. Findings A and B share one root cause — `84d407b5` changed two things while intending one — so they are best fixed together. The measurement citation, the openapi contract prose, the `pt` tests and the precedence test are all independently verified sound.

dudarenok-maker:
## PR review — pass 3 (head 977311eb, depth high)

Scope: the two new commits (`82aa9f27`, `977311eb`) against the whole branch diff — `server/src/routes/voice-library.{ts,test.ts}`, `openapi.yaml`, `src/lib/api-types.ts`, the three frontend files + their tests, both release-notes surfaces, `docs/features/275-clone-voice-language.md` · verified: `voice-library.test.ts` **191 passed** locally (13.8s), all 21 cloud checks green, tree clean at `977311eb` before and after.

Mutations were **not executed** this round — the environment blocked temp writes into the worktree, and I would not modify a tracked file to get around that. Claims below about *which assertion reddens* are analytic (traced through `normaliseBookLanguage` / `sidecarLanguageName` by hand) rather than observed, and I say so where it matters. Everything else is grep- or run-backed.

---

### Pass-2 dispositions

**A — unnormalised comparand · VERIFIED RESOLVED.** Split is correct: presence from `bookStateLanguageOrNull` (`voice-library.ts:1741` → `:1905`), value from `bookStateLanguage` (`:1721` → `:1906`, `:1911`). I attacked all three shapes the fix could have got wrong:

- *Registered but non-normalising book language* — impossible. Every registry key is a bare primary subtag (`tts/language-registry.ts:45-119`: `en ru es fr de zh ja`), and both persist paths gate on `isSupportedLanguage(normaliseBookLanguage(…))`.
- *Clone code failing `isSupportedLanguage` at write time* — yields `entry.languageCode: undefined` (`voice-library.ts:1219-1221`), so the first conjunct suppresses. Correct, and pinned by the `clone-undef` test.
- *`bookLanguage`'s `'en'` default reaching the comparison despite presence* — **one residual shape exists, and it is unreachable.** `primarySubtag` (`tts/language.ts:17-19`) returns `''` for any trimmed value starting with `-`, so a `state.language` of `'-ru'` gives a truthy `bookStateLanguageOrNull` (presence passes) while `bookStateLanguage` silently falls back to `'en'` — a `ru` clone would then be told "this book is English". Both writers normalise *and* registry-gate before persisting (`book-state.ts:851` `pickLanguage`, `import.ts:354-366`), so `'-ru'` cannot be written; it needs a hand-edited `state.json`. Recording it so a future round does not re-derive it, not raising it.

Also confirmed the *entry* side of the comparison, which the fix relied on without re-checking: `POST /design` does persist a raw client-supplied `languageCode` (`:217` → `:244`), but hardcodes `provenance: 'designed'` (`:241`), provenance is immutable (`:557-558`), and the only cloned writer normalises + registry-gates (`:1348`). So "`entry.languageCode` is always stored normalised" holds for every entry this guard can see.

**B — vacuous test · VERIFIED RESOLVED.** `:3329` now seeds `'en'`, so `entry.languageCode &&` is the sole suppressor.

**C — stale comparand claims · PARTIALLY RESOLVED.** Guard comment and describe header fixed; the two mutation recipes in the same file were not. See 🟡-1.

**D — `bookLanguage` used by both branches · CONFIRMED, reasoning holds.** It is now genuinely read by the designed branch (`:1746`) and the cloned branch (`:1906`, `:1911`). The `:1717` comment is accurate.

**E — release-notes reframing · VERIFIED RESOLVED.** Hint text (`clone-voice-wizard.tsx:104-108`) is book-independent, and the "only mount" premise holds — `CloneVoiceWizard`'s sole mount is `my-voices-section.tsx:98`, passing `onClose` alone; no `bookId` is in scope.

**F — rotting line citation · NOT FULLY RESOLVED.** The test-file citation was fixed; a second one in this same PR is now wrong, and `82aa9f27` is what staled it. See 🟡-2.

**G — openapi duplications · VERIFIED RESOLVED.** "one of three signals"; `grep -c "takes priority"` = **1** (`openapi.yaml:2654-2656`); the three negative fragments in the #1998 bullet are distinct and each true against `:1904`, `:1905`, `:1910-1912`. `src/lib/api-types.ts` is regenerated in exact sync — no schema shape changed, descriptions only.

---

### 🟡 Minor (cleanup) — 1. Both mutation recipes name a comparand the code no longer has; mutation 2 is now a strict no-op

`server/src/routes/voice-library.test.ts:3406-3407` and `:3410-3417`.

The string `entry.languageCode !== bookLanguageForClonedCheck` does not exist in `voice-library.ts` — `:1906` reads `entry.languageCode !== bookLanguage`. So:

- **Mutation 1** (`:3406-3407`) tells the reader to change `!== bookLanguageForClonedCheck` to `=== bookLanguageForClonedCheck`. Unperformable as written; the intent (invert the operator) is still clear and still valid.
- **Mutation 2** (`:3410-3417`) tells the reader to change `entry.languageCode !== bookLanguageForClonedCheck` to `entry.languageCode !== bookLanguage`. **That target is verbatim the current code.** Following the recipe changes nothing, and the regression tests it says "must go red" stay green.

Its stated rationale is stale too: what actually pins the unset-book behaviour is the `bookLanguageForClonedCheck &&` presence conjunct at `:1905`, not the comparand. `82aa9f27` updated the *production* comment to say exactly that (`:1899-1901`, "Mutation 2 pins … that the honest reader gates presence") but left the test file's own recipe describing the old comparand swap — so the two comments the same commit touched now contradict each other.

**Coverage is intact** — deleting `bookLanguageForClonedCheck &&` makes an unset book read `'en'`, so a `ru` clone warns and both unset tests fail on `expect(res.body.warning).toBeUndefined()` (no 500: `sidecarLanguageName` resolves `'ru'` and `'en'`). This is a documentation defect, not a hole. Worth fixing anyway because it is the *third* consecutive round to find a stale comparand/citation claim in this neighbourhood: pass 1 finding 6, pass 2 finding C, now this. Each round has fixed the instances it saw rather than sweeping the file — `git grep -n bookLanguageForClonedCheck` returns all six sites in two seconds.

### 🟡 Minor (cleanup) — 2. `release-notes-next.md` cites the designed-voice branch for the cloned-voice measurement

`docs/release-notes-next.md`, the #1998 assign entry:

> a cross-language source clip renders at ~0.23 ECAPA cosine identity against ~0.60 same-language (`voice-library.ts:1870-1873`; …)

At `adeaf12c` that citation was **correct**. `82aa9f27` inserted ~10 lines of split-readers comment at `:1722-1741`, pushing the measurement text down to **`voice-library.ts:1882-1883`**. Lines 1870-1873 are now the **designed-voice** branch — `designedManifest.language !== expectedSidecarLang` and the warning ending *"the audio will be **unintelligible**. Re-design the voice in …"*.

So a reader following the citation lands on the unintelligibility branch — precisely the conflation the same sentence warns against two clauses later ("The consequence is **loss of resemblance**, not unintelligibility … do not quote #1951's `avg_logprob` figures for this one"). `977311eb`, whose subject is "drop three duplications and **a rotting line citation**", fixed the other one and missed this. Cheapest fix is the one that commit already chose elsewhere: drop the line range, cite the expression or the issue's measurement table.

---

### House gates

- Paired tests real, not vacuous. The `profile-drawer.test.tsx:2321+` fixture string is byte-traceable to the template at `voice-library.ts:1914-1916` — not invented. The wizard test asserts present-then-absent across the phase boundary, so it fails if the hint is un-gated or removed.
- Release-notes pair present; both reframed per E and mutually consistent.
- `Closes #1998` plain, outside any code span; CI's issue-link check green.
- **No `docs/BACKLOG.md` row owed** — #1998 is labelled `bug`, and BACKLOG renders `type:feature` only. Its absence is not drift.
- On-box acceptance correctly declared not-owed — the measurement is cited to register row A1 / E-01 Run 3, not re-claimed as fresh.
- Derived artifact `src/lib/api-types.ts` regenerated. No `cast.json` lock surface changed; no new config knob.

### Verdict

**Nothing blocks merge.** No 🔴, no 🟠. Both findings are comment/prose defects with no runtime effect: the guard is correct, the entry and book sides of the comparison are both provably normalised, and the executable coverage genuinely pins the behaviours it names.

Both 🟡s are one-line edits and should be folded before merge — 🟡-1 in particular, because the recurrence pattern (three rounds, same neighbourhood, instances fixed rather than the file swept) is the part worth acting on, not the individual stale string.

dudarenok-maker:
## Pass 3 findings fixed — `df158924`

Both 🟡 fixed. No 🔴/🟠 in pass 3, so no fourth round; merging once CI is green on `df158924`.

**Finding 1 — swept, not patched.** Your observation that three consecutive rounds each fixed only the instance they saw is the actionable part, so I grepped every `bookLanguageForClonedCheck` mention and every line citation in the neighbourhood rather than editing the two you named.

The recipes are now described by **behaviour** instead of by a literal string to swap, which is what kept rotting. And you were right that mutation 2's instruction had become verbatim the shipped code — a recipe that is a strict no-op is the same "cannot fail" defect this PR keeps finding in tests, just in prose.

Three mutations are now documented, and **I executed all three against the real code before writing them down**:

| Mutation | Result |
|---|---|
| M1 — invert the comparison | match-case test red, plus 3 others |
| M2 — drop the presence conjunct | **both** unset-language tests red at `expect(res.body.warning).toBeUndefined()` |
| M3 — take the value from the presence reader | both `ru-RU` and `RU` tests red at the same assertion |

Production file restored byte-identical after each; `git diff` on it is empty. This also closes the gap your own caveat named — you flagged that pass 3's assertion-level claims were analytic because the environment blocked temp writes. They are now empirical.

**Finding 2 — confirmed and worse than a stale number.** `voice-library.ts:1870-1873` was correct at `adeaf12c`; `82aa9f27` inserted ~10 comment lines and moved it, so the citation now lands in the **designed** branch — whose warning ends "the audio will be unintelligible", the exact conflation that same sentence distinguishes itself from two clauses later. Replaced with #1998's title, which does not move. Also converted a "line 3150" reference to name its describe block; a line citation in this file has rotted twice now.

**On your finding-1 verification of A** — thank you for checking the side I relied on without re-verifying. I had not confirmed that `POST /design`'s raw client `languageCode` can never reach the cloned guard; `provenance: 'designed'` being hardcoded and immutable is the reason it cannot, and that belonged in my reasoning rather than in yours.

191 passed, tsc and eslint clean.

Issue #3040, opened 2026-09-06
[agent instructions][claude][verify] A29 re-verify the genuine install click-through after the sidecar-race fix, fold register


dudarenok-maker:
Verify step of the A29 install-path fix chain, tracking issue #3038.

## The gate — do not start early

**This issue stays `Standing` until #3039's PR is MERGED to `main`.** Promotion
to `Agent Todo` is a deliberate act by the operator or the coordinating
session, not something this issue's claimant does for itself. If you are
reading this as a claimed `Agent Todo` issue, first confirm independently
that #3039's PR actually merged (check its state, don't trust the promotion
alone) — if it hasn't, this is `AGENT BLOCKED`, not a run.

## Why this exists

Chain #2913 → #3020 → #3019 found a genuine, reproducible `WinError 5` on the
in-app Qwen3-TTS install click-through, root-caused to the installer racing
the live sidecar for the `onnxruntime` DLL (evidence:
`docs/testing/onbox-a29-results/step-2-genuine-install.md`; register row A29,
2026-09-06 entry in `docs/testing/onbox-acceptance-register.md`). #3039 fixes
that race. This issue re-runs the same genuine test against the fix and, if it
now passes clean, folds the register.

## What — repeat #3020's procedure exactly, against the fix

1. Create a fresh worktree off post-merge `main`:
   `node scripts/wt-new.mjs fix/ops-2192-a29-verify-fix --from main`, branch
   `fix/ops-2192-a29-verify-fix`.
2. Bootstrap this worktree's sidecar venv **without** installing Qwen3-TTS.
   Confirm it's genuinely absent via the sidecar's own live probe
   (`packageFault === 'missing'`) before proceeding — do not assume.
3. Start the app for real against this worktree's own isolated ports and
   fresh venv.
4. From a real browser (Playwright or claude-in-chrome, not curl): Model
   Manager → Qwen engine → click **Install**. Confirm the install genuinely
   runs (a near-instant return is evidence of a no-op, not a pass).
5. Watch for `WinError 5` / `Accès refusé` on any `.dll` under
   `site-packages/onnxruntime/capi/` during the install. This is the
   acceptance criterion: it must **not** occur now.
6. After install completes, load Kokoro and confirm `/health`'s
   `devices.kokoro` (the live ONNX session's actual provider, not
   `get_available_providers()`) — record it honestly whatever it reports;
   per the prior pass this is a separate, pre-existing gap
[...]

Issue #3039, opened 2026-09-06
[agent instructions][claude][task] Fix qwen install racing the live sidecar for the onnxruntime DLL (WinError 5)


dudarenok-maker:
Rework step of the A29 install-path fix chain, tracking issue #3038.

## Why this exists

Chain #2913 → #3020 → #3019 finally exercised A29's criterion 2 for real (step 1,
#3020, evidence `docs/testing/onbox-a29-results/step-2-genuine-install.md`) and
found a genuine, reproducible `WinError 5` / `Access is denied` on
`onnxruntime\capi\onnxruntime_providers_shared.dll` whenever the in-app **Install
Qwen3-TTS** action runs while the sidecar is alive. Root cause, isolated by a
controlled A/B in that evidence doc: `server/src/tts/qwen-install-bootstrap.ts`
→ `server/tts-sidecar/scripts/install-qwen3.mjs` never stops the sidecar before
running `pip install qwen-tts`; `qwen-tts` pulls a fresh `onnxruntime` wheel as a
transitive dependency, and the live sidecar holds that DLL memory-mapped, so the
DLL replace fails every time. The identical `pip` command succeeds once the
sidecar is stopped first.

## What

Fix `server/src/tts/qwen-install-bootstrap.ts` (and/or
`server/tts-sidecar/scripts/install-qwen3.mjs`) so the in-app install path no
longer races the live sidecar for that DLL. Two options the evidence doc
identified — pick whichever is more robust for this codebase, and document the
choice in the PR:

1. **Stop/pause the sidecar before running `pip`, restart it after** — mirrors
   the manual A/B that worked in the evidence doc (`taskkill` on the sidecar
   PID, then `pip install`, then respawn). Check how `sidecar-owner.ts`/
   `spawn-sidecar.ts` expose a stop/restart hook already used elsewhere
   (`POST /api/sidecar/restart` exists per the evidence doc — reuse it if it
   fits).
2. **Pin `qwen-tts`'s `onnxruntime` dependency against the already-installed
   `onnxruntime-gpu`** so pip never attempts the DLL swap while it's loaded —
   only viable if `qwen-tts`'s own dependency spec allows a constraint/extra
   that keeps it on the existing wheel.

Also worth fixing alongside (small, same file): the install job currently
surfaces only the last 2000 chars / 3 lines of the child process's stderr
(`qwen-install-bootstrap.ts`), which can truncate the actual `WinError 5` detail
[...]

Issue #3038, opened 2026-09-06
ops — A29: qwen install races the live sidecar for the onnxruntime DLL (WinError 5), install path fix needed


dudarenok-maker:
Register row **A29** ([#2192](https://github.com/dudarenok-maker/Castwright/issues/2192), plan [282](../features/282-ort-pip-consistency-marker.md)) chain #2913 → #3020 → #3019 finally exercised criterion 2 for real (step 1, #3020) — and it genuinely fails.

## What was found

`server/src/tts/qwen-install-bootstrap.ts` → `server/tts-sidecar/scripts/install-qwen3.mjs` never stops the sidecar process before running `pip install qwen-tts`. `qwen-tts` pulls a fresh `onnxruntime` wheel as a transitive dependency regardless of what the venv already has, and the live sidecar has that DLL memory-mapped — so the DLL replace fails with a genuine, reproducible `WinError 5` / `Access is denied` on `onnxruntime\capi\onnxruntime_providers_shared.dll` every time, in ~3 seconds. The identical `pip` command succeeds once the sidecar is stopped first.

This is #2192's exact reported scenario, now actually reproduced (not skipped by a pre-installed venv like the prior five attempts). It is distinct from — and not fixed by — the #2192/#2534 work already on `main` (that addressed the sidecar's own ORT provider selection at *startup*, not this installer's pip-vs-live-process DLL conflict).

A secondary, unrelated pre-existing gap was also confirmed honestly: `qwen-tts`'s pip resolution pulls the plain CPU `onnxruntime` wheel rather than `onnxruntime-gpu`, which is why Kokoro reports `cpu` afterward in this environment (same #2534-class CPU-fallback pattern) — not something this install attempt caused, and not in scope for the fix below.

Full evidence: `docs/testing/onbox-a29-results/step-2-genuine-install.md` (on branch `fix/ops-2192-a29-genuine-install-test`), and the 2026-09-06 register entry under A29 in `docs/testing/onbox-acceptance-register.md`.

## Why its own parent, and why the verify child WAITS

This is not a re-run, it's a code fix. Decomposed into a 2-step chain:

1. **Rework** — fix the installer's sidecar/pip race.
2. **Verify** — redo the genuine click-through (fresh venv, real browser, watch for `WinError 5`) against the fix once it's on `main`, then fold the register.

[...]

Issue #3020, opened 2026-09-06
[agent instructions][claude][task] A29 genuine-install-test step 1 - real click-through against a venv without Qwen3 preinstalled


dudarenok-maker:
Step 1 of the A29 genuine-install-test chain, campaign #2435, on-box acceptance
register.

## Why this exists

PR #3015 found the earlier A29 "discharge" (from chain #2913) rested on an
untested criterion: that worktree's sidecar venv already had Qwen3-TTS
installed from its own bootstrap, so clicking Install in the app ran a
no-op `pip install` — nothing was written under
`site-packages/onnxruntime/capi/`, so the actual #2192 scenario
(`WinError 5`/`Accès refusé` during a real package install) never fired.
This step exists to close that gap for real.

## What

1. **Bootstrap this worktree's sidecar venv WITHOUT installing Qwen3-TTS.**
   Read `server/tts-sidecar/scripts/install-qwen3.mjs` and whatever base
   bootstrap script this repo uses (`README`/`docs/features` for the sidecar
   setup) to install only the baseline deps a healthy app needs to boot —
   deliberately skip the Qwen3 package. Confirm directly (list
   `site-packages/`, or run the install script's own "already installed"
   check) that Qwen3-TTS is genuinely absent before proceeding — do not
   assume, verify.
2. **Start the app for real** against this worktree's own ports
   (`VITE=5373`, `API=8280`, `TTS=9200` — already written to `.env.local`/
   `server/.env` by `wt-new.mjs`) and this fresh venv (`LOCAL_TTS_PORT=9200`
   in `server/.env`, already set).
3. **From a real browser** (Chrome via claude-in-chrome or Playwright,
   real UI, not curl): Model Manager → Qwen engine → click **Install**.
   Confirm the install genuinely runs for real time (downloading/writing
   packages) — a near-instant return is itself evidence of a no-op and
   means this run has NOT closed the gap; if that happens, stop and report
   it rather than declaring success.
4. **Watch for `WinError 5` / `Accès refusé`** on any `.dll` under
   `site-packages/onnxruntime/capi/` during the install — this is the
   actual #2192 repro condition. Capture console/network evidence of the
   install's real duration and outcome (success or the exact failure).
5. **After install completes, load Kokoro** and confirm its `/health`
   response reports `devices.kokoro: "cuda"` from the live ONNX session
   (`server/tts-sidecar/main.py`'s health route — read it to find the exact
[...]

PR #3041, opened 2026-09-06
docs(ops): A29 step 2 - genuine WinError 5 found in qwen install, still owed

dudarenok-maker:
## Summary
- Chain #2913 → #3020 → #3019 finally exercised register row **A29**'s criterion 2 for real (a venv confirmed genuinely absent Qwen3-TTS before the in-app **Install** click, across six total attempts on this row).
- **Genuine, reproducible `WinError 5` / `Access is denied` found**: clicking Install in Model Manager (real Playwright browser click, not curl) triggers a real `pip install qwen-tts` job that fails on `onnxruntime\capi\onnxruntime_providers_shared.dll` every time the sidecar is alive. Root-caused by a controlled A/B: the installer (`server/src/tts/qwen-install-bootstrap.ts` → `install-qwen3.mjs`) never stops the sidecar before running `pip`, and `qwen-tts` pulls a fresh `onnxruntime` wheel that collides with the DLL the live sidecar has memory-mapped. Identical `pip` command succeeds once the sidecar is stopped first.
- **This is a real bug, not discharged.** Not fixed by the #2192/#2534 work already on `main` (that addressed sidecar-startup ORT provider selection, a different code path).
- Kokoro-afterward check read honestly from the live `/health` response (`devices.kokoro`, not `get_available_providers()`): `"cpu"` throughout — a separate, pre-existing gap (plain `onnxruntime` vs `onnxruntime-gpu` resolution for `qwen-tts`), not caused by this pass.
- Register row A29 narrowed with this finding and left **STILL OWED** (not removed — a genuine defect is not discharged).
- Rework filed as its own chain rather than retried silently: tracking parent #3038, rework task #3039 (fix the installer's sidecar/pip race), gated verify child #3040 (re-run this exact test against the fix; stays `Standing` until #3039 merges).

**Docs-only** — confined to `docs/testing/onbox-a29-results/step-2-genuine-install.md` (new evidence doc) and `docs/testing/onbox-acceptance-register.md` (narrowed A29 row). No source files touched. Claiming the `pr-review-gate` exemption on that basis.

## Test plan
- `npm run check:onbox-register` — passes, register/live-view agree (no live-view diff since no `###` heading changed).
- Evidence doc `docs/testing/onbox-a29-results/step-2-genuine-install.md` records the full repro: absence confirmation (twice), the real Playwright install click and its network trace, the WinError 5 root-cause A/B, the Kokoro health check, and worktree restoration/box-safety.

Refs #2192, #2435, #2913, #3019, #3020, #3038, #3039, #3040


dudarenok-maker:
Docs-only file set (`docs/testing/onbox-a29-results/step-2-genuine-install.md`, `docs/testing/onbox-acceptance-register.md`) — exempt from the mandatory pr-review-gate pass per CLAUDE.md's doc-only file-set test.

Issue #2914, opened 2026-09-04
[agent instructions][claude][verify] A29 step 2 - verify the retry, fold the register, open the PR


dudarenok-maker:
Step 2 of the A29 retry chain (#2913), on-box acceptance register campaign
(#2435).

## What

Verify step 1's run, fold the outcome into the register, and open the PR.

## Checks

1. Step 1's evidence shows the app bound **this worktree's own ports**
   (VITE 5253 / API 8160 / TTS 9080), confirmed from the server's own startup
   log — not adopted the box's shared `:9000`.
2. The install and Kokoro-afterward check were run against **this worktree's
   own bootstrapped venv**, never the primary checkout's live one, and never
   a junction to it.
3. Real browser screenshots or console/network evidence exist for the actual
   click-through — not a description of what should happen.
4. The verdict is honest: if collision or contention occurred anyway despite
   the distinct port, that is recorded as a **regression against #2632**, not
   silently retried or hidden.

## Fold

If DISCHARGED: remove row **A29** from
`docs/testing/onbox-acceptance-register.md` (2026-08-22 ruling — removed, not
annotated), recompute the "At a glance" table by counting `###` headings,
update the live-view HTML by hand. If STILL OWED or a new defect surfaced,
narrow the row or leave it, and file the defect separately per CLAUDE.md's
incidental-findings rule.

## FAIL ordering

Rework `[claude][task]` child → fresh `[claude][verify]` child → link both to
#2913 → then close this issue.

## On PASS

Open the PR: base `main`, head `docs/docs-2913-a29-retry`. `Closes #2913`,
`Refs #2435`. If the fold is a bare register removal with no code change,
this is a docs-only file set and exempt from `pr-review-gate` — state that
explicitly in the PR body.

Never close parent #2913 yourself.

Worktree `C:\Claude\Projects\wt-2913-a29-retry`, branch
`docs/docs-2913-a29-retry`.

Refs #2913, #2435

dudarenok-maker:
AGENT CLAIMED
Agent: claude

Taking [claude][verify] step 2 of the A29 retry chain. Will verify step 1's evidence, fold the register/PR outcome, and open the PR per the issue's PASS/FAIL instructions.

dudarenok-maker:
AGENT CLAIMED
Agent: claude

Taking this via the scheduler's claim fast path (kind: todo). Will verify step 1's
[...]

PR #3042, opened 2026-09-06
docs(docs): close #3011 as doctrine after three refuted designs

### LINKED ISSUE #3011: Parking a ticket at Agent Needs Input does not stop the resumable-claim fast path — real two-writer worktree collision
## What happened

**2026-09-05.** A lane parked issue #2914 at **`Agent Needs Input`** specifically to stop the queue
touching it while hand-committing in `wt-2913-a29-retry`. A fresh tick **resumed it anyway** —
posting `AGENT RESUMED` and starting its own `claude.exe`/ringer process **into the same worktree
as the in-flight commit.**

That is a genuine **two-writer** situation, not a hypothetical. It was caught and the duplicate
killed only because someone happened to be watching; the duplicate had not yet written, the
hand-commit had. A few seconds either way and the outcome is an interleaved commit to reconstruct.

Found by the `Onbox Mega Wave — 11 Chains` lane while working #2910/#2912/#2914.

## Mechanism

**The resumable-claim fast path reads the ticket's comment/claim history, not the board `Status`
field.** So a board state is *authoritative* to one code path and *invisible* to another, and
nothing at the call site tells you which one you are relying on.

The practical consequence is that **"I parked it" is not a guarantee.** An operator or lane that
sets `Agent Needs Input` to reserve a tree has taken an action that reads as protective and is
not.

## Why this is worth a ticket rather than a patch

**1. It is the failure mode this repo has hit three separate times.** Two writers in one worktree
(2026-08-29, 2026-08-30, 2026-09-04/05) produced bundled commits, contention flakes, and — worst —
a subagent concluding the contention it caused itself was an external blocker and reaching for
`--no-verify`. CLAUDE.md's "One writer per worktree" section exists because of those incidents.

**2. It is independent of the commit gate, and survives #2999.** Every other concurrency problem
chased on 2026-09-05 arrived through the git hooks — slow batteries, contention, deadlock — and
#2999 removed that source. **This route has nothing to do with hooks.** Merging every worktree onto
`main` does not touch it. A quiet box is not evidence this class is closed.

**3. The invariant being broken is not the one the code guards.** The resource that cannot take two
writers is the **worktree**, not the ticket. A ticket and a worktree are not 1:1 — they diverge as
soon as a chain reassigns or a claim is reset mid-flight, which **both #2910 and #2914 did on the
same night**. Exclusivity asserted on the ticket therefore does not protect the tree at all.

## The decision owed

Three defensible options; this is filed rather than fixed because choosing between them is a design
call, not a judgement call:

1. **Make resumable-claim consult board `Status` as a gate.** Smallest change. Leaves two sources of
   truth in place and only teaches one more path to read the other.
2. **Make `Agent Needs Input` write something the claim-history path already reads.** Also small.
   Keeps the fix inside the mechanism that is currently authoritative, but makes the board state's
   meaning depend on a side effect.
3. **Assert exclusivity on the WORKTREE rather than the ticket.** *(Preferred by both the finder and
   the filer, with 1 and 2 explicitly left on the table.)* Fixes the noun rather than one of its
   symptoms, and generalises past this bug — it holds when a chain reassigns, when a claim resets
   mid-flight, and for any future path that can start work in a tree. Larger change, and it needs a
   decision about what holds the lock and how it is released when a lane dies without cleanup —
   note that orphaned processes outliving their agent were separately observed the same night.

## Acceptance

1. A recorded decision among the three, with reasoning.
2. Parking a ticket in a "do not touch" board state reliably prevents **every** path from starting
   work on it — or, under option 3, a second writer cannot enter a worktree that already has one,
   whatever the ticket says.
3. A regression test that reproduces the reported shape: a claim parked in the blocking state, a
   tick fired, and no second process started.
4. If option 3 is taken, an answer for stale-lock release when a lane exits without cleanup.

## Key files

- `~/.claude/skills/open-agent-engine/SKILL.md` — claim/resume/receipt protocol
- `~/.claude/skills/oe-route/SKILL.md` — lane assignment and reroute
- the OE tick/runtime implementing the resumable-claim fast path
- `CLAUDE.md` → "One writer per worktree — commit/push is single-threaded"

## Benefit

*(technical / architectural)* Closes a live, reproduced route to the exact worktree-corruption
class this repo has hit three times, and one that no amount of commit-gate work can reach.

dudarenok-maker:
## Summary

Closes #3011 as **doctrine, not mechanism**. Three successive designs were investigated and all three were refuted — two by adversarial review, and the third (a defect *raised by* a review) by verification. **No code change anywhere.**

### What actually happened

A lane parked #2914 at `Agent Needs Input` to hold `wt-2913-a29-retry` while a human hand-committed in it. A tick launched a second lane into the same tree.

| Time (UTC) | Event |
|---|---|
| 22:55:15 | `AGENT BLOCKED` |
| 22:59:10 | board status → `Agent Needs Input` |
| **22:59:11** | **human comment: "Parking at Agent Needs Input temporarily…"** |
| 23:00:44 | `claimable: #2914[claude](answered)` |
| **23:08:44** | lane launched |
| 23:14:37 | `AGENT RESUMED` |

That branch's rule is *newest comment is not an `AGENT_*` receipt ⇒ a human replied ⇒ claimable*. The parking comment is human prose, so **announcing the park is what un-parked it**, 93 seconds later. The overlap was ~6 minutes, not the "few seconds" the issue describes.

**The issue's own stated mechanism is wrong.** It says the resumable-claim path ignores board `Status`; `oe-tick.ps1` branches on status explicitly and produces `resumable-claim` only under `Agent Working`. Its remedies 1 and 2 target a defect that does not exist.

### The three refutations

**A lock file in `.husky/pre-commit`** — would have *inverted* the incident. The collision was a lane against a **human**, who carries no lane identity, so the hook refuses the human while the intruder writes. Also: nothing would acquire it; `.husky/*` is tracked so each worktree keeps old hooks until it merges `main`, and tool-created trees run no hook silently; and the mandated commit path (a detached helper) never puts the worktree on a command line, so a process probe cannot see it.

**The gate refusing a second lane on an occupied tree** — reproduces the same blind spot (its predicate is *lane* liveness; the incident had a human), and measurement killed it independently: the name-convention resolver covers **6 of 13** sequences, and the failures span **76% of open tickets**; the tick records no worktree for a running lane; `Dirty`/`Ahead` reports **9 of 15** trees dirty with nothing live, and cannot attribute what it sees — the log's own "work left behind by lane claude" line describes files the *human* wrote.

**An `Agent Reserved` board state** — unnecessary. The board already carries `Parked`, `Agent Waiting` and `Waiting/Blocked`, all already unclaimable.

**And the "one real defect" was not one.** Review pass 2 reported `Select-Lanes`' `$seenSeq` as an unseeded cap letting a top-up start two lanes in one worktree. That fix was implemented, then **discarded**: the top-up call site already excludes in-flight sequences, deliberately, with a comment saying so. Both lane-start paths are covered. Recorded as method: *a review's claim is evidence, not a verdict.*

### What ships

Doctrine in CLAUDE.md's existing "One writer per worktree" section: never hand-commit in a tree with an open agent ticket; park at `Parked`, never `Agent Needs Input`; announcing a park does not park it.

Plus a **correction to that section's own process probe** — it documents `CommandLine -like "*<worktree-path>*"`, which matches the shell running the search. It reported five phantom occupants on a completely idle tree today. Now says to exclude your own ancestry *and its children*.

Stated as a limit rather than papered over: **nothing can see a non-lane writer** — a human, an interactive session, a dispatched fix agent, an orphaned battery — and that is what caused this incident.

## Test plan

Docs and process only; no runtime surface, no tests to add.

The substance was verified rather than asserted: every code claim was checked against `oe-tick.ps1`, `oe-doctor.ps1` and `runner/prompt.md`; the incident timeline against `~/.open-engine/tick.log`; the resolver and `Dirty`/`Ahead` figures measured against the live queue and the 15 non-primary worktrees; and the discarded fix confirmed unnecessary by enumerating every `Start-Lanes` and `Invoke-Gate` call site.

**Review gate: exempt** — the changed file set is entirely root-level `*.md` plus `docs/**` (`CLAUDE.md`, one spec), the same file-set test as CONTRIBUTING.md's doc-only fast-path. Noting deliberately that this *does* touch instructions agents follow literally, which the skill flags as the subclass not to wave through on substance — so for the record, the substance carried **two Premium-tier adversarial passes**, and their findings are what reversed the outcome from "ship a mechanism" to "ship a rule."

Closes #3011

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1271,6 +1271,32 @@ Working practice below; this holds even under contention).
   dispatch tool reported the agent "completed" — a completed agent
   notification can still have live background children of its own still
   running commits/pushes.
+- **That process probe matches ITSELF.** The pattern you search for is in the
+  command line of the shell running the search, so a naive
+  `CommandLine -like "*<worktree-path>*"` reports phantom occupants on a
+  completely idle tree (observed 2026-09-06: five "live processes", all of them
+  the probe's own wrappers). Exclude your own **ancestry and that ancestry's
+  children** — excluding just `$PID` is not enough, because the harness spawns a
+  wrapper shell per call that is not on your ancestor chain but is a child of
+  something that is. A probe that always fires is indistinguishable from one
+  that works, so confirm it reports zero on a tree you know is idle before
+  trusting a zero on one you are about to write to.
+
+**Holding a tree against the Open Engine queue** (#3011, design of record:
+[docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md](docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md)):
+
+- **Never hand-commit in a worktree that has an open agent ticket.** Commit from
+  the primary checkout, or from a tree no ticket points at.
+- **To hold one anyway, move the ticket to `Parked`** — an existing board state
+  the gate never makes claimable. **Never `Agent Needs Input`**, which means
+  *waiting on a person*: **any comment on it is a resume trigger, including the
+  comment saying not to resume.** That is not hypothetical — it is exactly how
+  a second lane was launched into a tree a human was committing in.
+- **Announcing a park does not park it.** Change the state first.
+- **The tooling cannot see a non-lane writer.** A human, an interactive session,
+  a dispatched fix agent, and an orphaned battery are all invisible to the
+  queue's own exclusivity checks. The rules above are the protection; there is
+  no mechanism behind them.
 
 ### Planning agents
 
--- a/docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md
+++ b/docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md
@@ -1,256 +1,140 @@
-# Worktree exclusivity — design
+# Worktree exclusivity — findings and doctrine
 
 - **Date:** 2026-09-06
 - **Issue:** [#3011](https://github.com/dudarenok-maker/Castwright/issues/3011)
-- **Status:** draft, revised after one adversarial review pass
-- **Implementation lands in:** [`open-engine`](https://github.com/dudarenok-maker/open-engine) — this repo carries the design of record only
+- **Status:** closed as **doctrine, not mechanism**
+- **Outcome:** an operator rule. **No code change anywhere** — three designs and one reported defect were investigated and all four were refuted.
 
-## Problem
+Two designs were written for this and **both were refuted by adversarial review**; a third
+finding, raised by a review itself, was refuted on verification. This document records what was
+learned, so none of them is proposed again, and states the rule that replaces them.
 
-Two writers in one worktree is the failure CLAUDE.md records this repo hitting **three separate
-times** (2026-08-29, 2026-08-30, 2026-09-04/05). It produces bundled commits, contention flakes,
-and — worst — a subagent concluding the contention it caused itself is an external blocker and
-reaching for `--no-verify`.
-
-### What actually happened on 2026-09-05
+## What actually happened on 2026-09-05
 
 A lane parked #2914 at `Agent Needs Input` to hold `wt-2913-a29-retry` while a human hand-committed
-in it. A fresh tick launched a second lane into the same tree.
-
-The sequence, from `~/.open-engine/tick.log` and the issue's own comment log — **the log is the
-evidence; the comment text alone is only an inference**:
+in it. A tick launched a second lane into the same tree.
 
 | Time (UTC) | Event | Source |
 |---|---|---|
-| 22:55:15 | `AGENT BLOCKED` posted | issue #2914 |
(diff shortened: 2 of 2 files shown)


dudarenok-maker:

Changed files are entirely root-level `*.md` and `docs/**` — `CLAUDE.md` and one spec under `docs/superpowers/specs/` — which is the same file-set test as CONTRIBUTING.md's doc-only fast-path. No review pass runs.

Recorded explicitly because this PR *does* touch instructions agents follow literally, the subclass the skill says is reviewed rather than waved through on substance. That standard was met by a different route: the content is the output of **two Premium-tier adversarial passes**, whose findings reversed the outcome twice — from "ship a hook-enforced lock" to "ship a gate-side check" to "ship no mechanism at all" — and a third finding raised by one of those passes was implemented and then discarded on verification.

PR #3045, opened 2026-09-06
docs(docs): record that EngineLaneCap 3 is settled, not a confounder to remove

dudarenok-maker:
## Summary

The worktree-exclusivity design doc described `EngineLaneCap = 3` as an experiment and *"a one-line revert"*. That is no longer true — [open-engine PR #66](https://github.com/dudarenok-maker/open-engine/pull/66) made 3 the settled value on the operator's decision.

**This changes what the section means, not just its wording.** It previously framed the cap as something to undo *before* measuring whether the two-writer class recurs. It is now the condition being measured *under* — so a future recurrence is evidence about the system as actually configured, rather than an artefact to be discounted.

Also records why the original caution was retired rather than carried forward: it watched for pre-commit hooks stalling under simultaneous vitest batteries, and ops-2997 removed batteries from hooks entirely, so concurrent lanes cannot collide that way at all. Plus the two errors corrected alongside it — the cap is **global**, not claude-scoped as its commit subject implied, and `MaxLanes` (4) still bounds total concurrency regardless.

## Test plan

Docs only; no runtime surface, no tests owed.

**Review gate: exempt** — changed file set is a single file under `docs/**`, the same file-set test as CONTRIBUTING.md's doc-only fast-path.

Refs #2997

### DIFF
--- a/docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md
+++ b/docs/superpowers/specs/2026-09-06-worktree-exclusivity-design.md
@@ -131,10 +131,21 @@ If this class recurs *between lanes*, that is new information — the two lane-s
 guarded, so a lane-versus-lane collision would mean one of those guards is wrong — and this
 document is wrong. If it recurs between a lane and a human, it is the known limit above.
 
-## Confounder, noted and not addressed
+## Confounder, and how it was settled
 
 `EngineLaneCap` went 1 → 3 in `open-engine` `f741933` at 2026-09-05T20:49Z, ~2h before the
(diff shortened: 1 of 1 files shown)


dudarenok-maker:

Changed file set is a single file under `docs/**`, which is the same file-set test as CONTRIBUTING.md's doc-only fast-path. No review pass runs.

Recording it because the change is a correction of meaning rather than wording: the section previously framed `EngineLaneCap = 3` as a confounder to remove before measuring, and it is now the condition being measured under.

Issue #3044, opened 2026-09-06
Dispatched fix agents write to the primary checkout instead of their assigned worktree


dudarenok-maker:
## What

**Dispatched fix agents write to the wrong absolute path** — the primary checkout, or another lane's tree — instead of the isolated worktree their brief names. **Four occurrences on record**, across at least two independent sessions.

The latest (2026-09-06) is the clearest, because it separates the two failure modes cleanly:

- Uncommitted edits to `RELEASE_NOTES.md` and `docs/release-notes-next.md` appeared in the **primary checkout, on `main`**.
- The content was **byte-for-byte correct** — near-identical to entries already properly committed on PR #3016's branch (`fix/server-2934-a36-audition-band`), plus coverage duplicating PR #3033's clone-language warning.
- The owning session confirmed it never wrote to the primary: every one of its own calls there was read-only, and all six of its dispatched fix agents were briefed with explicit absolute worktree paths.

**So the agent had the right content and the wrong destination. This is a path bug, not a content bug**, and that distinction is what makes it dangerous: the output looks correct on inspection, so nothing about the diff itself signals that it landed in the wrong repository state.

## Why it matters

Had those edits been committed, they would have landed on `main` **without a PR** — which `main`'s required status checks and the branching workflow both exist to prevent — and then collided with two open PRs carrying the same entries.

One earlier occurrence corrupted the primary checkout outright.

This also interacts badly with the "one writer per worktree" invariant: a stray agent writing into the primary is a second writer in a tree that another session is reading, and neither side has any signal that it happened.

## Why briefs alone do not fix it

Every occurrence involved a brief that **already stated the absolute worktree path**, in some cases with an explicit "do NOT create a worktree, do NOT work anywhere else" instruction. Strengthening the prose has been tried and has not held.

The failure is not the agent misunderstanding where to work — it is the agent resolving a relative path, or a remembered path, against the wrong root at some point mid-task.

## The decision owed

This is filed rather than fixed because the remedy has more than one defensible shape:

1. **Verify after every dispatch round.** The dispatching session runs `git status --porcelain` in the primary (and in any tree it does not own) after each fix-agent returns, and treats any unexpected change as a failed dispatch. Cheap, purely additive, catches every occurrence — but it is detection, not prevention, and it depends on the coordinator remembering.
[...]

PR #3046, opened 2026-09-06
docs(ops): verify where a dispatched agent actually wrote

dudarenok-maker:
## Summary

Records the **option 1** decision on #3044 — the dispatching session verifies the primary checkout's tree state around every dispatch round — as doctrine under CLAUDE.md's "One writer per worktree" section, which is the issue's own acceptance item 3 ("documented where dispatchers will actually read it").

Four bullets, each carrying a fact the failures established rather than general advice:

- **before/after `git status --porcelain`**, so the signal is a delta rather than a judgement about what looks like this branch's work. Cites the identical before/after tree check `pr-review-gate` already runs around a reviewer pass.
- **every tree the dispatch does not own**, not only the primary — the primary is the likeliest wrong root, not a special one.
- **the diff cannot reveal it.** In the clearest occurrence the content was byte-for-byte correct, near-identical to entries already properly committed on the branch that owned them. Path bug, not content bug.
- **the brief does not hold.** Every occurrence already named the absolute worktree path, some with an explicit "do NOT work anywhere else".

## Why `Refs`, not `Closes`

Option 1 is detection, and the issue's acceptance item 2 asks for a *check* that would have caught the 2026-09-06 occurrence. A probe run alongside this change established that **option 2 is buildable** — Claude Code's `PreToolUse` hooks fire for `Write`/`Edit`/`Bash`, receive `tool_input.file_path` (or the raw `command`) plus the session `cwd`, and can refuse a call via exit 2 or `permissionDecision: "deny"`. Notably the hook can live in **subagent frontmatter**, i.e. `.claude/agents/*.md` — the one part of `.claude/` this repo does not gitignore — so it can be version-controlled and reach a fresh clone, which a `.claude/settings.json` hook could not.

One unknown remains before option 2 ships: whether a frontmatter-declared hook actually fires for an Agent-tool dispatch is undocumented and needs an empirical test. Option 2 supersedes this as the gate when it lands; this stays as the backstop.

## Test plan

Documentation-only — no runtime surface, no test owed. Verified the rendered anchor
`.claude/skills/pr-review-gate/SKILL.md#the-tree-check` resolves against the existing heading.

Before-shipping steps not applicable and why: **1** (no behaviour change to plan), **2** (no code), **3** (no hardware-provable behaviour), **4** (no plan added or moved), **5** (process-only, no shippable delta), **7** (no runtime surface for the branch-scoped battery).

Refs #3044

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1298,6 +1298,30 @@ Working practice below; this holds even under contention).
   queue's own exclusivity checks. The rules above are the protection; there is
   no mechanism behind them.
 
+**Verify where a dispatched agent actually wrote**
+([#3044](https://github.com/dudarenok-maker/Castwright/issues/3044)):
+
+- **Capture the primary checkout's `git status --porcelain` before a dispatch
+  round and again after each agent returns.** Any entry that is not yours is a
+  **failed dispatch** — revert it and re-dispatch; do not adopt it. This is the
+  same before/after tree check the [`pr-review-gate`
+  skill](.claude/skills/pr-review-gate/SKILL.md#the-tree-check) already runs
+  around a reviewer pass, applied to the tree the agent was told *not* to touch.
+- **Check every tree the dispatch does not own**, not just the primary. The
+  observed writes went to the primary because that is the path most likely to be
+  resolved against by mistake, but nothing about the mechanism is specific to it.
+- **Inspecting the diff cannot tell you this happened.** Four occurrences are on
+  record across at least two sessions, and in the clearest one the content was
+  *byte-for-byte correct* — near-identical to entries already properly committed
+  on the branch that owned them. It is a path bug, not a content bug, so the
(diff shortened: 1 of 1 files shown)


dudarenok-maker:

Changed-file set is `CLAUDE.md` alone — a root-level `*.md`, so the PR meets the docs-only file-set test [`pr-review-gate`](.claude/skills/pr-review-gate/SKILL.md#exemption) shares with CONTRIBUTING.md's doc-only CI fast-path. No review pass runs.

Noting the tension deliberately rather than silently: this is instruction content agents follow literally, the subclass the skill's review-depth section says is reviewed rather than waved through. The exemption is by **file set** and is total, so it applies — but the assertions in the block (four occurrences, two sessions, byte-for-byte identical content, two colliding PRs) are all sourced verbatim from #3044's body rather than re-derived, which is where that subclass usually goes wrong.

Issue #3050, opened 2026-09-06
ops-74 — commit gate Part 6: scope-triggered local acceptance for the sidecar


dudarenok-maker:
## What

A required check that fails a PR touching `server/tts-sidecar/**` unless the body records the local `test:sidecar` run (command, date, outcome) or links an [on-box acceptance register](docs/testing/onbox-acceptance-register.md) row. Mirrors `.github/workflows/pr-issue-link.yml`. Design of record: **Part 6** of [`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md`](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md). Parent #2997. Independent of #3047/#3048.

## Why it must be mechanised, not a checklist item

38 `pytest.importorskip("torch")` tests across 14 files run only on this box. Part 1 kept an **automatic** `test:sidecar` push trigger that exercises them; anything replacing or supplementing it that is unenforced is strictly weaker than what it supplements. An unenforced checklist item would be waived by habit within a week.

## Trigger path: `server/tts-sidecar/**` only

An earlier draft also listed `server/src/tts/**`, `server/src/analyzer/**`, `server/src/gpu/**` — 177 TypeScript test files Ubuntu CI already covers. **That requirement would be waived by habit within a week and must not be added back.**

## Acceptance

1. A PR touching `server/tts-sidecar/**` with neither a recorded run nor a register-row link fails the check.
2. A PR touching `server/tts-sidecar/**` that records either one passes.
3. A PR touching nothing under `server/tts-sidecar/**` passes without needing anything in its body.
4. The recorded-run format is documented where a PR author will read it, and is specific enough to be checkable (command, date, outcome) rather than free prose.
5. Wired into `main`'s required status checks alongside `verify.yml` and `pr-issue-link.yml`.

Refs #2997

## Benefit

*(technical)* Keeps the only automatic executor of the torch-gated sidecar suite from silently going unexercised, with enforcement rather than a convention.

dudarenok-maker:
**AC5 done — closing.** `Verify sidecar acceptance is recorded` is now a required status check on `main`, appended to ruleset `17654264` alongside `npm run verify` and `Verify PR body links a GitHub issue`. Verified by re-reading the ruleset after the write; the ruleset's other two rules (`deletion`, `non_fast_forward`) and its `refs/heads/main` condition are unchanged.

[...]

Issue #2083, opened 2026-08-03
srv-83 — sweep the Promise.all-of-imports vs async vi.mock race out of the remaining server specs


dudarenok-maker:
## What

~19 server spec files still combine a hoisted **async-factory** `vi.mock(path, async (importOriginal) => …)` with a `beforeAll`/`beforeEach` that loads modules via `Promise.all` of dynamic `import()`s. That combination is a real, measured race: `Promise.all` starts all module evaluations in the same tick, which can begin **before vitest has finished resolving the async mock factory**. Whichever import wins the race binds the **real** export, and the spec then silently exercises unmocked code.

The fix is mechanical — sequential `await import(...)` statements in the setup hook, same binding names and order. No production change.

## Evidence this is real, not theoretical

Found on the cast-lock sweep (#1981). In `server/src/routes/voices.test.ts` it was **live**: 4 failures in 6 runs at `--retry=0`, presenting as `expected false to be true` because the module under test called the real `readJson` instead of the gated mock. Converting that hook to sequential `await`s took it to 8/8 clean. Restoring the `Promise.all` during review reproduced the flake at 4/6.

The cast-lock branch fixed 13 files (the 6 it touched plus 6 directly adjacent in the cast/voice-design family). The files below were confirmed to have the same shape but are **unrelated subject matter** — generation, chapter-splice variants, export, QA machinery — so widening that PR's diff into them was the wrong call. They are filed here instead.

Measured **latent, not live** across the ones sampled (0 failures in 14 runs) — they have the hazard but do not currently trip it. That is a property of current timing, not a guarantee; the same shape was live one file over.

## Files

Verified by reading the setup hook, not by a loose grep (a bare "Promise.all + import(" grep matches ~60 files, almost all false positives from unrelated live-request racing or per-test lazy imports). Spot-checked at file-and-hook level for `generation.test.ts` (`beforeAll` destructuring 6 concurrent imports), `export-disk-guard.test.ts`, `qa-report.test.ts`, `annotate-emotion.test.ts`.

- `routes/annotate-emotion.test.ts`
- `routes/chapter-qa-repair-spk.test.ts`
- `routes/chapter-splice-fresh-verdict.test.ts`
- `routes/chapter-splice-gate-off-preserves-verdict.test.ts`
- `routes/chapter-splice-rerecord-qa.test.ts`
- `routes/export-disk-guard.test.ts`
- `routes/generation.test.ts`
[...]

PR #3053, opened 2026-09-06
ci(ops,scripts): add a Windows PR-CI leg and a sidecar acceptance gate

### LINKED ISSUE #3049: ops-73 — commit gate Part 5: a Windows leg in PR-time CI
## What

A `windows-latest` job in `verify.yml` running `test` and `test:server`, with its scope conditions on **steps**. Design of record: **Part 5** of [`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md`](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md). Parent #2997. Independent of #3047/#3048.

Reuses `.github/actions/setup` (already used on `windows-latest` by `cross-os.yml`) and `.github/actions/install-ffmpeg-windows` for the server leg. Every step pins `shell: bash` — `verify.yml` sets no `defaults: run: shell:` and relies on ubuntu's default. Sets `timeout-minutes` like every other job in the file.

## Three wiring traps, all verified against the workflow

- **The job must join the aggregator's `needs:` list.** The only required context is the aggregator job `name: npm run verify`; a job outside that list is decorative. The "Check leg results" step iterates `jq -r 'keys[]'` over all of `needs`, so no second edit is required, and the required-check name does not change.
- **Scope conditions go on steps, never the job.** The aggregator fails on `'skipped'`; a job-level `if:` would permanently block every docs-only PR. No existing leg uses a job-level `if:`.
- **`concurrency: cancel-in-progress: true`** makes a slow Windows leg a new source of `cancelled`, which the aggregator also treats as failure. Budget the `timeout-minutes` accordingly.

## Honest payoff — read this before sizing the work

**Detection latency, not coverage.** `cross-os.yml` already runs full `verify:quick` + audits + build on `windows-latest` twice weekly, and `release.yml` gates every tag on Windows. This moves Windows detection from <=3.5 days to per-PR.

It does **not** close the tmpdir/handle race — that is a concurrency artifact and a `windows-latest` runner is single-tenant, so it structurally cannot reproduce there. Do not claim otherwise in the PR body.

## Acceptance

1. The Windows job appears in the aggregator's `needs:` and its result is read by the existing `jq` loop with no second edit.
2. A docs-only PR still completes green — verified by reasoning about the step-level conditions, and ideally by an actual docs-only PR run.
3. `shell: bash` on every step; `timeout-minutes` set.
4. The PR body states the payoff as detection latency, not coverage.

Refs #2997

## Benefit

*(technical)* Windows CRLF/path regressions surface on the PR that introduces them rather than up to 3.5 days later on a cron.

dudarenok-maker:
## Summary

Parts **5** and **6** of the [commit-gate rebalance design](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md), grouped exactly as its rollout order groups them.

### Part 5 — a Windows leg in PR-time CI (#3049)

A `windows-latest` job running `npm test` and `npm run test:server`, added to the aggregator's `needs:` list so it is not decorative — the `jq -r 'keys[]'` loop already iterates all of `needs`, so no second edit was required and the required-check name is unchanged.

Three wiring traps, all avoided deliberately and commented in place:

- **Scope conditions sit on STEPS, never the job.** The aggregator fails on `'skipped'`, so a job-level `if:` would leave this job permanently skipped on every docs-only PR and block merge forever. No other leg in the file uses one; this one doesn't either.
- **Every step pins `shell: bash`.** The file sets no `defaults: run: shell:`, so a bare `run:` on `windows-latest` would default to pwsh.
- **`cancel-in-progress: true`** makes a slow Windows leg a new source of `cancelled`, which the aggregator also treats as failure — `timeout-minutes: 30` is sized with that in mind.

**The payoff is detection latency, not coverage**, and the PR says so rather than overselling it. `cross-os.yml` already runs full `verify:quick` + audits + build on `windows-latest` twice weekly, and `release.yml` gates every tag on Windows; this moves Windows detection from ≤3.5 days to per-PR. It does **not** close the tmpdir/handle race — that is a concurrency artifact between co-running batteries, and a `windows-latest` runner is single-tenant, so it structurally cannot reproduce there.

### Part 6 — scope-triggered local acceptance for the sidecar (#3050)

`.github/workflows/sidecar-acceptance-gate.yml` mirrors `pr-issue-link.yml`'s shape: a PR touching `server/tts-sidecar/**` fails unless its body either records the local run (a backticked `` `npm run test:sidecar` ``, an ISO date, and an outcome) or links an on-box acceptance register row.

Mechanised rather than left as a checklist item, because it supplements an *automatic* trigger and an unenforced item is strictly weaker than what it supplements.

- **Trigger path is `server/tts-sidecar/**` only.** An earlier design draft also listed `server/src/{tts,analyzer,gpu}/**` — 177 TypeScript test files Ubuntu CI already covers. Verified the matcher does not fire on those, nor on an incidental substring.
- **Only `passed` satisfies the gate.** A recorded `failed` run parses correctly and still fails the check.
- **The register-link alternative requires the filename *and* a row id on the same `Sidecar acceptance:` line**, not a bare mention anywhere in the body.

Format documented where an author actually reads it: a new CONTRIBUTING.md section, sibling to "Doc-only PR fast-path", and the Test-plan comment block in the PR template.

## Test plan

`scripts/tests/validate-sidecar-acceptance.test.mjs` — **35 tests**, unit and CLI level, all green; full `npm run test:hooks` battery (1957 tests) green.

**Mutation-verified** — each mutation reddened a named test, then was restored:

| Mutation | Test that reddened |
|---|---|
| `touchesSidecar` forced false | its positive-match test |
| dropped the `outcome === 'passed'` restriction | the "failed run parses but doesn't satisfy" unit test **and** its CLI counterpart |
| widened the register-link regex to drop the same-line prefix | the "no prefix" reject case |
| broke `parseFileList`'s blank-line filter | both of its tests |
| `passesSidecarAcceptanceGate` short-circuited to `true` | both the unit and CLI "neither form fails" tests |

**Scope conditions traced through three diff shapes** against `ci-scope.mjs`'s existing machinery (unchanged here):

- **docs-only** — every scope key false, both Windows steps skip, the gate's file list is empty, both new checks report green fast.
- **`server/tts-sidecar/**`** — the gate now requires a recorded run or register link. (Correction, from review: the Windows steps do **not** uniformly skip. A sidecar path such as `server/tts-sidecar/main.py` also matches `server/**`, so `step_test_server` is true and the Windows server step runs. The original claim here was wrong.)
- **frontend-only** — Windows "Frontend tests" runs, "Server tests" skips; the gate passes with nothing required.

The YAML was parsed with `js-yaml` to confirm `windows-tests` is in `verify.needs` and carries no job-level `if:`.

## Not done, deliberately

Acceptance criterion 5 of #3050 — wiring the new check into `main`'s required status checks — is a branch-protection ruleset change, not a file in this repo, and it cannot be made until the check has run at least once. It is a follow-up action on the repo settings, tracked on #3050.

Before-shipping steps not applicable: **1** (no plan doc — the design doc is the plan and already covers these parts), **3** (no hardware-provable behaviour), **4** (no plan added or moved), **5** (CI/process only, no shippable delta).

Closes #3049
Refs #3050
Refs #2997

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -24,6 +24,13 @@ Checklist of what was run / what reviewers should look at. Examples:
 - [ ] cloud `verify.yml` (required status check) — green
 - [ ] Manual walkthrough in mock mode: <steps>
 - [ ] Reviewer: spot-check <file>:<line> against the regression plan
+
+If this PR touches server/tts-sidecar/**, the sidecar acceptance gate needs
+one of these two lines somewhere in this body (see CONTRIBUTING.md "Sidecar
+acceptance fast-path" for the full format):
+
+  Sidecar acceptance: `npm run test:sidecar` -- <YYYY-MM-DD> -- passed
+  Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row <ID>
 -->
 
 - [ ] `npm run verify` — green
--- a/.github/workflows/sidecar-acceptance-gate.yml
+++ b/.github/workflows/sidecar-acceptance-gate.yml
@@ -1,0 +1,71 @@
+name: Sidecar acceptance gate
+
+# ops-74 (#3050) — commit-gate rebalance design Part 6. Mirrors
+# .github/workflows/pr-issue-link.yml's shape: a small, always-triggered
+# validator gated on a specific, checkable PR-body format -- but this one
+# only requires that format when the PR's diff touches
+# server/tts-sidecar/**, the ONE directory whose 38
+# `pytest.importorskip("torch")` tests (14 files) run exclusively via a
+# local `npm run test:sidecar` on real hardware and have no automated CI
+# equivalent (see docs/superpowers/specs/2026-09-05-commit-gate-rebalance-
+# design.md, Part 6, and CONTRIBUTING.md "Sidecar acceptance fast-path").
+#
+# Trigger path is server/tts-sidecar/** ONLY -- deliberately narrower than an
+# earlier draft that also listed server/src/tts/**, server/src/analyzer/**,
+# server/src/gpu/** (177 TypeScript test files Ubuntu CI already covers via
+# verify.yml). Do not add those back.
+
+on:
+  pull_request:
+    # Same four types as pr-issue-link.yml, for the same reason: `edited`
+    # stays because the primary way a PR satisfies this gate after opening
+    # is editing the body to add the "Sidecar acceptance:" line, not pushing
+    # a new commit -- without `edited` that fix wouldn't clear the check
+    # until the next `synchronize` push.
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    name: Verify sidecar acceptance is recorded
+    runs-on: ubuntu-latest
+    # Tiny validator; cap explicitly so a hung run can't bill the 360-min
+    # default (same rationale as pr-title-lint.yml / pr-issue-link.yml).
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          # Full history so the scope diff below can reach the PR's merge
+          # base. A shallow clone wouldn't have it reachable.
+          fetch-depth: 0
+
+      - name: Set up Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: '24'
+
+      - name: List changed files
+        # Same merge-base pattern as verify.yml's `detect` job: diff against
+        # the merge base, not BASE itself, so a PR branch cut from a stale
+        # main doesn't pick up every file main moved past the fork point.
+        run: |
+          BASE="${{ github.event.pull_request.base.sha }}"
+          HEAD="${{ github.event.pull_request.head.sha }}"
+          MERGE_BASE="$(git merge-base "$BASE" "$HEAD" 2>/dev/null || echo "$BASE")"
+          git diff --name-only "$MERGE_BASE" "$HEAD" > "$RUNNER_TEMP/pr-files.txt"
+          echo "Changed files in this PR:"
+          cat "$RUNNER_TEMP/pr-files.txt"
+
+      - name: Write PR body to a temp file
+        # Same file-passing pattern as pr-issue-link.yml's PR_BODY handling
+        # — the body can't be interpreted as a shell expression this way.
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          printf '%s\n' "$PR_BODY" > "$RUNNER_TEMP/pr-body.txt"
+
+      - name: Validate sidecar acceptance
+        run: node scripts/validate-sidecar-acceptance.mjs "$RUNNER_TEMP/pr-files.txt" "$RUNNER_TEMP/pr-body.txt"
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -551,13 +551,86 @@ jobs:
         if: fromJSON(needs.detect.outputs.scopes).step_build || fromJSON(needs.detect.outputs.scopes).shared
         run: npm run build
 
+  # Windows leg (ops-73 / #3049, commit-gate rebalance design Part 5).
+  # Detection latency, not new coverage: cross-os.yml already runs the full
+  # `verify:quick` battery on windows-latest twice weekly, and release.yml
+  # gates every tag on Windows. This moves Windows detection from <=3.5 days
+  # to per-PR. It does NOT close the tmpdir/handle race the old pre-push
+  # hooks were built around — that race is a concurrency artifact between
+  # co-running batteries, and this runner is single-tenant, so it structurally
+  # cannot reproduce here.
+  #
+  # Scope conditions sit on STEPS, never the job — a job-level `if:` would
+  # leave this job (and therefore the aggregator's `needs:` entry for it)
+  # PERMANENTLY SKIPPED on every docs-only PR, and the aggregator fails on
+  # `'skipped'` below, which would block merge forever. No other leg in this
+  # file uses a job-level `if:`; this one doesn't either.
+  #
+  # `cancel-in-progress: true` (workflow-wide, top of file) makes a slow
+  # Windows leg a new source of `cancelled`, which the aggregator also treats
+  # as failure — timeout-minutes is sized with that in mind.
+  #
+  # Every step pins `shell: bash` explicitly: this file sets no workflow- or
+  # job-level `defaults: run: shell:`, so a bare `run:` on windows-latest
+  # would default to pwsh, not bash.
+  windows-tests:
+    name: Windows (test + test:server)
+    needs: detect
+    runs-on: windows-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Node + deps
+        uses: ./.github/actions/setup
+
+      # ffmpeg is needed by the server MP3 encoder (server/src/tts/mp3.ts
+      # shells out to it). See .github/actions/install-ffmpeg-windows for why
+      # this installs via choco with a winget fallback and a bounded retry.
+      # The Windows steps carry their OWN leg identifiers, not the Ubuntu
+      # ones: the two legs genuinely differ in scope. Ubuntu's `test:server`
+      # runs the fast AND slow suites, so it is gated on step_test_server_slow
+      # too; this job runs `npm run test:server`, which EXCLUDES the slow
+      # files, so a PR touching only a slow test file must not start a Windows
+      # leg that cannot execute it. Marker values also stay inside the wiring
+      # guard's grammar (lowercase, digits, `:` and `-` only) -- a ` (Windows)`
+      # suffix parses as nothing at all, which left both steps outside every
+      # guard in scripts/tests/workflow-wiring.test.mjs (#3053 review).
+      # supports: test:server:windows
+      - name: Install ffmpeg (Windows)
+        if: fromJSON(needs.detect.outputs.scopes).step_test_server || fromJSON(needs.detect.outputs.scopes).shared
+        uses: ./.github/actions/install-ffmpeg-windows
+
+      # leg: test:windows
+      - name: Frontend tests
+        if: fromJSON(needs.detect.outputs.scopes).step_test || fromJSON(needs.detect.outputs.scopes).shared
+        shell: bash
+        run: npm test
+
+      # leg: test:server:windows
+      - name: Server tests
+        if: fromJSON(needs.detect.outputs.scopes).step_test_server || fromJSON(needs.detect.outputs.scopes).shared
+        shell: bash
+        run: npm run test:server
+
   # Thin aggregator — see the top-of-file comment. Keeps the "npm run verify"
   # status-check name (required on main's branch-protection ruleset) stable
   # while the real work runs concurrently in the named jobs above.
   verify:
     name: npm run verify
     needs:
-      [detect, lint-and-checks, frontend-tests, server-tests, sidecar-tests, e2e, e2e-visual, build]
+      [
+        detect,
+        lint-and-checks,
+        frontend-tests,
+        server-tests,
+        sidecar-tests,
+        e2e,
+        e2e-visual,
+        build,
+        windows-tests,
+      ]
     runs-on: ubuntu-latest
     if: always()
     steps:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -590,6 +590,46 @@ docs-only push has no runtime surface for it to exercise, so paying even
 that scope computation is wasted time/CPU for zero signal. See
 [CLAUDE.md "Commit gate"](CLAUDE.md#commit-gate).
 
+### Sidecar acceptance fast-path
+
+A PR whose diff touches `server/tts-sidecar/**` is a required check
+([`.github/workflows/sidecar-acceptance-gate.yml`](.github/workflows/sidecar-acceptance-gate.yml),
+ops-74 / #3050) unless the PR body records that `npm run test:sidecar` was
+run locally, or links a row in the [on-box acceptance
+register](docs/testing/onbox-acceptance-register.md). This exists because 38
+`pytest.importorskip("torch")` tests (14 files, all under
+`server/tts-sidecar/`) run **only** on real hardware via a local
+`npm run test:sidecar` — never in CI (`verify.yml`'s sidecar leg deliberately
+installs the lean, torch-free dependency set, so those tests report as
+skipped there). A PR that doesn't touch `server/tts-sidecar/**` is unaffected
+and needs nothing in its body.
+
+Write one of these two lines plainly (not inside backticks or a code block)
+in the PR body:
+
+```
+Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- passed
+Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row <ID>
+```
+
+The first form is checked field-by-field — a literal
+`` `npm run test:sidecar` `` (flags after it are fine), an ISO date, and an
+outcome of `passed` or `failed`; only `passed` satisfies the gate. Free
+prose ("ran the sidecar tests, all good") does not. The second form points
+at a specific register row rather than re-running locally — useful when
+acceptance is already tracked there. `<ID>` must name a row that **actually
+exists** in the register: the gate parses the register (with
+`check-register-citations.mjs`'s own row parser) and rejects an id it does
+not find, so "blocked until Q3" or "will file in v2" is not a citation. And
+"plainly" is enforced, not merely asked for — a line inside a backtick or
+tilde fence, inside an HTML comment, indented by four spaces or a tab, or
+inside an inline code span opened on an earlier line, does not satisfy the
+gate. See
+[`scripts/validate-sidecar-acceptance.mjs`](scripts/validate-sidecar-acceptance.mjs)
+for the exact patterns and
+[docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md)
+("Part 6") for the rationale.
+
 ## When you ship a change
 
 The full checklist lives in [CLAUDE.md → "Before-shipping checklist"](CLAUDE.md#before-shipping-checklist).
--- a/scripts/tests/validate-sidecar-acceptance.test.mjs
+++ b/scripts/tests/validate-sidecar-acceptance.test.mjs
@@ -1,0 +1,777 @@
+// Tests for the sidecar-acceptance-gate validator (ops-74 / #3050).
+// Run via `npm run test:hooks` (node --test, no extra deps).
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, writeFileSync, rmSync, readFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { spawnSync } from 'node:child_process';
+import {
+  touchesSidecar,
+  unquoteGitPath,
+  parseFileList,
+  parseRecordedRun,
+  hasPassingRecordedRun,
+  hasRegisterLink,
+  parseRegisterLink,
+  passesSidecarAcceptanceGate,
+  helpMessage,
+  loadRegisterRowIds,
+  readRegisterRowIds,
+  failureDetail,
+} from '../validate-sidecar-acceptance.mjs';
+
+// Synthetic register row IDs, used ONLY as fixtures for this validator's
+// regex. Composed at runtime instead of written literally because
+// scripts/check-register-citations.mjs reads the literal "row <ID>" idiom in
+// any tracked file as a CITATION of a real register row -- and these are not
+// citations, they are inputs to a pattern. Written literally they made that
+// checker exit 1 on the whole tree and, through it, disarmed its own
+// mutation test (`CLI mutation: deleting Check C from fatalSections ...`),
+// which proves its fatality by asserting a mutant PASSES. The runtime string
+// each test actually exercises is unchanged.
+const row = (group, n) => `${group}${n}`;
+const FIXTURE_ROW_ID = row('A', 101);
+const FIXTURE_ROW_ID_ALT = row('E', 101);
+
+// The gate now checks that a cited row EXISTS (#3053 review pass 2, N1), so
+// every syntax-level assertion below hands it this explicit set rather than
+// the real register -- keeping the pattern fixtures synthetic, exactly as
+// the comment above requires.
+const FIXTURE_ROWS = new Set([FIXTURE_ROW_ID, FIXTURE_ROW_ID_ALT, row('C', 2)]);
+
+// The real register's row ids, for the existence half. Read through the
+// SAME parser the validator uses, so a fixture id can never drift from what
+// the register actually contains.
+const REAL_ROW_IDS = loadRegisterRowIds();
+const A_REAL_ROW_ID = [...REAL_ROW_IDS].sort()[0];
+
+const scriptPath = resolve(
+  dirname(fileURLToPath(import.meta.url)),
+  '..',
+  'validate-sidecar-acceptance.mjs',
+);
+
+// --- touchesSidecar -------------------------------------------------------
+
+test('touchesSidecar: true when any file is under server/tts-sidecar/', () => {
+  assert.equal(touchesSidecar(['server/tts-sidecar/main.py']), true);
+  assert.equal(
+    touchesSidecar(['src/App.tsx', 'server/tts-sidecar/tests/test_smoke.py']),
+    true,
+  );
+});
+
+test('touchesSidecar: false for an empty list, non-array, or no sidecar path', () => {
+  assert.equal(touchesSidecar([]), false);
+  assert.equal(touchesSidecar(undefined), false);
+  assert.equal(touchesSidecar(null), false);
+  assert.equal(touchesSidecar(['src/App.tsx', 'server/src/tts/index.ts']), false);
+});
+
+// The trigger path is deliberately narrow -- server/src/tts/**,
+// server/src/analyzer/**, server/src/gpu/** must NOT trip this gate (the
+// earlier draft this issue explicitly rejected). A path that merely
+// CONTAINS the substring "tts-sidecar" elsewhere in the tree (not as this
+// exact directory prefix) must not false-positive either.
+test('touchesSidecar: does not match the rejected wider paths or a substring elsewhere', () => {
+  assert.equal(touchesSidecar(['server/src/tts/index.ts']), false);
+  assert.equal(touchesSidecar(['server/src/analyzer/ollama.ts']), false);
+  assert.equal(touchesSidecar(['server/src/gpu/active-generation-gate.ts']), false);
+  assert.equal(touchesSidecar(['docs/notes/tts-sidecar-history.md']), false);
+});
+
+// --- parseFileList ----------------------------------------------------------
+
+test('parseFileList: splits newline-separated text and drops blank lines', () => {
+  assert.deepEqual(parseFileList('a.txt\nb.txt\n\nc.txt\n'), ['a.txt', 'b.txt', 'c.txt']);
+});
+
+test('parseFileList: empty diff (empty string) yields an empty array, not ["")', () => {
+  assert.deepEqual(parseFileList(''), []);
+});
+
+test('parseFileList: non-string input yields an empty array', () => {
+  assert.deepEqual(parseFileList(undefined), []);
+});
+
+// --- parseRecordedRun / hasPassingRecordedRun ------------------------------
+
+const acceptedRecordedRuns = [
+  'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- passed',
+  'Sidecar acceptance: `npm run test:sidecar -- --require-venv` -- 2026-01-05 -- passed',
+  'SIDECAR ACCEPTANCE: `npm run test:sidecar` -- 2026-09-06 -- PASSED',
+  'Intro text.\n\nSidecar acceptance: `npm run test:sidecar` — 2026-09-06 — passed\n\nMore text.',
+  // en dash -- the third accepted separator, and the one the code comment
+  // used to omit while naming a single ASCII hyphen it does not accept
+  'Sidecar acceptance: `npm run test:sidecar` – 2026-09-06 – passed',
+];
+
+for (const body of acceptedRecordedRuns) {
+  test(`hasPassingRecordedRun accepts: ${JSON.stringify(body)}`, () => {
+    assert.equal(hasPassingRecordedRun(body), true);
+  });
+}
+
+test('parseRecordedRun: extracts command, date, and outcome', () => {
+  const parsed = parseRecordedRun(
+    'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- passed',
+  );
+  assert.deepEqual(parsed, {
+    command: 'npm run test:sidecar',
+    date: '2026-09-06',
+    outcome: 'passed',
+  });
+});
+
+// A recorded FAILED run is a real, parseable record -- but must not satisfy
+// the gate. This is the outcome-vocabulary check, not merely "some text
+// present".
+test('a recorded run with outcome "failed" parses but does not satisfy the gate', () => {
+  const body = 'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- failed';
+  assert.deepEqual(parseRecordedRun(body), {
+    command: 'npm run test:sidecar',
+    date: '2026-09-06',
+    outcome: 'failed',
+  });
+  assert.equal(hasPassingRecordedRun(body), false);
+});
+
+const rejectedRecordedRuns = [
+  '',
+  'Ran the sidecar tests locally, all good.', // free prose, not the checkable format
+  'Sidecar acceptance: npm run test:sidecar -- 2026-09-06 -- passed', // command not backticked
+  'Sidecar acceptance: `npm test` -- 2026-09-06 -- passed', // wrong command
+  'Sidecar acceptance: `npm run test:sidecar` -- 09-06-2026 -- passed', // non-ISO date
+  'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- ok', // outcome not in vocabulary
+  // A SINGLE ASCII hyphen is not one of the three accepted separators
+  // (`--`, em dash, en dash). The code comment used to claim it was; this
+  // fixture is what keeps comment and code from drifting apart again.
+  'Sidecar acceptance: `npm run test:sidecar` - 2026-09-06 - passed',
+];
+
+for (const body of rejectedRecordedRuns) {
+  test(`hasPassingRecordedRun rejects: ${JSON.stringify(body)}`, () => {
+    assert.equal(hasPassingRecordedRun(body), false);
+  });
+}
+
+test('hasPassingRecordedRun rejects non-string input', () => {
+  assert.equal(hasPassingRecordedRun(undefined), false);
+  assert.equal(hasPassingRecordedRun(null), false);
+});
+
+// --- hasRegisterLink --------------------------------------------------------
+
+const acceptedRegisterLinks = [
+  `Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row ${FIXTURE_ROW_ID}`,
+  // Lowercase on purpose: GitHub lowercases heading anchors, so this is the
+  // form a real link into the register carries. It used to read #C2, which
+  // is the only form the `#` branch accepted and the one form that never
+  // occurs (#3053 review pass 3, P2).
+  `Sidecar acceptance: docs/testing/onbox-acceptance-register.md#${row('C', 2).toLowerCase()}`,
+  `sidecar acceptance: recorded at docs/testing/onbox-acceptance-register.md, row ${FIXTURE_ROW_ID_ALT}`,
+];
+
+for (const body of acceptedRegisterLinks) {
+  test(`hasRegisterLink accepts: ${JSON.stringify(body)}`, () => {
+    assert.equal(hasRegisterLink(body, FIXTURE_ROWS), true);
+  });
+}
+
+const rejectedRegisterLinks = [
+  '',
+  `See docs/testing/onbox-acceptance-register.md row ${FIXTURE_ROW_ID}`, // no "Sidecar acceptance:" prefix on the line
+  'Sidecar acceptance: docs/testing/onbox-acceptance-register.md', // no row ID
+  // The register is mentioned elsewhere in the body, but not on a line that
+  // also carries the "Sidecar acceptance:" prefix and a row id -- a bare,
+  // unrelated mention must not satisfy the gate.
+  'This references docs/testing/onbox-acceptance-register.md in passing.\n\nSidecar acceptance: pending.',
+];
+
+for (const body of rejectedRegisterLinks) {
+  test(`hasRegisterLink rejects: ${JSON.stringify(body)}`, () => {
+    assert.equal(hasRegisterLink(body, FIXTURE_ROWS), false);
+  });
+}
+
+// --- passesSidecarAcceptanceGate -------------------------------------------
+
+test('passesSidecarAcceptanceGate: a PR that does not touch the sidecar always passes', () => {
+  assert.equal(passesSidecarAcceptanceGate(['src/App.tsx'], ''), true);
+  assert.equal(passesSidecarAcceptanceGate([], 'no relevant text'), true);
+});
+
+test('passesSidecarAcceptanceGate: a sidecar-touching PR with neither form fails', () => {
+  assert.equal(
+    passesSidecarAcceptanceGate(['server/tts-sidecar/main.py'], 'Nothing relevant here.'),
+    false,
+  );
+});
+
+test('passesSidecarAcceptanceGate: a sidecar-touching PR with a passing recorded run passes', () => {
+  assert.equal(
+    passesSidecarAcceptanceGate(
+      ['server/tts-sidecar/main.py'],
+      'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- passed',
+    ),
+    true,
+  );
+});
+
+test('passesSidecarAcceptanceGate: a sidecar-touching PR with a register link passes', () => {
+  assert.equal(
+    passesSidecarAcceptanceGate(
+      ['server/tts-sidecar/main.py'],
+      `Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row ${FIXTURE_ROW_ID}`,
+      FIXTURE_ROWS,
+    ),
+    true,
+  );
+});
+
+// --- "written plainly": fenced blocks and HTML comments ---------------------
+// helpMessage() and CONTRIBUTING.md "Sidecar acceptance fast-path" both tell
+// the author to write the line plainly, NOT inside a code block. Without
+// stripping, that instruction was unenforced and the documented copy-paste
+// block was itself a passing body.
+
+const FENCE = '`'.repeat(3);
+
+test('a recorded run inside a fenced code block does NOT satisfy the gate', () => {
+  const body = [
+    'We did not get to the box this round.',
+    '',
+    FENCE,
+    'Sidecar acceptance: `npm run test:sidecar` -- 2026-09-06 -- passed',
+    FENCE,
+  ].join('\n');
+  assert.equal(hasPassingRecordedRun(body), false);
+  assert.equal(passesSidecarAcceptanceGate(['server/tts-sidecar/main.py'], body), false);
+});
+
+test('a register link inside a fenced code block does NOT satisfy the gate', () => {
+  const body = [
+    FENCE,
+    `Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row ${FIXTURE_ROW_ID}`,
+    FENCE,
+  ].join('\n');
+  assert.equal(hasRegisterLink(body, FIXTURE_ROWS), false);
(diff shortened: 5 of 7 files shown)


dudarenok-maker:
## PR review — pass 1 (head 829341a5, depth high)

Scope: all 6 changed files (`.github/workflows/verify.yml`, `.github/workflows/sidecar-acceptance-gate.yml`, `.github/pull_request_template.md`, `CONTRIBUTING.md`, `scripts/validate-sidecar-acceptance.mjs`, `scripts/tests/validate-sidecar-acceptance.test.mjs`), read against design-doc Parts 5/6, #3049, #3050, and the sibling machinery they claim to mirror (`pr-issue-link.yml` + `validate-pr-issue-link.mjs`, `ci-scope.mjs`, `verify-cache.mjs` `STEPS[]`, `workflow-wiring.test.mjs`, `check-register-citations.mjs`, `spawn-windows-hide.test.ts`).

Verified before probing: `node --test scripts/tests/validate-sidecar-acceptance.test.mjs` → 35/35 green; all five claimed mutations re-performed on an out-of-repo copy and each reddened its named test; the three diff shapes recomputed by calling `computeScopes()` directly; `node --test scripts/tests/workflow-wiring.test.mjs` → 16 pass / **1 fail**; `node scripts/check-register-citations.mjs` → **exit 1**; CI run 34023421467 read leg by leg.

**Precondition not met:** this gate assumes cloud `verify.yml` is green. It is red at this head — `npm run verify` (the required check) fails, along with `Lint, typecheck & fast checks`, `Server tests (fast + slow)`, and the new `Windows (test + test:server)` job. All three failures are caused by this PR's own diff. They are findings 1–3 below.

---

### 🔴 Blocking — the new validator's regex literal makes `scripts/**` unscannable, failing `test:server` on **both** OSes

`scripts/validate-sidecar-acceptance.mjs:47-48` — `RECORDED_RUN_PATTERN` embeds **three literal backticks** inside a regex literal.

`server/src/spawn-windows-hide.test.ts` scans `scripts/**/*.{mjs,cjs,js}` for spawn calls, and its `regexLiteralDesyncsQuoteTracking` guard (added for #2747) **throws** on exactly this shape, because an unpaired backtick inside a regex literal desyncs the scanner's quote tracking and would silently blank the rest of the file — hiding any real spawn call after it.

Exact input → wrong output:

- input: `npm run test:server` at head 829341a5
- output: `FAIL src/spawn-windows-hide.test.ts` → `Error: Ambiguous regex literal desyncs quote tracking at offset 2329: /^\s*sidecar acceptance:.../` thrown at `server/src/spawn-windows-hide.test.ts:412:17`, via `filterSpawningFiles:206` ← `externalFilesFloor:275`.
- Reproduced in CI on **both** platforms: job `101460108163` (Ubuntu, `Server tests (fast + slow)`) and job `101460108279` (Windows, `Server tests`) — 1 failed / 539 passed in each. This is not Windows-specific; the new Windows leg simply reports it twice.

The guard's own message states the fix: escape the character as ``` rather than using a literal backtick inside the regex. (The opposite direction was reviewed recently too — `c3a0bfd6` "revert two unnecessary \uXXXX escapes" — so the rule is *inside a regex literal only*, which is where these three sit.)

Correctness bug.

---

### 🔴 Blocking — the new test's placeholder register IDs are read as real citations, so `check-register-citations` exits 1

`scripts/tests/validate-sidecar-acceptance.test.mjs:131`, `:133`, `:144`, `:186`, `:234` use `A101` / `E101` as *fixture* row ids next to the literal string `docs/testing/onbox-acceptance-register.md`. `scripts/check-register-citations.mjs` treats that pairing as a citation and its **Check A (nonexistent row ID) is fatal**.

Exact input → wrong output:

- input: `node scripts/check-register-citations.mjs` on this branch
- output: **exit 1**, and the fatal section contains *only* this PR's file:

```
Check A — nonexistent row ID (5):
  scripts/tests/validate-sidecar-acceptance.test.mjs:131 — cited A101 — no such row in docs/testing/onbox-acceptance-register.md (nonexistent ID)
  scripts/tests/validate-sidecar-acceptance.test.mjs:133 — cited E101 — no such row ...
  scripts/tests/validate-sidecar-acceptance.test.mjs:144 — cited A101 — no such row ...
  scripts/tests/validate-sidecar-acceptance.test.mjs:186 — cited A101 — no such row ...
  scripts/tests/validate-sidecar-acceptance.test.mjs:234 — cited A101 — no such row ...
```

That non-zero exit is what reddens **six** CLI tests in `scripts/tests/check-register-citations.test.mjs` (they assert exit 0 against the real tree), which is the `Hooks tests` failure in job `101460108131`:

- `CLI: without --strict, Check C's exploratory half does not print...` → `1 !== 0`
- `CLI: with --strict, Check C's exploratory half runs...` → `1 !== 0`
- `CLI: Check C's wrongId half is FATAL...` → `expected exit 0 for args []`
- `CLI mutation: deleting Check C from fatalSections un-gates a real wrong-ID heading...` → `mutant should incorrectly pass once Check C is un-gated`
- `CLI: the real tree scan finds at least one binary file...` → `1 !== 0`
- `CLI: with --strict, Check D finds title drift but does NOT fail the gate` → `Check D drift findings are advisory, should not fail`

Note the fourth one: this PR's citations **disarm an existing mutation test**. That test proves Check C's fatality is load-bearing by asserting a mutant *passes*, and it can no longer pass because Check A is now fatal. So the damage is not only "red build"; it is a guard whose own proof-of-life has been neutralised.

I checked the two other new `A101` sites and they are *not* flagged: `CONTRIBUTING.md`'s example sits in a fenced block (`stripFences` blanks it) and `helpMessage()`'s copy is not picked up. Only the five test-file lines are fatal, so the fix is local to that file (a nearby discharge/removal annotation, ids that exist, or a fixture id the checker's grammar does not read as a citation — whichever its annotation rules make cleanest).

Correctness bug.

---

### 🔴 Blocking — the Windows job's `# leg:` markers are invisible to `workflow-wiring.test.mjs`, and its `# supports:` marker cross-binds to the **Ubuntu** leg

`.github/workflows/verify.yml:591` (`# supports: test:server`), `:596` (`# leg: test (Windows)`), `:602` (`# leg: test:server (Windows)`).

`scripts/tests/workflow-wiring.test.mjs:246` parses leg markers with `/# leg: ([a-z:0-9-]+)\n\s*- name: .../`. The character class excludes a space and `\n` must follow immediately, so a marker with a ` (Windows)` suffix matches **nothing**. Proven directly against the real file:

```
legs captured by the guard (17):
  lint, typecheck, config:check, openapi:types, test:hooks, check:budget-poll, audit,
  audit:server, test:scripts, test:pinokio, test, test:server, check:cycles,
  test:sidecar, test:e2e, test:e2e:visual, build
raw "# leg:" markers present in the file:
  ... "test (Windows)"          <- NOT captured
  ... "test:server (Windows)"   <- NOT captured
```

The `# supports:` marker at `:591` *is* captured (its value is the bare `test:server`), so it resolves to `legs.get('test:server')` — the **Ubuntu** `Server tests (fast + slow)` step at `:346`, whose keys are `shared|step_test_server|step_test_server_slow`. The Windows ffmpeg step's keys are `shared|step_test_server`.

Exact input → wrong output:

- input: `node --test scripts/tests/workflow-wiring.test.mjs`
- output: `✖ setup steps depend on the same scope keys as the leg they support` —
  `Install ffmpeg (Windows) / supports: test:server / has keys: shared|step_test_server / leg has keys: shared|step_test_server|step_test_server_slow`
- Same failure in CI, job `101460108131`, at `workflow-wiring.test.mjs:276`.

Two distinct defects, and fixing only the red one leaves the second armed:

1. **The red.** The Windows ffmpeg step is annotated as supporting a leg it does not sit next to. Its own condition is correct for its own leg (Windows runs `test:server` only, never `test:server-slow`), so the marker vocabulary — not the condition — is what is wrong.
2. **Silent loss of coverage.** Because both Windows `# leg:` markers are unparseable, the Windows job's two leg steps sit outside *every* wiring guard in that file, and `legs.size >= 11` still passes on the Ubuntu markers alone, so nothing reports the gap. This is the "guard that enumerates syntax, and loses one spelling per round" shape — the marker grammar became one-job-per-workflow the moment a second OS was added, and neither the grammar nor its guard was widened to match.

Correctness bug.

---

### 🟠 Significant — the sidecar gate is satisfied by a fenced code block or an HTML comment, which both the code's own help text and CONTRIBUTING.md say will not satisfy it

`scripts/validate-sidecar-acceptance.mjs:47` / `:70` do no code-span stripping, while `scripts/validate-sidecar-acceptance.mjs:94-95` (`helpMessage()`) and `CONTRIBUTING.md:603` both tell the author to write the line "plainly (**not inside backticks or a code block**)".

Exact input → wrong output. Files `['server/tts-sidecar/main.py']`, body = a paragraph saying nothing was done, followed by a triple-backtick fence containing the line `Sidecar acceptance: (backtick)npm run test:sidecar(backtick) -- 2026-09-06 -- passed` → `passesSidecarAcceptanceGate(...)` returns **`true`** and the CLI exits 0. Same result with the line inside `<!-- ... -->`.

Also verified: feeding **the whole of `CONTRIBUTING.md`** as a PR body passes the gate, because its documented example block is a fence containing a real date and a real row id — i.e. the copy-paste block the docs present *is itself* a passing body, and the one instruction that would have prevented that ("write it plainly") is unenforced. The likely path here is not gaming but honest copy-paste.

The stated model, `scripts/validate-pr-issue-link.mjs:29-70`, already solves this with `stripFencedBlocks` + `stripInlineSpans`, and its comments record that a code-span false positive was a real incident there. Given Part 6's own rationale — "an unenforced item is strictly weaker than what it supplements" — a gate that goes green on a body where nothing was run is the failure this part exists to prevent.

I checked the worst case first and it is clean: the **unedited PR template does not** satisfy the gate — `<YYYY-MM-DD>` fails the ISO date and `<ID>` carries no digits, so neither pattern matches. That was the one shape that would have made every sidecar PR auto-pass, and the placeholders defeat it.

Correctness bug.

---

### 🟠 Significant — `Closes #3050` closes the issue the PR body says is tracking the undone acceptance criterion

PR body: acceptance criterion 5 of #3050 (wiring the check into `main`'s required status checks) is "a follow-up action on the repo settings, **tracked on #3050**" — and the same body carries `Closes #3050`. On merge GitHub closes #3050, so the follow-up is tracked nowhere and the check stays advisory-only with nothing recording that it should not be.

This also contradicts the repo's own convention: CLAUDE.md's Before-shipping step 6 and the branching workflow both specify `Refs #NN` for a partial delivery and `Closes #NN` only for a full one. Four of #3050's five acceptance criteria are met; this is a partial.

Correctness bug (bookkeeping, but it silently drops a merge-gate wiring step).

---

### 🟡 Minor

**a. Stale comment: the separator does not accept a single hyphen.** `scripts/validate-sidecar-acceptance.mjs:44-46` says the separator is `--`, an em dash, "or `-` with spaces on both sides". The regex alternation is `(?:--|<em dash>|<en dash>)` — double hyphen, em dash, **en dash**; a single ASCII hyphen is not accepted. Repro: `hasPassingRecordedRun('Sidecar acceptance: (backtick)npm run test:sidecar(backtick) - 2026-09-06 - passed')` → `false`. The comment names an alternative the code does not have and omits the one it does. Cleanup nit — the code's behaviour matches CONTRIBUTING.md, which documents `--` only.

**b. The PR body's `server/tts-sidecar/**` scope trace is wrong for `main.py`.** The body asserts "**`server/tts-sidecar/**`** — Windows steps skip". Recomputed against `ci-scope.mjs`:

| diff | `step_test` | `step_test_server` |
|---|---|---|
| `docs/BACKLOG.md`, `README.md` | false | false |
| `server/tts-sidecar/engines/kokoro.py` | false | false |
| **`server/tts-sidecar/main.py`** | false | **true** |
| `src/views/listen.tsx` | true | false |

A `main.py` diff runs the Windows `Install ffmpeg` + `Server tests` steps, via the `extraFiles` exception that `verify.yml:319-325` documents 270 lines above the new job. The docs-only and frontend-only rows are exactly as claimed. Behaviour is harmless; the traced claim is not accurate. Cleanup nit.

**c. `touchesSidecar` misses git-quoted paths.** `scripts/validate-sidecar-acceptance.mjs:22` uses `startsWith('server/tts-sidecar/')`, but `git diff --name-only` (the producer, `sidecar-acceptance-gate.yml:58`) emits non-ASCII paths **quoted and C-escaped** under the default `core.quotepath=true`, e.g. a path beginning with a double-quote character rather than `s`. The gate then does not fire. Only bites a PR whose *entire* sidecar footprint is non-ASCII-named, so likelihood in a Python tree is very low. Cleanup nit.

---

### ✅ What is solid

- **The aggregator claim is true, and I checked it against the file rather than the description.** `verify.yml:641-698`: `jq -r 'keys[]' needs.json` iterates every `needs` key, and the `case` buckets `cancelled` / `failure` / **`skipped`** and exits 1 on any of them, with an explicit default arm treating an unknown result as failed. `windows-tests` carries no job-level `if:` — grep confirms the aggregator's `if: always()` is the only job-level `if:` in the file — so with step-level conditions the job result is `success` even when every step skips. A job-level `if:` really would have made it `skipped` and deadlocked the required check.
- **Docs-only completes green.** `computeScopes(['docs/BACKLOG.md','README.md'])` → every key false → all three conditional Windows steps skip → job success → aggregator green. #3049 AC2 holds.
- **The Part 6 matcher is correctly narrow in both directions.** Does not fire on `server/src/tts/index.ts`, `server/src/analyzer/ollama.ts`, `server/src/gpu/sidecar-health-gate.ts`, `docs/notes/tts-sidecar-history.md`, or `server/tts-sidecarX/foo.py`; does fire on `server/tts-sidecar/main.py` and `server/tts-sidecar/requirements/base.txt`. The trailing slash in `SIDECAR_PREFIX` is what defeats the incidental-substring case. The only path I found that defeats it is 🟡c above.
- **The mutation table is accurate.** I re-performed all five on an out-of-repo copy: `touchesSidecar` → false reddened its positive-match test (+3 more); dropping `outcome === 'passed'` reddened both the unit and CLI "failed run" tests; widening `REGISTER_LINK_PATTERN` to drop the same-line prefix reddened the "no prefix" reject case; breaking `parseFileList`'s filter reddened both of its tests; short-circuiting `passesSidecarAcceptanceGate` to `true` reddened both "neither form fails" tests. No claimed mutation was over-stated.
- **CLI-level coverage is real, not source-text.** The five CLI tests spawn the actual script and assert exit codes, so argv/exit wiring is covered rather than asserted about.
- **`windowsHide`.** The one new spawn call, `scripts/tests/validate-sidecar-acceptance.test.mjs:204-207`, carries the literal `windowsHide: true` inside its own balanced-paren argument text, which is what the syntactic scanner requires. `EXTERNAL_FILES_FLOOR` is scan-derived, not a hand-maintained count, so no floor bump is owed.
- **CRLF-safe.** GitHub PR bodies are CRLF-delimited and `sidecar-acceptance-gate.yml:68` preserves them; `\s*$` absorbs the `\r`, so a real GitHub body parses. Verified.
- **`fetch-depth: 0` on the gate job** is genuinely needed for `git merge-base` and is present. Its absence on `windows-tests` is correct — I grepped the frontend and server suites and neither shells out to git, so a depth-1 clone is sufficient there.
- The workflow-wiring guard's `^ every job with a derived condition is in the aggregator needs:` test passes, mechanically confirming the `needs:` edit.

### Verdict

**Do not merge.** Three blocking findings, all of them from this PR's own diff, all currently red in CI — the required `npm run verify` check is failing at head 829341a5. Findings 1 and 2 are self-contained one-file fixes. Finding 3 needs a decision about the marker grammar rather than just a green build: making `# supports:` / `# leg:` OS-aware, and teaching `workflow-wiring.test.mjs` to parse a suffixed marker (or to key on job as well as leg name), is what stops the Windows job's steps sitting outside every wiring guard. Findings 4 and 5 are correctness bugs that should land in the same round.

The design's three named Part 5 traps were genuinely avoided — I checked each against the file, not against the description. What bit this PR is the surrounding guard mesh — `spawn-windows-hide`, `check-register-citations`, `workflow-wiring` — none of which was run against the branch before it was pushed.

dudarenok-maker:
## PR review — pass 2 (head 827eee10, depth high)

Re-review of the fix commit `827eee10`. Deltas only — the solid items from pass 1 are not re-listed.

**Precondition met, and independently confirmed:** all 21 checks green at this head, and the new Windows leg is not vacuously green — `actions/jobs/101466018299` shows `Install ffmpeg (Windows)` ✔, `Frontend tests` **skipped** (correct: this diff touches no frontend file), `Server tests` ✔ in 5m04s. The leg actually executed the suite that was red in pass 1.

**Verified independently of the fix commit's claims** (each re-run, not read):

| repro | pass 1 | now |
|---|---|---|
| `npx vitest run src/spawn-windows-hide.test.ts` | 1 failed / 539 passed | **22/22 pass**, exit 0 |
| `node scripts/check-register-citations.mjs` | **exit 1**, 5 fatal Check-A rows | **exit 0**, "OK", no fatal rows from this diff |
| `node --test scripts/tests/workflow-wiring.test.mjs` | 16 pass / **1 fail** | **18/18 pass** |
| `node --test scripts/tests/validate-sidecar-acceptance.test.mjs` | 35/35 | **45/45** |
| whole `CONTRIBUTING.md` as a PR body | **gate passes** | **gate fails** |

---

## Pass-1 findings — disposition

### 1. 🔴 Backticks in a regex literal — ✅ VERIFIED RESOLVED

`scripts/validate-sidecar-acceptance.mjs:133` now uses ```. `regexLiteralDesyncsQuoteTracking` no longer throws; the server suite's scan of `scripts/**` completes. Re-run above, plus CI job `101466018287` (Ubuntu) and `101466018299` (Windows) both green.

### 2. 🔴 Placeholder register ids read as citations — ✅ VERIFIED RESOLVED

Ids are composed at runtime (`scripts/tests/validate-sidecar-acceptance.test.mjs:29-31`), no literal `row A101` text remains, and `check-register-citations.mjs` exits 0 with no fatal row from this file. The disarmed mutation test (`CLI mutation: deleting Check C from fatalSections …`) has its precondition back.

**Caveat, and it is the fixer's own report — I confirmed it.** The fix is over-strong for a reason nobody should rely on: the strings are now **template literals**, and `check-register-citations.mjs` cannot see a citation inside one at all. Measured against the real register (62 rows, `A101` absent):

```
SEEN as a citation (FATAL)   'Sidecar acceptance: … row A101',      <- single-quoted (the pass-1 shape)
SEEN as a citation (FATAL)   "Sidecar acceptance: … row A101",      <- double-quoted
not seen                     `Sidecar acceptance: … row A101`,      <- template literal, LITERAL id
not seen                     `Sidecar acceptance: … row ${ID}`,     <- what shipped
```

So simply wrapping the old literal in backticks would have passed too. That is a latent hole in the checker — see new finding **N4**.

### 3. 🔴 Windows `# leg:` markers outside the guard grammar — ✅ VERIFIED RESOLVED (for the four named legs)

Markers are now `test:windows` / `test:server:windows`, the ffmpeg step's `# supports:` binds to the Windows leg, and `REQUIRED_LEGS` (`scripts/tests/workflow-wiring.test.mjs:272`) asserts presence + exact keys. **Mutation-verified against an out-of-repo copy of the tree:** reverting `# leg: test:windows` back to `# leg: test (Windows)` reddens the new test with the right message —

```
✖ every required leg is present and gated on exactly its own scope keys
  '# leg: test:windows' is absent or unparseable (markers found: lint, typecheck, …, test:server:windows)
```

The differing scope keys are also correct, not just asserted: Windows runs `npm run test:server` (fast only), so gating it on `step_test_server_slow` would start a leg that cannot execute the changed file.

**Residual, and it is the half of the pass-1 finding that was about the guard rather than the red:** the general gap ("both Windows leg steps sat outside *every* wiring guard and nothing reported it") is closed only for the four legs now named. See new finding **N3**.

### 4. 🟠 The gate satisfied by a fenced block or HTML comment — ⚠️ PARTIALLY RESOLVED

Triple-backtick fences and HTML comments are stripped, and the motivating case (whole `CONTRIBUTING.md` as a body) now fails. Three other spellings of the same thing still pass — see **N2**.

### 5. 🟠 `Closes #3050` on a partial delivery — ❌ **STILL OPEN**

The **PR body** was corrected to `Refs #3050`. The **commit message was not.** `git log origin/main..HEAD` on this branch:

```
829341a5  ci(ops,scripts): add a Windows PR-CI leg and a sidecar acceptance gate
            …
            Closes #3049
            Closes #3050      <-- still here
            Refs #2997
```

Repo settings confirm the mechanism: `allow_merge_commit: true`, `allow_squash_merge: false`, `allow_rebase_merge: false` — so `829341a5` lands on `main` **verbatim**, and GitHub auto-closes on a commit-message keyword pushed to the default branch regardless of what the PR body says. #3050 is `OPEN` today and would close on merge, taking acceptance criterion 5 (wiring the new check into `main`'s required status checks) with it — the criterion this PR's own "Not done, deliberately" section says is *tracked on #3050*.

This exact failure has happened in this repo before (PR #2126 / #2083, 2026-08-05: body said `Refs`, a lane commit's trailer said `Closes`, the issue closed with six files unswept). Editing the body is not sufficient — the branch is unmerged, so amending `829341a5`'s trailer to `Refs #3050` is cheap.

Correctness bug (bookkeeping, but it silently drops the merge-gate wiring step — same as pass 1).

### Minors — all ✅ RESOLVED

- **a. Separator comment.** `scripts/validate-sidecar-acceptance.mjs:128-131` now names `--` / em dash / en dash and states a single ASCII hyphen is rejected; two paired fixtures added (accepted en dash, rejected single hyphen), so comment and code are now pinned together. Verified by running both.
- **b. PR body's scope trace.** The body now carries an explicit correction naming `main.py` → `step_test_server` true. Recomputed against `computeScopes` — matches.
- **c. `touchesSidecar` and git-quoted paths.** `unquoteGitPath` (`:31`) added with three paired tests; `touchesSidecar(['"server/tts-sidecar/tests/test_caf\\303\\251.py"'])` → `true`, and a quoted path outside the prefix still → `false`. Verified.

---

## New findings

### N1 — 🟠 Significant — the register-link half of the gate goes green on a body that says the tests were **not** run

`scripts/validate-sidecar-acceptance.mjs:155-156` — `REGISTER_LINK_PATTERN` requires the register **filename** and a **row-id-*shaped* token** on the same line. It never checks the id exists, and under the `/i` flag `[A-Z]` also matches lowercase. Nothing downstream validates it either: `check-register-citations.mjs` only scans **tracked files**, and a PR body is not one.

Exact inputs → wrong output, files `['server/tts-sidecar/main.py']`, run through `passesSidecarAcceptanceGate`:

```
GATE PASSES   rowId="Q3"    Sidecar acceptance: NOT RUN. docs/testing/onbox-acceptance-register.md
                            has no row yet; blocked on box time until Q3.
GATE PASSES   rowId="v2"    Sidecar acceptance: deferred, see docs/testing/onbox-acceptance-register.md
                            — will file in v2
GATE PASSES   rowId="Z9999" Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row Z9999
GATE PASSES   rowId="A101"  Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row A101
```

The last one is the one that matters: **`A101` does not exist** (the real register parses to 62 rows; `rows.has('A101') === false`), and it is the id printed by the gate's own failure output — `helpMessage()`, `scripts/validate-sidecar-acceptance.mjs:184` — and documented at `CONTRIBUTING.md:612`. Feeding **the entire `helpMessage()` string** as a PR body passes the gate. So the honest path is: check goes red → author reads the help → copies the example line → check goes green → nothing was run and no row exists.

This is the same defect pass 1 found for the fenced block, one door over: the documented copy-paste text is itself a passing body. The fence half was closed; this half was not, and it is worse, because a body containing the literal words `NOT RUN` satisfies a required check.

The fix is cheap and the material is already on disk: the workflow does a full `actions/checkout`, and `parseRegisterRows` is already exported from `scripts/check-register-citations.mjs` — read the register, assert the cited row exists. (Secondarily: `helpMessage()`/`CONTRIBUTING.md` should stop advertising an id that is deliberately nonexistent.)

Correctness bug.

### N2 — 🟠 Significant — three more spellings of "inside a code block" still satisfy the gate, and the comment defending the inline-span decision is false as written

`scripts/validate-sidecar-acceptance.mjs:89` (`stripFencedBlocks`) handles only ` ``` ` fences; `:79-84` argues inline spans need no stripping. Both patterns anchor on `^\s*sidecar acceptance:` — and `\s*` absorbs an indent.

Measured, files `['server/tts-sidecar/main.py']`, `RUN` = the documented recorded-run line:

| body | renders on GitHub as | gate |
|---|---|---|
| `Nothing run.` ⏎⏎ `····<RUN>` (4-space indent) | **indented code block** | **PASSES** |
| same with a tab indent | **indented code block** | **PASSES** |
| `~~~` ⏎ `<RUN>` ⏎ `~~~` | **fenced code block** (CommonMark tilde fence) | **PASSES** |
| `Draft, do not read as a record: ` + backtick ⏎ `<RUN>` ⏎ backtick | **one multi-line inline code span** | **PASSES** |
| ` ``` ` fence (pass-1 case) | fenced code block | fails ✔ |
| whole-line inline span | inline code span | fails ✔ |

The register-link form behaves identically (indent and `~~~` both pass).

The last row is the refutation of the comment at `:82-84`: *"An inline span wrapping the WHOLE line is already rejected regardless, because both patterns anchor on `^\s*sidecar acceptance:` and a leading backtick is neither whitespace nor the literal prefix."* That holds only for a span that **opens and closes on the same line**. A span opened by a trailing backtick on the *previous* line leaves the acceptance line starting with the bare prefix, so the anchor matches while the rendered body shows one code span. The stated reason for not stripping inline spans does not survive the case it was written to dismiss — and the sibling this file says it mirrors (`scripts/validate-pr-issue-link.mjs:60-66`) strips them per-paragraph precisely because of that shape.

This is the reviewer brief's shape 2 — a guard that enumerates syntax loses one spelling per round. The fix listed one of the four CommonMark ways to write a code block. Impact is narrower than pass 1's (the documented copy-paste block is closed), but the contract the code and `CONTRIBUTING.md:607` both state — *"plainly (not inside backticks or a code block)"* — is enforced for one spelling out of four.

Correctness bug.

### N3 — 🟡 Minor — `REQUIRED_LEGS` closes the named-leg gap; the gap it was written for is still open one step over

`scripts/tests/workflow-wiring.test.mjs:272` (allowlist) and `:318` (`legs.size >= 11`). The pass-1 finding had two halves; the second was *"both Windows leg steps sit outside every wiring guard and nothing reports the gap."* A four-entry allowlist cannot report a **fifth** leg.

Mutations run against an out-of-repo copy, both on top of the real `verify.yml`:

- **Add a fifth leg with a valid marker and a wrong gate** — `# leg: test:hooks:windows` on a `npm run test:hooks` step gated on `step_test_e2e` (a key that has nothing to do with it): **18/18 pass, exit 0.**
- **Add the same step with no marker at all**: **18/18 pass, exit 0.**

Measured headroom: the file carries 19 parseable `# leg:` and 10 `# supports:` markers against floors of 11 and 9 — **8 leg markers could go unparseable with both floors still green.** Today all 29 derived steps happen to carry a parseable marker, but nothing asserts that invariant, which is precisely what let this PR's first draft ship two markerless legs.

Cheap closure, if wanted: assert the invariant that already holds — every step whose `if:` references `fromJSON(needs.detect.outputs.scopes)` is immediately preceded by a parseable `# leg:`/`# supports:` marker (29 today, 0 exceptions). That is structural rather than an allowlist, and would have caught the original defect without anyone naming the Windows legs.

Correctness bug (in the guard, not the shipped workflow) — minor because the current file is correct and the four legs that matter are pinned.

### N4 — 🟡 Minor, latent + pre-existing — `check-register-citations.mjs` cannot see a citation inside a template literal

`scripts/check-register-citations.mjs:1183-1185` — `extractCitationsByLine` runs `stripInlineCodeSpans` (`:1092`) over **every tracked text file**, not just markdown. In a `.mjs`/`.ts` file every template literal is a backtick-delimited run, so its whole content is blanked before any citation surface is matched (the `SINGLE_ID_SPAN_REGEX` carve-out only rescues a span whose entire content is a bare id). Confirmed by the table under pass-1 finding 2 above: identical text is fatal in a `'…'` string and invisible in a `` `…` `` one.

Concrete failure: a wrong or discharged row id written into a template literal — an error message, a log line, a test fixture — is never checked, on a **fatal** gate whose own long coverage summary enumerates its blind spots and does not name this one. The checker demonstrably does treat source files as citation surfaces (it fired on five `.test.mjs` lines at pass 1, and its current output flags `server/tts-sidecar/tests/test_qwen_duration_bless_gating.py:230`).

Not introduced by this PR, and it is the one item here I would **not** hold the merge for — it plausibly needs a design pass (should source files be scanned under markdown code-span rules at all, or should the code-span blanking be markdown-only?) rather than a one-line fix, and CLAUDE.md's incidental-findings rule is explicit that a named decision is what distinguishes a filed finding from a deferred one. Naming it here so it does not vanish: the decision owed is *"which file types get markdown code-span blanking."*

Correctness bug (latent).

---

## Verdict

The three pass-1 blockers are genuinely fixed, mutation-verified, and CI is green with the new Windows leg actually executing rather than skipping. The two minors and the two documentation-accuracy items are fixed properly, with paired fixtures where a fixture was the right answer.

**Not ready to merge**, for two reasons that are cheap to close:

1. **Pass-1 finding 5 is still open** and is the one that cannot be caught after the fact — `829341a5`'s `Closes #3050` trailer overrides the corrected PR body and will close the issue tracking the undone required-status-check wiring. Amend the trailer.
2. **N1** — a required check that reports green on a body reading `Sidecar acceptance: NOT RUN` is the failure Part 6 exists to prevent, and its own help text is a passing body. This should land before AC5 wires the check into `main`'s ruleset, not after.

**N2** belongs in the same round (it is the unfinished half of pass-1 finding 4, and the comment defending the decision is false). **N3** is a judgement call. **N4** is pre-existing and needs a decision named, not a fix agent.

No fixes were applied. Tree unmodified: HEAD `827eee10bd1a46a5068a4f3f3a4a3db981bf3dce`, `git status --porcelain` empty before and after.

dudarenok-maker:
## PR review — pass 3 (head fc7359fc, depth high)

Final pass under the loop cap. Deltas only — solid items from passes 1 and 2 are not re-listed.

**Precondition met and independently confirmed:** `gh pr checks 3053` → **21/21 green** at this head, including `Windows (test + test:server)` at 4m11s (not vacuous) and `Verify sidecar acceptance is recorded`.

Re-run rather than read:

| repro | result |
|---|---|
| `node --test scripts/tests/validate-sidecar-acceptance.test.mjs` | **67/67** (was 45) |
| `node --test scripts/tests/workflow-wiring.test.mjs` | **19/19** (was 18) |
| `node scripts/check-register-citations.mjs` | **exit 0**, no fatal row from this diff |
| all six N1/N2 exploit bodies from pass 2 | **all now rejected** |
| `helpMessage()` fed whole as a body / whole `CONTRIBUTING.md` / whole PR template | **all rejected** |

**Import side effects — checked, clean.** `import { parseRegisterRows } from './check-register-citations.mjs'` cannot drag CLI behaviour into the workflow: that module's only top-level statement is its `isDirectlyInvoked(import.meta.url)` guard at `:2519`; everything else is imports and declarations (`REPO_ROOT` at `:301` is a `new URL`, not a `git` spawn). Measured: importing it produces no stdout, leaves `process.exitCode` undefined, and takes 10 ms. Its transitive imports (`git-env.mjs`, `lib/read-normalized.mjs`) have no top-level work either. Demonstrated end-to-end by the CLI runs below, which print only gate output.

---

## Pass-2 findings — disposition

### Pass-1 #5 / pass-2 #5 — `Closes #3050` on a partial delivery — ❌ **STILL OPEN (third pass)**

Unchanged at this head. `git log origin/main..HEAD` still shows commit **`829341a5`** ending:

```
Closes #3049
Closes #3050      <-- still here
Refs #2997
```

`gh issue view 3050` → **state `OPEN`**, and its acceptance criterion 5 ("Wired into `main`'s required status checks alongside `verify.yml` and `pr-issue-link.yml`") is the criterion this PR's own "Not done, deliberately" section says is *tracked on #3050*. Repo settings are unchanged (merge-commit only), so `829341a5` lands on `main` verbatim and GitHub closes #3050 from the commit trailer regardless of the corrected PR body. See finding **P0**.

### N1 — the register-link half green on a body that says the tests were not run — ✅ **RESOLVED**

Verified against the real register (62 rows parsed through `parseRegisterRows`), files `['server/tts-sidecar/main.py']`:

```
FAIL  "Sidecar acceptance: NOT RUN. ...register.md has no row yet; blocked ... until Q3."
FAIL  "Sidecar acceptance: deferred, see ...register.md — will file in v2"
FAIL  "Sidecar acceptance: see ...register.md row Z999"
FAIL  "Sidecar acceptance: see ...register.md row A101"   (named: "Cited register row A101 does not exist")
FAIL  "Sidecar acceptance: see ...register.md row a1"     (lowercase id no longer parses)
FAIL  helpMessage() fed whole as a PR body
PASS  "Sidecar acceptance: see ...register.md row A1"     (a row that exists)
```

The split into a case-insensitive prefix and a case-sensitive, `row`/`rows`-introduced id is the right shape, and `hasRegisterLink`'s existence check is pinned by a test that first proves the *syntax* half still matches (`hasRegisterLink(body, new Set([FIXTURE_ROW_ID])) === true`), so the rejection is proven to come from existence and not from the pattern. That is the correct way to write that test. One residual on the `#`-anchor branch — finding **P2**.

### N2 — three more spellings of "inside a code block", and a false comment — ✅ **RESOLVED for the four named spellings**

4-space indent, tab indent, `~~~` fence, and a span opened by a trailing backtick on the previous line all now fail, for both the recorded-run and register-link forms. The fence tracker's char+length rule is correct (a `~~~` line inside a backtick fence is content), and `stripLinesInsideOpenInlineSpan` correctly resets at a blank line (CommonMark forbids a code span crossing one) and correctly does not leak parity out of a fence or an HTML comment — I checked all three. The comment at `:113-122` now states the same-line/previous-line boundary accurately.

Two residuals, in opposite directions: the strip is too narrow in one place (**P3**) and the new `^ {0,3}` anchor is too broad in another (**P1**).

### N3 — the four-entry allowlist could not see a fifth leg — ✅ **RESOLVED for the shapes N3 named**

Mutation-verified against an out-of-repo copy of the tree (baseline 19/19 green):

```
RED (caught)   new derived step, NO marker at all
RED (caught)   new derived step, marker outside the grammar ("# leg: test:hooks (Windows)")
RED (caught)   marker present but an `id:` line sits between `- name:` and `if:`
```

Both mutations N3 explicitly measured as green now fail with a legible message. The anti-vacuity floor is real. Residual — the guard now enumerates a different syntax one level up — finding **P4**.

### N4 — `check-register-citations.mjs` blind to a citation inside a template literal — unchanged, and correctly so

Not touched by this commit. Pass 2's judgement stands: pre-existing, needs a named decision (*which file types get markdown code-span blanking*), not a fix agent. Carries to the user as-is; I am not re-litigating it.

---

## New findings

### P0 — 🔴 **Blocking** — the commit trailer still closes #3050

`git log origin/main..HEAD`, commit `829341a5` — see disposition above. This is the third pass on the same item and it is the only finding here I would hold the merge for, because it is the only one that cannot be corrected after the fact: once `829341a5` is on `main`, #3050 is closed and AC5 (wiring this check into the ruleset) is tracked nowhere. The branch is unmerged, so amending the trailer to `Refs #3050` costs one rebase. Prior incident in this repo: PR #2126 / #2083, 2026-08-05 — body said `Refs`, a lane commit's trailer said `Closes`, the issue closed with six files unswept.

Correctness bug (bookkeeping).

---

### P1 — 🟠 Significant — the new `^ {0,3}` anchor rejects the placement the PR template itself invites, and the failure text blames something else

`scripts/validate-sidecar-acceptance.mjs:207` (`RECORDED_RUN_PATTERN`) and `:242` (`REGISTER_LINK_PREFIX_PATTERN`). This is the first of the two deliberate behaviour changes, and it over-shoots.

The justification in the comment at `:107-109` — *"a line indented 4+ spaces, or by a tab, is a code block, not a paragraph"* — is **false in the two contexts where an author is most likely to indent**. CommonMark does not let an indented code block interrupt a paragraph, and inside a list item four spaces is content indentation, not code. Rendered through **GitHub's own `POST /markdown` (`mode: gfm`)**, not asserted:

| body | GitHub renders the acceptance line as | gate |
|---|---|---|
| `- [ ] Sidecar acceptance: …` in the Test-plan checklist | `<li>` — a list item | **FAILS** |
| `- Ran the suite.` ⏎⏎ `····Sidecar acceptance: …` | `<p>` inside the `<li>` — a paragraph | **FAILS** |
| `Everything ran clean.` ⏎ `····Sidecar acceptance: …` | `<p>…<br>…</p>` — one paragraph | **FAILS** |
| `····Sidecar acceptance: …` after a blank line | `<pre><code>` — genuinely code | fails ✔ (correct) |

Only the last row is a code block. The other three are plain text and are rejected anyway.

**The first row is the one that matters.** `.github/pull_request_template.md:27-33` (added by this PR) puts the guidance immediately above the `## Test plan` checklist, and `CONTRIBUTING.md:604` says only *"Write one of these two lines plainly … in the PR body"* — a `- [ ]` item is "plainly" by any ordinary reading. Measured author-facing output for exactly that body:

```
$ node scripts/validate-sidecar-acceptance.mjs files.txt body.txt
This PR touches server/tts-sidecar/**, which carries 38
...
Record acceptance in the PR body, written plainly (not inside backticks
or a code block), as one of -- filling in the <placeholders>:
...
exit=1
```

The author wrote it plainly and not in a code block, and the only diagnosis offered says otherwise. Nothing in the help text or in CONTRIBUTING.md mentions a list marker, a blockquote, or a bold label — all three fail.

This is also a deviation from the sibling the file says it mirrors: `scripts/validate-pr-issue-link.mjs:12` is `/\b(?:closes|refs)\s+#\d+/i` with **no line anchor at all**, so `- [ ] Closes #123` works there.

The over-rejection is not fixable by loosening the anchor back to `^\s*` — that reopens N2. The shape that actually holds is to strip *genuine* indented code blocks as a block pass (a 4+-space line that follows a blank line and is not inside a list), the way fences are already stripped, and leave the anchor tolerant of a list marker and blockquote prefix. At minimum the help text must name the real cause.

Correctness bug.

### P2 — 🟠 Significant — the `#`-anchor branch accepts the spelling GitHub never produces and rejects the one this repo documents

`scripts/validate-sidecar-acceptance.mjs:243-244`:

```js
/onbox-acceptance-register\.md(?:#([A-Z]\d{1,3})\b|\b.*\brows?:?\s+([A-Z]\d{1,3})\b)/
```

The `#` branch is case-**sensitive** `[A-Z]`. GitHub lowercases heading anchors, and this repo's own citation checker says so in as many words — `scripts/check-register-row-citations.mjs:26-28`: *"a markdown link into onbox-acceptance-register.md's own anchor for a row heading (`onbox-acceptance-register.md#a41`/`#e3`, **case-insensitive — GitHub lowercases heading anchors**)"*, with `scripts/tests/check-register-row-citations.test.mjs:114` using `#a1`.

Measured, files `['server/tts-sidecar/main.py']`, against the real 62-row register:

```
FAIL  Sidecar acceptance: see [A1](docs/testing/onbox-acceptance-register.md#a1)   <- the documented form
PASS  Sidecar acceptance: see [A1](docs/testing/onbox-acceptance-register.md#A1)   <- a form GitHub never emits
```

So the branch added to support anchor links supports only the spelling that does not occur. The paired fixture at `scripts/tests/validate-sidecar-acceptance.test.mjs:168` (`…register.md#C2`) is uppercase too, so the test set never touches the real shape — reviewer-brief shape 11, one door over: the fixture encodes a belief about GitHub's anchor form rather than the form the repo has already recorded.

Related, same line: the `row`/`rows` branch requires the id to come **after** the filename, so the two most natural link forms are rejected —

```
FAIL  Sidecar acceptance: see row A1 in docs/testing/onbox-acceptance-register.md
FAIL  Sidecar acceptance: tracked on [row A1](docs/testing/onbox-acceptance-register.md)
```

— and the second is the repo's own `Register row: [A1](onbox-acceptance-register.md#…)` label idiom (`docs/testing/fs38-wave3-onbox-acceptance.md:16`, `docs/testing/ort-marker-onbox-acceptance.md:19-21`). Making the `#` branch case-insensitive (uppercasing before `rowIds.has`) is safe and does not reopen N1: the literal `#` immediately before the id is the disambiguator that ordinary prose lacks. The ordering constraint is a separate, cheaper widening.

Correctness bug.

### P3 — 🟠 Significant — a doubled-backtick span still satisfies the gate; the comment says it cannot

`scripts/validate-sidecar-acceptance.mjs:169-184`, comment at `:164-165`: *"A run contributes its own length, so a doubled-backtick span stays even."*

That reasoning holds for a doubled span that **closes** on its own line. It fails for one that **opens**: an even-length unmatched run leaves parity even, so the counter reports "no span open" while CommonMark has one open, looking for the next equal-length run.

Exact input → wrong output, files `['server/tts-sidecar/main.py']`:

```
Nothing was run. ``
Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row A1
``
```

- `passesSidecarAcceptanceGate(...)` → **`true`**, CLI exits 0.
- Rendered through GitHub's `POST /markdown`, `mode: gfm`:
  `<p dir="auto">Nothing was run. <code class="notranslate">Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row A1</code></p>`

The acceptance line is inside `<code>` — the exact rendering N2 exists to reject — and the gate is green on a body whose visible words are "Nothing was run."

This is narrower than N2's cases: nothing documented produces it, so it takes a deliberate author rather than honest copy-paste. But `` `` `` is the standard CommonMark idiom for a span containing a backtick, so it is not exotic, and it is the reviewer brief's shape 2 exactly — *a guard that enumerates syntax loses one spelling per round*, now on the third round. Character parity is the wrong model; the fix is to track unmatched **runs by length** (a run of length *n* opens until a later run of length *n* closes it), which is what CommonMark actually specifies and which also makes the "doubled-backtick span stays even" claim true instead of approximately true.

Correctness bug.

### P4 — 🟡 Minor — the new "every derived step carries a marker" guard enumerates the single-line `if:` spelling

`scripts/tests/workflow-wiring.test.mjs:326` — `/^\s*if: .*fromJSON\(needs\.detect\.outputs\.scopes\)/` — and the floor at `:340`.

Mutations against an out-of-repo copy, each adding one new derived step to the Windows job with **no marker at all**:

```
RED   (caught)                 if: fromJSON(needs.detect.outputs.scopes).step_test_e2e
RED   (caught)                 if: ${{ fromJSON(needs.detect.outputs.scopes).step_test_e2e }}
RED   (caught)                 if: 'fromJSON(needs.detect.outputs.scopes).step_test_e2e'
GREEN (unmarked leg accepted)  if: ${{ fromJSON( needs.detect.outputs.scopes ).step_test_e2e }}
GREEN (unmarked leg accepted)  if: >-
                                 fromJSON(needs.detect.outputs.scopes).step_test_e2e ||
                                 fromJSON(needs.detect.outputs.scopes).shared
```

Both green shapes are valid GitHub Actions YAML, and the folded one is what a author naturally reaches for once a condition passes ~110 characters — which every existing one already does. A step written that way never enters `derived`, so it is neither counted nor marker-checked, and `derived.length >= 25` does not notice because the count stays at 29.

The guard's own comment (`:311-316`) claims *"A new leg with no marker … fails HERE, whether or not anyone remembers to add it to REQUIRED_LEGS"* — true for the marker grammar, not true for the `if:` grammar. Same class as N3, one level up. Secondary: the floor's headroom is now 4 (29 vs 25), better than the 8 it criticises but not zero.

Cheap structural closure, if wanted: assert that **every** `if:` line in `verify.yml` is either `always()` or matches the derived regex — i.e. forbid the shapes the scan cannot parse, rather than widening the scan to chase them. That invariant holds today (30 `if:` lines, 29 derived + `if: always()` at `:635`) and is one assertion.

Correctness bug (in the guard, not the shipped workflow) — minor, because the current file is correct and the four legs that matter are pinned by `REQUIRED_LEGS`.

### P5 — 🟡 Minor — `loadRegisterRowIds`'s fail-closed path reports a false cause

`scripts/validate-sidecar-acceptance.mjs:33-39` and `:320-326`.

Failing closed is the right call and the comment argues it well. But the CLI's diagnosis does not distinguish "you cited a row that does not exist" from "I could not read the register", and prints the former for both. Measured by running the real CLI from a tree where `docs/` is absent, body = `Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row A1` (a row that **does** exist):

```
Cited register row A1 does not exist in docs/testing/onbox-acceptance-register.md.
[...help text...]
exit=1
```

The statement is false and sends the author to change a correct line. Trigger is narrow — `actions/checkout@v7` on a `pull_request` event checks out the merge ref, so the register is present unless this same PR moves or renames it, or `parseRegisterRows` throws on it — but the `catch {}` at `:36` also swallows a genuine parser exception into the same wrong message. One `readFileSync` failure flag threaded to the CLI, or a distinct message when `rowIds.size === 0`, closes it.

Cleanup nit.

---

## Verdict

**One blocker: P0.** The `Closes #3050` trailer on `829341a5` is on its third pass, is the only finding here that cannot be corrected after merge, and costs one amend. Everything else is fixable later without loss.

**P1, P2 and P3 must land before acceptance criterion 5 wires this check into `main`'s ruleset** — that is the author's own stated bar for N1, and it applies with more force here, because these are false *negatives*: P1 rejects the placement this PR's own template invites and then misdiagnoses it, and P2 rejects the anchor form the repo has already documented. While the check is advisory a false negative costs an author a confusing red X; as a required check it costs a blocked merge on a correct body. P3 is the smallest of the three and is the one I would let slip if only two fit.

**P4** is a judgement call on a guard that is materially better than it was. **P5** is a nit. **N4** carries to the user unchanged, with the decision it needs already named.

For what it is worth on the substance: the three fixes in `fc7359fc` are real, are mutation-verified in both directions, and the test file is unusually honest — it guards its own guards (`assert.equal(hasRegisterLink(body, new Set([FIXTURE_ROW_ID])), true)` before asserting the existence rejection; the staleness assertion on `FIXTURE_ROW_ID`; the anti-vacuity floor on the derived-step scan; the `honestBodiesWithCodeElsewhere` set testing the false-positive direction). What P1 and P3 show is that that set was drawn from the shapes the author had already thought of, not from a model of CommonMark — which is why a fourth spelling and an over-rejection both survived it.

No fixes were applied. Tree unmodified: HEAD `fc7359fcca077ee4364ff14c670b3b55ed6366ad`, `git status --porcelain` empty before and after.

PR #3054, opened 2026-09-06
fix(ops): allocate worktree port slots from claims, not from a count

### LINKED ISSUE #3052: wt-new.mjs assigns colliding port slots — 10 of 17 live worktrees share a slot
## What

`scripts/wt-new.mjs` assigns each new worktree's port slot from a **live count** of existing worktrees:

```js
// Existing worktrees include the main one, so count == slot for the new tree.
const slot = countExistingWorktrees();
```

That comment holds only if worktrees are **never removed**. This repo removes them routinely — 28 orphaned trees were torn down on 2026-09-06 alone — so every teardown lowers the count and the next `wt-new` re-issues a slot a surviving tree already holds.

## Observed, 2026-09-06

Ten of the seventeen live worktrees are on a **shared** slot:

| Slot | Worktrees holding it |
|---|---|
| 15 | `wt-2card-pinokio-batch`, `wt-3044-dispatch-write-verification`, `wt-3047-stale-battery-reaper` |
| 16 | `wt-analyzer-render-batch`, `wt-3049-windows-leg-sidecar-gate` |
| **17** | `wt-mechanical-batch-2`, `wt-2192-qwen-install-sidecar-stop`, `wt-3004-analysis-rejoin-signal`, `wt-3051-worktree-gc` |
| 19 | `wt-fs38-a1-wave12`, `wt-human-checkpoint-batch` |

Four trees are all assigned `PORT=8250`. The highest slot in use is 20 against 17 linked worktrees, so the counter has drifted both up and down as trees came and went.

## Why this is a data hazard, not an annoyance

The whole point of the per-slot `.env.local` / `server/.env` pair is isolation, and CLAUDE.md's "Worktree setup" step 5 already spells out the failure when it is absent: a worktree gets "a frontend that silently proxies `/api` to whatever's already listening on `:8080` (typically the primary checkout's server and workspace) while its own correctly-isolated server sits idle."

A **collision** produces the same class of failure between two lanes rather than against the primary — and `WORKSPACE_DIR` collides with the ports, so two servers can share one `cast.json`/`state.json`. That is the exact "two writers, one workspace" shape this repo has repeatedly been bitten by.

Nothing surfaces it. Both trees look correctly configured in isolation; only comparing them across trees reveals it.

## Fix shape

Allocate the **lowest slot not already claimed**, by reading the `worktree slot <n>` marker `renderEnvLocal` already writes into every generated `.env.local`, rather than counting trees. Behaviour is not in question — do not re-issue a slot a live worktree holds — so this needs no design pass.

Two details worth getting right:

- A worktree with **no** `.env.local` (created by `EnterWorktree` or Agent `isolation: "worktree"`, which never write one) claims no slot, so it must not block allocation — but it also cannot be detected, which is a known and acceptable limit worth a comment.
- Slot 0 is the primary checkout and stays reserved.

**The existing collisions are not repaired by this change.** Say so in the PR; re-slotting live trees is a separate, manual call since some are mid-work.

## Acceptance

1. Allocation reads claimed slots and returns the lowest free one; a unit test proves that removing a middle worktree and creating a new one does **not** re-issue the surviving tree's slot.
2. A test that a tree lacking `.env.local` does not block allocation.
3. Both tests assert against input that would otherwise make the guard fire, and each is mutation-verified — delete the token, watch that named test go red.
4. `scripts/tests/wt-new.test.mjs` stays green.

## Key files

- `scripts/wt-new.mjs` — `countExistingWorktrees()`, `computePorts()`, `renderEnvLocal()`, `main()`
- `scripts/tests/wt-new.test.mjs`
- `CLAUDE.md` → "Worktree setup" step 5

## Benefit

*(technical)* Restores the per-worktree isolation the generated env files exist to provide, closing a silent route by which two lanes share a server, a TTS sidecar and a workspace.

dudarenok-maker:
## Summary

`scripts/wt-new.mjs` derived each new worktree's port slot from a **live count** of existing worktrees, with the comment *"Existing worktrees include the main one, so count == slot for the new tree."*

That holds only if worktrees are never removed. They are removed constantly here — 28 orphans were torn down on 2026-09-06 — so every teardown lowered the count and the next run re-issued a slot a surviving tree still held.

**Ten of seventeen live worktrees were sharing a slot** when this was found, four of them on `PORT=8250`, and the highest slot in use was 20 against 17 trees: the counter had drifted both up and down.

### The hazard, stated correctly

An earlier version of this body claimed `WORKSPACE_DIR` collides with the ports. **That was wrong**, and a review pass caught it: `WORKSPACE_DIR` is the constant `../castwright-workspace` (`scripts/wt-new.mjs:228`), resolved relative to each tree's own `server/`, so workspaces were always isolated.

The real route is the proxy. `vite.config.ts` resolves its `/api` target from `VITE_API_PORT ?? PORT`, so on a shared slot whichever server binds first owns the port and the other lane's frontend proxies into **that** server — and thence into that server's workspace. Two lanes still end up operating on one `cast.json`/`state.json`; they arrive there through the proxy, not through a shared `WORKSPACE_DIR`. The distinction matters, because isolating `WORKSPACE_DIR` would have fixed nothing.

## What changed

**Allocation reads claims instead of counting.** The lowest unclaimed slot wins, read from the `worktree slot N` marker `renderEnvLocal()` writes into every generated env file. `countExistingWorktrees()` is removed.

**Both claim files are scanned and unioned.** `.env.local`'s own generated header says *"Safe to edit; safe to delete"* — so the single file the original scan trusted is the one the tool invites you to delete, while `server/.env` carries the identical marker *and* the ports that matter. `server/.env` wins keys both carry. The asymmetry decides the direction: a false positive wastes a slot number, a false negative restores the data hazard.

**The reader does not trust line 1.** It scans the leading comment block, stopping at the first `KEY=value`, so a prepended lane note or a blank line cannot surrender a live claim. The original comment asserted these files are never hand-edited — contradicted by the file's own header and by CLAUDE.md, which instructs hand-writing them.

**The primary checkout is exempt from the silent-claimant count.** Every real clone's primary has a hand-written, marker-less `server/.env`, so without this the new warning fired on 100% of real runs — and its text was false for that tree besides, since slot 0 can never be issued. It is identified by a fact about the tree rather than its list position: a linked worktree's `.git` is a **file** containing `gitdir:`, the primary's is a **directory**.

**`main()` reports the scan** — trees scanned, slots claimed — and warns when a *linked* tree has a generated env file but no readable marker, so a silent skip is visible.

- **`scripts/wt-list.mjs` carried the same bug this PR removes.** `buildRows` used the array index as the slot, so the diagnostic tool for #3052 reported slots 0-16 all distinct while its own `PORT` column showed 8250 on three trees. It now reads the real marker.
- `readPortsFor` also read only `.env.local`, printing `PORT=(default)` for a tree whose server binds 8120. It uses the same shared reader.
- The marker reader lives in a new leaf module, `scripts/lib/worktree-slot.mjs`: `wt-new.mjs` already imports from `wt-list.mjs`, so putting it in either would have closed an import cycle.
- Dead defensive branches, a dead fixture call, a stale usage header, and the unused `relative`/`path` fields this change orphaned are all removed; the allocation-to-claim-write TOCTOU is documented at the site.

**Existing collisions are NOT repaired.** Re-slotting a live tree is a manual call and several are mid-work.

## Test plan

`scripts/tests/wt-new.test.mjs` — **86 tests, 86 pass**. `npm run test:hooks` — **1992 pass, 0 fail**. ESLint clean on all four changed files.

### The first version's tests could not fail

Worth recording, because it is this repo's recurring failure. The initial round shipped 12 tests that **reimplemented the allocation loop inside each test body** and asserted on their own local variable; `allocateNextSlot` and `findClaimedSlots` were imported but never called. Replacing `allocateNextSlot`'s entire body with `return 999;` left **all 63 green**.

Fixing that with a default-parameter seam then created a *second* blind spot: every test passed an explicit argument, so the zero-argument path production actually uses was exercised by nothing, and four mutations that disable the fix entirely shipped **70/70 green** — including the `main()` call site replaced by a constant. All four are worse than the bug being fixed, since the old count rule at least varied the slot.

Closing it needed **integration tests against a real throwaway git repo**: real `git worktree add` trees whose env files come from the production `renderEnvLocal`/`renderServerEnv`, git spawned through `scrubGitEnvForThrowawayRepo`, and cwd plus every `GIT_*` var restored in a `finally`, so the no-argument path and `main()` itself are exercised.

### Mutation coverage

Each mutation applied to production source, the named test run alone, then restored and re-run green.

| Mutation | Named test that reddened |
|---|---|
| `allocateNextSlot` returns `claimed.length` (**the old count rule**) | does not re-issue a surviving worktree slot after teardown (#3052) |
| `allocateNextSlot`'s default argument emptied | findClaimedSlots() and allocateNextSlot() read the live worktree list with NO arguments (#3052) |
| the `main()` call site replaced by a constant | wt-new main() stamps the allocated slot into the worktree it creates (#3052) |
| `isPrimaryCheckout` forced false | does not warn about the primary checkout's hand-written server/.env (#3052) |
| slot counter starts at 0 instead of 1 | allocateNextSlot never hands out slot 0, reserved for the primary checkout |
| marker read from the whole file instead of the leading comment block | extractSlotFromEnvLocal only reads the first line, not a later marker |
| `Set` accumulator replaced by an array | findClaimedSlots de-duplicates two worktrees claiming the same slot |
| claim-file precedence reversed | wt-list port columns read server/.env when both claim files carry different values |
| the fixture's primary `server/.env` removed | the exemption's own fixture assertion |

Two findings from running that campaign honestly. The first harness reported "all 15 survived" because it decoded node's UTF-8 summary as cp1252 and matched nothing — a parse-failure guard was added so a non-parsing run can never read as a pass again. And one mutation genuinely survived: starting the slot counter at 0 left its test green, because that test's two cases gave the same answer from either start. It was widened until it kills.

Before-shipping steps not applicable: **1** (localized fix — the issue plus its tests are the spec), **3** (no hardware-provable behaviour), **4** (no plan), **5** (dev tooling; none of the changed files is in `build-release-zip.mjs`'s ship allowlist).

Closes #3052


dudarenok-maker:
## PR review — pass 1 (head eed18a3e, depth medium)

Scope: `scripts/wt-new.mjs`, `scripts/tests/wt-new.test.mjs` (+ the sibling `scripts/wt-list.mjs` the change surfaces) · verified: `npm run test:hooks` **1976/1976 green**, `node --test scripts/tests/wt-new.test.mjs` **70/70 green**, and **9 independent mutations** applied to a scratch copy of `scripts/` (never to the worktree — tree clean before and after) before adversarially probing the allocation contract, the `lines[0]` premise, and the mutation campaign the PR body reports.

---

### 🔴 Blocking — correctness bug: the "20 mutations, 20 kills" campaign never mutated the wiring, and **four mutations that reinstate #3052 in its worst form survive 70/70 green**

The rewrite genuinely fixed the *pure core* — I reproduced that (see ✅ below). But the campaign mutated only the core. The **composition root** — the default parameters and the `main()` call that connect the core to production — is covered by no test at all, because every test passes an explicit argument and nothing ever exercises the zero-arg path.

Applied one at a time to `scripts/wt-new.mjs`, suite re-run after each:

| # | Mutation | `file:line` | Production effect | Suite |
|---|---|---|---|---|
| A | `allocateNextSlot(claimed = findClaimedSlots())` → `allocateNextSlot(claimed = [])` | `wt-new.mjs:160` | every new worktree gets slot 1 | **70 pass / 0 fail** |
| B | `findClaimedSlots(worktreePaths = listWorktreePaths())` → `= []` | `wt-new.mjs:129` | nothing ever claims a slot → every tree gets slot 1 | **70 pass / 0 fail** |
| C | `const slot = allocateNextSlot();` → `const slot = 1;` | `wt-new.mjs:349` | the entire fix bypassed | **70 pass / 0 fail** |
| D | `return parseWorktreePorcelain(porcelain).map((tree) => tree.path);` → `return [];` | `wt-new.mjs:115` | no worktrees scanned → every tree gets slot 1 | **70 pass / 0 fail** |

Each of these is strictly *worse* than the bug being fixed: the old count rule at least varied the slot, whereas A–D hand slot 1 to every worktree forever. All four ship green.

Concretely for C — delete the call the whole PR exists to introduce, and the suite that was rewritten specifically to catch "tests that cannot fail" reports 70/70.

This matters beyond the usual test-gap argument because the PR body's central merge argument is that the first version's tests could not fail, that they were rewritten, and that a 20-mutation campaign scored 20 kills. That claim is materially incomplete: the campaign's blind spot is *exactly the seam this PR introduced*. #3052's Acceptance criterion 1 ("Allocation **reads claimed slots** and returns the lowest free one") is unproven — only the "returns the lowest free one" half is under test.

The behaviour is agreed (the wiring must be regression-guarded), so this is an implementation choice, not a design pass — e.g. hoist the composition into an injectable (`resolveSlot({ listPaths })`) that `main()` calls and a test can drive against a fixture, or add a source-scan guard in the shape this repo already uses (`entry-point-guard-convention.test.mjs`, `cast-lock.guard.test.ts`).

---

### 🟠 Significant — correctness bug: `wt-list.mjs` still prints the array index as "slot"; the class this PR fixes is left armed in the repo's own diagnostic for it

`scripts/wt-list.mjs:87` — `const rows = trees.map((tree, slot) => {` — the `slot` column is the enumeration index, i.e. the *same* "slot == position in the worktree list" premise this PR removes from `wt-new.mjs`.

Repro, run just now against the live 17-worktree state (paths redacted; this is `node scripts/wt-list.mjs` output):

```
slot   branch                                       PORT
1      fix/ops-2192-a29-genuine-install-test        8280
2      fix/server-2192-qwen-install-sidecar-stop    8250   <--
5      docs/docs-2card-pinokio-batch                8230   <--
6      ci/ops-3047-stale-battery-reaper             8230   <--
7      ci/ops-3049-windows-leg-sidecar-gate         8240   <--
8      chore/ops-3051-worktree-gc                   8250   <--
9      fix/ops-3052-worktree-slot-allocation        8260   <--  (this PR's own tree)
10     fix/server-3056-sibling-installer-clobber    8240   <--
11     chore/ops-analyzer-render-batch              8240   <--
12     fix/server-fs-2054-clone-gate-uuid           8260   <--
13     chore/ops-fs38-a1-wave12                     8270   <--
14     docs/docs-human-checkpoint-batch             8270   <--
15     docs/docs-mechanical-batch-2                 8250   <--
```

The `slot` column reads `0..16`, all distinct, implying zero collisions — while its own `PORT` column three columns to the right shows 8250 held by three trees, 8240 by three, and 8230 / 8260 / 8270 by two each. The tool a maintainer reaches for to *diagnose* #3052 actively denies #3052 exists.

The true slot is now one call away: `extractSlotFromEnvLocal` is exported from `wt-new.mjs`, and `readPortsFor` already reads that exact file (`wt-list.mjs:60-66`). Per CLAUDE.md's incidental-findings rule this is a defect the work surfaced, so it is in scope for this round rather than a follow-up ticket.

---

### 🟠 Significant — correctness bug: the scan reads `.env.local`, the one file whose own generated header says **"safe to delete"** — while `server/.env`, which carries the identical marker *and* the ports that matter, is ignored

`renderServerEnv` (`wt-new.mjs:212-214`) writes the **same** `# Generated by scripts/wt-new.mjs — worktree slot N (branch …)` marker as line 1 of `server/.env`, and that is the file holding `PORT`, `LOCAL_TTS_PORT` and `WORKSPACE_DIR`. `findClaimedSlots` (`wt-new.mjs:133`) reads only `.env.local` — whose own last header line (`wt-new.mjs:191`) tells the reader:

> `# Safe to edit; safe to delete (defaults from vite.config.ts / sidecar apply).`

Repro (executed, output verbatim), building a tree exactly as `main()` does then taking the file at its word:

```
before delete: claimed = [ 1 ] -> next slot 2
after  delete: claimed = []    -> next slot 1
live tree server/.env still pins PORT=8090, LOCAL_TTS_PORT=9010
new  tree would get             PORT=8090, LOCAL_TTS_PORT=9010
COLLIDE: true
```

Deleting `.env.local` is safe for Vite (the file's claim is true in that narrow sense) but silently surrenders the slot claim while the tree's server and sidecar remain bound to it — reinstating the exact collision this PR closes, with no warning anywhere.

Compounding it, `findClaimedSlots` swallows all three failure modes silently (`:134-137` no file, `:145-148` read/parse error, `:141` no marker) and `main()` prints only the slot it chose (`renderLaunchBlock`, `wt-new.mjs:289`) — never how many trees were scanned or how many yielded no claim. A total detection failure is therefore indistinguishable from "slot 1 is genuinely free" (rubric shape 6, success reported while doing nothing). Scanning `server/.env` as a second source, and printing the claimed set, are both cheap.

---

### 🟠 Significant — correctness bug: `wt-new.mjs:100` states a premise the same file contradicts 91 lines later, and the failure is silent re-collision

> `// These files are never edited by hand or prepended to by external tools, so the first-line guarantee holds.`

`wt-new.mjs:191` says the opposite (`Safe to edit`), and CLAUDE.md's "Worktree setup" step 5 instructs the operator to hand-write `.env.local` for tool-made worktrees. Repro (executed):

```
D prepended line  claimed= [ 2 ] next= 1   <-- slot 1 IS live but re-issued
E leading blank   claimed= [ 2 ] next= 1
F UTF-16LE        claimed= [ 2 ]
```

Input: two live trees at slots 1 and 2, the slot-1 tree's `.env.local` given one prepended comment line (`# lane: keep this tree, mid-work`). Output: `findClaimedSlots` → `[2]`, `allocateNextSlot` → `1`, colliding head-on with the live tree. A leading blank line does the same; so does a re-save as UTF-16LE (Windows PowerShell 5.1 `>` redirection default), since `readFileSync(…, 'utf8')` mojibakes it.

Note the risk is asymmetric and the current design has it backwards: a *false positive* (matching a stray "worktree slot 9" further down, which `wt-new.mjs:104-107` and its test at `:432` are built to prevent) merely wastes a slot number; a *false negative* re-creates the data hazard. Scanning the leading comment block rather than `lines[0]`, or warning when a `.env.local` exists but yields no slot, fixes the asymmetry without loosening anything that matters.

---

### 🟡 Minor

1. **An assertion that cannot fail, cited in the PR body as load-bearing.** `scripts/tests/wt-new.test.mjs:557` — `assert.deepEqual(computePorts(paths.length), computePorts(3))`. `paths.length === 3` by construction (`:540`, `makeWorktreeFixture(t, [0, 3, 4])`), so this compares `computePorts(3)` to itself. Proven inert: I replaced `computePorts`'s body with `return { ...BASE_PORTS };` (slot ignored entirely — the strongest possible violation of the property the comment at `:555-556` claims it demonstrates) and this line **passed**; the failure came from `:558`'s `notDeepEqual`, which already carries the whole claim. The PR body cites `:557` as what "closes the loop"; it closes nothing.
2. **Stale usage header the change made false.** `scripts/wt-new.mjs:16` still reads `// 2. Picks a port-offset slot from existing worktree count (slot 0 = main).` — the file's own front-door documentation of the rule this PR removed. A chore the work made owed, not taste.
3. **Two dead defensive branches.** `wt-new.mjs:104` `if (lines.length === 0) return null;` is unreachable (`''.split(/\r?\n/)` → `['']`, length 1; the empty-content test passes via the regex, not this branch). `wt-new.mjs:141` `typeof slot === 'number'` is redundant — `extractSlotFromEnvLocal` returns `number | null` and `slot !== null` already settles it. Removing both leaves 70/70 green (verified). CLAUDE.md "Simplicity first": no error handling for impossible scenarios.
4. **Dead fixture call.** `scripts/tests/wt-new.test.mjs:469` — `makeWorktreeFixture(t, []);` discards its return value and the next line passes a literal `[]`. It builds and tears down an empty temp dir for nothing.
5. **The PR body and commit message state a mechanism the code does not have.** Both assert "`WORKSPACE_DIR` is assigned off the same slot as the ports, so a collision lets two lanes run one server against one workspace — the same `cast.json` and `state.json`." `WORKSPACE_DIR` is the constant `'../castwright-workspace'` (`wt-new.mjs:210`), identical for every slot and resolved *inside* each worktree — verified: slots 1 / 2 / 7 all emit `WORKSPACE_DIR=../castwright-workspace`, with only `PORT` and `LOCAL_TTS_PORT` varying. Two colliding trees get **different** workspace directories. The hazard is real by the other route the issue's own opening paragraph describes correctly (port collision → the second tree's Vite proxies `/api` to the first tree's server → the first tree's workspace), so the fix's justification survives — but the commit message is permanent, and as written it is wrong.
6. **TOCTOU between allocation and claim.** `wt-new.mjs:349` allocates, `:353` runs `git worktree add` (seconds), `:356` writes `.env.local`. Two concurrent `wt-new.mjs` runs both allocate before either claims, and both get the same slot — in a repo whose documented use case is multiple parallel sessions. Not a regression (the count rule raced identically) and not something this PR must close, but it is the residual half of the guarantee and deserves a line in the code rather than silence.

---

### ✅ What is solid

- **The "tests cannot fail" defect is genuinely fixed — reproduced independently.** Replacing `allocateNextSlot`'s body with `return 999;` gives **63 pass / 7 fail**, reddening 7 tests by name including `allocateNextSlot does not re-issue a surviving worktree slot after teardown (#3052)`. The PR's account of the first version is accurate and the rewrite does what it claims for the core.
- **The widened slot-floor test does kill its mutation.** `let slot = 1` → `let slot = 0` reddens `allocateNextSlot never hands out slot 0, reserved for the primary checkout` (plus 6 others). The `[[], [0], [0,1], [1,3], [0,2,3], [4]]` sweep at `:496` is what does it — the PR's account of that near-miss is correct and the fix is real.
- **The ascending / de-duplicated contract is honoured on every reachable path.** I could not construct a violating input that reaches `allocateNextSlot` in practice: `findClaimedSlots` de-dupes through a `Set` and sorts numerically (`:150`), `(\d+)` cannot yield `NaN`, a negative, or a non-integer, and the only production caller (`:349`) uses the default. The unsorted-input hazard (`allocateNextSlot([3, 1])` → `1`, already claimed) is real but unreachable; documenting it rather than sorting defensively is the right call here.
- **BOM and CRLF do *not* defeat `lines[0]` — the PR under-claims its own robustness.** Verified: a UTF-8 BOM prepended to a canonical file still yields `claimed=[1,2] next=3` (the regex is unanchored, so the BOM sits harmlessly before the match), and CRLF is handled by the `/\r?\n/` split. Only a prepended *line* breaks it (see above).
- **Correct on the live machine.** `listWorktreePaths()` sees 17 worktrees; 16 carry `.env.local` and **all 16** yield a readable marker; `findClaimedSlots()` returns `5,7,14,15,16,17,18,19,20` and `allocateNextSlot()` returns **1**. The one tree without `.env.local` is the primary checkout, which claims nothing — correct, since slot 0 is reserved and allocation starts at 1 regardless.
- **The injection seam increased coverage rather than hiding behaviour.** The parsing the old `countExistingWorktrees()` did inline is now delegated to `parseWorktreePorcelain`, which already has two tests (`:252`, `:271`); the old function itself had none. Only the trivial `.map(tree => tree.path)` is uncovered — the real gap is the wiring, not the extraction.
- **House gates.** `Closes #3052` present, plain, outside any code span. On-box N/A is correct (deterministic Node script, no hardware-provable behaviour). Release-notes N/A is defensible and I checked it rather than taking it on trust — `scripts/wt-new.mjs` is absent from `build-release-zip.mjs`'s ship allowlist, so it is maintainer-only tooling with no shipped delta. `countExistingWorktrees` is fully removed with zero remaining references tree-wide (`git grep`).
- Full `npm run test:hooks`: **1976 pass / 0 fail**.

---

### Verdict

**Request changes.** The allocation logic is correct and demonstrably works against the real 17-worktree state — this is the right fix for #3052 and it should land. What blocks it is that the fix's production path has no regression guard: four one-line mutations that give every future worktree slot 1 all ship 70/70 green, and the PR's stated mutation coverage does not extend to the seam the PR itself introduced. Given #3052 exists because nothing detected a slot collision for months, merging its fix with an unguarded composition root repeats the shape.

Alongside that: `wt-list.mjs:87` still hard-codes the index-as-slot premise and will keep reporting 17 distinct slots over a colliding fleet; and the claim scan reads only the file whose own header invites its deletion, while the marker-bearing `server/.env` that actually pins the ports goes unread.

dudarenok-maker:
## PR review — pass 2 (head 589e621c, depth medium)

Deltas only. Verified: `node --test scripts/tests/wt-new.test.mjs` **80/80**, full `npm run test:hooks` **1986/0**, `npx eslint` clean on all four touched files, cloud CI green at this head. **12 mutations** applied one at a time to a scratch copy of `scripts/` (never the worktree — HEAD `589e621c` and an empty `git status --porcelain` before and after; `git worktree list` 18 rows before and after).

---

### Pass-1 dispositions

**🔴 Blocking (four mutations disabling the fix shipped 70/70 green) — FIXED.** Re-run independently, one at a time, suite re-run after each:

| Mutation | `file:line` | Result |
|---|---|---|
| `allocateNextSlot(claimed = findClaimedSlots())` → `= []` | `wt-new.mjs:163` | **79/1** — `findClaimedSlots() and allocateNextSlot() read the live worktree list with NO arguments (#3052)` |
| `findClaimedSlots(worktreePaths = listWorktreePaths())` → `= []` | `wt-new.mjs:152` | **79/1** — same test |
| `collectSlotClaims(worktreePaths = listWorktreePaths())` → `= []` (the default `main()` now actually uses) | `wt-new.mjs:136` | **79/1** — `wt-new main() stamps the allocated slot into the worktree it creates (#3052)` |
| `const slot = allocateNextSlot(claims.slots);` → `const slot = 1;` | `wt-new.mjs:362` | **79/1** — same test |
| `return parseWorktreePorcelain(porcelain).map(…)` → `return [];` | `wt-new.mjs:113` | **78/2** — both integration tests |

The integration fixture is safe on all four counts asked. `gitAt` (`wt-new.test.mjs:703-716`) spawns every git through `scrubGitEnvForThrowawayRepo()`, so an inherited `GIT_DIR`/`GIT_INDEX_FILE` cannot redirect it; `main()`'s own `gitOrThrow` uses the narrower `scrubGitEnv`, but `inRepo` (`:750-766`) has already deleted every `GIT_*` from `process.env` by then, and restores cwd + env in a `finally`; `repoRoot()` resolves from the chdir'd fixture, and `main()`'s `resolve(root, '..', 'wt-<slug>')` lands inside the same `mkdtemp` root the `t.after` removes. Confirmed empirically — the real repo's `git worktree list` is 18 rows before and after a full `test:hooks` run.

Two further mutations on the leaf module also kill: `SLOT_CLAIM_FILES` → `['.env.local']` (**78/2**), `present++` → no-op (**79/1**), `if (slot !== null) slots.add(slot)` → unconditional (**78/2**).

**🟠 `wt-list.mjs:87` index-as-slot — FIXED.** `buildRows` now reads `readSlotClaims`; restoring the index (`slots.join('/')` → `String(trees.indexOf(tree))`) reddens **78/2**. But see 🟠 N2 below — the same table's PORT columns still carry the old premise.

**🟠 Scan read only `.env.local` — FIXED.** Unioned across `SLOT_CLAIM_FILES`; dropping `server/.env` reddens 2 by name. The union cannot *lose* a claim (a `Set` union with no removal path, `worktree-slot.mjs:67-86` / `wt-new.mjs:136-149`); one-file-only and disagreeing-file cases are both tested. The surfacing mechanism added alongside it is where 🟠 N1 lives.

**🟠 False "never edited by hand" comment — FIXED.** Comment replaced (`worktree-slot.mjs:31-47`), scan now walks the leading comment block and stops at the first `KEY=value`. Removing that stop reddens `extractSlotFromEnvLocal ignores a marker phrase below the settings, not just below line 1`. I could not defeat the reader with the inputs named in the brief: a marker inside a value is unreachable (the `KEY=value` stop runs *before* the regex, `worktree-slot.mjs:55-56`); CRLF is handled by the split; a UTF-8 BOM is stripped by `trim()` (U+FEFF is spec whitespace — verified on node v24.15.0, a BOM'd header line still reads as a comment); a UTF-16LE re-save still reads `null`, but now increments `silent`, so it is at least reported. A prepended note that *itself* contains a `worktree slot N` phrase wins over the generated marker (first match, not last), but the union means `server/.env` still holds the true claim, so it over-claims rather than under-claiming — no break to show.

**🟡 Minors 1 / 2 / 3 / 4 / 6 — FIXED.** The tautology is gone: `computePorts` mutated to ignore `slot` entirely now reddens `allocateNextSlot does not re-issue a surviving worktree slot after teardown (#3052)` via the new disk-read oracle (`:568-583`) — it did not before. Usage header corrected (`wt-new.mjs:16-19`); both dead branches gone; dead fixture call gone; TOCTOU documented at the site (`wt-new.mjs:353-360`).

**🟡 Minor 5 (`WORKSPACE_DIR` mechanism) — HALF fixed.** `589e621c`'s commit message carries the correction. The PR body does not — see 🟡 N4.

**Leaf-module cycle claim holds.** `npm run check:cycles` covers `server/src` only, so I walked `scripts/`'s own ESM graph directly: **207 `.mjs` files, 0 cycles**. `worktree-slot.mjs` imports node builtins only; `wt-new.mjs → wt-list.mjs → worktree-slot.mjs` is a DAG.

---

### 🟠 N1 — Significant, correctness bug: the new silent-skip warning fires on **every** real invocation, and what it says about the tree that trips it is false

`wt-new.mjs:145` counts a worktree as `silent` when it has any file from `SLOT_CLAIM_FILES` but yields no marker. `SLOT_CLAIM_FILES` (`worktree-slot.mjs:29`) includes `server/.env` — and **every real Castwright clone's primary checkout has a hand-written `server/.env` with no `wt-new` marker** (it is where `GEMINI_API_KEY` lives; `wt-new.mjs` never writes into the primary checkout). So the primary checkout is permanently `silent`.

Measured against the live 18-worktree fleet on this box (read-only, `collectSlotClaims(listWorktreePaths())`):

```
trees: 18  slots: 5,7,14,15,16,17,18,19,20  silent: 1  next: 1
  (none)  present=1  files=[server/.env]              Audiobook-Generator   <-- SILENT
  20      present=2  files=[.env.local,server/.env]   wt-2192-a29-…
  …16 more, all present=2, none silent
```

Reproduced end-to-end in a throwaway repo whose only marker-less file is the root's hand-written `server/.env` (`PORT=8080`, `WORKSPACE_DIR=…`) — the shape of every clone. `main(['feat/server-fresh','--no-install'])` exits 0 and prints, verbatim:

```
[wt-new] scanned 3 worktree(s); slots claimed: 1, 2
[wt-new] WARNING: 1 worktree(s) have a generated .env.local / server/.env but no readable slot marker — their slots are NOT protected and may be re-issued.
```

Two things are wrong here, not one:

1. **The warning is unconditional.** It will print on every `wt-new` run from a normal clone, forever. A warning with a 100% base rate carries no signal — and the *only* reason it was added this round was to make a genuine detection failure visible. The genuine case (a hand-written worktree env per CLAUDE.md "Worktree setup" step 5, marker stripped) now arrives as `2 worktree(s)` instead of `1`, which nobody will read as different.
2. **Its text is false about the tree that trips it.** "their slots are NOT protected and may be re-issued" — the primary checkout is slot 0, `allocateNextSlot` starts at 1 (`wt-new.mjs:164`) and can never issue it. The one tree structurally guaranteed to be safe is the one being reported as at risk.

Same root cause, second symptom: `wt-list` now prints `(none)` in the slot column for the primary checkout (it claims nothing), where it printed `0` before this PR. `0` was correct.

**Why the new integration test does not catch it:** `makeRealWorktreeRepo` (`wt-new.test.mjs:722-744`) commits a `server/package.json` into the fixture repo root but never a `server/.env`, so the fixture's "primary checkout" is `present=0` — a shape no real clone has. The test asserts `scanned 3` and reads stdout only; `process.stderr.write` is not captured (`:781-785`), so the warning is invisible to it either way.

The behaviour wanted is agreed (warn only when a claim was actually lost), so this is an implementation choice, not a design pass — e.g. exclude the primary checkout (`git worktree list --porcelain` always emits it first, and `repoRoot()` already resolves it), or count `silent` only on a tree that has `.env.local`, the file `wt-new` itself always writes.

---

### 🟠 N2 — Significant, correctness bug: `wt-list`'s PORT columns still read only `.env.local`, so the tool reports `(default)` for a tree whose server is bound to slot ports

`scripts/wt-list.mjs:60-68` — `readPortsFor` is unchanged and still reads `.env.local` alone. This PR's own argument (body, and `worktree-slot.mjs:15-28`) is that `.env.local` is deletable by design while `server/.env` "carries … the `PORT` / `LOCAL_TTS_PORT` the server and sidecar actually bind". That reasoning was applied to the slot column and not to the four columns beside it.

Repro (executed; two fixture trees, the second with `.env.local` deleted — the operation its own generated header at `wt-new.mjs:194` calls safe):

```
slot=3  tree=keeps-both          VITE=5203       PORT=8110       TTS=9030       E2E=5204
slot=4  tree=env-local-deleted   VITE=(default)  PORT=(default)  TTS=(default)  E2E=(default)

But the surviving server/.env for env-local-deleted actually pins:
  PORT=8120  LOCAL_TTS_PORT=9040
```

A maintainer reading row 2 concludes the tree is unconfigured and free to collide with; it is bound to 8120/9040. `VITE_PORT`/`PLAYWRIGHT_PORT` genuinely do fall back to defaults after that deletion, so only the two server-side columns lie — but those are the two the #3052 diagnosis turns on, and they are the ones the slot column was just fixed to agree with. This is pass 1's finding one column over: same premise, same file, same tool, left armed.

---

### 🟡 N3 — Minor, cleanup: the warning branch `main()` gained this round has no test — mutating it away leaves 80/80

`wt-new.mjs:370-375`. `if (claims.silent > 0) {` → `if (false && claims.silent > 0) {` → **80 pass / 0 fail**. The `silent` *counter* is guarded (`collectSlotClaims counts trees whose generated file yields no slot`), but the branch that surfaces it is not, and no test captures `process.stderr`. This is the same untested-wiring shape as pass 1's blocking finding, one level down — and it is why N1 shipped: a test that ran the warning against a realistic primary checkout would have shown it firing on the benign case.

---

### 🟡 N4 — Minor, cleanup: the PR body retracts a claim in one section and restates it verbatim in another, and its Test plan describes the previous head

The body's "Correction, and what review pass 1 changed" says the `WORKSPACE_DIR` mechanism "was stated wrong … it is the constant `'../castwright-workspace'` … so workspaces were always isolated." Four sections later, under **"Why this is a data hazard rather than an annoyance"**, unchanged:

> `WORKSPACE_DIR` is assigned off the same slot as the ports, so a collision lets **two lanes run one server against one workspace** — the same `cast.json` and `state.json`.

That is the exact sentence the correction retracts. The merge commit carries this body; a reader gets both.

Also stale in the same body, all describing `eed18a3e`:

- **"## Test plan"** — "`scripts/tests/wt-new.test.mjs` — **70 tests, 70 pass**. ESLint clean on both files." It is 80 tests and four files, which the Summary section already says, so the body disagrees with itself twice.
- **Mutation table** — cites `lines[0]` → whole file killing `extractSlotFromEnvLocal only reads the first line, not a later marker`. Neither exists at this head: `lines[0]` was replaced by the header-block scan, and that test is now named `…ignores a marker phrase below the settings, not just below line 1`.
- **"### The fix"** — still says allocation reads the marker "as line 1 of every generated `.env.local`" and that "a tree with no `.env.local` … claims nothing", both superseded by this head; and "with no call-site change", when `main()` now calls `collectSlotClaims()` + `allocateNextSlot(claims.slots)` rather than the bare default.

---

### 🟡 N5 — Minor, cleanup: `inRepo`'s `GIT_*` strip sits outside the region its `finally` protects

`scripts/tests/wt-new.test.mjs:750-766`. The delete loop and `process.chdir(repo)` both run *before* the `try`. If `chdir` throws (a removed fixture dir, a permissions failure), every `GIT_*` var is already gone from `process.env` and nothing restores it — every later test in the file then runs with a mutated environment, and the failure would surface as an unrelated test failing. Moving the strip inside the `try`, or wrapping both in the `finally`'s region, closes it. Unreachable in practice today; the dispatch asked specifically whether every `GIT_*` var is restored in a `finally`, and strictly it is not.

---

### Verdict

**Request changes.** The blocking finding is genuinely closed — I reproduced all four kills myself, plus a fifth on the default `main()` actually consumes, and the integration fixture is properly isolated from the real repository (18 worktrees before and after). The allocation core, the union, and the header-block reader all survive adversarial probing; I could not construct an input that makes allocation hand out a live slot.

What blocks is that the *reporting* added and touched this round is wrong in two ways that both matter for a tool whose entire job is diagnosis. The silent-skip warning — added specifically so a detection failure could not hide — fires on 100% of real invocations because of the primary checkout's own `server/.env`, and tells the operator that the one structurally-safe tree is at risk; its branch has no test, which is why that shipped. And `wt-list` still reports `(default)` ports for a tree bound to slot ports, because `readPortsFor` was left on the premise the slot column three lines above it just abandoned. Neither can produce a wrong allocation, but #3052 exists because the diagnostics agreed with the bug for months.

dudarenok-maker:
## PR review — pass 3 (head 8572efe8, depth medium)

Deltas only. Final pass under the loop cap.

Verified: `node --test scripts/tests/wt-new.test.mjs` **85/85**, full `npm run test:hooks` **1991/0**, `npx eslint` clean (exit 0) on all four touched files from the worktree root, cloud CI green at this head. **14 mutations** applied one at a time to a scratch copy of `scripts/` in `$TEMP` — never the worktree. HEAD `8572efe8` and an empty `git status --porcelain` before and after; `git worktree list` 18 rows before and after; the primary checkout's `git status` unchanged.

---

### Pass-2 dispositions

**🟠 N1 (the silent-marker warning fired on 100% of real runs; its text was false about the tree that tripped it; `wt-list` regressed to `(none)` for the primary) — FIXED.** Verified four independent ways rather than on the author's account:

| Check | Result |
|---|---|
| Exemption removed alone (`wt-new.mjs:160` → `if (present > 0 && slots.length === 0) silent++;`) | **83/2** — `…does not warn about the primary checkout’s hand-written server/.env`, `…DOES warn when a LINKED worktree lost its slot marker` |
| `isPrimaryCheckout` → always **false** (`worktree-slot.mjs:79`) | **82/3** — both of the above **plus** `wt-list buildRows prints slot 0 for the primary checkout, not (none)` |
| `isPrimaryCheckout` → always **true** | **82/3** — `collectSlotClaims counts trees whose generated file yields no slot`, `wt-list buildRows prints (none) for a tree with no generated env file`, `…DOES warn when a LINKED worktree lost its slot marker` |
| Warning branch disabled (`wt-new.mjs:385` → `if (false && claims.silent > 0)`) | **84/1** — `…DOES warn when a LINKED worktree lost its slot marker` (this is **N3**, below) |

Both directions of the exemption kill, which is the shape that matters — an exemption guarded only against "removed" would still let "always on" through.

Live read-only confirmation on this box, against the real 18-tree fleet: `collectSlotClaims(listWorktreePaths())` → `scanned=18 silent=0 next=1` (**was `silent=1`**), and `node scripts/wt-list.mjs` prints `0` in the slot column for `C:/Claude/Projects/Audiobook-Generator`. I checked the premise itself rather than taking it from the commit message: the primary's `server/.env` is a real 18 KB hand-written file with `GEMINI_API_KEY`/`WORKSPACE_DIR` and **no `PORT` key at all** — so `present=1, slots=[]`, exactly the shape that used to be counted `silent`. Warning text corrected to "linked worktree(s)".

**🟠 N2 (`readPortsFor` still read only `.env.local`) — FIXED.** Both claim files now go through one shared `readClaimFiles` (`worktree-slot.mjs:97-111`), consumed by `readSlotClaims` and by `wt-list`'s port columns alike, so there is no second reader to drift. Restricting the merge to `.env.local` reddens `wt-list port columns read server/.env when .env.local is deleted` (**84/1**). `wt-list.mjs` no longer imports `fs` or `path` at all.

**🟡 N3 (warning branch untested, nothing captured stderr) — FIXED.** `captureStderr` (`wt-new.test.mjs:769-782`) plus two tests covering both directions. Killed as tabulated above.

**🟡 N5 (`inRepo`'s `GIT_*` strip sat outside the region its `finally` protects) — FIXED.** Strip and `chdir` both moved inside the `try` (`wt-new.test.mjs:793-809`). Reverting that structure reddens `inRepo restores GIT_* vars even when the chdir into the fixture throws` (**84/1**) — the test is not decorative.

**🟡 N4 (PR body retracts a claim in one section and restates it in another; Test plan describes a two-heads-old run) — NOT FIXED.** Carried forward as finding 1.

---

### The five wiring mutations — re-run, all still kill

| Mutation | `file:line` | Suite |
|---|---|---|
| `allocateNextSlot(claimed = findClaimedSlots())` → `= []` | `wt-new.mjs:178` | **84/1** |
| `findClaimedSlots(worktreePaths = listWorktreePaths())` → `= []` | `wt-new.mjs:167` | **84/1** |
| `collectSlotClaims(worktreePaths = listWorktreePaths())` → `= []` | `wt-new.mjs:151` | **83/2** |
| `const slot = allocateNextSlot(claims.slots);` → `const slot = 1;` | `wt-new.mjs:377` | **84/1** |
| `return parseWorktreePorcelain(porcelain).map(…)` → `return [];` | `wt-new.mjs:118` | **82/3** |

Pass 1's blocking finding stays closed at this head.

---

### 🟡 Finding 1 — cleanup: the PR body still contradicts itself on the mechanism, and still reports the numbers of two heads ago

This is the only item I would spend two minutes on before pressing merge, because the merge commit carries this body permanently. None of it is a code defect.

**The retracted claim is still in the body, verbatim.** "Correction, and what review pass 1 changed" says the `WORKSPACE_DIR` mechanism "was stated wrong … it is the constant `'../castwright-workspace'` … so workspaces were always isolated." Four sections down, under **"Why this is a data hazard rather than an annoyance"**, unchanged:

> `WORKSPACE_DIR` is assigned off the same slot as the ports, so a collision lets **two lanes run one server against one workspace** — the same `cast.json` and `state.json`.

The code agrees with the correction, not with this: `wt-new.mjs:228` is `const WORKSPACE_DIR_RELATIVE = '../castwright-workspace';`, slot-independent, and `renderServerEnv` writes it verbatim for every slot. A reader of the merge commit gets both statements.

**Figures, measured at this head:**

| Body says | Actual at `8572efe8` |
|---|---|
| Summary: "**80 tests, 80 pass** (was 70)" | **85 / 85** |
| Summary: "`test:hooks` 1986/0" | **1991 / 0** |
| "## Test plan — `scripts/tests/wt-new.test.mjs` — **70 tests, 70 pass**. ESLint clean on **both** files." | 85 tests, **four** files — and the body's own Summary already disagrees with this line |

**Stale prose describing superseded code:**

- Mutation table cites `lines[0]` → whole file killing `extractSlotFromEnvLocal only reads the first line, not a later marker`. `git grep "only reads the first line"` returns **zero hits** — the test was renamed and `lines[0]` is gone.
- "### The fix" still says allocation reads the marker "as line 1 of every generated `.env.local`", and that "a tree with no `.env.local` … claims nothing". Both files are scanned now, and the scan walks the leading comment block.
- "### The fix" still says "with **no call-site change**". `main()` now calls `collectSlotClaims()` and passes `claims.slots` (`wt-new.mjs:376-377`) — that is a call-site change, and a deliberate one.
- **The pass-2/3 incidental fixes are not declared anywhere in the body.** "### Also fixed from that review" lists only pass-1's items. The primary-checkout exemption, `readPortsFor`'s union, the two stderr tests and the `inRepo` strip are absent. CLAUDE.md requires incidental fixes be declared in the PR body ("Also fixed, found in passing: …") precisely so they do not read as scope creep. The commit message for `8572efe8` covers all of them accurately — the body just never caught up.

---

### 🟡 Finding 2 — cleanup: the one fixture line that makes the exemption testable has no guard, and deleting it leaves the suite green

`scripts/tests/wt-new.test.mjs:743-747` — the marker-less `server/.env` written into the fixture repo root. The PR's own argument is that this line is load-bearing. **It is** — and I verified the proof rather than accepting it:

| Mutation | Suite |
|---|---|
| Remove the exemption (`wt-new.mjs:160`) alone | **83 / 2** |
| Remove the fixture's `server/.env` (`:743-747`) alone | **85 / 0 — survives** |
| Remove **both** together | **85 / 0 — survives** |

Row 3 is the author's claim and it holds: the fixture change is genuinely what makes the exemption tested. Row 2 is the exposure. Nothing asserts the fixture's primary checkout is `present=1`, so a future tidy-up that deletes those five lines silently converts both new exemption tests from "proves the primary checkout is exempt" to "passes because there was nothing to warn about" — with no test going red, and the exemption back to untested. That is the same control-never-measured shape this round has been chasing since pass 1; it has just moved into the fixture.

One line closes it, in `makeRealWorktreeRepo` or at the head of the "does not warn" test:

```js
assert.equal(readSlotClaims(repo).present, 1, 'fixture must give the primary a marker-less server/.env');
```

---

### 🟡 Finding 3 — cleanup: two micro-items in the new leaf module

1. **`readClaimFiles` returns two fields nothing reads.** `worktree-slot.mjs:97-111` returns `{ relative, path, content }`; `git grep` across all four files shows no consumer of `.relative` or `.path` — `readSlotClaims` uses `content` and `files.length`, `readPortsFor` uses `content`. CLAUDE.md "Simplicity first": no configurability that wasn't requested.
2. **The merge-precedence claim is untested.** `wt-list.mjs:70-72` states `server/.env` wins the keys both files carry. Reversing the merge order (`for (const file of [...files].reverse())`) leaves the suite **85/0**. The shipped order is the correct one — `server/.env` is what the server and sidecar actually read — and a divergence needs a hand edit, so this is a missing guard on a stated property, not a wrong behaviour.

---

### Probed hardest, and could not break

- **`isPrimaryCheckout`'s `.git`-is-a-directory premise.** Constructed and measured: no `.git` → `false`; `.git` as a FILE containing `gitdir: …` (the linked-worktree shape) → `false`; `.git` as a DIRECTORY → `true`; `.git` as a junction pointing at a directory → `true`; a real `git init` + real `git worktree add` → primary `true`, linked `false`; `git init --separate-git-dir` → working dir `false`, and `git worktree list --porcelain` reports the *gitdir* path for that repo anyway, so the working tree is never scanned. **I could not construct a linked worktree whose `.git` is a directory** — git does not produce that shape, and a hand-edit that made it one would break git for that tree first.
- **And the blast radius is bounded by construction, which is the part that settles it.** `isPrimaryCheckout` is referenced at exactly two sites — `wt-new.mjs:160` (the `silent` counter) and `wt-list.mjs:117` (one display cell). `collectSlotClaims` does `for (const slot of slots) claimed.add(slot);` **before** the exemption is consulted, so every claim a tree carries enters the claimed set regardless of how `isPrimaryCheckout` answers. A misfire in either direction can suppress or add a warning, or print `0` instead of `(none)`; it cannot re-issue a live slot. The exemption cannot reintroduce #3052.
- **Can the union lose a claim?** No input I could build does. Measured, all via `readSlotClaims`: unreadable `.env.local` (a directory) + `server/.env` claiming 7 → `{slots:[7], present:2}`; the two files disagreeing 4 vs 5 → `[4,5]`; marker-less `.env.local` + `server/.env` claiming 9 → `[9]`; empty `server/.env` + `.env.local` claiming 2 → `[2]`; and a decoy note reading `# reusing worktree slot 3 conventions` prepended above a real slot-6 marker → `[3,6]`, i.e. it **over**-claims. There is no removal path — a `Set` union across `readClaimFiles`, and files that fail to read are skipped for slots while still counting toward `present`, so an unreadable file surfaces as `silent` rather than as a free slot. The asymmetry the module documents is honoured: a false positive wastes a slot number, and I could not produce a false negative.
- **The port merge's direction is the safe one.** `.env.local` and `server/.env` both carry `PORT` and `LOCAL_TTS_PORT` (`renderEnvLocal` writes every `BASE_PORTS` key), and `.env.local`'s own header says `LOCAL_TTS_PORT` is not read from there. `server/.env` winning is therefore correct — see finding 3.2 for the missing guard, not for the behaviour.
- **`wt-list`'s `PORT_VARS` (`wt-list.mjs:14`) omits `VITE_API_PORT`**, which the PR body names as the actual proxy-hazard vector. Not raised as a finding: in every *generated* file `VITE_API_PORT === PORT` by construction (`BASE_PORTS`, same step), so the column would be redundant on every tree the tool will ever see unless a file was hand-edited. Noting it only because two passes have now found this tool one column short.
- **House gates.** `Closes #3052` present, plain, outside any code span. On-box acceptance N/A — deterministic Node scripts, nothing hardware-provable. Release-notes N/A re-checked rather than inherited: `scripts/wt-new.mjs`, `scripts/wt-list.mjs` and the new `scripts/lib/worktree-slot.mjs` are all absent from `build-release-zip.mjs`'s ship allowlist, so there is no shipped delta. `countExistingWorktrees` is gone with zero references tree-wide. `scripts/lib/worktree-slot.mjs` needs no `isDirectlyInvoked` guard or allowlist row — it is a leaf module, not an entry point, and `entry-point-guard-convention.test.mjs` passes inside the 1991.
- **Fixture isolation, re-confirmed at this head.** The four integration tests spawn `git worktree add` for real. `git worktree list` on the real repository is **18 rows before and after** a full `test:hooks` run, HEAD unmoved, `git status --porcelain` empty in this worktree and unchanged in the primary checkout.

---

### Verdict

**Approve.** No correctness bug found in the delta, and none carried over. Every pass-2 finding is genuinely closed — I reproduced each kill myself rather than reading the report, including both directions of the new exemption, and confirmed on the live fleet that `silent` went 1 → 0 and that `wt-list` prints `0` for the primary checkout. The exemption is the right shape: it keys on a fact about the tree rather than its position in an enumeration, and it sits on the warning path only, so even a misfire cannot put the allocator back on a colliding slot. The five wiring mutations that shipped green at pass 1 all still kill.

What remains is three cleanup items, none of which blocks: the PR body (finding 1) still carries the `WORKSPACE_DIR` sentence its own Correction section retracts, plus test counts from two heads ago and no declaration of this round's incidental fixes — worth fixing before merge only because the merge commit keeps it. Finding 2 is a one-line assertion that would stop a future edit from silently de-fanging the exemption tests. Finding 3 is two lines of tidying.

PR #3043, opened 2026-09-06
fix(server): stop sidecar before qwen pip install to avoid WinError 5

dudarenok-maker:
## Summary

Fixes the genuine, reproducible `WinError 5` / `Access is denied` on `onnxruntime_providers_shared.dll` that register row A29's fifth on-box attempt finally exercised for real (`docs/testing/onbox-a29-results/step-2-genuine-install.md`): the in-app **Install Qwen3-TTS** ran `pip install qwen-tts` under a live sidecar that has that DLL memory-mapped (the sidecar imports onnxruntime at boot), so pip's DLL replace failed every time the app was running — the only time the button is reachable.

Three earlier rounds on this PR patched it with `stop()`/`start()` and each left the supervisor in a state it could not recover from. This round restructures rather than re-patches:

- **A new, call-scoped supervisor primitive, `withSidecarHeld(fn)`** (`server/src/tts/sidecar-supervisor.ts`) — *not* `stop()`+`start()`, which own the supervisor's permanent shutdown state. It kills the child with auto-respawn suppressed, **waits for the child's actual `exit`** (`kill()` resolves when `taskkill` returns, not when the process is gone — a late exit landing after the resume would null the new handle and double-spawn), holds the queue (`recycling()` reads true throughout), runs `fn`, and always brings the sidecar back. The hold flag is released synchronously in the primitive's own `finally` *before* the respawn is attempted, so neither `fn` nor a failed respawn can leave it set: a throwing respawn is warned and left to `POST /api/sidecar/restart`; a refused one goes through the ordinary backoff/cap retry; a child that never exits within 30s is **rolled back** (handle restored, queue released, error thrown) rather than proceeded past. `stop()`/`start()` are byte-identical to `main`.
- **Adopted vs. not-started uses the supervisor's real signal**, `adoptedWatching` — the only state meaning "a sidecar is running that we didn't spawn". `current() === null` is shared with autoStart-off and mid-respawn; `recycling() === false` is shared with autoStart-off; neither accessor can tell them apart alone. The hold also refuses a code-43 trip, plain exhaustion, a respawn already in flight, a stopped supervisor, and a concurrent hold — all decided synchronously before any state moves.
[...]

PR #3055, opened 2026-09-06
chore(ops): add worktree GC with junction-first teardown

### LINKED ISSUE #3051: ops-75 — commit gate Part 4: worktree GC
## What

`scripts/wt-gc.mjs` / `npm run wt:gc`. Lists worktrees with commits-ahead-of-`main`, merged status, and (when `gh` is available) PR state; offline-tolerant. Reports by default, `--prune` acts. Design of record: **Part 4** of [`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md`](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md), which marks it "Independent of Parts 1-3 and cuttable". Parent #2997.

## Why it earns its place

On 2026-09-06 a manual sweep found **28 orphaned worktree directories, 8.27 GB**, against 14 registered worktrees. Twelve of the fourteen junctions in them pointed at the primary checkout's real `node_modules`/`.venv`/Kokoro weights. That is the hazard this automates away, and it is the reason the teardown order below is not a detail.

## Teardown order is load-bearing

**Junctions are dropped FIRST**, via `[System.IO.Directory]::Delete($p, $false)`, gated on the `ReparsePoint` attribute:

- **Never gate on `.LinkTarget`** — it is PowerShell 6.2+ and reads **empty** on this box's Windows PowerShell 5.1 even for a real junction, so a guard written against it skips the delete silently and the junction survives into the recursive remove.
- **Enumerate junctions recursively, not just at top level.** `server/tts-sidecar/.venv` sits three levels down; a top-level-only check misses it, and a `Remove-Item -Recurse` over a tree still holding it follows the link into the primary checkout's real venv. This has been survived by luck once already.
- `cmd /c rmdir` from a bash shell **silently no-ops and returns 0** — verify removal with `Test-Path` on both the junction and its target, never with an exit code.
- A path an `npm install` converted from junction to real directory needs `Remove-Item -Recurse -Force` instead.

## Refusals — all three required

Refuses to prune the primary checkout, any tree with uncommitted changes, and any tree with unpushed commits.

## Acceptance

1. Unit tests for the merged / ahead / dirty / unpushed refusal cases, each asserting against input that would otherwise make the guard fire.
2. Offline tolerance: `gh` absent or failing degrades to git-only reporting rather than erroring.
3. Report mode is the default; `--prune` is explicit.
4. The `ReparsePoint` gate is tested, and the test would fail if the gate were changed to `.LinkTarget`.
5. On-box acceptance row for a real prune against real junctions — the destructive path cannot be proven in-PR.

Refs #2997

## Benefit

*(technical)* Automates the sweep that last time reclaimed 8.27 GB by hand, with the junction-first teardown order that keeps it from deleting the primary checkout's real dependency trees.

dudarenok-maker:
## Summary

Part **4** of the [commit-gate rebalance design](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md) — `scripts/wt-gc.mjs`, exposed as `npm run wt:gc`. Reports by default; `--prune` is explicit.

Sized against a real problem: a manual sweep on 2026-09-06 found that **twelve of the fourteen registered worktrees' junctions pointed at the primary checkout's real `node_modules`, `.venv` and Kokoro weights** — a `Remove-Item -Recurse` teardown over one of those follows the link and deletes the target.

**Scope, stated up front because that same sweep is easy to misread as this tool's remit.** `wt:gc` operates on the worktrees `git worktree list` REGISTERS, and only those. The **28 already-orphaned directories (8.27 GB)** the sweep also counted — directories git no longer knows about — appear in no porcelain and are therefore structurally invisible here: `wt:gc` neither lists nor reclaims them. What this ships is the thing that stops registered worktrees from *becoming* those: the junction-first teardown recipe made routine, and refused on any tree that is not safe to lose. Clearing an already-orphaned directory remains the manual recipe in CLAUDE.md "Worktree teardown", and #3051's acceptance list does not ask for it.

### The teardown order is the whole ticket

Junctions are dropped **first**, via `[System.IO.Directory]::Delete($p, $false)`, gated strictly on the `ReparsePoint` attribute bit — **never `.LinkTarget`**, which is PowerShell 6.2+ and reads empty on Windows PowerShell 5.1 even for a real junction. A guard written against it skips the delete silently and the junction survives into the recursive remove, which then follows it into the primary checkout's real dependency tree.

Removal is verified with `Test-Path` on both link and target, never by exit code — `cmd /c rmdir` from a bash shell silently no-ops and returns 0. The finder never descends into a junction it found.

### Refusals — six, and every one fails closed

**Corrected in review pass 1 (see below): this section previously read "all three required", which was wrong on the merits and disagreed with the ticket.** Issue #3051's acceptance #1 names *"merged / ahead / dirty / unpushed refusal cases"* — merged **as a refusal case**. The design doc's Part 4 sentence listed only three, the implementation followed the doc, and the result was that `refusalReasons()` never read the `merged` or `PR` columns the report printed: on this box, six rows read `prunable? yes`, of which five were unmerged, four carried an open PR, and one was **this PR's own worktree**. The ticket is right, and the design doc is amended in this diff (`docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md`, Part 4) with the reconciliation stated inline rather than silently.

The shipped set:

1. **the primary checkout** — and if `git rev-parse --git-common-dir`/`--git-dir` cannot be answered, the tree is **treated as** the primary checkout. That check used to fail OPEN while both its siblings (`dirty`, `unpushedCount`) fail closed; git's own `fatal: is a main working tree` backstops only the step *after* the junction sweep has already run over the primary checkout.
2. **the worktree this process is running from.** Measured: `git worktree remove --force` from inside its own tree deletes every file and deregisters the worktree, then fails the final rmdir with exit 255 — leaving exactly the orphaned, no-longer-registered directory this tool exists to remove, reported to the user as a failed prune.
3. **uncommitted changes.**
4. **not merged into `main`.** Teardown destroys per-worktree state that does not travel with the branch — `server/.env` (`PORT`/`WORKSPACE_DIR`/`LOCAL_TTS_PORT`), `.env.local`, the `node_modules` junctions, `server/tts-sidecar/.venv`, `voices/` — which CLAUDE.md states outright. "The commits are pushed, so nothing is lost" is false: the commits survive, the environment does not.
5. **unpushed commits**, including a branch with **no upstream at all**, treated as unverifiable rather than assumed safe.
6. **an open PR, or a PR state that could not be determined at all.** `gh pr list --head <branch>` printing `[]` at exit 0 ("there is no PR") used to render identically to `gh` being uninstalled ("I could not ask"). Those are opposite answers for a destructive default, so they are now separate: `none` vs `unknown (gh unavailable)` — the first safe, the second a refusal.

`gh` calls still route through the `scripts/gh.mjs` chokepoint and never throw — offline tolerance means the **report** keeps working, not that the **prune** deletes trees it could not check. `gh` is also only queried for a row that has no other refusal, since a row that cannot become prunable cannot be changed by the answer (17 serial round-trips saved on this box; such rows render `not queried`).

**Live proof on this box, report mode, after the fix:** 18 rows, **exactly one** `prunable? yes` — `wt-2192-a29-genuine-install-test`, merged, clean, PR #3041 MERGED. This PR's own worktree now reads `no (the worktree this process is running from — never pruned; uncommitted changes; not merged into main)`.

### The junction scan fails closed too

`Get-JunctionsRecursive` used `-ErrorAction SilentlyContinue` plus a bare `catch { return }`, so whole subtrees were dropped without a signal — and an empty result meaning *the scan failed* is indistinguishable from one meaning *there are no junctions here*. The caller acts on the second by running `git worktree remove --force`, which then follows the junction the scan never saw. Demonstrated with a real junction at a 394-character path: `pwsh` finds it; **Windows PowerShell 5.1 returned `{"items": []}`, exit 0, no warning** — and `pickPowerShell()` falls back to `powershell`. Every enumeration error now throws, the `.ps1` turns that into a non-zero exit, and `run()` aborts that tree's prune.

### The `.ps1` ↔ JS JSON seam is pinned from both sides

Not a live defect — the keys matched — but renaming `TargetStillExists` would have made `j.TargetStillExists === false` permanently false, so the **catastrophic** case (junction unlinked *and* its real target destroyed) would read as success with zero tests reddening. Now: `JUNCTION_RESULT_KEYS` is exported from `wt-gc.mjs`, `validateJunctionEntry()` makes an off-contract entry a runtime failure, the JS suite asserts the `.psm1` source emits each key, and the Pester suite asserts the real emitted property set against that same JS constant. A one-sided rename reddens tests on **both** sides (verified — M17 below).

## Test plan

- `scripts/tests/wt-gc.test.mjs` — **62 tests**, all green (was 27).
- `scripts/tests/wt-gc-junctions.Tests.ps1` — **20 Pester cases**, all green (was 13; Pester 6.1.0).
- `npm run test:hooks` — **2019/2019**. `npx eslint scripts/wt-gc.mjs scripts/tests/wt-gc.test.mjs` — clean.
- `npm run check:onbox-register` OK; `npm run register:build -- --check` **up to date** (it fails on a committed `body-placeholder`, which is exactly what the E103 row was).

**Mutation-verified — 19 mutations**, each applied to production source, the *named* test run and watched go red, then restored and re-run green:

| # | Mutation (production source) | Named test that reddened |
|---|---|---|
| M1 | `if (!facts.mergedIntoMain)` → `if (false)` | `--prune REFUSES an unmerged worktree and never touches it` |
| M2 | `else if (facts.prOpen)` → `else if (false)` | `--prune REFUSES a worktree whose branch carries an OPEN PR` |
| M3 | `if (!facts.prAvailable)` → `if (false)` | `gh unavailable ... and REFUSES the prune` |
| M4 | `if (facts.isSelf)` → `if (false)` | `--prune REFUSES the worktree the process is running from` |
| M5 | `c.startsWith(``${wt}/``)` → `c.startsWith(wt)` | `isSelfWorktree: false for a sibling worktree with a shared prefix` |
| M6 | the fail-closed `isPrimary` → the old fail-open `!== null &&` form | both `REFUSES every tree when git cannot resolve ...` tests |
| M7 | `!prInfo \|\| prInfo.available !== true` → `!prInfo` | `"gh answered: no PR" and "gh could not be asked" are DIFFERENT states` |
| M8 | `validateJunctionEntry`'s missing-key list → `[]` | `a junction entry missing a contract key aborts` |
| M9 | `facts.unpushedCount ?? null` → `?? 0` | `an unverifiable unpushed count is reported as null` |
| M10 | `facts.unpushedVerified === true` → `true` | `an unverifiable unpushed count is reported as null` |
| M11 | `if (otherRefusals.length > 0)` → `if (false)` | `gh is NOT queried for a row that already carries another refusal` |
| M12 | `.psm1` enumerate-error `throw` → swallow (`$children = @()`) | `THROWS when a directory cannot be enumerated` |
| M13 | `.psm1` inspect-error `throw` → `return` | `THROWS when a directory cannot be inspected` |
| M14 | same as M12 | `propagates the failure out of Remove-JunctionsRecursive` |
| M15 | `.ps1` `ValidateSet('Remove')` → `('Find', 'Remove')` | `REJECTS -Action Find` (Pester) |
| M16 | same as M15 | `the .ps1 no longer offers the unreachable, shape-mismatched -Action Find` (JS) |
| M17 | `.psm1` `TargetStillExists` renamed — **PowerShell side only** | `the .psm1 emits every key JUNCTION_RESULT_KEYS names` (JS) **and** `emits exactly the property set wt-gc.mjs declares` (Pester) |
| M18 | `run()`'s junction-scan `catch` → swallow and carry on | `a THROWING junction scan aborts the prune for that tree` |
| M19 | `prInfo.queried === false` → `false` | `a row skipped for gh renders "not queried"` |

Every refusal test hands in an all-clear fact set with exactly one flag flipped, so none can pass because some *other* refusal fired; each also has a paired negative case proving the flag is read rather than always-on.

### One result worth stating plainly

Mutating `Test-IsReparsePoint` to use `.LinkTarget` reddened the source-pin tests **but the behavioural junction test still passed** — this box runs pwsh 7.6.5, where `.LinkTarget` happens to be populated. That is precisely the documented hazard: a `.LinkTarget`-based guard looks correct here and breaks on Windows PowerShell 5.1. The source pin is doing real work that no behavioural test on this box can do.

### Review pass 1 — all findings folded

The Blocking finding and all six Significant findings are fixed above, plus the four Minor ones: the header comment listing a non-existent `unpushed` report column is corrected to what `formatReportTable` actually emits; the dead, shape-mismatched `-Action Find` branch is deleted (it returned an array of **strings**, so `!j.Removed` would have been truthy for every one of them); a row whose unpushed count could not be read now reports `unpushedCount: null` plus `unpushedVerified: false` rather than claiming a count it does not have; and `gh` is no longer queried for rows that already carry a refusal.

Also carried in from the review's related note: `mergedIntoMain` is computed against the **local** `main`, so a stale local `main` reports a merged tree as unmerged — that errs toward refusing, and the file header now says so and names the fix (`git fetch && git merge --ff-only main`).

`findJunctions` was defined as a runner and destructured in `run()` but never called — `removeJunctions` handles find-and-remove internally. Removed from both, plus seven dead `node:fs`/`node:path` imports in the test file. Both were caught by `pre-commit`'s ESLint pass, which blocked the first commit attempt.

## On-box acceptance

Register row **E103** added for the destructive `--prune` path, which cannot be proven in-PR. Register and live-view HTML move in this diff; `check:onbox-register` passes and `register:build --check` reports up to date. **Review pass 1 caught that the live view's E103 entry was still the generator's unfilled `body-placeholder` stub, with raw markdown in its `iname`** — the only placeholder in the file, and what would have been published. It is now hand-authored in the style of its neighbours (plain-prose `iname`, `<ul>` criteria, a `flag` block for the fail-closed-scan half no fixture can prove, a `p.src` prerequisites line), and both the markdown row and the HTML carry the six-refusal criteria rather than the old three. **Publishing the live view is deliberately NOT done from this branch, and is not a merge precondition** — an earlier revision of this body said it was, which is now wrong. The page currently live at the recorded URL is **ahead of `main`** (publish token 11 against this file's 9): another lane published a page carrying rows `main` has not seen yet, exactly the AHEAD case the register's own procedure grew a section for in #3075 (merged into this branch in the same merge commit that resolved this round's conflicts). Republishing from here would silently delete that lane's rows with every local check green. The publish happens **once**, from a checkout that contains every lane touching the register — #3055 (E103), #3061 (A106), #3063 (E104) and the merged #3064/#3065 — and only after `npm run check:onbox-register -- --against-published <saved live page>` agrees. Recording the debt is what blocks a merge; publishing it is not, and doing it early is the failure mode.

Before-shipping steps not applicable: **1** (Part 4's design already lives in the cited spec), **5** (internal dev-ops tool, no end-user-visible effect — `docs/release-notes-next.md` gets the technical entry, `RELEASE_NOTES.md` deliberately does not).

Closes #3051
Refs #2997


dudarenok-maker:
## PR review — pass 1 (head fa1b1bf8, depth medium)

Scope: all 9 changed files — `scripts/wt-gc.mjs`, `scripts/lib/wt-gc-junctions.{ps1,psm1}`, `scripts/tests/wt-gc.test.mjs`, `scripts/tests/wt-gc-junctions.Tests.ps1`, `package.json`, and the three doc surfaces. Verified before probing: `node --test scripts/tests/wt-gc.test.mjs` **27/27**; `Invoke-Pester scripts/tests/wt-gc-junctions.Tests.ps1` **13/13** (Pester 6.1.0); `npm run check:onbox-register` **OK**. Then adversarially probed the junction-first ordering, the `ReparsePoint` gate on **both** engines (pwsh 7.6.5 and Windows PowerShell 5.1.26100), the three refusals, offline degradation, and what the injected-fake tests leave unproven — using throwaway `%TEMP%` fixtures with real `New-Item -ItemType Junction` links and a throwaway git repo. No tracked file modified; `--prune` never run against anything real.

---

### 🔴 Blocking — `--prune` marks unmerged worktrees with OPEN PRs as prunable. On this box that is four active lanes plus this PR's own worktree.

**Correctness bug.** `refusalReasons()` (`scripts/wt-gc.mjs:209-216`) reads only `isPrimary`, `dirty`, `hasUpstream`, `unpushedCount`. It never reads `facts.mergedIntoMain`, and `classifyWorktree` (`scripts/wt-gc.mjs:219-231`) computes `prState` from `prInfo` **for the report only** — `prInfo` is never passed to `refusalReasons`. The two columns the tool advertises as its safety signal are decoration.

Repro — the tool's own read-only report mode, run on this box at this head:

```
$ node scripts/wt-gc.mjs
path                                branch                                     merged ahead dirty PR           prunable?
wt-2192-qwen-install-sidecar-stop   fix/server-2192-qwen-install-sidecar-stop  false  12    false #3043 OPEN   yes
wt-2903-a34-repair                  fix/scripts-2903-a34-repair                false  3     false #3057 OPEN   yes
wt-3051-worktree-gc                 chore/ops-3051-worktree-gc                 false  1     false #3055 OPEN   yes
wt-onbox-mechanical-batch1          docs/docs-onbox-mechanical-batch1          false  11    false #3031 OPEN   yes
wt-fs38-a1-wave12                   chore/ops-fs38-a1-wave12                   false  4     false unknown      yes
wt-2192-a29-genuine-install-test    fix/ops-2192-a29-genuine-install-test      true   0     false #3041 MERGED yes
```

Six rows read `prunable? yes`. **Five are unmerged. Four carry an open PR. One is `wt-3051-worktree-gc` — the worktree this PR lives in and is being reviewed in.** A single `npm run wt:gc -- --prune` on this box today tears down four in-flight agent lanes mid-review, plus itself.

"But the commits are pushed, so nothing is lost" does not hold. The commits survive; the *environment* does not, and it is not in git. Each of those trees carries per-worktree `server/.env` (`PORT`, `WORKSPACE_DIR`, `LOCAL_TTS_PORT`), `.env.local`, the `node_modules` / `server/node_modules` junctions, and for sidecar lanes `server/tts-sidecar/.venv` + `voices/`. CLAUDE.md → Branching workflow → "Git-ignored artifacts" states outright that these are "destroyed by worktree teardown". Rebuilding one is `wt-new.mjs` + two `npm install`s + hand-written env files + junctions.

**The two specs disagree, and the PR resolved it silently in the destructive direction.** Issue #3051 acceptance #1: *"Unit tests for the **merged** / **ahead** / dirty / unpushed refusal cases"* — merged and ahead are named **as refusal cases**. The design doc's Part 4 "Refusals" sentence lists only three and omits merged. The implementation followed the design doc; the PR body asserts "Refusals — all three required" without noting that the ticket says four. Which refusal is right (refuse unmerged; refuse an open PR; require a second flag; interactive confirm) is a genuine design decision and may warrant its own pass — but shipping a default `--prune` that destroys four live lanes on the box it was written for is not a defensible resting state.

Related, and it bites the moment `merged` is wired in: `mergedIntoMain` is computed against the **local** `main` (`git merge-base --is-ancestor <head> main`, `scripts/wt-gc.mjs:242`), so a stale local `main` reports a genuinely-merged tree as unmerged.

---

### 🟠 Significant — the primary-checkout refusal fails OPEN when git can't answer, while both its siblings deliberately fail closed

**Correctness bug.** `scripts/wt-gc.mjs:314`:

```js
const isPrimary = commonDir !== null && gitDir !== null && isPrimaryWorktree(commonDir, gitDir);
```

If `git rev-parse --path-format=absolute --git-common-dir` (line 307) **or** `--git-dir` (line 312) exits non-zero, `isPrimary` becomes `false` and the primary checkout loses refusal #1 entirely. Contrast the two facts computed beside it, both of which fail *closed* with an explicit comment saying so: `dirty` (line 240, `// fail closed`) and `unpushedCount` (line 259, `// couldn't verify — fail closed`). This is the one that goes the other way, and it is the only one whose failure points the destructive path at the primary checkout.

Repro — the **production** `run()` imported from `scripts/wt-gc.mjs` (no logic reimplemented in the probe), driven by a stub git. The primary checkout is in its most ordinary post-merge state: clean, on `main`, `main` tracks `origin/main`, 0 unpushed. The *only* difference between A and B is the exit status of that one call:

```
=== A. --git-common-dir resolves ===
C:/repo  main  true  0  false  unknown  no (primary checkout — never pruned)
SKIP C:/repo — primary checkout — never pruned
removeJunctions called on: []

=== B. --git-common-dir FAILS (e.g. git < 2.31: no --path-format) ===
C:/repo  main  true  0  false  unknown  yes
Pruning C:/repo...
removeJunctions called on: ["C:/repo"]
removeWorktree  called on: ["C:/repo"]
```

(Scenario C — only `--git-dir` failing, e.g. a `safe.directory`/dubious-ownership refusal — is identical.)

Damage is bounded but real. I verified in a throwaway repo that git itself refuses the last step: `git worktree remove --force <main worktree>` → `exit=128`, `fatal: '...' is a main working tree`, files intact. So the directory survives — **but `removeJunctions(<primary checkout>)` has already run**, unlinking every reparse point found anywhere under the primary checkout after a full recursive walk of its real `node_modules`. Nothing in this tool stops that; git is the only backstop, and it only guards the step *after*.

No test covers a failing `--git-common-dir`/`--git-dir`. Fix is one line: unresolvable ⇒ treat as primary.

---

### 🟠 Significant — nothing excludes the worktree the tool is running in; pruning it half-destroys the tree and is then reported as a failure

**Correctness bug.** `run()` never compares `row.path` against `repoRoot` (exported at `scripts/wt-gc.mjs:58`, used only by `makeDefaultRunners`). Confirmed live in the report above: `C:/Claude/Projects/wt-3051-worktree-gc` — the tree the script lives in, and was invoked from — reads `prunable? yes`.

Measured in a throwaway repo, running the command exactly as `removeWorktree` does (`cwd` inside the tree):

```
$ git -C <wt> worktree remove --force <wt>
exit=255   error: failed to delete '<wt>': Permission denied
$ git -C <repo> worktree list      # <wt> no longer listed — deregistered
$ Test-Path <wt>                   # True
$ Get-ChildItem <wt> -Force -Recurse   # (empty) — every file already deleted
```

git deletes the contents and deregisters the worktree, failing only the final `rmdir` because the process CWD sits inside it. Applied to `npm run wt:gc -- --prune` from a worktree — the normal invocation here, since `package.json` is present in every worktree and CLAUDE.md mandates working in one — junctions are unlinked, the tree is emptied and deregistered, `removeWorktree` returns 255, and `scripts/wt-gc.mjs:362-366` prints `git worktree remove failed` and increments `failures`. **The user is told the prune failed for a tree that no longer exists in any usable sense.** `pathExists` (line 369) is never reached, because the non-zero branch `continue`s first. The leftover empty directory is then precisely the orphaned-directory litter this tool exists to remove — and since git deregistered it, `git worktree list` no longer reports it, so a re-run of `wt-gc` cannot clean it either.

---

### 🟠 Significant — the junction scan reports success while silently incomplete, and nothing re-scans before the recursive delete

**Correctness bug.** `wt-gc-junctions.psm1:69` (`Get-ChildItem ... -ErrorAction SilentlyContinue`) and `:60-64` (`try { Get-Item } catch { return }`) each drop an entire subtree without a word. `Remove-JunctionsRecursive` then reports only on what the scan found; `run()` (`scripts/wt-gc.mjs:348-358`) logs `removed N junction(s)` and proceeds to `git worktree remove --force`. A junction the scan missed is still live when the recursive delete runs — the exact failure this PR exists to prevent.

Demonstrated with a **live junction**, using the production `.ps1`, in a `%TEMP%` fixture (junction at a 394-char path; its target holds a real file):

```
$ pwsh -File scripts/lib/wt-gc-junctions.ps1 -Root <fixture>\wt -Action Find
{ "items": [ "...\\aaaa...\\aaaa...\\node_modules" ] }

$ powershell -File scripts/lib/wt-gc-junctions.ps1 -Root <fixture>\wt -Action Find
{ "items": [ ] }        # exit 0. no warning, no stderr, no error field.
```

Windows PowerShell 5.1 raises `PathTooLongException` inside `Get-ChildItem`; `-ErrorAction SilentlyContinue` eats it and the module answers "no junctions here" with a live junction sitting in the tree. This is not a hypothetical engine: `pickPowerShell()` (`scripts/wt-gc.mjs:157-163`) probes `pwsh` first and **falls back to `powershell`**, and the module's own `#requires -Version 5.1` declares 5.1 supported. This box runs pwsh 7.6.5, so the Pester suite passing here proves nothing about the 5.1 path. The same swallow also covers an ACL-denied directory, on either engine.

Cheapest fix, which also closes the scan-to-delete race: after removal, re-run `Get-JunctionsRecursive` and refuse to call `git worktree remove` if it returns anything; and surface enumeration errors as a per-root `Errors` list the JS side treats as a junction failure, instead of `SilentlyContinue`.

---

### 🟠 Significant — `"unknown"` PR state conflates "gh is unavailable" with "gh answered: there is no PR"

**Correctness bug.** `makeDefaultRunners().ghPrState` (`scripts/wt-gc.mjs:111-123`) returns `null` on `result.error || result.status !== 0` (gh missing / unauthenticated / offline) **and** on `rows[0] ?? null` (gh succeeded, empty list). `classifyWorktree:228` maps both to the same string `'unknown'`.

Confirmed live:

```
$ gh pr list --head chore/ops-fs38-a1-wave12 --state all --json number,state --limit 1
[]
$ echo $?
0
```

That row prints `unknown` in the real report above — character-for-character what it would print with `gh` uninstalled. Issue #3051 acceptance #2 frames PR state as the offline-degradation surface; collapsing "definitely no PR" into the same token as "couldn't ask" destroys the one signal a user could have used to sanity-check a `prunable? yes` row (see the Blocking finding). Two distinct sentinels cost one branch.

---

### 🟠 Significant — the `.ps1`-to-JS JSON seam is pinned on neither side

**Chore / test gap — not a live defect; I verified the contract holds today.** `wt-gc.test.mjs` hand-writes `{ Path, Removed, TargetStillExists, Error }` into the `removeJunctions` stub (lines 276, 289, 306, 331). `wt-gc-junctions.Tests.ps1` asserts the PowerShell side emits those properties. **Nothing asserts that `wt-gc-junctions.ps1`'s `{"items":[...]}` envelope — parsed at `scripts/wt-gc.mjs:180-184` and destructured as `.items` at line 129 — actually yields objects with those keys.**

Concrete consequence of one future rename: spell `TargetStillExists` any other way and `j.TargetStillExists === false` (line 349) is permanently false, so **the catastrophic case the test at line 300 exists to catch** — junction unlinked *and* its real target destroyed — silently passes and `git worktree remove --force` proceeds. Zero JS tests redden. (A `Removed` rename fails closed instead, so only one of the two keys is dangerous — worth knowing, and recorded nowhere.)

`makeDefaultRunners()`, `pickPowerShell()` and `runJunctionScript()` are untested in full. That includes the offline-tolerance claim: `wt-gc.test.mjs:343` proves only that `classifyWorktree` maps `null` to `'unknown'`, against a stub that returns `null` by construction — an assertion that cannot fail. The real swallow (`result.error || result.status !== 0`) is exercised by nothing.

---

### 🟠 Significant — on-box acceptance: the live view's E103 row is the generator's unfilled placeholder, and it is slated to be published

**Chore — house gate (before-shipping step 3) incomplete.** `docs/testing/onbox-acceptance-register-live-view.html:1284-1289`:

```html
<summary><span class="num">E103</span><span class="iname">`scripts/wt-gc.mjs --prune` — real junction-first teardown ([#3051](https://github.com/dudarenok-maker/Castwright/issues/3051), ops-75 Part 4)</span><span class="risk">Not yet published</span><span class="chev">&rsaquo;</span></summary>
<div class="body">
  <p class="body-placeholder">Not yet published — run `npm run register:build` after adding row content, or fill in manually and re-run --check.</p>
</div>
```

Two defects in the one row:

1. **The body is the tool's own stub, not content.** It is the *only* `body-placeholder` in the entire file (`grep -c body-placeholder` returns `1`); every neighbouring row (E9, E11, ...) is hand-authored prose with `<code>`/`<a>`/`<ul>`/`<p class="src">`. The markdown register's E103 has a good, concrete "what to observe" list — none of it reached the third surface.
2. **The `iname` carries raw markdown** — literal backticks and `([#3051](https://...))` — which renders verbatim in the browser. Every neighbouring `iname` is plain prose.

`npm run check:onbox-register` passes (I ran it: `OK — ... agree`) because it validates ids, counts and structure, never row content — so the mechanical gate cannot catch this. The PR body says *"The live view still needs publishing to its recorded artifact URL before merge"*, which means **this placeholder is what would go live**, on the surface this repo has already broken four times (2026-07-31/08-01).

---

### 🟡 Minor

- **`scripts/wt-gc.mjs:14-16` documents a report column that does not exist.** Header: *"Report columns per non-primary worktree: path, branch, merged-into-main, commits-ahead-of-main, dirty, **unpushed**, PR state"*. `formatReportTable:269` emits `['path','branch','merged','ahead','dirty','PR','prunable?']` — no `unpushed`. Unpushed state surfaces only inside the `prunable?` cell's refusal text. (Reviewer-brief catalogue #9 — the defect is in the document.)
- **`-Action Find` is unreachable and shape-mismatched.** `runJunctionScript` is only ever called with `'Remove'` (`scripts/wt-gc.mjs:129`); this PR removed the `findJunctions` runner. The surviving `Find` branch (`wt-gc-junctions.ps1:25-29`) returns an array of **strings**, not objects — so if anything ever routes it through `run()`'s loop, `!j.Removed` is truthy for every string and every junction reads as a failure. Delete the branch, or record the shape difference at both ends.
- **`classifyWorktree` can emit `unpushedCount: null`** (`wt-gc.mjs:258` sets `null`, `:259` flips `hasUpstream` false so the refusal itself still fires correctly). Harmless today only because the field is never rendered; the row object claims a count it does not have.
- **`ghPrState` is called once per worktree even in report mode** (`wt-gc.mjs:317`), serially, including for rows that already carry a refusal. 17 worktrees here means 17 sequential `gh` round-trips before the first row prints. Skipping already-refused rows, or a `--no-gh` flag, would pay for itself.

---

### ✅ What is solid

- **The junction-first teardown is genuinely correct, and correct on the engine that matters.** Verified end-to-end on **Windows PowerShell 5.1.26100** in a scratch fixture with a real `New-Item -ItemType Junction` at `wt/server/tts-sidecar/.venv`: the link is unlinked, `TargetStillExists: true`, and the target's `PRECIOUS.txt` survives byte-intact. `[System.IO.Directory]::Delete($p, $false)` and the `ReparsePoint`-attribute gate behave exactly as documented on 5.1, and `.Target` (reporting only, never the gate) populates there too.
- **The `.LinkTarget` source pin is the right call and I could not defeat it.** `Should -Not -Match '\.LinkTarget'` over comment-stripped source is the *only* test that can catch that regression on a pwsh-7 box, and the comment filter is correctly scoped (every real reference sits on its own line). The PR body's plain statement that the behavioural test still passes under a `.LinkTarget` mutation is exactly the right reading of its own evidence.
- **The three refusals work when git answers.** Live: the primary checkout refused on two grounds; `wt-2616-onbox-group-c` refused on **67 unpushed commits**; three trees refused with *"no upstream configured — cannot verify it is pushed"*. "No upstream at all" really is treated as unverifiable rather than assumed safe, as claimed.
- **The tests call the production functions — no reimplementation-in-test shape.** 27/27 node tests import `run`/`refusalReasons`/`classifyWorktree`/`isPrimaryWorktree` from `../wt-gc.mjs`; 13/13 Pester cases import and call the real module. Both re-run green here. The empty-array assertions (`($found -is [array])` on the captured variable, with the comment explaining why not `Should -BeOfType` off a pipe) avoid a vacuous-pass shape this repo has been burned by before.
- Partial junction failure (`Removed:false`) and target-destroyed (`TargetStillExists:false`) both correctly abort **before** `git worktree remove`, and `pathExists` correctly refuses to trust git's exit code alone. Recursion never descends into a found junction; three-levels-deep discovery covers the named `server/tts-sidecar/.venv` shape.
- House gates: `Closes #3051` plain and outside code spans; `docs/release-notes-next.md` entry present and accurate; `RELEASE_NOTES.md` skip declared with a reason; register row E103 plus glance/strip counts internally consistent; `npm run wt:gc` wired; `scripts/wt-gc.mjs` sits inside `spawn-windows-hide.test.ts`'s scanned `scripts/` tree, so that claim in the PR body is real; incidental findings (dead `findJunctions`, dead imports) declared.

---

### Verdict

**Request changes.**

The blocker is that `--prune`, as shipped, marks four open-PR agent worktrees and its own review worktree as prunable on this box at this head, because `refusalReasons` never reads the `merged` and PR columns the tool prints — and issue #3051's acceptance #1 names merged and ahead *as refusal cases* while the design doc names only three. Which refusal is correct is a real design decision worth its own pass; shipping the current default is not.

The four Significant correctness bugs are all small fixes: fail closed on an unresolvable primary check, exclude the running worktree, re-scan after junction removal instead of trusting a scan that swallows its own errors, and split the `unknown` PR sentinel in two. The seam test and the live-view row are chores the work made owed.

The core mechanism this PR exists to deliver — junction-first, `ReparsePoint`-gated, `Test-Path`-verified teardown — is right, and I confirmed it on 5.1 rather than taking the pwsh-7 suite's word for it.

dudarenok-maker:
## PR review — pass 2 (head 95d8ea4a, depth medium)

Scope: deltas only. Re-verified before probing: `node --test scripts/tests/wt-gc.test.mjs` **62/62**; `Invoke-Pester scripts/tests/wt-gc-junctions.Tests.ps1` **20/20** (Pester 6.1.0); `npm run check:onbox-register` **OK**; `npm run register:build -- --check` **up to date**. Then probed the six refusals against a purpose-built throwaway git repo with a bare remote and four worktrees (merged+pushed, detached, locked, never-pushed), driving the **production** `run()` in report mode with recorder stubs for `removeJunctions`/`removeWorktree`; the junction scan under **both** engines against real `New-Item -ItemType Junction` fixtures (394-char path, ACL-denied subtree at depth 2); and the seam pin by actually performing the one-sided rename. `--prune` never run against anything real. No tracked file modified.

---

### Pass 1 dispositions

|---|---|---|
| 🔴 | `--prune` marks unmerged / open-PR / own worktree prunable | **VERIFIED RESOLVED** |
| 🟠 1 | primary-checkout refusal fails OPEN on a git error | **VERIFIED RESOLVED** |
| 🟠 2 | no self-exclusion | **VERIFIED RESOLVED** |
| 🟠 3 | junction scan silently incomplete | **PARTIALLY RESOLVED — see 🔴 below and 🟡 2** |
| 🟠 4 | `unknown` conflates gh-unavailable with no-PR | **VERIFIED RESOLVED** |
| 🟠 5 | `.ps1` ↔ JS JSON seam unpinned | **VERIFIED RESOLVED** |
| 🟠 6 | live-view E103 is the generator's placeholder | **VERIFIED RESOLVED** |
| 🟡 a–d | header column list · dead `-Action Find` · `unpushedCount: null` · gh queried for refused rows | **VERIFIED RESOLVED** (d introduces 🟡 1 below) |

Evidence for the ones I did more than read:

- **Blocking.** I re-ran report mode myself and then checked the *reasons*, not the count, against git directly. The single `prunable? yes` row (`wt-2192-a29-genuine-install-test`) is independently confirmed: `merge-base --is-ancestor HEAD main` → exit 0, `@{upstream}` = `origin/fix/ops-2192-a29-genuine-install-test`, `rev-list --count @{upstream}..HEAD` → **0**, `status --porcelain` **empty**, PR #3041 **MERGED**. Every other row's refusal text matches its real state, including this PR's own tree (`the worktree this process is running from — never pruned; not merged into main`) and `wt-2616-onbox-group-c`'s `67 unpushed commit(s)`.
- **🟠 1.** Beyond the fail-closed `commonDir === null || gitDir === null` at `scripts/wt-gc.mjs:467`, I checked the *other* way `isPrimaryWorktree`'s raw string equality could fail open — a casing mismatch between `--git-common-dir` (taken with `cwd: repoRoot`) and `--git-dir` (taken with `cwd: tree.path`). It cannot: git 2.54 canonicalises `--path-format=absolute` output to the on-disk case regardless of the cwd spelling it was given (measured, invoking the same repo through a lowercased path). No finding.
- **🟠 2.** `isSelfWorktree` probed directly, both prefix directions: row `C:/wt/foo` + self `C:/wt/foo-bar` → **false**; row `C:/wt/foo-bar` + self `C:/wt/foo` → **false**; exact, cwd-inside, mixed-case, backslash-vs-forward and trailing-slash variants all correct. Sound.
- **🟠 5.** I performed the rename in a scratch copy of `scripts/` — `TargetStillExists` → `TargetSurvives`, **`.psm1` only**. JS: `the .psm1 emits every key JUNCTION_RESULT_KEYS names` → **✖**. Pester: `emits exactly the property set wt-gc.mjs declares` → **[-]** (`Error,Path,Removed,Target,TargetSurvives` vs expected), plus `reports TargetStillExists = $true after removal` → **[-]**. The two-sided claim holds exactly as stated.
- **🟠 6.** E103's live-view entry now matches its neighbours (plain-prose `iname`, `<ul>` criteria, `flag` block, `p.src`) and agrees with the markdown row; `body-placeholder` count in the file is back to **0**.
- **🟠 3 (partial).** The fail-closed throw is real and end-to-end under `pwsh`: an ACL-denied directory two levels below the root makes the production `.ps1` exit **3** with `wt-gc: junction scan could not enumerate '…' — the scan is INCOMPLETE…`, `makeDefaultRunners().removeJunctions()` throws on it, and no junction is unlinked before the throw (the scan completes before the delete loop). Recursive descent, `Get-Item` half and `Get-ChildItem` half all propagate. **But see the Blocking finding: on the other engine the module never loads at all.**

---

### 🔴 Blocking (new) — `wt-gc-junctions.psm1` does not parse under Windows PowerShell 5.1. `--prune` cannot prune anything on a box without `pwsh`, and this is a regression introduced by the pass-1 fix.

**Correctness bug.** `scripts/lib/wt-gc-junctions.psm1:79` and `:89` — the two `throw` strings added to close pass 1's 🟠 3 — each contain a literal **U+2014 EM DASH inside a double-quoted string**:

```powershell
throw "wt-gc: junction scan could not inspect '$Dir' — the scan is INCOMPLETE and must not be treated as 'no junctions found': $($_.Exception.Message)"
```

The file is UTF-8 **with no BOM** (`head -c 8` → `#require`, no `EF BB BF`), and `.gitattributes` has no rule for `*.psm1`. Windows PowerShell 5.1 decodes a BOM-less file as the ANSI code page, so `E2 80 94` becomes `â€"` — and that third byte **is a double quote**, which terminates the string literal mid-line.

Measured on this box (`powershell.exe`, 5.1.26100.9223):

```
$ powershell -NoProfile -Command "Import-Module .\scripts\lib\wt-gc-junctions.psm1 -Force; ..."
At ...\wt-gc-junctions.psm1:79 char:70
+ ...  throw "wt-gc: junction scan could not inspect '$Dir' â€" the scan is ...
+                                                               ~~~
Unexpected token 'the' in expression or statement.
ParserError: UnexpectedToken

Import-Module : ... was not loaded because no valid module file was found
The term 'Remove-JunctionsRecursive' is not recognized ...
```

End-to-end through the production `.ps1`, against the 394-char-junction fixture: exit **3**, junction still live, target's `PRECIOUS.txt` untouched. So it *is* fail-closed — but the consequence is that on any box where `pickPowerShell()` (`scripts/wt-gc.mjs:198-204`) falls through to `powershell`, **every** prunable tree fails with `junction removal failed` and `--prune` prunes nothing, ever. That fallback is not hypothetical: the module declares `#requires -Version 5.1`, `pickPowerShell()` explicitly falls back to it, and 5.1 is what a Windows box has unless someone installed PowerShell 7.

**It is a regression this PR introduced.** `git show fa1b1bf8:scripts/lib/wt-gc-junctions.psm1` has **no non-ASCII outside comments** and imports cleanly under 5.1 (measured: `LOADED OK`). Every other PowerShell file in `scripts/lib/` (`log-utils.psm1`, `prevent-sleep.psm1`, `sidecar-sweep-port.psm1`, `cast-reconcile.psm1`, `golden-bless-pytest-args.ps1`) keeps non-ASCII to comments only — where mojibake is harmless because `#` runs to end of line. These two lines are the only place in the repo's PowerShell where a non-ASCII character sits inside a string literal.

**Nothing in the suite can catch it**, which is why it shipped green: `pickPowerShell()` probes `pwsh` first, `scripts/run-powershell.mjs:38-39` (what `npm run test:scripts` uses) also prefers `pwsh`, and both this box and GitHub's `windows-latest` runners have it. The 20/20 Pester pass proves nothing about the engine the PR body calls "supported, not hypothetical".

**It also disarms the E103 acceptance criterion written to check exactly this.** The row says: *"run once under `pwsh` and once under `powershell` (5.1) and confirm both either find the junction or FAIL LOUDLY."* Under 5.1 today it does fail loudly — because the module will not load, not because the scan fails closed. The criterion cannot distinguish the two, so it would be recorded as discharged while the mechanism it exists to prove was never executed. (Separately: there is no CLI affordance to *select* the engine — `pickPowerShell()` always prefers `pwsh`, with no flag or env override — so "run it under 5.1" means renaming `pwsh` off PATH.)

Fix is one line each: `--` instead of `—` in those two strings (matching how every other `scripts/lib` PowerShell file already behaves), or a UTF-8 BOM on the file. A guard asserting `scripts/lib/*.ps*` carries no non-ASCII outside comments would keep the class shut — the same shape as the existing `.LinkTarget` source pin, which this PR already argues is the only kind of test that can catch a 5.1-only defect from a pwsh-7 box.

---

### 🟠 Significant (new) — a `git worktree lock`'d worktree reads `prunable? yes`; `--prune` strips its junctions and *then* discovers the lock

**Correctness bug.** `parseWorktreePorcelain` (`scripts/wt-list.mjs:25-43`) reads `worktree` / `branch` / `HEAD` / `detached` and **discards the `locked` line**, so `refusalReasons` (`scripts/wt-gc.mjs:276-289`) never sees git's own explicit "do not remove this" marker — even though it is right there in the porcelain the tool already parses.

Repro, in a throwaway repo, production `run()` in report mode:

```
$ git -C <repo> worktree lock <wt> --reason 'agent lane in flight'
$ git -C <repo> worktree list --porcelain
worktree .../FixLocked
branch refs/heads/feat/locked
locked agent lane in flight

$ node <probe driving production run()> <repo>
path           branch       merged  ahead  dirty  PR    prunable?
.../FixLocked  feat/locked  true    0      false  none  yes
```

Then the two prune steps, in the order `run()` runs them (`scripts/wt-gc.mjs:510`, then `:540`):

```
$ pwsh -File scripts/lib/wt-gc-junctions.ps1 -Root <wt> -Action Remove
{"items":[{"Path":".../FixLocked/node_modules","Removed":true,"TargetStillExists":true,...}]}
$ Test-Path .../FixLocked/node_modules            # False — junction gone

$ git -C <repo> worktree remove --force <wt>
exit=128
fatal: cannot remove a locked working tree, lock reason: agent lane in flight
use 'remove -f -f' to override or unlock first
```

The tree survives, **stripped of its junctions**, and `run()` reports `git worktree remove failed` + `failures += 1`. This is structurally the same defect as pass 1's 🟠 1 and 🟠 2: the destructive junction sweep runs *before* the only thing that honours the refusal, so git's backstop protects the directory and not the environment inside it. A locked tree is by definition one someone marked as in-flight; recovering it means re-junctioning `node_modules`, `server/node_modules`, `server/tts-sidecar/.venv` and `voices/` by hand.

Fix: parse `locked` in `parseWorktreePorcelain` (a fourth `startsWith`) and add it as refusal #7. This repo does not currently use `git worktree lock` in its own workflow — but the tool's whole thesis is that an explicit signal to leave a tree alone must be honoured *before* the destructive step, and this is the one such signal git itself provides.

---

### 🟡 Minor (new)

- **The gh-skip optimisation (pass 1 🟡 d) was applied to *report* mode too, where the answer is the product.** `scripts/wt-gc.mjs:477-486` sets `prQueried: false` for any row carrying another refusal regardless of `--prune`, so on this box **16 of 18 rows now render `not queried` in the PR column**. Pass 1's blocking repro was legible precisely because that column read `#3043 OPEN`, `#3057 OPEN`, `#3055 OPEN`, `#3031 OPEN`; today this PR's own row reads `not queried`. The file header (`:14-18`) still advertises "PR state (when `gh` is available and authenticated)" as a report column with no mention of the skip. Suggest: skip only under `--prune`, or say so in the header and the `--help` text. (cleanup/doc)
- **Pass 1's re-scan recommendation was half-taken, so the scan→delete window is still open.** `Remove-JunctionsRecursive` (`wt-gc-junctions.psm1:122-164`) enumerates once and then deletes; `run()` (`scripts/wt-gc.mjs:537-540`) goes straight from `removed N junction(s)` to `git worktree remove --force` with no re-scan. A junction created after the enumeration — an `npm install` or a `wt-new.mjs` finishing in that tree — is still live when the recursive remove runs. Narrow (the tree must be merged, pushed, clean and PR-closed for anything to be pruned at all), which is why this is Minor rather than a repeat of 🟠 3 — but it was the half of pass 1's suggested fix that closed the race, and it is now recorded nowhere. (cleanup)
- **The tool cannot see the thing the PR body sizes it against.** `git worktree list --porcelain` lists only *registered* worktrees, so an orphaned directory — one git no longer knows about, which is exactly what the 2026-09-06 sweep counted — is invisible to `wt-gc`. Concretely, right now: `ls -d C:/Claude/Projects/wt-*` → **18**, `git worktree list` → 17 `wt-*` plus the primary, and the difference is `C:\Claude\Projects\wt-2934-a36-audition-band`, which `npm run wt:gc` does not list. Issue #3051's acceptance list does not require orphan-directory cleanup, so this is not a missing feature — but the PR body ("Sized against a real problem: … 28 orphaned worktree directories totalling 8.27 GB"), `docs/release-notes-next.md`, `scripts/wt-gc.mjs:5-8` and both E103 rows all frame the tool as the answer to that sweep, and a reader will expect `wt:gc` to reclaim the 8.27 GB. One sentence in the header naming the boundary would settle it. (doc — reviewer-brief catalogue #9)

---

### Verdict

**Request changes**, on the new Blocking finding alone.

Every pass-1 finding is genuinely fixed, and the two I could most easily have taken on trust — the seam pin's two-sided rename and the "exactly one prunable row" claim — I reproduced independently, and both hold. The refusal set now reads the columns it prints, and I could not construct a tree that is unsafe to prune and clears all six *by the refusals' own logic*: detached HEAD, no upstream, a stale local `main`, and an unverifiable push count all land on the safe side, and the primary/self checks survive the path-shape attacks I could think of.

What I could construct is a tree that clears all six because a signal was never read (the `locked` marker), and — the blocker — a supported engine on which the fail-closed scan this round added cannot run at all, broken by the very commit that added it, invisibly to a 62 + 20 green suite and to the acceptance row written to check that engine.

dudarenok-maker:
## PR review — pass 3 (head 2620e873, depth medium)

Scope: deltas only — the pass-3 commit `2620e873` (`ps-ascii-scan.mjs` + its suite, the `locked` refusal, the post-delete re-scan, the report-mode `gh` revert, the E103 rewrite). Re-verified before probing: `node --test scripts/tests/wt-gc.test.mjs scripts/tests/ps-ascii-scan.test.mjs` **83/83**; `Invoke-Pester scripts/tests/wt-gc-junctions.Tests.ps1` **22/22** (Pester 6.1.0); `npm run check:onbox-register` **OK**; `npm run register:build -- --check` **up to date**. Then probed the new lexer with 31 adversarial sources, the guard's file set against every tracked `.ps1`/`.psm1` in the repo, its CI scheduling through the production `computeScopes`, the `locked` refusal against a purpose-built throwaway repo with three worktrees (locked-with-reason / locked-bare / unlocked), the re-scan through production `run()`, and E103's criteria by executing them verbatim under real Windows PowerShell **5.1.26100.9223**. `--prune` never run against anything real. No tracked file modified.

---

### Pass 2 dispositions

|---|---|---|
| 🔴 | `wt-gc-junctions.psm1` unparseable under Windows PowerShell 5.1 (em dash in two `throw` strings) | **VERIFIED RESOLVED** |
| 🟠 | a `git worktree lock`'d tree reads `prunable? yes` | **VERIFIED RESOLVED** |
| 🟡 1 | the `gh` skip blanked the PR column in *report* mode | **VERIFIED RESOLVED** |
| 🟡 2 | the scan→delete window left open | **VERIFIED RESOLVED** |
| 🟡 3 | orphaned directories are invisible and the framing did not say so | **VERIFIED RESOLVED** |

Evidence for the ones I did more than read:

- **🔴.** `powershell.exe -NoProfile -Command "Import-Module .\scripts\lib\wt-gc-junctions.psm1 -Force; …"` → **`LOADED OK`** at this head, on 5.1.26100.9223. Independently: I fed `git show 95d8ea4a:scripts/lib/wt-gc-junctions.psm1` to the new scanner and it reports exactly **2** findings, at `:79:66` and `:89:68` — the two lines pass 2 named — so the guard genuinely catches the historical defect rather than being green by construction. I also confirmed the *class* is real and wider than the em dash: a BOM-less file with an **EN DASH** (U+2013 → CP1252 `0x93` → U+201C) in a **double**-quoted string, and one with **U+2011** (→ `0x91` → U+2018) in a **single**-quoted string, each produce `ParserError: TerminatorExpectedAtEndOfString` under 5.1. The blanket "no non-ASCII in any string literal" rule is the right rule, and the three em-dash→`--` edits in `sidecar-sweep-port.Tests.ps1` are not over-strictness.
- **🟠 locked.** Throwaway repo, three worktrees, production `parseWorktreePorcelain` + `run({prune:true})` with recorder stubs. `locked agent lane in flight` → `locked=true, lockReason="agent lane in flight"`; bare `locked` → `locked=true, lockReason=""` (and `refusalReasons` still refuses on the empty reason, rendering `` no (locked by `git worktree lock`) `` with no dangling colon); the tree listed immediately after both reads `locked=false`, so there is no bleed. `--prune` printed `SKIP` for both and `removeJunctions` / `removeWorktree` were called **zero** times.
- **🟡 2 re-scan.** Confirmed it cannot be deduped away, from both sides. Module side: `$reported` only collects paths whose *delete failed*, so a path that was removed and then re-created is not suppressed — it lands as a second entry with `Removed=$false`. JS side: I drove production `run({prune:true})` with a `removeJunctions` stub returning the re-scan entry key-for-key as the `.psm1` emits it, alongside a successful entry for a different path. Result: `junction NOT cleanly removed: …`, exit **1**, `removeWorktree` calls **`[]`** — the abort really does happen before `git worktree remove`.
- **🟡 1 + the refusal set as a whole.** Live report mode on this box at this head: **18 rows, zero `prunable? yes`**, every PR cell populated (`#3065 OPEN`, `#3057 OPEN`, `#3031 MERGED`, `none`, …), and every refusal string matching real state — including this tree's own `no (the worktree this process is running from — never pruned; not merged into main; open PR #3055)`. 28.1 s wall clock for the 18 serial `gh` calls.

I could not construct a tree that is unsafe to prune and clears all seven refusals, beyond the `--limit 1` case in Minor 3 below. Detached HEAD, no upstream, an unverifiable push count, a stale local `main` (`wt-onbox-mechanical-batch1` reads `not merged into main` with PR #3031 **MERGED** — erring toward refusing, exactly as the header documents), a missing worktree directory, and a mid-rebase tree all land on the safe side.

**The shipped `wt-gc` tool itself I could not break this round.** Everything below is about the *guard* this round added and the *acceptance row* that is meant to prove it.

---

### 🟠 Significant (new) — the new ASCII guard is not scheduled by the diff class it exists to police. A PowerShell-only PR skips it in CI and reads `[cached]` locally.

**Correctness bug (wiring).** `ps-ascii-scan.test.mjs` reads `scripts/lib/*.ps1`, `*.psm1` and `scripts/{lib,tests}/*.Tests.ps1` with `readFileSync` at runtime — no module-graph edge. But `verify-cache.mjs`'s `STEPS[]` entry for `test:hooks` declares its inputs as:

```
scripts/**/*.{mjs,cjs,js,mts,cts,ts}
scripts/tests/fixtures/**
pinokio-scripts/**
.github/workflows/**  .github/actions/**  .husky/**
.claude/skills/**  .claude/agents/**  docs/testing/**  docs/features/**
```

No `.ps1`, no `.psm1`. Measured with the **production** resolver (`computeScopes` from `scripts/ci-scope.mjs`, which `verify.yml`'s `detect` job uses):

```
this PR                      : shared=true  step_test_hooks=true
a future .psm1-only fix      : shared=false step_test_hooks=false
a future .psm1 + .Tests.ps1  : shared=false step_test_hooks=false
scripts/start-app.ps1 only   : shared=false step_test_hooks=false
server/tts-sidecar/start.ps1 : shared=false step_test_hooks=false
```

`stepTouchedByDiff(test:hooks, ['scripts/lib/wt-gc-junctions.psm1'])` → **`false`**.

So: a follow-up PR that edits only `scripts/lib/wt-gc-junctions.psm1` — the exact diff that reintroduces pass 2's blocker — does **not** schedule the `test:hooks` leg in `verify.yml`, and locally `verify-cache.mjs` prints `test:hooks [cached]`. This PR runs the guard only because it touches root `package.json`, which trips `computeShared` and puts every step in scope; that is incidental to it, not a property of the guard.

This is the trap `verify-cache.mjs`'s own comments document twice on this very entry — *"a fixture-only diff would skip the very test it adds"* (#1847) and *"a pinokio-scripts-only diff prints `test:hooks [cached]` on exactly the diff that introduced it"* (#2216). Both were fixed by widening these globs; this one is the third spelling. Fix is one line: add `scripts/**/*.{ps1,psm1}` to that `inputs.globs` array.

---

### 🟠 Significant (new) — the guard's own test name is false of the repo today: `scripts/start-app.ps1` is under `scripts/` and carries four em dashes inside double-quoted strings

**Correctness bug (class left armed).** The test is named `no PowerShell source under scripts/ carries a non-ASCII character outside a comment`, and its failure message states the rule as a repo-wide invariant. `powerShellSources()` (`scripts/tests/ps-ascii-scan.test.mjs:31-42`) does not implement that: it `readdirSync`es exactly two directories, non-recursively, and keeps `scripts/lib/*.ps?1` plus `*.Tests.ps1`. I ran the production scanner over all **30** tracked `.ps1`/`.psm1` files. It covers 18; **12 are outside it**, four of them under `scripts/` itself:

```
scripts/oe-detached-commit.ps1        scripts/reconcile-broken-cast.ps1
scripts/start-app.ps1                 scripts/stop-app.ps1
scripts/tests/run.ps1                 server/tts-sidecar/{run-tests,run-golden-tests,start,sidecar-restart-policy}.ps1
server/tts-sidecar/scripts/install-{coqui,kokoro,qwen3}.ps1
```

`scripts/start-app.ps1` — reached by `npm start`, the repo's primary entry point, and `#requires -Version 5.1` — currently violates the stated rule four times:

```
scripts/start-app.ps1:40:41  U+2014 in double-quoted string -- Write-Status "`nCastwright v$pkgVersion — Any book, performed by a full cast.`n"
scripts/start-app.ps1:67:52  U+2014 in double-quoted string
scripts/start-app.ps1:141:100 U+2014 in double-quoted string -- Fail "Port :$($svc.Port) is occupied … — likely a stale or foreign server…"
scripts/start-app.ps1:144:120 U+2014 in double-quoted string
```

It is safe today only because it is the **one** PowerShell file in the repo carrying a UTF-8 BOM (`EF BB BF`), so 5.1 decodes it as UTF-8. The scanner does not model that — it reports the BOM itself as a `U+FEFF … in code` finding at `1:1`. Two consequences, and they pull against each other:

1. Widening the file set (the obvious fix for the gap) turns the guard **red** on `start-app.ps1` — five findings, none of them a live defect.
2. The *other* documented fix for this hazard, named in `wt-gc-junctions.psm1`'s own header and in pass 2's finding ("or a UTF-8 BOM on the file"), is unusable inside the guarded set: adding a BOM to any `scripts/lib` PowerShell file reddens the guard.

`server/tts-sidecar/start.ps1` and `scripts/stop-app.ps1` are BOM-less and carry non-ASCII in comments only, so nothing else is live today — but a future em dash in a `throw`/`Write-Host` string in either reproduces pass 2's defect exactly, on files this guard never opens. Either narrow the test's name and message to the set it actually scans, or widen the set and teach `isNonAscii` to skip a leading U+FEFF.

---

### 🟠 Significant (new) — E103's new "distinguish the two ways 5.1 can fail loudly" criterion cannot be executed as written, on either surface

**Chore — house gate (before-shipping step 3), on a merge-gate surface that has not been published yet.** Two independent defects in the one bullet, which appears verbatim in both `docs/testing/onbox-acceptance-register.md:4818-4830` and `docs/testing/onbox-acceptance-register-live-view.html:1298`.

**(a) The precondition command fails, with the exact signature the criterion says disqualifies a pass.** The criterion reads: *"First confirm `powershell.exe -NoProfile -Command "Import-Module scripts\lib\wt-gc-junctions.psm1 -Force"` loads clean, THEN exercise the scan"*, having just told the operator that *"a `ParserError` / `Import-Module` failure means the MODULE did not load … which is NOT a pass for this bullet."* Run verbatim from the repo root:

```
$ powershell.exe -NoProfile -Command "Import-Module scripts\lib\wt-gc-junctions.psm1 -Force"
Import-Module : The specified module 'scripts\lib\wt-gc-junctions.psm1' was not loaded because
no valid module file was found in any module directory.
    + FullyQualifiedErrorId : Modules_ModuleNotFound,…
exit=1
```

`Import-Module` treats a bare relative path as a *module name* and searches `$env:PSModulePath`; it needs a leading `.\`. Same failure on `pwsh` with the forward-slash spelling. With `.\` it is `LOADED OK`. (Verified the child process's cwd really was the repo root, so this is not a harness artefact.) As written, an operator following this criterion gets a guaranteed **false FAIL** whose text is character-for-character the disqualifying signature the criterion was rewritten this round to teach them to recognise.

**(b) The two-engine run is single-shot now that `-Action Find` is gone.** The criterion says *"run once under `pwsh` and once under `powershell` (5.1) and confirm both either find the junction or FAIL LOUDLY"*, and its own parenthetical says that means *"invoking the `.ps1` through `powershell.exe` by hand"*. But `wt-gc-junctions.ps1`'s `[ValidateSet('Remove')]` leaves `Remove` as the only action (pass 1's 🟡 fix). The `pwsh` run therefore **unlinks the junction the 5.1 run needs**, so the second run legitimately finds nothing — which the same bullet defines as a failure. The `.ps1` header already names the read-only alternative (*"import the module and call `Get-JunctionsRecursive` directly"*); the criterion does not, and the register is what the operator will follow. This matters more than usual because the live view **has not been published yet** — the PR body still lists that as owed — so this text is what would go onto the artifact URL.

---

### 🟡 Minor (new)

- **One lexer shape I could defeat: a `#` opening a continuation line of a multi-line PowerShell string.** PowerShell string literals *do* span lines; `scanPowerShellNonAscii` resets `state` to `'code'` at every EOL (`scripts/lib/ps-ascii-scan.mjs:135`), so the next line's leading `#` is read as a comment (`:111`) and everything after it is skipped. Measured, BOM-less UTF-8:

  ```powershell
  $msg = "Castwright start
    # step 2 — launch the sidecar
    done"
  ```
  → scanner: **`[]`** (zero findings). → `powershell.exe` 5.1: `ParserError: Unexpected token 'launch'` + `The string is missing the terminator: "`. A `<# … #>` on such a continuation line is missed the same way. This is the exact defect class the guard exists to close, passing clean. The source comment at `:131-134` does disclose the limitation ("does not span lines in PowerShell in any shape this repo writes" — true today, I checked), which is why this is Minor rather than higher; but the test name and the failure message state it as an unqualified invariant. Everything else held: 31 adversarial sources — `#` and `'` inside double-quoted strings, backtick escapes, `''`/`""` doubling, a here-string containing its own and the other terminator, leading-whitespace pseudo-terminators, `<#` inside strings, `#>` with no opener, `$( "nested" )`, `@{`/`@(`, `@"literal"@` on one line — all classified correctly, in both directions.
- **`ghSpawn` carries no `timeout`, and report mode now makes one blocking call per branched row.** `scripts/gh.mjs:86-95` passes no `timeout` to `spawnSync`, and `scripts/wt-gc.mjs:188-192` adds none. A `gh` that *hangs* rather than fails — blackholed DNS, a captive portal, a stalled TLS connect — wedges the read-only report indefinitely, and the file header's offline-tolerance paragraph promises the report "never aborts", which covers `gh` failing, not `gh` hanging. Measured cost of the (correct) pass-3 revert on this box: **28.1 s** for 18 serial calls. `spawnSync`'s `timeout` option treated as `available:false` is a one-line fix and preserves the fail-closed reading.
- **`gh pr list --state all --limit 1` answers "is the *most recent* PR open", not "is *any* PR open".** `scripts/wt-gc.mjs:190` requests `--limit 1` and `:196` takes `rows[0]` and `normalizePrInfo` sets `open` from that single row's `state`. A head branch with two PRs — one to `main` still open, a newer one to another base already merged — renders `#N MERGED` and clears refusal 6. I did not construct this (it needs two real PRs off one branch), so I am reporting the mechanism from the code, not a repro: `--limit 2` plus `rows.some(r => r.state === 'OPEN')` would close it.
- **`ps-ascii-scan.mjs`'s header states the mechanism in a way that understates the class.** `:6-7` says the em dash's third CP1252 character *"is `"` — which terminates an enclosing double-quoted string"*. Byte `0x94` is not ASCII `0x22`; it is U+201D, and it closes the literal because **PowerShell accepts smart quotes as string delimiters**. Stating it the accurate way is what makes the rest of the class visible: I measured U+2013 (`0x93` → U+201C) breaking a *double*-quoted string and U+2011 (`0x91` → U+2018) breaking a *single*-quoted one. A future reader could take the header at face value and conclude only em dashes in double-quoted strings matter. (Catalogue #9 — the defect is in the document; the rule the file enforces is correct and broader than its own example.)

---

### Verdict

**The tool is merge-ready; the guard's wiring is not.** I spent this pass trying to break `wt-gc` itself and could not: the seventh refusal reads git's own marker before the destructive step and does not bleed across records, the re-scan aborts the prune before `git worktree remove` and cannot be deduped away, report mode is honest again, and the 5.1 blocker is genuinely gone on the engine that had it. Live report at this head: 18 rows, **zero** prunable, every refusal string true of its tree.

Every finding above is about the *guard* and the *acceptance row*, and all three 🟠s are one- or two-line fixes:

1. add `scripts/**/*.{ps1,psm1}` to `test:hooks`'s `inputs.globs` — without it the guard is not scheduled on the only diff shape that can reintroduce the defect, and this PR runs it only by accident of touching `package.json`;
2. reconcile the guard's name with its file set (or widen the set and skip a leading BOM) — `scripts/start-app.ps1` is a live counter-example to the invariant the test claims;
3. fix E103's `Import-Module` command (`.\`) and its two-engine instruction — **this one before the live-view publish**, which the PR body already lists as an owed pre-merge step, since the criterion as written produces a guaranteed false FAIL and the second engine has no junction left to find.

This is pass 3 of the 3-pass cap, so per the runbook these go to the repo owner rather than round 4. My reading: none of them endangers anything at this head, and (1) is the one I would not merge without, because a guard that is not scheduled on its own subject matter is the thing this round is shipping.

Issue #2910, opened 2026-09-04
[agent instructions][claude][task] A34 step 2 - implement + test the wrong-direction repair script


dudarenok-maker:
Step 2 of the A34 repair-and-retest chain (#2903), on-box acceptance register
campaign (#2435).

## What

Implement, test-first, a repair script for the wrong-direction
`characterId` retirement defect: an already-established id that should be
ASCII per name-equivalence, but currently resolves to its non-ASCII
(typically Cyrillic) sibling instead — the pre-existing-corruption shape PR
#2640 (`server/src/analyzer/roster-dedup.ts`,
`stripEstablishedAsciiRewrites`) prevents going forward but cannot repair.

## Read first

- `server/src/analyzer/roster-dedup.ts` — `stripEstablishedAsciiRewrites` and
  `normaliseForMatch`, the exact comparator this repair must reuse for
  detecting the wrong-direction shape. **Do not invent a second
  name-equivalence rule** — import or replicate this one exactly, or the
  repair script and the fix it re-tests will disagree about what "the same
  character" means.
- `scripts/repair-cast-id-drift.mjs` — the house pattern for this class of
  script: liveness-probe refusal when a server is live (do not skip this —
  it is the safety net that stops a script from racing a live server's own
  writes), `parseArgs({apply})`, per-book planning function, dry-run by
  default, real filesystem reads via `WORKSPACE_DIR`/`BASE`/
  `AUDIOBOOK_WORKSPACE` env vars (default `<home>/AudiobookWorkspace`). Follow
  its shape; this is a **different defect class**, so write a new script
  (`scripts/repair-a34-wrong-direction-ids.mjs`) rather than bolting a second
  mode onto the existing one, unless you find the two share enough real code
  that a shared helper module is cleaner — your call, but don't force it.
- Issue #1's investigation output (step 1, `docs/testing/onbox-a34-results/
  step-1-scope.md`) for the actual shape of affected entries in this
  workspace, so your detection logic is grounded in real data, not a
  hypothetical.

## What to build

- A detector: given a book's `cast.json` + `cast-id-history.json`, find any
  character whose **currently-live** id is non-ASCII while `supersededBy` (or
  the equivalent live history) shows an ASCII id was retired in its favour,
  AND the two ids are the same character by `normaliseForMatch`. This is the
  reverse of what `stripEstablishedAsciiRewrites` prevents going forward.
- A repair action: reinstate the ASCII id as live, retire the non-ASCII one
[...]

Issue #2903, opened 2026-09-04
ops — A34: repair pass for wrong-direction characterId retirements, then re-test the fix


dudarenok-maker:
On-box acceptance register row **A34** ([#2584](https://github.com/dudarenok-maker/Castwright/issues/2584), [#2040](https://github.com/dudarenok-maker/Castwright/issues/2040), PR [#2640](https://github.com/dudarenok-maker/Castwright/pull/2640)) cannot be discharged as worded: its criterion asks PR #2640's fix to *repair* a book (*Заказ Коалфолла*) whose bad Cyrillic id (`одуван` instead of `oduvan`) was already established **before** the fix shipped. The fix (`stripEstablishedAsciiRewrites`, `server/src/analyzer/roster-dedup.ts`) is deliberately scoped to *prevent* a fresh wrong-direction retirement, not repair a pre-existing one — confirmed by a real re-analysis run (wave 8, 2026-08-27): the fresh analyzer run re-proposed `одуван` again, so no retirement event ever fired for the fix's guard to intercept.

## Owner ruling, 2026-09-05

Presented three options (re-scope to a different book / build a repair pass then re-test / discharge as unsatisfiable). **Chosen: build a repair pass, then re-test.**

## Scope

1. Investigate: find every book in the real workspace with this same wrong-direction shape (an established id that should be ASCII per the same name-equivalence comparator `stripEstablishedAsciiRewrites`/`normaliseForMatch` use, but is currently the non-ASCII survivor) — not just *Заказ Коалфолла*.
2. Implement a repair script, test-first, dry-run by default — house pattern is `scripts/repair-cast-id-drift.mjs` (liveness-probe refusal when the server is up, `--apply` flag, per-book snapshot evidence, `parseArgs`/`planBookRepairs`-shaped structure). This is a **different defect class** from that script's own (wrong-direction ASCII/non-ASCII retirement, not orphaned segment ids) — a new script, not an extension, unless investigation finds they're closely enough related to share code cleanly.
3. Dry-run against the real workspace, review.
4. Back up affected books' `cast.json`/`state.json` (byte-verified), apply for real, verify each write.
5. Re-test A34's actual criterion: re-analyze the repaired book and confirm the id now resolves ASCII and the fix's guard holds against a repeat flip.
6. Fold into the register; discharge or narrow A34 based on what's actually observed.

Decomposed into an agent-executable chain — see linked children.

PR #3057, opened 2026-09-06
feat(scripts): A34 step 2 - repair wrong-direction characterId retirement

### LINKED ISSUE #3095: A same-day retry of repair-cast-id-drift.mjs destroys the pre-repair backup it points at
## What

`scripts/repair-cast-id-drift.mjs`'s `backupCastIdHistory` (≈`:2381-2387`) stamps its pre-repair backup **date-only** and copies with a plain `copyFileSync` — no `COPYFILE_EXCL`. A second run on the same day silently overwrites the first run's pre-repair copy.

## Why it matters

This script rewrites real `cast.json` / `cast-id-history.json` identity data in a live workspace, and that backup is the only thing standing behind the write. The destructive sequence is the one the script's own error message invites:

1. A run dies part-way. In practice that is usually a `LockAcquisitionTimeoutError`, which has a 10 s per-acquisition budget and fires under ordinary contention — common, not exotic.
2. The error reads as transient, so the operator retries the same day.
3. Run 2 copies the **half-repaired** file over the only pre-repair copy.
4. Run 2's error still names that file as the thing to go back to.

An undo a retry can destroy is not an undo.

## How it was found

Surfaced in passing while fixing the **identical** defect in the sibling `scripts/repair-a34-wrong-direction-ids.mjs`, reported as finding D2 of PR #3057's review pass 2. Same class, second site — which is itself the pattern: pass 1 of that same PR found a discarded return field (`bakAvailable`), pass 2 found another one (`droppedBooks`) in the same file.

## Fix

Mirror what landed in the sibling in `a87c5dc2`: millisecond-resolution stamp, `COPYFILE_EXCL` on the copy, retry at a suffixed path on `EEXIST`, and throw rather than fall back to overwriting if no candidate is free. Invariant: an existing pre-repair copy is never overwritten, and an attempt to do so never passes silently.

Paired regression test required — nothing covers a second run today, in either script.

## Acceptance

1. A second `backupCastIdHistory` call on the same day lands at a distinct path; the first run's copy is byte-unchanged.
2. An exact stamp collision retries rather than overwriting.
3. Exhausting the candidate range throws.
4. Any scanner that picks up `*.bak.*` files as evidence still matches the new stamp shape.
5. The sibling's doc comment, which currently says it deliberately does *not* mirror this function, is corrected — it is stale once both are fixed.

Fixed in the same round as the finding that surfaced it, per CLAUDE.md "Incidental findings: report, fix, record"; this issue is the bookkeeping, not a deferral.

dudarenok-maker:
## Summary

Implements the detector + repair script for the A34 wrong-direction `characterId` retirement defect (step 2 of the #2903 repair-and-retest chain): an ASCII-kebab id already established as live, retired in favour of its non-ASCII sibling by prior analysis. Reuses `stripEstablishedAsciiRewrites`'s own name-equivalence comparator so detection and the forward-going fix (#2640) agree on "the same character." Dry-run by default; `--apply` refuses when a server is live on the resolved port, mirroring `repair-cast-id-drift.mjs`'s liveness-probe pattern.

**This PR also changes the already-shipped `scripts/repair-cast-id-drift.mjs`**, not only adds a new script. Both changes are declared below.

### Also in this PR: the shared liveness probe's port-skip list

`probePortRangeRefused` (in `repair-cast-id-drift.mjs`, used by both repair passes) carried a hardcoded `KNOWN_STANDING_PORTS = new Set([8090])` — an operator exception for `llama-swap`. 8090 is **also this repo's own worktree slot-1 `PORT`** (`wt-new.mjs`'s `BASE_PORTS` + `PORT_STEP`, asserted at `scripts/tests/wt-new.test.mjs:83`, documented in CLAUDE.md) and sits inside the default 8080 auto-rebind walk, so a global constant made an apply-time safety probe permanently blind to a port a real Castwright server binds — for **every** consumer of the helper, including the already-shipped sibling's own `--apply`. It is now a per-run `ALLOW_STANDING_PORTS` opt-in, empty by default, with tests (there were none). Chosen over simply deleting 8090 because the llama-swap collision is real on the operator's box; making it per-run and explicit confines the blindness to the run whose operator asserted it.

### Review pass 1 — five correctness bugs fixed

- **C1 · The whole `--apply` write path was untested**; the test re-implemented `applyBookPlan` inline and asserted on its own copy. Four mutations of the real function left the suite green, including swapping `retireCharacterId`'s last two arguments — which drives its *forward* branch and re-inflicts the exact A34 defect this script exists to repair. `applyBookPlan` and `backupBeforeApply` are now exported and driven directly; the inline copy is gone. All four mutations redden a named test.
- **C2 · The evidence guard failed open.** `collectBakNameEntries`'s `bakAvailable` was destructured away, so one unparseable `cast.json.bak.*` — swallowed to `null`, contributing zero entries — let `buildNameIndex`'s `normSet.size > 1` read the survivors as *unambiguous* rather than *unknown*, turning lost evidence into a **confirmed** repair. `planBookRepairs` now gates on it (`bak-evidence-unreadable`), as the sibling script already does, and the module doc's claim about the unreadable case is now true.
- **C3 · The same-name confirmation had no uniqueness rule.** A display name shared by two live rows (Солдат, Стражник, the `unknown-*` buckets) auto-confirmed, binding the retired id — and every attribution behind it — to possibly the wrong character, at a different `voiceUuid`. `resolveTierAName`'s tie rule is now mirrored (`live-name-not-unique`).
- **C4 · `KNOWN_STANDING_PORTS`** — see the section above.
- **C5 · No atomicity or backup across the two-file write.** Order is reversed to history-first: if `retireCharacterId` throws (`CastIdHistoryUnreadableError`, its own `withKeyLock` timeout, an AV-scanner EPERM), cast-first leaves the live id ASCII while the history still targets the dead non-ASCII one, which `buildCastResolver` drops — orphaning *every* attribution in the book. Both files are now copied to `<file>.bak.a34-<date>` before either write, and a part-way failure names the copies.

Cleanup nits from the same pass: the `ascii-id-already-live` branch now has a test; both release-notes documents carry an entry; the three `docs/testing/onbox-a34-results/` evidence docs are now reachable from register row A34 via its cited run sheet (new §10.2b in `cast-id-drift-onbox-acceptance.md`), and `step-2-implement.md` carries a review-update section so it stops describing the first draft as though it were shipped.

Refs #2903, #2435. Refs #2910 (this is step 2 of that chain; not closing it via this PR — the issue is closed directly per its own completion instructions).

## Review pass 2 — both blocking findings fixed, plus a third they surfaced

**D1 (🔴) — a corrupt book read as a clean workspace** (`6c9e5f20`). `main()` destructured
`droppedBooks` away from `collectBooks()`, so a book whose `cast.json` exists but is truncated was
dropped — its evidence LOST — and never mentioned. A one-book workspace in exactly that state
printed `books scanned: 0` and `0 confirmed wrong-direction pairs — nothing to repair`, exit 0,
**under `--apply`**. Dropped books are now named, `--apply` refuses outright on an unreadable one
before any write path is reached, and a zero-book scan reads as *nothing was examined* rather than
as an unremarkable `0`. The sibling `repair-cast-id-drift.mjs` already closed both halves (#2097,
#2108), so its helpers are imported rather than reimplemented — the same operator reads both
transcripts, and `docs/testing/onbox-a34-results/step-3-dry-run.md` is committed acceptance
evidence built from this exact summary line.

This is the same shape as pass 1's C2, where `bakAvailable` was destructured away from the same
kind of return value. Second site of one class.

**D2 (🟠) — the backup was destroyed by the retry its own error invites** (`a87c5dc2`).
`backupBeforeApply` stamped date-only and copied without `COPYFILE_EXCL`. The sequence: a run dies
part-way on pair 2; the error reads like the transient lock timeout it usually is (a
`LockAcquisitionTimeoutError` has a 10 s per-acquisition budget and fires under ordinary
contention); the operator retries the same day; run 2 copies the **half-repaired** files over the
only pre-repair copy — and run 2's error still names that file as "Pre-repair copies". An undo a
retry can destroy is not an undo, and `RELEASE_NOTES.md`'s "so there is always something to go
back to" was the sentence that failed. Now a millisecond stamp **and** `COPYFILE_EXCL`: resolution
alone leaves a same-millisecond window and a mocked clock defeats it outright, so EXCL closes it
unconditionally, a collision retries at a suffixed path, and an exhausted range throws rather than
falling back to overwriting.

`repair-cast-id-drift.mjs`'s `backupCastIdHistory` carried the **identical** defect (`c4615ecb`) —
surfaced while fixing D2. Fixed in this round rather than filed and left: it is a defect this work
surfaced, in a file this branch already touches, and CLAUDE.md rules out every reason to defer it.
**Closes #3095**, which records it rather than substituting for it. Two comments in the a34 script
asserting it deliberately does *not* mirror that function were true when written and false the
moment both matched; corrected in the same commit.

Confirmed rather than assumed: every message naming a backup interpolates the returned path, so
none needed updating; and `collectBakNameEntries` filters on names starting with `cast.json.bak`
while this backup is `cast-id-history.json.bak.id-drift-*`, so it was never bak evidence before
and still isn't — now stated in the doc comment so it doesn't get "fixed" later.

### Worth stating plainly

**Every finding on this PR has had a second site.** Pass 1 found a discarded return field; pass 2
found another one in the same file; D2's backup bug had a twin one file over. The shape is the
finding, not the site.

Each of the five new tests was watched go red against a revert of the fix it covers and green
again after. Suites: **48 pass** (a34) and **190 pass** (drift), 0 fail.

## Test plan

- [x] `node --test scripts/tests/repair-a34-wrong-direction-ids.test.mjs` — **43/43** passing (was 25 at PR open, 27 at review). Covers the real `applyBookPlan`, the `bakAvailable` gate end-to-end through `main()` with a genuinely truncated bak file on disk, the uniqueness tie rule, the `ascii-id-already-live` branch, the backup/order/part-way-failure contract, and `ALLOW_STANDING_PORTS` against a real `net` listener.
- [x] `node --test scripts/tests/repair-cast-id-drift.test.mjs` — 187/187, unchanged by the `ALLOW_STANDING_PORTS` swap.
- [x] `npm run test:hooks` — 2000/2000.
- [x] Mutation-verified: each of the ten production tokens above was broken, the named test watched go red, and the token restored. `--apply` was never pointed at a real workspace; every fixture is a fresh temp directory.
- [ ] `npm run verify` — required status check


dudarenok-maker:
## PR review — pass 1 (head c83c7c1c, depth medium)

Reviewed at `c83c7c1c`, worktree `C:\Claude\Projects\wt-2903-a34-repair`, tree clean before and after. No tracked file modified. All mutation work ran against a copy of `scripts/` in a temp directory, never the worktree, and the apply path was never pointed at a real workspace.

**5 correctness bugs, 4 cleanup nits.**

---

## Correctness bugs

### C1 · Critical · The entire `--apply` write path is untested. The test that claims to cover it re-implements it.

`scripts/tests/repair-a34-wrong-direction-ids.test.mjs:386-451` · `scripts/repair-a34-wrong-direction-ids.mjs:310-333`

The test named `apply path: cast.json id is renamed and retireCharacterId is called from non-ASCII to ASCII` never calls `applyBookPlan`. It is not in the import list (`:21-28`) and is not exported. The test copies `applyBookPlan`'s body into itself at `:423-432` and asserts on its own copy. The comment says so out loud — "Re-implements applyBookPlan's exact contract" — but a re-implementation is not coverage: it is a second, independent implementation that happens to agree with the first, and nothing in the suite compares them.

Mutations against a copy of this branch's `scripts/` in a temp dir, `node --test scripts/tests/repair-a34-wrong-direction-ids.test.mjs`, baseline **27/27 green**:

| Mutation to `repair-a34-wrong-direction-ids.mjs` | Result |
|---|---|
| `:331` → `retireCharacterId(bookPlan.bookDir, r.asciiId, r.nonAsciiId)` (arguments swapped) | **27/27 pass** |
| `:310-333` → `async function applyBookPlan(bookPlan, mods) { return; }` | **27/27 pass** |
| `:323` → `if (false) {` (partial-repair refusal deleted) | **27/27 pass** |
| *control:* `:359` → `if (false && notRefused.length) {` (liveness refusal neutered) | **26/27 — 1 fail** |

The control reddens, so the harness works. The three above are genuinely invisible.

The first mutation is the one that matters. Swapping those two arguments does not merely break the repair — it drives `retireCharacterId`'s **forward** branch instead of its direct-reversal branch and writes `supersededBy["oduvan"] = "одуван"`, i.e. it re-applies the exact A34 defect this script exists to remove, against a book whose `cast.json` it has already rewritten. The suite that ships with the script cannot tell that apart from a correct repair.

`applyBookPlan` is `main()`'s only writer of real book identity data and is executed by zero tests. Exporting it and driving it with the `fakeMods` already in the file is a small change; asserting on a private copy of it is not a substitute.

### C2 · High · `bakAvailable` is discarded, so a corrupt `cast.json.bak.*` turns ambiguous evidence into a confirmed repair.

`scripts/repair-a34-wrong-direction-ids.mjs:378`

```js
const { entries: bakEntries } = collectBakNameEntries(book.audiobookDir);
```

`collectBakNameEntries` returns `{ entries, bakAvailable }`. The second field is the fail-closed gate: it goes `false` when the directory cannot be enumerated **or when at least one bak file that exists fails to parse or shape-check** — see its own doc comment at `scripts/repair-cast-id-drift.mjs:1749-1799`, which spells out that this is #2135's exact repro and that "an unread file could have named ANY id ambiguous, so this function can't tell guard 2 which ids are safe." The sibling script consumes it at `repair-cast-id-drift.mjs:2605-2606` and `:2641-2648` and withholds every matched id for the book. This script never reads it.

Demonstrated. Fixture: one book, live `одуван` = "Одуван", history `{ oduvan: "одуван" }`, and two bak snapshots — `cast.json.bak.castfix` naming `oduvan` = "Одуван", plus `cast.json.bak.2026-01-01` truncated mid-JSON (its real content named `oduvan` a different character, so the true bak evidence for that id is ambiguous):

```
collectBakNameEntries -> {"entries":[{"id":"oduvan","name":"Одуван"}],"bakAvailable":false}
--- main() dry run ---
confirmed repairs: 1 across 1 book(s)
  - [A / S / T] would reinstate "oduvan" (was "одуван", "Одуван")
```

`bakAvailable: false` — lost evidence — and the script reports a **confirmed** repair that `--apply` would write. The corrupt snapshot is silently swallowed to `null` by `readJsonSync` and contributes zero entries, which `buildNameIndex`'s `normSet.size > 1` check then reads as *unambiguous* rather than *unknown*. One unparseable bak is all it takes.

The module doc comment claims the opposite, at `:43-45`:

> No bak evidence (missing, **unreadable**, or ambiguous across snapshots) means this script CANNOT confirm same-name and leaves the pair alone (`reportOnly`, never auto-repaired)

The "unreadable" half is not implemented. Either the doc moves to match the code, or — better, since this is the identity of record — the code gets the gate the doc already promises.

### C3 · High · The same-name confirmation carries no uniqueness rule, so a display name shared by two live rows auto-confirms.

`scripts/repair-a34-wrong-direction-ids.mjs:222`

```js
if (deps.normaliseForMatch(bakEntry.name) !== deps.normaliseForMatch(liveName)) {
```

A bare 1-1 string comparison. It never asks whether the matched name identifies a *unique* character, on either side. Minor cast routinely shares a display name (Солдат, Стражник, Голос, Женщина, and the `unknown-male`/`unknown-female` buckets the sibling script's doc comment already flags as a hazard).

Demonstrated. Fixture: two live rows both named "Солдат" — `soldier-one` (`voiceUuid: VOICE-A`) and `солдат` (`voiceUuid: VOICE-B`) — history `{ soldier: "солдат" }`, bak naming `soldier` = "Солдат":

```
confirmed repairs: 1 across 1 book(s)
  - [A / S / T] would reinstate "soldier" (was "солдат", "Солдат")
```

Nothing on disk says *which* "Солдат" the retired `soldier` was, and the script does not notice that its evidence cannot answer that. `--apply` renames `солдат` → `soldier`, permanently binding every pre-retirement `soldier` attribution and the retirement lineage to a row that may be the wrong one — with a different `voiceUuid`, so the mis-bound character renders in the wrong voice.

This is a divergence from established discipline in the same file family, not a novel judgement call. `resolveTierAName` (`scripts/repair-cast-id-drift.mjs:378-384`) does exactly this lookup and returns `undefined` on a tie, and its doc comment says why: *"the same 'a tie means stop' rule `buildCastResolver` and `remapFreshToPriorIds` both already apply; this function must not reintroduce a looser one."* This detector reintroduces a looser one.

The module doc's defence — that the gate "mirrors `stripEstablishedAsciiRewrites`'s own gate exactly" (`:17`) — does not carry over. There, a wrong answer means failing to strip one rewrite in a fresh analysis run that a later run revisits. Here, a wrong answer is a permanent write to `cast.json` and `cast-id-history.json`. Same comparator, different blast radius; the tie rule is what covers the difference.

### C4 · High · `KNOWN_STANDING_PORTS` makes the apply-time liveness probe permanently blind to this repo's own worktree slot-1 server port.

`scripts/repair-cast-id-drift.mjs:2052` and `:2073`

```js
export const KNOWN_STANDING_PORTS = new Set([8090]);
...
.filter((p) => p <= 65535 && !KNOWN_STANDING_PORTS.has(p));
```

8090 is not only llama-swap's port. It is **this repo's own port for worktree slot 1**, and the repo asserts that in its own test suite:

- `scripts/wt-new.mjs:42-49` — `BASE_PORTS.PORT = 8080`, `PORT_STEP = 10`.
- `scripts/tests/wt-new.test.mjs:83` — `assert.equal(ports.PORT, 8090);`
- `scripts/tests/sidecar-sweep-port.test.mjs:358,373` — the per-worktree `server/.env` fixtures are literally `PORT=8090`.
- `CLAUDE.md`, worktree setup step 5 — "slot 1 is `VITE_PORT=5183`, `PORT`/`VITE_API_PORT=8090`".

8090 also sits inside the default `PORT=8080` auto-rebind walk (8080–8099). `listenWithAutoRebind` (`server/src/crash-logging.ts:136-175`) increments by 1 on `EADDRINUSE` for up to 20 attempts, so a server started with the default `PORT` on a box where 8080–8089 are taken binds 8090 — precisely the #2090 scenario the range probe was widened to catch, per `repair-cast-id-drift.mjs:2000-2007`.

Concrete failure: a Castwright server is running in a slot-1 worktree (or has rebound to 8090). An operator runs `node scripts/repair-a34-wrong-direction-ids.mjs --apply`. `probePortRangeRefused(8080)` skips 8090 entirely, every other port in the range returns `ECONNREFUSED`, `notRefused` is empty, and the script writes `cast.json` and `cast-id-history.json` out-of-process with no in-process lock, underneath a live server writing the same two files. That is exactly the hazard the refusal message at `:360-366` describes.

The comment's own claim — "keeps every script that reuses `probePortRangeRefused` fail-closed for the ports that actually matter to it" — is false for 8090, and its guard rail ("standing and permanent, never a one-off") does not catch this, because 8090 *is* standing and permanent for llama-swap and simultaneously a Castwright port. The scoping needs to key on *what is listening*, not on the port number: probe 8090 and check whether the listener answers as a Castwright server, or take an explicit per-run opt-in (`ALLOW_STANDING_PORTS=8090`), rather than a hardcoded global set that every current and future consumer of this helper inherits forever.

Two aggravating factors. There is **no test** for `KNOWN_STANDING_PORTS` anywhere (`grep -rn KNOWN_STANDING_PORTS scripts/tests/` returns nothing), and this silently changes the apply-time safety of the already-shipped `repair-cast-id-drift.mjs` `main()` too — a different script with the same out-of-process-write hazard, whose behaviour change the PR body does not mention.

### C5 · Medium · The two-file write has no atomicity and no backup; a failure between the two leaves the book worse than unrepaired.

`scripts/repair-a34-wrong-direction-ids.mjs:329-332`

```js
await mods.writeJsonAtomic(bookPlan.castPath, cast);
for (const r of bookPlan.repairs) {
  await mods.retireCharacterId(bookPlan.bookDir, r.nonAsciiId, r.asciiId);
}
```

`cast.json` is written first, the history second, with no try/catch here or at the `main()` call site (`:415-418`). `retireCharacterId` can throw for reasons unrelated to this script: `loadHistoryOrThrow` raises `CastIdHistoryUnreadableError` on a degraded read (`server/src/store/cast-id-history.ts:550`), and it runs inside `withKeyLock`, which bounds acquisition at 10s and throws `LockAcquisitionTimeoutError`. An EPERM from an AV scanner does it too — the sibling script explicitly defends against that shape at `repair-cast-id-drift.mjs:2377-2390`.

The intermediate state is worse than the damage being repaired. After `cast.json` is renamed but before the history write, the live id is `oduvan` and the history still says `{ oduvan: "одуван" }`. In `buildCastResolver` (`server/src/store/cast-resolve.ts:113-118`) a history entry only counts when its target is still live — `одуван` no longer is, so the entry is dropped. `normaliseIdKey` (`server/src/util/character-id.ts:8-10`) is lowercase plus `[-_\s]+ → -` and does not transliterate, so `одуван` and `oduvan` do not collide there either. Net result: `resolve("одуван")` returns `undefined`. Every manuscript attribution and frozen render currently pointing at `одуван` — which is *all of them*, since that is the live id today — becomes an orphan. The operator sees a stack trace and a partially repaired book.

Two things are missing that the codebase already has:

1. **No backup.** `writeJsonAtomic` takes `{ rotate: { keep } }` and rotates `path.bak.N` before writing (`server/src/workspace/state-io.ts:105-132`); this call passes no options, so the pre-repair `cast.json` is gone. The sibling script this one says it mirrors takes an explicit timestamped copy before its write — `backupCastIdHistory` at `repair-cast-id-drift.mjs:2336-2343`. This script writes *two* files instead of one and backs up neither. The `--apply` refusal message tells the operator the write is unlocked; nothing tells them it is also irreversible. The files committed under `docs/testing/onbox-a34-results/backups/` are a manual one-off git commit, not a step the script performs — advisory, not real.
2. **No ordering or rollback argument.** Writing the history first would be strictly safer: the intermediate state would be live `одуван` with history `{ одуван: "oduvan" }`, where `resolve("одуван")` still hits `byId` exactly and only the already-retired `oduvan` dangles. The `:308-309` comment justifies the current order as "matching the order the issue itself lists the two repair actions in", which is not a correctness argument.

The loop also applies repairs one at a time with no rollback, so a book with two confirmed pairs can end half-retired, and a workspace with two planned books can end with book 1 written and book 2 untouched.

---

## Cleanup nits

### N1 · The `ascii-id-already-live` branch has no test.

`scripts/repair-a34-wrong-direction-ids.mjs:211-214`. No test produces `reason: 'ascii-id-already-live'`. Replacing that branch with `repairs.push({ asciiId, nonAsciiId, name: liveChar.name }); continue;` leaves **27/27 green** — and that mutation writes a `cast.json` with two rows sharing one `id`, which `buildCastResolver`'s `byId` silently resolves to whichever came first.

### N2 · Release-notes pair not updated, with no explicit not-applicable.

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the PR body does not say the step is skipped or why. Before-shipping step 5 permits skipping for a change with no shippable delta but requires saying so explicitly. A new operator-facing repair script plus a behaviour change to an existing one is at minimum an operator-visible delta worth a technical-register line.

### N3 · Three new on-box evidence docs are unreachable from the register.

`docs/testing/onbox-a34-results/step-{1,2,3}*.md` record a real dry run against the real `C:\AudiobookWorkspace`, but neither register row A34 (`docs/testing/onbox-acceptance-register.md:3403`) nor its cited run sheet (`cast-id-drift-onbox-acceptance.md` §10) is touched, so nothing in the register points at them. Someone working the A34 row will not find this evidence.

### N4 · The PR body does not declare the `repair-cast-id-drift.mjs` change.

The Summary describes only the new script. The `KNOWN_STANDING_PORTS` commit changes the apply-time safety of an already-shipped sibling script for every caller, and reads as undeclared scope creep. The test count is also stale — the body says 25/25; the suite is now 27.

---

## Checked and clean

Stated because the dispatch asked for these specifically.

- **The escaped regex rewrite is behaviour-identical.** The escaped class and the literal class were compared character by character over the **entire BMP** (0x0000–0xFFFF): **zero** membership differences. The class is the 26 `\s` code points plus U+0022, U+0027, U+0060. The paired tests are real, not decorative — dropping U+0060 from the class reddens 1 test, adding `-` reddens 2. `scripts/repair-a34-wrong-direction-ids.mjs:161`.
- **The merge-induced-failure explanation is accurate.** `regexLiteralDesyncsQuoteTracking` (`server/src/spawn-windows-hide.test.ts:117-132`) tracks one open quote at a time and treats the double quote, apostrophe and backtick alike, so the literal class — one of each, all odd counts — leaves `openQuote` non-null and throws at import time. The remediation matches the five sites main already fixed.
- **The merge really changed nothing on either side.** `git diff ddf1441c c83c7c1c -- server package.json package-lock.json .github` is empty. Claim holds.
- **The retirement direction is right.** `retireCharacterId(bookDir, nonAsciiId, asciiId)` satisfies `history.supersededBy[to] === from` and takes the direct-reversal branch (`cast-id-history.ts:565-579`): deletes the stale `asciiId → nonAsciiId`, repoints anything targeting `nonAsciiId` onto `asciiId`, writes `nonAsciiId → asciiId`. Correct, and correctly delegated rather than hand-rolled.
- **The dry run genuinely writes nothing.** The end-to-end fixture test drives real `collectBooks` plus the real detector and byte-compares both files afterwards; the liveness-refusal test is real (neutering the refusal reddens it) and fires before `collectBooks` and before any `server/dist` import.
- **`state.json` needs no rewrite.** The real book's committed `state.json` backup contains neither `oduvan` nor `одуван`; leaving it alone is correct.
- **`normaliseForMatch` cannot be made to fold two names apart or together incorrectly** beyond the tie problem in C3 — the fold set (case, smart quotes, dashes, ellipsis, edge trim, whitespace collapse) matches `server/src/util/text-match.ts:19-28` exactly.

---

*Findings only — nothing applied. `git rev-parse HEAD` and `git status --porcelain` unchanged across this pass.*

dudarenok-maker:
## PR review — pass 2 (head 3d85bef3, depth medium)

Re-review at `3d85bef3`, worktree `C:\Claude\Projects\wt-2903-a34-repair`, tree clean and HEAD unmoved before and after. No tracked file modified. Every mutation and every demo below ran against a **copy** of `scripts/` in `%TEMP%`; the apply path was never pointed at a real workspace.

Deltas only. Baseline is pass 1 (head `c83c7c1c`).

**2 new correctness bugs, 3 new cleanup nits. All nine pass-1 findings resolved.**

---

## Pass-1 dispositions

### C1 · apply path untested / test re-implemented it — **VERIFIED RESOLVED**

`applyBookPlan` and `backupBeforeApply` are exported (`repair-a34-wrong-direction-ids.mjs:445`, `:420`), imported at `scripts/tests/repair-a34-wrong-direction-ids.test.mjs:38-39`, and the inline copy is gone. Verified by re-running the pass-1 mutation battery plus eleven new tokens against the temp copy — baseline **43/43 green**:

| Mutation | Pass 1 | Pass 2 |
|---|---|---|
| `:478` → `retireCharacterId(bookDir, r.asciiId, r.nonAsciiId)` (args swapped) | 27/27 pass | **CAUGHT** (1 fail) |
| `applyBookPlan` → `return { backups: [] };` stub | 27/27 pass | **CAUGHT** (6 fail) |
| `:461` partial-repair refusal → `if (false && …)` | 27/27 pass | **CAUGHT** (1 fail) |
| `:297` `ascii-id-already-live` → push a repair (N1) | 27/27 pass | **CAUGHT** (1 fail) |

All four pass-1 blind spots now redden a named test. The argument-swap one — the mutation that re-inflicts A34 on an already-rewritten book — is pinned by an explicit `deepEqual` on the recorded call at `test.mjs:571-575`, not by a side effect.

### C2 · `bakAvailable` discarded, evidence guard failed open — **VERIFIED RESOLVED**

Threaded at `:541`/`:549`, gated at `:305-308`, and driven end-to-end through `main()` by a genuinely truncated `cast.json.bak.*` on disk (`test.mjs:711-741`). Verified by mutation, not by reading:

| Mutation | Result |
|---|---|
| `:305` → `if (false && bakAvailable === false)` | **CAUGHT** (2 fail) |
| `:541` → `const { entries } = …; const bakAvailable = true;` (re-discard the field at the wiring) | **CAUGHT** (1 fail) |

`undefined` is deliberately not read as `false` (`:305` is `=== false`, not falsy), which keeps the pure-helper tests that pass no flag on `no-name-evidence` rather than on the new reason — correct, and the doc says so.

### C3 · no uniqueness rule on the same-name confirmation — **VERIFIED RESOLVED**

Live-side census at `:270-276`, tie rule at `:324-327`. Mutating `:324` to `if (false && …)` reddens 2 tests. I also replayed the **shipped** detector against this branch's own committed copy of the one real hit (`docs/testing/onbox-a34-results/backups/castwright__standalones__заказ-коалфолла/`): 14 live rows, **zero** normalised-name collisions, `одуван` count = 1 — so the tightened rule does not make the fix unreachable for the case it was built for (`bakAvailable: true` → still confirmed; `false` → `bak-evidence-unreadable`, as designed).

### C4 · `KNOWN_STANDING_PORTS` — **VERIFIED RESOLVED**, on all three points the dispatch asked about

- **(a) default is fail-closed.** `KNOWN_STANDING_PORTS` is gone repo-wide (`grep` finds only prose). `parseStandingPorts(undefined)` → empty set. Mutating the default back (`String(raw ?? '8090')`) **CAUGHT** (2 fail); dropping the range validation **CAUGHT** (1 fail).
- **(b) the refusal is reachable.** `test.mjs:456-508` drives `main(['--apply'])` against a real `net` listener and asserts exit 1 with both files byte-identical; neutering the refusal (`:518` → `if (false && …)`) reddens it. I reproduced the refusal by hand too. *Clearly worded* — see 🟡 N1: it never names the escape hatch.
- **(c) nothing re-opens the hole.** `probePortRangeRefused` is the sole consumer and reads the env **per call** (`repair-cast-id-drift.mjs:2116`), so no module-load freeze; making it ignore the env entirely is **CAUGHT** (1 fail). Both `main()`s route through that one function. The only residue is an audit-trail gap, 🟡 N2.

### C5 · no backup, no ordering argument — **RESOLVED, with a new hole in the fix**

Order is now history-first (`:477-480`) with the two intermediate states argued at `:76-97`; both files are copied before either write (`:474`); a part-way failure names the copies (`:485-489`). Mutations: reversing the order **CAUGHT** (2 fail); `const backups = []` **CAUGHT** (2 fail); backing the cast file up twice so the history is never copied **CAUGHT** (2 fail); dropping the paths from the error **CAUGHT** (1 fail).

The backup itself is not durable across a same-day retry — see 🟠 D2.

### N1–N4 · **all VERIFIED RESOLVED**

`ascii-id-already-live` covered (mutation reddens); both release-notes files carry an entry (`RELEASE_NOTES.md` + `docs/release-notes-next.md`, two bullets — the script and the probe change); the three evidence docs are now reachable from register row A34 via the run sheet it cites (`docs/testing/cast-id-drift-onbox-acceptance.md` §10.2b); the PR body declares the `repair-cast-id-drift.mjs` change in its own section and the test count is current (43).

### Merge resolution — **VERIFIED HONEST**

`git diff 0a9bfe16 3d85bef3 -- RELEASE_NOTES.md` shows exactly one addition (main's fs-60 bullet) and zero deletions. `git diff 8e4025ee 3d85bef3 -- RELEASE_NOTES.md` shows exactly one addition (this branch's A34 bullet) and zero deletions. Neither side lost content. The rest of the merge is clean: `git diff origin/main...HEAD` is the 12 branch files only.

---

## 🔴 New — correctness bug

### D1 · Critical · `collectBooks`'s `droppedBooks` is discarded, so a book whose `cast.json` is present but corrupt is reported as a clean workspace — the same fail-open shape C2 just closed, one collector over

`scripts/repair-a34-wrong-direction-ids.mjs:531-532` and `:561-563`

```js
const { books } = collectBooks(workspaceDir);
console.log(`books scanned: ${books.length}\n`);
```

`collectBooks` returns `{ books, droppedBooks }`. A book whose `cast.json` or `state.json` is **present but unreadable or wrong-shaped** is pushed to `droppedBooks` with `reason: 'unreadable'` — evidence LOST, explicitly distinguished from the legitimate `'not-yet-analysed'` absence (`repair-cast-id-drift.mjs:1723-1731`). So is an author/series directory that cannot be enumerated (`:1686`, `:1692`, `:1698`). This script destructures the field away, exactly as it used to destructure `bakAvailable` away.

Demonstrated, `--apply`, ports pointed at free sockets so the liveness gate passes:

```
=== CASE B: 1 book, cast.json present but corrupt, --apply ===
mode: APPLY (writing cast.json + cast-id-history.json)
books scanned: 0

0 confirmed wrong-direction pairs — nothing to repair.
exitCode = 0
```

The fixture is one book with `cast.json` truncated mid-JSON, a valid `state.json`, and a `cast-id-history.json` carrying `{"oduvan": "x-nonascii"}` — i.e. **the one book in the workspace that has the exact defect this script exists to find**. It is invisible, the count is `0`, and the operator is told the workspace is clean. Case A (a workspace root with no `books/` at all — a forgotten `BASE`/`WORKSPACE_DIR`, which defaults to `<home>/AudiobookWorkspace`) prints the same two lines and exits 0.

The sibling this script says it mirrors handles both, and this repo already paid for both lessons:

- **#2097** — `repair-cast-id-drift.mjs:2519-2531` names every dropped book and why, on dry runs too, and `:2540-2550` **refuses `--apply` unconditionally** on any `'unreadable'` drop, because "this pass cannot scan them at all, so it cannot rule out damage sitting unprotected in them."
- **#2108** — `:2512` routes the count through `formatBooksScannedLine()` precisely so a zero-book scan reads as "nothing was examined" rather than as an unremarkable count, and `:2555-2560`'s own comment says why: *"a wrong `WORKSPACE_DIR` … scans ZERO books — and without this check, that reads as 'clean' rather than 'unknown'."*

Why this is not cosmetic: `docs/testing/onbox-a34-results/step-3-dry-run.md` is committed evidence built from this exact output, its Result section leads with "**23 books scanned**, matching step 1's manual count", and register row A34's discharge will rest on it. A dropped book silently moves that denominator with no line anywhere saying so — and step 3's own text already notes the real `C:\AudiobookWorkspace` contains a mojibake-named directory and a throwaway directory, i.e. the tree is exactly the shape where a `readdir`/parse failure is plausible rather than theoretical.

Minimum fix is the reporting half (name the dropped books and their reason, and make a zero-book scan read as "nothing examined"). Whether `--apply` should also refuse is the sibling's own precedent rather than a judgement call this PR gets to skip silently: a book that could not be read is a book this pass cannot rule A34 damage out of.

---

## 🟠 New — correctness bug

### D2 · Medium · The backup C5 added is destroyed by the retry the error message invites — same-day, same path, silently overwritten with post-repair content

`scripts/repair-a34-wrong-direction-ids.mjs:420-426`, used at `:474`

```js
const stamp = new Date().toISOString().slice(0, 10);
const backupPath = `${filePath}.bak.a34-${stamp}`;
deps.fs.copyFileSync(filePath, backupPath);
```

The stamp is date-only and `copyFileSync` overwrites without `COPYFILE_EXCL`, so every run on the same day writes the same two paths. That is fine for a first run and wrong for the exact scenario the backup was added for.

Demonstrated. Book with two confirmed pairs; `retireCharacterId` succeeds for pair 1 and throws `LockAcquisitionTimeoutError` on pair 2 (one of the three failure modes the module doc at `:78-80` names):

```
RUN 1 (--apply):
  threw: … repair failed part-way (LockAcquisitionTimeoutError…). Pre-repair copies: …
  backup after run 1 : {"schema":1,"supersededBy":{"oduvan":"одуван","soldier":"солдат"}}
  live history now   : {"schema":1,"supersededBy":{"soldier":"солдат","одуван":"oduvan"}}

Operator retries the transient failure the SAME DAY.
RUN 2 (--apply, replanned: only the soldier pair is still wrong-direction):
  threw: … Pre-repair copies: …
  backup after run 2 : {"schema":1,"supersededBy":{"soldier":"солдат","одуван":"oduvan"}}

  pre-repair state was: {"schema":1,"supersededBy":{"oduvan":"одуван","soldier":"солдат"}}
  backup still holds it? false
```

Retrying a transient lock timeout is the obvious operator response, and re-planning legitimately narrows the plan (pair 1's `supersededBy` entry was rewritten by run 1, so it no longer matches the wrong-direction shape). Run 2 then copies the **half-repaired** history over the only pre-repair copy. After that, `cast-id-history.json.bak.a34-<date>` restores the half-applied state, not the original — and run 2's own error message still points the operator at it as "Pre-repair copies". The undo silently became a no-op while continuing to advertise itself.

`RELEASE_NOTES.md`'s user-facing line for this PR — "*it now takes a dated copy of both files it changes before changing either — so there is always something to go back to*" — is the claim that does not hold.

`repair-cast-id-drift.mjs:2378-2384`'s `backupCastIdHistory` has the same date-only stamp, so this is an inherited shape rather than an invented one — but that script backs up **one** file and does not carry this PR's "there is always something to go back to" promise, and the class-fix rule points at both. Refusing when the target exists (or stamping to millisecond precision, or `.bak.a34-<date>.N`) all close it. Nothing in the suite covers a second run: `test.mjs:610-632` asserts the two backups only on a fresh fixture.

---

## 🟡 New — cleanup nits

### N1 · Neither `--apply` refusal names `ALLOW_STANDING_PORTS`, so the operator the opt-in exists for cannot find it

`scripts/repair-a34-wrong-direction-ids.mjs:519-525`, `scripts/repair-cast-id-drift.mjs:2474-2480`

Both messages end "Stop the server on `8080-8099` (and LAN HTTPS …), or point `PORT`/`LAN_HTTPS_PORT` elsewhere." The whole reason the hardcoded set became an opt-in rather than a deletion is that llama-swap really does hold 8090 on the operator's box and a bare deletion "would refuse every `--apply` there forever" (commit `0a9bfe16`). That operator now hits a refusal whose only suggested remedies are stopping a service that is not Castwright, or moving a port. The escape hatch is documented in two module doc comments and in neither refusal.

### N2 · The probe's log line claims a range it did not fully probe when the opt-in is set, and nothing records the skip

`scripts/repair-a34-wrong-direction-ids.mjs:509-512`

```
(a live listener really is on 58916, inside the probed range)
probing 127.0.0.1:58913-58932 and 127.0.0.1:58913-58932 for a live server...
books scanned: 1
```

`ALLOW_STANDING_PORTS=58916` was set; 58916 was skipped; the line still claims the whole range and no other line mentions the exception. These transcripts are pasted verbatim into `docs/testing/onbox-a34-results/*.md` as acceptance evidence, so a run that skipped a port is indistinguishable from one that probed it and got `ECONNREFUSED`. One line — `skipping port(s) N per ALLOW_STANDING_PORTS` — makes the assertion auditable, which is what "a visible act rather than an inherited constant" (`repair-cast-id-drift.mjs:2073-2077`) is supposed to mean.

### N3 · `step-3-dry-run.md` records a dry run of the pre-review script and is not annotated as such

`docs/testing/onbox-a34-results/step-3-dry-run.md`

`step-2-implement.md` gained an honest "Pass-1 review update" section saying everything above it describes the first draft. `step-3-dry-run.md` — the real-workspace run whose output is step 4's precondition — got no such note, and its pasted output was produced by a detector that has since gained two gates and two report-only reasons (three reasons became five). I replayed the shipped detector against this branch's own committed copy of the confirmed book and it still confirms, and `Playing with Fire` has no bak files at all so `bakAvailable` stays `true` and its two entries still read `no-name-evidence` — i.e. the recorded output does appear to survive the change. That is worth one sentence in the file, because right now nothing in it says the question was asked.

---

## Verdict

**Blocking.** D1 is a correctness bug: a workspace containing the exact damage this script exists to find is reported as clean, on the summary line that becomes committed acceptance evidence for register row A34. It is the same class as C2, which this round fixed — the second return field of a shared collector destructured away — and the sibling script already closed both halves of it under #2097 and #2108.

D2 is a correctness bug in the safety property C5 added, reachable by the retry its own error message invites.

Everything from pass 1 is genuinely fixed and genuinely pinned: fifteen mutations, including all four that survived pass 1, now redden a named test, and the control still behaves.

*Findings only — nothing applied. `git rev-parse HEAD` and `git status --porcelain` unchanged across this pass.*

dudarenok-maker:
## PR review — pass 3 (head c4615ecb, depth medium)

Final pass of the 3-pass cap. Worktree `C:\Claude\Projects\wt-2903-a34-repair`, tree clean and HEAD unmoved before and after. No tracked file modified. Every mutation ran against a **copy** of `scripts/` in `%TEMP%` (verified byte-identical to the worktree after each restore); `--apply` was never pointed at a real workspace — the one real-workspace run below is a **dry run**, which returns before `loadServerModules()` and writes nothing.

Deltas only. Baseline is pass 2 (head `3d85bef3`).

**1 new correctness bug (🔴), 1 new correctness bug (🟠), 1 new cleanup nit. Both pass-2 blocking findings verified resolved.**

---

## Pass-2 dispositions

### D1 · `droppedBooks` discarded — **VERIFIED RESOLVED**

Verified by mutation against the temp copy, baseline **48/48 green**, not by reading the fix commit:

| Mutation to `repair-a34-wrong-direction-ids.mjs` | Result |
|---|---|
| `:570` → `const { books } = …; const droppedBooks = [];` (re-discard the field) | **CAUGHT** (2 fail) |
| `:602` → `if (false && shouldRefuseApplyForUnreadableBooks(…))` | **CAUGHT** (1 fail) |
| `:576` → `` console.log(`books scanned: ${books.length}\n`) `` (drop the #2108 line) | **CAUGHT** (1 fail) |

Each reddens a *named* test, and the three tests are distinct.

**The helper reuse the dispatch asked me to check specifically — two of three correct, one partly wrong.**

- `shouldRefuseApplyForUnreadableBooks` — **cannot fire on a healthy workspace, and does fire on the shape it claims.** `collectBooks`'s `dirs()` filters `readdirSync(..., { withFileTypes: true })` to directories only (`repair-cast-id-drift.mjs:1674-1683`), so a stray file at any level never becomes a book, and a title directory with no `.audiobook` yields two `ENOENT`s → `'not-yet-analysed'`, never `'unreadable'`. Confirmed empirically with a **dry run against the real `C:\AudiobookWorkspace`**: 27 books scanned, 3 `not-yet-analysed`, **0 dropped-unreadable**, same 1 confirmed pair and same 2 report-only pairs as the committed step-3 evidence. No false refusal.
- `formatNotYetAnalysedLine` — correct here. Its text promises "a file that's PRESENT but unreadable is counted separately **below**", and the `books DROPPED` block genuinely is printed below it (`:594-600`). The cross-reference holds in this script, not just in the sibling.
- `formatBooksScannedLine` — see 🟠 E2. The line's own text is written for a summary block this script does not have.

### D2 · backup destroyed by a same-day retry — **VERIFIED RESOLVED**

Attacked directly, on each of the four axes named in the dispatch:

| Mutation | Result |
|---|---|
| `backupBeforeApply` reverted to date-only stamp **and** no `COPYFILE_EXCL` (the original bug) | **CAUGHT** (1 fail — `a same-day retry does not destroy the first run's pre-repair copy`) |
| `:454` → `throw err;` (remove the EEXIST retry) | **CAUGHT** (1 fail — `an exact stamp collision retries to a distinct path…`) |
| `backupCastIdHistory` reverted to date-only + no EXCL | **CAUGHT** (1 fail, in `repair-cast-id-drift.test.mjs`, baseline 190/190) |

Reverting **either half alone** survives (48/48, 190/190). That is correct rather than a gap — the two mechanisms are deliberately redundant and either one alone preserves the property — but it means the suite pins the *property*, not `COPYFILE_EXCL` itself. Worth knowing before someone "simplifies" the stamp later; not a finding.

- **Non-`EEXIST` error mid-loop**: rethrown unchanged (`:454`), same as the pre-fix `copyFileSync`. No new swallow.
- **A `deps.fs` that ignores the mode flag**: only reachable from tests — production callers take the `{ fs }` default. The collision test's fake does ignore `flags` and throws unconditionally, so it proves the *retry*, not that EXCL is passed; that is covered by the full-revert mutation above.
- **`collectBakNameEntries`' glob**: unaffected. It filters `f.startsWith('cast.json.bak')` (`repair-cast-id-drift.mjs:1813`), which matches `cast.json.bak.a34-<stamp>` in both stamp shapes and never matched `cast-id-history.json.bak.id-drift-*` in either. The doc comment now states this.
- **Can the 1000-attempt throw strand a caller mid-repair?** No. In `applyBookPlan` both backups are taken at `:513` *before* the `try` that does either write, so a throw there means zero writes. In the sibling `main()` the call sits above its own `try` too (`:2937`), and a copy failure could already abort that per-book loop before this change (EPERM, ENOSPC) — the exhaustion path adds no new reachable shape.

### Collateral from `c4615ecb` (the highest-risk commit) — **CHECKED, none found**

- The only importer of `repair-cast-id-drift.mjs` outside `scripts/` is `server/src/store/cast-resolve.repair-pass-contract.test.ts:8`, which imports two named symbols. Adding an export cannot affect it. No test anywhere asserts the module's export surface.
- The new `deps = { fs }` parameter matches the existing `backupBeforeApply` seam exactly, and the one production call site (`:2937`) passes a single argument, taking the default.
- The backup-name shape change reaches operators only through the returned path — every message interpolates `backupPath` rather than rebuilding it (`:2938`, `:2955`).
- Both suites re-run at this head in the temp copy: **48/48** and **190/190**. CI's `Hooks tests` step ran (not skipped) and succeeded on run `34202660564`, and `scripts/run-hooks-tests.mjs` globs `scripts/tests/*.test.mjs`, so the new file is genuinely picked up.

### The five new tests — **INDEPENDENTLY MUTATION-VERIFIED**

All five reddened under a targeted revert, each naming a different test: `droppedBooks` re-discarded (2), refusal neutered (1), `formatBooksScannedLine` reverted (1), backup fix fully reverted (1), EEXIST retry removed (1). Plus the sibling's under a full revert of `backupCastIdHistory` (1). Reported mutation-proof is real here.

### Pass-2 cleanup nits N1, N2 — **STILL OPEN**, untouched by the three commits

Neither `--apply` refusal names `ALLOW_STANDING_PORTS` (`repair-a34-wrong-direction-ids.mjs:558-565`), and the probe's log line still claims the whole range when a port was skipped (`:548-551`).

### Pass-2 nit N3 — **STILL OPEN, and materially worse than when it was filed**

`docs/testing/onbox-a34-results/step-3-dry-run.md` still records a transcript that no longer reproduces on this box. Its Result section leads with "**23 books scanned**, matching step 1's manual count", and `cast-id-drift-onbox-acceptance.md` §10.2b repeats the 23. My dry run at this head, same machine, same real workspace, prints:

```
books scanned: 27

books not yet analysed (cast.json and/or state.json genuinely missing, …): 3
  - ______ __________ / The Night Watch Tetralogy (C1 cloud throwaway) / ______ _____ (C1 cloud throwaway)
  - Сергей Лукьяненко / The Night Watch Tetralogy / Ночной дозор (C1 cloud throwaway 2894)
  - Сергей Лукьяненко / The Night Watch Tetralogy (C1 cloud throwaway 20260906) / Ночной дозор (C1 cloud throwaway 20260906)
```

The *substantive* findings are unchanged (1 confirmed pair, same book and ids; the same 2 `no-name-evidence` report-only entries), which is the important part and is worth recording. But step 4 on the box begins with a re-run, and its operator will meet 27 where the run sheet says 23 and a block the run sheet does not mention. One sentence in each file settles it.

---

## 🔴 New — correctness bug

### E1 · `cast-id-history.json` present-but-unreadable is read as "this book has no history", so the book carrying the exact damage this script exists to find reports clean — the **third** site of the class, on the primary evidence file

`scripts/repair-a34-wrong-direction-ids.mjs:616` and `:627`, helper at `:370-376`

```js
const history = readJsonSync(historyPath);
…
supersededBy: history?.supersededBy ?? {},
```

```js
const readJsonSync = (p) => {
  try { return JSON.parse(fs.readFileSync(p, 'utf8')); } catch { return null; }
};
```

`readJsonSync` collapses ENOENT, EACCES, ENOTDIR and a parse failure into one `null`. `supersededBy` then becomes `{}` — indistinguishable from the legitimate, common "book has no history file" case the doc at `:222` describes. `supersededBy` is not merely one input to the detector; it is the **only** thing `planBookRepairs` iterates (`:282`). Zero entries means zero candidate pairs, so the book contributes nothing to `repairs` *and* nothing to `reportOnly`, and — because `collectBooks` only shape-checks `cast.json` and `state.json` — it is still counted in `books scanned`.

Demonstrated. Two fixtures identical except for one file, driven through the real `main()`:

```
================ CONTROL — cast-id-history.json intact ================
books scanned: 1

confirmed repairs: 1 across 1 book(s)
  - [Author / Series / Title] would reinstate "oduvan" (was "одуван", "Одуван")

================ PROBE — cast-id-history.json PRESENT but truncated mid-JSON ================
books scanned: 1

0 confirmed wrong-direction pairs — nothing to repair.
```

And under `--apply`, with the liveness probe pointed at verified-free ranges so the gate passes:

```
mode: APPLY (writing cast.json + cast-id-history.json)
books scanned: 1

0 confirmed wrong-direction pairs — nothing to repair.
exitCode = 0
```

No `books DROPPED` line, no report-only line, no refusal, exit 0. The operator is told the workspace is clean by a run that never read the file the answer lives in.

This is not a new class — it is the same one this round closed twice. Pass 1's C2 was `bakAvailable` destructured away; pass 2's D1 was `droppedBooks` destructured away; both were "evidence LOST read as evidence ABSENT". The mechanism differs here (a swallowing reader rather than a discarded field) but the failure is identical, and the repo already built the instrument for it: `readJsonTriState` (`repair-cast-id-drift.mjs:1572-1584`) exists precisely because "`existsSync` swallows EVERY error to a bare `false` — indistinguishable from a genuinely absent file", which is #2097 defect 10, verbatim. `collectBooks` uses it for `cast.json`/`state.json`; this script re-opens the same blind spot one file over, on the file that matters most.

It also makes D1's fix under-deliver on its own promise. `--apply` now refuses when `cast.json` is unreadable *because* "this pass cannot scan them for wrong-direction characterId retirements at all" — but a book whose `cast.json` is pristine and whose `cast-id-history.json` is corrupt cannot be scanned for wrong-direction retirements either, and gets no refusal, no line, and a place in the scanned count.

Two knock-on documentation defects fall out of the same read: `:222`'s "`{}` if the book has no history file at all — the common case" and `planWorkspaceRepairs`'s "whether it has no history file at all, or a history file with no wrong-direction entry, or only report-only pairs" (`:341-346`) both enumerate the reachable causes of an empty result and both omit "or a history file that could not be read".

**Correctness bug**, not a nit. The fix is the shape the sibling already ships: read it tri-state, name the book, and route an `'unreadable'` history into the same `--apply` refusal `unreadableBooks` already drives.

---

## 🟠 New — correctness bug

### E2 · `shouldRefuseApplyForEmptyScan` was left behind: `--apply` against a mistyped `WORKSPACE_DIR` prints the borrowed warning and exits **0**, where the sibling refuses

`scripts/repair-a34-wrong-direction-ids.mjs:570-611` (import list `:152-158`)

D1's fix imported three of the four helpers the sibling's #2097/#2108 handling is built from — `formatBooksScannedLine`, `formatNotYetAnalysedLine`, `shouldRefuseApplyForUnreadableBooks` — and left the fourth. `formatBooksScannedLine`'s own doc comment names it as the other half of the mechanism: *"Dry-run only in practice: `--apply` refuses before ever reaching the summary when `booksScanned === 0` (see `shouldRefuseApplyForEmptyScan`)"* (`repair-cast-id-drift.mjs:1493-1497`). In this script that sentence is false — the line is reachable under `--apply` and nothing refuses.

Demonstrated, `--apply` against a directory with no `books/` at all (the shape a forgotten or mistyped `BASE`/`WORKSPACE_DIR` produces — it defaults to `<home>/AudiobookWorkspace`):

```
mode: APPLY (writing cast.json + cast-id-history.json)
probing 127.0.0.1:45824-45843 and 127.0.0.1:44974-44993 for a live server...
books scanned: 0 — WARNING: nothing was examined; every count below is a row of clean zeros because NOTHING WAS SCANNED, not because the workspace is healthy. Check WORKSPACE_DIR before trusting anything else here.

0 confirmed wrong-direction pairs — nothing to repair.

exitCode = 0
the sibling's own decision helper on the same input: shouldRefuseApplyForEmptyScan(true, 0) = true
```

The exit code is the part that matters. `shouldRefuseApplyForEmptyScan`'s doc comment says outright *"There is no legitimate `--apply` against an empty workspace"* (`repair-cast-id-drift.mjs:1478-1485`), and this script answers `0` — success — to exactly that invocation. These transcripts are pasted into `docs/testing/onbox-a34-results/*.md` as acceptance evidence, so a run that examined nothing exits indistinguishably from one that examined 27 books and found them clean. Nothing is written, so the blast radius is a wrong assertion rather than wrong data — that is why this is 🟠 and not 🔴.

Second, smaller half of the same finding, and the "helper wording may be subtly wrong here" case the dispatch asked about: the imported warning says *"every count below is a row of clean zeros"*. That describes the sibling's `--- Summary ---` block. This script prints one sentence below it, and when `books/` itself is unreadable the count below is `1`, not a zero. The borrowed sentence is close enough to be useful and not quite true here.

---

## 🟡 New — cleanup nit

### E3 · Four documents still describe the backup shape and the review history this round replaced

Under CLAUDE.md these are chores the work made owed, not taste — a doc sentence a change made false.

- `docs/testing/cast-id-drift-onbox-acceptance.md:978` — "the script copies both files it touches to `` `.bak.a34-<date>` ``". It is now `.bak.a34-2026-09-08T14-22-05-311Z`. This is the run sheet an operator follows for step 4, and it is where they will look to confirm the backups exist.
- `docs/testing/onbox-a34-results/step-2-implement.md:151` — same `` `<file>.bak.a34-<date>` `` claim.
- `docs/release-notes-next.md:184` — "both files copied to `` `.bak.a34-<date>` `` first", and "One hit across the **23 books** scanned". More substantively: this entry enumerates all five of pass 1's findings in detail and records **none** of pass 2's, and **#3095 — which this PR closes, in an already-shipped operator script — has no line in either release-notes file at all.** The entry's own precedent sets the bar it now fails.
- `RELEASE_NOTES.md:3` — "it now takes a **dated** copy of both files it changes". The substantive claim ("there is always something to go back to") is repaired and now true; "dated" is the residue. The operator-visible half that is genuinely missing is D1's new behaviour: `--apply` now *refuses outright* when a book cannot be read.

Also, the PR body's own **Test plan** still reads "**43/43**" and "187/187" while a later paragraph in the same body says "**48 pass** (a34) and **190 pass** (drift)". Pass 1's N4 flagged a stale test count in this body; it recurred.

---

## Verdict

**Not blocking, with one finding I would want fixed before this ships and would accept as a follow-on if the owner disagrees.** Passing this to the repo owner rather than opening a round 4, per the cap.

- **E1 (🔴) — I would not merge without it, but I would not hold the PR open for it either.** It is a real fail-open on the file the whole detector keys on, and it lands in the one place this PR's evidence is weakest: the summary line that becomes register row A34's discharge. It writes nothing wrong; it under-reports. Set against that, this is the third round in a row to find the same class, the fix is small and has an in-repo template (`readJsonTriState` + the `unreadableBooks` path that already exists ten lines above), and a fourth site is exactly what the next round would find. **My recommendation: fix it here.** If the owner ships without it, the honest statement in step 4's run sheet is that a book with an unreadable `cast-id-history.json` is invisible to this pass, and the dry run's `books scanned` is a count of books *opened*, not of books *examined*.
- **E2 (🟠) — I would merge without this.** It cannot corrupt data; it produces a wrong exit code on an operator error the warning line already shouts about. It is a three-line import-and-call, and it belongs in the same commit as E1 if E1 is taken.
- **E3 (🟡) — I would merge without this.** Documentation drift, no runtime effect. The one piece I would not leave is the missing release-notes line for #3095: this PR closes a bug in an already-shipped script and neither notes file mentions it, which is what Before-shipping step 5 exists to prevent.
- **Pass-2 N1/N2/N3 remain open.** N3 is the one with real cost — an operator starting step 4 meets 27 books where the run sheet promises 23.

Everything pass 2 blocked on is genuinely fixed and genuinely pinned: three mutations for D1, three for D2 across both scripts, each reddening a distinct named test, and the real-workspace dry run confirms the new refusal does not fire on a healthy box. The `c4615ecb` incidental carries no collateral I can find — no export-shape assertion, no changed default, no behaviour reaching an unreviewed caller.

*Findings only — nothing applied. `git rev-parse HEAD` and `git status --porcelain` unchanged across this pass.*

Issue #2155, opened 2026-08-05
srv-87 — the cast-reuse carryover file has two unlocked touchers the #2099 cast lock doesn't reach


dudarenok-maker:
## What

#2099 put `applyReparse`'s cast.json read, its derived carryover write, and the cast.json delete inside one `withCastLock` hold. That makes the span atomic **against other lock-taking writers only**. `cast-reuse-carryover.json` itself has two touchers that take no lock at all, so they are unaffected:

- **`server/src/routes/analysis.ts:186`** — `readPriorCastForMerge` **reads** the carryover, unlocked, and *prefers `cast.json` when present*:
  ```ts
  const fromCast = (await readJson(castJsonPath(bookDir)).catch(() => null))?.characters;
  if (fromCast?.length) return fromCast;
  const fromCarryover = (await readJson(castReuseCarryoverJsonPath(bookDir)).catch(() => null))?.characters;
  ```
  Because it prefers `cast.json`, a concurrent analysis run that reads while reparse holds the lock can still observe the intermediate state — carryover written, `cast.json` not yet deleted — and take the roster reparse is about to discard, exactly as before #2099.

- **`server/src/routes/analysis.ts:2928`** — an unlocked `rm(castReuseCarryoverJsonPath(...), { force: true })`. Nothing serialises it against reparse's carryover write, so a carryover snapshot can be deleted between being written and being read.

Neither is closed by #2099, and **neither is covered by #2015** — that issue is scoped to `analysis.ts`'s five cast.json *merge-base writes* and does not mention the carryover file. Filed so the residual is tracked rather than living only as prose in the plan doc and PR #2148.

## Why it was not fixed in #2099

Out of that ticket's scope, and not a mechanical wrap: `analysis.ts` is the file the cast-lock static guard allowlists (its five writes are deferred to #2015), so touching its locking is entangled with that ticket. It also needs a decision this issue does not pre-empt — whether the carryover deserves its own lock class, whether it should ride the existing `cast` key (it is per-book and written under the cast lock today, so that is plausible), or whether `readPriorCastForMerge`'s cast-then-carryover preference should be inverted or made atomic.

## Acceptance

- The carryover's read and delete in `analysis.ts` either participate in the same per-book serialisation as its writer, or the divergence is documented at each site with the reason.
[...]

PR #3060, opened 2026-09-06
test(server): retarget the fresh-cast-lock test at the invariant it can reach

### LINKED ISSUE #3022: analysis.fresh-cast-lock.test.ts cannot detect the lock removal it exists to catch
## What

`server/src/routes/analysis.fresh-cast-lock.test.ts` exists to prove one thing — that the `withCastLock` wrapper around the "Start fresh" `cast.json` delete (`server/src/routes/analysis.ts:3660`) is what stops a concurrent `add-alias` writer resurrecting the roster from a stale read (#1981 Task 11).

**It cannot detect the removal of that lock.** Verified twice, independently:

| Who | Mutation | Result |
|---|---|---|
| PR #3009 fix agent | removed the `withCastLock` wrapper at `analysis.ts:3660` | `Tests 1 passed` |
| This session, separately | replaced the same wrapper with a passthrough of identical arity | `Tests 1 passed (1)`, 7.12 s |

Not even the weaker `expect(resAlias!.status).toBe(200)` goes red. The whole test is green on the regression.

## Why — an upstream lock already forces the ordering

`readPriorCastForMerge` (`analysis.ts:220`) opens with `withCastLock(bookDir, …)`, and it is called at `analysis.ts:3534` whenever `recordRef.bookDir` is set (`recordRef.bookDir ? await readPriorCastForMerge(...) : ...`) — strictly *before* the `requestedFresh` block containing the delete.

In this test, `add-alias` is parked inside that same cast lock on the test's `gate`. So the analyzer job cannot get past `readPriorCastForMerge` until `add-alias` releases — and `add-alias`'s write happens *inside* that hold, via `writeJsonAtomic` before its HTTP response. By the time the job reaches the delete at all, the write has already landed on disk.

The ordering the delete's own lock is supposed to enforce is therefore **already guaranteed upstream**, whether or not that delete is wrapped. From this test's vantage point the delete's lock is unfalsifiable.

This is independent of how the test waits. It holds for the original fixed sleeps and for every synchronisation form tried in PR #3009.

## It also blocks the obvious repair

PR #3009's review (finding 3) proposed replacing the 100 ms head start with a real happens-before edge, suggesting an intercept on the job's own lock acquisition. That was tried and **deadlocks**, for the same structural reason: the job cannot reach `clearAnalysisCache` (`analysis.ts:3639`) until `add-alias` releases the lock, but the release was gated on that signal. Observed: `clearAnalysisCache never completed within 5000ms`, plus a genuine `LockAcquisitionTimeoutError` from the job's own 10 s budget.

So there is no earlier test-visible edge to await with the current interceptor shape, and no duration-free synchronisation point for that head start.

## The decision owed

More than one defensible outcome, which is why this is filed rather than fixed:

1. **Neutralise the upstream serialisation for this test** — stub or bypass `readPriorCastForMerge`'s lock so the delete's own lock becomes the only thing ordering the two writers, making it falsifiable again. Most faithful to the original intent; also the most invasive, and risks the test proving something the production path never exercises.
2. **Retarget the test** at the invariant that *is* reachable — that a "Start fresh" delete and a concurrent alias write cannot interleave to leave a resurrected roster, whichever lock enforces it. Honest and cheap, but stops pinning the specific line #1981 Task 11 was about.
3. **Accept that `analysis.ts:3535`'s lock subsumes it** and delete or downgrade this test, recording that the delete's own wrapper is defence-in-depth rather than the load-bearing guard. Cheapest. Note the caveat originally written here — “loses coverage if the upstream call is ever made conditional” — was already wrong: that call **is** conditional, on `recordRef.bookDir`. So the subsumption only holds for books with a directory on disk, which weakens this option rather than strengthening it.

Whichever is chosen, the head-start sleep at `analysis.fresh-cast-lock.test.ts:352` should be resolved in the same pass, since its only remaining justification is that it gates nothing meaningful.

## Scope note

This is **not** caused by PR #3009. That PR fixes a real, separate defect — a 300 ms fixed sleep that raced under load and failed the commit gate three times — and its changes are improvements on their own terms. It merely ran the mutation that exposed this. PR #3009 discloses this issue rather than claiming to fix it.

## Key files

- `server/src/routes/analysis.fresh-cast-lock.test.ts` — the test
- `server/src/routes/analysis.ts:3660` — the delete's `withCastLock`, the thing meant to be under test
- `server/src/routes/analysis.ts:220` + `:3534` — `readPriorCastForMerge`'s lock and its (conditional) call site
- `server/src/routes/book-state-preserve-voices.test.ts` — the sibling with the same interceptor shape; worth checking whether it has the same problem (it is the live contention row in the flaky register, #2226)

## Benefit

*Benefit (technical):* a test that cannot fail is worse than no test — it reports coverage of the #1981 Task 11 invariant that does not exist, and would let a real removal of that lock merge green. Resolving it either restores the coverage or removes a false signal.

---

**Correction (2026-09-06, from PR #3009 review pass 2).** This issue originally described the upstream call as *unconditional*. It is not — `analysis.ts:3534` guards it on `recordRef.bookDir`. The core finding is unaffected (in this test `bookDir` is always set, so the upstream lock does serialise the two writers and the delete's own lock remains unfalsifiable here), but option 3's reasoning changes: the subsumption is narrower than stated.

dudarenok-maker:
## Summary

`server/src/routes/analysis.fresh-cast-lock.test.ts` claimed to pin the `withCastLock` wrapper around the "Start fresh" `cast.json` delete (`analysis.ts:3825`, the #1981 Task 11 invariant).

**It could not.** `readPriorCastForMerge` takes the same cast lock upstream, before the delete block, so replacing that wrapper with a passthrough left the test **fully green** — verified twice by mutation during PR #3009. A test that cannot detect its own regression is worse than no test: it reports coverage that does not exist.

This is the fix for that, implementing **option 2** of the three #3022 named, as chosen by the repo owner.

## What changed

The test now pins the **outcome** rather than the mechanism: after a "Start fresh" delete races a concurrent `add-alias` write, `cast.json` must end up deleted and must not be resurrected — **whichever lock enforces it**.

- `describe` renamed to say that.
- The file header and the comments that carried the `#3022` warning now state what *is* pinned, rather than warning about a defect that no longer exists.
- No production code. `analysis.ts` and `cast-lock.ts` are untouched; `git diff` on both is empty.

## It is falsifiable — which was the entire bar

#3022 exists because a test could not fail. So the deliverable is not the rename, it is the demonstration.

Reducing `withCastLock` to a passthrough — removing protection from **every** reachable call site at once (at least three: `readPriorCastForMerge`, the delete block, and add-alias's own wrapper in `cast-aliases.ts` — "both" undercounted this in an earlier version of this body, review pass 2 finding B), which is the right mutation for "whichever lock enforces it" — turns the test red:

```
Error: cast.json is still present after 5000ms — either the "Start fresh" delete
never completed, or it completed and a concurrent writer resurrected the file.
Both are failures of this invariant; check the job log for a
cast_merge_base_stale advisory, which indicates the latter.
```

**Single-site removals, measured rather than assumed** (each one site at a time, same-arity passthrough, restored byte-identical between runs):

| Lock removed | Result |
|---|---|
| add-alias's own (`cast-aliases.ts`) | **RED** |
| the "Start fresh" delete's own (`analysis.ts`) | green |
| `readPriorCastForMerge`'s (`analysis.ts`) | green |
| both `analysis.ts` sites together, `cast-aliases.ts` untouched | **RED** (measured after review pass 2, not assumed — see the file's own comment) |

An earlier version of this body said *"removing only one lock still passes"*. That is false, and review pass 1 was right to flag it — add-alias's own lock is a single site whose removal the test does catch. The two that stay green are the pair on the job's side, and that is the honest shape of what option 2 bought: the test no longer pins the #1981 Task 11 delete wrapper specifically (disclosed, and the reason #3022 chose between three options), but it is not indifferent to locking either.

I re-ran all three myself rather than carrying the reviewer's numbers into a permanent code comment unverified.

## Two changes beyond the brief, both about trusting the signal

**The poll's failure message was misdiagnostic.** It said only `cast.json still exists; delete has not completed`. That is the assertion a real regression reddens — so a resurrection would have been reported as contention, the same misdiagnosis PR #3009 review pass 2 finding 1 caught in the previous wording. It now names both causes and points at the advisory that distinguishes them.

**My first mutation was broken and would have produced false evidence.** It rewrote `await withCastLock(dir, fn)` as `await (async () => {…})`, which awaits a *function* rather than calling it — so the body never ran and the "lock removal" tested nothing. Caught before drawing any conclusion from it, and redone by making `withCastLock` a passthrough, which is syntactically safe and removes every reachable site's protection at once.

## Deliberately unchanged

- **The 100 ms head start stays.** Under this invariant it is a legitimate interleaving window, not a synchronisation guess, and the comment now says so plainly instead of apologising for it. The signal-based replacement is **not** retried: it deadlocks (#3009), because the job cannot reach `clearAnalysisCache` until `add-alias` releases the very lock the signal was gating. That experiment has been paid for once.
- `SYNC_WAIT_TIMEOUT_MS`, `RESURRECTION_SETTLE_MS`, and the `60_000` `it()` budget — unrelated to what the test asserts.

## Test plan

- Test green in isolation, `--retry=0`; body ~4.7 s.
- **Red under the every-site passthrough mutation**, message above; `analysis.ts` and `cast-lock.ts` restored byte-identical, verified by empty `git diff` on each. Also red under the narrower two-`analysis.ts`-site mutation with `cast-aliases.ts` left intact (review pass 2 finding B), same restoration discipline.
- `server/src/workspace/cast-lock.guard.test.ts` still passes (4) — it pins unlocked `cast.json` writers and must not be disturbed by reasoning about locks.
- `tsc -p .` and `eslint` clean.

### Checklist notes

- **Regression plan** — not applicable; #3022 plus the changed test is the spec.
- **On-box acceptance** — not applicable; no hardware-only behaviour.
- **Release notes** — skipped: test-only, no user- or operator-visible delta.

Closes #3022
Refs #3007

https://claude.ai/code/session_013DFfsAoY1LtxjDgnGPSZkc

### DIFF
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -26,8 +26,25 @@
    between the "Start fresh" delete and Phase-0's own per-chapter cast.json
    mirror write (analysis.ts, `interim.length > 0` block) touches cast.json,
    so holding Phase 0 open guarantees the only write in play during the
-   race window is analysis.ts's locked delete and add-alias's own write —
-   never the job's own later, legitimate re-creation of cast.json. */
+   race window is analysis.ts's own writer(s) and add-alias's own write —
+   never the job's own later, legitimate re-creation of cast.json.
+
+   #3022 retarget (2026-09-06, owner decision "option 2"): this test pins the
+   OUTCOME — cast.json stays deleted after a "Start fresh" delete races a
+   concurrent add-alias write — not any one lock site. An earlier version of
+   this file asserted that outcome AND claimed it as proof the delete's own
+   `withCastLock` (analysis.ts, inside the `requestedFresh` block) was what
+   enforced it; that claim was false; replacing that wrapper with a
+   passthrough of identical arity left the assertion below green, because
+   `readPriorCastForMerge` (analysis.ts) takes the SAME cast lock earlier in
+   the same job, unconditionally whenever `recordRef.bookDir` is set, strictly
+   before the delete — so add-alias's write is already serialised against the
+   job well before the delete is reached, regardless of the delete's own
+   wrapper. The outcome is real and worth gating; which specific call site
+   delivers it is not something this test can pin without also breaking on a
+   configuration production never runs (removing the upstream lock, which
+   nothing in production ever does). See the assertion at the bottom of the
+   `it` body for what mutation actually reddens this test. */
 
 import { describe, it, expect, beforeAll, afterAll, afterEach, vi } from 'vitest';
 import { mkdtempSync, rmSync, mkdirSync, writeFileSync, existsSync } from 'node:fs';
@@ -162,7 +179,7 @@ function buildPhase1Analyzer(): Analyzer {
   };
 }
 
-describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent cast writer', () => {
+describe('#1981 Task 11 — "Start fresh" cannot be resurrected by a racing cast writer', () => {
   it(
     'an add-alias write does not resurrect cast.json after a concurrent "Start fresh" delete',
     async () => {
@@ -291,9 +308,10 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
              would release the delete while the read was still pending, which
              is harmless while the route holds withCastLock but gives strictly
              LESS slack than the 300ms sleep did. Deliberately NOT argued from
-             the lock-removed mutation, the way the sibling's comment is: per
-             #3022 this file cannot detect that mutation, so citing it here
-             would be reasoning from an experiment that does not run. */
+             a single named lock site, the way the sibling's comment is: this
+             file pins the outcome across whichever lock enforces it (#3022
+             retarget), not one call site, so citing a specific wrapper here
+             would overstate what this interceptor's timing actually proves. */
           signalIntercepted();
           await gate; // hold the RESOLUTION open until released below
           return value;
@@ -352,13 +370,13 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
 
         // Finding 3 (PR #3009 review pass 1) — a `clearAnalysisCache`-completion
         // signal was tried here as a replacement for this fixed sleep. It
-        // deadlocks: `readPriorCastForMerge` (analysis.ts:3535, unconditional
+        // deadlocks: `readPriorCastForMerge` (analysis.ts, unconditional
         // whenever bookDir exists, and unconditionally BEFORE this
         // `requestedFresh` block) takes the SAME cast lock add-alias is
         // holding open below, so the job cannot reach `clearAnalysisCache`
-        // (analysis.ts:3639) — let alone signal its completion — until
-        // add-alias's lock is released. But `released()` below is gated on
-        // that very signal, so nothing ever fires; observed
+        // (further down in the same block) — let alone signal its
+        // completion — until add-alias's lock is released. But `released()`
+        // below is gated on that very signal, so nothing ever fires; observed
         // `LockAcquisitionTimeoutError` after 10s and the test's own 5s wait
         // timing out. No duration-free synchronization point is AVAILABLE for
         // this head start with the current interceptor shape: the job cannot
@@ -367,16 +385,22 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         // not non-existence (PR #3009 review pass 2) — a lock-queue-ENTRY
         // edge (workspace/file-lock.ts:232, synchronous, fires while add-alias
         // still holds) would not deadlock; it is simply not exposed to a test
-        // today, and adding that seam to production is part of #3022's
-        // decision rather than something to smuggle in here.
-        // Kept as the pre-existing fixed sleep; the fact that it does
-        // not gate anything meaningful is the subject of the "CRITICAL CHECK"
-        // finding below the core assertion — this needs a design pass (filed as #3022), not a
-        // synchronization swap, so it is deliberately left rather than
-        // "fixed" into something else. Generous head start: the delete either
-        // completes immediately (unlocked — the bug window) or queues behind
-        // add-alias's held lock (locked — the fix). Not a tight window either
-        // way.
+        // today. Do not re-run this experiment: it was tried once already and
+        // deadlocked for the reason above, which does not change with #3022's
+        // retarget below.
+        //
+        // #3022 retarget: under the outcome-level invariant this file now
+        // pins, this fixed sleep has exactly one job — give the job and
+        // add-alias a chance to actually interleave, so a real regression has
+        // a window to occur in. That is a legitimate use of a fixed delay
+        // (it does not gate correctness of anything the test asserts; it only
+        // widens the odds a wrong writer gets to run during the race), not a
+        // synchronisation guess standing in for a signal that doesn't exist —
+        // the sleep below is that window, not a timing assumption to defend.
+        // Generous head start: the delete either completes immediately
+        // (unlocked at every site — the bug window) or queues behind
+        // add-alias's held lock (locked at any site that matters — the fix).
+        // Not a tight window either way.
         await new Promise((r) => setTimeout(r, 100));
 
         released();
@@ -391,7 +415,21 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
         await vi.waitFor(
           () => {
             if (existsSync(castPath)) {
-              throw new Error('cast.json still exists; delete has not completed');
+              /* Message names BOTH causes deliberately. Under the retargeted
+                 invariant this is the assertion a real regression reddens, and
+                 the two ways to get here are indistinguishable from inside the
+                 poll: the delete may not have run yet (contention), or it ran
+                 and a racing writer put the file back (the #1981 Task 11 bug).
+                 A message naming only the first blames contention for the
+                 second, which is exactly the misdiagnosis PR #3009 review pass
+                 2 finding 1 caught in the previous wording. */
+              throw new Error(
+                `cast.json is still present after ${SYNC_WAIT_TIMEOUT_MS}ms ` +
+                  '— either the "Start fresh" delete never completed, or it ' +
+                  'completed and a concurrent writer resurrected the file. ' +
+                  'Both are failures of this invariant; check the job log for ' +
+                  'a cast_merge_base_stale advisory, which indicates the latter.',
+              );
             }
           },
           { timeout: SYNC_WAIT_TIMEOUT_MS, interval: 10 },
@@ -428,19 +466,23 @@ describe('#1981 Task 11 — "Start fresh" cast.json delete races a concurrent ca
       }
 
       expect(resAlias!.status).toBe(200);
-      /* The core assertion: whichever side acquired the lock first,
-         cast.json ends up deleted, never resurrected with add-alias's stale
-         snapshot.
-
-         #3022 — READ THIS BEFORE TRUSTING THIS TEST AS A GATE. It does not
-         currently detect removal of the `withCastLock` at analysis.ts:3660,
-         the regression it was written for: replacing that wrapper with a
-         passthrough leaves this test fully green (verified twice, PR #3009).
-         `readPriorCastForMerge` (analysis.ts:220, called unconditionally at
-         :3535, BEFORE this block) takes the same cast lock, so add-alias's
-         write has already landed by the time the job reaches the delete at
-         all — the ordering is forced upstream whether or not the delete is
-         wrapped. #3022 names the decision owed. */
+      /* The core assertion, and what this test pins after the #3022 retarget:
+         an add-alias write cannot interleave with a "Start fresh" delete to
+         leave cast.json resurrected — whichever lock in analysis.ts's
+         "Start fresh" path is what actually enforces that. This test does
+         NOT claim to pin any one call site: `readPriorCastForMerge`
+         (analysis.ts) takes the cast lock unconditionally, before the delete
+         block, whenever `recordRef.bookDir` is set, so add-alias's write is
+         already serialised against the job well before the job reaches the
+         delete — which is exactly why asserting this outcome does not, by
+         itself, prove the delete's own `withCastLock` is doing anything (see
+         the file header). What this assertion DOES catch: removing cast-lock
+         protection from BOTH `readPriorCastForMerge` and the delete block
+         together, which lets the job race straight through to the delete
+         while add-alias is still mid-write — the actual "Start fresh
+         resurrects the roster" bug #1981 Task 11 exists to prevent. Verified
+         by mutation; see the PR description for the mutation and its
+         verbatim red output. */
       expect(castExistsAfterRace).toBe(false);
     },
     /* Runaway backstop, not a synchronisation deadline. 60_000 is this suite's
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -43,8 +43,16 @@
    wrapper. The outcome is real and worth gating; which specific call site
    delivers it is not something this test can pin without also breaking on a
    configuration production never runs (removing the upstream lock, which
-   nothing in production ever does). See the assertion at the bottom of the
-   `it` body for what mutation actually reddens this test. */
+   nothing in production ever does).
+
+   WHICH assertion gates this test: the `vi.waitFor` poll for the delete, not
+   the `expect(castExistsAfterRace)` at the bottom. Under the mutations that
+   break the invariant, add-alias's write lands before the poll ever observes
+   an absent file, so the poll throws and the bottom assertion is never
+   reached. An earlier version of this header sent the reader to the bottom
+   assertion; that was wrong (PR #3060 review pass 1, finding 1), and naming
+   the reddening assertion correctly is precisely what #3022's decision asked
+   for. */
 
 import { describe, it, expect, beforeAll, afterAll, afterEach, vi } from 'vitest';
 import { mkdtempSync, rmSync, mkdirSync, writeFileSync, existsSync } from 'node:fs';
@@ -437,17 +445,20 @@ describe('#1981 Task 11 — "Start fresh" cannot be resurrected by a racing cast
 
         /* Settle window, RESTORED (PR #3009 review pass 2, finding 1). The poll
            above returns at the first instant of absence, so sampling absence
-           immediately after it is tautological — that assertion could not
-           fail, and a real resurrection would instead surface as the poll's own
-           "delete has not completed" message, blaming contention for what is
-           actually a resurrection. The regression this file exists to catch is
-           a write landing AFTER the delete, so the sample has to be taken at a
-           moment the poll did not choose. Both properties kept: poll for the
-           delete (no fixed duration gating "has it happened yet"), THEN settle
-           before sampling. A fixed duration is correct here and is not the
-           mistake this PR fixes elsewhere — it is a window for a late
-           writer to be caught in, not a guess at how long something takes, so
-           erring long only makes the check stronger. */
+           immediately after it would be tautological, so the sample is taken
+           at a moment the poll did not choose.
+
+           Be honest about what this currently buys: NOTHING on today's code
+           path. Per the reasoning above, no writer exists between the poll
+           seeing the file gone and this sample, so this window waits 400ms for
+           something that cannot happen, and the assertion it feeds cannot fail
+           (PR #3060 review pass 1, finding 2). It is retained as a FORWARD
+           guard, not as the gate: the invariant is "must not be resurrected",
+           and a writer added later between the delete and stage 1 would be
+           caught here and NOT by the poll, which exits at the first instant of
+           absence. If that ever stops being worth 400ms, delete the window and
+           the assertion together rather than leaving an assertion that reads
+           like a gate. The gate is the poll above. */
         await new Promise((r) => setTimeout(r, RESURRECTION_SETTLE_MS));
 
         // Capture disk state NOW, before Phase 0 (still gated) is allowed to
@@ -476,13 +487,18 @@ describe('#1981 Task 11 — "Start fresh" cannot be resurrected by a racing cast
          already serialised against the job well before the job reaches the
          delete — which is exactly why asserting this outcome does not, by
          itself, prove the delete's own `withCastLock` is doing anything (see
-         the file header). What this assertion DOES catch: removing cast-lock
-         protection from BOTH `readPriorCastForMerge` and the delete block
-         together, which lets the job race straight through to the delete
-         while add-alias is still mid-write — the actual "Start fresh
-         resurrects the roster" bug #1981 Task 11 exists to prevent. Verified
-         by mutation; see the PR description for the mutation and its
-         verbatim red output. */
+         the file header).
+
+         This assertion is NOT what catches the mutation, despite reading like
+         it (PR #3060 review pass 1, finding 1). Removing cast-lock protection
+         from both `readPriorCastForMerge` and the delete block does redden the
+         test — at the POLL above, because add-alias's write lands before the
+         poll ever sees the file gone, so execution never reaches this line.
+         Review pass 1 measured which single-site removals redden it too:
+         add-alias's own wrapper (`cast-aliases.ts`) does; the delete's wrapper
+         alone and `readPriorCastForMerge`'s alone do not — the latter being
+         the #1981 Task 11 line this file is named for, and the disclosure
+         #3022 signed up for when it chose to retarget. */
       expect(castExistsAfterRace).toBe(false);
     },
     /* Runaway backstop, not a synchronisation deadline. 60_000 is this suite's
--- a/server/src/routes/analysis.fresh-cast-lock.test.ts
+++ b/server/src/routes/analysis.fresh-cast-lock.test.ts
@@ -479,7 +479,7 @@ describe('#1981 Task 11 — "Start fresh" cannot be resurrected by a racing cast
       }
 
       expect(resAlias!.status).toBe(200);
-      /* The `vi.waitFor` poll below is the gate that this test pins after the
+      /* The `vi.waitFor` poll above is the gate that this test pins after the
          #3022 retarget: an add-alias write cannot interleave with a "Start
          fresh" delete to leave cast.json resurrected — whichever lock in
          analysis.ts's "Start fresh" path is what actually enforces that. This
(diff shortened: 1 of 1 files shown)


dudarenok-maker:
## PR review — pass 1 (head bf599c3a, depth low)

Scope: the one changed file, `server/src/routes/analysis.fresh-cast-lock.test.ts`, read against `analysis.ts` (`readPriorCastForMerge` :223–245, its call site :3699–3701, the `requestedFresh` delete block :3803–3841), `workspace/cast-lock.ts`, `workspace/file-lock.ts` and `routes/cast-aliases.ts`. Verified before probing: baseline green (1 passed, 7.66 s body), and a five-cell mutation matrix run by hand — restore by file copy each time, `git status --porcelain` empty and HEAD still `bf599c3a` after every cell.

### Mutation matrix (measured, not asserted)

| Mutation | Result | Assertion that reddened |
|---|---|---|
| `withCastLock` → passthrough (all sites) | **RED** | poll, `:426` |
| `withCastLock` → passthrough **and** head start 100 ms → 0 ms | **RED** | poll, `:426` |
| delete-site wrapper only (`analysis.ts`, the #1981 Task 11 line) | green | — |
| `readPriorCastForMerge` wrapper only (`analysis.ts:225`) | green | — |
| add-alias's own wrapper only (`cast-aliases.ts:331`) | **RED** | poll, `:426` |

The red is red for the claimed reason, not a timeout or a 500: the mutant job log orders `phase=0 Discarded cached progress — starting from scratch` (emitted after the delete block) **before** `[cast-aliases] … added alias "Supernova" to nova`. The delete completed and add-alias put the file back — the #1981 Task 11 bug. `resAlias.status` never tripped. The PR's verbatim red output matches mine exactly.

**Every red is at the poll, `:426`. None reaches `:486`.** That is where the two findings below come from.

### 🟠 Significant (correctness) — the file attributes detection to an assertion that never runs

`analysis.fresh-cast-lock.test.ts:479–485`, the comment above `expect(castExistsAfterRace).toBe(false)`:

> What this assertion DOES catch: removing cast-lock protection from BOTH `readPriorCastForMerge` and the delete block together … Verified by mutation

Under that exact mutation `:486` is never evaluated. `vi.waitFor` throws at `:426` inside the `try`; control jumps to the `finally` at `:456`; `castExistsAfterRace` is assigned at `:455`, which is unreachable. Vitest's own frame in the failing run says so: `❯ vi.waitFor.timeout src/routes/analysis.fresh-cast-lock.test.ts:426:21`.

The file header compounds it — `:46–47`: "See the assertion at the bottom of the `it` body for what mutation actually reddens this test." The assertion at the bottom of the `it` body is the one assertion that does **not** redden.

Concrete break: a later maintainer trimming this test for speed (shorten `SYNC_WAIT_TIMEOUT_MS`, or swap the poll for a plain `await`) reads `:46–47` and `:479–485` as saying the final `expect` is the gate, removes or weakens the poll, and deletes the only assertion that can fail — reproducing #3022 in the same file. #3022's decision comment made *"the implementing PR must show a mutation that reddens it, and name which assertion reddened"* a requirement for exactly this reason. The **PR body** gets it right; the **file**, which is what survives the PR, names the wrong one.

### 🟠 Significant (correctness) — `expect(castExistsAfterRace).toBe(false)` (`:486`) cannot fail

By this file's own reasoning at `:411–414` — "No resurrection write is possible after resAlias resolves, since the add-alias write has already landed on disk (writeJsonAtomic happens before the HTTP response) and Phase 0 is still gated" — there is no writer at all between the poll's first observation of absence and the sample at `:455`. So `RESURRECTION_SETTLE_MS` at `:451` waits 400 ms for a writer the same file argues cannot exist, and the sample that follows it can only ever read `false`.

`:440–442` already concedes the point ("sampling absence immediately after it is tautological — that assertion could not fail") and the settle window was restored to repair it, but a settle window only helps if *something* can write during it, and here nothing can: between the delete and Phase 0's `interim.length > 0` mirror write nothing touches cast.json (the file header's own claim, `:26–30`), and Phase 0 stays gated until the `finally`. I could construct no mutation that reaches `:486` — all three reddening cells above stop at `:426`.

Net effect of the two together: this PR labels the file's un-failable assertion "The core assertion" and gives the assertion that actually gates the invariant a comment describing it as a message-wording fix. For a PR whose single deliverable is "a test that can FAIL", that inversion is worth correcting before merge, even though the test as a whole does fail.

### 🟡 Minor — the settle-window comment quotes a message this PR deleted

`:442–444` still reads: *a real resurrection would instead surface as the poll's own `"delete has not completed"` message, blaming contention for what is actually a resurrection.* That string no longer exists — this PR rewrote it at `:426–432` to name both causes. So the stated rationale for `RESURRECTION_SETTLE_MS` argues from a misdiagnosis the same diff removed. A comment the change made false.

### 🟡 Minor — the "which mutation reddens" claim is incomplete, and the PR body's "only one lock" generalisation is measurably false

Removing **add-alias's own** `withCastLock` (`cast-aliases.ts:331`) *alone* reddens the test — a single-site, entirely realistic regression, and better residual coverage than the file claims. `:479–483` names only the both-analysis-sites mutation. The PR body goes further: *"Removing only **one** lock still passes, exactly as #3009 found."* That holds for the two `analysis.ts` sites (both verified green above) and is false for the third. Relatedly, the passthrough mutation removes **three** relevant `withCastLock` sites, not two — add-alias's included — so "removing both locks at once" understates the experiment that was run.

### 🟡 Minor — line citations stripped in one place, kept in another

The PR correctly drops the rotted `analysis.ts:3535` / `:3639` (the real lines today are 3699 and 3804). It leaves `workspace/file-lock.ts:232` (`:386`) and `workspace/file-lock.ts:239-247` (`:342`). Both still resolve correctly today — `:232` is `chains.set(key, mine)`, `:239–247` is the `unref` block — so nothing is wrong; the rot rationale just applies equally.

### ✅ What is solid

- **The deliverable is met.** The retarget is genuinely falsifiable, and the reddening cause is the claimed cause — verified independently here, not taken from the PR body.
- **The disclosed blind spot is real and honestly disclosed.** Removing the #1981 Task 11 delete wrapper alone: green. Removing `readPriorCastForMerge`'s wrapper alone: green. The PR states this rather than hiding it, and it is what owner-decision option 2 signed up for.
- **`cast_merge_base_stale` is real and reachable.** `analysis.ts:3647–3656`, and it actually fires in the failing run: `[analysis] cast_merge_base_stale mns=… site=interim expected=ABSENT observed=ba3761616d76`. The poll's disambiguation hint is actionable, not decorative. Worth knowing it reaches the log only because the `finally` releases Phase 0 and awaits the job — the advisory is emitted after the assertion has already thrown.
- **The 100 ms head start survives adversarial probing.** With the passthrough *and* the head start cut to 0 ms the test is still RED — the real interleaving window is `writeJsonAtomic`'s own I/O, so the sleep is belt-and-braces rather than a load-bearing timing guess. And with the locks intact the job blocks at `readPriorCastForMerge` whatever the sleep length, so it cannot produce a red flake. Kept, correctly; the signal-based alternative is not re-proposed.
- **House gates.** No production code (`git diff` on `analysis.ts` and `cast-lock.ts` empty, confirmed). `Closes #3022` plain, outside any code span. Release-notes skip explicitly justified (test-only). On-box acceptance N/A. #3022 carries the standalone `bug` label, so no `docs/BACKLOG.md` row is owed. Depth `low` matches single-scope `test(server)`. `#2015` header claims re-checked: `readJsonWithFingerprint` really does use `readFile` (`cast-fingerprint.ts:85`), and add-alias really does 409 on an absent cast inside its own lock (`cast-aliases.ts:333–337`).
- Restored tree after every mutation: `git status --porcelain` empty, HEAD `bf599c3a`; `analysis.fresh-cast-lock` + `cast-lock.guard` + `cast-lock.race` = 3 files, 8 tests, all passing.

### Verdict

**Nothing blocks merge.** The test can fail, it fails for the right reason, and the coverage it does not have is disclosed rather than overclaimed — the bar #3022 set is cleared.

Two 🟠 correctness findings remain, and both are in the file's own comments rather than in behaviour: the retarget hands the "this is what catches the regression" label to `:486`, which cannot fail and never runs, while the assertion that actually catches it (`:426`) is annotated as a message-wording fix. That is the same defect class #3022 was filed for — a claim about what this test detects that its own mutation does not support — so it is worth closing in this round rather than leaving for the next reader to rediscover.

dudarenok-maker:
## PR review — pass 2 (head 77fa7b3b, depth low)

Re-review of pass 1 (head `bf599c3a`). Deltas only.

Scope: `server/src/routes/analysis.fresh-cast-lock.test.ts` at head, plus the two intervening commits (`c30a4a9c`, `77fa7b3b`) read against `analysis.ts`, `cast-aliases.ts` and `workspace/file-lock.ts`.

Verified before probing:

- **PR still touches exactly one file** — `git diff --stat main...77fa7b3b` = `analysis.fresh-cast-lock.test.ts | 154 +++---`, 1 file. `git diff main...77fa7b3b -- server/src/routes/analysis.ts server/src/workspace/cast-lock.ts server/src/routes/cast-aliases.ts` is **empty**. No production diff, as claimed.
- **Both commits since pass 1 are comment-only.** Every `+`/`-` line in `git diff bf599c3a..77fa7b3b` falls inside a `/* */` or `//` span; the one executable line in the changed hunks (`await new Promise((r) => setTimeout(r, RESURRECTION_SETTLE_MS));`) is context, unchanged. So pass 1's measured baseline and its five-cell mutation matrix still describe this head exactly. I did **not** re-run the matrix, and nothing below assumes I did.

---

**Pass 1 🟠 #1 (the file attributes detection to an assertion that never runs) — VERIFIED RESOLVED, but the fix introduced a new inversion (finding A below).** The header (`:48–55`) now names the `vi.waitFor` poll and explicitly retracts the old pointer; the bottom comment (`:494–498`) says outright that `expect(castExistsAfterRace)` is *not* what catches the mutation and that the red lands at the poll. Both corrections are accurate.

**Pass 1 🟠 #2 (`expect(castExistsAfterRace).toBe(false)` cannot fail) — VERIFIED RESOLVED.** `:453–463` now states plainly that the window "buys NOTHING on today's code path", that the assertion it feeds cannot fail, that it is retained as a FORWARD guard, and that the window and the assertion go together if the 400 ms ever stops being worth it. That is the labelled-un-failable-assertion shape the repo's own doctrine permits, and it is the right disposition.

**Pass 1 🟡 #3 (settle comment quotes a message this PR deleted) — VERIFIED RESOLVED.** The `"delete has not completed"` quote and the misdiagnosis argument are gone from `:448–463`.

**Pass 1 🟡 #5 (line citations stripped in one place, kept in another) — VERIFIED RESOLVED.** `grep -nE "\.ts:[0-9]"` over the head file returns nothing, so `77fa7b3b`'s own "no line citations remain" claim holds. Both replacement anchors resolve, and resolve uniquely: `file-lock.ts:239` is the "A pending timer keeps the event loop (and vitest) alive; unref so a…" note (the only unref note in the file), and `file-lock.ts:232` is the sole `chains.set(key, mine)`.

**Pass 1 🟡 #4 (incomplete mutation claim; PR body's "only one lock" generalisation false) — PARTIALLY RESOLVED.** The file now records add-alias's single-site red (`:499–501`) ✅, and the PR body replaced the false sentence with a measured three-row table and an explicit retraction ✅. The *other* half of that finding — "the passthrough removes three relevant sites, not two" — survives in the PR body ("removing **both** locks at once") and has now been carried into the file as a claim about a cell nobody ran. That is finding B below.

---

### 🟠 Significant (correctness) — A: the file now names the wrong lock site as "the #1981 Task 11 line this file is named for"

`analysis.fresh-cast-lock.test.ts:499–503`:

> Review pass 1 measured which single-site removals redden it too: add-alias's own wrapper (`cast-aliases.ts`) does; the delete's wrapper alone and `readPriorCastForMerge`'s alone do not — **the latter** being the #1981 Task 11 line this file is named for, and the disclosure #3022 signed up for when it chose to retarget.

The enumeration is [the delete's wrapper] then [`readPriorCastForMerge`'s], so "the latter" points at `readPriorCastForMerge`. That is inverted:

- `analysis.ts:3814` carries the marker, inside the `requestedFresh` delete block (`:3803`): `/* #1981 Task 11 — the delete must be serialised against the other 34 cast.json writers… */`, immediately above `await withCastLock(freshBookDir, …)` and its `rm(castJsonPath(freshBookDir))`.
- `readPriorCastForMerge` (`analysis.ts:223`) carries no Task 11 marker at all — its lock is a `#2015`/`#2155` addition, which this file's own header (`:17–22`) already says.
- `c30a4a9c`'s **commit message** gets it right — "The delete's wrapper is the #1981 Task 11 line this file is named for" — so the inversion was introduced between the message and the file. The file is what survives, which is the exact failure mode pass 1 named one paragraph earlier in this same PR.

Concrete break: a maintainer auditing residual coverage before touching `analysis.ts:3825` (the Task 11 wrapper) reads this line, concludes the wrapper this file is named for is `readPriorCastForMerge`'s, and mis-attributes both the named invariant and the coverage `#3022` deliberately gave up. Both cells are correctly reported green, so the coverage facts survive — but the file's single sentence about *which invariant is now unguarded* points at the wrong function, in the one file whose whole purpose is to stop making false claims about which lock it pins. One word: `latter` → `former`.

### 🟠 Significant (correctness) — B: the file states, as measured, a mutation cell the matrix does not contain, and the red it rests on is over-determined

`analysis.fresh-cast-lock.test.ts:494–498`:

> Removing cast-lock protection from both `readPriorCastForMerge` and the delete block **does redden the test** — at the POLL above…

Sitting immediately above "Review pass 1 measured which single-site removals redden it too", this reads as measured fact. No run in either the PR body's table or pass 1's matrix isolates that cell. What was actually run:

| Cell | Result | In the PR body / pass 1 matrix? |
|---|---|---|
| `withCastLock` → passthrough (**all** sites) | RED | yes |
| delete's wrapper alone | green | yes |
| `readPriorCastForMerge`'s alone | green | yes |
| add-alias's own alone (`cast-aliases.ts:331`) | RED | yes |
| **both `analysis.ts` sites, add-alias's intact** | **never run** | **no** |

And the passthrough red cannot stand in for it. `cast-aliases.ts:331` is itself a `withCastLock` call (`git grep withCastLock -- server/src/routes/cast-aliases.ts` → `:109`, `:240`, `:331`), so the passthrough removes it too — and the matrix separately shows that site alone is sufficient to redden. The passthrough red is therefore **fully explained without either `analysis.ts` site**. Attributing it to the two analysis-side wrappers is over-determination, and it is why the PR body's "removing **both** locks at once" (still present, pass 1 flagged the same wording) understates the experiment.

Concrete stake, not a fragility gesture: if that missing cell is green, then the only lock removal this test can detect is a `cast-aliases.ts` route's own wrapper, and the file's claim to guard the "Start fresh" outcome against analysis-side lock loss is an overclaim of exactly the shape `#3022` was filed to remove — one level up from the original. If it is red, the file's central sentence is true and the residual-coverage story holds. Nobody knows which, and this is a PR whose body claims "I re-ran all three myself rather than carrying the reviewer's numbers into a permanent code comment unverified" — the standard the file's own most load-bearing sentence does not meet. Cost to close: one mutation run (revert `analysis.ts`'s two wrappers to same-arity passthroughs, leave `cast-aliases.ts` alone), or reword to what was measured.

### 🟡 Minor (cleanup) — C: "The core assertion" survives as the lead-in to the assertion the same comment then says is not the gate

`analysis.fresh-cast-lock.test.ts:482` still opens the block with `/* The core assertion, and what this test pins after the #3022 retarget:` — 12 lines above `This assertion is NOT what catches the mutation` and `The gate is the poll above.` (`:463`). That phrase is the one pass 1's verdict quoted as the inversion ("this PR labels the file's un-failable assertion 'The core assertion'"). A skimming reader stops at the lead-in and gets the pre-`c30a4a9c` reading. Materially better than pass 1 — the correction is now unmistakably present in the same block — so cleanup, not correctness.

---

### Verdict

**Nothing blocks merge.** Every pass 1 finding is resolved or materially improved, no production code moved, and the honest-labelling work in `c30a4a9c` (the un-failable assertion, the poll named as the gate) is the right disposition rather than a cosmetic one.

Two 🟠 remain, both in the file's comments, both created or carried by the commits that answered pass 1: A is a one-word inversion with a five-minute fix; B is the sharper form of pass 1's 🟡 #4 and now sits in the permanent comment rather than only in the PR body. Neither changes what the test does — pass 1's verified red is still the red — but this is the one PR in which "the file claims something the mutation does not support" is the defect under repair, so both are worth closing in this round rather than shipping the class the PR exists to remove.

dudarenok-maker:
## PR review — pass 3 (head 4c985b25, depth low)

Re-review of pass 2 (head `77fa7b3b`). Deltas only. This is the final round under the loop cap.

Scope: `server/src/routes/analysis.fresh-cast-lock.test.ts` at head, plus `4c985b25` read against `analysis.ts`, `cast-aliases.ts`. Depth re-derived from the current commit set: `test(server)` + 3 × `docs(server)`, single-scope → `low`, unchanged.

Verified before probing:

- **Still test-file-only.** `git diff main...4c985b25 --stat` = `analysis.fresh-cast-lock.test.ts | 156 ++++---`, 1 file. `git diff main...4c985b25 -- server/src/routes/analysis.ts server/src/routes/cast-aliases.ts server/src/workspace/cast-lock.ts` is **empty**.
- **Baseline green at this head**, `--retry=0`, 1 passed. (Body 14.8 s on this box — 18 worktrees live, so not a quiet run and not comparable to the file's own ~5 s quiet-isolation figure. The 60 s backstop absorbs it; noted as context, not a finding.)
- CI green on all 19 checks at this head.

---

**Pass 2 🟠 A (the file named the wrong lock site as "the #1981 Task 11 line") — VERIFIED RESOLVED.** Checked against the source rather than the commit message:

- `analysis.ts:3814` is the `/* #1981 Task 11 — the delete must be serialised against the other 34 cast.json writers… */` marker, sitting immediately above `await withCastLock(freshBookDir, …)` at `:3825` and its `rm(castJsonPath(freshBookDir))` at `:3826`.
- `readPriorCastForMerge` (`analysis.ts:223–246`) carries no Task 11 marker; its lock at `:225` is the `#2015`/`#2155` addition, exactly as the file header `:17–22` says.
- The enumeration at `:502–503` is [the delete's wrapper alone] then [`readPriorCastForMerge`'s alone], and the sentence now reads **"the former"**. Former = the delete's wrapper = the Task 11 line. Correct.

**Pass 2 🟠 B (a measured claim about a cell nobody had run) — VERIFIED RESOLVED, and independently re-measured here rather than taken from the commit message.**

I ran the exact missing cell myself. Mutation shape — a same-arity passthrough introduced alongside the real import, then swapped in at exactly the two sites named:

```
+const withCastLockMUTANT = async <T>(_bookDir: string, fn: () => Promise<T>): Promise<T> => fn();
-  return withCastLock(bookDir, async () => {          // readPriorCastForMerge, analysis.ts:225
+  return withCastLockMUTANT(bookDir, async () => {
-        await withCastLock(freshBookDir, async () => { // the delete block, analysis.ts:3825
+        await withCastLockMUTANT(freshBookDir, async () => {
```

`cast-aliases.ts:331` untouched (verified by `git diff` and by grep: its three `withCastLock` sites all still call the real one). `analysis.ts:413`'s `reconcileRejectEdgesOnDisk` wrapper also left intact — it is outside the race window. Test file untouched.

**Result: RED, twice, at the poll** — `analysis.fresh-cast-lock.test.ts:436:21`, `vi.waitFor.timeout`:

```
Error: cast.json is still present after 5000ms — either the "Start fresh" delete
never completed, or it completed and a concurrent writer resurrected the file. …
```

**And red for the claimed reason, not contention.** The second run's job log orders it unambiguously:

```
 5: [analysis] … phase=0 Discarded cached progress — starting from scratch.
18: [cast-aliases] book=fresh-lock-author__… added alias "Supernova" to nova
32: [analysis] cast_merge_base_stale mns=… site=interim expected=ABSENT observed=ba3761616d76
```

The `Discarded cached progress` line is emitted *after* the delete block (`analysis.ts:3846`), and it precedes add-alias's write. The delete completed; add-alias put the file back. The `cast_merge_base_stale` advisory the poll's own message points at fires, exactly as advertised. That is the #1981 Task 11 bug, reproduced with `cast-aliases.ts`'s wrapper fully intact — so the file's central sentence is now measured, and the over-determination pass 2 flagged is closed: the red does **not** depend on add-alias's own lock.

`analysis.ts` restored byte-identically (sha256 `74590e8b…` before mutation and after restore), `cast-aliases.ts` unchanged (`8c07c842…`), `git status --porcelain` empty, HEAD still `4c985b25`.

**Pass 2 🟡 C ("The core assertion" lead-in) — RESOLVED as to the phrase; the replacement introduces two new wording defects.** `:482` no longer says "The core assertion". See 🟡 N2 below.

**Pass 2's PR-body carry-over ("removing both locks at once") — VERIFIED RESOLVED.** The body now reads "removing protection from **every** reachable call site at once (at least three: …)" with an explicit retraction, and the measured-cell table gained the fourth row. One attribution nit survives — 🟡 N3.

---

### 🟡 Minor (cleanup) — N1: `4c985b25` re-introduces the one line citation `77fa7b3b` had just removed

`analysis.fresh-cast-lock.test.ts:497`:

> block (while leaving `cast-aliases.ts:331`'s wrapper intact) DOES

`77fa7b3b` — the commit immediately before this one — closed pass 1 🟡 #5 by stripping the last two `file-lock.ts:NNN` citations, and its message states: *"No line citations remain in the file."* `grep -nE "\.(ts|mjs|js):[0-9]"` over the head file now returns exactly one hit, and it is this one, added one commit later. The citation resolves correctly today (`cast-aliases.ts:331` is `return withCastLock(bookDir, async () => {`), so nothing is wrong — it is the same "correct today, rotted tomorrow" shape the file argued itself out of one commit earlier. `cast-aliases.ts`'s add-alias wrapper names it uniquely without the number; two of its three `withCastLock` sites are in other handlers, so a bare `cast-aliases.ts` would be ambiguous and an anchor is genuinely wanted here — `cast-aliases.ts`'s own `add-alias` wrapper is the fix, not deletion of the reference.

### 🟡 Minor (cleanup) — N2: the reworded `:482` lead-in points the reader downward, and leaves a dangling "This assertion"

Two residues of the same edit, both in the block at `:482–492`:

1. `:482` reads `/* The `vi.waitFor` poll **below** is the gate…`. The poll is at `:425–446` — 57 lines **above** this comment. The same block says "at the POLL **above**" at `:498`, and `:463` says "The gate is the poll above". The only thing below `:482` is `expect(castExistsAfterRace).toBe(false)` at `:506`, which this block then says is *not* the gate. The referent is named (`vi.waitFor` occurs exactly once in the file), so a reader who checks cannot land wrong — which is why this is 🟡 and not the pass-1-finding-1 shape — but "below" is simply the wrong word, in the sentence whose whole job is to point at the gate.
2. `:492` `This assertion is retained as a FORWARD guard.` — the paragraph no longer contains the word "assertion" before this sentence; its subject is now "The `vi.waitFor` poll". So the nearest antecedent is the poll, which is the gate, not the forward guard (`:457–463` assigns the FORWARD-guard role to the settle window and `expect(castExistsAfterRace)`). It was correct while the paragraph opened with "The core assertion"; swapping the subject orphaned it. Naming the referent — "the assertion below is retained as a FORWARD guard" — closes both, and would fix (1) at the same time.

### 🟡 Minor (cleanup) — N3: the new measurement is credited to the review pass that reported it as never run

`:495` reads `Pass 2 verified: removing cast-lock protection from both…`, and the PR body's table row reads `**RED** (measured in review pass 2, not assumed)`. Pass 2 measured nothing — it is the pass that flagged the cell as unmeasured and priced the run. The run was done by the author in the fix round afterwards, which `4c985b25`'s own commit message states correctly ("Actually ran the missing cell"). In a file whose subject is "do not state as measured what was not measured", crediting the measurement to the wrong actor is worth one word: `Pass 2 verified` → `Verified after pass 2` (or `Measured in the pass-2 fix round`), and the same in the body's table cell.

---

### Verdict

**Nothing blocks merge, and the loop closes clean.** Both pass 2 🟠 findings are resolved: A against the source, B by re-running the exact cell independently — two RED runs, at the poll, with the delete-then-resurrect log ordering and the `cast_merge_base_stale` advisory that distinguishes resurrection from contention, with `cast-aliases.ts` intact throughout. The claim the file's most load-bearing sentence makes is now a measurement, and it is the author's *and* mine.

Three 🟡 remain, all cleanup, all in comment prose: one line citation re-added a commit after it was removed, one directional word plus a dangling pronoun in the block that names the gate, one misattributed measurement credit. **No correctness findings this pass**, so the re-review trigger does not fire and pass 3 is the end of the loop rather than a cap being hit. The 🟡s are worth folding before merge — this is the one PR where comment prose *is* the deliverable — but none of them is a reason to hold it.

PR #3064, opened 2026-09-06
docs(ops): run Group C on-box acceptance, fold C1-C4 into the register

### LINKED ISSUE #2616: ops — Group C: the Ночной дозор re-analysis session (C1-C4)
Split out of wave 5 (#2606) of the on-box acceptance register campaign (#2435), because it does not fit an unattended linear chain.

## Why its own parent

Group C is **one *Ночной дозор* re-analysis session**. The register's own record of the last one: **12 h 27 m of compute** across an overnight pause, 9 chapters, 15,069 sentences, local Ollama `qwen36-cw-iq4-32k`. Four rows of that inside wave 5's chain would stall every child behind them for days.

This parent exists so the session can be scheduled when the box is free, without blocking anything else.

## The four rows

- **C1** · Free-tier Gemma cloud pass completes end to end (#1685)
- **C2** · Dialogue-convention invariant end to end (#2253)
- **C3** · A deterministic stage-2 failure actually clears when the span is halved (#2304)
- **C4** · The dialogue-collapse guard fires on a real collapse and stays quiet on a healthy book (#2325, #2342)

C2/C3/C4 share the local-analyzer session. **C1 is different** — it needs the free-tier Gemini/Gemma cloud path, and therefore `GEMINI_API_KEY`.

## C1's authorised exception — read carefully

The repo owner has ruled (2026-08-22) that **C1 runs in the primary checkout** `C:\Claude\Projects\Audiobook-Generator`, so it picks up the existing `GEMINI_API_KEY` from `server/.env` already present there.

This is the **single authorised exception** to this campaign's isolation rule, and it is narrow:

- It means *running where the key already is*. It does **not** authorise reading, printing, copying or transcribing the key, and it does **not** authorise copying `server/.env` anywhere. #2345 remains in force.
- It applies to **C1 only**. No other row in this campaign runs in the primary checkout.
- C1 is a **real re-analysis** and therefore **writes** `cast.json` and `state.json` for the book. That is inherent to the row, not something to dry-run around. **Back both files up before starting, verify the backup byte-for-byte, and record the restore path** — so a bad run is recoverable.

## Not yet decomposed

No children are filed. Decompose when the box is schedulable — C2/C3/C4 as one session, C1 separately.

Refs #2435, #2606

dudarenok-maker:
Closes #2616
Refs #2435

## Summary

Group C on-box acceptance run (steps 1-5), folded into `docs/testing/onbox-acceptance-register.md`:

- **C1** (free-tier Gemma cloud pass) — **STILL OWED**. Two independent live-API attempts both died in Phase 0 on Chapter 1 with a genuine `PROHIBITED_CONTENT` block from `gemma-4-31b-it`, falsifying the row's "RECITATION-immune" premise. Run against a throwaway re-import via the primary checkout's already-running server; library book never touched.
- **C2** (dialogue-convention invariant) — **NARROWED further**. A real 9-chapter local re-analysis confirmed every structural criterion (unresolved/flagged populated at the right scale, no ch5 dash-opening collapse, no reproduction of #2306's 87.4% narrator collapse). Only the `state.json` persistence criterion remains, blocked on the run's own drift guard refusing to commit.
- **C3** (deterministic stage-2 failure clears when span halved) — **STILL OWED, no change**. The named ch8/offset19 repeat-loop reproducer did not occur this run; ch8 completed whole via a different failure path.
- **C4** (dialogue-collapse guard fires on collapse, quiet on health) — **DISCHARGED and removed**. The guard fired correctly on every real collapse/coverage breach and stayed quiet on every healthy chapter this run.

Register owed count recomputed by counting `### ` row headings: 65 → 64, Group C 4 → 3. Live view (`onbox-acceptance-register-live-view.html`) re-derived by hand to match.

## Test plan

- [x] `npm run check:onbox-register` — green (register and live view agree)
- [x] Evidence reviewed: `docs/testing/onbox-wave11-group-c-results/step-{1-setup,2-c2c3c4-run,3-c1-cloud}.md` — real executed commands with pasted log output, not descriptions
- [x] Data safety confirmed: C1's primary-checkout exception used a throwaway re-import (`mns_KwZCcqh6JG`), library book (`mns_oyK7Po6BiT`) never opened/touched; `server/.env` byte-for-byte unchanged
- [x] No discharged row left annotated-and-kept — C4 fully removed per the operator's 2026-08-22 standing ruling

This is a docs-only file set (`docs/**`) — the `pr-review-gate` mandatory review is exempt per CLAUDE.md's doc-only fast-path.

### DIFF
--- a/docs/testing/night-watch-reanalysis-onbox-acceptance.md
+++ b/docs/testing/night-watch-reanalysis-onbox-acceptance.md
@@ -647,8 +647,78 @@ Seven further gaps this attempt exposed, all folded into the sections above:
   clear after ~2.5 h. Fixed under
   [#2304](https://github.com/dudarenok-maker/Castwright/issues/2304).
 - **C1 — cloud pass on `gemma-4-31b-it` incl. script-review:** _Result:_
-- **C1 — per-minute 429 retried, not misclassified:** _Result:_
+  **BLOCKED (2026-09-07).** Two independent live attempts (07:03:59–07:04:00.9,
+  then 07:05:38–07:06:01.3) both died in Phase 0 on Chapter 1 with a byte-identical
+  `GeminiContentBlockedError (reason=PROHIBITED_CONTENT)`. Zero of nine chapters
+  completed either time; script-review was never reached (it runs only after a
+  roster and attributed sentences exist). The block is deterministic, per the
+  code's own design comment — a third attempt would reproduce it.
+- **C1 — per-minute 429 retried, not misclassified:** _Result:_ **NOT OBSERVED,
+  either way (2026-09-07).** The run never got far enough into real request volume
+  to hit Google's per-minute cap. The only rate-limiter event seen was the app's
+  own proactive TPM throttle (`{"reason":"tpm","waitMs":60196}`) before Chapter 2's
+  call even went out — correctly classified, but not a live 429 from Gemini.
 - **C1 — working `localInputFraction` for zero truncation drops:** _Result:_
+  **NOT EXERCISED (2026-09-07).** The run died in Phase 0, before any stage-2 call
+  that `localInputFraction` governs — no truncation data of any kind was produced.
+
+**2026-09-05 → 2026-09-07 — Group C step 4's predecessor session (steps 1-3,
+#2896/#2895/#2894). C2/C3/C4 (per the register's current numbering) batched on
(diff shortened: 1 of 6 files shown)


dudarenok-maker:
pr-review-gate: exempt — docs-only file set (`docs/**`), same test as CONTRIBUTING.md's doc-only fast-path. No review pass run.

Issue #3067, opened 2026-09-07
[agent instructions][claude][task] fix stale glance-table counts in onbox live-view (rework of #2956)


dudarenok-maker:
Rework child of batch step 7 verification (**#2956**), parent **#2952**, register campaign **#2435**.

## What broke

Step 6 (commit `6bf39eef` on `chore/ops-analyzer-render-batch`) hand-updated
`docs/testing/onbox-acceptance-register-live-view.html` to fold B2/D1/A37 out
and narrow A16/A31, instead of regenerating it from
`docs/testing/onbox-acceptance-register.md` with the tracked build script. The
"At a glance" summary table (around line 212) was not touched and still
carries the **pre-fold** counts:

```
<!-- BEGIN GENERATED:glance:A -->39<!-- END GENERATED:glance:A -->   (should be 38)
<!-- BEGIN GENERATED:glance:B -->2<!-- END GENERATED:glance:B -->    (should be 1)
<!-- BEGIN GENERATED:glance:D -->3<!-- END GENERATED:glance:D -->    (should be 2)
```

This contradicts the section headers further down the same page, which DO
show the correct post-fold counts (`38 rows` for Group A, etc.), and it
contradicts the top "Owed" stat box (correctly showing 62). The page is
internally self-contradictory. `npm run check:onbox-register` does not catch
this — it only checks glance-table letter presence, never the numeric cell
against the register's actual row count per group.

## What to do

1. Run `npm run register:build` (`scripts/build-register-live-view.mjs`) in
   the worktree at `C:\Claude\Projects\wt-analyzer-render-batch`, branch
   `chore/ops-analyzer-render-batch` — do not create another worktree. This
   is the tracked generator; it should produce byte-correct
   `GENERATED:glance:*` markers directly from the current
   `onbox-acceptance-register.md` state (Group A → 38, Group B → 1,
   Group D → 2), leaving prose sections it does not own untouched.
2. Diff the result before committing — confirm ONLY the glance-table
   generated markers change, nothing else in the file is disturbed.
3. Run `npm run check:onbox-register` and paste its output into the
   evidence/commit — must pass.
4. Commit on the existing branch (do not touch source outside this file).

## Source of the finding

Read-only verification agent run for **#2956** flagged this while confirming
step 6's fold; full verification report (all five rows B2/D1/A16/A31/A37
themselves are independently confirmed PASS with real evidence — this is the
only defect found) is in the closing comment of #2956.

[...]

PR #3063, opened 2026-09-06
ci(ops): add a stale-battery reaper invoked automatically from pre-push

dudarenok-maker:
## Summary

Part **3** of the [commit-gate rebalance design](docs/superpowers/specs/2026-09-05-commit-gate-rebalance-design.md) — `scripts/reap-stale-batteries.mjs`, exposed as `npm run doctor` **and invoked automatically** from `pre-push`. An earlier draft of that design left this as a manual, report-only CLI that nothing ever called, while the risk table credited it with automatic cleanup; closing that gap is the point.

### Two independent tests, because neither alone is sufficient

| Test | Catches | Misses |
|---|---|---|
| Two-sample subtree CPU rate | genuinely stalled work | busy work that cannot land |
| Reachability — top ancestor gone AND no `git.exe` in the subtree | doomed orphans | a live-parented battery that is truly wedged |

The reachability test is a fact about **reachability**, not a threshold about **speed**, which is why it catches what a rate cannot. Not hypothetical: on 2026-09-05 two orphaned subtrees at 5.67 and 2.59 CPU-s/min sat comfortably above any dead bar, doing genuine work that could never land, while starving a `git commit` that had been running 78 minutes. Killing them completed that commit in seconds.

### Three things the earlier census got wrong, fixed here

- **Ancestry resolves through a supervisor-bounded climb**, so a battery's alternating `node → cmd → node` chain collapses to one subtree without merging unrelated batteries all the way to the OS root. An unbounded climb is what produced the discredited "two concurrent batteries are enough" claim, by presenting one battery as several. Caught and fixed mid-implementation.
- **The child/subtree rate is what is judged**, never the supervisor's — a supervisor idles while its child works (~0.9 CPU-s/min is healthy), and three lanes were nearly reaped on that figure.
- **The census log records each root's command line**, which the 2026-09-05 census omitted.

### Never reapable

`python.exe` (TTS sidecars, Ringer), any subtree overlapping a live `git` process, and the caller's own ancestor chain.

### Wiring

`pre-push`, **before** the docs-only branch: one `Win32_Process` query — measured ~694 ms, and ~0.8–3.5 s for a whole census on a 415-root box — **no process pool** — so the hook invariant holds, and the new invocation is added to `hook-no-pool.test.mjs`'s allowlist deliberately. It never blocks a push (`runCli` returns 0 even when the census throws). It runs on docs-only pushes too, because an orphan from an earlier battery does not care what kind of diff is being pushed next.

[...]

PR #3061, opened 2026-09-06
feat(server,sidecar): add per-request X-Device-Hint for lazy Coqui derive

dudarenok-maker:
## Summary

- `server/tts-sidecar/main.py`: `xtts_clone_voice()` now reads an optional `X-Device-Hint` header (e.g. `cuda:1`) and passes it to `PlacementController.reservation()`'s `pinned` argument for that one request. Validated via the same grammar `_engine_env_pin` already uses for `COQUI_DEVICE`/`QWEN_DEVICE` (`_parse_device_hint`, reusing `_resolve_uuid_to_index`/`_parse_device`). Absent or invalid header falls through to today's `_engine_env_pin("coqui")` behavior — byte-for-byte unchanged.
- `server/src/tts/derive-engine-artifact.ts`: adds optional `deviceHint?: string` on `DeriveArtifactInput`, sent as `X-Device-Hint` only when truthy.
- `server/src/tts/clone-voice-resolver.ts`: the lazy Coqui self-heal derive inside `resolveDesignedVoicesForChapter` now computes `lazyCoquiDeriveDeviceHint()` — returns `'cuda:1'` only when `getLastKnownGpuDevices()` (kept warm by `GET /api/gpu/devices`) actually reports an index-1 card, else `undefined`. Every other call site (the qwen arm in the same function, and both arms of `resolveClonedVoicesForChapter`) is untouched and never sends the header.

This lets the lazy Coqui derive (used by designed-voice self-heal) land on GPU1 while Qwen stays resident on GPU0, avoiding the `vram-spill` failure that blocked `#2919`'s E-06 kit on boxes where GPU0 can't hold both engines at once.

## Test plan

- [x] `python -m pytest server/tts-sidecar/tests/test_device_parse.py server/tts-sidecar/tests/test_xtts_clone_voice.py server/tts-sidecar/tests/test_placement.py -q` — 148 passed, 1 pre-existing skip
- [x] `npx vitest run src/tts/derive-engine-artifact.test.ts src/tts/clone-voice-resolver.test.ts` — 113 passed
- [x] `npx tsc -p server/tsconfig.json --noEmit` — clean
[...]

PR #3076, opened 2026-09-07
docs(docs): batch step 7 - fold B2/D1/A37, regenerate live-view glance table

### LINKED ISSUE #2952: ops — analyzer + render batch: B2, D1, A16, A31, A37 on-box confirmation
On-box acceptance register campaign (#2435). Five owed rows, grouped into one
chain because they share one local dev server + TTS sidecar session (Ollama
analyzer, then Coqui/XTTS, then Qwen clone) rather than needing separate
setups. Corrected audit: `docs/testing/onbox-a-series-audit-2026-09-05.md`
(scratch, uncommitted in the primary checkout) — read it for context, but the
register (`docs/testing/onbox-acceptance-register.md`) is the source of
truth; cross-check each row there before acting.

## Rows in this chain

- **B2** — an unset book hits the analysis language gate, prompt resolves it,
  analysis then selects the right conventions table. Analyzer-only, no TTS.
  Register: `### B2` (search `docs/testing/onbox-acceptance-register.md`).
- **D1** — non-English ASR content-QA calibration (es/ru/fr/de). Render real
  audio, run the ASR content-QA gate, inspect WER distribution, set
  `qa.asr.maxWer.{es,fr,de,ru}` from observed data. Register: `### D1`.
- **A16** — a cloned voice renders a non-English book in the book's language
  (chapter-level re-run, post-#1972). Register: `### A16`.
- **A31** — Russian XTTS quality: leading-dash pause (measurable via
  ffprobe/silence-detect, not by ear despite the row's title), Coqui
  degeneracy guard, neuter `-ее` invariant. Register: `### A31`.
- **A37** — Russian dash-attributed dialogue doubled-comma collapse pause
  (same measurable-proxy framing as A31). Register: `### A37`.

## Standing rule for anything that reads "by ear" (2026-09-05, repo owner)

If a criterion genuinely needs a human ear, the agent runs everything it can
first — render the audio, run whatever automated proxy measurement the row's
own criteria describe (duration/silence-detect via ffprobe, ASR-transcript
text match) — and produces a clearly-labeled comparison artifact (before/after
clips, exact timestamps/measurements) ready for a human to judge. It does
**not** declare the row discharged itself. The fold step records this as
"STAGED FOR HUMAN REVIEW" with the artifact location — never as owed-with-no-
progress, never as falsely discharged. A31 and A37 both carry this shape for
their by-ear bullets; the rest of each row's criteria are fully agent-
checkable.

## Where the work happens

Worktree `C:\Claude\Projects\wt-analyzer-render-batch` (created, dependencies
installed, husky active). Branch `chore/ops-analyzer-render-batch`. Server
ports: VITE=5333, API=8240, TTS sidecar=9160 (slot 16, isolated
`server/.env`/`.env.local`, no collision with any other live worktree
checked 2026-09-05). TTS sidecar `.venv`/`voices/` junctioned from the
primary checkout.

Render steps use a **throwaway workspace copy**, never the live primary
checkout's `C:\AudiobookWorkspace` — copy the relevant per-language demo book
folders in, don't operate on the originals in place (memory: hand-editing
real workspace data has broken other lanes' in-flight work before).

Decomposed into an agent-executable chain — see linked children.

dudarenok-maker:
## Summary

Discharges the register campaign's batch step 7 fold: B2, D1, and A37 are
folded out of the acceptance register as discharged, A16 and A31 are
narrowed and staged for human review. Touches
`server/src/config/registry.ts` for D1's threshold defaults alongside the
docs changes, so this is not a docs-only PR.

This branch was very stale (based on an old `main` commit, `8e4025ee`) and
needed a large `origin/main` merge mid-review to pick up the rest of the
#2997 campaign — PR #3073's A1 discharge and C-15 fixes, PR #3096's A16
sentence-19 finding, PR #3083's unreachable-arm removal, and eleven
independent on-box discharges (A5, A29, A36, C4, D3, E4, E6, E8, E10, E101,
E12) plus a new row (E103) that landed on `main` after this branch diverged.
The register conflict was resolved by rebuilding the change-log and glance
table from `origin/main`'s current baseline rather than this branch's stale
one, then applying this branch's own three discharges (B2, D1, A37) on top.
Current, post-merge figures: **52 owed**, Group A 35, Group B 1, Group D 1,
Group E 8 (unchanged by this PR).

The merge itself was verified lossless by comparing row-ID sets across the
true git merge-base, `origin/main`, this branch, and the merged result — the
merged set is the exact expected set difference (main's 11 discharges +
E103, minus this branch's 3 discharges), confirmed by exact diff. A16's
conflict — both this branch and `origin/main` had independently re-evidenced
that still-owed row in separate investigation sessions, finding different
residual defects — was resolved by merging both findings rather than
picking one; they're complementary evidence about the same open row, not
competing claims.

The merge's own CI run then surfaced two real regressions from discharging
B2/D1: nine now-unannotated citations to those row IDs elsewhere in the
tree (`check-register-citations`), and a pre-existing bad relative link in
this PR's own new `a16.md` evidence file (`review-gate-mechanism`'s
intra-repo link checker). Both fixed in follow-up commits.

Closes #2952
Refs #2435
Refs #2956
Refs #3067

## Test plan

- [x] `npm run check:onbox-register` — passes: register.md and live-view.html agree.
- [x] `npm run register:build` — idempotent at the final head (no further diff).
- [x] `npm run check:register-row-citations` and `npm run check:register-citations` — both green (the latter was red immediately after the merge; fixed).
- [x] Row-ID-set diff across true merge-base / `origin/main` / this branch / merged result — exact match, no row silently lost.
- [x] `scripts/tests/check-register-citations.test.mjs` (143/143) and `scripts/tests/review-gate-mechanism.test.mjs` (29/29) — both were red right after the merge, now green.
- [x] `tsc --noEmit` (root + server) — clean.
- [x] Cloud `verify.yml` (required status check) — green.
- [x] `pr-review-gate` — **3 rounds run** (the mandated cap): pass 2 (head `6b80c45c`, before the merge) found 5 cleanup nits and no correctness bugs; pass 3 (head `9111abb6`, after the merge + citation/link fixes) found those 5 nits had never actually been fixed despite being noted, plus 2 new 🟠 editorial findings from the merge's own conflict resolution and 4 🟡 cleanup items — 11 total. All 11 fixed in `3da5cc46`, independently re-verified by re-reading every diff and re-running every check myself rather than trusting the fix report.

**From the merge itself:**
- `check-register-citations.mjs` flagged 9 unannotated citations to B2/D1 (discharged by this PR) across `cast-id-drift-onbox-acceptance.md`, `language-recurrence-onbox-acceptance.md`, `onbox-batch-results/{b2,d1}.md`, and `onbox-sitting-multilanguage.md`. Annotated each within the checker's 120-char proximity window.
- `onbox-batch-results/a16.md`'s relative link to plan 275 used one `../` where two are needed (the file sits two levels under `docs/testing/`, not one) — a genuine pre-existing bug in this PR's own new file, never caught until this merge pulled in a newer version of the link-checker test.

**From pass-3 review (11 findings, all fixed in `3da5cc46`):**
- Pass 2's 5 cleanup nits, genuinely never fixed despite this PR's own earlier claim otherwise: `isLanguageCalibrated()` was a dead distinction (both branches now behave identically) — deleted the helper, its call site, and its two tautological tests (`server/src/tts/segment-asr-qa.ts`/`.test.ts`); `docs/features/248-...md` still said the four WER knobs default to `0.4` (now `0.45`); `d1.md` and `registry.ts` characterized the same es/de percentile data in contradictory ways (fixed `d1.md` to match); the live view carried a dangling `B2` cross-reference (this PR discharges B2, so it resolved nowhere) — reworded; the PR body itself hadn't declared the `perLanguageMaxWer()` precedence change (this section is that declaration).
- My own A16 merge conflict resolution had dropped a lead-in sentence ("**What still stands** — the fix works, proven at the synthesis boundary") that introduced a results table, leaving the table orphaned and two later back-references dangling. Restored (verified word-for-word against `origin/main`'s original text) in both `register.md` and the live-view.
- B2's per-feature run sheet (`language-recurrence-onbox-acceptance.md`) was marked discharged but still had blank `Result:` lines and an unchecked outcome box, plus a header self-contradiction ("fill in as you run" vs. "historical record"). Filled in from the real evidence in `onbox-batch-results/b2.md`, reconciled the header.
- A16's own badge said "one new defect" in one place while carrying two independent open findings (the sentence-19 leak and the QA voice-mismatch flag) — synced to "2 open findings" everywhere.
- The change-log cited an evidence file for D1 only; added the same for B2 and A37 (both permanent, allocate-once removals with their own evidence files).
- `cast-id-drift-onbox-acceptance.md`'s historical "B2" citation is actually a *different, earlier* row that also happened to carry the id "B2" (#2313/PR #2317, a stage-1 cast-names check, discharged separately in 2026-08) — not the row this PR discharges. Disambiguated with the issue/PR numbers so a reader can't confuse the two discharge dates.

## On-box acceptance

**Publishing the live view is NOT a merge precondition, and should not be done from this branch alone.** `main`'s own tracked live view is currently at publish token 13 (PR #3073 published it post-merge). Other register-touching lanes (#3057, #3061, #3063 — not mine, run by a peer session) may still land further discharges before this PR merges. Whichever PR merges to `main` last among the currently-open register-touching lanes should be the one to republish, once, only after `npm run check:onbox-register -- --against-published <saved live page>` agrees against whatever is live at that time.


dudarenok-maker:
## PR review — pass 1 (head b33649e8850fb7df6117056e411df7856d62d2a0, depth high)

Depth `high` per the ladder's multi-scope clause (`fix(server)` + `docs`/`test`), as the PR body computes. Reviewed `git diff main...HEAD` with the focus on the `fix(server)` commit (5afe9fac) and on register/live-view internal consistency. `npm run check:onbox-register` passes at this head; the docs-fold half of the PR is internally consistent (Owed 65→62, A 39→38, B 2→1, D 3→2, glance markers agree with section headers, no discharged row left half-removed).

**4 correctness bugs (1 blocking), 5 cleanup nits.**

---

### 🔴 Correctness bugs

**1. BLOCKING — `server/.env.example:640-648`: derived artifact not regenerated; `npm run config:check` is red at this head.**

`server/.env.example` is generated from the registry by `server/scripts/sync-env-example.ts` (both the `default` value and the `help` string are copied verbatim). 5afe9fac changed `default` `0.4 → 0.45` and rewrote the `help` string for all four of `qa.asr.maxWer.{es,ru,fr,de}` in `server/src/config/registry.ts`, but `.env.example` was not regenerated — it still carries `# SEG_ASR_MAX_WER_ES=0.4` (line 641), `_RU=0.4` (643), `_FR=0.4` (645), `_DE=0.4` (647) under the old `"...defaults to the global ASR max WER until tuned on-box (#1084)."` help text.

Failure scenario, reproduced at this head:

```
$ npm run config:check
.env.example is out of sync with the config registry. Run: npm run config:sync
exit 1
```

`Config check` is a wired CI step whose path filter explicitly includes `server/src/config/` and `server/.env.example` (`scripts/tests/workflow-wiring.test.mjs:481,519`), so required verify reddens on this branch. Second, operator-facing failure: someone uncomments `SEG_ASR_MAX_WER_ES=0.4` believing they are pinning the documented default, and instead sets an *explicit* override that reverts the calibration — and, per finding 2, an explicit override is now the only precedence level that can move a calibrated language, so this silently sticks.

Fix: `npm run config:sync`, commit the result.

**2. `server/src/tts/segment-asr-qa.ts:177-180`: the global `qa.asr.maxWer` override no longer reaches es/fr/de/ru — the four languages this PR is about.**

Before 5afe9fac, `perLanguageMaxWer()` returned `undefined` for any language without an explicit per-language env/app override, so `resolveAsrThresholds` (line 188) fell through to `configValue('qa.asr.maxWer')` — the global knob *including its own live override*. After the change, for the four calibrated languages `state.source === 'default'` and `knob.default (0.45) !== globalKnob.default (0.4)`, so line 179 returns the hardcoded registry literal and the global is never consulted.

Failure scenario: an operator investigating a drift incident sets `SEG_ASR_MAX_WER=0.25` (or moves the "ASR max WER" slider in Settings, `apply: 'live'`) to tighten content-drift detection across the board. English and every uncalibrated language tighten to 0.25. Spanish, French, German and Russian silently stay at 0.45 — a Spanish line transcribed at WER 0.35 that the operator explicitly asked to be flagged `drift` classifies `ok` and ships. The only way to reach those four is four separate per-language env vars, which nothing tells the operator: `qa.asr.maxWer`'s own help is still the unqualified `'Word-error-rate above this threshold is flagged as content drift.'` (`registry.ts:289`) and `docs/wiki/Advanced-Settings.md:250` presents it the same way.

The docstring at `segment-asr-qa.ts:166-170` only claims the cascade still works "for every not-yet-calibrated language", so the asymmetry may be deliberate — but it is undocumented on every operator-facing surface, and **no test covers global-override × calibrated-language**: `segment-asr-qa.test.ts:692-707` deletes only the four `SEG_ASR_MAX_WER_{ES,FR,DE}` vars and never sets `SEG_ASR_MAX_WER` itself. Either pin the intended precedence with a test and say so in both help strings, or make an explicitly-set global win over an uncalibrated-by-override language default. (Catalogue shape 7 — a knob that is correct code but reaches zero of the four languages the operator most likely wants it for.)

**3. `server/src/tts/segment-asr-qa.ts:179-180` + `server/src/tts/segment-asr-qa.test.ts:695-706`: the "calibrated" sentinel is a comparison of two registry literals, and its false branch has zero test coverage.**

`knob.default !== globalKnob.default` infers "this language was calibrated" from two static constants. The `return undefined` at line 180 is unreachable for every knob that exists today (all four differ from the global), and the test that its own comment claims covers it does not: `expect(resolveAsrThresholds(undefined, 'zh').maxWer).toBe(0.4)` (test.ts:705) exits at line 175's `if (!knob) return undefined` — `registry.ts` defines only `qa.asr.maxWer.{es,ru,fr,de}`, there is no `qa.asr.maxWer.zh` or `.ja`. So the branch that decides "not calibrated → cascade" is asserted by nothing.

Failure scenario A: a future on-box row calibrates Italian, and the honest answer from its WER distribution is 0.4. The author adds `qa.asr.maxWer.it` with `default: 0.4`, the whole suite stays green, and Italian is silently uncalibrated — this is exactly the inert-calibration defect `d1.md` Step 4 was written to fix, reintroduced with no guard. Failure scenario B: someone later moves the *global* default to 0.45 (the natural follow-up, given all four calibrated there). In one edit all four per-language knobs flip to "uncalibrated"; behaviour is identical at that instant, so nothing goes red — but from then on a global override silently overrides four caps that were supposed to be pinned. A `calibrated: true` flag on the knob (or a constant recording the pre-calibration baseline) would make the signal explicit instead of inferred; at minimum, add a fixture knob covering the equal-defaults branch.

**4. `docs/wiki/Advanced-Settings.md:251-254`: the operator-facing default column is now false for all four knobs.**

Rows still read `| ASR max WER (Spanish) | ... | 0.4 | ...` etc. `knob-docs-sync.test.mjs` (#2012) joins registry `KNOBS[].label` against this table's *first column only*, so it cannot catch a stale value column — the row exists, the number is wrong, and the guard stays green. Failure scenario: an operator reads the wiki, believes the Spanish cap is 0.4, and sets `SEG_ASR_MAX_WER_ES=0.42` intending a small loosening — actually a *tightening* against the shipped 0.45. CLAUDE.md's requirement is the Settings row lands *in the same change*; here it landed stale.

---

### 🧹 Cleanup nits

**5. `server/src/tts/segment-asr-qa.ts:166-167` — the new JSDoc is factually wrong about its own example.** It says "A language whose knob default still EQUALS the global default (never calibrated, e.g. zh/ja) instead falls through to undefined". zh/ja have no `qa.asr.maxWer.*` knob at all and take the `!knob` path at line 175, not the equal-defaults path. `segment-asr-qa.test.ts:698` states the correct version ("no registry knob at all, e.g. zh/ja per D2"), so source and test now disagree about *why* zh cascades — and the source's version is the one a reader will use to reason about the branch at line 179. (Shape 9.)

**6. `server/src/config/registry.ts:304` overstates the data.** "raised only modestly (0.45, each language's own p75-p80 band)" — per `d1.md`'s own Step 2 table, es p75/p80 = 0.20/0.29 and de = 0.27/0.33. 0.45 is well *above* both bands; for es and de the entire observed p80 sits below even the *pre-existing* 0.4 cap, so 0.45 is a uniform value justified by fr (0.40/0.40) and ru (0.33/0.50), not by those two languages' own distributions. `d1.md` phrases this honestly ("within or just above"); the registry comment is what a future reader re-deriving the number will read, and it claims per-language support that the data does not give. Worth restating as "uniform 0.45, bounded above by the 0.5 substitution floor; es/de's own bands did not require any raise."

**7. `docs/testing/onbox-batch-results/d1.md:236-240` — the evidence file's own verification is left unresolvable.** `## Commit` is still the literal placeholder `` `<filled in after the detached commit completes>` ``, and the Verification section defers `npm run test`'s authoritative pass/fail to the pre-commit hook "(this file records the SHA below once committed)" — the SHA is never recorded. As shipped, the full-suite verification claim points at nothing.

**8. `docs/testing/onbox-acceptance-register-live-view.html:540` — dangling `B2` cross-reference.** A22's body reads "A22 and B2 each prove their mechanism against a single already-drifted chapter or book." After this PR no `B2` row exists at all. (The B2 that sentence originally meant was a *different* earlier B2, discharged at #2606 step 6 — so this was already stale before this PR and is not introduced here; flagging because a discharge fold is the round that should sweep it.) Same for the register's mirrored copy.

**9. House gate — release-notes pair unmet with no declared not-applicable.** `docs/release-notes-next.md` gets no entry for either half of 5afe9fac: four per-language ASR gate defaults loosened `0.4 → 0.45` on live behaviour, and a precedence change to `perLanguageMaxWer()` that removes four languages from the global knob's reach (finding 2). Both are user-visible. The PR body's Summary notes the registry touch ("so this is not a docs-only PR") but neither adds the note nor declares it N/A.

---

### Checked and clean

- `npm run check:onbox-register` — OK; register ↔ live view agree at this head.
- Glance markers (A 38, B 1, D 2) match the section-header row counts and the Owed stat (62); the three discharged rows are removed from both surfaces, A16/A31 narrowed rather than dropped.
- `Closes #2952` / `Refs #2435,#2956,#3067` present in the PR body, outside code spans.
- The `catastrophicWer` backstop is genuinely untouched: 0.45 < 0.85, so `Math.max(catastrophicWer, maxWer)` (`segment-asr-qa.ts:749`) is unchanged for all four languages — the comment's claim holds.
- The paired test is a real regression test for the claimed reason: before the `perLanguageMaxWer()` fix, `resolveAsrThresholds(undefined,'es').maxWer` returns the global 0.4, so the four new `0.45` assertions are red-before/green-after for the stated cause. `d1.md` Step 5 also records the two pinned German drift cases (WER 0.5 / 0.6) failing under the naive p90 pick — the 0.5 ceiling on the threshold choice is evidenced, not asserted.
- Threshold evidence is traceable to a recorded run, not invented: the four `d1-audio/*.segments.json` carry real per-line `asr.wer`/`asr.transcript` from a real faster-whisper pass (no fixture-invents-an-external-contract shape here), and `d1.md` names both residual risks it could *not* exercise rather than claiming them cleared.

dudarenok-maker:
## PR review — pass 2 (head 6b80c45c22a425e1a53bc9f01f47ad13c6e0e147, depth high)

Re-review after `eaee67aa` + `6b80c45c`. Depth `high` (multi-scope: `fix(server)` + `docs`/`test`), unchanged from pass 1. Deltas only — pass 1's "Checked and clean" list is not re-litigated. Read both new commits' diffs directly rather than their messages.

**No new correctness bugs. 4 cleanup nits (1 of them a carry-over).**

---

**1. BLOCKING — `server/.env.example` not regenerated / `config:check` red — ✅ VERIFIED RESOLVED.**
`eaee67aa` resyncs all four `SEG_ASR_MAX_WER_{ES,RU,FR,DE}` lines to `0.45` and copies the new registry `help` strings verbatim (compared against `registry.ts:326,336,346,356` — they match, including German's escaped apostrophe). `npm run config:check` passes at this head.

**2. Global `qa.asr.maxWer` override no longer reaches es/fr/de/ru — ✅ VERIFIED RESOLVED.**
`segment-asr-qa.ts:194` now short-circuits to `undefined` when `resolveKnob(globalKnob).source !== 'default'`, so an explicit `SEG_ASR_MAX_WER` (or the live Settings slider, `apply: 'live'`) cascades to the calibrated four again. Precedence is per-language override → global override → calibrated per-language default → global default, and both the new and the surviving ordering are pinned: `segment-asr-qa.test.ts:730-741` is a genuine red-before/green-after regression test for the claimed reason (pre-fix, `perLanguageMaxWer` returns the registry `0.45` at `state.source === 'default'`, so all four `toBeCloseTo(0.25)` assertions fail), and `:745-754` pins that a per-language override still outranks the global one. Env cleanup is sound — `SEG_ASR_MAX_WER` in each test's `finally`, `SEG_ASR_MAX_WER_ES` by the describe's existing `afterEach:690`, no leak into neighbouring cases.

**3. "Calibrated" sentinel inferred from two registry literals, false branch untested — ✅ RESOLVED, but by dissolution, not by the test that claims to cover it.** See nit 6 below. Both failure scenarios pass 1 named are now unreachable: scenario B (global default moved to 0.45) is moot because an explicit global override now wins over a calibrated default by design, and scenario A (a future `qa.asr.maxWer.it` calibrated to exactly the global default, silently inert) now produces a numerically identical result on both branches. Correct outcome; the accompanying tests are not what produces it.

**4. `docs/wiki/Advanced-Settings.md` default column stale — ✅ VERIFIED RESOLVED.** Lines 251-254 now read `0.45`.

**5. JSDoc wrong about zh/ja — ✅ VERIFIED RESOLVED.** `segment-asr-qa.ts:181-184` now correctly attributes zh/ja to the `!knob` guard, matching the test file's version.

**6. `registry.ts` overstated the p75/p80 bands — ✅ VERIFIED RESOLVED.** `registry.ts:304-308` now says uniform 0.45 bounded by the 0.5 floor, and states outright that es (0.20/0.29) and de (0.27/0.33) did not call for a raise. (See nit 9 — `d1.md` did not get the same correction.)

**7. `d1.md`'s `## Commit` placeholder — ✅ VERIFIED RESOLVED.** `6b80c45c` fills it with `dec2a43e` and cross-refs the `eaee67aa`/#3077 rework; #3077 exists and is closed, so the cross-reference resolves. The Verification section's "this file records the SHA below once committed" deferral now terminates somewhere real.

**8. Dangling `B2` cross-reference in the live view — 🟡 STILL OPEN.** See nit 10.

**9. Release-notes pair unmet — ✅ VERIFIED RESOLVED.** Both `docs/release-notes-next.md:241` (technical) and `RELEASE_NOTES.md:159` (user-facing) land, and both cover *each* user-visible half — the four `0.4 → 0.45` defaults **and** the precedence fix — rather than only the threshold change.

---

### 🔴 Correctness bugs

None. The two new commits land what they claim; nothing new introduced.

---

### 🧹 Cleanup nits

**6. LOW — `server/src/tts/segment-asr-qa.ts:170-172,195-198` + `segment-asr-qa.test.ts:757-774`: after fix 2, `isLanguageCalibrated` is a distinction without a difference, and its two new tests assert that `!==` is `!==`.**

By the time control reaches line 195, `resolveKnob(globalKnob).source === 'default'` is already established (line 194), so `configValue('qa.asr.maxWer')` — which is exactly `resolveKnob(globalKnob).effective` (`resolver.ts:176-180`) — is `globalKnob.default`. The `false` branch therefore returns `undefined`, and `resolveAsrThresholds:206`'s `?? configValue('qa.asr.maxWer')` supplies `globalKnob.default` — which in that branch *equals* `knob.default`, the value the `true` branch would have returned. Deleting lines 195-198 and returning `knob.default` unconditionally is behaviour-identical in every reachable state. Concretely: add `qa.asr.maxWer.it` with `default: 0.4` today and the effective Italian cap is 0.4 with the check and 0.4 without it, both with and without a global override set.

The consequence for the tests: `isLanguageCalibrated(0.45, 0.4) === true` / `(0.4, 0.4) === false` are assertions about the `!==` operator. They stay green against any implementation of the surrounding function, including one with the branch deleted — shape 3, "a test that cannot fail". They do not cover the hazard pass 1 named; they document a decision that no longer has a consequence. The JSDoc at `:164-169` compounds it by asserting the `false` case is a live risk ("would otherwise reach this branch for the first time with zero prior coverage") that fix 2 has already removed. Simplest honest resolution: drop the helper, the branch and the two tests, and let the JSDoc say the per-language default applies whenever the global knob is untouched — one rule instead of two, and no test claiming coverage it does not have.

**7. LOW — `docs/features/248-srv-asr-qa-non-english-normalization.md:77-80`: the run sheet's "Invariants to preserve" now states a fact this PR made false.** It reads "`qa.asr.maxWer.{fr,de}` (this plan) and the pre-existing `.es`/`.ru` all default to the global `qa.asr.maxWer` value (`0.4`)". All four now default to `0.45`. The invariant's *condition* is satisfied (the clause permits a different value "with real on-box calibration evidence behind it", which D1 supplies) — it is the parenthetical statement of current state that is now wrong, on the per-feature run sheet a reader consults to learn what the knobs do. Same shape at `docs/testing/onbox-sitting-multilanguage.md:115` ("every one still `default: 0.4`"), though that one reads as a dated observation inside a sitting plan and is defensible as-is. Shape 9.

**8. LOW — the PR body no longer describes the PR.** The Summary still frames the branch as a register fold that "touches `server/src/config/registry.ts` for D1's threshold defaults ... so this is not a docs-only PR". Since `eaee67aa` it also changes `perLanguageMaxWer()`'s precedence — live behaviour for four languages — and carries the release-notes pair. Failure scenario: whoever merges reads the body, sees "batch step 7 fold", and does not learn that this PR changes what the global ASR max WER slider does for es/fr/de/ru. The house gate asks for incidental findings fixed in-round to be *declared in the PR body*; they were fixed (good) but not declared.

**9. LOW — `docs/testing/onbox-batch-results/d1.md:202-206` and `server/src/config/registry.ts:304-308` now characterize the same numbers differently.** The registry comment was corrected (pass-1 finding 6) to "es (0.20/0.29) and de (0.27/0.33) both sit well under it ... es/de's own distributions did not call for any raise"; the evidence file it cites still says 0.45 "sits within or just above each language's own p75–p80 band" and "absorbs a real, data-backed slice" for all four. Pass 1 read d1.md's phrasing as the honest one, so this is not a new overstatement — but the correction landed on only one of the two surfaces, and the evidence file is the one a future recalibration starts from. One sentence in d1.md would close it.

**10. LOW (carry-over, pre-existing) — `docs/testing/onbox-acceptance-register-live-view.html:540`: dangling `B2` cross-reference, still open.** "A22 and B2 each prove their mechanism against a single already-drifted chapter or book" — no `B2` row exists after this PR's fold. Confirmed present only in the live view; the register `.md` has no mirrored copy, so pass 1's "same for the register" half does not apply. Not introduced here and cleanup-only, so it does not gate — but a discharge fold is the round that should sweep it.

---

### Delta verification performed

- Read `git show eaee67aa` and `git show 6b80c45c` in full; every claim in `eaee67aa`'s message is backed by its own diff (no claimed change missing, no unclaimed change smuggled in).
- `.env.example` help/default lines diffed against `registry.ts`'s four knob definitions — exact match.
- New precedence traced through `resolveKnob`/`resolveKnobInner` (`server/src/config/resolver.ts:40-59,87-89`) and `configValue` (`:176-180`) rather than assumed.
- Re-run independently at this head by the dispatching session: `npm run config:check` ✅, `npm run check:onbox-register` ✅, `tsc --noEmit` (root + server) ✅, scoped `segment-asr-qa.test.ts` 131/131 ✅, full `npm --prefix server run test` 536 files / 8145 tests exit 0 ✅.
- Per-finding severity, `file:line` and a concrete failure scenario are attached above. No finding in this pass is a correctness bug, so per `findings-triage.md` this pass does not re-trigger the gate.

dudarenok-maker:
## PR review — pass 3 (head 9111abb6, depth high)

Final round of the 3-pass cap. Depth `high` re-derived from the current commit set (multi-scope: `fix(server)` + `docs`/`test`) — unchanged from passes 1-2. Deltas only; the earlier passes' "checked and clean" lists are not re-litigated.

Scope: the two commits added since pass 2 — `c238415b` (the `origin/main` merge and its register conflict resolution) and `9111abb6` (the citation/link fixes) — plus an independent re-derivation of the register arithmetic from the row bodies, and a **disposition check on every pass-2 finding**.

**Headline: pass 2's entire 🟡 list is still open. All five, untouched.** The PR was staged for this round as though they were closed; they are not. Plus 2 new significant findings and 4 new cleanup findings from the merge resolution.

---

**6. `isLanguageCalibrated` is a distinction without a difference; its two tests assert that `!==` is `!==` — ❌ STILL OPEN, byte-for-byte untouched.**
`git diff 6b80c45c 9111abb6 -- server/src/tts/segment-asr-qa.ts server/src/tts/segment-asr-qa.test.ts` returns **empty**. The helper is still exported at `segment-asr-qa.ts:170` and still called at `:195`; the two tests are still at `segment-asr-qa.test.ts:757-772`. No new coverage, no deletion, no JSDoc correction.

**7. `docs/features/248-srv-asr-qa-non-english-normalization.md:77-80` states a fact this PR made false — ❌ STILL OPEN, untouched.**
Still reads "`qa.asr.maxWer.{fr,de}` (this plan) and the pre-existing `.es`/`.ru` all default to the global `qa.asr.maxWer` value (`0.4`)". All four ship at `0.45` (`registry.ts`).

**8. The PR body doesn't describe the PR — ❌ STILL OPEN.**
The Summary gained three paragraphs about the merge, but still frames the server touch as "`server/src/config/registry.ts` for D1's threshold defaults alongside the docs changes". The `perLanguageMaxWer()` precedence change — live behaviour for es/fr/de/ru, and the thing a merger most needs to know — is still undeclared, as is the release-notes pair.

**9. `d1.md` and `registry.ts` still characterise the same numbers differently — ❌ STILL OPEN, untouched.**
`d1.md:203-205`: "0.45 sits within or just above each language's own p75–p80 band (de 0.27–0.33, es 0.20–0.29, fr 0.40–0.40, ru 0.33–0.50), so it absorbs a real, data-backed slice…"
`registry.ts:305-308` (corrected in pass 1): "…**not** each language's own p75-p80 band: es (0.20/0.29) and de (0.27/0.33) both sit well under it… es/de's own distributions did not call for any raise."
Directly contradictory, on the two surfaces a future recalibration reads. The evidence file is the one it starts from.

**10. Dangling `B2` cross-reference in the live view — ❌ STILL OPEN, and now strictly worse.**
`onbox-acceptance-register-live-view.html:544` — "A22 and B2 each prove their mechanism against a single already-drifted chapter or book." When pass 1 flagged this, the id `B2` at least still resolved to *a* live row. This PR discharges B2, so the sentence now names a row that exists on neither surface.

Per [`findings-triage.md`](https://github.com/dudarenok-maker/Castwright/blob/main/.claude/skills/pr-review-gate/references/findings-triage.md): 🟡-only findings do not re-trigger a round, but they **are** fixed in the round that surfaced them — "This is a *fixing* rule, not a re-review rule". Five were surfaced; none were fixed.

---

### 🟠 Significant — new

**A. `docs/testing/onbox-acceptance-register.md:2214-2235` — the A16 conflict resolution deleted `main`'s introducing sentence, orphaning the evidence table and dangling two back-references to it.**

`main`'s A16 carried, immediately above the table:

> **What still stands — the fix works, proven at the synthesis boundary.** Three direct `POST /synthesize` calls on the same cloned voice, raw PCM transcribed with Whisper auto-detect and embedded with `/embed`:

`c238415b` replaced those three lines with this branch's chapter-level paragraph (`:2214-2228`) and left the table at `:2229-2233` in place. Consequences, all present at this head:

1. **The table has no antecedent.** `:2229` now follows a paragraph about a *full-chapter batch render* of a *Russian* character (Одуван, cosine 0.800), while the table's three rows are *direct `/synthesize` probes* in *English/German* (0.865 / 0.809 / 0.876). Nothing on the page any longer says what "Call" means or how those numbers were produced.
2. **`:2235` dangles.** "Row 3 reproduces the shipped bug live — German in, English phonetics out" now trails a paragraph that never mentioned three rows, a pre-fix condition, or German.
3. **`:2252` is a reference with no referent.** "…seen in the **synthesis-boundary table above** (0.809–0.865)". `grep -n "synthesis.boundary"` over the file at this head returns exactly two hits — `:2244` and `:2252` — **both of them back-references**. The phrase's only definition was in the deleted sentence.
4. **A claim `main` asserted is gone from the register.** "the fix works, proven at the synthesis boundary" is not restated anywhere.

Failure scenario: an auditor asks what proves the #1951 fix at the synthesis boundary, greps `synthesis-boundary`, lands on `:2252`, follows "the table above", and finds an unlabelled table sitting under a paragraph about a different render, in a different language, on a different call path. They cannot tell whether the 0.809 figure is this branch's chapter evidence or the older probe.

This is also the one place the PR body's losslessness claim doesn't hold. "The merge itself was verified lossless" is true as stated (it scopes itself to row-ID sets, and that check does pass — see ✅ below), but prose from `origin/main` *was* dropped, and a row-ID diff is structurally blind to it.

**B. `docs/testing/language-recurrence-onbox-acceptance.md` — B2's per-feature run sheet is annotated as discharged but records no run, on a PR that discharges B2.**

The removed B2 row named this file as its criteria source ("*Criteria:* [`language-recurrence-onbox-acceptance.md`](…) §Analysis language gate"), so it is the per-feature run sheet for CLAUDE.md's [before-shipping step 3](https://github.com/dudarenok-maker/Castwright/blob/main/CLAUDE.md#before-shipping-checklist): a PR that discharges owed acceptance "must leave the acceptance state recorded across all three surfaces, in the same diff — … the per-feature run sheet …: the criteria, and **the filled-in `Result:` lines once run**".

At this head, after `9111abb6` edited it:
- §3's four `Result:` lines (`:63`, `:69`, `:73`, `:84`) are **all still blank** — `Result: _______________________________________________`.
- `:118` — `- [ ] §3 run (register row B2, discharged 2026-09-07, removed from the register)` — the checkbox is **unchecked** while the annotation on the same line asserts the discharge.
- §2's `SHA: ____  Clean tree: ☐  Date: ____  Run by: ____` (`:51`) is blank.
- The file **never links `onbox-batch-results/b2.md`**, where the run that discharged B2 actually lives.
- `:122` — the sheet's own closing instruction — reads "Record what was observed, by whom, and when — **here** and in the register rows."

So the surface a reader is sent to for B2's criteria tells them the run has not happened, and offers no route to the evidence that says it did. Recording is the half of the on-box gate that blocks a merge; here it landed on two surfaces of three.

Compounding it, `9111abb6` introduced a self-contradiction in the header block: `:3` still says "**This is a working document.** Fill in the `Result:` lines AS you run this on the box. Do not pre-fill them," and `:10`, seven lines later, now says "this whole document is now historical record." §4 is A43's section and carries six of its own unfilled `Result:` lines (`:96`-`:114`), so "historical record" is not accurate for that half either.

---

### 🟡 Minor — new

**C. `onbox-acceptance-register.md:2199` — A16's headline still says "one new defect" while the row now carries two, and the two tracked surfaces disagree about the count.**
The heading note reads "still owed, **one new defect**". The body now carries two independent open findings: the sentence-19 English leak (`:2262-2266`, filed #3079) and this branch's 1-of-8 `voice-mismatch` flag (`:2272-2290`). The live view's `<summary>` badge *was* updated — `…-live-view.html:412` reads "still owed, **2 open findings**" — but its own flag body (`:414`) still says "one new defect", same as the register. Failure scenario: a reader scanning register row headings for what A16 still needs stops after #3079 and never reaches the flagged line awaiting a human listen — which is the second, independent blocker.

**D. `onbox-acceptance-register.md:564-573` — the change-log entry cites an evidence file for D1 only; B2's and A37's discharges leave the register with no pointer at all.**
The entry links `onbox-batch-results/d1.md`, and even that is cited for "why 0.45 and not each language's own p90" rather than as the discharge evidence. B2 and A37 get parenthetical summaries and nothing else. Row IDs are allocate-once and the rows are gone permanently, so this entry is the only register-side trace they will ever have. The register's own convention, stated in the prior entry at `:595-596`, is "citing the evidence file rather than keeping the row per the 2026-08-22 discharge ruling". `onbox-batch-results/b2.md` and `a37.md` both exist and are committed here — they are simply not linked.

**E. `docs/testing/onbox-batch-results/d1.md:3-6` — unmatched closing parenthesis, introduced by `9111abb6`.**
The edit removed the opening `(` from "Register row D1 (`…`, issues" but left the closing one:
```
Register row D1 — discharged and removed 2026-09-07 (this batch's own
evidence below is what discharged it) — issues
[#1527](…),
[#1084](…)). Step 5 of
```
Renders as "issues #1527, #1084**)**. Step 5 of…".

**F. `docs/testing/cast-id-drift-onbox-acceptance.md:271,284` — this is the one of the nine annotations that does not read true.**
`9111abb6` moved "discharged" inside the 120-char `ID_PROXIMITY_CHARS` window so Check A would pass. The window is genuinely satisfied and the sentence's meaning is unchanged — but the discharge it now advertises is **not the one that made the citation dangle**. The `B2` cited here is the *earlier* row of that id: `onbox-wave5-results/step-4-b2.md`'s own header reads "register row B2 (**stage-1 cast names in the manuscript's own script**) … Register row: B2 ([#2313], PR #2317)", and the register's change-log at `:1176` records "#2606 step 6, 69 → 66 (A27, A45, **B2** discharged)". The row this PR removes is #2246's analysis language gate. Two different rows have carried the id `B2`; nothing on either surface discloses it.
Failure scenario: someone auditing "when was B2 discharged, and on what evidence" gets 2026-08-23 / `onbox-wave5-results/step-4-b2.md` from `:271` and 2026-09-07 / `onbox-batch-results/b2.md` from the register change-log, with no way to tell they describe different rows. Before this PR the citation read as pointing at a live row; the annotation added here makes it read as a settled current-numbering statement instead. One clause naming which B2 closes it.
(The underlying id reuse is historical and not this PR's to fix — but this PR is the round that made the citation dangle and the round that annotated it.)

**G. `docs/testing/onbox-sitting-multilanguage.md:111` — "Criteria source: `onbox-acceptance-register.md` D1 (`:2854-2867`)" now points into unrelated prose, under a heading this PR just annotated as discharged.**
Register lines 2854-2867 at this head sit inside A23's Wave-3 `collectSegmentOrphans` narrative. (Pre-existing drift — D1 was at `:4572` on `main` too — but this PR is the discharge sweep that removes the row the pointer targets, and it edited `:109`, the heading directly above.) `:116` in the same block still asserts "`qa.asr.maxWer.{es,ru,fr,de}`, every one still `default: 0.4` — identical to the global default, confirmed unchanged", which this PR made false; pass 2 read that as a defensible dated observation, which it no longer is now that the line directly above declares the row discharged by this very PR.

---

### ✅ Verified independently this pass

- **Row-ID arithmetic is exact.** True merge-base computed as `git merge-base 6b80c45c 7c100c43` = `76aa7c6d` (not the PR's stale `baseRefOid`). Row-heading sets extracted from the file bodies at all four points: merge-base 65 → this branch pre-merge 62 (−A37, B2, D1) → `main` at merge 55 (−A5, A29, A36, C4, D3, E4, E6, E8, E10, E12, E101, +E103) → merged head **52**. The head's set is *exactly* `(merge-base) − (main's 11) − (this branch's 3) + E103`, with **no other row added or dropped** in either direction.
- **Glance table and change-log numbers are right, recomputed from the row bodies:** A 35, B 1, C 3, D 1, E 8, G 2, H 2 = **52 owed**. The entry's "55 → 52", "Group A 36 → 35", "B 2 → 1", "D 2 → 1" all check out against `main`'s state at `7c100c43`. `main`'s change-log history is preserved below, correctly re-labelled "Prior change".
- **"Batch step 6" in the change-log is correct, not a typo for the PR title's "step 7"** — #2964 is the fold step and #2956 is the verify/open-PR step. Checked against the chain's own tickets rather than assumed.
- **A36 and A37 are both fully absent as row headings**, on both surfaces (`A36` appears 0 times in the live view; only historical change-log mentions remain in the register). A36 was not resurrected by the merge; A37's evidence survives at `onbox-batch-results/a37.md`, correctly annotated.
- **A37's discharge is within its own bar, not a by-ear row self-discharged.** I checked #2976 before flagging it: the owner-written ticket says "The row's title says 'by ear' but its own criterion is a pause/duration comparison — **measure it, don't skip to a human**", and sets the bar as "does the pause survive the comma collapse". It does (0.194 s vs a 0.05–0.08 s no-punctuation band). The 2.5× shortening (0.495 → 0.194 s) is recorded honestly in `a37.md`.
- **A16's two findings are both present and genuinely complementary**, not competing: `expectedEngine`/`renderedEngine` both resolve to `qwen` in the clean run, so the #1969 stale-reference mode is excluded as the cause of the voice-mismatch flag, and neither finding's evidence contradicts the other. The stale `render-integrity.json`/`embeddings.json` trap is disclosed rather than buried. Only the headline count (finding C) is wrong.
- **`9111abb6`'s annotations are inside the real window, not gamed.** Read `ID_PROXIMITY_CHARS = 120` at `scripts/check-register-citations.mjs:746` and `DISCHARGE_ANNOTATION_REGEX` at `:698`; all nine sit within it, and eight of the nine also read true (finding F is the exception).
- **`a16.md`'s link now resolves.** `docs/testing/onbox-batch-results/` + `../../features/275-clone-voice-language.md` → a real file.
- **No file from `main` was dropped by the merge** (`git diff --diff-filter=D 7c100c43 9111abb6` is empty), and the non-docs surface vs `main` is exactly the five D1-threshold files — no frontend or server change of `main`'s was reverted.
- **No drift in the pass-2-reviewed server surfaces:** `registry.ts`, `segment-asr-qa.ts`, `segment-asr-qa.test.ts`, `.env.example` and `docs/wiki/Advanced-Settings.md` are byte-identical between `6b80c45c` and this head.
- Commands run at this head: `check:onbox-register` ✅ · `build-register-live-view.mjs --check` → "up to date" (idempotent) ✅ · `check-register-citations.mjs` exit 0 ✅ · `check-register-row-citations.mjs` exit 0 (57 citations resolve, 21 annotated-discharged) ✅ · `node --test scripts/tests/check-register-citations.test.mjs` **143/143** ✅ · `node --test scripts/tests/review-gate-mechanism.test.mjs` **29/29** ✅ · `tsc --noEmit` root ✅ and server ✅.

---

### Verdict

**Not mergeable as staged.** The merge itself is sound where it matters most — the arithmetic is exact, no row and no file was silently lost, and the A16/A36/A37 conflict resolutions all reach the right end state. The two new 🟠 findings are both in the *editorial* half of that merge (a deleted antecedent in A16, an unfilled run sheet for B2), and the 🟡s are small.

What blocks is the disposition table: **all five of pass 2's cleanup findings are still open, four of them byte-for-byte untouched.** Under the report-fix-record rule they were owed in that round, and "it's only a chore" is a named-void deferral. Eleven findings now stand open on this PR — five carried, six new.

This is round 3 of 3, so the loop cap is reached and the gate stops here rather than auto-looping. Handing back to the shipping session: fix all eleven, then merge on the strength of this pass — a further round is not available, and none of the eleven is a correctness bug in shipped behaviour that would warrant asking for the cap to be lifted.

Issue #1263, opened 2026-07-03
Non-English chapters offer "Render anyway" in the fallback-confirm queue gate, then deterministically fail


dudarenok-maker:
Found during adversarial review of the fe-46 plan (docs/features/240); pre-existing, NOT introduced by it.

**Repro:** non-English book with an undesigned Qwen speaker → enqueue a chapter → it parks at `awaiting_fallback_confirm` (`computeQwenKokoroFallbackSet` doesn't check language, `server/src/tts/qwen-fallback-set.ts`) → queue modal offers "Render anyway" (`src/modals/queue-modal.tsx` ~551-565, no language branch) → confirming re-dispatches with `fallbackConfirmed:true`, but `synthesiseChapter` runs with `forbidKokoroFallback` for non-English (`server/src/routes/generation.ts` ~1283) → the chapter still throws `MissingDesignedVoiceError` and fails.

**Expected:** a non-English chapter with undesigned speakers should never offer "Render anyway" — the gate should present design-the-voices as the only forward path (mirroring the fe-46 pre-flight gate's hard block), or fail fast with that message instead of a confirm→fail loop.

fe-46's pre-flight gate (#1262) reduces how often users reach this state but deliberately does NOT fix this per-chapter path — server semantics were out of its scope.

PR #3083, opened 2026-09-07
refactor(server): remove the unreachable !coquiEligible generation arm

### LINKED ISSUE #3059: The generation gate's !coquiEligible arm is unreachable and untested — keep it, guard it, or remove it
## What

`server/src/routes/generation.ts`'s per-chapter voice-readiness gate now reads:

```ts
if (nonEnglishBook && !coquiEligible) { /* hard-fail, naming every undesigned character */ }
```

PR #3031 narrowed it from bare `nonEnglishBook` — that was the fix, and it is correct. But review pass 1 of that PR established two things about what is left:

1. **The surviving `!coquiEligible` arm is unreachable.** `ENGINE_LANGUAGE_SUPPORT.coqui` is *exactly* the language registry's code set, and a language outside the registry throws at `generation.ts:851` before this gate is evaluated. So `nonEnglishBook && !coquiEligible` is provably empty on the current build.
2. **Its test coverage was deleted without replacement.** Four tests were removed in the same change; three of them covered logic that still exists — the `state.json` persist, the plural wording, and the narrator-note suppression. Neutralising the whole 58-line branch with `if (false && …)` leaves the file **8/8 green**.

So the branch is currently both unguarded and unreachable.

## Why this is filed rather than fixed

There is more than one defensible outcome and they differ in what they assert about the system, not merely in how they are written:

1. **Keep it as a fail-safe, and restore the coverage.** Re-add tests for the three still-live behaviours, driving the branch directly. Says: the registry/support-set coupling is not guaranteed forever, and a book that slips through should fail loudly and legibly rather than reaching `synthesiseChapter`. Costs: tests that pin a path production cannot currently take.
2. **Remove the branch entirely.** It is dead code by construction; delete it, and let `generation.ts:851`'s throw be the single place an unsupported language is refused. Says: the coupling IS the invariant, and a second refusal site is duplication that can drift. Costs: if `ENGINE_LANGUAGE_SUPPORT.coqui` ever diverges from the registry, the friendly all-characters-named failure is gone and users get the raw throw.
3. **Convert it into an explicit invariant check.** Keep the code path but make it assert-and-report (a named internal error, logged, surfaced through the existing failure taxonomy) rather than pretend to be ordinary user-facing behaviour. Says: this is a "cannot happen" that we want told about if it ever does. Costs: a third failure shape on a route that already has several.

**The decision owed:** whether the registry ↔ `ENGINE_LANGUAGE_SUPPORT.coqui` equality is an invariant we rely on (→ 2), a coincidence we defend against (→ 1), or a condition we want alarmed (→ 3).

Note the second-order effect on documentation: PR #3031's release note originally described the `!coquiEligible` arm as live user-facing behaviour. That was corrected to describe it as a retained fail-safe pending this decision — whichever option wins, that sentence needs a final pass.

## Key files

- `server/src/routes/generation.ts` — the gate, and `:851`'s earlier throw
- `server/src/routes/generation-fallback-gate.test.ts` — where the removed coverage lived
- `server/src/tts/voice-mapping.ts` — `ENGINE_LANGUAGE_SUPPORT`
- `server/src/tts/language-registry.ts` — the registry whose code set it mirrors

## Benefit

*Benefit (technical):* an unreachable branch with no tests is indistinguishable from a bug that has not been noticed yet. Either it earns its place and gets coverage, or it goes — and settling which also settles whether the registry/support-set equality is something the codebase may rely on, which is reusable well beyond this one gate.

Found in review of PR #3031, which narrowed the gate. Not caused by that PR — the `!coquiEligible` arm's unreachability predates it; the PR is what made the arm visible as a distinct case.

dudarenok-maker:
## Summary

PR #3031 narrowed `generation.ts`'s per-chapter voice-readiness gate from a bare `nonEnglishBook` to `nonEnglishBook && !coquiEligible`. That was the fix, and it was correct. Its review then established two things about what survived: **the `!coquiEligible` arm is unreachable**, and **its test coverage had been deleted** — neutralising the whole 58-line branch left the file 8/8 green. #3059 filed the decision that leaves: is the coupling the arm's deadness rests on an invariant we rely on, a coincidence we defend against, or a condition we want alarmed?

**Answer: an invariant — and it becomes one by being enforced, not by being asserted in a comment.** This removes the arm and adds the guard that keeps the removal true.

This PR went through two review rounds. Round 1 found the removal itself correct but the guard incomplete (findings A–E below); round 2 found the round-1 fix for finding A itself made a false claim about its own coverage (finding F) and flagged an undeclared incidental fix (finding G). Both are folded in below rather than left as separate history — this section describes the guard as it stands at HEAD, not as first shipped.

## The evidence, re-verified on `main` at `8e4025ee`

- `ENGINE_LANGUAGE_SUPPORT.coqui` is `['en','ru','es','fr','de','zh','ja']`. The registry's codes are exactly `en, ru, es, fr, de, zh, ja`, all `supported: true`. Equal sets ⇒ `nonEnglishBook && !coquiEligible` is empty, and a language outside the registry throws earlier in the same route.
- `generation-fallback-gate.test.ts` has 8 tests; every non-English case is explicitly a *Coqui-eligible* one. Nothing drove the arm.
- **Nothing anywhere asserted the two sets agree.** `ENGINE_LANGUAGE_SUPPORT` has three other readers — `engine-recommendation.ts`, `language.ts`'s `resolveEligibleEngines`, and two tests — and not one is a check.

That last point is what decided it against simply keeping the arm. The registry's own header documents that every language landed `supported: false` and was flipped after on-box validation; the table is designed to grow one language at a time, and the equality was maintained by whoever remembered.

## Why not the other two options

- **Keep it and restore coverage** would pin a path production provably cannot take. That is coverage that cannot fail — the failure class this repo has catalogued at length.
- **Convert it to a runtime invariant check** puts the alarm in front of a user mid-render, for a condition a build-time check catches before merge, and adds a third failure shape to a route that already has several.

## The guard

`nonEnglishBook && !coquiEligible` is empty only if three conjuncts all hold: (1) `sidecarLanguageName` throws before the per-chapter gate for any language outside the registry; (2) every validated registry code is in `ENGINE_LANGUAGE_SUPPORT.coqui`; (3) `generation.ts:830`'s `resolveEligibleEngines(bookLanguage, ALL_TTS_ENGINES)` call still passes `ALL_TTS_ENGINES` and `'coqui' ∈ ALL_TTS_ENGINES`. `server/src/tts/engine-language-coverage.guard.test.ts` now has **four** assertions covering (2) and (3):

1. Every **validated** (`supported: true`) non-English registry code appears in `ENGINE_LANGUAGE_SUPPORT.coqui` — the table check.
2. `resolveEligibleEngines(code, ALL_TTS_ENGINES)` includes `'coqui'` for every such code — additional coverage beyond #1, since it also pins `'coqui' ∈ ALL_TTS_ENGINES` intentionally (previously only incidental coverage in `language.test.ts`).
3. `generation.ts`'s own `resolveEligibleEngines(...)` call site still passes `ALL_TTS_ENGINES` as its second argument — a deliberately syntactic source-text scan (same idiom as `server/src/workspace/cast-lock.guard.test.ts`), parsed with the TypeScript compiler API rather than a regex.

**Round-1 correction:** the first cut of assertion #2 was proposed (in review) and implemented as sufficient to also catch a change to what `generation.ts:830` passes as the installed-engines set — the exact change #3059's own "Explicitly out of scope" section parks. Round-2 review measured that this is false: assertion #2 imports `ALL_TTS_ENGINES` itself and never reads the call site, so swapping it for a real installed-on-this-box set left both assertions green while the removed arm was live for every non-English book. Assertion #3 exists specifically to close that — it is what actually watches the call site.

**Keyed on the validation flip, not on registration** — deliberately, and the reasoning is in the guard's header. A `supported: false` language is blocked at the confirm-screen support gate and cannot reach generation, so it cannot make the removed arm reachable. Firing at registration would force whoever registers an unvalidated `ko` to either restore a refusal path they do not yet need, or — far more likely — add `'ko'` to `ENGINE_LANGUAGE_SUPPORT.coqui` to get green, which would be a false claim about what XTTS can render. Firing at the flip puts the guard where the choice is an honest one.

**The header now names four limits**, not one:
- Conjunct 1 (the `sidecarLanguageName` throw ordering) has no coverage in this file at all — named so the guard's scope record doesn't imply more than it has.
- A `supported: false` registry entry is out of scope by design; if the confirm-screen gate is ever bypassed, this guard's scope is too narrow.
- A language *demoted* from `supported: true` to `false` after books were already imported under it is not caught — `sidecarLanguageName` → `getLanguageEntry` doesn't consult `supported`, and `requireBookStateLanguage` returns the stored language raw.
- Assertion #3's syntactic scan has its own blind spots: an aliased import of either `resolveEligibleEngines` or `ALL_TTS_ENGINES`, indirection through a differently-named variable that happens to hold the same value, and a second call site elsewhere in the tree are all invisible to it.

Every failure message names both real remedies and links #3059, so a future red run does not need this PR to be actionable.

## Round-3 correction: the guard asserted a spelling, not a binding

Assertion #3 checked that the call site's second argument is the identifier `ALL_TTS_ENGINES`. That means nothing if the **name** no longer refers to the real engine set. Measured fail-open:

```diff
-import { ALL_TTS_ENGINES } from '../tts/model-keys.js';
+import { ALL_TTS_ENGINES as ALL_ENGINES_FULL } from '../tts/model-keys.js';
+  const ALL_TTS_ENGINES = ALL_ENGINES_FULL.filter((e) => e !== 'coqui');
   const eligibleEngines = resolveEligibleEngines(bookLanguage, ALL_TTS_ENGINES);
```

Type-correct, lint-clean, call site byte-identical — and **3/3 green** while `coquiEligible` was false for all six validated non-English languages. That is the *minimal* diff for the very change #3059 parks, so it is the one spelling this guard could least afford to miss.

A fourth assertion now walks the AST for an aliased import of `ALL_TTS_ENGINES` and for any re-declaration of that name. Re-running the mutation reddens it, naming both the alias and the shadowing line.

The declared blind-spot list was also wrong: it had aliasing backwards (both aliases actually fail **closed**, measured) and omitted the fail-open case above. What remains genuinely unguarded is indirection through an equivalent value under another name, and a second call site in another file — this scan reads only `generation.ts`.

- **Two frontend comments described a deleted mechanism.** `src/store/queue-dispatcher-middleware.ts` and `src/store/generation-stream-runner.ts` both described the old `#1263` `voice-not-designed` fail-fast as live, and one described it as having "replaced" the `awaiting_confirm` park. After this PR, no server path emits `voice-not-designed` (confirmed: `grep -rn 'voice-not-designed' server/src` returns only the taxonomy union member, a remediation-table key, and a table-completeness test — no emitter, and no dynamic `errorCode:` construction anywhere in `generation.ts`). Both comments are corrected to say so.
- **The `voice-not-designed` taxonomy plumbing is deliberately retained, not deleted** — `failure-remediations.ts`'s entry, `IMMEDIATE_TOAST_ERROR_CODES`, `help-failures.ts`, the srv-11 streak-exemption check, and their tests. This is a decision, not an oversight: the guard's own header names a live scenario (assertion #3's own scope, and more broadly the call-site conjunct) in which the removed fail-fast would need restoring, so deleting its downstream machinery now would only have to be rebuilt.
- **A stale line-number citation** in `generation-fallback-gate.test.ts` (`~805` → `~854`, the actual `sidecarLanguageName` throw site) — the one line changed in that file; no test in it was added, removed, or rewritten.
- **The release note's over-claim about the guard's own coverage** — an earlier draft of the fs-60 bullet in `docs/release-notes-next.md` claimed the guard's second assertion covered the call-site conjunct. It didn't (see "Round-1 correction" above); the bullet now describes all three assertions and their limits accurately.

- **A guard that the scoped CI leg could not select.** Assertion #3 reads `generation.ts` with `readFileSync`, so there is no module-graph edge and `verify.yml`'s `vitest run --changed` server leg never chose it — silent on exactly the call-site change it exists to catch. Added the `forceRerunTriggers` entry and its `force-rerun-triggers.test.ts` counterpart, following the `synthesise-chapter.ts` precedent. Measured: a `generation.ts`-only diff went from selecting **31 files** to the whole fast tier, with the guard among them. The general case — broad-tree scanners no per-file trigger covers — is filed as **#3085** with three options and the decision named.
- **A stale hardcoded count.** `state-language.guard.test.ts` pins per-file `stateJsonPath(` site counts; the removed arm held one of `generation.ts`'s two, so `2` became `1` (45 sites, not 46) with its audit-trail comment updated. Caught by CI, and only by the Windows leg — the Linux leg is `--changed`-scoped and ran 31 files.

## Test plan

**The guard was watched fail before it was trusted, twice — once per new assertion.**

Dropping `'ru'` from `ENGINE_LANGUAGE_SUPPORT.coqui` (assertions #1/#2):

```
FAIL  src/tts/engine-language-coverage.guard.test.ts
AssertionError: language(s) ru are validated (supported:true) in the language registry
but missing from ENGINE_LANGUAGE_SUPPORT.coqui in server/src/tts/voice-mapping.ts.
Either Coqui/XTTS can actually render this language — add it to
ENGINE_LANGUAGE_SUPPORT.coqui — or it can't, which means a non-Coqui-eligible
non-English book is now genuinely reachable through generation.ts's voice-readiness
gate and the refusal path removed by #3059 needs to be restored for it.
See https://github.com/dudarenok-maker/Castwright/issues/3059.
  expected [ 'ru' ] to deeply equal []
```

Removing `'coqui'` from `ALL_TTS_ENGINES` (assertion #2 only — #1 stays green, proving the red is for the claimed reason):

```
AssertionError: language(s) ru, es, fr, de, zh, ja are validated (supported:true) but
resolveEligibleEngines(code, ALL_TTS_ENGINES) does not include 'coqui' for them — ...
  expected [ 'ru', 'es', 'fr', 'de', 'zh', 'ja' ] to deeply equal []
```

Retargeting `generation.ts:830`'s call site to a local `installedEngines` array instead of `ALL_TTS_ENGINES` (assertion #3 — the one that actually closes finding A/F):

```
AssertionError: server/src/routes/generation.ts calls resolveEligibleEngines(...) with a
second argument that is no longer the bare identifier ALL_TTS_ENGINES:
  line 831: resolveEligibleEngines(bookLanguage, installedEngines)
This is the call-site conjunct the arm removed by #3059 depends on ...
```

All three reverted; all three assertions confirmed green again; `voice-mapping.ts`, `model-keys.ts`, and `generation.ts` each confirmed byte-identical to their pre-mutation state via `git diff`.

- `src/tts/engine-language-coverage.guard.test.ts` — **4 passed**
- `src/workspace/state-language.guard.test.ts` and `src/force-rerun-triggers.test.ts` — green (86 passed across the four guard suites)
- `src/routes/generation-fallback-gate.test.ts` — **8 passed**; differs from `origin/main` by exactly the one line-number citation fix, no test added/removed/rewritten
- `src/tts/` — 125 files, 1963 passed
- `src/routes/generation.test.ts` via `vitest.config.slow.ts` — 50 passed
- `src/store/queue-dispatcher-middleware.test.ts` + `src/store/generation-stream-runner.test.ts` — 53 passed
- Server `npx tsc -p . --noEmit` and frontend `npx tsc -p tsconfig.json --noEmit` — both clean
- `npx eslint` on every touched source file (server + frontend) — clean

### Checklist notes

- **Regression plan** — not applicable. No behaviour changes: the removed code was unreachable, and the guard is a build-time assertion.
- **On-box acceptance** — not applicable; nothing here needs hardware to prove.
- **Release notes** — no new entry. The existing fs-60 entry in `docs/release-notes-next.md` described the arm as "retained as a fail-safe … see the follow-up issue on whether it should be kept, guarded or removed", which this PR makes false; it is corrected in place to say the arm was removed as unreachable and describe the guard's three assertions and their limits accurately. The stale `generation.ts:851` line-number citation went with it. There is no user-visible delta, because the arm was never reachable.

Closes #3059
Refs #3031


dudarenok-maker:
## PR review — pass 1 (head d7c94b9c, depth high)

**Scope:** all three changed files (`server/src/routes/generation.ts`, `server/src/tts/engine-language-coverage.guard.test.ts`, `docs/release-notes-next.md`), reviewed against #3059's "Decision: option 2, plus the guard that makes it safe" comment as the spec.

**Verified by execution in `wt-3059-coqui-arm` (read-only; HEAD `d7c94b9c`, tree clean before and after):** `npx tsc -p . --noEmit` exit 0 · `npx eslint src/routes/generation.ts src/tts/engine-language-coverage.guard.test.ts` exit 0 · guard test **1 passed** · `src/routes/generation-fallback-gate.test.ts` **8 passed** and confirmed **byte-identical to `origin/main`** (`git diff origin/main...HEAD --` on that path is empty) · `src/routes/generation.test.ts` under `vitest.config.slow.ts` **50 passed** · `src/tts` **125 files / 1961 passed**. Every numeric claim in the PR body reproduces.

---

### 🔴 Blocking — none

---

### 🟠 Correctness — 1

#### A. The guard covers one of the three facts the removal rests on, and the follow-up #3059 itself names would break a different one silently

`server/src/tts/engine-language-coverage.guard.test.ts:31-53`

Decompose what the removal actually needs. At `server/src/routes/generation.ts:830-831`:

```ts
const eligibleEngines = resolveEligibleEngines(bookLanguage, ALL_TTS_ENGINES);
const coquiEligible = eligibleEngines.includes('coqui');
```

and `server/src/tts/language.ts:60-70` expands that to

```
coquiEligible  ==  ALL_TTS_ENGINES.includes('coqui')
                && (ENGINE_LANGUAGE_SUPPORT.coqui === '*' || ENGINE_LANGUAGE_SUPPORT.coqui.includes(norm(lang)))
```

so `nonEnglishBook && !coquiEligible` is empty only if **all three** hold:

1. every reachable non-English `bookLanguage` is a registry code — enforced by `sidecarLanguageName`'s throw (`server/src/tts/language.ts:34-43`, called at `generation.ts:854` inside the `if (nonEnglishBook)` at `:832`, i.e. before the per-chapter loop at `:1382`);
2. every such registry code is in `ENGINE_LANGUAGE_SUPPORT.coqui`;
3. `'coqui' ∈ ALL_TTS_ENGINES` **as `generation.ts:830` calls it**.

The guard asserts **(2) only**. Its entire input surface is `allLanguageEntries()` and `ENGINE_LANGUAGE_SUPPORT.coqui` (imports at `:27-28`) — it never touches `resolveEligibleEngines`, `ALL_TTS_ENGINES`, or the call site.

That matters because #3059's own "Explicitly out of scope" section parks the obvious next change to this exact code: *"Whether the park-and-confirm path's 'Proceed anyway' can therefore promise a Coqui fallback that a box without Coqui weights cannot deliver is a separate question."* The natural resolution is to stop passing `ALL_TTS_ENGINES` at `generation.ts:830` and pass the installed set instead — `resolveEligibleEngines` already takes one, and `server/src/tts/language.test.ts:120-126` already pins `resolveEligibleEngines('ru', ['kokoro','gemini']) === []`. Concrete chain on a box without Coqui weights after that change:

- `coquiEligible === false` for every language → `nonEnglishBook && !coquiEligible` is true for **every** non-English book;
- the guard stays **green** — nothing it reads moved;
- the chapter now parks at `generation.ts:1452-1462` and the UI offers "Proceed anyway — generic Coqui fallback voices" (`src/modals/voice-readiness-gate.tsx:137`);
- confirming re-dispatches with `fallbackConfirmed`, skipping the gate, into `synthesiseChapter` with `forbidKokoroFallback: true, coquiEligible: false` (`generation.ts:1632-1633`). `applyQwenFallback` substitutes Coqui only when `forbidKokoroFallback && coquiEligible`, so it throws `MissingDesignedVoiceError` (`server/src/tts/synthesise-chapter.ts:1517`; pinned by `synthesise-chapter-coqui-fallback.test.ts:84`);
- that error matches **no** taxonomy signature (`server/src/routes/failure-taxonomy.ts` `matchName` entries are only `AnalyzerTruncatedError`, `DailyQuotaExhaustedError`, `RequestExceedsTpmError`, `GeminiStreamIdleError`, `GeminiContentBlockedError`, `UnresolvableClonedVoiceError`), so `classifyFailure` falls through to `code: 'unknown'` (`failure-taxonomy.ts:368-374`);
- `'unknown'` is **not** on the breaker-exemption list at `src/store/queue-dispatcher-middleware.ts:164`, so three such chapters trip the srv-11 consecutive-failure breaker and pause the **whole queue**.

So the failure mode is not "the deleted arm's message is missing" — it is the exact trap the deleted comment said the arm existed to avoid ("parking would just offer a button that deterministically re-fails"), **plus** a queue-wide pause that the deleted arm's `voice-not-designed` code was explicitly exempted from.

Fact (1) is likewise unguarded: nothing pins that `generation.ts` still calls `sidecarLanguageName` before the gate, or that it still throws rather than defaulting.

Honest bound on this finding: it is not reachable at `d7c94b9c` — see the ✅ section, I could not construct a live case — and fact (3) has *incidental* coverage today (`language.test.ts:98-102` would go red if `'coqui'` were deleted from the `ALL_TTS_ENGINES` array literal at `model-keys.ts:24`). Neither of those covers the call-site change above, and neither would produce the guard's carefully-written "the refusal path removed by #3059 needs to be restored" message.

Cheap fix that closes all three at once, because it exercises the same function the removal depends on instead of the table underneath it:

```ts
for (const e of allLanguageEntries().filter((e) => e.supported && e.code !== 'en')) {
  expect(resolveEligibleEngines(e.code, ALL_TTS_ENGINES), e.code).toContain('coqui');
}
```

---

### 🟡 Cleanup — 4

#### B. The guard header names one limit; it has a second

`server/src/tts/engine-language-coverage.guard.test.ts:11-24`

The header states the guard "rests on the confirm-screen `supported` gate actually holding", and justifies keying on `supported: true` with *"A `supported: false` language (registered but not yet validated) is blocked at the confirm-screen support gate and can never reach generation."* The parenthetical silently assumes the only way to be `supported:false` is *not yet validated*. A **demotion** breaks it: demote `ja` to `supported:false` (fs-59's own registry note flags CJK attribution as analyzer-model-dependent) and drop `'ja'` from `ENGINE_LANGUAGE_SUPPORT.coqui`. The guard filters `supported:false` out, so it stays green. But a book already on disk with `language: 'ja'` still reaches the gate: `sidecarLanguageName` resolves through `getLanguageEntry` (`language.ts:34-43`), which **does not consult `supported`**, so it does not throw — and `requireBookStateLanguage` (`server/src/workspace/scan.ts:375`) returns `state.language` raw without re-validating. `nonEnglishBook && !coquiEligible` is then live with the arm gone. Worth one sentence in the header, since the header is presented as the record of the guard's scope.

#### C. `docs/release-notes-next.md:194` now contradicts itself, and overstates the guard

Two problems in the one bullet:

- The unchanged sentence *"The condition is now `nonEnglishBook && !coquiEligible`, which is the split this same file already uses twice…"* is **false as of this PR** for the gate it describes — that gate is now unconditional. The same paragraph then says the arm "was removed by #3059". #3059's scope item 3 called out the *"retained as a fail-safe"* sentence; this one describes the condition as current and was missed. (The `:938` / `:952` line citations inside it are still accurate — the removal is below both.)
- The new text says the guard *"fails the moment the registry and `ENGINE_LANGUAGE_SUPPORT.coqui` stop agreeing."* It does not. It ignores `supported:false` entries **by design** (the header's central argument) and never fires on a coqui-side-only addition. The PR body and the guard header are both precise about this; the release note — the durable technical register — is the one place it isn't, which is the "asserted in prose" failure mode this PR set out to end.

#### D. Two comments the diff made false, and the last textual producer of `voice-not-designed` is gone

`git grep 'voice-not-designed' origin/main -- server/src` shows `generation.ts:1471` was the **sole** server producer, and no taxonomy signature yields the code (see A). After this diff **no server path can emit `errorCode: 'voice-not-designed'`**. Two comments now describe a mechanism that no longer exists:

- `src/store/queue-dispatcher-middleware.ts:159-163` — *"voice-not-designed is exempt: it's a deterministic per-book cast-config issue (the #1263 fail-fast that replaced the awaiting_confirm park, which was ALSO exempt from this streak)"*. The fail-fast is gone and the park was not replaced; the exemption at `:164` now has no producer.
- `src/store/generation-stream-runner.ts:422-424` — *"the old chapter_awaiting_fallback_confirm park voice-not-designed replaces (#1263)"*. Same.

Left downstream with no producer: `FAILURE_REMEDIATIONS['voice-not-designed']` (`server/src/routes/failure-remediations.ts:174`), `IMMEDIATE_TOAST_ERROR_CODES` (`generation-stream-runner.ts:74`), `src/data/help-failures.ts:46,75`, and two frontend tests that drive the code end-to-end (`src/store/generation-stream-runner.test.ts:284`, `src/store/queue-dispatcher-middleware.test.ts:523`) — i.e. coverage of a state the product can no longer produce, the same class the PR body invokes to justify the deletion.

Fair caveat: the producer was *itself* unreachable before this PR, so nothing became dead at runtime — what changed is that the last textual producer went away and the two comments above became false. Under CLAUDE.md's incidental-findings rule ("a stale comment or doc sentence your change made false", "dead or duplicated code the work exposed") those are chores owed in this round; none is declared in the PR body.

#### E. The identical dormant condition one screen up is left, with the repo's opposite position still on record

`server/src/routes/generation.ts:938` — `if (qwenUnavailable && nonEnglishBook && !coquiEligible)` — is dead by the *identical* proof this PR relies on, and is left in place. `server/src/routes/generation-fallback-gate.test.ts:444-457` (unchanged) still records the repo's position on it: *"That route-level fatal branch's LOGIC is unchanged, but it now has NO direct test… The route branch stays dormant until a non-Coqui registered language exists (fs-70)."* The PR reverses that disposition for one site and not the other, with no stated distinction, so a reader now finds two identical dormant conditions treated oppositely. #3059's Scope names only the `:1452` arm, so this is a scoping observation, not a spec breach — but it deserves a sentence somewhere.

Micro-nit in the same block: `generation-fallback-gate.test.ts:452-453` cites the `sidecarLanguageName` throw as *"line ~805"*; it is at `generation.ts:854` (`:805` is `requireBookStateLanguage`). Pre-existing drift, but this PR is load-bearing on that exact throw and its body advertises retiring a stale line citation.

---

### ✅ What I attacked and could not break

**Unreachability holds at `d7c94b9c`. I could not construct a reachable `nonEnglishBook && !coquiEligible` state.** What I tried:

1. **Odd normalisation / region subtags / case / whitespace** — `isNonEnglish`, `resolveEligibleEngines` and `sidecarLanguageName` all route through the *same* `normaliseBookLanguage` (`language.ts:16-25`), so `'ru-RU'`, `'RU'`, `' ru'`, `'zh-Hans'` cannot make the three disagree.
2. **A language outside the registry (`ko`)** — `sidecarLanguageName` throws at `generation.ts:854`, inside the `if (nonEnglishBook)` at `:832`, and the handler `return res.end()`s. The gate is at `:1452` inside `processOneChapter`, defined at `:1382` in the *same* handler (`generationRouter.post` spans `:716-2324`). There is no second entry point.
3. **A `supported:false` entry reaching generation** — none exists (registry codes are exactly `en, ru, es, fr, de, zh, ja`, all `supported: true`), and both write paths validate: `import.ts:360` and `book-state.ts:814` call `isSupportedLanguage`; the auto-detect path only writes a language when `detectedLanguageSupported && !fallback` (`import.ts:349-356`), with `supported` read off the registry (`detect-language.ts`'s `resultFor`). I enumerated all 20 `writeStateJsonAtomic` call sites — every other one passes `state.language` through unchanged.
4. **`ALL_TTS_ENGINES` missing `'coqui'`** — static const at `model-keys.ts:24`, read only via `.filter`; no runtime mutation anywhere. (See finding A for the *maintenance-time* version of this.)
5. **`ENGINE_LANGUAGE_SUPPORT` mutated at runtime** — no writer exists; all references are reads.
6. **`bookLanguage` re-derived or `state` re-read between derivation and gate** — the only assignment is at `:805`; `nonEnglishBook`/`eligibleEngines`/`coquiEligible` are `const` at `:829-831`. No re-read.

**The rewritten comment's factual claim checks out.** *"every book language this app will let a user pick is validated (registry `supported:true`)"* — verified against the only two write surfaces (`import.ts:334-368`, `book-state.ts:800-822`), both `isSupportedLanguage`-gated, and the confirm selector is fed from `supportedLanguages()` (`import.ts:155`). Accurate as scoped ("let a user pick"); the on-disk case is finding B.

**The `coquiLanguages === '*'` early return is sound, not a hole.** If Coqui were modelled as universal, `resolveEligibleEngines` returns coqui for every language, `coquiEligible` is always true, and the removed arm stays unreachable a fortiori — so there genuinely is nothing left to check. The branch is also required for typing (`Record<TtsEngine, string[] | '*'>`).

**The guard's pasted red output is authentic** — it renders the source message at `:44-51` verbatim, including the `expected [ 'ru' ] to deeply equal []` tail.

**On "8 passed, unchanged" as evidence:** it is a valid one-directional claim as the PR words it, and no more. #3031's review already established the file is insensitive to the removed arm (neutralising the whole 58-line branch left it 8/8 green), so 8/8 says nothing about the arm. What it *does* prove is that the surviving behaviour is covered: `:397` pins the park for a Coqui-eligible Russian book, `:234` pins confirm-then-render, `:246` pins the no-`queueEntryId` bypass — i.e. the fall-through path the removal makes unconditional is exercised.

**House gates:** `Closes #3059` present and outside any code span · no `cast.json` writes touched · no config knob · no derived artifact staled · regression-plan / on-box-acceptance / new-release-note not-applicables are each stated explicitly with a reason · release-notes edit-in-place is the correct call, subject to finding C.

---

### Verdict

**One correctness finding (A) and four cleanup findings.** Nothing here says the deletion was wrong — the unreachability proof holds at this head and I could not falsify it. What A says is that the PR's own thesis — *"it becomes an invariant by being enforced, not by being asserted in a comment"* — is only two-thirds delivered: the guard enforces the table coupling and leaves the `ALL_TTS_ENGINES`/call-site conjunct asserted in exactly the prose the PR set out to replace, with the most likely next change to this code (the follow-up #3059 itself parks) landing precisely in the gap. A is a ~4-line addition to the guard.

dudarenok-maker:
## PR review — pass 2 (head 9f542097, depth high)

Re-review of the fix commit only. Pass 1's core conclusion — the removal is correct and `nonEnglishBook && !coquiEligible` is unreachable at this head — was re-checked for disturbance and holds: `9f542097` touches no production code path (`generation.ts` is byte-identical between `d7c94b9c` and `9f542097`), and `model-keys.ts`, `voice-mapping.ts`, `language.ts` are each **byte-identical to `origin/main`** (`git diff origin/main...HEAD --` on all three is empty). Deltas only below.

**Re-verified by execution** (worktree `wt-3059-coqui-arm`, read-only; HEAD `9f542097`, `git status --porcelain` empty before and after): server `npx tsc -p . --noEmit` exit 0 · frontend `npx tsc -p tsconfig.json --noEmit` exit 0 · `npx eslint` on all five touched source files exit 0 · `engine-language-coverage.guard.test.ts` + `generation-fallback-gate.test.ts` = **10 passed** (2 + 8) · `src/store/generation-stream-runner.test.ts` + `src/store/queue-dispatcher-middleware.test.ts` = **53 passed**.

I also ran a read-only probe (scratch file outside the repo; imports the real worktree modules and mutates only in-memory state) to settle finding A empirically rather than by argument. Its output is quoted below.

---

## Dispositions

### A — **STILL OPEN.** The new assertion closes the array-contents half of conjunct 3, not the call-site half — which is the scenario the finding was built around

`server/src/tts/engine-language-coverage.guard.test.ts:79-93`

**What it does close, and the red is honest.** Removing `'coqui'` from `ALL_TTS_ENGINES` reddens the new assertion and *only* the new assertion — the table-only first `it` stays green — so the red-then-green evidence in the commit message is for the reason claimed, not incidental:

```
--- SCENARIO 1: 'coqui' removed from ALL_TTS_ENGINES (the PR's claimed red) ---
  ALL_TTS_ENGINES now: piper,kokoro,gemini,qwen
  it#1 uncovered  = []   <- STILL GREEN (table untouched)
  it#2 ineligible = ["ru","es","fr","de","zh","ja"]   <- RED, and only it#2
```

That is real progress: `'coqui' ∈ ALL_TTS_ENGINES` went from *incidental* coverage in `language.test.ts` to intentional coverage carrying the #3059 message. And `model-keys.ts` is confirmed byte-identical to `origin/main` at HEAD, so the mutation was reverted.

**What it does not close.** The guard imports `ALL_TTS_ENGINES` directly (`:45`) and passes it as the argument itself (`:82`). Nothing in the guard reads, references, or is derived from what `generation.ts:830` actually passes. So pass 1's named scenario — swapping `ALL_TTS_ENGINES` at the call site for a real installed-on-this-box set, the change #3059's own "Explicitly out of scope" section parks — is still completely invisible. Constructed and measured:

```
--- SCENARIO 2: generation.ts:830 passes a real installed set (no coqui weights) ---
  lang=ru: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  lang=es: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  lang=fr: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  lang=de: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  lang=zh: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  lang=ja: coquiEligible=false  =>  (nonEnglishBook && !coquiEligible) = true
  it#1 uncovered  = []   <- GREEN
  it#2 ineligible = []   <- GREEN
  ==> guard is GREEN while the removed arm is live for EVERY non-English book.
```

Both assertions green, the removed arm live for all six validated non-English languages, and the downstream chain pass 1 traced (park → "Proceed anyway" → `MissingDesignedVoiceError` → `code: 'unknown'` → srv-11 breaker pauses the whole queue) unchanged. The new assertion restates the first through a longer path *plus* one extra bit (`'coqui' ∈ ALL_TTS_ENGINES`); it does not reach the call site.

To be fair to the fix: pass 1 proposed exactly this snippet and asserted it "closes all three at once". Pass 1 was wrong about that, and the author implemented it faithfully. The gap is pass 1's error carried into the fix — but it is still a gap, and `9f542097` now claims in three places that it is closed (finding F).

**Conjunct 1 is also still unguarded, and the header stopped recording it.** Nothing pins that `generation.ts` still calls `sidecarLanguageName` before the gate or that it still throws. The old header at `d7c94b9c` at least *stated* the fact ("a language outside the registry throws earlier in the same route (before this gate is ever evaluated)"); the rewritten header drops that sentence entirely, so the guard's scope record is now silent on the one conjunct that has no coverage at all and never did.

**Two ways to close it, both cheap and both repo-idiomatic** — neither is a design pass, since they are two implementations of one agreed behaviour:
- export the derivation from `generation.ts` (`export function deriveCoquiEligibility(bookLanguage)` wrapping `:830-831`) and have the guard call *that*, so the guard follows the call site instead of duplicating its argument; or
- a deliberately-syntactic source scan asserting the second argument at `generation.ts:830` is still `ALL_TTS_ENGINES` — the same idiom `server/src/workspace/cast-lock.guard.test.ts` already uses, headers-and-blind-spots included.

If neither is wanted, the honest alternative is to accept the bound and say so — which is finding F.

---

### B — **VERIFIED RESOLVED**

`server/src/tts/engine-language-coverage.guard.test.ts:29-39`. The header now names two limits, and the demotion one is stated correctly and specifically: `sidecarLanguageName` → `getLanguageEntry` (confirmed at `language.ts:34-43` — it does not consult `supported`) and `requireBookStateLanguage` returning the stored language raw. Matches what the code does.

### C — **RESOLVED for the contradiction; one over-claim clause remains** (rolled into F, not double-counted)

`docs/release-notes-next.md:194`. Both pass-1 defects are gone and the replacements are true, not merely different:
- the condition is now stated in the past tense ("was `nonEnglishBook && !coquiEligible` at the time of this fix") and the bullet closes with "so the gate is unconditional now" — no sentence in the bullet contradicts another any more. I read the whole bullet for this, not just the changed clause.
- "fails the moment the registry and `ENGINE_LANGUAGE_SUPPORT.coqui` stop agreeing" is replaced by an accurate description of the actual predicate, and the two carve-outs (`supported: false`, demotion) are now named in the register itself.
- the `generation.ts:938` / `:952` citations are still accurate at HEAD (verified: `:938` is `qwenUnavailable && nonEnglishBook && !coquiEligible`, `:952` its `&& coquiEligible` twin). The `generation.ts:851` citation the PR body says was retired is indeed gone. `RELEASE_NOTES.md` has no fs-60/`coquiEligible` twin to correct — checked, nothing owed there.

The residual: the clause "covering both the table … and the **`ALL_TTS_ENGINES` call-site conjunct** the removed arm's unreachability depended on". That is the same over-claim as F, in the durable technical register.

### D — **VERIFIED RESOLVED**

Both comments are now true, and the underlying factual claim checks out. `grep -rn 'voice-not-designed'` over `server/src` at HEAD returns only the union member (`failure-taxonomy.ts:47`), a comment (`:205`), the remediation table key (`failure-remediations.ts:174`), and a table-completeness test (`failure-taxonomy.test.ts:424`) — **no emitter**. There are also no dynamic `errorCode:` assignments in `generation.ts` (all are string literals), so nothing constructs the code indirectly. "No server path currently emits voice-not-designed today" is accurate.

The deliberate-retention decision is sound and, if anything, *strengthened* by A: both comments justify keeping the plumbing by pointing at "a live scenario where the fail-fast gets restored (see `engine-language-coverage.guard.test.ts`'s header)", and that scenario is exactly the one the guard cannot catch — so it is live, unguarded, and worth keeping the machinery for. The cross-reference resolves; `failure-taxonomy.test.ts:424` is a table-completeness assertion, not vacuous coverage of a dead state.

### E — **VERIFIED RESOLVED**

`server/src/routes/generation-fallback-gate.test.ts` differs from `origin/main` by exactly one line (`git diff --numstat origin/main...HEAD` → `1  1`), and the new number is correct: `grep -n sidecarLanguageName` puts the call at `generation.ts:854` at HEAD, with `:805` being `requireBookStateLanguage` as pass 1 said. The file is 8/8 green.

Noting for the record, without re-litigating: the *other* half of pass 1's E — `generation.ts:938` left in place while its twin is deleted, with the same file's `:444-457` comment still recording the repo's opposite disposition — is untouched at `9f542097`. That comment is still accurate about `:938`, so nothing is false; it was triaged out of scope for this round.

---

## New findings

### 🟠 Correctness — 1

#### F. The guard now asserts, in three places, that it catches the call-site swap. It cannot — and the claim will stop the real check ever being added

`server/src/tts/engine-language-coverage.guard.test.ts:14-16` and `:73-78`; `docs/release-notes-next.md:194`

The header sets up the gap precisely and then claims the fix for it:

> "…a future change to what `generation.ts:830` passes as the installed-engines set (e.g. swapping `ALL_TTS_ENGINES` for a real installed-on-this-box set …) could make `coquiEligible` false for every language while the table stays untouched. **The second `it` below asserts through `resolveEligibleEngines` itself so it can catch that**…"

The antecedent of "that" is unambiguous, and Scenario 2 above is that exact case: both assertions green, arm live for every non-English book. The claim is false. The reasoning it rests on is also broken on its own terms — "assert through `resolveEligibleEngines` itself" only helps if the *argument* tracks the call site, and `:82` hardcodes `ALL_TTS_ENGINES` from its own import at `:45`.

Restated more explicitly inside the test body at `:73-78`: *"so this also catches a future change to what installed-engines set generation.ts passes there (e.g. a real installed set instead of ALL_TTS_ENGINES)"* — same claim, same falsity. And a third time in the technical register at `docs/release-notes-next.md:194`: *"covering both the table … and the `ALL_TTS_ENGINES` **call-site** conjunct"*.

**Concrete failure scenario** (the reason this is correctness, not taste): someone resolves #3059's parked out-of-scope question, changes `generation.ts:830` to pass an installed set, and — being conscientious — opens the guard to check whether they have broken it. The header tells them in so many words that this guard exists to catch precisely their change. It is green. They ship, and every non-English book on a box without Coqui weights parks, offers "Proceed anyway", deterministically re-fails with `code: 'unknown'`, and trips the srv-11 breaker on the third chapter. The false claim is worse than the silence it replaced: at `d7c94b9c` the guard was merely narrow; at `9f542097` it actively certifies the coverage it lacks.

This is the failure shape the PR's own thesis names — *"it becomes an invariant by being enforced, not by being asserted in a comment"* — inverted: the comment now asserts an enforcement that does not exist. No runtime behaviour is wrong; the defect is entirely in the guard's own scope record, which is the one artifact whose accuracy the guard's value depends on.

**Fix:** either close the gap for real (the two options under A) and keep the prose, or keep the guard as-is and correct all three sites to say what it actually covers — `ALL_TTS_ENGINES`'s *contents*, not what the call site passes — and record the call-site conjunct as a third known limit alongside the two already listed at `:29-39`.

### 🟡 Cleanup — 1

#### G. The PR body was not updated for `9f542097`; three of its statements are now false and the incidental fixes are undeclared

PR #3083 body

CLAUDE.md's incidental-findings rule requires incidental fixes to be *declared in the PR body* ("Also fixed, found in passing: …"), because an unannounced fix reads as scope creep. `9f542097` adds two files to the diff — `src/store/queue-dispatcher-middleware.ts` and `src/store/generation-stream-runner.ts` — that the body never mentions at all, and the fix commit's own message is the only record of any of it.

Three specific statements in the body are now false against HEAD:
- **"## The guard"**: *"`engine-language-coverage.guard.test.ts` asserts every validated (`supported: true`) non-English registry code appears in `ENGINE_LANGUAGE_SUPPORT.coqui`"* — it now makes two assertions, and the second is the point of this commit. Same paragraph: *"The header also records the guard's own **limit**"* (singular) — there are now two.
- **"## Test plan"**: *"`src/tts/engine-language-coverage.guard.test.ts` — **1 passed**"* — it is 2. The pasted red output is still the `'ru'`-dropped table run; the ALL_TTS_ENGINES red evidence exists only in the commit message.
- **"## Test plan"**: *"`src/routes/generation-fallback-gate.test.ts` — 8 passed, **unchanged**"* — it is changed, by the one-line fix that discharges finding E. ("No test in it was added, removed or rewritten" remains true.)

A reviewer checking the body's numbers against the branch finds three mismatches, on a PR whose entire argument is that prose about code must be kept true.

---

## Verdict

**One finding carried over as still-open (A) and one new correctness finding (F); one cleanup finding (G).** B, D and E are verified resolved; C is resolved except for the clause folded into F.

The fix commit is a real improvement — the new assertion is a genuine test that genuinely fails, proven red for the right reason and only it; the demotion limit, the two frontend comments, the release-note contradiction and the line citation are all correctly and precisely fixed; nothing executable moved in the two frontend files and every leg is green. But the finding it was dispatched to close is not closed. The guard still cannot see the change #3059 itself parks, and `9f542097` converted a silent coverage gap into a coverage gap that certifies itself as covered. The cheapest honest close is to correct the three prose sites and list the call-site conjunct as a third known limit; the better one is to make the guard call the derivation `generation.ts` actually uses.

dudarenok-maker:
## Triage note — finding F is a defect I introduced (head 9f542097)

Pass 2's finding **F** is correct and worth being explicit about, because the chain of custody matters more than the fix.

Pass 1 proposed the `resolveEligibleEngines(code, ALL_TTS_ENGINES)` assertion and stated it "closes all three at once". **I repeated that claim in the fix brief**, and the implementing agent built exactly what it was told. All three of us were wrong in the same way: asserting *through* the function only helps if the argument tracks the call site, and the guard supplies the argument itself. Pass 2 measured the residual gap and it is total — an installed set of `['qwen','kokoro','gemini']` makes `coquiEligible` false for all six validated non-English codes while both assertions stay green.

The part that makes this a 🟠 rather than a 🟡 is the prose, not the predicate. Before `9f542097` the call-site conjunct was uncovered and unmentioned. After it, three places — the guard header, an inline comment, and the release note — tell a reader the guard catches exactly the change they are making. **A guard that certifies coverage it lacks is worse than one that is silent**, and it is the inversion of this PR's own thesis, which is that an unenforced coupling should stop being asserted in prose.

The likely reader is not hypothetical: whoever resolves the question #3059 parked as out of scope — whether "Proceed anyway" can promise Coqui on a box without weights — will open this guard first.

Going in:

- a third assertion that watches the call site, as a syntactic source scan following `server/src/workspace/cast-lock.guard.test.ts`'s established idiom, with that pattern's own discipline about declaring what a text scan cannot see (an aliased variable, a computed argument);
- conjunct 1 — the `sidecarLanguageName` throw ordering — restored as a **named limit**. `9f542097` dropped the sentence that stated it. It is not being guarded; naming it is the deliverable;
- the three prose sites corrected to describe what the guard actually does;
- the PR body rewritten (finding **G**) — it was never updated for `9f542097`, and the two frontend store files are in the diff with no "Also fixed, found in passing" declaration.

Not going in: the second `it` stays. It moved `'coqui' ∈ ALL_TTS_ENGINES` from incidental coverage in `language.test.ts` to intentional coverage carrying the #3059 message, which is worth keeping on its own terms.

dudarenok-maker:
## PR review — pass 3 (head f5e36cbf, depth high)

Final pass under the 3-pass cap. Deltas only: `3845e7da` and `f5e36cbf`, plus a disposition on every prior finding.

**Tree discipline.** Worktree `wt-3059-coqui-arm`, HEAD `f5e36cbf` and `git status --porcelain` **empty** before and after. I did mutate `server/src/routes/generation.ts` three times to measure the new guard — each from a pristine md5-checked copy held outside the repo, each restored immediately; final state re-verified `md5 = eb205dd8b92e25dd47023f791a2b8597`, `git status --porcelain` empty, `git diff --stat` empty, HEAD unmoved. The primary checkout is also byte-for-byte as I found it (`?? docs/testing/onbox-a-series-audit-2026-09-05.md`, HEAD `8e4025ee`); `wt-3081-flake-repro` untouched.

**Re-verified by execution.** `src/tts/engine-language-coverage.guard.test.ts` + `src/workspace/state-language.guard.test.ts` + `src/force-rerun-triggers.test.ts` + `src/routes/generation-fallback-gate.test.ts` = **86 passed (4 files)** at HEAD. `generation.ts` is byte-identical between `d7c94b9c` and `f5e36cbf` (`git diff` on that path across those two commits is empty), so pass 1's six-attack unreachability proof is undisturbed and was not redone.

---

## Dispositions

### A — **RESOLVED.** it #3 closes the call-site conjunct; re-derived, not taken on trust

Mutated `generation.ts:830` to `resolveEligibleEngines(bookLanguage, [...ALL_TTS_ENGINES])` — semantically identical, so nothing about the *program* changes — and ran the guard:

```
 ❯ src/tts/engine-language-coverage.guard.test.ts (3 tests | 1 failed)
   × generation.ts's resolveEligibleEngines call site still passes ALL_TTS_ENGINES
AssertionError: ... second argument that is no longer the bare identifier ALL_TTS_ENGINES:
  line 830: resolveEligibleEngines(bookLanguage, [...ALL_TTS_ENGINES])
      Tests  1 failed | 2 passed (3)
```

Only it #3 reddens, it #1 and it #2 stay green, and the message names the offending line. That is the call-site conjunct genuinely enforced, not restated through a longer path.

I also probed the *derivation-moved* case, which the earlier passes did not: aliasing the function import (`import { … resolveEligibleEngines as resolveEngines }` + `resolveEngines(bookLanguage, ALL_TTS_ENGINES)`) reddens the guard's **first** expect —

```
AssertionError: no resolveEligibleEngines(...) call found in server/src/routes/generation.ts —
the coquiEligible derivation this guard exists to watch appears to have moved or been removed.
: expected 0 to be greater than 0
```

— so the guard fails **closed** when the derivation leaves the file. That is the right shape.

Conjunct 1 (`sidecarLanguageName` throws before the gate) is still unguarded, and is now correctly **named as a limit** at `:75-79` — which was the triage note's stated deliverable, not a promise of coverage. A is discharged as scoped.

### B — **remains resolved.** The demotion limit is intact and still accurate (`engine-language-coverage.guard.test.ts:83-89`).

### C — **RESOLVED, including the residual clause.** I read the whole fs-60 bullet at `docs/release-notes-next.md:194`, not just the changed span. The pass-2 residual ("covering both the table … and the **`ALL_TTS_ENGINES` call-site conjunct**") is gone; the bullet now enumerates three assertions, describes each correctly, names the `supported:false` / demotion / conjunct-1 limits, and defers to the guard header for the rest. The `generation.ts:938` / `:952` citations in it are still accurate at HEAD (nothing in `generation.ts` moved after `d7c94b9c`).

### D — **remains resolved.** Both frontend comments are unchanged since `9f542097` and still true; no server emitter of `voice-not-designed` exists.

### E — **remains resolved.** The `~805` → `~854` citation stands; `generation-fallback-gate.test.ts` still differs from `origin/main` by exactly that one line and is 8/8 green. The `:938` twin is still untouched and still out of scope, as triaged.

### F — **RESOLVED.** All three prose sites are corrected, and the correction is specific rather than hedged: the header at `:21-38` now states outright that it #2 "does **NOT** see the call site", explains *why* (it imports `ALL_TTS_ENGINES` and passes its own import), spells out the exact scenario that would leave it green, and cites the pass-2 review that measured it. The inline comment at `:142-145` says the same. The release note describes the three assertions accurately. And the claimed coverage now actually exists (see A). This is the honest close, not the cheap one.

### G — **STILL OPEN — it recurred one commit later.** See the cleanup finding below.

---

## New findings

### 🟠 Correctness — 1

#### H. it #3's declared blind-spot list is wrong in one direction and incomplete in the dangerous one — and the one gap it omits is a measured fail-open

`server/src/tts/engine-language-coverage.guard.test.ts:90-101`

The header's KNOWN LIMITS block says:

> "an ALIASED import of either name (`import { resolveEligibleEngines as foo }` or `import { ALL_TTS_ENGINES as bar }`) is **invisible to it**, since it matches on the literal identifier text"

**Both halves of that are false, measured.** Aliasing the function name reddens the guard via the `calls.length > 0` assertion (output quoted under A above). Aliasing `ALL_TTS_ENGINES` makes the second argument read `bar`, which is not the literal `ALL_TTS_ENGINES`, so it lands in the `violations` list and reddens too (`:200-208` — the filter requires `secondArg.text === 'ALL_TTS_ENGINES'`). Neither is invisible; both fail closed. That direction of error is safe, but it is still a false statement in the artifact whose accuracy is this PR's entire thesis.

**The gap the list omits is the one that actually fails open: rebinding the identifier text.** The scan matches `secondArg.text`, so anything that makes the *text* `ALL_TTS_ENGINES` refer to a different value passes green. Measured, on a two-line, type-correct, lint-clean edit (`eslint.config.mjs` configures no `no-shadow`):

```ts
-import { ALL_TTS_ENGINES } from '../tts/model-keys.js';
+import { ALL_TTS_ENGINES as ALL_ENGINES_FULL } from '../tts/model-keys.js';
...
+  const ALL_TTS_ENGINES = ALL_ENGINES_FULL.filter((e) => e !== 'coqui');
   const eligibleEngines = resolveEligibleEngines(bookLanguage, ALL_TTS_ENGINES);
```

```
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

All three assertions green. `coquiEligible` is now false for all six validated non-English codes, `nonEnglishBook && !coquiEligible` is live for every non-English book, and pass 1's downstream chain follows unchanged (park → "Proceed anyway" → `MissingDesignedVoiceError` → `code: 'unknown'` → srv-11 breaker pauses the whole queue on the third chapter).

**Why this is a realistic spelling, not a contrived one.** The reader this guard is written for is whoever resolves the question #3059 parks — "can 'Proceed anyway' promise a Coqui fallback on a box without weights?". Narrowing the set *in place*, keeping the local name, is the **minimal** diff: the call site line and every downstream use of `eligibleEngines` / `coquiEligible` stay untouched. The guard does catch the other natural spelling (`resolveEligibleEngines(bookLanguage, installedEngines)` → red, measured under A), so this is a partial gap, not a total one — unlike F, which was total. But it is a gap in exactly the scenario the guard exists to catch, and the limits list a conscientious reader consults tells them the opposite of what is true about it.

Compare `server/src/workspace/cast-lock.guard.test.ts:105-170`, the idiom this guard's header says it borrows: that list is exhaustive *and* directional — it names the aliased writer import, the aliased path import, the hand-built path, lock identity vs. presence, call-graph indirection, the latent migration seam, and the `scripts/*.mjs` files outside the scan, and for each it says what the guard actually does. This one borrowed the framing and dropped that discipline.

**Fix — two options, neither a design pass:**
- **Prose only (~6 lines):** delete or correct the aliased-import sentence (say that aliasing either name reddens, and which assertion catches it), and add the rebinding case — a local shadow, or an import renamed *to* `ALL_TTS_ENGINES` — as the one known fail-open.
- **Close it (~10 lines, same AST walk):** additionally assert that in `generation.ts`, `ALL_TTS_ENGINES` is imported un-aliased from `'../tts/model-keys.js'` and is never re-declared as a variable, parameter or binding element. That makes the identifier text mean what the guard assumes it means.

I would take the second — it removes the need to be right in prose about a fail-open — but the first is an honest close and is what F's own resolution did for the analogous case.

---

### 🟡 Cleanup — 1

#### G (recurring). The PR body was rewritten at `3845e7da` and is accurate for that commit; `f5e36cbf` then landed undeclared

PR #3083 body

`3845e7da` did the rewrite G asked for, and it is a good one. Then `f5e36cbf` added four files' worth of change that the body does not mention anywhere:

- `server/src/workspace/state-language.guard.test.ts` — `'routes/generation.ts': 2 → 1` plus its audit-trail comment. This is squarely the CLAUDE.md incidental-findings class ("a bookkeeping row your change made owed"), fixed in-round, correctly — and undeclared.
- `server/vitest.config.ts` + `server/src/force-rerun-triggers.test.ts` + the guard-header note at `:48-56` — the `forceRerunTriggers` entry. Same class, same omission.
- **#3085**, filed in this round as the design-pass deferral for the general case, is not referenced in the body at all. It is a good ticket (three named options, the decision stated, a measured table) and it earns its deferral — but a reader of the PR has no way to reach it.
- The **Test plan** section lists neither `state-language.guard.test.ts` nor `force-rerun-triggers.test.ts`, both of which this PR changes.

Also missing from the body: **the trade the new trigger makes.** Measured below — it is small, and it is the right call; it is just nowhere stated in the artifact a reviewer reads.

This is the same defect as pass 2's G, one commit later, on a PR whose argument is that prose about code must be kept true. It is cleanup, not correctness — but it is the second occurrence, so the fix worth making is "update the body as the last step before merge", not "update it once".

---

## ✅ What I attacked and could not break

**The `2 → 1` count is right, and nothing else in that file was derived from the old 46.** Exactly one `stateJsonPath(` remains in `generation.ts` (`:2206`); `git diff origin/main...HEAD` confirms the removed arm held the other one (`const statePath = stateJsonPath(bookDir)` … `await writeStateJsonAtomic(statePath, …)` inside the `voice-not-designed` persist), which is precisely what the new audit-trail comment says. The map now sums to **45 across 23 keys** — I summed it by hand. `G3_FLOOR_SITES = 35` and `G3_FLOOR_FILES = 19` both still hold with margin. The assertion is genuinely both-ways (`:424-442`: map→observed *and* observed→map), so a file dropping to zero sites or gaining an unmapped one both redden. Incidentally, the header at `:45-46` already said "45 across 23" on `origin/main` while the map summed to 46 — a pre-existing prose/data disagreement that this change happens to resolve rather than create.

**The trigger claim is true, and I measured the corollary the prompt asked about.** With `generation.ts` dirty and nothing else, `npx vitest list --filesOnly --changed` from `server/` selects **543 files**. That is the *entire* fast tier: 554 `*.test.ts` under `server/src` minus the 11 in `SLOW_FILES_TO_EXCLUDE` = 543 exactly. So yes — this is a whole-suite trigger, by design: vitest's own code is `if (matcher && related.some((file) => matcher(file))) return specs;` (`vitest/dist/chunks/cli-api…js:11496`), which returns every spec.

I verified the "before" figure from CI rather than from the commit message: run `34167026051`'s Linux server leg (head `3845e7da`, `--changed`-scoped) logged `Test Files 31 passed (31)` and reported **success**, while the unscoped Windows leg ran the full suite and **failed** on the stale count. So a `generation.ts`-touching PR goes from **31 → 543** files on that leg.

**That cost is real but immaterial, and I could not turn it into a finding.** The current full run of that job is **3m40s** — below the critical path (slowest e2e shard on this PR is 5m17s), and `verify.yml`'s own header records that Actions minutes are free and uncapped on this public repo. Locally nothing changes at all: `scripts/verify-cache.mjs`'s `test:server` step is glob-based on `server/src/**` and already ran the whole suite for any `generation.ts` diff. The repo has direct precedent for triggering on an importable file (`synthesise-chapter.ts`, added "for consistency" by #1932), and `force-rerun-triggers.test.ts`'s `NOT_COVERED` cases — which guard against *widening* a trigger — all still pass. The general question is filed as **#3085** with three defensible options and the decision explicitly named, which is a legitimate design-pass deferral under CLAUDE.md, not a dodge. My only note is that none of this reaches the PR body (folded into G).

**Cross-check for other staled facts about `generation.ts`: clean.** I grepped `routes/generation` across `server/src`, `src`, `scripts` and `.github`; every remaining reference is a structural comment (the `gpu/` leaf-gate cycle notes, `capacity-retry.ts`, `analysis-cache-rebuild.ts`, `registry.ts:191`'s `resolveSegmentQaRerecords` default) or an import — none carries a count or a line citation the removal invalidated. The one candidate a search surfaces, `docs/superpowers/plans/2026-08-13-language-recurrence-and-prompt.md:103`'s `generation.ts:1453,2239`, is **not** owed: it is a dated snapshot ("re-measured 2026-08-14") of a task that has already landed (`generation.ts:42` imports `writeStateJsonAtomic`; `:2227` uses it), and the plan's own header instructs re-running the greps because "line numbers drift on every merge".

**The new `MAIN_COVERED` entry is sound.** `force-rerun-triggers.test.ts` proves the glob at all three checkout shapes (clean, dot-prefixed, real) and `absPathUnder` asserts the named file exists — so a rename of `generation.ts` reddens the trigger list as well as the guard. `typescript` is a declared devDependency in `server/package.json` (`^6.0.0`), and three other server guards already import it, so it #3's compiler-API use is not a new portability risk.

**House gates.** `Closes #3059` present, outside any code span, and the PR-issue-link check is green. No `cast.json` writes touched; no config knob; no derived artifact staled; regression-plan / on-box-acceptance not-applicables each stated with a reason. Release notes: the fs-60 bullet edit-in-place remains the right call, and `f5e36cbf` adds no shippable delta (CI scoping + a guard count) so no new entry is owed. CI at `f5e36cbf`: every leg green except **Windows (test + test:server)**, still pending as I write — that is the one leg that caught the stale count last time, so it is worth waiting on before merge.

---

## Verdict

**One correctness finding (H) and one recurring cleanup finding (G). A–F are resolved, and A and F are resolved properly rather than papered over.**

`3845e7da` is the right fix for F: it #3 is a real assertion that genuinely fails, I re-derived its red independently and it reddened *only* it #3, naming the line — and the prose that certified coverage the guard lacked is now precise about what each assertion does and does not see. `f5e36cbf` is a good catch on its own terms: the count fix is correct, and noticing that a runtime-scanning guard is invisible to the CI leg that is supposed to enforce it — then filing the general case with the decision named — is exactly the reasoning this PR set out to institutionalise.

**Would I ship it? Not quite as it stands, and the gap is small.** H is a prose defect in the one artifact this PR exists to make trustworthy: the guard's own record of what it can and cannot see. It gets one thing backwards in the safe direction and omits the only case I could measure as fail-open — a two-line edit to `generation.ts` that leaves the guard 3/3 green while the removed arm is live for every non-English book. The reader who will rely on that list is named in the guard's own header, and they will read it while making precisely that change. Fix is six lines of prose, or ten of AST if you want it enforced rather than declared; then update the body for `f5e36cbf` and merge. If you would rather merge now and fix forward, the residual risk is bounded — the guard still catches the *likelier* spelling — but this is the last pass under the cap, so the call is yours rather than another round's.

PR #3113, opened 2026-09-08
feat(server,docs): build GPU-pin auto-revert, discharge on-box rows A2/E7/E11/A18-4

dudarenok-maker:
## Summary

Final step (9) of the 2-card-boot + Pinokio batch chain, on-box acceptance
register campaign. Builds the GPU-pin auto-revert feature (`runAutoRevert`
+ operator toast, `server/src/gpu/auto-revert.ts`) and its supporting
supervisor/route wiring, verifies steps 1-8's real-hardware evidence against
the parent chain's own criteria, and folds the six affected register rows
individually into `docs/testing/onbox-acceptance-register.md`:

- **A2** — DISCHARGED (removed). Cross-card device-steer walkthrough step 9
  confirmed for real on both cards.
- **A3** — NARROWED, not discharged. The 10-item checklist and task 16/16.5's
  build+tests both passed for real, but the real-hardware trigger did not
  reach `runAutoRevert` in production — `start.ps1` absorbs the code-43
  streak before Node's own supervisor observes three distinct exits. Row
  narrows to that one remaining wiring gap, tracked as
  [#3121](https://github.com/dudarenok-maker/Castwright/issues/3121).
- **A12** — NARROWED, not discharged. Bullets 1/3/4 (pin-survives-respawn,
  codec-pin placement, codec-pin-falls-back-to-cpu) confirmed for real on
  both cards; bullet 2 (enumeration-order swap) deliberately left untouched —
  excluded for contention risk against the box's other live lanes, not a
  hardware gap.
- **E7** — DISCHARGED (removed). Pinokio Update's `pip-in-place` branch
  (this throwaway's v1.13.0→v1.15.0 span had genuinely changed `reqHash`;
  `pip check` reproduced BROKEN at the pre-fix baseline, then CLEAN
  immediately after Update with no server started), and a boot-time
  `ensureOrtMarker` run that separately reported `noop` (Update's own
  `pip-in-place` step had already fixed the install, so nothing was left
  for the self-heal to do — recorded honestly as such, not claimed as proof
  of the self-heal's own corrective branch), a fresh Install's
  `pip-in-place`, and Qwen3 install with no `WinError 5` all confirmed for
  real.
- **E11** — DISCHARGED (removed). A genuinely CRLF-mangled
  `requirements/*.txt` normalized to LF without a spurious force-reinstall;
  a subsequent fresh Install also normalized correctly.
- **A18** item 4 alone — DISCHARGED. `import torchcodec` runs clean inside
[...]

PR #3115, opened 2026-09-08
docs(docs): stamp publish token 14 for the owed live-view republish

### LINKED ISSUE #3114: on-box register: the published live view is 3 discharged rows behind main after #3076 merged without republishing
## What

The register's canonical live view at
https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a
is stale relative to `main`. It presents three already-discharged rows as
still-owed debt:

| Figure | Published page | `main` |
|---|---|---|
| Owed | 55 | 52 |
| Group A | 36 | 35 |
| Group B | 2 | 1 |
| Group D | 2 | 1 |

The three extra rows are **A37** (Russian dash-attributed dialogue), **B2**
(analysis language gate) and **D1** (non-English ASR content-QA calibration).
Rows **A16** and **A23** also carry pre-#3076 body text.

## Cause

PR #3076 (`docs(docs): batch step 7 - fold B2/D1/A37, regenerate live-view
glance table`, merged 2026-09-08, fold commit `e3bad2a6`) edited both tracked
register surfaces and merged, but **step 4 of the register's own four-step
publish procedure never ran** — the artifact was never republished.

## Why the publish token did not catch it

`#3076` edited `onbox-acceptance-register-live-view.html` **without running
`npm run stamp:publish-token`**. Tracked and published therefore both read
`data-published-as="13" data-publish-id="a3ef6817d22b"` while their contents
had diverged.

The register's documented one-line disambiguator — *"published HIGHER than
tracked → someone is AHEAD of you; EQUAL or LOWER → the BEHIND case"* — still
returns the right verdict here (EQUAL → BEHIND), but only because BEHIND is
the fallthrough. An unstamped edit makes the counter uninformative rather than
wrong, so the diagnosis has to fall back to a row-by-row diff, which is the
manual byte-diffing that #2599/A41 built the token to replace.

## Fix

Stamp a fresh token and republish `main`'s copy. Confirmed non-destructive:
`--against-published` reports **no** BEHIND rows, because `origin/main` also
lacks A37/B2/D1 — the #2199 signature of a deliberate discharge rather than a
competing lane's publish. The only report is content-drift on A16 and A23,
which are exactly the two rows #3076 rewrote.

## Worth considering separately

Nothing mechanically requires a stamp when the live view changes. `#2599`
records that the token is unenforced; this is the first observed case of an
unstamped edit actually costing a diagnosis.

dudarenok-maker:
## Summary

The on-box register's canonical live view has been three discharged rows
behind `main` since PR #3076 merged on 2026-09-08. That PR folded **A37**,
**B2** and **D1** out of the register and narrowed **A16**/**A31**, but never
ran step 4 of the register's own publish procedure — so the artifact still
shows **55 owed against main's 52**, group **A at 36 against 35**, and **B**
and **D** each one row high. Three rows that are genuinely discharged are
being presented to the reader as outstanding debt.

This PR is the one tracked change that republish requires: a fresh publish
token, stamped by `npm run stamp:publish-token` and committed **before** the
`--against-published` run, per the register's step 0. The artifact itself is
republished from this branch before merge, which is the documented order.

### Why this went unnoticed

`#3076` edited the live view without re-stamping, so tracked and published
both read `data-published-as="13" data-publish-id="a3ef6817d22b"` while their
contents had diverged. The register's one-line disambiguator (*published
HIGHER than tracked → AHEAD; EQUAL or LOWER → BEHIND*) still lands on the
correct verdict, but only because BEHIND is the fallthrough — an unstamped
edit makes the counter uninformative rather than wrong, and the diagnosis
falls back to the row-by-row diffing that #2599/A41 built the token to
replace. Noted in the issue as worth a separate look; `#2599` already records
that the token is unenforced.

### The publish is non-destructive, and that was checked, not assumed

`npm run check:onbox-register -- --against-published <saved live page>`
reports **no BEHIND rows**. A37, B2 and D1 are absent from `origin/main` too,
which is the #2199 signature of a deliberate discharge rather than a
competing lane's pre-merge publish — the case the checker deliberately does
not report. The only finding is content-drift on **A16** and **A23**, and
those are exactly the two rows `#3076` rewrote (A23's body carried the
"A22 and B2 each prove their mechanism" sentence that B2's fold invalidated).
Local matches the `origin/main` baseline; the published page does not. That
is the register's documented BEHIND signature, and republishing `main`'s copy
is its reconciliation.

No other lane has published ahead of this one: the three open PRs that touch
the register (#3113, #3061, #3063) are all still at the wave-12 page.

## Test plan

- `npm run check:onbox-register` — OK, the tracked pair agrees.
- `npm run register:build` — no diff; the derived figures on `main` were
  already correct. Only the *published* copy was stale.
- `npm run check:onbox-register -- --against-published <saved live page>` —
  no BEHIND rows; content-drift on A16/A23 only, diagnosed above.
- Republished to the canonical URL with `url` passed, from this branch,
  before merge. Verified afterward that the live page reads token 14
  (`b69a8bb984af`), owed 52, A 35, B 1, D 1.

Release notes skipped deliberately: bookkeeping only, no user- or
operator-visible product delta.

Closes #3114

### DIFF
--- a/docs/testing/onbox-acceptance-register-live-view.html
+++ b/docs/testing/onbox-acceptance-register-live-view.html
@@ -165,7 +165,7 @@
   <!-- Publish token. NEVER hand-edit: run `npm run stamp:publish-token`.
        The counter orders publishes; the id proves which branch produced one.
        Attributes, not a comment: checkLiveView removes comments before reading. -->
-  <div hidden data-published-as="13" data-publish-id="a3ef6817d22b"></div>
+  <div hidden data-published-as="14" data-publish-id="b69a8bb984af"></div>
 
   <p class="lede">
     Shipped behaviour that only real hardware can prove — a live GPU, a real sidecar, a real


dudarenok-maker:
**Review gate — exempt.** Changed-file set is `docs/testing/onbox-acceptance-register-live-view.html` alone, entirely under `docs/**`, so this PR meets the `pr-review-gate` skill's docs-only file-set test (the same test CONTRIBUTING.md's doc-only PR fast-path and `verify.yml`'s `paths-ignore` skip use). No review pass runs; recorded here so the absence is a fact rather than a silence.

The one substantive change is a machine-written publish token (`npm run stamp:publish-token`, 13 → 14, id `b69a8bb984af`) — never hand-edited, per the file's own header.

Issue #3089, opened 2026-09-08
[agent instructions][cline-free][task] Survey and triage newly-caught citations from markdown-only blanking scope


dudarenok-maker:
## What

The previous child (**#3088**) scoped `check-register-citations.mjs`'s
code-span blanking to markdown files only. Per issue #3062's acceptance item
2: "a widening that lights up existing files must be triaged, not merged
blind" — this checker gates CI. Your job: run the updated checker across the
real tracked tree, find what the narrower scope newly surfaces, and triage
every finding before this can ship.

## Background you need

Before the fix, every tracked file (not just markdown) had its backtick spans
blanked before citation-scanning, so a citation written inside a template
literal (in `.mjs`/`.ts`/`.py` source) was invisible to the checker. After the
fix, source files are scanned raw — any register-ID-shaped text inside a
template literal, anywhere in the tracked tree, is now a live citation the
checker will validate (nonexistent-ID check, conflicting-subject check,
title-drift check). This can surface **real** newly-caught defects (good — the
bug this checker exists to prevent) or **false positives** — a string that
merely looks like a citation shape (`` `see A12 in the notes` `` as a comment
or log message that was never meant as a register citation).

## The work

1. `git -C C:\Claude\Projects\wt-scripts-3062-blanking pull` isn't needed —
   you're continuing in the same worktree/branch as the prior child:
   `C:\Claude\Projects\wt-scripts-3062-blanking`,
   branch `fix/scripts-3062-markdown-only-blanking`. Confirm you're on it
   (`git status`) before touching anything.
2. Run `node scripts/check-register-citations.mjs` (no flags) and
   `node scripts/check-register-citations.mjs --strict` against the current
   tree. Compare the output to what `git stash` (temporarily reverting to the
   pre-fix `stripInlineCodeSpans` behaviour, i.e. `git show
   main:scripts/check-register-citations.mjs > /tmp/before.mjs` and diffing
   findings) would have reported — you're looking for **new** findings only,
   not re-litigating findings that already existed before this chain started.
3. For each newly-surfaced finding:
   - If it's a **genuine** citation defect (a real wrong/nonexistent register
     ID that happened to live inside a template literal and was previously
     invisible) — fix it at the source (correct the ID, or add the
     discharge/removal annotation the checker already recognises, matching
     the pattern used elsewhere in that file).
   - If it's a **false positive** (register-ID-shaped text that was never
     meant as a citation — e.g. an example/placeholder string in a script or
[...]

PR #2630, opened 2026-08-23
docs(docs): correct stale on-box register row IDs cited in code comments

dudarenok-maker:
## Summary

On-box acceptance register row IDs are **positional** — they renumber whenever a row is discharged. Two discharges landed while this branch was open, and citations to those IDs live all over the repo. Each rotted one points a reader at unrelated shipped behaviour.

This corrects them, and — because correcting them by hand demonstrably does not converge — **adds the check that makes the claim verifiable**.

### Why the check exists

Seven review passes. Findings per pass stayed flat, and every miss was the same defect: *one claim on several surfaces, only some edited.* Each round landed on a surface the previous round's instrument never covered — run-sheet prose, an active plan, a design spec, an HTML mirror, a plan-of-record table, a `status: stable` spec, an archived plan.

**The sweep itself was wrong, twice, in ways no pass caught for rounds.** A uniform −1 was applied to seven headings where four needed −2 — while the body claimed IDs were resolved individually. A plan-of-record table was then "corrected" to agree with the wrong pack, so two sources agreed because one derived from the other. A sixth ID was correct when written and rotted by someone else's merge landing on `main`.

### `npm run check:register-citations`

| Check | Status |
|---|---|
| **A** — nonexistent row ID | FATAL |
| **C** — existing ID cited for the wrong subject | **FATAL** — this is the class above |
| **C** — subject absent from every register heading | opt-in `--strict`, never fatal |
| **B** — bidirectional run-sheet linkage | FATAL — 13 owned header pairs |
| unclassified run-sheet mentions | printed, explicitly UNCHECKED |

Frozen paths are documented with per-prefix rationale. `status: stable` and `docs/features/archive/` are **not** excluded — each was found hiding a live "acceptance owed" pointer, and *stable* and *archived* describe the **document**, not every claim inside it.

### It was built hollow twice, and both were caught by injection

Check B first matched an ID **anywhere in a file**, so a run sheet mentioning its row twice passed with a wrong header — and it never recorded ownership for the half of the register's references written as bare filenames. Later, the success line reported a file count that included files the checks never scan, and the header comment described Check A's and Check C's surfaces **backwards** — which is precisely why the wrong-ID class went ungated for four rounds.

Every check is now validated by injecting the defect, observing the failure, and restoring byte-identically. **Re-injecting the original four wrong headings fails a default run**, naming each with "maps to X, not Y".

### What the sweep can now claim

[...]

PR #3124, opened 2026-09-09
fix(scripts): scope register-citation code-span blanking to markdown files

### LINKED ISSUE #3062: check-register-citations blanks markdown code spans in source files, hiding citations in template literals
## What

`scripts/check-register-citations.mjs` reads essentially every tracked file (its own header says so) and applies **`stripInlineCodeSpans`** — a *markdown* construct — to all of them before scanning for register-row citations.

In a `.mjs` file, a **template literal**'s backticks are read as inline-code delimiters, so the text between them is blanked. A `row A101` sitting inside a template literal is therefore invisible to the checker, while the identical text inside a `'…'` or `"…"` string is fatal.

## How it surfaced

Found while fixing PR #3053. A fix agent's first mutation attempt "looked green" for exactly this reason — it had moved a fixture id into a template literal, and the checker stopped seeing it. That is the [instrument-cannot-fail](https://github.com/dudarenok-maker/Castwright/issues) shape: the mutation appeared not to matter, when in fact the detector had gone blind.

PR #3053's landed fix does **not** depend on this — the literal id text is gone entirely — so nothing is currently broken by it.

## Why it matters anyway

**"Wrap it in backticks" is not a reliable way to hide a fixture from this checker, and it is not a reliable way to expose one either.** Whichever direction someone reasons in, the behaviour is accidental rather than designed:

- a genuine wrong-id citation inside a template literal ships undetected;
- a deliberate test fixture inside a single-quoted string is fatal and has to be worked around.

Both have now happened in the same round.

## The decision owed

**Which file types get markdown code-span blanking.** There is more than one defensible answer and the checker cannot choose on its own:

1. **Markdown only** (`.md`, `.html`) — blanking is a markdown affordance and source files should be scanned raw. Cleanest semantically; may surface existing citations in source comments that were previously invisible, so it needs a survey of what it newly catches before it can land.
2. **All tracked files, as today, but documented** — keep the behaviour and state it explicitly at the scanner and in the checker's header, so neither direction is a surprise. Cheapest; leaves the blind spot in place deliberately.
3. **Language-aware** — strip real code spans in markdown, and in JS/TS strip nothing but instead skip *comments* (where an example citation legitimately lives). Most correct, most work, and needs a parser rather than a regex.

Option 1 is the writer's instinct; option 2 is honest about cost; option 3 is what the checker would look like if it were designed today. Pick one before touching the scanner.

## Acceptance

1. A recorded decision among the three, with reasoning.
2. If the blanking scope changes, a survey of what the new scope newly catches across the tracked tree — this checker gates CI, so a widening that lights up existing files must be triaged, not merged blind.
3. A test asserting the chosen behaviour in **both** directions: a citation inside a template literal, and one inside a quoted string, each landing on the correct side of the line.
4. The checker's own header states the rule, since its current header describes the file set but not this transformation.

## Key files

- `scripts/check-register-citations.mjs` — `stripInlineCodeSpans` (~`:1092`), and its call sites at `:1022`, `:1185`, `:1427`
- `scripts/tests/check-register-citations.test.mjs`

## Benefit

*(technical)* Removes an accidental, undocumented blind spot in a checker that gates CI, and stops "put it in backticks" from being a silent behaviour change in either direction.

dudarenok-maker:
## Summary

- Per the operator's 2026-09-08 decision on #3062 (Option 1), `check-register-citations.mjs`'s single-backtick inline code-span blanking (`stripInlineCodeSpans`) is now scoped to markdown-scanned files only (`.md`, `.html`) via a new `isMarkdownScanPath()` helper threaded through every call site. Non-markdown tracked sources (`.mjs`, `.ts`, `.py`, ...) are now scanned raw, so a template literal's backticks are no longer misread as markdown code-span delimiters and can no longer silently hide a register-row citation inside one.
- Widening the scan surface newly caught 2 previously-hidden citations of the nonexistent row ID `A101`, both in explanatory comments (`scripts/validate-sidecar-acceptance.mjs` and its paired test) describing the gate's placeholder example text. #3089's survey confirmed both were false positives, not defects — no real citation-defect fixes were needed. They were reworded to drop the "row" prefix so the bare ID no longer sits on any citation surface, leaving the checker's detection logic itself untouched.
- The checker's own header comment and `stripInlineCodeSpans`/`isMarkdownScanPath`'s own comments state the markdown-only rule explicitly.

## Also fixed, found in passing (pr-review-gate)

- **F1 (correctness, 🟠)**: `titleDriftTokens(row.title, isMarkdown)` was blanking the register row's own title using the *scanned file's* markdown flag instead of treating the register (always markdown) as markdown unconditionally — silently suppressing title-drift detection for 9 backticked-title rows (A17, A20, A22, A23, A25, A32, A33, E3, E9) whenever the citing file was non-markdown. Fixed to `titleDriftTokens(row.title)` (defaults to markdown), paired regression test added.
- **F2 (chore)**: `stripFences` (triple-backtick fence blanking) wasn't threaded through `isMarkdown` alongside `stripInlineCodeSpans`, so a non-markdown file's fenced-looking content (and an unpaired fence line blanking to EOF) could still silently hide a citation — exactly the class of bug this PR set out to fix. Gated `stripFences` behind `isMarkdown` across its 6 call sites, with regression tests.
- **F3 (chore)**: documented the known, accepted residual risk that `.html` (treated as fully markdown per the #3062 decision) can blank backtick spans inside embedded `<script>`/`<style>` content — no behavior change, comment only.
- **F4 (chore)**: added test coverage pairing markdown vs. non-markdown behavior for a previously-untested `isMarkdown` call site (`checkConflictingSubjects`), since thin coverage there is how F1 survived to review.
- **N1 (correctness, 🟠, found on re-review)**: the same `isMarkdown` threading accidentally loosened discharge-annotation blanking too — a real, fatal citation of a nonexistent register row could be silently excused as "discharged" in a non-markdown file if the discharge word sat inside a backtick span, purely due to file extension (the mirror-image bug of F1/F2). Discharge-annotation blanking isn't markdown-specific (its false-positive-prevention rationale applies to any file type), so the two affected call sites (`idSpecificAnnotationPresent`, `dischargeAnnotationPresentAnywhere`) were made unconditional again, matching the pre-existing correct pattern elsewhere in the file. Latent in the real tree (zero live instances); paired regression test added.
- **N2 (chore, found on re-review)**: updated `isMarkdownScanPath()`'s header comment, which described only the inline-code-span gating it was originally added for and had gone stale once it also started gating `stripFences` (F2) — now documents both mechanisms it controls.

## Test plan

- `node --test scripts/tests/check-register-citations.test.mjs` — 152/152 pass, including new regression tests for F1, F2, F4, and N1.
- `node scripts/check-register-citations.mjs --strict` — `check:register-citations: OK`, no fatal findings on the real tree.
- `npm run typecheck` (root, covers `server/` too) — clean.
- `node --test scripts/tests/validate-sidecar-acceptance.test.mjs` — 88/88 pass, confirming the reworded placeholder comments still satisfy every existing assertion.
- Reproduced the fixed-vs-unchanged behaviour directly via the suite's own CLI-mutation tests: a nonexistent-ID citation inside a template-literal-shaped backtick span in a `.mjs` file is now caught, while the identical shape in a `.md` file remains exempt, unchanged; a discharge-annotation excuse inside a backtick span no longer masks a fatal citation in either file type.

## Not applicable
- **Release notes (before-shipping step 5)**: skipped — internal dev-tooling/checker fix, no shippable user- or operator-visible product delta.
- **On-box acceptance (before-shipping step 3)**: skipped — nothing here requires real hardware to prove; all behavior is verified by the automated test suite above.

Closes #3062

### DIFF
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -2115,7 +2115,7 @@ function stripMarkdownLinkUrls(text) {
   return text.replace(MARKDOWN_LINK_REGEX, '$1');
 }
 
-function titleDriftTokens(text, isMarkdown) {
+function titleDriftTokens(text, isMarkdown = true) {
   const withoutLinks = stripMarkdownLinkUrls(text);
   const withoutCodeSpans = stripInlineCodeSpans(withoutLinks, { isMarkdown });
   const raw = withoutCodeSpans.toLowerCase().match(TITLE_DRIFT_TOKEN_REGEX) ?? [];
@@ -2267,7 +2267,7 @@ export function checkCitationTitleDrift(text, filePath, registerRows) {
     for (const id of ids) {
       const row = registerRows.get(id);
       if (!row) continue;
-      const titleTokens = titleDriftTokens(row.title, isMarkdown);
+      const titleTokens = titleDriftTokens(row.title);
       const { ratio, shared } = titleDriftScore(titleTokens, proseTokens);
       if (ratio > TITLE_DRIFT_RATIO_THRESHOLD || shared >= TITLE_DRIFT_MIN_SHARED_TOKENS) continue;
       const message =
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -1027,7 +1027,7 @@ function isDischargeAssertionNegated(scanText, dm, clauseStart) {
  * this flips zero of today's citations — see #2858 for the false-positive
  * analysis of the 4 citations a looser, unshared polarity scan misflagged.
  */
-function idSpecificAnnotationPresent(sectionText, id, _isMarkdown) {
+function idSpecificAnnotationPresent(sectionText, id) {
   const dischargeScanText = stripInlineCodeSpans(sectionText);
   const dischargeMatches = [
     ...dischargeScanText.matchAll(new RegExp(DISCHARGE_ANNOTATION_REGEX.source, 'gi')),
@@ -1109,23 +1109,34 @@ const SINGLE_ID_SPAN_REGEX = new RegExp(`^${ROW_ID_TOKEN}$`);
 /**
  * Whether a repo-relative scanned-file path (e.g. `docs/foo.md`,
  * `scripts/foo.mjs`) should have its single-backtick inline code spans
- * blanked as markdown. Blanking is scoped to markdown sources only (`.md`,
- * `.html`): in a source file (`.mjs`/`.ts`/`.py`/...) a template literal's
- * backticks would be misread as markdown delimiters and a register citation
- * inside one (e.g. `` `row A101` `` in a template string) would be silently
- * blanked and never checked — while the identical text inside a `'...'` or
- * `"..."` string is scanned and can be fatal. Every other tracked file is
- * therefore scanned raw, unblanked. (Decided 2026-09-08 per #3062.)
+ * blanked as markdown. Returns true for `.md` and `.html` files (the
+ * markdown-scanned set); all other tracked files are scanned raw, unblanked.
+ *
+ * Per the operator's #3062 decision, `.html` is treated as fully markdown,
+ * including embedded `<script>` and `<style>` block content — both
+ * `stripFences` (triple-backtick blocks) and `stripInlineCodeSpans`
+ * (single-backtick spans) apply within .html files, so backtick-wrapped
+ * citations inside embedded code blocks are also blanked. This is an
+ * accepted trade-off of the current scope decision (see real-tree example
+ * at docs/features/277-v115-bug-chore-sweep-board.html around lines 1062
+ * and 1085).
+ *
+ * BLANKING SCOPE: `stripFences` is markdown-conditional (gated by `isMarkdown`
+ * at every call site). `stripInlineCodeSpans` is markdown-conditional for
+ * citation-scanning purposes (gated at most call sites: lines 1229, 1859,
+ * 2037, 2135, 2173) but unconditional for discharge-annotation checks
+ * (lines 1031, 2216), where the blanking prevents false positives from
+ * backtick-wrapped example citations. In non-markdown sources
+ * (`.mjs`/`.ts`/`.py`/...), a template literal's backticks would otherwise
+ * be misread as markdown delimiters — a citation inside one (e.g. `` `row
+ * A101` `` in a template string) would be silently blanked by a
+ * markdown-gated scan while remaining visible to downstream checks, creating
+ * inconsistency. Unconditional discharge-annotation blanking trades that
+ * inconsistency for false-positive immunity (the register-owned text and
+ * discharge annotations are not markdown-syntax-specific; a row ID is either
+ * cited for history or it isn't, regardless of whether the document is
+ * markdown).
  */
-// Returns true for .md and .html files (the markdown-scanned set). Per the
-// operator's #3062 decision, .html is treated as fully markdown, including
-// embedded `<script>` and `<style>` block content — both triple-backtick
-// `stripFences` and single-backtick `stripInlineCodeSpans` blanking apply
-// (the latter unconditionally for discharge-annotation false-positive prevention,
-// the former gated by isMarkdown), and backtick-wrapped spans inside those blocks
-// are blanked, which could theoretically hide a citation there. This is an
-// accepted trade-off of the current scope decision (see real-tree example at
-// docs/features/277-v115-bug-chore-sweep-board.html around lines 1062 and 1085).
 function isMarkdownScanPath(relPath) {
   return /\.(md|html)$/i.test(relPath);
 }
@@ -1265,7 +1276,7 @@ export function checkNonexistentIds(text, filePath, registerRows) {
     for (const id of [...byLine.get(i)].sort()) {
       if (registerRows.has(id)) continue;
       const message = `${filePath}:${i + 1} — cited ${id} — no such row in ${REGISTER_PATH} (nonexistent ID)`;
-      if (idSpecificAnnotationPresent(enclosingSectionText(lines, i), id, isMarkdown)) {
+      if (idSpecificAnnotationPresent(enclosingSectionText(lines, i), id)) {
         annotated.push(`${message} — annotated as discharged/removed, not failing`);
       } else {
         errors.push(message);
@@ -1788,7 +1799,7 @@ function recordSubjectConflict(
     }
     if (currentSubjects && currentSubjects.size > 0) {
       const currentSubjectsText = [...currentSubjects].sort((a, b) => a - b).join('/');
-      if (idSpecificAnnotationPresent(enclosingSectionText(lines, lineIndex), id, isMarkdownScanPath(filePath))) {
+      if (idSpecificAnnotationPresent(enclosingSectionText(lines, lineIndex), id)) {
         annotatedDischarge.push(
           `${filePath}:${lineIndex + 1} — cited ${id} for #${subject}, but ${id} now tracks ` +
             `#${currentSubjectsText} (#${subject}'s row has discharged and ${id} was re-minted) — ` +
@@ -2207,7 +2218,7 @@ function extractHeadingTitleEchoes(text, isMarkdown) {
  * A31, etc., which do). This relaxation moves A8 from `findings` to
  * `annotatedFindings` by treating a single-id case more generously.
  */
-function dischargeAnnotationPresentAnywhere(sectionText, id, _isMarkdown) {
+function dischargeAnnotationPresentAnywhere(sectionText, id) {
   // For single-ID headings, look for discharge in the header section only
   // (first ~300 chars) to avoid false positives from body text mentioning
   // other IDs. This covers the criteria-source blockquote region.
@@ -2296,8 +2307,8 @@ export function checkCitationTitleDrift(text, filePath, registerRows) {
       // anywhere in section, or next to this heading's ID). For multi-ID headings,
       // require ID-proximity to avoid one ID's discharge excusing another's drift.
       const isAnnotated = ids.length === 1
-        ? dischargeAnnotationPresentAnywhere(sectionText, id, isMarkdown)
-        : idSpecificAnnotationPresent(sectionText, id, isMarkdown);
+        ? dischargeAnnotationPresentAnywhere(sectionText, id)
+        : idSpecificAnnotationPresent(sectionText, id);
 
       if (isAnnotated) {
         annotatedFindings.push(`${message} — annotated as discharged/removed, not flagged as drift`);
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -33,6 +33,14 @@
 // unconditional/uncached local leg (independent of scope-gating), and whether
 // a git hook should wire it.
 //
+// Scanned file set, and how it is read: the checker reads every non-frozen,
+// non-self-referential git-tracked file (see `gitLsFiles` /
+// `runCheckRegisterCitationsCli`). The single-backtick inline code-span
+// blanking (`stripInlineCodeSpans`) applies to MARKDOWN files only (`.md`,
+// `.html`): non-markdown sources are read raw, unblanked, so a template
+// literal's backticks can never silently hide a citation. See
+// `isMarkdownScanPath` and `stripInlineCodeSpans`'s own comments.
+//
 // Four checks, ordered by precision (least to most likely to need
 // judgment):
 //
@@ -1015,8 +1023,8 @@ function isDischargeAssertionNegated(scanText, dm, clauseStart) {
  * this flips zero of today's citations — see #2858 for the false-positive
  * analysis of the 4 citations a looser, unshared polarity scan misflagged.
  */
-function idSpecificAnnotationPresent(sectionText, id) {
-  const dischargeScanText = stripInlineCodeSpans(sectionText);
+function idSpecificAnnotationPresent(sectionText, id, isMarkdown) {
+  const dischargeScanText = stripInlineCodeSpans(sectionText, { isMarkdown });
   const dischargeMatches = [
     ...dischargeScanText.matchAll(new RegExp(DISCHARGE_ANNOTATION_REGEX.source, 'gi')),
   ];
@@ -1083,10 +1091,36 @@ function idSpecificAnnotationPresent(sectionText, id) {
 // DOES occur, as a composite-key separator, in two real tracked files) — so
 // `\s+`-based regexes can no longer bridge across a blanked span the way a
 // plain-space blank did.
+//
+// SCOPE (per #3062): blanking applies to MARKDOWN scanned files only (`.md`,
+// `.html`). In a non-markdown source file a template literal's backticks are
+// read by this transform as the same markdown delimiter and would silently
+// blank a citation inside one (e.g. `` `row A101` `` in a template string),
+// so `stripInlineCodeSpans(text, { isMarkdown: false })` returns `text`
+// unchanged. Every call site threads `isMarkdown` (from the scanned path)
+// through — see `isMarkdownScanPath`.
 const CODE_SPAN_BLANK_CHAR = '';
 const SINGLE_ID_SPAN_REGEX = new RegExp(`^${ROW_ID_TOKEN}$`);
 
-function stripInlineCodeSpans(text) {
+/**
+ * Whether a repo-relative scanned-file path (e.g. `docs/foo.md`,
+ * `scripts/foo.mjs`) should have its single-backtick inline code spans
+ * blanked as markdown. Blanking is scoped to markdown sources only (`.md`,
+ * `.html`): in a source file (`.mjs`/`.ts`/`.py`/...) a template literal's
+ * backticks would be misread as markdown delimiters and a register citation
+ * inside one (e.g. `` `row A101` `` in a template string) would be silently
+ * blanked and never checked — while the identical text inside a `'...'` or
+ * `"..."` string is scanned and can be fatal. Every other tracked file is
+ * therefore scanned raw, unblanked. (Decided 2026-09-08 per #3062.)
+ */
+function isMarkdownScanPath(relPath) {
+  return /\.(md|html)$/i.test(relPath);
+}
+
+// `stripInlineCodeSpans` takes an explicit `{ isMarkdown }` flag: when false
+// (a non-markdown scanned source), it returns `text` unchanged.
+function stripInlineCodeSpans(text, { isMarkdown = true } = {}) {
+  if (!isMarkdown) return text;
   return text.replace(/\u0060([^\u0060\n]*)\u0060/g, (m, inner) => {
     const trimmed = inner.trim();
     if (SINGLE_ID_SPAN_REGEX.test(trimmed)) {
@@ -1177,9 +1211,9 @@ function deBold(text) {
  * span, so this costs nothing on a real citation today.
  * @returns {Map<number, Set<string>>}
  */
-function extractCitationsByLine(text) {
+function extractCitationsByLine(text, isMarkdown) {
   const stripped = deBold(stripFences(text));
-  const scanLines = stripInlineCodeSpans(stripped).split('\n');
+  const scanLines = stripInlineCodeSpans(stripped, { isMarkdown }).split('\n');
   const byLine = new Map();
   const add = (i, ids) => {
     if (!ids.length) return;
@@ -1210,14 +1244,15 @@ function extractCitationsByLine(text) {
 export function checkNonexistentIds(text, filePath, registerRows) {
   const errors = [];
   const annotated = [];
+  const isMarkdown = isMarkdownScanPath(filePath);
   const lines = deBold(stripFences(text)).split('\n');
-  const byLine = extractCitationsByLine(text);
+  const byLine = extractCitationsByLine(text, isMarkdown);
   const sortedLineIndexes = [...byLine.keys()].sort((a, b) => a - b);
   for (const i of sortedLineIndexes) {
     for (const id of [...byLine.get(i)].sort()) {
       if (registerRows.has(id)) continue;
       const message = `${filePath}:${i + 1} — cited ${id} — no such row in ${REGISTER_PATH} (nonexistent ID)`;
-      if (idSpecificAnnotationPresent(enclosingSectionText(lines, i), id)) {
+      if (idSpecificAnnotationPresent(enclosingSectionText(lines, i), id, isMarkdown)) {
         annotated.push(`${message} — annotated as discharged/removed, not failing`);
       } else {
         errors.push(message);
@@ -1740,7 +1775,7 @@ function recordSubjectConflict(
     }
     if (currentSubjects && currentSubjects.size > 0) {
       const currentSubjectsText = [...currentSubjects].sort((a, b) => a - b).join('/');
-      if (idSpecificAnnotationPresent(enclosingSectionText(lines, lineIndex), id)) {
+      if (idSpecificAnnotationPresent(enclosingSectionText(lines, lineIndex), id, isMarkdownScanPath(filePath))) {
         annotatedDischarge.push(
           `${filePath}:${lineIndex + 1} — cited ${id} for #${subject}, but ${id} now tracks ` +
             `#${currentSubjectsText} (#${subject}'s row has discharged and ${id} was re-minted) — ` +
@@ -1807,7 +1842,7 @@ export function checkConflictingSubjects(fileTexts, registerRows) {
     // citation (see `stripInlineCodeSpans`'s own comment) — anchored
     // headings can't appear inside a code span at all (`^#{2,6}`), so this
     // only ever changes behaviour on the `Criteria source:` surface.
-    const text = stripInlineCodeSpans(deBold(stripFences(rawText)));
+    const text = stripInlineCodeSpans(deBold(stripFences(rawText)), { isMarkdown: isMarkdownScanPath(filePath) });
     const lines = text.split('\n');
     lines.forEach((line, i) => {
       const citedIds = citationShapedLineIds(line);
@@ -1984,7 +2019,7 @@ export function measureWrongIdEligibleLines(fileTexts, registerRows) {
     // an unblanked copy here would silently disagree with Check C again the
     // moment a `Criteria source:`-shaped example command inside a code span
     // is counted as eligible here but is (correctly) blanked away there.
-    const text = stripInlineCodeSpans(deBold(stripFences(rawText)));
+    const text = stripInlineCodeSpans(deBold(stripFences(rawText)), { isMarkdown: isMarkdownScanPath(filePath) });
     for (const line of text.split('\n')) {
       if (extractSubjectNumbers(line).size === 0) continue;
       const shapedIds = citationShapedLineIds(line);
@@ -2080,9 +2115,9 @@ function stripMarkdownLinkUrls(text) {
   return text.replace(MARKDOWN_LINK_REGEX, '$1');
 }
 
-function titleDriftTokens(text) {
+function titleDriftTokens(text, isMarkdown) {
   const withoutLinks = stripMarkdownLinkUrls(text);
-  const withoutCodeSpans = stripInlineCodeSpans(withoutLinks);
+  const withoutCodeSpans = stripInlineCodeSpans(withoutLinks, { isMarkdown });
   const raw = withoutCodeSpans.toLowerCase().match(TITLE_DRIFT_TOKEN_REGEX) ?? [];
   const tokens = new Set();
   for (const t of raw) {
@@ -2118,9 +2153,9 @@ function titleDriftScore(titleTokens, proseTokens) {
  * surface is invented here.
  * @returns {{ lineIndex: number, ids: string[], titleEcho: string }[]}
  */
-function extractHeadingTitleEchoes(text) {
+function extractHeadingTitleEchoes(text, isMarkdown) {
   const stripped = deBold(stripFences(text));
-  const lines = stripInlineCodeSpans(stripped).split('\n');
+  const lines = stripInlineCodeSpans(stripped, { isMarkdown }).split('\n');
   const citations = [];
   lines.forEach((line, i) => {
     const m = line.match(HEADING_ID_REGEX);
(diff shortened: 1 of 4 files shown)


dudarenok-maker:
## PR review — pass 1 (head 99b4c921f7d18da7ee1a0db85ed5173f81b46d20, depth medium)

Depth `medium` per CLAUDE.md's review-depth ladder (single-scope `fix`). Not docs-only, so the gate applies.

Verification actually run (isolated clone in OS temp; primary checkout untouched):

- `node scripts/check-register-citations.mjs --strict` at head → `OK`, exit 0 (3564 scanned files). Also run on **head merged with current `main` (`d3c3d904`)** → still `OK`, exit 0 — no post-merge fatal regression, and `build-register-live-view.test.mjs` is now self-referential on `main`, which drops Check D's advisory count 17 → 5.
- `node --test scripts/tests/check-register-citations.test.mjs` on the merged tree → 146/146 pass.
- **Control measured** for both new tests, against the merge-base checker: `BASE: test1(.mjs) errors=0, test2(.md) errors=0` → `HEAD: test1(.mjs) errors=1, test2(.md) errors=0`. The regression test is red before / green after **for the reason claimed**, and the paired `.md` control discriminates (same inner span, different verdict keyed only on the path). House gate satisfied.
- **Control measured** for the 17 Check-D advisories in `build-register-live-view.test.mjs`: they are present on the merge-base too, so they are **not** caused by this PR. Not reported.
- **Control measured** for Prettier: `prettier --check` fails on all three changed files at the merge-base as well, and `format:check` is wired into neither `verify-cache.mjs` nor any workflow. Not reported.

---

### 🟠 F1 — correctness bug — the register row's own title is now blanked (or not) according to the *citing* file's extension

`scripts/check-register-citations.mjs:2273`

```js
const titleTokens = titleDriftTokens(row.title, isMarkdown);
```

`row.title` comes from `docs/testing/onbox-acceptance-register.md` — always markdown, no matter who cites it. `isMarkdown` here is `isMarkdownScanPath(filePath)` for the **scanned** file (line 2266). So the register's own inline code spans stop being blanked whenever the citing file happens to be non-markdown, and Check D scores a different token set for the same row.

The same PR gets this right one call site over: `stripInlineCodeSpans(rowBodyText)` at line 1462 is deliberately left unconditional because the register body is always markdown. Line 2273 is the same category and was threaded anyway.

Concrete failure, reproduced against the real register (row `E3 · Pair from \`castwright.local\` (plan 256)`), identical heading text in both runs:

| scanned path | base checker | head checker |
|---|---|---|
| `docs/x.md` | 1 drift finding | 1 drift finding |
| `scripts/x.mjs` | 1 drift finding | **0 findings** |

Input: `### E3 · castwright local`. Under the head checker with a `.mjs` path, `castwright` and `local` leak out of the register's ``` `castwright.local` ``` span into `titleTokens`, `shared` reaches 2 = `TITLE_DRIFT_MIN_SHARED_TOKENS`, and the drift is suppressed. Under `.md` (and under the base checker for both paths) it fires. Nine register rows today carry backticked titles (A17, A20, A22, A23, A25, A32, A33, E3, E9), so this is not a contrived shape.

Check D is advisory/never-fatal, so this does not break the gate — but it is a behaviour regression this PR introduces, in the direction of *hiding* drift, and the fix is to stop threading the scanned file's flag into register-sourced text (`titleDriftTokens(row.title)`).

Not currently firing on the real tree only because no non-excluded non-markdown file carries a `^#{2,6} <ID> ·` heading right now — `build-register-live-view.test.mjs` had 17 such lines until `main` made it self-referential. That is timing, not immunity.

---

### 🟡 F2 — cleanup — the new header comment states an invariant the code does not hold

`scripts/check-register-citations.mjs:39-45`

> non-markdown sources are read raw, unblanked, so a template literal's backticks can never silently hide a citation.

They are not read raw. `stripFences` (line 517) and `deBold` (line 1184) still run unconditionally on every scanned file, markdown or not — see lines 1218, 1251, 1848, 2025, 2160, 2267. `stripFences` blanks from any line whose `trimStart()` begins with ` ``` `/`~~~` to the next such line, or to EOF if unpaired.

Reproduced on the head checker:

| input, scanned as `scripts/x.mjs` | errors |
|---|---|
| `// see row A999` | 1 |
| same line between two ` ``` ` lines | **0** |
| same line after **one** unpaired ` ``` ` line | **0** |
| same file with the fence line replaced by ordinary text | 1 |

`openapi.yaml:2976` and `:2987` already carry fence-shaped lines in a tracked non-markdown file, so the transform is live on this corpus. A citation added inside that block — or anywhere after a single unpaired fence line in any `.yaml`/`.py`/`.sh`/`.ps1` file — is silently blanked, which is the exact #3062 class.

Either scope the sibling transforms the same way `stripInlineCodeSpans` was scoped, or narrow the comment to the single transform actually changed. As written the comment is a promise the file does not keep, and the next reader will trust it.

---

### 🟡 F3 — cleanup — `.html` is classified as markdown, leaving the #3062 class armed in the one format that embeds JS

`scripts/check-register-citations.mjs:1119-1121`

```js
function isMarkdownScanPath(relPath) {
  return /\.(md|html)$/i.test(relPath);
}
```

The docstring justifies the set as "markdown sources only (`.md`, `.html`)". HTML is not markdown, and a backtick in HTML is a literal character. Repro on the head checker:

- `<script>lines.push("see \`row A999\` here");</script>` scanned as `docs/x.html` → **0 errors** (blanked)
- the identical `lines.push(...)` line scanned as `scripts/x.mjs` → 1 error

All three tracked HTML files use backticks, and `docs/features/277-v115-bug-chore-sweep-board.html:1062` and `:1085` are backtick-delimited spans **inside JavaScript string literals in a `<script>` block** — structurally the same construct #3062 was filed about. No citation is hidden today, so nothing is broken yet; the class is armed and the stated rationale does not cover it. Either drop `.html` from the set or say in the comment why HTML is deliberately blanked despite not being markdown.

Adjacent, same helper: the extension test is exact-suffix, so `.mdx`, `.markdown` and `.htm` are all classified non-markdown. No such files are tracked today, so this is a note rather than a defect — worth one line in the docstring so the set reads as deliberate.

---

### 🟡 F4 — cleanup — one of eight threaded call sites is tested

`scripts/tests/check-register-citations.test.mjs:636-660`

The two new cases exercise `checkNonexistentIds` (Check A) only. The change threads `isMarkdown` through eight sites — `extractCitationsByLine`, `idSpecificAnnotationPresent`, `recordSubjectConflict` (1781), `checkConflictingSubjects` (1848), `measureWrongIdEligibleLines` (2025), `titleDriftTokens` (2123), `extractHeadingTitleEchoes` (2161), `dischargeAnnotationPresentAnywhere` (2204) — and none of the other seven has a paired assertion. F1 is one of the untested seven, which is how a one-token category error survived to head. `isMarkdownScanPath` itself also has no direct test, so the `.md`/`.html`-vs-everything-else boundary is asserted nowhere.

At minimum: a Check C case (`checkConflictingSubjects` on a `.mjs` path) and a Check D case pinning F1's fix.

---

### 🟡 F5 — cleanup — release-notes pair neither landed nor declared not-applicable

PR body / `docs/release-notes-next.md`

Before-shipping step 5 requires an entry in `docs/release-notes-next.md` + `RELEASE_NOTES.md`, or an explicit "not applicable" **stated** rather than silently omitted. The body has `## Summary`, `## Test plan`, `Closes #3062`, and no mention of release notes; the diff touches neither file.

This checker's own prior behaviour changes each landed one — `docs/release-notes-next.md:443` (#2721/#2846), `:444` (#2878), `:592` (#2603/#2840) — and this PR changes what the gate flags tree-wide, which is the same class of operator-visible delta. Either add the entry or say in the body why it is exempt.

---

### Gates checked and clean

- `Closes #3062` present, outside any code span.
- Paired regression test is real: red before, green after, red for the claimed reason (control measured above).
- Post-merge fatal-clean against current `main`; 146/146 tests pass on the merged tree.
- The two `A101` rewordings the widening surfaced are declared in the PR body, per the incidental-findings rule.
- No `cast.json` writes, no config knob, no derived artifact (`api-types.ts`, `BACKLOG.md`, brand PNGs) touched. On-box acceptance not applicable — no hardware-provable behaviour, no register row moved.

dudarenok-maker:
## PR review — pass 2 (head e7e468bf, depth medium)

**Depth re-derived from the current commit set**, not inherited from pass 1: `3cbc45f1 fix(scripts)`, `99b4c921 fix(scripts)`, `3bad5627 fix(scripts)`, `00c9dbf5 fix(scripts)`, `e7e468bf test(scripts)` — five commits, single scope (`scripts`), types `fix`/`test`. CLAUDE.md's ladder puts a single-scope `fix` at **medium**; the `test` commit does not raise it. Confirmed **medium**.

Every claim below is measured against a disposable clone of `refs/pull/3124/head` (verified `e7e468bfd51b0f…`), not against the primary checkout, which I did not modify.

---

**🟠 F1 — `titleDriftTokens(row.title, isMarkdown)` passed the scanned file's flag to a register-owned input — ✅ VERIFIED RESOLVED.**

`scripts/check-register-citations.mjs:2134` now reads `function titleDriftTokens(text, isMarkdown = true)`, and `:2286` calls `titleDriftTokens(row.title)` — the register is always markdown, matching the unconditional treatment already given to `stripInlineCodeSpans(rowBodyText)` at `:1473` and `stripFences(text)` at `:669` (`parseRegisterRows`, register input). I checked the concern you flagged explicitly: `titleDriftTokens` has exactly **two** callers (`:2282` `titleDriftTokens(titleEcho, isMarkdown)` and `:2286`), so no other caller silently changed meaning by acquiring the new default. The `:2282` call is correct — `titleEcho` came out of `extractHeadingTitleEchoes(text, isMarkdown)`, which already applied (or correctly skipped) the same blanking.

Mutation-verified red-before-green: reverting `:2286` to `titleDriftTokens(row.title, isMarkdown)` fails the new test at `scripts/tests/check-register-citations.test.mjs:2421` with `AssertionError: drift detected when scanning .mjs file (same as .md, not file-dependent)` — red for exactly the claimed reason, not incidentally.

**🟡 F2 — `stripFences` ran unconditionally on non-markdown files — ✅ VERIFIED RESOLVED, gating complete.**

`stripFences(text, { isMarkdown = true } = {})` at `:520`. I enumerated **all seven** call sites, not the six the PR body names: `:1229`, `:1262`, `:1860`, `:2038`, `:2173`, `:2280` all thread `{ isMarkdown }`; the seventh, `:669` inside `parseRegisterRows`, correctly keeps the `true` default because its input is the register itself. No site missed.

Mutation-verified: deleting the `if (!isMarkdown) return text;` guard at `:521` turns both new tests red (`check-register-citations.test.mjs:2634` and `:2646`) with `actual: 0, expected: 1` — the citation goes back to being invisible, which is the claimed failure.

**🟡 F3 — `.html` treated as fully markdown, `<script>`/`<style>` content included — ✅ RESOLVED, with one gap (see N2).**

The comment at `:1123-1129` is accurate on its own terms and its real-tree citation checks out: `docs/features/277-v115-bug-chore-sweep-board.html:1062` contains `` `blocked` `` and `:1085` contains `` `gh issue view --comments` ``, both inside `<script>` block JS string literals. The "could theoretically hide a citation there" hedge is right — neither carries a register ID today. Not an overclaim. It does underclaim in one respect — see N2.

**🟡 F4 — thin markdown-vs-non-markdown coverage — ⚠️ PARTIALLY RESOLVED. The judgment call was not reasonable in hindsight: the untested subset is exactly where the remaining defect sits (N1).**

The added Check C pair (`check-register-citations.test.mjs:1881` / `:1899`) is real and load-bearing — removing the `stripInlineCodeSpans` guard at `:1137` turns the non-markdown half red (`actual: 0, expected: 1`) while the markdown control stays green. Good test.

Where the flag is now covered: Check A (`:1261`), Check C (`:1859`), Check D (`:2279`, via F1's test), and transitively `extractCitationsByLine` (`:1228`) and `extractHeadingTitleEchoes` (`:2172`).

Where it is still uncovered: **`idSpecificAnnotationPresent` (`:1033`) and `dischargeAnnotationPresentAnywhere` (`:2211`) — the discharge-annotation/excuse paths**, plus `measureWrongIdEligibleLines` (`:2037`). Those are not an arbitrary remainder: they are the only threaded sites that decide whether a finding is *suppressed* rather than whether it is *seen*, i.e. the fail-open direction — and N1 below shows a proven behaviour regression sitting in exactly that gap. "Acceptable to triage by judgment" would have been defensible if the untested remainder were homogeneous with the tested part; it isn't.

**🟡 F5 — missing before-shipping declarations — ✅ VERIFIED RESOLVED, and the body matches the diff.**

`Closes #3062` present, outside any code span; #3062 is open, labelled `bug` + `area:ops`; the `pr-issue-link` check is green. Release-notes and on-box-acceptance not-applicable declarations both present and both correct (dev-tooling checker, no product delta, no hardware-provable behaviour). Cross-checked the body against the actual four-file diff: all four incidental fixes are declared and all four are genuinely in the diff; the `validate-sidecar-acceptance.mjs` / `.test.mjs` rewords are declared in the Summary. Nothing undeclared, nothing declared-but-absent.

Test-plan claims independently reproduced on the PR head: `check-register-citations.test.mjs` **150/150**, `validate-sidecar-acceptance.test.mjs` **88/88**, `node scripts/check-register-citations.mjs --strict` → `check:register-citations: OK`, exit 0.

Two body-accuracy notes, neither a finding: (a) F2's bullet says "6 call sites"; there are 7, the 7th correctly defaulted — the count is of *threaded* sites, which is fine but reads as an enumeration; (b) F1's bullet says the bug was "silently suppressing title-drift detection for 9 backticked-title rows (A17, A20, …)". The *mechanism* claim is right (I confirmed the direction is suppression), but the count is unmeasured: running the checker over the real tree with the F1 bug reintroduced produces a **byte-identical** report, so zero of those 9 rows actually had a finding suppressed today. It reads as nine lost findings; it is nine rows whose detection was degraded, with no measured loss.

---

### New findings

#### 🟠 N1 — correctness. The same `isMarkdown` threading also *loosened* the discharge-annotation scan, so a discharge word inside a backtick span can now disarm two FATAL checks in any non-markdown file. Undeclared, untested, and it reverses a defect this file already fixed once.

`scripts/check-register-citations.mjs:1033-1034` (`idSpecificAnnotationPresent`), reached from `:1269` (Check A, fatal), `:1792` (Check C, fatal/non-fatal branch), `:2301` (Check D); same shape at `:2211-2217` (`dischargeAnnotationPresentAnywhere`), reached from `:2300`.

The blanking at `:1034` is not there for the same reason as the citation scan. Its own docstring at `:1008-1023` records why it exists — PR #2630 pass-8 finding O:

> a real-tree paired injection showed a discharge word sitting inside an example command (`` `grep "register row A99 was discharged" docs/` ``) disarming the check, even though the text is an instruction to run a search, not an assertion about anything.

That rationale is **not markdown-specific**. A `.mjs`/`.ts`/`.py` file is *more* likely to hold a quoted shell command than a markdown doc is, and `execFileSync(`grep "register row A99 was discharged" docs/`)` is the same non-assertion. Threading `isMarkdown` in here was mechanical, not decided — #3062's scope call was about citations being *hidden*, and says nothing about what may *excuse* one.

Measured, not reasoned. Same input text, same register, `checkNonexistentIds` only:

```
text = 'See register row A999 — `discharged` in the 2026-08 sweep.\n'

origin/main's checker:   docs/foo.md → errors=1   scripts/foo.mjs → errors=1
this PR's checker:       docs/foo.md → errors=1   scripts/foo.mjs → errors=0, annotated=1
```

A citation `main` fails on is silently excused at head, purely because the file extension is `.mjs`. The citation itself sits outside the backticks — only the discharge word is inside — so this is not the intended widening, it is its mirror image. At `:1792` the same helper picks between `annotatedDischarge` (non-fatal) and `wrongId` (**fatal**), so Check C is weakened the same way.

Note the resulting incoherence *within one file*: three discharge-word scans now disagree about what a backtick means — `:1034` and `:2217` gate on the scanned file's extension, `:1473` (`dischargedSubjectsMentionedIn`) is unconditional. And the same author writing the same sentence gets a fatal citation in `docs/foo.md` and an excused one in `scripts/foo.mjs`.

Zero real-tree instances today (I diffed `origin/main`'s checker against this PR's checker over the full current tree — identical output, 137 lines each), so this is a latent gate weakening, not an active one. That does not downgrade it: the gate exists for the file that has not been written yet, and this is a documented defect being re-armed for half the corpus.

**Suggested fix**: drop the `isMarkdown` parameter from `idSpecificAnnotationPresent` and `dischargeAnnotationPresentAnywhere` and let their `stripInlineCodeSpans` calls stay unconditional (matching `:1473`), which restores finding O's protection tree-wide and keeps the PR's actual intent — the *citation* scan — untouched. Paired test: the probe above, both directions, as a `.md`/`.mjs` pair. If instead the widening is deliberate, it needs to be said out loud in the header comment at `:39-45` (which today claims only the tightening: "so a template literal's backticks can never silently hide a citation") and pinned by a test — but I'd want the repo owner's call on that, since it re-opens a defect they already paid to close.

#### 🟡 N2 — cleanup. F3's comment names only inline code spans, but the predicate it sits on now gates fence blanking too.

`scripts/check-register-citations.mjs:1123-1129`. The comment was written to record the accepted `.html` trade-off and says "backtick-wrapped spans inside those blocks are also blanked". Since `00c9dbf5`, `isMarkdownScanPath` also gates `stripFences` (`:520`), whose failure mode is categorically larger: a single unpaired ` ``` ` inside an `.html` `<script>` blanks **to the next fence marker or EOF**, not one span. A reader of this comment comes away believing the worst case is one hidden span. No tracked `.html` file contains a triple backtick today (I checked all three), so there is no live instance — it is a comment the F2 fix made incomplete, one clause short.

---

### Checked and explicitly *not* findings

Recording these so a later pass doesn't re-spend the time:

- **A 9th same-class call site.** I enumerated every markdown-shaped transform applied to scanned text. `deBold` (`:1195`) and `stripMarkdownLinkUrls` (`:2130`) are still unconditional on non-markdown files, but neither can hide or manufacture a citation — `deBold` only removes `**` (it cannot delete ID characters, and `row **A9**` → `row A9` preserves intent), and `stripMarkdownLinkUrls` is applied symmetrically to both sides of the drift comparison. No failure scenario. `parseRegisterRows` (`:669`) and `dischargedSubjectsMentionedIn` (`:1473`) are unconditional *correctly* — both take register-owned input.
- **Extension enumeration.** `/\.(md|html)$/i` misses `.htm` and `.markdown`; neither exists in the tree, and the miss direction is fail-loud (scan raw → more citations surfaced), not fail-open.
- **Path shape.** `isMarkdownScanPath` receives `git ls-files` repo-relative paths at `:2430`/`:2433` and `filePath` map keys at `:1859`/`:2037`/`:1792` — all consistent, no absolute/relative mismatch.
- **Real-tree effect.** Running `origin/main`'s checker and this PR's checker over the *same* tree produces byte-identical output. The PR is a pure hardening today; nothing newly caught, nothing newly excused.
- **`stripFences` gating changes nothing observable today either** — head vs. head-with-the-guard-removed produced identical real-tree output.
- **Merge safety, given the branch is 162 commits behind `origin/main`.** `main` has since touched this very file, so I checked rather than assumed: its only change since the merge-base (`8e4025ee`) is adding `scripts/tests/build-register-live-view.test.mjs` to `SELF_REFERENTIAL_PATHS` plus a paired mutation test — **no new `stripInlineCodeSpans`/`stripFences` call sites**, so the merge cannot land an ungated one. I performed the merge locally: clean, no conflicts, `check-register-citations.mjs --strict` exits 0 on the merged tree, and the suite is **151/151** (150 + main's new mutation test). Worth knowing: reviewing the PR head *in isolation* shows 17 spurious Check D drift findings on `build-register-live-view.test.mjs`; those are a stale-branch artifact of the missing 6th exclusion, and they vanish on merge. Not a defect in this PR.

---

### Re-review trigger

**Triggered.** N1 is a 🔴/🟠-class correctness finding, so fixing and pushing it re-opens the loop — pass 3 would be the second of the two permitted re-review rounds. N2 is 🟡 and does not itself trigger, but is still fixed this round per the report-fix-record rule.

**Not clear to merge as it stands.**

dudarenok-maker:
## PR review — pass 3 (head 3d8745fc, depth medium)

Final allowed round under the loop cap (initial + 2 re-reviews). Reviewed from an isolated shallow clone of the PR ref at `3d8745fc`; the primary checkout was not touched.

**Verdict: no 🔴/🟠 correctness findings. Two 🟡 cleanup findings. The loop cap is NOT tripped — fixing these does not re-trigger a pass, and this round closes the gate.**

---

| Finding | Status | How verified |
|---|---|---|
| **F1** (🟠 register title blanked by scanned file type) | **VERIFIED RESOLVED** | Re-derived red-before independently: reverting `titleDriftTokens(row.title)` → `titleDriftTokens(row.title, isMarkdown)` in a scratch copy fails the paired test with `drift detected when scanning .mjs file (same as .md, not file-dependent)`. Green at head. |
| **F2** (🟡 `stripFences` not gated) | **VERIFIED RESOLVED** | Re-derived red-before: neutering `if (!isMarkdown) return text;` in `stripFences` fails **both** paired tests (fenced citation in `.mjs`; unpaired fence not blanking to EOF). All 7 `stripFences` sites accounted for — 6 gated, 1 (`parseRegisterRows:666`) correctly unconditional because its input is the register itself. |
| **F3** (🟡 `.html` residual-risk note) | **VERIFIED RESOLVED** | Comment present at `scripts/check-register-citations.mjs:1120-1128`. Accurate for the `.html` claim it makes — but see 🟡 **P1** below, which is about a *different* clause N2 later appended to the same block. |
| **F4** (🟡 missing paired coverage for `checkConflictingSubjects`) | **VERIFIED RESOLVED** | Both the `.ts` case and the `.md` control are present and genuinely distinguish the two paths. |
| **N1** (🟠 discharge-annotation blanking wrongly gated) | **VERIFIED RESOLVED** | Re-derived red-before rather than trusting the report: reverting both call sites to `stripInlineCodeSpans(…, { isMarkdown })` fails the paired test **for exactly the claimed reason** — `AssertionError: non-markdown should also report A999 as fatal / actual: 0, expected: 1`, with the `.md` half still passing. Green at head. |
| **N2** (🟡 stale `isMarkdownScanPath` comment) | **PARTIALLY RESOLVED — see 🟡 P1** | The `stripFences` half was added correctly. The clause added about `stripInlineCodeSpans` states the opposite of what the code does, and the JSDoc block N2 was raised against is still unrepaired. |

Test plan re-run independently in the clone, all figures in the PR body confirmed:
- `node --test scripts/tests/check-register-citations.test.mjs` → **152/152 pass** ✅ (body says 152)
- `node --test scripts/tests/validate-sidecar-acceptance.test.mjs` → **88/88 pass** ✅ (body says 88)
- `node scripts/check-register-citations.mjs --strict` → **exit 0**, 131 lines of output ✅

---

### The fourth-bug hunt

Given three `isMarkdown`-threading bugs across two rounds, I enumerated **every** blanking site and classified each by the mechanism's own semantics — "normalise markdown syntax" (must be gated) vs. "prevent a false positive/negative for a reason unrelated to markdown syntax" (must be unconditional) — rather than pattern-matching on which sites were already touched.

**`stripInlineCodeSpans` — 8 call sites, 5 gated / 3 unconditional:**

| Line | Site | State | Correct? |
|---|---|---|---|
| 1031 | `idSpecificAnnotationPresent` | unconditional | ✅ discharge-word scan; finding-O rationale is idiom-based (a code span is an instruction, not an assertion) and holds in any file type |
| 1229 | `extractCitationsByLine` | gated | ✅ the PR's headline change |
| 1472 | `dischargedSubjectsMentionedIn` | unconditional | ✅ input is register body text, always markdown |
| 1859 | `checkConflictingSubjects` | gated | ✅ scanned-file text |
| 2037 | `measureWrongIdEligibleLines` | gated | ✅ and byte-identical to 1859, preserving the denominator-must-match invariant its own comment demands |
| 2135 | `titleDriftTokens` | gated by param | ✅ |
| 2173 | `extractHeadingTitleEchoes` | gated | ✅ scanned-file text |
| 2216 | `dischargeAnnotationPresentAnywhere` | unconditional | ✅ same as 1031 |

**`stripFences` — 7 call sites, 6 gated / 1 unconditional (666, the register). ✅**

The resulting rule is coherent and every site conforms: **register-sourced text → always markdown; scanned-file text → gated; discharge-word scans → always blanked.** I found no site on the wrong side of it, and no site that should have the flag threaded but doesn't (`checkRunSheetLinkage` / `findUnclassifiedRunSheetMentions` do no blanking at all and read only `*-onbox-acceptance.md` run sheets).

I also confirmed empirically that reverting each fix independently (`isMarkdownScanPath → true`, `stripFences` unconditional, N1 reverted, F1 reverted) produces **byte-identical real-tree checker output** to head — i.e. all three are latent with zero live instances, matching the PR body's claim for N1 and consistent with the two `A101` citations having been reworded away in commit `6e2f8de`.

---

### New findings

#### 🟡 P1 — cleanup (chore): N2's new comment states the *opposite* of what the code does, and contradicts two other comments in the same file

`scripts/check-register-citations.mjs:1122-1125`:

```
// embedded `<script>` and `<style>` block content — both triple-backtick
// `stripFences` and single-backtick `stripInlineCodeSpans` blanking apply
// (the latter unconditionally for discharge-annotation false-positive prevention,
// the former gated by isMarkdown), and backtick-wrapped spans inside those blocks
```

"the latter [`stripInlineCodeSpans`] unconditionally" is false as a description of what `isMarkdownScanPath` gates. Per the table above, `isMarkdownScanPath` gates `stripInlineCodeSpans` at **5 of its 8 call sites** — including the PR's entire headline change (`:1229`, `:1859`, `:2037`, `:2135`, `:2173`). Only the two discharge-scan sites (`:1031`, `:2216`) and the register-body site (`:1472`) are unconditional.

It also directly contradicts the JSDoc three lines above it (`:1110-1113` — "should have its single-backtick inline code spans blanked as markdown. Blanking is scoped to markdown sources only") and the file header (`:38-42`).

**Concrete failure scenario** — not hypothetical for this file: an agent briefed to "make the blanking consistent with its own documentation" reads `stripInlineCodeSpans … unconditionally`, greps the five `{ isMarkdown }` sites as non-conforming, and removes the gate — reverting #3062's fix and re-arming the exact bug this PR exists to close. Three threading bugs have already been produced here in two rounds, each a mis-read of which mechanism the flag controls; this comment now actively teaches that mis-read.

Two smaller parts of the same finding:
- `isMarkdownScanPath` now carries **two stacked comment blocks** — a JSDoc at `:1109-1119` and a `//` block at `:1120-1128` — whose opening sentences restate each other.
- N2's edit landed on the `//` block, so the **JSDoc half is still stale in exactly the way N2 was raised to fix**: it describes only inline-code-span gating and never mentions `stripFences`.

Suggested shape: one block that says `isMarkdownScanPath` gates **both** `stripFences` and the general citation-scanning `stripInlineCodeSpans` calls, and that the **discharge-word scans** (`idSpecificAnnotationPresent`, `dischargeAnnotationPresentAnywhere`) and the **register-body scan** deliberately bypass the flag — with the reason, so the next reader cannot re-derive N1.

#### 🟡 P2 — cleanup (dead code): the N1 fix left `_isMarkdown` as a dead parameter, and four call sites still compute and pass a discarded argument

`scripts/check-register-citations.mjs:1030` and `:2210` now take `_isMarkdown` and never read it. Four call sites still pass a third argument into that dead slot:

- `:1268` — `idSpecificAnnotationPresent(enclosingSectionText(lines, i), id, isMarkdown)`
- `:1791` — `idSpecificAnnotationPresent(enclosingSectionText(lines, lineIndex), id, isMarkdownScanPath(filePath))` — this one **calls `isMarkdownScanPath(filePath)` purely to feed the discarded slot**
- `:2299` — `dischargeAnnotationPresentAnywhere(sectionText, id, isMarkdown)`
- `:2300` — `idSpecificAnnotationPresent(sectionText, id, isMarkdown)`

No wrong output — this is residue of N1 rather than its removal. But a reader at any of those four call sites sees `isMarkdown` being threaded into a discharge check and reasonably concludes the check honours it, which is precisely the misreading that produced N1. The underscore satisfies `eslint.config.mjs:128`'s `argsIgnorePattern: '^_'` (silencing the linter), not the reader. Dropping the parameter and the four arguments removes the trap.

---

### Examined and deliberately NOT raised as findings

Recording these so a future round does not "discover" them and reopen a closed question.

1. **Fence/discharge asymmetry introduced by F2.** Reproduced: a discharge word inside an unindented triple-backtick block with the citation outside it yields `errors=1` in `docs/foo.md` but `errors=0, annotated=2` in `scripts/foo.mjs`. This is *not* a bug — it is the #3062 decision working as specified. In `.mjs` the fence markers are inert characters, so citation **and** annotation are both read raw; the treatment is internally consistent, and removing the fences entirely from the `.mjs` input produces the same result. `stripFences`'s sole stated purpose ("an example heading inside a fence can't be mistaken for a real one") is markdown-syntax-specific, unlike `stripInlineCodeSpans`'s idiom-based discharge rationale, which is why N1 was a bug and this is not.

2. **`titleDriftTokens` prose side still gated while the register side (post-F1) is not.** Deliberate and correct: the prose echo *is* the scanned file's own text, so gating it on that file's type is the #3062 decision; the register title is not, so it must never depend on scan type. F1's argument is precise and complete.

3. **`isMarkdown = true` defaulting toward blanking** in both `stripFences` and `stripInlineCodeSpans` means a future call site that forgets the flag silently gets the citation-hiding behaviour — which is literally how F2 happened. Not raised: five sites legitimately rely on that default, and F1's own fix deliberately *uses* it as the "register is always markdown" idiom, so making the flag required is a design preference with more than one defensible outcome, not a defect at head.

4. **`Closes #3124 finding F1` in commit `9a73b5c`'s body.** Checked against the `r_commit_trailer_beats_pr_body` precedent: `gh pr view 3124 --json closingIssuesReferences` returns **only #3062**, so the self-referential trailer created no spurious linkage and does not displace the intended close.

5. **The `A101` rewording in commit `6e2f8de`.** Both changed lines are `//` comments describing the help text, not the help text itself — no behavioural surface. `grep A101` confirms no remaining occurrence sits on a citation surface in either file, and the 88/88 sidecar-acceptance suite is green.

---

### House gates

- Paired tests: real regression tests, red-before re-derived independently for F1, F2 and N1 ✅
- `Closes #3062` present, outside any code span, and confirmed as the PR's only closing reference ✅
- On-box acceptance: explicitly N/A — nothing hardware-provable here ✅
- Release notes: explicitly skipped as internal dev-tooling with no user/operator-visible delta ✅ (reasonable)
- Incidental findings F1-F4 and N1-N2 all fixed in-round and declared under "Also fixed, found in passing" ✅
- Test-plan figures in the body match reality (152, 88, strict exit 0) ✅
- No `cast.json` writes, no config knob, no derived artifact touched — N/A ✅

Neither P1 nor P2 is a correctness bug, so per findings-triage they are fixed in this round but do **not** re-trigger a review pass. Once they land, this PR is clear to merge.

Issue #3122, opened 2026-09-08
check:register-citations has no CI wiring at all — a broken citation only surfaces on manual run


dudarenok-maker:
## What

`scripts/check-register-citations.mjs` (the checker for
`docs/testing/onbox-acceptance-register.md` row-ID citations scattered
across the repo) is a real, well-tested script (144 tests as of PR #3113)
with a real npm script (`check:register-citations`) — but no CI workflow,
no `test:all`/`verify` step, and no git hook ever runs it. Its sibling
checks (`check:onbox-register`, `register:build --check`) ARE wired into
`.github/workflows/` and CI's "Check on-box acceptance register
consistency" leg; this one is manual-only.

Found in PR #3113's review (2026-09-09): a discharge in that PR broke real
citations elsewhere in the tree (a stale mention of a now-removed row), and
nothing but a human running `npm run check:register-citations` by hand would
have caught it. The fix landed in the same PR, but only because the reviewer
happened to run the script — there's no mechanical backstop.

## What's owed

A design decision, not a one-line fix:

- **Which CI leg it joins** — a new dedicated leg, or folded into the
  existing "Check on-box acceptance register consistency" job.
- **Scope-gating** — should it run on every PR, or only PRs touching
  `docs/testing/onbox-acceptance-register.md` / files it cites (the same
  `git diff` scope-gating `verify.yml`'s other legs already use)? The
  checker scans the WHOLE repo tree for citations, not just changed files,
  so a PR that discharges a row but doesn't otherwise look
  register-related could still break a citation anywhere.
- **Cost** — the CLI-mutation test suite alone took ~33s locally; the real
  full-tree scan itself is comparatively cheap, but this needs measuring
  before picking a gating strategy (a Ubuntu runner's cold-cache time may
  differ from a warm local run).
- **Docs-only fast-path interaction** — should this run on the doc-only PR
  fast path (CONTRIBUTING.md), since a docs-only PR is exactly the shape
  most likely to discharge a register row and break a citation elsewhere?

Refs #3113 (where the gap was found and worked around by hand), #2629
(the checker's original issue).

PR #3134, opened 2026-09-09
chore(ops): wire check:register-citations into verify.yml

dudarenok-maker:
## Summary
- Adds an unconditional `check:register-citations` step to verify.yml's
  lint-and-checks job, closing the gap where a broken register-row citation
  outside test:hooks' own scope (docs/features/**, src/**, server/**, e2e/**)
  was never caught by CI.

## Test plan
- [x] New step runs `npm run check:register-citations`, no `if:` scope-gate
- [x] Verified NOT folded into the path-filtered onbox-register-check.yml
- [x] Mutation: broke a real citation (pointed it at a nonexistent register
      row), confirmed check:register-citations catches it, reverted, clean
- [x] npm run check:register-citations passes on the tree as committed

## Not applicable
- **Release notes (before-shipping step 5)**: skipped — CI-only change, no
  shippable user- or operator-visible delta.
- **On-box acceptance (before-shipping step 3)**: skipped — this wires an
  existing, already-tested script into CI; nothing here requires real
  hardware to prove.

Refs #3122 — this PR closes the CI-workflow gap #3122 named, but #3122 also
named a `test:all`/`verify` step and a git hook as unwired surfaces; those
carry a real design decision (how an always-run, whole-tree-scanning check
fits `verify-cache.mjs`'s scope-gated-and-cached `STEPS[]` convention, and
ops-2997's "hooks run no battery" shape) and are tracked separately in #3140
rather than decided here.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -63,12 +63,16 @@ name: Verify
 #
 # No `paths-ignore` here (deliberately removed — see design spec): keeping it
 # while this is a required status check would deadlock every docs-only PR,
-# since a required check that never fires blocks merge forever. Every leg's
+# since a required check that never fires blocks merge forever. Most legs'
 # own `if:` scope condition already evaluates false for a docs-only diff, so
-# every job still completes in roughly the time of "Setup Node + deps" alone
-# and the aggregator reports green — no deadlock, negligible added time. Same
-# reasoning is why no job gates on PR `draft` status either — a draft-gated
-# required check would deadlock every draft PR the same way.
+# most jobs still complete in roughly the time of "Setup Node + deps" alone
+# and the aggregator reports green — no deadlock, negligible added time.
+# ONE exception: `check:register-citations` (see #3122) runs unconditionally,
+# since citations can appear in any file and diff-scope cannot predict whether
+# one broke; this step still runs on docs-only diffs, but remains negligible
+# (a few seconds), so overall timing is unaffected. Same reasoning is why no
+# job gates on PR `draft` status either — a draft-gated required check would
+# deadlock every draft PR the same way.
 
 on:
   pull_request:
@@ -215,8 +219,8 @@ jobs:
         if: fromJSON(needs.detect.outputs.scopes).step_check_budget_poll || fromJSON(needs.detect.outputs.scopes).shared
         run: npm run check:budget-poll
 
-      # Register-citation check (unconditional — see #3122: the checker
-      # scans the WHOLE tree for citations, so no diff-scope reliably
+      # Register-citation check (unconditional — see #3122 and #3140): the
+      # checker scans the WHOLE tree for citations, so no diff-scope reliably
       # predicts whether one broke; must run on every PR, docs-only included.
       # leg: check:register-citations
       - name: Register citation check
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -222,7 +222,7 @@ jobs:
       # Register-citation check (unconditional — see #3122 and #3140): the
       # checker scans the WHOLE tree for citations, so no diff-scope reliably
       # predicts whether one broke; must run on every PR, docs-only included.
-      # leg: check:register-citations
+      # (No scope gate: dedicated guard in workflow-wiring.test.mjs, not via parseLegs())
       - name: Register citation check
         run: npm run check:register-citations
 
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -215,6 +215,13 @@ jobs:
         if: fromJSON(needs.detect.outputs.scopes).step_check_budget_poll || fromJSON(needs.detect.outputs.scopes).shared
         run: npm run check:budget-poll
 
+      # Register-citation check (unconditional — see #3122: the checker
+      # scans the WHOLE tree for citations, so no diff-scope reliably
+      # predicts whether one broke; must run on every PR, docs-only included.
+      # leg: check:register-citations
+      - name: Register citation check
+        run: npm run check:register-citations
+
       # #2434 → #2779: npm audit against the root lockfile, gated so it only
       # runs on a root package-lock.json diff (or a shared/global change).
       # leg: audit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1013,8 +1013,9 @@ OS blocks the public-beta release, so you no longer fire cross-OS by hand
 before a release. `cross-os.yml` (`workflow_dispatch` + twice-weekly cron on
 `main`) stays as the between-releases pulse + ad-hoc cross-OS/mobile run.
 Docs-only PRs still complete `verify.yml` in seconds rather than deadlocking
-the required check — every leg's own scope condition is false for a
-docs-only diff, so the job just does env setup and reports green (see
+the required check — most legs' scope conditions evaluate false for a
+docs-only diff; one step (`check:register-citations`, PR #3134) runs
+unconditionally. Either way, the job completes in seconds and reports green (see
 [docs/superpowers/specs/2026-07-06-verify-ci-rebalance-design.md](docs/superpowers/specs/2026-07-06-verify-ci-rebalance-design.md)
 for why `paths-ignore` was deliberately removed rather than kept).
 See [docs/features/215-ci-label-gated-verify.md](docs/features/215-ci-label-gated-verify.md)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -576,9 +576,11 @@ A PR whose changed-file set lives entirely under `docs/**`, root-level
 deliberately removed — see
 [docs/superpowers/specs/2026-07-06-verify-ci-rebalance-design.md](docs/superpowers/specs/2026-07-06-verify-ci-rebalance-design.md)
 for why: keeping it while the check is required would deadlock every
-docs-only PR forever). Every leg's own scope condition evaluates false for
-a docs-only diff, though, so the job completes in roughly the time of "Setup
-Node + deps" alone (~20-40s) and reports green — the gate stays "PR required
+docs-only PR forever). Most legs' own scope conditions evaluate false for a
+docs-only diff, though; one step (`check:register-citations`, added via PR
+#3134) runs unconditionally. Either way, the job still completes in roughly
+the time of "Setup Node + deps" alone (~20-40s) and reports green — the gate
+stays "PR required
 + title valid + no conflicts + a fast green required check", not the 10-15
 min full battery. Rationale and the exact glob list for the underlying
 scope-matching (unrelated to the removed `paths-ignore`):
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -580,10 +580,9 @@ docs-only PR forever). Most legs' own scope conditions evaluate false for a
 docs-only diff, though; one step (`check:register-citations`, added via PR
 #3134) runs unconditionally. Either way, the job still completes in roughly
 the time of "Setup Node + deps" alone (~20-40s) and reports green — the gate
-stays "PR required
-+ title valid + no conflicts + a fast green required check", not the 10-15
-min full battery. Rationale and the exact glob list for the underlying
-scope-matching (unrelated to the removed `paths-ignore`):
+stays "PR required + title valid + no conflicts + a fast green required check",
+not the 10-15 min full battery. Rationale and the exact glob list for the
+underlying scope-matching (unrelated to the removed `paths-ignore`):
 [docs/features/archive/101-docs-only-ci-skip.md](docs/features/archive/101-docs-only-ci-skip.md).
 
 The same file-set test also skips the **local** pre-push `test:sidecar`
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -16,23 +16,21 @@
 // (1) `scripts/tests/check-register-citations.test.mjs`'s CLI-integration
 // tests, run as part of `npm run test:hooks`; and
 // (2) `.github/workflows/verify.yml`'s unconditional `check:register-citations`
-// step (wired in PR #3134, closing #3122).
+// step (PR #3134 closes the CI-wiring gap at #3122; other surfaces remain at #3140).
 //
-// CI WIRING (closed): The dedicated `.github/workflows/verify.yml` step now
-// exists (see `.github/workflows/verify.yml` lines ~218-223); it runs
-// unconditionally on every PR, including docs-only diffs, because citations
-// can live in any file and diff-scope cannot reliably predict whether one
-// broke.
+// CI WIRING (closed): The dedicated `.github/workflows/verify.yml` "Register
+// citation check" step now exists; it runs unconditionally on every PR,
+// including docs-only diffs, because citations can live in any file and
+// diff-scope cannot reliably predict whether one broke.
 //
-// LOCAL WIRING (open, tracked at #3140): The checker is still NOT invoked
-// from `npm run verify` or `npm run test:all` locally, or from any git hook.
-// Those remaining integration points are tracked separately at #3140.
-// The challenge: this checker's real-tree run scans essentially every tracked
-// file, so declaring it as a `test:hooks` input would make that step
-// un-cacheable for everyone, defeating the scope-gating `verify-cache.mjs`
-// exists for. The right fix is a design decision about whether to add a
-// dedicated local step, and if so, whether it belongs in `verify` or wrapped
-// elsewhere — not something to wire in blind here.
+// LOCAL WIRING (open, tracked at #3140): The checker IS reachable locally as
+// part of `npm run test:hooks` (which `npm run verify` and `npm run test:all`
+// both run), via the test suite `scripts/tests/check-register-citations.test.mjs`
+// that spawns the checker CLI against the real tree. However, `test:hooks`'s own
+// input globs don't span the whole tree, so on a diff outside those globs, the
+// step can be marked `[cached]`/skipped rather than actually re-scanning. A
+// dedicated local step that runs unconditionally (independent of scope) is
+// tracked separately at #3140.
 //
 // Four checks, ordered by precision (least to most likely to need
 // judgment):
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -11,30 +11,28 @@
 // claim living on several surfaces, with only some of them corrected. This
 // script exists to make that mechanical instead of eyeballed.
 //
-// WIRING GAP, stated explicitly rather than left implicit (pass-9 review of
-// PR #2630, finding AA): `package.json`'s `check:register-citations` script
-// is invoked from exactly one place today —
-// `scripts/tests/check-register-citations.test.mjs`'s own CLI-integration
-// tests, run as part of `npm run test:hooks`. That means this checker only
-// actually EXERCISES on a diff `verify-cache.mjs`'s `test:hooks` step
-// considers in-scope: `docs/testing/**`, the register itself, `CLAUDE.md`,
-// and `scripts/**` — NOT `docs/features/**`, `docs/superpowers/**`,
-// `src/**`, `server/**`, or `e2e/**`, even though a citation can live in any
-// of those and this checker's own real-tree run scans every one of them.
-// There is no dedicated `.github/workflows/*.yml` step for this checker the
-// way the sibling `check-onbox-register.mjs` has
-// (`onbox-register-check.yml`) — `#2629`'s option 3 ("catches rot at PR
-// time") is not fully true yet: rot in a file outside `test:hooks`' own
-// scope is caught only the NEXT time some in-scope file changes too, or on
-// a manual `npm run check:register-citations`. Widening `test:hooks`'
-// inputs to the whole tree isn't the fix — this checker's own real-tree run
-// reads essentially every tracked file, so declaring that as a `test:hooks`
-// input would make the step un-cacheable for everyone, defeating the
-// scope-gating `verify-cache.mjs` exists for. The right fix is a dedicated
-// CI step (mirroring `onbox-register-check.yml`) that always runs this
-// checker regardless of diff scope — a genuine design decision (schedule,
-// gating, whether it belongs in `verify.yml` or its own workflow), not
-// something to wire in blind here; tracked at `#2721`.
+// WIRING STATUS (updated 2026-09-10, PR #3134): `package.json`'s
+// `check:register-citations` script is now invoked from TWO places:
+// (1) `scripts/tests/check-register-citations.test.mjs`'s CLI-integration
+// tests, run as part of `npm run test:hooks`; and
+// (2) `.github/workflows/verify.yml`'s unconditional `check:register-citations`
+// step (wired in PR #3134, closing #3122).
+//
+// CI WIRING (closed): The dedicated `.github/workflows/verify.yml` step now
+// exists (see `.github/workflows/verify.yml` lines ~218-223); it runs
+// unconditionally on every PR, including docs-only diffs, because citations
+// can live in any file and diff-scope cannot reliably predict whether one
+// broke.
+//
+// LOCAL WIRING (open, tracked at #3140): The checker is still NOT invoked
+// from `npm run verify` or `npm run test:all` locally, or from any git hook.
+// Those remaining integration points are tracked separately at #3140.
+// The challenge: this checker's real-tree run scans essentially every tracked
+// file, so declaring it as a `test:hooks` input would make that step
+// un-cacheable for everyone, defeating the scope-gating `verify-cache.mjs`
+// exists for. The right fix is a design decision about whether to add a
+// dedicated local step, and if so, whether it belongs in `verify` or wrapped
+// elsewhere — not something to wire in blind here.
 //
 // Four checks, ordered by precision (least to most likely to need
 // judgment):
--- a/scripts/check-register-citations.mjs
+++ b/scripts/check-register-citations.mjs
@@ -23,14 +23,15 @@
 // including docs-only diffs, because citations can live in any file and
 // diff-scope cannot reliably predict whether one broke.
 //
-// LOCAL WIRING (open, tracked at #3140): The checker IS reachable locally as
-// part of `npm run test:hooks` (which `npm run verify` and `npm run test:all`
-// both run), via the test suite `scripts/tests/check-register-citations.test.mjs`
-// that spawns the checker CLI against the real tree. However, `test:hooks`'s own
-// input globs don't span the whole tree, so on a diff outside those globs, the
-// step can be marked `[cached]`/skipped rather than actually re-scanning. A
-// dedicated local step that runs unconditionally (independent of scope) is
-// tracked separately at #3140.
+// LOCAL WIRING (open, tracked at #3140): The checker IS reachable locally via
+// two paths: (1) `npm run verify` reaches it via `test:hooks` (which is scope-gated
+// in verify-cache.mjs, so it can be marked `[cached]`/skipped on out-of-scope diffs);
+// and (2) `npm run test:all` and `npm run verify:quick` invoke test:hooks directly
+// with NO caching, so the checker runs unconditionally as part of every local
+// `test:all`/`verify:quick` invocation. The actual decision at #3140 is narrower:
+// whether `npm run verify` itself (beyond test:all) should also have an
+// unconditional/uncached local leg (independent of scope-gating), and whether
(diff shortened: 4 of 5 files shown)


dudarenok-maker:
## PR review — pass 1 (head bda4a2aa1d4f1f5b882a4e267631c98d14fe263b, depth low)

Mandatory pre-merge gate. Depth `low` per CLAUDE.md's review-depth ladder (single-scope `ci`/`chore(ops)` PR, one file, +7/-0).

### What checks out

- The step carries **no `if:`**, and `lint-and-checks` itself has no job-level `if:` (only `needs: detect`) — so it genuinely runs on every PR including docs-only. The PR's stated purpose is actually delivered.
- Position is right: after `actions/checkout` (`fetch-depth: 0`) and `./.github/actions/setup`, whose `npm ci` is unconditional-on-cache-miss, so node + deps are present.
- Failure propagates: `scripts/check-register-citations.mjs:2523` throws `CliExitError(1)` → `process.exitCode = 1` (`:2531`) → default `bash -e` `run:` fails the step → job → `verify` aggregator (`lint-and-checks` is in its `needs:` list). Not a fail-open.
- **No conflict/duplication with `onbox-register-check.yml`** — that workflow runs `check-onbox-register.mjs` + `build-register-live-view.mjs --check`, never this script. The PR body's claim is accurate.
- No existing guard breaks. `workflow-wiring.test.mjs`'s `parseLegs` requires an `if:` after the `- name:`, so the new `# leg:` marker parses as nothing and `legs.size >= 11` is unaffected; `scanDerivedSteps` only inspects steps whose `if:` references the derived scopes; `ci-scope.mjs` emits no new key, so the `->`/`<-` assertions are untouched.
- Ran `node scripts/check-register-citations.mjs` against `main` (`d3c3d904`): **EXIT=0**, 3630 files scanned. The step will not redden every PR on day one.

---

### 🟠 1. Nothing mechanically stops the `if:` from being added back — the PR's entire promise is untested

`.github/workflows/verify.yml:218-223` (new step) · `scripts/tests/workflow-wiring.test.mjs:272-302` (`REQUIRED_LEGS`), `:363-395` (`scanDerivedSteps`)

The whole value of this change is the *absence* of a scope gate, and no assertion anywhere pins that absence.

Concrete scenario: a later cost-trimming pass on `verify.yml` (exactly the kind recorded in `docs/features/118-ci-cost-round-2.md`) adds one line to this step:

```yaml
      # leg: check:register-citations
      - name: Register citation check
        if: fromJSON(needs.detect.outputs.scopes).step_test_hooks || fromJSON(needs.detect.outputs.scopes).shared
        run: npm run check:register-citations
```

Every assertion in `workflow-wiring.test.mjs` still passes:

- `scanDerivedSteps` requires `- name:` on the line above the `if:` and a `MARKER_LINE` two above — both are present (`check:register-citations` is inside the `[a-z:0-9-]+` grammar), so the "every derived step carries a parseable marker" test is satisfied;
- `REQUIRED_LEGS` names only `test`, `test:server`, `test:windows`, `test:server:windows` — this leg's keys are never compared against anything;
- `legs.size >= 11` only *grows* when a leg gains an `if:`.

Result: #3122's gap silently reopens with a fully green suite. That is the same failure mode this test file's own `#3053` comments document ("A leg that must exist therefore belongs in `REQUIRED_LEGS` below, not behind a count") — the PR adds a leg that belongs in a named guard and puts it in none.

CLAUDE.md "Testing discipline (REQUIRED for every change)" is unmet: the Test plan is four manual checkboxes, two of which ("no `if:` scope-gate", "NOT folded into the path-filtered `onbox-register-check.yml`") are restatements of assertions that belong in the static-scan suite this repo already maintains over this exact file.

Fix shape: an assertion in `workflow-wiring.test.mjs` that the `Register citation check` step exists **and** its step body contains no `if:` key — plus, per that file's own anti-vacuity habit, a synthetic fixture proving the assertion actually goes red when an `if:` is added. `test:hooks`' scope includes `.github/workflows/**`, so such a guard really would fire on the offending edit.

---

### 🟡 2. Four prose sites now assert the opposite of what this PR just made true

This is CLAUDE.md's named chore "a stale comment or doc sentence your change made false", and it is the "one instance fixed, the class left armed" shape — the first site is **in the file this PR edits**.

1. **`.github/workflows/verify.yml:64-69`** (head file), the workflow's own header:
   > "Every leg's own `if:` scope condition already evaluates false for a docs-only diff, so every job still completes in roughly the time of "Setup Node + deps" alone"

   False as of line 218 of the same file, ~150 lines below it.

2. **`scripts/check-register-citations.mjs:14-37`**, the "WIRING GAP, stated explicitly" block — three separate false claims now:
   - `:15-17` "`check:register-citations` … is invoked from exactly one place today — `scripts/tests/check-register-citations.test.mjs`'s own CLI-integration tests";
   - `:24-29` "There is no dedicated `.github/workflows/*.yml` step for this checker … `#2629`'s option 3 ("catches rot at PR time") is not fully true yet";
   - `:33-37` "The right fix is a dedicated CI step (mirroring `onbox-register-check.yml`) that always runs this checker regardless of diff scope … not something to wire in blind here; **tracked at `#2721`**." That pointer is doubly stale — #2721 is **CLOSED** and is titled "ops-64 — rehome the wrongId-widening deferral that #2629's closure orphans", i.e. a different subject. *This PR is the thing that block says is owed.*

3. **`CONTRIBUTING.md:579-581`** — "Every leg's own scope condition evaluates false for a docs-only diff, though, so the job completes in roughly the time of "Setup Node + deps" alone (~20-40s)". Both halves are now wrong: one leg has no scope condition, and the docs-only job now additionally pays a full-tree scan of 3630 files.

4. **`CLAUDE.md:1015-1017`** — "every leg's own scope condition is false for a docs-only diff, so the job just does env setup and reports green".

While fixing (3)/(4), the new always-on behaviour deserves one recorded consequence that no site currently states: because the checker scans the **whole tree**, not the diff, a citation broken by lane A merging to `main` will redden lane B's unrelated PR with an error naming files lane B never touched, and block lane B's merge until lane A is fixed. That is an intended property of "must run on every PR", but it is a shared-fate behaviour change for a required status check in a repo that runs many parallel lanes, and it should be written down rather than discovered.

Nit while in the file: the new comment's `(unconditional — see #3122: …` opens a parenthesis at `verify.yml:218` that is never closed.

---

### 🟡 3. `Closes #3122` overclaims — the local half of the gap is still unwired

`.github/workflows/verify.yml:218-223` · issue #3122

#3122's "What" names three surfaces with no wiring: "no CI workflow, **no `test:all`/`verify` step**, and **no git hook** ever runs it." This PR wires one. `npm run verify` — which CLAUDE.md's Before-shipping step 7 calls for before a release cut, and which every other step in `lint-and-checks` (`lint`, `typecheck`, `config:check`, `test:hooks`, `check:budget-poll`, `audit`, …) participates in via `verify-cache.mjs`'s `STEPS[]` — still does not run this checker. Concrete consequence: a developer runs the full local battery green, pushes, and CI reddens on a leg the local battery structurally cannot reach.

This one is not a naive add, and the PR should say which way it went rather than leaving it implied:

- `ci-scope.mjs:16,39` derives `step_<slug>` for **every** `STEPS[]` entry;
- `workflow-wiring.test.mjs:72-80` (`<- every emitted scope key is referenced by the workflow`) then requires `verify.yml` to reference `step_check_register_citations` — which an intentionally-unconditional step has no `if:` to do from.

So there is a genuine decision here (add the STEPS entry and carve the key out of the `<-` assertion / reference it in an always-true form / hang it off `test:all` instead / keep it CI-only on purpose). If it is being deferred, that is the design-pass carve-out: change `Closes #3122` to `Refs #3122`, keep the issue open, and **name that decision in it** — "needs design" without the decision named is a deferral in disguise. If it is deliberately out of scope forever, say so in #3122 and close it with that reasoning recorded.

---

### 🟡 4. PR body omits the two required not-applicable declarations

PR body, `## Test plan`

CLAUDE.md's Before-shipping checklist: "Skipping a step is fine when the step genuinely does not apply … **but say so explicitly rather than silently omitting.**" Two gates are silently absent rather than declared:

- **step 5, the release-notes pair** (`docs/release-notes-next.md` + `RELEASE_NOTES.md`). A CI-only change plausibly qualifies for the "no shippable delta" skip — but it is an operator-visible change to what every PR's required check does, so the call should be made in writing, not inferred. (Note `scripts/release-notes-gate.mjs` only runs at tag time in `release.yml`, so nothing mechanical catches this.)
- **step 3, on-box acceptance** — plainly N/A here, but the checklist asks for the explicit "N/A", and a reader currently cannot tell considered-and-N/A from forgotten.

---

### Triage summary

| # | Severity | Class | Site |
|---|---|---|---|
| 1 | 🟠 Significant | correctness bug (unguarded regression) | `.github/workflows/verify.yml:218-223`, `scripts/tests/workflow-wiring.test.mjs` |
| 2 | 🟡 Minor | chore (stale docs, 4 sites) | `verify.yml:64-69`, `check-register-citations.mjs:14-37`, `CONTRIBUTING.md:579-581`, `CLAUDE.md:1015-1017` |
| 3 | 🟡 Minor | chore (issue linkage / unwired half) | PR body, `verify-cache.mjs` `STEPS[]` |
| 4 | 🟡 Minor | chore (house-gate declaration) | PR body |

Finding 1 is a correctness finding and re-triggers a re-review pass once fixed and pushed. Findings 2–4 are fixed this round but do not on their own reopen the loop.

dudarenok-maker:
## PR review — pass 2 (head 24c79eec, depth low)

Re-review over the full three-commit set (`bda4a2aa` → `78e32eea` → `24c79eec`), not just the fix commits. Depth re-derived from the current commit set: all three are single-scope `ops` (`ci` / `test` / `docs`) → **low**.

---

**F1 (🟠, pass 1) — no test pinned the step's unconditional-ness.** → **Resolved for the shape it named, but the fix has its own fail-open path.**

The new test at `scripts/tests/workflow-wiring.test.mjs:1494` does go red on the regression F1 described. I re-derived this independently rather than taking the commit message's word for it, by replaying the assertion body verbatim against mutated copies of `verify.yml`:

| mutation | result |
|---|---|
| control (tree as committed) | GREEN ✅ |
| `if:` inserted **before** `run:` (the house style every other gated step in this file uses) | **RED** ✅ |
| `if:` inserted **after** `run:` | **RED** ✅ |
| `if:` as a folded block scalar (`if: >-`, the spelling #3053 pass 3 found a sibling regex could not see) | **RED** ✅ |
| step deleted outright | **RED** ✅ |

So the dominant regression is genuinely caught. The residual hole is new finding **N1** below.

**Finding 2 (🟡, pass 1) — four stale prose sites.** → **Resolved for the three I could verify mechanically; the fourth introduced three new inaccuracies (N2–N4).**

Verified true rather than assumed:

- `verify.yml:70`'s new *"ONE exception"* claim is accurate. I scanned every `- name:` in the file and resolved each step body: the only work-doing steps carrying no `if:` are `Checkout`, `Setup Node + deps`, `Detect changed scopes`, the two aggregator steps, and `Register citation check` (`:226`). It is the only *leg* that runs unconditionally.
- The step really does run on docs-only PRs: `lint-and-checks` carries no job-level `if:` (only `name` / `needs: detect` / `runs-on`), so nothing upstream skips it.
- The *"negligible (a few seconds)"* cost claim — which #3122 explicitly asked to be **measured**, not asserted — checks out against this PR's own run: job `102799178530`, step 10 `Register citation check`, `08:26:19Z → 08:26:23Z` = **4s**. CLAUDE.md's "completes in seconds" and CONTRIBUTING.md's "~20-40s" both survive it.

**Finding 3 (🟡, pass 1) — `Closes #3122` overclaimed.** → **VERIFIED RESOLVED at the PR-body level.** Body now reads `Refs #3122` in plain text (outside any code span; `pr-issue-link` green). #3140 is OPEN, labelled `area:ops` + `type:chore`, and names the decision owed with three enumerated options rather than "needs design" — which is what the design-pass carve-out requires. `type:chore` ⇒ no `docs/BACKLOG.md` row owed, so nothing missing there. *But* the same overclaim is re-asserted in code — see **N2**.

**Finding 4 (🟡, pass 1) — missing not-applicable declarations.** → **VERIFIED RESOLVED.** The `## Not applicable` block covers before-shipping step 5 (release notes) and step 3 (on-box acceptance), each with a reason. Step 1 (regression plan) and step 4 (INDEX.md) are covered by their own carve-outs and are not silent omissions.

---

### New findings

#### 🟠 N1 — correctness (in the guard) — the new test silently disarms itself if the `run:` line ever changes

`scripts/tests/workflow-wiring.test.mjs:1519-1522`

```js
const stepSection = source.slice(
  source.indexOf('- name: Register citation check'),
  source.indexOf('run: npm run check:register-citations') + 50,
);
```

When the literal `run: npm run check:register-citations` is not present verbatim, `indexOf` returns `-1`, the slice end becomes `49`, the start is ~8100, and `String.prototype.slice(8100, 49)` returns `''`. `assert.doesNotMatch('', /^\s*if:/m)` then passes. The step's whole invariant is checked against an empty string, and **nothing goes red to announce it**.

The first assertion does not rescue this: its regex (`:1507-1508`) is `/- name: Register citation check\n(?:\s*[^\n]*\n)?(?:\s*if:|\s*run:)/` — it accepts `\s*if:` as an alternative to `\s*run:`, so a step that *has* acquired an `if:` still satisfies `assert.ok`. (The comment at `:1504-1506` says it looks for "`run:` (no `if:` line in between)"; the regex it describes does the opposite.)

Concrete two-step scenario, both steps empirically confirmed:

1. A pure rename chore renames the npm script (`check:register-citations` → `check:citations`) and updates the step. The step is still unconditional and still correct. Guard: **GREEN**, `stepSection.length === 0`. The guard is now permanently inert and no test says so.
2. Any later PR adds `if: fromJSON(needs.detect.outputs.scopes).…` to the step — the exact regression #3122 names. Guard: **GREEN**. #3122 is silently reopened, which is the outcome this test exists to prevent.

Two more confirmed fail-opens off the same anchor:

- `if:` added **together with** any rewording of the `run:` line (e.g. `run: node scripts/check-register-citations.mjs`) → GREEN.
- The step kept unconditional but neutered — `run: "true"` → GREEN. Nothing in this file pins that the step still *invokes the checker*, so the guard does not actually establish "the checker runs on every PR".

Separately, the `+ 50` window is a magic number with ~1 character of slack: it clears `\n` + 8 spaces + `if:` by one byte. Reindent the step body to 10 spaces and the after-`run:` case (currently RED) goes GREEN.

The comment at `:1502` — *"Mutation proof: adding any `if:` condition to the step makes this test fail"* — is therefore false as written; the four cases above are counterexamples. This is the repo's own recurring shape (reviewer-brief #1 "a guard that fails open on absent evidence", #3 "a test that cannot fail", #6 "success reported while doing nothing").

Shape of a fix: assert on the parsed step rather than on a byte window anchored to a mutable literal — e.g. slice from the `- name:` line to the next line at or below the step's own indent that starts a sibling `- `, assert that block contains no `/^\s*if:/m`, **and** assert it still contains `check:register-citations`. That last assertion is what makes the guard fail loudly instead of vacuously when the command is renamed.

This matters more than it otherwise would because the bespoke test is the *only* thing binding this step — see N5.

#### 🟡 N2 — cleanup — the code comment re-asserts the `Closes` overclaim that pass 1 removed from the PR body

`scripts/check-register-citations.mjs:19`

```js
// step (wired in PR #3134, closing #3122).
```

#3122 is OPEN and stays open. #3140's own body states the disposition explicitly: *"PR #3134's `Closes #3122` should read `Refs #3122` — the CI-leg gap is closed, the local-wiring gap named in #3122 is not."* The commit that fixed the PR body wrote the opposite claim into the source tree in the same round. (The following paragraphs do split "CI WIRING (closed)" from "LOCAL WIRING (open)", so the block contradicts itself as well as the ticket.)

#### 🟡 N3 — cleanup — the comment's `verify.yml` line citation points at the wrong step

`scripts/check-register-citations.mjs:22` — *"see `.github/workflows/verify.yml` lines ~218-223"*.

At this head, `verify.yml:218-220` is the **Budgeted-poll guardrail** step; `:222-224` is the new step's comment block and the step itself is at `:225-227`. The cited range excludes the thing it points at and includes an unrelated step.

#### 🟡 N4 — cleanup — the comment's load-bearing "not invoked locally" claim is false for `test:all`

`scripts/check-register-citations.mjs:27-29` — *"The checker is still NOT invoked from `npm run verify` or `npm run test:all` locally, or from any git hook."*

It is invoked from both. `test:all` and `verify` both run `test:hooks`, which runs `scripts/tests/check-register-citations.test.mjs`, which spawns the CLI **against the real tree** — see that file's `:2665` (*"The real tree at HEAD carries no existing-ID-cited-for-the-wrong-subject…"*) and `:2801` (*"CLI: the real tree scan finds at least one binary file"*).

The accurate statement is the one this commit deleted: the checker is reachable *only* via `test:hooks`, whose `verify-cache.mjs` input globs do not span the tree, so on a diff outside those globs the step prints `[cached]` and the real-tree scan never runs. That distinction is the actual premise of #3140's open question, and replacing it with a flat "not invoked" makes the remaining gap harder to reason about, not easier. (Worth noting `docs/testing/**` and `docs/features/**` *are* already `test:hooks` inputs — `verify-cache.mjs:190` and `:197` — added for the sibling `check-register-row-citations.mjs`, so the gap is narrower than "the whole tree".)

#### 🟡 N5 — cleanup — the new `# leg:` marker is inert

`.github/workflows/verify.yml:225` carries `# leg: check:register-citations`, but `parseLegs` (`scripts/tests/workflow-wiring.test.mjs:255`) matches `/# leg: (…)\n\s*- name: [^\n]+\n\s*if: ([^\n]+)\n/` — it **requires** an `if:` line, which this step deliberately does not have. So the marker binds to no guard, is not counted by the `legs.size >= 11` floor at `:452`, and cannot be listed in `REQUIRED_LEGS` (it has no scope keys by design).

That is precisely the failure mode this file's own doctrine at `:246-252` was written about after #3053 — *"A marker outside it parses as NOTHING, silently, which is how both Windows leg steps came to sit outside every guard in this file."* Nothing breaks today, but a reader reasonably assumes the marker means the step is registered, and it is the reason N1's fail-open is load-bearing rather than redundant: this step is bound by exactly one assertion and nothing else. Either drop the marker or note in-line why it is documentation-only.

---

### Checked and clean

- No behavioural change to `check-register-citations.mjs` — the diff there is `//` comment lines only, so no paired test is owed for it.
- The old comment's `#2721` citation was removed; #2721 is CLOSED and is about an unrelated wrongId-widening deferral, so it was mis-cited to begin with. Removing it is a net fix and leaves no bookkeeping owed.
- Adding an unconditional step to `lint-and-checks` needs no aggregator change — `Check leg results` buckets *job* conclusions, not steps, so a red step still surfaces.
- Correct sibling wired: `check:register-citations` (#3122's subject), not the narrower `check:register-row-citations` (#2831), which is separately covered as a `test:hooks` `extraFiles` input at `verify-cache.mjs:233`.
- `CLAUDE.md` and `CONTRIBUTING.md` edits are accurate; the ragged `stays "PR required` wrap in CONTRIBUTING.md renders identically and the pre-existing `+ title valid…` line-leading-`+` is untouched.

---

### Re-review trigger

**Yes.** N1 is a 🔴/🟠-class correctness finding (a guard that fails open), so fixing and pushing it re-triggers a pass. N2–N5 are 🟡 cleanup — fixed this round per the report-fix-record rule, but they do not themselves reopen the loop.

dudarenok-maker:
## PR review — pass 3 (head b728d847, depth low)

Final allowed round (initial + 2 re-reviews = 3). Depth re-derived from the current commit set: all six commits are single-scope `ops` (`ci`/`test`/`docs`) → `low`.

Every claim below was re-derived against the actual PR head, not from the prior passes' self-reports. The new guard was tested empirically against 19 mutated copies of `verify.yml`.

---

**🟠 N1 (pass 2) — VERIFIED RESOLVED.** The byte-offset/literal-`run:`-substring anchoring is gone; the guard now locates the step by its `- name:` line, derives `stepIndent` from that line, and extracts the block by indentation. I ran the rewritten logic verbatim against mutated copies of the real `verify.yml`. It goes **RED** on all of these:

| mutation | guard result |
|---|---|
| `if:` added after `run:` | RED (no-`if:` assertion) |
| `if:` added before `run:` | RED |
| `if: >-` folded block scalar | RED |
| `if: ${{ false }}` inline | RED |
| npm script renamed in `run:` | RED (`run:` assertion) |
| `run: "true"` (neutered) | RED |
| whole step deleted | RED (step-not-found — **loud**, not vacuous) |
| step display-name renamed | RED (step-not-found) |
| step commented out | RED |
| flow-mapping step `- { name: …, if: false, … }` | RED |
| `- if: false` with `name:` demoted below it | RED |
| whole job re-indented +2 | GREEN (correct — no false positive) |

The specific pass-2 failure mode — a rename silently producing an empty search window and a vacuous pass — is closed: the block always contains at least the located `- name:` line, and if the step is not located the test throws at `workflow-wiring.test.mjs:1526` before either assertion runs. The extraction also cannot over-run into a sibling step, since any sibling's `- name:` sits at exactly `stepIndent` and breaks the loop. No new vacuous-pass mode introduced.

**🟡 N2–N4 (pass 2) — VERIFIED RESOLVED.** The `#3122`-fully-closed overclaim is gone (header now scopes it to the CI-wiring gap and routes the rest to #3140; the PR body correctly says `Refs #3122`, not `Closes`). The wrong line-number citation is gone (no line reference remains). #3140 exists, is OPEN, is labelled `area:ops`/`type:chore`, and its body matches what the header says it tracks. See 🟡 C below — the "never runs locally" correction now overshoots slightly in the other direction.

**🟡 N5 (pass 2) — VERIFIED RESOLVED.** The inert `# leg:` marker is replaced with a plain comment, and no machinery broke. Re-measured against `verify.yml` at this head: `parseLegs` → 19 legs (floor 11), `scanDerivedSteps` → 29 derived conditions (floor 25) with **0** unmarked, `fromJSON` references → 64 (floor 35). The replacement comment correctly does not match `MARKER_LINE`, and does not need to: the step carries no `if:`, so `scanDerivedSteps` never reaches it.

**Live wiring confirmed, not just statically asserted.** The step actually executed on this PR's own CI run — job `Lint, typecheck & fast checks` (run 34457566267) shows `Register citation check: success` sitting between five `skipped` scope-gated siblings. That is the property #3122 needs, observed rather than inferred.

---

### New findings

#### 🟡 A — cleanup — `continue-on-error: true` disarms the step and the guard stays green

`scripts/tests/workflow-wiring.test.mjs:1555-1574` / `.github/workflows/verify.yml:226-227`

The guard asserts two things: no `if:`, and the exact `run:`. `continue-on-error: true` satisfies both while achieving the identical outcome the guard exists to prevent.

Concrete scenario, empirically verified:

```yaml
      - name: Register citation check
        continue-on-error: true
        run: npm run check:register-citations
```

Guard result: **GREEN**. And no other assertion in the file catches it either — I re-ran the whole set against this mutation: `derived=29`, `unmarked=0`, `legs=19`, all floors satisfied. In CI the checker would run, exit 1 on a broken register citation, be recorded as failed-but-continued, and `lint-and-checks` would still conclude `success` → the `verify` aggregator green → required check green. #3122's gap fully reopened, with the step still present, still unconditional, and still running the right command.

This is not exotic. The repo already uses this exact idiom deliberately: `.github/workflows/quarantine-health.yml:67` and `.github/workflows/release.yml:271`. It is the standard reach when a check reports a false positive and someone wants to unblock a merge — which is precisely the moment this guard is supposed to fire.

Same class, one spelling further out: `if : false` (space before the colon) slips `/^\s*if:/m` and also stays GREEN.

Fix is one assertion plus one character:

```js
  assert.ok(!/^\s*if\s*:/m.test(stepBlock), /* existing message */);
  assert.ok(
    !/^\s*continue-on-error\s*:/m.test(stepBlock),
    'Register citation check step must not carry `continue-on-error` — the step would run, ' +
      'fail, and still report the job green, reopening #3122 with the step apparently intact.',
  );
```

**Severity rationale (this matters for the loop cap):** classified 🟡, not 🟠. Nothing in the PR as it stands behaves wrongly — `verify.yml` carries no `continue-on-error`, the step gates, and CI proves it. The guard does not self-disarm and does not misreport anything about the current tree; it fails loud on every disarm its own stated invariant names. That is categorically different from pass-2's N1, which could go **vacuous** — asserting nothing while reporting green — through an ordinary rename. This is a guard-coverage gap, which this repo's own taxonomy files as a chore ("a seam you touched that has no test"), so it is fixed in this round but does **not** reopen the review loop.

#### 🟡 B — cleanup — the comment describes a helper the code does not use

`scripts/tests/workflow-wiring.test.mjs:1556-1557`

> `// Use the same ifConditions pattern as the rest of this file: extract all`
> `// if: lines (including block scalars), and assert none are found.`

It does not use `ifConditions` (defined at `:335`). It uses a bare `/^\s*if:/m`. The code is *right* here — detecting the `if:` **key** doesn't need the value-normalisation `ifConditions` does, which is why the block-scalar mutation still goes red — but the comment claims a provenance the code doesn't have, which is exactly the trap that makes the next reader believe this guard inherits `ifConditions`' hardening. Reword to state what it actually does and why the bare key match suffices.

Relatedly, the docstring at `:1508-1510` lists "Scenarios caught" with no counterpart list of what is *not* caught — the omission that let A through. Worth adding the boundary explicitly once A is fixed.

#### 🟡 C — cleanup — the corrected LOCAL WIRING prose overshoots in the other direction

`scripts/check-register-citations.mjs` (header, "LOCAL WIRING (open, tracked at #3140)")

The corrected text says the checker "IS reachable locally … (which `npm run verify` and `npm run test:all` both run)" but then that "the step can be marked `[cached]`/skipped rather than actually re-scanning", and that "A dedicated local step that runs unconditionally (independent of scope) is tracked separately at #3140."

True of `npm run verify` (cache-aware; `verify-cache.mjs`'s `test:hooks` input globs don't span the whole tree). **Not true of `npm run test:all` / `npm run verify:quick`, which the same sentence names.** Verified:

- `package.json:85` — `"test:all": "npm run test:hooks && …"` — a direct invocation, no `verify-cache.mjs`, no scope gate, no cache. `package.json:90` — `"verify:quick": "npm run test:all"`.
- `scripts/run-hooks-tests.mjs:10` — `fg('scripts/tests/*.test.mjs')`, so `check-register-citations.test.mjs` is always selected.
- That suite runs the CLI against the **real tree** and asserts `result.status === 0` (e.g. `check-register-citations.test.mjs:2665-2674`, `:2801-2810`), so real-tree citation rot turns it red.

So an unconditional, whole-tree local run already exists today via `test:all`/`verify:quick`; the residual local gap is narrower than the header states — it is the *cached* `npm run verify` path plus the git hook, not local coverage in general. Two concrete consequences: a reader takes the header at face value, concludes local runs can't catch citation rot, and skips a `verify:quick` that would have caught it; and #3140's option *"Leave local `verify`/`test:all` without this check permanently"* is asking about something `test:all` already has. Narrow the header sentence, and post a correcting note on #3140 so its decision is framed against the real remaining gap.

#### 🟡 D — cleanup — the reflow left an orphan line above a paragraph-breaking `+`

`CONTRIBUTING.md` (~line 583, the docs-only fast-path paragraph)

The reflow ends a line with `stays "PR required` and leaves the following context line starting `+ title valid + no conflicts …`. Under CommonMark a `+ ` at line start interrupts the paragraph and becomes a bullet. Verified via GitHub's own renderer:

```html
<p>and reports green — the gate
stays "PR required</p>
<ul>
<li>title valid + no conflicts + a fast green required check", not the 10-15
min full battery.</li>
</ul>
```

The sentence renders split across a paragraph and a bogus list item, and the `+` is eaten. The `+ title valid` line itself is pre-existing (a context line in this diff), but the reflow rewrote the line directly above it and left the orphan, so the one-word fix — pull `+ title valid` up onto the previous line — belongs here rather than in a follow-up.

---

### Verdict

**No 🔴 / 🟠 correctness bugs.** N1 is genuinely closed — I could not construct a mutation that disarms the step *and* leaves the guard green through any path the guard is scoped to cover, and the one path outside that scope (A) is a coverage gap in a guard that is otherwise correct, not a live defect. N2–N5 are all verified resolved. CI is green on every leg, and the new step is observed running on this PR's own run.

**The loop cap is not tripped.** Per `findings-triage.md`, a pass surfacing only 🟡 cleanup findings does not re-trigger a re-review. Fix A–D in this round per the report-fix-record rule, then merge — no pass 4 is owed or permitted.

<sub>House gates: paired test present and load-bearing (mutation-verified above); release notes explicitly N/A (CI-only, no shippable delta) — correct; on-box acceptance explicitly N/A — correct, no hardware-provable behaviour; `Refs #3122` present, outside any code span, and the narrower link is the right call given #3140; no `cast.json` writes, no config knob, no derived artifact staled by this diff.</sub>

Issue #3100, opened 2026-09-08
[agent instructions][claude][verify] Verify #3086 VD/Kokoro arbiter fix


dudarenok-maker:
[agent instructions][claude][verify] Verify #3086 VD/Kokoro arbiter fix

## What

Verify the fix for #3086 (the parent) landed on branch
`fix/sidecar-3086-vd-kokoro-arbiter-race` in worktree
`C:\Claude\Projects\wt-3086-vd-kokoro-arbiter-race`, then open the PR.

## Parent acceptance, restated as a checklist

- [ ] The task child identified a concrete mechanism (in code, with file:line)
      for why a raw Kokoro `/synthesize` call can complete while
      `_VD_KOKORO._design_active_count > 0` — either a call site that bypasses
      `_VD_KOKORO.kokoro_synth()` entirely, or a real bug in
      `_VdKokoroArbiter`/its manual `design()` entry/exit management at
      `server/tts-sidecar/main.py:7158-7220`.
- [ ] A pure-Python unit test exists (no GPU, no real models needed — the
      `_VdKokoroArbiter` class in isolation is just `threading.Condition`
      bookkeeping) that reproduces the race: two threads simulate
      `design()` held while a third simulates `kokoro_synth()`, and the test
      asserts the Kokoro thread blocks until both designs release. The test
      FAILS against the pre-fix code and PASSES after the fix.
- [ ] If the root cause was a bypassed call site rather than the arbiter
      class itself, the test instead pins that the actual `/synthesize` route
      handler for Kokoro provably goes through `_VD_KOKORO.kokoro_synth()` (a
      call-path assertion, e.g. via a spy/mock on the context manager).
- [ ] Re-run the task child's claimed mutation yourself and paste the observed
      failing output — do not trust an unverified receipt.
- [ ] `server/tts-sidecar/run-tests.ps1` (or the scoped pytest invocation
      covering the changed test file) passes.
- [ ] The fix does not touch `server/tts-sidecar/tests/test_*` files unrelated
      to this race, and does not touch any GPU-requiring code path in a way
      that would need real hardware to validate (see "On-box acceptance"
      below — this fix should be provable with a pure concurrency unit test).

## Re-run any mutation whose receipt asserts it without showing output

If the task child's `AGENT DONE` comment claims a mutation test without
pasting the actual failing assertion output, re-run it yourself before
trusting PASS.

## Read-only for source

You report, you do not fix.

## FAIL ordering (exact)

[...]

Issue #3058, opened 2026-09-06
[agent instructions][claude][task] Add per-request GPU device hint for the lazy Coqui derive so E-06 isn't blocked by vram-spill (#2915)


dudarenok-maker:
## What

Split off `#2919`'s E-06 blocker per the operator's decision there. E-06 ("a designed voice
on Coqui, judged against the stock catalogue voice it replaces") needs a **lazy Coqui derive**
that holds Qwen (~6.1 GB) and Coqui (~5.4 GB) resident on the GPU **at the same time**. On this
box GPU0 (8188 MB total, ~3.3-3.5 GB already used by other processes) cannot fit both even at
100% idle — two live attempts both failed `vram-spill`. GPU1 (RTX 5070 Ti, 16 GB, ~15.4 GB free
throughout both attempts) sits idle because nothing routes this one operation to it.

The operator's instruction on `#2919` was explicit: find or add a lever to pin **this one
derive** to GPU1 without moving the whole app's default device there (other workloads on this
box may expect GPU0), and if no such lever exists today, file it as its own task — which is
what this issue is.

## What's already been investigated — do not re-derive this

A same-day investigation (recorded on `#2919`) read the actual code and found:

- **Sidecar launch**: `server/src/tts/spawn-sidecar.ts` `spawnSidecar` (~line 605) starts one
  long-lived uvicorn process (`server/tts-sidecar/main.py`). `buildSidecarEnv` (~line 501)
  injects `*_DEVICE` env vars from the config registry, but changing any of them requires a
  full sidecar **restart** — there is no live per-request override today.
- **A device-pin mechanism already exists, but it's per-engine, not per-operation**:
  `server/src/config/registry.ts` has `tts.coqui.device` → `COQUI_DEVICE` (~line 666) and
  `tts.qwen.device` → `QWEN_DEVICE` (~line 686), accepting `auto|cpu|cuda|cuda:N|cuda-uuid:UUID`.
  In `main.py`, `_engine_env_pin()` (~line 4055) reads these at admission time. When an engine's
  device is left `auto`, `PlacementController.admit()`/`reservation()` (~line 4774, ~5082) via
  `_gpu_candidates()`/`best_fit` (~line 4624) **already does cross-GPU capacity-aware
  placement** — an unpinned, not-yet-resident engine can already land on GPU1 on its own. The
  gap is only when you need ONE specific HTTP call routed differently from the rest of that
  engine's traffic.
[...]

PR #3133, opened 2026-09-09
ci(ops): enforce the live view's publish token via --stamped-since

dudarenok-maker:
## Summary

The on-box register's live view carries a publish token — `data-published-as`
/ `data-publish-id`, written only by `npm run stamp:publish-token`. It orders
publishes of the register's artifact and lets an operator tell *"the live page
is ahead of me"* from *"the live page is behind"* in one line. Stamping it is
mandatory per the register's own step 0, and **nothing enforced it.**

This adds an opt-in `--stamped-since <ref>` to
`scripts/check-onbox-register.mjs`, invoked by `onbox-register-check.yml` with HEAD^1 (the merge commit's first parent). If the live view's *rendered* content changed since the base
and the counter did not move, the check **reports** the issue and names
`npm run stamp:publish-token`. **The check is not currently a required merge
gate** (see #3138 for the enforcement decision).

### Why now — it has already cost real time, twice

**#3076** folded rows A37/B2/D1 out and merged unstamped. Tracked and published
both read `13` while their contents had diverged, so the register's own one-line
disambiguator was uninformative and the diagnosis needed a manual byte-diff —
exactly what #2599/A41 built the token to replace. Fixed by #3114 / PR #3115.

It then happened **again while this PR was being written**: #3061 merged with an
A106 row added and no stamp, so `main`'s live view now carries A106 at token 16
— the artifact published at 16 does not show that row.

### The predicate, and why the obvious one is wrong

Content comparison strips HTML comments (the file header and the
`<!-- BEGIN GENERATED:... -->` markers are not rendered; the counts *between*
them are) and blanks the token's own attribute values. Identical → pass. Different
→ require `data-published-as` to be **strictly greater than the base ref's value**.

The predicate is: fail if `counter < base`, fail if `counter === base`, fail if
`counter > base AND nonce unchanged` (hand-edited counter; `stamp-publish-token`
always mints a fresh nonce). Anything else passes.

That matters because a check asking merely *"does this PR touch the live view
and also change the token?"* would have passed #3076, #3061 and #3063, because
a stamp commit anywhere in a branch's history proves nothing about its tip. Only
comparison against the base catches an unstamped edit landed after an earlier
stamp — and needing the base ref is what forced the flag design over folding this
into the no-flag run, which deliberately compares only the two tracked files and
must keep working offline.

Fails closed on a missing or malformed token on either side, and on any `git
show` failure other than "path absent at that ref" — an unresolvable ref must
never read as "no change".

## Test plan

[...]

PR #3142, opened 2026-09-10
fix(sidecar): close /load's bypass of the VD/Kokoro arbiter

### LINKED ISSUE #3086: VdKokoroArbiter does not block a raw Kokoro /synthesize call while a VoiceDesign forward is active
## What

`_VdKokoroArbiter` (`server/tts-sidecar/main.py:1530-1619`) is supposed to
guarantee a Kokoro `/synthesize` forward never overlaps a Qwen VoiceDesign
forward on a shared card (`shares_device` true) — `kokoro_synth()` waits
while `_design_active_count > 0` before proceeding
(`main.py:1577-1591`), and `KokoroEngine.synthesize()` wraps its entire
forward (including lazy load) in that guard (`main.py:3800-3804`).

On-box confirmation for `docs/testing/onbox-acceptance-register.md` row
**A105 bullet 4** (batch 2 step 2, dudarenok-maker/Castwright#2986) found
that a raw `POST /synthesize` Kokoro call completes successfully **while
two VoiceDesign forwards are still concurrently resident**, instead of
blocking until both release.

## Repro (this box, worktree `wt-mechanical-batch-2`, sidecar port 9170)

1. `POST /qwen/design-voice {"voiceId":"qwen-a105b4-E","instruct":"a calm, steady middle-aged narrator, neutral pace"}` — fired 01:31:06 UTC. `/health` 3s later: `qwen_design_resident: true`, `inflight_synth: 1`.
2. `POST /qwen/design-voice {"voiceId":"qwen-a105b4-F","instruct":"a bright, energetic young woman, fast pace"}` — fired 01:31:31 (25s after E). `/health` immediately after: `qwen_design_resident: true`, `inflight_synth: 2` (both confirmed concurrently in flight).
3. `POST /synthesize {"engine":"kokoro","model":"kokoro-v1","voice":"af_bella","text":"This is a raw arbiter probe sentence for bullet four."}` — fired 01:31:36 (5s after step 2), while both designs were confirmed resident.

## Observed

- The Kokoro call returned `HTTP 200` with a real 142,336-byte PCM payload
  after 44.11s (completed ~01:32:20 UTC).
- Design E returned `HTTP 200` after 94.59s (completed ~01:32:41 UTC) —
  **21s after** the Kokoro call had already finished.
- Design F returned `HTTP 200` after 98.75s (completed ~01:33:10 UTC) —
  **50s after** the Kokoro call had already finished.

`_VD_KOKORO.design()` is entered manually in `design_voice()`
(`main.py:7126-7127`) and released only in that function's own `finally`,
spanning the full model load *and* GPU forward per the inline comment at
`main.py:7149-7165` ("remain inside `_VD_KOKORO.design()`" through the
forwards). The design route only sends its HTTP response after
`design_voice()` returns (`main.py:11352-11375`), so a design's HTTP
completion time is an authoritative lower bound on how long it held
`_design_active_count`. The Kokoro call's completion strictly precedes both
designs' completion, so it ran and finished while `_design_active_count`
was still ≥ 1.

Ruled out `_shares_device` being `False`: `server/.env` in this worktree has
both `QWEN_DEVICE`/`KOKORO_DEVICE` unset, `/health` confirms both engines
resolve to unindexed `cuda` (same card), and `shares_device()`
(`main.py:5646-5661`) computes `True` for that case — there's no config on
this box that would put the arbiter in its no-op mode.

## Expected

The Kokoro call should have blocked (via `self._cv.wait()` in
`kokoro_synth()`) until both designs released `_design_active_count`, per
the class's own docstring ("this arbiter guarantees the two heaviest-combined
ops never co-reside").

## Not investigated further

Root cause not diagnosed — could be a missed `notify_all()`/wait race, the
condition variable not being shared correctly across the `asyncio.to_thread`
worker threads involved, or something else. Filed to hand off rather than
fixed, per this task's scope (on-box confirmation only).

## Evidence

Full run writeup: `docs/testing/onbox-mechanical-batch2-results/step-2-qwen-lifecycle.md`,
"11th run (2026-09-08)" section, on branch `docs/docs-mechanical-batch-2`
(worktree `wt-mechanical-batch-2`, not yet merged to `main`).

Refs #2986, #2960, #2435

dudarenok-maker:
## Summary

- `POST /load {"engine":"kokoro"}` and the `PRELOAD_KOKORO` startup path called `KokoroEngine._ensure_loaded()` directly, bypassing `_VD_KOKORO.kokoro_synth()` entirely — unlike `KokoroEngine.synthesize()`, which already wraps its whole forward in that gate.
- On the DirectML profile a cold Kokoro load is not just bookkeeping: `_directml_selftest_or_fallback` runs a real one-shot forward (`kokoro.create("ok", ...)`) to prove the provider works, which could land mid-design with zero exclusion — this is the confirmed mechanism behind #3086.
- Added `_kokoro_ensure_loaded_guarded()`, which takes `_VD_KOKORO.kokoro_synth()` around `_ensure_loaded`, and routed both bypass call sites (`/load` and the startup preload) through it.
- Recorded the residual on-box debt as register row **A107** (`docs/testing/onbox-acceptance-register.md`): the unit-level fix is proven, but a real Kokoro `/load` racing a resident VoiceDesign forward on a shared-device box still needs a real on-box re-run.

Closes #3086

## Test plan

- New test `server/tts-sidecar/tests/test_load_kokoro_arbiter_gate.py::test_load_kokoro_blocks_while_design_active`: a `/load {"engine":"kokoro"}` call now blocks while `_VD_KOKORO.design()` is held elsewhere, and completes once released.
- **Mutation verified (re-run independently during #3100 verification):** reverted both guarded call sites back to the raw `kokoro._ensure_loaded(...)` call and reran the new test — it failed exactly as claimed:
  ```
  AssertionError: /load completed while a VoiceDesign forward was still active — the arbiter gate was bypassed (#3086/#3101)
  assert False
   +  where False = is_alive()
  ```
  Restored the fix and reran — green again.
- Scoped suite green after restoring the fix: `test_load_kokoro_arbiter_gate.py` + `test_design_kokoro_exclusion.py` + `test_load_admission.py` (27 passed, 2 skipped).
- `npm run check:onbox-register` passes after adding row A107 and rebuilding the live view via `npm run register:build`.

**On-box acceptance note:** this fix ships behaviour that a real GPU/real sidecar run originally surfaced (#3086, docs/testing/onbox-mechanical-batch2-results). The unit-level fix + regression test proves the arbiter logic in isolation; the original symptom (a live Kokoro `/load` racing a live VoiceDesign forward on real hardware) is recorded as owed acceptance under register row A107 (Group A), per CLAUDE.md's before-shipping checklist step 3.

**Rebased onto main (2026-09-10):** this branch was cut before PR #3061 minted **A106** for #3058's X-Device-Hint row, so the two collided on the same allocate-once id. Resolved during the rebase by minting the next free Group A id, **A107**, and bumping `next-id` to A108.

## Known limitations, accepted

- **B4:** A CPU-admitted Kokoro load (when capacity admission can't fit the GPU) still blocks behind the arbiter gate waiting for any active design to finish, even though a CPU load doesn't contend with a GPU-resident design for VRAM. Accepted because the blanket exclusion is simpler than device-aware gating, and this edge case (capacity hitting the GPU limit) is rare; the delay is acceptable overhead.
- **B5:** The arbiter's `cv.wait()` is unbounded and starvable if designs queue continuously (since `design()` does not wait on other designs, only drains Kokoro). This is NOT bounded by any external timeout on the dominant caller — `KokoroEngine.synthesize()` (ordinary TTS generation) has no wall-clock ceiling at all (`sidecar.ts` sets `headersTimeout: 0, bodyTimeout: 0` deliberately, plan 137). The `/load` endpoint's 90s Node-side client timeout only bounds the HTTP client giving up; it doesn't release the server-side wait, `kokoro._load_lock`, the `_loading` flag, or the VRAM reservation. Accepted anyway: a bounded wait would have to either proceed past a still-active design (reintroducing the exact GPU co-residency the arbiter exists to prevent) or raise and fail a chapter sentence mid-render — both worse than the current fairness cost. This is suboptimal fairness, not a correctness bug.

Sidecar acceptance: see docs/testing/onbox-acceptance-register.md row A107


dudarenok-maker:
## PR review — pass 1 (head e09cc760ea44413393306e3e42457907ca945d60, depth high)

**Scope / verified.** All 5 changed files against `git merge-base origin/main e09cc760` = `80354c75`. Read `main.py`'s `_VdKokoroArbiter` (`:1530-1617`), `KokoroEngine._ensure_loaded`/`synthesize`/`unload`, the `/load` and `/unload` routes, `design_voice`'s arbiter/`card_lock` interleave, the lifespan handler order, and `PlacementController.reservation`. Swept every `_ensure_loaded` test double in the sidecar suite for the staleness the `device=` kwarg introduces. Checked the register/live-view arithmetic by hand. Depth `high` — two findings below are labelled **uncertain** and flagged as such.

**Verified clean** (so a later pass doesn't re-derive it):

- **Register/live-view arithmetic is internally consistent.** Group A body rows = 33 (counted `^### A[0-9]`), glance table A = 33, live-view `glance:A` = 33 and `<span class="gcount">33 rows</span>`; group sum 33+1+3+1+7+2+2 = **49** = the strip's `owed` and the markdown's "**49 owed.**". `next-id` bumped to A108, strictly above the highest in-use A id (A107). The A106→A107 renumber is correct: A106 is taken on `main` by #3058 / PR #3061, and A107 appears nowhere else in the register.
- **`Closes #3086` is present in the PR body, outside any code span**; #3086 is open. (#3101 is already closed, so no `Refs` is owed.)
- **The `test_kokoro.py` stub fix is correct and minimal.** `lambda model, device=None: calls.append(model)` still appends only `model`, so `test_preload_skips_kokoro_when_disabled`/`_when_unset`/`_loads_kokoro_when_enabled` keep their `[]`/`[]`/`["v1"]` contracts; the wrapper calls `kokoro._ensure_loaded(model, device=device)` against a `monkeypatch.setattr`-installed *instance* attribute, so no `self` is passed and the 2-parameter lambda binds correctly.
- **No other stale `_ensure_loaded` double.** Only two doubles in the suite touch Kokoro: `test_design_kokoro_exclusion.py:373`'s `lambda model: None` (reached only through `KokoroEngine.synthesize`'s positional `self._ensure_loaded(model)`, never the wrapper) and `test_load_admission.py:153`'s `_FakeLoadKokoro._ensure_loaded(self, model, device=None)`. Both fine. `test_kokoro.py:2094`'s `kokoro_unloaded_client` drives the real method with fake weights. Nothing else needs widening.
- **No deadlock in the new nesting that I could construct.** `KokoroEngine.unload()` is lock-free, `/unload`'s kokoro arm (`main.py:11411`+) doesn't take `_load_lock`, and `design_voice`'s in-`design()` Kokoro eviction calls `unload()` (never a load) — so `kokoro_synth()` is never entered from inside `design()`, which is the one nesting that would hang.
- **Out of scope, correctly:** no `cast.json` writes, no config knob, no derived artifact (`api-types.ts` / `docs/BACKLOG.md` / brand PNGs).

---

### 🔴 Correctness bugs

#### B1 — the `PRELOAD_KOKORO` half of the fix has no test that can detect its removal, and the PR's mutation evidence can't discriminate it

`server/tts-sidecar/main.py:10260`, `server/tts-sidecar/tests/test_kokoro.py:2254-2258`

Revert **only** line 10260 back to `await asyncio.to_thread(kokoro._ensure_loaded, "v1")` and the whole sidecar suite stays green:

- `test_load_kokoro_arbiter_gate.py` posts `/load` and nothing else — it never runs the startup preload path.
- `_run_preload_capturing_kokoro`'s stub, *as widened by this very PR*, is `lambda model, device=None: calls.append(model)`. `device` is optional, so the reverted 1-positional-arg call binds fine, and all three assertions are on `model` alone (`[]`, `[]`, `["v1"]`).
- `git grep _kokoro_ensure_loaded_guarded -- server/tts-sidecar/tests/` returns one hit: prose in the new file's module docstring. No test names the wrapper.

The PR body's mutation proof reverted **both** call sites together and re-ran one test. That measures the `/load` site and reports the result for both — the "control never measured" shape. Half of the claimed fix is unpinned.

Second half of the same finding, and the reason B1 may resolve as a *doc* fix rather than a test: **the preload site may not be a live bypass at all.** `_preload_default_engines()` runs inside the FastAPI lifespan startup (`main.py:1023-1032`: `_startup_ort_cuda_preload` → `_configure_vd_kokoro_coupling` → … → `_preload_default_engines`), which completes before uvicorn accepts a request. No `design_voice()` can be in flight, so `_design_active_count` is necessarily 0 and the guard is unconditionally a no-op there. If that's right, three places overstate it: the wrapper docstring (`main.py:11218` — "`/load` **and** the startup preload path … bypassing it entirely"), register row A107 ("routes both bypass call sites through it"), and the PR body ("routed both bypass call sites"). Pick one — pin the preload path with a real test, or reword all three to say the preload guard is defensive symmetry rather than a closed bypass.

#### B2 — the new test's `_shares_device` pin is dead: the lifespan overwrites it, so the gate's verdict is decided by the box's env

`server/tts-sidecar/tests/test_load_kokoro_arbiter_gate.py:50-55`

```python
    # Pin the single-card-box default regardless of what the real startup
    # coupling hook resolves on this test box (mirrors
    # test_design_kokoro_exclusion.py's autouse fixture).
    main._VD_KOKORO._shares_device = True          # line 53
    main._reset_poison_for_test()
    with TestClient(main.app) as c:                # line 55 — lifespan runs HERE
```

`TestClient.__enter__` enters the lifespan, whose second handler is `await _configure_vd_kokoro_coupling()` (`main.py:1024`), which does `_VD_KOKORO._shares_device = _compute_vd_kokoro_shares_device()` (`main.py:1627`). The pin at line 53 is discarded before the test body ever runs, so the comment is false on its own terms. It does **not** mirror `test_design_kokoro_exclusion.py:23-35` — that fixture works because its tests never enter a lifespan, and its docstring documents this exact hazard verbatim: *"Other test files' `TestClient(main.app)` usage fires the real `_configure_vd_kokoro_coupling` startup hook against the module-level `_VD_KOKORO` singleton, which can flip `_shares_device` to False on a box with no CUDA (or QWEN_DEVICE/KOKORO_DEVICE pinned to different cards)."*

Concrete failure. `shares_device` (`main.py:5826-5841`) returns **False** whenever either side resolves to a non-`cuda` family or a different index. `_compute_vd_kokoro_shares_device`'s `except: return True` escape hatch only fires when `import torch` raises — i.e. on CI's deliberately lean sidecar install. It does **not** fire on the box `npm run test:sidecar` actually runs on (full venv, torch present — that is the whole reason pre-push keeps that step local). Two real configurations there:

- **(a)** `QWEN_DEVICE=cuda:1` — this box's *documented standing pin*, per this PR's own neighbouring register row A106: "`QWEN_DEVICE` must be temporarily set to `cuda:0` instead of the box's own standing `cuda:1` pin" — with `KOKORO_DEVICE` unset → `auto` → `cuda:0`. Index 1 ≠ 0 → **False**.
- **(b)** torch importable, CUDA unavailable → both `auto` resolve to `cpu` → `fam_a != "cuda"` → **False**.

With `_shares_device` False, `kokoro_synth()` and `design()` are no-op context managers (`main.py:1579-1581`, `:1595-1597`), `/load` completes immediately, `load_thread.join(timeout=0.5)` returns, and line 89 fails with `"/load completed while a VoiceDesign forward was still active — the arbiter gate was bypassed (#3086/#3101)"`. A red that reads exactly like the fix regressing, on a box where the fix is fine.

Fix: pin *inside* the `with` (after `__enter__`), or `monkeypatch.setattr(main, "_compute_vd_kokoro_shares_device", lambda: True)` before constructing the client — and restore the prior value on teardown (see N2).

#### B3 — the fixture doesn't neutralise `PRELOAD_KOKORO`, so an operator env with it set fails the test on the `/load` fast path

`server/tts-sidecar/tests/test_load_kokoro_arbiter_gate.py:45-58`

The fake is installed into `main.ENGINES` at line 49, **before** the lifespan at line 55. `_preload_default_engines()` reads `_parse_bool(os.environ.get("PRELOAD_KOKORO"), False)` straight off the process env (`main.py:10255`), and `server/tts-sidecar/conftest.py` neutralises `PRELOAD_COQUI`, `COQUI_PIN_IMPORT_ORDER`, `QWEN_DEGEN_GUARD`, `COQUI_DEGEN_GUARD` and `SEG_CAPACITY_ADMISSION` — but **not** `PRELOAD_KOKORO`.

With `PRELOAD_KOKORO=1` exported (a registry-backed knob, `server/src/config/registry.ts:789`, surfaced in Advanced Settings as "Preload Kokoro at startup"), startup calls `_kokoro_ensure_loaded_guarded(fake, "v1")` → `fake._kokoro = object()` and `load_calls == [None]` before the test body starts. `/load` then short-circuits on the `if kokoro._kokoro is not None: return {"status": "ready"}` fast path just above `main.py:11278`, never touching the arbiter → line 89 (`assert load_thread.is_alive()`) fails, and so does line 93 (`assert load_calls == []`).

Both in-repo precedents guard this explicitly and this fixture does neither: `test_load_admission.py:47-49` (`monkeypatch.delitem(main.ENGINES, "kokoro", raising=False)` — *"so TestClient's startup event doesn't try to eager-preload it"*) and `test_kokoro.py:2094-2105`'s `kokoro_unloaded_client` (unloads again after `__enter__`, *"would otherwise warm Kokoro out from under these tests"*).

Fix: `monkeypatch.setenv("PRELOAD_KOKORO", "0")` in the fixture.

#### B4 — *(uncertain)* the guard is unconditional on the admitted device: a `cpu`-admitted Kokoro load now blocks behind a GPU design it cannot contend with

`server/tts-sidecar/main.py:11226` (the `with`), `:11292-11293` (the admitted call site)

`SEG_CAPACITY_ADMISSION` defaults ON. During a design the VoiceDesign 1.7B holds ~4-5 GB of the 8 GB card, so `_placement.reservation("kokoro", …, cpu_capable=True, heavy=False)` frequently can't fit a GPU and admits `device="cpu"` — the documented fallback, pinned by `test_load_admission.py::test_load_kokoro_falls_back_to_cpu_when_no_gpu_room` (asserts `load_calls == [("v1", "cpu")]`, 200 not 503). `_kokoro_ensure_loaded_guarded` then takes `kokoro_synth()` anyway and parks for the whole design.

A CPU-EP ONNX session allocates no VRAM and never reaches `_directml_selftest_or_fallback`'s forward (that branch is gated on `"DmlExecutionProvider" in providers`), so there is nothing for the arbiter to exclude — and the CPU fallback exists precisely so Kokoro can make progress when the GPU is full. The guard removes exactly that property.

Uncertain because `KokoroEngine.synthesize()` is likewise unconditional and `_shares_device` is env-derived rather than admission-derived, so this may be a deliberate consistency choice. But nothing in the PR body, the wrapper docstring, or register row A107 says the decision was made. Either gate the `with` on the resolved device, or state the trade.

#### B5 — *(uncertain)* an unbounded, starvable wait now sits on a user-driven request path with a hard 90s client ceiling and a sticky `_loading` flag

`server/tts-sidecar/main.py:11226` (the wait), `:11278`/`:11281`/`:11300` (what stays held)

`kokoro_synth()`'s `self._cv.wait()` (`main.py:1584`) has no timeout, and `design()` deliberately does **not** wait on `_design_active_count` (its own comment, `main.py:1594-1612`) — so during a bulk "Design full cast" loop a new design can re-enter before the parked `/load` thread reacquires `_cv`. The `/load` can be starved for the length of the whole run.

Node's `/api/sidecar/load` aborts at 90s (`server/src/gpu/capacity-retry.ts:71`, `:340`, `:359`). Concretely: the operator clicks Load Kokoro mid-cast-review, the button reports failure at 90s, and the sidecar keeps `kokoro._load_lock` held and `kokoro._loading = True` — surfaced as `/health`'s `kokoro_loading` (`main.py:10907`), so the pill spins indefinitely — until the design run ends, then loads Kokoro anyway. A Stop pressed during that window does nothing: `/unload`'s kokoro arm (`main.py:11411`+) is lock-free and no-ops on a not-yet-resident engine, so the load lands after it.

Every other contended wait in this file is bounded and reports a retryable error (`_BASE17_CONTENTION_WAIT_S_DEFAULT` → `Base17ContentionTimeoutError`, "Retry the design shortly"; `withKeyLock`'s 10s cap), and `/load` already has a 503 "retry shortly" shape (`_restart_pending`).

Uncertain because `/synthesize` already carries the identical unbounded wait, so this may be an accepted house pattern. The distinction I'd want ruled on: `/synthesize` is a server-driven retryable call, `/load` is a user button with a hard client ceiling and a flag that stays true after the client is gone.

---

### 🟠 House gates

#### H1 — release-notes pair missing, with no explicit not-applicable

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the PR body contains no "no shippable delta" statement. CLAUDE.md before-shipping step 5 requires one or the other, explicitly.

This isn't a borderline internal chore: the delta is operator-visible (a Kokoro load during a cast-design session no longer races the resident VoiceDesign forward on a shared-device box), `docs/release-notes-next.md` is an open, actively per-PR-maintained v1.15.0 cycle file — not a cut-time reconstruction — and it already carries entries naming this same arbiter.

*(Not a finding, a pre-merge reminder: the live view carries a fresh publish token — `data-published-as="18"`, id `995345f313ac`. Per CLAUDE.md step 3 the styled page still has to be published to the URL recorded in the register header, and `npm run check:onbox-register -- --against-published <saved copy>` run immediately before that publish. The PR body records `check:onbox-register` and `register:build` but not the `--against-published` comparison.)*

---

### 🔵 Cleanup nits

**N1 — `PR TBD` in the A107 heading.** `docs/testing/onbox-acceptance-register.md:4522`, and the live view's A107 `.src` line likewise names only #3086/#3101. It is the **only** `PR TBD` in the whole register (exactly 1 occurrence); A105 and A106 both carry `PR [#NNNN]`. The number is known — backfill `PR [#3142]` in both surfaces before merge, or whoever picks up A107 on the box can't find the change that owed it.

**N2 — `_shares_device` is written to the module singleton and never restored.** `test_load_kokoro_arbiter_gate.py:53`. `test_design_kokoro_exclusion.py:33-36` — the fixture this one cites — saves `prior` and restores it on teardown. Fold into B2's fix.

---

### Verdict

**Changes requested.** Three correctness bugs (B1, B2, B3) all sit in the paired regression test and its coverage — which is the part of this PR that has to hold, because the behaviour itself can't be proven off-box and is being carried as owed acceptance (A107). Two further correctness findings (B4, B5) are labelled uncertain and are decisions rather than defects; a one-line answer either way is enough. One house gate (H1, release notes) and two nits.

The production change itself — `_kokoro_ensure_loaded_guarded` and the two call-site rewrites — is correct as written, and the diagnosis in the PR body and A107 (the DirectML self-test forward as the real hazard, not mere bookkeeping) checks out against `_ensure_loaded`'s actual code.

Per the triage rules: B1-B3 are actual correctness bugs, so once they're fixed and pushed this **re-triggers a review pass**. H1/N1/N2 alone would not have.

dudarenok-maker:
## PR review — pass 2 (head 9e71f554b57d3f918916b502e9977c73a6d2115f, depth high)

**How the re-review was done.** Every disposition below is verified by execution, not by reading the diff. I copied `server/tts-sidecar/` to an OS-temp scratch tree and ran the real suite against mutated copies there, so nothing in the worktree was touched (`git rev-parse HEAD` + `git status --porcelain` identical before and after). Baseline on a **torch-equipped** box: full sidecar suite `1162 passed, 5 skipped, 10 deselected` — worth stating because CI's sidecar leg installs without torch, where `_compute_vd_kokoro_shares_device`'s `except: return True` fires, so **CI could never have caught B2 and cannot confirm its fix either.**

---

### Prior findings

#### B1 — test half **VERIFIED RESOLVED**; doc half **still open** (see B6)

Mutation-verified independently. In the scratch copy I reverted **only** `main.py:10260` back to `await asyncio.to_thread(kokoro._ensure_loaded, "v1")`, leaving both `/load` guard sites (`:11293`, `:11296`) intact, and ran the file:

```
FAILED tests/test_load_kokoro_arbiter_gate.py::test_startup_preload_kokoro_uses_guarded_path
E  AssertionError: startup preload didn't use _kokoro_ensure_loaded_guarded correctly; calls = []
E  assert [] == [('v1', None)]
```

…while `test_load_kokoro_blocks_while_design_active` stayed green. That is exactly the discrimination pass 1 said was missing. The two tests are complementary: the new one pins the startup **call-site wiring** (it monkeypatches the wrapper out, so it says nothing about the wrapper's body), and the `/load` test pins the wrapper's **body** — emptying `_kokoro_ensure_loaded_guarded` of its `with _VD_KOKORO.kokoro_synth():` reds the `/load` test. Between them, no single-site revert of this fix survives.

**The fix agent's structural claim is correct — I checked it independently, and it is stronger than stated.** `uvicorn/server.py:104` awaits `self.lifespan.startup()` before `loop.create_server(...)` at `:142`, so no socket exists until every startup handler has returned; `_preload_default_engines()` is the 9th of ten sequential `await`s in `_lifespan` (`main.py:1023-1032`) and has no other caller in the module; the only two `_VD_KOKORO.design()` entries in `main.py` are `:7306` (`design_voice`) and `:7701` (`mint_variant`), both request-driven. So `_design_active_count` is necessarily `0` at `main.py:10260`. Cross-process doesn't rescue it either: `_VD_KOKORO` is a **process-local in-memory** arbiter (`main.py:1619`), so a second sidecar's design is invisible to it. The startup guard is defensive symmetry, permanently. That is the right thing to keep — but three surfaces still say otherwise, and that is now a confirmed defect rather than a suspicion: **B6**.

#### B2 — **VERIFIED RESOLVED**, and load-bearing

Static: `_shares_device` is written in exactly two places — the constructor (`main.py:1569`) and `_configure_vd_kokoro_coupling` (`main.py:1627`) — and the latter has exactly one caller, the lifespan at `main.py:1024`. Nothing re-runs it after `TestClient.__enter__`, so the pin at `test_load_kokoro_arbiter_gate.py:61` holds for the whole test body.

Empirical, on this torch box with pass 1's configuration (a) exported — `QWEN_DEVICE=cuda:1`, `KOKORO_DEVICE=cuda:0`:

- current fixture → **2 passed**;
- scratch copy with pass 1's pre-`__enter__` placement restored → **FAILED**, `/load completed while a VoiceDesign forward was still active — the arbiter gate was bypassed (#3086/#3101)`.

So pass 1's B2 was a real red on a real configuration, and the move closes it.

Restore is correct too: the `try/finally` sits *inside* the `with`, so it runs before `TestClient.__exit__`, and `prior` is the **post-lifespan** value — i.e. the fixture hands back exactly what the box's own coupling hook computed, not a stale module default. No cross-file leak: the full 1162-test suite is green in file order.

#### B3 — **VERIFIED RESOLVED**, ordering correct and load-bearing

`monkeypatch.setenv("PRELOAD_KOKORO", "0")` is at `test_load_kokoro_arbiter_gate.py:50`, five lines before `TestClient(main.app)` at `:55`, and `_preload_default_engines` reads `os.environ.get("PRELOAD_KOKORO")` at call time (`main.py:10255`) — so the neutralisation lands before the lifespan reads it. Confirmed empirically both directions with `PRELOAD_KOKORO=1` exported in the process env:

- current fixture → **2 passed**;
- scratch copy with that one line deleted → **FAILED** at `assert load_thread.is_alive()`, the fast-path short-circuit pass 1 predicted.

Also confirmed `conftest.py` neutralises `PRELOAD_COQUI` / `COQUI_PIN_IMPORT_ORDER` / `QWEN_DEGEN_GUARD` / `COQUI_DEGEN_GUARD` / `SEG_CAPACITY_ADMISSION` but **not** `PRELOAD_KOKORO`, so the fixture is the right place for it and matches the two in-repo precedents.

#### H1 — **RESOLVED as a gate**; the user-facing half is wrong (see B7)

Both files moved, in this PR. `docs/release-notes-next.md:592` is accurate and sits with its own chain (#2790, #2810). `RELEASE_NOTES.md:47` is in the in-progress `# Castwright 1.15.0` section. The gate is satisfied; the *content* of the user-facing line is not — **B7**.

#### N1 — **VERIFIED RESOLVED**

`git grep -c "PR TBD"` is now **0** in both surfaces. Register heading (`:4522`) carries `PR [#3142](…/pull/3142)`; the live view's A107 `.src` carries the matching `<a href=…/pull/3142>PR #3142</a>`.

#### N2 — **VERIFIED RESOLVED**

`try/finally` at `test_load_kokoro_arbiter_gate.py:63-66`. Covered under B2.

#### B4 / B5 — **still open: no disposition recorded anywhere**

Pass 1 labelled both uncertain and asked for a one-line ruling, not code. Neither `c3dbed69` nor `9e71f554` nor the PR body nor any comment answers either:

- **B4** — `_kokoro_ensure_loaded_guarded` parks a **cpu-admitted** Kokoro load behind a GPU design it cannot contend with (`main.py:11226` / `:11292-11293`). `test_load_admission.py::test_load_kokoro_falls_back_to_cpu_when_no_gpu_room` is the path.
- **B5** — an **unbounded, starvable** `self._cv.wait()` (`main.py:1584`) now sits on a user-driven request path with a hard 90s client ceiling (`server/src/gpu/capacity-retry.ts:71`) and a sticky `kokoro._loading` flag.

Either answer is fine ("deliberate, mirrors `synthesize()`" closes both). What is not fine is merging with them unanswered — a later reader cannot tell a decision from an oversight.

---

### 🔴 New correctness findings

#### B6 — A107's acceptance criteria offers an on-box repro arm that **cannot be constructed**, and would be signed off on a null observation

`docs/testing/onbox-acceptance-register.md:4540` (Criteria), `:4535` (body), `:4526` (body), `docs/testing/onbox-acceptance-register-live-view.html:728`, `server/tts-sidecar/main.py:11217-11219` (wrapper docstring).

The Criteria reads: *"re-run the #3086 repro — concurrent VoiceDesign forward + raw Kokoro `/load` **(or startup preload)** on a shared-device box — and confirm the Kokoro load now blocks until the design releases."* The body likewise: *"a live Kokoro `/load` **(or `PRELOAD_KOKORO=1` startup)** racing a real, resident VoiceDesign forward."*

Per the trace under B1, the second arm is unconstructible. The concrete failure: an operator picks up A107, chooses the `PRELOAD_KOKORO=1` arm, and goes looking for "a real, resident VoiceDesign forward" to race the startup preload against. There is no such state — the preload only runs in a **freshly started** sidecar whose `_VD_KOKORO` is a brand-new object with `_design_active_count == 0`, and any design running in a *different* process is invisible to an in-memory arbiter. They observe Kokoro loading straight through, and record either **a false pass** ("no interference — gate works") or **a false regression** ("it didn't block"). Both outcomes are worthless, and A107 is precisely the instrument that carries this PR's unprovable half.

This is the one place where the fix round contradicts itself: `c3dbed69`'s own commit message concludes *"PRELOAD_KOKORO startup race is structurally no-op … making the guard defense-in-depth rather than a closed real-world race, consistent with A107's body documentation"* — but A107's body is not consistent with that. `:4526` says *"the actual bypass was `POST /load {"engine":"kokoro"}` **and the `PRELOAD_KOKORO` startup path**"*, and `main.py:11217-11219`'s docstring frames both call sites as live bypasses. The conclusion was reached and then not written down anywhere it would be read.

**Fix (docs only, this PR):** drop `(or startup preload)` / `(or PRELOAD_KOKORO=1 startup)` from the Criteria and the body sentence in **both** register surfaces — the live view is a separate hand-authored edit, and moving the register alone leaves the published page still instructing the unconstructible run — and add one clause to the wrapper docstring saying the preload site is symmetry, a no-op today because the lifespan completes before uvicorn opens the socket. The register/live-view edit needs a fresh `npm run stamp:publish-token` on top, same as this round's.

#### B7 — the user-facing release note states the precondition **backwards** and names a control that does not exist

`RELEASE_NOTES.md:47`.

> *"On a two-graphics-card machine, manually loading Kokoro (the Manual Load button on the Status panel) while a voice design was running could land both on the same card…"*

Two errors:

1. **The affected population is inverted.** The arbiter is live only when `_compute_vd_kokoro_shares_device()` is true (`main.py:5853-5864` → `shares_device`, `:5826-5841`), i.e. when `QWEN_DEVICE` and `KOKORO_DEVICE` resolve to the **same** `cuda:N`. That is the single-card default — and it is A107's own stated prerequisite (*"single 8 GB GPU card"*) and #3086's repro box. On a genuine two-card machine carrying this repo's documented standing pin (`QWEN_DEVICE=cuda:1`, per neighbouring register row A106), `shares_device` returns **False**, the arbiter is a no-op pair of context managers (`main.py:1579-1581`, `:1595-1597`), and this entire fix is inert. So the note tells the single-card users who *are* affected that this is about somebody else's hardware. (The sentence looks like it was patterned on the existing two-card entry six bullets above it, which genuinely is multi-GPU.)
2. **"the Manual Load button on the Status panel" does not exist.** `src/components/ModelControlPill.tsx:160` renders `Load model`, with `aria-label` `Load model Kokoro` when `engineLabel` is set. A user told to look for "Manual Load" will not find it.

**Fix:** scope it to "a machine where voice design and Kokoro share one graphics card (the default setup)" and name the control by its real label. While reworking it, the trailing clause *"protecting the design's real-time interactivity the same way a background chapter render already was"* is ambiguous in the wrong direction — what was already true is that a background chapter render (`KokoroEngine.synthesize()`) **waits for** the design, not that the render was protected.

---

### 🔵 Cleanup nits

**N3 — the new test's docstring hardcodes a line number.** `server/tts-sidecar/tests/test_load_kokoro_arbiter_gate.py:119`: *"if someone removes ONLY the guard from the startup preload path (line 10260)"*. Correct today; silently false after any insertion above it in an 11k-line file. Name the symbol (`_preload_default_engines`'s Kokoro branch) instead.

**N4 — the PR body's Test plan is now stale evidence.** It still presents only the pass-1 mutation run (*"reverted **both** guarded call sites … and reran the new test"*) — the exact evidence pass 1 flagged as unable to discriminate the two sites — and never mentions `test_startup_preload_kokoro_uses_guarded_path`, which is the test that actually closes B1. Update it so the body's claim matches the branch.

**N5 — publish/`--against-published` step still unrecorded** (restating pass 1's pre-merge reminder, now that the token moved again). The PR body records `check:onbox-register` and `register:build` but not `npm run check:onbox-register -- --against-published <saved live page>` run immediately before publishing, which CLAUDE.md step 3 requires. B6's fix will bump the token a third time, so do this once, at the end.

---

### Verified this pass (so a later pass doesn't re-derive it)

**The publish-token mechanics are genuinely correct this time, not passing by luck.** `origin/main` (`2b63b451`) carries `data-published-as="17" data-publish-id="00e4ce4cce1b"`; the branch carries `"19"` / `"cf680620f38e"`. `checkStampedSince` (`scripts/check-onbox-register.mjs:569`) returns clean only when content is unchanged, or the counter strictly increased **and** the nonce changed — both hold, and neither the "counter behind", "counter equal", nor "hand-edited counter, nonce unchanged" branch is reachable here. Ran locally, all green: `check-onbox-register.mjs` (plain), `--stamped-since origin/main`, `--stamped-since 80354c75` (the merge-base), and `build-register-live-view.mjs --check` → *"up to date"*. I also confirmed `main`'s live view is **byte-identical** between the merge-base and `origin/main`, so the `merge(base, head)` comparison CI actually performs is equivalent to the one I ran by hand. All 22 PR checks are green, including *Check on-box acceptance register consistency*.

**Nothing new introduced by the fix commits.** `c3dbed69` touches 5 files (no production code — `main.py` is untouched by this round), `9e71f554` touches only the token. The co-author drift the dispatch asked about (`Claude Haiku 4.5` on `c3dbed69`, `Claude Sonnet 5` on `9e71f554`; author/committer is the repo owner on both) is within this repo's routing table for a well-specified fix and is not a defect — though it is worth noting that the *one* part of this round that came out wrong (B6) is the part that needed a judgment call rather than a mechanical edit.

---

### Verdict

**Changes requested — but narrowly.** All three of pass 1's correctness bugs (B1, B2, B3) are **verified resolved by execution**, not by inspection, and B2's and B3's fixes are each demonstrably load-bearing under a real box configuration. H1, N1 and N2 are resolved. The register/live-view publish-token handling is correct.

What remains is **documentation, not code**: B6 (an acceptance criterion that cannot be performed — the highest-stakes of the three, because A107 is the only instrument carrying this PR's unprovable half), B7 (a user-facing note that misstates who is affected and invents a button), and a one-line ruling on B4/B5. The production change and its test coverage are, as of this head, sound.

dudarenok-maker:
## PR review — pass 3 (head 1d880cbff3f66bdb61b9c38d02c1e1269f156268, depth high)

**How this pass was done.** Same method as pass 2. Nothing in the worktree was modified: `git rev-parse HEAD` + `git status --porcelain` identical before and after, re-checked after every test run. All mutation work ran against copies of `server/tts-sidecar/` in an OS-temp scratch tree, driven by the primary checkout's venv python (the worktree has no `.venv`). Every disposition below is verified by execution or by reading the cited code, not by reading the diff.

---

### Prior findings

#### B6 — **VERIFIED RESOLVED** on all three surfaces, and the docstring's claim re-derived independently

All four edits landed, and nothing of the removed language survives — `git grep -n "startup preload\|PRELOAD_KOKORO=1 startup\|or startup"` over both register surfaces returns **0 hits**.

- `docs/testing/onbox-acceptance-register.md:4538` (Criteria) — `(or startup preload)` gone.
- `docs/testing/onbox-acceptance-register.md:4525-4533` (body) — `and the PRELOAD_KOKORO startup path` gone; `routes both bypass call sites` → `routes the /load bypass`; the closing "unit tests cannot prove" sentence no longer offers the `PRELOAD_KOKORO=1` arm.
- `docs/testing/onbox-acceptance-register-live-view.html:726` (body `<p>`) **and** `:728` (the Criteria `<li>`) — both corrected, matching the markdown. Pass 2 named this file as one surface; it actually carries the claim twice and **both** were fixed.
- `server/tts-sidecar/main.py:11222-11233` (wrapper docstring) — now says the preload site goes through the wrapper "for consistency, though the lifespan startup completes before uvicorn accepts requests".

**I re-derived the structural claim from scratch rather than inheriting pass 2's.** `uvicorn/server.py:104` (`await self.lifespan.startup()`) runs before `loop.create_server(...)` at `:142`/`:149`, so no socket exists until startup returns; `start.ps1:137` confirms the sidecar really is `python -m uvicorn main:app`. `_preload_default_engines()` is the 9th of ten sequential awaits in `_lifespan` (`main.py:1023-1032`) and has exactly one caller (`git grep` gives `:1031` plus its own `def` at `:10223`). `git grep _VD_KOKORO` over `main.py` shows exactly two `design()` entries — `:7313` (`design_voice`) and `:7708` (`mint_variant`) — both request-driven; no watchdog or lifespan task enters it. And `main.py:10162` already carried an independent statement of the same fact ("uvicorn binds only after lifespan startup returns"), so the new docstring is consistent with the file's own prior finding rather than a fresh assertion. The docstring is accurate.

#### B7 — **VERIFIED RESOLVED** on both named errors; two of pass 2's fix instructions were not carried out (→ N6, N7)

`RELEASE_NOTES.md:47`, both corrections checked against the source, not taken on trust:

1. **Population.** `_compute_vd_kokoro_shares_device` (`main.py:5860-5871`) is `shares_device(_read_device_env("QWEN_DEVICE"), _read_device_env("KOKORO_DEVICE"))` — true exactly when the two resolve to the **same** card. "a machine where voice design and Kokoro share one graphics card (the default setup)" is the right condition, and it matches A107's own `single 8 GB GPU card` prerequisite. Correct.
2. **Control label.** `ModelControlPill.tsx:160` — `case 'idle': return { label: 'Load model', handler: 'load', disabled: false }`, and `actionText` (`:192-194`) is the bare `action.label` unless `leadWithAction` is set. The Kokoro pill is constructed at `layout.tsx:1237-1251` **without** `leadWithAction`, and that cluster is `ttsPillElement` → `statusDetail.ttsControls` (`layout.tsx:1590`), i.e. the Status pill's popover. So both "Load model" and "on the Status panel" are accurate. Correct.

#### B4 — ruling **RECORDED**; the decision is sound, its stated premise is not (→ N8)

`main.py:11234-11238`. The load-bearing half of the ruling checks out: with `device="cpu"`, `_ensure_loaded` hits `family == "cpu"` → `providers = ["CPUExecutionProvider"]` (`main.py:3587-3588`), so the `"DmlExecutionProvider" in providers` gate at `:3724` is false and `_directml_selftest_or_fallback`'s forward never runs. **A CPU-admitted Kokoro load genuinely does not contend**, exactly as the comment says. The blanket-lock decision stands. See N8 for the one clause that does not.

#### B5 — ruling **RECORDED**, but the ruling's own justification is wrong and it is written into the wrong function (→ **B9**, below)

The decision ("accepted, not a correctness bug") is defensible. The three sentences supporting it are not, and they are now the only record of the decision. Filed as a new finding rather than a resolution.

#### B1 / B2 / B3 / H1 / N1 / N2 — still resolved

`git diff 9e71f554 HEAD --stat` is 4 files; `test_load_kokoro_arbiter_gate.py` is untouched this round, so pass 2's execution-verified dispositions stand. Re-ran `test_load_kokoro_arbiter_gate.py` + `test_design_kokoro_exclusion.py` + `test_load_admission.py` + `test_mint_variant_kokoro_stall.py` + `test_lifespan_order.py` against the worktree: **35 passed**. `main.py`'s diff this round is comment/docstring text only — no logic moved.

---

### 🔴 New correctness findings

#### B8 — `/load`'s **default-configuration** call site is unpinned: removing its guard leaves the entire sidecar suite green

`server/tts-sidecar/main.py:11305-11307` (the guarded call), `server/tts-sidecar/tests/test_load_kokoro_arbiter_gate.py:51`, `server/tts-sidecar/conftest.py:40`

`/load`'s Kokoro branch has **three** call sites of `_kokoro_ensure_loaded_guarded`, not the two this PR's body, the register and the reviews have all been reasoning about:

```python
if _capacity_admission_enabled():                      # main.py:11296
    async with _placement.reservation(...) as adm:
        ...
        await asyncio.to_thread(
            _kokoro_ensure_loaded_guarded, kokoro, "v1", adm["device"]   # :11306  ← ADMITTED
        )
else:
    await asyncio.to_thread(_kokoro_ensure_loaded_guarded, kokoro, "v1") # :11309  ← admission OFF
```

`_capacity_admission_enabled()` is `os.environ.get("SEG_CAPACITY_ADMISSION", "1") != "0"` (`main.py:5377-5378`) — **ON by default in production**, so `:11306` is the line a real operator's Load-model click executes. `:11309` is the opt-out branch.

The new regression test only ever reaches `:11309`. `conftest.py:40` sets `SEG_CAPACITY_ADMISSION=0` suite-wide, and `test_load_kokoro_arbiter_gate.py:51` sets it to `"0"` again in its own fixture (which is also why `load_calls == [None]` at `:112`, not `("v1", <device>)`).

**Mutation-proven, with the control measured.** In a scratch copy I reverted **only** `:11306` to `await asyncio.to_thread(kokoro._ensure_loaded, "v1", adm["device"])`, leaving `:11309` and the preload site guarded:

- `test_load_kokoro_arbiter_gate.py` + `test_design_kokoro_exclusion.py` + `test_load_admission.py` + `test_mint_variant_kokoro_stall.py` → **30 passed, exit 0.**
- Whole `tests/` directory → the only failures were `test_footprints.py` and `test_speaker_embed.py`, and both are artifacts of my copy excluding `spikes/` and `docs/` — I confirmed both files pass in the real worktree (**28 passed, exit 0**). Nothing else went red.
- Reporting the one run that did go red elsewhere, since omitting it would be selective: a third run of the same mutant with those two files `--ignore`d failed on `test_asr_footprint_measurement.py::test_non_asr_engine_still_uses_the_torch_allocator_path`. That is not the mutation — the test drives a **coqui** reservation and stubs `PlacementController._device_free_mb`, touches no Kokoro or `/load` path, **passed** on the same mutant in the full-collection run above, and **passes alone on the mutant** (15 passed, exit 0). It is collection-order-dependent, consistent with `reservation`'s own docstring hazard about `asyncio.Lock` binding to the first loop that awaits it. A mutation detector does not pass and fail on the same mutant according to which unrelated files are collected.

So a partial revert of the default-configuration half of this fix is invisible to the whole suite. This is pass 1's B1 one rung deeper: B1 established that the preload site needed its own pin because the `/load` test could not discriminate it; the same argument applies to the admitted `/load` site, and nobody has made it until now. It matters here more than it normally would, because A107 exists precisely to say the behaviour itself cannot be proven off-box — the test coverage *is* the load-bearing part of this PR.

Note `test_load_admission.py::test_load_kokoro_falls_back_to_cpu_when_no_gpu_room` (`:191-210`) already builds the whole admitted-path harness (`monkeypatch.setenv("SEG_CAPACITY_ADMISSION", "1")` plus a stubbed `main._placement.probe`), so closing this is cheap: one test that holds `_VD_KOKORO.design()` on a thread while posting `/load` with admission **ON**.

#### B9 — the B5 ruling's stated reason contradicts its own next sentence, and it is written into a function `/load` is not the main caller of

`server/tts-sidecar/main.py:1583-1589` (the comment), `:3811` and `:11239` (the two callers), `:11291` / `:11294` / `:11298` / `:11313` (what actually stays held)

Two separate defects in the same seven lines.

**(a) Wrong location.** The comment sits inside `_VdKokoroArbiter.kokoro_synth()`. `git grep -n "with _VD_KOKORO.kokoro_synth():"` returns exactly two call sites: `main.py:3811` — `KokoroEngine.synthesize()`, i.e. every ordinary Kokoro sentence of every chapter render — and `main.py:11239`, the new `/load` wrapper. Stating there that the wait is "bounded in practice by the Node-side 90s client timeout on `/load` requests" is simply false for the first and dominant caller, and tells the next reader of `kokoro_synth()` that `/load` is the only thing that waits here. A ruling scoped to `/load` belongs at `_kokoro_ensure_loaded_guarded` (`:11234`, next to B4's), not in the shared arbiter.

**(b) The reason contradicts the sentence after it.** "bounded in practice by the Node-side 90s client timeout" followed immediately by "the sidecar thread remains in this wait until `design_active_count` drops". The 90s ceiling bounds *the client's* wait. It bounds neither this wait nor any of the state the wait holds. Concretely, on the single-card default box during a bulk "Design full cast" run: the operator clicks Load model, the button fails at 90s, and for the rest of the run —

- `kokoro._load_lock` stays held (`:11291` wraps the `to_thread`), so a retry click just queues behind the same park;
- `kokoro._loading` stays `True` (set `:11294`, cleared only in the `finally` at `:11313`), which `/health` publishes as `kokoro_loading` (`:10914` → `:10962`), so the pill spins with nothing behind it;
- Stop is inert — `/unload`'s Kokoro arm (`:11450-11454`) takes no lock and `KokoroEngine.unload()` no-ops on a non-resident engine, so the load still lands afterwards;
- **and, not previously named:** the `_placement.reservation` at `:11298` wraps the `to_thread`, so the VRAM hold is outstanding for the whole park. `PlacementController.reservation` releases in its `finally` after the `yield` (`:5307-5353`), and a *held reservation* — unlike a *resident* engine — cannot be evicted to make room. `mint_variant` explicitly unloads a resident Kokoro at `:7709-7711` for exactly this reason; it cannot do that to a hold. Pre-PR this hold lasted the ~1s of the load; it now lasts the design.

Every one of those is strictly outside what the 90s ceiling bounds, and the comment offers that ceiling as the reason the limitation is acceptable. The same wording is in the PR body's "Known limitations, accepted" section, so both copies need it.

**This is a documentation-class correctness finding, the same class pass 2's B6 and B7 were.** I am not asking for the ruling to be reversed — "accepted, suboptimal fairness" is a defensible call. I am saying the recorded justification is wrong, and it is the only record there is.

---

### 🔵 Cleanup nits

**N3 — now actually false, not merely fragile.** `tests/test_load_kokoro_arbiter_gate.py:119` still reads *"removes ONLY the guard from the startup preload path (line 10260)"*. This round's B5 comment inserted 7 lines at `main.py:1583`, so that call site is now at **`main.py:10267`**. Pass 2 predicted this would silently go false; this commit is what made it so. Under the repo's own rule a comment a change made false is a chore owed in the same round. Name `_preload_default_engines`'s Kokoro branch instead of a line number.

**N4 — PR body Test plan still stale** (restating pass 2). It still presents only the pass-1 two-site mutation — the evidence pass 1 flagged as unable to discriminate the sites — and still never names `test_startup_preload_kokoro_uses_guarded_path`, the test that actually closed B1. The new "Known limitations, accepted" section does not fix this.

**N5 — the `--against-published` step is still unrecorded** (restating pass 2's N5). The token moved again this round (see below). CLAUDE.md step 3 wants `npm run check:onbox-register -- --against-published <saved live page>` run immediately before publishing the styled page to the URL in the register header. Do it once, at the end.

**N6 — `docs/release-notes-next.md:592` still carries the framing B6 removed everywhere else.** It says the preload path was one of two things "bypassing `_VD_KOKORO.kokoro_synth()`" and that the fix routes "both bypass call sites through the same gate". Structurally that sentence is true and pass 2 passed it, but the register now deliberately says "the `/load` bypass" — so the two now disagree about whether the preload site was ever a bypass. Align the wording or add the same one-clause qualifier the docstring got.

**N7 — the reworked release-notes sentence is now circular, and the clause pass 2 asked to rework is unchanged.** `RELEASE_NOTES.md:47`: *"On a machine where voice design and Kokoro share one graphics card (the default setup), clicking the Load model button … could interrupt the design with no warning **if both ended up on the same card**."* The trailing conditional restates the premise, and reads as though there were a chance they would not. Separately, pass 2's B7 also asked for the trailing *"protecting the design's real-time interactivity the same way a background chapter render already was"* to be reworked — what was already true is that a background chapter render **waits for** the design, not that the render was protected — and it is verbatim unchanged.

**N8 — B4's ruling rests on a frequency claim that is unevidenced and probably inverted.** `main.py:11237-11238` calls the CPU-admitted path "this rare path (capacity admission hitting the GPU limit)". The gate only blocks while a design is active — i.e. while the VoiceDesign 1.7B holds ~4-5 GB of the 8 GB card, which is precisely when `reservation("kokoro", …, cpu_capable=True)` cannot fit the GPU and admits `"cpu"`. *Conditional on the guard blocking at all*, CPU admission is more likely than baseline, not rare; pass 1's B4 called it "frequently" and nothing has rebutted that. The decision survives the correction — a blanket lock is still the simpler thing — but the comment currently justifies it with the opposite of the conditional probability that matters.

---

### Verified this pass (so it is not re-derived)

**Publish-token mechanics are correct, for the third time and on the mechanism rather than by luck.** `origin/main` (`2b63b451`, re-fetched this pass, unmoved since pass 2) carries `data-published-as="17" data-publish-id="00e4ce4cce1b"`; the merge-base `80354c75` carries the **byte-identical** line, so CI's `merge(base, head)` comparison is equivalent to the by-hand one; HEAD carries `"20"` / `"300d8fd24341"`. Counter strictly increased **and** nonce changed — the only branch of `checkStampedSince` that returns clean on changed content. The nonce moved, so the stamp came from `stamp:publish-token`, not a hand-edit, and it landed in the same commit as the content it stamps. Ran locally, all green:

```
node scripts/check-onbox-register.mjs                            → OK
node scripts/check-onbox-register.mjs --stamped-since origin/main → OK
node scripts/check-onbox-register.mjs --stamped-since 80354c75    → OK   (the merge-base)
node scripts/build-register-live-view.mjs --check                 → up to date
```

**The branch is behind `origin/main` but will not revert the register.** `git diff 80354c75 origin/main` over the two register surfaces is one line, in `onbox-acceptance-register.md:86` (PR #3133's `HEAD^1` prose), nowhere near A107 — a clean three-way merge, no lost row.

**All 22 PR checks are green at this head**, including *Sidecar tests (pytest)* (1m25s) — which is itself part of B8's point: CI's sidecar leg would not have caught that mutation either.

**Nothing else regressed.** `1d880cbf` touches 4 files. The `main.py` half is a docstring rewrite plus two comment blocks; no statement moved. No test asserts on either docstring. No `cast.json` write, no config knob, no derived artifact (`api-types.ts` / `docs/BACKLOG.md` / brand PNGs) in scope. `Closes #3086` is still present in the body, outside any code span.

---

### Verdict

**Changes requested — and this is the third pass, so per the `pr-review-gate` 3-pass cap it does not auto-trigger a fourth. The two open items below go to the repo owner to decide, not to another automatic round.**

- **B6 and B7 are genuinely resolved**, verified against the code and not the diff, including re-deriving B6's structural claim from uvicorn's own source rather than inheriting it.
- **B4's ruling is recorded and technically sound** (the "does not contend" half is verified true); only its frequency premise is wrong, which is N8, a nit.
- **B5's ruling is recorded but its justification is wrong** — B9. The conclusion can stand; the stated reason cannot, and it is in the wrong function.
- **B8 is new and is the substantive one**: the production-default `/load` path — the one an operator's click actually takes — is not pinned by any test, mutation-proven with the control measured. That is the same class of gap pass 1 raised as B1 and the reason pass 1's B1 was treated as blocking.

My recommendation, for the owner rather than for another agent round: **B8 is worth one more commit** (a second test in the existing file, admission ON, reusing `test_load_admission.py`'s stub-probe harness) — this PR's whole premise is that the behaviour is unprovable off-box, which makes the tests the deliverable. **B9 is a two-line edit** (move the comment to `_kokoro_ensure_loaded_guarded` and say what actually stays held instead of citing the 90s ceiling as a bound). N3 is owed under this repo's own stale-comment rule. N4-N8 are optional.

The production change itself — `_kokoro_ensure_loaded_guarded` and all three call sites — remains correct as written. Nothing found in this pass says the code is wrong; what is wrong is what pins it and what records why.

dudarenok-maker:
### Pass 3 addendum — the B8 control, now measured

Closing the one loose end in the pass-3 comment above (head `1d880cbff3f66bdb61b9c38d02c1e1269f156268`). B8's finding is unchanged; this only completes its control arm, since "a control group labelled clean but never measured" is on this repo's own trap list.

The unmutated copy, run with the **identical** command that produced the one stray red on the mutant (`pytest tests/ -q --ignore=tests/test_footprints.py --ignore=tests/test_speaker_embed.py`), finished **exit 0 — fully green**. So the four observations on `test_asr_footprint_measurement.py::test_non_asr_engine_still_uses_the_torch_allocator_path` are:

| run | collection | result |
|---|---|---|
| mutant | whole `tests/` | **pass** |
| mutant | `--ignore` the two copy artifacts | fail |
| mutant | that file alone | **pass** (15 passed) |
| clean control | `--ignore` the two copy artifacts | **pass** (exit 0) |

Two passes out of three on the mutant, and no causal path from the mutated line — the test drives a **coqui** reservation and stubs `PlacementController._device_free_mb`, and never reaches `/load`'s Kokoro branch. It is non-deterministic, not a detector of the mutation. Reporting it in full rather than only the runs that agree with me.

**B8's own evidence is untouched by this**, and it is the narrow run that matters: with `main.py:11306` reverted to a raw `kokoro._ensure_loaded`, `test_load_kokoro_arbiter_gate.py` + `test_design_kokoro_exclusion.py` + `test_load_admission.py` + `test_mint_variant_kokoro_stall.py` are **30 passed, exit 0**. The file written to catch precisely that bypass does not catch it at the configuration production actually runs.

dudarenok-maker:
## PR review — pass 4 (head 73ea30effb3ea09520abf15888bc8c7ff9bb4e9e, depth high)

**Owner-authorized round beyond the 3-pass cap.** Pass 3 closed with B8 and B9 open and explicitly routed them to the repo owner rather than to another automatic round. This pass exists because the owner authorized one more, and it is scoped to verifying `73ea30ef` independently — not to re-deriving pass 1–3's dispositions, which stand.

**How this pass was done.** Same method as passes 2 and 3, and nothing in either tree was modified. Worktree `C:\Claude\Projects\wt-3086-vd-kokoro-arbiter-race`: `git rev-parse HEAD` = `73ea30ef` and `git status --porcelain` **empty**, before and after, re-checked at the end. Primary checkout `C:\Claude\Projects\Audiobook-Generator`: `HEAD` = `2b63b451`, `git status --porcelain` = the one pre-existing untracked `docs/testing/onbox-a-series-audit-2026-09-05.md` and nothing else, unchanged by this pass. All mutation work ran against a copy of `server/tts-sidecar/` in an OS-temp scratch tree (hash-verified identical to the worktree before each run and restored by re-copy, not by rewrite, so no newline drift), driven by the primary checkout's venv python — **torch 2.11.0+cu128, `torch.cuda.is_available()` True**, i.e. the box where `_compute_vd_kokoro_shares_device`'s `except: return True` escape hatch does *not* fire. I did not trust the fix agent's claims; every disposition below is from execution or from reading the cited code.

---

### B8 — **VERIFIED RESOLVED.** The guard was already there; the test that pins it is now real, and I proved it myself

**The shipped-code half first, read rather than taken on trust.** `main.py:11310-11321` — inside `if _capacity_admission_enabled():`, inside `async with _placement.reservation(...) as adm:` — is `await asyncio.to_thread(_kokoro_ensure_loaded_guarded, kokoro, "v1", adm["device"])`, and `_kokoro_ensure_loaded_guarded` (`:11233`) wraps its `kokoro._ensure_loaded(model, device=device)` in `with _VD_KOKORO.kokoro_synth():` at `:11253`. So the fix agent's claim is correct: **there was never a live bypass at the admission-ON site; only the coverage was missing.** Pass 3 said the same thing ("the production change itself … remains correct as written") — B8 was always a test-coverage finding, and it is closed as one.

**The class is closed, not just the instance.** `git grep -n "_ensure_loaded"` over `main.py` gives exactly five Kokoro load sites: `:3826` (inside `synthesize()`'s own `with _VD_KOKORO.kokoro_synth():` at `:3825`), `:11254` (inside the wrapper), and the three wrapper calls at `:10281` (startup preload), `:11320` (admission ON), `:11323` (admission OFF). There is no sixth, unguarded entry point.

**My own mutation, run independently of the fix agent's.** In the scratch copy I reverted **only** the admission-ON site back to `await asyncio.to_thread(kokoro._ensure_loaded, "v1", adm["device"])`, leaving `:11323` and `:10281` guarded:

```
FAILED tests/test_load_kokoro_arbiter_gate.py::test_load_kokoro_admission_on_blocks_while_design_active
E  AssertionError: /load completed while a VoiceDesign was still active (admission ON) —
   the arbiter gate was bypassed in the admission-enabled code path (#3086/#3101)
E  assert False
E   +  where False = is_alive()
```

One failure, and it is the new test, failing **for the reason claimed** — not for a collection artifact, an import error or a fixture bug. Restored by re-copying the pristine file (hash-verified equal to the worktree's) → green again.

**Complementary control, so the two tests are shown to discriminate rather than merely coexist.** I ran the mirror mutation: revert **only** the admission-OFF site (`:11323`), leave `:11320` and `:10281` guarded. That kills **only** `test_load_kokoro_blocks_while_design_active`, and the new admission-ON test stays green. So each `/load` branch is now pinned by its own test, and the startup site by `test_startup_preload_kokoro_uses_guarded_path` (pass 2 verified that one by the same method). **No single-site revert of this fix survives the suite any more** — which is what pass 3 said was not true at `1d880cbf`.

**The fixture genuinely reaches the admission-ON path; the env var is not the evidence.** `conftest.py:40` sets `SEG_CAPACITY_ADMISSION=0` via `os.environ.setdefault` at collection, and `load_client_with_admission` overrides it with `monkeypatch.setenv(..., "1")` before `TestClient(main.app)`; `_capacity_admission_enabled()` (`main.py:5391-5392` at *this* head — pass 3's `:5377-5378` is one of the citations the +14 shift moved) reads the env at call time. But the load-bearing proof is the *terminal* assertion, not the setenv: `assert fake_kokoro.load_calls == ["cuda:0"]`. `_FakeLoadKokoro._ensure_loaded` records the **device**, and the admission-OFF branch passes no device at all (the original test asserts `[None]`). A run that silently fell through to `:11323`, or that hit `noCapacity` and returned early, would produce `[None]` or `[]` and fail. The test cannot pass without the admission branch having supplied a resolved device.

**Not box-dependent.** `SEED_FOOTPRINTS_MB["kokoro"]` is 1200 (`main.py:4360`) against the stub probe's 5000 MB free, and `FootprintTable` has no on-disk persistence (`grep` for a footprints path over `main.py` returns nothing), so the admission outcome is deterministic given the stubbed probe. Confirmed at both extremes: green here on a torch+CUDA box, and green on CI's torchless Linux runner — I pulled the *Sidecar tests (pytest)* job log for this head (job `103181706015`) rather than trusting the check's green tick, and the pytest progress line reaches `[100%]` with no `F`, so the new test really is collected and really passes there. The stub-probe shape is also the one `test_load_admission.py` already uses at seven sites, not a second divergent harness.

**Regression sweep.** Scratch copy, full `tests/` with `-m "not golden"`: the only failure was `test_footprints.py::test_seed_parity_with_local_llm_doc`, a `FileNotFoundError` on `<repo-root>/docs/local-llm.md` — a copy artifact, since the scratch tree has no repo root. Rather than assert that from the error text (pass 3 hit the same two artifacts and had to reason around them), I placed the real `docs/local-llm.md` at the path the test resolves and re-ran: **11 passed**. Full suite then re-run end-to-end: **exit 0, fully green.** Targeted set `test_load_kokoro_arbiter_gate` + `test_design_kokoro_exclusion` + `test_load_admission` + `test_mint_variant_kokoro_stall` + `test_lifespan_order`: **36 passed.**

---

### B9 — comment half **RESOLVED and independently verified honest**; the ruling is **sound and I would have made the same call**; the PR body's copy is **still unfixed**

#### The comment is now accurate — I checked the claim that carries it, not just that the words changed

`main.py:1583-1603`. The false framing pass 3 flagged is gone: the 90s ceiling is no longer offered as a *bound on the wait*, it is explicitly narrowed to "only bounds what the CLIENT observes."

The new comment's load-bearing factual claim is that the dominant caller has **no** external timeout, and that is the part worth checking independently rather than accepting. It is **true, and true by deliberate design**: `server/src/tts/sidecar.ts:141-145` builds the synth dispatcher as

```ts
const SIDECAR_DISPATCHER = new Agent({
  headersTimeout: 0,
  bodyTimeout: 0,
  connectTimeout: 10_000,
});
```

with the comment above it stating outright that "the only valid cancellation for a synth is the caller's AbortSignal, never a wall-clock timeout" (plan 137, locked by `server/src/tts/sidecar-timeout.test.ts`). So `KokoroEngine.synthesize()`'s park has no wall-clock ceiling anywhere in the stack. The `/load` side of the claim also holds: `server/src/gpu/capacity-retry.ts:340` and `:360` both name "/api/sidecar/load's 90s" ceiling. And the "what stays held" sentence is right — `kokoro._loading` is set at `:11308` and cleared only in the `finally` at `:11327`.

Pass 3's B9(a) asked for the ruling to be **moved** to `_kokoro_ensure_loaded_guarded`. The fix agent instead kept it in `kokoro_synth()` and made it caller-aware. **That is a better resolution than the one requested**, and I am recording that deliberately: the defect was that a reader of the shared arbiter was told `/load` is the only thing that waits there. Enumerating both callers at the shared site fixes that; relocating the comment would have left `kokoro_synth()` with no note at all.

#### My independent opinion on the ruling: comment-only is correct, and a `cv.wait(timeout=…)` would have been the worse change

The dispatch asked me to form my own view rather than accept or silently re-open. I think the fix agent's call is right, and not merely defensible:

1. **A bounded wait has only two terminations and both are worse than waiting.** Proceed-anyway defeats the arbiter outright — it readmits the exact Kokoro/VoiceDesign GPU co-residency this whole PR exists to prevent (the plan-108 OOM class). Raise-on-timeout converts a slow-but-correct `synthesize()` into a *failed sentence mid-chapter-render*, which is a strictly worse operator outcome than a slow one; and on the `/load` side the client has already been told it failed at 90s, so raising buys the operator nothing.
2. **The exclusion is a safety property; fairness is not.** Trading a correctness guarantee for a fairness improvement, with no production report in hand, is the wrong direction of travel.
3. **The starvation precondition is narrow.** It needs a gapless stream of `design()` entries overlapping an active chapter render — i.e. bulk "Design full cast" running concurrently with generation. Those are separate operator modes in this app's own model; the overlap is constructible but is not the dominant shape.
4. **The deferral is legitimately specified rather than shrugged off.** The comment names the concrete trigger ("user reports starved TTS during bulk design") and the concrete remedy (a `timeout=` on the `cv.wait()`), which is the difference between an accepted limitation and an unexamined one.

So: **no new finding here, and I am not asking for the ruling to be re-opened.** If a future round does revisit it, the right shape is almost certainly a bounded wait on the **`/load` path only** (where failing loudly is already the client-visible outcome) and an unbounded one on `synthesize()` — but that is a design question, not a defect.

#### What is still open: the PR body's copy of the sentence B9 was filed against

**🔴 B9-residual — the "Known limitations, accepted" section of the PR body still carries the exact false claim verbatim**, in the PR body (not a file, so no `file:line`).

> *"**B5:** … Accepted as a known limitation bounded in practice by the Node-side 90s client timeout on `/load` requests. If a client times out, the sidecar continues waiting until active designs finish…"*

Pass 3 named this explicitly — *"The same wording is in the PR body's 'Known limitations, accepted' section, so both copies need it."* One copy moved; this one did not. It is false for the same reason and in the same two ways: the 90s ceiling is presented as bounding the wait rather than the client's observation of it, and the whole bullet is scoped to `/load` while `kokoro_synth()`'s dominant caller is `KokoroEngine.synthesize()` (`main.py:3825`), which has no ceiling at all.

Concrete failure scenario: this is the copy that survives into the merge commit and into issue archaeology on #3086. Someone triaging a future "Kokoro load hangs during cast design" report reads the PR body's accepted-limitations bullet, sees a 90s bound asserted, and rules the hang out of scope — while the in-file comment two clicks away says the opposite.

I am classifying this as a **correctness finding of the documentation class**, the same class passes 2 and 3 used for B6, B7 and B9 — but with one practical note the owner should weigh: **fixing it is a PR-body edit, which does not move the head SHA.** It invalidates none of the test evidence above and needs no re-push and no pass 5.

---

### 🔵 Cleanup nits

**N9 — the fix commit introduced three new hardcoded line numbers, and all three were false the moment they were written.** `tests/test_load_kokoro_arbiter_gate.py:149-150` ("line ~11306 … line ~11309") and `:181-185` ("the PRODUCTION-DEFAULT code path (~main.py:11306)", "if you revert line 11306's `_kokoro_ensure_loaded_guarded` call"). Those were the correct numbers at `1d880cbf`; this same commit expanded the B9 comment block by +14 lines at `main.py:1583`, so the sites are now **`:11320`** and **`:11323`**. Verified both ends: `git show 1d880cbf:server/tts-sidecar/main.py | grep -n` gives 11306/11309/10267, and the worktree gives 11320/11323/10281. The docstring is a **mutation recipe** — its stated purpose is to tell a future maintainer how to re-verify the test — and current `main.py:11306` is `if kokoro._kokoro is not None:`, an unrelated statement. Name the branch (`/load`'s `if _capacity_admission_enabled():` arm) instead of a line.

**N3 — restating pass 3's, now materially worse, and this commit is what made it worse.** `tests/test_load_kokoro_arbiter_gate.py:119` still reads *"removes ONLY the guard from the startup preload path (line 10260)"*. Pass 3 recorded that the real site had already moved to `:10267`; it is now `:10281`, and **`main.py:10260` is today `await asyncio.to_thread(coqui._ensure_loaded, coqui_model)` — the Coqui preload**. A maintainer following that instruction literally mutates a different engine's load path, watches the suite stay green, and concludes the test is broken. Under this repo's own stale-comment rule this was owed last round and is owed again; fold it into N9's fix.

**N10 — the new fixture drops the two "why" comments that keep `load_client` correct.** `load_client` carries the B3 rationale at `:47-49` and, critically, the B2 rationale at `:56-59` explaining that `_shares_device` must be pinned **after** `TestClient.__enter__` because the lifespan's `_configure_vd_kokoro_coupling` (`main.py:1024` → the assignment at `:1648`) overwrites it. `load_client_with_admission` (`:152-174`) does the same two things correctly but says why for neither. Concrete regression vector: someone tidying the duplication hoists the pin above the `with`, and on any box with `QWEN_DEVICE=cuda:1` (this repo's documented standing pin, per register row A106) the test goes red with a message that reads exactly like the fix regressing — pass 1's B2, re-created. Two comment lines, or a `# see load_client` pointer.

**N11 — the ruling's justification clause is a frequency claim where the real reason is termination.** `main.py:1600`: *"not a correctness bug since designs are expected to be infrequent."* The reason it is not a correctness bug is that each `design()` is a bounded GPU forward, so the wait terminates whenever the design stream has any gap at all; frequency only governs how *long* a park can get. This is a milder instance of the shape pass 3 filed as N8 against B4's "rare path". The conclusion is right either way.

**N4 / N5 / N6 / N7 / N8 — all five of pass 3's remaining nits are untouched.** This round's diff is two files (`main.py`, `tests/test_load_kokoro_arbiter_gate.py`), so none of them could have moved. N4 is now doubly stale: the Test plan still presents only pass 1's two-site mutation, and now also omits both `test_startup_preload_kokoro_uses_guarded_path` (which closed B1) and `test_load_kokoro_admission_on_blocks_while_design_active` (which closes B8) — the two tests that are actually the deliverable here. If the PR body is being edited for B9-residual anyway, N4 costs nothing extra.

---

### Verified clean this pass

- **Nothing else regressed.** `73ea30ef` touches exactly two files. `main.py`'s change is the comment block at `:1583-1603` — no statement moved, and `git diff` confirms the `while self._design_active_count > 0: self._cv.wait()` body is byte-identical. No test asserts on that comment.
- **Register and live view are untouched this round**, so pass 3's publish-token analysis stands unchanged and no re-stamp is owed. `origin/main` re-fetched this pass is still `2b63b451` — unmoved since pass 2 — so the three-way merge pass 3 verified is still the one that will happen. PR reports `MERGEABLE`.
- **All 22 checks green at this head**, including *Sidecar tests (pytest)*, *Check on-box acceptance register consistency*, and *Verify PR body links a GitHub issue*. `Closes #3086` is still present in the body and outside any code span.
- **Out of scope, correctly:** no `cast.json` write, no config knob, no derived artifact (`src/lib/api-types.ts` / `docs/BACKLOG.md` / brand PNGs). Release-notes pair landed in an earlier round and needs no delta for a test + comment change.
- **Commit hygiene:** author and committer are the repo owner; `Co-Authored-By: Claude Haiku 4.5` is within the routing table's Cheap tier for a well-specified fix with a clear repro and no design decisions, which is what B8 was.

---

### Verdict

**B8 is resolved — verified by my own mutation, with the complementary control measured.** The admission-ON guard was already in the shipped code (so no user-visible bug ever existed there), all three call sites are now individually pinned, and the new test is neither vacuous nor box-dependent: it passes on a torch+CUDA box and on CI's torchless runner, and its terminal `load_calls == ["cuda:0"]` assertion is what makes "the admission branch actually ran" load-bearing rather than assumed.

**B9's code-side half is resolved and its ruling is correct.** The comment is now honest about the dominant caller, I verified the claim it rests on against `sidecar.ts`'s deliberately-disabled timeouts rather than accepting it, and — asked for my own opinion — I would have made the same call. A bounded `cv.wait()` would have traded a safety property for a fairness one, and the deferral is specified with a real trigger and a real remedy rather than waved at. **I am not re-opening it.**

**One item stays open: the PR body's B5 bullet still asserts the sentence B9 was filed against.** That is the only correctness-class finding in this pass, it is a PR-body edit rather than a code change, and it does not move the head SHA — so it needs no further review round.

**Is the PR clean?** On code: **yes.** I found no correctness bug in `main.py`, in the test, or in their interaction, and the coverage gap that made pass 3 block is closed and proven closed. What remains is one false sentence in the PR body plus four docstring/comment nits, two of which (N9, N3) this commit either created or worsened and which this repo's own stale-comment rule says are owed in the same round. My recommendation to the owner: **edit the PR body (B9-residual, and N4 while there), fix the four line-number references in the test file's docstrings in one commit, and merge.** A pass 5 would have nothing left to measure.

Issue #3026, opened 2026-09-06
Generation queue permanently stuck behind a stale/unanswered awaiting_confirm entry


dudarenok-maker:
A stale/unanswered `awaiting_confirm` entry in the generation queue permanently blocks every entry queued behind it, with no error surfaced to the user — the chapter shows "Generating"/"Stalled" at 0/N lines, the Activity log gains zero `[tts]` activity, and the sidecar's own `/health` never shows `inflight_synth` for it. The FIFO queue processor appears unable to skip past a parked `awaiting_confirm` entry to reach later, real work.

## Where this came from

Found twice during the on-box mechanical batch 1 campaign (#2949, #2435):

- Step 1 (docs/testing/onbox-mechanical-batch1-results/step-1-a4-a14.md, "A14 — real-book QA/badge agreement"): clicking "Generate this chapter" on a cold, freshly-restarted server produced this exact symptom on a clean cast, blocking row A14 outright.
- Step 3 (docs/testing/onbox-mechanical-batch1-results/step-3-a8-a9-a10.md, "Finding 2"): traced a live reproduction of the same symptom to three leftover `awaiting_confirm` entries sitting in `castwright-workspace/.queue.json` from an earlier, unrelated session — a fourth, real "Generate this chapter" click queued behind them and could never be reached. Using the Queue modal's "Clear queue" (with "Also stop generation in progress") to drop all four, then re-clicking, dispatched immediately and rendered normally.

Step 3 could not confirm this was byte-for-byte step 1's own root cause (that session's queue state was gone by the time it was investigated), but it is a real, reproducible way to hit the identical symptom, and is worth ruling out first on any future "queue not dispatching" report before treating it as a dead pipeline.

## Suggested fix direction

Either the FIFO queue processor should be able to skip a parked/unanswered `awaiting_confirm` entry to reach later queued work, or a parked entry that never gets an answer should time out / surface a clear "queue is blocked on an unanswered confirmation" state to the user instead of a silent stall.

Filed per CLAUDE.md's incidental-findings rule, at the direction of #2958 (docs-only register fold) — reported, not fixed, since it needs a design decision (skip-vs-timeout) and isn't in scope for a docs-only fold.

dudarenok-maker:
## Exposure to this widens with PR #3031 — recording the interaction

[...]

Issue #3105, opened 2026-09-08
[agent instructions][claude][verify] Verify #3026 stuck-queue surfacing fix


dudarenok-maker:
[agent instructions][claude][verify] Verify #3026 stuck-queue surfacing fix

## What

Verify the fix for #3026 landed on branch `fix/frontend-3026-stuck-queue-confirm`
in worktree `C:\Claude\Projects\wt-3026-stuck-queue-confirm`, then open the PR.

## The decision (already made — do not re-litigate)

#3026's own body named an open decision: skip past a parked `awaiting_confirm`
entry, OR surface a clear blocked state to the user. **Decision: surface, do
NOT auto-skip.** Auto-skip risks silently reordering/abandoning a user's
queued work; a clear "queue is blocked — an earlier chapter is waiting on your
confirmation" message plus a pointer to the existing "Clear queue" control
(already in the Queue modal per the issue body) is lower-risk and keeps the
user in control. Reject any rework that proposes auto-skip instead — that
would need a NEW design pass, not a rework of this fix.

## Parent acceptance, restated as a checklist

- [ ] When one or more `queued` entries sit behind an `awaiting_confirm` entry
  that has NOT been answered for some bounded, reasonable duration, the Queue
  modal (or a toast) surfaces a clear message naming the blocked state — e.g.
  "N chapters are queued but blocked behind an unanswered fallback
  confirmation" — distinct from the ordinary "Queued"/"Generating" labels.
- [ ] The message points the user at the existing resolution: confirm or skip
  the blocking entry (the existing per-row controls), or "Clear queue".
- [ ] A regression test reproduces: enqueue an `awaiting_confirm` entry, enqueue
  a second `queued` entry behind it, advance time/ticks past whatever
  threshold the task child chose, and assert the blocked-state signal fires.
  FAILS pre-fix (no such signal exists today), PASSES post-fix.
- [ ] No auto-skip logic was added — reordering/dropping queued entries without
  user action is explicitly out of scope for this fix.

## FAIL ordering (exact)

1. Create a fresh `[cline][task]` rework child.
2. Create a fresh `[claude][verify]` child for the rework.
3. Link both as sub-issues of parent #3026.
4. Only then close this issue, with your verdict.

## On PASS

Open the PR:
- Base `main`, head `fix/frontend-3026-stuck-queue-confirm`.
- Title matches commit-convention format (`fix(frontend): ...`).
- `Closes #3026`.
[...]

Issue #3106, opened 2026-09-08
[agent instructions][claude][task] Surface a blocked-state signal when the queue is stuck behind an unanswered awaiting_confirm entry


dudarenok-maker:
[agent instructions][cline][task] Surface a clear blocked-state signal when the queue is stuck behind an unanswered awaiting_confirm entry

## What

Fix #3026: a stale/unanswered `awaiting_confirm` entry permanently blocks
every entry queued behind it, with the chapter showing "Generating"/"Stalled"
at 0/N lines and zero activity — no error surfaced to the user.

## The decision (already made — implement this, don't re-decide)

Two shapes were possible: auto-skip the parked entry, or surface a clear
blocked state. **Decision: surface, do not auto-skip.** Silently reordering or
abandoning a user's queued chapter is worse than telling them what's wrong and
pointing at the existing manual controls (per-row confirm/skip, and "Clear
queue" in the Queue modal — already exist per the parent issue body). Do not
implement a skip/timeout-that-drops-work mechanism.

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3026-stuck-queue-confirm`
Branch: `fix/frontend-3026-stuck-queue-confirm` (already checked out there)

Do NOT create another worktree. Do NOT work in the primary checkout. Commit on
this branch. Do NOT open a PR — the verify child does that.

## Background you need

`src/store/queue-dispatcher-middleware.ts`'s `tick()` STEP 2 fill loop (lines
219-281) only considers entries with `status === 'queued'` (line 228) — an
`awaiting_confirm` entry is deliberately skipped (it's not in that status), so
later `queued` entries SHOULD still be independently claimable up to the
worker limit... unless the entry actually queued behind it never reaches
`status:'queued'` for some other reason, or the modal/UI just doesn't
communicate that a chapter IS blocked specifically on confirmation versus
merely "next in line." Re-read the parent issue's two repro bullets carefully
— Step 3 traced a REAL case where a fourth genuine request was blocked behind
three leftover `awaiting_confirm` entries from an unrelated earlier session.
Confirm whether the blocking is (a) a real dispatcher gap (queued entries
genuinely never get claimed because of how many workers vs. how many parked
entries exist), or (b) purely a UI-legibility gap (the entries ARE
independently dispatchable, but nothing tells the user WHY the chapter they
care about looks stuck, because they don't know an unrelated earlier
`awaiting_confirm` entry exists at all).

[...]

PR #3143, opened 2026-09-10
fix(frontend): surface blocked-state signal for stale awaiting_confirm queue entries

dudarenok-maker:
## Summary

Surfaces a clear stale-signal (persistent Queue-modal banner + one-shot warn toast) when one or more `queued` entries sit elsewhere in the workspace queue while an `awaiting_confirm` entry has gone unanswered for 60s, naming the affected chapter's characters and pointing at the existing per-row confirm/skip controls or "Clear queue". This implements "surface, don't auto-skip" per the decision recorded in this verify child's brief (#3105) — no auto-skip/reorder logic was added; the dispatcher's `tick()` STEP 2 fill loop already continues past non-queued entries independently. Refs #3026.

## Test plan

- [x] Regression tests `src/modals/queue-modal.stale-awaiting.test.tsx` (8 cases: banner+toast fire past threshold, do not fire before threshold, do not fire when nothing else is queued in the workspace, clear on resolution, no repeated toast re-push on later polls, fires immediately on mount when already stale, plus 2 clock-skew cases pinning that a device clock reading behind the server doesn't permanently suppress or instantly-fire the signal) — fails pre-fix, passes post-fix.
- [x] Server regression tests `server/src/workspace/queue-io.test.ts` — `markAwaitingConfirm` stamps `parkedAt` with the current time (bounded between a `before` mark and `Date.now()`), and `fallbackCharacters` is carried through unchanged.
- [x] `parkedAt` added to `openapi.yaml`'s `QueueEntry` schema and `src/lib/api-types.ts` regenerated (`npm run openapi:types`); the hand-written widening for it in `src/store/queue-slice.ts` was removed now that the generated type carries the field.
- [x] Both release-notes entries (`docs/release-notes-next.md`, `RELEASE_NOTES.md`) updated to match the corrected signal semantics (see below) rather than the "blocked"/"queued behind it"/"updates in place" language the first round shipped.
- [x] Mutation-verified (this round): reverted the N2 clock-skew clamp in `src/modals/queue-modal.tsx` while keeping the new skew test — 1/8 failed with `TestingLibraryElementError` (banner never renders for the skewed-behind case) for exactly the reason the test exists to catch; restored the fix, reran — 8/8 passed again. Same pattern previously run against the S1–S5 mechanism in the prior round (pre-fix reversion → 2/3 failed with `TestingLibraryElementError`; restored → 3/3 passed).
- [x] `npx tsc --noEmit` (frontend + server) — clean, no errors.

Also fixed, found in passing (pr-review-gate pass 2, N1/N2/N3/N4/N5): the "queued behind it" gate never actually checked ordering, book, or pause state, so the copy was softened from "blocked" to "needs your input" everywhere (banner, toast, docstrings, both release-notes files) to match what the gate actually establishes; a browser-clock-behind-server skew could permanently suppress the stale signal, now clamped to fall back to client-observation time; `parkedAt` is now in `openapi.yaml` (was incorrectly documented as spec-absent, on a false precedent about `fallbackCharacters` that was also corrected); the "updates in place" toast description and the "one-shot per entry" docstring (true only per-mount, not across reloads) were corrected to describe actual behavior.


dudarenok-maker:
## PR review — pass 1 (head 2ec109a48f857609a44cc950dac6e3af97d7c017, depth medium)

Scope: both changed files read in full at head (`src/modals/queue-modal.tsx`, `src/modals/queue-modal.stale-awaiting.test.tsx`), plus the context they depend on — `src/store/queue-slice.ts` (entry shape, `selectQueueByBook` memoisation, `selectQueueCount`), `src/store/queue-dispatcher-middleware.ts` (STEP 1 park reconcile + STEP 2 fill loop), `src/store/notifications-slice.ts` (`pushToast` dedupe semantics), `src/components/toast-stack.tsx` (auto-dismiss), `src/store/generation-stream-runner.ts` (the pre-existing park toast), `src/components/layout.tsx` (mount site + cold-boot `loadQueue`), `openapi.yaml`'s `QueueEntry`, and the linked issue #3026 with both agent briefs #3105/#3106. No suites run — this pass is read-only and the primary checkout is on `main`; every finding below names the exact fixture edit or input that flips it.

### 🔴 Blocking — none

### 🟠 Significant

**S1 — [correctness] The signal fires with nothing queued behind the parked entry, and then tells the user their queue is blocked.**
`src/modals/queue-modal.tsx:82-91` and `:104-126` — the scan asks only `entry.status === 'awaiting_confirm'`. There is no predicate for a `queued` entry sitting behind it, even though #3105's acceptance ("When one or more `queued` entries sit behind an `awaiting_confirm` entry…") and this PR's own Summary both state the trigger that way.

Repro: a book whose only queued work is one chapter with an undesigned Qwen voice. Click "Generate this chapter" → the worker parks it → `.queue.json` holds exactly one entry, `awaiting_confirm`, nothing behind it. 60 s later — while the user is still reading the existing `Needs confirmation · no designed Qwen voice for Narrator → would render in Kokoro` row and deciding — they get a warn toast reading `Queue blocked: a chapter is waiting for voice confirmation (Narrator)…` (`:120`) plus a persistent amber banner reading `1 chapter blocked` (`:333-335`). Nothing is blocked. This is the loud-fallback gate working exactly as designed, and the copy reports it as a fault.

The new suite pins the wrong behaviour rather than catching it: `queue-modal.stale-awaiting.test.tsx:116` builds a store containing a single `awaiting_confirm` entry and `:122` asserts the banner **is** rendered.

**S2 — [correctness] The one test that carries the acceptance fixture does not depend on it.**
`src/modals/queue-modal.stale-awaiting.test.tsx:77-78` — test 1 is the only case with the `queued`-behind entry #3106's acceptance criterion 1 requires. Delete line 77 and change line 78 to `makeStore([awaiting])`: the test stays green, because nothing in `useStaleAwaitingConfirm` ever reads `status === 'queued'`. The fixture is decorative, so the criterion is uncovered by anything in the repo.

Red-before/green-after does not settle this: the test went red pre-fix because *no* signal existed at all, not because the queued-behind condition held. Same reason applies to the mutation receipt in the PR body — reverting the whole file proves the banner exists, not that it fires for the stated reason.

**S3 — [correctness] The load-bearing one-shot toast guard is untested, and deleting it is silently green.**
`src/modals/queue-modal.tsx:110-111` (`toastFiredRef`) is not a nicety. `pushToast`'s reducer dedupes on `dedupeKey` and **bumps `createdAt`** on a repeat (`src/store/notifications-slice.ts:78-83`), and `ToastItem`'s auto-dismiss effect keys on `[toast.id, toast.createdAt, dispatch]` (`src/components/toast-stack.tsx:41-48`). So a toast re-pushed every `STALE_CHECK_INTERVAL_MS` (5 s, `:63`) resets its own 6 s dismissal window forever — it becomes a permanent, undismissable toast.

Repro of the test gap: delete lines 110-111 and their closing brace so the dispatch runs on every tick. All three tests still pass. Test 1's only toast assertion is `toasts.find((x) => x.dedupeKey === 'stale-awaiting-confirm-e10')` (`:90-92`), which the reducer's dedupe keeps at exactly one element no matter how many pushes land. Nothing counts toasts, and nothing asserts `kind === 'warn'`.

**S4 — [correctness] "Unanswered for 60 s" is actually "present in this browser tab for 60 s", so every reload re-nags.**
`src/modals/queue-modal.tsx:73-74` — `firstSeenRef` and `toastFiredRef` are per component instance with no persistence, and `:86-88` stamps `firstSeen = now` on first observation.

Repro: a user deliberately leaves a parked entry alone (they intend to design the voice later). Every time they load the app, Layout's cold-boot `loadQueue` (`src/components/layout.tsx:469-473`) hydrates the entry, `firstSeen` is stamped fresh, and 60 s later the identical "Queue blocked" toast fires again — on every reload, indefinitely. The mirror of the same mechanism: an entry parked four hours ago in a previous session — i.e. exactly #3026's step-3 scenario, the case this PR exists for — has its signal withheld for a further 60 s after the app opens.

The comment at `:49-57` and the docstring at `:65-68` describe this as "unanswered past the threshold" and a "one-shot warn toast per entry". Neither is what the code measures. `addedAt` is not a drop-in substitute (it is enqueue time, not park time — a long-queued chapter that parked one second ago would fire instantly), so the honest fixes are a server-side park timestamp, or correcting those two comments to describe the actual mechanism.

**S5 — [correctness] Duplicate, strictly less informative toast in the live-park case.**
`src/store/generation-stream-runner.ts:523-546` already pushes a warn toast the instant the worker parks a chapter: `Chapter 3 paused — no designed Qwen voice for Narrator. Confirm or skip it in the generation queue.`, with `dedupeKey: fallback-confirm:<entryId>`. The new toast uses `dedupeKey: stale-awaiting-confirm-<id>` (`src/modals/queue-modal.tsx:121`), so the two do not collapse.

Repro: with the app open, park a chapter. At t the user gets an amber toast naming the chapter number; at t+60 s they get a second amber toast, `Queue blocked: a chapter is waiting for voice confirmation (Narrator)…`, which does *not* name the chapter. Two toasts for one event, the second less useful than the first. The new toast only adds information in the case the SSE toast cannot cover (a park from an earlier session), which argues for gating it on "this tab never saw the park event", or at minimum reusing `fallback-confirm:<entryId>` as the dedupeKey so it replaces rather than stacks.

**S6 — [house gate] Release-notes pair missing, with no not-applicable stated.**
`docs/release-notes-next.md` (marker `release-notes-next-version: 1.15.0`) and `RELEASE_NOTES.md` are both untouched and the PR body says nothing about them. This ships a user-visible banner and a user-visible toast, so CLAUDE.md before-shipping step 5 applies and the "no shippable delta" carve-out does not.

**S7 — [house gate] `Closes #3026` rests on a premise this PR asserts rather than demonstrates.**
#3106's brief specifically required determining whether the blocking is (a) a real dispatcher gap or (b) a UI-legibility gap, and fixing the dispatcher directly if (a). The PR's evidence for (b) is `src/modals/queue-modal.tsx:52-57`, which restates the brief's own hypothesis.

I checked the two places that decide it, and they do support (b) for the frontend dispatcher: `src/store/queue-dispatcher-middleware.ts:228` is `if (e.status !== 'queued') continue;`, and `:122-142` shows a park does `inFlight.delete(entryId)` before `continue`, so it frees its worker slot rather than holding it. That is genuinely reassuring — but it leaves #3026's step-3 field observation ("a fourth, real 'Generate this chapter' click queued behind them and could never be reached") unexplained, and merging this auto-closes the issue.

One mechanism that fits the reported symptom exactly and is not addressed here: the 4th entry is claimed normally and then immediately re-parked server-side by the same loud-fallback gate (same book, same undesigned voices), producing "Generating at 0/N lines, zero `[tts]` activity" with no FIFO defect anywhere. Under that mechanism this banner does help — but #3026's own "the FIFO queue processor appears unable to skip past a parked entry" framing is then wrong, and closing the issue silently inherits it. Either land the repro/evidence, or switch to `Refs #3026` and leave the issue open with the residual question named.

### 🟡 Minor

**M1** — `src/modals/queue-modal.tsx:103,128`: `setStaleEntries(stale)` allocates a fresh array every check, so React never bails via `Object.is`. `QueueModal` re-renders every 5 s for the entire app session even with zero `awaiting_confirm` entries and the modal closed — it is mounted unconditionally at `src/components/layout.tsx:2336`. Compare ids and skip the `setState` when the set is unchanged.

**M2** — `src/modals/queue-modal.stale-awaiting.test.tsx:133`: no trailing newline (`\ No newline at end of file`). CI will not catch it (`npm run lint` is ESLint only; `format:check` is not wired into `verify.yml`), but nothing else in the tree is shaped that way.

**M3** — `src/modals/queue-modal.stale-awaiting.test.tsx:20-21`: two separate `import … from './queue-modal'` statements; fold into one.

**M4** — `src/modals/queue-modal.stale-awaiting.test.tsx:124-128`: `store.dispatch(...)` outside `act()`. The following `rerender` makes it work today, but it is a React-18 act warning waiting to be triggered by a scheduling change.

**M5** — `src/modals/queue-modal.tsx:339-345`: the banner's character list is `flatMap`ped across all stale entries and joined without de-duplication, so three chapters parked on the same Narrator render `(Narrator, Narrator, Narrator)`.

**M6** — Neither `docs/features/archive/102-global-queue-modal.md` (which owns the modal's "Invariants to preserve" / "Automated coverage" sections and is cited from `src/store/queue-slice.ts:49`) nor an e2e spec was touched. The e2e is a SHOULD in CLAUDE.md and this change crosses the redux + layout seams; updating an archived/stable plan is a judgement call. Either is fine to skip — but say so in the PR body rather than silently.

### ✅ What is solid

- **The hook is mounted where it can actually fire.** `QueueModalContainer` is unconditional at `src/components/layout.tsx:2336`, `if (!open) return null` sits *after* the hook call (`queue-modal.tsx:182` vs `:206`), and Layout cold-boot-hydrates the queue at `:469-473`. So the toast is not gated on the modal being open — the "fix unreachable at the default configuration" shape was checked and is not present.
- **The banner's enclosing `count > 0` guard is satisfiable.** `selectQueueCount` is a raw `entries.length` (`src/store/queue-slice.ts:104`) and counts `awaiting_confirm`, so an all-parked queue still reaches the banner branch rather than the empty state.
- **No infinite render loop.** `selectQueueByBook` is a `createSelector` memoised on the entries reference (`src/store/queue-slice.ts:118-132`), so the `setStaleEntries` re-render does not re-run the effect that produced it.
- **Eviction is correct.** Both refs are cleared when an entry leaves `awaiting_confirm` (`queue-modal.tsx:95-100`), so a confirm-then-re-park cycle re-arms rather than staying permanently silent.
- **Types and imports check out.** `kind: 'warn'` is a real `ToastKind` (`src/store/notifications-slice.ts:19`), `IconWarning` exists (`src/lib/icons.tsx:54`), and all eight `queue-thunks` exports the modal imports are covered by the test's `vi.mock` factory.
- **No auto-skip/reorder logic was added**, per the brief's explicit constraint. Confirmed by reading the whole diff.
- **Threshold arithmetic in the tests is right.** With a 5 s interval, test 2's 59 s advance lands its last check at 55 s (< 60 s, correctly silent) and test 1's 66 s advance crosses at exactly 60 s.

### Verdict

**Request changes.** Five correctness findings. S1 is the one that changes shipped behaviour for an ordinary user: the "Queue blocked" / "N chapters blocked" copy fires in the single-parked-chapter case where nothing is blocked, and the new suite pins that as expected. S2 and S3 are the test-cannot-fail pair — the acceptance criterion's queued-behind fixture is decorative, and the one-shot guard that keeps the toast dismissable can be deleted with all three tests still green. S4 and S5 are user-facing nag defects (re-fires on every reload; duplicates an existing, better toast). S6 and S7 are house gates: the release-notes pair, and `Closes #3026` on a root cause that was never reproduced.

dudarenok-maker:
## PR review — pass 2 (head 69c8e3f2, depth medium)

Deltas only — pass 1's "what is solid" is not restated. This pass ran the suite: 6/6 green at head, then five targeted mutations and six behavioural probes in a throwaway detached worktree at `69c8e3f2` (removed afterwards; no tracked file touched). Each disposition below is re-derived, not taken from the fix report.

- **S1** (signal fires with nothing queued behind it) — **RESOLVED as reported.** `src/modals/queue-modal.tsx:101` gates on a `queued` entry existing, and `queue-modal.stale-awaiting.test.tsx:225` pins the lone-parked-entry negative. Mutation-verified: replacing the gate with `true` fails exactly that test and nothing else. The gate admits more than the "queued **behind** it" it claims — see **N1**.
- **S2** (acceptance fixture decorative) — **VERIFIED RESOLVED.** Mutation-verified: flipping test 1's sibling `e11` from `status: 'queued'` to `'failed'` turns test 1 red. The fixture is now load-bearing.
- **S3** (untested `toastFiredRef`) — **VERIFIED RESOLVED, red for the claimed reason.** Mutation-verified: replacing `if (!toastFiredRef.current.has(entry.id))` with `if (true)` fails only the new poll test, with `AssertionError: expected 1788998485000 to be 1788998465000` — a 20 000 ms `createdAt` bump, i.e. exactly the four 5 s re-pushes the guard exists to stop.
- **S4** (staleness per mount, not per entry) — **RESOLVED for the substantive half.** `parkedAt` is stamped at the single `awaiting_confirm` assignment site in the tree (`server/src/workspace/queue-io.ts:317`), survives `writeQueueFile`/`migrateQueueJson` (both pass-through spreads, no field whitelist), and reaches the client raw via `res.json({ entries: file.entries, … })` (`server/src/routes/queue.ts:61`) into `setSnapshot`'s direct assignment — so the field is live end-to-end, not just in the store fixture. Mutation-verified: narrowing `entry.parkedAt ?? entry.addedAt` to `entry.addedAt` fails the already-stale-on-mount test. The `?? addedAt` fallback is sound: `markAwaitingConfirm` is the only writer of `awaiting_confirm`, so the fallback is reachable only for an entry parked by a pre-#3106 server and still parked after the upgrade restart — for which `addedAt <= parkedAt < restart`, so it over-reports staleness on an entry already guaranteed stale. Two residuals: **N2** (new cross-clock comparison) and **N3** (the reload half is unchanged and the docstring still denies it).
- **S5** (duplicate toast stacking) — **RESOLVED.** `src/modals/queue-modal.tsx:133` now uses `fallback-confirm:${entry.id}`, matching `generation-stream-runner.ts:546`'s `fallback-confirm:${ev.queueEntryId ?? ev.chapterId}`; the dispatcher always supplies `queueEntryId` on a queue-driven claim (`queue-dispatcher-middleware.ts:265`), so the keys collapse on every path that can produce both. The *mechanism* described in the release notes is wrong — see **N4**.
- **S6** (release-notes pair) — **VERIFIED RESOLVED.** Both files carry an entry, and the `69c8e3f2` conflict resolution is clean: diffing the merge against both parents shows main's four new lines and the branch's one line all present, nothing dropped or corrupted.
- **S7** (`Closes #3026` on an unreproduced root cause) — **STILL OPEN, untouched.** The body still says `Closes #3026`, no repro or evidence was landed, and it was not switched to `Refs`. Merging still auto-closes #3026 on the strength of a hypothesis.
- **M1–M6** — M2 fixed (trailing newline present). **M1, M3, M4, M5, M6 all unchanged and still open**; none blocks.

### 🔴 Blocking — none

### 🟠 Significant

**N1 — [correctness] The S1 gate is "any `queued` entry anywhere", not "queued behind this entry", and over-fires three demonstrable ways.**
`src/modals/queue-modal.tsx:101` — `allEntries.some((e) => e.status === 'queued')` is evaluated over `groupedByBook.flatMap(...)`, i.e. every entry in every book, with no reference to the parked entry's `order` or `bookId`. All three of the following were **run** against `69c8e3f2` and render the banner:

1. **Cross-book.** Book A's only entry is the parked chapter; book B has one `queued` chapter. Book A's parked entry blocks nothing — the dispatcher drains book B fine — yet the modal shows `1 chapter blocked` and the toast says `Queue blocked`. This is S1's exact user-facing wrongness, reachable as soon as a second book has work.
2. **Ordered ahead, not behind.** `queued` at `order: 0`, `awaiting_confirm` at `order: 1` → fires, with literally nothing behind it. Reachable through the modal's own drag-reorder: `queue-io.ts:150` treats every non-`in_progress` entry as reorderable, so an `awaiting_confirm` entry can be dragged below a queued one.
3. **Paused queue.** With `queue.paused === true` the signal fires and the copy says `Open the Queue to confirm or skip` — but confirming flips the entry to `queued` and the dispatcher returns at `queue-dispatcher-middleware.ts:217` without claiming it. The advice cannot resolve what the banner claims is blocking.

The stronger claim is asserted in three places the code does not implement: the docstring at `src/modals/queue-modal.tsx:71-73` ("only fires when something is actually queued behind the stale entry"), `docs/release-notes-next.md` ("only fires when something is actually queued behind the parked entry"), and `RELEASE_NOTES.md` ("It only speaks up when something's actually being held back"). Worth noting which way to fix it: STEP 2 is `if (e.status !== 'queued') continue;` (`queue-dispatcher-middleware.ts:228`), so a parked entry never blocks anything under any ordering — there is no predicate that makes "blocked" true. The defensible fix is the copy (say "waiting for your answer", not "blocked"), not a stricter gate.

**N2 — [correctness] S4's fix introduces a cross-clock subtraction that defeats the 60 s threshold on a skewed device.**
`src/modals/queue-modal.tsx:113` computes `now - Date.parse(entry.parkedAt)` where `now` is the **browser's** `Date.now()` and `parkedAt` is stamped from the **server's** `new Date()` (`server/src/workspace/queue-io.ts:317`). The pre-fix mechanism was wholly client-side and therefore skew-immune; this one is not. Both directions were run:

- **Browser clock ahead of the server by 2 min** → the banner renders on the very first `check()`, with zero elapsed time, i.e. the instant the chapter parks. That is precisely the "don't nag a user who is actively deciding" case the 60 s constant exists for (`:49-57`), and precisely the scenario S1 objected to.
- **Browser clock behind the server by 2 min** → the signal is suppressed past the threshold entirely (still no banner at t+66 s); the feature silently does nothing.

CLAUDE.md's mobile-testing protocol has this app driven from phones and tablets over LAN HTTPS, where a device whose clock is minutes off the server box is ordinary — a tablet that has been powered off, or one that never reached NTP. Fixes: carry a server `now` alongside the queue snapshot and difference against that, or fall back to the observation-time clock whenever `parkedAt` is in the browser's future. Either needs a paired test — the current suite cannot see this, because it builds `parkedAt` from the same fake clock the component reads.

**N3 — [correctness] `parkedAt` is a wire field the frontend reads, but it was kept out of `openapi.yaml` on a false precedent.**
`server/src/workspace/queue-io.ts:75-82` justifies the omission with "SERVER queue shape only (NOT in openapi.yaml's QueueEntry — like fallbackCharacters/requiredEngines above)". `fallbackCharacters` **is** in `openapi.yaml` — line 5840, with `fallbackConfirmed` at 5853. Only `requiredEngines`/`multiTts` are genuinely spec-absent. The pre-existing comment at `queue-io.ts:67-68` makes the same false claim about `fallbackCharacters`, which is where the new comment got it; that stale comment is a chore this work surfaced and should be fixed in the same round rather than propagated.

`fallbackCharacters` is the exact analogue of `parkedAt` — stamped by the same `markAwaitingConfirm` call, present only on a parked entry, read by the frontend off `GET /api/queue` — and it is in the spec. Concrete failure: CLAUDE.md states this frontend is "designed to swap to a real backend without changing component code" and that `openapi.yaml` is the "source of truth for backend shapes". A backend implemented from the spec emits no `parkedAt`, so `entry.parkedAt ?? entry.addedAt` silently falls through to `addedAt` for every parked entry — reinstating by omission the exact substitution pass 1 rejected as unsound, with no test anywhere able to notice. Add the field to `openapi.yaml`, regenerate `src/lib/api-types.ts`, and drop the hand-written widening at `src/store/queue-slice.ts:31-35`.

### 🟡 Minor

**N4 — [doc]** `docs/release-notes-next.md` and `RELEASE_NOTES.md` both say the stale toast "updates in place" / "updates the original notification instead of piling up a second one", and `queue-modal.tsx:130-132` says it "fires as an update (message + createdAt bump) on that existing toast". That is not what happens in the running app: `ToastStack` auto-dismisses the park toast after 6 s and the threshold is 60 s, so the existing toast is always gone by then and the shared key produces a **fresh** toast, not an update. The outcome (never two at once) is right; the stated mechanism is wrong in all three places, and it is only "in place" in the test, where no `ToastStack` is mounted to dismiss anything.

**N5 — [doc]** `src/modals/queue-modal.tsx:66` — "Fires a one-shot warn toast per entry" is still false across page loads. `toastFiredRef` is per component instance, so a reload re-arms it; verified by unmounting and re-rendering against the same store snapshot, which re-pushes the toast with a new `createdAt`. With `parkedAt` now already past the threshold, the re-fire is also immediate rather than 60 s in. This is S4's "every reload re-nags" half, unchanged — arguably acceptable product behaviour, but the docstring asserts the opposite.

**N6 — [house gate]** The PR body was not updated for this round. Its Test plan still describes "3 cases" (there are 6) and still carries the pre-fix mutation receipt ("2/3 tests failed"); it never mentions the new server-side `parkedAt` field, `server/src/workspace/queue-io.ts`/`queue-io.test.ts` changing at all, or the release-notes pair. CLAUDE.md requires fixes found in passing to be declared in the body — a reviewer reading it would not know the server was touched. Separately, `docs/release-notes-next.md` cites "(#3106/#3105)" — the two agent-child briefs — and never #3026, the bug the PR closes.

### Verdict

**Request changes.** Pass 1's S2, S3, S5 and S6 are verified fixed, S4's substantive half is fixed and its plumbing verified end-to-end, and S1 is fixed for the case as reported — all mutation- or probe-verified rather than read. Three findings stand: N1 (the new gate is a weaker predicate than the code, both release-notes files, and the tests all claim, with three run-and-confirmed over-fires), N2 (the S4 fix replaced a skew-immune client measurement with a browser-vs-server subtraction that fires instantly on a 2-minute-skewed device — a regression introduced by this round), and N3 (a wire field kept out of `openapi.yaml` on a comment that mis-cites `fallbackCharacters`, which is in the spec). S7 remains untouched from pass 1: `Closes #3026` still rests on a hypothesis rather than a repro.

dudarenok-maker:
## PR review — pass 3 (head 98d2ab36, depth medium)

Deltas only — pass 1's "what is solid" and pass 2's verified-resolved items are not restated. Depth re-derived from the current commit set: three `fix(frontend)` commits, single scope → `medium`.

This pass ran in a throwaway detached worktree at `98d2ab36` (junctions dropped, worktree removed afterwards; no tracked file touched, primary checkout `git status --porcelain` byte-identical before and after). Verified: 8/8 green on `src/modals/queue-modal.stale-awaiting.test.tsx`; 113/113 on all frontend queue specs (`src/modals/queue-modal*`, `src/store/queue*`); 51/51 on `server/src/workspace/queue-io.test.ts`; `npx tsc --noEmit` (frontend) clean. Then one mutation and three behavioural probes. Every disposition below is re-derived, not taken from the fix report.

- **N1** (gate is "any `queued` entry anywhere", not "queued behind it") — **RESOLVED for the copy, which is the fix pass 2 recommended.** No user-facing string anywhere on the branch still says "blocked": `git grep -E 'chapters? blocked|Queue blocked'` at head returns nothing. Banner is `1 chapter needs your input` / `N chapters need your input` (`src/modals/queue-modal.tsx:398-400`), toast is `Still waiting: a chapter needs voice confirmation (…)` (`:177`), and both release-notes files now state the gate honestly ("it isn't a claim that the parked entry itself is blocking that other work"). Re-checked the copy against all three over-fire cases pass 2 ran — cross-book, ordered-ahead, paused-queue — and it stays true in each. Residual "blocked" language survives in comments and test names only — see **P4**, and a new copy/gate mismatch in the release notes at **P3**.
- **N2** (cross-clock subtraction) — **RESOLVED in the suppression direction, verified by mutation; the instant-fire direction is unfixed and undeclared, and the clamp introduces a new defect.** Mutation: replacing `src/modals/queue-modal.tsx:150-162` with the pass-2 body (`const stampMs = Date.parse(...); if (Number.isNaN(stampMs) || now - stampMs < STALE_AWAITING_CONFIRM_MS) continue;`) turns exactly one test red — `clock skew (N2) > does not permanently suppress the signal when the device clock reads BEHIND the server`, with `TestingLibraryElementError: Unable to find an element by: [data-testid="queue-stale-awaiting-banner"]`, i.e. red for precisely the reason claimed. The test does **not** share the component's clock in the way that matters: `parkedAt` is built as `Date.parse(NOW_ISO) + 2 * 60_000` (`:245`) and never re-derived from the advancing fake clock, so the gap is a real fixed skew. That half is genuinely closed and genuinely covered. See **P1** (new) and **P5** (the accepted half).
- **N3** (`parkedAt` kept out of `openapi.yaml` on a false precedent) — **VERIFIED RESOLVED, including the derived artifact.** `parkedAt` is at `openapi.yaml:5859` inside `QueueEntry`'s `properties`, correctly left out of `required:` (line 5795 is unchanged: `[id, bookId, chapterId, scope, addedAt, status, order]`). `src/lib/api-types.ts` was **actually regenerated, not hand-edited** — I ran `openapi-typescript` against head's own `openapi.yaml` into a temp file and diffed it against head's `src/lib/api-types.ts`: **zero bytes differ**. The hand-written widening is genuinely gone, not left redundant: `src/store/queue-slice.ts:30-33` now widens only `requiredEngines`/`multiTts`, and `git grep parkedAt -- src/store/` returns only the explanatory comment. Both stale comments in `queue-io.ts` are corrected (`:64-68`, `:74-83`). Nothing else consumes the schema in a way the additive optional property can break — frontend `tsc --noEmit` is clean and all 113 queue specs pass.
- **N4** ("updates in place") — **VERIFIED RESOLVED.** All three sites now say "replaces" rather than "updates in place": `queue-modal.tsx:110-116` and `:178-180`, `docs/release-notes-next.md` ("replaces any lingering park-time notification"), `RELEASE_NOTES.md` ("a fresh reminder replaces the original notification").
- **N5** ("one-shot per entry") — **VERIFIED RESOLVED.** `queue-modal.tsx:66-73` now says "per entry, per component mount" and spells out that a reload re-arms it. Accurate.
- **N6** (stale PR body) — **RESOLVED.** Body now describes 8 cases (the file has 8), carries this round's mutation receipt (which I reproduced exactly: 1/8 red, `TestingLibraryElementError`), names the server-side `parkedAt` work and both release-notes files, and carries an "Also fixed, found in passing" paragraph. Two residues: it does not declare N2's accepted residual (**P5**), and `docs/release-notes-next.md` still cites `(#3106/#3105)` — the two agent-child briefs — rather than #3026 or this PR, in a register CLAUDE.md describes as PR-refed.
- **S7** (`Closes #3026` on an unreproduced root cause) — **VERIFIED RESOLVED, and checked at the level that actually decides it.** The body says `Refs #3026`. I also checked every commit message on the branch for an auto-close trailer, since a commit trailer beats the PR body on GitHub: `git log 733ee305..98d2ab36` contains exactly one issue reference, `Refs #3026` in `98d2ab36`, and no `Closes`/`Fixes`/`Resolves` anywhere. #3026 is confirmed still `OPEN` and will stay open on merge.
- **M1, M3, M4, M5, M6** (pass 1) — all still open, none blocking, none re-stated here. One update: **M4 is no longer theoretical** — the run at head prints `An update to QueueModal inside a test was not wrapped in act(...)` twice to stderr from the `clears banner when the entry is resolved` case.

### 🔴 Blocking — none

### 🟠 Significant

**P1 — [correctness] The N2 clamp discards its own observation the moment the skew stops reading as future, so the banner appears, vanishes for a full minute, then reappears.**
`src/modals/queue-modal.tsx:159-161` — the `else { staleObservedAtRef.current.delete(entry.id); }` branch throws away the client-clock fallback and reverts `stampMs` to the server stamp, re-importing exactly the skew the clamp exists to neutralise.

Reproduced (throwaway probe at head, skew 120 s behind: `parkedAt = NOW_ISO + 120s`, one sibling `queued` entry, one `check()` tick per sample):

```
t=65s banner=true | t=125s banner=false | t=185s banner=true | t=245s banner=true
```

Derivation, with the client behind by `S` and the park at server time `P`: the client's `now` at elapsed `t` is `P + t − S`. While `t < S` the condition `serverStampMs > now` holds, the clamp is used, elapsed reads `t`, and the banner correctly fires at `t = 60 s`. At `t ≥ S` the `else` branch fires: the observation is deleted, `stampMs` reverts to `P`, and elapsed collapses to `t − S ≈ 0` — so the entry drops out of `stale`, `setStaleEntries` (`:188`) unmounts the banner, and it stays gone until `t = S + 60 s`. Only bites when `S > 60 s`; below that the clamp never reaches the threshold first anyway, so the observable effect is just a delay to `60 + S`. `toastFiredRef` is not evicted during the gap, so there is no second toast — the defect is banner-only.

The paired test cannot see this, and the reason is structural rather than incidental: both skew cases stop at `STALE_AWAITING_CONFIRM_MS + 6_000` (`:262`) — i.e. `t = 66 s`, well before the crossover at `t = S = 120 s`. Advancing that one test to `S + 6_000` and asserting the banner is *still* present would cover it.

Fix is one line — never let a later reading move the effective park time forward, e.g. keep the observation once recorded and use `stampMs = Math.min(serverStampMs, staleObservedAtRef.current.get(entry.id) ?? serverStampMs)`.

Proportionality, stated plainly so it can be weighed rather than inferred: this is confined to the clock-skew path, it is strictly better than the pre-round behaviour (which showed nothing at all in this scenario), and it never displays *wrong* information — it withholds correct information for one interval. It is nonetheless a wrong output for a specified input (at `t = 125 s` the entry has been unanswered for 125 s and the hook reports it as not-stale), and it was introduced by this round.

### 🟡 Minor

**P2 — [doc] `RELEASE_NOTES.md` promises the banner names the chapter; it names only characters — and the two release-notes files contradict each other on this.**
`RELEASE_NOTES.md` (this release's new bullet): "the Queue now shows a clear "needs your input" banner **naming the chapter and characters**". Probed the rendered banner at head with a parked `chapterId: 7`:

```
"1 chapter needs your inputWaiting for voice confirmation (Narrator). Confirm or skip the row below, or use "Clear queue" above."
```

No chapter number, no title — and the toast is the same (`Still waiting: a chapter needs voice confirmation (Narrator)…`). With two chapters parked on the same character the user is told which character, never which chapter. `docs/release-notes-next.md` describes the same banner correctly ("naming the characters"), so the pair disagree. Note this is a real downgrade against the park-time toast the shared dedupe key now replaces, which *does* say `Chapter 3 paused` (`generation-stream-runner.ts:546`) — so the user-visible net effect of the S5 fix is losing the chapter number. Either fix the sentence or put the chapter in the copy.

**P3 — [doc] Both release-notes bullets open by describing the lone-parked-entry symptom, which is the one case the shipped gate excludes.**
`docs/release-notes-next.md`: "a lone parked entry with nothing else queued could sit there indefinitely looking like it had simply stopped". `RELEASE_NOTES.md`: "a parked chapter with nothing else queued could sit there looking stalled instead of waiting". Both then state the gate correctly two sentences later ("whenever there's other queued work waiting"), so each bullet self-corrects — but each leads with a symptom the change does not address, and `queue-modal.stale-awaiting.test.tsx:140-154` pins that exclusion as intended. Concretely: a user who reads either note, parks a single chapter with an empty queue and waits five minutes gets no banner and no toast. Worth noting the underlying tension rather than only the wording — after N1 the banner says "needs your input" and never mentions other queued work, so `hasQueuedEntry` (`:134`) no longer has any relationship to the message it gates; whether that gate should still exist is a product call, not a review finding, but the notes should not describe the excluded case as the problem being solved.

**P4 — [doc] Residual "blocked" / "queued behind it" language, now self-contradicting.**
`src/modals/queue-modal.tsx:49-57` — the comment on `STALE_AWAITING_CONFIRM_MS` still calls this a "blocked-state signal", says it "surfaces the block", and says it tells the user "WHY a chapter looks stuck". Seven lines further down, `:80-85` states the opposite as an established fact: "a parked entry never technically blocks anything under any ordering/book/pause combination". Two adjacent comments on the same mechanism now assert contradictory things, and the one N1 corrected is the second. Same residue in the test file: the describe block is `#3106 stale awaiting_confirm blocked-state signal` (`:89`) and two test names still read "with something queued behind it" (`:96`, `:119`) — the exact predicate N1 established the gate does not implement, so a future reader mutating the gate will read those names as the contract. The PR title also still reads "surface **blocked-state** signal", which becomes the merge commit subject.

**P5 — [house gate] N2's unfixed half is real and user-visible, and is recorded in neither an issue nor the PR body.**
The client-clock-**ahead** direction is unchanged and I reproduced it at head: with the client 120 s ahead of the server (`parkedAt = NOW_ISO − 120s`) and the entry *just* parked, `bannerOnMount = true, toastOnMount = true` — the banner and the warn toast both fire on the very first `check()`, defeating the entire 60 s grace period that `:49-57` exists to provide. `queue-modal.tsx:105-109` documents this as accepted and names the fix ("needs the server's own 'now' alongside the snapshot"), which is a defensible design-pass call — adding a field to `GET /api/queue`'s response is a spec decision. But CLAUDE.md's design-pass carve-out is the *only* deferral it allows and it requires the issue plus the decision owed; `gh issue list` finds no issue for it, and the PR body's "Also fixed, found in passing" paragraph describes N2 as simply fixed ("now clamped to fall back to client-observation time") with no mention of a residual. A reviewer reading the body would not know half of N2 is still open. File the issue naming the decision (carry a server `now` in the queue snapshot vs. track the in-session `in_progress → awaiting_confirm` transition client-side) and say so in the body.

### Verdict

**Request changes — and this is the loop cap.** Per `pr-review-gate` SKILL.md's "Re-review trigger and loop cap" (initial pass + 2 re-review rounds, 3 total), pass 3 is the last automatic round. **Do not open a pass 4** — this hands to the repo owner to decide.

Substance, so that decision has something to weigh. All six pass-2 findings are genuinely addressed and I verified each by re-derivation rather than by report: N3 is the strongest of them (`api-types.ts` is byte-identical to a fresh regeneration, the widening is gone, the false precedent is corrected in both comments), N2's suppression half is mutation-verified with a test that genuinely fails without the clamp and for the stated reason, and S7 is closed at the commit-trailer level, not just the body.

One correctness finding stands (**P1**), and it is new this round rather than a re-trip of N2: the clamp reverts to the skewed server stamp at `t = S` and drops the banner for a minute. It is a one-line fix, its test gap is a single advanced timer in an existing case, and it is confined to a path that was strictly broken before this round. Four minors (**P2–P5**) are documentation and bookkeeping — P2 and P3 are wrong statements in shipped release notes, P5 is an undeclared, un-ticketed deferral of half a pass-2 finding.

My read: nothing here is a merge-blocker on shipped behaviour for a correctly-clocked device, and P1 strictly improves on the pre-round state. A fourth fix round for P1 + P2 + P5 would be cheap and would close the branch properly; merging as-is ships a skew-path banner flicker and two release-notes sentences that are false. That is the owner's call, not mine.

Issue #1284, opened 2026-07-04
Queue dispatcher stalls after confirmFallbackEntry — chapter never claimed for synthesis


dudarenok-maker:
## What happened

Generating audio for a chapter whose speaker has no designed Qwen voice trips a per-chapter safety gate: the queue entry parks in `awaiting_confirm` ("Chapter N paused — no designed Qwen voice for X. Confirm or skip it in the generation queue."), resolved via the Queue modal's "Render anyway" / "Skip" buttons (`src/modals/queue-modal.tsx` — `onConfirmFallback`/`onSkipFallback`, `confirmFallbackEntry`/`skipFallbackEntry`).

After confirming the fallback, the chapter's queue entry transitions to `queued` but is **never claimed by the dispatcher** — generation stays at "line 0 of N" indefinitely.

## Reproduction (3/3)

Book: *The Coalfall Commission* demo book, Chapter 4 ("The Pour", 199 lines, 14 speakers), a chapter needing a Qwen voice for "Oduvan".

1. Start generation with Chapter 4 in the queue.
2. Chapter 4 pauses at `awaiting_confirm` ("no designed Qwen voice for Oduvan").
3. Confirm the fallback via the Queue modal ("Render anyway").
4. Chapter 4 transitions to `queued` but never progresses past "line 0 of 199" — reproduced 3 separate times, including once with the queue entry deleted and re-created from scratch.

## Evidence collected

- `curl :8080/api/sidecar/health` and the sidecar's own `/health` showed `qwenLoaded: false`, `vramReservedMb: ~94` (essentially idle) — the GPU was doing nothing.
- `nvidia-smi` showed 0% utilization throughout.
- `server.log` had **zero** log lines for a Chapter 4 synthesis attempt despite three real dispatch attempts spanning ~25 minutes; the sidecar's own log showed only routine `/health` polls, never a `/synthesize` call.
- Toggling queue pause/resume did not help.

Meanwhile Chapter 3 (same book, same session, no fallback-confirm needed) rendered to completion cleanly and quickly.

## Suspected area

The `awaiting_confirm` → `confirmFallbackEntry` → `queued` → (dispatcher claim) transition in the queue/generation dispatcher — looks like the confirmed entry isn't being picked back up by whatever polls/claims `queued` entries for synthesis.

## Key files

- `src/modals/queue-modal.tsx` (`onConfirmFallback`, `confirmFallbackEntry`)
[...]

PR #3144, opened 2026-09-10
fix(frontend): close stream on park so re-claim isn't blocked

dudarenok-maker:
## Summary

A freshly `fallbackConfirmed: true` queue entry could sit at `queued` for minutes: the stream handle stayed open until the server's `idle` tick arrived, so `hasOpenStreamForChapter` still returned true for the just-parked chapter and the dispatcher's STEP 2 skipped re-claiming it until an unrelated action fired the dispatcher. Closes the stream handle synchronously on the `chapter_awaiting_fallback_confirm` park event instead of waiting for the subsequent `idle` tick. Refs #3029.

## Test plan

- [x] `npx vitest run src/store/queue-dispatcher-middleware.test.ts` — 26/26 passed, including the new regression test `re-claims a parked chapter after confirm even when idle has not yet arrived (#3029)`, which parks a chapter (no idle tick) and then confirms it, asserting a second stream opens for the same chapter.
- [x] Mutation test (verify child, re-run independently of the task child's own claim): reverted `src/store/generation-stream-runner.ts` to its pre-fix state via `git checkout cf7f2829~1 -- src/store/generation-stream-runner.ts`, re-ran the scoped test — RED, with the expected assertion `AssertionError: expected [ [ 1 ] ] to have a length of 2 but got 1`. Restored the fix via `git checkout cf7f2829 -- src/store/generation-stream-runner.ts` and re-ran — GREEN, 26/26.
- [x] `npm run typecheck` (frontend `tsc --noEmit` + `server` typecheck) — clean.
- [x] `npm run lint` (`eslint . --max-warnings 0`) — clean.
- [ ] `npm run verify` — not run standalone; the scoped typecheck/test/lint above cover the changed files and their commit's own pre-commit/pre-push hooks already gated this on push.


dudarenok-maker:
## PR review — pass 1 (head cf7f2829, depth medium)

Scope: both changed files (`src/store/generation-stream-runner.ts`, `src/store/queue-dispatcher-middleware.test.ts`) read in full, plus their live context — `src/store/queue-dispatcher-middleware.ts`, `src/store/queue-thunks.ts`, `src/lib/api.ts`'s `realStreamGeneration`/`mockStreamGeneration`, `src/store/chapters-slice.ts`'s `applyGenerationTick`, `server/src/routes/generation.ts` (park emission, `res.on('close')` orphan recovery, subscribe/catch-up replay, job teardown), `server/src/routes/queue.ts` + `server/src/workspace/queue-io.ts` (`confirmFallback`/`retry`/`completeEntry`/`resetEntryToQueued`), and `docs/features/146-loud-fallback-gate.md`. Verified: all 21 required checks green at `cf7f2829` (`gh pr checks 3144`), `Closes #3029` present and plain, issue #3029 open and labeled `bug`, before adversarially probing the re-entrant cancel, stale-tick, cross-boundary and class-left-armed shapes below.

### 🔴 Blocking

None.

### 🟠 Significant

None. I found no correctness bug in the change itself. The probes I ran against it are listed under "What is solid" so the record shows what was and was not interrogated.

### 🟡 Minor

**M1 — cleanup — `docs/features/146-loud-fallback-gate.md` is `status: active`, names both changed files under `Key files`, documents the exact behaviour this PR inverts, and is not updated.**

- `docs/features/146-loud-fallback-gate.md:40` — "…the Redux copy isn't guaranteed to have caught up to `awaiting_confirm` by the time the reconcile runs off **the following `idle` tick**." After this PR a parked chapter's reconcile never runs off an `idle` tick: the park's own `close(key)` dispatches `clearActiveStream`, which is what wakes `tick()`; the `idle` that follows is dropped at `generation-stream-runner.ts:349` before it reaches anything.
- `docs/features/146-loud-fallback-gate.md:52` — describes the vitest coverage as exercising "the REAL SSE tick ordering (park tick then idle, no interleaved queue-snapshot dispatch)". The new test (`queue-dispatcher-middleware.test.ts:731`) deliberately omits the idle and is not listed.
- `docs/features/146-loud-fallback-gate.md:30` — "srv-12 orphan-reset: the worker flips to `awaiting_confirm` (serialized) BEFORE the stream closes, so `resetEntryToQueued` (in_progress-only) no-ops on the parked entry." This sentence is now **load-bearing in a way it was not before this PR**, and the change records that nowhere. Pre-PR the client never aborted at the park, so `res.on('close')` (`server/src/routes/generation.ts:1308`) only fired after `deregisterJob` (`:2306`) and its `inFlightByChapter.get(key) === job` guard was already false — the orphan-recovery branch was unreachable on the park path. Post-PR `close()` → `handle.cancel()` → `controller.abort()` runs synchronously inside the park tick, so if the chapter-boundary `await getSidecarRecyclePending()/triggerSidecarRecycle()` at `:2301-2302` holds the loop open, that branch now genuinely runs. It is still **safe** — `resetEntryToQueued` (`server/src/workspace/queue-io.ts:381`) requires `status === 'in_progress'` and the entry is `awaiting_confirm` by then (`generation.ts:1454` awaits the flip before `broadcast` at `:1457`) — but the PR now leans on that guard and the plan doesn't say so.

CLAUDE.md before-shipping step 1: "Changed behaviour cited in an existing plan → update that plan in the same diff."

**M2 — cleanup — two comments the change made false, both in the touched file.**

- `src/store/generation-stream-runner.ts:527-529` — "Record the park so the dispatcher's reconcile **(which runs off the `idle` tick that follows)** knows this stream close was a park, not a completion". This now sits *immediately above* the new comment at `:531-537`, which asserts the opposite ("A subsequent `idle` tick (if it arrives) is a no-op"). Two adjacent comments describing the same eight lines contradict each other; a reader arriving at this branch cannot tell which one is current.
- `src/store/generation-stream-runner.ts:203-207` — the `chapterAwaitingConfirm` declaration: "read+cleared by the dispatcher (`takeChapterAwaitingConfirm`) **during the reconcile that follows the park's `idle` close**". There is no longer a park `idle` close at all.

Per CLAUDE.md, "a stale comment or doc sentence your change made false" is a chore fixed in the same round, not taste.

**M3 — cleanup — the release-notes pair is missing, with no explicit not-applicable.**

The PR ships a user-visible delta (a fallback you confirmed re-dispatches immediately instead of sitting stalled). Neither `docs/release-notes-next.md` (open `release-notes-next-version: 1.15.0` cycle) nor `RELEASE_NOTES.md`'s in-progress `# Castwright 1.15.0` section is touched, and the PR body neither adds an entry nor states the step is N/A — it only lists test/typecheck/lint. Sibling `fix(...)` commits already on `main` do carry the entry (`8eaf844e`, `9b7067ae` both touch `docs/release-notes-next.md`). Before-shipping step 5 requires both files or an explicit skip statement.

**M4 — cleanup — `parkStream`'s trailing `idle` is now unreachable, and its docstring still claims it drives the close.**

`src/store/queue-dispatcher-middleware.test.ts:185-193`. Post-fix, line 192's `onTick({ type: 'idle' })` hits `handles.get(key) === undefined` at `generation-stream-runner.ts:349` and returns before touching anything. Concretely: delete line 192 and both consumers (`:671` and `:696`) still pass — the helper's second half now measures nothing, while its comment (`:178-184`) still says the pair of ticks is what "#1284's regression test relies on". If the trailing tick is retained deliberately to pin the new "a subsequent idle is a no-op" claim, nothing asserts that — no test observes the second tick's effect either way. Either drop it or say in the comment why it stays.

**M5 — cleanup (house gate) — no on-box acceptance row, and the PR's own mechanism does not account for the duration #3029 reports.**

#3029's evidence is an on-box observation (mechanical batch 1 step 4, row A5 bullet 4): a confirmed entry sat unclaimed "for minutes". The window this PR closes is, on the ordinary server path, sub-millisecond — after the park returns, `generation.ts` reaches `deregisterJob` (`:2306`) and `endAllSubscribers(job, { type: 'idle' })` (`:2312`) with only a microtask boundary in between, so `hasOpenStreamForChapter` is true for that long and no longer. A minutes-long stall therefore requires the `idle` to be delayed or never sent — plausibly the `await triggerSidecarRecycle()` at `:2301-2302`, or an exception escaping the chapter loop — and the PR names neither, rules neither out, and ships no evidence from the real flow.

The change is a genuine defect fix and its jsdom test proves the redux seam. What it does not prove is that the *reported* symptom is gone, which needs a real book, a real Qwen fallback set and the browser confirm path. Per CLAUDE.md before-shipping step 3 that does not block the merge — but recording it does: a row is owed in `docs/testing/onbox-acceptance-register.md` (plus the live view, republished to the URL in the register header) saying concretely what to observe (park a chapter on the loud-fallback gate on a real mixed-cast book, click "Render anyway", observe the chapter starts rendering without any other UI interaction), or the PR should say why no row is owed. Alternatively, downgrade `Closes #3029` to `Refs #3029` and keep the issue open pending on-box confirmation.

### ✅ What is solid

Probed and cleared — each of these was a plausible way for this change to break something, and none of them does:

- **Re-entrant `controller.abort()` from inside `onTick`.** `close()` → `handle.cancel()` → `api.ts:5933`'s `controller.abort()` fires while the parse loop at `api.ts:5873-5889` is still mid-buffer. `cancelled` is set by the abort listener (`:5838-5840`) before control returns, and `shouldReconnect` (`:5895`) is `!cancelled && …`, so the fix cannot provoke a spurious reconnect. Frames already in the same buffer are then dropped at `generation-stream-runner.ts:349` — which is the intended no-op, and on this path only `idle` can follow a park.
- **A stale tick landing on a *new* handle under the same key** (park → close → user confirms → dispatcher re-opens `${bookId}::${chapterId}` → old stream's `idle` arrives and tears the new one down). Not reachable: the real path aborts the fetch so no further frames arrive, and the mock path never emits `chapter_awaiting_fallback_confirm` at all (no occurrence anywhere in `src/lib/api.ts` or `src/mocks/`), with a synchronous `clearInterval` canceller besides.
- **Server catch-up replay re-delivering the park to a reconnecting client.** `generation.ts:1226-1258` replays only `progress` (or a synthetic `progress` marker), never the park tick, so a reattach cannot close a freshly-opened stream.
- **A premature `generation_run_complete`.** `close()` only emits the rollup when `handle.completedChapterIds.length > 0`; under queue-sole concurrency one stream renders one chapter and the server parks *before* rendering, so the array is always empty at a park.
- **The back-compat `${bookId}::*` handle**, where closing on one chapter's park would kill the sibling chapters of a multi-chapter run. Not reachable: the only non-test `runner.open(...)` call site is `queue-dispatcher-middleware.ts:266`, which always passes `chapterId`.
- **One instance fixed, class left armed** (rubric #10). The only other transition that can leave an entry `queued` while its chapter's stream is still open is Retry — and `retry()` (`server/src/workspace/queue-io.ts:275-283`) no-ops unless the entry is already `failed`, a state only reachable *after* STEP 1 ran, i.e. after the stream closed. So `confirm-fallback` is genuinely the only site in this class; the fix is complete for it.
- **The reconcile still reaches STEP 1.** `close()`'s `clearActiveStream` is in the dispatcher's wake-up list (`queue-dispatcher-middleware.ts:296`), so `takeChapterAwaitingConfirm` is still consumed exactly once and the entry is neither `/complete`d nor added to `completed` — #1284 stays fixed.
- **The paired test is a real regression test, red for the reason claimed.** Pre-fix, the park leaves `handles` populated, so STEP 1's `!runner.hasOpenStreamForChapter(...)` is false, `inFlight` keeps `a1`, and STEP 2 skips at `:229`/`:231` — one open, not two, exactly the `expected [ [ 1 ] ] to have a length of 2` the PR body reports.
- Derived artifacts: no `openapi.yaml` change, and #3029 is a `bug` issue, which correctly renders in no `docs/BACKLOG.md` row. No config knob, no `cast.json` write, no lock-timeout seam touched.

### Verdict

**No correctness bugs.** The change is sound and the regression test is real. Five 🟡 findings, of which M1, M3 and M5 are house gates that should land before merge (stale active plan, missing release-notes pair, unrecorded on-box acceptance / unproven `Closes`), and M2/M4 are stale comments the change made false — chores under CLAUDE.md's incidental-findings rule, not taste. Per the re-review trigger, a cleanup-only pass does **not** re-trigger a review once these are fixed and pushed.

Issue #3137, opened 2026-09-09
[agent instructions][claude][verify] Verify: guard-exported scan targets drive forceRerunTriggers


dudarenok-maker:
[agent instructions][claude][verify] Verify: guard-exported scan targets drive forceRerunTriggers

## What

Verify the combined work of #3135 and #3136 (children of parent #3085): four runtime-scanning guards (`state-language.guard.test.ts`, `spawn-windows-hide.test.ts`, `cast-lock.guard.test.ts`, `engine-language-coverage.guard.test.ts`) should each export their own scan-target scope, `server/vitest.config.ts`'s `forceRerunTriggers` should cover all four correctly, and `force-rerun-triggers.test.ts` should check the real triggers against each guard's own export — never a re-typed literal that could independently drift.

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3085-guard-trigger-generation`
Branch: `chore/ops-3085-guard-trigger-generation` (already checked out there, with both children's commits on it)

Do NOT create another worktree. Do NOT work in the primary checkout. Read-only
for source: report a verdict, do not fix anything yourself.

## Parent's acceptance, restated as a checklist

1. Four sibling scope-constant modules (or however child 1's shared-module shape actually landed) exist, one per guard, each exporting a named constant describing that guard's real scan scope.
2. Each of the four guard test files imports its constant rather than hard-coding its scan root independently of it.
3. `server/vitest.config.ts`'s `forceRerunTriggers` correctly covers: `server/src/**` (for `state-language.guard.test.ts` and `cast-lock.guard.test.ts`), `spawn-windows-hide.test.ts`'s actual scan root(s) (read the file yourself to confirm — it scans outside `server/src` too, per #3085's own text), and `src/routes/generation.ts` (for `engine-language-coverage.guard.test.ts`).
4. `server/src/force-rerun-triggers.test.ts` imports all four constants (not four re-typed literals) and checks the real triggers against them.
5. The old hand-typed `MAIN_COVERED` entry for `src/routes/generation.ts` (the #3059 one) is gone or now DERIVED from the new constant — not duplicated alongside it.
6. `git grep -n "readFileSync"` for guard-shaped test files that still hand-type their scope (i.e., confirm nothing was silently left half-converted) — cross-check against the four names above; anything claimed as converted must actually be converted.
[...]

Issue #3136, opened 2026-09-09
[agent instructions][claude][task] Apply the guard-exported-scan-target convention to the remaining runtime-scanning guards


dudarenok-maker:
[agent instructions][cline][task] Apply the guard-exported-scan-target convention to the remaining runtime-scanning guards

## What

Child 1 (#3135, merged onto this branch before you start) established the convention for #3085: a runtime-scanning guard exports its own scan-target scope from a small sibling module, and `server/src/force-rerun-triggers.test.ts` checks the real `forceRerunTriggers` array against that export rather than a hand-typed, independently-maintained literal. Apply the SAME convention — same module-naming shape, same export style, same wiring into `force-rerun-triggers.test.ts` — to the remaining runtime-scanning guards named in #3085:

- `server/src/spawn-windows-hide.test.ts`
- `server/src/workspace/cast-lock.guard.test.ts`
- `server/src/tts/engine-language-coverage.guard.test.ts`

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3085-guard-trigger-generation`
Branch: `chore/ops-3085-guard-trigger-generation` (already checked out there, with child 1's commit(s) on it — read that diff first with `git log -p` or `git show` to see the exact convention established)

Do NOT create another worktree. Do NOT work in the primary checkout. Commit on
this branch. Do NOT open a PR — the verify child does that.

## Background you need

- Read child 1's actual committed diff on this branch FIRST (`git log --oneline` then `git show <sha>` for its commit) — do not guess the convention from the issue text alone; follow exactly what was actually built (module name pattern, export name pattern, how `force-rerun-triggers.test.ts` was restructured to consume it).
- `server/src/spawn-windows-hide.test.ts` scans the tree OUTSIDE `server/src` too, including `scripts/**` (per #3085's own body) — read the whole file to find its actual scan root(s) before writing its scope constant; do not assume it matches `state-language.guard.test.ts`'s shape.
- `server/src/workspace/cast-lock.guard.test.ts` scans the `cast.json` write sites across `server/src/**` via `collectSourceFiles(SRC_ROOT)` — same tree-wide shape as child 1's guard. Its scope constant should look like child 1's (`server/src/**`), following the exact convention already established.
[...]

Issue #3135, opened 2026-09-09
[agent instructions][claude][task] Make state-language.guard.test.ts export its own scan targets; generate its forceRerunTriggers entries from that export


dudarenok-maker:
[agent instructions][claude][task] Make state-language.guard.test.ts export its own scan targets; generate its forceRerunTriggers entries from that export

## What

Establish the convention chosen for #3085: a runtime-scanning guard (one that reads its target as TEXT via `readFileSync` rather than importing it, so vitest's module graph has no edge to it and `--changed` can't select it) exports its own scan-target set, and `server/src/force-rerun-triggers.test.ts` checks the real `forceRerunTriggers` array against THAT export instead of a hand-typed, independently-maintained list. This removes the drift class at its root: the guard becomes the single source of truth for what it needs, and it becomes structurally impossible for the trigger list and the guard's actual scan scope to disagree.

This first child establishes the pattern on ONE guard, `server/src/workspace/state-language.guard.test.ts` — a tree-wide scanner (walks all of `server/src/**` via `collectSourceFiles(SRC_ROOT)`), the clearest and most-cited example in #3085's own body. A second child applies the same established pattern to the remaining guards.

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3085-guard-trigger-generation`
Branch: `chore/ops-3085-guard-trigger-generation` (already checked out there)

Do NOT create another worktree. Do NOT work in the primary checkout. Commit on
this branch. Do NOT open a PR — the verify child does that.

## Background you need

- `server/src/force-rerun-triggers.test.ts` (whole file) is the guard that pins `server/vitest.config.ts`'s `forceRerunTriggers` array (`server/vitest.config.ts:193-...`). It currently hand-maintains `MAIN_COVERED`/`SLOW_COVERED`/`NOT_COVERED` arrays and cross-checks them against the real `forceRerunTriggers` glob patterns via `picomatch`. This works for files that genuinely need a fixed, small trigger (manifests, config files) but is a hand-typed list for the runtime-scanning-guard entries too — e.g. the entry `{ rel: 'src/routes/generation.ts', file: 'the generation route (resolveEligibleEngines call site)', base: SERVER_ROOT }` exists ONLY because someone remembered to add it when `engine-language-coverage.guard.test.ts` was written (#3059). Nothing enforces that every runtime-scanning guard actually HAS a corresponding entry — that's exactly the gap #3085 exists to close.
[...]

PR #3151, opened 2026-09-10
chore(ops): pin forceRerunTriggers to each guard's own scan-target export via drift check

### LINKED ISSUE #3085: ops — a runtime-scanning guard is not selected by --changed, so scoped CI legs cannot enforce it
## What

`verify.yml`'s server leg runs `vitest run --changed`, which selects tests whose **module graph** touches the diff. A guard that reads its target as **text at runtime** (`readFileSync`) has no module-graph edge to that target, so a diff touching only the scanned file does not select the guard.

`server/vitest.config.ts` already knows this class — `forceRerunTriggers` exists for it and its comment names the trap — but the list is **per-file and hand-maintained**, so it covers only the targets someone remembered to add.

## Measured, not inferred

PR #3083 removed a block from `server/src/routes/generation.ts`. That stales `state-language.guard.test.ts`'s hardcoded per-file count of `stateJsonPath(` sites (`'routes/generation.ts': 2`, now 1). On run `34167026051`:

| Leg | Files run | Result |
|---|---|---|
| Server tests (fast + slow) — Linux, `--changed`-scoped | **31** | **success** |
| Windows (test + test:server) — full suite | 543 | failure, caught it |

The guard was caught **by accident**, on the one leg that happens not to scope. Had the Windows leg been scoped the same way, a stale guard count would have merged green.

## The gap

`git grep -l "readFileSync" server/src --include=*.guard.test.ts` (plus `spawn-windows-hide.test.ts`, which is not `.guard.`-named) finds several scanners whose targets are broad trees, not single files:

- `state-language.guard.test.ts` — scans ~23 files across `server/src/**`
- `spawn-windows-hide.test.ts` — scans the tree outside `server/src`, including `scripts/**`
- `cast-lock.guard.test.ts` — scans the `cast.json` write sites
- `engine-language-coverage.guard.test.ts` (new in #3083) — scans `routes/generation.ts`

A single-file trigger works for the last one and is being added there. It does not generalise: adding `server/src/**` to `forceRerunTriggers` would force a full server run on essentially every server PR, which is the cost `--changed` exists to avoid.

## The decision owed

**How should a runtime-scanning guard declare its inputs, given that `forceRerunTriggers` is a hand-maintained per-file list and the scan targets are trees?** At least three defensible shapes:

1. **Widen the triggers to the scanned trees** and accept that most server PRs run the full suite. Simplest, honest, and the most expensive.
2. **Have each guard export its scan target set, and generate the triggers from that** — the guard becomes the single source of truth, and `force-rerun-triggers.test.ts` checks the generated list rather than a hand-typed one. Removes the class of drift entirely; costs a build step and a convention every future guard must follow.
3. **Route runtime-scanning guards into their own always-run lane** — cheap and unscoped, since these guards are pure text scans and take milliseconds. Keeps `--changed` narrow for the real suite; costs a second lane and a rule about what belongs in it.

Not filed as a defect to fix in passing: whichever way this goes changes a convention every future guard has to follow, which is a design pass rather than a chore.

## Key files

- `server/vitest.config.ts` — `forceRerunTriggers`, and the comment describing the trap
- `server/src/force-rerun-triggers.test.ts` — the covered/`NOT_COVERED` case list that pins the triggers
- `.github/workflows/verify.yml` — the `--changed` server leg
- the guards listed above

## Benefit

*Benefit (technical):* a guard that cannot be selected by the change it exists to catch is indistinguishable from no guard, and green CI actively asserts the opposite. This repo has invested heavily in static guards; this is the mechanism that decides whether they fire. Two separate guards went red on only one CI leg in a single session's work — the pattern is already recurring.

Found while diagnosing why #3083's stale-count failure appeared on the Windows leg only.

dudarenok-maker:
## Summary
- state-language.guard.test.ts, spawn-windows-hide.test.ts,
  cast-lock.guard.test.ts, and engine-language-coverage.guard.test.ts each
  now export their own real scan-target scope from a sibling module.
- server/vitest.config.ts's forceRerunTriggers covers all four from those
  exports; force-rerun-triggers.test.ts checks the real triggers against
  the SAME exported constants, so a guard's scope and its CI trigger can
  no longer independently drift (#3085).
- coqui-residency-policy.guard.test.ts (same runtime-scanning shape) was
  NOT converted in this PR — flagged as a follow-up (#3170), not silently dropped.

## Test plan
- [x] 4 mutations, each showing force-rerun-triggers.test.ts catch a
      deliberately-wrong scope constant, then reverted
- [x] npm run test:server passes (1 pre-existing, unrelated flaky failure in
      library-cast-override.test.ts's AB/BA deadlock race test, untouched by
      this branch; all 5 in-scope files — force-rerun-triggers.test.ts,
      state-language.guard.test.ts, spawn-windows-hide.test.ts,
      cast-lock.guard.test.ts, engine-language-coverage.guard.test.ts —
      pass cleanly: 120/120)
- [x] npm run typecheck passes

## Release notes
N/A — internal test/guard infrastructure with no user- or operator-visible effect.

Closes #3085

### DIFF
--- a/server/src/force-rerun-triggers.test.ts
+++ b/server/src/force-rerun-triggers.test.ts
@@ -250,10 +250,14 @@ const SLOW_NOT_COVERED = [
 /* #3085: state-language.guard.test.ts's new trigger deliberately covers the
    whole server/src/** tree, so "an ordinary server source file" is no longer
    a valid not-covered case for the MAIN config — matching it is the point.
-   Swap in a file outside every guard's declared scope instead. */
+   Swap in files outside every guard's declared scope instead — one for each
+   major extension type (non-source, source-extension) so widening a trigger
+   to a broad glob (e.g. ** + any-extension) is caught rather than only the
+   crudest widening. */
 const MAIN_NOT_COVERED = [
   ...NOT_COVERED,
   { rel: 'CONTRIBUTING.md', file: 'an ordinary repo file outside every guard scope', base: REPO_ROOT },
+  { rel: 'src/lib/account-defaults.ts', file: 'a frontend source file outside every guard scope', base: REPO_ROOT },
 ];
 
 const crossProduct = (covered: typeof MAIN_COVERED) =>
@@ -307,6 +311,22 @@ describe('server/vitest.config.ts forceRerunTriggers', () => {
       `{**/${ENGINE_LANGUAGE_COVERAGE_GUARD_SCAN_GLOB},` +
       `**/.*/**/${ENGINE_LANGUAGE_COVERAGE_GUARD_SCAN_GLOB}}`;
     expect(mainTriggers).toContain(expected);
+  });
+
+  /* coqui-residency-policy.guard.test.ts (#1932, side-18) reads
+     server/src/tts/synthesise-chapter.ts at RUNTIME — the file is importable
+     but the guard doesn't import it, it scans its source text for eviction
+     policy cross-references, so there is no module-graph edge from the guard
+     to this file. The new broad server/src/** trigger added in #3085 now
+     covers this file, making the specific entry able to be deleted without
+     failing this test suite — but the guard still declares
+     synthesise-chapter.ts as its own scope. This exact-entry pin guards against
+     the entry being silently deleted on the assumption that the broader trigger
+     handles it (it does, today, but the guard's scope is still the specific
+     file). */
+  it('main forceRerunTriggers has the exact entry for coqui-residency-policy.guard.test.ts synthesise-chapter.ts (#3085)', () => {
+    const coquiSynthesiseEntry = '{**/server/src/tts/synthesise-chapter.ts,**/.*/**/server/src/tts/synthesise-chapter.ts}';
+    expect(mainTriggers).toContain(coquiSynthesiseEntry);
   });
 
   /* Only entries 1, 4 and 6 have their OWN literal brace-glob trigger to pin
--- a/server/src/force-rerun-triggers.test.ts
+++ b/server/src/force-rerun-triggers.test.ts
@@ -30,6 +30,7 @@ import { existsSync } from 'node:fs';
 import { relative, resolve } from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS } from './tts/coqui-residency-policy.guard-targets.js';
+import { STATE_LANGUAGE_GUARD_SCAN_GLOB } from './workspace/state-language.guard-targets.js';
 
 const SERVER_ROOT = resolve(__dirname, '..');
 const REPO_ROOT = resolve(SERVER_ROOT, '..');
@@ -170,6 +171,18 @@ const MAIN_COVERED = [
      #1847 runtime-read trap as the entries above. */
   { rel: 'src/analyzer/rate-limit.ts', file: 'the rate-limit dynamic-reader lookup', base: SERVER_ROOT },
   { rel: 'src/tts/segment-asr-qa.ts', file: 'the per-language maxWer dynamic-reader lookup', base: SERVER_ROOT },
+  /* state-language.guard.test.ts (#3085): a tree-wide scanner —
+     collectSourceFiles(SRC_ROOT) reads every non-test .ts file under
+     server/src/** at RUNTIME, no module-graph edge to any of them. The rel
+     path below is DERIVED from STATE_LANGUAGE_GUARD_SCAN_GLOB (the same
+     constant the guard itself imports), not a second hand-typed literal that
+     happens to agree with it today — so this assertion and the guard's
+     declared scope can never independently drift. */
+  {
+    rel: STATE_LANGUAGE_GUARD_SCAN_GLOB.replace(/\*\*$/, 'index.ts'),
+    file: 'a file under the state-language guard scan scope (STATE_LANGUAGE_GUARD_SCAN_GLOB)',
+    base: REPO_ROOT,
+  },
 ];
 
 const SLOW_COVERED = [
@@ -181,9 +194,24 @@ const SLOW_COVERED = [
    trigger to something like `**` + a suffix glob is caught rather than only
    the crudest `**`. */
 const NOT_COVERED = [
-  { rel: 'src/index.ts', file: 'an ordinary server source file', base: SERVER_ROOT },
   { rel: 'tsconfig.json', file: 'a JSON file that is not a manifest', base: REPO_ROOT },
   { rel: 'apps/android/pubspec.yaml', file: 'a YAML file that is not the contract', base: REPO_ROOT },
+];
+
+/* The slow config's triggers are untouched by #3085, so an ordinary server
+   source file is still a valid not-covered case there. */
+const SLOW_NOT_COVERED = [
+  ...NOT_COVERED,
+  { rel: 'src/index.ts', file: 'an ordinary server source file', base: SERVER_ROOT },
+];
+
+/* #3085: state-language.guard.test.ts's new trigger deliberately covers the
+   whole server/src/** tree, so "an ordinary server source file" is no longer
+   a valid not-covered case for the MAIN config — matching it is the point.
+   Swap in a file outside every guard's declared scope instead. */
+const MAIN_NOT_COVERED = [
+  ...NOT_COVERED,
+  { rel: 'CONTRIBUTING.md', file: 'an ordinary repo file outside every guard scope', base: REPO_ROOT },
 ];
 
 const crossProduct = (covered: typeof MAIN_COVERED) =>
@@ -210,10 +238,22 @@ describe('server/vitest.config.ts forceRerunTriggers', () => {
     },
   );
 
+  /* #3085: the file-coverage case above only proves the CURRENT scope is
+     covered — narrowing STATE_LANGUAGE_GUARD_SCAN_GLOB to a subtree (e.g.
+     'server/src/tts/**') would still pass it, because a narrower scope is
+     still a subset of the real (unchanged) trigger below. This assertion
+     checks the trigger array contains the EXACT brace-glob built from the
+     imported constant, so narrowing OR widening the constant without
+     updating vitest.config.ts's literal entry to match is caught either way. */
+  it('main forceRerunTriggers has the exact entry derived from STATE_LANGUAGE_GUARD_SCAN_GLOB (#3085)', () => {
+    const expected = `{**/${STATE_LANGUAGE_GUARD_SCAN_GLOB},**/.*/**/${STATE_LANGUAGE_GUARD_SCAN_GLOB}}`;
+    expect(mainTriggers).toContain(expected);
+  });
+
   /* Guards against "fixing" a dead trigger by widening it to something that
      matches everything — that would force a full run on every diff and
      quietly undo the point of --changed. */
-  it.each(crossProduct(NOT_COVERED))('does not match $file from $shape', ({ rel, base, root }) => {
+  it.each(crossProduct(MAIN_NOT_COVERED))('does not match $file from $shape', ({ rel, base, root }) => {
     expect(matchesTrigger(mainTriggers, absPathUnder(root, base, rel))).toBe(false);
   });
 });
@@ -226,7 +266,7 @@ describe('server/vitest.config.slow.ts forceRerunTriggers', () => {
     },
   );
 
-  it.each(crossProduct(NOT_COVERED))('does not match $file from $shape', ({ rel, base, root }) => {
+  it.each(crossProduct(SLOW_NOT_COVERED))('does not match $file from $shape', ({ rel, base, root }) => {
     expect(matchesTrigger(slowTriggers, absPathUnder(root, base, rel))).toBe(false);
   });
 });
--- a/server/src/force-rerun-triggers.test.ts
+++ b/server/src/force-rerun-triggers.test.ts
@@ -31,6 +31,9 @@ import { relative, resolve } from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS } from './tts/coqui-residency-policy.guard-targets.js';
 import { STATE_LANGUAGE_GUARD_SCAN_GLOB } from './workspace/state-language.guard-targets.js';
+import { SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS } from './spawn-windows-hide.guard-targets.js';
+import { CAST_LOCK_GUARD_SCAN_GLOB } from './workspace/cast-lock.guard-targets.js';
+import { ENGINE_LANGUAGE_COVERAGE_GUARD_SCAN_GLOB } from './tts/engine-language-coverage.guard-targets.js';
 
 const SERVER_ROOT = resolve(__dirname, '..');
 const REPO_ROOT = resolve(SERVER_ROOT, '..');
@@ -105,26 +108,43 @@ const MAIN_COVERED = [
   { rel: 'vitest.config.slow.ts', file: 'the slow-tier config', base: SERVER_ROOT },
   /* Pins the `vite` half of the brace. Without a case for it, narrowing the
      trigger to `vitest.config.ts` would leave every assertion green while
-     silently dropping coverage for the vite build config. */
+     silently dropping coverage for the vite build config. Also one of the
+     concrete root files SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[5] names
+     (#3085) — its coverage here doubles as that guard's file-coverage case. */
   { rel: 'vite.config.ts', file: 'the vite build config', base: REPO_ROOT },
   { rel: 'openapi.yaml', file: 'the API contract', base: REPO_ROOT },
-  /* #2567 review round 3: spawn-windows-hide.test.ts (this suite) reads
-     these trees as TEXT at RUNTIME to scan for a missing windowsHide —
-     no module-graph edge, so a diff confined to any of them selected zero
-     tests under `vitest run --changed` before these triggers existed. */
-  { rel: 'scripts/verify-cache.mjs', file: 'an arbitrary scripts/** file', base: REPO_ROOT },
+  /* #2567 review round 3, re-derived from SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS
+     (#3085): spawn-windows-hide.test.ts (this suite) reads these trees as
+     TEXT at RUNTIME to scan for a missing windowsHide — no module-graph
+     edge, so a diff confined to any of them selected zero tests under
+     `vitest run --changed` before these triggers existed. Each `rel` below
+     is DERIVED from the guard's own declared scope, not a second hand-typed
+     literal that happens to agree with it today. */
   {
-    rel: 'server/tts-sidecar/scripts/install-ort.mjs',
-    file: 'a server/tts-sidecar/scripts/** file (via the scripts/** trigger)',
+    rel: SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[1].replace(/\*\*$/, 'verify-cache.mjs'),
+    file: 'an arbitrary scripts/** file (SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS)',
     base: REPO_ROOT,
   },
   {
-    rel: 'pinokio-scripts/lib/resolve-release.js',
-    file: 'a pinokio-scripts/** file',
+    rel: SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[2].replace(/\*\*$/, 'install-ort.mjs'),
+    file: 'a server/tts-sidecar/scripts/** file (via the scripts/** trigger; SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS)',
     base: REPO_ROOT,
   },
-  { rel: 'e2e/global-teardown.ts', file: 'the e2e Playwright teardown', base: REPO_ROOT },
-  { rel: 'launch.mjs', file: 'the versioned-dir launcher', base: REPO_ROOT },
+  {
+    rel: SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[3].replace(/\*\*$/, 'resolve-release.js'),
+    file: 'a pinokio-scripts/** file (SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS)',
+    base: REPO_ROOT,
+  },
+  {
+    rel: SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[4],
+    file: 'the e2e Playwright teardown (SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS)',
+    base: REPO_ROOT,
+  },
+  {
+    rel: SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS[6],
+    file: 'the versioned-dir launcher (SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS)',
+    base: REPO_ROOT,
+  },
   /* #2588 pass-2 review: venv-migration.test.ts (this suite) reads BOTH of these
      at RUNTIME — .gitattributes to assert the requirements/*.txt LF pin rule is
      declared, the requirements files themselves to assert the pin materialised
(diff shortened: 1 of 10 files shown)


dudarenok-maker:
## PR review — pass 1 (head e04af668f3996330bcf6df48588c6a7afc932f89, depth low)

**Scope / verified:** all 10 changed files read at head. Cross-checked against the real `forceRerunTriggers` arrays in `server/vitest.config.ts` (15 entries) and `server/vitest.config.slow.ts` (2 entries); against `externalFilesFloor()`'s actual scan roots in `server/src/spawn-windows-hide.test.ts`; and against #3085's option list plus #3136's implementation brief. The four new exact-entry `toContain` assertions were each matched by hand to a real literal in `server/vitest.config.ts` (lines 216, 218, 219, 252, 276) — all four resolve. CI is green on all 21 checks; `Closes #3085` is present and outside any code span. House gates N/A and confirmed N/A: no `cast.json` writes, no config knob, no derived artifact (`api-types.ts` / `BACKLOG.md` / brand PNGs untouched), no hardware-provable behaviour, so no on-box register row is owed.

### Correctness bugs

**1 — Three of the seven "scan scope" assertions are compile-time tautologies and cannot observe the guard's real scan roots.** `server/src/spawn-windows-hide.test.ts:502-516`

The new test declares its own copies of the scan roots inside the test body:

```ts
const scriptsDir = join(REPO_ROOT, 'scripts');
const ttsDir     = join(REPO_ROOT, 'server', 'tts-sidecar', 'scripts');
const pinokioDir = join(REPO_ROOT, 'pinokio-scripts', 'lib');
```

The guard's *real* roots are function-local `const`s inside `externalFilesFloor()` at `:263`, `:267` and `:271` — not exported, not module-level, and not read by this test. So `expect(toRepoRel(scriptsDir)).toBe(GLOBS[1].replace(/\/\*\*$/, ''))` reduces to `expect('scripts').toBe('scripts')`: two independently-typed copies of the same string, neither of which is the thing being scanned. The test's own comment at `:498-501` claims it "Ties this guard's ACTUAL scan roots to the scope it DECLARES" — for indices `[1]`, `[2]`, `[3]` it does not.

The other four indices are done correctly and are the contrast: `[0]` reads module-level `SRC_ROOT` (the same binding `listSourceFiles(SRC_ROOT)` uses at `:493`), `[4]` checks `EXTERNAL_FILES_MANUAL` **exhaustively** with `toEqual([...])`, and `[5]`/`[6]` read the dynamically-built `EXTERNAL_FILES_FLOOR` — with a comment at `:519-522` stating exactly the right principle.

*Concrete break:* add a fourth candidate directory to `externalFilesFloor()` — say `const toolsFiles = listFilesRecursive(join(REPO_ROOT, 'tools'), ['.mjs'])`, spread into `candidates` at `:275` — and drop one spawning `.mjs` in it. The function has grown its root list twice already (its own comment at `:234-241` records the history), so this is the expected next edit. Result: the guard now scans `tools/**`, `SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS` still lists seven entries with no `tools`, `server/vitest.config.ts` gets no trigger for it, and a diff confined to `tools/foo.mjs` selects zero tests under `vitest run --changed` — the guard is silent on precisely the file it exists to guard. Every assertion in this PR stays green: this test compares its own re-typed literals to the unchanged constant, `force-rerun-triggers.test.ts:122-136` derives its covered paths from the same unchanged constant, and the "old hardcoded list" test at `:717` is a subset check that a widening passes. That is #3085's exact failure mode, reintroduced inside the fix for it.

*Fix:* hoist `scriptsDir` / `ttsDir` / `pinokioDir` to module scope (or have `externalFilesFloor()` return/expose its root list) so `externalFilesFloor()` and this test read one binding — and assert the root set **exhaustively** with `toEqual`, the way `EXTERNAL_FILES_MANUAL` is already handled at `:517`, so an added root fails rather than being ignored.

---

**2 — `MAIN_NOT_COVERED` lost its only source-extension negative case, so the widening this file exists to catch is no longer detectable on the main config.** `server/src/force-rerun-triggers.test.ts:241-244` (and the now-false comment at `:222-224`)

The PR moves `{ rel: 'src/index.ts', base: SERVER_ROOT }` out of the main config's negative set (correctly — the new `server/src` trigger matches it by design) and substitutes `CONTRIBUTING.md`. That leaves `MAIN_NOT_COVERED` as `tsconfig.json` + `apps/android/pubspec.yaml` + `CONTRIBUTING.md` — `.json`, `.yaml`, `.md`. Not one `.ts` / `.js` / `.mjs` file remains. The comment at `:222-224` still asserts "More than one shape, so widening a dead trigger to something like `` `**` `` + a suffix glob is caught rather than only the crudest `` `**` ``". That claim is now false for the main config, for exactly the suffix a TypeScript server repo would reach for.

*Concrete break:* widen `server/vitest.config.ts:276` from `'{**/server/src/**,**/.*/**/server/src/**}'` to `'{**/*.ts,**/.*/**/*.ts}'` — a plausible "fix" given the guard scans `.ts` files, and one that forces a full server rerun on every TypeScript change anywhere in the repo, frontend included, killing `--changed` outright. All three `MAIN_NOT_COVERED` cases are non-`.ts`, so none of them match. Every `MAIN_COVERED` case still passes: the `.ts` entries are matched by the widened glob, and the rest (`package.json`, `openapi.yaml`, `.gitattributes`, `main.py`, `base.txt`, the `.mjs`/`.js` script cases, the docs `.md`) are matched by the other 14 triggers, which are untouched. The four new exact-entry `toContain` assertions also pass, since they only require their literals to be *present*. Suite green; `--changed` dead.

This matters more in this PR than it would in any other, because this is the PR that introduces the repo's broadest trigger — the bound on how much broader it can get is the negative set, and the negative set was weakened in the same diff.

*Fix:* add a source-extension file outside every declared scope, e.g. `{ rel: 'src/lib/account-defaults.ts', file: 'a frontend source file outside every guard scope', base: REPO_ROOT }`. I checked it against all 15 main triggers at all three `ROOT_SHAPES` — no match (no `server/src`, `scripts`, or `pinokio-scripts` segment, and not a named root file).

---

**3 — The new broad trigger makes a pre-existing coverage case unable to fail.** `server/vitest.config.ts:276` vs. `server/src/force-rerun-triggers.test.ts:164` — *low severity, but introduced by this diff*

`{ rel: 'src/tts/synthesise-chapter.ts', base: SERVER_ROOT }` resolves to `/repo/server/src/tts/synthesise-chapter.ts`, which the new `server/src` entry now matches. Delete `'{**/server/src/tts/synthesise-chapter.ts,**/.*/**/server/src/tts/synthesise-chapter.ts}'` from `server/vitest.config.ts:235` and the suite stays green — before this PR that deletion reddened this case. The Coqui-residency guard's declared trigger is now silently riding an entry added for three different guards, with nothing pinning it.

This PR added an exact-entry `toContain` pin for each of its four constants precisely because file-coverage alone stopped being sufficient (the comment at `:257-263` says so). `:252`'s generation.ts trigger got one; `:235` did not — the same reasoning applies to it and it is now in the same position. Given the whole framing of the broad trigger is that it is expensive, the person who eventually narrows it back inherits a config whose per-file `server/src` triggers may already have been deleted unnoticed.

*Fix:* one more `toContain` assertion for `:235`'s literal, in the shape already written four times in this PR.

### Cleanup nits

**4 — Nothing is generated; the title and the config comment say otherwise.** `server/vitest.config.ts:265-276`, PR title. The trigger at `:276` is a hand-typed literal that a test *checks* against the constants. That is a sound mechanism, but the comment labels it "#3085's chosen design, option 2", and #3085's option 2 reads "generate the triggers from that … costs a build step". What shipped is option 1's trigger content (widen to the scanned tree) plus a drift check — a defensible fourth shape, but a reader who goes looking for the generator this PR's title promises will not find one. Say what it is.

**5 — No release-notes statement in the PR body.** CLAUDE.md before-shipping step 5 requires an explicit not-applicable rather than silent omission. CI-only, so N/A is the right call — it just has to be said.

**6 — The declared `coqui-residency-policy.guard.test.ts` follow-up cites no issue number.** The PR body says it was "flagged as a follow-up, not silently dropped", and #3136's "Not in scope" section told the implementer to name it "so a follow-up can pick it up". I found no matching issue. Under CLAUDE.md's incidental-findings rule a deferral has to land as an issue number; without one this is the deferral-in-disguise shape that rule exists to close. (Its three targets do already carry triggers at `:235-237`, so there is no CI gap today — only the same two-sources-of-truth drift #3085 exists to close, and note finding 3 above shows `:235`'s pin has just weakened.)

### Considered and not filed

- **The `server/src` trigger forces a full server suite on essentially every server PR.** #3085's own body names this as option 1's cost and #3136's brief sanctions the glob; it is documented in three places in the diff. Deliberate design, not a defect — flagged here only so the record shows it was weighed.
- **`externalFilesFloor()`'s repo-root non-recursive `.mjs`/`.ts` scan is broader than the two files the constant names.** Pre-existing, and honestly disclosed in `spawn-windows-hide.guard-targets.ts`'s own header rather than implied closed.
- Index-based access (`GLOBS[0]`…`GLOBS[6]`) across two files is fragile, but every misalignment I traced fails loud via `absPathUnder`'s `existsSync` gate or the `toEqual`/`toContain` checks. Not a finding.

### Verdict

**Changes requested.** Three correctness bugs — findings 1 and 2 are the substantive ones: finding 1 leaves three of the seven declared scan roots unverified against the code they claim to describe, and finding 2 removes the bound on the broadest trigger in the repo in the same diff that adds it. Both are the drift class this PR was written to close. Re-review is triggered once they are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head b6dfb8f95f6b5ac5638d2453619a8e7add69ede9, depth low)

**Scope.** Re-derived each of pass 1's three correctness bugs against the head tree by replaying the exact failure scenario pass 1 named, not by inspecting whether the code changed. Findings 2 and 3 were settled empirically — finding 2 by running the real 15-entry `forceRerunTriggers` array through `picomatch` at all three `ROOT_SHAPES`, both as-shipped and with the widening pass 1 described applied; finding 3 by byte-comparing the pinned literal against `server/vitest.config.ts:235`. Finding 1 was traced statically because the linkage it needs either exists in the data flow or does not. Plus a normal-depth read of everything `b6dfb8f9` touched (2 files, +51/−17). CI green on all 21 checks at this head.

**Finding 1 — STILL OPEN. Correctness bug.** `server/src/spawn-windows-hide.test.ts:515-526`

The three directory constants were hoisted to module scope, and that closes a real hole — a **modification** of an existing root is now caught. Change `EXTERNAL_PINOKIO_DIR` (`:163`) to `join(REPO_ROOT, 'pinokio-scripts')`, dropping the `lib` segment, and `externalRootsRel[2]` becomes `pinokio-scripts` against an expected `pinokio-scripts/lib`; the `toEqual` at `:526` goes red. Before `b6dfb8f9` that passed. Real improvement, correctly done.

But that is not the failure mode pass 1 filed. Pass 1's break was **adding a fourth root**, and it survives verbatim. The `toEqual` is exhaustive over an array the test declares about itself:

```ts
const externalRoots = [
  EXTERNAL_SCRIPTS_DIR,
  EXTERNAL_TTS_DIR,
  EXTERNAL_PINOKIO_DIR,
];
```

`externalFilesFloor()` does not read that array. It calls `listFilesRecursive` three separate times (`:273`, `:276`, `:279`) and spreads four lists into `candidates` at `:282`. There is no data flow from the function's root set into `externalRoots` — the array is a third hand-maintained copy of the list, just now spelled with constant names instead of `join(...)` calls.

*Concrete break, re-derived at this head.* Apply pass 1's mutation to `externalFilesFloor()`:

```ts
const toolsFiles = listFilesRecursive(join(REPO_ROOT, 'tools'), ['.mjs']);
const candidates = [...rootFiles, ...scriptsFiles, ...ttsFiles, ...pinokioFiles, ...toolsFiles];
```

Then, at head:

- `:526` — `externalRootsRel` is still the same 3 strings, `expectedExternalRootsRel` is still `GLOBS[1..3]`. **Passes.**
- `:528`, `:531` — `SRC_ROOT` and `EXTERNAL_FILES_MANUAL` untouched. **Pass.**
- `:537-539` — `EXTERNAL_FILES_FLOOR` grows, but these are `toContain`. **Pass.**
- `:702`, `:719`, `:731` — I checked every other assertion against the floor in this file: all three are presence/superset checks (`filter(...).length`, "old hardcoded list ⊆ floor"). Every one tolerates growth. **Pass.**
- `SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS` is unchanged, so `force-rerun-triggers.test.ts` derives no new case and its `[1,4,6]` exact-entry check is unaffected. **Passes.**

Suite fully green; `server/vitest.config.ts` gets no `tools` trigger; a diff confined to `tools/foo.mjs` selects zero tests under `vitest run --changed`, and the guard is silent on the tree it now scans. That is #3085's failure mode, still live inside the fix for it. It is also the *expected* next edit — `externalFilesFloor()`'s own header at `:244-251` records that this root list has already grown twice.

A smaller variant survives too: adding the fourth root as a module-scope constant `EXTERNAL_TOOLS_DIR` alongside the other three — i.e. following the convention the new comment establishes — still passes, because the test enumerates three constants by name rather than reading the module's scan set.

**The two comments and the commit message now assert the opposite, and a future editor will believe them.** `:158-160` — "an edit that adds a fourth root to `externalFilesFloor()` fails the test rather than silently agreeing with a stale constant." `:512-514` — "an added fourth root fails this exhaustive `toEqual` check rather than being silently ignored." `b6dfb8f9`'s message — "assert exhaustively with `toEqual` so an added root fails rather than being silently ignored." All three are false as written. This is worse than the original tautology, which at least did not claim to cover the case.

*Fix.* Make the scan set a single module-level value that `externalFilesFloor()` **consumes** — the shape `EXTERNAL_FILES_MANUAL` already has at `:189`, and the reason its check at `:531` is genuinely exhaustive:

```ts
const EXTERNAL_SCAN_ROOTS = [
  { dir: join(REPO_ROOT, 'scripts'), extensions: ['.mjs', '.cjs', '.js'] },
  { dir: join(REPO_ROOT, 'server', 'tts-sidecar', 'scripts'), extensions: ['.mjs'] },
  { dir: join(REPO_ROOT, 'pinokio-scripts', 'lib'), extensions: ['.js', '.mjs'] },
] as const;
// in externalFilesFloor():
const externalCandidates = EXTERNAL_SCAN_ROOTS.flatMap(r => listFilesRecursive(r.dir, [...r.extensions]));
```

Then `expect(EXTERNAL_SCAN_ROOTS.map(r => toRepoRel(r.dir))).toEqual([...])` is exhaustive for real: appending a root is the only way to make the scan see it, and doing so reddens the test. Either land that, or correct all three claims to say what the check actually covers — but the first is what pass 1 asked for and what the PR's own premise requires.

---

**Finding 2 — VERIFIED RESOLVED.** `server/src/force-rerun-triggers.test.ts:247`

`{ rel: 'src/lib/account-defaults.ts', base: REPO_ROOT }` restores a source-extension negative case. Verified empirically rather than by reading:

- **As shipped**, `src/lib/account-defaults.ts` matches **none** of the 15 real main triggers, at `/repo`, at `/repo/.claude/worktrees/wt`, and at this checkout's real root — so the `does not match` case is green today for the right reason, not because the path is unresolvable. `absPathUnder`'s `existsSync` assertion at `:98` is satisfied (the file is real, 5001 bytes), so this is not a vacuous pass.
- **With pass 1's widening applied** — `server/vitest.config.ts:276` replaced by `'{**/*.ts,**/.*/**/*.ts}'` — it matches at **all three** root shapes, so `expect(matchesTrigger(...)).toBe(false)` fails at every shape. The break pass 1 named is now caught.

The comment at `:240-243` was updated to match what the set now does.

---

**Finding 3 — VERIFIED RESOLVED.** `server/src/force-rerun-triggers.test.ts:301-304`

The pinned literal is byte-identical to `server/vitest.config.ts:235` (compared mechanically, not by eye), so deleting that config entry now reddens this test. Confirmed the reason the pin was needed still holds: `{**/server/src/**,...}` does match `/repo/server/src/tts/synthesise-chapter.ts`, so the file-coverage case at `:164` cannot catch the deletion on its own.

Also confirmed the pin describes reality rather than a stale path — `coqui-residency-policy.guard.test.ts:22` computes `SYNTHESISE_CHAPTER_PATH` as `join(process.cwd(), 'src', 'tts', 'synthesise-chapter.ts')` and reads it with `readFileSync` at `:49`, so the runtime-read gap the trigger exists to close is genuine and the path is the one pinned.

The literal is hand-typed on both sides rather than derived from a guard constant, unlike the other four pins — but `coqui-residency-policy.guard.test.ts` deliberately was not converted in this PR, so there is no constant to derive from. The two literals must agree or the test is red, so they cannot drift apart; the residual gap is that neither is tied to the guard's own `SYNTHESISE_CHAPTER_PATH`. That is the declared follow-up, not a new defect here.

### New findings

None beyond the false claims folded into finding 1 above. I re-read both files `b6dfb8f9` touched at normal depth: the constant hoisting is value-for-value identical to the inlined `join(...)` calls it replaced, so `externalFilesFloor()`'s behaviour is unchanged; the `externalRoots` ↔ `GLOBS[1..3]` index correspondence is correct in order (`scripts`, `server/tts-sidecar/scripts`, `pinokio-scripts/lib`) and a reorder of the exported array would fail loudly; moving the `SRC_ROOT` assertion below the new block only changes which failure reports first. The new `MAIN_NOT_COVERED` entry couples the server suite to a frontend file's existence via `existsSync`, which is the same cross-tree coupling `CONTRIBUTING.md` and `apps/android/pubspec.yaml` already have in that array, and it fails closed.

### Still open from pass 1's cleanup nits

Noted for the record, not re-argued — none blocks merge: nit 4 (the PR title and `server/vitest.config.ts:263` describe the mechanism as generation / "#3085's chosen design, option 2" when nothing is generated), nit 5 (no explicit release-notes not-applicable in the PR body — CLAUDE.md before-shipping step 5 wants it said, not omitted), and nit 6 (the declared `coqui-residency-policy.guard.test.ts` follow-up still cites no issue number) are all unchanged at this head.

### Verdict

**Changes requested.** Findings 2 and 3 are genuinely closed and were verified by replaying the breaks, not by reading the diff. Finding 1 is not: the hoisting closed the modification case and left the addition case — the one pass 1 filed, and the one the function's own history says is next — fully undetected, while adding two comments and a commit message that state it is covered. One correctness bug open, so re-review is triggered once it is fixed and pushed.

dudarenok-maker:
## PR review — pass 3 (head 6db98b5444cbcec9c3f4a78d777db680c547337b, depth low)

**Scope.** Final allowed automatic round under the gate's 3-pass cap. Finding 1 was settled **empirically**, not by reading the diff — three separate throwaway mutations run against the real suite, each reverted. Findings 2 and 3 were re-confirmed unaffected (`6db98b54` touches exactly one file, `server/src/spawn-windows-hide.test.ts`, +21/−34). Plus a runtime-equivalence check of the refactor against its own parent commit. All 21 CI checks pass at this head. Tree verified clean before and after: `HEAD` unchanged at `6db98b54`, `git status --porcelain` empty, and the file's SHA256 byte-identical to the backup taken before the first mutation.

**Finding 1 — VERIFIED RESOLVED.** `server/src/spawn-windows-hide.test.ts:161,269,507`

The fix is the shape pass 2 prescribed, and it holds. `EXTERNAL_SCAN_ROOTS` is declared once at `:161` and has exactly three references in the whole tree — the declaration, the guard's consumption at `:269`, and the test's assertion at `:507`:

```
:161  const EXTERNAL_SCAN_ROOTS = [ ... ] as const;
:269  const externalCandidates = EXTERNAL_SCAN_ROOTS.flatMap((root) =>
:507  const externalRootsRel = EXTERNAL_SCAN_ROOTS.map((r) => toRepoRel(r.dir));
```

Same module, same binding — not a copy and not a re-declaration. `git grep` confirms the three removed constants (`EXTERNAL_SCRIPTS_DIR` / `EXTERNAL_TTS_DIR` / `EXTERNAL_PINOKIO_DIR`) have no remaining references anywhere.

**Mutation A — the exact break pass 1 filed and pass 2 found still live: adding a fourth root.** The fix agent's own report said their mutation failed via "ENOENT during import", which would mean the protection was an accident of using a non-existent path rather than the assertion. It is not. Added a fourth entry pointing at a **real, existing** directory — `{ dir: join(REPO_ROOT, 'public'), extensions: ['.mjs'] }` — so no filesystem error is possible. Result:

```
FAIL  scan scope matches the declared SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS (#3085)
AssertionError: expected [ 'scripts', …(3) ] to deeply equal [ 'scripts', …(2) ]
  [
    "scripts",
    "server/tts-sidecar/scripts",
    "pinokio-scripts/lib",
+   "public",
  ]
 ❯ src/spawn-windows-hide.test.ts:514:30
```

A clean `toEqual` length/content mismatch at the assertion — not a crash, not an import error. The protection mechanism is the one the code claims. 22 passed, 1 failed, and the one that failed is the scope test.

**Mutation B — the consumption proof.** A `toEqual` on an array the guard doesn't actually read would still pass mutation A, so that alone does not prove `externalFilesFloor()` consumes this array rather than merely sharing a name with it. Removed the `scripts` entry from `EXTERNAL_SCAN_ROOTS` and re-ran:

```
× scan scope matches the declared SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS (#3085)
× EXTERNAL_FILES_FLOOR from function covers the old hardcoded list
AssertionError: old hardcoded entries missing from EXTERNAL_FILES_FLOOR (coverage regression):
scripts/wt-new.mjs: expected [ 'scripts/start-app-prod.mjs', …(40) ] to deeply equal []
```

41 real files dropped out of the floor. The array genuinely drives the scan; the declaration the test asserts against and the list the guard walks are the same object.

**Mutation C — runtime equivalence.** Instrumented both this head and its parent `b6dfb8f9` identically to dump the full computed `EXTERNAL_FILES_FLOOR` (repo-relative, sorted) to a temp file, ran each, and hashed the results:

```
HEAD-HASH:   0C2129C93891F456FA63286E64D3EFB73AE23BF62300E940BEEA54CC355E9830
PARENT-HASH: 0C2129C93891F456FA63286E64D3EFB73AE23BF62300E940BEEA54CC355E9830
IDENTICAL: True   (93 entries, 0 diff lines)
```

The three separate `listFilesRecursive` calls and the `flatMap` produce the same 93-file floor. Ordering is equivalent by construction too — `[a,b,c].flatMap(f)` is `[...f(a), ...f(b), ...f(c)]`, and the array preserves the original `scripts` → `tts-sidecar/scripts` → `pinokio-scripts/lib` order with identical extension lists. No behaviour change.

**The loop #3085 exists to close is now genuinely closed for the addition case.** Adding a fourth scanned root forces this chain, with a red test at every step it is skipped: `EXTERNAL_SCAN_ROOTS` → `toEqual` at `:513` → `SPAWN_WINDOWS_HIDE_GUARD_SCAN_GLOBS` → `force-rerun-triggers.test.ts` (which derives coverage cases from indices `[1][2][3][4][6]` and pins exact config entries for `[1][4][6]`) → a real `forceRerunTriggers` entry. I also checked the evasion: *appending* the new glob at index 7 instead of inserting at 4 does not escape, because `expectedExternalRootsRel` reads `[1..3]` and stays 3 long against 4 actual roots; inserting at 4 instead reddens the `EXTERNAL_FILES_MANUAL` assertion at `:518`. Both directions fail loud.

**The three claims pass 2 called false are now true as written** — the comments at `:155-160` and `:503-506` describe the mechanism this pass just demonstrated.

**Findings 2 and 3 — still resolved.** `6db98b54` touches only `spawn-windows-hide.test.ts`; `force-rerun-triggers.test.ts` and `server/vitest.config.ts` are byte-unchanged since pass 2 verified them. Re-ran all five in-scope files at this head: **124/124 pass** (`spawn-windows-hide`, `force-rerun-triggers`, `cast-lock.guard`, `state-language.guard`, `engine-language-coverage.guard`).

### Cleanup nits

**7 — `6db98b54` left a dangling cross-reference.** `server/src/spawn-windows-hide.test.ts:251`

The commit deleted the block comment inside `externalFilesFloor()` that explained why `scripts/tests/` is scanned. The function's header still points at it:

```
:248   - scripts/ recursive, .mjs/.cjs/.js, INCLUDING scripts/tests/ (previously
:249     excluded — that exclusion let an unguarded module-load-time pwsh spawn
:250     in cross-os-ffmpeg-install.test.mjs pop a visible window on every
:251     `npm run test:hooks` run; see externalFilesFloor()'s scripts/ comment)
```

There is no longer a `scripts/` comment in `externalFilesFloor()` — the only comments left in it are `:259`, `:267-268`, `:273`, `:277`. No knowledge was lost (the rationale survives in full at both `:22-29` and `:248-251` itself), so this is a dangling pointer, not a dropped explanation. Fix is to end `:251` at `run)`. Under CLAUDE.md's incidental-findings rule this is a chore the change made owed — cheap to fold before merge.

**8 — the commit message names the wrong protection mechanism.** `6db98b54`

Its body states "Adding a 4th root fails immediately (directory not found during import)". That is what happens only if the added path does not exist; it describes a filesystem accident, not the guard. Mutation A above shows the real mechanism is the `toEqual` array comparison at `:513`, which fires on a perfectly valid directory. The code is right and the message is wrong — recorded here rather than filed, since the body is already in history and the shipped comments at `:155-160` / `:503-506` describe it correctly.

### Still open from passes 1–2

Unchanged at this head, re-checked rather than assumed; none blocks merge:

- **Nit 4** — the PR title says "generate forceRerunTriggers from …" and `server/vitest.config.ts:263` says "#3085's chosen design, option 2", but nothing is generated: the trigger is a hand-typed literal that a test checks against the constants. Sound mechanism, wrong name.
- **Nit 5** — still no explicit release-notes not-applicable in the PR body. CI-only, so N/A is right; CLAUDE.md before-shipping step 5 wants it *said*.
- **Nit 6** — the declared `coqui-residency-policy.guard.test.ts` follow-up still cites no issue number. I searched open and closed issues; the only related ones are #3085, #3135, #3136, #3137, none of which is this follow-up.

### Considered and not filed

- **A future editor could bypass the array entirely** by adding a bare `listFilesRecursive(...)` call next to the `flatMap`. True, but there is no longer a three-call pattern to copy — the single `flatMap` over a named array, plus the comment at `:267-268`, is the convention now, and this is the shape pass 2 asked for. Not a demonstrable break.
- **The `REPO_ROOT` non-recursive `.mjs`/`.ts` scan at `:260-265` is outside `EXTERNAL_SCAN_ROOTS`** and so outside the exhaustive check — widening it would not redden anything. Pre-existing, predates #3085, and disclosed in `spawn-windows-hide.guard-targets.ts:24-33` rather than implied closed. Pass 1 weighed and did not file it; unchanged here.

### Verdict

**No correctness bugs. Clean to merge.** Finding 1 — open since pass 1 and the sole blocker after pass 2 — is verified closed by direct experiment rather than by inspection: the fourth-root break fails via the assertion, on a valid directory, and the array demonstrably drives the real scan. Findings 2 and 3 remain closed. Runtime output is byte-identical to pre-refactor.

Only cleanup nits remain (7 and 8 new, 4–6 carried), which per the gate's triage rule do not trigger another round. Nit 7 is a stale comment this commit made false, so it is owed in this PR under the incidental-findings rule; the rest are PR-body/naming hygiene. This was the third and final automatic pass — anything further goes to the repo owner rather than another fix-and-review cycle, and on this head nothing needs to.

dudarenok-maker:
**Merge coordination note from #3191 (PR just closed):**

#3191 and this PR (#3151) both modify  in a hard-conflicting file. That PR deleted a duplicate literal assertion at line 302 () as cleanup; this PR may reinstate it.

Whichever of the two merges second should delete the now-redundant assertion in the other's merged commit before completing the merge (or immediately after via a follow-up commit if it's already merged). The test suite will help catch this — if both assertions exist, the it.each loop will fail with a duplicate-entry error.

Issue #2292, opened 2026-08-12
chore(server): nine more sites still swallow or downgrade a lock-acquisition timeout


dudarenok-maker:
## What

#2260 (PR #2284) gave `withKeyLock`'s acquisition expiry a typed shape (`LockAcquisitionTimeoutError` / `isLockAcquisitionTimeout`) and made **seven** best-effort `catch` blocks rethrow it instead of swallowing it — the cast-identity family, where swallowing produced a 200 with `cast.json` written and the retirement never recorded.

**Nine further sites still swallow or downgrade a lock-acquisition timeout.** They were found by that PR's implementation pass, deliberately left alone because each needs a behaviour decision rather than an obvious fix, and are filed here rather than folded so the diff stayed scoped.

## The one that matters most

**`server/src/routes/cast-reject-orphan.ts:435`** — `catch (forgetErr)` around `forgetSupersededId`, marked "non-fatal". The route falls through and still returns `res.json({...})` **success**.

This is the closest remaining match to the bug #2260 was filed against: a timeout here silently reports a completed reject with the stash cleanup skipped. The genuine question is whether it *should* fail the reject, given the pair is already durably recorded by the time this runs — which is exactly why it was not fixed inline.

## The interim-snapshot family

Three `catch (persistErr)` blocks around interim `cast.json` writes (`castBase.writeChecked` → `withCastLock`):

- `server/src/routes/analysis.ts:4036` — main route, interim write
- `server/src/routes/analysis.ts:4249` — main route, stage1 write
- `server/src/routes/analysis.ts:6151` — subset route, interim write

These snapshots are optional by design: a final write follows. Rethrowing would fail a whole analysis run because a *progress snapshot* could not be taken. Two defensible answers, so it fails the fix-now bar — but the current behaviour means a run can lose every interim checkpoint to contention and report nothing.

## The report-as-per-item-failure family

Five routes surface a lock timeout as an ordinary per-item failure inside a 200/207 body rather than escalating it. Reasonable for a genuine per-item error; questionable for a lock timeout, which says nothing about that item:

- `server/src/routes/script-review.ts:996`
- `server/src/routes/cast-design.ts:613`
- `server/src/routes/voice-style.ts:172`
- `server/src/routes/cast-series-patch.ts:178`
- `server/src/routes/voice-override-linked.ts:164`

## Also worth deciding: `cast-merge.ts`'s 500 is HTML, not JSON

[...]

Issue #1057, opened 2026-06-24
Design full cast: picking a scope (esp. Emotion variants) races 0→100% instantly and designs nothing


dudarenok-maker:
## Symptom
On the Cast view, **Design full cast → "What should I design?"** popover shows e.g. *Base voices: all done · Emotion variants: 35 ready now · Both: 35 tasks*. Picking a scope makes the progress pill **jump from 0% to 100% in ~1 second with nothing generated** — no voices/variants designed.

Repro screenshot: cast view for "Keeper of the Lost Cities", chips show Generated 13 / Carried 6 / Has variants 17 / Needs variants 23; Emotion variants reports 35 ready.

## Investigation so far (NOT yet root-caused / fixed)
Two candidate mechanisms, both consistent with "instant 100%, nothing happens":

1. **POST carries no actionable work → server immediately idles.** In `server/src/routes/cast-design.ts`, `hasWork = (characterIds?.length>0) || (scope!=='bases' && variantTasks.length>0)`. If the frontend doesn't send a populated `variantTasks` (or the server's strict per-task validation — `emotions.every(e ∈ VARIANT_EMOTIONS)` — drops them), `hasWork` is false, the POST is treated as a *subscribe*, and the server sends `{type:'idle', total:0}` at once. Need to confirm what `cast-design-stream-middleware` actually POSTs for the `variants`/`both` scopes and whether the emotion strings match `VARIANT_EMOTIONS`.

2. **All variant tasks silently skipped in the loop.** The design loop skips a variant when `!character.overrideTtsVoices?.qwen?.name` or the variant already exists. Client `hasBase` (`!!qwen.name`) *matches* the server gate, so this shouldn't fire for the "ready" set — UNLESS the on-disk cast.json differs from the frontend's view (e.g. carried/reused voices where the qwen base name isn't denormalised onto this book's cast row — cf. "reused voice lives on the matched Voice, not the character"). Also note: for the **local** persona engine, a variant whose character lacks a `voiceStyle` is `continue`d **without** incrementing `done` or `skipped` and **without** any broadcast (`cast-design.ts` ~line 315) — a silent drop that can desync the pill.

## Next steps
- Capture the actual POST body for scope=variants (devtools / server log) and confirm `variantTasks` + emotion enum.
- Confirm whether the affected characters have `overrideTtsVoices.qwen.name` on disk vs. only a carried/matched voice.
- Reproduce on the GPU box (this path needs the Qwen sidecar).

## Notes
[...]

PR #3161, opened 2026-09-10
fix(frontend): surface orphaned design requests and stop Gemini persona failures halting bulk jobs

### LINKED ISSUE #3027: Design full cast can halt silently mid-run, and an orphaned concurrent single-design request never resolves

dudarenok-maker:
## Summary

Fixes both halves of #3027. **Half 1:** the cast-design stream middleware silently dropped a design request (single or bulk) when another job already owned the stream's `handle` — it now dispatches a `pushToast` error so the drop is visible. **Half 2:** a throw from the Gemini persona fallback (`generateVoiceStylePersona`, non-local engine) escaped the per-character try/catch and landed in the route backstop as a job-wide halt with no designed/failed/skipped summary; it's now wrapped in its own try/catch that routes the failure through the same `job.failures` push + `character_failed` broadcast pattern the ride-out loop already uses, so one persona failure no longer aborts the remaining characters. The LOCAL-engine persona pre-pass (`runPersonaPrePass`) already handled this correctly per-character and needed no change.

Closes #3027

## Test plan

- [x] `src/store/cast-design-stream-middleware.test.tsx` — 17/17 passing, including regression tests for the busy-toast path (single-while-busy, bulk-while-busy) and a "first failure vs. all-failures-share-a-reason" toast-wording test (N4, added in review pass 2) that pins which arm of the ternary is used when failures differ — verified red against a swapped-arms mutation and green against current code.
- [x] `server/src/routes/cast-design.test.ts` — 41/41 passing, including a regression test (`#3027/2`) that fails pre-fix and passes post-fix (re-verified the same way: reverting only the fix file turns exactly that one test red).

Found by the pr-review-gate re-review pass(es) on this PR:

- **Cast-lock message leak** (`b8b12be5`) — the Gemini persona-write catch was surfacing a raw lock-timeout error message (which embeds an absolute workspace path) instead of the curated one.
- **Toast didn't show the failure reason** (`48a1a7ab`) — the bulk cast-design summary toast collapsed multiple failures into a bare count; now surfaces the first failure's actual error text (and says once when every failure shares it).
- **Stale re-entrancy comment + weak test assertions** (`f8256049`) — a comment describing the old silent-no-op behavior, and busy-toast tests that weren't pinning the actual toast content.
- **Release-notes / on-box register bookkeeping** (folded into `914cc2e8`) — recorded this fix in `docs/release-notes-next.md`, `RELEASE_NOTES.md`, and the on-box acceptance register.
- **`ensureCharacterVoiceUuid` per-character-catch bug** (`f3d3a341`) — a call outside the per-character try/catch could still abort the whole ride-out loop; moved inside it.
- **Stale comments elsewhere describing the old silent-no-op behavior** (this round) — `src/modals/voice-readiness-gate.tsx` and `src/components/voice-nudge-toast.tsx` still described a second design start as silently no-oping; corrected to describe the actual busy-toast behavior.
- **Duplicated busy-toast string** (this round) — the "already running" message literal was copy-pasted at two call sites in `cast-design-stream-middleware.ts`; extracted into a shared `DESIGN_ALREADY_RUNNING_MESSAGE` constant.

https://claude.ai/code/session_01HHmS2UNg7pN6rp7F7tfEqD


dudarenok-maker:
## PR review — pass 1 (head ffe43ce1d713b8fc423e1416b82e7dee639f7675, depth high)

**Scope:** 4 files — `server/src/routes/cast-design.ts`, `server/src/routes/cast-design.test.ts`, `src/store/cast-design-stream-middleware.ts`, `src/store/cast-design-stream-middleware.test.tsx`. Cross-cutting (frontend Redux middleware + server SSE route), so reviewed at `high` rather than the per-commit single-scope tier.

Verified locally in a detached scratch worktree at the head SHA (not the primary checkout): `server` 39/39 green, frontend middleware 17/17 green. Reverting **only** `cast-design.ts` to `fd151582` turns exactly `#3027/2` red, on the `character_failed`-for-`hart` assertion — red for the reason claimed. Reverting **only** `cast-design-stream-middleware.ts` to `80354c75` turns exactly the two new frontend tests red. Both paired tests are real regression tests.

---

### 🔴 Blocking

**1. [correctness bug] The new persona catch bypasses `itemFailureReason`, so a cast-lock timeout now reaches the client verbatim — absolute workspace path included.**
`server/src/routes/cast-design.ts:533`

```ts
const message = (e as Error).message || 'Persona generation failed.';
job.failures.push({ characterId, name: character.name ?? characterId, error: message });
broadcast(job, { type: 'character_failed', characterId, name: ..., errorReason: message });
```

`writeVoiceStylePersona` (same file, `:189`) is `withCastLock(bookDir, …)` → `withKeyLock(castJsonPath(bookDir), …)`, which throws `LockAcquisitionTimeoutError` after 10s. Its message embeds the lock key, and a cast key **is** the absolute path to `cast.json`. The new `try` now spans that write, and its `catch` ships the raw message to two client-facing surfaces (`character_failed.errorReason`, `idle.failures[].error`) plus `GET …/cast/design/status`'s JSON body (`:942`).

This is a regression, not a pre-existing gap: **before this commit that throw escaped the loop into the route backstop at `:900`, which curates it** — `isLockAcquisitionTimeout(e)` → `requestFailureMessage(…)`. The fix moved the path from a curated seam to an uncurated one.

It also contradicts the block comment directly above it, which claims to "Route through the SAME per-character-failure shape the inner ride loop uses (lines 688-696)". That loop's catch is at `:714` and does `itemFailureReason(e, message)`. The helper is already imported at `:42` and goes unused here.

Concrete, reproduced. Driving a `LockAcquisitionTimeoutError` out of the first cast-lock acquisition of a `['hart','aria']` run (a spy on `withCastLock`, `hart` being the persona-less character) at this head SHA:

```
errorReason   => "withKeyLock: timed out after 10000ms waiting to acquire
                  \"C:\\…\\books\\Della Renwick\\The Hollow Tide\\The Hollow Tide\\cast.json\" --
                  either a cast-lock.ts rule 1 … or rule 4 … violation, OR ordinary contention …"
idle.failures => [{"characterId":"hart","name":"Hart","error":"withKeyLock: timed out after 10000ms …"}]
```

`file-lock.ts:175`'s own header states why this is forbidden: *"every key this class can carry embeds an absolute workspace path … This app is served over LAN HTTPS by design, so returning that verbatim hands any paired phone the filesystem layout of the user's library plus the internal locking vocabulary."* CLAUDE.md states it as an absolute: *"No client-facing failure ever carries a lock timeout's own message — not an escalating body, not a per-item reason, and not an SSE `error` event."*

The sibling site already has a test that would catch this shape — `cast-design.test.ts:1615`, `expect(idle?.failures?.[0].error).not.toContain('withKeyLock')`. The new site has no equivalent, which is why CI is green.

Fix: `const message = itemFailureReason(e, (e as Error).message || 'Persona generation failed.');`, plus a test mirroring `describe('cast-design — per-character reason on a lock timeout (#2292)')` for the persona site.

---

### 🟠 Significant

**2. [correctness bug] Routing the persona failure per-character removes the only surface that showed *why* it failed — for the exact configuration that produced #3027.**
`server/src/routes/cast-design.ts:534-541`, with `src/store/cast-design-stream-middleware.ts:100-134`

`failures[].error` / `character_failed.errorReason` are **never rendered anywhere in the frontend.** `onCharacterFailed` dispatches `charFailed`, which only appends to `castDesign.active.failures`; the sole consumer of that array is `layout.tsx:1481`'s `failureCount: failures.length`. `git grep` over `src/**/*.tsx` finds no render of a failure's `error` string. The terminal summary is a count: `"Designed ${done}"` + `"${failed} failed"` (`cast-design-stream-middleware.ts:115-117`).

So, for the configuration #3027 itself names as the leading trigger — `server/.env` with no `GEMINI_API_KEY`, 12-character cast:

- **Before this commit:** the first persona call throws at `voice-style.ts:200`, escapes the loop, backstop emits `{type:'error', message}`, middleware `onError` toasts it. The user reads *"GEMINI_API_KEY is required to generate voice-style personas. Set it from Account → Server configuration → Gemini API key, or in server/.env for CI / power users."* — immediately, and actionable. #3027's own step-3 note records exactly this: *"failed immediately with a clean, structured SSE error."*
- **After this commit:** 12 identical per-character failures, none rendered, and one toast reading **"Designed 0 · 12 failed."** The cause is on the wire and nowhere on screen.

The issue asked for the failure to be *"caught and surfaced as a per-character failure in the terminal summary (designed/failed/skipped)"*. A count is not a surfacing — the PR makes a loud, correctly-diagnosed failure quiet, which is the same class of defect #3027 was filed about.

Secondary, same site: the file's own doctrine for a class that fails every remaining character identically is to stop, not grind — `:672-681` halts with `code:'sidecar_unavailable'` precisely *"instead of grinding through N timeouts."* A missing/exhausted Gemini key is that class (the PR's own fixture calls it `'Gemini quota exceeded'`, and a quota failure goes through `geminiRateLimiter.acquire` on every one of the N retries).

If the remedy chosen is "render the reasons" vs "halt on a job-wide persona-config class" vs "both", that is a decision with more than one defensible outcome and may belong in the design-pass carve-out — **the decision owed is: where does a per-character design failure's reason become visible to the user, and does a persona-config failure halt the run or ride out?** What is *not* a judgement call is that the PR currently leaves the cause invisible.

---

### 🟡 Minor

**3. [cleanup] The middleware's module header still documents the retired behaviour.**
`src/store/cast-design-stream-middleware.ts:13-15`

> *"Re-entrancy: one open `handle` at a time … A second start while one runs **is ignored** — the Cast view also disables the button, so this is belt-and-braces."*

This change is precisely that a second start is no longer ignored. A stale comment the change made false is a chore under CLAUDE.md's incidental-findings rule, not taste.

**4. [cleanup] The pre-existing re-entrancy test keeps the retired name, and the new bulk test duplicates it rather than extending it.**
`src/store/cast-design-stream-middleware.test.tsx:209` and `:577`

`it('re-entrancy: a second designAllRequested while one runs is ignored')` asserts only `expect(startCalls).toHaveLength(1)`. The new `it('dispatches an error toast when a bulk design request arrives while another job is running')` asserts the same `startCalls` length plus the toast. Two tests now cover one path, one of them named for semantics the PR retired. Also `:7` of the same file's header ("guarding against a second concurrent run (re-entrancy)").

**5. [cleanup] Both new frontend assertions are weaker than their own test names, and weaker than this file's established bar.**
`src/store/cast-design-stream-middleware.test.tsx:573-574` and `:609-610`

```ts
const toastActions = recorded.filter((a) => a.type === 'notifications/pushToast');
expect(toastActions.length).toBeGreaterThan(0);
```

The tests are titled "dispatches an **error** toast" but check neither `kind`, nor `message`, nor `dedupeKey`, nor that the toast reached `store.getState().notifications.toasts`. They count `pushToast` actions recorded anywhere in the test, from any source. Concretely: if `designAllRequested` ever grows a toast of its own on `begin` (a "Designing N characters…" info toast, say), both tests pass with the fix fully reverted — they would stop being able to fail. This file already has the stronger pattern: `[C-6]` at `:245` renders the real `<ToastStack>` off the same store to prove the text reaches the DOM. Assert `kind: 'error'` and the message at minimum.

**6. [cleanup — merge gate] The release-notes pair is missing, with no explicit not-applicable.**
`docs/release-notes-next.md`, `RELEASE_NOTES.md`

Before-shipping step 5 requires both, PR-by-PR, or an explicit statement that the change has no shippable delta. This PR has a clear user-visible delta (a new error toast; a bulk run that no longer dies with a bare "Halted"), touches neither file, and the PR body says nothing about them. The open `# Castwright 1.15.0` section in `RELEASE_NOTES.md` already carries entries for smaller, internal-only items, so the bar is not the obstacle here.

**7. [cleanup — merge gate] On-box acceptance rows A6 and A7 name #3027 as their live blocker; this PR closes it, and the register is untouched.**
`docs/testing/onbox-acceptance-register.md:1998-2030`

A7's `*Needs:*` line reads *"Blocked on the same halt as A6 — [#3027]"*, and both rows' 2026-09-06 PARTIAL notes end *"Filed as [#3027]"*. On merge both rows point at a closed issue as their current blocker. Separately, the behaviour this PR ships (a bulk run surviving a real Gemini persona failure; the cross-book single-design refusal) is the "real analyzer, real book" class that Before-shipping step 3 converts into a recorded row. Recording blocks the merge even though running does not — and all three surfaces move together (register, run sheet, live view via its recorded `url`).

---

### ✅ Checked and clean

- **Paired tests are real.** Both halves verified red-before / green-after by reverting each fix file alone at the head SHA, and red for the claimed reason.
- **`Closes #3027`** present, on its own line, outside any code span.
- **`cast.json` locking rules 1–4** — `writeVoiceStylePersona` is unchanged and still the innermost read-through-write under `withCastLock`; the new `try` sits outside the lock, so no nesting is introduced. Only the *curation* of its timeout is wrong (finding 1).
- **No new config knob, no derived artifact staled** — `openapi.yaml` untouched (these SSE events are deliberately unspecced per plan 195), no `api-types.ts` / BACKLOG / brand-PNG impact. `bug`-labelled issue, so no `docs/BACKLOG.md` row is owed.
- **Control-flow of the new `continue`** — inside `try { … } finally { clearInterval(heartbeat) }`, so the heartbeat is cleared; `job.done`/`job.skipped` accounting matches the ride loop's `break`; a persona-failed base leaves its variants to the existing `!baseName` skip at `:475`, so scope `'both'` records one failure and not one per variant.
- **No stale snapshot on a dropped frontend request** — `designSingleRequested`/`designAllRequested` are no-op reducers, and `beginSingle`/`begin` are dispatched after the `handle` check, so a refused request leaves no spinner. The drawer's `sliceDesigning` reads only the slice.
- **Toast dedupe is not permanent** — `notifications-slice.ts:78-99` bumps `createdAt` on a live toast only; a later drop after auto-dismiss re-toasts. No silent-second-drop.
- **Test-mock hygiene** — the new server test's `personaMock.mockReset()` drops the `beforeEach` default, but `aria` carries a `voiceStyle` so the fallback is never reached for it; contained to this test.
- **Plan 195 is not staled by the server half** — `docs/features/195-design-full-cast.md:39` already reads *"Per-character failures are recorded and the loop continues"*, which is what this commit finally makes true.

---

### Verdict

**Do not merge as-is.** Finding 1 is a proven, this-diff regression against a named CLAUDE.md invariant and is a one-line fix plus a test. Finding 2 means the PR does not yet deliver #3027's stated acceptance — the failure is surfaced on the wire but not to the user. Findings 3–7 are chores this work made owed and are fixed in the same round per the incidental-findings rule.

2 correctness findings (1 🔴, 1 🟠) — this pass re-triggers a review round once they are fixed and pushed.

dudarenok-maker:
## PR review — pass 2 (head 914cc2e8943e47efbe6ccddd70be76269b2a307b, depth high)

Delta-only re-review. Prior pass: `ffe43ce1…`, 7 findings. Depth re-derived from the current commit set — still multi-scope frontend + server, so `high`, unchanged.

Verified in a detached scratch worktree at this head (not the primary checkout, and not the branch's own worktree). Merge-base against `origin/main` is `2b63b451`, i.e. exactly `main`'s tip — the rebase is current and the diff is 7 files, additive on every doc file (`RELEASE_NOTES.md` 1+/0-, `docs/release-notes-next.md` 1+/0-, `docs/testing/onbox-acceptance-register.md` 16+/0-). No conflict markers; the two new register blockquotes appear exactly once each. **The rebase dropped and duplicated nothing.** `server/src/routes/cast-design.test.ts` 40/40 green, `src/store/cast-design-stream-middleware.test.tsx` 16/16 green (was 17 — the duplicate re-entrancy test was removed, two busy-toast tests added). All 22 CI checks green.

---

### Disposition of the 7 prior findings

**1. 🔴 → ✅ VERIFIED RESOLVED.** `server/src/routes/cast-design.ts:534` now reads `const reason = itemFailureReason(e, message);`, and the curated `reason` — not the raw message — is what reaches both `job.failures[].error` (`:535`, which is also what `GET …/cast/design/status` serves at `:943`) and the `character_failed` broadcast's `errorReason` (`:540`). Re-proved independently, the same way pass 1 did: with the fix mutated back to `const reason = message`, the new test at `cast-design.test.ts:1632` goes red on

```
Expected: "Timed out waiting for another operation on this book to release its file lock — that is
           contention, not a problem with this item. Retry once the other operation has finished."
Received: "withKeyLock: timed out after 10000ms waiting to acquire "cast:/w/hollow-tide" -- either a
           cast-lock.ts rule 1 … violation, OR ordinary contention …"
```

— exactly one test, red for exactly the claimed reason, green with the fix. The test drives a real `LockAcquisitionTimeoutError` through a `withCastLock` spy and asserts `LOCK_CONTENTION_ITEM_REASON` on **both** surfaces plus `not.toContain('withKeyLock')` / the lock key. Real regression test.

**2. 🟠 → ✅ VERIFIED RESOLVED.** `src/store/cast-design-stream-middleware.ts:137-141` appends the failure reason to the summary toast. Confirmed against the actual #3027 configuration: `generateViaGemini` (`server/src/analyzer/voice-style.ts:200-204`) throws a byte-identical message for every character when the key is missing, so `sameReason` is `true` and a 12-character run toasts *"Designed 0 · 12 failed. GEMINI_API_KEY is required to generate voice-style personas. Set it from Account → Server configuration → Gemini API key, …"* — the actionable string #3027 asked for, not a bare count. Mutating the whole `if (failed > 0)` block out turns `cast-design-stream-middleware.test.tsx:206` red on `expected 'Designed 1 · 1 failed.' to contain 'no gemini key'`. Red-before / green-after confirmed. The zero-failure case is safe (`failed = failures.length`, so `failures[0]` is only read when it exists).

**3. 🟡 → ✅ VERIFIED RESOLVED** in the file it was raised against (`cast-design-stream-middleware.ts:13-17` now says "dropped and surfaces an error toast"). **The class is not resolved — see N3 below.**

**4. 🟡 → ✅ VERIFIED RESOLVED.** The duplicate `re-entrancy: … is ignored` test is gone; its `startCalls` assertion lives on inside the new bulk test. The test-file header at `:7` ("guarding against a second concurrent run (re-entrancy)") was left as-is, which is correct — the guard still exists, only its feedback changed, so that sentence was never made false.

**5. 🟡 → ✅ VERIFIED RESOLVED.** Both new tests now assert `toHaveLength(1)` + `payload.kind === 'error'` + `message` match, replacing `toBeGreaterThan(0)`. Confirmed they can fail: mutating both `if (handle) { dispatch(…) }` blocks back to `if (handle) return result` turns exactly those two red (`expected [] to have a length of 1`).

**6. 🟡 → ✅ VERIFIED RESOLVED.** Both release-notes files moved. `docs/release-notes-next.md` entry is appended to the in-progress section with `(#3027, PR #3161)`; `RELEASE_NOTES.md` carries a matching brand-voice line in the open `# Castwright 1.15.0` section.

**7. 🟡 → ⚠️ PARTIALLY RESOLVED.** The register rows moved (A6 `:2019`, A7 `:2062`), correctly as dated blockquotes that record the code fix and keep the debt open rather than deleting the rows. `npm run check:onbox-register` and `npm run register:build --check` both pass. **But only one of the three surfaces moved — see N2 below.**

---

### New findings

Four of these are on surface pass 1 never saw (the four fix commits + the rebase). N1 is on pre-existing code the fix commits' own reasoning points at.

#### 🟠 Significant

**N1. [correctness bug] The fix closed the persona instance of "one character's lock contention kills the whole run" and left the sibling instance armed — in the call the ride-loop's own comment claims is already covered.**
`server/src/routes/cast-design.ts:575` (the call), `:706-714` (the comment that says it is handled)

The ride-out loop's per-character catch carries this:

```ts
/* Per-character synthesis failure — record it and move on.
   #2292 (owner decision) — a `LockAcquisitionTimeoutError` out of
   the persist steps in this try (`applyOverrideToCastFiles`,
   `persistEmotionVariant`, `ensureCharacterVoiceUuid`,
   `writeVoiceStylePersona`) keeps this per-character shape — one
   contended character must not fail the other N — …
```

Two of those four calls are **not in that try**. `ensureCharacterVoiceUuid` is at `:575`, three lines above `for (;;) { try {` at `:579-580`; `writeVoiceStylePersona` is at `:527`, now inside this PR's own new catch. So the comment names a handler for neither.

For `writeVoiceStylePersona` that is only misleading — this PR gave it a correct handler one level up. For `ensureCharacterVoiceUuid` it is a live behavioural gap, and it is reachable exactly the way finding #1 was: `ensureCharacterVoiceUuid` (`server/src/routes/qwen-voice.ts:258-298`) takes `withDesignLock` then `withCastLock` on `castJsonPath(bookDir)` for any character with no `voiceUuid`, so ordinary contention behind a long cast.json holder throws `LockAcquisitionTimeoutError` after 10s.

Reproduced at this head. Spying `ensureCharacterVoiceUuid` to throw a real `LockAcquisitionTimeoutError` on a two-character run (`['aria','brann']`) yields this SSE transcript:

```json
[ { "type": "progress", "characterId": "aria", "done": 0 },
  { "type": "error", "code": "lock-contention",
    "message": "Timed out waiting for another operation on this book to release its file lock — …" } ]
```

That is the whole job: no `character_failed`, no `idle`, no designed/failed/skipped summary, and `brann` never attempted. It is the same terminal shape #3027 was filed about — a bare halt — for a different call in the same loop. (The message itself is curated by the route backstop, so unlike finding #1 there is no disclosure leak; the defect is purely the halt.)

Not a design decision: #2292's owner decision already settled that this class stays per-character, and the comment above already asserts the intended behaviour — the code just doesn't implement it for this call. Choosing between "move the call inside the ride-loop try" and "give it its own catch beside the persona one" is two implementations of one agreed behaviour, which CLAUDE.md explicitly says is not a design pass. Yes, it is pre-existing; "it's pre-existing" is a void deferral reason, and this is the reviewer-brief shape *"one instance fixed, the class left armed"*. Fix it here, with a paired test mirroring the `#3027/2` one, and correct the `:706-714` comment so it names where each of the four calls is actually handled.

#### 🟡 Minor

**N2. [cleanup — merge gate] The register moved; the live view did not. Both mechanical checks pass because this is the register's own documented blind spot.**
`docs/testing/onbox-acceptance-register.md:2019`, `:2062` vs `docs/testing/onbox-acceptance-register-live-view.html:313`, `:320`

The live view mirrors each row's dated notes as `<div class="flag">` blocks — A6's and A7's 2026-09-06 PARTIAL notes are both there, verbatim in substance. The two new **2026-09-11** notes are not, and the HTML is untouched by this PR. Anyone reading the published page still sees A6/A7 blocked on an open #3027.

`check:onbox-register` passing is not evidence against this. The register's own "Live view" section states the gap outright:

> **A wording-only edit to a numbered Group row does not fail** … recording a run result, changing a hardware note … changes nothing either check compares for a Group row, since those are matched by ID. **The live view mirrors that prose in its own row bodies and will silently fall behind.**

And `check-onbox-register.mjs:246-252` confirms the prose is deliberately never compared markdown-against-HTML. The repo's own precedent is the other way: commit `9156bade`, which *added* the 2026-09-06 notes these two extend, moved both files in one commit.

Owed in this PR: the two matching `<div class="flag">` blocks in the HTML, then `npm run stamp:publish-token` (the register's "Any PR that changes the live view's RENDERED content must re-stamp it"), commit, and publish that file with the recorded `url`.

**N3. [cleanup] Finding #3's stale comment has two siblings the fix left behind — the same sentence, in the two components that dispatch these actions.**
`src/modals/voice-readiness-gate.tsx:52-54` and `src/components/voice-nudge-toast.tsx:3-6`

```ts
/* Mirrors cast.tsx's own guard: only one book's bulk design can run at a
   time (cast-design-stream-middleware.ts ignores a second start), so
   dispatching here while another book owns the run would silently no-op. */
```

The middleware no longer "ignores a second start" and dispatching no longer "silently no-op[s]" — it raises an error toast. The comment's stated *reason* for the early return at `:60-68` is now false (the real reason is now "would raise a second, confusing error toast"). `voice-nudge-toast.tsx:4-5` carries the same claim — "a tap can never silently no-op against the single-stream middleware". Both are stale comments this change made false, which CLAUDE.md's incidental-findings rule lists as a chore, and they are the same class finding #3 already fixed in the middleware's own header.

**N4. [cleanup] The `First failure:` half of the new toast ternary is completely unpinned — swapping the two branches leaves the whole suite green.**
`src/store/cast-design-stream-middleware.ts:140`

```ts
message += sameReason ? ` ${firstError}` : ` First failure: ${firstError}`;
```

`grep "First failure"` across `src/` and `e2e/` returns exactly one hit: the implementation. The only test that reaches this code has one failure, so it only ever exercises the `sameReason === true` branch. Verified: rewriting the line as `!sameReason ? … : …` — i.e. every mixed-reason run losing the "First failure:" qualifier and every single-root-cause run gaining it — leaves 16/16 green. A branch this PR introduced, with no test that can fail on it. One test with two differing `failures[].error` values closes it.

**N5. [cleanup] The PR body's Test plan is now factually wrong, and the four incidental fixes are undeclared.**

The body still claims `cast-design-stream-middleware.test.tsx` "17/17 passing" (it is 16 — one test was deleted) and `cast-design.test.ts` "39/39 passing, including a new regression test (`#3027/2`)" (it is 40, and there are two new regression tests — the second is the lock-timeout one that closes finding #1). It also says nothing about the summary-toast reason, the busy toast, the release-notes pair, or the register rows. CLAUDE.md requires incidental fixes to be declared in the PR body ("Also fixed, found in passing: …") precisely so a reviewer doesn't read them as scope creep.

**N6. [cleanup] The busy-toast message is a 100-character user-facing sentence duplicated verbatim at two sites in one file.**
`src/store/cast-design-stream-middleware.ts:354` and `:414`

Both tests match only `/already running/`, so a copy-edit to one — the likely edit for a sentence like this — diverges the two and ships green, leaving the same condition worded two ways depending on whether the user hit "Design full cast" or the drawer's single-design button. One module-level const, the way `LOCK_CONTENTION_FACT` (`server/src/workspace/file-lock.ts:149`) already handles the same problem.

---

### Verdict

The two correctness findings from pass 1 are genuinely fixed, both re-proved by mutation rather than taken on report. Findings 3–6 are fully resolved; finding 7 is half-resolved (N2).

**1 new correctness finding (N1, 🟠)** — so this pass re-triggers a review round once it is fixed and pushed. That would be pass 3, the last one the loop cap allows.

dudarenok-maker:
## PR review — pass 3 (head 01b4bb9d3e30343dea197de4c41cc275df9ac32d, depth high)

Delta-only re-review, and the **last round the loop cap allows**. Prior passes: `ffe43ce1…` (7 findings), `914cc2e8…` (6 findings, N1–N6). Depth re-derived from the current commit set — 11 commits, still multi-scope frontend + server, so `high`, unchanged.

Verified in a **detached scratch worktree at this head** (`C:\Claude\Projects\scratch-3161-pass3`), not the primary checkout and not the branch's own worktree. Merge-base against `origin/main` is `2b63b451` — exactly `main`'s tip, so the branch is current. `npm run typecheck` exit 0. `server/src/routes/cast-design.test.ts` 41/41, `src/store/cast-design-stream-middleware.test.tsx` 17/17 — both matching the PR body's updated counts. All 22 CI checks green at this head.

---

### Disposition of the 6 prior findings

**N1. 🟠 → ✅ VERIFIED RESOLVED — re-derived, not read off the diff.**

I drove the scenario myself rather than trusting the committed test's assertions: spied `ensureCharacterVoiceUuid` to throw a real `LockAcquisitionTimeoutError` on the first of two characters and dumped the **whole SSE transcript**, heartbeats filtered:

```json
[ {"type":"progress","characterId":"aria","name":"Aria","done":0,"total":2},
  {"type":"character_failed","characterId":"aria","name":"Aria",
   "errorReason":"Timed out waiting for another operation on this book to release its file lock — that is contention, not a problem with this item. Retry once the other operation has finished."},
  {"type":"progress","characterId":"brann","name":"Brann","done":0,"total":2},
  {"type":"character_designed","characterId":"brann","voiceId":"qwen-v_brann"},
  {"type":"idle","done":1,"total":2,"skipped":0,"clonedSkips":[],
   "failures":[{"characterId":"aria","name":"Aria","error":"Timed out waiting…"}]} ]
```

Curated `character_failed` for the failing character, the second character **still designed**, terminal `idle` with a real designed/failed summary, and **no `error` event at all**. Pass 2's transcript for the same input was `[progress, error]` and nothing else.

The paired test is real: restoring `server/src/routes/cast-design.ts` to `f3d3a341^` (reverting *only* the N1 hunk, nothing else — no other commit touches that file after `914cc2e8`) turns **exactly one** test red, `cast-design.test.ts:1680`, on `expected undefined to be defined` for the missing `character_failed` — red for precisely the claimed reason, green with the fix.

**N2. 🟡 → ✅ VERIFIED RESOLVED — I read the live page, I did not take the prior report's word for it.**

Fetched `https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a` directly. It carries `data-published-as="18" data-publish-id="ffc00c300f36"` — matching this branch's HTML exactly, so the live page **is** this branch's file, not a same-token coincidence — and both new blocks are present verbatim on the rendered page:

- A6: *"2026-09-11 — #3027 fixed (PR #3161), on-box re-run still owed."*
- A7: the matching block, naming the three bullets still owed.

Three mechanical checks pass against this head: `npm run check:onbox-register` OK, `npm run register:build -- --check` up to date, and — the one that actually compares against the artifact — `check:onbox-register -- --against-published <saved live page>` reports the register is **not behind** the published page. The rows are dated blockquotes that record the code fix and keep the debt open, which is the correct shape: this PR is not the on-box run.

**N3. 🟡 → ✅ VERIFIED RESOLVED, and the class is clear this time.** Both comments now describe what the code does, and both are *accurate*, which is the part worth checking rather than assuming:

- `voice-readiness-gate.tsx:52-56` — "would just duplicate that toast" is true: `onDesignFullCast` (`:62-70`) already pushes its own `kind:'warn'` "A design run is already in progress for another book." toast on the `designRunningElsewhere` branch, so falling through would produce a *second* toast about one condition.
- `voice-nudge-toast.tsx:1-8` — "the action is disabled" is true: `disabled={designRunning}` at `:55`, off `s.castDesign.active?.state === 'running'` (any book).

`grep -i "ignores a second start|silently no-op|is ignored|no-ops"` across `src/` returns no remaining claim about this middleware. No sibling left armed.

**N4. 🟡 → ✅ VERIFIED RESOLVED — mutation-proved.** Rewriting `cast-design-stream-middleware.ts:146` as `!sameReason ? …` — the exact mutation that left 16/16 green at pass 2 — now reddens exactly one test:

```
AssertionError: expected 'Designed 0 · 2 failed. no gemini key' to contain 'First failure: no gemini key'
```

The new test forces two *different* reasons across `c1`/`c2` and asserts both `toContain('First failure: no gemini key')` and `not.toContain('rate limited')`, so it pins which branch runs **and** which failure's text is surfaced.

**N5. 🟡 → ⚠️ PARTIALLY RESOLVED.** The counts are now right (17 and 41, both confirmed locally) and all seven incidental fixes are declared. **But two of the commit SHAs the body cites are not on this branch** — see P2 below.

**N6. 🟡 → ✅ VERIFIED RESOLVED — mutation-proved.** `grep` over the file shows the literal survives only as `DESIGN_ALREADY_RUNNING_MESSAGE` (`:45-46`) with both call sites (`:360`, `:419`) referencing the constant. Changing the constant's text to a sentinel reddens **both** busy tests (`2 failed | 15 passed`), which proves both sites genuinely read it rather than one having been left behind.

---

### New findings

Both 🟡. **No new correctness finding — nothing in this pass re-triggers the loop.**

#### 🟡 Minor

**P1. [cleanup] The ride-out catch's comment was named in pass 2's N1 as owed a correction, and it still names a call that is not in the try it describes.**
`server/src/routes/cast-design.ts:713-721`

```ts
/* Per-character synthesis failure — record it and move on.
   #2292 (owner decision) — a `LockAcquisitionTimeoutError` out of
   the persist steps in this try (`applyOverrideToCastFiles`,
   `persistEmotionVariant`, `ensureCharacterVoiceUuid`,
   `writeVoiceStylePersona`) keeps this per-character shape …
```

`f3d3a341` fixed the `ensureCharacterVoiceUuid` half — that call is now genuinely in this try at `:583-585`. `writeVoiceStylePersona` is not, and never was: `grep` gives exactly two call sites in this file, `:337` (inside `runPersonaPrePass`) and `:527` (inside *this PR's own* new persona try/catch at `:525-543`). Neither is inside the `for(;;)` try this comment sits in.

This is not pedantry about wording — **this exact comment is what produced pass 2's N1.** It asserted `ensureCharacterVoiceUuid` was covered when it was not, and the gap survived two prior reviews because the comment said otherwise. Leaving half of it still wrong re-arms the same trap for the next reader, who will look inside this try for `writeVoiceStylePersona`, not find it, and have to re-derive where it is handled. Pass 2's N1 asked for exactly this ("correct the `:706-714` comment so it names where each of the four calls is actually handled") and the fix commit did not touch it. Owed: name the persona pair's real handler (`:528`, the outer catch) rather than listing it among "the persist steps in this try".

**P2. [cleanup] The PR body's "Also fixed, found in passing" section cites two commits that are not on this branch.**

```
$ git merge-base --is-ancestor 95f5feca HEAD  →  NOT on branch
$ git merge-base --is-ancestor 6257a355 HEAD  →  NOT on branch
```

Both are pre-rebase orphans. Their on-branch equivalents are **`48a1a7ab`** ("Toast didn't show the failure reason") and **`f8256049`** ("Stale re-entrancy comment + weak test assertions"). Concretely: after merge, `git show 95f5feca` from the merged history finds nothing reachable, so the two entries the body offers as the audit trail for those fixes point at commits that will not exist in `main`'s ancestry. This is the residue of N5 — the counts were corrected by the same `gh pr edit` pass that left these two stale. A `gh pr edit` swapping the two SHAs closes it.

**P3. [cleanup] The technical release-notes register omits the `ensureCharacterVoiceUuid` half of what shipped.**
`docs/release-notes-next.md:598`

The entry (landed in `914cc2e8`, i.e. **before** `f3d3a341`) covers the persona-fallback per-character routing, the lock-timeout curation on the persona write, the summary-toast reason and the busy toast. It says nothing about the second call site — a `LockAcquisitionTimeoutError` out of `ensureCharacterVoiceUuid` no longer halting the job. That is a distinct behavioural change with its own commit and its own regression test, and `docs/release-notes-next.md` is the PR-refed technical register of exactly that. The user-facing line in `RELEASE_NOTES.md` is general enough to cover it ("one voice's failure doesn't hold up any of the others"); the technical one is not. One clause appended to the existing entry closes it — no new entry needed.

---

### Checked and clean — the combined-effect pass

This is the last round, so these are the "what did the fix commits do *together* that no single commit's review could see" checks, run rather than reasoned:

- **The ride-out loop now re-invokes `ensureCharacterVoiceUuid`, and it does not re-mint.** This is the highest-stakes consequence of `f3d3a341`: `voiceUuid`/`characterForDesign` moved from "computed once above `for(;;)`" to "computed inside the try", so a sidecar recycle calls the mint again. A second mint would flip the base's storage key while its embedding sat at the old one — the orphaning the surrounding `srv-43`/`#1057` comment exists to prevent. The suite's own ride-out test (`:552`) uses `aria`, which has a **pre-seeded** `voiceUuid`, so it structurally cannot see this. I drove a character with **no** `voiceUuid` through one forced ride-out with a counting passthrough spy: `calls=2 returned=["0CLV7wd0in5vMLbwDHCoZ","0CLV7wd0in5vMLbwDHCoZ"] persisted="0CLV7wd0in5vMLbwDHCoZ"` — two invocations, one uuid, and the persisted value matches. The early-out at `qwen-voice.ts:259-262` re-reads `cast.json` fresh on every call, so the retry sees the first call's write. Cost is one extra `design`+`cast` lock acquisition per retry; no correctness effect.
- **No new lock-order or curation breach from the move.** `ensureCharacterVoiceUuid` takes `design` → `cast` (correct global order) and is not nested inside another lock at its new position. A `LockAcquisitionTimeoutError` out of it cannot be misrouted into the ride-out branches first: `SIDECAR_DOWN_RE` is `/unreachable|did not complete within|stopped responding|recycling/i` and `isGpuBusyClass` keys on `code === 'GPU_BUSY'`; the lock message (`file-lock.ts:84`) matches neither, so it reaches the per-character `itemFailureReason(e, message)` at `:722`, which is what the transcript above shows.
- **The "first failure" toast text cannot conflict with a per-character error shown elsewhere — because there is no elsewhere.** `grep` over `src/**` (non-test) finds `failures[].error` read in exactly two places: the summary toast (`cast-design-stream-middleware.ts:144`) and `cast-design-slice.ts:209`'s `failures.push`, consumed only as `failures.length`. The toast is the only render of a failure reason anywhere in the app, so `sameReason`/`First failure:` cannot disagree with a second surface.
- **Accounting still closes.** The persona catch pushes one `job.failures` entry and broadcasts one `character_failed` without touching `job.done`, and `continue`s inside `try { … } finally { clearInterval(heartbeat) }`, so `done + failed + skipped = total` holds and no heartbeat leaks. A character failing at the uuid mint leaves its base undesigned, so scope `'both'` sends its variants to the existing `!baseName` skip at `:475` — one failure plus N skips, not N+1 failures.
- **`resubscribe` correctly did *not* grow a busy toast.** `:392` is still `if (handle || !bookId) return result;`. A cold-boot re-subscribe against a live handle must stay silent; only `REQUESTED_TYPE` and `SINGLE_REQUESTED_TYPE` toast.
- **A refused request still leaves no orphaned spinner.** Both new toast dispatches sit before `begin`/`beginSingle`, and `designSingleRequested`/`designAllRequested` are no-op reducers — so the drop leaves the pill untouched, as pass 1 checked for the pre-toast version.
- **Raw non-lock error text now reaching a toast is not a new breach.** `48a1a7ab` renders `failures[0].error`, so e.g. an `EPERM`/`ENOSPC` out of `writeJsonAtomic` would put an absolute `cast.json` path on screen. I checked this against the repo's actual rule rather than a general instinct: CLAUDE.md's invariant is scoped to the **lock-timeout class**, and `file-lock.ts:187-196` documents `requestFailureMessage` as deliberately curating "this one class and leav[ing] every other body verbatim". Both new sites route that class through `itemFailureReason`. Widening curation to every fs error would also directly contradict pass 1's finding 2 and #3027's own acceptance criterion, which is that the reason must be visible. Not a finding.
- **No plan prose staled.** `docs/features/195-design-full-cast.md:39` ("Per-character failures are recorded and the loop continues") is made *more* true; `:65`'s "re-entrancy" test-coverage line still holds (the guard exists, only its feedback changed). The 2026-06-07 single-design spec's "a second start … **409s**" describes the *server* route, which is unchanged — the middleware now short-circuits earlier, it did not remove the 409.
- **Register bookkeeping is otherwise complete.** `Closes #3027` present and outside any code span; release-notes pair both moved; no new config knob; no derived artifact staled (`openapi.yaml` untouched, these SSE events deliberately unspecced per plan 195); `bug`-labelled issue, so no `docs/BACKLOG.md` row owed.

---

### Verdict

**Nothing blocking. This PR is merge-ready once the three 🟡s are folded.**

Stated unambiguously, because this is the final round the gate allows: **I found no correctness bug in this pass.** N1 is genuinely fixed and I re-derived the proof end-to-end rather than reading the diff; N2's live view is genuinely published and I read the page rather than trusting the report; N3, N4 and N6 are each verified by mutation or by checking the claim against the code it describes; N5 is fixed apart from two stale SHAs.

**0 🔴 · 0 🟠 · 3 🟡** (P1, P2, P3 — plus N5's residual, which *is* P2). Per the findings-triage split, a 🟡-only pass does **not** re-trigger a review round, so fixing and pushing these does not reopen the loop — they are fixed in this round under the report-fix-record rule and the PR merges. All three are one-edit chores: a comment clause, a `gh pr edit` SHA swap, and a clause appended to an existing release-notes entry.

The one I would not merge without is **P1** — not because it breaks anything, but because that comment is the artifact that hid the pass-2 bug for two review rounds, and it is still half wrong.

Issue #3141, opened 2026-09-10
Five analyzer-model Advanced Settings overrides are ignored — Account-tab fields win silently


dudarenok-maker:
## What

Same defect class as #3139 (a registry knob whose read site bypasses the resolver). Five knobs in Advanced Settings → **Analyzer models & endpoints** accept and display a saved override that nothing reads:

| Knob (`server/src/config/registry.ts`) | Actual read path |
|---|---|
| `analyzer.ollama.url` (:1131) | `getResolvedOllamaUrl` — Account `ollamaUrl` ?? `OLLAMA_URL` ?? default (`server/src/workspace/user-settings.ts:597-600`). `ollamaUrl` is a required settings field, so the env var is also ignored whenever the settings cache is warm. |
| `analyzer.ollama.model` (:1141) | `getResolvedOllamaModel` — Account `defaultAnalysisModel` (if it contains `:`) ?? `OLLAMA_MODEL` ?? default (`user-settings.ts:766-771`) |
| `analyzer.phase0.model` / `analyzer.phase1.model` (:1204, :1214) | `selectAnalyzerForPhase` — env → per-request → Account field → default (`server/src/analyzer/select-analyzer.ts:68-100`) |
| `analyzer.phase1.minLagChapters` (:1224) | `resolvePhase1MinLagChapters` — env → Account field → default (`select-analyzer.ts:113-130`) |

None of the five keys is read via `configValue`/`resolveKnob` anywhere, so an Advanced Settings override is silently ignored while the row shows it as effective.

## Decision (owner, 2026-09-10)

- **Advanced Settings owns editing** for `analyzer.ollama.url`, `analyzer.phase0.model`, `analyzer.phase1.model` and `analyzer.phase1.minLagChapters`. Resolution becomes the standard resolver chain: env → Advanced override → registry default. Env rules are unchanged.
- **The Account tab shows those four read-only**, with a pointer to Advanced Settings (`#/advanced`).
- `analyzer.ollama.model` keeps its documented role as the **fallback Ollama tag** behind Account's "Default analysis model" (which stays editable in Account, since it also selects Gemini models and the engine). Only its fallback branch is wired through `configValue`.
- **Migration:** values already saved in the Account fields move into `configOverrides` on read, never clobbering an existing override. Precedent: `migrateLegacyEagerLoadFields` (`user-settings.ts:69-88`).

## Writers that change

[...]

Issue #3146, opened 2026-09-10
[agent instructions][claude][task] Add a guard that fails when a server-resolved registry knob is never read (#3139 step 2 of 2)


dudarenok-maker:
## What

Parent: #3139. The rate-limit bug was one instance of a class: a registry knob that Advanced Settings lets a user override, but that no server code reads through the config resolver, so the saved override is silently ignored. Step 1 (already committed on this branch) fixed the rate-limit instance. Add a guard test so the class cannot recur silently.

## Background you need

- Registry: `server/src/config/registry.ts`. `allKnobs()` / `getKnob(key)` are exported at ~lines 1366-1375. Each `ConfigKnob` (`server/src/config/types.ts`) has `key`, `env`, `apply: 'live' | 'restart-sidecar' | 'restart-server' | 'rebuild'`, and optional `isPrompt`.
- Resolver: `server/src/config/resolver.ts` — `configValue(key)` (~line 176) and `resolveKnob(knob)` (~line 87). Read the file for any other exported resolver-backed read helpers.
- `apply: 'restart-sidecar'` knobs are read by the Python sidecar, forwarded generically by iterating `allKnobs()` in `server/src/tts/spawn-sidecar.ts` (~lines 564-600). They are legitimately never read by name on the Node side — exclude them.
- Two knob families are read through a computed key, not a literal:
  - `qa.asr.maxWer.<lang>` → `server/src/tts/segment-asr-qa.ts` (~line 175 builds `` `qa.asr.maxWer.${lang}` ``).
  - `rate.{rpm,tpm,rpd}.*` → `server/src/analyzer/rate-limit.ts`, via the generic env-name → knob lookup that step 1 added to `resolveLimits`. Read the committed code to see exactly what it calls.
- Five knobs are **known to be unread** today and are being fixed separately in #3141: `analyzer.ollama.url`, `analyzer.ollama.model`, `analyzer.phase0.model`, `analyzer.phase1.model`, `analyzer.phase1.minLagChapters`.
- Precedent for a count/allowlist-shaped textual guard in this directory: `server/src/config/direct-env-reader-guard.test.ts` (read its header for the WALK / BLIND SPOTS documentation style and its stale-allowlist rule).
[...]

Issue #3145, opened 2026-09-10
[agent instructions][claude][verify] Verify #3139: saved rate-limit overrides reach the limiter, and unread registry knobs are guarded


dudarenok-maker:
## What

Verify the combined work of the two task children of parent #3139, committed on one branch: (1) `resolveLimits` honours saved Advanced Settings rate-limit overrides with env behaviour unchanged, and (2) a new guard test fails when a server-resolved registry knob is never read. Judge the committed diff, not the issue text. On PASS, open the PR.

## Where the work happens

- Worktree: `C:\Claude\Projects\wt-rate-limit-overrides`
- Branch: `fix/server-rate-limit-overrides` (already checked out, carrying both children's commits on top of `main` @ `733ee305`)
- Do NOT create another worktree. Do NOT work in the primary checkout `C:\Claude\Projects\Audiobook-Generator`.
- **Read-only for source:** report a verdict; do not fix anything yourself. Temporary mutations for re-running proofs must be reverted, and `git status --porcelain` must be empty when you finish.
- **Never run a commit or push in the background, and never use `--no-verify`.** This branch touches no `server/tts-sidecar/**` path, so `git push` runs in the foreground.

## Parent's acceptance, restated as a checklist

1. `resolveLimits` (`server/src/analyzer/rate-limit.ts`) resolves each of rpm/tpm/rpd as **env → saved override → `BUILTIN_LIMITS` → `FALLBACK_LIMITS`**. The env step is the unchanged `readEnvNumber` / `readTpmEnv` calls (confirm in the diff that those helpers and their `unlimited` / TPM `0` handling are untouched).
2. The override is found **generically** — by matching the computed env name to a registry knob's `env` — with no hand-written model→key map, and the resolver is consulted only when env is absent.
3. A saved TPM override of `0` resolves to `Infinity`.
4. The six `rate.{rpm,tpm,rpd}.gemma{,26}` knobs in `server/src/config/registry.ts` are `apply: 'live'`, and `npm run config:check` passes.
5. `server/src/analyzer/rate-limit.test.ts` has regression tests for: override enforced; env beats override; TPM override `0` removes the gate. Step 1's receipt shows red-before-fix output.
6. `server/src/config/registry-knob-read.guard.test.ts` exists and:
   - scopes to `allKnobs()` with `!isPrompt && apply !== 'restart-sidecar'`;
[...]

PR #3163, opened 2026-09-10
fix(server): honour saved rate-limit overrides in the Gemini limiter

### LINKED ISSUE #3139: Gemini rate-limit overrides saved in Advanced Settings never reach the limiter
## What

The six `rate.{rpm,tpm,rpd}.gemma*` knobs (Advanced Settings → **Gemini rate limits**, `server/src/config/registry.ts:1016-1075`) accept and display a saved override, but the limiter never reads it.

- `resolveLimits` (`server/src/analyzer/rate-limit.ts:76-84`) reads only `process.env` (`GEMINI_{RPM,TPM,RPD}_<slug>`), then `BUILTIN_LIMITS`, then `FALLBACK_LIMITS`.
- No code reads any `rate.*` key via `configValue`/`resolveKnob` (the keys appear only in the registry), and nothing copies saved overrides into `process.env` — the resolver's precedence is env → override → default (`server/src/config/resolver.ts:16-58`), and the only `.env` writer is `/env-cleanup` (`server/src/routes/config.ts:239`).
- Net effect: a limit raised in Settings (e.g. for a paid key) is silently ignored; analysis keeps waiting at the built-in free-tier caps. The Settings row shows the override as effective, so the UI misreports what the limiter enforces.
- Secondary: the knobs are marked `apply: 'restart-server'`, but `resolveLimits` runs on every `acquire` (`rate-limit.ts:203`), so once wired they apply live.

Found during the #3084 design pass. It is a plausible contributor to that reporter's stalled Gemini run, but that is **not confirmed**.

The same defect class (a registry knob whose read site bypasses the resolver) also affects five analyzer-model knobs; that is tracked separately because it needs a UI ownership change. `direct-env-reader-guard.test.ts` documents `rate-limit.ts`'s computed key as a known blind spot, which is why the existing guard did not catch this.

## Repro

1. No `GEMINI_RPM_GEMMA_4_31B_IT` in the environment.
2. Advanced Settings → Gemini rate limits → set **Gemma 4 31B RPM** to a value below 30 (e.g. 2).
3. Run an analysis on `gemma-4-31b-it`: the limiter still admits up to 30 requests/minute.

## Acceptance

- Limits resolve **env (unchanged, including `unlimited` and the TPM `0` sentinel) → saved override → `BUILTIN_LIMITS` → `FALLBACK_LIMITS`**; a TPM override of `0` means unlimited, matching the knob's help text.
- The six knobs' `apply` reflects live behaviour.
- Regression test that fails before the fix and passes after (override honoured; env still wins over an override).
- A guard that fails when a non-`restart-sidecar`, non-prompt registry knob has no resolver-backed read site, so this class cannot recur silently.

dudarenok-maker:
## Summary
- Advanced Settings rate-limit overrides for the Gemini limiter were stored and shown as effective but never read; `resolveLimits` now resolves env → saved override → built-in → fallback, with env behaviour unchanged.
- The six `rate.*` knobs now apply live.
- New guard `registry-knob-read.guard.test.ts` fails when a server-resolved registry knob is never read; the five analyzer-model knobs tracked by #3141 are allowlisted.

Found in passing: `analyzer.engine` (env `ANALYZER`) — `getResolvedAnalysisEngine()` reads a separate legacy user-settings field, never the `analyzer.engine` config-override store, so a saved Advanced Settings override for it is silently ignored, same shape as the original #3139 bug. Reported, not fixed — out of scope for this change and tracked by #3141 (the same open, correctly-labeled issue the five allowlisted analyzer-model knobs above already cite; a separate active branch, `fix/server-analyzer-model-overrides`, is addressing that whole knob group). Recorded in the guard's `KNOWN_UNREAD` allowlist.

## Test plan
- [x] Regression tests in `rate-limit.test.ts` (red before the fix, green after)
- [x] Mutation checks observed (re-verified during this review, each confirmed red then reverted):
  - Removing the `overrideValue` fallback from `resolveLimits` → `rate-limit.test.ts`'s "enforces a saved rate.rpm.gemma override" test fails (`onWait` never called)
  - Removing the `readEnvNumber(...) ??` precedence term from `resolveLimits` → `rate-limit.test.ts`'s "env still beats a saved override" test fails (`onWait` never called)
  - Removing `analyzer.phase1.minLagChapters` from `KNOWN_UNREAD` → guard test fails (set mismatch)
  - Adding `analyzer.stage2.minCoverage` (an actually-read knob) to `KNOWN_UNREAD` → guard test fails (set mismatch)
- [x] `npm run typecheck`, `npm run check:cycles`, `npm run config:check`, `npx vitest run src/analyzer src/config` (83 files / 1450 tests passed)
- [x] Mandatory `pr-review-gate` pass findings addressed in a follow-up commit (citation fix, two strengthened assertions, a stripped-source guard fix, a corrected docblock, a docs update, a typo)
- [ ] Mandatory `pr-review-gate` re-review pass (operator)
- [x] Before-shipping step 3 (on-box acceptance): **not applicable.** The saved-override precedence and live-apply behaviour is deterministic sliding-window logic, fully covered by `rate-limit.test.ts`'s fake-timer suite (temp `user-settings.json` + `writeConfigOverride`) — it needs no live GPU/sidecar/analyzer/book to prove. The register's existing `C1` row (a real, multi-hour Gemini cloud pass) additionally exercises the underlying limiter under genuine throttling, on the env-override path that predates this PR.

Closes #3139

https://claude.ai/code/session_01HHmS2UNg7pN6rp7F7tfEqD


dudarenok-maker:
## PR review — pass 1 (head 262cce29dc50a5090f0d872c915372b1d0877aef, depth medium)

Scope: all 9 changed files (`server/src/analyzer/rate-limit.ts` + `.test.ts`, `server/src/config/registry.ts`, `server/src/config/registry-knob-read.guard.test.ts`, `server/src/force-rerun-triggers.test.ts`, `server/vitest.config.ts`, `server/.env.example`, `RELEASE_NOTES.md`, `docs/release-notes-next.md`), plus the resolver/override chain they now depend on (`config/resolver.ts`, `workspace/user-settings.ts`) · verified in a clean detached checkout of the head SHA: `npm run typecheck` (root + server) exit 0; `npm run config:check` → `.env.example in sync ✓`; `npm run check:cycles` → OK, 15 cycles, all allowlisted (the two new `config/*` imports into `analyzer/rate-limit.ts` add none); `npx vitest run src/analyzer src/config src/force-rerun-triggers.test.ts` → 84 files / 1528 tests passed; `npx eslint` on all 6 changed source files → clean. Then six mutations, run individually and reverted:

| mutation | result |
|---|---|
| drop `overrideValue(...)` from `rate-limit.ts:99` | **red** — "enforces a saved rate.rpm.gemma override" fails (`onWait` never called). The #3139 regression test is real. |
| rename the lambda param inside `overrideValue`'s `allKnobs().find` | **red** — guard reports all six `rate.*` keys unread. The dynamic-reader declaration is genuinely verified. |
| add a new in-scope knob (`zzz.review.probe`, `apply:'live'`) with no read site | **red** — guard reports `zzz.review.probe`. The guard's core claim holds. |
| memoise `resolveLimits` per model | red — but only via **three pre-existing** env tests, not the new ones |
| memoise `overrideValue` per env name (saved override resolves once per process) | **GREEN — 22/22 pass.** See 🟠-1. |
| raise test 1's override `2` → `10` so the third acquire never blocks | `expect(settled).toBe(false)` still passes; first failure is line 309. See 🟡-1. |

### 🟠 Significant — the `apply: 'restart-server' → 'live'` reclassification, and the release note that depends on it, have no test at all

`server/src/config/registry.ts:1021,1031,1041,1051,1061,1071` move six knobs to `apply: 'live'`; `RELEASE_NOTES.md:250` promises "The new limit takes effect right away, with no restart needed"; `docs/release-notes-next.md:335` says "All six knobs move from `apply: 'restart-server'` to `apply: 'live'`, since `resolveLimits` is read on every `acquire` call."

Nothing pins that. Concrete repro — memoise the override lookup so a saved override resolves exactly once per process (i.e. restore `restart-server` semantics for the override path while leaving env live):

```ts
const _ovCache = new Map<string, number | undefined>();
function overrideValue(envName: string): number | undefined {
  if (_ovCache.has(envName)) return _ovCache.get(envName);
  const knob = allKnobs().find((k) => k.env === envName);
  const state = knob ? resolveKnob(knob) : undefined;
  const v = state && state.source === 'override' && typeof state.effective === 'number' ? state.effective : undefined;
  _ovCache.set(envName, v);
  return v;
}
```

Result: `npx vitest run src/analyzer/rate-limit.test.ts src/config/registry-knob-read.guard.test.ts --retry=0` → **Test Files 2 passed, Tests 22 passed**. A user who changes the RPM in Advanced Settings mid-session now sees nothing until a server restart, the registry's `apply: 'live'` badge in the Settings row is a lie, and the release note is false — and the suite is green. (The blunter mutant, memoising `resolveLimits` itself, *is* caught, but only by three pre-existing env tests that share module state — `honors GEMINI_RPM_/TPM_/RPD_ env overrides`, `treats env TPM 0 as unlimited…`, `cancels in-flight acquires…`. None of the three new tests notice either mutant.)

Why the new tests can't see it: `limiterWithOverrides` (`rate-limit.test.ts:260-277`) always writes the override **before** `await import('./rate-limit.js')` and `new GeminiRateLimiter()`, so every one of the three cases would pass under resolve-once semantics.

To be clear: the shipped behaviour is correct — `resolveLimits` runs inside `acquire()` (`rate-limit.ts:231`) and `readConfigOverrides()` reads the live `getCachedUserSettings()` cache. This is an untested behaviour claim, not a live defect. The missing test is one case: acquire once, then `writeConfigOverride('rate.rpm.gemma', 2)` against the already-imported settings module, then assert the next acquire throttles.

### 🟠 Significant — the `analyzer.engine` finding is deferred on a void reason, and its only record is a code comment pointing at a CLOSED agent ticket

`server/src/config/registry-knob-read.guard.test.ts:228` (`'analyzer.engine', // UNTRIAGED — reported on #3146`) and `:49-57`; PR body: "Reported, not fixed (out of scope for this change)".

The underlying defect is real and user-reachable, and I verified each hop:
- `analyzer.engine` is an editable enum knob — `server/src/config/registry.ts:1121`, `group: 'analyzer-models'`, `apply: 'live'`, `options: ['local','gemini']` — so Advanced Settings renders it as a live `<select>` (`src/views/advanced.test.tsx:471-477` says as much: including it in `descriptors` "would render a second, genuinely-editable 'Analyzer engine' `<select>`").
- `PUT /api/config` accepts it (`server/src/routes/config.test.ts:525`), so the value lands in `configOverrides['analyzer.engine']` and `/api/config` reports it `overridden: true`.
- The only thing that selects the engine is `getResolvedAnalysisEngine()` — `server/src/workspace/user-settings.ts:781-783` — `return getCachedUserSettings().analysisEngine === 'gemini' ? 'gemini' : 'local';`. It never reads the override store.

So: user sets Analyzer engine → `gemini` → Save → UI shows it as effective → every analysis still runs `local`. That is exactly the #3139 silent-ignore class this PR exists to close.

The gate problem is the disposition, not the discovery. Per CLAUDE.md → "Incidental findings: report, fix, record", "it would expand the scope of this PR" is a **verbatim void deferral reason**; the only finding allowed to leave the round unfixed is one needing a design pass, and then *its issue must name the decision owed*. There is no such issue — `gh issue list --search "analyzer.engine override" --state all` returns only #3145 and #3146, both **closed** `agent-instructions` tickets. A `KNOWN_UNREAD` entry reading "reported on #3146" is not the backlog.

The remedy is already open: **#3141** ("Five analyzer-model Advanced Settings overrides are ignored — Account-tab fields win silently", OPEN, `bug`, `area:srv`) is the same defect, same knob group, same Account-tab-field-wins mechanism. `analyzer.engine` is its sixth key, not a separate untriaged thing. Either fix all six under #3141's decision in this round, or at minimum move this key onto #3141 and change the comment to `// tracked by #3141` like its five siblings — the current split (five "tracked by #3141", one "UNTRIAGED — reported on #3146") reads as if they were different problems.

### 🟡 Minor

1. **`expect(settled).toBe(false)` can never fail** — `server/src/analyzer/rate-limit.test.ts:302-308`. `pending.then(...)` is registered *after* `await vi.advanceTimersByTimeAsync(10)`, and its callback is a microtask, so `settled` is unconditionally `false` on the very next synchronous line. Proven: raising the override from `2` to `10` so the third acquire resolves immediately leaves line 308 green — the first failure is line 309 (`expect(onWait).toHaveBeenCalled()`). Fix: register `.then` before advancing, then `await Promise.resolve()` before asserting.

2. **"env still beats a saved override for the same knob" cannot detect an env-precedence regression** — `server/src/analyzer/rate-limit.test.ts:317-328`. Proven: deleting the `readEnvNumber(...) ??` term from the `rpm:` line at `rate-limit.ts:99` leaves this test green (the only red was the unrelated pre-existing "cancels in-flight acquires"). The chosen env value `30` equals **both** `BUILTIN_LIMITS['gemma-4-31b-it'].rpm` and the registry default for `rate.rpm.gemma`, so env-honoured / env-ignored→default / env-ignored→builtin are indistinguishable; the test proves only that the override does *not* win. Fix: use an env value distinct from both (e.g. `5`) and assert the wait lands on the 6th acquire, not the 3rd.

3. **The dynamic-reader declaration is verified against un-stripped source, so commented-out code satisfies it** — `server/src/config/registry-knob-read.guard.test.ts:259-260` does `readFileSync(...)` then `raw.includes(dynamic.contains)`, while every other scan in the same file goes through `stripComments` (`:234`). Comment out `overrideValue`'s call sites in `rate-limit.ts` but leave `allKnobs().find((k) => k.env === envName)` inside the comment block and the guard stays green with #3139 fully reintroduced. Backstopped by `rate-limit.test.ts` today, but the header claims each entry is "verified against the ACTUAL file content below (not just trusted)". Fix: `stripComments(raw).includes(dynamic.contains)` — neither `contains` string carries a `//` or `/*`, so stripping is safe.

4. **The `stripComments` docblock misdescribes its own code** — `server/src/config/registry-knob-read.guard.test.ts:130-132` says "A docblock that merely quotes `configValue('foo.bar')` as prose is the one false-positive this guard accepts". It isn't: comments are blanked *before* the scan. I ported the function verbatim and ran it — `/* see configValue('zzz.docblock') */` → `[]`, `// configValue('zzz.linecomment')` → `[]`, `const msg = "call configValue('zzz.instring') first"` → `["zzz.instring"]`. The false positive the preceding sentence actually sets up is a **string literal** containing the text (an error message, a doc string), not a docblock. In a brand-new file whose whole job is to be read by the next person hitting this class, the example should match the rule.

5. **The regression plan's override sentence is now incomplete** — `docs/features/archive/06-analyzer-gemini.md:29` still reads "Built-in per-model limits … override via `GEMINI_RPM_<slug>`, `GEMINI_TPM_<slug>`, `GEMINI_RPD_<slug>` env vars; see `server/.env.example`". After this PR there is a second, now-live override source for the six `rate.*.gemma[26]` knobs that the plan does not mention. CLAUDE.md before-shipping step 1: "Changed behaviour cited in an existing plan → update that plan in the same diff."

6. **Before-shipping step 3 (on-box acceptance) is silently omitted rather than explicitly N/A'd** — the PR body's Test plan has four items, none of which addresses it, and the register / run sheet / live view are untouched. CLAUDE.md: "Skipping a step is fine when the step genuinely does not apply … but say so explicitly rather than silently omitting." There is precedent cutting both ways: `docs/testing/onbox-acceptance-register.md` Group C is the `GEMINI_API_KEY`-gated group and **C1** already covers "the limiter, the retries and the fallback interacting over hours" on a real Gemma run — so "a saved `rate.rpm.gemma` override throttles a live run with no restart" is at least arguably Group C territory. Either a row or one sentence, but not silence.

7. **Comment typo** — `server/src/analyzer/rate-limit.test.ts:333` reads "Request ExceedsTpmError"; the class is `RequestExceedsTpmError`.

### ✅ What is solid

- **The fix itself is right and the regression test has teeth.** Dropping `overrideValue` from `rate-limit.ts:99` goes red on the named test for the named reason (`onWait` never called). Resolution order matches the PR's claim exactly: env wins, then a saved override, then the built-in — and env behaviour is genuinely unchanged, because `readTpmEnv` already maps `0`/`unlimited` to `Infinity` before `tpmLimit`'s own `t === 0` sentinel is ever reached, so the new `tpmLimit` cannot alter any env-sourced value.
- **`overrideValue`'s `source !== 'override'` check is the right shape** — it means a knob whose env var is set but *invalid* still falls through to the saved override (matching `resolveKnobInner`'s own precedence at `config/resolver.ts:16-58`) instead of silently landing on the built-in, and it keeps `resolveKnob`'s env coercion/warning off the hot path for values the env read already accepted.
- **No hot-path cost and no new cycles.** `readConfigOverrides()` is an in-memory read off `getCachedUserSettings()` (`workspace/user-settings.ts:903-905`), not a disk read; `allKnobs()` returns the array directly (`registry.ts:1366-1368`). `check:cycles` is unchanged at 15 allowlisted cycles despite `analyzer/rate-limit.ts` now importing both `config/registry.ts` and `config/resolver.ts`.
- **The guard is a real guard, not a decoration.** Both mutations aimed at it went red for the right reason, including the "declared dynamic reader stopped matching reality" path — and the exact-set-equality assertion makes the allowlist un-staleable in both directions. Its `collectFiles` also fails *closed*: an unresolvable walk target yields zero read keys and a huge unread set, not a silent pass.
- **`apply: 'live'` needs no frontend change** — `src/components/settings/override-row.tsx:22-36` and `src/store/config-slice.ts:246-250` read `apply` from the server descriptor, so the badge and the "restart needed" banner follow automatically. `config:check` confirms the six `.env.example` `[live]` markers are generated, not hand-edited.
- **The `forceRerunTriggers` additions are correctly reasoned and don't collide** with `force-rerun-triggers.test.ts`'s `NOT_COVERED` pin (which names `src/index.ts`, not these two files), and the vitest.config comment is right that a blanket `server/src/**` trigger would.
- Release-notes pair present and correctly placed (the new `RELEASE_NOTES.md` bullet is the last line of the in-progress `# Castwright 1.15.0` section); `Closes #3139` is plain text outside any code span.

**Observation, not a finding against this PR:** `server/src/config/direct-env-reader-guard.test.ts` tripped the suite's `[retry-hazard]` reporter during my 84-file run and then failed once standalone under `--retry=0` (43.6 s of test time) before passing on the next standalone run (2.3 s). It is untouched here, excludes test files from its own walk (`:150`), and this PR adds no literal `process.env.<REGISTERED_VAR>` to non-test code — so this reads as a cold-FS-cache timing flake, not something this branch caused. Flagging it because it is not in `docs/testing/flaky-register.md`.

### Verdict

Not clean. Two 🟠 findings — one untested behaviour claim that the PR itself puts in the release notes, one CLAUDE.md incidental-findings gate violation with an open issue (#3141) already sitting there to absorb it — plus seven 🟡. The core fix is correct and well-tested for the bug it names; everything above is about the *other* claims the PR makes alongside it. Per the re-review trigger, fixing and pushing the 🟠 items re-triggers a pass.

Issue #3172, opened 2026-09-11
Analysing view: phase card still shows spinner + streaming while a run is paused, halted, or errored (decision owed)


dudarenok-maker:
Follow-up to #3169 (the idle analysing view showed phase 0 as running before a run started).

## Current behaviour

#3169 made the phase cards stay `pending` until a run has started. Once a run has started, the frontier phase card is `active`: a spinner, a progress bar, and a `· streaming` model chip. That stays true when the run is:
- **paused** by the user;
- **halted**, e.g. after a reload with a halted snapshot;
- **failed**: the view dispatches `setHalted` + `setError` and shows the error banner, yet the card beside it still spins and says "streaming". A setup failure now ends the job with a proper error (#3169), so this state is easier to reach than before.

This is the same complaint as #3169 (the page claims work is happening when it isn't), in a different state.

## Decision owed

What should the frontier phase card show while the run is paused, halted, or has errored? Options:
- (a) revert to `pending`, the same as never started;
- (b) a distinct paused/halted visual: no spinner, a "paused" or "stopped" chip, with the progress bar kept at its last value;
- (c) keep today's active rendering for paused, and (a) or (b) only for halted/error.

`derivePhaseState` (`src/lib/analysis-phase-state.ts`) and `PhaseCard` / `PhaseModelChip` (`src/components/analysing/`) are the seams. Two view tests from #3169 currently assert only that paused/halted cold boots render phase 0 as "not pending", so they don't lock in any of these options.

dudarenok-maker:
## Decision ratified: option (b)

The frontier phase card gets a distinct paused/halted render state, reusing this app's existing paused/halted visual vocabulary rather than inventing new tones or collapsing into `pending`:

- **Paused**: no spinner, progress bar frozen at its last value, chip goes neutral/amber ("· paused") — matching `top-bar.tsx:959` / `library-grid.tsx:265`'s existing paused treatment.
- **Halted**: same frozen-progress treatment, rose tone + `IconWarning` ("· halted") — matching `top-bar.tsx:964-968` / `library-grid.tsx:264`.
- **Errored**: same rose/halted treatment (server already conflates halted+error via `setHalted`/`setError`, so no separate visual is needed).

Rejected (a) revert-to-pending: throws away real signal the app already models (`activeStream.state`) and surfaces elsewhere in the same view. Rejected (c) keep active-for-paused: puts the frontier card at odds with the top-bar/library-card badges sitting right next to it.

[...]

Issue #3084, opened 2026-09-07
[FEATURE REQUEST] Please Add OpenAI-Compatible Endpoint for Script Generation Similar to Alexandria Audiobook


REFLEXGAMING007:
### What's on your mind?

I would like to request an OpenAI-compatible API endpoint for script generation, similar to the way Alexandria Audiobook handles LLM-based script generation.

This would make it much easier to use Castwright with OpenAI-compatible providers, local LLM servers, and other API gateways that expose an OpenAI-style "/chat/completions" interface.

Requested Features

1. OpenAI-Compatible Endpoint for Script Generation

Please add the ability to configure an OpenAI-compatible endpoint for the script-generation/analysis LLM.

Ideally, the configuration would allow users to specify:

- API endpoint / Base URL
- API key
- Model name
- Standard generation parameters

For example:

Base URL: http://localhost:1234/v1
API Key: ...
Model: ...

The implementation should be compatible with APIs following the OpenAI Chat Completions format.

2. Chunk Control

Please add a configurable Chunk setting for script generation.

This would allow users to control how much of the manuscript is sent to the LLM per generation request.

For example:

Chunk: 50,000 characters

This is particularly useful for large novels and for models/providers with different context-window limits.

3. Max Tokens Control

Please expose a configurable Max Tokens parameter for the LLM request.

Users should be able to control the maximum number of output tokens generated for each script-generation request.

Example:

Max Tokens: 65,536

4. Reasoning Effort Control

Please add a Reasoning Effort setting for providers/models that support it.

For example:

Reasoning Effort:
- Low
- Medium
- High

The value should be passed through appropriately when supported by the selected OpenAI-compatible API.

5. Custom JSON Payload

Please add an Advanced Custom JSON Payload field that allows users to add, modify, or remove arbitrary JSON parameters from the request sent to the LLM.

For example:

{
  "top_k": 40,
  "min_p": 0.05,
  "presence_penalty": 0.2,
  "custom_parameter": true
}

Ideally, this should be merged into the request payload so that users can:

- Add provider-specific parameters
- Override default parameters
- Remove parameters that a particular provider does not support
- Use experimental/model-specific options without requiring a Castwright code change

Why This Would Be Useful

OpenAI-compatible APIs are now commonly provided by local LLM servers, proxies, inference platforms, and alternative model providers.

Having these controls would make Castwright much more flexible and would allow users to select the LLM backend that works best for their particular novel, hardware, and workflow.

[...]

PR #3173, opened 2026-09-11
fix(frontend,server): keep analysis phases idle until started and end jobs that fail in setup

### LINKED ISSUE #3169: fix(frontend): analysing view shows phase 0 as running ("streaming" + spinner) before analysis starts
Refs #3084 — split out of the reporter's "analysis buffers forever on Detecting characters" follow-up. The #3084 feature design (OpenAI-compatible endpoints etc.) is separate and not part of this bug.

## What happens

Before a run is started, the analysing view renders phase 0 ("Detecting characters") as running: a spinner in the phase badge and a `<model> · streaming` chip, while the status pill reads **Idle**, Overall is **0%**, and the **Start analysis** button is still shown. An idle page that looks busy invites the user to wait instead of clicking Start.

## Evidence

- The reporter's screenshot (https://github.com/user-attachments/assets/6d107764-2adb-46c7-abe6-f39236393db3): Idle pill + Start analysis button + 0% alongside a spinner and a "Gemini 3.6 Flash · streaming" chip on Detecting characters.
- The reporter's `server.log` (https://github.com/user-attachments/files/32063745/server.log): after the upload's cover lookup at 20:08:35 there is not a single `[analysis]` line, although a started run logs every phase milestone and `Chapter N — <title>: detecting cast…` for the first chapter. The UI kept polling `/health` and `/capacity` past 20:15.

## Cause (confirmed in code on `main` at `2b63b451`, same logic at `v1.14.0`)

- `derivePhaseState` (`src/lib/analysis-phase-state.ts:46`) returns `'active'` for `phaseId === maxPhase`, and `maxPhase` is `0` before any start.
- `src/views/analysing.tsx:1437-1447` passes that to `PhaseCard` as `isPhaseActive` without consulting `analysisStarted` (`analysing.tsx:235`) or the resume state.
- `PhaseCard` then renders `IconSpinner` (`src/components/analysing/phase-card.tsx:458-462`) and `chipState = 'streaming'` (`phase-card.tsx:442-448`) → "· streaming" (`src/components/analysing/phase-model-chip.tsx:110`).

## Still open (to rule in or out in the fix)

The screenshot and log are also consistent with a Start click that failed silently before the server logged anything (e.g. an early-out in the start path, a GPU guard modal, series-prior scan, stage-1 preflight, or an unsurfaced request error). The fix must walk the Start click path for Gemini with default settings and confirm every early-out or failure either shows the user something or logs.

## Acceptance

- Before a run has started (and when not re-attaching to a running job), no phase card is active: no spinner, no progress bar, phase 0 chip shows the selected model as pending (not streaming). Phase 1's grey "warms up after ch. N" chip is a static schedule label, not an activity claim, and stays.
- Clicking Start still shows phase 0 active (spinner + streaming chip).
- The cold-boot rehydrate path (`analysing.tsx:253-290`, a `running` snapshot after reload) and the #865 resuming bridge still show phase 0 active.
- A paused/halted run keeps its phase progress rendering correctly.
- Paired unit test (red before, green after) and a Playwright mock-mode spec covering idle → Start.
- Any silent start failure found is fixed with its own paired test.

### LINKED ISSUE #3171: fix(server): single-design leaks its job and design-busy flag when the cast read throws before its try
Found in passing while fixing #3169, the same kind of bug fixed there in the analysis job runners.

## What happens

`server/src/routes/single-design.ts`:
- `runSingleDesign` awaits `readJson(castJsonPath(job.bookDir))` (~line 111) before its `try`.
- The route launches it with a bare `void runSingleDesign(...)` (~line 361) and no `.catch`. The sibling bulk design runner (`cast-design.ts` ~874) and `script-review.ts` (~438) both have one.

If that read throws, for example on a corrupt `cast.json` or a Windows `EBUSY`/`EPERM` on `readFile` (`readJson` rethrows parse and read errors), the rejection is unhandled. The consequences:
- `endJob` never runs;
- the SSE response stays open;
- the job stays in `inFlightByBook`;
- `markDesignBusy(bookDir)` (~351) is never cleared.

Until the server restarts, every main or chapter-subset analysis POST for that book answers `design_in_progress`, and single-design itself answers 409. The only log line is `[server] FATAL unhandledRejection` in `server.err.log`.

## Acceptance

- A throw anywhere in `runSingleDesign`, including before its current `try`, ends the job terminally:
  - subscribers receive an `error` event with a curated message, never a raw error message (lock-timeout messages carry workspace paths);
  - `inFlightByBook` is released;
  - the design-busy flag is cleared;
  - the error is logged.
- Nothing reaches the process-level `unhandledRejection` handler.
- A paired regression test forces the pre-`try` read to throw. It fails before the fix and passes after.
- Confirm no other bare `void run…(` detached dispatch under `server/src/routes/` has the same shape.

### LINKED ISSUE #3174: fix(server): remaining detached promises that can reject uncaught (analysis snapshots, script-review catch, upgrade catch, boot settings read)
Found while fixing #3169 and #3171: a sweep of every detached (never-awaited) promise under `server/src/` turned up more sites that can reject with nothing catching them. The full audit table is in PR #3173.

Each rejection lands in the process-level handler as `[server] FATAL unhandledRejection … (survived)`. Some also leave state wrong.

| Site | Rejection path | Consequence |
|---|---|---|
| `routes/analysis.ts` `void persistRunningSnapshot(...)`, the `void (async () => { tryResolveVerifiedBookDir… deleteAnalysisState })()` block, and `void persistTerminalSnapshot(job, 'paused' / 'halted', …)` | `withVerifiedBookDir` / `resolveVerifiedBookDir` slow path → `ensureWorkspace()` `mkdirSync` throwing on an fs failure | Snapshot not written or not deleted; a finished book's stale analysis state can linger and be offered as resumable |
| `routes/script-review.ts` `.catch` on `void runScriptReviewJob(...)` | Its handler sends the raw `err.message` to clients, and its `sub.res.end()` loop isn't guarded the way sibling `endJob`s are | A raw error message, possibly a lock-timeout path, reaches the client; a throw in the loop leaves later subscribers hanging |
| `routes/upgrade.ts` apply IIFE | The catch block's own `writeState(...)` is unguarded | A disk failure there leaves self-upgrade stuck at `applying` with no error shown, so retry is blocked |
| `index.ts` `void readUserSettings()` and `void sidecarSupervisor.start()` | `readJson` rethrows `JSON.parse` on a malformed `user-settings.json`, and nothing on either path catches it. The boot comment claims a malformed file "falls through to defaults"; it doesn't | Unhandled rejection at boot; on the start path the sidecar silently never starts |

## Acceptance

- Every site above either catches and logs the rejection, or routes it through its existing terminal or error path. Client-facing messages are curated.
- Each site has a paired regression test that forces the rejection.
- The false boot comment in `index.ts` is corrected.
- What a malformed `user-settings.json` should DO (fall back, recover from backup, or refuse to boot) is a separate design decision, tracked in its own issue. This issue only makes the failure loud and contained.

dudarenok-maker:
## Summary

Fixes #3169, split out of the #3084 report "analysis buffers forever on Detecting characters". The reporter's screenshot shows the status pill reading **Idle** and the **Start analysis** button still visible. At the same time, "Detecting characters" shows a spinner and a `· streaming` model chip, and their `server.log` has no `[analysis]` line at all. An idle page that looks busy invites exactly that wait. The leading explanation is that no run was ever started. A Start request that never reached the server has not been ruled out, so this PR also makes that case visible in the log.

**Frontend: idle analysing view** (`src/lib/analysis-phase-state.ts`, `src/views/analysing.tsx`)
- `derivePhaseState` treated the frontier phase as active before any run existed, because `maxPhase` is `0` at mount. It now takes a required `started` input, and the frontier stays `pending` until a run has started.
- The view passes `analysisStarted || resuming || hasStartedOnceRef.current || (!!manuscriptId && activeStreamSnapshot?.manuscriptId === manuscriptId)`.
  - The snapshot term makes a reloaded running, paused or halted run render as started from the very first render.
  - A snapshot only ever holds `running | paused | halted`, and completion clears it.
- Real progress and live data still override the gate.
- Phase 1's grey "warms up after ch. N" chip is intentionally unchanged: it's a schedule label, not a claim that anything is running.

**Server: analysis jobs that fail during setup** (`server/src/routes/analysis.ts`)
- Both detached job runners, `runMainAnalyzerJob` and `runSubsetAnalyzerJob`, ran about 150 lines of setup before their only `try`: phase-1 analyzer selection, the cloud escalation analyzer, language and author resolution, and more.
- A throw there became a process-level `unhandledRejection`:
  - `endJob` never ran;
  - the job stayed registered;
  - every subscriber's SSE stayed open forever;
  - nothing was logged under `[analysis]`.
- The job's existing `try` now covers setup. A setup failure ends the job through the same terminal handling as any other failure: `aborted`, `STALE_BOOK_DIR`, or `classifyAnalysisFailure`. The client therefore gets the same message it gets for any other analysis failure, including the curated sentence for a lock timeout.
- This did **not** cause the #3084 reporter's stall: their `server.err.log` has no `FATAL` line.

**Server: every analysis POST leaves a log line, and every run says how it began**
- `[analysis] request received manuscript="…" model="…" fresh=…` is logged at the top of the handler on every POST. That includes the stream middleware's own body-less subscribe POST and any reload or rejoin.
- One outcome line follows:
  - `[analysis] start manuscript="…" engine=… model="…"` for a new job;
  - `[analysis] subscribe manuscript="…"` when attaching to a live job. It deliberately names no model, so it can never print one the job isn't running.
- The chapter-subset route logs `[analysis-subset] request received …` on every POST, with `chapters=N` in place of `fresh`. It logs a `start` or `subscribe` outcome line once its chapter-id validation passes; a request that fails validation gets its `bad_request` SSE error and no outcome line. The route previously logged nothing on entry.
- The old resolved `engine=… model=…` line, which only printed when the request named a model, is gone. Request-supplied values (the manuscript id from the URL and the requested model) are `JSON.stringify`d in every one of these lines, so a LAN client can't inject a newline and forge a log line.

**Also fixed, found in passing:**
- **#3171:** `single-design.ts` had the same pre-`try` shape. `runSingleDesign` read `cast.json` before its `try`, and the route dispatched it with a bare `void`. A read failure leaked the job and left the book's design-busy flag set until restart, so every analysis POST for that book answered `design_in_progress`. The read now happens inside the existing `try`, so a failure ends the job with an SSE error whose message goes through `requestFailureMessage` (a lock-timeout message, which carries a workspace path, is replaced with the curated sentence; other errors pass their own message through, unchanged), frees `inFlightByBook`, clears the design-busy flag and the heartbeat, and logs `[single-design] failed` with the raw error.
- A comment in `runSubsetAnalyzerJob` claimed "no outer catch", and a rewritten comment omitted the #2260 lock-timeout rethrow. Both are corrected, and comment history dropped by the setup hoist is restored.

- **#3174 (the rest of the class):** an audit of every detached `void …` promise under `server/src/` found more sites that could reject with nothing catching them. All of them are now contained, each with a paired test:
  - the analysis state-file writes;
  - `script-review`'s start-failure path, which passed any error's own message to clients, including a lock-timeout message that carries a workspace path;
  - the self-upgrade apply task's state writes. When a write failed, the actual apply error went unlogged, and a failed `restarting` write skipped the restart. The state can still stick at "applying"; this PR doesn't change that;
  - the boot-time settings read and sidecar start. The boot comment claimed a malformed `user-settings.json` "falls through to defaults"; it doesn't.

  Two sites the audit couldn't classify (the catalog audit and the shutdown sequence) were traced separately and are safe. How each of the four is contained:
  - Analysis state writes: `persistRunningSnapshot`, `persistTerminalSnapshot` and the stale-snapshot delete now catch internally and log a non-fatal warning. The one awaited caller (`/pause`) doesn't depend on a rejection.
  - `script-review`: its start-failure path now goes through `requestFailureMessage`, guards its `res.end()` loop the way the other job runners do, and logs the raw error.
  - Self-upgrade: all three writes of the apply task's upgrade state now go through `tryWriteState`. The real apply failure is still logged when the state write fails too, and a successful upgrade still restarts.
  - Boot: the settings warm-up and sidecar start run through two small helpers, `bootWarmUserSettings` and `bootStartSidecarSupervisor`, that log loudly. `readUserSettings` itself is unchanged; see #3175.

**Filed, not changed here:**
- #3202: a chapter-subset analysis request that names different chapters than the subset job already running silently joins that job. What should happen instead (join with a warning, reject, or queue) is a decision owed. This PR only corrects the two comments that claimed the route rejects a mismatch.
- #3175: what a malformed `user-settings.json` should do (recover from backup, run on defaults without overwriting, or refuse to boot). Today every settings read rejects; this PR only makes the boot-time failures loud and contained.
- #3172. While a run is paused, halted or errored, the phase card still shows a spinner and "streaming". That behaviour predates this PR, and a design decision is owed on what it should show. The two new paused/halted tests assert only "not pending", so they don't lock it in.

**Not changed (ruled out of scope on #3169):** client fetch timeouts and disk-read timeouts. No occurrence has been observed, and choosing timeout values would need a design pass.

**Before-shipping notes:**
- **Regression plan:** none. This is a small, localized fix; the issues plus their brief comments are the spec.
- **On-box acceptance:** not applicable. Nothing here needs real hardware, and mock-mode e2e plus server tests cover it.
- **Release notes:** included, in `docs/release-notes-next.md` and `RELEASE_NOTES.md`.

Closes #3169
Closes #3171
Closes #3174
Refs #3084

## Test plan

- [x] cloud `verify.yml` (required status check): green at `efdaea30`, 21/21 checks (Linux verify, server fast+slow, 8 E2E shards, visual baselines, Windows, sidecar pytest)
- [x] `npm run verify:fast:branch` locally at `9ff0edac`: exit 0. Commits after it (`efdaea30` onward, including the review-gate pass-1 fixes) were each verified with their own targeted suites and typecheck. Cloud `verify.yml` on the final head is the full-battery check for them.
  - `lint` and `typecheck` passed, served from cache (their inputs were unchanged since an earlier green run on this head); `config:check`, `audit`, `audit:server` and `test:sidecar` were out of scope.
  - `test:hooks` 2402/2402, `check:budget-poll`, `test` 4960/4960 (346 files) and `build` passed.
  - `test:server` passed on its 2nd attempt: 8388 passed, 9 skipped, 549 files passed and 1 skipped. Attempt 1 was a vitest fork-pool crash ("Worker exited unexpectedly" ×2), not a test failure. It ran with reduced concurrency while 3 other vitest batteries were running.
  - `[retry-hazard]` retries appeared in pre-existing files this branch doesn't touch: `analysis.rename-midrun.test.ts`, `voice-style.test.ts`, `chapter-splice.test.ts`.
- [x] `npm run typecheck`: clean
- [x] Frontend vitest (`src/lib/analysis-phase-state.test.ts`, `src/views/analysing.test.tsx`, `src/components/analysing`): green. New tests cover:
  - idle → `pending`, Start → `streaming`;
  - a cold-boot running snapshot → `streaming`;
  - cold-boot paused and halted snapshots → not `pending`, asserted synchronously on first render;
  - undefined `manuscriptId` → `pending`;
  - a non-zero-frontier `started` pair.
- [x] Playwright `e2e/analysing-idle-phase-state.spec.ts` (new, mock mode): the idle chip is `pending`, and it's `streaming` after Start. The `analysing-progress` and `analysing-multi-model` specs were re-run green.
- [x] Server vitest:
  - new `analysis.setup-throw.test.ts`: a setup throw produces a terminal error through `classifyAnalysisFailure`, the promise resolves, the job is deregistered, the aborted path works, lock-timeout messages stay curated, and the subset runner behaves the same;
  - new `analysis.request-log.test.ts`: `request received` is logged before hydration, `start` exactly once for a new job, `subscribe` (not `start`, and no saved-default model) for a subscribe-shaped POST against a live job on both routes (body-less on the main route; `{ chapterIds: [1] }` on the subset route, which must pass its chapter-id validation), and a newline-injection case;
  - `single-design.test.ts`: a new #3171 regression test makes `runSingleDesign`'s own `cast.json` read throw, while the route's upfront read still succeeds. It asserts:
    - a terminal error;
    - the design-busy flag cleared;
    - a second design request not answered with 409;
    - a `[single-design] failed` line logged;
    - no unhandled rejection.

    A sibling case shows that a lock-timeout error on that read reaches the client as the curated sentence rather than the workspace path. The regression test hangs to timeout against the pre-fix file. Each of those assertions fails when its fix is removed, and the file passes 15/15;
  - #3174 detached-promise sweep. Each test forces its rejection, and each one fails against the unfixed code, run with retries off:
    - new `analysis.snapshot-detach.test.ts`: a snapshot write fails, the warning is logged, the job still ends, and no unhandled rejection escapes.
    - `script-review.test.ts`: when a lock timeout makes the review fail to start, clients get the curated sentence instead of a message containing the workspace path, and the connections close. Other errors still pass their own message through, by design.
    - `upgrade.test.ts`: the upgrade state write fails on the catch path and on the success path. The failure is logged, and the restart is still scheduled on success.
    - `index.test.ts`: the boot settings read and the sidecar start both fail, and both failures are logged loudly.
  - `analysis.test.ts` and the other job-runner suites were re-run green, and so was `cast-lock.guard.test.ts`.
- [x] Mutation proofs (`--retry=0`), each restored afterwards: the `started` gate, the ref-only `started`, the setup widening, the old unconditional log line, the `!!manuscriptId` guard, the snapshot term, and the single-design pre-`try` read. Every new test went red on its assertion.
- [ ] Reviewer: read `server/src/routes/analysis.ts` with `-w`. Most of its line count is re-indentation from widening the `try`.


dudarenok-maker:
## PR review — pass 1 (head f32e94e740026ec8faf3b510cdf8cfae09f82444, depth high)

Scope: all 13 changed files (`server/src/routes/analysis.ts`, `single-design.ts`, the three new/changed server suites, `src/lib/analysis-phase-state.ts`, `src/views/analysing.tsx` + their suites, `e2e/analysing-idle-phase-state.spec.ts`, both release-notes files), reviewed in a detached scratch worktree at the PR head.

Verified before probing: `npm run typecheck` clean (frontend + server). Server — the 3 new/changed suites green (24 tests), plus 13 analysis-adjacent route suites (83), `analysis.test.ts` + `cast-lock.guard.test.ts` + `openapi-design-parity.test.ts` (239). Frontend — `analysis-phase-state.test.ts`, `analysing.test.tsx`, `components/analysing` green (134). Playwright `e2e/analysing-idle-phase-state.spec.ts` green. ESLint clean on every changed file.

Red-before-green re-derived independently, not taken from the PR body:
- base `analysis.ts` + the two new server suites → **10/10 red**, each "promise rejected … instead of resolving" — the exact D1 shape;
- base `single-design.ts` + the #3171 test → **red by 20s timeout** (SSE never closes) — the exact leak shape;
- base `analysis-phase-state.ts` + `analysing.tsx` → **4 red** (the two `started`-gate unit cases + two view cases); the new paused/halted pair correctly stays green there, since it guards the F3 fix, not the original bug;
- base frontend + the new e2e spec → **red** at the idle `data-phase-state=pending` assertion;
- mutation: dropping `!!manuscriptId &&` → 1 red; dropping the `activeStreamSnapshot` term → 2 red (the synchronous paused/halted pair). Both guards are really pinned.

### 🔴 Blocking

None.

### 🟠 Significant — the new log lines close the injection hole for `model` and leave it wide open for `manuscriptId` (correctness bug)

`server/src/routes/analysis.ts:3231-3234` (also `:3374`, `:3468`, `:6484-6487`, `:6599`, `:6679`)

F6 wraps `model` in `JSON.stringify` with the stated threat model "so a LAN client can't inject a forged `[analysis]` line via a newline in it". `manuscriptId` sits in the *same* template literal and is equally client-supplied: it is `req.params.id`, and Express percent-decodes route params, so `%0A` arrives as a literal newline.

This is not adjacent pre-existing code — the `request received` line is the **first** `console` statement in the handler, above the language pre-flight and above `getOrHydrateManuscript`, so it is the one `[analysis]` line in the file reachable with a *completely unvalidated* id. Test (a) in the new suite proves that reachability with an unregistered id; the only missing step is a newline in it. Every pre-existing `[analysis] … manuscriptId` line sits behind a successful hydrate.

Concrete repro, run against the real `analysisRouter` (not a toy router), body `{}`:

```
POST /api/manuscripts/mns_probe%0A%5Banalysis%5D%20start%20manuscript%3Dmns_victim%20engine%3Dgemini%20model%3D%22gemma-4-31b-it%22/analysis
```

server.log receives **two** lines:

```
[analysis] request received manuscript=mns_probe
[analysis] start manuscript=mns_victim engine=gemini model="gemma-4-31b-it" model="(saved/default)" fresh=false
```

The second is a fully-formed, forged `start` outcome line naming a different manuscript. That is precisely the signal this PR exists to make trustworthy: an operator reading `server.log` to tell "a run started and stalled" from "no run was ever started" can now be handed a fabricated answer by any LAN client. The new test (f) cannot catch it — its per-line `expect(line).not.toMatch(/\n/)` loop is exactly the right assertion, but the id it uses contains no newline.

Fix: stringify (or sanitise) `manuscriptId` at all six new sites, the same way `model` already is; and extend test (f) with a `%0A`-bearing id so the loop it already has actually gets exercised on this half of the class.

### 🟠 Significant — #3171 ships with no release-notes entry, and the PR body states the opposite (house gate)

`RELEASE_NOTES.md`, `docs/release-notes-next.md`

At this head, grepping either file for `3171`, `single-design`, or the design-busy symptom returns nothing. Both new bullets cover #3169 only.

#3171 has a plainly user-visible delta — the PR body describes it itself: a failed cast read left the book's design-busy flag set until restart, so *every* subsequent analysis POST for that book answered `design_in_progress`. A user hitting that sees "A 'Design full cast' run is in progress for this book" forever, with no design running. That is not a no-shippable-delta change, so before-shipping step 5 applies to it in full.

What makes this a finding rather than an omission to note: the PR body's Before-shipping notes assert **"Release notes: included, in `docs/release-notes-next.md` and `RELEASE_NOTES.md`."** The claim is true for #3169 and false for #3171, so the gate reads as discharged when it is not.

### 🟡 Minor

**1. `analysis.request-log.test.ts:407` — test (e)'s title claims a condition it does not exercise, and the release note inherits the overstatement.** The title is "a subscribe-shaped POST (no body) against a live subset job…", but the body sent is `{ chapterIds: [1] }`. It has to be: the subset route's `chapterIds` validation (`analysis.ts:6544`) sends `bad_request` and ends the response *before* the subscribe dispatch at `:6594`, so a genuinely body-less POST can never reach `[analysis-subset] subscribe`. The consequence outside the test: `docs/release-notes-next.md` says the subset route logs "the same three shapes", but the middleware's own resubscribe posts `snap.subsetChapterIds ?? []` (`src/store/analysis-stream-middleware.ts:159`), and a snapshot without `subsetChapterIds` — documented as reachable for "pre-D1 snapshots" at `src/lib/types.ts:609` — therefore logs `request received` and then `bad_request`, never an outcome line. Either retitle the test to match what it drives and soften the note, or (better) move the subset `subscribe` line above the chapter-id validation so the shape the note promises is real.

**2. `single-design.test.ts` — `expect(consoleErrorSpy).toHaveBeenCalled()` does not check what its comment says it checks.** The comment reads "The raw error was logged server-side", but any `console.error` from anywhere in the request satisfies it; deleting or renaming the new `[single-design] failed` line would not turn this red. The sibling suite added in the same PR does it correctly — `analysis.setup-throw.test.ts:237-240` finds the call whose first argument `startsWith('[analysis] failed')`. Mirror that.

### ✅ What is solid

- **The D1 widening is done properly, not just moved.** `activeModelId` / `analyzerLabel` / `lastStep` are hoisted with their initialisers unchanged, so the catch reads the same values for a setup-span throw as for a pipeline throw. The `runUserSettings` rename is *required*, not cosmetic: the later `const userSettings = await readUserSettings()` at `:5750` sits at the same block depth, so after the widening the two would collide in TDZ — and the four rewired call sites (`:3647`, `:3666`, `:3672`, `:3675`) all still read the read-once opts snapshot, while `:5806`/`:5817` still read the fresh live re-read. No semantic swap.
- **`derivePhaseState` has exactly one production caller** (`analysing.tsx:1439`), so the new required `started` input cannot be silently defaulted anywhere. The three real-run-data rules (`done` by progress, `active` by live, `done` by later-phase advance) are all ordered *above* the `started` gate, and two new unit cases pin that they keep winning when `started` is false.
- **The `!!manuscriptId` guard (F3) is a genuine fix, not defensive noise** — `undefined === undefined` really did read as started, and dropping the guard reproduces it in one test.
- **Reading `activeStreamSnapshot` directly rather than trusting `hasStartedOnceRef`** is the right call for the ref-write-triggers-no-render problem, and the two paused/halted tests assert synchronously (`getBy…`, no `findBy`/`waitFor`) so a regression to the ref-only form is a hard failure rather than a papered-over flash.
- **The subscribe path deliberately omits `model`**, and test (d) asserts the absence across *every* line the POST logged, not just the subscribe line — which is what actually catches a regression to the old unconditional resolved-model line.
- **The single-design `not_found` early return moved inside the try safely**: the heartbeat now starts before it, and `finally { clearInterval(heartbeat) }` plus `endJob`'s `inFlightByBook.delete` + `clearDesignBusy` cover it. The follow-up POST in the regression test asserting a non-409 is the right proof that the leak is gone.
- **The superseded regression test was caught and replaced rather than kept.** `d92c40d5` explicitly documents that the earlier `applyOverrideToCastFiles`-throwing test sat inside the try on both sides of the fix and so could never have detected #3171 — and the replacement really does time out against the pre-fix file. Keeping the old one, relabelled as post-setup coverage, is the right disposition.
- **The comment corrections are real, not cosmetic.** The subset runner's "fire-and-forget with no outer catch" claim was made false by this very change; F7 restores the 2026-06-06 ch12 `lastStep` history the hoist dropped; F4 adds the #2260 lock-timeout carve-out the code already granted. The source-scanning guard in `analysis-reject-edge-reconcile.test.ts` still passes across the re-indentation.
- **Scope discipline held.** #3172 is a real design-pass carve-out with the decision named ("what should a paused/halted phase card show"), it carries the `bug` label so no `docs/BACKLOG.md` row is owed, and F5 correctly loosened the two paused/halted tests to "not pending" so the PR does not lock in the behaviour it just deferred. `Closes #3169` / `Closes #3171` / `Refs #3084` are all present and outside code spans.

### Verdict

Two 🟠 findings, both in scope for this round: the `manuscriptId` half of the log-injection class is still armed at six new call sites, and #3171 ships undocumented while the PR body says otherwise. Neither is a design question. The 🟡 pair is a mislabelled test title (with a real behavioural tail in the subset route) and a log assertion that cannot fail for the reason its comment gives.

The engineering underneath is strong — every paired test in this PR went red against base for the reason claimed, including the e2e, and the two mutation probes I ran independently both fired.

One operational note: this review is pinned to `f32e94e7`, the PR's current head. The `fix/frontend-analysing-idle-phase-active` branch has commits beyond it that were not pushed to this PR at review time; if any of them address the above, say so on re-review rather than re-fixing.

dudarenok-maker:
## PR review — pass 1 (head efdaea30, depth high)

Scope: all 21 files in `2b63b451..efdaea30` (full diff, plus `git diff -w` for `server/src/routes/analysis.ts` with the non-`-w` re-indent spot-checked). Verified: CI at the time of review shows 20 checks passing and `Windows (test + test:server + test:hooks)` still **pending** (the PR body says 21/21 green). Suites were not re-run locally. One out-of-tree probe was executed against the worktree's own Express 5.2.1. Adversarially probed:
- the setup-span `try` widening in both runners (binding scope, catch/terminal paths, snapshot side effects);
- the new request/outcome log lines, including injection;
- the `started` gate across idle / Start / cold-boot running, paused and halted / no manuscript / analyzer not ready;
- #3171's heartbeat and `not_found` path;
- #3174 sweep completeness over every `void` site under `server/src`;
- the curation seams;
- release-note claims against `crash-logging.ts`.

### 🟠 Significant — F6's newline guard misses `manuscriptId`, which the new pre-validation `request received` lines print raw (correctness)

`server/src/routes/analysis.ts:3262-3265` (main) and `:6515-6518` (subset).

`manuscriptId = req.params.id` is URL-decoded by Express, so `%0A` becomes a literal newline. Both new `request received` lines interpolate it raw at the top of the handler, before `getOrHydrateManuscript` validates anything. The removed resolved line (`[analysis] manuscript=… engine=…`) only printed after hydration had succeeded. The two new lines are the ones this PR puts ahead of any validation.

Repro, executed with the worktree's Express 5.2.1 against the same mount (`/api/manuscripts` + `/:id/analysis`) and the exact template:

`POST /api/manuscripts/x%0A%5Banalysis%5D%20start%20manuscript%3Dforged%20engine%3Dgemini/analysis` with body `{}`

→ `req.params.id === "x\n[analysis] start manuscript=forged engine=gemini"`, and the log call emits:

```
[analysis] request received manuscript=x
[analysis] start manuscript=forged engine=gemini model="(saved/default)" fresh=false
```

The second line is a well-formed forged `start` outcome line. That is exactly the signal D2/F2 exist to make trustworthy, and exactly what F6 claims to rule out ("Request-supplied strings are `JSON.stringify`d so a LAN client can't inject newlines into the log").

Nothing upstream stops it. No `.param()` validator exists under `server/src`. `app.ts`'s pre-router middleware doesn't touch the path: body parsers, `apiLimiter`, `requireLanToken` (which loopback bypasses, and which is off unless LAN HTTPS plus a token are set), and `requireSameOrigin`.

The F6 test (`analysis.request-log.test.ts:461-497`) injects only through `model`, so it can't see this. One field was hardened; the class stays armed.

### 🟡 Minor (cleanup)

**1. The #3171 "curated message" claim is false for the non-lock errors this fix newly sends to clients, and the test that claims to lock it can't fail.**

Three texts make the claim:
- `docs/release-notes-next.md:337`: "…ends the job through the normal terminal-error path: a curated client-facing message…"
- the PR body: "an error message that never exposes the raw error (it goes through `requestFailureMessage`)"
- `single-design.ts:206-208`: "unlike the curated client-facing message below"

`requestFailureMessage` (`file-lock.ts:214-216`) replaces only `LockAcquisitionTimeoutError`. Everything else passes `(e as Error).message` through verbatim, per its own doc at `:210-213`. `readJson` (`state-io.ts:46-47`) rethrows `readFile` errors unwrapped.

Scenario: the very case the note names, "a Windows file lock" on `cast.json` at `runSingleDesign`'s read. The read throws `EBUSY: resource busy or locked, open 'C:\Users\<name>\…\.audiobook\cast.json'`, and that message goes verbatim into the SSE `design_failed` event to every subscriber. Before the fix it reached no client at all, because the job hung.

The test at `single-design.test.ts:576-578` ("Curated: a real JSON.parse error, never a raw filesystem path", plus a path regex) injects a `JSON.parse` `SyntaxError`. That message never contains a path, so the assertion passes whether or not anything is curated.

Correct the three texts to say only lock timeouts are curated. S3 already made this exact correction for script-review on this branch; single-design was missed. Whether ordinary fs errors *should* be curated is the separate judgement `requestFailureMessage`'s own doc defers. That decision isn't owed here; the text correction is.

**2. The request-log tests claim "printed before hydration" but can't detect the line moving after hydration.**

Test (a) (`analysis.request-log.test.ts:205-246`) and test (c) (`:297-339`) are affected. Their comment at `:234-238` says the absent `start` line is "proof the request-received line above is logged strictly BEFORE hydration".

Move the `console.log` to just after `const record = await getOrHydrateManuscript(manuscriptId)` (`analysis.ts:3327`), still before the `!record` return. Test (a) keeps passing all four assertions: the 200, `unknown_manuscript`, the line and its fields, and no `start` line. The missing `start` line only proves no job started. The same holds for (c) against `:6545`.

D2's whole point is a line that precedes the slow cold re-parse, and nothing locks that ordering. One way to lock it is to make hydration observable in the test, for example by recording whether `request received` had already printed when `getOrHydrateManuscript` was entered.

**3. The boot settings failure log hardcodes a cause it can't know.**

`server/src/index.ts:139` logs `(malformed, not merely missing -- not auto-recovered, see #3175)` for *any* rejection. But `readUserSettings` (`user-settings.ts:355-360`) also rejects on errors that aren't a malformed file:
- `migrateLegacyUserSettings`'s `mkdir`/`copyFile` (`:49-50`);
- `readFile` errors such as `EBUSY`/`EPERM`, via `readJson` (`state-io.ts:47`).

Scenario: OneDrive or antivirus holds `user-settings.json` at boot, producing `EBUSY`. The loud line tells the operator the file is corrupt and points at the malformed-file recovery design, yet the next read retries (`cached` is unset) and succeeds. The doc comment at `:120-126` makes the same assumption. The raw `err` already carries the real cause, so the fixed text only needs to stop asserting one.

**4. `RELEASE_NOTES.md:7` misstates the before and after for the upgrade and sidecar changes.**

It says "that failure used to disappear silently … with no explanation anywhere" and "the log used to show only a generic crash with no mention of what actually broke". Before this PR both rejections reached `crash-logging.ts:72-74`, which logs `formatCrash` (`:44-48`), i.e. the error's full stack. The log did name what broke.

What was actually lost before, and isn't mentioned:
- On the upgrade catch/else paths, `writeState` ran before the `console.error`, so the apply failure's own reason was never logged.
- On the success path, a failed `restarting` write meant SIGTERM never fired, so the new release never started.

The sentence also reads as if the stuck "applying" state is fixed. The PR's own tests assert it still sticks (`upgrade.test.ts:239`, `:290`). The comment at `index.ts:150-152` ("no logged reason beyond the generic FATAL unhandledRejection line") understates the pre-fix log the same way.

**5. Comments this change made false.**
- `analysis.test.ts:8724-8726` says `analyzerLabel`/`activeModelId` "are now computed from `selection` as the first statement inside it". They are computed *before* the `try` (`analysis.ts:3606-3610`, `:6756-6759`).
- `analysis.ts:6507-6510` (subset route) says the stream middleware's subscribe POST "reaches this handler too, with no body". The middleware sends `chapterIds` for subset (`analysis-stream-middleware.ts:158-160`), and must: without them the POST ends at the `bad_request` guard (`:6575`) before the subscribe dispatch.
- Test (e)'s title (`analysis.request-log.test.ts:407`) says "(no body)" while sending `{ chapterIds: [1] }`.

### ✅ What is solid

- **Setup widening.** No try-scoped binding is read by either catch, and neither runner has a `finally`. The local `language_unset` catch still returns. Setup throws now reach the aborted / `STALE_BOOK_DIR` / `classifyAnalysisFailure` paths. The setup-throw suite goes red on the pre-fix ordering, because the promise rejects.
- **`started` gate.** Start is disabled while `!isAnalyzerReady` (`analysing.tsx:1222`), so the card can't show streaming before a POST can fire. The snapshot term and the `!!manuscriptId` guard are correct. The synchronous paused/halted assertions are well-shaped.
- **#3171.** The `not_found` return now inside the `try` is covered by `finally { clearInterval(heartbeat) }`. `endJob` clears design-busy. The regression test isolates the second read.
- **#3174 sweep.** I checked every `void` site under `server/src` myself; all are contained:
  - with a `.catch`: `index.ts:192/211`, `auto-backup.ts:260`;
  - with internal try/catch: `import.ts:453`, `generation.ts:1342`, `runAutoRevert`, `triggerScoring`;
  - callees that never reject: `updates.ts:131`, `device-tokens.ts:578`, `sidecar-supervisor.ts:355`.

  `sidecar-supervisor.ts:195`/`:470` and `design-voice-core.ts:146` are pre-existing and weren't traced to completion. `withVerifiedBookDir` in `drop` mode never throws `BookDirUnresolvedError`, so the new outer catches swallow nothing load-bearing.
- **Bookkeeping.**
  - CLAUDE.md's "thirteen sites" matches `git grep requestFailureMessage(`.
  - `Closes`/`Refs` sit outside code spans.
  - The release-notes pair is present.
  - On-box acceptance isn't applicable.

### Verdict

**Not ready to merge.** One 🟠 correctness finding: a log-forgery path this PR introduces, contradicting its own F6 hardening. Fixing and pushing it re-triggers a pass. The five 🟡 items are cleanup, fixed this round; they don't re-trigger. Also confirm the pending Windows leg lands green before merge.

dudarenok-maker:
## PR review — pass 2 (head 4123e895, depth high)

Scope: the fix delta `efdaea30..4123e895` (7 files), plus what its new text makes claims about:
- `failure-taxonomy.ts`, the stream middleware, and the subset route's validation/subscribe order;
- tracked consumers of the log shape that `JSON.stringify` changed;
- a sweep of every remaining `${manuscriptId}` log line.

Verified: `gh pr checks` shows 21/21 green at this head. I didn't re-run the suites locally. I also didn't repeat the implementers' mutation proofs, because the tree is read-only for this pass. Instead I read each changed assertion to confirm it can fail for the reason it claims.

### Prior findings: dispositions

**Pass 1 at `efdaea30`**

- 🟠 **`manuscriptId` printed raw in pre-validation log lines: VERIFIED RESOLVED.**
  - All six sites now stringify it: `analysis.ts:3265`, `:3407`, `:3503`, `:6522`, `:6636`, `:6721`.
  - Tests (g) `analysis.request-log.test.ts:557` and (h) `:600` send an id containing `%0A` through both routes, and the reported red output shows the forged `start` line.
  - The rest of the class is closed:
    - every other `${manuscriptId}` log line (`:3704`, `:3739`, `:3751`, `:6401`, `:6418`, `:6816`, `:6828`, `:8009`, `:8024`) runs only after `getOrHydrateManuscript` (`:3329` / `:6551`) returned a record;
    - `selection.engine` comes from `inferEngineFromModelId`, so it's an enum;
    - `store/manuscripts.ts` doesn't log.
  - No tracked script or doc parses the changed shape. The only hit, `docs/testing/onbox-wave11-group-c-results/step-3-c1-cloud.md:74`, is a historical capture of the removed line.
- 🟡1 **#3171 curation claim: RESOLVED, with one leftover.**
  - `docs/release-notes-next.md:337` and the PR body now say only lock timeouts are curated.
  - The new case at `single-design.test.ts:626` asserts `code: 'lock-contention'` (`:655`) and the exact `LOCK_CONTENTION_REQUEST_ERROR` (`:658`), so it fails if the raw message passes through.
  - `single-design.ts:206-208` is unchanged, but it reads correctly in context: `:204-205` directly above states the scope.
  - *Leftover (🟡):* `single-design.test.ts:583-585` still labels its path regex `// Curated: a real JSON.parse error, never a raw filesystem path`. The new case's own comment (`:628-629`) says that assertion "can't fail regardless of curation". Drop the "Curated" label or the assertion, so nobody later deletes the real test believing this one covers curation.
- 🟡2 **before-hydration proof: VERIFIED RESOLVED.**
  - The mock at `analysis.request-log.test.ts:47-59` records, as `getOrHydrateManuscript` is entered, whether `request received` has already been logged.
  - (a) and (c) assert it at `:270-273` and `:380-383`.
  - The wrapper drops no arguments: the real function takes only one (`store/manuscripts.ts:71-73`).
- 🟡3 **boot settings message: VERIFIED RESOLVED.** `index.ts:143` no longer asserts a cause, and `:121-133` lists the parse, read and migration rejections.
- 🟡4 **`RELEASE_NOTES.md` before/after: PARTIALLY RESOLVED.**
  - The upgrade half now matches `upgrade.ts:183-192`: the apply error is lost on the else and catch paths, the restart is skipped on the success path, and "applying" is stated as still stuck. The `index.ts:154-159` comment is corrected.
  - *Still open (🟡):* the boot clause "the log used to note only that something had crashed, without saying what" (`RELEASE_NOTES.md:7`) is the flagged claim reworded. It also contradicts the same bullet's opening, which says Castwright "wrote its full detail to the server log".
    - Before this PR, `crash-logging.ts:72-73` logged `[server] FATAL unhandledRejection — <err.stack> (survived — server continues)`.
    - An `EBUSY` on the settings file at boot carries the file path in that message.
    - What was actually missing is that the entry didn't name the step: the settings read, or the speech-engine start.
- 🟡5 **comments made false: items 1 and 3 VERIFIED RESOLVED.**
  - Item 1: `analysis.test.ts:8722-8729` now matches `analysis.ts:3610-3614` / `:6764-6767`.
  - Item 3: test (e)'s title is fixed (`:465`).
  - Item 2: the rewrite at `analysis.ts:6511-6515` removed "no body" but introduced a new false claim; see New 3.

**Earlier pass 1 at `f32e94e7`**

- 🟠1 **`manuscriptId` log injection: VERIFIED RESOLVED.** Same fix and tests as above.
- 🟠2 **#3171 had no release notes: VERIFIED RESOLVED** since `b24636e5` / `ea8ba067`: `RELEASE_NOTES.md:5` and `docs/release-notes-next.md:337`.
- 🟡1 **test (e) title and the subset "same three shapes" note.**
  - The title is VERIFIED RESOLVED.
  - The softened note adds a false reachability claim; see New 2.
  - *Leftover (🟡):* the PR body's Test plan still says "`subscribe` … for a body-less POST against a live job, both routes", but the subset case sends `{ chapterIds: [1] }`.
- 🟡2 **`consoleErrorSpy` assertion couldn't fail: VERIFIED RESOLVED.** It now matches the `[single-design] failed` prefix (`single-design.test.ts:593-596`).

### 🔴 Blocking
None.

### 🟠 Significant
None.

### 🟡 Minor (cleanup): new, all introduced by `4123e895` or surfaced by checking its claims

**New 1. `docs/release-notes-next.md:336` says only lock-timeout messages get rewritten. `classifyAnalysisFailure` rewrites more than that.**

The parenthetical reads "(lock-timeout messages curated; every other error still reaches the client with its own message)". The classifier also replaces the message in other branches:
- A signature hit returns the remediation copy with no raw text and no `detail` (`failure-taxonomy.ts:567-569`, via `scanSignatures` at `:348-351`).
- `AnalyzerTruncatedError` and `DailyQuotaExhaustedError` get constructed sentences (`:526-541`).
- Quota envelopes are trimmed (`:551-554`).

Scenario: `new Error("ENOSPC: no space left on device, write '…'")` has no envelope and no status, so it matches `disk-full` (`:245-250`, `source: 'both'`). The client receives "The workspace volume is out of disk space — the chapter audio could not be written." (`failure-remediations.ts:57`), not its own message.

The M5 report's survey of the classifier's branches omits the signature-hit return. Suggested wording: lock timeouts get the curated sentence; other errors go through the classifier, which may reword them.

**New 2. The same bullet describes a path the middleware can't take, and the inserted clause breaks the sentence.**

The note says a subset resubscribe with no chapter ids is "reachable for a pre-D1 snapshot without `subsetChapterIds`". It isn't:
- `analysis-stream-middleware.ts:83` picks the subset route only when `snap.kind === 'subset'`.
- A pre-D1 snapshot has no `kind` (`src/lib/types.ts:607-611`), so it resubscribes through the **main** route.
- `kind` and `subsetChapterIds` were added in the same commits (`30715f71`, `d0fdff03`).
- The server sets both on every subset job (`analysis.ts:6683-6684`).
- Every snapshot writer pairs them (`:2741-2742`, `:2851-2852`, `:2901-2902`).
- All three frontend `kind: 'subset'` literals set the ids (`analysing.tsx:813-814`, `generation.tsx:435-436`, `:604-605`).

So the `?? []` at middleware `:159` is a defensive default, not a documented path. The misreading started with the `f32e94e7` pass's reading of `types.ts:609`, which says pre-D1 snapshots omit `kind`, not `subsetChapterIds`.

The insertion also leaves the sentence reading "…logs no outcome line at all: `[analysis] start manuscript=…` … for a new job", so the shape list now attaches to the no-outcome case.

Fix: drop the reachability parenthetical, keep "the subset route logs an outcome line only once its chapter-id validation passes", and put the shape list back before that qualification.

**New 3. `server/src/routes/analysis.ts:6513-6514`: the new comment says the subscribe POST sends `chapterIds` "(needed to validate against the job's subsetChapterIds)". The route never compares them.**

- The subscribe dispatch at `:6630-6639` checks only `existing && !existing.controller.signal.aborted`.
- The route reads `subsetChapterIds` only in the job literal (`:6684`).
- The route has already flushed 200 SSE headers at `:6530`, so it can't answer 409. The file's only `res.status(409)` is `language_unset` at `:3283`.

Scenario: a subset job for chapter 3 is running, and a POST arrives with `{ chapterIds: [7] }` for an existing chapter 7. It passes `:6581-6605`, logs `[analysis-subset] subscribe`, and is attached to the chapter-3 job. There's no conflict, and the requested ids are ignored.

The claim was copied from the middleware's own comment at `analysis-stream-middleware.ts:153-155` ("mismatched would 409"), which is stale in the same way. Correct both: the ids only need to pass the route's non-empty and known-chapter checks.

**New 4. `server/src/workspace/user-settings.ts:350-352`: `readUserSettings`'s own doc comment still says it "falls back to defaults when the file is missing or malformed".**

Both this PR's corrected `index.ts:121-133` and the PR body ("The boot comment claimed a malformed `user-settings.json` 'falls through to defaults'; it doesn't") state the opposite, and the code agrees with them:
- `readJson` throws on unparseable JSON (`state-io.ts:53`), and nothing catches it, so `readUserSettings` rejects.
- Only parseable content that fails the schema falls back to defaults (`:375-376`).

The PR fixed the downstream copy of the claim and left the source copy on the function itself. Suggested wording: "missing or schema-invalid → defaults; unparseable or unreadable → rejects (see #3175)."

### Verdict

**No 🔴 or 🟠 at this head.** Every correctness finding from both earlier passes is verified resolved.

What remains is 🟡 cleanup only:
- the four new items above;
- three leftovers: the "Curated" test comment, the `RELEASE_NOTES.md` boot clause, and the PR body's "body-less … both routes" phrase.

Under the gate's rules these get fixed this round and don't trigger another review pass. CI is 21/21 green at `4123e895`.

Issue #3182, opened 2026-09-11
[agent instructions][claude][verify] Verify #3170 coqui-residency-policy guard-targets conversion and open the PR


dudarenok-maker:
## What

Verify the #3170 task child, then open the PR. Read-only for source — you
report and open the PR, you do not fix.

## Parent's acceptance, restated as a checklist

- [ ] `server/src/tts/coqui-residency-policy.guard-targets.ts` exists,
      exports `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS` (three entries),
      matches the brace-glob string style of the four sibling
      `.guard-targets.ts` modules from PR #3151.
- [ ] `coqui-residency-policy.guard.test.ts` imports from the new module
      instead of declaring the three path literals inline, and has a new
      self-consistency test (mirroring the four sibling guards' own).
- [ ] `force-rerun-triggers.test.ts`'s three `MAIN_COVERED` entries for these
      files are DERIVED from the imported constant, not re-typed literals,
      and there's a new drift-check test asserting `vitest.config.ts`'s real
      `forceRerunTriggers` array matches.
- [ ] `vitest.config.ts:229-237`'s comment names the new `.guard-targets.ts`
      module; the three glob string literals themselves are unchanged.
- [ ] `cd server && npx vitest run src/force-rerun-triggers.test.ts
      src/tts/coqui-residency-policy.guard.test.ts` passes.
- [ ] `npx tsc --noEmit` (server) passes.
- [ ] The task child's `AGENT DONE` receipt showed REAL vitest failure output
      for its mutation check, not just a claim. **Re-run any mutation whose
      receipt asserts it without showing output**: apply the same one-
      character typo to the exported constant, confirm the drift-check test
      actually reddens with the expected message, then revert.
- [ ] `git diff --stat` against `main` touches only: the new
      `.guard-targets.ts` file, `coqui-residency-policy.guard.test.ts`,
      `force-rerun-triggers.test.ts`, `vitest.config.ts`. Nothing else.

## FAIL ordering (exactly this order if ANY check above fails)

1. Create a rework `[claude][task]` child, briefed from what specifically
   failed and citing the exact command/output.
2. Create a fresh `[claude][verify]` child.
3. Link both to the parent (#3170).
4. **Only then** close this issue with the verdict.

## On PASS

Open the PR yourself:

- `Closes #3170`.
- Use `.github/pull_request_template.md`'s `## Summary` / `## Test plan`
[...]

PR #3191, opened 2026-09-11
chore(ops): convert coqui-residency-policy guard to guard-targets export shape

### LINKED ISSUE #3170: chore(ops): add coqui-residency-policy.guard.test.ts to force-rerun-triggers mechanism
Add coqui-residency-policy.guard.test.ts to the force-rerun-triggers drift-detection mechanism (#3085), same shape as cast-lock.guard.test.ts, state-language.guard.test.ts, spawn-windows-hide.guard.test.ts, and engine-language-coverage.guard.test.ts.

This guard is a runtime-scanning shape (reads files at test startup) and should export its own scan-target scope from a sibling module, with force-rerun-triggers.test.ts checking the real trigger against the exported constant.

Flagged as follow-up in PR #3151.

### LINKED ISSUE #3181: [agent instructions][claude][task] Convert coqui-residency-policy.guard.test.ts to the guard-targets export shape (#3151 follow-up)
## What

Convert `server/src/tts/coqui-residency-policy.guard.test.ts` to the same
`<name>.guard-targets.ts` sibling-module-export shape PR #3151 gave its four
sibling guards (`state-language.guard.test.ts`, `spawn-windows-hide.test.ts`,
`cast-lock.guard.test.ts`, `engine-language-coverage.guard.test.ts`), so this
guard's real runtime scan-scope and its `server/vitest.config.ts`
`forceRerunTriggers` coverage are checked against ONE shared constant instead
of two independently-maintained literal copies. PR #3151 deliberately did NOT
convert this guard, flagging it as this follow-up (#3170).

## Background you need

`coqui-residency-policy.guard.test.ts` (`server/src/tts/`) is a runtime-
scanning guard (#1932, side-18): it `readFileSync`s three files at test
startup to check a cross-reference token/heading survives — no module-graph
edge to any of them, so `vitest run --changed` needs an explicit
`forceRerunTriggers` entry per file or a diff confined to one of them selects
zero tests. Those three entries **already exist** in
`server/vitest.config.ts:235-237`:

```
'{**/server/src/tts/synthesise-chapter.ts,**/.*/**/server/src/tts/synthesise-chapter.ts}',
'{**/server/tts-sidecar/main.py,**/.*/**/server/tts-sidecar/main.py}',
'{**/docs/features/264-vram-aware-gpu-placement.md,**/.*/**/docs/features/264-vram-aware-gpu-placement.md}',
```

...and `server/src/force-rerun-triggers.test.ts:136-153` (`MAIN_COVERED`)
already has matching hardcoded entries. **What's missing is the dedup**: the
guard test's own three path constants (`SYNTHESISE_CHAPTER_PATH`/
`SIDECAR_MAIN_PATH`/`POLICY_DOC_PATH`, currently inline literals at
`coqui-residency-policy.guard.test.ts:22-23,41`) and
`force-rerun-triggers.test.ts`'s copies can drift independently — the exact
gap #3085 closed for the other four guards.

**Read PR #3151 in full before writing anything**
(`gh pr diff 3151 --repo dudarenok-maker/Castwright`), specifically
`server/src/tts/engine-language-coverage.guard-targets.ts` (closest shape —
single scan target, `_GUARD_SCAN_GLOB` singular) as your primary template,
and `server/src/spawn-windows-hide.guard-targets.ts` (plural
`_GUARD_SCAN_GLOBS`, for when a guard scans more than one tree) for how a
multi-target guard is documented. This guard has THREE independent scan
targets — closer to the plural shape, but each target here is a single file,
not a glob tree, so don't over-generalize spawn-windows-hide's tree-glob
style; follow engine-language-coverage's single-file literal-path style,
just three of them in one array.

## The fix

1. Create `server/src/tts/coqui-residency-policy.guard-targets.ts` exporting
   `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS` — an array of the three
   repo-root-relative paths, in the SAME brace-glob string style
   `server/vitest.config.ts`'s other `forceRerunTriggers` entries use (see
   `engine-language-coverage.guard-targets.ts`'s single-string style, applied
   three times). Doc comment explains why each has no module-graph edge
   (mirror the existing comment at `coqui-residency-policy.guard.test.ts:1-13`
   and `vitest.config.ts:229-234` rather than inventing new prose — that
   explanation is already correct, just needs to live behind the export).
2. Update `coqui-residency-policy.guard.test.ts` to import the three paths
   from the new module instead of declaring `SYNTHESISE_CHAPTER_PATH`/
   `SIDECAR_MAIN_PATH`/`POLICY_DOC_PATH` as local literals (the `REPO_ROOT`/
   `__dirname`-relative resolution stays local to the test — only the
   relative-path STRINGS move to the shared module). Add a self-consistency
   test mirroring the pattern in the other four guards' own `.test.ts` files
   (`gh pr diff 3151` shows each guard gained an
   `it('scan scope matches the declared X_GUARD_SCAN_GLOB(S) (#3085)', ...)`
   case) — adapt for three targets instead of one.
3. Update `server/src/force-rerun-triggers.test.ts`: import
   `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`, replace the three hardcoded
   `MAIN_COVERED` entries at lines 136-153 with entries whose `rel:` is
   DERIVED from the imported constant (not re-typed), and add a dedicated
   test asserting `server/vitest.config.ts`'s real `forceRerunTriggers` array
   contains the entry derived from each of the three constants — mirroring
   the `it('main forceRerunTriggers has the exact entry derived from
   ENGINE_LANGUAGE_COVERAGE_GUARD_SCAN_GLOB (#3085)', ...)` test PR #3151
   added (`gh pr diff 3151` line ~175).
4. Update the `vitest.config.ts:229-237` comment to cite the new
   `coqui-residency-policy.guard-targets.ts` constant by name, mirroring how
   PR #3151 updated the `engine-language-coverage` and `state-language`
   comments at `vitest.config.ts` (see the PR diff) to name their own
   `.guard-targets.ts` module and say the entry's literal text is checked
   against it. **The three glob string literals themselves in
   `vitest.config.ts` do not need to change** — only the explanatory comment
   above them needs to name the new source of truth.

**Mutation check, required, with observed output:** change one character in
`COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`'s middle entry (e.g. typo
`main.py` → `main.pyy`) and re-run
`npx vitest run src/force-rerun-triggers.test.ts` from `server/` — the new
drift-check test must FAIL, naming the mismatch. Paste the actual failure
output in your `AGENT DONE` receipt. Then revert and re-confirm green.

## Where the work happens

This is small enough for its own fresh worktree (no pre-existing work to
build on, unlike #3056). Cut it yourself first:

```bash
node scripts/wt-new.mjs chore/ops-3170-coqui-guard-scope
```

Run from `C:\Claude\Projects\Audiobook-Generator` (the primary checkout).
This creates `C:\Claude\Projects\wt-3170-coqui-guard-scope` (or whatever
`wt-new.mjs` names it — check its output) on branch
`chore/ops-3170-coqui-guard-scope`, off latest `main`, with dependencies
installed and hooks active. Work there. Do not work in the primary checkout.

## Acceptance

- `cd server && npx vitest run src/force-rerun-triggers.test.ts
  src/tts/coqui-residency-policy.guard.test.ts` passes, all tests, including
  the new drift-check and self-consistency tests.
- `npx tsc --noEmit` (server) passes.
- Mutation check above shows real FAIL output in the receipt.
- `git diff --stat` shows: new
  `server/src/tts/coqui-residency-policy.guard-targets.ts`; modified
  `coqui-residency-policy.guard.test.ts`, `force-rerun-triggers.test.ts`,
  `vitest.config.ts` (comment only, not the glob literals). Nothing else.

## Key files

- PR #3151 (`gh pr diff 3151 --repo dudarenok-maker/Castwright`) — the
  pattern for all four steps above, cite it don't reinvent it.
- `server/src/tts/coqui-residency-policy.guard.test.ts` (convert)
- `server/src/force-rerun-triggers.test.ts:136-153` (MAIN_COVERED entries
  to derive from the new export)
- `server/vitest.config.ts:229-237` (comment to update)

## Not in scope

- Changing the three glob string LITERALS in `vitest.config.ts` — they're
  already correct, only the comment needs updating.
- Touching any of the four guards PR #3151 already converted.
- PR #3151 itself — separate branch, don't rebase onto it or wait for it;
  this is an independent, parallel change.

## Commit/push gate

Commit and push normally in the foreground — this branch touches no
`server/tts-sidecar/**` files. Never `--no-verify`.

## On completion

1. Commit and push to the same branch.
2. Flip the successor (#3182, the verify child) to `Agent Todo` FIRST:

```bash
ITEM_ID=$(gh api graphql -F owner=dudarenok-maker -F repo=Castwright -F num=3182 -f query='query($owner:String!,$repo:String!,$num:Int!){ repository(owner:$owner,name:$repo){ issue(number:$num){ projectItems(first:10){ nodes{ id } } } } }' --jq '.data.repository.issue.projectItems.nodes[0].id')
gh project item-edit --id "$ITEM_ID" --project-id PVT_kwHOEOX6_c4Bcf9a --field-id PVTSSF_lAHOEOX6_c4Bcf9azhXISes --single-select-option-id 9d9fa565
```

3. THEN leave `AGENT DONE` (with the mutation-check output and the worktree
   path/branch you actually used) and set THIS issue's Status to
   `Agent Review`. Do not close this issue yourself.

dudarenok-maker:
## Summary
- Adds `server/src/tts/coqui-residency-policy.guard-targets.ts` exporting `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`, the three repo-root-relative paths `coqui-residency-policy.guard.test.ts` reads at runtime.
- `coqui-residency-policy.guard.test.ts` now imports those paths instead of declaring them as local literals, plus a self-consistency test tying its actual scan targets to the declared constant.
- `force-rerun-triggers.test.ts`'s three `MAIN_COVERED` entries for this guard now derive `rel` from the imported constant, plus a drift-check test pinning `server/vitest.config.ts`'s literal `forceRerunTriggers` entries to it.
- `server/vitest.config.ts`'s comment above those three entries now names the new `guard-targets.ts` module as the source of truth (the glob literals themselves are unchanged).

Follow-up to #3151 (PR #3151 deliberately did not convert this guard) and #3085/#1932 side-18.

Closes #3170
Closes #3181

## Release notes
N/A — internal guard/test infrastructure with no user- or operator-visible effect.

## Merge coordination (blocking with #3151)
**This PR has a merge interaction with still-open #3151.** Both edit `force-rerun-triggers.test.ts` and `vitest.config.ts`. #3151 adds a hand-typed assertion for the same `synthesise-chapter.ts` string this PR's constant derives. After both merge, that entry becomes a second independently-maintained literal copy — the opposite of what this PR exists to close. 

If #3151 has already merged: rebase this PR onto main and eliminate the duplicate literal it introduced.

If #3151 is still open: whichever merges first, the second must delete the now-redundant hand-typed copy on merge (the `vitest.config.ts` half auto-merges silently, so it won't conflict-warn). Exact entry to look for: the coquiSynthesiseEntry literal in #3151's force-rerun-triggers.test.ts, which duplicates the brace-glob built from GLOBS[0].

## Mutation check (required by #3170/#3181, independently re-verified by #3182)
Changed `server/tts-sidecar/main.py` → `server/tts-sidecar/main.pyy` in the guard-targets constant, re-ran `npx vitest run src/force-rerun-triggers.test.ts src/tts/coqui-residency-policy.guard.test.ts` from `server/`:

```
FAIL  src/force-rerun-triggers.test.ts > server/vitest.config.ts forceRerunTriggers > main forceRerunTriggers has the exact entry derived from COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS entry server/tts-sidecar/main.pyy (#3085, #3151 follow-up)
AssertionError: expected [ …(14) ] to include '{**/server/tts-sidecar/main.pyy,**/.*…'
```
Also failed the 3 file-coverage cases and the guard's own `main.py` cross-reference case for the same entry (existsSync/ENOENT — the mutated path doesn't exist). Reverted; re-ran green (79/79).

## Test plan
- [x] `cd server && npx vitest run src/force-rerun-triggers.test.ts src/tts/coqui-residency-policy.guard.test.ts` — 79/79 pass
- [x] `npx tsc --noEmit` (server) — clean
- [x] Mutation check above shows real FAIL output, reverted, reconfirmed green
- [x] `git diff --stat` matches acceptance: new `coqui-residency-policy.guard-targets.ts`; modified `coqui-residency-policy.guard.test.ts`, `force-rerun-triggers.test.ts`, `vitest.config.ts` (comment only)
- [x] Independently re-verified by #3182 (verify child): re-ran the vitest suite, tsc, and the mutation check from a clean worktree state; all matched the claims above exactly.

### DIFF
--- a/scripts/verify-cache.mjs
+++ b/scripts/verify-cache.mjs
@@ -470,16 +470,17 @@ export const STEPS = [
         'server/package.json',
         'e2e/global-teardown.ts',
         '.gitattributes',
-        /* #1932 (side-18): coqui-residency-policy.guard.test.ts reads THREE of these at RUNTIME
+        /* #1932 (side-18): coqui-residency-policy.guard.test.ts reads TWO of these at RUNTIME
            to guard cross-reference rot across Coqui eviction mechanisms and their policy doc
            — no module-graph edge, same #1847 runtime-read trap as openapi.yaml/scripts/** above.
            This is item 5 of that guard's own NOTE at server/src/tts/coqui-residency-policy.guard.test.ts.
            The three paths are DECLARED via COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS
            (server/src/tts/coqui-residency-policy.guard-targets.ts), the single source of truth
            for all three locations (server/vitest.config.ts, force-rerun-triggers.test.ts, and
-           this entry). This entry fails SILENTLY (hashFile returns a sentinel for a missing path
-           rather than throwing), unlike the test's own POLICY_DOC_PATH which throws on readFileSync. */
-        'server/src/tts/synthesise-chapter.ts',
+           this entry). synthesise-chapter.ts is inert here (covered by server/src/** glob already),
+           while main.py and the policy doc are not covered by any glob and would fail SILENTLY
+           (hashFile returns a sentinel for a missing path rather than throwing), unlike the
+           test's own POLICY_DOC_PATH which throws on readFileSync. */
         'server/tts-sidecar/main.py',
         'docs/features/264-vram-aware-gpu-placement.md',
       ],
--- a/scripts/verify-cache.mjs
+++ b/scripts/verify-cache.mjs
@@ -470,12 +470,16 @@ export const STEPS = [
         'server/package.json',
         'e2e/global-teardown.ts',
         '.gitattributes',
-        /* #1932 (side-18): coqui-residency-policy.guard.test.ts (server suite) reads both
-           of these at RUNTIME to guard cross-reference rot across Coqui eviction mechanisms
-           and their policy doc — no module-graph edge, same #1847 runtime-read trap as
-           openapi.yaml/scripts/** above. Without these entries, a diff confined to either
-           (exactly the shape that could break the Coqui eviction contract) reports [cached]
-           and the guard never re-runs. */
+        /* #1932 (side-18): coqui-residency-policy.guard.test.ts reads THREE of these at RUNTIME
+           to guard cross-reference rot across Coqui eviction mechanisms and their policy doc
+           — no module-graph edge, same #1847 runtime-read trap as openapi.yaml/scripts/** above.
+           This is item 5 of that guard's own NOTE at server/src/tts/coqui-residency-policy.guard.test.ts.
+           The three paths are DECLARED via COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS
+           (server/src/tts/coqui-residency-policy.guard-targets.ts), the single source of truth
+           for all three locations (server/vitest.config.ts, force-rerun-triggers.test.ts, and
+           this entry). This entry fails SILENTLY (hashFile returns a sentinel for a missing path
+           rather than throwing), unlike the test's own POLICY_DOC_PATH which throws on readFileSync. */
+        'server/src/tts/synthesise-chapter.ts',
         'server/tts-sidecar/main.py',
         'docs/features/264-vram-aware-gpu-placement.md',
       ],
--- a/server/src/force-rerun-triggers.test.ts
+++ b/server/src/force-rerun-triggers.test.ts
@@ -193,13 +193,10 @@ describe('server/vitest.config.ts forceRerunTriggers', () => {
   );
 
   /* #3085/#3151 follow-up: the file-coverage case above only proves the
-     CURRENT scope is covered — narrowing one of
-     COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS's entries to a different path
-     would still pass it, because the (unchanged) trigger below matches a
-     superset. This assertion checks the trigger array contains the EXACT
-     brace-glob built from each imported constant, so narrowing OR widening
-     the constant without updating vitest.config.ts's literal entry to match
-     is caught either way. */
+     CURRENT scope is covered. This assertion checks the trigger array contains
+     the EXACT brace-glob built from each imported constant, so narrowing OR
+     widening the constant without updating vitest.config.ts's literal entry
+     to match is caught either way — a drift in either direction fails this test. */
   it.each(COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)(
     'main forceRerunTriggers has the exact entry derived from COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS entry %s (#3085, #3151 follow-up)',
     (glob) => {
--- a/server/src/force-rerun-triggers.test.ts
+++ b/server/src/force-rerun-triggers.test.ts
@@ -29,6 +29,7 @@ import picomatch from 'picomatch';
 import { existsSync } from 'node:fs';
 import { relative, resolve } from 'node:path';
 import { pathToFileURL } from 'node:url';
+import { COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS } from './tts/coqui-residency-policy.guard-targets.js';
 
 const SERVER_ROOT = resolve(__dirname, '..');
 const REPO_ROOT = resolve(SERVER_ROOT, '..');
@@ -139,16 +140,23 @@ const MAIN_COVERED = [
      module-graph edges (same #1847 runtime-read trap); synthesise-chapter.ts
      IS importable but is included here for consistency with the guard's uniform
      readFileSync approach rather than being split into two separate tracking
-     mechanisms. */
-  { rel: 'src/tts/synthesise-chapter.ts', file: 'the Node Coqui eviction mechanism', base: SERVER_ROOT },
+     mechanisms. Each `rel` below is DERIVED from
+     COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS (#3085, #3151 follow-up), the
+     same constant the guard itself declares, not a second hand-typed literal
+     that happens to agree with it today. */
   {
-    rel: 'server/tts-sidecar/main.py',
-    file: 'the sidecar Coqui eviction mechanism (via the server/tts-sidecar/main.py trigger)',
+    rel: COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[0],
+    file: 'the Node Coqui eviction mechanism (COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)',
     base: REPO_ROOT,
   },
   {
-    rel: 'docs/features/264-vram-aware-gpu-placement.md',
-    file: 'the Coqui residency policy doc',
+    rel: COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[1],
+    file: 'the sidecar Coqui eviction mechanism (COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)',
+    base: REPO_ROOT,
+  },
+  {
+    rel: COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[2],
+    file: 'the Coqui residency policy doc (COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)',
     base: REPO_ROOT,
   },
   /* #3059 — engine-language-coverage.guard.test.ts's third assertion reads
@@ -181,6 +189,22 @@ describe('server/vitest.config.ts forceRerunTriggers', () => {
     'covers $file from $shape so a config-only diff still forces a full --changed run',
     ({ rel, base, root }) => {
       expect(matchesTrigger(mainTriggers, absPathUnder(root, base, rel))).toBe(true);
+    },
+  );
+
+  /* #3085/#3151 follow-up: the file-coverage case above only proves the
+     CURRENT scope is covered — narrowing one of
+     COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS's entries to a different path
+     would still pass it, because the (unchanged) trigger below matches a
+     superset. This assertion checks the trigger array contains the EXACT
+     brace-glob built from each imported constant, so narrowing OR widening
+     the constant without updating vitest.config.ts's literal entry to match
+     is caught either way. */
+  it.each(COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)(
+    'main forceRerunTriggers has the exact entry derived from COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS entry %s (#3085, #3151 follow-up)',
+    (glob) => {
+      const expected = `{**/${glob},**/.*/**/${glob}}`;
+      expect(mainTriggers).toContain(expected);
     },
   );
 
--- a/server/src/tts/coqui-residency-policy.guard-targets.ts
+++ b/server/src/tts/coqui-residency-policy.guard-targets.ts
@@ -1,0 +1,18 @@
+/** The real scan scope of coqui-residency-policy.guard.test.ts (#1932,
+    side-18): it `readFileSync`s these three files at RUNTIME to check a
+    cross-reference token/heading survives, rather than importing them —
+    `main.py` and the policy doc have no module-graph edge for that reason
+    (the same #1847 runtime-read trap), and `synthesise-chapter.ts` is
+    included here too for consistency with the guard's uniform readFileSync
+    approach rather than being split into two separate tracking mechanisms.
+    Each entry is a repo-root-relative path, in the same brace-glob style
+    `server/vitest.config.ts`'s other `forceRerunTriggers` entries use, so
+    this constant is the single source of truth both
+    `server/vitest.config.ts` and `force-rerun-triggers.test.ts` check
+    against, instead of two independently-maintained literal copies (#3085,
+    #3151 follow-up). */
+export const COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS = [
+  'server/src/tts/synthesise-chapter.ts',
+  'server/tts-sidecar/main.py',
+  'docs/features/264-vram-aware-gpu-placement.md',
+] as const;
--- a/server/src/tts/coqui-residency-policy.guard.test.ts
+++ b/server/src/tts/coqui-residency-policy.guard.test.ts
@@ -14,13 +14,14 @@
 
 import { describe, it, expect } from 'vitest';
 import { readFileSync } from 'node:fs';
-import { join } from 'node:path';
+import { join, relative, sep } from 'node:path';
+import { COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS } from './coqui-residency-policy.guard-targets.js';
 
 const TOKEN = 'COQUI-RESIDENCY-POLICY';
 const REPO_ROOT = join(process.cwd(), '..');
 
-const SYNTHESISE_CHAPTER_PATH = join(process.cwd(), 'src', 'tts', 'synthesise-chapter.ts');
-const SIDECAR_MAIN_PATH = join(process.cwd(), 'tts-sidecar', 'main.py');
+const SYNTHESISE_CHAPTER_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[0]);
+const SIDECAR_MAIN_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[1]);
 /* NOTE: When plan 264 is archived (status → stable), this path will move to
    docs/features/archive/264-vram-aware-gpu-placement.md. Keep the path
    hardcoded so the guard fails-closed if archival forgets to update it.
@@ -38,13 +39,26 @@ const SIDECAR_MAIN_PATH = join(process.cwd(), 'tts-sidecar', 'main.py');
       forgotten — hashFile returns a sentinel for a missing path rather than
       throwing, so a docs-only PR would just silently stop invalidating the
       verify cache, reopen the bug this PR #2715 fixed, and never error) */
-const POLICY_DOC_PATH = join(REPO_ROOT, 'docs', 'features', '264-vram-aware-gpu-placement.md');
+const POLICY_DOC_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[2]);
 
 function occurrences(haystack: string, needle: string): number {
   return haystack.split(needle).length - 1;
 }
 
 describe('Coqui residency policy cross-references (side-18, #1932)', () => {
+  it('scan scope matches the declared COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS (#3085, #3151 follow-up)', () => {
+    // Ties this guard's ACTUAL scan targets to the scope it DECLARES via the
+    // sibling module — the same constant force-rerun-triggers.test.ts checks
+    // its forceRerunTriggers entries against — so the two statements of this
+    // guard's scope can never independently drift.
+    const toRepoRel = (p: string) => relative(REPO_ROOT, p).split(sep).join('/');
+    expect([
+      toRepoRel(SYNTHESISE_CHAPTER_PATH),
+      toRepoRel(SIDECAR_MAIN_PATH),
+      toRepoRel(POLICY_DOC_PATH),
+    ]).toEqual([...COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS]);
+  });
+
   it('keeps the cross-reference in synthesise-chapter.ts (mechanism A)', () => {
     const source = readFileSync(SYNTHESISE_CHAPTER_PATH, 'utf8');
     expect(
--- a/server/src/tts/coqui-residency-policy.guard.test.ts
+++ b/server/src/tts/coqui-residency-policy.guard.test.ts
@@ -20,10 +20,14 @@ import { COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS } from './coqui-residency-polic
 const TOKEN = 'COQUI-RESIDENCY-POLICY';
 const REPO_ROOT = join(process.cwd(), '..');
 
-const SYNTHESISE_CHAPTER_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[0]);
-const SIDECAR_MAIN_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[1]);
-/* NOTE: When plan 264 is archived (status → stable), this path will move to
-   docs/features/archive/264-vram-aware-gpu-placement.md. Keep the path
+/* These three paths are the ACTUAL scan targets of this guard (what it reads
+   at runtime). They are hardcoded independently so the test at lines 49-60
+   can compare them against the DECLARED scan scope (COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)
+   and have a second independent side — the two statements of this guard's
+   scope must never independently drift from each other.
+
+   When plan 264 is archived (status → stable), this path will move to
+   docs/features/archive/264-vram-aware-gpu-placement.md. Keep all three
    hardcoded so the guard fails-closed if archival forgets to update it.
    On archival, update all FIVE of these in the same commit. ITEMS 1-5 all fail
    SILENTLY when archival forgets them (stale comments stay stale, CI entries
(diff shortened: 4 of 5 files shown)


dudarenok-maker:
## PR review — pass 1 (head 7b5a85e9ce6c6ba024353643609e977759f2a80d, depth low)

Scope: all 4 changed files (`server/src/tts/coqui-residency-policy.guard-targets.ts`, `server/src/tts/coqui-residency-policy.guard.test.ts`, `server/src/force-rerun-triggers.test.ts`, `server/vitest.config.ts`), reviewed in an isolated worktree off `pull/3191/head`.

Verified independently rather than assumed: the guard's path arithmetic re-executed standalone against six candidate constant values; the branch's real 14-entry `forceRerunTriggers` array replayed through the same `picomatch` build the suite uses; `git merge-tree` run against still-open #3151; sibling pattern read from #3151's actual diff. **Not** run: the server vitest suite (no `node_modules` in the review worktree) — cloud `verify.yml` green and the PR body's mutation transcript are taken as the evidence it passes. The findings below are about what that green does *not* prove.

### 🟠 Significant — the new guard-side "scan scope matches" test cannot fail

`server/src/tts/coqui-residency-policy.guard.test.ts:49-60`

The test compares `toRepoRel(SYNTHESISE_CHAPTER_PATH | SIDECAR_MAIN_PATH | POLICY_DOC_PATH)` against `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`. But this PR also rewrote those three bindings to be *derived from that same constant* (lines 23, 24, 42):

```ts
const SYNTHESISE_CHAPTER_PATH = join(REPO_ROOT, COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS[0]);
```

So the assertion reduces to `relative(R, join(R, x)) === x`, which holds for any normalised relative path. There are no longer "two statements of this guard's scope" for it to compare — there is one, compared to itself.

Measured, replicating lines 21/23/24/42/54-59 exactly and varying only the constant:

```
PASSES  <- REAL (as shipped)
PASSES  <- ALL THREE WRONG files   ['server/src/tts/model-keys.ts','server/tts-sidecar/run-tests.ps1','docs/features/INDEX.md']
PASSES  <- NONEXISTENT paths       ['does/not/exist.ts','nope/nope.py','nope/nope.md']
PASSES  <- GLOBS, not files        ['server/src/**','**/*.py','docs/**']
PASSES  <- scope EMPTIED to a dir  ['server','server','docs']
```

Concrete scenario: point all three entries at the wrong files and this spec stays green while reporting `scan scope matches the declared COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`. The PR body's own mutation transcript corroborates this — it attributes the `main.pyy` failure to the three file-coverage cases and to a `readFileSync` ENOENT, and never to this test.

This also inverts the #3151 pattern the change cites. In every #3151 sibling the guard keeps its real binding independent of the constant, so the comparison genuinely has two sides:

- `engine-language-coverage.guard.test.ts` — `GENERATION_ROUTE_PATH = join(__dirname, '..', 'routes', 'generation.ts')`
- `cast-lock.guard.test.ts` / `state-language.guard.test.ts` — `SRC_ROOT = join(__dirname, '..')`
- `spawn-windows-hide.test.ts` — `SRC_ROOT`, `EXTERNAL_SCAN_ROOTS`, `EXTERNAL_FILES_MANUAL`, `EXTERNAL_FILES_FLOOR`

Each of those fails if the constant alone moves. This one does not. Two ways out, both implementations of one agreed behaviour (so not a design-pass carve-out): keep the three path bindings independent so the test regains a second side, or drop the test and state in the header that single-sourcing makes guard↔constant drift structurally impossible. What is not acceptable is a green spec whose name asserts a check it does not perform.

### 🟠 Significant — merge interaction with still-open #3151 reinstates the literal copy this PR removes

#3151 is still `OPEN`, so the pattern this PR follows is not on `main`, and both PRs edit the same two files. `git merge-tree --write-tree review-pr-3191-pass1 review-pr-3151-ref`:

```
Auto-merging server/src/force-rerun-triggers.test.ts
CONFLICT (content): Merge conflict in server/src/force-rerun-triggers.test.ts
Auto-merging server/vitest.config.ts
```

`vitest.config.ts` auto-merges silently, so nothing forces a look at it. More importantly, #3151 adds a **hand-typed** exact-entry assertion for the identical string this PR derives:

```ts
// #3151
const coquiSynthesiseEntry = '{**/server/src/tts/synthesise-chapter.ts,**/.*/**/server/src/tts/synthesise-chapter.ts}';
expect(mainTriggers).toContain(coquiSynthesiseEntry);
```

`it.each(COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS)` at `force-rerun-triggers.test.ts:203-209` builds exactly that string for `GLOBS[0]`. After both land, `force-rerun-triggers.test.ts` holds a hand-typed literal copy of the one string this PR's constant is meant to be the single source of truth for — the PR body's "instead of two independently-maintained literal copies" becomes false.

Concrete: rename `synthesise-chapter.ts` and update the constant. The derived `it.each` follows automatically; #3151's hand-typed assertion goes red; the fix is to hand-edit the literal — the exact drift class this PR exists to close. Whichever merges second needs to delete the other's now-redundant assertion, and that requirement is recorded in neither PR body.

(Flagging at 🟠 rather than 🟡 because the post-merge state actively violates the invariant this PR ships, and the dangerous half — `vitest.config.ts` — merges without a conflict marker to prompt review. Downgrade if triage prefers to treat it purely as merge ordering.)

### 🟡 Minor — the drift test's justification comment describes a trigger set this branch does not have

`server/src/force-rerun-triggers.test.ts:195-202` says narrowing a `GLOBS` entry "to a different path would still pass it, because the (unchanged) trigger below matches a superset."

Replayed against this branch's real 14-entry `forceRerunTriggers` with the repo's own `picomatch`:

```
matches NOTHING          <- server/src/tts/model-keys.ts
matches NOTHING          <- docs/features/INDEX.md
matches NOTHING          <- server/tts-sidecar/run-tests.ps1
matches NOTHING          <- server/src/index.ts
matches NOTHING          <- CONTRIBUTING.md
STILL MATCHES a trigger  <- scripts/verify-cache.mjs
```

Those narrowings already fail the file-coverage case today. The three coqui triggers are exact-path globs with no superset relationship; the sentence only becomes true once #3151 adds `{**/server/src/**,...}`. The test is worth keeping — the real gap it closes is narrowing to a path covered by a *different* existing trigger, e.g. `scripts/verify-cache.mjs` — but the stated mechanism is not the one in effect on this branch.

### 🟡 Minor — the archival NOTE's enumerated checklist is made false by this change

`server/src/tts/coqui-residency-policy.guard.test.ts:25-41` still reads "update all FIVE of these" and "ITEMS 1-5 all fail SILENTLY … The ONE exception that fails LOUDLY is this file's own POLICY_DOC_PATH constant". Two of its claims are no longer true after this diff:

- **Item 4** (`force-rerun-triggers.test.ts` MAIN_COVERED row) is not a separate thing to update any more — it is derived from the constant at lines 148/153/158.
- **Item 3** (`vitest.config.ts` forceRerunTriggers entry) no longer fails silently; this PR's own new drift test is what makes it fail loudly, so the "ONE exception" sentence is contradicted by the test added three lines away.

A comment the change made false is a chore under CLAUDE.md, not taste.

### 🟡 Minor — a third hand-typed copy is left unpinned and unnamed

`scripts/verify-cache.mjs:479-480` hand-types two of the three paths:

```js
'server/tts-sidecar/main.py',
'docs/features/264-vram-aware-gpu-placement.md',
```

This is item 5 of the guard's own NOTE, documented there as failing **silently** (`hashFile` returns a sentinel for a missing path rather than throwing). Neither `coqui-residency-policy.guard-targets.ts`'s header nor the `vitest.config.ts` comment mentions it, and both say the constant is "the single source of truth". #3151's own `spawn-windows-hide.guard-targets.ts` sets the precedent for handling exactly this — it names its residual gap explicitly ("This gap is pre-existing … and is named here rather than silently implied closed"). Pin it or name it; leaving the loudest-claiming header silent about the one copy that fails quietly is the wrong half to omit.

### 🟡 Minor — no release-notes line in the PR body

CLAUDE.md before-shipping step 5 wants the `docs/release-notes-next.md` + `RELEASE_NOTES.md` pair **or an explicit not-applicable**. Sibling #3151's body carries "N/A — internal test/guard infrastructure with no user- or operator-visible effect"; this body has Summary / Mutation check / Test plan and no such line. The change is genuinely N/A — the fix is the sentence, not the files.

### ✅ What is solid

- **The `force-rerun-triggers.test.ts` drift test (lines 203-209) is the real one**, and it earns the PR. It is the only assertion here that pins `vitest.config.ts`'s literal entries to the constant, and the PR body's transcript shows it going red for the reason claimed.
- **The `MAIN_COVERED` rewrite preserves resolution.** Entry 0's `base` correctly moved `SERVER_ROOT` → `REPO_ROOT` alongside its `rel`, so `absPathUnder` composes the same absolute path; easy to get wrong, and it isn't.
- **The three `file:` descriptions stayed aligned with their paths** across the reorder — a classic off-by-one spot, clean here.
- **The fail-closed property survives.** `POLICY_DOC_PATH` still resolves to a literal path and `readFileSync` still throws if plan 264 is archived without updating it, which the NOTE depends on.
- **No new runtime-read trap introduced by the fix itself** — the new module has real module-graph edges to both test files, so `--changed` selects them on a change to it.
- `Closes #3170` / `Closes #3181` both present, plain text, outside code spans.
- Mutation check documented with real FAIL output and independently re-verified by #3182 — the right discipline, and it is what let me confirm finding 1 from the PR's own evidence.

### Verdict

**Changes requested.** The headline self-consistency assertion — the one new test on the guard side — cannot fail, and its name and comment claim a drift check it does not perform; that is this repo's most-repeated defect shape, landing inside a guard written to prevent drift. Separately, merging alongside still-open #3151 reinstates the hand-typed literal this PR exists to eliminate, via a `vitest.config.ts` auto-merge that produces no conflict marker. The four 🟡 items are cheap documentation/bookkeeping chores. The `force-rerun-triggers` drift test itself is sound and should survive whatever the fix to finding 1 looks like.

dudarenok-maker:
## PR review — pass 2 (head a23d318940aa967c4e5ae0710b983f028f7e5b0a, depth low)

**Scope.** Each of pass 1's six findings re-derived against the head tree by replaying the failure scenario it named, not by reading whether the code changed. This pass had `node_modules` junctioned into the review worktree, so unlike pass 1 the suites were actually executed rather than inferred from CI:

- `npx vitest run src/force-rerun-triggers.test.ts src/tts/coqui-residency-policy.guard.test.ts` — **79/79 pass**
- `node --test scripts/tests/verify-cache.test.mjs` — **142/142 pass** (the fix touches `verify-cache.mjs`)
- `node --test scripts/tests/workflow-wiring.test.mjs` — **21/21 pass**
- `npx tsc --noEmit` (server) — exit 0; `npx eslint` over all 5 changed files — exit 0
- All 22 PR checks green at this head.

Findings 1 and NEW-1 were settled by **measurement** against the real modules, not by inspection. `git merge-tree` re-run against #3151 at its current head. Tree verified clean before and after (`HEAD` = `a23d3189`, `git status --porcelain` empty both times); no tracked file modified.

---

**Finding 1 (🟠 the guard-side "scan scope matches" test cannot fail) — VERIFIED RESOLVED.**
`server/src/tts/coqui-residency-policy.guard.test.ts:46-48, 55-66`

The three bindings are now independent hardcoded literals (`join(REPO_ROOT, 'server/src/tts/synthesise-chapter.ts')` etc.), so `toEqual` genuinely has two sides. Re-ran pass 1's exact measurement — lines 21/46-48/60-65 replicated verbatim, varying **only** the constant — and every mutant that passed before now fails:

```
actual (left side of toEqual): ["server/src/tts/synthesise-chapter.ts","server/tts-sidecar/main.py","docs/features/264-vram-aware-gpu-placement.md"]
PASSES  <- REAL (as shipped)
FAILS   <- ALL THREE WRONG files            (all PASSED in pass 1)
FAILS   <- NONEXISTENT paths                (all PASSED in pass 1)
FAILS   <- GLOBS, not files                 (all PASSED in pass 1)
FAILS   <- scope EMPTIED to a dir           (all PASSED in pass 1)
FAILS   <- ONE entry drifted (archival shape)
FAILS   <- main.pyy typo (PR-body mutant)
FAILS   <- REORDERED (same members)
FAILS   <- a FOURTH entry appended
```

This is now the #3151 sibling shape (guard keeps its real binding independent of the constant), and it catches reordering and arity changes as well as substitution. The module-graph edge survives — `guard.test.ts:18` still imports the constant, so `--changed` on `guard-targets.ts` still selects this file.

**Finding 2 (🟠 merge interaction with still-open #3151) — PARTIALLY RESOLVED; the half that matters is still open. Downgraded to 🟡.**

Re-checked at this head. #3151 is still `OPEN` (head `eee76ba7`, untouched since 2026-09-11T03:41Z — i.e. it has not moved since pass 1), and `git merge-tree --write-tree review-pr-3191-pass2 <3151 head>` still reports the same shape:

```
CONFLICT (content): Merge conflict in server/src/force-rerun-triggers.test.ts
Auto-merging server/vitest.config.ts
```

The new "Merge coordination" section is **accurate**, down to the symbol: `coquiSynthesiseEntry` is real and still present at `force-rerun-triggers.test.ts:302` on #3151's head, and it is byte-identical to what this PR's `it.each` derives from `GLOBS[0]`. Two things earn a downgrade from 🟠: the code fix is genuinely unavailable here (the line does not exist on `main` or on this branch, so there is nothing to delete), and the duplicate literal lands in the file that **hard-conflicts**, not in the silently-auto-merging one — pass 1's severity rested partly on the collision being invisible, and it is not.

What is still open is the reach. **#3151's body and comments contain no reciprocal note** (verified: its body ends at `Closes #3085`; none of its three review comments mention #3191). A PR body is read on *its own* PR and does not survive into the tree or the merge commit. If #3191 merges first — the likely order, since it is the one under active review — #3151's lane resolves that conflict with this note nowhere in view, and the default resolution of a conflict is "keep both sides", which is exactly the wrong outcome.

Concrete close, cheap and durable, either or both: (a) `gh pr comment 3151` with the same paragraph, so the note reaches the lane that has to act on it; (b) put one line inside `force-rerun-triggers.test.ts` next to the `it.each` — "these entries are derived from `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS`; do not add a hand-typed exact-entry assertion for them" — since that is the file the second merger is forced to open. Not a design pass: one agreed behaviour (the derived `it.each` survives, the hand-typed copy goes), only the delivery to choose.

**Finding 3 (🟡 drift test's justification comment describes a trigger set this branch does not have) — VERIFIED RESOLVED.**
`server/src/force-rerun-triggers.test.ts:195-199`

The false "because the (unchanged) trigger below matches a superset" clause is gone. What remains ("the file-coverage case above only proves the CURRENT scope is covered … a drift in either direction fails this test") is true of this branch as shipped — confirmed against the real 14-entry array. The justification is thinner than it could be (it no longer names *why* the file-coverage case is insufficient, which is narrowing to a path a *different* existing trigger already covers, e.g. `scripts/verify-cache.mjs`), but thinner is not false, and 🟡 is satisfied by removing the untrue claim.

**Finding 4 (🟡 the archival NOTE's enumerated checklist is made false) — PARTIALLY RESOLVED; still open.**
`server/src/tts/coqui-residency-policy.guard.test.ts:32-45`

Item 4 was reworded. Three claims in the same comment block are still wrong, and one of them is now contradicted three lines below itself:

- **"ITEMS 1-5 all fail SILENTLY … The ONE exception that fails LOUDLY is this file's own POLICY_DOC_PATH constant"** (lines 32-37) vs. **item 4's own new parenthetical "(will fail if GLOBS changes)"** (line 41). The umbrella sentence and the item it summarises now assert opposite things, in one comment.
- **Item 3 is also loud now**, and still classified silent. On archival, if `guard-targets.ts` moves and `vitest.config.ts`'s literal does not, the `it.each` at `force-rerun-triggers.test.ts:200-206` goes red. That is the whole point of the test this PR adds; the NOTE still describes the pre-PR world.
- **`guard-targets.ts` is absent from the list entirely**, despite being a mandatory archival edit. Concrete scenario: follow this checklist *exactly* on archival — update items 1-5 plus `POLICY_DOC_PATH` — and the suite is red, because `COQUI_RESIDENCY_POLICY_GUARD_SCAN_GLOBS` still names `docs/features/264-…` while the guard's own literal names `docs/features/archive/264-…`, so the line-55 scan-scope test fails. A checklist whose faithful execution leaves the build red is the defect. (Update the constant instead and item 3 goes red for the mirror reason.) The NOTE says "update all FIVE of these"; it is six.

Minor, same block: the retained sentence "When plan 264 is archived … **this path** will move to …" now sits above three constants rather than directly above `POLICY_DOC_PATH`, so "this path" no longer has a referent.

**Finding 5 (🟡 a third hand-typed copy left unpinned and unnamed) — RESOLVED on the letter; see NEW-1 for what the fix introduced.**
`scripts/verify-cache.mjs:473-484`

Pass 1 asked to pin it or name it. It is named, at the site itself and as item 5 of the guard's NOTE, including the reason it fails silently. That discharges the finding. (`guard-targets.ts`'s own header still enumerates only two consumers — `vitest.config.ts` and `force-rerun-triggers.test.ts` — and still says "single source of truth" without naming the `verify-cache.mjs` gap, which is the site pass 1 pointed at first; naming it at the gap instead is a defensible read of "name it", so I am not holding this open.)

**Finding 6 (🟡 no release-notes line) — VERIFIED RESOLVED.** The body now carries "## Release notes — N/A — internal guard/test infrastructure with no user- or operator-visible effect."

---

### New findings

**🟡 NEW-1 — the fix added an inert third literal to `verify-cache.mjs`, enlarging the exact surface this PR exists to shrink.**
`scripts/verify-cache.mjs:482`

The fix commit added `'server/src/tts/synthesise-chapter.ts'` to `test:server`'s `extraFiles`. That path is already inside the step's own `server/src/**` glob (`verify-cache.mjs:417`), so the entry changes nothing. Measured against the real module, stripping the entry and re-running `stepTouchedByDiff` on a diff confined to that one file — with the other two entries as a control group, to distinguish "this one is redundant" from "extraFiles does nothing here":

```
test:server globs: ["server/src/**","*.{mjs,ts}","scripts/**/*.{mjs,cjs,js}", …]
selected WITH new extraFiles entry   : true
selected WITHOUT new extraFiles entry: true      <- inert
control: server/tts-sidecar/main.py                        -> with=true without=false   <- load-bearing
control: docs/features/264-vram-aware-gpu-placement.md     -> with=true without=false   <- load-bearing
```

Two reasons this is not merely harmless. First, it runs against this block's own documented practice, stated eight lines above it: "launch.mjs and vite.config.ts used to be named here too, but the `*.{mjs,ts}` root glob above now covers both" — entries this file makes redundant get *removed*, and `workflow-wiring.test.mjs:745` records a prior round doing exactly that pruning. Second, pass 1 counted two hand-typed copies of these paths outside the constant; this makes three, and the comment directly above now claims the constant is "the single source of truth for all three locations … and this entry" while nothing pins this entry to it. Concrete: rename `synthesise-chapter.ts` and update the constant — the guard test, the drift test and the three `MAIN_COVERED` rows all follow or fail loudly; this literal goes stale silently (`hashFile` returns a sentinel, it does not throw), and because it was never load-bearing, no behaviour changes to surface it. It is a copy that can only ever rot. Delete it, or if it is kept deliberately for symmetry, say in the comment that it is glob-covered and inert rather than describing it as one of three entries doing the same job.

**🟡 NEW-2 — the new comment cites line numbers, in a file whose own header forbids exactly that, and the citation was already wrong when written.**
`server/src/tts/coqui-residency-policy.guard.test.ts:24`

The added text reads "They are hardcoded independently so the test at **lines 49-60** can compare them against the DECLARED scan scope". The test is at lines **55-66** (`it('scan scope matches …')` through its closing `});`); line 49 is blank and 50-52 are `function occurrences`. The same file's header, lines 11-13, states: "Line numbers are asserted NOWHERE here on purpose: they rot (this repo's register row IDs already did, tree-wide)." Name the test by its title, not its coordinates — that is the convention the file already sets for itself, and the citation has already drifted in its first commit.

---

### Verdict

**No 🔴/🟠 correctness bugs remain.** The headline defect — a green spec whose name asserted a drift check it did not perform — is genuinely fixed, and the fix is measurably stronger than the minimum: it catches substitution, reordering and arity changes, and it restores the #3151 sibling shape. Finding 2 keeps only a 🟡 residue because the code half cannot be written here.

Four 🟡 cleanups remain (finding 2's outreach, finding 4's contradicted NOTE, NEW-1, NEW-2). Per findings-triage they are fixed in this round, but 🟡-only findings do **not** re-trigger the loop — fold them and merge; no pass 3 is owed. The one with a deadline attached is finding 2: whichever of #3191/#3151 merges second must delete the other's redundant assertion, and right now only #3191's body says so.

PR #3199, opened 2026-09-11
fix(server): pass the engine to stage 2 in attribution eval runs

### LINKED ISSUE #3196: fix(server): attribution eval sizes stage-2 chunks with the cloud budget for local (qwen) runs
## What happens

`evalFixture` in `server/src/analyzer/attribution-eval/run-eval.ts` calls `attributeChapterStage2` (`run-eval.ts:188-197`) without passing `engine`.

`attributeChapterStage2` sizes chunks with `resolveStage2ChunkCharBudget(opts.engine, opts.chapter.body)` (`server/src/routes/analysis.ts:2308`). With `engine` undefined, the resolver takes its non-`local` branch (`server/src/analyzer/stage2-chunk.ts:65-74`) and sizes the body with the Gemini cloud budget.

So an eval run with `engine: 'qwen'` (local Ollama) chunks stage 2 differently from a production local run, while the same function's opt-in review stage does derive the engine (`chunkEngine`, `run-eval.ts:214`, passed at `:238`). Its own comment says eval chunk boundaries should approximate production's.

## Expected

Stage 2 in the eval uses the same engine mapping as the review stage (`qwen` → `local`, `gemma` → `gemini`), so eval stage-2 chunk boundaries match a production run on that engine.

## Found

In passing while planning #3084 wave 2 (capacity pinning).

Refs #3084

dudarenok-maker:
## Summary

`evalFixture` (`server/src/analyzer/attribution-eval/run-eval.ts`) called `attributeChapterStage2` without `engine`. `resolveStage2ChunkCharBudget(undefined, body)` takes its non-local branch, so a `qwen` (local Ollama) eval sized stage-2 chunks with the Gemini cloud budget — unlike a production local run and unlike the same function's review stage, which already mapped `qwen → local` / `gemma → gemini`.

The mapping is now derived once, before stage 2, and passed to both stage 2 and the review stage. An eval with no engine still resolves to `gemini`, which takes the same cloud branch the missing engine did, so that case is unchanged.

Found in passing while planning #3084 wave 2 (capacity pinning).

No release-notes entry: the attribution eval is a developer tool with no user- or operator-visible effect.

## Test plan

- [x] New `run-eval.test.ts` cases: `passes qwen engine as "local" to attributeChapterStage2` and `passes gemma engine as "gemini" to attributeChapterStage2` — red before the fix, green after.
- [x] Mutation proof: removing `engine: chunkEngine` from the stage-2 call turns both red again.
- [x] All 142 tests across the 17 `attribution-eval` test files pass.

Closes #3196
Refs #3084

https://claude.ai/code/session_013DFfsAoY1LtxjDgnGPSZkc

### DIFF
--- a/server/src/analyzer/attribution-eval/run-eval.test.ts
+++ b/server/src/analyzer/attribution-eval/run-eval.test.ts
@@ -1,8 +1,9 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { evalFixture, rosterToStage1, familyBreakdown, aggStage, aggregateFixture, rosterAliasMap, scoreStage, type StageScore, type ReviewScore, type FixtureResult } from './run-eval.js';
 import type { LabelledChapter } from './schema.js';
 import type { RosterSnapshot } from './roster-schema.js';
 import type { SentenceOutput } from '../../handoff/schemas.js';
+import * as analysisModule from '../../routes/analysis.js';
 
 const roster: RosterSnapshot = { characters: [
   { id: 'narrator', name: 'Narrator' },
@@ -171,6 +172,45 @@ describe('evalFixture', () => {
     expect('reviewed' in res).toBe(false);
     expect(res.final.recall).toBeCloseTo(1);
   });
+
+  describe('engine parameter mapping to attributeChapterStage2', () => {
+    let attributeChapterStage2Spy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(() => {
+      attributeChapterStage2Spy = vi.spyOn(analysisModule, 'attributeChapterStage2');
+    });
+
+    afterEach(() => {
+      attributeChapterStage2Spy.mockRestore();
+    });
+
+    it('passes qwen engine as "local" to attributeChapterStage2', async () => {
+      // Call with engine: 'qwen', which should be mapped to 'local' when passed to attributeChapterStage2
+      await evalFixture({
+        analyzer: fakeAnalyzer,
+        manuscriptId: 'm', title: 'T', truth, roster, chapterId: 44,
+        stageCall: { language: 'en' } as never,
+        engine: 'qwen',
+      });
+
+      // Verify that attributeChapterStage2 was called with engine: 'local'
+      expect(attributeChapterStage2Spy).toHaveBeenCalledWith(expect.objectContaining({ engine: 'local' }));
+    });
+
+    it('passes gemma engine as "gemini" to attributeChapterStage2', async () => {
+      // Call with engine: 'gemma', which should be mapped to 'gemini' when passed to attributeChapterStage2
+      await evalFixture({
+        analyzer: fakeAnalyzer,
+        manuscriptId: 'm', title: 'T', truth, roster, chapterId: 44,
+        stageCall: { language: 'en' } as never,
+        engine: 'gemma',
+      });
+
+      // Verify that attributeChapterStage2 was called with engine: 'gemini'
+      expect(attributeChapterStage2Spy).toHaveBeenCalledWith(expect.objectContaining({ engine: 'gemini' }));
+    });
+  });
+
 });
 
 describe('evalFixture — reviewed char-stage (opt-in)', () => {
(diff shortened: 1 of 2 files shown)


dudarenok-maker:
## PR review — pass 1 (head 83c359fe, depth medium)

Scope: `server/src/analyzer/attribution-eval/run-eval.ts`, `server/src/analyzer/attribution-eval/run-eval.test.ts`, plus the code they reach: `attributeChapterStage2` (`server/src/routes/analysis.ts:2201-2435`), `resolveStage2ChunkCharBudget` (`server/src/analyzer/stage2-chunk.ts:70-82`) and `run-eval-cli.ts`. Verified first: `run-eval.test.ts` 16/16 green at `--retry=0` on this head; all 17 `attribution-eval` test files 142/142 green at `--retry=0` (matches the PR body's count); cloud CI green (typecheck/lint, server fast+slow, Windows). Then probed adversarially: can the new test fail, did the no-engine path change, and does the same omission exist at any other call site.

### ✅ No findings

What was probed, and what came back:

- **The test observes the call it claims to.** `vi.spyOn(analysisModule, 'attributeChapterStage2')` spies on the module namespace, while `run-eval.ts` uses a named import. Under vitest 4's transform, that named import reads the namespace property at call time, so the spy does intercept. Confirmed by running it, not just by reasoning: if the spy did not intercept, it would never be called and both new tests would be red on this head. They are green. The spy passes through to the real function, so the rest of `evalFixture` still runs for real.
- **Red before the fix, for the claimed reason.** Mutation run with no tracked file touched: a scratch vitest config outside the repo whose Vite `transform` hook removes the stage-2 `engine: chunkEngine,` line in memory only. Exactly the two new tests fail (`Received: - "engine": "local"` / `- "engine": "gemini"`, meaning the key is absent), and the other 14 pass.
- **The assertions tell the two mappings apart.** A second in-memory mutant forces stage 2 to `engine: 'local'` for every input. The gemma→`gemini` test goes red (`+ "engine": "local"`) and the qwen→`local` test stays green, so a mapping collapsed to one constant would not pass.
- **No-engine behaviour is unchanged; the PR body's claim holds.** `attributeChapterStage2` reads `opts.engine` in exactly one place, `analysis.ts:2308` → `resolveStage2ChunkCharBudget(opts.engine, body)`. That resolver branches only on `engine !== 'local'` (`stage2-chunk.ts:72`), so `undefined` and `'gemini'` take the same cloud branch with the same inputs. Nothing else in the function (escalation mode, prompt building, the structure engine, `onStages`) reads `engine`. The review stage gets the same `chunkEngine` value as before; the declaration just moved above the stage-2 call.
- **The fix is reachable as shipped.** The only production caller of `evalFixture` is `run-eval-cli.ts:125`, which passes `engine` (`:135`) for every `--engine` in its loop, so a qwen CLI run now reaches `stage2ChunkBudgetForEngine(..., 'local', ...)`.
- **No other call site has the same omission.** Outside the route file, the only non-test caller of `attributeChapterStage2` is `run-eval.ts`. Both production `attributeChapterStage2WithEval` sites pass `engine: phase1Selection.engine` (`analysis.ts:5375`, `:7221`). The only non-test caller of `runStage2ChapterChunked` is `analysis.ts:2306`. The other optional-engine budget resolver, `resolveStage1ChunkCharBudget`, gets `selection.engine` at both production sites (`analysis.ts:4411`, `:6927`) and a literal `'local'` at `chapter-chunker.ts:136`. The other `attribution-eval` modules that import route code (`review-run.ts`, `capture.ts`, `capture-cli.ts`) only import `script-review.js` helpers, and `runReviewOverChapter` types `engine` as required (`review-run.ts:47`). Nothing outside `server/src` calls any of these.
- **House gates.** `Closes #3196` is present and outside any code span. The release-notes pair is explicitly marked not applicable, with a reason (developer-only eval tool). No on-box acceptance row is owed: the change is argument wiring into a deterministic budget function that `stage2-chunk.test.ts` already unit-tests, so nothing here needs hardware to prove. No config knob, no `cast.json` write, no derived artifact. No comment was made false: the review-roster comment at `run-eval.ts:220-225` still describes the code accurately.

### Verdict

**Clean: no correctness bugs and no cleanup items.** The fix does what #3196 asks. The paired test fails without the fix and for the claimed reason, and the no-engine path is provably unchanged. This pass does not trigger a re-review.

Issue #3107, opened 2026-09-08
[agent instructions][claude][verify] Verify #3097 device-hint tolerance fix


dudarenok-maker:
[agent instructions][claude][verify] Verify #3097 device-hint tolerance fix

## What

Verify the fix for #3097 landed on branch `fix/sidecar-3097-device-hint-tolerance`
in worktree `C:\Claude\Projects\wt-3097-device-hint-tolerance`, then open the PR.

## The decision (already made — do not re-litigate)

#3097's own body named the owed decision: keep `preferred` as a hard
"wins if it fits" preference (a), or weaken it to only win when competitive
with the unconstrained winner (b). **Decision: (b), with a concrete
tolerance.** `preferred` wins if it fits AND its free headroom (`free_mb`) is
at least 75% of the unconstrained winner's free headroom (i.e. within a 25%
relative tolerance); otherwise placement falls back to the unconstrained
winner. This directly fixes the reported inversion (a hint pointing at the
card Qwen already occupies loses to the genuinely free card) while still
letting a hint break near-ties in its own favor, which is what "advisory but
not ignorable" should mean. Reject any rework proposing option (a)
(keep-as-is) or a materially different tolerance shape without a fresh
decision from a human — a rework may adjust the exact 75% constant if the
task child's own reasoning found it clearly wrong (e.g. it doesn't fix the
issue's own repro numbers), but must stay within the "compare against
unconstrained, tolerance-based" shape.

## Parent acceptance, restated as a checklist

- [ ] `PlacementController._resolve_admission`'s `preferred is not None and
  constraint is None` branch (`server/tts-sidecar/main.py`, near line 5166)
  now also computes the unconstrained `try_hold` candidate and compares free
  headroom before committing to the hinted device.
- [ ] The issue's own repro numbers are fixed: `cuda:0` 24000/18000 (free),
  `cuda:1` 16000/7000 (free, Qwen resident) → a `preferred=cuda:1` request now
  resolves to `cuda:0` (7000 is not within 75% of 18000: 7000 < 0.75×18000 =
  13500).
- [ ] A case where the hint is CLOSE to the unconstrained winner (e.g. hinted
  free headroom ≥ 75% of the winner's) still resolves to the hinted device —
  confirming this isn't a silent full revert to "always ignore preferred."
- [ ] A case with no competing candidate (hint is the only fitting device, or
  IS the unconstrained winner) is unaffected — same behavior as before.
[...]

Issue #3166, opened 2026-09-10
[agent instructions][claude][verify] Verify #3165 preferred-tolerance fix (#3097 follow-up)


dudarenok-maker:
## What

Verify #3165's work: adding a 75%-of-winner tolerance check to the `preferred`
(X-Device-Hint) branch in `PlacementController._resolve_admission()`
(`server/tts-sidecar/main.py`), mirroring #3097's existing `pinned` fix.

This is a mid-chain verify, not the chain's final one — #3160 (already filed)
verifies the doc-comment update that #3159 resumes once this passes. **Do
not open a PR from this issue.**

## Checklist

- [ ] Read the actual diff on `fix/sidecar-3097-device-hint-tolerance`
  (worktree `C:\Claude\Projects\wt-3097-device-hint-tolerance`) — not just
  #3165's receipt — for the change to `_resolve_admission()`.
- [ ] The new tolerance block only affects the `preferred` branch
  (`main.py` around line 5187-5205) — confirm the existing `pinned` block
  (lines 5171-5185) is byte-for-byte unchanged.
- [ ] The mechanism is correct: when `preferred` is set, `constraint is
  None`, there is more than one candidate GPU, and the hinted device's free
  headroom is < 75% of the unconstrained winner's free headroom, the code
  falls through to the UNCONSTRAINED `try_hold` (full candidate set) rather
  than restricting `candidates` the way the `pinned` block does. Read the
  actual code path, don't take a comment's word for it.
- [ ] 3 new tests exist in `server/tts-sidecar/tests/test_placement.py`,
  named `test_preferred_tolerance_issue_repro_overrides_hint`,
  `test_preferred_tolerance_near_tie_honours_hint`,
  `test_preferred_tolerance_single_gpu_unaffected`, and match the numbers
  in #3165's brief (cuda:0 24000/18000, cuda:1 16000/7000 → cuda:0;
  cuda:1 16000/13600 → cuda:1; single GPU → unaffected).
- [ ] Run `pytest server/tts-sidecar/tests/test_placement.py -k
  "preferred_tolerance or pinned_tolerance"` yourself — must be 6/6 pass.
- [ ] **Re-run the mutation** #3165's receipt claims: comment out the new
  tolerance block yourself, re-run the 3 `preferred_tolerance` tests, confirm
  `test_preferred_tolerance_issue_repro_overrides_hint` actually fails with
  the expected assertion mismatch (not a hang, not an unrelated error) —
[...]

Issue #3159, opened 2026-09-10
[agent instructions][claude][task] Merge current main into #3097's device-hint tolerance fix and update the clone-voice-resolver.ts doc comment


dudarenok-maker:
[agent instructions][cline][task] Merge current main into #3097's device-hint tolerance fix and update the clone-voice-resolver.ts doc comment

## What

Branch `fix/sidecar-3097-device-hint-tolerance` (worktree
`C:\Claude\Projects\wt-3097-device-hint-tolerance`, commit `b401d41f`)
already implements the 75%-of-winner tolerance decision from #3097 correctly
in `server/tts-sidecar/main.py`'s `PlacementController._resolve_admission` —
this was independently re-verified in #3107 (this rework's parent verify):
all 3 new unit tests in `server/tts-sidecar/tests/test_placement.py` pass,
and reverting the tolerance check (commenting out the comparison block) makes
`test_pinned_tolerance_issue_repro_overrides_hint` fail exactly as the
mutation-test receipt on #3108 claimed. **Do not change that logic.**

## The actual problem

The branch's merge-base with `origin/main` is `a50c16bf` — **96 commits
behind `origin/main`** as measured on 2026-09-10. It was branched before
commit `b7717074` ("feat(server,sidecar): add per-request X-Device-Hint for
lazy Coqui derive") landed on `main`, so it is missing that entire feature,
including the doc comment #3097 explicitly requires be re-checked:
`server/src/tts/clone-voice-resolver.ts:887-908` (line numbers as of current
`origin/main`; they will shift after your merge). That doc comment currently
reads (on `origin/main`):

> "...`_resolve_admission`'s hinted-device try_hold only checks whether
> `cuda:1` itself has room, never whether some other card is materially
> freer, so a `cuda:1` that merely FITS wins outright — even while `cuda:0`
> sits nearly empty... Whether 'advisory' should instead mean 'wins only when
> competitive with the [unconstrained winner]' is the open question #3097
> resolves."

This is now stale: #3097 resolved that open question as decision (b), and
`main.py`'s fix changes the exact semantics this comment describes. The
comment needs updating wherever it currently sits (search near `#3058` /
`lazyCoquiDeriveDeviceHint` in `clone-voice-resolver.ts`) — it must describe
the new 75%-of-winner tolerance behavior, not the old "wins outright if it
fits" behavior it currently documents.

The task child on #3108 ran into this same gap and reported "the function
appears to be part of a different sub-task... that has not yet been
[...]

Issue #3108, opened 2026-09-08
[agent instructions][cline-qwen-cloud][task] Weaken X-Device-Hint placement to a 75%-of-winner tolerance


dudarenok-maker:
[agent instructions][cline][task] Weaken X-Device-Hint placement to a 75%-of-winner tolerance

## What

Fix #3097: `PlacementController._resolve_admission`'s `preferred` branch
(`server/tts-sidecar/main.py`, the `preferred is not None and constraint is
None` case, currently around line 5166) restricts `try_hold` to the hinted
device alone and takes it if it merely fits — never comparing against a
materially freer alternative. This defeats `#3058`'s whole point (steering the
Coqui derive away from the card Qwen is generating on) whenever a stale
device-list cache or an operator's device pin makes the hint point AT that
exact occupied card.

## The decision (already made — implement this, don't re-decide)

**`preferred` wins if it fits AND its free headroom is at least 75% of the
unconstrained winner's free headroom.** Otherwise, fall back to the
unconstrained winner. This is a deliberate choice (see the verify child's
brief, issue #3107, for the full rationale) — implement exactly this shape,
not option (a) "keep as hard preference" from the original issue.

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3097-device-hint-tolerance`
Branch: `fix/sidecar-3097-device-hint-tolerance` (already checked out there)

Do NOT create another worktree. Do NOT work in the primary checkout. Commit on
this branch. Do NOT open a PR — the verify child does that.

## Background you need

`server/src/gpu/ensure-gpu-device-list-warm.ts:33` already fetches `free_mb`/
`total_mb` per device and currently discards them — this is the data your
comparison needs. Read `PlacementController._resolve_admission` and
`PlacementController.reservation` in `server/tts-sidecar/main.py` (the issue
names line ~5166 as the current `preferred` branch, but locate it fresh — line
numbers drift) to understand `try_hold`'s existing shape before adding the
comparison.

## The fix

In the `preferred is not None and constraint is None` branch:

1. Compute `try_hold` against the hinted device (existing behavior).
2. If it fits, ALSO compute the unconstrained placement result (no `preferred`
   constraint) — i.e. run the same admission logic without pinning to the
   hint, to find what device it would otherwise have chosen and that
   candidate's free headroom.
3. Compare: if the hinted device's free headroom is `>= 0.75 *
   unconstrained_winner.free_mb`, honor the hint (existing return). Otherwise,
[...]

PR #3207, opened 2026-09-11
fix(sidecar): apply 75% headroom tolerance to preferred device hints

### LINKED ISSUE #3097: device hint: a fitting-but-worse card beats a materially freer one
## What

`PlacementController._resolve_admission` in `server/tts-sidecar/main.py` (the `preferred is not None and constraint is None` branch, currently around line 5166) restricts `try_hold` to the hinted device alone and takes it if it merely fits — it never compares the hinted device's headroom against any other candidate's. So `X-Device-Hint` (today only the `#3058` lazy-Coqui-derive self-heal on `/xtts/clone-voice`) wins over a materially freer card whenever the hinted card fits at all.

Surfaced by PR #3061 review R2. Repro (from the review): `cuda:0` at 24000 total / 18000 free, `cuda:1` at 16000 total / 7000 free because Qwen is resident there, peak 4000:

```
no hint           -> {'device': 'cuda:0'}   # the FREE card
preferred=cuda:1  -> {'device': 'cuda:1'}   # the card Qwen is on
```

This is the opposite of what the feature exists for: `#3058`'s whole point is steering the Coqui derive AWAY from the card Qwen is generating on, but a stale-but-still-populated GPU device-list cache (`#3061` review N1) or an operator's own `tts.qwen.device` pin can make the hint point AT that exact card, and today's admission logic happily parks the derive there instead of on the genuinely free one — silently, no error, no retry.

The corrected doc comment (this PR, `server/src/tts/clone-voice-resolver.ts:887-908`) now states the real semantics: a `preferred` hint wins whenever the hinted device fits, even against a materially better alternative.

## The decision that is owed

Whether `preferred` should keep "wins if it fits" semantics, or should only win when it is *also* competitive with (not worse than) what unconstrained placement would pick. Two shapes, at least:

- (a) Keep as-is: a hint is a hard preference among fitting devices, full stop. Document it and move on (already done here).
- (b) Weaken it: compute the unconstrained `try_hold` result too, and only honor `preferred` if its headroom is within some tolerance of (or not worse than) the winner — i.e. only override "the free card" when the hinted card is roughly as good, not merely adequate.

(b) has a real cost: it means computing/discarding the unconstrained candidate list on every hinted request (cheap), and it changes behavior for every existing/future caller of `preferred`, not just `#3058`'s. It's a genuine "what does advisory mean" product decision, not a one-line bug fix — that's why this is filed rather than fixed on the spot (CLAUDE.md's incidental-findings design-pass carve-out).

## Key files

- `server/tts-sidecar/main.py` — `PlacementController._resolve_admission` (the `preferred` branch), `PlacementController.reservation`
- `server/src/tts/clone-voice-resolver.ts` — `lazyCoquiDeriveDeviceHint`, the corrected doc comment
- `server/src/gpu/ensure-gpu-device-list-warm.ts:33` — already fetches `free_mb`/`total_mb` and discards them; whichever option is chosen, this is the data a smarter comparison would need already sitting there.

## Depends on

None — self-contained.

### LINKED ISSUE #3165: [agent instructions][claude][task] Add 75%-of-winner tolerance to the preferred (X-Device-Hint) branch (#3097 follow-up)
## What

Mirror #3097's `pinned` 75%-of-winner tolerance fix onto the `preferred`
(advisory, X-Device-Hint) branch in `PlacementController._resolve_admission()`.
Today `preferred` wins if it merely fits, with no headroom comparison against
the unconstrained winner — this is the actual gap #3097 was filed to close for
the Coqui-derive self-heal (#3058), which uses `preferred` (via
`X-Device-Hint`), not `pinned`. #3159 found that `b401d41f` applied the fix to
the wrong parameter (`pinned` instead of `preferred`) and is blocked on this
landing before it can honestly write the doc comment #3097 asked for.

## Background you need

`server/tts-sidecar/main.py`, `PlacementController._resolve_admission()`:

- Lines 5171-5185: the existing `pinned` tolerance block (#3097, **do not
  modify** — already independently verified in #3107, out of scope here).
- Lines 5187-5205: the `preferred` advisory branch — one `try_hold` restricted
  to the hinted device (5200-5203), no headroom comparison, falls through to
  the full unconstrained `try_hold` at 5204-5205 only if the hinted device
  doesn't fit at all.

#3097's ratified decision (see its issue body): a hinted device wins only when
its free headroom is ≥75% of the unconstrained winner's free headroom;
otherwise use the unconstrained winner. That decision applies to the
production `X-Device-Hint` path, which is `preferred`, confirmed at
`main.py:12038` (`reservation(preferred=device_hint, ...)`).

3 existing tests at `server/tts-sidecar/tests/test_placement.py:1538-1591`
(`test_pinned_tolerance_*`) cover the `pinned` case — model your 3 new tests
on these, calling `pc.reservation(..., preferred="cuda:1")` instead of
`pinned=`.

## The fix

In `_resolve_admission()`, immediately before the existing
`if preferred is not None and constraint is None:` block (line 5200), add a
headroom comparison mirroring the `pinned` block at 5171-5185, adapted for
`preferred`. Read lines 5171-5205 in full before writing the patch — in
particular note the `pinned` block RESTRICTS `candidates` when the tolerance
check fails, but `preferred`'s whole contract (comment at 5187-5194) is
"try the hint, and if it doesn't fit OR isn't materially competitive, fall
through to the full unconstrained candidate set" — so the correct mechanism
here is to null out `preferred` when the check fails (so the `if preferred is
not None` guard at line 5200 is skipped and execution falls straight to the
existing unconstrained `try_hold` at 5204-5205), NOT to rewrite `candidates`
the way the `pinned` block does.

Shape (adapt exact form/comment style to match the file, cite #3097 and this
issue number for the "why"):

```python
if preferred is not None and constraint is None and candidates:
    all_gpus = self._gpu_candidates(devices, None)
    if len(all_gpus) > 1:
        preferred_candidates = [c for c in candidates if c[0] == preferred]
        if preferred_candidates:
            preferred_free = preferred_candidates[0][1]
            winner_free = max(c[1] for c in all_gpus)
            if preferred_free < 0.75 * winner_free:
                preferred = None
if preferred is not None and constraint is None:
    held = self.ledger.try_hold(
        [c for c in candidates if c[0] == preferred], peak, reserve_cap, engine
    )
```

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3097-device-hint-tolerance`
Branch: `fix/sidecar-3097-device-hint-tolerance` (already carries #3159's
merge commit `3a6eb27e` — this worktree is already on the right commit; do
not create another worktree, do not work in the primary checkout).

## Acceptance

Add 3 new tests to `server/tts-sidecar/tests/test_placement.py`, directly
after `test_pinned_tolerance_single_gpu_unaffected` (line 1591), shaped
exactly like their `pinned` counterparts but using `preferred=`:

1. `test_preferred_tolerance_issue_repro_overrides_hint` — cuda:0 24000/18000
   free, cuda:1 16000/7000 free, `preferred="cuda:1"` → resolves to `cuda:0`
   (7000 < 0.75×18000 = 13500).
2. `test_preferred_tolerance_near_tie_honours_hint` — cuda:0 24000/18000,
   cuda:1 16000/13600, `preferred="cuda:1"` → resolves to `cuda:1`
   (13600 ≥ 13500).
3. `test_preferred_tolerance_single_gpu_unaffected` — one GPU,
   `preferred="cuda:0"` → resolves to `cuda:0` (tolerance check skipped, no
   competing candidate).

Run `pytest server/tts-sidecar/tests/test_placement.py -k
"preferred_tolerance or pinned_tolerance"` — expect 6/6 pass (3 new, plus the
3 existing `pinned` ones untouched and still green).

**Mutation check, required, with observed output:** comment out your new
tolerance-check block (so `preferred` is used unconditionally, as before)
and re-run the 3 new `preferred_tolerance` tests —
`test_preferred_tolerance_issue_repro_overrides_hint` must FAIL (asserts
`cuda:0`, gets `cuda:1`). Paste the actual pytest failure output in your
`AGENT DONE` receipt, not just a claim that it failed. Then restore your real
fix and re-confirm all 6 pass before committing.

## Key files

- `server/tts-sidecar/main.py:5187-5205` (the branch you're changing)
- `server/tts-sidecar/tests/test_placement.py:1538-1591` (pattern to mirror)

## Not in scope

- The `pinned` tolerance block (lines 5171-5185) or its 3 existing tests —
  already shipped and independently verified in #3097/#3107. Do not touch.
- `admit()` (`main.py:4912-4974`) — a separate, parallel `preferred`/`pinned`
  decision function with **no tolerance check at all on either parameter**.
  Its own docstring says it has "no production caller today" and demands
  parity with `_resolve_admission()`, which means it is *already* out of
  parity (missing the #3097 `pinned` fix too, not just this one). That is a
  pre-existing gap this chain did not create — leave `admit()` untouched.
  Note it in your `AGENT DONE` receipt as an aside; do not fix it.
- `clone-voice-resolver.ts` or any doc comment — that is #3159's job, which
  resumes once this lands (its successor flip below re-queues it).
- `_VdKokoroArbiter`/`design()` (#3086) — separate chain, do not touch.

## Commit/push gate

Never run `git commit` or `git push` in a backgrounded/detached process by
default — both complete in seconds in the foreground now. The ONE exception:
this branch touches `server/tts-sidecar/**`, so `git push` on it runs the
full sidecar pytest suite at push time (~6.85 minutes) — launch that push
detached and poll for its `EXIT=` sentinel rather than waiting in the
foreground or assuming it worked from launching it. Never `--no-verify`.

## On completion

1. Commit and push to the same branch.
2. Flip the successor (the verify child linked to this issue) to `Agent
   Todo` FIRST:

```bash
ITEM_ID=$(gh api graphql -F owner=dudarenok-maker -F repo=Castwright -F num=3166 -f query='query($owner:String!,$repo:String!,$num:Int!){ repository(owner:$owner,name:$repo){ issue(number:$num){ projectItems(first:10){ nodes{ id } } } } }' --jq '.data.repository.issue.projectItems.nodes[0].id')
gh project item-edit --id "$ITEM_ID" --project-id PVT_kwHOEOX6_c4Bcf9a --field-id PVTSSF_lAHOEOX6_c4Bcf9azhXISes --single-select-option-id 9d9fa565
```

3. THEN leave `AGENT DONE` and set THIS issue's Status to `Agent Review`. Do
   not close this issue yourself.

dudarenok-maker:
## Summary

**Removed the `pinned` tolerance block** — it was an unreverted misapplication of #3097. Per #3165's own statement and the PR review findings, the design decision #3107 ratified applies only to `preferred` (X-Device-Hint), not `pinned`. Reverted `pinned` back to hard-honor semantics: an operator's `COQUI_DEVICE` env pin or resident engine always wins, with no tolerance comparison. Only `preferred` gets the 75%-of-winner free-headroom tolerance weakening.

**Fixed headroom comparison bug in the `preferred` block**: the tolerance check was comparing raw `freeMb` instead of actual `_headroom` values (which account for ledger-reserved VRAM and per-device reserves). Added `ReservationLedger.get_headroom()` public method and use it in the tolerance check, so a card that is already booked by a pending reservation doesn't get treated as free.

**Fixed test fixtures**: 
- Removed the three `pinned`-tolerance unit tests from `test_placement.py` (they test behavior that no longer exists).
- Kept the three `preferred`-tolerance unit tests and adjusted one fixture (`test_preferred_tolerance_near_tie_honours_hint`) to account for per-device reserves in the headroom calculation.
- Reverted COQUI_DEVICE (pinned) route tests back to original values since pinned no longer has a tolerance window.
- Kept X-Device-Hint (preferred) route test fixtures adjusted so they stay within the 75% tolerance band and keep testing plumbing, not edge cases.

**Fixed doc comments** in `clone-voice-resolver.ts` to correctly attribute the tolerance to #3097/#3107/#3165 (not just b401d41f) and remove references to pinned tolerance.

Closes #3097
Closes #3165

## Test plan

Four `preferred`-tolerance cases in `server/tts-sidecar/tests/test_placement.py`:

- `test_preferred_tolerance_issue_repro_overrides_hint` — the original issue repro: hinted device (X-Device-Hint: cuda:1) is far below the unconstrained winner's headroom, tolerance check fails, placement falls through to the freer device (FAIL → PASS transition).
- `test_preferred_tolerance_near_tie_honours_hint` — hinted device's headroom is within the 75% band of the winner, hint is honored (PASS, regression guard — tolerance must not break this case).
- `test_preferred_tolerance_single_gpu_unaffected` — only one GPU present, no comparison possible, hint path behaves exactly as before (PASS, regression guard).
- `test_preferred_tolerance_ledger_reservation_reduces_effective_headroom` — a card with free space but pending ledger-reserved VRAM still fails the tolerance check when its actual headroom (accounting for the reservation) falls below 75% of the winner's (PASS, regression guard).

All 132 sidecar placement + route tests pass.

Sidecar acceptance: `npm run test:sidecar` -- 2026-09-12 -- passed

https://claude.ai/code/session_01LGHSyr5oVgaBmZuU7z3NNj


dudarenok-maker:
## PR review — pass 1 (head 2523e836, depth high)

Scope: all 4 changed files (`server/tts-sidecar/main.py`, `tests/test_placement.py`, `tests/test_xtts_clone_voice.py`, `server/src/tts/clone-voice-resolver.ts`) plus the surrounding admission path (`_gpu_candidates`, `ReservationLedger._headroom`/`try_hold`, `_evict_until`, `admit()`, all 10 `reservation(pinned=…)` call sites), the issue chain (#3097 → #3107 → #3108 → #3159 → #3165 → #3166), and the two on-box surfaces that already track this behaviour (`docs/testing/onbox-acceptance-register.md` row A106, `docs/testing/device-hint-placement-onbox-acceptance.md`).

Verified before probing: `pytest tests/test_placement.py -k tolerance` → 6 passed; `pytest tests/test_xtts_clone_voice.py` → 73 passed. All three failure scenarios below were **executed**, not reasoned about, against this exact HEAD via read-only scratch drivers in `$TEMP` (no repo file touched; `git status --porcelain` clean before and after).

---

### 🔴 Blocking (correctness) — the `pinned` half of this change is not the decision #3097 ratified, and it inverts a written on-box acceptance criterion

`server/tts-sidecar/main.py:5179-5185`

**#3097 decided `preferred`, not `pinned`.** The issue body names "the `preferred is not None and constraint is None` branch" and asks "Whether **`preferred`** should keep 'wins if it fits' semantics". #3107, the verify child where the decision was actually ratified, restates it as: "**Decision: (b)** … **`preferred`** wins if it fits AND its free headroom is at least 75% of the unconstrained winner's" — and every one of its acceptance checkboxes names `preferred`. Nothing in either issue ratifies weakening `pinned`.

That is not my inference; it is recorded in this chain's own tickets. #3165's body, verbatim: *"#3159 found that `b401d41f` applied the fix to the **wrong parameter** (`pinned` instead of `preferred`)"*. `b401d41f` was then left in place and #3165 added the `preferred` block alongside it. So the `pinned` block ships as an unreverted misapplication, and the PR body presents it as the ratified design ("this PR implements option (b) for both paths").

**What it actually changes.** `pinned` is `_engine_env_pin(...)` at all 10 reservation sites — `COQUI_DEVICE`, `QWEN_DEVICE`, `KOKORO_DEVICE`, `ASR_DEVICE`, `SPK_DEVICE`. This PR makes every one of those operator-set `risk: 'high'` knobs advisory: the engine silently loads on a different card whenever the pinned card's free VRAM is under 75% of another's. Blast radius is all five engines, not the Coqui derive #3097 was filed about — and the production X-Device-Hint path does not use `pinned` at all (#3165 confirms this at `main.py:12038`), so the `pinned` block buys #3097 nothing.

**Concrete failure — the repo's own acceptance criterion now fails.** `docs/testing/device-hint-placement-onbox-acceptance.md`, **Criterion 4 — "the operator's `COQUI_DEVICE` pin still wins"**:

> 1. Set `tts.coqui.device` to `cuda:0` … **Pass:** the derive lands on **GPU0** despite the hint naming `cuda:1`.
> A per-request preference must not overrule a `risk: 'high'` registry knob the operator set on purpose.

Register row A106 names the box: `cuda:0` is the 8 GB 4070, `cuda:1` the 16 GB 5070 Ti. Criterion 4 is not in the set that requires `QWEN_DEVICE=cuda:0`, so it runs with the box's standing layout. On an idle 2-card boot that is roughly `cuda:0` ≈ 7500 MB free, `cuda:1` ≈ 15500 MB free → `7500 < 0.75 × 15500 = 11625` → the new block drops the pin → **the derive lands on GPU1 and Criterion 4 FAILS**, on the exact hardware the register names, in its default state. The criterion's pass condition is inverted by this PR, and none of the three acceptance surfaces (register, run sheet, live view) moves in this diff — CLAUDE.md before-shipping step 3.

`docs/release-notes-next.md:207` (the shipped #3058/#3061 entry) carries the same now-false sentence: *"An operator's own `COQUI_DEVICE` pin still wins outright over the hint."*

**This is a judgment call, not something to auto-fix.** Weakening an operator's explicit device pin is defensible on the same contention argument that motivated `preferred` — which is exactly what makes it a design decision with more than one defensible outcome, and there is a written human-authored criterion asserting the opposite. Per the gate's judgment-call carve-out this suspends the fix loop: the repo owner decides whether to (a) drop `b401d41f`'s block and ship only the ratified `preferred` fix, or (b) keep it and update Criterion 4, A106, and the release-notes sentence in this diff. Do not resolve it by editing the code alone.

---

### 🟠 Significant (correctness) — the tolerance check reads raw `freeMb` while placement is decided by `_headroom`, so the fix misses its own headline scenario

`server/tts-sidecar/main.py:5182-5183` and `:5212-5213`

Both blocks compare `c[1]` — the raw `freeMb` off `probe()`. But the winner they claim to compare against is chosen by `ReservationLedger.try_hold` → `_headroom` (`main.py:4550-4558`), which is `min(free_mb, total_mb − ledger_reserved) − _device_reserve_mb(total, cap)`. The ledger term is the whole point: a `reservation()` holds VRAM from admission time, **before** the model allocates it, so a card that is already booked still reads as free to `probe()`.

That is precisely the case #3097 exists to catch — "steering the Coqui derive AWAY from the card Qwen is generating on". Measured at this HEAD:

```
Qwen holds 9000 MB on cuda:1 via the ledger (in-flight, not yet allocated).
devices: cuda:0 24000/18000 free, cuda:1 16000/14000 free, peak 4000, cap 500

  raw freeMb    : cuda:0=18000  cuda:1=14000  -> tolerance sees 14000 >= 0.75*18000 (13500) = True
  REAL headroom : cuda:0=17500  cuda:1=6500   -> a headroom-based check would see 6500 >= 0.75*17500 (13125) = False

  A) pinned='cuda:1'    -> {'device': 'cuda:1'}     <-- lands on the busy card
  B) preferred='cuda:1' -> {'device': 'cuda:1'}     <-- lands on the busy card
  C) CONTROL (no pin/hint) -> {'device': 'cuda:0'}  <-- what placement should pick
```

`cuda:0` has 2.7× the real headroom and unconstrained placement picks it, but both tolerance paths wave the hint through. The issue's own repro passes only because it describes Qwen as already **resident** (allocated), which `freeMb` does reflect; the reserved-but-not-yet-allocated window — a cold Qwen load reserving ~4–5 GB, or the stale device-list cache #3097 itself cites — does not.

Recurring shape #4, *a metric blind to a case it must score*. The codebase already has the correct primitive three lines away; the fix is to compare `_headroom` on both sides rather than `c[1]`, and to add the ledger-hold case to `test_placement.py` (`make(...)` already accepts a ledger).

---

### 🟠 Significant (correctness) — the widened `candidates` is discarded by the evict ladder and the stranded-cache retry

`server/tts-sidecar/main.py:5185` vs `:5229-5231` and `:5250-5252`

When the `pinned` tolerance fires, the fix widens the local `candidates` to all GPUs. But both recovery paths rebuild the candidate list from `constraint` — which is still the pin — rather than from that widened local:

```python
def _fits():
    probed = self.probe()
    return self.ledger.try_hold(
        self._gpu_candidates(probed, constraint), peak, reserve_cap, engine   # :5230
    )
```

So the pin is non-binding for exactly one `try_hold` and binding again for everything after it. Measured at this HEAD (peak 20000, `cuda:0` 32000/18000, `cuda:1` 16000/7000, pinned `cuda:1`; `7000 < 0.75 × 18000` so the tolerance fires; the evict step frees 8000 MB on `cuda:0`, the card `_worst_device_key` targets):

```
  A) pinned='cuda:1'   free after evict: [('g0', 26000), ('g1', 7000)]
                       admission: {'noCapacity': {'neededMb': 20000, 'deviceKey': 'cuda:1'}}
  B) CONTROL pinned=None  free after evict: [('g0', 26000), ('g1', 7000)]
                       admission: {'device': 'cuda:0'}
```

`cuda:0` ends with 26000 MB free — comfortably fitting the 20000 peak — and the request is still refused, on the card the code just decided was not binding. Downstream that is the ~60 s `withCapacityRetry` stall ending in a stock-catalogue-voice substitution that `_parse_device_hint`'s own docstring (`:4127-4133`) describes as "strictly worse than sending no hint". `:5285` then reports `deviceKey: cuda:1` for the same reason.

Not a regression against `main` (the pin was hard everywhere before), but the fix is unreachable past the first `try_hold` — recurring shapes #7 and #10. `preferred` is unaffected: its guard requires `constraint is None`, so the ladder is genuinely unconstrained there. Nulling out `constraint` alongside widening `candidates`, or threading the widened list into `_fits`, closes it; the paired test is the scenario above.

---

### 🟡 Minor (cleanup)

**M1 — `clone-voice-resolver.ts:906-917` mis-cites both the commit and the issue, and calls branch-local code "already applies".** This paragraph is the entire deliverable of `b33bd42e`, and three of its claims are checkably wrong:
- *"#3097 (implemented by `b401d41f`) closed this gap: `_resolve_admission`'s **`preferred`** handling now weighs…"* — `git show b401d41f -- server/tts-sidecar/main.py | grep -c preferred` → **0**. `b401d41f` is "weaken **pinned**-device placement"; `preferred` was implemented by `c01ea35d`.
- *"the same tolerance check `_resolve_admission` already applies to `pinned` (**#3165**)"* — #3165 is titled "Add 75%-of-winner tolerance to the **preferred** (X-Device-Hint) branch". The citation is inverted.
- *"**already applies**"* — `git show origin/main:server/tts-sidecar/main.py | grep 3097` → no hits. Neither block exists on `main`; both ship first in this PR. A reader after merge will look for a prior `pinned` tolerance and find none.
- The closing phrase *"instead of `pinned`'s hard restriction"* contradicts the block it just credited with removing that restriction.

This is recurring shape #12: #3159's brief (written before `c01ea35d` existed) instructed "Cite this decision as made in #3097 / implemented by `b401d41f`", and the comment conforms to the stale brief rather than to the code.

**M2 — four in-file docstrings now state false invariants about `pinned`.** None updated: `_gpu_candidates` (`main.py:4765-4769`, "restricted to ONLY that device (no migration)"); `_parse_device_hint` (`:4125-4126`, `pinned` "restricts `_gpu_candidates` to one device with **no fallback at all**"); `_resolve_admission`'s own preferred comment (`:5193-5198`, "`pinned`, which restricts `_gpu_candidates` for the entire admission"; "an operator's env pin … the operator has said where it goes"); the clone-voice route (`:12039-12041`, "The operator's own `COQUI_DEVICE` pin still wins outright"). A106's row body also cites `main.py:5183-5187` for "under a `COQUI_DEVICE` pin the hint is a no-op" — post-PR those exact lines are the new tolerance block, i.e. the citation now lands on the opposite mechanism. Same for the run sheet's Criterion 3 citation of `main.py:5184-5187`.

**M3 — `admit()`'s parity invariant is now false, unmentioned.** `admit()` (`main.py:4911-4970`) has neither tolerance check, and its docstring says: *"Kept in parity with `reservation()`/`_resolve_admission` … an advisory decision function that silently dropped a parameter its binding twin honors would disagree with that twin the moment something calls it with a hint, **which is exactly the invariant this docstring already claims to hold**."* After this PR `admit(preferred='cuda:1')` and `reservation(preferred='cuda:1')` disagree on #3097's own repro numbers. `admit()` has no production caller, so this is a false docstring rather than a live bug — but #3165 scoped the *code* out, not the *docstring*, and the PR body doesn't mention the widened gap at all.

**M4 — release-notes pair missing, with no not-applicable statement.** Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the PR body does not say why. This changes which card models load on for a multi-GPU operator and (per the Blocking finding) overrides operator env pins — operator-visible by any reading, so CLAUDE.md before-shipping step 5 applies. The `#3058` entry at `docs/release-notes-next.md:207` additionally needs its "pin still wins outright" sentence corrected if the Blocking finding resolves as "keep".

**M5 — test gaps around the new blocks.** None of these has coverage: the `resident is None` guard at `:5179` (dropping it would let a resident engine be migrated and its VRAM booked on a card the model isn't on); the evict/reclaim paths (the Significant finding above); a ledger-hold case (the other Significant finding); and the *control* half of the new unit tests — none asserts what unconstrained placement would have picked, so a future fixture edit can make `near_tie`/`single_gpu` vacuous without anything going red. `test_pinned_tolerance_near_tie_honours_hint` and both `single_gpu` tests pass identically with the fix reverted; that's fine as guards, but the PR body should not read them as evidence the tolerance works.

---

### ✅ What is solid

- **The `2523e836` fixture fix is correct, and I verified the control rather than the arithmetic alone.** Threshold recomputed per test, then each fixture executed with the pin/hint removed to confirm it still discriminates:
  - `test_clone_voice_route_x_device_hint_overrides_best_fit_placement` — `cuda:0` 24000/18000, `cuda:1` 16000/14000. `0.75 × 18000 = 13500`; `14000 ≥ 13500` ✓ (500 MB inside the band). Measured: **no hint → `cuda:0`; `preferred=cuda:1` → `cuda:1`**. Still discriminating.
  - `test_clone_voice_route_engine_env_pin_still_wins_when_header_absent` and `test_clone_voice_route_invalid_hint_falls_back_to_engine_env_pin` — `cuda:0` 8192/5000 (pinned), `cuda:1` 24000/6000. `0.75 × 6000 = 4500`; `5000 ≥ 4500` ✓ (500 MB inside the band). Measured: **no pin → `cuda:1`; `pinned=cuda:0` → `cuda:0`**. Still discriminating.
  - The adjustment was necessary, not cosmetic: replayed against the **pre-adjustment** fixture (`cuda:1` free 20000), `pinned='cuda:0'` now resolves to **`cuda:1`** — the original assertion `device == "cuda:0"` really would have failed.
  - The docstrings' *conclusion* holds but their stated *reason* doesn't: the unconstrained winner is chosen by `_headroom`, not raw free. Real values at the route's live `GPU_RESERVE_MB=500` and `peak=3584`: `cuda:0` = `5000 − 410 = 4590`, `cuda:1` = `6000 − 500 = 5500`, so `cuda:1` is genuinely the better pick and `cuda:0` still clears the 3584 peak by 1006 MB. I also checked this is robust rather than lucky — the free-MB gaps (1000 and 4000 MB) exceed the maximum possible per-device reserve differences (790 and 400 MB), so the ordering holds for *every* `GPU_RESERVE_MB` value, not just 500. Worth rewording the two docstrings to say "headroom" instead of "free", but the numbers are right.
- **The `preferred` block's mechanism is the right one.** Nulling `preferred` rather than rewriting `candidates` preserves #3061 review C3's fall-through contract exactly, and means the evict ladder stays genuinely unconstrained on that path — which is why the Significant evict finding hits only `pinned`.
- `test_preferred_tolerance_issue_repro_overrides_hint` and `test_pinned_tolerance_issue_repro_overrides_hint` are real discriminators (both fail on a reverted block, per #3165's pasted mutation output), and `run_case`'s `_RAN` sentinel means a body that never reached its assertions fails rather than passing quietly.
- `Closes #3097` is present in the PR body, plain, outside any code span.
- The `preferred_candidates` empty-guard at `:5211` is correct: a hint naming an absent card leaves `preferred` set and lets the existing try_hold/fall-through handle it, rather than nulling on a missing input (avoids the fail-open shape #1).

---

### Verdict

**Do not merge as-is.** One Blocking finding needs a human decision before any code changes: the `pinned` tolerance is an unreverted misapplication of #3097 (recorded as such in #3165), it makes five operator `*_DEVICE` pins advisory, and it inverts the pass condition of an existing on-box acceptance criterion (Criterion 4 / row A106) that no surface in this diff updates. Either drop `b401d41f`'s block and ship the ratified `preferred` fix alone, or keep it and move Criterion 4, A106 and the release-notes sentence in the same diff.

The two Significant findings are ordinary defects and should be fixed this round whichever way the Blocking one resolves — the `freeMb`-vs-`_headroom` one especially, since it lets the feature's own headline scenario through on *both* paths. Both re-trigger a review pass once fixed and pushed.

The fixture fix in `2523e836` that prompted this review specifically is correct on every count I could measure, and needed doing.

dudarenok-maker:
## PR review — pass 2 (head dba38813, depth high)

Re-review of the 4-file diff after `dba38813`. Deltas only — pass 1's solid items are not re-listed. Everything below was **executed** against this exact HEAD via read-only drivers and pytest plugins in `$TEMP` (no repo file touched; `git status --porcelain` clean before and after; `npm run test:sidecar` scope: `tests/test_placement.py` + `tests/test_xtts_clone_voice.py` → **131 passed**). CI is green on all 21 checks.

---

## Pass-1 findings — disposition

### 🔴 Blocking (unratified `pinned` tolerance) — ✅ VERIFIED RESOLVED

The `pinned` block is gone from `_resolve_admission`; only the `preferred` block survives (`main.py:5194-5215`), guarded by `constraint is None` so residency **and** the env pin both still hard-restrict `_gpu_candidates` at `:5177`. Verified by replay rather than by reading the diff:

```
PROBE 2 -- pass 1's Criterion 4 scenario (A106's box, idle: cuda:0 8188/7500, cuda:1 16302/15500)
  pinned='cuda:0'  -> {'device': 'cuda:0'}      <-- pin honored
  CONTROL (no pin) -> {'device': 'cuda:1'}      <-- and the pin is what moved it
```

Criterion 4 replayed green at `GPU_RESERVE_MB` 768 **and** 500. Criterion 5 (hint names an absent card) also still resolves cleanly — the `preferred_candidates` empty-guard and the `len(all_gpus) > 1` guard both hold. The three `pinned` unit tests are gone, the two `COQUI_DEVICE` route fixtures are back to `8192/5000` + `24000/20000`, and `docs/release-notes-next.md`'s *"An operator's own `COQUI_DEVICE` pin still wins outright over the hint"* is true again and correctly left untouched. No drift on that surface.

### 🟠 Significant #1 (raw `freeMb` vs `_headroom`) — ⚠️ CODE FIX RESOLVED, but it shipped with **no regression test** (see N2)

The accounting is genuinely right, not just renamed. `get_headroom` (`main.py:4602-4608`) delegates to the same `_headroom` (`:4550-4558`) under the same `self._lock` that `try_hold` (`:4560-4583`) uses — identical ledger + per-device-reserve arithmetic, not a parallel implementation. Pass 1's exact scenario now inverts:

```
PROBE 1 -- Qwen holds 9000 MB on cuda:1 via the ledger (in-flight, not yet allocated)
  raw freeMb    : cuda:0=18000  cuda:1=14000
  REAL headroom : cuda:0=17500  cuda:1=6500
  preferred='cuda:1' -> {'device': 'cuda:0'}   (was 'cuda:1' at pass-1 HEAD)
  CONTROL (no hint)  -> {'device': 'cuda:0'}
```

I also checked the half the fix could have got wrong and didn't: the block's `winner_headroom` loop takes `argmax(headroom)` **unfiltered**, while `try_hold` takes `argmax(headroom)` subject to `headroom >= peak`. Those agree whenever any card fits (the max-headroom card *is* the fitting winner), and when none fits the tolerance is a no-op because both `try_hold`s fail anyway. So the "unconstrained winner" it compares against is the real one.

### 🟠 Significant #2 (evict ladder / reclaim discard the widened candidates) — ✅ VERIFIED RESOLVED, and the author's reasoning checks out

Not accepted on the claim. Traced: the guard at `:5194` requires `constraint is None`, which means `resident is None` **and** `pinned is None`. So `_fits`'s rebuild at `:5230` is `_gpu_candidates(probed, None)` = every GPU, the reclaim retry at `:5251` likewise, and `reclaim_key` falls through to `_worst_device_key`. The `preferred` path nulls `preferred` and never rewrites `candidates`, so there is nothing for the ladder to discard. The finding was pinned-only and died with the pinned block.

### 🟡 M1 (resolver mis-citation) — ⚠️ PARTIALLY RESOLVED

Pass 1's three checkable errors are fixed, re-probed:
- `git show b401d41f -- main.py | grep -c preferred` → **0** (unchanged); the comment no longer credits `b401d41f`. It now reads "#3097 (decided in #3107, implemented via #3165)" — correct: `c01ea35d` is #3165's commit, 15 `preferred` hits.
- The inverted #3165 citation is gone; so are "already applies" and "instead of `pinned`'s hard restriction".

**Still open** — see N7: the paragraph's own opening sentence, 20 lines above the corrected bullet, still states the pre-#3097 mechanism and now contradicts it.

### 🟡 M2 (four stale `main.py` docstrings) — ✅ MOSTLY MOOT, with one correction to pass 1 and one residue

With `pinned` hard again, `_gpu_candidates`, `_parse_device_hint` and the clone-voice route docstrings are all true as written. **Correcting pass 1:** I claimed A106's `main.py:5183-5187` citation would come to land on the new tolerance block. It doesn't, and it was *already* stale — on `origin/main` those same lines are the tail of `_resolve_admission`'s ASR docstring, not the hint logic. That is pre-existing drift, not something this PR caused. Residue is the run sheet's citations (N1).

### 🟡 M3 (`admit()` parity) — ❌ STILL OPEN, and unmentioned

`admit()` (`main.py:4920`) still has no tolerance check, and its docstring at `:4940-4944` still says `preferred` "tries that device first via the read-only `best_fit` … then falls back … if it doesn't fit", *"Kept in parity with `reservation()`/`_resolve_admission`"*. After this PR `reservation()` also drops `preferred` on the 75% test and `admit()` never does, so the stated parity is false. No production caller, so a false docstring rather than a live bug — unchanged from pass 1, and the PR body still doesn't mention it.

### 🟡 M4 (release-notes pair) — ❌ NOT ADDRESSED

Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff, and the updated PR body still contains no not-applicable statement. Additionally, the shipped #3058 entry in `docs/release-notes-next.md` now describes the mechanism wrongly: *"it becomes `reservation(preferred=...)`, which tries the hinted card first and then falls back to ordinary unconstrained best-fit."* After this PR it tests the 75% tolerance *before* trying the hinted card at all.

### 🟡 M5 (test gaps) — ❌ UNCHANGED, now mutation-measured

Mutating `get_headroom` to a constant so the tolerance can never fire (pre-#3097 "wins if it fits") reddens **exactly one** test:

```
FAILED tests/test_placement.py::test_preferred_tolerance_issue_repro_overrides_hint
```

`near_tie`, `single_gpu` and every route test stay green — guards, not discriminators, exactly as pass 1 said. Still no coverage for the `resident is None` guard and no ledger-hold case.

---

## New findings

### 🔴 N1 — Blocking (correctness, in the instrument) — the **ratified** `preferred` tolerance invalidates Criterion 2, the run sheet's only discriminating criterion, and no acceptance surface moves in this diff

`docs/testing/device-hint-placement-onbox-acceptance.md:116-315`

This is the same shape pass 1 found on the `pinned` path, now on the path that *is* ratified — so dropping code cannot resolve it.

Criterion 2 is the run sheet's own designated proof: *"Criterion 2 below is what actually proves the mechanism"* (`:86`). It deliberately constructs the band where **`cuda:1` still fits the derive but is not the roomier card** (`:171`: `target_headroom1 = peak + 200` = 3784 MB at the seed peak, with step 4 requiring `headroom0 > peak + 400`, and step 6 requiring `cuda:0` *"still be fully free"*). That band is precisely the state #3097's tolerance exists to override. Replayed on the box row A106 names (`cuda:0` 8192 idle, `cuda:1` 16303, peak 3584, this box's `GPU_RESERVE_MB=768` per the run sheet's own step 4):

```
  C2  free0=7500 headroom0=7090 headroom1=3784 bar=0.75*h0=5317.5 | hinted=cuda:0 ctrl=cuda:0 -> *** FAIL (same card) ***
  C2  free0=7000 headroom0=6590 headroom1=3784 bar=0.75*h0=4942.5 | hinted=cuda:0 ctrl=cuda:0 -> *** FAIL (same card) ***
  C2  free0=6500 headroom0=6090 headroom1=3784 bar=0.75*h0=4567.5 | hinted=cuda:0 ctrl=cuda:0 -> *** FAIL (same card) ***
  C2  discriminating ONLY while headroom0 in (3984, 5045.3]  ->  free0 in (4394, 5455.3]
```

Criterion 2's stated **Fail** condition (`:310`) is *"either run lands on the same card as the other"* — which is what both halves now do. It only still discriminates if `cuda:0` is itself ~60-70% occupied, directly contradicting its own step 6. Identical at `GPU_RESERVE_MB=500`.

**This is fixable in-round, not a design pass** — the band is re-derivable, just narrower (`headroom1` must now sit in `[0.75 × headroom0, headroom0)` and still clear `peak`). Verified:

```
  target_headroom1=3784 target_free1=4552 | hinted=cuda:0 ctrl=cuda:0 -> SAME CARD (criterion dead)  <-- current value
  target_headroom1=6000 target_free1=6768 | hinted=cuda:1 ctrl=cuda:0 -> diverge (criterion works)
  target_headroom1=6800 target_free1=7568 | hinted=cuda:1 ctrl=cuda:0 -> diverge (criterion works)
```

Three further consequences on the same surfaces, all unmoved in this diff:

1. **Criterion 3's diagnostic bullet is now unreachable** (`:393-399`). It prescribes `log.info("preferred=%s held=%s", preferred, held)` and expects `preferred="cuda:1"` with `held=None`. After this PR the tolerance nulls `preferred` *before* that try, so the restricted `try_hold` never runs — measured, the first `try_hold` sees `['cuda:0', 'cuda:1']` — and the line would read `preferred=None`. The criterion's other three bullets still pass, so this is the weaker half of N1.
2. **Every `main.py` line citation in the run sheet has drifted** by this PR's +35 lines: `:49` and `:275` (`main.py:5184`), `:326-327` (`:5184-5187`, `:5188-5189`), `:347`/`:349` (`:5187`, `:5189`), `:387` (`:5199`). All now land on comment text rather than the code they name.
3. **Register row A106's own body is now false** — `docs/testing/onbox-acceptance-register.md:4492`: *"a hint can equally park the derive on the exact card Qwen is generating on when that card merely fits (**the corrected comment at `clone-voice-resolver.ts:906-916` is the authority here**, not this row's earlier wording)."* That is the pre-#3097 behaviour, and the comment it names as the authority is the exact block this PR rewrites to say the opposite. The claim is now conditional at best ("equally" is wrong), and its cited authority contradicts it.

Per CLAUDE.md before-shipping step 3 the recording is a merge gate: the register row, the per-feature run sheet, **and** the live view (`docs/testing/onbox-acceptance-register-live-view.html:711` carries A106) all move in the shipping PR. None does.

### 🟠 N2 — Significant (correctness / house gate) — the Significant headroom fix shipped with **zero** regression coverage

`server/tts-sidecar/tests/test_placement.py:1538-1597`

Mutating `ReservationLedger.get_headroom` to `lambda …: free_mb` — byte-for-byte the pre-fix comparison this commit exists to replace — leaves the suite **fully green**:

```
[mutation] APPLIED: get_headroom -> raw free_mb (pre-fix semantics)
........................................................................ [ 54%]
...........................................................              [100%]     131 passed
```

A no-op control run of the same plugin (`MUT=0`) is also 131 passed, so the harness itself perturbs nothing. None of the three surviving tests holds a ledger reservation, so headroom and `freeMb` differ only by the per-device reserve — which never crosses the 75% bar in any of these fixtures. CLAUDE.md testing discipline: *"Bug fix → ship a regression test that fails before the fix and passes after."* The fix is correct and completely unprotected; `make(...)` already takes a `ledger`, and Probe 1's numbers above are a ready-made case.

### 🟡 N3 — Minor (cleanup) — the new route-test docstring's arithmetic uses the wrong reserve cap

`server/tts-sidecar/tests/test_xtts_clone_voice.py:1362-1365`

The docstring states GPU1's headroom as `14000-768=13232` against a threshold of `0.75 * (18000-768) = 12924`. But `768` is `test_placement.make()`'s `reserve_cap` default; this test drives `main._placement`, whose cap is `int(os.environ.get("GPU_RESERVE_MB", 500))` and no test sets it. Measured:

```
  main._placement.reserve_mb() = 500
  route fixture: cuda:0 headroom=17500  cuda:1 headroom=13500  threshold=13125.0  honored=True  margin=375.0
```

The conclusion (hint honored) holds; every number offered as evidence for it is from a different rule. Recurring shape #9.

### 🟡 N4 — Minor (cleanup) — `near_tie`'s adjusted fixture sits **exactly** on the bar, not "just above" it

`server/tts-sidecar/tests/test_placement.py:1559-1563`

The 13600 → 13692 adjustment was necessary (13600 gives headroom 12832 < 12924 and the test would have failed), and the docstring's inequality is literally true. But the margin is zero, not "just above":

```
  cuda:0 headroom=17232  cuda:1 headroom=12924  0.75*winner=12924.0   margin = 0.0
  cuda:1 free=13692 (as committed)       -> {'device': 'cuda:1'}
  cuda:1 free=13691 (committed minus 1)  -> {'device': 'cuda:0'}
```

Fine as a deliberate boundary test of `<` vs `<=`, but then say so — as written a reader cannot tell whether the test asserts the band or the inclusivity of its edge.

### 🟡 N5 — Minor (cleanup) — the new test block's section header still describes the deleted `pinned` mechanism

`server/tts-sidecar/tests/test_placement.py:1529-1536`

Left verbatim over three `preferred`-only tests: *"#3097 **pinned-device** 75 % tolerance"*, *"When a **pinned device** is set…"*, *"Otherwise **the full candidate set is used**"* (that is the removed `pinned` mechanism; `preferred` nulls itself and never widens `candidates`), *"the stale-cache / **operator-pin** failure mode"*. Every clause is now false, in the same hunk the commit edited.

### 🟡 N6 — Minor (cleanup) — the hinted route test's control no longer shares its probe

`server/tts-sidecar/tests/test_xtts_clone_voice.py:1409-1412`

`test_clone_voice_route_without_hint_header_is_unaffected_regression` says it is the *"Regression companion to the test above … **same probe as the hinted test**"*, but `2523e836` moved the hinted test's probe to `18000`/`14000` while this one stayed at `20000`/`10000`. Both still resolve to `cuda:0`, so no assertion breaks — but the pair no longer controls for the configuration it claims to. Recurring shape #8.

### 🟡 N7 — Minor (cleanup) — the resolver comment contradicts itself twenty lines apart (M1 residue)

`server/src/tts/clone-voice-resolver.ts:887-893` vs `:906-917`

The corrected bullet at `:906-917` is preceded by an opening sentence that was not updated: *"the sidecar threads `X-Device-Hint` into `reservation(preferred=...)`, which restricts its ONE try_hold to that card alone and **falls back to ordinary unconstrained placement only if the hinted card cannot fit the derive at all**"*, followed by *"That degrade-on-can't-fit is real, but it does not make a WRONG hint cheap in general — **only in the case where the hinted card is actually out of room**."* Both are the pre-#3097 contract, and the bullet immediately below now says the opposite. Half the pattern was mirrored.

### 🟡 N8 — Minor (cleanup) — `Closes #3097 #3165` links only #3097

GitHub needs a keyword per reference; `gh pr view 3207 --json closingIssuesReferences` returns **`3097` only**. Harmless in practice — #3165 and #3107 are already CLOSED — but the body states a linkage it does not create.

### 🟡 N9 — Minor (cleanup) — redundant second source of truth for the candidate list

`server/tts-sidecar/main.py:5195`

`all_gpus = self._gpu_candidates(devices, None)` is element-wise identical to `candidates` inside this branch, since the guard already requires `constraint is None`. Harmless today; it is a second copy of the same list that a future change to `candidates` would not track.

---

## Verdict

**Not mergeable yet — one Blocking, one Significant.**

Pass 1's Blocking and both Significant findings are genuinely resolved, verified by replay rather than by reading the diff: `pinned` is hard-honored again (Criterion 4 passes), the headroom comparison uses the same ledger accounting `try_hold` does (pass 1's repro now inverts correctly), and the discarded-candidates finding is correctly assessed as pinned-only. The code in this diff is, as far as I can measure, correct.

What blocks is that the change is unrecorded where this repo requires it to be recorded, and unprotected where it requires a test:

- **N1** — the ratified `preferred` tolerance makes Criterion 2 of the live on-box run sheet fail by its own explicit Fail condition on the exact box A106 names, invalidates Criterion 3's diagnostic bullet, staled six `main.py` citations, and falsifies A106's own body text — with none of the three acceptance surfaces moving. Unlike pass 1's Blocking this needs **no design decision**: the band is re-derivable (measured above), so per CLAUDE.md it is fixed this round, not filed.
- **N2** — the Significant headroom bug was fixed with no test that can detect the fix being reverted. Mutation-proven: 131/131 green against the pre-fix comparison.

M4 (release-notes pair, plus the now-inaccurate #3058 entry) and M3 (`admit()` parity docstring) remain open from pass 1; neither blocks on its own, but M4 is a checklist item with no not-applicable statement, and both are cheap to fold into the same round as N1/N2.

N2 is a correctness finding, so fixing and pushing re-triggers a pass 3.

dudarenok-maker:
## PR review — pass 3 (head fdafbd93, depth high)

Re-review of the 6-file diff after the five commits since `dba38813`. Deltas only — pass 1's and pass 2's solid items are not re-listed. Everything below was **executed** against this exact HEAD via read-only drivers and a pytest plugin in `$TEMP` (no repo file touched; `git status --porcelain` clean before and after; `rev-parse HEAD` = `fdafbd93` at both ends). CI is green on all 21 checks. Test scope run locally: `tests/test_placement.py` + `tests/test_xtts_clone_voice.py` → **132 passed** (was 131 — the new regression test).

---

## Pass-2 findings — disposition

### 🔴 N1 (Blocking — Criterion 2 invalidated) — ⚠️ **PARTIALLY RESOLVED.** The band is fixed and now discriminates; three of its four stated consequences are not.

**The re-derivation itself is correct — verified by replay, not by reading the number.** `target_headroom1 = 6000` with `target_free1 = 6000 + reserve(total1)` (6500 at `GPU_RESERVE_MB=500`, 6768 at this box's 768) makes both halves diverge on the hardware row A106 names, at every plausible idle `free0`:

```
cap=500 reserve0=410 reserve1=500 target_free1=6500 free0=7500 headroom0=7090 headroom1=6000 bar=5317.5 | hinted=cuda:1 ctrl=cuda:0 -> DISCRIMINATES
cap=768 reserve0=410 reserve1=768 target_free1=6768 free0=7500 headroom0=7090 headroom1=6000 bar=5317.5 | hinted=cuda:1 ctrl=cuda:0 -> DISCRIMINATES
cap=768 reserve0=410 reserve1=768 target_free1=6768 free0=8192 headroom0=7782 headroom1=6000 bar=5836.5 | hinted=cuda:1 ctrl=cuda:0 -> DISCRIMINATES
cap=768 reserve0=410 reserve1=768 target_free1=6768 free0=6500 headroom0=6090 headroom1=6000 bar=4567.5 | hinted=cuda:1 ctrl=cuda:0 -> DISCRIMINATES
swept: free0 range that discriminates = 6410 .. 8192  (headroom0 6000 .. 7782)
```

The `6500` / `6768` arithmetic in steps 4 and 5 is right on both sides, and the criterion's Fail condition ("either run lands on the same card as the other") is no longer met. Pass 2's core Blocking defect is genuinely closed.

**Criterion 3's diagnostic bullet — the claim is right, the instruction that produces it is not.** Replayed with a spy on `try_hold`, filling `cuda:1` below `peak`:

```
free1=3000 headroom1=2232 -> {'device': 'cuda:0'} try_hold candidate-lists seen: [['cuda:0','cuda:1']]
free1=2000 headroom1=1232 -> {'device': 'cuda:0'} try_hold candidate-lists seen: [['cuda:0','cuda:1']]
```

One `try_hold`, unconstrained — so `preferred` really is nulled before the try, and the new bullet's `preferred=None` is correct. But the instrumentation the bullet depends on is mis-placed (P2 below): the second log line it tells you to add never fires on the passing path.

**Consequence 3 (register row A106) — not addressed.** See P1.

**Consequence 2 (line citations) — addressed, and wrong.** See P2.

### 🟠 N2 (Significant — headroom fix shipped with zero coverage) — ✅ **VERIFIED RESOLVED**

Not accepted on the claim. Mutation-proved with a `$TEMP` pytest plugin, with a no-op control to show the harness perturbs nothing:

```
MUT=0  CONTROL: no patch applied                                    -> exit 0, 132 passed
MUT=1  get_headroom -> raw free_mb (byte-for-byte the pre-fix rule) -> exit 1, 1 FAILED
       FAILED test_placement.py::test_preferred_tolerance_ledger_reservation_reduces_effective_headroom
       E AssertionError: assert 'cuda:1' == 'cuda:0'
MUT=2  get_headroom -> constant (tolerance inert, pre-#3097)        -> exit 1, 2 FAILED
       FAILED ...::test_preferred_tolerance_issue_repro_overrides_hint
       FAILED ...::test_preferred_tolerance_ledger_reservation_reduces_effective_headroom
```

The new test reddens under exactly the mutation pass 2 used to prove the gap, and reddens **for the reason claimed** (`cuda:1` chosen where `cuda:0` is right), not incidentally. Its arithmetic checks out against the real `_headroom`: ledger hold 9000 on `cuda:1` → `min(16000, 16000−9000) − 768 = 6232` vs `cuda:0` `18000 − 768 = 17232`, bar 12924. M5's mutation count moves from 1 discriminating test to 2.

### 🟡 N3 (route docstring used the wrong reserve cap) — ✅ **VERIFIED RESOLVED**
`test_xtts_clone_voice.py:1362-1363` now reads `14000-500=13500` vs `0.75 * (18000-500) = 13125`. `main._placement`'s `reserve_mb` is `int(os.environ.get("GPU_RESERVE_MB", 500))` (`main.py:5707`) and `_device_reserve_mb` caps both cards at 500 here, so 17500 / 13500 / 13125 / margin 375 are the real numbers.

### 🟡 N4 (near-tie fixture sat exactly on the bar) — ✅ **VERIFIED RESOLVED**
13692 → 13700 gives `cuda:1` headroom 12932 against a bar of `0.75 × 17232 = 12924`, margin **+8 MB**, and the docstring now states that margin explicitly.

### 🟡 N5 (stale `pinned` section header) — ✅ **VERIFIED RESOLVED**
`test_placement.py:1529-1535` is rewritten to `preferred`. The surviving "operator-pin failure mode" clause is fine — it means an operator's `tts.qwen.device` pin making the hinted card the busy one (the resolver comment's own framing), not `COQUI_DEVICE`, which still hard-wins.

### 🟡 N6 (control test no longer shared its probe) — ✅ **VERIFIED RESOLVED**
Both `test_clone_voice_route_x_device_hint_overrides_best_fit_placement` (`:1373-1374`) and `..._without_hint_header_is_unaffected_regression` (`:1420-1421`) now probe `18000` / `14000`, and the pair asserts `cuda:1` vs `cuda:0` off one configuration.

### 🟡 N7 (resolver comment contradicted itself) — ✅ **VERIFIED RESOLVED**
`clone-voice-resolver.ts:887-896` now describes the tolerance and agrees with the bullet at `:907-916`. M1 is fully closed.

### 🟡 M3 (`admit()` parity docstring) — ✅ **RESOLVED**
`main.py:4941-4949` replaces the false parity claim with an explicit asymmetry NOTE. Independently confirmed the premise: `grep -rn "\.admit(" server/tts-sidecar/ --include=*.py` returns test call sites only, no production caller.

### 🟡 M4 (release-notes pair) — ⚠️ **HALF ADDRESSED**
`docs/release-notes-next.md:207`'s #3058 entry now describes the tolerance correctly. But `RELEASE_NOTES.md`'s matching user-facing line for the same feature — in the **in-progress 1.15.0 section** (`package.json` is 1.14.0, latest tag `v1.14.0`, so that section is unreleased) — still states the pre-#3097 rule: *"It's a request, not a rule: **if that card is unavailable**, or you've told Castwright yourself which card a voice engine should use, your choice still wins and the repair falls back to landing wherever it can."* After this PR the far more common trigger is "that card is merely busier than the alternative", which is the user-visible delta of this very PR. The PR body still carries no not-applicable statement for the pair either, and its Summary doesn't mention the release-notes edit at all.

### 🟡 N8 (`Closes #3097 #3165` links only #3097) — ❌ **STILL OPEN, unchanged**
Body still reads `Closes #3097 #3165`; `gh pr view 3207 --json closingIssuesReferences` still returns **`3097` only**. (#3165 is already CLOSED so nothing breaks, but the body states a linkage it does not create.)

### 🟡 N9 (redundant `all_gpus`) — ❌ **STILL OPEN, unchanged** (`main.py:5194`; cleanup only)

### 🟡 M5 (test gaps) — ⚠️ **IMPROVED, residue**
Two discriminating tests now (see N2's MUT=2). The `resident is None` guard at `:5193` still has no coverage.

## Pass-1 findings — still resolved

Re-checked because this round touched the same files. The `pinned` block is still absent (`git diff origin/main...HEAD -- main.py` contains only `get_headroom`, the `admit()` docstring, and the `preferred` block); the tolerance guard still requires `constraint is None`, so the evict ladder at `:5226-5232` and the reclaim retry at `:5249-5251` remain genuinely unconstrained on that path; M1/M2's resolver and `main.py` docstrings are consistent.

---

## New findings

### 🟠 P1 — Significant (correctness, in the instrument) — register row A106 and the live view still assert what this PR falsified, and name as their authority a comment that now says the opposite

`docs/testing/onbox-acceptance-register.md:4490-4494` and `docs/testing/onbox-acceptance-register-live-view.html:713` (verbatim in both):

> "…and a hint can **equally** park the derive on the exact card Qwen is generating on **when that card merely fits** (the corrected comment at `clone-voice-resolver.ts:906-916` **is the authority here**, not this row's earlier wording)."

The cited authority, as rewritten by this PR (`clone-voice-resolver.ts:913-916`), says:

> "A `cuda:1` that **merely fits no longer wins outright** while `cuda:0` sits nearly empty — only a `cuda:1` that is competitive with the alternative does."

A106 asserts X, names comment C as the authority for X, and C now asserts ¬X — in the same diff. "Equally" was true before #3097 and is the exact behaviour the tolerance exists to stop. (The citation `:906-916` is also now off by one — `:906` is the tail of the preceding bullet; the bullet begins at `:907`.)

A106 is **owed, not discharged**, and this PR changes the placement decision it exists to prove, so CLAUDE.md before-shipping step 3 applies to all three surfaces. The run sheet moved; the register and the live view did not. Fixing the live view means editing `onbox-acceptance-register-live-view.html` and publishing **that file** with `url = https://claude.ai/code/artifact/adf22b7b-12dd-49fe-874c-4a340585b26a`, after `npm run check:onbox-register -- --against-published <saved copy>` — per the register's own "Live view" section. No design decision is owed, so per CLAUDE.md this is fixed in this round.

### 🟠 P2 — Significant (correctness, in the instrument) — the citation refresh is wrong at **every** site it touched and absent at three more, and one of the wrong ones makes Criterion 3's diagnostic unobtainable

Measured against this PR's own **merge-base** (`352ff664`), every citation was *exact* before this PR; this PR's `main.py` insertions shift them all by **+33**. The refresh applied +10 / +34 / +35 instead of re-reading the file:

| run sheet | base (correct then) | doc now says | correct at HEAD |
|---|---|---|---|
| `:49`, `:276` "hint is skipped outright" | `5184` = `if preferred is not None and constraint is None:` | **`5194`** | `5217` |
| `:326` `_resolve_admission` span | `5126-5253` (def…return) | **`5134-5287`** | `5133-5286` |
| `:327` preferred `try_hold` | `5184-5187` | **`5219-5221`** | `5217-5220` |
| `:327-328` unconstrained fallback | `5188-5189` | **`5222-5223`** | `5221-5222` |
| `:348` insert log after preferred try | `5187` (the closing `)`) | **`5221`** | `5220` |
| `:350` insert log after fallback try | `5189` (the fallback call) | **`5223`** | `5222` |
| `:388` `_evict_until` | `5199` | **`5233`** | `5232` |

`:5194` is additionally the *wrong statement*, not just the wrong line: it is inside the tolerance block whose own guard is `constraint is **None**` — the opposite of the condition Setup step 5 is describing.

**The `:5223` one breaks a Pass bullet.** HEAD `:5222` is the fallback `try_hold`; `:5223` is `if held is None and not (cpu_capable and not heavy):`. The only parseable place to add a statement "right after `:5223`" is **inside** that branch, which is entered only when the fallback *failed*. In Criterion 3's passing scenario the fallback succeeds — my replay shows a single `try_hold` over `['cuda:0','cuda:1']` returning a hold on `cuda:0` — so `held is not None`, the branch is skipped, and `log.info("fallback held=%s", held)` **never prints**. The bullet requires "the second shows `held` on `cuda:0`". An operator following the sheet gets no second line and, per the bullet, would read that as a failure. (The `:5221` insertion survives by luck: it lands inside `if held is None:` and still prints `preferred=None`.)

**Three more sites this PR staled and the refresh missed entirely** — all exact at the merge-base:

- run sheet `:94` (Criterion 1's instrumentation point) — `main.py:12011` was `device_hint = _parse_device_hint(...)`; at HEAD that line is **12044**, and `:12011` is now blank.
- register A106 `:4483` — `main.py:5183-5187` → **`5216-5220`** at HEAD.
- register A106 `:4483` — `main.py:11994-11998` → **`12027-12031`** at HEAD.

**This corrects pass 2's own M2 note**, which said A106's citations were pre-existing drift. They were not: pass 2 compared against a `origin/main` that had itself moved past this branch's base. Measured against `352ff664` they were correct, and this PR staled them.

### 🟡 P3 — Minor (cleanup) — the run sheet's own "What is being accepted" still states the pre-#3097 contract

`docs/testing/device-hint-placement-onbox-acceptance.md:22-23`: *"the sidecar threads it into `reservation(preferred=...)`, which **tries that card first and falls back to ordinary unconstrained placement if it does not fit**."* This is the identical sentence the round corrected in `clone-voice-resolver.ts` under N7, left standing 145 lines above the criterion it re-derived. Half the pattern mirrored again — same shape as N7 itself.

### 🟡 P4 — Minor (cleanup) — step 4's re-derivation drops its own guard, leaves a dangling reference to it, and mis-states why 6000 is the number

`docs/testing/device-hint-placement-onbox-acceptance.md:166-175`, three parts:

1. **The dropped guard.** The old step 4 opened with *"this criterion needs `headroom0 > peak + 400` to have room to construct a discriminating band at all — if it doesn't, record the measured `headroom0` and `peak` in the Result line rather than forcing a pass."* Nothing replaces it. The new construction needs `0.75 × headroom0 ≤ 6000 < headroom0`, i.e. `headroom0 ∈ (6000, 8000]`; measured, `free0` must be **≥ 6410** or both halves land on `cuda:1` and the operator records a Fail against correct code. Step 2 still computes `headroom0` and nothing now consumes it.
2. **A dangling reference the same edit created.** Setup step 5 (`:78`) still says *"pick a 0.6B-only book if this render also has to compete with the `headroom0 > peak + 400` check at step 4"* — a check step 4 no longer contains.
3. **The stated reason for 6000 is false on this box.** `:171` and `:173-175` call it *"the 75%-tolerance threshold boundary"* and *"the concrete boundary where the tolerance switches from rejecting to accepting the hint."* At the box's idle `headroom0 = 7090` the real boundary is **5317.5**; 6000 is the boundary only at `headroom0 = 8000`, which an 8192 MB card cannot reach. The number works — it just isn't a boundary, it's a point with ~680 MB of margin, which is the property worth stating.

### 🟡 P5 — Minor (cleanup) — the new test's docstring states a headroom formula that only coincides here

`server/tts-sidecar/tests/test_placement.py:1587-1588`, mirrored in the inline comment at `:1610`: *"actual headroom after ledger reserve is `16000-768-9000=6232`"*. `_headroom` (`main.py:4558`) is `min(free_mb, total_mb − reserved) − reserve`, i.e. `min(16000, 7000) − 768`. The two agree only because this fixture sets `free == total`. Copied to a fixture where `free < total` the stated formula gives the wrong number — the same "right conclusion, wrong stated reason" shape N3 was about.

### 🟡 P6 — Minor (cleanup) — PR body Test plan is stale by the very test this round added

*"All 131 sidecar placement + route tests pass"* — it is **132** now, and the three listed bullets don't include `test_preferred_tolerance_ledger_reservation_reduces_effective_headroom`, which is the round's headline deliverable.

---

## Verdict

**Not mergeable yet — two Significant correctness findings, no Blocking.**

The **code** is in good shape and I could not break it: `pinned` is still hard-honored, the tolerance still uses the same ledger accounting `try_hold` does, and the headroom fix now has a regression test that is mutation-proven to redden against the exact pre-fix comparison (and only against it). Pass-2's N2 and all five of its minors are genuinely resolved, verified by execution rather than by reading the diff. Pass-1's findings remain resolved.

What is still wrong is entirely in the **acceptance instruments**, and both parts need no design decision, so per CLAUDE.md they are fixed in this round rather than filed:

- **P1** — register row A106 and the published live view still assert the pre-#3097 behaviour and cite as their authority a comment this PR rewrote to say the opposite. A106 is owed, this PR changes the behaviour it proves, and two of the three surfaces step 3 names do not move.
- **P2** — the citation refresh was done by offset, not by re-reading: all seven refreshed citations are off by 1-2 lines, one of them (`:5223`) makes Criterion 3's second diagnostic log line unreachable on a passing run, and three further citations this PR staled were missed entirely.

P3-P6 and the open N8/N9/M4/M5 residues are cleanup and would not on their own re-trigger a pass; folding them into the same round is cheap. P1 and P2 are correctness findings, so fixing and pushing re-triggers a pass 4.

dudarenok-maker:
## PR review — pass 4 (head 1e9fd06db64a063db0d4ff3f09ccb52fb5ac67ab, depth medium)

Reviewed at `1e9fd06d` against merge-base `352ff664`. Working tree unmodified (`git status --porcelain` empty before and after this pass).

### What I checked mechanically rather than assumed

- **The tolerance block is correct.** `all_gpus`'s max-headroom is monotone with `try_hold`'s fit-winner, so a `preferred` that *is* the unconstrained winner can never be dropped (`h < 0.75·h` is false for `h ≥ 0`), and a card whose headroom exceeds the winner's is impossible by construction. The all-cards-negative case *can* drop a winning `preferred`, but `try_hold` would have refused it anyway (`headroom >= peak` fails), so the resolved device is byte-identical. Float comparison is exact at these magnitudes. **No correctness bug in `main.py`.**
- **The paired tests discriminate.** Two counterfactual runs against the real module (harness in `$TEMP`, nothing written into the repo):
  - tolerance block neutralised → `test_preferred_tolerance_issue_repro_overrides_hint` and `test_preferred_tolerance_ledger_reservation_reduces_effective_headroom` go **red**; the other two stay green.
  - pass-2 bug restored (`get_headroom` → raw `free_mb`) → **only** `..._ledger_reservation_reduces_effective_headroom` goes red.
  So the pass-2 regression test is red for exactly the reason claimed, and `..._near_tie_honours_hint` does guard the `0.75` constant (red at `0.80`). All four green as shipped; fixture arithmetic in the docstrings (17232 / 12932 / 12924 / 6232) checks out against `_device_reserve_mb` at `reserve_cap=768`.
- **`admit()` has no production caller** — only `test_placement.py` and `test_asr_footprint_measurement.py`. The docstring's load-bearing claim holds.
- House gates: `Closes #3097` / `Closes #3165` present and outside code spans; release-notes pair both moved; no new config knob; no `cast.json` write; no derived artifact staled by the code change; A106 present on all three surfaces.

---

## Correctness bugs

### C1 — Criterion 2's fill target is a hardcoded constant whose validity the operator is never asked to check; the precondition it *is* asked to check doesn't gate it. Produces a **false FAIL against correct code.** — `docs/testing/device-hint-placement-onbox-acceptance.md:168-201` — **must-fix before merge**

Criterion 2 discriminates only inside `max(peak, 0.75 × headroom0) ≤ headroom1 < headroom0`. The hinted run needs `headroom1 ≥ 0.75 × headroom0` (tolerance) and `≥ peak` (fit); the unhinted control needs `headroom0 > headroom1`.

This pass replaced the derived target (`peak + 200`) with the constant `target_headroom1 = 6000` and left the operator-facing precondition at **`headroom0 > peak + 200` = 3784**. That gate does not imply either bound the constant depends on:

- **Lower bound violated.** `cuda:0` is the 8 GB card and this box runs the app's own frontend on it. Measure `free0 ≈ 6100` → `headroom0 = 6100 − 410 = 5690`. That passes the stated precondition (`5690 > 3784`, "with room to spare"), so the operator proceeds, fills `cuda:1` to `headroom1 = 6000`, and now `headroom1 > headroom0`. Step 8's unhinted control lands on `cuda:1`, matching step 6 → the documented **Fail** ("either run lands on the same card as the other") fires against correct code, after two full chapter renders.
- **Upper bound violated, and unrecomputable.** Step 1 explicitly invites recomputation for different cards. On a box whose `cuda:0` is 16 GB, `headroom0 ≈ 15500` → `0.75 × headroom0 = 11625 > 6000` → the tolerance *rejects* the hint, step 6's hinted derive lands on `cuda:0`, false FAIL again. Every other figure in step 4 is given as a formula (`reserve(total_mb)`, `target_free1 = target_headroom1 + reserve(total1)`); `6000` is the one value with no derivation, so there is nothing to recompute from.

Two supporting statements are also wrong as written, and are what let the omission through:

- `:180` — "the band defined by `0.75 × headroom0` and `peak`" names **two lower bounds**. `peak` (3584) sits *below* `0.75 × headroom0` (≈5300–5800 on this box), and the band's actual **upper** bound, `headroom0`, is omitted entirely — precisely the bound the fixed constant can cross.
- `:179` — "with ~680 MB of margin from the real boundary" is not derivable from anything in the document. The real boundary is `0.75 × headroom0`; for the margin to be 680, `headroom0` must be ≈7093, a figure that appears nowhere (step 2 leaves `headroom0` as the un-worked `free0 − 410`). A measured-once constant is being presented as a general property.

The fix is to state the band the operator must verify — `6000 < headroom0` *and* `0.75 × headroom0 ≤ 6000` — or to keep `target_headroom1` derived, e.g. `max(peak + 200, ceil(0.75 × headroom0) + margin)`.

### C2 — Nine `main.py:NNNN` citations left at their pre-PR values across all three acceptance surfaces; this PR shifted `main.py` by exactly +33 lines below 5216 — `docs/testing/*` — **must-fix before merge**

Pass 3 fixed seven citations, all of them inside the A106/tolerance prose cluster. Every citation *outside* that cluster was missed, and the whole live-view surface was missed. I verified each against `352ff664:server/tts-sidecar/main.py` (correct at base) and against HEAD (off by +33):

| file:line | cited | actually at HEAD | should be |
|---|---|---|---|
| `device-hint-placement-onbox-acceptance.md:132` | `main.py:9067` | `demoted = _parse_device(...)` | `9100` (`_DESIGN_IDLE_TTL_DEFAULT = 120.0`) |
| `device-hint-placement-onbox-acceptance.md:132` | `:9098` | a comment header | `9131` (`async def _qwen_design_idle_watchdog`) |
| `device-hint-placement-onbox-acceptance.md:133` | `:8141-8142` | a "near-empty audio" error string | `8174-8175` (the `unload_design()` guard) |
| `device-hint-placement-onbox-acceptance.md:162` | `main.py:11075-11172` | `"committed_mb": _process_commit_mb(),` | `11108-11205` (`@app.get("/debug/memory")` … `snapshot()`) |
| `device-hint-placement-onbox-acceptance.md:444` | `main.py:5787-5797` | `_parse_device`/rocm re-tag helpers | `5820-5830` (`def _validate_cuda_index`) |
| `onbox-acceptance-register.md:2493` | `main.py:8151`, `:8228` | `load_start = time.perf_counter()` | `8184`, `8261` |
| `onbox-acceptance-register-live-view.html:713` | `main.py:5183-5187`, `:11994-11998` | `:11994` is the **X-Audition-Text base64 decode** comment — unrelated code | `5216-5220`, `12027-12031` |

The live-view row is the worst of these on two counts: it is the surface an operator actually reads when running A106, its `:11994-11998` now points at an entirely unrelated block, and **this PR restamps the publish token (`data-published-as` 18 → 19)**, so the stale text is about to become the published page — while the markdown register's *identical sentence* was corrected. The `onbox-acceptance-register.md:2493` row is a different, live row that this PR staled as collateral.

Nothing mechanical catches this class: `check-register-citations.mjs` and `check-register-row-citations.mjs` validate row **IDs** only, and `check-onbox-register.mjs`'s content hash compares tracked-HTML against published-HTML, never markdown against HTML (its own header says the two documents are deliberately worded differently). That is why CI is green on it.

(I excluded the dated transcript directories — `onbox-*-results/`, `onbox-acceptance-staleness-audit.md` — which the repo's own frozen-path convention treats as historical records.)

---

## Cleanup nits

### N1 — PR title contradicts the PR — **must-fix before merge (title edit, zero cost)**
Title: *"apply 75% headroom tolerance to **pinned**/preferred device hints"*. The PR's first body paragraph says the pinned block was **removed** and pinned reverted to hard-honor. The merge commit carries the title permanently.

### N2 — `docs/release-notes-next.md:207` cites "PR #3165", which is an issue, not a PR — **must-fix before merge (cheap)**
`gh pr view 3165` → *"Could not resolve to a PullRequest"*; #3165 is the `agent-instructions` task issue. The neighbouring citation in the same sentence uses the convention correctly (`#3058, PR #3061`). As it stands, **PR #3207 appears nowhere in either release-notes file**, so the technical register loses the delivering-PR provenance that is its whole purpose. Should read `(#3097, PR #3207)`.

### N3 — `admit()` now disagrees with `reservation()` on the same inputs, reversing an invariant an earlier review deliberately installed — `server/tts-sidecar/main.py:4941-4949` vs `:4959-4962` — *follow-up acceptable*
The docstring this PR deleted argued parity was load-bearing ("an advisory decision function that silently dropped a parameter its binding twin honors would disagree with that twin the moment something calls it with a hint"); the replacement documents the asymmetry instead. Eleven lines below, the surviving `#2094 review R11` comment still calls `reservation()` "the binding decision this advisory `admit()` **must not disagree with**". I confirmed there is no production caller, so this is unreachable today — but the tolerance is ~10 lines and the class is otherwise left armed (defect shape 10).

### N4 — `main.py:5194` recomputes `candidates` under a different name — *follow-up acceptable*
`all_gpus = self._gpu_candidates(devices, None)` inside a block already guarded by `constraint is None`, where `candidates = self._gpu_candidates(devices, constraint)`. The two lists are identical by construction, so `len(all_gpus) > 1` and the winner loop could both read `candidates`. As written it implies the tolerance compares against a wider set than `candidates`, which it cannot.

### N5 — `test_preferred_tolerance_single_gpu_unaffected` does not guard what its docstring claims — `server/tts-sidecar/tests/test_placement.py:1632-1645` — *follow-up acceptable*
Verified by mutation: it stays green with the tolerance block neutralised *and* with the `len(all_gpus) > 1` early-out removed (single GPU ⇒ `preferred_headroom == winner_headroom` ⇒ `h < 0.75·h` is false). Its docstring says "tolerance check is skipped"; the assertion cannot observe whether it was. Harmless as an end-to-end smoke, but it is not the guard it is labelled as.

### N6 — Criterion 3's temporary-log-line instruction now depends on an indentation the doc doesn't state — `docs/testing/device-hint-placement-onbox-acceptance.md:354` vs `:403-408` — *follow-up acceptable*
"right after the `preferred` try at `main.py:5220`" was unambiguous before: `preferred` was never nulled, so the line printed wherever it went. Now the Pass bullet expects `preferred=None`, which only prints if the line is inserted at **method-body** indentation (after the `if` block closes at 5220). At the try_hold's own indentation it sits *inside* `if preferred is not None …`, which in Criterion 3's constructed state (cuda:1 filled below `peak`, so the tolerance drops the hint) never executes — the operator sees no first log line at all and the documented expectation is unobservable.

### N7 — `RELEASE_NOTES.md:10` puts a raw registry key in the user-facing notes — *follow-up acceptable*
`tts.coqui.device` is the only registry-key-shaped token in the entire file (every other backticked token is a filename, extension or hostname). The sentence it replaced — "you've told Castwright yourself which card a voice engine should use" — was in voice. The added "at least 75% of the best unconstrained alternative's free headroom" is likewise an implementation figure in a brand-voice note.

---

**Summary:** no runtime defect. Two documentation-correctness bugs on the acceptance surfaces this PR ships (C1 can produce a false FAIL against correct code; C2 mis-points nine citations, two of them on the page about to be published), plus two zero-cost must-fixes (N1 title, N2 wrong PR ref) and four genuine follow-ups.

PR #3201, opened 2026-09-11
fix(server,frontend): remove the retired analyzer engine knob from Advanced Settings

dudarenok-maker:
## Summary

`analyzer.engine` (env `ANALYZER`) stopped selecting the analyzer engine on 2026-07-15. `getResolvedAnalysisEngine` reads only the saved account setting. The knob stayed behind in three places:

- **Advanced Settings** still rendered an "Analyzer engine" control that did nothing. Its help text said `gemini` "always goes direct to the Gemini API".
- **The Ollama analyzer-device row** was shown only when that dead knob's effective value was `local` (`src/views/advanced.tsx`). A stale `ANALYZER=gemini` in an old `.env` hid the row for a user whose real engine is local.
- **Docs** still described `ANALYZER` as the engine control:
  - `INSTALL.md` told users to set `ANALYZER=local`;
  - the Model Manager wiki said the Advanced knob applies "when no per-account preference is set", with the wrong per-account default;
  - the Advanced Settings wiki listed the row;
  - plan 221 used `ANALYZER=local` as the fallback condition.

Changes:
- **Registry:** removes the knob and regenerates the managed `server/.env.example` block (`npm run config:sync`).
- **Device row:** gated on the saved account analysis engine (`account.analysisEngine`).
- **Docs:**
  - `INSTALL.md` and the wiki now point at Admin → Model Manager → Analyzer engine;
  - the fallback conditions now include Cloud fallback;
  - the Advanced Settings wiki row is removed;
  - plan 221's fallback wording is corrected.
- **Test fixtures** no longer carry the knob. `config.test.ts`'s "no device knob in patch" case now uses another non-device knob.

Found in passing while planning #3084 wave 3 (engine-literal inventory).

## Also fixed, found in passing

Review passes 1–6 surfaced a whole class of stale `Account → …` paths for the settings that moved to Admin → Model Manager on 2026-07-15 (fs-23 A7), plus the guard meant to catch them. All fixed on this branch:

[...]

PR #3215, opened 2026-09-11
fix(server,frontend): reject subset re-analysis chapter mismatch instead of silently joining

### LINKED ISSUE #3202: Subset analysis POST silently joins a running subset job for different chapters (decision owed: join, reject, or queue)
Found by the review gate on PR #3173 (pass 2). This behaviour predates that PR, which only corrects the comments that misdescribe it.

## Current behaviour

`POST /api/manuscripts/:id/analysis/chapters` (`server/src/routes/analysis.ts`, the subset route):
- validates the requested `chapterIds` against the manuscript (non-empty, known, not excluded);
- then, if a live subset job is already registered for that manuscript (`inFlightSubsetByManuscript`), **joins it**. It never compares the request's `chapterIds` with the running job's `subsetChapterIds`.

So while a chapter-3 subset job is running, a POST with `{ chapterIds: [7] }` returns 200. The server logs `[analysis-subset] subscribe` and streams the chapter-3 job's events, and the chapter-7 request is silently dropped.

By then the route has already sent its SSE headers, so it cannot answer with an HTTP 409.

`src/store/analysis-stream-middleware.ts` has long claimed "mismatched would 409". That was never implemented.

## Decision owed

When a subset request names chapters different from the subset job already running for that manuscript, what should happen?
- (a) **Keep joining**, which is today's behaviour, but tell the client: e.g. an SSE `warning` naming the chapters actually running.
- (b) **Reject.** Either move the running-job/id comparison before the SSE headers so the route can answer a real 409, or send a terminal SSE `error` with a code such as `subset_in_progress`, and have the view surface it.
- (c) **Queue or merge**: run the new chapters after, or add them to, the running job.

The options differ in UX, in what the frontend must handle, and in whether a user's click is ever silently dropped.

dudarenok-maker:
## Summary

Ratified decision from #3202: when a subset re-analysis request's chapter ids
don't match the manuscript's already-running subset job, the server now sends
a terminal SSE `subset_in_progress` error and does not join — instead of
silently attaching the new request to the wrong job and streaming/reporting
results for chapters it never asked about.

This round adds backend validation and frontend fixes on top of the
server-side change from HEAD commit:

1. **Frontend crash fix** — the analysing and generation views were calling
   `store.getState()` without passing a store context, causing a ReferenceError.
   Both views now hold the store as a hook-injected ref and call through it
   safely.

2. **Frontend teardown guard (C1)** — when a rejected subset request tries to
   clear the active stream, the middleware's sticky subscriber guard now
   correctly gates the clear and lets the prior job's ticks continue to update
   the cross-view snapshot.

3. **Re-arm guard for subset_in_progress (C2)** — when the subset rejection
   lands, the main run is NOT re-armed. Suppressing the re-arm alone leaves a
   stale snapshot permanently frozen at the rejected chapter ids (B2).

4. **Pre-POST snapshot restoration (B2)** — the snapshot is captured before
   each optimistic `setActiveStream` call (Include, Re-analyse, and Retry
   flows). On `subset_in_progress` rejection, the prior snapshot is restored
   so the UI does not display a run that is not actually happening. Three sites
   fixed: `generation.tsx` Include (line 547-551), `generation.tsx` Re-analyse
   (line 708-716), `analysing.tsx` Retry (line 929-933).

5. **Paused state on retry collision (B3)** — when a subset rejection occurs
   in the analysing view after pausing the main run, the restored snapshot is
   set to `state: 'paused'` instead of preserving the original `state: 'running'`.
   This prevents the stall-detection logic from incorrectly marking the pill
   as stalled.

6. **Release note repair (N5)** — "Casting" → "Resuming" in RELEASE_NOTES.md.

7. **N3/N4 assertions (N7)** — log line assertion for the `subset_in_progress`
   outcome, and empty-title fallback test for the error message.

8. **Test coverage for B2, B3, B4** — regression tests verify that snapshot
   restoration works for all three code paths (Re-analyse, Include, Retry).

Two-scope change: server and frontend, plus test coverage for the rejection
cases.

Closes #3202.

## Test plan

- [x] `npx vitest run server/src/routes/analysis.test.ts` (server) — 223/223 pass
- [x] `npx vitest run server/src/routes/analysis.request-log.test.ts` — 11/11 pass (fixed N8: empty-fallback test now uses unresolvable chapter ids only)
- [x] `npx vitest run src/views/generation.test.tsx` — 84/84 pass (added B4: Include flow regression test)
- [x] `npx vitest run src/views/analysing.test.tsx` — 74/74 pass (extended C2 test with B3 assertion)
- [x] `npx tsc --noEmit` at repo root — clean
- [x] `npx tsc --noEmit` in `server/` — clean
- [x] Regression tests for C1 (re-arm guard, covered in `analysing.test.tsx`), C2 (re-arm),
      B2 (snapshot restore), B3 (paused state), B4 (Include + Retry): written and
      verified to redden when fixes are temporarily reverted. (No separate
      `analysis-stream-middleware.test.ts` changes — C1's coverage lives in the
      view-level tests above.)


dudarenok-maker:
## PR review — pass 1 (head 7cc39f0d, depth high)

Server-side guard is correct and fails closed. The two findings below are both in the frontend: the views treat `subset_in_progress` as a terminal failure *of a run*, and run the failure teardown — but this rejection means "your request never started a job, and nothing about the other run changed". The teardown is therefore aimed at somebody else's live job.

### 🐞 Correctness bugs

**C1 — a rejected collision tears down the *other*, still-running subset's global snapshot and kills the middleware's sticky subscriber.**
`src/views/generation.tsx:533` (include flow), `src/views/generation.tsx:674` (re-analyse flow), and `src/views/analysing.tsx:966` (retry `.finally`).

Concrete scenario, entirely inside the Generate view, no navigation and no second tab:

1. Click **+ Include in book** on chapter 3. A subset job `[3]` starts; `generation.tsx:423` seeds `setActiveStream({ kind:'subset', subsetChapterIds:[3] })`; the middleware opens its sticky subscriber on the first tick.
2. A second later, click **Re-analyse** on chapter 5. The idempotency guard at `generation.tsx:397` / `:583` is keyed on `subsetByChapter[chapterId]`, i.e. per chapter — a *different* chapter is not blocked, so the POST fires. This is exactly the collision the PR now rejects.
3. Server sends `{kind:'error', code:'subset_in_progress'}`; `src/lib/api.ts:5707` throws `AnalysisError(msg, 'subset_in_progress')`.
4. The catch at `generation.tsx:671` runs `dispatch(analysisActions.clearActiveStream())` at **line 674 — above** the new branch's early return at `:685`. (Same ordering in the include flow: clear at `:533`, new branch at `:541`.)

Effect: `analysis-stream-middleware.ts:270-275` closes the handle on `CLEAR_TYPE`, aborting the sticky SSE subscription to job `[3]`; `analysis.activeStream` becomes `null`, and `applyAnalysisSnapshotTick` no-ops on a null snapshot (`src/store/analysis-slice.ts:130-131`). Job `[3]` keeps running server-side (sticky — `server/src/routes/analysis.ts:6726-6733`), but the AnalysisPill is gone for the rest of the session: nothing re-seeds `activeStream` in-session, the only re-seed being layout's once-at-mount cold-boot hydrate (`src/components/layout.tsx:604-644`). Navigate away and the live run is invisible everywhere.

This is introduced by this PR: pre-fix, step 2's POST joined job `[3]` (the bug being fixed), the promise did not reject, and the snapshot survived.

Fix: move the `subset_in_progress` early return **above** `clearActiveStream()` in both `generation.tsx` catches, and skip the clear in `analysing.tsx`'s `.finally` for this code. A request the server refused before starting a job must not touch global snapshot state at all. Related and worth the same round: `setActiveStream` at `generation.tsx:423` / `:601` and `analysing.tsx:801-816` clobbers the live run's `subsetChapterIds` *before* the POST — pre-existing, but a refused request now leaves that clobber uncompensated.

**C2 — the rejection re-arms the main analysis run while a foreign subset job is still live, which is the cache-clobber race the view's own PAUSE-AND-RETRY contract exists to prevent.**
`src/views/analysing.tsx:938-961`, contract comment at `:751-762`.

Scenario:

1. Start **Re-analyse chapter 5** from the Generate view. The server job is sticky and survives both navigation and client disconnect (`analysis.ts:6726-6733`).
2. Navigate to the analysing view with the main run streaming. A chapter fails; click **Retry** on chapter 3.
3. `pausedMainForRetry` is true (`analysing.tsx:784-789`), so the main SSE is aborted and the subset POST fires.
4. The POST is now rejected *instantly* with `subset_in_progress`, so `.finally` re-arms the main run immediately (`:938-961`) — while job `[5]` is still writing.

`analysing.tsx:751-762` states why that is not allowed: "both routes load their own snapshot of the disk-backed analysis cache and write back independently, so the second-finisher's stale view of `cache.failedChapterIds` / `cache.chapters` silently clobbered the first's progress… after a successful retry, reload restored the failed rows and showed previously-attributed chapters as still-not-parsed." Nothing on the server prevents it either: main and subset live in separate maps, `markAnalysisBusy` is a refcount for the rename/design locks rather than a run gate, and only `/pause` aborts both (`analysis.ts:6461-6482`).

Before this PR the colliding retry joined job `[5]` and main stayed paused until it ended. `subset_in_progress` is the only rejection code that *guarantees* another analyzer job is live at the moment the resume fires (`bad_request` / `chapter_excluded` / `design_in_progress` imply no such thing), so this is a new reachability, not a pre-existing one.

Fix: on `subset_in_progress`, do not re-arm the main run — leave it paused and surface the message; the user has to wait for the other subset either way.

### 🧹 Cleanup

**N1 (house gate) — the release-notes pair is missing and no not-applicable is declared.** Neither `docs/release-notes-next.md` nor `RELEASE_NOTES.md` is in the diff. The delta is user-visible (a click that used to silently return another chapter's results now shows "A different subset re-analysis is already in progress for this manuscript: …"), the 1.15.0 section of `RELEASE_NOTES.md` is open, and its **first bullet is the sibling #3169/#3173 entry** — this belongs next to it. Before-shipping step 5 permits a skip only with an explicit N/A in the PR body; the body says nothing.

**N2 — a doc sentence this change made false.** `docs/features/archive/32-sticky-analysis.md:185` still says "concurrent re-POSTs attach as second subscribers via the shared `replayCatchUp` helper" — now true only for a matching chapter set. Same class of stale prose #3173 was opened to correct. If archived plans are treated as historical ship records here, say so explicitly and close it rather than leaving it undecided.

**N3 — the reject path logs no outcome line.** `server/src/routes/analysis.ts:6644-6655` sends the error and returns without any `[analysis-subset]` outcome line, so a collision reads in the server log as `request received` then silence — the shape #3169's D2/F2 wave set out to make distinguishable from a wedge. The branch sits inside the describe that asserts that contract (`analysis.request-log.test.ts:235`), with test (j) asserting a `subscribe` line and test (i) asserting none. The sibling error paths (`unknown_manuscript`, `design_in_progress`, `bad_request`, `chapter_excluded`) are equally silent, so the new branch widens a pre-existing gap rather than opening it.

**N4 — the error message can degrade to a dangling colon.** `analysis.ts:6645-6651` filters out any running id `hintsById` cannot resolve and has no fallback, so a record re-hydrated with different chapter ids (or the absent-`subsetChapterIds` case the `?? []` guards) yields "…already in progress for this manuscript: **.** Wait for it to finish…". A raw-id fallback, or a generic clause when the list is empty, is one line. The guard itself fails closed in that case, which is the right direction.

### Checked and clean

- **Set comparison is symmetric with what registration stores.** `toRun.map(t => t.id)` at `:6708` vs `new Set(toRun.map(t => t.id))` at `:6639` — same normalisation, so dedupe and unknown-id dropping (`targets` at `:6597-6599`) apply identically on both sides. A request of `[3, 999]` against a running `[3]` still joins, correctly.
- **The guard fails closed on absent evidence** — `subsetChapterIds` undefined ⇒ empty running set ⇒ mismatch ⇒ reject, never join.
- **The new code is reachable as shipped.** `src/lib/api.ts:5707` propagates `payload.code ?? 'unknown'`, so the three view branches really do see `code === 'subset_in_progress'`. Both frontend test files mock `../lib/api` (generation.test.tsx supplies its *own* `AnalysisError` class at `:115`), so this integration is the one thing they cannot prove — verified by reading api.ts.
- **The middleware reattach genuinely matches**, so the rewritten comment at `analysis-stream-middleware.ts:154-158` is accurate: all three real callers post `[chapterId]` and seed the snapshot with the same single id (`analysing.tsx:814`/`:828`, `generation.tsx:436`/`:444`, `:614`/`:618`), and the handle identity is `(manuscriptId, kind)` so a colliding `setActiveStream` does not re-fire a subscribe POST.
- **The paired tests can fail, for the claimed reason.** Test (i): with the comparison neutered, the request joins the live stub whose stream never terminates ⇒ supertest hangs ⇒ timeout, and `subscribers.size` is a second independent way to redden. Test (j): an array-equality comparison reddens it, which is what makes the order-independence claim real.
- **Ratified decision honoured** — option (b) narrow variant per #3202's decision comment, terminal SSE `error` rather than a 409, join-on-match unchanged, plus the middleware comment repair that comment explicitly asked for. `Closes #3202` present and outside a code span.
- **Not applicable, checked:** no `cast.json` writes and no lock-timeout seam; no config knob; no derived artifact (openapi's subset-route `200` `oneOf` enumerates no error events for *any* existing code, so this introduces no contract drift); no on-box acceptance owed — nothing here needs a GPU, sidecar, real analyzer or real book.
- **Tree check:** `git status --porcelain` empty before and after this pass; HEAD unchanged at `7cc39f0d`.

dudarenok-maker:
## PR review — pass 2 (head 02c6cd31, depth high)

Deltas only. Pass-1 items marked "Checked and clean" were re-checked only where this round's two commits (`a93ebc64`, `02c6cd31`) touched them.

| # | Disposition |
|---|---|
| **C1** — rejected collision tears down the other job's snapshot / sticky subscriber | **PARTIALLY RESOLVED — still open.** The three `clearActiveStream()` sites are correctly gated, but C1's other half — the pre-POST `setActiveStream` clobber, which pass 1 named in the same finding as "worth the same round" — is untouched, and in the `analysing.tsx` path it now produces a new wrong state. See **B2**. |
| **C2** — rejection re-arms the main run while a foreign subset is live | **VERIFIED RESOLVED** (mechanism). Untested — see **B1**. |
| **N1** — release-notes pair missing | **VERIFIED RESOLVED**, with a copy defect — see **N5**. |
| **N2** — stale sentence in `32-sticky-analysis.md` | **VERIFIED RESOLVED.** |
| **N3** — reject path logs no outcome line | **VERIFIED RESOLVED**, unasserted — see **N7**. |
| **N4** — dangling colon in the error message | **VERIFIED RESOLVED**, unasserted — see **N7**. |

**C1 — what I verified.** `src/views/generation.tsx:529-548` (include) and `:678-698` (re-analyse) now dispatch `clearActiveStream()` only inside the `isAbort` branch and the generic terminal-failure tail; the `subset_in_progress` branch returns before either. `src/views/analysing.tsx:948-950` returns out of `.finally` before both the re-arm and the `else` clear. Traced against `analysis-stream-middleware.ts:270-275` (`CLEAR_TYPE` → `closeHandle`) and `:277-295` (displacement on `(manuscriptId, kind)`): in the **generation.tsx** scenario pass 1 described — job `[3]` live, `Re-analyse` on chapter 5 rejected — the colliding `setActiveStream` keeps `manuscriptId` and `kind: 'subset'`, so the displacement guard does not fire, no `CLEAR_TYPE` is dispatched, and the handle on job `[3]` genuinely survives. Job `[3]`'s own ticks then restore `phaseId`/`phaseProgress`/`lastTickAt` via `applyAnalysisSnapshotTick`. That path is fixed. The `analysing.tsx` path is not — see B2.

**C2 — what I verified.** `subsetInProgressRef` (`analysing.tsx:196`) is reset to `false` at `:773`, at the top of every `handleRetryChapter` and after the `retryingChapterId !== null` re-entry guard at `:770`; it is set `true` only inside `if (err instanceof AnalysisError && err.code === 'subset_in_progress')` at `:919-920`. No other rejection code reaches it, so `bad_request` / `chapter_excluded` / `design_in_progress` / "no result" still re-arm as before — the specific suppression risk you asked about is not present. The guard sits above the `pausedMainForRetry` block at `:958-981`, which is the exact site pass 1 named. Re-entrancy is safe: `setRetryingChapterId(null)` at `:940` is an async state update, so no second `handleRetryChapter` can run between it and the ref read at `:948`. I also checked the only other resume path, the `castIncomplete` auto-resume effect at `:744-749` — the `subset_in_progress` catch appends a `failedChapters` row at `:921-924`, so `failedChapters.length > 0` short-circuits it. C2 is genuinely closed.

**N3/N4 — what I verified.** The log line at `analysis.ts:6658` sits between `send(...)` and `clearInterval`/`res.end()` inside the `!sameChapterSet` branch, so it executes on exactly the rejection and nowhere else; its shape `[analysis-subset] <outcome> manuscript=${JSON.stringify(manuscriptId)}` matches the `subscribe` line at `:6666` and the `start` line at `:6751`. It does not break test (i) (which asserts response text and `subscribers.size`, not log absence) or test (j) (whose `subset_in_progress` absence assertion is on the *matching* path) — `analysis.request-log.test.ts` 10/10 green at this head. N4: `titlePart` at `:6649` collapses to `''` when `runningTitles` is empty, producing "…for this manuscript. Wait for it…" with no dangling fragment.

---

### 🐞 Correctness bugs

**B1 (house gate, blocking) — neither C1 nor C2 has a regression test. Both fixes are green before and green after, so nothing in the suite holds them.**
`src/views/generation.test.tsx:2057-2068`, `:2151`, `:2314`; `src/views/analysing.test.tsx:1536-1541`.

This is not an inference from the commit messages — it is the diff:

```
$ git diff 7cc39f0d..02c6cd31 --stat
 RELEASE_NOTES.md                            |  2 ++
 docs/features/archive/32-sticky-analysis.md |  2 +-
 docs/release-notes-next.md                  |  1 +
 server/src/routes/analysis.ts               |  8 +++++++-
 src/views/analysing.tsx                     | 22 +++++++++++++++++++++-
 src/views/generation.tsx                    | 26 ++++++++++++++++++--------
```

Zero test files. The "82 and 73 tests passing" figure reported for this round is the **same** figure this PR's own body records against `7cc39f0d` — i.e. against the code *before* the C1/C2 fix. I re-ran both at this head from inside the worktree: `src/views/generation.test.tsx` 82/82, `src/views/analysing.test.tsx` 73/73 — identical counts, because the files are byte-identical to their `7cc39f0d` versions.

The three `#3202` tests that do exist were added in `7cc39f0d` and assert only that the server's message reaches the row. They are **structurally incapable** of observing either fix:

- `generation.test.tsx`'s `makeIncludeStore` (`:2057-2068`) registers `ui, chapters, manuscript, changeLog, cast, library, queue, bookMeta` — **no `analysis` reducer**. `analysisActions.clearActiveStream()` dispatched into that store is an inert no-op. (The `activeStream` parameter at `:2052-2055` is the *generation* stream — `{bookId, modelKey}` — not the analysis snapshot.)
- `analysing.test.tsx:1540-1541` builds `configureStore({ reducer: { ui, cast, account, bookMeta } })` — **no `analysis` reducer** either, and it asserts nothing about the main run being re-armed. Worse for C2: in that test `analysisStarted` is false and no main SSE is in flight, so `pausedMainForRetry` (`analysing.tsx:791-792`) evaluates **false** and the re-arm branch the fix guards is never entered at all.
- Neither file wires `analysisStreamMiddleware`. `git grep analysisStreamMiddleware src/` returns exactly one test file — `src/store/analysis-stream-middleware.test.ts` — and this PR does not touch it. So the sticky-subscriber teardown, which *is* C1's mechanism, has no coverage anywhere in this diff.

This is defect shape #3 ("a test that cannot fail") in its purest form, and it is a CLAUDE.md merge gate, not a nicety: "Bug fix → ship a regression test that fails before the fix and passes after." Two correctness bugs found by an adversarial review round were fixed with subtle ordering and ref-guard changes — precisely the kind that a later refactor re-breaks silently — and nothing in CI would notice.

Concretely reproducible remedies, both cheap:
- **C1**: `src/store/analysis-stream-middleware.test.ts` already has the harness (`:103` concats the real middleware). Seed `setActiveStream({kind:'subset', subsetChapterIds:[3]})`, tick to open the handle, then dispatch a second `setActiveStream({kind:'subset', subsetChapterIds:[5]})` followed by *no* `clearActiveStream`, and assert the handle's `AbortController` was not aborted. Revert `generation.tsx:529-548` and it reddens.
- **C2**: add `analysis: analysisSlice.reducer` to the store at `analysing.test.tsx:1540` (the file already imports `analysisSlice` at `:9` and uses it in a dozen other stores), drive the view into a streaming main run so `pausedMainForRetry` is true, reject the retry with `subset_in_progress`, and assert `api.analyseManuscript` is **not** called again and `analysis.activeStream` still carries `kind: 'subset'`. Revert `analysing.tsx:948-950` and it reddens.

**B2 (correctness, medium) — C1's other half is unfixed: the pre-POST `setActiveStream` clobber is still uncompensated, and in the `analysing.tsx` path the rejection now leaves the pill permanently claiming a subset that is not running, with no subscriber anywhere.**
`src/views/analysing.tsx:808-823` (the clobber), `:948-950` (the guard that now preserves it), `src/store/analysis-stream-middleware.ts:288-293` (the displacement close), `src/components/layout.tsx:1427-1436` (where it surfaces).

Pass 1 named this in the same finding as C1 — "*`setActiveStream` at `generation.tsx:423` / `:601` and `analysing.tsx:801-816` clobbers the live run's `subsetChapterIds` before the POST — pre-existing, but a refused request now leaves that clobber uncompensated. Related and worth the same round*". Only the `clearActiveStream()` half was addressed. Suppressing the clear without restoring the snapshot is what creates the new state:

1. Subset job `[3]` is live server-side (sticky — started from Generate's Include, or a prior Retry; it survives navigation and client disconnect, `analysis.ts:6732-6738`).
2. User is on the analysing view with the main run streaming: `analysis.activeStream = {kind:'main', state:'running'}`, middleware handle `= {m1, 'main'}`.
3. User clicks **Retry** on chapter 5. `pausedMainForRetry` is true (`:791-792`) → the main SSE is aborted and `analysisStarted` goes false (`:793-796`). Then `:808-823` dispatches `setActiveStream({kind:'subset', subsetChapterIds:[5], state:'running', phaseId:0, phaseProgress:0, lastTickAt: now})`.
4. The middleware's displacement guard fires on the `main → subset` **kind** change (`analysis-stream-middleware.ts:290-292`) and calls `closeHandle()`. Handle is now `null` — *before* the POST is even sent, so the new `subset_in_progress` gating in `.finally` cannot save it.
5. The POST is rejected. The guard at `:948-950` correctly skips the re-arm — and also skips the `else` branch's `clearActiveStream()` at `:986`.

End state: `analysis.activeStream` is frozen at `{kind:'subset', subsetChapterIds:[5], state:'running', phaseProgress:0, lastTickAt:T0}`. The middleware handle is `null` and can never reopen — `openHandle` is only reachable from `APPLY_TICK_TYPE` (`:297-308`), and no SSE exists in this client to produce a tick. `layout.tsx:1427-1436` derives `stalled` from `state === 'running' && Date.now() - lastTickAt > STALL_THRESHOLD_MS`, so the top-bar pill renders "Retrying · 0%" and then flips to **stalled**, permanently, naming a subset that is not running — while the subset that *is* running (`[3]`) is invisible everywhere. Clicking the pill navigates to the analysing route, which shows a Resume button and no evidence of job `[3]`.

At pass-1 head this same sequence produced a *disappearing* pill (the `clearActiveStream()` that was C1) plus the C2 race. C2 is genuinely fixed; the cost is a false permanent "stalled" status, which is a new wrong output introduced by this round.

Two mitigating facts, stated so severity isn't inflated: (a) it self-heals the moment the user clicks **Resume analysis** (`analysing.tsx:1240-1254`), because the main analysis effect dispatches its own `setActiveStream({kind:'main'})` — at the cost of re-opening the very race C2 closed, now as an explicit user action rather than an automatic one; (b) the same clobber in `generation.tsx:423-438` / `:605-620` is currently **benign** only by coincidence — `layout.tsx:1447` reads `subsetChapterIds.length`, never the ids, and every real caller posts exactly one id, so `[5]` and `[3]` render identically. That is defect shape #10: the class is armed at all three sites and the two in `generation.tsx` are one multi-chapter caller away from surfacing, plus `analysis-stream-middleware.ts:162` would re-POST the stale ids on any future handle reopen — which the server now *rejects*.

The fix is the one pass 1 pointed at: a request the server refused before starting a job must not have mutated the global snapshot in the first place. Either seed the snapshot after the POST is known to have started a job, or capture the prior snapshot and restore it on `subset_in_progress` (the same three sites, plus the `analysing.tsx` kind-flip).

---

### 🧹 Cleanup

**N5 — the user-facing release note has a garbled clause and over-claims one of its two promises.** `RELEASE_NOTES.md` (1.15.0 section, the `#3202` bullet). "**Casting** a main analysis job to resume while that retry is happening is also blocked now" — "Casting" is not a word that works here (Causing? Resuming?), and it lands in the brand-voice document users read. Separately, "the running retry stays visible and on-screen so you can wait for it or cancel it" is true for the Generate-view collision and false for the analysing-view one per **B2**, where the running retry becomes invisible and the pill goes stalled. Placement and section are correct (top in-progress `# Castwright 1.15.0`, adjacent to the sibling #3169/#3173 entry), and `docs/release-notes-next.md:338` is correctly inside `## 🗣️ Analyzer, script review & manuscript` — N1 is otherwise discharged.

**N6 — the PR body no longer describes the PR.** Its Summary and Test plan still describe only the `7cc39f0d` server-side change; nothing mentions the frontend teardown fix, the `subsetInProgressRef` re-arm guard, the new log line, or the message repair, and the recorded 82/73 counts now read as if they covered this round's fixes when they predate them. CLAUDE.md's house gate — "incidental findings fixed in this round … declared in the PR body" — is unmet, and a reviewer arriving cold is pointed at the wrong diff.

**N7 — N3's log line and N4's empty-title fallback are both unasserted.** `analysis.ts:6658` adds an outcome line to the one file whose `D2/F2 (#3169)` describe block exists specifically to pin log-line shapes; test (i) at `analysis.request-log.test.ts:524` does not spy on `console.log`, so nothing holds it. `analysis.ts:6648-6649`'s `titlePart` fallback is the branch that only fires when `hintsById` cannot resolve the running ids — the exact case N4 was raised about — and test (i) exercises only the resolvable path (`'Chapter 3'`). Both are one assertion each on tests that already exist.

---

### Verdict

**Not mergeable as it stands.** C2, N1–N4 are discharged and verified. C1 is half-discharged: the teardown is correctly gated at all three sites, but the companion clobber pass 1 flagged in the same finding is untouched and now produces a permanently-stalled pill in the analysing-view collision (**B2**). And independently of B2, **both** C1 and C2 shipped with no regression test at all (**B1**) — the fixes are green-before and green-after, which is the repo's own merge gate, not a style preference.

**Tree check:** `git status --porcelain` empty in `C:\Claude\Projects\wt-3202-subset-chapter-mismatch` before and after this pass; `HEAD` unchanged at `02c6cd318c1cb8e6e2601d771dc876178f1d48f4`. No tracked file modified.

dudarenok-maker:
## PR review — pass 3 (head 8c37c9ed, depth high)

Deltas only. Everything below was re-derived from the tree at this head — nothing was taken from the round summary or the PR body. Every disposition that says "verified" was measured by reverting the specific fix and re-running the suite, not by reading the test.

| # | Disposition |
|---|---|
| **B1** — neither C1 nor C2 had a regression test | **VERIFIED RESOLVED for C1 + C2.** Both new tests are mutation-proven red for the claimed reason. But the *same* gate is now unmet for this round's own B2 fix at two of its three sites — see **B4**. |
| **B2** — pre-POST `setActiveStream` clobber uncompensated | **PARTIALLY RESOLVED — still open in the `analysing.tsx` path.** Both `generation.tsx` sites are genuinely fixed. The `analysing.tsx` site restores the *snapshot* but not the *subscription*, so the permanently-stalled pill B2 named survives there — see **B3**. |
| **N5** — garbled clause + over-claim in `RELEASE_NOTES.md` | **HALF RESOLVED.** "Casting" → "Resuming" is fixed and reads correctly. The over-claim half is unchanged and still false for the analysing-view path — see **N9**. |
| **N6** — PR body no longer describes the PR | **RESOLVED, with new inaccuracies introduced — see N10.** |
| **N7** — N3's log line and N4's fallback unasserted | **N3 VERIFIED RESOLVED. N4 NOT RESOLVED** — the test added for it does not reach the branch, proven by mutation — see **N8**. |

**B1 — what I measured.** Four mutations, each run from inside the worktree at this head:

| Mutation | Result |
|---|---|
| Neuter the `subsetInProgressRef` re-arm guard (`analysing.tsx:961`) | `analysing.test.tsx` **1 failed / 73 passed** — `#3215 C2`, `AssertionError: expected {…(14)} to be undefined`, i.e. a second `analyseManuscript` landed. Red for exactly the claimed reason. |
| Delete the `priorSnapshot` restore at **all three** sites | `generation.test.tsx` **1 failed / 82 passed** — `#3215 C1/B2`, `expected {…(11)} to deeply equal {…(9)}`. **Only one test reddens** → see B4. |
| Disable only `handleReanalyse`'s `subset_in_progress` branch (`generation.tsx:708`) | Same single test reddens, `expected null to deeply equal {…}` (the generic tail's `clearActiveStream()` ran). This confirms the routing claim: `chapter-row-1-reanalyse` (`generation.tsx:1905`) → `setReanalyseChapter` → the confirm modal at `:1262-1281` → `handleReanalyse`, **not** the include flow. It also shows the one test holds C1's gating *and* B2's restore at that site. |
| Disable the whole `isSubsetInProgress` branch in the **Include** flow (`generation.tsx:542`) | `generation.test.tsx` **83/83 green** — see B4. |

**N3 — what I verified.** `analysis.ts:6658` emits `` `[analysis-subset] subset_in_progress manuscript=${JSON.stringify(manuscriptId)}` ``; the request-received line at `:6525` begins `[analysis-subset] request received`. The test's filter is now `lines.find(l => l.startsWith('[analysis-subset] subset_in_progress'))` (`analysis.request-log.test.ts:554`), which cannot match the request-received prefix — the pass-2 `includes()` bug is genuinely gone, and the assertion is `toBeDefined()` on that line plus a `manuscript=` containment check, so deleting the `console.log` reddens it.

**Local verification at this head** (all run by me, in `C:\Claude\Projects\wt-3202-subset-chapter-mismatch`):

- `npx tsc --noEmit` (root) — exit 0. `npx tsc --noEmit` (`server/`) — exit 0.
- `npx eslint` on the five changed frontend files — exit 0.
- `npx vitest run` (full frontend suite) — **348 files / 5005 tests, all green.** Run deliberately because adding `store.getState().analysis.*` to a view is the exact shape that has broken hand-built test stores here before (#3201); the 13 added `analysis: analysisSlice.reducer` entries cover it, and the four other test files that render these views (`layout.test.tsx`, `routes/index.test.tsx`, `prosody-autotrigger.test.tsx`, `analysing-prosody-toggle.test.tsx`) are unaffected.
- `npx vitest run src/routes/analysis.request-log.test.ts` (server) — 11/11.
- No bare `getState()` survives: the only three call sites are `generation.tsx:426`, `:619`, `:736` and `analysing.tsx:814`, all `store.getState()`; `useStore<RootState>()` is wired at `generation.tsx:203` and `analysing.tsx:135`. The crash this round fixed is real and is gone.
- Merge state is coherent: no conflict markers in either release-notes file; exactly one `#3202` entry in `docs/release-notes-next.md:341` (inside `## 🗣️ Analyzer, script review & manuscript`) and one in `RELEASE_NOTES.md`'s open `# Castwright 1.15.0` section; the N2 edit to `docs/features/archive/32-sticky-analysis.md:185` survived both merges intact. Server production code is untouched since `02c6cd31` (only `analysis.request-log.test.ts` changed), so pass 2's server verification still stands.

---

### 🐞 Correctness bugs

**B3 — B2's fix restores the snapshot but not the subscription, so the analysing-view collision still ends in a permanently-stalled pill — and it now stalls on a main run that is genuinely still running server-side, hiding real progress.**
`src/views/analysing.tsx:814` (capture), `:929-933` (restore), `:961-963` (the re-arm skip), `src/store/analysis-stream-middleware.ts:288-293` and `:297-308`, `src/components/layout.tsx:1428-1436`.

Same scenario pass 2 gave for B2, re-walked against this head:

1. A subset job `[3]` is live server-side (started from Generate's Include/Re-analyse; sticky, survives navigation and client disconnect).
2. The user is on the analysing view with the main run streaming. `analysis.activeStream = {kind:'main', state:'running'}`; the middleware handle is `{m1,'main'}` (opened by the main path's own `applyAnalysisSnapshotTick` at `analysing.tsx:498-510` / `:531-537`).
3. Retry on a failed chapter. `pausedMainForRetry` is true (`:793-794`) → the view aborts **its own** main SSE and sets `analysisStarted` false.
4. `priorSnapshot` is captured (`:814`), then the optimistic `setActiveStream({kind:'subset',[5]})` fires (`:815-830`). The middleware's displacement guard sees a `main → subset` kind change and calls `closeHandle()` (`analysis-stream-middleware.ts:290-292`) — **before the POST**, so nothing in the `.catch` can save it.
5. The POST is rejected. The new code restores `priorSnapshot` (`:929-933`) and `subsetInProgressRef` suppresses the re-arm (`:961-963`).

End state: `analysis.activeStream = {kind:'main', state:'running', lastTickAt: T₀}` where T₀ is the last tick *before* the click. The client now holds **no subscription to anything** — its own main SSE is aborted, the middleware handle is `null`, and it cannot reopen: `openHandle` is reachable only from `APPLY_TICK_TYPE` (`:297-308`) and nothing in this client will produce a tick. `layout.tsx:1428-1429` computes `stalled = streamState === 'running' && Date.now() - lastTickAt > STALL_THRESHOLD_MS` (30 s, `chapters-slice.ts:36`), so the top-bar pill flips to **stalled** 30 seconds later and stays there.

The restore therefore changes *which* job the permanently-stalled pill names (main instead of subset `[5]`) without removing the symptom pass 2 raised. It is arguably worse in one respect: the main job **is** still running. `analysis.ts:3480-3485` is explicit — *"Sticky: do NOT abort the controller on subscriber disconnect… Only `/pause` or a `fresh: true` displacement aborts"* — so step 3's client-side abort drops a subscriber and nothing else. Real, ongoing analysis progress is now reported to the user as "stalled", while the view beside it offers "Resume analysis" (`analysing.tsx:1257-1266`). Pill and view disagree about the same run.

Mitigant, stated so severity isn't inflated: clicking **Resume** self-heals it — the analysis effect dispatches its own `setActiveStream({kind:'main'})` and re-POSTs, which takes the server's subscribe path onto the still-live sticky job. So this is a wrong indicator plus a manual recovery step, not lost work.

The cheap fix is one property, not a redesign: restore with `state: 'paused'` when `pausedMainForRetry` was true. `AnalysisStreamSnapshot.state` already has `'paused'` (`analysis-slice.ts:63`) and `layout.tsx:1430-1431` already renders it as a paused pill rather than running-then-stalled, which is exactly what the view itself is showing. Note the obvious-looking alternative is wrong: dispatching `analysisActions.setPaused` instead would hit the middleware's `PAUSE_TYPE` branch (`analysis-stream-middleware.ts:258-267`), which POSTs `/pause` — and `/pause` aborts main **and** the other subset for the manuscript, killing the very job this whole PR exists to protect.

**B4 (house gate) — this round's own B2 fix is held by a test at exactly one of the three sites it touches. Deleting it at the other two leaves the suite green, which is the same gate B1 invoked.**
Unheld: `src/views/generation.tsx:542-554` (Include flow) and `src/views/analysing.tsx:929-933` (Retry flow). Held: `src/views/generation.tsx:708-716` (Re-analyse flow), by `generation.test.tsx:2185`.

Measured, not inferred (table above): deleting the `priorSnapshot` restore at **all three** sites reddens exactly **one** test — the `handleReanalyse` one. And disabling the Include flow's entire `if (isSubsetInProgress)` branch — C1's gating *and* B2's restore together — leaves `generation.test.tsx` at **83/83 green**. The reason is that *nothing* exercises the include flow's `subset_in_progress` path: despite living in the `Include in book (subset re-analysis)` describe block, **both** `#3202` frontend tests there (`:2153` and `:2185`) click `chapter-row-1-reanalyse` and route through `handleReanalyse`. The include path's own rejection branch has never had a test.

This is the same "green before, green after" condition that made B1 blocking, applied to the fix B1's round produced. The remedies are near-copies of the test that already works:

- **Include flow**: the `#3215 C1/B2` test's own store and seed, but click the `+ Include in book` button on the excluded chapter-3 row that `makeIncludeStore` already seeds (the trigger every other include test in that block uses, e.g. `:2279`) instead of the Re-analyse confirm, and assert `analysis.activeStream` still equals the seeded `[5]` snapshot. Reverting `generation.tsx:547-551` must redden it.
- **Retry flow**: extend the new `#3215 C2` test — it already drives main into `streaming`, pauses it, and rejects with `subset_in_progress` — with one assertion on `store.getState().analysis.activeStream` after the rejection. That assertion is also what would pin B3's fix, whichever form it takes.

---

### 🧹 Cleanup

**N8 — the test added for N4 does not reach N4's branch; deleting the fallback entirely keeps the suite green.**
`server/src/routes/analysis.request-log.test.ts:564-596` (test `(i2)`), guarding `server/src/routes/analysis.ts:6649`.

The test's own comment says it is *"testing the titlePart empty-fallback path"*. It is not. It seeds `subsetChapterIds: [3, 100]` with a comment that `100` is non-existent — but chapter **3 resolves**, so `runningTitles` is `"Chapter 3"` and `titlePart` takes the populated branch, identical to test `(i)`. The fallback (`runningTitles === ''` → no colon) is still never executed.

Measured: changing `:6649` from ``runningTitles ? `: ${runningTitles}` : ''`` to ``` `: ${runningTitles}` ``` — i.e. deleting the fallback so an unresolvable id list yields the dangling `"…for this manuscript: . Wait for it…"` N4 was raised about — leaves `analysis.request-log.test.ts` at **11/11 green**.

Two contributing causes, both worth fixing rather than just the seed: the assertion regex `/for this manuscript(: .+)?. Wait for it to finish/` makes the colon clause optional *and* leaves `.` unescaped, so it is written to accept either shape; and the fixture can only ever produce one of them. To actually pin the branch, register the live job with **only** unresolvable ids (e.g. `[101, 102]` against a 10-chapter stub) and assert the exact string `for this manuscript. Wait for it` with no colon. The production code is correct — this is a test that documents coverage it does not have, which is the shape N7 was raised to close.

**N9 — the release-notes over-claim N5 flagged is unchanged, and per B3 it is still false for one of the two collision paths.**
`RELEASE_NOTES.md` (1.15.0, `#3202` bullet) and `docs/release-notes-next.md:341`.

`RELEASE_NOTES.md` still says *"the running retry stays visible and on-screen so you can wait for it or cancel it and try again"*, and `release-notes-next` still says the frontend *"(b) leaves the other job's SSE subscription untouched so it remains visible and streaming in the UI"*. Both are true for the Generate-view collision — verified: the handle is `(manuscriptId, kind)`-keyed, the colliding `setActiveStream` keeps both, so no `closeHandle` fires and the restored snapshot keeps ticking. Neither is true for the analysing-view collision, where this client never held a subscription to the other job at all, its main handle is closed by the kind flip, and the pill ends up naming main (B3). Either narrow both sentences to the Generate-view path, or fix B3 and keep them.

**N10 — the PR body's new Summary and Test plan contain four claims the tree does not support.**

- *"Regression tests for C1 (middleware) and C2"* — there is no middleware test. `src/store/analysis-stream-middleware.test.ts` is not in this PR's diff at all (`git diff 46e62a34..HEAD --name-only | grep middleware` returns only the non-test `analysis-stream-middleware.ts`). C1's regression test lives in `generation.test.tsx`. Pass 2's B1 was caused by exactly this kind of unverified test claim in this same body.
- *"`generation.tsx` Include (~line 426), `generation.tsx` Re-analyse (~line 426)"* — the Re-analyse site is `:619` / `:708-716`, not `~426`.
- Test plan counts `82/82` and `73/73` are stale; at this head they are **83** and **74**. Same failure mode as the pass-2 body, whose unchanged counts were the first signal that no tests had been added.
- The body describes neither the `getState` crash this round fixed nor the two `origin/main` merges. CLAUDE.md's house gate is that incidental fixes are *declared* in the body — a `ReferenceError` in shipped view code, fixed in the same round, is the clearest possible case.

---

### Verdict

**Not mergeable as it stands.** C2, N1–N3, N6 and the crash fix are genuinely done and independently verified; the `getState` repair is real and both typechecks are clean. Two things are still open: **B2 is not fully discharged** — the `analysing.tsx` restore fixes which job the pill names without removing the permanently-stalled pill, and it now mis-reports a main run that is still live server-side (**B3**); and **the B2 fix is unheld at two of its three sites** (**B4**), which is the same green-before/green-after condition that made B1 blocking one round ago, now applied to B1's own remedy.

**Tree check:** `git status --porcelain` empty in `C:\Claude\Projects\wt-3202-subset-chapter-mismatch` before and after this pass; `HEAD` unchanged at `8c37c9ede404e8dc55ee87a98fe2b1bfed22bad6`. The five mutations above were applied, measured, reverted with `git checkout --`, and each revert confirmed byte-identical against a copy taken before the first edit. No tracked file modified.

dudarenok-maker:
## PR review — pass 4 (head 0e013b0d, depth high)

Deltas only. Every "verified" below was **measured by mutation** — the specific fix reverted, the suite re-run, the failure message read — reusing pass 3's exact mutations where it named them, so the two that left the suite green last round are directly comparable. Nothing was taken from the PR body or the round summary.

| # | Disposition |
|---|---|
| **B3** — analysing-view restore leaves a permanently-stalled pill on a still-live main run | **VERIFIED RESOLVED.** Mutation-proven, and the prescribed trap (`setPaused`, which would POST `/pause`) was correctly avoided. |
| **B4** — B2's fix held at 1 of 3 sites | **VERIFIED RESOLVED at both previously-unheld sites.** Pass 3's own two green-leaving mutations now both redden. One residual weakness in the Retry-flow assertion — see **N11**. |
| **N8** — `(i2)` never reached the `titlePart` empty-fallback branch | **VERIFIED RESOLVED.** Pass 3's mutation (which left 11/11 green) now reddens with the exact dangling-colon string N4 was raised about. |
| **N9** — release-notes over-claim | **PARTIALLY RESOLVED.** `RELEASE_NOTES.md` is now defensible; `docs/release-notes-next.md`'s clause (b) is unchanged and still false for the analysing-view path. |
| **N10** — four unsupported PR-body claims | **PARTIALLY RESOLVED — 3 of 4 fixed and independently verified; the "middleware test" claim survives verbatim.** |

---

**B3 — what I measured.**

| Mutation | Result |
|---|---|
| Revert `analysing.tsx:934-938` to `dispatch(analysisActions.setActiveStream(priorSnapshot))` | `analysing.test.tsx` **1 failed / 73 passed** — `#3215 C2`, `AssertionError: expected 'running' to be 'paused'` at `:1495`. Red for exactly the claimed reason, and it also proves `pausedMainForRetry` really is `true` in that test (pass 2's B1 found the opposite). |

Mechanism verified independently of the test:

- `activeStream.state === 'paused'` has exactly **two** readers in non-test `src/` — `layout.tsx:1433` (pill state, which short-circuits the `stalled` branch at `:1428-1429`) and `top-bar.tsx:192` (aggregate status → "Paused"). Both are display-only. Nothing dispatches, POSTs, or persists off the value.
- The fix routes through `setActiveStream`, not `setPaused`, so `analysis-stream-middleware.ts:258-267`'s `PAUSE_TYPE` branch — which POSTs `/pause` and would abort the sibling subset job this whole PR exists to protect — is never entered. That was the trap pass 3 named explicitly; it was avoided.
- The restore no longer displaces anything: `SET_ACTIVE_TYPE` at `:288-293` guards on `if (handle && snap)`, and the handle is already `null` by then, so the restore is inert at the middleware. No `/pause`, no subscribe, no re-POST.
- Pill and view now agree. After the collision `conn === 'idle'`, `isAnalysisRunning` is false and `hasStartedOnceRef.current` is true, so `analysing.tsx:1265-1274` renders **"Resume analysis"** while the pill reads **"Paused"** — where pass 3 found "Stalled" next to "Resume". Clicking Resume still re-subscribes to the live sticky job.

One acknowledged residual, not a finding: "Paused" is a white lie about a job the server is still running. It is the semantics the pause-for-retry path already uses (the client abort drops a subscriber; `/pause` is never called), it is what pass 3 prescribed, and it is now self-consistent with the view. Re-litigating it here would be manufacturing.

**B4 — what I measured.** Four mutations, all from inside the worktree at this head:

| Mutation | Pass 3 | Pass 4 |
|---|---|---|
| Delete the `priorSnapshot` restore in the **Include** flow (`generation.tsx:547-551`) | — | `generation.test.tsx` **1 failed / 83 passed** — `B4 — Include flow: a rejected subset_in_progress collision restores the other job's snapshot`, diff `subsetChapterIds: [5] → [3]` and `phaseProgress: 40 → 0`, i.e. the clobber persisting. |
| Disable the **whole** `if (isSubsetInProgress)` branch in the Include flow (`generation.tsx:542`) | **83/83 GREEN** | **1 failed / 83 passed** — same test, `expected null` (the generic tail's `clearActiveStream()` ran). So C1's gating *and* B2's restore are both held at this site now. |
| Delete the entire `if (priorSnapshot) {…} else {…}` restore in the **Retry** flow (`analysing.tsx:929-941`) | *(green — unheld)* | `analysing.test.tsx` **1 failed / 73 passed** — `#3215 C2`, `expected 'running' to be 'paused'` (the clobbered subset snapshot survives). Held. |
| Restore the **clobbered** snapshot with `state: 'paused'` instead of `priorSnapshot` (Retry flow) | — | **74/74 GREEN** — see **N11**. |

The second row is the load-bearing one: it is pass 3's own mutation, verbatim, and it flipped from green to red. The new Include test genuinely routes through `handleInclude` (not `handleReanalyse`, which is what silently held every earlier `#3202` test in that describe block).

**N8 — what I measured.** Pass 3's mutation, verbatim — change `analysis.ts:6649` from ``runningTitles ? `: ${runningTitles}` : ''`` to ``` `: ${runningTitles}` ```:

- Pass 3: **11/11 green.** Pass 4: **1 failed / 10 passed** — `(i2)`, received `"A different subset re-analysis is already in progress for this manuscript: . Wait for it to finish…"`. That is the exact dangling colon N4 was raised about, now caught.

The test is also self-proving at this head: the tightened regex requires `for this manuscript\. Wait` with the `.` escaped and the colon clause no longer optional, so it can only pass on the empty-fallback branch. A fixture that resolved titles would produce `manuscript: Chapter 101.` and redden it. Both of N8's two contributing causes (permissive regex, resolvable fixture) are fixed, not just the fixture.

**Local verification at this head** (all run by me, in `C:\Claude\Projects\wt-3202-subset-chapter-mismatch`):

- `npx vitest run src/views/analysing.test.tsx` — **74/74**.
- `npx vitest run src/views/generation.test.tsx` — **84/84** (158 across both files).
- `npx vitest run src/routes/analysis.request-log.test.ts` (server) — **11/11**.
- `npx vitest run src/routes/analysis.test.ts` (server) — **223/223**.
- `gh pr checks 3215` — all 21 legs green, including the full frontend suite, server fast+slow, and the Windows leg.

---

### 🐞 Correctness bugs

**None.** No new correctness defect found this pass, and both of pass 3's are discharged under mutation.

One thing I probed and could **not** reach, recorded so the record shows it was considered rather than missed: B3's `state: 'paused'` repair is applied only at `analysing.tsx`; the two `generation.tsx` sites restore `priorSnapshot` verbatim. The same false-stall would follow there if `priorSnapshot.kind === 'main'` at the click (the optimistic `kind: 'subset'` write closes the main handle at `analysis-stream-middleware.ts:290-292`, and `openHandle` is reachable only from `APPLY_TICK_TYPE`, which nothing in the Generate view fires for a main run). I could not construct that state: `snapshotInFlightAnalysis` returns the *subset* job when one is live, so cold-boot hydrate seeds `kind: 'subset'`; `handleRetryChapter`'s `.finally` only flips back to `main` after the subset settles; and the cross-tab path carries the subset snapshot through `applyExternalAnalysisSnapshot`, which is a different action type and does not trip the displacement guard. Reported as unreachable, not as a finding — per the brief, a risk I cannot demonstrate is not a finding.

### 🧹 Cleanup

**N11 (new) — the Retry-flow half of B4 is held against *deletion* but not against *substitution*: the assertion reads only `.state`, so restoring the wrong snapshot with the right state keeps the suite green.**
`src/views/analysing.test.tsx:1495`, guarding `src/views/analysing.tsx:934-938`.

Measured, not inferred. Replacing the restore with

```ts
pausedMainForRetry
  ? { ...store.getState().analysis.activeStream!, state: 'paused' }
  : priorSnapshot,
```

— i.e. paused-ifying the **clobbered** subset snapshot (`kind: 'subset'`, `subsetChapterIds: [44]`) instead of restoring the prior main one, which re-arms B2 in full at that site — leaves `analysing.test.tsx` at **74/74 green**. The two `generation.test.tsx` siblings assert `toEqual(otherJobSnapshot)` on the whole object and would catch it; this one asserts `store.getState().analysis.activeStream?.state`.

Pass 3's remedy asked for "one assertion on `store.getState().analysis.activeStream`"; what landed narrows that to one property. The fix is to widen the existing line to the whole object, e.g. `toEqual({ ...capturedMainSnapshot, state: 'paused' })`, matching the shape the other two sites already use. Not blocking — the fix itself is correct and its deletion *is* caught.

**N9 (carried, half-open) — `RELEASE_NOTES.md` is now defensible; `docs/release-notes-next.md`'s clause (b) is not.**
`docs/release-notes-next.md:341`.

- `RELEASE_NOTES.md` — "the running retry stays visible and on-screen so you can wait for it or cancel it and try again" now reads inside its own worked example, *"If you click Re-analyse on chapter 5 while chapter 3 is already retrying"*, which is the Generate-view collision. I re-verified the claim holds there: the handle is `(manuscriptId, kind)`-keyed, the colliding `setActiveStream` changes neither, so no `closeHandle` fires and the restored snapshot keeps ticking. Scoped to the scenario it describes, it is true. Call this discharged.
- `docs/release-notes-next.md:341` — "*the frontend now … (b) leaves the other job's SSE subscription untouched so it remains visible and streaming in the UI*" carries no such scoping. In the analysing-view Retry collision this client never held a subscription to the other job, its own main handle is closed by the `main → subset` kind flip before the POST, and the end state is a **paused main** pill — the other job is visible nowhere. B3 removed the false *stalled* alarm; it did not make the other job visible, which is what this sentence promises.

One clause, e.g. "…leaves a Generate-view collision's other job untouched so it remains visible and streaming; an analysing-view collision leaves the main run paused and re-attachable via Resume." Cost: a tracked-file commit and a CI round.

**N10 (carried, one of four still open) — the "middleware" regression-test claim survives verbatim in the Test plan.**

Re-checked each of pass 3's four bullets against the tree:

- ❌ *"Regression tests for C1 (**middleware**), C2 (re-arm), B2 (snapshot restore), B3 (paused state), B4 (Include + Retry): written and verified to redden when fixes are temporarily reverted."* — `git diff main...0e013b0d --name-only` returns ten files; `src/store/analysis-stream-middleware.test.ts` is not among them, and the file is untouched on this branch. C1's regression test lives in `generation.test.tsx`. This is the third consecutive round in which this body has carried a test claim the tree does not support, and it is the exact signal that surfaced B1 in pass 2.
- ✅ Line numbers — `generation.tsx` Include **547-551** and Re-analyse **708-716** are both exactly right at this head. (`analysing.tsx` Retry is cited as **929-933**; the block is 929-941 now that B3 widened it. Trivial.)
- ✅ Counts — **84/84** and **74/74** both measured and correct; **223/223** for `analysis.test.ts` also measured and correct.
- ✅ The `getState` crash fix is declared (Summary item 1).

The open half costs one `gh pr edit` — no push, no CI, no tracked file. Deleting the word `(middleware)` or re-pointing it at `generation.test.tsx` closes it.

---

### Verdict

**Mergeable.** All three of pass 3's open correctness/house-gate items — **B3**, **B4**, **N8** — are discharged and mutation-proven, including the two mutations that left the suite green one round ago and now redden. No correctness bug found this pass.

Triage for the three remaining cleanup items, since this is the final round:

- **N10's "(middleware)" claim — fix before merge.** It is a PR-body edit, not a code change: no commit, no push, no CI. A body that names a test file the PR does not contain is the precise artefact that hid B1 for a round.
- **N9's `release-notes-next.md` clause — follow-up.** Real inaccuracy, but it needs a tracked-file commit and a CI round to correct one sentence in a technical register; not worth holding a green PR.
- **N11 — follow-up.** The fix is correct and its deletion is caught; only the substitution case slips through.

**Tree check:** `git status --porcelain` empty in `C:\Claude\Projects\wt-3202-subset-chapter-mismatch` before and after this pass; `HEAD` unchanged at `0e013b0de924b25fbac7fd984a6469b6e98ceaf4`. Six mutations were applied, measured, reverted with `git checkout --`, and every one of the four touched files confirmed **byte-identical** afterwards by `git hash-object` against blobs captured before the first edit (`c5636486`, `febb361f`, `d26b5cc7`, plus the three test files unmodified throughout). No tracked file modified.

Issue #3116, opened 2026-09-08
ops — enforce the live-view publish token: fail a PR that changes the register's live view without re-stamping


dudarenok-maker:
## What

Nothing mechanically requires `npm run stamp:publish-token` when a PR changes
`docs/testing/onbox-acceptance-register-live-view.html`. The register's own
step 0 mandates it; no check enforces it. Add one.

## Why now — it just cost a real diagnosis

PR #3076 folded rows A37/B2/D1 out of the register and rewrote A16/A23, then
merged **without stamping and without republishing**. The result (#3114, fixed
by PR #3115):

- the artifact sat three discharged rows behind `main` — owed **55** against
  main's **52** — for a day;
- **both** the tracked file and the published page read
  `data-published-as="13" data-publish-id="a3ef6817d22b"` while their contents
  had diverged.

That second point is the damage. The register documents a one-line
disambiguator — *published HIGHER than tracked ⇒ another lane is AHEAD; EQUAL
or LOWER ⇒ the live page is BEHIND* — and **EQUAL is a fallthrough, not a
measurement**. An unstamped edit does not make the counter wrong, it makes it
uninformative, and the next operator falls back to the row-by-row byte-diffing
that #2599/A41 built the token to replace. The register's own text already
concedes the gap: *"this works even though nothing yet enforces the token
(#2599)"* — #2599 is closed and the enforcement never landed.

The cost lands on whoever publishes **next**, not on the lane that skipped the
stamp, which is why prose alone has not held.

## The decision this needs — three defensible shapes

Filing rather than fixing because the mechanism forks, and the fork changes
behaviour operators depend on.

**Where the baseline comes from.** `check:onbox-register` with no flag
deliberately compares only the two tracked files and needs no network — the
register documents that, and `onbox-register-check.yml` runs it on a
default-depth `actions/checkout@v7` with no base ref available. Options:

1. **A new opt-in flag** (`--stamped-since <ref>`), invoked from the workflow
   with the PR base after a shallow fetch of it. Keeps the no-flag run offline
   and unchanged. Costs a workflow step and does nothing locally.
2. **Fold it into the no-flag run**, fetching `origin/main` the way
   `--against-published` already does. Catches it locally too, but makes the
   default path network-dependent — a real change for offline work, and the
   no-flag run is a required-adjacent check.
3. **Enforce in the workflow only**, comparing via `git diff` against the base
   without touching the script.

[...]

Issue #3221, opened 2026-09-11
[agent instructions][claude][verify] Verify on-box register CI-fold fix and open PR closing #3138


dudarenok-maker:
## What

Verify the CI-fold fix from the previous `[claude][task]` child in this chain, then open the PR closing #3138.

## Parent's acceptance, restated as a checklist

1. `.github/workflows/onbox-register-check.yml` no longer exists on this branch.
2. `.github/workflows/verify.yml`'s `lint-and-checks` job has three new steps (the two unconditional register checks, plus the PR-only `--stamped-since HEAD^1` one) gated on a `step_check_onbox_register || shared` condition.
3. `scripts/verify-cache.mjs` has a new `STEPS[]` entry named `check:onbox-register` with `globs`/`extraFiles` mirroring the deleted workflow's path filter.
4. `node --test scripts/tests/verify-cache.test.mjs` passes, including the "every STEPS[] entry is covered by a local --steps CSV or explicitly allowlisted" test.
5. `npm run lint` and `npm run typecheck` pass.
6. **Re-run the mutation, don't trust the receipt:** temporarily rename/comment out the new `STEPS[]` entry (or break its `globs`) and confirm `node --test scripts/tests/verify-cache.test.mjs` goes RED with the coverage test specifically. Then restore it and confirm green again. Report the actual observed failure text, not just "it failed."
7. Commit message follows `<type>(<scope>): <subject>`.

You are **read-only for source** beyond the mutation probe above (which you must revert) — report, do not permanently fix.

## Where the work happens

Worktree: `C:\Claude\Projects\wt-3138-onbox-register-required-check`
Branch: `chore/ops-3138-onbox-register-required-check`

Do NOT create another worktree. Read the actual committed diff (`git diff main`, `git log`) — not the issue description.

## On FAIL (any acceptance item not met)

In this exact order:

1. Create a rework `[agent instructions][claude][task]` child describing precisely what's wrong (cite line numbers / diff hunks / the exact failing test output).
2. Create a fresh `[agent instructions][claude][verify]` child (copy of this brief).
3. Link both as sub-issues of #3138.
4. **Only then** close this issue, stating the verdict (FAIL) and why.

## On PASS

Open the PR yourself:

- Base: `main`. Head: `chore/ops-3138-onbox-register-required-check`.
- Title matching commit-convention subject format, e.g. `chore(ci): fold on-box register checks into the required verify.yml job`.
[...]

PR #3229, opened 2026-09-11
chore(ci): fold on-box register checks into the required verify.yml job

### LINKED ISSUE #3138: ops — the on-box register check is not a required status check, so a red publish-token gate cannot block a merge
## What

`main`'s ruleset (`17654264`) requires exactly three contexts:

```
npm run verify
Verify PR body links a GitHub issue
Verify sidecar acceptance is recorded
```

**`Check on-box acceptance register consistency` is not among them** — verified
against the live ruleset, 2026-09-10. That job carries `check:onbox-register`,
`register:build --check`, and (as of PR #3133) the `--stamped-since`
publish-token gate. All three go red without blocking the merge button.

Both incidents #3116 was filed for were **merged-anyway** incidents: #3076 and
#3061 each merged with the register in a state the check would have objected to.
Replaying #3061 with #3133 merged gives a red job and an enabled merge button.

## The decision this needs

Adding the context to the ruleset is **not** a one-line change.
`onbox-register-check.yml` is `paths:`-filtered, and a path-filtered required
check never reports on a PR that doesn't touch those paths — it sits permanently
pending and blocks every unrelated PR. That is the same trap
`docs/superpowers/specs/2026-07-06-verify-ci-rebalance-design.md` records for
`verify.yml`, which is why that workflow **dropped `paths-ignore`** and instead
runs always with per-leg scope conditions, reporting green in seconds on a
docs-only diff.

Three shapes, all defensible:

1. **Mirror `verify.yml`'s pattern** — drop the `paths:` filter, keep the job
   always-running, and make each step internally scope-gated so an unrelated PR
   reports green immediately. Then add the context to the ruleset. Most
   consistent with the repo's existing solution; costs a job spin-up per PR.
2. **Fold the register checks into `verify.yml`** as a scope-gated leg, so they
   inherit the already-required `npm run verify` context. No new required
   context, no pending-check trap; couples register bookkeeping to the big
   battery and slows its critical path slightly.
3. **Accept advisory-only** and stop describing it as enforcement. Cheapest, and
   honest — but it leaves the merged-anyway shape exactly as it is, which is the
   thing #3116 existed to close.

## Not a defect in PR #3133

The `--stamped-since` gate itself works and is proven in real CI. This issue is
about the *enforcement wiring around it*, which predates #3133 and applies
equally to the two register checks that were already in that job. PR #3133's
description has been corrected to say the check reports rather than blocks, so
nothing in the tree overclaims while this is open.

Found by the `pr-review-gate` pass on #3133.

Refs #3116

dudarenok-maker:
## Summary

Folds the three register checks into the already-required `verify.yml` `lint-and-checks` job per #3138's decision — no ruleset edit needed. Deletes the redundant, path-filtered `onbox-register-check.yml`, which could never become a required status check and let the register go red without blocking a merge (#3076, #3061).

Closes #3138
Refs #3116

## Test plan

Verified against the parent's 7-item acceptance checklist:

- [x] `.github/workflows/onbox-register-check.yml` deleted
- [x] `verify.yml`'s `lint-and-checks` job has the three new steps (two unconditional register checks, plus the PR-only `--stamped-since HEAD^1` one), gated on `step_check_onbox_register || shared`
- [x] `scripts/verify-cache.mjs` has a new `check:onbox-register` `STEPS[]` entry with `globs` mirroring the deleted workflow's path filter exactly
- [x] `node --test scripts/tests/verify-cache.test.mjs` passes — 142/142, including the coverage test
- [x] `npm run lint` and `npm run typecheck` pass
- [x] Mutation probe re-run (not trusted from the receipt): renamed the new `STEPS[]` entry to `check:onbox-register-MUTATED-PROBE`, re-ran the suite — went RED with `AssertionError [ERR_ASSERTION]: STEPS[] entr(y/ies) absent from every local --steps CSV and not allowlisted as cloud/full-verify-only: check:onbox-register-MUTATED-PROBE`, 1 failing test, exit 1. Restored the entry and confirmed green again — 142/142, exit 0.
- [x] Commit message follows `<type>(<scope>): <subject>`

Once CI runs on this PR, will confirm `npm run verify`'s new register-check steps actually execute (the diff touches `scripts/verify-cache.mjs`/`.github/workflows/verify.yml`, which should trigger `shared` scope).


dudarenok-maker:
Closing without merge: this branch does **not** actually pass full CI, despite passing the narrower local checks in the parent's acceptance list (verify-cache.test.mjs, lint, typecheck). I opened this PR before letting the full CI suite finish — CI's own "Lint, typecheck & fast checks" job (the `Hooks tests` step, i.e. `npm run test:hooks`) fails with two real defects:

1. `scripts/tests/onbox-register-check-workflow-wiring.test.mjs` throws `ENOENT` reading `.github/workflows/onbox-register-check.yml` — that file was deleted by this branch, but this test (which pins the fetch-depth:2 / `--stamped-since HEAD^1` fix from #3116) was never ported to target `verify.yml` instead.
2. `workflow-wiring.test.mjs`'s "every derived step carries a parseable `# leg:`/`# supports:` marker" test fails: only the first of the three new register-check steps in `verify.yml` (`Check on-box register consistency`, line ~227) has a `# leg: check:onbox-register` marker; the other two (`Check the generated live-view surfaces are up to date` at line 233, and `Check the live view was re-stamped if its content changed` at line 243) have none.

See #3221 for the full verify verdict and the rework/re-verify children opened against #3138.

